Issue #936, opened 2016-01-29
Reduce 1_notmnist.ipynb memory requirements.


vincentvanhoucke:
Lots of people are having trouble downloading and pickling the data. need to look into ways to make that easier.

bhack:
Please Vincent add also an intermediate pickle of the dataset. The first cell where train and test is formed is slow to run every time you break your notebook session.

vincentvanhoucke:
@bhack under which conditions do you 'lose' the pickle? Is it because you're using an ephemeral docker container?

bhack:
I don't lose the pickle but I think it is better to save also one at an intermediate point. What happens if people doesn't run the pickle step? You have safe condition  for not run again the download but some cell are also heavy so it is safer to save train and test set as soon as formed.

vincentvanhoucke:
@rthouvenin your approach is a net improvement over the status quo.
Would you be ok cleaning it up (in a way that can be easily diffed) and sending me a PR?
https://github.com/rthouvenin/tensorflow/blob/notmnist-lowmem/tensorflow/examples/udacity/1_notmnist.ipynb

rthouvenin:
Sure, just opened the PR (I got what you meant by easily diffed, notebook doesn't play nicely on that aspect... :)).

Though I did not have trouble to download the files (it just took time), I was also a bit scared it would be difficult as my connection is quite erratic sometimes. I don't think there is much to be done, but I would suggest warning the reader about it, and suggest to download the file separately if they have a slow / not reliable connection (e.g. with a download manager that support pause/resume). 
To help with that, maybe a torrent link could be set up (I didn't check if there is already one).

bhack:
@vincentvanhouck Ok the intermediate pickling is introduced in @rthouvenin PR.

PR #940, opened 2016-01-30
Notmnist lowmem

rthouvenin:
As explained in this [Udacity thread](https://discussions.udacity.com/t/deep-learning-notmnist-data-curation/45051) and as requested by Vincent in Issue #936, I am issuing this PR to update the way the notMNIST dataset is loaded into memory, so that it requires less memory.

The idea is to load the dataset in pieces, curate each piece separately, and merge them.


rthouvenin:
I signed it!

tensorflow/examples/udacity/1_notmnist.ipynb
```diff
@@ -283,83 +283,140 @@
...
+        "    \n",
+        "    print('Full dataset tensor:', dataset.shape)\n",
+        "    print('Mean:', np.mean(dataset))\n",
+        "    print('Standard deviation:', np.std(dataset))\n",
+        "    return dataset\n",
+        "        \n",
+        "def load(data_folders, min_num_images, num_classes):\n",
+        "  dataset_names = []\n",
+        "  for folder in data_folders:\n",
+        "    dataset = load_letter(folder, min_num_images / num_classes)\n",
```
vincentvanhoucke: For clarity, I would slightly prefer load() to be passed a min_num_images_per_class, to be forwarded directly to load_letter()

tensorflow/examples/udacity/1_notmnist.ipynb
```diff
@@ -496,22 +502,61 @@
         "outputId": "8af66da6-902d-4719-bedc-7c9fb7ae7948"
       },
       "source": [
+        "def make_arrays(nb_rows, img_size):\n",
+        "    if nb_rows:\n",
+        "        dataset = np.ndarray((nb_rows, img_size, img_size), dtype=np.float32)\n",
+        "        labels = np.ndarray(nb_rows, dtype=np.int32)\n",
+        "    else:\n",
```
vincentvanhoucke: Why do you require this else clause here? I would prefer simpler.

tensorflow/examples/udacity/1_notmnist.ipynb
```diff
@@ -496,22 +502,61 @@
...
+        "        dataset, labels = None, None\n",
+        "    return dataset, labels\n",
+        "\n",
+        "def merge_datasets(pickle_files, train_size, valid_size=0):\n",
```
vincentvanhoucke: default to zero goes against my pestering that having a validation set is a good thing :) No default is fine.

tensorflow/examples/udacity/1_notmnist.ipynb
```diff
@@ -496,22 +502,61 @@
...
+        "    num_classes = len(pickle_files)\n",
+        "    valid_dataset, valid_labels = make_arrays(valid_size, image_size)\n",
+        "    train_dataset, train_labels = make_arrays(train_size, image_size)\n",
+        "    vsize_per_class = int(valid_size / num_classes)\n",
```
vincentvanhoucke: I think this should be
vsize_per_class = valid_size // num_classes
for Python 3 compatibility.
Same below.

tensorflow/examples/udacity/1_notmnist.ipynb
```diff
@@ -496,22 +502,61 @@
...
-        "print('Validation', valid_dataset.shape, valid_labels.shape)"
+        "print('Training:', train_dataset.shape, train_labels.shape)\n",
+        "print('Validation:', valid_dataset.shape, valid_labels.shape)\n",
+        "print('Testing:', test_dataset.shape, test_labels.shape)"
       ],
       "outputs": [
         {
           "output_type": "stream",
           "text": [
             "Training (200000, 28, 28) (200000,)\n",
```
vincentvanhoucke: Have you verified that subsequent assignments run well and give sensible results with this default training set size?

tensorflow/examples/udacity/1_notmnist.ipynb
rthouvenin: This and the test in `make_arrays` is so that the same function can be called for the training data and the test data. I can:

 1. Keep it like this
 2. Remove the default and make it explicit in the call for the test set
 3. Move the size test from `make_array` to `merge_datasets`
 4. Have a different function for the training+validation set and the test set
 5. Something else?

tensorflow/examples/udacity/1_notmnist.ipynb
rthouvenin: The call to `int()` make it Python3-compatible, but you're right it is clearer with //.

tensorflow/examples/udacity/1_notmnist.ipynb
rthouvenin: I haven't had the time yet to do the subsequent assignments. But this default training size is already in the current notebook.

tensorflow/examples/udacity/1_notmnist.ipynb
vincentvanhoucke: You're right, it's simpler this way.

vincentvanhoucke:
Cool, can you wrap the couple of lines that go past the edge and squash all the commits? I'll run a test in my local env this weekend or Monday and pull it. Thanks again!

vincentvanhoucke:
Merged, thanks!

PR #943, opened 2016-01-30
Udacity word2vec notebook: more explicit examples for batch generation algo

cjauvin:
@vincentvanhoucke As I struggled a bit (with pen & paper) to make sense of what's really happening in the `generate_batch` function of the word2vec notebook, I'd like to offer this revision, with examples which make it a bit more explicit IMO. Here's a link to the rendered version of what I'm proposing:

https://github.com/cjauvin/tensorflow/blob/b0b231a7202db3d4b4e31fa361cd1078719b9ba9/tensorflow/examples/udacity/5_word2vec.ipynb


vincentvanhoucke:
Merged. Thanks!

PR #948, opened 2016-02-01
Improved 1_notmnist

shreyasva:
- Added a .gitignore for files created by ipython
- Changed data_folders to have only folders.


tensorflow/examples/udacity/.gitignore
```diff
@@ -0,0 +1,5 @@
+notMNIST.pickle
```
vincentvanhoucke: Note that we now also have one pickle per class.

tensorflow/examples/udacity/1_notmnist.ipynb
```diff
@@ -183,15 +183,17 @@
       "source": [
         "num_classes = 10\n",
         "\n",
-        "def extract(filename):\n",
-        "  tar = tarfile.open(filename)\n",
+        "def maybe_extract(filename):\n",
```
vincentvanhoucke: No, I don't want that behavior. If someone accidentally cancels the decompression in the middle, they have no way to recover cleanly.
If you don't want to re-decompress, simply don't rerun this frame.
Note that the situation is different for the download since we want to avoid DDOSing the server hosting the data, and we have a check for the integrity of the downloaded file to warn if the file is only half downloaded.

tensorflow/examples/udacity/.gitignore
shreyasva: Sorry I have not made my way through the other classes.
So is it the case that it's better to regex this as *.pickle?

tensorflow/examples/udacity/1_notmnist.ipynb
shreyasva: Good catch.
There are multiple ways to solve this but all of them seem like overkill. Removed this logic.

tensorflow/examples/udacity/.gitignore
vincentvanhoucke: No, I mean 1_notmnist has changed and has more pickles now.

vincentvanhoucke: Never mind, they're covered by the *. Looks like you still have a stray edit in the .ipynb.

shreyasva: It's an intended edit :). 
It's trying to change |data_folders| to have only folders, as opposed to the relatively flimsy approach of excluding .DS_Store.

shreyasva: It's an intended edit :)
Changed data_folders to check for folders rather than somewhat flimsy check for .DS_Store.

tensorflow/examples/udacity/1_notmnist.ipynb
```diff
@@ -191,7 +191,8 @@
         "  tar.extractall()\n",
         "  tar.close()\n",
         "  data_folders = [\n",
-        "    os.path.join(root, d) for d in sorted(os.listdir(root)) if d != '.DS_Store']\n",
+        "    os.path.join(root, d) for d in sorted(os.listdir(root))\n",
```
vincentvanhoucke: A similar change should be pushed to github soon. (I couldn't take in PRs for the last few days due to sync issues)

vincentvanhoucke: Pushed. This is no longer needed. Happy to merge the .gitignore change when you revert this.

vincentvanhoucke:
Closing due to inactivity.

PR #951, opened 2016-02-01
fix gradient of tf.floor

### LINKED ISSUE #897: gradient of tf.floor
```python
import numpy as np
import tensorflow as tf
f = np.array([3.8], dtype='float32')
vif = tf.placeholder(tf.float32, shape=[1])

#out = vif + tf.cast(tf.cast(vif, tf.int32), tf.float32)
out = vif + tf.floor(vif)
grad = tf.gradients(tf.reduce_sum(out), vif)

sess = tf.Session()
print sess.run(grad, feed_dict={vif: f})  # print 2

import theano
import theano.tensor as T
vif = T.fvector()
out = vif + T.floor(vif)
grad = T.grad(out.sum(), vif)
func = theano.function([vif], grad)
print func(f) # print 1
```
It looks like the gradient of `tf.floor(x)` w.r.t `x` is 1. But I'm expecting it to be 0, as in theano.
Right now I could use a double casting as a work around, but why is it different?

ppwwyyxx:
Fix #897 .

### DIFF
--- a/tensorflow/python/ops/math_grad.py
+++ b/tensorflow/python/ops/math_grad.py
@@ -488,7 +488,7 @@ def _SparseMatMulGrad(op, grad):
 
 @ops.RegisterGradient("Floor")
 def _FloorGrad(_, grad):
-  return grad
+  return [None]
 
 
 @ops.RegisterGradient("BatchMatMul")


vrv:

girving:
LGTM if tests pass.

vrv:
Merged

PR #963, opened 2016-02-02
Autoload folders and pickle files if already exist (avoid extract)

mtamburrano:
improved `extract(filename)` and `load(data_folders, min_num_images_per_class)` to avoid extractions of .tar.gz if folders already exist and  the creation of already dumped pickle files is also avoided


mtamburrano:
CLA signed

vincentvanhoucke:
I don't want that behavior. It will silently cause irrecoverable issues if the user aborts the tarball extraction in the middle. If you don't want to re-extract, you can simply skip this cell. Note that this is different from the maybe_download() case since it's easy to check for download size.

mtamburrano:
How can you skip these cells if you reopen your notebook in a second moment? The `train_folders` and `train_datasets` variables are not defined until cells execution.
It's true that if the user aborts the extraction in the middle issues can happen, but this is already happening during the dumping of the pickle files where if the user aborts that operation, the pickle files are incomplete.

vincentvanhoucke:
Aha, not having these variables defined is indeed a problem.
There are some changes to this notebook that are about to be pushed to the public github, let me think about it in the meantime, it seems to be a problem for several people. Thanks for clarifying.

mtamburrano:
It would not be enough to write a file after the execution of `extractall` and verify the existence of this file when trying to read from folders? In this way if the user aborts the extraction the file doesn't exist and successive runs will try to extract the files again

vincentvanhoucke:
A fix similar to yours should be pushed shortly (sorry, my tree was not in a state where I could accept PRs easily). Thanks!

Issue #933, opened 2016-01-29
thread pool does not execute all remaining tasks on shutdown


dvyukov:
ThreadPool claims to finish all remaining work on shutdown. This is not the case. If remaining tasks submit new tasks, the new tasks do not get executed. This happens because the sentinel tasks are fetched before the newly submitted real tasks.

ThreadPool must either execute all pending work, or execute nothing on shutdown. It does not make sense to execute half of work.

Here is a repro:
```c
static void BM_ParallelDivide(int iters, const char* impl) {
  THREAD_POOL_IMPL_NAME = impl;
  for (int i = 0; i < iters; i++) {
    const int kTasks = 10;
    const int kLevels = 22;
    std::atomic<unsigned> count(kTasks * (1 << kLevels));
    mutex done_lock;
    condition_variable done;
    bool done_flag = false;
    std::function<void(int)> work;
    ThreadPool pool(Env::Default(), "test", kNumThreads);
    work = [&pool, &work, &count, &done_lock, &done, &done_flag](int level) {
      if (level-- > 0) {
        pool.Schedule([&work, level]() { work(level); });
        pool.Schedule([&work, level]() { work(level); });
        return;
      }
      delay();
    };
    for (int t = 0; t < kTasks; ++t) {
      pool.Schedule([&work]() {
        work(kLevels);
      });
    }
  }
}
```

jeremybarnes:
Personally, I find this behaviour of performing potentially unlimited work in the destructor to be strange.  I think it would make more sense for the caller to explicitly ask for all work to be performed, and for the destructor to finish as fast as possible, even if it meant leaving work in the queue.

dvyukov:
@jeremybarnes I fine either way as long as it is documented. It is just the current implementation that does not make sense.

jeremybarnes:
Note that there are some places in the existing codebase (and tests) that assume this behavior.  I had to implement it in my PR in order to avoid deadlocks.

girving:
@dvyukov, @rmlarsen: Is this resolved?

dvyukov:
Yes, it's fixed. New thread pool waits for completion of all work.

Issue #932, opened 2016-01-29
thread pool deadlocks on shutdown


dvyukov:
ThreadPool dtor does not pop waiters from waiters_ list. As the result dead waiters are left on the list. If remaining tasks submit new tasks, thread pool deadlocks because some notifications are consumed by the leftover dead waiters instead of alive threads that should receive the notifications.

Here is a simple test that does classical parallel decomposition and reliably deadlocks:
```c
static void BM_ParallelDivide(int iters, const char* impl) {
  THREAD_POOL_IMPL_NAME = impl;
  for (int i = 0; i < iters; i++) {
    const int kTasks = 10;
    const int kLevels = 22;
    std::atomic<unsigned> count(kTasks * (1 << kLevels));
    mutex done_lock;
    condition_variable done;
    bool done_flag = false;
    std::function<void(int)> work;
    ThreadPool pool(Env::Default(), "test", kNumThreads);
    work = [&pool, &work, &count, &done_lock, &done, &done_flag](int level) {
      if (level-- > 0) {
        pool.Schedule([&work, level]() { work(level); });
        pool.Schedule([&work, level]() { work(level); });
        return;
      }
      delay();
    };
    for (int t = 0; t < kTasks; ++t) {
      pool.Schedule([&work]() {
        work(kLevels);
      });
    }
  }
}
```

girving:
@zheng-xq: Do you have thoughts on this?  The non-blocking thread pool PR is problematic since it breaks Eigen's FIFO requirement, but the deadlock issue seems like something we should fix.

dvyukov:
@zheng-xq @girving 

On Tue, Mar 8, 2016 at 8:08 PM, Zhifeng Chen wrote:
> The thread calling 'delete p' should have the sole ownership of p, where p is generally any c++ object. 
> So, it's not valid if one thread is calling a threadpool p while another thread calling p->Add().

That's true. But the thread pool waits for the existing tasks to finish in destructor. If you combine these two points, you conclude that tasks are not allowed to submit child tasks to the thread pool. Ever. That's just does not make sense. So you need to either (1) wait for _all_ tasks to finish, including subtasks of existing tasks, or (2) don't wait in thread pool destructor at all, stop threads as soon as possible, assert that the queue is empty now and exit.

I am working on a fix.

girving:
@dvyukov, @rmlarsen: Did the recent thread pool changes resolve this?

girving:
[...]

PR #967, opened 2016-02-02
add distributed, non-blocking thread pool implementation

### LINKED ISSUE #551: Lock contention in thread pool for CPU-based tensor operations
Running the December 15 Inception model, with either serial or parallel inference steps, is showing serious lock contention on the thread pool lock, with about 50% of the CPU being spent in futex operations:

jeremy_khan____projects_mldb

As an experiment, replacing the mutex with a basic spinlock reduces latency by around 35% (whilst obviously increasing CPU usage due to all of the CPUs constantly spinning even with nothing to do).  Obviously a pure spinlock is not a viable solution, but it shows that there is some scope for improvement.

This leads to three related questions:

1.  Would it be possible to allow the platform to provide its own thread pool implementation or factory?  The ThreadPool methods are virtual, so it would be possible to make it pluggable, except that the class is directly created wherever used.  In our case we have a separate thread pool that has a broadly compatible interface and we'd like to share with TensorFlow.
2.  Would it make sense to reduce the granularity of the Eigen tensor operation multithreading?
3.  If not, is there any interest in a lock free thread pool implementation?  Or is this something that somebody is already working on?

### LINKED ISSUE #583: Multiple CPU usage ineffective
I'm running tensorflow on a machine with 64 CPUs and no GPUs. I noticed that although tf detects the number of CPU's correctly (
I tensorflow/core/common_runtime/local_device.cc:40] Local device intra op parallelism threads: 64
I tensorflow/core/common_runtime/direct_session.cc:58] Direct session inter op parallelism threads: 64
), the total CPU usage is less than 16 most of the time. It increases to 20-30 occasionally but for a very short period of time. And even that is much lower than 64.

dvyukov:
Current thread pool implementation is centralized and non-scalable.
This change adds distributed, non-blocking thread pool implementation.
Both implementations co-exist and can be chosen using
TF_THREAD_POOL env var.

Fixes #551
Fixes #583
Update #932
Update #933

### DIFF
--- a/tensorflow/core/lib/core/eventcount.h
+++ b/tensorflow/core/lib/core/eventcount.h
@@ -1,0 +1,170 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//==============================================================================
+
+#ifndef TENSORFLOW_LIB_CORE_EVENTCOUNT_H_
+#define TENSORFLOW_LIB_CORE_EVENTCOUNT_H_
+
+#include <atomic>
+#include <vector>
+
+#include "tensorflow/core/platform/logging.h"
+#include "tensorflow/core/platform/macros.h"
+#include "tensorflow/core/platform/mutex.h"
+
+namespace tensorflow {
+namespace thread {
+
+// EventCount allows to wait for arbitrary predicates in non-blocking
+// algorithms. Think of condition variable, but wait predicate does not need to
+// be protected by a mutex. Usage:
+// Waiting thread does:
+//
+//   if (predicate)
+//     return act();
+//   EventCount::Waiter w;
+//   ec.Prewait(&w);
+//   if (predicate)
+//     return act();
+//   ec.Wait(&w);
+//
+// Notifying thread does:
+//
+//   predicate = true;
+//   ec.Notify(&count, true);
+//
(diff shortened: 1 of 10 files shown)


dvyukov:
I've been told that Eigen library requires thread pool to execute tasks in a strict FIFO order. This thread pool does not execute tasks in FIFO order. So I am not sure how to proceed.

I am also not sure about benchmarks. I did not find any. On various tests that I executed this thread pool showed 2-6x speedup.

@jeremybarnes

martinwicke:
@benoitsteiner, can you answer this question or reassign?

dvyukov:
We could use a wrapper along the following lines to ensure FIFO ordering on top of a scheduler that does not give any ordering guarantees. I've tried to fix deadlocks in label_image example by applying the FIFO wrapper in TensorContractionThreadPool, but no success (even if I use FIFO for all tasks submitted in TensorContractionThreadPool, it still deadlocks).

```c
class FIFO {
 public:
  typedef std::function<void()> F;

  F operator()(F f) {
    {
      std::unique_lock<std::mutex> lock(m_mutex);
      m_queue.push(std::move(f));
    }
    return [this]() {
      F f;
      {
        std::unique_lock<std::mutex> lock(m_mutex);
        f = std::move(m_queue.front());
        m_queue.pop();
      }
      f();
    };
  }

 private:
  std::mutex m_mutex;
  std::queue<F> m_queue;
};
```

jeremybarnes:
It seems fundamentally wrong to require ordering of operations that are submitted to a thread pool, since even if the jobs are scheduled on a thread in order, they won't necessarily start or complete in any order.  The only thing that we can really guarantee is that at least one job from n-1 to n-numThreads will complete before job n is scheduled.  I wonder if the FIFO-ness is really the issue here.

dvyukov:
Yes, FIFO is the issue.
Consider you have a thread pool with N threads and 2*N tasks where first N tasks wait for completion of the last N tasks. If you start running last tasks first, everything is OK. If you start running first tasks, you deadlock.

girving:
@benoitsteiner: Can you comment on this?

vincentvanhoucke:
As of change `ab02c5`, we consider the CPU scaling issue generally resolved, so I'm going to close this PR. Please file new issues with benchmarks and steps to reproduce if you still see any performance issues.

PR #973, opened 2016-02-03
Fix wrong variable name (pickle_name) in 1_notmnist.ipynb

sasbluesea:
When raise Exception at line 327, 'pickle_file' at line 328 is not predefined and raise error.

NameError: global name 'pickle_file' is not defined


sasbluesea:
I signed it!

PR #974, opened 2016-02-03
Fix wrong variable name (pickle_name) in 1_notmnist.ipynb

sasbluesea:
When raise Exception at line 327, 'pickle_file' is not predefined and raise error.

`
NameError: global name 'pickle_file' is not defined
`


vincentvanhoucke:
Thanks for pointing it out. I've actually fixed it upstream and the fix should propagate soon. (The tree wasn't in a state I could simply pull a PR at the time.)

Issue #736, opened 2016-01-09
Build tools and versions listed in Jenkins build log.


EricGT:
Since TensorFlow is built regularly and it success it reported on the GitHub page, it would be nice to see in the logs e.g. (http://ci.tensorflow.org/job/tensorflow-master/47/configure) the OS configuration, tools used and their version. This way if people want to know what tools and version of the tools currently work they would be available and automatically updated.

Example:
OS: Ubuntu 14.04 LTS
Memory: 2G
Swap space: 2G
Processor: i7
Free hard disk space: 2G
Javac: Oracle Javac 1.8.0_66
Python: 2.7
g++: 4.9
Bazel: 0.1.1
SWIG: 2.0.11
TensorFlow:  f9514a917265f0e98c8fb44abc51158685f72ca6

martinwicke:
We'll look into modifying the build scripts to that information lands in the logs.

EricGT:
Thanks. I may have more details as I find other tools/options that effect the success of the builds.

EricGT:
Please add the Bazel --local_resouces configuration to the list as it is significant:
 --local_resouces: http://bazel.io/docs/bazel-user-manual.html#flag--local_resources
Results of changing values: http://stackoverflow.com/a/34770965/1243762

martinwicke:
@caisq can you add some output to the scripts? I think they're all run through configured, so maybe that's the right place to put it.

caisq:
Please close the issue. See: https://github.com/tensorflow/tensorflow/pull/979.

Jenkins builds now shows descriptions that contain JSON objects with info about the build, including build container type, command, platform/OS, build tools and source version. For example: 
http://ci.tensorflow.org/job/tensorflow-master/171/

jendap:
I have removed the machine configuration from build descriptions. It was making too much noise. The description stays in the output log.

@EricGT what is your use case to ask for this? Note you are able to replicate any of the linux builds trivially. The only thing you need is docker. See [ci_build/README.md](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/README.md) These scripts can be used for great dev setup. I will update the documentation how to do it on windows and mac as well.

[...]

PR #979, opened 2016-02-03
Adding build info print-out for Jenkins Description Setter

caisq:
This is addressing issue https://github.com/tensorflow/tensorflow/issues/736

The print-out JSON object line can be used by Jenkins Description Setter to extract and display information about the build, including source version, platform and build tools. See example at: 

http://ci.tensorflow.org/view/Experimental/job/experimental-cais-tensorflow-cpu-python27-copt_pip_install-test/29/

### DIFF
--- a/tensorflow/tools/ci_build/builds/print_build_info.sh
+++ b/tensorflow/tools/ci_build/builds/print_build_info.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+
+# Print build info, including info related to the machine, OS, build tools
+# and TensorFlow source code. This can be used by build tools such as Jenkins.
+# All info is printed on a single line, in JSON format, to workaround the
+# limitation of Jenkins Description Setter Plugin'sthat multi-line regex is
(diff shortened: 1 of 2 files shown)


martinwicke:
Can you add the bazel --local_resources setting? I'm not sure how to get it out of bazel, but since we set it, we should know anyway.

caisq:
Added container type and command, which cover @martinwicke's comment about local_resources.

martinwicke:
Jenkins, test this please.

martinwicke:
Oops. Test this please.

martinwicke:
Can you squash?

tensorflow/tools/ci_build/builds/print_build_info.sh
```diff
@@ -0,0 +1,94 @@
...
+#     e.g.,
+#       print_build_info.sh GPU bazel test -c opt --config=cuda //tensorflow/...
+
+# Information about the command
+CONTAINER_TYPE=$1
+shift 1
+COMMAND=("$@")
+
+# Information about machine and OS
+OS=`uname`
```
jendap: You should change all the way the subshell is run - see https://github.com/koalaman/shellcheck/wiki/SC2006

(because shellcheck is eventually running on the scirpts)

tensorflow/tools/ci_build/builds/print_build_info.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# NVIDIA & CUDA info
+NVIDIA_DRIVER_VER=""
+if [[ -f /proc/driver/nvidia/version ]]; then
+  NVIDIA_DRIVER_VER=`head -1 /proc/driver/nvidia/version | awk '{print $(NF-6)}'`
+fi
+
+CUDA_DEVICE_COUNT="0"
+CUDA_DEVICE_NAMES=""
+if [[ ! -z `which nvidia-debugdump` ]]; then
```
jendap: "if which nvidia-debugdump >/dev/null; then" does not print error when the command is not found

martinwicke:
Squashed and merged.

tensorflow/tools/ci_build/builds/print_build_info.sh
caisq: @jendap It got merged before I had a chance to address your comments. I'll submit a separate PR for those.

PR #984, opened 2016-02-04
Shuffling letters before assigning them to training and validation datasets

lukas-krecan:


lukas-krecan:
I signed it!

vincentvanhoucke:
Sounds like a good idea. Can you:
- replace "lets" with "let's"
- move the call to np.random.seed(133) to the top of the file, so that the seed is fixed for everything
- add a:
valid_dataset, valid_labels = randomize(valid_dataset, valid_labels) where the other calls are

I think that would take care of all randomization and reproducibility issues.

lukas-krecan:
Sure, I just do not get what do you mean by the third bullet point "add a: valid_dataset, valid_labels = randomize(valid_dataset, valid_labels) where the other calls are"

vincentvanhoucke:
We don't appear to be re-randomizing the validation set here:
train_dataset, train_labels = randomize(train_dataset, train_labels)
test_dataset, test_labels = randomize(test_dataset, test_labels)

lukas-krecan:
I see. Should be OK now

vincentvanhoucke:
Merged. Thanks!

PR #981, opened 2016-02-03
Add FillFunctor definition for bool

martinwicke:

PR #986, opened 2016-02-04
Fix 952

TronicLT:


martinwicke:
I think all these commits are in HEAD -- are you sure this is still needed?

martinwicke:
See #981

PR #991, opened 2016-02-04
Made the fixed point code compile when AVX2 instructions ar enabled

benoitsteiner:


martinwicke:
Jenkins, test this please.

vrv:
your set of commits seems to have a protobuf change in it -- you might want to remove that since it looks like it's causing build failures?

benoitsteiner:
@vrv: I don't see a protobuf change in the diffs.In fact the files that I changed don't have any protobufs.

martinwicke:
Jenkins, test this please.

vrv:
@benoitsteiner can you resolve the conflict and rebase?  Looks good otherwise.

martinwicke:
When you resolve the conflict, please omit the "external/eigen_archive/" part of the paths, a change to bazel makes this the preferred way to referencing those files, so they can be used from targets outside tensorflow.

benoitsteiner:
I have resolved the merge conflicts and rebased.

peterbraden:
@martinwicke I'm hitting this now, how does an external build reference these files?

martinwicke:
@peterbraden, are you asking about the include paths? What are you trying to do?

These are changes made necessary in order to allow tensorflow to be included in another bazel project (for example, as a submodule).

peterbraden:
@martinwicke I'm trying to reference tensorflow from a non-bazel project (I'm trying to create nodejs bindings to tensorflow). Because of the references to the bazel-downloaded files, I can't successfully link.

martinwicke:
If you built tensorflow with bazel (just to get all the downloads), the
downloads will be symlinked in `tensorflow/bazel-tensorflow/external...`,
and the built libraries will be symlinked in
`tensorflow/bazel-out/external/...`.

If you need to fiddle with what bazel links in, you should probably ask
over at bazel, they're more likely to know the internals.

There's an effort in PR728 to make a cmake build, which I'd love to accept
to contrib. That may also solve your problem, at least partially.

PR #997, opened 2016-02-04
Python3 tests: getting easy ones out of the way

caisq:
Adding PY2AND3 label for a Python test file. Getting rid some obvious Python 2-3 compatibility issues that cause test errors, including comparing lists with integers, string-byte array difference, and use of xrange. After this CL, there are still 11 tests that fail in Python3. Those are more involved issues and will be addressed separately.

### DIFF
--- a/tensorflow/python/kernel_tests/py_func_test.py
+++ b/tensorflow/python/kernel_tests/py_func_test.py
@@ -24,6 +24,7 @@ import tensorflow as tf
 
 from tensorflow.python.framework import errors
 from tensorflow.python.ops import script_ops
+from six.moves import xrange
 
 
 class PyOpTest(tf.test.TestCase):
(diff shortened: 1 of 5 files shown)


tensorflow/python/kernel_tests/py_func_test.py
```diff
@@ -91,7 +91,7 @@ def testLarge(self):
       x = tf.zeros([1000000], dtype=np.float32)
       y = tf.py_func(lambda x: x + 1, [x], [tf.float32])
       z = tf.py_func(lambda x: x * 2, [x], [tf.float32])
-      for _ in xrange(100):
+      for _ in range(100):
```
martinwicke: For this size of range it doesn't matter much, but better to add a "from six.moves import xrange" to get xrange back.

martinwicke:
Jenkins, test this please

caisq:
Using six.moves.xrange now. Commits squashed.

martinwicke:
Jenkins, test this please.

martinwicke:
merged

PR #998, opened 2016-02-04
Script for Jenkins parameterized matrix build

caisq:
The script takes a few environment variables as input and invoke
ci_build.sh in the same directory in the proper way. It takes
information about the build type (cpu | gpu | android), Python version,
whether "-c opt" is used and whether PIP install-test is to be
performed. It automatically examines the parameters to determine the
correct build script to call and the argument flags to use with it. It
also examines whether Docker is available on the system to determine
whether the build will be done inside Docker or not. If GPU Docker build
is detectd, it'll map the CUDA devices and libraries automatically.
In addition, it looks at some optional environment variables to set
additional command flags for Docker and the build command.

See experimental Jenkins matrix builds at: 
For CPU (Linux and Mac): http://ci.tensorflow.org/view/Experimental/job/experimental-cais-matrix-linux-cpu-test/8/
For GPU (Linux): http://ci.tensorflow.org/view/Experimental/job/experimental-cais-matrix-linux-gpu-test/
For Android: http://ci.tensorflow.org/view/Experimental/job/experimental-cais-matrix-linux-android-test/


tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -0,0 +1,204 @@
...
+#
+# The script obeys the following required environment variables:
+#   TF_BUILD_CONTAINER_TYPE:   (CPU | GPU | ANDROID)
+#   TF_BUILD_PYTHON_VERSION:   (PYTHON2 | PYTHON3)
+#   TF_BUILD_IS_OPT:           (NO_OPT | OPT)
+#   TF_BUILD_IS_PIP:           (NO_PIP | PIP)
+#
+# Note: certain combinations of parameter values are regarded
+# as invalid and will cause the script to exit with code 0. For example:
+#   NO_OPT & PIP     (PIP builds should always use OPT)
```
martinwicke: Why is that?

martinwicke:
I like this. @jendap?

tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -0,0 +1,204 @@
...
+# Default configuration
+CI_BUILD_DIR="tensorflow/tools/ci_build"
+
+# Command to call when Docker is available
+DOCKER_MAIN_CMD="${CI_BUILD_DIR}/ci_build.sh"
+# Command to call when Docker is unavailable
+NO_DOCKER_MAIN_CMD="${CI_BUILD_DIR}/builds/configured"
+
+# Additional option flags to apply when Docker is unavailable (e.g., on Mac)
+NO_DOCKER_OPT_FLAG="-j 1 --genrule_strategy=standalone "\
```
jendap: Do we need these arguments outside of mac? Why just one job? It will take forever to build, will it not? And why the other two?

tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -0,0 +1,204 @@
...
+if [[ -z "$(which docker)" ]]; then
+  echo "It appears that Docker is not available on this system. "\
+"Will perform build without Docker."
+  echo "In addition, the additional option flags will be applied to the build:"
+  echo "  ${NO_DOCKER_OPT_FLAG}"
+  MAIN_CMD=${NO_DOCKER_MAIN_CMD}
+  OPT_FLAG="${OPT_FLAG} ${NO_DOCKER_OPT_FLAG}"
+
+elif [[ ${CTYPE} == "gpu" ]]; then  # Docker GPU: Map devices and libraries
+  CUDA_DEVICES=$(ls /dev/nvidia* | xargs -I{} echo '--device {}:{}' | xargs)
```
jendap: You do not have to do this. It is already in ci_build.sh

tensorflow/tools/ci_build/ci_parameterized_build.sh
martinwicke: If it works, let's omit the -j1.

tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -0,0 +1,204 @@
...
+  CUDA_LIBS=$(ls /usr/lib/x86_64-linux-gnu/libcuda* | xargs -I{} echo '-v {}:{}' | xargs)
+
+  EXTRA_PARAMS="${EXTRA_PARAMS} ${CUDA_DEVICES[@]} ${CUDA_LIBS[@]}"
+fi
+
+# Process Bazel "-c opt" flag
+if [[ ${TF_BUILD_IS_OPT} == "no_opt" ]]; then
+  # PIP builds are done only with the -c opt flag
+  if [[ ${TF_BUILD_IS_PIP} == "pip" ]]; then
+    echo "Skipping parameter combination: ${TF_BUILD_IS_OPT} & "\
```
jendap: Why not?

tensorflow/tools/ci_build/ci_parameterized_build.sh
caisq: Removed as suggested.

tensorflow/tools/ci_build/ci_parameterized_build.sh
caisq: When you generate the install package, typically you want to use the optimized C++ build. If there is any build configuration that we want to leave out for efficiency, these are the first ones to go.

caisq: In fact, if you look at pip.sh (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/builds/pip.sh), the "-c opt" option is always used when running the target "//tensorflow/tools/pip_package:build_pip_package". So specifying NO_OPT here would have no effect anyway.

tensorflow/tools/ci_build/ci_parameterized_build.sh
caisq: I copied those from the mac build on Jenkins assuming they were necessary. I confirmed that except "--linkopt=-headerpad_max_install_names", the flags can be omitted without causing the build to fail. The "--linkopt" flag seems to be necessary. See: 
http://ci.tensorflow.org/job/tensorflow-master-mac/146/

Keeping it ensures that both the clean and non-clean bazel build pass.
http://ci.tensorflow.org/job/tensorflow-master-mac/147/
http://ci.tensorflow.org/job/tensorflow-master-mac/148/

So I'll remove the option flags except "--linkopt=-headerpad_max_install_names".

martinwicke: I've had trouble without --spawn_strategy=standalone but a lot has changed
since then.

caisq:
I addressed comments from @jendap. In addition, I realized that there was an issue with BUILD_TAG values from matrix builds. Those tags have special characters such as "," and "=", which make the Docker image names illegal. I modified ci_build.sh a little to solve that problem.

tensorflow/tools/ci_build/ci_parameterized_build.sh
caisq: See my answer to @jendap below.

caisq:
@martinwicke Please hold on merging this one. There are some issues related to Docker runs for parameterized builds. I'm looking into it. I'll let you know when it's okay to merge.

martinwicke:
Ok

tensorflow/tools/ci_build/ci_build.sh
```diff
@@ -35,7 +35,9 @@ fi
 
 # Optional arguments - environment variables. For example:
 # CI_DOCKER_EXTRA_PARAMS='-it --rm' CI_COMMAND_PREFIX='' tensorflow/tools/ci_build/ci_build.sh CPU /bin/bash
-CI_DOCKER_EXTRA_PARAMS=("${CI_DOCKER_EXTRA_PARAMS[@]:---rm}")
+if [[ "${CI_DOCKER_EXTRA_PARAMS}" != *"--rm"* ]]; then
+  CI_DOCKER_EXTRA_PARAMS="--rm ${CI_DOCKER_EXTRA_PARAMS}"
+fi
```
caisq: Making sure that the "--rm" flag for Docker is always set. In the case of automatic parametric build, CI_DOCKER_EXTRA_PARAMS may already been set to an non-empty value. So the previous approach needs to be replaced.

caisq:
@martinwicke @jendap Feel free to merge now. The issue on Jenkins matrix build is now resolved. The change set is now squashed and ready.

martinwicke:
I merged this, apparently five hours ago.

PR #1002, opened 2016-02-05
Set visibility of all targets to //visibilty:public

damienmg:
Bazel has a bug that prevent correct resolution of visibility labels
and this change is using //visiblity:\* targets to be able to
load tensorflow as a remote repository.

### DIFF
--- a/third_party/gpus/cuda/BUILD
+++ b/third_party/gpus/cuda/BUILD
@@ -144,19 +144,8 @@ genrule(
         "OUTPUTDIR=`readlink -f $(@D)/../../..`; cd third_party/gpus/cuda; OUTPUTDIR=$$OUTPUTDIR ./cuda_config.sh --check;",
 
         # Under non-cuda config, create all dummy files to make the build go through
-        ";".join([
-          "mkdir -p $(@D)/include",
-          "mkdir -p $(@D)/lib64",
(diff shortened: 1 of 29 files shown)


third_party/gpus/cuda/BUILD
```diff
@@ -144,19 +144,8 @@ genrule(
...
-          "touch $(@D)/include/cublas.h",
-          "touch $(@D)/include/cudnn.h",
-          "touch $(@D)/lib64/libcudart_static.a",
-          "touch $(@D)/lib64/libcublas.so." + tf_get_cuda_version(),
-          "touch $(@D)/lib64/libcudnn.so." + tf_get_cudnn_version(),
-          "touch $(@D)/lib64/libcudart.so." + tf_get_cuda_version(),
-          "touch $(@D)/lib64/libcufft.so." + tf_get_cuda_version(),
-            ]),
-    ),
+        "touch $(OUTS)"
```
martinwicke: Will this create the directories?

damienmg: Oops this change is not belonging there sorry.

Directories are automatically created by bazel.

damienmg:
No need for that, sorry for the noise. Bazel has fixed that

PR #1004, opened 2016-02-05
Strip out external/XXXX/ prefix from GetPath in cc_op_gen

damienmg:
GetPath was not stripping external/XXX/ which make C++ compilation
of the generated header fail. All external repositories are added
to the include path in C++ compilation by Bazel.


martinwicke:
Jenkins, test this please.

martinwicke:
merged.

PR #1007, opened 2016-02-06
TensorFlow: fix word2vec timeouts on GPU by pinning model on CPU

vrv:
Right now word2vec is not well supported on GPU (missing
scatter sub, etc).  We can remove this limitation later when
ScatterSub and other ops are all available on GPU.

With this fix, the two word2vec tests now run in 5 seconds on a machine with GPU.


caisq:
LGTM. The test failures you see on the PR gpu-pip build is due to a separate known issue.

martinwicke:
merged.

PR #1009, opened 2016-02-06
Example code for SummaryIterator using `tf.train.summary_iterator`.

eerwitt:
I couldn't find [`tensorflow.summary_iterator`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/__init__.py#L92) but found it under [`tensorflow.train`](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/python/training). I've updated the documentation to match [this test](https://github.com/tensorflow/tensorflow/blob/3b3b06278cc7971bb273563bc917a89d49d19e7d/tensorflow/python/training/summary_writer_test.py#L49) which also calls it by `tf.train.summary_iterator`.

### DIFF
--- a/tensorflow/python/training/summary_io.py
+++ b/tensorflow/python/training/summary_io.py
@@ -234,7 +234,7 @@ def summary_iterator(path):
   Example: Print the contents of an events file.
 
   ```python
-  for e in tf.summary_iterator(path to events file):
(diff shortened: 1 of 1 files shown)


martinwicke:
Merged. Thanks!

PR #1010, opened 2016-02-07
fix minor filename typo

moutai:
fix minor filename typo


moutai:
I signed it!
Thanks for the project!

PR #1011, opened 2016-02-07
Adding option for parallel build and serial tests

caisq:
In ci_parameterized_build.sh, allow the option of parallel bazel build
followed by serial (--jobs=1) bazel test.

Potentially useful for limiting the build time for GPU builds, where the
tests cannot run in parallel due to contention of GPU memory.

### DIFF
--- a/tensorflow/tools/ci_build/ci_parameterized_build.sh
+++ b/tensorflow/tools/ci_build/ci_parameterized_build.sh
@@ -41,6 +41,11 @@
 #   TF_BUILD_BAZEL_TARGET:
 #                      Used to override the default bazel build target:
 #                      //tensorflow/...
+#   TF_BUILD_SERIAL_TESTS:
+#                      Build parallely, but test serially
+#                      (i.e., bazel test --job=1), potentially useful for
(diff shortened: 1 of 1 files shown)


tensorflow/tools/ci_build/builds/configured
```diff
@@ -37,4 +37,18 @@ fi
 SCRIPT_DIR=$( cd ${0%/*} && pwd -P )
 ${SCRIPT_DIR}/print_build_info.sh ${CONTAINER_TYPE} ${COMMAND[@]}
 
-${COMMAND[@]}
+# Execute the individual commands, separated with &&
+COMMAND+=("&&")
```
jendap: Let's not do this. Those script are getting super complicated every day.

What if this whole feature was one or two lines in ci_parameterized_build.sh... just add "bash -c" to the command and you do not have to touch this script.

```bash
bash -c 'bazel build ... && bazel test -j 1 ...'
```

caisq:
@jendap I addressed your suggestion of keeping tensorflow/tools/ci_build/builds/configured untouched. The commits are squashed.

martinwicke:
merged

PR #1022, opened 2016-02-08
Python3 install-test in pip.sh

caisq:


martinwicke:
Is this file already used in our tests?

caisq:
@martinwicke It is used in only some of the experimental builds right now.

martinwicke:
Then I'll save ourselves the testing. Merged.

PR #1025, opened 2016-02-09
More explicit directory structure instructions

martinwicke:
More documentation about directory structure so I can point at it.


martinwicke:
merged.

PR #1031, opened 2016-02-09
Improvements to print_build_info.sh and pip.sh

caisq:
This is in response to @jendap's earlier comments. Using the $() subscript syntax.

Also, making --upgrade default option for pip

This is addressing the issue that if --upgrade is not used, preexisting installation of TensorFlow PIP on the system will make the install-test results obsolete (e.g., on Mac, where we don't use Docker to run the tests)


martinwicke:
merged.

PR #1036, opened 2016-02-09
Fix all but one Python 3 test

girving:
`saver_test` is still broken.

### DIFF
--- a/tensorflow/python/framework/load_library.py
+++ b/tensorflow/python/framework/load_library.py
@@ -61,7 +61,7 @@ def load_op_library(library_filename):
 
   op_list_str = py_tf.TF_GetOpList(lib_handle)
   op_list = op_def_pb2.OpList()
-  op_list.ParseFromString(bytes(op_list_str))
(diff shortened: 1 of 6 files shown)


caisq:
This PR contains a lot edits in the same spots as https://github.com/tensorflow/tensorflow/pull/1034 We can look and decide which PR to merge.

girving:
It's much better to open the file correctly as binary or not than to decode manually, so this PR seems better.

caisq:
@girving I agree that this PR is better with one exception: In scatter_ops_test.py, use 
`from six.moves import xrange`

so that we can get both 2-3 compatibility and the performance of xrange.

caisq:
I'm closing https://github.com/tensorflow/tensorflow/pull/1034 in favor of this one.

girving:
The performance difference is negligible for small loops and this is just a test, so I went for simplicity.

martinwicke:
Jenkins, test this please.

martinwicke:
merged.

PR #1051, opened 2016-02-11
 A few bug fixes to pip.sh and ci_parameterized_build.sh

caisq:
1) Perform pip install twice, first time with --upgrade and second time without, to fix sporadic test failures related to protobuf version
2) Correctly determine the local pip install directory based on Python version being used
3) Always perform "bazel test" on non-Docker build environments, such as Mac

### DIFF
--- a/tensorflow/tools/ci_build/builds/pip.sh
+++ b/tensorflow/tools/ci_build/builds/pip.sh
@@ -100,6 +100,14 @@ if [[ -z "$PYTHON_BIN_PATH" ]]; then
   exit 1
 fi
 
+# Determine the major and minor versions of Python being used (e.g., 2.7)
+# This info will be useful for determining the directory of the local pip
+# installation of Python
+PY_MAJOR_MINOR_VER=$(${PYTHON_BIN_PATH} -V 2>&1 | awk '{print $NF}' | cut -d. -f-2)
+
+echo "Python binary path to be used in PIP install-test: ${PYTHON_BIN_PATH} "\
+"(Major.Minor version: ${PY_MAJOR_MINOR_VER})"
+
 # Build PIP Wheel file
 PIP_WHL_DIR="pip_test/whl"
 PIP_WHL_DIR=`abs_path ${PIP_WHL_DIR}`  # Get absolute path
@@ -119,7 +127,10 @@ echo "whl file path = ${WHL_PATH}"
 # Install, in user's local home folder
 echo "Installing pip whl file: ${WHL_PATH}"
 
+# Call pip install twice, first time with --upgrade and second time without it
+# This addresses the sporadic test failures related to protobuf version
 ${PYTHON_BIN_PATH} -m pip install -v --user --upgrade ${WHL_PATH} &&
+${PYTHON_BIN_PATH} -m pip install -v --user ${WHL_PATH} &&
 
 # If NO_TEST_ON_INSTALL is set to any non-empty value, skip all Python
 # tests-on-install and exit right away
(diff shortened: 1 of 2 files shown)


tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -119,7 +127,10 @@ echo "whl file path = ${WHL_PATH}"
 # Install, in user's local home folder
 echo "Installing pip whl file: ${WHL_PATH}"
 
+# Call pip install twice, first time with --upgrade and second time without it
+# This addresses the sporadic test failures related to protobuf version
 ${PYTHON_BIN_PATH} -m pip install -v --user --upgrade ${WHL_PATH} &&
+${PYTHON_BIN_PATH} -m pip install -v --user ${WHL_PATH} &&
```
martinwicke: What is the point of the &&? Maybe my bash is rusty.

martinwicke: If it means what I think it does, should we just add `set -e` to the top of the file and remove all these?

caisq: I know "set -e" is used in some other files in tensorflow/tools/ci_build. But it is considered unreliable (http://mywiki.wooledge.org/BashFAQ/105/Answers). This is why I prefer to check for error manually using && at critical places in the script.

caisq:
I added some further changes to fix a bug related to ci_build.sh arguments containing double ampersand (&&).

tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -120,15 +128,20 @@ fi
...
+  echo "Also, the additional option flags will be applied to the build:"
   echo "  ${NO_DOCKER_OPT_FLAG}"
-  MAIN_CMD=${NO_DOCKER_MAIN_CMD}
+  MAIN_CMD="${NO_DOCKER_MAIN_CMD} ${CTYPE}"
   OPT_FLAG="${OPT_FLAG} ${NO_DOCKER_OPT_FLAG}"
 
+  # Perform bazel clean in non-Docker environments to ensure a clean build
+  echo "Also, \"${BAZEL_CLEAN_CMD}\" will be performed prior to "\
+"the main build step"
+  DO_BAZEL_CLEAN=1
```
jendap: Why do you ever need this? Bazel should be good enough to never need it. And it seems so. No?

martinwicke: Apparently, if a commit changes the WORKSPACE file, bazel is not good enough to figure it out, leading to inconsistencies for instance in the Eigen version. This is a bazel bug.

caisq:
Let me know if there are any further comments. If there is none, please merge it to master and r0.7 branches for the release.

tensorflow/tools/ci_build/ci_parameterized_build.sh
jendap: I see. That is unfortunate. But then we should do it all the time, even inside docker? It will be slow though.

caisq: @jendap Yes, that would slow things down. But I see your point. I guess the Bazel caching somehow didn't work when the folder is mapped from outside Docker to inside, which is the main reason why we haven't needed to do bazel clean for Docker builds? @martinwicke has their been any changes regarding the caching recently?

jendap: Actually everything from outisde the container is mapped to inside. The filesystem should look like there is no docker at all.

Can it be that you are not running with standalone argument outside docker? The server process of bazel is what I believe reads WORKSPACE file. Not afterwards. I believe the standalone flags can solve with without bazel clean.

caisq: @jendap @martinwicke Alright, how about this? I'll add --genrule_strategy=standalone to without-Docker and get rid of the bazel clean. In the meantime, I'll add bazel clean as an option that can be activated through environment variable "TF_BUILD_BAZEL_CLEAN" (any non-empty and non-0 value).

caisq:
I removed "bazel clean" as a compulsory step for non-Docker builds and added "--genrule_strategy=standalone" for such builds.  In addition, I added TF_BUILD_BAZEL_CLEAN as an optional environment variable.

caisq:
Rebased and squashed

martinwicke:
merged.

martinwicke:
And cherry-picked into r0.7

PR #1057, opened 2016-02-11
Clean up references

Styrke:
Clean up references and add explicit links to pdf files for references
that are displayed on www.tensorflow.org.

Because having to download a pdf when you just want to see the title and abstract is mildly annoying.


vrv:
I think a general issue here is that very often new versions of the papers are uploaded to arxiv, and hardcoding a link to a specific version will often go stale (at least, that's what appears to be the case from the URLs).  Does arxiv have a 'latest' pdf link so that we don't have to hardcode a specific version?

vrv:
In other words, some of your changes are great -- do you think we should just always link to the arxiv front page instead of pdfs, consistently?

JonasMeinertz:
Sometimes it's nice to have both, but I always prefer a link to the arxiv page over a link directly to the pdf. As I mentioned in the introduction, I get annoyed when i have to wait 5-20 seconds for a pdf to download if I just want to read the title and abstract.

Playing around with arxiv just now, it seems that removing the version number from a pdf link will point to the latest version. So the most recent version of http://arxiv.org/abs/1409.3215 is http://arxiv.org/pdf/1409.3215.pdf.

JonasMeinertz:
http://arxiv.org/pdf/1409.3215 works too, but then it's not clear that it's a pdf.

vrv:
I too prefer the arxiv page over pdf.  Want to make that change to this PR?  If you want to have the pdf link too, include the version-less link and we're good :)

JonasMeinertz:
I will fix the pdf links here. :-)

JonasMeinertz:
@vrv Links to pdf files are version-less now.

vrv:
Merged, thanks!

PR #1058, opened 2016-02-11
Fix a bug for an unavailable argument

yoneken:
The _num_classes_ argument of the _dense_to_one_hot_ function is never called by the other function.

IHMO, we should set the number of classes in _extract_labels_.
Since, the _extract_labels_ is a task-specific-function but the _dense_to_one_hot_ is not.


vincentvanhoucke:
Merged, thanks!

PR #1075, opened 2016-02-12
Fixing Python 3-breaking xrange in r0.7 branch

caisq:


vrv:

tensorflow/tensorboard/backend/tensorboard_handler_test.py
```diff
@@ -25,6 +25,7 @@
 from tensorflow.python.platform import googletest
 from tensorflow.tensorboard.backend import tensorboard_handler
 
+from six.moves import xrange
```
jendap: That should work. I have tried yesterday the same. Only ... should the import of six be above tensorflow? do we have some ordering rules?

vrv: I think it would be best to have this above googletest actually

martinwicke: If we want to avoid issues later, put it above. It'll just create a merge conflict when some overzealous linter gets a hold of it.

caisq:
Fixed python import order. Rebased and squashed. Should I double-submit to master?

martinwicke:
I'm thinking we'll likely merge r0.7 into master once we've release, so no
harm done either way.

caisq:
@martinwicke OK. Please merge it to r0.7 so we can get this release issue out of the way.

vrv:

vrv:
Merged

PR #1077, opened 2016-02-12
Fix tensorboard serving of external assets

teamdandelion:
This is a quick fix to have TensorBoard correctly load external assets, after we have renamed the directories for compatibility with new Bazel.


vrv:

PR #1081, opened 2016-02-13
added return state to bidirectional rnn, added unit tests

lucaswiser:
Also modified other unit tests in BidirectionalRNNTest to fit new return value.


lucaswiser:
Please look at and run unit tests. I created three new functions in the test class BidirectionalRNNTest, and modified two more.

lucaswiser:
Don't merge, the tests I added have a bug and fail. Will fix now.

PR #1082, opened 2016-02-13
Make failing builds fail

martinwicke:


martinwicke:
@caisq @jendap, PTAL.

martinwicke:
Let's see what Jenkins thinks about this.

martinwicke:
Seems to work as intended.

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -69,17 +69,22 @@ abs_path() {
...
-  bazel build -c opt ${PIP_BUILD_TARGET}
+  bazel build -c opt ${PIP_BUILD_TARGET} || die "Build failed."
 elif [[ ${CONTAINER_TYPE} == "gpu" ]]; then
-  bazel build -c opt --config=cuda ${PIP_BUILD_TARGET}
+  bazel build -c opt --config=cuda ${PIP_BUILD_TARGET} || die "Build failed."
 else
-  echo "Unrecognized container type: \"${CONTAINER_TYPE}\""
-  exit 1
+  die "Unrecognized container type: \"${CONTAINER_TYPE}\""
 fi
```
caisq: I think a easier and cleaner change would be adding " &&" after this line, (line 88, fi)

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -96,8 +101,7 @@ source tools/python_bin_path.sh
 
 # Assume: PYTHON_BIN_PATH is exported by the script above
 if [[ -z "$PYTHON_BIN_PATH" ]]; then
-  echo "PYTHON_BIN_PATH was not provided. Did you run configure?"
-  exit 1
+  die "PYTHON_BIN_PATH was not provided. Did you run configure?"
```
caisq: I'm fine with the rest of the change though. They reduce the code in size a little and achieve the same purpose.

tensorflow/tools/ci_build/builds/pip.sh
martinwicke: Because those lines are so important, I'd like to be very sure that they fail properly, i.e., I think it's better if each line alone tells you that it will either succeed or exit the script. Having the potential failure two branches away is much harder to read (at least for me).

caisq: OK. But to be a little nit-picky, say we add more container types in the future (e.g., Windows, various CUDA versions, etc.), there will be more logical branches and this approach will get a little verbose. 

If clear failure message is desired, I would suggest adding the following lines after line 88:

if [[ $? != "0" ]]; then
  die "FAILED at build step for container type ${CONTAINER_TYPE}"
fi

martinwicke: I really prefer the default to be the other way (set -e), such that if something fail, I get to know, instead of saying everywhere that if stuff fails, please tell me. You are correct, if it gets more complex, catching all of the if/fi construct may be a good idea.

martinwicke:
merged into master.

caisq:
Should this be cherry-picked into r0.7 so that the release testing can get the benefit as well?

martinwicke:
Yes, it must be.

PR #1085, opened 2016-02-13
Update 0.6.0 docs with fix to tutorial.  Fixes #1084

### LINKED ISSUE #1084: Missing variable on MNIST tutorial
I'm writing this here because I can't find a better way to contact someone about this.

In short, first Tensorflow tutorial about MNIST:
https://www.tensorflow.org/versions/v0.6.0/tutorials/mnist/beginners/index.html#the-mnist-data

First piece of code:
```
import tensorflow.examples.tutorials.mnist.input_data
mnist = input_data.read_data_sets("MNIST_data/", one_hot=True) 
```

where it should be:
```
import tensorflow.examples.tutorials.mnist.input_data as input_data
mnist = input_data.read_data_sets("MNIST_data/", one_hot=True) 
```
**"as input_data"** is missing, so when running the second line of code, python doesn't know whats input_data.whatever.

Hope someone can edit that, because it's a begginer tutorial, and can be very frustrating to fail running the very first piece of code of the first and most basic tutorial.

vrv:


martinwicke:
This won't actually fix the docs. I think we should revisit your idea of
using branch head for generating docs.

vrv:
Right, but at least the head of the branch is now correct.

PR #1087, opened 2016-02-14
Clean out assets left over from a prior run

martinwicke:
Before, assets were not cleaned out if the build failed. Since some of them are created with mode 400, the next run will fail because unzip won't overwrite them. This makes sure we always remove the directory before we try unzipping.


martinwicke:
merged.

vrv:
LGTM.

PR #1089, opened 2016-02-14
Enable GPU reduction for doubles

ibab:
I was just trying to apply `reduce_sum` to a `float64` array on a GPU and noticed that the reduction operators are currently only defined for `float32`.
Is there a reason for limiting the type?

I've added reduction definitions for `double` in this PR.

I've also played with adding reduction ops for all `realnumbertypes` (the reduction ops in [core/ops/math_ops.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc) specify `numbertype`).
I can add them to this PR if there's interest.


vrv:

vrv:
I think this is great, thanks for doing this.

We could use some tests to validate this though:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/reduction_ops_test.py#L371 is only testing for float, we should add one for double.

(I'm hoping we'll have our GPU machine to test this by the time you get this done :)  @martinwicke )

martinwicke:
I hope that we have our GPU machine replaced with a cloud thingy by then. :)

vrv:
cool, once we have some tests for these, we can merge

ibab:
I've adjusted the existing test for `Sum` and added tests for the other three reducers.

One thing I've noticed is that all of the python tests make use of `test_session(use_gpu=True)` instead of `force_gpu=True` (With the exception of one function in `scatter_ops_test.py`).
When an op isn't defined for the GPU, this will happily run it on the CPU, kind of bypassing the original intent of the test.
But changing `use_gpu` to `force_gpu` everywhere would probably require a lot of cleanup work to make the tests pass again.

What's your opinion on switching to `force_gpu` to make the tests more strict?

vrv:
It's difficult because then the test won't run when the client doesn't have a GPU.  This is basically a way to say 'if you have a GPU, test it.'  Since we now run GPU tests on Jenkins, we should get coverage.

ibab:
Ah, okay, thanks!
It seems that updating the branch here on Github requires the tests to be run again.

vrv:
Don't worry about updating the branch as long as there's no conflict :)

vrv:
well, I guess that's not technically safe, but in this case it would have likely been fine.

vrv:
Merged

PR #1094, opened 2016-02-14
Add a CNN sample for MNIST

yoneken:
The purpose of this PR is,
- Separate net structure (net.py, mnist.py) and train process (fully_connected_feed.py) clearly.
- Add a CNN sample (mnist_cnn.py) for MNIST

Any advices and comments would be greatly appreciated.
Thanks!

### DIFF
--- a/tensorflow/examples/tutorials/mnist/fully_connected_feed.py
+++ b/tensorflow/examples/tutorials/mnist/fully_connected_feed.py
@@ -27,7 +27,8 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 import tensorflow as tf
 
 from tensorflow.examples.tutorials.mnist import input_data
-from tensorflow.examples.tutorials.mnist import mnist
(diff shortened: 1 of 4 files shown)


vrv:
My high level comment is that, while this is nice, I think the original intent of the tutorials was to keep the examples minimal and self-contained, rather than factored out -- the longer term intent is to convert these into IPython Notebooks that have most of the relevant content in one.

If you feel strongly this should be factored out, can you explain what the goal of your refactoring is?

yoneken:
> the original intent of the tutorials was to keep the examples minimal and self-contained
> the longer term intent is to convert these into IPython Notebooks

I strongly agreed.
Thank you for the comment.

When I leaned TensorFlow, I felt a big barrier between the beginner's mnist tutorial and the expert's one.
It's a reason that I wrote the code which has the same style between those.
I am looking forward to see such a sophisticated tutorial and code.

PR #1106, opened 2016-02-15
Python API docs fix array syntax

noporpoise:
Fix array syntax in comments for examples `tf.pow` and `tf.accumulate_n`


noporpoise:
I've also updated the docstrings in `tensorflow/python/ops/math_ops.py`.

vrv:
Thanks!

PR #1109, opened 2016-02-15
shrink pip package content

jendap:
@martinwicke @caisq we need this and the questions is if we want to keep the "external" there.

### DIFF
--- a/tensorflow/tools/pip_package/build_pip_package.sh
+++ b/tensorflow/tools/pip_package/build_pip_package.sh
@@ -33,8 +33,10 @@ function main() {
     exit 1
   fi
   cp -R \
-    bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/* \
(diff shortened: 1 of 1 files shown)


jendap:
The external contains a lot of things - all the javascript files for tensorboard (we probably want those), but also all the eigen header files (we don't need those), and perhaps many other things. I'm going to look at those directories and probably remove a few of them now. But that's hacky, not long term solution.

jendap:
It is updated to remove the eigen headers. All javascript stay.

martinwicke:
@benoitsteiner @danmane This is in, but I'd like your opinions on what can/should stay and what can be removed.

danmane:
We can absolutely shrink the pip package further. Right now we don't minify TensorBoard's HTML and JS, and I'm not sure whether we are taking the minified js or full js (or both) into the pip package. So there are some very low hanging fruit here.

PR #1112, opened 2016-02-16
TensorFlow: add thread annotations to mutex classes to prevent

vrv:
thread annotation warnings from popping up (mostly on Mac with
clang).

Tests passed with this change, but this seems a bit ugly, and I wonder
if there's a better way to do the annotation.  Suggestions welcome.


jendap:
Perfect! I was looking at this a while back but never finished it. Merge it! It will kill gadzillion of warnings in mac build :-)

martinwicke:
merged.

PR #1114, opened 2016-02-16
update docker for 0.7.0

jendap:


cancan101:
Why change the URLs to `http` from `https`?

martinwicke:
Odd problem with urllib.

cancan101:
Care to elaborate?

martinwicke:
@jendap, what was the problem there?

PR #1116, opened 2016-02-16
keep numpy version in pip.sh

jendap:
We should keep building against numpy 1.8.2 - the version in ubuntu 14.04.


jendap:
Note: user home directory is cached between the builds - bazel is picking up the newer version installed after first run of pip.sh

Also this should be able to automatically downgrade all the installations in all the jobs on ci.tensorflow.org. But it will need two builds. First one will still pick the new version ob numpy.

jendap:
Ok, it works. But the first build fails. Just like on this PR. The first bazel build took 1.10.4 cached from previous run and then failed while running test with downgraded 1.8.2. Next runs should work.

jendap:
Also the docker container build with this works :-)

martinwicke:
So we need to re-run the release twice to propagate this to the binaries?

PR #1118, opened 2016-02-16
use string::append to catenate strings

cg31:
With VC++, when std::string is empty, there will be an error -- "string iterator not dereferencable" -- to use string::begin.

And using raw pointer to manipulate std::string doesn't look like a good idea.


cg31:
something is wrong, closing it.

PR #1119, opened 2016-02-16
Normalization of image

Macbull:
Generally pixel intensity or features are normalised to [-1,1] range. But, here the range was [-0.5,0.5]. 
Sorry for inconvenience, if it was intentional.


Macbull:
I signed it!

vincentvanhoucke:
The range does not matter, as long as it's O(1.0).

PR #1120, opened 2016-02-16
use string::append to catenate strings

cg31:
With VC++, when std::string is empty, there will be an error -- "string iterator not dereferencable" -- to use string::begin.

And using raw pointer to manipulate std::string doesn't look like a good idea.

### DIFF
--- a/tensorflow/core/lib/strings/strcat.cc
+++ b/tensorflow/core/lib/strings/strcat.cc
@@ -56,61 +56,47 @@ AlphaNum::AlphaNum(Hex hex) {
 // Append is merely a version of memcpy that returns the address of the byte
 // after the area just overwritten.  It comes in multiple flavors to minimize
 // call overhead.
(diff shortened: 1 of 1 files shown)


concretevitamin:
Jenkins, test this please.

vrv:

google/protobuf
```diff
@@ -1 +1 @@
-Subproject commit caf1fb7197ee94c07108fc7cfbca07432b185a28
```
vrv: you'll want to undo this

vrv:
(the branch has conflicts, can you address the conflict? we can take a look after that)

caisq:
@vrv The conflicts needs to be resolved first before the tests can pass on Jenkins. The base is a few weeks old.

vrv:
My guess is we'll get to this as part of our Windows support.

PR #1133, opened 2016-02-17
don't need a '+' before lambda

cg31:
I think the + before the lambda is a typo. It can't compile by MSVC. On Linux, after removing it, gcc compile the code without problem and all tests pass as well.


vrv:

vrv:
(It's actually not a typo according to the spec, but it is probably unnecessary.)

vrv:
Merged

PR #1137, opened 2016-02-17
Embed __version__ into package

rasbt:
Hi,
it would be nice to embed a version into the tensorflow package for reproducibility. The common Python convention is to do this via `__version__` attribute in the `__init__.py` file so that one can import the package and verify its version. 

```
import tensorflow as tf
print(tf.__version__)
```

This is very helpful for reproducibility and managing different virtual environments.

### DIFF
--- a/tensorflow/__init__.py
+++ b/tensorflow/__init__.py
@@ -21,3 +21,5 @@ from __future__ import division
 from __future__ import print_function
 
 from tensorflow.python import *
+
+__version__ = '0.8.0dev'


vrv:
This already works since 0.6 -- do you have evidence this is not the case?

rasbt:
Thanks for the quick feedback. I just upgraded to 0.7 and wanted to check the version, which didn't work (must have been still 0.5). I tried another `pip --upgrade` which still wasn't able to install it properly; this time, I noticed the error during the upgrade. 

    Cannot remove entries from nonexistent file /Users/Sebastian/miniconda3/envs/py27/lib/python2.7/site-packages/easy-install.pth

which may have prevented the correct installation. So, it must be some problem on my side (just installed in my python 3 environment were it works perfectly fine. Sorry for bothering.

vrv:
No problem, always good to double check that we didn't mess something up :)

PR #1144, opened 2016-02-17
Use standard types and add missing returns to fix compitable with MSVC

cg31:


vrv:

PR #1151, opened 2016-02-17
Adding tests and options to pip.sh

caisq:
1) Adding test-on-intall for tensorflow/python/{contrib,examples,models,tensorboard}

2) Adding option for TF_BUILD_IS_PIP=BOTH, which performs both the "bazel test" step and the "pip.sh" test-on-install step in one single build

3) Adding "bazel clean" switch to pip.sh


tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -199,11 +216,11 @@ rm -rf ${PY_TEST_DIR}/tensorflow/core/lib/png
 cp -r tensorflow/core/lib/png ${PY_TEST_DIR}/tensorflow/core/lib
 
 # Run tests
-DIR0=`pwd`
-ALL_PY_TESTS=`find tensorflow/python -name "*_test.py"`
+DIR0=$(pwd)
+ALL_PY_TESTS=$(find tensorflow/{contrib,examples,models,python,tensorboard} -name "*_test.py" | sort)
```
jendap: nice!

jendap:

caisq:
@jendap: I know some tests will faiure (Mac+Python3). Those are unrelated to this change set. Also, the commits are messed up and difficult to rebase. I'll create another PR with only one single commit (with the same changes, of course).

jendap:

jendap:
I think the linux build was killed by mac py3 failing. I have fixed that.

Can you squash and rebase it please?

Otherwise it looks good. (although I'm getting a bit lost in those scripts, but I may have idea about that)

PR #1152, opened 2016-02-17
fix master cpu build url

jendap:
because we have renamed the master to master-cpu


jendap:
@caisq I wanted to rename it some time ago. But then ... one would have to to bother updating this.

PR #1154, opened 2016-02-17
Adding tests and options to pip.sh

caisq:
1) Adding test-on-intall for
tensorflow/python/{contrib,examples,models,tensorboard}

2) Adding option for TF_BUILD_IS_PIP=BOTH, which performs both the
"bazel test" step and the "pip.sh" test-on-install step in one single
build

3) Adding "bazel clean" switch to pip.sh

(This is a squashed version of https://github.com/tensorflow/tensorflow/pull/1151, with the same changes)


caisq:
@jendap has LGTM'ed this in #1151. Ready to merge.

vrv:
Merged

PR #1164, opened 2016-02-18
Fix tensorboard import path

Vesnica:


Vesnica:
I signed it!

vrv:
@danmane: is this correct? 

https://github.com/tensorflow/tensorflow/commit/880011919c7b113802eb23c54ed5b8e2218e4416 specifically changed this a few days ago.

vrv:
Hm, with the reorganization of tensorboard, I think this is right.  Currently  tensorboard install is broken at HEAD, and I think this fixes it.

vrv:
(Merged a few days ago)

PR #1172, opened 2016-02-18
pip.sh: die on pip install failure

caisq:


tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -141,8 +142,10 @@ echo "Installing pip whl file: ${WHL_PATH}"
 
 # Call pip install twice, first time with --upgrade and second time without it
 # This addresses the sporadic test failures related to protobuf version
-${PYTHON_BIN_PATH} -m pip install -v --user --upgrade ${WHL_PATH} numpy==1.8.2 &&
-${PYTHON_BIN_PATH} -m pip install -v --user ${WHL_PATH} &&
+${PYTHON_BIN_PATH} -m pip install -v --user --upgrade ${WHL_PATH} numpy==1.8.2 \
+|| die "pip install (1st step) FAILED"
```
jendap: nit: you can put here "...(1st step - pip install --upgrade)..." or something like that (one may not remember which is is 1st and which is 2nd)

jendap:

caisq:
@jendap's comment addressed. Squashed.

vrv:
Merged

PR #1183, opened 2016-02-19
Clean up unused variables in a example

izeye:
This PR removes unused variables in `mnist_with_summaries.py` and makes its document align with it.

### DIFF
--- a/tensorflow/examples/tutorials/mnist/mnist_with_summaries.py
+++ b/tensorflow/examples/tutorials/mnist/mnist_with_summaries.py
@@ -59,16 +59,16 @@ def main(_):
     y = tf.nn.softmax(tf.matmul(x, W) + b)
 
   # Add summary ops to collect data
-  _ = tf.histogram_summary('weights', W)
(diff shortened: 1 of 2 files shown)


vrv:
I think for illustration purposes, the current code might be useful to show that these functions do return objects that *could* be used, even though at the moment the side effects are that they add to the graph. What do you think?

izeye:
@vrv If they have been added intentionally, this PR could be closed. Thanks for the feedback!

vrv:
@danmane: what do you think?  Originally they were given names (which led to unused_var lint errors).   Do you still think it's worth capturing the return values?

martinwicke:
The summary ops are magic in that the canonical summary use depends on the collection. They are ops just like all the others, but their return values are very rarely used (I guess you could depend on them, but typically you're really happy that they don't run, and what are you going to do with the serialized proto anyway except for write it to disk?). I would omit the return types in the example code, I don't think it's all that enlightening.

vrv:
Ok, merged

Issue #970, opened 2016-02-02
Work nvidia-docker rather than using custom docker start script


cancan101:
Right now the docs recommend using a [custom start command](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/docker#running-the-container) ([or here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#docker-installation)) rather than using the [`nvidia-docker` program published by nvidia](https://github.com/NVIDIA/nvidia-docker). I suggest having the image work with the more standardized approach to starting docker containers with nvidia support.

See: https://github.com/tensorflow/tensorflow/issues/808#issuecomment-175934093.

craigcitro:
I agree that getting the container started with GPU support still isn't the smoothest experience, especially when it comes to matching library versions. That said, I feel like "now go read the nvidia wiki and install this extra package" has its own tradeoffs as a getting started experience.

Probably the thing to do here is clear up the instructions:
1. make sure to clearly route people to either `docker_run_gpu.sh` or the NVidia docker plugin, and
2. add a few diagnostic steps (eg "here's how you run `deviceQuery` ...") for troubleshooting.

@ebrevdo do you have time to take a look at this?

jendap:
I would not rush it. I have spend a couple of hours with nvidia-docker two weeks ago. I have tried to use it for ci.tensorflow.org. It was not fun. I would not recommend that at this time. It would, imho, make more harm than good.

Simply try to install latest stable drivers. Check you have /dev/nvidia0, /dev/nvidiactl and /dev/nvidia-uvm outside docker and that you have /usr/lib/x86_64-linux-gnu/libcudart.so and /usr/lib/x86_64-linux-gnu/libcuda.so. If so the docker_run_gpu.sh script should work.

BTW: I believe docker and cuda together will get better. But it is not going to be a great experience any time soon if ever.

cancan101:
I would be curious to know more about what were the issues that you had with `nvidia-docker` that took a few hours. The program "just worked for me". The one caveat I would add is they rewrote the program ~ Jan 1 and also changed the base CUDA docker images (same images that TF extends for GPU). I think there are issues if you mix and match old images and new `nvidia-docker` or vice versa.

Otherwise, using new versions of both, my experience has been great. I just run `nvidia-docker run` and it sets the volumes and devices.

[...]

PR #1188, opened 2016-02-19
remove ld_library_path from docker image

jendap:
as suggested by @3XX0 in https://github.com/tensorflow/tensorflow/issues/970#issuecomment-185954223


vrv:
Merged

PR #1189, opened 2016-02-19
add samples to docker and start jupyter by default

jendap:


jendap:
Should this not be the default?

Note, people can still "docker run -it --rm tensorflow/tensorflow /bin/bash".

BTW: Perhaps we could for 0.8 bundle by default [tensorflow/examples/udacity](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/examples/udacity).

vrv:
Merged

PR #1190, opened 2016-02-19
update release version to 0.7.1

jendap:
That should be all that but RELEASE.md


caisq:
LGTM. As we do this more often, we should perhaps write a script to do this automatically.

PR #1202, opened 2016-02-19
Bidirectional rnn state

lucaswiser:
Added return state from forward and backward rnn in bidirectional_rnn. Updated tests to pass in kernel_tests/rnn_test.py


martinwicke:
Can you squash the commits and make sure the email you have used for your CLA is the author of the remaining commit?

lucaswiser:
@martinwicke Yes I can do that. I will probably close this and commit from another user name

PR #1207, opened 2016-02-19
Avoid copying invalid large constant data.

dongjoon-hyun:
Tensorflow cannot create an operation with a NodeDef larger than 2GB.
(Python Unit Test Case: ConstantTest.testTooLargeConstant)
However, before throwing a meaningful exception, it consumes lots of memory
to copy the invalid input data. This causes `std::bad_alloc` errors sometimes.

This PR prevents those invalid copying. The allocated memory for the following
testcase decreases from over 4GB into less than 1GB.

bazel-bin/tensorflow/python/constant_op_test ConstantTest.testTooLargeConstant

### DIFF
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -1038,7 +1038,7 @@ class Operation(object):
       raise TypeError("node_def needs to be a NodeDef: %s" % node_def)
     if node_def.ByteSize() >= (1 << 31) or node_def.ByteSize() < 0:
       raise ValueError(
-          "Cannot create an Operation with a NodeDef larger than 2GB.")
+          "Cannot create a tensor proto whose content is larger than 2GB.")
     if not _VALID_OP_NAME_REGEX.match(node_def.name):
       raise ValueError("'%s' is not a valid node name" % node_def.name)
     if not isinstance(g, Graph):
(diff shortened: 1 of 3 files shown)


keveman:
I am not sure expanding the API of `make_tensor_proto` for this purpose is a good idea, but that's my opinion. Adding josh11b@ and mrry@ for opinions.

josh11b:
I'm fine with this, but the code needs to be wrapped at 80 columns.

Mistobaan:
I agree with @keveman, it feels weird to inject the error message and the limit. I would add it as a special check in a try catch on the same operation ...
```
try:
    tensor_proto.tensor_content = nparray.tostring()
catch:
    // check error and check size and if is greater than 2GB raise error.
```

vrv:
Agreed, I prefer try/catching rather than adding args to the API.  Otherwise looks great.  Probably worth adding a test of the core functionality (without necessarily creating a 2GiB array)

dongjoon-hyun:
@vrv . Oh, it's not. We should avoid run the `try` part. 
When you call the following statements, it really consume all memory.
```
tensor_proto.tensor_content = nparray.tostring()
```

dongjoon-hyun:
HI, @Mistobaan and @keveman .
For the injection, the exception message depends on the situation.
For this PR, it's for `constant` op generation. 
For the other code, the error message will be different.

dongjoon-hyun:
I updated the PR and rebased to the master again.
  * Remove the added parameter: `limit` and `limit_msg`.
  * Use `try` statement in `constant_op.py` instead.
  * Keep the original behavior (including other `ValueError` exceptions).
Thank you, @vrv , @keveman , @josh11b , and @Mistobaan .

dongjoon-hyun:
Hi, @keveman . I updated the code according to your advice. Also, it rebased to the master and passed the test. Thanks.
(By the way, during rebasing, the messages on the code are gone. Sorry for that.)

PR #1215, opened 2016-02-20
Retraining Example with Notebook

kmader:
The code from the retraining demo morphed into a IPython Notebook to make it easier to use interactively.


kmader:
I signed it!

vrv:
CLA check still fails :(  Closing for now but comment here and fix the CLA issue and we can take a look.

PR #1219, opened 2016-02-21
Fixed bug in example: s/len(num_steps)/num_steps/g

NHDaly:
num_steps is a value (an integer), not a list, so it was a typo to take `len(num_steps)`.

This fixes that typo.


NHDaly:
I signed it!

vrv:
Merged

PR #1226, opened 2016-02-21
MNIST example using TFRecordWriter context manager.

eerwitt:
The `TFRecordWriter` wasn't being explicitly closed. Updated the example to use the context while writing the examples.

**NOTE**: This code will throw an error on any install of TensorFlow before [this commit](https://github.com/tensorflow/tensorflow/commit/67b2abcd673f466d32c4017ad8bab0d0b9ac86c5). The exception will likely be `AttributeError: 'NoneType' object has no attribute 'write'`.

Would it be better to call `TFRecordWriter#close()` explicitly on this example code since that would work with older version of TF as well?

### DIFF
--- a/tensorflow/examples/how_tos/reading_data/convert_to_records.py
+++ b/tensorflow/examples/how_tos/reading_data/convert_to_records.py
@@ -68,6 +68,7 @@ def convert_to(images, labels, name):
         'label': _int64_feature(int(labels[index])),
         'image_raw': _bytes_feature(image_raw)}))
     writer.write(example.SerializeToString())
+  writer.close()
 
 
 def main(argv):


josh11b:
My recollection is that when I put logging statements into the C++, I observed that the file was being closed properly.  This may have been a result of the Python object being unreferenced, though, and I suspect this is a better approach.  However, given that [the commit you mention](https://github.com/tensorflow/tensorflow/commit/67b2abcd673f466d32c4017ad8bab0d0b9ac86c5) does not appear to have been included in 0.7, I am a little worried about causing trouble for people who get the example from master but installed 0.7 via PiP or something.

eerwitt:
Updated to explicitly close the writer in the example.

I wouldn't doubt that the file is properly closed on the Python object being dereferenced. I'm afraid of people copying the pattern and leading to unintentional leaks.

I'm OK with this being closed and not merged but I think it'd be good to highlight that `TFRecordWriter` should be closed.

vrv:
Can you update to master and ping this thread when ready?

vrv:
Actually looks like there's no conflict.

@tensorflow-jenkins: test this please.

vrv:
Merged

PR #1229, opened 2016-02-21
Conditional logging for core platform flow

rekhajoshm:
Conditional logging for core platform flow(skeletal), quick take as i find users needing this.This pull is based on using re2 flags and allowing severity check before the logging.Please let me know your thoughts.

### DIFF
--- a/tensorflow/core/platform/default/logging.h
+++ b/tensorflow/core/platform/default/logging.h
@@ -18,6 +18,7 @@ limitations under the License.
 
 #include <sstream>
 #include "tensorflow/core/platform/port.h"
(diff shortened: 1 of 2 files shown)


vrv:
Thanks for this contribution @rekhajoshm, but we don't want to introduce command line flags to our library like this -- most options should be passed via ConfigProto or another proto.  We could maybe add a LogOptions structure to hold options related to logging.

Feel free to file an issue asking for a way to control log message output and we'll try to get it done.

rekhajoshm:
ok thanks @vrv raised #1258 for library handling the log control.thanks

PR #1223, opened 2016-02-21
Added skflow link to README

terrytangyuan:

PR #995, opened 2016-02-04
Added a link to some tutorial examples.

amcnamara:

PR #1232, opened 2016-02-22
Add list of community created projects

martinwicke:
... and point to it from the README. This addresses #995, #1223. We'd like to link only to things in this repo from the README, but it's a good idea to add a place for projects using or enhancing TensorFlow.


vrv:
Merged

PR #1237, opened 2016-02-22
Go bindings

tmc:
This adds initial support for go via swig.


tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (t *Tensor) Dim(n int) int {
+	return int(tf.TF_Dim(t.tensor, n))
+}
+
+func (t *Tensor) DataSize() int {
+	return int(tf.TF_TensorByteSize(t.tensor))
+}
+
```
dave-andersen: It feels a little more go-like to use int64 for datasize, in keeping with other interface that change size_t to an exported size.

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+		offset += copy(data[offset:], inLen)
+		offset += copy(data[offset:], inBytes)
+	}
+	return out
+}
+
+func Constant(value interface{}) (*Tensor, error) {
+	switch v := value.(type) {
+	case string:
+		//		defer C.free(unsafe.Pointer(str))
```
dave-andersen: lingering comment no longer needed?  'str' has no reference here.

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/golang/protobuf/proto"
+	tf "github.com/tensorflow/tensorflow/tensorflow/go"
+)
+
+type Session struct {
+	session tf.TF_Session
+}
+
+func NewSession() (*Session, error) {
```
dave-andersen: General comment:  Most of the functions provided here are missing doc-producing comments.  cf: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+
+type Tensor struct {
+	tensor tf.TF_Tensor
+	buf    []byte
+}
+
+type TensorShape [][]int64
+
+var (
+	TensorShapeScalar = TensorShape{{1}}
```
dave-andersen: could you comment why you're using this as a var?  I assume it's just optimization for Constant?

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+func NewTensor(dataType tf.DataType, shape TensorShape, data interface{}) (*Tensor, error) {
+	// TODO(tmc): ensure data is a slice
+	v := reflect.ValueOf(data)
+	if v.Kind() != reflect.Slice {
+		return nil, fmt.Errorf("tensorflow: 'data' argument must be a slice")
+	}
+	dataSize := v.Len() * int(v.Type().Elem().Size())
```
dave-andersen: I'm perplexed why this is getting cast to an int when it's going to be up-casted to an int64 when used.  (In general, I suspect go+tensorflow should just be all int64 for the sizes of things, unless there's a critical speed optimization I'm overlooking.)

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+
+func newTensor(dataType tf.DataType, shape TensorShape, data uintptr, size int) (*Tensor, error) {
+	t := &Tensor{
+		tensor: tf.TF_NewTensor_wrapper(tf.TF_DataType(dataType), &(shape[0][0]), len(shape), data, int64(size)),
+	}
+
+	return t, nil
+}
+
+func encodeStrings(in []string) []byte {
```
dave-andersen: This function could use a unit test.  (In general, I'd love to see more unit tests throughout the Go bindings.)

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+	if code != 0 {
+		return fmt.Errorf("tensorflow: %d: %v", code, message)
+	}
+	return nil
+}
+
+func (s *Session) Run(inputs map[string]*Tensor, outputs []string, targets []string) ([]*Tensor, error) {
+	inputNames := tf.NewStringVector()
+	inputValues := tf.NewTensorVector()
+	if inputs != nil {
```
dave-andersen: This test is unnecessary - range over a nil map just skips over it, so you can elide it and simply enter into the range loop.

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, n := range targets {
+		targetNames.Add(n)
+	}
+
+	outputValues := tf.NewTensorVector()
+	status := tf.TF_NewStatus()
+
+	tf.TF_Run_wrapper(s.session, inputNames, inputValues, outputNames, outputValues, targetNames, status)
+
+	result := []*Tensor{}
```
dave-andersen: Because you know the size (outputValues.Size()) in advance, I'd be tempted to make() the result the right size and directly set result[i] in the below list, but this is relatively minor.

tensorflow.go
```diff
@@ -0,0 +1,172 @@
...
+	return int(tf.TF_Dim(t.tensor, n))
+}
+
+func (t *Tensor) DataSize() int {
+	return int(tf.TF_TensorByteSize(t.tensor))
+}
+
+func (t *Tensor) Data() []byte {
+	length := t.DataSize()
+	return (*[1 << 30]byte)(unsafe.Pointer(tf.TF_TensorData(t.tensor)))[:length:length]
```
dave-andersen: Is 1 << 30 the right maximum size for a Tensor data?  I may be misreading, but the underlying c++ implementation has a 1<<40 max elements limit.  In any event, this should be commented - the 1<<30 is idiomatic to people who do cgo, but I'm not certain it's *correct*, and if it is, it should be explained for the next person to come along. :)

dave-andersen:
Thanks so much for contributing this, Travis!  I've put some code comments above;  most are nitpicky, but some should be given a bit of thought (such as the Right Way(tm) to represent sizes, and adding more tests -- we depend heavily upon having good test coverage to not accidentally break things during refactoring.)

tensorflow_test.go
```diff
@@ -0,0 +1,107 @@
...
+}
+version: 5`
+
+func TestNewSession(t *testing.T) {
+	graph := &tf.GraphDef{}
+	if err := proto.UnmarshalText(exampleGraph, graph); err != nil {
+		t.Fatal(err)
+	}
+	s, err := tensorflow.NewSession()
+	if err := s.ExtendGraph(graph); err != nil {
```
dave-andersen: To mirror the C++ API, maybe a quick bit of sugar to also support s.Create(graph)?

tensorflow_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.Fatal(err)
+	}
+	output, err := s.Run(nil, []string{"output1", "output2"}, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	fmt.Println(output)
+	for _, out := range output {
```
dave-andersen: Instead of Spewing the output - it seems like this could be turned into a more automated test that's also more helpful for users by adding a type switch and actually testing that the types of the outputs are what you expected based upon the input proto?

tensorflow/go/tf_tensor_helper.cc
```diff
@@ -0,0 +1,30 @@
...
+
+#include "tf_session_helper.h"
+
+namespace tensorflow {
+
+TF_Tensor* TF_NewTensor_wrapper(TF_DataType dtype, long long* dims, int num_dims,
+                   void* data, size_t len) {
+  
+  return TF_NewTensor(dtype, dims, num_dims, data, len, [](void *data, size_t len, void* arg){
+      puts("in tensor dealloc");
```
dave-andersen: This feels potentially incomplete?

dave-andersen:
Final final question (i hope):  What's the story for memory management w.r.t. calling Data on a tensor, and did you verify that the C allocations are getting properly deleted when the Go objects are destroyed?  (This may be obvious to someone with more recent swig experience than I!)

tmc:
Thank you for the insightful comments and review. Great suggestions. I'll get the amended.

martinwicke:
Also, can you move the whole tree to contrib/go please? We'd like for fast-moving code to live in contrib.

tmc:
@martinwicke I was considering the aesthetics of the import path on use

martinwicke:
I don't actually know much about how that would look like from the go side. We're giving guarantees about the API stability for things not in contrib, so we need to be confident things will stay as they are, and for large chunks like this one, I'd rather have them live in contrib for a while until things have settled with some actual use.

Another alternative would be a tensorflow/tensorflow-go repo which could point to tensorflow/tensorflow as a submodule.

dave-andersen:
Agreed.  I think a final question for this - perhaps one that's deferrable, particularly if we put it in contrib - is:  What to do with Tensors once they're in Go?  A quick check doesn't suggest that there are immediately obvious libraries to use beyond 2D matrices, but I might be missing something.

Serving in many cases can probably get by with extracting arrays of basic types, but it's worth thinking a bit about what the better general solution is.  I don't have a great idea yet.

dave-andersen:
Hi, tmc - just wanted to check on this, and note that if you'd like, I'm happy to help with some of the additional test code.  LMK.  I'd love to see this happen. :)

tensorflow.go
tmc: I intended it to be helpful to show the most basic use of the type.

tensorflow.go
tmc: No you're right, 1<<40 is correct

tensorflow_test.go
tmc: Yes I agree. This work was really just a quick push to show that the approach was functional and this is a prime example.

jmhodges:
(Looks like tmc responded to feedback but didn't know that other folks don't get notified when code is pushed up. Have to drop a comment like this one!)

vrv:
(Indeed, GitHub is quite annoying.  We'll try to get to this soon!)

dave-andersen:
Agh - I fail github 101 and didn't have my notifications configured properly.  I'll get on this towards the end of this week - I'm OOO.  Very sorry about that.

dave-andersen:
Reviewing and checking now!

tensorflow.go
```diff
@@ -24,11 +26,13 @@ func NewSession() (*Session, error) {
 	}, statusToError(status)
 }
 
+// Tensor represents a value created from an Operation
```
dave-andersen: missing period at end of comment.

tensorflow.go
```diff
@@ -24,11 +26,13 @@ func NewSession() (*Session, error) {
...
 type Tensor struct {
 	tensor tf.TF_Tensor
 	buf    []byte
 }
 
+// TensorShape represents the shapre of a Tensor.
```
dave-andersen: typo shapre->shape

dave-andersen:
Looks good aside from a few very minor typos, Martin's suggestion of moving to contrib/, and probably needing a few more tests.

One thought is that we could pull it in here and add a few tests as part of the commit process - I'd be happy to do that - or if you have them ready you could add.  LMK.  If I don't respond promptly, it's probably me failing github again, and you're welcome to ping me at dga (at google) dot com.

olostan:
Great work, @tmc ! Hope that will be merged soon. 

It would be great if someone, who successfully used TF with Go add a tiny how-to text about start using it (in tensorflow/go/README.md just description what it is, and test just verify bindings as I see). I am not as good with Python as with Go, but would like to start experimenting (and if I can, contribute some more docs). 

Thnx in advance!

alonsovidales:
Hello,
I've opened this PR today: https://github.com/tensorflow/tensorflow/pull/1771/
It includes most of the modifications commented here:
- All the code is now contained into contrib/go
- Improved test coverage
- Added methods to create tensors, populate them, read the content, etc.
- Graph generation, now it is possible to build the Graph from Go
- Modified the "Image Recognition" tutorial to include a section similar to the C++ API section but using the Go API, also the code was ported to Go
- Code updated to work with the 7.1 version
- Code commented, GoLint doesn't complains

Thanks,

dave-andersen:
@alonsovidales - got it, moving my attention over there.  thank you!

tmc:
Updated my goog CLA with my github username. Commenting to hopefully trigger the bot.

gunan:

tmc:
Closing this as @alonsovidales has continued building on this at #1771

PR #1255, opened 2016-02-23
update_version.sh: automatic version update in source

caisq:
Usage example:
  update_version.sh 0.8.0

### DIFF
--- a/tensorflow/tools/ci_build/update_version.sh
+++ b/tensorflow/tools/ci_build/update_version.sh
@@ -1,0 +1,126 @@
+#!/usr/bin/env bash
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


tensorflow/tools/ci_build/update_version.sh
```diff
@@ -0,0 +1,108 @@
...
+# ==============================================================================
+# 
+# Automatically update TensorFlow version in source files
+# 
+# Usage:  update_version.sh <new_major_ver>.<new_minor_ver>.<new_patch_ver>
+#         e.g.,
+#           update_version 0.7.2
+
+# Helper functions
+check_existence() {
```
martinwicke: What does this add over test?

test -d $1 || die "Error..."

martinwicke:
Yes!

caisq:
Addressed @martinwicke's comment. Added safety check for lingering old version strings and warning if those are found.

martinwicke:
merged.

PR #1260, opened 2016-02-23
Added sqrt to LRN doc equation.

cesarsalgado:


cesarsalgado:
Actually there should be no sqrt in the formula, but the beta should be moved outside the parentheses. I will close this one and make another one.

Issue #824, opened 2016-01-20
Feature parity of tf.diag with np.diag


yaroslavvb:
np.diag can

1) take 2d array and produce 1d vector of diagonal entries
2) take 1d array and produce diagonal 2d array (inverse of case 1)

tf.diag only supports 2) and the behavior for case 1) is undocumented

Currently to do 1) you need to do something like this:

    def identity_matrix(n):
      """Returns nxn identity matrix."""
      # note, if n is a constant node, this assert node won't be executed,
      # this error will be caught during shape analysis 
      assert_op = tf.Assert(tf.greater(n, 0), ["Matrix size must be positive"])
      with tf.control_dependencies([assert_op]):
        ones = tf.fill(n, 1)
        diag = tf.diag(ones)
      return diag
    
    def extract_diagonal(tensor):
      """Extract diagonal of a square matrix."""
      
      shape = tf.shape(tensor)
      n = shape[0]
      assert_op = tf.Assert(tf.equal(shape[0], shape[1]), ["Can't get diagonal of "
                                                           "a non-square matrix"])
      
      with tf.control_dependencies([assert_op]):
      return tf.reduce_sum(tf.mul(tensor, identity_matrix(n)), [0])

josh11b:
I believe tf.diag is only intended to support 2).  It would be reasonable to create a new op for 1).

girving:
Yeah, having the same op arbitrarily be its own one-sided inverse seems pretty strange.  Two ops sound better to me.

yaroslavvb:
Besides numpy, this also how it's done in Theano, Torch, Matlab, ViennaCL

girving:
In my defense, that's four different things copying Matlab's arguably sketchy choice, but I see your point.  I'm okay if we bow to peer pressure.

chemelnucfin:
@girving This issue may be closed?

girving:
Yep, closing.

girving:
@chemelnucfin: For future use, you can include a line "Fixes #824" in the commit message in order to automatically close the issue when it's merged.

PR #1265, opened 2016-02-24
DiagPart Operator

chemelnucfin:
Implemented the get diagonal GetDiag operator #824

I believe numpy is also trying to have two different operators, but they are using diagonal and diagflat for their names.  GetDiag sounds reasonable to me, but discussion welcome.

### DIFF
--- a/tensorflow/core/kernels/diag_op.cc
+++ b/tensorflow/core/kernels/diag_op.cc
@@ -45,6 +45,28 @@ class DiagonalGenerator {
  private:
   Tensor diagonal_;
 };
+
+template <typename T, size_t NumDims>
+class DiagonalExtractor {
+ public:
+  explicit DiagonalExtractor(const Tensor& tensor) : tensor_(tensor) {
+    CHECK_EQ(tensor.dims(), 2 * NumDims);
+  }
+  T operator()(const Eigen::array<Eigen::Index, NumDims>& coordinates) const {
+    Eigen::array<Eigen::Index, 2 * NumDims> index;
+    for (size_t j = 0; j < NumDims; ++j){
+      index[j] = coordinates[j];
+    }
+    for (size_t j = NumDims; j < 2 * NumDims; ++j){
+      index[j] = index[j - NumDims];
+    }
+    return tensor_.tensor<T, 2 * NumDims>()(index);
+  }
+
+ private:
+  Tensor tensor_;
+};
+  
 }  // namespace
 
 // Generate the diagonal tensor with the diagonal set to the input tensor.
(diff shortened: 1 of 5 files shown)


girving:
Thank you for the contribution!  I added a bunch of comments.  Let me know when I should take another look.

chemelnucfin:
wow, thanks a lot for the detailed comments.  I'll work on the fixes and resubmit.  Since it hasn't merged yet, is it ok if I do a force push to squash the commits or leave them separately?

girving:
A force push is good.

chemelnucfin:
Seems like a forced push erases your line comments, unfortunately.  I have addressed your concerns earlier.  Including indentation issues, removed unnecessary template parameter, documentation, and test matrices. 

One thing I realized from looking at numpy code, shouldn't diagonals be where all the indices are the same instead of how it is now?  I mostly reversed the diag operator for this commit though.

Thanks.

girving:
Thanks, getting much closer!  I guess add commits for now if forcing removes old comments, and once everything is good we can squash into one.

Mistobaan:
I was thinking if `get_diag` a good name for the operation ?  reading the doc and having `diag` and `get_diag` feels a bit confusing

girving:
@Mistobaan: I don't much like it either, but I haven't thought of something good yet.  Do you have suggestions?

girving:
Ah, got it!  `diag_part`.

chemelnucfin:
Diagflat in numpy is actually the "diag" operation here, (which diagflat doesn't make sense to me either).  Having diag_part might be confusing to numpy users.  I thought of using ExtractDiag for this commit and CreateDiag for the existing "diag" operation, but not sure if you wanted names that long.

girving:
diag_part is mathematically unambiguous and matches common usage ("the diagonal part of A").  It's unfortunately that numpy is different, but `get_diag` is just as bad there.

chemelnucfin:
Ok, give me a few minutes to finish working on the rintop pull request and
I'll change the name in this one.

chemelnucfin:
Thanks for the time spent reviewing the code!  Hopefully this is good for now.

Should I add offset in the future in both diag and diag_part?

girving:
Looks good, thank you!  @vrv: How does one run tests and get this merged?

girving:
Yep, an offset would be a reasonable addition for a future CL, but it has the unfortunate issue that it would ideally be an `input` rather than an `attr`, and adding new inputs is awkward to do in a backwards compatible way.

Also, before we merge (but perhaps after the tests?) we should squash this into one commit.  Thanks for keeping it separated for now to make it easier to incrementally review.

chemelnucfin:
@girving I think you just say "jenkins, test this please" and it will trigger the tests.

So, should I work on offset for this first before we merge?  

Also, for both of these operations should there be a value to indicate the diagonal that is created and extracted?
For example:
dim=0 => A[i, i, i, i, i, i]
dim=1 => A[i, j, i, j, i, j]
dim=2 => A[i, j, k, i, j, k]
etc.

vrv:
only admins can trigger it :)

girving:
Let's leave generalizations to a separate CL.  The symmetry of your existing change is good, since it doesn't change the existing op and adds it's one sided inverse.

chemelnucfin:
I just squashed the commits.

vrv:

vrv:
Merged

PR #1275, opened 2016-02-24
Adding permission x to test_tutorials.sh

caisq:


caisq:
This change needs to be made so that tutorial tests can work on Jenkins. I guess the file permission was somehow not set correctly when it is pushed from internal.

PR #1278, opened 2016-02-24
In LRN docs eq, moved beta outside the parenthesis

cesarsalgado:
In Krizhevsky 2012 paper, the beta in LRN equation is the exponent
of everything inside the parenthesis and not just sqr_sum.


cesarsalgado:
Closing because it was fixed in ab48dbd4a.

PR #1294, opened 2016-02-26
Automated Docker image build and test

caisq:
1) Breaking up the pip.sh into the install and test-on-install

This is aimed at automating testing of docker images.
Also, a step is added in pip.sh to uninstall existing versions of
protobuf and tensorflow if they exist. This addresses pip install-test
issues on non-Docker environments (e.g., Mac)

2) Using virtualenv to perform pip test-on-install

The built pip-package is installed in a virtualenv, then the pip install test and the tutorial
test are carried out inside the virutal env. This is done inside the same docker container as the one used for bazel build, if docker is available. 

3) Automating Docker image build and test

ci_parameterized_build now has a new env-var parameter:
TF_BUILD_DOCKER_TEST, that can be used to trigger Docker image build
and test following the PIP build (i.e., TF_BUILD_IS_PIP=PIP or BOTH).

The image is automatically tagged as ${USER}/tensorflow:${VERSION},
wherein the version is extracted from version.h.

The main script for Docker build and test is the newly added:
docker_test.sh

It uses Dockerfiles in tools/ci_build/docker and installs the whl files
inside the Docker image. After the pip installation step, the Python
unit tests and tutorial tests are inside the Docker container.


tensorflow/tools/docker/Dockerfile
```diff
@@ -29,8 +30,8 @@ RUN pip --no-cache-dir install \
 
 # Install TensorFlow CPU version.
 ENV TENSORFLOW_VERSION 0.7.1
-RUN pip --no-cache-dir install \
-    http://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-${TENSORFLOW_VERSION}-cp27-none-linux_x86_64.whl
+ADD whl/tensorflow-*.whl /tmp/
```
jendap: The problem with doing this is that people can simply build docker in tools/docker folder anymore. They would have to know they have to run multiple hops of other scripts. Complicated.

tensorflow/tools/docker/Dockerfile.gpu
```diff
@@ -1,9 +1,10 @@
-FROM nvidia/cuda:7.5-cudnn4-runtime
+FROM nvidia/cuda:7.5-cudnn4-devel
```
jendap: This container should not depend on devel as it would make it a lot bigger. And this container is what we recommend people to use for runtime. It should be small.

tensorflow/tools/ci_build/builds/pip_test.sh
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env bash
```
jendap: Nice. It is good to pull out the test only script. No compilation or installation here.

Should it be called test_installation.sh? Or something like that? I can be used to run python test on tf installed anywhere by any means.

tensorflow/tools/ci_build/builds/docker_test.sh
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env bash
```
jendap: What is the goal of this? Are we trying to automate the release process? Or are we trying to run python tests for every build (master, pull requests, ...) in better isolation?

For the better isolation it would be better (simpler, multiplatform, easier and faster) to use virtualenv. We can make release more automated later.

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -72,12 +37,56 @@ abs_path() {
...
 
+
 # Exit after a failure
 die() {
     echo $@
     exit 1
 }
 
+# Uninstall existing pip packages
+uninstall_existing_pip_packages() {
```
jendap: The uninstall sounds a bit dangerous. You would never need it if it was intended to run only in inside per build docker container. Because there would be nothing to uninstall in the first place. Also this does not allow concurrent builds outside docker (think mac os).

It would be great to think about any developer being able to use it on his/her computer. Without being afraid of messing up system. Hence no uninstall. Would it work if we would create per build virtualenv for python?

One can have working version of tensorflow installed on system and develop tensorflow on the same machine then.

tensorflow/tools/ci_build/builds/docker_test.sh
caisq: The main goal of this file (docker_test.sh) is to automate the docker image release process. I have now made further changes to use virtualenv.

tensorflow/tools/docker/Dockerfile.gpu
caisq: I changed it to devel because otherwise there will be a couple of test failures. Some tests depend on devel headers and libraries. Maybe I can just blacklist those tests here?

tensorflow/tools/docker/Dockerfile
caisq: Unfortunately it's difficult to parameterize Dockerfiles. How about this, I can move this content into Dockerfile.devel (replacing the existing ones) and restore the http URL in this file?

jendap: It would be nice if all the docker files just work. But we do not need this do we? We will use virtualenv for normal builds anyway. We will want to test this only when we do a release, right? We can script that later and differently. We can simply build and upload the pips before building the docker images.

tensorflow/tools/docker/Dockerfile.gpu
jendap: Really? Which tests depends on it? I thought it runs python scripts. The devel stuff should be only needed if you compile some c/c++ code. And that should not be happening.

If it really is a problem for a few tests then I would ignore them here.

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -140,189 +125,47 @@ echo "whl file path = ${WHL_PATH}"
 # Install, in user's local home folder
 echo "Installing pip whl file: ${WHL_PATH}"
 
-# Call pip install on the whl file. We are doing it without the --upgrade
-# option. So dependency updates will need to be performed separately in
-# the environment.
-${PYTHON_BIN_PATH} -m pip install -v --user ${WHL_PATH} \
+# Create temporary directory for install test
+VENV_DIR="${PIP_TEST_ROOT}/venv"
+rm -rf "${VENV_DIR}" && mkdir -p "${VENV_DIR}"
```
jendap: Does pip inside virtualenv work reliably without deleting the whole env? It would be nice for caching. We would reinstall only tf and dependencies only if their version would change. But I'm not sure if it would be reliable.

tensorflow/tools/ci_build/builds/pip_test.sh
```diff
@@ -0,0 +1,297 @@
...
+# Determine the major and minor versions of Python being used (e.g., 2.7)
+# This info will be useful for determining the directory of the local pip
+# installation of Python
+PY_MAJOR_MINOR_VER=$(${PYTHON_BIN_PATH} -V 2>&1 | awk '{print $NF}' | cut -d. -f-2)
+
+echo "Python binary path to be used in PIP install-test: ${PYTHON_BIN_PATH} "\
+"(Major.Minor version: ${PY_MAJOR_MINOR_VER})"
+
+# Avoid permission issues outside container
+umask 000
```
jendap: That's weird. What issues?

tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -50,9 +50,18 @@
...
 #                      resource contention (e.g., for GPU builds)
 #   TF_BUILD_TEST_TUTORIALS:
-#                      Perform tutorials test (Applicable only if
-#                      TF_BUILD_IS_PIP is PIP or BOTH).
-#                      See build/test_tutorials.sh
+#                      If set to any non-empty and non-0 value, will perform
+#                      tutorials tests (Applicable only if TF_BUILD_IS_PIP is
+#                      PIP or BOTH).
+#                      See builds/test_tutorials.sh
+#   TF_BUILD_DOCKER_TEST:
```
jendap: Do you really want to put the release docker testing into this script? There is already way too many things here, is it not?

tensorflow/tools/ci_build/install/install_deb_packages.sh
```diff
@@ -38,3 +38,5 @@ apt-get install -y \
     zlib1g-dev
 apt-get clean
 rm -rf /var/lib/apt/lists/*
+
+pip install virtualenv
```
jendap: Would python-virtualenv from apt-get work? It would be better to use the linux distro version if possible.

jendap:
Splitting the test out of pip.sh is good. Using virtualenv is good. Testing the docker images is, imho, nice to have but not critical for now. We should not change the docker files. Also putting docker files testing into parametrized build is, imho, an overkill.

tensorflow/tools/ci_build/builds/pip_test.sh
caisq: When the docker container runs, the user is root. It turns out that directories created inside the container will retain root owner ship even outside the container.

caisq:
Addressed @jendap's comments. Closing this PR and opening a new fresh one with all changes squashed. https://github.com/tensorflow/tensorflow/pull/1380

PR #1314, opened 2016-02-27
fix #1303 doc link broken

### LINKED ISSUE #1303: Broken links in C++ API docs
Apparently most of the links in the main part of the [index of the C++ API documentation](https://www.tensorflow.org/versions/r0.7/api_docs/cc/index.html) are broken.

The problem seems to be that the first letter of each file name is in lowercase, so the links look like this:
[https://www.tensorflow.org/versions/r0.7/api_docs/cc/classEnv.html](https://www.tensorflow.org/versions/r0.7/api_docs/cc/classEnv.html)
Instead of this:
[https://www.tensorflow.org/versions/r0.7/api_docs/cc/ClassEnv.html](https://www.tensorflow.org/versions/r0.7/api_docs/cc/ClassEnv.html)

chemelnucfin:
capitalize name of type in C++ doc links, referencing #1303

### DIFF
--- a/tensorflow/tools/docs/gen_cc_md.py
+++ b/tensorflow/tools/docs/gen_cc_md.py
@@ -275,7 +275,8 @@ class Page(object):
     return self.type
 
   def get_md_filename(self):
-    return self.get_type() + anchorize(self.get_short_name()) + '.md'
+    deftype = self.type[0].upper() + self.type[1:]
+    return deftype + anchorize(self.get_short_name()) + '.md'
 
 
 def main(unused_argv):


tensorflow/tools/docs/gen_cc_md.py
```diff
@@ -275,7 +275,8 @@ def get_type(self):
     return self.type
 
   def get_md_filename(self):
-    return self.get_type() + anchorize(self.get_short_name()) + '.md'
+    deftype = self.type[0].upper() + self.type[1:]
```
martinwicke: I just made that same change. I think that's the second commit we're crossing on just today. :)

My change is upstream, it'll get to github in the next push out of google.

vrv:
Already did: https://github.com/tensorflow/tensorflow/commit/55be24524167c1d1879dba64b492c33c82244d1f

tensorflow/tools/docs/gen_cc_md.py
chemelnucfin: @martinwicke  

:)
For the rint, I just thought that you were working on more important things.  Didn't mean to cross your commits.  
:)

PR #1335, opened 2016-02-29
Fix word2vec_basic.py example to be python 3 compatible

timsl:
Fixed word2vec_basic.py example to be both Python 2 & 3 compatible.


vrv:
I think I like this one more than #1345.  can you rebase?

vrv:
Nm, the icon they switched to looks like the conflict icon.  @tensorflow-jenkins: test this please

vrv:
Merged

PR #1338, opened 2016-03-01
Master

hanyo:
translate.pyの中の
outputs=[int(np.argmax(logit,axis=1))for logit in output_logits]
ないにenumerate関数を組み込む


sesse:
こんにちは,

This pull request appears to be empty, consisting of just two merges. Also, the description is in Japanese—did you intend to send it in its current form? If so, please rewrite to English :-)

PR #1357, opened 2016-03-02
Retraining Example with Notebook

kmader:
The code from the retraining demo morphed into a IPython Notebook to make it easier to use interactively.


kmader:
CLA has been signed for kmader

vrv:
@kmader: I like this, but right now we don't have a good way to test ipython notebooks, which means it's very hard to keep updated if we make changes.  The udacity notebooks already often break, and I'm worried about adding more notebooks at this point, but I could be convinced otherwise.  @martinwicke

kmader:
@vrv not sure how @googlebot works but we have jenkins setup to [extract the python code](https://ipython.org/ipython-doc/1/interactive/nbconvert.html) from the notebooks and make sure it executes without throwing any errors. Alternatively it is fairly easy to construct the notebooks from annotated python code, so maybe that is a more viable option. The biggest issue there is the ```FLAGS``` in TF needs to be reworked.

```bash
ipython nbconvert --to python notebook.ipynb notebook.py
```

martinwicke:
Flags is just a thin wrapped around argparse, in what way would it have to
be reworked?

I'm quite excited by the idea of tested ipython notebooks, it's been
something we've been thinking about for our tutorials (although we've gone
with regular tests for now).

vrv:
friendly ping -- what needs to be reworked?

Also, @craigcitro for general advice on migrating our tutorials to testable notebooks rather than having copies that will inevitably diverge.

martinwicke:
Is this PR still active?

vrv:
We want to move our tutorials to notebooks, but I guess we're not ready now.  Thanks for this PR -- we'll try to get to this for all of our tutorials eventually.

PR #1358, opened 2016-03-02
Mean of cross entropy vs sum in summaries tutorial

basveeling:
The `mnist_with_summaries.py` and `mnist_softmax.py` tutorials compute the sum of the cross_entropy over the data points in the batch. This seems to deviate from the more common approach of taking the mean, such as in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/mnist/mnist.py#L99 
The current form works as a loss function, but might confuse newcomers who are figuring out how gradients are aggregated in TensorFlow (as experienced by me, see http://stackoverflow.com/questions/35731506/unaggregated-gradients-gradients-per-example-in-tensorflow :-))


yaroslavvb:
Perhaps one reason that reduce_sum was used is that reduce_mean is not yet implemented on GPU.

basveeling:
Fair enough. Do you think it's a good idea to make a note about this in the example code or the [documentation](https://www.tensorflow.org/versions/r0.7/api_docs/python/math_ops.html#reduce_mean) to clear this up? Looking at some other frameworks, taking a mean over cross entropy seems to be customary[1,2], so I can imagine this could potentially confuse others as well. 
- [1] https://github.com/Lasagne/Lasagne/blob/master/README.rst
- [2] https://github.com/mila-udem/blocks/blob/58994253579d30fb63a2cfa53499a68b1237c3c0/blocks/bricks/cost.py#L59

vrv:
Just pushed b3dfff2a23b435f14cca7377f8e0e5ad7583b45e which added reduce_mean support for GPU, so that will no longer be the reason.  @vincentvanhoucke does this look good?

vincentvanhoucke:
Does this converge to the same result though? Our learning rate is not scale invariant.

vrv:
@colah: friendly ping?

Sohl-Dickstein:
@basveeling I agree that switching from a sum to a mean across datapoints is better. Can you:
- Make the corresponding changes to the tutorial text https://github.com/tensorflow/tensorflow/tree/master/tensorflow/g3doc/tutorials/mnist
- Confirm that the model converges to the same result after making this change. (You will probably need to multiply the learning rate by the batch size of 100.)

basveeling:
@Sohl-Dickstein: done. I determined the new learning rate for the softmax tutorial by cross validation:
https://gist.github.com/basveeling/0b43d5d9fe533e9aaf710d13356931f4
I haven't validated the results for the convolutional mnist network in the pros tutorial due to lack of gpu, but someone can verify this with https://gist.github.com/basveeling/80b11bf657891c0c8f253c899a4be99d .

tensorflow/g3doc/tutorials/mnist/pros/index.md
```diff
@@ -157,11 +157,12 @@ easily. Our cost function will be the cross-entropy between the target and the
...
 
 ```python
-cross_entropy = -tf.reduce_sum(y_*tf.log(y))
+cross_entropy = tf.reduce_mean(-tf.reduce_sum(y_ * tf.log(y), reduction_indices=[1]))
 ```
 
-Note that `tf.reduce_sum` sums across all images in the minibatch, as well as
-all classes. We are computing the cross entropy for the entire minibatch.
+Note that `tf.reduce_sum` sums across all classes and `tf.reduce_mean` takes 
+the average over these sums. We are computing the cross entropy for the entire
```
Sohl-Dickstein: Delete the last sentence "We are computing the cross entropy for the entire minibatch."

(It was a bit of an odd phrasing -- but it was intended to convey that we were computing the cross entropy as if a sample consisted of an entire minibatch -- and this is no longer true.)

tensorflow/g3doc/tutorials/mnist/beginners/index.md
```diff
@@ -311,12 +311,14 @@ y_ = tf.placeholder(tf.float32, [None, 10])
...
 
 First, `tf.log` computes the logarithm of each element of `y`. Next, we multiply
-each element of `y_` with the corresponding element of `tf.log(y)`. Finally,
-`tf.reduce_sum` adds all the elements of the tensor.
+each element of `y_` with the corresponding element of `tf.log(y)`. Then 
+`tf.reduce_sum` adds the elements in the second dimension of y, due to the 
+`reduction_indices=[1]` parameter. Finally,  `tf.reduce_mean` computes the mean
+over all the examples in the batch.
 
 Note that this isn't just the cross-entropy of the truth with a single
```
Sohl-Dickstein: Delete this sentence "Note that this isn't just the cross-entropy of the truth with a single prediction, but the sum of the cross-entropies for all the images we looked at." (it's no longer true)

Sohl-Dickstein:
Thank you for the submission @basveeling.

See the inline requests for minor text changes. Also, both tutorials reference a performance of the naive model of 91%. Thanks to your cross validation, this is now 92% -- so you should make that change too.

Otherwise it looks good. Thanks again.

tensorflow/g3doc/tutorials/mnist/beginners/index.md
basveeling: Done. Thanks, missed that part.

tensorflow/g3doc/tutorials/mnist/pros/index.md
basveeling: Done.

basveeling:
@Sohl-Dickstein Thanks, should be good now I left the performance at 91% in the docs so users wouldn't feel bad when they incidentally get a lower score :-), but updated it to 92 now.

martinwicke:
Jenkins, test this please.

martinwicke:
Thanks!

PR #1380, opened 2016-03-04
Changes to pip.sh; adding docker_test.sh

caisq:
1) Breaking pip.sh into the install step (pip.sh) and the test step
(newly added test_installation.sh)

pip.sh now performs only the building and install of the pip package,
with the install happening inside virtualenv. (vritualenv address the
reinstallation issue on Mac pip tests). Then it calls
test_installation.sh and test_tutorials.sh to test the virtualenv pip
install of TensorFlow.

2) Adding a new file, docker_test.sh, as a first step toward automation
of TensorFlow docker image build and test process.

This script builds and tests docker images with TensorFlow pip whl
installed. The tests include the Python unit tests and tutorial tests
against the (non-virtualenv) install. This is not fully functional yet, because
we need to further automate the pip package uploading process, so the
Dockerfiles in tensorflow/tools/docker can point to the latest pip
files.

### DIFF
--- a/tensorflow/tools/ci_build/builds/docker_test.sh
+++ b/tensorflow/tools/ci_build/builds/docker_test.sh
@@ -1,0 +1,127 @@
+#!/usr/bin/env bash
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+#
+# Build and test TensorFlow docker images.
+# The tests include Python unit tests-on-install and tutorial tests.
+#
+# Usage: docker_test.sh <IMAGE_TYPE> <TAG> <WHL_PATH>
+# Arguments:
+#   IMAGE_TYPE : Type of the image: (CPU|GPU)
+#   TAG        : Docker image tag
+#   WHL_PATH   : Path to the whl file to be installed inside the docker image
+#
+#   e.g.: docker_test.sh CPU someone/tensorflow:0.8.0 pip_test/whl/tensorflow-0.8.0-cp27-none-linux_x86_64.whl
+#
+
+# Helper functions
+# Exit after a failure
+die() {
+  echo $@
+  exit 1
+}
+
+# Convert to lower case
+to_lower () {
+  echo "$1" | tr '[:upper:]' '[:lower:]'
+}
+
+
+# Helper function to traverse directories up until given file is found.
+function upsearch () {
+  test / == "$PWD" && return || \
+      test -e "$1" && echo "$PWD" && return || \
+      cd .. && upsearch "$1"
+}
+
+
+# Verify command line argument
+if [[ $# != "3" ]]; then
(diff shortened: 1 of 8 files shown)


tensorflow/tools/ci_build/builds/docker_test.sh
```diff
@@ -0,0 +1,126 @@
...
+DOCKERFILE_PATH="${BASE_DIR}/${DOCKERFILE}"
+DOCKERFILE_DIR="$(dirname ${DOCKERFILE_PATH})"
+
+# Check to make sure that the whl file exists
+test -f ${WHL_PATH} || die "whl file does not exist: ${WHL_PATH}"
+
+TMP_WHL_DIR="${DOCKERFILE_DIR}/whl"
+mkdir -p "${TMP_WHL_DIR}"
+cp "${WHL_PATH}" "${TMP_WHL_DIR}/" \
+|| die "FAILED to copy whl file from ${WHL_PATH} to ${TMP_WHL_DIR}/"
```
jendap: nit: the || is usually written at the end of first line ( ... || \ ) and next line have some (four) spaces at the beginning

tensorflow/tools/ci_build/builds/docker_test.sh
```diff
@@ -0,0 +1,126 @@
...
+
+docker build -t "${DOCKER_IMG_TAG}" -f "${DOCKERFILE_PATH}" \
+"${DOCKERFILE_DIR}" \
+|| die "FAILED to build docker image from Dockerfile ${DOCKERFILE_PATH}"
+
+# Clean up
+rm -rf "${TMP_WHL_DIR}" ||
```
jendap: \ at the end?

tensorflow/tools/ci_build/builds/docker_test.sh
```diff
@@ -0,0 +1,126 @@
...
+die "Failed to remove temporary directory ${TMP_WHL_DIR}"
```
jendap: should these lines have four spaces at the beginning? (two indents for continuations of previous line)

tensorflow/tools/ci_build/builds/test_installation.sh
```diff
@@ -0,0 +1,293 @@
...
+LIB_PYTHON_DIR=$(dirname $(dirname $(dirname "${TF_FILE}")))
+
+if [[ -z ${LIB_PYTHON_DIR} ]]; then
+  die "Failed to find Python library directory where TensorFlow is installed."
+else
+  echo "Found Python library directory where TensorFlow is installed: "\
+"${LIB_PYTHON_DIR}"
+fi
+
+PACKAGES_DIR=$(ls -d ${LIB_PYTHON_DIR}/*-packages | head -1)
```
jendap: Why are the 20 lines above and bellow so complicated? The PACKAGES_DIR computation currently looks fragile. Would this work?
TF_INSTALLATION_PATH=$(dirname $("${PYTHON_BIN_PATH}" -c "import tensorflow as tf; print(tf.__file__)"))

Note: No need of PACKAGES_DIR in the first place as TF_INSTALLATION_PATH=${PACKAGES_DIR}/tensorflow

And everything from pushd can be gone.

jendap:
Nice!

But let's make the line wrapping good. I see it was broken on many places even before this commit (at least in parametrized_build). We normally indent by two spaces. Hence when the line is too long and continues on next one it should have the same indentation PLUS 4 spaces. And all the || and && should be at the end of the lines not beginnings. Right?

caisq:
Addressed @jendap's latest comments. Tested on Jenkins. This PR is ready to be merged.

jendap:
Thank you!

Nit: There are still a few multiline statements without intending the next line by two indents. For example
```
  die "FAILED: Unable to find the base directory where the dockerfile "\
"${DOCKERFFILE} resides"
```
is missing 6 spaces on the second line. And there are others.

jendap:

PR #1384, opened 2016-03-04
Improvements to Issue template language

vrv:
Points people to StackOverflow for non bugs / features  as an attempt to try to keep github issues mostly for bugs / feature tracking instead of general support.


vrv:
Merged

PR #1393, opened 2016-03-05
Added GPU implementation for resize nearest neighbor grad.

panmari:
And also adapted/added appropriate tests.

### DIFF
--- a/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
@@ -24,6 +24,7 @@ limitations under the License.
 #include "tensorflow/core/framework/tensor_types.h"
 #include "tensorflow/core/kernels/maxpooling_op.h"
 #include "tensorflow/core/kernels/maxpooling_op_gpu.h"
+#include "tensorflow/core/util/cuda_kernel_helper.h"
 
 namespace tensorflow {
 namespace {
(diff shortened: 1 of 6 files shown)


panmari:
I needed to adapt the test so they use `np.float` instead of `np.float32`, e. g. [here](https://github.com/tensorflow/tensorflow/pull/1393/files#diff-bb1fdb8f5be7a85b4505589762579bf7R31). Isn't that a bit concerning? Shouldn't these types be equivalent? Or at least the op fall back to CPU if it's not supported on the GPU?

martinwicke:
np.float is an alias for float64, not float32. Surprised me too. http://docs.scipy.org/doc/numpy-1.10.1/user/basics.types.html

panmari:
Oh, that's bad news. That means the tests are not using the new written kernel, but are falling back to the CPU version. I'll have to look into that and add a [WIP] for the time being.

panmari:
@martinwicke I fixed the PR and tests. Could you take another look at this? I also added the kernel for `SetZero` from `maxpooling_op_gpu.cu.cc` to `tensorflow/core/util/cuda_kernel_helper.h` and consequently removed some duplicated code there after including `cuda_kernel_helper.h`.

martinwicke:
Thanks for moving SetZero, that's good. Can you add a blank line after it? Otherwise looks good to me. You can amend your last commit with that, then I'll merge.

panmari:
Done.

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please

PR #1395, opened 2016-03-06
#1392 fix mnist tutorial doc typo

### LINKED ISSUE #1392: Minor Error in Tutorial
On [this tutorial page](https://www.tensorflow.org/versions/master/tutorials/mnist/download/index.html), it says that the MNIST data has been rescaled to [-.5, .5]:

"The image data is extracted into a 2d tensor of: [image index, pixel index] where each entry is the intensity value of a specific pixel in a specific image, rescaled from [0, 255] to [-0.5, 0.5]."

However, in [the code](https://github.com/tensorflow/tensorflow/blob/r0.7/tensorflow/examples/tutorials/mnist/input_data.py), the data is rescaled to [0,1] as per the comment on line 123 (and confirmed by loading a sample image and inspecting):

```# Convert from [0, 255] -> [0.0, 1.0].```

Just a minor error, but thought it was worth pointing out! I'm actually curious which is the better method of rescaling, but that's a topic for another discussion..

cameronphchen:
fix #1392 tutorial typo

### DIFF
--- a/tensorflow/g3doc/tutorials/mnist/download/index.md
+++ b/tensorflow/g3doc/tutorials/mnist/download/index.md
@@ -50,7 +50,7 @@ unpacked (following the instructions available at the website) by the
 
 The image data is extracted into a 2d tensor of: `[image index, pixel index]`
 where each entry is the intensity value of a specific pixel in a specific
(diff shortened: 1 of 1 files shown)


martinwicke:
Merged. Thanks!

PR #1400, opened 2016-03-06
Fix raising wrong exception.

dongjoon-hyun:


martinwicke:
Good catch. Clearly, this wasn't properly tested (that line never executed), could you add a unit test for the failure behavior?

dongjoon-hyun:
Oh, in this case, I'm afraid that it might be too much work(lines) for this tiny error. :-)

By the way, is `Tensorflow` tested by any coverage test tool like `Coverity` now?

martinwicke:
Not coverity, but we do measure coverage. It was around 96% or so last time I checked.

I'll merge this.

dongjoon-hyun:
Thank you, @martinwicke !

PR #1403, opened 2016-03-06
Fix types image ops tests

panmari:
Thanks to a recent comment by @martinwicke I realized that there's something wrong with the tests of `test_image_ops.py`. Combined with the use of `use_gpu` instead of `force_gpu`, the tests that were supposed to test the GPU implementation actually ran the CPU implementation. 

After rectifying this, two bugs were found in the GPU implementation and corrected.

### DIFF
--- a/tensorflow/core/kernels/resize_nearest_neighbor_op.cc
+++ b/tensorflow/core/kernels/resize_nearest_neighbor_op.cc
@@ -281,9 +281,7 @@ class ResizeNearestNeighborGPUOp : public OpKernel {
                               .HostMemory("size"),                \
                           ResizeNearestNeighborGPUOp<T>);
 
-REGISTER_KERNEL(float);
-// TODO(panmari): Fix kernel for double.
-//REGISTER_KERNEL(double);
+TF_CALL_GPU_NUMBER_TYPES(REGISTER_KERNEL);
 
 #undef REGISTER_KERNEL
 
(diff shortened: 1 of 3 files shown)


martinwicke:
`force_gpu` will fail the test if there is no GPU available (e.g. if you compile without cuda, or if you don't have a supported GPU). It's really for debugging, not particularly useful in tests. The absence of something in between of `force_gpu` and `use_gpu` is a bit of a bug: we do need something that will fail only if there is a GPU, and TF is compiled with --config=cuda, but it's not used.

panmari:
I see, thanks for your insight. I changed the tests back to using `use_gpu`. I agree, it would really be useful to have a flag along the lines of `force_gpu_if_available`.

martinwicke:
I made #1423 to track this.

martinwicke:
Jenkins, test this please.

martinwicke:
hmm... not sure what happened with the tests. Checking.

panmari:
For me it looks like it's properly executing the tests...

martinwicke:
Yeah, it was just waiting in the queue.

chemelnucfin:
@martinwicke I do get saver_test problems on my python3.4.3 (fedora) also.  I was going to take a look, but maybe you'll get to it before I will.

panmari:
@chemelnucfin is that concerned to this PR?

chemelnucfin:
@panmari, I don't think so, but I noticed the mac test failing.

martinwicke:
The mac/python3 tests had been failing already, so I merged anyway.

PR #1404, opened 2016-03-06
DO NOT MERGE: Replaced python configuration by a Skylark Remote Repository

damienmg:
This is just an example on how to delete the ./configure script
with Skylark Remote Repository (available in Bazel @HEAD).

This was tested by building the pip package on OS X but should
be tested further on various platform.

Protobuf fast python header is deactivated because protobuf
needs to be updated to work correctly with that.

It should be pretty straightforward to do the same things for
cuda too.


vrv:
closing for now, this is a useful reference for the future when we decide to use this mechanism for configuring.

davidzchen:
For my reference, here is the tracking bug for CUDA autoconf: #2873

PR #1435, opened 2016-03-08
Fixing Broken Link

johcarvajal:
Fixed broked links


johcarvajal:
I fixed C++ documentation links

girving:
@martinwicke: Are these files autogenerated?

chemelnucfin:
@girving @martinwicke I think those files are autogenerated and we both made changes which should have fixed the links.  Do you need to run something internally and push to production?  This also references the comment @girving made on #1303

martinwicke:
I need to redeploy the website. Otherwise this should be fixed.

PR #1447, opened 2016-03-10
ci_build: fixes

jendap:
- do not require $USER variable
- do not contact github in print_build_info.sh
  - there is a easier and faster way to get the clone url


martinwicke:
Merged.

PR #1455, opened 2016-03-10
Fixed broken link in "Links"

cachanzheng:
In "Links" paragraph there is a broken link to tensorflow.org. 
Fixed it by changing from tensorflow.org to 
https://www.tensorflow.org/


girving:
@danmane: Looks reasonable to me, is there any website subtlety here?

PR #1456, opened 2016-03-10
Fixing broken link in "Using the Op in Python"

cachanzheng:
In "Using the Op in Python" section: the load_op_library link is broken.  Changed from ../../api_docs/python/framework#load_op_library to https://www.tensorflow.org/versions/r0.7/api_docs/python/framework.html#load_op_library


girving:
The broken link is unfortunate, but this PR is bad because it would make all versions of the docs link to the r.07 version.  Cc @martinwicke.

martinwicke:
Yes, the absolute path is bad, can you just add a .md after framework? That should fix it as well.

vrv:
(feel free to comment if you've updated this as above, otherwise closing due to inactivity)

PR #1460, opened 2016-03-11
Update README.md

shekkizh:
Scores obtained on running label_image does not display normalized value.

### DIFF
--- a/tensorflow/examples/label_image/README.md
+++ b/tensorflow/examples/label_image/README.md
@@ -43,15 +43,15 @@ This uses the default example image that ships with the framework, and should
 output something similar to this:
 
 ```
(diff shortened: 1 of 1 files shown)


shekkizh:
I signed it!

girving:
@petewarden: These scores do look pretty different.  Is there a potential issue here?

petewarden:
Those new scores look correct. The difference comes because I was originally using Inception v1 (that we include with the Android example) to generate the scores, and we switched to v3, but I forgot to update this README, though the tutorial has the correct values:
https://www.tensorflow.org/versions/master/tutorials/image_recognition/index.html

petewarden:
Thanks for this change @shekkizh! We appreciate the fix.

PR #1462, opened 2016-03-11
added shortcuts to access tensorshape dims and ndims

chemelnucfin:
for less typing.  
**len** seems to cause problems though, will look into it.

### DIFF
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -227,6 +227,11 @@ class Tensor(object):
     return self._op
 
   @property
+  def dims(self):
+    """Shortcut to return dims of the `TensorShape` of this Tensor."""
+    return self._shape.dims
+
+  @property
   def dtype(self):
     """The `DType` of elements in this tensor."""
     return self._dtype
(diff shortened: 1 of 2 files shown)


girving:
@mrry: Was there some plan to change `get_shape()` to be less verbose as well?

mrry:
Yes, but it depends on fixing all of the legacy code that uses the vestigial `Tensor.shape` property internally. Assuming this eventually happens, I'd be inclined to add `Tensor.shape` as an alias for `Tensor.get_shape()`, instead of accepting this PR.

girving:
@mrry: What does the vestigial thing do?  I can't find it in the code.

chemelnucfin:
@mrry Are there any reasons to have convenience accessors on C++ but not python?

girving:
It does seem like once we rename `get_shape()` to `shape`, these aliases fit naturally into the picture.

vrv:
Ok, looks like we have some internal work to do instead.  Thanks!

vrv:
(We can revive this later when it becomes possible to get to our end goal)

PR #1472, opened 2016-03-12
pip test-on-install changes for skflow

caisq:
1) Expanding the search scope for Python unit tests in order to include
the new tests in python/skflow/tests
2) By default, installing scipy and sklearn in virtualenv for the
aforementioned Python unit tests
3) Enabling caching of virtualenv directory to speed up pip test-on-install

### DIFF
--- a/tensorflow/tools/ci_build/builds/pip.sh
+++ b/tensorflow/tools/ci_build/builds/pip.sh
@@ -22,10 +22,19 @@
 #   pip.sh CONTAINER_TYPE [--test_tutorials]
 #
 # When executing the Python unit tests, the script obeys the shell
-# variables: TF_BUILD_BAZEL_CLEAN, NO_TEST_ON_INSTALL
+# variables: TF_BUILD_BAZEL_CLEAN, TF_BUILD_INSTALL_EXTRA_PIP_PACKAGES,
+# TF_BUILD_NO_CACHING_VIRTUALENV, NO_TEST_ON_INSTALL
 #
 # TF_BUILD_BAZEL_CLEAN, if set to any non-empty and non-0 value, directs the
 # script to perform bazel clean prior to main build and test steps.
+#
+# TF_BUILD_INSTALL_EXTRA_PIP_PACKAGES overrides the default extra pip packages
+# to be installed in virtualenv before test_installation.sh is called. Multiple
+# pakcage names are separated with spaces.
+#
+# TF_BUILD_NO_CACHING_VIRTUALENV: If set to any non-empty and non-0 value,
+# will cause the script to force remove any existing (cached) virtualenv
+# directory.
 #
 # If NO_TEST_ON_INSTALL has any non-empty and non-0 value, the test-on-install
 # part will be skipped.
@@ -34,6 +43,8 @@
 # tutorial tests (see test_tutorials.sh) after the PIP
 # installation and the Python unit tests-on-install step.
 #
+
+INSTALL_EXTRA_PIP_PACKAGES=${TF_BUILD_INSTALL_EXTRA_PIP_PACKAGES}
 
 # Helper functions
 # Get the absolute path from a path
@@ -111,7 +122,7 @@ PIP_WHL_DIR="${PIP_TEST_ROOT}/whl"
 PIP_WHL_DIR=$(abs_path ${PIP_WHL_DIR})  # Get absolute path
 rm -rf ${PIP_WHL_DIR} && mkdir -p ${PIP_WHL_DIR}
 bazel-bin/tensorflow/tools/pip_package/build_pip_package ${PIP_WHL_DIR} || \
-die "build_pip_package FAILED"
+    die "build_pip_package FAILED"
 
 # Perform installation
 WHL_PATH=$(ls ${PIP_WHL_DIR}/tensorflow*.whl)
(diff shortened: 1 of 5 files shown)


tensorflow/tools/ci_build/install/install_deb_packages.sh
```diff
@@ -23,6 +23,10 @@ apt-get install -y \
     build-essential \
     curl \
     git \
+    gfortran \
```
caisq: These four new packages are required by scipy.

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -111,7 +120,7 @@ PIP_WHL_DIR="${PIP_TEST_ROOT}/whl"
 PIP_WHL_DIR=$(abs_path ${PIP_WHL_DIR})  # Get absolute path
 rm -rf ${PIP_WHL_DIR} && mkdir -p ${PIP_WHL_DIR}
 bazel-bin/tensorflow/tools/pip_package/build_pip_package ${PIP_WHL_DIR} || \
-die "build_pip_package FAILED"
+    die "build_pip_package FAILED"
```
jendap: nice

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -127,25 +136,43 @@ echo "Installing pip whl file: ${WHL_PATH}"
 
 # Create temporary directory for install test
 VENV_DIR="${PIP_TEST_ROOT}/venv"
-rm -rf "${VENV_DIR}" && mkdir -p "${VENV_DIR}"
-echo "Create directory for virtualenv: ${VENV_DIR}"
+if [[ -d "${VENV_DIR}" ]]; then
+  echo "virtualenv directory already exists: ${VENV_DIR}"
+else
+  echo "Creating virtualenv directory: ${VENV_DIR}"
+  mkdir -p ${VENV_DIR} || \
```
jendap: You can do this, but is not overly verbose? I mean "mkdir -p ..." does not fail even if the directory is there already. Then all the if statements and echo statements are just noise.

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -127,25 +136,43 @@ echo "Installing pip whl file: ${WHL_PATH}"
...
-die "FAILED: Unable to activate virtualenv"
 
 # Install the pip file in virtual env
 pip install -v ${WHL_PATH} \
 && echo "Successfully installed pip package ${WHL_PATH}" \
 || die "pip install (without --upgrade) FAILED"
 
+# Install extra pip packages required by the test-on-install
+if [[ ! -z "${TF_BUILD_INSTALL_EXTRA_PIP_PACKAGES}" ]]; then
+  INSTALL_EXTRA_PIP_PACKAGES="${TF_BUILD_INSTALL_EXTRA_PIP_PACKAGES}"
```
jendap: Why not to do this as the other things... at top use TF_BUILD_INSTALL_EXTRA_PIP_PACKAGES=${TF_BUILD_INSTALL_EXTRA_PIP_PACKAGES:-scipy sklearn}"

It makes it clear at top what arguments it takes, what are defaults and makes the code easier to follow if things are immutable (in any language).

tensorflow/tools/ci_build/install/install_deb_packages.sh
jendap: and why don't we install python-scipy instead? Do we need newer version?

caisq: We need scipy inside virtualenv, for the install+test. This file is for outside virtualenv. Without these new packages, pip install scipy and pip install sklearn will fail in the virutalenv.

tensorflow/tools/ci_build/builds/pip.sh
caisq: Changed as suggested.

caisq:
@jendap's comments and questions addressed. Rebased and squashed. Ready to merge.

caisq:
I added an additional environment variable switch "TF_BUILD_NO_CACHING_VIRTUALENV" to allow removal of cached virtualenv folders if necessary.

tensorflow/tools/ci_build/install/install_deb_packages.sh
jendap: Take a look at https://github.com/tensorflow/tensorflow/pull/1479 please. That's what I was suggesting before. Is it not better?

Should we merge that first? And use here the python-scipy and no gfortran or dev libraries?

shanqing-cai: I think using --system-site-packages is a good idea. I will incorporate it into this PR and make other related, necessary changes in Dockerfiles to make it work, if that's okay with you.

jendap:

tensorflow/tools/ci_build/install/install_pip_packages.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+set -e
+
+pip install sklearn
```
jendap: Hmm, sklearn in available as python-sklearn package as well. Unfortunately python3-sklearn is not (on 14.04). So ok, let's keep them both here.

tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -125,27 +136,46 @@ echo "whl file path = ${WHL_PATH}"
...
-echo "Create directory for virtualenv: ${VENV_DIR}"
+if [[ -d "${VENV_DIR}" ]] &&
+   [[ ! -z "${TF_BUILD_NO_CACHING_VIRTUALENV}" ]] &&
+   [[ "${TF_BUILD_NO_CACHING_VIRTUALENV}" != "0" ]]; then
+  echo "TF_BUILD_NO_CACHING_VIRTUALENV=${TF_BUILD_NO_CACHING_VIRTUALENV}:"
+  echo "Removing existing virtualenv directory: ${VENV_DIR}"
+
+  rm -rf "${VENV_DIR}" || \
+      die "Failed to remove existing virtualenv directory: ${VENV_DIR}"
+fi
```
jendap: Ok, this style is here on multiple places. But for next time... why not simple remove all these features. The scripts can be considerably simpler by removing all this. And one still have it all. Look...

```bash
export TF_BUILD_NO_CACHING_VIRTUALENV=1
tensorflow/tools/ci_build/builds/pip.sh CPU
```

vs

```bash
rm -rf pip_test/venv
tensorflow/tools/ci_build/builds/pip.sh CPU
```

It is shorter even outside of he script and you have one less api feature (TF_BUILD_NO_CACHING_VIRTUALENV) to support. 10+ lines of script less.

PR #1479, opened 2016-03-13
virtualenv tests should inherit system site-packages

jendap:
This way we do not need compile and install anything. We will test with the same version of packages we have compiled with (and the same as bazel test is running). Also we do not need install gcc and compile pip packages.


caisq:
@jendap I incorporated this change into 1472, along with some related changes. So this PR can be closed if it's okay with you.

jendap:
Closing this as it is incorporated into https://github.com/tensorflow/tensorflow/pull/1472

PR #1485, opened 2016-03-13
Fix the download of the notMNIST dataset.

abhinav-upadhyay:
It seems the Apache server at http://yaroslavvb.com is configured to
allow access to only specific user agents. By passing a user agent
string like wget or Mozilla, it could be tricked into allowing the
notebook to download the dataset. It is an ugly hack for urllib, would
have been simpler if the notebook used the requests module :-)


vincentvanhoucke:
We're moving the data to a better host, stay tuned.

jendap:
It sounds like somebody enabled the filtering for a reason. And downloads from tensorflow could be that reason.

:+1: for a different host

vincentvanhoucke:
The new URL:
`url = 'http://commondatastorage.googleapis.com/books1000/'`
Let me know if that one is still an issue. I'll update the container soon.

abhinav-upadhyay:
The new URL works. I will close the PR. Thanks :)

PR #1487, opened 2016-03-13
make xrange python 2 and 3 compatible in sdca_ops.py

chemelnucfin:
use six.moves.range instead of xrange for python2/3 compatibility.

By the way, are there ways to use jenkins remotely to test on personal forks?  It takes a long time to run the entire test suite and renders my computer almost unusable for other tasks.


vrv:
@tensorflow-jenkins: test this please.

chemelnucfin:
@vrv yup, those are the exact same tests on python3 that failed on mine too.  I just fixed an easy one.

vrv:
sounds good, thanks!

as for remote testing -- unfortunately only admins can trigger tests right now.

PR #1489, opened 2016-03-14
Fixing mac-pip builds

caisq:
The --system-site-package option flag in PR 1472 broke Mac pip builds due to existing packages on the system. For example, see: http://ci.tensorflow.org/view/Experimental/job/experimental-cais-tensorflow-mac-python2-copt_both/17/console

Generally speaking, the --system-site-package option makes the virtualenv environment depend on the packages on the Mac host. Removing the option gets rid of this dependency and the need to maintain the host packages, which will save work (especially if there are multiple Mac slaves). 

The cost of installing the packages in the virtualenv should be minimal for both Mac and Linux, now that we cache the virtualenv folder by default. 

The PR has been tested manually for both python2 and python3 on Mac. See:
http://ci.tensorflow.org/view/Experimental/job/experimental-cais-tensorflow-mac-python2-copt_both/18/console
http://ci.tensorflow.org/view/Experimental/job/experimental-cais-tensorflow-mac-python3-copt_both/4/console

### DIFF
--- a/tensorflow/tools/ci_build/builds/pip.sh
+++ b/tensorflow/tools/ci_build/builds/pip.sh
@@ -163,10 +163,10 @@ source "${VENV_DIR}/bin/activate" || \
     die "FAILED: Unable to activate virtualenv"
 
 
-# Install the pip file in virtual env
-pip install -v --force-reinstall ${WHL_PATH} \
-&& echo "Successfully installed pip package ${WHL_PATH}" \
-|| die "pip install (without --upgrade) FAILED"
+# Install the pip file in virtualenv
+pip install -v --upgrade --force-reinstall ${WHL_PATH} || \
+    die "pip install FAILED"
+echo "Successfully installed pip package ${WHL_PATH}"
 
 # Install extra pip packages required by the test-on-install
 for PACKAGE in ${INSTALL_EXTRA_PIP_PACKAGES}; do


jendap:
And how is "blaze test" without pip.sh going to work?

caisq:
@jendap For Mac, "blaze test" should work as long as the correct versions of dependent python packages (numpy, scipy and sklearn) are installed on he system. I think the main problem with "--system-site-packages" is that if TensorFlow is already installed on the system, "pip install" in virtualenv from pip.sh will not update it, if that the version number has not increased. Even the "--force-reinstall" option does not work.

jendap:
It would be great to have consistency with what bazel see and what virtualenv see.

Looking at the output from your build 17 - it says "Requirement already satisfied (use --upgrade to upgrade): tensorflow==0.7.1". Have you tried adding "--upgrade" to pip install?

Ideally it should upgrade tensorflow but not the other packages.

jendap:
This is what we need: `pip install --upgrade --no-deps --force-reinstall <packagename>`

(second answer on http://stackoverflow.com/questions/19548957/can-i-force-pip-to-reinstall-the-current-version)

jendap:
Take a look at https://github.com/tensorflow/tensorflow/pull/1495. It does exactly what we need.

caisq:
Added the --upgrade suggested in #1495. Ready to merge.

caisq:

martinwicke:
I merged #1495 so I'm closing this.

PR #1493, opened 2016-03-14
fixes #1423 - add check for tests to see if tensorflow was built with gpu

### LINKED ISSUE #1423: Need force_gpu_if_available for tests
For GPU tests, we have the `force_gpu` and `use_gpu` flags. `force_gpu=True` is good for manual testing during development, but if an op cannot be placed on a gpu because there isn't a GPU or because TensorFlow was built without GPU support, it will make the test fail. 

For tests, we really need a way to say "fail if a GPU is available, but it's not used". Could be as simple as providing the equivalent of `force_gpu=tf.compiled_with_GPU_support() and tf.get_GPU_devices()`. Both functions don't exist, but really should.

chemelnucfin:
EDIT:
The changes worked on CPU, but I would like to test GPU just to make sure.  I opened an issue for GPU test failure #1501

### DIFF
--- a/tensorflow/python/framework/test_util.py
+++ b/tensorflow/python/framework/test_util.py
@@ -222,7 +222,7 @@ class TensorFlowTestCase(googletest.TestCase):
       config: An optional config_pb2.ConfigProto to use to configure the
         session.
       use_gpu: If True, attempt to run as many ops as possible on GPU.
-      force_gpu: If True, pin all ops to `/gpu:0`.
+      force_gpu: If True and built with GPU, pin all ops to `/gpu:0`.
 
     Returns:
       A Session object that should be used as a context manager to surround
@@ -246,7 +246,7 @@ class TensorFlowTestCase(googletest.TestCase):
                                                config=prepare_config(config))
       sess = self._cached_session
       with sess.graph.as_default(), sess.as_default():
-        if force_gpu:
+        if force_gpu and test.is_built_with_gpu():
           with sess.graph.device("/gpu:0"):
             yield sess
         elif use_gpu:
@@ -256,7 +256,7 @@ class TensorFlowTestCase(googletest.TestCase):
             yield sess
     else:
       with session.Session(graph=graph, config=prepare_config(config)) as sess:
-        if force_gpu:
+        if force_gpu and test.is_built_with_gpu():
           with sess.graph.device("/gpu:0"):
             yield sess
         elif use_gpu:
--- a/tensorflow/python/framework/test_util_test.py
+++ b/tensorflow/python/framework/test_util_test.py
@@ -23,6 +23,7 @@ import threading
 import numpy as np
 from six.moves import xrange  # pylint: disable=redefined-builtin
 import tensorflow as tf
+import warnings
 
 from google.protobuf import text_format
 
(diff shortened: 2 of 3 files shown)


tensorflow/python/framework/test_util.py
```diff
@@ -27,6 +27,7 @@
 
 import numpy as np
 import six
+import tensorflow as tf
```
vrv: I don't think this will work -- this will create a circular dependency since tf needs test_util.  You'll have to get the internal test_util module using the fully qualified path.

martinwicke:
Yes, you cannot import tensorflow from anywhere inside tensorflow, except for inside tests.

chemelnucfin:
I see.  I couldn't import test before, but I guess "test" was already imported.  I've finished a fix, just running it through my CPU tests before uploading it (it takes about an hour now after rebase) :( .  I still have not figured out how to the GPU tests yet though.

martinwicke:
Jenkins, test this please.

chemelnucfin:
@martinwicke Hello, I saw that the GPU test was failing.  I apologize, I couldn't build with GPU to test on my machine.  It's a little strange that the CPU test can import test but not the GPU test.

I'll try to resolve the testing locally, but I reinstalled everything from the beginning trying to fix it and followed the instructions.  Currently I can't even test the CPU, but I did built the wheel to install locally.  I can't build the GPU though.

chemelnucfin:
@martinwicke let me try setting up the virtualenv

martinwicke:
Any luck?

gunan:

chemelnucfin:
@martinwicke Sorry, nope.  Sometimes it's with protobuf, sometimes it's with cuda, sometimes it's with bazel, sometimes it's with blaze.  Kinda frustrating.  How do you guys deal with build config issues internally?

martinwicke:
Hm... I usually use virtualenv which makes this pretty straightforward. For maximum isolation from problems, you can try the ci_build scripts. They set up a docker container and install all the prerequisites inside. Then you can work inside there.

You do need the latest bazel, and the key is to have no incompatible protobuf and/or tensorflow installed. That's surpsingly hard to enforce, but virtualenv takes away most of the pain according to my experience.

martinwicke:
Do you want to update this PR? It seems it's a simple python syntax thing in test_util_test.

chemelnucfin:
@martinwicke I haven't figured out how to get tensorflow from source (especially with GPU).   For one thing, there seems to be a bazel issue upstream. https://github.com/bazelbuild/bazel/issues/1193

There was one issue that was fixed with the GPU using --genrule=standalone from #2053 though.

I looked over the tests, it is a bit strange that the CPU tests pass, but not the GPU tests.  Perhaps you might know the reason for that and/or can help me fix it and test it quickly?  I feel bad holding you guys up on this too.

martinwicke:
@caisq: This fails only GPU_PIP, because tf.test is not available there, is that expected? It's possible we simply have to blacklist this test, but I may be missing something.

martinwicke:
Ping @caisq

vrv:
ping @caisq @martinwicke

chemelnucfin:
@vjv @martinwicke I did get testing working on gpu also last weekend.
However I mostly can only work on tensorflow on weekends now.

martinwicke:
@caisq should we blacklist this test in the PIP build? Or should it be able to run?

vrv:
Closing due to inactivity, we can re-open if there's any more traction or debugging on this.  I suspect our testing team will at some point try to address these issues for load reasons.

PR #1495, opened 2016-03-14
fix virtualenv pip testing

jendap:


tensorflow/tools/ci_build/builds/pip.sh
```diff
@@ -164,9 +164,9 @@ source "${VENV_DIR}/bin/activate" || \
 
 
 # Install the pip file in virtual env
-pip install -v --force-reinstall ${WHL_PATH} \
-&& echo "Successfully installed pip package ${WHL_PATH}" \
-|| die "pip install (without --upgrade) FAILED"
+pip install -v --upgrade --no-deps --force-reinstall ${WHL_PATH} || \
+    die "pip install (without --upgrade) FAILED"
```
caisq: Thanks for the suggestion for "--upgrade". But I think "--no-deps" needs to be removed, otherwise we'll run into issues related to protobuf being unavailable.  
Also, the substring "(without --upgrade )" needs to be removed. 
I'll fix these in my PR 1489 and test it thoroughly.

jendap: right, the comment should change :)

I'm afraid the upgrade and no-deps has to stay both otherwise we would upgrade all the dependencies. And we do not want that.

I'm going to update this PR a bit...

jendap:
@caisq how about this? IT is similar to what you had there some time ago. It should work.

BTW: It may work completely without the second pip install if I do another thing I wanted to do ... one day ... for now this should work, right?

PR #1496, opened 2016-03-14
fix crash for empty dev_set bucket

yuwen-yan:


vrv:

PR #1507, opened 2016-03-15
quick python3 fix

chemelnucfin:
use as instead of comma for python3 compatibility

### DIFF
--- a/tensorflow/python/training/session_manager.py
+++ b/tensorflow/python/training/session_manager.py
@@ -326,7 +326,7 @@ class SessionManager(object):
       try:
         sess.run(self._ready_op)
         return None
-      except errors.FailedPreconditionError, e:
+      except errors.FailedPreconditionError as e:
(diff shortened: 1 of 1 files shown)


chemelnucfin:
apparently only python 2.6 and above though.
http://stackoverflow.com/questions/12682558/how-to-write-an-exception-catching-code-works-in-python2-4-to-python3

martinwicke:
Thanks! Just did the same without seeing yours. I'll merge yours (tests are in #1508). We'll need more fixed though.

chemelnucfin:
@martinwicke ya, I fixed one other simple one with the b missing in front of the string.  A lot of the others require diving into the C++ part and/or server/clients and haven't gone deep into them yet.
Thanks for merging mine and hopefully you don't mind I stepped on your path again.  :)

martinwicke:
No problem. Always nice to have backup. :)

PR #1508, opened 2016-03-15
Fix python3 breakage (old-style exception block)

martinwicke:


vrv:
#1507 does the same?

martinwicke:
Ah. I'll merge that.

martinwicke:
Merged #1507 instead.

PR #1512, opened 2016-03-15
some learning decays from Stanford CS231n Karpathy lecture 6

chemelnucfin:
exponential time decay
step decay
1/t decay
from Lecture 6-39 Stanford CS231n

### DIFF
--- a/tensorflow/python/training/initial_learning_rate_decay.py
+++ b/tensorflow/python/training/initial_learning_rate_decay.py
@@ -1,0 +1,190 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Various learning rate decay functions."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from tensorflow.python.framework import dtypes
+from tensorflow.python.framework import ops
+from tensorflow.python.ops import math_ops
+from tensorflow.python.ops import constant_op
+
+
+def exponential_time_decay(initial_lr, epoch, decay_rate, name=None):
+  """Applies exponential time decay to the initial learning rate.
+
+  When training a model, it is often recommended to lower the learning rate as
+  the training progresses.  This function applies an exponential decay function
+  to a provided initial learning rate.  It requires an `epoch` value to compute
+  the decayed learning rate.  You can just pass a TensorFlow variable that you
+  increment at each training step.
+
+  The function returns the decayed learning rate.  It is computed as:
+
+  ```python
+  decayed_learning_rate = initial_lr * exp(-decay_rate * epoch)
+  ```
+
+  Example: decay exponetially with a base of 0.96:
+
+  ```python
+  ...
+  epoch = tf.Variable(0, trainable=False)
+  initial_lr = 0.1
+  k = 0.5
+  learning_rate = tf.train.exponential_time_decay(initial_lr, epoch, k)
+
+  # Passing epoch to minimize() will increment it at each step.
+  learning_step = (
+      tf.GradientDescentOptimizer(learning_rate)
+      .minimize(...my loss..., global_step=epoch)
+  )
+  ```
+
+  Args:
+    initial_lr: A scalar `float32` or `float64` `Tensor` or a
+      Python number.  The initial learning rate.
+    epoch: A Python number.
+      Global step to use for the decay computation.  Must not be negative.
+    decay_rate: A Python number.  The decay rate.
+    name: String.  Optional name of the operation.  Defaults to
+      'ExponentialTimeDecay'
+
+  Returns:
+    A scalar `Tensor` of the same type as `learning_rate`.  The decayed
+    learning rate.
+  """
+  with ops.op_scope([initial_lr, epoch, decay_rate],
+                    name, "ExponentialTimeDecay") as name:
+    initial_lr = ops.convert_to_tensor(initial_lr, name="learning_rate")
+    decay_rate = math_ops.cast(decay_rate, initial_lr.dtype)
+    exponent = math_ops.exp(math_ops.mul(math_ops.neg(decay_rate), epoch))
+    return math_ops.mul(initial_lr, exponent, name=name)
+
+
+def inverse_time_decay(initial_lr, epoch, decay_rate, name=None):
+  """Applies inverse time decay to the initial learning rate.
+
+  When training a model, it is often recommended to lower the learning rate as
+  the training progresses.  This function applies an inverse decay function
+  to a provided initial learning rate.  It requires an `epoch` value to compute
+  the decayed learning rate.  You can just pass a TensorFlow variable that you
+  increment at each training step.
+
+  The function returns the decayed learning rate.  It is computed as:
+
+  ```python
+  decayed_learning_rate = learning_rate / (1 + decay_rate * t)
+  ```
+
+  Example: decay 1/t with a rate of 0.5:
+
+  ```python
+  ...
+  epoch = tf.Variable(0, trainable=False)
+  initial_lr = 0.1
+  k = 0.5
+  learning_rate = tf.train.inverse_time_decay(initial_lr, epoch, k)
+
+  # Passing epoch to minimize() will increment it at each step.
+  learning_step = (
+      tf.GradientDescentOptimizer(learning_rate)
+      .minimize(...my loss..., global_step=epoch)
+  )
+  ```
+
+  Args:
+    initial_lr: A scalar `float32` or `float64` `Tensor` or a
+      Python number.  The initial learning rate.
+    epoch: A Python number.
+      Global step to use for the decay computation.  Must not be negative.
+    decay_rate: A Python number.  The decay rate.
+    name: String.  Optional name of the operation.  Defaults to
+      'InverseTimeDecay'
+
+  Returns:
+    A scalar `Tensor` of the same type as `learning_rate`.  The decayed
+    learning rate.
+  """
+
+  with ops.op_scope([initial_lr, epoch, decay_rate],
+                    name, "InverseTimeDecay") as name:
+    initial_lr = ops.convert_to_tensor(initial_lr, name="learning_rate")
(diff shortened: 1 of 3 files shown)


tensorflow/python/training/learning_rate_decay.py
```diff
@@ -19,7 +19,8 @@
 from __future__ import print_function
 
 from tensorflow.python.framework import ops
-from tensorflow.python.ops import math_ops
+from tensorflow.python.ops.math_ops import add, cast, div, exp, floor, mod, mul, neg
```
vrv: We generally don't import this way: we import the module and fully qualify them.

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -77,10 +78,173 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
...
+    A scalar `Tensor` of the same type as `learning_rate`.  The decayed
+    learning rate.
+  """
+  with ops.op_scope([initial_lr, epoch, decay_rate],
+                   name, "ExponentialTimeDecay") as name:
+    initial_lr = ops.convert_to_tensor(initial_lr, name="learning_rate")
+    decay_rate = cast(decay_rate, initial_lr.dtype)
+    exponent = exp(mul(neg(decay_rate), epoch))
+    return mul(initial_lr, exponent, name=name)
+
```
vrv: We use two line spaces between top-level definitions

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -77,10 +78,173 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
...
+    decay_rate: A Python number.  The decay rate.
+    name: String.  Optional name of the operation.  Defaults to
+      'ExponentialTimeDecay'
+
+  Returns:
+    A scalar `Tensor` of the same type as `learning_rate`.  The decayed
+    learning rate.
+  """
+  with ops.op_scope([initial_lr, epoch, decay_rate],
+                   name, "ExponentialTimeDecay") as name:
```
vrv: indentation is misaligned

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -77,10 +78,173 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
...
+  Returns:
+    A scalar `Tensor` of the same type as `learning_rate`.  The decayed
+    learning rate.
+  """
+  with ops.op_scope([learning_rate, epoch, num_steps_for_decay, decay_rate],
+                   name, "StepTimeDecay") as name:
+    learning_rate = ops.convert_to_tensor(learning_rate, name="learning_rate")
+    dtype = learning_rate.dtype
+    decay_rate = cast(decay_rate, dtype)
+    if mod(epoch, num_steps_for_decay).eval():
```
vrv: Functions shouldn't require an eval in them to work -- can you express this symbolically?

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -77,10 +78,173 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
...
+               name=None):
+  """Applies step decay to the learning rate.
+
+  When training a model, it is often recommended to lower the learning rate as
+  the training progresses.  This function applies a step decay function
+  to a provided initial learning rate.  It requires a `global_step` value to
+  compute the decayed learning rate.  You can just pass a python number
+  that you increment at each training step.
+
+  The function returns the decayed learning rate.  It is computed as:
```
vrv: Can you describe in more detail what type of behavior you want this to have?   The function below suggests that the learning rate only changes on exactly one step every epoch, which I don't think is what you wanted.

tensorflow/python/training/learning_rate_decay.py
chemelnucfin: @vrv Thanks Vijay for your comments.  I see, I just thought it makes the equation easier to read.  ie. Line 137 exponent = exp(mul(neg(decay_rate), epoch)).  I'll change it and resubmit.

tensorflow/python/training/learning_rate_decay.py
chemelnucfin: Ok, I just thought if I was working with pure numbers, I don't need to go through the compute mechanism.  I'll change this too.

tensorflow/python/training/learning_rate_decay.py
chemelnucfin: My intention is to decrease the learning rate for every num_steps_for_decay.  So when the step(or epoch) number mod num_steps_for_decay has no remainder, then decay. ie.  0.96 for first 1000 steps, then 0.48 for the next 1000, then 0.24...  

Let me try to reword the function a little better.

chemelnucfin: Maybe I used the wrong word, I should have kept the name as "step".

tensorflow/python/training/learning_rate_decay_test.py
```diff
@@ -72,6 +73,36 @@ def testVariables(self):
...
+      
+class StepDecayTest(test_util.TensorFlowTestCase):
+  
+  def testDecay(self):
+    lr = 0.1
+    k = 0.5
+    num_steps = 5
+    with self.test_session():
+      for step in range(1, 11):
+        lr = learning_rate_decay.step_time_decay(lr, step, num_steps, k)
```
vrv: The reason I was concerned about this function is that you are assuming the 'lr' passed into the function is updated like this test is doing on every call -- the other functions are always passing an initial learning rate and computing the final learning rate based on the step and steps per epoch, so the result doesn't have to be passed back into a later call of the function.  I think keeping the same type of behavior for this function would be ideal.

chemelnucfin: Oh I see, let me see if I can make it better. I'll get back to you.

chemelnucfin:
@vrv I've addressed your comments and made the step time decay function take the initial learning rate as a parameter instead of passing the learning rate back in every time.  Additional comments welcome.  Thanks.

tensorflow/python/training/initial_learning_rate_decay.py
```diff
@@ -0,0 +1,189 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Various learning rate decay functions."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from tensorflow.python.framework import ops, dtypes
```
vrv: I think we prefer importing one module per line (like 22 and 23)

chemelnucfin: Fixed, thanks.

tensorflow/python/training/initial_learning_rate_decay.py
```diff
@@ -0,0 +1,190 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
```
vrv: is there any reason why you split this up into a separate file ?

tensorflow/python/training/learning_rate_decay_test.py
```diff
@@ -23,8 +23,11 @@
 from tensorflow.python.ops import state_ops
 from tensorflow.python.ops import variables
 from tensorflow.python.platform import googletest
+from tensorflow.python.training import initial_learning_rate_decay
 from tensorflow.python.training import learning_rate_decay
 
+import math
```
vrv: we typically import system / python libraries before tensorflow ones, so you can add this on line 20 like:

    import math

    from tensorflow.python. ....

tensorflow/python/training/initial_learning_rate_decay.py
chemelnucfin: I decided to do so because of the confusion you mentioned with using
initial learning rate versus feeding current learning rate into the
function.  I can change it back if you prefer.

vrv: I guess I don't see how exponential_decay is different: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/learning_rate_decay.py#L51 shows the same type of pattern -- let me know if I'm missing something subtle!

chemelnucfin: Hmm...  It is just an option to use exp as base rather than power.  But it
does make the naming a little confusing.  Do you have suggestions?  Thanks.

vrv: sorry, i meant that exponential_decay is also using the initial_lr pattern, so they might as well all be in the same file.

chemelnucfin: @vrv I see.  I somehow had the impression that the learning rate was continuously fed in and updated in the original file.  It seems like I remembered wrong.  Let me look at it again and I'll put it in the same file if I confirmed my thinking.  Thanks!

chemelnucfin: @vrv do you have any suggestions on differentiating the names between the existing exponential function and the new exponential function?

vrv: I'm not sure -- maybe natural_exponential_decay to more strongly indicate that it's related to exp ?

tensorflow/python/training/initial_learning_rate_decay.py
```diff
@@ -0,0 +1,190 @@
...
+
+  Returns:
+    A scalar `Tensor` of the same type as `learning_rate`.  The decayed
+    learning rate.
+  """
+  with ops.op_scope([initial_lr, epoch, decay_rate],
+                    name, "ExponentialTimeDecay") as name:
+    initial_lr = ops.convert_to_tensor(initial_lr, name="learning_rate")
+    decay_rate = math_ops.cast(decay_rate, initial_lr.dtype)
+    exponent = math_ops.exp(math_ops.mul(math_ops.neg(decay_rate), epoch))
```
vrv: after more closely looking at the differences, I realize that all of these functions have a 'staircase' effect: they only drop after each epoch.

I think it might be worth looking at how it is done in exponential_decay:

- You always pass the global_step and the 'decay steps' (essentially the number of steps in an epoch)
- If staircase is True, then you take the floor of global_step / decay_steps when computing the equivalent of 'epoch' here
- If staircase is False, then you use global_step / decay_steps as a floating point value, so you have smooth behavior.  That way all of the interfaces to the learning rate functions are basically identical, they each take an initial learning rate, an global step value, and a decay_steps. 

In fact, where possible, it might be preferable to keep the signature (the names of the variables) the same as in exponential_decay.  Does this seem reasonable?

chemelnucfin: Yes.  I was just on my way to naming them consistently.  Thanks for your
time in reviewing.

vrv:
It's been over a month since the last activity, so I'm closing this -- feel free to re-open when you have something to update!

chemelnucfin:
@vrv Seems like I got testing working again.  I have updated this pull request with the changes you suggested.  Please let me know if there are any problems.

PR #1525, opened 2016-03-16
fix python3 supervisor and server_lib tests

chemelnucfin:
as title.


martinwicke:
We've fixed those internally, they should come out shortly. Sorry we keep stepping on your feet. Everything will get better once we have the python3 lint running :)

PR #1531, opened 2016-03-17
Replace build_opener with urlretrieve

lekaha:
Since used 'urlretrieve' without header to download the file from give url will cause fail it should use another way it is able to attach headers instead.


lekaha:
I signed the CLA. But the signed name isn't the name I committed.

lekaha:
One of the datasets, notMNIST_large.tar.gz, is quite big. It will take time for downloading.

vincentvanhoucke:
Is this still an issue since we moved from Yaroslav's domain to:
url = 'http://commondatastorage.googleapis.com/books1000/'
?

lekaha:
@vincentvanhoucke
Sorry, I tested it on little bit older commit and pasted the fix to new repo that I just clone yesterday. I've tested the url = 'http://commondatastorage.googleapis.com/books1000/' , it's ok for me. 
At the point, I'd would like to discuss, it's a best way just change url? Even though it is just an assignment we're able to adopt the code and use other url instead(data set). It's just my opinion.

vincentvanhoucke:
If that's all right, I marginally prefer the simpler original code. It's more readable to novice coders. I might be proven wrong in case we hit other DOS protection issues, in which case we should definitely revive this patch. Thanks for your help!

PR #1536, opened 2016-03-18
test without long test_timeout

jendap:
We should remove --test_timeout=3600.

### DIFF
--- a/tensorflow/python/BUILD
+++ b/tensorflow/python/BUILD
@@ -1127,7 +1127,7 @@ cuda_py_tests(
         "kernel_tests/linalg_grad_test.py",
         "kernel_tests/pooling_ops_test.py",
     ],
-    shard_count = 2,
+    shard_count = 4,
 )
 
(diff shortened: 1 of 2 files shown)


jendap:
@martinwicke @vrv do not merge yet, I will remove the bazel clean hack. It will make sure the tests are not cached. I should have used --cache_test_results=no probably. Anyway, we will remove that hack.

jendap:
It has build just fine without insanely long test timeouts. We can merge this. But expect to find some flaky tests in next few builds.

vrv:

martinwicke:
cwise_ops_test is just on the boundary it seems -- runs 295-305s -- should we make that test bigger?

jendap:
How about just increase the sharding - https://github.com/tensorflow/tensorflow/pull/1536/commits/ef3409d0eb76e9f3ec0c181616ab91330f4805b1?

That sharding is actually very cool feature! Bazel is pretty smart :)

PR #1554, opened 2016-03-19
Clarify that skip-gram and CBOW are both Word2Vec models

lopespm:
Text changes made in the Udacity Deep Learning Assignment 5 to clarify that skip-gram and CBOW are both Word2Vec models

### DIFF
--- a/tensorflow/examples/udacity/5_word2vec.ipynb
+++ b/tensorflow/examples/udacity/5_word2vec.ipynb
@@ -283,7 +283,7 @@
       ],
       "execution_count": 0
     },
-    { skip-gram mod
+    {
       "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


tensorflow/examples/udacity/5_word2vec.ipynb
```diff
@@ -283,7 +283,7 @@
       ],
       "execution_count": 0
     },
-    {
+    { skip-gram mod
```
vincentvanhoucke: stray text here

lopespm: ok, fixed

vincentvanhoucke:
Merged. Thanks!

PR #1556, opened 2016-03-19
Fix broken link in add_summary() doc

davidkretch:
Fix incorrect closing paren in link to Session.run() in doc for
tf.train.SummaryWriter.add_summary()


shanqing-cai:
Please squash the commits to one and then we can test it. Thanks.

caisq:
Please squash the commits to one so we can test it. Thanks.

vrv:
@caisq: merge commits are ignored, so this is technically already squashed.

However, https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/summary_io.py#L119 is the right place to make this change, since our metadata documentation is generated from the original source.  Can you make that change @davidkretch ?

PR #1566, opened 2016-03-21
Update README.md for Udacity Deep Learning Course

waitingkuo:
Add a FAQ for adding a new virtual machine instead of the default one. 
The default one is only 1G. Here we create a new vm with 8G which is recommended


waitingkuo:
I signed it!

vincentvanhoucke:
Merged. Thanks!

PR #1580, opened 2016-03-22
Fix docs for cross-entropy loss functions

### LINKED ISSUE #1234: Update documentation for softmax-with-cross-entropy loss functions
`softmax_cross_entropy_with_logits` and `sparse_softmax_cross_entropy_with_logits` both have useful  behavior that conflicts with current documentation.

In `softmax_cross_entropy_with_logits`: "All that is required is that each row of `labels` is
a valid probability distribution." In `sparse_softmax_cross_entropy_with_logits`: "labels: Each entry `labels[i]` must be an index in `[0, num_classes)`."

Neither of these statements is true, and using all 0s in the case of `softmax_cross_entropy_with_logits` or -1s in the case of `sparse_softmax_cross_entropy_with_logits` is useful when signals include entries for which we don't want to compute loss.

`softmax_cross_entropy_with_logits` example:

```
logits_list = [tf.Variable([[0.0, 0.0, 0.0]]), tf.Variable([[99.0, 0.0, 0.0]])]
labels = tf.Variable([[0.0, 0.0, 0.0]])
loss_list = [tf.nn.softmax_cross_entropy_with_logits(logits, labels)
             for logits in logits_list]

with tf.Session() as sess:
    print(sess.run(loss_list))

# [array([ 0.], dtype=float32), array([ 0.], dtype=float32)]
```

`sparse_softmax_cross_entropy_with_logits` example:

```
logits_list = [tf.Variable([[0.0, 0.0, 0.0]]), tf.Variable([[99.0, 0.0, 0.0]])]
labels = tf.cast(tf.Variable([-1]), tf.int64)
loss_list = [tf.nn.sparse_softmax_cross_entropy_with_logits(logits, labels)
             for logits in logits_list]

with tf.Session() as sess:
    print(sess.run(loss_list))

# [array([ 0.], dtype=float32), array([ 0.], dtype=float32)]
```

We could filter invalid entries out before computing logits/loss, but in my use cases this would add a lot of boilerplate code and make only a negligible difference performance wise. (RNNs with long sequences, a fairly small number of time steps, multiple predictions per time step, and batch entries with varying `sequence_length`s. Using this trick lets us avoid having to break up logits/targets etc. into valid chunks vs. invalid chunks.

Edit: I ended up applying a boolean mask anyway since I wanted to view the loss over time (not just take gradients and minimize). In any case I like the current behavior so that I can apply masks after I obtain loss values over all targets, both valid and invalid.

rdipietro:
Should close https://github.com/tensorflow/tensorflow/issues/1234


josh11b:
Jenkins, test this please.

PR #1584, opened 2016-03-22
Note about compiling user op with gcc 5

keveman:
Added a note about compiling user op library with gcc 5


keveman:
I signed it!

PR #1591, opened 2016-03-23
fix test sizes

jendap:


jendap:
@girving we have set the timeouts way too strict... see http://ci.tensorflow.org/job/tensorflow-master-cpu/403/console

@girving @martinwicke @vrv why have we shrink down the timeouts? It will imho bring a lot of flakiness to the tests. Plus even if it runs well on our ci servers it may be always failing on contributor's notebook which has probably slower cpu. Why don't we have default "medium" or even "large"? 99% of tests will be fine and the 1% currently flaky will fail anyway. We should, imho, fix tests like session_test instead of chasing timeouts in BUILD files.

girving:
Jenkins, test this please.

@martinwicke: Preferences?  I'm not the right person to judge the usefulness, since I'm not familiar with the details of timeouts.

martinwicke:
I just want to avoid the endless build when we have an error that causes a
deadlock in session.run or some such. It happened before.

PR #1599, opened 2016-03-23
Add %matplotlib inline on the first cell

lekaha:
Config the matlotlib backend as plotting inline in IPython


vincentvanhoucke:
This is a huge diff: +782 −754. Can you make it such that the only change is that one line?

lekaha:
@vincentvanhoucke 
Updated.

vincentvanhoucke:
Merged. Thank you!

PR #1601, opened 2016-03-23
fix: add selective_registration.h header to framework file group

awni:


keveman:
@awni Thanks for the fix. As it happens, I have the same fix in #1608 that is farther along in the tests. Do you mind if I drop this one? #1608 has other commits and extricating changes to the one file would be painful.

awni:
@keveman No problem!

PR #1603, opened 2016-03-23
Benchmark tests

caisq:
To run the tests, do (for example):
export TF_BUILD_CONTAINER_TYPE=CPU
export TF_BUILD_PYTHON_VERSION=PYTHON2
export TF_BUILD_IS_OPT=OPT
export TF_BUILD_IS_PIP=NO_PIP
export TF_BUILD_RUN_BENCHMARKS=1
tensorflow/tools/ci_build/ci_parameterized_build.sh

Use the environment variable TF_BUILD_BENCHMARK_HOOK to specify the path
to a binary/script that handles the benchmark test log, to, e.g., enter
the benchmark information into a database.


tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -240,17 +247,22 @@ if [[ ${TF_BUILD_IS_PIP} == "pip" ]] ||
...
-  MAIN_CMD="${NO_PIP_MAIN_CMD}"
-elif [[ ${TF_BUILD_IS_PIP} == "pip" ]]; then
-  MAIN_CMD="${PIP_MAIN_CMD}"
-elif [[ ${TF_BUILD_IS_PIP} == "both" ]]; then
-  MAIN_CMD="${NO_PIP_MAIN_CMD} && ${PIP_MAIN_CMD}"
+
+if [[ ! -z "${TF_BUILD_RUN_BENCHMARKS}" ]] &&
+   [[ "${TF_BUILD_RUN_BENCHMARKS}" != "0" ]]; then
+  MAIN_CMD="${BENCHMARK_CMD} ${OPT_FLAG}"
 else
```
ebrevdo: Why not elif here?

tensorflow/tools/test/performance.bzl
```diff
@@ -0,0 +1,55 @@
+# -*- Python -*-
+
+load("//tensorflow:tensorflow.bzl", "tf_py_test")
```
ebrevdo: Add the tags "local", "regression-test" by default, and you'll have to modify jenkins testing to run:

bazel test $(bazel query "//tensorflow/... except attr(tags, 'local', //tensorflow/...)")

instead of what's currently there:

bazel test //tensorflow/...

tensorflow/tools/test/performance.bzl
```diff
@@ -0,0 +1,55 @@
...
+  if not target:
+    fail("Must provide a target")
+  if (not ":" in target
+      or not target.startswith("//")
+      or target.endswith(":all")
+      or target.endswith(".")):
+    fail(" ".join(("Target must be a single well-defined test, e.g.,",
+                   "//path/to:test. Received: %s" % target)))
+
+  all_tags = list(set(tags) + set(["benchmark-test"]))
```
ebrevdo: you'll be able to find benchmark tests via the query

bazel query "attr(tags, 'benchmark-test', //tensorflow/...)"

tensorflow/tools/test/run_and_gather_logs_lib.py
```diff
@@ -27,17 +27,50 @@
...
+  """Get git commit SHA for this build.
+
+  Attempt to get the SHA from environment variable GIT_COMMIT, which should
+  be available on Jenkins build agents.
+
+  Returns:
+    SHA hash of the git commit used for the build, if available
+  """
+
+  return os.getenv("GIT_COMMIT")
```
ebrevdo: will this return None if not found?  we want it to fail cleanly.

tensorflow/tools/test/run_and_gather_logs.py
```diff
@@ -40,10 +40,11 @@
 # pylint: enable=g-bad-import-order
 # pylint: enable=unused-import
 
-from google.protobuf import text_format
 from tensorflow.core.util import test_log_pb2
 from tensorflow.tools.test import run_and_gather_logs_lib
 
+from google.protobuf import text_format
```
ebrevdo: why the move?

caisq: Suggested by internal linter.

ebrevdo: That's weird, don't do that.  google comes before tensorflow
lexicographically.

tensorflow/tools/ci_build/ci_parameterized_build.sh
caisq: Revised as suggested.

tensorflow/tools/test/performance.bzl
caisq: Not sure I understand this comment and the previous one. This is already what is done in the newly added file tools/ci_build/builds/benchmark.sh.

tensorflow/tools/test/run_and_gather_logs_lib.py
caisq: Yes, this returns None if not found.

tensorflow/tools/ci_build/builds/benchmark.sh
```diff
@@ -0,0 +1,135 @@
...
+echo ""
+echo "====== Benchmark tests start ======"
+
+OPT_FLAG=""
+if [[ "$1" == "-c" ]] && [[ "$2" == "opt" ]]; then
+  OPT_FLAG="${OPT_FLAG} -c opt"
+fi
+
+BENCHMARK_TESTS=$(bazel query \
+    'attr("tags", "benchmark-test", //tensorflow/tools/test/...)')
```
ebrevdo: search in //tensorflow/...

tensorflow/tools/ci_build/builds/benchmark.sh
```diff
@@ -0,0 +1,135 @@
...
+FAILED_TESTS=""
+COUNTER=0
+
+for BENCHMARK_TEST in ${BENCHMARK_TESTS}; do
+  ((COUNTER++))
+
+  echo ""
+  echo "Running benchmark test (${COUNTER} / ${N_TESTS}): ${BENCHMARK_TEST}"
+
+  bazel test ${OPT_FLAG} --cache_test_results=no "${BENCHMARK_TEST}"
```
ebrevdo: Do you need to set up bazel in any special way here?  Or do you expect there to be a properly configured ~/.bazelrc?

tensorflow/tools/ci_build/builds/benchmark.sh
```diff
@@ -0,0 +1,135 @@
...
+#
+# Usage: benchmark [-c opt]
+# Option flags
+#    -c opt:  Use optimized C++ build ("-c opt")
+#
+# This script obeys the following environmental variables:
+#   TF_BUILD_BENCHMARK_HOOK:
+#     Path to a binary / script that will handle the test log and other related
+#     info after the completion of each benchmark test.
+
```
ebrevdo: set -e
set -u

tensorflow/tools/ci_build/builds/benchmark.sh
```diff
@@ -0,0 +1,135 @@
...
+echo ""
+echo "====== Benchmark tests start ======"
+
+OPT_FLAG=""
+if [[ "$1" == "-c" ]] && [[ "$2" == "opt" ]]; then
```
ebrevdo: consider using bash's getopts in the future.

tensorflow/tools/ci_build/builds/benchmark.sh
```diff
@@ -0,0 +1,135 @@
...
+
+      ${TF_BUILD_BENCHMARK_LOG_HOOK} "${OPT_FLAG}" "${TEST_LOG}"
+    else
+      echo "WARNING: No hook binary is specified to handle test log ${TEST_LOG}"
+    fi
+  else
+    # Mark as failure if the test log file cannot be found
+    TEST_RESULT=2
+
+    echo "ERROR: Cannot find log file from benchmark ${BENCHMARK_TEST} @ "\
```
ebrevdo: This should probably exit hard?  also echo >&2 ?

tensorflow/tools/test/performance.bzl
ebrevdo: I see it.

You'll have to update the regular tests to except local tag as above.

caisq: Added "local" and "regression-test" tags as suggested. 

However, ci_paramerized_build.sh has its own test tag filtering mechanism, for other build configurations such as GPUs. It uses the --test_tag_filters flag of "bazel test". I will use --test_tag_filters=-benchmark-test to filter out the benchmark tests for non-benchmark jobs. Let me know it is okay.

tensorflow/tools/ci_build/builds/benchmark.sh
caisq: Revised as suggested.

tensorflow/tools/ci_build/builds/benchmark.sh
caisq: Do you mean set -e? If so, we've decided not use 'set -e', but check for errors explicitly. See: http://mywiki.wooledge.org/BashFAQ/105

tensorflow/tools/ci_build/builds/benchmark.sh
caisq: I think it's better to finish running all the test even if there are some test failures. This is how other tests (e.g., bazel tests, pip test-on-install) are done.

caisq: Added >&2 to "cat ${TEST_LOG}"

tensorflow/tools/ci_build/builds/benchmark.sh
caisq: Revised the script to use getopts.

tensorflow/tools/test/run_and_gather_logs.py
caisq: This change is necessary: tensorflow is considered a third-party library, which goes before google libraries.

tensorflow/tools/ci_build/builds/benchmark.sh
ebrevdo: Nope, "set -u" causes the script to fail if you try to access a variable that is undefined.  Different type of strictness.  You can then use "${VAR:blah}" for "blah if $VAR is undefined".

caisq: OK. Added it. Thanks.

caisq:

caisq:
@vrv Does this look good to you? I tested the changes internally (with proper sync-repo mappings) and the tests pass. If so, please merge it.

caisq:
For internal mapping see CL 117950736

PR #1612, opened 2016-03-23
Add matplotlib inline

lekaha:
@vincentvanhoucke 
Sorry for leak comma.


vincentvanhoucke:
I've already fixed it.

PR #1617, opened 2016-03-24
Typo in comment, random weights init

amchercashin:
According to the code and tensorflow reference standard deviation would be 0.1
Excuse me if something is wrong, have little to no experience in python and tensorflow


amchercashin:
Signed the CLA.

PR #1620, opened 2016-03-24
remove dropout in MNIST tutorial

fayeshine:
remove dropout in MNIST tutorial since it didn't improve the accuracy but make training slower,
what's more, it make the tutorial long and tedious. A simpler demo which can show same results will be cool.
After removing dropout, the accuracy remains around 99.2% and sometimes even higher (test 3 times, 0.9919, 0.9922, 0.9929).

### DIFF
--- a/tensorflow/g3doc/tutorials/mnist/pros/index.md
+++ b/tensorflow/g3doc/tutorials/mnist/pros/index.md
@@ -333,20 +333,6 @@ h_pool2_flat = tf.reshape(h_pool2, [-1, 7*7*64])
 h_fc1 = tf.nn.relu(tf.matmul(h_pool2_flat, W_fc1) + b_fc1)
 ```
 
-#### Dropout
-
-To reduce overfitting, we will apply dropout before the readout layer.
-We create a `placeholder` for the probability that a neuron's output is kept
-during dropout. This allows us to turn dropout on during training, and turn it
-off during testing.
-TensorFlow's `tf.nn.dropout` op automatically handles scaling neuron outputs in
-addition to masking them, so dropout just works without any additional scaling.
-
-```python
-keep_prob = tf.placeholder(tf.float32)
-h_fc1_drop = tf.nn.dropout(h_fc1, keep_prob)
-```
-
 ### Readout Layer
 
 Finally, we add a softmax layer, just like for the one layer softmax regression
@@ -356,7 +342,7 @@ above.
 W_fc2 = weight_variable([1024, 10])
 b_fc2 = bias_variable([10])
 
-y_conv=tf.nn.softmax(tf.matmul(h_fc1_drop, W_fc2) + b_fc2)
+y_conv=tf.nn.softmax(tf.matmul(h_fc1, W_fc2) + b_fc2)
 ```
 
 ### Train and Evaluate the Model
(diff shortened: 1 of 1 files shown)


vrv:
@vincentvanhoucke: feel free to re-assign to reassess

samjabrahams:
To throw in my two cents: I think that, while training materials shouldn't be overly tedious in terms of time-to-completion, in this case it's more important to demonstrate the usage of an important API component. Since dropout is such a common and important tool, showcasing how to use it properly with TensorFlow in the MNIST tutorial sets users up for more success (and less frustration) in the long term.

While it's nice that the model in this example appears to work well without dropout, the purpose of the tutorial is to demonstrate how to use the TensorFlow API, not create the fastest-to-train accurate model.

fayeshine:
@samjabrahams 
I agree with you, we should demo the dropout api but maybe in later chapters where dropout can truly work. However, if it cannot work in this demo, it does not seems to be convincing to let people use it. This is the very begining chapter so I think we can keep it as simple as we could.

samjabrahams:
@fayeshine - If this was the "MNIST for Beginners" tutorial, I'd agree with you, but seeing as this is tutorial is "for experts", we should assume that users will have some experience building neural networks in other frameworks or languages, and will likely have used dropout before. Dropout is such common step to take in these sorts of models that taking it out of what is supposed to be a "moderately sophisticated" model would be conspicuous. This would lead to users to go looking for dropout in the API and figure out how to implement it anyway.

Users get the "keep it simple" parts of the model from "MNIST for beginners" and the first half of "MNIST for experts". Not only does removing dropout from the more advanced model do a disservice to those trying to learn the API, but it also gives cause for data scientists to be skeptical of the library.

If we want to make TensorFlow seem more friendly in terms of training time, our efforts would be better spent making TensorFlow more efficient instead of trying to cut corners when teaching the library.

fayeshine:
@samjabrahams 
I understand what you mean, truly dropout is very important and common for expert.
It seems to be very necessary to teach users how to use dropout.
However, the main point is not to reduce the training time, but dropout seems didn't useful or work at this MNIST tutorial. It may not be a good example here and may mislead users to use dropout all time without thinking the proper situation to use dropout.
So I just wonder whether it is suitable to put a non-useful procedure here, or maybe put it in later chapter of tutorial where it can truly work and in this case we can also teach the intuitive why it work: dropout can reduce the overfitting problem in complicated models with many parameters. In such manner, people can understand why dropout work and where dropout can work. :smiley:

samjabrahams:
@fayeshine Absolutely- I think a separate tutorial with a targeted example showcasing when and how dropout is useful would be a great resource for those learning about neural network models. I think our disagreement about this particular section stems from our thoughts on what the primary purpose of this tutorial should be. Correct me if I'm wrong, but from what I gather your opinion is that the most important aspect of this tutorial is to teach users how to design and setup a model that A, works and B. is as efficient as possible. I completely respect this, and it is an important skill for anyone who's learning machine learning to get a grasp of. My opinion, in contrast, is that this tutorial's primary goal should be to teach the fundamental concepts and frequently-used components of the API that one would expect to see in a convolutional neural network. In this way, it can serve as an alternative way of learning the API- showing users how different components are used in practice.

Again- I think your suggestion is valid, but the commit as it stands just removes any mention of dropout anywhere in the tutorial. I don't like the idea of completely removing valuable educational material about the API for the sake of teaching users that dropout isn't always necessary.

Perhaps this change could be bundled with the aforementioned separate dropout tutorial :)

fayeshine:
@samjabrahams , Yes, you are right, I think I will submit a commit bundled with the separate dropout tutorial or add dropout usage and demo in later tutorial chapter where it can truly work and improve the accuracy.

Sohl-Dickstein:
@fayeshine I agree that it makes more sense to introduce dropout in a context where it actually helps performance. Good catch! Adding it to a later tutorial sounds like a good idea. Another (possibly simpler) approach would be to modify the model architecture in the MNIST for Experts tutorial in such a way that dropout actually provides a benefit.

I don't think we should devote an entire tutorial to dropout. The goal of the tutorials is to teach people how to use TensorFlow rather than how to do ML research. We definitely want our examples to be scientifically correct and relevant, but we want to focus on the code. In the case of using dropout the code consists of the one function tf.nn.dropout(...), so an entire tutorial would be overkill.

Feel free to bounce ideas or an outline of your planned changes off of me!

vrv:
Closed since you've opened up other PRS

Issue #1413, opened 2016-03-07
error when building from source


fayeshine:
bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package

Warning: ignoring LD_PRELOAD in environment.
ERROR: /home/wenjian/pkgs/tensorflow/tensorflow/core/distributed_runtime/rpc/BUILD:87:1: no such package '@grpc//': Error cloning repository: https://boringssl.googlesource.com/boringssl: cannot open git-upload-pack caused by https://boringssl.googlesource.com/boringssl: cannot open git-upload-pack caused by https://boringssl.googlesource.com/boringssl: cannot open git-upload-pack and referenced by '//tensorflow/core/distributed_runtime/rpc:grpc_channel'.
ERROR: Loading failed; build aborted.
INFO: Elapsed time: 342.460s

girving:
@mrry: Looks like a distributed-related issue.

mrry:
Are you able to run `git clone https://boringssl.googlesource.com/boringssl` locally?

vrv:
#1387 might be relevant...

Vesnica:
I can run git clone https://boringssl.googlesource.com/boringssl locally, but in bazel that didn't work.

Maybe bazel didn't honor the proxy env vars when download the deps of deps. (tensorflow -> grpc -> boringssl)

fayeshine:
@mrry @Vesnica I can only run it with proxy since I'm behind the firewall. However, when I run bazel with proxy, it seems cannot work

mrry:
Can you please try upgrading to the latest HEAD of Bazel? It looks like a fix was added for proxy support in https://github.com/bazelbuild/bazel/commit/d4c00d42b7ac3939f7041ba3bcb920088857adbd.

Vesnica:
Latest bazel didn't fix this problem:

```
ERROR: /home/vesnica/tensorflow/tensorflow/core/distributed_runtime/rpc/BUILD:158:1: no such package '@grpc//': Error cloning repository: https://boringssl.googlesource.com/boringssl: cannot open git-upload-pack caused by https://boringssl.googlesource.com/boringssl: cannot open git-upload-pack caused by https://boringssl.googlesource.com/boringssl: cannot open git-upload-pack and referenced by '//tensorflow/core/distributed_runtime/rpc:grpc_remote_master'.
ERROR: Loading failed; build aborted.
INFO: Elapsed time: 161.436s
```

[...]

PR #1626, opened 2016-03-24
solve the boringssl problem in grpc

fayeshine:
boringssl in grpc was previously held in googlesouce which block people under the wall,
upgrade to this commit version will change its source to https://github.com/google/boringssl.git
and this solve all the problems.
Detail can be found in https://github.com/tensorflow/tensorflow/issues/1413#issuecomment-200804155

### DIFF
--- a/tensorflow/workspace.bzl
+++ b/tensorflow/workspace.bzl
@@ -74,7 +74,7 @@ def tf_workspace(path_prefix = "", tf_repo_name = ""):
 
   native.git_repository(
     name = "grpc",
-    commit = "73979f4",
+    commit = "3d62fc6",
     init_submodules = True,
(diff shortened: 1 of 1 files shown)


fayeshine:
@mrry Could you verify this? I think you have the background on this problem.

girving:
@mrry is probably the right person for grpc questions.

vrv:

mrry:
Should be fine as long as (i) it builds, and (ii) the tests pass. I'm not aware of any breaking API changes in gRPC since we first pinned to a commit.

mrry:
The timeout look suspicious but is unrelated to this change (and will be fixed when one of @vrv's patches gets in). I'm going to merge this now.

PR #1627, opened 2016-03-24
Make configure hard code path to swig

### LINKED ISSUE #1329: Our shell wrapper around swig is a bug according to bazel
Bazel likes to strip things it doesn't like out of `PATH`, as discussed at https://github.com/bazelbuild/bazel/issues/957.  For TensorFlow, this is a problem since we have a shell wrapper around swig that just says `swig "$@"`.  In my case, `swig` is in `~/homebrew/bin`.

Since Bazel considers this a bug in TensorFlow rather than in Bazel, it may be more useful to have the discussion here than on the Bazel site.

girving:
I've set it up to just try "swig" if configure hasn't been run since
syncing to this change, so users who sync and have no swig issues won't
have to do anything new.

Fixes #1329.


ebrevdo:
Jenkins, test this please.

girving:
@vrv: Is the server_lib_test timeout resolved?  http://ci.tensorflow.org/job/tensorflow-pull-requests-cpu/459/console

vrv:

(not sure, I think it's being "fixed")

vrv:
Our triggering seems broken right now.

PR #1631, opened 2016-03-24
Initial iOS Support and Example

petewarden:
Compiling this will require the top-of-tree version of Bazel, until it makes it into the released version in about a week.

To compile the iOS example, run the following command:
`bazel build -c opt -s //tensorflow/examples/ios:ios-app-binary-cc --ios_sdk_version=9.2 --ios_cpu="arm64"`

You will need to put your mobile provisioning certificate in tools/objc/default_provisioning_profile.mobileprovision, and the Inception v3 model file and labels in a data directory inside tensorflow/examples/ios/data, e.g.
`ls tensorflow/examples/ios/data/`
`grace_hopper.jpg  imagenet_comp_graph_label_strings.txt tensorflow_inception_graph.pb`


tensorflow/examples/ios/BUILD
```diff
@@ -0,0 +1,68 @@
+# Description:
+#   Tensorflow camera demo app for Android.
```
Mistobaan: for iOS?

petewarden:
Jenkins, test this please.

martinwicke:
@petewarden The failures still show bazel 0.2, are the new docker files not pushed out yet? Or did something go wrong with the bazel update? Or is this just a too old test?

vrv:

petewarden:
This was a recent test, I need to debug what's going wrong on the Jenkins server. The install script does appear to be updated:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/install/install_bazel.sh

vrv:

martinwicke:
I just checked, and at least on the Mac, bazel 0.2.1 is definitely installed

jendap:
You have the same error in linux build with 0.2.1 as well as on mac where is 0.2.0.

@martinwicke you have checked only mac0-slave, right?

@caisq can you upgrade bazel on mac1-slave, please?

martinwicke:
Oh, right, I have only updated mac0-slave to 0.2.1.

petewarden:
I've installed the 0.2.1 binaries locally, and it seems like the fix we need didn't make it in, despite being in top-of-tree ahead of time! :( I'm following up with the Bazel team to find out what happened, and what we can do to fix this.

Mistobaan:
any updates on this patch?

petewarden:
We're currently blocked because the top-of-tree protobuf is broken, and we need that for iOS support. I'm working on fixing that right now.

dauba-dauba:
I've been attempting to use this patch locally to see if I could manually work around the issues in protobuf. I've run into issues with Xcode 7 where object files built using the bazel cc_library rule can't be linked into an iOS simulator application due to metadata that's now being written into the object files. Going back to the bazel iOS example (http://bazel.io/docs/tutorial/ios-app.html) and adding some C++ code, I found that I was only able to get it to build properly if the C++ was added using an objc_library rule, not a cc_library rule.

I note that this patch currently only defines a cc_library rule for the tensorflow code. Does it also need to define an objc_library rule for the benefit of the simulator?

martinwicke:
@petewarden is this PR still active or can it be closed?

PR #1638, opened 2016-03-25
mu to learning_rate

amchercashin:
New var `mu` defined as _Learning rate_:
`mu = 0.002`
 but later in the same code cell var `learning_rate` is used instead:
`update_weights = tf.train.GradientDescentOptimizer(learning_rate).minimize(loss)`
, which was defined in previous cell:

Suggest to rename because cell uses var which is not defined in it, but
defines var which is not used.

Suggest to rename `mu` to `learning_rate` (not `learning_rate` to `mu` later in
code) because this code cell is supposed to be like previous but with
comments. And in previous cell the name `learning_rate` was used.

Commit also change var name in comment to reflect changes in var
definition.


girving:
Looks like you left a couple of `mu` references in the file.  Can you get rid of those too?  Also, please rebase and squash to remove the merge commit from the pull request.

amchercashin:
@girving: Yes, i have left them in the last code cell. Thought that `mu `was written there intentionally to distinguish manual calculations of weights update. And later, at the bottom text cell there are references to `mu` too. And I understand that in text and formulas `mu` looks more compact then `learning_rate.`

If you think that it is better to replace `mu` with `learning_rate` there too, I will do it.

girving:
Yeah, we should be uniform throughout the page.  `learning_rate` is more verbose but obviously less confusing as a name.

amchercashin:
Okay, done.

girving:
Thanks!  Jenkins, test this please.

vrv:
I should have double checked the merge commit - it was bogus.  You can send this PR in again with the right content!

amchercashin:
I'm very sorry, somehow some waste lines leaked in the file while I was trying to update my brach from master. Made a new PR: #1706. Hope it will be ok now.

PR #1641, opened 2016-03-25
gradient calc description change it text cell

amchercashin:
I have found the previous explanation of manual gradient calculation a
little confusing because we compute a vector of gradients with the size
of number of neurons (with bias) and in the explanation it looks like we
compute a scalar.

Propose more detailed variant if community finds it's better. Excuse me
if  it is convenient to write x_i even for bias unit, just discard this
then.

Attach old and new variant of just last notebook text cell saved to pdf (cannot attach ipynb here). The differences in the description of second line of code.

[Old.pdf](https://github.com/tensorflow/tensorflow/files/189573/Old.pdf)
[New.pdf](https://github.com/tensorflow/tensorflow/files/189574/New.pdf)


girving:
So that I understand, how does the old description imply the gradient is a scalar?

keveman:
Jenkins, test this please

vrv:

amchercashin:
@girving: I'm sorry, not a scalar, but a vector of size of number of neurons (without bias unit), and in the case of one neuron - a scalar. It was says that partial derivatives of loss function are _SUM( y_hat - y ) * x_i_ . And then I look at loss function and see only one _x_: 1/2*SUM( w2*x + w1 - y)^2. So then the _SUM( y_hat - y ) * x_i_ becomes _SUM( y_hat - y ) * x_ which looks like one number for me. Then I look at the code and see that the result is two numbers as it should be. But there is no partial derivative with respect to w1 in description

I understand that maybe it is convenient to just omit this dedicated equation for the bias unit thinking of it like it is an _x_whichisalwaysone_ and thus _SUM( y_hat - y ) * x_i_ will return two values here (with bias). But for a newbie like me it was a little confusing at start so I humbly propose more detailed description. If you found it bad, pls just discard it.

girving:
The line `x_with_bias = np.array([(1., a) for a in x]).astype(np.float32)` makes the simpler formula correct. Your new description is inconsistent with the simplified formulas above.  If you want to clarify that `x` contains a 1 entry corresponding to the bias, just add a parenthetical comment like "(Note that the first entry of `x` is 1 corresponding to the bias)`.

amchercashin:
Yes,I see your logic now! But in this python code string the name of variable is `x_with_bias` which means "x with something else", or "not only x." And, according to that string, x does not contains a 1 entry, it's `x_with_bias`. It's all just a matter of taste of course and not that important :)

girving:
Very true!  Let's use `x_with_bias` in the formula; that even avoids the need for an extra comment.

PR #1646, opened 2016-03-25
Get tensorflow/core link down to 2.2G

girving:
Most tests now depend on only the kernels they desperately need.


girving:
Jenkins, test this please.

PR #1648, opened 2016-03-25
Fix wrong parenthesis.

dongjoon-hyun:


vrv:
Thanks for the fix!  Would it be possible to introduce a test that exercises this codepath, so we can prevent it from regressing?

vrv:

(well, add a test later if you can, this is better than nothing)

PR #1649, opened 2016-03-26
Fix building android_tensorflow_lib.

darrengarvey:
This commit fixes a build time failure of the Android TF library that is
missing some dependencies since they have been made more fine-grained.
For reference, the relevant errors (one per required dep) look something
like:

```
ERROR: tensorflow/tensorflow/core/BUILD:663:1: Couldn't build file tensorflow/core/_objs/android_tensorflow_lib/tensorflow/core/kernels/softsign_op.o: undeclared inclusion(s) in rule '//tensorflow/core:android_tensorflow_lib':
this rule is missing dependency declarations for the following files included by 'tensorflow/core/kernels/softsign_op.cc':

ERROR: tensorflow/tensorflow/core/BUILD:663:1: undeclared inclusion(s) in rule '//tensorflow/core:android_tensorflow_lib':
this rule is missing dependency declarations for the following files included by 'tensorflow/core/kernels/maxpooling_op.cc':

ERROR: tensorflow/tensorflow/core/BUILD:663:1: undeclared inclusion(s) in rule '//tensorflow/core:android_tensorflow_lib':
this rule is missing dependency declarations for the following files included by 'tensorflow/core/kernels/ctc_decoder_ops.cc':

ERROR: tensorflow/tensorflow/core/BUILD:663:1: undeclared inclusion(s) in rule '//tensorflow/core:android_tensorflow_lib':
this rule is missing dependency declarations for the following files included by 'tensorflow/core/kernels/avgpooling_op.cc':
```

### DIFF
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -666,6 +666,7 @@ cc_library(
         ":android_op_registrations_and_gradients",
         "//tensorflow/core/kernels:android_core_ops",
         "//tensorflow/core/kernels:android_extended_ops",
+        "//tensorflow/core/util/ctc:android_srcs",
     ],
     copts = select({
         ":android": ANDROID_TF_COPTS,
@@ -679,7 +680,10 @@ cc_library(
     deps = [
         "@re2//:re2",
         ":android_tensorflow_lib_lite",
+        ":core_cpu",
+        ":framework",
         ":protos_cc",
+        "//tensorflow/core/kernels:pooling_ops",
         "//third_party/eigen3",
     ],
 )


vrv:
@andrewharp since he's been dutifully maintaining this target

@damienmg because I don't understand why our Android test target is fine but according to Jenkins but users still run into this problem :)

darrengarvey:
I'd add a patch to make this show up in your default CI builds but I'm not sure what your build steps are. tools/ci_build doesn't seem to be the right place to be looking.

andrewharp:
@darrengarvey Hi, are you building tensorflow/core:tensorflow_android_lib directly (as opposed to //tensorflow/examples/android:tensorflow_demo)? If so, you need to tell Bazel to actually build for Android, e.g. with the flag --fat_apk_cpu=armeabi-v7a

@vrv There's a select in android_tensorflow_lib_lite that nullifies attempts to build for non-Android, but not for android_tensorflow_lib. This is probably what's causing the misleading errors people are seeing. Jenkins always builds in an Android configuration so it's not catching this. I'll send a CL to fix this.

andrewharp:
@darrengarvey You'll need a couple more flags in addition to fat_apk_cpu if you want to build the library by itself. Here's what worked for me:
`bazel build -c opt tensorflow/core:android_tensorflow_lib --crosstool_top=//external:android/crosstool --cpu=armeabi-v7a --host_crosstool_top=@bazel_tools//tools/cpp:toolchain`

andrewharp:
Note that if you are building this for an app you'll probably want to wrap the cc_library in a cc_binary (see tensorflow/examples/android/BUILD for an example) so that unused code can be stripped out. Otherwise you'll end up with over 200mb of unnecessary size. Just make sure you export the symbols you want to reference so they don't get stripped as well.

darrengarvey:
@andrewharp - Thanks for the tips. I can build it using the command line above.

I replicated the switch in `srcs` from `android_tensorflow_lib_lite` and that also prevented the errors I was seeing while building correctly for android when I asked for that. I can raise a PR for that if you want / haven't already?

While this change makes it better, it might be more intuitive for users (like me) if this target failed to build with a message saying it should be built with android tools explicitly. Does it makes sense to build `android_tensorflow_lib` for anything except an android target? Then again I'm only new to bazel so it may well be just right as it is given the trade-off between build system complexity and user-friendliness.

Btw. I didn't have a problem building an app for Android using [the docs](http://bazel.io/docs/tutorial/android-app.html) and the `android/BUILD` example. I came across this error when trying to understand what steps bazel was doing internally in the `android_binary` command, just for curiosity's sake really.

I'm happy to close the PR if we can replace it with another as suggested by @andrewharp.

andrewharp:
I've submitted an internal patch that will nullify the src for android_tensorflow_lib in the same way as android_tensorflow_lib_lite. I agree that an informative error message might be ideal, but I'm worried about erroneous failures in automated builds and tests (not everything we run TF through seems to respect manual and notap at this point). As a compromise I added a comment on the targets specifying the targets must be built for Android, and how to do so.

darrengarvey:
Sounds good to me. Thanks for the help @andrewharp and @vrv.

PR #1654, opened 2016-03-26
replace libpng from google source to github source

fayeshine:
since google source will be blocked for people behind the wall, we could use the official release of libpng in github


vrv:
@girving: glennrp maintains libpng so this seems like a trustworthy github dependency -- what do you think?

girving:
Seems fine to me.  Note that the only risk of attack occurs when we change the hash.

girving:
Jenkins, test this please.

PR #1659, opened 2016-03-26
Updated year on License

the872:


the872:
I signed it!

vrv:
Thanks for the contribution!  IANAL, but I don't think this is needed -- you have to maintain at least the copyright year of the first distribution (2015).  Some resources online say it is okay to update on major updates / releases though, but I'd rather just keep the original year unless lawyercats suggest otherwise.   git history is a more accurate source of truth than what we put in this file anyway ;)

@martinwicke might know some lawyercats who say otherwise, so we can re-open if needed.

Issue #1453, opened 2016-03-10
Multidimensional RNN


phvu:
Hi all,

This is a feature request for [Multidimensional LSTM](http://people.idsia.ch/~juergen/nips2009.pdf). Is there any plan to support this in tensorflow?

I might be able to help with implementation if there is such a plan.

Cheers,

girving:
@ebrevdo: Want to comment on this?  Fancier variants of LSTMs would be great to have, the only question is whether they should go in core tensorflow, the contrib directory, or the models repo.  If you can make the interface similar to Eugene's existing RNN cell setup contrib might be the best place at first.

ebrevdo:
Right now it's possible to implement various types of multidimensional RNNs by feeding in your data as time being one direction (say, x), taking the output of the RNN, transposing it, and feeding it into a second RNN.  etc.  Alternatively feed your data & its transpose into separate RNNs (possibly with tied weights) and depth-concatenate the results.  And maybe feed the result into another RNN.

This allows one to implement both separable and non-separable multidimensional RNNs.  Not sure we need the sugar until people show it's superior to convolutional networks for something, and there's lots of use cases.

**That said**, it would be really nice to implement GridLSTM (also by Alex Graves, and according to their paper showing a lot of promise with fewer parameters).  GridLSTM uses standard LSTM cells, so it would be a function that uses RNNCell objects.  It connects them in interesting, convolutional, ways.

GridLSTM would probably go well in contrib, until enough people use it and it shows to be a very successful technique.

phvu:
Thanks, GridLSTM looks interesting. I will give it a shot. [This](https://github.com/coreylynch/grid-lstm) seems to be a good starting point.

I do have a use-case for this, but it is still in a pilot study. Will be nice to make it into contrib until there are more use-cases.

selcouthlyBlue:
> Right now it's possible to implement various types of multidimensional RNNs by feeding in your data as time being one direction (say, x), taking the output of the RNN, transposing it, and feeding it into a second RNN. etc.

How exactly do I transpose the output? Say I have this code:

```
input = tf.placeholder(tf.float32, [batch_size, time_steps, num_features]
cell = rnn.BasicLSTMCell(num_units=1)
lstm1 = tf.nn.dynamic_rnn(cell, input, dtype=tf.float32)
tranposed_lstm_output = some_transpose_function(lstm1) # how does this work?
[...]

PR #1665, opened 2016-03-27
Added implementation for GridRNN

phvu:
As discussed here: https://github.com/tensorflow/tensorflow/issues/1453

The implementation is generic, users can specify the number of dimensions and various configurations for those dimensions (input/output/priority/non-recurrent). The type of the cells along dimensions can also be selected among LSTM, GRU, vanilla RNN.

Come with unittests for basic types: 2LSTM (tied weights, non-recurrent), 2BasicLSTM and 2RNN.

I made a simple test of Grid2LSTM for character-level language modeling: https://github.com/phvu/grid-lstm-tensorflow/tree/master/char-rnn.


ebrevdo:
Will take a look tomorrow.

tensorflow/contrib/grid_rnn/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+## This package provides classes for GridRNN
+
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+# pylint: disable=unused-import,wildcard-import, line-too-long
+from tensorflow.contrib.grid_rnn.python.ops.grid_rnn_cell import *
```
ebrevdo: add newline

tensorflow/contrib/grid_rnn/python/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
ebrevdo: add newline and probably also the copyright (remember 2016)

tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
```diff
@@ -0,0 +1,324 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
```
ebrevdo: 2016 here and elsewhere

tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
```diff
@@ -0,0 +1,324 @@
...
+        self.assertEqual(cell.output_size, 2)
+        self.assertEqual(cell.input_size, 3)
+        self.assertEqual(cell.state_size, 8)
+        g, s = cell(x, m)
+        sess.run([tf.initialize_all_variables()])
+        res = sess.run([g, s], {x.name: np.array([[1., 1., 1.]]),
+                             m.name: np.array([[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8]])})
+        self.assertEqual(res[0].shape, (1, 2))
+        self.assertEqual(res[1].shape, (1, 8))
+        self.assertAllClose(res[0], [[ 0.36617181, 0.36617181]])
```
ebrevdo: how are you getting these numbers?  are they just a smoke test?

if so, convert all these tests to just making sure that they're not nan/inf and verifying that the shapes are correct.

also add checks on g.get_shape() and s.get_shape()

tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
```diff
@@ -0,0 +1,324 @@
...
+      with tf.variable_scope('root', initializer=tf.constant_initializer(0.2)) as root_scope:
+        x = tf.zeros([1, 3])
+        m = tf.zeros([1, 8])
+        cell = tf.contrib.grid_rnn.Grid2BasicLSTMCell(2, input_size=3)
+        self.assertEqual(cell.output_size, 2)
+        self.assertEqual(cell.input_size, 3)
+        self.assertEqual(cell.state_size, 8)
+        g, s = cell(x, m)
+        sess.run([tf.initialize_all_variables()])
+        res = sess.run([g, s], {x.name: np.array([[1., 1., 1.]]),
```
ebrevdo: change x.name to just x (can pass the object itself); here and elsewhere.

tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
```diff
@@ -0,0 +1,324 @@
...
+
+class GridRNNCellTest(tf.test.TestCase):
+
+  def testGrid2BasicLSTMCell(self):
+    with self.test_session() as sess:
+      with tf.variable_scope('root', initializer=tf.constant_initializer(0.2)) as root_scope:
+        x = tf.zeros([1, 3])
+        m = tf.zeros([1, 8])
+        cell = tf.contrib.grid_rnn.Grid2BasicLSTMCell(2, input_size=3)
+        self.assertEqual(cell.output_size, 2)
```
ebrevdo: we're getting rid of both input_size and output_size; output_size is not needed and input_size can be replaced with using get_shape() on the input inside the cell.  please remove these from tests and update the cell accordingly.

tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
```diff
@@ -0,0 +1,324 @@
...
+
+  def testGridRNNEdgeCasesNoOutput(self):
+    with self.test_session() as sess:
+      with tf.variable_scope('root', initializer=tf.constant_initializer(0.5)):
+        x =  tf.zeros([1, 2])
+        m = tf.zeros([1, 4])
+
+        # This cell produces no output
+        cell = tf.contrib.grid_rnn.GridRNNCell(num_units=2, num_dims=2, input_dims=0, output_dims=None,
+                                               non_recurrent_dims=0, non_recurrent_fn=tf.nn.relu)
```
ebrevdo: as elsewhere, remove input_dims and output_dims

tensorflow/contrib/grid_rnn/python/ops/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
ebrevdo: nit: add newlines here and elsewhere; maybe also the license header (2016 etc)

tensorflow/contrib/grid_rnn/python/ops/grid_rnn_cell.py
```diff
@@ -0,0 +1,353 @@
...
+  for i in dim_indices:
+    d = conf.dims[i]
+    if d.non_recurrent_fn:
+      linear_args = [cell_inputs, last_dim_output] if conf.num_dims > 1 else last_dim_output
+      if conf.tied:
+        with vs.variable_scope('non_recurrent', reuse=not(first_call and i == dim_indices[0])):
+          new_output[d.idx] = d.non_recurrent_fn(rnn_cell.linear(args=linear_args,
+                                                                 output_size=conf.num_units, bias=True))
+      else:
+          new_output[d.idx] = d.non_recurrent_fn(rnn_cell.linear(args=linear_args,
```
ebrevdo: can you use the contrib.layers.fully_connected here instead?  we're trying to get rid of rnn_cell.linear with fire.

tensorflow/contrib/grid_rnn/python/ops/grid_rnn_cell.py
```diff
@@ -0,0 +1,353 @@
...
+
+class Grid2GRUCell(GridRNNCell):
+  """2D LSTM cell
+    This creates a 2D cell which receives input and gives output in the first dimension.
+    The first dimension can optionally be non-recurrent if `non_recurrent_fn` is specified.
+  """
+
+  def __init__(self, num_units, input_size=None, tied=False, non_recurrent_fn=None):
+    super(Grid2GRUCell, self).__init__(num_units=num_units, input_size=input_size, num_dims=2,
+                                        input_dims=0, output_dims=0, priority_dims=0, tied=tied,
```
ebrevdo: nit: whitespace.

ebrevdo:
Thanks for the hard work!  Some comments.

tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
phvu: `input_dims` and `output_dims` are the dimension indices of the gridRNN cell. These are different from `input_size` and `output_size`.

phvu:
@ebrevdo Thanks for the comments. I updated the code.

- Good to know about `input_size` and `output_size`, I updated the code to not depend on those.
However as the `LSTMCell` is still accepting `input_size`, I still keep `input_size` as a parameter in the `cell_fn` callback. Should it be also removed?

- The tests are real. I computed the values using [this](https://docs.google.com/spreadsheets/d/18bkheoJbmMUqdRFrviUy_TiooSjvvpDqiti7hm2EASY/edit#gid=0). Since the initializer is fixed in the tests (weights are initialized to 0.5 or 0.2, depending on the tests; bias are initialized to zero), the output are deterministic. But of course I can switch to lightweight asserts if that is preferred.

tensorflow/contrib/grid_rnn/__init__.py
phvu: done

tensorflow/contrib/grid_rnn/python/__init__.py
phvu: done

tensorflow/contrib/grid_rnn/python/ops/__init__.py
phvu: done

ebrevdo:
Without a unit test that calls tf.nn.rnn or tf.nn.dynamic_rnn, it's not clear that these cells interact correctly with those methods.  Can you add such a test for your classes?

phvu:
ok will do.

phvu:
I added the tests for `Grid1LSTM`, `Grid2LSTM`, `Grid3LSTM` (with ReLU), trained with `tf.nn.rnn`. Can you take a look?

ebrevdo:
Will look tomorrow-thanks!

ebrevdo:
Jenkins, test this please?

gunan:

ebrevdo:
Sorry for not reviewing earlier.  Let's retest this and we can get it merged.

ebrevdo:
@phvu can you rebase on HEAD and run the tests?

vrv:
@ebrevdo: why does he have to rebase to HEAD?  Do you think there will be a conflict in doing so?

Also, only we can trigger tests.

vrv:

phvu:
Cool, tests passed. So I don't need to rebase master for now?

ebrevdo:
Excellent, we're good to go.  Thanks for the contribution!  If you have any example code using this we can consider adding it elsewhere in the repo (e.g. under models).

phvu:
Cool, thanks for all the help.
Unfortunately I can't share what I am working on with this yet. I have an application of this to character-level modeling (https://github.com/phvu/grid-lstm-tensorflow/tree/master/char-rnn) but it isn't entirely my code so I put it in a separated repo.

I think at some point I can try reproducing some experiments in the paper. Will submit PRs by then.

jstaker7:
Hi @phvu,

Thanks very much for your work. I'm a little confused as to how this can be applied generally. For example, in the paper a 3-LSTM network was applied to image patches of the MNIST dataset where each patch is c * m units long (which is equal to the depth dimension). So we have a 3D grid of c * m vectors. But in your implementation of __call__, we are expecting a tensor of size batch * c * m * num_dims. For MNIST, shouldn't it support b * c * m * dim1_size * dim2_size ? This would require input_dims to be something like [1, 2] (where dim 0 is batch); but I don't see any tests covering cases besides input_dim=0. Am I missing something?

phvu:
Hi @jstaker7,
There are some nuances here.
First, the `__call__` method expects the state tensor of `batch_size * ( (c + m) * num_dims)`. When c=m=2 then c*m = c+m, so sorry for the bad test case, I should have use other values.

Second, in the MNIST case, (as far as I understand) I suppose you could use Grid3LSTMCell (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/grid_rnn/python/ops/grid_rnn_cell.py#L242). This cells receives input and gives output in the first dimension (index 0), and uses LSTM in the other 2 dimensions.
In order to replicate Figure 11 in the paper, you will need to construct 4 3-LSTM cells, each cell will handle 1 scan direction of the input image. However the first hidden LSTM layer receives original pixels as input, while other hidden LSTM layers receives the output of the LSTM just below it.
How to do that is up to you. One simple way is to have a loop to scan the image in a given scan direction (say left-to-right top-to-bottom), which will give a sequence of vectors, and then feed that sequence into the corresponding 3-LSTM cell for that scan direction. For bigger images, you might want to use `tf.scan()`

So your assumption that input_dims should be `[1,2]` is not true. The cells should receive input at dimension 0. For me this is the only reasonable interpretation of the paper, which also take 1-LSTM and 2-LSTM cells into account. (unless the authors release their implementation so that we can do a comparison).

jstaker7:
Hi @phvu,

Thank you, this helps quite a lot. Since each dimension will always have its own LSTM, does that mean we will never use an `input_dims` other than 0?

phvu:
To replicate the experiments in the paper, yes. But I am not sure about that generally.

When we set dimension i to be input_dims, it simply means for each __call__, the cell expects an input tensor for dimension i. For non-input dimensions, the cell will take the recurrent values (extracted from the state tensor) as input.

So I guess we can be creative and feed inputs into more than one dimension. I used dimension 0 as input and output dimensions just as a convention. You can as well construct your own GridRNNCell with any configuration.

jstaker7:
Perfect, thanks so much for the clarification. Super helpful and makes a lot of sense.

This might also be interesting to look into in the future. Grid LSTMs were mentioned at it looks like there was a recent merge: https://github.com/tensorflow/tensorflow/issues/2560

PR #1667, opened 2016-03-27
Update fully_connected_feed.py

zhongzyd:
"os.path" and "numpy" was unused, so I removed it.
The tfevents-output-file was not completely written in IPython, so I Added "summary_writer.flush()" to fix it.


vrv:
This is a useful change, but can you make it to the master branch instead of r0.7 ?

zhongzyd:
Thanks for reminded me, I will create a pull request to master.

PR #1679, opened 2016-03-28
Add ability to check for a specific bazel version. Supports

vrv:
bazel versions pre-the ability to check for versions.


vrv:
@damienmg fyi

martinwicke:
I thought we needed 2.0?

vrv:
We currently don't need 2.0 as checked in:

2.0 is needed for protobuf as a submodule
The next pending release is needed for iOS support.

So this is just encoding the requirements at HEAD

PR #1685, opened 2016-03-29
Re-add ability to check for a specific bazel version

vrv:
I tested this by upgrading to bazel master -- when i set minimum version to 0.2.1, it failed, and when I set it to 0.2.0, it passed.  Setting it to 0.1.4 currently also passes.


vrv:
FYI to @damienmg, @petewarden @martinwicke and @fayeshine

WORKSPACE
```diff
@@ -16,6 +16,10 @@
 load("//tensorflow:workspace.bzl", "tf_workspace")
 tf_workspace()
 
+# Specify the minimum required bazel version.
+load("//tensorflow:tensorflow.bzl", "check_version")
+check_version("0.1.4")
```
damienmg: FWIW 0.1.4 will have no effect here because the bazel_version was added after 0.2.0 (so only > 0.2.1 will have an effect)

PR #1673, opened 2016-03-28
replace googlemock source from google site to github

fayeshine:
google site is blocked for people under the wall, we could help them by using official google github repo

PR #1687, opened 2016-03-29
replace googlemock source from google site to github

fayeshine:
@vrv , this is to replace #1673

google site is blocked for people under the wall, we could help them by using official fedora's site
This should work since the SHA256 are the same

### DIFF
--- a/tensorflow/workspace.bzl
+++ b/tensorflow/workspace.bzl
@@ -6,7 +6,7 @@
 def tf_workspace(path_prefix = "", tf_repo_name = ""):
   native.new_http_archive(
     name = "gmock_archive",
-    url = "http://pkgs.fedoraproject.org/repo/pkgs/gmock/gmock-1.7.0.zip/073b984d8798ea1594f5e44d85b20d66/gmock-1.7.0.zip",
+    url = "https://archive.openswitch.net/gmock-1.7.0.zip",
     sha256 = "26fcbb5925b74ad5fc8c26b0495dfc96353f4d553492eb97e85a8a6d2f43095b",
     build_file = path_prefix + "google/protobuf/gmock.BUILD",
   )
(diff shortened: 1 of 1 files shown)


tensorflow/workspace.bzl
```diff
@@ -6,7 +6,7 @@
 def tf_workspace(path_prefix = "", tf_repo_name = ""):
   native.new_http_archive(
     name = "gmock_archive",
-    url = "https://googlemock.googlecode.com/files/gmock-1.7.0.zip",
+    url = "http://pkgs.fedoraproject.org/repo/pkgs/gmock/gmock-1.7.0.zip/073b984d8798ea1594f5e44d85b20d66/gmock-1.7.0.zip",
```
vrv: is there a site that has https at least?

fayeshine:
@vrv 
"https://archive.openswitch.net/gmock-1.7.0.zip" should work
openswitch is Community-Based, Open Source Operating System, should be trusty, and it should be safe as long as the SHA256 is unchanged

vrv:
@martinwicke: why don't we just host this ourselves on tensorflow.org instead?

martinwicke:
No reason apart from fear of accumulating things requiring maintenance.
Gmock is stable enough. We can add this along with our hooks.

vrv:
@fayeshine: can you try accessing https://storage.googleapis.com/download.tensorflow.org/deps/gmock-1.7.0.zip or is that also blocked?

fayeshine:
@vrv , this link (https://storage.googleapis.com/download.tensorflow.org/deps/gmock-1.7.0.zip) is blocked for me, :(
besides, "tensorflow.org" is very slow here, sometimes unreachable

vrv:
Well that's unfortunate.  I think openswitch is okay for now, but we should probably find a better long-term solution.  One option to use: https://github.com/google/googletest, if that works, since that's the more-or-less official github one now

vrv:
(Specifically, using git_repository instead of http_archive, and changing the build_file to something that works for that github repo, rather than the hardcoded 1.7 paths)

vrv:

PR #1697, opened 2016-03-29
Update fully_connected_feed.py

zhongzyd:
"os.path" and "numpy" were unused, so I removed them.
The tfevents-output-file was not completely written in IPython, so I added "summary_writer.flush()" to fix it.


vrv:

PR #1709, opened 2016-03-30
Update build_pip_package.sh for generating a tarball

bug-fixed:
Update build_pip_package.sh for generating a tarball

One can generate a single wheel file using:

```
    bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
```

One can also generate an extra tarball file using:

```
    bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg tarball
```

### DIFF
--- a/tensorflow/tools/pip_package/build_pip_package.sh
+++ b/tensorflow/tools/pip_package/build_pip_package.sh
@@ -59,7 +59,20 @@ function main() {
   mkdir -p ${DEST}
   cp dist/* ${DEST}
   popd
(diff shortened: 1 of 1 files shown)


zszhong:
Hello, I signed it!

zszhong:
Close this as it is a mistake. The tarball only contains the generated dynamic libraries that might be not suitable for other environment, e.g., the `GLIBC_2.14 not found` error.

PR #1710, opened 2016-03-30
move dropout to later chapters where it can truly work

fayeshine:
@Sohl-Dickstein 
This is to replace #1620 ,
remove dropout in the very beginning tutorial on MNIST since it doesn't work in that network architecture,
and move it to Tensorflow Mechanics 101, where it can truly work and we give more detailed explanation on dropout:

`After all training epochs, the accuracy on test data will be around 98.5%.
However if we remove dropout in the network architecture, the accuracy will
decrease to 98.2%. This is because dropout can reduce the effect of overfitting.
If we do not have dropout, the accuracy on training data will eventually become
100%, however the accuracy on test data will soon increase to 98.2% but get
stalled there. By adding dropout, the accuracy on training data increase slower,
but the generalization ability will become better.`

### DIFF
--- a/tensorflow/examples/tutorials/mnist/fully_connected_feed.py
+++ b/tensorflow/examples/tutorials/mnist/fully_connected_feed.py
@@ -34,7 +34,7 @@ from tensorflow.examples.tutorials.mnist import mnist
 flags = tf.app.flags
 FLAGS = flags.FLAGS
 flags.DEFINE_float('learning_rate', 0.1, 'Initial learning rate.')
-flags.DEFINE_float('keep_prob', 0.5, 'Dropout keep rate.')
+flags.DEFINE_float('keep_prob', 0.5, 'Dropout keep probability.')
 flags.DEFINE_integer('max_steps', 20000, 'Number of steps to run trainer.')
 flags.DEFINE_integer('hidden1', 1024, 'Number of units in hidden layer 1.')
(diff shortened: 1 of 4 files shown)


fayeshine:
@Sohl-Dickstein ,what's your opinion?

Sohl-Dickstein:
Sorry for the delay. I will get a change review back to you in two days
(Monday PST). Thank you for the contribution!

Sohl-Dickstein:
I have some concerns about this pull request. This modification requires training for 10 times more learning steps, with twice the batch size, and with a model that has several hundred times the number of parameters. We want the models in these tutorials to run as quickly as possible, and this seems like a big step backwards in that respect. Even with that size increase, the performance change due to dropout is very small. There are also two images in the tutorial (graph visualization, and tensorboard screen shot) that will no longer be accurate after this change.

Let me suggest a much more light weight change. How about placing a footnote link at the end of the first sentence in the dropout section in the original MNIST for Experts tutorial, then adding the following footnote text to the end of the tutorial:

"For this small convolutional network, performance is actually nearly identical with and without dropout. Dropout is often very effective at reducing overfitting, but it is most useful when training very large neural networks."

Instructions for adding footnotes in markdown are here:

http://stackoverflow.com/questions/25579868/how-do-i-add-footnotes-to-github-flavoured-markdown

PR #1713, opened 2016-03-30
Update to correct method name in local example.

etrain:
There is an error in the current distributed TensorFlow documentation which caused me a little trouble when verifying that installation was correct.


vrv:
Thanks! (I know it sucks to have to sign a CLA for a doc change).

etrain:
Oh man - didn't see that one coming. I don't want to sign this before running this by a lawyer. I'm going to close this PR for now. If someone who has signed it (or is an Alphabet Employee) gets motivated they can make a similar change.

PR #1719, opened 2016-03-31
Reduction of CPU test time for cwise_ops_test and rnn_test.

vrv:
Reduce test scenarios for CPU-only tests by checking for whether
cuda has been enabled in the binary to avoid running CPU tests twice.

Also avoids testing the cross product of broadcasting shapes, dtypes,
and functions in the bcast code.  Instead, we only test broadcasting
of all the various shapes for exactly one datatype and function (add),
which should test the broadcasting behavior (which should not be function
or type specific), and then testing only the cross product of
datatypes and functions with one shape.  Reduces cwise_ops_test
in non-opt mode from 100 seconds to 20.

Also reduce the number of timesteps in one of the rnn_test tests
from 8 to 3 (enough to check the recurrency).  The changes reduce
the runtime from 65 seconds to 19, and there are probably other
changes that can be made to the test to reduce it further.

### DIFF
--- a/tensorflow/python/BUILD
+++ b/tensorflow/python/BUILD
@@ -1084,16 +1084,19 @@ medium_kernel_test_list = glob([
     "kernel_tests/concat_op_test.py",
     "kernel_tests/division_future_test.py",
     "kernel_tests/fft_ops_test.py",
-    "kernel_tests/rnn_test.py",
     "kernel_tests/scatter_ops_test.py",
     "kernel_tests/seq2seq_test.py",
     "kernel_tests/slice_op_test.py",
 ])
 
 sharded_kernel_test_list = glob([
-    "kernel_tests/cwise_ops_test.py",
     "kernel_tests/embedding_ops_test.py",
     "kernel_tests/linalg_grad_test.py",
+])
+
+exclusive_kernel_test_list = glob([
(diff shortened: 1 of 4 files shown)


vrv:
rnn_test and cwise_ops_test timed out at 65 seconds, even though I ran them locally and they run in far less.

testing again, I have a feeling there's interference due to recent changes that cause multiple tests to run at the same time.  Since all the other tests are cached, only these two tests will run and we'll see the actual runtime.

vrv:
Interesting, I tried running rnn_test and cwise_ops_test as exclusive mode.  on the cpu_python3 suite they both pass in ~20 seconds.  on normal cpu python2 they were running exclusively (I watched the log) but they still timed out -- not sure what's going on anymore :(

@caisq, @jendap for advice?

caisq:

caisq:
@vrv You think --jobs=1 will help? We could change the scripts so that they run these two tests with --jobs=1 and the rest in the normal (parallel) way.

vrv:
It's possible --jobs=1 will help -- I'm still unsure myself why python3 linux is fine but the python2 linux one isn't :(

caisq:

vrv:
Well, I tried.  Maybe someone else can figure out what's broken about this.

PR #1726, opened 2016-03-31
Add support for development version of Bazel when detecting the version

damienmg:
Development version of Bazel (especially built on ci.bazel.io) are setting their version to empty, making the test fails with error.

/cc @vrv


vrv:

PR #1730, opened 2016-03-31
update description for Tensorflow Mechanics 101 Tutorial

rcullito:
I changed: "We use again MNIST as the example." to "We again use MNIST as the example." as this seemed to make more sense.


rcullito:
will submit again with signed git username

PR #1721, opened 2016-03-31
[bugfix] chase version to tensorflow/workspace.bzl

clsung:
Fix #1720 . With original version we will have errors like:

> /Users/clsung/git/tensorflow/third_party/eigen3/unsupported/Eigen/CXX11/Tensor:1:10: fatal error:
>       'eigen-eigen-3d9f227afae2/unsupported/Eigen/CXX11/Tensor' file not found
> # include "eigen-eigen-3d9f227afae2/unsupported/Eigen/CXX11/Tensor"
> 
> ```
>      ^
> ```
> 
> 1 error generated.

PR #1738, opened 2016-04-01
Fix Linux CPU Tests CMAKE

clsung:
Changes:
- target_compile_features require CMake 3.1.0
- add install_proto3.sh for protobuf3, which required by the CMake
- add ${CMAKE_DL_LIBS} for -ldl
- "make all test" to "make all" since there's no add_test() in tensorflow/contrib/cmake/

Ref:
https://cmake.org/cmake/help/v3.1/release/3.1.0.html

Related PR: #1721


clsung:
Alternative of protobuf3 installation is setup --recurse-submodules in Jenkins to clone/update google/protobuf. However that can not be done here.

caisq:

caisq:
There is still a failure in the cmake PR build: http://ci.tensorflow.org/job/tensorflow-pull-requests-cpu-cmake/9/console

clsung:
I've merged #1721 in this PR.

caisq:

caisq:
The renaming "tensorflow/core/ops/compat/update_ops.cc → ...orflow/core/ops/compat/update_ops_main.cc" has broken CPU builds.

clsung:
Just fixed it. Can you please give a try? I may also need to setup another jenkins task for it. :-/

caisq:
Will cmake work for gpu builds? If so, the install_proto3.sh needs to be added to tensorflow/tools/ci_build/Dockerfile.gpu as well.

caisq:

clsung:
I'll need to grab a AWS AMI to setup Jenkins to setup with GPU support to test it. How about separate gpu-specifics to another PR?

caisq:
@clsung Let me help you test it later. We have GPU machines available. Stay tuned.

clsung:
Since nvidia/cuda:7.5-cudnn4-devel is from ubuntu:14.04, we should only need to copy one line from Dockerfile.cpu to Dockerfile.gpu:

`RUN /install/install_proto3.sh`

Where we already install autoconf, automake and cmake 3.x in install_deb_packages.sh.

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # Minimum CMake required
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
```
jendap: Does it need 3.1? Would 3.0 work? That is what is in stable debian.

tensorflow/tools/ci_build/install/install_deb_packages.sh
```diff
@@ -19,17 +19,22 @@ set -e
 # Install FFmpeg from an alternate repository.
 apt-get install -y software-properties-common
 add-apt-repository -y ppa:mc3man/trusty-media
+# Install CMake 3.x
+add-apt-repository -y ppa:george-edison55/cmake-3.x
```
jendap: @vrv This ppa should be moved to Dcokerfile.cpu, Dockerfile.gpu and Dockerfile.android. As well as the ffmpeg ppa above. They are workaround for using old trusty. Not needed on new ubuntu or debian image. Debian nightly is broken ever since the ffmpeg pa has made it in!

jendap:
@clsung if cmake 3.0 would be enough then please put it into your next PR.

@vrv I will make PR to move the ppa repositories.

Both... 48 lines in 17 commits? Should we squash it next time? :-)

vrv:
https://github.com/tensorflow/tensorflow/commit/2c2f422a628d83a591fb5d20d46deecc47655ead it was squashed.

jendap:
Right! That's why I had to look for the PR and could not just click through :-)

Github should really make squash+rebase a feature! (even a default option)

vrv:
It is a feature -- it's what we used here.  I guess it doesn't modify the original pull request, that's all.

tensorflow/contrib/cmake/CMakeLists.txt
clsung: According to https://cmake.org/cmake/help/v3.1/release/3.1.0.html

> New target_compile_features() command allows populating the COMPILE_FEATURES target property, just like any other build variable.

So it should only works from 3.1

PR #1746, opened 2016-04-02
Upstream changes for april fools

vrv:


caisq:
So that this is a april fool's prank is an april fool's prank? :p

vrv:
android demo app seems to have had an infrastructure failure (groupadd problem).  Trying again.

vrv:
Looks persistent -- not sure what we can do -- any ideas @jendap or @caisq ?  This has happened once before, it would be good to root cause why the issue comes back.

```
Running '/tmp/tf_build.sh' inside jenkins-tensorflow-pull-requests-android-549.android...
Adding group `jenkins' (GID 120) ...
groupadd: failure while writing changes to /etc/group
addgroup: `/usr/sbin/groupadd -g 120 jenkins' returned error code 10. Exiting.
```

caisq:
@vrv @jendap On the android slave, the file system inside docker container have become read-only. This is the reason why addgroup fails in with_the_same_user.sh. I'm looking into why docker locks the file system.

caisq:

caisq:
@vrv @jendap OK, I re-created /var/lib/docker on the android slave and restarted the docker service. The problem seems to be gone for now. Why the file systems became read-only in docker containers is still unclear to me. If you have any thoughts or clues, please let me know. We'll keep an eye on the android builds.

vrv:
Using a merge commit here instead of squashing.  I wish we could not squash but just rebase.

PR #1753, opened 2016-04-03
correcting bottleneck residual function

daviddao:
As described in He _et al._ [1] the first 1x1 convolution inside the bottleneck architecture reduces the dimension to save FLOPs. Hence the correct number of filters is `block.bottleneck_size` instead of `block.num_filters`.

06cf6894-d197-11e5-8f2f-33d48c225961 2

[1] http://arxiv.org/pdf/1512.03385.pdf


ilblackdragon:

ilblackdragon:
Thanks, good catch!

PR #1754, opened 2016-04-03
fix cuda warning

Mistobaan:
I see this error message when compiling with nvcc 7.5

### DIFF
--- a/tensorflow/tensorflow.bzl
+++ b/tensorflow/tensorflow.bzl
@@ -256,7 +256,7 @@ def tf_cuda_cc_tests(tests, deps, tags=[], size="medium"):
 def tf_gpu_kernel_library(srcs, copts=[], cuda_copts=[], deps=[], hdrs=[],
                           **kwargs):
   cuda_copts = ["-x", "cuda", "-DGOOGLE_CUDA=1",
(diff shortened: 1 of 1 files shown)


tensorflow/tensorflow.bzl
```diff
@@ -256,7 +256,7 @@ def tf_cuda_cc_tests(tests, deps, tags=[], size="medium"):
 def tf_gpu_kernel_library(srcs, copts=[], cuda_copts=[], deps=[], hdrs=[],
                           **kwargs):
   cuda_copts = ["-x", "cuda", "-DGOOGLE_CUDA=1",
-                "-nvcc_options=relaxed-constexpr", "-nvcc_options=ftz=true",
+                "-nvcc_options=expt-relaxed-constexpr", "-nvcc_options=ftz=true",
```
vrv: Have you tried seeing whether this breaks earlier nvcc's?  I don't know if at this point we're willing to break everybody but nvcc 7.5, given how many people have difficulty getting admins to upgrade software on clusters ;)

vrv:
Please verify that this doesn't break nvcc 7.0 and we'll re-open, thanks!

Issue #1502, opened 2016-03-15
BatchNorm kernel?


lukemetz:
Hello,
Adding batch norm to a model results in a significant slowdown (using tf.nn.batch_normalize).

I was wondering why a dedicated kernel was removed in favour of a symbolic approach?
(seemingly unrelated commit in question: https://github.com/tensorflow/tensorflow/commit/03869e6e9cfac85d3e0f87ded01f4dd8de10ed67)

Was this a design decisions to not want these "one off" style optimizations/kernels?

Having a faster kernel, (like cudnn #714, or something like what torch does) would be great if its not against tensorflows design. I will look into it if it is not.

Thank you.

ajtulloch:
Yeah, we've been fusing graph ops manually for performance, it'd be great to understand the rationale for going the other way.

mrry:
It looks like two commits were inadvertently squashed. Here's the intended description for the batch norm parts of that commit:

> General batchnorm op, suitable to both global normalization for convolutions and per-depth normalization of fully-connected layers. It also supports arbitrary layouts and orderings of the axes.
> 
> Nothing deep in these essentially 4 lines of code, but note:
> - The ops are ordered in a way that makes the computation much more efficient (thx Sergey). None of the existing wrappers that didn't use the fused kernel did that, and as a result were almost 2x slower according to the benchmark.
> - The benchmark confirms that this implementation is competitive with the handcrafted kernel.
> - I confirmed on inception that the step time or accuracy are not affected.
> - Added documentation and tests showing how the op can be used for arbitrary layouts.
> - The change in API compared with batch_norm_with_global_normalization makes it possible to not have a scale tensor when it's not used, which caused some confusion and unnecessary additional saved variables in the past.

@vincentvanhoucke is working on the related issue https://github.com/tensorflow/tensorflow/issues/1122 and might have more to add.

vincentvanhoucke:
If you're seeing a significant regression from the non-fused kernel, I'd be interested in hearing about it. Note that there were two changes:
- removing the fused kernel, which did not affect performance on any benchmark I could run. In particular, this one:
https://github.com/tensorflow/tensorflow/blob/0249729a26b2cd3bdbbc5880f8d634a56860b0fd/tensorflow/python/ops/batch_norm_benchmark.py
- replacing the two-pass moment accumulation algorithm with a one-pass algorithm. This was a tad slower on GPU, and faster on CPU.
On balance, those two changes gave us a lot more flexibility:
[...]

PR #1759, opened 2016-04-04
[WIP] CuDNN Batch Normalization Op

lukemetz:
This pull request is a WIP for adding a cudnn batchnorm op that works on NCHW data layout. Others and myself (https://github.com/tensorflow/tensorflow/issues/1502) have noticed very poor performance. Currently, the code is very rough and is just enough to do performance evaluations.

Code used to create the following numbers can be found here (https://gist.github.com/lukemetz/bdb58d2f006b64fe234cb04f7f0e8185). I tested 3 variations: current batch norm implementation plus moments in NHWC (tf_NHWC), current batch norm implementation and moments in NCHW (tf_NCHW), and my proposed op in NCHW format (cudnn_NCHW). All tests done were done on a 980, and the number listed is time taken for 100 iterations computing forward and backward passes.

| Shape | tf_NHWC | tf_NCHW | cudnn_NCHW | percent faster tf_NHWC | percent faster tf_NCWH |
| --- | --- | --- | --- | --- | --- |
| (32, 8, 8, 32) | 125.5 ms | 196.8 ms | 85.6 ms | 47% | 130% |
| (128, 4, 4, 32) | 133.4 ms | 171.7 ms | 77.1 ms | 73% | 123% |
| (256, 16, 16, 32) | 648.7 ms | 4624.5 ms | 388.6 ms | 67% | 1090% |
| (128, 64, 64, 32) | 4900.5 ms | 37107.5 ms | 3011.5 ms | 63 % | 1132% |
| (128, 128, 128, 32) | 19087.1 ms | 171175.2 ms | 11815.6 ms | 62% | 1349% |

(Note:  in my experience it is faster to transpose batchnorm transpose instead of using NCHW moments and batchnorm.)

When testing on an largish net on a single GPU, there are also performance improvements. My test network is an 8 layer residual-like network on 32x32 images with batchsize 128. 

NCHW batchnorm and HCHW convs: approximately 2700 examples a second. 

NHWC with Tensorflow implementation and NHWC convs: approximately 1900 examples a second. 

(Note: this difference is in large part to do with using NCHW convolutions with no transposes, but this was not possible before a fast NCHW batchnorm.)

The one odd thing I have encountered is that performance takes a hit when using multiple GPUs with data parallelism. . . It is actually slower than using a single GPU. If anybody has an idea as to why, or solutions, I would love to know. To my knowledge each GPU has one cuda stream so the cudnn locks shouldn't interfere with each other ???

If the concept behind this PR looks good, let me know. I wanted to share results early but there is a still a lot to do in terms of cleanup/standardization, error checking, testing, and documentation as well as a front end python api.

Thanks, all. Any comments appreciated.


vrv:
Awesome!  At a high level these numbers sound great to me, but we can't really take too much of a look until the CLA is signed.  If this makes sense, we'd be happy to work with you to get these changes in.

@vincentvanhoucke @zheng-xq @benoitsteiner as an early heads up.

lukemetz:
Thanks for the help!
I signed it!

vincentvanhoucke:
This looks very good from my POV. @vrv how do we handle stream executor changes like this?

vrv:
I think we now just accept them and we'll figure out how to backport them to the open source stream executor.  Unfortunately the best person to review this change is @zheng-xq who is out of the office this week and next.  I could offer style / tests suggestions, but I'd rather do that only after the high level review has been done.

ppwwyyxx:
Did you try using cudnn for NHWC (by adding a transpose, for example) as well? It will be very helpful if cudnn can beat the current kernel in both cases.

lukemetz:
@ppwwyyxx I have not yet. There is some special code to do fast NHWC -> NCHW used in their convolutions I need to add. My guess is that it will be comparable speeds but a well designed fused kernel with no transposes should always do the best.

vincentvanhoucke:
@zheng-xq did you have a chance to look at this?

tensorflow/core/ops/batchnorm_training_op.cc
```diff
@@ -0,0 +1,80 @@
...
+    .Input("running_mean: Ref(T)")
+    .Input("running_inv_var: Ref(T)")
+    .Output("out: T")
+    .Output("save_mean: T")
+    .Output("save_inv_var: T")
+    .Attr("T: {float}")
+    .Attr("epsilon: float")
+    .Attr("exponential_average_factor: float")
+    .Doc(R"doc(
+Perform batch normalization using batch statistics.
```
zheng-xq: TensorFlow default data_format is NHWC. So we need to support that at least. 

There are two ways: 

1. Call the Cudnn version of NHWC. 
2. Sometimes #1 is too slow, it is okay to call NHWCToNCHW first, do the op in NCHW, and then transform back to NHWC.

tensorflow/core/BUILD
```diff
@@ -305,6 +305,7 @@ tf_gen_op_libs(
         "state_ops",
         "string_ops",
         "training_ops",
+        "batchnorm_training_op",
```
zheng-xq: Keep them sorted as much as possible. Same things below.

tensorflow/core/kernels/cudnn_batchnorm_op.cc
```diff
@@ -0,0 +1,329 @@
...
+      OP_REQUIRES_OK(context, context->MatchSignature({dt, dt, dt, dt_ref, dt_ref}, {dt, dt, dt}));
+
+      //Do some type checking here
+      OP_REQUIRES_OK(context, context->GetAttr("epsilon", &epsilon_));
+      OP_REQUIRES_OK(context, context->GetAttr("exponential_average_factor",
+                                               &exponential_average_factor_));
+    }
+
+    void Compute(OpKernelContext* context) override {
+      //TODO a whole bunch of error checking
```
zheng-xq: Agree. We should check all the dimension compatibilities here.

tensorflow/core/ops/batchnorm_training_op.cc
```diff
@@ -0,0 +1,80 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "tensorflow/core/framework/op.h"
+
+namespace tensorflow {
+
+REGISTER_OP("BatchNormalizeTraining")
```
zheng-xq: It is okay to have a separate BatchNormalizeInference op for now. However, in the future, we really want a single "BatchNorm" that can choose "inference" and "training" variants depending on whether gradients are needed. That will be an improvement down the road. 

Regarding the name, TF calls it BatchNorm. Both Cudnn and the original paper use the term BatchNormalization. I would prefer not adding a new variant, either one works for me.

tensorflow/core/ops/batchnorm_training_op.cc
```diff
@@ -0,0 +1,80 @@
...
+    .Input("input: T")
+    .Input("scale: T")
+    .Input("bias: T")
+    .Input("running_mean: Ref(T)")
```
zheng-xq: We shouldn't have this op modify the running_mean and running_inv_var directly. A few problems: 

1. making distributed training difficult, since all replicas needs to directly access the running mean; 
2. synchronized training difficult, since multiple replicas do all the update without proper locking.

tensorflow/core/ops/batchnorm_training_op.cc
```diff
@@ -0,0 +1,80 @@
...
+    .Input("running_inv_var: Ref(T)")
+    .Output("out: T")
+    .Output("save_mean: T")
+    .Output("save_inv_var: T")
+    .Attr("T: {float}")
```
zheng-xq: The float-only type constraint seems a bit restrictive. We should at least support {float, double}. It is okay not to register a kernel that implements it, but the interface should be there.

tensorflow/core/ops/batchnorm_training_op.cc
```diff
@@ -0,0 +1,80 @@
...
+    .Attr("epsilon: float")
+    .Attr("exponential_average_factor: float")
```
zheng-xq: Cudnn asks for double as epsilon, and TF uses "float" here in the regular "BatchNorm". It is entirely possible for users to use different epsilon for float and double. So I slightly prefer "double" or even "T" here. 

@vrv, what do you think?

tensorflow/core/ops/batchnorm_training_op.cc
```diff
@@ -0,0 +1,80 @@
...
+  During training this reference is updated via a rolling average from 1 over the
+  current batch variance.
+out: A 4D output Tensor. The input after applying batch normalization.
+save_mean: A 1D Tensor. Computed means. To be used in the backward pass.
+save_inv_var: A 1D Tensor. Computed inverse variance. To be used in the backward pass.
+epsilon: float
+exponential_average_factor: float. Exponential decay on rolling average.
+)doc");
+
+REGISTER_OP("BatchNormalizeTrainingGrad")
```
zheng-xq: Unless we want people to call this op directly, it is better to mark this op as hidden, and use "_BatchNormalizeTrainingGrad".

tensorflow/python/ops/nn.py
```diff
@@ -801,6 +803,71 @@ def batch_norm_with_global_normalization(t,
...
+      mean: 1D tensor with computed means for the batch.
+      inv_var: A 1D tensor with computed inverse variances of the batch.
+    """
+    if data_format != "NCHW":
+        raise AttributeError("Currently only NCHW data_format is supported.")
+    output, mean, inv_var = batch_normalize_training(t,
+        gamma, beta, running_mean, running_inv_var, 0.01, 0.8)
+    return output, mean, inv_var
+
+def batch_norm_inference(t, beta, gamma, variance_epsilon, population_means,
```
zheng-xq: Have you measured the performance difference here, against the Cudnn implementation? 

The straightforward implementation writes and then reads intermediate data multiple time, and may have a performance implication. It is a good idea to actually check out.

zheng-xq:
@lukemetz, thank you for the contribution!

I've put in my first round of comments. We can go through more details later.

tensorflow/core/ops/batchnorm_training_op.cc
lukemetz: This was concerning to me as well but I am unsure of a good solution. What alternative do you propose? I see a few possible options:
1.) have a old running mean that gets passed in and a updated running mean that gets returned.
2.) ignore this option all together
3.)??

tensorflow/python/ops/nn.py
lukemetz: Nope I have not. I am working under the assumption that Cudnn will be a lot faster but am lazy for now and want to have one less thing to test while still working on this. I will do the test some point later.

lukemetz:
Thank you for the comments @zheng-xq. When I can find time I will fix and add where necessary and will ping back.

tensorflow/core/ops/batchnorm_training_op.cc
lukemetz: Correct me if I am wrong, but I do not see any support for Cudnn batch norm in the NHWC format in the header.
For now I will do the NHWCToNCHW calls and profile.

tensorflow/core/ops/batchnorm_training_op.cc
zheng-xq: I would recommend to output the current mean and variance directly. The update itself can happen in a different node in the graph with tf.assign_moving_average.

vrv:
Hi @lukemetz, any updates?

tensorflow/core/ops/batchnorm_training_op.cc
lukemetz: When shifting to T or double, somewhere in TF's code generation fails. Does this make sense? An alternative could be to make epsilon a tensor input and not an attr but that seems a little extreme.

```
  /bin/bash -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; bazel-out/host/bin/tensorflow/python/gen_batchnorm_training_op_py_wrappers_cc BatchNormTrainingGrad 0 > bazel-out/local_linux-opt/genfiles/tensorflow/python/gen_batchnorm_training_op.py').
F tensorflow/core/framework/op.cc:132] Check failed: ::tensorflow::Status::OK() == (RegisterAlreadyLocked(op_def)) (OK vs. Invalid argument: Unrecognized type '' in attr 'epsilon'; in OpDef: name: "BatchNormTraining" input_arg { name: "input" description: "A 4D input Tensor." type_attr: "T" } input_arg { name: "scale" description: "A 1D Tensor with size equal to the number of channels.\nThis is the learned scale value multiplied post normalization.\nAlso known as gamma." type_attr: "T" } input_arg { name: "bias" description: "A 1D Tensor with size equal to the number of channels.\nThis is the learned bias value added post normalization.\nAlso known as beta." type_attr: "T" } output_arg { name: "out" description: "A 4D output Tensor. The input after applying batch normalization." type_attr: "T" } output_arg { name: "save_mean" description: "A 1D Tensor. Computed means. To be used in the backward pass." type_attr: "T" } output_arg { name: "save_inv_var" description: "A 1D Tensor. Computed inverse variance. To be used in the backward pass." type_attr: "T" } attr { name: "T" type: "type" allowed_values { list { type: DT_FLOAT type: DT_DOUBLE } } } attr { name: "epsilon" description: "float" } attr { name: "data_format" type: "string" } summary: "Perform batch norm using batch statistics.")Attempting to register: Op<name=BatchNormTraining; signature=input:T, scale:T, bias:T -> out:T, save_mean:T, save_inv_var:T; attr=T:type,allowed=[DT_FLOAT, DT_DOUBLE]; attr=epsilon:; attr=data_format:string>
```

tensorflow/core/ops/batchnorm_training_op.cc
lukemetz: I have marked as hidden in bazel.
When adding the underscore to the Op definition I cannot find where the code generation puts it nor an example op that uses this underscore notation in op definition. Any ideas?

lukemetz:
Work / life has been a bit hectic lately. Sorry for the long delay.
I think I addressed all of the comments as well as added a few tests.

What still needs to be done (possibly not this PR):
1.) Adding support for inference.
2.) fully connected (2D) batch normalization -- should this be a separate op? If not, what should be done with data_format?

Let me know what else needs to be fixed and thanks for the help.
@vrv @zheng-xq

metap:
So I recompliled from source based on the pull request.
Several problems with the code:
1. `tf.nn.batch_norm_training` only returns `output` while the docstring says it returns `output`, `mean` and `inv_var`. Also several typos in the docstring
2. Shape info is lost on the three variables mentioned above after `batch_norm_training`; had to manually set the shape afterwards
3. Seems to be sensitive against small epsilon. Something as "big" as 1e-5 gives cudnn launch error
4. This one is important: **gives OOM error** after fixed number of executions. On my 980 Ti, the test code in OP stops at `i=81` in the warmup when I set `shape = [128, 32, 32, 128]`

metap:
Took me a while but I found the culprit of 4. Hope this gets merged soon; it literally doubled the training speed of my models.

metap:
Also, I had to manually normalize the input using the returned `mean` and `inv_var` to make the back pass work. Simply using `batch_norm_training` will lead to the correct `output` but the gradients aren't actually back propagated past the BN layer. Can anyone confirm this?

lukemetz:
1. yep, forgot to update doc string.

metap:
The gradients computed via `tf.gradients()` are actually correct as in the test file in OP, but they aren't being back propagated future down to earlier layers in actual networks.

metap:
Could `tensorflow/core/kernels/cudnn_batchnorm_op.cc` line 393 to 395 be related to 4? I don't see it being used after allocation

lukemetz:
That is certainly wrong / shouldn't be there. Good catch. It might fix the issue?

metap:
Yeah removing the three lines fixed 4. The gradient issue is still unfixed, unfortunately

zheng-xq:
@lukemetz, thank you very much for the code improvement! It is awesome to hear about the performance gain. 

There are a few comments at the bottom. Some of them are regarding the interface of this op. Because TensorFlow promises backward compatibility on TF core, it is difficult to change the op interface once the code is in TF core. 

On the other hand, we would like to merge in this code as soon as possible so more people can benefit and help improve. How about checking the same code into TF contrib for now? I think we would be able to merge with very little modifications. And after we iron out the details, the TF team will help integrate the change back into the TF core. 

Detailed comments: 

1. Lack of CPU implementation. 
2. We do not want a separate training and inference op. The same op should check whether "save_mean" and "saved_inv_var" are actually used before picking the "training" and "inference" version. 
3. We don't want stream-executor to allocate any memory. TensorFlow has a much more efficient memory management system. So this should be done through ScratchAllocator. 
4. Ongoing discussion whether to expose the running mean and average, and measure the performance impact. 
5. We need to test more common shapes to verify the results.

metap:
By the way the same API function `cudnnBatchNormalizationForwardTraining` uses running var in CuDNN 5004 but running `inv` var in CuDNN 4007.

lukemetz:
Thanks for the comments @zheng-xq. I will move the ops and kernels to contrib and make updates there. Correct me if I am wrong, but there is no stream executor equivalent in contrib?

Good catch @metap. I think I will just move to V5 then.

zheng-xq:
@lukemetz, unfortunately, there is no contrib equivalent of stream_executor. However, since your change is mostly addition, so that should be fine. 

Also there are a lot of people still using V4, so we should make sure both are functionally correct. TensorFlow in general makes sure at least two versions are supported.

vrv:
@lukemetz any updates?

sguada:
Any progress?

lukemetz:
I am very sorry I haven't had time to work on this. I am in the midst of finishing one job and moving across country to start a new one (at Google). As a result I have had very little free time to devote to this. I will try to find some time to make the updates soon.

lukemetz:
In the process of moving to contrib I ran into an issue with dynamic library loading that I am unsure how to fix. It appears my shared library is trying to register stream executor plugins that have already been registered. Is it something todo with the deps? (https://github.com/tensorflow/tensorflow/pull/1759/files#diff-246fc971a7ed6b0336891e4933f47651R15)
```
E tensorflow/stream_executor/cuda/cuda_blas.cc:2275] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
E tensorflow/stream_executor/cuda/cuda_dnn.cc:2128] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
F tensorflow/stream_executor/cuda/cuda_platform.cc:180] Check failed: ::perftools::gputools::port::Status::OK() == (MultiPlatformManager::RegisterPlatform(std::move(platform))) (OK vs. Internal: platform is already registered with name: "CUDA")
Aborted (core dumped)
```

Thanks for the help.

tensorflow/contrib/cudnn_batchnorm/BUILD
```diff
@@ -0,0 +1,69 @@
...
+  name = "cudnn_batchnorm_op",
+  out = "gen_cudnn_batchnorm.py",
+  require_shape_functions=False,
+  hidden=[
+    "BatchNormTrainingGrad"
+  ],
+  deps = ["//tensorflow/contrib/cudnn_batchnorm:cudnn_batchnorm_op_lib"]
+)
+
+tf_kernel_library(
```
vrv: My guess is this should just be a normal cc_library, not a tf_kernel_library, since this is just a 'package' that ncludes the underlying tf_kernel_library

lukemetz: Sorry for the delay. Good catch. Made this change but still does not appear to fix the duplicate plug-in registration problems.

vrv:
Not sure what the status of this is, but presumably it's almost ready now?  We don't get notifications when you push commits, unfortunately.

rmlarsen:
What is the status of this?

vrv:
Closing -- we're going to hopefully handle this internally -- it requires some fairly sophisticated changes to do this in a clean way, from what I've been told, since you want to do different things when you know the output during training is used or not.

alexgkendall:
What is the expected time frame for this to be integrated?

PR #1765, opened 2016-04-04
Update TF 101 Tutorial to match recent code

samjabrahams:
This patch updates two portions of the TensorFlow 101 tutorial:

The first change updates the Loss section to match the changes made in
5c5b29adda87f6fb, which switched from manually creating 1-hot labels to
using `tf.nn.sparse_softmax_cross_entropy_with_logits()`

The second removes a reference to the non-existent `get_data()`
function.


vrv:
Thanks!

Issue #313, opened 2015-11-20
Allow learning rates of specific variables to be scaled


alexatknit:
When repurposing a network, it can be useful to set layer-wise learning rates so that the final layer (which has random weights) does most of the learning initially. A useful pattern for applying learning rate scaling would be tf.stop_gradient(input, name=None), which can be seen as effectively scaling the learning rate by a factor of zero (obviously it also stops computation).

vrv:
You can use different optimizers for any subset of variables in your graph to get per-layer learning rates.

The 'minimize' function of an optimizer http://www.tensorflow.org/api_docs/python/train.html#optimizers allows you to specify which variables should be updated.  Let us know if this isn't what you need ...

alexatknit:
nope, its perfect, thanks

ascust:
hi guys, I would like to know more about this layer-wise learning rate since I am using this technique for fine-tuning. It seems by using stop_gradient, we can only be able to specify which layer should be prevented from updating. What if I want to give low level layers some low learning rate like 0.0001 and give high level layers high value like 0.01. How am I supposed to do this. Thank you

vrv:
This should really be posted on StackOverflow -- you'll get more community help there :)

PR #1780, opened 2016-04-05
refactor learning rate related code for the optimizers

raingo:
This is one step forward to solve #313. It's overshot to use multiple optimizers just for different learning rate.

Control the learning rate on the root class of the optimizers.


raingo:
I signed it!

martinwicke:
I'm conflicted -- the optimizer interface is too narrow as it is, and learning rate is not necessarily universal (although arguably it is for the optimizers we actually support). @rmlarsen, an opinion?

rmlarsen:
We might want  to implement other optimizers where the concept of "learning rate" does not make sense (such as second order methods), so I find moving it to the base class is not the right design.

raingo:
Gotta.

PR #1794, opened 2016-04-06
Attempt at adding support for cudnnr5 API only (no LSTM/RNN additions

vrv:
yet).

Tested only on R5 on a subset of tests so far.  Have not yet
tested that it doesn't break R4 -- Jenkins tests should cover
that.


vrv:
Jason from Stream Executor looked at this and gave his LGTM, so merging.

PR #1795, opened 2016-04-06
Add Cuda/Cudnn version request

martinwicke:
Helps with GPU issues.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -10,6 +10,8 @@ help and advice.
 ### Environment info
 Operating System:
 
+Installed version of CUDA and cuDNN (if any):
+
 If installed from binary pip package, provide:
 
 1. Which pip package you installed.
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -10,6 +10,8 @@ help and advice.
 ### Environment info
 Operating System:
 
+Installed version of CUDA and cuDNN (if any):
```
vrv: Would it be worth asking for something like

ls -l /path/to/cuda/lib/libcud* ?

(It helps to find out where the symlinks are pointing to, permissions, etc).

martinwicke:
Done.

ISSUE_TEMPLATE.md
```diff
@@ -10,7 +10,8 @@ help and advice.
 ### Environment info
 Operating System:
 
-Installed version of CUDA and cuDNN (if any):
+Installed version of CUDA and cuDNN: 
+(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):
```
vrv: probably want lib64, not lib.  Otherwise LGTM.

PR #1827, opened 2016-04-08
Upstream changes from past few days

vrv:


vrv:

vrv:
well, cmake is failing for unknown reasons, everything looks right.  Merging anyway since we're way behind on upstreaming, and everything else works.

Issue #1741, opened 2016-04-01
Polygamma and zeta function for tensorflow


tillahoffmann:
The polygamma function is necessary to evaluate the derivative of the digamma function which is already supported by tensorflow. As suggested by @ebrevdo in #291, I have ported the code for the two-parameter zeta function from cephes which allows us to evaluate the polygamma function. A pull request for eigen is [here](https://bitbucket.org/eigen/eigen/pull-requests/173/added-zeta-function-of-two-arguments-and). 

What is the best way to proceed integrating the new functionality into tensorflow? I'm happy to contribute the code with a bit of input regarding the right procedure. In particular, how do you update the eigen sources within the tensorflow repo?

ebrevdo:
Let's get that code through review on bitbucket, then we can update the references and I'll push my changes adding igamma/igammac to TensorFlow.  You'll be able to use those as reference for the TF implementations of polygamma/zeta.

tillahoffmann:
The zeta and polygamma functions are now in eigen. Happy to have a look at implementing them in tensorflow whenever you have time to update the references.

ebrevdo:
Great!  I need a couple of days to import your changes on our end.  In the meantime look at my recent push of igamma/igammac as a reference for wrapping zeta/polygamma.

ebrevdo:
(note the unit tests I added: they require scipy and cause a soft failure if it's not found.  this is fine - just make sure you have scipy installed when testing locally)

tillahoffmann:
That sounds good. I have implemented the wrappers locally. Just let me know when you've integrated the changes and I'll submit a PR.

benoitsteiner:
The Eigen implementations of the polygamma and zeta functions are now available in TensorFlow

akashgit:
@benoitsteiner is the build containing these functions already available ? thanks

tillahoffmann:
@akashgit, the eigen code in TensorFlow supports the functions but they won't be available from python until the PR above is merged. You can have a look there if you need them urgently.

PR #1834, opened 2016-04-09
Add polygamma and zeta function to tensorflow

tillahoffmann:
This PR adds the polygamma and zeta functions as discussed in #1741. It also implements the derivative of the digamma function. Tests are provided and can be run provided `scipy` is installed.


tillahoffmann:
I signed it!

tensorflow/core/ops/math_ops.cc
```diff
@@ -433,6 +433,37 @@ Note, above `Q(a, x)` (`Igammac`) is the upper regularized complete
 Gamma function.
 )doc");
 
+REGISTER_OP("Zeta")
+    .Input("x: T")
+    .Input("q: T")
+    .Output("z: T")
+    .Attr("T: {float, double}")
+    .Doc(R"doc(
+Compute the Hurwitz zeta function `\zeta(x, q)`.
```
ebrevdo: Strange code here?, \z.  Is this valid ascii?

tensorflow/core/ops/math_ops.cc
```diff
@@ -433,6 +433,37 @@ Note, above `Q(a, x)` (`Igammac`) is the upper regularized complete
...
+
+The Hurwitz zeta function is defined as:
+
+```
+\zeta(x, q) = \sum_{n=0}^{\infty} (q + n)^{-x}
```
ebrevdo: ditto here; and elsewhere.

ebrevdo:
Overall looks good, though I am worried about the integer first argument.  It is mainly an issue because of the proto serialization and working across multiple machines; it's not always gonna be the case that if i create a float 2.0, and then it undergoes proto serialization, then gets passed to the op on a different architecture, that floor(n) == n anymore.

Would you consider making a functor that adds a small epsilon (e.g. 0.01) , then floors the integer input, just before calling the Eigen tensor method?

gunan:

tensorflow/core/kernels/cwise_ops.h
```diff
@@ -512,6 +512,12 @@ template <typename T>
 struct igammac : base<T, Eigen::internal::scalar_igammac_op<T>> {};
 
 template <typename T>
+struct zeta : base<T, Eigen::internal::scalar_zeta_op<T>> {};
```
ebrevdo: I think this is where you'll need to extend the functors a bit.

tensorflow/core/ops/math_ops.cc
tillahoffmann: Just realised that you use \\( math \\) for inline math. Shall I fix all the other documentation characters in the process? E.g. see line [122](https://github.com/tillahoffmann/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc#L122) in the same file.

tillahoffmann:
What would be your opinion on demanding `abs(round(n) - n) < epsilon` in the eigen implementation or do you think this is TensorFlow specific because of the serialisation?

ebrevdo:
i think the abs(round(n) - n) < epsilon, where epsilon is e.g. 10*machine
epsilon is a good solution.

tillahoffmann:
I'm having second thoughts about the suggested implementation. I'm sure it works fine but I'm a bit worried about consistency: a number of functions in `SpecialFunctions.h` in eigen use equality tests. For example, `igamma` tests for equality with zero and I also test for equality with one in the implementation of the zeta function. Is there something special about the representation of one and zero in memory that we don't have to worry about serialisation issues? Would it be worthwhile adding a helper function `integer_equal_impl` for all such comparisons?

ebrevdo:
Perhaps I jumped the gun.  Looks like float and double can represent
integers up to and including 2^{mantissa bits + 1}.  So there seems to be
no need to worry in this case.  Even with the vagaries of
serializing/deserializing, I don't believe that a  number with exact FP
representation will be modified.

tillahoffmann:
Great, it seems we can just leave the implementations as they are. I have fixed the inline math. The problem with unescaped characters appears to be a [new feature for raw strings](https://solarianprogrammer.com/2011/10/16/cpp-11-raw-strings-literals-tutorial/) in C++11. E.g. the following will compile and print the expected string:

    const char* test = R"doc(This is the \zeta function.)doc";
    std::cout << test;

ebrevdo:
I think we're ready to import this, but I'll be away for a while.  Assigning to @benoitsteiner for final merge.

tillahoffmann:
Just wanted to follow up on this PR.

ebrevdo:

ebrevdo:
I was looking at the eigen implementation and saw that there may be a missing registration for the CUDA (float4 or double2) packet ops.  Can you revisit and make sure you added all the correct items?

ebrevdo:
@tensorflow-jenkins : test this please.

ebrevdo:
@martinwicke jenkins doesn't like me? :(

ebrevdo:
@tillahoffmann ignore my comment about Eigen.  I re-checked and it looks good.  Waiting on jenkins to review.

.idea/encodings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ebrevdo: please remove the .idea directory.

build.sh
```diff
@@ -0,0 +1,4 @@
+git submodule update --init
```
ebrevdo: please remove this

tensorflow/python/kernel_tests/.idea/misc.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ebrevdo: please remove all .idea directories

tensorflow/workspace.bzl~
```diff
@@ -0,0 +1,92 @@
+# TensorFlow external dependencies that can be loaded in WORKSPACE files.
```
ebrevdo: please remove all files ending with tilde

tillahoffmann:
Sorry, they sneaked past my .gitignore. Should be all cleaned up now.

tensorflow/python/ops/math_grad.py
```diff
@@ -295,8 +295,11 @@ def _LgammaGrad(op, grad):
 
 
 @ops.RegisterGradient("Digamma")
-def _DigammaGrad(op, grad):  # pylint: disable=unused-argument
-  raise NotImplementedError("grad(Digamma) == Polygamma(1) is not implemented")
+def _DigammaGrad(op, grad):
+  """Compute gradient of the digamma function with respect to its argument."""
+  x = op.inputs[0]
+  with ops.control_dependencies([grad.op]):
+    return grad * math_ops.polygamma(array_ops.ones(array_ops.shape(x), dtype=x.dtype), x)
```
ebrevdo: i *think* polygamma will distribute correctly if you just use constant(1, dtype=x.dtype)

tillahoffmann: Yes, `constant(1)` seems to work fine.

ebrevdo:
I'll try one running tests more time.  It's possible you have to remerge and fix any merge conflicts

ebrevdo:
@tensorflow-jenkins : test this please.

martinwicke:
Jenkins, test this please. And stop ignoring @ebrevdo.

PR #1837, opened 2016-04-10
Changed the dim computation in a compact form. It is also very easy t…

hunkim:
Changed the dim computation in a compact form. 

I think this is much easy to read and understand. I am not sure about the performance, but it's only a 3-element production, so it won't matter much.


hunkim:
Done!

hunkim:
I made a better patch.

PR #1838, opened 2016-04-10
Let reshape guess the dim.

hunkim:
Let 'tf.reshape' guess the dim and use it.


vrv:

(nice!)

PR #1847, opened 2016-04-11
Modified tests for gradients

tillahoffmann:
While implementing gradients for some new functions, I forgot to propagate the `grad` argument in the definition of the derivative in [`tensorflow/python/ops/math_grad.py`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py). The wrong implementation passed all the tests because the derivative of the function is evaluated in isolation. 

Multiplying the TensorFlow operation by a constant `a` ensures that such errors are caught by the tests.  This PR implements the modified tests with `a = 1.1`.


concretevitamin:
Jenkins, test this please.

caisq:

PR #1858, opened 2016-04-11
Fix "naming" anchor in "Adding an Op" How-To

samjabrahams:
The file had a braces-style declared anchor, {#naming}, which wasn't
being rendered or pre-processed properly. This changes it to an HTML `<a>`
tag. This should also fix the broken link pointing to this section on
line 874 in the same file.


martinwicke:
Thanks!

gunan:

martinwicke:
Will go  out with the next website push.

PR #1868, opened 2016-04-12
Fix conv2d with kernel 1x1 stride > 1

### LINKED ISSUE #889: Large Strides for 1x1 Convolutions
When a conv2d operator has a stride greater than the kernel size, the following error is thrown:
```ValueError: ('stride must be less than or equal to filter size', 'stride: [2x2] filter: [Dimension(1)xDimension(1)]')```
This makes implementing the 1x1 convolutions used to reduce spatial resolution in several papers (MSRA 2015 among others) awkward. Also, a convolution with stride greater than kernel size may be unusual but is it still well defined. Fixing this may be as simple as removing this assertion.
Thanks to everyone who developed TensorFlow, it's a fascinating tool.

### PR DESCRIPTION
**Pull request by ry**
Fixes #889 - seems the problem was an optimization to call directly into cublas when the kernel is 1x1. I suspect that was meant for 1x1 and stride 1, as it calculates the wrong thing with stride2. This adds 3 tests - two that demo the bug, one that I did while debugging but figured to submit too for extra coverage.

I signed the CLA

### DIFF
--- a/tensorflow/python/kernel_tests/conv_ops_test.py
+++ b/tensorflow/python/kernel_tests/conv_ops_test.py
@@ -319,6 +319,20 @@ class Conv2DTest(tf.test.TestCase):
                        strides=[1, 2], padding="VALID",
                        expected=expected_output)
 
+  def testConv2DKernelSmallerThanStrideValid(self):
+    expected_output = [65, 95, 275, 305]
+    self._VerifyValues(tensor_in_sizes=[1, 7, 7, 1],
+                       filter_in_sizes=[2, 2, 1, 1],
+                       strides=[3, 3], padding="VALID",
+                       expected=expected_output)
+
+  def testConv2DKernelSmallerThanStrideSame(self):
+    expected_output = [1, 3, 7, 9]
+    self._VerifyValues(tensor_in_sizes=[1, 3, 3, 1],
+                       filter_in_sizes=[1, 1, 1, 1],
+                       strides=[2, 2], padding="SAME",
+                       expected=expected_output)
+
   # Testing for backprops
   def _RunAndVerifyBackpropInput(self, input_sizes, filter_sizes, output_sizes,
                                  strides, padding, expected, data_format,
@@ -862,21 +876,6 @@ class Conv2DTest(tf.test.TestCase):
                                           shape=[21, 20, 3, 2]),
                     strides=[1, 1, 1, 1], padding="SAME")
 
-    # Stride larger than filter.
-    with self.assertRaisesRegexp(ValueError,
-                                 "stride must be less than or equal to filter"):
-      tf.nn.conv2d(tf.placeholder(tf.float32,
-                                          shape=[32, 20, 20, 3]),
-                    tf.placeholder(tf.float32,
-                                          shape=[4, 5, 3, 2]),
-                    strides=[1, 5, 5, 1], padding="SAME")
-    with self.assertRaisesRegexp(ValueError,
-                                 "stride must be less than or equal to filter"):
-      tf.nn.conv2d(tf.placeholder(tf.float32,
-                                          shape=[32, 20, 20, 3]),
-                    tf.placeholder(tf.float32,
-                                          shape=[5, 4, 3, 2]),
-                    strides=[1, 5, 5, 1], padding="SAME")
 
 
 # This is only a very simple test. More comprehensive tests live in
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: tensorflow/python/kernel_tests/conv_ops_test.py [hunk 1]
```diff
@@ -319,6 +319,20 @@ def testConv2D2x2FilterStride1x2(self):
...
                        expected=expected_output)
 
+  def testConv2D1x1FilterStride2(self):
+    expected_output = [1.0, 3.0, 7.0, 9.0]
+    self._VerifyValues(tensor_in_sizes=[1, 3, 3, 1],
+                       filter_in_sizes=[1, 1, 1, 1],
+                       strides=[2, 2], padding="VALID",
+                       expected=expected_output)
+
+  def testConv2D1x1FilterStride2Same(self):
```
vrv: This test isn't exactly testing anything different than SAME -- perhaps modify the example to exercise this?  Maybe try:

tensor_in_sizes=[1, 5, 5, 1]
filter_in_sizes=[1, 2, 2, 1]
strides=[2, 2]

So that the expected result is that you have some padding to exercise the difference between same and valid padding.

tensorflow/python/kernel_tests/conv_ops_test.py
```diff
@@ -319,6 +319,20 @@ def testConv2D2x2FilterStride1x2(self):
...
+    expected_output = [1.0, 3.0, 7.0, 9.0]
+    self._VerifyValues(tensor_in_sizes=[1, 3, 3, 1],
+                       filter_in_sizes=[1, 1, 1, 1],
+                       strides=[2, 2], padding="SAME",
+                       expected=expected_output)
+
```
vrv: Might also be nice to have a test with unequal strides > filter sizes, just to make sure the arguments are correctly propagated consistently.  Optional.

vrv:
Very nice, and good catch there!  Just a few small comments, and then we can test / merge.

tensorflow/python/kernel_tests/conv_ops_test.py
ry: Do you mean filter_in_sizes=[2, 2, 1, 1] ?

vrv: Oops, yeah.  Square filter.

ry: Ok - but that doesn't test stride > ksize ?

vrv: Grr, yes.  Ok, what about:

tensor_in_sizes=[1, 7, 7, 1]
filter_in_sizes=[2, 2, 1, 1]
strides = [3, 3]

If that doesn't work, forget it, and just add a TODO to add a more meaningful "SAME" test for stride > ksize.

gunan:

ry:
@vrv I updated the patch - that test seems to work.

vrv:
(@ry: tell me when I'm supposed to look at this again)

vrv:
alright then!  LGTM.  @tensorflow-jenkins: test this please

tensorflow/python/kernel_tests/conv_ops_test.py
ry: Oh forgot to do that... but maybe i'll pass on that for now if you don't mind. I'm curious to see if the build bots pass the tests, as my GPU/CUDA setup is a bit wacky.

vrv: That's fine, now that I think about it, this isn't really taking any new code paths and we should have coverage for the non-equal stride case already.

vrv:
Doh, pooling_ops_test also has a similar function for checking whether stride > ksize that you'd have to disable, though I'm now wondering whether pooling also supports stride > ksize the same way conv does.

ry:
oops- sorry! ok I'll look at this and get back to you

vrv:
If pooling doesn't natively support it, it's still probably worth getting this CL in, but will require a bit more work, in that you'd have to retain the stride > ksize checking for pooling (essentially forking the python shape code that's currently shared between conv2d and pooling ops).  Let's hope Eigen / Cudnn already support this:)

ry:
@vrv I squashed the changes into one commit. Pooling seems to work out of the box.

vrv:
This is great!  @tensorflow-jenkins: test this please

Issue #1799, opened 2016-04-06
One hot embedding?


vladfi1:
I'd like to embed integer labels into a fixed dimension space via the classic one-hot embedding. Here is an example:

```
onehot(inputs=[0, 2], num_labels=4) -> [[1, 0, 0, 0], [0, 0, 1, 0]]
```

It seems that sparse_to_dense does something similar, but not quite what I want. I've found some solutions online but they are rather convoluted. Is there a straightforward way to do this in tensorflow?

drgriffis:
`sparse_to_dense` will do what you want, you just have to index your data first.  For your example, you would want
```python
inputs, num_labels = [0,2], 4
indexedInputs = [[i, inputs[i]] for i in range(len(inputs))]
tf.sparse_to_dense(indexedInputs, [len(inputs),num_labels], 1)  # produces [[1,0,0,0], [0,0,1,0]]
```

vladfi1:
Is there any way to do this with variable-length inputs (shape=[None])?

drgriffis:
Sure, the `tf.range` function will give you an integer range in the length of your input node, then you can reshape them both as matrices and `tf.concat` them together to get the right data matrix that you can run through `tf.sparse_to_dense`.  The documentation for those functions can help you get things set up properly.

In future, [StackOverflow](https://stackoverflow.com/questions/tagged/tensorflow) is a much better place for a question like this.

vladfi1:
I did check StackOverflow but found only convoluted answers, and I wanted to ask if there wasn't a more idiomatic way to perform this fairly ubiquitous operation. Since that doesn't appear to exist, perhaps this could be a feature request for it?

I've also tried tf.range but have encountered some errors when trying to reshape it #1801. Even so, I don't see how this would allow for variable length inputs, since tf.range needs to take a concrete limit.

samjabrahams:
As a quick note, the [`tf.one_hot()`](https://www.tensorflow.org/versions/master/api_docs/python/array_ops.html#one_hot) method appears to be arriving in the next release of TensorFlow. I haven't checked, but it may also be available already in the nightly build- see the instructions to install it [here](https://github.com/tensorflow/tensorflow#installation)

vladfi1:
@samjabrahams That's great, thanks for the tip!

vladfi1:
Not a huge deal, but for some reason tf.one_hot insists on taking int64 as input - it should probably also work with the other int types. I'll just add in the cast myself for now.

[...]

PR #1869, opened 2016-04-12
Add QOL improvements for tf.one_hot()

samjabrahams:
This commit adds a Python wrapper for tf.one_hot(), and introduces two
main, backwards compatible, improvements, as mentioned in #1799:
1. The parameters `on_value` and `off_value` now have default values 1 and 0
   respectively, as 32-bit integers. Because the most common use case of
   one_hot() is to create Tensors for one-hot encoding, it will save users
   some headache to not have to manually type in these values by hand each
   time it is used. The parameters can still be given different values as
   before, and this does not any previous behavior.
2. `indices` and `depth` no longer have to be cast by hand as `int64` and
   `int32` values, respectively. Previously, one_hot() would complain if
   these values weren't of the exact right type- it now casts the values to
   the appropriate type on its own.

Pinging @ebrevdo for review.

### DIFF
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -1757,6 +1757,111 @@ def _DepthToSpaceShape(op):
       [input_shape[0], height, width, new_depth])]
 
 
+def one_hot(indices, depth, on_value=1, off_value=0,
+            axis=None, dtype=dtypes.float32, name=None):
+  """Returns a one-hot tensor.
+
+  The locations represented by indices in `indices` take value `on_value`,
+  while all other locations take value `off_value`. By default, `on_value` is 1,
+  and `off_value` is 0. The type of the output tensor is specified by `dtype`, 
+  which defaults to `tf.float32`.
+
+  If the input `indices` is rank `N`, the output will have rank `N+1`. The
+  new axis is created at dimension `axis` (default: the new axis is appended
+  at the end).
+
+  If `indices` is a scalar the output shape will be a vector of length `depth`
+
+  If `indices` is a vector of length `features`, the output shape will be:
+  ```
+    features x depth if axis == -1
+    depth x features if axis == 0
+  ```
+
+  If `indices` is a matrix (batch) with shape `[batch, features]`, the output
+  shape will be:
+  ```
+    batch x features x depth if axis == -1
+    batch x depth x features if axis == 1
+    depth x batch x features if axis == 0
+  ```
+
+
+  Examples
+  =========
+
+  Suppose that
+
+  ```
+    indices = [0, 2, -1, 1]
+    depth = 3
+    on_value = 5.0
+    off_value = 0.0
+    axis = -1
+  ```
+
+  Then output is `[4 x 3]`:
+
+  ```output =
+    [5.0 0.0 0.0]  // one_hot(0)
+    [0.0 0.0 5.0]  // one_hot(2)
+    [0.0 0.0 0.0]  // one_hot(-1)
+    [0.0 5.0 0.0]  // one_hot(1)
+  ```
+
+  Suppose that
+
+  ```
+    indices = [[0, 2], [1, -1]]
+    depth = 3
+    on_value = 1.0
+    off_value = 0.0
+    axis = -1
+  ```
+
+  Then output is `[2 x 2 x 3]`:
+
+  ```
+    output =
+    [
+      [1.0, 0.0, 0.0]  // one_hot(0)
+      [0.0, 0.0, 1.0]  // one_hot(2)
+    ][
+      [0.0, 1.0, 0.0]  // one_hot(1)
+      [0.0, 0.0, 0.0]  // one_hot(-1)
+    ]
+  ```
+
+  Args:
+    indices: A `Tensor` of indices.
+    depth: A scalar defining the depth of the one hot dimension.
+    on_value: A scalar defining the value to fill in output when `indices[j]
+      = i`. (default: 1)
+    off_value: A scalar defining the value to fill in output when `indices[j]
+      != i`. (default: 0)
+    axis: The axis to fill (default: -1, a new inner-most axis).
+    dtype: The data type of the output tensor.
+
+  Returns:
+    output: The one-hot tensor.
+
+  Raises:
+    TypeError: If dtype is `tf.string`
+  """
+  # Check for bad dtype specification
+  if dtype == dtypes.string:
+    raise TypeError("dtype must be a numeric type")
+
+  with ops.op_scope([indices, depth, on_value, off_value,
+            axis, dtype], name, "one_hot") as name:
+    on_value = ops.convert_to_tensor(on_value, dtype=dtype, name="on_value")
+    off_value = ops.convert_to_tensor(off_value, dtype=dtype, name="off_value")
+    indices = ops.convert_to_tensor(indices, dtype=dtypes.int64, name="indices")
+    depth = ops.convert_to_tensor(depth, dtype=dtypes.int32, name="depth")
+    return gen_array_ops._one_hot(indices, depth, on_value, 
+                                  off_value, axis, name)
+
+
 @ops.RegisterShape("OneHot")
 def _OneHotShape(op):
   """Shape function for the OneHot op.
--- a/tensorflow/python/BUILD
+++ b/tensorflow/python/BUILD
@@ -488,6 +488,7 @@ tf_gen_op_wrapper_py(
         "EditDistance",
         "MirrorPad",
         "MirrorPadGrad",
+        "OneHot",
         "Pack",
         "Pad",
         "Placeholder",
--- a/tensorflow/python/kernel_tests/one_hot_op_test.py
+++ b/tensorflow/python/kernel_tests/one_hot_op_test.py
@@ -24,20 +24,25 @@ import tensorflow as tf
 
 class OneHotTest(tf.test.TestCase):
 
-  def _testOneHot(self, truth, use_gpu=False, expected_err_re=None, **inputs):
+  def _testOneHot(self, truth, use_gpu=False, expected_err_re=None, 
+                  raises=None, **inputs):
     with self.test_session(use_gpu=use_gpu):
-      ans = tf.one_hot(**inputs)
-      if expected_err_re is None:
-        tf_ans = ans.eval()
-        self.assertAllClose(tf_ans, truth, atol=1e-10)
-        self.assertEqual(tf_ans.shape, ans.get_shape())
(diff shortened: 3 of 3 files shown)


samjabrahams:
@ebrevdo: pull request is here at #1869. Will work on your comments and update when it's ready to look at again.

gunan:

samjabrahams:
I've been working on this on and off for a few days- the one thing that I'm getting hung up on is the goal functionality for string data types. Correct me if I'm wrong, but I'm not sure the base one_hot operation supports strings. Running the current implementation of one_hot, I get the following error:

```
>>> import tensorflow as tf
>>> sess = tf.Session()
>>> a = tf.one_hot([1,0,2],3,"1","0")
>>> sess.run(a)
...
tensorflow.python.framework.errors.InvalidArgumentError: No OpKernel was registered to support Op 'OneHot' with these attrs
	 [[Node: OneHot_1 = OneHot[T=DT_STRING, axis=-1](OneHot_1/indices, OneHot_1/depth, OneHot_1/on_value, OneHot_1/off_value)]]
```

I'll update the code that I have for review once I finish writing a few tests.

samjabrahams:
Additionally, do you think it might be better to pass dtype along to convert_to_tensor instead of checking for a type error?

 `on_value = ops.convert_to_tensor(on_value, dtype=dtype)`

Since dtype is already given a default value of tf.float32, we're always going to have to do that type check, which ends up with unintuitive errors. For example:

`tf.one_hot([0,1,2], 3, on_value=1, off_value=-1)` 

throws a type error because it's expecting float values for on_value and off_value.

samjabrahams:
@ebrevdo See the latest changes- is this closer to what you're looking for?

ebrevdo:
Sorry, I'm away for a while.  Assigning @girving to finish the review.

girving:
Happy to finish the review, but can you squash the commits into one?

tensorflow/python/ops/array_ops.py
```diff
@@ -1608,6 +1608,138 @@ def _DepthToSpaceShape(op):
       [input_shape[0], height, width, new_depth])]
 
 
+def one_hot(indices, depth, on_value=None, off_value=None,
+            axis=None, dtype=dtypes.float32, name=None):
```
ebrevdo: on_value and off_value and dtype should be None by default.

if on_value and off_value are provided; then use their dtype.  otherwise if dtype is provided, use on_value 1 and off_value 0 for that dtype.  if dtype is tf.string, raise an error because it's not clear what the on_value should be.  if all 3 are provided, make sure the dtype of on_value and off_value match dtype.

update the Args: section below to reflect the changes.

update the Args: section below to reflect this.

samjabrahams:
Thanks @girving - I've made the changes suggested, and also made it possible for the user to specify only one of `on_value` or `off_value` if they choose to. Added more tests to go along with this as well.

girving:
Thanks!  Taking a look now.

tensorflow/python/ops/array_ops.py
```diff
@@ -1757,6 +1757,176 @@ def _DepthToSpaceShape(op):
       [input_shape[0], height, width, new_depth])]
 
 
+def one_hot(indices, depth, on_value=None, off_value=None,
+            axis=None, dtype=None, name=None):
```
girving: Can we say `on_value=1` and `off_value=0` rather than `None`?

tensorflow/python/ops/array_ops.py
```diff
@@ -1757,6 +1757,176 @@ def _DepthToSpaceShape(op):
...
+  ```
+    output =
+    [0.0 3.0 3.0 3.0]
+    [3.0 3.0 3.0 0.0]
+    [3.0 3.0 3.0 3.0]
+    [3.0 0.0 3.0 3.0]
+  //  ^                one_hot(0)
+  //      ^            one_hot(2)
+  //          ^        one_hot(-1)
+  //              ^    one_hot(1)
```
girving: I don't think this second example adds anything compared to the first one.

tensorflow/python/ops/array_ops.py
girving: Also `axis=-1`.

tensorflow/python/ops/array_ops.py
```diff
@@ -1757,6 +1757,176 @@ def _DepthToSpaceShape(op):
...
+    # Create off_value tensor 
+    if off_exists:
+      off_value = ops.convert_to_tensor(off_value, name="off_value")
+      if dtype_exists and off_value.dtype.base_dtype != dtype:
+        raise TypeError("The type of off_value must match dtype")
+    elif dtype_exists:
+      off_value = constant(0, dtype, name="off_value")
+    else:
+      off_value = constant(0, dtypes.float32, name="off_value")
+
```
girving: Here's some simplified logic for all the exists cases, once the defaults are fixed to be their actual values rather than `None`.

    if dtype is None:
      try:
        dtype = on_value.dtype
      except AttributeError:
        try:
          dtype = off_value.dtype
        except AttributeError:
          pass
    on_value = ops.convert_to_tensor(on_value, dtype=dtype, name="on_value")
    off_value = ops.convert_to_tensor(on_value, dtype=dtype, name="off_value")

tensorflow/python/ops/array_ops.py
girving: Oh, I see, the crazy case analysis structure is so that the defaults can be 1.0 and 0.0 but setting just one of them to an integer makes the whole thing an integer.  Gah.  Let me think to see if there's a better way to do this.

tensorflow/python/ops/array_ops.py
```diff
@@ -1757,6 +1757,176 @@ def _DepthToSpaceShape(op):
...
+      off_value = ops.convert_to_tensor(off_value, name="off_value")
+      if dtype_exists and off_value.dtype.base_dtype != dtype:
+        raise TypeError("The type of off_value must match dtype")
+    elif dtype_exists:
+      off_value = constant(0, dtype, name="off_value")
+    else:
+      off_value = constant(0, dtypes.float32, name="off_value")
+    
+    indices = constant(indices, dtype=dtypes.int64, name="indices")
+    depth = constant(depth, dtype=dtypes.int32, name="depth")
```
girving: Should be `ops.convert_to_tensor`, not `constant`.  `constant` breaks if `indices` is already a `Tensor`.

tensorflow/python/ops/array_ops.py
girving: Disgard the above code, I hadn't absorbed the full horror of the one but not both of `on_value` and `off_value` case.  It's not your fault, I'm just depressed about type conversion behavior in this language. :)

tensorflow/python/ops/array_ops.py
girving: Okay, how about the following compromise: make the default `dtype=tf.float32`.  Once we always have a dtype, the whole case analysis becomes the two lines

    on_value = ops.convert_to_tensor(on_value, dtype, name="on_value")
    off_value = ops.convert_to_tensor(off_value, dtype, name="off_value")

tensorflow/python/ops/array_ops.py
samjabrahams: I copied these examples from the original C++ documentation- I believe the second example is showcasing the difference between `axis = 0` and `axis = -1`. I'll still take it out if you think that's best.

girving: Ah, that's fine then.  A bit verbose for my taste, but preserving the existing documentation seems reasonable.

tensorflow/python/ops/array_ops.py
samjabrahams: I had a similar implementation when I first had a go at this wrapper. I think the reason @ebrevdo asked me to keep the inputs as `None` was to keep the behavior of `one_hot()` outputting a tensor of the same type as `on_value` and `off_value` (without having to re-specify it in `dtype`) . This led to all of the case checking below.

Personally, I think what you've suggested works well- it makes sense for `dtype` to act as a toggle switch rather than a type-checking device. It definitely makes the wrapper more readable too :D

samjabrahams:
@girving I made the changes! I'm going to hold off on squashing until it looks good to go- that way I can revert anything easily (if necessary).

girving:
Looks good!  Merging and I'll off the tests.  Unfortunately I have to do an errand soon, so there may be a delay.

samjabrahams:
Squashed- thank you, good sir!

girving:
Jenkins, test this please.

jtolds:
I installed the latest Tensorflow whl file through pip and attempted to use tf.one_hot. Now I'm getting 
```
tensorflow.python.framework.errors.InvalidArgumentError: No OpKernel was registered to support Op 'OneHot' with these attrs
	 [[Node: OneHot = OneHot[T=DT_BOOL, axis=-1](ArgMax, OneHot/depth, OneHot/on_value, OneHot/off_value)]]
```

Did I do something wrong?

samjabrahams:
Two questions:

1. When you say the latest TensorFlow wheel, do you mean `sudo pip install --upgrade https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.8.0-cp27-none-linux_x86_64.whl`, or did you download the nightly build?
2. What was the context for this error? i.e. what did you input to the function?

jtolds:
I ran `pip install --upgrade https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.8.0-cp27-none-linux_x86_64.whl` inside of a virtualenv and
```
import tensorflow as tf
sess = tf.Session()
a = tf.one_hot([1,0,2],3,"1","0")
sess.run(a)
```
reproduces it

samjabrahams:
Great, thanks! The quality of life changes aren't actually part of TensorFlow 0.8, so in order to test it out, you'll have to either build TensorFlow from source, or install it from the nightly binaries [available here](https://github.com/tensorflow/tensorflow#installation)

The other issue is that, unfortunately, string values aren't working properly in `one_hot` at this time. Try running this command instead and see if it works:

```
a = tf.one_hot([1,0,2],3,1,0)
```

There are a few bugs/headaches that need to be addressed as well- hopefully by the next release it'll be working super smoothly!

jtolds:
Oh! Turns out, using integers totally fixes it. My initial use case before I ended up reducing down to the test case I posted used booleans for the on/off values. That also doesn't work, but that's totally fine, since I can get the same behavior with integers and casts.
Great, thanks so much!

samjabrahams:
No problem! Hopefully by the next release you won't have to fiddle around with casts for non-numeric values. 

Glad it's working for you!

PR #1895, opened 2016-04-12
Fix bugs in the indexing operator with latest NumPy

mrry:
`np.random.random_integers()` is deprecated and recently started returning non-`int` values. This PR fixes this so that the `tf.Tensor` indexing operator can handle non-`int` (but convertable-to-`int`) values, and removes the deprecated method from the test.


mrry:

PR #1901, opened 2016-04-13
Marge internal changes

martinwicke:


martinwicke:
This contains an image which should not go into github.

PR #1903, opened 2016-04-13
update more places with version to 0.8

jendap:


jendap:
@martinwicke have we already generated new documentation?

martinwicke:
Working on that

PR #1854, opened 2016-04-11
fix ppa repositories in ci_build

jendap:
Those repositories are just workaround for old versions of libraries on old ubuntu (trusty). Those ppas do not make sense on newer ubuntu or debian.

PR #1904, opened 2016-04-13
fix ppa repositories in ci_build

jendap:
fix debian build


jendap:
@martinwicke this replace #1854. It is done against r0.8 now so that we can build debian there as well.

(it should work now - I put there the && instead of multiple RUN statements - it creates less layers)

martinwicke:
:+1:

PR #1906, opened 2016-04-13
add a reference on log-likelihood

fayeshine:
add a reference on log-likelihood in word2vec tutorial


gunan:

PR #1910, opened 2016-04-13
switch docker links from b.gcr.io to gcr.io

jendap:


jendap:
This should go to r0.8, opening new PR...

PR #1911, opened 2016-04-13
switch docker links from b.gcr.io to gcr.io

jendap:


jendap:
Do we want to switch the "latest" tags to point to release candidate versions? I think it is better pointing to old stable release. But that means we should during the release candidate period edit the documentation to contain explicit tags - i.e. gcr.io/tensorflow/tensorflow:0.8.0rc0 instead of gcr.io/tensorflow/tensorflow. Then for final release we should edit the links again. Too complicated?

Anyway, this time we can switch "latest" to the RC because people reading old documentation will se b.gcr.io where the latest will still point to 0.7.1 ;-)

jendap:
@vincentvanhoucke you should probably also upload the next version of udacity container to gcr.io and update the readme accordingly.

Svikashini:
how to know which version of tensorflow been installed

martinwicke:
```python
import tensorflow as tf
tf.__version__
```

Svikashini:
while importing sklearn i got an error "No module named sklearn". How to install it

PR #1912, opened 2016-04-13
add a reference for logistic regression

fayeshine:
add a reference for logistic regression in word2vec tutorial


gunan:

PR #1926, opened 2016-04-13
Fix TensorBoard lib/css dependency

teamdandelion:


martinwicke:
Can you make another one against r0.8?

kchen92:
How long would it take for this fix to reflect in the pip install? Thanks.

martinwicke:
Should happen tomorrow.

PR #1957, opened 2016-04-15
add a reference for dropout in mnist tutorial

fayeshine:
add a reference for dropout in mnist tutorial


gunan:

PR #1958, opened 2016-04-15
add a reference for `building from source` and `bazel`

fayeshine:
add a reference for `building from source` and `bazel` in RNN tutorial


gunan:

PR #1963, opened 2016-04-15
add a wiki reference for Mandelbrot set

fayeshine:
add a wiki reference for Mandelbrot set


gunan:

PR #1964, opened 2016-04-15
add a wiki reference for partial differential equation

fayeshine:
add a wiki reference for partial differential equation


gunan:

vrv:
btw, feel free to batch all of these changes into one commit.

PR #1977, opened 2016-04-16
Revert change to r0.8 branch that points to new URLs

vrv:
They haven't been uploaded yet, and so our website is pointing people at the wrong location.


martinwicke:
Did we ever actually deploy the website with those changes? Better to be safe, for sure. Possibly in the future, we should point the website at a tag again (possibly the latest tag on each branch, we can find those easily).

vrv:
(We did, so I rolled it back and re-deployed the website).

martinwicke:
Thanks. :(

PR #1979, opened 2016-04-16
Fix tb r0.8

dongs0104:


vrv:
These are already in master  (#1926)

PR #1987, opened 2016-04-16
Stream Executor: perform a cudnn library vs. binary check, add relu6/x support.

vrv:
During initialization, we check whether the loaded version
of the cudnn library matches the version of the library that
the binary was compiled against.  When there is a mismatch,
throw an error.

Example: I built with cudnnv5 but I am loading V4, and I get:

E tensorflow/stream_executor/cuda/cuda_dnn.cc:298] Loaded cudnn library: 4007 but source was compiled against 5004.  If using a binary install, upgrade your cudnn library to match.  If building from sources, make sure the library loaded matches the version you specified during compile configuration.

I then copy the v5 version of libcudnn.so into /usr/local/cuda/lib64
and when I re-run the program, the error goes away.

Thoughts, @zheng-xq, @leary-google, @zffchen78 ?

### DIFF
--- a/tensorflow/stream_executor/cuda/cuda_dnn.cc
+++ b/tensorflow/stream_executor/cuda/cuda_dnn.cc
@@ -283,6 +283,22 @@ port::Status CudnnSupport::Init() {
   auto status = dynload::cudnnCreate(
       parent_, reinterpret_cast<cudnnHandle_t*>(&dnn_handle_));
   if (status == CUDNN_STATUS_SUCCESS) {
+    // Check whether loaded version of CuDNN matches what the source
+    // was built with.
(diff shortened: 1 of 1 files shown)


martinwicke:
+1. Should I merge this?

martinwicke:
However, aren't cuDNN backwards compatible, at least to some degree? Should we check >=?

vrv:
Not really. Even when the api is the same it sometimes doesn't work

zheng-xq:
Ideally, we only want to check the major versions, since the minor versions
and patch levels do not affect backward compatibility. That is true for all
official binaries starting from RC.

However, given we have a number of EA binaries floating around, this is not
necessarily true any more. So the current check is a bit too strict, but
fine in practice.

PR #1991, opened 2016-04-17
Fix SAME padding calc when stride>ksize

ry:
Adds a failing test case for conv2d ksize=1, stride=2, padding=same, input=4x4

Adds tests convering that case for pooling ops too.

It seems cudnn doesn't like `ksize<stride` when `ksize>1`, thus the commented out test.

@vrv

### DIFF
--- a/tensorflow/python/kernel_tests/conv_ops_test.py
+++ b/tensorflow/python/kernel_tests/conv_ops_test.py
@@ -327,11 +327,26 @@ class Conv2DTest(tf.test.TestCase):
                        expected=expected_output)
 
   def testConv2DKernelSmallerThanStrideSame(self):
-    expected_output = [1, 3, 7, 9]
     self._VerifyValues(tensor_in_sizes=[1, 3, 3, 1],
                        filter_in_sizes=[1, 1, 1, 1],
                        strides=[2, 2], padding="SAME",
-                       expected=expected_output)
+                       expected=[1, 3, 7, 9])
+
+    self._VerifyValues(tensor_in_sizes=[1, 4, 4, 1],
+                       filter_in_sizes=[1, 1, 1, 1],
+                       strides=[2, 2], padding="SAME",
+                       expected=[1, 3, 9, 11])
+
+    self._VerifyValues(tensor_in_sizes=[1, 4, 4, 1],
+                       filter_in_sizes=[2, 2, 1, 1],
+                       strides=[3, 3], padding="SAME",
+                       expected=[44, 28, 41, 16])
+
+    # TODO this currently fails.
+    #self._VerifyValues(tensor_in_sizes=[1, 8, 8, 1],
+    #                   filter_in_sizes=[2, 2, 1, 1],
+    #                   strides=[4, 4], padding="SAME",
+    #                   expected=[72, 112, 392, 432])
 
   # Testing for backprops
   def _RunAndVerifyBackpropInput(self, input_sizes, filter_sizes, output_sizes,
--- a/tensorflow/core/kernels/ops_util.cc
+++ b/tensorflow/core/kernels/ops_util.cc
@@ -54,16 +54,14 @@ Status Get2dOutputSizeVerbose(const int in_height, const int in_width,
       *new_width = ceil(in_width / static_cast<float>(col_stride));
       // Calculate padding for top/bottom/left/right, spilling any excess
       // padding to bottom and right.
-      const int pad_needed_height =
-          (*new_height - 1) * row_stride + filter_height - in_height;
+      const int pad_needed_height = std::max(0,
+          (*new_height - 1) * row_stride + filter_height - in_height);
       *pad_top = pad_needed_height / 2;
-      CHECK_GE(pad_needed_height, 0);
       *pad_bottom = pad_needed_height - *pad_top;
 
-      const int pad_needed_width =
-          (*new_width - 1) * col_stride + filter_width - in_width;
+      const int pad_needed_width = std::max(0,
+          (*new_width - 1) * col_stride + filter_width - in_width);
       *pad_left = pad_needed_width / 2;
-      CHECK_GE(pad_needed_width, 0);
       *pad_right = pad_needed_width - *pad_left;
       break;
   }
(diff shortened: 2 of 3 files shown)


tensorflow/python/kernel_tests/conv_ops_test.py
```diff
@@ -327,11 +327,28 @@ def testConv2DKernelSmallerThanStrideValid(self):
                        expected=expected_output)
 
   def testConv2DKernelSmallerThanStrideSame(self):
-    expected_output = [1, 3, 7, 9]
+    #[[1 4 7 0]
```
vrv: It's weird that the padding calculation computes a padding of 1 here -- without your changes to ops_util.cc, would all of these new tests still pass?

tensorflow/python/kernel_tests/conv_ops_test.py
```diff
@@ -327,11 +327,28 @@ def testConv2DKernelSmallerThanStrideValid(self):
...
+    #  [2  6 10 14]
+    #  [3  7 11 15]
+    #  [4  8 12 16]]
+    self._VerifyValues(tensor_in_sizes=[1, 4, 4, 1],
+                       filter_in_sizes=[1, 1, 1, 1],
+                       strides=[2, 2], padding="SAME",
+                       expected=[1, 3, 9, 11])
+
+    # TODO this currently fails.
+    #self._VerifyValues(tensor_in_sizes=[1, 8, 8, 1],
```
vrv: We should check what other frameworks do for this case.  I could imagine padding by 1 on both sides so:

[ 1 2 3 4 5 6 7 8 ]
[ 1 2 3 4 5 6 7 8 ]   
...

Your patches for the first row would be:
 [0 0    [0 0   [0 0
  0 1]    4 5]    8 0]

So the output size could be 3x3, rather than 2x2.

tensorflow/python/kernel_tests/conv_ops_test.py
ry: All pass but the 4x4 stride 2 one, which fails and the patch fixes.

tensorflow/python/kernel_tests/conv_ops_test.py
ry: sorry your formatting didn't turn out and not sure what you mean - can you revise the comment?

vrv: Perhaps ignore that comment.  I'm just thinking about what's the right thing to do here, and I think instead of the change to use std::max(filter_size, stride_size), I think it might just be better to do real_padding = std::max(0, calculated_padding).

Why: the equations I usually go back to, from @Yangqing, master of understanding padding, is (for SAME):

R' = ceil(float(R) / float(S))
C' = ceil(float(C) / float(S))
Pr = ((R' - 1) * S + K - R) / 2
Pc = ((C' - 1) * S + K - C) / 2

In one case above (R=C=3, S=2, K=1), we have:  R' = C' = 2, Pr = Pc = 1 * 2 + 1 - 3 = 0, so it feels like the padding calculation is wrong for that case, since it's now calculated as '1' even though it's unneeded.

In the case that failed for you: R=C=4, S=2, K=1 we have R'=C'=2,  Pr = Pc = 1 * 2 + 1 - 4 = -1, so we have negative padding.  My guess is that you changed the code to avoid tripping the check failure -- if you use std::max(0, calculated_padding), I'm thinking the math will generally work out.  What do you think?

ry:
> In one case above (R=C=3, S=2, K=1), we have: R' = C' = 2, Pr = Pc = 1 * 2 + 1 - 3 = 0, so it feels like the padding calculation is wrong for that case, since it's now calculated as '1' even though it's unneeded.
>
> In the case that failed for you: R=C=4, S=2, K=1 we have R'=C'=2, Pr = Pc = 1 * 2 + 1 - 4 = -1, so we have negative padding. My guess is that you changed the code to avoid tripping the check failure -- if you use std::max(0, calculated_padding), I'm thinking the math will generally work out. What do you think?

Let's consider another case (padding=SAME) where the kernel is bigger than 1x1 and the strides are bigger than the kernel, such that the kernel lays across the boundary of the image.

So take R=C=4, K=2, S=3. R'=C'=2 and 
```
pad_needed_width = (R' - 1)S + K - R = 1*3 + 2 - 4 = 1
pad_left =  pad_needed_width / 2 = 0
pad_right = pad_needed_width - pad_left  = 1
```

That seems okay, it calculates some 2x2 output matrix. In the case of the tests we'd have `expected == [44, 28, 41, 16]`.

But in Theano with `border_mode='full'` it pads with one on all sides and calculates `expected == [4, 25, 70, 144]`.

Using this patch, TF would calculate:
```
pad_needed_width = (R' - 1)S + max(S,K) - R = 1*3 + 3 - 4 = 2
pad_left = pad_needed_width / 2 = 1
pad_right = pad_needed_width - pad_left  = 1
```

See the second test here: https://gist.github.com/ry/273e0a8b83e33f891afbe09a6130c72c

vrv:
@zheng-xq: thoughts?  I know we don't do padding exactly the same as other frameworks, so I'm not entirely sure what to do here.

zheng-xq:
To me, the right thing to do seems to be:

// keep pad_needed_height the unchanged.  
      int pad_needed_height =
          (*new_height - 1) * row_stride + filter_height - in_height;
// if the padding is negative, truncate it to zero. 
      if (pad_needed_height < 0) {
        pad_needed_height = 0;
      }
// The same goes to "pad_needed_width"

zheng-xq:
I've read the example more carefully, and I still like a clipping-based approach. 

If R=C=5, K=2, S=3. R'=C'=2
The clipping method:
pad_needed_width = (R' - 1)S + K - R = 1*3 + 2 - 5 = 0

The proposed method:
pad_needed_width = (R' - 1)S + max(S,K) - R = 1*3 + 3 - 5 = 1

I would say the clipping method is better. 

The weird thing TF does with R=C=4 can also be attributed to the fact TF prefers adding an extra padding on the right. If it used the other way around, a few corner cases would be slightly easier to handle and matches Theano in that case. But at this point, too many models have already been trained this way. And this benefit seems small to make a global switch.

ry:
I'm okay with that. I think the difference is minimal. I made a graphic showing the two cases with  R=C=4, K=2, S=3. The colored areas show the kernel. The first is the algorithm i proposed (and what theano uses, I think) and the other is the "clipping"

padding

ry:
I've updated the patch with the clipping method, and another test.

vrv:
Awesome, thanks for helping us work through this!.  @tensorflow-jenkins: test this please

PR #1994, opened 2016-04-17
(Do not merge - developing) Add tests for user ops

caisq:
as a part of test-on-install in pip.sh


caisq:
CL submitted internally for this. Will be synced to external shortly. This PR will be closed without merging.

Issue #1606, opened 2016-03-23
Confusion Matrix


Muaazbinsaeed:
Is there any operation which generates Confusion Matrix in Tensorflow?
Can't find it on http://stackoverflow.com/questions/tagged/tensorflow

fayeshine:
You can do it in this way, maybe a little tedious

res = tf.pack([tf.argmax(y,1), tf.argmax(y_,1)])
ans = res.eval(feed_dict={x: mnist.test.images, y_: mnist.test.labels})
import numpy as np
import pandas as pd
confusion = np.zeros([10,10],int)
for p in ans.T:
    confusion[p[0],p[1]]+=1
print(pd.DataFrame(confusion))

Muaazbinsaeed:
@fayeshine thanks
There could be many ways to implement it using other libraries 
but that's not the way tensorflow is.
Tensorflow should make this operation too.

girving:
Fixed by #1999.

syedk01:
I am hoping to print the confusion matrix in my code
In the line where it says 
Tgtc=11
How do i add something like 
Tgtc=[2,5,7,8,11] where i want to do a confusion matrix for the 5 variables

>>>>>>>>>>>>
import sys

if sys.version_info.major != 3:
    raise Exception('Please run with python3')

import os
import time
import gc
import numpy as np

import afib_common

from afib_common import (
    build_model, get_batch, train_test_split, get_batch2)
from afib_data import reload_settings

# Load constants defined in afib_data
reload_settings()
if not os.path.isdir('modelsave'):
    os.path.mkdir('modelsave')
model = build_model()

def excld(N, i):
    ''' Generates a list of numbers 0..N-1 without the number i'''
    return tuple(_ for _ in range(N) if _ != i)

# Target class for performance metric
tgtC = 11
# (normal = "N", is the 11th item in afib_common.symbols_keep)

def calc_confusion_matrix(test_y, test_yhat):
    ''' Generates a confusion matrix from actual classes and predicted classes.

    Columns are the predicted class, while rows are the actual.
    '''
    N = test_y.shape[1]
    cm = np.zeros((N, N))
    cls_y = test_y.argmax(1)
    cls_yhat = test_yhat.argmax(1)
    for actual, predicted in zip(cls_y, cls_yhat):
        cm[actual, predicted] += 1
    return cm / cm.sum()

def estimate_f1_score():
    ''' Estimate the F1 score using many samples.

[...]

PR #1999, opened 2016-04-17
Add confusion matrix to tensorflow

lucasmoura:
This MR add a confusion matrix implementation for tensorflow. The function is named confusion_matrix and is a metric operation.

The function receives three parameters, both predictions and targets labels and an optional num_classes:

``` python
tf.contrib.metrics.confusion_matrix(predictions, targets, num_classes=None)
```

If num_classes is not provided, it will be computed using both predictions and targets parameters.

Both predictions and targets must be an 1-D array with dtype int32 or int64. This function will then infer the valid labels from both arrays and create a confusion matrix for them. The confusion matrix will always be a square matrix with shape L X L, considering L as the number of valid labels for a given classification task. The dtype of the confusion matrix will be the same as the predictions and target ones.

For example, given a prediction array: [1, 2, 3, 1], and a target array: [1, 2, 1, 1], the generated confusion matrix would be:

``` python
[[0 0 0],
 [2 0 0],
 [0 1 0],
 [1 0 0]]
```

The rows of the matrix represent the predictions array and the columns, the targets one.
Finally, this function was implemented only for CPU use. This MR resolves the issue #1606, but there is still work to be done on tensorboard considering this function.

### DIFF
--- a/tensorflow/contrib/metrics/python/kernel_tests/confusion_matrix_ops_test.py
+++ b/tensorflow/contrib/metrics/python/kernel_tests/confusion_matrix_ops_test.py
@@ -1,0 +1,127 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for confusion_matrix_ops."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+
+
+class ConfusionMatrixTest(tf.test.TestCase):
+
+  def _testConfMatrix(self, predictions, targets, truth):
+    with self.test_session():
+      ans = tf.contrib.metrics.confusion_matrix(predictions, targets)
+      tf_ans = ans.eval()
+      self.assertAllClose(tf_ans, truth, atol=1e-10)
+
+  def _testBasic(self, dtype):
+    predictions = np.arange(5, dtype=dtype)
+    targets = np.arange(5, dtype=dtype)
+
+    truth = np.asarray(
+      [[1, 0, 0, 0, 0],
+       [0, 1, 0, 0, 0],
+       [0, 0, 1, 0, 0],
+       [0, 0, 0, 1, 0],
+       [0, 0, 0, 0, 1]],
+      dtype=dtype)
+
+    self._testConfMatrix(
+      predictions=predictions,
+      targets=targets,
+      truth=truth)
+
+  def testInt32Basic(self, dtype=np.int32):
+    self._testBasic(dtype)
+
+  def testInt64Basic(self, dtype=np.int64):
+    self._testBasic(dtype)
+
+  def _testDiffentLabelsInPredictionAndTarget(self, dtype):
+    predictions = np.asarray([1, 2, 3], dtype=dtype)
+    targets = np.asarray([4, 5, 6], dtype=dtype)
+
+    truth = np.asarray(
+      [[0, 0, 0, 0, 0, 0, 0],
+       [0, 0, 0, 0, 1, 0, 0],
+       [0, 0, 0, 0, 0, 1, 0],
+       [0, 0, 0, 0, 0, 0, 1],
+       [0, 0, 0, 0, 0, 0, 0],
+       [0, 0, 0, 0, 0, 0, 0],
+       [0, 0, 0, 0, 0, 0, 0]],
+      dtype=dtype)
+
+    self._testConfMatrix(
+      predictions=predictions,
+      targets=targets,
+      truth=truth)
+
+  def testInt32DifferentLabels(self, dtype=np.int32):
+    self._testDiffentLabelsInPredictionAndTarget(dtype)
+
+  def testInt64DifferentLabels(self, dtype=np.int64):
+    self._testDiffentLabelsInPredictionAndTarget(dtype)
+
+  def _testMultipleLabels(self, dtype):
+    predictions = np.asarray([1, 1, 2, 3, 5, 6, 1, 2, 3, 4], dtype=dtype)
+    targets = np.asarray([1, 1, 2, 3, 5, 1, 3, 6, 3, 1], dtype=dtype)
+
+    truth = np.asarray(
+      [[0, 0, 0, 0, 0, 0, 0],
+       [0, 2, 0, 1, 0, 0, 0],
+       [0, 0, 1, 0, 0, 0, 1],
+       [0, 0, 0, 2, 0, 0, 0],
+       [0, 1, 0, 0, 0, 0, 0],
+       [0, 0, 0, 0, 0, 1, 0],
+       [0, 1, 0, 0, 0, 0, 0]],
+      dtype=dtype)
+
+    self._testConfMatrix(
+      predictions=predictions,
+      targets=targets,
+      truth=truth)
+
+  def testInt32MultipleLabels(self, dtype=np.int32):
+    self._testMultipleLabels(dtype)
+
+  def testInt64MultipleLabels(self, dtype=np.int64):
+    self._testMultipleLabels(dtype)
+
+  def testInvalidRank(self):
+    predictions = np.asarray([[1, 2, 3]])
+    targets = np.asarray([1, 2, 3])
+    self.assertRaisesRegexp(
+        ValueError, "are not compatible",
+        tf.contrib.metrics.confusion_matrix, predictions, targets)
+
+    predictions = np.asarray([1, 2, 3])
+    targets = np.asarray([[1, 2, 3]])
+    self.assertRaisesRegexp(
+        ValueError, "are not compatible",
+        tf.contrib.metrics.confusion_matrix, predictions, targets)
+
+  def testInputDifferentSize(self):
+    predictions = np.asarray([1, 2, 3])
+    targets = np.asarray([1, 2])
+    self.assertRaisesRegexp(
+          ValueError,
+          "are not compatible",
+          tf.contrib.metrics.confusion_matrix, predictions, targets)
+
+if __name__ == '__main__':
+  tf.test.main()
--- a/tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
+++ b/tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
@@ -1,0 +1,86 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from tensorflow.python.framework import ops
+from tensorflow.python.framework import dtypes
+from tensorflow.python.ops import array_ops
+from tensorflow.python.ops import math_ops
+from tensorflow.python.ops import sparse_ops
+
+
+"""Confusion matrix related metrics."""
+
+
+def confusion_matrix(predictions, targets, num_classes=None, name=None):
+  """Computes the confusion matrix from predictions and targets
+
+  Calculate the Confusion Matrix for a pair of prediction and
+  target 1-D int arrays.
+
+  Considering a prediction array such as: `[1, 2, 3]`
+  And a target array such as: `[2, 2, 3]`
+
+  The confusion matrix returned would be the following one:
+      [[0, 0, 0]
+       [0, 1, 0]
+       [0, 1, 0]
+       [0, 0, 1]]
+
+  Where the matrix rows represent the prediction labels and the columns
+  represents the target labels. The confusion matrix is always a 2-D array
+  of shape [n, n], where n is the number of valid labels for a given
+  classification task. Both prediction and target must be 1-D arrays of
+  the same shape in order for this function to work.
+
+  Args:
+    predictions: A 1-D array represeting the predictions for a given
+                 classification.
+    targets: A 1-D represeting the real labels for the classification task.
+    num_classes: The possible number of labels the classification task can
+                 have. If this value is not provided, it will be calculated
+                 using both predictions and targets array.
+
+  Returns:
+    A l X l matrix represeting the confusion matrix, where l in the number of
+    possible labels in the classification task.
+
+  Raises:
+    ValueError: If both predictions and targets are not 1-D vectors and do not
+                have the same size.
+  """
+  with ops.op_scope([predictions, targets, num_classes], name,
+                    'confusion_matrix') as name:
+    predictions = ops.convert_to_tensor(
+      predictions, name="predictions", dtype=dtypes.int64)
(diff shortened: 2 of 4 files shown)


srjoglekar246:
+1 to this. Nice, simple API, similar to the scipy version.

Marty-the-Red:
FUCK OFF

kashif:
@lucasmoura you also have to run:

```
bazel run tensorflow/core/ops/compat:update_ops
```

to update the `ops_history` file I think

vrv:
@kashif: no longer, we update the ops docs ourselves now.

kashif:
@vrv ah cool good to know! thanks

girving:
@concretevitamin: Are you working on a out-of-place version of scatter?  I'd prefer that confusion matrix was implemented as a Python wrapper on top of something more fundamental like that.

concretevitamin:
@girving: not sure if these are exactly what you want, but we have a [ScatterNdAdd functor](https://github.com/tensorflow/tensorflow/blob/cc9bfbf8ef4a3dea6514ad939d238f7442188247/tensorflow/core/kernels/sparse_tensor_dense_add_op.h) and a [SparseTensorDenseAddOp](https://github.com/tensorflow/tensorflow/blob/cc9bfbf8ef4a3dea6514ad939d238f7442188247/tensorflow/core/ops/sparse_ops.cc#L393) now. Will they work for this purpose?

girving:
The functor looks good; `SparseTensorDenseAdd` isn't quite the right since you have to pass zeros but otherwise works.  For now, let's implement the confusion matrix on top of `SparseTensorDenseAdd` in Python, maybe with a TODO to use something better if it comes around.

lucasmoura:
Okay, I will change my implementation to be built on top of SparseTensorDenseAdd

lucasmoura:
@girving Should I close this merge request ? Since confusion matrix will be just a python wrapper for an existing tensorflow op, the code I have written will no longer be useful, right ?

girving:
Would you be interested in modifying your PR instead?  I think it's just a few lines of Python.  I would call it `confusion_matrix`, but otherwise the interface is good.

lucasmoura:
@girving Wait, I think I misunderstood what I should do. I originally thought that I should reimplement the confusion matrix using only a python wrapper for SparseTensorDenseAdd, but should I left the c++ implementation and add this new python code as well ?

girving:
@lucasmoura: Your original interpretation was correct: it's better to remove the C++ implementation and just have the few lines of Python.

lucasmoura:
@girving Okay, I will start working on it today

lucasmoura:
@girving Done

tensorflow/python/ops/nn_ops.py
```diff
@@ -42,6 +43,96 @@
 local_response_normalization = gen_nn_ops.lrn
 
 
+def confusion_matrix(predictions, targets, name=None):
+  """'
+  Calculate the Confusion Matrix for a pair of prediction and target 1-D int arrays.
```
girving: Put on one line:

"""Computes the confusion matrix from predictions and targets.

tensorflow/python/ops/nn_ops.py
```diff
@@ -42,6 +43,96 @@
...
+
+  Considering a prediction array such as: [1, 2, 3]
+  And a target array such as: [2, 2, 3]
```
girving: Backticks around code bits: `[1, 2, 3]`.

tensorflow/python/ops/nn_ops.py
```diff
@@ -42,6 +43,96 @@
...
+
+  The confusion matrix returned would be the following one:
+  [[0, 1, 0]
+   [0, 1, 0]
+   [0, 0, 1]]
```
girving: Indent the matrix by 4 spaces to format it as code.

tensorflow/python/ops/nn_ops.py
```diff
@@ -42,6 +43,96 @@
...
+
+  Where the matrix rows represent the prediction labels and the columns represents
+  the target labels. The confusion matrix is always a 2-D array of shape l X l, where
```
girving: of shape `[n, n]`, where `n` is the number ...

tensorflow/python/ops/nn_ops.py
```diff
@@ -42,6 +43,96 @@
...
+        position = value * num_labels + target
+
+        values[position] += 1
+
+    indexes = [ [i, j] for i in range(num_labels) for j in range(num_labels)
+                if values[i * num_labels + j] != 0]
+    values = [ value for value in values if value != 0]
+    shape = (num_labels, num_labels)
+    dense_matrix = np.zeros(shape=(num_labels, num_labels))
+
```
girving: I have to run off now, but I think you should be able to replace most of the above looping code with one line.  Will comment more later.

lucasmoura: Okay, I will try to find a better way to compute these variables. For the other problems, I have already fixed them.

lucasmoura: I have managed to create both indices and values array with two lines of code. I don't know if my approach is the way you originally thought about in order to reduce the lines of code, but if it is not, please let me know.

ilblackdragon: Would it make sense to add this to metrics folder instead of nn? We started
one in contrib/metrics -
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/metrics/python/

ilblackdragon: Would it make sense to add this to metrics folder instead of nn? We started
one in contrib/metrics -
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/metrics/python/
On Apr 21, 2016 3:35 PM, "lucasmoura" <notifications@github.com> wrote:

In tensorflow/python/ops/nn_ops.py
<https://github.com/tensorflow/tensorflow/pull/1999#discussion_r60666190>:

> +    values = [ 0 for i in range(num_labels * num_labels)]
> +
> +    for index, value in enumerate(predictions):
> +        target = index_to_label[targets[index]]
> +        value = index_to_label[value]
> +
> +        position = value * num_labels + target
> +
> +        values[position] += 1
> +
> +    indexes = [ [i, j] for i in range(num_labels) for j in range(num_labels)
> +                if values[i * num_labels + j] != 0]
> +    values = [ value for value in values if value != 0]
> +    shape = (num_labels, num_labels)
> +    dense_matrix = np.zeros(shape=(num_labels, num_labels))
> +

I have managed to create both indices and values array with two lines of
code. I don't know if my approach is the way you originally thought about
in order to reduce the lines of code, but if it is not, please let me know.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub
<https://github.com/tensorflow/tensorflow/pull/1999/files/303d2e059eefe8e545de59ccf08a21f397720a74#r60666190>

lucasmoura: Yeah, I think it makes more sense if confusion matrix is in the metrics folder. I will start moving the code to the appropriate module.

girving:
Here's an alternative implementation (details such as name skipped):

    def confusion_matrix(predictions, target, n=None):
      if n is None:
        n = tf.max(tf.reduce_max(predictions), tf.reduce_max(target)) + 1
      return tf.sparse_add(tf.zeros(tf.pack([n, n]), tf.transpose(tf.pack([predictions, target])))

tensorflow/python/ops/nn_ops.py
lucasmoura: @ilblackdragon Done

lucasmoura:
@girving I have tested the code locally and there is a problem in the fact that both sparse_add parameters are not SparseTensors.

concretevitamin:
@lucasmoura could you paste the errors somewhere?  `tf.sparse_add()` should support both S+S and S+D.

lucasmoura:
@concretevitamin No problem

```python
import tensorflow as tf
import numpy as np

predictions = np.array([1, 2, 3])
targets = np.array([3, 2, 1])
n = 4
print tf.sparse_add(tf.zeros(tf.pack([n, n])), tf.transpose(tf.pack([predictions, targets]))).eval()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-e7c79bd8a1b6> in <module>()
----> 1 print tf.sparse_add(tf.zeros(tf.pack([n, n])), tf.transpose(tf.pack([predictions, targets]))).eval()

/home/lucas/Envs/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/ops/sparse_ops.pyc in sparse_add(a, b, thresh)
    193   """
    194   if not any(isinstance(inp, ops.SparseTensor) for inp in [a, b]):
--> 195     raise TypeError("At least one input should be SparseTensor; do you mean to"
    196                     " use tf.add()?")
    197 

TypeError: At least one input should be SparseTensor; do you mean to use tf.add()?
```

concretevitamin:
Yeah, in your snippet both operands are passed as dense Tensors.  Geoffrey's snippet actually does Dense + Sparse.

girving:
Apologies: I had a broken version first and then edited it to be more
correct.

lucasmoura:
@girving I have implemented confusion matrix according to your implementation. Thanks for displaying this approach, I should have used tensorflow built-in functions to implement confusion matrix.

Furthermore, I could not use use values=[1] in SparseTensor. It is required that values has the same amount of elements as the first dimension of the indices matrix. 

Finally, if there is anything that I can improve in the function or its tests, please let me know.

tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
```diff
@@ -0,0 +1,99 @@
...
+  predictions_shape = predictions_tensor.get_shape()
+  targets_shape = targets_tensor.get_shape()
+
+  if predictions_shape.ndims != 1:
+    raise ValueError(
+      "targets is not a 1-D vector")
+
+  if targets_shape.ndims != 1:
+    raise ValueError(
+      "targets is not a 1-D vector")
```
girving: These checks don't work if the rank isn't known at runtime, so they should be removed.

tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
```diff
@@ -0,0 +1,99 @@
...
+    ValueError: If both predictions and targets are not 1-D vectors and do not
+                have the same size.
+  """
+  predictions_tensor = ops.convert_to_tensor(
+    predictions, name="inputs_tensor", dtype=dtypes.int64)
+  targets_tensor = ops.convert_to_tensor(
+    targets, name="targets_tensor", dtype=dtypes.int64)
+
+  predictions_shape = predictions_tensor.get_shape()
+  targets_shape = targets_tensor.get_shape()
```
girving: get_shape isn't always known, so it shouldn't occur here.

tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
```diff
@@ -0,0 +1,99 @@
...
+    possible labels in the classification task.
+
+  Raises:
+    ValueError: If both predictions and targets are not 1-D vectors and do not
+                have the same size.
+  """
+  predictions_tensor = ops.convert_to_tensor(
+    predictions, name="inputs_tensor", dtype=dtypes.int64)
+  targets_tensor = ops.convert_to_tensor(
+    targets, name="targets_tensor", dtype=dtypes.int64)
```
girving: Don't add "_tensor" to names.  "targets" > "target_tensor_" > "targets_tensor_that_is_the_name_of_a_tensor_in_tensorflow".

tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
```diff
@@ -0,0 +1,99 @@
...
+      "predictions and targets must be the same size")
+
+  if num_classes is None:
+    num_classes = math_ops.maximum(math_ops.reduce_max(predictions_tensor),
+                                   math_ops.reduce_max(targets_tensor)) + 1
+
+  shape = array_ops.pack([num_classes, num_classes])
+  indices = array_ops.transpose(
+    array_ops.pack([predictions_tensor, targets_tensor]))
+  values = array_ops.ones(shape=[indices.get_shape()[0], ], dtype=dtypes.int32)
```
girving: `arrays_ops.ones(shape=shape, dtype=dtypes.int32)` so that it works if shape is only known at runtime.

girving:
Thanks!  Can you merge the two commits into one?  Tests for new functionality should be added in the same commit as the functionality.

tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
lucasmoura: I think this approach has one problem. In the SparseTensor documentation it is stated that the indices array must have a shape [N, ndims] and values must have a shape of [N]. I believe that this N variable is not always equal to the num_classes one.

girving: Ah, yes, you're absolutely right.  Do `array_ops.ones_like(predictions, dtype=dtypes.int32)` instead.

lucasmoura:
@girving Done, I have fixed the issues and merged the commits into one.

tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
```diff
@@ -0,0 +1,86 @@
...
+    possible labels in the classification task.
+
+  Raises:
+    ValueError: If both predictions and targets are not 1-D vectors and do not
+                have the same size.
+  """
+  with ops.op_scope([predictions, targets, num_classes], name,
+                    'confusion_matrix') as name:
+    predictions = ops.convert_to_tensor(
+      predictions, name="inputs_tensor", dtype=dtypes.int64)
```
girving: `name="predictions"`

tensorflow/contrib/metrics/python/ops/confusion_matrix_ops.py
```diff
@@ -0,0 +1,86 @@
...
+    targets = ops.convert_to_tensor(
+      targets, name="targets_tensor", dtype=dtypes.int64)
```
girving: `name="targets"`

tensorflow/contrib/metrics/python/kernel_tests/confusion_matrix_ops_test.py
```diff
@@ -0,0 +1,127 @@
...
+        ValueError, "are not compatible",
+        tf.contrib.metrics.confusion_matrix, predictions, targets)
+
+  def testInputDifferentSize(self):
+    predictions = np.asarray([1, 2, 3])
+    targets = np.asarray([1, 2])
+    self.assertRaisesRegexp(
+          ValueError,
+          "are not compatible",
+          tf.contrib.metrics.confusion_matrix, predictions, targets)
```
girving: These tests probably fail now.

girving:
Excellent!  The only code comment is the names, but you may want to rerun and update the tests.  Then I'll test and merge.

tensorflow/contrib/metrics/python/kernel_tests/confusion_matrix_ops_test.py
lucasmoura: I have rerun both tests and both of them are passing. Maybe another operation is raising the exceptions. But if it is better, I can remove or rewrite them.

lucasmoura:
@girving Done, I have fixed both names.

girving:
Jenkins, test this please.

girving:
Ug, build timed out.  @martinwicke: Should I just retry?

martinwicke:
Yes, try again.

girving:
Jenkins, test this please yet again.

lucasmoura:
Weird, I don't see any relation of both build failures with the code I am adding. I will look into the BUILD file I have modified to run the confusion matrix tests.

lucasmoura:
@girving Should I keep rebasing this branch or wait for another jenkins build to see the errors ?

girving:
Sorry, slow start today.  Unfortunately it looks like the extra rebase destroyed the link to the tests, so I'll rerun them again. :/

girving:
Jenkins, test this please.

lucasmoura:
@girving Sorry for that, I realized that after I have already applied the rebase on my branch. But the errors given by Jenkins are still the same ones as the last run. However, I don't see how my code is affecting the matmul tests.

girving:
Thanks!  I don't think that failure is related.

jinghuangzhu:
So is this documented in the API? Are the confusion matrix accumulated and updated through multiple testing batches at testing time? Thanks.

concretevitamin:
@jinghuangzhu: https://www.tensorflow.org/versions/master/api_docs/python/contrib.metrics.html#confusion_matrix

PR #2006, opened 2016-04-18
Duplicated description

jcyk:
No need to point out twice.


vincentvanhoucke:
Shape and type are different things.

martinwicke:
The wording isn't great. However, this is an autogenerated file. You'd have to modify the source of this doc which is the docstring of the function in question.

PR #2009, opened 2016-04-18
Include TensorBoard TAG in pip_package

teamdandelion:


danmane:
Tested building the pip package with this, and the TAG file is now present.

PR #2019, opened 2016-04-19
docs only: removes sudo from pip/easy_install in instructions

alanzoppa:
`sudo`, where removed in the below instances, is superfluous, nonstandard, and is not a sensible default.  When using a virtualenv or similar, this can cause unsafety later on. Below examples work when pip/easy_install/npm are installed in userland. `pip` will complain if it needs root.

### DIFF
--- a/tensorflow/examples/skflow/README.md
+++ b/tensorflow/examples/skflow/README.md
@@ -4,8 +4,8 @@ Scikit Flow is high level API that allows to create,
 train and use deep learning models easily with well
 known Scikit Learn API.
 
-To run these examples, you need to have `scikit learn` library installed (`sudo pip install sklearn`).
-Some examples use the `pandas` library for data processing (`sudo pip install pandas`).
+To run these examples, you need to have `scikit learn` library installed (`pip install sklearn`).
+Some examples use the `pandas` library for data processing (`pip install pandas`).
 
 * [Deep Neural Network Regression with Boston Data](boston.py)
 * [Convolutional Neural Networks with Digits Data](digits.py)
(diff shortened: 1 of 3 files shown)


alanzoppa:
PS: sorry for being a pedant.

alanzoppa:
Also, I signed it!

alanzoppa:
Wrong email address vis-a-vis CLA is fixed.

martinwicke:
We probably put it in because on managed corporate machines the sudo can be necessary. Let me test on a clean machine. I'm all for removing it if it works for people in general.

vrv:
Actually we left `sudo` off in our initial docs, and many of our users who weren't installing in virtualenvs kept running into permissions problems, so we added `sudo` and the ones who knew what they were doing left it off in virtualenvs ;).

martinwicke:
We already have no sudo in the virtualenv section of the docs. Sadly, as @vrv has said, in the non-virtualenv case, it is terribly common to require either a targeted chmod/chown or sudo. 

I'd rather not go back to fixing people's permissions. :(

alanzoppa:
Oof. @martinwicke @vrv I am sympathetic to your position. To summarize:

1. On a vanilla install, most users will need root.
2. pip/easy_install/anything will give pretty clear permission denied errors
3. Non-experienced users will hassle the maintainers personally if confused about #2
4. #3 is a show-stopper at google scale

I am sympathetic to your position and I can't even say you're making the wrong choice. That said, it gives me a sad. :(

Thanks in any event for your attention to this matter.

martinwicke:
Well "personally" is a little harsh -- nobody has yelled at me on the street about this yet. But it is a support burden we'd rather not go back to. I'm sad about it too.

I'll close this sad PR.

PR #2026, opened 2016-04-19
One more version update for 0.8.0 final

caisq:


martinwicke:
Are there other occurences of 0.8.0rc0? It's safe to assume they're all supposed to be converted (except in RELEASES.md).

martinwicke:
The timing is, as long as we don't deploy the website between this merge and the release, we're good.

Issue #1786, opened 2016-04-06
feature request: support for cudnn v5 rc


fayeshine:
Anyone can help this?
https://developer.nvidia.com/rdp/cudnn-download

fayeshine:
FYI, https://github.com/Theano/Theano/issues/4342#

vrv:
I looked into this just now briefly -- there are a lot of API changes that are somewhat annoying to #ifdef around but I made progress.  The biggest changes have to do with convolution APIs: they exposed the convolution algorithms as required parameters, and from the Theano example, it seems they decided to plumb the algo as an option through their API, which seems kind of ugly.

There seems to be a "FindConvolution*Algorithm" function in the API which might be able to choose the best one automatically but there's no documentation so I have no idea if this is what we should be calling (and why the cudnn APIs don't call this automatically underneath the hood).

If anybody has any ideas about what we should do here, please let us know.

vrv:
https://github.com/vrv/tensorflow/commit/a76449737fed046b7e9f9871dd2d1d5c2f51ead0

Built for me with R5, I was able to run a few benchmarks, though not a huge difference, presumably because I'm not choosing the 'optimal algorithms'.

alquraishi:
They claim to be getting rather [substantial speed improvements](https://pbs.twimg.com/media/CfTGrS9XIAENPCv.jpg) on RNNs and LSTMs.

vrv:
#1794 added basic support for v5 RC, so you can at least install the library on your system and it will work.  We haven't performance optimized it or added the RNN/LSTM API yet.

esd100:
I'm not sure why but I am unable to do a Bazel PIP build past cuDNN.

I'm using the instructions and forked TF from here.
https://gist.github.com/ageitgey/819a51afa4613649bd18

TF_UNOFFICIAL_SETTING=1 ./configure
bazel build -c opt --config=cuda //tensorflow/cc:tutorials_example_trainer

[TF_BazelPIPBuild_TerminalOutput.txt](https://github.com/tensorflow/tensorflow/files/211039/TF_BazelPIPBuild_TerminalOutput.txt)

vrv:
I'm not sure stream executor works for OS X yet (hence #664).  I did test cudnn v5 on my local machine and it seemed to work, so I'm not entirely sure why it's still failing for you -- might double check that the symlinks and cudnn header file and ./configure options have been set properly.

wchan:
[...]

PR #2027, opened 2016-04-19
set cudnnSetFilter4dDescriptor format using filter layout for cudnn_v5

kashif:
`CUDNN_TENSOR_NHWC` the only other cudnn filter format is still not supported since `kInputYXOutput` would correspond to `CUDNN_TENSOR_CHWN` which doesn’t exist

for issue #1786

### DIFF
--- a/tensorflow/stream_executor/cuda/cuda_dnn.cc
+++ b/tensorflow/stream_executor/cuda/cuda_dnn.cc
@@ -382,28 +382,14 @@ class ScopedFilterDescriptor {
                  << ToString(status);
     }
 
-#if CUDNN_VERSION >= 5000
-    cudnnTensorFormat_t format;
-    switch (batch_descriptor.layout()) {
-      case dnn::DataLayout::kBatchYXDepth:
-        format = CUDNN_TENSOR_NHWC;
-        break;
-      case dnn::DataLayout::kBatchDepthYX:
-        format = CUDNN_TENSOR_NCHW;
-        break;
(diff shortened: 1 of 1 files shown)


vrv:
@tensorflow-jenkins : test this please

vrv:
googlebot where are you?

vrv:
@googlebot, wakey wakey

@willnorris in case you know why CLA check is not triggering anymore.

willnorris:
I'll dig into it.  But in the meantime, this is fine to merge ([signcla/kashif](http://signcla/kashif))

willnorris:
ah, looks like a race condition between when GitHub sent us the webhook and when we tried to fetch the commits for the pull request.

Also, I see that you have the CLA check required, so you can't just merge past this (probably not a good idea until we have better tools to manually override the CLA check).  In the meantime, @kashif would you mind leaving a comment here?  Doesn't matter what it is, but a comment from you will force the CLA check to run again.

vrv:
Actually I'm an admin so I can merge it forcefully -- I'd be happy to wait to get the official seal if you think it's for some reason better.

martinwicke:
We're already merging things we know are kosher (PRs auto-created from internal CLs, cherry-picks from another branch, ...), so I think there's no reason to wait if clabot has trouble re-checking this particular PR.

vrv:
Done.

kashif:
thanks @vrv

PR #2042, opened 2016-04-21
R0.7

mikyungjeon:
please i want download tensorflow data


vrv:
(assuming this is an accident).

PR #2058, opened 2016-04-21
use cudnnGetErrorString to get cudnn error string

kashif:
the built-in cudnnGetErrorString does exactly the same as the ToString helper


vrv:
@leary-google any ideas why you did it the other way?

vrv:

kashif:
ah damn @vrv the reason it was done the other way is because the macro is for the cudnn functions that return status, and this function is the only one that returns a `char *`.... ok closing

PR #2059, opened 2016-04-22
Add caffe importer to resources

martinwicke:


1LISA2:
Why do black people exist, black people still have a lot of AIDS, black, ugly, poor and violent

PR #2069, opened 2016-04-22
FIFOBucketedQueue

wchan:
New FIFOBucketedQueue! This queue is similar to the existing FIFOQueue but w/ buckets. This allows you to use the same graph for all your buckets rather than creating multiple graphs for each bucket (i.e., as shown in the TF examples for seq2seq). The main advantage is having a single graph (makes creating a graph much faster and simplier python code). Finally, for those unfamiliar w/ bucketing, see Sutskever et al., 2014 "Sequence to Sequence Learning with Neural Networks", it is a strategy to sample by say sequence length to reduce the variance of each minibatch, and consequently be much more efficient in computation (at the cost of losing uniform sampling, however this doesn't seem to be a big problem w/ large datasets).

FIFOBucketedQueueTest also included ; )


ebrevdo:
I've been able to replicate this behavior by constructing n+1 padded queues and using tf.case (default a noop).  n queues, one for each bucket, deququeing batch_size elements and enqueueing the result in a single enqueue to the final queue you will read from.

ebrevdo:
You'll also need Queue.from_list

vrv:
update on this?  @ebrevdo @wchan

ebrevdo:
We'll work together to see if we can build a simple implementation using
queues.  That said, this approach has a certain appeal to me.

ebrevdo:
At the very least I'd like this to work more like PaddingFIFOQueue, if we go with this approach.

vrv:
Closing due to inactivity

Issue #1346, opened 2016-03-01
CUDA compilation on ArchLinux fails


e14159:
## Overview

Build fails with 

```
/usr/include/string.h: In function 'void* __mempcpy_inline(void*, const void*, size_t)':
/usr/include/string.h:652:42: error: 'memcpy' was not declared in this scope
     return (char *) memcpy (__dest, __src, __n) + __n;
```

## Specific information:
- ArchLinux system
- Compute Capability 3.0 necessary
- CuDNN 4, CUDA 7.5
- Bazel 0.2.0

#### Followed the installation from source instructions:
- CuDNN 4, CUDA 7.5 are installed in /opt/cuda
- `git clone -b v0.7.1 --recurse-submodules https://github.com/tensorflow/tensorflow`
- `cd tensorflow`
- `./configure` Details as attached. Building with cuda support and compute capability 3.0
- `~/bin/bazel build -c opt --config=cuda //tensorflow/cc:tutorials_example_trainer`
This results in the following error:
```

/usr/lib/gcc/x86_64-unknown-linux-gnu/5.3.0/include/mwaitxintrin.h(36): error: identifier "__builtin_ia32_monitorx" is undefined

/usr/lib/gcc/x86_64-unknown-linux-gnu/5.3.0/include/mwaitxintrin.h(42): error: identifier "__builtin_ia32_mwaitx" is undefined

2 errors detected in the compilation of "/tmp/tmpxft_00006c12_00000000-7_avgpooling_op_gpu.cu.cpp1.ii".
```
This is probably due to cuda requiring gcc 4.9, while I am using 5.3. I fixed this by editing `crosstool_wrapper_driver_is_not_gcc` in `tensorflow/third_party/gpus/crosstool/clang/bin`, changing line 49 to the explicit
`GCC_HOST_COMPILER_PATH = ('/usr/bin/gcc-4.9')`
This allows compilation to continue; however, it then fails with 
```
/usr/include/string.h: In function 'void* __mempcpy_inline(void*, const void*, size_t)':
/usr/include/string.h:652:42: error: 'memcpy' was not declared in this scope
   return (char *) memcpy (__dest, __src, __n) + __n;
                                          ^
ERROR: /home/RED/src/tensorflow/tensorflow/core/BUILD:334:1: output 'tensorflow/core/_objs/gpu_kernels/tensorflow/core/kernels/aggregate_ops_gpu.cu.o' was not created.
[...]

PR #2073, opened 2016-04-23
solve compatibility issue for ubuntu 16.04

fayeshine:
Ubuntu 16.04 LTS has update default `gcc` to 5.3 and `string.h` as well, even we install a `gcc-4.9`,
we will get a bug like

```
usr/include/string.h: In function 'void* __mempcpy_inline(void*, const void*, size_t)':
/usr/include/string.h:652:42: error: 'memcpy' was not declared in this scope
     return (char *) memcpy (__dest, __src, __n) + __n;
```

this bus is described in https://github.com/tensorflow/tensorflow/issues/1346
and can be solved by 
insert `cxx_flag: "-D_FORCE_INLINES"` into `tensorflow/third_party/gpus/crosstool/CROSSTOOL`

### DIFF
--- a/third_party/gpus/crosstool/CROSSTOOL
+++ b/third_party/gpus/crosstool/CROSSTOOL
@@ -50,6 +50,7 @@ toolchain {
   # Use "-std=c++11" for nvcc. For consistency, force both the host compiler
   # and the device compiler to use "-std=c++11".
   cxx_flag: "-std=c++11"
+  cxx_flag: "-D_FORCE_INLINES"
   linker_flag: "-lstdc++"
   linker_flag: "-B/usr/bin/"
 


myme5261314:
I've confirmed this commit solved my same issue. :-)

zheng-xq:
This change has more impact that this issue alone. So I would like to find out whether there is another way. 

1. Is aggregate_ops_gpu.cu.cc the only *.cu.cc translation unit that has this problem? You can continue to build with "-k". 
2. Could you list the header file inclusion order from aggregate_ops_gpu.cu.cc to the offending header file on your system? We should check whether all of them are necessary.

hawkinsp:
I can confirm that the problem is not confined to aggregate_ops_gpu.cu.cc. On my home Ubuntu 16.04 machine, if I
bazel build -k -c opt --config=cuda //tensorflow/cc:tutorials_example_trainer  
I get ~190 instances of the same error (presumably that's almost all of the CUDA code).

The suggested fix works for me.

vrv:
@zheng-xq @fayeshine any updates?

vrv:
friendly ping for updates?

fayeshine:
@vrv @zheng-xq sorry for the late reply, I'm sorry that I don't have enough to further work on this problem. This will touch too much fundamental code in c++ and cuda. If someone can help this, welcome.

jendap:
Have we changed this recently? I have just build "bazel test //tensrflow/..." without any error on 16.04...

I will setup CI build for it tomorrow then we will know for sure :)

yaroslavvb:
@jendap what if you do`bazel test --config=cuda`

vrv:
@zheng-xq: apparently discussion on https://github.com/torch/torch7/issues/670 suggests that others have

1) added this flag FORCE_INLINES

2) Suggested people move to cuda 8.0 RC which apparently fixes the problem.

zheng-xq:
#2 seems to be good path right now.

vrv:
ok, closing, since it sounds like a cuda / nvidia problem that has a workaround (upgrade), and we'd rather not add the large hammer of FORCE_INLINES in our entire build

yaochx:
It's the right way to fix it

PR #2074, opened 2016-04-23
Add a link for installing from sources in os_setup.md

fayeshine:
Add a link for installing from sources in os_setup.md


caisq:
(and as test for yourself along the way)

PR #2078, opened 2016-04-24
remove unused array assignment

rasbt:
The `array` in

```
array = np.random.rand(32, 100, 100)
```

in the context 

```
import numpy as np
array = np.random.rand(32, 100, 100)

def my_func(arg):
  arg = tf.convert_to_tensor(arg, dtype=tf.float32)
  return tf.matmul(arg, arg) + arg

# The following calls are equivalent.
value_1 = my_func(tf.constant([[1.0, 2.0], [3.0, 4.0]]))
value_2 = my_func([[1.0, 2.0], [3.0, 4.0]])
value_3 = my_func(np.array([[1.0, 2.0], [3.0, 4.0]], dtype=np.float32))
```

is never used. I suggest removing it for clarity.

### DIFF
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -524,7 +524,6 @@ def convert_to_tensor(value, dtype=None, name=None, as_ref=False):
 
   ```python
   import numpy as np
-  array = np.random.rand(32, 100, 100)
 
   def my_func(arg):
     arg = tf.convert_to_tensor(arg, dtype=tf.float32)


mrry:
The markdown file is auto-generated from a docstring in `tensorflow/python/framework/ops.py`. Can you please make the change there?

rasbt:
@mrry Ah, sorry about that ... that was quite stupid; reverted the previous change and applied the fix to the ops.py file now

mrry:

mrry:
Thanks for the fix! It looks like Jenkins isn't listening to me, so I'm going to go ahead and merge it directly.

Issue #512, opened 2015-12-15
Possible bug in attention seq2seq


wchan:
source file is seq2seq.py

    def attention(query):
      """Put attention masks on hidden using hidden_features and query."""
      ds = []  # Results of attention reads will be stored here.
      for a in xrange(num_heads):
        with vs.variable_scope("Attention_%d" % a): 
          y = rnn_cell.linear(query, attention_vec_size, True)
          y = array_ops.reshape(y, [-1, 1, 1, attention_vec_size])
          # Attention mask is a softmax of v^T * tanh(...).
          s = math_ops.reduce_sum(
              v[a] * math_ops.tanh(hidden_features[a] + y), [2, 3]) 
          a = nn_ops.softmax(s)
          # Now calculate the attention-weighted vector d.
          d = math_ops.reduce_sum(
              array_ops.reshape(a, [-1, attn_length, 1, 1]) * hidden,
              [1, 2]) 
          ds.append(array_ops.reshape(d, [-1, attn_size]))
      return ds
                                       

at line 440, you compute the energies, ok… then at line 441 you normalize the energies… you need to do a -FLT_MAX mask on only the part of the encoder sequence (i.e., you don’t want to pay attention to “empty” part of the sequence)... I don't see the masking op? Or am i missing something?

I can wrap an op patch for this if nobody is working on this (let me know).

Thanks all!

lukaszkaiser:
I was thinking about it and there were 2 things that stopped me from working on it. The first and most important one is that I get good results without any special handling for masks, just as it is. Are you sure that a -FLT_MAX really helps, did you try without it? (I think the net just learns to never attend to the padding easily.) The second reason was that one day, when "while" works ok, we might not need to pad at all -- and I did not want to add complexity to the API that is not applicable to that case.

That said, if you find it important and want to do it -- by all means, send a patch :).

LeavesBreathe:
> I get good results without any special handling for masks, just as it is

I too have been using this for about a month, and I also get good results just as it is. I think it learns to never attend to padding as that is a simple if statement. In keras, we also talked about this, and concluded that masking was not necessary because the network learns to ignore padding.

ogrisel:
[...]

PR #2079, opened 2016-04-24
Attention Mask Ops

wchan:
The current attention seq2seq example (i.e., in seq2seq.py function attention_decoder) has a bug. The code (quoted below):

520     def attention(query):
521       """Put attention masks on hidden using hidden_features and query."""
522       ds = []  # Results of attention reads will be stored here.
523       for a in xrange(num_heads):
524         with variable_scope.variable_scope("Attention_%d" % a):
525           y = rnn_cell.linear(query, attention_vec_size, True)
526           y = array_ops.reshape(y, [-1, 1, 1, attention_vec_size])
527           # Attention mask is a softmax of v^T \* tanh(...).
528           s = math_ops.reduce_sum(
529               v[a] \* math_ops.tanh(hidden_features[a] + y), [2, 3])
530           a = nn_ops.softmax(s)
531           # Now calculate the attention-weighted vector d.
532           d = math_ops.reduce_sum(
533               array_ops.reshape(a, [-1, attn_length, 1, 1]) \* hidden,
534               [1, 2])
535           ds.append(array_ops.reshape(d, [-1, attn_size]))
536       return ds

does not do a masking op before the softmax. This git pull request will fix this problem, i.e., we should mask the attention energies before the softmax based on the encoder sequence length.


wchan:
this is in reference to this issue: https://github.com/tensorflow/tensorflow/issues/512

lukaszkaiser:
I agree that it would be good to add masking to the attention model and API. But I think adding an op and a separate cuda kernel just for that is a huge overkill. Why not just multiply pointwise by the mask in python before the softmax? Without a good reason, I find a PR like this not acceptable, the code complexity and maintenance cost looks much larger than the simple problem it solves.

wchan:
sometimes the mask is generated dynamically (i.e., the mask is a function
of the previous attention)...

for example in speech recognition: Bahdanau/Chorowski et. al.,
http://arxiv.org/pdf/1508.04395.pdf
http://papers.nips.cc/paper/5847-attention-based-models-for-speech-recognition.pdf

--
William Chan
Carnegie Mellon University
(650) 450-9455
williamchan.ca

lukaszkaiser:
It is possible that I don't understand the problem, so please, clarify. But why would it need a special op to make the mask a function? If it's a tensor of 0s and 1s, it can be changed at every step, why can't this be done the simple way in python? Is there a very big speed advantage to having an extra op?

wchan:
Well, the python side does not know what mask to generate (i.e., it doesnt
know the sequence_len of each sample in the minibatch). This is especially
true in speech where your minibatch has a high variance in sequence_len.
However, this problem can be somewhat alleviated to generate the mask of 1
and -FLT_MAX (not 1 and 0), if our input layer generates that for us.

However, this is not true for all attention masks, for example in speech
recognition where the encoder length can be thousands of frames, we want
the attention mask to be a function of the previous attention alignment,
i.e., in the papers I cited above, they generate the mask by looking at the
median of the previous attention and then use a window around it (in
essence, you have a somewhat monotonic sliding window on where to attend
to). Correct me if I am wrong, but I believe this can not be generated
python side?

Its not a speed advantage, (correct me if I am wrong), but I don't think
you can mask properly w/o an op like this.

--
William Chan
Carnegie Mellon University
(650) 450-9455
williamchan.ca

lukaszkaiser:
I was thinking about it a bit and I still don't see the problem. Here is how I'd do it:
* add a parameter to attention_decoder called mask_function
* this parameter is a python function that takes the step number and the previous attention mask tensor
  and outputs a new attention mask tensor
* every attention is multiplied by mask_function(prev_attention, step_number) (one-liner in python)

Would that work? If you want to feed-in attention, you'd just provide mask_function(_, i) = length-tensor[i] so you can feed the example lengths. And if you want to average by previous (or move it with a step), you just need to write a lambda that does it.

Or am I missing something crucial here?

wchan:
sorry, didnt quite catch the last step:

> you'd just provide mask_function(_, i) = length-tensor[i]
>
how does "length - tensor[i]" generate the mask? length is a tensor of the
sample length? and tensor[i] is what? and doesn't this result in a scalar
not a mask?

thanks!

wchan:
ok. i think ur right, just need a tf.cond or something similar in the
mask_function right? and then use that to generate the mask?

--
William Chan
Carnegie Mellon University
(650) 450-9455
williamchan.ca

lukaszkaiser:
To answer your question, let me be a little bit more precise. What I suggest is to modify the [seq2seq.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/seq2seq.py) in two ways.

(1) on lines 546, 565, 567 instead of `attns = attention(state)` let's do `attns = attention(state, attns, inp, i)` so the `attention()` function now takes the arguments `attention(query, previous_attention, decoder_input, step_number)`

(2) on line 530 instead of `a = nn_ops.softmax(s)` let's do `a = nn_ops.softmax(s + mask_f(previous_attention, decoder_input, step_number) * -LARGE_NUMBER` where `mask_f` is an extra argument to the decoder.

In this way the mask_f function gets all the current information in the step. Its task is to generate 1.0 (or anything positive) if we want to mask something and 0.0 otherwise. The 1.0 will be turned into a -LARGE_NUMBER bias in the softmax and effectively disable that field for attention, right?

The tensor s has shape [batch_size x attn_length], where attn_length is the size of the attention_states (length of the encoder in a seq2seq model). Let me focus on the basic case when we want mask_f to mask an input iff the decoder input symbol is a padding, which means that it is all zeros. We can do this in two ways, I think. If we're sure the decoder input is all 0s only if it's padded and has a reasonably large norm otherwise, then we can do `mask_f(_, dec_inp, _) = tf.minimum(1.0, tf.nn.l2_loss(dec_inp) * LARGE)`. If we're not sure of that, then we can do `mask_f(_, _, step) = self.decoder_mask[i]` when declaring the `mask_f` function somewhere in our model code. In this case `decoder_mask` are placeholders of the same kind as `decoder_inputs` that are 1 in case of padding and 0 otherwise -- and we need to make sure to feed the appropriate values.

I'm sure I missed some things. In particular, `previous_attention` coming to the `attention` function should probably be the previous attention mask, not the previous attention vector as I suggested above (so the new mask needs to be returned in the function and the pointer saved in the loop, a few more lines). But does the general idea sound reasonable and flexible enough?

wchan:
I agree in general your idea works. Just some minor things...

> The tensor s has shape [batch_size x attn_length], where attn_length is
> the size of the attention_states (length of the encoder in a seq2seq
> model). Let me focus on the basic case when we want mask_f to mask an input
> iff the decoder input symbol is a padding, which means that it is all
> zeros. We can do this in two ways, I think. If we're sure the decoder input
> is all 0s only if it's padded and has a reasonably large norm otherwise,
> then we can do mask_f(_, dec_inp, _) = tf.minimum(1.0,
> tf.nn.l2_loss(dec_inp) * LARGE). If we're not sure of that, then we can
> do mask_f(_, _, step) = self.decoder_mask[i] when declaring the mask_f
> function somewhere in our model code. In this case decoder_mask are
> placeholders of the same kind as decoder_inputs that are 1 in case of
> padding and 0 otherwise -- and we need to make sure to feed the appropriate
> values.
>
The problem w/ this approach is (where self.decoder_inputs is generated by
the input layer), is that the encoder may have a pyramid architecture
(i.e., time subsampling) which is very common in speech. i.e., depending on
the graph, we may reduce it by 4/8/16 multipliers. so we need to know this
ahead of time (the time factor reduction) and pass that in to the input op
that is generating the mask.

Where as w/ a op that does the masking for you where you only need to pass
in the sequence_length, every time we do hierarchical subsampling in the
graph, we can just do sequence_length_tensor = sequence_length_tensor / 2;
energies = attention_mask(energies, sequence_length);

However, I think this problem can be alleviated by tf.cond(), where we can
do something like tf.cond(sequence_length < encoder_idx, 0, -FLT_MAX),
something like that... however we need to do this T times where T is the
length of the encoder and encoder_idx \in range(0, T).

But I agree w/ you, I think (havent completely tested it out) you are right
this can be done almost completely w/ the existing python ops. Personally,
I'll still be using my op since I already wrote it and it works and it
should be slightly faster (and simplier graph). But we can close this PR if
you think this is adding too much code to maintain.

PR #2094, opened 2016-04-25
tensorflow.bzl: strip bazel version number to work with patch bazel release

meteorcloudy:
The newest bazel version number is 0.2.2b, when need to strip 2b

### DIFF
--- a/tensorflow/tensorflow.bzl
+++ b/tensorflow/tensorflow.bzl
@@ -12,7 +12,7 @@ def _parse_bazel_version(bazel_version):
   # Turn "release" into a tuple of integers
   version_tuple = ()
   for number in parts[0].split('.'):
-    version_tuple += (int(number.strip('abcdefghij')),)
+    version_tuple += (str(number),)
   return version_tuple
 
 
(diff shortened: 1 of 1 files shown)


tensorflow/tensorflow.bzl
```diff
@@ -12,7 +12,7 @@ def _parse_bazel_version(bazel_version):
   # Turn "release" into a tuple of integers
   version_tuple = ()
   for number in parts[0].split('.'):
-    version_tuple += (int(number),)
+    version_tuple += (int(number.strip('abcdefghij')),)
```
vrv: can you replace 'int' with 'str' here ?

version_tuple += (str(number),)

We'll compare strings lexicographically, which will probably solve your problem here.

Ideally we'd all use semantic versioning conventions though.

vrv:
There are probably better ways to do this, but in the interest of unbreaking people, this seems fine.  Going to test and merge if successful.

vrv:

yvirin:
hurray

PR #2095, opened 2016-04-25
Update os_setup.md

hephaex:
Fixed command to Installing from sources on OS X

In file "os_setup.md"

When try to build from sources on OS X, it need to root permission during

> sudo bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
> sudo pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-any.whl

### DIFF
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -617,10 +617,10 @@ $ bazel build -c opt //tensorflow/tools/pip_package:build_pip_package
 # To build with GPU support:
 $ bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package
 
-$ bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
+$ sudo bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
 
 # The name of the .whl file will depend on your platform.
-$ pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-linux_x86_64.whl
+$ sudo pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-any.whl
 ```
 
 ## Setting up TensorFlow for Development


tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -617,10 +617,10 @@ $ bazel build -c opt //tensorflow/tools/pip_package:build_pip_package
 # To build with GPU support:
 $ bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package
 
-$ bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
+$ sudo bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
```
martinwicke: You really shouldn't need sudo for this one, unless your /tmp folder isn't writable for your user, or wherever you downloaded tensorflow to isn't.

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -617,10 +617,10 @@ $ bazel build -c opt //tensorflow/tools/pip_package:build_pip_package
...
 
 # The name of the .whl file will depend on your platform.
-$ pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-linux_x86_64.whl
+$ sudo pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-any.whl
```
martinwicke: We can make it more explicit that the name depends on the platform, maybe in the text below? We can also add a note there that depending on your pip and python installation, you may need to use `sudo` on this command. I'm a little wary about adding it for everybody, it really shouldn't be required.

tensorflow/g3doc/get_started/os_setup.md
hephaex: > @@ -617,10 +617,10 @@ $ bazel build -c opt //tensorflow/tools/pip_package:build_pip_package
>  # To build with GPU support:
>  $ bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package
>
> -$ bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
> +$ sudo bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg

In case of this, "/tmp/tensorflo_pkg" is as same as usage of other OS
platform like Ubuntu. but OS X environment   "/tmp" location of file system
is need to sudo permission. current package location occur to OS
dependency. I think that build and install from source code is not so good
specific location. If it is possible to change other description from
"/tmp/tensorflow_pkg" to other location, it is more flexible way to build
and installation.

tensorflow/g3doc/get_started/os_setup.md
hephaex: >
>  # The name of the .whl file will depend on your platform.
> -$ pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-linux_x86_64.whl
> +$ sudo pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-any.whl

Sure it is depend on the platform, In case of OS X, the pip and python
installation need to sudo permission.
In this reason, using OS X enviroment, the pip install description on "
os_setup.md", almost add sudo command.
This commit is just add sudo command on OS X. I do same install from source
code on Ubuntu 14.04. it is not use sudo permission.

martinwicke: Ok, can you change the command line back to without sudo, and add text below that command to point out that on some platforms, you may need sudo?

tensorflow/g3doc/get_started/os_setup.md
martinwicke: /tmp doesn't need sudo on my machine, which is a Mac with El Capitan. Maybe it does for non-admin users? Even that would very much surprise me though.

hephaex: I try to again using "bazel-bin/tensorflow/tools/pip_package/build_pip_package
/tmp/tensorflow_pkg"

A error occurred, I captured that message

IOError: [Errno 13] Permission denied:

'/Library/Python/2.7/site-packages/protobuf-3.0.0b2-py2.7.egg/EGG-INFO/namespace_packages.txt'

because of namespace_package.txt is allow to admin (root)
I change add sudo command "sudo
bazel-bin/tensorflow/tools/pip_package/build_pip_package
/tmp/tensorflow_pkg".

$ ls -al
/Library/Python/2.7/site-packages/protobuf-3.0.0b2-py2.7.egg/EGG-INFO/

total 56

drwxr-xr-x  2 root  wheel   306  4 20 18:13 .

drwxr-xr-x  4 root  wheel   136  4 20 18:13 ..

-rw-r-----  1 root  wheel   662  4 20 18:13 PKG-INFO

-rw-r-----  1 root  wheel  2801  4 20 18:13 SOURCES.txt

-rw-r-----  1 root  wheel     1  4 20 18:13 dependency_links.txt

-rw-r-----  1 root  wheel     7  4 20 18:13 namespace_packages.txt

-rw-r-----  1 root  wheel    19  4 20 18:13 requires.txt

-rw-r-----  1 root  wheel     7  4 20 18:13 top_level.txt

-rw-r--r--  1 root  wheel     1  4 20 18:13 zip-safe


the result of add sudo command, it is no problem. complete progress.
and I also found that "/tmp" is possible to us non-admin user at OS x EL
Capitan version.

How did you think about? it is special case on my environment? or general
case on OS X environment.

martinwicke: build_pip_package shouldn't really access an installed protobuf at all. Can you uninstall protobuf before running build_pip_package? 

@keveman, ideas?

keveman: This is really strange. This is some deep setuptools arcana that needs someone more expert than me. I went through our build_pip_package and setup.py, nothing there seems like it should need to access installed protobuf.

martinwicke:
I'll close this for inactivity. If you update the PR, I'm happy to reopen.

PR #2100, opened 2016-04-25
Branch 120724543

benoitsteiner:


caisq:

caisq:
Regarding the quantization related errors, please make another push from internal. There is a recent CL (120733385) that fixes it.

We'll also have to look into the remaining errors.

PR #2102, opened 2016-04-25
Those aren't floats...

kholbekj:
Am I misunderstanding something here? I fail to see floats.


kholbekj:
Signed CLA

martinwicke:
I believe these are actually floats. Sure, floats that each carry only one bit of information, but partly for historical reasons, partly for convenience, they are floats.

PR #2132, opened 2016-04-27
Fix sec_per_batch in a multi-GPU CIFAR-10 example

kibum14:
In a multi-GPU CIFAR-10 example, 'sec_per_batch' should be computed in the same way as a single GPU example. It's identical with the [inception-v3 example](https://github.com/tensorflow/models/blob/master/inception/inception/inception_train.py#L346-L351).


kibum14:
I signed it!

vrv:
closing due to inactivity and lack of CLA

PR #2149, opened 2016-04-28
Fix typos in example 'adding an op'

tvogels:
The colons after `@tf.RegisterShape("ZeroOut")` should not be there.


mrry:
Thanks for the fix!

PR #2152, opened 2016-04-28
fix scaler transform in boston skflow example

mheilman:
fit_transform was being used instead of transform for the test set in the boston example

### DIFF
--- a/tensorflow/examples/skflow/boston.py
+++ b/tensorflow/examples/skflow/boston.py
@@ -40,6 +40,6 @@ regressor = skflow.TensorFlowDNNRegressor(hidden_units=[10, 10],
 regressor.fit(X_train, y_train)
 
 # Predict and score
-score = metrics.mean_squared_error(regressor.predict(scaler.fit_transform(X_test)), y_test)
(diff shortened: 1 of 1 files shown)


mheilman:
My company (Civis Analytics) signed the CLA a while back.

poxvoculi:
@ilblackdragon Should transform or fit_transform be preferred here?

vrv:
@willnorris if users sign the corp CLA, is there any way to get the label to be set to yes?

willnorris:
The commit in this PR shows `mheilman@users.noreply.github.com` as the commit author.  You would need to use the email address that is associated with the corporate CLA in order to be recognized (presumably an `@civisanalytics.com` email?)

ilblackdragon:
@poxvoculi transform should be used usually - as in serving you don't actually know what would be mean / std.dev of the distribution.

ilblackdragon:

mheilman:
Gah, sorry, I forgot that GH would use that generated email address since I just made a simple fix in the in-browser interface.  The email is correct now.

vrv:

PR #2157, opened 2016-04-28
bug fix in ApplyAdadelta update rule

dpaiton:
Following the Zeiler ADADELTA paper (http://arxiv.org/pdf/1212.5701v1.pdf), numerator of update equation should be RMS(accum_update) \* grad. The code previously did not perform the RMS operation.


vrv:
does this change cause any tests to fail?  if not, shouldn't it have? :)

cc @Mistobaan

Mistobaan:
Is definitely missing the RMS part. The test pass because is missing the RMS there too.

dpaiton:
I can also fix the test in this PR if you would like?

vrv:
Yes please!

dpaiton:
@vrv can you point me to where the faulty test is? I looked in [adadelta_test.py] (https://github.com/tensorflow/tensorflow/blob/16d395e5dea687ab3aece0a462e631de25c8d77d/tensorflow/python/training/adadelta_test.py), but the update rule there looks correct. Also, I rebased my pull request to be up to date, which may have messed up the CLA process. I'm not sure how to proceed there...

vrv:
Unfortunately because you rebased I've lost the history of the tests, and also you probably rebased in the incorrect way, since a bunch of other commits are showing up in this PR.  Try restarting the PR from scratch?

dpaiton:
Yeah, I agree. I will do that. Thank you!

PR #2162, opened 2016-04-29
Fix variable names on confusion matrix metric

lucasmoura:
Considering that tensorflow metrics documentation states that every metric receives the parameters predictions and labels, this merge request fix the name of one of the confusion matrix variable, which is named "targets" and now is named "labels".


vrv:

PR #2172, opened 2016-04-30
Blacklist py tests in python/platform/default

caisq:
These tests are superseded by the corresponding ones in python/platform
in the OSS tests. They are currently used only internally and will be
eventually removed in the future.

This PR also fixes a bug in test_installation.sh that could cause infinite
loops in certain cases.


caisq:
Ready to merge. This should fix an error we are currently seeing in the nightly builds right now.

PR #2173, opened 2016-04-30
Import numpy before a call to sys.setdlopenflags

### LINKED ISSUE #2034: segmentation fault
I install tensorflow use:sudo pip install --upgrade https://storage.googleapis.com/tensorflow/linux/gpu/tensorflow-0.8.0rc0-cp27-none-linux_x86_64.whl

linux version is :Linux version 3.10.0-229.el7.x86_64 (builder@kbuilder.dev.centos.org) (gcc version 4.8.2 20140120 (Red Hat 4.8.2-16) (GCC) ) #1 SMP Fri Mar 6 11:36:42 UTC 2015

when I learn python, and import tensorflow,segmentation fault

>>> import tensorflow
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcublas.so locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcudnn.so locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcufft.so locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcuda.so.1 locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcurand.so locally
Segmentation fault (core dumped)

stephenroller:
Must import numpy before a call to sys.setdlopenflags, or there may be a segfault in certain situations. Fixes #2034.

The situation occurs when users have compiled TensorFlow, but use binary distributions of SciPy. Alternatives to this patch would be documentation stating that users who compile TensorFlow from source should also recompile SciPy if they need it.

### DIFF
--- a/tensorflow/python/__init__.py
+++ b/tensorflow/python/__init__.py
@@ -40,6 +40,10 @@ import traceback
 # the mode to RTLD_GLOBAL to make the symbols visible, so libraries such
 # as the ones implementing custom ops can have access to tensorflow
 # framework's symbols.
+# one catch is that numpy *must* be imported before the call to
+# setdlopenflags(), or there is a risk that later c modules will segfault
+# when importing numpy (gh-2034).
+import numpy as np
 _default_dlopen_flags = sys.getdlopenflags()
 sys.setdlopenflags(_default_dlopen_flags | ctypes.RTLD_GLOBAL)
 from tensorflow.python import pywrap_tensorflow


yaroslavvb:
Nice investigation...how did you find it/any idea why this happens? Adding @keveman who added this part of code (although he slow to respond due to ICLR)

martinwicke:
Jenkins, test this please.

Nice catch. Very obscure, much wow.

stephenroller:
Yeah so this is just like some horrible edge case and part of the reason for the UNIQUE_SYMBOL mess in NumPy I guess.

Anyway when you include NumPy, you have to call import_array to populate some function pointers. Each C module needs to do this separately, and is supposed to have its own housekeeping for these function pointers. With the GLOBAL flag on though, there is overlap in some of these shared symbols (the UNIQUE_SYMBOL is the same in NumPy and Scipy:  PyArray_API). The result is the initialization code generated by any f2py module will clobber the housekeeping and dereference a bad pointer. Whoops! But if you do the import *before* the GLOBAL flag is set, then there is no shared symbol conflict, so the modules happily do their own housekeeping like they expect. O_o

@yaroslavvb I have just really, really wanted to use TensorFlow on the Texas supercomputing clusters, which use too old GLIBC to use the binary distribution, so I've been stuck compiling TF myself :stuck_out_tongue:

martinwicke:
Hm... timeout, probably infrastructure failure. Will try again later.

vrv:

PR #2178, opened 2016-04-30
Remove loading sklearn even if it's present and only load when env flag is set

ilblackdragon:
Ref #2034 to reduce load time of import tf when people are not interested in tf.learn / sklearn integration.


martinwicke:
Can you fix the error?

ilblackdragon:
@martinwicke Fixed.

martinwicke:
Thanks!

PR #2183, opened 2016-05-01
In the doc example, commas in tensors are missing.

hunkim:
Missing commas in the doc example are added to show tensors more precisely. Also, they are now consistent with other reshape examples.


caisq:
Can you combine this with 2184 please?

hunkim:
Merged with PR #2184. Thanks!

PR #2184, opened 2016-05-01
Added clear description: -1 can also be used to infer the shape

hunkim:
The current description, "also be used with higher dimensional shapes" is not very clear.

-1 very commonly used to infer the shape. So, I added clear descriptions and examples how to use -1 in reshape.

### DIFF
--- a/tensorflow/core/ops/array_ops.cc
+++ b/tensorflow/core/ops/array_ops.cc
@@ -679,9 +679,21 @@ reshape(t, [2, 4]) ==> [[1, 1, 2, 2]
 # tensor 't' has shape [3, 2, 3]
 # pass '[-1]' to flatten 't'
 reshape(t, [-1]) ==> [1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6]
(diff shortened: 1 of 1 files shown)


tensorflow/core/ops/array_ops.cc
```diff
@@ -679,9 +679,21 @@ reshape(t, [2, 4]) ==> [[1, 1, 2, 2]
 # tensor 't' has shape [3, 2, 3]
 # pass '[-1]' to flatten 't'
 reshape(t, [-1]) ==> [1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6]
-# -1 can also be used with higher dimensional shapes
+
+# -1 can also be used to infer the shape
+# -1 is inferred to be 9
```
caisq: nit: Add a colon at the end of each line like this, so that it can be clearer whether it is talking about the example before or after it.

hunkim:
Thanks for the wonderful comments. It is merged with PR #2183 and colons are added.

caisq:

caisq:
Merged. Thanks.

PR #2187, opened 2016-05-01
Fix convert_to_tensor error in tf.one_hot

samjabrahams:
Addresses more issues found in #1799. When a user passes in a TensorFlow constant that is not of type `tf.int64` for `indices` or type `tf.int32` for `depth`, `one_hot()` throws a "Tensor conversion requested..." error due to using `convert_to_tensor` for those parameters. This PR addresses this by using `tf.cast()` instead for `indices` and `depth`.

Additionally, this PR adds more test coverage (has tests using `tf.constant`) and renames some tests introduced in #1869 to be more descriptive.

### DIFF
--- a/tensorflow/core/kernels/one_hot_op.cc
+++ b/tensorflow/core/kernels/one_hot_op.cc
@@ -74,7 +74,7 @@ class OneHotOp : public OpKernel {
     const int axis = (axis_ == -1) ? indices_dims : axis_;
 
     // The one-hot dimension.
-    const int32 depth_v = depth.scalar<int32>()();
+    const TI depth_v = depth.scalar<TI>()();
 
     TensorShape output_shape = indices_shape;
     output_shape.InsertDim(axis, depth_v);
--- a/tensorflow/core/kernels/one_hot_op.cc
+++ b/tensorflow/core/kernels/one_hot_op.cc
@@ -151,6 +151,7 @@ TF_CALL_GPU_NUMBER_TYPES(DECLARE_GPU_SPEC);
 
 }  // namespace functor
 
+// Registration of the GPU implementations.
 #define REGISTER_ONE_HOT_GPU_INDEX(type, index_type)              \
   REGISTER_KERNEL_BUILDER(Name("OneHot")                          \
                               .Device(DEVICE_GPU)                 \
--- a/tensorflow/core/kernels/one_hot_op.cc
+++ b/tensorflow/core/kernels/one_hot_op.cc
@@ -39,7 +39,7 @@ namespace tensorflow {
 typedef Eigen::ThreadPoolDevice CPUDevice;
 typedef Eigen::GpuDevice GPUDevice;
 
-template <typename Device, typename T>
+template <typename Device, typename T, typename TI>
 class OneHotOp : public OpKernel {
  public:
   explicit OneHotOp(OpKernelConstruction* ctx) : OpKernel(ctx) {
@@ -88,22 +88,22 @@ class OneHotOp : public OpKernel {
     // prefix_dim_size == # of elements before the axis
     // depth_v == # of elements per axis
     // suffix_dim_size == # of elements after the axis
-    int64 prefix_dim_size = 1;
+    TI prefix_dim_size = 1;
     for (int i = 0; i < axis; ++i) {
       prefix_dim_size *= indices_shape.dim_size(i);
     }
-    int64 suffix_dim_size =
+    TI suffix_dim_size =
         indices_shape.num_elements() / prefix_dim_size;
 
     // Split indices into matrix of size prefix_dim_size x suffix_dim_size
     auto indices_t =
-        indices.shaped<int64, 2>({prefix_dim_size, suffix_dim_size});
+        indices.shaped<TI, 2>({prefix_dim_size, suffix_dim_size});
     // Split output into 3-Tensor of size:
     //   prefix_dim_size x depth x suffix_dim_size.
     auto output_t =
         output->shaped<T, 3>({prefix_dim_size, depth_v, suffix_dim_size});
 
-    functor::OneHot<Device, T>::Compute(ctx->eigen_device<Device>(), indices_t,
+    functor::OneHot<Device, T, TI>::Compute(ctx->eigen_device<Device>(), indices_t,
                                         on_value_t, off_value_t, &output_t);
   }
 
@@ -113,44 +113,59 @@ class OneHotOp : public OpKernel {
   TF_DISALLOW_COPY_AND_ASSIGN(OneHotOp);
 };
 
-#define REGISTER_ONE_HOT(type)                           \
-  REGISTER_KERNEL_BUILDER(Name("OneHot")                 \
-                              .Device(DEVICE_CPU)        \
-                              .TypeConstraint<type>("T") \
-                              .HostMemory("depth"),      \
-                          OneHotOp<CPUDevice, type>);
+#define REGISTER_ONE_HOT_INDEX(type, index_type)                  \
+  REGISTER_KERNEL_BUILDER(Name("OneHot")                          \
+                              .Device(DEVICE_CPU)                 \
+                              .TypeConstraint<index_type>("TI")   \
+                              .TypeConstraint<type>("T")          \
+                              .HostMemory("depth"),               \
+                          OneHotOp<CPUDevice, type, index_type>);
 
-TF_CALL_NUMBER_TYPES(REGISTER_ONE_HOT);
+#define REGISTER_ONE_HOT(type)          \
+  REGISTER_ONE_HOT_INDEX(type, int32);  \
+  REGISTER_ONE_HOT_INDEX(type, int64)
+
+TF_CALL_ALL_TYPES(REGISTER_ONE_HOT);
 
 #if GOOGLE_CUDA
 
 // Forward declarations of the functor specializations for GPU.
 namespace functor {
-#define DECLARE_GPU_SPEC(T)                                                   \
-  template <>                                                                 \
-  void OneHot<GPUDevice, T>::Compute(                                         \
-      const GPUDevice& d, const typename TTypes<int64>::ConstMatrix& indices, \
-      const typename TTypes<T>::ConstScalar& on_value,                        \
-      const typename TTypes<T>::ConstScalar& off_value,                       \
-      typename TTypes<T, 3>::Tensor* output);                                 \
-  extern template struct OneHot<GPUDevice, T>;
+#define DECLARE_GPU_SPEC_INDEX(T, TI)                                       \
+  template <>                                                               \
+  void OneHot<GPUDevice, T, TI>::Compute(                                   \
+      const GPUDevice& d, const typename TTypes<TI>::ConstMatrix& indices,  \
+      const typename TTypes<T>::ConstScalar& on_value,                      \
+      const typename TTypes<T>::ConstScalar& off_value,                     \
+      typename TTypes<T, 3>::Tensor* output);                               \
+  extern template struct OneHot<GPUDevice, T, TI>;
+
+#define DECLARE_GPU_SPEC(T)          \
+  DECLARE_GPU_SPEC_INDEX(T, int32);  \
+  DECLARE_GPU_SPEC_INDEX(T, int64);  \
 
 TF_CALL_GPU_NUMBER_TYPES(DECLARE_GPU_SPEC);
 
+#undef DECLARE_GPU_SPEC_INDEX
 #undef DECLARE_GPU_SPEC
 
 }  // namespace functor
 
-// Registration of the GPU implementations.
-#define REGISTER_ONE_HOT_GPU(type)                       \
-  REGISTER_KERNEL_BUILDER(Name("OneHot")                 \
-                              .Device(DEVICE_GPU)        \
-                              .TypeConstraint<type>("T") \
-                              .HostMemory("depth"),      \
-                          OneHotOp<GPUDevice, type>);
+#define REGISTER_ONE_HOT_GPU_INDEX(type, index_type)              \
+  REGISTER_KERNEL_BUILDER(Name("OneHot")                          \
+                              .Device(DEVICE_GPU)                 \
+                              .TypeConstraint<index_type>("TI")   \
+                              .TypeConstraint<type>("T")          \
+                              .HostMemory("depth"),               \
+                          OneHotOp<GPUDevice, type, index_type>);
+
+#define REGISTER_ONE_HOT_GPU(type)          \
+  REGISTER_ONE_HOT_GPU_INDEX(type, int32);  \
+  REGISTER_ONE_HOT_GPU_INDEX(type, int64);  \
 
 TF_CALL_GPU_NUMBER_TYPES(REGISTER_ONE_HOT_GPU);
 
+#undef REGISTER_ONE_HOT_GPU_INDEX
 #undef REGISTER_ONE_HOT_GPU
 
 #endif  // GOOGLE_CUDA
--- a/tensorflow/core/kernels/one_hot_op.cc
+++ b/tensorflow/core/kernels/one_hot_op.cc
@@ -74,7 +74,7 @@ class OneHotOp : public OpKernel {
     const int axis = (axis_ == -1) ? indices_dims : axis_;
 
     // The one-hot dimension.
-    const TI depth_v = depth.scalar<TI>()();
+    const int32 depth_v = depth.scalar<int32>()();
 
     TensorShape output_shape = indices_shape;
     output_shape.InsertDim(axis, depth_v);
--- a/tensorflow/core/kernels/one_hot_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/one_hot_op_gpu.cu.cc
@@ -21,17 +21,24 @@ limitations under the License.
 
 #include "tensorflow/core/kernels/one_hot_op.h"
 #include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/framework/types.h"
 
 namespace tensorflow {
 
 typedef Eigen::GpuDevice GPUDevice;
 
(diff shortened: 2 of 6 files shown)


girving:
This is not a good change, since it makes type errors silently work with strange behavior.  What are you trying to do?

girving:
For example, if you pass some floats in as indices, it is a type error.  Floor is not what the user expects.

samjabrahams:
Thanks, good to know.

The issue, as seen [here](https://github.com/tensorflow/tensorflow/issues/1799#issuecomment-215913080), is that when a tensor with a defined data-type is passed to `convert_to_tensor` with a different datatype, we get a ValueError. For example:

```python
>>> import tensorflow as tf
>>> a = tf.constant(3, dtype=tf.int32)
>>> b = tf.convert_to_tensor(a, dtype=tf.int64)

  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/framework/ops.py", line 565, in convert_to_tensor
    ret = conversion_func(value, dtype=dtype, name=name, as_ref=as_ref)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/framework/ops.py", line 510, in _TensorTensorConversionFunction
    % (dtype.name, t.dtype.name, str(t)))
ValueError: Tensor conversion requested dtype int64 for Tensor with dtype int32: 'Tensor("Const_2:0", shape=(), dtype=int32)'
```

You get similar results with Tensors of different data types. Numpy arrays work fine, however. I'm not sure whether this is intended behavior for `convert_to_tensor`. 

I understand the primary use case for `one_hot` will be when reading in data from a file, so having to convert TensorFlow objects isn't crucial, but the thought was to support as many data types as possible.

girving:
I believe this should be fixed by making `one_hot` natively work for `int32` tensors and then removing the `int64` default.

samjabrahams:
That sounds good. I guess I'll try to dive into this one.

girving:
Thanks!  Let me know if you have questions.

ebrevdo:
@girving assigning to you for now.  i can pick up after may 10th if it's still open.

samjabrahams:
Sorry it took a while to get something posted! I've pushed the WIP, and the code is working as intended. Not sure if this is the most elegant solution, however. I have a couple of questions:

* Is there a better pattern to iterate over combinations of data types when using `REGISTER_KERNEL_BUILDER`? I'm wondering if we should just extend `indices` compatibility to be of any `int` type. Ideally, I'd like to use both `TF_CALL_NUMBER_TYPES` and `TF_CALL_INTEGRAL_TYPES` (or modified versions of those macros), but they aren't designed for builders with multiple generic types. For example, in [lines 116-133](https://github.com/tensorflow/tensorflow/pull/2187/files#diff-9069716f50988af959734fd72f49c31fR116), I declared nearly identical code in order to register an implementation for both `int32` and `int64`. I could copy and paste a bunch for `int8`, `int16`, etc., but I figured I would ask first.
* Am I going overboard with the tests? I modified them to iterate over NumPy dtypes and TensorFlow dtypes, as we weren't testing `tf.constant` objects before ([which gave us surprise errors](https://github.com/tensorflow/tensorflow/issues/1799#issuecomment-215913080)). However, this jumps the test time from ~2.5 seconds to ~5 seconds for CPU-only, and from ~4.5 seconds to ~9 seconds with GPU.

Thanks!

tensorflow/core/kernels/one_hot_op.cc
```diff
@@ -39,7 +39,7 @@ namespace tensorflow {
 typedef Eigen::ThreadPoolDevice CPUDevice;
 typedef Eigen::GpuDevice GPUDevice;
 
-template <typename Device, typename T>
+template <typename Device, typename T, typename I>
```
ebrevdo: typename TI instead of typename I.  trust me, it'll be easier to read.

tensorflow/core/kernels/one_hot_op.cc
```diff
@@ -113,45 +113,77 @@ class OneHotOp : public OpKernel {
...
-TF_CALL_NUMBER_TYPES(REGISTER_ONE_HOT);
+#define REGISTER_ONE_HOT_64(type)                        \
+  REGISTER_KERNEL_BUILDER(Name("OneHot")                 \
+                              .Device(DEVICE_CPU)        \
+                              .TypeConstraint<int64>("I")\
+                              .TypeConstraint<type>("T") \
+                              .HostMemory("depth"),      \
+                          OneHotOp<CPUDevice, type, int64>);
+
+TF_CALL_NUMBER_TYPES(REGISTER_ONE_HOT_32);
```
ebrevdo: i wonder why this is not just TF_CALL_ALL_TYPES.  one_hot should work OK with strings too

tensorflow/core/kernels/one_hot_op.cc
```diff
@@ -113,45 +113,77 @@ class OneHotOp : public OpKernel {
   TF_DISALLOW_COPY_AND_ASSIGN(OneHotOp);
 };
 
-#define REGISTER_ONE_HOT(type)                           \
+#define REGISTER_ONE_HOT_32(type)                        \
```
ebrevdo: you can always just have both int32 and int64 definitions inside a single "#define".

you can also "#define REGISTER_ONE_HOT_INDEX(type, index_type) ...." then "#define REGISTER_ONE_HOT(type) REGISTER_ONE_HOT_INDEX(type, int32); REGISTER_ONE_HOT_INDEX(type, int64)

tensorflow/core/ops/array_ops.cc
```diff
@@ -1733,13 +1733,14 @@ dimension be equal to sizeof(`type`)/sizeof(`T`). The shape then goes from
 )doc");
 
 REGISTER_OP("OneHot")
-    .Input("indices: int64")
+    .Input("indices: I")
```
ebrevdo: indices: TI

tensorflow/python/kernel_tests/one_hot_op_test.py
```diff
@@ -45,62 +45,74 @@ def _testBothOneHot(self, truth, expected_err_re=None, raises=None, **inputs):
     self._testOneHot(truth, False, expected_err_re, raises, **inputs)
 
   def _testBasic(self, dtype):
-    indices = np.asarray([0, 2, -1, 1], dtype=np.int64)
```
ebrevdo: i dont think you have to modify all these tests.  just add one new test where the indices dtype is an int32.  (say, take the most complicated of the existing tests and copy it in a new method)

tensorflow/python/ops/array_ops.py
```diff
@@ -1917,7 +1917,6 @@ def one_hot(indices, depth, on_value=1, off_value=0,
             axis, dtype], name, "one_hot") as name:
     on_value = ops.convert_to_tensor(on_value, dtype=dtype, name="on_value")
```
ebrevdo: these should all be cast instead of ops.convert_to_tensor

girving:
@ebrevdo: Assigning to you since you seem to be looking over it already.  Let me know if you want me to take over, but one reviewer seems sufficient.

tensorflow/core/kernels/one_hot_op.cc
samjabrahams: Will do- should I change `typename T` to `typename TV` for the values, as well?

tensorflow/core/kernels/one_hot_op.cc
samjabrahams: I had similar thoughts, but didn't want to change it for this first pass. I'll try to get strings working with the next push.

tensorflow/python/ops/array_ops.py
samjabrahams: @ebrevdo When these are switched to `cast`, we end up with an `UnimplementedError` when trying to cast standard Python numbers to `complex64`, while `convert_to_tensor` has code coverage for this. Might we stick with `convert_to_tensor` for now and switch after the appropriate `cast` implementations are covered? From what I understand, the main downside of `convert_to_tensor` is that it may not work properly when `on_value` and `off_value` are passed in as Tensor objects, but I imagine most people just passing in normal Python types.

ebrevdo: No. They should with with tensors.  Sounds like a bug in cast.  Let's fix
that either by adding support for cast integers to complex or remove
dtype=complex from the test.

samjabrahams: Ok, I'll comment out complex dtype tests for now, and look to fix cast in another PR.

tensorflow/core/kernels/one_hot_op.cc
ebrevdo: T and TI are more typical, but TV and TI would be fine too.

samjabrahams:
I've pushed the next WIP commit. Here's the summary:

* Refactored the various kernel registration macros
* Changed generic typename `I` to `TI`
* Added support for int8, int16, int32, and uint8 for the `indices` parameter
* Separated out test for various indices types into `testIndicesTypes`
* Added base test for string support - _not passing_
* Adjusted Python wrapper to maintain backwards compatibility. If `dtype` is not passed in, `one_hot`'s output will now take on the dtype of `on_value` and/or `off_value` (instead of converting to `float32`).
    * If none of `on_value`, `off_value`, or `dtype` is provided, the output will be of type `float32`

---

I'm having trouble getting strings to work properly. In [the current, simple string test](https://github.com/tensorflow/tensorflow/pull/2187/files#diff-305762e23976349d47a199ab48ded4e4R309), I get the following error:

```
AssertionError: Shape mismatch: expected (3, 3), got ().
```

@ebrevdo Thoughts, when you get the chance?

ebrevdo:
The test failure with the string test: truth should not be a tf.constant but a np.array([[b"1.0", b"0.0", ...]], ...)

samjabrahams:
Sweet! Everything appears to be in order, now. Just had to add an extra check inside of `_testOneHot` to make sure it didn't try to run `assertAllClose` on a string tensor. Let me know if there's anything else that looks fishy.

tensorflow/core/kernels/one_hot_op.cc
```diff
@@ -113,44 +113,68 @@ class OneHotOp : public OpKernel {
...
+  template <>                                                               \
+  void OneHot<GPUDevice, T, TI>::Compute(                                   \
+      const GPUDevice& d, const typename TTypes<TI>::ConstMatrix& indices,  \
+      const typename TTypes<T>::ConstScalar& on_value,                      \
+      const typename TTypes<T>::ConstScalar& off_value,                     \
+      typename TTypes<T, 3>::Tensor* output);                               \
+  extern template struct OneHot<GPUDevice, T, TI>;
+
+#define DECLARE_GPU_SPEC(T)          \
+  DECLARE_GPU_SPEC_INDEX(T, int8);   \
```
ebrevdo: we only need int32 and int64.  no one uses indexing with other types and this bloats the C++ objects (every instantiation of a template creates a complete copy of the code, * all the types)

tensorflow/core/kernels/one_hot_op_gpu.cu.cc
```diff
@@ -21,17 +21,26 @@ limitations under the License.
...
-#define DEFINE_GPU_SPEC(T)                   \
-  template class generator::OneGenerator<T>; \
-  template struct functor::OneHot<GPUDevice, T>;
+#define DEFINE_GPU_SPEC_INDEX(T, TI)                   \
+  template class generator::OneGenerator<T, TI>;       \
+  template struct functor::OneHot<GPUDevice, T, TI>;
+
+#define DEFINE_GPU_SPEC(T)           \
+  DEFINE_GPU_SPEC_INDEX(T, int8);    \
+  DEFINE_GPU_SPEC_INDEX(T, int16);   \
```
ebrevdo: ditto

tensorflow/core/ops/array_ops.cc
```diff
@@ -1733,13 +1733,14 @@ dimension be equal to sizeof(`type`)/sizeof(`T`). The shape then goes from
...
 REGISTER_OP("OneHot")
-    .Input("indices: int64")
+    .Input("indices: TI")
     .Input("depth: int32")
     .Input("on_value: T")
     .Input("off_value: T")
     .Attr("axis: int = -1")
     .Output("output: T")
     .Attr("T: type")
+    .Attr("TI: {int8, int16, int32, int64, uint8}")
```
ebrevdo: TI: {int32, int64}

tensorflow/python/ops/array_ops.py
```diff
@@ -1817,8 +1817,8 @@ def _DepthToSpaceShape(op):
       [input_shape[0], height, width, new_depth])]
 
 
-def one_hot(indices, depth, on_value=1, off_value=0,
-            axis=None, dtype=dtypes.float32, name=None):
+def one_hot(indices, depth, on_value=1.0, off_value=0.0,
```
ebrevdo: set default on_value and off_value = None; that makes the checks below much easier.

tensorflow/python/ops/array_ops.py
```diff
@@ -1907,18 +1907,35 @@ def one_hot(indices, depth, on_value=1, off_value=0,
     output: The one-hot tensor.
 
   Raises:
-    TypeError: If dtype is `tf.string`
+    TypeError: If dtype of provided `on_value` and `off_value` don't match
   """
-  # Check for bad dtype specification
-  if dtype == dtypes.string:
-    raise TypeError("dtype must be a numeric type")
-
```
ebrevdo: first, check if dtype is None [and on_value and off_value are None].  if so, set dtype to float32.

next you can check "if on_value is None" and set it to 1.0 via convert_to_tensor; same with off_value.

then check to make sure their dtypes match.

samjabrahams:
Alright, I've pushed the next patch. Here's what's included:

* Altered handling of `on_value`, `off_value`, and `dtype` in array_ops.py to match requested implementation
* Removed int8, int16, uint8 registration
* Modified default value tests to match how dtype handles mismatches
* Add tests to assert TypeErrors when `on_value` and `off_value` are of different types, or when either of their types differs from `dtype`
* Added a documentation to describe new behavior and provide example of default value use

ebrevdo:
Jenkins, test this please.

tensorflow/python/kernel_tests/one_hot_op_test.py
```diff
@@ -34,7 +34,8 @@ def _testOneHot(self, truth, use_gpu=False, expected_err_re=None,
         ans = tf.one_hot(**inputs)
         if expected_err_re is None:
           tf_ans = ans.eval()
-          self.assertAllClose(tf_ans, truth, atol=1e-10)
+          if inputs.get('dtype') is not tf.string:
+            self.assertAllClose(tf_ans, truth, atol=1e-10)
```
ebrevdo: i think one_hot should probably return exact values.  have you tried converting this to self.assertAllEqual, including for strings?

tensorflow/python/kernel_tests/one_hot_op_test.py
```diff
@@ -214,45 +213,47 @@ def _testTypeBatch(self, dtype):
...
+        indices=indices,
+        on_value=on_value,
+        off_value=off_value,
+        depth=depth,
+        dtype=dtype,
+        truth=truth)
 
     # axis == 1
     self._testBothOneHot(
-            indices=indices,
```
ebrevdo: can you revert the changes that are only whitespace changes?

tensorflow/python/ops/array_ops.py
```diff
@@ -1908,20 +1938,51 @@ def one_hot(indices, depth, on_value=1, off_value=0,
...
+    if not off_exists:
+      # off_value not provided: assign to value 0 of type dtype
+      off_value = ops.convert_to_tensor(0, dtype, name="off_value")
+      off_dtype = dtype
+
+    if on_dtype != off_dtype:
+      raise TypeError("dtype {0} of on_value does not match " \
+                      "dtype {1} of off_value".format(on_dtype, off_dtype))
+
+    depth = gen_math_ops.cast(depth, dtypes.int32, name="depth")
```
ebrevdo: do you still need this cast?

samjabrahams:
Thank you @ebrevdo for your continued notes. How does this most recent patch look? Here are the highlights:

* Changed `assertAllClose` to `assertAllEqual`, and removed the check for string data-type. The `assertAllClose` check was from the initial implementation of the Op, but it's a good catch to change it.
* All white space changes should be gone
* Removed the `cast` on depth in the Python wrapper. Letting depth use the same generic type as indices, TI, doesn't work with the backwards compatibility check. But you're right in that the cast is unnecessary in general
* Added a default value of `DT_INT64` for the generic type TI- this was needed for the test to pass the backwards compatibility test

Let me know if you have anything else!

ebrevdo:
Uh-oh.  We can't break the backwards compatibility.  What this means is folks who have created a graph with a previous version of this op will not be able to run it on a more recently compiled version of TF.  That's a big no-no.  Sorry for suggesting it.  Can we revert the change on dtype of depth?

samjabrahams:
I think I miscommunicated what I did (or I've done goofed)- the latest patch was made to _maintain_ backwards compatibility. We couldn't implement the suggestion to pass `TI` as the type of `depth` without breaking backwards compatibility, so I've already reverted it. The rest of that bullet indicates that I simply removed the `depth = tf.cast(...)` line in the Python wrapper, as it A. was unnecessary and B. allowed for bad `depth` inputs to work when they shouldn't (as both you and @girving pointed out). I also registered the required default dtype for `TI` so that it matches the `int64` type used by `indices` previously.

Let me know if I've done goofed, though :)

ebrevdo:
Jenkins, test this please.

ebrevdo:
Jenkins, test this please.

ebrevdo:
Thank you for bearing with me!

samjabrahams:
Absolutely! Thank you for your guidance with this PR- I've definitely learned a ton from this process.

vrv:
@ebrevdo next time squash before merging

Issue #37, opened 2015-11-09
Node.js (JavaScript) Wrapper API


keon:
Because JavaScript is Awesome

ZECTBynmo:
+1!

ripper2hl:
+1 ! :+1:

dhritzkiv:
+1

fiws:
:+1:

miguelalche:
+1

jagandecapri:
Just what I was searching for. :+1: 

As quoted from the offical website http://www.tensorflow.org/
>we’re hoping to entice you to contribute SWIG interfaces to your favorite language -- be it Go, Java, Lua, Javascript, or R.

I am new to this whole SWIG thing but searched around and found this. http://www.swig.org/Doc3.0/Javascript.html

Not really sure how this works. Do we need to write swig interface file specifically for Javascript or is it auto-generated when running some commands or is somebody already working on this (this would be awesome) ?

imgntn:
+1 :+1:

prathamesh7pute:
+1

keonkim:
@jagandecapri i am also not an expert on wrapping things with javascript, but i am willing to contribute!  (but might take some time)

Luna4:
+1

aki-russia:
+1

marcbaetica:
+1

lowe0292:
👍

mattkosoy:
:+1:

nikhilk:
+1!

Just starting out on one, but new to writing a nodejs addon. Checking out the swig interface files to see if they're going to be helpful, or if I should just use the c++ API.

vnglst:
+1

jalona:
+1

voodooattack:
+1

Foorack:
+1

tngan:
+1

vincentvanhoucke:
This is something the core TensorFlow team is unlikely to tackle in the near future, so if you want to contribute it, please go ahead! I would recommend circulating a proposed implementation on the discuss mailing list early on, so that a consensus about where such API might live (in repo / off repo / in 'contrib' directory) can be reached ahead of time.

Foorack:
Anyone up to code a NodeJSlibrary? :+1:

augbog:
:+1:

tngan:
[...]

PR #2206, opened 2016-05-03
[WIP] Initial proof-of-concept NodeJS binding.

peterbraden:
Retrieves the tensorflow versions string.

Test in the contrib/nodejs directory with:

```
npm run generate && npm install && npm test
```


danmane:
@ry would you be open to reviewing this?

ry:
@danmane yea sure. I'll check it out tomorrow

danmane:
Thanks!

kenokabe:
True progress! thanks @peterbraden

peterbraden:
I'm working on a further test, which will actually call some tensorflow code, but I have a bunch of outstanding problems, which I'm documenting in this gist: https://gist.github.com/peterbraden/1f6e52045c27cfcb8a5c13b1ef82c94f - not sure the best way to collaborate on this, but if anyone has any suggestions or workarounds, please let me know.

ry:
@danmane @peterbraden it looks fine to me as a proof-of-concept... I was able to build an run the test on ubuntu. This might be better as a separate project outside the TF tree? Also I'd have some style nitpicks like in `test/suite.js`.

peterbraden:
@ry Thanks for the review, I did this as part of contrib as a response to this comment: https://github.com/tensorflow/tensorflow/issues/37#issuecomment-216272659

Would love to hear your style nitpicks too.

I'm continuing work in this branch: https://github.com/tensorflow/tensorflow/compare/master...peterbraden:nodejs-wip - I think that until I have a useful subset of functionality, this isn't ready for merging.

Would welcome all feedback or collaboration though. I think we need to wait for this too: https://github.com/nodejs/node-gyp/pull/919

tensorflow/contrib/nodejs/test/suite.js
```diff
@@ -0,0 +1,7 @@
+var test = require('tape')
+  , tensorflow = require('../index')
```
ry: Don't use comma first style, add semi-colons, add space between `(t)` and `{` in the function

I assume TF will want to follow this style guide https://google.github.io/styleguide/javascriptguide.xml ?

danmane: For Python, we are generally following Google style guide, but are using the naming conventions from PEP8, to be more consistent with open-source users. So Google style guide is a good default, but if there are reasons to use a different style guide that's more familiar or widely adopted in the nodejs community I think we'd be open to that - especially since I doubt think the node bindings will see much use inside of Google. 

If you think the google js style guide is a good choice then let's go with that.

@martinwicke feel free to comment more authoritatively... (since I know you have such passionate, passionate thoughts on javascript styles)

martinwicke: Let's try for the google style guide. As far as I know, nothing resembling
a standard has evolved in the JS community.

billautomata:
@martinwicke https://www.npmjs.com/package/standard this has evolved as a standard in the node community, powerful linter and node js focused static analysis

martinwicke:
@danmane how different is this from google style? I'm happy to use a standard if there is one.

peterbraden:
I'm not so interested in specific code lint styles, I can defer to people there. 

I've been working in this branch https://github.com/tensorflow/tensorflow/compare/master...peterbraden:nodejs-wip - I'm not sure the best way to have transparency here, should I add to this pull-request, even though it's not ready to be merged, and have a long-running PR, or should I wait until I have something more complete? I don't have a huge amount of time to work on this, and would love to attract some collaborators.

martinwicke:
You can make a PR, add [WIP] to the title and reference this thread in the
description.

danmane:
@martinwicke @billautomata
There are a few notable diffs, e.g. [googlejs](https://google.github.io/styleguide/javascriptguide.xml) calls for semicolons always, standard calls for no semicolons at all. But given that node is basically unused within Google, I see no benefit in tying the node frontend to Google's style guide. And it looks like standard has pretty good adoption based on the download stats.  So let's go with package/standard.

martinwicke:
Wow. No semicolons ever? That's insane. I would have expected semicolons to
be part of any style guide for JS given how brittle JS is around line
breaks.

Anyway. package/standard it is.

ry:
Node itself uses a style similar (same?) to Google's JS style... package/standard seems rather non-standard despite its name.
https://github.com/nodejs/node/blob/395cc885f4eb21c2002dbfe405d31da275699aa8/lib/path.js

camiloazula:
I like the approach of http://standardjs.com/

rmlarsen:
Any progress on this?

guymguym:
Hey
Using google's eslint config is pretty easy and provides good defaults.
It's also easy to override rules from the base.
Quick start - ready for copy-paste in terminal:
```
npm install --save eslint eslint-config-google
cat << EOF >.eslintrc.js
module.exports = {

    // extending google's eslint config as base
    extends: 'google',

    rules: {
        // overriding rules from base
    }
};
EOF
```

danmane:
@peterbraden what should the next steps be with this PR? Is this in a state where it should be merged into the TensorFlow repo? You mentioned wanting to attract collaborators; that doesn't seem to have happened so far. 

We're trying to control the number of open pull requests so things are easier to manage, so if we don't have a story on merging this code we may close it due to inactivity.

guymguym:
@peterbraden @danmane 

I am not sure how this project is being managed, but in my opinion it would be best to merge this as a beta feature rather than closing, otherwise how would you ever get enough collaborators to try it out and contribute issues/fixes?

I would love to help and contribute, but not sure what are the gaps, so if you have some breakdown it would be helpful.

Perhaps contributing tests would be a good starting point?

BTW regarding https://github.com/feross/standard, while it is popular it's also very opinionated, so unless you feel very opinionated about specific coding style, you are probably better of using eslint with https://github.com/google/eslint-config-google as base which provides great starting point, and can be adapted to specific needs. But that's not something to stop from merging, and can be applied later as well.

chrisber:
My guess is that  tensorflow is currently developing a c-api which then can be used with swig to build wrappers for multiple languages.
Take a look at: https://github.com/tensorflow/tensorflow/commits/master/tensorflow/c/c_api.h

martinwicke:
That is correct -- and while the C API in 0.10.0RC0 is almost there, it's
not quite done.

peterbraden:
@danmane The next steps would be to start using SWIG to wrap the appropriate features. I was working in this branch: https://github.com/tensorflow/tensorflow/compare/master...peterbraden:nodejs-wip

I don't really have much time to work on this at present. If you need to close the PR that's fine, but if somebody continues the attempt in the future, then this will be a good starting point, so some way of pointing them to this would be good.

rmlarsen:
@peterbraden @danmane Should we close the PR or perhaps leave it open with a contributions welcome tag? Or would you prefer to open an issue with a reference to the closed PR?

peterbraden:
Either work for me. Thanks!

guymguym:
Hey, just note that SWIG might be good for setting up TF sessions from node.js code, but I think that running a session doesn't make sense by SWIG function call, since it will block the node.js event loop from processing events for the entire session run time.

The run call can use async workers:
https://github.com/nodejs/nan/blob/master/doc/asyncworker.md

Or a similar model for offloading the work with threads away from the main event loop:
http://docs.libuv.org/en/v1.x/threading.html

jhseu:
As mentioned earlier in the thread, we're discouraging new uses of the C++ API through SWIG. Any language bindings should now go through the C API, which is pretty close to feature-complete with the C++ API:
https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/c/c_api.h

You can see an in-progress binding that uses the C API here:
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/go

martinwicke:
I'm closing this. We'd like to not use SWIG going forward (and move to the C API instead). 

Re #37: We're realistically not able to support node.js bindings at the moment and therefore currently see no path for this to get merged out of contrib. This would be more suitable for a separate repo.

Issue #1661, opened 2016-03-26
conv4d and higher dimension generalization


jerabaul29:
Are there some plans to include a 4D (and possibly higher dimension generalization) convolution op? Would be nice for doing 3D + time processing for example (for the 4D version).

Also, should something be done to offer the user the possibility to choose the shape of the kernel? For example allowing both the natural hyper cube kernel nodes location or a hyper thetrahedron location? Could have consequences on the computing power requested:

http://arxiv.org/pdf/1505.02890v2.pdf

keveman:
@jerabaul29, generalizing convolutions to higher dimensions would be a great addition to TensorFlow, especially the GPU implementation. It will be great if someone in the community can take this up.

bhack:
I think that this impact Eigen upstream

akors:
I'm no expert, but I think that Eigen supports tensors with arbitrary dimensions (albeit compilation gets really slow when it's more than 8 or something), and judging by the cuDNN manual, it can also take 5D Tensors (as in 3D+1D-color+1D-batchsize) or maybe even higher-dimensional.

The code in tensorflow/core/kernels/conv_ops.cc looks fairly generic, and might even be extensible to include higher-dimensional and GPU-accelerated convolutions.

I'm afraid it would require a serious amount of digging, though.

vrv:
https://github.com/tensorflow/tensorflow/commit/6a187ccddaebb741ea77fc3201c6e36625f0aadb for an example of adding 3d convolutions, which uses cudnn Nd tensor interface and eigen for CPU.  Something similar could be done for conv4d, conv1d, etc.

If someone wants to tackle this, please let us know (cc @rewonc), since we now have conv2d and conv3d to provide a template for generalization.  We may still be missing something though, so no guarantees it will be straightforward.

rewonc:
@vrv Cool -- definitely seems like there's an opportunity to generalize a template. Will take a look and let you

mhuen:
Are there any news on this?
A 4D operation would be a really nice feature. Especially now that more and more people from other fields, where 4D (3D + time)  is the common use case (i.e. physics), are jumping in on deep learning.
Even if the n-dimensional generalization is not easily implementable, I think a separate conv4d would be very helpful.

mayank1123:
I know it hasn't been long but is there any news on this ? We are currently using Caffe for 4D light-fields convolutions but would love to shift to tensorflow.

yaroslavvb:
[...]

Issue #1136, opened 2016-02-17
Support for 1D convolutions


ghost:
There is no conv1d method and it would be very welcome.

Or is conv2d implemented in such a way that (h=1, w, channels) would work with no issues and no compute penalties?

It might even be a good idea to implement an arbitrary convxd function, there are instances where high dimensional inputs have spatial relevance.

vrv:
Yes, conv2d can work with h=1, though it's possible one could write a faster kernel if you knew it was 1d.  If someone was willing to write a kernel (CPU and GPU) and benchmarks for this, we'd be happy to accept it.  Same with convxd.  #150 for a similar request, but for 3D, which we'd also love contributions for.

alphaf52:
conv2d can work with h=1 indeed. However, when i want to set stride > 1 for 1d conv using conv2d with h = 1, i got an error "ValueError: Current implementation only supports equal length strides in the row and column dimensions". And as i set h = 1, i can't have a stride wrt h which is greater than 1. I want to know if there is a way i can do this.

jramapuram:
Fractional striding on 1d convolutions would be helpful to downsample 1d signals. Currently there is no way to do this with 2d convolutions as mentioned by @alphaf52 .

girving:
@jramapuram Fractional striding is hard to implement efficiently using anything like direct convolution; normally I'd just write it using FFTs (which already works).

vrv:
@alphaf52 I believe conv2d now supports unequal strides in the h and w dimensions.  separable/depthwise don't yet, but for most people's purposes, that should be addressed.

renesax14:
are there examples or docs that explain how to actually use the 1D convolution?

renesax14:
example: http://stackoverflow.com/questions/38114534/basic-1d-convolution-in-tensorflow/38117279#comment65079952_38117279

shoyer:
I believe this is actually done. See:
https://www.tensorflow.org/versions/r0.11/api_docs/python/nn.html#conv1d

renesax14:
@shoyer I'm pretty sure thats new, and it seems it doesn't have BN automatically included in it like the other convolution, which is a big draw back.

shoyer:
@renesax14 sorry, "BN"?

renesax14:
sorry, I didn't even realize I wrote the acronym. Batch Normalization.

shoyer:
[...]

Issue #150, opened 2015-11-11
Is there 3D ConvNets support ? 


sasadep:
(Is there) **or** (will there be) a **3D ConvNets** support in **tensorflow** ?  
Which helps for spatiotemporal features.

vrv:
We don't have a conv3d op yet, but there will eventually be!   We'll use this issue to track this feature request.

sasadep:
:+1:

PatrickChrist:
Theano offers serveral 3D Convolutional Implementations on CUDA. https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/GpuConv3D.py
https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/GpuConvGrad3D.py
https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/fftconv.py 
I would really appreciate, if you guys can port this stuff to tensorflow.

edward339:
Any word on this? Would be very useful capability.

bhack:
:+1:

daeyun:
What is a good way to support 3D convolution in an existing code base that already uses tensorflow? I'm expecting to have to just port to a different library, but wanted to ask in case there's an easier way. I would want GPU support, but less optimized is fine.

omarcr:
+1 on this request.

noureldien:
Please, support 3D convnets. Thanks for help.

bhack:
Ops are arriving with https://github.com/tensorflow/tensorflow/pull/1571/commits/784778e76d3f5fd75b9ec99138770795907f1616

jerabaul29:
Great!

I guess a problem with 3 (or higher) D convnets will be the cost of computations. Do you think it would be possible / useful (or maybe it is already possible in your ops?) to implement a tetrahedral instead of square grid convnet as suggested in this article? Interesting also for later generalization to higher dimension: if I understand well, this approach (generalized if necessary to higher dimension) will make the number of nodes in the kernel linear instead of exponential in the dimension of the initial tensor.

http://arxiv.org/pdf/1505.02890v2.pdf

jerabaul29:
Can the people at the origin of the present enhancement look at the following one:

https://github.com/tensorflow/tensorflow/issues/1661

The generalization asked may (?) be easier to implement by the same people who have already done it in 3D...

akors:
[...]

PR #2213, opened 2016-05-04
Add support for 3d convolutions

rewonc:
This patch is a WIP, but adds functionality for 3d convolutions as requested in #150.
### Approach

This patch adds 3d kernels and glue code, largely based off the 2d kernels, as separate modules. We add `conv_3d`, `conv3d_ops`, `conv3d_grad_ops`, and others to the original `conv_2d`, `conv_ops`, and `conv_grad_ops`, which remain unchanged.

As there is logic in the 2d kernels which the 3d kernel simply replicates, this patch is quite large. Generalizing the conv operator to n dimensions might solve 1d and 4d use cases (#1661 and #1136) as well as reducing the overall quantity of code in the codebase.  As this patch includes a comprehensive test suite, this patch should help if or when that happens. 
### Progress
- [x] Test suite for GPU/CPU
- [x] CPU kernel implementation
- [ ] GPU kernel implementation 

Currently, there are ~40 tests in `tensorflow/python/kernel_tests/conv3d_ops_test.py` which test shapes, outputs, strides, and gradients for a variety of input. They pass on my CPU instance: 

image

I currently lack access to a GPU so I haven't worked on the GPU kernel yet, but hopefully that should happen soon.
### Todo

This is a learning project for me, so I'd love to get feedback on how this might become more helpful, or how it might fit in better with the TF team's architectural design.  Thanks!

### DIFF
--- a/tensorflow/core/kernels/BUILD
+++ b/tensorflow/core/kernels/BUILD
@@ -72,6 +72,16 @@ cc_library(
 )
 
 cc_library(
+    name = "conv_3d",
+    hdrs = ["conv_3d.h"],
+    deps = [
+        ":eigen_helpers",
+        "//tensorflow/core:framework",
+        "//third_party/eigen3",
+    ],
+)
+
+cc_library(
     name = "fill_functor",
     hdrs = ["fill_functor.h"],
     deps = [
@@ -1103,6 +1113,21 @@ tf_kernel_library(
 )
 
 tf_kernel_library(
+    name = "conv3d_ops",
+    srcs = ["conv3d_grad_ops.cc"],
+    prefix = "conv3d_ops",
+    deps = [
+        ":bounds_check",
+        ":conv_3d",
+        ":ops_util",
+        "//tensorflow/core:core_cpu",
+        "//tensorflow/core:framework",
+        "//tensorflow/core:lib",
+        "//tensorflow/core:nn_ops_op_lib",
+    ],
+)
+
+tf_kernel_library(
     name = "depthwise_conv_op",
     prefix = "depthwise_conv_op",
     deps = [
@@ -1148,6 +1173,8 @@ tf_kernel_libraries(
         ":bounds_check",
         ":conv_2d",
         ":conv_ops",
+        ":conv_3d",
+        ":conv3d_ops",
         ":depthwise_conv_grad_op",
         ":depthwise_conv_op",
         ":ops_util",
--- a/tensorflow/core/kernels/conv3d_grad_ops.cc
+++ b/tensorflow/core/kernels/conv3d_grad_ops.cc
@@ -1,0 +1,292 @@
+/* Copyright 2015 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+// See docs in ../ops/nn_ops.cc.
+
+#define USE_EIGEN_TENSOR
+#define EIGEN_USE_THREADS
+
+#include <vector>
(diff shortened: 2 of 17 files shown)


zheng-xq:
A 3D convolution is almost done internally. I'll add the internal author to comment on the progress.

vrv:
Yeah it's unfortunate that you went through all this effort since it's probably arriving with the next push (including GPU support).  I would have suggested mentioning you were working on it earlier! :(

vrv:
Also, our fault for not indicating on that bug that it was being worked on internally. It came together pretty quickly in the last week or so and we forgot to update this

rewonc:
Oh bummer! Well was still valuable to learn about it. What about 1D and 4D?

vrv:
Sure!  conv1d and conv4d would be nice!  Once the conv3d implementation is available you can generally see how to do it for both CPU and GPU -- there's opportunities for lots of code sharing across all of the ops then.

vrv:
3d convolutions added in https://github.com/tensorflow/tensorflow/commit/6a187ccddaebb741ea77fc3201c6e36625f0aadb, you can take a look to see how it was done and maybe look for opportunities for code sharing with something like conv1d if you think it's worth doing!

carrasqu:
1d convolution would be good to have. Also I wonder if anyone would be interested in having different boundary conditions in all types of convolutions. for instance I needed to have periodic boundary conditions where the pixels on the edges are connected to pixels on opposite edges, and so on. I had to modify my training/test sets in order to account for that.

zhongying811:
Does GPU kernel of 3d convolutions have been implemented?

zheng-xq:
@zhongying811, yes. It is tf.nn.conv3d

zhongying811:
Who has some 3d convolution(3DCNN) example codes on GPU.
I'm a beginner. Thanks.

akors:
@zhongying811 The code on GPU is the same as on CPU, where it runs only depends on the TensorFlow version installed.

I don't have a simple example, but basically, you can follow the [tutorial for deep MNIST](https://www.tensorflow.org/versions/master/tutorials/mnist/pros/index.html#deep-mnist-for-experts) for 2D data and go to 3D data, and replace

    tf.nn.conv2d(x, W, strides=[1, 1, 1, 1], padding='SAME')

with

    tf.nn.conv3d(x, W, strides=[1, 1, 1, 1, 1], padding='SAME')

and

    tf.nn.max_pool(x, ksize=[1, 2, 2, 1], strides=[1, 2, 2, 1], padding='SAME')

with

    tf.nn.max_pool3d(x, ksize=[1, 2, 2, 2, 1], strides=[1, 2, 2, 2, 1], padding='SAME')

It is really quite similar.

If you are a beginner, here are two good online courses about machine learning in general:

https://www.coursera.org/learn/machine-learning
https://www.udacity.com/course/machine-learning-engineer-nanodegree-by-google--nd009

I suggest that you ask any other questions on [StackOverflow](http://stackoverflow.com/) (using the tensorflow tag).

zhongying811:
@akors Thanks a lot. And the online courses are just what I need.

wolfbill:
Hi,I'd like to get the demo of conv3d of tensorflow. But I can't find it  and I hope I can get one. Is anyone  familiar with it.

akors:
@wolfbill 

Here's an example of conv3d usage:

<script src="https://gist.github.com/akors/52bba57f724e38229da9d85fbc85b673.js"></script>

https://gist.github.com/akors/52bba57f724e38229da9d85fbc85b673

zhongying811:
Who has some 3d convolution or 2d convolution example codes with C++ interface.

wolfbill:
@akors Hi man. That's really happy to get your message. We can't  access gist.github.com in mainland.We can only access github.com。Ｃould you place your example on github？

wolfbill:
＠akors I get the file from url. Thank you.

sankethvedula:
Hi all,

Request for a feature: Is there a function called as separable_conv3d? 

I think it is more important in the case of 3D convolutions to be separate. 

Could anyone please let me know if it exists?

wookaka:
Hi all
I am trying to produce a batch for 3DCNN , then i try this code:

image_batch, label_batch = tf.train.batch([image, label],
                                          batch_size = 50,
                                          shapes = [[12,250,115,3], label.get_shape()])

raise ValueError("Shapes %s and %s are incompatible" % (self, other))
ValueError: Shapes (?, ?, 3) and (12, 250, 115, 3) are incompatible

then i got this error, have you seen such error?

PR #2214, opened 2016-05-04
Fix init value in inner_product

legrosbuffle:
inner_product accumulates into the type of the initial value
meaning that the result actually has type int instead of int64.
This will lead to truncation errors resulting in invalid writes
on buffers larger than 4GB.


ebrevdo:

ebrevdo:
@tensorflow-jenkins : test this please.

ebrevdo:
Test this please, Jenkins.

mrry:

PR #2221, opened 2016-05-04
API Doc: Moved the usual termination code into a finally block

qbx2:
Moved the usual termination code into a finally block so that the code can be executed even when SystemExit, KeyboardInterrupt, GeneratorExit raised (These exceptions derive not Exception but BaseException.)


qbx2:
I signed it!

petewarden:
Test this please, Jenkins.

PR #2222, opened 2016-05-04
Add port mapping in the setup documentation to make sure Jupyter notebook server port is exposed.

hsaputra:
Need to add `-p 8888:8888` to expose the port to host in the Docker documentation.

### DIFF
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -309,7 +309,7 @@ After Docker is installed, launch a Docker container with the TensorFlow binary
 image as follows.
 
 ```bash
-$ docker run -it gcr.io/tensorflow/tensorflow
+$ docker run -it -p 8888:8888 gcr.io/tensorflow/tensorflow
 ```
 
 If you're using a container with GPU support, some additional flags must be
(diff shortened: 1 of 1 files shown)


ilblackdragon:

Need to pass checks even for documentation change.

martinwicke:
Can you also add the flags to the docker_run script? And maybe add a sentence about what it's for?

hsaputra:
This PR is for updating documentation.
Will send another PR to update the script. 

Would that be ok?

martinwicke:
That would be ok. Can you add a sentence explaining why the -p argument is
necessary?

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -309,7 +309,7 @@ After Docker is installed, launch a Docker container with the TensorFlow binary
 image as follows.
```
ilblackdragon: how about adding here just simple to explain port, what it's for and that people can change it if they want to something else "(`-p 8888:8888` used to share internal port to the host machine, in this case to ensure to ensure Jupiter notebook connection)".

hsaputra: Sounds good. Will do

hsaputra: @martinwicke : thx, will do

hsaputra:
Update PR with description about the Docker -p option.

hsaputra:
Thx for merging, @martinwicke

PR #2230, opened 2016-05-05
Merge changes from the last several days

vrv:
This was a pretty hard conflict to resolve from internal to external, so we'll see if I did it right.


vrv:
Okay all tests passed, so merging (looks like my conflict resolution worked, as far as tests could determine).

Issue #2225, opened 2016-05-04
InvalidArgumentError: No OpKernel T=DT_INT32 for log, exp, etc.


hunkim:
### Environment info
Operating System: Mac/Ubuntu
Installed version of CUDA and cuDNN: None
commit b8883a237b71e877759327fb4b9077847d4cb16c

### Steps to reproduce
Thanks to http://stackoverflow.com/questions/37027762:
```
import tensorflow as tf
import numpy as np
x = np.array([1], dtype=np.int32)
sess = tf.Session()
print(sess.run(tf.log(x)))
```

### Logs or other output that would be helpful
(If logs are large, please upload as attachment).

https://www.tensorflow.org/versions/r0.8/api_docs/python/math_ops.html#log says "x: A Tensor. Must be one of the following types: float32, float64, int32, complex64, int64", but int32 and in64 throws InvalidArgumentError:

InvalidArgumentError: No OpKernel was registered to support Op 'Log' with these attrs
	 [[Node: Log_4 = Log[T=DT_INT32](Log_4/x)]].

I haven't checked all, but `exp`, `cos` and `sin` have the same issue.

I added a simple test:

```
--- a/tensorflow/python/ops/math_ops_test.py
+++ b/tensorflow/python/ops/math_ops_test.py
@@ -28,6 +28,21 @@ from tensorflow.python.platform import googletest
 exp = np.exp
 log = np.log

+class LogTest(test_util.TensorFlowTestCase):
+
+  def testLog(self):
+    x = np.array([1], dtype=np.int32)
+    with self.test_session():
+      y_tf = math_ops.log(x).eval()
+      self.assertEqual(y_tf, 0)

```
which clearly reveals this issue:

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
..E..........
======================================================================
ERROR: testLog (__main__.LogTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/hunkim/.cache/bazel/_bazel_hunkim/6785122ce03f8c29e79889062219c7d3/tensorflow/bazel-out/local-fastbuild/bin/tensorflow/python/math_ops_test.runfiles/tensorflow/python/ops/math_ops_test.py", line 44, in testLog
    y_tf = math_ops.log(x).eval()
[...]

PR #2227, opened 2016-05-05
Fix Issue  #2225 InvalidArgumentError: No OpKernel T=DT_INT32 for log, exp, etc.

hunkim:
This PR fixes  #2225,  InvalidArgumentError: No OpKernel T=DT_INT32 for log, exp, etc.
## Issue

https://www.tensorflow.org/versions/r0.8/api_docs/python/math_ops.html says, "A Tensor. Must be one of the following types: float32, float64, int32, complex64, int64", but code does not:

`REGISTER4(UnaryOp, CPU, "Cos", functor::cos, float, Eigen::half, double, complex64);`

They don't support int32 and int64. So this simple code fails:

```
x = np.array([1], dtype=np.int32)
sess = tf.Session()
print(sess.run(tf.log(x)))
```
## Solution

Thanks to a hint from @yaroslavvb, the patch is rather simple:

```
-REGISTER4(UnaryOp, CPU, "Log", functor::log, float, Eigen::half, double,        
-          complex64);       
+REGISTER6(UnaryOp, CPU, "Log", functor::log, float, Eigen::half, double, int32,
+         int64, complex64);
```
## Tests

I also added several tests to verify the patch. They all pass in my cpu-only machine:

```
...
//tensorflow/python:math_grad_test                                       PASSED in 2.8s
//tensorflow/python:math_ops_test                                        PASSED in 1.5s
//tensorflow/python:matmul_op_test                                       PASSED in 3.1s
...
//tensorflow/python:xent_op_test                                         PASSED in 1.4s

Executed 186 out of 187 tests: 187 tests pass.
```

If I just run math_ops_test without this patch, it fails with InvalidArgumentError: 

```
//tensorflow/python:math_ops_test                                        FAILED in 1.4s
  /home/hunkim/.cache/bazel/_bazel_hunkim/6785122ce03f8c29e79889062219c7d3/tensorflow/bazel-out/local-fastbuild/testlogs/tensorflow/python/math_ops_test/test.log

Executed 186 out of 187 tests: 186 tests pass and 1 fails locally.
```

PR #2244, opened 2016-05-06
Updated API docs to sync with supported data types in csiwe_op_*.

hunkim:
Based on excellent comments from @yaroslavvb and @zheng-xq, this PR changes API docs to sync with supported data types in csiwe_op_*.

This addresses #2225

### DIFF
--- a/tensorflow/core/ops/math_ops.cc
+++ b/tensorflow/core/ops/math_ops.cc
@@ -131,11 +131,11 @@ tf.complex_abs(x) ==> [5.25594902, 6.60492229]
 
 // Declares cwise unary operations signature: 't -> 't
 #define UNARY()                      \
-  Input("x: T").Output("y: T").Attr( \
-      "T: {half, float, double, int32, complex64, int64}")
+  Input("x: T").Output("y: T")
 
 REGISTER_OP("Neg")
     .UNARY()
+    .Attr("T: {half, float, double, int32, int64, complex64}")
     .Doc(R"doc(
 Computes numerical negative value element-wise.
 I.e., \\(y = -x\\).
@@ -143,6 +143,7 @@ I.e., \\(y = -x\\).
 
 REGISTER_OP("Inv")
     .UNARY()
+    .Attr("T: {half, float, double, complex64}")
     .Doc(R"doc(
 Computes the reciprocal of x element-wise.
 I.e., \\(y = 1 / x\\).
@@ -150,6 +151,7 @@ I.e., \\(y = 1 / x\\).
 
 REGISTER_OP("Square")
     .UNARY()
+    .Attr("T: {half, float, double, int32, int64, complex64}")
     .Doc(R"doc(
 Computes square of x element-wise.
 I.e., \\(y = x * x = x^2\\).
@@ -157,6 +159,7 @@ I.e., \\(y = x * x = x^2\\).
 
 REGISTER_OP("Sqrt")
     .UNARY()
+    .Attr("T: {half, float, double, complex64}")
     .Doc(R"doc(
 Computes square root of x element-wise.
 I.e., \\(y = \sqrt{x} = x^{1/2}\\).
@@ -164,6 +167,7 @@ I.e., \\(y = \sqrt{x} = x^{1/2}\\).
 
 REGISTER_OP("Rsqrt")
     .UNARY()
+    .Attr("T: {half, float, double, complex64}")
     .Doc(R"doc(
 Computes reciprocal of square root of x element-wise.
 I.e., \\(y = 1 / \sqrt{x}\\).
@@ -171,12 +175,14 @@ I.e., \\(y = 1 / \sqrt{x}\\).
 
 REGISTER_OP("Exp")
     .UNARY()
+    .Attr("T: {half, float, double, complex64}")
     .Doc(R"doc(
 Computes exponential of x element-wise.  \\(y = e^x\\).
 )doc");
 
 REGISTER_OP("Log")
     .UNARY()
+    .Attr("T: {half, float, double, complex64}")
     .Doc(R"doc(
 Computes natural logarithm of x element-wise.
 I.e., \\(y = \log_e x\\).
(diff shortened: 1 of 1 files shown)


martinwicke:
@girving Will this be a backward-compatibilty issue?

girving:
It will break the backwards compatibility test, since it reduces the list of datatypes the ops are registered for.  However, it's not otherwise a backwards compatibility issue if the removed ops never ran, and it's a good change to make since it stops us from lying.  We should arguably make an exception and do the necessary surgery to fix the test.

I believe the backwards compatibility test is open source, so @hunkim: am I correct that you haven't tried running the unit tests yet with y our change?

girving:
@josh11b: If you agree, what's the best way to perform the necessary surgery to clarify that `tf.sqrt` has never worked on `int32`?

hunkim:
@girving, @martinwicke Not at all. Now this API doc is synced with actual implementation.

martinwicke:
This is not a doc file. This is the file that actually defined which
kernels are compiled.

girving:
@hunkim: I understand that your change is good, but by "Not at all" do you mean that you definitely haven't run the tests?  I believe they will fail, since we have a backwards compatibility check that registered ops only get more permissive (ignoring the kernels).

hunkim:
@girving I just generated API docs and it was successful. So I assumed it was OK. It was my bad.

I just ran the python tests. They all passed, but as you said //tensorflow/python:nn_test failed. 

`TypeError: DataType int32 for attr 'T' not in list of allowed values: float16, float32, float64, complex64`

How to update docs without breaking this? Currently, supported types in doc and code are different as discussed PR #2227.

> A Tensor. Must be one of the following types: float32, float64, int32, complex64, int64

`
REGISTER4(UnaryOp, CPU, "Cos", functor::cos, float, Eigen::half, double, complex64);`

tensorflow/core/ops/math_ops.cc
```diff
@@ -131,90 +131,102 @@ tf.complex_abs(x) ==> [5.25594902, 6.60492229]
 
 // Declares cwise unary operations signature: 't -> 't
 #define UNARY()                      \
-  Input("x: T").Output("y: T").Attr( \
-      "T: {half, float, double, int32, complex64, int64}")
+  Input("x: T").Output("y: T")
```
zheng-xq: I suggest to add another macro such as UNARY_FP around the list of types. So it is easier to add types in the future. 

{half, float, double, complex64}

zheng-xq:
Adding @girving to give guidance on how to deal with the backward compatibility test. 

In general, it is a good idea to run the TensorFlow tests locally.

hunkim:
@zheng-xq Thanks for the tip. I'll always run tests locally even for doc changes.

However, testing takes a really long time for me: about 40 min in my Mac, and over 10 min in a good google cloud machine (16 vCPUs, 60 GB memory) even if I change only one code line.

I wonder why it runs all tests again and again for just one-line-change? Can it only run affected test cases by the change?

zheng-xq:
>> I wonder why it runs all tests again and again for just one-line-change? Can it only run affected test cases by the change?

The public bazel does not have the capability to achieve that automatically. 

When you get to the point that you know the test structure, you can just run the tests that are most likely to fail.

tensorflow/core/ops/math_ops.cc
hunkim: Good idea. However, there is one more kind:

```
REGISTER_OP("Lgamma")
    .UNARY()
    .Attr("T: {half, float, double}")
    .Doc(R"doc(
Computes the log of the absolute value of `Gamma(x)` element-wise.
)doc");
```
Should we have three MACROS in total?

- UNARY:  {half, float, double, int32, complex64, int64}
- UNARY_FP:  {half, float, double, int64}
- UNARY_FPC:  {half, float, double, complex64}

I think it's getting complicated. For somehow, just adding `int32, int64` in the cwise_op* might be simpler.

` REGISTER4(UnaryOp, CPU, "Cos", functor::cos, float, Eigen::half, double, int32, int64, complex64);
`
 It won't add much as we discussed, but does it hurt? I guess there is no backwards compatibility issue, and no need to add two more macros.

girving: We definitely don't want trigonometric functions of integers.  I think the macros should be

* `UNARY`: Explicit list of types (what you did)
* `UNARY_ALL`: the current list of types
* `UNARY_REAL`: half, float, double (not sure why you stuck int64 in their)
* `UNARY_COMPLEX`: half, float, double, complex64, complex128

where the last three macros are implemented in terms of the first.

girving:
Verifying with @josh11b, but I think we should do the surgery to make the backwards compatibility test pass.  However, you seem to say above that the backwards compatibility test doesn't fail with your change, which is suspicious.  To confirm, can you do

    bazel test core/ops/compat:backwards_compatibility_test

That should fail.  Assuming it does, it should be reasonably easy to fix by manually editing

     core/ops/compat/ops_history.v0.pbtxt

to remove the types that are invalid.  I will then look at the changes to that file very carefully. :)

tensorflow/core/ops/math_ops.cc
hunkim: Sounds good. 
" (not sure why you stuck int64 in their)" was a typo.

girving:
@hunkim: Go for stripping the bad dtypes out of `ops_history.v0.pbtxt`.  Let me know if you have questions or want help.

tensorflow/core/ops/math_ops.cc
hunkim: @girving two quick questions for you:

- UNARY: {half, float, double, int32, complex64, int64}
- UNARY_ALL: {half, float, double, uint8, int8, int16, int32, int64, complex64}

Something like this? I am not sure UNARY_ALL is necessary.

- UNARY_REAL: {half, float, double}
- UNARY_COMPLEX: {half, float, double, complex64, ~~complex128~~}

I don't see `complex128` used in math_ops or cwise_op_\*? Delete `complex128` or add it in the MACRO and cwise_op_\*?

Thanks!

girving: Since you're touching the code, let's go ahead and add `complex128` support.  It should just mean adding it to the macro here and a few register macros in `cwise_op_*` as you say.  Thanks!

Your list of macros seems good, except that we should go ahead and allow `complex128` everywhere that `complex64` is allowed.

hunkim: Thanks! Then, let me submit two separate PRs:

1. adding the `complex128` support
2. adding 'UNARY' MACROS

for easy review and testing. I'm working on it.

hunkim: For 1, I have submitted PR #2261. Thanks!

hunkim:
I have submitted a new PR #2435.

Issue #2243, opened 2016-05-06
Error when building on Mac with GPU


LostString:
I am new to this so feel free to delete this if this is trivial.

I encountered the following errors when building from source on Mac with GPU support

`tensorflow/core/kernels/cudnn_pooling_gpu.cc:71:32: error: implicit instantiation of undefined template 'std::__1::array<long long, 3>'
  for (size_t i = 0; i < window.size(); ++i) {
                               ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__tuple:95:65: note: template is declared here
template <class _Tp, size_t _Size> struct _LIBCPP_TYPE_VIS_ONLY array;
                                                                ^
tensorflow/core/kernels/cudnn_pooling_gpu.cc:73:42: error: implicit instantiation of undefined template 'std::__1::array<long long, 3>'
    pooling_desc.set_window(dim_i, window.rbegin()[i]);
                                         ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__tuple:95:65: note: template is declared here
template <class _Tp, size_t _Size> struct _LIBCPP_TYPE_VIS_ONLY array;
`
and seven other similar errors.

Including `<array>` in `tensorflow/tensorflow/core/kernels/cudnn_pooling_gpu.cc` seems to fix the problem

zheng-xq:
In that case, tensorflow/core/kernels/cudnn_pooling_gpu.h should have included <array>

Please feel free to send a PR.

PR #2250, opened 2016-05-06
fix compile issue on os x

kashif:
for issue #2243


zheng-xq:

PR #2257, opened 2016-05-06
Update os_setup.md

hephaex:
From PR#2095, I've checked command to Installing from sources on OS X.

To set up the TensorFlow form source on OS X, pip installation need root permission. 

> sudo pip install /tmp/tensorflow_pkg/tensorflow-0.8.0-py2-none-any.whl


petewarden:
Test this please, Jenkins.

martinwicke:
I'm still not happy about just blanket adding sudo. We already have a sentence there saying the command (whl file name) may depend on the platform. I suggest we add there that "on some platforms (some versions of MacOS in particular), you may have to use sudo", without actually putting that into the code.

If you add sudo on a linux machine, it will actually mess with your pip install and prevent updates without sudo and all kinds of other nasty stuff.

martinwicke:
Never mind. I got confused where in the doc this was. Sorry about that.

PR #2265, opened 2016-05-07
Parameter keep_prob now dropout

codingyourlife:
Seems as parameter keep_prob is now called dropout. Because of the comment above I'm pretty sure droupout is meant.


caisq:
@tensorflow-jenkins : test this please

caisq:
Merged. Thanks, @codingyourlife

PR #2272, opened 2016-05-08
Renamed `vocabulary_size` to `character_set_size`.  "Vocabulary" mean…

Jakobovski:
…s "a set of words" not a "set of characters" and thus the variable name was misleading.


petewarden:
Test this please, Jenkins.

martinwicke:
vocabulary is a very generic term in this context, just like alphabet for coding theory. I don't think we'd want to deviate from standard nomenclature for character level models, but @vincentvanhoucke should decide.

vincentvanhoucke:
Yes, 'vocabulary' in the context of NLP defines the set of entities that are being predicted. They're often words but can be sub-word units or characters. This is very standard practice, so I'll close this PR.

PR #2273, opened 2016-05-08
Markdown links using `[This link](http://example.net/)` syntax

eerwitt:
Noticed these spots are using a syntax similar to markdown links but weren't rendering in the markdown under `g3doc`.

There may be a valid reason these links are different, they do cross new line boundaries but there are other links which do as well and use `[This link](http://example.net/)` syntax.


caisq:

petewarden:
Test this please, Jenkins.

PR #2275, opened 2016-05-08
Added comments to clarify how/why the embeddings are being minimized.

Jakobovski:
Reason for commit: For a student who has been going through the udacity examples, this particular use of the minimize function is a bit confusing as its difference between the previous examples is subtle. To the student it appears that only the softmax weights are being modified.


petewarden:
Test this please, Jenkins.

PR #2297, opened 2016-05-09
Fix botched merge in conv_grad_ops.cc.

vrv:
We returned in the wrong location, so that we ended up doing twice the
work: the optimal gemm and then the suboptimal conv (for 1x1 filter).

No tests failed because the result was the same.

### DIFF
--- a/tensorflow/core/kernels/conv_grad_ops.cc
+++ b/tensorflow/core/kernels/conv_grad_ops.cc
@@ -888,8 +888,9 @@ class Conv2DSlowBackpropInputOp : public OpKernel {
       if (!blas_launch_status) {
         context->SetStatus(errors::Internal("Blas SGEMM launch failed : m=", m,
                                             ", n=", n, ", k=", k));
-        return;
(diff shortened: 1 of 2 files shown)


tensorflow/core/kernels/conv_grad_ops.cc
```diff
@@ -888,8 +888,9 @@ class Conv2DSlowBackpropInputOp : public OpKernel {
       if (!blas_launch_status) {
         context->SetStatus(errors::Internal("Blas SGEMM launch failed : m=", m,
                                             ", n=", n, ", k=", k));
-        return;
       }
+
+      return;
```
zheng-xq: Could you do the one in conv_ops.cc as well? 

tensorflow/core/kernels/conv_ops.cc

vrv: Done.  Can't say I wasn't consistent!

zheng-xq:
LGTM

PR #2302, opened 2016-05-10
Fixed Tensorboard minimap not being drawn on Safari (OS X, iOS)

qbx2:
Fixed Tensorboard minimap not being drawn on Safari (OS X, iOS)
Tested on Chrome (OS X), Safari (OS X, iOS)

### DIFF
--- a/tensorflow/tensorboard/components/tf-graph-common/lib/scene/minimap.ts
+++ b/tensorflow/tensorboard/components/tf-graph-common/lib/scene/minimap.ts
@@ -267,8 +267,11 @@ export class Minimap {
       downloadContext.drawImage(image, 0, 0,
         this.downloadCanvas.width, this.downloadCanvas.height);
     };
-    let blob = new Blob([svgXml], {type: 'image/svg+xml;charset=utf-8'});
-    image.src = URL.createObjectURL(blob);
+    image.onerror = () => {
+      let blob = new Blob([svgXml], {type: 'image/svg+xml;charset=utf-8'});
+      image.src = URL.createObjectURL(blob);
+    }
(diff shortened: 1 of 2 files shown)


caisq:

qbx2:
This patch also makes "Download PNG" work on Safari.

vrv:
@dsmilkov: look ok?  We'll merge if so.

dsmilkov:
Thanks for the trying to fix this!

I tested it and the minimap doesn't work on Firefox now. Can you get a solution that works on Chrome/Firefox/MacOS? Thanks!

dsmilkov:
@qbx2 I did some testing and having the old way as a fallback makes it work on Firefox/Safari/Chrome. Can you change your commit to:
```
image.src = 'data:image/svg+xml;charset=utf-8,' + svgXml.replace(/\n/g,'');
image.onerror = () => {
  let blob = new Blob([svgXml], {type: 'image/svg+xml;charset=utf-8'});
  image.src = URL.createObjectURL(blob);
};
```

qbx2:
@dsmilkov I committed as you suggested. Test please.

dsmilkov:
Thanks! One last question: any reason why in the new commit, you removed the "replace" as in svgXml.replace(/\n/g,''); Is that not needed? Thanks!

qbx2:
I used "replace" for debugging because URL must not contain any new-line characters. After checking if it would work if i remove "replace", I removed it for clearer code because both worked well.

dsmilkov:
Great - was just curious! @vrv Good to merge!

qbx2:
Url-encoding svgXml data would be good

vrv:
there are conflicts unfortunately -- can you rebase so we can test and merge?

qbx2:
@vrv test please.

vrv:

PR #2321, opened 2016-05-11
Confusion with QueueRunner doc

cgel:
When reading the threading_and_queues how_to and the doc of QueueRunner the phrase "repeatedly run an enqueue op" confused me.
I think these minor changes would save others a few minutes of confusion.

### DIFF
--- a/tensorflow/g3doc/api_docs/python/train.md
+++ b/tensorflow/g3doc/api_docs/python/train.md
@@ -1332,6 +1332,7 @@ Wait till the Coordinator is told to stop.
 ### `class tf.train.QueueRunner` {#QueueRunner}
 
 Holds a list of enqueue operations for a queue, each to be run in a thread.
+It asynchronously runs the enqueue ops in different threads when the queuue is not full.
 
 Queues are a convenient TensorFlow mechanism to compute tensors
 asynchronously using multiple threads. For example in the canonical 'Input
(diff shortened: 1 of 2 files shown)


cgel:
I signed it!

tensorflow/g3doc/api_docs/python/train.md
```diff
@@ -1332,6 +1332,7 @@ Wait till the Coordinator is told to stop.
 ### `class tf.train.QueueRunner` {#QueueRunner}
 
 Holds a list of enqueue operations for a queue, each to be run in a thread.
+It asynchronously runs the enqueue ops in different threads when the queuue is not full.
```
vrv: 1) queuue -> queue

2) I think to be precise, the thread always runs the enqueue op, but the enqueue op itself blocks when the queue is full.  @josh11b or @mrry to confirm

mrry: Confirmed.

Is it obvious from the comment that enqueue operations run multiple times in a tight loop? Perhaps would be worth adding that?

mrry:
Hi there - just so we can make the comment as clear as possible, what was it about the remark in threading_and_queues.md that confused you?

cgel:
If you are not accustomed to the queue class it is not obvious that repeatedly runing an enqueue op will block.
My initial idea is that it would overflow the queue.

vrv:
Ping for @mrry, is the existing comment okay?

vrv:
closing due to inactivity, please comment and we'll reopen!

PR #2329, opened 2016-05-11
Additional instructions on how to fix missing submodules

KarlAmort:
This adds a better hint on how to fix a repository that was cloned without `--recurse-submodules`.

Currently, it recommends to run `git clone` again, which may be a somewhat involved operation if the user has made modifications, pulled changesets from other branches etc. before their first attempt to compile.
##### Old:

`ERROR: It appears that the required submodule google/protobuf is not available in this TensorFlow git clone.
Please be sure to use the --recurse-submodules flag when performing git clone of TensorFlow.`
##### New:

`ERROR: It appears that the required submodule google/protobuf is not available in this TensorFlow git clone.
Please re-run your git clone of Tensorflow with the --recurse-submodules flag or try git pull && git submodule init && git submodule update && git submodule status`

### DIFF
--- a/configure
+++ b/configure
@@ -7,8 +7,8 @@ DO_NOT_SUBMIT_WARNING="Unofficial setting. DO NOT SUBMIT!!!"
 if [[ ! -f "google/protobuf/protobuf.bzl" ]]; then
   echo "ERROR: It appears that the required submodule google/protobuf is not "\
 "available in this TensorFlow git clone."
-  echo "Please be sure to use the --recurse-submodules flag when performing "\
-"git clone of TensorFlow."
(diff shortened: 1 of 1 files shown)


vrv:
would "git submodule update --init" be sufficient ?

caisq:
Tried "git submodule update --init --recursive" and it works.

vrv:
Okay I think git submodule update --init --recursive might be simpler then.  Feel free to update the PR with that

vrv:
Now that we don't use submodules for protobuf, we can actually close this, but thanks for the doc suggestion!

PR #2336, opened 2016-05-12
Set workspace name

### LINKED ISSUE #2040: building pip package issue
```
bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
2016年 04月 21日 星期四 07:34:12 CST : === Using tmpdir: /tmp/tmp.cMsLMGw3bz
cp: cannot stat ‘bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/tensorflow’: No such file or directory
cp: cannot stat ‘bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/external’: No such file or directory
```

kchodorow:
Adds the workspace name to runfiles paths, removes a hack introduced in
https://github.com/tensorflow/tensorflow/commit/8cda66e19c0bdc6d8fc373fdcefcc644ab5ecfa3,
and fixes #2040.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,3 +1,5 @@
+workspace(name = "org_tensorflow")
+
 # Uncomment and update the paths in these entries to build the Android demo.
 #android_sdk_repository(
 #    name = "androidsdk",
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -650,7 +650,7 @@ bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_packag
 
 mkdir _python_build
 cd _python_build
-ln -s ../bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/* .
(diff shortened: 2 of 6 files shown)


vrv:
Does this require a specific version of bazel to be installed, btw?

kchodorow:
Yes, this will only work with 0.2.3+ (see https://groups.google.com/d/msg/bazel-discuss/_Xo14sf7fHw/g7wdatTGBgAJ for more info about the change).  For this to work with older versions, I can make it check both paths and use whichever exists.

vrv:
Given that 0.2.3 isn't officially released yet, I think we'd prefer to try to make it work with both versions, at least for a little transition period.

kchodorow:
Okay, finally updated this, at least according to the test suite this should now work with old and new versions of Bazel.

vrv:

kchodorow:
Whoops, I didn't realize your CI was still on 0.2.1, should be fixed it for that version (and previous versions), too.

vrv:
We should probably upgrade our CI to use a newer version of bazel, but supporting older is good too, particularly since check_version requires 0.2.1 right now

PR #2337, opened 2016-05-12
Update seq2seq

raingo:
This PR brings updates to the seq2seq examples, removing all of the `with ops.device("/cpu:0"):`. 

There are many ops implemented with cuda kernels recently, so maybe it's time to remove some cpu device specifications from the translate.py examples.

I don't have profile numbers for the translate.py example, but according to a similar experiment, this will boost the speed 5~6 times on a NVIDIA Titan GPU, especially when there are more than one independent runs going together.

Side note: when training with multiple cards (independent runs), the frequent data transfer between gpu and cpu is particularly harmful in my case.

### DIFF
--- a/tensorflow/python/ops/seq2seq.py
+++ b/tensorflow/python/ops/seq2seq.py
@@ -257,9 +257,8 @@ def embedding_rnn_decoder(decoder_inputs, initial_state, cell, num_symbols,
     proj_biases.get_shape().assert_is_compatible_with([num_symbols])
 
   with variable_scope.variable_scope(scope or "embedding_rnn_decoder"):
-    with ops.device("/cpu:0"):
-      embedding = variable_scope.get_variable("embedding",
-                                              [num_symbols, embedding_size])
+    embedding = variable_scope.get_variable("embedding",
+            [num_symbols, embedding_size])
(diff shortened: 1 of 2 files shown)


raingo:
ping ...

Any feedbacks are appreciated.

tensorflow/python/ops/seq2seq.py
```diff
@@ -257,9 +257,8 @@ def embedding_rnn_decoder(decoder_inputs, initial_state, cell, num_symbols,
     proj_biases.get_shape().assert_is_compatible_with([num_symbols])
 
   with variable_scope.variable_scope(scope or "embedding_rnn_decoder"):
-    with ops.device("/cpu:0"):
-      embedding = variable_scope.get_variable("embedding",
-                                              [num_symbols, embedding_size])
+    embedding = variable_scope.get_variable("embedding",
+                                          [num_symbols, embedding_size])
```
vrv: can you align these?

vrv:

raingo:
@vrv Thanks for your attention. Updated. Do you know why the test failed on the matmul?

vrv:
no idea, let me try again.

PR #2341, opened 2016-05-12
Upstream changes from internal

vrv:


vrv:
wow, 1 hour to test GPU .... at least it passed.

PR #2346, opened 2016-05-13
Basic implementation of immediate execution mode for TensorFlow

yaroslavvb:
This version implements Immediate execution for native TensorFlow ops with graph caching:

IE, you can do

```
    tf = immediate.Env()
    val = np.ones(())
    tensor1 = immediate.Tensor.numpy_to_tensor(env, val)
    tensor2 = immediate.Tensor.numpy_to_tensor(env, val)
    tensor3 = tf.add(tensor1, tensor2)   # executes operation immediately
    tensor3 = tf.add(tensor3, tensor2)   # executes operation without modifying graph
    print tensor3

```

What's not implemented yet (rough implementation detail is in the design doc):
- support for "tf.nn" namespace
- support for manually created Python ops
- caching for get_session_tensor and get_session_handle ops

@yuanbyu @keveman


yaroslavvb:
I signed it!

yaroslavvb:
If I click on the link it gives me, it shows that I signed
screen shot 2016-05-12 at 6 01 40 pm

yaroslavvb:
rats, I did all my commits as "yaroslavvb@Yaroslavs-MacBook-Pro.local"

yaroslavvb:
@adarob using `tf` or some other name is up to the user, the framework itself doesn't change anything in the original `tf` or `tensorflow` namespace. You could use `tfi` instead of `tf` for instance. I've been using `tf` in order to reuse existing tests, but run them in immediate mode, but when you want to mix things, I would use `tfi`. Also, Python's symbol resolution happens during module loading time, so when SyntaxNet is loaded it remembers what `tf` was during that time, so if you redefine it later, it won't be affected.

yaroslavvb:
@martinwicke  I just did rebase to "p4 sync" my version to head, and now it won't pass "cla" because there are commits from non-CLA-signed googlers in the history for my Pull Request ... any tips? (maybe I shouldn't be doing rebase? Just started learning git 2 weeks ago)

martinwicke:
You should do a rebase, I think something went wrong in the one you did. Try making a clean branch from tensorflow/tensorflow, cherry-picking your changes there, and then force-pushing it onto your master.

martinwicke:
Something is wrong with your fork. I at this point, I would recommend to keep your changes someplace safe, delete the fork, and re-fork. Then add the new fork as a remote to your existing local copy, and do something like
```
git remote add new-fork git@github.com/yaroslavvb/tensorflow
git checkout new-fork/master
git cherry-pick all-your-actual-changes
git push new-fork
```

Or similar. In the end, you should have only a few commits on top of master when you make the PR.

yaroslavvb:
Thanks for the tip, I'll give that a try

yaroslavvb:
I accidentally committed debug changes to files outside of contrib/ which makes this pull request hard to merge, closing this request to replace it with a fresh one

PR #2347, opened 2016-05-13
Set -headerpad_max_install_names on Darwin for targets that need binary header padding to rename paths

### LINKED ISSUE #2332: install_name_tool error when running reduction_ops_test_gpu on OS X
I ran into this while testing my patch tensorflow/tensorflow#1289 on OS X. At first, I thought that it was caused by my patch but then I found that it also reproduces at HEAD.

I am not entirely sure whether this is a TensorFlow issue (the error seems to indicate that we need to add a `ld` flag) or a Bazel issue in the gcc wrapper.

+cc @lberki @damienmg

### Environment info
Operating System: Mac OS X 10.11.4

Installed version of CUDA and cuDNN: Building without GPU support

If installed from sources, provide the commit hash: d3d1a63a8b8ad99347d95756e88ae0589ed1a9b0

### Steps to reproduce
1. Run `bazel test //tensorflow/core/kernels:reduction_ops_test_gpu` on OS X

### What have you tried?
1. This does not reproduce on Linux

```
❯❯❯ bazel test //tensorflow/core/kernels:reduction_ops_test_gpu --test_output=errors
WARNING: /private/var/tmp/_bazel_dzc/50d95479571c8b29601dcbccdd342764/external/re2/WORKSPACE:1: Workspace name in /private/var/tmp/_bazel_dzc/50d95479571c8b29601dcbccdd342764/external/re2/WORKSPACE (@__main__) does not match the name given in the repository's definition (@re2); this will cause a build error in future versions.
WARNING: /Users/dzc/Projects/tensorflow/tensorflow/google/protobuf/BUILD:59:16: in includes attribute of cc_library rule //google/protobuf:protobuf_lite: 'src/' resolves to 'google/protobuf/src' not in 'third_party'. This will be an error in the future.
WARNING: /Users/dzc/Projects/tensorflow/tensorflow/google/protobuf/BUILD:124:16: in includes attribute of cc_library rule //google/protobuf:protobuf: 'src/' resolves to 'google/protobuf/src' not in 'third_party'. This will be an error in the future.
WARNING: /Users/dzc/Projects/tensorflow/tensorflow/google/protobuf/BUILD:266:16: in includes attribute of cc_library rule //google/protobuf:protoc_lib: 'src/' resolves to 'google/protobuf/src' not in 'third_party'. This will be an error in the future.
INFO: Found 1 test target...
ERROR: /Users/dzc/Projects/tensorflow/tensorflow/tensorflow/core/kernels/BUILD:1044:1: Linking of rule '//tensorflow/core/kernels:reduction_ops_test_gpu' failed: cc_wrapper.sh failed: error executing command external/local_config_cc/cc_wrapper.sh -o bazel-out/local-fastbuild/bin/tensorflow/core/kernels/reduction_ops_test_gpu '-Wl,-rpath,$ORIGIN/../../../_solib_darwin/' ... (remaining 53 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
clang: warning: argument unused during compilation: '-pthread'
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: changing install names or rpaths can't be redone for: bazel-out/local-fastbuild/bin/tensorflow/core/kernels/reduction_ops_test_gpu (for architecture x86_64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)
Target //tensorflow/core/kernels:reduction_ops_test_gpu failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.623s, Critical Path: 0.49s

Executed 0 out of 1 test: 1 fails to build.
```

davidzchen:
Fixes #2332

### DIFF
--- a/tensorflow/contrib/util/BUILD
+++ b/tensorflow/contrib/util/BUILD
@@ -46,6 +46,10 @@ cc_test(
         "//tensorflow/core:test",
         "//tensorflow/core:test_main",
     ],
+    linkopts = select({
+        "//tensorflow:darwin": ["-headerpad_max_install_names"],
+        "//conditions:default": [],
+    }),
 )
 
 cc_binary(
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -1276,6 +1276,10 @@ tf_cc_tests(
         "//tensorflow/core/kernels:ops_util",
         "//third_party/eigen3",
     ],
+    linkopts = select({
+        "//tensorflow:darwin": ["-headerpad_max_install_names"],
+        "//conditions:default": [],
+    }),
 )
 
 # GPU-related tests
--- a/tensorflow/core/distributed_runtime/rpc/BUILD
+++ b/tensorflow/core/distributed_runtime/rpc/BUILD
@@ -347,6 +347,10 @@ tf_cuda_cc_tests(
         "//tensorflow/core:testlib",
         "//tensorflow/core/distributed_runtime:server_lib",
     ],
+    linkopts = select({
+        "//tensorflow:darwin": ["-headerpad_max_install_names"],
+        "//conditions:default": [],
+    }),
 )
 
 tf_cuda_cc_tests(
--- a/tensorflow/core/kernels/BUILD
+++ b/tensorflow/core/kernels/BUILD
@@ -485,6 +485,10 @@ tf_cc_test(
         "//tensorflow/core:test_main",
         "//tensorflow/core:testlib",
     ],
+    linkopts = select({
+        "//tensorflow:darwin": ["-headerpad_max_install_names"],
+        "//conditions:default": [],
+    }),
 )
 
 tf_cc_test(
@@ -786,6 +790,10 @@ tf_cc_tests(
         "//tensorflow/core:test_main",
         "//tensorflow/core:testlib",
     ],
+    linkopts = select({
+        "//tensorflow:darwin": ["-headerpad_max_install_names"],
+        "//conditions:default": [],
+    }),
 )
 
 tf_cuda_cc_test(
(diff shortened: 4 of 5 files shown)


vrv:
why this test only?

vrv:
also this doesn't select based on platform

davidzchen:
Oops. Accidentally did a force push on a different machine without pulling. The select should be there now.

I only noticed it on this test the first time through but I'll run everything again to see if there are others that need this too.

vrv:
Do you understand why it fixes anything, and what the implications are for setting this globally?  Cause I don't :)

davidzchen:
All of the targets that have this problem should be covered now.

It seems that the problem is occurring when [Bazel's cc_wrapper script](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/osx_cc_wrapper.sh) renames all the paths in the binary with the paths relative to the binary (since relpath in OS X doesn't really work). For these targets, the load commands in the binary end up being [larger after the renaming](http://stackoverflow.com/questions/28324785/install-name-tool-cant-use-change-because-larger-updated-load-commands-do-not), which would require relinking. The `-headerpad_max_install_names` flag would be needed to ensure there is enough room in the binary for the larger load statements.

On that note, @damienmg, do you think it would be a good idea to set `-headerpad` or `-headerpad_max_install_names` by default in `osx_cc_wrapper.sh` or should we just set the linkopt for targets that need them?

damienmg:
@davidzchen I think this make more sense to add it to the CROSSTOOL file for darwin in Bazel.

martinwicke:
Jenkins, test this please.

martinwicke:
We probably want to accept this anyway, and remove it once the fix has made its way into a bazel release.

davidzchen:
Is this patch ready to be merged?

davidzchen:
Rebased on master.

Do we still want to merge this PR?

vrv:

Issue #1140, opened 2016-02-17
Math ops support for float64


smcantab:
I am raising this issue following [this stackoverflow discussion](https://stackoverflow.com/questions/35443080/tensorflow-critical-graph-operations-assigned-to-cpu-rather-than-gpu?lq=1). I found that there is not that much support for float64 ops, that is needed for some numerical applications. A number of other issues  #1061 #761 #547 and PR #1089 have been raised around this.

On the basis of how my graph is initialised, things like **MatMul** (also raised in this other [stackoverflow question](https://stackoverflow.com/questions/35428297/tensorflow-issue-with-gpu-on-matmul-gpu-isnt-recognized?lq=1)), **Mean, L2Loss, Mean_grad/Prod, Mul_grad/Sum, AddN, SoftmaxCrossEntropyWithLogits** are not implemented for float64.

Would it be a good idea to extend PR #1089 to more ops?

girving:
Looks like #1089 was already merged, so we'll need more PRs.

siddharth-agrawal:
I want to start contributing to Tensorflow, so thought this issue would be a good start. I started with MatMul from among the ops mentioned by @smcantab. In the op implementation I found [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/matmul_op.cc#L210), float64 seems to be purposely commented out. Is there a reason for this?

smcantab:
as @dave-andersen pointed out [here](https://stackoverflow.com/questions/35443080/tensorflow-critical-graph-operations-assigned-to-cpu-rather-than-gpu) some operations have not been registered for double simply to reduce the bloat in the binaries, given that, as far as I understand, it was not clear whether there'd be much interest in supporting double precision. Obviously there's been quite some interest so we should go ahead. For some operations it should be as simple as registering the kernel for double, and checking that the tests are there. Some other operations might require tailored optimization. I have been busy with some other stuff, so I have not been able to work on a PR yet, but @siddharth-agrawal if you start one I'll follow it and contribute as soon as possible.

siddharth-agrawal:
Sounds good. I will try to get something out by the end of this week.

girving:
@smcantab: Many people have requested better double support and it's critical for a lot of numerical applications, so we should do it even it increases binary size.  complex128 is rarer but the same principle applies.

siddharth-agrawal:
[...]

PR #2352, opened 2016-05-13
Enable GPU for L2Loss float64

siddharth-agrawal:
Enabled GPU registration for L2Loss operations of type double. This partially addresses #1140.

Tested locally. I'm assuming that there won't be any change in the tests because of [this commit](https://github.com/tensorflow/tensorflow/commit/a1bc10f9e28cd3d33e91bfaedd8199e4590f2893).

### DIFF
--- a/tensorflow/core/kernels/l2loss_op.cc
+++ b/tensorflow/core/kernels/l2loss_op.cc
@@ -68,6 +68,7 @@ namespace functor {
   extern template struct L2Loss<GPUDevice, T>;
 
 DECLARE_GPU_SPEC(float);
+DECLARE_GPU_SPEC(double);
 DECLARE_GPU_SPEC(Eigen::half);
 #undef DECLARE_GPU_SPEC
 }  // namespace functor
@@ -79,6 +80,7 @@ DECLARE_GPU_SPEC(Eigen::half);
       L2LossOp<GPUDevice, T>);
 
 REGISTER_GPU_KERNEL(float);
+REGISTER_GPU_KERNEL(double);
 REGISTER_GPU_KERNEL(Eigen::half);
 #undef REGISTER_GPU_KERNEL
 
--- a/tensorflow/core/kernels/l2loss_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/l2loss_op_gpu.cu.cc
@@ -25,6 +25,7 @@ namespace tensorflow {
 
 typedef Eigen::GpuDevice GPUDevice;
 template struct functor::L2Loss<GPUDevice, float>;
+template struct functor::L2Loss<GPUDevice, double>;
 template struct functor::L2Loss<GPUDevice, Eigen::half>;
 
 }  // namespace tensorflow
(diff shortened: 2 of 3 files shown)


vrv:
I think we do want to add a test for double -- that commit you linked was removing test cases, I think.

vrv:
Actually I'm not sure what's going on now -- if nn_test is passing right now without the GPU double kernel registered, then that kernel isn't being tested.

siddharth-agrawal:
@vrv Is there something I need to do?

vrv:
What I'm asking is for a test of the L2Loss op -- you've added the registration but we have no idea if it works for double!  So you should be able to write a test that fails without the registration, and passes after it.

siddharth-agrawal:
@vrv I found something strange. Even if I change [this line](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_test.py#L129) to `with self.test_session(use_gpu=True)`, the test succeeds; without the registration of the float64 GPU kernel. How is this possible?

vrv:
I'm not sure, can you try passing 'log_device_placement=True' to the session options passed to test_session to see what's going on?

siddharth-agrawal:
@vrv Gives me the following output:
`Device mapping:
/job:localhost/replica:0/task:0/gpu:0 -> device: 0, name: GeForce GTX 950M, pci bus id: 0000:01:00.0
Const: /job:localhost/replica:0/task:0/gpu:0
Identity: /job:localhost/replica:0/task:0/gpu:0
x: /job:localhost/replica:0/task:0/gpu:0
gradients/L2Loss_grad/mul: /job:localhost/replica:0/task:0/gpu:0
L2Loss: /job:localhost/replica:0/task:0/cpu:0`

L2Loss on the CPU, this is with `use_gpu=True`. Any hints as to where I should dig for this?

siddharth-agrawal:
@vrv I don't think that we should add any more tests here, just to be consistent. For example, consider the changes in my commit [here](https://github.com/tensorflow/tensorflow/pull/2506/commits/00d402d350302764595ce5cd6aec03e33471baae).  Even if I remove the registration from `matmul_op.cc`, the test still succeeds as it is able to find the kernel for the CPU (same is true for all the other tests in TF). The only way we will get the behavior you want is by using `force_gpu=True`, which has not been used in other tests. Does this make sense?

vrv:
You should still add a test for float64 though, otherwise it's just completely untested.

You can then use log_device_placement temporarily to verify it's running on the GPU.

siddharth-agrawal:
@vrv There is a test for float64 [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_test.py#L128). I verified using `force_gpu=True`

vrv:
Nice, let's also add one https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_test.py#L119 as well.  something like

    for dtype in [tf.float32, tf.float64]:
      with self.test_session():
        x = tf.constant(..., dtype=dtype)
        ... same as before ...

siddharth-agrawal:
@vrv Modified the test.

vrv:

siddharth-agrawal:
@vrv The failure was due to a timeout, seems unrelated to the change.

PR #2378, opened 2016-05-16
Enable password support for Jupyter in a Docker image

victorhcm:
This small update enables Jupyter to support passwords passed as an environment variable when instantianting a docker container. 

In the same way, `jupyter_notebook_config.py` could also be modified to support HTTPS by default. If there is an interest, I could also include these updates in the same PR.

### DIFF
--- a/tensorflow/tools/docker/README.md
+++ b/tensorflow/tools/docker/README.md
@@ -40,13 +40,13 @@ accomplished via
 
     $ export CUDA_SO=$(\ls /usr/lib/x86_64-linux-gnu/libcuda.* | xargs -I{} echo '-v {}:{}')
     $ export DEVICES=$(\ls /dev/nvidia* | xargs -I{} echo '--device {}:{}')
-    $ docker run -it -p 8888:8888 $CUDA_SO $DEVICES gcr.io/tensorflow/tensorflow-devel-gpu
+    $ docker run -it -p 8888:8888 $CUDA_SO $DEVICES gcr.io/tensorflow/tensorflow-gpu
 
 Alternately, you can use the `docker_run_gpu.sh` script in this directory.
 
 In order to set Jupyter Notebook to require a password, the `-e PASSWORD=pass` option must be provided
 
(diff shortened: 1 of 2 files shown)


tensorflow/tools/docker/jupyter_notebook_config.py
```diff
@@ -17,3 +17,10 @@
 c.NotebookApp.port = 8888
 c.NotebookApp.open_browser = False
 c.MultiKernelManager.default_kernel_name = 'python2'
+
+# sets a password if PASSWORD is set in the environment
+if 'PASSWORD' in os.environ:
+    import os
+    from IPython.lib import passwd
```
caisq: Move the imports to the beginning of the code.

tensorflow/tools/docker/README.md
```diff
@@ -44,6 +44,10 @@ accomplished via
 
 Alternately, you can use the `docker_run_gpu.sh` script in this directory.
 
+In order to set Jupyter Notebook to require a password, the `-e PASSWORD=pass` option must be provided
+
+    $ docker run -it -p 8888:8888 $CUDA_SO $DEVICES -e PASSWORD=pass gcr.io/tensorflow/tensorflow-devel-gpu
```
caisq: Users of Jupyter Notebook docker images of TensorFlow typically want to use the non-devel images. Does it make sense to change the image name to "gcr.io/tensorflow/tensorflow-gpu" or "gcr.io/tensorflow/tensorflow"?

victorhcm: That makes sense. Should we change line 43 too?

caisq: Yes, please change both to gcr.io/tensorflow/tensorflow-gpu

caisq:

caisq:
Merged. Thanks.

victorhcm:
Thanks! Do you think we also should add support to HTTPS? It would be passed as `-e USE_HTTPS=1`, otherwise it would use simple HTTP by default.

PR #2379, opened 2016-05-16
[skflow] Added verbosity support to Validation monitor

nmiculinic:
Aim of this diff is adding verbosity switch to Validation monitor class,
whcih previosly had none (It was automatically set to level 1).

I've tested locally on my machine, and works as expected:

``` python
skflow.monitors.ValidationMonitor
regressor = skflow.TensorFlowDNNRegressor(hidden_units=[10, 10],
    steps=5000, learning_rate=0.1, batch_size=1,
                                          verbose=0)

regressor.fit(X_t, Y_t, monitor=skflow.monitors.ValidationMonitor(X_cv, Y_cv, verbose=0))

```

Previous code is a bit edited example ones.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/monitors.py
+++ b/tensorflow/contrib/learn/python/learn/monitors.py
@@ -190,10 +190,12 @@ class ValidationMonitor(BaseMonitor):
                val_y,
                n_classes=0,
                print_steps=100,
-               early_stopping_rounds=None):
+               early_stopping_rounds=None,
+               verbose=1):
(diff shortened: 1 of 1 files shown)


ilblackdragon:
Hi @nmiculinic Thanks for contribution, but we actually refactored both code to run evaluation (via `evaluate` method which now is verbose by default) and monitoring (change a API as well).

I'll close this one for today, please try new APIs and see if there is missing functionality there.
Thanks!

PR #2404, opened 2016-05-17
Updated installation instructions for conda.

ziky90:
This PR updates documentation of installation instructions for conda users.

The new version was tested so it seems that nobody has changed documentation of installation instructions.


mrry:
Thanks for the update!

PR #2263, opened 2016-05-07
Add complex128 support

hunkim:
Added the complex128 support, based on the guidance from @girving and others in PR #2244. All tests (in CPU) including backwards_compatibility_test are passed! I couldn't test the GPU version.

The complex128 support is not added for _FFT_ ops with input: complex64 and output: complex64.

PR #2417, opened 2016-05-18
Add more changes for complex128 support

ibab:
This PR contains all ops from @hunkim's #2263, rebased on `master` with a few adjustments.
I've extended all relevant tests, and added a few more ops that were missing from the PR above.
With this PR, support for `complex128` is almost finished (documentation changes and a few ops are missing).


hunkim:
Looks good! Thanks!!

tensorflow/core/kernels/cwise_ops_gpu_common.cu.h
```diff
@@ -150,6 +150,9 @@ struct BinaryFunctor<GPUDevice, Functor, NDIMS, has_errors> {
 #define DEFINE_BINARY9(F, T0, T1, T2, T3, T4, T5, T6, T7, T8) \
   DEFINE_BINARY4(F, T0, T1, T2, T3);                          \
   DEFINE_BINARY5(F, T4, T5, T6, T7, T8)
+#define DEFINE_BINARY10(F, T0, T1, T2, T3, T4, T5, T6, T7, T8, T9) \
+  DEFINE_BINARY4(F, T0, T1, T2, T3, T4);                           \
```
girving: DEFINE_BINARY5

tensorflow/python/kernel_tests/matmul_op_test.py
```diff
@@ -45,9 +45,13 @@ def _testGpuMatmul(self, x, y, transpose_x=False, transpose_y=False):
...
-      real = self._randMatrix(rows, cols, np.float32)
-      imag = self._randMatrix(rows, cols, np.float32)
+    if dtype in (np.complex64, np.complex128):
+      if dtype == np.complex64:
+          float_dtype = np.float32
+      else:
+          float_dtype = np.float64
+      real = self._randMatrix(rows, cols, float_dtype)
+      imag = self._randMatrix(rows, cols, float_dtype)
       return real + np.complex(0, 1) * imag
```
girving: `real + 1j * imag`

girving:
Can you add a `complex128` version of the the binary op version of `testComplex64Basic` in `cwise_ops_test.py`?

girving:
Looks like `complex128` `tf.conj` needs a test too (copy the `complex64` one).

tensorflow/core/kernels/cwise_ops_gpu_common.cu.h
ibab: Ouch, thanks for spotting this one.

girving:
And `complex128` `tf.select` and `tf.pow` (see testOverload for the latter).  Finally `testOverloadComparisons` should test both `complex64` and `complex128`.

girving:
One more: `sparse_tensor_dense_matmul` needs to test `complex128`.  That looks like it; otherwise the testing is good.

tensorflow/core/kernels/cwise_op_gpu_equal_to.cu.cc
```diff
@@ -19,8 +19,8 @@ limitations under the License.
 
 namespace tensorflow {
 namespace functor {
-DEFINE_BINARY9(equal_to, float, Eigen::half, double, uint8, int8, int16, int64,
-               complex64, bool);
+DEFINE_BINARY10(equal_to, float, Eigen::half, double, uint8, int8, int16, int64,
+               complex64, bool, complex128);
```
hunkim: It a small thing, but perhaps put complex64 and complex128 together?  -> complex64, complex128, bool

tensorflow/core/kernels/cwise_op_gpu_not_equal_to.cu.cc
```diff
@@ -19,8 +19,8 @@ limitations under the License.
 
 namespace tensorflow {
 namespace functor {
-DEFINE_BINARY9(not_equal_to, float, Eigen::half, double, uint8, int8, int16,
-               int64, complex64, bool);
+DEFINE_BINARY10(not_equal_to, float, Eigen::half, double, uint8, int8, int16,
+               int64, complex64, bool, complex128);
```
hunkim: -> complex64, complex128, bool

tensorflow/core/kernels/cwise_op_neg.cc
```diff
@@ -16,8 +16,8 @@ limitations under the License.
 #include "tensorflow/core/kernels/cwise_ops_common.h"
 
 namespace tensorflow {
-REGISTER6(UnaryOp, CPU, "Neg", functor::neg, float, Eigen::half, double, int32,
-          complex64, int64);
+REGISTER7(UnaryOp, CPU, "Neg", functor::neg, float, Eigen::half, double, int32,
+          complex64, int64, complex128);
```
hunkim: For better readability, perhaps int32, int64, complex64, complex128?

tensorflow/core/kernels/cwise_op_not_equal_to.cc
```diff
@@ -16,8 +16,9 @@ limitations under the License.
 #include "tensorflow/core/kernels/cwise_ops_common.h"
 
 namespace tensorflow {
-REGISTER11(BinaryOp, CPU, "NotEqual", functor::not_equal_to, float, Eigen::half,
-           double, uint8, int8, int16, int32, int64, complex64, string, bool);
+REGISTER12(BinaryOp, CPU, "NotEqual", functor::not_equal_to, float, Eigen::half,
+           double, uint8, int8, int16, int32, int64, complex64, string, bool,
+           complex128);
```
hunkim: complex64, complex128, string, bool?

tensorflow/core/ops/math_ops.cc
```diff
@@ -37,7 +37,7 @@ REGISTER_OP("BatchMatMul")
     .Input("x: T")
     .Input("y: T")
     .Output("output: T")
-    .Attr("T: {float, double, int32, complex64}")
+    .Attr("T: {float, double, int32, complex64, complex128}")
```
hunkim: int64 missing?

tensorflow/core/ops/math_ops.cc
```diff
@@ -134,7 +134,7 @@ tf.complex_abs(x) ==> [5.25594902, 6.60492229]
 // Declares cwise unary operations signature: 't -> 't
 #define UNARY()                      \
   Input("x: T").Output("y: T").Attr( \
-      "T: {half, float, double, int32, complex64, int64}")
+      "T: {half, float, double, int32, complex64, int64, complex128}")
```
hunkim: perhaps int32, int64, complex64, complex128?

tensorflow/core/ops/math_ops.cc
girving: Let's leave that out of this PR, since it's about complex128.

tensorflow/core/ops/math_ops.cc
```diff
@@ -306,7 +306,7 @@ REGISTER_OP("Add")
     .Output("z: T")
     .Attr(
         "T: {half, float, double, uint8, int8, int16, int32, int64, complex64, "
-        "string}")
+        "string, complex128}")
```
hunkim: complex64, complex128?

tensorflow/core/ops/math_ops.cc
```diff
@@ -375,7 +375,7 @@ REGISTER_OP("Pow")
     .Input("x: T")
     .Input("y: T")
     .Output("z: T")
-    .Attr("T: {half, float, double, int32, complex64, int64}")
+    .Attr("T: {half, float, double, int32, complex64, int64, complex128}")
```
hunkim: complex64, complex128?

tensorflow/core/ops/math_ops.cc
```diff
@@ -504,7 +504,7 @@ Returns the truth value of (x >= y) element-wise.
 #define EQUALITY_COMPARISON()                                                  \
   Input("x: T").Input("y: T").Output("z: bool").SetIsCommutative().Attr(       \
       "T: {half, float, double, uint8, int8, int16, int32, int64, complex64, " \
-      "quint8, qint8, qint32, string, bool}")
+      "quint8, qint8, qint32, string, bool, complex128}")
```
hunkim: complex64, complex128?

ibab:
Thanks for the comments!

`testOverloadComparisons` doesn't test for `complex64` because we can't define `greater/less` ops for it.
The second part of `testOverloadComparisons` only seems to operate on booleans.
(And, strangely, tests `tf.equal` and `tf.not_equal` instead of the overloaded operators).
Should I add `_EQ` and `_NEQ` lambdas to test the overloading for `==` and `!=`?
Then it would make sense to add `complex64` and `complex128` to the test.

girving:
Unfortunately, TensorFlow doesn't overload the `==` and `!=` operators: they are the same as `is` for `Tensor` objects in Python.  However, it's worth refactoring that routine so that we do test `equal` and `not_equal` for the two complex types.

Putting `complex64` and `complex128` next to each other is good as he suggests.

girving:
Go ahead and add new commits resolving the comments to make it easier to review, and we can squash at the end once everything looks good.

ibab:
Ah, right, `==` and `!=` aren't overloaded.
In this case, would it make more sense to extend `testTensorCompareTensor` and friends?
(Which also test `tf.equal` and `tf.not_equal`, but don't contain `complex64` and `complex128` at the moment)

girving:
Yep, `testTensorCompareTensor` sounds good.

ibab:
I think all comments have been addressed and the tests are passing for me on the CPU.

girving:
Looks great!  Jenkins, test this please.

ibab:
Looks like one of the tests failed.
This one was `(cached) PASSED` when I ran the tests, so maybe `bazel` didn't pick up the changes.
Will fix the problem and push again in a moment.

girving:
Yep, looks like transpose complex128 needs to be registered on GPU.

ibab:
I'm having some problems with extending the GPU kernel for `transpose` to `complex128`.
Calling `__ldg` on a `complex128*` doesn't compile for me, which means that I can't call
```
internal::Transpose<Device, complex128>(d, in, perm, out)
```
as I did with the CPU version.
The call to `__ldg` happens here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/transpose_functor_gpu.cu.cc#L40
Should I just define a template instance for `complex128` that doesn't use the `__ldg` intrinsic?

ibab:
Seems like I'm able to call `__ldg` on  `float4`, which should have the right size.
Is that a correct solution, or did I miss something?
I've pushed a commit that uses `float4` and makes the `transpose` tests run.

girving:
@ibab: What's the error message with `__ldg` on `complex128`?  We do want to preserve `__ldg` semantics here since it can improve performance, so we probably want to write a little wrapper around __ldb in `cuda_kernel_helper.h` that calls `__ldg` twice for `complex128`.

@zffchen78: You recently refactored transpose to be lower compile time; do you have suggestions for how to add `complex128`?  

@ibab: If you're unable to run GPU tests, it's going to be hard to iterate, so it may be worth splitting the changes out of this PR.  I could take over that bit of it if you want.

girving:
@ibab: Cool, `float4` sounds perfect.  I'll take a look.

girving:
Jenkins, test this please.

ibab:
Looks like we have some conflicts now.
They're very simple in case you want to solve this on your end, otherwise I can push my rebased version.

girving:
@ibab: Please push your rebased and squashed version (reduced to one commit).

ibab:
@girving: Okay, it's pushed.

girving:
Woot.  Jenkins, test this please.

girving:
Merged.  Thanks again for the contributions!  This will be great to have for applications of TensorFlow outside standard deep nets.

ibab:
@girving @hunkim: Thanks for reviewing the code!

PR #2418, opened 2016-05-18
R0.8

ByungKeon-Ko:


caisq:
We have been seeing this type of misoperation PRs from a couple of GH users recently. Not sure what might be the cause of those.

mrry:
@caisq: In that case I'll close this request, because it seems spurious.

terrytangyuan:
@caisq @mrry Seems like some of theses changes are not on master

PR #2440, opened 2016-05-20
Experimental makefile support

petewarden:
I'm working on automatically generating a makefile from our Bazel dependencies, to make it easier to cross-compile for mobile platforms, and build on systems that can't handle Bazel's resource requirements or dependencies. Supports OS X, Ubuntu, iOS, and Android currently.


cg31:
@petewarden, does the makefile also generate the ops files in tensorflow/cc/ops?

I checked the makefile but didn't find it,

petewarden:
@cg31 No, it's focused on the C++ core needed to load and run pre-trained inference models, which doesn't include the generated cc ops and headers. In this it's similar to the Android libraries we build with Bazel. We haven't defined this core subset of files explicitly before, so I'm hoping to document what we're doing a bit more.

It would also be possible to add the cc ops to the makefile, but I'm trying to keep it as minimal as possible since my focus is porting to resource-limited platforms.

timbrucks:
@petewarden Thanks for your contributions on this: the makefile works as advertised. I am running on OS X and can use this to produce the libtensorflow-core.a. I can build and run the example benchmark tool and it works.

Now I am trying to create a tool in my development paradigm (which utilizes CMake) that links against this lib. I can load a graph in (using ReadBinaryProto), but when I try to create a TF session like this:

> tensorflow::GraphDef reinspect_graph_def;
> tensorflow::Status graph_load_status = ReadBinaryProto(tensorflow::Env::Default(), reinspect_graph_file, &reinspect_graph_def);
> ...
> tensorflow::SessionOptions sess_opts;
> std::unique_ptr<tensorflow::Session> session_reinspect(tensorflow::NewSession(sess_opts));
> tensorflow::Status session_create_status = session_reinspect->Create(reinspect_graph_def);

I get this run time error:

> E tensorflow/core/common_runtime/session.cc:58] Not found: No session factory registered for the given session options: {target: "" config: } Registered factories are {}.
> Segmentation fault: 11

Any idea what us causing that error? My tool links against libtensorflow-core.a as well as the protobuf library that is installed on my machine. Perhaps I need to link against some other libs?

petewarden:
@timbrucks thanks for the update on your progress! Unfortunately that's a problem I know well, and the summary is that the linker is stripping out global constructors it mistakenly thinks aren't needed.

The longer version is that most components in TF are registered using C++ macros that look like REGISTER_KERNEL("SomeOp", SomeKernelClass).

Under the hood, those create a global variable of a C++ class that registers the component with the main registry as part of its constructor. The problem is, that global is just an implementation detail, and is never referenced by any code. This leads many linkers to conclude that it's not needed and can be removed, even though its constructor has an important side-effect.

The only reliable solution we have is to force the linker to avoid stripping any code from the library, even if it thinks it's not used. There are a variety of different ways to specify that, depending on your platform. On OS X/iOS you use -force_load with the path to the static library. On Ubuntu, we use --whole-archive:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/makefile/Makefile#L124

In your case, you should be able to pick the correct linker flags for the platform and it will avoid stripping the global that does the factory registration that's causing your error. Hopefully the linker flags in the makefile should help there.

timbrucks:
@petewarden thanks for the very clear explanation, I was pretty confused. Per your suggestion I was able to get the linker to stop stripping code from the library. I now have the TF core code integrated into my development paradigm (ahead of schedule too ... I did not think I could get it done before TF arrived at version 1.0).

If anyone is trying to do something similar via CMake, the step that overcame the linker issues was adding "-all_load" within the target_link_libraries() call in the CMakeLists.txt file

martinwicke:
@timbrucks you're not considering how slow we will be getting to 1.0. :)

jendap:
This looks useful. Does it work?

If so, should we create a build for it?

martinwicke:
We should. Its main purpose is to create an iOS library, for which we could
create a build on the Mac.

jendap:
I thought it is there for ios. More pain to maintain macs. Ok, I'll leave macs to caisq or gunan ;-)

AmmarkoV:
thank you

PR #2442, opened 2016-05-20
Add piecewise constant op

### LINKED ISSUE #2432: Feature Request: Staircase function
Right now for learning rates we have `exponential_decay`, which is useful but doesn't handle fine-tuned scheduling. For example, regardless of whether we did this manually or with `exponential_decay`, there'd be boilerplate code to use a learning rate of 1.0 for 100000 steps, 0.5 for the next 10000 steps, and 0.1 for all remaining steps.

This could be handled with a staircase function, maybe with usage like this:

```
boundaries = [100000, 110000]
values = [1.0, 0.5, 0.1]
learning_rate = staircase(global_step, boundaries, values)
```

This would create a `Tensor` that evaluates to 1.0 when `x <= 100000`, 0.5 when `x > 100000` and `x <= 110000`, and 0.1 when `x > 110000`.

Shouldn't be hard to build using `tf.case`.

Is this of interest or too specific?

rdipietro:
This should close https://github.com/tensorflow/tensorflow/issues/2432

This is more general than for learning-rate decay, but at the same time it didn't feel like it should be a native math op, so I kept it in `tensorflow/python/training/learning_rate_decay.py`.

### DIFF
--- a/tensorflow/python/training/learning_rate_decay.py
+++ b/tensorflow/python/training/learning_rate_decay.py
@@ -20,6 +20,7 @@ from __future__ import print_function
 
 from tensorflow.python.framework import ops
 from tensorflow.python.ops import math_ops
+from tensorflow.python.ops import control_flow_ops
 
 
 def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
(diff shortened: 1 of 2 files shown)


rdipietro:
The tests run, but I just noticed that `piecewise_constant` isn't available as `tf.train.piecewise_constant`, and I don't know how to fix this as I don't see any relevant entries in `tensorflow/python/training/__init__.py` or `tensorflow/python/__init__.py`. I'll fix if you advise.

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -84,3 +86,49 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
     if staircase:
       p = math_ops.floor(p)
     return math_ops.mul(learning_rate, math_ops.pow(decay_rate, p), name=name)
+
```
mrry: Minor style nit: can you add another blank line here (our linter will complain otherwise...)?

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -84,3 +86,49 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
...
+  # Later, whenever we perform an optimization step, we increment global_step.
+  ```
+  
+  Args:
+    x: A 0-D scalar Tensor. Must be one of the following types: `float32`,
+    `float64`, `uint8`, `int8`, `int16`, `int32`, `int64`.
+    boundaries: A list of `float`s or `int`s with strictly increasing entries.
+    values: A list of `float`s or `int`s that specifies the values for the
+      intervals defined by `boundaries`. (It should therefore have one more
+      element than `boundaries`.
```
mrry: How about accepting a list of scalar tensors for each element of `boundaries` and `values`? We can then use `ops.convert_n_to_tensor()` internally, to avoid converting the scalars so many times, and reduce the number of nodes this function adds to the graph.

Note that I think the elements of `boundaries` have to have the same dtype as `x`, and all elements of `values` must have the same type.

tensorflow/python/training/learning_rate_decay_test.py
```diff
@@ -72,6 +72,30 @@ def testVariables(self):
...
+      self.assertAllClose(pc.eval(), 1.0, 1e-6)
+      assign_105.op.run()
+      self.assertAllClose(pc.eval(), 0.1, 1e-6)
+      assign_110.op.run()
+      self.assertAllClose(pc.eval(), 0.1, 1e-6)
+      assign_120.op.run()
+      self.assertAllClose(pc.eval(), 0.01, 1e-6)
+      assign_999.op.run()
+      self.assertAllClose(pc.eval(), 0.001, 1e-6)
+
```
mrry: Can you add some tests for error conditions (e.g. mismatched types in the list of boundaries, values)?

tensorflow/python/training/learning_rate_decay.py
rdipietro: It seems like `convert_n_to_tensor` simply loops over the list, calling `convert_to_tensor` on each element. Does this really end up adding fewer nodes to the graph?

If not, I suppose it could be useful to accept a list of `Tensor`s, but maybe it'd be worth waiting to see if this is a needed use case? I say this because it's easy to include error checks for Python lists, for example throwing an `InvalidArgumentError` if types don't match, or if `boundaries` isn't strictly increasing, but it's not straightforward (at least for me) to do this error checking with a list of `Tensor`s.

If you'd very much prefer the list-of-`Tensor`s route, can you please point to some code which checks `Tensor`s for conditions and generates an `InvalidArgumentError` if necessary?

rdipietro:
@mrry current solution is to

- Accept lists of `Tensor`s
- Make sure that all `Tensor`s in `boundaries` have the same `dtype` as `x` (with test)
- Make sure that all `Tensor`s in `values` have the same `dtype` (with test)

I haven't made sure that `boundaries`'s entries are strictly increasing. Not sure of the canonical way to do this since `boundaries` is a `Tensor`.

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -120,15 +123,26 @@ def piecewise_constant(x, boundaries, values, name=None):
...
                     name, 'PiecewiseConstant') as name:
+    x = ops.convert_to_tensor(x)
+    # Avoid explicit conversion to x's dtype. This could result in faulty
+    # comparisons, for example if floats are converted to integers.
+    boundaries = ops.convert_n_to_tensor(boundaries)
+    if not all([b.dtype == x.dtype for b in boundaries]):
+      raise ValueError('boundaries must have the same dtype as x.')
+    # TODO(rdipietro): Ensure that boundaries' elements are strictly increasing.
+    values = ops.convert_n_to_tensor(values)
+    if not all([v.dtype == values[0].dtype for v in values]):
```
mrry: Nit: don't need the square brackets as this can be a generator expression:

```python
if not all(v.dtype == values[0].dtype for v in values):
```

tensorflow/python/training/learning_rate_decay.py
```diff
@@ -120,15 +123,26 @@ def piecewise_constant(x, boundaries, values, name=None):
   
   with ops.op_scope([x, boundaries, values, name],
                     name, 'PiecewiseConstant') as name:
+    x = ops.convert_to_tensor(x)
+    # Avoid explicit conversion to x's dtype. This could result in faulty
+    # comparisons, for example if floats are converted to integers.
+    boundaries = ops.convert_n_to_tensor(boundaries)
+    if not all([b.dtype == x.dtype for b in boundaries]):
```
mrry: Same note about generator expressions from below.

mrry:
I think we can defer adding the strictly-increasing assertions for now. Thanks for this!

mrry:

rdipietro:
Sure, glad to contribute.

List vs. generator fixed. Also caught/fixed a bug in the tests, as the different-`values`-dtypes `ValueError` wasn't actually getting tested.

mrry:

vrv:
ok to merge ?

vrv:
learning_rate_decay_test.py appears to be a binary file...

mrry:
LGTM (modulo binary file concerns).

rdipietro:
Can you please elaborate? Is this something I need to fix on my end? I edited the files over an sshfs mount, so I'm wondering if that caused problems.

vrv:
I'm not sure -- if you look at the file diffs, I can't even see the contents of the test code

maniteja123:
Hi, from my previous experience I found that it would have been accidentally made an executable. Changing it using chmod should hopefully solve the issue.

rdipietro:
The problem was that I somehow ended up appending a null character to the end of learning_rate_decay_test.py.

I tried to reset to the previous commit, but the null character remained. Instead I reset back to your original source, copied over all changes to the most recent, and cleanly committed once.

Sorry for the confusion.

vrv:
Cool, looks good.  One more test and we can merge.

PR #2446, opened 2016-05-20
[tf.learn] Fix logistic_regression() summary name conflict.

waleedka:
If you call logistic_regression() more than once in the same model, you get this error.

> Duplicate tag logistic_regression.X found in summary inputs

Fixed by using the variable scope in the summary name.


ilblackdragon:

PR #2468, opened 2016-05-22
Learning rates based on Karpathy CS231n lecture 6

chemelnucfin:
Continuation of #1512 
@vrv seems like I can test now.  Updated the pull request as you suggested.  Please let me know if you have any comments.

### DIFF
--- a/tensorflow/python/training/learning_rate_decay.py
+++ b/tensorflow/python/training/learning_rate_decay.py
@@ -18,6 +18,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+from tensorflow.python.framework import constant_op
 from tensorflow.python.framework import ops
 from tensorflow.python.ops import math_ops
 from tensorflow.python.ops import control_flow_ops
@@ -40,7 +41,7 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
                           decay_rate ^ (global_step / decay_steps)
   ```
 
-  If the argument `staircase` is `True`, then `global_step /decay_steps` is an
+  If the argument `staircase` is `True`, then `global_step / decay_steps` is an
   integer division and the decayed learning rate follows a staircase function.
 
   Example: decay every 100000 steps with a base of 0.96:
@@ -67,15 +68,16 @@ def exponential_decay(learning_rate, global_step, decay_steps, decay_rate,
       Must be positive.  See the decay computation above.
     decay_rate: A scalar `float32` or `float64` `Tensor` or a
       Python number.  The decay rate.
-    staircase: Boolean.  It `True` decay the learning rate at discrete intervals.
-    name: String.  Optional name of the operation.  Defaults to 'ExponentialDecay'
+    staircase: Boolean.  It `True` decay the learning rate at discrete intervals
+    name: String.  Optional name of the operation.  Defaults to 
+      'ExponentialDecay'
 
   Returns:
     A scalar `Tensor` of the same type as `learning_rate`.  The decayed
     learning rate.
   """
   with ops.op_scope([learning_rate, global_step, decay_steps, decay_rate],
-                   name, "ExponentialDecay") as name:
+                    name, "ExponentialDecay") as name:
     learning_rate = ops.convert_to_tensor(learning_rate, name="learning_rate")
     dtype = learning_rate.dtype
     global_step = math_ops.cast(global_step, dtype)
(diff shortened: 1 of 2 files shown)


tensorflow/python/training/learning_rate_decay.py
```diff
@@ -19,7 +19,7 @@
 from __future__ import print_function
 
 from tensorflow.python.framework import ops
-from tensorflow.python.ops import math_ops
+from tensorflow.python.ops import math_ops, constant_op
```
vrv: please import one module per line

tensorflow/python/training/learning_rate_decay_test.py
```diff
@@ -73,5 +75,67 @@ def testVariables(self):
...
 
+class ExponentialDecayTest(test_util.TensorFlowTestCase):
+
+  def testDecay(self):
+    initial_lr = 0.1
+    k = 10
+    decay_rate = 0.96
+    with self.test_session():
+      for step in range(0, k+1):
+        lr = learning_rate_decay.natural_exp_decay(
```
vrv: these tests are creating 10 copies of the graph -- can you rewrite the test so that it creates the graph once and passes the proper parameters as inputs to the function? 


You can see how it's done for the exponential case above: all the ops are created first rather than creating them in a loop.

rwill128:
I find it helpful to phrase commit titles in the imperative, with a verb that describes what merging that commit will do. Basically I try to write commit messages that can complete this sentence: Adding this commit will <do something useful>. Using this commit as an example, it would be helpful for me to know what's being done with the learning rates -- is a new library method being added, or is it becoming the new default learning rate, etc.

I'm a new guy and I submit this as a humble comment.

tensorflow/python/training/learning_rate_decay_test.py
chemelnucfin: @vrv is the way I wrote it now better?  Thanks.

tensorflow/python/training/learning_rate_decay_test.py
```diff
@@ -108,5 +110,75 @@ def testPiecewiseConstantEdgeCases(self):
...
+class ExponentialDecayTest(test_util.TensorFlowTestCase):
+
+  def testDecay(self):
+    initial_lr = 0.1
+    k = 10
+    decay_rate = 0.96
+    with self.test_session():
+      step = state_ops.variable_op([], dtypes.int32)
+      for i in range(k+1):
+        assign = state_ops.assign(step, i)
```
vrv: to make it clear, it is best practice to create the entire graph ahead of time and then invoke it by calling .eval() or sess.run().

This is adding a new op (assign) for every iteration of the loop.

What about moving this statement out of the loop and changing it to assign_add(step, 1) instead ?

chemelnucfin:
@vrv Is this what you mean?

tensorflow/python/training/learning_rate_decay_test.py
```diff
@@ -108,5 +110,82 @@ def testPiecewiseConstantEdgeCases(self):
...
+    initial_lr = 0.1
+    k = 10
+    decay_rate = 0.96
+    step = state_ops.variable_op([], dtypes.int32)
+    assign = state_ops.assign(step, 0)
+    add = state_ops.assign_add(step, 1)
+    with self.test_session():
+      assign.op.run()
+      for i in range(k+1):
+        decayed_lr = learning_rate_decay.natural_exp_decay(initial_lr, step,
```
vrv: This calls natural_exp_delay once for every value of 'i', which means you are still adding 'k' versions of the ops in this loop.

The idea is to define the entire graph completely, before even with self.test_session(), so that the code within self.test_session() is just driving the existing graph, rather than modifying it.  Does that make sense?

tensorflow/python/training/learning_rate_decay_test.py
```diff
@@ -215,5 +217,83 @@ def testBeyondEndWithCycle(self):
...
+    decay_rate = 0.96
+    step = state_ops.variable_op([], dtypes.int32)
+    assign_step = state_ops.assign(step, 0)
+    increment_step = state_ops.assign_add(step, 1)
+    decayed_lr = learning_rate_decay.natural_exp_decay(initial_lr,
+                                                       step,
+                                                       k,
+                                                       decay_rate,
+                                                       staircase=True)
+    with self.test_session():
```
chemelnucfin: @vrv I submitted a new pull request, is that what you mean?  Thanks a lot!

vrv: Hm, I don't see a new pull request! :(

chemelnucfin: @vrv oops, i meant as a force push...

Is this what you mean by moving the decay rate outside?

Thanks!

tensorflow/python/training/learning_rate_decay_test.py
vrv: I still see this being called within the range loop, am I not seeing the latest version?

tensorflow/python/training/learning_rate_decay_test.py
chemelnucfin: @vrv let me check the code again..   I think i didn't change it for all of
the tests, but only for test_staircase..  I'll get back to you when i am
back on my laptop.

chemelnucfin: @vrv apolgies for the confusion.

chemelnucfin: @vrv please try this and let me know if there are any problems!  Thanks for your patience in dealing with this PR.

vrv:
Great!

PR #1771, opened 2016-04-04
Go API

alonsovidales:
Hello,
During the last weeks I have been working on this implementation of the Go API. **This is still a work in progress**, I'm doing this PR just to show what I'm working on and to get feedback.

I forked from https://github.com/tmc/tensorflow/tree/go_bindings where the bindings was implemented by Travis Cline. This Go library allows to interact with the tensors in an easy way from Go being able to create, populate and retrieve the content from the tensors. This library also provides the API for Graph generation from Go, create sessions, run this graphs and so on. The goal is to achieve the same functionality provided by the C++ API.

In order to show how the API works, I updated the label_image example with an implementation using the [Go API](https://github.com/alonsovidales/tensorflow/tree/go_bindings_tensors/tensorflow/examples/label_image_go) and I also updated the [tutorial](https://github.com/alonsovidales/tensorflow/blob/go_bindings_tensors/tensorflow/g3doc/tutorials/image_recognition/index.md#usage-with-the-go-api)

I'm trying to keep all the code covered with tests, I still have to cover some corner cases and functionality like the data type recognition for the graph, etc.
By the moment the supported Tensor data types are:
- TF_FLOAT
- TF_DOUBLE
- TF_INT32
- TF_UINT8
- TF_UINT32
- TF_INT16
- TF_INT8
- TF_STRING
- TF_INT64
- TF_BOOL
  I'm planing to add support for TF_QINT\* TF_BFLOAT16 and TF_COMPLEX soon.

I have some doubts:
- Would it be better to add the API docu as md in tensorflow/g3doc/ , or just with GoDoc it would be enough? Having the docu on the MarkDown files could be better for the people without too much experience on Go, but having the docu on GoDocs is going to be updated in an easiest way and the Go developers are more used to read the documentation on GoDoc format. We can also keep it in both places, or just add a link from the MarkDown docu to GoDoc.
- In order to get the definition for all the available operations I'm loading this info from tensorflow/core/ops/ops.pbtxt . The problem is that I'm doing this [cp of the file from go gen](https://github.com/alonsovidales/tensorflow/blob/go_bindings_tensors/tensorflow/go/gen.go#L3). I think that this problem could be approached on three different ways:
  - 1. Converting ops.pbtxt into a .go file: converting the content into a constant string we don't need to move this file, this option will remove this dependency, but if the file changes it will requiere to regenerate the libs that shouldn't be too much problem, we could even generate Go code for each operation so we don't need to parse the file and it would improve the performance.
[...]

PR #2479, opened 2016-05-23
Go bindings

alonsovidales:


martinwicke:
I merged this without CLA. We checked with all authors, and all are both ok with merging their commits and have CLAs in place.

jmhodges:
This was merged into the `go` branch. I'm guessing that was intentional. Is there another PR to merge the `go` branch into `master`?

alonsovidales:
Hi @jmhodges , yes, this PR comes from this other one: https://github.com/tensorflow/tensorflow/pull/1771
I think that is because the code still has to be integrated on Jenkins and reviewed a bit more. By the moment, I think that there is no other PR to `master`.

martinwicke:
There is no PR to merge it into master right now. More infrastructure work
has to happen before we can merge it to master (from where it would
automatically be sucked into a lot of systems that are not ready for it).

tmc:
@martinwicke it looks like this got merged with 1429abd -- did that lose commit history? I was looking for how much of this still was made up of my initial contribution but it looks like that history may have been lost.

martinwicke:
Yes. I am sorry, I did squash those commits. I should not have. You can
still look at #2479 for the history.

Issue #2463, opened 2016-05-22
Using `tf.nn.rnn_cell.LSTMCell` with `tf.nn.rnn` breaks when `state_is_tuple` is True


mouradmourafiq:
### Environment info
Operating System: mac

If installed from sources, provide the commit hash: 7d9ab3eb485e6eb1778bad4ef01a1cd95b2d22d9

### Steps to reproduce
Creating an rnn based on a list of cell specified by `MultiRNNCell` breaks at the `zero_state` function:

```python
num_units = 3
input_size = 5
num_proj = 4
max_length = 8
sequence_length = [4, 6]
initializer = tf.random_uniform_initializer(-0.01, 0.01)
inputs = max_length * [tf.placeholder(tf.float32, shape=(None, input_size))]
inputs_c = tf.pack(inputs)
cell = tf.nn.rnn_cell.LSTMCell(num_units, input_size, use_peepholes=True,
                               num_proj=num_proj, initializer=initializer, 
                               state_is_tuple=True)
cells = tf.nn.rnn_cell.MultiRNNCell([cell] * 2, state_is_tuple=True)
outputs_static, state_static = tf.nn.rnn(cells, inputs, dtype=tf.float32, 
                                         sequence_length=sequence_length)
```

The same model definition would work if `state_is_tuple=False`, the code break at this [line](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/rnn_cell.py#L107)

The same code would work if we pass the cell directly to `tf.nn.rnn` i.e. without using `tf.nn.rnn_cell.MultiRNNCell`, the reason is that `tf.nn.rnn_cell.MultiRNNCell` would have a `state_size` value as tuple of tuples, in my example the value of the `state_size` is `<type 'tuple'>: ((3, 4), (3, 4))`

### What have you tried?
.1. Tried the same model definition without `state_is_tuple=False`, which works but shows a deprication warning.

```python
num_units = 3
input_size = 5
num_proj = 4
max_length = 8
sequence_length = [4, 6]
initializer = tf.random_uniform_initializer(-0.01, 0.01)
inputs = max_length * [tf.placeholder(tf.float32, shape=(None, input_size))]
inputs_c = tf.pack(inputs)
cell = tf.nn.rnn_cell.LSTMCell(num_units, input_size, use_peepholes=True,
                               num_proj=num_proj, initializer=initializer)
cells = tf.nn.rnn_cell.MultiRNNCell([cell])
[...]

PR #2490, opened 2016-05-24
Modify zero_state to work with MultiRNNCell when state_is_tuple is True

jihunchoi:
#2463

### DIFF
--- a/tensorflow/tools/ci_build/builds/integration_tests.sh
+++ b/tensorflow/tools/ci_build/builds/integration_tests.sh
@@ -108,10 +108,10 @@ mkdir -p "${TEST_DIR}" || \
 test_ffmpeg_lib() {
   # If FFmpeg is not installed then run a test that assumes it is not installed.
   if [[ -z "$(which ffmpeg)" ]]; then
-    bazel test tensorflow/contrib/ffmpeg/kernels:ffmpeg_lib_uninstalled_test
+    bazel test tensorflow/contrib/ffmpeg/default:ffmpeg_lib_uninstalled_test
     return $?
   else
-    bazel test tensorflow/contrib/ffmpeg/kernels:ffmpeg_lib_installed_test \
+    bazel test tensorflow/contrib/ffmpeg/default:ffmpeg_lib_installed_test \
         tensorflow/contrib/ffmpeg:decode_audio_op_test
     return $?
   fi


jihunchoi:
@googlebot I signed a CLA.

vrv:
Thanks for the contribution, can you add a test for these changes so that we don't regress in the future?

jihunchoi:
Implemented a test on my implementation.
Since the current version of `testBasicLSTMCellWithStateTuple` in rnn_cell_test.py is not properly using state_is_tuple, (it is using state_is_tuple only on `MultiRNNCell`, not on `BasicLSTMCel`l), I modified `testBasicLSTMCellWithStateTuple` to use state_is_tuple on both cells.

vrv:
Looks good!  Thanks.  @tensorflow-jenkins test this please

tensorflow/models/rnn/ptb/ptb_word_lm.py
```diff
@@ -245,13 +245,24 @@ def run_epoch(session, m, data, eval_op, verbose=False):
...
+    fetches.append(eval_op)
+    for c, h in m.final_state:
+        fetches.append(c)
+        fetches.append(h)
+    feed_dict = {}
+    feed_dict[m.input_data] = x
+    feed_dict[m.targets] = y
+    for i, (c, h) in enumerate(m.initial_state):
+        feed_dict[c], feed_dict[h] = state[i]
+    cost, _, *states = session.run(fetches, feed_dict)
```
vrv: I think this is an error according to our linter

jihunchoi: My text editor inserted 4 spaces for indent instead of 2 spaces...
Updated the codes, and checked that it passes the sanity checks in my local machine.

jihunchoi: Wait; it only runs on Python 3.
I will modify soon.

jihunchoi: I've just modified the codes to be compatible with Python 2 and checked that it passes sanity checks on my 2.7.11 and 3.5.1.
Can you test via jenkins please?

vrv:
Sure, will kick off a test:  @tensorflow-jenkins test this please

@ebrevdo should still look at this though

vrv:
Some test failures -- you can test locally by running bazel test tensorflow/... to make sure you catch all failures.

ebrevdo:
We have a change internally that should fix nested tuples.  It obsoletes
the rnn changes in this pr.  At our next public push you should rebase and
we can see about the ptb example updates.

jihunchoi:
Wow, the test works properly after completely reinstalling the whole library...

I didn't know that this is internally fixed, so I will left only PTB examples.

Issue #2525, opened 2016-05-26
the cudnn lib directory is not in the tensorflow gpu docker image's LD_LIBRARY_PATH


yeison:
libcudnn4 is installed in the gcr.io/tensorflow/tensorflow:latest-gpu docker image, but I get the following errors from the jupyter tensorflow tutorial:

F tensorflow/stream_executor/cuda/cuda_dnn.cc:204] could not find cudnnCreate in cudnn DSO; dlerror: /usr/local/lib/python2.7/dist-packages/tensorflow/python/_pywrap_tensorflow.so: undefined symbol: cudnnCreate

running env shows that LD_LIBRARY_PATH does not have a reference to the libcudnn.so location 

```
LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64:
```

yeison:
https://github.com/tensorflow/tensorflow/pull/2527

yeison:
See pull request for on simple fix: https://github.com/tensorflow/tensorflow/pull/2527

yeison:
More detail here: https://github.com/NVIDIA/nvidia-docker/issues/45

flx42:
That's because `LD_LIBRARY_PATH` is not the only path used for lookup.
The library is present in `/usr/lib/x86_64-linux-gnu/libcudnn.so.4`, it's just that the `dlopen` should be using the `SONAME`.

3XX0:
Probably a issue with the configuration script again.

`GetCudnnVersion` needs to return `4`, see:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/stream_executor/dso_loader.cc#L57
https://github.com/tensorflow/tensorflow/blob/14ac2235699509f512b44b71160239c153ab413d/configure#L243

And you might want to hardcode `"1"` for the CUDA library too
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/stream_executor/dso_loader.cc#L75

jendap:
It does not return 4. It will return empty string. It should (probably) look for /usr/lib/x86_64-linux-gnu/libcudnn.so

@3XX0 what waas the story why the container doe not contain a symlink from `/usr/lib/x86_64-linux-gnu/libcudnn.so.4` to `/usr/lib/x86_64-linux-gnu/libcudnn.so`?

[...]

PR #2527, opened 2016-05-27
fix missing cudnn dependencies

yeison:
This is the simplest solution to fix missing cudnn dependencies.  Just use the devel image from nvidia/cuda


yeison:
issue 2525

yeison:
https://github.com/tensorflow/tensorflow/issues/2525

vrv:
@tensorflow-jenkins test this please  (not sure if our tests require docker)

vrv:
I'm optimistically merging this.  We can rollback if there are problems.

flx42:
But the library is **already** present in the `runtime` image, see my comment:
https://github.com/tensorflow/tensorflow/issues/2525#issuecomment-223170383

jendap:
wait, why? that would make the image large. we should change it back

PR #2529, opened 2016-05-27
[tf.learn] logic for pandas import and fixed test

terrytangyuan:
1. pandas should be imported by users when needed
2. fixed test_categorical.py when pandas is installed


terrytangyuan:
@martinwicke Seems like tests are running automatically for me now?

martinwicke:
Someone may have fixed auth, if that's the case, thank you, @jendap?

jendap:
I have not changed anything. But the email address in the commits has domain @users.noreply.github.com so maybe the "pull request builder plugin" will then resolve properly the user belonging to tensorflow org.

Actually I believe the domain does not matter. I guess it will take the user name before domain and match that agains the list of users ion tensorflow org. Hence if someone has martinwicke on github and commit as wicke@ then it will not recognize. (but of course wicke is admin so for that user it does not matter whether he is resolved to be in the org or not)

That would also explain why @zheng-xq had to be added to admins. He has different user name in email then on github.

jendap:
Ok, Gunan has looked up the source code of the jenkins plugin. It is not cheating with user names. It asks github api. All public members of tensorflow org should build automatically. Hence this has to return 2xx

curl -s -o /dev/null -w "%{http_code}" https://api.github.com/organizations/15658638/public_members/terrytangyuan

terrytangyuan:
Thanks for looking into it!

PR #2537, opened 2016-05-27
Enable complex number types for tile op

ibab:
I've enabled `complex64` and `complex128` on the CPU for `tf.tile` and added them to the tests.


vrv:

PR #2541, opened 2016-05-27
Add meshgrid function

ibab:
I needed to use something similar to [`np.meshgrid`](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.meshgrid.html) in tensorflow and decided to write my own version.
This can be implemented in pure Python by making use of `tf.reshape` and `tf.tile`.
I've placed the function in `array_ops.py` and added a simple test that compares the output to the numpy version.

One thing that's still missing is testing whether the inputs all have rank 1.
Seems like I might have to use `tf.rank` to check this at runtime.
Would `Assert` ops be the right tool for that?

### DIFF
--- a/tensorflow/python/kernel_tests/array_ops_test.py
+++ b/tensorflow/python/kernel_tests/array_ops_test.py
@@ -157,7 +157,7 @@ class MeshgridTest(test_util.TensorFlowTestCase):
   def _compare(self, n, np_dtype, use_gpu):
     inputs = []
     for i in range(n):
-      x = np.linspace(-10, 10, 5, dtype=np_dtype)
+      x = np.linspace(-10, 10, 5).astype(np_dtype)
       if np_dtype in (np.complex64, np.complex128):
         x += 1j
       inputs.append(x)
--- a/tensorflow/python/kernel_tests/array_ops_test.py
+++ b/tensorflow/python/kernel_tests/array_ops_test.py
@@ -166,7 +166,6 @@ class MeshgridTest(test_util.TensorFlowTestCase):
     with self.test_session(use_gpu=use_gpu):
       tf_out = array_ops.meshgrid(*inputs)
       for X, _X in zip(numpy_out, tf_out):
-        print(X, _X)
         self.assertAllEqual(X, _X.eval())
 
   def testCompare(self):
@@ -178,8 +177,12 @@ class MeshgridTest(test_util.TensorFlowTestCase):
       self._compare(4, t, False)
       self._compare(5, t, False)
 
-    with self.assertRaises(ValueError):
-      self._compare(6, np.float32, False)
+    # Test for inputs with rank not equal to 1
+    x = [[1, 1], [1, 1]]
+    with self.assertRaises(errors.InvalidArgumentError):
+      with self.test_session():
+        X, _ = array_ops.meshgrid(x, x)
+        X.eval()
 
 
 if __name__ == "__main__":
--- a/tensorflow/python/kernel_tests/array_ops_test.py
+++ b/tensorflow/python/kernel_tests/array_ops_test.py
@@ -152,5 +152,35 @@ class ReverseTest(test_util.TensorFlowTestCase):
         self.assertAllEqual(x_tf, np.asarray(x_np)[::-1])
 
 
+class MeshgridTest(test_util.TensorFlowTestCase):
+
+  def _compare(self, n, np_dtype, use_gpu):
+    inputs = []
+    for i in range(n):
+      x = np.linspace(-10, 10, 5, dtype=np_dtype)
+      if np_dtype in (np.complex64, np.complex128):
+        x += 1j
+      inputs.append(x)
+
+    numpy_out = np.meshgrid(*inputs)
+    with self.test_session(use_gpu=use_gpu):
+      tf_out = array_ops.meshgrid(*inputs)
+      for X, _X in zip(numpy_out, tf_out):
+        print(X, _X)
+        self.assertAllEqual(X, _X.eval())
+
+  def testCompare(self):
+    for t in (np.float16, np.float32, np.float64, np.int32, np.int64,
+            np.complex64, np.complex128):
+      self._compare(1, t, False)
+      self._compare(2, t, False)
+      self._compare(3, t, False)
+      self._compare(4, t, False)
+      self._compare(5, t, False)
+
+    with self.assertRaises(ValueError):
+      self._compare(6, np.float32, False)
+
+
 if __name__ == "__main__":
   googletest.main()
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/array_ops.py
```diff
@@ -986,6 +987,62 @@ def pad(tensor, paddings, mode="CONSTANT", name=None):  # pylint: disable=invali
...
+
+  Args:
+    *args: `Tensor`s with rank 1
+
+  Returns:
+    outputs: A list of N `Tensor`s with rank N
+
+  """
+  if indexing not in ("xy", "ij"):
+    raise ValueError("indexing parameter must be either 'xy' or 'ij'")
```
girving: What do those mean?  We should probably document that above.

tensorflow/python/ops/array_ops.py
```diff
@@ -986,6 +987,62 @@ def pad(tensor, paddings, mode="CONSTANT", name=None):  # pylint: disable=invali
...
+
+  num_inputs = len(args)
+
+  if num_inputs > 5:
+    raise ValueError("meshgrid only supports up to 5 dimensions")
```
girving: I think it's fine to drop this assertion and let tile break or not as it pleases.  In particular, if tile becomes less constrained later it'd be nice if this routine didn't need to change.

tensorflow/python/ops/array_ops.py
```diff
@@ -986,6 +987,62 @@ def pad(tensor, paddings, mode="CONSTANT", name=None):  # pylint: disable=invali
...
+
+  ones = (1,) * num_inputs
+
+  with ops.op_scope(args, name, "meshgrid") as name:
+    # Prepare reshape by inserting dimensions with size 1 where needed
+    shapes = [ones[:i] + (-1,) + ones[i + 1:] for i in range(num_inputs)]
+    # Create parameters for broadcasting each tensor to the full size
+    sizes = [shape(x)[0] for x in args]
```
girving: sizes = [size(x) for x in args]

is faster since we know x has rank 1 (at least in the correct case).

tensorflow/python/kernel_tests/array_ops_test.py
```diff
@@ -152,5 +152,35 @@ def testReverse1DimAuto(self):
...
+      x = np.linspace(-10, 10, 5, dtype=np_dtype)
+      if np_dtype in (np.complex64, np.complex128):
+        x += 1j
+      inputs.append(x)
+
+    numpy_out = np.meshgrid(*inputs)
+    with self.test_session(use_gpu=use_gpu):
+      tf_out = array_ops.meshgrid(*inputs)
+      for X, _X in zip(numpy_out, tf_out):
+        print(X, _X)
```
girving: Remove print before submitting.

girving:
Thanks!  Yes, `tf.Assert` is the right tool.  I'm tempted to not require it since it mucks up the code quite a bit and doesn't affect the correct path, but I agree it's the right thing to do.

tensorflow/python/ops/array_ops.py
ibab: I ported these from the numpy function.
The `'xy'` stands for cartesian indexing, where the first two indices are swapped.
I'll add this to the docstring.

ibab:
Thanks for the comments!
I've pushed a new commit that addresses them.
I'm requiring all `tf.Assert`s for each of the outputs in case the user doesn't use all of them.

tensorflow/python/ops/array_ops.py
```diff
@@ -1040,7 +1047,12 @@ def meshgrid(*args, indexing="xy", name="meshgrid"):
       shapes[0], shapes[1] = shapes[1], shapes[0]
       bcast[0], bcast[1] = bcast[1], bcast[0]
 
-    return [tile(reshape(a, r), e) for a, r, e in zip(args, shapes, bcast)]
+    results = []
+    for a, r, e in zip(args, shapes, bcast):
+      with ops.control_dependencies(asserts):
```
girving: Move the ops.control_dependencies outside the while loop.

tensorflow/python/kernel_tests/array_ops_test.py
```diff
@@ -178,8 +177,12 @@ def testCompare(self):
       self._compare(4, t, False)
       self._compare(5, t, False)
 
-    with self.assertRaises(ValueError):
-      self._compare(6, np.float32, False)
+    # Test for inputs with rank not equal to 1
+    x = [[1, 1], [1, 1]]
+    with self.assertRaises(errors.InvalidArgumentError):
```
girving: Use `self.assertRaisesRegexp` to verify you're catching the expected error string.

girving:
Cool, two more small comments and then I'll fire off the tests.

ibab:
Thanks!
I've pushed another commit.

girving:
Jenkins, test this please.

girving:
@martinwicke: Do you understand the rather unhelpful "invalid syntax" lint error?  The line looks perfectly reasonable to me.

ibab:
Just noticed that one of the new lines was longer than 80 characters

ibab:
@chrisburr just pointed out that you can't use `*args` with explicit keyword arguments in Python 2,
and that's what's causing the syntax error.
I'll update the PR to use `**kwargs`.

ibab:
Okay, I've adjusted the code to work with Python 2 and improved the docstring slightly.
This should be ready to go now.

tensorflow/python/ops/array_ops.py
```diff
@@ -1023,6 +1023,9 @@ def meshgrid(*args, indexing="xy", name="meshgrid"):
   Returns:
     outputs: A list of N `Tensor`s with rank N
   """
+  indexing = kwargs.get("indexing", "xy")
+  name = kwargs.get("name", "meshgrid")
```
girving: We should bail if there are unknown keyword args.  Not sure how best to express that.

girving:
Currently unknown keyword args are silently dropped; we should report an error instead.  It's unfortunate that Python 2 doesn't have keyword-only arguments.

ibab:
Yeah, it's always sad when this happens.
I've pushed a new commit that tries to mimic the error that Python usually raises.

girving:
Thanks!  Jenkins, test this please.

girving:
Jenkins, test this please.

girving:
@caisq, @gunan: Do you know what's going on with the ci.tensorflow.org failure?

ibab:
Looks like Jenkins is healthy again.
Can someone re-run the tests?

girving:
Jenkins, test this please.

ibab:
Seems like Jenkins is using a numpy version before `1.9.0` that doesn't have the `dtype` kwarg for `np.linspace`

ibab:
I've pushed a new commit that should fix the failing test

girving:
Jenkins, test this please.

girving:
@ibab: Looks like you'll have to drop the 1 argument test since numpy doesn't support it for the version Jenkins uses.  Ours should, though.

ibab:
Okay, I've removed the 1 dim case.

girving:
Jenkins, test this please.

girving:
@vrv: Yes!  Thank you for yet another contribution @ibab!

PR #2544, opened 2016-05-27
Updated Eigen version in Makefile

petewarden:
This cleans up assignment in the makefile, and switches to the latest Eigen version.


vrv:
LGTM, merging.

PR #2545, opened 2016-05-28
Work around for problem with Thumb linking

petewarden:
There's an occasional problem with linking armv7 code with the makefile, which results in a bogus linker error. One way around it is to disable Thumb code. See https://discussions.apple.com/thread/6037322?tstart=0 for more details.


petewarden:
Jenkins, test this please.

Issue #1727, opened 2016-03-31
GPU resources not released when session is closed


gustavla:
As I understand from the documentation, running `sess.close()` is supposed to release the resources, but it doesn't. I have been running the following test:

``` python
with tf.Session() as sess:
    with tf.device('/gpu:0'):
        matrix1 = tf.constant([[3., 3]])
        matrix2 = tf.constant([[2.], [2.]])
        product = tf.matmul(matrix1, matrix2)
        result = sess.run(product)
        print(result)
```

This allocates all the free memory of gpu0, but it is not released when `sess` is closed (both using a context manager as in the code above, but also when calling `sess.close()` manually). The memory usage persists until that Python process is terminated. The way I have been checking memory usage is through `nvidia-smi`, but I have also confirmed that other processes can't allocate that GPU memory until the process terminates (not the session closes). I would like to be able to free the resources and still keep the Python process running.
### Environment info

I am running a 64-bit Linux (CentOS) with a computer that has two Tesla K40c (driver 346.46, CUDA 7.0). I installed the 0.7.1 tensorflow for Linux and Python 3.4 through pip. The output of `tf.__version__` is 0.7.1.
### Steps to reproduce

Simply running the code above should according to the document allocate and then release the memory. However, the GPU memory is still allocated and thus unusable by other processes. However, it can be re-used by the same Python process, meaning that I can re-run the snippet over and over as long as I do it from the same Python process.
### Logs or other output that would be helpful

Here is a log of the session. At the end, the memory is still allocated. Note that another user is connected to both GPUs through Torch7, and is actively using gpu0.

``` python
In [1]: import tensorflow as tf
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcublas.so locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcudnn.so locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcufft.so locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcuda.so.1 locally
I tensorflow/stream_executor/dso_loader.cc:105] successfully opened CUDA library libcurand.so locally

In [2]: with tf.Session() as sess:
    with tf.device('/gpu:1'):
        matrix1 = tf.constant([[3., 3]])
        matrix2 = tf.constant([[2.], [2.]])
        product = tf.matmul(matrix1, matrix2)
        result = sess.run(product)
[...]

PR #2555, opened 2016-05-28
cleans up warning/errors tensorflow/stream_executor 

Mistobaan:
- the delete variables are not used anywhere (why keep them around?)
- tests seem to be running as before.

### DIFF
--- a/tensorflow/python/ops/rnn.py
+++ b/tensorflow/python/ops/rnn.py
@@ -126,7 +126,8 @@ def rnn(cell, inputs, initial_state=None, dtype=None,
       state = initial_state
     else:
       if not dtype:
-        raise ValueError("If no initial_state is provided, dtype must be.")
+        raise ValueError("If no initial_state is provided, "
+                           "dtype must be specified")
       state = cell.zero_state(batch_size, dtype)
 
     if sequence_length is not None:  # Prepare variables
--- a/tensorflow/stream_executor/cuda/cuda_activation.cc
+++ b/tensorflow/stream_executor/cuda/cuda_activation.cc
@@ -27,8 +27,7 @@ CudaContext* ExtractCudaContext(CUDAExecutor *cuda_exec);
 CUDAExecutor *ExtractCudaExecutor(StreamExecutor *stream_exec);
 
 ScopedActivateExecutorContext::ScopedActivateExecutorContext(
-    CUDAExecutor *cuda_exec)
-    : cuda_exec_(cuda_exec),
+    CUDAExecutor *cuda_exec):
       driver_scoped_activate_context_(
           new ScopedActivateContext{ExtractCudaContext(cuda_exec)}) { }
 
--- a/tensorflow/stream_executor/cuda/cuda_dnn.cc
+++ b/tensorflow/stream_executor/cuda/cuda_dnn.cc
@@ -457,6 +457,7 @@ class ScopedFilterDescriptor {
                  << ToString(status);
     }
 
+#if CUDNN_VERSION >= 5000
     // TODO(b/23032134): Even if the filter layout is not supported,
     // cudnnSetFilter4DDescriptor_v4 will return CUDNN_STATUS_SUCCESS because it
     // does not take layout as an input. Maybe force cuDNN by giving wrong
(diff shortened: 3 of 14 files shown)


tensorflow/stream_executor/cuda/cuda_activation.cc
```diff
@@ -27,8 +27,7 @@ CudaContext* ExtractCudaContext(CUDAExecutor *cuda_exec);
 CUDAExecutor *ExtractCudaExecutor(StreamExecutor *stream_exec);
 
 ScopedActivateExecutorContext::ScopedActivateExecutorContext(
-    CUDAExecutor *cuda_exec)
-    : cuda_exec_(cuda_exec),
+    CUDAExecutor *cuda_exec):
```
vrv: I don't think this is what clang-format would do, revert?

vrv: Actually I'm not sure why you made this change at all...

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -457,6 +457,7 @@ class ScopedFilterDescriptor {
                  << ToString(status);
     }
 
+#if CUDNN_VERSION >= 5000
```
vrv: Why is this needed?  We build with cudnnv4 right now and nothing was failing...

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -666,7 +668,7 @@ class ScopedActivationDescriptor {
         mode = CUDNN_ACTIVATION_TANH;
         break;
       default:
-        LOG(ERROR) << "unrecognized activation mode: "
+        LOG(FATAL) << "unrecognized activation mode: "
```
vrv: Let's revert this -- we don't want to crash the program -- we'd prefer to return a bad status message.

tensorflow/stream_executor/kernel.h
```diff
@@ -349,6 +349,7 @@ class TypedKernel : public KernelBase {
         port::InlinedVector<uint8, 4>{arg_ptr, arg_ptr + sizeof(arg)}, sizeof(arg)});
   }
 
+#ifndef GOOGLE_CUDA
```
vrv: this looks wrong -- or rather, there's a bug somewhere else that needs to be fixed, rather than adding this here

tensorflow/stream_executor/cuda/cuda_dnn.cc
Mistobaan: otherwise for CUDNNv4 and less  the `format` variable will be defined but not used (see the other block below:
```c++
    status = dynload::cudnnSetFilterNdDescriptor(parent_, handle_, elem_type,
#if CUDNN_VERSION >= 5000
                                                 format,
#endif
                                                 dims.size(), dims.data());
```

tensorflow/stream_executor/cuda/cuda_dnn.cc
Mistobaan: ok

tensorflow/stream_executor/kernel.h
Mistobaan: removed

tensorflow/stream_executor/cuda/cuda_activation.cc
Mistobaan: @vrv I think this is suspicious, not sure who owns the stream_executor but could it be that because we are not deleting the cuda_exec once it scopes out, will cause #1727 ? I did not have time to dig in the code to find out .

vrv: I don't think this object owns the cuda_exec pointer, so it's probably fine.  I guess it was just unused.

tensorflow/stream_executor/cuda/cuda_dnn.cc
vrv: Good point, this class is not written very well.  One day we should probably refactor this so that we can check the status rather than just kill the program.

tensorflow/python/ops/rnn.py
```diff
@@ -126,7 +126,8 @@ def rnn(cell, inputs, initial_state=None, dtype=None,
       state = initial_state
     else:
       if not dtype:
-        raise ValueError("If no initial_state is provided, dtype must be.")
+        raise ValueError("If no initial_state is provided, "
+                           "dtype must be specified")
```
vrv: alignment is messed up here -- this seems unrelated to this commit, maybe put in a different PR?

vrv:

martinwicke:
Looks ready to merge, vrv@?

PR #2556, opened 2016-05-28
cuda 8 api uses cudaDataType for 16bit float type

kashif:


tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -1680,10 +1680,25 @@ bool CUDABlas::DoBlasGemm(
   return DoBlasInternal(
       dynload::cublasSgemmEx, stream, true /* = pointer_mode_host */,
       CUDABlasTranspose(transa), CUDABlasTranspose(transb), m, n, k, &alpha,
-      CUDAMemory(a), CUBLAS_DATA_HALF, lda,
-      CUDAMemory(b), CUBLAS_DATA_HALF, ldb,
-      &beta,
-      CUDAMemoryMutable(c), CUBLAS_DATA_HALF, ldc);
+      CUDAMemory(a),
```
vrv: I wonder if it makes sense defining our own "half type" somewhere in another header that we typedef once, so we centralize the definition of the symbol.  Thoughts @zheng-xq  ?

kashif: yes or I can also `#define CUBLAS_DATA_HALF CUDA_R_16F` for `CUDA_VERSION >= 8000` which might make the code look a bit cleaner...

vrv: That seems fine to me

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -25,6 +25,10 @@ limitations under the License.
 #define EIGEN_HAS_CUDA_FP16
 #endif
 
+#if CUDA_VERSION >= 8000
+#define CUBLAS_DATA_HALF CUDA_R_16F
+#endif
+
```
zheng-xq: This seems fine with me too. 

The downside is that it introduces an outdated symbol back into CUBLAS space, which might interact with other code when we move it around. I would second vrv@'s suggestion that it is better to have our own symbol that is conditionally defined.

I will leave it up to you whether you want to follow through in this CL.

zheng-xq:

zheng-xq:
@vrv, @martinwicke, I've confirmed that with this patch, TensorFlow would build with Cuda 8. Feel free to merge this change. We can always improve it later.

vrv:
test this please

tensorflow/stream_executor/cuda/cuda_blas.cc
kashif: I will make the change in another pull request if that is ok?

vrv:
@martinwicke @caisq our Jenkins is broken   

Merging since this change was verified by @zheng-xq

kashif:
@zheng-xq  would something like:

```c++
#if CUDA_VERSION >= 8000
#define TF_CUDA_DATA_HALF CUDA_R_16F
#else
 #define TF_CUDA_DATA_HALF CUBLAS_DATA_HALF
#endif
```

and then I use `TF_CUDA_DATA_HALF` in the `cublasSgemmEx()`?

zheng-xq:
@kashif, thanks for getting to this quickly. 

Please name it SE_CUDA_DATA_HALF, since this is stream-executor code, which TensorFlow only hosts a fork temporarily. Other than that, it looks fine.

jendap:

jendap:
@kashif can you rename it please?

jendap:
Thank you @kashif!

kashif:
thank you! @jendap

Issue #2476, opened 2016-05-23
Feature request: RMSPropOptimizer support for sparse gradients


333caowei:
I use RMSPropOptimizer to optimize parameters, I get NotImplementedError. **But when I change to use AdamOptimizer, it works fine.**

So I try to fix the problem and I find some key point, may be it can help.
Here is my code:
```
self.x = tf.placeholder(tf.int32, [None, sequence_length])  
point = tf.get_variable([len(embedding_matrix)])
............
............
outputs, states = rnn.rnn(lstm_cell, x, initial_state=initial_state, sequence_length=real_length)
index     =    self.x[:, 0]
index  = tf.reshape(index, [-1,1])
index_point  =    tf.gather(pointt, index)
output   =  tf.mul(outputs[-1] ,   index_point)

scores = tf.nn.xw_plus_b(output, self.W, b)
losses = tf.nn.softmax_cross_entropy_with_logits(scores, self.input_y)
self.loss = tf.reduce_mean(losses) 
optimizer = tf.train.RMSPropOptimizer(1e-3, decay = 0.9)
grads_and_vars = optimizer.compute_gradients(self.loss)
self.train_op = optimizer.apply_gradients(grads_and_vars)
```
When I try to change the line ```output   =  tf.mul(outputs[-1] ,   index_point)``` to some others such as ```output   =  tf.mul(outputs[-1] ,   2)```, **the error disappear**. And I try to change to use ```bidirectional_rnn```, **the error also disappear.**

vrv:
This is known: there is no implementation of RMSPropOptimizer for sparse gradients.  I'll retitle to a feature request.

suiyuan2009:
I'll work on it

PR #2564, opened 2016-05-29
add sparse gradients support for RMSPropOptimizier

suiyuan2009:
#2476


tensorflow/core/ops/ops.pbtxt
```diff
@@ -11111,6 +11111,104 @@ op {
   description: "That is for rows we have grad for, we update var and accum as follows:\n\naccum = accum * momentum + grad\nvar -= lr * accum"
 }
 op {
+  name: "SparseApplyRMSProp"
```
vrv: you can drop this change -- we automatically update the ops now, I believe

tensorflow/python/training/rmsprop_test.py
```diff
@@ -23,9 +23,132 @@
 import numpy as np
 import tensorflow as tf
 
+def rmsprop_update_numpy(var, g, rms, mom, lr, decay, momentum, epsilon):
```
vrv: you can define this within the class RMSPropOptimizerTest

lukaszkaiser:
George Dahl agreed to take a look, he's just joining the right groups on github.

georgedahl:
I took a very quick glance and I believe this PR does the "wrong" thing efficiently, however it is a pretty reasonable "wrong" thing to do. What I mean is that for sparse updates, I don't think this will behave the same was as treating them as dense would. So doing the same optimization, but treating the updates as sparse will produce different results than treating them as dense, even if all the gradients are the same. This property needs to be documented, because one might somehow assume that using the sparse version would do the same thing as the dense one, just more efficiently when you have sparse updates.

Essentially, RMSProp and similar algorithms generate an update for each weight at each step and the approach here discards the updates for weights that have a zero gradient on that step. This gives the rarely updated indices a lower learning rate than they would otherwise have in the dense version of the algorithm since "correct" RMSprop would divide by gradient statistics and amplify the updates for weights that had many zero gradients. However, that said, for sparse updates this might work ok sometimes and it isn't clear what the best thing to do is, so doing the easy and efficient thing could be worth implementing.

I am pretty sure Adam in TF for sparse updates has the same problem as exhibited in this PR, and since this PR was for a feature request spawned by the inconsistency between ADAM supporting sparse updates and RMSProp not supporting them, it makes sense to have an analogous behavior. Of course it would be good to add something to the ADAM docs explaining that for sparse updates it only updates weights with non-zero gradient that step and makes no attempt at correcting for this.

So to summarize, the basic approach of the PR seems OK, but I would like to see the exact algorithm documented better in the sparse case so users can be warned not to think it works the same way as the dense update algorithm. ADAM might also need a doc improvement.

tensorflow/core/kernels/training_ops.cc
```diff
@@ -1428,4 +1428,140 @@ REGISTER_KERNELS(GPU, double);
...
+      auto ms_flat = ms.flat_outer_dims<T>();
+      auto mom_flat = mom.flat_outer_dims<T>();
+      auto grad_flat = grad.flat_outer_dims<T>();
+      const T lr_scalar = lr.scalar<T>()();
+      const T rho_scalar = rho.scalar<T>()();
+      const T epsilon_scalar = epsilon.scalar<T>()();
+      const T momentum_scalar = momentum.scalar<T>()();
+
+      for (Tindex i = 0; i < N; i++) {
+	const Tindex index = indices_vec(i);
```
vrv: can you run clang-format (style google) on this file to fix a lot of the alignment issues in this file?

suiyuan2009:
@georgedahl thank you very much! I took a look at 'RMSProp algorithm with Nesterov momentum' presented in Bengio's 'deep learning', you can record a variable like timestamp to calculate how many times you have not multiplied decay rate, but the simple implement is good enough maybe.

tensorflow/g3doc/api_docs/python/functions_and_classes/tf.train.AdamOptimizer.md
```diff
@@ -33,6 +33,10 @@ The default value of 1e-8 for epsilon might not be a good default in
 general. For example, when training an Inception network on ImageNet a
 current good choice is 1.0 or 0.1.
 
+Note that in dense implement of this algorithm, m_t, v_t and variable will
```
vrv: you can drop the changes to the .md files, they are autogenerated

vrv:
Sorry, you should comment when you've pushed new changes, otherwise we don't know when the code is ready.  You'll have to resolve conflicts and then update us when you've done so.

suiyuan2009:
@vrv modification of .md has been removed and conflicts have been solved.
but merges too many commits and someone has not signed CLA

vrv:
I think you rebased incorrectly, can you try again?

suiyuan2009:
@vrv seems ok now

vrv:

suiyuan2009:
@vrv MacOS CPU Tests failed, but Linux CPU Tests succeeded

```
//tensorflow/core:common_runtime_direct_session_with_tracking_alloc_test FAILED in 0.1s
  /private/var/tmp/_bazel_jenkins/13e370a18c169b19baeafefb05212b85/tensorflow-pull-requests-mac/bazel-out/local_darwin-opt/testlogs/tensorflow/core/common_runtime_direct_session_with_tracking_alloc_test/test.log

Executed 402 out of 517 tests: 516 tests pass and 1 fails locally.
```

vrv:
I think it's a flaky test, i will try again  @tensorflow-jenkins test this please

PR #2574, opened 2016-05-29
Remove unnecessary check from docker_run_gpu.sh

moof2k:
docker_run_gpu.sh was checking for /usr/local/cuda, which is not the default install location on ubuntu 16.04. The check was not being used by anything else anyways, so removing.


moof2k:
I signed it!

vrv:

vrv:

PR #2578, opened 2016-05-30
Support GIF image decode ops by FreeImage library

buaaliyi:
Use FreeImage library to decode GIF images. This is the initial commit with simple implementation first, will add more features and optimize later. Thanks.


buaaliyi:

martinwicke:
We're not comfortable with the FreeImage dependency. Is there a gif reader library we can use instead?

buaaliyi:
@martinwicke  
For pure gif decoder C++ library, there are:

giflib  http://giflib.sourceforge.net/
libnsgif  http://www.netsurf-browser.org/projects/libnsgif/

to be selected from.

martinwicke:
giflib it is. Can you use that instead? Thanks!

buaaliyi:
OK, I'll use giflib to refactor this code, and submit another PR after future several days. Decoding gif image files is really important in my tensorflow use case.

At the same time, thank you for your advise.

martinwicke:
Closing this in favor of #3020.

PR #2579, opened 2016-05-30
fixed atrous_conv2d to work with unknown shape of the input tensor

ThomasWollmann:


ThomasWollmann:
I signed it!

martinwicke:
This is great. Could you add a test to make sure it keeps working?

ThomasWollmann:
The test should still work. I ran them and fixed some imports. Still they are failing, because the calculated padding is somehow wrong. Currently I have no time to investigate. Maybe someone sees the cause immediately.

shampool:
The current atrous_conv2d seems requiring image size should be divisible by the rate.
```
InvalidArgumentError: Image height 232 and width 248should be divisible by block_size: 16
	 [[Node: atrous_conv2d_7/SpaceToBatch = SpaceToBatch[T=DT_FLOAT, block_size=16, _device="/job:localhost/replica:0/task:0/cpu:0"](transpose_21, atrous_conv2d_7/concat)]]
Caused by op u'atrous_conv2d_7/SpaceToBatch', defined at:
```

ThomasWollmann:
Thanks @shampool! Works for me.

vrv:
sorry, some conflicts have sprung up, can you rebase?

ThomasWollmann:
Seems like somebody else applied a fix in the meanhwile: https://github.com/tensorflow/tensorflow/commit/3a744a417178646712e9cda0a85655d5d8968e19

PR #2592, opened 2016-05-31
dynamic split

wan-wei:


vrv:
(Reopen if you can sign the CLA)

j-wilson:
Am I correct in assuming that this PR provided a variant of `tf.split` which supported dynamic arguments --- i.e. for which arguments `num_or_size_splits` and/or `num` can be integer-typed tensors? 

If so, did something along these lines ever end up being incorporated into TensorFlow (esp. a differentiable version thereof)?  I'm aware of `tf.dynamic_partition`; but, am under the impression that a `dynamic_split` method would be decidedly more efficient (owing to implicit contiguity assumptions).

Issue #2586, opened 2016-05-30
Persistent tensors get placed onto GPU unexpectedly


yaroslavvb:
Running `get_session_handle` ops seems to affect where subsequent `get_session_tensor` is placed.
The following fails when built with `config=cuda` with

```
E tensorflow/stream_executor/cuda/cuda_driver.cc:1239] failed to enqueue async memcpy from device to host: CUDA_ERROR_INVALID_VALUE; host dst: 0x208a00000; GPU src: 0x1b1a4a0; size: 4=0x4
F tensorflow/core/common_runtime/gpu/gpu_util.cc:296] GPU->CPU Memcpy failed

```
Looking at device placement, it seems that `GetSessionTensor` ops are placed on GPU. Changing `failure_case` to `False` makes the test below pass as everything is placed on CPU as expected.

```
import tensorflow as tf                                                         
                                                                                
class HandleTest(tf.test.TestCase):                                             
                                                                                
  def testHandle(self):                                                         
    with self.test_session() as sess:                                           
      a = tf.constant(1.0)                                                      
      a_handle_op = tf.get_session_handle(a)                                    
      b = tf.constant(2.0)                                                      
      b_handle_op = tf.get_session_handle(b)                                    
                                                                                
      failure_case = True                                                       
      if failure_case:                                                          
        a_p, a_t = tf.get_session_tensor(tf.float32)                            
        b_p, b_t = tf.get_session_tensor(tf.float32)                            
        a_handle = sess.run(a_handle_op)                                        
        b_handle = sess.run(b_handle_op)                                        
      else:                                                                     
        a_handle = sess.run(a_handle_op)                                        
        b_handle = sess.run(b_handle_op)                                        
        a_p, a_t = tf.get_session_tensor(tf.float32)                            
        b_p, b_t = tf.get_session_tensor(tf.float32)                            
                                                                                
      c = tf.add(a_t, b_t)                                                      
      c_handle = sess.run(                                                      
        tf.get_session_handle(c),                                               
        feed_dict={a_p: a_handle.handle,                                        
                   b_p: b_handle.handle})                                       
      self.assertEqual(3.0, c_handle.eval())                                    
                                                                                
                                                                                
if __name__ == "__main__":                                                      
  tf.test.main()                                                                

```

yaroslavvb:
@yuanbyu

yaroslavvb:
[...]

PR #2595, opened 2016-05-31
Immediate Mode execution in tensorflow

yaroslavvb:
Implementation of immediate mode execution in TensorFlow. The idea is to wrap original tensorflow API, but provide session/run management logic so that commands execute immediately, and graph caching to avoid modifying graph when same op is run repeatedly. Data is kept in TensorFlow runtime whenever possible using persistent tensors and transferred to Python runtime on demand when needed for printing/control flow.

```
import tensorflow as tf
from tensorflow.contrib import immediate

tfi = immediate.Env(tf).tf        # wraps "tf" namespace, saves it as "tfi"
val1 = tfi.ones((3,))             # creates tensor on GPU
val2 = val1 + val1                # runs into tf.add, keeps result on GPU
val3 = tfi.ones((2, 2, 2, 2))
val4 = tfi.nn.conv2d(val3, val3, [1, 1, 1, 1], "VALID")   # run CuDNN conv2d
if (tfi.reduce_sum(val3)>0.):     # runs reduce_sum on GPU, transfers bool to CPU
  print(val4)                     # transfers whole tensor to CPU for display

```

This is a single commit rebase of a previous pull request from https://github.com/tensorflow/tensorflow/pull/2346

@keveman


martinwicke:
Jenkins, test this please.

yaroslavvb:
The failures are due to https://github.com/tensorflow/tensorflow/issues/2586, @yuanbyu  are you able to reproduce?

yaroslavvb:
Hm, I may need to move this to a new pull request, every time I do rebase, it includes all the changes fast-forwarded by the rebase into the pull request

yaroslavvb:
Sorry, I'm opening new request since I don't know how to fix this:

https://github.com/tensorflow/tensorflow/pull/2747

In my current work-flow, when I rebase, every commit that's included during fast forwarding goes into the pull request. Here's the sequence of commands I've been using so that people know not to do this: http://stackoverflow.com/questions/37707605/doing-rebase-adds-replays-master-branch-commits-into-my-pull-request

PR #2624, opened 2016-06-02
bug fix in ApplyAdadelta update rule

dpaiton:
Following the Zeiler ADADELTA paper, numerator of update equation
should be RMS(accum_update) \* grad

### DIFF
--- a/tensorflow/python/training/adadelta_test.py
+++ b/tensorflow/python/training/adadelta_test.py
@@ -20,104 +20,90 @@ from __future__ import print_function
 import tensorflow.python.platform
 
 import numpy as np
-from six.moves import xrange  # pylint: disable=redefined-builtin
 import tensorflow as tf
 
-
 class AdadeltaOptimizerTest(tf.test.TestCase):
-
   def testBasic(self):
+    num_updates = 4 # number of ADADELTA steps to perform
     for dtype in [tf.half, tf.float32]:
-      with self.test_session():
-        var0 = tf.Variable([1.0, 2.0], dtype=dtype)
-        var1 = tf.Variable([3.0, 4.0], dtype=dtype)
-        grads0 = tf.constant([0.1, 0.1], dtype=dtype)
-        grads1 = tf.constant([0.01, 0.01], dtype=dtype)
-        lr = 1.0
-        rho = 0.95
-        epsilon = 1e-8
+        for grad in [0.2, 0.1, 0.01]:
+            for lr in [1.0, 0.5, 0.1]:
+              with self.test_session():
+                var0_init = [1.0, 2.0]
+                var1_init = [3.0, 4.0]
+                var0 = tf.Variable(var0_init, dtype=dtype)
+                var1 = tf.Variable(var1_init, dtype=dtype)
 
-        adadelta_opt = tf.train.AdadeltaOptimizer(lr, rho=rho, epsilon=epsilon)
-        adadelta_update = adadelta_opt.apply_gradients(zip(
-            [grads0, grads1], [var0, var1]))
-        tf.initialize_all_variables().run()
+                grads = tf.constant([grad, grad], dtype=dtype)
 
-        # Check we have slots
-        self.assertEqual(["accum", "accum_update"],
-                         adadelta_opt.get_slot_names())
-        slot0 = adadelta_opt.get_slot(var0, "accum")
-        self.assertEquals(slot0.get_shape(), var0.get_shape())
-        self.assertFalse(slot0 in tf.trainable_variables())
+                accum = 0.0
+                accum_update = 0.0
 
-        slot0_update = adadelta_opt.get_slot(var0, "accum_update")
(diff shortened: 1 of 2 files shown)


dpaiton:
@vrv had requested that I also merge a fix to the test. I looked in [training_ops_test.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/training_ops_test.cc) and did not find any testing for adadelta. I also checked [adadelta_test.py](https://github.com/tensorflow/tensorflow/blob/16d395e5dea687ab3aece0a462e631de25c8d77d/tensorflow/python/training/adadelta_test.py), which has the correct rule in place. I do not know where else I should look for the bug.

vrv:
Correct me if I'm wrong, but you're saying that adadelta_test has the right equation, but that this test was passing before your change, and still passes after your change?

dpaiton:
Correct, the file [adadelta_test.py](https://github.com/tensorflow/tensorflow/blob/16d395e5dea687ab3aece0a462e631de25c8d77d/tensorflow/python/training/adadelta_test.py) passes before and after the fix.

vrv:
doesn't that mean that the existing test is not really sufficient?  Can we modify the test so it would have failed before your fix?

dpaiton:
Probably - I suspect that decreasing the tolerance in the asserts (from 1e-3 to a smaller number) would cause the test to fail, although I don't know if there is a testing convention in place for tolerance. Is there a reason 1e-3 was chosen in the first place, or is that a default value? These tests use numerical approximations, right? So at some point a low enough tolerance will always cause a failure.

vrv:
There's no really good default, since every test has a different number of operations being performed.  Ideally we would test based off of expected number of ULP differences for every test, but that's somewhat impractical.

If you can find a setting that passes with your change and fails before it, that would be awesome.

dpaiton:
I'll look into it and get back to you.

dpaiton:
Ok, I added additional test variable options (learning rates, gradients, number of ADADELTA steps) in order to find a use case where the previously broken code fails and the newly patch code passes.

tensorflow/python/training/adadelta_test.py
```diff
@@ -20,104 +20,90 @@
...
+                rho = 0.95
+                epsilon = 1e-8
+                adadelta_opt = tf.train.AdadeltaOptimizer(lr, rho, epsilon)
+                adadelta_update = adadelta_opt.apply_gradients(zip(
+                    [grads, grads], [var0, var1]))
+
+                tf.initialize_all_variables().run()
+
+                # Assign slots
+                slot = [None]*2
```
vrv: space between ] and *  and 2

tensorflow/python/training/adadelta_test.py
```diff
@@ -20,104 +20,90 @@
...
+                tot_update = 0
+                for run_no in range(num_updates):
+                    # Run adadelta update for comparison
+                    adadelta_update.run()
+
+                    # Perform initial update without previous accum values
+                    accum = accum * rho + (grad**2) * (1 - rho)
+                    update[run_no] = (np.sqrt(accum_update + epsilon) *
+                        (1. / np.sqrt(accum + epsilon)) * grad)
+                    accum_update = accum_update * rho + (update[run_no]**2) * (1.0 - rho)
```
vrv: our linters go nuts when lines go over 80 characters, can you try to wrap these to 80 cols as necessary?)

tensorflow/python/training/adadelta_test.py
```diff
@@ -20,104 +20,90 @@
...
-            var0.eval(),
-            rtol=1e-3)
-
-        self.assertAllCloseAccordingToType(
-            np.array([3.0 - update1 - update2, 4.0 - update1 - update2]),
-            var1.eval(),
-            rtol=1e-3)
-
+        for grad in [0.2, 0.1, 0.01]:
+            for lr in [1.0, 0.5, 0.1]:
```
vrv: for better or worse, we use 2 space indents

tensorflow/python/training/adadelta_test.py
```diff
@@ -20,104 +20,90 @@
...
+                self.assertEquals(slot_update[1].get_shape(), var1.get_shape())
+                self.assertFalse(slot_update[1] in tf.trainable_variables())
+
+                # Fetch params to validate initial values
+                self.assertAllClose(var0_init, var0.eval())
+                self.assertAllClose(var1_init, var1.eval())
+
+                update = [None]*num_updates
+                tot_update = 0
+                for run_no in range(num_updates):
```
vrv: "step" is a more common parlance in TF code

vrv:
Thanks for this!  mostly just lint / style issues, otherwise this looks great.

dpaiton:
Ok, I fixed the linting/style errors. Should I squash the commits or can you do that in the merge?

vrv:
We can squash now when we merge :)

martinwicke:
Thanks!

dpaiton:
Thanks for your help @vrv

Issue #2233, opened 2016-05-05
protobuf message overflow on trying distributed 


smartcat2010:
I'm trying to build an RNN on multi-machines following the [Distributed Tensorflow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/how_tos/distributed/index.md). 
when I use "with sv.managed_session(server.target) as sess:", it shows error:
AttributeError: 'Supervisor' object has no attribute 'managed_session'
So I follow the code of "Inception":
with sv.prepare_or_wait_for_session(server.target, config = sess_config) as sess :

Then it starts to run, but hangs immediately after reporting the following error:

[libprotobuf WARNING google/protobuf/src/google/protobuf/io/coded_stream.cc:569] Reading dangerously large protocol message.  If the message turns out to be larger than 67108864 bytes, parsing will be halted for security reasons.  To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.
[libprotobuf ERROR google/protobuf/src/google/protobuf/io/coded_stream.cc:207] A protocol message was rejected because it was too big (more than 67108864 bytes).  To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.
[libprotobuf WARNING google/protobuf/src/google/protobuf/io/coded_stream.cc:81] The total number of bytes read was 67108864
E tensorflow/core/framework/tensor.cc:105] Input size was 67108839 and expected 72000800

Would you please help me on this?
Thanks a lot in advance!

mrry:
Thanks for letting us know about this - it sounds like a bug in the tensor serialization code, but to figure out what's going wrong, we're going to need more information:

1. Can you pinpoint the line of Python on which it hangs? (Does it hang on the `sv.prepare_or_wait_for_session()` or does it hang when you try to run something?)
2. Is there an obviously-large tensor in your program? Does it exist as a `tf.constant()`, a numpy array that is implicitly converted to a `tf.constant()`, a value that is fed into the graph, or could it be a result that you're fetching? (Looking at the expected size, 72000800 bytes, I'm guessing that it's a tensor with size 90001 in one dimension, maybe a 200 x 90001 matrix of floats?) All of these cases should be handled, but this will help to construct a minimal failing test case.
[...]

PR #2632, opened 2016-06-03
Updated gRPC to release 0.14.

mrry:
This patch updates the version of gRPC used to the latest release. It adds dependencies on nanopb and the same boringssl library as used in many of the models' workspaces.


mrry:
@keveman: This update is a prerequisite to the fix for #2233. If the tests pass, alright to merge?

jendap:
@mrry can you please poke grpc folks to make the build more modular so we can remove the copied BUILD later?

martinwicke:
Yeah, maintaining build files for other people is asking for trouble.

PR #2657, opened 2016-06-04
Android permission check

ErnestG:
Added a simple permission check and request to prevent crashing on api 23+


ErnestG:
I signed it!

Issue #2631, opened 2016-06-03
error building py package in format of tar.gz


ericxsun:
OS: OS X Yosemite 10.10.4

After installation of dependencies, I build it as following cmds:

delete `bdist_wheel` in `tensorflow/tensorflow/tools/pip_package/build_pip_package.sh`

`${PYTHON_BIN_PATH:-python} setup.py bdist_wheel >/dev/null` to `${PYTHON_BIN_PATH:-python} setup.py`

```


bazel build -c opt //tensorflow/tools/pip_package:build_pip_package

sh +x tensorflow/tensorflow/tools/pip_package/build_pip_package.sh /tmp
```

then I got the package named `tensorflow-0.8.0.tar.gz`, after that I installed it using pip, succeed.

However, when I imported it, I got the error:

```
>>> import tensorflow as tf
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named tensorflow
```

Could anybody help me? Any help will be highly appreciated.

Could staff in Google build it or give some valuable instructions?

aselle:
I think this might be related to the problems affecting "setting up for development" #2497 i.e. in 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md
section "Setting up TensorFlow for Development."

Specifically, The instructions have to be different on what you symlink into the python build directory dependent on the version of bazel (which ones to do when are outlined in #2497).  Are you able to get those instructions to work and sucessfully import tensorflow from python?

ericxsun:
Yes, I have made it worked. I'll update the doc later. 

I have another question:  where can I find the code before generating by `SWIG` of ` bazel-bin/tensorflow/python/pywrap_tensorflow.py`. I want to make it more general when packaging it as tar.gz.

Thanks a lot. Awaiting for your reply.

jendap:
For the generated file location see bazel-out...

For example [pywrap_tensorflow.cc](http://ci.tensorflow.org/job/tensorflow-master-cpu/ws/bazel-out/local_linux-opt/bin/tensorflow/python//pywrap_tensorflow.cc) on [ci.tensorflow.org](http://ci.tensorflow.org/job/tensorflow-master-cpu/ws/bazel-out/local_linux-opt/bin/tensorflow/python/).

aselle:
Closing since it seems to be of no further problem. Reopen if further bugs are discovered.

PR #2666, opened 2016-06-05
add doc for packaging TF in format of tar.gz

airlsyn:
tackle issue: #2631


ericxsun:
I signed it!

jendap:
This file is already big. It does not feel right to put there all those instructions...

Why to build tar.gz anyway?

PR #2673, opened 2016-06-05
temporarily disable contrib/session_bundle

jendap:
The genrule makes problems in integration testing. We disable it for now and fix it after 0.9 release.


jendap:
@nfiedel - we are disabling this for now. It have problems to build in [ci.tensorflow.org](http://ci.tensorflow.org). It is not needed for 0.9, is it?

FYI @caisq @martinwicke @gunan

jendap:
The problem is with the shell script inside genrule in session_bundle/example. It picks wrong version of swig or python or something like that. See [log from a failed build](http://ci.tensorflow.org/view/Release/job/release-matrix-cpu/TF_BUILD_CONTAINER_TYPE=CPU,TF_BUILD_IS_OPT=OPT,TF_BUILD_IS_PIP=PIP,TF_BUILD_PYTHON_VERSION=PYTHON3,label=cpu-slave/45/console).

Also, such a shell script in genrule may cause problems for builds on windows (no "cp") or when two builds would run on the same machine (hardcoded /tmp path).

PR #2675, opened 2016-06-05
temporarily disable contrib/session_bundle from test_installation.sh

jendap:


jendap:
@caisq does this look right? or do wait for the builds?

jendap:
closing as it is also added in #2676

Issue #1968, opened 2016-04-15
Common shape ops for SparseTensor


concretevitamin:
It should be easy to extend/modify `tf.rank()`, `tf.shape()`, and `tf.size()` to have them work on `SparseTensor`s.

Happy to review PRs!

AnishShah:
Hi! 
I'm working on this.

AnishShah:
@concretevitamin I'm having some problem with bazel-build. Not able to build from source. Is there any other way I can build from source?

concretevitamin:
What are the errors?  I trust that the latest 0.8 docs on the website did
not help?

siddharth-agrawal:
@concretevitamin Could you provide some hints as to how this might be done? I spent some time looking at some sparse operations to understand how `SparseTensor`s are handled. In most of the cases, `SparseTensor`s are converted to `Tensor`s for the underlying Op. One exception I found is `sparse_matmul`. Are `a_is_sparse`, `b_is_sparse` somehow relevant?

concretevitamin:
@siddharth-agrawal Essentially, SparseTensor in Python can be thought of as a 3-tuple of dense tensors, `(indices, values, shape)`.  You're right that most underlying Ops take these three dense Tensors separately, but there's no "conversion" going on. 

As for `sparse_matmul()`: due to historical reasons it's actually a misnomer.  It does not operate on `SparseTensor`s, but instead use an optimized algorithm depending on the two flags.  So that op is irrelevant.

siddharth-agrawal:
@concretevitamin If we take the special case of `tf.rank()`, then we need some way making [this statement](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/shape_ops.cc#L132) generic so that it works for a `SparseTensor` as well. Are there any utilities that can differentiate between a `Tensor` and a `SparseTensor`? If not, can you tell me how to approach this (since you said that i should be easy in the description)? I did a lot of digging, but I still have no idea how to solve this. Even a 'where to look' reply would be fine.

concretevitamin:
[...]

PR #2681, opened 2016-06-06
Enable tf.size() for SparseTensor

siddharth-agrawal:
Added an override for `tf.size()`, that takes care of `SparseTensor` objects as well. Added tests and verified locally. This partially addresses #1968.

### DIFF
--- a/tensorflow/python/kernel_tests/shape_ops_test.py
+++ b/tensorflow/python/kernel_tests/shape_ops_test.py
@@ -91,6 +91,15 @@ class ShapeOpsTest(tf.test.TestCase):
     self.assertAllEqual(np_ans, result)
     self.assertShapeEqual(np_ans, tf_ans)
 
+  def _compareSizeSparse(self, x_np, use_gpu=False):
+    np_ans = np.asarray(np.size(x_np))
+    x_tf, unused_nnz = _sparsify(x_np)
+    with self.test_session(use_gpu=use_gpu):
+      tf_ans = tf.size(x_tf)
+      result = tf_ans.eval()
+    self.assertAllEqual(np_ans, result)
+    self.assertShapeEqual(np_ans, tf_ans)
+
   def _testCpu(self, x):
     self._compareShape(x, use_gpu=False)
     self._compareShapeN(x, use_gpu=False)
@@ -98,6 +107,7 @@ class ShapeOpsTest(tf.test.TestCase):
     self._compareSize(x, use_gpu=False)
     self._compareShapeSparse(x, use_gpu=False)
     self._compareRankSparse(x, use_gpu=False)
+    self._compareSizeSparse(x, use_gpu=False)
 
   def _testGpu(self, x):
     self._compareShape(x, use_gpu=True)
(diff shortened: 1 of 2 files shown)


ebrevdo:
Jenkins, test this please.

tensorflow/python/ops/array_ops.py
```diff
@@ -125,6 +125,32 @@ def shape(input, name=None):
...
+  Args:
+    input: A `Tensor` or `SparseTensor`.
+    name: A name for the operation (optional).
+
+  Returns:
+    A `Tensor` of type `int32`.
+  """
+  with ops.op_scope([input], name, "Size") as name:
+    if isinstance(input, ops.SparseTensor):
+      return gen_math_ops._prod(input.shape, 0, name=name)
```
ebrevdo: why not math_ops.reduce_prod ?

siddharth-agrawal: This may be wrong, but I didn't do it because it may lead to cyclic dependencies. `math_ops.py` imports `array_ops`, so importing `math_ops` here will lead to a cycle. I'm not sure how dependencies work in Python, so this may be entirely wrong. Also, `tf.reduce_prod()` uses `gen_math_ops._prod()` within the implementation, so I guess it doesn't make a difference.

siddharth-agrawal:
@ebrevdo Is there something I need to do here? I mean apart from the rebase.

tensorflow/python/ops/array_ops.py
```diff
@@ -125,6 +125,32 @@ def shape(input, name=None):
...
+def size(input, name=None):
+  """Returns the size of a tensor.
+
+  This operation returns an integer representing the number of elements in
+  `input`.
+
+  For example:
+
+  ```python
+  # 't' is [[[1, 1,, 1], [2, 2, 2]], [[3, 3, 3], [4, 4, 4]]]]
```
concretevitamin: I see that the double-comma typo has managed to live its life for a while in our official doc...Mind fixing it while you're at it?

tensorflow/python/ops/array_ops.py
concretevitamin: Agreed with @yaroslavvb.

One minor thing: `input.shape` for `SparseTensor` is actually int64, so to avoid inconsistency we should either (1) change the "Returns:" part in our documentation, or (2) cast to int32 here.  

I am leaning towards (2) to keep it simple.  @ebrevdo any thoughts?

concretevitamin:
@siddharth-agrawal I took a quick look and left a few minor comments.  Let's wait for Eugene's thought on the returned dtype.

siddharth-agrawal:
@ebrevdo Could you comment on @concretevitamin's comment on dtype?

tensorflow/python/ops/array_ops.py
ebrevdo: I'm also leading towards (2).  It is unfortunate that size returns an int32, but not much we can do about it now.

concretevitamin:
Jenkins, test this please.

concretevitamin:
Merged into master. Thanks.

tensorflow/python/kernel_tests/shape_ops_test.py
```diff
@@ -106,6 +116,7 @@ def _testGpu(self, x):
     self._compareSize(x, use_gpu=True)
     self._compareShapeSparse(x, use_gpu=True)
     self._compareRankSparse(x, use_gpu=True)
+    self._compareSizeSparse(x, use_gpu=True)
```
OneDay18: False

PR #2686, opened 2016-06-06
Fix formatting issue in docs caused by lack of backtick

Undo1:
Missing backtick was causing unhappiness.


Undo1:
@googlebot  I signed it!

girving:
@Undo1: Thank you for the fix!  Backticks are depressing hard to see. :)

PR #2691, opened 2016-06-06
Gated Feedback LSTM implementation

bogdanteleaga:
This PR aims to add a GFLSTM implementation in contrib.

Some of the reasoning for adding a new MultiCell is explained. The new cells cannot be used without a MultiCell, since rnn doesn't know how to pass the state properly.

I am not sure whether splitting all the h's before is better or not, but I think it should use less memory overall on a high number of layers.

The state_is_tuple implementation has not been tested but follows what's being done in the current MultiRNN and LSTM cell so they should be fine.

I'm also open to suggestions for the TODO item, I'm not sure how to handle that.

### DIFF
--- a/tensorflow/contrib/rnn/python/ops/rnn_cell.py
+++ b/tensorflow/contrib/rnn/python/ops/rnn_cell.py
@@ -552,7 +552,11 @@ class MultiGFRNNCell(rnn_cell.MultiRNNCell):
         with vs.variable_scope("Cell%d" % i):
           cur_inp, new_state = cell(cur_inp, (prev_states[i], prev_hs))
           new_states.append(new_state)
-      return cur_inp, array_ops.concat(1, new_states)
+
+      if self._state_is_tuple:
+          return cur_inp, tuple(new_states)
+      else:
+          return cur_inp, array_ops.concat(1, new_states)
 
 
 class GFLSTMCell(rnn_cell.BasicLSTMCell):
--- a/tensorflow/contrib/rnn/python/ops/rnn_cell.py
+++ b/tensorflow/contrib/rnn/python/ops/rnn_cell.py
@@ -29,6 +29,9 @@ from tensorflow.python.ops import rnn_cell
 from tensorflow.python.ops import variable_scope as vs
 from tensorflow.python.ops.math_ops import sigmoid
 from tensorflow.python.ops.math_ops import tanh
+from tensorflow.python.ops.rnn_cell import LSTMStateTuple
+from tensorflow.python.ops.rnn_cell import _is_sequence
+from tensorflow.python.ops.rnn_cell import _linear as linear
 
 
 def _get_concat_variable(name, shape, dtype, num_shards):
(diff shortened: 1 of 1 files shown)


tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -478,3 +481,137 @@ def _make_tf_features(self, input_feat):
...
+          if self._state_is_tuple:
+            if not _is_sequence(state):
+              raise ValueError(
+                  "Expected state to be a tuple of length %d, but received: %s"
+                  % (len(self.state_size), state))
+            prev_state = state[i]
+          else:
+            prev_state = array_ops.slice(state, [0, curr_state_pos], [-1, cell.state_size])
+            curr_state_pos += cell.state_size
+          if cell._state_is_tuple:
```
bogdanteleaga: This part will only work properly for some cell types, namely LSTM and I think GRU as well. To make it more general we could unpack inside the cell itself, but that means more unpacks.

It also breaks when wrappers are used

lukaszkaiser:
The PR looks ok to me, but it will need tests. Please add tests to rnn_cell_test and make sure to test both with and without state_is_tuple.

bogdanteleaga:
Is there a guide on setting up an environment where I can test? The contributing page doesn't seem to say anything.

Also any suggestions for the TODO item and the slicing part?

lukaszkaiser:
I think you can just run rnn_cell_test to test locally, and then github/jenkins will test all other test for you, right? As for TODO: do you mean that MultiGFRNNCell needs to work with other cells than GFXXX cells? Maybe not, we have MultiRNNCell for that in the end. Or did I misunderstand you?

bogdanteleaga:
So for the TODO bit: I'm using this piece of code to test whether the cell that's passed in is a GFLSTMCell `if not isinstance(cell, GFLSTMCell)`. However, that GFLSTMCell can be wrapper by a dropout wrapper for example and then I'd need to test `cell._cell` and so on. I was wondering if there's an easier way to do this. I don't know beforehand how many wrappers are around the cell. I guess I could just do `if cell._cell and isinstance(cell._cell)` and recurse on that but it feels yucky.

I refactored the MultiCell a bit to allow it to potentially work with other GF cells. It now passes the whole previous state along with an index to identify the input previous state for the particular cell.

lukaszkaiser:
Would it be possible to extend MultiRNNCell by an optional argument rather than making a separate MultiGFXXX Cell? Maybe that'll be cleaner?

ebrevdo:
Initial comment:

Don't import _linear.  Use contrib.layers.linear with the appropriate initializer instead.

Lulasz: I prefer not making MultiRNNCell more complicated, so this is a good place for this code for now.

ebrevdo:
Oh, and get rid of the state_is_tuple. That's for backwards compatibility. Assume it's true.

rmlarsen:
Ping. Any progress on this?

bogdanteleaga:
Not really. I was busy finishing my thesis, which I just presented today :)

Hopefully, I can finish it by the end of this or the next week, with the main issue being the tests themselves. I started using `state_is_tuple` in my project so that bit works. Also, still unsure of how to deal with accepting only the proper cell types.

vrv:
(another friendly ping!)

danmane:
Since it's been about a month since the last ping with no activity, I'm going to close this pull request. @bogdanteleaga please feel free to re-open it when you have the time!

PR #2705, opened 2016-06-07
Update Cmake build files to reflect recent dependency change

lilac:
Add external library dependencies including
- jsoncpp
- boringssl
- farmhash
- highwayhash

Also create targets to generate the `proto_text` binary, and add a custom rule for generating proto text parser code using the previous output tool.


lilac:
I signed it!

mrry:

mrry:
Thanks for doing this! I've been trying to resuscitate the CMake files, and this saves me a lot of effort :).

mrry:
@jendap @caisq: Does either of you know why the "Linux CPU Tests CMAKE" test might be failing with the following error?

```
[  0%] Running C++ protocol buffer compiler on tensorflow/core/framework/log_memory.proto
make[2]: execvp: /usr/local/bin/protoc: Permission denied
make[2]: *** [tensorflow/core/framework/log_memory.pb.cc] Error 127
make[1]: *** [CMakeFiles/tf_protos_cc.dir/all] Error 2
make: *** [all] Error 2
```

It looks like the `install_proto3.sh` script is running. Is there something else we need to do to chmod `protoc`?

tensorflow/contrib/cmake/tf_core_framework.cmake
```diff
@@ -136,7 +204,7 @@ file(GLOB_RECURSE tf_core_framework_test_srcs
 
 list(REMOVE_ITEM tf_core_framework_srcs ${tf_core_framework_test_srcs})
 
-add_library(tf_core_framework OBJECT ${tf_core_framework_srcs})
+add_library(tf_core_framework OBJECT ${tf_core_framework_srcs} ${PROTO_TEXT_HDRS})
```
mrry: I'm not sure of the best way to do this, but I think you need to add a dependency on `${PROTO_TEXT_SRCS}` here.

mrry: I think the layering has (unfortunately) changed so that we need a few more source files in here. A bit of experimentation reveals that the following might be necessary:

```
    "${tensorflow_source_dir}/tensorflow/core/client/tensor_c_api.cc"
    "${tensorflow_source_dir}/tensorflow/core/common_runtime/session.cc"
    "${tensorflow_source_dir}/tensorflow/core/common_runtime/session_factory.cc"
    "${tensorflow_source_dir}/tensorflow/core/common_runtime/session_options.cc"
```

...and we'll need to exclude them from `tf_core_cpu` as well.

tensorflow/contrib/cmake/tf_core_kernels.cmake
```diff
@@ -19,12 +19,14 @@ list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_exclude_srcs})
 
 add_library(tf_core_kernels OBJECT ${tf_core_kernels_srcs})
 
-add_dependencies(tf_core_kernels tf_core_cpu)
+add_dependencies(tf_core_kernels tf_core_cpu farmhash highwayhash)
```
mrry: We also need dependencies on `farmhash_copy_headers_to_destination` and  `highwayhash_copy_headers_to_destination` here.

mrry:
@lilac: I've forked this PR (#2839) and added the necessary fixes to make it build the tutorial example.

mrry:
I'm going to merge this (even though it is incomplete), and follow up with #2839, which makes the build work again.

Issue #813, opened 2016-01-19
Cumulative sum (cumsum) op needed


langmore:
C.f. [`numpy.cumsum`](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.cumsum.html)

benoitsteiner:
You can achieve what you're looking for by padding your input with zeros and then convolving the padded tensor with ones. For example, if you have a vector of N elements, you can start by padding it with N-1 elements on the left, and the launch a 1D convolution with a kernel of size N.

Of course that won't be very efficient, but that can get you going for the time being.

ibab:
I'd like to learn how to write ops by implementing this.
I'll work based on the [adding_an_op](https://www.tensorflow.org/versions/0.6.0/how_tos/adding_an_op/index.html) section in the docs.

Can someone give me a few pointers on where this should end up in the tensorflow source tree?
The op itself would probably be registered in [`core/ops/math_ops.cc`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc).
Should I put the kernel into  a new `core/kernels/cumsum_op.cc` file?

girving:
If we do this, it should probably have a name pattern similar to `reduce_sum`, `reduce_prod`, etc.  Maybe `accumulate_sum`, `accumulate_prod` ([somewhat following numpy](https://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.ufunc.accumulate.html))?  @dave-andersen: Do you have opinions here?

Also, I'm not sure if it should be inclusive or exclusive.

Once we figure that out: yes, the op definitions would go in `core/ops/math_ops.cc`, and the kernel files would have a similar pattern to the reduction ops in `core/kernels`.

leconteur:
I did the following in a personal project. It is probably inefficient, but probably less than a convolution.

`def cumsum(softmax):
    values = tf.split(1, softmax.get_shape()[1], softmax)
    out = []
    prev = tf.zeros_like(values[0])
    for val in values:
        s = prev + val
        out.append(s)
        prev = s
    cumsum = tf.concat(1, out)
    return cumsum`

leconteur:
I did the following in a personal project. It is not very efficient but it's probably a little better than the convolution.

https://gist.github.com/leconteur/82f968f483ddd2234a03

langmore:
Hello ibab.  Have you been able to make any progress on this?  Thanks!

[...]

Issue #2641, opened 2016-06-03
Error in gradient of reduce_prod


RafaelCosman:
```
vars = tf.Variable([1., 2.])
tf.initialize_all_variables().run()
tf.gradients(tf.reduce_prod(vars), vars)[0].eval()
```

yields `[ 2.,  1.]` which is correct. But

```
vars = tf.Variable([0., 2.])
tf.initialize_all_variables().run()
tf.gradients(tf.reduce_prod(vars), vars)[0].eval()
```

yields `[ nan,   0.]` which is incorrect. The correct gradient is `[ 2.,  0.]`

RafaelCosman:
Note that this is not an issue with regular TensorFlow multiplication. The following works correctly:

```
var1 = tf.Variable(0.)
var2 = tf.Variable(2.)
tf.initialize_all_variables().run()
print tf.gradients(var1 * var2, var1)[0].eval()
print tf.gradients(var1 * var2, var2)[0].eval()
```

It prints `2.0` and `0.0` which is correct.

aselle:
I have reproduced this.

RafaelCosman:
Thanks @aselle.

Idea: This could be caused `reduce_prod` is implemented somewhat like `tf.exp(tf.reduce_sum(tf.log(.)))`

girving:
The gradients are computed by computing the full prod and then dividing, which is broken as you point out.  For example, for a length two produce `x *  y` the gradient w.r.t. `x` is `x * y / y`.  Indeed, there's an ancient TODO in the code that it is broken.

@ibab, @benoitsteiner: The easiest way to fix this would be to do two scan products to get the sequence of partial products from both directions, then multiply them together to get all products with one element removed.  Unfortunately, we don't yet have scan.

girving:
Adding contributions welcome, but note that it'll have to wait until after #2711.

ibab:
I've had a shot at solving this today using `cumprod`, and it's working great for full reductions.
But if `reduce_prod` is performed with `reduction_indices`, things get more difficult.
In that case, the `cumprod` needs to be performed over the reduced dimensions, and not over the remaining ones.
Any ideas on how this could be solved?
Maybe if there was a `tf.reshape_selected` op that allows you to group all selected dimensions into a single dimension, and all remaining ones into a second one, like this:
```
x.shape = (a, b, c, d, e)
reduction_indices = [1, 3]
y = tf.reshape_selected(x, reduction_indices)
[...]

PR #2711, opened 2016-06-07
Add cumsum and cumprod ops

ibab:
This implements `tf.scan_sum` (cumulative sum) and `tf.scan_prod` (cumulative product) ops.
Unfortunately, this won't compile with the current version of eigen, as I have to make some small changes to `TensorScanOp.h`
But this should be useful to discuss how the API should look and whether we want to have inclusive/exclusive ops, and so on.

I'll also add tests and gradients to the two ops.

This relates to #813.

### DIFF
--- a/tensorflow/core/kernels/scan_ops.cc
+++ b/tensorflow/core/kernels/scan_ops.cc
@@ -1,0 +1,175 @@
+/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#define EIGEN_USE_THREADS
+#if GOOGLE_CUDA
+#define EIGEN_USE_GPU
+#endif  // GOOGLE_CUDA
+
+#include "tensorflow/core/framework/numeric_op.h"
+#include "tensorflow/core/framework/op_kernel.h"
+#include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/framework/tensor.h"
+#include "tensorflow/core/framework/types.h"
+#include "tensorflow/core/kernels/bounds_check.h"
+
+#include "third_party/eigen3/Eigen/Core"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
+#include "tensorflow/core/kernels/scan_ops.h"
+
+namespace tensorflow {
+
+typedef Eigen::ThreadPoolDevice CPUDevice;
+typedef Eigen::GpuDevice GPUDevice;
+
+template <typename Device, class T, typename Reducer>
+class ScanOp : public OpKernel {
+public:
+  explicit ScanOp(OpKernelConstruction* ctx) : OpKernel(ctx) {
+    OP_REQUIRES_OK(ctx, ctx->GetAttr("reverse", &reverse_));
+  }
+
+  void Compute(OpKernelContext* ctx) override {
+    const Tensor& input = ctx->input(0);
+    const Tensor& tensor_axis = ctx->input(1);
+
+    OP_REQUIRES(ctx, TensorShapeUtils::IsScalar(tensor_axis.shape()),
+                errors::InvalidArgument("ScanOp: axis must be a scalar, not ",
+                                        tensor_axis.shape().DebugString()));
+
+    const int axis = internal::SubtleMustCopy(tensor_axis.scalar<int>()());
+
+    OP_REQUIRES(
+        ctx, FastBoundsCheck(axis, input.dims()),
+        errors::InvalidArgument("ScanOp: Expected scan axis in the range [", 0,
+                                ", ", input.dims(), "), but got ", axis));
+
+    TensorShape output_shape = input.shape();
+    Tensor* output = nullptr;
+    OP_REQUIRES_OK(ctx, ctx->allocate_output(0, output_shape, &output));
+
+    const Device& d = ctx->eigen_device<Device>();
+    Reducer reducer;
+
+#define HANDLE_SCAN(NDIMS)                                                  \
+  case NDIMS:                                                               \
+    functor::Scan<Device, Reducer, T, NDIMS>()(d, input.tensor<T, NDIMS>(), \
+                                               output->tensor<T, NDIMS>(),  \
+                                               reducer, axis, reverse_);    \
+    return;
+
+    switch (input.dims()) {
+      // input.dims() == 0 can't occur as there
+      // is no valid axis parameter in this case
+      HANDLE_SCAN(1);
+      HANDLE_SCAN(2);
+      HANDLE_SCAN(3);
+      HANDLE_SCAN(4);
+      HANDLE_SCAN(5);
+      HANDLE_SCAN(6);
+      HANDLE_SCAN(7);
+      HANDLE_SCAN(8);
+      default:
+        OP_REQUIRES(ctx, false, errors::InvalidArgument(
+                                    "Scan does not support tensors with "
+                                    "more than 8 dimensions",
+                                    input.dims()));
+    }
+#undef HANDLE_SCAN
+  }
+
+private:
+  bool reverse_;
+};
+
+#ifdef GOOGLE_CUDA
+namespace functor {
+
+// Forward declarations of GPU functors
+#define DECLARE(REDUCER, T, D)                          \
+  template <>                                           \
+  void Scan<GPUDevice, REDUCER, T, D>::operator()(      \
+      const GPUDevice& d, TTypes<T, D>::ConstTensor in, \
+      TTypes<T, D>::Tensor out, const REDUCER& reducer, \
+      const Eigen::Index& axis, bool reverse);          \
+  extern template struct Scan<GPUDevice, REDUCER, T, D>;
+
+#define DECLARE_FOR_ALL_DIMS(REDUCER, T) \
+  DECLARE(REDUCER, T, 1);                \
+  DECLARE(REDUCER, T, 2);                \
+  DECLARE(REDUCER, T, 3);                \
+  DECLARE(REDUCER, T, 4);                \
+  DECLARE(REDUCER, T, 5);                \
+  DECLARE(REDUCER, T, 6);                \
+  DECLARE(REDUCER, T, 7);                \
+  DECLARE(REDUCER, T, 8);
+
+#define DECLARE_FOR_ALL_REDUCERS(T)                        \
+  DECLARE_FOR_ALL_DIMS(Eigen::internal::SumReducer<T>, T); \
+  DECLARE_FOR_ALL_DIMS(Eigen::internal::ProdReducer<T>, T);
+
+TF_CALL_GPU_NUMBER_TYPES(DECLARE_FOR_ALL_REDUCERS);
+
+#undef DECLARE_FOR_ALL_REDUCERS
+#undef DECLARE_FOR_ALL_DIMS
+#undef DECLARE
+
+}  // namespace functor
+#endif  // GOOGLE_CUDA
+
+
+// Register Cumsum kernels
+#define REGISTER_CPU_KERNELS(type)                                 \
+  REGISTER_KERNEL_BUILDER(                                         \
+      Name("Cumsum").Device(DEVICE_CPU).TypeConstraint<type>("T"), \
+      ScanOp<CPUDevice, type, Eigen::internal::SumReducer<type>>)
+TF_CALL_NUMBER_TYPES(REGISTER_CPU_KERNELS);
+#undef REGISTER_CPU_KERNELS
+
+#if GOOGLE_CUDA
+#define REGISTER_GPU_KERNELS(type)   \
+  REGISTER_KERNEL_BUILDER(           \
+      Name("Cumsum")                 \
+          .Device(DEVICE_GPU)        \
+          .TypeConstraint<type>("T") \
+          .HostMemory("axis"),       \
+      ScanOp<GPUDevice, type, Eigen::internal::SumReducer<type>>)
+TF_CALL_GPU_NUMBER_TYPES(REGISTER_GPU_KERNELS)
+#undef REGISTER_GPU_KERNELS
+#endif // GOOGLE_CUDA
+
+
+// Register Cumprod kernels
+#define REGISTER_CPU_KERNELS(type)                                  \
+  REGISTER_KERNEL_BUILDER(                                          \
+      Name("Cumprod").Device(DEVICE_CPU).TypeConstraint<type>("T"), \
+      ScanOp<CPUDevice, type, Eigen::internal::ProdReducer<type>>)
+TF_CALL_NUMBER_TYPES(REGISTER_CPU_KERNELS);
+#undef REGISTER_CPU_KERNELS
+
+#if GOOGLE_CUDA
+#define REGISTER_GPU_KERNELS(type)   \
+  REGISTER_KERNEL_BUILDER(           \
+      Name("Cumprod")                \
+          .Device(DEVICE_GPU)        \
+          .TypeConstraint<type>("T") \
+          .HostMemory("axis"),       \
+      ScanOp<GPUDevice, type, Eigen::internal::ProdReducer<type>>)
+TF_CALL_GPU_NUMBER_TYPES(REGISTER_GPU_KERNELS)
+#undef REGISTER_GPU_KERNELS
+#endif // GOOGLE_CUDA
+
+}  // namespace tensorflow
--- a/tensorflow/core/ops/math_ops.cc
+++ b/tensorflow/core/ops/math_ops.cc
@@ -1812,4 +1812,56 @@ b: Another tensor, of same type and shape as `a`.
 product: Pairwise cross product of the vectors in `a` and `b`.
 )doc");
 
+// --------------------------------------------------------------------------
+
+REGISTER_OP("Cumsum")
+    .Input("x: T")
+    .Input("axis: int32")
+    .Attr("reverse: bool")
+    .Output("out: T")
+    .Attr("T: numbertype")
+    .Doc(R"doc(
+Compute the cumulative sum of the tensor `x` along `axis`.
+
+The output `out` at any given index i is equal to the sum of all
+elements `x_j` of `x` with j <= i.
+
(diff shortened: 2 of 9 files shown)


tensorflow/core/ops/math_ops.cc
```diff
@@ -1447,4 +1447,52 @@ b: Another tensor, of same type and shape as `a`.
 product: Pairwise cross product of the vectors in `a` and `b`.
 )doc");
 
+// --------------------------------------------------------------------------
+
+REGISTER_OP("ScanSum")
```
vrv: Would cumsum/cumprod be more equivalent to numpy / matlab naming?

ibab: I decided to go with `scan_sum` and `scan_prod` based on this comment: https://github.com/tensorflow/tensorflow/issues/813#issuecomment-179365203
The idea was to keep it consistent with `reduce_*`.
But if the consensus is to go with `cumsum`/`cumprod` I'd be happy to change it.

ibab:
Here is the corresponding PR in Eigen, which needs to propagate into TensorFlow first before this PR can compile: https://bitbucket.org/eigen/eigen/pull-requests/195/add-small-fixes-to-tensorscanop

tensorflow/core/kernels/scan_ops_common.h
```diff
@@ -0,0 +1,106 @@
...
+      HANDLE_SCAN(2);
+      HANDLE_SCAN(3);
+      HANDLE_SCAN(4);
+      HANDLE_SCAN(5);
+      HANDLE_SCAN(6);
+      HANDLE_SCAN(7);
+      HANDLE_SCAN(8);
+      default:
+        OP_REQUIRES(ctx, false,
+                    errors::InvalidArgument("Scan only supports tensors with "
```
Mistobaan: what about: `Scan does not support tensors with more than 8 dimensions`

tensorflow/core/ops/math_ops.cc
```diff
@@ -1447,4 +1447,52 @@ b: Another tensor, of same type and shape as `a`.
...
+    .Input("x: T")
+    .Attr("axis: int32")
+    .Output("out: T")
+    .Attr("T: numbertype")
+    .Doc(R"doc(
+Compute the cumulative sum of the input tensor `x` along the axis `axis`.
```
Mistobaan: I agree with @vrv `cumsum` is easier to remember and understand what it does. Your description also confirms that. To describe the operation you don't use 'scan' anywhere.

ibab:
Seeing as `tf.reverse(tf.scan_X(tf.reverse(x)))` is useful in quite a few situations, would it make sense to add a `reverse=True/False` kwarg to the scan ops?
This could either be done in the Python wrappers, or in C++.

Also, I haven't yet figured out how to write the gradient for `scan_prod` without using the same division trick that's also been used in #2641.

tensorflow/core/kernels/scan_ops_common.h
ibab: Thanks! I've changed the message accordingly.

tensorflow/core/ops/math_ops.cc
ibab: I've now changed the op names to `cumsum` and `cumprod` 👍

ibab:
I've made a few changes:
 - Renamed the ops to `cumsum` and `cumprod`
 - Added GPU support
 - Added a `reverse` kwarg to efficiently compute backwards scans
 - Added gradients (the `cumprod` gradient still uses the `/ x` hack)
 - Added tests

My Eigen PR has also been merged, so we will be able to test the PR as soon as Eigen is updated to the newest version.

girving:
Sorry for the delay.  Looking over this now.

tensorflow/core/kernels/scan_ops.cc
```diff
@@ -0,0 +1,175 @@
...
+
+  void Compute(OpKernelContext* ctx) override {
+    const Tensor& input = ctx->input(0);
+    const Tensor& tensor_axis = ctx->input(1);
+
+    OP_REQUIRES(ctx, TensorShapeUtils::IsScalar(tensor_axis.shape()),
+                errors::InvalidArgument("ScanOp: axis must be a scalar, not ",
+                                        tensor_axis.shape().DebugString()));
+
+    int axis = tensor_axis.tensor<int, 0>()();
```
girving: Do the following in order to avoid irritating race condition technicalities:

    const int axis = internal::SubtleMustCopy(tensor_axis.scalar<int>()());

We try to avoid any possibility of crashing TensorFlow in cases of asynchronous race conditions.

tensorflow/core/ops/math_ops.cc
```diff
@@ -1447,4 +1447,56 @@ b: Another tensor, of same type and shape as `a`.
...
+    .Attr("T: numbertype")
+    .Doc(R"doc(
+Compute the cumulative sum of the tensor `x` along `axis`.
+
+The output `out` at any given index i is equal to the sum of all
+elements `x_j` of `x` with j <= i.
+
+By setting the `reverse` operation to `True`, the sum is performed in the
+reverse order. In contrast to using `tf.reverse`, this avoids copying the
+tensor.
```
girving: Despite numpy convention, typically when I do a scan I want the exclusive version with `output[0] = identity`.  In particular, that's the kind of scan necessary to do safe gradients of `reduce_prod`.

How difficult would a possibly optional exclusive scan be to arrange?

tensorflow/core/ops/math_ops.cc
```diff
@@ -1447,4 +1447,56 @@ b: Another tensor, of same type and shape as `a`.
...
+
+For example:
+
+```prettyprint
+# tensor 'x' is [1, 2, 3, 4, 5]
+tf.cumsum(x)               ==> [1, 3, 6, 10, 15]
+tf.cumsum(x, reverse=True) ==> [5, 9, 12, 14, 15]
```
girving: Is that right?  I would have assumed `[15, 14, 12, 9, 5]`?

girving: We should have doc lines for the inputs and outputs: `x`, `axis`, `reverse`, and `out`.

tensorflow/python/kernel_tests/scan_ops_test.py
```diff
@@ -0,0 +1,219 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+
+from tensorflow.python.framework import errors
+from tensorflow.python.framework import tensor_shape
+from tensorflow.python.ops import math_ops
```
girving: Use `tf.` rather than individual imports of `errors`, `tensor_shape`, and `math_ops`.

tensorflow/python/kernel_tests/scan_ops_test.py
```diff
@@ -0,0 +1,219 @@
...
+      for n in range(8):
+        self._compareAll(x, n)
+
+  def testInvalidAxis(self):
+    x = np.arange(0, 10).reshape([2, 5]).astype(np.float32)
+    input_tensor = tf.convert_to_tensor(x)
+    with self.assertRaisesWithPredicateMatch(
+        errors.InvalidArgumentError,
+        lambda e: "Expected scan axis in the range" in str(e)):
+      with self.test_session():
```
girving: Flip the order of the two withs.

tensorflow/python/ops/math_grad.py
```diff
@@ -815,3 +815,22 @@ def _CrossGrad(op, grad):
   u = op.inputs[0]
   v = op.inputs[1]
   return (math_ops.cross(v, grad), math_ops.cross(grad, u))
+
+
+@ops.RegisterGradient("Cumsum")
+def _CumsumGrad(op, grad):
+  axis = op.inputs[1]
+  reverse = op.get_attr("reverse")
+  return [math_ops.cumsum(grad, axis=axis, reverse=(not reverse)), None]
```
girving: That is a lovely gradient function. :)

tensorflow/python/ops/math_ops.py
```diff
@@ -1427,6 +1435,74 @@ def tanh(x, name=None):
...
+    By setting the `reverse` operation to `True`, the sum is performed in the
+    reverse order. In contrast to using `tf.reverse`, this avoids copying the
+    tensor.
+
+    For example:
+
+    ```prettyprint
+    # tensor 'x' is [1, 2, 3, 4, 5]
+    tf.cumsum(x)               ==> [1, 3, 6, 10, 15]
+    tf.cumsum(x, reverse=True) ==> [5, 9, 12, 14, 15]
```
girving: Is that backwards?

tensorflow/python/ops/math_ops.py
```diff
@@ -1719,6 +1795,12 @@ def _LinspaceShape(op):
   return [tensor_shape.vector(num)]
 
 
+@ops.RegisterShape("Cumsum")
+@ops.RegisterShape("Cumprod")
```
girving: `@ops.RegisterShape("Cumsum")(common_shapes.unchanged_shape)`

tensorflow/core/ops/math_ops.cc
ibab: Are those not generated automatically when creating the Python functions in `gen_math_ops`?
I remember this because I copied the docstrings of the generated ones into the Python wrappers.
But I'll be happy to add these if there's a way to avoid duplication in the generated docstrings.

tensorflow/python/ops/math_ops.py
ibab: Yup, it's the same mistake as above.
Thanks for spotting this!

tensorflow/core/ops/math_ops.cc
ibab: I've though about this when writing the op.
I'm not sure how to neatly make the exclusive version accessible in Eigen, except by creating a new `cumsum_exclusive(...)` method.
Maybe it's enough to only add this on the TensorFlow side by optionally padding the tensor with the value that we get from `Reducer::initialize()` ?

girving: Let's get this one in and worry about the exclusive version later if necessary.

tensorflow/core/ops/math_ops.cc
girving: If all the information about them is already there in the prose section above, and the autogenerated docs are fine, it's okay to leave as is.

ibab: The only thing that I added are the default values (like `default: 0`).
But these don't apply to the autogenerated ops, so I think it might be best to leave things as they are now.

bryongloden: Yup @vrv

girving:
Jenkins, test this please.

tensorflow/core/ops/math_ops.cc
ibab: We won't be able to compile this PR right now, as Eigen hasn't been updated to the latest version yet :(
I'll have a look at implementing the `exclusive=True/False` kwarg

tensorflow/python/ops/math_grad.py
bryongloden: @ibab The best. Your going to have to explain this to us at RSA 2017

ibab:
Don't think it'll compile, as the current Eigen version doesn't contain the two fixes that were merged into Eigen two days ago.

girving:
@ibab Alas, re the eigen change.  Is there something I should do to help that process along?  If not, can you send me a mention here when we're ready to go?

girving:
@benoitsteiner Does something need to happen to get those Eigen changes in?

ibab:
We could simply modify the bazel workspace and assorted files to pull the newest Eigen version, but that could interfere with @benoitsteiner's work (might produce a merge/rebase conflict for the affected files).

bryongloden:
Or not. Go for it @ibab

tensorflow/core/ops/math_ops.cc
samjabrahams: @girving I believe the `reverse=True` output is correct. It cumulatively adds up the tensor starting from the end and iterates backwards, but it doesn't reverse the output tensor:

```
cumsum([1, 2, 3, 4, 5])
= [(1), (1 + 2), (1 + 2 + 3), (1 + 2 + 3 + 4), (1 + 2 + 3 + 4 + 5)]
= [1, 3, 6, 10, 15]

cumsum([1, 2, 3, 4, 5], reverse=True)
= [(5), (5 + 4), (5 + 4 + 3), (5 + 4 + 3 + 2), (5 + 4 + 3 + 2 +1)] 
= [5, 9, 12, 14, 15]
```

ibab: @samjabrahams 
It's implemented right now by reversing, calculating `cumsum` and reversing again, so:
```
initial:
  [1, 2, 3, 4, 5]
reverse:
  [5, 4, 3, 2, 1]
cumsum:
  [5, 9, 12, 14, 15]
reverse:
  [15, 14, 12, 9, 5]
```
This mainly exists to simplify the gradients, and can also be used when fixing the gradient for `reduce_prod`.

samjabrahams: I see- very cool stuff. My gut wants to suggest splitting it up into separate `reverse_sum` and `reverse_order` parameters, but I don't want to over-complicate things if it's unnecessary.

If you think that change would be beneficial, I'd be happy to take a crack at coding it; I don't want to burden you with the annoying overhead a change like this would cause. If this change is going to happen, it'd be smart to do this now to avoid backwards compatibility issues.

ibab: I'm slightly against making the reverse option a separate op, as things would get complicated once we introduce an `exclusive` option.
(Leading to  `reverse_exclusive_cumsum` if we continue on the same path)

samjabrahams: I don't think it would require an entirely separate Op- I'm just suggesting splitting up the `reverse` parameter into two distinct reversal options. Right now it indicates that it should both reverse the order of the summation (starting from the back) as well as reverse the order of the output tensor.

My thought is to add two more cases in [this section](https://github.com/tensorflow/tensorflow/pull/2711/files#diff-b6fb1bb5fdce94575b577ae6c5082e2fR32) (one each for just `reverse_sum == true` and just `reverse_order == true`) as well as changing `reverse` into those two separate parameters. The current `if (reverse)` would be converted to `if (reverse_sum && reverse_order)`

ibab: Sorry, I misunderstood your suggestion.
I think that if you simply want to reverse the order of the output (or input), using `tf.reverse` would be much more appropriate.

girving: Let's not split the parameters.  The existing parameter makes intuitive sense: it's whether you accumulate from the left or the right.  If you want something else, use `tf.reverse` directly.

ibab:
I've added exclusive scan functionality to Eigen here: https://bitbucket.org/eigen/eigen/pull-requests/197/implement-exclusive-scan-option-for-tensor/diff.
I also have a version of `ScanOp` that has an optional `exclusive=True/False` argument, which I can push in a new PR once we finish this one.

ibab:
My changes to Eigen have passed in the nightly tests.
@benoitsteiner: Would it be okay to update the Eigen version now?
Is that something I should do as part of this PR?

ibab:
@girving: The Eigen version in TensorFlow has been updated, so we can have a shot at testing this.
I've rebased my changes on top of `master` and squashed them into a single commit.
I've also changed the copyright holder from `Google` to `The TensorFlow Authors`.

The GPU test is segfaulting for me, but I've debugged this a bit and it seems to be a problem that affects many of the tests on my machine.
Everything works if I test it manually and force the device with `tf.device('/gpu:0')`, and the CPU tests are working, so I suspect the tests will simply pass if we run them.

girving:
Jenkins, test this please.

ibab:
Seems like it's also failing on travis.
I can reproduce the segfault on my machine when setting the `opt_level` to `-1` with the following code:
```python
from tensorflow.core.protobuf import config_pb2
config = config_pb2.ConfigProto()
config.graph_options.optimizer_options.opt_level = -1
s = tf.Session(config=config)
with tf.device('/gpu:0'):
  x = np.arange(1, 6).astype(np.float32)
  s.run(tf.cumsum(x, 0, False))
```
Strangely, it works without problems when running with the default opt level.
The segfault occurs when accessing the memory of the input tensor.

girving:
Unpleasant!  Probably a job for a debug+optimized compile and valgrind; I don't have any particular smart ideas for what it might be.

benoitsteiner:
Can you try with the latest version of Eigen (version [3f47637d1e7c1](https://bitbucket.org/eigen/eigen/commits/3f47637d1e7c199b10c66bf1f8d3bf046e479d26)) ? I just fixed a potential issue in the implementation of the evaluator of the scan operation that may result in dangling references once the evaluator is copied.

ibab:
Thanks for fixing the evaluator!
I've compiled with the changes in `3f47637d1e7c1`, but unfortunately the tests are still segfaulting.
I'll try to hunt down the problem using valgrind.

benoitsteiner:
@ibab  The problem is that evalSubExprIfNeeded is executed on CPU, while both your input data and the temporary buffer reside on GPU memory. You're going to have to launch a CUDA kernel in which you call actually call accumulateTo for the scan operations to work on GPU.

ibab:
Thanks!
I also noticed this when I tried to reproduce the problem with `cuda-gdb`, but didn't actually end up in a kernel.
I had assumed that the kernel launch would be automatically handled by the framework (e.g. in `TensorExecutor`)
I can add the launch in a new PR to eigen.

benoitsteiner:
@ibab. TensorExecutor will call the coeff and packet methods from a CUDA kernel. However it will call evalSubExprIfNeeded and cleanup from the host. This makes it possible to do preprocessing on the host if needed. Looking forward to your pull request.

Mistobaan:
@ibab Any update on this ? from the thread seems like there is some work to be done in the eigen library before this PR can be successfully merged. Is that correct ?

ibab:
@Mistobaan I've made the necessary changes in Eigen here: https://bitbucket.org/eigen/eigen/commits/bbffe8ed5f26dfc7166e9fd97ce15bc2655af769
Currently I'm just waiting for an update of the Eigen version that's used in tensorflow.

ibab:
@girving @benoitsteiner Maybe I should update the Eigen version as part of this PR, or in a separate PR?

benoitsteiner:
@ibab I'd recommend that you create a separate PR to update Eigen. From time to time we hit a compatibility issue, and we need to update TensorFlow as well. It's best to have a standalone PR that contains both the Eigen update and the corresponding TF changes.

ibab:
I've updated the Eigen version, so the tests should pass now (they're passing locally for me).

vrv:

vrv:
Cool, @girving this looks ready to review.

girving:
Looks good.  I read through it again, and nothing seems to have changed since the last review.

@ibab Thank you for the contribution!

ibab:
@girving Thanks for reviewing!

PR #2722, opened 2016-06-08
Fixed tensorboard not to use any lambda functions for compatibility with Safari

qbx2:
Fixed tensorboard not to use any lambda functions for compatibility with Safari

Tested on Safari (OS X, iOS), Chrome (Windows, OS X)


dsmilkov:
Thanks! We are fixing this internally (upstream) since it requires updates to other source files as well. Will close this issue once the internal change surfaces as a commit here and will reference it.

PR #2723, opened 2016-06-08
Enable whole batch whitening.

MarvinTeichmann:
The function tf.image.per_image_whitening() can now be applied to a whole batch. `Mean` and `adjusted_stddev` is then computed over all batch dimensions.

I find it very useful to be able to ably tf.image.whitening to an entire batch.
1) In some situations a batch is given (due to previous computation in the graph). Splitting the batch into individual images, in order to be able to apply whitening seams to be a bit redundant.
2) Applying whitening to a batch is statistically superior (and more elegant when used in combination with Batch Normalization).

It might be considered to rename the function in `per_batch_whitening`. As this would now be more accurate.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,7 +10,7 @@
 
 
 ## Big Fixes and Other Changes
- * `tf.nn.per_image_whitening(images)` is now able to handel batches.
+ * `tf.nn.per_image_whitening(images)` is now able to handle batches.
     Parameter got renamed from `image` to `images`.
 
 # Release 0.9.0
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -8,6 +8,11 @@
 * `env.h` replaces use of `New*File()` functions to use `std::unique_ptr`
   return arguments, removing the old raw pointer returns.
 
+
+## Big Fixes and Other Changes
+ * `tf.nn.per_image_whitening(images)` is now able to handel batches.
+    Parameter got renamed from `image` to `images`.
+
 # Release 0.9.0
 
 ## Major Features and Improvements
--- a/tensorflow/python/ops/image_ops.py
+++ b/tensorflow/python/ops/image_ops.py
@@ -780,38 +780,41 @@ def resize_images(images,
   return images
 
 
-def per_image_whitening(image):
-  """Linearly scales `image` to have zero mean and unit norm.
+def per_image_whitening(images):
+  """Linearly scales `images` to have zero mean and unit norm.
 
   This op computes `(x - mean) / adjusted_stddev`, where `mean` is the average
-  of all values in image, and
-  `adjusted_stddev = max(stddev, 1.0/sqrt(image.NumElements()))`.
+  of all values in images, and
+  `adjusted_stddev = max(stddev, 1.0/sqrt(images.NumElements()))`.
 
-  `stddev` is the standard deviation of all values in `image`. It is capped
+  `stddev` is the standard deviation of all values in `images`. It is capped
   away from zero to protect against division by 0 when handling uniform images.
 
   Note that this implementation is limited:
-  *  It only whitens based on the statistics of an individual image.
+  *  It only whitens based on the marginal statistics of the supplied images.
+      A single standard deviation and mean is calculated for all images provided
+      as input to this function.
   *  It does not take into account the covariance structure.
 
   Args:
-    image: 3-D tensor of shape `[height, width, channels]`.
+    images: 4-D Tensor of shape [batch, height, width, channels] or
+            3-D Tensor of shape [height, width, channels].
 
   Returns:
-    The whitened image with same shape as `image`.
+    An output tensor with same shape as `images`.
 
   Raises:
-    ValueError: if the shape of 'image' is incompatible with this function.
+    ValueError: if the shape of 'images' is incompatible with this function.
   """
-  image = ops.convert_to_tensor(image, name='image')
-  _Check3DImage(image, require_static=False)
-  num_pixels = math_ops.reduce_prod(array_ops.shape(image))
+  images = ops.convert_to_tensor(images, name='images')
+  _CheckAtLeast3DImage(images)
+  num_pixels = math_ops.reduce_prod(array_ops.shape(images))
 
-  image = math_ops.cast(image, dtype=dtypes.float32)
-  image_mean = math_ops.reduce_mean(image)
+  images = math_ops.cast(images, dtype=dtypes.float32)
+  images_mean = math_ops.reduce_mean(images)
 
-  variance = (math_ops.reduce_mean(math_ops.square(image)) -
-              math_ops.square(image_mean))
+  variance = (math_ops.reduce_mean(math_ops.square(images)) -
+              math_ops.square(images_mean))
   variance = gen_nn_ops.relu(variance)
   stddev = math_ops.sqrt(variance)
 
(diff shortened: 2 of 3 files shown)


MarvinTeichmann:
I signed it!

mrry:

tensorflow/python/ops/image_ops.py
```diff
@@ -686,50 +686,52 @@ def resize_images(images,
...
+  of all values in images, and
+  `adjusted_stddev = max(stddev, 1.0/sqrt(images.NumElements()))`.
 
-  `stddev` is the standard deviation of all values in `image`. It is capped
+  `stddev` is the standard deviation of all values in `images`. It is capped
   away from zero to protect against division by 0 when handling uniform images.
 
   Note that this implementation is limited:
-  *  It only whitens based on the statistics of an individual image.
+  *  It only whitens based on the statistics of the individual images
```
shlens: Perhaps update this statement?

It only whitens based on the marginal statistics of the supplied images. A single standard deviation and mean is calculated for all images provided as input to this function.

MarvinTeichmann: Good Idea, thx! I have amended it.

vrv:
It would be nice to mention that the name of the attribute has changed in RELEASE.md ... people who use named arguments are going to break (even though it doesn't make sense for them to for a one argument function).

vrv:
Or maybe it's not worth changing the name right now and just leave it as 'image' -- the documentation should hopefully be sufficient.

vrv:
Also, shouldn't there be a test of this behavior?

rmlarsen:
Ping. Any progress on this?

rmlarsen:

rmlarsen:
The test failure seems to be an unrelated timeout of //tensorflow/contrib/learn:linear_test

MarvinTeichmann:
Are you waiting for my response? Whether we should leave the name as `image` or rename to `images` is a design decision. I would leave that up to you guys. Personally I slightly prefer `images`, as this is consistent with the name convention of all other function definition in that file.

I will add a test to `python/ops/image_ops_test.py` tomorrow. I can also add a note to RELEASE.md, if that is desired.

MarvinTeichmann:
Tests are now added as well as a short note to release.md.

MarvinTeichmann:

RELEASE.md
```diff
@@ -8,6 +8,11 @@
 * `env.h` replaces use of `New*File()` functions to use `std::unique_ptr`
   return arguments, removing the old raw pointer returns.
 
+
+## Big Fixes and Other Changes
+ * `tf.nn.per_image_whitening(images)` is now able to handel batches.
```
vrv: handle

vrv:

vrv:

vrv:
@shlens: do you think we should have a per_image_whitening separate from the batch version, since some people might want whitening done per image in a batch, rather than using the batch statistics?

shlens:
After staring hard at the PR, I am starting to wonder whether we might already have this functionality in TF.

Namely, if we wish to whiten all images based on the *batch* mean and *batch* variance, then this is precisely what `tf.nn.batch_normalization` does. That is, I think you might mimic `per_image_whitening` where the mean and variance are computed over the entire batch via a call to `tf.nn_batch_normalization`. For example, in your PR this call:

```python
whitened_images = per_image_whitening(images)
```

becomes

```python
# Compute the minimum variance to mimic per_image_whitening()
num_pixels = tf.reduce_prod(array_ops.shape(images))
min_stddev = tf.inv(tf.sqrt(math_ops.cast(num_pixels, tf.float32)))

# Calculate the mean and variance of the image batch.
axis = list(range(images.get_shape().ndims - 1))
mean, variance = nn.moments(images, axis)

# Whiten the images.
whitened_images = tf.nn_batch_normalization(images, mean, variance, scale=None, offset=None, variance_epsilon=min_stddev)
```

Take a look at the API for [tf.nn.batch_normalization](https://www.tensorflow.org/versions/r0.9/api_docs/python/nn.html#batch_normalization) and please let me know if this works for you. If not, then we should expand the API for `per_image_whitening`.

danmane:
@MarvinTeichmann can you please reply to shlens' comment?

MarvinTeichmann:
Yes, your code should provide a similar / the same result as per_batch_whitening. Did not found the the time yet, to test this.

The use case is somewhat different though. The whitening is supposed to be part of the preprocessing while the batch normalization is used as a regularize. From an API perceptive it might not be the niced to make the user "abuse" the batch normalization part. Like, most user would expect this functionality in the `tf.image` package, where all the other preprocessing can be found.

Additionally, the function `per_image_whitening` is already part of the module. The code you posted would also work for the single image case (just use a batch size of 1). I am not sure, whether it is nice to have completely different approaches for single image/batch size. 

I could imagine, that the `tf.nn_batch_normalization` is faster (more code in C++?). In this case it might be a nice option to rewrite the  `tf.image.per_image_whitening` function, to take use of this. I would however vote for keeping the `tf.image.per_image_whitening` functionality.

vrv:
I guess my only source of confusion is that per_image_whitening makes it sound like whitening is done independently of all other images -- any type of batch statistics no longer makes it 'per image'.  I think I would prefer per_batch_whitening for this new function, and leaving per_image_whitening as is? 

What do you think @shlens ?

MarvinTeichmann:
@vrv Yes, you are write, an `per_image_whitening` is not the optimal naming.

On the other hand, I don't think it is nice to have two functions, `per_image_whitening` and `per_batch_whitening`, as both of them are applying exactly the same tensorflow ops. The _very same computation_ is done in both cases. The only difference is the dimension of the input and output.

Keep in mind, the only real change I have done is to remove the line, which throws an error, if a 4-d tensor is passed to the function. The rest is syntactic sugar (Tests, Parameter Naming, Documentation). This line does violate the general python design idea of being permissive, as it artificially restricts the function to a specific use-case. Therefor I think that this line should not be there.

Function naming is a different issue, but having two restrictive functions does not sound like a great idea, if it is possible to solve the same think with one permissive function.

vrv:
The reason I suggest the different name is, as written, the results of the whitening of one image out of a batch of N depends on the contents of the other images in the batch.   per_image_whitening that operated on batches of images would compute a result where each whitened image is computed independently of the other samples in the batch.

So as written, I would say that this function is per_batch_whitening.  a per_image_whitening that operated on batches could be implemented, by changing the computation to reduce independently of the batch dimension.

shlens:
I think the @MarvinTeichmann  makes some nice points -- namely that it is a simple code change to permit whitening across multiple images. Likewise, the functionality is very similar to the 3-D version of `tf.image.per_image_whitening`. So, from an API perspective, this is very simple and clean.

From a functionality perspective I do share @vrv  concern. This function would be performing in effect a different operation with a different statistical goal then the name of the function suggests (i.e. an operation akin to batch-normalization). I think that the danger of not conveying this subtlety to the user would cause lots of pain down the road so much so that I think it is worth sacrificing some simplicity in the API.

If adding this functionality is important, then my suggestion would be to create a new Python function, i.e.` tf.image.per_batch_whitening` and have it be a light wrapper around `tf.batch_normalization` per the code snippet above. This would provide an efficient implementation and avoid any ambiguity about how statistics are computed.

MarvinTeichmann:
@shlens Thanks for your Input.

Additionally, what do you think about renaming the current `tf.image.per_image_whitening` to `tf.image.per_batch_whitening`?  (i.e remove the current `tf.image.per_image_whitening`).

This would be clean from both API side and also user side perspective. The only disadvantage is, that it breaks compability.

This solution would work even if we use `tf.batch_normalization` as backend. (i.e. the posted code snippet should to the same as `tf.image.per_image_whitening` , if it is applied to a single image).

shlens:
I think we are in agreement then having `tf.image.per_batch_whitening` in the API call `tf.batch_normalization` would be a nice addition to the API.

The secondary question is what to do with `tf.image.per_image_whitening`. I think that we have some policy about retaining backward compatibility within some limits. @vrv would you be able to comment on our degree of backward compatibility in the API?

vrv:
If you want to deprecate it (logging warning), and remove all the existing internal users of it (switching them to per_batch_whitening), go ahead :).

vrv:
(closing due to lack of activity, but will happily re-open if desired -- maybe a new PR will bump this up in the list).

PR #2724, opened 2016-06-08
Fix typo

tvogels:


vrv:
Thanks for this -- though you should make the fix in the original documentation: https://github.com/tensorflow/tensorflow/blob/9078909131d5e91e8bd1878eac54136885656043/tensorflow/core/ops/array_ops.cc#L1718  since the others are automatically generated.

tvogels:
Oops. I tried to update it in that document, but now there is a new pull request: #2752

vrv:
No problem -- you can drop the change from ops.pbtxt -- we auto generate that afterwards.

PR #2733, opened 2016-06-08
MNIST downloader doesn't work as expected

MicaelCarvalho:
See: https://www.tensorflow.org/versions/r0.9/tutorials/mnist/pros/index.html

> For your convenience, we've included a script which automatically downloads and imports the MNIST dataset. It will create a directory 'MNIST_data' in which to store the data files.

Linked to this file: https://github.com/tensorflow/tensorflow/blob/r0.9/tensorflow/examples/tutorials/mnist/input_data.py

Which currently doesn't create any folder.

### DIFF
--- a/tensorflow/examples/tutorials/mnist/input_data.py
+++ b/tensorflow/examples/tutorials/mnist/input_data.py
@@ -26,4 +26,6 @@ import numpy
 from six.moves import urllib
 from six.moves import xrange  # pylint: disable=redefined-builtin
 import tensorflow as tf
-from tensorflow.contrib.learn.python.learn.datasets.mnist import read_data_sets
+
+from tensorflow.examples.tutorials.mnist import input_data
+mnist = input_data.read_data_sets('MNIST_data', one_hot=True)
--- a/tensorflow/examples/tutorials/mnist/input_data.py
+++ b/tensorflow/examples/tutorials/mnist/input_data.py
@@ -27,5 +27,6 @@ from six.moves import urllib
 from six.moves import xrange  # pylint: disable=redefined-builtin
 import tensorflow as tf
 
-from tensorflow.examples.tutorials.mnist import input_data
-mnist = input_data.read_data_sets('MNIST_data', one_hot=True)
+from tensorflow.contrib.learn.python.learn.datasets.mnist import read_data_sets
+
+mnist = read_data_sets('MNIST_data', one_hot=True)


MicaelCarvalho:
Signed.

vrv:
Assigning to tf learn people, since this seems like a bug they should fix instead?

martinwicke:
@ilblackdragon this will likely break some tests. At the very least this change would need some changes to the examples? Maybe it's better to adjust the mnist tutorial text to change the reference to the input.

ilblackdragon:

@martinwicke not sure if it will break tests, let's see. Indeed changing the tutorial text makes sense.

martinwicke:
Jenkins, test this please? For real?

vrv:
Indeed, some failures unfortunately.  For example:

FAIL: //tensorflow/examples/tutorials/mnist:fully_connected_feed_test (see /private/var/tmp/_bazel_jenkins/13e370a18c169b19baeafefb05212b85/tensorflow-pull-requests-mac/bazel-out/local_darwin-opt/testlogs/tensorflow/examples/tutorials/mnist/fully_connected_feed_test/test.log).
INFO: From Testing //tensorflow/examples/tutorials/mnist:fully_connected_feed_test:
==================== Test output for //tensorflow/examples/tutorials/mnist:fully_connected_feed_test:
  File "/private/var/tmp/_bazel_jenkins/13e370a18c169b19baeafefb05212b85/tensorflow-pull-requests-mac/bazel-out/local_darwin-opt/bin/tensorflow/examples/tutorials/mnist/fully_connected_feed_test.runfiles/tensorflow/examples/tutorials/mnist/fully_connected_feed.py", line 27, in <module>
    from tensorflow.examples.tutorials.mnist import input_data
  File "/private/var/tmp/_bazel_jenkins/13e370a18c169b19baeafefb05212b85/tensorflow-pull-requests-mac/bazel-out/local_darwin-opt/bin/tensorflow/examples/tutorials/mnist/fully_connected_feed_test.runfiles/tensorflow/examples/tutorials/mnist/input_data.py", line 30, in <module>
    from tensorflow.examples.tutorials.mnist import input_data
ImportError: cannot import name input_data

MicaelCarvalho:
The initial pull request was plain bad code (can't really describe it better -- shame on me). I fixed the problem with another commit, and I believe it should work now without crashing stuff.

However, I also believe this pull request should be refused, and the tutorial should be corrected instead, just to make clear that the user must run the code to download the dataset himself -- the same code that is already there. The difference in the text is that the said script won't download anything, but will help with it.

Here is why I think it should be refused: This script is used in 3 more files (fully_connected_feed.py, mnist_softmax.py and mnist_with_summaries.py). These files use flags to specify the directory in which the dataset should be saved. Saving the dataset in the current folder would be too strict, and the said scripts would still download it later on their own folders.

TL;DL: It will work for the tutorial, but it will download the dataset without using it. Two command lines are enough, and they are already in the tutorial. Just change the text so it won't say that the script downloads things. Refuse this pull request, please. :-)

vrv:
Sounds good -- we'll close this request.  Feel free to send us a PR to correct our tutorials as you best see fit!

Issue #521, opened 2015-12-16
resize_image_with_crop_or_pad doesn't work with input pipelines


ry:
Because `tf.image.resize_image_with_crop_or_pad()` requires its input to have a fully defined shape, it's not useful as part of an input pipeline since the size of images loaded will not be known in advance.

``` python
  queue = tf.train.string_input_producer(filenames)
  reader = tf.WholeFileReader()
  _, contents = reader.read(queue)
  image = tf.image.decode_jpeg(contents, channels=3)
  cropped = tf.image.resize_image_with_crop_or_pad(image, 224, 224)
```

Ideally the above would work...

martinwicke:
Yes, that's a bug.

eerwitt:
I needed to use the method `resize_image_with_crop_or_pad` as part of an input pipeline and ran into a similar issue.

I'm not yet proficient enough with TensorFlow to tell if this is useful. I made a small workaround for personal use. If this workaround would be a route to use in fixing this issue I'd be happy to update the tests, docs and change out the logic in the existing implementation. If anyone can advise on its usefulness I'd appreciate the feedback.

[Example Implementation of resize_image_with_crop_or_pad](https://gist.github.com/eerwitt/51aba4bffd9ddd5c581c#file-resize_image_with_crop_or_pad_pipeline-py-L6)

po0ya:
@eerwitt It works flawlessly! thanks.

martinwicke:
Send a PR?

kYroL01:
So any news to solve the problem ?

martinwicke:
Other than the workaround posted above, no.

gaohuazuo:
Here is a monkey patching module based on PR https://github.com/tensorflow/tensorflow/pull/2737

[resize_image_patch.py](https://gist.github.com/gaohuazuo/4093eee2aee4d5b8746bff698506ee57#file-resize_image_patch-py)

**Note**
* Remember to pass argument `dynamic_shape=True`
* Input image must have static rank of 3. This can be done by `tf.Tensor.set_shape([None, None, None])`

yongtang:
Looks like the issue has been fixed by PR #2737. Maybe this issue could be closed?

martinwicke:
Thanks!

PR #2737, opened 2016-06-08
Dynamically shaped image support for `tf.image.resize_image_with_crop_or_pad`

gaohuazuo:
This patch is intended for issue https://github.com/tensorflow/tensorflow/issues/521

A new keyword argument `dynamic_shape=False` is added to these functions in `tensorflow.python.image_ops`:
`crop_to_bounding_box` `pad_to_bounding_box` `resize_image_with_crop_or_pad` `_ImageDimensions`

The change is fully backward compatible. The new functionality of `resize_image_with_crop_or_pad` is tested and worked well. However, the dynamic version of `crop_to_bounding_box` `pad_to_bounding_box` is missing many argument checks compared to the static version.

Keeping both a dynamic and a static version of the same function is quite ugly. Maybe we should keep only the dynamic version?

**TODO**
Test cases


tensorflow/python/ops/image_ops.py
```diff
@@ -478,40 +492,46 @@ def crop_to_bounding_box(image, offset_height, offset_width, target_height,
...
 
   Raises:
     ValueError: If the shape of `image` is incompatible with the `offset_*` or
-    `target_*` arguments
+    `target_*` arguments, and `dynamic_shape` is set to `False`.
   """
   image = ops.convert_to_tensor(image, name='image')
-  _Check3DImage(image, require_static=True)
-  height, width, _ = _ImageDimensions(image)
+  _Check3DImage(image, require_static=(not dynamic_shape))
```
martinwicke: These not... are ugly. Can you change the argument from dynamic_shape=False to static_shape_check=True? Then all the functions have the same kind of parameter.

martinwicke:
This is great. Can you add tests for this? I think having the ability to use static checks is useful (sometimes), and backwards compatibility is nice, so I don't think you should remove he static checks.

gaohuazuo:
Should these functions simply skip checks that cannot be performed at graph construction time? That would eliminate the need  of a new argument.

martinwicke:
That would mean that tensors of unknown size will pass all the static tests? That's a good idea. I think the danger is minimal here. The errors in the resize kernels are not too bad anyway.

tensorflow/python/ops/image_ops.py
```diff
@@ -427,44 +433,57 @@ def pad_to_bounding_box(image, offset_height, offset_width, target_height,
...
-  if target_width < width:
-    raise ValueError('target_width must be >= width')
-  if target_height < height:
-    raise ValueError('target_height must be >= height')
-
   after_padding_width = target_width - offset_width - width
   after_padding_height = target_height - offset_height - height
 
-  if after_padding_width < 0:
+  if not isinstance(offset_height, ops.Tensor) and offset_height < 0:
```
gaohuazuo: Using `isinstance(something, ops.Tensor)` here is too clumsy.

Solution 1:
Use some helper function to check whether the value is a "literal" or "symbol".

Solution 2:
Make the framework automatically evaluate all ops that have constant inputs at graph construction time (like C++ constexpr).

martinwicke: Solution 2 is not something that will happen right now (although constant folding does happen before execution)

How about making _is_tensor (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/framework/python/framework/tensor_util.py#L218) public and using that?

vrv:
@gaohuazuo you'll have to ping us when this is ready for review

gaohuazuo:
@martinwicke Ready for review.

tensorflow/contrib/framework/python/framework/tensor_util.py
```diff
@@ -215,8 +216,19 @@ def with_same_shape(expected_tensor, tensor):
     return with_shape(expected_shape, tensor)
 
 
-def _is_tensor(t):
-  return isinstance(t, (ops.Tensor, ops.SparseTensor, variables.Variable))
+def is_tensor(x):
+  """Check for tensor types.
+  If the object represents a tensor, returns True. Equivalent to
```
martinwicke: `True`

tensorflow/contrib/framework/python/framework/tensor_util.py
```diff
@@ -215,8 +216,19 @@ def with_same_shape(expected_tensor, tensor):
...
+  `isinstance(x, [tf.Tensor, tf.SparseTensor, tf.Variable])`.
+
+  Args:
+    x: An python object to check.
+
+  Returns:
+    True if `x` represents a tensor, False if not.
```
martinwicke: `True`, `False`

tensorflow/python/ops/image_ops.py
```diff
@@ -194,12 +195,18 @@ def _ImageDimensions(images):
     images: 4-D Tensor of shape [batch, height, width, channels]
 
   Returns:
-    list of integers [batch, height, width, channels]
+    list of integers [batch, height, width, channels], when static shape is
```
martinwicke: `[batch, height, width, channels]`

tensorflow/python/ops/image_ops.py
```diff
@@ -194,12 +195,18 @@ def _ImageDimensions(images):
...
+    fully defined.
+    list of integer scalar tensors [batch, height, width, channels], when
```
martinwicke: here too

tensorflow/python/ops/image_ops.py
```diff
@@ -427,36 +434,47 @@ def pad_to_bounding_box(image, offset_height, offset_width, target_height,
...
   after_padding_width = target_width - offset_width - width
   after_padding_height = target_height - offset_height - height
 
-  if after_padding_width < 0:
+  if not is_tensor(offset_height) and offset_height < 0:
+    raise ValueError('offset_height must be >= 0')
+  if not is_tensor(offset_width) and offset_width < 0:
+    raise ValueError('offset_width must be >= 0')
+
+  if (all(not is_tensor(i) for i in [target_width, width]) and
```
martinwicke: for sets of 2, is_tensor(target_width) and is_tensor(width) is much easier to read

tensorflow/python/ops/image_ops.py
```diff
@@ -427,36 +434,47 @@ def pad_to_bounding_box(image, offset_height, offset_width, target_height,
...
+      target_width < width):
+    raise ValueError('target_width must be >= width')
+  if (all(not is_tensor(i) for i in [target_height, height]) and
```
martinwicke: here too

tensorflow/python/ops/image_ops_test.py
```diff
@@ -551,21 +567,31 @@ def testError(self):
 class PadToBoundingBoxTest(test_util.TensorFlowTestCase):
```
martinwicke: Can you add a test that checks that it is possible to fail these various checks? I know that should have existed before but now we need it even more.

martinwicke:
Thanks! Added bunch of comments. Mainly: Some doc nits, some nit nits, and, could you add another test that assertRaises appropriately if you pass in illegal values, and with also asserts that if you pass in tensors which have illegal values it still fails (although not with ValueError, but probably with OpError).

gaohuazuo:
@martinwicke Test cases updated. Also added corresponding run time check for every static check, though not sure if these checks worth the cost of more complexity in the graph.

tensorflow/python/ops/image_ops.py
```diff
@@ -212,16 +247,26 @@ def _Check3DImage(image, require_static=True):
 
   Raises:
     ValueError: if image.shape is not a [3] vector.
```
gaohuazuo: `image.shape`

tensorflow/python/ops/image_ops.py
```diff
@@ -212,16 +247,26 @@ def _Check3DImage(image, require_static=True):
...
+
+  Returns:
+    An empty list, if image has fully defined dimensions. Otherwise, a list
```
gaohuazuo: `image`

tensorflow/python/ops/image_ops.py
```diff
@@ -484,24 +532,38 @@ def crop_to_bounding_box(image, offset_height, offset_width, target_height,
...
+                        'height must be >= target + offset.')
+  image = control_flow_ops.with_dependencies(assert_ops, image)
+
+  cropped = array_ops.slice(
+    image,
+    array_ops.pack([offset_height, offset_width, 0]),
+    array_ops.pack([target_height, target_width, -1]))
+
+  cropped_shape = [None if is_tensor(i) else i
+                  for i in [target_height, target_width, depth]]
```
gaohuazuo: misalignment

tensorflow/python/ops/image_ops.py
```diff
@@ -531,43 +593,74 @@ def resize_image_with_crop_or_pad(image, target_height, target_width):
...
+  assert_ops = []
+  assert_ops += _Check3DImage(image, require_static=False)
+  assert_ops += _assert(target_width > 0, ValueError,
+                        'target_width must be > 0.')
+  assert_ops += _assert(target_height > 0, ValueError,
+                        'target_height must be > 0.')
+
+  image = control_flow_ops.with_dependencies(assert_ops, image)
+  # `crop_to_bounding_box` and `pad_to_bounding_box` have their own checks.
+  # Make sure our checks comes first, so that error messages are clearer.
```
gaohuazuo: \# Make sure our checks **come** first ...

tensorflow/python/ops/image_ops.py
```diff
@@ -531,43 +593,74 @@ def resize_image_with_crop_or_pad(image, target_height, target_width):
...
+  if is_tensor(target_height):
+    target_height = control_flow_ops.with_dependencies(
+      assert_ops, target_height)
```
gaohuazuo: line wrap not necessary

tensorflow/python/ops/image_ops_test.py
```diff
@@ -550,45 +655,148 @@ def testError(self):
...
+
+    # Each line is a test configuration:
+    #   (offset_height, offset_width, target_height, target_width), err_msg
+    test_config = (([-1, 0, 4, 4], 'offset_height must be >= 0'),
+                   ([0, -1, 4, 4], 'offset_width must be >= 0'),
+                   ([2,  0, 4, 4], 'height must be <= target - offset'),
+                   ([0,  2, 4, 4], 'width must be <= target - offset'))
+
+    for params, err_msg in test_config:
+      self._assertRaises(x, x_shape, *params, err_msg)
```
gaohuazuo: python 2 compatibility

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

martinwicke:
There are some failing tests in this, can you check?

gaohuazuo:
@martinwicke `//tensorflow/python/image_ops_test` and `//tensorflow/contrib/framework:tensor_util_test` fixed. There are other failing tests, such as `//tensorflow/core/platform/cloud:oauth_client_test`, but I believe they are irrelevant to the PR.

martinwicke:
Sorry, apparently we accumulated conflicts. Can you rebase?

gaohuazuo:
@martinwicke The conflict was caused by `import` statements inserted into the same location. Fixed.

rmlarsen:
It looks like there are still conflicts. Can you check?

gaohuazuo:
Conflicts resolved.

martinwicke:
Jenkins, test this please

martinwicke:
Excellent! Thank you!

PR #2752, opened 2016-06-09
Fix typo 'puth' instead of 'put'

tvogels:
Extract `patches` from `images` and **put** them in the "depth" output dimension.


tvogels:
I updated my original pull request: #2724

PR #2755, opened 2016-06-09
Apply fully connected activation function on flat data

danijar:
`layers.fully_connected` flattens all dimensions of the input but the last, transforms the flat rows using the same weight matrix, and unflattens the result. One use case is to apply the same layer to each output of an RNN like in this example.

``` python
from tensorflow.models.rnn import rnn
from tensorflow.contrib import layers

output, _ = rnn.dynamic_rnn(...)
prediction = layers.fully_connected(output, out_size, tf.nn.softmax)
```

However, the example failed using `tf.nn.softmax` since the activation function was applied on the unflattened result already. This pull request applies the activation function before unflattening so that any activation function that works with normal layers (without flattening) can be used.


martinwicke:
Jenkins, test this please!

PR #2757, opened 2016-06-09
Logistic regression weights: Using a `tf.Tensor` as a Python `bool` is not allowed

petrjanda:
Trying to use logistic_regression weights led me to the exception:

```
  File "train.py", line 22, in <module>
    results = run(X, y, hyperparams, args.model_name)
  File "/opt/app/app/logistic_classifier.py", line 104, in run
    logdir='/opt/app/log/' + model_name
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/estimators/base.py", line 227, in fit
    self._setup_training()
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/estimators/base.py", line 148, in _setup_training
    self._inp, self._out)
  File "/opt/app/app/logistic_classifier.py", line 204, in model
    init_mean=params['init_mean'], init_stddev=params['init_stddev']
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/models.py", line 158, in logistic_regression
    if not class_weight:
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/framework/ops.py", line 475, in __nonzero__
    raise TypeError("Using a `tf.Tensor` as a Python `bool` is not allowed. "
TypeError: Using a `tf.Tensor` as a Python `bool` is not allowed. Use `if t is not None:` instead of `if t:` to test if a tensor is defined, and use the logical TensorFlow ops to test the value of a tensor.
```

which is fixed by attached change.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/models.py
+++ b/tensorflow/contrib/learn/python/learn/models.py
@@ -153,7 +153,7 @@ def logistic_regression(x,
                                   bias)
     # If no class weight provided, try to retrieve one from pre-defined
     # tensor name in the graph.
-    if not class_weight:
+    if class_weight is not None:
       try:
         class_weight = ops.get_default_graph().get_tensor_by_name(
             'class_weight:0')


petrjanda:
I signed it!

vrv:
@ilblackdragon @martinwicke  are there really no tests for this? :)

petrjanda:
@vrv Let me know if you want me to add any. 

I stumbled upon this quite randomly and might need some guidance on where it would be appropriate to put the tests for this.

vrv:
I feel like some test should be written that exercises this behavior, but I'm not too familiar with this code so I've assigned those who might know more.

tensorflow/contrib/learn/python/learn/models.py
```diff
@@ -153,7 +153,7 @@ def logistic_regression(x,
                                   bias)
     # If no class weight provided, try to retrieve one from pre-defined
     # tensor name in the graph.
-    if not class_weight:
+    if class_weight is not None:
```
mikowals: this reverses the intent of the original code and attempts to reassign the class_weight tensor if it exists.

vrv: Exactly why there needs to be a test!

martinwicke:
That's why we're having a special event next week to add tests. :/

ilblackdragon:
@vrv @martinwicke Let's start with class weights were tested until we stopped supporting them. (see tensorflow/contrib/learn/python/learn/tests/base_test.py?l=58) and this before was a deprecation warning which wasn't failing them.

ilblackdragon:
PS. @vrv @martinwicke  really need that test coverage showing here if you really want to keep test coverage up. Jenkins will do

vrv:
@ilblackdragon the code is wrong and there is no test for it, so we need a test to validate this is workign as intended before accepting.  Can you suggest a good place for a test?

ilblackdragon:
We deprecated it in our interface so it's quiet hard to test :)

ilblackdragon:
@petrjanda you can implement test that uses custom Estimator with class weight - that should hit this code.

vrv:
you probably want to revert this, the check is the reverse of what you had before

Issue #1828, opened 2016-04-08
SparseTensor: common unary ops


concretevitamin:
Currently, we have a slew of common unary ops that work on Tensors, but not SparseTensors ([ref](https://www.tensorflow.org/versions/r0.7/api_docs/python/math_ops.html#basic-math-functions)):

```
tf.pow()
tf.exp()
tf.log()

# lower priority?
tf.abs()
tf.neg()
tf.sign()
tf.inv()
tf.square()
tf.round()
tf.sqrt()
tf.ceil()
tf.floor()
```
and so on.

We'd like these ops to work on SparseTensor. These do not change the indices nor shape of `SparseTensor`s, so all that's needed is transform the `.values` field on Python side in O(1) line.

mrry:
Are `log()` and `inv()` well-defined for sparse tensors? Also, doesn't `exp()` make the whole tensor dense? (Or do other frameworks deal with the implicitly-zero elements differently in those cases?)

concretevitamin:
As a first cut, I think having these ops transform only the non-empty entries makes sense and is simple to start with.  (This is what Theano calls "structured", as opposed to "regular" where the implicitly-zero elements are worked on as well.)  

So, `log(), inv(), exp()` should all work on the provided elements and thus will not make the tensor dense.
.

siddharth-agrawal:
@concretevitamin @ebrevdo Added PR for `tf.log()`, let me know whether I'm on the right track.

maniteja123:
Hi everyone, I am new to tensorflow but have been working on it recently and was hoping to contribute. This task might help me understand the basic functionalities like adding ops. Would it be fine if I try implementing these. I suppose @siddharth-agarwal has already worked on some of these. Please let me know whether I can work on this. Thanks.

siddharth-agrawal:
@maniteja123 I would suggest looking at how to implement `tf.pow()` as it is a binary operator. I have already figured out how the unitary operator code and tests need to look like, so I will be putting them out one by one.

ebrevdo:
pow(0, x) is undefined for x == 0, so that requires subtle work.

exp(0) == 1, so that densifies the SparseTensor (therefore it's not very useful)
log(0) = -inf, and i see you just pushed that; we will have to revert it because it densifies the sparse tensor.
inv(0) is not defined, and so is not implementable.

ebrevdo:
@vrv have to assign you for now, to revert the log() change.  feel free to unassign yourself after.

[...]

PR #2763, opened 2016-06-09
Enable tf.log() for SparseTensor

siddharth-agrawal:
Enabled `tf.log()` for `SparseTensor`. Added tests and verified locally. This partially addresses #1828.

### DIFF
--- a/tensorflow/python/kernel_tests/cwise_ops_test.py
+++ b/tensorflow/python/kernel_tests/cwise_ops_test.py
@@ -46,6 +46,19 @@ _XOR = lambda x, y: x ^ y
 _INV = lambda x: ~x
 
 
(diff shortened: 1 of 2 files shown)


tensorflow/python/kernel_tests/cwise_ops_test.py
```diff
@@ -24,6 +24,8 @@
 import numpy as np
 import tensorflow as tf
 
+from tensorflow.python.framework import ops
```
vrv: is this necessary?  Does tf.SparseTensor not work?

siddharth-agrawal: It does work, updated.

vrv:

PR #2767, opened 2016-06-09
Initializer for Input->H & H->H weight matrices for RNNCells

PhABC:
This modification allows initializer (with trainability) for the Input -> Hidden and Hidden to Hidden layer when using the _linear function. Currently, only BasicRNNCell has the new arguments in it's __call__ function, but this could easily be solved. The motivation for such a modification is that it would easily allow different weight architecture for each component of the cell. For example, one could have a one-to-one connection from input -> H, all-to-all from H -> H and one-to-one for H->Output. One could also easily create sparse weight matrices for the input->H and H->H.  

**Example code:**

```
# Initializer ~ 1st matrix acts as an index for which unit is trainable, 2st is the weight of the units
init  = [ tf.constant(np.identity(n_input+n_hidden), dtype = 'float32'), tf.random_normal([n_input+n_hidden,1]) ]

# Whether matrix1 and matrix2 are trainable
train = [False, True]

#Creating cell
cell  = rnn_cell.BasicRNNCell(n_hidden, initializer = init, trainable = train)

#Executing cell
outputs, states = rnn.rnn(cell, Input, initial_state=_istate, dtype=tf.float32)
```

### DIFF
--- a/tensorflow/python/ops/rnn_cell.py
+++ b/tensorflow/python/ops/rnn_cell.py
@@ -13,42 +13,18 @@
 # limitations under the License.
 # ==============================================================================
 
-"""Module for constructing RNN Cells.
-
-## Base interface for all RNN Cells
-
-@@RNNCell
-
-## RNN Cells for use with TensorFlow's core RNN methods
-
-@@BasicRNNCell
-@@BasicLSTMCell
-@@GRUCell
-@@LSTMCell
-
-## Classes storing split `RNNCell` state
-
-@@LSTMStateTuple
-
-## RNN Cell wrappers (RNNCells that wrap other RNNCells)
-
-@@MultiRNNCell
-@@DropoutWrapper
-@@EmbeddingWrapper
-@@InputProjectionWrapper
-@@OutputProjectionWrapper
-"""
+"""Module for constructing RNN Cells."""
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
 import collections
 import math
+import numpy as np
 
 import six
 
 from tensorflow.python.framework import ops
-from tensorflow.python.framework import tensor_shape
 from tensorflow.python.ops import array_ops
 from tensorflow.python.ops import clip_ops
 from tensorflow.python.ops import embedding_ops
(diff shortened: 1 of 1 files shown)


PhABC:
@googlebot Signed it

vrv:
Assigning to RNN API people -- I suspect they are going to push back on these changes without a very compelling reason.  In general, our base libraries should strive to be simple and this seems to be trying to do a little too much.  But we'll see what they say.

PhABC:
@vrv I see. Although all of the added features here are optional, which means it wouldn't impact people that don't want to use them. I could always make it more explicit with a trainIndex matrix argument and the initializer as the second argument, which would get rid of the trainable argument and would make the two matrices separate. Let's see what the RNN group think first, but thank you for your feedback!

lukaszkaiser:
I'm afraid I don't understand the problem this PR is attempting to solve. Initializers are part of variable scope, and all RNN cells rely on this mechanism. If you want to change the initializer, you can set your own function for that in variable scope. Is that mechanism not flexible enough? Why can't you write your own (maybe complex) initializer function and just pass it as initializer to variable scope, so it gets used in all RNN cells?

PhABC:
@lukaszkaiser Yes, you can use initializer for the RNN cells, I did miss that. On the other hand, my feature does not only allow initializers. Here, initializers can be a list of 2 matrices that are multipled together before going through the transfer function. The first matrix can be used as an logical index matrix (composed of 1 and 0) which allow you to have various kind of connectivity patterns ( if you set the learnability for this matrix to zero obviously). For example, you can now have a sparse weight matrix or only allow self-connections (identity matrix), but other designs could be used. The second matrix is your real weight matrix, that you put as learnable. Hence, when they are multipled, only the weights that have a non zero index in the first matrix are used. 

This is convenient as you do not need to reassign the weights at every iteration, which is the case if you want a self-connection (identity) constraint with the current initializer. With a few hundred thousands iterations and multiple variables, this could have an impact on performance.

lukaszkaiser:
It looks like what you're proposing is kind-of a new cell. Maybe a separate cell class for that would be more appropriate? It would be clear then that these are new designs rather than just parameters to re-set initializers -- I think you might prefer that too, right?

PhABC:
Hum, yes that would be interesting. On the other hand, this is something that could be implemented for any type of cell (e.g. LSTM, GRU) and transfer function (not just linear). Also, I don't think I am the only one who want such a feature, as I remember reading a stackoverflow question related to having an identity connectivity matrix, only learnable on the diagonal.

ebrevdo:
If you put it into contrib, you can also use the linear operators in tf.contrib.layers (which we can't move to because rnn is in core).

vrv:
@ebrevdo if you don't see this eventually making it into core, it's best that it lives as a useful library outside of this repo.

ebrevdo:
This does seem to be very specialized code that belongs in its own repository.

ebrevdo:
Closing this for now. If you feel strongly that this belongs in tf.contrib.rnn, consider resubmitting a pr to that location and meeting the python pep8 guidelines.

PR #2792, opened 2016-06-10
Enable tf.exp() for SparseTensor

siddharth-agrawal:
Enabled `tf.exp()` for `SparseTensor`. Added tests and verified locally. This partially addresses #1828.

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -330,6 +330,25 @@ def pow(x, y, name=None):
     return gen_math_ops._pow(x, y, name=name)
 
 
+def exp(x, name=None):
+  """Computes exponential of x element-wise.  \\(y = e^x\\).
+
+  Args:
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/math_ops.py
```diff
@@ -330,6 +330,25 @@ def pow(x, y, name=None):
     return gen_math_ops._pow(x, y, name=name)
 
 
+def exp(x, name=None):
```
vrv: @concretevitamin: crazy thought, but do you think it might be possible to do this within gen_*_ops autogeneration?  It seems unfortunate that we have to add this wrapper for every op in TensorFlow.

vrv:

ebrevdo:
exp(0) == 1, and this implementation is incorrect.  a correct implementation would basically make one giant dense SparseTensor (and is therefore not very useful).

siddharth-agrawal:
@ebrevdo Please look at my comment [here](https://github.com/tensorflow/tensorflow/issues/1828#issuecomment-225338777).

Issue #594, opened 2015-12-23
Will the tensorflow support nesterov momentum optimization?


qixianbiao:
Does anybody have experience about "momentum" and "nesterov momentum"?
 
does the "nesterov momentum" perform much better than than "momentum"?

eduardo-elizondo:
@qixianbiao working on this right now.

Ivanopolo:
@eduardo-elizondo are there any news on that? Probably a branch with a draft implementation?

mschonwe:
+1

metap:
+1

beopst:
+1. Any update on this?

suiyuan2009:
ApplyNesterovMomentumOp's computation of grad is differ from other optimization op, we should overide some functions in optimizer.py

siddharth-agrawal:
@vrv Has this issue been resolved?

vrv:
Yup!

Issue #2292, opened 2016-05-09
Bug on specifying GPU to tutorial example minist


smartcat2010:
I tried to specify GPU ID to run the tutorial example [mnist](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/mnist/fully_connected_feed.py). I change the code to:

    with tf.device('/gpu:3‘):
        # Generate placeholders for the images and labels.
        images_placeholder, labels_placeholder = placeholder_inputs(
            FLAGS.batch_size)
        # Build a Graph that computes predictions from the inference model.
        logits = mnist.inference(images_placeholder,
                                    FLAGS.hidden1,
                                    FLAGS.hidden2)
        # Add to the Graph the Ops for loss calculation.
        loss = mnist.loss(logits, labels_placeholder)

        # Add to the Graph the Ops that calculate and apply gradients.
        train_op = mnist.training(loss, FLAGS.learning_rate)

Then it reports error when running:

tensorflow.python.framework.errors.InvalidArgumentError: Cannot assign a device to node 'global_step': Could not satisfy explicit device specification '/device:GPU:3' because no supported kernel for GPU devices is available
	 [[Node: global_step = Variable[container="", dtype=DT_INT32, shape=[], shared_name="", _device="/device:GPU:3"]()]]
Caused by op u'global_step', defined at:
  File "fully_connected_feed.py", line 232, in <module>
    tf.app.run()
  File "/usr/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 30, in run
    sys.exit(main(sys.argv))
  File "fully_connected_feed.py", line 228, in main
    run_training()
  File "fully_connected_feed.py", line 150, in run_training
    train_op = mnist.training(loss, FLAGS.learning_rate)
  File "/search/guangliang/package/tensorflow/tensorflow/examples/tutorials/mnist/mnist.py", line 125, in training
    global_step = tf.Variable(0, name='global_step', trainable=False)
  File "/usr/lib/python2.7/site-packages/tensorflow/python/ops/variables.py", line 209, in __init__
    dtype=dtype)
...

Then I fix the line 125 in "mnist.py" with the following code:

  with tf.device('/cpu:0'):
    global_step = tf.Variable(0, name='global_step', trainable=False)

Then it reports the following error on rerunning:

[...]

Issue #2285, opened 2016-05-09
the bug of using multiple GPUs, related to tf.Variable pinned to CPU


myme5261314:
### Environment info
Operating System: Ubuntu 14.04

Installed version of CUDA and cuDNN: 7.5 and 4.0.7
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):

If installed from sources, provide the commit hash: 4a4f2461533847dde239851ecebe5056088a828c

### Steps to reproduce
Run the following code
```python
import tensorflow as tf

def main():
    a = tf.Variable(1)
    init_a = tf.initialize_all_variables()
    with tf.Session() as sess:
        sess.run(init_a)

    with tf.device("/gpu:0"):
        b = tf.constant(2)
        init_b = tf.initialize_all_variables()
    with tf.Session() as sess:
        sess.run(init_b)

    with tf.device("/cpu:0"):
        c = tf.Variable(2)
        init_c = tf.initialize_all_variables()
    with tf.Session() as sess:
        sess.run(init_c)

    with tf.device("/gpu:0"):
        d = tf.Variable(2)
        init_d = tf.initialize_all_variables()
    with tf.Session() as sess:
        sess.run(init_d)

if __name__ == '__main__':
    main()
```

### Logs or other output that would be helpful
(If logs are large, please upload as attachment).
```
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcublas.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcudnn.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcufft.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcuda.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcurand.so locally
I tensorflow/core/common_runtime/gpu/gpu_init.cc:102] Found device 0 with properties: 
name: GeForce GTX TITAN X
major: 5 minor: 2 memoryClockRate (GHz) 1.266
pciBusID 0000:05:00.0
Total memory: 12.00GiB
Free memory: 11.02GiB
I tensorflow/core/common_runtime/gpu/gpu_init.cc:102] Found device 1 with properties: 
name: GeForce GTX 980
major: 5 minor: 2 memoryClockRate (GHz) 1.2785
pciBusID 0000:09:00.0
[...]

PR #2798, opened 2016-06-11
add support for nesterov momentum

suiyuan2009:
#594
- I use an indirect way. After every minimal step, var will be var_true - lr \* momentum \* accum, so we need to recover to var_true before compute new var.
- I find it's difficult to implement  nesterov momentum in a direct way due to the code structure of minimal step.
- The first commit only include dense nesterov momentum, I'll add sparse version and docs.
- The code will not pass test currently becase of accuracy problem of tf.half.
- **Anyone has a better suggestion?**

### DIFF
--- a/tensorflow/core/kernels/training_ops.cc
+++ b/tensorflow/core/kernels/training_ops.cc
@@ -177,7 +177,7 @@ struct ApplyMomentum<CPUDevice, T> {
                   typename TTypes<T>::ConstFlat grad,
                   typename TTypes<T>::ConstScalar momentum, bool use_nesterov) {
     accum.device(d) = accum * momentum() + grad;
-    if (use_nesterov == true) {
+    if (use_nesterov) {
       var.device(d) -= grad * lr() + accum * momentum() * lr();
     } else {
       var.device(d) -= accum * lr();
@@ -1678,7 +1678,7 @@ class SparseApplyMomentumOp : public OpKernel {
         auto g = grad_flat.template chip<0>(i);
         auto v = var_flat.template chip<0>(index);
         a = a * a.constant(momentum_scalar) + g;
-        if (use_nesterov_ == true) {
+        if (use_nesterov_) {
           v -= g.constant(lr_scalar) * g +
                a.constant(lr_scalar) * a.constant(momentum_scalar) * a;
         } else {
--- a/tensorflow/core/kernels/training_ops.cc
+++ b/tensorflow/core/kernels/training_ops.cc
@@ -176,9 +176,15 @@ struct ApplyMomentum<CPUDevice, T> {
                   typename TTypes<T>::Flat accum,
                   typename TTypes<T>::ConstScalar lr,
                   typename TTypes<T>::ConstFlat grad,
-                  typename TTypes<T>::ConstScalar momentum) {
-    accum.device(d) = accum * momentum() + grad;
-    var.device(d) -= accum * lr();
+                  typename TTypes<T>::ConstScalar momentum,
+		  bool use_nesterov) {
+    if(use_nesterov){
+    	accum.device(d) = accum * momentum() + grad;
+    	var.device(d) -= grad * lr() + accum * momentum() * lr();
+    }else{
+    	accum.device(d) = accum * momentum() + grad;
+    	var.device(d) -= accum * lr();
+    }
   }
 };
 
@@ -1515,6 +1521,7 @@ class ApplyMomentumOp : public OpKernel {
  public:
   explicit ApplyMomentumOp(OpKernelConstruction* ctx) : OpKernel(ctx) {
     OP_REQUIRES_OK(ctx, ctx->GetAttr("use_locking", &use_exclusive_lock_));
+    OP_REQUIRES_OK(ctx, ctx->GetAttr("use_nesterov", &use_nesterov_)); 
   }
 
   void Compute(OpKernelContext* ctx) override {
@@ -1554,12 +1561,13 @@ class ApplyMomentumOp : public OpKernel {
     const Device& device = ctx->template eigen_device<Device>();
     functor::ApplyMomentum<Device, T>()(device, var.flat<T>(), accum.flat<T>(),
                                         lr.scalar<T>(), grad.flat<T>(),
-                                        momentum.scalar<T>());
+                                        momentum.scalar<T>(), use_nesterov_);
     ctx->forward_ref_input_to_ref_output(0, 0);
   }
 
  private:
   bool use_exclusive_lock_;
+  bool use_nesterov_;
 };
 
 typedef Eigen::ThreadPoolDevice CPUDevice;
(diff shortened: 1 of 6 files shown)


georgedahl:
Can you add a derivation/explanation on how the code correctly implements the Nesterov-style momentum updates? On the face of it, it isn't clear how the correct gradient gets computed so explaining the reparameterizaiton you use and proving it is equivalent to the obvious implementation would be helpful.

suiyuan2009:
@georgedahl 
the obvious implementation of nesterov momentum is as follow
```
1. var' = var - lr_ * momentum_ * accum
2. grad = some_function(var')
3. accum = momentum_ * accum + grad
4. var = var - lr_ * accum
```
In tensorflow, a minimal step contains two parts, one is compute_gradients, another is apply_gradients. In my implenmentation, after apply_gradients, the var stored is actually var - lr_ * momentum_ * accum, so the value of var' will pass to compute_gradients. In the next apply_gradients step, we should first recover var by var = var + lr_ * momentum_ * accum. I combined formulas to simplify codes.

suiyuan2009:
@georgedahl I added sparse version and docs, removed tf.half in test.

georgedahl:
I am going to ICML and I won't get to look at this until I return.

vrv:
ping for @georgedahl

georgedahl:
Sorry to take so long to look at this. I believe this logic is correct and is fine to merge. I defer to @vrv for any other cleanup.

tensorflow/core/ops/training_ops.cc
```diff
@@ -356,6 +358,9 @@ out: Same as "var".
 use_locking: If `True`, updating of the var and accum tensors will be protected
   by a lock; otherwise the behavior is undefined, but may exhibit less
   contention.
+use_nesterov: If `True`, the var pass to compute grad will be
```
vrv: the tensor passed to compute

tensorflow/core/kernels/training_ops.cc
```diff
@@ -176,9 +175,13 @@ struct ApplyMomentum<CPUDevice, T> {
                   typename TTypes<T>::Flat accum,
                   typename TTypes<T>::ConstScalar lr,
                   typename TTypes<T>::ConstFlat grad,
-                  typename TTypes<T>::ConstScalar momentum) {
+                  typename TTypes<T>::ConstScalar momentum, bool use_nesterov) {
     accum.device(d) = accum * momentum() + grad;
-    var.device(d) -= accum * lr();
+    if (use_nesterov == true) {
```
vrv: just if (use_nesterov) {  should be fine

tensorflow/core/kernels/training_ops.cc
```diff
@@ -1672,7 +1678,12 @@ class SparseApplyMomentumOp : public OpKernel {
         auto g = grad_flat.template chip<0>(i);
         auto v = var_flat.template chip<0>(index);
         a = a * a.constant(momentum_scalar) + g;
-        v -= a.constant(lr_scalar) * a;
+        if (use_nesterov_ == true) {
```
vrv: (same here)

suiyuan2009:
@vrv cleanups done

tensorflow/core/ops/training_ops.cc
vrv: (i don't think this was fixed?)

suiyuan2009: fixed now

suiyuan2009:
ping for @vrv

vrv:

vrv:
Looks like the GPU implementation also needs these changes -- can you make the appropriate changes to the GPU version of the functor?

suiyuan2009:
I'll make change to the functor in `tensorflow/core/kernels/training_ops_gpu.cu.cc`, may be I need to find a machine with nvidia gpu since my laptop doesn't support it.

suiyuan2009:
gpu version functor is ok now

suiyuan2009:
ping for @vrv

tensorflow/core/ops/training_ops.cc
```diff
@@ -341,8 +341,10 @@ REGISTER_OP("ApplyMomentum")
     .Output("out: Ref(T)")
     .Attr("T: numbertype")
     .Attr("use_locking: bool = false")
+    .Attr("use_nesterov: bool = false")
     .Doc(R"doc(
-Update '*var' according to the momentum scheme.
+Update '*var' according to the momentum scheme. Set use_nesterov = True if you 
+want to use nesterov momentum.
```
vrv: capitalize Nesterov

tensorflow/core/ops/training_ops.cc
```diff
@@ -369,8 +374,10 @@ REGISTER_OP("SparseApplyMomentum")
     .Attr("T: numbertype")
     .Attr("Tindices: {int32, int64}")
     .Attr("use_locking: bool = false")
+    .Attr("use_nesterov: bool = false")
     .Doc(R"doc(
 Update relevant entries in '*var' and '*accum' according to the momentum scheme.
+Set use_nesterov = True if you want to use nesterov momentum.
```
vrv: same here

vrv:

vrv:

avostryakov:
@suiyuan2009 It's pity but Nesterov Momentum doesn't work with embedding matrix on GPU. The error: "Cannot assign a device to node 'gradients/networks/transpose_1_grad/transpose/x': Could not satisfy explicit device specification '/device:GPU:0' because no supported kernel for GPU devices is available."

Maybe you can fix it?

suiyuan2009:
@avostryakov have you seen #2285 and #2292? if not, try adding "allow_soft_placement=True".

avostryakov:
@suiyuan2009 It's exactly what I meant. I can use allow_soft_placement=True but things are become extremely slow during training because embedding matrix will be stored on CPU device

avostryakov:
I saw that you worked with RMSProb optimizer on the same problem in #2564. It's why I thought that you can do the same here.

suiyuan2009:
@avostryakov the RMSProb optimizer that I write is a sparse version, now it only supports CPU device.
@vrv can you give some advice to @avostryakov ?
I'd like to take a deep look at this GPU bug.

suiyuan2009:
@avostryakov you can try distributed tensorflow, but take care of memory management of your GPUs.

avostryakov:
@suiyuan2009 What is the point to use distributed tensorflow? Do you mean on several machines with GPUs?

vrv:
Yeah, I don't think any of the sparse optimizer training ops are implemented on GPU.  It would be awesome if someone wanted to do this, though it probably requires either cuda or Eigen programming knowledge to implement.

avostryakov:
@vrv I don't have skills in this field but yeah, it could be awesome! By the way, AdamOptimizer can train sparse tensors on GPU, isn't it?

vrv:
I don't see an implementation of SparseApplyAdam on GPU, so unfortunately not.  Until such operations are implemented, it's best to keep the variables on CPU and the computation on GPU.

suiyuan2009:
@avostryakov you can use distributed tensorflow on single machine,[distributed tensorflow](https://www.tensorflow.org/versions/r0.9/how_tos/distributed/index.html#distributed-tensorflow), I guess you run your code on one machine with many GPUs.

avostryakov:
@vrv Thank you for the reply!

PR #2819, opened 2016-06-12
Fixed bug in train_test_split operation.

marckhoury:
If scikit-learn is installed, tensorflow attempts to import the operation train_test_split from scikit-learn. Otherwise it provides a custom implementation. The custom implementation correctly computes the position to split the dataset, but returns the same set of indices for both the training set and the test set. Effectively this operations returns the training set for both the training and test sets.


benoitsteiner:
Jenkins, test this please.

PR #2822, opened 2016-06-13
R0.9

pk449:


martinwicke:
We're not quite done with r0.9, we'll merge it when we publish the non-RC.

PR #2824, opened 2016-06-13
Shrink iOS binaries

martinwicke:


martinwicke:
Internal commits cherry-picked from master. Ignoring CLAbot.

PR #2832, opened 2016-06-13
Branch 124731683

benoitsteiner:


martinwicke:
Ignoring CLA, all contributions came from internal.

PR #2842, opened 2016-06-13
Update version number in pip package setup.py

Robin-des-Bois:
Updated version number. 
Minor change. But I was confused for a short time about the old version number in the newly generated .whl file...

### DIFF
--- a/tensorflow/tools/pip_package/setup.py
+++ b/tensorflow/tools/pip_package/setup.py
@@ -27,7 +27,7 @@ from setuptools import find_packages, setup, Command, Extension
 from setuptools.command.install import install as InstallCommandBase
 from setuptools.dist import Distribution
 
-_VERSION = '0.8.0'
+_VERSION = '0.9.0'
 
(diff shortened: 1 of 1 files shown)


benoitsteiner:
Jenkins, test this please.

vrv:
this is incomplete -- there are a bunch of other places to make this change.

@martinwicke @caisq should we merge the version change from the r0.9 back into master?  Should this have been done first before branching?

martinwicke:
In the past we have merged back after the actual release. There's no reason
not to merge back more often. I'd like the actual version number change to
be a commit on the release branch but that's really only an aesthetic
preference.

martinwicke:
We'll revert this, it breaks a bunch of stuff.

Robin-des-Bois:
I did not realize that the version number should also get updated in all these other places. I should have checked that. I was also not sure about where to base the pull request. 

I am sorry for your inconvenience.

martinwicke:
Don't worry. We shouldn't have merged it.

PR #2847, opened 2016-06-14
[WIP] Gradient reversal op

pumpikano:
"[Unsupervised Domain Adaptation by Backpropagation](http://sites.skoltech.ru/compvision/projects/grl/files/paper.pdf)" introduces a very simple but effective "gradient reversal layer" that is the identity for the forward computation and flips the gradient during backpropagation. This PR implements this as a user op called `FlipGradientOp`. The gradient registration is [here](https://github.com/pumpikano/tf-dann/blob/master/flip_gradient.py) (with a small MNIST experiment in that repo that uses the op).

I created the PR to gauge if there is any interest in getting something like this into core or at least contrib. As you can see, the operation itself is dead simple and very general. The same effect can be achieved by processing gradients before application, but the op implementation means that everything works naturally as a result of autodiff. I may have missed a way to achieve this functionality with existing operations - if so, let me know!

The example implementation here of `FlipGradientOp` effectively multiplies the gradient by -1. It may be worth generalizing this as a `ScaleGradientOp` that scales the incoming gradient by an arbitrary scalar tensor.

I'm happy to get this in mergeable shape if the tensorflow team approves.

### DIFF
--- a/tensorflow/core/user_ops/BUILD
+++ b/tensorflow/core/user_ops/BUILD
@@ -1,0 +1,6 @@
+load("//tensorflow:tensorflow.bzl", "tf_custom_op_library")
+
+tf_custom_op_library(
+    name = "flip_gradient.so",
+    srcs = ["flip_gradient.cc"],
+)
--- a/tensorflow/core/user_ops/flip_gradient.cc
+++ b/tensorflow/core/user_ops/flip_gradient.cc
@@ -22,6 +22,7 @@ class FlipGradientOp : public OpKernel {
 
 REGISTER_OP("FlipGradient")
     .Input("input: T")
+    .Input("s: T")
     .Output("output: T")
     .Attr("T: type")
     .Doc(R"Doc(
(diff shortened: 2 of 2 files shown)


girving:
Do you know about `tf.Graph.gradient_override_map`?  It's a bit ugly, but it lets you do this with unmodified TensorFlow.  Can you check if it suffices for your purposes?

pumpikano:
Thanks @girving, I wasn't aware of `tf.Graph.gradient_override_map`. For my use case, I need to provide a scalar tensor to scale the gradients. To do this, I need this tensor to exist already and use it in the gradient function. I came up with this approach, which I think is acceptable, if a little hacky:

```python
class FlipGradientsBuilder(object):
    def __init__(self):
        self.num_calls = 0

    def __call__(self, x, l):
        grad_name = "FlipGradient%d" % self.num_calls
        @ops.RegisterGradient(grad_name)
        def _flip_gradients(op, grad):
            return [tf.neg(grad) * l]

        g = tf.get_default_graph()
        with g.gradient_override_map({"Identity": grad_name}):
            y = tf.identity(x)
            
        self.num_calls += 1
        return y
    
flip_gradients = FlipGradientsBuilder()
```

This allows me to call `y = flip_gradients(x, l)` any number of times when building the graph with minimal overhead as far as I can tell. I wanted to share this solution in case others find it useful.

girving:
@pumpikano Yep, that's how you'd have to use it.  I agree that `gradient_override_map` is pretty clunky, but given that it works I think it's better to use it than to introduce a special `flip_gradient` op.

pumpikano:
Cool, thanks. I'm closing this PR.

PR #2861, opened 2016-06-14
Fix bugs in Residual Network implementation

ilai-deutel:
- Fix `res_net` implementation by moving  upscaling outside of the `layer_i` loop (bug was introduced with commit 5c145f0e3cabc7e61440ddf32c1ac28f5b9d499e)
- `learn.TensorFlowEstimator.restore()` raises `NotImplementedError`; disable restoration until it is implemented

### DIFF
--- a/tensorflow/examples/skflow/resnet.py
+++ b/tensorflow/examples/skflow/resnet.py
@@ -26,7 +26,6 @@ from __future__ import print_function
 
 from collections import namedtuple
 from math import sqrt
-import os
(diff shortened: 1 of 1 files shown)


gideonite:
I think that I've duplicated this change in https://github.com/tensorflow/tensorflow/pull/2863 without noticing that it had just been fixed by this PR. I do think that my change does a better job at clarifying the ResNet implementation though =)

PR #2863, opened 2016-06-14
fix dim mismatch bug in resnet example.

gideonite:
- There was a bug which confused residual blocks with residual groups,
  which is just  a list of blocks. The output of each block should be of
  the same dimension as the input so that a residual connection can be
  made. Upscaling only occurs between groups.
- Changed some of the variable names to clarify the distinction between
  groups, blocks and layers.

### DIFF
--- a/tensorflow/examples/skflow/resnet.py
+++ b/tensorflow/examples/skflow/resnet.py
@@ -140,22 +140,12 @@ def res_net(x, y, activation=tf.nn.relu):
 
   return learn.models.logistic_regression(net, y)
 
-
 # Download and load MNIST data.
 mnist = input_data.read_data_sets('MNIST_data')
 
 # Restore model if graph is saved into a folder.
 if os.path.exists('models/resnet/graph.pbtxt'):
   classifier = learn.TensorFlowEstimator.restore('models/resnet/')
-else:
-  # Create a new resnet classifier.
-  classifier = learn.TensorFlowEstimator(
-      model_fn=res_net, n_classes=10, batch_size=100, steps=100,
-      learning_rate=0.001, continue_training=True)
-
-classifier = learn.TensorFlowEstimator(
-  model_fn=res_net, n_classes=10, batch_size=100, steps=100,
-  learning_rate=0.001, continue_training=True)
 
 while True:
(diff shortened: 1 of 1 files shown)


gunpowdergreentea:
This patch is EPIC. No more crazy memory crashed on GPU. 💯 % recommend. ez win

gideonite:
Interesting, I do not get the `NotImplementedError` when I run the code on tensorflow v0.8.0. The code to load the saved model is definitely hit when there is a model at specified path.

ilai-deutel:
I think this error has to do with commit 150005ec3f995473ee11913d055fb6fe7e93479f which changed the behavior of `_restore` in `learn/estimators/base.py` with v0.9.0 RC0, but I am not familiar enough with TensorFlow to know if this is an issue or a misuse of `TensorFlowEstimator` (I have encountered this error on two machines with different configurations so I guess this issue is not machine-specific).
That is why I have caught the exception `NotImplementedError` on #2861.

ilblackdragon:
Ref #2950

ilblackdragon:
@gideonite @ilai-deutel I pretty much disabled restore code, because it's not really working well with current way of handling the graph. This example should remove usage of restore and just point into the same directory. If you want to change it or I can create a PR with your commits and the change in saving/restoring.

tensorflow/examples/skflow/resnet.py
```diff
@@ -152,6 +153,10 @@ def res_net(x, y, activation=tf.nn.relu):
       model_fn=res_net, n_classes=10, batch_size=100, steps=100,
       learning_rate=0.001, continue_training=True)
 
+classifier = learn.TensorFlowEstimator(
```
ilblackdragon: you should remove previous lines and `save` call - fit now auto-saves/restores into logdir.

gideonite:
@ilblackdragon Cool. Just pushed a small change. Does this address your comments?

This is probably not the right place to open this conversation but I'm thinking that the examples should somehow get plugged into the automatic testing system. Even if it's just to make sure that the syntax is valid. Should I open a new issue?

rmlarsen:

PR #2869, opened 2016-06-14
Added iOS CI build script

petewarden:


tensorflow/tools/ci_build/builds/ios.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+set -e
+
```
caisq: Add the lines to make sure that this script doesn't get run accidentally on non-OSX environments.
if [[ $(uname) != "Darwin" ]]; then
  echo "ERROR: This makefile build requires OS X, which the current system "\
"is not."
  exit 1
fi

tensorflow/tools/ci_build/builds/ios.sh
```diff
@@ -0,0 +1,37 @@
...
+# Remove any old files first.
+make -f tensorflow/contrib/makefile/Makefile clean
```
caisq: Rather than assuming that this script is run from the root folder of the git repo, consider doing this:
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
make -f ${SCRIPTS_DIR}/../../../contrib/makefile/Makefile clean

tensorflow/tools/ci_build/builds/ios.sh
```diff
@@ -0,0 +1,37 @@
...
+./configure
+make
+sudo make install
+cd ../../../../..
+
+# Compile protobuf for the target iOS device architectures.
+tensorflow/contrib/makefile/compile_ios_protobuf.sh
+
+# Build the iOS TensorFlow libraries.
+tensorflow/contrib/makefile/compile_ios_tensorflow.sh
```
caisq: Please add a comment indicating the path of the artifact generated.

tensorflow/tools/ci_build/builds/ios.sh
```diff
@@ -0,0 +1,37 @@
...
+make -f tensorflow/contrib/makefile/Makefile clean
+rm -rf tensorflow/contrib/makefile/downloads
+
+tensorflow/contrib/makefile/download_dependencies.sh
+
+# Make sure the installed system version of protobuf is up to date.
+cd tensorflow/contrib/makefile/downloads/protobuf/
+./autogen.sh
+./configure
+make
```
caisq: I remember we mentioned options flags to do parallel make so that things can be sped up? Is this still possible?

tensorflow/tools/ci_build/builds/ios.sh
caisq: I think it's probably better to add the file in "tensorflow/contrib/makefile". First, it would make the relative path above less funny. Second, it makes it easier for users interested in iOS builds to find this script.

tensorflow/tools/ci_build/builds/ios.sh
petewarden: Done.

tensorflow/tools/ci_build/builds/ios.sh
petewarden: Done. I've cd-ed to the right directory though, to avoid the repetition of the location variable throughout the rest of the file.

tensorflow/tools/ci_build/builds/ios.sh
petewarden: I'd prefer to leave off those optimizations until we have made sure we're reliably building.

tensorflow/tools/ci_build/builds/ios.sh
petewarden: Done.

petewarden:
Jenkins, test me please.

petewarden:
Jenkins, test this please.

PR #2871, opened 2016-06-15
Make sure that concat complains about concat_dim < 0

### LINKED ISSUE #2868: tf.concat(-1, list_of_tensors) should fail at graph build time, not eval time
Instead it gives an erroneous result:
```
>>> tf.concat(-1, [tf.range(3), tf.range(3)])
<tf.Tensor 'concat_2:0' shape=(6, 3) dtype=int32>
```

When you run `.eval()` you get the expected error message:
`InvalidArgumentError: ConcatOp : Expected concatenating dimensions in the range [0, 1), but got -1`

Of course, it would be even better if negative indices counted axes from the end (like in NumPy), but that's a different matter...

ibab:
`tf.concat` already checks whether `concat_dim` is too large for
the provided arguments at graph construction (if that's possible), but it
should also make sure that `concat_dim >= 0`.

This fixes #2868.

### DIFF
--- a/tensorflow/python/kernel_tests/concat_op_test.py
+++ b/tensorflow/python/kernel_tests/concat_op_test.py
@@ -274,22 +274,22 @@ class ConcatOpTest(tf.test.TestCase):
     # Rank doesn't match.
     with self.assertRaises(ValueError):
       tf.concat(1, [tf.constant(10.0, shape=[4, 4, 4, 4]),
-                           tf.constant(20.0, shape=[4, 4, 4])])
+                    tf.constant(20.0, shape=[4, 4, 4])])
 
     # Dimensions don't match in a non-concat dim.
     with self.assertRaises(ValueError):
       tf.concat(1, [tf.constant(10.0, shape=[1, 2, 1]),
-                           tf.constant(20.0, shape=[3, 2, 1])])
+                    tf.constant(20.0, shape=[3, 2, 1])])
 
     # concat_dim out of range.
     with self.assertRaises(ValueError):
       tf.concat(3, [tf.constant(10.0, shape=[4, 4, 4]),
-                           tf.constant(20.0, shape=[4, 4, 4])])
+                    tf.constant(20.0, shape=[4, 4, 4])])
 
     # concat_dim < 0
     with self.assertRaises(ValueError):
       tf.concat(-1, [tf.constant(10.0, shape=[4, 4, 4]),
-                           tf.constant(20.0, shape=[4, 4, 4])])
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/array_ops.py
```diff
@@ -524,12 +524,15 @@ def _ConcatShape(op):
     # Merge all the non-concat dims, and sum the concat dim to make an
     # output shape.
     concat_dim = int(concat_dim)
+    if concat_dim < 0:
+        raise ValueError("Expected concat_dim >= 0, but got %d" % concat_dim)
```
vrv: two space indent

tensorflow/python/ops/array_ops.py
```diff
@@ -524,12 +524,15 @@ def _ConcatShape(op):
...
+
     output_shape = op.inputs[1].get_shape()
     for value in op.inputs[2:]:
       value_shape = value.get_shape()
       if value_shape.ndims is not None and concat_dim >= value_shape.ndims:
-        raise ValueError("concat_dim is out of range (values rank = %d)" %
-                         value_shape.ndims)
+        raise ValueError("Expected concat_dim in range [0, %d), but got %d" %
+                (concat_dim, value_shape.ndims))
```
vrv: indent here too.

I also suspect you got the order of the arguments incorrect

vrv:
A very simple test to exercise this code path would be great!

tensorflow/python/ops/array_ops.py
ibab: Sorry for messing up the indentation and thanks for spotting the swapped arguments.

tensorflow/python/kernel_tests/concat_op_test.py
```diff
@@ -286,6 +286,11 @@ def testShapeError(self):
       tf.concat(3, [tf.constant(10.0, shape=[4, 4, 4]),
                            tf.constant(20.0, shape=[4, 4, 4])])
 
+    # concat_dim < 0
+    with self.assertRaises(ValueError):
+      tf.concat(-1, [tf.constant(10.0, shape=[4, 4, 4]),
+                           tf.constant(20.0, shape=[4, 4, 4])])
```
vrv: I know you just copy pasted this, but can you fix the indenting here too? :)

ibab: Sure, no problem :)

vrv:
Woohoo!  Thanks.  @tensorflow-jenkins test this please

vrv:

PR #2881, opened 2016-06-15
Qualify references with @org_tensorflow to make function importable

dmansfield:
Qualify references with @org_tensorflow to make function importable by other workspaces

E.g. by syntaxnet which uses this function.

I have signed CLA with Cobite, INC. but googlebot is confused.

### DIFF
--- a/third_party/gpus/cuda/build_defs.bzl
+++ b/third_party/gpus/cuda/build_defs.bzl
@@ -8,7 +8,7 @@ def if_cuda(if_true, if_false = []):
 
     """
     return select({
-        "//third_party/gpus/cuda:using_nvcc": if_true,
-        "//third_party/gpus/cuda:using_gcudacc": if_true,
+        "@org_tensorflow//third_party/gpus/cuda:using_nvcc": if_true,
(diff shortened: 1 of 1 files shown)


dmansfield:
I signed it! Company is Cobite, INC.  If I visit linked page (check your CLA) it's not listed, but when I fill out the form (again) and submit it says it's already been signed.  Not sure what's up.

martinwicke:
CLAbot can be slow with company CLAs, don't worry about it.

dmansfield:
The approach used in the PR doesn't work in practice.  See a this stack overflow which asks the generic question about how to solve: http://stackoverflow.com/questions/37884031/how-to-write-a-bazel-select-macro-that-works-inside-or-outside-of-defining-works

(Also, the email address in the commit was wrong, causing additional problems for CLAbot, which, BTW, did go through! eventually).

I'll submit a different PR once a more robust solution has been found.

dmansfield:
Turns out the code is probably correct, but Bazel is buggy.  See:  https://github.com/bazelbuild/bazel/issues/1425

Issue #2880, opened 2016-06-15
tf.flags.DEFINE_boolean is broken for flags containing dashes


jdoerrie:
`tf.flags.DEFINE_boolean` does not replace dashes with underscores when adding the negated flag, which results in a bug for flags containing dashes.

### Environment info
Operating System: Mac OSX 10.11.5
Pip Package: https://storage.googleapis.com/tensorflow/mac/tensorflow-0.9.0rc0-py3-none-any.whl
TensorFlow Version: 0.9.0rc0

### Steps to reproduce
- `foo.py`:
```python
import tensorflow as tf

tf.flags.DEFINE_boolean('foo-bar', True, "Foo Bar")
tf.flags.FLAGS._parse_flags()
print(tf.flags.FLAGS.__flags)
```

### What have you tried?
- `$ python foo.py`: `{'foo_bar': True, 'foo-bar': True}`
- `$ python foo.py --foo-bar`: `{'foo_bar': True, 'foo-bar': True}`
- `$ python foo.py --nofoo-bar`: `{'foo_bar': True, 'foo-bar': False}`

In the last example it is impossible to get the False value of the flag, `tf.flags.FLAGS.foo-bar` results in an AttributeError, because the dash is interpreted as a minus and `FLAGS.foo` does not exist.

In order to support flags with dashes, Python's argparse replaces dashes with underscores ([docs](https://docs.python.org/dev/library/argparse.html#dest), [code](https://github.com/python/cpython/blob/master/Lib/argparse.py#L1471)).

Fixing this bug should be as simple as substituting `flag_name` with `flag_name.replace('-', '_')` in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/platform/flags.py#L106, but I haven't tested this yet.

craigcitro:
@jdoerrie Yep, good catch. Would you have time to send a PR?

jdoerrie:
I'm not quite sure whether referencing an issue triggers a notification, but I created a PR in the meantime. @craigcitro could you have a look at it when you get the chance?

craigcitro:
This should be fixed by #2888.

PR #2888, opened 2016-06-15
Fix negation of flags with dashes, add test.

jdoerrie:
This fixes issue #2880 by replacing dashes with underscores when adding the negated version of a boolean flag. I also added a test making sure it behaves as expected.

### DIFF
--- a/tensorflow/python/platform/flags.py
+++ b/tensorflow/python/platform/flags.py
@@ -101,9 +101,12 @@ def DEFINE_boolean(flag_name, default_value, docstring):
                               help=docstring,
                               default=default_value,
                               type=str2bool)
+
+  # Add negated version, stay consistent with argparse with regard to
+  # dashes in flag names.
   _global_parser.add_argument('--no' + flag_name,
                               action='store_false',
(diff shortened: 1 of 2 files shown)


vrv:
Thanks!

vrv:
@craigcitro our internal flags library doesn't do this though, so we're unlikely to ever be able to accept code into our repo that has dashes in the flag name, but I guess it's still useful for everyone else?

craigcitro:
@vrv That was my thought -- actually, flags with dashes work just fine in both cases. The only wrinkle here is that externally, we can refer to `FLAGS.name_with_dashes`, whereas internally it would only be `FLAGS['name-with-dashes'].value`. Given that there *is* a sane way to make it work in both cases, does that seem reasonable enough?

vrv:
Yeah, that's what I meant -- we wouldn't be able to use code in TensorFlow that has FLAGS.name_with_dashes, and it would be a pain to convert.  I'm happy to accept this change to be more consistent with argparse, but tf code should avoid FLAGS.name_with_dashes references.

PR #2893, opened 2016-06-16
Deprecate contrib.learn.ops{dropout, dnn}

terrytangyuan:
For dnn, simply added a deprecation message. BTW, there's a bug in `TensorFlowEstimator` where dropout is still happening during testing. But I am guessing that's not critical at this moment since it will be deprecated anyways. 

Question: do we want a separate layer that combines both dnn and dropout in `contrib.layers`, e.g. `dnn_with_dropout`?

### DIFF
--- a/tensorflow/contrib/learn/python/learn/ops/dnn_ops.py
+++ b/tensorflow/contrib/learn/python/learn/ops/dnn_ops.py
@@ -21,7 +21,9 @@ from __future__ import print_function
 
 from tensorflow.contrib import layers
 from tensorflow.contrib.learn.python.learn.ops import dropout_ops
+
 from tensorflow.python.framework import ops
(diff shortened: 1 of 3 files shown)


benoitsteiner:
Jenkins, test this please.

benoitsteiner:
Jenkins, test this please.

rnditdev:
Hello, I'm adjusting some examples to account for this deprecation, could you clarify how the new dnn method works? Because it isn't parameter compatible with the old one?
Maybe a simple call with the old method, and the current equivalent?

terrytangyuan:
Some examples are updated as well. Please check them out.

rnditdev:
Great, I just updated the examples. Many thanks.

PR #2904, opened 2016-06-16
Fix markdown list in dtypes documentation

ibab:
The markdown processor can't deal with empty lines in between list items, which leads to a broken list here:
https://www.tensorflow.org/versions/master/api_docs/python/framework.html#DType

Just removing the empty lines should be fine I think.


mrry:

mrry:
Thanks for the cleanup!

PR #2913, opened 2016-06-16
Merging release notes changes from r0.9 back into master.

vrv:
@martinwicke not sure if we should merge the entire tree back or whether I should cherry-pick one at a time.


martinwicke:
Entire tree is fine. We'll do that after exiting rc, but we might as well
do it in between.

PR #2933, opened 2016-06-17
Make RNN api public.

ebrevdo:
Change: 124305799


ebrevdo:
@martinwicke can you review?

PR #2936, opened 2016-06-17
Fixed makefile problems with new files and zlib dependency

petewarden:


tensorflow/contrib/makefile/Makefile
```diff
@@ -66,7 +66,8 @@ HOST_LIBS := \
 -lstdc++ \
 -lprotobuf \
 -lpthread \
--lm
+-lm \
+-lz
```
vrv: why add this if you strip it out anyway in the code?

petewarden: This section is for the host part of the cross-compilation (which we only do to create the proto_text tool). Since this isn't a 'special' build, and zlib is available on all systems that can handle compilation, I didn't want to add the IS_SLIM_BUILD flag here.

vrv: Got it.  The rest of this change seems fine and unavoidable until we can get better factor out the zlib input buffer.

PR #2947, opened 2016-06-18
Added basic Raspberry Pi example

petewarden:
Also fixed a math error in the iOS examples image decoding.


vrv:
I really hope it de-duped the image, we really don't want images checked into git. :(((

petewarden:
I checked when I used it multiple times in other examples earlier, and it was deduped, so I think we're good. I understand why it's not a good idea in general though!

tensorflow/contrib/pi_examples/README.md
```diff
@@ -0,0 +1,39 @@
+# TensorFlow Raspberry Pi Examples
+
+This folder contains examples of how to build applications for the Raspberry Pi using TensorFlow.
+
```
wolffg: Maybe:
"This folder contains an example of a Raspberry Pi application that uses TensorFlow", as there's only one example.

tensorflow/contrib/pi_examples/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Building the Examples
+
+ - Follow the Raspberry Pi section of the instructions at [tensorflow/contrib/makefile](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/makefile) to compile a static library containing the core TensorFlow code.
+
+ - Install libjpeg, so we can load image files:
+
+```
+sudo apt-get install -y libjpeg-dev
+```
```
wolffg: If we're only providing Ubuntu examples, call that out: "You'll need to make libjpg and its headers available. On Ubuntu, do [x]".

PR #2954, opened 2016-06-20
tf.learn update

martinwicke:
Update contrib/{learn,losses,layers,metrics,framework,slim} to reflect changes since r0.9 branch split.


martinwicke:
Ignoring CLA because all commits are cherry-picks.

martinwicke:
Giving up. This is not worth the risk of breaking all sorts of other things.

PR #2971, opened 2016-06-21
Enable tf.sqrt() for SparseTensor

siddharth-agrawal:
Enabled `tf.sqrt()` for `SparseTensor`. Added tests and verified locally. This partially addresses #1828.

### DIFF
--- a/tensorflow/python/kernel_tests/cwise_ops_test.py
+++ b/tensorflow/python/kernel_tests/cwise_ops_test.py
@@ -115,18 +115,22 @@ class UnaryOpTest(tf.test.TestCase):
                                                     x_init_value=x)
         self.assertAllClose(jacob_t, jacob_n, rtol=1e-5, atol=1e-5)
 
-  def _check(self, result_tensor, result_np, input_sp_t):
+  def _check(self, result_tensor, result_np, input_sp_t, tol):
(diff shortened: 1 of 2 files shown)


maciekcc:
Thanks!

Can you please change this a little and make sure that we use the higher tolerance threshold (1e-3) only for testing sqrt? If the other tests pass with the default value of: 1e-6 let us leave it this way.

siddharth-agrawal:
@maciekcc Updated the PR. Let me know if this is okay.

maciekcc:
Thanks for the update!

Sorry for nit-picking :). Can you please pass the tolerance explicitly in the test body, say:

`self._compareBothSparse(z, np.sqrt, tf.sqrt, tolerance=1e-3)`

pass it down to _check and use `None` as default value. This way you can rely on the defaults defined in the super class. You do not need to keep them synced in two places, and overload only when you want to. Maybe something like this will do:

'
def _check(self, result_tensor, result_np, input_sp_t, tolerance=None):
  ..
  if tolerance is None:
    self.assertAllClose(result_np, result_tensor.values.eval())
  else:
    self.assertAllClose(result_np, result_tensor.values.eval(), rtol=tolerance, atol=tolerance)

`

siddharth-agrawal:
@maciekcc Updated.

maciekcc:
Jenkins, test this please.

PR #3003, opened 2016-06-22
Update Eigen to new version with cumsum/cumprod fixes

ibab:
Would it be okay to update the version of Eigen to one that contains the fixes needed for #2711 to get merged?
This PR changes the version to the one right after the fixes were merged.


ibab:
Eigen has now been updated at https://github.com/tensorflow/tensorflow/commit/bb401633a9a1e5cfc3b6f051e000d9739395d6db, so this isn't needed anymore.

PR #3011, opened 2016-06-23
R0.9 ver2

caisq:
Reduces the test time from about 33 s to 24 s.


caisq:
Will submit as CL internally.

PR #3020, opened 2016-06-24
Support GIF image decode ops by giflib library

buaaliyi:
Refactor the code from the former PR #2578, to use giflib instead of FreeImage. 
@martinwicke Please review this. Thanks


martinwicke:
Cool. While I look at the code in detail, can you remove earth.gif? If you need a gif for tests (I'm assuming you do), can you make or use one that is <20k? We're trying to keep the repo of manageable size.

buaaliyi:
The gif file for test case has been replaced. Thanks

martinwicke:
Jenkins, test this please.

tensorflow/python/ops/image_ops_test.py
```diff
@@ -1279,6 +1279,29 @@ def testShape(self):
...
+    # Read some real GIFs, converting to different channel numbers
+    prefix = 'tensorflow/core/lib/gif/testdata/'
+    inputs = (3, 'cock.gif'),
+    for channels_in, filename in inputs:
+      for channels in 0, 1, 3, 4:
+        with self.test_session() as sess:
+          gif0 = io_ops.read_file(prefix + filename)
+          image0 = image_ops.decode_gif(gif0, channels=channels)
+          gif0, image0 = sess.run([gif0, image0])
+          self.assertEqual(image0.shape, (55, 63, 3))
```
martinwicke: Can you add some tests to make sure the contents of the read gif is what we would expect?

tensorflow/core/ops/image_ops.cc
```diff
@@ -350,6 +350,33 @@ contents: 0-D. PNG-encoded image.
...
+REGISTER_OP("DecodeGif")
+    .Input("contents: string")
+    .Attr("channels: int = 0")
+    .Attr("dtype: {uint8, uint16} = DT_UINT8")
+    .Output("image: dtype")
+    .Doc(R"doc(
+Decode a GIF-encoded image to a uint8 or uint16 tensor.
+
+The attr `channels` indicates the desired number of color channels for the
+decoded image.
```
martinwicke: Can you mention what happens to animated gifs?

raingo:
Hi @buaaliyi 

If only single frame can be decoded, what's the point of using GIF? It will be really helpful, if this can be extended to load all frames. In that case, you may need to take care of the "coalesce". I am happy to help in this effort.

buaaliyi:
Hi @raingo ,
You are right. However, I'm not quite familiar with the 'Tensor' representation to store multiple images in just one Tensor, which compatible with current image decoding method such as jpeg and png. So the code just decode the first frame at this moment, and to be fixed. I'll try in the future. Furthermore, thank you for your help.

raingo:
@martinwicke do you have comments on how to feed in variable length images? Examples from existing operations can help. Thanks!

martinwicke:
If we add reading all frames (which I think would be great, it's the main use of the gif format these days after all), the output should simply be a 4D tensor [frames, height, width, channels]. In case the gif has only a single image, the output would be [1, height, width, channels]. This is different from the decode_jpg, and _png functions, but we can provide a Python convenience function decode_single_frame_gif or similar which collapses the first dimension.

martinwicke:
Jenkins, test this please.

martinwicke:
Since this is included in #3264, I'd rather accept that (it also fixes the test failures). I'll then close this PR. @buaaliyi could you reply to #3264 to make our CLAbot happy?

Issue #2180, opened 2016-05-01
SparseTensor   transpose


nationalflag:
Is  there a way to transpose a SparseTensor without converting it to dense?

mrry:
There's no built-in op for doing this, but you could do something like the following:

```python
input_st = ...  # object of type `tf.SparseTensor`, with 2-D indices

indices = input_st.indices
# Reverse the columns of `indices`
transposed_indices = tf.concat(1, [indices[:, 1:2], indices[:, 0:1]])

dense_shape = input_st.dense_shape
# Reverse the elements of `dense_shape`
transposed_dense_shape = tf.pack([dense_shape[1], dense_shape[0]])

transposed_st = tf.SparseTensor(transposed_indices, input_st.values, transposed_dense_shape)
```

mrry:
(Assigning this to @concretevitamin, since he is looking at the sparse ops right now.)

ebrevdo:
Don't forget to call Reorder Sparse on the result of Derek's code.

girving:
@concretevitamin: Are you planning to work on this?  If not, we should mark it as contributions welcome.

@mrry: `tf.reverse` is a better way to reverse a Tensor that extracting each component and manually reassembling them in the opposite order.

concretevitamin:
I see Eugene has marked it as contributions welcome.  I am happy to review any PRs!

maniteja123:
Hi, I would like working on this. Please let me know if it would be fine to proceed. Thanks.

concretevitamin:
@maniteja123 please feel free to take it.  Besides all of the above suggestions, I'd imagine adding an additional bool flag (`reorder`?) to the Python function can be useful, with default to True.

maniteja123:
Thanks for the response. Could you please clarify some questions ? Would adding a wrapper suffice ? I was thinking ``sparse_ops.py`` where similar functions are there ? Or should it be added to the core code ?

Also the implementation AFAICU would be something like 

```
transposed_indices  = tf.reverse(sp_input.indices), [False, True])
transposed_values = sp_input.values
transposed_shape = tf.reverse(sp_input.shape), [True])
sp_output = tf.SparseTensor(transosed_indices, transposed_values, transposed_shape)
sp_output = tf.sparse_reorder(sp_output)
```

Please let me know if this is  the right approach here ? Thanks.

concretevitamin:
AFAICT, adding a wrapper function to sparse_ops.py sounds good.

[...]

PR #3031, opened 2016-06-24
Add sparse_transpose wrapper

maniteja123:
Attempt to add `transpose` function for SparseTensor as mentioned in #2180.

The commented code is an attempt to use `tf.reverse` to reverse `indices` and `shape`. But right now it raises error since `reverse` doesn't support `int64`.

I did test the changes locally. This is my first attempt to contribute. Please let me know if I have to do something here. Thanks.

### DIFF
--- a/tensorflow/python/kernel_tests/sparse_ops_test.py
+++ b/tensorflow/python/kernel_tests/sparse_ops_test.py
@@ -751,38 +751,20 @@ class SparseTransposeTest(tf.test.TestCase):
         tf.placeholder(tf.float64),
         tf.placeholder(tf.int64))
 
-  def _SparseInputTensorValue(self):
-    ind = np.array([
-        [0, 0],
-        [1, 0], [1, 3], [1, 4],
-        [3, 2], [3, 3]]).astype(np.int64)
-    val = np.array([0, 10, 13, 14, 32, 33]).astype(np.float64)
-
-    shape = np.array([5, 6]).astype(np.int64)
-    return tf.SparseTensorValue(ind, val, shape)
-
-  def _SparseOutputTensorValue(self):
-    ind = np.array([
-        [0, 0], [0, 1],
-        [2, 3],
-        [3, 1], [3, 3],
-        [4, 1]]).astype(np.int64)
-    val = np.array([0, 10, 32, 13, 33, 14]).astype(np.float64)
-
-    shape = np.array([6, 5]).astype(np.int64)
-    return tf.SparseTensorValue(ind, val, shape)
-  
   def testTranspose(self):
     with self.test_session(use_gpu=False) as sess:
-      sp_input = self._SparseTensorPlaceholder()
-      input_val = self._SparseInputTensorValue()
-      sp_output = tf.sparse_transpose(sp_input)
-      expected_output_val = self._SparseOutputTensorValue()
-      
-      output_val = sess.run(sp_output, {sp_input: input_val})
-      self.assertAllEqual(output_val.indices, expected_output_val.indices)
-      self.assertAllEqual(output_val.values, expected_output_val.values)
-      self.assertAllEqual(output_val.shape, expected_output_val.shape)
+      np.random.seed(1618)
+      shapes = [np.random.randint(1, 10, size=rank) for rank in range(1, 6)]
+      for shape in shapes:
+        for dtype in [np.int32, np.int64, np.float32, np.float64]:
+          dn_input = np.random.randn(*shape).astype(dtype)
+          rank = tf.rank(dn_input).eval()
+          perm = np.random.choice(rank, rank, False)
+          sp_input, unused_a_nnz = _sparsify(dn_input)
+          sp_trans = tf.sparse_transpose(sp_input, perm=perm)
+          dn_trans = tf.sparse_tensor_to_dense(sp_trans).eval()
+          expected_trans = tf.transpose(dn_input, perm=perm).eval()
+          self.assertAllEqual(dn_trans, expected_trans)
 
 if __name__ == "__main__":
   googletest.main()
--- a/tensorflow/python/kernel_tests/sparse_ops_test.py
+++ b/tensorflow/python/kernel_tests/sparse_ops_test.py
@@ -743,5 +743,46 @@ class SparseMinimumMaximumTest(test_util.TensorFlowTestCase):
         tf.sparse_maximum(sp_zero, sp_one).eval()
 
 
+class SparseTransposeTest(tf.test.TestCase):
+
+  def _SparseTensorPlaceholder(self):
+    return tf.SparseTensor(
+        tf.placeholder(tf.int64),
+        tf.placeholder(tf.float64),
+        tf.placeholder(tf.int64))
+
+  def _SparseInputTensorValue(self):
+    ind = np.array([
+        [0, 0],
+        [1, 0], [1, 3], [1, 4],
+        [3, 2], [3, 3]]).astype(np.int64)
+    val = np.array([0, 10, 13, 14, 32, 33]).astype(np.float64)
+
+    shape = np.array([5, 6]).astype(np.int64)
+    return tf.SparseTensorValue(ind, val, shape)
+
+  def _SparseOutputTensorValue(self):
+    ind = np.array([
+        [0, 0], [0, 1],
+        [2, 3],
+        [3, 1], [3, 3],
+        [4, 1]]).astype(np.int64)
+    val = np.array([0, 10, 32, 13, 33, 14]).astype(np.float64)
+
+    shape = np.array([6, 5]).astype(np.int64)
+    return tf.SparseTensorValue(ind, val, shape)
+  
+  def testTranspose(self):
+    with self.test_session(use_gpu=False) as sess:
+      sp_input = self._SparseTensorPlaceholder()
+      input_val = self._SparseInputTensorValue()
+      sp_output = tf.sparse_transpose(sp_input)
+      expected_output_val = self._SparseOutputTensorValue()
+      
+      output_val = sess.run(sp_output, {sp_input: input_val})
+      self.assertAllEqual(output_val.indices, expected_output_val.indices)
+      self.assertAllEqual(output_val.values, expected_output_val.values)
+      self.assertAllEqual(output_val.shape, expected_output_val.shape)
+
 if __name__ == "__main__":
   googletest.main()
--- a/tensorflow/python/ops/sparse_ops.py
+++ b/tensorflow/python/ops/sparse_ops.py
@@ -1585,8 +1585,13 @@ def _SparseSparseMaximumMinimumShape(op):  # pylint: disable=invalid-name
   return [tensor_shape.unknown_shape(2), tensor_shape.unknown_shape(1)]
 
 
-def sparse_transpose(sp_input, name=None):
+def sparse_transpose(sp_input, perm=None, name=None):
   """Transposes a `SparseTensor`
+
+  The returned tensor's dimension i will correspond to the input dimension
+  `perm[i]`. If `perm` is not given, it is set to (n-1...0), where n is
+  the rank of the input tensor. Hence by default, this operation performs a
+  regular matrix transpose on 2-D input Tensors.
 
   For example, if `sp_input` has shape `[4, 5]` and `indices` / `values`:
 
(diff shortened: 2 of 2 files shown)


rmlarsen:
Thanks for the contribution. A couple of issues with the current code:

1. We need to support tensors with more than 2 dimensions.
2. tf.transpose is more general than simply reversing the indices, but takes an optional argument with a permutation of [0...n-1]. I think we should support this functionality for sparse tensors as well. I think your can use tf.gather to reorder the columns of the index tensor.

maniteja123:
Thank you for the suggestions. I will look into the code for transpose and try to extend it to n dimensions for sparse tensor too. If I understand ``tf.gather`` correctly, it can permute along the axis=0, but here it needs to be done along axis=1. I couldn't think of any approach other than transposing the indices matrix and gather and transpose back. Could you please let me know how to do this ? Sorry for my questions but I am new to contributing here and am trying to understand the codebase better. Thanks.

maniteja123:
Hi, sorry for disturbing again but was hoping someone can give insight on how to use ``gather`` to reorder the columns (along axis=1). Thanks!

concretevitamin:
Sorry for the delay.

I don't quite understand why `tf.gather()` comes into place here, given the two new requirements?

Here's one potential solution -- I haven't thought through the details! -- recall `indicies : [NNZ, NDIM]` and `values : [NNZ]`.  

We first unpack/split `indices` into `[NNZ, 1, ..., 1]` with `NDIM` ones in the end.  We concatenate it with `values`, obtaining a tensor with shape `[NNZ] + [1]*NDIM + [1]`.  

Now we call `tf.transpose()` on this concatenated tensor, with `perm = [0] + user_perm + [NDIM + 1]`, namely we permute the middle indices.  Finally, transform the shape back and split again.  

If this doesn't work, perhaps we'll need a C++ native kernel.

maniteja123:
Sorry if I am understanding it wrong, but @rmlarsen has mentioned that probably ``tf.gather`` can be used to reorder the columns in terms of ``perm`` but AFAIU it seemed that ``tf.gather`` can only reorder only along axis=0. That was the reason I asked for the help.

Thanks for suggesting the approach. Sorry but I am unable to understand it correctly here.

>
We first unpack/split indices into [NNZ, 1, ..., 1]
>
Indices has shape ``[NNZ, ndim]``. So how does it unpack into [NNZ] + [1]*ndim ? Unpack reduces the rank by at most 1 right. Could you clarify it here please ? Thanks a lot.

concretevitamin:
@rmlarsen gentle ping, we might need a small clarification here ;)

maniteja123:
@rmlarsen Sorry to bother you again but it would be really helpful to have your input here. Thanks a lot !

vrv:
ping @rmlarsen ?

rmlarsen:
Yes, tf.gather only operates on the first dimension, so what I had in mind was:

...make sure that perm is a permutation of [0:ndims-1]...
transposed_indices = tf.transpose(tf.gather(tf.transpose(indices), perm))

instead of the less general

transposed_indices = array_ops.concat(1, [indices[:, 1:2], indices[:, 0:1]])

rmlarsen:
Sorry about the delay BTW.

maniteja123:
No problem! Thanks a lot for the clarification. I will try to make the changes today and get back to you.

maniteja123:
Hi, I have used ``array_ops.transpose(array_ops.gather(array_ops.transpose(indices), perm))`` instead of ``array_ops.concat(1, [indices[:, 1:2], indices[:, 0:1]])`` according to your suggestion. I have run the tests locally and are passing. I have added a simple test, so please do have a look at it and let me know if there are any better tests. Thanks!

danmane:
@rmlarsen please re-review at your convenience

rmlarsen:

tensorflow/python/ops/sparse_ops.py
```diff
@@ -1582,3 +1583,49 @@ def _SparseSparseMaximumMinimumShape(op):  # pylint: disable=invalid-name
...
+    sp_input: The input `SparseTensor`.
+    perm: A permutation of the dimensions of `sp_input`.
+    name: A name prefix for the returned tensors (optional)
+  Returns:
+    A transposed `SparseTensor`.
+
+  Raises:
+    TypeError: If `sp_input` is not a `SparseTensor`.
+  """
+  with ops.op_scope([sp_input], name, "SparseTranspose") as name:
```
rmlarsen: This looks pretty good. Now, I think it would be even better if moved this to be part of the regular transpose op. You could put this code inside an "if isinstance(input, ops.SparseTensor):" check in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/array_ops.py#L995

It would be great if you could update the corresponding gradient and its tests at the same time:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/array_grad.py#L322

it should be pretty easy.

tensorflow/python/kernel_tests/sparse_ops_test.py
```diff
@@ -743,5 +743,46 @@ def testMismatchedShapes(self):
         tf.sparse_maximum(sp_zero, sp_one).eval()
 
 
+class SparseTransposeTest(tf.test.TestCase):
```
rmlarsen: We'd definitely want tests for tensors of rank 0, 1, 2, and 3, and various value datatypes. I'd suggest just generating random tensors and write a generic test that is called or each, instead of hard-coded tests. Move to/merge with https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/transpose_op_test.py.

tensorflow/python/ops/sparse_ops.py
maniteja123: thanks for the help and sorry for the delay. I have tried to put it in array_ops itself but this code uses the transpose function itself on the dense tensor and also the ``sparse_reorder`` function needs to reorder the final values in the transposed sparse tensor. But AFAICS, none of the sparse functions are used in the array_ops, probably because that would create a cyclic dependency. It would be really helpful if you could guide me here. 
Also could you let me know about updating the grad for the sparse tensor. Thanks.

tensorflow/python/kernel_tests/sparse_ops_test.py
maniteja123: Okay sure, just wanted to have a basic test to see if I was doing it correct. Will move the tests to the ``transpose_op_test.py``.

rmlarsen:
@maniteja123 looking into your question...

tensorflow/python/ops/sparse_ops.py
rmlarsen: @maniteja123 You are right, let's keep it as a separate function for now. I think if you extend the tests to cover rank > 2 tensors, we should be good. Maybe it would be easier to generate random sparse tensors and use tf.sparse_tensor_to_dense(t) to convert it to dense, transpose the dense and compare with the "densified" version of the output from your function. See e.g. https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/sparse_ops_test.py#L721

maniteja123: @rmlarsen Thanks a lot for the clarification. I tried to add a test for tensors of rank from 1 to 5 and generate a permutation. I still have not added a gradient as I wasn't able to proceed and have never written one before. Please do let me know if that has to be implemented and any advice is greatly helpful. Thanks.

tensorflow/python/kernel_tests/sparse_ops_test.py
maniteja123: I have changed this test now. Now that it is in ``sparse_ops``,  shall I still move it to ``transpose_op_test.py`` now ?

maniteja123:
Hi I have added generalised tests. Can someone please have a look at the tests and run jenkins if they seem appropriate ? Thanks!

rmlarsen:

tensorflow/python/kernel_tests/sparse_ops_test.py
```diff
@@ -743,5 +743,28 @@ def testMismatchedShapes(self):
...
+
+  def _SparseTensorPlaceholder(self):
```
concretevitamin: Nit: you can replace this with `array_ops.sparse_placeholder()`.

tensorflow/python/kernel_tests/sparse_ops_test.py
```diff
@@ -743,5 +743,28 @@ def testMismatchedShapes(self):
...
+    return tf.SparseTensor(
+        tf.placeholder(tf.int64),
+        tf.placeholder(tf.float64),
+        tf.placeholder(tf.int64))
+
+  def testTranspose(self):
+    with self.test_session(use_gpu=False) as sess:
+      np.random.seed(1618)
```
concretevitamin: Thanks for using my favorite seed.

tensorflow/python/ops/sparse_ops.py
```diff
@@ -1582,3 +1583,49 @@ def _SparseSparseMaximumMinimumShape(op):  # pylint: disable=invalid-name
...
+    if perm is None:
+      rank = array_ops.rank(sp_input)
+      perm = (rank - 1) - math_ops.range(0, rank, 1)
+    indices = sp_input.indices
+    transposed_indices = array_ops.transpose(array_ops.gather(array_ops.transpose(indices), perm))
```
concretevitamin: This line is longer than 80 cols.

rmlarsen:
@maniteja123 Thanks for your contribution!

tensorflow/python/kernel_tests/sparse_ops_test.py
rmlarsen: It's not as good as 42.

maniteja123:
@rmlarsen @concretevitamin Thanks a lot for patiently guiding me throughout the PR. Sorry for failing to address the nitpicks. Is that okay ?

concretevitamin:
No worries!

amrit-poudel:
Hi @maniteja123,

Can we now transpose sparse tensors (dtype float64 and/or complex128) of order/rank greater than 2 ? I just wanted to make sure before I download the most recent version of tensorflow library. Thanks.

maniteja123:
Hi @amrit-poudel Yeah, I believe it should be possible to transpose sparse tensors of rank till the maximum, which I suppose is 5.

amrit-poudel:
Hi @maniteja123, 

Thanks for the quick note and for your input in developing this feature. I appreciate it.

I was just wondering if there was any specific reason why it won't work beyond 5th order.

Thanks.

maniteja123:
Hi @amrit-poudel, even I am not aware of the exact reason but I remember reading that it is because higher ranks don't have common use case. Probably someone experienced can give a better explanation.

amrit-poudel:
Hi @maniteja123, 
I am currently working with sparse tensors of rank 7 and was wondering if it is too difficult, in your experience, to extend transpose function to tensors of higher ranks.  Thanks.

amrit-poudel:
Hi, I get these warnings when using the recently added sparse_transpose function:

WARNING:tensorflow:tf.op_scope(values, name, default_name) is deprecated, use tf.name_scope(name, default_name, values)
WARNING:tensorflow:tf.op_scope(values, name, default_name) is deprecated, use tf.name_scope(name, default_name, values)

Is there a way to update sparse_transpose so that these warnings are avoided?

amrit-poudel:
Hi @maniteja123,

I just checked that spares_transpose function seems to work fine without any error even for sparse tensor of order/rank >5, unlike what you previously stated.

I tested this on build  # 234 (last stable build on cpu only macosx)

Could you or someone please clarify if sparse_transpose function is correctly implemented for sparse tensors of rank greater than 5? Thanks.

rmlarsen:
@amrit-poudel sparse_transpose works correctly for high-rank tensors. You can ignore the deprecation messages about op_scope. They do not affect the result at all.

amrit-poudel:
Thanks for confirming.

amrit-poudel:
How do I suppress this warning:

WARNING:tensorflow:tf.op_scope(values, name, default_name) is deprecated, use tf.name_scope(name, default_name, values) 

This warning originated from recently updated sparse_transpose function in tensorflow.

I tried python -W ignore myprogram.py . However, this warning does not go away.

I also tried export PYTHONWARNINGS="ignore". This did not work either.

Then I tried:

import warnings
 warnings.filterwarnings('ignore')

Did not work.

 My output file is polluted by this warning message. I absolutely want to get rid of it. Please suggest. Thanks!

PR #3053, opened 2016-06-26
Templatize dtype in RGBToHSV and HSVToRGB

siddharth-agrawal:
`RGBToHSV` and `HSVToRGB` currently work only for `float32` inputs. Made the data type a template parameter so that other input types like `float16` and `float64` can be used. This change is towards resolving #1140 for the above mentioned operations.

This change does not enable `float64`, as incorporating new types will also need new tests, something I'm not sure about at this point. (Addressed to whoever looks at the PR) Could you explain how the tests in `core/kernels/colorspace_op_test.cc` can be executed, and how they need to be changed in case `float64` is enabled?

### DIFF
--- a/tensorflow/core/kernels/colorspace_op.h
+++ b/tensorflow/core/kernels/colorspace_op.h
@@ -48,17 +48,18 @@ struct RGBToHSV {
 
     range.device(d) = V - input_data.minimum(channel_axis);
 
-    S.device(d) = (V > 0.f).select(range / V, V.constant(0.f));
+    S.device(d) = (V > T(0)).select(range / V, V.constant(T(0)));
 
-    auto norm = range.inverse() * (1.f / 6.f);
+    auto norm = range.inverse() * (T(1) / T(6));
     // TODO(wicke): all these assignments are only necessary because a combined
     // expression is larger than kernel parameter space. A custom kernel is
     // probably in order.
     H.device(d) = (R == V).select(norm * (G - B),
-                                  (G == V).select(norm * (B - R) + 2.f / 6.f,
-                                                  norm * (R - G) + 4.f / 6.f));
-    H.device(d) = (range > 0.f).select(H, H.constant(0.f));
-    H.device(d) = (H < 0.f).select(H + 1.f, H);
+                                  (G == V).select(
+                                      norm * (B - R) + T(2) / T(6),
+                                      norm * (R - G) + T(4) / T(6)));
+    H.device(d) = (range > T(0)).select(H, H.constant(T(0)));
+    H.device(d) = (H < T(0)).select(H + T(1), H);
   }
 };
 
@@ -72,11 +73,11 @@ struct HSVToRGB {
     auto V = input_data.template chip<1>(2);
 
     // TODO(wicke): compute only the fractional part of H for robustness
-    auto dh = H * 6.f;
-    auto dr = ((dh - 3.f).abs() - 1.f).cwiseMax(0.f).cwiseMin(1.f);
-    auto dg = (-(dh - 2.f).abs() + 2.f).cwiseMax(0.f).cwiseMin(1.f);
-    auto db = (-(dh - 4.f).abs() + 2.f).cwiseMax(0.f).cwiseMin(1.f);
-    auto one_s = -S + 1.f;
+    auto dh = H * T(6);
+    auto dr = ((dh - T(3)).abs() - T(1)).cwiseMax(T(0)).cwiseMin(T(1));
+    auto dg = (-(dh - T(2)).abs() + T(2)).cwiseMax(T(0)).cwiseMin(T(1));
+    auto db = (-(dh - T(4)).abs() + T(2)).cwiseMax(T(0)).cwiseMin(T(1));
+    auto one_s = -S + T(1);
 
     auto R = output_data.template chip<1>(0);
     auto G = output_data.template chip<1>(1);
--- a/tensorflow/core/kernels/colorspace_op.h
+++ b/tensorflow/core/kernels/colorspace_op.h
@@ -24,18 +24,19 @@ namespace tensorflow {
 
 namespace functor {
 
-template <typename Device>
+template <typename Device, typename T>
 struct RGBToHSV {
-  void operator()(const Device &d, TTypes<float, 2>::ConstTensor input_data,
-                  TTypes<float, 1>::Tensor range,
-                  TTypes<float, 2>::Tensor output_data) {
-    auto H = output_data.chip<1>(0);
-    auto S = output_data.chip<1>(1);
-    auto V = output_data.chip<1>(2);
+  void operator()(const Device &d,
+                  typename TTypes<T, 2>::ConstTensor input_data,
+                  typename TTypes<T, 1>::Tensor range,
+                  typename TTypes<T, 2>::Tensor output_data) {
+    auto H = output_data.template chip<1>(0);
+    auto S = output_data.template chip<1>(1);
+    auto V = output_data.template chip<1>(2);
 
-    auto R = input_data.chip<1>(0);
-    auto G = input_data.chip<1>(1);
-    auto B = input_data.chip<1>(2);
+    auto R = input_data.template chip<1>(0);
+    auto G = input_data.template chip<1>(1);
+    auto B = input_data.template chip<1>(2);
 
 #if !defined(EIGEN_HAS_INDEX_LIST)
     Eigen::array<int, 1> channel_axis{{1}};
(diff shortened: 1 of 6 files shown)


tensorflow/core/kernels/colorspace_op.h
```diff
@@ -61,13 +62,14 @@ struct RGBToHSV {
...
-    auto V = input_data.chip<1>(2);
+  void operator()(const Device &d,
+                  typename TTypes<T, 2>::ConstTensor input_data,
+                  typename TTypes<T, 2>::Tensor output_data) {
+    auto H = input_data.template chip<1>(0);
+    auto S = input_data.template chip<1>(1);
+    auto V = input_data.template chip<1>(2);
 
     // TODO(wicke): compute only the fractional part of H for robustness
     auto dh = H * 6.f;
```
rmlarsen: Here I think you should use T(6) to avoid potentially costly promotion for float16. Same everywhere.

siddharth-agrawal: Ah, that's a nice catch. I will fix this.

rmlarsen:

siddharth-agrawal:
@rmlarsen Should I update this PR, or keep it for the follow up PR?

rmlarsen:
Please update this PR.

siddharth-agrawal:
@rmlarsen Updated. Could you respond to the question I posted in the PR description?

rmlarsen:
You should be able to run the specific test with

bazel test ${my_tf_dir}/tensorflow/core/kernels:colorspace_op_test

In term of adding test coverage for all types in colorspace_op_test.cc, I would probably 

1. templatize class RGBToHSVOpTest on the data type 
2. replace the constructor with a MakeOp method (see scatter_op_test.cc for example)
3. make the test code members of class RGBToHSVOpTest 

e.g. something along the lines of

```
template <typename T>
class RGBToHSVOpTest : public OpsTestBase {
 protected:
  void MakeOp(DataType datatype) {
    TF_EXPECT_OK(NodeDefBuilder("rgb_to_hsv_op", "RGBToHSV")
                     .Input(FakeInput(datatype))
                     .Finalize(node_def()));
    TF_EXPECT_OK(InitOp());
  }

  void CheckBlack(DataType datatype) {
      // Black pixel should map to hsv = [0,0,0]
    AddInputFromArray<float>(TensorShape({3}), {T(0), T(0), T(0)});
    TF_ASSERT_OK(RunOpKernel());

  Tensor expected(allocator(), datatype, TensorShape({3}));
  test::FillValues<T>(&expected, {T(0.0), T(0.0), T(0.0)});
  test::ExpectTensorEqual<T>(expected, *GetOutput(0));
  }

  void  CheckGray(DataType datatype) {
    ...
  }
};

typedef RGBToHSVOpTest<float> RGBToHSVOpTest_float;

TEST_F(RGBToHSVOpTest_float, CheckBlack_float) {
  MakeOp(DT_FLOAT);
  CheckBlack(DT_FLOAT);
}

...

// You can use macros to get rid of the duplication here...
typedef RGBToHSVOpTest<double> RGBToHSVOpTest_double;

TEST_F(RGBToHSVOpTest_double, CheckBlack_double) {
  MakeOp(DT_DOUBLE);
  CheckBlack(DT_DOUBLE);
}

```

siddharth-agrawal:
@rmlarsen Is there something I need to do here?

samjabrahams:
I went ahead and gave making these changes a shot. Opened up a pull request on @siddharth-agrawal's fork [here](https://github.com/siddharth-agrawal/tensorflow/pull/1)- I figured that would be better than opening up a separate pull request here. Here are the bullet points of what I added in:

* Registers both HSVToRGB and RGBToHSV for `float` and `double`, setting default value to `DT_FLOAT` for backwards compatibility
* Refactors tests to be reusable with different data types (as @rmlarsen suggested)
* Adds HSVToRBG CheckNegativeDifference test, which was missing for some reason previously
* Cleans up a few missed `float` -> `T` bugs

I also found that the Eigen::half type doesn't want to play nice with [these three lines of code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/colorspace_op.h#L74-L76), which is why I didn't include the `half` type in these changes. Trying to add it in spits out [this error log](http://pastebin.com/qXnqU5K6).

rmlarsen:
@siddharth-agrawal @samjabrahams As commented on the private repo, Sam's refactoring of the tests looks good to me. If you merge the changes into this PR, I think we are ready to merge this.

siddharth-agrawal:
@rmlarsen The change is incomplete according to me, there are some Python tests that will be affected too. This is why I made a separate PR for this change, other the set of changes would just be too big. @samjabrahams if you don't mind, I will make the follow up changes as I have already made all the necessary investigations. There are other ops that you can help with, complete list [here](https://github.com/tensorflow/tensorflow/issues/1140).

rmlarsen:
@siddharth-agrawal SGTM. I'll wait for your followup changes.

samjabrahams:
That's fine, hopefully the bits I added in will help with the rest of the changes you're hoping to do.

I'll add the list of "needs 64-bit floating point support" ops to my list of things I could help out with!

siddharth-agrawal:
@samjabrahams Yes, your changes will make the refactoring a lot easier, thanks!

siddharth-agrawal:
@rmlarsen Do you mean I should update this PR? I was thinking that I will send out a new PR once this is merged in.

samjabrahams:
@siddharth-agrawal I think you might be sending mixed messages- above you said that you don't want to  merge this PR in until you update the related Python tests, but now you're saying it should be merged-in as-is?

Usually pull requests contain a complete set of features, or at least a complete feature on it's own. The changes that you currently have don't change any functionality on their own. This pull-request should probably include all changes related to HSVToRGB and RGBToHSV you'd like to see (including C++/Python tests). If you're worried about the pull-request getting "too big": plenty of pull requests are several-hundred lines long, this one is pretty neatly contained thus far.

siddharth-agrawal:
@samjabrahams Maybe I wasn't clear earlier, in this change I just wanted to templatize the `RGBToHSV` and `HSVToRGB` ops and retain the functionality that they already provide, that is work for type `float`; thus not changing any tests. In the follow up changes (new PR), I wanted to do what you did, add new types and make changes to the corresponding tests.

I don't mind making all those changes in this PR itself, but I prefer keeping PRs short (easy for both the reviewer and the author). If @rmlarsen wants me to make the changes in this PR itself, I will do it.

rmlarsen:
@siddharth-agrawal I had the same impression as @samjabrahams, i.e. that you wanted to add the missing tests. I do prefer that PR bundle new code (in this case new classes) and associated new tests, so I prefer you add the tests to this PR. Since I've already reviewed most of the code it will not be a a lot of work.

siddharth-agrawal:
@rmlarsen I will try to update the PR this weekend, I'm traveling tomorrow.

rmlarsen:
@siddharth-agrawal sounds good, thanks! I'll keep an eye on the PR.

siddharth-agrawal:
@rmlarsen Updated.
@samjabrahams Thanks for the PR! Your changes made the process a lot easier.

rmlarsen:

siddharth-agrawal:
@rmlarsen Looks like `RGBToHSV` and `HSVToRGB` are "core" ops, do I need to change something for the test to pass?

rmlarsen:
@siddharth-agrawal it looks like the backwards compatibility test fails because you have not set a default value for the type.

See http://ci.tensorflow.org/job/tensorflow-pull-requests-cpu-python3/932/consoleFull
<snip>
Getting all registered ops...
Reading op history from tensorflow/core/ops/compat/ops_history.v0.pbtxt...
Verifying updates are compatible...
tensorflow/core/ops/compat/backwards_compatibility_test.cc:36: Failure
Value of: (compatibility.ValidateCompatible(env, &changed_ops, &added_ops, nullptr))
  Actual: Invalid argument: Incompatible Op change: Attr 'T' added without default; old: Op<name=HSVToRGB; signature=images:float -> output:float>; new: Op<name=HSVToRGB; signature=images:T -> output:T; attr=T:type,allowed=[DT_FLOAT, DT_DOUBLE]>
Expected: ::tensorflow::Status::OK()


Can you please set the default to float to ensure backwards compatibility?

siddharth-agrawal:
@rmlarsen Updated the PR.

rmlarsen:
@siddharth-agrawal test this please

rmlarsen:
Oops! :-)

rmlarsen:
@vrv  I reviewed this, just waiting for all tests to pass, then I'll merge it.

siddharth-agrawal:
@rmlarsen Can I delete the branch that I used? Generally, Github shows a message saying that it is safe to delete the branch, but not here. Is it because of the multiple commits?

PR #3058, opened 2016-06-27
Branch 125939528

maciekcc:


martinwicke:
CLAbot is wrong, all commits are from internal.

PR #3063, opened 2016-06-27
Fix Nightly Python3 error in graph_io_test.py

caisq:
Change: 125988941


caisq:

PR #3064, opened 2016-06-27
Explanation of blank label in ctc_loss

igormq:
The doc string of ctc_loss is lacking of several details. I tried to explain a little better of what is needed to get it working, but english is not my native language, so I'm sure that you can do much better!

### DIFF
--- a/tensorflow/python/ops/ctc_ops.py
+++ b/tensorflow/python/ops/ctc_ops.py
@@ -47,6 +47,18 @@ def ctc_loss(inputs, labels, sequence_length,
   max(labels.indices(labels.indices[:, 1] == b, 2))
     <= sequence_length(b) for all b.
   ```
+  
+  Useful Tips:
+  
+  The ctc_loss has as input an affine projection of RNN state:
+    inputs_t = W*h^L_{t} + b, for t = {1,...,max_time}
+  where h is the hidden state of last rnn calculated at timestep t.
+  
+  The `inputs` argument has size num_labels + 1 where +1 is reserved to
+  blank label, e.g., for a vocab = {a, b, c} the num_classes will be 4 
+  where {a: 0, b: 1, c: 2, blank: 3}.
+  
+  `ctc_loss` function already interleaves blank label between the target labels.
 
   Regarding the arguments `preprocess_collapse_repeated` and
   `ctc_merge_repeated`:
@@ -84,7 +96,7 @@ def ctc_loss(inputs, labels, sequence_length,
 
   Args:
     inputs: 3-D `float` `Tensor` sized
-      `[max_time x batch_size x num_classes]`.  The logits.
+      `[max_time x batch_size x num_classes]`.  The logits. 
     labels: An `int32` `SparseTensor`.
       `labels.indices[i, :] == [b, t]` means `labels.values[i]` stores
       the id for (batch b, time t).  See `core/ops/ctc_ops.cc` for more details.


vrv:
ping for @ebrevdo

ebrevdo:
Can you rebase the conflict?

ebrevdo:
Jenkins, test this please.

tensorflow/python/ops/ctc_ops.py
```diff
@@ -47,6 +47,18 @@ def ctc_loss(inputs, labels, sequence_length,
   max(labels.indices(labels.indices[:, 1] == b, 2))
     <= sequence_length(b) for all b.
   ```
+  
+  Useful Tips:
+  
+  The ctc_loss has as input an affine projection of RNN state:
```
ebrevdo: The input to ctc_loss does not need to be an affine projection of RNN state.  It can be anything; so long as it's a 3-D tensor.

tensorflow/python/ops/ctc_ops.py
```diff
@@ -84,7 +96,7 @@ def ctc_loss(inputs, labels, sequence_length,
 
   Args:
     inputs: 3-D `float` `Tensor` sized
-      `[max_time x batch_size x num_classes]`.  The logits.
+      `[max_time x batch_size x num_classes]`.  The logits.
```
ebrevdo: undo this whitespace addition.

tensorflow/python/ops/ctc_ops.py
```diff
@@ -47,6 +47,18 @@ def ctc_loss(inputs, labels, sequence_length,
...
     <= sequence_length(b) for all b.
   ```
+  
+  Useful Tips:
+  
+  The ctc_loss has as input an affine projection of RNN state:
+    inputs_t = W*h^L_{t} + b, for t = {1,...,max_time}
+  where h is the hidden state of last rnn calculated at timestep t.
+  
+  The `inputs` argument has size num_labels + 1 where +1 is reserved to
```
ebrevdo: **Note**: The `inputs` Tensor's innermost dimension size, `num_classes`, represents
`num_labels + 1` classes, where `num_labels` is the number of true labels, and the
largest value (`num_classes - 1`) is reserved for the blank label.

For example, for a vocabulary containing 3 labels `[a, b, c]`, `num_classes = 4` and the
labels indexing is `{a: 0, b: 1, c: 2, blank: 3}`.

tensorflow/python/ops/ctc_ops.py
```diff
@@ -47,6 +47,18 @@ def ctc_loss(inputs, labels, sequence_length,
...
+  blank label, e.g., for a vocab = {a, b, c} the num_classes will be 4 
+  where {a: 0, b: 1, c: 2, blank: 3}.
+  
+  `ctc_loss` function already interleaves blank label between the target labels.
```
ebrevdo: i don't understand this line; perhaps remove it?

tensorflow/python/ops/ctc_ops.py
ebrevdo: (remember to use backticks appropriately)

tensorflow/python/ops/ctc_ops.py
ASDen: I think this is an important note, he points out that you *don't* need to encode the target labels (by interleaving with blank symbol) before passing to ctc_loss, this is internally done for you!

it took me a while to grasp this from the Cpp code, since there is no examples for ctc loss/decode currently (you can only find the straight forward usage in the test code)

tensorflow/python/ops/ctc_ops.py
ASDen: a more precise description is on ctc_ops.cc
"This class performs the softmax operation for you, so inputs should be e.g. linear projections of outputs by an LSTM."
this also gives the important undocumented info that there is no need to apply softmax before supplying inputs, this is done internally for you

ebrevdo: Indeed; perhaps it would be good to bring some of the documentation out of ctc_ops.cc into here as part of this PR.

tensorflow/python/ops/ctc_ops.py
ebrevdo: yes - the test code is meant to be an example of usage.

if you had to provide the blank labels yourself, it wouldn't be CTC :)

perhaps the appropriate documentation would be around line 101 saying that `labels.values[i]` must take on values in `[0, num_labels)`.

igormq: Yes, I agree @ebrevdo, but only looking the ctc_ops.py documentation turns unclear how the blank label is inserted (and its's index).

danmane:
@igormq can you resolve the comments from code review please?

igormq:
I messed up my github and deleted my fork of tensorFlow (and i can't edit this PR). So, I'll have to submit it again and do another pull request. That is ok for you?

vrv:
sure!

igormq:
PR #3713. :)

PR #3072, opened 2016-06-28
Update framework.md

wulingdong:
# Create a nested scope called "inner_1".

```
with g.name_scope("inner"): 
```

code conflict with comment。shall be "inner_1"


tensorflow/g3doc/api_docs/python/framework.md
```diff
@@ -330,7 +330,7 @@ with tf.Graph().as_default() as g:
       assert nested_inner_c.op.name == "nested/inner/c"
 
     # Create a nested scope called "inner_1".
-    with g.name_scope("inner"):
```
rmlarsen: This will actually automatically append "_1" to the name to make it unique, since there is already an existing scope called "inner". See the discussion of the name in the documentation above.

Issue #110, opened 2015-11-11
Support for Redhat, Centos and many superclusters


trungnt13:
Many clusters system using module with Redhat or Centos < 7 which is **glibc 2.12**

Since, **bazel** requires **glibc 2.14** and the prebuilt version for linux requires **glibc 2.17**. It is hopeless to make tensorflow run on clusters.

Referred to this issue reported on bazel: https://github.com/bazelbuild/bazel/issues/583

vrv:
Since we depend on bazel, this sounds like a bazel issue.

Feel free to re-open if bazel ends up supporting 2.12 or lower, and we can see what we can do.

alantus:
Am I right that you depend on bazel only at build-time? If this is true then why don't you drop the glibc-2.17 dependency? You could also release static-linked packages that would be very useful to people stuck on clusters with old libraries...

urimerhav:
So did anyone find some way past this problem? I'm using redhat 6.4, as is my entire corporation. We're stuck on redhat 6.4. I'm not sure how to end up running tensorflow on such a machine...

ttrouill:
I managed to have it running on a CentOS 6.7 : http://stackoverflow.com/a/34897674/1990516 :)
Tell me if it works for you

urimerhav:
Thanks man! I'll look into it as soon as I can.

altaetran:
Could you let me know if this worked? I can't seem to get any of these other solutions working.

urimerhav:
Since @ttrouill only says he got it working on 6.7 so I didn't check whether this works on 6.4 actually...

rdipietro:
Both solutions seem to work, but they're not optimal. TensorFlow and Python seem to run okay, but if I try and run IPython, then with the first solution I get an Invalid ELF error, and with the second solution there is a memory leak and IPython continues to absorb all memory with time. I believe that this can also happen with other Python imports that rely on libraries that were compiled using the older libc.

I'd love to see a straightforward how-to-compile-bazel-with-old-glibc guide, but I haven't come across one yet.

rdipietro:
Also https://github.com/bazelbuild/bazel/issues/760 is relevant, but it's far from straightforward and my attempt to build bazel using this guide failed. Hopefully within the next few weeks I can give it some more time and continue that thread with the errors I end up getting.

rdipietro:
[...]

PR #3077, opened 2016-06-28
Remove explicit dependency on Python 2.7 from crosstool wrapper

jdoerrie:
Many superclusters need to compile TensorFlow from source due to an outdated glibc version (see #110). In @rdipietro's excellent workaround post (https://github.com/tensorflow/tensorflow/issues/110#issuecomment-219790730) he mentions issues with the referenced Python version in this file. I have issues as well, but of a different nature. In my case the build script is unable to find `libpython2.7.so.1.0`, since only Python 3 is present on my machine. The issue originates from `crosstool_wrapper_driver_is_not_gcc` where the only Python 2.7 exclusive feature is the `print` statement. By `import`ing `print_function from __future__` the explicit dependency can be dropped and both versions of Python are supported.


vrv:

PR #3090, opened 2016-06-28
Fix the TensorBoard README link on the how to page.

RenatoUtsch:
Change: 125799092

Backporting this change from master to r0.9 to fix the link in the r0.9 TensorBoard how-to too.


danmane:
LGTM

PR #3091, opened 2016-06-28
Resolve the libcudnn library to its SONAME, when possible,

vrv:
and build the binary with that SONAME, which is supposed to be
ABI-compatible.

When we build our pip packages, we will now build with precisely
the version of the SONAME that we intend to support with the
binary packages.  This is the solution recommended by @3XX0.

Tested by installing cudnn 5.1.3 RC and seeing that the source code
modifications show the use of version "5" (the SONAME version
pointed to by libcudnn.so symlink).

In addition, installing the pip package shows:

> > > import tensorflow as tf
> > > I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcublas.so locally
> > > I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcudnn.so.5 locally
> > > I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcufft.so locally
> > > I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcuda.so.1 locally
> > > I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcurand.so locally

### DIFF
--- a/configure
+++ b/configure
@@ -174,6 +174,19 @@ while true; do
 
   if [[ -z "$TF_CUDNN_VERSION" ]]; then
     TF_CUDNN_EXT=""
+    # Resolve to the SONAME of the symlink.  Use readlink without -f
+    # to resolve exactly once to the SONAME.  E.g, libcudnn.so ->
+    # libcudnn.so.4
+    REALVAL=`readlink ${CUDNN_INSTALL_PATH}/lib64/libcudnn.so`
+
+    # Extract the version of the SONAME, if it was indeed symlinked to
+    # the SONAME version of the file.
+    if [[ "$REALVAL" =~ .so[.]+([0-9]*) ]];
+    then
+      TF_CUDNN_EXT="."${BASH_REMATCH[1]}
+      TF_CUDNN_VERSION=${BASH_REMATCH[1]}
+      echo "libcudnn.so resolves to libcudnn${TF_CUDNN_EXT}"
+    fi
   else
     TF_CUDNN_EXT=".$TF_CUDNN_VERSION"
   fi


vrv:
@3XX0: is this what you were thinking?

configure
```diff
@@ -174,6 +174,19 @@ while true; do
 
   if [[ -z "$TF_CUDNN_VERSION" ]]; then
     TF_CUDNN_EXT=""
+    # Resolve to the SONAME of the symlink.  Use readlink without -f
+    # to resolve exactly once to the SONAME.  E.g, libcudnn.so ->
+    # libcudnn.so.4
+    REALVAL=`readlink ${CUDNN_INSTALL_PATH}/lib64/libcudnn.so`
```
zheng-xq: Out of curiosity, why only one level? Why not down to the real version, such as "4.0.7"? 

Also do we want to handle the darwin case, which is used a few lines below.

vrv: Because the SONAME implies the binary compatibility -- if a user has 4.0.7 on their system and we build with 4.0.4, it should be compatible -- that's the ABI compatibility.  This is going off of what @3XX0 said https://github.com/tensorflow/tensorflow/issues/2525#issuecomment-223523008 there

I'm not entirely sure what I'm supposed to do for OS X -- @yaroslavvb do you know what the dylib's look like?

zheng-xq: LGTM

yaroslavvb: Installing CuDNN 5.0 only gave me this file on MacOS
`-rwxr-xr-x@ 1 yaroslavvb  staff  58270280 Apr 23 03:19 /usr/local/cuda/lib/libcudnn.5.dylib
`
I then manually added this symlink to make it work
`lrwxr-xr-x  1 yaroslavvb  admin        16 Jun 17 04:24 /usr/local/cuda/lib/libcudnn.dylib -> libcudnn.5.dylib
`

vrv: Ah, if there's no symlink by default, then this branch won't execute, so we're okay.  If the symlink does exist, the existing codepath will work too.

vrv:
Ok, merging, hopefully this will help in the future.

PR #3094, opened 2016-06-29
Fix path to boringssl_err_data_c in workspace.bzl.

lesniewski:
Passing in an absolute path for `path_prefix` gave me this error:

```
invalid package name '.../boringssl': package names may not start with '/'.
```

I believe the intent was to refer to the target within the TF repo, and
I think that `tf_repo_name` is supposed to accomplish this.


rmlarsen:

martinwicke:
@kirilg you added this rule, can you make sure this works for you?

kirilg:
Looks good, verified that this works for TensorFlow Serving.

PR #3096, opened 2016-06-29
Wide and Deep docs update

martinwicke:


martinwicke:
Failure is infrastructure flake, rerun here: http://ci.tensorflow.org/job/tensorflow-pull-requests-mac/1021/

msperlich:
This Wide and Deep tutorial looks really interesting, it makes use of `tf.contrib.layers.embedding_column`, which doesn't seem to be part of tensorflow yet. Is the code for the embedding column going to be added or made available somewhere else?

martinwicke:
It's on master, but not on 0.9. Arguably these docs should not be on 0.9 either. :/

That confusion should be resolved with the next release.

Issue #3083, opened 2016-06-28
ValueError in tensorflow/examples/skflow/mnist_weights.py


tcfuji:
### Environment info
OS: OS X 10.11.5

Cuda version: 7.5.30

cuDNN version: v5.1rc (or 5103)

Tensorflow version: 0.9.0rc0

Installed with cuda support using these [instructions](https://medium.com/@fabmilo/how-to-compile-tensorflow-with-cuda-support-on-osx-fd27108e27e1#.mf4zcsmu3).

### Steps to reproduce:

1. cd tensorflow/examples/skflow/
2. $ python mnist_weights.py

### Output:
```
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcublas.7.5.dylib locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcudnn.5.dylib locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcufft.7.5.dylib locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcuda.dylib locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcurand.7.5.dylib locally
Extracting MNIST_data/train-images-idx3-ubyte.gz
Extracting MNIST_data/train-labels-idx1-ubyte.gz
Extracting MNIST_data/t10k-images-idx3-ubyte.gz
Extracting MNIST_data/t10k-labels-idx1-ubyte.gz
WARNING:tensorflow:TensorFlowLinearClassifier class is deprecated. Please consider using LinearClassifier as an alternative.
Traceback (most recent call last):
  File "mnist_weights.py", line 37, in <module>
    classifier.fit(mnist.train.images, mnist.train.labels)
  File "/Users/tcf/.virtualenvs/dnn/lib/python2.7/site-packages/tensorflow/contrib/learn/python/learn/estimators/base.py", line 458, in fit
    batch_size=batch_size or self.batch_size, monitors=monitors)
  File "/Users/tcf/.virtualenvs/dnn/lib/python2.7/site-packages/tensorflow/contrib/learn/python/learn/estimators/estimator.py", line 182, in fit
    monitors=monitors)
  File "/Users/tcf/.virtualenvs/dnn/lib/python2.7/site-packages/tensorflow/contrib/learn/python/learn/estimators/estimator.py", line 449, in _train_model
[...]

PR #3109, opened 2016-06-29
fixed ValueError in tensorflow/examples/skflow/mnist_weights.py

tcfuji:
Solution to #3083. 

I replaced `mnist.train.images` with `mnist.train.images.astype(np.float32)` and `mnist.train.labels` with `mnist.train.labels.astype(np.int)`.

### DIFF
--- a/tensorflow/examples/skflow/mnist_weights.py
+++ b/tensorflow/examples/skflow/mnist_weights.py
@@ -21,6 +21,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import numpy as np
 from sklearn import metrics
 import tensorflow as tf
 from tensorflow.examples.tutorials.mnist import input_data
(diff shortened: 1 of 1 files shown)


tensorflow/examples/skflow/mnist_weights.py
```diff
@@ -34,8 +35,10 @@
 
 classifier = learn.TensorFlowLinearClassifier(
     n_classes=10, batch_size=100, steps=1000, learning_rate=0.01)
-classifier.fit(mnist.train.images, mnist.train.labels)
-score = metrics.accuracy_score(mnist.test.labels, classifier.predict(mnist.test.images))
+X_train, y_train = mnist.train.images.astype(np.float32), mnist.train.labels.astype(np.int)
```
rmlarsen: Should this be handled in learn.datasets.load_dataset('mnist')?

martinwicke: That would be better. Is the cast necessary?

rmlarsen: From the bug #3083 it appears that the labels should be int32, int64 but are in fact uint8. Perhaps the right fix is for tflearn to seamlessly support uint8 labels?

martinwicke: This is an outdated example, it's in line to be updated, just has to wait its turn. If the new code cannot handle uint8s it should.

rmlarsen:
I'm closing this PR, as we seem to agree that this is not the right fix.

PR #3121, opened 2016-06-30
change boringssl resource from googlesource to github

Huarong:
Due to the network issue in China, the `https://boringssl.googlesource.com/boringssl` can not be downloaded while using Bazel. So I suggest to change it to a github mirror `https://github.com/google/boringssl.git`.


Huarong:
I signed it!

rmlarsen:

PR #3130, opened 2016-06-30
Remove mentions of `with_dependencies` from API docs

mrry:
We no longer (perhaps never did?) expose this function in our public API, so it shouldn't be mentioned in our public API docs.


mrry:

PR #3136, opened 2016-07-01
Fixed run_config example

terrytangyuan:
- Switched to use `DNNClassifier`
- Example runs forever if steps is not specified
- Updated copyright info


terrytangyuan:
Closing this since it's already changed from internal merge

PR #2998, opened 2016-06-22
Enable tf.tanh() for SparseTensor

siddharth-agrawal:
Enabled `tf.tanh()` for `SparseTensor`. Added tests and verified locally. This partially addresses #1828.

PR #3145, opened 2016-07-01
Add missing tf.tanh() tests for SparseTensor

siddharth-agrawal:
More information in comment [here](https://github.com/tensorflow/tensorflow/pull/2998#issuecomment-229743288). Tested locally.

### DIFF
--- a/tensorflow/python/kernel_tests/cwise_ops_test.py
+++ b/tensorflow/python/kernel_tests/cwise_ops_test.py
@@ -48,7 +48,9 @@ _INV = lambda x: ~x
 
 # TODO(zongheng): it'd be great to factor out this function and various random
 # SparseTensor gen funcs.
-def _sparsify(x, thresh=0.5, index_dtype=np.int64):
+def _sparsify(x, thresh, index_dtype=np.int64):
+  if thresh is None:
+    thresh = 0.5
   x[x < thresh] = 0
 
   non_zero = np.where(x)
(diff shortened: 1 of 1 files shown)


tensorflow/python/kernel_tests/cwise_ops_test.py
```diff
@@ -220,8 +222,10 @@ def testFloatBasic(self):
   def testFloatTanhEdge(self):
     x = np.arange(40, 40 + 6).reshape(6).astype(np.float32)
     self._compareBoth(x, np.tanh, tf.tanh)
+    self._compareBothSparse(x, np.tanh, tf.tanh)
     x = np.arange(-40, -40 + 6).reshape(6).astype(np.float32)
     self._compareBoth(x, np.tanh, tf.tanh)
+    self._compareBothSparse(x, np.tanh, tf.tanh, thresh=-40)
```
rmlarsen: With a threshold of -40 the sparsification will do nothing. Is that intentional?

siddharth-agrawal: Yes, I think the point of this test is to check numerical stability of the Tanh for large values of x, so I chose to not sparsify.

rmlarsen: @siddharth-agrawal I see. In that case I don't see any point in this change at all, since the dense and sparse versions call the same underlying tanh function, and there are already tests for the sparse version in the non-edge case.

siddharth-agrawal: @rmlarsen Yeah, that's true. Sorry for this PR.

PR #3164, opened 2016-07-03
Add Float64 support to CTC Decoder Ops

samjabrahams:
**PR incomplete! Have question about best way to get around one line**

Chipping away at ops listed in #1140

This PR adds 64-bit floating point support for CTCGreedyDecoder and provides tests for it. However, CTCDecoderBeamSearch doesn't have support yet, as there's [a line that needs to be changed](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/ctc_decoder_ops.cc#L293), and I need help with it. What would be the best way to dynamically select the correct `DataType` here? Currently `DT_FLOAT` is hard-coded in, but it should be `DT_DOUBLE` when the template type is `double`. Thanks!

### DIFF
--- a/tensorflow/core/kernels/ctc_decoder_ops.cc
+++ b/tensorflow/core/kernels/ctc_decoder_ops.cc
@@ -22,6 +22,7 @@ limitations under the License.
 #include "tensorflow/core/util/ctc/ctc_beam_search.h"
 #include "tensorflow/core/framework/op.h"
 #include "tensorflow/core/framework/op_kernel.h"
+#include "tensorflow/core/framework/register_types.h"
 #include "tensorflow/core/framework/types.h"
 #include "tensorflow/core/lib/core/status.h"
 #include "tensorflow/core/platform/logging.h"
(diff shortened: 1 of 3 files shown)


ibab:
`DataTypeToEnum<T>::value` from
https://github.com/tensorflow/tensorflow/blob/b8652a2efea3a04cbaa00b5210ba58a6c5854307/tensorflow/core/framework/types.h#L150 should work.

andrewharp:

ebrevdo:
Is there a good reason for having a float64  version of CTC decoders, and possibly even loss?  Can you ought point to a reference it or example that shows this improves numerical stability/convergence speed?

samjabrahams:
Honestly, there wasn't much thought in terms of the total benefit for adding double support for these specific Ops- it was pretty much a random draw from [this list](https://github.com/tensorflow/tensorflow/issues/1140#issuecomment-227012645) in #1140. You can follow the discussion there to see the context.

That said, I still have more work on this to implement and register a GPU kernel. If you think my efforts are better spent on something else, I'm happy to close this for now and move on to something more beneficial.

samjabrahams:
Pulling this off for now- doesn't look like it's necessary.

PR #3219, opened 2016-07-07
docker exec -it to start the interactive terminal

vivekjuneja:
As per the current documentation, 
$ docker run -it -p 8888:8888  gcr.io/tensorflow/tensorflow should be used to access the CLI. 
However, the previous command starts the Notebook App. The CLI is not available. To access the CLI, we need to access the container by creating a new interactive terminal to it. 
Hence, I added a second step using docker exec.

### DIFF
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -370,7 +370,9 @@ After Docker is installed, launch a Docker container with the TensorFlow binary
 image as follows.
 
 ```bash
-$ docker run -it -p 8888:8888 gcr.io/tensorflow/tensorflow
+$ docker run -d -p 8888:8888 --name="tensor" gcr.io/tensorflow/tensorflow
+
+$ docker exec -it tensor bash
 ```
 
 The option `-p 8888:8888` is used to publish the Docker container᾿s internal port to the host machine, in this case to ensure Jupyter notebook connection.
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -380,7 +380,7 @@ The format of the port mapping is `hostPort:containerPort`. You can specify any 
 To start a command-line interface:
 
 ```bash
-$ docker run -it -p 8888:8888 --name="tensor" --entrypoint=/bin/bash gcr.io/tensorflow/tensorflow
+$ docker run --rm -it -p 8888:8888 --name="tensor" --entrypoint=/bin/bash gcr.io/tensorflow/tensorflow
 ```
 
 
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -371,13 +371,18 @@ image as follows.
 
 ```bash
 $ docker run -d -p 8888:8888 --name="tensor" gcr.io/tensorflow/tensorflow
-
-$ docker exec -it tensor bash
 ```
 
 The option `-p 8888:8888` is used to publish the Docker container᾿s internal port to the host machine, in this case to ensure Jupyter notebook connection.
 
 The format of the port mapping is `hostPort:containerPort`. You can specify any valid port number for the host port but have to use `8888` for the container port portion.
+
+To start a command-line interface:
+
+```bash
+$ docker run -it -p 8888:8888 --name="tensor" --entrypoint=/bin/bash gcr.io/tensorflow/tensorflow
+```
+
 
 For NVidia GPU support install latest NVidia drivers and
 [nvidia-docker](https://github.com/NVIDIA/nvidia-docker).


vivekjuneja:
I signed it !

andrewharp:

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -370,7 +370,9 @@ After Docker is installed, launch a Docker container with the TensorFlow binary
 image as follows.
 
 ```bash
-$ docker run -it -p 8888:8888 gcr.io/tensorflow/tensorflow
+$ docker run -d -p 8888:8888 --name="tensor" gcr.io/tensorflow/tensorflow
+
+$ docker exec -it tensor bash
```
vrv: Maybe keep both sets of instructions:

"To start the Notebook App: 
$ docker run -it -p 8888:8888 gcr.io/tensorflow/tensorflow

To start a command-line interface:
$ docker run -it -p 8888:8888 --entrypoint /bin/bash gcr.io/tensorflow/tensorflow
"

(I think --entrypoint /bin/bash might be nicer in this case -- does that work for you?).

vivekjuneja: I agree. This one is nicer. 

```$ docker run -it -p 8888:8888 --entrypoint /bin/bash gcr.io/tensorflow/tensorflow```

Thanks for consideration.

Comment (vrv) [inline tensorflow/g3doc/get_started/os_setup.md]: Sounds good -- @craigcitro any thoughts on this?

Comment (craigcitro) [inline tensorflow/g3doc/get_started/os_setup.md]: So I think there are two questions here, if I'm reading correctly:

1. If we start the notebook as a server, should we use `-it` or `-d`? I agree that `-d` is nice, the only downside is that it means it's harder to find the logs. Maybe I'm more interested in reading the logs than other folks, so that's probably fine. 😉 

1. If we do want to start a CLI, how should we do it? I think there are actually *two* use cases: one is "I **just** want a bash prompt", and the other is "I want to get a bash prompt to my **already running** container". For the former, 

  ```
$ docker run -it -p 8888:8888 --entrypoint /bin/bash gcr.io/tensorflow/tensorflow
```
  seems nice; for the latter, I think we want to `docker exec --name=tensor bash`, right?

vivekjuneja: @craigcitro - You have raised good points. 

As per the documentation (https://www.tensorflow.org/versions/r0.9/get_started/os_setup.html#docker-installation), it is assumed that the developer would use tensorflow docker image to isolate their host environment from the tensorflow packages. 

It is better to scope the two docker invocations separately. 

1. If you would wish to run the notebook as a server, we use the ```-it```. This will avoid separate invocation of ```docker logs``` on the existing container. 

2. The need for CLI is ideally to isolate the host (local) environment from the tensorflow packages. The user may want to use the python installation, and the installed packages in the container to test out code. In that case, ```--entrypoint /bin/bash``` is more cleaner. 

@vrv  - what do you think ?

vrv: Sure, that sounds fine -- as long as these cases are documented well (why you would use one vs. the other), having both instructions seems helpful.  We don't want to embed all of docker documentation in TF, so we should link to Docker documentation where it makes sense

danmane:
@vivekjuneja it looks like this PR has stalled out - my impression is the ball is in your court to resolve the comments from code review. or do we need to take another look?

vivekjuneja:
@danmane - I have changed the documentation based on the feedback from @vrv

rmlarsen:
@vrv please review this.

rmlarsen:

vrv:
the branch has conflicts, so they need to be resolved first...  otherwise looks fine, but I don't use docker so I'm probably the wrong person to review this :)

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -370,13 +370,20 @@ After Docker is installed, launch a Docker container with the TensorFlow binary
...
 ```bash
-$ docker run -it -p 8888:8888 gcr.io/tensorflow/tensorflow
+$ docker run -d -p 8888:8888 --name="tensor" gcr.io/tensorflow/tensorflow
 ```
 
 The option `-p 8888:8888` is used to publish the Docker container᾿s internal port to the host machine, in this case to ensure Jupyter notebook connection.
 
 The format of the port mapping is `hostPort:containerPort`. You can specify any valid port number for the host port but have to use `8888` for the container port portion.
 
+To start a command-line interface:
```
vrv: maybe say "To start a command-line interface attaching to the container you created above:"

danmane:
@craigcitro perhaps you'd be kind enough to take a look :)

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -370,13 +370,20 @@ After Docker is installed, launch a Docker container with the TensorFlow binary
 image as follows.
 
 ```bash
-$ docker run -it -p 8888:8888 gcr.io/tensorflow/tensorflow
+$ docker run -d -p 8888:8888 --name="tensor" gcr.io/tensorflow/tensorflow
```
craigcitro: If we're going to leave it running in daemon mode, we need to include a sample command to stop it. And probably one to remove it?

tensorflow/g3doc/get_started/os_setup.md
craigcitro: Actually, this is an *alternate* approach, right? This is starting a new container.

We also probably want a sample in line with what @vrv was expecting here.

vrv:
I'm closing this due to lack of response but comment and we'll reopen after you've addressed these changes.  Thanks!

PR #3222, opened 2016-07-07
instructions for saving progress using the udacity docker image

hrldcpr:
People not used to Docker might be surprised when their "saved" changes suddenly disappear.

Additionally we could add instructions for mounting a local directory at `/notebooks`, but first you have to copy the initial contents into your local directory, so it's a bit complicated.

### DIFF
--- a/tensorflow/examples/udacity/README.md
+++ b/tensorflow/examples/udacity/README.md
@@ -19,6 +19,21 @@ On mac, find the virtual machine's IP using:
 
 Then go to: http://IP:8888 (likely http://192.168.99.100:8888)
 
+Saving Your Progress
+--------------------
+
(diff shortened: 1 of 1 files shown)


andrewharp:

pcasaretto:
[This commit](https://github.com/tensorflow/tensorflow/commit/8ebb0c9b0025fc5cf327a5152ae09df4b6df19b7) commit removed the `--rm` flag making these somewhat confusing.

vincentvanhoucke:
@pcasaretto indeed. PR #3303 removes this section.

hrldcpr:
Haha, looks like we were all pouncing on that `--rm` on the same day, from different angles.

PR #3229, opened 2016-07-07
maximize function added to Optimizer

Mazecreator:
I have added a Maximize() function to the Optimizer.  This makes efforts where maximizing a function straight forward to implement.

I have used the "minimize()" function as a model, created "maximize()" and simply inverted the sign to the cost of the "compute_gradients()" function.

This is my first attempt at an update on Github, please let me know if there is an error in the request.

### DIFF
--- a/tensorflow/python/training/optimizer.py
+++ b/tensorflow/python/training/optimizer.py
@@ -84,13 +84,14 @@ class Optimizer(object):
   @@__init__
 
   @@minimize
+  @@maximize
   @@compute_gradients
   @@apply_gradients
 
   ### Gating Gradients
 
-  Both `minimize()` and `compute_gradients()` accept a `gate_gradient` argument
-  that controls the degree of parallelism during the application of the
+  The `minimize()`, `maximize()` and `compute_gradients()` accept a `gate_gradient` 
+  argument that controls the degree of parallelism during the application of the
   gradients.
 
   The possible values are: `GATE_NONE`, `GATE_OP`, and `GATE_GRAPH`.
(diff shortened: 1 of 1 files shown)


Mazecreator:
I signed it!

girving:
I'm not sure this is worth adding.  Can't you just pass `-value` to `minimize`?  Having a separate function is a nontrivial cognitive burden since `Optimizer` is part of a class hierarchy.  For example, your current implementation is wrong, since if someone overrides `minimize` to do something new `maximize` will no longer do the right thing for that derived class.

Mazecreator:
I am sure I can now us (-COST), but I am just trying to make it easier for people in the future and have code more readable.  When maximizing, it really isn't a COST that is being maximized so that isn't intuitive and certainly -COST was not clear to me and not likely clear to many readers of future code so it seems like Maximize() would simply round of Optimization.  You can then minimize a cost or maximize a value function.  It seems like this will help make the python wrapper easier to use.

I did think about changes to "minimize()" in the future, so in my implementation, I didn't call "minimize()" from the "maximize()" function but use it as a reference.  "maximize" only depends upon the "compute_gradients" & "apply_gradients" functions.  I think this is a simple addition that will have benefit in the long run.

I can write a test based upon "testBasic" in "optimizer_test.py" if that is required for a core change.  I have done a simple bench test to test maximize() within my local repository.  I can call it "testBasicMaximize" but I am unsure how to get it to run during the build test phase.

girving:
It is natural to think it would help, but I do think this would cause more confusion for users down the road rather than less, in particular for users who write their own optimizers.  As I said above, your current implementation of `maximize` is wrong in the presence of derived classes.  If you fix it to call `minimize` the situation improves, but not enough: the result is still fragile if someone writes an optimizer and decides to override both minimize and maximize, then a third person overrides just minimize, etc.  This fears are based on experience: it's already surprisingly difficult to robustly inherit from optimizers in nonstandard ways, so I'm leery of making it worse in order to save 1 character.

vrv:
I would prefer not to add another API to this class too.  Users should be able to easily understand that maximizing a value is the same as minimizing the negative value (or even the reciprocal of the value).  Keeping our core classes simple is going to be a better benefit to everyone in the long term.  You are free to develop a higher-level library with this functionality and API though!

Mazecreator:
I have no problem moving it to a contribute section, but I would still like to keep the same context:
        sgd_op = tf.train.GradientDescentOptimizer(3.0).maximize(value, global_step, [var0, var1])

I will try to inherit Optimizer and add the maximize() function.  Is that acceptable?

girving:
You're free to do that in your own code, but I would recommend against it.  If you do that, your code will break if you want to switch optimizers.  If you really do want to replace one character with a whole function, I'd recommend:

    def maximize(optimizer, value, ...):
      return optimizer.minimize(-value, ...)

PR #3231, opened 2016-07-08
Fix doc typo from `Tensor.run` to `Tensor.eval`

samjabrahams:
Quick documentation in session.py to fix to point to correct function name and permalink id.


caisq:

caisq:
Looks good to me.

Issue #3213, opened 2016-07-07
Accompany TF_LoadLibrary with TF_DeleteLibrary


IvanUkhov:
The C API provides the [following function](https://github.com/tensorflow/tensorflow/blob/r0.9/tensorflow/core/public/tensor_c_api.h#L357) for loading libraries:

```c
extern TF_Library* TF_LoadLibrary(const char* library_filename, TF_Status* status);
```

The documentation says that “the caller owns the library handle” returned by the function. Unlike other functions allocating memory and returning pointers, `TF_LoadLibrary` doesn’t have a counterpart for properly freeing the memory. Note that it’s not about unloading the library but about deallocating the `TF_Library` struct; although, a function for unloading the library would also be nice. Thanks!

Regards,
Ivan

aselle:
I suppose it is possible that the caller cannot know what free-store the memory of TF_LoadLibrary was allocated from. Unloading libraries is generally unreliable as dlclose isn't really required by POSIX to do much except invalidate the handle.

PR #3235, opened 2016-07-08
Accompany TF_LoadLibrary with TF_DeleteLibrary

IvanUkhov:
The pull request addresses issue #3213. Please let me know
- if I should add a comment to the new function, and
- if I should also change [tf_session.i](https://github.com/tensorflow/tensorflow/blob/r0.9/tensorflow/python/client/tf_session.i#L239) or [load_library.py](https://github.com/tensorflow/tensorflow/blob/r0.9/tensorflow/python/framework/load_library.py#L60).

Thanks!

Regards,
Ivan

### DIFF
--- a/tensorflow/core/client/tensor_c_api.cc
+++ b/tensorflow/core/client/tensor_c_api.cc
@@ -568,6 +568,10 @@ TF_Library* TF_LoadLibrary(const char* library_filename, TF_Status* status) {
   return lib_handle;
 }
 
+void TF_DeleteLibrary(TF_Library* library) {
+  delete library;
+}
+
 TF_Buffer TF_GetOpList(TF_Library* lib_handle) { return lib_handle->op_list; }
 
 }  // end extern "C"
--- a/tensorflow/core/public/tensor_c_api.h
+++ b/tensorflow/core/public/tensor_c_api.h
@@ -361,6 +361,8 @@ typedef struct TF_Library TF_Library;
 extern TF_Library* TF_LoadLibrary(const char* library_filename,
                                   TF_Status* status);
 
+extern void TF_DeleteLibrary(TF_Library*);
+
 // Get the OpList of OpDefs defined in the library pointed by lib_handle.
 //
 // Returns a TF_Buffer. The memory pointed to by the result is owned by


andrewharp:

keveman:
@IvanUkhov Thanks a lot for the PR. Although your `TF_DeleteLibrary` frees up the memory owned by a `TF_Library` handle, I am not sure how useful that is. The reason that we didn't bother adding this function in the first place is that we don't know of a reliable way to unload the library from the process' address space. If you know of an implementation for unloading the library on linux and mac, that would make this a much stronger PR.

IvanUkhov:
@keveman, please have a look at issue #3213 and the comment there. This function is not for unloading libraries; it’s for freeing memory. Freeing memory is the caller’s responsibility, but the caller doesn’t know what it was allocated and, hence, cannot do it properly. For unloading libraries, one might add something like `TF_UnloadLibrary` in the future.

keveman:
@IvanUkhov, sure, I read your comment. What I'm wondering about is the utility of just deleting the handle without unloading the library.

IvanUkhov:
@keveman, I see. Thanks for the clarification! I think it depends on what `TF_DeleteLibrary` is supposed to do. If we agree that its sole purpose is to dispose `TF_Library` then the current implementation of the function satiates the purpose. It eliminates a memory leakage, and it makes one more confident about the usage of the API as one knows exactly what to do this the pointer given by `TF_LoadLibrary`. Imagine what people calling the function from other languages think when they are given a pointer and told that they own it and the underlaying data. If, however, `TF_DeleteLibrary` is also responsible for unloading the library then, I agree, it’s better to do not have any function at all than a half-working one.

It might be interesting to note that there is a certain analogy between `TF_DeleteSession` and `TF_DeleteLibrary` and between `TF_CloseSession` and `TF_UnloadLibrary` (potential).

dauba-dauba:
Is dlclose() not sufficient for a TF_UnloadLibrary implementation?

keveman:
@dauba-dauba Reading the documentation [here](http://pubs.opengroup.org/onlinepubs/009695399/functions/dlclose.html), it doesn't seem like there are any guarantees about the shared library being unmapped from the memory space. Also, with a multithreaded application like TensorFlow, I don't know how we can reason about `dlclose` when there is another thread executing a kernel loaded from the shared library. @IvanUkhov, while I share your sentiment about avoiding memory leaks, I can't imagine people loading hundreds of shared libraries via `TF_LoadLibrary` and leaking the handles. If we get into a situation like that, we can definitely revisit this, but until I see some compelling reason, I would rather have libraries loaded and remain in memory.

IvanUkhov:
@keveman, I phrased it poorly, but I was actually referring to confidence. Leakage is an unpleasant but indeed minor issue. In any case, I appreciate your time and respect your opinion. Please feel free to close both the issue and the pull request. Thank you.

dauba-dauba:
IIRC, the spec reads that way because some implementations on now-rarely-used flavors of Unix (e.g. AIX, or maybe HP-UX) never removed libraries from the process address space. Both Linux and OS X man pages for dlclose() state that it decrements a ref count and, if zero, unloads the library.

That said, if semantically there's no good way of tracking symbol references into the loaded library, then I agree it's easier to just declare that such libraries are "load-only".

keveman:
@dauba-dauba, right, even if the OS removed it from the address space, it would be hard to track currently running threads that have ops and kernels from the library.
@IvanUkhov Thanks a lot for your time. Expecting more PRs in the future :)

PR #3237, opened 2016-07-08
Fix issue 3186: buffer alignment needs to be 8-byte for uint64 vals 

### LINKED ISSUE #3186: Insufficient alignment in u_ union in tensorflow/core/lib/gtl/inlined_vector.h
### Environment info

Operating System: Linux

Installed version of CUDA and cuDNN: None

If installed from sources, provide the commit hash: aa2cacd6627ffb296bedc910c957a0fd4a2f957f
### Steps to reproduce
1. Find an architecture with 32 bit pointers but strict alignment requirements for 64 bit
2. Compile and run "benchmark" from tensorflow/contrib/makefile
3. Get bus error from misaligned pointer
### What have you tried?
1. Fixing the issue by increasing the u_ alignment.  This works.

In tensorflow/core/lib/gtl/inlined_vector.h, the data buf is aligned by adding a pointer.  However, it is cast to other types, including uint64_t.  On architectures where pointers are 32 bits, this cast causes errors.

The fix is to add another component to the union that will force alignment to the largest size, such as uint64_t.

For example:

diff --git a/tensorflow/core/lib/gtl/inlined_vector.h b/tensorflow/core/lib/gtl/inlined_vector.h
index e8fe66c..518b421 100644
--- a/tensorflow/core/lib/gtl/inlined_vector.h
+++ b/tensorflow/core/lib/gtl/inlined_vector.h
@@ -276,6 +276,8 @@ class InlinedVector {
     unsigned char data[kSize];
     // Force data to be aligned enough for a pointer.
     T\* unused_aligner;
-    // EJP: force 8 byte alignment, as pointers could be 4-byte but data elements 8-byte?
-    uint64_t unused_64;
  } u_;

You won't see this issue on architectures where pointers are the size of the largest type (aarch64, x86-64) or where a 32-bit architecture loads 64-bit things but can do so at arbitrary 32-bit boundaries (x86, armv7 with misaligned pointers enabled).  

But it might improve performance to have 64-bit values on 64-bit boundaries.

eplondke:
This Closes #3186 

Add uint64_t to the u_ union to ensure 8-byte alignment, pointers are not 8-byte aligned on 32 bit architectures.

### DIFF
--- a/tensorflow/core/lib/gtl/inlined_vector.h
+++ b/tensorflow/core/lib/gtl/inlined_vector.h
@@ -276,6 +276,9 @@ class InlinedVector {
     unsigned char data[kSize];
     // Force data to be aligned enough for a pointer.
     T* unused_aligner;
+    // Some values placed here are uint64, which may require 8 byte alignment
+    // but on 32 bit archtectures, pointers are insufficiently aligned
+    uint64_t unused_aligner_64;
   } u_;
 
   inline void InitRep() { u_.data[kSize - 1] = 0; }


vrv:
(we can only take a look if the CLA is signed).

eplondke:
Working on CLA.

vrv:
any updates on CLA?

rmlarsen:
@eplondke any update?

rmlarsen:
Closing due to inactivity.

PR #3248, opened 2016-07-09
NUM_CHANNELS instead of '1' for consistency

marcomarchesi:
Even if the MNIST dataset is made by one-channels images, NUM_CHANNELS variable is used for creating the model. To be consistent and for generality, shall we use NUM_CHANNELS also for extracting data from images into the 4D tensor.


marcomarchesi:
I signed it!

vrv:

PR #3250, opened 2016-07-09
Remove support for CuDNN v2 runtime/compile from TensorFlow.

vrv:
Major changes in the APIs between CuDNN versions make it
hard to update to newer versions of the library as they
come out every few months (as can be seen by the lines
removed by this commit).  As CuDNN APIs become more stable
we can likely leave support for more versions, but V2 was
particularly problematic, and versions V3+ are a large
enough improvement over V2 that we don't need to support V2
anymore.

When V6 comes out, we'll re-evaluate the list of supported
versions and see whether to drop V3 at that time.

### DIFF
--- a/tensorflow/stream_executor/cuda/cuda_dnn.cc
+++ b/tensorflow/stream_executor/cuda/cuda_dnn.cc
@@ -138,13 +138,6 @@ size_t cudnnGetVersion() {
   return callable();
 }
 
-// Returns whether the currently loaded cuDNN version is R2.
-bool IsCudnnR2() {
-  static auto version = cudnnGetVersion();
-  DCHECK_GE(version, 2000);
-  return version < 3000;
-}
-
 #define PERFTOOLS_GPUTOOLS_CUDNN_WRAP(__name)                        \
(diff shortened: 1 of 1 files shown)


tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -138,13 +138,6 @@ size_t cudnnGetVersion() {
   return callable();
 }
 
-// Returns whether the currently loaded cuDNN version is R2.
-bool IsCudnnR2() {
```
zheng-xq: Also those:

#define CUDNN_DNN_ROUTINE_EACH_R2(__macro)                \
  __macro(cudnnAddTensor_v2)                              \
  __macro(cudnnConvolutionBackwardData_v2)                \
  __macro(cudnnConvolutionBackwardFilter_v2)
#else

vrv: Done, moved these 3 into the define above.

zheng-xq:
LGTM

PR #3269, opened 2016-07-11
Add cuda_configure repository rule to autodetect cuda.

### LINKED ISSUE #2873: Perform CUDA autoconfiguration in Skylark
Moving the CUDA autoconfiguration to a Skylark repository rule, similar to Bazel's [`cc_configure`](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl), will fix many pain points and build issues that have been reported. This will also eventually allow projects such as Serving and Magenta to include TensorFlow as a Bazel workspace dependency rather than a Git submodule.

Related to this is moving Python detection to a Skylark repository rule (see #1404).

Internal tracking bug: b/29006900

davidzchen:
Fixes #2873


third_party/gpus/crosstool/CROSSTOOL.tpl
```diff
@@ -121,6 +121,9 @@ toolchain {
   # linker_flag: "-Wl,--warn-execstack"
   # linker_flag: "-Wl,--detect-odr-violations"
 
+  # Include directory for cuda headers.
+  cxx_builtin_include_directory: "/usr/local/cuda%{cuda_version}/include"
```
davidzchen: @damienmg is this sufficient? If not, I will still need to figure out what is the `nvcc` equivalent of `cc -E -xc++ - -v` to add an equivalent of [`_get_cxx_inc_directories`](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl#L125).

third_party/gpus/cuda_configure.bzl
```diff
@@ -0,0 +1,248 @@
...
+
+
+_DEFAULT_CUDA_TOOLKIT_PATH = "/usr/local/cuda"
+_DEFAULT_CUDNN_INSTALL_PATH = "/usr/local/cuda"
+
+
+def _cuda_toolkit_path(repository_ctx):
+  """Finds the cuda toolkit directory."""
+  cuda_toolkit_path = _DEFAULT_CUDA_TOOLKIT_PATH
+  if "CUDA_TOOLKIT_PATH" in repository_ctx.os.environ:
```
davidzchen: Params that were previously configured using the `./configure` script prompts are now being passed using environment variables. For example:

```sh
CUDA_VERSION=7.5 bazel test --config=cuda //tensorflow/...
```

The problem with this is that changing the environment variables would not automatically trigger a re-fetch of the `@local_config_cuda` external repository.

Is there a way to pass these values using command line flags and apply them using Bazel's configuration mechanism? I'm not sure how to do this with `config_setting` since it seems to only match specific values.

third_party/gpus/cuda/BUILD.tpl
```diff
@@ -0,0 +1,235 @@
...
+    visibility = ["//visibility:public"],
+)
+
+
+# TODO(opensource): for now, we have to invoke the cuda_config.sh manually in the source tree.
+# This rule checks if Cuda libraries in the source tree has been properly configured.
+# The output list makes bazel runs this rule first if the Cuda files are missing.
+# This gives us an opportunity to check and print a meaningful error message.
+# But we will need to create the output file list to make bazel happy in a successful run.
+#genrule(
```
davidzchen: TODO: Still need to implement a replacement for this.

davidzchen:
Building using the system's default version of cuda libraries (running `bazel test --config=cuda //tensorflow/...` without `CUDA_VERSION=7.5) causes all the Python targets to raise the following import error:

```
❯❯❯ bazel test --config=cuda //tensorflow/python:summary_writer_test --test_output=errors
WARNING: /usr/local/google/home/dzc/Projects/tensorflow/tensorflow/util/python/BUILD:11:16: in includes attribute of cc_library rule //util/python:python_headers: 'python_include' resolves to 'util/python/python_include' not in 'third_party'. This will be an error in the future
INFO: Analysed target //tensorflow/python:summary_writer_test (0 packages loaded).
INFO: Found 1 test target...
FAIL: //tensorflow/python:summary_writer_test (see /usr/local/google/home/dzc/.cache/bazel/_bazel_dzc/97010c3bd7554ebe07410261582c4d5e/execroot/tensorflow/bazel-out/local_linux-fastbuild/testlogs/tensorflow/python/summary_writer_test/test.log)
INFO: From Testing //tensorflow/python:summary_writer_test:
==================== Test output for //tensorflow/python:summary_writer_test:
  File "/usr/local/google/home/dzc/.cache/bazel/_bazel_dzc/97010c3bd7554ebe07410261582c4d5e/execroot/tensorflow/bazel-out/local_linux-fastbuild/bin/tensorflow/python/summary_writer_test.runfiles/org_tensorflow/tensorflow/python/training/summary_writer_test.py", line 26, in <module>
    import tensorflow as tf
  File "/usr/local/google/home/dzc/.cache/bazel/_bazel_dzc/97010c3bd7554ebe07410261582c4d5e/execroot/tensorflow/bazel-out/local_linux-fastbuild/bin/tensorflow/python/summary_writer_test.runfiles/org_tensorflow/tensorflow/__init__.py", line 23, in <module>
    from tensorflow.python import *
  File "/usr/local/google/home/dzc/.cache/bazel/_bazel_dzc/97010c3bd7554ebe07410261582c4d5e/execroot/tensorflow/bazel-out/local_linux-fastbuild/bin/tensorflow/python/summary_writer_test.runfiles/org_tensorflow/tensorflow/python/__init__.py", line 48, in <module>
    from tensorflow.python import pywrap_tensorflow
  File "/usr/local/google/home/dzc/.cache/bazel/_bazel_dzc/97010c3bd7554ebe07410261582c4d5e/execroot/tensorflow/bazel-out/local_linux-fastbuild/bin/tensorflow/python/summary_writer_test.runfiles/org_tensorflow/tensorflow/python/pywrap_tensorflow.py", line 28, in <module>
    _pywrap_tensorflow = swig_import_helper()
  File "/usr/local/google/home/dzc/.cache/bazel/_bazel_dzc/97010c3bd7554ebe07410261582c4d5e/execroot/tensorflow/bazel-out/local_linux-fastbuild/bin/tensorflow/python/summary_writer_test.runfiles/org_tensorflow/tensorflow/python/pywrap_tensorflow.py", line 24, in swig_import_helper
    _mod = imp.load_module('_pywrap_tensorflow', fp, pathname, description)
ImportError: libcudart.so.7.5: cannot open shared object file: No such file or directory
Target //tensorflow/python:summary_writer_test up-to-date:
  bazel-bin/tensorflow/python/summary_writer_test
INFO: Elapsed time: 0.447s, Critical Path: 0.10s
INFO: Build completed, 1 test FAILED, 2 total actions
//tensorflow/python:summary_writer_test                                  FAILED in 0.1s
  /usr/local/google/home/dzc/.cache/bazel/_bazel_dzc/97010c3bd7554ebe07410261582c4d5e/execroot/tensorflow/bazel-out/local_linux-fastbuild/testlogs/tensorflow/python/summary_writer_test/test.log

Executed 1 out of 1 test: 1 fails locally.
```

As expected, the `local_config_cuda/cuda/lib64` directory only contains the symlinks to the `.so`'s without the version suffix (`libcudart.so` rather than `libcudart.so.7.5`). However, it appears that the SWIG wrapper still ends up looking for the versioned libraries. I'm not familiar with how SWIG or `imp` works, but the string `"7\.5"` does not appear any of the files under the files generated by `cuda_configure()` under `bazel-tensorflow/external/local_config_cuda`. Is the cuda version getting hard-coded somewhere else, causing the SWIG wrapper to look for the versioned library instead of the unversioned one?

third_party/gpus/crosstool/CROSSTOOL.tpl
damienmg: In think that should be enough. Using _get_cxx_inc_directories is definitely safer.

third_party/gpus/cuda_configure.bzl
damienmg: This is WAI. I need to add an explicit option to bazel to invalidate. Maybe run bazel clean --expunge for now on ./configure script?

WORKSPACE
```diff
@@ -18,6 +18,9 @@ workspace(name = "org_tensorflow")
 load("//tensorflow:workspace.bzl", "tf_workspace")
 tf_workspace()
 
+load("//third_party/gpus:cuda_configure.bzl", "cuda_configure")
```
damienmg: Shouldn't this be at the begining of the workspace file?

Maybe you actually want to add the cuda_configure to the tf_workspace macros for enabling it for ppl depending on it (like serving).

configure
```diff
@@ -68,251 +68,7 @@ echo "$SWIG_PATH" > tensorflow/tools/swig/swig_path
 # Invoke python_config and set up symlinks to python includes
 (./util/python/python_config.sh --setup "$PYTHON_BIN_PATH";) || exit -1
 
-## Set up Cuda-related environment settings
-
-while [ "$TF_NEED_CUDA" == "" ]; do
```
damienmg: We should keep that user interface, and provide the good export when running bazel fetch. This way, the configure script interface will stay the same.

tensorflow/stream_executor/cuda/cuda_fft.h
```diff
@@ -23,7 +23,7 @@ limitations under the License.
 #include "tensorflow/stream_executor/fft.h"
 #include "tensorflow/stream_executor/platform/port.h"
 #include "tensorflow/stream_executor/plugin_registry.h"
-#include "third_party/gpus/cuda/include/cufft.h"
+#include "external/local_config_cuda/cuda/include/cufft.h"
```
damienmg: we should not need the external/local_config_cuda prefix, this is more stable

tensorflow/stream_executor/dso_loader.cc
```diff
@@ -33,54 +33,63 @@ limitations under the License.
...
-// TensorFlow OSS configure uses the following lines to configure versions. For
-// any modifications of the format, please make sure the script still works.
-string GetCudaVersion() { return ""; }
-string GetCudnnVersion() { return ""; }
-
 /* static */ port::Status DsoLoader::GetCublasDsoHandle(void** dso_handle) {
-  return GetDsoHandle(FindDsoPath(tensorflow::internal::FormatLibraryFileName("cublas", GetCudaVersion()),
-                                  GetCudaLibraryDirPath()),
-                      dso_handle);
+  return GetDsoHandle(
```
damienmg: What is that change?

third_party/gpus/cuda/BUILD.tpl
damienmg: Does that still needs to exists? If you depends on cuda library then the cuda_configure should just write a build file that have a genrule that fails.

damienmg:
Ok this mostly looks good, I added some comments.

@davidzchen: the tests are all broken, do you know why?

configure
davidzchen: One thing I noticed is that after running `CUDA_VERSION=7.5 bazel fetch --config=cuda //...` (which I put in the `./configure` script), the `local_config_cuda` external repository does not show up in `bazel-tensorflow/external` until after building certain targets.

Further, I found that I need to add `CUDA_VERSION=7.5` to the `bazel build` invocation in order for the `@local_config_cuda` to be configured with the `CUDA_VERSION=7.5`, which seems to suggest that `@local_config_cuda` wasn't fetched by `bazel fetch //...`. Is this a bug or is the `bazel fetch` invocation incorrect here?

damienmg: This sounds like a bug to me. I'll investigate it as soon as I am done with sheriffing today.

tensorflow/stream_executor/dso_loader.cc
davidzchen: Made some style fixes since I have changes in this file anyway. I can revert these changes if need be.

tensorflow/stream_executor/cuda/cuda_fft.h
davidzchen: Done.

davidzchen:
@damienmg The tests are failing due to the following error:

```
ERROR: com.google.devtools.build.lib.packages.BuildFileContainsErrorsException: error loading package '': Encountered error while reading extension file 'cuda/build_defs.bzl': no such package '@local_config_cuda//cuda': Traceback (most recent call last):
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 219
		_cuda_toolkit_path(repository_ctx)
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 72, in _cuda_toolkit_path
		fail("Cannot find cuda toolkit path.")
Cannot find cuda toolkit path.
```

I still need to figure out why the CI machine cannot find cuda.

damienmg:
@jendap might knows more about the CI setup itself.

configure
damienmg: So has seen on another case this local_config_cuda should be fetched but is not visible unless we build a dep of it. I'll try to download your repo later today to do more test on it.

damienmg:
So the PR LGTM, just needs to fix the test and figure out the problem with fetching.

davidzchen:
Here is the current breakdown of test failures. The good news is that the vast majority of tests are passing. Curiously, there are some additional test failures when the we are configuring using `CUDA_VERSION=7.5` than when the autoconf is configured to use the system default cuda version (i.e. the non-versioned files), and some of these appear to be `missing dependency declarations` errors. I am still investigating these.

## Without `CUDA_VERSION=7.5` set

Bazel invocation:

```sh
bazel fetch //...
bazel test --config=cuda //tensorflow/...
```

6 / 624 tests failing:

```
//tensorflow/contrib/distributions:beta_test
//tensorflow/contrib/distributions:dirichlet_test
//tensorflow/contrib/distributions:gamma_test
//tensorflow/contrib/distributions:inverse_gamma_test
//tensorflow/contrib/distributions:mvn_test
//tensorflow/examples/skflow:examples_test
```

## With `CUDA_VERSION=7.5` set

Bazel invocation:

```sh
CUDA_VERSION=7.5 bazel fetch //...
bazel test --config=cuda //tensorflow/...
```

13 / 624 tests failing:

```
//tensorflow/core:debug_debug_gateway_test (TIMEOUT)
//tensorflow/contrib/distributions:beta_test
//tensorflow/contrib/distributions:dirichlet_test
//tensorflow/contrib/distributions:gamma_test
//tensorflow/contrib/distributions:inverse_gamma_test
//tensorflow/contrib/distributions:mvn_test
//tensorflow/core:common_runtime_gpu_gpu_bfc_allocator_test
//tensorflow/core:common_runtime_gpu_gpu_debug_allocator_test
//tensorflow/core:common_runtime_gpu_gpu_event_mgr_test
//tensorflow/core:common_runtime_gpu_gpu_tracer_test
//tensorflow/core:common_runtime_gpu_pool_allocator_test
//tensorflow/examples/skflow:examples_test
//tensorflow/g3doc/how_tos/adding_an_op:cuda_op_test
```

configure
damienmg: So I am testing right now.

First issue: it wants to have CUDA, I think this should be made so we can avoid having it. I don't have CUDA on my machine, how can I avoid having it?

Second issue: I don't have the interface to select cuda from ./configure, it should set the corresponding env variable. I install cuda and it's on the path, it stills need to set the environment variable. Also it wants cudNN which I don't have on my machine so I am stopping there right now.

davidzchen: Working on fixing these two issues.

By the way, is there a reason why the `fetch` command doesn't create any Bazel output dir symlinks? Seems that it might be useful to create the `bazel-<workspace-name>` symlink at least so that we can look at the external repositories that were fetched.

third_party/gpus/cuda_configure.bzl
davidzchen: I have added the clean command to the configure script.

Is it safe to use `--expunge_async` and then immediately run the fetch command? That is, does `--expunge_async` move/rename the root of the output directory and deletes asynchronously so that any new fetch/build commands that are run while the delete is still happening will be run in a clean output dir?

third_party/gpus/crosstool/CROSSTOOL.tpl
davidzchen: Understood. Thanks.

third_party/gpus/cuda/BUILD.tpl
davidzchen: It makes sense why we might want to remove this, but what do you mean by having `cuda_configure` "write a build file that has a genrule that fails?"

configure
damienmg: +Kristina Chodorow <kchodorow@google.com> to answer that question.

third_party/gpus/cuda/BUILD.tpl
damienmg: If cuda configuration is off, you just write a genrule cuda_check that does "exit 1" and if cuda is on, it does exit 0

davidzchen: I'm not completely following. This is a sanity check to determine whether the cuda files are set up correctly. In the context of this change, we would want a similar sanity check to determine whether `@local_config_cuda` is reasonable, which, come to think of it, might be more suitable implemented in `cuda_configure.bzl` rather than as a build target. WDYT?

damienmg: I was always considering this as a way to warn the user about cuda
configuration being unavailable. If cuda is deactivating, TF should fails
with a nice message when trying to build for cuda, right?

damienmg:
Is this stalled? Maybe try to get some help from people from the TensorFlow team

vrv:
Try building with --config=cuda -c opt  ?

configure
davidzchen: +cc @kchodorow

third_party/gpus/cuda_configure.bzl
davidzchen: Ping?

davidzchen:
Apologies for the wait. I have been busy with some work for my team these past few days. I'm working on this right now.

third_party/gpus/cuda_configure.bzl
damienmg: Sorry yes it is better not to use the async version.

Grrr. It is really hard to follow a code-review on that stuff.

configure
```diff
@@ -107,7 +109,6 @@ while true; do
   # Retry
 done
 
-
 # Find out where the CUDA toolkit is installed
 OSNAME=`uname -s`
```
damienmg: Ultimately we should simplify the part below so that only the user interaction remains in that script.

tensorflow/core/platform/default/build_config/BUILD
```diff
@@ -91,20 +91,20 @@ filegroup(
...
-        "//third_party/gpus/cuda:{}".format(cuda_library_path("cudart")),
+        "@local_config_cuda//cuda:{}".format(cuda_library_path("cudart")),
     ],
     linkopts = select({
-        "//third_party/gpus/cuda:darwin": [
-            "-Wl,-rpath,third_party/gpus/cuda/lib",
-            "-Wl,-rpath,third_party/gpus/cuda/extras/CUPTI/lib",
+        "@local_config_cuda//cuda:darwin": [
+            "-Wl,-rpath,external/local_config_cuda/cuda/lib",
+            "-Wl,-rpath,external/local_config_cuda/cuda/extras/CUPTI/lib",
```
damienmg: Does that work? With the deepexecroot change it will break, can you try ../local_config_cuda instead?

damienmg:
I added a couple of other comment. This change LGTM if it passes the test suite.

third_party/gpus/cuda_configure.bzl
davidzchen: Yeah, if GitHub had the ability to mark whether a comment is resolved, it would be much better. The way it tries to automatically hide "outdated" comments makes the review pretty confusing. :/

third_party/gpus/cuda/BUILD.tpl
davidzchen: Actually, it seems that even with the build configured with GPU support turned off, `tensorflow/platform/default/build_config:cuda` still expects the CUDA toolkit files to be present. I have added some code to create dummy CUDA toolkit files in `cuda_configure` for the codepath where CUDA support is disabled, which allowed the sanity tests to pass.

configure
davidzchen: Agreed. I'll do that in a follow-up change.

tensorflow/core/platform/default/build_config/BUILD
davidzchen: Done.

davidzchen:
Making good progress. All of the CI build are passing except for the Linux GPU PIP build.

@jendap - The Linux GPU PIP build is failing due to the following:

```
ERROR: com.google.devtools.build.lib.packages.BuildFileContainsErrorsException: error loading package '': Encountered error while reading extension file 'cuda/build_defs.bzl': no such package '@local_config_cuda//cuda': Traceback (most recent call last):
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 381
		_create_cuda_repository(repository_ctx)
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 337, in _create_cuda_repository
		repository_ctx.symlink(cudnn_header_dir + "/cudnn.h", "cu...")
java.io.IOException: Could not create symlink from /usr/include/cudnn.h to /opt/jenkins/workspace/tensorflow-pull-requests-gpu_pip/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/local_config_cuda/cuda/include/cudnn.h: /opt/jenkins/workspace/tensorflow-pull-requests-gpu_pip/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/local_config_cuda/cuda/include/cudnn.h (Permission denied).
Configuration finished
```

It seems that there is a permission error when `cuda_configure` attempts to create the symlinks for the CUDA toolkit files. Do we need to make some changes to the CI scripts to get this to work?

gunan:
Curious, the directory is owned by the jenkins user. however, there is already a cuda.h file in that folder, so I am guessing that is the reason for header symlink failing to be created.

gunan:
To me, it looks like the error is thrown during the configure script.
Is it possible one of your changes is causing an unintended side effect?

martinwicke:
If there's already a cuda.h, could it be a caching/previous build issue? The GPU build changed, so I'll run this again.

Jenkins, test this please.

gunan:
It failed the same way, interesting.
David, could you check your changes to the configure script?

martinwicke:
As @terrytangyuan has pointed out, this seems to be a pervasive issue, not related to any particular PR. I suspect we have to change the blocking presubmits in the branch setup. Let me see.

martinwicke:
Jenkins, test this please

davidzchen:
That is interesting. Before the `configure` script runs the `bazel fetch` command (where the error is happening), it runs `bazel clean --expunge`, which should have deleted all of the files Bazel generated for the workspace. Is it possible that not all of the files were deleted?

martinwicke:
Looking at the Linux-GPU failure now, it does look different. Can you take a look?

davidzchen:
I see the `no such package '@local_config_cuda//crosstool': BUILD file not found on package path.` error at the end of the log, but in the middle, we still have the permissions error:

```
ERROR: com.google.devtools.build.lib.packages.BuildFileContainsErrorsException: error loading package '': Encountered error while reading extension file 'cuda/build_defs.bzl': no such package '@local_config_cuda//cuda': Traceback (most recent call last):
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 381
		_create_cuda_repository(repository_ctx)
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 337, in _create_cuda_repository
		repository_ctx.symlink(cudnn_header_dir + "/cudnn.h", "cu...")
java.io.IOException: Could not create symlink from /usr/include/cudnn.h to /opt/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/local_config_cuda/cuda/include/cudnn.h: /opt/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/local_config_cuda/cuda/include/cudnn.h (Permission denied).
```

My understanding is that because `cc_configure` was not able to complete due to the permission issue when trying to symlink `cudnn.h` the workspace it created is incomplete, and the "BUILD file not found on package" error occurs because the BUILD file is generated [at a later step](https://github.com/davidzchen/tensorflow/blob/cuda-configure/third_party/gpus/cuda_configure.bzl#L342). @damienmg, is my understanding of this correct? If so, shouldn't the first error cause the loading phase to abort?

martinwicke:
I wiped the workspace and tried again, but same problem.

davidzchen:
That is odd. I'll investigate further tomorrow after I am back from vacation.

configure
damienmg: Sorry, missed this.  So, there are a couple problems with having fetch
create output dir symlinks. bazel-<workspace-name> is a sylink to the
execution root, which is created right before execution (not during the
loading phase, which is all that fetch runs). Also, it's sort of an
accident that the external repos are under bazel-<workspace-name> at all,
hopefully one that will be corrected in the next couple of days!

However, the point stands, it would be nice to see what the heck fetch
did.  I could see a couple possible options:

1. fetch could print paths to the repositories it fetched (that might be a
bit weird, because I think it would have to print their transitive closure
which might include repositories the user wasn't aware of).
2. Bazel could generally create a convenience symlink to external.

gunan:
Jenkins, test this please

davidzchen:
The header symlinking issue is fixed, but it looks like the Linux GPU build is unable to find the cuDNN libraries:

```
ERROR: com.google.devtools.build.lib.packages.BuildFileContainsErrorsException: error loading package '': Encountered error while reading extension file 'cuda/build_defs.bzl': no such package '@local_config_cuda//cuda': Traceback (most recent call last):
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 406
		_create_cuda_repository(repository_ctx)
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 343, in _create_cuda_repository
		_find_cudnn_lib_path(repository_ctx, cudnn_install_base..., ...)
	File "/workspace/third_party/gpus/cuda_configure.bzl", line 252, in _find_cudnn_lib_path
		fail("Cannot find %s or %s under %s" ...))
Cannot find lib64/libcudnn.so.5 or libcudnn.so.5 under /usr/local/cuda.
```

I have dumped the list of files under both `/usr/local/cuda` and `/usr/local/cuda/lib64` (see [log](http://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/1520/console)), and curiously, `libcudnn.so.5` is under neither directory (though all other cuda libraries appear to be present).

@jendap, @martinwicke - where are the cuDNN libraries installed on the system for the GPU CI Docker containers?

Also, for my understanding, it seems that [`CUDA_DNN_LIB_ALT_PATH`](https://github.com/tensorflow/tensorflow/blob/master/third_party/gpus/cuda/cuda_config.sh#L128) in the existing CUDA detection implementation seems to suggest that `libcudnn.so` may be installed on some systems directly under `/usr/local/cuda` rather than under `/usr/local/cuda/lib` or `/usr/local/cuda/lib64`. Is my understanding of this correct?

gunan:
I just logged into our docker container used for a gpu pip test, and here is what I see.
libcudnn.so location: /usr/lib/x86_64-linux-gnu/libcudnn.so
# echo $LD_LIBRARY_PATH
/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64

# ls -l /usr/local/cuda/lib64
lrwxrwxrwx 1 root root 24 Aug 15  2015 /usr/local/cuda/lib64 -> targets/x86_64-linux/lib

# ls -l /usr/local/cuda/lib  
ls: cannot access /usr/local/cuda/lib: No such file or directory

davidzchen:
@gunan - To confirm: on the docker container for the GPU tests, `libcudnn.so` is installed under `/usr/lib/x86_64-linux-gnu` and not `/usr/local/cuda/lib`? Is this expected?

This seems odd to me since [`Dockerfile.gpu`](https://github.com/tensorflow/tensorflow/blob/73ced9d797056c7e67a06ed2098dd809d85ec44a/tensorflow/tools/ci_build/Dockerfile.gpu#L25) seems to specify that `libcudnn.so` is also installed under `/usr/local/cuda/lib`:

```dockerfile
ENV CUDNN_INSTALL_PATH /usr/local/cuda
```

gunan:
Correct. At least this is what I am seeing. I do not know if this is expected or not, but my bet is it is unexpected.
I also downloaded the clean image from nvidia's dockerhub, and it is the same as what I saw on our jenkins workers.

Does this issue look related?
https://github.com/NVIDIA/nvidia-docker/issues/37

davidzchen:
It's possible. We did just upgraded to cudnn 5 (tensorflow/tensorflow#3773), but it is curious that the existing GPU CI build does not seem to be failing with this issue.

I have opened NVIDIA/nvidia-docker#172 to confirm whether the location of `libcudnn.so` is expected in the image.

gunan:
On my workstation I tried with nvidia/cuda 7.5-cudnn4-devel image and it is in the same location, not under /usr/lib/cuda
So the issue is not specific to cudnn5 image

davidzchen:
I see. Also, @flx42 just [confirmed](https://github.com/NVIDIA/nvidia-docker/issues/172#issuecomment-240015793) on NVIDIA/nvidia-docker#172 that `/usr/lib/x86_64-linux-gnu` is, in fact, the correct install location for `libcudnn.so`.

In that case, is the `ENV CUDNN_INSTALL_PATH` line in `Dockerfile.gpu` is incorrect then?

It is still unclear to me why the CI GPU builds for the existing code is not failing due to this since the existing code [searches for `libcudnn.so` under a `$CUDNN_INSTALL_BASEDIR/lib64` directory](https://github.com/tensorflow/tensorflow/blob/master/third_party/gpus/cuda/cuda_config.sh#L127).

third_party/gpus/cuda/platform.bzl.tpl
```diff
@@ -26,7 +26,7 @@ def cuda_static_library_path(name):
   else:
     return "lib64/lib{}_static.a".format(name)
 
-def cudnn_library_path(version = cudnn_sdk_version()):
+def cudnn_library_path_foo(version = cudnn_sdk_version()):
```
vrv: foo?

third_party/gpus/cuda/platform.bzl.tpl
```diff
@@ -38,6 +38,12 @@ def cudnn_library_path(version = cudnn_sdk_version()):
     else:
       return "lib64/libcudnn.so.{}".format(version)
 
+def cudnn_library_path(version = cudnn_sdk_version()):
+  print("11 version %s" % version)
```
vrv: remove prints?

tensorflow/stream_executor/dso_loader.cc
```diff
@@ -33,54 +33,63 @@ limitations under the License.
...
+      FindDsoPath(
+          tensorflow::internal::FormatLibraryFileName("cublas",
```
vrv: can you clang-format --style=google this file?

davidzchen: Done.

third_party/gpus/cuda/platform.bzl.tpl
davidzchen: Done. I had removed the prints and rebased last night but wanted to wait for the results of the test first. :)

third_party/gpus/cuda/platform.bzl.tpl
davidzchen: This was part of the debug print. Removed.

gunan:
Jenkins, test this please

gunan:
Any progress here?
r0.10 full release is blocked on this PR.

davidzchen:
I was able to get the Linux GPU build to work and made sure that both of the possible install paths for `libcudnn.so` are supported. I also started looking into changes needed internally to merge this into the internal repo (@gunan, I have added you to that thread). I am also oncall this week and have been a bit busy with that as well.

All of the Linux GPU tests are passing except for these:

```
//tensorflow/core:common_runtime_gpu_gpu_tracer_test
//tensorflow/g3doc/how_tos/adding_an_op:cuda_op_test
//tensorflow/python:device_lib_test
```

These tests seem to all be failing due to the following:

```
Actual: Invalid argument: Cannot assign a device to node 'n/_2': Could not satisfy explicit device specification '/job:localhost/replica:0/task:0/gpu:0' because no devices matching that specification are registered in this process; available devices: /job:localhost/replica:0/task:0/cpu:0
	 [[Node: n/_2 = MatMul[T=DT_FLOAT, transpose_a=false, transpose_b=false, _device="/job:localhost/replica:0/task:0/gpu:0"](n/_0, n/_1)]]
Expected: ::tensorflow::Status::OK()
```

These error message seem to be related:

```
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 0 to device ordinal 1
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 0 to device ordinal 2
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 0 to device ordinal 3
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 1 to device ordinal 0
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 1 to device ordinal 2
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 1 to device ordinal 3
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 2 to device ordinal 0
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 2 to device ordinal 1
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 2 to device ordinal 3
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 3 to device ordinal 0
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 3 to device ordinal 1
I tensorflow/core/common_runtime/gpu/gpu_init.cc:61] cannot enable peer access from device ordinal 3 to device ordinal 2
I tensorflow/core/common_runtime/gpu/gpu_init.cc:138] DMA: 0 1 2 3 
I tensorflow/core/common_runtime/gpu/gpu_init.cc:148] 0:   Y N N N 
I tensorflow/core/common_runtime/gpu/gpu_init.cc:148] 1:   N Y N N 
I tensorflow/core/common_runtime/gpu/gpu_init.cc:148] 2:   N N Y N 
I tensorflow/core/common_runtime/gpu/gpu_init.cc:148] 3:   N N N Y 
I tensorflow/core/common_runtime/gpu/gpu_device.cc:842] Ignoring gpu device (device: 0, name: GRID K2, pci bus id: 0000:00:04.0) with Cuda compute capability 3.0. The minimum required Cuda capability is 3.5.
I tensorflow/core/common_runtime/gpu/gpu_device.cc:842] Ignoring gpu device (device: 1, name: GRID K2, pci bus id: 0000:00:05.0) with Cuda compute capability 3.0. The minimum required Cuda capability is 3.5.
I tensorflow/core/common_runtime/gpu/gpu_device.cc:842] Ignoring gpu device (device: 2, name: GRID K2, pci bus id: 0000:00:06.0) with Cuda compute capability 3.0. The minimum required Cuda capability is 3.5.
I tensorflow/core/common_runtime/gpu/gpu_device.cc:842] Ignoring gpu device (device: 3, name: GRID K2, pci bus id: 0000:00:07.0) with Cuda compute capability 3.0. The minimum required Cuda capability is 3.5.
tensorflow/core/debug/debug_gateway_test.cc:140: Failure
Value of: (s)
  Actual: Invalid argument: Cannot assign a device to node 'n/_2': Could not satisfy explicit device specification '/job:localhost/replica:0/task:0/gpu:0' because no devices matching that specification are registered in this process; available devices: /job:localhost/replica:0/task:0/cpu:0
	 [[Node: n/_2 = MatMul[T=DT_FLOAT, transpose_a=false, transpose_b=false, _device="/job:localhost/replica:0/task:0/gpu:0"](n/_0, n/_1)]]
Expected: ::tensorflow::Status::OK()
Which is: OK
```

Is the message about the CUDA compute capability being on the CI machine 3.0 correct?

gunan:
Yes, Cuda compute capability is correct.
We have k2 cards on jenkins' test machines.

Your error looks like those tests were not able to access the GPUs

davidzchen:
Understood. In that case, we need to set the minimum compute capability to 3.0 for the tests. I have made a change to set an env variable for that in Dockerfile.gpu and re-running the tests now.

davidzchen:
I have fixed Dockerfile.gpu to use the compute capability, and Linux GPU tests are now passing. The status of the MacOS CPU Tests is still "Waiting for status to be reported." Are those tests stalled?

This PR itself is ready. There is a bit of work I need to do internally to allow this PR to be correctly imported into the internal repo before we can merge this.

damienmg:
Hi, just coming back from holidays. David: do you still need my help on
this PR?

vrv:
Woohoo!  Mac test machines are currently disabled, but hopefully the changes here should apply there as well.

vrv:
Congrats Rasmus, you'll have to work with David this week to pull this change in internally :)

rmlarsen:
Uh oh, that have I done!?!

gunan:
Now, what is the best way to pull this PR into r0.10?

vrv:
git cherry-pick :)

davidzchen:
Thank you everyone for taking the time to help me understand the intricacies of the TensorFlow CUDA build setup and taking the time to review this change! :)

martinwicke:
This change is big enough in terms of the impact on installation that we
may want to wait another full 7 days until we exit RC (assuming everything
else is clear). I'd like to see the effect on issues first and have this
field tested by the community on more systems than we can muster here.

gunan:
Martin, do you think we should re-release the docker images and pip wheel files for the RC?

martinwicke:
To make RC1? This change won't affect the binaries at all, so I don't think
that's necessary.

flx42:
@davidzchen With this PR, it looks like setting variable `TF_CUDA_COMPUTE_CAPABILITIES` before the configure doesn't work anymore, am I missing something? Now you have to set `CUDA_COMPUTE_CAPABILITIES` variable instead.

davidzchen:
@flx42 Setting `TF_CUDA_COMPUTE_CAPABILITIES` before running `configure` should still work since that part of the `configure` script (showing the Cuda compute capabilities prompt only if `TF_CUDA_COMPUTE_CAPABILITIES` is not set) is still there: https://github.com/tensorflow/tensorflow/blob/master/configure#L239

Running `TF_CUDA_COMPUTE_CAPABILITIES=3.0,5.2 ./configure`, then running a bazel build command, then looking at the compute capabilities generated in `bazel-tensorflow/external/local_config_cuda/cuda/cuda_config.h` confirms that the correct compute capabilities were set.

What is the issue that you are observing?

flx42:
> Running TF_CUDA_COMPUTE_CAPABILITIES=3.0,5.2 ./configure, then running a bazel build command, then looking at the compute capabilities generated in bazel-tensorflow/external/local_config_cuda/cuda/cuda_config.h confirms that the correct compute capabilities were set.

That's the two default values, so I'm not sure you're testing anything here :).

On my side, I tried with a modified version of the current Dockerfile.devel-gpu:
```
diff --git a/tensorflow/tools/docker/Dockerfile.devel-gpu b/tensorflow/tools/docker/Dockerfile.devel-gpu
index 4877f8c..3b542b0 100644
--- a/tensorflow/tools/docker/Dockerfile.devel-gpu
+++ b/tensorflow/tools/docker/Dockerfile.devel-gpu
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:7.5-cudnn5-devel
+FROM nvidia/cuda:8.0-cudnn5-devel
 
 MAINTAINER Craig Citro <craigcitro@google.com>
 
@@ -89,6 +89,7 @@ WORKDIR /tensorflow
 ENV CUDA_TOOLKIT_PATH /usr/local/cuda
 ENV CUDNN_INSTALL_PATH /usr/local/cuda
 ENV TF_NEED_CUDA 1
+ENV TF_CUDA_COMPUTE_CAPABILITIES "6.0"
 
 RUN ./configure && \
     bazel build -c opt --config=cuda tensorflow/tools/pip_package:build_pip_package && \
```
And I saw 3.5, 5.2 in cuda_config.h

flx42:
@davidzchen Look like after the `configure` I do have 6.0 in `cuda_config.h`, but during the `bazel build` step, it becomes `3.5,5.2` instead.

davidzchen:
@flx42 That is odd. I'll take a look. #4002 is the tracking bug opened for that issue.

PR #3143, opened 2016-07-01
Misc improvements to Dockerfile.devel-gpu

cancan101:

PR #3272, opened 2016-07-11
Revert "Revert "Misc improvements to Dockerfile.devel-gpu""

martinwicke:
Reverts tensorflow/tensorflow#3157 which reverted #3143, but which now has proper CLA.

### DIFF
--- a/tensorflow/tools/docker/Dockerfile.devel-gpu
+++ b/tensorflow/tools/docker/Dockerfile.devel-gpu
@@ -80,7 +80,7 @@ RUN mkdir /bazel && \
 
 # Download and build TensorFlow.
 
-RUN git clone --recursive https://github.com/tensorflow/tensorflow.git && \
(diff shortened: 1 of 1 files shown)


tensorflow/tools/docker/Dockerfile.devel-gpu
```diff
@@ -80,7 +80,7 @@ RUN mkdir /bazel && \
 
 # Download and build TensorFlow.
 
-RUN git clone --recursive https://github.com/tensorflow/tensorflow.git && \
+RUN git clone -b r0.9 --recursive --recurse-submodules https://github.com/tensorflow/tensorflow.git && \
```
vrv: --recurse-submodules isn't needed anymore, right? we don't have submodules..

martinwicke:
Yeah, it's unnecessary, you're right. Oh well. Doesn't hurt either.

PR #3275, opened 2016-07-11
Update Eigen to version that includes scan op fix

ibab:
This updates Eigen to a newer version that includes a fix needed to merge #2711.

### DIFF
--- a/tensorflow/contrib/cmake/external/eigen.cmake
+++ b/tensorflow/contrib/cmake/external/eigen.cmake
@@ -7,7 +7,7 @@
 
 include (ExternalProject)
 
-set(eigen_archive_hash "802d984ade26")
+set(eigen_archive_hash "bbffe8ed5f26")
 
 set(eigen_INCLUDE_DIRS
     ${CMAKE_CURRENT_BINARY_DIR}
--- a/eigen.BUILD
+++ b/eigen.BUILD
@@ -1,6 +1,6 @@
 package(default_visibility = ["//visibility:public"])
 
-archive_dir = "eigen-eigen-334b1d428283"
+archive_dir = "eigen-eigen-bbffe8ed5f26"
 cc_library(
     name = "eigen",
     hdrs = glob([archive_dir+"/**/*.h", archive_dir+"/unsupported/Eigen/CXX11/*", archive_dir+"/Eigen/*"]),
--- a/tensorflow/workspace.bzl
+++ b/tensorflow/workspace.bzl
@@ -6,8 +6,8 @@
 def tf_workspace(path_prefix = "", tf_repo_name = ""):
   native.new_http_archive(
     name = "eigen_archive",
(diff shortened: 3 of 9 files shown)


vrv:
https://github.com/tensorflow/tensorflow/blob/dfb71ea206eb9f61e5d97c9727caa1a6449e39cb/tensorflow/contrib/cmake/external/eigen.cmake#L10 can you update that too?

ibab:
Yeah, sorry for forgetting about the cmake build. I've updated it too, now.

vrv:

vrv:
@tensorflow-jenkins test this please (once more, I don't expect the result to change.  cmake is always broken anyway :/).  Will merge if everything else passes.

ibab:
Cool, the cmake test build is a good idea.
The error
```
execvp: /usr/local/bin/protoc: Permission denied
```
seems a bit bizarre. Maybe the filesystem is mounted with `noexec`, or SELinux is interfering?

ibab:
The protobuf installation is unpacked from a zip file here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/install/install_proto3.sh#L29
Maybe the executable permissions aren't preserved when unpacking. (Although it seems to have the right permissions when I unzip that file on my laptop)

tensorflow/contrib/cmake/external/eigen.cmake
```diff
@@ -7,7 +7,7 @@
 
 include (ExternalProject)
 
-set(eigen_archive_hash "802d984ade26")
+set(eigen_archive_hash "bbffe8ed5f26")
```
vrv: the eigen hash below also needs to be updated :(

vrv:
Bah, I am just going to merge this anyway, benoit has another update that should hopefully fix this anyway.

tensorflow/contrib/cmake/external/eigen.cmake
ibab: :(

vrv: updating eigen is hard.

ibab:
Thanks! That means #2711 should work now after a rebase.

tensorflow/contrib/cmake/external/eigen.cmake
ibab: Yeah, it would be better if the unpacked archive would be renamed to something fixed, then we wouldn't have to touch all those files.

ibab: Maybe this new bazel option can help: http://www.bazel.io/docs/be/workspace.html#new_http_archive.strip_prefix

vrv: Interesting, would this work @benoitsteiner ?

ibab: I've added a working example of the idea as a WIP pull request in #3288.

Issue #3271, opened 2016-07-11
Reverse for complex numbers


el3ment:
tf.reverse does not seem to be implemented for complex numbers

```
w_real = tf.Variable(tf.truncated_normal([10], stddev=0.1, dtype=tf.float32))
w_imag = tf.Variable(tf.truncated_normal([10], stddev=0.1, dtype=tf.float32))
w = tf.complex(w_real, w_imag)
w_rev = tf.reverse(w, [True])
```

I came across this needing to construct a matrix that is conjugate symmetric, but this was also referenced on StackOverflow : [http://stackoverflow.com/questions/38145235/why-cant-i-reverse-a-sequence-of-complex-numbers-in-tensorflow](http://stackoverflow.com/questions/38145235/why-cant-i-reverse-a-sequence-of-complex-numbers-in-tensorflow)

sjperkins:
Related to https://github.com/tensorflow/tensorflow/issues/2577 which tracks general support for complex numbers in tensor transformation functions.

ibab:
I've enabled `reverse` for complex numbers in #3281, so this can be closed if everyone agrees.

Issue #2577, opened 2016-05-30
Support complex numbers in Tensor Transformation Operations


sjperkins:
A number of changes have been made to support complex128, but the the tensor transformation op suite seems to be missing this support (At least in tf.pack and tf.concat). See the example script below which works for float32 and float64, but not complex64 and complex128.
### Environment info

Operating System: Ubuntu 14.04.4

Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):

``` bash
$ ls -l /usr/local/cuda-7.5/lib64/libcud*
-rw-r--r-- 1 root root    322936 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudadevrt.a
lrwxrwxrwx 1 root root        16 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart.so -> libcudart.so.7.5
lrwxrwxrwx 1 root root        19 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart.so.7.5 -> libcudart.so.7.5.18
-rwxr-xr-x 1 root root    383336 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart.so.7.5.18
-rw-r--r-- 1 root root    720192 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart_static.a
lrwxrwxrwx 1 3319 users       13 Feb  9 19:48 /usr/local/cuda-7.5/lib64/libcudnn.so -> libcudnn.so.4
lrwxrwxrwx 1 3319 users       17 Feb  9 19:48 /usr/local/cuda-7.5/lib64/libcudnn.so.4 -> libcudnn.so.4.0.7
-rwxrwxr-x 1 3319 users 61453024 Feb  9 00:12 /usr/local/cuda-7.5/lib64/libcudnn.so.4.0.7
-rw-rw-r-- 1 3319 users 62025862 Feb  9 00:12 /usr/local/cuda-7.5/lib64/libcudnn_static.a
```

If installed from binary pip package, provide:
1. Which pip package you installed. **Python 2 GPU nightly**
2. The output from python -c "import tensorflow; print(tensorflow.**version**)". **0.8.0**

If installed from sources, provide the commit hash:
### Steps to reproduce
1. Vary **dtype** in the following test script between `float32, float64, complex64, complex128`
   
   ``` python
   import numpy as np
   import tensorflow as tf
   
   # Data type for this test case
[...]

PR #3281, opened 2016-07-12
Enable reverse op for complex values

ibab:
I've enabled complex values for the reverse op based on #3271.
This is part of the more general #2577.

I've also extended the tests to cover all enabled dtypes.
One thing that I've noticed is that we're compiling GPU code for `bool` and `int32`, but we're not enabling GPU kernels for these dtypes.
I've added commented out versions of the kernel registrations for these dtypes to point out that they're currently not enabled.


vrv:
@girving can I assign this one to you?  If not, feel free to re-assign.

tensorflow/core/kernels/reverse_op.cc
```diff
@@ -149,9 +145,13 @@ TF_CALL_double(DECLARE_GPU_SPEC);
                           ReverseOp<GPUDevice, T>)
 TF_CALL_uint8(REGISTER_GPU_KERNEL);
 TF_CALL_int8(REGISTER_GPU_KERNEL);
+//TF_CALL_int32(REGISTER_GPU_KERNEL);
+//TF_CALL_bool(REGISTER_GPU_KERNEL);
```
girving: Why are these commented out?  If there are broken reasons why they can't be added, maybe leave them commented out but add a TODO comment saying why.

girving:
Looks good except for the comment about the comment.

tensorflow/core/kernels/reverse_op.cc
ibab: I've tried enabling them, but had a problem with `int32`.
The tests were crashing without an error message for me :/
So I wasn't quite sure what to put in the TODO comment.
Should I add a comment saying `// TODO Find out why int32 doesn't work` ?

ibab:
Okay, I've added a `TODO` to the commented out op registrations.
I could have a look at enabling those in a second PR.

girving:
Jenkins, test this please.

girving:
Jenkins, test this please.

girving:
@vrv Do you know the tests wouldn't be triggering here?

vrv:
i dunno, Jenkins is not exactly reliable ;).

girving:
Failures seem unrelated.  Merging.  Thank you for the stream of contributions!

PR #3282, opened 2016-07-12
Add Install step in Installation for Linux.

raix852:
- Add Install step in in Installation for Linux for better following installation.
- Fix layout in RNN tutorials.

### DIFF
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -679,7 +679,7 @@ Every time you change the Cuda library paths you need to run this step again bef
 you invoke the bazel build command. For the cuDNN libraries, use '6.5' for R2, '7.0'
 for R3, and '4.0.4' for R4-RC.
 
-##### Build your target with GPU support
+#### Build your target with GPU support
 From the root of your source tree, run:
 
 ```bash
@@ -695,7 +695,7 @@ $ bazel-bin/tensorflow/cc/tutorials_example_trainer --use_gpu
 
 Note that "--config=cuda" is needed to enable the GPU support.
 
-##### Known issues
+#### Known issues
 
 * Although it is possible to build both Cuda and non-Cuda configs under the same
 source tree, we recommend to run `bazel clean` when switching between these two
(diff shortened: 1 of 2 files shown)


tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -623,6 +623,8 @@ Every time you change the Cuda library paths you need to run this step again bef
 you invoke the bazel build command. For the cuDNN libraries, use '6.5' for R2, '7.0'
 for R3, and '4.0.4' for R4-RC.
 
+#### Install
+[Create the pip package and install](#create-the-pip-package-and-install)
 
 ##### Build your target with GPU support
 From the root of your source tree, run:
```
vrv: Actually, I think this next section here is redundant, and would propose removing it.  Thoughts @martinwicke  ?

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -623,6 +623,8 @@ Every time you change the Cuda library paths you need to run this step again bef
 you invoke the bazel build command. For the cuDNN libraries, use '6.5' for R2, '7.0'
 for R3, and '4.0.4' for R4-RC.
 
+#### Install
+[Create the pip package and install](#create-the-pip-package-and-install)
```
martinwicke: Add a period at the end?

tensorflow/g3doc/get_started/os_setup.md
martinwicke: Yes, it's redundant. Also the ./configure step should just be moved to the "Configure the installation" section, to replace the description of the configure step there.

raix852:
Merged configure the installation as suggested. But I didn't have an experience on configure GPU because lacked of environment.  So it maybe have some incorrect parts.

martinwicke:
I think this is fine. Reading the whole thing I noticed it's still pretty garbled, but this is a good start. We should probably rewrite this to factor the GPU story out completely -- it's currently repeated between MacOS and Linux.

Anyway, LGTM.

martinwicke:
Jenkins, test this please.

Issue #2826, opened 2016-06-13
A 1D and 2D Gaussian implementation in TensorFlow


RobRomijnders:
Feature request:

Mixture Density networks, attention models and variational inference getting more popular, we are defining more and more Gaussians in our model. My feature request would be to have a native Gaussian implementation in TensorFlow.

Up to now, I re-use [this](https://github.com/hardmaru/write-rnn-tensorflow/blob/master/model.py) code:

    def tf_2d_normal(x1, x2, mu1, mu2, s1, s2, rho):
      # eq # 24 and 25 of http://arxiv.org/abs/1308.0850
      norm1 = tf.sub(x1, mu1)
      norm2 = tf.sub(x2, mu2)
      s1s2 = tf.mul(s1, s2)
      z = tf.square(tf.div(norm1, s1))+tf.square(tf.div(norm2, s2))-2*tf.div(tf.mul(rho, tf.mul(norm1, norm2)), s1s2)
      negRho = 1-tf.square(rho)
      result = tf.exp(tf.div(-z,2*negRho))
      denom = 2*np.pi*tf.mul(s1s2, tf.sqrt(negRho))
      result = tf.div(result, denom)
      return result

However, it might be nice for Tensorflow to implement it in the library. They might be able to apply it more computationally efficient? Also, they might go for N-dimensional Gaussians, not only two.

What do you think? Is this feasible?

suiyuan2009:
I'll do it, implement 1D and 2D Gaussian

ebrevdo:
See tf.contrib.distributions for normal and multivariate normal distribution classes.

PR #3284, opened 2016-07-12
add norm1d and 2d support

suiyuan2009:
#2826
- add norm1d support
- I'll write norm2d, and add test

### DIFF
--- a/tensorflow/core/kernels/norm_ops.cc
+++ b/tensorflow/core/kernels/norm_ops.cc
@@ -1,0 +1,69 @@
+
+#define EIGEN_USE_THREADS
+
+#include "tensorflow/core/kernels/norm_ops.h"
+#include "tensorflow/core/framework/op.h"
+#include "tensorflow/core/framework/op_kernel.h"
+#include "tensorflow/core/kernels/bounds_check.h"
+#include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/kernels/fill_functor.h"
+
+namespace tensorflow {
+typedef Eigen::ThreadPoolDevice CPUDevice;
+
+namespace functor {
+
+template <typename T>
+struct Norm1D<CPUDevice, T> {
+   void operator()(const CPUDevice& d, typename TTypes<T>::ConstFlat x,
+		   typename TTypes<T>::ConstFlat mean,
+		   typename TTypes<T>::ConstFlat stdd,
+		   typename TTypes<T>::Flat out) {
+	   auto new_x = x - mean;
+	   auto z = - new_x.square() / (stdd.constant(2.0) * stdd.square());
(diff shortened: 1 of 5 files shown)


vrv:
I'm not sure how much of this CL overlaps (if any) with the existing ops, assigning to those who are more familiar

suiyuan2009:
@vrv I see #2826 , so I think it's not implemented before

tensorflow/core/kernels/norm_ops.cc
```diff
@@ -0,0 +1,69 @@
...
+namespace functor {
+
+template <typename T>
+struct Norm1D<CPUDevice, T> {
+   void operator()(const CPUDevice& d, typename TTypes<T>::ConstFlat x,
+		   typename TTypes<T>::ConstFlat mean,
+		   typename TTypes<T>::ConstFlat stdd,
+		   typename TTypes<T>::Flat out) {
+	   auto new_x = x - mean;
+	   auto z = - new_x.square() / (stdd.constant(2.0) * stdd.square());
```
benoitsteiner: I would force the 2.0 constant to be of the same type as T by writing the code as:
   stdd.constant(static_cast<T>(2.0))

tensorflow/core/kernels/norm_ops.cc
```diff
@@ -0,0 +1,69 @@
...
+
+namespace functor {
+
+template <typename T>
+struct Norm1D<CPUDevice, T> {
+   void operator()(const CPUDevice& d, typename TTypes<T>::ConstFlat x,
+		   typename TTypes<T>::ConstFlat mean,
+		   typename TTypes<T>::ConstFlat stdd,
+		   typename TTypes<T>::Flat out) {
+	   auto new_x = x - mean;
```
benoitsteiner: you could define new_x as mean-x: this would avoid the need for the negation in the computation of z below

ebrevdo:
This is a premature optimization. See tf.contrib.distributions.

This op doesn't support broadcasting and has no registered gradients.  It's not clear that it's any faster the equivalent tensorflow code.

suiyuan2009:
@benoitsteiner @ebrevdo thank you for the code review, learned  a lot,  I'll close this pr.

suiyuan2009:
hope I can help if optimization in tf.contrib is needed

tensorflow/core/kernels/norm_ops.cc
suiyuan2009: thank you anyway, but negation can't be avoided becase `square()`

vrv:
I think some numbers about why tf.contrib.distributions is not fast enough would be needed before accepting this (in general we only try to add custom ops when we can't get the equivalent from compositions of primitive ops at a reasonable speed).

PR #3288, opened 2016-07-13
Simplify Eigen package config

ibab:
I think we can avoid specifying the Eigen commit hash in multiple places by using the `strip_prefix` option to `new_http_archive`.

Here, I've declared the real Eigen includes without a prefix, while the includes in `third_party` can only be reached by including `third_party/eigen3/...` (to avoid an infinite inclusion loop).
I've used this setup to successfully build tensorflow.

This can probably still be improved by not declaring the real Eigen includes at the top level.
Maybe it's possible to access them under `eigen_archive/...`, although I'm not sure how to set that up.

### DIFF
--- a/eigen.BUILD
+++ b/eigen.BUILD
@@ -1,9 +1,8 @@
 package(default_visibility = ["//visibility:public"])
 
-archive_dir = "eigen-eigen-b4fa9622b809"
 cc_library(
     name = "eigen",
-    hdrs = glob([archive_dir+"/**/*.h", archive_dir+"/unsupported/Eigen/*", archive_dir+"/unsupported/Eigen/CXX11/*", archive_dir+"/Eigen/*"]),
-    includes = [ archive_dir ],
+    hdrs = glob(["**/*.h", "unsupported/Eigen/CXX11/*", "Eigen/*"]),
+    includes = [ '.' ],
     visibility = ["//visibility:public"],
 )
--- a/eigen.BUILD
+++ b/eigen.BUILD
@@ -2,7 +2,7 @@ package(default_visibility = ["//visibility:public"])
 
 cc_library(
     name = "eigen",
-    hdrs = glob(["**/*.h", "unsupported/Eigen/CXX11/*", "Eigen/*"]),
+    hdrs = glob(["**/*.h", "unsupported/Eigen/*", "unsupported/Eigen/CXX11/*", "Eigen/*"]),
     includes = [ '.' ],
     visibility = ["//visibility:public"],
 )
--- a/tensorflow/contrib/makefile/Makefile
+++ b/tensorflow/contrib/makefile/Makefile
@@ -38,8 +38,9 @@ HOST_OBJDIR := $(MAKEFILE_DIR)/gen/host_obj/
 HOST_BINDIR := $(MAKEFILE_DIR)/gen/host_bin/
 HOST_GENDIR := $(MAKEFILE_DIR)/gen/host_obj/
 
-# Find the current Eigen version name from the Bazel build file
-EIGEN_HASH := $(shell cat eigen.BUILD | grep archive_dir | head -1 | cut -f3 -d- | cut -f1 -d\")
+# Find the current Eigen version name from the cmake configuration
+EIGEN_HASH := $(shell cat tensorflow/contrib/cmake/external/eigen.cmake | \
+	                  grep eigen_archive_hash | head -1 | sed -e 's/.*"\(.*\)".*/\1/')
 
 # Settings for the host compiler.
 HOST_CXX := $(CC_PREFIX) gcc
--- a/tensorflow/contrib/makefile/Makefile
+++ b/tensorflow/contrib/makefile/Makefile
@@ -38,9 +38,8 @@ HOST_OBJDIR := $(MAKEFILE_DIR)/gen/host_obj/
 HOST_BINDIR := $(MAKEFILE_DIR)/gen/host_bin/
 HOST_GENDIR := $(MAKEFILE_DIR)/gen/host_obj/
 
-# Find the current Eigen version name from the cmake configuration
-EIGEN_HASH := $(shell cat tensorflow/contrib/cmake/external/eigen.cmake | \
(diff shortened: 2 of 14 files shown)


vrv:
Sorry, there's a conflict I just introduced from another update to eigen libraries -- can you rebase and we can try this again?

ibab:
No problem, I've rebased the PR.

eigen.BUILD
```diff
@@ -1,9 +1,8 @@
 package(default_visibility = ["//visibility:public"])
 
-archive_dir = "eigen-eigen-b4fa9622b809"
 cc_library(
     name = "eigen",
-    hdrs = glob([archive_dir+"/**/*.h", archive_dir+"/unsupported/Eigen/*", archive_dir+"/unsupported/Eigen/CXX11/*", archive_dir+"/Eigen/*"]),
-    includes = [ archive_dir ],
+    hdrs = glob(["**/*.h", "unsupported/Eigen/CXX11/*", "Eigen/*"]),
```
vrv: are we missing an unsupported/Eigen/* ?

ibab: Yeah, it seems like it was added in one of the last few commits, so it didn't make it into the PR before.

vrv:
Sweet, let's give this a shot.  @tensorflow-jenkins test this please

vrv:
hm, that's an interesting failure.  Have you successfully been able to compile with this?

ibab:
I'm seeing the same problem now.
I'm pretty sure managed to compile it before the rebase, though :/

vrv:
The cpu test failed with a more appropriate error -- maybe we can resolve the CPU ones first and then figure out what's going on with GPU?

ibab:
I noticed that there's a new `third_party` Eigen header (at `third_party/eigen3/unsupported/Eigen/SpecialFunctions`) but adjusting it didn't help.

ibab:
Okay, the CPU tests are failing because I haven't adjusted the code that creates the pip package (It still assumes that eigen is located in `eigen_archive/eigen-eigen-\w+`).
Not sure about the GPU tests, though.

third_party/eigen3/unsupported/Eigen/CXX11/Tensor
```diff
@@ -1 +1 @@
-#include "eigen-eigen-b4fa9622b809/unsupported/Eigen/CXX11/Tensor"
+#include "unsupported/Eigen/CXX11/Tensor"
```
benoitsteiner: You also need to adjust the new unsupported/Eigen/SpecialFunctions header

ibab:
I'm still not sure what's causing the problem with the GPU build.
Maybe the fact that Eigen is including the curand headers.
I'll have a closer look later.

ibab:
Finally got around to this.
It turns out that that the problem is caused by [a workaround in the GPU crosstool wrapper](https://github.com/tensorflow/tensorflow/blob/e95f4e760c6b6713b6b686ebeff9a1586a5831dd/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc#L103) that's used to support CUDA <= 6.0. Does TensorFlow still support those versions? If not, I could remove it.
I've added some extra code that makes the build work with the workaround.

Both problems from before should be solved now, so we can have another go at the tests :)

vrv:
I am fine dropping support for cuda 6.0, so go ahead!

ibab:
Okay, I've removed that part of the crosstool script and pushed my changes.

tensorflow/contrib/makefile/Makefile
```diff
@@ -38,8 +38,9 @@ HOST_OBJDIR := $(MAKEFILE_DIR)/gen/host_obj/
 HOST_BINDIR := $(MAKEFILE_DIR)/gen/host_bin/
 HOST_GENDIR := $(MAKEFILE_DIR)/gen/host_obj/
 
-# Find the current Eigen version name from the Bazel build file
-EIGEN_HASH := $(shell cat eigen.BUILD | grep archive_dir | head -1 | cut -f3 -d- | cut -f1 -d\")
+# Find the current Eigen version name from the cmake configuration
+EIGEN_HASH := $(shell cat tensorflow/contrib/cmake/external/eigen.cmake | \
```
vrv: I think it would still be nice to have the eigen hash in a core file rather than in contrib.

Since workspace.bzl is where the actual git repository is fetched from, what do you think about changing this script to look there?

vrv:
Going to kick off a run just to see.  @tensorflow-jenkins test this please

tensorflow/contrib/makefile/Makefile
ibab: Yeah, that's a good idea.
I started doing that, but found it a bit brittle because the `workspace.bzl` doesn't state "this is the eigen hash" in a clear way.

I could define an `eigen_hash` variable in `workspace.bzl` and add a comment that this line is parsed from the makefile and cmake builds.

tensorflow/workspace.bzl
```diff
@@ -8,6 +8,7 @@ def tf_workspace(path_prefix = "", tf_repo_name = ""):
     name = "eigen_archive",
     url = "https://bitbucket.org/eigen/eigen/get/b4fa9622b809.tar.gz",
     sha256 = "2862840c2de9c0473a4ef20f8678949ae89ab25965352ee53329e63ba46cec62",
+    strip_prefix = "eigen-eigen-b4fa9622b809",
```
vrv: yeah, somehow separating the hash here would be great -- then we only have to replace the hash and the sha256 in one file and updating eigen becomes super easy :)

ibab:
I've pushed a new commit that isolates the `eigen_version` and `eigen_sha256` in the bazel config and adds some code to the cmake and make builds to extract those values.
I've used built-in tools in cmake to make sure that the config stays portable.

vrv:
Woohoo!  This is awesome.  

cc @rmlarsen  -- our hacky scripts for updating eigen just got easier.

vrv:
Seems like an unrelated failure with cc_op_gen.cc, so merging.

ibab:
Thanks!
If there should be an issue with the cmake build as a result of this PR, you can post it here and I'll fix it.

PR #3292, opened 2016-07-13
Minor code doc fix - cwise_ops_common.h

siddharth-agrawal:
Found a minor error in code documentation while digging for something. The file `cwise_functors.h` does not exist, and the `functor` definitions are inside `cwise_ops.h`. Also, couldn't find the `add2 functor`, so replaced it with `add`.


caisq:

PR #3297, opened 2016-07-13
time_major parameter for ctc_loss op

igormq:
Just as dynamic_rnn and to be more regular across the TensorFlow's code I added a flag capable of transpose the input data to be in time_major, required by gen_ctc_ops._ctc_loss. Note that, instead of set time_major=False (as the most of TensorFlow data), it was set False to be compatible with previous ctc_loss function call (but I don't think that this is a good idea).

### DIFF
--- a/tensorflow/python/ops/ctc_ops.py
+++ b/tensorflow/python/ops/ctc_ops.py
@@ -28,7 +28,7 @@ from tensorflow.python.ops.nn_grad import _BroadcastMul
 
 # pylint: disable=protected-access, invalid-name
 def ctc_loss(inputs, labels, sequence_length,
-             preprocess_collapse_repeated=False, ctc_merge_repeated=True):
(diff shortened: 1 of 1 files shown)


danmane:
@ebrevdo Can you review this please?

rmlarsen:
Friendly ping @ebrevdo

ebrevdo:
Can you add a quick unit test ensuring that given input with batch_size > 1, max time > 1,  transposing & setting time_major=True results in the same loss?  i just want to exercise this branch of the code.

ebrevdo:
Jenkins, test this please.

rmlarsen:
@igormq Please rebase and resolve conflicts.

igormq:
I'll have to create another PR, because this repo was erased by mistake, so I'm unable to make the corrections. Sorry for that.

igormq:
Sorry again for the mess. I made the PR #3993  with the test.

PR #3303, opened 2016-07-13
Remove confusing section in Udacity doc.

vincentvanhoucke:
Removing the --rm is simpler. If users want to use a named container, they can, but it's not essential.

### DIFF
--- a/tensorflow/examples/udacity/README.md
+++ b/tensorflow/examples/udacity/README.md
@@ -19,21 +19,6 @@ On mac, find the virtual machine's IP using:
 
 Then go to: http://IP:8888 (likely http://192.168.99.100:8888)
 
-Saving Your Progress
---------------------
-
-Because of the `--rm` flag above, stopping the docker container removes it, so any changes you've made will disappear. One way around this is to remove the `--rm` flag, and name the container for easy restarting:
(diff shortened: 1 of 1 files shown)


hrldcpr:
I worry that even without the `--rm`, people unfamiliar with Docker still won't know how to resume the stopped container. I expect they'll try the `docker run` command again, which will create another fresh container.

Realizing that you have to use `docker ps -a` to get the id of the stopped container and then use `docker start` to resume it seems tricky with no instructions, unless I'm missing something obvious?

hrldcpr:
(Overall the most user-friendly approach might be to mount the local filesystem in the container, something like:

    git clone https://github.com/tensorflow/tensorflow
    cd tensorflow/tensorflow/examples/udacity
    docker run -v $PWD:/notebooks -p 8888:8888 -it b.gcr.io/tensorflow-udacity/assignments:0.5.0

Then people who don't know Docker would be able to see the files in their normal filesystem, could simply run the `docker run` command again whenever they want, edit files locally, et cetera…

But anyway these are just some rambling thoughts! 🙉 It's a great course so far, thanks. I'm sure everyone figures out the nuances of Docker one way or another.)

Issue #508, opened 2015-12-14
Document the broadcasting behavior for `tf.mul` etc.


mrry:
The fact that many of the elementwise operators in TensorFlow support numpy-style broadcasting is not well documented or discoverable. (e.g. [This SO question.](http://stackoverflow.com/questions/34220374/apply-1-channel-mask-to-3-channel-tensor-in-tensorflow/34220963)) We should improve the documentation on this.

girving:
@mrry: Are you still planning to work on this?

mrry:
Not urgently, so perhaps "contributions welcome" is more appropriate.

girving:
If anyone takes this on, it should go in the doc string at the top of `math_ops.py`.

siddharth-agrawal:
@girving Should this be a generic comment saying that TF uses numpy-style broadcasting? Or should it be specific to the operations that are affected by it (alternatively just mention the operations)?

girving:
TF doesn't use numpy style broadcasting for all ops, just for elementwise operations (specifically binary elementwise operations).  Referring to elementwise ops is sufficient; there's no need to list them all.

siddharth-agrawal:
@girving Should this issue be closed? Or are there other changes that need to be made?

vrv:
We could have documentation on every elementwise op that supports broadcasting (numpy does mention broadcasting somewhere in the notes of every op.

girving:
@vrv Putting it in every op seems unpleasant to maintain.

siddharth-agrawal:
Final thoughts on this? I agree with @vrv, I think we should have one sentence for each op that supports broadcasting.

girving:
I'll reverse my previous position: putting a sentence in every op, ideally with a link to a more detailed discussion of broadcasting, would be great.

michaelisard:
@zffchen78 would you update the doc based on the comments in #3534?

tatatodd:
@zffchen78 Friendly ping, can you update this issue once it's fixed?

vrv:
I believe this has been "fixed" with a commit that added this to all of the element wise ops in their cc files.

PR #3314, opened 2016-07-14
Documentation for broadcasting behavior of binary elementwise operations

siddharth-agrawal:
PR for #508. Uses @mrry's example in the SO question referenced in the issue.

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -15,22 +15,6 @@
 
 """Note: Elementwise binary operations in TensorFlow follow [numpy-style
 broadcasting](http://docs.scipy.org/doc/numpy/user/basics.broadcasting.html).
-Let us take an example of `tf.mul()` which falls in the above mentioned
-category, to clarify broadcasting and also give example usage of
-`tf.expand_dims()` when broadcasting is desired.
-
-Ex: Let us say that we have a batch of images in the tensor `image_batch` which
-is of size `b x h x w x c`, and a mask tensor `image_mask` of size `h x w` that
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/math_ops.py
```diff
@@ -13,7 +13,26 @@
...
 
-"""## Arithmetic Operators
+"""Note: Elementwise binary operations in TensorFlow follow [numpy-style
+broadcasting](http://docs.scipy.org/doc/numpy/user/basics.broadcasting.html).
+Let us take an example of `tf.mul()` which falls in the above mentioned
+category, to clarify broadcasting and also give example usage of
+`tf.expand_dims()` when broadcasting is desired.
+
+Ex: Let us say that we have a batch of images in the tensor `image_batch` which
+is of size `b x h x w x c`, and a mask tensor `image_mask` of size `h x w` that
```
vrv: 'which has shape [b, h, w, c], and as mask tensor of size '[h, w]' ...

tensorflow/python/ops/math_ops.py
```diff
@@ -13,7 +13,26 @@
...
+we need to apply to all the images in the batch across all the channels. We thus
```
vrv: Do you think it's necessary to have additional text here?  Why isn't the link to the numpy page enough?  Thoughts @martinwicke  ?

martinwicke: I'd be done with just the reference to numpy... No example here is complete, and may actually be confusing.

siddharth-agrawal: @vrv @martinwicke Updated the PR.

PR #3323, opened 2016-07-15
Allow tensorboard to follow symlinks (feature request #2140)

### LINKED ISSUE #2140: Feature request: support symbolic links for tensorboard
It would be nice if one could use symbolic links to organize groups of training files. Tensorboard does not search over symbolic links for event files because the following line 178 of python/summary/event_multiplexer.py:

```python
subdirs = [
        subdir
        for (subdir, files) in io_wrapper.ListRecursively(path)
        if list(filter(event_accumulator.IsTensorFlowEventsFile, files))
]
```

does not walk across symbolic links. This would be easy to fix by adding a symlink option to ListRecursively().

Current difficulties: 
1) Over time, the number of distinct training runs in a directory can grow to over 10, at which point tensorboard becomes slow and too information rich. Reorganizing a subset with symbolic links would be quite nice.

2) Sometimes it would be nice to compare a small subset of training runs that exist in different directories. Copying these files with `cp -r` is quite slow.

rohitgirdhar:
Fixes #2140

### DIFF
--- a/tensorflow/python/platform/gfile.py
+++ b/tensorflow/python/platform/gfile.py
@@ -382,13 +382,15 @@ def ListDirectory(directory, return_dotfiles=False):  # pylint: disable=invalid-
   return files
 
 
-def Walk(top, topdown=1, onerror=None):
+def Walk(top, topdown=1, onerror=None, followlinks=False):
   """Recursive directory tree generator.
 
   Args:
     top: string, a pathname.
     topdown: bool, should traversal be pre-order (True) or post-order (False)
     onerror: function, optional callback for errors.
+    followlinks: bool, should walk into symbolic links that resolve into
+      directories (true)
 
   By default, errors that occur when listing a directory are ignored.
   (This is the same semantics as Python's os.walk() generator.)  If the
@@ -401,7 +403,7 @@ def Walk(top, topdown=1, onerror=None):
     # by lists of all its subdirectories and leaf files.
     (dirname, [subdirname, subdirname, ...], [filename, filename, ...])
   """
-  return os.walk(top, topdown=topdown, onerror=onerror)
+  return os.walk(top, topdown=topdown, onerror=onerror, followlinks=followlinks)
 
 
 def Stat(path):   # pylint: disable=invalid-name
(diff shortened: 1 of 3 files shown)


rohitgirdhar:
I signed it!

tensorflow/python/platform/gfile.py
```diff
@@ -382,13 +382,15 @@ def ListDirectory(directory, return_dotfiles=False):  # pylint: disable=invalid-
   return files
 
 
-def Walk(top, topdown=1, onerror=None):
+def Walk(top, topdown=1, onerror=None, followlinks=False):
```
vrv: I don't think there are very many uses of Walk right now, and it might make more sense to make followlinks the default, so we don't have to pass that everywhere.  Maybe just change line 404 to add followlinks=True and nothing else is needed?

vrv: This was also referenced by Dan in https://github.com/tensorflow/tensorflow/issues/2140#issuecomment-215888562, so I guess that's another vote :)

rohitgirdhar:
@vrv Sounds good, done!

danmane:
Change looks good to me.

tensorflow/python/platform/gfile.py
```diff
@@ -401,7 +401,7 @@ def Walk(top, topdown=1, onerror=None):
     # by lists of all its subdirectories and leaf files.
```
vrv: I would definitely add to the documentation that Walk follows symlinks by default though.

rohitgirdhar:
@danmane @vrv Done

vrv:
great, thanks!

rohitgirdhar:
@vrv: Do you think the failing rnn test could be related to this change?

vrv:
No it shouldn't be, I'm going to merge in some newer changes from internal first, and see if the flakiness goes away.

vrv:

vrv:
If we discover that the symlink loops are common, which I hope they are not, we can change the function to use the os.walk iterator and check whether a symlink causes a loop.

cipri-tom:
Any idea on when this will be merged ? I suppose it belongs to the `tensorboard` repo now, right ?

PR #3351, opened 2016-07-17
Fix gradient of prod op using cumprod

ibab:
This is an initial shot at fixing #2641 (NaNs in the gradient of `tf.reduce_prod` when the input is `0`).
@girving suggested to use two calls to `cumprod` and to multiply the outputs to produce an array containing the products of all values without the current one, like this:

```
input: [0, 1, 2, 3, 4]
cumprod1: [1, 0, 0, 0, 0]
cumprod2 (reverse): [24, 24, 12, 4, 1]
gradient: [24, 0, 0, 0, 0]
```

The difficulty with this is handling the fact that `tf.reduce_prod` can calculate the product over a subset of the tensor dimensions.
When solving this in Python, quite a bit of transposing and reshaping is required.
The idea is to reshape the input into a tensor with shape `[N, M]`, where the first dimension contains all the entries that we reduced over, and the second dimension contains the remaining ones.
I've pushed a Python solution using the approach that makes the reduction op tests pass.
As you can see, the gradient code is a bit complicated.
Maybe someone knows a shortcut that makes this simpler?

There might be other solutions that could be better here:
@benoitsteiner suggested that `tf.cumprod` could be changed to take a list of dimensions to scan over instead of a single axis. This would bring it more in line with the reduction ops.
I think this could be implemented without modifying Eigen, by using various Eigen methods to reshape and transpose the input.

### DIFF
--- a/tensorflow/python/kernel_tests/reduction_ops_test.py
+++ b/tensorflow/python/kernel_tests/reduction_ops_test.py
@@ -435,7 +435,7 @@ class ProdReductionTest(tf.test.TestCase):
   def testGradient(self):
     s = [2, 3, 4, 2]
     # NOTE(kearnes): divide by 20 so product is a reasonable size
-    x = np.arange(1.0, 49.0).reshape(s).astype(np.float32) / 20.
+    x = np.arange(0.0, 48.0).reshape(s).astype(np.float32) / 20.
     with self.test_session():
       t = tf.convert_to_tensor(x)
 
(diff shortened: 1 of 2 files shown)


tensorflow/python/kernel_tests/reduction_ops_test.py
```diff
@@ -435,7 +435,7 @@ def testFloatReduce3D(self):
   def testGradient(self):
     s = [2, 3, 4, 2]
     # NOTE(kearnes): divide by 20 so product is a reasonable size
-    x = np.arange(1.0, 49.0).reshape(s).astype(np.float32) / 20.
+    x = np.arange(0.0, 48.0).reshape(s).astype(np.float32) / 20.
```
girving: We should test that it works with the zero at the beginning, middle, and end.

girving:
Looks good besides the testing comment.  I don't see any obvious simplifications.

tensorflow/python/kernel_tests/reduction_ops_test.py
ibab: Good idea, I'll implement that together with an array that doesn't have a zero, as arrays with a zero have a very simple gradient which could prevent bugs from showing up.

girving:
Jenkins, test this please.

girving:
Awesome, nice to have this fixed!

ibab:
Cheers 🎉

prb12:
#3957 seems to be related to this commit.

PR #3395, opened 2016-07-19
standardize BUILD files format using buildifier

Mistobaan:
I just run `buildifier -mode=fix`find . -iname BUILD -type f`. 

`go get github.com/bazelbuild/buildifier` to install the tool.

I like to have one format for the build files. 
What is the general feeling about this ?

### DIFF
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -824,6 +824,7 @@ cc_library(
     hdrs = [
         "lib/core/blocking_counter.h",
         "lib/core/refcount.h",
+        "lib/gif/gif_io.h",
         "lib/gtl/edit_distance.h",
         "lib/gtl/int_type.h",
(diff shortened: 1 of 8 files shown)


martinwicke:
This is awesome. @gunan @caisq can we make this part of the ci-sanity? It causes all kinds of conflicts.

gunan:
Maybe without the autofix option, but I agree that we should have this.

gunan:
Jenkins, test this please.

martinwicke:
Actually, I was thinking about adding a test to cl-sanity that simply runs
buildifier and fails if any changes are suggested. The submitter can then
run it themselves to make sure it passes.

Mistobaan:
I think a CI check would be sufficient, you can run the command with `-mode=check` to only check if the files are well formatted. 

buildifier -mode=check `find . -iname BUILD -type f`

PR #3409, opened 2016-07-20
Broadcast doc for ops in math_ops.cc

siddharth-agrawal:
Document broadcasting behavior for each binary elementwise operation, as discussed in #508.

### DIFF
--- a/tensorflow/core/ops/math_ops.cc
+++ b/tensorflow/core/ops/math_ops.cc
@@ -467,7 +467,8 @@ REGISTER_OP("Add")
     .Doc(R"doc(
 Returns x + y element-wise.
 
-*NOTE*: Add supports broadcasting. AddN does not.
+*NOTE*: Add supports broadcasting. AddN does not. More about broadcasting
+[here](http://docs.scipy.org/doc/numpy/user/basics.broadcasting.html)
 )doc");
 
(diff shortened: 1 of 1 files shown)


martinwicke:
I think there's a bunch of backticks missing, can you look at https://www.tensorflow.org/versions/r0.9/how_tos/documentation/index.html? Most of it is not relevant here, but the formatting for functions names should be consistent.

siddharth-agrawal:
@martinwicke Are you referring to op names? I'm following the documentation [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc#L470).

xmbrst:
Yep, @martinwicke is referring to op names. BatchMatMul in [math_ops.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc#L129) illustrates the right style: op names, class names, etc should be in backticks. It's true the style isn't applied completely consistently at the moment, but we should try to be consistent going forward.

siddharth-agrawal:
Updated the PR.

xmbrst:
Admins, this PR looks good to me.

martinwicke:
Jenkins, test this please.

PR #3412, opened 2016-07-20
Update the quantization how-to

bafu:
- Prevent the compiling errors of the label_image program.
- Fix the parameters of the label_image program.
- Fix the broken image links.


martinwicke:
We shouldn't check in these images. @petewarden, talk to me if images on tensorflow.org need to be updated, or if it's just the links.

martinwicke:
The images are on tensorflow.org/images/quantization0.png (etc.), it's just the links that need fixing, image links shouldn't be relative if we want them to work on github.

bafu:
Hi @martinwicke , thank you for the feedback.  I have updated the image links from the relative to absolute ones, and remove images from `tensorflow/g3doc/images/`.  Please let me know if there is any place needed to improve.

danmane:
@martinwicke / @petewarden Please take a look.

PR #3415, opened 2016-07-20
Speed up retraining times  approximately 2x

sergejsrk:
Speed up is achieved by:
1. saving (and especially loading) bottleneck arrays in binary files instead of .txt files
2. caching already calculated bottleneck arrays in RAM, instead of reading them every time from disk

P.S.
2x speed up is achieved on SSD disks, on HDD disks speed up is even greater

### DIFF
--- a/tensorflow/examples/image_retraining/retrain.py
+++ b/tensorflow/examples/image_retraining/retrain.py
@@ -406,6 +406,9 @@ def read_list_of_floats_from_file(file_path):
   return list(s)
 	
 
+bottleneck_path_2_bottleneck_values = {}
+
+	
 def get_or_create_bottleneck(sess, image_lists, label_name, index, image_dir,
                              category, bottleneck_dir, jpeg_data_tensor,
                              bottleneck_tensor):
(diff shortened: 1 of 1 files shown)


sergejsrk:
I signed it!

martinwicke:
Could you reformat this to match pep8, except for the indent, which should be 2 spaces?

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -371,6 +372,44 @@ def ensure_dir_exists(dir_name):
...
+
+  Args:
+    list_of_floats: List of floats we want to write to a file.
+    file_path: Path to a file where list of floats will be stored.
+
+  """
+
+  s = struct.pack('d' * BOTTLENECK_TENSOR_SIZE, *list_of_floats)
+
+  f = open(file_path, 'wb')
```
Mistobaan: use `with open(fpath) as f:`

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -371,6 +372,44 @@ def ensure_dir_exists(dir_name):
...
+  """Reads list of floats from a given file.
+
+  Args:    
+    file_path: Path to a file where list of floats was stored.
+  Returns:
+    Array of bottleneck values (list of floats).
+
+  """
+
+  f = open(file_path, 'rb')
```
Mistobaan: use `with open(fpath) as f:`

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -371,6 +372,44 @@ def ensure_dir_exists(dir_name):
...
+  s = []
+  s = struct.unpack('d' * BOTTLENECK_TENSOR_SIZE, f.read())
+  f.close()
+  # print (list(s))
```
Mistobaan: remove comment

sergejsrk:
Reformated according to PEP8

sergejsrk:
One more thing considering creation of bottlenecks. 

If a new version of retrain.py is run in a bottleneck directory, that contains bottlenecks files created with a previous version of retrain.py then retrain.py will crash (it will try to read list of floats from .txt file and fail). 

To prevent this problem we can either:

- delete all bottlenecks created with a previous version of retrain.py (it is remove .txt files from bottleneck dir)
- or modify get_bottleneck_path function, so it appends not ".txt", but ".bin" postfix.

Which one is better option?

danmane:
@petewarden can you review this please?

petewarden:
Jenkins, test this please.

petewarden:
I've reviewed the changes, and they look good to me! Thanks for contributing this @sergejsrk

PR #3416, opened 2016-07-20
Make the default command use a named container

sushabhardy:
It seems that it's a more common workflow than I assumed, and should be relatively foolproof.


vrv:
I'm not sure what this change is -- it looks like an existing commit  -- if this is supposed to be a real change, can send it using the normal github pull request process?

PR #3417, opened 2016-07-20
Prevent evaluating stale output node.

parvizp:
When quantizing to 8-bit the Identity ops are removed ("spliced"). For graph's with a parameter-less output node like SoftMax, there is no issue. However, if the graph's output node received a constant input (i.e. a bias add with weights) the output node referenced [here](https://github.com/parvizp/tensorflow/blob/502ecd2c6c324a1c5c3f7b195c36dbafb45dafb3/tensorflow/contrib/quantization/tools/quantize_graph.py#L318) will be stale (still has Identity input) and result in a error.

### DIFF
--- a/tensorflow/contrib/quantization/tools/quantize_graph.py
+++ b/tensorflow/contrib/quantization/tools/quantize_graph.py
@@ -328,6 +328,8 @@ class GraphRewriter(object):
         self.quantize_nodes_recursively(output_node)
     elif self.mode == "eightbit":
       self.set_input_graph(self.remove_unneeded_nodes(self.input_graph))
+      output_nodes = [self.nodes_map[output_node_name]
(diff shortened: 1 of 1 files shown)


parvizp:
I signed it!
The cooperation is NanoWatt Design.

danmane:
We should get the CLABot happy before we go forward with this CL. @parvizp can you try poking it again?

parvizp:
I signed it!

I had the email address used on the commits and CLA as secondary email on my GitHub profile. I changed to be primary in case that helps.

parvizp:
I also need to write a test.

rmlarsen:
@parvizp Any update?

vrv:
Closing due to lack of CLA + lots of time passing

PR #3428, opened 2016-07-20
Fix eigen hash detection in download_dependencies.sh

satok16:
Fixes #3427

PR #3452, opened 2016-07-21
Update download_dependencies.sh for changed Eigen config

ibab:
The `download_dependencies.sh` in `contrib/make` also needs to be updated to work with the Eigen config changes from #3288.
Thanks to @apcode for pointing this out!


vrv:
Is #3428 sufficient?

ibab:
Oh, I missed #3428. It looks sufficient.
I'll close this one, then.

Issue #3457, opened 2016-07-22
Problem in distributed tensorflow demo?


perhapszzy:
### Environment info

Operating System:
Kubernetes (using the 0.8.0 server image)

Installed version of CUDA and cuDNN: 
No

When I start the [mnist_replica demo](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/dist_test/python/mnist_replica.py), it stuck after `prepare_or_wait_for_session`

But if I run `start_queue_runners` for both chief worker and all other workers, it runs successfully. (By removing `and is_chief` part [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/dist_test/python/mnist_replica.py#L224) and [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/dist_test/python/mnist_replica.py#L193))

Is this a bug in the demo?

perhapszzy:
made a pr https://github.com/tensorflow/tensorflow/pull/3458 to fix this

perhapszzy:
Update： If I start workers this way:
```
  if FLAGS.job_name == 'ps':
    server.join()
  else:
   # Do training
```

Instead of connecting to worker each time:
```
sess = sv.prepare_or_wait_for_session(FLAGS.worker_grpc_url)
```

I can successfully run the code. Any explanation about this?

perhapszzy:
Way to repeat the problem:
On my own laptop run the following three cmd:
```
python mnist_replica.py --worker_grpc_url="grpc://43.254.53.118:30001" --worker_index=0 --sync_replicas
python mnist_replica.py --worker_grpc_url="grpc://43.254.53.118:30002" --worker_index=1 --sync_replicas
python mnist_replica.py --worker_grpc_url="grpc://43.254.53.118:30003" --worker_index=2 --sync_replicas
```
The log for worker1 is:
```
Worker GRPC URL: grpc://43.254.53.118:30001
Worker index = 0
Number of workers = 3
Worker 0: Initializing session...
Worker 0: Session initialization complete.
Starting chief queue runner and running init_tokens_op
Training begins @ 1469418277.317969
1469418277.659315: Worker 0: training step 1 done (global step: 0)
1469418277.961008: Worker 0: training step 2 done (global step: 0)
1469418278.261021: Worker 0: training step 3 done (global step: 0)
1469418292.345302: Worker 0: training step 4 done (global step: 0)
```

The log for worker2 is:
```
[...]

PR #3458, opened 2016-07-22
fix sync update in distributed mnist demo

perhapszzy:


perhapszzy:
I signed it!

yifeif:
@caisq could you take a look?

caisq:
Looking

caisq:
Adding @jmchen-g

perhapszzy:
https://github.com/tensorflow/tensorflow/issues/3457 provided more information

danmane:
@caisq / @jmchen-g should we go ahead to merge this?

jmchen-g:
This fix is not right. The chief queue runner can only run by the chief. This seems to expose a bug somewhere but we have not figured it yet. We are in the middle of changing the sync_rep_optimizer so will come back to this after that change.

perhapszzy:
Any updates on the sync mode?

rmlarsen:
@jmchen-g any updates on this?

jmchen-g:
Most of the updates are implemented internally and after they are verified, we will push it outside. That will take some time though. We can close this PR for now as the fix is not a real fix but happens to get around some issues in the current implementation. The new implementation should be more robust and easier to debug.

renato2099:
hi @jmchen-g any updates on this?

jmchen-g:
@caisq, any update on this please?

PR #3490, opened 2016-07-25
R0.9

moontree:


martinwicke:
The link changes are actually backports, but those are fixes only for 0.9. 

Can you add a commit reverting  tensorflow/g3doc/tutorials/wide/index.md and  tensorflow/g3doc/tutorials/wide-n-deep/index.md?

vrv:
I'm pretty sure this is an accident -- something about github's instructions end up making users try to push branches back to master.

PR #3516, opened 2016-07-26
Fix an error message.

ClumsyLee:
Minimum tensor rank should be `dim + 1` instead of `dim`.

Example:

``` python
tf.argmin(x, 1)  # x is a 1-d vector.
```

before:

> InvalidArgumentError: Minimum tensor rank: 1 but got: 1

after:

> InvalidArgumentError: Minimum tensor rank: 2 but got: 1


yifeif:

PR #3521, opened 2016-07-27
Clarify docs for tensorflow/python/ops/rnn_cell.py

adamcrume:
What the docs and code refer to as a "cell" corresponds to what the literature
calls a layer of cells/units. This clarifies that e.g. RNNCell is an array of
cells/units.

### DIFF
--- a/tensorflow/python/ops/rnn_cell.py
+++ b/tensorflow/python/ops/rnn_cell.py
@@ -87,6 +87,11 @@ def _state_size_with_prefix(state_size, prefix=None):
 class RNNCell(object):
   """Abstract object representing an RNN cell.
 
+  The definition of cell in this package differs from the definition used in the
+  literature. In the literature, cell refers to an object with a single scalar
+  output. The definition in this package refers to a horizontal array of such
+  units.
+
   An RNN cell, in the most abstract setting, is anything that has
   a state and performs some operation that takes a matrix of inputs.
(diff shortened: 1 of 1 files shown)


adamcrume:
Sorry, used my personal email address the first time around.  Fixed.

tensorflow/python/ops/rnn_cell.py
```diff
@@ -85,9 +85,11 @@ def _state_size_with_prefix(state_size, prefix=None):
 
 
 class RNNCell(object):
-  """Abstract object representing an RNN cell.
+  """Array of RNN cells.
```
ebrevdo: sounds like you want to change the notation of cell to mean something else here.  however, we already have a meaning in this code for "Cell" and it's:

An RNN cell, in the most abstract setting, is anything that has a state and performs some operation that takes a matrix of inputs.

So - don't change all the names of RNNCell objects to mean "cell" as something else.  An RNNCell creates one unit of computation, it is not an "Array of RNN cells".

Rather, add a paragraph explaining what the word "cell" means in the literature and how that differs from what we mean by Cell.  That's it.

adamcrume: I was only trying to make the usage match the definition used by the rest of the community, including papers linked to in this file.  I'll change it as per your request, though.

PR #3522, opened 2016-07-27
fix for BM_UNARY test

kashif:
here is a fix for the test. Can you kindly review @rryan ?

### DIFF
--- a/tensorflow/core/kernels/cwise_ops_test.cc
+++ b/tensorflow/core/kernels/cwise_ops_test.cc
@@ -23,8 +23,8 @@ limitations under the License.
 
 namespace tensorflow {
 
-// Creates a Graph which applies a unary "func" on a 3D float tensor
-// of "num" elements.
+// Creates a Graph which applies a unary "func" on a 3D tensor of
+// type T with "num" elements.
 template <typename T>
 static Graph* Unary(const string& func, int num, DataType dtype) {
   Graph* g = new Graph(OpRegistry::Global());
--- a/tensorflow/core/kernels/cwise_ops_test.cc
+++ b/tensorflow/core/kernels/cwise_ops_test.cc
@@ -25,20 +25,12 @@ namespace tensorflow {
 
 // Creates a Graph which applies a unary "func" on a 3D float tensor
 // of "num" elements.
+template <typename T>
 static Graph* Unary(const string& func, int num, DataType dtype) {
   Graph* g = new Graph(OpRegistry::Global());
   Tensor data(dtype, TensorShape({64, 64, num / (64 * 64)}));
   CHECK_GT(data.NumElements(), 0);
-  switch (dtype) {
-    case DT_DOUBLE:
-      data.flat<double>().setRandom();
-    case DT_COMPLEX64:
-      data.flat<complex64>().setRandom();
-    case DT_COMPLEX128:
-      data.flat<complex128>().setRandom();
-    default:
-      data.flat<float>().setRandom();
-  }
+  data.flat<T>().setRandom();
   test::graph::Unary(g, func, test::graph::Constant(g, data), 0);
   return g;
 }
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/cwise_ops_test.cc
```diff
@@ -29,7 +29,16 @@ static Graph* Unary(const string& func, int num, DataType dtype) {
   Graph* g = new Graph(OpRegistry::Global());
   Tensor data(dtype, TensorShape({64, 64, num / (64 * 64)}));
   CHECK_GT(data.NumElements(), 0);
-  data.flat<float>().setRandom();
+  switch (dtype) {
```
rryan: I think it would be less error prone if we made this a template function of T -- then the C++ type could be part of the BM_UNARY macro below. You can see an example of this in the matmul benchmark:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/matmul_op_test.cc#L23

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/matmul_op_test.cc#L23

kashif: ok yes since there were only two types we were checking... but yes i prefer templates too. Sending a patch soon.

tensorflow/core/kernels/cwise_ops_test.cc
```diff
@@ -49,21 +41,23 @@ static int RowsAndColsArg(int r, int c) { return r * kRows + c; }
...
-    const int64 tot = static_cast<int64>(iters) * num;            \
-    testing::ItemsProcessed(tot);                                 \
-    testing::BytesProcessed(tot * sizeof(float));                 \
-    test::Benchmark(#DEVICE, Unary(#FUNC, num, TYPE)).Run(iters); \
-  }                                                               \
+#define BM_UNARY(DEVICE, FUNC, T, TYPE)                              \
+  static void BM_##DEVICE##_##FUNC_##TYPE(int iters, int num) {      \
+    const int64 tot = static_cast<int64>(iters) * num;               \
+    testing::ItemsProcessed(tot);                                    \
+    testing::BytesProcessed(tot * sizeof(float));                    \
```
rryan: I think this should probably be ```sizeof(T)```

tensorflow/core/kernels/cwise_ops_test.cc
```diff
@@ -25,11 +25,12 @@ namespace tensorflow {
 
 // Creates a Graph which applies a unary "func" on a 3D float tensor
 // of "num" elements.
```
rryan: while you're here -- could you please update the comment to say 'on a 3D tensor of type T with "num" elements' ?

tensorflow/core/kernels/cwise_ops_test.cc
kashif: indeed! Funnly enough the test passed which is kind of confusing... I am actually having issues testing the gpu part on my mac (it crashes while loading the curand dynamic library) and on ubuntu it didnt fail so I don't know how to see a failure...

What bazel command do you use to test?

rryan: Yea, I think this wouldn't be covered by any test -- it's simply used to report the number of bytes processed (for calculating throughput of the benchmark I think?). 

Annoyingly, the benchmark isn't run at all by the test suite! You have to build the test binary then run the test binary with the ```--benchmarks=all``` flag.

Something like:
```
$ bazel build //tensorflow/core/kernels:cwise_ops_test
$ ./bazel-bin/tensorflow/core/kernels/cwise_ops_test --benchmarks=all
```

rryan:
The code LGTM in its current state. I'm not sure if I have permission to verify the patch for Jenkins but here goes:

tensorflow/core/kernels/cwise_ops_test.cc
kashif: ah great!! that helps me tons!

kashif:
thanks again! the tests now run on the gpu on the mac! I also fixed an issue with the function names of the benchmark... now it shows them:

```
Benchmark                                                   Time(ns) Iterations
-------------------------------------------------------------------------------
BM_cpu_Floor_DT_FLOAT/4096                                     26842      27131	 610.4MB/s 152.6M items/s
BM_cpu_Floor_DT_FLOAT/32768                                    95100       7409	 1378.3MB/s 344.6M items/s
BM_cpu_Floor_DT_FLOAT/262144                                  259694       2739	 4037.7MB/s 1009.4M items/s
BM_cpu_Floor_DT_FLOAT/1048576                                 847080        700	 4951.5MB/s 1237.9M items/s
...
```

rryan:
> thanks again! the tests now run on the gpu on the mac! I also fixed an issue with the function names of the benchmark... now it shows them:

Thanks!

rryan:
Looks like @tensorflow-jenkins doesn't like me :) @vrv could you trigger a build please?

yifeif:
Thanks @rryan and @kashif! @tensorflow-jenkins test this please

yifeif:

PR #3542, opened 2016-07-28
Fix go build errors

damienstanton:
Two minor changes:
- enum declaration should use C++ syntax
- swig can't port `int64_t` to Go's type system. `long long` used instead.

### DIFF
--- a/tensorflow/core/client/tensor_c_api.cc
+++ b/tensorflow/core/client/tensor_c_api.cc
@@ -132,7 +132,7 @@ struct TF_Tensor {
   TensorBuffer* buffer;
 };
 
-TF_Tensor* TF_NewTensor(TF_DataType dtype, const int64_t* dims, int num_dims,
+TF_Tensor* TF_NewTensor(TF_DataType dtype, const long long* dims, int num_dims,
                         void* data, size_t len,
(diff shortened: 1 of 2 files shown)


yifeif:

yifeif:

damienstanton:
@tensorflow-jenkins test this, pelase

yifeif:

zheng-xq:
Adding @josh11b to review. 

The files had all been moved to tensorflow/c

josh11b:
- We really want a 64-bit int type, not whatever long long happens to be
- The whole point of this file is to use C not C++ so the enum change is counter productive.

damienstanton:
@josh11b fair point, but I was not able to use the Go API bindings until I made these changes. I should probably take a different approach and try to modify the swig helpers rather than the C API itself.

danmane:
@damienstanton should we keep this PR open for you to try a different approach, or close and wait for a fresh PR?

damienstanton:
@danmane since I don't want my branch to get too stale, let's close this and I will open a new PR down the line. Thanks!

danmane:
OK, look forward to the new one :)

PR #3564, opened 2016-07-29
Update index.md

ssyrain:
I just test it as I want to assign a, b and c to gpu3. Something wrong happened.
if the code "c = tf.matmul(a, b)" is not include in "with tf.device('/cpu:0')", then even you find b and a are assigned to cpu0，they just be assigned default. Because c is assigned to cpu0 default. a and b are used by c, so they are assigned to cpu0.


ssyrain:
I signed it !!!!!!!!!!!

yifeif:
Thanks @ssyrain!

PR #3566, opened 2016-07-29
Extended adjust_hue and adjust_saturation to work on 4D tensors

ppries:
I noticed that adjust_hue and adjust_saturation only worked on single images (3D tensors), but I couldn't really see a reason for that, so I extended it to work on 4D. I don't know if you want this in the main repo?

I based the work on the r0.9 branch because that's what I'm using, but I can do it for the master branch if that is more suitable.

### DIFF
--- a/tensorflow/python/ops/image_ops.py
+++ b/tensorflow/python/ops/image_ops.py
@@ -252,13 +252,12 @@ def _slice_channels(image):
     list of tensor, one for each channel.
   """
 
-  image_shape = image.get_shape()
-  image_rank = len(image_shape)
+  image_rank = array_ops.rank(image)
 
   begin = [0] * (image_rank - 1)
   size = [-1] * (image_rank - 1) + [1]
 
-  channels = [array_ops.slice(image, begin + [i], size) for i in range(image_shape[-1])]
+  channels = [array_ops.slice(image, begin + [i], size) for i in range(3)]
   return channels
 
 
@@ -1152,7 +1151,7 @@ def adjust_hue(image, delta, name=None):
     # floating point number since delta is [-0.5, 0.5].
     hue = math_ops.mod(hue + (delta + 1.), 1.)
 
-    hsv_altered = array_ops.concat(len(image.get_shape()) - 1, [hue, saturation, value])
+    hsv_altered = array_ops.concat(array_ops.rank(image) - 1, [hue, saturation, value])
     rgb_altered = gen_image_ops.hsv_to_rgb(hsv_altered)
 
     return convert_image_dtype(rgb_altered, orig_dtype)
@@ -1224,7 +1223,7 @@ def adjust_saturation(image, saturation_factor, name=None):
     saturation *= saturation_factor
     saturation = clip_ops.clip_by_value(saturation, 0.0, 1.0)
 
-    hsv_altered = array_ops.concat(len(image.get_shape()) - 1, [hue, saturation, value])
+    hsv_altered = array_ops.concat(array_ops.rank(image) - 1, [hue, saturation, value])
     rgb_altered = gen_image_ops.hsv_to_rgb(hsv_altered)
 
     return convert_image_dtype(rgb_altered, orig_dtype)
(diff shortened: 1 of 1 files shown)


ppries:
CLA signed.

ppries:
I just noticed pretty much all of the methods has a check for 3D image in them. So is this a design choice or?

yifeif:
@martinwicke could you take a look at this one?

tensorflow/python/ops/image_ops.py
```diff
@@ -242,6 +242,26 @@ def _CheckAtLeast3DImage(image):
...
+  """Slices a tensor into one tensor for each 'channel' (the last dimension).
+
+  Args:
+    images: Tensor containing the images to be sliced.
+
+  Returns:
+    list of tensor, one for each channel.
+  """
+
+  image_shape = image.get_shape()
```
martinwicke: Can you use tf.rank here? That would work also on tensors whose rank not statically defined.

tensorflow/python/ops/image_ops.py
```diff
@@ -1126,15 +1146,13 @@ def adjust_hue(image, delta, name=None):
...
-    saturation = array_ops.slice(hsv, [0, 0, 1], [-1, -1, 1])
-    value = array_ops.slice(hsv, [0, 0, 2], [-1, -1, 1])
+    hue, saturation, value = _slice_channels(hsv)
 
     # Note that we add 2*pi to guarantee that the resulting hue is a positive
     # floating point number since delta is [-0.5, 0.5].
     hue = math_ops.mod(hue + (delta + 1.), 1.)
 
-    hsv_altered = array_ops.concat(2, [hue, saturation, value])
+    hsv_altered = array_ops.concat(len(image.get_shape()) - 1, [hue, saturation, value])
```
martinwicke: Also tf.rank elsewhere

martinwicke:
This should be a patch against master -- we don't really want to backport features at this point. The code hasn't changed much so it will probably work unaltered against master. I have some comments on the patch itself (use tf.rank), otherwise, could you write some unit tests that exercise the newly enabled behavior?

ppries:
I've changed the appropriate lines with calls to tf.rank instead. I will have to assume there will always be three channels, then -- but I guess that's fine?

I don't have much experience with unit testing, but I will take a look at it later this week or start of next week. Any recommendations are welcome.

Should I also change the other ops to be usable on batches of images as well?

I will continue working on this branch for now. Once we agree it's ready I'll close this pull request, merge with master, and create a new one.

tensorflow/python/ops/image_ops.py
```diff
@@ -252,13 +252,12 @@ def _slice_channels(image):
...
 
-  image_shape = image.get_shape()
-  image_rank = len(image_shape)
+  image_rank = array_ops.rank(image)
 
   begin = [0] * (image_rank - 1)
   size = [-1] * (image_rank - 1) + [1]
 
-  channels = [array_ops.slice(image, begin + [i], size) for i in range(image_shape[-1])]
+  channels = [array_ops.slice(image, begin + [i], size) for i in range(3)]
```
martinwicke: You can use tf.unpack(image, axis=image_rank-1) instead.

martinwicke:
Sounds good. 

I left a comment about the 3 channels issue.

If you would remove the restriction to 3D from other ops, that would be amazing.

As for unit tests: Please add tests that make sure that these ops actually work as expected when applied to batches. You can probably simply extend the existing tests to add cases for higher input rank. The key is to use predictable input data with known outputs, and keep the tests fast.

ppries:
Just to give an update: I'm working on removing the 3D restriction from the other ops, but I'm having difficulty doing it in a way that would work with tensors whose shape is not statically defined. Right now I'm thinking something like creating a flat tensor of ones (or whatever fits the op) of same size as the rank of the input tensor, and then change whatever index needs to be different for that op with tf.scatter_update, but I'm not sure how to get scatter_update to work with constant tensors. I'm sure I will figure it out though, just wanted to inform you I haven't forgotten about this.

tensorflow/python/ops/image_ops.py
ppries: I was looking for this! Turns out it didn't exist (with the axis argument) in r0.9, but very nice that it was added in the master branch.

ppries:
I'm closing this, since I've opened a new pull request for the master branch instead of r0.9 ( #4150).

PR #3573, opened 2016-07-29
update os_setup for r0.10

jendap:
mac directories are changed to be similar to linux and to avoid collision of cpu/gpu mac wheel files


jendap:
@gunan @caisq

PR #3597, opened 2016-08-01
Use tf.contrib.layers.conv2d instead of learn.ops.conv2d which is dep…

yangwang166:
Use tf.contrib.layers.conv2d instead of learn.ops.conv2d which is deprecated and the code will be removed after 2016-08-15.


terrytangyuan:

Issue #3549, opened 2016-07-28
Android: NoOpKernel was registered to support Op 'TensorflowArray' DynamicRNN


lingz:
I'm trying to load a graph I trained in python, which includes only using dynamicRNN, and the most basic RNN cell. 

When I try execute the graph on the Android, I get the following error message:

```
Invalid argument: No OpKernel was registered to support Op 'TensorArray' with these attrs
	 [[Node: c96ce52cccd5408aad2fb356a8246023/RNN/TensorArray_1 = TensorArray[clear_after_read=true, dtype=DT_FLOAT, dynamic_size=false, tensor_array_name="c96ce52cccd5408aad2fb356a8246023/RNN/dynamic_rnn/input"](c96ce52cccd5408aad2fb356a8246023/RNN/unpack)]]
```

### Environment info
Operating System:
Ubuntu x64 14.04
Android 6.0

Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):

If installed from binary pip package, provide:

1. Which pip package you installed.
2. The output from `python -c "import tensorflow; print(tensorflow.__version__)"`.

If installed from source, provide 

1. The commit hash (`git rev-parse HEAD`)
tag v0.9.0
2. The output of `bazel version`
0.3.0

### Steps to reproduce
1. Create RNN model in python
2. Compile it with tensorflow_android_lib targeting armv7
3. Try run the model on android

### What have you tried?
1. 

### Logs or other output that would be helpful
(If logs are large, please upload as attachment).

lingz:
It seems it that the RNN ops are default not included in the `android_extended_ops_group` under /tensorflow/cc/kernel/BUILD.

a hack is adding the operations manually but the more correct way seems to use selective_registration.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/selective_registration.h

Is there an easy way to determine the names of all the ops I would need, perhaps writing a grep on parsing a txt version of the saved graph?

lingz:
And on the same note, is there an example of how to use selective_registration, in particular how to differentiate between `SHOULD_REGISTER_OP_KERNEL` `SHOULD_REGISTER_OP` and `SHOULD_REGISTER_OP_GRADIENT`

andrewharp:
So long as the op builds and runs correctly, it's fine to add it to the default ops file group (we've just been adding them on an as-needed basis now that we have a way to manage binary size).

[...]

PR #3602, opened 2016-08-01
Changed visibility of protos_cc to public

lingz:
The only tensorflow package that is private and preventing us from creating custom kernels (with a custom set of operations), is the protos_cc. If we make this public, we can make our own kernels outside of the tensorflow repo, which means for Android/iOS we can make compact libraries with just the right amount of kernel ops.

There seems to be another approach which is `SELECTIVE_REGISTRATION`, but firstly according to #3549 no one has managed to provide a working example for it (tried myself but was not able to yet), and it also seems to be Android only. For the meantime, it seems the best method is to define our own kernel and therefore our own tensorflow_lib bazel files, but to do this we need to changes protos_cc to public.

### DIFF
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -1027,6 +1027,7 @@ tf_cuda_library(
 cc_library(
     name = "protos_cc",
     deps = ["//tensorflow/core/platform/default/build_config:protos_cc"],
+    visibility = ["//visibility:public"],
 )
 
 tf_cuda_library(


zheng-xq:
Adding @josh11b to review this change.

keveman:
Sorry for the delay. `//tensorflow/core:framework_headers_lib` is already public and should give you all the headers you need to build a custom op. Is that not working for some reason?

lingz:
The use case here is not for building a custom op, rather it is a building a custom kernel, containing a specific collection of ops. The current kernels all seem to depend on `protos_cc`.

keveman:
My previous question applies equally to custom kernels. By current kernels, if you mean kernels that are part of core, then sure, all of them have a `bazel` dependency of `protos_cc`. But that doesn't mean they are really using any from `protos_cc`. Custom kernels are supposed to build with just the headers in `framework_headers_lib`. Have you tried building with just that dependency?

lingz:
But isn't `framework_headers_lib` just a collection of headers? Right now I'm including also all the core tensorflow code for example from `//tensorflow/core:android_tensorflow_lib_lite`.

Here's my example build file, do you mean that I can replace all the deps with just `framework_headers_lib`?

```
cc_library(
    name = "myapp_android_tensorflow_lib",
    srcs = if_android(["@org_tensorflow//tensorflow/core:android_op_registrations_and_gradients"]),
    copts = tf_copts(),
    deps = [
        "@org_tensorflow//tensorflow/core:android_tensorflow_lib_lite",
        "@org_tensorflow//tensorflow/core:protos_cc",
        "@org_tensorflow//third_party/eigen3",
        ":ml_android_tensorflow_kernels",
    ],
    alwayslink = 1,
)

cc_library(
    name = "ml_android_tensorflow_kernels",
    srcs = select({
        "@org_tensorflow//tensorflow:android": [
            ":android_myapp_ops",
            "@org_tensorflow//tensorflow/core/kernels:android_core_ops",
            "@org_tensorflow//tensorflow/core/kernels:android_extended_ops",
        ],
        "//conditions:default": [],
    }),
    copts = tf_copts(),
    deps = [
        "@org_tensorflow//tensorflow/core:android_tensorflow_lib_lite",
        "@org_tensorflow//tensorflow/core:protos_cc",
        "@org_tensorflow//third_party/eigen3",
    ],
    alwayslink = 1,
)
```

lingz:
Btw, this is related to issue #3549 which is about how to correctly build a custom collection of ops right now.

keveman:
@lingz Yes, you would only need the headers to build your custom collection of ops and kernels.

rmlarsen:
@lingz Any update on this?

lingz:
Yep so I tried and you can replace everything except android_tf_lib_lite with dramework_headers_lib.

wew2012:
@lingz Could you show me the complete code about how to use SELECTIVE_REGISTRATION feature?

lingz:
I was not able to make it work. I would like to see a working example also.

wew2012:
@keveman I tried to build with just the headers in `framework_headers_lib`. However, the following error occurs:

>bazel-out/arm-linux-androideabi-4.9-v7a-gnu-libstdcpp-fastbuild/genfiles/tensorflow/core/framework/graph.pb.h:9:42: fatal error: google/protobuf/stubs/common.h: No such file or directory
 #include <google/protobuf/stubs/common.h>
>                                          ^
>compilation terminated.

PR #3604, opened 2016-08-01
Added esn cell

m-colombo:
I wrote an Echo State Network cell for my personal need, hope that this can be useful to someone else. I tried to be compliant with the project coding style, let me know if I missed something


m-colombo:
About the CLA: I've signed it as a corporation (Biobeats) and it result in the CLAs page. Do I need to edit my commits in order to have the company email which is in the CLA google group ? I thought it was enough to have it in the github account.

zheng-xq:
Adding @ebrevdo for review.

ebrevdo:
Hi, this is a good contribution to tf.contrib.rnn.  can you move the class there?  Also please add better documentation including literature reference in the doc string, and use contrib.layers instead of _linear.  Do you have any unit tests?

m-colombo:
Moved to tf.contrib. As soon as I can I'll improve the doc (sorry but I'm doing this in the spare time after work).
I've written no unit testing, I just tested it on MackeyGlass dataset. I'm not sure about how to write effective unit testing since it's mostly based on randomness, however if it's needed I can try to write something.

ebrevdo:
Here's a suggestion for now as a smoke test:

1. create inputs with np.random (use a seed) and use a non-random
initializer for the variables
2. create a small ESA cell and run it forward a few time steps & save to
file the output & final state numbers
3. write your test to ensure that the same random seeds & deterministic
initializer lead to the exact same output as you had before (use
self.assertAllClose with the tensor values you saved from step #2)

this will ensure that future changes to ESA cell do not break its numeric
behavior.

m-colombo:
So I improved the docs and added a smoke test using tf.test.TestCase, hope everything's allright :)

ebrevdo:
Will likely get to this on Monday.

rmlarsen:
@ebrevdo when do you think you'll be able to review this?

tensorflow/contrib/rnn/__init__.py
```diff
@@ -24,6 +24,9 @@
 @@TimeFreqLSTMCell
 @@GridLSTMCell
 
+### ESN cells
```
ebrevdo: ### Echo State Networks

tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -661,6 +663,109 @@ def _make_tf_features(self, input_feat):
       freq_inputs.append(cur_input)
     return freq_inputs
 
+class ESNCell(rnn_cell.RNNCell):
+  """Echo State Network Cell.
+
+    Based on http://www.faculty.jacobs-university.de/hjaeger/pubs/EchoStatesTechRep.pdf
+    It doesn't consider output feedback, specifically it models just the reservoir.
```
ebrevdo: This comment, without context to someone who doesn't know about ESNs, makes no sense.  Can you expand?

tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -661,6 +663,109 @@ def _make_tf_features(self, input_feat):
...
+
+    Here a practical guide to use Echo State Networks:
+    http://minds.jacobs-university.de/sites/default/files/uploads/papers/PracticalESN.pdf
```
ebrevdo: Are these links guaranteed to stay good for a Really Long Time?

tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -661,6 +663,109 @@ def _make_tf_features(self, input_feat):
...
+
+    Since at the moment TF doesn't provide a way to compute spectral radius
+    of a matrix the echo state property necessary condition `max(eig(W)) < 1` is approximated
+    scaling the norm 2 of the reservoir matrix which is an upper bound of the spectral radius.
+    See https://en.wikipedia.org/wiki/Matrix_norm#Induced_norm
```
ebrevdo: This probably isn't guaranteed to stay static.  Can you update the link to something like ".../Matrix_norm, the section on Induced norms."

tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -661,6 +663,109 @@ def _make_tf_features(self, input_feat):
...
+      bias_init: initializer for biases
+    """
+    self._num_units = num_units
+    self._leaky = leaky
+    self._activation = activation
+
+    def _wr_initializer(shape, dtype):
+      wr = wr_init(shape, dtype=dtype)
+
+      connectivity_mask = \
```
ebrevdo: backslashes aren't accepted in TF python code; can you find another way to perform the multi-line?  e.g.:

```
connectivity_mask = math_ops.cast(
  ...
  ...)
```

tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -661,6 +663,109 @@ def _make_tf_features(self, input_feat):
...
+        math_ops.cast(
+          math_ops.less_equal(
+            random_ops.random_uniform(shape),
+            connectivity),
+          dtype)
+
+      wr = math_ops.mul(wr, connectivity_mask)
+
+      wr_norm2 = math_ops.sqrt(
```
ebrevdo: if this can fit on one line, reduce to one line.  same elsewhere.

tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -661,6 +663,109 @@ def _make_tf_features(self, input_feat):
...
+        Raises:
+          ValueError: if `inputs` or `state` tensor size mismatch the previously provided dimension.
+          """
+
+    input_size = inputs.get_shape().as_list()[1]
+    dtype = inputs.dtype
+
+    with vs.variable_scope(scope or type(self).__name__):  # "ESNCell"
+
+      win = vs.get_variable("InputMatrix", [input_size, self._num_units], dtype=dtype,
```
ebrevdo: it is often leads to faster computation e.g. on GPUs to create a single matrix, e.g., InputMatrix equivalent to concatenating ReservoirMatrix, perform a single matmul, and then split the results.

In fact, here you should probably just use contrib.layers.linear instead of calling the matmul + get_variable yourself.

tensorflow/contrib/rnn/python/ops/rnn_cell_test.py
```diff
@@ -0,0 +1,52 @@
...
+      state, _ = cell(inputs[i, :, :], state)
+
+    with self.test_session() as S:
+      S.run(tf.initialize_all_variables())
+      final_states = S.run([state])
+
+    expected_final_states = [[[0.56619251, 0.90794069, -0.99339789, -0.38217574],
+                              [0.97225142, 0.98931664, -0.65432256, -0.66463804]]]
+
+    self.assertAllClose(final_states, expected_final_states)
```
ebrevdo: nit: newline at end of file.

tensorflow/contrib/rnn/python/ops/rnn_cell_test.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
ebrevdo: we have a test in contrib/rnn/python/kernel_tests/rnn_cell_test.  you should add your test class there instead of here.

tensorflow/contrib/rnn/python/ops/rnn_cell_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+    cell = rnn_cell.ESNCell(n_units)
+    inputs = tf.get_variable("inputs", [input_length, batch_size, input_size], initializer=random_normal_initializer())
+
+    state = cell.zero_state(batch_size, inputs.dtype)
+    for i in range(input_length):
+      if i == 1 : tf.get_variable_scope().reuse_variables()
+      state, _ = cell(inputs[i, :, :], state)
+
+    with self.test_session() as S:
```
ebrevdo: replace "S" with "sess" or "session"

tensorflow/contrib/rnn/python/ops/rnn_cell_test.py
```diff
@@ -0,0 +1,52 @@
...
+  def test_esn_dynamics(self):
+    """ A simple smoke test """
+
+    input_size = 4
+    input_length = 4
+    batch_size = 2
+    n_units = 4
+
+    cell = rnn_cell.ESNCell(n_units)
+    inputs = tf.get_variable("inputs", [input_length, batch_size, input_size], initializer=random_normal_initializer())
```
ebrevdo: just use a numpy array here, like np.random.randn...

tensorflow/contrib/rnn/python/ops/rnn_cell_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+    state = cell.zero_state(batch_size, inputs.dtype)
+    for i in range(input_length):
+      if i == 1 : tf.get_variable_scope().reuse_variables()
```
ebrevdo: if i > 0:

tensorflow/contrib/rnn/python/ops/rnn_cell.py
m-colombo: There's no guarantees on those links lifetime, unfortunately on arxiv I didn't find paper introducing esn, just using them.

danmane:
@ebrevdo PTAL

danmane:

ebrevdo:
After looking at the contribution guidelines, I realized that without an internal maintainer for this architecture, we cannot accept this type of submission. It is probably a better fit in its own dedicated repository.

ebrevdo:
@vrv, can you confirm that my understanding is correct?

vrv:
Yeah, this is a cool contribution, but best for an external repository -- if it became critical for state-of-the-art models, we'd probably want to integrate it then.  In the meantime, feel free to point us at where you are maintaining this and we can link to it somewhere!

m-colombo:
I took the contribution out of TF and put in a [repository](https://github.com/m-colombo/Tensorflow-EchoStateNetwork) by it self.

PR #3606, opened 2016-08-02
Restore a working CMake build on Linux

mrry:
Updates CMake build files as follows:
- Adds giflib dependency.
- Adds support for the new C++ codegen API.
- Adds support for the `tensorflow/core/debug` libraries.


mrry:
N.B. The CMake Docker build is broken, but this works locally, on a machine with protobuf correctly installed.

zheng-xq:
Adding @vrv for review. Please feel free to redirect if necessary.

vrv:
LGTM

PR #3609, opened 2016-08-02
bug fix: the benchmark only run FLAGS.num_batches-1 times

yanchen036:
I found that if I set num_batches = 100, then it only runs 99 batches.
Because i is [0, 110), and it starts to count the time when i > 10, so valid i is (10, 110), it only contains 99 elements. So i should be [10, 110), which means '> ' in line 179 should be changed with '>='.


yanchen036:
I signed it!

Yan Chen ( 陈琰 )

zheng-xq:
LGTM.

@tensorflow-jenkins, test this please

danmane:

PR #3365, opened 2016-07-18
Remove read_analogies() from word2vec class initialization

raix852:
- Fix inconsistent between word2vec and word2vec_optimized.
- Remove read_analogies() from word2vec class initialization.

When I used word2vec model for training my target words directly, it lacked of analogies file for evaluation. Thus, it would raise file not found exception when model initialization. So I made read_analogies() public and called it after initialization.

PR #3637, opened 2016-08-04
Fix word2vec_optimized excepction when words contain unicode.

muik:
When word2vec_optimized deal with Korean language, it will break out below exception.

```
  File "word2vec_optimized.py", line 432, in <module>
    tf.app.run()
  File "/Library/Python/2.7/site-packages/tensorflow/python/platform/app.py", line 30, in run
    sys.exit(main(sys.argv))
  File "word2vec_optimized.py", line 417, in main
    model = Word2Vec(opts, session)
  File "word2vec_optimized.py", line 146, in __init__
    self.save_vocab()
  File "word2vec_optimized.py", line 242, in save_vocab
    opts.vocab_counts[i]))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```

I fix this exception by adding encode('utf-8') , and then it successful run word2vec_optimized model.

### DIFF
--- a/tensorflow/models/embedding/word2vec_optimized.py
+++ b/tensorflow/models/embedding/word2vec_optimized.py
@@ -238,7 +238,8 @@ class Word2Vec(object):
     opts = self._options
     with open(os.path.join(opts.save_path, "vocab.txt"), "w") as f:
       for i in xrange(opts.vocab_size):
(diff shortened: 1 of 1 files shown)


raix852:
I have done the same job in #3365 sub-commit.
But it still pending.

muik:
There is the same job in #3365 @raix852 mentioned.

Issue #2234, opened 2016-05-05
`_tanh(x)` returns different values for the same `x`


ninotoshi:
Why is the last one (`0.99505478`) different from the first four?

```python
>>> import tensorflow as tf
>>> from tensorflow.python.ops.gen_math_ops import _tanh
>>> a = tf.ones([1, 3], tf.float32)
>>> b = tf.ones([3, 5], tf.float32)
>>> c = tf.matmul(a, b)
>>> sess = tf.Session()
>>> print(sess.run(_tanh(c)))
[[ 0.99505472  0.99505472  0.99505472  0.99505472  0.99505478]]
```

With `tf.float64`, the values are all `0.99505475`.

yaroslavvb:
Can't reproduce, tried on Z420 and Z840 machines and TF at head.

Are you using GPU?

```
import tensorflow as tf
with tf.device("/cpu:0"):
  a = tf.ones([1, 3], tf.float32)
  b = tf.ones([3, 5], tf.float32)
sess = tf.Session()
print(sess.run(tf.tanh(tf.matmul(a, b))))

[[ 0.99505478  0.99505478  0.99505478  0.99505478  0.99505478]]
```

ninotoshi:
I get the same result even with your code specifying `tf.device("/cpu:0")`.
I am using 0.8.0 in the VirtualBox image of Ubuntu (`vagrant-ubuntu-vivid-64 3.19.0-10-generic #10-Ubuntu SMP`).

yaroslavvb:
Can you try and see if it's still non-deterministic when using with less threads

```
import tensorflow as tf
with tf.device("/cpu:0"):
  a = tf.ones([1, 3], tf.float32)
  b = tf.ones([3, 5], tf.float32)
  c = tf.tanh(tf.matmul(a, b))
config = tf.ConfigProto(inter_op_parallelism_threads=1, intra_op_parallelism_threads=1)
sess = tf.Session(config=config)
print(sess.run(c))
```

hunkim:
I can reproduce it (in python 3.5.1):
![image](https://cloud.githubusercontent.com/assets/901975/15057830/71bdc75e-134b-11e6-83d2-f451ee11d611.png)

It's getting interesting when I change the input size 10:

![image](https://cloud.githubusercontent.com/assets/901975/15057856/890356b8-134b-11e6-8265-a45179fbe818.png)

with 100, the problem is gone:
[...]

PR #3642, opened 2016-08-04
fix tanh inconsistent

suiyuan2009:
#2234
1. use tag dispatching to dispatch complex number and non-complex number, because no max/min operator for complex number, if not use dispatching, build will fail 
2. packetOp is inherited from scalar_tanh_op in eigen
3. complex number can't be packet, so just use numext::tanh(std::tanh) to compute

### DIFF
--- a/tensorflow/core/kernels/cwise_ops.h
+++ b/tensorflow/core/kernels/cwise_ops.h
@@ -236,6 +236,110 @@ struct functor_traits<scalar_compose_op<Scalar, UnaryFunctor, BinaryFunctor>> {
   };
 };
 
+// tag for not complex number pass to tanh
+struct tanh_not_complex_tag {};
+// tag for complex number pass to tanh
+struct tanh_complex_tag {};
+// struct for tag dispatching
+template <typename T>
+struct tanh_dispatching_tag {
+  typedef tanh_not_complex_tag type;
+};
+// struct specify for complex<double>
+template <>
+struct tanh_dispatching_tag<std::complex<double>> {
+  typedef tanh_complex_tag type;
+};
+// struct specify for complex<float>
+template <>
+struct tanh_dispatching_tag<std::complex<float>> {
+  typedef tanh_complex_tag type;
+};
+// compute tanh for complex number
+template <typename T>
+const T tanh_dispatching(tanh_complex_tag, const T& _x) {
+  return numext::tanh(_x);
+}
+// compute tanh for not complex number
+template <typename T>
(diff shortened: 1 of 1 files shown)


zheng-xq:
Adding @benoitsteiner to review. 

It is generally not a good idea to put so much operator specific logic in a common header file.

suiyuan2009:
@zheng-xq we can write two more scalar_tanh_op_google template for complex<double> and complex<float> specifically, so we can remove  tag dispatching

tensorflow/core/kernels/cwise_ops.h
```diff
@@ -236,6 +236,110 @@ struct functor_traits<scalar_compose_op<Scalar, UnaryFunctor, BinaryFunctor>> {
...
+  typedef tanh_complex_tag type;
+};
+// compute tanh for complex number
+template <typename T>
+const T tanh_dispatching(tanh_complex_tag, const T& _x) {
+  return numext::tanh(_x);
+}
+// compute tanh for not complex number
+template <typename T>
+const T tanh_dispatching(tanh_not_complex_tag, const T& _x) {
```
benoitsteiner: It would be a simpler and we wouldn't loose the vectorized implementation if you instead fixed the eigen code to return consistent results with a without vectorization. How about updating https://bitbucket.org/eigen/eigen/src/c44a78394f3f9d9bfc75caad97feb14c4183c942/Eigen/src/Core/functors/UnaryFunctors.h?at=default&fileviewer=file-view-default#UnaryFunctors.h-501 to use the code you implemented in tanh_dispatching instead of relying on std::tanh?

suiyuan2009: I'll move it to bitbucket, and write two other struct for complex number, tag dispatching is complicated.

suiyuan2009: I opened a pr on bitbucket [eigen](https://bitbucket.org/eigen/eigen/pull-requests/215/fix_tanh_inconsistent_for_tensorflow/diff)

danmane:
Taking over sync rotation for the week. Should we wait for the eigen PR to go through and then update / merge this afterwards? Or can this PR be closed if we solve the issue in eigen?

benoitsteiner:
The fix is now merged in Eigen. All that remains is to upgrade TensorFlow to pull the latest version of Eigen. This should happen in the next few days.

danmane:
OK, closing this PR per @benoitsteiner's comment.

PR #3655, opened 2016-08-04
Starting on a tf.contrib.learn notebook

random-forests:
I'd like to add an IPython notebook using tf.contrib.learn.

I've started on this one. Content is similar to the MNIST For ML Beginners tutorial: https://www.tensorflow.org/versions/r0.10/tutorials/mnist/beginners/index.html

Can continue fleshing it out down the road, and open to suggestions on how to improve.


terrytangyuan:
Awesome! One concern is that notebooks can get really large if we continue adding more in the future.

girving:
@martinwicke I'm not the right person for a `tf.learn` / docker PR.  Want to reroute?

martinwicke:
Jenkins, test this please.

Yeah, your assignment may have been a mis-click.

martinwicke:
Good point -- these are big. Can we clear the cells? Or would that defeat the purpose? Otherwise I would suggest we move all these tutorials into a separate repo.

danmane:
@craigcitro maybe you can opine? :)

random-forests:
I think we should keep the content of the cells displayed, but open to
hosting notebooks elsewhere to save on space. If we do host elsewhere,
let's put a single notebook in this directory that makes it clear how to
pull in the other notebooks pain-free.

craigcitro:
I think a separate repo could be good, especially if it had a `requirements.txt` with an explicit version of tensorflow they were written against will also be a big win.

Note that keeping the outputs in place has two big advantages:

1. github does a great job displaying them (yay nbviewer)
1. Makes it possible to write tests (at least internally)

I think end-users will benefit a lot from (1).

martinwicke:
Separate repo it is.

danmane:
@martinwicke Please close this PR when you've set up another repo for the tutorials.

rmlarsen:
@martinwicke what's the status here?

martinwicke:
Still working on someone to maintain the other repo. Any volunteers?

vrv:
I'm going to close this since it sounds like we're going to move this to another repo.  No value in having the PR in the open status

PR #3664, opened 2016-08-05
Update jupyter_notebook_config.py

theely:
Adding the capability to set jupyter port via the PORT environment variable.
This is especially useful if this Docker container runs on clusters managers like Mesos, DCOS , etc.

### DIFF
--- a/tensorflow/tools/docker/jupyter_notebook_config.py
+++ b/tensorflow/tools/docker/jupyter_notebook_config.py
@@ -16,7 +16,7 @@ import os
 from IPython.lib import passwd
 
 c.NotebookApp.ip = '*'
-c.NotebookApp.port = 8888
+c.NotebookApp.port = int(os.getenv('PORT', '') or 8888)
 c.NotebookApp.open_browser = False
 c.MultiKernelManager.default_kernel_name = 'python2'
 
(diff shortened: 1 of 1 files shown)


tensorflow/tools/docker/jupyter_notebook_config.py
```diff
@@ -16,7 +16,7 @@
 from IPython.lib import passwd
 
 c.NotebookApp.ip = '*'
-c.NotebookApp.port = 8888
+c.NotebookApp.port = int(os.getenv('PORT', '') or 8888)
```
craigcitro: You could even do `int(os.getenv('PORT', 8888))` if you want.

craigcitro:
LGTM

danmane:

tensorflow/tools/docker/jupyter_notebook_config.py
theely: @craigcitro would you like me to change it to int(os.getenv('PORT', 8888))? I don't have a strong preference, actually your proposal seems cleaner to me.

craigcitro: yeah, let's go ahead and change it.

vrv:

PR #3666, opened 2016-08-05
Fix analogy response of word2vec interactive when the response is unicode.

muik:
I run word2vec.py with interactive mode.

```
$ python word2vec.py --interactive --train_data=t --eval_data=q --save_path=save
...
In [3]: model.analogy(b'남자', b'여자', b'아빠')
Out[3]: '\xec\x97\x84\xeb\xa7\x88'
```

The response of analogy couldn't be readable.

So I fix the response and then rerun the code.

```
In [3]: model.analogy(b'남자', b'여자', b'아빠')
엄마
```

Now I can see the response.

### DIFF
--- a/tensorflow/models/embedding/word2vec.py
+++ b/tensorflow/models/embedding/word2vec.py
@@ -477,8 +477,8 @@ class Word2Vec(object):
     for c in [self._id2word[i] for i in idx[0, :]]:
       if c not in [w0, w1, w2]:
         print(c)
-        return
-    return "unknown"
+        break
+    print("unknown")
 
   def nearby(self, words, num=20):
     """Prints out nearby words given a list of words."""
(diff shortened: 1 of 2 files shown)


danmane:
This doesn't seem quite right to me. Why should the analogy function sometimes return a string, but sometimes print a string and then return nothing? That is a strange contract for a function, and will confuse users of this code who are depending on the return value (i.e. not in interactive mode).

Instead of changing the type signature of the function, can you instead make it handle unicode correctly?

muik:
@danmane  I agree a strange contract for the analogy function. But IPython always prints an unreadable unicode output which is decoded or not.

IPython examples:
```
In [1]: '한글'
'\xed\x95\x9c\xea\xb8\x80'

In [2]: '한글'.decode('UTF-8')
u'\ud55c\uae00'
```

So I fixed the function to be printing a result like the nearby function which is also a function for an interactive mode.

rmlarsen:
@danmane please give this another look.

danmane:
Ok, @zffchen78 (who has a bit more context on this code) gave it LGTM.

Issue #3621, opened 2016-08-03
layer normalization


suiyuan2009:
recently Hinton published layer normalization on [arxiv layer normalization](https://arxiv.org/pdf/1607.06450v1.pdf?), any plan to add an Op?

suiyuan2009:
I see c++ implement of batch_normalization is deprecated

suiyuan2009:
@xodus7 great

xcyan:
@suiyuan2009 Have you ever tried this layer_norm op with most recent TF (1.3 or 1.4)? 
#3671 #14562

PR #3671, opened 2016-08-05
Add layer_norm op to contrib.layers.

xodus7:
Addresses #3621 for python at least. The new op is modeled after batch_norm and borrows code from it heavily.

### DIFF
--- a/tensorflow/contrib/layers/python/layers/layers.py
+++ b/tensorflow/contrib/layers/python/layers/layers.py
@@ -52,6 +52,7 @@ __all__ = ['avg_pool2d',
            'dropout',
            'flatten',
            'fully_connected',
+           'layer_norm',
            'linear',
            'max_pool2d',
            'one_hot_encoding',
@@ -837,6 +838,92 @@ def fully_connected(inputs,
       outputs.set_shape(static_shape)
     return utils.collect_named_outputs(outputs_collections,
                                        sc.original_name_scope, outputs)
+
+
+@add_arg_scope
+def layer_norm(inputs,
+               center=True,
+               scale=False,
+               epsilon=0.001,
+               activation_fn=None,
+               reuse=None,
+               variables_collections=None,
+               outputs_collections=None,
+               trainable=True,
+               scope=None):
+  """Adds a Layer Normalization layer from https://arxiv.org/abs/1607.06450.
+
+    "Layer Normalization"
+
+    Jimmy Lei Ba, Jamie Ryan Kiros, Geoffrey E. Hinton
+
+  Can be used as a normalizer function for conv2d and fully_connected.
+
+  Args:
+    inputs: a tensor of size `[batch_size, height, width, channels]`
+            or `[batch_size, channels]`.
+    center: If True, subtract `beta`. If False, `beta` is ignored.
+    scale: If True, multiply by `gamma`. If False, `gamma` is
+      not used. When the next layer is linear (also e.g. `nn.relu`), this can be
+      disabled since the scaling can be done by the next layer.
+    epsilon: small float added to variance to avoid dividing by zero.
+    activation_fn: Optional activation function.
+    reuse: whether or not the layer and its variables should be reused. To be
+      able to reuse the layer scope must be given.
+    variables_collections: optional collections for the variables.
+    outputs_collections: collections to add the outputs.
+    trainable: If `True` also add variables to the graph collection
+      `GraphKeys.TRAINABLE_VARIABLES` (see tf.Variable).
+    scope: Optional scope for `variable_op_scope`.
+
+  Returns:
+    A `Tensor` representing the output of the operation.
+
+  Raises:
+    ValueError: if rank or last dimension of `inputs` is undefined.
+  """
+  with variable_scope.variable_op_scope([inputs],
+                                        scope, 'LayerNorm', reuse=reuse) as sc:
+    inputs = ops.convert_to_tensor(inputs)
+    inputs_shape = inputs.get_shape()
+    inputs_rank = inputs_shape.ndims
+    if inputs_rank is None:
+      raise ValueError('Inputs %s has undefined rank.' % inputs.name)
+    dtype = inputs.dtype.base_dtype
+    axis = list(range(1, inputs_rank))
+    params_shape = inputs_shape[-1:]
+    if not params_shape.is_fully_defined():
+      raise ValueError('Inputs %s has undefined last dimension %s.' % (
+          inputs.name, params_shape))
+    # Allocate parameters for the beta and gamma of the normalization.
+    beta, gamma = None, None
+    if center:
+      beta_collections = utils.get_variable_collections(variables_collections,
+                                                        'beta')
+      beta = variables.model_variable('beta',
+                                      shape=params_shape,
+                                      dtype=dtype,
+                                      initializer=init_ops.zeros_initializer,
+                                      collections=beta_collections,
+                                      trainable=trainable)
+    if scale:
+      gamma_collections = utils.get_variable_collections(variables_collections,
+                                                         'gamma')
+      gamma = variables.model_variable('gamma',
+                                       shape=params_shape,
+                                       dtype=dtype,
+                                       initializer=init_ops.ones_initializer,
+                                       collections=gamma_collections,
+                                       trainable=trainable)
+    # Calculate the moments on the last axis (layer activations).
+    mean, variance = nn.moments(inputs, axis, keep_dims=True)
+    # Compute layer normalization using the batch_normalization function.
+    outputs = nn.batch_normalization(
+        inputs, mean, variance, beta, gamma, epsilon)
+    outputs.set_shape(inputs_shape)
+    if activation_fn:
+      outputs = activation_fn(outputs)
+    return utils.collect_named_outputs(outputs_collections, sc.name, outputs)
 
 
 @add_arg_scope
--- a/tensorflow/contrib/layers/python/layers/layers.py
+++ b/tensorflow/contrib/layers/python/layers/layers.py
@@ -843,8 +843,7 @@ def fully_connected(inputs,
 @add_arg_scope
 def layer_norm(inputs,
                center=True,
-               scale=False,
-               epsilon=0.001,
+               scale=True,
                activation_fn=None,
                reuse=None,
                variables_collections=None,
@@ -866,7 +865,6 @@ def layer_norm(inputs,
     scale: If True, multiply by `gamma`. If False, `gamma` is
       not used. When the next layer is linear (also e.g. `nn.relu`), this can be
       disabled since the scaling can be done by the next layer.
-    epsilon: small float added to variance to avoid dividing by zero.
     activation_fn: Optional activation function.
     reuse: whether or not the layer and its variables should be reused. To be
       able to reuse the layer scope must be given.
@@ -918,8 +916,9 @@ def layer_norm(inputs,
     # Calculate the moments on the last axis (layer activations).
     mean, variance = nn.moments(inputs, axis, keep_dims=True)
     # Compute layer normalization using the batch_normalization function.
+    variance_epsilon = 1E-12
     outputs = nn.batch_normalization(
-        inputs, mean, variance, beta, gamma, epsilon)
+        inputs, mean, variance, beta, gamma, variance_epsilon)
     outputs.set_shape(inputs_shape)
     if activation_fn:
       outputs = activation_fn(outputs)
--- a/tensorflow/contrib/layers/python/layers/layers.py
+++ b/tensorflow/contrib/layers/python/layers/layers.py
@@ -859,8 +859,8 @@ def layer_norm(inputs,
   Can be used as a normalizer function for conv2d and fully_connected.
 
   Args:
-    inputs: a tensor of size `[batch_size, height, width, channels]`
-            or `[batch_size, channels]`.
+    inputs: a tensor with 2 or more dimensions. The normalization
+            occurs over all but the first dimension.
     center: If True, subtract `beta`. If False, `beta` is ignored.
     scale: If True, multiply by `gamma`. If False, `gamma` is
       not used. When the next layer is linear (also e.g. `nn.relu`), this can be
@@ -880,8 +880,8 @@ def layer_norm(inputs,
   Raises:
     ValueError: if rank or last dimension of `inputs` is undefined.
   """
-  with variable_scope.variable_op_scope([inputs],
-                                        scope, 'LayerNorm', reuse=reuse) as sc:
+  with variable_scope.variable_scope(scope, 'LayerNorm', [inputs],
+                                     reuse=reuse) as sc:
     inputs = ops.convert_to_tensor(inputs)
     inputs_shape = inputs.get_shape()
     inputs_rank = inputs_shape.ndims
@@ -922,7 +922,9 @@ def layer_norm(inputs,
     outputs.set_shape(inputs_shape)
     if activation_fn:
       outputs = activation_fn(outputs)
-    return utils.collect_named_outputs(outputs_collections, sc.name, outputs)
+    return utils.collect_named_outputs(outputs_collections,
+                                       sc.original_name_scope,
+                                       outputs)
 
 
 @add_arg_scope
--- a/tensorflow/contrib/layers/python/layers/layers_test.py
+++ b/tensorflow/contrib/layers/python/layers/layers_test.py
@@ -1571,9 +1571,8 @@ class LayerNormTest(tf.test.TestCase):
       # output_train and output_eval should be the same.
       self.assertAllClose(sess.run([output_train]), sess.run([output_eval]))
 
-  def testOutput(self):
+  def doOutputTest(self, input_shape):
     with self.test_session() as sess:
-      input_shape = (10, 100)
       input_values = np.random.rand(*input_shape)
       inputs = tf.constant(input_values, shape=input_shape, dtype=tf.float32)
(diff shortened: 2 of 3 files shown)


xodus7:
Added the proper email addresses to my account. Do I need to do anything else to resolve the CLA auto-reject?

suiyuan2009:
@xodus7 have you use this op on any example such as seq2seq translate

xodus7:
@suiyuan2009 I've tested it with good results on a custom character RNN using a modified GRU cell (as described in the appendix of the paper). Training performance was ~50% slower due to the 4 extra layer normalizations per cell, but the network converged much quicker and with a lower perplexity overall.

xodus7:
@suiyuan2009 Good catch on the test functions. I'll fix...

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -836,6 +837,92 @@ def fully_connected(inputs,
...
+    Jimmy Lei Ba, Jamie Ryan Kiros, Geoffrey E. Hinton
+
+  Can be used as a normalizer function for conv2d and fully_connected.
+
+  Args:
+    inputs: a tensor of size `[batch_size, height, width, channels]`
+            or `[batch_size, channels]`.
+    center: If True, subtract `beta`. If False, `beta` is ignored.
+    scale: If True, multiply by `gamma`. If False, `gamma` is
+      not used. When the next layer is linear (also e.g. `nn.relu`), this can be
```
martinwicke: Is this only true for linear and relu? Any fully connected layer learn the scale, right?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -836,6 +837,92 @@ def fully_connected(inputs,
...
+      disabled since the scaling can be done by the next layer.
+    epsilon: small float added to variance to avoid dividing by zero.
```
martinwicke: Is the variance is zero, you don't have to change the values since they're all the same. Can you do that instead: if variance < float_epsilon, do not divide by it? You can find the appropriate epsilon by asking python, and then remove this argument.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -836,6 +837,92 @@ def fully_connected(inputs,
 
 
 @add_arg_scope
+def layer_norm(inputs,
+               center=True,
+               scale=False,
+               epsilon=0.001,
+               activation_fn=None,
+               reuse=None,
```
martinwicke: Should this have an is_training boolean like batch_norm?

martinwicke:
Left some comments

tensorflow/contrib/layers/python/layers/layers.py
xodus7: @martinwicke  I think the general intuition is that normalizing values with a variance of 1.0 before non-linearities such as tanh limits most activations to the linear portion of the function. This causes the network to loss it's expressive power, which the learned gamma and beta terms counteract. I think the idea is if you are applying a "linear" activation function, such as the identity transformation or relu, then these terms are not strictly necessary (though still beneficial?) and can be optimized out. Perhaps the default for this parameter should be True?

tensorflow/contrib/layers/python/layers/layers.py
xodus7: One of the features on layer normalization is that normalization happens across features/neurons in a layer and so each sample in a batch is independent. This means that the mean and variance is calculated the same during both training and inference and we don't need moving averages.

tensorflow/contrib/layers/python/layers/layers.py
xodus7: This seems like a good idea and should be achievable with application of the tf.select() function. I guess the only question is if there is a significant performance impact with that approach? I'll give it a shot and do some benchmarks. Also, I think this could be done for batch_norm as well?

tensorflow/contrib/layers/python/layers/layers.py
martinwicke: You're right, I wrote that before I noticed the absence of update_ops and saw what you were doing, sorry for the noise.

tensorflow/contrib/layers/python/layers/layers.py
martinwicke: I don't quite understand: The data flow goes

layer -> layer_norm -> activation_fn -> next layer -> activation_fn or another layer_norm or whatnot

So maybe if the activation_fn given to this is a linear one (None, or relu, or reluX), you may want this to be false?

Anyway, I think a default of True makes more sense -- it seems to me you're not losing much even in terms of performance if it's on.

tensorflow/contrib/layers/python/layers/layers.py
martinwicke: It may well be slower. A benchmark would help a lot. If it turns out to be slower, I'd still much prefer if we eliminated this parameter by replacing it with flt_min, the smallest positive float (depending on the internally used dtype), instead of an arbitrary number (with a fairly large default).

mikowals: I think adding epsilon before the square root and division is the correct approach.  If variance were 1 the division could be skipped but for variance near or at 0 the impact of division is  to change the values significantly.

xodus7: @mikowals That's what the tf.batch_normalization() op does (which I use here). Thinking this through a bit more:

* This can only occur if all values are identical.
* If all values are identical then the variance is zero since we first subtract the mean.
* Therefore the value of epsilon is unimportant since the end result of division will be zero.

Is there a flaw in my logic or some corner case on certain hardware I'm not considering? If not then I think the correct thing to do here is to remove the parameter and pass in a sane default to the batch_normalization op.

xodus7: I push a change; thanks for the comments.

sguada:
Jenkins test this please

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -914,8 +912,9 @@ def layer_norm(inputs,
     # Calculate the moments on the last axis (layer activations).
     mean, variance = nn.moments(inputs, axis, keep_dims=True)
     # Compute layer normalization using the batch_normalization function.
+    variance_epsilon = 1E-12
```
sguada: an epsilon so small can create huge gradients if the activations are 0.

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -1491,6 +1491,88 @@ def testNoneUpdatesCollectionIsTrainingVariable(self):
...
+                                                  scope='LN')
+      output_eval = tf.contrib.layers.layer_norm(images,
+                                                 scope='LN',
+                                                 reuse=True)
+      # Initialize all variables
+      sess.run(tf.initialize_all_variables())
+      # output_train and output_eval should be the same.
+      self.assertAllClose(sess.run([output_train]), sess.run([output_eval]))
+
+  def testOutput(self):
```
sguada: Can you add another test of outputs when the input is a 4-D Tensor.

tensorflow/contrib/layers/python/layers/layers.py
xodus7: @sguada I had originally set this to 0.001 but receive feedback that it is too large. I'm not sure what the correct value to put here is. Do you have a suggestion?

tensorflow/contrib/layers/python/layers/layers_test.py
xodus7: Sure, I will add it soon.

xodus7: I've added the test as requested.

sguada:
Jenkins, please test this.

sguada:
@martinwicke Why I cannot test this? and why I cannot merge it either?

rmlarsen:

sguada:

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -836,6 +837,91 @@ def fully_connected(inputs,
...
+      `GraphKeys.TRAINABLE_VARIABLES` (see tf.Variable).
+    scope: Optional scope for `variable_op_scope`.
+
+  Returns:
+    A `Tensor` representing the output of the operation.
+
+  Raises:
+    ValueError: if rank or last dimension of `inputs` is undefined.
+  """
+  with variable_scope.variable_op_scope([inputs],
```
sguada: replace 
with variable_scope.variable_scope(scope, 'LayerNorm', [inputs], reuse=reuse) as sc:

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -836,6 +837,91 @@ def fully_connected(inputs,
...
+    # Calculate the moments on the last axis (layer activations).
+    mean, variance = nn.moments(inputs, axis, keep_dims=True)
+    # Compute layer normalization using the batch_normalization function.
+    variance_epsilon = 1E-12
+    outputs = nn.batch_normalization(
+        inputs, mean, variance, beta, gamma, variance_epsilon)
+    outputs.set_shape(inputs_shape)
+    if activation_fn:
+      outputs = activation_fn(outputs)
+    return utils.collect_named_outputs(outputs_collections, sc.name, outputs)
```
sguada: replace sc.name with sc.original_name_scope

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -1491,6 +1491,93 @@ def testNoneUpdatesCollectionIsTrainingVariable(self):
...
+      images = np.random.uniform(size=(5, height, width, 3))
+      output = tf.contrib.layers.layer_norm(images)
+      self.assertTrue(output.op.name.startswith('LayerNorm/batchnorm'))
+      self.assertListEqual(output.get_shape().as_list(), [5, height, width, 3])
+
+  def testCreateVariables(self):
+    height, width = 3, 3
+    with self.test_session():
+      images = tf.random_uniform((5, height, width, 3), seed=1)
+      tf.contrib.layers.layer_norm(images, scale=True)
```
sguada: the default of scale is already True, remove it.

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -1491,6 +1491,93 @@ def testNoneUpdatesCollectionIsTrainingVariable(self):
...
+      beta = tf.contrib.framework.get_variables_by_name('beta')[0]
+      gamma = tf.contrib.framework.get_variables_by_name('gamma')[0]
+      self.assertEquals(beta.op.name, 'LayerNorm/beta')
+      self.assertEquals(gamma.op.name, 'LayerNorm/gamma')
+
+  def testReuseVariables(self):
+    height, width = 3, 3
+    with self.test_session():
+      images = tf.random_uniform((5, height, width, 3), seed=1)
+      tf.contrib.layers.layer_norm(images, scale=True, scope='ln')
```
sguada: Remove scale=True

sguada:
@xodus7 Please update the last set of comments

xodus7:
@sguada I've addressed the comments. Some of the changes were due to recent commits it seems (variable_scope) so I went ahead and re-based on master. That seems to have reset the googlebot. I hope that it didn't create a major inconvenience for you.

rmlarsen:
@sguada can you confirm that this is ready to merge (pending tests)?

sguada:

sguada:
@rmlarsen once test pass is ready for merge

rmlarsen:
the test failure is a known flaky test. Merging.

helinwang:
Thanks, on my experiment generalizes much better than batch norm.

MycChiu:
Great stuff, Thank you! Just one question though, has any one used this layer on recurrent networks and found some major slow down in terms of wall clock time? (about 3-5x slower compared to vanilla rnn without any form of normalization.) 
Although I think this is not the problem of this particular implementation, since I found [a thread on reddit ](https://www.reddit.com/r/MachineLearning/comments/4ufmxy/layer_normalization_implemented_in_tensorflow/), where other people have observed same slowdown with a different implementation. However, one of the co-authors from the original layer norm paper [commented in the same thread ](https://www.reddit.com/r/MachineLearning/comments/4ufmxy/layer_normalization_implemented_in_tensorflow/d5q95oe) stating that the implementation in Theano does not cause any slowdown with the help of CNMeM.
I am really eager to solve this, since layer norm does help the model to converge faster in terms of iteration number, and if there is a way to incorporate it into rnn without any slowdown, I would love to add it to my future experiments more often.

xodus7:
@MycChiu Yeah I'm also seeing a significant slowdown, though not quite that dramatic. I'm not too well versed on the Eigen library but it could be that my use of the batch_normalization op in this way is less than optimal.

MycChiu:
I found that this slowdown is not limited to rnn, it also happens to fully connected layers. Here's a quick benchmark:
```python
from tensorflow.contrib.layers import layer_norm
from timeit import default_timer

import tensorflow as tf
import numpy as np
slim = tf.contrib.slim

batch_size = 64
nb_units = 512
nb_layers = 20
nb_epoch = 3
itrs = 100


def benchmark(norm_fn):
    epo_times = np.zeros([nb_epoch])
    with tf.Graph().as_default():
        # fake data
        x = tf.random_uniform([batch_size, 30])
        y = tf.random_uniform([batch_size], maxval=10)
        labels = tf.to_int32(y)
        # define graph, need to pass scale argument since batch_norm's default
        # option for scale is False while that of layer_norm is True.
        out = slim.repeat(x, nb_layers, slim.fully_connected, nb_units,
                          # normalizer_fn=layer_norm,
                          normalizer_fn=norm_fn,
                          normalizer_params={"scale": True},
                          scope="fc")
        # calculate loss
        loss = tf.nn.sparse_softmax_cross_entropy_with_logits(out, labels)
        # define optimizer
        optimizer = tf.train.AdamOptimizer(1e-3)
        # create train_op
        train_op = optimizer.minimize(loss)
        # initialization op
        init = tf.initialize_all_variables()

        with tf.Session("") as sess:
            sess.run(init)
            # warm up
            for i in range(itrs):
                sess.run(train_op)
            # start benchmark
            for e in range(nb_epoch):
                epo_start = default_timer()
                for i in range(itrs):
                    sess.run(train_op)
                eT = (default_timer() - epo_start)
                epo_times[e] = eT
            return epo_times



vanilla_t = benchmark(None)
batchNorm_t = benchmark(slim.batch_norm)
layerNorm_t = benchmark(layer_norm)
print("vanilla times in seconds:")
print(vanilla_t)
print("batch_norm times in seconds:")
print(batchNorm_t)
print("layer_norm times in seconds:")
print(layerNorm_t)
print("batch_norm is %1.2fX slower" %
          (batchNorm_t.sum() / vanilla_t.sum()))
print("layer_norm is %1.2fX slower" %
          (layerNorm_t.sum() / vanilla_t.sum()))
```
Here's my output:
```
vanilla times in seconds:
[ 0.42624067  0.42290944  0.43818542]
batch_norm times in seconds:
[ 0.74915753  0.74786358  0.75310942]
layer_norm times in seconds:
[ 1.13074295  1.12952367  1.13170891]
batch_norm is 1.75X slower
layer_norm is 2.63X slower
```
I just built tensorflow from source using the current master branch, so it's probably not the infamous batch_norm regression from few weeks ago. I wonder if this slowdown is inevitable for layer normalization or there are actually ways to improve the speed.

@xodus7 I too can't think of a better way to implement this, I tried a naive approach using lower level ops, but it was even slower than yours.

xodus7:
I'm no expert on low-level GPU architecture but I wonder if the slowdown is due to caching effects as we normalize across layer activations rather than across features.

xodus7:
@MycChiu Just to add another data point, I ran your benchmark on my system and got very similar results:

vanilla times in seconds:
[ 0.392066    0.39233494  0.39980507]
batch_norm times in seconds:
[ 0.68308187  0.68399501  0.68293214]
layer_norm times in seconds:
[ 1.08977199  1.08793688  1.08768916]
batch_norm is 1.73X slower
layer_norm is 2.76X slower

MycChiu:
@xodus7 Yeah, I wonder the same thing, but too bad I have no idea how to improve this. Hopefully someone capable of improving this will see this thread soon. (and thanks for your benchmark result, at least now I know the slowdown is probably not caused my machine's configuration.)

MycChiu:
If anyone is still interested, I have written a [GPU kernel for layer normalization](https://github.com/MycChiu/fast-LayerNorm-TF) which reduce the slowdown dramatically.

ilblackdragon:
@MycChiu would you send a PR with your implementation of layer norm? It probably can go where the batch_norm_fused kernel is. Thanks!

PS, if you are at NIPS we can meet up.

sguada:
I think it would need a bit more work to make it more general but seems a good beginning.

MycChiu:
phew...after some hours of battling with bazel's build rules, I finally merged the codes into `tf.contrib.layers`. Here's the PR #6205.

I didn't put them in `core/kernels`, because as @sguada suggested, it currently has quite a lot limitations, and since I also wrote the layer function mimicking current `tf.contrib.layers.layer_norm`, I decided to put them under `tf.contrb.layers`

@ilblackdragon I would love to! Too bad my current financial state doesn't allow me to travel to Spain :(

xcyan:
@xodus7 One question about the layer_normalization layer in this PR:
Did you really implement the layer_norm based on Ba et al 2016?

As far as I can see, scalar offset/scale was used in Ba et al 2016 but there is no such option in your implementation. Instead, you normalized each individual dimensional separately (1D offset/scale vector).

Issue #2921, opened 2016-06-17
No gradient defined for operation 'ExtractImagePatches'


jyshee:
I tried to  use [tf.extract_image_patches](https://www.tensorflow.org/versions/r0.9/api_docs/python/array_ops.html#extract_image_patches) to implement a locally layer as #2605 said. However an error occurred:
LookupError: No gradient defined for operation 'local/ExtractImagePatches' (op type: ExtractImagePatches)  

I checked that tf.space_to_depth has the gradient definition but tf.extract_image_patches does not. And it seems not easy for me to implement the gradient. Any advice? 
@martinwicke @benoitsteiner @mrry

sherrym:
@jyshee go for it if you need it.

jyshee:
I tried to implement the gradient of tf.extract_image_patches using tf.reshape and tf.pad Ops as:

 @ops.RegisterGradient("ExtractImagePatches")
   def _ExtractImagePatchesGrad(op,grad):
      ksize=op.get_attr("ksizes")[1]
      im=grad
      batch_size = im.get_shape()[0].value
      h = im.get_shape()[1].value
      w = im.get_shape()[2].value
      ch = im.get_shape()[3].value
      h_o = h + (ksize - 1)
      w_o = w + (ksize - 1)
      ch_o = ch / ksize / ksize
      im_o = tf.Variable(tf.ones([batch_size, h_o, w_o, ch_o]), trainable=False)
      for i in xrange(h):
        for j in xrange(w):
           res=tf.reshape(im[:,i,j,:],[-1,ksize,ksize,int(ch/ksize/ksize)])
           padres=tf.pad(res,[[0,0],[i,h_o-i-ksize],[j,w_o-j-ksize],[0,0]])
           im_o+=padres
     return tf.identity(im_o)


But when I trained the model, the weights parameters of both local layer and Conv layers become very large (about 30 times larger than before). I'm still working on this problem.

hughsalimbeni:
I've hit this problem also. A quick test of the @jyshee solution fails in my example in the reshape line due to a mismatched number of elements

jyshee:
@hughsalimbeni could you show more details about your error when you test the codes above?

hughsalimbeni:
My code is rather messy and very far from a minimal example, but here it is:

```
    def make_x_hat(self, x):
        L, H = self.L, self.H
        #put two cols of zeros
        # TODO sort out messy tranpose
[...]

PR #3672, opened 2016-08-05
added image patch grad

nikhilmishra000:
This solves: https://github.com/tensorflow/tensorflow/issues/2921

I ran unit tests using tensorflow.python.ops.gradient_checker to verify that it works :D

### DIFF
--- a/tensorflow/python/ops/array_grad.py
+++ b/tensorflow/python/ops/array_grad.py
@@ -481,7 +481,8 @@ def _ExtractImagePatchesGrad(op, grad):
     cols_out = int(ceil((cols_in - ksize_c_eff + 1) / stride_h))
     pad_rows = (rows_out - 1) * stride_r + ksize_r_eff - rows_in
     pad_cols = (cols_out - 1) * stride_h + ksize_c_eff - cols_in
-    pad_rows, pad_cols = max(0, pad_rows), max(0, pad_cols)
+
+  pad_rows, pad_cols = max(0, pad_rows), max(0, pad_cols)
 
   grad_expanded = array_ops.transpose(
     array_ops.reshape(grad, (batch_size, rows_out,
--- a/tensorflow/python/ops/array_grad.py
+++ b/tensorflow/python/ops/array_grad.py
@@ -19,10 +19,13 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+from math import ceil
+
 from tensorflow.python.framework import constant_op
 from tensorflow.python.framework import ops
 from tensorflow.python.framework import tensor_util
 from tensorflow.python.ops import array_ops
+from tensorflow.python.ops import sparse_ops
 from tensorflow.python.ops import gen_array_ops
 from tensorflow.python.ops import math_ops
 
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/array_grad.py
```diff
@@ -448,3 +451,77 @@ def _MirrorPadGradGrad(op, grad):
 @ops.RegisterGradient("QuantizeAndDequantize")
 def _QuantizeAndDequantizeGrad(_, grad):
   return grad
+
+
+@ops.RegisterGradient("ExtractImagePatches")
+def _ExtractImagePatchesGrad(op, grad):
```
benoitsteiner: Thanks for the contribution, this looks pretty good. Can you add a regression test to make sure the code is correct? You can use https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/conv2d_backprop_filter_grad_test.py#L25 as an example.

nikhilmishra000:
added the tests, let me know if there's anything else

danmane:

benoitsteiner:
Jenkins, test this please.

danmane:
The tests are failing :(

nikhilmishra000:
I'm looking at the logs, and the one that's failing is //tensorflow/tools/test:check_futures_test.
I think it's because my test file "tensorflow/python/kernel_tests/extract_image_patches_grad_test.py" doesn't import "absolute_import"?

danmane:
Push the fix, and I'll re-trigger the tests.

danmane:

danmane:
@benoitsteiner Can you LGTM this now that tests were added?

tensorflow/python/ops/array_grad.py
```diff
@@ -448,3 +451,78 @@ def _MirrorPadGradGrad(op, grad):
...
+  ksize_r_eff = ksize_r + (ksize_r - 1) * (rate_r - 1)
+  ksize_c_eff = ksize_c + (ksize_c - 1) * (rate_c - 1)
+
+  if padding == 'SAME':
+    rows_out = int(ceil(rows_in / stride_r))
+    cols_out = int(ceil(cols_in / stride_h))
+    pad_rows = ((rows_out - 1) * stride_r + ksize_r_eff - rows_in) // 2
+    pad_cols = ((cols_out - 1) * stride_h + ksize_c_eff - cols_in) // 2
+
+  elif padding == 'VALID':
```
gunan: Looks like in python 3 PIP setups, there is a problem caused by these lines.
It looks like "padding" can be something other than 'SAME' or 'VALID'
any ideas what might be happening?

http://ci.tensorflow.org/view/Tensorflow%20Jenkins%20Monitored%20builds/job/nightly-matrix-cpu/TF_BUILD_CONTAINER_TYPE=CPU,TF_BUILD_IS_OPT=OPT,TF_BUILD_IS_PIP=PIP,TF_BUILD_PYTHON_VERSION=PYTHON3,label=cpu-slave/194/consoleFull

gunan:
Looks like the BUILD file change for the test is missing from this PR.
Therefore Jenkins did not run the new feature, and I suspect this is broken.
I am investigating now.

gunan:
I am preparing a fix for the issue.
It will be pushed to github later today, or tomorrow.

mohamedadaly:
The current gradient seems to have a problem when `batches` is `None` i.e. when using a variable size batches using Keras. Simply added this line
```python
  batch_size = array_ops.shape(op.inputs[0])[0]
```
after this
```python
  batch_size, rows_in, cols_in, channels = [
    dim.value for dim in op.inputs[0].get_shape()
  ]
```
and it seems to have fixed the problem. I am new to TensorFlow, so you probably have a better solution than this.

PR #3679, opened 2016-08-06
Fixes #3677: Added missing save_summary_steps in BaseEstimator

### LINKED ISSUE #3677: TFLearn estimators not passing "save_summary_steps"
I think this should be an easy fix (or it was to hack it in locally), but it looks like [supervisor.py:L560](https://github.com/tensorflow/tensorflow/blob/r0.10/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L560) for release **r0.10** is calling ```graph_actions._supervised_train``` without passing the ```save_summary_steps``` parameter now configurable in RunConfig.

I was able to correct by adding ```supervisor_save_summaries_steps=self._config.save_summary_steps``` in the call to graph_actions._supervised_train and this worked as expected, with events showing in TensorBoard on a higher configured frequency than the default 100 steps.

terrytangyuan:
Reported in #3677

### DIFF
--- a/tensorflow/contrib/learn/python/learn/estimators/estimator.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/estimator.py
@@ -499,6 +499,7 @@ class BaseEstimator(
           supervisor_is_chief=(self._config.task == 0),
           supervisor_master=self._config.master,
           supervisor_save_model_secs=self._config.save_checkpoints_secs,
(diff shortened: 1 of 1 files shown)


danmane:
@martinwicke @ilblackdragon it looks like this could be reviewed in like 30 seconds, and then we'd have one fewer open pull requests! :)

PR #3695, opened 2016-08-08
Added code to raise ValueError if parent directory of  path Saver.save…

PuchatekwSzortach:
Current version of Saver.save(session, path) fails with a rather mysterious error code when parent directory of path doesn't exist. It throws a `NotFoundError`, listing a bunch of nodes variables, e.g.:

```
(...)
tensorflow.python.framework.errors.NotFoundError: /tmp/tensorflow/autoencoder/cifar/cifar.ckpt.tempstate2833829603959566611
   [[Node: save/save = SaveSlices[T=[DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT], _device="/job:localhost/replica:0/task:0/cpu:0"](_recv_save/Const_0, save/save/tensor_names, save/save/shapes_and_slices, Variable, Variable/Adam, Variable/Adam_1, beta1_power, beta2_power, biases, biases/Adam, biases/Adam_1, biases_1, biases_1/Adam, biases_1/Adam_1)]]
```

This makes user believe problem is with some ops not being found by `Saver` object and doesn't point to path problem at all.

I added code that checks that parent folder of output path exists and if not, fails with a `ValueError` and appropriate message.


danmane:
@PuchatekwSzortach please satisfy CLABot and we'll move forward with review.

PuchatekwSzortach:
@danmane I did sign CLA on both accounts I used to write the code, but for one of them after the pull request was submitted. I assume this didn't trigger CLABot reevaluation, as the action was taken outside of Github. Is there a way to force CLABot to run its check again?

PuchatekwSzortach:
@danmane @googlebot I authored all commits for this pull request, though some code comes from a different account I use. If needed I can post a comment testifying to that with the other account. If there is a another "proper" way of satisfying the CLA requirements apart for already having sign it with all participating accounts, please let me know how.

vrv:
If you rewrite the email address to the one you signed the CLA with (git commit --amend --author="Your Name <your email>"), then push, the CLA bot will be happy

PuchatekwSzortach:
@vrv Done

vrv:
Try squashing your commits

tensorflow/python/training/saver.py
```diff
@@ -1070,6 +1070,10 @@ def save(self, sess, save_path, global_step=None, latest_filename=None,
     if not isinstance(sess, session.SessionInterface):
       raise TypeError("'sess' must be a Session; %s" % sess)
 
+    # Note a few lines above save_path was set to os.path.dirname(save_path)
```
danmane: I'm a bit leery of this type of comment that references the structure of the code, since code changes over time and the comment may become outdated and confusing.

Why not move this check a few lines up, so it immediatley follows setting save_path = os.path.dirname(save_path)? The test that checks if the session is a Session can occur after the save path is setup. And the comment will no longer be necessary, since the immediately pre-ceeding line will be the save_path setup.

danmane:
In addition to the line comment - can you add a test to saver_test.py that verifies that the expected exception is raised when saver is given a non-existent path?

tensorflow/python/training/saver.py
PuchatekwSzortach: You have a good point here. I moved the check up and changed it to work on unmodified `os.path.dirname(save_path)`.
I think it's more explicit than setting `save_path = os.path.dirname(save_path)`, which is just bound to cause confusion sooner or later.

benoitsteiner:
Jenkins, test this please.

vrv:

benoitsteiner:
Jenkins, test this please.

tensorflow/python/training/saver.py
muddassir235: save_file = 'model.ckpt'

When I try to do 

```
tf.saver = tf.train.Saver()
with tf.Session() as sess:
     # Do something 
     saver.save(sess, save_file)
```

I get this error: 
```
C:\Program Files\Anaconda3\envs\IntroToTensorFlow\lib\site-packages\tensorflow\python\training\saver.py in save(self, sess, save_path, global_step, latest_filename, meta_graph_suffix, write_meta_graph, write_state)
   1312     if not gfile.IsDirectory(os.path.dirname(save_path)):
   1313       raise ValueError(
-> 1314           "Parent directory of {} doesn't exist, can't save.".format(save_path))
   1315 
   1316     save_path = os.path.dirname(save_path)

ValueError: Parent directory of model.ckpt doesn't exist, can't save.
```

I am using tensorflow 0.12.0 in ancaconda 3 using windows 8.

PuchatekwSzortach: I don't have a Windows machine to try, but

    os.path.exists(os.path.dirname("whatever"))

indeed returns False on my machine. A solution would be to use

    if not gfile.IsDirectory(os.path.abspath(os.path.dirname(save_path))):

on line 1312. I'll make that change.

PuchatekwSzortach: Addressed in https://github.com/tensorflow/tensorflow/pull/6601

PR #3718, opened 2016-08-09
Add complex dtype support to cast

### LINKED ISSUE #3346: Casting from real to complex lacks gradient
The operation of casting from `float32` to `complex64` lacks a gradient:
```
>>> x = tf.Variable(tf.zeros([10]))
>>> tf.gradients(tf.cast(x, dtype=tf.complex64), x)
[None]
```
Surely this is not intentional?

(FWIW, I'm on Tensorflow 0.9.0, running on the CPU on OS X.)

ibab:
This PR enables the `tf.cast` op to convert from/to complex numbers.
This is something that users might expect to work, as it's possible in numpy.
I've also extended the tests and gradient code.

I didn't quite know what to put as the cost for the cast operation, so I picked `NumTraits<To>::AddCost`, which is similar to what's used by `bfloat16`.

This fixes #3346.

### DIFF
--- a/tensorflow/core/kernels/cast_op.h
+++ b/tensorflow/core/kernels/cast_op.h
@@ -43,6 +43,111 @@ struct CastFunctor {
 namespace Eigen {
 namespace internal {
 
+// Eigen can't convert to/from complex numbers, because it is limited to cases
+// that can be static_casted. But numpy is able to cast to/from complex, which
+// we want to emulate. So we add specializations for complex here.
+typedef std::complex<float> complex64;
+typedef std::complex<double> complex128;
+using tensorflow::uint8;
+using tensorflow::int8;
+using tensorflow::uint16;
+using tensorflow::int16;
+using tensorflow::int32;
+using tensorflow::int64;
+
+// These are seperate definitions of what should happen when we cast from/to
+// complex numbers. The actual template specializations are instantiated below.
+template<typename From, typename To>
+struct cast_from_complex_impl {
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE To operator()(const From& a) const {
+    return static_cast<To>(a.real());
+  }
+};
+
+template<typename From, typename To>
+struct cast_to_complex_impl {
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE To operator()(const From& a) const {
+    return To(static_cast<typename To::value_type>(a), typename To::value_type(0));
+  }
+};
+
+template<typename From, typename To>
+struct cast_from_to_complex_impl {
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE To operator()(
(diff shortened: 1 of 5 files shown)


danmane:
@benoitsteiner Can you review?

benoitsteiner:
Jenkins, test this please.

tensorflow/core/kernels/cast_op.h
```diff
@@ -43,6 +43,111 @@ struct CastFunctor {
...
+using tensorflow::uint16;
+using tensorflow::int16;
+using tensorflow::int32;
+using tensorflow::int64;
+
+// These are seperate definitions of what should happen when we cast from/to
+// complex numbers. The actual template specializations are instantiated below.
+template<typename From, typename To>
+struct cast_from_complex_impl {
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE To operator()(const From& a) const {
```
benoitsteiner: Consider adding a comment stating that we're replicating the numpy behavior here.

tensorflow/core/kernels/cast_op.h
```diff
@@ -43,6 +43,111 @@ struct CastFunctor {
...
+    return static_cast<To>(a.real());
+  }
+};
+
+template<typename From, typename To>
```
benoitsteiner: The code would be simpler if you used partial specialization instead: you could for example write this to implement the casting from complex to real.
template <typename From, typename To>
struct scalar_cast_op<std::complex<From>, To> {
 EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE To operator()(const std::complex<From>& a) const {
  return static_cast<To>(a.real());
  }
};

ibab: Ah, I spent a lot of time trying to make a partial specialization approach work.
But I was trying to use SFINAE to specialize for both `complex64` and `complex128` at once, which isn't really possible here.
I should have realized that I can do
```c++
template<typename From, typename To>
struct scalar_cast_op<std::complex<From>, To> {
  To operator()(const std::complex<From>& a) const {
     ...
   }
};
```
as you've suggested.
This works really well, and should make the code a lot more elegant.
I'll update the PR.

danmane:
The test failures so far are known issues; they don't appear related to this PR.

ibab:
I've pushed a new commit that uses partial specialization instead of macros.

benoitsteiner:
Jenkins, test this please.

benoitsteiner:
The test failures appear unrelated, and the CPU tests were already failing when Jenkins started to test this PR (http://ci.tensorflow.org/job/tensorflow-master-cpu/926/). However to be on the safe side I'll wait until TensorFlow tests turn green again and trigger the tests again.

benoitsteiner:
Jenkins, test this please.

benoitsteiner:
Jenkins, test this please.

PR #3741, opened 2016-08-11
Support a separate namespace for the protobuf library

petewarden:
Some clients are already using protobuf v2 as part of their application and would need a lot of work to move, but TensorFlow requires v3. This script patches the protobuf source code after downloading it to put it into a google::protobuf3 namespace, and alters all the necessary TensorFlow code to compile with this new namespace. This allows the v3 library to be linked into applications alongside v2 without causing linker errors.


petewarden:
Jenkins, test this please.

petewarden:
The only testing failure is the unrelated dnn_test, which is a known issue, so I believe the testing for this is passed.

petewarden:
Jenkins, test this please.

tensorflow/contrib/makefile/rename_protobuf.sh
```diff
@@ -0,0 +1,86 @@
...
+
+# This script modifies the downloaded protobuf library and the TensorFlow source
+# to put all protobuf-related symbols into the google::protobuf3 namespace
+# instead of the default google::protobuf. This is necessary to work around
+# linking issues for applications that use protobuf v2 already, and want to
+# adopt TensorFlow, since otherwise the two libraries have duplicate function
+# symbols that clash. It also renames all the include paths to google/protobuf3
+# throughout the protobuf and TensorFlow code.
+# This is a massive hack, and if possible it's recommended that you switch your
+# whole application to protobuf v3 so there's no mismatch with TensorFlow.
```
vrv: (Is it worth mentioning that, although you've gotten something to work, there's no guarantee that this solution is bulletproof?)

vrv:

tensorflow/contrib/makefile/rename_protobuf.sh
petewarden: Good point Vijay, I've added more comments about that to the documentation.

Issue #3464, opened 2016-07-22
Extract the element from the output list of tf.py_func when length 1


omoindrot:
`tf.py_func` could extract the element from the list when the length is 1.

In this example:

    def func(x):
        return 2*x

    x = tf.constant(1.)
    res = tf.py_func(func, [x], [tf.float32])

`res` is a list containing 1 Tensor. This can be confusing, because users expect to get only one tensor as output.

`tf.cond` handles this case well:
> **Returns:**
> Tensors returned by the call to either fn1 or fn2. If the callables return a singleton list, the element is extracted from the list.

yaroslavvb:
Maybe it could be 
```
tf.py_func(func, [x], [tf.float32]) -> returns list of tf.float32
tf.py_func(func, [x], tf.float32) -> returns single tf.float32
```
This is similar to `session.run` which returns either list of numpy arrays or single numpy array depending on whether input is a list

omoindrot:
That would work for me.

concretevitamin:
We are unlikely to work on this soon, so marking as contributions welcome.  IMO @yaroslavvb's suggested behavior is saner than checking for the length of the result list.

suiyuan2009:
dbs

suiyuan2009:
op's attr can't be {type, list(type)}, so I suggest add a boolean to distinguish

suiyuan2009:
@concretevitamin since attr can't be set as {type, list(type)}, if it's set as list(type), then this op will always return [...] instead of ... , I want to make C++ api and python api same as each other, I need help.

TimZaman:
I think #3744 closes this one.

PR #3744, opened 2016-08-11
extract element from list when py_func's output type is a single tensorflow type

suiyuan2009:
#3464


rmlarsen:

zffchen78:
LGTM

Issue #654, opened 2015-12-30
Beam Search 4 Translate.py


mfederico:
Hi, 
I'm wondering which steps are necessary to move from the greedy decoder currently implemented to an actual beam search decoder. Is this enhancement already in someone's roadmap? If not, could anyone tell me which is the right point in the code where to add this functionality?  
Thanks a lot! 
Marcello

marcotrombetti:
I would like to contribute too.

@vrv @lukaszkaiser knowing where the proper place to add it will be very useful.

lukaszkaiser:
When writing the seq2seq module, the idea was that the loop_function argument (e.g. of attention_decoder) could be used to provide various forms of decoding. It works for the greedy case, but we have not implemented a beam-search yet. I think it should be possible using loop_function and the top-k op from tensorflow. But we'll see - if it's too hard to do it inside the graph, then we can change the design and go with a python-side decoder. Having a decoder in the graph has advantages though, esp. when building more complex models, so I'd like to try that first. All ideas, comments, remarks and code are welcome of course!

PrajitR:
@lukaszkaiser 
Here's a self contained example demonstrating a possible beam search implementation:
```python
from __future__ import division
import tensorflow as tf

with tf.Graph().as_default():
    beam_size = 3 # Number of hypotheses in beam.
    num_symbols = 5 # Output vocabulary size.
    embedding_size = 10
    num_steps = 3
    embedding = tf.zeros([num_symbols, embedding_size])
    output_projection = None
    
    # log_beam_probs: list of [beam_size, 1] Tensors
    #  Ordered log probabilities of the `beam_size` best hypotheses
    #  found in each beam step (highest probability first).
    # beam_symbols: list of [beam_size] Tensors 
    #  The ordered `beam_size` words / symbols extracted by the beam
    #  step, which will be appended to their corresponding hypotheses
    #  (corresponding hypotheses found in `beam_path`).
    # beam_path: list of [beam_size] Tensor
    #  The ordered `beam_size` parent indices. Their values range
    #  from [0, `beam_size`), and they denote which previous
    #  hypothesis each word should be appended to.
    log_beam_probs, beam_symbols, beam_path  = [], [], []
    def beam_search(prev, i):
        if output_projection is not None:
            prev = tf.nn.xw_plus_b(
                prev, output_projection[0], output_projection[1])
        
        # Compute 
        #  log P(next_word, hypothesis) = 
[...]

PR #3756, opened 2016-08-11
Beam Search Wrapper

anair13:
I've been working with the beam search implementation @nikitakit wrote to fix issue #654 , and found the outputs aren't exactly what they should be (unless I'm misunderstanding something). It seems the top output is correct but the rest of the beams are not reordered correctly in the computation. This is my fixed version with a more detailed test case to prove correctness, as well as an attention wrapper to be able to use beam search with attention without much effort. I think it would be great to nail down some official solution.

### DIFF
--- a/tensorflow/contrib/beam_search/python/ops/attention_rnn_cell.py
+++ b/tensorflow/contrib/beam_search/python/ops/attention_rnn_cell.py
@@ -1,0 +1,80 @@
+import tensorflow as tf
+
+from tensorflow.python.framework import dtypes
+from tensorflow.python.framework import ops
+from tensorflow.python.ops import array_ops
+from tensorflow.python.ops import control_flow_ops
+from tensorflow.python.ops import embedding_ops
+from tensorflow.python.ops import math_ops
+from tensorflow.python.ops import nn_ops
+from tensorflow.python.ops import rnn
+from tensorflow.python.ops import rnn_cell
+from tensorflow.python.ops import variable_scope
+
+linear = rnn_cell._linear
+
+class AttentionWrapper(tf.nn.rnn_cell.RNNCell):
+    def __init__(self, cell, attention_states, batch_size, embedding, initializer=None, num_heads=1, scope=None):
+        if not isinstance(cell, tf.nn.rnn_cell.RNNCell):
+            raise TypeError("The parameter cell is not RNNCell.")
+        self._cell = cell
+        self._attention_states = attention_states
+        self.embedding = embedding
+
+        with variable_scope.variable_scope(scope or "attention_decoder"):
+            # batch_size = attention_states.get_shape()[0].value
+            attn_length = attention_states.get_shape()[1].value
+            attn_size = attention_states.get_shape()[2].value
+
+            hidden = array_ops.reshape(
+            attention_states, [-1, attn_length, 1, attn_size])
+            hidden_features = []
+            v = []
+            attention_vec_size = attn_size  # Size of query vectors for attention.
+            for a in xrange(num_heads):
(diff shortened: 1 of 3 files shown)


therne:
`cand_symbols` and `cand_probs` are not updated? unwrap functions use these states as output but I can't see any updating code...

vrv:
@ebrevdo can you take a look at this?  Does this make sense for TensorFlow core, or does this maybe belong in a separate, more focused repo?

ebrevdo:
@lukaszkaiser let's discuss this one on Monday.

lukaszkaiser:
Ok, maybe around 3pm?

Lukasz

ebrevdo:
Unfortunately, without the original author of the beam search code having provided a clear license in their original gist/github, we cannot accept code except from that author.

That said, the attention wrapper looks like it's original (is it?  if not, do you have details about the original license?)  So if you want, you can submit that on its own with corresponding unit tests for review.  Please look at some other code and unit tests to ensure you conform to code and documentation guidelines.

Thanks!

nikitakit:
I want to chime in here and say that I'm happy to have my code included in contrib, just like I said in the original issue.

I've been away on vacation until recently, which is why you've only been hearing from @anair13 so far.

@ebrevdo I can resubmit my parts under my name if you'd like. (Though I wouldn't be comfortable doing so without first fixing the bugs and getting the tests into shape.) That said, I would find it extremely useful to hear some comments on the high-level approach (i.e. cell wrapping, as opposed to dynamic loops or custom ops) and API design before writing any more code.

anair13:
Thanks for looking at this everyone. As @nikitakit said, I think it would be good to discuss the approach here (personally, I think this way of implementing both attention and beam search as wrappers is simpler to understand and easier to modify) and make changes as we go.

I can fix the issue mentioned by @therne as I had commented out parts I had not tested, and also contribute some new tests and a sequence to sequence model with attention and beam search.

ebrevdo:
Have you seen the `AttentionCellWrapper` we have in [contrib.rnn](https://github.com/tensorflow/tensorflow/blob/ad9e5b29aeee6638476cb47522bc880876eaab86/tensorflow/contrib/rnn/python/ops/rnn_cell.py#L670)?

`tf.nn.raw_rnn` is now safe to use; so any decoding algorithms should probably use it.

ebrevdo:
@nikitakit if you could take a look at AttentionCellWrapper and `raw_rnn` and let us know whether you would like to add onto these or build on top of these, we can decide whether your work should go into core, contrib, or is better suited for its own, higly focused, "model" repo.

nikitakit:
@ebrevdo All attention code here is by @anair13. To be honest I'd prefer to keep the beam search and attention discussions separate, because their implementations should be orthogonal. (The beam search code can work with any attention implementation, including the built-in AttentionCellWrapper).

I would love to build on top of `raw_rnn`, but unfortunately its API is still not expressive enough to support beam search. Beam search needs to have access to intermediate cell states (not just inputs/outputs), because items that fall off the beam need to have their cell state removed and replaced with that of another item. 

Now if `raw_rnn` could be modified to receive `cell_state` into the loop function and output a `new_cell_state` from it, then it's a different story.

anair13:
Thanks for the reply @ebrevdo. I hadn't seen the AttentionCellWrapper before, and thats exactly what the one included here does so I'll take it out.

ebrevdo:
Nikita I'll consider adding control of state in raw_rnn.  The interface is
still flexible

nikitakit:
@ebrevdo If you could add state control to raw_rnn, that would be much appreciated. Using that API should also provide performance improvements for beam search, because it will allow early termination using dynamic_rnn primitives.

PR #3778, opened 2016-08-12
Reduce memory usage and increase performance for convolution on iOS

petewarden:
We've had lots of problems with large convolutions hitting memory limits on iOS. This new implementation of the operator breaks the work into chunks so we never use more than 16 MB, and uses Apple's Accelerate framework to optimize the matrix multiplication.

Testing shows that it's between 5% to 10% faster than the existing implementation on various models, and keeps memory usage to a minimum.


bhack:
For quantized models will be used gemmlowp also on IOS? /cc @wangyida

petewarden:
@bhack gemmlowp can be used on iOS, though we haven't investigated optimizing it for those devices in particular so I expect we'll need to do more work there. This is primarily a fix for memory issues when running float models.

petewarden:
Jenkins, test this please.

bhack:
Could be interesting to benchmark this against [BNNS](https://developer.apple.com/reference/accelerate/1912851-bnns).

Wangyida:
@bhack I can apply gemm on FC layer in tiny-dnn of IOS platform now, some memory issue seems related to the batch size and network structure rather than the parametric model itself.

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+// This file contains a set of different implementations of the two-dimensional
+// convolution operation. The standard TensorFlow Conv2d kernel uses EigenTensor
+// to implement the computation, but here there are a variety of different ways
```
cwhipkey: change "here there" to "there"

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+#include "tensorflow/core/framework/tensor_slice.h"
+#include "tensorflow/core/kernels/bounds_check.h"
+#include "tensorflow/core/util/padding.h"
+#include "tensorflow/core/util/tensor_format.h"
+
+#if defined(__APPLE__)
+#include <Accelerate/Accelerate.h>
+#define USE_ACCELERATE_GEMM
+#endif  // __APPLE__
+
```
cwhipkey: Can you add a test that uses the new functor, so we can make sure the padding/stride computations match those of the existing implementation?  One way may be to register the kernel in all cases but register it with a different name or a specific label when it's not meant to be the default conv implementation

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+    // produce output values where the filters can read all their values from
+    // within the input image. It effectively removes the margins of the output
+    // image compared to the one produced by SAME. Stride complicates this
+    // definition though, because it can result in the right and bottom filter
+    // patches sampling from outside the borders if it's greater than 1.
+    // Most of the logic for sorting this all out is done before this function,
+    // when we calculate the output size, but the positioning of the origin of
+    // the filters is different between the two modes, since SAME positions the
+    // first filter off the edge of the input.
+    int filter_left_offset;
```
cwhipkey: use int64 for filter_left_offset and filter_top_offset

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+// memory order very explicit, which is important for both inspecting memory
+// contents during debugging and for specifying what we expect to others.
+// The memory layout of the data is, from biggest stride to smallest:
+// input_data = [input_batches, input_height, input_width, input_depth]
+// filter_data = [filter_height, filter_width, input_depth, filter_count]
+// output_data = [input_batches, output_height, output_width, filter_count]
+template <class T1, class T2, class T3>
+class ReferenceConvFunctor {
+ public:
+  void operator()(OpKernelContext* context, const T1* input_data,
```
cwhipkey: int is used in many places.  Should it be int64, or check the width and the height fit in int32?  (I haven't checked below - maybe there is such a check).

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+      filter_top_offset =
+          ((output_height - 1) * stride_rows + filter_height - input_height) /
+          2;
+    }
+
+    // If we've got multiple images in our input, work through each of them.
+    for (int batch = 0; batch < input_batches; ++batch) {
+      // Walk through all the output image values, sliding the filter to
+      // different
+      // positions in the input.
```
cwhipkey: move this line to previous line

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+        const size_t c_index = ((i * c_i_stride) + (j * c_j_stride));
+        c[c_index] = total;
+      }
+    }
+  }
+};
+
+// Uses the optimized Eigen library to implement the matrix multiplication
+// required by the Im2ColConvFunctor class.
+template <class T1, class T2, class T3>
```
cwhipkey: why T1,T2,T3 instead of a single template parameter T ?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+    // count |                     |
+    //   v   +---------------------+
+    // Each patch row contains a filter_width x filter_height patch of the
+    // input, with the depth channel as the most contiguous in memory, followed
+    // by the width, then the height. This is the standard memory order in the
+    // image world if it helps to visualize it.
+    const int filter_value_count = filter_width * filter_height * input_depth;
+    const int patch_count = input_batches * output_width * output_height;
+
+    CHECK_GT(patch_count, 0);
```
cwhipkey: I've seen it mentioned elsewhere that we should not CHECK inside tf.  Some kernels do CHECK, but it seems most try to use OP_REQUIRES_OK?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+    CHECK_GT(filter_count, 0);
+    CHECK_GT(filter_value_count, 0);
+
+    // We don't want to allocate a buffer to hold all the patches if the size is
+    // going to be extremely large, so break it into chunks if it's bigger than
+    // a limit. Each chunk will be processed serially, so we can refill the
+    // buffer for the next chunk and reuse it, keeping maximum memory size down.
+    // In this case, we've picked 16 megabytes as a reasonable limit.
+    const size_t max_chunk_size = (16 * 1024 * 1024);
+    const size_t patches_per_chunk =
```
cwhipkey: make sure patches_per_chunk >= 1 ?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+    // allocate a persistent buffer that will be kept around between calls. We
+    // use TensorFlow's resource management to ensure that the memory will be
+    // released when the session is over.
+    Im2ColBufferResource<T1, max_chunk_size>* im2col_buffer_resource;
+    std::function<Status(Im2ColBufferResource<T1, max_chunk_size>**)> creator =
+        [](Im2ColBufferResource<T1, max_chunk_size>** resource) {
+          *resource = new Im2ColBufferResource<T1, max_chunk_size>();
+          return Status::OK();
+        };
+    TF_CHECK_OK(context->resource_manager()->LookupOrCreate(
```
cwhipkey: Does this also then assume that only one of the kernels is working at once?  Can we guard against misuse in this case?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+          input_data + (batch * input_height * input_width * input_depth);
+      for (int out_y = 0; out_y < output_height; ++out_y) {
+        const int in_y_origin = (out_y * stride_rows) - filter_top_offset;
+        for (int out_x = 0; out_x < output_width; ++out_x) {
+          const int in_x_origin = (out_x * stride_cols) - filter_left_offset;
+          const int patch_index = (batch * output_width * output_height) +
+                                  (out_y * output_width) + out_x;
+          const int chunk_index = patch_index % patches_per_chunk;
+          T1* im2col_patch_start =
+              im2col_buffer /*.get()*/ + (chunk_index * filter_value_count);
```
cwhipkey: remove the .get() comment?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+    for (int batch = 0; batch < input_batches; ++batch) {
+      const T1* input_batch_start =
+          input_data + (batch * input_height * input_width * input_depth);
+      for (int out_y = 0; out_y < output_height; ++out_y) {
+        const int in_y_origin = (out_y * stride_rows) - filter_top_offset;
+        for (int out_x = 0; out_x < output_width; ++out_x) {
+          const int in_x_origin = (out_x * stride_cols) - filter_left_offset;
+          const int patch_index = (batch * output_width * output_height) +
+                                  (out_y * output_width) + out_x;
+          const int chunk_index = patch_index % patches_per_chunk;
```
cwhipkey: chunk_index was confusing to me - I thought there would be a single chunk at once?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+            const int k = filter_value_count;
+            const int lda = filter_value_count;
+            const int ldb = filter_count;
+            const int ldc = filter_count;
+            const size_t start_patch_index =
+                patch_index - (how_many_patches - 1);
+            T3* chunk_output_data =
+                output_data + (start_patch_index * filter_count);
+            TGemmFunctor gemm_functor;
+            gemm_functor(m, n, k, im2col_buffer /*.get()*/, lda, filter_data,
```
cwhipkey: remove .get() comment?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+                         ldb, chunk_output_data, ldc);
+          }
+        }
+      }
+    }
+
+    // Finish up any remaining work left over from the loop.
```
cwhipkey: if it would help to have fewer copies of this code, this could be merged into the main loop (e.g. update patches_per_chunk when the last patch is visited).

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+                    im2col_row_start +
+                    ((left_zero_count + center_copy_count) * input_depth);
+                T1* im2col_right_end =
+                    im2col_right_start + (right_zero_count * input_depth);
+                std::fill(im2col_right_start, im2col_right_end, T1(0));
+              }
+            }
+          }
+          // Now we've assembled a set of image patches into a matrix, apply a
+          // GEMM to them to get partial results in the output matrix.
```
cwhipkey: it might be useful to say what the GEMM is of, like     [a, b] X [b, c]   (defining a,b,c in there).

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+  explicit Conv2DUsingGemmOp(OpKernelConstruction* context)
+      : BinaryOp<T>(context) {
+    OP_REQUIRES_OK(context, context->GetAttr("strides", &strides_));
+    string data_format;
+    OP_REQUIRES_OK(context, context->GetAttr("data_format", &data_format));
+    OP_REQUIRES(context, FormatFromString(data_format, &data_format_),
+                errors::InvalidArgument("Invalid data format"));
+    OP_REQUIRES(
+        context, data_format_ == FORMAT_NHWC,
+        errors::InvalidArgument("Data format not supported by this kernel"));
```
cwhipkey: maybe do  ...by this kernel: ", data_format);

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+        errors::InvalidArgument("input and filter must have the same depth: ",
+                                in_depth, " vs ", filter.dim_size(2)));
+
+    // The last dimension for filter is out_depth.
+    const int out_depth = static_cast<int>(filter.dim_size(3));
+
+    // The second dimension for input is rows/height.
+    // The first dimension for filter is rows/height.
+    const int64 input_rows_raw = GetTensorDim(input, data_format_, 'H');
+    OP_REQUIRES(context, FastBoundsCheck(input_rows_raw,
```
cwhipkey: This doesn't validate the filter rows+cols.  Should it?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+ public:
+  void operator()(size_t m, size_t n, size_t k, const T1* a, size_t lda,
+                  const T2* b, size_t ldb, T3* c, size_t ldc) {
+    Eigen::Map<const Eigen::Matrix<T1, Eigen::Dynamic, Eigen::Dynamic,
+                                   Eigen::RowMajor>>
+        a_matrix(a, m, k);
+    Eigen::Map<const Eigen::Matrix<T2, Eigen::Dynamic, Eigen::Dynamic,
+                                   Eigen::RowMajor>>
+        b_matrix(b, k, n);
+    Eigen::Map<
```
andydavis1: Maybe typedef these: 

public:
  typedef Eigen::Map<const Eigen::Matrix<T2, Eigen::Dynamic, Eigen::Dynamic,
￼                                   Eigen::RowMajor>> ConstMatrixMap;
  typedef Eigen::Map<Eigen::Matrix<T3, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>> MatrixMap
   void operator()(size_t m, size_t n, size_t k, const T1* a, size_t lda,
￼ +                  const T2* b, size_t ldb, T3* c, size_t ldc) {
        ConstMatrix a_matrix(a, m, x);
         ...                                        
￼

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+        "Conv2d", "im2col_buffer", &im2col_buffer_resource, creator));
+    // This means that multiple ops can't be run simultaneously on different
+    // threads, because we have a single shared resource. The platforms this is
+    // aimed at have intra-op parallelism as their focus though, so it shouldn't
+    // be an issue.
+    mutex_lock lock_buffer(im2col_buffer_resource->mu);
+    core::ScopedUnref unref_buffer(im2col_buffer_resource);
+    T1* im2col_buffer = im2col_buffer_resource->data;
+
+    for (int batch = 0; batch < input_batches; ++batch) {
```
andydavis1: There is an Im2Col implementation at the top of conv_grad_ops.cc, that used to be exposed in a header file (could be re-exposed if used elsewere). Any chance that could be re-used here for part of this loop?

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+    // We don't want to allocate a buffer to hold all the patches if the size is
+    // going to be extremely large, so break it into chunks if it's bigger than
+    // a limit. Each chunk will be processed serially, so we can refill the
+    // buffer for the next chunk and reuse it, keeping maximum memory size down.
+    // In this case, we've picked 16 megabytes as a reasonable limit.
+    const size_t max_chunk_size = (16 * 1024 * 1024);
+    const size_t patches_per_chunk =
+        max_chunk_size / (filter_value_count * sizeof(T1));
+    const size_t im2col_size = patches_per_chunk * filter_value_count;
+    // Because memory allocation is very expensive on mobile platforms, try to
```
andydavis1: What would happen if you allocated a using the DeviceContext::allocate_temp? Allocate a temporary tensor once (before the loop), then re-use it in the loop...

andydavis1: NEVERMIND. I remember our previous conversation, that you needed this memory to live across OpKernel invocations....

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,612 @@
...
+          *resource = new Im2ColBufferResource<T1, max_chunk_size>();
+          return Status::OK();
+        };
+    TF_CHECK_OK(context->resource_manager()->LookupOrCreate(
+        "Conv2d", "im2col_buffer", &im2col_buffer_resource, creator));
+    // This means that multiple ops can't be run simultaneously on different
+    // threads, because we have a single shared resource. The platforms this is
+    // aimed at have intra-op parallelism as their focus though, so it shouldn't
+    // be an issue.
+    mutex_lock lock_buffer(im2col_buffer_resource->mu);
```
andydavis1: Holding this lock for the entire op scares me. Can you change the Im2ColBufferResource to do this locking internally (around some simple free list), and return a buffer to the caller? This way, you wont have to hold a lock for the entire op, and concurrent access to the Im2ColBufferResource can be managed internally (i.e. it could just allocate another buffer to a concurrent caller, rather than just blocking).
What kinds of concurrent access patterns do you expect?  You could serialize things with "intra_op_threads" flag (if needed).

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: I've updated the line to read "this module has a variety...". Is that clearer?

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: We've discussed this off-line. The ideal solution is to run all our tests with multiple implementations, but it's not clear how to do that right now. I've manually run the complete set of OS X tests with all the functor combinations, and they pass, so that may be the best approach for now

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Good point. We do actually make sure that the int64 input parameters fit into int's as part of the common Compute() code below.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Done.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: With lower-precision formats, it's common to need to accumulate into higher-precision variable (e.g. float16 inputs, with a float32 output, or uint8 inputs with a uint32 output). Even though the currently-supported types don't take advantage of this, I have a mild preference for keeping the types specified separately. I've added a comment to make this a bit clearer, if that helps?

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Good idea, I've taken that approach and it cleans up the code.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Thanks for the tip, I've now moved most of the checks to the top of the function now, and made it an early return. I am able to call OP_REQUIRES_OK for the resource allocation though, since I have a context at this point, so I added that in too.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Done.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Yes, that's right, the goal is to allocate once per graph, so that overall memory usage is minimized.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: For the particular use case I'm trying to address, memory is an extremely scarce resource, and also the expected pattern is to have no inter-op parallelism, only intra-op. This seems to be the best way to get conv performance on the mobile and embedded platforms I'm focused on, and from talking to XQ that seems true even for the server GPU side. My preference would be to stick to this blocking approach for now, unless we run into situations where it's a problem. What do you think?

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: That implementation only fills in a single 'pixel' at a time, whereas we've found we get best performance by copying or setting a row of pixels in one call. That does mean some extra complexity in the calculations used by this version though.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Good point, I've changed this to patch_index_within_chunkl

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Done.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: I've expanded the comment a bit more.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Done.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Good idea, done.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: Done.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: The filter rows and cols get checked above, in a loop over the filter dims.

tensorflow/core/kernels/conv_ops_using_gemm.cc
andydavis1: You could make the max free list size 1, and have callers block and wait if the free list size is zero.  Then it's clear that callers are serializing on that resource. Anyway, maybe it's a minor nitpick, so up to you...

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,623 @@
...
+
+    // We don't want to allocate a buffer to hold all the patches if the size is
+    // going to be extremely large, so break it into chunks if it's bigger than
+    // a limit. Each chunk will be processed serially, so we can refill the
+    // buffer for the next chunk and reuse it, keeping maximum memory size down.
+    // In this case, we've picked 16 megabytes as a reasonable limit.
+    const size_t max_chunk_size = (16 * 1024 * 1024);
+    OP_REQUIRES(context, (filter_value_count * sizeof(T1)) <= max_chunk_size,
+                errors::InvalidArgument("Im2Col patch too large for buffer"));
+    const size_t patches_per_chunk =
```
andydavis1: Could this be zero for a very large filter? I've seen large filters sizes that get close to this limit. You probably want to check your self here and have a min of one:
  const size_t patches_per_chunk = min(1,
    max_chunk_size / (filter_value_count * sizeof(T1));
)

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,623 @@
...
+    const int filter_value_count = filter_width * filter_height * input_depth;
+    const int patch_count = input_batches * output_width * output_height;
+
+    // We don't want to allocate a buffer to hold all the patches if the size is
+    // going to be extremely large, so break it into chunks if it's bigger than
+    // a limit. Each chunk will be processed serially, so we can refill the
+    // buffer for the next chunk and reuse it, keeping maximum memory size down.
+    // In this case, we've picked 16 megabytes as a reasonable limit.
+    const size_t max_chunk_size = (16 * 1024 * 1024);
+    OP_REQUIRES(context, (filter_value_count * sizeof(T1)) <= max_chunk_size,
```
andydavis1: Could pull filter_value_count * sizeof(T1) out into a constant and re-use it below.

andydavis1:
Looks good to me (my latest round of comments were minor)...

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -0,0 +1,623 @@
...
+    }
+  }
+};
+
+// A readable but slow implementation of matrix multiplication, useful for
+// debugging and understanding the algorithm. Use instead of FastGemmFunctor in
+// the Im2ColConvFunctor template definition inside the op registration to
+// enable. Assumes row-major ordering of the values in memory.
+template <class T1, class T2, class T3>
+class ReferenceGemmFunctor {
```
rmlarsen: Why do we need to include this? The problem with including slow reference implementations, is that they end up being used and are hard to get rid of.

petewarden: We discussed this offline, but to summarize it's useful for bootstrapping porting to new platforms, though I agree it's a little awkward here.

petewarden:
Once the tests have passed, could the admins merge this since we have LGTMs?

PR #3779, opened 2016-08-12
Allows running "-m tensorflow.tensorboard"

gustavla:
This is a suggested change that moves the main entry point of Tensorboard from `tensorboard.py` to `__main__.py`, allowing tensorboard to be run as a package:

```
python -m tensorflow.tensorboard --logdir=/path
```

This can be really useful on environments with multiple versions of python. Of course, virtualenv is a much better way of dealing with this, but I still find this option to be useful at times.

For instance, if you maintain both Python 2 and Python 3 and do a `pip install --user` install of tensorflow for both versions. The python files will be neatly separated by the version in `~/.local/lib/pythonX.X/site-packages`. However, there is no such mechanism for `~/.local/bin`, so the second pip install will override the first one. You can also get around this by running the `tensorboard.py` file directly, but this is not nearly as convenient and safe as letting python specify the location of the package.

Another less drastic option here is to keep `tensorboard.py` and add a new file for `__main__.py` that simply imports `main` from `tensorboard.py` and runs it. I'm happy to arrange such a PR too.


vrv:
I think I would prefer the last approach, instead of moving the contents from tensorboard.py.

Assigning to Dan

gustavla:
Closing in favor of #3811.

PR #3780, opened 2016-08-12
Added hungarian layer.

ghost:
Please see initial discussion [here](https://github.com/Russell91/TensorBox/issues/47) for reference. I am submitting a custom layer that is necessary to implement [this paper](http://www.cv-foundation.org/openaccess/content_cvpr_2016/papers/Stewart_End-To-End_People_Detection_CVPR_2016_paper.pdf) from CVPR 2016. Users of [Tensorbox](https://github.com/Russell91/TensorBox) currently must compile Tensorflow from source to use these features, so accepting a variant of this pull request would at least help out those existing users. It would ideally be useful to others as well.

That being said, there are 2 known problems with the pull request:

1) The code is messy
2) The hungarian algorithm library itself is not something I can contribute under the Tensorflow license. There is no publicly available hungarian implementation that would satisfy this constraint either. I am holding out hope that Google may have a private implementation of the hungarian algorithm that could be packaged into the repository and satisfy the license constraints. If not, and the only way forward is to write a new implementation, someone (perhaps me) could write a novel hungarian c++ implementation from the details on wikipedia.

Problem (2) has prevented me from really worrying about problem (1).

### DIFF
--- a/tensorflow/core/user_ops/hungarian.cc
+++ b/tensorflow/core/user_ops/hungarian.cc
@@ -1,0 +1,755 @@
+#include <assert.h>
+#include "tensorflow/core/framework/op.h"
+#include "tensorflow/core/framework/op_kernel.h"
+
+using namespace tensorflow;
+using std::vector;
+
+#ifndef HUNGARIAN_H
+#define HUNGARIAN_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+  
+#define HUNGARIAN_NOT_ASSIGNED 0 
+#define HUNGARIAN_ASSIGNED 1
+
+#define HUNGARIAN_MODE_MINIMIZE_COST   0
+#define HUNGARIAN_MODE_MAXIMIZE_UTIL 1
+
+#define MIN(a,b) (((a)<(b))?(a):(b))
+#define MAX(a,b) (((a)>(b))?(a):(b))
+
+using std::vector;
+
+class Rect {
+ public:
+  int cx_;
+  int cy_;
+  int width_;
+  int height_;
+  float confidence_;
+  float true_confidence_;
+
+  explicit Rect(int cx, int cy, int width, int height, float confidence) {
+    cx_ = cx;
+    cy_ = cy;
+    width_ = width;
+    height_ = height;
+    confidence_ = confidence;
+    true_confidence_ = confidence;
+  }
+
+  Rect(const Rect& other) {
+    cx_ = other.cx_;
+    cy_ = other.cy_;
+    width_ = other.width_;
+    height_ = other.height_;
+    confidence_ = other.confidence_;
+    true_confidence_ = other.true_confidence_;
+  }
+
+  bool overlaps(const Rect& other) const {
+    if (fabs(cx_ - other.cx_) > (width_ + other.width_) / 1.5) {
+      return false;
+    } else if (fabs(cy_ - other.cy_) > (height_ + other.height_) / 2.0) {
+      return false;
+    } else {
+      return iou(other) > 0.25;
+    }
+  }
+
+  int distance(const Rect& other) const {
+    return (fabs(cx_ - other.cx_) + fabs(cy_ - other.cy_) +
+            fabs(width_ - other.width_) + fabs(height_ - other.height_));
+  }
+
+  float intersection(const Rect& other) const {
+    int left = MAX(cx_ - width_ / 2., other.cx_ - other.width_ / 2.);
+    int right = MIN(cx_ + width_ / 2., other.cx_ + other.width_ / 2.);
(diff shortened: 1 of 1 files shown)


vrv:
Thanks for the pull request.  Unfortunately the licensing issues are a blocker.

Perhaps we can sidestep this issue?  I think we now have a new way of supporting user-implemented ops now -- we should no longer need the user_ops directory to load custom ops, and we should figure out a way for your users to be able to build and load your custom op without having it be added to the official TF repository.  https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/how_tos/adding_an_op/index.md#with-tensorflow-binary-installation should be the way you can get this done.

Assigning to @keveman so he can help with custom ops issues.

vrv:
(Marking as CLA: no so this doesn't get merged by accident)

Russell91:
Okay, cool that there is an additional pathway under development. It seems that each user would still have to compile the op from source on their machine (and install bazel) though?

Also, were there no CLA issues, what would be the probability of getting this sort of op in? I didn't mean to imply in my initial post that recoding a hungarian algorithm would be some sort of intractable effort. Everything outside that contiguous region of code is 100% owned by me.

vrv:
Well, the hope is that users don't have to install bazel -- just need to run 'gcc' on the custom op that, say, lives in the TensorBox library (which you could package up via pip yourself, etc).  That's the situation I'd love to be in :)

As for its suitability in TF: we could get a set of opinions on this, but speaking as someone who has to essentially forever maintain any code that gets into TF core, I'm always wary of having to maintain (and handle bug reports for, debug) code for ideas that aren't "core" functions of TensorFlow.  I guess I'd have to understand how generally this op could be used.  One idea is that, we leave this external for now, and as more feature requests come in that could be solved by having this op in the core, we add it to core at that point.  What do you think?

Russell91:
@vrv Makes sense to me. I guess we'll just have to keep an eye on this as the extension system matures?

keveman:
@Russell91, perhaps you didn't fully grok what @vrv was trying to say. The extension system is indeed quite mature. Look [here](https://www.tensorflow.org/versions/r0.10/how_tos/adding_an_op/index.html) for complete instructions on how to add an op to TensorFlow. You have two options, depending on how you want your op to be compiled and used.

- Add it to tensorflow/contrib and send a PR. See [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/linear_optimizer/BUILD) for an example BUILD file that compiles a custom op and makes it available TensorFlow users via `tensorflow.contrib`.
- Maintain your op's source code independent of TensorFlow's source code. In this case, you would need the users to install the binary version of TensorFlow via the PIP package. Look [here](https://www.tensorflow.org/versions/r0.10/how_tos/adding_an_op/index.html#with-tensorflow-binary-installation) for how to compile your custom op without using the Bazel build system, say, using plain vanilla Makefile.

Let me know if you need help with either of those.

vrv:
@keveman: tensorflow.contrib is for stuff that is likely to go into core, and as discussed, we're not yet ready for that.

But that second option is what I was trying to suggest, yes.

keveman:
Whoops sorry for the confusion.
What I had in mind was the ability to use the Bazel build system to build the custom op. @Russell91, you can follow the model used by some projects in tensorflow/models, look [here](https://github.com/tensorflow/models/tree/master/syntaxnet)  for example. You setup your custom op as an independent git repository with its own BUILD file, and use TensorFlow as an external project.

vrv:
I still think he would prefer to use a Makefile in his own project to build the custom op, which is also possible.

Russell91:
@kaveman My quote 

> I guess we'll just have to keep an eye on this as the extension system matures?

was a reaction to the comment of @vrv that 

> Well, the hope is that users don't have to install bazel -- just need to run 'gcc' on the custom op that, say, lives in the TensorBox library (which you could package up via pip yourself, etc). That's the situation I'd love to be in :)

I guess I'll have to look into exactly what is possible with the current system, but you have to understand that I'm not only interested in a system which is _possible_ for users to add the hungarian layer, but one where it is _easy_. That's just a much higher bar, and say, even requiring them to install bazel would miss the mark entirely. Bazel is so rapidly developed that there's at least a 20% chance of an install taking > 15 minutes due to versioning problems. So I'm happy to learn about a bazel extension system, and it may indeed be much better than forcing them to recompile tensorflow from scratch (as they currently have to with my project). But @vrv is absolutely correct that I would prefer to just have my users say "make hungarian", which would call the native gcc to create a .so, and link that in at runtime with some python magic than to have them muck around with a brand new build system (bazel). Note also that my project currently suggests people to get started by ignoring the hungarian layer and just use a simpler model.

PR #3792, opened 2016-08-13
Fix Eigen-related compilation error when using -mavx2.

darrengarvey:
Since updating to use a newer version of Eigen [1], compilation with
-mavx2 causes an error caused by an undefined type. Eigen removed
the type `scalar_multipl2_op` [2] so `scalar_product_op` should now
be used.

This commit replaces uses of `scalar_multiple2_op` similarly to how
[2] did this.

[1] See chage 127264575, commit 10211a6c.
[2] See https://bitbucket.org/eigen/eigen/commits/03556a17eb548275bc9404d7cda8303ff6ca5c13

### DIFF
--- a/third_party/eigen3/unsupported/Eigen/CXX11/src/FixedPoint/PacketMathAVX2.h
+++ b/third_party/eigen3/unsupported/Eigen/CXX11/src/FixedPoint/PacketMathAVX2.h
@@ -344,29 +344,29 @@ EIGEN_STRONG_INLINE QInt8 predux_max<Packet32q8i>(const Packet32q8i& a) {
 
 // Vectorized scaling of Packet32q8i by float.
 template<>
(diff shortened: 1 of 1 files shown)


darrengarvey:
It would be good if someone familiar with Eigen could review this. This fixes #3792.

benoitsteiner:
Jenkins, test this please.

terrytangyuan:
@benoitsteiner there seems to be problem with Jenkins or infrastructure, which is also happening to other PRs.

darrengarvey:
Does normal testing ever set -mavx2?

benoitsteiner:
Jenkins, test this please.

vrv:

I don't think all of our test machines have AVX2, so yes, we don't test with all possible architectures, sadly.

Issue #3800, opened 2016-08-14
Problems in Tensorflow Seq2Seq Tutorial


sahiliitm:
### Environment info
Operating System: Ubuntu 14.04

Installed version of CUDA and cuDNN: 
Cuda 7.5 and CuDNN v4
Output of ls -l /path/to/cuda/lib/cuda*:
```
-rwxrwxrwx 1 root root   322936 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudadevrt.a
lrwxrwxrwx 1 root root       16 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart.so -> libcudart.so.7.5
lrwxrwxrwx 1 root root       19 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart.so.7.5 -> libcudart.so.7.5.18
-rwxrwxrwx 1 root root   383336 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart.so.7.5.18
-rwxrwxrwx 1 root root   720192 Aug 15  2015 /usr/local/cuda-7.5/lib64/libcudart_static.a
-rwxrwxrwx 1 root root 61272736 Jul 31 21:45 /usr/local/cuda-7.5/lib64/libcudnn.so
-rwxrwxrwx 1 root root 61272736 Jul 31 21:45 /usr/local/cuda-7.5/lib64/libcudnn.so.4
-rwxrwxrwx 1 root root 61272736 Jul 31 21:45 /usr/local/cuda-7.5/lib64/libcudnn.so.4.0.4
```

pip package installed from:
```
export TF_BINARY_URL=https://storage.googleapis.com/tensorflow/linux/gpu/tensorflow-0.10.0rc0-cp27-none-linux_x86_64.whl
```
The output from `python -c "import tensorflow; print(tensorflow.__version__)"`.
```
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcublas.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcudnn.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcufft.so locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcuda.so.1 locally
I tensorflow/stream_executor/dso_loader.cc:108] successfully opened CUDA library libcurand.so locally
0.10.0rc0
```

The problem:
[...]

PR #3802, opened 2016-08-14
seq2seq.py: Fixed the documentation to be consistent with the code

sahiliitm:
A fix for https://github.com/tensorflow/tensorflow/issues/3800

### DIFF
--- a/tensorflow/python/ops/seq2seq.py
+++ b/tensorflow/python/ops/seq2seq.py
@@ -318,8 +318,9 @@ def embedding_rnn_seq2seq(encoder_inputs,
   Returns:
     A tuple of the form (outputs, state), where:
       outputs: A list of the same length as decoder_inputs of 2D Tensors with
-        shape [batch_size x num_decoder_symbols] containing the generated
-        outputs.
+        shape [batch_size x output_size] containing the generated
+        outputs. Each row of the 2D tensors is a dense representation of the
+        predicted token.
       state: The state of each decoder cell in each time-step. This is a list
         with length len(decoder_inputs) -- one item for each time-step.
         It is a 2D Tensor of shape [batch_size x cell.state_size].
@@ -1083,7 +1084,8 @@ def model_with_buckets(encoder_inputs, decoder_inputs, targets, weights,
   Returns:
     A tuple of the form (outputs, losses), where:
       outputs: The outputs for each bucket. Its j'th element consists of a list
-        of 2D Tensors of shape [batch_size x num_decoder_symbols] (jth outputs).
+        of 2D Tensors of shape [batch_size x output_size] (jth outputs). Each
+        row of the 2D tensors is a dense representation of the predicted token.
       losses: List of scalar Tensors, representing losses for each bucket, or,
         if per_example_loss is set, a list of 1D batch-sized float Tensors.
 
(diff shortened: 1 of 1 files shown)


sahiliitm:
I signed it!

lukaszkaiser:
Hi sahiliitm . I looked at your changed comments, but I don't understand them. What is `output_size`, what part of arguments is it? Are the current comments not correct? Can you explain?

sahiliitm:
Hi Lukas,

I have documented the bug [here](https://github.com/tensorflow/tensorflow/issues/3800). Please let me know if it is not clear. :) 
In short, the outputs of the two functions (for which I have changed the comments) do not return tensors of size `(batch_size, num_decoder_symbols)` but rather of size `(batch_size, output_size)` where `output_size` is the same parameter as [documented for the function](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/seq2seq.py#L252) `embedding_rnn_decoder`

lukaszkaiser:
Oh, I see, what is meant by output_size is cell.output_size (the comment just missed that). But I think you're not fully correct -- the functions do output num_decoder_symbol-sized tensors sometimes, it depends on the output_projection. If you're correcting it, please make it fully correct. I'll be away from github for a few weeks now, so there might be a delay in my answers, sorry.

sahiliitm:
Are you sure that is the case? 
The reason I doubt that is because of this [line of code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/models/rnn/translate/seq2seq_model.py#L161). We do pass the parameters `output_projection` to the function `tf.nn.seq2seq.model_with_buckets` (indirectly through the function defined at [this line](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/models/rnn/translate/seq2seq_model.py#L125)) in [seq2seq_model.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/models/rnn/translate/seq2seq_model.py) but the output returned by it is still a `cell.output_size` sized tensor. Which is why the line of code I pointed out is needed to project it back to the size `num_decoder_symbols`.

I guess what I'm asking is that if `output_projection` parameters are not `None`, is that the case when the functions under question would return `num_decoder_symbol` sized tensors? That does not seem to be the case.

lukaszkaiser:
I think the seq2seq module behaves like you say in the end: if output_projection is not None, it returns cell.output_size - sized tensors. If it is None, it returns num_decoder_symbol - sized tensors. Is it not so?

The seq2seq_model example is a bit different. It uses sampled softmax, but that's only good for training. When decoding, it cannot sample from the softmax, so it just returns the dense logits. But that's seq2seq_model -- let's keep it separate from this issue, which is changing the seq2seq module.

Or am I missing something? Let me know!

sahiliitm:
Hi Lukas, 

Sorry for the delayed reply. I was a bit busy. I agree with you. I've changed the comments accordingly.
thanks

lukaszkaiser:
It looks good to me now, great thanks for clarifying and making the code better! There is one small nit: it looks like the indentation of your comments is off. I'm not even sure if that's the case of if it's just the online viewer, but could you take a look and make sure they're indented ok? (I mean: the multi-line argument comments should be a few spaces to the left -- same as other comments in the file.)

Except for this, we should merge this. Some checks seem to still be running...

sahiliitm:
Hi,
Sorry! Indeed, some of the comments were badly formatted. I've fixed that now.
Thanks for pointing it out.

lukaszkaiser:
Looks good to me, thanks! Can we merge this one? Thanks! (Assigned vrv@ just for merging.)

vrv:

PR #3804, opened 2016-08-14
retrain.py, changed graph_util import statement

Travis42:
First attempt at a Pull Request, please be gentle:

I tried the TensorFlow For Poets tutorial today and followed the instructions, but retrain.py kept crashing because graph_util was being imported from the wrong location.  That's line 80.

After the change, the tutorial went as planned!

I also ran 2to3 on it...which apparently got rid of the from **future_** statements.  Hmm.

The rest is just formatting changes, PEP8 stuff.

### DIFF
--- a/tensorflow/examples/image_retraining/retrain.py
+++ b/tensorflow/examples/image_retraining/retrain.py
@@ -60,9 +60,9 @@ Visualize the summaries with this command:
 tensorboard --logdir /tmp/retrain_logs
 
 """
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
+
+
+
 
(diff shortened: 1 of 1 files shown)


Travis42:
I signed it!

Travis42:
I signed it!

vrv:
Thanks for the attempt -- unfortunately the existing code is correct: graph_util.py was moved into framework a month ago, so most likely you're using an older version of TensorFlow installed against a newer copy of the retrain.py code.  You can either upgrade your existing TF install to 0.10.0.rc0 or check out the copy of retrain.py at the version branch you have installed.

PR #3837, opened 2016-08-16
Concurrent access lock

ylongqi:
sess.run() is called concurrently by multiple threads in word2vec. Set use_locking=True in GradientDescentOptimizer to prohibit concurrent parameter read and update.

### DIFF
--- a/tensorflow/models/embedding/word2vec.py
+++ b/tensorflow/models/embedding/word2vec.py
@@ -281,7 +281,7 @@ class Word2Vec(object):
     lr = opts.learning_rate * tf.maximum(
         0.0001, 1.0 - tf.cast(self._words, tf.float32) / words_to_train)
     self._lr = lr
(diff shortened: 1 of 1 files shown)


yaroslavvb:
That may be by design (locking makes things slower)

ylongqi:
Yes, I agree that lock will make things slower. But the program without lock sometimes has behavior that is unrepeatable. IMHO, the lock is not needed when the sess.run is only called inside a single thread but for multithread program, it's better to add it.

yaroslavvb:
Typical application of TensorFlow is non-deterministic even in single-threaded .run because of non-deterministic ops such as reduce_sum on GPU. But I've seen cases when locking was justified because it made training faster (ie, each iteration was slower, but bigger cost decrease)

vrv:
We'll keep this as is, but you are free to add locking as you see fit into your own code.

PR #3861, opened 2016-08-16
enable python setup.py test using nose

Mistobaan:
when developing for the python package might be useful to use standard python testing 

`python setup.py test`


vrv:
From nose's website: "New projects should consider using Nose2, py.test, or just plain unittest/unittest2."

I'd rather keep using unittest for now, unless there's a clear reason for using something else.

vrv:
(Closing for now, feel free to comment -- after reading a bunch of the documentation, it seems like unittest + bazel test accomplishes a lot of what nose/nose2 does).

PR #3876, opened 2016-08-17
Tensorboard log plot option

bdilday:
An option to switch between modified-log and linear y-axis scales for event charts in tensorboard. 

The iron-icons are not being displayed properly, but I believe that issue is unrelated to this pull request.

### DIFF
--- a/tensorflow/tensorboard/components/tf-dashboard-common/dashboard-style.html
+++ b/tensorflow/tensorboard/components/tf-dashboard-common/dashboard-style.html
@@ -36,7 +36,7 @@
       }
       .card .card-bottom-row {
         position: absolute;
-        left: 50px;
+        left: 75px;
         bottom: 0;
         padding-right: 10px;
       }
@@ -49,6 +49,14 @@
       .expand-button {
         position: absolute;
         left: 0px;
+        bottom: 0px;
+        color: #2196F3;
+        display: block;
+      }
+
+      .log-option-button {
+        position: absolute;
+        left: 25px;
         bottom: 0px;
         color: #2196F3;
         display: block;
--- a/tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
+++ b/tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
@@ -232,17 +232,19 @@ The #center div contains tf-line-charts embedded inside tf-collapsable-panes.
           chartScaffold.chart().redraw();
         }
       },
+
       toggleYScale: function(e) {
         var currentTarget = Polymer.dom(e.currentTarget);
 
         var b = currentTarget.parentNode.querySelector('.log-option-button');
         var c = currentTarget.parentNode.querySelector('vz-line-chart');
         if (c !== null) {
-          c.yScale = c.yScale === 'log' ? 'linear' : 'log';
-          b.icon = c.yScale === 'log' ? 'line-weight' : 'reorder';
+          c.yScaleType = c.yScaleType === 'log' ? 'linear' : 'log';
+          b.icon = c.yScaleType === 'log' ? 'line-weight' : 'reorder';
           c.redraw();
         }
       },
+
       validRuns: function(tag, runsChange, run2tagChange) {
         var _this = this;
         var result = this.selectedRuns.filter(function(r) {
--- a/tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
+++ b/tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
@@ -128,6 +128,13 @@ The #center div contains tf-line-charts embedded inside tf-collapsable-panes.
                       icon="fullscreen"
                       on-tap="toggleSelected"
                     ></paper-icon-button>
+                      <paper-icon-button
+                              class="log-option-button"
+                              icon="line-weight"
+                              on-tap="toggleYScale"
+                              title="Toggle y-axis scale (log, linear)"
+                              ></paper-icon-button>
+
                   </div>
                   <template is="dom-if" if="[[_showDownloadLinks]]">
                     <div class="card-bottom-row">
@@ -223,6 +230,17 @@ The #center div contains tf-line-charts embedded inside tf-collapsable-panes.
         var chartScaffold = currentTarget.previousElementSibling;
         if (chartScaffold) {
           chartScaffold.chart().redraw();
+        }
+      },
+      toggleYScale: function(e) {
+        var currentTarget = Polymer.dom(e.currentTarget);
+
+        var b = currentTarget.parentNode.querySelector('.log-option-button');
+        var c = currentTarget.parentNode.querySelector('vz-line-chart');
+        if (c !== null) {
+          c.yScale = c.yScale === 'log' ? 'linear' : 'log';
+          b.icon = c.yScale === 'log' ? 'line-weight' : 'reorder';
+          c.redraw();
         }
       },
       validRuns: function(tag, runsChange, run2tagChange) {
--- a/tensorflow/tensorboard/components/vz-line-chart/vz-line-chart.html
+++ b/tensorflow/tensorboard/components/vz-line-chart/vz-line-chart.html
@@ -173,6 +173,16 @@ such as different X scales (linear and temporal), tooltips and smoothing.
         },
 
         /**
+         * The scale for the y-axis. Allows:
+         * - "linear" - linear scale (Plottable.Scales.Linear)
+         * - "log" - modified-log scale (Plottable.Scales.ModifiedLog)
+         */
+        yScaleType: {
+          type: String,
+          value: 'linear'
+        },
+
+        /**
          * Change how the tooltip is sorted. Allows:
          * - "name" - Sort the tooltip by series name.
          * - "ascending" - Sort the tooltip by ascending value.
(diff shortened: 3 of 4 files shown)


bdilday:
I signed it!

poindextrose:
The CLA was signed under the company name Gravity Jack, Inc. The commit messages have been updated to an email that is in the Google group on the CLA.

tensorflow/tensorboard/components/vz-line-chart/vz-line-chart.ts
```diff
@@ -71,7 +71,7 @@ module VZ {
       this.xScale = xComponents.scale;
       this.xAxis = xComponents.axis;
       this.xAxis.margin(0).tickLabelPadding(3);
-      this.yScale = new Plottable.Scales.Linear();
+      this.yScale = yScale === 'log' ? new Plottable.Scales.ModifiedLog() : new Plottable.Scales.Linear();
```
danmane: Since "log" and "linear" are the accepted options, I think it would be better to do:

function getYScaleFromType(yScaleType: string): Plottable.QuantitativeScale {
  if (yScaleType === "log") {
    return new Plottable.Scales.ModifiedLog();
  } else if (yScaleType === "linear") {
    return new Plottable.Scales.Linear();
  } else {
    throw new Error("Unrecognized yScale type " + yScaleType);
  }

tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
```diff
@@ -110,6 +110,12 @@
                       icon="fullscreen"
                       on-tap="toggleSelected"
                     ></paper-icon-button>
+                      <paper-icon-button
+                              class="log-option-button"
+                              icon="line_weight"
```
danmane: this should be "line-weight"; that's why it isn't displaying.

I think the icon should reflect the current state of the UI, not the state that can be toggled to. So if we go with this UI approach, it should start with "re-order" (because it is a linear scale) and then swap to "line-weight" when it is a log scale.

tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
```diff
@@ -194,6 +201,17 @@
           chartScaffold.chart().redraw();
         }
       },
+
+      toggleYScale: function(e) {
+        var b = this.parentNode.querySelector('.log-option-button');
+        var c = this.parentNode.querySelector('vz-line-chart');
+        if (c !== null) {
+          c.yScale = c.yScale === 'log' ? 'linear' : 'log';
+          b.icon = c.yScale === 'log' ? 'reorder' : 'line_weight';
```
danmane: should be "line-weight"

tensorflow/tensorboard/components/vz-line-chart/vz-line-chart.html
```diff
@@ -154,6 +154,11 @@
           value: 'step'
         },
 
+        yScale: {
+          type: String,
+          value: 'linear'
+        },
```
danmane: property needs documentation in a comment that specifies allowable values, etc.

tensorflow/tensorboard/dist/tf-tensorboard.html
```diff
@@ -3812,6 +3830,17 @@ <h3 id="tooltip-help" class="tooltip-container">
           chartScaffold.chart().redraw();
         }
       },
+
+      toggleYScale: function(e) {
+        var b = this.parentNode.querySelector('.log-option-button');
+        var c = this.parentNode.querySelector('vz-line-chart');
+        if (c !== null) {
```
danmane: As currently written, clicking any log/linear button always modifies the first chart. Here is a fix:
```
      toggleYScale: function(e) {
        var currentTarget = Polymer.dom(e.currentTarget);

        var b = currentTarget.parentNode.querySelector('.log-option-button');
        var c = currentTarget.parentNode.querySelector('vz-line-chart');
        if (c !== null) {
          c.yScale = c.yScale === 'log' ? 'linear' : 'log';
          b.icon = c.yScale === 'log' ? 'line-weight' : 'reorder';
          c.redraw();
        }
      },
```

tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
```diff
@@ -194,6 +201,17 @@
           chartScaffold.chart().redraw();
         }
       },
+
+      toggleYScale: function(e) {
+        var b = this.parentNode.querySelector('.log-option-button');
+        var c = this.parentNode.querySelector('vz-line-chart');
```
danmane: As currently written, clicking any log/linear button always modifies the first chart. Here is a fix:
```
      toggleYScale: function(e) {
        var currentTarget = Polymer.dom(e.currentTarget);

        var b = currentTarget.parentNode.querySelector('.log-option-button');
        var c = currentTarget.parentNode.querySelector('vz-line-chart');
        if (c !== null) {
          c.yScale = c.yScale === 'log' ? 'linear' : 'log';
          b.icon = c.yScale === 'log' ? 'line-weight' : 'reorder';
          c.redraw();
        }
      },
```

tensorflow/tensorboard/components/tf-dashboard-common/dashboard-style.html
```diff
@@ -54,6 +54,14 @@
         display: block;
       }
 
+      .log-option-button {
```
danmane: this CSS should actually be in tf-event-dashboard, not tf-dashboard-common, since it is only for the event dashboard.

jart:
This is amazing. Thank you for taking the time to send us this pull request.

Would you happen to have screenshots? Or better yet, could you set up a running demo server on a VM somewhere?

tensorflow/tensorboard/components/vz-line-chart/vz-line-chart.ts
```diff
@@ -45,7 +45,7 @@ module VZ {
     private smoothingEnabled: Boolean;
 
     constructor(
-        xType: string, colorScale: Plottable.Scales.Color,
+        xType: string, yScale: string, colorScale: Plottable.Scales.Color,
```
danmane: "yScale" as a variable name makes me expect a scale. Please rename to "yScaleType" or similar.

tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
```diff
@@ -110,6 +110,12 @@
                       icon="fullscreen"
                       on-tap="toggleSelected"
                     ></paper-icon-button>
+                      <paper-icon-button
```
danmane: To make this slightly more self-documenting, please add a title text like "Toggle between Linear and Logarithmic y scale"

Or you can make a title like "swap to log scale" and then change the title when the button is clicked.

tensorflow/tensorboard/components/tf-dashboard-common/dashboard-style.html
RenatoUtsch: tf-event-dashboard doesn't currently have CSS rules of its own, all css rules are being dumped in dashboard-style.

danmane:
Also, can you revert the dist/tf-tensorboard.html file in this pull request? It is going to generate a lot of conflicts, and we have tooling for regenerating it when we sync the GitHub repository.

tensorflow/tensorboard/components/tf-event-dashboard/tf-event-dashboard.html
RenatoUtsch: I feel like it is confusing both ways. Maybe changing the icon background to show that the option is 'selected'  is better than changing between different icons. This way, it would become obvious if the log scale is enabled or not.

tensorflow/tensorboard/components/tf-dashboard-common/dashboard-style.html
danmane: Maybe we should just put a style block in tf-event-dashboard?

Renato you own refactoring the dashboards so I'll let you decide what, if anything, should happen here :)

danmane:
@jart To make things easier for bdilday here are some screenshots: 
image

(The bottom chart is in log mode).

The lack of a top line on the log plot is a bit weird, but really a manifestation of a different issue.

tensorflow/tensorboard/components/tf-dashboard-common/dashboard-style.html
RenatoUtsch: I think keeping it on dashboard-style for now is ok, I'll have to move a bunch of css rules in the future anyway.

tensorflow/tensorboard/components/tf-dashboard-common/dashboard-style.html
```diff
@@ -54,6 +54,14 @@
...
+          position: absolute;
+          margin-left: 26px;
```
RenatoUtsch: using left instead of margin-left here would be preferable. Did you check if the new button and the div that appears when 'show download links' is enabled overlap?

bdilday: the log-option button did overlap the "download data" div. I moved the `.card-bottom-row` element over by 25px to get rid of the overlap.

rmlarsen:

rmlarsen:

rmlarsen:
@bdilday please rebase your PR and resolve conflicts.

danmane:

poindextrose:
@bdilday It looks like you wiped out important changes in the rebase you did. Check the files changed on this PR.

danmane:

rmlarsen:
The failing test is flaky. Merging.

PR #3877, opened 2016-08-17
Rename Tensor "summary_op" to "summary" in the 101 tutorial

tijmentieleman:
"summary_op" is a Tensor, so it's incorrectly labeled as an op. This change renames it to "summary", and makes the corresponding changes in the explanation.


josh11b:
Looks good to me.

danmane:
LGTM too :)

josh11b:

tijmentieleman:
One of the test errors is "No space left on device". I don't feel that I understand the testing system well enough to comment intelligently, though I'm speculating that the error is not in the code change because it's just a variable rename. Who could take a look and diagnose the problem?

vrv:
known infra issues, merged since all other tests passed

PR #3880, opened 2016-08-17
Change "var1 == var2" to "var1 is var2"

tijmentieleman:
This is a detail and might be just a matter of taste. My reason for proposing this change is that the Python "is" operator stands for a stronger type of equality than the "==" operator, so perhaps it drives home the equality point more strongly. However, again, this could just be a matter of taste; I will certainly not be offended if this proposal is rejected.


lukaszkaiser:
Looks good to me, thanks!

PR #3883, opened 2016-08-17
Fixing comment, wrong data type in argument description.

dcastro9:
Usage can be seen in lines 445->446, the file is read in and passed directly in, it is not a Numpy array, but rather just a string of raw JPEG data passed into the DecodeJpeg layer of an arbitrary network (by default Inception v3).


shlens:
LGTM

danmane:
@tensorflow-jenkins PTAL

petewarden:
I believe this correction is right, LGTM, thanks!

danmane:
Since this only changes a docstring, I am going to live dangerously and merge it now.

PR #3914, opened 2016-08-19
Fix typo in documentation of `gate_gradients`

gibiansky:
The documentation said `gate_gradient` in several places; however, the
option is named `gate_gradients`. This PR fixes this in all the
places that `gate_gradient` occurs.


danmane:
Thanks!

PR #3927, opened 2016-08-19
Docs(Queue): Fix a PriorityQueue API missing bug

DjangoPeng:
As we all know there are 4 subclass of `class QueueBase`, but only 3 subclass(FIFOQueue, PaddingFIFOQueue, RandomShuffleQueue) descriptions in the API Docs [here](https://www.tensorflow.org/versions/r0.10/api_docs/python/io_ops.html#QueueBase)

So I add the description of  `class PriorityQueue()` which are the comments in ../python/ops/data_flow_ops.py.
Start line 714 to 759 and you can check [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/data_flow_ops.py#L714)

### DIFF
--- a/tensorflow/g3doc/api_docs/python/io_ops.md
+++ b/tensorflow/g3doc/api_docs/python/io_ops.md
@@ -2079,6 +2079,57 @@ queue has been closed.
 
 
 
+- - -
+
+### `class tf.PriorityQueue` {#PriorityQueue}
+
+A queue implementation that dequeues elements in prioritized order.
+
+See [`tf.QueueBase`](#QueueBase) for a description of the methods on
+this class.
+
+- - -
+
+
+#### `tf.PriorityQueue.__init__(capacity, types, shapes=None, names=None, shared_name=None, name='priority_queue')` {#PriorityQueue.__init__}
+
+Creates a queue that dequeues elements in a first-in first-out order.
+
+A `PriorityQueue` has bounded capacity; supports multiple concurrent
+producers and consumers; and provides exactly-once delivery.
+
+A `PriorityQueue` holds a list of up to `capacity` elements. Each
+element is a fixed-length tuple of tensors whose dtypes are
+described by `types`, and whose shapes are optionally described
+by the `shapes` argument.
+
+If the `shapes` argument is specified, each component of a queue
+element must have the respective fixed shape. If it is
+unspecified, different queue elements may have different shapes,
+but the use of `dequeue_many` is disallowed.
+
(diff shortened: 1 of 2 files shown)


danmane:
Thanks for the pull request! Unless I'm much mistaken, these files should be autogenerated from the source in data_flow_ops. @josh11b can you comment on what the right fix is?

josh11b:
The source for the documentation of tf.PriorityQueue lives in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/data_flow_ops.py#L715

tensorflow/g3doc/api_docs/python/io_ops.md is automatically generated.  Any changes to it will be overwritten the next time the documentation generator runs.  In fact, its very first line is:


The fact that PriorityQueue is not being generated is an issue, which we can hopefully fix by adding PriorityQueue to the list here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/io_ops.py#L82
(And hopefully then the doc generator will generate docs from the data_flow_ops.py file -- see https://www.tensorflow.org/versions/r0.10/how_tos/documentation/index.html for instructions on how to run the doc generator yourself. TL/DR: tools/docs/gen_docs.sh)

If, once the documentation is being generated, there are still improvements to be made, those changes should be made to data_flow_ops.py.

DjangoPeng:
@josh11b So it is a missing mistake of the list, and when could you run the doc generator script? 
And do you mean I don't need to push this commits this time?

rmlarsen:
@DjangoPeng @josh11b I think Josh's point was that if you add the correct entry to

 https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/io_ops.py#L82

then the documentation will get automatically generated (visibility pending our push). Please go ahead and make that change as well as any documentation updates to data_flow_ops.py in this PR.

DjangoPeng:
@rmlarsen So do I need to create a new pull request or not? I don't know if it's right to add the commit in this request.

rmlarsen:
@DjangoPeng  Can you please revert the changes from io_ops.md, since this is an autogenerated file, and instead update the documentation in 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/data_flow_ops.py#L715

if needed.

DjangoPeng:
@rmlarsen I have reverted the change from `io_ops.md`, and it's unnecessary to update the `data__flow_ops.py` this time. Plz check.

rmlarsen:

PR #3928, opened 2016-08-19
Fix broken link for gif_archive

caisq:


caisq:

PR #3955, opened 2016-08-22
Change app.py() to pass through unparsed flags.

vrv:
This is so that a client that is using our flags library can pass their own flags as well.

### DIFF
--- a/tensorflow/python/BUILD
+++ b/tensorflow/python/BUILD
@@ -71,6 +71,26 @@ tf_py_test(
     ],
 )
 
+tf_py_test(
+    name = "flags_test",
+    size = "small",
+    srcs = ["platform/flags_test.py"],
+    additional_deps = [
+        ":platform",
+        ":platform_test",
+    ],
+)
+
+tf_py_test(
+    name = "app_test",
+    size = "small",
+    srcs = ["platform/app_test.py"],
+    additional_deps = [
+        ":platform",
+        ":platform_test",
+    ],
+)
+
 cc_library(
     name = "numpy_lib",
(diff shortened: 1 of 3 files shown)


tensorflow/python/platform/app_test.py
```diff
@@ -0,0 +1,45 @@
...
+import sys
+
+from tensorflow.python.platform import app
+from tensorflow.python.platform import flags
+
+FLAGS = flags.FLAGS
+flags.DEFINE_boolean('myflag', False, '')
+
+def main(argv):
+  print(argv)
```
craigcitro: debugging?

tensorflow/python/BUILD
```diff
@@ -71,6 +71,31 @@ tf_py_test(
     ],
 )
 
+tf_py_test(
+    name = "flags_test",
+    size = "small",
+    srcs = ["platform/flags_test.py"],
```
craigcitro: wait, is there a flags_test that just wasn't getting run? 😞

craigcitro:
LGTM, just one or two little questions.

We could also tweak `app_test.py` to use the `unittest` module, but I actually feel like it's a little nicer to see that it's getting actual command-line args routed through correctly.

rmlarsen:

tensorflow/python/platform/app_test.py
vrv: Removed

tensorflow/python/BUILD
vrv: Some refactoring must have caused it to stop running :(

PR #3966, opened 2016-08-22
Add cuda_configure repository rule to autodetect cuda.

### LINKED ISSUE #2873: Perform CUDA autoconfiguration in Skylark

gunan:
This change reimplements the CUDA autoconfiguration mechanism in Skylark,
providing a `cuda_configure` workspace rule. We keep the same user interface,
the ./configure script, but rather than modifying source files within the
source tree, `cuda_configure` generates a `@local_config_cuda` workspace
containing:
- Symlinks to the CUDA headers and libraries
- BUILD files generated with the correct CUDA and cuDNN versions
- CROSSTOOL config with CUDA include dirs populated
- crosstool_wrapper_driver_is_not_gcc wrapper script with compiler paths and
  CUDA compute capabilities set.
- cuda_config.h header file with CUDA versions and compute capabilities set,
  which can be `#include`d by source files.

This change also makes the following fixes to `Dockerfile.gpu`:
- Change the `CUDNN_INSTALL_PATH` to point to `/usr/lib/x86_64-linux-gnu`
  rather than `/usr/local/cuda` since NVIDIA's image installs `libcudnn.so`
  under `/usr/lib/x86_64-linux-gnu`.
- Add env variable to set the minimum compute capability to 3.0.

Fixes #2873


tensorflow/tensorflow.bzl
```diff
@@ -316,29 +316,29 @@ def _cuda_copts():
     common_cuda_opts = ["-x", "cuda", "-DGOOGLE_CUDA=1"]
     return select({
         "//conditions:default": [],
-        "//third_party/gpus/cuda:using_nvcc": (
+        "@local_config_cuda//cuda:using_nvcc": (
             common_cuda_opts +
             [
                 "-nvcc_options=relaxed-constexpr",
```
Mistobaan: it would be nice if you could detect the version of `nvcc` and if is greater or equal to 7.5 then use `--expt-relaxed-constexpr`. unless we make 7.5 min requirement and just use that flag.  This will remove an annoying warning and all the tons of lines created inside logs.

gunan: @davidzchen 
This is just a cherrypick for this change into r0.10 release, but we will note this request for changes into master and following releases.

davidzchen: I have opened #3980 for improvements to the CUDA autoconf. If there are any feature requests for `cuda_configure`, feel free to add them there.

Issue #3945, opened 2016-08-21
Feature request: extend tf.select to broadcast a scalar condition


NeilGirdhar:
Consider allowing tensorflow.select to accept scalar conditions.  I.e., `tf.select(condition, t, e, name=None)` with `condition` a scalar, and `t`, `e` having the same shape.

vrv:
I think that is perhaps better accomplished via tf.cond?  https://www.tensorflow.org/versions/r0.10/api_docs/python/control_flow_ops.html#cond

(The reason it is "better" is that, if implemented carefully, tf.cond doesn't evaluate both branches, whereas tf.select with this feature would always evaluate both branches).  Please comment if you disagree!

NeilGirdhar:
Isn't `cond` done on the CPU?  I mean, even if the condition is a placeholder, cond has to copy the result of `pred` back to the CPU and then call either `fn1` or `fn2` — all on the CPU.  `select` seems better if everything is already on the GPU and needs to stay on the GPU.

vrv:
That's a good point, and select does do some amount of broadcasting already (it can broadcast row-wise when it is a vector), so it seems reasonable to extend that all the way to the scalar case.

This probably isn't super high priority though, but would be a welcome improvement if someone in the community wanted to implement this, so marking this as contributions welcome.  In the mean time, you could probably use tf.tile to broadcast the scalar into a vector of the right length without the same penalty as the full broadcast.

NeilGirdhar:
Thanks for reopening the issule, could you point me to where select is defined in the source?  My search for "def select" didn't find it.

vrv:
You'd have to implement the change in math_ops.py _SelectShape to allow the new shape signature, math_ops.cc's ShapeFn to do the same, and then you'd have to edit cwise_op_select.cc and cwise_op_gpu_select.cu.cc to implement the new feature, and then probably cwise_ops_test.py to validate the change.

NeilGirdhar:
Okay, thanks.  I may work around this for now using `expand_dims` on my condition.

Mistobaan:
Taking a look at this. The additional case when the condition is just a scalar it will return either the full tensor from `t` or `e` based on the single value of `cond`, right?

vrv:
Yup!

albertz:
Can someone comment on the state here? This was reverted? Why?

What broadcasting is supposed to work?

E.g. I get the error:

[...]

PR #3995, opened 2016-08-23
extend tf.select to broadcast a scalar condition #3945

Mistobaan:
This is WIP. The test fails on my machine for the GPU version. Comments are welcome.

### DIFF
--- a/tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
+++ b/tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
@@ -33,6 +33,20 @@ struct SelectFunctor<GPUDevice, T> {
 };
 
 template <typename T>
+struct SelectScalarFunctor<GPUDevice, T> {
+  void operator()(const GPUDevice& d, typename TTypes<T>::Flat out,
+                  typename TTypes<bool>::ConstScalar cond,
+                  typename TTypes<T>::ConstFlat then_flat,
+                  typename TTypes<T>::ConstFlat else_flat) {
+    if (cond()){
+        To32Bit(out).device(d) = To32Bit(then_flat);
+    } else {
+        To32Bit(out).device(d) = To32Bit(else_flat);
+    }
+  }
+};
+
+template <typename T>
 struct BatchSelectFunctor<GPUDevice, T> {
   void operator()(const GPUDevice& d,
                   typename TTypes<T>::Matrix output_flat_outer_dims,
@@ -69,6 +83,7 @@ struct BatchSelectFunctor<GPUDevice, T> {
 
 #define SELECT_FUNCTOR(T)                      \
   template struct SelectFunctor<GPUDevice, T>; \
+  template struct SelectScalarFunctor<GPUDevice, T>; \
   template struct BatchSelectFunctor<GPUDevice, T>;
 
 SELECT_FUNCTOR(Eigen::half);
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -1788,8 +1788,10 @@ def _AddNShape(op):
 def _SelectShape(op):
   """Shape function for SelectOp."""
   # The inputs 'then' and 'else' must have the same shape.
-  # The input 'cond' must either have the same shape as 'then' and
-  # 'else', or be a vector if 'then' and 'else' are at least vectors.
+  # The input 'cond' must either:
+  # 1. be a scalar
+  # 2. have the same shape as 'then' and 'else'
+  # 3. be a vector if 'then' and 'else' are at least vectors
   c_shape = op.inputs[0].get_shape()
   t_shape = op.inputs[1].get_shape()
   e_shape = op.inputs[2].get_shape()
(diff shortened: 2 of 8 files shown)


rmlarsen:

Mistobaan:
from the CI_BUILD Is hard to understand where is the problem. But debugging locally I find it segfaults on the gpu's 'cond()' statement.  I tried to trace down what actually happens but that is dark Eigen magic.

keveman:
I will have to patch this in and try to repro on my machine. Will try to do so later this evening.

tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
```diff
@@ -33,6 +33,20 @@ struct SelectFunctor<GPUDevice, T> {
 };
 
 template <typename T>
+struct SelectScalarFunctor<GPUDevice, T> {
+  void operator()(const GPUDevice& d, typename TTypes<T>::Flat out,
+                  typename TTypes<bool>::ConstScalar cond,
+                  typename TTypes<T>::ConstFlat then_flat,
+                  typename TTypes<T>::ConstFlat else_flat) {
+    if (cond()){
```
vrv: I believe this is evaluating cond() as if it were in CPU memory, but it's likely in GPU memory, so it's segfaulting trying to derefernce GPU memory as CPU memory, I believe.

tensorflow/python/ops/math_ops.py
```diff
@@ -1788,18 +1788,24 @@ def _AddNShape(op):
 def _SelectShape(op):
```
vrv: You also have to update the C++ shape function in the array_ops.cc definition of the Select op

Mistobaan: I am assuming you meant math_ops.cc:810.  The Shape function in C++ seems much more complex than necessary. Are those SetShapeFn tested somewhere in C++ ?

vrv: yes, math_ops.cc.  Should be tested in math_ops_test.cc.  The shape function in C++ should be similar to the python one, it's just more verbose cause it's not python.

tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
Mistobaan: so is on the Eigen side or I am using the wrong API to access the bool value  ?

vrv: I guess I would recommend using the broadcasting approach used in the BatchSelectFunctor.  Or you could write a custom cuda kernel instead of using Eigen, since this one isn't that hard.

Mistobaan: even that kernel has a comment from @ebrevdo that leads to erroneous memory access. Seems something pesky is going on underneath in Eigen.
The second solution to just pass a bool to the kernel. But I wonder if that would require CPU sync with the op to transfer that single boolean. @benoitsteiner Do you have any insight in why a call to a ScalarBool on gpu leads to memory access errors ?

Mistobaan: maybe is because only float{16,32,64} are supported on the GPU ?

danmane:
@Mistobaan What's the status here? Should I poke some Googlers to give you some more feedback, e.g. on SelectScalarFunction thread?

Mistobaan:
@danmane this was so long ago that I have to check. I think the GPU kernel fails.

danmane:

tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
ebrevdo: You can put cond on the host by marking that tensor to be in HostMemory during the GPU kernel registration.

I think for the assignments, you don't need to To32Bit(out)...

just do:

out.device(d) = then_flat;

etc.

ebrevdo:
Can you rebase on head & fix conflicts?

tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
vrv: I think the whole point of this op was to have the 'cond' be on GPU though.  Otherwise you might as well just use tf.cond()

ebrevdo: In this case you want to broadcast the scalar value to a vector and pass
the ref to the standard select functor.  I'd like to see that there's a
performance benefit to not copying a single scalar from GPU to CPU though,
to justify all this.

martinwicke:
@Mistobaan the simplest thing to do is the HostMemory annotation, but @ebrevdo's solution should work as well. The copying should be justifiable if the other inputs don't have to be moved off of the GPU as a result.

Can you make those changes and rebase?

vrv:
As mentioned above, if the cond scalar is forced on host memory, then there's little benefit to using this over tf.cond(), which is exactly the same thing.

I don't think this PR makes sense if the 'cond' input is host memory, it's just adding more complexity for no benefit.

tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
vrv: Or you could write a custom cuda kernel instead of using Eigen.  That was my suggestion, since there's no complex selecting going on here, it's just 'A or B'.

Mistobaan:
I will go with the custom cuda kernel solution

drpngx:
Thanks Mistobaan, closing since you're going to go another path. Please re-open if I got it wrong.

Mistobaan:
@drpngx can yo reopen this? I just commit the changes that @benoitsteiner suggested and seem to work on my machine.

drpngx:
github says it can't be re-opened because it was force-pushed or re-created. Did you do that?

Mistobaan:
@drpngx Yes I did rebase and force push to my branch. I will open a new PR then

Issue #3996, opened 2016-08-23
Fix CMake build due to cuda_configure changes


davidzchen:
The new changes from #3269 broke the CMake build due to the header include path changes and the new `cuda_config.h` file generated by `cuda_configure`.

mrry:
This is now fixed, thanks to @guschmue's great work on the Windows GPU build. There are still some wrinkles regarding CMake-on-Linux (see #5465), but I'm going to close this issue for now.

davidzchen:
Thanks for fixing this, @guschmue!

PR #3997, opened 2016-08-23
Disable CMake build of tf_stream_executor

mrry:
Temporary mitigation for #3996.


rmlarsen:
Thanks @mrry I'll merge once this passes.

mrry:
Running tests again - needed to fix up a dependency in debug_ops.h.

mrry:
Urgh, that did not go as planned. Looks like the easiest solution will be to remove debug_ops from the CMake build. I'll send another PR for that.

Issue #3762, opened 2016-08-12
PS is bound in GPU:0 by default, we can't change!


DjangoPeng:
I modified the seq2seq , mnist_replica and ptb model into distributed training model. But all of them have the same problem, that is when I start the server and `server.join()` the ps. The ps was bound in GPU:0 by default, and I can't change the setting. I follow the tutorial of distributed tensorflow [here](https://www.tensorflow.org/versions/r0.10/how_tos/distributed/index.html#putting-it-all-together-example-trainer-program). 
This is part of my translate.py :
```
def train():

  # set distributed configs
  ps_hosts = ["9.91.9.130:2222"]
  worker_hosts = ["9.91.9.130:2223", "9.91.9.130:2224"]
  #worker_hosts = ["9.91.9.130:2223"]

  cluster = tf.train.ClusterSpec({"ps":ps_hosts, "worker":worker_hosts})
  server = tf.train.Server(cluster,
                            job_name=FLAGS.job_name,
                            task_index=FLAGS.task_index)
  if FLAGS.job_name == "ps":
        server.join()
  elif FLAGS.job_name == "worker":
      # Worker server 
      is_chief = (FLAGS.task_index == 0)      
      gpu_num = FLAGS.task_index + 1
      #with tf.Graph().as_default():
      with tf.device(tf.train.replica_device_setter(cluster=cluster,
          worker_device="/job:worker/task:%d/gpu:%d" % (FLAGS.task_index, gpu_num))):
      #with tf.device("/gpu:%d" % FLAGS.task_index):
          """Train a en->fr translation model using WMT data."""
```
This is the GPU info:
```
Fri Aug 12 09:17:07 2016
+------------------------------------------------------+
| NVIDIA-SMI 352.39     Driver Version: 352.39         |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla K80           On   | 0000:06:00.0     Off |                    0 |
| N/A   67C    P0    63W / 149W |  11099MiB / 11519MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Tesla K80           On   | 0000:07:00.0     Off |                    0 |
[...]

PR #4008, opened 2016-08-24
Change device_setter.py for choose ps device

HW-ZZ:
Add arguments "ps_device_name" for replica_device_setter(), in order to choose ps decive .
Reference the issue[#3762](https://github.com/tensorflow/tensorflow/issues/3762)

### DIFF
--- a/tensorflow/python/training/device_setter_test.py
+++ b/tensorflow/python/training/device_setter_test.py
@@ -77,6 +77,23 @@ class DeviceSetterTest(tf.test.TestCase):
       self.assertDeviceEqual("/job:moon/task:1", w.device)
       self.assertDeviceEqual("/job:moon/task:1", w.initializer.device)
       self.assertDeviceEqual("/job:sun", a.device)
+  
+  def testPS2TasksWithCPUConstraint(self):
+     cluster_spec = tf.train.ClusterSpec({
+        "sun": ["sun0:2222", "sun1:2222", "sun2:2222"],
+        "moon": ["moon0:2222", "moon1:2222"]})
+
+    with tf.device(tf.train.replica_device_setter(
+        ps_device="/job:moon/cpu:0", worker_device="/job:sun",
+        cluster=cluster_spec.as_cluster_def())):
+      v = tf.Variable([1, 2])
+      w = tf.Variable([2, 1])
+      a = v + w
+      self.assertDeviceEqual("/job:moon/task:0/cpu:0", v.device)
+      self.assertDeviceEqual("/job:moon/task:0/cpu:0", v.initializer.device)
+      self.assertDeviceEqual("/job:moon/task:1/cpu:0", w.device)
+      self.assertDeviceEqual("/job:moon/task:1/cpu:0", w.initializer.device)
+      self.assertDeviceEqual("/job:sun", a.device)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


mrry:
I'm not sure I see the point of this change. Can't you pass (e.g.) `"/job:ps/cpu:0"` to the existing `ps_device` argument if (as in #3762) you want to put all the variables on the CPU by default?

HW-ZZ:
This is the code block in def replica_device_setter()
```
 ps_job_name = ps_device.lstrip("/job:")
 if ps_job_name not in cluster_spec or cluster_spec[ps_job_name] is None:
      return None
```
if the argument `ps_device="/job:ps/cpu:0"` it will return None

mrry:
OK, that `lstrip()` is completely wrong. Thanks for pointing it out.

More properly, it should do:

```python
ps_job_name = pydev.DeviceSpec.from_string(ps_device).job
```

Do you want to try making that change in this PR?

HW-ZZ:
@mrry  Yes, and thanks for your reply

HW-ZZ:
@mrry I have changed it

mrry:
Great, thanks! Can you also add a test that covers this case to `device_setter_test.py`?

I imagine you could copy `testPS2TasksWithDevice()` and make it `testPS2TasksWithCPUConstraint()`. The code would look something like:

```python
  def testPS2TasksWithCPUConstraint(self):
    cluster_spec = tf.train.ClusterSpec({
        "sun": ["sun0:2222", "sun1:2222", "sun2:2222"],
        "moon": ["moon0:2222", "moon1:2222"]})

    with tf.device(tf.train.replica_device_setter(
        ps_device="/job:moon/cpu:0", worker_device="/job:sun",
        cluster=cluster_spec.as_cluster_def())):
      v = tf.Variable([1, 2])
      w = tf.Variable([2, 1])
      a = v + w
      self.assertDeviceEqual("/job:moon/task:0/cpu:0", v.device)
      self.assertDeviceEqual("/job:moon/task:0/cpu:0", v.initializer.device)
      self.assertDeviceEqual("/job:moon/task:1/cpu:0", w.device)
      self.assertDeviceEqual("/job:moon/task:1/cpu:0", w.initializer.device)
      self.assertDeviceEqual("/job:sun", a.device)
```

HW-ZZ:
@mrry  I have added testPS2TasksWithCPUConstraint() device_setter_test.py, any other test should be add?

tensorflow/python/training/device_setter_test.py
```diff
@@ -77,6 +77,23 @@ def testPS2TasksWithDevice(self):
       self.assertDeviceEqual("/job:moon/task:1", w.device)
       self.assertDeviceEqual("/job:moon/task:1", w.initializer.device)
       self.assertDeviceEqual("/job:sun", a.device)
+  
+  def testPS2TasksWithCPUConstraint(self):
+     cluster_spec = tf.train.ClusterSpec({
+        "sun": ["sun0:2222", "sun1:2222", "sun2:2222"],
+        "moon": ["moon0:2222", "moon1:2222"]})
```
mrry: It looks like the indentation is slightly inconsistent here: to satisfy our linter (which follows [these style rules](https://google.github.io/styleguide/pyguide.html)), the code should be indented two spaces at each block, and statement continuations (such as L83 and L84) should be indented four spaces.

Can you please fix that up? Then we should be good to go!

mrry:

mrry:
Looks good to me! Thanks for making this change.

rmlarsen:
Thanks for the contribution!

PR #4017, opened 2016-08-24
Add PSGLD optimizer

yzqcjdtc:
I add an algorithm, Preconditioned Stochastic Gradient Langevin Dynamics (PSGLD), for training optimization, which
- derives from Stochastic Gradient Langevin Dynamics (SGLD) and SGD,
- requires only the gradient on mini-batches of data,
- incorperates adaptive preconditioners from RMSProp.

And it effectively accelerates the convergence speed compared with the existed algorithms, while basically maintaining the accuracy of prediction.


rmlarsen:
Thank you for your contribution. This is really nice work, but we think that for now it may be better for this to live outside of the official repo; we can add it later if it proves to be useful to the wider community.

yzqcjdtc:
Thankd for your comments! @rmlarsen

PR #4028, opened 2016-08-25
Branch 131232000

rmlarsen:


tensorflow/core/framework/shape_inference.h
```diff
@@ -28,6 +28,7 @@ limitations under the License.
 namespace tensorflow {
 namespace shape_inference {
 
+struct DimensionOrConstant;
```
Mistobaan: shouldn't this be `class` ? it will generate a warning otherwise.

caisq: I'm merging the PR first. Will address this later soon.

PR #4040, opened 2016-08-25
fix small typo in contrib.metrics.confusion_matrix docs, and use less…

ichuang:
… confusing variable


ichuang:
CLA signed

rmlarsen:
What do you have against the letter l?

ichuang:
When rendered by github, in this context, the letter I looks like a vertical bar:

image

and this makes it appear unfortunately like incomplete [density matrix notation](https://en.wikipedia.org/wiki/Density_matrix).

rmlarsen:
OK, I don't object to k either.

PR #4041, opened 2016-08-25
Cherry-pick into r0.10 cuda configure fix and gif.BUILD missing headers

vrv:


vrv:
Doh, I squashed them instead of cherry-picking them individually.  Oh well :/

PR #4046, opened 2016-08-25
Fuse resize and mirror padding ops into convolutions

petewarden:
Spatial transformations like padding and bilinear resizing can be merged into the im2col stage of conv2d. This reduces the memory usage considerably (from 338MB to 224MB) and latency (by 15%) on some models, and helps us avoid OOM crashes on iOS. This PR has all the changes needed to fuse these particular ops, including the kernels themselves and integration into the optimize_for_inference script.


petewarden:
Chad has agreed to review this.

cwhipkey:
Should conv_ops_fused go in contrib instead?  In core, there is a higher commitment to keep this approach around longer.

tensorflow/contrib/makefile/tf_proto_files.txt
```diff
@@ -19,7 +19,6 @@ tensorflow/core/framework/tensor.proto
 tensorflow/core/framework/summary.proto
 tensorflow/core/framework/step_stats.proto
 tensorflow/core/framework/op_def.proto
-tensorflow/core/framework/node_def.proto
```
cwhipkey: is this a bad merge?

tensorflow/core/kernels/conv_ops.h
```diff
@@ -38,6 +39,14 @@ class LaunchConv2DOp {
               TensorFormat data_format);
 };
 
+// Used to keep track of persistent memory buffers used within the op.
+template <class T, size_t size>
+struct Im2ColBufferResource : public ResourceBase {
+  mutex mu;
+  T data[size];
```
cwhipkey: comment what this protects?  (it protects the full use of data buffers?

tensorflow/core/kernels/conv_ops_test.cc
```diff
@@ -0,0 +1,240 @@
+/* Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
cwhipkey: 2016

tensorflow/core/kernels/gemm_functors.h
```diff
@@ -0,0 +1,105 @@
+/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
cwhipkey: these were verbatim copies from the other file?

tensorflow/core/kernels/image_resizer_state.h
```diff
@@ -27,13 +27,13 @@ limitations under the License.
 #include <algorithm>
 #include <array>
 
-#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
```
cwhipkey: for some reason, existing code puts this first in a lot of places.  do you know why?

(same on other files, if this ends up getting moved back).

tensorflow/core/ops/nn_ops.cc
```diff
@@ -425,6 +426,31 @@ data_format: Specify the data format of the input and output data. With the
...
+    .Input("input: T")
+    .Input("size: int32")
+    .Input("paddings: int32")
+    .Input("filter: T")
+    .Output("output: T")
+    .Attr("T: {half, float, double}")
+    .Attr("resize_align_corners: bool = false")
+    .Attr(GetMirrorPadModeAttrString())
+    .Attr("strides: list(int)")
+    .Attr(GetPaddingAttrString())
```
cwhipkey: we'll need to add a C++ shape function at some point (this change or soon after).

tensorflow/python/ops/nn_ops.py
```diff
@@ -794,6 +794,71 @@ def _BatchNormGradShape(op):
 ops.RegisterShape("MaxPool")(common_shapes.max_pool_shape)
 
 
+@ops.RegisterShape("FusedResizeAndPadConv2D")
```
cwhipkey: maybe not worth it here, since the cpp shape function would be the final place - but is there overlap between this and the other conv ops?

tensorflow/python/tools/optimize_for_inference_lib.py
```diff
@@ -303,3 +307,92 @@ def fold_batch_norms(input_graph_def):
...
+  Raises:
+    ValueError: If the graph is badly formed with duplicate node names.
+  """
+
+  input_node_map = {}
+  for node in input_graph_def.node:
+    if node.name not in input_node_map.keys():
+      input_node_map[node.name] = node
+    else:
+      raise ValueError("Duplicate node names detected.")
```
cwhipkey: include    node.name  in the exception message?

tensorflow/python/tools/optimize_for_inference_lib.py
```diff
@@ -303,3 +307,92 @@ def fold_batch_norms(input_graph_def):
...
+    nodes_to_skip[resize_op.name] = True
+
+    fused_conv_op = tf.NodeDef()
+    fused_conv_op.op = "FusedResizeAndPadConv2D"
+    fused_conv_op.name = conv_op.name
+    if mirror_pad_op:
+      mirror_paddings_name = mirror_pad_op.input[1]
+      mirror_paddings_mode = mirror_pad_op.attr["mode"]
+    else:
+      paddings_op = tf.NodeDef()
```
cwhipkey: can you add a 1-line comment for why this is done?

tensorflow/python/tools/optimize_for_inference_lib.py
```diff
@@ -303,3 +307,92 @@ def fold_batch_norms(input_graph_def):
...
+      mirror_paddings_mode = tf.AttrValue(s="REFLECT")
+    fused_conv_op.input.extend([resize_op.input[0], resize_op.input[1],
+                                mirror_paddings_name, conv_op.input[1]])
+    fused_conv_op.attr["T"].CopyFrom(conv_op.attr["T"])
+    fused_conv_op.attr["resize_align_corners"].CopyFrom(
+        resize_op.attr["align_corners"])
+    fused_conv_op.attr["mode"].CopyFrom(mirror_paddings_mode)
+    fused_conv_op.attr["strides"].CopyFrom(conv_op.attr["strides"])
+    fused_conv_op.attr["padding"].CopyFrom(conv_op.attr["padding"])
+    new_ops.extend([fused_conv_op])
```
cwhipkey: is there a way (and is it even worth it) to protect the code from there being new attributes in any of the three ops being combined, that this code then doesn't handle?

tensorflow/python/tools/optimize_for_inference_lib.py
```diff
@@ -303,3 +307,92 @@ def fold_batch_norms(input_graph_def):
...
+
+  result_graph_def = tf.GraphDef()
+  for node in input_graph_def.node:
+    if node.name in nodes_to_skip:
```
cwhipkey: we should consider asserting that the nodes being skipped are no longer used by any other nodes at this point.

tensorflow/core/ops/nn_ops.cc
```diff
@@ -425,6 +426,31 @@ data_format: Specify the data format of the input and output data. With the
...
+
+It's often possible to do spatial transformations more efficiently as part of
+the packing stage of a convolution, so this op allows for an optimized
+implementation where these stages are fused together. This prevents the need to
+write out the intermediate results as whole tensors, reducing memory pressure,
+and we can get some latency gains by merging the transformation calculations.
+
+strides: 1-D of length 4.  The stride of the sliding window for each dimension
+  of `input`. Must be in the same order as the dimension specified with format.
+padding: The type of padding algorithm to use.
```
cwhipkey: format is not supported on this op?  Maybe doc what the supported format is.

tensorflow/core/ops/nn_ops.cc
```diff
@@ -425,6 +426,31 @@ data_format: Specify the data format of the input and output data. With the
...
+    .Input("paddings: int32")
+    .Input("filter: T")
+    .Output("output: T")
+    .Attr("T: {half, float, double}")
+    .Attr("resize_align_corners: bool = false")
+    .Attr(GetMirrorPadModeAttrString())
+    .Attr("strides: list(int)")
+    .Attr(GetPaddingAttrString())
+    .Doc(R"doc(
+Performs a resize and padding as a preprocess during a convolution.
```
cwhipkey: indent doc string 2 spaces (I think that's what other ones do).

tensorflow/core/ops/nn_ops.cc
cwhipkey: add docs for input, size, paddings, and filter (listing rank/dim info is useful).

tensorflow/python/ops/nn_ops.py
cwhipkey: note that this shape fn is not used since there is no python code for creating the op in the graph.  But the C++ version of the shape function would eventually be used server side.

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+  void operator()(OpKernelContext* context, const Tensor& input,
+                  int input_batches, int resized_height, int resized_width,
+                  int padded_height, int padded_width, int input_depth,
+                  const T2* filter_data, int filter_height, int filter_width,
+                  int filter_count, int stride_rows, int stride_cols,
+                  Padding padding, T3* output_data, int output_height,
+                  int output_width, const ImageResizerState& st,
+                  int top_padding, int bottom_padding, int left_padding,
+                  int right_padding, int pad_offset) {
+    if ((input_batches <= 0) || (padded_width <= 0) || (padded_height <= 0) ||
```
cwhipkey: Are these also verified in the kernel?

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+    OP_REQUIRES_OK(context, context->GetAttr("padding", &padding_));
+  }
+
+  void Compute(OpKernelContext* context) override {
+    // Input tensor is of the following dimensions:
+    // [ batch, in_rows, in_cols, in_depth ]
+    const Tensor& input = context->input(0);
+
+    ImageResizerState st(align_corners_);
+    st.ValidateAndCalculateOutputSize(context, input);
```
cwhipkey: if validation fails, this should now return (I think you can  check the context's stats after this call to see if there was an error).

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+    const TensorShape resized_shape(
+        {input.dim_size(0), st.out_height, st.out_width, input.dim_size(3)});
+
+    const Tensor& paddings = context->input(2);
+
+    const int dims = resized_shape.dims();
+    static const int kMinDims = 0;
+    static const int kMaxDims = 5;
+    OP_REQUIRES(context, kMinDims <= dims && dims <= kMaxDims,
```
cwhipkey: is this true, or is only rank-4 supported?  (I haven't double-checked elsewhere yet).

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+                errors::Unimplemented("inputs rank not in [", kMinDims, ",",
+                                      kMaxDims, "]: ", dims));
+    OP_REQUIRES(
+        context, TensorShapeUtils::IsMatrix(paddings.shape()) &&
+                     paddings.dim_size(1) == 2,
+        errors::InvalidArgument("paddings must be a matrix with 2 columns: ",
+                                paddings.shape().DebugString()));
+    const int fixed_dims =
+        (allow_legacy_scalars() && dims == 0 && paddings.dim_size(0) == 1)
```
cwhipkey: I think we can require !allow_legacy_scalars() instead of checking it here.

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+    for (int d = 0; d < dims; ++d) {
+      const int32 before =
+          paddings_matrix(d, 0);  // Pad before existing elements.
+      const int32 after =
+          paddings_matrix(d, 1);  // Pad after exisitng elements.
+      OP_REQUIRES(context, before >= 0 && after >= 0,
+                  errors::InvalidArgument("paddings must be non-negative: ",
+                                          before, " ", after));
+      if (offset_ == 0) {  // SYMMETRIC mode.
+        OP_REQUIRES(
```
cwhipkey: This could be written in one check instead of the conditional - like with   (before <= resized_shape.dim_size(d) - offset_ ...), not sure if it's much better that way though.

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+                         after < resized_shape.dim_size(d),
+            errors::InvalidArgument("paddings must be less than"
+                                    " the dimension size: ",
+                                    before, ", ", after, " not less than ",
+                                    resized_shape.dim_size(d)));
+      }
+      padded_shape.AddDim(before + resized_shape.dim_size(d) + after);
+    }
+
+    OP_REQUIRES(
```
cwhipkey: consider moving these before the general loop over dims.

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+    OP_REQUIRES(context, padded_shape.dims() == 4,
+                errors::InvalidArgument("input must be 4-dimensional",
+                                        padded_shape.DebugString()));
+    OP_REQUIRES(context, filter.dims() == 4,
+                errors::InvalidArgument("filter must be 4-dimensional: ",
+                                        filter.shape().DebugString()));
+
+    for (int i = 0; i < 3; i++) {
+      OP_REQUIRES(context, FastBoundsCheck(filter.dim_size(i),
+                                           std::numeric_limits<int>::max()),
```
cwhipkey: int32 instead of int?

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+
+    // For 2D convolution, there should be 4 dimensions.
+    OP_REQUIRES(context, padded_shape.dims() == 4,
+                errors::InvalidArgument("input must be 4-dimensional",
+                                        padded_shape.DebugString()));
+    OP_REQUIRES(context, filter.dims() == 4,
+                errors::InvalidArgument("filter must be 4-dimensional: ",
+                                        filter.shape().DebugString()));
+
+    for (int i = 0; i < 3; i++) {
```
cwhipkey: < 4 ?

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+    // threads, because we have a single shared resource. The platforms this is
+    // aimed at have intra-op parallelism as their focus though, so it shouldn't
+    // be an issue.
+    mutex_lock lock_buffer(im2col_buffer_resource->mu);
+    core::ScopedUnref unref_buffer(im2col_buffer_resource);
+    T1* im2col_buffer = im2col_buffer_resource->data;
+
+    typename TTypes<T1, 4>::ConstTensor input_data = input.tensor<T1, 4>();
+
+    for (int batch = 0; batch < input_batches; ++batch) {
```
cwhipkey: I should have asked on the other change too - is it possible to do these directly in eigen?  (would it be faster to run that way?)

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+              in_x *= st.width_scale;
+              const int64 left_x_index = static_cast<int64>(floorf(in_x));
+              const int64 right_x_index =
+                  std::min(static_cast<int64>(ceilf(in_x)), (st.in_width - 1));
+              const T1 x_lerp = in_x - left_x_index;
+              T1* im2col_row_pixel =
+                  im2col_row_start + (filter_x * input_depth);
+              for (int in_channel = 0; in_channel < input_depth; ++in_channel) {
+                T1 in_value;
+                if ((conv_in_x >= 0) && (conv_in_x < padded_width) &&
```
cwhipkey: this checks conv_in_x and conv_in_y, but then uses top_y_index and left_x_index, and right_x_index, etc.

When the conv sees a value off the image, it should have the same mirror padding applied for that conv input pixel?

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -0,0 +1,478 @@
...
+              if (in_x < 0) {
+                in_x = -(in_x + 1.0f - pad_offset);
+              } else if (in_x >= resized_width) {
+                in_x = (resized_width * 2.0f) - (in_x + 1.0f + pad_offset);
+              }
+              in_x *= st.width_scale;
+              const int64 left_x_index = static_cast<int64>(floorf(in_x));
+              const int64 right_x_index =
+                  std::min(static_cast<int64>(ceilf(in_x)), (st.in_width - 1));
+              const T1 x_lerp = in_x - left_x_index;
```
cwhipkey: I think the lerp needs to be computed to be a fraction between 0 and 1 differently but am not positive.  Like if we are scaling down the image by width_scale = 100, then this computation needs to be different.

rmlarsen: Please use the std::floor instead of floorf. Same for ceilf etc. Could you use regular in for *_x_index?

petewarden:
We switched over to reviewing this internally, and have now got the changes checked in there, so they should appear here in github soon. Closing this PR.

PR #4055, opened 2016-08-26
Deprecate parameters from tf_workspace

damienmg:
There is no need for either of those parameters, use Bazel label
resolution instead.


rmlarsen:

PR #4061, opened 2016-08-26
Use auto-detected gcc host compiler include paths in CROSSTOOL file.

### LINKED ISSUE #4058: Use header include directories auto-detected from host compiler
As mentioned by @akors in #2109 and #3980, we should generate `cxx_builtin_include_directory` entries for the header include directories used by the user's host compiler of choice rather than hard-coding them as we are currently doing.

davidzchen:
This change removes the hard-coded cxx_builtin_include_directory entries and
instead generates cxx_builtin_include_directory entries using the compiler
include dirs detected using get_cxx_inc_directories.

This change also adds the auto_configure_fail function and uses it to improve
auto-configuration error messages.

Fixes #4058


martinwicke:
Thanks!

PR #4140, opened 2016-09-01
update reader._file_to_word_ids

afshinrahimi:
_file_to_word_ids throws a key error if there is a word in the validation or test data that doesn't exist in the training data.

The unknown words won't exist in word_to_id dictionary and so will throw a key error while converting validation and test files into ids.

Changed to way the file id sequence is built by checking if the word exists in the word_to_id dictionary to prevent the key error.


danmane:
@afshinrahimi Please sign the CLA so we can move forward with the review process.

danmane:

danmane:
Looks like the build failure was a flake due to connection timeout. I'm going to merge it. Thanks!

PR #4141, opened 2016-09-01
Fix bug where TensorBoard will crash if it can't determine host ip address

teamdandelion:


jart:
LGTM

In a future change I'd like to have it display the hostname (if it's in DNS) rather than the IP address.

Issue #1888, opened 2016-04-12
Feature request: Ability to specify some GPUs and ignore all others


rdipietro:
Right now, to specify a particular GPU, I do something like this:

```
gpu = 0
available_devices = os.environ['CUDA_VISIBLE_DEVICES'].split(',')
os.environ['CUDA_VISIBLE_DEVICES'] = available_devices[gpu]
```

after which TensorFlow will only allocate resources on the first GPU device. It would be nice to include a native way to do this.

Side note: There isn't much documentation (any?) on `device_count`, so I'm not sure if it's meant to handle this. Either way, I have experimented with the `device_count`, but with no luck: if I use `device_count = {'GPU': 1}`, TensorFlow still allocates memory on all available GPUs.

gibiansky:
I would like to implement this, as it's important for some things besides just convenience; for example, you cannot use `CUDA_VISIBLE_DEVICES` with CUDA IPC, so Tensorflow cannot be used with anything that needs access to the GPU memory in other processes (without copying it to CPU first). 

Is there a particular API that would make sense for this?

Some options:

1. Another environment variable, besides `CUDA_VISIBLE_DEVICES`, which tells Tensorflow which GPUs to use. For example, `TENSORFLOW_CUDA_DEVICES`, which *must* be a subset of `CUDA_VISIBLE_DEVICES` (if `CUDA_VISIBLE_DEVICES` is specified).
2. Modify the `GPUOptions` proto with a `repeated` field along the lines of `gpu_id`, which select which of the `CUDA_VISIBLE_DEVICES` to use. This seems cleaner, but I'm not sure how invasive that change would be (when does Tensorflow allocate GPU memory, call `cuGetDevice`, and otherwise initialize the StreamExecutor? Is it upon `tf.Session()`? Or is it earlier? If it's earlier, this change may be tricky to implement.)

Finally, if I were to implement option (2) or (1), what would I need to do in order to get the contribution accepted to Tensorflow?

gibiansky:
@zffchen78 If I were to implement this, what strategy would be most likely to get accepted as a patch to tensorflow?

[...]

PR #4185, opened 2016-09-02
Use valid GPU for allocating CUDA host memory

gibiansky:
This PR fixes (again!) issue #1888.

Before this commit, when allocating CUDA host memory, device 0 would
always be used to allocate the memory (because the memory is allocated
on the host with DMA enabled in the same way for all devices; it does
not matter which stream executor allocates the host memory). This
works, but if device 0 is not a valid device in this session (e.g. it
is turned off using GPUOptions.visible_device_list), doing so would
allocate a context on GPU 0, even when it was not being used. This
change fixes this, so that a visible device is found by looking at
which GPU allocators exist, and then using that device to allocate host
memory.

@vrv

### DIFF
--- a/tensorflow/core/common_runtime/gpu/process_state.cc
+++ b/tensorflow/core/common_runtime/gpu/process_state.cc
@@ -181,11 +181,23 @@ Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
   // different numa_nodes.  For now, just one.
   numa_node = 0;
   mutex_lock lock(mu_);
+
+  // CUDAHost alloc the same across all gpus, so just get the executor for the
+  // first valid device. We can't just use 0 because 0 may not be a valid device
+  // in this tensorflow session.
+  gpu::Platform* gpu_platform = GPUMachineManager();
+  gpu::StreamExecutor* se = nullptr;
+  for (int gpu_id = 0;
+          gpu_id < static_cast<int64>(gpu_allocators_.size());
+          ++gpu_id) {
+      if(gpu_allocators_[gpu_id] != nullptr) {
+          se = gpu_platform->ExecutorForDevice(gpu_id).ValueOrDie();
+          break;
+      }
+  }
+  CHECK_NE(se, nullptr);
+
   while (static_cast<int>(cuda_host_allocators_.size()) <= numa_node) {
-    // CUDAHost alloc the same across all gpus, so just get the
-    // executor for the first device.
-    gpu::Platform* gpu_platform = GPUMachineManager();
-    gpu::StreamExecutor* se = gpu_platform->ExecutorForDevice(0).ValueOrDie();
     CHECK(se);
     Allocator* allocator = nullptr;
     static constexpr bool kCudaHostMemoryUseBFC = true;


tensorflow/core/common_runtime/gpu/process_state.cc
```diff
@@ -181,11 +181,23 @@ Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
   // different numa_nodes.  For now, just one.
   numa_node = 0;
   mutex_lock lock(mu_);
+
+  // CUDAHost alloc the same across all gpus, so just get the executor for the
```
vrv: Maybe start this with "Any valid stream executor works for allocating CUDA host memory, so just ..."

tensorflow/core/common_runtime/gpu/process_state.cc
```diff
@@ -181,11 +181,23 @@ Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
...
+  // first valid device. We can't just use 0 because 0 may not be a valid device
+  // in this tensorflow session.
```
vrv: nit: s/tensorflow/TensorFlow :)

tensorflow/core/common_runtime/gpu/process_state.cc
```diff
@@ -181,11 +181,23 @@ Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
...
+  gpu::StreamExecutor* se = nullptr;
+  for (int gpu_id = 0;
+          gpu_id < static_cast<int64>(gpu_allocators_.size());
+          ++gpu_id) {
+      if(gpu_allocators_[gpu_id] != nullptr) {
+          se = gpu_platform->ExecutorForDevice(gpu_id).ValueOrDie();
+          break;
+      }
+  }
+  CHECK_NE(se, nullptr);
```
vrv: swap these arguments, since CHECK_NE expects "expected" to be the first argument (the error message will be confusing otherwise).

tensorflow/core/common_runtime/gpu/process_state.cc
```diff
@@ -181,11 +181,23 @@ Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
   // different numa_nodes.  For now, just one.
   numa_node = 0;
   mutex_lock lock(mu_);
+
+  // CUDAHost alloc the same across all gpus, so just get the executor for the
+  // first valid device. We can't just use 0 because 0 may not be a valid device
+  // in this tensorflow session.
+  gpu::Platform* gpu_platform = GPUMachineManager();
+  gpu::StreamExecutor* se = nullptr;
+  for (int gpu_id = 0;
```
vrv: If you could, try running clang-format --style=google on this file, it should automatically fix up spacing and indentation to conform to existing style.

tensorflow/core/common_runtime/gpu/process_state.cc
```diff
@@ -181,11 +181,23 @@ Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
...
+          gpu_id < static_cast<int64>(gpu_allocators_.size());
```
vrv: did you mean to statically cast to int instead, since you're comparing gpu_id which is an int ?

vrv:
This is pretty much what I was thinking, so thanks!  I wonder what happened to the CLA bot though, we need that to go through...

vrv:
@googlebot where are you

vrv:
I guess if this is being done as part of your work from Baidu, we'd need Baidu to sign the corp CLA.  Perhaps I should just implement this fix myself instead, unless you want to push that through ;)

gibiansky:
Probably best  to just have you re-implement it, then, especially given how small it is; this is not being done for Baidu (hence non corp CLA) and I do not think I can get a corp CLA there for this work at this point, but given how small this change is, may as well play it safe.

vrv:
ah, the commit had a baidu.com email address, so I wasn't sure.  Okay, sorry about that, legal stuff makes me sad too.  I'll send in a fix soon.

PR #4188, opened 2016-09-03
Documentation updates for r0.10(non-rc) release.

gunan:


gunan:
Not sure about the change I made for TF_Buffer.

vrv:
https://github.com/tensorflow/tensorflow/blob/r0.10/tensorflow/g3doc/api_docs/cc/StructTF_Buffer.md  you'd have to cherrypick the change that removed the generated docs too :/

gunan:
Will do right away.
However, I see that TF_Buffer is still here:
https://github.com/tensorflow/tensorflow/blob/f9ae4fe749c5d7dbf1851cffc94730086e588c35/tensorflow/c/c_api.h#L146

Is removing it (and what I am doing in this cl the right approach?

vrv:
I don't have the context for what you are doing exactly

martinwicke:
He's just rebuilding the docs -- TF_Buffer doesn't show up any more.

vrv:
LGTM.

gunan:
I confirmed with Manjunath that c_api file is still not compatible with doxygen. That is the reason for TF_Buffer to be missing. removing it from docs for now is the correct course of action.

Issue #3994, opened 2016-08-23
Cannot use Variables as gradients in apply_gradients


lightcatcher:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/optimizer.py#L287 prevents the user from using a variable as a gradient when calling apply_gradients.

I'd like to do this to accumulate gradients over multiple minibatches, and then do a single gradient update.

The current code I have is

```python
opt = tf.train.AdamOptimizer()                                                                                                   

tvs = tf.trainable_variables()
accum_vars = [tf.Variable(tf.zeros_like(tv.initialized_value()), trainable=False) for tv in tvs]                                        
zero_ops = [tv.assign(tf.zeros_like(tv)) for tv in accum_vars]

gvs = opt.compute_gradients(rmse, tvs)
accum_ops = [accum_vars[i].assign_add(gv[0]) for i, gv in enumerate(gvs)]

train_step = opt.apply_gradients([(accum_vars[i], gv[1]) for i, gv in enumerate(gvs)])  
```

that I'd like to run with logic like
```
while True:
    sess.run(zero_ops)

    for i in xrange(n_minibatches):
        sess.run(accum_ops, feed_dict=dict(X: Xs[i], y: ys[i]))

    sess.run(train_step)
```

Is there any reason variables cannot be used as arguments to apply_gradients? It seems to me that they should be able to be used as gradients. If there is a good reason, is there a recommended way to have the pattern I desire? I'm currently using the ugly hack of replacing the ``train_step`` definition with 
```
train_step = opt.apply_gradients([(accum_vars[i].assign(accum_vars[i]), gv[1]) 
                                  for i, gv in enumerate(gvs)])
```
because ``Variable.assign`` returns a tensor.

lightcatcher:
I found the ``Variable.value()`` method that seems like a nicer workaround than the ``assign`` (no-)op.
From reading the docs, it seems like like a good fix would be for ``apply_gradients`` to call ``convert_to_tensor``.

prb12:
This is a general usage question best suited to StackOverflow.  Could you please re-ask there.

lightcatcher:
Unless there is a technical reason variables cannot be used, I think this is a bug (or at least unintuitive behavior) in ``apply_gradients`` that should be fixed.

mrry:
[...]

PR #4191, opened 2016-09-04
Fix a bug #3994

DjangoPeng:
I have fixed the bug[#3994](https://github.com/tensorflow/tensorflow/issues/3994). So, now, we can use `Variables` as gradients with the update.
Check if gradient is `tf.Variable`:
- Yes, convert it to tensor.
- No, check its type as before.

Then, remake the `grads_and_vars` after checking and converting.

### DIFF
--- a/tensorflow/python/training/optimizer.py
+++ b/tensorflow/python/training/optimizer.py
@@ -27,7 +27,7 @@ from tensorflow.python.ops import gradients
 from tensorflow.python.ops import state_ops
 from tensorflow.python.ops import variables
 from tensorflow.python.training import slot_creator
-
+import tensorflow as tf
 
 class Optimizer(object):
   """Base class for optimizers.
@@ -282,8 +282,14 @@ class Optimizer(object):
     # This is a default implementation of apply_gradients() that can be shared
     # by most optimizers.  It relies on the subclass implementing the following
     # methods: _create_slots(), _prepare(), _apply_dense(), and _apply_sparse().
+
+    grads_to_tensors_list = []
+    vars_list = []
     grads_and_vars = tuple(grads_and_vars)  # Make sure repeat iteration works
     for g, v in grads_and_vars:
+      # Check if gradient is Variable. If so, convert it to tensor
+      if isinstance(g, (variables.Variable)):
+        g = tf.convert_to_tensor(g)
       if not isinstance(g, (ops.Tensor, ops.IndexedSlices, type(None))):
         raise TypeError(
             "Gradient must be a Tensor, IndexedSlices, or None: %s" % g)
@@ -292,6 +298,13 @@ class Optimizer(object):
             "Variable must be a tf.Variable: %s" % v)
       if g is not None:
         self._assert_valid_dtypes([g, v])
+      # If all is well, append g and v to corresponding list separately 
+      grads_to_tensors_list.append(g)
+      vars_list.append(v) 
+
+    # Remake the grads_and_vars after checking and converting 
+    grads_and_vars = tuple(zip(grads_to_tensors_list, vars_list))
+    
     var_list = [v for g, v in grads_and_vars if g is not None]
     if not var_list:
       raise ValueError("No gradients provided for any variable: %s" %
--- a/tensorflow/python/training/optimizer.py
+++ b/tensorflow/python/training/optimizer.py
@@ -27,7 +27,7 @@ from tensorflow.python.ops import gradients
 from tensorflow.python.ops import state_ops
 from tensorflow.python.ops import variables
 from tensorflow.python.training import slot_creator
-import tensorflow as tf
+
 
 class Optimizer(object):
   """Base class for optimizers.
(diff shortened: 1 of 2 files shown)


danmane:
@mrry Can you review this please?

tensorflow/python/training/optimizer.py
```diff
@@ -27,7 +27,7 @@
 from tensorflow.python.ops import state_ops
 from tensorflow.python.ops import variables
 from tensorflow.python.training import slot_creator
-
+import tensorflow as tf
```
mrry: Remove this import (as it creates a circular dependency, and the relevant method is available from `ops`).

tensorflow/python/training/optimizer.py
```diff
@@ -282,8 +282,14 @@ def apply_gradients(self, grads_and_vars, global_step=None, name=None):
     # This is a default implementation of apply_gradients() that can be shared
     # by most optimizers.  It relies on the subclass implementing the following
     # methods: _create_slots(), _prepare(), _apply_dense(), and _apply_sparse().
+
+    grads_to_tensors_list = []
+    vars_list = []
     grads_and_vars = tuple(grads_and_vars)  # Make sure repeat iteration works
```
mrry: I think this line is redundant now. (It was previously necessary in order to be able to iterate over the argument twice, when the argument was a generator expression. However, you now only iterate over it once.)

tensorflow/python/training/optimizer.py
```diff
@@ -282,8 +282,14 @@ def apply_gradients(self, grads_and_vars, global_step=None, name=None):
...
     for g, v in grads_and_vars:
+      # Check if gradient is Variable. If so, convert it to tensor
+      if isinstance(g, (variables.Variable)):
+        g = tf.convert_to_tensor(g)
```
mrry: It seems like it would be simpler (and cover more cases) to do:

```python
if g is not None:
  try:
    g = ops.convert_to_tensor_or_indexed_slices(g)
  except TypeError:
    raise TypeError(
        "Gradient must be convertible to a Tensor or IndexedSlices, or None: %s" %g)
```

tensorflow/python/training/optimizer.py
```diff
@@ -292,6 +298,13 @@ def apply_gradients(self, grads_and_vars, global_step=None, name=None):
             "Variable must be a tf.Variable: %s" % v)
       if g is not None:
         self._assert_valid_dtypes([g, v])
+      # If all is well, append g and v to corresponding list separately 
+      grads_to_tensors_list.append(g)
+      vars_list.append(v)
```
mrry: Instead of maintaining two lists in sync, change this to maintain a single list of tuples, called e.g. `converted_grads_and_vars` (then use `converted_grads_and_vars` in place of `grads_and_vars` in the remainder of this function).

mrry:
Please add a test that passes a `tf.Variable` to `Optimizer.apply_gradients()` in `optimizer_test.py`.

DjangoPeng:
@mrry  I have updated the `optimizer.py` and added the test that passes a `tf.Variable` to `Optimizer.apply_gradient()` in `optimizer_test.py`. But, I added an import of `ops`, because I have to use the `ops.control_dependencies()` to assist that convert gradient to `tf.Variable`. :(

tensorflow/python/training/optimizer_test.py
```diff
@@ -113,6 +114,32 @@ def testNoGradients(self):
...
+        var0 = tf.Variable([1.0, 2.0], dtype=dtype)
+        var1 = tf.Variable([3.0, 4.0], dtype=dtype)
+        cost = 5 * var0 + 3 * var1
+        global_step = tf.Variable(tf.zeros([], tf.int64), name='global_step')
+        sgd_op = tf.train.GradientDescentOptimizer(3.0)
+        grads_and_vars = sgd_op.compute_gradients(cost, [var0, var1])
+        # Convert gradients to tf.Variables
+        converted_grads = [tf.Variable(tf.zeros([2], dtype)) for i in grads_and_vars]
+        convert_ops = [tf.assign(converted_grads[i], gv[0]) for i,gv in enumerate(grads_and_vars)]
+
```
DjangoPeng: Convert the gradient(tensor) to `tf.Variable`

tensorflow/python/training/optimizer_test.py
```diff
@@ -113,6 +114,32 @@ def testNoGradients(self):
...
+        sgd_op = tf.train.GradientDescentOptimizer(3.0)
+        grads_and_vars = sgd_op.compute_gradients(cost, [var0, var1])
+        # Convert gradients to tf.Variables
+        converted_grads = [tf.Variable(tf.zeros([2], dtype)) for i in grads_and_vars]
+        convert_ops = [tf.assign(converted_grads[i], gv[0]) for i,gv in enumerate(grads_and_vars)]
+        
+        converted_grads_and_vars = list(zip(converted_grads, [var0, var1]))
+        with ops.control_dependencies(convert_ops):
+            opt_op = sgd_op.apply_gradients(converted_grads_and_vars, global_step)
+
```
DjangoPeng: Add the `ops.control_dependencies(convert_ops):` to ensure the `converted_grads` has been initialized.

tensorflow/python/training/optimizer_test.py
```diff
@@ -113,6 +114,32 @@ def testNoGradients(self):
...
+        cost = 5 * var0 + 3 * var1
+        global_step = tf.Variable(tf.zeros([], tf.int64), name='global_step')
+        sgd_op = tf.train.GradientDescentOptimizer(3.0)
+        grads_and_vars = sgd_op.compute_gradients(cost, [var0, var1])
+        # Convert gradients to tf.Variables
+        converted_grads = [tf.Variable(tf.zeros([2], dtype)) for i in grads_and_vars]
+        convert_ops = [tf.assign(converted_grads[i], gv[0]) for i,gv in enumerate(grads_and_vars)]
+        
+        converted_grads_and_vars = list(zip(converted_grads, [var0, var1]))
+        with ops.control_dependencies(convert_ops):
```
mrry: `with tf.control_dependencies(convert_ops):` should work here.

Alternatively, you could run the `convert_ops` and `opt_op` in separate steps, which might make the test clearer.

tensorflow/python/training/optimizer_test.py
```diff
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+from tensorflow.python.framework import ops
```
mrry: This import shouldn't be necessary.

tensorflow/python/training/optimizer_test.py
DjangoPeng: Sure thing, I agree to that the separate run is better than the `control_dependencies`. I'll update it!

tensorflow/python/training/optimizer_test.py
```diff
@@ -128,10 +127,11 @@ def testGradientsAsVariables(self):
         convert_ops = [tf.assign(converted_grads[i], gv[0]) for i,gv in enumerate(grads_and_vars)]
         
         converted_grads_and_vars = list(zip(converted_grads, [var0, var1]))
-        with ops.control_dependencies(convert_ops):
-            opt_op = sgd_op.apply_gradients(converted_grads_and_vars, global_step)
+        opt_op = sgd_op.apply_gradients(converted_grads_and_vars, global_step)
```
DjangoPeng: Remove the `control_dependencies`

tensorflow/python/training/optimizer_test.py
```diff
@@ -128,10 +127,11 @@ def testGradientsAsVariables(self):
...
 
         tf.initialize_all_variables().run()
+        # Run convert_ops to achieve the gradietns converting
+        sess.run(convert_ops)
         # Fetch params to validate initial values
```
DjangoPeng: separate run~

tensorflow/python/training/optimizer_test.py
```diff
@@ -128,10 +127,11 @@ def testGradientsAsVariables(self):
         convert_ops = [tf.assign(converted_grads[i], gv[0]) for i,gv in enumerate(grads_and_vars)]
         
         converted_grads_and_vars = list(zip(converted_grads, [var0, var1]))
-        with ops.control_dependencies(convert_ops):
-            opt_op = sgd_op.apply_gradients(converted_grads_and_vars, global_step)
+        opt_op = sgd_op.apply_gradients(converted_grads_and_vars, global_step)
 
         tf.initialize_all_variables().run()
+        # Run convert_ops to achieve the gradietns converting
+        sess.run(convert_ops)
```
DjangoPeng: Separately run `convert_ops` and `opt_op`!~

DjangoPeng:
@mrry I have removed the unnecessary import and `control_dependencies`

mrry:
Great, thanks! This LGTM after the tests pass.

mrry:

DjangoPeng:
@mrry Sounds good~BTW, what is LGTM :)

mrry:
It means "Looks Good To Me." And now the tests have passed, so we can merge it, thanks!

PR #4198, opened 2016-09-05
remove redundancy in moment ops

thuyen:
We subtracted the shift twice for both the first and second moments. Removing one subtraction slightly improves batch_norm performance.

### DIFF
--- a/tensorflow/python/ops/nn.py
+++ b/tensorflow/python/ops/nn.py
@@ -757,10 +757,9 @@ def sufficient_statistics(x, axes, shift=None, keep_dims=False, name=None):
     if shift is not None:
       shift = ops.convert_to_tensor(shift, name="shift")
       m_ss = math_ops.sub(x, shift)
-      v_ss = math_ops.squared_difference(x, shift)
     else:  # no shift.
       m_ss = x
-      v_ss = math_ops.square(x)
+    v_ss = math_ops.square(m_ss)
     m_ss = math_ops.reduce_sum(m_ss, axes, keep_dims=keep_dims, name="mean_ss")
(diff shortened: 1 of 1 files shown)


vincentvanhoucke:
This change is wrong, for subtle reasons. The purpose of taking the difference before squaring is that it's a lot more stable numerically. See e.g.: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
This is not a mere theoretical concern: it happens in very common scenarios.

thuyen:
This change still effectively takes the difference before squaring. It just doesn't recompute the difference (m_ss itself is the difference)

vincentvanhoucke:
Sorry, I had skimmed through your change too quickly. The issue here is different: if you use two separate kernels to compute `x-shift` and `square()` instead of the fused `squared_difference()`, you end up with two large tensors that need to be preserved through the forward *and* backward pass: 1) the `x` tensor itself needed for the backward pass through `m_ss` 2) `x-shift`, which is needed for the backward pass through `square()`. So, while your change is a bit more efficient computationally, it also effectively doubles the amount of memory that's required to keep the activations around, and that's a killer for large models.

thuyen:
Thanks Vincent. But am still a little bit confused. Isn't `x-shift` currently needed to backpropagate through reduce_sum as well? 
Currently:
`m_ss = math_ops.sub(x, shift)`
`m_ss = math_ops.reduce_sum(m_ss, axes, keep_dims=keep_dims, name="mean_ss")`

thuyen:
Oh I see, `x-shift` might not be needed for backpropagation of `reduce_sum`!
Thanks!

vincentvanhoucke:
It might be worth documenting this subtlety in the code. It's not obvious at all and has very bad consequences when done wrong.

Issue #4009, opened 2016-08-24
tf.image.decode_image would be nice (handling both png and jpeg)


untom:
`tf.image.decode_jpeg` crashes on a seemingly valid input image with an  `InvalidArgumentError: Invalid JPEG data, size 107746`. I've used both the current nightly from tonight, as well as TF 0.9.0, using Python 3.5.2 on CentOS 7 on two different machines.

The image  can be opened/displayed without problems in Firefox, GIMP and other image viewers, as well as with the `PIL` image library within Python. The image is part of the ILSVRC2015 dataset, `n02105855/n02105855_2933.JPEG`. I've  uploaded it to http://imgur.com/a/pblKL for your reference (i hope imgur doesn't recode the image, let me know). 

This is a minimal code example:

```
import tensorflow as tf
fn = './imagenet/ILSVRC2015/Data/CLS-LOC/train/n02105855/n02105855_2933.JPEG'
with tf.Graph().as_default():
    image_contents = tf.read_file(fn)
    image = tf.image.decode_jpeg(image_contents, channels=3)
    init_op = tf.initialize_all_tables()
    with tf.Session() as sess:
        sess.run(init_op)
        tmp = sess.run(image)
```

Which crashes with the following error:

```
InvalidArgumentError: Invalid JPEG data, size 107746
 [[Node: DecodeJpeg = DecodeJpeg[acceptable_fraction=1, channels=3, fancy_upscaling=true, ratio=1, try_recover_truncated=false, _device="/job:localhost/replica:0/task:0/cpu:0"](ReadFile)]]
Caused by op 'DecodeJpeg', defined at:
```

For reference, here the full stack trace is here: http://pastebin.com/GZMWDjge

girving:
That is a PNG.

girving:
Actually, I'll leave this open in case and renamed in case someone wants to finally make `decode_image` which handles both.  It's almost possible to do it in Python, or it could be done in C++ with a bunch of code sharing with `decode_png` and `decode_jpeg`.

girving:
Cc @dave-andersen to trigger painful memories. :)

girving:
Also, anyone is welcome to make the error message for `decode_png` or `decode_jpeg` include a suitably escaped version of the first few bytes, which generally give it away.

samjabrahams:
@girving would using Python's [imghdr](https://docs.python.org/3.4/library/imghdr.html) library and calling the appropriate `decode_<type>` function suffice?

[...]

PR #4222, opened 2016-09-06
Add decode_image Op

samjabrahams:
For #4009: Automatically detects whether the bytes passed in are JPEG, PNG, or neither and returns a `decode_jpeg` op, `decode_png` op, or ValueError respectively.

Before merging, need to hard-code the [`imghdr.what()`](https://docs.python.org/2/library/imghdr.html#imghdr.what) function, whose source is available as [pure Python](https://hg.python.org/cpython/file/2.7/Lib/imghdr.py)


tensorflow/python/ops/image_ops.py
```diff
@@ -1041,6 +1043,63 @@ def _DecodeGifShape(op):
...
+    if res:
+      return res
+  return None
+
+
+def decode_image(contents, channels=None, ratio=None, fancy_upscaling=None,
+                 try_recover_truncated=None, acceptable_fraction=None,
+                 name=None):
+  """Detects whether an image is a JPEG or PNG and decodes it to a uint8 tensor
+
```
samjabrahams: Need to add more documentation here.

samjabrahams:
The `_image_type()` function placeholder logic that needs to go into a separate TensorFlow operation. The code currently pushed works if pre-loaded bytes are sent to the Op, but will crash and burn if the output from another op is passed in (say, from `tf.read_file()`)

danmane:
I'm assuming that since this is still a WIP we should hold off on reviewing and testing it. Let me know if that isn't the case.

samjabrahams:
@danmane That's correct. The current code works if you load the bytes of a file ahead of time, but would throw an error if you passed in code from a function like `tf.read_file`.

The current game plan is to write a `tf.substr` Op and then utilize it in this function. [Here's that conversation for context.](https://github.com/tensorflow/tensorflow/issues/4009#issuecomment-244863602)

Thanks!

rohan100jain:
Any updates on this?

samjabrahams:
@rohan100jain Waiting for feedback on #4338: I can close this in the meantime if you'd prefer.

samjabrahams:
@rohan100jain @danmane - `tf.substr` has been merged, and I've update this Op to utilize it. There are a few things left to do, but the main functionality of the op is here. Left to do:

* More/better documentation
* More tests (should we test with small jpeg/png/gif files in addition to hard-coding in bytes?)
* Tests for gifs

vrv:
Thanks @samjabrahams !  Ping us when you think this is ready for a final review.

samjabrahams:
@vrv ready for review! Couple of judgement calls:

* Included GIF support. `decode_gif` returns a 4-D tensor as opposed to 3-D for `decode_jpeg` and `decode_png` due to the extra "number-of-frames" dimension. I'm also not sure if `decode_gif` is supposed to be exposed publicly yet (currently not in the API).
* Did not add actual image files to be read from disk and tested- stuck with hard-coded byte strings.

vrv:
@girving let me know if you have time to look at this.  I worry that the variable shape output of this convenience op will make it inconvenient :), but maybe it'll be alright.  Also, it might make sense to think about how to pass the options to decode_image in a better way -- otherwise the op's arguments are going to keep increasing to account for every image option that is added.

samjabrahams:
The different shaped output definitely seems a bit clunky. Couple options (which may or may not be any less clunky):

1. Return `tf.squeeze(tf.decode_gif(...))`. Since the `decode_gif` option always has three color channels, this will return a 3-D tensor for single-framed gifs (assuming the image is larger than 1x1) and 4-D tensors otherwise. If `decode_gif` ever gets an option for number of color channels, this will have to be modified.
2. Reshape the png and jpeg results to be 4-D. This keeps it consistent, but means that this Op requires a different workflow to the standard `decode_png` and `decode_jpeg`
3. Remove the `decode_gif` portion 
4. Leave it as is and have the end user figure it out.

For the growing number of parameters: I could use `**kwargs` and either let the user look up the options, or just leave the options out of the function signature but include them in the documentation.

girving:
I can look at it, but could we squash it before the initial review?

samjabrahams:
@girving Squashed!

tensorflow/python/kernel_tests/decode_image_op_test.py
```diff
@@ -0,0 +1,116 @@
...
+import tensorflow as tf
+
+
+class DecodeImageOpTest(tf.test.TestCase):
+
+  def testGif(self):
+    # GIF encoded 2x2 image
+    image_bytes = (b'GIF89a\x02\x00\x02\x00\xf1\x01\x00\x00\x00\x00\xff\xff\xff'
+                   b'\xff\xff\xff\x00\x00\x00!\xf9\x04\x00\x00\x00\x00\x00,\x00'
+                   b'\x00\x00\x00\x02\x00\x02\x00\x00\x02\x03D4\x05\x00;')
```
girving: Please use the same test files as `image_ops_test.py`: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/image_ops_test.py#L1953

tensorflow/python/kernel_tests/decode_image_op_test.py
```diff
@@ -0,0 +1,116 @@
...
+                   b'\x01?! \xd5N\xe1!\xc4\x9c\x00\x05\xff\xda\x00\x0c\x03\x01'
+                   b'\x00\x02\x00\x03\x00\x00\x00\x10\x9f\xff\xc4\x00\x14\x11'
+                   b'\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+                   b'\x00\x00\x00\xff\xda\x00\x08\x01\x03\x01\x01?\x10\x7f\xff'
+                   b'\xc4\x00\x14\x11\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+                   b'\x00\x00\x00\x00\x00\x00\x00\xff\xda\x00\x08\x01\x02\x01'
+                   b'\x01?\x10\x7f\xff\xc4\x00\x14\x10\x01\x00\x00\x00\x00\x00'
+                   b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\xff\xda\x00'
+                   b'\x08\x01\x01\x00\x01?\x10\x0f\xcaHM\xe0\x84\x89\xe8\xff'
+                   b'\xd9')
```
girving: Use test files as above.

tensorflow/python/kernel_tests/decode_image_op_test.py
```diff
@@ -0,0 +1,116 @@
...
+
+  def testPng(self):
+    # PNG encoded 2x2 image
+    image_bytes = (b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x02\x00\x00'
+                   b'\x00\x02\x08\x06\x00\x00\x00r\xb6\r$\x00\x00\x00\tpHYs\x00'
+                   b'\x00\x0b\x13\x00\x00\x0b\x13\x01\x00\x9a\x9c\x18\x00\x00'
+                   b'\x00\x07tIME\x07\xe0\t\x06\x023\x1c\x11\x05a\x9e\x00\x00'
+                   b'\x00\x1bIDAT\x08\xd7cd``\xf8\xdf\xd1\xd1\xc1\xc0p\xe6\xcc'
+                   b'\x99\xff\xff\xff\xff\xff\x0f\x00@\xe3\t\xf8tP\x9c\xec\x00'
+                   b'\x00\x00\x00IEND\xaeB`\x82')
```
girving: Use test files.

tensorflow/python/ops/image_ops.py
```diff
@@ -1388,6 +1391,66 @@ def _crop_and_resize_shape(op):
 ops.RegisterShape('NonMaxSuppression')(common_shapes.call_cpp_shape_fn)
 
 
+def decode_image(contents, channels=None, ratio=None, fancy_upscaling=None,
+                 try_recover_truncated=None, acceptable_fraction=None,
```
girving: I think we should just drop the `ratio`, `fancy_upscaling`, `try_recover_truncated`, and `acceptable_fraction` arguments.  If you want these details you should use the underlying format op.

tensorflow/python/ops/image_ops.py
```diff
@@ -1388,6 +1391,66 @@ def _crop_and_resize_shape(op):
...
+    acceptable_fraction: The minimum required fraction of lines before a 
+      truncated input is accepted (only used when decoding JPEG images).
+  
+  Returns:
+    `Tensor` with type `uint8`. Shape `[height, width, num_channels]` for JPEG 
+      and PNG images. Shape `[num_frames, height, width, 3]` for GIF images.
+  """
+  with ops.name_scope(name, 'decode_image') as scope:
+    def _gif():
+      return gen_image_ops.decode_gif(contents)
```
girving: The `_gif` case should handle channels somehow.

tensorflow/python/ops/image_ops.py
```diff
@@ -1388,6 +1391,66 @@ def _crop_and_resize_shape(op):
...
+
+    def _jpeg():
+      return gen_image_ops.decode_jpeg(contents, channels, ratio, 
+                                       fancy_upscaling, try_recover_truncated, 
+                                       acceptable_fraction)
+    def _png():
+      return gen_image_ops.decode_png(contents, channels, dtypes.uint8)
```
girving: `dtypes.uint8` should be unnecessary since it's the default, right?

tensorflow/python/ops/image_ops.py
girving: Don't use `gen_image_ops` and `gen_string_ops` in this function.  It should use the normal exposed names.

tensorflow/python/ops/image_ops.py
```diff
@@ -1388,6 +1391,66 @@ def _crop_and_resize_shape(op):
...
+
+    is_gif = math_ops.equal(gen_string_ops.substr(contents, 0, 4),
+                            b'\x47\x49\x46\x38')
+    is_jpeg = math_ops.equal(gen_string_ops.substr(contents, 0, 4), 
+                            b'\xff\xd8\xff\xe0')
```
girving: Only do `substr(contents, 0, 4)` once.

tensorflow/python/ops/image_ops.py
```diff
@@ -1388,6 +1391,66 @@ def _crop_and_resize_shape(op):
...
+                                               [b'Unable to decode bytes as a '
+                                                b'PNG or JPEG. Is the file '
+                                                b'encoded properly?'])
+    # Leaving default case to be decode_png
+    cases = [(is_gif, _gif),
+             (is_jpeg, _jpeg),
+            ]
+    with ops.control_dependencies([assert_decodable]):
+      return control_flow_ops.case(cases, _png, exclusive=True, 
+                                   name=scope)
```
girving: Use nested cases in the following order for speed:

1. Check for jpeg, decode jpeg if so.
2. Check for png, decode png if so.
3. Check for gif, decode gif if so.
4. Throw the assertion.

The point is that we should waste time checking for gif or png in the jpeg case, which is the most common.

girving (changes requested):

tensorflow/python/ops/image_ops.py
samjabrahams: For `_gif`: Should I use `tf.image.rgb_to_grayscale` and `tf.image.grayscale_to_rgb`, and limit `channels` to 0, 1, or 3 (exclude alpha channels for PNG)?

tensorflow/python/ops/image_ops.py
samjabrahams: @girving Was something like this what you had in mind?

```python
def _fail():
    return tf.Assert(False, [b'Not decodable!'])
def _try_gif():
    return tf.cond(is_gif, _gif, _fail)
def _try_png_gif():
    return tf.cond(is_png, _png, _try_gif)

return tf.cond(is_jpeg, _jpeg, _try_png_gif)
```

The reason I haven't set it up this way was that `tf.cond` and `tf.case` need all of the callable functions to return `Tensor` objects with the same type (so the above code wouldn't work when it tries to use the `_fail` function). I might be able to do it in a hacky way by having `_fail` call a dummy Op nested inside of a control dependency:

```python
def _fail():
    with tf.control_dependencies(tf.Assert(False, b'Not decodable!')):
        return tf.zeros([], dtype=tf.uint8)
```

girving: You can put an assert that it's a gif before the decode gif path.

tensorflow/python/ops/image_ops.py
samjabrahams: I don't think so? I'm digging around the[ `decode` implementation in tensorflow/core/lib/gif](https://github.com/tensorflow/tensorflow/blob/6b1d4fd8090d44d20fdadabf06f1a9b178c3d80c/tensorflow/core/lib/gif/gif_io.cc#L35), which currently has the number of channels hard coded to three. If you try to pass in a `channels` argument to `decode_gif`, you get an error, as [`channels` isn't registered as a parameter](https://github.com/tensorflow/tensorflow/blob/816ecb7a342c25c19daa8b19627346e1fb38e56f/tensorflow/core/ops/image_ops.cc#L522-L524). Additionally, the [`DecodeGifOp` only reads one value (the image contents) from the context](https://github.com/tensorflow/tensorflow/blob/6b1d4fd8090d44d20fdadabf06f1a9b178c3d80c/tensorflow/core/kernels/decode_gif_op.cc#L36).

samjabrahams: Got it!

tensorflow/python/kernel_tests/decode_image_op_test.py
```diff
@@ -25,84 +25,62 @@
...
+      for frame_idx, frame in enumerate(image0):
+        gt = np.zeros(shape[1:], dtype=np.uint8)
+        start = frame_idx * STRIDE
+        end = (frame_idx + 1) * STRIDE
+        if end <= WIDTH:
+          gt[:, start:end, :] = 255
+        else:
+          start -= WIDTH
+          end -= WIDTH
+          gt[start:end, :, :] = 255
```
girving: What does this bit do?

tensorflow/python/ops/image_ops.py
samjabrahams: I had to do the channels check with another assert message due to how `cond` runs it's function parameters. It seems like the Op-returning functions (`_gif`, `_png`,` _jpeg`) all run at Op creation time, which means that `_jpeg` gets unhappy when `channels` is 4 (granted, the user should probably just be using `decode_png` at that point). How does `if channels is not None or channels not in (0, 1, 3)`, throw an error at the beginning sound?

tensorflow/python/ops/image_ops.py
```diff
@@ -1406,49 +1403,37 @@ def decode_image(contents, channels=None, ratio=None, fancy_upscaling=None,
...
+    def _gif():
+      return decode_gif(contents)
+
+    header = string_ops.substr(contents, 0, 4)
+    is_jpeg = math_ops.equal(header, b'\xff\xd8\xff\xe0')
+    is_png = math_ops.equal(header, b'\211PNG')
+    is_gif = math_ops.equal(header, b'\x47\x49\x46\x38')
+
+    def _try_gif():
+      if channels is not None and channels not in (0, 3):
```
girving: `if channels not in (None, 0, 3):`

girving (changes requested):

tensorflow/python/kernel_tests/decode_image_op_test.py
samjabrahams: I grabbed this from the GIF tests in python/ops/image_ops_test.py. The [test image](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/gif/testdata/scan.gif) is a multi-framed image of a white bar travelling across a black background. It looks like this test starts with an all-black (zeros) array and then fills in the appropriate pixels with white (255) depending on which frame it is.

I can comment an explanation in or remove this check if you want.

danmane:
@samjabrahams What's the status here? I think @girving is on vacation but I can likely find someone else to review if you're ready to move forward. Also, please rebase and fix conflicts.

girving:
@danmane Sorry, Github state machine failure.  I had forgotten about this.  Looking now.

girving (approved):

girving:
Jenkins, test this please.

girving:
@samjabrahams Looks like I've delayed long enough that it needs conflict resolution.  Sorry about that. :/

samjabrahams:
@girving No worries- I've rebased and cleaned up the implementation a bit. The last major headache here is dealing with `channels` along the GIF path. Unfortunately, creating the path isn't as simple as the following:

```python
def _gif():
    if channels not in (None, 0, 3):
        raise ValueError('GIF images must be decoded with channels in (None, 0, 3)
    # ... do the op
```

Because `tf.cond` always ends up running the `_gif` function, the check happens even if there are no GIF files being decoded. That means that `channels` can't be set to 1 if we go with this approach. 

If we try using `tf.Assert` instead, it might look like this:

```python
good_op = tf.not_equal(channels, 1)   # Already checked that channels in (None, 0, 1, 3)
assert_good = tf.Assert(good_op, ['channels can't be 1!'])
with tf.control_dependencies([assert_good]):
    # ...do the rest
```
This doesn't work either, as we run into the issue of `channels` being set to `None`, which can't be passed into `equal` or `not_equal`.

In short, I'm not sure how I'd allow different channels for different image types. Here are two different ways to handle `channels` I know I can make work:

1. Use `rgb_to_grayscale` to convert the GIF image to black/white if `channels` is set to 1. It might be a little bit much going on under-the-hood, but I imagine it accomplishes the intent of what users want. The current code does this.
2. Only allow channels to be in (None, 0, 3), excluding 1 for all image types. This is a bit cleaner, though it reduces the functionality.

samjabrahams:
Also, after the move to image_ops_impl.py, it's no longer possible to use `decode_*` without prefixing with `gen_image_ops`, hence their return.

girving:
I think you should be able to use a `tf.Assert(false, ...)` as long as it has enough control dependencies to only run in the gif case (which should happen automatically if the assert is placed in try_gif).

girving:
I'm also fine with only allowing `(None, 0, 3)` channels.

samjabrahams:
@girving I believe I've got it working with an Assert checking that `channels` isn't set to 1 along the gif path.   I was being a dummy, since `None` can be interpreted as `0` by default. Let me know if this looks alright!

girving (approved):

girving:
Jenkins, test this please.

girving (approved):

girving:
@mrry: The newly added `decode_image` tests are failing to find the test files on Windows.  Do you know why that would happen on Windows but not on other platforms?

samjabrahams:
Looks like it's a path issue: `tensorflow/core/lib/gif/testdata/scan.gif : The system cannot find the path specified.` Curse you backslashes! I'll wrap the files with `os.path.join` and see if that works.

mrry:
I assume it's because the working directory isn't what the test expects. (It looks like the similar code `image_ops_test.py` isn't running as part of the Windows build because it's in `python/ops` rather than `python/kernel_tests`.)

As a quick workaround, you might try adding `WORKING_DIRECTORY ${tensorflow_source_dir}` to [this line](https://github.com/tensorflow/tensorflow/blob/9d5c3b35f59c42445a0d1e01d46e606d27ae6f7c/tensorflow/contrib/cmake/tf_tests.cmake#L102) and see if it fixes things.

samjabrahams:
@girving @mrry I pushed the change to `tf_tests.cmake` so we can see if that does the trick.

girving:
Jenkins, test this please.

samjabrahams:
May need to rerun- the GPU test had an aneurysm when trying to `git fetch`

girving:
Jenkins, test this please.

girving:
@zffchen78 The Linux GPU build has twice failed during git fetch.  Do you know what might be going wrong?

Mistobaan:
The fact that it returns different tensors shapes based on the  file format is not the 'usual' tensoflow way.
Can we just return the first frame in case of a gif image instead? If the user *really* wants the whole gif sequence then they can use the specialized version. This function will be very useful when you have mixed images formats in your data folder.

girving:
@Mistobaan What fixed shape do the other image decode functions return?

samjabrahams:
Right now `decode_jpeg` and `decode_png` return tensors with rank 3, while `decode_gif` returns one with rank 4, as it can decode multiple frames from a gif image.

girving:
Jenkins, test this please.

samjabrahams:
@Mistobaan you can always `tf.squeeze` the output of this Op of you know that you have gifs that only contain a single frame. Personally, I don't really know what the "TensorFlowian" way to do it would be, but it seems wrong to discard image data (by automatically only keeping the first frame of multi-frame gifs). That said, I wasn't sure I should even include gif support when I started this Op due to the rank differences. Here are my thoughts for options:

1. Leave it as is. Let users figure out how they want to handle gifs
2. Don't allow multi-frame gifs. Throw an assertion error if the number of frames is greater than 1, and squeeze the first dimension out
3. Leave it as is, and create a `contrib` layer that does extra work to ensure that the output ranks match for jpeg, png, and gif

girving:
I'd prefer to either leave it as is or automatically squeeze the first dimension of the gif if it's a single frame (unfortunately that requires another `tf.cond`).  Silently discarding data by picking the first frame is not a good idea.

girving:
If we want to add the optional squeeze to gif I think it can be done in a second pass.  I'm going to merge for now.

PR #4239, opened 2016-09-07
Improve documentation about custom GPU op compilation

### LINKED ISSUE #4050: Possible improvement in documentation
I'm following the documentation [here](https://www.tensorflow.org/versions/r0.10/how_tos/adding_an_op/index.html#compiling_the_kernel_for_the_gpu_device). Specifically, I find that running the second command in the code block of the section "Compiling the kernel for the GPU device" returns a linker error complaining about the lack of the `lcudart` library. I have solved this issue by adding the option `-L /usr/local/cuda-8.0/lib64/` to the command, since the `libcudart.so` library is located in that folder on my machine. I'm not sure how else this library would be located (LD_LIBRARY_PATH is only searched on execution, not compile-time), so I'm thinking this problem might not be specific to my configuration. If so, adding mention of this option on the documentation might be valuable.

zplizzi:
Fixes #4050. See issue for discussion.

### DIFF
--- a/tensorflow/g3doc/how_tos/adding_an_op/index.md
+++ b/tensorflow/g3doc/how_tos/adding_an_op/index.md
@@ -1000,6 +1000,11 @@ cuda_op_kernel.cu.o -I $TF_INC -fPIC -lcudart
 `cuda_op_kernel.so` produced above can be loaded as usual in Python, using the
 `tf.load_op_library` function.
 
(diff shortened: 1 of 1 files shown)


danmane:
@keveman  PTAL

martinwicke (approved):
LGTM

martinwicke:
Saving the environment by not testing this docs only change.

PR #4254, opened 2016-09-07
Bug in decode

chiphuyen:
Need to change the batch_size before create the model. Right now, the model has placeholders for decode to be [batch_size]. It should be 1.


danmane:
I think overriding the FLAG is not a good pattern. The FLAGS are intended to be a way for the user to specify options to the program, not as a way for the program to communicate arbitrary global state within itself. Please take a different approach.

PR #4262, opened 2016-09-08
Exclude external/local_config_cuda from being built into the PIP package

### LINKED ISSUE #4261: PIP package contains CUDA libraries
As discussed offline: the PIP package size has greatly increased since the `cuda_configure` change because the CUDA libraries under `external/local_config_cuda` are now being included by `build_pip_package.sh`.

The simplest fix for this is to special-case `local_config_cuda` to exclude it from being copied in `build_pip_package.sh` since there does not seem to be a way to specifically exclude all files under `@local_config_cuda//` in the `sh_binary` rule itself.

The reason why we did not experience this previously is because the CUDA libraries were symlinked under `third_party`, and `build_pip_package.sh` rsyncs `third_party/eigen3` specifically.

davidzchen:
Fixes #4261

### DIFF
--- a/tensorflow/tools/pip_package/build_pip_package.sh
+++ b/tensorflow/tools/pip_package/build_pip_package.sh
@@ -17,6 +17,14 @@
 
 set -e
 
+function cp_external() {
+  local src_dir=$1
+  local dest_dir=$2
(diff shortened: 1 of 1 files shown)


vrv:
LGTM

PR #4280, opened 2016-09-08
Fix type error in python_config.sh

josharian:
In the getsitepackages code path,
library_paths is a list.
Make the distutils code path a list as well.
Remove extraneous whitespace while we're here.

Without this, running ./configure:

$ ./configure 
Please specify the location of python. [Default is /Users/josh/.virtualenvs/tf/bin/python]: 
Do you wish to build TensorFlow with Google Cloud Platform support? [y/N] n
No Google Cloud Platform support will be enabled for TensorFlow
  File "<stdin>", line 18, in <module>
TypeError: can only concatenate list (not "str") to list
Found possible Python library paths:
Please input the desired Python library path to use.  Default is []

ln: util/python/python_lib: Invalid argument

### DIFF
--- a/util/python/python_config.sh
+++ b/util/python/python_config.sh
@@ -60,10 +60,10 @@ python_paths = []
 if os.getenv('PYTHONPATH') is not None:
   python_paths = os.getenv('PYTHONPATH').split(':')
 try:
-  library_paths =  site.getsitepackages()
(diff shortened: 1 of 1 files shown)


danmane:
@itsmeolivia PTAL

itsmeolivia:
LGTM

danmane:

PR #4294, opened 2016-09-09
Fix a typo in g3doc/how_tos/tool_developers

shinh:
Add `()' after f.read so we pass a string rather than a
function.


caisq:

caisq:
LGTM

PR #4300, opened 2016-09-09
Add -D_FORCE_INLINES to nvcc opts

### LINKED ISSUE #4103: Build failure while following TensorFlow from source following TF documentation
### Problem

TensorFlow build  of the pip package with CUDA support using the following command, taken from the [TensorFlow documentation](https://www.tensorflow.org/versions/r0.10/get_started/os_setup.html#installing-from-sources), fails:

```bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package```

### Error message

```
/usr/include/string.h: In function 'void* __mempcpy_inline(void*, const void*, size_t)':
/usr/include/string.h:652:42: error: 'memcpy' was not declared in this scope
   return (char *) memcpy (__dest, __src, __n) + __n;
                                          ^
ERROR: /home/marek/src/tensorflow/tensorflow/contrib/rnn/BUILD:63:1: output 'tensorflow/contrib/rnn/_objs/python/ops/_gru_ops_gpu/tensorflow/contrib/rnn/kernels/gru_ops_gpu.cu.pic.o' was not created.
ERROR: /home/marek/src/tensorflow/tensorflow/contrib/rnn/BUILD:63:1: not all outputs were created.
```

### Environment

- OS: Ubuntu 16.04 (64-bit)
- CUDA 8.0
- GCC 4.9.3
- Java 1.8 (r101_b13)
- Python 2.7.12 (Anaconda 4.1.1)
- Bazel 0.3.1

### Related problems for CUDA-based frameworks reported on StackOverflow and Github
https://github.com/tensorflow/tensorflow/issues/1346
https://github.com/torch/torch7/issues/670
https://github.com/BVLC/caffe/issues/4046
https://github.com/opencv/opencv/issues/6500

### Solutions tried (none of them worked)

- downgrading to CUDA 7.5 and CUDNN 4
- changing downgrading and upgrading GCC away from 4.9.3
- changing GCC and cmake options to include ```-D_FORCE_INLINES``` flag (see related Github issues above)

### Solution that worked

Build succeeds when the ```-c opt``` option is removed from the Bazel build command. That is, the final command should be

```bazel build --config=cuda //tensorflow/tools/pip_package:build_pip_package```

martinwicke:
Fixes #4103.

### DIFF
--- a/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc.tpl
+++ b/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc.tpl
@@ -229,7 +229,7 @@ def InvokeNvcc(argv, log=False):
   out = ' -o ' + out_file[0]
 
   supported_cuda_compute_capabilities = [ %{cuda_compute_capabilities} ]
-  nvccopts = ''
+  nvccopts = '-D_FORCE_INLINES '
   for capability in supported_cuda_compute_capabilities:
     capability = capability.replace('.', '')
     nvccopts += r'-gencode=arch=compute_%s,\"code=sm_%s,compute_%s\" ' % (


martinwicke:
@zheng-xq please let me know if this is a bad idea (it seems to fix a problem on recent Ubuntu)

@damienmg let me know if there's a better way to set this define

damienmg:
@martinwicke I like your mention bot, can we have the same? :) Can we train tensorflow to do review instead of us :D 

Anyway LGTM AFAICT but I don't now about the nvcc flag itself. For gcc we call the compiler to know if the option is supported but for NVCC it is probably ok to just include it raw.

martinwicke:
Since it's just a `-D` I'm pretty confident it will be supported. Whether it's honored is another question, but it seems to solve problems. :)

martinwicke:
Oh, and the mention bot, just add it as a webhook!

martinwicke:
The mention bot can just be added as a webhook. It's awesome. You should
add it too. I wish I had the same for issues, and we'll be working on
something like that soon. :)

PR #4310, opened 2016-09-09
Add PIP package creation to the CMake build.

mrry:


martinwicke:
Since the setup.py is mostly identical, is it possible to somehow reuse the existing one?

mrry:
@martinwicke I'd rather keep this simplified one as a temporary measure (until the two PIP packages are at parity) than refactor the main `setup.py` with a bunch of conditional flags that we're eventually not going to need. Does that make sense?

martinwicke:
Makes sense to me. Just wanted to check that this is deliberate.

gunan:
I think we need a new build for this in jenkins, the existing pull-requests-cmake would probably not cut it.
Do we have any scripts for building the pip package for this build?
What is the command to build it?

gunan:
Found it, we will first merge this, and refer to this one through a script similar to what we have here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/build_pip_package.sh

Issue #4312, opened 2016-09-10
Can't configure due to undeclared packages inside if_android/if_ios


gustavla:
I am having trouble configuring the latest master branch (dbe7ee0dfa9e5ab26284522379f2747510fc267b). When I run `./configure`, I get:
```
ERROR: [...]/tensorflow/tensorflow/contrib/session_bundle/BUILD:107:1: no such target '//tensorflow/core:android_lib_lite': target 'android_lib_lite' not declared in package 'tensorflow/core' defined by [...]/tensorflow/tensorflow/core/BUILD and referenced by '//tensorflow/contrib/session_bundle:session_bundle'.
ERROR: [...]/tensorflow/tensorflow/contrib/session_bundle/BUILD:213:1: no such target '//tensorflow/core:android_lib_lite': target 'android_lib_lite' not declared in package 'tensorflow/core' defined by [...]/tensorflow/tensorflow/core/BUILD and referenced by '//tensorflow/contrib/session_bundle:signature'.
ERROR: [...]/tensorflow/tensorflow/contrib/session_bundle/BUILD:107:1: no such target '//tensorflow/core:meta_graph_portable_proto': target 'meta_graph_portable_proto' not declared in package 'tensorflow/core' defined by [...]/tensorflow/tensorflow/core/BUILD and referenced by '//tensorflow/contrib/session_bundle:session_bundle'.
ERROR: [...]/tensorflow/tensorflow/contrib/session_bundle/BUILD:213:1: no such target '//tensorflow/core:meta_graph_portable_proto': target 'meta_graph_portable_proto' not declared in package 'tensorflow/core' defined by [...]/tensorflow/tensorflow/core/BUILD and referenced by '//tensorflow/contrib/session_bundle:signature'.
ERROR: [...]/tensorflow/tensorflow/contrib/session_bundle/BUILD:213:1: no such target '//tensorflow/core:meta_graph_portable_proto': target 'meta_graph_portable_proto' not declared in package 'tensorflow/core' defined by [...]/tensorflow/tensorflow/core/BUILD and referenced by '//tensorflow/contrib/session_bundle:signature'.
[...]

Issue #3743, opened 2016-08-11
Bazel missing dependency declaration for Eigen


gustavla:
Compilation error that suggests Bazel is not recognizing the Eigen header files. Occurs consistently for me on all versions starting from 21716d8f6e175cd6e8cd97a84e48497574268b0c up until the current master (00726750c9f7dedfc57685c5011e21df3b5b3706).

### Environment info
64-bit CentOS / cuda 7.5.18 / cudnn 5.0.5 / bazel 0.2.1 / java 1.8.0_91

### Steps to reproduce
Run `./configure` and set up compilation with CUDA 7.5 and CuDNN 5.0.

    bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package

Compiles fine without `--config=cuda`.

### What have you tried?
I ran a `git bisect` and traced it to 21716d8f6e175cd6e8cd97a84e48497574268b0c. More specifically, it is moving Eigen out of the `eigen-eigen-b4fa9622b809` folder that for some reason is causing this. Reverting only these changes from the master makes it compile just fine (see https://github.com/gustavla/tensorflow/commit/058a6517576171bb0c7dcf11ec1aac04bd59aedf). I have so far been unable to fix it while keeping the `strip_prefix` in.

It suspect it might be a quirk in bazel 0.2.1. It would be great if anyone else running that version can confirm.

### Error
The error message (abridged and personal paths removed):
```
ERROR: /.../tensorflow/core/kernels/BUILD:1527:1: undeclared inclusion(s) in rule '//tensorflow/core/kernels:depth_space_ops_gpu':
this rule is missing dependency declarations for the following files included by 'tensorflow/core/kernels/spacetodepth_op_gpu.cu.cc':
  '/.../.cache/external/eigen_archive/Eigen/Core'
  '/.../.cache/external/eigen_archive/Eigen/src/Core/util/DisableStupidWarnings.h'
  '/.../.cache/external/eigen_archive/Eigen/src/Core/util/Macros.h'
  ...
  '/.../.cache/external/eigen_archive/Eigen/src/Core/ArrayWrapper.h'
  '/.../.cache/external/eigen_archive/Eigen/src/Core/GlobalFunctions.h'
  '/.../.cache/external/eigen_archive/Eigen/src/Core/util/ReenableStupidWarnings.h'.
```
Issues with related error messages include #1157 and  #3589.

girving:
[...]

PR #4315, opened 2016-09-10
Pass through -fno-canonical-system-headers

gustavla:
This fixes the critical build issue #3743. After I figured out what was causing the issue, I couldn't really get a discussion of possible solutions going. I figured I would just go ahead and create a PR with the simplest and most immediate fix to get the ball rolling.

The fix is simple, gcc needs to be passed `-fno-canonical-system-headers`, and this PR makes it recognize this flag if it's passed to `crosstool_wrapper_driver_is_not_gcc` and pass it along to the compiler command.

However, one could argue that there are more general and better solutions, such as passing through all `-f...` flags (there are actually currently many flags that are passed into the Python script that are not passed along). If anyone feels strongly about such an alternative solution, I would like to discuss the details of it before putting together a replacement PR.

I might need to rebase this build before it can be merged. I cloned an older commit since the master head is broken for me (#4312).

### DIFF
--- a/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc.tpl
+++ b/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc.tpl
@@ -95,6 +95,7 @@ def GetHostCompilerOptions(argv):
   parser.add_argument('-iquote', nargs='*', action='append')
   parser.add_argument('--sysroot', nargs=1)
   parser.add_argument('-g', nargs='*', action='append')
+  parser.add_argument('-fno-canonical-system-headers', action='store_true')
 
   args, _ = parser.parse_known_args(argv)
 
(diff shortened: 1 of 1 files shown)


martinwicke:
Thanks for breaking out the PR!

@davidzchen @damienmg what's the "proper" thing to do here? Should all -f options be passed through? All options? Is there a reason not to pass through certain options?

davidzchen:
If `crosstool_wrapper_driver_is_not_gcc` is meant to provide a command line interface that is more compatible with gcc, my intuition is that all -f options should be passed through.

+@lberki for his thoughts on this.

damienmg:
I would pull people with good C++ knowledge for a general answer that but for this specific flag it is correct to pass it through.

Some option might not be supported by nvcc and you might not want to pass them through.

martinwicke:
Jenkins, test this please

martinwicke:
gustavia@ can you rebase?

gustavla:
@martinwicke Done.

martinwicke:
Jenkins, test this please

martinwicke (approved):
LGTM

PR #4316, opened 2016-09-10
WMT en-fr training corpus filename change

gokceneraslan:
`giga-fren.release2.fr.gz` file has been renamed to `giga-fren.release2.fixed.fr.gz`


gokceneraslan:
Ah, reported already. Closing.

Issue #4232, opened 2016-09-06
Create tf.substr Op


samjabrahams:
Would be useful for a number of purposes, such as extracting data, substring matches (#4009), etc. Proposed API (matches [C++ style substrings](http://www.cplusplus.com/reference/string/string/substr/)):

```
Args:
  inputs: a `Tensor` of type `string`. The strings that will have substrings extracted
  pos: `int32` or `int64` `Tensor`. Position of the first character to be copied as a substring
  len: `int32` or `int64` `Tensor`. Number of characters to include in the substring

Returns:
  A `Tensor` of type `string` containing substrings of `inputs`
```

DjangoPeng:
@samjabrahams  I want to try adding the substr Op :)

samjabrahams:
@DjangoPeng Oh no! I already sent in a PR two days ago with #4338. I'm planning on using the Op to finish work on the `tf.decode_image` Op, so I knocked out a version over the weekend.

DjangoPeng:
@samjabrahams Sounds good~ Look forward to your update~

samjabrahams:
Substr (#4338) has been merged into master

PR #4338, opened 2016-09-12
Add Substr Op

samjabrahams:
Adds the `tf.substr` Op, as proposed in #4232. Once merged, work can continue on the `tf.decode_image` Op (#4009).


girving:
OOO, please reassign.

andrewharp:

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,77 @@
...
+      // Validate size of tensors
+      OP_REQUIRES(context, TensorShapeUtils::IsScalar(pos_tensor.shape()),
+                  errors::InvalidArgument("pos must be a scalar, but got: ",
+                                          pos_tensor.shape().DebugString()));
+      OP_REQUIRES(context, TensorShapeUtils::IsScalar(len_tensor.shape()),
+                  errors::InvalidArgument("len must be a scalar, but got: ",
+                                          len_tensor.shape().DebugString()));
+
+      auto input = input_tensor.flat<string>();
+      size_t pos = pos_tensor.scalar<T>()(0);
```
vrv: I think you want this to be just ()();  Scalars have no offset

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,77 @@
...
+      size_t len = len_tensor.scalar<T>()(0);
```
vrv: These should be type T, otherwise your OP_REQUIRES check doesn't work for >= 0.

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,77 @@
...
+
+      // Allocate output
+      Tensor* output_tensor = nullptr;
+      OP_REQUIRES_OK(context,
+                     context->allocate_output("output", input_tensor.shape(),
+                                              &output_tensor));
+      auto output = output_tensor->flat<string>();
+
+      // Set output to be substrings of input strings
+      for (int i = 0; i < input.size(); i++) {
```
vrv: size_t i = 0; i < input.NumElements(); ++i?

tensorflow/core/ops/string_ops.cc
```diff
@@ -280,4 +280,32 @@ input: Base64 strings to decode.
 output: Decoded strings.
 )doc");
 
+REGISTER_OP("Substr")
+  .Input("input: string")
+  .Input("pos: T")
+  .Input("len: T")
+  .Output("output: string")
+  .Attr("T: {int32, int64}")
+  .SetShapeFn(shape_inference::UnchangedShape)
```
vrv: I think you can do a bit more here, validating that pos and len are both scalars:

```
.SetShapeFn([](InferenceContext* c) {
  ShapeHandle unused;
  TF_RETURN_IF_ERROR(c->WithRank(c->input(1), 0, &unused));
  TF_RETURN_IF_ERROR(c->WithRank(c->input(2), 0, &unused));
  return shape_inference::UnchangedShape(c);
}
```

If you do this, you might be able to just elide the shape checks in the kernel itself.

Of course, this depends on the discussion above about the shapes, so just FYI for now.

tensorflow/core/ops/string_ops.cc
```diff
@@ -280,4 +280,32 @@ input: Base64 strings to decode.
 output: Decoded strings.
 )doc");
 
+REGISTER_OP("Substr")
```
vrv: I'm a little dubious about this op as written -- I can see the use case if all inputs were scalars, but applying pos/len scalars to all elements of an input seems potentially limiting.

I was maybe thinking that pos and len could be equal-shaped tensors, and have pos and len be broadcastable against 'input'.  For example, if they were scalars, then they would apply to every element.

If the shape of 'pos' and 'len' matched 'input', then the entries in pos and len would be elementwise applied to input.  Otherwise, they would follow our standard broadcasting rules.  (e.g., you could have a matrix [MxN] of strings, have pos and len be [1xN], and then pos and len would apply row-wise to every element of the input matrix).

From an API perspective, that seems more flexible, even if for now, you only implement the scalar or equal-shaped versions and threw errors for the rest.

tensorflow/python/ops/string_ops.py
```diff
@@ -117,6 +118,7 @@ def string_split(source, delimiter=" "):  # pylint: disable=invalid-name
 ops.NotDifferentiable("ReduceJoin")
 ops.NotDifferentiable("StringJoin")
 ops.NotDifferentiable("StringSplit")
+ops.NotDifferentiable("Substr")
```
vrv: Is this needed?  Typically you only need to add this if the op will ever be used in a context where its gradient might be asked for.  Given the use cases, I imagine this will never be backpropped on, and it would be better for this to raise an error due to 'no gradient function found' instead of silently truncating.

vrv (changes requested):
Thanks -- I have partial context from the decode_image op, and @girving is OOO for a while, so I'll do my best.

tensorflow/core/kernels/substr_op.cc
samjabrahams: Got it- giving this a shot.

tensorflow/core/ops/string_ops.cc
samjabrahams: I was thinking about doing a broadcasted version of this Op, but I wasn't exactly sure how to approach it, so I pushed what I had. I'll give this a go over the next few days!

tensorflow/python/ops/string_ops.py
samjabrahams: This was added as I was cleaning up this PR, since most of the string ops are registered as non-differentiable. I'll remove it if you think it's unnecessary.

samjabrahams:
Still working on this; I've been exploring the TensorFlow core to see what I could pick up. I've added in the element-wise version of this Op and have been working at making broadcasting work.

I'm trying to use the functionality in `core/util/bcast.h`, but it's not quite there yet. I've been able to get the appropriate Tensor shape calculations, but I don't know how to convert the `TensorBroadcastingOp` back into a `Eigen::TensorMap` (or similar) so that I can actually access the values associated with the broadcasted Tensor. I've commented out my latest attempt on [line 99 of substr_op.cc](https://github.com/tensorflow/tensorflow/pull/4338/files#diff-8ce346dacb5764a51c596536a50de93eR99)- is this close to how I should attempt this, or am I barking up the wrong tree?

I'll need to update the shape function to match the broadcast result, as well.

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,145 @@
...
+        //                context->allocate_output("output", output_shape,
+        //                                         &output_tensor));
+        // switch (ndims) {
+        //   case 2: {
+        //     auto output = output_tensor->shaped<string, 2>(bcast.x_reshape());
+        //     auto input_reshaped = input_tensor.shaped<string, 2>(bcast.x_reshape());
+        //     auto pos_reshaped = pos_tensor.shaped<T, 2>(bcast.y_reshape());
+        //     auto len_reshaped = len_tensor.shaped<T, 2>(bcast.y_reshape());
+              
+        //     typename TTypes<string, 2>::Tensor input = input_reshaped.broadcast(BCast::ToIndexArray<2>(bcast.x_bcast()));
```
samjabrahams: This is the section where I've been having trouble. Compilation returns errors along the lines of the following:

```
tensorflow/core/kernels/substr_op.cc:117:120: error: conversion from 'const Eigen::TensorBroadcastingOp<const Eigen::array<long int, 2ul>, const Eigen::TensorMap<Eigen::Tensor<const std::basic_string<char>, 2, 1, long int>, 16> >' to non-scalar type 'tensorflow::TTypes<std::basic_string<char>, 2, long int>::Tensor {aka Eigen::TensorMap<Eigen::Tensor<std::basic_string<char>, 2, 1, long int>, 16>}' requested
             typename TTypes<string, 2>::Tensor input = input_reshaped.broadcast(BCast::ToIndexArray<2>(bcast.x_bcast()));
```

vrv:
I admit that my brain shuts off when I see Eigen compilation failures, bringing in @benoitsteiner to help. :P

samjabrahams:
Glad to hear it's not just me :P Looking forward to seeing how to get this implemented.

samjabrahams:
Pinging @benoitsteiner if you have time. Any recommendations for accessing data from a `Eigen::TensorBroadcastingOp`? I don't have a huge amount of Eigen experience, so I'm not sure where I'm going wrong in my code above (or if I'm barking up the wrong tree entirely)

tensorflow/core/kernels/substr_op.cc
vrv: Hmm, have you considered trying to look at cwise_ops.h and cwise_ops_common.h for helpful structures or templates to do the broadcasting?

samjabrahams: I have- my first attempts involved trying to use that code. The main problem I was encountering is that those end up using the Eigen::Matrix::binaryExpr function under-the-hood to do the actual computation. Since substring takes in three parameters (input string, start position of the substring, and length of substring), I was having trouble finding a clean way to pass in all of the necessary data using the BinaryOp pattern.

vrv: Ah, good point.  Well, if @benoitsteiner doesn't get back to you soon, perhaps stick with the scalar implementation for now, since I suspect that's good enough for your usecase.

Perhaps:

1) Implement the shape spec using broadcastable shapes.  E.g., have the C++ shape spec use BroadcastBinaryOpShapeFn in common_shape_fns.h so that the "spec" allows broadcasting: it's probably sufficient to check that shape(pos) == shape(len) and then do BroadcastBinaryOpShapefn on input and pos.

2) In your kernel implementation, check that the pos/len are scalar, and if not, return an errors::Unimplemented, so it's clear that it *should* be implemented but is not yet.

Probably also mention that true ternary broadcasting might be necessary to make it efficient, since the looping you're doing even for the scalar approach isn't parallel.

samjabrahams: Sounds good. Do you want me to leave in the case where pos/len are the same shape as the input string?

vrv: Yeah, whatever cases you can handle easily seems fine.

samjabrahams:
I think this is ready for a re-review. Here's the highlights of the changes:

* Moved check to see if pos/len are the same shape inside of shape function
* Shape function uses `BroadcastBinaryOpShapeFn`
* `substr_op.cc` checks for scalar pos/len or pos/len with the same shape as input tensor at the beginning of the Op. If it is neither, sets an Unimplemented error status
* Cleaned up `substr_op.cc` to remove most of duplicated code
* Added tests to make sure pos/len shape mismatches throw appropriate errors

I have a few questions about some judgement calls:

* Since pos/len shapes are checked by the shape function, I removed checking again inside of the kernel implementation. Is that fine, or should we double check?
* I left a second Unimplemented error in the final `else` block toward the bottom of the kernel implementation. It's identical to the one at the top of the code, and shouldn't be reachable. I left it in to make it extra-extra explicit that this is where broadcasting implementations would go. I can take this out and just leave the comments explaining the situation if that's preferable.

tensorflow/core/kernels/substr_op.cc
benoitsteiner: The issue is the following: you need to initialize the TensorMap first, and then you can populate its values by assigning the result of the broadcast operation to it. That should work:
 typename TTypes<string, 2>::Tensor input(buffer, first_dim, second_dim);
 input = input_reshaped.broadcast(BCast::ToIndexArray<2>(bcast.x_bcast()));

That means that you need to manually allocate a temporary buffer to hold the result. You can do this by call OpKernelContext::allocate_temp.

samjabrahams:
Thanks @benoitsteiner for your notes- from them I was able to get broadcasting to work for the Op. I've pushed the latest changes to the Op along with updated tests which pass on my machine. I attempted to do some clean-up with a macro (as different dimensionality broadcasts have to be defined separately and similarly), but was hitting some cross-initialization hiccups. Here are the main notes for @vrv:

* Added 1 and 2 dimension broadcasts in `substr.cc`. They still end up iterating through the tensors sequentially, however
* Added/updated `_testBroadcast` and `_testBadBroadcast` in `substr_op_test.py`
* I left in the base cases of pos/len with either scalar or matching shape to the input strings. This way, we avoid allocating the additional memory used in broadcasting

Assuming things look good, the last thing I need to do is update the documentation to indicate that 1/2 dimensional broadcasting is supported.

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,211 @@
...
+        auto input = input_tensor.flat<string>();
+        // Allocate output
+        Tensor* output_tensor = nullptr;
+        OP_REQUIRES_OK(context,
+                       context->allocate_output("output", input_tensor.shape(),
+                                                &output_tensor));
+        auto output = output_tensor->flat<string>();
+        if (is_scalar) {
+          // Perform Op with scalar pos/len
+          T pos = pos_tensor.scalar<T>()();
```
vrv: Can you use SubtleMustCopy on this and 'len'?  It's to ensure that the input tensors can't be manipulated to do out-of-bounds access due to timing attacks.

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,211 @@
...
+          T len = len_tensor.scalar<T>()();
+          for (size_t i = 0; i < input_tensor.NumElements(); ++i) {
+            // Make sure pos won't cause a runtime error
+            OP_REQUIRES(context, pos >= 0 && pos < input(i).size(),
```
vrv: One faster way to do this check is to compute the min and max of the pos tensor outside this loop using Eigen, and validating that they are within bounds (that way, the reductions will be faster).  One thing to do later if this ever becomes performance critical.

vrv (commented):
The non-broadcasting path looks good!  will let benoit look at the broadcasting side.  I'll kick off a test to verify.

vrv:

samjabrahams:
I forgot to add the `from __future__ import print_function` statement to the kernel tests. Fixed.

vrv:

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,220 @@
...
+                                                input(i), "' at index ", i));
+            output(i) = input(i).substr(pos, len);
+          }
+        } else {
+          // Perform Op element-wise with tensor pos/len
+          auto pos_flat = pos_tensor.flat<T>();
+          auto len_flat = len_tensor.flat<T>();
+          // Use SubtleMustCopy on pos/len to prevent async attacks
+          const size_t num_elements = pos_tensor.NumElements();
+          std::vector<T> pos(num_elements);
```
samjabrahams: @vrv Double checking if I should use `SubtleMustCopy` on the element-wise (and eventually broadcasting) path. I borrowed this code from [`space_to_batch`](https://github.com/tensorflow/tensorflow/blob/d24e64227653ab3cab7405c79ca166063a42f0ff/tensorflow/core/kernels/spacetobatch_functor.h#L52-L60).

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,220 @@
...
+          std::vector<T> len(num_elements);
+          for (size_t i = 0; i < num_elements; ++i) {
+            pos[i] = tensorflow::internal::SubtleMustCopy(pos_flat(i));
+            len[i] = tensorflow::internal::SubtleMustCopy(len_flat(i));
+          }
+          for (size_t i = 0; i < input_tensor.NumElements(); ++i) {
+            // Make sure pos won't cause a runtime error
+            OP_REQUIRES(context, pos[i] >= 0 && pos[i] < input(i).size(),
```
samjabrahams: Should I be using [`FastBoundsCheck`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/bounds_check.h#L30) here instead of the boolean expression I have?

samjabrahams (commented):
Couple of questions related to using `SubtleMustCopy` and `FastBoundsCheck`

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,220 @@
...
+                                                  output_shape,
+                                                  &len_buffer));
+            typename TTypes<T,2>::Tensor len_bcast = len_buffer.shaped<T,2>(
+                                                          bcast.result_shape());
+            len_bcast = len.broadcast(BCast::ToIndexArray<2>(bcast.y_bcast()));
+            
+            // Iterate through broadcasted tensors and perform substr
+            for (int i = 0; i < output_shape.dim_size(0); ++i) {              
+              for (int j = 0; j < output_shape.dim_size(1); ++j) {
+                output(i, j) = input_bcast(i, j).substr(pos_bcast(i, j),
```
vrv: and another subtlemustcopy here too

vrv (commented):
Great, thanks for going through this process!  Just a few more from my side, but then this probably looks good.

samjabrahams:
Absolutely- thanks for reviewing! I'll have these changes made later today.

tensorflow/core/kernels/substr_op.cc
```diff
@@ -0,0 +1,220 @@
...
+                           context->allocate_temp(DataTypeToEnum<T>::v(),
+                                                  output_shape,
+                                                  &len_buffer));
+            typename TTypes<T,1>::Tensor len_bcast = len_buffer.shaped<T,1>(
+                                                          bcast.result_shape());
+            len_bcast = len.broadcast(BCast::ToIndexArray<1>(bcast.y_bcast()));
+            
+            // Iterate through broadcasted tensors and perform substr
+            for (int i = 0; i < output_shape.dim_size(0); ++i) {              
+              output(i) = input_bcast(i).substr(pos_bcast(i), len_bcast(i));
```
samjabrahams: Quick question: should I create vectors for pos/len ahead before this loop, or just call SubtleMustCopy as we loop through the tensors?

vrv: I think you can just copy them as you loop through -- that's apparently how it's used elsewhere in the codebase, so it should be okay!

samjabrahams: Got it, thanks!

samjabrahams:
Alright! I've added the calls to `SubtleMustCopy` and `FastBoundsCheck`. Additionally, I modified the loops to avoid multiple accessor calls to the string tensor (not sure if it will make much of a difference) and added examples to the docstring in string_ops.cc. If this works, then I believe it just needs the LGTM from @benoitsteiner

tensorflow/python/kernel_tests/substr_op_test.py
```diff
@@ -0,0 +1,235 @@
...
+# limitations under the License.
+# ==============================================================================
+
+"""Tests for Substr op from string_ops."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
```
vrv: can you import numpy before tensorflow  for alphabetical ordering?

vrv:
(No need to squash, we'll squash when we merge).  Just one suggestion above and I'll kick off tests and merge.

samjabrahams:
Fixed!

vrv:
Great!  @tensorflow-jenkins test this please

vrv (approved):

samjabrahams:
Not sure what's up with the `tensorflow/contrib/learn/python/learn/monitors_test.py` failure. I'm not able to replicate the test failure on my machine with the PR branch (I'm not sure how this code would affect that test, regardless).

vrv:

samjabrahams:
Rebased to master to see if that helps.

vrv:

samjabrahams:
Awesome- thanks for your help and guidance @vrv and @benoitsteiner!

PR #4345, opened 2016-09-13
Add code to close session

vra:
Hi all, I read the tensorflow tutorial and find that we must close session when finishing using it. However, when I review the code in [cifar10 tutorial](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/models/image/cifar10), I find that in some files,  there is no `sess.close()` line. I don't know whether it's a function of `tf.app` utilities, so I add the session closing code and make you know about this issue.  Hope this can be helpful.


andrewharp:
@vra Thanks for the commit! My assumption though is that session.close() will be called automatically by the __del__ method when session goes out of scope, so functionally the logic here would be identical.  @shlens can you confirm?

shlens:
Correct, the Session is closed when it goes out of scope so this change is not necessary.

vra:
Okay I got it. Thanks for your explanation. I will close this PR.

PR #4351, opened 2016-09-13
Changes in default value

pajayrao:


ajayraop3495:
I signed it!

tensorflow/python/training/optimizer.py
```diff
@@ -385,7 +385,7 @@ def _assert_valid_dtypes(self, tensors):
   def _valid_dtypes(self):
     """Valid types for loss, variables and gradients.
 
-    Defaults to `float32`. Subclasses should override to allow other types.
+    Subclasses should override to allow other types.
```
andrewharp: Would it be more informative but still correct to say "types other than floats"?

tensorflow/python/training/optimizer.py
```diff
@@ -385,7 +385,7 @@ def _assert_valid_dtypes(self, tensors):
   def _valid_dtypes(self):
     """Valid types for loss, variables and gradients.
 
-    Subclasses should override to allow other types.
+    Subclasses should override to allow other float types.
```
andrewharp: Just to confirm, are float types the only types that could potentially added here? Or are doubles, ints, etc also possible?

tensorflow/python/training/optimizer.py
pajayrao: not really, as loss and gradients values will be float always.

tensorflow/python/training/optimizer.py
andrewharp: Nevermind, just saw your other reply, thanks.

PR #4364, opened 2016-09-13
Fix to allow compilation for Android API 19

steverichey:
This is a (very) small change to allow compiling TensorFlow for Android API 19. Without this change, you get `tensorflow/core/platform/posix/net.cc:44:51: error: 'errno' was not declared in this scope` when compiling the example Android project with the `WORKSPACE` `android_ndk_repository` parameter `api_level` set to 19. This change resolves that error and allows compilation for this target.


mrry:

mrry:
LGTM.

PR #4370, opened 2016-09-14
[Test]Fix the array type bug

DjangoPeng:
Fix the `PS_ARRAY` and `WORKER_ARRAY`  bug. They should be arrays, but in fact they are strings. The problem is the wrong way for assigning bash array. 
The log info before fixing:

```
2 worker process(es) running in parallel...
Worker 0: 
  WORKER HOST: localhost:2223 localhost:2224 
  log file: /tmp/worker0.log
Worker 1: 
  WORKER HOST: 
  log file: /tmp/worker1.log
```

After:

```
2 worker process(es) running in parallel...
Worker 0: 
  WORKER HOST: localhost:2223
  log file: /tmp/worker0.log
Worker 1: 
  WORKER HOST: localhost:2224
  log file: /tmp/worker1.log
```

### DIFF
--- a/tensorflow/tools/gcs_test/Dockerfile
+++ b/tensorflow/tools/gcs_test/Dockerfile
@@ -16,7 +16,7 @@ RUN ./install_google_cloud_sdk.bash --disable-prompts --install-dir=/var/gcloud
 
 # Install nightly TensorFlow pip
 RUN pip install \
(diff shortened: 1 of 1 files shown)


tensorflow/tools/dist_test/scripts/dist_mnist_test.sh
```diff
@@ -147,7 +147,7 @@ done
 # Get N_WORKERS by WORKER_HOSTS
 N_WORKERS=$(echo ${WORKER_HOSTS} | awk -F "," '{printf NF}')
 # Replace the delimiter with " "
-WORKER_ARRAY=$(echo ${WORKER_HOSTS} | awk -F "," '{for(i=1;i<=NF;i++){printf $i" "}}')
+WORKER_ARRAY=($(echo ${WORKER_HOSTS} | awk -F "," '{for(i=1;i<=NF;i++){printf $i" "}}'))
```
DjangoPeng: The right way for assign bash array: `array=($string)`. It's obvious that the old `WORKER_ARRAY` gets a string instead of an array. The same as `PS_ARRAY`.

andrewharp:

PR #4372, opened 2016-09-14
Typo fix in Android Example (gradle)

eerwitt:
It seems like the `t` is a mistake, although it may be a feature which I'm unfamiliar with.


andrewharp:
@eerwitt Correct, that was an accidental keypress just prior to submit. Thanks for the catch!

PR #4383, opened 2016-09-14
Removed duplicate section in Tensorboard README

floriandotpy:
There were two sections present, both talking about multiple/distributed summary writes. The sections were almost identical. I have removed one of them.

### DIFF
--- a/tensorflow/tensorboard/README.md
+++ b/tensorflow/tensorboard/README.md
@@ -191,15 +191,6 @@ tutorial](https://www.tensorflow.org/versions/r0.10/how_tos/graph_viz/index.html
 
 # Frequently Asked Questions
 
-### Does TensorBoard support distributed summary writers?
-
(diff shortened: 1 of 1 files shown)


florianletsch:
I signed it!

tensorflow/tensorboard/README.md
```diff
@@ -191,15 +191,6 @@ tutorial](https://www.tensorflow.org/versions/r0.10/how_tos/graph_viz/index.html
 
 # Frequently Asked Questions
 
-### Does TensorBoard support distributed summary writers?
```
andrewharp: Can you please add:
"Violating the assumption that there is a single events file may cause TensorBoard to lose data or stop updating unexpectedly." to the other section, as this seems like an important thing to note.

danmane:
Thanks for the PR. However, this was fixed internally a few days ago, so this PR will be redundant once we sync changes. I'm closing it.

Issue #3825, opened 2016-08-15
A bug with a default initialiser and tf.nn.rnn_cell.LSTMCell/tf.contrib.rnn.LSTMFusedCell


avostryakov:
### Environment info
Operating System:
Mac OS X 10.9.5 

Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):
No CUDA

If installed from binary pip package, provide:
pip install http://ci.tensorflow.org/view/Nightly/job/nightly-matrix-cpu/TF_BUILD_CONTAINER_TYPE=CPU,TF_BUILD_IS_OPT=OPT,TF_BUILD_IS_PIP=PIP,TF_BUILD_PYTHON_VERSION=PYTHON2,label=mac1-slave/lastSuccessfulBuild/artifact/pip_test/whl/tensorflow-0.10.0rc0-py2-none-any.whl

1. Which pip package you installed.
2. The output from `python -c "import tensorflow; print(tensorflow.__version__)"`.
0.10.0rc0

### Steps to reproduce
1. A minimum example to reproduce the bug:

```
import numpy as np
import tensorflow as tf

with tf.variable_scope("rnn", initializer=tf.contrib.layers.xavier_initializer()):
    out_cell = tf.nn.rnn_cell.LSTMCell(num_units=100, state_is_tuple=True, use_peepholes=True)
    #out_cell = tf.contrib.rnn.LSTMFusedCell(num_units=100)
    input_var = tf.placeholder(dtype=tf.float32, shape=(32, 50, 100))

    d, _ = tf.nn.dynamic_rnn(out_cell, input_var, dtype=tf.float32, parallel_iterations=32,
                                 scope="output_rnn")
    init = tf.initialize_all_variables()
    with tf.Session() as sess:
        input = np.random.uniform(-1.0, 1.0, [32, 50, 100])
        sess.run(init)
        o = sess.run([d], {input_var: input})

```
2. The same bug if you comment LSTMCell and uncomment LSTMFusedCell which was renamed to LSTMBlockCell recently

3. If I remove ", initializer=tf.contrib.layers.xavier_initializer()" in tf.variable_scope code will work without errrors with normal distribution by default. Or if I remove use_peepholes=True in LSTMCell the bug will disappear too.

### What have you tried?
1.

### Logs or other output that would be helpful
Traceback (most recent call last):
....
[...]

PR #4405, opened 2016-09-16
Deal with 1d shape in variance_scaling_initializer.

imironhead:
[issue#3825](https://github.com/tensorflow/tensorflow/issues/3825)

An initializer could be used anywhere. If the shape for variance_scaling_initializer is 1d, we can use the same value for both fan_in & fan_out.


imironhead:
change author

PR #4411, opened 2016-09-16
Add gradient to BiasAddGrad op.

### LINKED ISSUE #4174: tf.nn.bias_add does not support multiple derivatives (using with contrib.learn)
Apparently, `tf.nn.bias_add` does not support differentiating twice: I created a neural network with `tf.contrib.layers`, and attempting to compute a Hessian matrix (actually Hessian-vector product, based on https://en.wikipedia.org/wiki/Hessian_automatic_differentiation) yields a `LookupError` saying that `BiasAddGrad` has no gradient.

From what I understand, `bias_add` is basically `add` with support for different types.

Is there any way to allow `bias_add` to be differentiated multiple times? Perhaps at the very least we could make the use of `bias_add` optional in `layers`.

admcrae:
A rather inelegant solution to fix #4174.

### DIFF
--- a/tensorflow/python/kernel_tests/bias_op_test.py
+++ b/tensorflow/python/kernel_tests/bias_op_test.py
@@ -20,7 +20,6 @@ from __future__ import print_function
 
 import numpy as np
 import tensorflow as tf
-from tensorflow.python.ops import gen_nn_ops
 
 
 def GetTestConfigs():
@@ -131,8 +130,8 @@ class BiasAddTest(tf.test.TestCase):
           bias_tensor, bias.shape, output_tensor, np_input.shape)
          
       # Test gradient of BiasAddGrad
-      bias_add_grad = gen_nn_ops.bias_add_grad(out_backprop=output_tensor,
-                                               data_format=data_format)
+      bias_add_grad = tf.gradients(tf.nn.l2_loss(output_tensor),
+                                   bias_tensor)[0]
       
       grad_jacob_t, grad_jacob_n = tf.test.compute_gradient(
           output_tensor, np_input.shape, bias_add_grad, bias.shape
@@ -156,8 +155,8 @@ class BiasAddTest(tf.test.TestCase):
             bias_tensor, bias.shape, output_tensor, np_input.shape)
         
         
-        bias_add_grad = gen_nn_ops.bias_add_grad(out_backprop=output_tensor,
-                                                 data_format=data_format)
+        bias_add_grad = tf.gradients(tf.nn.l2_loss(output_tensor),
+                                     bias_tensor)[0]
                                                  
         
         _, grad_jacob_n = tf.test.compute_gradient(
--- a/tensorflow/python/kernel_tests/bias_op_test.py
+++ b/tensorflow/python/kernel_tests/bias_op_test.py
@@ -132,11 +132,8 @@ class BiasAddTest(tf.test.TestCase):
       # Test gradient of BiasAddGrad
       bias_add_grad = tf.gradients(tf.nn.l2_loss(output_tensor),
                                    bias_tensor)[0]
-      
       grad_jacob_t, grad_jacob_n = tf.test.compute_gradient(
-          output_tensor, np_input.shape, bias_add_grad, bias.shape
-      )
-      
+          output_tensor, np_input.shape, bias_add_grad, bias.shape)
       
       if dtype == np.float16:
         # Compare fp16 theoretical gradients to fp32 numerical gradients,
@@ -154,14 +151,10 @@ class BiasAddTest(tf.test.TestCase):
         _, bias_jacob_n = tf.test.compute_gradient(
             bias_tensor, bias.shape, output_tensor, np_input.shape)
         
-        
         bias_add_grad = tf.gradients(tf.nn.l2_loss(output_tensor),
                                      bias_tensor)[0]
-                                                 
-        
         _, grad_jacob_n = tf.test.compute_gradient(
-            output_tensor, np_input.shape, bias_add_grad, bias.shape
-        )
+            output_tensor, np_input.shape, bias_add_grad, bias.shape)
         
       threshold = 2e-3
       if dtype == tf.float64:
--- a/tensorflow/python/kernel_tests/bias_op_test.py
+++ b/tensorflow/python/kernel_tests/bias_op_test.py
@@ -20,6 +20,7 @@ from __future__ import print_function
 
 import numpy as np
 import tensorflow as tf
+from tensorflow.python.ops import gen_nn_ops
 
 
 def GetTestConfigs():
@@ -128,7 +129,16 @@ class BiasAddTest(tf.test.TestCase):
           input_tensor, np_input.shape, output_tensor, np_input.shape)
       bias_jacob_t, bias_jacob_n = tf.test.compute_gradient(
           bias_tensor, bias.shape, output_tensor, np_input.shape)
-
+         
+      # Test gradient of BiasAddGrad
+      bias_add_grad = gen_nn_ops.bias_add_grad(out_backprop=output_tensor,
+                                               data_format=data_format)
+      
+      grad_jacob_t, grad_jacob_n = tf.test.compute_gradient(
+          output_tensor, np_input.shape, bias_add_grad, bias.shape
+      )
+      
+      
       if dtype == np.float16:
         # Compare fp16 theoretical gradients to fp32 numerical gradients,
         # since fp16 numerical gradients are too imprecise unless great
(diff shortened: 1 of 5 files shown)


tensorflow/python/ops/nn_grad.py
```diff
@@ -203,6 +204,27 @@ def _BiasAddGrad(op, received_grad):
...
+    is the gradient of the bias term in BiasAdd)
+  """
+  
+  try:
+    data_format = op.get_attr("data_format")
+  except ValueError:
+    data_format = None
+    
+  zeros = gen_array_ops._zeros_like(op.inputs[0])
+  return gen_nn_ops._bias_add(zeros, received_grad, data_format=data_format)
```
zheng-xq: In cases where the shapes are known, I think a broadcasting op such as tf.tile will be better. Since it avoids allocating the zero input array and add the unnecessary zeros. However, tf.tile in its current form may be slower. So a bit benchmarking is needed for that path. I'll leave it up to you whether you want to try that. 

Also we need to add a test using compute_gradient_errort to make sure this passes the gradient checker.

admcrae: I don't see a file in tensorflow/python/ops for testing nn gradient ops. Is it somewhere else? Otherwise, should I create a new nn_grad_test file or just use the current nn_test file?

zheng-xq: Please use bias_op_test. Thanks.

tensorflow/python/kernel_tests/bias_op_test.py
```diff
@@ -128,7 +129,16 @@ def _testGradient(self, np_input, bias, dtype, data_format, use_gpu):
           input_tensor, np_input.shape, output_tensor, np_input.shape)
       bias_jacob_t, bias_jacob_n = tf.test.compute_gradient(
           bias_tensor, bias.shape, output_tensor, np_input.shape)
-
+         
+      # Test gradient of BiasAddGrad
+      bias_add_grad = gen_nn_ops.bias_add_grad(out_backprop=output_tensor,
```
zheng-xq: Instead of calling the underlying function, maybe calling "tf.Gradient(loss, bias)". So even if the underlying implementation changes, the test might still be valid.

tensorflow/python/ops/nn_grad.py
```diff
@@ -203,6 +203,48 @@ def _BiasAddGrad(op, received_grad):
...
+    data_format = op.get_attr("data_format")
+  except ValueError:
+    data_format = None
+    
+  #zeros = array_ops.zeros_like(op.inputs[0])
+  #return gen_nn_ops._bias_add(zeros, received_grad, data_format=data_format)
+  
+  shape = array_ops.shape(op.inputs[0])
+  rank = array_ops.rank(op.inputs[0])
+  bias_shape = array_ops.shape(received_grad)
```
zheng-xq: If the shape is not fully defined at this point, this adds an unnecessary dependency at runtime. So it is better to do this when the shape is known, and do your previous bias_add when it is not. For example:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/array_ops.py#L154

If possible, please also include a benchmark result to show this is not actually slower for non-trivially large shape. tf.tile may have a few performance problems on GPU, and it is on our roadmap for improvement.

admcrae: Some crude benchmarks:
[benchmark.zip](https://github.com/tensorflow/tensorflow/files/479926/benchmark.zip)

The benchmarks were done on a workstation with an i7-965 (yes, we laughed too when we got it) and a GTX 1080.

The results are a bit odd. Using `bias_add` with `zeros_like` seems to give a negligible to moderate performance advantage in most cases, but it is 2-5x slower on the CPU with statically determined shapes (it looks like something is not being optimized properly here). I did notice that on the GPU, using `bias_add` seemed to result in somewhat higher GPU usage overall.

I'm inclined to stay with `tf.tile` for now, since (1) it is a more direct solution (using `bias_add` seems extremely hacky), (2) its performance is more consistent, and (3) you mentioned that the performance of `tf.tile` might be improved in the future.

Of course, the optimal solution would be to write another kernel for this operation (and then we could finally end the chain of gradient ops, since this operation's gradient is identical to `bias_add_grad`), but I'm not sure if that is worthwhile for something that most machine learning people aren't going to need.

tensorflow/python/ops/nn_grad.py
```diff
@@ -203,6 +203,43 @@ def _BiasAddGrad(op, received_grad):
...
+    is the gradient of the bias term in BiasAdd)
+  """
+  
+  try:
+    data_format = op.get_attr("data_format")
+  except ValueError:
+    data_format = None
+  
+  shape = array_ops.shape(op.inputs[0])
+  rank = array_ops.rank(op.inputs[0])
```
zheng-xq: Could you assert that rank is 4 for now? For conv3d and 5D tensors from videos, we will have to make adjustment. However, there are plenty of other code that needs to be modified. So just a check would suffice for now.

zheng-xq (commented):
I agree with the assessment on tf.tile is better for now. With a bit more nitpicking, I think we will be ready to go.

martinwicke:
Changing assignment to @zheng-xq to reflect the reality of the review.

tensorflow/python/ops/nn_grad.py
admcrae: `contrib.layers` [uses `bias_add` even for fully-connected layers](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/layers/python/layers/layers.py#L857) (where the tensors are 2-D). I'd be hesitant to mess that up if anyone wanted to do higher-order gradients with that code (that's where I originally noticed this issue, although I'm not using it right now).

zheng-xq: Thanks a lot for getting to that. It is indeed an important use case. Could you add a test case to cover that as well? We need to make sure future changes do not break that.

admcrae: [`bias_op_test.py`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/bias_op_test.py) appears to have plenty of tests for ranks 2 and 3 but, curiously, no test for rank-4 tensors (for the bias_add op itself; it has gradient tests for ranks 2 and 4). Is this what you meant?

tensorflow/python/kernel_tests/bias_op_test.py
```diff
@@ -128,7 +128,16 @@ def _testGradient(self, np_input, bias, dtype, data_format, use_gpu):
...
-
+         
+      # Test gradient of BiasAddGrad
+      bias_add_grad = tf.gradients(tf.nn.l2_loss(output_tensor),
+                                   bias_tensor)[0]
+      
+      grad_jacob_t, grad_jacob_n = tf.test.compute_gradient(
+          output_tensor, np_input.shape, bias_add_grad, bias.shape
+      )
+
```
zheng-xq: One less new line please.

tensorflow/python/kernel_tests/bias_op_test.py
```diff
@@ -144,12 +153,22 @@ def _testGradient(self, np_input, bias, dtype, data_format, use_gpu):
             input_tensor, np_input.shape, output_tensor, np_input.shape)
         _, bias_jacob_n = tf.test.compute_gradient(
             bias_tensor, bias.shape, output_tensor, np_input.shape)
-
+
```
zheng-xq: Same here.

tensorflow/python/kernel_tests/bias_op_test.py
```diff
@@ -144,12 +153,22 @@ def _testGradient(self, np_input, bias, dtype, data_format, use_gpu):
             input_tensor, np_input.shape, output_tensor, np_input.shape)
         _, bias_jacob_n = tf.test.compute_gradient(
             bias_tensor, bias.shape, output_tensor, np_input.shape)
-
+        
+        
+        bias_add_grad = tf.gradients(tf.nn.l2_loss(output_tensor),
+                                     bias_tensor)[0]
+
```
zheng-xq: And here.

zheng-xq (commented):
Looks good, other than cosmetic suggestions.

admcrae:
That better? Note I haven't made any changes to the tests since your previous comment--I wasn't entirely sure what you wanted there.

zheng-xq:
LGTM.

martinwicke (commented):
reusing zheng-xq's LGTM

martinwicke (approved):

zheng-xq:
@martinwicke, could you confirm the mac test failure is unrelated? Thanks.

martinwicke:
Yes, unrelated. Good to merge.

martinwicke:
@jhseu feel free to use your powers and override the failed test.

tensorflow/python/kernel_tests/bias_op_test.py
```diff
@@ -128,7 +128,13 @@ def _testGradient(self, np_input, bias, dtype, data_format, use_gpu):
           input_tensor, np_input.shape, output_tensor, np_input.shape)
       bias_jacob_t, bias_jacob_n = tf.test.compute_gradient(
           bias_tensor, bias.shape, output_tensor, np_input.shape)
-
+         
+      # Test gradient of BiasAddGrad
+      bias_add_grad = tf.gradients(tf.nn.l2_loss(output_tensor),
+                                   bias_tensor)[0]
+      grad_jacob_t, grad_jacob_n = tf.test.compute_gradient(
+          output_tensor, np_input.shape, bias_add_grad, bias.shape)
```
gunan: I think we have a python3 issue here. Our nightlies are failing on this test:
http://ci.tensorflow.org/view/Nightly/job/nightly-matrix-linux-gpu/TF_BUILD_IS_OPT=OPT,TF_BUILD_IS_PIP=NO_PIP,TF_BUILD_PYTHON_VERSION=PYTHON3,label=gpu-linux/247/consoleFull

I was also able to reproduce the problem locally using the following command:
$ bazel test tensorflow/python/kernel_tests:bias_op_test -c opt --config=cuda

admcrae: Is there anything I can do here? This appears to be limited to Python3 with a GPU, so it seems unlikely that this is due to my code.

admcrae: To clarify, right now the data_format of the bias op is being compared to a Python string literal. Looking at [another similar code snippet](https://github.com/tensorflow/tensorflow/blob/450e9386e0cc4b81b49adb9184859411f43ef259/tensorflow/python/framework/common_shapes.py#L185) suggests that should explicitly make it an ASCII string. I'll fix that shortly.

gunan: Sorry, I was reading through my email backlog, and just saw your update.
Thank you very much for looking into this!

PR #4054, opened 2016-08-26
Fix stddev computation in TensorBoard examples.

bamos:

PR #4426, opened 2016-09-17

shohad25:


martinwicke:
I believe this has been fixed at head.

martinwicke:
(and in r0.10)

vrv:
(If it was in r0.10, wouldn't this PR diff to nothing?)

martinwicke:
Who knows what that diff is doing, but it's not a diff against r0.10: 
[retrain.py](https://github.com/tensorflow/tensorflow/blob/r0.10/tensorflow/examples/image_retraining/retrain.py) has no mention of `_sum(` in it.

martinwicke:
I did that on the weekend. Should be fixed.

Issue #3495, opened 2016-07-25
Show the collection that a variable belongs to in TensorBoard


cgarciae:
This is specially useful when using functions like `tf.contrib.layers.fully_connected` that generate the weight and bias variables on their own. With this you could simply inspect in TensorBoard and get that information right away, no need to look in the docs or ask.

shkr:
I can work on this

danmane:
@dsmilkov @jameswex Can either of you give some guidance on how shkr could approach working on this?

jameswex:
Is this for viewing variables in TensorBoard in the Events tab, or in the Graph tab? I'd like to understand a specific example of what this ask is for. Thanks.

cgarciae:
@jameswex The request would be for the Graph tab.
When trying to implement regularization using `tf.contrib.layers.fully_connected`, it accepts some parameters handle regularization for weights and biases, but the Tensors needed for regularization end up in the `REGULARIZATION_LOSSES` collection, but its hard to know unless you ask. It would be nice if you could see the collection(s) that a Tensor belongs e.g. REGULARIZATION_LOSSES in the TensorBoard Graph viewer.

If it helps, here is the code I use to get those regularization terms
```
tf.get_collection(tf.GraphKeys.REGULARIZATION_LOSSES)
```

jameswex:
I'm sorry I'm still not quite following the issue and proposed solution. Can you provide a screenshot of the confusion caused by how the variables are displayed in the graph in your use case?

shkr:
My understanding is that @cgarciae is requesting to add a collections field in the node for each variable either present in the red outlined box or the top right box for the selected variable. Currently the collections that a tensor belongs cannot be seen on tensorboard. Please point out if I am missing any section where it is present.

![screen shot 2016-09-04 at 11 25 45 am](https://cloud.githubusercontent.com/assets/5184832/18233055/6115d206-7292-11e6-8406-520f099d8038.png)
](url)

@cgarciae @jameswex if that is the correct formulation of the feature request, then I can move ahead. Please confirm.

cgarciae:
Exactly what @shkr said!

jameswex:
That explanation helps. Sounds like you would want to add the collections a variable belongs to as an entry in the attr map in the NodeDef for a variable node so that that information is saved for every node in the graph definition proto. Then the graph visualizer can use that information to add a collections list to the tf_node_info info card for any selected variable.

shkr:
[...]

PR #4433, opened 2016-09-18
Add a collections attribute to node_def

shkr:
This updates add_to_collection method in the Python API to update (or create) a ListValue of `collections` in the attr map of each node def. This has been prescribed by @jameswex to solve Issue #3495, in order to have tensorboard show the collections that a node belongs to in the node info box.

### DIFF
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -2665,6 +2665,13 @@ class Graph(object):
     """
     self._check_not_finalized()
     with self._lock:
+      # if node does not have collections attr, create the key
+      if not value._variable.op.node_def.attr.__contains__('collections'):
+          value._variable.op.node_def.attr['collections'].list\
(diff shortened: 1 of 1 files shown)


mrry:
Thanks for proposing this change. From my reading of issue #3495, it looks like we already have a more general representation for graph collections, in the [`MetaGraphDef.collection_def`](https://github.com/tensorflow/tensorflow/blob/7df9c6860e00b91eda0e550b11d9be52d9341d85/tensorflow/core/protobuf/meta_graph.proto#L65) field, and a better solution would be to enhance the graph visualizer so that it could accept `MetaGraphDef` protocol buffers in addition to `GraphDef` protos. Since a `MetaGraphDef` includes a `GraphDef` as one of its fields, this should not be too difficult.

I'm going to close this PR now, but we'd be glad to review a contribution that re-uses the existing tools for representing graph collections in the `MetaGraphDef`.

PR #4435, opened 2016-09-18
Fix cmake files for highwayhash and re2.

gunan:
Tested via the following:
$ cd tensorflow/contrib/cmake
$ mkdir build
$ cd build
$ cmake ..
$ make

### DIFF
--- a/tensorflow/contrib/cmake/external/highwayhash.cmake
+++ b/tensorflow/contrib/cmake/external/highwayhash.cmake
@@ -12,9 +12,9 @@ set(highwayhash_STATIC_LIBRARIES
 )
 set(highwayhash_INCLUDES ${highwayhash_BUILD})
 
-set(highwayhash_HEADERS
-    "${highwayhash_BUILD}/highwayhash/*.h"
-)
+file(GLOB highwayhash_HEADERS RELATIVE
+     "${highwayhash_BUILD}/highwayhash"
+     "${highwayhash_BUILD}/highwayhash/*.h")
 
 ExternalProject_Add(highwayhash
     PREFIX highwayhash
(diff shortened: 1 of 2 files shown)


tensorflow/contrib/cmake/external/highwayhash.cmake
```diff
@@ -12,9 +12,9 @@ set(highwayhash_STATIC_LIBRARIES
 )
 set(highwayhash_INCLUDES ${highwayhash_BUILD})
 
-set(highwayhash_HEADERS
-    "${highwayhash_BUILD}/highwayhash/*.h"
-)
+file(GLOB highwayhash_HEADERS RELATIVE
+     "${highwayhash_BUILD}/highwayhash"
+     "${highwayhash_BUILD}/highwayhash/*.h")
```
mrry: I'm not sure why this makes a difference - and it seems to be failing on the CI build :(. Is it a matter of CMake version inconsistency?

I also had to modify this file to get the right thing to happen on Windows: the replacement I used there was:

```cmake
add_custom_command(TARGET highwayhash_copy_headers_to_destination PRE_BUILD
COMMAND ${CMAKE_COMMAND} -E copy_directory ${highwayhash_INSTALL}/include/ ${highwayhash_INCLUDE_DIR}/highwayhash)
```

Would that maybe work in both cases?

gunan: I have no idea why it fails in CI build.
I have logged into the machine it tests and tried running cmake.
It seems to work nicely. I will try rerunning ci build. 
I will also test this on windows.

gunan:
Jenkins, test this please.

PR #4448, opened 2016-09-19
Fixed cuda_configure.bzl for Windows

meteorcloudy:
`cuda_configure` is the first blocker when building tensorflow on Windows. We shouldn't even run cuda_configure if cuda support is not enabled, but this patch could unblock us easily. @mrry @damienmg @dslomov


davidzchen:
Jenkins, test this please.

davidzchen (approved):

davidzchen:
@mrry - This change looks good to me. Can you take a look and merge if there are no other feedback? Thanks!

martinwicke:
Thanks for the fix!

PR #4452, opened 2016-09-19
contrib/quantization: Open file in binary mode

olesalscheider:
This fixes the following error:
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 53: invalid start byte


petewarden (approved):
This looks good to me, thanks for the fix!

petewarden:
Jenkins, test this please.

PR #4454, opened 2016-09-19
Make grpc buildable on Windows

meteorcloudy:
grpc++_unsecure needs zlib to build, adding @zlib_archive as a dependency

With this patch `@grpc//:grpc_cpp_plugin` and `@grpc//:grpc++_unsecure` is buildable on Windows
@mrry @damienmg @dslomov

### DIFF
--- a/grpc.BUILD
+++ b/grpc.BUILD
@@ -1224,6 +1224,7 @@ cc_library(
   deps = [
     ":gpr",
     "//external:nanopb",
+    "@zlib_archive//:zlib",
   ],
   copts = [
     "-std=gnu99",
--- a/grpc.BUILD
+++ b/grpc.BUILD
@@ -1224,7 +1224,7 @@ cc_library(
   deps = [
     ":gpr",
     "//external:nanopb",
-    "@zlib_archive//:zlib",
+    "//external:zlib",
   ],
   copts = [
     "-std=gnu99",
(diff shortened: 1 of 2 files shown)


mrry:
@meteorcloudy Is there a corresponding issue against https://github.com/grpc/grpc for adding this? We'd like to keep the diff between our forked/fixed build file and the standard one small, so that we can switch back to using it instead.

mrry:

damienmg:
Don't we have gRPC on Bazel too? Having just to vendor gRPC would be preferable for Bazel also.

mrry:
@damienmg Yes, but there are two open issues on the Bazel support that prevent us from using it directly: https://github.com/grpc/grpc/issues/7707 and https://github.com/grpc/grpc/issues/7851.

meteorcloudy:
@mrry I see.  I can send a PR to grpc to add `//external:zlib` as a dependency of `grpc++_unsecure` which [they already did for `grpc` target](https://github.com/grpc/grpc/commit/f2f24e2f93bba8ae939beb53980ac8e042c372c0). Then we bind `@zlib_archive//:zlib` as `zlib` in `workspace.bzl`. This should slove the problem without touching `grpc.BUILD`

mrry (approved):
LGTM, but it'd be great to get this upstream in gRPC.

mrry:

PR #4457, opened 2016-09-19
Improved iOS camera example and binary footprint optimizations

petewarden:
This PR has a variety of cleanup changes and fixes to help demonstrate advanced features of iOS development like memory mapping and loading retrained models. It also includes some related changes that help shrink the binary footprint of the final executable on iOS.


petewarden:
@cwhipkey this PR isn't urgent, but I was wondering if you could review it?

tensorflow/contrib/ios_examples/README.md
```diff
@@ -13,7 +13,17 @@ This folder contains examples of how to build applications for iOS devices using
...
    [Inception v1](https://storage.googleapis.com/download.tensorflow.org/models/inception5h.zip),
    and extract the label and graph files into the data folders inside both the
-   simple and camera examples.
+   simple and camera examples:
+
+```bash
+mkdir -p ~/graphs
+curl -o ~/graphs/inception.zip \
+ https://storage.googleapis.com/download.tensorflow.org/models/inception5h.zip \
+ && unzip ~/graphs/inception.zip -d ~/graphs/inception
```
cwhipkey: should this unzip take inception5h.zip instead of inception.zip?

tensorflow/contrib/ios_examples/camera/CameraExampleViewController.mm
```diff
@@ -22,6 +22,27 @@
...
 
+// If you have your own model, modify this to the file name, and make sure
+// you've added the file to your app resources too.
+static NSString* model_file_name = @"tensorflow_inception_graph";
+static NSString* model_file_type = @"pb";
+// This controls whether we'll be loading a plain GraphDef proto, or a
+// file created by the convert_graphdef_memmapped_format utility that wraps a
+// GraphDef and parameter file that can be mapped into memory from file to
+// reduce overall memory usage.
+const bool model_uses_memory_mapping = false;
```
cwhipkey: should this and the ones further below (wanted_input_* through output_layer_name) be static too?

tensorflow/contrib/ios_examples/camera/camera_example.xcodeproj/project.pbxproj
```diff
@@ -304,12 +304,6 @@
 				OTHER_LDFLAGS = (
 					"-force_load",
 					"$(SRCROOT)/../../makefile/gen/lib/libtensorflow-core.a",
-					"-Xlinker",
```
cwhipkey: are these changes intentional?

tensorflow/contrib/ios_examples/camera/tensorflow_utils.mm
```diff
@@ -119,57 +121,108 @@ bool PortableReadFileToProto(const std::string& file_name,
...
+      &tensorflow_graph);
+  if (!load_graph_status.ok()) {
+    LOG(ERROR) << "MMap load graph failed with "
+               << load_graph_status.error_message();
+    return load_graph_status;
+  }
+
+  tensorflow::SessionOptions options;
+  // Disable optimizations on this graph so that constant folding doesn't
+  // increase the memory footprint by creating new constant copies of the weight
```
cwhipkey: Does it do this even if the constant is not used in some other expression?  We should probably make it so that single ImmutableConst nodes don't get constant-folded into single constant nodes.

cwhipkey: Does it do this even if the constant is not used in some other expression?  We should probably make it so that single ImmutableConst nodes don't get constant-folded into single constant nodes.

That said, skipping optimization on all mobile models would make sense anyway if we just run the optimizations as an offline process?

tensorflow/contrib/ios_examples/simple/tf_ios_makefile_example.xcodeproj/project.pbxproj
```diff
@@ -302,14 +306,18 @@
...
 				PRODUCT_BUNDLE_IDENTIFIER = "com.google.TF-Test";
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SEPARATE_STRIP = NO;
 			};
 			name = Debug;
 		};
 		591157B41CF4011D00C31E3A /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CLANG_DEBUG_INFORMATION_LEVEL = default;
```
cwhipkey: do we want all of these?

For debug_information_level, is the default for Release essentially 'none'?

exceptions and rtti are not needed by tensorflow and at least rtti adds code size .  (although maybe rtti is needed if we are not using proto lite runtime here).

tensorflow/contrib/makefile/Makefile
```diff
@@ -140,14 +140,15 @@ $(shell mkdir -p $(DEPDIR) >/dev/null)
...
-CXXFLAGS := --std=c++11 -DIS_SLIM_BUILD $(OPTFLAGS)
+CXXFLAGS := --std=c++11 -DIS_SLIM_BUILD -fno-exceptions -DNDEBUG $(OPTFLAGS)
 LDFLAGS := \
 -L/usr/local/lib
 DEPFLAGS = -MT $@ -MMD -MP -MF $(DEPDIR)/$*.Td
 
 INCLUDES := \
 -I. \
 -I$(MAKEFILE_DIR)/downloads/ \
+-I$(MAKEFILE_DIR)/downloads/gemmlowp/ \
```
cwhipkey: for this change, are there instructions needed elsewhere for downloading gemmlowp, or do those already exist.

tensorflow/core/kernels/conv_ops_using_gemm.cc
```diff
@@ -262,14 +262,15 @@ class Im2ColConvFunctor {
                 errors::InvalidArgument("Im2Col patch too large for buffer"));
     const size_t patches_per_chunk =
         max_chunk_size / (filter_value_count * sizeof(T1));
+    const size_t chunk_value_count = max_chunk_size / sizeof(T1);
```
cwhipkey: why this change?

is max_chunk_size definitely a multiple of sizeof(T1) here?

cwhipkey (changes requested):

tensorflow/contrib/ios_examples/README.md
petewarden: Thanks, I've updated all of the directions to refer to inception5h rather than just inception.

tensorflow/contrib/ios_examples/camera/CameraExampleViewController.mm
petewarden: I believe constant variables have internal linkage (like file-level static) by default[1], so it's prefered to avoid prefixing them with static. The NSStrings are defined that way because they're not const, mostly because they're based on old-school C constructs through ObjectiveC, so I explicitly make them static to avoid possible linker conflicts. 

[1] http://stackoverflow.com/questions/6173872/why-is-multiple-definition-of-a-const-global-variable-allowed-in-c-and-not-in

tensorflow/contrib/ios_examples/camera/tensorflow_utils.mm
petewarden: This only happens if it's used in a subsequent expression, so in general it's doing the right thing. It would be good to have an option to disable this optimization just for subgraphs that have ImmutableConsts as their source though. I'd also like to get constant folding into optimize_graph as an offline process too.

tensorflow/contrib/ios_examples/simple/tf_ios_makefile_example.xcodeproj/project.pbxproj
petewarden: > For debug_information_level, is the default for Release essentially 'none'?

Yes, that's correct.

> exceptions and rtti are not needed by tensorflow and at least rtti adds code size .
> (although maybe rtti is needed if we are not using proto lite runtime here).

I've turned off exceptions (though not rtti) for the protobuf and tensorflow library builds. I found I hit problems when trying it with the actual app though, but I'm guessing since it's only a small amount of code it wouldn't make a massive difference. Disabling exceptions on TF and protobuf did help a lot though.

Thanks for the comment though, I'd forgotten that RTTI was separate from exceptions, so I can try disabling that too!

tensorflow/contrib/makefile/Makefile
petewarden: We actually already download gemmlowp, but this change shouldn't be checked in since it's only necessary when we're running quantized models, so I've removed it.

tensorflow/core/kernels/conv_ops_using_gemm.cc
petewarden: This is a bug fix because I was actually using max_chunk_size (which is in bytes) as the number of _elements_ to allocate below. This led to a 64MB buffer being allocated for 16m floats, rather than the 16MB buffer with 4m floats that I was expecting.

Good point on the size calculation! I've updated it to effective do a ceil on the integer calculation, to make sure we allocate enough space even for odd-sized types.

tensorflow/contrib/ios_examples/simple/tf_ios_makefile_example.xcodeproj/project.pbxproj
petewarden: As an update, I tried disabling RTTI and the full protobuf failed with multiple errors, as did TensorFlow when I tried to compile against it. It's still worth investigating since this could be a big size win, but I'll leave that for a later change.

petewarden:
Jenkins, test this please.

cwhipkey (approved):

PR #4464, opened 2016-09-19
[mnist]Fix the array type bug

DjangoPeng:
Fix the `PS_ARRAY` and `WORKER_ARRAY`  bug. They should be arrays, but in fact they are strings. The problem is the wrong way for assigning bash array. 
The log info before fixing:

```
2 worker process(es) running in parallel...
Worker 0: 
  WORKER HOST: localhost:2223 localhost:2224 
  log file: /tmp/worker0.log
Worker 1: 
  WORKER HOST: 
  log file: /tmp/worker1.log
```

After:

```
2 worker process(es) running in parallel...
Worker 0: 
  WORKER HOST: localhost:2223
  log file: /tmp/worker0.log
Worker 1: 
  WORKER HOST: localhost:2224
  log file: /tmp/worker1.log
```

I made a mistake on my branch, so I have to open a new PR to fix the bug. 
@caisq  Plz check


caisq:

caisq (approved):
LGTM

PR #4472, opened 2016-09-19
Revised CMake rules for SWIG wrapper generation.

mrry:
Instead of generating a separate `python_deps.so` and `_pywrap_tensorflow.so`, we now generate a single `_pywrap_tensorflow.so` containing the entire runtime, like the Bazel build. This simplifies the linkage on other platforms.

### DIFF
--- a/tensorflow/contrib/cmake/setup.py
+++ b/tensorflow/contrib/cmake/setup.py
@@ -177,7 +177,6 @@ setup(
     include_package_data=True,
     package_data={
         'tensorflow': ['python/_pywrap_tensorflow.so',
-                       'python/libpython_deps.so',
                      ] + matches,
     },
     zip_safe=False,
--- a/tensorflow/contrib/cmake/tf_python.cmake
+++ b/tensorflow/contrib/cmake/tf_python.cmake
@@ -12,11 +12,7 @@
 # Resolve installed dependencies
 ########################################################
 
-# 1. Resolve the installed version of SWIG.
-FIND_PACKAGE(SWIG REQUIRED)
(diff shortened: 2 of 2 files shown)


mrry:

mrry:
@gunan PS. Is there some way to configure the presubmits to only run the CMake build if the changes are all under `tensorflow/contrib/cmake`? Sorry that this wastes so much time on the Jenkins cluster!

mrry:
@gunan - I think the workspace for the CMake build might have some leftover files in it that the ci_build doesn't have permission to delete. Is that possible?

gunan:
I think cmake build is configured to do exactly that, only run if 1-everything else passes, 2-it changes something in cmake folder. 
I think what we want to do is just not wait for other tests, run anytime there is a change in cmake folder.

For the 2nd question. It runs inside a docker container. Inside that container, it runs as root. 
Ddid you see some unexpected behaviour? I can check if there is a permissions issue on the cmake test.

gunan:
Jenkins, test this please.

mrry:
AFAIK, the build [start by deleting any leftover build directory](http://stackoverflow.com/questions/39582974/regarding-the-tensor-shape-is-1) and this seems to be failing with `Permission denied` errors. Do we deprivilege before running a build? (Is that what [`with_the_same_user`](https://github.com/tensorflow/tensorflow/blob/56eb8376260ef0ddb6389920dabc7f636d21cbff/tensorflow/tools/ci_build/builds/with_the_same_user) is meant to do?)

gunan (approved):

PR #4488, opened 2016-09-20
Fuse MirrorPadding into Conv2D to reduce memory usage and decrease latency

petewarden:
This is similar to the work for fusing bilinear filtering into Conv2D, but this version supports nearest neighbor lookups as part of im2col, so that mirror padding without a bilinear resize present can be efficiently fused.


martinwicke:
Why PR?

petewarden:
> Why PR?

Because this is aimed at iOS, so it's more straightforward to test changes directly in github. I won't make this into a habit, and I've asked cwhipkey@ if he minds the extra reviewing burden.

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -46,9 +46,15 @@ namespace {
...
+enum SamplingMode {
+  BILINEAR = 0,
+  NEAREST = 1,
+};
+
 // Combines bilinear resizing and mirror padding into the im2col transformation
-// stage of convolution,
-template <class T1, class T2, class T3, class TGemmFunctor>
+// stage of convolution.
+template <class T1, class T2, class T3, class TGemmFunctor, int SampleMode>
```
cwhipkey: can       int SampleMode      be             SamplingMode SampleMode    ?

tensorflow/core/kernels/conv_ops_fused.cc
```diff
@@ -183,18 +192,24 @@ class FusedResizeAndPadConvFunctor {
...
-                  const T1 bottom_left(input_data(batch, bottom_y_index,
-                                                  left_x_index, in_channel));
-                  const T1 bottom_right(input_data(batch, bottom_y_index,
-                                                   right_x_index, in_channel));
-                  const T1 top = top_left + (top_right - top_left) * x_lerp;
-                  const T1 bottom =
-                      bottom_left + (bottom_right - bottom_left) * x_lerp;
-                  in_value = top + (bottom - top) * y_lerp;
+                  if (SampleMode == NEAREST) {
+                    const T1 top_left(input_data(batch, top_y_index,
```
cwhipkey: NEAREST doesn't seem like the right name here, since it always takes the top-left one.

perhaps comment (or check) somewhere that DoResize and SampleMode need to be in sync?  Seems like using DoResize==true with SampleMode=NEAREST would be wrong.   Or maybe just use DoResize for this choice here, and remove SampleMode for now?

tensorflow/core/kernels/conv_ops_test.cc
```diff
@@ -173,6 +173,61 @@ class FusedResizePadConvOpTest : public OpsTestBase {
...
+                                      int input_depth, int y_padding,
+                                      int x_padding, int filter_size,
+                                      int filter_count, string pad_mode,
+                                      int stride, string padding) {
+    auto root = tensorflow::Scope::NewRootScope();
+    using namespace ::tensorflow::ops;  // NOLINT(build/namespaces)
+
+    const size_t input_data_size = input_height * input_width * input_depth;
+    Tensor input_data(DT_FLOAT,
+                      TensorShape({1, input_height, input_width, input_depth}));
```
cwhipkey: you could use FillIota from tensor_testutil.h instead (if filling 0, ...  is okay).

cwhipkey: (and same below on line 199)

tensorflow/core/ops/nn_ops.cc
```diff
@@ -499,6 +499,40 @@ strides: 1-D of length 4.  The stride of the sliding window for each dimension
...
+    .Attr(GetMirrorPadModeAttrString())
+    .Attr("strides: list(int)")
+    .Attr(GetPaddingAttrString())
+    .Doc(R"doc(
+Performs a padding as a preprocess during a convolution.
+
+Similar to FusedResizeAndPadConv2d, this op allows for an optimized
+implementation where the spatial padding transformation stage is fused with the
+im2col lookup, but in this case without the bilinear filtering required for
+resizing. This prevents the need to write out the intermediate results as whole
```
cwhipkey: the "This" in "This prevents...".  Maybe say "Fusing the padding prevents ..."

tensorflow/core/ops/nn_ops.cc
```diff
@@ -499,6 +499,40 @@ strides: 1-D of length 4.  The stride of the sliding window for each dimension
...
+tensors, reducing memory pressure, and we can get some latency gains by merging
+the transformation calculations.
+The data_format attribute for Conv2D isn't supported by this op, and defaults to
```
cwhipkey: default to ...   maybe say  'NHWC' order is used.   instead.

tensorflow/python/ops/nn_ops.py
```diff
@@ -945,6 +945,61 @@ def _FusedResizeAndPadConv2DShape(op):
   return [tensor_shape.TensorShape(output_shape)]
 
 
+@ops.RegisterShape("FusedPadConv2D")
+def _FusedPadConv2DShape(op):
```
cwhipkey: can this use the same function as the other fused one?  This will make it easier when porting to C++ (which will happen soon...)

tensorflow/python/tools/optimize_for_inference_lib.py
```diff
@@ -361,7 +361,15 @@ def fuse_resize_and_conv(input_graph_def):
     else:
       raise ValueError("Duplicate node names detected for ", node.name)
 
-  nodes_to_skip = {}
+  node_reference_count = {}
```
cwhipkey: if you use a collections.defaultdict  then you don't need to check for existence in the loop - just do  node_reference_count[stripped_name] += 1

tensorflow/python/tools/optimize_for_inference_lib.py
```diff
@@ -405,19 +424,24 @@ def fuse_resize_and_conv(input_graph_def):
...
-        resize_op.attr["align_corners"])
     fused_conv_op.attr["mode"].CopyFrom(mirror_paddings_mode)
     fused_conv_op.attr["strides"].CopyFrom(conv_op.attr["strides"])
     fused_conv_op.attr["padding"].CopyFrom(conv_op.attr["padding"])
     new_ops.extend([fused_conv_op])
 
   result_graph_def = tf.GraphDef()
   for node in input_graph_def.node:
-    if node.name in nodes_to_skip:
+    if (node.name in node_reference_count.keys() and
```
cwhipkey: can remove first condition if using a defaultdict.

cwhipkey (changes requested):

tensorflow/python/tools/optimize_for_inference_lib.py
petewarden: Thanks, that is cleaner!

tensorflow/python/tools/optimize_for_inference_lib.py
petewarden: It did change the logic a bit, because output nodes were never mentioned in the inputs of other nodes (so had no key) but we wanted to keep them. I like the cleanliness of this approach though, so I switched to explicitly passing in the names of the outputs.

tensorflow/python/ops/nn_ops.py
petewarden: That's a good idea, I've merged them together.

petewarden (commented):
Thanks for the review! I've responded to all the comments, and addressed most of them I think.

cwhipkey (approved):

PR #4489, opened 2016-09-20
Fix gif build on Windows

meteorcloudy:
It fails on Windows because "'unistd.h': No such file or directory". "unistd.h" is included in some source files but never used. I can manually comment out these includes then the target is buildable. Since it's a new_http_archive, the only fix I can think of is to write a genrule to remove those include statments
Ideally, we want the include could be configurable, but that's not true in the source code yet. @mrry @damienmg @dslomov

### DIFF
--- a/gif.BUILD
+++ b/gif.BUILD
@@ -15,13 +15,49 @@ HEADERS = [
     "gif_lib_private.h",
 ]
 
+config_setting(
+        name = "windows",
+        values = {
+            "cpu": "x64_windows_msvc",
+        },
+        visibility = ["//visibility:public"],
+)
+
 prefix_dir = "giflib-5.1.4/lib"
+prefix_dir_windows = "windows/giflib-5.1.4/lib"
+
+genrule(
+  name = "srcs_without_unistd",
+  srcs = [prefix_dir + "/" + source for source in SOURCES],
+  outs = [prefix_dir_windows + "/" + source for source in SOURCES],
+  cmd = "for f in $(SRCS); do " +
+        "  sed 's/#include <unistd.h>//g' $$f > $(@D)/%s/$$(basename $$f);" % prefix_dir_windows +
+        "done",
+)
+
+genrule(
+  name = "hdrs_without_unistd",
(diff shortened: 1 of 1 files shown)


gif.BUILD
```diff
@@ -15,13 +15,49 @@ HEADERS = [
...
+
 prefix_dir = "giflib-5.1.4/lib"
+prefix_dir_windows = "windows/giflib-5.1.4/lib"
+
+genrule(
+  name = "srcs_without_unistd",
+  srcs = [prefix_dir + "/" + source for source in SOURCES],
+  outs = [prefix_dir_windows + "/" + source for source in SOURCES],
+  cmd = "for f in $(SRCS); do " +
+        "  sed 's/#include <unistd.h>//g' $$f > $(@D)/%s/$$(basename $$f);" % prefix_dir_windows +
```
mrry: Is `sed` installed by default in MSYS, or would we have to add it to the dependencies?

mrry:

gif.BUILD
meteorcloudy: Yes, it is installed by default.

martinwicke: Are we relying on msys, implicitly or explicitly?

mrry: @martinwicke: IIUC running Bazel on Windows currently requires msys, but the generated binaries would not depend on it (since they're compiled using MSVC).

damienmg:
Iirc we still build bazel with msys. I am sure we still needs msys to run.

mrry (approved):
LGTM, despite the misgivings about relying on `sed`, though it would be great to find a cross-platform source for GIFLIB to avoid having to do this hack. (Unfortunately the top link I found for "GIFLIB for Windows" is 9 years and one major version out of date.)

martinwicke:
It's interesting since ​giflib (allegedly) is included in major browsers,
who was we know do work on Windows. GIF hasn't changed much so maybe a 9
year old source isn't so bad?

gif.BUILD
meteorcloudy: @mrry Yes, that's right. And we intend to remove msys dependency from Bazel as well the next quarter. :)

PR #4501, opened 2016-09-21
Updating CallWithRetries to do exponential backoff

moontails:
- setting maximum backoff seconds to be 32 seconds.


tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -24,6 +24,8 @@ namespace {
 
 // In case of failure, every call will be retried kMaxAttempts-1 times.
 constexpr int kMaxAttempts = 4;
+// Maximum backoff time in seconds
+constexpr int maximumBackoffSeconds = 32;
```
rinugun: Could you please rename this to kMaximumBackoffSeconds in order to be consistent with the naming conventions for constants?

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -39,14 +41,18 @@ bool IsRetriable(Status status) {
 
 Status CallWithRetries(const std::function<Status()>& f) {
   int attempts = 0;
+  int delay;
```
rinugun: Since the variable 'delay' is not used outside of the 'while' loop, could you please move it inside the while loop?

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -39,14 +41,18 @@ bool IsRetriable(Status status) {
...
   while (true) {
+    delay = 1 << attempts;
     attempts++;
     auto status = f();
     if (!IsRetriable(status) || attempts >= kMaxAttempts) {
       return status;
     }
     LOG(ERROR) << "The operation resulted in an error and will be retried: "
                << status.ToString();
+    delay = delay > maximumBackoffSeconds ? maximumBackoffSeconds : delay;
```
rinugun: I wonder if this line and the line 46 could be merged also using std::min, e.g
int delay = std::min(1 << attempts, kMaximumBackoffSeconds) ? This could result in shorter code.

rinugun (commented):
Hi! Thanks for improving this.

This is the first time I'm doing a code review on GitHub, so apologies if I'm doing something wrong.

One high level comment: retrying_file_system_test.cc contains test cases for the scenarios when all retry attempts failed. With this change, those tests will take more than 30 seconds to run and might actually time out, depending on what timeout is set for unittests in the test framework (I can imagine it being exactly 30 seconds).

I'd suggest adding an optional constructor parameter initial_delay_seconds, which defaults to 1 but can be overridden to 0 in the unittest, so that the unittest can complete faster.

This code would that use 
initial_delay_seconds << attempts
instead of hardcoding '1' as a constant for the initial delay.

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: I wanted to reuse the delay variable in each loop instead of creating a new one every time.

Should I still go ahead and make the change?

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Good idea!

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Did not know that. Will do.

tensorflow/core/platform/cloud/retrying_file_system.cc
martinwicke: The compiler will take care of optimization if it is beneficial. Please do make the change.

moontails:
@rinugun Thank you for reviewing my PR.

Parameterizing the intial_delay_seconds seems like a good approach, I have addressed the feedback and pushed the changes. Please take a look.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -24,6 +24,8 @@ namespace {
 
 // In case of failure, every call will be retried kMaxAttempts-1 times.
 constexpr int kMaxAttempts = 4;
+// Maximum backoff time in seconds
```
rinugun: A minor comment: please add a period at the end for consistency.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -40,20 +42,22 @@ bool IsRetriable(Status status) {
 Status CallWithRetries(const std::function<Status()>& f) {
   int attempts = 0;
   while (true) {
+    int delay = std::min(initial_delay_seconds << attempts, kMaximumBackoffSeconds);
```
rinugun: Minor comments: 
1) could you please add a 'const' modifier?
It's recommended by the style guide, see https://google.github.io/styleguide/cppguide.html#Use_of_const for details. 

2) Could you also move this line to right before line 53, since 'delay' is not used anywhere besides the last line in the loop. This should make the code more readable.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -40,20 +42,22 @@ bool IsRetriable(Status status) {
...
                << status.ToString();
+    sleep(delay);
   }
 }
 
 class RetryingRandomAccessFile : public RandomAccessFile {
  public:
-  RetryingRandomAccessFile(std::unique_ptr<RandomAccessFile> base_file)
-      : base_file_(std::move(base_file)) {}
+  RetryingRandomAccessFile(std::unique_ptr<RandomAccessFile> base_file, int delay_seconds = 1)
```
rinugun: Minor: could you please split the line to keep the line length under 80 chars?
See https://google.github.io/styleguide/cppguide.html#Line_Length for details

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -86,6 +91,7 @@ class RetryingWritableFile : public WritableFile {
 
  private:
   std::unique_ptr<WritableFile> base_file_;
+  int initial_delay_seconds;
```
rinugun: Please add a trailing underscore to the name, see this section for details
https://google.github.io/styleguide/cppguide.html#Variable_Names

Please also add a 'const' modifier, as this member is not supposed to be changed in the lifetime of the object.

Also applies to other places where you're adding the new member.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -40,20 +42,22 @@ bool IsRetriable(Status status) {
 Status CallWithRetries(const std::function<Status()>& f) {
```
rinugun: Did you want to add 'initial_delay_seconds' as a parameter to pass it to CallWithRetries? Otherwise it's not clear to me how initial_delay_seconds inside CallWithRetries is available, or am I missing something?

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -96,7 +102,7 @@ Status RetryingFileSystem::NewRandomAccessFile(
   TF_RETURN_IF_ERROR(CallWithRetries(std::bind(&FileSystem::NewRandomAccessFile,
                                                base_file_system_.get(),
                                                filename, &base_file)));
-  result->reset(new RetryingRandomAccessFile(std::move(base_file)));
+  result->reset(new RetryingRandomAccessFile(std::move(base_file), 0));
```
rinugun: It looks like if you pass zero for the initial_delay, it will effectively disable exponential backoff. Did you want to pass initial_delay_seconds_ which is a member of RetryingFileSystem?

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Should `initial_delay_seconds` be made a optional arg to `CallWithRetries` with a default of 1? (Rather than it being a member variable and optional constructor parameter)

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Regarding 
1. I will add the `constexpr` for delay.

2. I originally added this line here as attempts is incremented at the beginning of the loop. Adding near line 53 would have a slightly uglier way of doing this -
`int delay = std::min(initial_delay_seconds << attempts - 1, kMaximumBackoffSeconds);`

Do you propose to make this change? Or should we increment at the end of the loop and initialize `kMaxAttempts` to 3?
```
constexpr int kMaxAttempts = 3;
...

Status CallWithRetries(const std::function<Status()>& f) {
  int attempts = 0;
  while (true) {
    auto status = f();
    if (!IsRetriable(status) || attempts >= kMaxAttempts) {
      return status;
    }
    LOG(ERROR) << "The operation resulted in an error and will be retried: "
               << status.ToString();
    constexpr int delay = std::min(initial_delay_seconds << attempts, kMaximumBackoffSeconds);
    attempts++;
    sleep(delay);
  }
}
```

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: My understanding was that you wanted to pass zero for the initial_delay so that the unit tests could run faster?

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Will do

martinwicke:
assigning @rinugun to reflect reality. Thanks for reviewing!

tensorflow/core/platform/cloud/retrying_file_system.cc
rinugun: I don't think CallWithRetries is called from the unittest directly, so how would you make sure that initial_delay_seconds is set to 0 for tests?

tensorflow/core/platform/cloud/retrying_file_system.cc
rinugun: 1. Will 'constexpr' compile though? I thought 'constexpr' has to be computable at compile time. 
2. I think it's fine to move 'attempts++' to make it the last line in the loop (i.e it could be after sleep())
3. In that case its semantics changes from "the number of made attempts" to "the current attempt number", so I suggest renaming 'attempts' to 'attempt' to reflect that
4. Also, 'attempts >= kMaxAttempts" in the 'if' line must be replaced with "attempt > kMaxAttempt" to keep the total amount of allowed attempts the same.

tensorflow/core/platform/cloud/retrying_file_system.cc
rinugun: Yes, but zeros should only be passed from the unittest code (as you have implemented in the _test.cc file), whereas the non-test code should still use the actual non-zero values to  have the backoff logic working.

moontails: I see what you mean. Should have updated only in tests. Accidentally updated here as well. My bad. Will revert these changes.

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: 1. That is correct. Will use`const`

2, 3, 4. Agreed.

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: makes sense to keep it as optional constructor parameter

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Added as a parameter to pass it to CallWithRetries

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -23,7 +23,9 @@ namespace tensorflow {
 namespace {
 
 // In case of failure, every call will be retried kMaxAttempts-1 times.
-constexpr int kMaxAttempts = 4;
+constexpr int kMaxAttempts = 3;
```
rinugun: Is the change from 4 to 3 intentional?

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +39,70 @@ bool IsRetriable(Status status) {
...
-Status CallWithRetries(const std::function<Status()>& f) {
-  int attempts = 0;
+Status CallWithRetries(const std::function<Status()>& f,
+                       const initial_delay_seconds) {
+  int attempt = 0;
   while (true) {
-    attempts++;
     auto status = f();
-    if (!IsRetriable(status) || attempts >= kMaxAttempts) {
+    if (!IsRetriable(status) || attempt > kMaxAttempts) {
```
rinugun: Sorry, I think I gave you a wrong suggestion. Actually the previous >= condition was more correct here.
So if the attempt number is 1, this means we're making the second attempt. So if kMaxAttempts was 1 and attempt == kMaxAttempt, this means we should fail.

rinugun (commented):
Looks great, thanks for making all the fixes -- just one minor remaining thing re the kMaxAttempt constant and the way attempts are counted -- sorry for giving you a misleading comment in the past.

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Is there a preference between `>=` or `>=` ?

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: `attempt++` happens after the check now. Hence this has been changed from 4 to 3. I need to update the comment however has now it retries exactly `kMaxAttempts` times (if the comparison check is changed to a `>=` or a `==`)

tensorflow/core/platform/cloud/retrying_file_system.cc
rinugun: I think >= feels safer, even though in practice there should be no difference. So I'd prefer >=

moontails:
@rinugun Thank you for reviewing my code quickly each time. Learned a few things along the way while helping enhance this :)

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -22,8 +22,10 @@ namespace tensorflow {
 
 namespace {
 
-// In case of failure, every call will be retried kMaxAttempts-1 times.
-constexpr int kMaxAttempts = 4;
+// In case of failure, every call will be retried kMaxAttempts times.
+constexpr int kMaxAttempts = 3;
```
rinugun: Ok, I think I got it. It looks like kMaxAttempts has become a "retry count" instead of an "attempt count" that it used to be. So this needs to be fixed to be made consistent again. 

One way would be to rename it to kMaxRetries and rename 'attempt' to 'retry' respectively, keeping the constant value '3', the comment and the rest of the code as it is now.

Another, maybe a simpler way, would be to replace "attempt >= kMaxAttempts" with "attempt >= kMaxAttempts -1" to solve the "off by one" error.  And then the lines 25-26 can be reverted back to the previous state.

What do you think?

moontails: I see how this can be an issue due to the how the variable is named. 

Personally I prefer the first approach. Although it involves more code change, it is easier to read as retry count makes everything slightly more readable.

rinugun: Sounds good to me, thanks!

rinugun: Actually then 'attempt' should become 'retries' (not 'retry') and will represent the amount of retries made so far, so it will make sense that it is  0 in the beginning.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -41,18 +41,18 @@ bool IsRetriable(Status status) {
 
 Status CallWithRetries(const std::function<Status()>& f,
                        const initial_delay_seconds) {
-  int attempt = 0;
+  int retry = 0;
```
rinugun: Just one last comment -- suggest renaming to 'retries', since it's going to be a retry count now.

moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Agreed. `retries` is just more readable

rinugun (approved):
Great, thanks for bearing with me!

Let me see if I can figure out how to make your changes get submitted to the repository, I've never done this before.

moontails:
@rinugun the pleasure was mine! Your feedback helped polish the code I had submitted. Really appreciate your quick feedback on each commit :) Thank you!

rinugun:
Actually I'm not authorized to merge pull requests, so hopefully something good will happen with this PR automatically now when I approved. If not, please ping this thread again, I'll try to investigate.

moontails:
I will keep that in mind. Thank you once again @rinugun :)

martinwicke:
Jenkins, test this please (@rinugun, you can trigger tests as well)

moontails:
Jenkins test this please

martinwicke:
Jenkins, test this please

moontails:
Error details -
```
tensorflow/core/platform/cloud/retrying_file_system.cc: In member function 'virtual tensorflow::Status tensorflow::RetryingFileSystem::IsDirectory(const string&)':
tensorflow/core/platform/cloud/retrying_file_system.cc:200:76: error: too few arguments to function 'tensorflow::Status tensorflow::{anonymous}::CallWithRetries(const std::function<tensorflow::Status()>&, int)'
       std::bind(&FileSystem::IsDirectory, base_file_system_.get(), dirname));
                                                                            ^
tensorflow/core/platform/cloud/retrying_file_system.cc:42:8: note: declared here
 Status CallWithRetries(const std::function<Status()>& f,
        ^
tensorflow/core/platform/cloud/retrying_file_system.cc:201:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

I checked and see that the `IsDirectory` method is not overridden. But I fail to see why this causes an error as the base class implementation does not wrap any call with `CallWithRetries`.

martinwicke:
Not quite yet. :)

rinugun:
@moontails, I think your local version of the repository is outdated, the function is overridden in the latest version on GitHub https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/cloud/retrying_file_system.cc#L170

moontails:
@rinugun I will sync with master and make the changes.

rinugun:
Jenkins, test this please

moontails:
@rinugun @martinwicke the tests werent trigerred

martinwicke:
Jenkins, test this please?

moontails:
looks like the failed jobs lost connection?

jhseu:
Jenkins, test this please

jhseu:
Hmm, that failure looks unrelated. Can you try rebasing on top of the latest master and we'll retest?

moontails:
Sure will do. Do you want me to rebase or merge with latest master?

Best Regards,
Balachander Ramachandran

jhseu:
Merge is fine too (and probably easier). We have some changes in master that are needed for tests to work.

rinugun:
@moontails, I just realized that sleep() might not be the canonical way of making a delay, e.g it might not be supported on Windows. I was wondering if you could replace it with 
Env::Default()->SleepForMilliseconds, e.g see this example:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/cloud/gcs_file_system.cc#L379

This also prepares the code for the subsequent implementation of random millisecond part, that we discussed in another thread with @davidj-github.

Also I had another comment related to this, even though this contradicts to what I suggested before. Instead of making 'initial_delay' a ctor parameter, you could have 'Env* env' a ctor parameter instead, which would default to Env::Default(). This way you could provide a fake implementation of SleepForMilliseconds from the test code. The fake implementation would return immediately, but would record how it was called.
The benefit of the latter is that the test code can still work fast, but also check the sleep intervals that were passed to SleepForMilliseconds from the main code and make sure they all look as expected.

moontails:
@jhseu I have merged the latest changes from master. Could you please trigger the tests?

jhseu:
Sure, but mind making the requested changes before we commit?
Jenkins, test this please

moontails:
@jhseu @rinugun committed the changes. Please review and trigger tests!

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -14,16 +14,21 @@ limitations under the License.
...
 
 namespace tensorflow {
 
 namespace {
 
-// In case of failure, every call will be retried kMaxAttempts-1 times.
-constexpr int kMaxAttempts = 4;
+// In case of failure, every call will be retried kMaxRetries times.
+constexpr int kMaxRetries = 3;
+// Maximum backoff time in seconds.
```
rinugun: Please update the comment.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -14,16 +14,21 @@ limitations under the License.
...
+constexpr int kMaximumBackoffMilliSeconds = 32000;
```
rinugun: A minor thing, but "millisecond" is one word, so it should be kMaximumBackoffMilliseconds (small "s")

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +42,81 @@ bool IsRetriable(Status status) {
...
-Status CallWithRetries(const std::function<Status()>& f) {
-  int attempts = 0;
+void WaitBeforeRetry(const int delay_seconds) {
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_int_distribution<> dist(0, 999);
+
+  const int delay_milliseconds = delay_seconds * 1000;
+  const int random_milliseconds = dist(gen);
+  const int delay = std::min(delay_milliseconds + random_milliseconds,
```
rinugun: Another minor comment: since 'delay' here in fact also means "delay in milliseconds", this now conflicts with the variable at line 50. To resolve this, I suggest renaming 'delay' to 'delay_milliseconds' and inlining the previous 'delay_milliseconds', e.g

const int64 delay = std::min(delay_seconds * 1000 + random_milliseconds)

tensorflow/core/platform/cloud/retrying_file_system.cc
rinugun: Also, please update to int64 to be on the safe side.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +42,81 @@ bool IsRetriable(Status status) {
...
 }
 
-Status CallWithRetries(const std::function<Status()>& f) {
-  int attempts = 0;
+void WaitBeforeRetry(const int delay_seconds) {
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_int_distribution<> dist(0, 999);
+
+  const int delay_milliseconds = delay_seconds * 1000;
```
rinugun: Could you update the types to  int64 to be sure, also the next line.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +42,81 @@ bool IsRetriable(Status status) {
...
+                       const int initial_delay_seconds) {
+  int retries = 0;
   while (true) {
-    attempts++;
     auto status = f();
-    if (!IsRetriable(status) || attempts >= kMaxAttempts) {
+    if (!IsRetriable(status) || retries >= kMaxRetries) {
       return status;
     }
     LOG(ERROR) << "The operation resulted in an error and will be retried: "
```
rinugun: I wonder if you could also print the delay in seconds in the LOG message, e.g 
const int delay_seconds = initial_delay_seconds << retries;
... will be retried after " << delay_seconds << " sec:"

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
rinugun: Sorry, what I meant was
const int64 delay_milliseconds = std::min(delay_seconds * 1000 + random_milliseconds,
                                                                      kMaximumBackoffMilliseconds);

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Yup got it, Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

rinugun:
Sorry for not sending all comments at once -- hit a wrong button on Github.
Please see my other comments as well.

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

moontails:
@rinugun I have addressed all your comments

rinugun (approved):
Thanks, looks great!

moontails:
@jhseu @martinwicke could one of you please trigger the tests

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +42,81 @@ bool IsRetriable(Status status) {
...
+void WaitBeforeRetry(const int delay_seconds) {
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_int_distribution<> dist(0, 999);
+
+  const int64 random_ms = dist(gen);
+  const int64 delay_ms = std::min(delay_seconds * 1000 + random_ms,
+                                  kMaximumBackoffMilliseconds);
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(delay_ms));
```
jhseu: Please use Env::SleepForMicroseconds instead:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/env.h#L220

There are internal reasons why we need to use a non-standard sleep call.

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +42,81 @@ bool IsRetriable(Status status) {
   }
 }
 
-Status CallWithRetries(const std::function<Status()>& f) {
-  int attempts = 0;
+void WaitBeforeRetry(const int delay_seconds) {
+  std::random_device rd;
+  std::mt19937 gen(rd());
```
jhseu: Please call Rand64() instead:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/random/random.h

It uses a static device, so won't reconstruct every call. Something like `const int64 random_ms = Rand64() % 1000`

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +42,81 @@ bool IsRetriable(Status status) {
...
+
+  const int64 random_ms = dist(gen);
+  const int64 delay_ms = std::min(delay_seconds * 1000 + random_ms,
+                                  kMaximumBackoffMilliseconds);
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(delay_ms));
+}
+
+Status CallWithRetries(const std::function<Status()>& f,
+                       const int initial_delay_seconds) {
```
jhseu: Can you pass around microseconds for consistency with the rest of tensorflow?

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -133,49 +168,58 @@ bool RetryingFileSystem::FileExists(const string& fname) {
...
+      std::bind(&FileSystem::Stat, base_file_system_.get(), fname, stat),
+      initial_delay_seconds_);
 }
 
 Status RetryingFileSystem::GetChildren(const string& dir,
                                        std::vector<string>* result) {
   return CallWithRetries(std::bind(&FileSystem::GetChildren,
-                                   base_file_system_.get(), dir, result));
+                                   base_file_system_.get(), dir, result),
+                         initial_delay_seconds_);
```
jhseu: nit: here and below, please align

jhseu (commented):
Sorry, I'm a little pickier than @rinugun :)

A few more comments.

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Isn't the alignment right here? `std::bind(&FileSystem::GetChildren, base_file_system_.get(), dir, result)` is one parameter and `initial_delay_seconds_` is another.

Parameters to `std::bind` are aligned and parameters to `CallWithRetries` are algined

jhseu: Oh, you're right, ignore this comment.

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -14,16 +14,23 @@ limitations under the License.
 ==============================================================================*/
 
 #include "tensorflow/core/platform/cloud/retrying_file_system.h"
+#include <chrono>
```
jhseu: nit: remove chrono random and thread includes?

jhseu (changes requested):

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: my bad, should have cleaned up unused imports

tensorflow/core/platform/cloud/retrying_file_system.h
```diff
@@ -16,6 +16,8 @@ limitations under the License.
 #ifndef TENSORFLOW_CORE_PLATFORM_RETRYING_FILE_SYSTEM_H_
 #define TENSORFLOW_CORE_PLATFORM_RETRYING_FILE_SYSTEM_H_
 
+#include <unistd.h>
```
jhseu: Delete?

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -16,14 +16,18 @@ limitations under the License.
...
 
 namespace tensorflow {
 
 namespace {
 
-// In case of failure, every call will be retried kMaxAttempts-1 times.
-constexpr int kMaxAttempts = 4;
+// In case of failure, every call will be retried kMaxRetries times.
+constexpr int kMaxRetries = 3;
+// Maximum backoff time in milliseconds.
```
jhseu: milliseconds -> microseconds

tensorflow/core/platform/cloud/retrying_file_system.cc
```diff
@@ -37,55 +41,76 @@ bool IsRetriable(Status status) {
   }
 }
 
-Status CallWithRetries(const std::function<Status()>& f) {
-  int attempts = 0;
+void WaitBeforeRetry(const int64 delay_micros) {
+  const int64 random_micros = random::New64() % 1000000L;
```
jhseu: Here and below: L is not necessary for these constants (and isn't sufficient to make it an int64)

jhseu (commented):
Looks good. A few more nitpicks :)

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: so i should remove L ?

jhseu: Yes, please

moontails: quick question should i be removing it from every place i have used? (like `kMaximumBackoffMicroseconds` , constructor args default value etc... )

tensorflow/core/platform/cloud/retrying_file_system.h
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Done

tensorflow/core/platform/cloud/retrying_file_system.cc
moontails: Removed it everywhere I introduced it

jhseu (approved):

jhseu:
Jenkins, test this please

jhseu:
Jenkins, test this please

moontails:
@jhseu @rinugun thank you for all the feedback during this code review and responding to my questions so quickly each time :)

jhseu:
Thanks you for your contribution!

Issue #217, opened 2015-11-14
Is pep8 compatibility necessary right now for tensorflow?


taranjeet:
On running the pep8 test, I found out around 18k warnings. I think we should strict to some guideline, so that as the project grows, a proper standard is maintained. We can use flake8 for this.

I can work on this and submit a PR, if pep8 compatibility is required.

vrv:
Thanks for the report -- we'll look into this at some point!  Hopefully our code is at least internally consistent :)

staranjeet:
ok, then i work on this later on.

girving:
@martinwicke: Do we have a plan in terms of linting and/or pep8 compliance?  Should this be closed?

martinwicke:
Plans, yes, time to do it, no. I would like to make much more of the pylint
we're already running, and ideally add cpplint as well.

girving:
@martinwicke: I'll leave it assigned to you for now, and you can reassign to other testing folk as appropriate.

staranjeet:
I can work on this.

girving:
@staranjeet: I think we may need to figure out a policy of just how pep8 compliant we want to be first.  For example, I'd hate to give up our 2 space indent paradise.

martinwicke:
Definitely not while we still maintain a 80-character hard limit.

martinwicke:
@staranjeet We'd like to ignore some thing (in particular, the 4 space indent), but otherwise, proper linting would be great. We already run pylint as part of the build, but we disabled all but the most terrible errors (syntax errors, unknown names) in order to make it blocking without damaging productivity.

It would be nice to have a stricter pylintrc that we can use for informational purposes. Especially if we can generate statistics on how many lint issues were added (and make that a blocking error). 

I haven't used flake8 miself, if it supports both python 2 and python 3 and is sufficiently flexible we could use that as well.

martinwicke:
Closing this -- if there are specific violations, or a nicer linter for our tests, I'd love more concrete issues or PRs.

staranjeet:
I ran a pep8 test and found that most of the warnings were for spaces as 2
and 80 char limit. As for the spaces, I think that as per the above
conversation we are going with 2 spaces convention. Char limit can be
discussed.

Also let me know if there are any other linters. I was a little busy, will
also look myself today.

martinwicke:
[...]

PR #4511, opened 2016-09-21
adding flake8 config

esc:
If you are going to use 2 spaces instead of four, at least reconfigure the
linter to not complain.

:)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+ignore = E111


jhseu:
We're not using flake8 yet, and I don't think we want to merge this until we get a handle on the configuration we want to use for it. A related issue: https://github.com/tensorflow/tensorflow/issues/217

It'll otherwise give too many warnings besides the indentation that would make it pretty useless.

Note: we do run pylint with this config:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/pylintrc

esc:
ok, i'll see if I can make a user specific config work for me.

Issue #4359, opened 2016-09-13
Feature Request: plug-in support for new devices


vrv:
We need to be able to support using new devices in TensorFlow without requiring editing the TensorFlow binary source code.

This requires a few changes.  Among many others:
- The ability to dynamically register an implementation / factories for alternative devices
- New or better APIs for allowing OpKernel's to use arbitrary device resources (e.g., right now we assume either the use of StreamExecutor or EigenDevice implementations, but that's obviously not general).

aidan-plenert-macdonald:
Few questions.

## About Feature Request
By dynamically, I assume you are referring to at compile time, but self-contained code changes. If you mean at runtime via dynamic module loading, I don't know off hand how to do this, but I think we can figure it out.

How do you wish to handle the kernel code? Should I plan to have new hardware supported by developers modifying the `tensorflow/core/kernels/*` and registering the new hardware? Or do you want this also separated. Although this currently is handled with macros and registration systems in the C++, but I ask from a build perspective. We could make a new folder that contains all the code for extra devices and contains all the build scripts over there. It this is done, I can write template files that have all the basic structural code (many projects do this for configuration scripts) that a user could basically rename and be ready to go.

## About TensorFlow

[TensorFlow Device Contexts, Streams, and Context Switching](http://stackoverflow.com/questions/39481453/tensorflow-device-contexts-streams-and-context-switching)

vrv:
Yes, I do mean self-contained code changes: a user should be able to use an existing binary install of TensorFlow that doesn't contain the custom device code, but then call a function to "load a device" and then another function to "load the kernels" for that device.

The latter is already possible for existing devices via the tf.load_op_library() mechanism, so theoretically something similar could be done for a new tf.load_device().

I'll answer your other question on SO, but I don't think the answer there will be instructive for new devices.  Every device has its own execution model, and so the way the device code is written needs to take into account the execution model.  On CPU, the ThreadPoolDevice is just a device that uses Eigen's multi-threaded CPU support to implement operations.  On GPU, we have to manage streams and execution order carefully because GPU devices are asynchronous, not synchronous.

If you told us more about the execution model of your hardware, we might be able to suggest something more relevant -- it's likely that copying the CPU or the GPU way of doing things is not the right solution.

aidan-plenert-macdonald:
[...]

PR #4520, opened 2016-09-21
Develop device framework

aodhan-domhnaill:
Beginning of solution to [Issue #4359](https://github.com/tensorflow/tensorflow/issues/4359)


vrv:
I'm going to close this for now since we're making progress on the https://github.com/tensorflow/tensorflow/issues/4359.  Once we have everything in the framework to add a device, we can add the docs.  Thanks!

PR #4523, opened 2016-09-21
make text in word2vec tutorial match image

hrldcpr:
The embedded image has male-female, verb tense, and country-capital, so the text should too.


martinwicke (approved):

martinwicke:
Thanks!

PR #4541, opened 2016-09-23
Fix recently-introduced bug in ci_build.sh re. cmake builds

caisq:
Commit 4316aeb inadvertently made cmake builds depend on bazel. This PR removes this dependency.


vrv (approved):

caisq:
Thanks, @vrv

PR #4563, opened 2016-09-24
Fix to download eigen library

tyfkda:
URL for eigen library is extracted from `tensorflow/workspace.bzl`,
but expression is used to construct the URL in the code,
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl#L22
so extracted URL using simple `grep` is not correct.

This change get version code for eigen library from the file,
and construct URL with it.

It looks cmake handles the URL in same way:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/external/eigen.cmake#L13

### DIFF
--- a/tensorflow/contrib/makefile/download_dependencies.sh
+++ b/tensorflow/contrib/makefile/download_dependencies.sh
@@ -19,9 +19,12 @@ set -e
 DOWNLOADS_DIR=tensorflow/contrib/makefile/downloads
 BZL_FILE_PATH=tensorflow/workspace.bzl
 
-# Temporarily specify the Eigen URL manually while we're waiting on the build
-# file changes to be synced.
-EIGEN_URL=http://bitbucket.org/eigen/eigen/get/8106cca06137.tar.gz
+EIGEN_VERSION="$(sed -ne 's/^[ \t]*eigen_version = "\(.*\)".*$/\1/p' "${BZL_FILE_PATH}")"
+if [ "${EIGEN_VERSION}" == '' ]; then
+    echo "Cannot extract eigen_version from ${BZL_FILE_PATH}" >&2
+    exit 1
+fi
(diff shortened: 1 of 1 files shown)


tyfkda:
I find the change is done in the commit: 65038b084059cf934df50fa86dba5b0e765f9d65

tyfkda:
I have updated the commit
since conflict occurred with #4575 .

jart:
Thank you for your contribution. Ideally I would like to update this by having it pull the URL out of the workspace.bzl file the same way it gets pulled out of other files.

tyfkda:
@jart 
I'm not sure what is your expected way,

1.
Unlike URLs for other libraries in download_dependencies.sh (`GEMMLOWP_URL`, `GOOGLETEST_URL`, `PROTOBUF_URL`, `RE2_URL`), single `grep` cannot extract `EIGEN_URL`, because the URL is declared as

```py
    url = "http://bitbucket.org/eigen/eigen/get/" + eigen_version + ".tar.gz",
```

in [workspace.bzl](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl#L22).

2.
Unlike CMake, [string(REGEX MATCH ...)](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/external/eigen.cmake#L13) cannot use in Makefile.

3.
Do you mean it is preferable to modify the declaration for eigen library in same way in workspace.bzl, not to use varible?

```py
  native.new_http_archive(
    name = "eigen_archive",
    url = "http://bitbucket.org/eigen/eigen/get/46ee714e25d5.tar.gz",
    sha256 = "d2ba02303c20d6ddc1a922f7e0e176ef841514545e053388845359aa62176912",
    strip_prefix = "eigen-eigen-46ee714e25d5",
    build_file = str(Label("//:eigen.BUILD")),
  )
```

to enable to use single `grep` to extract eigen URL?

jhseu:
Jenkins, test this please

jhseu (approved):

PR #4564, opened 2016-09-24
Fix header path of iOS examples for eigen

tyfkda:
I failed to build iOS example projects.
It seems download destination path for eigen library is changed
from `tensorflow/contrib/makefile/downloads/eigen-latest`
to `tensorflow/contrib/makefile/downloads/eigen`.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/makefile/download_dependencies.sh#L49

Fixing `Header Search Paths` solves the problem.


petewarden:
I think these changes make sense to me, Jenkins test this please.

jart (approved):

PR #4569, opened 2016-09-25
fix Gfile method size() to Size()

mslovy:
Gfile object has method Size(), not size()

Signed-off-by: yaoning yaoning@unitedstack.com


vrv:
Are you sure?

Gfile now just wraps FileIO, which defines it as size(): https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/lib/io/file_io.py#L83

vrv:
It was Size() in branch r0.10.  If you want to send a PR for that branch, please do, after signing the CLA.  However, we since changed it to size() in r0.11 / master, so I'm going to close this.

https://github.com/tensorflow/tensorflow/blob/r0.10/tensorflow/python/platform/gfile.py#L179

PR #4575, opened 2016-09-25
Temporary workaround for iOS build problem

petewarden:
The download_dependencies.sh script fails, because the pattern it's searching for in workspace.bzl has changed but the workspace file itself hasn't been updated yet. This is a minimal fix to allow users to continue building for iOS until we get the syncing issue sorted out.

### DIFF
--- a/tensorflow/contrib/makefile/download_dependencies.sh
+++ b/tensorflow/contrib/makefile/download_dependencies.sh
@@ -19,7 +19,9 @@ set -e
 DOWNLOADS_DIR=tensorflow/contrib/makefile/downloads
 BZL_FILE_PATH=tensorflow/workspace.bzl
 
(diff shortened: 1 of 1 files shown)


jart (approved):
Ideally I'd like to re-fix workspace.bzl so eigen's version is specified in the url like all the other externals. But this could be a temporary fix.

Mr. Jenkins test this please.

jart:
By the way, are you aware that the Eigen version you specified in the download file is not the same as the one in the [workspace.bzl](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl) file?

jart:
test this please

gunan:
http://ci.tensorflow.org/job/tensorflow-master-makefile/
and this PR fixed the master makefile build.

PR #4580, opened 2016-09-26
Support older versions of git

Phhere:
The Parameter -C is unknown in git 1.7.1 but --exec-path= can be used.

### DIFF
--- a/tensorflow/tools/git/gen_git_source.py
+++ b/tensorflow/tools/git/gen_git_source.py
@@ -143,7 +143,7 @@ def get_git_version(git_base_path):
 
   unknown_label = "unknown"
   try:
-    val = subprocess.check_output(["git", "-C", git_base_path, "describe",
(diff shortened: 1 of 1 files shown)


jhseu:
Andrew, mind taking a look?

aselle:
Thank you for the patch.

You should squash your two commits into one using `git rebase -i` since the first commit is incorrect (it confused me at first.

Also, there is an internal committed change that will be sync'd soon that adds a 
`tensorflow/tools/git/gen_git_source.sh`. It would be great if you could make the change to that (which is used for the Makefile case to avoid the python dependency).

Phhere:
Ok, I will rebase it and wait for the internal change

Phhere:
I think the new version is ok

drpngx:
Jenkins, test this please.

drpngx:
@aselle , are we good to go?

drpngx:
(I don't think that the gen_git_source.sh in its present form requires modifications)

aselle (approved):

PR #4594, opened 2016-09-27
Update optimizers.py

gautam1858:


gautam1858:
I signed it !

Regards,
Gautam.R

jhseu:
Thanks for the change! We can't change the default yet because some ops may not be colocateable with the gradient (e.g., if an implementation doesn't exist on the device). Also, changing the default would require some analysis of performance impact.

PR #4608, opened 2016-09-28
optional arguments should be optional

pfaucon:
the input_op_fn appears to be optional (it has a default value), but the default value is trash and causes the function to crash with a less than useful argument.

I don't have the skill to fix the rnn.py function at the moment, but it seems to me that the null_input_op_fn should make some guess at how the data should be unwrapped rather than just return something that will cause a crash.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/estimators/rnn_test.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/rnn_test.py
@@ -133,5 +133,16 @@ class RNNTest(tf.test.TestCase):
     predictions = classifier.predict(data[:2])
     self.assertAllClose(predictions, labels[:2])
 
+def testNoInputFnRNN(self):
+    # Classification
+    classifier = tf.contrib.learn.TensorFlowRNNClassifier(rnn_size=2,
+                                                          cell_type="lstm",
+                                                          n_classes=2,
+                                                          steps=100)
+    classifier.fit(data, labels)
+    predictions = classifier.predict(data[:2])
+    self.assertAllClose(predictions, labels[:2])
+
+    
 if __name__ == "__main__":
   tf.test.main()


pfaucon:
I signed it!

jhseu:
Thanks for the change! This behavior is intentional. You can only avoid passing input_op_fn if your input is already in the right format (sequence) for the RNN.

pfaucon:
Could you clarify what you mean "if your input is already in the right format"?  I would be very happy to write a passing test case that does not use any of the optional methods if you show me the way.

I've tried making a new np.array containing a list (but not list of lists) and that fails.
I've tried passing a list and that fails (no dtype).

jhseu:
Ah, you're right that there's no way to work without input_op_fn. That's because RNN's expect an unpacked (python) list of tensors, but the fit function takes in an np.array.

A minimal example would be something like:
`classifier = tf.contrib.learn.TensorFlowRNNClassifier(rnn_size=2, cell_type="lstm", n_classes=2, steps=100, input_op_fn=lambda x: tf.unpack(x, axis=1))
classifier.fit(np.array([[[1, 2, 3], [4, 5, 6]], [[7, 8 , 9], [10, 11, 12]]], dtype=np.float32), np.array([0, 1], dtype=np.float32))`

Keep in mind that we can't merge in failing tests because our test suite ensures that every commit passes all tests before merging pull requests. So any pull request with a previously failing unit test would also require a fix. If you simply have an example that fails without the fix, please file an issue instead.

pfaucon:
I'm new to Tensorflow, so I'm not sure what the best approach would be to go from here but in your opinion would it be reasonable to modify the set input_op_fn as an unpack?  Maybe read in the input tensor and guess the direction of the unpack/slice? Or just remove the default and force people to provide an input function?

PR #4612, opened 2016-09-28
Remove RE2 from the CMake build.

mrry:
It is no longer used internally.


gunan (approved):
If the cmake test passes, I'm happy :)

PR #4621, opened 2016-09-28
Update gradiesnt_descent.py

gautam1858:


jhseu:
Thanks for the change! In the future, please describe the reasoning for the change in your pull request. In this case, passing grad.indices is incorrect because that function is for dense tensors, but .indices only works for sparse tensors.

PR #4625, opened 2016-09-28
spelling mistake of hight is fixed

tusharsoni08:
At link https://www.tensorflow.org/versions/r0.10/tutorials/image_recognition/index.html 

Fix for the typo mistake of **hight** in line "it holds a 299 pixel **high**, 299 pixel width,".


petewarden (changes requested):
Thanks for the update! The spelling should actually be "height", rather than "hight", but to be more idiomatic we should probably change the sentence to read "it holds a 299 pixel high, 299 pixel wide, 3 channel image" instead. Does that make sense?

tusharsoni08:
Ohh sorry! that should be "height" instead of "high" and "hight". It perfectly make sense. :)

jhseu:
Jenkins, test this please

jhseu (approved):

PR #4627, opened 2016-09-28
Resolve urllib dependency for py3.5

jusjusjus:
After this pull, <..>/tutorials/estimators/abalone.py works with python 3.5.

It resolves the dependency on urllib.urlretreive, which shows a different syntax under python 3.5.

### DIFF
--- a/tensorflow/examples/tutorials/estimators/abalone.py
+++ b/tensorflow/examples/tutorials/estimators/abalone.py
@@ -17,8 +17,15 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


mrry:
Thanks for sending in this fix! Elsewhere in the project ([for example](https://github.com/tensorflow/tensorflow/blob/d42facc3cc9611f0c9722c81551a7404a0bd3f6b/tensorflow/models/image/imagenet/classify_image.py)), we use the following rename:

```python
from six.moves import urllib
```

...and then use `urllib.request.urlretrieve()` throughout.

Can you please revise your change to use this style instead?

PR #4630, opened 2016-09-28
fix issues in "nn_layer()" and "cross_entropy"

ziliwang:
1. the act argument in nn_layer() didn't work, I remove it, and assign actions singly.
2. the cross entropy was not correct.


jhseu:
Thanks for the change! These are definitely both issues in the 0.10 release, and they're fixed in the latest master:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/mnist/mnist_with_summaries.py

We're not doing anymore 0.10 releases, and 0.11rc0 will be out soon which will pick up those fixes.

Issue #3856, opened 2016-08-16
Need Protobuf pip wheel package for Python 3.4, which is missing the setup doc


ye:
Current documentation only gives `protobuf-3.0.0b2.post2-cp35-none-any.whl` but nothing for `python 3.4`.

### Environment info
Operating System: Mac OS X El Capitan 10.11.6

Installed version of CUDA and cuDNN: None (Intel Iris Graphics 6100)

If installed from binary pip package, provide:
TF_BINARY_URL=https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-0.10.0rc0-py3-none-any.whl

### Steps to reproduce
1. Follow the Mac [setup steps here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md)
2. Installation completed fine.  
3. Protobuf's package is the pure python one, which is VERY SLOW. The [setup guide](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#protobuf-library-related-issues) only gives a pip wheel package for Python 3.5

### What steps you have tried

```bash
$ pip3 install --upgrade https://storage.googleapis.com/tensorflow/mac/protobuf-3.0.0b2.post2-cp35-none-any.whl
protobuf-3.0.0b2.post2-cp35-none-any.whl is not a supported wheel on this platform.

$ pip3 install --upgrade https://storage.googleapis.com/tensorflow/mac/protobuf-3.0.0b2.post2-cp34-none-any.whl
Collecting protobuf==3.0.0b2.post2 from https://storage.googleapis.com/tensorflow/mac/protobuf-3.0.0b2.post2-cp34-none-any.whl
  HTTP error 404 while getting https://storage.googleapis.com/tensorflow/mac/protobuf-3.0.0b2.post2-cp34-none-any.whl
  Could not install requirement protobuf==3.0.0b2.post2 from https://storage.googleapis.com/tensorflow/mac/protobuf-3.0.0b2.post2-cp34-none-any.whl because of error 404 Client Error: Not Found for url: https://storage.googleapis.com/tensorflow/mac/protobuf-3.0.0b2.post2-cp34-none-any.whl
[...]

PR #4642, opened 2016-09-29
Update instructions for installing protobuf package.

gunan:
To triage #3856

### DIFF
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -902,6 +902,7 @@ $ make -j12
 $ export PROTOC=$PWD/src/protoc
 $ cd python
 $ python setup.py bdist_wheel --cpp_implementation --compile_static_extension
+$ pip uninstall protobuf
 $ pip install dist/<wheel file name>
 ```
 
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -901,8 +901,8 @@ $ CXXFLAGS="-fPIC -g -O2" ./configure
 $ make -j12
 $ export PROTOC=$PWD/src/protoc
 $ cd python
-$ python3.4 setup.py bdist_wheel --cpp_implementation --compile_static_extension
-$ cd dist
+$ python setup.py bdist_wheel --cpp_implementation --compile_static_extension
+$ pip install dist/<wheel file name>
 ```
 
 Install the above package _after_ you have installed TensorFlow via pip, as the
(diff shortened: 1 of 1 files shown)


tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -877,14 +877,32 @@ $ pip install --upgrade https://storage.googleapis.com/tensorflow/linux/cpu/prot
...
+```bash
+$ git clone https://github.com/google/protobuf.git
+$ cd protobuf
+$ ./autogen.sh
+$ CXXFLAGS="-fPIC -g -O2" ./configure
+$ make -j12
+$ export PROTOC=$PWD/src/protoc
+$ cd python
+$ python3.4 setup.py bdist_wheel --cpp_implementation --compile_static_extension
+$ cd dist
```
vrv: what does cd'ing into the directory as the last thing do?

gunan: That is where the wheel file is located. Maybe instead of cd, I can have a comment there to say "whl file is built in "dist" folder" ?

jhseu: Put in the command to pip3 install the whl from dist instead?

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -877,14 +877,32 @@ $ pip install --upgrade https://storage.googleapis.com/tensorflow/linux/cpu/prot
...
+```bash
+$ git clone https://github.com/google/protobuf.git
+$ cd protobuf
+$ ./autogen.sh
+$ CXXFLAGS="-fPIC -g -O2" ./configure
+$ make -j12
+$ export PROTOC=$PWD/src/protoc
+$ cd python
+$ python setup.py bdist_wheel --cpp_implementation --compile_static_extension
+$ pip install dist/<wheel file name>
```
jhseu: pip3 install --upgrade ?

martinwicke:
Probably even more than upgrade, you have to uninstall first otherwise pip
may not install anything find you have 3.0.0 already.

tensorflow/g3doc/get_started/os_setup.md
gunan: not pip3. This should be version independent.
If a user only has python3 and pip3 installed, pip command should still point to pip3 anyway.

But --upgrade, maybe.

gunan:
simply pip uninstall, then pip install?

martinwicke:
That's what I'm thinking but I can't test it right now. Not sure if we need
something like apt's --purge.

tensorflow/g3doc/get_started/os_setup.md
gunan: Tried in virtualenv.
only --upgrade did not work.
uninstall and install did work,
did not need --upgrade

ye:
@martinwicke you are right about uninstall and reinstall. Exactly what I did to get it work - and it's a good habit too.

jhseu (approved):

PR #4647, opened 2016-09-29
Fix gcs_test issues

caisq:
1) pip install issue related to recent numpy upgrade
2) OutOfRange package issue caused by recent changes in TF Python code

### DIFF
--- a/tensorflow/tools/gcs_test/Dockerfile
+++ b/tensorflow/tools/gcs_test/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
 MAINTAINER Shanqing Cai <cais@google.com>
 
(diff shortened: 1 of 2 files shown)


tensorflow/tools/gcs_test/Dockerfile
```diff
@@ -14,6 +14,10 @@ RUN curl -O https://dl.google.com/dl/cloudsdk/channels/rapid/install_google_clou
 RUN chmod +x install_google_cloud_sdk.bash
 RUN ./install_google_cloud_sdk.bash --disable-prompts --install-dir=/var/gcloud
 
+RUN curl -O https://bootstrap.pypa.io/get-pip.py && \
```
gunan: Can we add a comment to explain what this does?

caisq: This change was initially there to make sure that the pip install of tensorflow works. There are some recent changes (probably the numpy version update) which caused this to stop working. But I realized that the problem can be more easily solved with using ubuntu 16 instead of 14. See new changes.

gunan (approved):

PR #4660, opened 2016-09-29
link to better jupyter notebook preview

hrldcpr:
As mentioned in [the deep dream tutorial readme](https://github.com/tensorflow/tensorflow/blob/bc64f05d4090262025a95438b42a54bfdc5bcc80/tensorflow/examples/tutorials/deepdream/README.md) the jupyter notebook preview on nbviewer is better than the one on github. Specifically, github removes the network graph visualizations.

### DIFF
--- a/tensorflow/g3doc/tutorials/index.md
+++ b/tensorflow/g3doc/tutorials/index.md
@@ -112,7 +112,7 @@ Building on the Inception recognition model, we will release a TensorFlow
 version of the [Deep Dream](https://github.com/google/deepdream) neural network
 visual hallucination software.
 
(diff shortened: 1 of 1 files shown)


jhseu:
We probably want to leave it the way it is for consistency with the rest of the tutorials. It should probably be a relative link like the others, though, so feel free to make it a relative link instead.

hrldcpr:
@jhseu As far as I can tell, this is the only tutorial which is a jupyter notebook, and clicking that link just redirects to github, so I think it's doomed to inconsistency regardless 😉

jhseu (approved):
After thinking about it, you're right that this link is better.

jhseu:
Jenkins, test this please

PR #4666, opened 2016-09-30
autocomplete and image read fix

neilhan:
Added pyreadline to enable ipython autocomplete
Added Pillow to fix an image loading error


neilhan:
I signed it!

PR #4670, opened 2016-09-30

yli192:


jhseu:
Thanks for the change! These are already fixed in master and are in the latest 0.11.0rc0 release, which we're pushing out right now. We're not making anymore 0.10 releases.

PR #4676, opened 2016-09-30
Generate additional cudart linkopts in cuda_configure.

davidzchen:
Currently, the target `@local_config_cuda//cuda:cudart_static` contains a
`select()` condition that references the `@//tensorflow:darwin` config
condition.

While this is fine when building out of the `tensorflow` workspace, this breaks
the build for serving and other projects that vendor TF since those workspaces
do not have a `//tensorflow:darwin` target (under serving, that label would be
`//tensorflow/tensorflow:darwin` under the current setup).

This change removes the `select()` and moves the logic for generating additional
platform-specific linkopts for cudart_static into cuda_configure, making the
`@local_config_cuda` workspace more self-contained.

### DIFF
--- a/third_party/gpus/cuda/BUILD.tpl
+++ b/third_party/gpus/cuda/BUILD.tpl
@@ -58,10 +58,8 @@ cc_library(
     linkopts = [
         "-ldl",
         "-lpthread",
-    ] + select({
-        "@//tensorflow:darwin": [],
-        "//conditions:default": ["-lrt"],
-    }),
+        %{cudart_static_linkopt}
+    ],
     visibility = ["//visibility:public"],
 )
 
--- a/third_party/gpus/cuda_configure.bzl
+++ b/third_party/gpus/cuda_configure.bzl
@@ -315,6 +315,11 @@ def _find_cudnn_lib_path(repository_ctx, cudnn_install_basedir, symlink_files):
         cudnn_install_basedir))
 
 
+def _cudart_static_linkopt(cpu_value):
+  """Returns additional platform-specific linkopts for cudart."""
+  return "" if cpu_value == "Darwin" else "\"-lrt\","
+
+
 def _tpl(repository_ctx, tpl, substitutions={}, out=None):
   if not out:
     out = tpl.replace(":", "/")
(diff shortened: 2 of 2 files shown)


third_party/gpus/cuda/BUILD.tpl
```diff
@@ -58,10 +58,8 @@ cc_library(
     linkopts = [
         "-ldl",
         "-lpthread",
-    ] + select({
-        "@//tensorflow:darwin": [],
```
jhseu: This does exist, but it's just a typo, right?
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/BUILD#L41

Probably should be "//tensorflow:darwin" without the @

davidzchen: Ah, sorry, you're right. I was building from within the workspace of TF serving. I'll investigate this further.

davidzchen: It looks like we still need this change.

This `BUILD` file is generated in the `@local_config_cuda` workspace, which is created by `cuda_configure`. The `@//` syntax refers to a label in the main workspace, which is fine when building from the `tensorflow` workspace since `//tensorflow:darwin` exists. 

Changing this from `@//` to `//` breaks the build because then the `//tensorflow:darwin` resolves to `@local_config_cuda//tensorflow:darwin` does not exist.

Similarly, when building out of the `serving` workspace, this also breaks because `//tensorflow:darwin` does not exist in `serving` (it would be `//tensorflow/tensorflow:darwin` based on the way the repository is set up currently).

In any case, since `cuda_configure` already performs platform-specific checks, it would be more consistent to keep this change which moves the platform-specific check into `cuda_configure`.

davidzchen:
I have updated the commit message based on the discussion.

PTAL.

davidzchen:
Looks like Jenkins is having problems:

```
javax.servlet.ServletException: org.apache.commons.jelly.JellyTagException: jar:file:/var/cache/jenkins/war/WEB-INF/lib/jenkins-core-2.24.jar!/hudson/model/AbstractBuild/index.jelly:110:84: <st:include> org.apache.commons.jelly.JellyTagException: jar:file:/var/cache/jenkins/war/WEB-INF/lib/jenkins-core-2.24.jar!/lib/hudson/summary.jelly:105:31: <d:invokeBody> Cannot get property 'fullName' on null object
...
```

davidzchen:
Jenkins, test this please.

yifeif:

jhseu (approved):

PR #4678, opened 2016-09-30
dynamic_raw_rnn

alrojo:
Working example of the recent [raw_rnn](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/rnn.py#L1032) by @ebrevdo .

This working example implements the [dynamic_rnn](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/rnn.py#L685) without tuple support (yet) using the `raw_rnn` interface.

The working example builds on this [snippet](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/rnn.py#L1076), but is fully functional and provides the same interface as `dynamic_rnn` using the raw_rnn implementation.

The purpose is not to replace `dynamic_rnn` (at least not yet?), but to give a code base with commentaries for implementing custom RNNs using the `raw_rnn` interface.

### DIFF
--- a/tensorflow/python/ops/rnn.py
+++ b/tensorflow/python/ops/rnn.py
@@ -1453,7 +1453,7 @@ def dynamic_raw_rnn(cell, inputs, sequence_length=None, initial_state=None,
     outputs_ta, final_state, _ = \
       raw_rnn(cell=cell, loop_fn=loop_fn,
               parallel_iterations=parallel_iterations,
-              swap_memory=swap_memory, scope=scope)
+              swap_memory=swap_memory, scope=varscope)
     outputs = outputs_ta.pack()
     if not time_major:
       # [seq, batch, features] -> [batch, seq, features]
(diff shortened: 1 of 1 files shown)


ebrevdo:
Can you talk about the reasoning behind this PR?  It seems to duplicate a subset dynamic_rnn's functionality.  We're not going to replace dynamic_rnn with raw_rnn wrappers because raw_rnn doesn't have access to the input size; thereby disallowing certain optimizations that dynamic_rnn can make now and in the future.

alrojo:
@ebrevdo I was not aware of thoese optimizations. Perhaps this is more suitable as an "introduction to raw_rnn" in the commentary like your [snippet](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/rnn.py#L1076). If that is of interest I will re-PR it as such?

PR #4687, opened 2016-09-30
Fixed documentation for batch_normalization

DreSimpelo:
Fixed batch_normalization part of the nn python Documentation.

Fixed syntax errors restricting the latex code to load in the markdown documentation file for "tf.nn.batch_normalization(...)".

### DIFF
--- a/tensorflow/g3doc/api_docs/python/nn.md
+++ b/tensorflow/g3doc/api_docs/python/nn.md
@@ -3221,9 +3221,9 @@ Batch normalization.
 
 As described in http://arxiv.org/abs/1502.03167.
 Normalizes a tensor by `mean` and `variance`, and applies (optionally) a
-`scale` \\\\(\gamma\\\\) to it, as well as an `offset` \\\\(\\beta\\\\):
+`scale` \\(\gamma\\) to it, as well as an `offset` \\(\beta\\):
 
-\\\\(\\frac{\gamma(x-\mu)}{\sigma}+\\beta\\\\)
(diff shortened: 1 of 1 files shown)


tensorflow/g3doc/api_docs/python/nn.md
```diff
@@ -3221,9 +3221,9 @@ Batch normalization.
 
 As described in http://arxiv.org/abs/1502.03167.
 Normalizes a tensor by `mean` and `variance`, and applies (optionally) a
-`scale` \\\\(\gamma\\\\) to it, as well as an `offset` \\\\(\\beta\\\\):
+`scale` \\(\gamma\\) to it, as well as an `offset` \\(\beta\\):
 
-\\\\(\\frac{\gamma(x-\mu)}{\sigma}+\\beta\\\\)
+$$\frac{\gamma(x-\mu)}{\sigma}+\beta$$
```
jhseu: Can you leave it as \\( and \\) instead of switching to $$?

jhseu (changes requested):
Thanks for the change! The API docs are autogenerated. You'll have to change this file instead:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn.py#L939

tensorflow/g3doc/api_docs/python/nn.md
DreSimpelo: Done! (The equation is no longer centered after that change)

DreSimpelo:
Moved: https://github.com/tensorflow/tensorflow/pull/4689

PR #4691, opened 2016-09-30
Merge r0.11 branch to get any fixes in the branch.

gunan:


gunan:
Dropping in favor of #4694

PR #4684, opened 2016-09-30
Fix skflow resnet example to work on v0.10.0

tyfkda:
I tried to run `resnet.py` skflow example on TensorFlow v0.10.0,
but an error occurred.

I have fixed the code to run on v0.10.0:
- `tensorflow.contrib.learn.ops.conv2d` was deprecated and has been removed in #4373 .
  So changed to use `contrib.layers.convolution2d` .
- `classifier` wasn't created if no previous model existed in `models/resnet/graph.pbtxt`,
  but it was failed to run at first time.
  `TensorFlowEstimator` creation was removed in #2863 , so revert the code removal.

PR #4696, opened 2016-09-30
Enable to pass feed_fn to fit method

tyfkda:
There was no way to use `fit` method to pass `feed_fn`.

I tried to use [fit](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L314) method of `learn.estimator.Estimator` (actually the method is defined in its parent `BaseEstimator`).

Also I tried to use [DataFeeder](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py#L217) to feed data. I can use `input_builder` and `get_feed_dict_fn()` of `DataFeeder` instance to extract data. But when I tried to feed data to `fit` method, I couldn't pass `feed_fn`, extracted from `DataFeeder`.

This change add `feed_fn` keyword argument to `fit` method, to handle above situation.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/estimators/dnn.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/dnn.py
@@ -423,8 +423,8 @@ class DNNClassifier(evaluable.Evaluable, trainable.Trainable):
         },
         feature_engineering_fn=feature_engineering_fn)
 
-  def fit(self, x=None, y=None, input_fn=None, steps=None, batch_size=None,
-          monitors=None, max_steps=None):
+  def fit(self, x=None, y=None, input_fn=None, feed_fn=None, steps=None,
+          batch_size=None, monitors=None, max_steps=None):
     """See trainable.Trainable."""
     # TODO(roumposg): Remove when deprecated monitors are removed.
     if monitors is not None:
@@ -436,9 +436,9 @@ class DNNClassifier(evaluable.Evaluable, trainable.Trainable):
         monitor.set_estimator(self)
         monitor._lock_estimator()  # pylint: disable=protected-access
 
-    result = self._estimator.fit(x=x, y=y, input_fn=input_fn, steps=steps,
-                                 batch_size=batch_size, monitors=monitors,
-                                 max_steps=max_steps)
+    result = self._estimator.fit(x=x, y=y, input_fn=input_fn, feed_fn=feed_fn,
+                                 steps=steps, batch_size=batch_size,
+                                 monitors=monitors, max_steps=max_steps)
 
     if monitors is not None:
       for monitor in deprecated_monitors:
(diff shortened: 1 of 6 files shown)


tyfkda:
@terrytangyuan, would you please take a look? Thanks!

terrytangyuan:
I think you also need to change it in Trainable with docstring. @martinwicke should be able to help more.

abhitopia:
Is there a need to support feed_fn because as @martinwicke pointed out [here](https://groups.google.com/a/tensorflow.org/forum/#!topic/discuss/ZEzEa1TyYuE), `x` and `y` already support iterators by using `StreamingDataFeeder` internally. I wonder what would be use case for needing feed_fn explicitly?

tyfkda:
@abhitopia 
Thank you for your information, I am going to read the thread.

I am wondering that I can replace `tensorflow.contrib.learn.TensorFlowEstimator` to `Estimator`, suggested by @terrytangyuan in [#4684](https://github.com/tensorflow/tensorflow/pull/4684#issuecomment-250772081).
`TensorFlowEstimator#fit` uses `feed_fn` ([ref.](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/estimators/base.py#L255)) But I cannot do the same thing using `Estimator` because it doesn't accept `feed_fn` for `fit` method.

martinwicke:
Y​ou should be able to use the x and y arguments.

abhitopia:
@martinwicke - I thought about it a bit more. Although `fit()` support `x` and `y` to be iterables, it constraints them to be tensors(Dask or Numpy). On the other hand, with `input_fn`, you can pass arbitrary dictionary. Many NLP problems require `x` to be more than just one tensor. For instance, we might need to pass the length of variable length batch. How can work around this without having to use some sort of `Reader`?

martinwicke:
You can use a `py_func` inside your input_fn. There's lots of downsides to using a `py_func`, but in this case you probably don't care much. Would that work for you?

I don't think we want to add more arguments to the signature of fit.

tyfkda:
I can use `Estimator#fit` with passing `x` and `y` argument,
so solved the problem.

I am going to close this pull request.

abhitopia:
@martinwicke - Thanks. Would you be kind enough to suggest a little blueprint on writing your proposed `input_fn` using `py_func`, please? For context you can assume, there is some function `get_data()` which fetches one sample of data point from database on every call and returns `None` when one epoch is completed.

martinwicke:
That is a question best asked on StackOverflow, where others will be able
to find it more easily.

abhitopia:
@martinwicke - I have posted the question here-http://stackoverflow.com/questions/40033667/how-to-pass-data-from-arbitrary-sources-not-files-though-tensorflow-contrib-le

PR #4699, opened 2016-09-30
Fail configure script immediately on error.

### LINKED ISSUE #4662: ./configure should fail if the bazel commands fail
The configure script now calls bazel clean and bazel fetch (as of ~Aug 2016).  If they fail, the script should exit with a nonzero exit status rather than continuing onwards.  This is important for automated build systems to detect failed configure calls.

davidzchen:
For lines such as `default_python_bin_path=$(which python)`, add a `|| true`
expression so that if the executable does not exist, the command will not exit
with a non-zero exit code, and the variable will be set to an empty string.

Also, use `yes` to automate the `./configure` script when running CI tests
rather than piping in an empty file, which causes subsequent `read` commands
to fail.

This change also augments the code in the configure script for resolving the
cuDNN library to search both the case where libcudnn is installed to a lib/lib64
directory (such as /usr/local/cuda-7.5/lib64) and when it is not (such as
/usr/lib/x86_64-linux-gnu), as well as when the platform is Darwin.

Fixes #4662

### DIFF
--- a/configure
+++ b/configure
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -e
+set -o pipefail
+
 # Find out the absolute path to where ./configure resides
 pushd `dirname $0` #> /dev/null
 SOURCE_BASE_DIR=`pwd -P`
@@ -14,7 +17,7 @@ function bazel_clean_and_fetch() {
 while true; do
   fromuser=""
   if [ -z "$PYTHON_BIN_PATH" ]; then
-    default_python_bin_path=$(which python)
+    default_python_bin_path=$(which python || true)
     read -p "Please specify the location of python. [Default is $default_python_bin_path]: " PYTHON_BIN_PATH
     fromuser="1"
     if [ -z "$PYTHON_BIN_PATH" ]; then
@@ -47,7 +50,6 @@ while [ "$TF_NEED_GCP" == "" ]; do
 done
 
 if [ "$TF_NEED_GCP" == "1" ]; then
-
   ## Verify that libcurl header files are available.
   # Only check Linux, since on MacOS the header files are installed with XCode.
   if [[ $(uname -a) =~ Linux ]] && [[ ! -f "/usr/include/curl/curl.h" ]]; then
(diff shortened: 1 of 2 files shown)


jhseu (commented):
Thanks for the change! Looks like the GPU build failed during configure. Ping me in chat if you need help debugging that.

Also adjust the python_config... || exit -1 line to remove the || exit -1?

davidzchen:
Interesting, it looks like it is failing at the following line:

```sh
    read -p "Please specify which gcc should be used by nvcc as the host compiler. [Default is $default_gcc_host_compiler_path]: " GCC_HOST_COMPILER_PATH
```

which is exiting with exit code `1`.

It seems that the current approach of piping in an empty file to the invocation of the configure script makes subsequent `read` commands to exit with an error code.

I have added a `-d` flag to have the `configure` script use defaults when finding libraries and toolchains on the system and updating the `configured` script to invoke `configure` with that flag.

Another command that causes `configure` to fail is when the script tries to resolve `libcudnn`. I found that the script only covers the case where `libcudnn` is installed to a `lib`/`lib64` directory, such as `/usr/local/cuda-7.5/lib64`, but not case where it isn't, such as `/usr/lib/x86_64-linux-gnu`, which is, in fact, the cuDNN install dir on the GPU docker image.

davidzchen:
The tests are now passing. PTAL.

configure
```diff
@@ -1,5 +1,23 @@
 #!/usr/bin/env bash
 
+set -e
+set -o pipefail
+
+function usage {
+  echo "Usage configure [-d]"
+  echo "-d Use default configuration settings."
```
jhseu: Instead of adding a new flag, we could just run `yes "" | ./configured" in the CI build, right?

configure
```diff
@@ -259,6 +304,7 @@ while true; do
   if [ "$OSNAME" == "Linux" ]; then
     echo "${CUDNN_PATH_FROM_LDCONFIG}${TF_CUDNN_EXT}"
   fi
+  echo " g"
```
jhseu: Leftover from debugging? :)

jhseu (changes requested):

configure
davidzchen: Great idea!

configure
davidzchen: Yes :)

jhseu (approved):
Thanks!

Jenkins, test this please

davidzchen:
The failing test in the Linux CPU Tests (Python 3) run, `//tensorflow/contrib/learn:graph_io_test`, does not appear to be related to this change.

jhseu:
Weird, I've seen that before where graph_io_test "fails", but all the tests in it passed. Feel free to ignore that test and merge @yifeif.

yifeif:
I have seen similar failure sporadically in nightly py3 builds as well. Will investigate separately. Merging. Thanks @davidzchen and @jhseu!

PR #4717, opened 2016-10-02
Adding missing module

ultraklon:
In the library testing section the PIL module was missing


vincentvanhoucke:
What's the problem exactly? There is already:
`from IPython.display import display, Image`

ultraklon:
I tried unistalling pillow and running the steps again and they work
I could not reproduce the problem, don't mind me

PR #4720, opened 2016-10-02
check_version: fails if native.bazel_version is undefined

damienmg:
This feature is in Bazel since 0.2.1, TensorFlow requires 0.3.0
at least and should error out if the feature is not present.

Also give a warning when the version number is not present
(when Bazel is built from HEAD) for better debugging.

See http://stackoverflow.com/questions/39811568/error-function-repository-rule-does-not-exist-while-trying-to-compile-tensor.

### DIFF
--- a/tensorflow/tensorflow.bzl
+++ b/tensorflow/tensorflow.bzl
@@ -25,7 +25,7 @@ def src_to_test_name(src):
 def check_version(bazel_version):
   if "bazel_version" not in dir(native):
     fail("\nCurrent Bazel version is lower than 0.2.1, expected at least %s\n" % bazel_version)
-  if not native.bazel_version:
+  elif not native.bazel_version:
     print("\nCurrent Bazel is not a release version, cannot check for compatibility.")
     print("Make sure that you are running at least Bazel %s.\n" % bazel_version)
   else:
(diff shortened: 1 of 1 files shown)


damienmg:
I would also suggest @gunan :)

Also if I ask, do I get ci to test it:
Jenkins test this please

tensorflow/tensorflow.bzl
```diff
@@ -23,13 +23,17 @@ def src_to_test_name(src):
 
 # Check that a specific bazel version is being used.
 def check_version(bazel_version):
-  if "bazel_version" in dir(native) and native.bazel_version:
+  if "bazel_version" not in dir(native):
+    fail("\nCurrent Bazel version is lower than 0.2.1, expected at least %s\n" % bazel_version)
+  if not native.bazel_version:
```
gunan: small nit
"elif not native.bazel_version:"

tensorflow/tensorflow.bzl
```diff
@@ -23,13 +23,17 @@ def src_to_test_name(src):
...
+    print("\nCurrent Bazel is not a release version, cannot check for compatibility.")
+    print("Make sure that you are running at least Bazel %s.\n" % bazel_version)
```
gunan: nit:
call "pass" after this line?

damienmg: Done, I am a bit curious, why add the "pass"?

tensorflow/tensorflow.bzl
damienmg: Done

tensorflow/tensorflow.bzl
gunan: To me, it looked just a little more readable.
But I can be convinced otherwise

gunan (approved):

PR #4724, opened 2016-10-03
Fix python bath in python_config script

viirya:
When runing `./configure`, two places in the script `python_config` try to invoke `python` but don't use the python-related environment variable `PYTHON_BIN_PATH`. So it will cause the error that can't find python during configuration.

### DIFF
--- a/util/python/python_config.sh
+++ b/util/python/python_config.sh
@@ -46,7 +46,7 @@ function main {
 }
 
 function python_path {
-  python - <<END
+  $PYTHON_BIN_PATH - <<END
 from __future__ import print_function
 import site
 import os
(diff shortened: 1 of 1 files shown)


yifeif:
@itsmeolivia could you take a look? Thanks.

gunan:
Jenkins, test this please

util/python/python_config.sh
```diff
@@ -46,7 +46,7 @@ function main {
 }
 
 function python_path {
-  python - <<END
+  $PYTHON_BIN_PATH - <<END
```
gunan: In my system, $PYTHON_BIN_PATH seems to be undefined.
Are you sure this will work in all systems?

viirya: Are you running this script directly? I encounter this issue when I run `./configure` and provide custom python path in when it asks for that. `configure` will invoke this `python_config.sh`.

gunan: Missed that. You are right.

gunan (approved):
Yifei, could you also cherrypick this into r0.11?
This change help fix installation issues on systems that only have python3.

PR #4745, opened 2016-10-04
Added changes to scoping to the release notes

jonasrauber:
The changes to name scoping and variable scoping are important breaking changes and should be listed as major changes.


martinwicke:
This should also be in the r0.11 branch. Can you make another PR against r0.11?

PR #4751, opened 2016-10-04
fix comment in notebook

davidbrai:


davidbrai:
I signed it!

yifeif (approved):
LG, thanks @davidbrai !

PR #4776, opened 2016-10-05
Fix python3 encoding issue with data format string

admcrae:
This fixes the python 3 issue found by @gunan in #4411.


yifeif:
Thanks @admcrae! @tensorflow-jenkins test this please.

yifeif (approved):

gunan:
Thank you very much for the Fix @admcrae !

PR #4777, opened 2016-10-05
bug fix when iterators stops at multiple of batch_size

abhitopia:
In contrib.learn framwork,  `Estmator.fit()` supports `x`, `y` and `batch_size` arguments where `x` and `y` could be iterators/generators. However, when iterator finishes right at the edge, after exactly integer multiple of batch_size, a run is made anyway with batch_size = 0. This causes NaN errors with following code snippet. 

[Gist Link](https://www.google.com/url?q=https%3A%2F%2Fgist.github.com%2Fabhitopia%2Fb6b60cb8bf9bab58a78b221fb3c4473b&sa=D&sntz=1&usg=AFQjCNEgS-muOb4Z0UC3SDBQ-F5KcfBMbQ) - to reproduce error.

For more information on the issue see
[Tensorflow google group discussion](https://groups.google.com/a/tensorflow.org/forum/#!topic/discuss/ZEzEa1TyYuE)

Simply skipping session.run when feed_dict is empty solves the problem.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/graph_actions.py
+++ b/tensorflow/contrib/learn/python/learn/graph_actions.py
@@ -285,8 +285,13 @@ def _monitored_train(graph,
         hooks=all_hooks) as super_sess:
       loss = None
       while not super_sess.should_stop():
-        _, loss = super_sess.run([train_op, loss_op], feed_fn() if feed_fn else
-                                 None)
+        feed_dict = feed_fn() if feed_fn else None
+        # check for case when iterator stopped exactly at multiple of batch_size
+        # In such case, all the returned numpy array are len zero.
+        edge_condition = True if feed_dict and all(len(v) == 0 for v in feed_dict.values()) else False
+        if not edge_condition:
+          _, loss = super_sess.run([train_op, loss_op], feed_dict)
+          
       return loss
 
 
--- a/tensorflow/contrib/learn/python/learn/graph_actions.py
+++ b/tensorflow/contrib/learn/python/learn/graph_actions.py
@@ -285,13 +285,8 @@ def _monitored_train(graph,
         hooks=all_hooks) as super_sess:
       loss = None
       while not super_sess.should_stop():
-        feed_dict = feed_fn() if feed_fn else None
-        # check for case when iterator stopped exactly at multiple of batch_size
-        # In such case, all the returned numpy array are len zero.
-        edge_condition = True if feed_dict and all(len(v) == 0 for v in feed_dict.values()) else False
-        if not edge_condition:
-          _, loss = super_sess.run([train_op, loss_op], feed_dict)
-          
+        _, loss = super_sess.run([train_op, loss_op], feed_fn() if feed_fn else None)
+        
       return loss
 
 
--- a/tensorflow/contrib/learn/python/learn/graph_actions.py
+++ b/tensorflow/contrib/learn/python/learn/graph_actions.py
@@ -287,7 +287,6 @@ def _monitored_train(graph,
       while not super_sess.should_stop():
         _, loss = super_sess.run([train_op, loss_op], feed_fn() if feed_fn else
                                  None)
-        
       return loss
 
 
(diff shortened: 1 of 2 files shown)


martinwicke:
Thanks for sending a PR!

I think this is the wrong place to fix this. Instead, I think there is a bug in `StreamingDataFeeder`, `data_feeder.py` around line 510.

abhitopia:
@martinwicke  - Thanks for pointing out, please check new commits and let me know if that makes sense.

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
```diff
@@ -507,10 +507,13 @@ def _feed_dict_fn():
           inp[i, :] = six.next(self._x)
         except StopIteration:
           self.stopped = True
-          inp = inp[:i, :]
-          if self._y is not None:
-            out = out[:i]
-          break
+          if i==0:
+            raise StopIteration
```
martinwicke: Should this just be `raise` in to preserve the original stack?

tensorflow/contrib/learn/python/learn/graph_actions.py
```diff
@@ -285,8 +285,8 @@ def make_saver():
         hooks=all_hooks) as super_sess:
       loss = None
       while not super_sess.should_stop():
-        _, loss = super_sess.run([train_op, loss_op], feed_fn() if feed_fn else
-                                 None)
+        _, loss = super_sess.run([train_op, loss_op], feed_fn() if feed_fn else None)
+
```
martinwicke: Can you revert this completely? I think the break was there for the 80 character limit.

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
```diff
@@ -507,10 +507,13 @@ def _feed_dict_fn():
           inp[i, :] = six.next(self._x)
         except StopIteration:
           self.stopped = True
-          inp = inp[:i, :]
-          if self._y is not None:
-            out = out[:i]
-          break
+          if i==0:
```
martinwicke: Please add spaces around `==`

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
```diff
@@ -507,10 +507,13 @@ def _feed_dict_fn():
...
+            raise StopIteration
+          else:
```
martinwicke: You don't need the `else:` (since `raise` never returns), and I would find it more legible without.

martinwicke (changes requested):

martinwicke:
This is the right change. Just a few small comments.

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
abhitopia: @martinwicke - done

tensorflow/contrib/learn/python/learn/graph_actions.py
abhitopia: @martinwicke - done.

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
abhitopia: @martinwicke - Not sure if I understand your question very well. I presume StopIteration is automatically picked by `super_session.should_stop()` in graph_action.py line 287

martinwicke: I meant, omit the `StopIteration`. `raise StopIteration` will create a new `StopIteration` instance with a stack trace to line 511, while the original `StopIteration is discarded. If you instead simply write `if i == 0: raise`, you will reuse the original exception, which contains the full stack trace.

It doesn't matter much in this case (not much useful information in a `StopIteration` anyway, and it'll be caught shortly by ourselves), but it's better to preserve it anyway.

abhitopia: @martinwicke - Agreed. Omitted new `StopIteration` now. Thanks for explaining.

martinwicke:
Jenkins, test this please

martinwicke (approved):

PR #4780, opened 2016-10-05
contrib/quantization: Do not open dot file in binary mode

olesalscheider:
The Python 3 documentation states:

Since printed arguments are converted to text strings,
print() cannot be used with binary mode file objects.


yifeif:

yifeif (approved):

yifeif:
Thanks @olesalscheider !

PR #4783, opened 2016-10-05
Fix: Update nightly links in readme.md.

gunan:


caisq (approved):
Thanks for fixing this!

PR #4789, opened 2016-10-05
Overwrite the target file in HDFS Rename operation; it removes the ta…

jhseu:
…rget file first if it exists.

Change: 134682344


gunan (approved):
Blind approval for cherrypick :)

PR #4796, opened 2016-10-06
Make tutorial_example_trainer build on Windows with Bazel

meteorcloudy:
With these changes, we are able to build the C++ example trainer by 
`bazel build --host_cpu=x64_windows_msvc --cpu=x64_windows_msvc -c opt tensorflow/cc:tutorials_example_trainer` 
with Bazel from HEAD on Windows.
I modified the BUILD files carefully, hopefully it won't break TF build on other platform.
@mrry @damienmg @dslomov

### DIFF
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -981,10 +981,11 @@ cc_library(
                     "**/*test*",
                     "platform/**/cuda.h",
                     "platform/**/stream_executor.h",
-                ] +
+                ] 
                 # Protobuf deps already included through the ":lib_proto_parsing"
                 # dependency.
-                tf_additional_proto_srcs(),
+                # Excluding logging.h from srcs will cause a linking error
+                # + tf_additional_proto_srcs(),
     ),
     hdrs = tf_additional_lib_hdrs() + [
         "lib/core/blocking_counter.h",
@@ -1057,6 +1058,7 @@ tf_version_info_genrule()
 cc_library(
     name = "version_lib",
     srcs = ["util/version_info.cc"],
+    copts = tf_copts(),
 )
 
 tf_cuda_library(
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -141,6 +141,7 @@ cc_library(
         "platform/protobuf.h",
         "platform/types.h",
     ] + glob(tf_additional_proto_hdrs()),
+    copts = tf_copts(),
     deps = [
         ":protos_all_cc",
         "//tensorflow/core/platform/default/build_config:proto_parsing",
@@ -979,11 +980,10 @@ cc_library(
                     "**/*test*",
                     "platform/**/cuda.h",
                     "platform/**/stream_executor.h",
-                ] 
+                ] +
                 # Protobuf deps already included through the ":lib_proto_parsing"
                 # dependency.
-                # Excluding logging.h from srcs will cause a linking error
-                # + tf_additional_proto_srcs(),
+                tf_additional_proto_srcs(),
     ),
     hdrs = tf_additional_lib_hdrs() + [
         "lib/core/blocking_counter.h",
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -535,8 +535,6 @@ cc_library(
         "//tensorflow/core/kernels:fact_op",
         "//tensorflow/core/kernels:array_not_windows",
         "//tensorflow/core/kernels:math_not_windows",
-        "//tensorflow/core/kernels:linalg_not_windows",
-        "//tensorflow/core/kernels:logging_not_windows",
     ]),
 )
 
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -64,6 +64,7 @@ load(
     "//tensorflow:tensorflow.bzl",
     "if_android",
     "if_ios",
+    "if_not_windows",
     "tf_copts",
     "tf_cc_test",
     "tf_cc_tests",
@@ -871,12 +872,12 @@ cc_library(
 # Libraries with GPU facilities that are useful for writing kernels.
 cc_library(
     name = "gpu_lib",
-    srcs = [
+    srcs = if_not_windows([
         "common_runtime/gpu/gpu_event_mgr.cc",
-    ],
-    hdrs = [
+    ]),
+    hdrs = if_not_windows([
         "common_runtime/gpu/gpu_event_mgr.h",
-    ],
+    ]),
     copts = tf_copts(),
     visibility = ["//visibility:public"],
     deps = [
@@ -886,8 +887,7 @@ cc_library(
         ":lib_internal",
         ":proto_text",
         ":protos_all_cc",
-        ":stream_executor",
-    ],
+    ] + if_not_windows([":stream_executor"]),
 )
 
 cc_library(
@@ -1341,7 +1341,7 @@ tf_cuda_library(
 
 tf_cuda_library(
     name = "gpu_runtime",
-    srcs = [
+    srcs = if_not_windows([
         "common_runtime/gpu/gpu_bfc_allocator.cc",
         "common_runtime/gpu/gpu_debug_allocator.cc",
         "common_runtime/gpu/gpu_device.cc",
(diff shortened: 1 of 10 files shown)


mrry:
Thanks for sending this in, Yun!

mrry:

tensorflow/core/BUILD
```diff
@@ -1348,9 +1380,8 @@ tf_cuda_library(
         ":lib",
         ":lib_internal",
         ":protos_all_cc",
-        ":stream_executor",
         "//third_party/eigen3",
-    ],
+    ] + if_not_windows([":stream_executor"]),
```
mrry: Instead of making this an (almost) no-op when building on Windows, could we modify `:core` so that it only pulls in `:gpu_runtime` on not-Windows?

There are a few targets that depend on `:gpu_runtime` in this file, but they all appear to depend on `:core_cpu`, so AFAICT this could be replaced with a dependency on `:core`.

meteorcloudy:
You are welcome! Looks like there are many failures, I'll look into it.

tensorflow/core/kernels/BUILD
```diff
@@ -1608,7 +1657,9 @@ tf_kernel_libraries(
         "//tensorflow/core:nn_grad",
         "//tensorflow/core:nn_ops_op_lib",
         "//third_party/eigen3",
-    ],
+    ] + if_not_windows([
```
mrry: I'm confused why this is treated differently from the `tf_kernel_libraries()` target, and thus not in `:math_not_windows`. (It may have been in the wrong place originally, but it'd be good to reduce the complexity if possible.)

I believe it should be possible to remove `:depthwise_conv_op` from the deps here, and move it into the `prefixes` of `:math_not_windows`. Does that make sense?

tensorflow/core/platform/default/build_config.bzl
```diff
@@ -90,17 +90,31 @@ def tf_proto_library(name, srcs = [], has_services = None,
...
+        "platform/posix/error.h",
+      ], exclude = exclude),
+    "//conditions:default" : native.glob([
+        "platform/default/*.h",
+        "platform/posix/*.h",
+      ], exclude = exclude),
+  })
+
+def tf_additional_lib_srcs(exclude = []):
+  return  select({
```
mrry: Nit: only one space between `return` and `select`

meteorcloudy: Thanks~

tensorflow/core/kernels/BUILD
meteorcloudy: Sounds good, I'll verify if it works

tensorflow/core/BUILD
meteorcloudy: I found there are  4 targets we need to modify if we want to cut the dependency on `gpu_runtime`.
`//tensorflow/core:core`, 
`//tensorflow/core:tensorflow_opensource`, 
`//tensorflow/core/util/ctc:ctc_beam_search_lib`, 
`//tensorflow/core/kernels:array`
As you can see, they are scattered in 3 different files, I kind of prefer the original change, so it will be easier when we enable gpu support later. WDYT?

tensorflow/core/BUILD
```diff
@@ -943,26 +951,41 @@ tf_proto_library_cc(
...
+                ]),
+    }) + tf_additional_lib_srcs(
+                exclude = [
+                    "**/*test*",
+                    "platform/**/cuda.h",
+                    "platform/**/stream_executor.h",
+                ] +
+                # Protobuf deps already included through the ":lib_proto_parsing"
+                # dependency.
+                tf_additional_proto_srcs(),
```
meteorcloudy: Thanks, I'll reformat all the BUILD files with buildifier.

tensorflow/core/kernels/BUILD
meteorcloudy: Hmm..  As you can see the definition [here](https://github.com/meteorcloudy/tensorflow/blob/fe74dfaf000ecef0b36b847f50bb0283e8a8ac71/tensorflow/core/kernels/BUILD#L1602)
Looks like `math_now_windows` lacks dependencies `depthwise_conv_op` needs.
        `:conv_ops`,        `:ops_util`, `//tensorflow/core:nn_ops_op_lib`
Do you think it's fine to add them into deps of `math_not_window`?

meteorcloudy:
Can we test it again?

yifeif:

yifeif:
@mrry could you take another look?

tensorflow/core/BUILD
mrry: Some of these dependencies look spurious (especially for `ctc_beam_search_lib`), but fixing that is probably orthogonal to this change. (`"gpu_runtime"` seems like a strange thing for CPU-only code to depend on :)....) Fine to leave as is.

tensorflow/core/kernels/BUILD
meteorcloudy: Yes, I agree!

meteorcloudy:
@mrry Just pushed, please review again. :)

mrry:

mrry (approved):
Looks good to me - thanks for the cleanup!

meteorcloudy:
From the error message in Linux GPU, looks like we should also pull `depthwise_conv_grad_op` out of deps, but  `depthwise_conv_grad_op` [needs `depthwise_conv_op.h`](https://github.com/meteorcloudy/tensorflow/blob/fe74dfaf000ecef0b36b847f50bb0283e8a8ac71/tensorflow/core/kernels/BUILD#L1617). What should we do?

meteorcloudy:
Or maybe they are in the deps for a reason? See these two commits b51ef0cd06e1bfb529b272e55010790ff3ead31f, 99e298a84775fcfe91ce0e2d40566d4bd00f53f9

mrry:
I'm not sure what the proper solution would be for them. Assigning to @jmchen-g, since he wrote those ops originally.

jmchen-g:
Which deps are not clear here? If you mean why it is in the deps of nn, Depthwise is one of the convs that we need to support in TF and is commonly used to reduce the computing needs for a model.

meteorcloudy:
OK, I see. Sounds like `depthwise_conv_ops` is supposed to be in deps of nn. I've deleted my last commit. Please test again?

meteorcloudy:
@mrry Are you OK with keeping `depthwise_conv_ops` in deps of `nn`? If so, I think this PR is ready to merge.

mrry:
@meteorcloudy Yes, that's fine. I suspect our existing build rules are not as clean as they could be for this op, but that shouldn't derail what you're trying to do :).

I'll go ahead and merge now.

meteorcloudy:
Nice, thanks! \o/

rohan100jain:
This change is causing problems in testing now with BUILD rules. Apparently select() and glob() don't play nicely with each other.

meteorcloudy:
@rohan100jain Can you show me the error message? I wonder why it's not causing problem before.. Users are beginning to try Windows build with Bazel now, reverting this will completely break them.

rohan100jain:
Even after fixing the bug at  https://github.com/tensorflow/tensorflow/pull/4796/files#diff-15609f78c0cd35bfdc31f9d8bd9b1e91R984 

there are failures like: 

https://ci.tensorflow.org/job/tensorflow-cl-sanity/14995/console

I tried various combinations of glob() and select() and none seem to work.

bssrdf:
Just want to follow up on this. I managed to build label_image example using the same procedure. Have to fix a bug at https://github.com/bssrdf/tensorflow/commit/58f93bd1d5cf47e7cffa63b59569233e5d2faa34#diff-f6254560279f78a9d74e66dfd2d85977 for it to build on win. But it failed to run due to some file path issue

$ bazel-bin/tensorflow/examples/label_image/label_image
E C:\tools\msys64\var\tmp\Bazel\s5nfxnhX\execroot\tensorflow\tensorflow\examples\label_image\main.cc:279] Not found: Failed to load compute graph at 'tensorflow/examples/label_image/data/tensorflow_inception_graph.pb'

although the file tensorflow_inception_graph.pb is present in that directory.

meteorcloudy:
@rohan100jain Hmm.. This is weird.
Here's what I tested:
I reset my branch to c20da148738df38c589fefc6287b64194b0ddbee, which is just before reverting this change.
And I ran `bazel build --nobuild tensorflow/...` on Mac, Linux and Windows. It all passed.

For `glob()` and `select()`, as long as you don't use `select` as arguments of `glob`, they will work just fine. So I'm not sure why it's failing like that on ci.

Can you give me some instructions to reproduce this error?

meteorcloudy:
@rohan100jain OK, I know what's wrong. After reviewed the internal change, It looks like a bad merge. The internal change is not the same as this PR. I will send a PR again.

meteorcloudy:
I believe there is nothing wrong with this PR, but I am not sure how to get it back. Another PR or revert the revert commit? @mrry

mrry:
I think the best way forward is another PR. Should be quick to review if it's the same as the last one :).

meteorcloudy:
OK, PR sent. https://github.com/tensorflow/tensorflow/pull/5009

PR #4800, opened 2016-10-06
Updated the links in docs to match the markdown sytnax.

haosdent:
Refer to [markdown syntax](https://daringfireball.net/projects/markdown/syntax#link),
the `[` should come after `)` without spaces or new line character. This fixs
the broken links when generate PDF/EPUB books by Gitbook.

### DIFF
--- a/tensorflow/contrib/session_bundle/README.md
+++ b/tensorflow/contrib/session_bundle/README.md
@@ -4,8 +4,8 @@
 
 ## Overview
 
-This document describes the data formats and layouts for exporting [TensorFlow]
-(https://www.tensorflow.org/) models for inference.
+This document describes the data formats and layouts for exporting
+[TensorFlow](https://www.tensorflow.org/) models for inference.
 
 These exports have the following properties:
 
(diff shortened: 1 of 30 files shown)


haosdent:
@terrytangyuan Sorry for the delay, may you help review this patch? Thank you in advance.

terrytangyuan:
Thanks! @tensorflow-jenkins Test this please.

terrytangyuan (approved):
LGTM

yifeif (approved):

rohan100jain:
For the changes in the python API, I believe it would make sense to update the actual py code instead of the docs since the docs are automatically generated.

haosdent:
Hi, @rohan100jain Thanks a lot for review! I change both shard documents in python API and python source code. e.g., For `Candidate Sampling Algorithms Reference`, I change `tensorflow/python/ops/nn.py` and `tensorflow/g3doc/api_docs/python/functions_and_classes/shard1/tf.nn.sampled_softmax_loss.md`

Should I need to revert anything?

rohan100jain:
Thats great! But I think you might have missed out some such as https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/ops.py#L2128
etc. 

Don't think you need to revert anything but what I'm saying is that if the .py files for all the ops are not changed this problem with re-occur because someone will regenerate and we'll get the broken links back again. So I guess it would be great if you can look through all the python API docs that were changed in this PR and see if there are corresponding .py documentation that might need to be changed as well.

haosdent:
@rohan100jain Thanks a lot! My bad use `\]\s\([#h]` to search it, I should use `\]\s+\([#h]` Would you like to help review my patch to fix this issue?

PR #4803, opened 2016-10-06
Adjusted buffer size to reduce memory usage and increase performance

petewarden:
This small tweak to the im2col buffer size shows improvements in some of our benchmark's latency, and reduces overall memory used.


yifeif:

yifeif (approved):

Issue #2630, opened 2016-06-03
Getting "readlink: illegal option -- f" when switching builds in MacOS


yaroslavvb:
I'm running into this error when building my targets with regular config, and after that trying to build with `config=cuda` in the same directory on latest MacOS (10.11.5) and bazel (0.2.3-homebrew). A work-around is to git-clone into a new directory, and build there with `config=cuda`
```

`/private/var/tmp/_bazel_yaroslavvb/1401dea20c857793b13cff5c293c0db6/tensorflow/'
____[8 / 88] Compiling external/protobuf/python/google/protobuf/internal/api_implementation.cc
____[8 / 165] Compiling external/re2/re2/filtered_re2.cc
____[8 / 313] Compiling external/boringssl_git/crypto/x509/x509_ext.c
____From Executing genrule //third_party/gpus/cuda:cuda_config_check:
readlink: illegal option -- f
usage: readlink [-n] [file ...]
ERROR: /Users/yaroslavvb/tfimmediate_fresh/tensorflow/third_party/gpus/cuda/BUILD:204:1: declared output 'third_party/gpus/cuda/cuda.config' was not created by genrule. This is probably because the genrule actually didn't create this output, or because the output was a directory and the genrule was run remotely (note that only the contents of declared file outputs are copied from genrules run remotely).
ERROR: /Users/yaroslavvb/tfimmediate_fresh/tensorflow/third_party/gpus/cuda/BUILD:204:1: not all outputs were created.
____[195 / 385] Compiling external/protobuf/src/google/protobuf/wrappers.pb.cc [for host]

```

aselle:
Hm, yes -f doesn't exist on the Mac OS readlink.  We should probably emulate the canonicalization this way...
http://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac
third_party/gpus/cuda/BUILD:181

yaroslavvb:
A side-effect of this is that if you make a mistake during "./configure" step, you must start over in a new directory. IE, running configure to change capability to "3.0" after setting it to default at first, doesn't have the desired effect, and "bazel clean" doesn't work because of this error

gunan:
@aselle any updates here?

rxwei:
On macOS you can use `greadlink` from homebrew'ed coreutils, which supports `-f`.

gunan:
@yifeif has a fix for this. Yifei, could you close this issue if your PR is merged?

[...]

PR #4806, opened 2016-10-06
Fix configure script readlink incompatibility on Darwin.

davidzchen:
#4699 introduced a change that broken the build on OS X since `readlink` works

differently.

This change fixes the `readlink` invocations so that they work on both platforms
and avoid failing the `configure` script if the path is not a symlink.

### DIFF
--- a/configure
+++ b/configure
@@ -216,9 +216,6 @@ while true; do
 
   if [[ -z "$TF_CUDNN_VERSION" ]]; then
     TF_CUDNN_EXT=""
-    # Resolve to the SONAME of the symlink.  Use readlink without -f
-    # to resolve exactly once to the SONAME.  E.g, libcudnn.so ->
-    # libcudnn.so.4
     cudnn_lib_path=""
     cudnn_alt_lib_path=""
     if [ "$OSNAME" == "Linux" ]; then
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -228,10 +225,15 @@ while true; do
...
       cudnn_alt_lib_path="${CUDNN_INSTALL_PATH}/libcudnn.dylib"
     fi
+    # Resolve to the SONAME of the symlink.  Use readlink without -f
+    # to resolve exactly once to the SONAME.  E.g, libcudnn.so ->
+    # libcudnn.so.4.
+    # If the path is not a symlink, readlink will exit with an error code, so
+    # in that case, we return the path itself.
     if [ -f "$cudnn_lib_path" ]; then
-      REALVAL=`readlink -f ${cudnn_lib_path}`
+      REALVAL=`readlink ${cudnn_lib_path} || echo "${cudnn_lib_path}"`
```
yifeif: Are we looking for the full path? echo seems to be not doing the same thing as readlink -f.

yifeif: I see we just want .so file name :)

yifeif (approved):

configure
davidzchen: Looking into this more, I now understand why the comment says that we only want to resolve one level.

On my installation, `libcudnn.so` resolves to `libcudnn.so.4`, which then resolves to `libcudnn.so.4.0.7`, and this code extracts the 4 from `libcudnn.so.4`. Though, the regex that we are using would still work if we resolve all the way to `libcudnn.so.4.0.7`.

In any case, let's merge this since `readlink -f` is invalid on macOS. Once Bazel 0.3.2 (see bazelbuild/bazel#1840) is released, I am planning to move this logic into `cuda_configure` and use the Skylark `realpath()` method, which implements `readlink -f` and works cross-platform.

yifeif:

yifeif:
This also fixes #2630. closes #2630.

PR #4819, opened 2016-10-07
Fix /WHOLEARCHIVE linking on Windows in Bazel build

meteorcloudy:
/WHOLEARCHIVE has a bug linking static library compiled from empty source file.

Fixed related source files by adding some include statements as placeholder.

This change will make users able to use Bazel to build C++ example trainer with Visual Studio 2015 update 2 (or newer verison). Older versions of VS works fine without this change because Bazel will directly link object files instead of using /WHOLEARCHIVE (it's not supported)
@mrry

### DIFF
--- a/tensorflow/core/kernels/fused_batch_norm_op.cu.cc
+++ b/tensorflow/core/kernels/fused_batch_norm_op.cu.cc
@@ -67,4 +67,8 @@ template class InvVarianceToVariance<float>;
 }  // namespace functor
 }  // namespace tensorflow
 
+#else
+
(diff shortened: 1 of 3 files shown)


mrry:

mrry (approved):
LGTM.

yifeif:
jenkins was having issues. @tensorflow-jenkins test this please.

yifeif:

meteorcloudy:
Not sure why it's still failing, can you test it again please?

rohan100jain:

Issue #4419, opened 2016-09-16
tf.get_variable without an explicit initializer fails for integer types


ethereon:
The following fails (shape and name are arbitrary):

```
tf.get_variable(name='foo', shape=(42,), dtype=tf.int32)
```

Exception: `TypeError: Expected int32, got -1.7320508075688772 of type 'float' instead.`

In contrast, using `tf.float32` works just fine.

The problem appears to be https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/variable_scope.py#L658

If an initializer is not provided (and a default one is not set), a uniform unit scaling init is used (notice that [`sqrt(3)==1.7320...`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/init_ops.py#L275)), which of course conflicts with the requested integer type.

While this can be mitigated by doing something like:

```
tf.get_variable(name='foo', dtype=tf.int32, initializer=tf.zeros_initializer(shape=(42,), dtype=tf.int32))
```

it feels like a smarter default behavior based on the variable type is warranted (or at least a less cryptic error).

Tested on the current master.

concretevitamin:
@lukaszkaiser - Hey Lukasz, are you the right person to look at this?

lukaszkaiser:
A smarter default behaviour would be nice, it's not too clear what it should be. I think we'd welcome a contribution from outside here.

DjangoPeng:
@lukaszkaiser I'm working to fix it

soloice:
Any progress on this? Still get unclear TypeError message for tf.int32 on Mac OS X, tf 1.0.0.

girving:
A better error message would be nice, but I don't think a default initializer is a good idea: integers frighten and confuse me.

PR #4826, opened 2016-10-07
Better support for initializing without explicit initializer(issue#4419)

DjangoPeng:
When using `tf.get_variable(name='foo', shape=(42,), dtype=tf.int32)` without explicit initializer, it causes the error:

```
TypeError: Expected int32, got -1.7320508075688772 of type 'float' instead.
```

The reason is as follows:
If no initializer provided, it would use the default initializer `uniform_unit_scaling_initializer` which just fit with `float`(notice that sqrt(3)==1.7320...). While, it always conflict with the required integer dtype.

The current solution is initialize the integer by `init_ops.zeros_initializer` when no initializer provided.

reference the [issus#4419](https://github.com/tensorflow/tensorflow/issues/4419)

### DIFF
--- a/tensorflow/python/framework/dtypes.py
+++ b/tensorflow/python/framework/dtypes.py
@@ -59,6 +59,7 @@ class DType(object):
   @@name
   @@base_dtype
   @@real_dtype
+  @@is_bool
   @@is_floating
   @@is_complex
   @@is_integer
@@ -140,6 +141,11 @@ class DType(object):
   def as_datatype_enum(self):
     """Returns a `types_pb2.DataType` enum value based on this `DType`."""
     return self._type_enum
+
+  @property
+  def is_bool(self):
+    """Returns whether this is a boolean data type"""
+    return self.base_dtype == bool
 
   @property
   def is_integer(self):
--- a/tensorflow/python/kernel_tests/variable_scope_test.py
+++ b/tensorflow/python/kernel_tests/variable_scope_test.py
@@ -22,6 +22,7 @@ import numpy
 import tensorflow as tf
 
 from tensorflow.python.framework import dtypes
+from tensorflow.python.ops import init_ops
 from tensorflow.python.ops import control_flow_ops
 from tensorflow.python.ops import variable_scope
 
@@ -94,6 +95,21 @@ class VariableScopeTest(tf.test.TestCase):
 
       with self.assertRaises(TypeError):
         tf.get_variable("x", initializer={})
+
+  def testInitFromNonInitializer(self):
+    with self.test_session() as sess:
+      # Test various dtypes with zeros initializer as following:
+      types = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64, 
+              tf.bool]
+
+      # Use different varibale_name to distinguish various dtypes
+      for (i, dtype) in enumerate(types):
+        x = tf.get_variable(name='x%d' % i, shape=(3, 4), dtype=dtype)
+        y = tf.get_variable(name='y%d' % i, dtype=dtype, 
+            initializer=init_ops.zeros_initializer(shape=(3, 4), dtype=dtype))
+        
+        tf.global_variables_initializer().run()
+        self.assertAllEqual(x.eval(), y.eval())
 
   def testVarScopeCachingDevice(self):
     with self.test_session():
@@ -671,6 +687,27 @@ def axis0_into3_partitioner(shape=None, **unused_kwargs):
 
 
 class VariableScopeWithPartitioningTest(tf.test.TestCase):
+
+  def testInitFromNonInitializer(self):
+    with self.test_session() as sess:
+      # Test various dtypes with zeros initializer as following:
+      types = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64, 
+              tf.bool]
+
+      # Use different varibale_name to distinguish various dtypes
+      for (i, dtype) in enumerate(types):
+        x = tf.get_variable(name='x%d' % i, shape=(3, 4), dtype=dtype,
+            partitioner=axis0_into2_partitioner)
+        y = tf.get_variable(name='y%d' % i, dtype=dtype, 
+            partitioner=axis0_into2_partitioner,
+            initializer=init_ops.zeros_initializer(shape=(6, 4), dtype=dtype))
+
+        tf.global_variables_initializer().run()
+        # x and y would become var list after partition
+        val_x = sess.run(list(x))
+        val_y = sess.run(list(y))
+
+        self.assertAllEqual(val_x, val_y)
 
   def testResultNameMatchesRequested(self):
     with tf.variable_scope("scope0", partitioner=axis0_into2_partitioner):
--- a/tensorflow/python/ops/variable_scope.py
+++ b/tensorflow/python/ops/variable_scope.py
@@ -525,9 +525,14 @@ class _VariableStore(object):
 
       var_full_name = "%s/part_%d" % (name, i)
       with ops.name_scope(var_full_name + "/PartitionedInitializer"):
+        # Create the tensor to initialize the variable with default value.
         if initializer is None:
-          init = init_ops.uniform_unit_scaling_initializer()
-          init_shape = var_shape
+          init, initializing_from_value = self._get_default_initializer(
+              name=name, shape=shape, dtype=dtype)
+          if initializing_from_value:
+            init_shape = None
+          else:
+            init_shape = var_shape
         elif callable(initializer):
           init = initializer
           init_shape = var_shape
@@ -652,9 +657,10 @@ class _VariableStore(object):
       raise ValueError("Shape of a new variable (%s) must be fully defined, "
                        "but instead was %s." % (name, shape))
 
-    # Create the tensor to initialize the variable.
+    # Create the tensor to initialize the variable with default value.
     if initializer is None:
-      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializer, initializing_from_value = self._get_default_initializer(
+          name=name, shape=shape, dtype=dtype)
     # Clear control dependencies while creating the initializer.
     with ops.control_dependencies(None):
       if initializing_from_value:
@@ -690,6 +696,39 @@ class _VariableStore(object):
           ops.add_to_collection(ops.GraphKeys.REGULARIZATION_LOSSES, loss)
 
     return v
+
+
+  # Initialize variable when no initializer provided
+  def _get_default_initializer(self, name, shape=None, dtype=dtypes.float32):
+    """Provide a default initializer and a corresponding value.
+
+    Args:
+      name: see get_variable.
+      shape: see get_variable.
+      dtype: see get_variable.
+
+    Returns:
+      initializer and initializing_from_value. See get_variable above.
+
+    Raises:
+      ValueError: When giving unsupported dtype.
+    """
+    # If dtype is DT_FLOAT, provide a uniform unit scaling initializer
+    if dtype.is_floating:
+      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializing_from_value = False
+    # If dtype is DT_INT/DT_UINT, provide a default value `zero`
+    # If dtype is DT_BOOL, provide a default value `FALSE`
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool:
+      initializer = init_ops.zeros_initializer(
+          shape=shape, dtype=dtype.base_dtype)
+      initializing_from_value = True
+    # NOTES:Do we need to support for handling DT_STRING and DT_COMPLEX here?
+    else:
+      raise ValueError("An initializer for variable %s of %s is required"
+          % (name, dtype.base_dtype))
+
+    return initializer, initializing_from_value
 
 
 # To stop regularization, use this regularizer


ebrevdo:
Perhaps we should just throw a cleaner error?  Initializer should be explicit imo

ebrevdo:
Nevermind. That's not backwards compatible.

DjangoPeng:
@ebrevdo But I think Tensorflow should provide the default initializer for integer, such as float. And in many platform and language, they have the build-in mechanism to handle the case

tensorflow/python/ops/variable_scope.py
```diff
@@ -652,6 +652,10 @@ def _get_single_variable(self,
       raise ValueError("Shape of a new variable (%s) must be fully defined, "
                        "but instead was %s." % (name, shape))
 
+    # Fill the integer array with zeros
+    if initializer is None and dtype.is_integer:
+      initializer = init_ops.zeros_initializer(shape=shape, dtype=dtype.base_dtype)
```
rohan100jain: Wouldn't it be better for this to be a function instead? In that case, for partitioned variables and such, there would be better support.

rohan100jain (changes requested):
Can you also add test cases in variable_scope_test.py to make sure that the issue is fixed?

DjangoPeng:
@rohan100jain Where should I define the function to handle the issue, init_op? or variable_scope?

DjangoPeng:
@rohan100jain  Please review

DjangoPeng:
@lukaszkaiser @rohan100jain @ebrevdo, @theweiho @keveman  Can one of you review the PR?

tensorflow/python/framework/dtypes.py
```diff
@@ -136,27 +138,37 @@ def as_datatype_enum(self):
...
 
   @property
   def is_floating(self):
     """Returns whether this is a (real) floating point type."""
     return issubclass(self.as_numpy_dtype, np.floating)
 
   @property
-  def is_complex(self):
-    """Returns whether this is a complex floating point type."""
-    return self.base_dtype in (complex64, complex128)
```
ebrevdo: can you avoid moving around methods?  revert this part of your change?

tensorflow/python/ops/array_ops.py
```diff
@@ -1152,6 +1152,37 @@ def matrix_transpose(a, name="matrix_transpose"):
...
+    shape: Either a list of integers, or a 1-D `Tensor` of type `string`.
+    dtype: The type of an element in the resulting `Tensor`.
+    name: A name for the operation (optional).
+
+  Returns:
+    A `Tensor` with all elements set to ''(empty string)
+  """
+  with ops.name_scope(name, 'strings', [shape]) as name:
+    string = ''
+    try:
```
ebrevdo: i dont think we need a dtype argument.

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,34 @@ def _get_single_variable(self,
...
+    Raises:
+      TypeError: When giving unsupported dtype.
+    """
+    if dtype.is_floating:
+      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializing_from_value = False
+    elif dtype.is_string:
+      initializer = init_ops.strings_initializer(shape=shape, dtype=dtype.base_dtype)
+      initializing_from_value=True
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool or dtype.is_complex:
```
ebrevdo: why do complex values get zeros initialized?  i dont think it'll even reach this point since complex types are is_floating.

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,34 @@ def _get_single_variable(self,
...
+    Returns:
+      initializer and initializing_from_value. See documentation of get_variable above.
+
+    Raises:
+      TypeError: When giving unsupported dtype.
+    """
+    if dtype.is_floating:
+      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializing_from_value = False
+    elif dtype.is_string:
```
ebrevdo: can we punt on string initialization till another PR?  i'm not in love with the tf.strings method and would prefer you just leave that out of this one.  keep it focused.

tensorflow/python/ops/variable_scope_test.py
```diff
@@ -0,0 +1,100 @@
+# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
ebrevdo: there is a variable scope test in kernel_tests; please put your tests there following the existing style of the file.

ebrevdo: (that includes using indentation and whitespace consistent with how we do it everywhere else)

ebrevdo (changes requested):

tensorflow/python/ops/variable_scope.py
DjangoPeng: Yep, complex is not tf.float, so I used `dtype.is_complex`, and the argument passed to `zeros_initializer` is `dtype.base_type=tf.complex64/128`. At the same time, initialize complex by zeros is reasonable in general.

tensorflow/python/ops/variable_scope_test.py
DjangoPeng: @rohan100jain So I don't need to add the test cases in `variable_scope_test.py`?

tensorflow/python/ops/variable_scope.py
DjangoPeng: I think tf.string is the same issue regarding initialization, so put them in one PR. If you insist that they are different, I can split them to reopen a PR for tf.string

tensorflow/python/ops/array_ops.py
DjangoPeng: Yep, you're right~ This function is just for `tf.string`, so it doesn't need a dtype argument

ebrevdo:
Please send a separate PR for tf.string, I think I'm not the best person to
review that (maybe andrew selle)

tensorflow/python/ops/variable_scope.py
ebrevdo: i'm not convinced that initializing complex with zeros is reasonable in general.  do you have some references showing neural networks that use complex parameters and suggest that initializing them with zeros is the right way to do it?

certainly having this be an undocumented feature is problematic; and since complex are complex floating point, it also seems very reasonable to initialize them like regular floating point.

tensorflow/python/ops/variable_scope_test.py
ebrevdo: you **need** to add test cases; but put them in kernel_tests/variable_scope_test; not in this new file.

tensorflow/python/ops/variable_scope.py
DjangoPeng: Yep, zeros initialization may not be the most appropriate if you just focus on NN. But I always    think tensorflow as a general Machine Learning platform, not only for NN, but also for other cases and solutions. As we can see the updated tutorials for [Non-ML Applications](https://www.tensorflow.org/versions/r0.11/tutorials/mandelbrot/index.html#mandelbrot-set), so I prefer to keep the initializer for complex.

Regarding zeros or sqrt(3), I think both are ok. Do you think so?

tensorflow/python/ops/variable_scope_test.py
DjangoPeng: @ebrevdo Gotcha~I'm on it.

tensorflow/python/framework/dtypes.py
DjangoPeng: Fine, I do the adjustment cause there are two assigners told me this is a custom of Google, and let me fix it in the previous PRs. So, this time, I adjusted them in passing, while it looks like it's not fit for everyone. I'll revert that~

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,31 @@ def _get_single_variable(self,
     return v
 
 
+  # Initialize variable when no initializer provided
+  def _get_default_initializer(self, shape=None, dtype=dtypes.float32):
+    """Use this function to provide a default and corresponding initializer of dtype
+
+    Args:
+      dtype: Type of the new or existing variable.(defaults to `DT_FLOAT`).
```
ebrevdo: s/variable ./variable /

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,31 @@ def _get_single_variable(self,
...
+      TypeError: When giving unsupported dtype.
+    """
+    if dtype.is_floating:
+      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializing_from_value = False
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool or dtype.is_complex:
+      initializer = init_ops.zeros_initializer(shape=shape, dtype=dtype.base_dtype)
+      initializing_from_value=True
+    else:
+      raise TypeError("You need to give an initializer for %s" % dtype)
```
ebrevdo: Raise a ValueError, and instead of dtype provide the name of the variable being initialized.  Otherwise it's not a very helpful error string.

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,31 @@ def _get_single_variable(self,
...
+    Returns:
+      initializer and initializing_from_value. See documentation of get_variable above.
+
+    Raises:
+      TypeError: When giving unsupported dtype.
+    """
+    if dtype.is_floating:
+      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializing_from_value = False
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool or dtype.is_complex:
```
ebrevdo: since the complex case is ambiguous enough for us to have a discussion about what value it should be, let's just fall back on raising an error in the complex case.

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,31 @@ def _get_single_variable(self,
...
+      initializer = init_ops.zeros_initializer(shape=shape, dtype=dtype.base_dtype)
+      initializing_from_value=True
```
ebrevdo: add spaces around the equality sign.

ebrevdo (changes requested):

tensorflow/python/ops/variable_scope.py
DjangoPeng: Yep, so we can handle the complex dtype here as well

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,31 @@ def _get_single_variable(self,
     return v
 
 
+  # Initialize variable when no initializer provided
+  def _get_default_initializer(self, shape=None, dtype=dtypes.float32):
+    """Use this function to provide a default and corresponding initializer of dtype
+
+    Args:
+      dtype: Type of the new or existing variable (defaults to `DT_FLOAT`).
+
```
DjangoPeng: Fixed

tensorflow/python/ops/variable_scope.py
```diff
@@ -690,6 +695,31 @@ def _get_single_variable(self,
...
+    """
+    if dtype.is_floating:
+      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializing_from_value = False
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool:
+      initializer = init_ops.zeros_initializer(shape=shape, dtype=dtype.base_dtype)
+      initializing_from_value = True
+    else:
+      raise ValueError("You need to give an appropriate initializer for %s" % dtype)
+
```
DjangoPeng: Raise ValueError, but I really don't know how to give an appropriate error string. 
:(

ebrevdo: Can't you pass name directly through to this function so you can use it in the error message?

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -655,6 +672,28 @@ def axis0_into3_partitioner(shape=None, **unused_kwargs):
...
 
+  def testInitFromNonInitializer(self):
+    with self.test_session() as sess:
+      # Test various dtypes with zeros initializer as following:
+      TYPES = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64, 
+              tf.bool, tf.complex64, tf.complex128]
+      i = 0
+      # Use different varibale_name to distinguish various dtypes
+      for dtype in TYPES:
+        x = tf.get_variable(name='x'+str(i), shape=(3,4), dtype=dtype,
```
ebrevdo: 'x%d' % i

here and elsewhere

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -655,6 +672,28 @@ def axis0_into3_partitioner(shape=None, **unused_kwargs):
...
+                partitioner=axis0_into2_partitioner(shape=(3,4)))
+        y = tf.get_variable(name='y'+str(i), dtype=dtype, 
+                partitioner=axis0_into2_partitioner(shape=(3,4)),
+                initializer=init_ops.zeros_initializer(shape=(3,4), dtype=dtype))
+
+        sess.run(tf.initialize_all_variables())
+        # x and y become var list after partition
+        val_x = sess.run(x._get_variable_list())
```
ebrevdo: don't use _get_variable_list; which is private!

use list(x)

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -680,16 +681,16 @@ def testInitFromNonInitializer(self):
...
+                partitioner=axis0_into2_partitioner),
+                initializer=init_ops.zeros_initializer(shape=(6,4), dtype=dtype))
 
         sess.run(tf.initialize_all_variables())
-        # x and y become var list after partition
-        val_x = sess.run(x._get_variable_list())
-        val_y = sess.run(y._get_variable_list())
+        # x and y would become var list after partition
+        val_x = sess.run(list(x))
+        val_y = sess.run(list(y))
```
DjangoPeng: Replace private function `_get_variable_list()` with `list()`

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -680,16 +681,16 @@ def testInitFromNonInitializer(self):
...
       # Use different varibale_name to distinguish various dtypes
       for dtype in TYPES:
-        x = tf.get_variable(name='x'+str(i), shape=(3,4), dtype=dtype,
-                partitioner=axis0_into2_partitioner(shape=(3,4)))
-        y = tf.get_variable(name='y'+str(i), dtype=dtype, 
-                partitioner=axis0_into2_partitioner(shape=(3,4)),
-                initializer=init_ops.zeros_initializer(shape=(3,4), dtype=dtype))
+        x = tf.get_variable(name='x%d' % i, shape=(3,4), dtype=dtype,
+                partitioner=axis0_into2_partitioner)
+        y = tf.get_variable(name='y%d' % i, dtype=dtype,
```
DjangoPeng: Set string format to `'x%d' % i`

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -680,16 +681,16 @@ def testInitFromNonInitializer(self):
...
-        x = tf.get_variable(name='x'+str(i), shape=(3,4), dtype=dtype,
-                partitioner=axis0_into2_partitioner(shape=(3,4)))
-        y = tf.get_variable(name='y'+str(i), dtype=dtype, 
-                partitioner=axis0_into2_partitioner(shape=(3,4)),
-                initializer=init_ops.zeros_initializer(shape=(3,4), dtype=dtype))
+        x = tf.get_variable(name='x%d' % i, shape=(3,4), dtype=dtype,
+                partitioner=axis0_into2_partitioner)
+        y = tf.get_variable(name='y%d' % i, dtype=dtype, 
+                partitioner=axis0_into2_partitioner),
+                initializer=init_ops.zeros_initializer(shape=(6,4), dtype=dtype))
```
DjangoPeng: Because using `axis0_into2_partitioner`,  `y` is required a replica shape in 1-D of `x`.

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -682,10 +682,10 @@ def testInitFromNonInitializer(self):
       # Use different varibale_name to distinguish various dtypes
       for dtype in TYPES:
         x = tf.get_variable(name='x%d' % i, shape=(3,4), dtype=dtype,
-                partitioner=axis0_into2_partitioner)
+            partitioner=axis0_into2_partitioner)
         y = tf.get_variable(name='y%d' % i, dtype=dtype, 
-                partitioner=axis0_into2_partitioner),
-                initializer=init_ops.zeros_initializer(shape=(6,4), dtype=dtype))
+            partitioner=axis0_into2_partitioner),
+            initializer=init_ops.zeros_initializer(shape=(6,4), dtype=dtype))
```
DjangoPeng: Delete redundant spaces here and elsewhere.

DjangoPeng:
@ebrevdo Please review.

drpngx:
Jenkins, test this please.

DjangoPeng:
@drpngx @ebrevdo It seems like everything goes well. ⊙▽⊙

drpngx:
Did you address all changes?

DjangoPeng:
yep，including necessary comments

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -94,6 +94,23 @@ def testInitFromNonTensorValue(self):
       with self.assertRaises(TypeError):
         tf.get_variable("x", initializer={})
 
+  def testInitFromNonInitializer(self):
+    with self.test_session() as sess:
+      # Test various dtypes with zeros initializer as following:
+      TYPES = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64,
```
ebrevdo: capital letters only allowed for module-level static constants.  this should be just 'types'.

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -94,6 +94,23 @@ def testInitFromNonTensorValue(self):
...
 
+  def testInitFromNonInitializer(self):
+    with self.test_session() as sess:
+      # Test various dtypes with zeros initializer as following:
+      TYPES = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64, 
+              tf.bool]
+
+      i = 0
+      # Use different varibale_name to distinguish various dtypes
+      for dtype in TYPES:
```
ebrevdo: for (i, dtype) in enumerate(types):

tensorflow/python/ops/variable_scope.py
```diff
@@ -652,9 +656,10 @@ def _get_single_variable(self,
       raise ValueError("Shape of a new variable (%s) must be fully defined, "
                        "but instead was %s." % (name, shape))
 
+
     # Create the tensor to initialize the variable.
     if initializer is None:
-      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializer, initializing_from_value = self._get_default_initializer(name=name, shape=shape, dtype=dtype)
```
ebrevdo: i can't tell but is this within the 80 character limit?

ebrevdo:
More comments.

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -97,19 +97,17 @@ def testInitFromNonTensorValue(self):
   def testInitFromNonInitializer(self):
     with self.test_session() as sess:
       # Test various dtypes with zeros initializer as following:
-      TYPES = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64, 
+      types = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64,
```
DjangoPeng: Fix capital here and eleswhere.

tensorflow/python/ops/variable_scope.py
```diff
@@ -527,7 +527,8 @@ def _get_partitioned_variable(
       with ops.name_scope(var_full_name + "/PartitionedInitializer"):
         # Create the tensor to initialize the variable.
         if initializer is None:
-          init, initializing_from_value = self._get_default_initializer(name=name, shape=shape, dtype=dtype)
+          init, initializing_from_value = self._get_default_initializer(
+              name=name, shape=shape, dtype=dtype)
```
DjangoPeng: Fix typesetting for 80 chars limit here and elsewhere.

DjangoPeng:
@ebrevdo Could you review and merge this PR, please?

DjangoPeng:
@ebrevdo Would you mind testing the PR?

drpngx:
Jenkins, test this please.

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -94,6 +94,21 @@ def testInitFromNonTensorValue(self):
...
 
+  def testInitFromNonInitializer(self):
+    with self.test_session() as sess:
+      # Test various dtypes with zeros initializer as following:
+      types = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64, 
+              tf.bool]
+
+      # Use different varibale_name to distinguish various dtypes
+      for (i, dtype) in enumerate(types):
+        x = tf.get_variable(name='x%d' % i, shape=(3,4), dtype=dtype)
```
ebrevdo: whitespace: (3,4) should be (3,<space>4)

here and everywhere else.

ebrevdo (changes requested):

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -94,6 +94,21 @@ def testInitFromNonTensorValue(self):
...
 
+  def testInitFromNonInitializer(self):
+    with self.test_session() as sess:
+      # Test various dtypes with zeros initializer as following:
+      types = [tf.int8, tf.uint8, tf.int16, tf.uint16, tf.int32, tf.int64, 
+              tf.bool]
+
+      # Use different varibale_name to distinguish various dtypes
+      for (i, dtype) in enumerate(types):
+        x = tf.get_variable(name='x%d' % i, shape=(3, 4), dtype=dtype)
```
DjangoPeng: fixed typesetting here and elsewhere

DjangoPeng:
@ebrevdo Please check.

DjangoPeng:
@ebrevdo @drpngx  I revert all commits before and re-push the branch with more informative comments. Now, it would be better to review and merge. 
## Change log:
### dtypes.py
Add `is_bool()`  function for `if` judgement in `_get_default_initializer`. And `DT_BOOL` can use `zeros_initializer` to initialize with `FALSE`, so I combine it with `DT_INT` and `DT_UINT` to avoid duplicating code block.

### variable_scope.py
Add `_get_default_initializer()` private function to provide a default initializer and a corresponding value when no explicit initializer provided. Now, it can handle `DT_FLOAT/DT_INT/DT_UINT/DT_BOOL`. In other dtype case, it would raise a `ValueError` with info `\"An initializer for variable %s is required\" % name`.

### variable_scope_test.py
Add `testInitFromNonInitializer()` for single_variable and partitioned_variable.

tensorflow/python/ops/variable_scope.py
```diff
@@ -692,6 +698,38 @@ def _get_single_variable(self,
     return v
 
 
+  # Initialize variable when no initializer provided
+  def _get_default_initializer(self, name, shape=None, dtype=dtypes.float32):
+    """Provide a default initializer and a corresponding value
```
vrv: End one-line docstrings with a period

tensorflow/python/ops/variable_scope.py
```diff
@@ -692,6 +698,38 @@ def _get_single_variable(self,
...
+      initializing_from_value = False
+    # If dtype is DT_INT/DT_UINT, provide a default value `zero`
+    # If dtype is DT_BOOL, provide a default value `FALSE`
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool:
+      initializer = init_ops.zeros_initializer(
+          shape=shape, dtype=dtype.base_dtype)
+      initializing_from_value = True
+    # NOTES:Do we need to support for handling DT_STRING and DT_COMPLEX here?
+    else:
+      raise ValueError("An initializer for variable %s is required" % name)
```
vrv: maybe print the type name so they know why it doesn't have a default initializer

tensorflow/python/ops/variable_scope.py
```diff
@@ -692,6 +698,39 @@ def _get_single_variable(self,
     return v
 
 
+  # Initialize variable when no initializer provided
+  def _get_default_initializer(self, name, shape=None, dtype=dtypes.float32):
+    """Provide a default initializer and a corresponding value.
```
DjangoPeng: Give a period in end-line.

tensorflow/python/ops/variable_scope.py
```diff
@@ -692,6 +698,39 @@ def _get_single_variable(self,
...
+    # If dtype is DT_INT/DT_UINT, provide a default value `zero`
+    # If dtype is DT_BOOL, provide a default value `FALSE`
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool:
+      initializer = init_ops.zeros_initializer(
+          shape=shape, dtype=dtype.base_dtype)
+      initializing_from_value = True
+    # NOTES:Do we need to support for handling DT_STRING and DT_COMPLEX here?
+    else:
+      raise ValueError("An initializer for variable %s of %s is required"
+          % (name, dtype.base_dtype)
```
DjangoPeng: Give dtype name in `ValueError` for more info to debug.

DjangoPeng:
@vrv Please review.

vrv:
Running a test but going to let @ebrevdo and @rohan100jain give the final approval.

DjangoPeng:
@vrv It seems like there are some internal errors?

tensorflow/python/ops/variable_scope.py
```diff
@@ -726,7 +726,7 @@ def _get_default_initializer(self, name, shape=None, dtype=dtypes.float32):
     # NOTES:Do we need to support for handling DT_STRING and DT_COMPLEX here?
     else:
       raise ValueError("An initializer for variable %s of %s is required"
-          % (name, dtype.base_dtype)
+          % (name, dtype.base_dtype))
```
DjangoPeng: I found the reason of testing failure. Fixed.

DjangoPeng (commented):
@vrv Please test again, thanks!

vrv:

vrv:
@rohan100jain @ebrevdo does this look good?  I ca merge if so.

tensorflow/python/ops/variable_scope.py
```diff
@@ -692,6 +698,39 @@ def _get_single_variable(self,
...
+  def _get_default_initializer(self, name, shape=None, dtype=dtypes.float32):
+    """Provide a default initializer and a corresponding value.
+
+    Args:
+      name: see get_variable.
+      shape: see get_variable.
+      dtype: see get_variable.
+
+    Returns:
+      initializer and initializing_from_value. See get_variable above.
```
rohan100jain: I think initializing_from_value isn't documented in get_variable. So perhaps clarify that here?

tensorflow/python/ops/variable_scope.py
```diff
@@ -692,6 +698,39 @@ def _get_single_variable(self,
...
+      ValueError: When giving unsupported dtype.
+    """
+    # If dtype is DT_FLOAT, provide a uniform unit scaling initializer
+    if dtype.is_floating:
+      initializer = init_ops.uniform_unit_scaling_initializer()
+      initializing_from_value = False
+    # If dtype is DT_INT/DT_UINT, provide a default value `zero`
+    # If dtype is DT_BOOL, provide a default value `FALSE`
+    elif dtype.is_integer or dtype.is_unsigned or dtype.is_bool:
+      initializer = init_ops.zeros_initializer(
```
rohan100jain: vrv@, ebrevdo@ - does it make a difference whether this is a function or a Tensor? Or performance / memory wise its all the same?

tensorflow/python/ops/variable_scope.py
DjangoPeng: yep, `initializing_from_value` isn't documented in `get_variable`. While, `get_variable` calls `_get_single_variable` and `_get_partitioned_variable`, and both them define it. From my perspective, the comment is a little unclear, but I think the name itself is easy to understand. Do we really need to clarify here? instead of above?

tensorflow/python/ops/variable_scope.py
DjangoPeng: If we want to use a function to handle this case, we have to define a set of functions to cooperate with it. While it would only run once and then copy value to the variable, which is not expensive. Do you think so?

DjangoPeng:
@ebrevdo @rohan100jain @vrv @drpngx Can one of you verify the patch?

danmane:
@DjangoPeng What's the status here - should I poke rohan/ebrevdo to re-review?
Also, it now has conflicts, so you will need to rebase.

DjangoPeng:
@danmane I think it's ready to merge, and I will rebase it after work today. I hope them can accept the PR, which is helpful from my perspective.

tensorflow/python/ops/variable_scope.py
ebrevdo: i believe zeros_initializer was recently converted to a function instead of a tensor.

ebrevdo:
once you merge we can run a test.

ebrevdo (approved):

DjangoPeng:
@danmane @ebrevdo I have solved the conflict and update `tf.global_variables_initializer()` API. Please check and test.

drpngx:
Jenkins, test this please.

DjangoPeng:
@ebrevdo @drpngx @danmane It looks everything goes well, would you mind merging it?

Issue #4722, opened 2016-10-03
einsum not fully implemented


xuancong84:
I am glad to see the newly added einsum function. The documentation claims that its usage is the same as numpy. However, it can do almost nothing as compared to numpy. For example, it only supports subscripts in the form of '_->_'. Unfortunately, even matrix transpose does not work, i.e., 'ij->ji'. 
Numpy works:

```
>>> A
array([[ 0.3828997 , -0.39114848, -0.09727838, -0.20430113],
       [ 0.48020577, -0.47122706,  0.42830791,  0.25665744],
       [-0.30885863,  0.21669025,  0.31648793,  0.22417514],
       [ 0.32505724,  0.30478035,  0.48655034,  0.20040547]])
>>> einsum('ij->ji',A)
array([[ 0.3828997 ,  0.48020577, -0.30885863,  0.32505724],
       [-0.39114848, -0.47122706,  0.21669025,  0.30478035],
       [-0.09727838,  0.42830791,  0.31648793,  0.48655034],
       [-0.20430113,  0.25665744,  0.22417514,  0.20040547]])
```

Tensorflow does not work:

```
pseudo-code:
M_ = tf.Variable(tf.random_normal([4,4]))
N_ = tf.einsum('ij->ji',M_)              
print [M_, N_]

output:
[array([[ 0.80474716, -1.38590837, -0.3379252 , -1.24965811],
       [ 2.57852983,  0.05492432,  0.23039417, -0.74263287],
       [-2.42627382,  1.70774114,  1.19503212,  0.43006262],
       [-1.04652011, -0.32753903, -1.26430523,  0.8810069 ]], dtype=float32), 
array([[ 0.80474716, -1.38590837, -0.3379252 , -1.24965811],
       [ 2.57852983,  0.05492432,  0.23039417, -0.74263287],
       [-2.42627382,  1.70774114,  1.19503212,  0.43006262],
       [-1.04652011, -0.32753903, -1.26430523,  0.8810069 ]], dtype=float32)]
```

I want to multiply a matrix with every frame vector in every batch. Or similar operations which can be done by a simple tensor product. It seems that I still have to duplicate the matrix so many times and perform a batch_matmul, which is very inconvenient and slow and memory consuming.

I suggest tensorflow to implement either the tensordot or einsum function which can perform tensor product.

It is quite a shame that tensorflow cannot even perform basic tensor product so far :(

tatatodd:
[...]

PR #4837, opened 2016-10-08
updates to einsum

nikhilmishra000:
Resolving https://github.com/tensorflow/tensorflow/issues/4722:
-- Fix bug where output indices are expected to be in alphabetical order
-- Add to docstring

### DIFF
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -96,17 +96,24 @@ def einsum(axes, *inputs):
 
   Like `numpy.einsum`, but does not support:
   * Ellipses (subscripts like `ij...,jk...->ik...`)
-  * Subscripts that reduce to scalars
+  * Subscripts where an axis appears twice within a single input (e.g. 'ijj,jk->ik').
 
-  Raises a ValueError if:
-  * The format of `axes` is incorrect.
-  * The number of inputs implied by `axes` does not match `len(inputs)`.
-  * An axis appears in the output subscripts but not in any of the inputs.
-  * The number of dimensions of an input differs from the number of indices in its subscript.
-  * The input shapes are inconsistent along a particular axis.
+  Args:
+    axes: a `str` describing the contraction, in the same format as `numpy.einsum`.
+    inputs: the `Tensor`s to contract, whose shapes should be consistent with `axes`.
+
+  Returns:
+    The contracted `Tensor`, with shape determined by `axes`.
+
+  Raises:
+    ValueError: If the format of `axes` is incorrect,
+                or the number of inputs implied by `axes` does not match `len(inputs)`,
+                or an axis appears in the output subscripts but not in any of the inputs,
+                or the number of dimensions of an input differs from the number of indices in its subscript,
+                or the input shapes are inconsistent along a particular axis.
   """
 
-  match = re.match('([a-z,]+)->([a-z]+)', axes)
+  match = re.match('([a-z,]+)(->[a-z]*)?', axes)
   if not match:
     raise ValueError(
       "Indices have incorrect format: %s" % axes
@@ -114,13 +121,27 @@ def einsum(axes, *inputs):
 
   inputs = list(inputs)
   idx_in = match.group(1).split(',')
-  idx_out = match.group(2)
   idx_all = set(''.join(idx_in))
   indices = str(idx_all)
 
+  if match.group(2):
+    idx_out = match.group(2)[2:]
+
+  else:
+    # infer the output subscripts if not given, assume alphabetical order
+    counts = {ax: 0 for ax in indices}
+    for axes in idx_in:
+      for ax in axes:
+        counts[ax] += 1
+
+    idx_out = ''.join(sorted(
+      ax for ax in indices
+      if counts[ax] == 1
+    ))
+
   if len(idx_in) != len(inputs):
     raise ValueError(
-      "Expected %d inputs but only got %d" % (len(idx_in), len(inputs))
+      "Expected %d inputs but got %d" % (len(idx_in), len(inputs))
     )
 
   missing_idx = set(idx_out).difference(idx_all)
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -94,7 +94,9 @@ def einsum(axes, *inputs):
   """
   A generalized contraction between tensors of arbitrary dimension.
 
-  Like numpy.einsum.
+  Like numpy.einsum, but does not support:
+  -- ellipses (subscripts like 'ij...,jk...->ik...')
+  -- subscripts that reduce to scalars
   """
 
   match = re.match('([a-z,]+)->([a-z]+)', axes)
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/special_math_ops.py
```diff
@@ -94,7 +94,9 @@ def einsum(axes, *inputs):
   """
   A generalized contraction between tensors of arbitrary dimension.
 
-  Like numpy.einsum.
+  Like numpy.einsum, but does not support:
+  -- ellipses (subscripts like 'ij...,jk...->ik...')
```
martinwicke: Can you turn those apostrophes into backticks? Then it'll render nicely as code. Also, please put `numpy.einsum` into backticks. All docstrings get rendered as markdown.

tensorflow/python/ops/special_math_ops.py
```diff
@@ -105,37 +107,43 @@ def einsum(axes, *inputs):
...
-
+  indices = str(idx_all)
 
   assert len(idx_in) == len(inputs), \
     "Expected %d inputs but only got %d" % (len(idx_in), len(inputs))
 
-  # transpose inputs so axes are in alphabetical order
+  missing_idx = set(idx_out).difference(idx_all)
+  assert not missing_idx, \
+    "Unknown ouput axes: %s" % missing_idx
```
martinwicke: Can you make this into an exception (ValueError is fine), and add it to the docstring?

martinwicke (changes requested):

tensorflow/python/ops/special_math_ops.py
nikhilmishra000: @martinwicke should I do the same for all of the other assertions too?

tensorflow/python/ops/special_math_ops.py
```diff
@@ -94,27 +94,40 @@ def einsum(axes, *inputs):
   """
   A generalized contraction between tensors of arbitrary dimension.
 
-  Like numpy.einsum, but does not support:
-  -- ellipses (subscripts like 'ij...,jk...->ik...')
-  -- subscripts that reduce to scalars
+  Like `numpy.einsum`, but does not support:
+  * Ellipses (subscripts like `ij...,jk...->ik...`)
+  * Subscripts that reduce to scalars
+
```
martinwicke: Can you add an "Args:" section here, and after the args section, add a "Returns:" section. See other op docstrings for examples.

tensorflow/python/ops/special_math_ops.py
```diff
@@ -94,27 +94,40 @@ def einsum(axes, *inputs):
...
+  Raises a ValueError if:
```
martinwicke: Can you reformat this to
```
Raises:
  ValueError: <all your text, reindented to flow after ValueError>
```

tensorflow/python/ops/special_math_ops_test.py
```diff
@@ -176,7 +180,7 @@ def test_invalid(self):
 
       try:
         result = tf.einsum(axes, *inputs)
-      except AssertionError as e:
+      except ValueError as e:
         print(e)
       assert result is None, \
         "An exception should have been thrown."
```
martinwicke: Can you rewrite this to use self.assertRaises?

tensorflow/python/ops/special_math_ops_test.py
```diff
@@ -190,7 +194,7 @@ def test_dim_mismatch(self):
       result = None
       try:
         result = tf.einsum(axes, *inputs)
-      except AssertionError:
+      except ValueError:
         pass
       assert result is None, "An exception should have been thrown."
```
martinwicke: This too

martinwicke (commented):
Sorry, missed some things earlier.

jart:
Einsum is not a trivial abstraction. Numpy has 900 lines of test code for einsum ([test_einsum.py](https://github.com/numpy/numpy/blob/master/numpy/core/tests/test_einsum.py)) and 100 lines of Pydoc tests ([einsumfunc.py](https://github.com/numpy/numpy/blob/master/numpy/core/einsumfunc.py)). Ideally we would just use their test cases to show that the behavior is identical. Especially if we document the function as being "like numpy.einsum."

Thank you for the work you put into bringing einsum to TensorFlow @nikhilmishra000 and for being responsive in addressing the concerns in #4722. I just can't help but wonder if this function is more appropriate for the contrib package, since it isn't on par with numpy in terms of functionality and test comprehensiveness. What do you think @martinwicke?

See also: #4886

martinwicke:
I think it is fine in core. There's no experimentation with the interface needed, which is usually the concern. If its functionality is not a complete mirror of numpy's we have to document the differences clearly. 

In this case, it should be noted that transpose doesn't work, are there other differences that are not documented yet?

nikhilmishra000:
The docstring has been updated, and I have added some more tests.

Following this PR, I think the outstanding differences from numpy will be that:
* Ellipses will not be recognized. This shouldn't be too hard to implement, but I don't know that I have the bandwidth to get it done in the near future.
* Subscripts where an axis appears more than once (example: `ii->i` computes the trace of a matrix). Adding support for this seems a little bit more involved.

Exceptions are raised for both of these cases.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

nikhilmishra000:
The previous failure in python3 was my fault (I didn't realize that passing a comparator to `sorted()` had been disallowed). Not too sure what happened on linux gpu, but it seems unrelated?

martinwicke:
Jenkins, test this please!

tensorflow/python/ops/special_math_ops.py
```diff
@@ -94,48 +94,100 @@ def einsum(axes, *inputs):
...
       inputs[i] = array_ops.transpose(input_, permuted)
       idx_in[i] = sorted_idx
 
-  missing_idx = set(idx_out).difference(idx_all)
-  assert not missing_idx, \
-    "Unknown ouput axes: %s" % missing_idx
-
   reduction_idx = []
   shapes = [[dim if dim else -1
              for dim in tensor.get_shape().as_list()]
```
eaplatanios: I would like to propose one more fix. I think this should be changed from:

```
shapes = [[dim if dim else -1
           for dim in tensor.get_shape().as_list()]
           for tensor in inputs]
```

to:

```
shapes = [[dim if dim is not None else -1
           for dim in array_ops.unpack(array_ops.shape(tensor))]
           for tensor in inputs]
```

in order to allow for dynamic tensor shapes. I run into an issue while using `tf.einsum` and this change fixed it for me.

yunita15 (commented):
Good

tensorflow/python/ops/special_math_ops.py
eaplatanios: Sounds good! I made PR #5174 but without any test coverage. I have a paper deadline approaching and I'm not sure about your test conventions, so I thought I'd leave it there for now. This change alone seems to be working fine for me, although I haven't tried using it with any edge cases.

PR #4843, opened 2016-10-08
Add PIL formats(gif,png...) support for retrain.py

elviswf:
- Add PIL formats(gif,png...) support for retrain.py.
- Use PIL.Image to read images and io.BytesIO to serialize them.

### DIFF
--- a/tensorflow/examples/image_retraining/retrain.py
+++ b/tensorflow/examples/image_retraining/retrain.py
@@ -14,37 +14,51 @@
 # ==============================================================================
 """Simple transfer learning with an Inception v3 architecture model which
 displays summaries in TensorBoard.
+
 This example shows how to take a Inception v3 architecture model trained on
 ImageNet images, and train a new top layer that can recognize other classes of
 images.
+
 The top layer receives as input a 2048-dimensional vector for each image. We
 train a softmax layer on top of this representation. Assuming the softmax layer
 contains N labels, this corresponds to learning N + 2048*N model parameters
 corresponding to the learned biases and weights.
+
 Here's an example, which assumes you have a folder containing class-named
 subfolders, each full of images for each label. The example folder flower_photos
 should have a structure like this:
+
 ~/flower_photos/daisy/photo1.jpg
 ~/flower_photos/daisy/photo2.jpg
 ...
 ~/flower_photos/rose/anotherphoto77.jpg
 ...
 ~/flower_photos/sunflower/somepicture.jpg
+
 The subfolder names are important, since they define what label is applied to
 each image, but the filenames themselves don't matter. Once your images are
 prepared, you can run the training with a command like this:
+
 bazel build third_party/tensorflow/examples/image_retraining:retrain && \
 bazel-bin/third_party/tensorflow/examples/image_retraining/retrain \
 --image_dir ~/flower_photos
+
 You can replace the image_dir argument with any folder containing subfolders of
 images. The label for each image is taken from the name of the subfolder it's
 in.
+
 This produces a new model file that can be loaded and run by any TensorFlow
 program, for example the label_image sample code.
+
+
 To use with TensorBoard:
+
 By default, this script will log summaries to /tmp/retrain_logs directory
+
 Visualize the summaries with this command:
+
 tensorboard --logdir /tmp/retrain_logs
+
 """
 from __future__ import absolute_import
 from __future__ import division
@@ -158,13 +172,16 @@ MAX_NUM_IMAGES_PER_CLASS = 2 ** 27 - 1  # ~134M
 
 def create_image_lists(image_dir, testing_percentage, validation_percentage):
   """Builds a list of training images from the file system.
+
   Analyzes the sub folders in the image directory, splits them into stable
   training, testing, and validation sets, and returns a data structure
   describing the lists of images for each label and their paths.
+
   Args:
     image_dir: String path to a folder containing subfolders of images.
     testing_percentage: Integer percentage of the images to reserve for tests.
     validation_percentage: Integer percentage of images reserved for validation.
+
   Returns:
     A dictionary containing an entry for each label subfolder, with images split
     into training, testing, and validation sets within each label.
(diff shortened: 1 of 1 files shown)


elviswf:
**I signed it!**
Thank you!

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -73,6 +73,8 @@
 import sys
 import tarfile
 
+from PIL import Image
```
rohan100jain: Might want to add in the comments above that PIL would need to be installed in order to be able to run this.

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -450,7 +452,14 @@ def get_or_create_bottleneck(sess, image_lists, label_name, index, image_dir,
                                 category)
     if not gfile.Exists(image_path):
       tf.logging.fatal('File does not exist %s', image_path)
-    image_data = gfile.FastGFile(image_path, 'rb').read()
+    jpgext = ['jpg', 'jpeg', 'JPG', 'JPEG']
+    if (image_path.split('.')[-1] in jpgext):
```
rohan100jain: Use os.path.splitext() instead to figure out extension?

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -450,7 +452,14 @@ def get_or_create_bottleneck(sess, image_lists, label_name, index, image_dir,
                                 category)
     if not gfile.Exists(image_path):
       tf.logging.fatal('File does not exist %s', image_path)
-    image_data = gfile.FastGFile(image_path, 'rb').read()
+    jpgext = ['jpg', 'jpeg', 'JPG', 'JPEG']
```
rohan100jain: Make this a separate helper method that takes in file_name and returns back image_data.

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -195,7 +197,7 @@ def create_image_lists(image_dir, testing_percentage, validation_percentage):
     if is_root_dir:
       is_root_dir = False
       continue
-    extensions = ['jpg', 'jpeg', 'JPG', 'JPEG']
+    extensions = ['jpg', 'jpeg', 'JPG', 'JPEG', 'gif', 'png']
```
rohan100jain: what about GIF and PNG (caps)

Maybe share this list between here and below? Might also not be a bad idea to make it a dictionary mapping extension to color depth (https://en.wikipedia.org/wiki/Color_depth). That way you can handle the conversion more generally...

rohan100jain (changes requested):

tensorflow/examples/image_retraining/retrain.py
elviswf: Sure. Thank you!

tensorflow/examples/image_retraining/retrain.py
elviswf: I have tested these formats. So it goes like:
`extensions = ['jpg', 'jpeg', 'JPG', 'JPEG', 'gif', 'GIF', 'png', 'PNG', 'bmp', 'BMP']`

tensorflow/examples/image_retraining/retrain.py
elviswf: Sure. I will try it.

tensorflow/examples/image_retraining/retrain.py
elviswf: os.path.splitext()  returns '.jpg', so it has to be set [1:] to return 'jpg' like:
`os.path.splitext(image_path)[1][1:]` 
I think `image_path.split('.')[-1]` might be  clear enough.

elviswf: @rohan100jain @petewarden I have updated the changes. Please check it. Thank you!

elviswf:
@petewarden I have updated the code according to the reviews. Thank you!

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -464,6 +469,29 @@ def get_or_create_bottleneck(sess, image_lists, label_name, index, image_dir,
   return bottleneck_values
 
 
+def read_image2RGBbytes(image_path):
+  """Read the image into bytes stream.
```
drpngx: Minor nit: Reads (as per PEP8). Also, "into a bytes stream".

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -464,6 +469,29 @@ def get_or_create_bottleneck(sess, image_lists, label_name, index, image_dir,
...
+
+   For jpg/jpeg, gfile.FastGFile is enough.
```
drpngx: Minor nit, this line and the next: remove one leading space.

drpngx (changes requested):

drpngx:
Jenkins, test this please.

elviswf:
` Log @: /workspace/pip_test/tests/logs/tensorflow/examples/image_retraining/retrain_test.py.log

============== BEGINS failure log content ==============
  File "/workspace/pip_test/tests/retrain_test.py", line 23, in <module>
    from tensorflow.examples.image_retraining import retrain
  File "/workspace/pip_test/venv/local/lib/python2.7/site-packages/tensorflow/examples/image_retraining/retrain.py", line 81, in <module>
    from PIL import Image
ImportError: No module named PIL

============== ENDS failure log content ==============`

It shows `No module named PIL`. I wonder if there is a way to avoid using PIL package.

vrv:
@petewarden @gunan Should we install PIL as part of our CI infrastructure and add this as a dependency?

Part of me thinks that, serving as an example, this is not going to do everything that people want to do, so perhaps we leave a comment instead suggesting other ways to read images instead.

gunan:
I am not a fan of adding the dependency, as I think everyone will have their own favorite ways of reading images. And PIL is not a real requirement to core TF, so I don't think it is right to brand it like a requirement to TF.

As for installing it to ci, sure.
But that means it will be installed to TF docker images.

vrv:
Okay, closing for now but discussion can continue here if needed.

elviswf:
Actually I fully agree that adding this dependency is not necessary for now. I just used it for a deep learning competition with png, gif data and wanted to shared a way to handle it directly. 
 I will go on with excellent TF. Thank you all!

passerbythesun:
I have the same problem, so this piece code is useful for me.

PR #4860, opened 2016-10-09
Fixed leak `_EventLoggerThread` threads

### LINKED ISSUE #4820: Thread created by SummaryWriter not killed
Hi,

I noticed that the _EventLoggerThread created by summary writer does not get killed by the close method, which will make the number of threads keep increasing until it exceeds the system capacity. Is there any particular reason to not kill these threads?

Best

haosdent:
This fixes #4820 .

### DIFF
--- a/tensorflow/python/summary/summary_iterator.py
+++ b/tensorflow/python/summary/summary_iterator.py
@@ -281,9 +281,14 @@ class SummaryWriter(object):
 
     Call this method when you do not need the summary writer anymore.
     """
+    # Need to mark it closed first then event_queue would stop to accept new events.
+    self._closed = True
+
+    self._worker.stop()
+    self._worker.join()
+
     self.flush()
     self._ev_writer.Close()
-    self._closed = True
 
 
 class _EventLoggerThread(threading.Thread):
@@ -301,15 +306,16 @@ class _EventLoggerThread(threading.Thread):
     """
     threading.Thread.__init__(self)
     self.daemon = True
+    self._stop_flag = False
     self._queue = queue
     self._ev_writer = ev_writer
     self._flush_secs = flush_secs
     # The first event will be flushed immediately.
     self._next_event_flush_time = 0
 
-  def run(self):
+  def _run(self):
     while True:
-      event = self._queue.get()
+      event = self._queue.get(False)
       try:
         self._ev_writer.WriteEvent(event)
         # Flush the event writer every so often.
@@ -320,6 +326,20 @@ class _EventLoggerThread(threading.Thread):
           self._next_event_flush_time = now + self._flush_secs
       finally:
         self._queue.task_done()
+
+  def run(self):
+    while True:
+      try:
+        self._run()
+      except:
+        # We check `_stop_flag` first to ensure there is no active producers
+        # when we check if `self._queue` is empty.
+        if self._stop_flag and self._queue.empty():
+          self._ev_writer.Flush()
+          break
+
+  def stop(self):
+    self._stop_flag = True
 
 
 def summary_iterator(path):
--- a/tensorflow/python/summary/summary_iterator.py
+++ b/tensorflow/python/summary/summary_iterator.py
@@ -102,6 +102,7 @@ class SummaryWriter(object):
     if not gfile.IsDirectory(self._logdir):
       gfile.MakeDirs(self._logdir)
     self._event_queue = six.moves.queue.Queue(max_queue)
+    self._add_events_counter = 0
     self._ev_writer = pywrap_tensorflow.EventsWriter(
         compat.as_bytes(os.path.join(self._logdir, "events")))
     self._closed = False
@@ -180,13 +181,16 @@ class SummaryWriter(object):
     """
     if not self._closed:
       self._event_queue.put(event)
+      self._add_events_counter += 1
+    else:
+      raise Exception("Event queue is closed.")
 
   def _add_graph_def(self, graph_def, global_step=None):
     graph_bytes = graph_def.SerializeToString()
     event = event_pb2.Event(wall_time=time.time(), graph_def=graph_bytes)
     if global_step is not None:
       event.step = int(global_step)
-    self._event_queue.put(event)
+    self.add_event(event)
 
   def add_graph(self, graph, global_step=None, graph_def=None):
     """Adds a `Graph` to the event file.
@@ -265,7 +269,7 @@ class SummaryWriter(object):
                             tagged_run_metadata=tagged_metadata)
     if global_step is not None:
       event.step = int(global_step)
-    self._event_queue.put(event)
+    self.add_event(event)
 
   def flush(self):
     """Flushes the event file to disk.
@@ -312,6 +316,7 @@ class _EventLoggerThread(threading.Thread):
     self._flush_secs = flush_secs
     # The first event will be flushed immediately.
     self._next_event_flush_time = 0
+    self._success_events_counter = 0
 
   def _run(self):
     while True:
@@ -324,6 +329,8 @@ class _EventLoggerThread(threading.Thread):
           self._ev_writer.Flush()
           # Do it again in two minutes.
           self._next_event_flush_time = now + self._flush_secs
+
+        self._success_events_counter += 1
       finally:
         self._queue.task_done()
 
--- a/tensorflow/python/summary/writer/event_file_writer.py
+++ b/tensorflow/python/summary/writer/event_file_writer.py
@@ -114,9 +114,14 @@ class EventFileWriter(object):
 
     Call this method when you do not need the summary writer anymore.
     """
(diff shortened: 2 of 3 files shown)


haosdent:
Hi, @tatatodd Would you available to review this? Thank you in advance.

tatatodd:
Sorry, I'm not a good reviewer for this, since I don't actually know Python that well.

Re-assigning to @mrry since it seems related to stuff that he's worked on.

tensorflow/python/summary/summary_iterator.py
```diff
@@ -301,26 +306,35 @@ def __init__(self, queue, ev_writer, flush_secs):
...
     self._ev_writer = ev_writer
     self._flush_secs = flush_secs
     # The first event will be flushed immediately.
     self._next_event_flush_time = 0
+    self._fetch_timeout = 1
 
   def run(self):
     while True:
-      event = self._queue.get()
+      event = self._queue.get(timeout=self._fetch_timeout)
```
mrry: IIUC, `self._queue.get(timeout=...)` will raise a `Queue.Empty` exception if there are no elements after the timeout. Let's not use a timeout here (see other comment in this method).

tensorflow/python/summary/writer/event_file_writer.py
```diff
@@ -114,9 +114,14 @@ def close(self):
...
+
+    self._worker.stop()
+    self._worker.join()
+
     self.flush()
     self._ev_writer.Close()
-    self._closed = True
 
 
 class _EventLoggerThread(threading.Thread):
```
mrry: Same comments from the other thread class apply here.

tensorflow/python/summary/summary_iterator.py
```diff
@@ -301,26 +306,35 @@ def __init__(self, queue, ev_writer, flush_secs):
...
       try:
-        self._ev_writer.WriteEvent(event)
+        if event:
+          self._ev_writer.WriteEvent(event)
         # Flush the event writer every so often.
         now = time.time()
-        if now > self._next_event_flush_time:
+        if now > self._next_event_flush_time or self._stop_flag:
```
mrry: Is it necessary to flush when `self._stop_flag` is true? The `stop()` method also calls `flush()`.

tensorflow/contrib/learn/python/learn/tests/summary_writer_test.py
```diff
@@ -0,0 +1,43 @@
+# pylint: disable=g-bad-file-header
```
mrry: Thanks for adding a test! Please move it next to the implementation, i.e. to `tensorflow/python/summary/summary_writer_test.py`.

tensorflow/contrib/learn/python/learn/tests/summary_writer_test.py
```diff
@@ -0,0 +1,43 @@
...
+class SummaryWriterTest(tf.test.TestCase):
+  """SummaryWriter tests."""
+
+  def testLoggerThreadExit(self):
+    writer = SummaryWriter(tempfile.mkdtemp(), graph=ops.get_default_graph())
+    previous_thread_count = threading.active_count()
+    writer.close()
+    current_thread_count = threading.active_count()
+    self.assertTrue(previous_thread_count + 1 == current_thread_count)
+
```
mrry: Could you please add a test that calls `stop()` on the main thread, while a separate thread calls `add_event()` (and catches the appropriate exception when the queue is closed)? For bonus points, please then read back the written log, and ensure that the expected events are found within.

mrry (changes requested):

tensorflow/python/summary/summary_iterator.py
haosdent: @mrry Thanks a lot for your reviews! For `self._queue.get()`, it would block forever and would not raise exception according to my test.

haosdent: Looks `self._queue.get(false)` would not block and raise empty exception.

tensorflow/python/summary/summary_iterator.py
```diff
@@ -301,26 +306,35 @@ def __init__(self, queue, ev_writer, flush_secs):
...
           self._ev_writer.Flush()
           # Do it again in two minutes.
           self._next_event_flush_time = now + self._flush_secs
       finally:
         self._queue.task_done()
 
+      if self._stop_flag:
+        break
```
haosdent: Thank a lot for nice analyze, actually I attempt to drop messages before because I would like to exit immediately. But I agree with all your ideas, let me change to make sure all events flush before exit. Thank you very much for your inputs!

tensorflow/contrib/learn/python/learn/tests/summary_writer_test.py
haosdent: Sury thing, if this duplicated with

```
Could you add a test that has a thread periodically writing events to the writer, then the main thread closes it, and the appropriate exceptions are raised?
```
?

tensorflow/python/summary/summary_iterator.py
```diff
@@ -301,15 +306,16 @@ def __init__(self, queue, ev_writer, flush_secs):
...
     self._ev_writer = ev_writer
     self._flush_secs = flush_secs
     # The first event will be flushed immediately.
     self._next_event_flush_time = 0
 
-  def run(self):
+  def _run(self):
     while True:
-      event = self._queue.get()
+      event = self._queue.get(False)
```
mrry: Calling `self._queue.get(False)` when the queue is empty will result in this thread busy-waiting for a new element (or for the queue to be closed), which will have unacceptable performance.

Instead of relying on exceptions, can you implement the sentinel object version?

tensorflow/contrib/learn/python/learn/tests/summary_writer_test.py
mrry: Yes, sorry for the duplicated comment. I'd prefer the version described in this comment.

mrry (approved):

mrry (changes requested):
Approved by mistake... please make the suggested changes.

tensorflow/contrib/learn/python/learn/tests/summary_writer_test.py
haosdent: Got it, thank you!

haosdent:
@mrry Sorry for the delay, I have addressed your comments and add new test cases, would you help to review this again? Thank you in advance.

tensorflow/python/summary/summary_iterator.py
mrry: This version still has the busy waiting problem. Can you address this as I suggested above?

tensorflow/python/summary/summary_iterator.py
```diff
@@ -102,6 +102,7 @@ def __init__(self, logdir, graph=None, max_queue=10, flush_secs=120,
     if not gfile.IsDirectory(self._logdir):
       gfile.MakeDirs(self._logdir)
     self._event_queue = six.moves.queue.Queue(max_queue)
+    self._add_events_counter = 0
```
mrry: This private member does not appear to be read anywhere in this class.

tensorflow/python/summary/summary_iterator.py
```diff
@@ -318,9 +329,25 @@ def run(self):
           self._ev_writer.Flush()
           # Do it again in two minutes.
           self._next_event_flush_time = now + self._flush_secs
+
+        self._success_events_counter += 1
```
mrry: This private member does not appear to be read anywhere in this class.

tensorflow/python/training/summary_writer_test.py
```diff
@@ -0,0 +1,72 @@
...
+          event = tf.Event(wall_time=int(time.time()), step=i)
+          writer.add_event(event)
+
+    producer = threading.Thread(target=put_events)
+    producer.start()
+
+    # Ensure there are enough events be flushed because close the writer,
+    # then we could continue to compare the number of published events and
+    # flushed events.
+    while writer._worker._success_events_counter <= 0:
```
mrry: Ah, now I see why you added those apparently unused private fields to the various objects.

Instead of adding state to every instance of this class, keep a count in this test case of how many objects are written, and compare it to the number read back from the file.

mrry (changes requested):

vrv:
Any updates @haosdent ?

rmlarsen:
@haosdent When would you be able to update this?

danmane:
@haosdent Any update? If there isn't any activity on this soon I will likely close the PR.

mrry:
I'm going to close this PR due to inactivity, and because it still needs quite a lot of work before it'd be ready to merge. Feel free to open a new PR if you recommence working on this, and thanks for contributing!

PR #4874, opened 2016-10-10
Fixed and improved ./configure on Windows

meteorcloudy:
1. Make the script work with path containing spaces (If users install python by Anaconda, the python path will probably contain space)
2. On Windows, disable bazel clean --expunge(it's broken)
3. Create junctions instead of copying python lib, include, numpy folders
4. fixed symlink creation in gen_git_source.py with python 3.5


tensorflow/tools/git/gen_git_source.py
```diff
@@ -112,9 +112,14 @@ def configure(src_base_path, debug=False):
     if src is None:
       open(os.path.join(gen_path, target), "w").write("")
     else:
-      if hasattr(os, 'symlink'):
-        os.symlink(src, os.path.join(gen_path, target))
-      else:
+      try:
+        # In python 3.5, symlink function exists even on Windows
+        # But requires previliaged right, otherwise a OSError will be thrown
```
rohan100jain: s/priviliaged/privileged 

Also can you expand on what privileges are needed?

rohan100jain (commented):
You seem to have added a lot of quotes around variables etc. Can you summarize why that is necessary for windows?

configure
```diff
@@ -8,8 +8,22 @@ pushd `dirname $0` #> /dev/null
 SOURCE_BASE_DIR=`pwd -P`
```
mrry: Could this include spaces? Do we need quotes around each use of `$SOURCE_BASE_DIR` as well?

tensorflow/tools/git/gen_git_source.py
mrry: IIRC, Windows requires a special user right for creating symbolic links. It's available if you have Administrator privileges, and although it can be extended to regular users, it isn't by default....

util/python/python_config.sh
```diff
@@ -108,26 +108,27 @@ function setup_python {
...
-  local python_include=$("${PYTHON_BIN_PATH}" -c 'from __future__ import print_function; from distutils import sysconfig; print(sysconfig.get_python_inc());')
+  local python_include="$("${PYTHON_BIN_PATH}" -c 'from __future__ import print_function; from distutils import sysconfig; print(sysconfig.get_python_inc());')"
   if [ "$python_include" == "" ]; then
     echo -e "\n\nERROR: Problem getting python include path.  Is distutils installed?"
     exit 1
   fi
-
-  local python_lib_path=$(python_path)
+  local python_lib_path
+  # Split python_path into an array of paths, this allows path cotaining spaces
```
mrry: Nit: s/cotaining/containing/

util/python/python_config.sh
```diff
@@ -142,24 +143,29 @@ function setup_python {
...
       rm -rf "$x"
     fi
   done
 
-# ln -sf is acutally implemented as copying in msys since creating symbolic links is privileged on Windows
-# So we need -rf to remove them above.
-  ln -sf "${python_include}" util/python/python_include
-  ln -sf "${python_lib}" util/python/python_lib
-  ln -sf "${numpy_include}" third_party/py/numpy/numpy_include
+# ln -sf is acutally implemented as copying in msys since creating symbolic links is privileged on Windows.
```
mrry: Nit: 80 characters, s/acutally/actually/.

util/python/python_config.sh
```diff
@@ -142,24 +143,29 @@ function setup_python {
...
+# But copying is too slow, so invoke mklink to create junctions on Windows.
+  if is_windows; then
+    cmd  /c "mklink /J util\\python\\python_include \"${python_include}\""
```
mrry: Nit: one space between `cmd` and `/c` (same on two following lines)

util/python/python_config.sh
```diff
@@ -183,7 +189,12 @@ function check_python {
       echo -e "\n\nERROR: '${x}' is not a symbolic link.  Internal error.\n\n" 1>&2
       exit 1
     fi
-    true_path=$(readlink "${x}")
+    if is_windows; then
+      # In msys, readlink path just doesn't work, because no symbolic link on Windows
```
mrry: Nit: 80 characters per line.

mrry (changes requested):
With the `is_windows()` function, would it be worth guarding the other configuration options that aren't currently supported on Windows (e.g. maybe GCP, HDFS, CUDA)?

configure
meteorcloudy: Yes, I believe so.

meteorcloudy:
@mrry Yes, I think it's a good idea to use `is_windows` to opt out GCP, HDFS, CUDA support on Windows for now.

meteorcloudy:
@rohan100jain For example, if users install python using Anaconda(which is recommended in CMake build), the python path will probably contain spaces like this `C:\Program Files\Anaconda3\lib\site-packages`, so we have to add quotes to the variable.

meteorcloudy:
@mrry Thanks for catching so many typos.. Please review again. :)

mrry (approved):
Looks good to me!

mrry:

mrry:

rohan100jain:

meteorcloudy:
The configure script does cause a bug on Mac, I'll look into it.

meteorcloudy:
`read -r -a python_lib_path <<< $(python_path)` doesn't really split the string into a bash array on OSX, so switch to another way.
Please test again! :)

DevinCheung:
I install the tensorflow on windows via the cmake/README. It seems everything goes well. And after I "activate tensorflow" and "import tensorflow as tf". It says "No module named 'tensorflow'" . I do not know how to fix it.

mrry:

meteorcloudy:
Great, looks like I fixed it!

PR #4889, opened 2016-10-11
Switch stream_executor to use portable dlopen()

mrry:
Use the functions from `port::Env` rather than `dlfcn.h`, to make it easier to port stream_executor to Windows.


mrry:
@vrv This is a step towards making stream_executor build on Windows. I haven't edited code in here before... is there any special process for modifying stream_executor code?

vrv (approved):
Seems good, but I'm not sure how we expect to upstream this change to StreamExecutor when it moves into LLVM. 

cc/ @henline

PR #4896, opened 2016-10-11
[bug]unmatched quote causes big problem

DjangoPeng:
An unmatched quote causes thousands lines typesetting error. Look at the api docs, from [here](https://www.tensorflow.org/versions/r0.11/api_docs/python/state_ops.html#constant_initializer) to [here](https://www.tensorflow.org/versions/r0.11/api_docs/python/state_ops.html#IndexedSlices) has the problem, which influence almost 30 functions api docs from `constant_initializer` to `IndexedSlices`. 

We'd better fix it ASAP

### DIFF
--- a/tensorflow/g3doc/api_docs/python/state_ops.md
+++ b/tensorflow/g3doc/api_docs/python/state_ops.md
@@ -2382,10 +2382,6 @@ tensor shape, the initializer will raise a `ValueError`.
 
   An initializer that generates tensors with constant values.
 
-##### Raises:
-
-*  <b>`ValueError`</b>: Too many elements provided. Needed at most 6, but received 8
-
 ##### Examples:
 
   The following example can be rewritten using a numpy.ndarray instead
@@ -2429,8 +2425,9 @@ tensor shape, the initializer will raise a `ValueError`.
   >>> with tf.Session():
   >>>   x = tf.get_variable('x', shape=[2, 3], initializer=init)
 
-```
-  
+
+*  <b>`ValueError`</b>: Too many elements provided. Needed at most 6, but received 8
+  ```
 - - -
 
 ### `tf.random_normal_initializer(mean=0.0, stddev=1.0, seed=None, dtype=tf.float32)` {#random_normal_initializer}
--- a/tensorflow/g3doc/api_docs/python/state_ops.md
+++ b/tensorflow/g3doc/api_docs/python/state_ops.md
@@ -2382,6 +2382,10 @@ tensor shape, the initializer will raise a `ValueError`.
 
   An initializer that generates tensors with constant values.
 
+##### Raises:
+
+*  <b>`ValueError`</b>: Too many elements provided. Needed at most 6, but received 8
+
 ##### Examples:
 
   The following example can be rewritten using a numpy.ndarray instead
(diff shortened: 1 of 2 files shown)


rohan100jain (changes requested):
I don't think this is the right fix for this issue. We want the python docs to be automatically generated. So either the documentation in init_ops.py should be updated or the script that generates the docs (gen_docs.sh) should be fixed. More details about documentation at https://www.tensorflow.org/versions/master/how_tos/documentation/index.html

DjangoPeng:
@rohan100jain Gotcha~ I updated the `init_ops.constant_initializer()`, and I think it would work when you generate the doc next time.

tensorflow/g3doc/api_docs/python/state_ops.md
```diff
@@ -2382,6 +2382,10 @@ tensor shape, the initializer will raise a `ValueError`.
 
   An initializer that generates tensors with constant values.
 
+##### Raises:
```
rohan100jain: revert this change?

tensorflow/python/ops/init_ops.py
```diff
@@ -95,7 +95,10 @@ def constant_initializer(value=0, dtype=dtypes.float32):
 
   Returns:
     An initializer that generates tensors with constant values.
-
+    
+  Raises:
```
rohan100jain: This is explained above I think...

"If the total number of elements in `value` is greater than the number of elements required by the tensor shape...."

so perhaps you could say something like...

Raises:
  ValueError: When total number of elements in `value` is more than the number of elements required by the tensor shape as explained above.

tensorflow/python/ops/init_ops.py
```diff
@@ -138,8 +141,8 @@ def constant_initializer(value=0, dtype=dtypes.float32):
     >>> with tf.Session():
     >>>   x = tf.get_variable('x', shape=[2, 3], initializer=init)
 
-    ValueError: Too many elements provided. Needed at most 6, but received 8
-    ```
+    ValueError: Too many elements provided. Needed at most 6, but received 8.
```
rohan100jain: So what was it that was causing the formatting bug? The lack of a period at the end of the sentence or the extra space in front of the ```? 

Can you run the script locally and verify that the docs are generated correctly?

rohan100jain (changes requested):

tensorflow/python/ops/init_ops.py
DjangoPeng: The extra space caused the bug. It's easy to test in the markdown editor, and it would be considered as `code` instead of quote flag if you give the extra space.

tensorflow/g3doc/api_docs/python/state_ops.md
DjangoPeng: sure

tensorflow/python/ops/init_ops.py
```diff
@@ -95,7 +95,10 @@ def constant_initializer(value=0, dtype=dtypes.float32):
 
   Returns:
     An initializer that generates tensors with constant values.
-
+    
+  Raises:  
+    ValueError: When total number of elements in value is more than the number of elements required by the tensor shape as explained above.
```
rohan100jain: 80 chars

rohan100jain (changes requested):

tensorflow/python/ops/init_ops.py
DjangoPeng: ...

DjangoPeng:
It seems like there is a conflict between `state_ops.md` and `init_ops.py`. So, I want to reopen a PR to fix the bug.

PR #4892, opened 2016-10-11
Add fused_batch_norm layer

thuyen:
Add data_format option for convolution2d, bias_add, max_pool2d, and avg_pool2d as well. This helps to fully utilize the cudnn speed for normal convnets with 'NCHW' data layout.

PR #4899, opened 2016-10-11
Branch 135828145

rohan100jain:


rohan100jain:

thuyen:
I just clone and tested this branch. For new batch_norm layer, when `fused` is `True` and `is_training` is a symbolic place_holder, gradients computation failed. Pull request #4892 addresses a little bit of this, but that won't be merged due to overlaps with this one I guess.

zhangyaobit:
Hi thuyen, thank you very much for your contribution and the help on testing. Unfortunately we had similar internal changes at about the same time. I noticed you added is_training to the gradient op in nn_ops.cc. Is it to solve this "shape not compatible" problem? There is no gradient computation for inference (is_training = False), in other words, is_training is not applicable to the gradient op. So adding it is a bit confusing to me. Could you explain why it is needed (or how it solved the shape incompatible issue)? Thanks!

zhangyaobit:
Hi thuyen, could you also point out which test could reproduce this error? Does this test cover this case? https://github.com/tensorflow/tensorflow/pull/4899/commits/df75f38eb7e1aacfbb85c637c72ccdfffc1ada78#diff-fa0c7294c4b5300fddcdc8753433406aR1616

Thanks.

thuyen:
Hi, Yao

You should see the error with this snippet:
```
import tensorflow as tf
from tensorflow.contrib import slim
import numpy as np
is_training = tf.placeholder(tf.bool)
inputs = tf.constant(np.zeros((10, 3, 100, 100)).astype('float32'))
outputs = slim.batch_norm(inputs, is_training=is_training, fused=True, data_format='NCHW')
loss = tf.reduce_sum(outputs)
grad = tf.gradients(loss, tf.trainable_variables())
sess = tf.Session()
p = sess.run(grad, feed_dict={is_training:True})
```
It's like: 
```
File "..../tensorflow/python/ops/gradients.py", line 491, in gradients
in_grad.set_shape(t_in.get_shape())
File "..../tensorflow/python/framework/ops.py", line 408, in set_shape
self._shape = self._shape.merge_with(shape)
File "..../tensorflow/python/framework/tensor_shape.py", line 583, in merge_with
(self, other))
ValueError: Shapes (0,) and (3,) are not compatible
```
Notice in the fused layer:
```
def _fused_batch_norm_training():
  return nn.fused_batch_norm(
      inputs, gamma, beta, epsilon=epsilon, data_format=data_format)
def _fused_batch_norm_inference():
  return nn.fused_batch_norm(
      inputs,
      gamma,
      beta,
      mean=moving_mean,
      variance=moving_variance,
      epsilon=epsilon,
      is_training=False,
      data_format=data_format)
outputs, mean, variance = utils.smart_cond(is_training,
                                           _fused_batch_norm_training,
                                           _fused_batch_norm_inference)

```
So when when `is_training` is a placeholder, we need both branches of the condition for the `outputs`. The same is true for gradients. So the tf.gradients function will actually go into the branch `is_training=False`, and perform gradient computation symbolically (even though at run time we never call the gradient op when `is_training=False`). Since we are not calling that branch at run time we shouldn't worry too much about it. But right now the shapes for  `moving_mean` and `moving_variance` gradients are set incorrectly for `is_training=False` (they are all `0` now, should be `0` for `is_training=True` only) and the tf.gradients catch that. I modified shape inference for gradient op in `nn_ops.cc` for that reason.

zhangyaobit:
Hi thuyen, thank you for catching this subtle issue, as well as the detailed explanation! After the merge, I think we can still have your changes on the fix (adding is_training to the gradient op) and the data_format support for bias_add in your pull request #4892

rohan100jain:
Yao, let me know if this is something you want to approve. After that I could merge this in and we can fix the other issues that have been brought up (thanks @thuyen !) through the other PR.

zhangyaobit:
Hi Rohan, yes, this sounds good to me.

PR #4905, opened 2016-10-12
Fix Mac CUDNN library path issue in configure script

caisq:


gunan (approved):

davidzchen:
Thanks for fixing this!

PR #4941, opened 2016-10-13
FIXBUG: convolution use error GFile API

hahnjiang:
This place should use Size() method.


pooorman:
I had signed a CLA and add email address in my github. How to pass the authorization. Thanks

rohan100jain:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/lib/io/file_io.py#L83

defines a size() and not Size() method.

PR #4951, opened 2016-10-14
Minor Python code health improvements

frankfqchen:
replace is with == for len()

PR #4955, opened 2016-10-14
refine benchmark_test.py

frankfqchen:
delete .keys()
no need to get keys()
just use 'in' instead


rohan100jain:
Could you actually merge all these small fixes in one PR? It'll reduce load on the testing framework :)

rohan100jain:
Thanks for all the fixes btw!

jart:
Thank you for your contribution @frankfqchen. We love code health improvements, but we'd prefer that it not be done in seven separate one-liner commits. It reduces the signal to noise ratio of our change history and puts stress on our CI system. Please merge this pull request (along with the others) into #4951 since that's the only one with an actionable comment.

PR #4962, opened 2016-10-14
Fix too long sentence

ofilip:
models/rnn/translate.py with --decode option raises exception if input sentence is longer than 20 (default lower bound on last bucket)

> 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5
> Traceback (most recent call last):
> File "translate_edit.py", line 298, in
> tf.app.run()
> File "/home/modrymar/venv/lib/python3.5/site-packages/tensorflow/python/platform/app.py", line 30, in run
> sys.exit(main(sys.argv))
> File "translate_edit.py", line 293, in main
> decode()
> File "translate_edit.py", line 241, in decode
> bucket_id = min([b for b in xrange(len(_buckets))
> ValueError: min() arg is an empty sequence

### DIFF
--- a/tensorflow/models/rnn/translate/translate.py
+++ b/tensorflow/models/rnn/translate/translate.py
@@ -238,8 +238,12 @@ def decode():
       # Get token-ids for the input sentence.
       token_ids = data_utils.sentence_to_token_ids(tf.compat.as_bytes(sentence), en_vocab)
       # Which bucket does it belong to?
-      bucket_id = min([b for b in xrange(len(_buckets))
-                       if _buckets[b][0] > len(token_ids)])
+      for i, bucket in enumerate(_buckets):
+        if bucket[0]>=len(token_ids):
+          bucket_id = i
+          break
+        else:
+          bucket_id = len(_buckets)-1
       # Get a 1-element batch to feed the sentence to the model.
       encoder_inputs, decoder_inputs, target_weights = model.get_batch(
           {bucket_id: [(token_ids, [])]}, bucket_id)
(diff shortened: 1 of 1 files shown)


lukaszkaiser (approved):
Looks good, but maybe let's print some warning in the else part so it's clear we're truncating the sentence?

ofilip:
Warning sounds good. I'm not sure how it works when some change on pull request is required. Who'll add the warning?

jart (dismissed):
This would be more idiomatic:

``` py
      bucket_id = len(_buckets) - 1
      for i, bucket in enumerate(_buckets):
        if bucket[0] >= len(token_ids):
          bucket_id = i
          break
      else:
        logging.warning("Sentence truncated: %s", sentence)
```

drpngx:
Feel free to address @jart's comments and respond.

ofilip:
Sorry for slow reaction. I'm not really used to github review's. Code is updated according to @jart's suggestion.

jart:
Mr. Jenkins test this please

The logging import needs to be defined.

drpngx:
Jenkins, test this please

tensorflow/models/rnn/translate/translate.py
```diff
@@ -238,8 +239,14 @@ def decode():
...
       # Which bucket does it belong to?
-      bucket_id = min([b for b in xrange(len(_buckets))
-                       if _buckets[b][0] > len(token_ids)])
+      bucket_id = len(_buckets) - 1
+      for i, bucket in enumerate(_buckets):
+        if bucket[0] >= len(token_ids):
+          bucket_id = i
+          break
+      else:
+        logging.warning("Sentence truncated: %s", sentence)
```
vrv: One alternative is to use the 'tf.logging' module and then you can remove the 'logging' import.  Can do in a follow up PR.

Issue #706, opened 2016-01-06
swig command not found


wenhuchen:
I try to build the GPU version on uBuntu 64 machine with command, 
```
bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package
```
I installed swig and bazel successfully, but it still reports that swig not found.
```
src/main/tools/namespace-sandbox.c:476: arg: tensorflow/python/tensorflow.i
bazel-out/host/bin/tensorflow/swig: line 17: swig: command not found
ERROR: /userdata01/Users/wenhuchen/tensorflow/tensorflow/tensorflow.bzl:298:3: SWIGing tensorflow/python/tensorflow.i failed: Error during execution of spawn: Process exited with status 127: Error during execution of spawn: Process exited with status 127.
Target //tensorflow/tools/pip_package:build_pip_package failed to build
```
Can anyone help me with that?

wenhuchen:
swig version is 3.0.8

mrry:
Where is your copy of `swig` installed (i.e. what result do you get when you run `which swig`)?

As a short-term fix, you can modify [`swig.sh`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/swig/swig.sh) to invoke SWIG by its absolute path.

wenhuchen:
I installed it locally, I specify the absolute path in the swig.sh and got the same error
```
 /home/wenhuchen/.local/bin/swig: No such file or directory
ERROR: /userdata01/Users/wenhuchen/tensorflow/tensorflow/tensorflow.bzl:298:3: SWIGing tensorflow/python/tensorflow.i failed: Error during execution of spawn: Process exited with status 127: Error during execution of spawn: Process exited with status 127.
Target //tensorflow/tools/pip_package:build_pip_package failed to build
```
I'm sure swig is usable, and my swig.sh looks like below
```
 /home/wenhuchen/.local/bin/swig "$@"
```

mrry:
Just to confirm, if you run `/home/wenhuchen/.local/bin/swig` from the command line, does that work without error?

psycharo:
try modifying [tensorflow.bzl here ](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorflow.bzl#L261) by adding an extra parameter:
```
use_default_shell_env=True,
```
and building with: 
```
--genrule_strategy=standalone --spawn_strategy=standalone
```

[...]

PR #4983, opened 2016-10-15
use use_default_shell_env for SWIGing action

AndreasMadsen:
When swig is in the PATH but not in a standard bin location, bazel can't
find it unless the PATH is propagated.

_I have only spend two days on becoming familiar with tensorflow and bazel, so I could easily be wrong._

### DIFF
--- a/tensorflow/tensorflow.bzl
+++ b/tensorflow/tensorflow.bzl
@@ -563,7 +563,8 @@ def _py_wrap_cc_impl(ctx):
              inputs=sorted(set([src]) + cc_includes + ctx.files.swig_includes +
                          ctx.attr.swig_deps.files),
              outputs=outputs,
-             progress_message="SWIGing {input}".format(input=src.path))
+             progress_message="SWIGing {input}".format(input=src.path),
+             use_default_shell_env=True)
   return struct(files=set(outputs))
 
 _py_wrap_cc = rule(


jart:
If you run `./configure` does it solve the problem? See also: #706

AndreasMadsen:
I'm not sure what you mean. If apply this PR and run configure then it fixes the problem. Yes.

I have other compiler problems after this, as described in https://github.com/tensorflow/tensorflow/issues/4984. I have yet to solve those.

jart:
The `./configure` script in the root of the TensorFlow repository creates a symlink to the system swig command inside `tensorflow/tools/swig`. The only reason that `tensorflow/tools/swig.sh` would try to run the swig command from the environment is if that symlink doesn't exist. So the solution is most likely to run `./configure`. Could you please confirm if that solves your problem?

I would ideally prefer to not use `use_default_shell_env` because our Bazel build should ideally be hermetically sealed. That's one of the goals I'm working towards. I'm actually working on a change right now that has Bazel fetch the sources for SWIG and compiles it from scratch.

jart:
According to #4984 you're using CentOS 6 which isn't in our support matrix. I really appreciate the time you put into sending this pull request. But I'm going to close this issue. I encourage you to send us a pull request adding a check to our `./configure` for `.el6.` in the `uname -a` output that shows:

```
CentOS 6 is not supported and probably won't work. Consider Docker or the tensorflow/contrib/makefile build system. Want to continue anyway? [Yn]
```

This should hopefully help save time for many people in the future.

Also I'm going to update this issue when my change is committed regarding SWIG hermeticism.

jart:
Also I'll reopen this issue provided new information.

PR #4992, opened 2016-10-16
Deprecated the `is_training` argument in vgg.py.

### LINKED ISSUE #4887: resnet model in tf slim does not take is_training param
Compare the [resnet_v2 model](https://github.com/tensorflow/tensorflow/blob/8e48ec6ea0492e2cb9fd19c0a2ccf41afc7b4dc6/tensorflow/contrib/slim/python/slim/nets/resnet_v2.py) to the [vgg model](https://github.com/tensorflow/tensorflow/blob/8e48ec6ea0492e2cb9fd19c0a2ccf41afc7b4dc6/tensorflow/contrib/slim/python/slim/nets/vgg.py). The vgg model takes `is_training` whereas the resnet model does not. This param should be taken and passed to the `batch_norm` layers.

The docs for the v1 model do reference `is_training`, but I don't see it used.

haosdent:
This fixes #4887.

### DIFF
--- a/tensorflow/contrib/slim/python/slim/nets/vgg.py
+++ b/tensorflow/contrib/slim/python/slim/nets/vgg.py
@@ -77,7 +77,8 @@ def vgg_a(inputs,
   Args:
     inputs: a tensor of size [batch_size, height, width, channels].
     num_classes: number of predicted classes.
-    is_training: whether or not the model is being trained.
+    is_training: whether or not the model is being trained. This argument is
+      being deprecated. Instead use `tf.contrib.framework.arg_scope`.
     dropout_keep_prob: the probability that activations are kept in the dropout
       layers during training.
     spatial_squeeze: whether or not should squeeze the spatial dimensions of the


haosdent:
Hi, @jart May you help to review this, thank you in advance.

jart (changes requested):
Thanks for taking the time to send us this.
1. I think this change is worth doing. But @sguada and @nathansilberman are owners of contrib/slim and have veto power.
2. If we're going to make this change, let's do it thoroughly. Here's what I'd like to see happen:
   1. Identify every single function in slim that accepts args purely to pass them along to other functions with the `@scopes.add_arg_scope` decorator. For example, in this file alone there's also vgg_16 and vgg_19 which haven't been updated by this PR.
   2. Remove those params.
   3. Add `**kwargs` param
   4. Add `logging.warning()` if deprecated params are passed, telling user to use arg_scope.
   5. Refactor every call site to use arg_scope that is passing those params.

sguada:
Actually I think it is actually better to let the is_training argument as part of the network, see 
https://github.com/tensorflow/models/tree/master/slim/nets which have unified API and would be merge soon into tf.contrib

haosdent:
Thanks @sguada and @jart I think we should discard this, right?

PR #5000, opened 2016-10-16
Add support for dict Input to DataFeeder, StreamingDataFeeder, Estimator.fit() and Estimator.evaluate()

abhitopia:
# Problem

The `Estimator.fit()` function takes as argument either
- (`x`, `y`, and `batch_size`) where `x` and `y` could be numpy arrays or iterators.
  - **PROS**
    1.   Easy to use.
    2.  Allows feeding data from arbitrary source as long as problem can be decomposed into `x` and `y`.
  - **CONS**
    1.  No provision to provide epoch
    2.  In case, `x` and `y` are arrays, the data aggregate must be available as opposed to reading on the fly ( say from database)
    3. Whether array or iterator, `x` and `y` can't be dictionaries. Most complex problems cannot be reduced to input matrix and output matrix and may require multiple input features matrices.
- `input_fn` - this is callback function which must return `features` and `target` tensors or dictionary of tensors.
  - **PROS**
    1. Allows feeding data from arbitrary source (in theory).
    2. returned features and targets can be dictionary thus allowing to solve complex problems which takes multiple inputs.
  - **CONS**
    1. Only found support for reading files using `read_batch_examples()`, `read_batch_features()`, `read_batch_record_features()`, etc.
    2. No support for passing placeholder and feed_fn to allow for arbitrary source of input data which don't require queue.
# Relevant discussions
1. https://github.com/tensorflow/tensorflow/pull/4696#issuecomment-253632403
2. http://stackoverflow.com/questions/39855375/how-to-use-streamingdatafeeder-as-contrib-learn-estimator-fits-input-fn
# Additional problem

Currently the self.feature_info (feature signature) must be same for training as well as evaluation. However, there can be cases where evaluation if done differently than training. For example, DualEncoderLSTM model where training just requires Context and Utterance, while Evaluation (Validation) requires context and multiple utterances.
# Solution

This PR basically combines the pros of using (`x`, `y`, and `batch_size`) with pros of input_fn by allowing `x`, `y` to be dictionary of multiple arrays. This also works for `x` and `y` iterators returning dictionary of streaming data. 

Also, added support for different feature/target signature for training and evaluation.


abhitopia:
@martinwicke -  Would be great to have your thoughts on this too.

drpngx:
@tyfkda is that something you would be interested in looking at perhaps?

abhitopia:
Just also added support for different feature_info (Signature) for train and evaluation.

abhitopia:
@martinwicke , @drpngx , @tyfkda - Have you guys had a chance to review it yet?

drpngx:
Sorry, not my area, but please resolve changes and ping again.

martinwicke:
Sorry -- we're thinking about predict quite a bit right now. This comes at a good time. Please give us some time to discuss.

abhitopia:
@martinwicke - Do I need to fix the conflicts first or perhaps I do it once you have reviewed it?

abhitopia:
@martinwicke - Any way I can contribute to the discussion?

pyrytakala:
+1 definitely see the use case of the functionality, when will get accepted?

abhitopia:
@vrv, @martinwicke - It would be great to get it reviewed soon. Please respond.

kamilryszkowski:
Exactly the what I was looking for. When will this get merged in master?

vrv:
@abhitopia Sorry, I'm not very familiar with the tf learn components at all, so I'll let @martinwicke guide :)

ilblackdragon:
Here is an update on current state:
 - we just split up SK compatibility from `Estimator` into separate class `SKCompat`. E.g. if one wants to use `fit(x, y)` they need to wrap their estimator with `SKCompat(Estimator(...))`. Otherwise preferred way will be .
  - there is a pending change (hopefully goes in today) to add `pandas_input_fn` that would allow you to `est.fit(pandas_input_fn(x_df, y_series), ...)` which would pass all the columns of the data frame as dict of features. It will also use local queue to feed data, which increases performance a bit compared to current data feeder solution.

Going forward we are thinking to remove data feeders, and rely on input functions similar to `pandas_input_fn` for other types of data. And `SKCompat` interface can just seamlessly wrap the passed data into this functions.

abhitopia:
Reponing in #5546

PR #5001, opened 2016-10-17
[g3doc] Fix image link error

jybaek:


jybaek:
I signed it!

drpngx:
Thanks for spotting the broken link!

We understand that the images are not checked in into github at the moment and that github looks broken right now. If you are building your internal version of the docs, we'd recommend you post-process the markdowns with a script.

We don't really want an absolute link since it will impair our ability to have multiple sites, for instance, a doc site for v1.0, and the last of the v0.xx.

jybaek:
@drpngx Thank you for your answer. I understand.

PR #5034, opened 2016-10-18
Fix hdfs scheme parsing issue in tensorboard

llhe:
Tensorboard HDFS support is broken because it's treated as a local relative path.


llhe:
Sorry, one typo in this patch, resubmit again.

PR #5055, opened 2016-10-18
Suppress all warnings that appear >100 times in windows cmake build.

gunan:


mrry (approved):
Looks good to me! How much does it reduce the log size?

drpngx:
Nice, thanks!

gunan:
We will see right now, after the above tests run :)

drpngx:
Roughly: 
* build 134 (which includes other commits), 33MB, 107k lines
* build 139, 3.7MB, 21k lines

mrry:
That's a nice improvement :). It looks like there's a lot of other chaff arising from an unknown compiler option - looking at the vcxproj files it seems as if two of the flags are being concatted together... I'll see what I can do in another PR.

gunan:
ah, I need to resolve conflicts now.
Will do that

PR #5057, opened 2016-10-18
CMake: Set the /MP flag correctly on Windows

mrry:
This should fix the "invalid option" warning we were seeing in the CMake logs.


mrry:
This builds on the nice work in #5055.

gunan (approved):

gunan:
GPU issue is a flake, so force merging this one.

PR #5058, opened 2016-10-18
fixed issue #4903

willdzeng:
as mentioned in #4903,
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py#L47
it directly run main function which doesn't run the argument parser in if __name__=="__main__".
I am not sure if FLAGS is used in somewhere else, so I moved the parser part into a function to update FLAGS in main() function.

Not sure if this is the right way to fix, but it works for both running through
`python tesnsorboard.py --logdir=AAA`
and
`tensorbaord --logdir=AAA`

### DIFF
--- a/tensorflow/tensorboard/tensorboard.py
+++ b/tensorflow/tensorboard/tensorboard.py
@@ -1,4 +1,4 @@
-# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2015-2016 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


danmane:
@ellibixby PTAL

drpngx:
Jenkins, test this please.

tobegit3hub:
Looks good to fix the issue.

drpngx:
For the tests, PR #5018 is in review.

elibixby:
While those tests look great, I was actually referring to intergration tests that verify the CLI can be run properly. (which I don't think I see there).

drpngx:
Ah, didn't realize we didn't have those.

tensorflow/tensorboard/tensorboard.py
```diff
@@ -38,8 +38,8 @@
 
 FLAGS = None
 
-
 def main(unused_argv=None):
+  FLAGS = parse_argument()
```
willdzeng: The flags are actually not consumed in app.run() for some reasons, it can run both by 
`$ python tensorboard.py() --help`
and
`$ tensorboard --help`
Maybe it's different on other platforms, but on ubuntu it works.

I think an easy way to fix is to let setup tools setup the link directly run the tensorboard.py file instead of the main function.

Thank you anyway. Wish it's helpful.

elibixby:
We don't I was saying we need them =)

Issue #4833, opened 2016-10-07
Slow IDCT used in JPEG decode


mkolod:
This issue is related to #4807, yet a bit distinct. Regardless of whether libjpeg or libjpeg-turbo is used, I was wondering if there was a reason to choose the slow but more accurate variant of IDCT in libjpeg ([code link](https://github.com/tensorflow/tensorflow/blob/6b1d4fd8090d44d20fdadabf06f1a9b178c3d80c/tensorflow/core/lib/jpeg/jpeg_mem.cc#L143)). For human image viewing pleasure, accuracy may matter, but I was wondering if for model training, the decision was made to use the slower IDCT because that improved model convergence? In case the training is immune to using the faster but less accurate IDCT, it may be worth changing the default. Has anyone done any studies as to the trained model impact of the faster IDCT? From a performance point of view, I'm getting a significant performance gain (2x GTX 1080, 6-core high-clock Core i7, 500 MB/s SSD), around 20% for AlexNet against precomputed ImageNet AlexNet may be more CPU bound for decode since the network passes themselves are relatively cheap, so maybe the gains for Inception v3 or ResNet would be less, but they would still likely be significant.

tatatodd:
@mkolod Thanks for another interesting point.  These are good areas to investigate wrt runtime performance!

I don't know of any studies of setting JDCT_IFAST, either wrt runtime performance or model accuracy.

As before, @prb12 and @poxvoculi might be interested.

prb12:
Interesting ... when training models for real (i.e. where time-to-accuracy is often more important than step-time), people often deliberately introduce noise and random distortions after decoding the image, so I can't imagine this would hurt!

TimZaman:
Choosing for JPEG is already a choice for lossy compression, so flipping the toggle makes sense.

mkolod:
@TimZaman Indeed.

vrv:
I made the fast IDCT the default in febdc1d13576a28382122e718bc8ab22ba9e6a45

Issue #4807, opened 2016-10-06
Upgrade to libjpeg-turbo


mkolod:
TensorFlow seems to be using the classic libjpeg v 0.9, but libjpeg-turbo is heavily optimized to leverage SIMD on various platforms (x86, ARM, POWER, etc.), and is both API and ABI compatible with libjpeg, at least up to libjpeg 0.8. Given TensorFlow's goal of portability across a wide varierty of platforms, I'm wondering if libjpeg was decided upon instead of libjpeg-turbo for portability, or was this just an omission? 

libjpeg-turbo speeds up JPEG decoding by about 2x, and since that reduces decoding latency and increases throughput, it not only improves performance of TensorFlow on the CPU, but especially on the GPU, which tends to wait for the CPU to provide decoded images. Moreover, that could result in a major performance improvement on more CPU-underpowered platforms such as ARM (e.g. via NEON SIMD on the NVIDIA TX1 platform which uses the NEON-capable Cortex A57, or on ARM Cortex A8 on a Samsung Galaxy S).

tatatodd:
@mkolod Thanks for filing this issue!

Switching to libjpeg-turbo sounds like a reasonable thing to investigate.  We welcome contributions!

If anyone is going to attempt this, probably the two areas to focus on are:
1. Benchmarks showing the actual performance when used within TensorFlow.
2. A breakdown of the portability tradeoffs between the two libraries.

Adding @prb12 @poxvoculi since they might be interested, and might know some of the history behind why libjpeg was chosen over libjpeg-turbo (if any).

jhseu:
We should definitely switch. libjpeg-turbo supports all the platforms we're on, and it seems that both Chrome and Firefox are already using it.

jhseu:
Self-assigning so I can test it out internally as well.

mkolod:
[...]

PR #5072, opened 2016-10-19
Use fast IDCT for JPEG decoding by default

mkolod:
Please refer to issue #4833 for details.

### DIFF
--- a/tensorflow/core/lib/jpeg/jpeg_mem.cc
+++ b/tensorflow/core/lib/jpeg/jpeg_mem.cc
@@ -139,8 +139,11 @@ uint8* UncompressLow(const void* srcdata, FewerArgsForCompiler* argball) {
   cinfo.do_fancy_upsampling = boolean(flags.fancy_upscaling);
   cinfo.scale_num = 1;
   cinfo.scale_denom = ratio;
-  // Activating this has a quality/speed trade-off implication:
-  // cinfo.dct_method = JDCT_IFAST;
+  // Activating this has a quality/speed trade-off implication.
+  // However, most JPEGs are already compressed, and so the faster IDCT
+  // should have no effect on training. The fast setting speeds up training on the
+  // GPU, e.g. by about 30% for smaller networks such as AlexNet.
+  cinfo.dct_method = JDCT_IFAST;
 
   jpeg_start_decompress(&cinfo);
 
--- a/tensorflow/core/lib/jpeg/jpeg_mem_unittest.cc
+++ b/tensorflow/core/lib/jpeg/jpeg_mem_unittest.cc
@@ -165,7 +165,7 @@ TEST(JpegMemTest, Jpeg2) {
     // Compare the two images
     const int totalerr = ComputeSumAbsoluteDifference(
         imgdata1.get(), refdata1.get(), in_w, in_h, stride1, stride1);
-    CHECK_LE(totalerr, 85000);
+    CHECK_LE(totalerr, 120000);
   }
 
   // check the second image too. Should be bitwise identical to the first.
--- a/tensorflow/python/ops/image_ops_test.py
+++ b/tensorflow/python/ops/image_ops_test.py
@@ -1803,10 +1803,10 @@ class JpegTest(test_util.TensorFlowTestCase):
       jpeg0, image0, image1, image2 = sess.run([jpeg0, image0, image1, image2])
 
       # The decoded-encoded image should be similar to the input
-      self.assertLess(self.averageError(image0, image1), 0.6)
+      self.assertLess(self.averageError(image0, image1), 0.7)
 
       # We should be very close to a fixpoint
-      self.assertLess(self.averageError(image1, image2), 0.02)
+      self.assertLess(self.averageError(image1, image2), 0.6)
 
       # Smooth ramps compress well (input size is 153600)
       self.assertGreaterEqual(len(jpeg0), 5000)


TimZaman:
Weird potential reviewers given by the bot. A human (#4833) suggested @prb12 and @poxvoculi..

mkolod:
BTW I'm from Nvidia (the org membership is part of my profile) and Nvidia has a corporate CLA. If you need me to do anything else (e.g. file a personal CLA as well) to keep the bots happy, please let me know. :)

drpngx:
@keveman Not sure what the implications are. It looks like we disabled this intentionally.

mkolod:
@drpngx @keveman For the record, Caffe and other frameworks use the default (i.e. fast) IDCT. With compressed JPEGs, the slow IDCT is very unlikely to give any accuracy gains, and slows down processing dramatically, especially when the CPU is the bottleneck for another component in the system that has much higher throughput, e.g. a GPU.

mkolod:
Here's a screenshot for ResNet-50, batch size 64, of convergence with default optimizer, learning rate, etc. I'm not sure if I can get full determinism even with the fixed seed, but in any case, the convergence seems similar after a bunch of iterations even if there's some early noisiness. Again, since the other frameworks use the fast IDCT, I'd think that it's rather innocuous. 
fast_vs_slow

mkolod:
I tried convergence on ResNet-50 and I don't see whether one is markedly better than the other. This was with the data loading queue and preprocessing queue all set to 1 thread, and using fixed initialization seeds, so this is as much determinism as I could imagine having control over. I'd imagine the Inception results would be very similar, and if anything, smaller nets like AlexNet would be even less sensitive to decoding issues.

fast_vs_slow

mkolod:
@drpngx @keveman Note that for e.g. AlexNet, the speed-up due to this small change is about 30% on a GTX 1080 and a Core i7-5930K, running off of a 500 MB/s SSD. That's a pretty significant speed-up. For multi-GPU cases, this will be even more important since then the bottleneck will be the CPU component even more so than in the single-GPU case. Also, it may matter even more on say ARM than on x86, including for the JPEG-based inference speeds.

drpngx:
I don't know about the implications of making that modification in all of users. My intuition would usually be to provide a flag, and consider flipping the default to new behavior.

@keveman any comment on the analysis provided?

mkolod:
Also, see [here](https://github.com/tensorflow/tensorflow/issues/4807#issuecomment-256718844) for perf impact. The importance of fast IDCT varies by image size, and of course its impact on training depends on how compute-heavy the model is (less impact for Inception, more for AlexNet).

keveman:

keveman (commented):
Seems good. Will merge after testing.

TimZaman:
Fails due to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/image_ops_test.py#L1806
Which error `AssertionError: 0.64828776041666669 not less than 0.6`
Barely overshoots the threshold - bearing in mind the range is [0:255]. I bet the magical thershold value of 0.6 is set because the non-IDCT gets something just below 0.6.

mkolod:
@TimZaman Updated to 0.7. I agree that the assertion's tolerance was based on what the slow IDCT would typically get. The fast IDCT's result is barely different, and that is to be expected due to algo difference, but isn't different for practical purposes.

keveman:

keveman:
@mkolod Can you please take a look at the failures?

drpngx:
It looks like they all fail on jpeg unit test, and mac additionally fails on env test

@gunan @caisq not sure if we bazel rc with test output on failure on macs

mkolod:
@keveman @drpngx I looked at the test and my findings are as follows. There isn't any bug in libjpeg, but, I would argue, the tolerance on the JPEG fixed point test (encode/decode/encode several times and see if there is stability in that the diffs go to zero) is pretty tight. For the first encode-decode, the diff is about 0.55 out of 255, but with the slow IDCT, the diff goes to 0.02 or so on the second encode/decode pass. The fast IDCT does generate more error, this is a known thing. The error is actually basically the same on the first pass in fast and slow IDCT, but the difference is that slow IDCT is more stable on the second decode and re-encode. I'd argue that since TensorFlow is not an image editing platform but a machine learning platform, it would be a real corner case if someone ended up decoding and re-encoding the same image dozens of times over (also in that case, why not use a lossless format?). If there is say one decoding and re-encoding (e.g. due to the need to preprocess an existing dataset offline for more optimized use later, e.g. pre-resizing ImageNet), that would be a single such pass, not counting the final decoding at learning time. So, while the fixed point test would matter a lot for image editing, I don't think it's that important for machine learning applications. If so, I can update the test case and re-submit. However, whether that is OK is of course up to you. Please let me what you think.

drpngx:
Sorry, I had lost context. Yes, we should bump up the epsilons in the tests to make them all pass.

mkolod:
@drpngx Will do, thanks!

mkolod:
@drpngx @keveman Should be fine for re-running tests now, I ran 
bazel test //tensorflow/python/...
locally on Linux and everything seemed fine.

mkolod:
@drpngx @keveman Should be fine for re-running tests now, I ran 
bazel test //tensorflow/python/...
locally on Linux and everything seemed fine. Could you re-run the CI pipeline?

keveman:

mkolod:
@keveman Apologies, looks like the Python tests pass but the C++ tests still fail. I thought all the testing of the transforms was in Python, hence I ran only the Python test suite. Will re-run all the tests including C++ ones.

mkolod:
@keveman I re-ran all the tests now (Python and C++) using 
```bazel test //tensorflow/...```
and I don't see anything failing in the mainline code on my Linux box.

vrv:
Okay let's kick off another test to find out!

vrv (approved):
I heartily approve of this PR.

vrv:
Okay, sadly I have to roll this change back -- it looks like some of our groups internally might rely on having higher quality decoding.  Let's re-make this change possibly via an option to DecodeJpegOp, rather than flipping the default everywhere.

vrv:
@mkolod: are you still working on getting this change back in again via the configurable option?  If not, would you like us to do it instead?

prb12:
> Let's re-make this change possibly via an option to DecodeJpegOp, rather than flipping the default everywhere.

+1   This should obviously be the default, but retaining the option for slow high quality is a good thing.
BTW, this op contributes significantly to multi-GPU inception training being bottlenecked on host CPU.

vrv:
Yes, we'll first add the option, then we can try to quickly find those that depend on the slow path for accuracy (not many, but there are some), and then we can switch the default after converting them to explicitly depend on the slow option.

PR #5076, opened 2016-10-19
windows specific eigen patch to enable gpu support for windows

guschmue:
compiling cuda kernels for windows fails. The following windows specific patch does temporarily fix this. Going to send the change also to eigen.

### DIFF
--- a/tensorflow/contrib/cmake/external/eigen.cmake
+++ b/tensorflow/contrib/cmake/external/eigen.cmake
@@ -46,3 +46,15 @@ ExternalProject_Add(eigen
         -DCMAKE_INSTALL_PREFIX:STRING=${eigen_INSTALL}
         -DINCLUDE_INSTALL_DIR:STRING=${CMAKE_CURRENT_BINARY_DIR}/external/eigen_archive
 )
+
+if(WIN32)
+    # temporarily patch eigen to workaround msvc+nvcc compile error
+    ExternalProject_Add_Step(eigen eigen_windows_patch
+        COMMAND ${CMAKE_COMMAND} -E copy
+            ${CMAKE_SOURCE_DIR}/patches/eigen/windows/Macros.h ${CMAKE_CURRENT_BINARY_DIR}/external/eigen_archive/Eigen/src/Core/util/
+        COMMAND ${CMAKE_COMMAND} -E copy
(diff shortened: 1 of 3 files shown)


drpngx:
@mrry maybe interested?
Auto notifier on original patch went to @ebrevdo and @ibab

mrry:
This change looks fine to me, but it'd be great if @benoitsteiner or @rmlarsen could comment on how we could get these changes upstream in Eigen (or if there's another workaround for compiling with NVCC on Windows (MSVC 2015)).

benoitsteiner (commented):
Doesn't https://bitbucket.org/eigen/eigen/commits/59db0512e12d98e8e223483f785931a076984288 address the problem with Macros.h ? If that's the case we only need to upgrade and fix the Triangular Solver code, which is a lot less risky.

guschmue:
going to take a look if that will work.

guschmue:
Yes, https://bitbucket.org/eigen/eigen/commits/59db0512e12d98e8e223483f785931a076984288 fixes the Macros.h issue and is the better fix. If you are planning to update the master with a new Eigen version that contains that fix I could take Macros.h out of this patch.

schmuell:
Triangular Solver is a side effect of how I changed Macros.h. If I take , https://bitbucket.org/eigen/eigen/commits/59db0512e12d98e8e223483f785931a076984288  instead, it seems fine. Clean master with only 59db0512e12d98e8e223483f785931a076984288  on top builds and runs fine.
So I could either update this pull request to only include 59db0512e12d98e8e223483f785931a076984288 or better if there is a chance that 59db0512e12d98e8e223483f785931a076984288  goes into the master  any time soon I could just close this PR.
Please advice.

drpngx:
Would this

  eigen_version = "1c7159a65db4"  eigen_sha256 =
"b089a6eae493c32703c6beb5fdae9d64a7667c3a5440bae00ac8e517cc822e62"work?
This is pending merge, probably today.

guschmue:
I think this will do. Going to watch for the merge and test it. If all is good I'll close this PR.

benoitsteiner:
1c7159a65db4 contains  https://bitbucket.org/eigen/eigen/commits/59db0512e12d98e8e223483f785931a076984288, so once the Eigen upgrade is merged the issue should be solved.

drpngx:
Merged #5105, let us know how this works out.

guschmue:
Verified: master builds and works on windows without this PR. Thank You!

drpngx:
Nice!

PR #5079, opened 2016-10-19
Fix typo in conv2d doc

Quarazy:
Currently, docs for conv2d:

``` markdown
For the most common case of the same
horizontal and **vertices** strides, `strides = [1, stride, stride, 1]`.
```

I believe it should be:

``` markdown
For the most common case of the same
horizontal and **vertical** strides, `strides = [1, stride, stride, 1]`.
```


caisq:

caisq (approved):

PR #5093, opened 2016-10-20
Do not set testonly to //tensorflow/python:construction_fails_go

damienmg:
This target is shipped into the pip_package so it is not testonly. Current
version of Bazel does not enforce testonly but the next version will.
Tested by building build_pip_package.

Tracking bug: bazelbuild/bazel#1967


gunan:
The change looks benign to me, but the "test" prefix in filename looks fishy to me.
Vijay, can you comment on this OP, is it a test only OP?

jart:
Closing in favor of cl/136944330. Thank you Herr Damien.

PR #5097, opened 2016-10-20
Update doc and release note for 0.11.0rc1

yifeif:


tensorflow/g3doc/api_docs/cc/ClassSession.md
```diff
@@ -6,36 +6,7 @@ When a Session is created with a given target, a new Session object is bound to
...
-// with it.
-auto output_tensor = outputs[0].flat<float>();
-if (output_tensor(0) > 0.5) { ... }
-
-// Close the session to release the resources associated with
-// this session.
-session->Close();
-
-```
+{c++}   tensorflow::GraphDef graph;  // ... Create or load graph into "graph".   // This example uses the default options which connects  // to a local runtime.  tensorflow::SessionOptions options;  std::unique_ptr<tensorflow::Session>  session(tensorflow::NewSession(options));   // Create the session with this graph.  tensorflow::Status s = session->Create(graph);  if (!s.ok()) { ... }   // Run the graph and fetch the first output of the "output"  // operation, and also run to but do not return anything  // for the "update_state" operation.  std::vector<tensorflow::Tensor> outputs;  s = session->Run({}, {"output:0"}, {"update_state"}, &outputs);  if (!s.ok()) { ... }   // Map the output as a flattened float tensor, and do something  // with it.  auto output_tensor = outputs[0].flat<float>();  if (output_tensor(0) > 0.5) { ... }   // Close the session to release the resources associated with  // this session.  session->Close();
```
drpngx: Not sure what happened to this

gunan: Interesting, can this be a bug in gen_docs.py?

yifeif: hmm looks like gen_docs hasn't changed for a while. Let me try re-generate again.

gunan (approved):

PR #5107, opened 2016-10-21
Fix cmake build by restoring changes from c25f125.

gunan:
Accidentally cmake build started depending on system zlib.
This PR fixes that.


drpngx:
Nice, thanks!

drpngx (approved):

PR #5103, opened 2016-10-21
Updating README adjusting to GPU support 

Carmezim:
Updating current limitations and adding instructions to use GPU on Cmake build.

PR #5124, opened 2016-10-21
Made mnist tutorial data storage folder consistent

willdzeng:
modified fully_connnected_feed.py train_dir to data_dir as other files
and made the default as /tmp/data
to avoid downloading data in the source folder.

### DIFF
--- a/tensorflow/examples/tutorials/mnist/fully_connected_feed.py
+++ b/tensorflow/examples/tutorials/mnist/fully_connected_feed.py
@@ -116,7 +116,7 @@ def run_training():
   """Train MNIST for a number of steps."""
   # Get the sets of images and labels for training, validation, and
   # test on MNIST.
-  data_sets = input_data.read_data_sets(FLAGS.train_dir, FLAGS.fake_data)
+  data_sets = input_data.read_data_sets(FLAGS.input_data_dir, FLAGS.fake_data)
 
   # Tell TensorFlow that the model will be built into the default Graph.
   with tf.Graph().as_default():
@@ -145,13 +145,13 @@ def run_training():
     init = tf.initialize_all_variables()
 
     # Create a saver for writing training checkpoints.
-    saver = tf.train.Saver()
+    saver = tf.train.Saver(write_version=tf.train.SaverDef.V2)
 
     # Create a session for running Ops on the Graph.
     sess = tf.Session()
 
     # Instantiate a SummaryWriter to output summaries and the Graph.
-    summary_writer = tf.train.SummaryWriter(FLAGS.train_dir, sess.graph)
+    summary_writer = tf.train.SummaryWriter(FLAGS.log_dir, sess.graph)
 
     # And then after everything is built:
 
@@ -189,7 +189,7 @@ def run_training():
 
       # Save a checkpoint and evaluate the model periodically.
       if (step + 1) % 1000 == 0 or (step + 1) == FLAGS.max_steps:
-        checkpoint_file = os.path.join(FLAGS.train_dir, 'checkpoint')
+        checkpoint_file = os.path.join(FLAGS.log_dir, 'model.ckpt')
         saver.save(sess, checkpoint_file, global_step=step)
         # Evaluate against the training set.
         print('Training Data Eval:')
@@ -215,6 +215,9 @@ def run_training():
 
 
 def main(_):
+  if tf.gfile.Exists(FLAGS.log_dir):
+    tf.gfile.DeleteRecursively(FLAGS.log_dir)
+  tf.gfile.MakeDirs(FLAGS.log_dir)
   run_training()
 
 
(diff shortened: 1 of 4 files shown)


tensorflow/examples/tutorials/mnist/fully_connected_feed.py
```diff
@@ -151,7 +151,7 @@ def run_training():
     sess = tf.Session()
 
     # Instantiate a SummaryWriter to output summaries and the Graph.
-    summary_writer = tf.train.SummaryWriter(FLAGS.train_dir, sess.graph)
+    summary_writer = tf.train.SummaryWriter(FLAGS.data_dir, sess.graph)
```
drpngx: This should probably stay in train_dir

tensorflow/examples/tutorials/mnist/fully_connected_feed.py
```diff
@@ -189,7 +189,7 @@ def run_training():
 
       # Save a checkpoint and evaluate the model periodically.
       if (step + 1) % 1000 == 0 or (step + 1) == FLAGS.max_steps:
-        checkpoint_file = os.path.join(FLAGS.train_dir, 'checkpoint')
+        checkpoint_file = os.path.join(FLAGS.data_dir, 'checkpoint')
```
drpngx: This probably belongs in train

tensorflow/examples/tutorials/mnist/fully_connected_feed.py
```diff
@@ -251,9 +251,9 @@ def main(_):
       help='Batch size.  Must divide evenly into the dataset sizes.'
   )
   parser.add_argument(
-      '--train_dir',
+      '--data_dir',
```
drpngx: We probably want to keep those separate. It could be an issue with naming, for instance,
output_dir
input_data_dir
?
(open to ideas -- but generally, I think it's a bad idea to mix input and output in the same tree)

tensorflow/examples/tutorials/mnist/fully_connected_feed.py
willdzeng: it doesn't matter, I think, it basically write the graph to the directory, it has nothing to do with whether it's training data or not

tensorflow/examples/tutorials/mnist/fully_connected_feed.py
willdzeng: I get what you saying here, maybe we need to create another directory which storage all the graph and checkpoint by the model name and type
 for example
/tmp/data/mnist/input_data
/tmp/data/mnist/output_data
/tmp/data/imagenet/input_data
/tmp/data/imagenet/output_data
so that when we run different model in a day we don't need to remove the other graph file to make it clean

tensorflow/examples/tutorials/mnist/fully_connected_feed.py
drpngx: Basically, I think it's good practice to separate the input from the output.

tensorflow/examples/tutorials/mnist/fully_connected_feed.py
willdzeng: something like this?
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/mnist/mnist_with_summaries.py#L201

willdzeng:
@drpngx Can you check if the new commit is good?

One thing I noteiced is the copyright year of all the code is still 2015, and my sublime keep updating the copyright year to 2015-2016 itself. I am wondering if you guys want to update that.

drpngx:
Nice, thanks!

drpngx:
Jenkins, test this please.

drpngx (approved):

drpngx:
It fails the test, can you fix?

willdzeng:
I am trying to fix, it's weird. I will switch to python 3 and test myself.

willdzeng:
Added Make directory of log_dir to avoid non-existed directory that may cause failure on other platforms.

`if tf.gfile.Exists(FLAGS.log_dir):
    tf.gfile.DeleteRecursively(FLAGS.log_dir)
  tf.gfile.MakeDirs(FLAGS.log_dir)`

drpngx:
Jenkins, test this please.

drpngx:
Still broken. Do you want to take another stab at it?

willdzeng:
Is there any way to pull the log? I tested on my with Linux python3, it works for me.

drpngx:
Sorry, build appears to be broken right now, possibly due to #5127. See discussion on #5103.

willdzeng:
I see. I just updated the Saver into V2.

willdzeng:
@drpngx I pulled the recent changes from master into the branch, does the build fixed? can you ask Jenkins to give another trial?

drpngx:
Jenkins, test this please.

drpngx:
Looks like there's some kind of windows error. I think we're waiting on a fix for that.

drpngx:
Nice find, yes!

willdzeng:
@drpngx I found out it's because the BUILD file having extra arguments and causing the failure on python3 test. The new commit should fix my test failure.

drpngx:
Jenkins, test this please.

PR #5135, opened 2016-10-22
Option to config python lib path from env var

villasv:
The simplest way to give the option to configure python library path from environment variables, targeting the issue https://github.com/tensorflow/serving/issues/216 about a scriptable complete configuration for tensorflow + serving.

### DIFF
--- a/util/python/python_config.sh
+++ b/util/python/python_config.sh
@@ -113,28 +113,32 @@ function setup_python {
     echo -e "\n\nERROR: Problem getting python include path.  Is distutils installed?"
     exit 1
   fi
-  local python_lib_path
-  # Split python_path into an array of paths, this allows path containing spaces
-  IFS=','
-  python_lib_path=($(python_path))
-  unset IFS
-  echo "Found possible Python library paths:"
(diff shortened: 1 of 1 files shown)


villasv:
@googlebot Done.

util/python/python_config.sh
```diff
@@ -113,29 +113,33 @@ function setup_python {
...
+    if [ "$b" == "" ]; then
+      PYTHON_LIB_PATH="$(default_python_path "${python_lib_path[0]}")"
+      echo PYTHON_LIB_PATH
     else
-      echo -e "\n\nERROR: The path you have entered does not exist."
-      exit 1
+      PYTHON_LIB_PATH="$b"
     fi
   fi
+  if test -d "$PYTHON_LIB_PATH" -a -x "$PYTHON_LIB_PATH"; then
```
drpngx: Why not an else branch? I guess this will only work with a single directory?

drpngx: Ah, ok, that's fine then.

drpngx:
Jenkins, test this please.

drpngx (approved):

PR #5127, opened 2016-10-21
extend tf.select to broadcast a scalar condition #3945

Mistobaan:
continuation of pull request GH-3995

PR #5139, opened 2016-10-22
Revert "extend tf.select to broadcast a scalar condition #3945"

drpngx:
Reverts tensorflow/tensorflow#5127


drpngx:
Original PR might have broken the build.

drpngx:
@guschmue working on a forward fix. See discussion in #5103.

PR #5140, opened 2016-10-22
Do not use non-portable `cp --parents`

fp:
Mac OS `cp` command does not support the `--parents` option, so
use `mkdir -p` + `cp` instead.

### DIFF
--- a/tensorflow/tools/pip_package/build_pip_package.sh
+++ b/tensorflow/tools/pip_package/build_pip_package.sh
@@ -107,7 +107,8 @@ function main() {
   mkdir -p ${TMPDIR}/third_party
   pushd ${RUNFILES%org_tensorflow}
   for header in $(find protobuf -name \*.h); do
-    cp --parents "$header" ${TMPDIR}/google;
+    mkdir -p "${TMPDIR}/google/$(dirname ${header})"
+    cp "$header" "${TMPDIR}/google/$(dirname ${header})/"
   done
   popd
   cp -R $RUNFILES/third_party/eigen3 ${TMPDIR}/third_party


fp:
I signed it!

tensorflow/tools/pip_package/build_pip_package.sh
```diff
@@ -107,7 +107,7 @@ function main() {
   mkdir -p ${TMPDIR}/third_party
   pushd ${RUNFILES%org_tensorflow}
   for header in $(find protobuf -name \*.h); do
-    cp --parents "$header" ${TMPDIR}/google;
+    rsync -R "$header" ${TMPDIR}/google;
```
caisq: Thanks for offering the fix. 
This script is used on multiple platforms. I'm a little worried that rsync may be missing from some of them. Can we do the safe thing here, namely something like
DEST_DIR="${TMPDIR}/google/$(dirname ${header})"
mkdir -p "${DEST_DIR}"
cp "${header}" "${DEST_DIR}/"

caisq: Just noticed that there is a variable called "dest_dir" in the same file. So "DEST_DIR" should be renamed to something else to avoid confusion.

fp:
@caisq As suggested, but without extra variable
```
...
++ find protobuf -name '*.h'
+ for header in $(find protobuf -name \*.h)
++ dirname protobuf/src/google/protobuf/any.h
+ mkdir -p /var/folders/kf/thhjvp791mj9vh08w165m2jw0000gp/T/tmp.XXXXXXXXXX.wrghdibj/google/protobuf/src/google/protobuf
+ cp protobuf/src/google/protobuf/any.h /var/folders/kf/thhjvp791mj9vh08w165m2jw0000gp/T/tmp.XXXXXXXXXX.wrghdibj/google
+ for header in $(find protobuf -name \*.h)
++ dirname protobuf/src/google/protobuf/any.pb.h
+ mkdir -p /var/folders/kf/thhjvp791mj9vh08w165m2jw0000gp/T/tmp.XXXXXXXXXX.wrghdibj/google/protobuf/src/google/protobuf
...
```

fp:
@caisq apologies for that. now it should do the right thing:
```
...
++ find protobuf -name '*.h'
+ for header in $(find protobuf -name \*.h)
++ dirname protobuf/src/google/protobuf/any.h
+ mkdir -p /var/folders/kf/thhjvp791mj9vh08w165m2jw0000gp/T/tmp.XXXXXXXXXX.wMGQ1CqG/google/protobuf/src/google/protobuf
++ dirname protobuf/src/google/protobuf/any.h
+ cp protobuf/src/google/protobuf/any.h /var/folders/kf/thhjvp791mj9vh08w165m2jw0000gp/T/tmp.XXXXXXXXXX.wMGQ1CqG/google/protobuf/src/google/protobuf/
+ for header in $(find protobuf -name \*.h)
++ dirname protobuf/src/google/protobuf/any.pb.h
+ mkdir -p /var/folders/kf/thhjvp791mj9vh08w165m2jw0000gp/T/tmp.XXXXXXXXXX.wMGQ1CqG/google/protobuf/src/google/protobuf
++ dirname protobuf/src/google/protobuf/any.pb.h
+ cp protobuf/src/google/protobuf/any.pb.h /var/folders/kf/thhjvp791mj9vh08w165m2jw0000gp/T/tmp.XXXXXXXXXX.wMGQ1CqG/google/protobuf/src/google/protobuf/
...
```

drpngx:
@gunan who pointed out the issue.

You should change the title since it no longer matches the change.

tensorflow/tools/pip_package/build_pip_package.sh
```diff
@@ -107,7 +107,8 @@ function main() {
   mkdir -p ${TMPDIR}/third_party
   pushd ${RUNFILES%org_tensorflow}
   for header in $(find protobuf -name \*.h); do
-    cp --parents "$header" ${TMPDIR}/google;
+    mkdir -p "${TMPDIR}/google/$(dirname ${header})"
+    cp "$header" "${TMPDIR}/google/$(dirname ${header})/"
```
gunan: Not very important, but a minor nit.
Do we need to call "dirname" here?
The destination file can be specified, I do not see an issue with that in this line.

gunan:
Thank you very much for the fix.
@meteorcloudy FYI.
The change looks good to me except for the one nit.
I will leave the approval to @caisq as I am more confident with his bash-fu :)

tensorflow/tools/pip_package/build_pip_package.sh
caisq: I think calling dirname at least once (in the first newly added line) is necessary. But gunan@ is right that the second new line doesn't need it. It can just be something like,
cp "$header" "${TMPDIR}/google/${header}"

But it's not a big deal. I'll approve the PR as is.

caisq (approved):

caisq:
@tensorflow-jenkins , test this please.

meteorcloudy:
@fp  Thanks for the fix!
@gunan Can you also make ci run `build_pip_package.sh` so that we can catch the breakage earlier?

caisq:
@meteorcloudy The reason why this was not caught before merging was because we do not test PIP install on Mac in PR tests. We test that only in nightly builds. The PR tests test only PIP install on Linux. It's a tradeoff between the speed of testing and thoroughness of testing.

PR #5148, opened 2016-10-23
Fix merging of summaries to match the tutorial "TensorFlow Mechanics 101"

nikikilbertus:
The example does not work with `tf.summary.merge_all()`. Changing it to `tf.merge_all_summaries()` like in the [tutorial](https://www.tensorflow.org/versions/r0.11/tutorials/mnist/tf/index.html#train-loop) fixes the issue.


nikikilbertus:
I signed it.

drpngx (approved):

zheng-xq:

tensorflow/models/image/cifar10/cifar10_eval.py
```diff
@@ -81,36 +81,36 @@ def eval_once(saver, summary_writer, top_k_op, summary_op):
       print('No checkpoint file found')
       return
 
-    # Start the queue runners.
```
drpngx: That code block is not reachable since it's after `return`?

drpngx (changes requested):

tensorflow/models/image/cifar10/cifar10_eval.py
nikikilbertus: Sorry, my bad. I had it dedented in my fork and then accidentally indented one level too much.

danmane:
I made this change originally. Using tf.summary.merge_all() is definitely intended, as for tf1.0 we are moving all the summary ops into the tf.summary namespace. I will dig into why the example is broken and fix it while keeping usage of tf.summary.merge_all()

PR #5157, opened 2016-10-24
move type2index into EIGEN_HAS_INDEX_LIST

guschmue:
https://github.com/tensorflow/tensorflow/pull/5127 has a small issue that type2index / IndexList is used without checking for EIGEN_HAS_INDEX_LIST which broke the windows gpu build.
This makes it build again.

### DIFF
--- a/tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
+++ b/tensorflow/core/kernels/cwise_op_gpu_select.cu.cc
@@ -39,7 +39,11 @@ struct SelectScalarFunctor<GPUDevice, T> {
                   typename TTypes<T>::ConstFlat then_flat,
                   typename TTypes<T>::ConstFlat else_flat) {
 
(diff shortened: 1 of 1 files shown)


Mistobaan:
can you invert the if condition? `#if defined #else ... #endif`. This to be able to remove the `!`. I had a feeling that I had to do something like this from the single example that I found in TF using these functions.

guschmue:
certainly can switch the ifdef. But there is the same ifdef in line 67. Switch both or none ?

Mistobaan:
oh I see, no then is better to leave it like this for consistency.

guschmue:
Cool, thank you. Btw, the missing ifdef: I'd not have seen it either. But next time we know :)

mrry:
Thanks for the fix, Guenther!

mrry:

mrry (approved):

benoitsteiner:
Jenkins, test this please.

PR #5164, opened 2016-10-24
Orthogonal initializer

tillahoffmann:
This PR adds an orthogonal initializer based on the paper [Exact solutions to the nonlinear dynamics of learning in deep linear neural networks](https://arxiv.org/abs/1312.6120).


tensorflow/python/ops/init_ops.py
```diff
@@ -340,3 +341,47 @@ def __call__(self, shape, dtype=dtypes.float32, partition_info=None):
...
+    elif isinstance(seed, int):
+      rng = np.random.RandomState(seed)
+    else:
+      assert hasattr(seed, 'normal'), "rng must be None, an integer seed, or an object with a member function 'normal'"
+      rng = seed
+
+    # Flatten the input shape with the last dimension remaining its original shape so it works for conv2d
+    flat_shape = (np.prod(shape[:-1]), shape[-1])
+    a = rng.normal(0, 1, flat_shape)
+    u, _, v = np.linalg.svd(a, full_matrices=False)
```
zheng-xq: My biggest concern is why using np for this svd operation, instead of going through the tf.svd.

tillahoffmann: Mostly out of familiarity with numpy. I'll have a look at porting to tensorflow.

tensorflow/python/ops/init_ops.py
```diff
@@ -340,3 +341,45 @@ def __call__(self, shape, dtype=dtypes.float32, partition_info=None):
...
+    num_cols = shape[-1]
+    flat_shape = (num_rows, num_cols)
+
+    if gain == 'relu':
+      _gain = math.sqrt(2)
+    else:
+      _gain = gain
+
+    # Generate a random matrix
+    a = random_ops.random_normal(flat_shape, dtype=dtype, seed=seed)
```
tillahoffmann: What convention do you use for naming nested operations? Would a `name_scope` be appropriate here?

vrv:
ping for @zheng-xq

vrv:
Or ping for @rmlarsen

tensorflow/python/ops/init_ops.py
```diff
@@ -340,3 +341,45 @@ def __call__(self, shape, dtype=dtypes.float32, partition_info=None):
...
+  def _initializer(shape, dtype=_assert_float_dtype(dtype), partition_info=None):
+    # Flatten the input shape with the last dimension remaining its original shape so it works for conv2d
+    num_rows = 1
+    for dim in shape[:-1]:
+      num_rows *= dim
+    num_cols = shape[-1]
+    flat_shape = (num_rows, num_cols)
+
+    if gain == 'relu':
+      _gain = math.sqrt(2)
```
zheng-xq: Why is 'relu' special here? It is not documented in the arguments.

tensorflow/python/ops/init_ops.py
```diff
@@ -340,3 +341,45 @@ def __call__(self, shape, dtype=dtypes.float32, partition_info=None):
...
                                    seed=self._seed)
+
+
+def orthogonal_initializer(gain=1.0, dtype=dtypes.float32, seed=None):
+  """Returns an initializer that generates an orthogonal rotation matrix.
+
+  Initialize a tensor using an orthogonal matrix.
+
+  Args:
+    gain: multiplicative factor to apply to the orthogonal matrix
```
zheng-xq: Since the user can do a multiplication easier outside of this op, I don't see why it should belong in this op. The implementation doesn't take advantage of it to provide any performance benefit anyway.

tensorflow/python/ops/init_ops.py
zheng-xq: random_normal is significantly more expensive than random_uniform. Is there any reason not to use uniform?

tensorflow/python/ops/init_ops.py
```diff
@@ -340,3 +341,45 @@ def __call__(self, shape, dtype=dtypes.float32, partition_info=None):
...
+    else:
+      _gain = gain
+
+    # Generate a random matrix
+    a = random_ops.random_normal(flat_shape, dtype=dtype, seed=seed)
+    # Compute the svd
+    _, u, v = linalg_ops.svd(a, full_matrices=False)
+    # Pick the right singular value decomposition
+    q = u if num_cols < num_rows else v
+    return _gain * array_ops.reshape(q, shape)
```
zheng-xq: I think u is (num_rows x num_rows), and v is (num_cols x num_cols). tf.reshape requires the old shape and the new shape to be identical. So there is no guarantee this would work.

tensorflow/python/ops/init_ops.py
```diff
@@ -340,3 +341,45 @@ def __call__(self, shape, dtype=dtypes.float32, partition_info=None):
...
+      for behavior.
+
+  Returns:
+    An initializer that generates orthogonal tensors
+
+  Raises:
+    ValueError: if `dtype` is not a floating point type.
+  """
+  def _initializer(shape, dtype=_assert_float_dtype(dtype), partition_info=None):
+    # Flatten the input shape with the last dimension remaining its original shape so it works for conv2d
```
zheng-xq: I think we need to clearly document what this shape means in the docstring: what is supported, and what is orthogonal as the output. 

Since even conv has multiple data formats, NHWC and NCHW, it is a good idea to separate them clearly. It is a good idea to keep this op has simple as possible, and requires the shape to be square matrices.

tensorflow/python/kernel_tests/init_ops_test.py
```diff
@@ -444,5 +444,47 @@ def testDevice(self):
...
+      init2 = tf.orthogonal_initializer(gain=3.14, seed=1, dtype=dtype)
+      with self.test_session(graph=tf.Graph()):
+        t1 = init1([num]).eval()
+      with self.test_session(graph=tf.Graph()):
+        t2 = init2([num]).eval()
+      return np.allclose(t1, t2 / 3.14, rtol=1e-15, atol=1e-15)
+
+  def testShapes(self):
+    for dtype in [tf.float32, tf.float64]:
+      for shape in [(10, 10), (10, 9, 8), (100, 5, 5)]:
```
zheng-xq: You need to test the results are actual orthogonal.

tensorflow/python/kernel_tests/init_ops_test.py
```diff
@@ -444,5 +444,47 @@ def testDevice(self):
...
+      tf.orthogonal_initializer, dtype=tf.string)
+
+  def testGain(self):
+    num = 100
+    for dtype in [tf.float32, tf.float64]:
+      init1 = tf.orthogonal_initializer(seed=1, dtype=dtype)
+      init2 = tf.orthogonal_initializer(gain=3.14, seed=1, dtype=dtype)
+      with self.test_session(graph=tf.Graph()):
+        t1 = init1([num]).eval()
+      with self.test_session(graph=tf.Graph()):
```
zheng-xq: If you want to test both the GPU and CPU implementation, add "use_gpu=True" here. This invokes the gpu kernels when "--config=cuda" is used, and use the cpu kernels when it is not.

tensorflow/python/ops/init_ops.py
tillahoffmann: Fair point. Will drop the gain.

tensorflow/python/kernel_tests/init_ops_test.py
vrv: It's redundant now -- we haven't cleaned all the other cases up though.

use_gpu=True will try to run on GPU if available, and pass otherwise.

We have CI builds that run on GPU and CPU, so they will eventually get tested on both.

tensorflow/python/ops/init_ops.py
tillahoffmann: [`svd`](https://www.tensorflow.org/versions/r0.11/api_docs/python/math_ops.html#svd), given a matrix of shape `[..., M, N]`, returns `u` of shape `[..., M, P]` and `v` of shape `[..., N, P]`, where `P` is the minimum of `M` and `N` (assuming `full_matrices=False`). So I still need to roll axes here to get things right.

As an aside, is it intentional, to have the return value of `svd` in tensorflow be different from the return value of `np.svd`?

tensorflow/python/ops/init_ops.py
tillahoffmann: I would be keen to keep the functionality as is but document more clearly as you suggested. The main reason is that I find

```python
weights = tf.get_variable('weights', (5, 5, 8, 32), tf.float32, 
                          tf.orthogonal_initializer())
```

much more straightforward than

```python
weights = tf.reshape(
  tf.get_variable('weights', (200, 32), tf.float32, tf.orthogonal_initializer()), 
  (5, 5, 8, 32)
)
```

tensorflow/python/ops/init_ops.py
tillahoffmann: No, I think uniform should be fine here.

tensorflow/python/ops/init_ops.py
tillahoffmann: On second thoughts, I think keeping the gain is useful because it simplifies the setup of variables (similar to not having to call `reshape`). It's more or less equivalent to the standard deviation for the `random_normal` initializer. Thoughts?

tensorflow/python/ops/init_ops.py
```diff
@@ -340,3 +341,55 @@ def __call__(self, shape, dtype=dtypes.float32, partition_info=None):
...
+    num_rows = 1
+    for dim in shape[:-1]:
+      num_rows *= dim
+    num_cols = shape[-1]
+    flat_shape = (num_rows, num_cols)
+
+    # Generate a random matrix
+    a = random_ops.random_uniform(flat_shape, dtype=dtype, seed=seed)
+    # Compute the svd
+    _, u, v = linalg_ops.svd(a, full_matrices=False)
```
rmlarsen: The SVD is an overly expensive algorithm to use for this purpose, we should have a QR factorization op for this. [No action required.]

rmlarsen (approved):

rmlarsen:

rmlarsen:

tillahoffmann:
The failing tests on Windows emit error messages of the form 

```
11:16:03 ======================================================================
11:16:03 ERROR: testGain (__main__.OrthogonalInitializerTest)
11:16:03 ----------------------------------------------------------------------
11:16:03 Traceback (most recent call last):
11:16:03   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\client\session.py", line 1021, in _do_call
11:16:03     return fn(*args)
11:16:03   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\client\session.py", line 999, in _run_fn
11:16:03     self._extend_graph()
11:16:03   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\client\session.py", line 1048, in _extend_graph
11:16:03     self._session, graph_def.SerializeToString(), status)
11:16:03   File "C:\Program Files\Anaconda3\lib\contextlib.py", line 66, in __exit__
11:16:03     next(self.gen)
11:16:03   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\framework\errors_impl.py", line 469, in raise_exception_on_not_ok_status
11:16:03     pywrap_tensorflow.TF_GetCode(status))
11:16:03 tensorflow.python.framework.errors_impl.InvalidArgumentError: No OpKernel was registered to support Op 'Svd' with these attrs.  Registered kernels:
11:16:03   <no registered kernels>
11:16:03 
11:16:03 	 [[Node: Svd = Svd[T=DT_FLOAT, compute_uv=true, full_matrices=false](random_uniform)]]
```

Unfortunately, I'm not quite sure what to make of that.

rmlarsen:
@mrry Derek, do you understand why this failure occurs on Windows but not Linux or Mac?

mrry:
Yes, IIRC there was a problem compiling the SVD op, so it was originally excluded from the (CMake) build. I can look into it later, or you could try commenting out [this line](https://github.com/tensorflow/tensorflow/blob/7f3d8e1f8736f4c812e4a0a96154af3bd3750180/tensorflow/contrib/cmake/tf_core_kernels.cmake#L93) in this CL and running the tests again....

rmlarsen:
@mrry Thanks for the explanation, Derek. I think we should probably get the SVD op working on Windows before merging this PR then?

mrry:
@rmlarsen Sounds good. I'll look at getting the SVD op compiled today.

rmlarsen:

tensorflow/python/ops/init_ops.py
poulson-google: Agreed. And, further, I would recommend generating the entries with a normal distribution and then taking the Q from the QR factorization so that a Haar matrix is generated. Diaconis has a nice paper that discusses this and an alternative implicit version: http://www.ams.org/journals/bull/2003-40-02/S0273-0979-03-00975-3/S0273-0979-03-00975-3.pdf

rmlarsen:

rmlarsen:
Our Mac tests are rather flaky at them moment :-1: . I'll keep running the tests until they pass at least once...

PR #5166, opened 2016-10-24
Fix incorrect link for Building a Random Forest Model

wangyum:
Replace `random_forest.py` with `random_forest_mnist.py`, because `random_forest.py` is not exist.


caisq:

caisq (approved):

caisq:
Thank you for the fix, @wangyum

PR #5167, opened 2016-10-24
Fix bad var echoing introduced by #5135

villasv:
A minor mistake slipped through PR #5135


caisq (approved):

caisq:

caisq:
Thank you for the fix, @villasv

PR #5171, opened 2016-10-24
Fixed typo in train api documentation

oskopek:
"It `True`" -> "If `True`" in the documentation for `tf.train.exponential_decay`


drpngx:
Thanks for the fix, could fix upstream tensorflow/python/training/learning_rate_decay.py?

PR #5146, opened 2016-10-23
Cherrypicks and rc2.

gunan:

PR #5173, opened 2016-10-24
Fix contrib/makefile builds in 0.11

avalys:
contrib/makefile builds include version_info.cc twice and end up with duplicate symbols in the resulting library.  As a result none of the iOS examples work and the resulting library can't be used.

This is simply the fix from master, https://github.com/tensorflow/tensorflow/commit/1283b84a49a9f5e14aca833cf981b61848aaf916.


gunan:
Already cherrypicked into the branch:
https://github.com/tensorflow/tensorflow/pull/5146
Closing this one.

PR #5176, opened 2016-10-24
Improved mnist/convolution.py performance

willdzeng:


yaroslavvb:
LGTM

tensorflow/models/image/mnist/convolutional.py
```diff
@@ -296,13 +297,15 @@ def eval_in_batches(data, sess):
...
-      _, l, lr, predictions = sess.run(
-          [optimizer, loss, learning_rate, train_prediction],
-          feed_dict=feed_dict)
+      # Run the optimizer.
+      sess.run(optimizer, feed_dict = feed_dict)
+      # print some extra information once reach the evaluation frequency
       if step % EVAL_FREQUENCY == 0:
-        elapsed_time = time.time() - start_time
-        start_time = time.time()
+        # fetch some extra nodes data
```
vincentvanhoucke: This is not just fetching extra nodes. It's rerunning the graph with these nodes, which means it consumes a batch of data that's not used in training.

vincentvanhoucke (commented):
Since this is an example, I prefer simplicity and readability over performance.

willdzeng:
@vincentvanhoucke sure, it's done.
This code is very helpful in learning tenrsorflow and convolution neural network, my modification is trivial.
Thank you.

tensorflow/models/image/mnist/convolutional.py
```diff
@@ -296,11 +296,13 @@ def eval_in_batches(data, sess):
       # node in the graph it should be fed to.
       feed_dict = {train_data_node: batch_data,
                    train_labels_node: batch_labels}
-      # Run the graph and fetch some of the nodes.
-      _, l, lr, predictions = sess.run(
-          [optimizer, loss, learning_rate, train_prediction],
-          feed_dict=feed_dict)
+      # Run the optimizer to update weights.
+      sess.run(optimizer, feed_dict = feed_dict)
```
tillahoffmann: Does omitting the three fetches really make that much of a difference? They should have been computed already so the only overhead should be the transfer from the session to numpy arrays.

vincentvanhoucke: It's 3 fetches plus the 10-way softmax for train_predictions, traded off against recomputing the whole graph once every 1/100 steps. I am surprised it makes any sort of difference, to be honest. @willdzeng what's your setup?

vincentvanhoucke: Style guide says `feed_dict=feed_dict` here and below.

tensorflow/models/image/mnist/convolutional.py
```diff
@@ -296,11 +296,13 @@ def eval_in_batches(data, sess):
...
+      # print some extra information once reach the evaluation frequency
       if step % EVAL_FREQUENCY == 0:
+        # fetch some extra nodes' data
+        l, lr, predictions = sess.run([loss, learning_rate, train_prediction],
```
vincentvanhoucke: wrap to 80 cols

tensorflow/models/image/mnist/convolutional.py
willdzeng: I am using i76900k and GTX980. The major improvement is from [_train_prediction_,](https://github.com/willdzeng/tensorflow/blob/6063e6062d920cb4b666c29b5c49f37521cf3edc/tensorflow/models/image/mnist/convolutional.py#L255) which is not calculated every time in _optimizer_.
It's around 10% improvement from 110s to 95s for me.

vincentvanhoucke (approved):
LGTM

vincentvanhoucke:

PR #5194, opened 2016-10-25
Fix confusing CTC documentation

al-muammar:
From the docs it seems that CTC decoder have problems and it can't handle labelings like `A B B`, but if you will look at the [code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/ctc/ctc_decoder.h#L84), it is not the case, everything is OK, just typo in the docs.

### DIFF
--- a/tensorflow/core/ops/ops.pbtxt
+++ b/tensorflow/core/ops/ops.pbtxt
@@ -3536,7 +3536,7 @@ op {
     description: "If True, merge repeated classes in output."
   }
   summary: "Performs greedy decoding on the logits given in inputs."
-  description: "A note about the attribute merge_repeated: if enabled, when\nconsecutive logits\' maximum indices are the same, only the first of\nthese is emitted.  Labeling the blank \'*\', the sequence \"A B B * B B\"\nbecomes \"A B\" if merge_repeated = True and \"A B B B B\" if\nmerge_repeated = False.\n\nRegardless of the value of merge_repeated, if the maximum index of a given\ntime and batch corresponds to the blank, index `(num_classes - 1)`, no new\nelement is emitted."
+  description: "A note about the attribute merge_repeated: if enabled, when\nconsecutive logits\' maximum indices are the same, only the first of\nthese is emitted.  Labeling the blank \'*\', the sequence \"A B B * B B\"\nbecomes \"A B B\" if merge_repeated = True and \"A B B B B\" if\nmerge_repeated = False.\n\nRegardless of the value of merge_repeated, if the maximum index of a given\ntime and batch corresponds to the blank, index `(num_classes - 1)`, no new\nelement is emitted."
 }
 op {
   name: "CTCLoss"
--- a/tensorflow/core/ops/ops.pbtxt
+++ b/tensorflow/core/ops/ops.pbtxt
@@ -3536,7 +3536,7 @@ op {
     description: "If True, merge repeated classes in output."
   }
   summary: "Performs greedy decoding on the logits given in inputs."
-  description: "A note about the attribute merge_repeated: if enabled, when\nconsecutive logits\' maximum indices are the same, only the first of\nthese is emitted.  Labeling the blank \'*\', the sequence \"A B B * B B\"\nbecomes \"A B B\" if merge_repeated = True and \"A B B B B\" if\nmerge_repeated = False.\n\nRegardless of the value of merge_repeated, if the maximum index of a given\ntime and batch corresponds to the blank, index `(num_classes - 1)`, no new\nelement is emitted."
+  description: "A note about the attribute merge_repeated: if enabled, when\nconsecutive logits\' maximum indices are the same, only the first of\nthese is emitted.  Labeling the blank \'*\', the sequence \"A B B * B B\"\nbecomes \"A B\" if merge_repeated = True and \"A B B B B\" if\nmerge_repeated = False.\n\nRegardless of the value of merge_repeated, if the maximum index of a given\ntime and batch corresponds to the blank, index `(num_classes - 1)`, no new\nelement is emitted."
 }
 op {
   name: "CTCLoss"
--- a/tensorflow/g3doc/api_docs/python/functions_and_classes/shard6/tf.nn.ctc_greedy_decoder.md
+++ b/tensorflow/g3doc/api_docs/python/functions_and_classes/shard6/tf.nn.ctc_greedy_decoder.md
@@ -11,8 +11,8 @@ This means that if consecutive logits' maximum indices are the same,
 only the first of these is emitted.  The sequence `A B B * B * B` (where '*'
 is the blank label) becomes
 
-  * `A B` if `merge_repeated=True`.
-  * `A B B B B B` if `merge_repeated=False`.
+  * `A B B B` if `merge_repeated=True`.
+  * `A B B B B` if `merge_repeated=False`.
 
 ##### Args:
 
--- a/tensorflow/g3doc/api_docs/python/functions_and_classes/shard6/tf.nn.ctc_greedy_decoder.md
+++ b/tensorflow/g3doc/api_docs/python/functions_and_classes/shard6/tf.nn.ctc_greedy_decoder.md
@@ -11,8 +11,8 @@ This means that if consecutive logits' maximum indices are the same,
 only the first of these is emitted.  The sequence `A B B * B * B` (where '*'
 is the blank label) becomes
 
-  * `A B B B` if `merge_repeated=True`.
-  * `A B B B B` if `merge_repeated=False`.
+  * `A B` if `merge_repeated=True`.
+  * `A B B B B B` if `merge_repeated=False`.
 
 ##### Args:
 
--- a/tensorflow/python/ops/ctc_ops.py
+++ b/tensorflow/python/ops/ctc_ops.py
@@ -180,7 +180,7 @@ def ctc_greedy_decoder(inputs, sequence_length, merge_repeated=True):
   only the first of these is emitted.  The sequence `A B B * B * B` (where '*'
   is the blank label) becomes
 
-    * `A B` if `merge_repeated=True`.
+    * `A B B B` if `merge_repeated=True`.
     * `A B B B B B` if `merge_repeated=False`.
 
   Args:
--- a/tensorflow/python/ops/ctc_ops.py
+++ b/tensorflow/python/ops/ctc_ops.py
@@ -181,7 +181,7 @@ def ctc_greedy_decoder(inputs, sequence_length, merge_repeated=True):
   is the blank label) becomes
 
     * `A B B B` if `merge_repeated=True`.
-    * `A B B B B B` if `merge_repeated=False`.
+    * `A B B B B` if `merge_repeated=False`.
 
   Args:
     inputs: 3-D `float` `Tensor` sized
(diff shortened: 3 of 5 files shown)


Jihadik:
I signed it!

igormq:
Hi, @Jihadik.

If you try this piece of code

```python
import numpy as np
import tensorflow as tf

#A B B * B * B, where * means blank
abb_b_b = np.array([[1, 0, 0], [0, 1, 0], [0, 1, 0], [0, 0, 1], [0, 1, 0], [0, 0, 1], [0, 1, 0]]) # One hot encoding
abb_b_b = abb_b_b[None, :, :].transpose((1,0,2)) # Added batch axis and transpose (it's time major notation)

merge = tf.sparse_tensor_to_dense(tf.nn.ctc_greedy_decoder(abb_b_b, seq_len)[0][0]).eval()
non_merge = tf.sparse_tensor_to_dense(tf.nn.ctc_greedy_decoder(abb_b_b, seq_len, merge_repeated=False)[0][0]).eval()

print(merge)
print(non_merge)
```

you'll get this output

```python
[[0 1 1 1]] # Which means ABBB
[[0 1 1 1 1]] # Which means ABBBB. There is a extra 1 in the docs!
```
So, the docs seems to have a problem, but not the problem that you've point to. If you look at [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/ctc/ctc_decoder.h#L88) line of code, you can see that if `merge_repeated = False`, then the greedy decoder will append to output only the maximum target if it isn't blank. In case of `merge_repeated=True`, the maximum target will only be appended to output if it isn't the blank label and if it isn't equal the previous label. 

Look at the comment [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/ctc_ops.py#L184). As you can see, there is a typo. Instead of * `A B B B B B` if `merge_repeated=False`., it should be * `A B B B B` if `merge_repeated=False`.

So, I recommend you to change your PR to fix this typo! :)

tensorflow/python/ops/ctc_ops.py
```diff
@@ -180,7 +180,7 @@ def ctc_greedy_decoder(inputs, sequence_length, merge_repeated=True):
   only the first of these is emitted.  The sequence `A B B * B * B` (where '*'
   is the blank label) becomes
 
-    * `A B` if `merge_repeated=True`.
+    * `A B B B` if `merge_repeated=True`.
     * `A B B B B B` if `merge_repeated=False`.
```
igormq: Could you change this to * `A B B B B` if `merge_repeated=False`?

igormq (approved):

zheng-xq:

tensorflow/python/ops/ctc_ops.py
```diff
@@ -180,8 +180,8 @@ def ctc_greedy_decoder(inputs, sequence_length, merge_repeated=True):
   only the first of these is emitted.  The sequence `A B B * B * B` (where '*'
   is the blank label) becomes
 
-    * `A B` if `merge_repeated=True`.
-    * `A B B B B B` if `merge_repeated=False`.
+    * `A B B B` if `merge_repeated=True`.
```
ebrevdo: This seems at odds with the unit test [here](https://github.com/petewarden/tensorflow_ios/blob/master/tensorflow/contrib/ctc/ctc_decoder_ops_test.py#L132), no?

ebrevdo: sorry - that was to the wrong repo, but the correct copy of the same file, [same line](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/ctc_decoder_ops_test.py#L133).

Jihadik: No, everything is fine. In those two examples, if you take `argmax` you will obtain:
1. `A * * B` which will transform into `A B`.
2. `B B * B A` which will transform into `B B A`.

Jihadik: In fact it is really confusing since from docs it seems that there is no way to obtain two similar labels in a row like `B B A` (that is one of the reasons why we need `blank`) which is not true for CTC.

Jihadik:

ebrevdo:
You're right, thanks!

ebrevdo:

ebrevdo (approved):

Jihadik:
Are we merging it, @ebrevdo?

ebrevdo:
Let's pause on it.  I need to find out how many people the true fix would break.

ebrevdo (dismissed):
Please hold on a few days.

vrv:
Ping for @ebrevdo -- let me know approximately when you think you'll have found your answer?

Jihadik:
Btw, what is the problem with merging comments? :)

tensorflow/core/ops/ops.pbtxt
```diff
@@ -3536,7 +3536,7 @@ op {
     description: "If True, merge repeated classes in output."
   }
   summary: "Performs greedy decoding on the logits given in inputs."
-  description: "A note about the attribute merge_repeated: if enabled, when\nconsecutive logits\' maximum indices are the same, only the first of\nthese is emitted.  Labeling the blank \'*\', the sequence \"A B B * B B\"\nbecomes \"A B\" if merge_repeated = True and \"A B B B B\" if\nmerge_repeated = False.\n\nRegardless of the value of merge_repeated, if the maximum index of a given\ntime and batch corresponds to the blank, index `(num_classes - 1)`, no new\nelement is emitted."
+  description: "A note about the attribute merge_repeated: if enabled, when\nconsecutive logits\' maximum indices are the same, only the first of\nthese is emitted.  Labeling the blank \'*\', the sequence \"A B B * B B\"\nbecomes \"A B B\" if merge_repeated = True and \"A B B B B\" if\nmerge_repeated = False.\n\nRegardless of the value of merge_repeated, if the maximum index of a given\ntime and batch corresponds to the blank, index `(num_classes - 1)`, no new\nelement is emitted."
```
vrv: This file is autogenerated, you can remove it from the PR and it'll get updated after merging sometime.

tensorflow/g3doc/api_docs/python/functions_and_classes/shard6/tf.nn.ctc_greedy_decoder.md
```diff
@@ -11,8 +11,8 @@ This means that if consecutive logits' maximum indices are the same,
 only the first of these is emitted.  The sequence `A B B * B * B` (where '*'
 is the blank label) becomes
 
-  * `A B` if `merge_repeated=True`.
-  * `A B B B B B` if `merge_repeated=False`.
+  * `A B B B` if `merge_repeated=True`.
```
vrv: same with this one and nn.md

Jihadik:
As I thought :), thanks!

Jihadik:
@vrv, I fixed it!

vrv:
(Thanks, now will wait for @ebrevdo to review)

Jihadik:
@ebrevdo, ping!

ebrevdo:
It's not the documentation that's wrong - it's the code.  Would you like to fix the code instead?

Jihadik:
No problems, I will look at the code a little bit later.

lingz:
Can confirm that I ran into this the other day. I was doing some training with labels that had repeated symbols, but it would output the warning "No Valid Path Found", and would return a cost of infinity.

ebrevdo:
Ling, I'm not sure that is related.

lingz:
Actually yeah, I investigated further and it was just that I did not make the seq_len long enough to allow for the blanks to be between the labels, so thus "No Valid Path Found"

Though I'm confused about your earlier comment @ebrevdo where you said it's not the documentation but the code that is wrong. The unit test linked and @igormq seems to show merge_repeated demonstrating the correct behavior?

ebrevdo:
The unit test is either wrong, or doesn't actually test repeats correctly.

martinwicke:
Ping! What's the next step for this?

Jihadik:
Soon I will try to fix it. I was really busy last few weeks.

saseptim:
I have run into several issues in the CTC decoding.

1) The greedy decoder seems to work right but has a problem with the documentation (still). Documentation online still shows 
A B B * B * B (where '*' is the blank label) becomes
    A B if merge_repeated=True.
    A B B B B B if merge_repeated=False.

Both results are wrong.
This should read:
    A B B B if merge_repeated=True.
    A B B B B  if merge_repeated=False.

2) I had a problem with the beamsearch decoder. Running the code of **igormq** above, but switching in the beamsearch decoder in place of the greedy decoder via:
`merge = tf.sparse_tensor_to_dense(tf.nn.ctc_beam_search_decoder(abb_b_b, seq_len, merge_repeated=False)[0][0]).eval()

non_merge = tf.sparse_tensor_to_dense(tf.nn.ctc_beam_search_decoder(abb_b_b, seq_len, merge_repeated=False)[0][0]).eval()

print(merge)
print(non_merge)
`

the results for both are:
[[0 1 0 1]] or A B A B

which makes no sense.

Additionally, I have also had the problem where the beamsearch decoder did not properly deal with repeated characters and had to switch merge_repeated to False. 

In summary, documentation problems in greedy decoder,  code problems in beamsearch decoder.

ebrevdo:
You have merge_repeated=false in both cases for your beam search code. For
greedy decoder, the documentation isn't wrong, the code is. I'll look into
the beam search decoder in early January.

saseptim:
Regarding the beamsearch, you are right, I copied the code wrong (now it is fixed) but the results are the same - [[0 1 0 1]] - still incorrect.

As for the greedy decoder, I do not understand why you think the code is wrong. The code acts for me as expected. The comments, on the other hand, are certainly wrong, just the fact that the result has 5 B's and the input has 4.

ebrevdo:
For greedy beam search, the code should act like the documentation... Not
the other way around. It's a bug.

On Dec 13, 2016 11:12 PM, "saseptim" <notifications@github.com> wrote:

Regarding the beamsearch, you are right, I copied the code wrong (now it is
fixed) but the results are the same - [[0 1 0 1]] - still incorrect.

As for the greedy decoder, I do not understand why you think the code is
wrong. The code acts for me as expected. The comments, on the other hand,
are certainly wrong, just the fact that the result has 5 B's and the input
has 4.

—
<https://github.com/tensorflow/tensorflow/pull/5194#issuecomment-266981939>,
or mute the thread

Jihadik:
@ebrevdo, I really disagree with you. If `merge_repeated=True` then it is exactly [Graves' CTC](ftp://ftp.idsia.ch/pub/juergen/icml2006.pdf) and in case `A B B * B * B` firstly it merges repeated labels and make `A B * B * B` then it removes all blanks and our sequence become `A B B B`, not `A B` as in docs now. If it firstly remove all blanks and the merge labels then we will never have two consecutive labels in a row which is not good and was not inteded by Graves.

If I'm wrong, please prove it.

saseptim:
I agree, there is a problem with the documentation of the greedy decoder and the code of the beamsearch decoder.

drpngx:
@Jihadik I really appreciate your willingness to contribute. Basically, the way it works is that google staff is maintaining the code and needs to make the final decision on changes we're comfortable with. We can't really merge @ebrevdo has a different opinion.

ebrevdo:
I'll revisit tomorrow.

ebrevdo:
@Jihadik, you are right.  lgtm.

drpngx:
Jenkins, test this please.

rmlarsen (approved):

PR #5196, opened 2016-10-25
Changed context creation to use primary context

ptrendx:
Using cuCtxCreate/Destroy does not give any benefits over using primary context and is not advised by the CUDA documentation: http://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#multiple-contexts.
Moreover, it breaks interaction with some runtime API calls, like cudaPointerGetAttributes and cudaEnablePeerAccess (see section Context Interoperability in: http://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER.html#group__CUDART__DRIVER).
This PR changes how the context are obtained and released from cuCtxCreate/Destroy to cuDevicePrimaryCtxRetain/Release

### DIFF
--- a/tensorflow/stream_executor/cuda/cuda_driver.cc
+++ b/tensorflow/stream_executor/cuda/cuda_driver.cc
@@ -75,6 +75,9 @@ namespace dynload {
   const char *DynLoadShim__##__name::kName = #__name;
 
 PERFTOOLS_GPUTOOLS_LIBCUDA_WRAP(cuCtxCreate_v2);
+PERFTOOLS_GPUTOOLS_LIBCUDA_WRAP(cuDevicePrimaryCtxRetain);
+PERFTOOLS_GPUTOOLS_LIBCUDA_WRAP(cuDevicePrimaryCtxRelease);
+PERFTOOLS_GPUTOOLS_LIBCUDA_WRAP(cuDevicePrimaryCtxSetFlags);
 PERFTOOLS_GPUTOOLS_LIBCUDA_WRAP(cuCtxDestroy);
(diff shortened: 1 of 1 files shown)


ptrendx:
I signed it!

zheng-xq:
@leary-google, @henline, could you review this change?

tensorflow/stream_executor/cuda/cuda_driver.cc
```diff
@@ -584,7 +587,8 @@ bool DeviceOptionsToContextFlags(DeviceOptions device_options, int *flags) {
     // TODO(leary) Need to see if NVIDIA can expunge the leakiness in their
     // context creation: see http://b/13248943
 
-    res = dynload::cuCtxCreate_v2(&new_context, flags, device);
+    res = dynload::cuDevicePrimaryCtxSetFlags(device, flags);
```
henline: This looks great, but I think these primary context functions are only available in CUDA 7.5 and newer, and unfortunately one of the major TensorFlow users is currently stuck on CUDA 7.0. Could we check the libcuda version here and use the old calls for 7.0 and earlier?

ptrendx: These functions were introduced in 7.0. I will add checks for earlier versions of CUDA.

henline (approved):
Thanks, this is a great change!

zheng-xq:

gunan:
Wait, this PR is created against the r0.11 release branch.
It should be submitted to master. I think there has been a mistake.

We are not accepting anything but bugfixes into r0.11 branch, therefore I will close this PR.

Issue #5169, opened 2016-10-24
Windows GPU random initialization not working


ToraxXx:
Using `tf.truncated_normal` (and possibly the other random functions too) on my GPU (GTX 1060) with the new Windows build produces only zeros whereas running it on my CPU works as expected.

Example code:
```Python
with tf.device('/cpu:0'):
    with tf.Session() as sess:
        # Something random, as expected
        print("CPU:", sess.run(tf.truncated_normal([5])))

with tf.device('/gpu:0'):
    with tf.Session() as sess:
        # All zeros
        print("GPU:", sess.run(tf.truncated_normal([5])))
```

I used the same setup as [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/README.md) with CMake 3.6 and rolled back to https://github.com/tensorflow/tensorflow/commit/41ba1e0e6fbf443f1a972ff0130ba6741b9b7a50 to fix the issue with eigen not compiling.

// Also it seems the result of any calculation involving a tf.Variable becomes 0, while just running / printing the variable itself shows its correct value (CPU works fine though).

drpngx:
@mrry @guschmue for any insight

guschmue:
let me take look. I still have a bunch of gpu fixes - working on a pr for those but not sure of this one is part of it.

guschmue:
Works on my box but I have some idea why its not working for you: on windows there might be some issue that it doesn't handle correctly if the exact cuda version is not compiled in. By default we compile with cuda 3.5+5.2 but a 1060 is on 6.1. I belief on linux it would downgrade you to 5.2 but on windows it does something undefined. Let me debug this a little - should not be to hard to find. 
If you need a solution right away: you can edit the CMakefile and add 6.1 to the list of cuda capabilities.

ToraxXx:
Alright I'll try it tomorrow and let you know if it worked, thanks for your help and great work!

guschmue:
If you change your CMakefile: search for all occurrences of 5.2 and add a 6.1. I should have a better solution by tomorrow.

ToraxXx:
I added 6.1 everywhere and it worked. I also tried neural style which worked too.

drpngx:
OK, closing this for now. Thanks much @guschmue !

guschmue:
I found the reason why on windows we are not falling back to 5.2 if we find a 6.1 gpu. Going to send a PR so in the future it will behave just like linux.

PR #5207, opened 2016-10-26
fix for issue #5169

guschmue:
in nvcc -gencode we need to specify both sm_ and compute_, else we can not
fallback to a lower compute capability.


mrry:

mrry (approved):
Looks good to me!

PR #5218, opened 2016-10-26
Fix: For mac non-docker GPU builds fix handling of test concurrency.

gunan:
@mrry This should fix mac-gpu tests.
@yifeif, let me test this. After that we should also cherrypick this into r0.11.

@caisq I keep messing up this change, please feel free to point out any problems or concerns you see.

PR #5227, opened 2016-10-27
Fix: For mac non-docker GPU builds fix handling of test concurrency. …

gunan:
…(#5218)

### DIFF
--- a/tensorflow/tools/ci_build/builds/test_installation.sh
+++ b/tensorflow/tools/ci_build/builds/test_installation.sh
@@ -47,9 +47,6 @@
 # TF_BUILD_BAZEL_CLEAN, if set to any non-empty and non-0 value, directs the
 # script to perform bazel clean prior to main build and test steps.
 #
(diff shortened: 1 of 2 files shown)


tensorflow/tools/ci_build/builds/test_installation.sh
```diff
@@ -410,7 +407,7 @@ SKIP_COUNTER=0
 FAILED_TESTS=""
 FAILED_TEST_LOGS=""
 
-if [[ "${TF_BUILD_USE_GPU}" == "1" ]]; then
+if [[ "${IS_GPU}" == "1" ]]; then
```
martinwicke: Where does IS_GPU come from? It's not documented any more with the other variables.

martinwicke (approved):

tensorflow/tools/ci_build/builds/test_installation.sh
martinwicke: Never mind. I see the flag.

gunan:
All tests except windows passing.
Merging.

PR #5229, opened 2016-10-27
reimplement viterbi decoding in tensorflow way

oohoops:
The original viterbi decoding in `tensorflow.contrib.crf` is implemented in numpy way and can only be used in testing. This update reimplement viterbi decoding in tensorflow style, can provide more flexibility during model building.


jungle-cat:
I signed it!

jungle-cat:
this implementation contain some import issues to be resolved.

PR #5238, opened 2016-10-27
gemmlopw has been added to external dependencies.

kamcpp:
Hi guys,
I was building tf using CMake in ubuntu. I found out that an external dependency (gemmlowp) is missing from the CMakeLists.txt file. The building failed while compiling tf_core_kernels files. I just added the dependency and everything went OK.


kamcpp:
I signed it!

kamcpp:
I signed it!

PR #5267, opened 2016-10-28
Initial support for OpenCL

benoitsteiner:


benoitsteiner:
I signed it !

martinwicke:
@lukeiwanski Can you comment here to let CLAbot know you're fine with contributing these commits?

martinwicke:
@jart This change has a lot of bazel (mostly parallel to the GPU case), which I thought you'd be most qualified to review.

tensorflow/core/common_runtime/device_factory.cc
```diff
@@ -97,11 +97,20 @@ Status DeviceFactory::AddDevices(const SessionOptions& options,
         gpu_factory->CreateDevices(options, name_prefix, devices));
   }
 
+  // Then SYCL.
```
vrv: I don't think we need this code, the loop below should handle this.

tensorflow/core/common_runtime/rendezvous_mgr.cc
```diff
@@ -65,10 +65,10 @@ void IntraProcessRendezvous::SameWorkerRecvDone(
     StatusCallback done) {
   // Do a quick copy (sharing the underlying buffer) if both tensors
   // are on host memory.
-  const bool src_host =
-      (send_args.alloc_attrs.on_host() || parsed.src.type == "CPU");
-  const bool dst_host =
-      (recv_args.alloc_attrs.on_host() || parsed.dst.type == "CPU");
+  const bool src_host = (send_args.alloc_attrs.on_host() ||
+                         parsed.src.type == "CPU" || parsed.src.type == "SYCL");
```
vrv: Let's see if we can change the alloc_attrs to set on_host() for SYCL, instead of adding the special case here.

tensorflow/core/framework/types.h
```diff
@@ -68,8 +68,9 @@ class DeviceType {
 std::ostream& operator<<(std::ostream& os, const DeviceType& d);
 
 // Convenient constants that can be passed to a DeviceType constructor
-extern const char* const DEVICE_CPU;  // "CPU"
-extern const char* const DEVICE_GPU;  // "GPU"
+extern const char* const DEVICE_CPU;   // "CPU"
+extern const char* const DEVICE_GPU;   // "GPU"
+extern const char* const DEVICE_SYCL;  // "SYCL"
```
vrv: Can we define DEVICE_SYCL in a sycl-specific header? 

It would be nice to demonstrate by example that you don't need to add code to core TensorFlow for new devices.

tensorflow/core/util/device_name_utils.cc
```diff
@@ -162,6 +162,16 @@ bool DeviceNameUtils::ParseFullName(StringPiece fullname, ParsedName* p) {
       }
       progress = true;
     }
+    if (str_util::ConsumePrefix(&fullname, "/sycl:") ||
```
vrv: You can remove this: device assignments can just use "/device:SYCL:0" instead and you don't need a special-case here

vrv (commented):
Hi Benoit, just a few suggestions -- we should be able to avoid adding SYCL-specific code to the core and get everything to work, so some suggestions for cleaning this up.

lukeiwanski:
I signed it !

tensorflow/core/util/device_name_utils.cc
lukeiwanski: sorry, I must be missing something. Where should I use /task:0/device:sycl:0 ?

vrv: In whatever program you are using, so for example:

```
with tf.device("/job:localhost/replica:0/task:0/device:sycl:0"):
  tf.add(...)
```

tensorflow/core/common_runtime/rendezvous_mgr.cc
vrv: This also seems like a bug, btw.  This is saying that device memory and host memory are the same for SYCL, when it comes to memcopying.  E.g., that when you are copying from say, /CPU:0 to /device:SYCL:0, the tensor being pointed to is in CPU memory.  Is that intended?

tensorflow/core/util/device_name_utils.cc
lukeiwanski: #4359 is very useful, thanks for the link! :)
Does it mean we need to copy and pasting all the existing TF tests and duplicated them with SYCL ones?

We do plan on supporting all the existing tests in the future.

tensorflow/core/framework/types.h
lukeiwanski: Do you have any preferences on where DEVICE_SYCL should go? 

types.h seemed like a good place since it is included by all Ops. 
That makes Op registration easy.

tensorflow/core/common_runtime/rendezvous_mgr.cc
lukeiwanski: This is just a temporary work around.

tensorflow/core/util/device_name_utils.cc
vrv: I think in this particular case, since we're supporting SYCL in our official repo (optionally), having the kernel registrations / tests in the same files makes some sense.

tensorflow/core/framework/types.h
vrv: I guess since we're registering all the kernels in the same files as CPU and GPU, this is fine.  Otherwise I'd have suggested defining something like a sycl/defs.h header where common SYCL-specific defines could go.

gunan:
CLA bot still reporting failure with "CLA signed, but unable to verify author consent".
@martinwicke how do we convince clabot?

vrv:
Luke, have you signed the CLA?  Can you:

- Sign it under the email address you used to author the commits
- Make sure that email is added to your github acct?

lukeiwanski:
I have 2 emails connected with my github .. both are added to google group for my company. Not quite sure what is happening with CLA.

vrv:
@willnorris for additional help with corp CLA stuff.

(Will, let me know if there's anything we can do on your behalf to validate -- the corp agreement side of things is something I have no visibility into)

willnorris:
@vrv: <http://go/cla#github-consent>.  googlebot is never going to mark this PR as okay.  And that's fine, it's not supposed to.  It's up to you as the merger to confirm that commit authors have consented.

benoitsteiner:
I've updated the pull request to pick up the latest TensorFlow changes: this should take care of the MacOS failures.

tensorflow/core/common_runtime/sycl/sycl_device.h
```diff
@@ -0,0 +1,59 @@
...
+  Allocator* GetAllocator(AllocatorAttributes attr) override;
+  Status MakeTensorFromProto(const TensorProto& tensor_proto,
+                             const AllocatorAttributes alloc_attrs,
+                             Tensor* tensor) override;
+
+  Status Sync() override { return Status::OK(); }
+  static string GetShortDeviceDescription(/*int device_id,
+                                          const DeviceDescription& desc*/) {
+    return strings::StrCat("device: 0, name SYCL, pci bus id: 0");
+    // return strings::StrCat("device: ", device_id, ", name: ", desc.name(),
```
vrv: remove

tensorflow/core/common_runtime/sycl/sycl_device_factory.cc
```diff
@@ -0,0 +1,44 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#if TENSORFLOW_USE_SYCL
+
+#include "sycl_device.h"
```
vrv: full include paths

tensorflow/core/kernels/constant_op.cc
```diff
@@ -50,6 +50,7 @@ void ConstantOp::Compute(OpKernelContext* ctx) { ctx->set_output(0, tensor_); }
 ConstantOp::~ConstantOp() {}
 
 REGISTER_KERNEL_BUILDER(Name("Const").Device(DEVICE_CPU), ConstantOp);
+REGISTER_KERNEL_BUILDER(Name("Const").Device(DEVICE_SYCL), ConstantOp);
```
vrv: Should these be protected by "TENSORFLOW_USE_SYCL" ?  Otherwise we're registering constants for devices that don't exist.

tensorflow/core/kernels/cwise_ops_sycl_common.h
```diff
@@ -0,0 +1,145 @@
...
+  void operator()(const SYCLDevice& d, typename Functor::tout_type out,
+                  typename Functor::tin_type in0,
+                  typename Functor::tin_type in1, bool* error) {
+    Assign(d, out, in0.binaryExpr(in1, typename Functor::func()));
+  }
+
+  void Left(const SYCLDevice& d, typename Functor::tout_type out,
+            typename Functor::tscalar_type scalar,
+            typename Functor::tin_type in, bool* error) {
+    // typedef typename Functor::out_type Tout;
```
vrv: remove all these comments?

tensorflow/core/kernels/cwise_ops_sycl_common.h
```diff
@@ -0,0 +1,145 @@
...
+
+  void BCast(const SYCLDevice& d,
+             typename TTypes<typename Functor::out_type, NDIMS>::Tensor out,
+             typename TTypes<typename Functor::in_type, NDIMS>::ConstTensor in0,
+             typename Eigen::array<Eigen::DenseIndex, NDIMS> bcast0,
+             typename TTypes<typename Functor::in_type, NDIMS>::ConstTensor in1,
+             typename Eigen::array<Eigen::DenseIndex, NDIMS> bcast1,
+             bool* error) {
+    LOG(FATAL) << "BinaryFunctor::BCast NOT IMPLEMENTED ";
+    //    printf("BinaryFunctor::BCast NOT IMPLEMENTED ! \n");
```
vrv: same here

tensorflow/core/common_runtime/device_factory.cc
vrv: This still isn't necessary, I think, and we should remove it.

tensorflow/core/common_runtime/rendezvous_mgr.cc
vrv: Can you elaborate what it works around, and where we intend to be in the future?

tensorflow/core/util/device_name_utils.cc
vrv: We still need to remove this.

third_party/eigen3/BUILD
```diff
@@ -23,5 +23,7 @@ cc_library(
         "unsupported/Eigen/CXX11/FixedPoint",
     ],
     visibility = ["//visibility:public"],
-    deps = ["@eigen_archive//:eigen"],
+    deps = ["@eigen_archive//:eigen",
```
vrv: can you run buildifer on this?  It's going to fail our sanity check otherwise.

vrv (changes requested):
Cool, just a few more comments on the C++ side of things.  I don't know much about the bazel stuff, so maybe @jart could take a look

tensorflow/core/common_runtime/rendezvous_mgr.cc
lukeiwanski: My understanding was that the SYCL part of Eigen will take care of data management and TensorFlow will deal only with the host memory.
Unfortunately, I was proven wrong  by very simple python test that I created. 
To this end I am studying the codebase in order to implement missing functionality in the SYCL TensorFlow code. Which seems to be pointing to device specific allocator.
And yes you are right, SYCL has different address space for the data. SYCL manages data copies and dependencies internally via buffers and accessors and will maintain a CPU copy of the data when the buffer was initialised with a pointer.

> the tensor being used as input to the sycl device must remain in CPU memory.
Could you clarify exactly what you mean by that? Does that mean CUDA is deallocating memory once pushed to the Device memory?

vrv: Yeah, you probably need to have a device-specific allocator.

The change here is basically saying that from the perspective of the framework, whenever you try to copy to a SYCL device, the destination should be treated as "host memory" (e.g., on the CPU).  This probably isn't what you want: you likely want the data to be copied into "device memory" which will have been allocated using a custom device allocator for SYCL.

In the CUDA case, when you have a send/recv pair between a CPU and GPU subgraph, you would want to transfer from CPU memory into GPU memory allocated via cuMalloc.  Our GPU allocator implementation for CUDA does this allocation, so you can see that our existing GPUDevice uses a GPUBFCAllocator implementation to allocate device memory via cuMalloc.

There are some cases where a kernel is written such that some of its inputs are not really needed in "device memory" -- they are more like 'control' bits (e.g., the index arguments to reshape).  Those are marked as "HostMemory" in their kernel registrations.  So this piece of code checks whether the source and destination are in 'host/cpu' memory.  If so, no copy needs to be made because the memory address spaces are the same.  Does that make sense?

benoitsteiner:
Jenkins, test this please.

tensorflow/core/kernels/identity_op.cc
```diff
@@ -24,6 +24,7 @@ limitations under the License.
 namespace tensorflow {
 
 REGISTER_KERNEL_BUILDER(Name("Identity").Device(DEVICE_CPU), IdentityOp);
+REGISTER_KERNEL_BUILDER(Name("Identity").Device(DEVICE_SYCL), IdentityOp);
```
vrv: protect this with #if TENSORFLOW_USE_SYCL ? (same with some of the other files)

benoitsteiner: Done. I'll update the PR in the next few minutes.

vrv (approved):

vrv:
Approving the C++ changes.  Don't know about build changes.

jart:
I'm not sure if I'm qualified to review this. But I would like to have a better understanding of whether or not we considered alternatives to adding more stuff to the configure script. I'd also like to know how we plan to regression test this functionality.

gunan:
At the moment, we are running "bazel test tensorflow/..." with config=sycl.
We have a few machines that can have opencl support, we will utilize those to run tests continuously.

rickyhan:
Thank you for supporting OpenCL. How can I test it out? I have a triple AMD R9 290X setup.

gunan:
This is just the bare minimum changes we need for opencl.
Full support is still not there. Once we have full support, we will have documentation , and a few toy examples you can use.

PR #5279, opened 2016-10-29
- fix for memory corruption: on windows use _aligned_free() for memor…

guschmue:
…y allocated with aligned_alloc()
- enabled sparse_matmul and immutable_constant_op
- don't print LD_LIBRARY_PATH on error for windows (its null and crashes)
  with this python kernel tests pass (well, some error left related to stl float Inf)

### DIFF
--- a/tensorflow/contrib/cmake/tf_core_kernels.cmake
+++ b/tensorflow/contrib/cmake/tf_core_kernels.cmake
@@ -85,33 +85,34 @@ list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_exclude_srcs})
 if(WIN32)
   file(GLOB_RECURSE tf_core_kernels_windows_exclude_srcs
       # not working on windows yet
-      "${tensorflow_source_dir}/tensorflow/core/kernels/depthwise_conv_op.cc"  # Cannot find symbol: tensorflow::LaunchConv2DOp<struct Eigen::ThreadPoolDevice, double>::launch(...).
       "${tensorflow_source_dir}/tensorflow/core/kernels/fact_op.cc"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/immutable_constant_op.cc"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/immutable_constant_op.h"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/meta_support.*"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/sparse_matmul_op.cc"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/sparse_matmul_op.h"
       "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.h"
       "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.cc"
+      "${tensorflow_source_dir}/tensorflow/core/kernels/meta_support.*"
       "${tensorflow_source_dir}/tensorflow/core/kernels/svd*.cc"
       "${tensorflow_source_dir}/tensorflow/core/kernels/avgpooling_op.*"
   )
   list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_windows_exclude_srcs})
 endif(WIN32)
 
-file(GLOB_RECURSE tf_core_gpu_kernels_srcs
-   "${tensorflow_source_dir}/tensorflow/core/kernels/*.cu.cc"
-   "${tensorflow_source_dir}/tensorflow/contrib/rnn/kernels/*.cu.cc"
-)
+if(WIN32 AND tensorflow_ENABLE_GPU)
+  file(GLOB_RECURSE tf_core_gpu_kernels_srcs
+     "${tensorflow_source_dir}/tensorflow/core/kernels/*.cu.cc"
+     "${tensorflow_source_dir}/tensorflow/contrib/rnn/kernels/*.cu.cc"
+  )
 
-if(WIN32)
   file(GLOB_RECURSE tf_core_gpu_kernels_exclude_srcs
       # not working on windows yet
       "${tensorflow_source_dir}/tensorflow/core/kernels/avgpooling_op_gpu.cu.cc"
   )
   list(REMOVE_ITEM tf_core_gpu_kernels_srcs ${tf_core_gpu_kernels_exclude_srcs})
-endif(WIN32)
+
+  # if we build for gpu include those  
+  list(APPEND tf_core_kernels_srcs 
+     "${tensorflow_source_dir}/tensorflow/core/kernels/debug_ops.h"
+     "${tensorflow_source_dir}/tensorflow/core/kernels/debug_ops.cc"
+  )
+endif(WIN32 AND tensorflow_ENABLE_GPU)
 
 add_library(tf_core_kernels OBJECT ${tf_core_kernels_srcs})
 add_dependencies(tf_core_kernels tf_core_cpu)
--- a/tensorflow/core/common_runtime/gpu/pool_allocator.h
+++ b/tensorflow/core/common_runtime/gpu/pool_allocator.h
@@ -173,7 +173,11 @@ class BasicCPUAllocator : public SubAllocator {
   void* Alloc(size_t alignment, size_t num_bytes) override {
     return port::aligned_malloc(num_bytes, alignment);
   }
+#if defined(PLATFORM_WINDOWS)
+  void Free(void* ptr, size_t num_bytes)  override { _aligned_free(ptr); }
+#else
   void Free(void* ptr, size_t num_bytes) override { free(ptr); }
+#endif
 };
 
 // Allocator for pinned CPU RAM that is made known to CUDA for the
--- a/tensorflow/core/kernels/debug_ops.h
+++ b/tensorflow/core/kernels/debug_ops.h
@@ -16,9 +16,12 @@ limitations under the License.
 #ifndef TENSORFLOW_KERNELS_DEBUG_OP_H_
 #define TENSORFLOW_KERNELS_DEBUG_OP_H_
 
+#include <type_traits>
+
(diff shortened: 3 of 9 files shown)


drpngx:
Jenkins, test this please.

tensorflow/contrib/cmake/tf_core_kernels.cmake
```diff
@@ -85,33 +85,34 @@ list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_exclude_srcs})
...
   file(GLOB_RECURSE tf_core_gpu_kernels_exclude_srcs
       # not working on windows yet
       "${tensorflow_source_dir}/tensorflow/core/kernels/avgpooling_op_gpu.cu.cc"
   )
   list(REMOVE_ITEM tf_core_gpu_kernels_srcs ${tf_core_gpu_kernels_exclude_srcs})
-endif(WIN32)
+
+  # if we build for gpu include those  
+  list(APPEND tf_core_kernels_srcs 
+     "${tensorflow_source_dir}/tensorflow/core/kernels/debug_ops.h"
```
drpngx: So I guess the windows behavior will be different from the linux one? One has debug ops, only in GPU, and the other one doesn't? We probably want the debug ops to be included (or not) not matter orthogonally to whether we're win32 or gpu.

tensorflow/core/common_runtime/gpu/pool_allocator.h
```diff
@@ -173,7 +173,11 @@ class BasicCPUAllocator : public SubAllocator {
   void* Alloc(size_t alignment, size_t num_bytes) override {
     return port::aligned_malloc(num_bytes, alignment);
   }
+#if defined(PLATFORM_WINDOWS)
```
drpngx: there should be a port::aligned_free which is valid for all cases. ASAN or valgrind should be able to flag that.

tensorflow/core/kernels/debug_ops.h
```diff
@@ -127,7 +130,16 @@ class DebugNanCountOp : public OpKernel {
       const T* input_flat = input.template flat<T>().data();
 
       for (int64 i = 0; i < input_shape.num_elements(); ++i) {
-        if (Eigen::numext::isnan(input_flat[i])) {
+#ifdef _MSC_VER
+        // VC complains about ambiguous resolution when isnan is called
+        // with integral values, so we explicitly convert an integral value to
+        // double before calling isnan.
+        typedef std::conditional<std::is_integral<T>::value, double, T>::type R;
+        R value = static_cast<R>(input_flat[i]);
```
drpngx: ```const R value```

Would just `static_cast<double>(input_flat[i])` work with all types?

tensorflow/core/kernels/debug_ops.h
```diff
@@ -127,7 +130,16 @@ class DebugNanCountOp : public OpKernel {
...
+#else
+        T& value = input_flat[i];
```
drpngx: const

tensorflow/core/kernels/sparse_matmul_op.h
```diff
@@ -18,6 +18,22 @@ limitations under the License.
 
 #include "third_party/eigen3/Eigen/Core"
 #include "tensorflow/core/platform/types.h"
+#ifdef _MSC_VER
```
drpngx: is it possible to find a better home for this, somewhere in platform? Maybe platform/windows/intrinsics_port.h?

drpngx (changes requested):
Nice!

drpngx:
Woops, it doesn't like `std::bind` and other things. Please take a look at the errors:

https://ci.tensorflow.org/job/tensorflow-pull-requests-multijob/2416/

Also, if you're comfortable with git, you could split the commits.

guschmue:
ah, should have tried on linux first. Working on it.

tensorflow/contrib/cmake/tf_core_kernels.cmake
```diff
@@ -85,33 +85,34 @@ list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_exclude_srcs})
...
-      "${tensorflow_source_dir}/tensorflow/core/kernels/depthwise_conv_op.cc"  # Cannot find symbol: tensorflow::LaunchConv2DOp<struct Eigen::ThreadPoolDevice, double>::launch(...).
       "${tensorflow_source_dir}/tensorflow/core/kernels/fact_op.cc"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/immutable_constant_op.cc"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/immutable_constant_op.h"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/meta_support.*"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/sparse_matmul_op.cc"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/sparse_matmul_op.h"
       "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.h"
       "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.cc"
+      "${tensorflow_source_dir}/tensorflow/core/kernels/meta_support.*"
```
guschmue: there was a mege conflict on meta_support

tensorflow/contrib/cmake/tf_core_kernels.cmake
```diff
@@ -85,33 +85,34 @@ list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_exclude_srcs})
...
       "${tensorflow_source_dir}/tensorflow/core/kernels/avgpooling_op.*"
   )
   list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_windows_exclude_srcs})
 endif(WIN32)
 
-file(GLOB_RECURSE tf_core_gpu_kernels_srcs
-   "${tensorflow_source_dir}/tensorflow/core/kernels/*.cu.cc"
-   "${tensorflow_source_dir}/tensorflow/contrib/rnn/kernels/*.cu.cc"
-)
+if(WIN32 AND tensorflow_ENABLE_GPU)
```
guschmue: it was collecting the files in tf_core_gpu_kernels_srcs but for cpu no target use this variable. That's maybe ok but its risky breaking somebody else using cmake so I figured its better to ifdef it.

tensorflow/contrib/cmake/tf_core_kernels.cmake
```diff
@@ -85,33 +85,34 @@ list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_exclude_srcs})
...
 
-if(WIN32)
   file(GLOB_RECURSE tf_core_gpu_kernels_exclude_srcs
       # not working on windows yet
       "${tensorflow_source_dir}/tensorflow/core/kernels/avgpooling_op_gpu.cu.cc"
   )
   list(REMOVE_ITEM tf_core_gpu_kernels_srcs ${tf_core_gpu_kernels_exclude_srcs})
-endif(WIN32)
+
+  # if we build for gpu include those
```
guschmue: those don't compile yet, some ambiguous operator error when compiling with msvc. I have a fix but I'm looking for a better fix.

tensorflow/core/common_runtime/gpu/pool_allocator.h
guschmue: yes, there is  - let me switch to this.

tensorflow/core/kernels/debug_ops.h
guschmue: let me look for  a better way

tensorflow/core/kernels/debug_ops.h
guschmue: ok

tensorflow/core/kernels/sparse_matmul_op.h
guschmue: fair.

tensorflow/core/platform/test.cc
```diff
@@ -13,6 +13,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
+#ifdef PLATFORM_WINDOWS
+#include <Windows.h>
```
guschmue: ok

guschmue:
looking over this, you are right: let me split this into a couple of commits. That also helps to get this aligned_free fix in.

guschmue:
closing this one and file some more targeted pr's instead. Going to reference this pr so we don't lose the review feedback.

drpngx:
Please merge with #5207, which will probably create some conflicts on the `meta_support.*`

guschmue:
I closed this pr and split it - too many things in it. I think I'll split it in 4 prs:
1. the memory corruption (addressed your review comments and send a new pr  #5288)
2. enable sparse mat ops (nearly done)
3. enable immutable ops
4. enable debug_op 
I'll make sure sure they are on top of the meta_support.* changes 
Hope that works.

drpngx:
nice, thanks!

drpngx:
BTW, for the `std::bind` thing, we're seeing an error in #5286, possibly the easiest fix would be to use `SparseSlice<TL>*` instead of `auto*`.

drpngx:
Oh, BTW, flipping through some code today, it occurred to me that we should probably `#define WIN32_LEAN_AND_MEAN 1` before including windows.h headers. I usually don't like to have them in header files, only in cpp files, because there a ton of symbols in there.

guschmue:
Ah, lean and mean we have: gets added in CMakeLists.txt to the compiler defines for windows.
Going to try SparseSlice<TL>* shortly. Would love to get rid of the std::bind() without a ifdef.

guschmue:
turns out I did not need the std::bind(). We had made this change before nvidia support vs2015.update3 (we used vs2015.update1 until the official cuda 8.0 came out). Now that the std:bind() made trouble for gcc I started over on sparse_maxmul_op.cc and its fine without std:bind(): builds on linux and windows.

drpngx:
Ah, OK, great to hear! Thanks for making these changes!

PR #5292, opened 2016-10-31
reduce_logsumexp fix for reduction_indices.  Fixes issue #5291

### LINKED ISSUE #5291: [Bug] reduce_logsumexp examples do not work in 0.11.0rc1
Version: 0.11.0rc1

These examples do not work:

```
# 'x' is [[0, 0, 0]]
#         [0, 0, 0]]
tf.reduce_logsumexp(x, 0) ==> [log(2), log(2), log(2)]
tf.reduce_logsumexp(x, 1) ==> [log(3), log(3)]
```

Because:
`tf.squeeze(t, squeeze_dims)` does not take integer for `squeeze_dims`

hoangmit:
Fix #5291

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -1389,7 +1389,7 @@ def reduce_logsumexp(input_tensor, reduction_indices=None, keep_dims=False,
         reduction_indices,
         keep_dims=True)) + my_max
     if not keep_dims:
-      reduction_indices = _ReductionDims(input_tensor, reduction_indices)
+      reduction_indices = [reduction_indices] if isinstance(reduction_indices, int) else reduction_indices
       result = array_ops.squeeze(result, reduction_indices)
     return result
 
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -1389,6 +1389,7 @@ def reduce_logsumexp(input_tensor, reduction_indices=None, keep_dims=False,
         reduction_indices,
         keep_dims=True)) + my_max
     if not keep_dims:
+      reduction_indices = [reduction_indices] if isinstance(reduction_indices, int) else reduction_indices
       result = array_ops.squeeze(result, reduction_indices)
     return result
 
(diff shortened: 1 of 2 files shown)


hoangmit:
I signed it!

tensorflow/python/ops/math_ops.py
```diff
@@ -1389,6 +1389,7 @@ def reduce_logsumexp(input_tensor, reduction_indices=None, keep_dims=False,
         reduction_indices,
         keep_dims=True)) + my_max
     if not keep_dims:
+      reduction_indices = [reduction_indices] if isinstance(reduction_indices, int) else reduction_indices
```
vrv: Change this to:

```
reduction_indices = _ReductionDims(input_tensor, reduction_indices)
```

That's what we do for the other reduction functions in this file.

tensorflow/python/ops/math_ops.py
```diff
@@ -1389,6 +1389,7 @@ def reduce_logsumexp(input_tensor, reduction_indices=None, keep_dims=False,
         reduction_indices,
         keep_dims=True)) + my_max
     if not keep_dims:
+      reduction_indices = _ReductionDims(input_tensor, reduction_indices)
```
vrv: Oh, is it possible to have a test for triggering this case?

vrv: (See where we currently test reduce_logsumexp and add another case for this).

tensorflow/python/ops/math_ops_test.py
```diff
@@ -68,6 +68,17 @@ def testReductionIndices(self):
         y_tf_np = y_tf.eval()
         self.assertAllClose(y_tf_np, y_np)
 
+
```
vrv: only one line space between tests

vrv:

Nm, I did it myself.

vrv:
Well that's unfortunate, it looks like the argument to squeeze has to be an array, not a Tensor, which means we can't use _ReductionDims().

hoangmit:
Use earlier change instead

vrv:

Issue #4701, opened 2016-10-01
Unable to run "bazel build tensorflow/examples/image_retraining:retrain"


atwj:
Hi! I was going through the re-trainer tutorial but i keep encountering the following error.

<code>Executing genrule //tensorflow/core:version_info_gen failed: bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
error: Could not expand include path '~/.gitcinclude'
fatal: bad config file line 49 in /usr/local/git/etc/gitconfig
Traceback (most recent call last):
  File "tensorflow/tools/git/gen_git_source.py", line 254, in <module>
    generate(args.generate)
  File "tensorflow/tools/git/gen_git_source.py", line 207, in generate
    write_version_info(dest_file, git_version)
  File "tensorflow/tools/git/gen_git_source.py", line 160, in write_version_info
    if b"\"" in git_version or b"\" in git_version:
TypeError: 'in <string>' requires string as left operand, not bytes
Target //tensorflow/examples/image_retraining:retrain failed to build
Use --verbose_failures to see the command lines of failed build steps. </code>

Am i doing something incorrectly?

olivermg:
This happens when using Python 3, as that code in gen_git_source.py ('b"\"" in ...' etc.) works only for Python 2.

As a workaround, I've patched my gen_git_source.py locally by removing these 'b' characters (thus passing a string into the "in" operation instead of a byte). However, this is probably just a workaround, as this change may make it incompatible with Python 2 and only work with Python 3.

ceocookie:
Hi @olivermg Could you advise how do you run tensorflow using Python 2? After installing with tensorflow using conda (anaconda2-4.1.1) , I proceed to clone the tensorflow repo, run /.configure before attempting to execute "bazel build tensorflow/examples/image_retraining:retrain". However I still get the following issues:

<code>
ERROR: /Users/AMOS/Documents/PROJECTS/iot-tensorflow-py2/tensorflow/tensorflow/core/BUILD:1030:1: Executing genrule //tensorflow/core:version_info_gen failed: bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
error: Could not expand include path '~/.gitcinclude'
fatal: bad config file line 49 in /usr/local/git/etc/gitconfig
[...]

PR #5318, opened 2016-11-01
Consistently use bytes for git version tracking for Python3 compatibi…

nsuke:
…lity

I'm getting the same TypeError as in #4701 on current master.
This patch fixes it by making sure that `write_version_info` always receives `bytes` for `git_version` param.

### DIFF
--- a/tensorflow/tools/git/gen_git_source.py
+++ b/tensorflow/tools/git/gen_git_source.py
@@ -147,8 +147,10 @@ def get_git_version(git_base_path):
   """
   unknown_label = b"unknown"
   try:
-    val = subprocess.check_output(["git", str("--git-dir="+git_base_path+"/.git"), str("--work-tree="+git_base_path), "describe",
-                                   "--long", "--dirty", "--tags"]).strip()
+    val = bytes(subprocess.check_output([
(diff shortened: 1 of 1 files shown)


zafartahirov:
I would personally make a minor edit using

```python
import builtins
val = builtins.bytes(subprocess.check_output...
```
because the word `bytes` might get overridden by some custom defined method. [Source](https://docs.python.org/3/library/builtins.html)

vrv:
Is builtins available in python2? I  couldn't find the docs in the 2.7 version.  https://docs.python.org/2/library/functions.html

zafartahirov:
Oh yeah -- I stand corrected, my bad! Python 2.x doesn't have `builtins`.

aselle (approved):
LGTM.
One question is whether it will work on Windows @mrry, can you comment?

mrry:
Looks like it should work on Windows (AIUI, `subprocess.check_output()` returns the same type on Windows and Linux, and the previous Windows failures were because we were hitting an untested failure case). The Windows presubmit should reveal whether it works or not....

vrv:

vrv:
Cool, kicking off tests.  Will merge if it's all passing.

Issue #3767, opened 2016-08-12
Incorrect Hessian of the tf.nn.sigmoid function


cjf00000:
GitHub issues are for bugs / installation problems / feature requests.  
For general support from the community, see [StackOverflow](https://stackoverflow.com/questions/tagged/tensorflow).
To make bugs and feature requests more easy to find and organize, we close issues that are deemed
out of scope for GitHub Issues and point people to StackOverflow.

For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.

### Environment info
Operating System: Linux 3.13.0-85-generic #129-Ubuntu SMP Thu Mar 17 20:50:15 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):
-rw-r--r-- 1 root root 322936 Aug 15  2015 /usr/local/cuda/lib64/libcudadevrt.a
lrwxrwxrwx 1 root root     16 Aug 15  2015 /usr/local/cuda/lib64/libcudart.so -> libcudart.so.7.5
lrwxrwxrwx 1 root root     19 Aug 15  2015 /usr/local/cuda/lib64/libcudart.so.7.5 -> libcudart.so.7.5.18
-rwxr-xr-x 1 root root 383336 Aug 15  2015 /usr/local/cuda/lib64/libcudart.so.7.5.18
-rw-r--r-- 1 root root 720192 Aug 15  2015 /usr/local/cuda/lib64/libcudart_static.a

If installed from binary pip package, provide:

1. Which pip package you installed.
tensorflow

2. The output from `python -c "import tensorflow; print(tensorflow.__version__)"`.
0.10.0rc0

If installed from source, provide 

1. The commit hash (`git rev-parse HEAD`)
2. The output of `bazel version`

### Steps to reproduce
I am trying to evaluate hessian of a logistic function via the following program.
```
# x = 1
# logistic(x) = 0.7310585786300049
# gradient(x) = logistic(x) * (1 - logistic(x)) = 0.19661193324148185
# hessian(x) = logistic(x) * (1 - logistic(x)) ^ 2 + -logistic(x) * (1 - logistic(x)) * logistic(x) = -0.09085774767294841
import tensorflow as tf

x = tf.Variable(1, name='x', dtype=tf.float32)
#logistic = 1. / (1 + tf.exp(-x))
[...]

Issue #2598, opened 2016-05-31
Hessian (calling tf.gradients twice) of tf.scan fails


dementrock:
GitHub issues are for bugs / installation problems / feature requests.  
For general support from the community, see [StackOverflow](https://stackoverflow.com/questions/tagged/tensorflow).
To make bugs and feature requests more easy to find and organize, we close issues that are deemed
out of scope for GitHub Issues and point people to StackOverflow.

For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.

### Environment info
Operating System: Mac OS X 10.11.2

Installed version of CUDA and cuDNN: None

If installed from sources, provide the commit hash:

4455f81c3dc07a77ac133dec24690968e121370a

### Steps to reproduce
Run the following script:

```python
import tensorflow as tf

theta = tf.Variable(initial_value=1.)


def fn(x, prev):
    return prev + x * theta

result = tf.scan(fn, [1., 2., 3.])

grad_theta = tf.gradients(result, theta)

tf.gradients(grad_theta, theta)
```

will result in the following error:

```
Traceback (most recent call last):
  File "sandbox/rocky/tf/small_example.py", line 13, in <module>
    tf.gradients(grad_theta, theta)
  File "/Users/dementrock/anaconda/envs/rllab/lib/python2.7/site-packages/tensorflow/python/ops/gradients.py", line 379, in gradients
    to_ops, from_ops)
  File "/Users/dementrock/anaconda/envs/rllab/lib/python2.7/site-packages/tensorflow/python/ops/gradients.py", line 185, in _PendingCount
    between_ops)
  File "/Users/dementrock/anaconda/envs/rllab/lib/python2.7/site-packages/tensorflow/python/ops/control_flow_ops.py", line 874, in MaybeCreateControlFlowState
    loop_state.AddWhileContext(op, between_op_list, between_ops)
  File "/Users/dementrock/anaconda/envs/rllab/lib/python2.7/site-packages/tensorflow/python/ops/control_flow_ops.py", line 738, in AddWhileContext
    for loop_exit in forward_ctxt.loop_exits:
TypeError: 'NoneType' object is not iterable
```

### What have you tried?
Nothing beyond creating this minimal reproducible example

### Logs or other output that would be helpful
(If logs are large, please upload as attachment).

concretevitamin:
Adding @yuanbyu@, @ebrevdo.

girving:
[...]

Issue #675, opened 2016-01-04
Gradients of non-scalars (higher rank Jacobians)


zackchase:
Currently if you call gradients(ys, xs), it will return the sum of dy/dx over all ys for each x in xs. I believe this doesn't accord with an a priori mathematical notion of the derivative of a vector. I'd like the way to take the derivative of ys wrt xs where both are vectors and have a Jacobian matrix returned. By extension, I'd like to take the derivative of a vector wrt a matrix and get back a 3-tensor. There doesn't seem to be a convenient tensorflow function to compute the Jacobian or higher order derivatives. Am I missing something or is this functionality that we could add?

keveman:
zackchase@, you are right about the current `gradients` function. Currently, you can compute the Jacobian of, say, a vector, by calling `gradients` multiple times, one for every scalar component (obtained by slicing) of the original vector, and reassembling the results. Contributions are welcome to make this nicer and efficient.

girving:
It'd be pretty hard to support gradients of non-scalars with our current setup, since it would require every gradient function to handle extra rank input.  The one possibility I could see would be if we add some sort of map facility to register how to add extra ranks to ops, then compute gradients with respect to extra rank by computing lower rank and calling the registered map transformations.

Someone asked for map a while back, so if anyone wanted to tackle this task that might be the way to go.  Handling it at the gradient function level is probably bad, since it would add required complexity to an existing feature.  Warning: This is a pretty large change, so a good deal of discussion would be in order before starting.

zackchase:
Hi Geoffrey, thanks for taking an interest in this issue. I was initially confused by the use of "rank" to describe the number of dimensions of the array. Should we avoid this name in the thread title and documentation to preempt confusion via overloading the linear algebra notion of rank?

girving:
Tensor rank is very standard terminology: http://mathworld.wolfram.com/TensorRank.html

zackchase:
Cool. The terminology gets funny when we talk about rank-R decompositions of tensors, meaning the tensor can be represented as a sum of R outer products of rank-1 tensors, but probably not a problem for us to solve here.

One thing I thought of is that I would like to compute the frobenius norm of the Jacobian of the log probabilities for use as a smoothness penalty as in a contractive autoencoder. In this case, as we only seek a scalar at the end, is there a more efficient method than separately calculating the derivative of each output with respect to the inputs?

girving:
[...]

PR #5329, opened 2016-11-01
Hessian with respect to one-dimensional tensors

tillahoffmann:
This PR adds an operation to compute the Hessian with respect to one-dimensional tensors.


goodfeli:
Looking at it

tillahoffmann:
The `HessianVectorProduct`, if I understand correctly, computes the inner product of the Hessian with some vector which is useful for some optimizers that make use of the Hessian for more efficient steps. It does not allow for the evaluation of the Hessian itself but I may have misunderstood.

I have restricted to the hessian with respect to one dimension because the reshaping necessary to do it for arbitrary-shape tensors would be quite involved.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+    The hessian of `sum(ys)` with respect to `x`.
+
+  Raises:
+    LookupError: if one of the operations between `x` and `ys` does not
+      have a registered gradient function.
+    ValueError: if the arguments are invalid
+  """
+  ndims = x.get_shape().ndims
+  if ndims is not None and ndims != 1:
+    raise ValueError("Hessians can only be computed with respect to one-dimensional tensors")
```
goodfeli: OK, I get what's going on here.

Usually the Hessian is defined in terms of a function that takes a tensor of shape [n] as input and returns a tensor of shape [] as output. The Hessian has shape [n, n]. Part of the definition of the Hessian is that it is a matrix.

One could imagine generalizing the concept of a Hessian matrix to be more tensorflow-ish by returning a Hessian tensor of shape `input_shape + input_shape`, so a function that takes a matrix of shape [m, n] as input would have a Hessian tensor of shape [m, n, m, n]. But I think such a feature would be very rarely used and there's not much point in implementing it unless someone actually requests it.

tillahoffmann: Yep, that's correct. Being able to compute the Hessian with respect to one-dimensional vectors is quite useful for sensitivity analysis because it allows one to test how tightly the loss function constrains the parameters of interest. But I don't think that feature would regularly be used with higher-dimensional objects (although a reshape would allow for that use-case).

vrv: Thanks for the comments @goodfeli, and thanks for the follow ups @tillahoffmann !  So is this something that we need a new function / interface for?  Is it too hard for users to just call gradients twice instead?

In general, TensorFlow tries to promote composition when possible, and only provides higher level interfaces when it is an extremely common pattern, or when there's something particularly tricky to get correct and we don't want users shooting themselves in the foot.  From what I can tell, this function doesn't really fit either of those criteria, but I could be convinced otherwise.

My inclination is to avoid adding this function just yet, mostly to keep maintenance burden low, unless we get an large number of people who think this is useful to have.  Thoughts?

goodfeli:
@vrv: tl;dr This looks like the correct way to implement the full Hessian matrix. You wouldn't want to use _hessian_vector_product to do it.

Long version:
For most problems, the Hessian matrix would be too big to store, so we use _hessian_vector_product to compute only a small Krylov subspace of the Hessian matrix. However, it can be useful to compute the full Hessian matrix for a small problem, for debugging or research purposes.

_hessian_vector_product takes the product between the Hessian "matrix" and a "vector". I use scare quotes here because the "vector" can actually be a list of tensors of arbitrary shape. We imagine that we reshape and concatenate these tensors to get one large 1-D tensor. It's not possible to call _hessian_vector_product with just a scalar `1` or an identity matrix as its argument; the argument needs to match the shapes of the parameter list.

To compute a Hessian matrix of shape [n, n] using _hessian_vector_product, you would need to call _hessian_vector_product n times. On call `i`, you pass a "vector" that is all zeros except for position `i`, and is 1 at position `i`. This is not very efficient because you have to construct those one-hot vectors.

Computing the full Hessian is not a very common operation, but if someone wants to do it, this PR looks like the right way to do it.

tensorflow/python/ops/gradients.py
tillahoffmann: I thought it might be useful because the (un)packing is a bit fiddly and when I was looking for an implementation I came across a number of people who wanted to do the same. Here are some examples:

* #2598
* [TensorFlow: Compute Hessian matrix (and higher order derivatives)](http://stackoverflow.com/questions/35266370/tensorflow-compute-hessian-matrix-and-higher-order-derivatives)
* [How to compute all second derivatives (only the diagonal of the Hessian matrix) in Tensorflow?](http://stackoverflow.com/questions/38200982/how-to-compute-all-second-derivatives-only-the-diagonal-of-the-hessian-matrix)
* #3767
* [Hessian computation using tensorflow](http://runopti.github.io/blog/2016/07/07/HessianComp/)
* #675

goodfeli:
@vrv : let me know if you need anything more from me---I wasn't sure if you wanted me to do a full TensorFlow code review or just to comment on algorithmic correctness.

tensorflow/python/ops/gradients.py
vrv: Okay, that's fair, and I can see that at some point, if someone wants to add support for the higher dimensional hessians, this would be a good place to handle all of the complicated shaping.  As long as you don't mind us assigning bugs / issues related to this function to you, we can accept it :).  I'll do a full review.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
 
   # Second backprop
   return gradients(elemwise_products, xs)
+
+
+def hessian(ys, x, name="hessian", colocate_gradients_with_ops=False, 
+            gate_gradients=False, aggregation_method=None):
+  """Constructs the Hessian of sum of `ys` with respect to `x`.
```
vrv: Somewhere in the docs it might be good to add a link to the definition of Hessian.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+
+  `hessian()` adds ops to the graph to output the Hessian of `ys` with 
+  respect to `x` and returns a `Tensor`.
+
+  Args:
+    ys: A `Tensor` or list of tensors to be differentiated.
+    x: A `Tensor` to be used for differentiation.
```
vrv: It would be great to make this interface future-proof.  Does it ever make sense for this to be a list of 'x's, like we do for gradients?  I know the current implementation would not work with that, but if it makes theoretical sense at some level, we should add it here, but check for the existence of only one tensor in the implementation for now.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+    name: Optional name to use for grouping all the gradient ops together.
+      defaults to 'gradients'.
```
vrv: 'hessian'

tensorflow/python/ops/gradients.py
tillahoffmann: Happy to do some maintenance. Am fiddling with tensorflow most days anyway.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+    colocate_gradients_with_ops: If True, try colocating gradients with
```
vrv: For the next three args, maybe say 'See 'gradients()' documentation for details', so we don't have to duplicate changes.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+    aggregation_method: Specifies the method used to combine gradient terms.
+      Accepted values are constants defined in the class `AggregationMethod`.
+
+  Returns:
+    The hessian of `sum(ys)` with respect to `x`.
+
+  Raises:
+    LookupError: if one of the operations between `x` and `ys` does not
+      have a registered gradient function.
+    ValueError: if the arguments are invalid
```
vrv: How about:

"ValueError: If the arguments are invalid or not supported.  Currently, this function only works when 'x' is known to be a vector."

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+  """
+  ndims = x.get_shape().ndims
+  if ndims is not None and ndims != 1:
```
vrv: This will do the wrong thing if 'x''s shape is not known.  I think you want to do:

if ndims is None or ndims != 1

So that it fails when we can't tell it's a vector or not.

We should also add a test for these error conditions.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+    raise ValueError("Hessians can only be computed with respect to one-dimensional tensors")
+  kwargs = {
+    'colocate_gradients_with_ops': colocate_gradients_with_ops,
+    'gate_gradients': gate_gradients,
+    'aggregation_method': aggregation_method
+  }
+  # Compute the regular gradients
+  _gradients, = gradients(ys, x, **kwargs)
```
vrv: For consistency with below, should you do _gradients = gradients(ys, x, **kwargs)[0] instead?

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
+  # Apply the gradients again
+  _hess = [gradients(_gradient, x, **kwargs)[0] for _gradient in array_ops.unpack(_gradients)]
```
vrv: Okay, I now appreciate a bit more about the subtlety of packing/unpacking now that I'm reviewing the code :).  

As far as I understand, _gradients is a single element Tensor containing the sum(dy/dx), and you're unpacking the tensor along the first dimension (which is guaranteed to be a 1D tensor) into its scalar components, and then performing gradients again for each element in the vector with respect to 'x'.  Finally, you pack the results together.

If this is correct, can you add some comments to describe what is being done here?  Future readers will greatly appreciate it!

tensorflow/python/ops/gradients_test.py
```diff
@@ -419,6 +419,27 @@ def testHessianVectorProduct(self):
       self.assertAllClose(hess_v_value, hess_v_actual)
 
 
+class HessianTest(test_util.TensorFlowTestCase):
+
+  def testHessian(self):
```
vrv: testHessian1D

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,45 @@ def _hessian_vector_product(ys, xs, v):
...
   # Second backprop
   return gradients(elemwise_products, xs)
+
+
+def hessian(ys, x, name="hessian", colocate_gradients_with_ops=False, 
+            gate_gradients=False, aggregation_method=None):
+  """Constructs the Hessian of sum of `ys` with respect to `x`.
+
+  `hessian()` adds ops to the graph to output the Hessian of `ys` with 
+  respect to `x` and returns a `Tensor`.
```
vrv: Let's add a comment somewhere here saying that currently this only supports 1D hessians.

tensorflow/python/ops/gradients.py
vrv: On this line, can we rephrase to say 'computing hessians is only currently supported on 1D tensors', to make it clear that it's possible but not yet implemented?

tensorflow/python/ops/gradients.py
tillahoffmann: I will make changes such that `hessian` returns a list of tensors of the same length of `xs` for consistency with `gradients` and will rename to `hessians`.

tensorflow/python/ops/gradients.py
tillahoffmann: That's correct. I've added a few more comments.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,64 @@ def _hessian_vector_product(ys, xs, v):
...
+    # Compute the partial derivatives of the input with respect to all 
+    # elements of `x`
+    _gradients = gradients(ys, x, **kwargs)[0]
+    # Unpack the gradients into a list so we can take derivatives with 
+    # respect to each element
+    _gradients = array_ops.unpack(_gradients)
+    # Compute the partial derivatives with respect to each element of the list
+    _hess = [gradients(_gradient, x, **kwargs)[0] for _gradient in _gradients]
+    # Pack the list into a matrix and add to the list of hessians
+    hessians.append(array_ops.pack(_hess, name=name))
```
vrv: I suspect this won't work when you have multiple x's, because there will be a name collision :).  A test that passes a list of 'x's should be added, and will catch this bug, I think

Alternatively, if you don't want to support multiple x's yet, accept it in the interface but throw a ValueError with 'unsupported' for now.

tensorflow/python/ops/gradients_test.py
```diff
@@ -419,6 +419,36 @@ def testHessianVectorProduct(self):
...
+  def testHessian1D(self):
+    # Manually compute the Hessian explicitly for a low-dimensional problem
+    # and check that `hessian` matches. Specifically, the Hessian of 
+    # f(x) = x^T A x is H = A + A^T.
+    m = 4
+    rng = np.random.RandomState([1, 2, 3])
+    mat_value = rng.randn(m, m).astype("float32")
+    x_value = rng.randn(m).astype("float32")
+    hess_value = mat_value + mat_value.T
+    for use_gpu in [False, True]:
```
vrv: here and below, you can avoid the loop and just do: use_gpu=True.  It will try to use gpu when available and use CPU otherwise.

vrv (changes requested):

tensorflow/python/ops/gradients.py
tillahoffmann: I think this works because tensorflow will rename the operations when they clash. E.g. changing the line in the tests to `hessians(quadratic, [x, x])` gives me

```python
[<tf.Tensor 'hessians_1:0' shape=(4, 4) dtype=float32>,
 <tf.Tensor 'hessians_2:0' shape=(4, 4) dtype=float32>]
```

Will double check and extend the tests.

tensorflow/python/ops/gradients.py
```diff
@@ -829,3 +829,66 @@ def _hessian_vector_product(ys, xs, v):
...
+      'gate_gradients': gate_gradients,
+      'aggregation_method': aggregation_method
+    }
+  # Compute a hessian matrix for each x in xs
+  hessians = []
+  for x in xs:
+    # Check dimensions
+    ndims = x.get_shape().ndims
+    if ndims is None:
+      raise ValueError('cannot compute Hessian because the dimensionality of ' \
```
vrv: Sorry, one more change: would it be possible to print some debug info for 'x'?  If this is a large list, it's going to be hard for the user to figure out which 'x' violated this.  Same in the elif.

vrv (changes requested):
Awesome!  Just one more suggestion and then I can kick off tests.

vrv:

vrv (approved):

lucfra:
Hello to all. 
I was wondering if there any specific reason why the` _hessian_vector_product` function is not present in the "public" api?

tillahoffmann:
I imagine the method exists for optimisation algorithms like [this one](http://machinelearning.wustl.edu/mlpapers/paper_files/icml2010_Martens10.pdf) which are not (yet) implemented in tensorflow.

PR #5330, opened 2016-11-01
Implementing WriteFile Op

### LINKED ISSUE #4471: Add `tf.write_file` op
Tensorflow currently has a `tf.read_file` op, as documented [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/api_docs/python/functions_and_classes/shard1/tf.read_file.md).

Can we have a `tf.write_file` as well?

It could be defined as something like:

``` python
def write_file(tensor, filename, name=None):
    def write_data(data):
        with open(filename, "wb") as handle:
            handle.write(data)

    with tf.name_scope("WriteFile" or name):
        tensor = tf.convert_to_tensor(tensor, name="Values")
        return tf.py_func(write_data, tensor)
```

There are some usecases where this would be helpful, such as generating images and audio and then writing them to disk. Presumably, this would be done after an op such as `encode_jpeg` or `encode_audio`.

If this already exists in some form, please close this – my apologies, but some googling as well as searching on Github did not find it.

malzantot:
Hello,
I created a pull request which implements `tf.write_file` OP requested [here](https://github.com/tensorflow/tensorflow/issues/4471). Please take a look.

Thanks,
Moustafa

Fixes #4471

### DIFF
--- a/tensorflow/core/kernels/whole_file_read_ops.cc
+++ b/tensorflow/core/kernels/whole_file_read_ops.cc
@@ -46,14 +46,6 @@ static Status ReadEntireFile(Env* env, const string& filename,
   return Status::OK();
 }
 
-static Status WriteFileContent(Env* env, const string& filename,
-		const string& contents) {
-	std::unique_ptr<WritableFile> file;
-	TF_RETURN_IF_ERROR(env->NewWritableFile(filename, &file));
-	TF_RETURN_IF_ERROR(file->Append(contents));
-	return Status::OK();
-}
-
 class WholeFileReader : public ReaderBase {
  public:
   WholeFileReader(Env* env, const string& node_name)
@@ -128,8 +120,8 @@ class ReadFileOp : public OpKernel {
 REGISTER_KERNEL_BUILDER(Name("ReadFile").Device(DEVICE_CPU), ReadFileOp);
 
 class WriteFileOp : public OpKernel {
+  using OpKernel::OpKernel;
   public:
-    using OpKernel::OpKernel;
     void Compute(OpKernelContext* context) override {
 	    const Tensor* filename_input;
 	    const Tensor* contents_input;
@@ -144,12 +136,15 @@ class WriteFileOp : public OpKernel {
 			    errors::InvalidArgument(
 				    "Contents tensor must be scalar, but had shape: ",
 				    contents_input->shape().DebugString()));
-	    OP_REQUIRES_OK(context, WriteFileContent(context->env(),
+	    OP_REQUIRES_OK(context, context->allocate_output("output", TensorShape({}),
+				    &output));
+	    if(WriteStringToFile(context->env(),
 				    filename_input->scalar<string>()(),
-				    contents_input->scalar<string>()()));
-	    OP_REQUIRES_OK(context, context->allocate_output("output", TensorShape({}), &output));
-	     output->scalar<int>();
-
+				    contents_input->scalar<string>()()).ok()) {
+	    	output->scalar<int>()(0) = 0;
+	    } else {
+	    	output->scalar<int>()(0) = -1;
+	    }
     }
 };
 
--- a/tensorflow/core/kernels/whole_file_read_ops.cc
+++ b/tensorflow/core/kernels/whole_file_read_ops.cc
@@ -125,19 +125,25 @@ class WriteFileOp : public OpKernel {
     void Compute(OpKernelContext* context) override {
 	    const Tensor* filename_input;
 	    const Tensor* contents_input;
-	    OP_REQUIRES_OK(context, context->input("filename", &filename_input));
-	    OP_REQUIRES_OK(context, context->input("contents", &contents_input));
-	    OP_REQUIRES(context, TensorShapeUtils::IsScalar(filename_input->shape()),
+	    OP_REQUIRES_OK(context,
+			   context->input("filename", &filename_input));
+	    OP_REQUIRES_OK(context,
+			   context->input("contents", &contents_input));
+	    OP_REQUIRES(context,
+			TensorShapeUtils::IsScalar(filename_input->shape()),
+			errors::InvalidArgument(
+			"Input filename tensor must be scalar, but had shape: ",
+			filename_input->shape().DebugString()));
+	    OP_REQUIRES(context,
+			TensorShapeUtils::IsScalar(contents_input->shape()),
 			    errors::InvalidArgument(
-				    "Input filename tensor must be scalar, but had shape: ",
-				    filename_input->shape().DebugString()));
-	    OP_REQUIRES(context, TensorShapeUtils::IsScalar(contents_input->shape()),
-			    errors::InvalidArgument(
-				    "Contents tensor must be scalar, but had shape: ",
-				    contents_input->shape().DebugString()));
-	    TF_CHECK_OK(WriteStringToFile(context->env(),
-				    filename_input->scalar<string>()(),
-				    contents_input->scalar<string>()()));
+			    "Contents tensor must be scalar, but had shape: ",
+			    contents_input->shape().DebugString()));
+	    OP_REQUIRES_OK(context,
+			   WriteStringToFile(context->env(),
+				             filename_input->scalar<string>()(),
+				             contents_input->scalar<string>()())
+			   );
     }
 };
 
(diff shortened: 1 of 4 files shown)


malzantot:
PS: I just submitted the CLA

tensorflow/core/ops/io_ops.cc
```diff
@@ -582,6 +582,15 @@ REGISTER_OP("ReadFile")
 Reads and outputs the entire contents of the input filename.
 )doc");
 
+REGISTER_OP("WriteFile")
+    .Input("filename: string")
+    .Input("contents: string")
+    .Output("output: int32")
+    .SetShapeFn(shape_inference::ScalarShape)
+    .Doc(R"doc(
+Writes contents to the end of input filename. Creates file if not existing.
```
concretevitamin: Could you add documents on the args and outputs?  In particular, it's worth pointing out what `output`'s shape and semantics are.

tensorflow/core/kernels/whole_file_read_ops.cc
```diff
@@ -119,4 +127,31 @@ class ReadFileOp : public OpKernel {
 
 REGISTER_KERNEL_BUILDER(Name("ReadFile").Device(DEVICE_CPU), ReadFileOp);
 
+class WriteFileOp : public OpKernel {
+  public:
+    using OpKernel::OpKernel;
```
concretevitamin: Is this line needed?

tensorflow/core/kernels/whole_file_read_ops.cc
```diff
@@ -119,4 +127,31 @@ class ReadFileOp : public OpKernel {
...
+				    filename_input->shape().DebugString()));
+	    OP_REQUIRES(context, TensorShapeUtils::IsScalar(contents_input->shape()),
+			    errors::InvalidArgument(
+				    "Contents tensor must be scalar, but had shape: ",
+				    contents_input->shape().DebugString()));
+	    OP_REQUIRES_OK(context, WriteFileContent(context->env(),
+				    filename_input->scalar<string>()(),
+				    contents_input->scalar<string>()()));
+	    OP_REQUIRES_OK(context, context->allocate_output("output", TensorShape({}), &output));
+	     output->scalar<int>();
```
concretevitamin: What's `output` really for?  Should we do a 0/1 indicator?

tensorflow/core/kernels/whole_file_read_ops.cc
```diff
@@ -46,6 +46,14 @@ static Status ReadEntireFile(Env* env, const string& filename,
   return Status::OK();
 }
 
+static Status WriteFileContent(Env* env, const string& filename,
```
concretevitamin: This helper is not needed -- see core/platform/env.h's `WriteStringToFile`.

tensorflow/core/ops/io_ops.cc
concretevitamin: Semantics: I believe "Write" inherently refers to "overwrite".  Let's change the semantics from appending to overwriting?

concretevitamin (changes requested):
Thanks!  A couple of minor comments.

tensorflow/core/kernels/whole_file_read_ops.cc
malzantot: If I remove this line compilation fails with "no matching constructor " error message.
`error: no matching constructor for initialization of 'tensorflow::WriteFileOp'
 note: candidate constructor (the implicit copy constructor) not viable: no known conversion from '::tensorflow::OpKernelConstruction *' to 'const tensorflow::WriteFileOp' for 1st argument
class WriteFileOp : public OpKernel {
`

tensorflow/core/ops/io_ops.cc
malzantot: I agree, The `write_file` actually overwrites the file content. Updated the documentation.

tensorflow/core/kernels/whole_file_read_ops.cc
malzantot: Done.

malzantot:
Hey @concretevitamin, 

Thanks for your review. PTAL!

tensorflow/core/kernels/whole_file_read_ops.cc
malzantot: done.

tensorflow/core/ops/io_ops.cc
vrv: Why does this op need an output?  Who is going to consume the output?

Should we just have it produce no outputs and do an OP_REQUIRES(WriteStringToFile...) in its implementation?

tensorflow/core/kernels/whole_file_read_ops.cc
```diff
@@ -119,4 +119,34 @@ class ReadFileOp : public OpKernel {
...
+			    errors::InvalidArgument(
+				    "Input filename tensor must be scalar, but had shape: ",
+				    filename_input->shape().DebugString()));
+	    OP_REQUIRES(context, TensorShapeUtils::IsScalar(contents_input->shape()),
+			    errors::InvalidArgument(
+				    "Contents tensor must be scalar, but had shape: ",
+				    contents_input->shape().DebugString()));
+	    OP_REQUIRES_OK(context, context->allocate_output("output", TensorShape({}),
+				    &output));
+	    if(WriteStringToFile(context->env(),
```
concretevitamin: Nit: Space after if

tensorflow/core/kernels/whole_file_read_ops.cc
```diff
@@ -119,4 +119,34 @@ class ReadFileOp : public OpKernel {
...
+				    filename_input->scalar<string>()(),
+				    contents_input->scalar<string>()()).ok()) {
+	    	output->scalar<int>()(0) = 0;
+	    } else {
+	    	output->scalar<int>()(0) = -1;
```
concretevitamin: On second thought, I think it'll be less burden on user of this Op just throw an error if the write is unsuccessful.  (Otherwise every responsible caller of this Op need to do something like `tf.assert_equal(tf.write_file(...), 0)`.)

Let's get rid of the output, and do `OP_REQUIRES_OK(context, WriteStringToFile(...))` directly here (when it fails, a nice stack trace will be printed out).

malzantot: Thanks,
I have a question; If we get rid of output, what would be the appropriate update we set to the op `SetShapeFn` in `tensorflow/core/kernels/io_ops.cc`, and also for the shape registration in 
`tensorflow/python/ops/io_ops.py` ?

concretevitamin: Take a look at the shape functions of `MergeV2Checkpoints` and `Save`: both have no outputs, so their shape fns just enforce some properties about the inputs.

The one in the Python side does not change -- it just means "use the shape fn registered in C++'s side".

malzantot:
Hi @concretevitamin ,

Thanks for your review. PTAL !

tensorflow/core/kernels/whole_file_read_ops.cc
malzantot: Done. Thanks.

tensorflow/core/kernels/whole_file_read_ops.cc
```diff
@@ -136,15 +135,9 @@ class WriteFileOp : public OpKernel {
 			    errors::InvalidArgument(
 				    "Contents tensor must be scalar, but had shape: ",
 				    contents_input->shape().DebugString()));
-	    OP_REQUIRES_OK(context, context->allocate_output("output", TensorShape({}),
-				    &output));
-	    if(WriteStringToFile(context->env(),
+	    TF_CHECK_OK(WriteStringToFile(context->env(),
```
concretevitamin: Prefer `OP_REQUIRES_OK(context, ...)` over `TF_CHECK_OK(...)` here?

concretevitamin (approved):
Just one final nit - but otherwise looks awesome.  Thanks for your contribution!

tensorflow/core/kernels/whole_file_read_ops.cc
malzantot: Done. Thanks for your reviews.

tensorflow/core/kernels/whole_file_read_ops.cc
```diff
@@ -119,4 +119,33 @@ class ReadFileOp : public OpKernel {
 
 REGISTER_KERNEL_BUILDER(Name("ReadFile").Device(DEVICE_CPU), ReadFileOp);
 
+class WriteFileOp : public OpKernel {
+  public:
+    using OpKernel::OpKernel;
+    void Compute(OpKernelContext* context) override {
+	    const Tensor* filename_input;
```
vrv: can you run clang-format --style=google on all of these files?  That way it'll conform to the rest of the codebase.

malzantot: Done. Thanks for the tip. PTAL.

concretevitamin:
Jenkins test this please

PR #5346, opened 2016-11-02
Fix sourceforge links for TF external dependencies.

gunan:


gunan:
tensorflow/tensorboard/backend:server_test  flaked.
Jenkins, test this please.

caisq (approved):

snnn:
Why rollback this change? ufpr.dl.sourceforge.net is not reachable again.

martinwicke:
Actually this change seems to have been clobbered by a bad merge. 

We have some redundancy now though, and we mirror the package and shouldn't need the sourceforge server at all. Do you see issues with it?

snnn:
@martinwicke , Yes, it broke the cmake build.  Later on, it was fixed by 6446895aa6e35b88c0344ac1f976b06826617cbb

Issue #5347, opened 2016-11-02
Change http://ufpr.dl.sourceforge.net links for http_archive to another more stable one


llhe:
TensorFlow build dependens on external links, for example:
```python
  native.new_http_archive(                                                      
    name = "gif_archive",                                                       
    url = "http://ufpr.dl.sourceforge.net/project/giflib/giflib-5.1.4.tar.gz",
    sha256 = "34a7377ba834397db019e8eb122e551a49c98f49df75ec3fcc92b9a794a4f6d1",
    strip_prefix = "giflib-5.1.4/lib",                                          
    build_file = str(Label("//:gif.BUILD")),                                    
  )
```
Looks like http://ufpr.dl.sourceforge.net is just a SF mirror located in Brazil, and not quite stable, see #3929, #4085. I also encountered this some times. Considering change to another stable one, for example: http://downloads.sourceforge.net

A list of SF sites/mirrors:
https://github.com/fink/fink-mirrors/blob/master/sourceforge

llhe:
By the way, I can't access it both in China and US-West, now, so the reason is that mirror is down instead of my network connection issue.

aselle:
Thanks for reporting the issue! @gunan, I don't know if there was a specific reason we chose than link, but the fix seems to make sense.

gunan:
#5346 already addressed this.
To the best of my knowledge, debian links should also be ok in China.

PR #5348, opened 2016-11-02
Change sourceforge links from the unstable mirror to the primary site

llhe:
This change address this issue: https://github.com/tensorflow/tensorflow/issues/5347


caisq:
PR https://github.com/tensorflow/tensorflow/pull/5346 should have addressed the same issue. Thank you for the pull request all the same, @llhe

PR #5353, opened 2016-11-02
Changing Bazel dependencies

fomiciov:


jart:
For future reference, please do not use git_repository. Always use {new_,}http_archive.

fomiciov:
This PR was opened by mistake, but thanks for the advice @jart 👍

Issue #3682, opened 2016-08-07
Consider adding expm1 and log1p


NeilGirdhar:
`expm1` is useful for inverting `tf.nn.softplus`:

```
def inverse_softplus(x):
    return tf.select(tf.gt(x, 80.0),
                     x,
                     tf.log(tf.expm1(x)))
```

`log1p` is useful for implementing `logaddexp`:

```
def logaddexp(x, y):
    temp = x - y
    return tf.select(tf.gt(temp, 0.0),
                     x + tf.log1p(tf.exp(-temp)),
                     y + tf.log1p(tf.exp(temp)))
```

girving:
Yep, this would both be nice to have.

ibab:
I've had a quick look at Eigen to see whether these two are already supported as functors.
`log1p` is there, but `expm1` is missing.
That means a good start for `expm1` would be to add it to Eigen.
I think some extra work in Eigen is also needed to get `log1p` to work with GPU ops (e.g. adding it to `Eigen/src/Core/arch/CUDA/MathFunctions.h`)

kbrems:
As a potential alternative, we have implemented a python-only framework for adding custom operators to tensorflow, the Operator Vectorization Library: 
https://github.com/opveclib/opveclib. Multi-threaded C++ and CUDA code is automatically generated from the python operators, which can be used in a Tensorflow application with only a binary install of tensorflow. We have implemented the log1p and expm1 operators as an example. See documentation at http://opveclib.readthedocs.io/en/latest/expm1.html

girving:
@kbrems Interesting!

tillahoffmann:
This should be fixed by 6731e21360ac2fd2fa4999a12e87ae72acfcb2d2.

PR #5356, opened 2016-11-02
Add log1p

tillahoffmann:
This PR adds the `log1p` function and partially addresses #3682. It also uses `log1p` in the computation of the sigmoid cross entropy to improve numerical stability (I had issues with rare events).


tensorflow/core/kernels/cwise_op_log1p.cc
```diff
@@ -0,0 +1,24 @@
+/* Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
vrv: Probably 2016 now

vrv (commented):
Very nice!!

vrv:

vrv:
Yes, seems reasonable to me.  I don't know why the tests are written like that anyway.

tillahoffmann:
Could just change them to perform an equivalent computation using numpy routines.

vrv:
@zakariahaque can you comment?

zakariahaque:
@tillahoffmann, thanks for the change.  Feel free to update the loss in the tests.
@vrv all the test needs to check is that the correct loss was used for that type of head/target_column.

vrv:

vrv (approved):

Issue #5357, opened 2016-11-02
Bazel fails to clean


Randl:
Bazel fails to clean with following message
```
WARNING: Output base '/home/username/.cache/bazel/_bazel_username/acf2af86672e1552dfd6edd47d54a950' is on NFS. This may lead to surprising failures and undetermined behavior.
.
INFO: Starting clean (this may take a while). Consider using --expunge_async if the clean takes more than several minutes.
ERROR: /home/username/.cache/bazel/_bazel_username/acf2af86672e1552dfd6edd47d54a950/server (Directory not empty).
```
Replacement of `bazel clean --expunge` with `bazel clean --expunge_async` [here](https://github.com/tensorflow/tensorflow/blob/master/configure#L25) fixes the issue.
### Environment info
Operating System:
Ubuntu 16.04

Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):
```
$ ls -l /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcud*
-rw-r--r-- 1 root root 558720 Sep 15 02:02 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudadevrt.a
lrwxrwxrwx 1 root root     16 Sep 15 02:05 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so -> libcudart.so.8.0
lrwxrwxrwx 1 root root     19 Sep 15 02:05 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so.8.0 -> libcudart.so.8.0.44
-rw-r--r-- 1 root root 415432 Sep 15 02:02 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so.8.0.44
-rw-r--r-- 1 root root 775162 Sep 15 02:02 /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart_static.a
``` 

1. The commit hash (`git rev-parse HEAD`)
```
$ git rev-parse HEAD
7443479c9261b893d00c923bf53d924bbae9bc64
```
2. The output of `bazel version`
```
$ bazel version
WARNING: Output base '/home/username/.cache/bazel/_bazel_chaimb/acf2af86672e1552dfd6edd47d54a950' is on NFS. This may lead to surprising failures and undetermined behavior.
Build label: 0.3.2
[...]

Issue #5253, opened 2016-10-28
Let user to choice whether to do bazel clean during configure


EricFisher:
Currently running the configure will do 'bazel clean --expunge' by default.  It is OK for the first time to run configure.  But when I need to rerun the configure (because some errors occur, I will retry), I have to spend much time to download files from internet again each time during to the execution of 'bazel clean --expunge'. 

So, should we let user to choice whether to do bazel clean during configure?

drpngx:
@jart 

I have also been puzzled by that. It prevents using the cache, and doesn't really belong in a configure script. You can use sed s/bazel.clean/true/ <configure|bash as a workaround.

jart:
The `bazel clean --expunge` and `bazel fetch //tensorflow/...` need to happen unfortunately to work around bugs in Bazel. There's issues relating to cache invalidation on remote repositories. There's also issues relating to passing environment variables to repository rules. CC: @davidzchen

It'd probably be better if we didn't have an option to disable the clean and fetch. We want to guarantee that builds work, even if that means things go a little slower. We're going to be improving the state of the build real soon though. We're actually planning to make the configure script optional. I think the Bazel team is doing amazing work helping to make that happen.

Thanks for bringing this to our attention. Feel free to post any other questions you might have and I'll be happy to answer them.

jart:
One other note, if you're 100% certain you want to roll the dice by disabling the clean and fetch, it shouldn't be too difficult to comment out those lines in the configure script.

drpngx:
If you have a link to the issue, please post here and/or in the code.

jart:
It's possible that the related issues are https://github.com/bazelbuild/bazel/issues/1022 and https://github.com/bazelbuild/bazel/issues/1595 and https://github.com/bazelbuild/bazel/issues/974. There was another one, but I'm having trouble digging it up.

Also here is an explanation that @davidzchen gave me a week ago regarding the necessity of fetch.

> The reason why we fetch in the configure script is because the way the script
> passes values to the cuda_configure workspace rule is via environment variables
> (see comment at the top of [cuda_configure.bzl](https://github.com/tensorflow/tensorflow/blob/d95a5b7cc6dca2d3603d73da1078274b29639ec7/third_party/gpus/cuda_configure.bzl)).

> The configure script exports these environment variables before calling bazel
> fetch, which will set up external repositories, such as @local_config_cuda,
> according to the values provided by the user. However, if we do not do this,
[...]

PR #5359, opened 2016-11-02
Replace expunge with expunge_async in bazel clean

Randl:
https://github.com/tensorflow/tensorflow/issues/5357


Randl:
Signed CLA

drpngx:
@jart, related to #5253.

damienmg:
This change might be problematic (generate some race conditions), and is not solving the underlying problem.

vrv:
Thanks @damienmg. I'll close this and we should either follow up on the referenced bug, or at bazelbuild/bazel, since this is more about bazel?

Issue #5321, opened 2016-11-01
Filesystem FileExists interface is broken


llhe:
The FileSystem define the following interface, which can't handle temporary runtime errors properly but just swallow them (like GCS, HDFS and any customized DFS),
```c++
virtual bool FileExists(const string& fname) = 0;
```
It should be changed to
```c++
virtual Status FileExists(const string& fname) = 0;
```
I can provide a patch if no one is working on this.

aselle:
@jhseu, I'm marking this contributions welcome, assuming you agree with that change.

jhseu:
Yeah, I agree that's an issue. I can make the change, though, because we also have to change the implementation for our internal filesystem.

llhe:
This is more friendly for the caller:
```c++
virtual Status FileExists(const string& fname, bool* result) = 0;
```

jhseu:
Fixed internally. It'll show up during the next sync.

As mentioned in your pull request, we decided on:
```c++
virtual Status FileExists(const string& fname);
```
for consistency with how we've been using these error codes elsewhere.

PR #5370, opened 2016-11-03
Change file system FileExists interface to return Status

llhe:
The pull request addresses issue https://github.com/tensorflow/tensorflow/issues/5321.

The calling code are modified to handle error condition at best effort, if the outer caller function does not return error, it's simply changed to
```c++
FileExists(name, &result).ok() && result
```

Tests:
Tests run except windows implementation. HDFS is tested with 2.7.3 pseudo-distributed mode.

PS: The folowwing tests may fail when all tests run together, but pass when run standalone, should not related to this change
```
//tensorflow/contrib/session_bundle:exporter_test
//tensorflow/tensorboard/backend:server_test
//tensorflow/contrib/tensor_forest/hybrid:k_feature_routing_function_op_test
//tensorflow/contrib/learn:random_forest_test
```

### DIFF
--- a/tensorflow/cc/saved_model/loader.cc
+++ b/tensorflow/cc/saved_model/loader.cc
@@ -42,13 +42,17 @@ constexpr char kLoadAttemptSuccess[] = "success";
 Status ReadSavedModel(const string& export_dir, SavedModel* saved_model_proto) {
   const string saved_model_pb_path =
       io::JoinPath(export_dir, kSavedModelFilenamePb);
-  if (Env::Default()->FileExists(saved_model_pb_path)) {
+  bool result;
+  TF_RETURN_IF_ERROR(Env::Default()->FileExists(saved_model_pb_path, &result));
+  if (result) {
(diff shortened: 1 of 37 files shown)


vrv:
Thanks for making this change!  It's going to be easier for us to make this change internally, since it changes a core public interface that a lot of people are using internally.  

cc @jhseu @rohan100jain, perhaps use this change as a seed for the internal change, and make sure to credit @llhe ?

Alternatively, we can keep the old interface around so we can pull this change internally without too many problems, and then get rid of it once all of the internal uses are changed.

jhseu:
I actually already have an equivalent change internally that I was going to send out for review today. There are quite a few call sites internally, so it's easier to make the change there.

Note that I'm going to change the signature to:
`Status FileExists(const string& fname);` where it returns NOT_FOUND when it doesn't exist. That matches some existing conventions that we're using.

vrv:
Okay, cool, I will close this and we'll have the internal fix pushed out soon.  Thanks @llhe !

Issue #5322, opened 2016-11-01
Remove unnecessary GCS customized python code


llhe:
We have a GCS-like file system (blob store) and added the support without any customized python code (event summary input for tensorboard is not needed either).

It's better to clean them up. At least, refactor and clean up the following code because wen shouldn't change any python code when we adding a new file system:
```python
    if (io_wrapper.IsGCSPath(specification) or                                  
        specification.startswith('hdfs://')): 
```

I can contribute if no one is working on this.

aselle:
@jhseu, since you are doing HDFS work, do you have any thoughts on this or know who might be a good person to consult. It seems like these checks could be encapsulated into some kind of filesystem registry, but maybe we already have plans to do that?  If not, please mark this contributions welcome if you agree with the premise.

jhseu:
@llhe Yeah, if you have any ideas about what to do there, contributions are welcome. Perhaps we should just check if it's a URI?

This is very tensorboard-specific code, so I'm not sure it's worth plumbing through filesystem registry from C++ to make it work.

jhseu:
If you send a pull request, mention me on it.

llhe:
@aselle @jhseu I think just check the URI is ok, check if it's start with xyz://, which won't conflict with the event file lists like group_name:/path/to/directory (Looks like the current code has a bug for group_name:xys://?). So it can be fixed without touching the C++ code.

PR #5376, opened 2016-11-03
Remove hard coded file system scheme checking in tensorboard code

llhe:
This pull request addresses https://github.com/tensorflow/tensorflow/issues/5322.

This patch also fixes the case `run_name:gs://path` incorrect parsing issue.

There is still one file not cleaned up: [directory_watcher.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/summary/impl/directory_watcher.py#L181). However, this does not affecting adding other file systems.

That one can't be removed now because in GCS `Stat` for folder is not supported. However, according to [GCS API document](https://cloud.google.com/storage/docs/json_api/v1/objects/list), stat for a folder can be supported by summing sizes of all objects with that common prefix (In fact, our blob store has similar interface, and stat for folder is implemented likewise).

### DIFF
--- a/tensorflow/tensorboard/backend/server.py
+++ b/tensorflow/tensorboard/backend/server.py
@@ -25,6 +25,7 @@ import functools
 import os
 import threading
 import time
+import re
 
 import six
 from six.moves import BaseHTTPServer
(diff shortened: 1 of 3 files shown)


llhe:
@jhseu I added a simple pure python fix.

tensorflow/tensorboard/backend/server.py
```diff
@@ -67,21 +68,19 @@ def ParseEventFilesSpec(logdir):
   files = {}
   if logdir is None:
     return files
+  uri_pattern = re.compile("[^:/]+://.*")
```
jhseu: Can we make this:
"[a-zA-Z][0-9a-zA-Z.]*://.*" to match ParseURI?
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/file_system.cc#L82

(Note that function might get moved to core/lib/io/path.cc by the time you see it)

llhe: Thanks for the comments, @jhseu.
I rebased against the latest code and fix this.

jhseu (approved):

jhseu:
Jenkins, test this please

PR #5382, opened 2016-11-03
Fix a potential bug for android demo

philokey:
Size of the priority queue will decrease in every iteration, so the terminal condition will change,  but i increase in every iteration, it will lead to some problems. For example, when `MAX_RESULTS = 10 `and` pq.size() = 4` in the beginning, after 2 iterations,  `i = 2`, the  terminal condition will become `i < 2`, but it  is wrong.

### DIFF
--- a/tensorflow/examples/android/src/org/tensorflow/demo/TensorFlowImageClassifier.java
+++ b/tensorflow/examples/android/src/org/tensorflow/demo/TensorFlowImageClassifier.java
@@ -160,8 +160,8 @@ public class TensorFlowImageClassifier implements Classifier {
       }
     }
     final ArrayList<Recognition> recognitions = new ArrayList<Recognition>();
-    int recognitions_size = Math.min(pq.size(), MAX_RESULTS);
-    for (int i = 0; i < recognitions_size; ++i) {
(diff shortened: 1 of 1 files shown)


vrv:
Do you plan on signing the CLA?

tensorflow/examples/android/src/org/tensorflow/demo/TensorFlowImageClassifier.java
```diff
@@ -160,7 +160,8 @@ public int compare(Recognition lhs, Recognition rhs) {
       }
     }
     final ArrayList<Recognition> recognitions = new ArrayList<Recognition>();
-    for (int i = 0; i < Math.min(pq.size(), MAX_RESULTS); ++i) {
+    int recognitions_size = Math.min(pq.size(), MAX_RESULTS);
+    for (int i = 0; i < recognitions_size; ++i) {
```
philokey: I fixed the problem.

philokey:
@vrv Yes, The application has just been submitted。

andrewharp:
Jenkins, test this please.

andrewharp:
Thanks for the fix!

PR #5385, opened 2016-11-03
Create a script for building tensorflow pip package on windows using bazel.

gunan:
Scripts created by @meteorcloudy

### DIFF
--- a/tensorflow/tools/ci_build/windows/cpu/pip/build_tf_windows.sh
+++ b/tensorflow/tools/ci_build/windows/cpu/pip/build_tf_windows.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# This script assumes the standard setup on tensorflow Jenkins windows machines.
+# It is NOT guaranteed to work on any other machine. Use at your own risk!
+#
+# REQUIREMENTS:
+# * All installed in standard locations:
+#   - JDK8, and JAVA_HOME set.
+#   - Microsoft Visual Studio 2015 Community Edition
+#   - Msys2
+#   - Anaconda3
+# * Bazel windows executable copied as "bazel.exe" and included in PATH.
+
+# This script is under <repo_root>/tensorflow/tools/ci_build/windows/cpu/pip/
+# Change into repository root.
+script_dir=$(dirname $0)
+cd ${script_dir%%tensorflow/tools/ci_build/windows/cpu/pip}
+
+# Use a temporary directory with a short name.
+export TMPDIR="C:/tmp"
+
+# Set bash path
+export BAZEL_SH="C:/tools/msys64/usr/bin/bash"
+
+# Set Python path for ./configure
+export PYTHON_BIN_PATH="C:/Program Files/Anaconda3/python"
+
+# Set Python path for cc_configure.bzl
+export BAZEL_PYTHON="C:/Program Files/Anaconda3/python"
+
+# Set Visual Studio path
+export BAZEL_VS="C:/Program Files (x86)/Microsoft Visual Studio 14.0"
+
+# Add python into PATH, it's needed because gen_git_source.py uses
+# '/usr/bin/env python' as a shebang
+export PATH="/c/Program Files/Anaconda3:$PATH"
+
+# bazel clean --expunge doesn't work on Windows yet.
+# Clean the output base manually to ensure build correctness
+bazel clean
+output_base=$(bazel info output_base)
+bazel shutdown
+rm -rf ${output_base}
+
+echo "" | ./configure
+
+bazel build -c opt --cpu=x64_windows_msvc --host_cpu=x64_windows_msvc\
+    --copt="/w" --verbose_failures --experimental_ui\
+      tensorflow/tools/pip_package:build_pip_package || exit $?
+
+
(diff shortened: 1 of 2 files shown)


gunan:
Manually killed all tests, jenkins is overloaded and I needed the PR to start testing windows bazel build.

meteorcloudy:
This change looks good to me!

gunan:
http://ci.tensorflow.org/job/tensorflow-pull-requests-windows-bazel/4/console

@meteorcloudy How about the failure above.
Is it caused by this change, or something missing here?
Or is it simply a breakage existing in the repo and we address it separately.

meteorcloudy:
Ah.. Looks like we are hitting the [Windows long path issue](http://stackoverflow.com/questions/1880321/why-does-the-260-character-path-length-limit-exist-in-windows).
Basically, Windows doesn't support file path longer than 260 characters. Unfortunately, AFAIK, there's no workaround for cl.exe. 

This error is because `tensorflow-pull-requests-windows-bazel` is long enough to make the file path exceed that limitation.

We'll have to work around this by setting `export TMPDIR=c:/temp`, so compare to `C:\\tools\\msys64\\var\\tmp` we can save some characters...

meteorcloudy:
I just noticed there's no need to use double back slash `\\`, this could save us a lot of characters, I'll fix it in the next Bazel release.

meteorcloudy:
```
15:18:35 ERROR: C:/tmp/Bazel/3SovMC5L/external/protobuf/BUILD:579:1: Linking of rule '@protobuf//:internal/_api_implementation.so' failed: msvc_link.bat failed: error executing command 
15:18:35   cd C:/tmp/Bazel/3SovMC5L/execroot/tensorflow-pull-requests-windows-bazel
15:18:35   SET PATH=external/local_config_cc/wrapper/bin
15:18:35   external/local_config_cc/wrapper/bin/msvc_link.bat /DLL /WHOLEARCHIVE -m64 -Xcompilation-mode=opt -Wl,@bazel-out/vc_14_0_x64-py3-opt/bin/external/protobuf/internal/_api_implementation.so-2.params: com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1120
15:18:35 LINK : warning LNK4001: no object files specified; libraries used
15:18:35 LINK : error LNK2001: unresolved external symbol _DllMainCRTStartup
15:18:35 bazel-out/vc_14_0_x64-py3-opt/bin/external/protobuf/internal/_api_implementation.so : fatal error LNK1120: 1 unresolved externals
15:18:35 ERROR: C:/tmp/Bazel/3SovMC5L/external/zlib_archive/BUILD:5:1: output 'external/zlib_archive/libzlib.a' was not created
15:18:35 ERROR: C:/tmp/Bazel/3SovMC5L/external/grpc/BUILD:69:1: output 'external/grpc/libgpr.a' was not created
15:18:35 ERROR: C:/tf_jenkins/home/workspace/tensorflow-pull-requests-windows-bazel/tensorflow/core/BUILD:1112:1: declared output 'tensorflow/core/util/version_info.cc' was not created by genrule. This is probably because the genrule actually didn't create this output, or because the output was a directory and the genrule was run remotely (note that only the contents of declared file outputs are copied from genrules run remotely)
```
I saw the new error message on ci, I can reproduce the same error when using "tensorflow-pull-requests-windows-bazel" as the workspace folder name.
Although the error message is not obvious, the underlying problem is still caused by long path. Is it ok to rename the job name to something like `tf-pr-win-bazel`?
So sorry for the trouble, I will fix this soon. (actually, we can only ease this by not using `\\` to save some characters)

gunan:
A-ha, now it is running:
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/1/console

gunan:
Hehe, I don't think I can approve this PR :)
@meteorcloudy with the new job name, do we need the TMPDIR environment variable?
Also, should I "set -e" in the script?
Not sure how msys bash works.

meteorcloudy:
@gunan Probably yes, it's better we still have it for now.
I can only confirm we don't need it when job name is as short as `tensorflow`
Setting `set -e` sounds good to me, just add it in the beginning of the script, it should be the same as linux.

Why can't you approve this? What else need to be done?

gunan:
Since I am the author of the PR, it wont let me approve.
I will add set -x and set -e to the bash piece.

meteorcloudy:
Haha, I see~  :)

gunan:
@meteorcloudy Final question:
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/2/consoleFull

We see this error, it does not seem to affect the outcome, but all bazel builds have it:

00:00:36.893 ERROR: CreateFile(C:\tmp\Bazel\hzVGLx29\install): The system cannot find the file specified.

Any idea what it is?

meteorcloudy:
@gunan It happens when running Bazel in a new workspace for the first time, Bazel tries to unzip itself to a install directory. But yes, we should get rid of this error message since it's not actually causing any problem.
We have an issue here: https://github.com/bazelbuild/bazel/issues/1744

Vijay just invited me to join TensorFlow organization, does that mean I can approve this? :)

mrry (approved):
LGTM.

PR #5395, opened 2016-11-04
Update examples/tutorials/mnist/mnist_softmax.py

littlekfc:
The comment is wrong. tf.softmax(y) is predicted probability distribution while y_ is the label. Check more about cross-entropy(http://colah.github.io/posts/2015-09-Visual-Information/#cross-entropy)

### DIFF
--- a/tensorflow/examples/tutorials/mnist/mnist_softmax.py
+++ b/tensorflow/examples/tutorials/mnist/mnist_softmax.py
@@ -46,7 +46,7 @@ def main(_):
 
   # The raw formulation of cross-entropy,
   #
-  #   tf.reduce_mean(-tf.reduce_sum(y_ * tf.log(tf.softmax(y)),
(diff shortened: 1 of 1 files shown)


littlekfc:
I signed it!

vrv:
Thanks for the PR @littlekfc.  I think the existing formulation is correct.  See http://ufldl.stanford.edu/wiki/index.php/Softmax_Regression#Cost_Function for some more details.

In this case, y_ is a matrix of [batch_size, 10] where each row is a one-hot vector indicating the class labels 1-10.  Each one-hot vector is the indicator variable vector for the sample.

So when computing the log-likelihood, we apply the softmax function on the logits, and then take a log of the output of the softmax, and then multiply that (batch-wise) with the one-hot vector indicating the label.

As far as I can tell, this equation is correct, though I admit it can be a bit confusing.  Definitely let me know if we're wrong after looking into this more!

PR #5397, opened 2016-11-04
Add environment variable to configure usage of default python library path

andreas-repp:
Some changes to `util/python_config.sh` since release version r0.10 introduced the problem that the user is asked to choose the python library path during `./configure`. The problem with this is that automated scripts / builds have no option to choose the default python library path. 

In PR #5135, an environment variable has been introduced to configure the python library path via an environment variable to get around this problem. However, it leaves the script using `./configure` with the task to find a correct python library path. In many scenarios this cannot be hard coded and determining it with code means to replicate the logic that is already in the script.

Therefore this PR adds an additional environment variable named `USE_DEFAULT_PYTHON_LIB_PATH`. Like the name says, if this variable is set to `1`, the default python library path is used without user interaction.

### DIFF
--- a/util/python/python_config.sh
+++ b/util/python/python_config.sh
@@ -120,20 +120,28 @@ function setup_python {
     IFS=','
     python_lib_path=($(python_path))
     unset IFS
-    echo "Found possible Python library paths:"
-    for x in "${python_lib_path[@]}"; do
-      echo "  $x"
-    done
-    set -- "${python_lib_path[@]}"
-    echo "Please input the desired Python library path to use.  Default is ["$1"]"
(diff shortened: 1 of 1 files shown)


villasv:
It is technically less general, but if hardcoding is really not always possible (I wonder when) I think this is good too. In the end, I can't imagine why someone would want to automate this but *not* want to use the default lib path... so I see no problem here.

vrv:

caisq (approved):

gunan:
not sure where the failure is coming from.
Jenkins, test this please.

PR #5405, opened 2016-11-04
Ignore tools/git/gen and fix a bash error in python_config.sh

girving:
https://github.com/tensorflow/tensorflow/commit/b4798133adfe8fcd9d138ddb86d5b3aa0c4901e9 only works if the new feature is used; otherwise it prints an obscure bash error.


vrv:
python 3 CI failure is known flaky test.  If the CPU test passes ,I'll merge.

PR #5418, opened 2016-11-05
Pylint `disable` fix

z-a-f:
I have added `# pylint: enable=wildcard-import` wherever I found an unclosed `disable`, but only for the `wildcard-import`. There are many more for other parameter, including the custom definitions (`g-...` parameters).


caisq:
@zafartahirov, can you please squash your six commits into a single one? Thanks.

zafartahirov:
@caisq I thought it is automatically squashed once it is merged, according to comment by @ilblackdragon [here](https://github.com/tensorflow/tensorflow/pull/5315#pullrequestreview-6716529)
I can squash it if it is the requirement

vrv:
@caisq we can squash for our users, it's better for PR analysis to have them not squash commits.

tensorflow/contrib/factorization/python/ops/factorization_ops.py
```diff
@@ -745,3 +745,4 @@ def _process_input_helper(self, update_row_factors,
                                 update_indices,
                                 new_left_values,
                                 sharding_func))
+# pylint: enable=wildcard-import,undefined-variable
```
vrv: Thanks for making these changes!  It would be nice to try to keep these re-enables as close to where they were disabled as possible.  So perhaps move this up near the imports, and preferably just after the last wildcard import?  

Same with some of the other files.

zafartahirov:
@vrv done -- Please, note that reenabling the `pylint enable=...` might cause the pylint checks to fail if the `disable` line was suppressing the fail before

tensorflow/contrib/bayesflow/__init__.py
```diff
@@ -29,3 +29,4 @@
 from tensorflow.contrib.bayesflow.python.ops import stochastic_tensor
 from tensorflow.contrib.bayesflow.python.ops import stochastic_variables
 from tensorflow.contrib.bayesflow.python.ops import variational_inference
+# pylint: enable=unused-import,wildcard-import,line-too-long
```
vrv: i think wildcards aren't used in this one, so it removable

vrv:

(testing pylint)

zafartahirov:
There is a merge conflict for `tensorflow/contrib/framework/python/framework/__init__.py`, I am not sure why it is a conflict -- it is a single line addition in the end of the file

vrv:
I'm not sure why it's a conflict, but if you rebase your changes on HEAD it should go away.

vrv:
(after you rebase I can kick off tests and merge.  Thanks for these cleanups!)

zafartahirov:
Done, I think the conflicts are resolved -- not sure what caused them. There are some parameters left that are not closed -- I was only explicitly searching for `wildcard-import` ones.

vrv:

vrv (approved):

PR #5421, opened 2016-11-06
fix windows gpu build

guschmue:


icoxfog417:
This fix helps me! Thank you.

vrv:

vrv (approved):
Maybe add a comment to line 109 explaining why this is being done?

guschmue:
Wonder if a comment makes sense for the T(2) thing. It is very common in this file and was only missing in the 2 places I changed.

I looked at the CI failure: the file must be read only for some reason. I remember seeing this 2 month back but not since. Never found out the root cause.
22:41:30          C:/tf_jenkins/home/workspace/tensorflow-pull-requests-windows-cmake/tensorflow/contrib/cmake/build/tensorflow/core/lib/core/error_codes.pb.cc: Permission denied
22:41:30     19>C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(171,5): error MSB6006:

mrry:

benoitsteiner (approved):

PR #5424, opened 2016-11-06
added full relative path to compile_ios_protobuf

codingyourlife:
The command before and after this command have the full path. That was confusing for me and is a mistake if you ask me.


petewarden (approved):
Thanks for catching this, it is a good fix, thanks!

PR #5431, opened 2016-11-06
fix path error for libcudnn.x

atom2ueki:
fix Invalid path to cuDNN  toolkit. Neither of the following two files can be found when using cuda configure

### DIFF
--- a/configure
+++ b/configure
@@ -278,8 +278,8 @@ while true; do
     CUDA_DNN_LIB_PATH="lib64/libcudnn.so${TF_CUDNN_EXT}"
     CUDA_DNN_LIB_ALT_PATH="libcudnn.so${TF_CUDNN_EXT}"
   elif [ "$OSNAME" == "Darwin" ]; then
-    CUDA_DNN_LIB_PATH="lib/libcudnn${TF_CUDNN_EXT}"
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -278,8 +278,8 @@ while true; do
     CUDA_DNN_LIB_PATH="lib64/libcudnn.so${TF_CUDNN_EXT}"
     CUDA_DNN_LIB_ALT_PATH="libcudnn.so${TF_CUDNN_EXT}"
   elif [ "$OSNAME" == "Darwin" ]; then
-    CUDA_DNN_LIB_PATH="lib/libcudnn${TF_CUDNN_EXT}"
-    CUDA_DNN_LIB_ALT_PATH="libcudnn${TF_CUDNN_EXT}"
+    CUDA_DNN_LIB_PATH="lib/libcudnn${TF_CUDNN_EXT}.dylib"
+    CUDA_DNN_LIB_ALT_PATH="libcudnn${TF_CUDNN_EXT}.dylib"
```
vrv: See lines 248 and 269 -- this should already be done above.  If not, can you figure out why it's not working for you?  I believe it is working for others...

vrv:
(Closing since we can't accept PRs without CLA signed, please sign if you would like this PR to go through, but first check that it is needed :)

PR #5433, opened 2016-11-06
Fix misleading SaveDef V2 format warning

raingo:
Adding a line: **Notice: the V2 format output different set of files**

The original warning made me think I can switch simply by adding the `write_graph` argument, but it turns out the outputs of `saver.save(sess, "model.ckpt")` are also different with the new format.

Originally, it writes to the file `model.ckpt`, but now it writes to `model.ckpt.index` and `model.ckpt.data.*`, which breaks the logic.

### DIFF
--- a/tensorflow/python/training/saver.py
+++ b/tensorflow/python/training/saver.py
@@ -1289,6 +1289,7 @@ class Saver(object):
       logging.warning("Consider switching to the more efficient V2 format now:")
       logging.warning("   `tf.train.Saver(write_version=tf.train.SaverDef.V2)`")
       logging.warning("to prevent breakage.")
+      logging.warning("Notice: the V2 format output different set of files")
       logging.warning("*******************************************************")
 
     if os.path.split(latest_filename)[0]:


concretevitamin:
Hey @raingo: thanks for your PR.  I don't think the deprecation warning is the best place to convey the technical details of the V2 format.  I'll think about what's a more appropriate place to document the different naming scheme.

grooms1122:
This error can occur when you link to a cell that has one style of formatting applied, but the destination cell is another format. For instance, cell A1 is formatted [as](https://loginzilla.com/) currency, but a linked cell with =A1 is formatted as a date. This behavior is limited to linked cells and [data types](https://support.microsoft.com/en-us/office/excel-data-types-stocks-and-geography-61a33056-9935-484f-8ac8-f1a89e210877).
If you select the Update format option, Excel will automatically apply the source cell's format for you.

You can change this setting in File > Options > Formulas > Misleading number formats. 
If you're using a Mac, the setting is under Excel > Preferences > Error Checking > Misleading number formats.

Note: The Misleading number format error can be evaluated in VBA (Visual Basic for Applications) through the Application.ErrorCheckingOptions.MisleadingNumberFormats property, which will return a Boolean (True/False) value.

PR #5442, opened 2016-11-07
phrase suggestion in doc

leonardgithub:
I am not very sure, but "as well" is adverb, "as well as" is conjunction or preposition. The "as well as" is used for the two nouns: the "total loss" and the "speed".


vrv:
Thanks!

PR #5456, opened 2016-11-07
[Windows/CMake] Remove dependency on zlib.dll.

### LINKED ISSUE #5275: [Windows] Cannot import _pywrap_tensorflow when zlib.dll is not in PATH
### What related GitHub issues or StackOverflow threads have you found by searching the web for your problem?

[#1013](https://github.com/tensorflow/tensorflow/issues/1013)
etc...
### Environment info

Operating System:

Windows 10
Installed version of CUDA and cuDNN: 
No CUDA ("CPU support only")

Pip wheel was self-build. Here's the output of `python -c "import tensorflow; print(tensorflow.__version__)"`: [in attatchment](https://github.com/tensorflow/tensorflow/files/559847/output.txt)

If installed from source, provide 
git-rev output - 3737ac321e67410bf061257d5f644eae8abbf79b
No bazel (is it required?)
### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)

`import tensorflow`
### What other attempted solutions have you tried?

Install with `python setup.py install` and generally avoid any tedious building/making but with no success.

(If logs are large, please upload as attachment or provide link).
[output.txt](https://github.com/tensorflow/tensorflow/files/559849/output.txt)

mrry:
Instead, use the static `zlibstatic.lib`, as originally intended.

Fixes #5275.

### DIFF
--- a/tensorflow/contrib/cmake/external/zlib.cmake
+++ b/tensorflow/contrib/cmake/external/zlib.cmake
@@ -8,7 +8,7 @@ set(ZLIB_TAG 50893291621658f355bc5b4d450a8d06a563053d)
 
 if(WIN32)
   set(zlib_STATIC_LIBRARIES
-      ${CMAKE_CURRENT_BINARY_DIR}/zlib/install/lib/zlib.lib)
+      ${CMAKE_CURRENT_BINARY_DIR}/zlib/install/lib/zlibstatic.lib)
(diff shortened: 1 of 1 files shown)


mrry:

vrv (approved):

Issue #5147, opened 2016-10-23
Add 'dynamic_pad' flag to `shuffle_batch` as implemented in `batch`


TimZaman:
The `tf.train.batch()` function had a nice `dynamic_pad` flag:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/input.py#L567
Is there anything holding us batch from implementing it in `tf.train.shuffle_batch()`:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/input.py#L778

asimshankar:
@ebrevdo : What do you think?

TimZaman:
Alternatively, is there a reason shuffle isnt just a flag for 'batch'?

ebrevdo:
This is nontrivial because the c++ logic for dynamic_pad is implemented by adding a new queue type (PaddingFIFOQueue) in addition to the standard FIFOQueue.  The same amount of work should be required to extend the shuttle queue c++ class.  Contributions welcome.

cbockman:
Is there a recommended way to handle this right now, strictly within tensorflow?  Or is the answer to wait for https://github.com/tensorflow/tensorflow/pull/5463 to get completed.

I.e., to use tf.train.shuffle_batch to get inputs--e.g., sentences--of varying lengths, and then pad them (as if dynamic_pad=True was passed)?  I'm unclear if the dynamic padding operation can be elegantly implemented from within TF.

standy66:
@cbockman, I use the following workaround:

```
dtypes = list(map(lambda x: x.dtype, inputs))
shapes = list(map(lambda x: x.get_shape(), inputs))
queue = tf.RandomShuffleQueue(CAPACITY, MIN_AFTER_DEQUEUE, dtypes)
enqueue_op = queue.enqueue(inputs)
qr = tf.train.QueueRunner(queue, [enqueue_op] * NUM_THREADS)
tf.add_to_collection(tf.GraphKeys.QUEUE_RUNNERS, qr)
inputs = queue.dequeue()
for tensor, shape in zip(inputs, shapes):
    tensor.set_shape(shape)
```

Don't forget to start queue runners. If you use SparseTensors there is some extra work of packing and unpacking them into a list of plain tensors.

cbockman:
Hi @standy66 , thanks, I may not be fully following this, but does this work when using TFRecords format w/ very large data sets which are loaded incrementally?

[...]

PR #5463, opened 2016-11-07
Add dynamic_pad to shuffle_batch

SriramRamesh:
Add 'dynamic_pad' flag to `shuffle_batch` as implemented in `batch`
Added the c++ implementation of padding_random_shuffle_queue 
#5147

### DIFF
--- a/tensorflow/core/graph/graph_constructor.cc
+++ b/tensorflow/core/graph/graph_constructor.cc
@@ -358,7 +358,7 @@ Status GraphConstructor::ValidateShape(Node* node) {
       const string& op = node->def().op();
       const std::vector<string> whitelist = {
           // To be removed after 2017/03/08.
-          "RandomShuffleQueue", "PaddingFIFOQueue", "FIFOQueue",
+          "PaddingRandomShuffleQueue","RandomShuffleQueue", "PaddingFIFOQueue", "FIFOQueue",
           "PriorityQueue", "QueueSize", "Stack", "Barrier", "BarrierReadySize",
           "BarrierIncompleteSize", "HashTable", "MutableHashTable",
           "MutableHashTableOfTensors", "Mutex", "CuckooTable", "IndexTable",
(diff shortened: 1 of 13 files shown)


ebrevdo:
Derek, do we want the added complexity of a padded shuffle queue?  PTAL.

mrry:
According to #5147, it's a contribution we'd welcome. I assume there's a use case?

(I'm not thrilled by adding another queue implementation, but it's compatible with the current way things are implemented. It looks like there's a lot of repeated logic in the new shuffle queue... perhaps the padding and shuffling could be implemented orthogonally so there isn't as much repetitive code?)

ebrevdo:
+1 Sriram, is there a way to refactor the two queues in question (shuffle + paddingshuffle) to use one common base set of functions for shuffling?

SriramRamesh:
RandomShuffleQueue can take dynamic_pad as a default argument and then implement padding if it is true

ebrevdo:
@mrry are you ok merging the logic for dynamic_padding and w/o into a single kernel impl?  if so, we could do the same for PaddingFIFOQueue and regular FIFOQueue.

mrry:
@ebrevdo: I don't think we want a single kernel (at the very least, we need separate kernels for backwards compatibility, but I would also prefer it aesthetically)... but the aim is to cut down on repeated code: so we might want to refactor the queue implementation so that (e.g.) the padding (output preparation) logic is separate from the coordination and the element selection code.

ebrevdo:
Sriram, want to try to factor the c++ implementations of the two ops to
share a common codebase for everything but the padded copies?

SriramRamesh:
Sorry,I didn't get you @ebrevdo 
Should i create a common base for PaddingRandomShuffleQueue and PaddingFIFOQueue?

ebrevdo:
This seems reasonable, yes.

danmane:
@SriramRamesh Please also rebase when you make the changes ebrevdo suggested.

martinwicke:
Friendly ping. Can you rebase?

PR #5471, opened 2016-11-08
Remove a premature checkpoint existence test in Saver.restore().

yifeif:
The correct way to perform such a check is inside the Op kernel.
Change: 138212174

(cherry picked from commit a2f0f07f59b318d255ba86c8ab3486466dece5a8)

### DIFF
--- a/tensorflow/contrib/slim/python/slim/evaluation_test.py
+++ b/tensorflow/contrib/slim/python/slim/evaluation_test.py
@@ -26,6 +26,7 @@ import time
 import numpy as np
 import tensorflow as tf
 
+from tensorflow.python.framework import errors
 from tensorflow.python.platform import flags
 from tensorflow.python.platform import gfile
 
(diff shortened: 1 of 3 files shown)


martinwicke (approved):

martinwicke:
Well, this has not fixed it. I do not know what's wrong. @sguada, any idea?

concretevitamin:
In `...slim/learning_test.py` ~Line 197 and other places, try replacing
```
    self._logdir = os.path.join(self.get_temp_dir(), 'tmp_bnlogs/')
```
with a process-unique tmp dir.  For instance `... + uuid.uuid4().hex`.

gunan:
@concretevitamin Actually, that also won't work.
For example, assume 2 mac test runs for 2 different PRs.
If these end up running on the same machine, they can end up with the same paths.
so we need something more like this:
```
self._logdir = tempfile.mkdtemp(prefix=os.path.join(self.get_temp_dir(), 'tmp_bnlogs/'))
```

yifeif:
We are removing this test to unblock the release. Meanwhile @cshallue is working on a fix. Closing this PR.

PR #5472, opened 2016-11-08
Fix a typo in export meta graph sample code.

nagachika:
The sample code in How-tos "Exporting and Importing a MetaGraph" cause NameError: name 'saver0_ckpt' is not defined.
I think it should be `'my-save-dir/my-model-10000'` to be restored in the subsequent sample code.

I've signed the CLA before for other Google's project (https://github.com/GoogleCloudPlatform/DataflowJavaSDK). Do I have to make another agreement?


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

PR #5475, opened 2016-11-08
Wait for jvm to shutdown completely before remove bazel install directory

meteorcloudy:
@gunan


gunan (approved):

gunan:
I am not really happy with the amount of bazel magic we are calling here.
Ultimately, are we going to be able to simply run the following in this script?
```
$ echo "" | configure
$ bazel test tensorflow/...
```

meteorcloudy:
@gunan Yes, I believe so, but of course the tests still need to be fixed first.
And the bazel magics are only for working around `bazel clean --expunge` problem, they can go away after we fix it. https://github.com/bazelbuild/bazel/issues/1586

gunan:
OK, sounds good.
The test failure is unrelated, therefore merging.

PR #5494, opened 2016-11-09
Refactor cuda_configure to better support for different CUDA installations

davidzchen:
This change implements cuda and cudnn version autodetection via running
nvcc --version and reading cudnn.h. As a result, TF_CUDA_VERSION and
TF_CUDNN_VERSION no longer have to be set. If TF_CUDA_VERSION and
TF_CUDNN_VERSION are set but do not match the detected versions,
cuda_configure will fail.

After the cuda and cudnn versions are detected, cuda_configure then
attempts to find the needed cuda and cudnn libraries in a set of known
installation directories.

### DIFF
--- a/third_party/gpus/cuda_configure.bzl
+++ b/third_party/gpus/cuda_configure.bzl
@@ -120,12 +120,11 @@ def _enable_cuda(repository_ctx):
   return False
 
 
-def _cuda_toolkit_path(repository_ctx, cuda_version):
+def _cuda_toolkit_path(repository_ctx):
   """Finds the cuda toolkit directory.
 
   Args:
     repository_ctx: The repository context.
-    cuda_version: The cuda toolkit version.
 
   Returns:
     A speculative real path of the cuda toolkit install directory.
@@ -135,16 +134,6 @@ def _cuda_toolkit_path(repository_ctx, cuda_version):
     cuda_toolkit_path = repository_ctx.os.environ[_CUDA_TOOLKIT_PATH].strip()
   if not repository_ctx.path(cuda_toolkit_path).exists:
     auto_configure_fail("Cannot find cuda toolkit path.")
-
-  if cuda_version:
-    # Handle typical configuration where the real path is
-    # <basedir>/cuda-<version> and the provided path is <basedir>/cuda.
-    version_suffixed = "%s-%s" % (cuda_toolkit_path, cuda_version)
-    if repository_ctx.path(version_suffixed).exists:
-      cuda_toolkit_path = version_suffixed
-  # Returns the non-versioned path if cuda version is not provided or if the
-  # installation does not use a cuda- directory, such as on ArchLinux where
-  # CUDA installs directly to /opt/cuda.
   return str(repository_ctx.path(cuda_toolkit_path).realpath)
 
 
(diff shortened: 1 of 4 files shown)


davidzchen:
FYI there is still a bit of manual testing I am doing, but I'm opening this to run it through the CI tests.

+cc @damienmg

third_party/gpus/cuda_configure.bzl
```diff
@@ -158,20 +147,63 @@ def _cudnn_install_basedir(repository_ctx):
...
+    if len(basename) == len("lib%s.dylib" % lib_name):
+      auto_configure_fail("Could not find versioned %s lib. Found: %s" %
+                          (lib_name, lib_path))
+    full_version = (
+        basename.replace("lib%s" % lib_name, "").replace(".dylib", ""))
+  elif cpu_value == "Windows":
+    if len(basename) == len("%s.dll" % lib_name):
+      auto_configure_fail("Could not find versioned %s lib. Found: %s" %
+                          (lib_name, lib_path))
+    full_version = basename.replace(lib_name, "").replace(".dll", "")
```
damienmg: Seems like this can be factorized:

```
EXTENSIONS = {"Linux": "so", "Darwin": "dylib", "Windows": "dll"}
PREFIXES = {"Linux": "lib", "Darwin": "lib", "Windows": ""}
if not cpu_value in EXTENSIONS.keys():
 auto_configure_fail("Not supported CPU value %s" % cpu_value)
expected_basename = "%s%s" % (PREFIXES[cpu_value], lib_name)
extension = EXTENSIONS[cpu_value]
expected_name = "%s.%s" % (expected_basename, extension)
if len(basename) == len(expected_name):
    auto_configure_fail("Could not find versioned %s lib. Found: %s" % (lib_name, lib_path))
full_version = basename.replace(expected_basename, "").replace("." + extension, "")

File: third_party/gpus/cuda_configure.bzl [hunk 2]
```diff
@@ -158,20 +147,63 @@ def _cudnn_install_basedir(repository_ctx):
   return cudnn_install_path
 
 
-def _cuda_version(repository_ctx):
+def _lib_version(lib_name, repository_ctx, cpu_value, base_dir):
+  lib = _find_cuda_lib(lib_name, repository_ctx, cpu_value, base_dir, "")
+  lib_path = lib.path
+  basename = repository_ctx.path(lib_path).basename
+  if cpu_value == "Linux":
+    if len(basename) == len("lib%s.so" % lib_name):
```
damienmg: What is that check? why not do a string comparison directly?

third_party/gpus/cuda_configure.bzl
```diff
@@ -200,75 +232,96 @@ def _cpu_value(repository_ctx):
...
-        cuda_lib_path = "lib64",
-        cuda_rt_lib = "lib64/libcudart.so%s" % cuda_ext,
-        cuda_rt_lib_static = "lib64/libcudart_static.a",
-        cuda_blas_lib = "lib64/libcublas.so%s" % cuda_ext,
-        cuda_dnn_lib = "lib64/libcudnn.so%s" % cudnn_ext,
-        cuda_dnn_lib_alt = "libcudnn.so%s" % cudnn_ext,
-        cuda_rand_lib = "lib64/libcurand.so%s" % cuda_ext,
-        cuda_fft_lib = "lib64/libcufft.so%s" % cuda_ext,
-        cuda_cupti_lib = "extras/CUPTI/lib64/libcupti.so%s" % cuda_ext)
+    if static:
```
damienmg: Same as before, using a map for extension/prefixes it would make the code a bit cleaner IMO

damienmg (commented):
Just a few clean-up comments.

jart (approved):
Thanks for doing this. You have my approval. Please address Damien's comments.

third_party/gpus/cuda_configure.bzl
davidzchen: Premature optimization. :)

Done.

third_party/gpus/cuda_configure.bzl
davidzchen: Done.

meteorcloudy: Hey, David. I am currently fixing the GPU build on Windows, I realized that the extension should be .lib instead of .dll.
Because on Windows, there is a corresponding .lib file for each .dll, we need only need the .lib file for compiling and the dll is only needed during runtime.

third_party/gpus/cuda_configure.bzl
davidzchen: Done, though Linux would still need a special case since it follows a different naming convention wrt where the version is located in the name.

third_party/gpus/cuda_configure.bzl
```diff
@@ -200,75 +287,143 @@ def _cpu_value(repository_ctx):
...
-        cuda_rt_lib_static = "lib/cudart_static.lib",
-        cuda_blas_lib = "lib/cublas%s.dll" % cuda_ext,
-        cuda_dnn_lib = "lib/cudnn%s.dll" % cudnn_ext,
-        cuda_dnn_lib_alt = "cudnn%s.dll" % cudnn_ext,
-        cuda_rand_lib = "lib/curand%s.dll" % cuda_ext,
-        cuda_fft_lib = "lib/cufft%s.dll" % cuda_ext,
-        cuda_cupti_lib = "extras/CUPTI/lib/cupti%s.dll" % cuda_ext)
-  else:
-    auto_configure_fail("Not supported CPU value %s" % cpu_value)
+    if static:
```
damienmg: To go further in the factorisation to diminish the number of if statement you could use templates in the map.

Not sure it would give a nicer code though.

damienmg (approved):

davidzchen:
Are the windows cmake tests broken currently? Most of the previous runs seem to be failing: https://ci.tensorflow.org/job/tensorflow-pr-win-cmake-py/

third_party/gpus/cuda_configure.bzl
davidzchen: We could add templates to the map, but there is still the corner case for Linux where the unversioned library does not have the period at the end (`libcudart.so.7.5` vs `libcudart.so`), which the other platforms do not have. Let's leave this as is for now.

davidzchen:
Rebased and resolved merge conflicts.

@jart - Any more feedback on this before it can be merged?

davidzchen:
@meteorcloudy - Is there a TF Windows test that builds with GPU or an instance that I can use to test this? If not, would it be possible for you to test this PR on your Windows machine? I currently do not have access to a Windows machine.

meteorcloudy:
@davidzchen Yes, I can test this, and everything looks correct to me except the library extension for Windows should be ".lib" instead of ".dll".

davidzchen:
@meteorcloudy Thanks! Fixed, rebased, and pushed. That's interesting. Out of curiosity, is that a convention that other libraries on Windows are starting to follow now?

third_party/gpus/cuda_configure.bzl
```diff
@@ -158,20 +147,110 @@ def _cudnn_install_basedir(repository_ctx):
...
+  """
+  lib = _find_cuda_lib(lib_name, repository_ctx, cpu_value, base_dir)
+  lib_path = lib.path
+  basename = repository_ctx.path(lib_path).basename
+
+  if not cpu_value in _EXTENSIONS.keys():
+    auto_configure_fail("Not supported CPU value %s" % cpu_value)
+  expected_basename = "%s%s" % (_PREFIXES[cpu_value], lib_name)
+  extension = _EXTENSIONS[cpu_value]
+  if basename == "%s.%s" % (expected_basename, extension):
```
davidzchen: This is when cuda_configure attempts to detect the version of cuda by resolving the symlinks from the non-versioned libraries to the versioned ones (as is the case on macos and linux, i.e. `libcudart.so.7.5` and `libcudart7.5.dylib`).

Does Windows have versioned cuda and cudnn libraries?

meteorcloudy: The situation on Windows is a little bit complicated.
But basically, every `dll` file has a corresponding `lib` file.
During linking time, we only need to link against the `lib` files, as long as the "dll" files are in PATH, the executable will run.
And the `lib` files are non-versioned, they are just like `cuda.lib`, `cudann.lib` without version number. But the `dll` files are versioned, they are like `cudart32_80.dll`, `cudart64_80.dll`, `cudann64_5.dll`. When loading libraries in [dso_loader.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/stream_executor/dso_loader.cc#L63), we are actually loading the dlls, so `GetCudaVersion()` and `GetCudnnVersion()` must return correct version number.
But the `lib` files specified in BUILD files have to be non-versioned.

davidzchen: Interesting. Thanks for the explanation!

I have reimplemented the cuda and cudnn detection code to read the output of `nvcc --version` and `cudnn.h` to detect the versions instead and special-cased the names of the windows libraries.

Can you try running it again?

meteorcloudy: Thanks David! The previous error is gone. Just two small problems.
1. When finding nvcc_path in [`_cuda_version`](https://github.com/tensorflow/tensorflow/pull/5494/files#diff-27368da6eb4c2514a27a4d9733bc9b57R164), the nvcc executable has to end with `.exe`, otherwise skylark can't find it.
2. The version numbers returned by parsing output of `nvcc --version` are `8.0` for cuda and `5` for cudnn, they don't have the `64_` prefix(We only support 64 bit on Windows, right?). I think we need to strip `.` in it and add `64_` as prefix, so that  `cudart64_80.dll`, `cudnn64_5.dll` can be loaded at runtime.

gunan:
The windows failure that shows up is a flake that is already fixed.
Jenkins, test this please.

meteorcloudy:
@davidzchen Hi David, can you take a look at the two Windows problems I mentioned? If you are worried about no way to test the change, I can also send another PR to fix them after this is merged.

danmane:

third_party/gpus/cuda_configure.bzl
```diff
@@ -158,20 +147,96 @@ def _cudnn_install_basedir(repository_ctx):
...
+
+  Args:
+    repository_ctx: The repository context.
+    cuda_toolkit_path: The CUDA install directory.
+
+  Returns:
+    String containing the version of CUDA.
+  """
+  # Run nvcc --version and find the line containing the CUDA version.
+  nvcc_path = repository_ctx.path("%s/bin/nvcc" % cuda_toolkit_path)
```
davidzchen: Done.

davidzchen:
Sorry, I just got back from the Thanksgiving holiday. I have fixed the `nvcc.exe` issue. I am a bit unclear on the `.dll` issue. Where are the build rules that reference the `.dll`s?

meteorcloudy:
@davidzchen No problem! I can send a fix later, thanks for giving these fixes! :)

davidzchen:
@meteorcloudy Sounds great. Thanks, Yun!

@jart @gunan All remaining review comments have been addressed and tests are passing. Can we go ahead and merge this?

Issue #5501, opened 2016-11-09
ExtractImagePatches Gradient doesn't work if batch size is unknown


liuyipei:
See #3672 comment by @mohamedadaly
His fix is to add 
```
  batch_size = array_ops.shape(op.inputs[0])[0]
```
after this:
```
  batch_size, rows_in, cols_in, channels = [
    dim.value for dim in op.inputs[0].get_shape()
  ]
```
in `tensorflow/python/ops/array_grad.py`

PR #5502, opened 2016-11-09
[Minor] ExtractImagePatches grad fails when batch size is unknown

liuyipei:
Mentioned in
#5501
#3672
Solution was by @mohamedadaly

### DIFF
--- a/tensorflow/python/ops/array_grad.py
+++ b/tensorflow/python/ops/array_grad.py
@@ -504,6 +504,8 @@ def _ExtractImagePatchesGrad(op, grad):
   batch_size, rows_in, cols_in, channels = [
     dim.value for dim in op.inputs[0].get_shape()
   ]
+  batch_size = array_ops.shape(op.inputs[0])[0]
+
   _, rows_out, cols_out, _ = [
     dim.value for dim in op.outputs[0].get_shape()
   ]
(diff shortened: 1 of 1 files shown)


liuyipei:
I signed it!

liuyipei:
I signed it!

tensorflow/python/ops/array_grad.py
```diff
@@ -504,6 +504,8 @@ def _ExtractImagePatchesGrad(op, grad):
   batch_size, rows_in, cols_in, channels = [
     dim.value for dim in op.inputs[0].get_shape()
   ]
+  batch_size = array_ops.shape(op.inputs[0])[0]
+
```
benoitsteiner: To be consistent, you should also use array_ops.shape() to extract rows_in, cols_in, and channels

benoitsteiner (changes requested):

liuyipei:
@benoitsteiner 
I changed `channels` as per your request;  sadly, there is a distinction for the others. 
`rows_out`, `rows_in`, and etc are actually used as scalar underneath, where they do scalar-ish things like serving as for-loop constraints, getting passed into `range`, and etc.  So, a significant rewrite would be required if I wanted to let `rows_in` to be a tensor.

benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

PR #5505, opened 2016-11-09
Xmbrst patch 1

xmbrst:
Will replace with an include in an internal change, but for now just removing.


xmbrst:
I did author these commits.

vrv:
Probably want to send a PR from an up to date branch.

PR #5520, opened 2016-11-10
Make sure to uninstall the old pip package and install the newly built

gunan:
one when running windows tests on TF CI.


mrry (approved):
LGTM! Thanks for fixing that :).

gunan:
both failures seem to be flakes.
The consistently failing 4 targets in cmake tests are fixed.
Merging.

PR #5526, opened 2016-11-10
Add support for broadcasting shapes.

tillahoffmann:
This PR adds functionality to `TensorShape` to allow for broadcasting. In particular,

* `TensorShape.broadcast_with(other)` returns the shape resulting from broadcasting a tensor of the given shape with a tensor of shape `other`. It raises a `ValueError` if the shapes are not broadcastable.
* `TensorShape.is_broadcastable_with(other)` determines whether two shapes are broadcastable
* `TensorShape.assert_is_broadcastable_with(other)` asserts that two shapes are broadcastable

### DIFF
--- a/tensorflow/python/framework/tensor_shape.py
+++ b/tensorflow/python/framework/tensor_shape.py
@@ -18,6 +18,8 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import itertools as it
+
 from tensorflow.core.framework import tensor_shape_pb2
 from tensorflow.python.util import compat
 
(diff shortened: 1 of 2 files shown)


vrv:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/common_shapes.py#L501

Do you know why this is different from that?

tillahoffmann:
No, they are in fact identical (up to the implementation).

jchadwickson264:
The term broadcasting describes how NumPy treats arrays with different shapes during arithmetic operations. Subject to certain constraints, the smaller array is “broadcast” across the larger array so that they have compatible shapes. Broadcasting provides a means of vectorizing array operations so that looping occurs in C instead of Python.
 It does this without making needless copies of data and usually leads to efficient algorithm implementations. There are, however, cases where broadcasting is a bad idea because [it](https://x8spedersapk.id/snaptik/) leads to inefficient use of memory that slows computation.
NumPy operations are usually done on pairs of arrays on an element-by-element basis. In the simplest case, the two arrays must have exactly the same shape, as in the following example:

a = np.array([1.0, 2.0, 3.0])

b = np.array([2.0, 2.0, 2.0])

a * b
array([2.,  4.,  6.])

NumPy’s broadcasting rule relaxes this constraint when the arrays’ shapes meet certain constraints. The simplest broadcasting example occurs when an array and a scalar value are combined in an operation:

a = np.array([1.0, 2.0, 3.0])

b = 2.0

a * b
array([2.,  4.,  6.])

The result is equivalent to the previous example where b was an array. We can think of the scalar b being stretched during the arithmetic operation into an array with the same shape as a. The new elements in b, as shown in Figure 1, are simply copies of the original scalar. The stretching analogy is only conceptual. NumPy is smart enough to use the original scalar value without actually making copies so that broadcasting operations are as memory and computationally efficient as possible.

PR #5529, opened 2016-11-10
Branch 138793949

benoitsteiner:


benoitsteiner:
Jenkins, test this please.

terrytangyuan:
@ispirmustafa Is there a reason why we are deprecating monitors? SessionRunHook doesn't offer all functionalities monitors have and it might be hard for users to do implement their own.

terrytangyuan:
cc: @martinwicke @ilblackdragon I understand that most of these changes are more for internal compatibility and flexibility, but is there any roadmap that you can share by any chance?

martinwicke:
The problem with `Monitors` was that they were too specific to the `Estimator` class. We wanted a similar solution for training loops not written as an `Estimator`, so we pulled `Monitor`s further down to attach them more directly to the run calls in a session.

They should be able to do everything the Monitors did (in fact, they should be able to do more). What functionality is missing?

terrytangyuan:
Thanks for the clarification! That actually makes sense. I missed some functionalities that actually exist so yeah they are able to do more than monitors. Perhaps we need to update the tutorials as well.

martinwicke:
yes. A lot of the docs are at least somewhat stale.

PR #5531, opened 2016-11-10
[Windows/CMake] Add support for tf.contrib.

mrry:
This change enables all libraries apart from tf.contrib.tfprof to be used from the CMake-built PIP package.


gunan:
I see a lot of test failures which look like this:

00:15:14.624 AttributeError: module 'tensorflow.python.ops.gen_array_ops' has no attribute '_list_diff'

They do not seem to happen in master branch.
Are we excluding gen_array_ops from cmake build in windows?

mrry:
The previous Jenkins build is showing the same failures: https://ci.tensorflow.org/job/tensorflow-pr-win-cmake-py/32/console

So I don't think it's caused by this CL, but something is messed up. We're blowing away the `cmake_build` directory between runs, right? (Otherwise maybe the Python codegen for `gen_array_ops` isn't rerunning for some reason, and thus we aren't picking up the new op setdiff1d or whatever it's called?)

guschmue:
if it helps: I have a jenkins for cpu+gpu that track master and they are both green all day, including tests.

mrry:

gunan:
Now we also have avgpooling ops change in.
Derek, do we need to rebase?

gunan:
Jenkins, test this please.

mrry:
I think we shouldn't need to rebase, as this touches a disjoint set of files.

gunan:
Now it looks like we are having a problem with gen_array_ops:
```
00:16:42.127     import tensorflow as tf
00:16:42.127   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\__init__.py", line 24, in <module>
00:16:42.127     from tensorflow.python import *
00:16:42.127   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\__init__.py", line 71, in <module>
00:16:42.127     from tensorflow.python.framework.framework_lib import *
00:16:42.127   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\framework\framework_lib.py", line 99, in <module>
00:16:42.127     from tensorflow.python.framework.subscribe import subscribe
00:16:42.127   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\framework\subscribe.py", line 23, in <module>
00:16:42.127     from tensorflow.python.ops import array_ops
00:16:42.127   File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\ops\array_ops.py", line 136, in <module>
00:16:42.127     listdiff.__doc__ = gen_array_ops._list_diff.__doc__ + "\n" + listdiff.__doc__
00:16:42.127 AttributeError: module 'tensorflow.python.ops.gen_array_ops' has no attribute '_list_diff'
```

mrry:

gunan:
reader_op_test is fixed, and there seems to be a flake, so:
Jenkins, test this please.

mrry:

guschmue (approved):
LGTM. 
I also tested this on my box, works fine.
the only question I have if this is good to turn on for all cmake builds or does it need to be under tensorflow_BUILD_CONTRIB.

gunan (approved):

gunan:
For everything else, we run contrib tests together with test for core.
I would like to change that, but for now having them run at the same location seems like the best approach.

PR #5532, opened 2016-11-10
[Windows/CMake] Enable avgpooling_op.cc.

### LINKED ISSUE #5517: Don't support avgpooling op on Windows ?
## Environment info
Operating System: Windows 10, 64bit, Visual Studio 2015

## Issue
I try compile tensorflow label image sample on windows 10 64bit with Visual Studio 2015.
I got error 'AvgPool' op kernel not found message while loading tensorflow_inception_graph.pb graph.
I check tf_core_kernels.cmake and found follow line.

```
if(WIN32)
  file(GLOB_RECURSE tf_core_kernels_windows_exclude_srcs
      # not working on windows yet
      "${tensorflow_source_dir}/tensorflow/core/kernels/depthwise_conv_op.cc"  # Cannot find symbol: tensorflow::LaunchConv2DOp<struct Eigen::ThreadPoolDevice, double>::launch(...).
      "${tensorflow_source_dir}/tensorflow/core/kernels/fact_op.cc"
      "${tensorflow_source_dir}/tensorflow/core/kernels/meta_support.*"
      "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.h"
      "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.cc"
      "${tensorflow_source_dir}/tensorflow/core/kernels/svd*.cc"
      "${tensorflow_source_dir}/tensorflow/core/kernels/avgpooling_op.*"
  )
  list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_windows_exclude_srcs})
endif(WIN32)
```

Is this mean don't support above ops on windows?
How can I running graph that using these ops?

mrry:
Fixes #5517.

### DIFF
--- a/tensorflow/contrib/cmake/tf_core_kernels.cmake
+++ b/tensorflow/contrib/cmake/tf_core_kernels.cmake
@@ -91,7 +91,6 @@ if(WIN32)
       "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.h"
       "${tensorflow_source_dir}/tensorflow/core/kernels/*quantiz*.cc"
       "${tensorflow_source_dir}/tensorflow/core/kernels/svd*.cc"
-      "${tensorflow_source_dir}/tensorflow/core/kernels/avgpooling_op.*"
   )
   list(REMOVE_ITEM tf_core_kernels_srcs ${tf_core_kernels_windows_exclude_srcs})
 endif(WIN32)
--- a/tensorflow/contrib/cmake/tf_tests.cmake
+++ b/tensorflow/contrib/cmake/tf_tests.cmake
@@ -149,11 +149,8 @@ if (tensorflow_BUILD_PYTHON_TESTS)
       # issues related to windows fs
       "${tensorflow_source_dir}/tensorflow/python/kernel_tests/io_ops_test.py"
       # missing kernel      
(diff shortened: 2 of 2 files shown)


gunan:
Looks like the cmake build fails?
is the failure caused by this change?

mrry:
This one looks like it builds successfully but after installing the PIP package, it can't find the `tensorflow` module. Not clear why a C++ change would affect this… could the Jenkins worker be in a bad state? (Looks like the build has been red all day, even with this morning's fix….)

gunan:
hmm, looks like pip install flaked?
Let's retry this.
Jenkins, test this please.

mrry:
Everything but `reader_ops_test.py` passed this time... which is a bit odd since it passes locally for me :(, but it shouldn't be affected by the addition of `avgpooling_op.cc` to the build.

gunan:
That test has been flaking wildly in our jenkins.
We can debug that separately.

guschmue:
Ah, I know why I did not enable avgpooling_op.cc: breaks the gpu build. If I remember correctly a bunch on changes needed to the .cu.cc files. 
I can take care of it.
--
 avgpooling_op.obj : error LNK2019: unresolved external symbol "public: void __cdecl tensorflow::functor::SpatialAvgPooling

PR #5547, opened 2016-11-11
optional libjpeg-turbo support

mkolod:
This is in reference to #4807.

### DIFF
--- a/configure
+++ b/configure
@@ -54,6 +54,7 @@ if is_windows; then
   TF_NEED_HDFS=0
   TF_NEED_CUDA=0
   TF_NEED_OPENCL=0
+  TF_NEED_LIBJPEG_TURBO=0
 fi
 
 while [ "$TF_NEED_GCP" == "" ]; do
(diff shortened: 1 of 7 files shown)


mkolod:
@jart @tensorflower-gardener @keveman BTW, in #4807, I was asked to provide some benchmarks. When using a jpeg-decode-only graph [source here](https://github.com/mkolod/tf_perf_eval/blob/master/io/jpeg/tensorflow/multithreaded_jpeg_decode.py) and limiting processing to 1 thread, I got 1,257 images/sec on a machine with an SSD and an Intel Core i7-5930K CPU @ 3.50GHz with libjpeg-turbo, and 632 images/second with libjpeg. This represents a 1.99x speedup, which is consistent with the expectations based on the [benchmarks](http://www.libjpeg-turbo.org/About/Performance) provided by the libjpeg-turbo team. The range of speed-up factors for x86-64   is 1.89 to 4.31, depending on image characteristics, and my test shows that the speed-up factor is indeed in that range. For other platforms such as ARM v8, the speed-up factors are basically the same as for x86-64 (see [here](http://www.libjpeg-turbo.org/About/Performance)). A very exhaustive spreadsheet with lots of benchmark scenarios can be found [here](http://www.libjpeg-turbo.org/pmwiki/uploads/About/libjpegturbo-1.5.ods). My finding of a 1.99x speedup with the library called from TensorFlow itself should be convincing enough that this constitutes a major improvement.

configure
```diff
@@ -100,6 +101,40 @@ while [ "$TF_NEED_HDFS" == "" ]; do
   esac
 done
 
+while [ "$TF_NEED_LIBJPEG_TURBO" == "" ]; do
```
jart: Is there any reason why we can't just always use libjpeg-turbo and get rid of the other one? Then we don't need more configure script technical debt.

jart (changes requested):
Excellent work. Thank you for making your argument with numbers.

configure
mkolod: @jart The challenge here is that libjpeg-turbo is an assembly-based code base, rather than pure C. This means that the configuration for it is pretty tricky, and the default build for libjpeg is based on autotools (autoconf, automake) as well as nasm. This makes it messy because suddenly for one library substitution, the build grows from Bazel to 3 additional build tools. If the lib substitution is optional, users who don't care about the improved performance can keep building like they always have, just with Bazel. Users who might benefit from the extra performance and don't mind installing extra tools could do so, but I think forcing extra tools on every user would be painful. The other issue here is that as far as I know, doing assembly builds by themselves directly in Bazel would also be non-trivial, hence I'm reusing the build definitions from libjpeg-turbo's code base.

mkolod: @jart Also see [here](https://github.com/tensorflow/tensorflow/issues/4807#issuecomment-255889425) and [here](https://github.com/tensorflow/tensorflow/issues/4807#issuecomment-256783132) regarding the previously stated preference by a Googler that this be made optional.

jart: There might be things @jhseu hadn't considered.

World will not end if we depend on nasm. It's 0.8MB and can be built very easily with Bazel. https://gist.github.com/jart/68b1fd09d48197850981d45e85758936 (Although that BUILD file would need a genrule to create config.h and the copts would need to be tuned for MSVC. We do not need to depend on autotools.) We already download and build others tools from scratch, e.g. [SWIG](https://github.com/tensorflow/tensorflow/blob/master/third_party/swig.BUILD). It adds three seconds to build time on my workstation. That's better IMHO than increasing the cognitive burden on users with configuration questions and action items.

However building libjpeg-turbo inside Bazel is not trivial. We have a build file that does it internally. It's a very complete build file that appears to support x86_64, x86, armv7a, armv8a, Android, GCC, Clang, and MSVC. I might be able to copy this file onto GitHub. Thoughts @gunan and @wicke?

gunan: Any additional dependencies are sources of flakiness, as a mirror ca be down, or slow to respond, etc. So it would be better to keep the dependencies minimal if we can.
In terms of the build file, @martinwicke  or @vrv can comment better on how we can make an internal build file available externally.

gunan:
Jenkins, test this please.

configure
jart: I'm going to be adding a feature to Bazel in the next month or so that will allow us to mirror dependencies. So any flakiness an extra dependency adds would only be temporary.

mkolod: @jart If you do indeed have an internal build file that builds libjpeg-turbo within Bazel, then that would definitely be a preferable way to go. If you could either share that so I could fix this PR, or substitute your solution with a different commit, that would render this entire PR unnecessary. :) Please keep me posted!

martinwicke: I'm happy with either solution. The caveat being that we have to make sure that this works on all platforms we (want to) support (incl. Windows and ARM). I agree depending on nasm isn't a problem assuming it's portable. I also agree that it would be nice to avoid an extra configuration question. I wasn't aware we had a complete bazel build. jart@ can you add that file to tensorflow so mkolod@ can use it?

jart:
We're grateful that you took the time to send us this pull request @mkolod. I've just written up a change internally that upgrades TensorFlow to libjpeg-turbo, using the secret build knowledge other Googlers have figured out before me. When my change gets exported to GitHub (in probably a week tops) #4807 will be updated automatically.

PR #5558, opened 2016-11-11
Add string tensor tags to new summary interface

thuyen:
Old/Deprecated summary ops can accept `string tensors` for `tags`. The new summary interface only allows `strings` for `names` (subsequently used as tags). This pull adds `prefix` option, which can be a `string tensor`. 

This is important because sometimes we want to define the tags symbolically based on some conditions of input placeholders.

### DIFF
--- a/tensorflow/python/summary/summary.py
+++ b/tensorflow/python/summary/summary.py
@@ -104,7 +104,9 @@ def scalar(name, tensor, collections=None, prefix=''):
     tensor: A real numeric Tensor containing a single value.
     collections: Optional list of graph collections keys. The new summary op is
       added to these collections. Defaults to `[GraphKeys.SUMMARIES]`.
-    prefix: Optional string tensor
+    prefix: Optional string tensor. If provided, it will be a prefix to the
+      summary's tag in TensorBoard. This can be used when the tag need to be
+      defined dynamically at run time.
 
   Returns:
     A scalar `Tensor` of type `string`. Which contains a `Summary` protobuf.
@@ -160,7 +162,9 @@ def image(name, tensor, max_outputs=3, collections=None, prefix=''):
     max_outputs: Max number of batch elements to generate images for.
     collections: Optional list of ops.GraphKeys.  The collections to add the
       summary to.  Defaults to [_ops.GraphKeys.SUMMARIES]
-    prefix: Optional string tensor
+    prefix: Optional string tensor. If provided, it will be a prefix to the
+      summary's tag in TensorBoard. This can be used when the tag need to be
+      defined dynamically at run time.
 
   Returns:
     A scalar `Tensor` of type `string`. The serialized `Summary` protocol
@@ -196,7 +200,9 @@ def histogram(name, values, collections=None, prefix=''):
       build the histogram.
     collections: Optional list of graph collections keys. The new summary op is
       added to these collections. Defaults to `[GraphKeys.SUMMARIES]`.
-    prefix: Optional string tensor
+    prefix: Optional string tensor. If provided, it will be a prefix to the
+      summary's tag in TensorBoard. This can be used when the tag need to be
+      defined dynamically at run time.
 
   Returns:
     A scalar `Tensor` of type `string`. The serialized `Summary` protocol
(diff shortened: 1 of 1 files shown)


thuyen:
@danmane @drpngx Can one of the admins have a look at this pull request?

drpngx:
@danmane knows best.

danmane:
Is there a need for the "prefix" argument? Why not just have the summary wrapper accept a string tensor as name, and convert it into a string?

Adding prefix as another arg seems orthogonal to the main goal of the pr

thuyen:
@danmane Could you tell me the function for `string_tensor to string` conversion. I couldn't find it in current API. If we have that function then we don't need this PR at all.

thuyen:
I am still not sure if such function exists since `tensor` value would be determined dynamically at run time while `string` needs static literals.

martinwicke:
What would the prefix be used for? Can't you achieve the same by wrapping the call to the summary in a name_scope?

thuyen:
`name_scope` couldn't be a `string tensor`. 
With the old interface, one can create a single graph for both training and validation (like the following). In the new interface, it's only possible with some new argument `prefix` (a `tensor`) because the `name` argument can only be `string`. An example:
```
is_training = tf.placeholder(tf.bool)
inputs, labels = data(is_training)
outputs = model(inputs, is_training)
loss = tf.reduce_mean(tf.square(output-labels))
train_op = ....

prefix = tf.cond(is_training, lambda: tf.constant('train/'), lambda: tf.constant('valid/'))
tf.summary.scalar('loss', loss, prefix=prefix)
summary_op = tf.summary.merge_all()

# Each training iteration
sess.run([train_op, summary_op], feed_dict={is_training: True})

# Validation
sess.run(summary_op, feed_dict={is_training: False})
```

foxik:
+1 for implementing some way of using string tensor as summary name.

I frequently used the following scenario, where I want to measure accuracy on train/devel/test data (to appear in different graphs in TensorBoard):
```python
    dataset_name = tf.placeholder(tf.string, [])
    summary = tf.scalar_summary(self.dataset_name+"/accuracy", accuracy)
```
That way I could use the same `summary` with any of the data:
```python
   summary = session.run([summary], {dataset_name:"train" / "devel" / "test"})
```
A `prefix` option would be ideal for me, but generaly any way of using string tensor is. Currently I have to create separate summary operations for every dataset I want to evaluate on.

ppwwyyxx:
+1 for passing an optional string/string tensor (but not prefix) as names. This would also resolve #6150

drpngx:
@dandelionmane is that the right way to go about this?

rmlarsen:
@dandelionmane any advice on how to proceed here?

dandelionmane:
@thuyen Thanks for the pull request, and sorry for the really slow review on my part.

I think this is a fine proposal. I would just like more descriptive docstrings that explain why the new argument should be used.

tensorflow/python/summary/summary.py
```diff
@@ -104,6 +104,7 @@ def scalar(name, tensor, collections=None):
     tensor: A real numeric Tensor containing a single value.
     collections: Optional list of graph collections keys. The new summary op is
       added to these collections. Defaults to `[GraphKeys.SUMMARIES]`.
+    prefix: Optional string tensor
```
dandelionmane: Optional string tensor. If provided, it will be a prefix to the summary's tag in TensorBoard.

teamdandelion (changes requested):
I'd like the docstring for "prefix" to explain how the prefix will be used. Otherwise, LGTM

vrv:
Adding @martinwicke for API review triage, if needed.

Also, a test is warranted here, probably.

dandelionmane:
Please update it for the (tensor_summary)[https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/summary_ops.py#L38] as well. This will be more tricky because the tensor_summary has a different implementation and things are a little inconsistent. 
Please also add tests.

In the long run, I think we may want to have all of the summaries set the node_name part of the summary protobuf according to the kernels' node name, and set the tag according to things like the prefix and the original text the user provided to the summary op (as opposed to the deduplicated, name-scope'd name)

ppwwyyxx:
Why the solution here is prefix instead of just giving the whole string? Giving the whole string would also fix #6150 together

martinwicke:
@dandelionmane, what about @ppwwyyxx's comment?

drpngx:
@dandelionmane any progress?

thuyen:
@dandelionmane `tensor_summary` has different interface, taking only the `tensor` as [input](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/logging_ops.cc#L68), no `tag`, and thus couldn't take a string tensor as tag. This is probably inline with the new summary interface . I should wait for more unified solution from google and close this?

PR #5566, opened 2016-11-12
Fix syntax error in UnchangedShape example code

ghost:

PR #5567, opened 2016-11-12
Fix syntax error in UnchangedShape example code

ghost:
This is the same patch as #5566, but applied to master.


rmlarsen (approved):

rmlarsen:

PR #5571, opened 2016-11-13
Add support for decoding RGBA images in tfexample_decoder.

xodus7:
Tensorflow's image decoding API's support RGBA images with png/raw formats, but the tfexample_decoder module cannot handle them due to an exception that is thrown when the decode_jpg() case is evaluated with channels >3. This PR works around that limitation by eliminating JPEG as a decoding option when the number channels is >3.


rmlarsen:

rmlarsen (approved):

rmlarsen:
@gunan I assume the Windows cmake failures here can be ignored?

gunan:
This one looks like a different failure. But I think it is a flake.
To make sure:
Jenkins, test this please.

PR #5581, opened 2016-11-13
Fix CMake config if used as a subproject

NewProggie:
With this fix we don't assume that the tensorflow project is the build
root. Instead `CMAKE_CURRENT_SOURCE_DIR`/`CMAKE_CURRENT_BINARY_DIR` is used
instead of just `CMAKE_SOURCE_DIR`/`CMAKE_BINARY_DIR`, which now builds
successfully, if it is built inside another, larger project.


mrry (approved):
Thanks for the fix! LGTM.

mrry:

NewProggie:
Seems that the one failing test has nothing to do with my PR?

mrry:
Yes, the test failure looks unrelated. I'll merge the PR now.

PR #5590, opened 2016-11-14
Remove argv argument from tf.app.run()

leoybkim:
The second argument in tf.app.run is causing a TypeError and I think we can remove this.  

    $ python classify_image.py 
      File "classify_image.py", line 227, in <module>
        tf.app.run(main=main, argv=[sys.argv[0]] + unparsed)
    TypeError: run() got an unexpected keyword argument 'argv'


vrv:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/platform/app.py#L26 run() indeed has an argv parameter now -- you are probably trying to use classify_image.py at HEAD when you're running an earlier version of the binary.

PR #5604, opened 2016-11-14
Fix: Change the lock file location for GPU tests.

gunan:
/var/lock does not exist on mac, so execution fails on mac.


gunan:
Looks like macos does not have flock at all. 
So I will need to skip this completely on macosx

PR #5610, opened 2016-11-15
Add missing backticks to documentation.

shirriff:
The formatting for [extract_image_patches](https://www.tensorflow.org/versions/r0.11/api_docs/python/array_ops.html#extract_image_patches) is missing backticks, which messes up the code vs non-code formatting.


shirriff:
I signed the CLA earlier and I'm the commit author. Is there something I need to do to pass the check or can you override it?

vrv:
Your git commits should use the email you signed the CLA with (right now it's using a 'local' git address).

PR #5614, opened 2016-11-15
[Windows/CMake] Use a more recent version of SWIG

mrry:
This fixes `reader_ops_test.py`, which was suffering from a bug in older versions of SWIG when using Python 3.5 and a SWIG-wrapped generator object.


mrry:
@gunan, just trying this out directly, now that the Jenkins pool is upgraded.

mrry:
@gunan @rmlarsen This PR appears to fix the Windows build. I propose merging this rather than my internal CL, since this will be more useful sooner....

gunan (approved):

PR #5616, opened 2016-11-15
Remove version_info.cc from framework_internal

javidcf:
version_info.cc was being compiled into both version_lib and framework_internal, resulting in duplicated symbols. The header version.h should also be associated with version_lib.

### DIFF
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -1117,6 +1117,7 @@ tf_version_info_genrule()
 cc_library(
     name = "version_lib",
     srcs = ["util/version_info.cc"],
+    hdrs = ["public/version.h"],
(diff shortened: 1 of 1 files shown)


girving:
Jenkins, test this please.

tensorflow/core/BUILD
```diff
@@ -1128,7 +1129,6 @@ tf_cuda_library(
             "example/**/*.cc",
             "framework/**/*.h",
             "framework/**/*.cc",
-            "public/version.h",
```
girving: I half expect removing this line to cause a bunch of build errors for things that don't depend on `version_lib`, but we'll see.

girving (approved):

PR #5620, opened 2016-11-15
Enable a mechanism for filtering out log messages using

### LINKED ISSUE #1258: Create a way to control log message output.
Create a way to control log message output., most options should be passed via ConfigProto or another proto. Maybe add a LogOptions structure to hold options related to logging.  #1229
### Environment info

Operating System: Mac 10.9.5
### Feature request
1. Have a way to control logs that you record.As in gflags/re2 flags.Refer  #1229 
### What have you tried?
1. #1229

vrv:
an environment variable.  Reads the environment variable
on the first call to LOG, uses it to filter out based on the
severity of the log message.

By default, min log level is 0, mapping to INFO.
If you set the env var to 1, INFO is filtered out b t
WARNING still gets printed.

Parsing code is rudimentary, to avoid bringing in
dependencies into platform (e.g., strings/lib/numbers.h,
env_var.h, etc).  For this use-case it is probably fine.

Fixes #1258.

### DIFF
--- a/tensorflow/core/platform/default/logging.cc
+++ b/tensorflow/core/platform/default/logging.cc
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
+#include "tensorflow/core/lib/core/stringpiece.h"
 #include "tensorflow/core/platform/default/logging.h"
 #include "tensorflow/core/platform/macros.h"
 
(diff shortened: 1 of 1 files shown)


tensorflow/core/platform/default/logging.cc
```diff
@@ -81,7 +82,38 @@ void LogMessage::GenerateLogMessage() {
...
+  const char* tf_env_var_val = getenv("TF_CPP_MIN_LOG_LEVEL");
+  if (tf_env_var_val == nullptr) {
+    return 0;
+  }
+
+  // Ideally we would use env_var / safe_strto64, but it is
+  // hard to use here without pulling in a lot of dependencies,
+  // so we do a poor-man's parsing.
+  StringPiece min_log_level_sp(tf_env_var_val);
+  string min_log_level_str(min_log_level_sp.data(), min_log_level_sp.size());
```
zheng-xq: How about this? 

string min_log_level_str(tf_env_var_val)

tensorflow/core/platform/default/logging.cc
```diff
@@ -81,7 +82,38 @@ void LogMessage::GenerateLogMessage() {
...
+  if (min_log_level_str == "1") {
```
zheng-xq: How about also their actual names? For example, case insensitive, "info", "warning", "error", etc.

vrv: info / warning /error don't indicate the order of INFO vs. WARNING and ERROR, so I think we should (for now) keep it as the integers.  I've added some comments at least.

zheng-xq (commented):
LGTM

PR #5640, opened 2016-11-16
Simplify CMake thirdparty library target names

NewProggie:
In order to build platform independent static libraries, we can use CMakes
builtin `CMAKE_STATIC_LIBRARY_PREFIX` and `CMAKE_STATIC_LIBRARY_SUFFIX`
variables to separate the naming convention for Windows and Unix host
systems.

### DIFF
--- a/tensorflow/contrib/cmake/external/farmhash.cmake
+++ b/tensorflow/contrib/cmake/external/farmhash.cmake
@@ -1,4 +1,4 @@
-include (ExternalProject)
+include(ExternalProject)
 
 set(farmhash_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/external/farmhash_archive ${CMAKE_CURRENT_BINARY_DIR}/external/farmhash_archive/util)
 set(farmhash_URL https://github.com/google/farmhash/archive/34c13ddfab0e35422f4c3979f360635a8c050260.zip)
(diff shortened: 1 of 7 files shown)


mrry (dismissed):
Thanks for cleaning this up!

mrry:

mrry:

mrry:

mrry:
@NewProggie It looks like `libjsoncpp.a` is the wrong name for the Linux CPU CMake build. Can you fix that up and we'll run the tests again? Thanks!

mrry:

mrry:
It looks like there's still a problem with the gRPC library name on Linux. My guess is that it only contains `Release/` on Windows.

NewProggie:
Ah, shoot. I'll return to this one time in the future and fix all the cross references, as there's no technical reason for all the thirdparty libs to build in different subfolders (as far as I can see).

mrry:
OK - thanks for the effort here though. I think the inconsistency stems from the fact that some of the modules are built in tree (then we link the libraries directly from wherever the build output put them), whereas others are also installed (which puts them in a more predictable location).

It'd be great to sort this out at some point!

danmane:
Looks like this PR is dead at the moment, so I'm going to close it. Feel free to re-open or create a new one when its time to move forward.

PR #5644, opened 2016-11-16
Add Windows GPU support with Bazel

meteorcloudy:
@mrry @dslomov @damienmg 

The change needed for Bazel is also ready to review:
https://bazel-review.googlesource.com/?polygerrit=0#/c/7351/
Once these changes are checked in, we should be able to build TensorFlow with GPU support on Windows by
`bazel build --confg=win-cuda -c opt --cpu=x64_windows_msvc --cpu=x64_windows tensorflow/tools/pip_package:build_pip_package`

I excluded `depthwise_conv_grad_op` which doesn't seem to be excluded in CMake build.
It caused a linking error like this, and only happens in GPU build:
```
depthwise_conv_grad_op.o : error LNK2019: unresolved external symbol "public: static void __cdecl tensorflow::DepthwiseConv2dBackpropInputGPULaunch<float>::Run(struct Eigen::GpuDevice const &,struct tensorflow::DepthwiseArgs,float const *,float const *,float *)" (?Run@?$DepthwiseConv2dBackpropInputGPULaunch@M@tensorflow@@SAXAEBUGpuDevice@Eigen@@UDepthwiseArgs@2@PEBM2PEAM@Z) referenced in function "public: static void __cdecl tensorflow::LaunchDepthwiseConvBackpropInputOp<struct Eigen::GpuDevice,float>::launch(class tensorflow::OpKernelContext *,struct tensorflow::DepthwiseArgs,float const *,float const *,float *)" (?launch@?$LaunchDepthwiseConvBackpropInputOp@UGpuDevice@Eigen@@M@tensorflow@@SAXPEAVOpKernelContext@2@UDepthwiseArgs@2@PEBM2PEAM@Z)
[...]

PR #5646, opened 2016-11-16
Maunally setting TF_NEED_CUDA=0 in build_tf_windows.sh

meteorcloudy:
Soon TF_NEED_CUDA won't be False by default on Windows, we need this to make sure [tf-master-win-bzl ](http://ci.tensorflow.org/job/tf-master-win-bzl/) won't fail. @gunan


gunan:
I would like to create new scripts for those under windows/**gpu**/
I would like to make what we execute in CI visible to developers.

gunan:
And we will have a separate GPU build for windows, as in linux.

meteorcloudy:
Oh, I didn't mean to use this script for GPU build, I just want keep the CPU build work after #5644.
Because after that, if we don't set `TF_NEED_CUDA=0`, `echo "" | ./configure` will just fail.
Sorry I didn't make that clear.

gunan (approved):

gunan:
OK, now I understand :)
Jenkins, test this please.

meteorcloudy:
Jenkins, test this please.

PR #5647, opened 2016-11-16
Improved support for OpenCL devices

benoitsteiner:


lukeiwanski:
I signed it !

gunan (approved):

gunan:
Will have to override CLA check.
CLA check ok, but it wont accept as PR creator does not own all commits, or something in those lines.

PR #5656, opened 2016-11-17
Make windows test script version independent.

gunan:
This should fix the windows cmake build issues.

### DIFF
--- a/tensorflow/tools/ci_build/windows/cpu/cmake/run_py.bat
+++ b/tensorflow/tools/ci_build/windows/cpu/cmake/run_py.bat
@@ -30,8 +30,14 @@ SET PIP_EXE="C:\Program Files\Anaconda3\Scripts\pip.exe"
 :: Uninstall tensorflow pip package, which might be a leftover from old runs.
 %PIP_EXE% uninstall -y tensorflow
 
+
+:: Since there are no wildcards in windows command prompt, use dark magic to get the wheel file name.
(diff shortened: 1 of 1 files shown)


mrry:
I like this! Is it responsible for the flakes?

mrry (approved):

gunan:
I think it is possible because of this we had a bunch of breakages today with the version update.
I will wait for cmake python tests and then merge.

tensorflow/tools/ci_build/windows/cpu/cmake/run_py.bat
```diff
@@ -30,8 +30,14 @@ SET PIP_EXE="C:\Program Files\Anaconda3\Scripts\pip.exe"
 :: Uninstall tensorflow pip package, which might be a leftover from old runs.
 %PIP_EXE% uninstall -y tensorflow
 
+
+:: Since there are no wildcards in windows command prompt, use dark magic to get the wheel file name.
+DIR %REPO_ROOT%\%BUILD_DIR%\tf_python\dist\ /S /B > wheel_filename_file
+set WHEEL_FILENAME= <wheel_filename_file
```
mrry: I think this might need to be `set /p WHEEL_FILENAME=<wheel_filename_file`. The `/p` means it reads a line of input (the "p" stands for "prompt"!) into the named variable.

mrry (commented):
Looks like this didn't get picked up properly....

gunan:
OK now it is working with /p and no space between = and <

PR #5657, opened 2016-11-17
0.11.0rc2 -> 0.11.0, fix broken windows cmake ci.

Oneplus:


gunan:
Thanks for the change, however I just prepared #5656 which should be a more generic update.
If the tests pass, I will merge that and abandon this one.

PR #5658, opened 2016-11-17
Produce binary release tarballs for the TensorFlow C API

asimshankar:
These scripts are intended to be run with every release to
produce libtensorflow.tar.gz for CPU and GPU on Linux and OS X
for x86_64 architecture machines.
(Eventually there will be other operating systems and architectures).

These binary releases are then intended to make use of other language
bindings (such as [Rust](https://github.com/tensorflow/rust), [Haskell](https://github.com/tensorflow/haskell), [Go](https://godoc.org/github.com/tensorflow/tensorflow/tensorflow/go)) easier
as the common case would be to download the binary C-library release and
avoid the need to build TensorFlow from source (and all the time and
external dependencies doing so entails).

Files:
- tensorflow/tools/ci_build/builds/libtensorflow.sh - Baseline common script to build a tarball
- tensorflow/osx/libtensorflow_{cpu,gpu}.sh - Build tarballs for OS X with and without GPU support
- tensorflow/linux: Has similar top level scripts, but the build happens in a docker container so it contains 4 files - the two top level builds, one shared libtensorflow_docker.sh that is used to build and execute the docker container and libtensorflow.sh which is the script run inside the container.

### DIFF
--- a/tensorflow/tools/ci_build/builds/libtensorflow.sh
+++ b/tensorflow/tools/ci_build/builds/libtensorflow.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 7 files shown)


gunan (approved):

asimshankar:
Jenkins, test this please

The test failures seemed unrelated, retrying.

gunan:
Jenkins, test this please

gunan:
both failures are known issues. Merging.

PR #5685, opened 2016-11-18
Fix integer overflow error

MikalaiDrabovich:
This sample crashes if used on big text files (>2B words).
Changing corpus_size_ to 64-bit int solves the problem.


MikalaiDrabovich:
CLA submitted

benoitsteiner:
Jenkins, test this please

benoitsteiner (approved):

PR #5689, opened 2016-11-18
Ensure stochastic_tensor_test runs.

darrengarvey:
stochastic_tensor_test.py was missing a call to tf.test.main() so the
test wasn't actually running. This commit fixes that and the test
passes.


rmlarsen:

PR #5690, opened 2016-11-18
Replace squeeze/split with unpack

khryniewicz:
In [Recurrent Neural Networks](https://www.tensorflow.org/versions/r0.11/tutorials/recurrent/index.html) tutorial file `ptb_word_lm.py` I've replaced the alternative version of the code with `unpack` in place of `squeeze` and `split` (as in [documentation](https://www.tensorflow.org/versions/r0.11/api_docs/python/array_ops.html#split)).

### DIFF
--- a/tensorflow/models/rnn/ptb/ptb_word_lm.py
+++ b/tensorflow/models/rnn/ptb/ptb_word_lm.py
@@ -130,7 +130,7 @@ class PTBModel(object):
     #
     # The alternative version of the code below is:
     #
-    # inputs = tf.unpack(inputs, num_steps, 1)
+    # inputs = tf.unstack(inputs, num_steps, 1)
     # outputs, state = tf.nn.rnn(cell, inputs, initial_state=self._initial_state)
     outputs = []
     state = self._initial_state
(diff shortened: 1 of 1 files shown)


tensorflow/models/rnn/ptb/ptb_word_lm.py
```diff
@@ -130,8 +130,7 @@ def __init__(self, is_training, config, input_):
     #
     # The alternative version of the code below is:
     #
-    # inputs = [tf.squeeze(input_step, [1])
-    #           for input_step in tf.split(1, num_steps, inputs)]
+    # inputs = tf.unpack(inputs, num_steps, 1)
```
vrv: I think there's an ongoing rename to move from unpack to unstack (to match numpy).  Can you make that change here?

tensorflow/models/rnn/ptb/ptb_word_lm.py
```diff
@@ -130,8 +130,7 @@ def __init__(self, is_training, config, input_):
     #
     # The alternative version of the code below is:
     #
-    # inputs = [tf.squeeze(input_step, [1])
-    #           for input_step in tf.split(1, num_steps, inputs)]
+    # inputs = tf.unstack(inputs, num_steps, 1)
```
vrv: One more suggestion: might be nice to use named arguments here for clarity:

num=num_steps, axis=1

vrv (approved):
Approval after the suggested change.

PR #5691, opened 2016-11-18
change validation_metrics as  MetricSpec

BoyuanJiang:
tensorflow:Please specify metrics using MetricSpec. Using bare functions or (key, fn) tuples is deprecated and support for it will be removed on Oct 1, 2016.
so run the previous code will get a mistake!


BoyuanJiang:
I signed it!

PR #5693, opened 2016-11-18
install Pillow in Dockerfile

foreverfaint:
Hi team,

When i used **tensorflow** in python, sometimes i try the code like 
`scipy.misc.imread('test.tif')`
but i get an error:
`AttributeError: 'module' object has no attribute 'imread'. `

After some digging, i find i need to install PIL from [the docs](https://docs.scipy.org/doc/scipy/reference/misc.html) on scipy.misc. i guess there may be some other developers using tensorflow in my way (e.g. [fast-style-transfer src/utils.py/#9](https://github.com/lengstrom/fast-style-transfer/blob/master/src/utils.py)). 

So i would like to suggest adding Pillow into docker image build.

### DIFF
--- a/tensorflow/tools/docker/Dockerfile
+++ b/tensorflow/tools/docker/Dockerfile
@@ -30,6 +30,7 @@ RUN pip --no-cache-dir install \
         numpy \
         scipy \
         sklearn \
+        Pillow \
         && \
     python -m ipykernel.kernelspec
 


caisq:
@foreverfaint , can you let me know the approximate docker size increase (`docker images`) due to this change?

foreverfaint:
I have signed it. I used foreverfaint@gmail.com for signing. 
Google Individual CLA	Hao	Nov 18, 2016 06:14 PST	Edit Contact Information

but when i submit this PR, my local git global config is wrong. It was *foreverfailt@gmail.com* due to a typo on failt (should be faint)

foreverfaint:
@caisq, it increases from 958.6 to 981.6. about 23MB up. Is it acceptable?

REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
gcr.io/tensorflow/tensorflow   Pillow              dff6ac3189bd        10 hours ago        981.6 MB
gcr.io/tensorflow/tensorflow   latest              5547120ff897        9 days ago          959.6 MB

caisq:
Since is <2.5% increase, I think it is acceptable. For symmetry, can you also add it to the following file:
tensorflow/tools/docker/Dockerfile.gpu

foreverfaint:
no problem, will do it.

foreverfaint:
because i used a wrong author name (not CLA signed one), i can't pass CLA check. so i am going to incorprate the change mentioned by @caisq into Dockerfile.gpu with a right author name, and create a new PR for review. will at you again in new PR. @mention-bot @tensorflow-jenkins @googlebot @caisq 

close this one now

PR #5699, opened 2016-11-18
Use passthrough args rather than wrapper functions

elibixby:
Googlers can look see [this doc](https://docs.google.com/document/d/1gVcn_r5njuEOEZDXXW3HubWqAGJDs4M5APX1d_JLr4o/edit#heading=h.mgiweq7i3oas) for an explanation of why this change is being made.

This change was made in a backwards compatible way. Other changes referenced in that doc will likely depend on this change but not be backwards compatible.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/estimators/estimator.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/estimator.py
@@ -964,7 +964,7 @@ def _identity_feature_engineering_fn(features, labels):
 class Estimator(BaseEstimator):
   """Estimator class is the basic TensorFlow model trainer/evaluator.
   """
-  @deprecated_args('2017-2-1', 'Pass parameters as keyword arguments to')
+  @deprecated_args(PARAMS_DATE, PARAMS_INSTRUCTIONS, 'params')
   def __init__(self,
                model_fn=None,
                model_dir=None,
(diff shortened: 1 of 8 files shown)


tensorflow/contrib/learn/python/learn/estimators/estimator.py
```diff
@@ -343,7 +348,7 @@ def config(self):
       SCIKIT_DECOUPLE_DATE, SCIKIT_DECOUPLE_INSTRUCTIONS, 'x', 'y', 'batch_size'
   )
   def fit(self, x=None, y=None, input_fn=None, steps=None, batch_size=None,
-          monitors=None, max_steps=None):
+          monitors=None, max_steps=None, *input_fn_args, **input_fn_kwargs):
```
martinwicke: High-level: Can we allow only kwargs as a matter of policy?

elibixby: I suppose, although it would be nice for users to be able to specify require arguments without having to resort to:
```
if arg is None:
  raise ValueError()
```
or a downstream error

Still I'm sensitive to the backwards compat issues taking arbitrary positional args raises. I still prefer allowing a catchall pass through here, but if you feel strongly I can remove it (or replace it with:
```
def fit(..., input_fn_args=[], **input_fn_kwargs)
```
even.

danmane:
@elibixby this will need to be rebased

drpngx:
@elibixby is this still current? We need to merge conflicts.

drpngx:
@elibixby Are we still doing this?

martinwicke:
We are in the middle of a redesign. You can close this, we're aware of it
and can use it as a prototype.

PR #5718, opened 2016-11-19
Return to initialize_all_variables()

roee058:
Return to initialize_all_variables() instead of global_variables_initializer() 
Got the following error while using global_variables_initializer:
AttributeError: 'module' object has no attribute 'global_variables_initializer'


vincentvanhoucke (approved):
LGTM

vrv:
This code is correct at HEAD.  Check out the code at the branch you have installed otherwise.

PR #5721, opened 2016-11-19
R0.12

gunan:
Update release version @asimshankar just to check if the new string complies with all our usecases.
Adding windows GPU build script to the branch.

### DIFF
--- a/tensorflow/contrib/cmake/setup.py
+++ b/tensorflow/contrib/cmake/setup.py
@@ -26,7 +26,7 @@ from setuptools import find_packages, setup, Command
 from setuptools.command.install import install as InstallCommandBase
 from setuptools.dist import Distribution
 
-_VERSION = '0.11.0-cmake-experimental'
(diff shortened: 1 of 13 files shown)


asimshankar:
You mean if the version string is semver compliant? 
Looking at semver.org, it seems that "0rc" is not a valid patch since it can't have leading zeros.
Also, the whl files are currently a bit funny - 0.12.0-0rc-0rc... (so 0rc twice).

I think what we might want to do is to set `TF_PATCH_VERSION` in `tensorflow/core/public/version.h` to `0-rc0` but remove the hyphen from the filenames of the `.whl`s

gunan:
Thanks for the explanation.
I made a change to have the version string as '0.12.0-rc0', but make pip convert this to '0.12.0rc0'

asimshankar (commented):
'TF_BINARY_URL' looks funny

gunan:
You are right,
looks like we will need to handle version updates in a special way.

PR #5724, opened 2016-11-20
enabling VocabularyProcessor.transform to also return document length.

abhitopia:
Currently, `VocabularyProcessor.transform()` ( and `VocabularyProcessor.fit_transform()` ) do not return the sequence length of ( number of tokens in )  individual raw documents. sequence lengths are required when dealing with variable length sequences ( eg. like `dynamic_rnn` where we need to also pass the lengths of individual sequences).

After the changes in PR, both `fit_transform()` and `fit()` has following signature - 
- `VocabularyProcessor.fit_transform( raw_documents, return_docs_len=False, unused_y=None)`
- `VocabularyProcessor.transform( raw_documents, return_docs_len=False)`

Setting `return_docs_len=True` makes above function to also yield the number of token in individual documents.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/preprocessing/text.py
+++ b/tensorflow/contrib/learn/python/learn/preprocessing/text.py
@@ -155,13 +155,13 @@ class VocabularyProcessor(object):
     self.vocabulary_.freeze()
     return self
 
-  def fit_transform(self, raw_documents, return_docs_len=False, unused_y=None):
+  def fit_transform(self, raw_documents, unused_y=None, return_docs_len=False):
     """Learn the vocabulary dictionary and return indexies of words.
 
     Args:
       raw_documents: An iterable which yield either str or unicode.
-      return_docs_len: Boolean if True yields number of tokens also.
       unused_y: to match fit_transform signature of estimators.
+      return_docs_len: Boolean if True yields number of tokens also.
 
     Returns:
       x: iterable, [n_samples, max_document_length]. Word-id matrix.
--- a/tensorflow/contrib/learn/python/learn/preprocessing/text.py
+++ b/tensorflow/contrib/learn/python/learn/preprocessing/text.py
@@ -155,20 +155,22 @@ class VocabularyProcessor(object):
     self.vocabulary_.freeze()
     return self
 
-  def fit_transform(self, raw_documents, unused_y=None):
+  def fit_transform(self, raw_documents, return_docs_len=False, unused_y=None):
     """Learn the vocabulary dictionary and return indexies of words.
 
     Args:
       raw_documents: An iterable which yield either str or unicode.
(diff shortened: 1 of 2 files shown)


enizhibitsky:
According to `fit_transform(X, y)` semantic, you should place your flag after `unused_y`.

abhitopia:
@enizhibitsky - Thanks for pointing out. Made the changes.

abhitopia:
@martinwicke , @ilblackdragon - Have you had a chance to look at it yet? It's a very minor patch.

abhitopia:
@martinwicke - Any comments on this?

martinwicke:
@ilblackdragon I suppose we're keeping VocabularyProcessor around for lack of an alternative, would you agree? In that case, I think this is sensible, but you know the code much better.

abhitopia:
@ilblackdragon - what do you think?

abhitopia:
@martinwicke , @ilblackdragon , @danmane - been a while, please someone review.

abhitopia:
@martinwicke - It seems that @ilblackdragon is unavailable, can you assign someone else please?

martinwicke:
Sorry, everybody's at NIPS.

tensorflow/contrib/learn/python/learn/preprocessing/text.py
```diff
@@ -155,38 +155,45 @@ def fit(self, raw_documents, unused_y=None):
...
     """Learn the vocabulary dictionary and return indexies of words.
 
     Args:
       raw_documents: An iterable which yield either str or unicode.
       unused_y: to match fit_transform signature of estimators.
+      return_docs_len: Boolean if True yields number of tokens also.
 
     Returns:
       x: iterable, [n_samples, max_document_length]. Word-id matrix.
+        Additional token count also returned is `return_docs_len= True`.
```
martinwicke: add backticks around pseudocode in brackets. Make the second sentence: "If `return_doc_len` is `True`, returns a tuple `(x, n_tokens)`, where x is the word-id matrix as before, and `n_tokens` is the number of tokens in the document."

martinwicke (changes requested):
Small doc change, otherwise looks good to me.

ilblackdragon:
@abhitopia Thanks for the contribution. Just a question - what do you think about `return_docs_len` to be argument of the `VocabularyProcessor` constructor? As I would expect that if `fit` returns then `transform` should return as well. Also how are you passing this (x, doc_length) into Estimator? Probably quick example would be great.

@martinwicke Yes, there is no real alternative (even after we get feature_to_id op, we still need vocabulary generator).

drpngx:
@abhitopia could you address @ilblackdragon concerns?

drpngx:
Closing due to inactivity. Feel free to reopen when you're able to address review comments.

PR #5727, opened 2016-11-20
Fix for the code in the tf.cond example.

b0noI:
Small problem, however without the fix it creates some confusion when copy/pasting code to the datalab Notebook (or any other place).


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please

PR #5734, opened 2016-11-20
Force precision to be float for the MNIST fully connected example.

tiriplicamihai:
Right now, when the example runs precision is always 0 because it computed by dividing two ints and the result is always less than 1.


tiriplicamihai:
I signed it!

caisq:

caisq (approved):

caisq:
PR merged. Thank you, @tiriplicamihai

PR #5739, opened 2016-11-21
[Windows/CMake] Fix some C++ tests

vit-stepanovs:
After recent changes, some enabled C++ tests do not build on Windows. This PR fixes that, plus has fixes to some other C++ tests to build under VC++. I have further fixes for other C++ tests too - will send them in subsequent PRs.


tensorflow/examples/label_image/main.cc
```diff
@@ -93,7 +93,8 @@ Status ReadTensorFromImageFile(string file_name, const int input_height,
 
   string input_name = "file_reader";
   string output_name = "normalized";
-  auto file_reader = ReadFile(root.WithOpName(input_name), file_name);
+  auto file_reader = tensorflow::ops::ReadFile(root.WithOpName(input_name),
```
mrry: Out of curiosity, shouldn't the `using namespace ::tensorflow::ops;` on L92 make this change unnecessary?

mrry (dismissed):
Changes look good to me, but I have one clarifying questions.

mrry:

tensorflow/contrib/cmake/tf_cc_ops.cmake
```diff
@@ -84,3 +84,35 @@ add_library(tf_cc_ops OBJECT
...
+    "${tensorflow_source_dir}/tensorflow/cc/framework/gradients.cc"
+)
+
+file(GLOB_RECURSE tf_cc_test_srcs
+    "${tensorflow_source_dir}/tensorflow/cc/*test*.cc"
+)
+
+list(REMOVE_ITEM tf_cc_srcs ${tf_cc_test_srcs})
+
+add_library(tf_cc OBJECT ${tf_cc_srcs})
```
mrry: I think this needs a dependency on some or all of the generated code (and that's what's causing the Linux CMake build to fail).

mrry (changes requested):

tensorflow/examples/label_image/main.cc
vit-stepanovs: There is ReadFile function defined in some Windows header, which results in ambiguous name error.  Fixed this by specifying namespace explicitly.

tensorflow/contrib/cmake/tf_cc_ops.cmake
vit-stepanovs: Added the missing dependencies.

tensorflow/contrib/cmake/tf_cc_ops.cmake
```diff
@@ -84,35 +84,36 @@ add_library(tf_cc_ops OBJECT
...
+add_dependencies(tf_cc tf_cc_framework tf_cc_ops)
```
vit-stepanovs: This is the only change in the commit. All other lines are highlighted because there were Windows-style line endings in the file, and I fixed them in this commit.

mrry:

mrry (approved):

PR #5750, opened 2016-11-21
Default optimizer initialization in the LinearRegressor fix

b0noI:
Default optimizer of the LinearRegressor should be set only if the users have not provided custom optimizer.


martinwicke (approved):

martinwicke:
This is a cosmetic change, right? Is there a situation in which setting the default and then overwriting it is harmful?

martinwicke:
Jenkins, test this please.

b0noI:
martinwicke@ this is a tiny performance improvement, ```_get_default_optimizer``` allocates memory and executes small calculation.  However the improvement should be so tiny so I guess changes can be called as pure cosmetic.

Issue #4856, opened 2016-10-09
TensorBoard charts blank on Firefox


stepseazy:
NOTE: Only file GitHub issues for bugs and feature requests.  All other topics will be closed.

For general support from the community, see [StackOverflow](https://stackoverflow.com/questions/tagged/tensorflow).
To make bugs and feature requests more easy to find and organize, we close issues that are deemed
out of scope for GitHub Issues and point people to StackOverflow.

For bugs or installation issues, please provide the following information.
The more information you provide, the more easily we will be able to offer
help and advice.
### What related GitHub issues or StackOverflow threads have you found by searching the web for your problem?

I have not found any related issues
### Environment info

Operating System:

Ubuntu 16.04

Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):

```

rik@rik-MS-7971:~$ ls -l /usr/local/cuda/lib64/libcud*
-rw-r--r-- 1 root root   560184 Sep 23 09:09 /usr/local/cuda/lib64/libcudadevrt.a
lrwxrwxrwx 1 root root       16 Sep 23 09:09 /usr/local/cuda/lib64/libcudart.so -> libcudart.so.8.0
lrwxrwxrwx 1 root root       19 Sep 23 09:09 /usr/local/cuda/lib64/libcudart.so.8.0 -> libcudart.so.8.0.27
-rwxr-xr-x 1 root root   394472 Sep 23 09:09 /usr/local/cuda/lib64/libcudart.so.8.0.27
-rw-r--r-- 1 root root   737516 Sep 23 09:09 /usr/local/cuda/lib64/libcudart_static.a
lrwxrwxrwx 1 rik  rik        13 Jul 27 01:55 /usr/local/cuda/lib64/libcudnn.so -> libcudnn.so.5
lrwxrwxrwx 1 rik  rik        17 Jul 27 01:55 /usr/local/cuda/lib64/libcudnn.so.5 -> libcudnn.so.5.1.5
-rwxrwxr-x 1 rik  rik  79337624 Sep 26 21:55 /usr/local/cuda/lib64/libcudnn.so.5.1.5
-rw-rw-r-- 1 rik  rik  69756172 Sep 26 21:55 /usr/local/cuda/lib64/libcudnn_static.a
```

If installed from binary pip package, provide:
1. A link to the pip package you installed:
2. The output from `python -c "import tensorflow; print(tensorflow.__version__)"`.

```
[...]

PR #5755, opened 2016-11-21
Fix Plottable css importing, so that TB works in FF/Safari.

teamdandelion:
Fixes https://github.com/tensorflow/tensorflow/issues/4856


jart:
I recommend also putting "Fixes #4856" on the third line of the commit message so it gets closed automatically when this is merged.

jart (approved):

danmane:

PR #5756, opened 2016-11-21
Import Plottable css file so that TensorBoard works in FF/Safari

teamdandelion:
This is the r0.12 fix for https://github.com/tensorflow/tensorflow/issues/4856


jart (approved):

gunan:
This is also separately merged to master, right?

danmane:
Yeah, I separately merged it to master

PR #5760, opened 2016-11-21
Add deprecation warnings to tf.X_summary ops and tf.train.SummaryWriter

teamdandelion:


gunan:
FYI, rc0 binaries are already built, looks like we missed this PR.
this will make into rc1, is that OK?
Jenkins, test this please.

gunan (approved):

PR #5778, opened 2016-11-22
Add maximize method to Optimizer.

tillahoffmann:
This PR adds the convenience method `maximize` to the `Optimizer` class.


vrv:
Sorry Till, but we had this discussion in https://github.com/tensorflow/tensorflow/pull/3229 and the result was that as an interface, it makes sense to only implement one of the two methods.

(If python had the equivalent of a "final" keyword to prevent overriding, we'd be okay with it...)

tillahoffmann:
Makes sense. 👍

PR #5794, opened 2016-11-22
Make windows gpu build project name tensorflow_gpu.

yifeif:


mrry (approved):
LGTM!

gunan:
Could you also change the script for windows gpu build?
we only uninstall (pip uninstall) tensorflow at the moment. To avoid weird behaviour, we should uninstall both tensorflow and tensorflow_gpu

gunan:
Also, we need to modify windows GPU installation instructions.

yifeif:
ok, let me do that

PR #5798, opened 2016-11-23
make cuDNN linux instructions copy-pastable

cwlbraa:
previously, we were trying to cp files over directories


cwlbraa:
I signed it!

caisq:

caisq (approved):

Issue #5479, opened 2016-11-08
theano.clone feature


4SkyNet:
Are there some plans to do extension like a **theano.clone** ?

[Description from a original API:](http://deeplearning.net/software/theano/library/#theano.clone)

`theano.clone(output, replace=None, strict=True, share_inputs=True)`

Function that allows replacing subgraphs of a computational graph.
It returns a copy of the initial subgraph with the corresponding substitutions.

**Parameters:**

- **output** (Theano Variables (or Theano expressions)) – Theano expression that represents the computational graph.

- **replace** (dict) – Dictionary describing which subgraphs should be replaced by what.

- **share_inputs** (bool) – If True, use the same inputs (and shared variables) as the original graph. If False, clone them. Note that cloned shared variables still use the same underlying storage, so they will always have the same value.

It would be great if we can do similar things in **tensorflow** with easy.

aselle:
Closing since this seems to be fixed with pull request #557.

Issue #5477, opened 2016-11-08
Deal with _control_flow_context when copying op


thjashin:
In the current implementation of copying ops (both `tf.contrib.copy_graph` and `tf.contrib.graph_editor`)
The code of copying an op looks like this
```python
# copy inputs
inputs_ = copy_func(op.inputs)
# copy control_inputs
control_inputs_ = copy_func(control_inputs)
# copy _node_def, _op_def
node_def_ = deepcopy(op._node_def)
op_def_ = deepcopy(op._op_def)
output_types_ = op._output_types[:]
input_types_ = op._input_types[:]
# copy name
name_ = copy_func(op.name)
# init the new op with above copies
new_op = tf_ops.Operation(node_def_, ...)
# ... copy shape and add to graph
```
But the `op._control_flow_context` is not copied at all. This causes problems when trying to compute gradients on a copied subgraph with control flow op like `tf.cond`. The error looks like
```python
    grads = optimizer.compute_gradients(-lower_bound)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/optimizer.py", line 253, in compute_gradients
    colocate_gradients_with_ops=colocate_gradients_with_ops)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/ops/gradients.py", line 461, in gradients
    out_grads[i] = control_flow_ops.ZerosLikeOutsideLoop(op, i)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/ops/control_flow_ops.py", line 1310, in ZerosLikeOutsideLoop
    pred = op_ctxt.pred
AttributeError: 'NoneType' object has no attribute 'pred'
```
This is because the function `ZerosLikeOutsideLoop` uses `_control_flow_context` when the op is `tf.switch`
```
def ZerosLikeOutsideLoop(op, index):
  """Create zeros_like for the specified output of an op."""
  val = op.outputs[index]
  if not IsSwitch(op):
    return array_ops.zeros_like(val, optimize=False)
  else:
    op_ctxt = op._get_control_flow_context()
    pred = op_ctxt.pred
    branch = op_ctxt.branch
    switch_val = switch(op.inputs[0], pred)[1 - branch]
    zeros_shape = array_ops.shape_internal(switch_val, optimize=False)
    return array_ops.zeros(zeros_shape, dtype=val.dtype)
```
I tried setting `new_op. _control_flow_context` as `op._control_flow_context`
[...]

Issue #1070, opened 2016-02-12
Request: Equivalent of theano.clone() to dynamically replace subgraph


sotelo:
I would like to know if this feature exist in the library or if there's any plan to have it in the future.

Thanks!

vrv:
Hopefully someone else knows more about the literature, but is "weight noise" a generalized stochastic operation applied specifically to weights?  For example, is dropout a form of "weight noise" if applied to weights?

You can implement dropout as a composition of existing ops here: https://github.com/tensorflow/tensorflow/blob/8048088bfb82846078f8023bc6199e6424926624/tensorflow/python/ops/nn_ops.py#L581, so if you can express the specific noise algorithm you want as a composition of existing ops, then you can do that today.  If there's something more specific you are asking for, let us know, and feel free to contribute it to the library.

sotelo:
Thanks for answering :)

Dropout it's very similar to what I would like to do. The only problem is that I would like to apply it AFTER the rest of the computation graph is defined. As far as I understand, the current dropout function has to be applied before. This is what I am referring to when I say graph replacement. In theano, the mechanism to do this is use the theano.clone function with a replace dictionary as argument. The reason that I would like to do it like that, is that I want to use this kind of function in all the weights. So, for example, the weights that correspond to a LSTM that I don't have access to. Ideally, I would like to do something like this:

vars = tf.variables.all_variables()
new_vars = [tf.nn.dropout(var) for var in vars ]
tf.replace(vars, new_vars)

For a less general example: How can I apply dropout to the weights of a lstm?

I saw this discussion that is related, but I would like to know if there is a simpler way of doing things:
http://stackoverflow.com/questions/33748552/tensorflow-how-to-replace-a-node-in-a-calculation-graph

Thanks!

vrv:
OK great, I've renamed the title to reflect the more general feature request.

I can think of various ways to do this, but none of them are the same as theano.clone. They involve duplicating the entire graph as mentioned in the SO post, or decomposing your graph into "functions" (a currently experimental and evolving feature) and using tf.cond to select which branch to run at runtime, etc, or using sess.partial_run or tf.py_func (also both evolving and experimental) to embed the logic in the user program rather than in the graph.

Thanks for the feature request!

rdipietro:
+1 :)

girving:
Closing since the merged PR #557 seems related.

PR #557, opened 2015-12-19
Recursively copying elements from one graph to another

srjoglekar246:
Allows for easy portability of elements (Variables and Ops) from one Tensorflow Graph to another. If called on a top-level root in a dataflow graph, automatically copies all required instances.

Provides an API to retrieve the copied elements in the other graph, using a namespace.

Reference: https://codesachin.wordpress.com/2015/11/20/recursively-copying-elements-from-one-graph-to-another-in-tensorflow/

PR #5802, opened 2016-11-23
A theano.clone() equivalent for Tensorflow

thjashin:
This an implementation of a `clone()` function for Tensorflow based on `meta_graph` utilities, inspired by discussion with @sherrym in #5477 .

Feature requests on this can be found in #1070  #5479 . Previous implementations of copying graphs #557 does copy naively and doesn't have `replace` feature as in [theano.clone()](http://deeplearning.net/software/theano/library/#theano.clone).

Though this implementation is almost good. It still has problems for now with control flow copies. To be specific, I'm not sure how to deal with collections of `cond_context` and `while_context`, because both of its `to_proto` functions only consider the case of stripping `export_scope`.

Not copying `cond_context` collection to new operations (to their `_control_flow_context`) will cause problems described in #5477 , also demonstrated in the last test (`testCloneBatchNorm`) in this pull request, which is failed for now. I'm opening this PR early before that test pass, wondering if someone can have advice on dealing with control flow collections when cloning subgraphs. Thanks.


thjashin:
Anyone help with this PR? @sherrym , @keveman , @vrv , @ebrevdo , especially about the control flow context part.

yaroslavvb:
@yuanbyu may know about control flow parts

martinwicke:
Ping all! @sherrym, @yuanbyu can you look at this?

thjashin:
@yaroslavvb @martinwicke Thanks for help. I'm always willing to keep working on this feature.

ebrevdo:
How does this work related to the graph view and replace operations in contrib?  Can those tools be used to perform cloning?

thjashin:
@ebrevdo Thanks for the question. **The short answer**:
They cannot do exactly the same thing due to bugs, but to fix bugs touches the design. While using meta_graph, from my view, is the most elegant way to implement and maintain this feature.

**Long answer**:
I used to rely on `tf.contrib.graph_editor`. It seems to have everything. But I have to say it's still buggy, and maintenance seems to have stopped for some time. The `graph_replace` function provided in `tf.contrib.graph_editor` tries to do similar things with this PR, however, there are big cases it fails, for example, when one node in the given replacements is a descendant of another, the copied subgraph won't cover the uprooting graph of the descendant node even there are target nodes between this two nodes. I tries to fix but it turns out things are not easy under that framework. This is not the only problem, you could see I had a PR ( #5451 ) merged on problems about control inputs. It's still an easy one to fix, but there exist much harder ones.

Though this owes much to lack of extensive tests in `tf.contrib` modules, I'm still wondering whether `tf.contrib.graph_editor` is the right way to achieve this clone() ability. Then I found the `_control_flow_context` problem, which I raised in #5477 . This is because copying operations in `tf.contrib.graph_editor` is by directly copying members of a python objects. In that issue @sherrym pointed out to me that I could use meta_graph utilities to achieve what I want. So I made this trial. Though I still don't succeed in copying control flow contexts, I find the codes more elegant in this way, compared to implementations in `tf.contrib.graph_editor`.

I believe this is at least a right direction to implement this theano equilvalent. And I also want to point out that **this function is of great value to people building higher level libraries based on Tensorflow, it helps on operation level reuse (which cannot be achieved either by using name scope or by experimental function utilities in TF)**.

yaroslavvb:
@sherrym  ping. Can you comment if this is feasible to get merged?

@thjashin note that getting added to contrib means it's a candidate to be moved into core, which means that core TensorFlow team would have to take over support and maintenance. Hence, the code has to be easy enough to transfer over (ie, design docs can help sometimes)

drpngx:
@thjashin could you merge and push again?

thjashin:
@yaroslavvb Thanks. Any templates for design docs?

thjashin:
@drpngx Sure, I'll do it tomorrow.

yaroslavvb:
@thjashin I was thinking more of an informal thing that'll help someone understand for someone reading the code for the first time (ie, motivation, brief outline of how code is structured and why)

thjashin:
@drpngx Done.

drpngx:
Jenkins, test this please.

drpngx:
We might have to move that to contrib first.

thjashin:
@drpngx I think I've modified things in the framework, not just adding other functions...Can you give some advice on how to move them into contrib? btw, I guess if you can get some team member to review, can we directly merge? I guess I don't change the original behavior of other parts of code. Also if you need design docs. I'm very happy to write one. In fact it's just a DFS in the graph with some terminal conditions.

drpngx:
@sheerym or @yuanbyu are the best to review this.

drpngx:
@sherrym sorry, typo.

dandelionmane:
@thjashin I'll try pinging sherrym and yuanbyu directly to review this. In the mean time, can you fix the merge conflict that's appeared?

thjashin:
@dandelionmane Sure. I will do this tomorrow.

yaroslavvb:
cc @purpledog

thjashin:
@dandelionmane I've done the merge.

drpngx:
Jenkins, test this please.

drpngx:
@yuanbyu @sherrym Could you take a look? I'm not super familiar with this part of the code.

yuanbyu:
I am not very familiar with this part of the codebase. For the control flow context, it seems to strip name scope when exporting and to add back the "current" name scope when importing. Why doesn't it work in your case?

vrv:
I think for now, given that we don't really have any volunteer maintainers of this code internally, we're going to close this PR;  I think the current implementation of GraphDef and state in TensorFlow's python implementation is not very amenable to analysis and correctness, and while what you've done is very neat, it will probably make it very difficult or impossible to move the existing complex python code (control flow, state management) to better implementations (not a proof, but just a gut feeling).

I believe some of the work on improving functions and state management will make something equivalent to clone() possible and maintainable, and that's what we really need to make sure we do if we intend to make our users happy.

ebrevdo:
+1 I think for now we need to reserve the right to make changes to this
part of the tf codebase without maintaining a .clone implementation,
especially given the lack of internal maintainers.

PR #5814, opened 2016-11-23
Manual Inception folder

Research2:
Manually downloaded inception folder.


caisq:
With this change, all users would have to manually download the inception data, which is a non-starter for this tutorial. If I'm missing something, please kindly point it out and we can re-open this PR.

PR #5817, opened 2016-11-23
Passing exit code of the test back to bazel

ptrendx:
Script used for parallel GPU execution does not pass the exit code of the test to bazel (always return with exit 0), so failures are shown as PASS in the summary.


gunan:
Jenkins, test this please.

gunan:
Interesting catch. It seems like our CI was able to report failures fine without this though.
Your analysis makes sense, but I wonder how our CI reported failures.

gunan (approved):

ptrendx:
This is strange indeed - even running false gives exit code 0 when running under old version of the script.

gunan:
The GPU tests ended up on a faulty machine, rerunning:
Jenkins, test this please.

PR #5824, opened 2016-11-24
Be explicit about setting up CUDA

Mistobaan:
Now that OPENCL support is initiated I think makes sense to be explicit about the configuration option.


gunan:
Jenkins, test this please.

gunan (approved):

PR #5835, opened 2016-11-24
Fixing bug in Adadelta implementation

luheng:
The updated accum vector should be computed as:
\rho * accum + (1.0 - \rho) * grad^2,
according to Algorithm1 in Zeiler 2012 (https://arxiv.org/pdf/1212.5701v1.pdf).

### DIFF
--- a/tensorflow/core/kernels/training_ops_gpu.cu.cc
+++ b/tensorflow/core/kernels/training_ops_gpu.cu.cc
@@ -64,7 +64,8 @@ struct ApplyAdadelta<GPUDevice, T> {
     bcast[0] = grad.dimension(0);
     Eigen::Sizes<1> single;
 
-    accum.device(d) = accum_update * rho.reshape(single).broadcast(bcast) +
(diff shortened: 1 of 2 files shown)


luheng:
I signed it!

vrv:
Thanks!  I suspect you'll have to update the tests in adadelta_test.py, but I'll kick off the test anyway.

luheng:
Thanks! Weirdly the updates in adadelta_test.py are implemented correctly. Do I still need to update the tests?

vrv:
Do you know why the tests were passing before and after?  Should we come up with a better test that would have failed before and passes with your change?

luheng:
I updated the precision in adadelta_test.py so it will be able to test the difference.

vrv:
That's great!  Gives us more confidence it's doing the right thing now.

vrv (approved):

caisq:
PR merged. Thanks, @luheng

PR #5810, opened 2016-11-23
Improved support for OpenCL

benoitsteiner:

PR #5836, opened 2016-11-24
Fix: remove the leftover delta addition in cwise_ops_test.

gunan:
small fix for #5810.
Should fix mac tests on master.

### DIFF
--- a/tensorflow/python/kernel_tests/cwise_ops_test.py
+++ b/tensorflow/python/kernel_tests/cwise_ops_test.py
@@ -792,7 +792,7 @@ class BinaryOpTest(tf.test.TestCase):
         else:
           self._compareGradientX(x, y, np_func, tf_func)
           self._compareGradientY(x, y, np_func, tf_func)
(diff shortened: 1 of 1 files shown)


yifeif (approved):

gunan:
CC @prb12 for approval, if online.
This PR fixed the mac test failures we have been seeing.

lukeiwanski:
@gunan All the others _compare methods for float32 and true_divide op are adding 0.1 to y component. We assumed that this one was missing. Isn't that the case?

gunan:
@girving seems to have added some of the 0.1's.
Geoffrey, it looks like it has been a long time since yu made the changes, but do you remember why we have the 0.1's added as @lukeiwanski described above?
Do we need the `+ 0.1` Which I removed here?

girving:
@gunan Presumably the choice of rounding was different when the result was exactly a half integer.  I'm not sure this is different, but the only way I'd be able to tell would be removing them and trying.  

@aselle Do you think we should match exactly now?  It might depend on processor flags, unfortunately.

PR #5839, opened 2016-11-25
Add sess.close() in document (g3doc/get_started/index.md)

jangsoopark:
I read from tensorflow Session management document that 

A session may own resources, such as variables, queues, and readers. It is important to release these resources when they are no longer required. 
To do this, either **invoke the close() method on the session**, or **use the session as a context manager.** The following two examples are equivalent:

but I found g3doc/get_started/index.md didn't

so I made the pull request for add sess.close() at the end of code.

Thank you for making tensorflow


jangsoopark:
I signed CLA!

gunan:
Jenkins, test this please.

vrv (approved):

PR #5844, opened 2016-11-25
[CMake] Add TensorBoard dependencies to PIP package.

vit-stepanovs:
Currently, TensorBoard external dependencies (JS scripts, css, images etc.) are not part of PIP package built by CMake. As a result, when you navigate to TensorBoard on Windows, it shows a blank screen. Added CMake scripts to download TensorBoard dependencies and make them part of PIP package.


gunan:
Jenkins, test this please.

vit-stepanovs:
Looks like build failed because multiple dependencies have the same downloadable archive name... Not sure why this passed for me locally. Will make a change to download each archive in a separate directory.

mrry:

tensorflow/contrib/cmake/tf_python.cmake
```diff
@@ -374,10 +374,6 @@ add_python_module("tensorflow/contrib/tensor_forest/hybrid/python/ops")
 add_python_module("tensorflow/contrib/tensor_forest/python")
 add_python_module("tensorflow/contrib/tensor_forest/python/kernel_tests")
 add_python_module("tensorflow/contrib/tensor_forest/python/ops")
-add_python_module("tensorflow/contrib/tensorboard")
-add_python_module("tensorflow/contrib/tensorboard")
-add_python_module("tensorflow/contrib/tensorboard/plugins")
-add_python_module("tensorflow/contrib/tensorboard/plugins/projector")
```
mrry: These lines (with the exception of the duplicated line 378) still seem important... without them I'd expect importing `tf.contrib` to fail. Are they covered by some other rule now?

tensorflow/contrib/cmake/external/tensorboard.cmake
```diff
@@ -0,0 +1,49 @@
...
+  add_custom_command(
+    TARGET tensorboard_copy_dependencies PRE_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy ${src_dir}/${src_file} ${dst_dir}/${src_file}
+  )
+  
+  add_custom_target(${_TB_NAME} DEPENDS ${src_dir}/${src_file})
+  set(tensorboard_dependencies ${tensorboard_dependencies} ${_TB_NAME} PARENT_SCOPE)
+endfunction()
+
+include(external/tensorboard_deps.cmake)
```
mrry: This file seems horrific - from a maintenance standpoint - but we should be able to generate it automatically (in the same manner as [this file](https://github.com/tensorflow/tensorflow/blob/a31752e88407507df2a6d61b6a0d7c940d63ecc5/tensorflow/tensorboard/bower/BUILD)).

tensorflow/contrib/cmake/tf_python.cmake
vit-stepanovs: The other lines are also duplicates - the same lines can be found above at 358-360.

tensorflow/contrib/cmake/external/tensorboard.cmake
vit-stepanovs: I generated this file myself off WORKSPACE/bower.BUILD files, hence it is separate from tensorboard.cmake. I am happy to share my script if needed, but I think it may be better to modify the one script that updates WORKSPACE/bower.BUILD to do the same to this file (from a maintenance perspective, I guess it is better to have one script).

tensorflow/contrib/cmake/tf_python.cmake
mrry: Ack, thanks for catching that!

mrry (approved):
LGTM! Test failure is unrelated.

tensorflow/contrib/cmake/external/tensorboard.cmake
jart: I'm a TensorBoard developer. I've never used cmake before. I don't know how I would maintain this. What would you recommend for me?

mrry: @jart We already produce [a similar file](https://github.com/tensorflow/tensorflow/blob/5657d0dee8d87f4594b3e5902ed3e3ca8d6dfc0a/bower.BUILD) for Bazel. We could modify the script that auto-generates `bower.BUILD` (and the suffix of `WORKSPACE`) so that it also produces the contents of `external/tensorboard_deps.cmake`. Unfortunately that script isn't in the git repository, so we can't easily do it in this PR.

jart: I'm concerned because the TensorBoard team changes these external dependency definitions [about once a week](https://github.com/tensorflow/tensorflow/commits/master/bower.BUILD). I want to make sure the additional burden on the team is as minimal as possible.

It would be nice if this information could be extracted from [WORKSPACE](https://github.com/tensorflow/tensorflow/blob/master/WORKSPACE) and [bower.BUILD](https://github.com/tensorflow/tensorflow/blob/master/bower.BUILD) without making any modifications to either of those files. The cmake build already [surgically extracts](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/external/eigen.cmake) similar information from the workspace.bzl file for Eigen.

Also is Jenkins going to be regression [testing](https://github.com/tensorflow/tensorflow/blob/de6bbda2353b50944bd06d5b04c86f8c0a62792a/tensorflow/tools/ci_build/windows/cpu/cmake/run_build.bat) this change?

jart: Also take into consideration that [Bazel TensorFlow is officially coming to Windows](https://developers.googleblog.com/2016/11/tensorflow-0-12-adds-support-for-windows.html). The TensorBoard team is committed to making the new TensorBoard Bazel build fully functional on Windows, so Windows users can participate in TensorBoard development if they choose.

vit-stepanovs: I have changed tensorboard.cmake to parse dependencies from Bazel files.

jart:
Mr. Jenkins: test this please

jart (approved):
Thank you for putting the time and effort into making this configuration maintainable for us. As a result, the TensorBoard team will put forth the effort to keep them in a working state if Jenkins reports to us that they're broken. That is a promise we'll make you. But we only promise that up until the point when we get our canonical Bazel build to serve all your needs.

vit-stepanovs:
It looks like test failures are unrelated...

gunan:
For good measure, let me rerun tests, than we can merge right away.
Jenkins, test this please.

DomenicD:
How do I install this new PIP package or how can I create/build one with this fix?

vit-stepanovs:
@DomenicD, for now you can build a PIP package manually by following [these instructions](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/README.md). Otherwise, it looks like this fix will be included in 0.12rc1 release, which may be available some time soon.

ferrouswheel:
@DomenicD if you don't want to build manually you can also get a whl of the nightly build from here: http://ci.tensorflow.org/view/Nightly/job/nightly-win/lastSuccessfulBuild/DEVICE=gpu,OS=windows/

DomenicD:
@ferrouswheel thank you, nightly worked great.

PR #5846, opened 2016-11-25
Fix an error in 'Adding a New Op' example code

llhe:
There is an error in the 'Adding a New Op' example code. By chance, the test code failed to catch it because only 1-D tensor is tested.


vrv:

vrv (approved):

Issue #5832, opened 2016-11-24
Still 0.11.RC0 after upgrading to RC2


zuoxingdong:
I've used Anaconda-Pip installation in the tutorial to upgrade my TensorFlow

```
# Ubuntu/Linux 64-bit, CPU only, Python 3.5
(tensorflow)$ export TF_BINARY_URL=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.11.0rc2-cp35-cp35m-linux_x86_64.whl

# Python 3
(tensorflow)$ pip3 install --ignore-installed --upgrade $TF_BINARY_URL

....> ipython

In [1]: import tensorflow as tf
In [2]: tf.__version__
Out[2]: '0.11.0rc0'

```

But it seems it is still in RC0, in which doesn't include new method such as `tf.summary.merge_all()`

prb12:
Could you please check the TF version from python3 directly?

omtcyfz:
@prb12 
```
# Ubuntu/Linux 64-bit, CPU only, Python 3.5
(tf) omtcyfz@omtcyfz-VW:~$ export TF_BINARY_URL=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.11.0rc2-cp35-cp35m-linux_x86_64.whl
(tf) omtcyfz@omtcyfz-VW:~$ pip3 install --ignore-installed --upgrade $TF_BINARY_URL
Collecting tensorflow==0.11.0rc2 from https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.11.0rc2-cp35-cp35m-linux_x86_64.whl
(tf) omtcyfz@omtcyfz-VW:~$ python3
Python 3.5.2+ (default, Sep 22 2016, 12:18:14) 
[GCC 6.2.0 20160927] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tensorflow
>>> tensorflow.__version__
'0.11.0rc2'
```
My `__version__` is just fine.

zuoxingdong:
@prb12 

`
...> python3
Python 3.5.2 |Continuum Analytics, Inc.
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tensorflow as tf
>>> tf.__version__
'0.11.0rc0'
>>> `

I've tried to show TF version directly on python3, it still returns RC0

[...]

PR #5856, opened 2016-11-25
[Anaconda Installation Tutorial]As in issue #5832, pip installed TF s…

zuoxingdong:
…hould be uninstalled if one want to use TF with Anaconda environment, because Anaconda will search system site-packages with higher priority, this might result in the latest upgrade of TF within Anaconda environment inaccessible.

### DIFF
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -252,7 +252,8 @@ Follow the instructions on the [Anaconda download
 site](https://www.continuum.io/downloads).
 
 Note: If tensorflow has been installed via pip previously, then one should 
-uninstall it before installation in Anaconda environment. 
+uninstall it before installation in Anaconda environment. Because Anaconda 
+searches system site-packages from `.local` with higher priority.
 ```
 # Python 2
 $ pip uninstall tensorflow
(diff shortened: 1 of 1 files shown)


tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -251,6 +251,18 @@ Install Anaconda:
 Follow the instructions on the [Anaconda download
 site](https://www.continuum.io/downloads).
 
+Note: If tensorflow has been installed via pip previously, then one should
```
caisq: Do you mean "has been installed via pip outside the Anaconda environment previously"? If so, it will be good to make it explicit.

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -251,6 +251,18 @@ Install Anaconda:
 Follow the instructions on the [Anaconda download
 site](https://www.continuum.io/downloads).
 
+Note: If tensorflow has been installed via pip previously, then one should 
+uninstall it before installation in Anaconda environment. Because Anaconda
```
caisq: grammar nit: ". Because" --> ", because"

caisq:

PR #5860, opened 2016-11-25
Add load method to tf.Variable

dm0:
Allows loading new value into variable without adding new operations to the graph.

Sometimes it is convenient to just load new value into variable without adding new operations to the graph. For example one may want to load values into some weights from other model or model trained with other framework.

### DIFF
--- a/tensorflow/python/kernel_tests/variables_test.py
+++ b/tensorflow/python/kernel_tests/variables_test.py
@@ -372,12 +372,11 @@ class VariablesTestCase(tf.test.TestCase):
 
   def testLoad(self):
     with self.test_session():
-      var = tf.Variable(np.random.randn(5, 5))
+      var = tf.Variable(np.zeros((5,5), np.float32))
       tf.global_variables_initializer().run()
-      data = np.random.randn(5, 5)
-      var.load(data)
+      var.load(np.ones((5, 5), np.float32))
 
-      self.assertAllClose(data, var.eval())
+      self.assertAllClose(np.ones((5, 5), np.float32), var.eval())
 
 
 class IsInitializedTest(tf.test.TestCase):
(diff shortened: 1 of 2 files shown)


caisq:
Should the new method have unit-test coverage? @dm0

dm0:
@caisq Probably yes. Is it correct place to add unit test to: tensorflow/python/kernel_tests/variable_ops_test.py?

lukaszkaiser (changes requested):
This code looks good to me, but please add at least a simple unit test in python/kernel_tests/variables_test.py. Thanks!

lukaszkaiser:
I think python/kernel_tests/variables_test.py will be better. Thanks!

tensorflow/python/kernel_tests/variables_test.py
```diff
@@ -370,6 +370,15 @@ def testInitializerFunctionDevicePlacement(self):
       for i in v2.initializer.inputs:
         self.assertEqual(expected_group_v2, i.op.colocation_groups())
 
+  def testLoad(self):
+    with self.test_session():
+      var = tf.Variable(np.random.randn(5, 5))
```
lukaszkaiser: Let's make it np.zeros first, and then load np.ones and check that it gets ones, would that be ok?

lukaszkaiser (changes requested):
Let's change it so that we're loading sth different than what was there at first.

tensorflow/python/kernel_tests/variables_test.py
dm0: No problem.

There is a chance that two random sequences will be the same but it is insignificant. Or am I missing something?

lukaszkaiser (approved):
Looks good, thanks!

danmane:

alexgorban:

martinwicke:
Failure is a flake. Thank you!

PR #5864, opened 2016-11-25
Fixed wrong documentation on ctc_beam_search_decoder

lingz:
The documentation for `ctc_beam_search_decoder` has an error. It claims that it is equivalent to `ctc_greedy_decoder` when `top_paths=1`. This gives the impression that if we only care about the best result, we should always use `ctc_greedy_decoder`. However, `ctc_greedy_decoder` just does a simple argmax at each timestep, which is not equivalent result to a beam_search with `beam_width=100` and `top_paths=1`.

The only case they are equivalent is when both `beam_width` and `top_paths` are `1`. This pull request corrects this documentation error.


caisq:

ebrevdo:
Thanks for catching this!

ebrevdo (approved):

caisq:
PR merged. Thank you, @lingz !

PR #5866, opened 2016-11-25
variable_summaries(...) missing parameter 'name'

oliveirabc:
Versions 0.11 and 0.10 had a parameter named 'name' on variable_summaries function but this is now missing on master and 0.12


danmane:
The removal of this argument was deliberate. The new summary ops under the tf.summary module use the TensorFlow name scoping system for name de-duplication, so client code no longer needs to manage name scopes by hand.

PR #5879, opened 2016-11-27
Upload installation instructions.

yifeif:


tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -61,7 +61,17 @@ $ sudo easy_install pip
 $ sudo easy_install --upgrade six
 ```
 
-Then, select the correct binary to install:
+We have also uploaded the CPU version of the binaries to Pypi, so you can
+simply install on Linux, Mac or Windows with:
+
+```bash
+$ pip install tensorflow
```
caisq: Should we also mention tensorflow_gpu?

caisq (changes requested):

tensorflow/g3doc/get_started/os_setup.md
gunan: Gpu versions are too big to upload to pypi. So at the moment, we only have CPU in pypi.
I think this should be mentioned.

caisq:
Oops - I merged the PR before I noticed your new comment, @gunan @yifeif, maybe you can do another PR to mention the reason why tensorflow_gpu wheels are not available via PyPi?

PR #5874, opened 2016-11-26
changed global_variables_initializer line

Goddard:
Changed global_variables_initializer to initialize_all_variables on master.

PR #5889, opened 2016-11-27
Variable initialization method changed in tensorflow 0.11.0

tangshuran:
otherwise, variables not initialized


tangshuran:
I signed it!

zuoxingdong:
@tangshuran It is similar to the issue #5874, `tf.global_variables_initializer()` is the new API for TF, and `initialize_all_variables` probably to be deprecated soon. So I think it is better to keep the new API.

tangshuran:
Thanks~I thought it was old~

PR #5897, opened 2016-11-28
Fix params docstring of embedding_lookup

llhe:
The first parameter of embedding_lookup actually accept a single tensor.
```py
  if not isinstance(params, list):
    params = [params]
```
However, in the docstring it requires a list which is confusing.
This PR fix and unify the doc.


teamdandelion (approved):

danmane:
Thanks!

PR #5899, opened 2016-11-28
Make placement of constants follow consumers if they are all on the same device

DavidNorman:
A generator node (for instant a constant) will now go to the same device as its consumers, when those consumers are all on the same device.   Previously a constant would only follow its consumer when there was only a single consumer.


DavidNorman:
I am in the process of getting the CLA signed and set up.

DavidNorman:
I signed it!

DavidNorman:
The company's name is Graphcore.   Signed yesterday.

vrv:
So I can see GraphCore is in our corp CLA list, and you've signed your commits with your company email.  Can you make sure your email is part of the googlegroup registered with us when you sign up for the corp CLA? That way our bot will come back happy.

DavidNorman:
Is the problem that the google group contains davidjpnorman@gmail.com, and not davidjpnorman@googlemail.com (my GitHub email)?   I have tried adding davidjpnorman@googlemail.com, but googlegroups won't have it, recognizing that one is an alias of the other.

I think I will transfer ownership of the group, remove davidjpnorman@gmail.com, and add davidjpnorman@googlemail.com, and see if that works.

DavidNorman:
Hmm.  It seems that the email address is converted from @googlemail.com into @gmail.com by the google groups system.

Much as I hate to do it, I will see if I can alter the email address in the github account.

DavidNorman:
I signed it!

DavidNorman:
ok - i changed the GitHub email address to davidjpnorman@gmail.com - hopefully that will be ok

vrv:
So I think the following need to match:

1) git commit email on all your commits
2) email must be added to your github account
3) email must be registered on the internal list your company has to associate emails with the corp CLA account.

If those don't match, our CLA bot can't connect all of the pieces together.  Can you give this one more try?  If it doesn't work, I'll rope in one of the CLA people at Google to help :)

DavidNorman:
Ok.

so perhaps the problem is that the original commit comes from davidn@graphcore.ai, while the merge request is by the github account (now davidjpnorman@gmail.com).  

although both of these addresses are in the google group, perhaps they need to be the same.  Can you confirm?

DavidNorman:
Or - i guess that I might have to redo the merge request, because maybe it is tagged with the previous github account email address (davidjpnorman@googlemail.com), rather than the new one.   I'll investigate that.

DavidNorman:
Ok - i'll redo this whole request to see what happens.

DavidNorman:
I tried creating another:  https://github.com/tensorflow/tensorflow/pull/6615

however it didn't work either.  I will look through the pages that it is referring to in the error message

PR #5900, opened 2016-11-28
Fix sandboxed build

martinwicke:
There are a number of missing dependency declarations, which break the sandboxed bazel build. This PR adds a few, hopefully all.


martinwicke:
@jart, I think the proto rules are easier to use this way, but what I did may be unidiomatic. If there's a better way let me know.

jart (approved):
In the future, the Bazel team wants to use aspects to generate language specific code for proto_library but this is a nice and simple way to do it that you've discovered which should hold us over.

martinwicke:
Thanks! The GPU failure looks unrelated.

martinwicke:
Sorry @danmane in advance for the pain you'll have syncing this.

PR #5911, opened 2016-11-28
Fix loading of libhdfs.so when installed in non-standard directory

skavulya:
Some Hadoop distributions do not install libhdfs.so in $HADOOP_HDFS_HOME/lib/native. For example, Cloudera Hadoop installs libhdfs.so in /opt/cloudera/parcels/CDH/lib64/libhdfs.so.

This patch allows the dynamic loader to search for libhdfs.so if it fails to find it in $HADOOP_HDFS_HOME/lib/native


jhseu (approved):

jhseu:
Jenkins, test this please

PR #5926, opened 2016-11-29
Replace parser with FLAGS

DjangoPeng:


tensorflow/examples/tutorials/mnist/mnist_softmax.py
```diff
@@ -23,16 +23,19 @@
 from __future__ import print_function
 
 import argparse
```
caisq: Remove this import?

caisq (changes requested):

vrv:
Sorry, we are trying to move away from flags library

PR #5937, opened 2016-11-29
fixed run_path_pairs AttributeError

drcege:
In python3,  `dict.items()` is a little different.
```
D.items() -> a set-like object providing a view on D's items
```
Otherwise, `AttributeError: 'dict_items' object has no attribute 'append'` will be raised.

### DIFF
--- a/tensorflow/tensorboard/plugins/projector/plugin.py
+++ b/tensorflow/tensorboard/plugins/projector/plugin.py
@@ -136,7 +136,7 @@ class ProjectorPlugin(TBPlugin):
   @property
   def configs(self):
     """Returns a map of run paths to `ProjectorConfig` protos."""
-    run_path_pairs = self.run_paths.items()
+    run_path_pairs = list(self.run_paths.items())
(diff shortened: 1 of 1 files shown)


tensorflow/tensorboard/plugins/projector/plugin.py
```diff
@@ -136,7 +136,7 @@ def get_plugin_handlers(self, run_paths, logdir):
   @property
   def configs(self):
     """Returns a map of run paths to `ProjectorConfig` protos."""
-    run_path_pairs = self.run_paths.items()
+    run_path_pairs = list(self.run_paths.items())
```
mrry: Instead of converting to a `list` here, it might be more readable to remove the use of `list.append()` and replace L143 with:

```python
      run_path_pairs = [('.', self.logdir)]
```

martinwicke:
ping?

mrry:
Looks like this fix has already been applied in 0cbf1f0955af2ea9a2f106c4f82460f7531d9ed1, so we can close this PR.

PR #5943, opened 2016-11-29
R0.12

roseyyan:


tensorflow/python/ops/logging_ops.py
```diff
@@ -149,6 +160,11 @@ def image_summary(tag, tensor, max_images=3, collections=None, name=None):
   return val
 
 
+@deprecated(
+    "2016-11-30", "Please switch to tf.summary.audio. Note that "
+    "tf.summary.histogram uses the node name instead of the tag. "
```
rversaw: Should this be `tf.summary.audio`? Same for `tf.summary.image` up above.

danmane: Good catch. Looksl ike this branch is just bringing all the r12 changes into master, so maybe I should fix this in a separate PR to master.

PR #5954, opened 2016-11-29
Enable sandboxed builds for tensorflow.

gunan:


gunan:
@martinwicke on MacOS sandboxed builds are taking much slower and timing out.
Any idea what is happening there?

martinwicke:
@damienmg Any idea why sandboxed builds would be slower than standalone builds on MacOS?

Context: After this fix we want to enable sandboxing on our test machines again, to prevent our dependencies from rotting, and also in preparation for distributed builds.

martinwicke:
@gunan I suppose that does mean we cannot enable them on MacOS just yet. We could just enable them for Linux assuming they're not slower there. The dependencies would be assured that way.

damienmg:
@hermione451 can explain a bit more but with sandboxing on OS X we need to set-up a lot of hard-links and that can be very slow. According to @philwo, you should not see the regression on Linux.

philwo:
Linux sandboxing has currently one known performance issue left, which is only relevant though if your actions have lots (thousands) of input files each. An improvement for that is in code review (https://bazel-review.googlesource.com/#/c/7538/) at the moment.

Regarding macOS sandboxing, I haven't had time myself to look into why it is slow.

gunan:
Looking at the logs closer, Linux GPU is both massively slow, and cannot handle GPU libraries.
Is it known on bazel side?
@damienmg @philwo

gunan:
@damienmg @phillwo Ping!

Any idea why Linux GPU build took more than twice it usually takes?

philwo:
Hi @gunan,

sorry for the delay! I will try to help you out. :) But I don't know where to get the build time data / logs regarding your builds and what to compare them to (what does "twice" mean? Also, I only see a failed build for "Linux GPU", so how can we compare performance to a usually succeeding build here?).

Here are my best guesses:

1) Regarding the build failure: The only thing I can see related to something called "Linux GPU" is a failed build mentioned above in the GitHub merge thingy. There is an error in the console log of that build:

process-wrapper: execvp("external/local_config_cuda/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc", ...): No such file or directory

(from https://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/2801/consoleFull)

My guess it that this Skylark file: https://github.com/tensorflow/tensorflow/blob/48fb73a1c94ee2409382225428063d3496dc651e/third_party/gpus/cuda_configure.bzl somehow doesn't add that "clang/bin/crosstool_wrapper_driver_is_not_gcc" file to the input files required when building with Cuda. This means it won't be included in the sandbox for actions using Cuda, but will appear to work just fine when not using the standalone strategy. The fix is to add that file (+ its dependencies) to the Skylark repository you're building in that script.

2) Regarding the build performance, if you have many input files and a slow disk with high latency (and cloud based VMs have in my experience a ridiculously high disk latency when compared to real hardware, even when they're using "SSDs") then you will very probably hit the currently only known performance issue with the sandbox, which is that you're limited by your disk / filesystem when building the symlink tree. The long-term fix for this will very probably involve using FUSE to generate the sandbox directory without actually doing any I/O. A short-term fix is something I'm currently working on. I hope I can get it done this or next week. From benchmarks run by one of our users who suffered from a similar issue, this fully resolves the performance problem.

Hope that helps,
Philipp

drpngx:
@gunan close?

gunan:
let's close, and Ill reopen once I address the issues.

yifeif:
Digging this up :).

Tried sandboxed build and I'm still seeing both issues:
(1) Build failure with execvp("external/local_config_cuda/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc", ...): No such file or directory. 
@philwo do you mind clarify "add that file to the Skylark repository you're building in that script"? This file has already been added to https://github.com/tensorflow/tensorflow/blob/48fb73a1c94ee2409382225428063d3496dc651e/third_party/gpus/crosstool/CROSSTOOL.tpl#L171. @davidzchen helped taking a look and we are not sure where else we can add it to make it available since it is part of the toolchain.

(2) Sandboxed build is still slow on mac. It took about twice as long as our normal mac build, and we are using bazel 0.4.4. https://ci.tensorflow.org/job/tensorflow-cl-mac/16842/. @philwo is the fix you mentioned available and how do we use it? 

Appreciate the help!

damienmg:
For (1) you need to add the wrapper to the generated cc_toolchain rule in the all_files attribute I believe (cc_configure do that in Bazel).

PR #5958, opened 2016-11-29
[Windows] Re-enable the special function ops.

mrry:
These were previously disabled under `#ifdef EIGEN_HAS_C99_MATH` due to a compiler error, which has since disappear. Re-enable these ops and their test.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -8,9 +8,7 @@
   acceleration. Known limitations include: It is not currently possible to load
   a custom op library. The GCS and HDFS file systems are not currently
   supported. The following ops are not currently implemented:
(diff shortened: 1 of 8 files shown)


mrry:
@gunan This is for the `r0.12` branch... would it be alright to include these (relatively obscure) ops in the final release, and is it alright to retrospectively modify the release notes before the final release is cut?

gunan:
I will tentatively accept this.
Do you think it will cause problems with any other builds?

gunan (approved):

mrry:
Thanks! Its effect should be limited to the Windows builds, since the only effect of the `#ifdef`s was to disable building those kernels with MSVC.

PR #5826, opened 2016-11-24
[Windows/CMake] Some more C++ test fixes

vit-stepanovs:
A new batch of fixes to be able to build some C++ tests on Windows.

PR #5961, opened 2016-11-30
Run cc tests on Windows with Bazel

meteorcloudy:
@gunan @dslomov @damienmg 

To filter out all the broken and failing tests on Windows, we add "fail_no_windows" tag for them.

### DIFF
--- a/tensorflow/core/platform/default/build_config.bzl
+++ b/tensorflow/core/platform/default/build_config.bzl
@@ -152,7 +152,16 @@ def tf_additional_test_deps():
   return []
 
 def tf_additional_test_srcs():
-  return ["platform/default/test_benchmark.cc", "platform/posix/test.cc"]
+  return [
+      "platform/default/test_benchmark.cc",
+  ] + select({
+      "//tensorflow:windows" : [
+          "platform/windows/test.cc"
+        ],
+      "//conditions:default" : [
+          "platform/posix/test.cc",
+        ],
+    })
 
 def tf_kernel_tests_linkstatic():
   return 0
--- a/tensorflow/tools/ci_build/windows/cpu/bazel/run_cc_test_windows.sh
+++ b/tensorflow/tools/ci_build/windows/cpu/bazel/run_cc_test_windows.sh
@@ -1,0 +1,140 @@
+#!/bin/bash
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+#
+# This script assumes the standard setup on tensorflow Jenkins windows machines.
+# It is NOT guaranteed to work on any other machine. Use at your own risk!
+#
+# REQUIREMENTS:
+# * All installed in standard locations:
+#   - JDK8, and JAVA_HOME set.
+#   - Microsoft Visual Studio 2015 Community Edition
+#   - Msys2
+#   - Anaconda3
+# * Bazel windows executable copied as "bazel.exe" and included in PATH.
+
+# All commands shall pass, and all should be visible.
+set -x
+set -e
+
+# This script is under <repo_root>/tensorflow/tools/ci_build/windows/cpu/bazel
+# Change into repository root.
+script_dir=$(dirname $0)
+cd ${script_dir%%tensorflow/tools/ci_build/windows/cpu/bazel}.
+
+# Setting up the environment variables Bazel needs
+source "tensorflow/tools/ci_build/windows/cpu/bazel/common_env.sh" \
+  || { echo "Failed to source common_env.sh" >&2; exit 1; }
+
+# bazel clean --expunge doesn't work on Windows yet.
+# Clean the output base manually to ensure build correctness
+bazel clean
+output_base=$(bazel info output_base)
+bazel shutdown
+# Sleep 5s to wait for jvm shutdown completely
+# otherwise rm will fail with device or resource busy error
+sleep 5
+rm -rf ${output_base}
+
+export TF_NEED_CUDA=0
+echo "" | ./configure
+
+failing_tests="\
+    //tensorflow/core:example_example_parser_configuration_test + \
+    //tensorflow/core/kernels:sparse_dense_binary_op_shared_test + \
+    //tensorflow/core/kernels:sparse_reduce_sum_op_test + \
+    //tensorflow/core:lib_core_status_test + \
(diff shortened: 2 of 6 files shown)


vit-stepanovs (commented):
FYI, I have fixed some of the tests listed in FAILING_CC_TESTS/BROKEN_CC_TESTS recently on Windows in PRs #5826 and #5739. They build and run successfully with CMake currently.

tensorflow/tensorflow.bzl
```diff
@@ -48,6 +48,10 @@ load(
     "if_cuda",
     "cuda_path_flags"
 )
+load(
```
gunan: nit: fits in one line?

meteorcloudy:
@vit-stepanovs Thanks for reminding me that! I do have a list of tests which run with CMake but not with Bazel, I'll look into them.

tensorflow/windows.bzl
```diff
@@ -0,0 +1,84 @@
...
+  "//tensorflow/core:platform_profile_utils_cpu_utils_test",
+  "//tensorflow/core:platform_port_test",
+  "//tensorflow/core:lib_strings_strcat_test",
+  "//tensorflow/core:lib_jpeg_jpeg_mem_unittest",
+  "//tensorflow/core:lib_core_notification_test",
+  "//tensorflow/core:framework_partial_tensor_shape_test",
+  "//tensorflow/core/debug:debug_io_utils_test",
+]
+
+def if_fail_on_windows(test_name):
```
gunan: I have bad feelings about this construct, but I am not sure how to do things better.
Maybe we can have the list live in a bash or a python file, and not use bazel query for constructing the test list to run?
Or maybe even manually exclude them directly in bazel test run?

meteorcloudy:
@gunan I think you are right, I removed all the logic of adding tags to the test targets, instead, we filter them out in a shell script for test. PTAL.

tensorflow/tools/ci_build/windows/cpu/bazel/run_cc_test_windows.sh
```diff
@@ -0,0 +1,140 @@
...
+# This script is under <repo_root>/tensorflow/tools/ci_build/windows/cpu/bazel
+# Change into repository root.
+script_dir=$(dirname $0)
+cd ${script_dir%%tensorflow/tools/ci_build/windows/cpu/bazel}.
+
+# Setting up the environment variables Bazel needs
+source "tensorflow/tools/ci_build/windows/cpu/bazel/common_env.sh" \
+  || { echo "Failed to source common_env.sh" >&2; exit 1; }
+
+# bazel clean --expunge doesn't work on Windows yet.
```
gunan: unfortunately, for now yes.
for full details, @damienmg and @davidzchen can elaborate.
I think this is because of some bugs in cache invalidation?

tensorflow/tools/ci_build/windows/cpu/pip/build_tf_windows.sh
```diff
@@ -32,28 +32,11 @@ set -e
 # This script is under <repo_root>/tensorflow/tools/ci_build/windows/cpu/pip/
 # Change into repository root.
 script_dir=$(dirname $0)
-cd ${script_dir%%tensorflow/tools/ci_build/windows/cpu/pip}
+cd ${script_dir%%tensorflow/tools/ci_build/windows/cpu/pip}.
```
gunan: extra "." at the end?

tensorflow/tools/ci_build/windows/cpu/pip/build_tf_windows.sh
```diff
@@ -32,28 +32,11 @@ set -e
...
-
-# Set Visual Studio path
-export BAZEL_VS="C:/Program Files (x86)/Microsoft Visual Studio 14.0"
-
-# Add python into PATH, it's needed because gen_git_source.py uses
-# '/usr/bin/env python' as a shebang
-export PATH="/c/Program Files/Anaconda3:$PATH"
-
-export TF_NEED_CUDA=0
+# Setting up the environment variables Bazel needs
```
gunan: Nit.
bazel and ./configure script needs.

tensorflow/core/platform/default/build_config.bzl
```diff
@@ -152,7 +152,16 @@ def tf_additional_test_deps():
...
 def tf_additional_test_srcs():
-  return ["platform/default/test_benchmark.cc", "platform/posix/test.cc"]
+  return [
+      "platform/default/test_benchmark.cc",
+  ] + select({
+      "//tensorflow:windows" : [
+          "platform/windows/test.cc"
+        ],
+      "//conditions:default" : [
+          "platform/posix/test.cc",
```
gunan: @vit-stepanovs @guschmue @mrry did you also run into a problem with this file?
Are we skipping posix/test.cc in our cmake build, too?

meteorcloudy: windows/test.cc file is introduced in https://github.com/tensorflow/tensorflow/commit/91359470eb2c940733497f2bc6b6e23e3b8d76ef
So I guess it is needed by cmake?

tensorflow/tools/ci_build/windows/cpu/pip/build_tf_windows.sh
meteorcloudy: Ha, that's because if we run this script under TF source root.
It would be 
`cd ${tensorflow/tools/ci_build/windows/cpu/pip%%tensorflow/tools/ci_build/windows/cpu/pip}`
, which is essentially `cd`, then it will just jump to your home directory.

gunan: no, I mean there is now an extra dot (.) at the end of the line.
That is probably not needed.

tensorflow/core/platform/default/build_config.bzl
meteorcloudy: I think the same logic in CMake build is here:https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/tf_tests.cmake#L204

tensorflow/tools/ci_build/windows/cpu/pip/build_tf_windows.sh
meteorcloudy: Oh, the extra dot is for addressing the mentioned problem. We don't want to jump to home directory if we run the script under TF source root.
If we add the extra dot, it would be `cd .`, which stays at the source root.

gunan (approved):

gunan:
Jenkins, test this please

gunan:
Jenkins is not updating the review thread, but the test is running here:
http://ci.tensorflow.org/job/tensorflow-pull-requests-multijob/2880/

gunan:
I think the failure is just a flake.
Jenkins, test this please.

gunan:
Tests seem to be flaky, as different ones failed each time.
Merging.

tensorflow/tools/ci_build/windows/cpu/bazel/run_cc_test_windows.sh
davidzchen: Yes, unfortunately, we need this for now. I have a more detailed explanation for this in #5253.

meteorcloudy: I see, thanks!

PR #5979, opened 2016-11-30
Translate.py - Added the option to translate a file 

rmeertens:
With this pull request you can easily translate/test a batch of sentences, which makes it easier to see how good your trained model is. 

If you specify a filename this functionality is used, if you don't specify anything the script continues to function as it used to.


rmeertens:
I signed it!

martinwicke:
Can you move this PR over to github.com/tensorflow/models? We have moved the models to that repo. You can call out @nealwu for a review.

PR #6006, opened 2016-12-01
split will move leading or trailing whitespace when sep is whitespace

unship:
split will move leading or trailing whitespace when sep is whitespace, so strip() is useless

If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace. Consequently, splitting an empty string or a string consisting of just whitespace with a None separator returns [].

see https://docs.python.org/3/library/stdtypes.html?highlight=split#str.split


andrewharp:

martinwicke:
@biolee can you fix the conflicts?

gunan:
AFAICS this file is not in tensorflow/tensorflow repository anymore.
Please create a new PR to tensorflow/models repository for this change.

PR #6009, opened 2016-12-01
Use correct variable scope for weights in rnn_cell._linear.

darrengarvey:
If a scope is passed in it should be used for the variable scope
rather than always falling back to the default scope.

This change in behaviour was a consequence of recent cleanup in
92da8abf. The commit may alter the name of "weights" variable in saved
checkpoints but is necessary if two `rnn_cell._linear` layers exist in
the same parent variable scope.


darrengarvey:
@ebrevdo - I guess you might be interested in this one since I don't think you intended to change the behaviour here (it breaks some uses of `_linear`, even though that may not be part of the public API atm).

ebrevdo:
It is hidden for a reason; you should not be using it.  Use tf.contrib.layers.linear instead.  The scope argument will be removed soon.

PR #6013, opened 2016-12-01
Fix incorrect version string in gcs_test/Dockerfile

caisq:
Also avoid hard-coding version string in the Dockerfile.
The pip wheel is now specified as an argument to gcs_smoke.sh.


caisq:
Passed experimental build (viewing requires jenkins login):
http://ci.tensorflow.org/view/Experimental/job/experimental-cais-gcs-smoke/1/console

gunan (approved):

PR #6014, opened 2016-12-01
[loss_op]Denote the returned value is already mean, it happens quite a few tim…

zuoxingdong:
…es that we often add tf.reduce_mean after that since tf.nn.sparse... returns loss without averaging it.

### DIFF
--- a/tensorflow/contrib/losses/python/losses/loss_ops.py
+++ b/tensorflow/contrib/losses/python/losses/loss_ops.py
@@ -413,7 +413,7 @@ def softmax_cross_entropy(
     weight: Deprecated alias for `weights`.
 
   Returns:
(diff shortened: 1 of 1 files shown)


martinwicke:
Can you make this change to python/ops/losses/loss.py as well? And for both the sparse and non-sparse versions?

zuoxingdong:
@martinwicke Since I didn't manage to find python/ops/losses/losses.py in the patch for this pull request, so I created a new patch with a new request #6359 for this additional change.

martinwicke:
Ok, I'll close this one then.

martinwicke:
Oh wait, I see. Can you add the content of this PR to the other one? It'll be cleaner that way.

PR #6016, opened 2016-12-01
fix deprecation date

sugyan:


danmane:
Thanks for the change. Can you revert the markdown changes, those are autogenerated from the source file.
@gunan do we want to merge this into r0.12? if so do we also need to get it separately into master?

gunan:
Thanks for bringing this to my attention.
I agree with all your comments @danmane 
Markdown changes should be automated, so should be rolled back,
Yes, we want this in 0.12
And yes, we want this also separately merged into master. My preference is to merge it to master before 0.12, but I am not picky about the order.

gunan:
Jenkins, test this please.

danmane:
Given that this is such a trivial change, I feel OK just merging it in.

PR #6031, opened 2016-12-01
Use gather on floats instead of int32 to keep the kernel on GPU when possible.

### LINKED ISSUE #6027: tracing batch normalization: apparently unnecessary gpu-cpu-gpu transfer for gather op.
I am running:
CUDA 8.0. CUDNN 5.1. tensorflow 0.11. Ubuntu 12.04.

I was experiencing low GPU utilization. As per #1824 I did a trace and found many instances of the following interaction, as shown by the screenshot in chrome:

The op I have highlighted (with arrows going in and out in the screenshot) is a `gather` op. I provide a link to the context in which it is called.
https://github.com/tensorflow/tensorflow/blob/eb56a8af24695bf8258addf28b0c53fbabff72e6/tensorflow/python/ops/nn_impl.py#L472

The call stack (at graph construction) time is:
`tf.contrib.layers.batch_norm`, `python.ops.nn_impl.moments`, `python.ops.nn_impl.sufficient_statics`

If I understand correctly, I am seeing that the GPU gets blocked as a result of shuffling a trivial `gather` op unnecessarily off to the CPU only to wait forever for it to come back. 

screenshot from 2016-12-01 11 45 17

vincentvanhoucke:
gather() on an int32 type forces the op to go to CPU.
Fixes #6027
@liuyipei can you confirm that this addresses what you observe?

### DIFF
--- a/tensorflow/python/ops/nn_impl.py
+++ b/tensorflow/python/ops/nn_impl.py
@@ -463,7 +463,7 @@ def sufficient_statistics(x, axes, shift=None, keep_dims=False, name=None):
   with ops.name_scope(name, "sufficient_statistics", [x, shift]):
     x = ops.convert_to_tensor(x, name="x")
     x_shape = x.get_shape()
-    if x_shape.is_fully_defined():
+    if all(x_shape[d].value is not None for d in axes):
       counts = 1
       for d in axes:
         counts *= x_shape[d].value
--- a/tensorflow/python/ops/nn_impl.py
+++ b/tensorflow/python/ops/nn_impl.py
@@ -469,9 +469,9 @@ def sufficient_statistics(x, axes, shift=None, keep_dims=False, name=None):
         counts *= x_shape[d].value
       counts = constant_op.constant(counts, dtype=x.dtype)
     else:  # shape needs to be inferred at runtime.
-      x_dims = array_ops.gather(array_ops.shape(x), axes)
-      counts = math_ops.cast(
-          math_ops.reduce_prod(x_dims), x.dtype, name="count")
+      x_dims = array_ops.gather(
+          math_ops.cast(array_ops.shape(x), x.dtype), axes)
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/nn_impl.py
```diff
@@ -469,9 +469,9 @@ def sufficient_statistics(x, axes, shift=None, keep_dims=False, name=None):
         counts *= x_shape[d].value
       counts = constant_op.constant(counts, dtype=x.dtype)
     else:  # shape needs to be inferred at runtime.
-      x_dims = array_ops.gather(array_ops.shape(x), axes)
-      counts = math_ops.cast(
-          math_ops.reduce_prod(x_dims), x.dtype, name="count")
+      x_dims = array_ops.gather(
+          math_ops.cast(array_ops.shape(x), x.dtype), axes)
```
vrv: gather's second argument can't be a float, it's an index.

vincentvanhoucke:
Not sure why this one test fails. Works fine locally.

tensorflow/python/ops/nn_impl.py
```diff
@@ -469,9 +469,9 @@ def sufficient_statistics(x, axes, shift=None, keep_dims=False, name=None):
         counts *= x_shape[d].value
       counts = constant_op.constant(counts, dtype=x.dtype)
     else:  # shape needs to be inferred at runtime.
```
mrry: Not sure if it would help with the original issue, but it might also be worth making L466 more permissive. e.g. Instead of:

```
if x_shape.is_fully_defined():
```

...we could instead have:

```
if all(x_shape[d].value is not None for d in axes):
```

vincentvanhoucke: Seems like a no brainer. Done

mrry (approved):
LGTM!

Issue #6036, opened 2016-12-02
File object not being closed properly in 6_lstm.ipynb


ronrest:
The [**6_lstm.ipynb**](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/udacity/6_lstm.ipynb) file contains the following code: 

```python
def read_data(filename):
  f = zipfile.ZipFile(filename)
  for name in f.namelist():
    return tf.compat.as_str(f.read(name))
  f.close()
```

This code has the following issues: 

- The file object `f` does not get to close properly, since the function is 
  exited before the `f.close()` method ever gets to be called. 
- The `for` loop is uneccesary, since there is only one item in the zip file, 
  *and* since the `return` operator forces the loop to end on the first 
  loop anyway. 
  
I propose the following code instead: 

```python
def read_data(filename):
  with zipfile.ZipFile(filename) as f:
    name = f.namelist()[0]
    data = tf.compat.as_str(f.read(name))
  return data
```

This code fixes the issues in the following ways: 

- The `with` statement will handle the file `f` being closed automatically, even 
  if there is an unexpected early termination. 
- Since there is actually only one item in the zip file, then we can just get 
  the zeroeth element in `f.namelist()`

I will create a pull request with this suggested code shortly.

PR #6038, opened 2016-12-02
Fix issue #6036

ronrest:
As per the issue raised [here](https://github.com/tensorflow/tensorflow/issues/6036), changed the following code to the  [**6_lstm.ipynb**](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/udacity/6_lstm.ipynb) file.

**FROM:**
```python
def read_data(filename):
  f = zipfile.ZipFile(filename)
  for name in f.namelist():
    return tf.compat.as_str(f.read(name))
  f.close()
```

**TO:**

```python
def read_data(filename):
  with zipfile.ZipFile(filename) as f:
    name = f.namelist()[0]
    data = tf.compat.as_str(f.read(name))
  return data
```

### DIFF
--- a/tensorflow/examples/udacity/6_lstm.ipynb
+++ b/tensorflow/examples/udacity/6_lstm.ipynb
@@ -167,10 +167,10 @@
       },
       "source": [
         "def read_data(filename):\n",
-        "  f = zipfile.ZipFile(filename)\n",
-        "  for name in f.namelist():\n",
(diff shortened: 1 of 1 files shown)


vincentvanhoucke (approved):

danmane:
AFAIK we have no ipynb tests, so I will just merge this.

PR #6049, opened 2016-12-02
Fix matmul() 3-D tensor example code

ghost:
It is necessary to specify an int32 dtype because the code as written
would use int64 dtype by default, and int64 is not currently supported
by matmul().

The "2-D tensor" example code is fine.

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -1682,13 +1682,15 @@ def matmul(a,
 
 
   # 3-D tensor `a`
-  a = tf.constant(np.arange(1,13), shape=[2, 2, 3]) => [[[ 1.  2.  3.]
+  a = tf.constant(np.arange(1, 13, dtype=np.int32),
+                  shape=[2, 2, 3])                  => [[[ 1.  2.  3.]
                                                          [ 4.  5.  6.]],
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/math_ops.py
```diff
@@ -1682,18 +1682,18 @@ def matmul(a,
 
 
   # 3-D tensor `a`
-  a = tf.constant(np.arange(1,13), shape=[2, 2, 3]) => [[[ 1.  2.  3.]
-                                                         [ 4.  5.  6.]],
-                                                        [[ 7.  8.  9.]
-                                                         [10. 11. 12.]]]
+  a = tf.constant(np.arange(1,13,dtype=np.int32), shape=[2, 2, 3]) => [[[ 1.  2.  3.]
```
martinwicke: Can you please break these to 80 characters?

martinwicke: Simplest by adding a newline before shape, I think.

martinwicke (changes requested):

tensorflow/python/ops/math_ops.py
dtrebbien: Done

martinwicke (approved):

PR #6050, opened 2016-12-02
Run python tests on Windows with Bazel

meteorcloudy:
The major problem with running TF test on Windows is python tests depend on too many and too big things (eg. pywrap_tensorflow.dll). Since py_test is a zip file on Windows, it's too slow to unzip for running them concurrently.

We solve this problem by getting rid of all the deps of py_test, so that every py_test zip only contains it's source code(several of them also contains some data files). Then we install tensorflow pip package into the system before running them.

By doing this we can get most of the python test running on Windows, but the disadvantage is the tests are not hermetic in Bazel.
@gunan


gunan:
Jenkins, test this please.

gunan:
Jenkins, test this please.

gunan:
Jenkins is failing to fetch our repo, great.
Trying again.
Jenkins, test this please.

gunan:
For good measure, one more.
Jenkins, test this please.

gunan:
@caisq @martinwicke @yifeif looks like this is a solution we can generalize for all our pip tests.
This is creating py test rules with to dependencies, so we can run these with bazel test with these rules after doing pip install.
What do you think?

martinwicke:
This is very cool, I'm all for it. Is redefining py_test an ok thing to do? @caisq FYI.

@meteorcloudy does this rely on the new and improved sandboxing behavior (allowing access to the installed system by not undeclared dependencies in the source)?

caisq:
@wicke @gunan @meteorcloudy Cool. I think it's definitely worth a shot to replace the pip tests with this. If it works, it'll bring benefits like better concurrency, better timeout and incrementality.

gunan:
I think sandboxing can complicate things here.
What happens is, we build tensorflow pip package, we pip install that package, and blaze test these (in the virtual environment we installed the pip package). So the test has to have access to the python pip packages installed. Would sandboxing prevent access to those?

meteorcloudy:
Actually, I don't think sandbox is a problem for us. Because sandboxing only happens at build time, we don't need the pip package at build time, and there is no sandbox when running the test.

I tried to use this way to run all the py_test under `//tensorflow/python/...`, only 2 of them are failing because they depend on something not included in the pip package. And they are also excluded in [test_installation.sh](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/builds/test_installation.sh)
```
//py_test/tensorflow/python:framework_ops_test
//py_test/tensorflow/python:protobuf_compare_test
```

meteorcloudy:
Sorry, what I said is wrong. I confirmed with our sandbox expert, there is sandboxing when running tests. But the system installed packages will still be accessible...

caisq:
@meteorcloud Thanks for looking into it. It's probably going to work, but it might be good to double check that virtualenv-installed packages work too.

meteorcloudy:
@caisq you are welcome, I was actually running under virtualenv, and it worked. I didn't specify any sandbox flag, so I believe sandbox is also enabled by default.

meteorcloudy:
Jenkins, test this please.

meteorcloudy:
@gunan We just had [Bazel 0.4.2 release](https://github.com/bazelbuild/bazel/releases), after this is merged, I'll send the scripts for GPU tests, so that you can set up the CI jobs!

gunan (approved):

gunan:
I think the failures are all known, but just in case.
Jenkins, test this please.

gunan:
Jenkins, test this please.

Linux CPU and GPU issues are known. Windows cmake should be fixed.

gunan:
Results as expected, only known issues remaining. merging changes.

PR #6052, opened 2016-12-03
Fix tutorial docs

bkeyvani:
Dear TensorFlow maintainers,

I noticed a missing closing `)` for the `tf.constant` function in `r0.12` and `r0.11` of `tensorflow/g3doc/tutorials/input_fn/index.md`. This is a fix for the `r0.12` docs.

Should I submit a separate PR for the `r0.11` documentation too, or will this suffice?

Thanks


bkeyvani:
Sorry for the confusion. I noticed the issue in `r0.12` and `r0.11` tags, but it seems that another user has already sent a patch for this and the issue was resolved in the `master` branch. 

my confusion arose from the fact that the tutorial page seems to default to the latest `r0.` version instead of `master`.

Please ignore/close my PR.

PR #6060, opened 2016-12-04
fix the number on ImageNet classification validation error

fayeshine:
According to the three paper, Inception(https://arxiv.org/pdf/1409.4842v1.pdf), Inception V2(https://arxiv.org/pdf/1502.03167v3.pdf), Inception V3(https://arxiv.org/pdf/1512.00567v3.pdf).
Their ImageNet Classification Validation Error should be 6.67%, 4.9%, 3.46%。


fayeshine:
@vrv

vincentvanhoucke (approved):

PR #6074, opened 2016-12-05
Change base datasets dtype default to float32

terrytangyuan:
So this following warning can be avoided in many examples:

`WARNING:tensorflow:float64 is not supported by many models, consider casting to float32.`


terrytangyuan:
Aborting this since it seems like a lot of tests rely on this. Not important change.

PR #6089, opened 2016-12-05
Add missing parenthesis

vincentvanhoucke:


ghost (commented):
I don't think that this is right. The existing right parenthesis is matched with the left parenthesis of "(e.g., we ...".

PR #6123, opened 2016-12-06
Disable contrib/integrate/python/ops/odes_test due to timeout

caisq:
The test hangs during PIP tests on GPU (but not CPU).


gunan:
Do we have a corresponding bug opened about this?

caisq:
@gunan bug opened internally for @shoyer

gunan (approved):

caisq:
DO NOT MERGE this for now. I'm seeing some indication that this test doesn't hang when the tests are run serially, instead of concurrently.

caisq:
nvm. the test times out even when run exclusively.

PR #6127, opened 2016-12-06
Clarify about 'SAME' padding in docs

### LINKED ISSUE #5203: Explanation about 'SAME' padding doesn't cover edge cases
In the [docs](https://www.tensorflow.org/versions/r0.11/api_docs/python/nn.html#convolution), it explains the padding as follows:
```
For the 'SAME' padding, the output height and width are computed as:

out_height = ceil(float(in_height) / float(strides[1]))
out_width  = ceil(float(in_width) / float(strides[2]))

and the padding on the top and left are computed as:

pad_along_height = ((out_height - 1) * strides[1] +
                    filter_height - in_height)
pad_along_width = ((out_width - 1) * strides[2] +
                   filter_width - in_width)
pad_top = pad_along_height / 2
pad_left = pad_along_width / 2
```
When `in_height=20, stride=2, filter_height=1`, we will end up with `pad_along_height=-1`. I think it's unclear what this means (is there padding or not?). Further explanation are needed to document the behavior in this case.

Also, when `in_height=19, stride=2, filter_height=1`, we will end up with `out_height=10, pad_along_height=0, pad_top=0`.  But with 19 inputs and stride=2 how can you have 10 output without padding? This doesn't seem right.

ppwwyyxx:
Fix #5203. Also use `//` instead of `/` to be consistent with Python2/3.

### DIFF
--- a/tensorflow/python/ops/nn.py
+++ b/tensorflow/python/ops/nn.py
@@ -68,12 +68,15 @@ For the `'SAME'` padding, the output height and width are computed as:
 
 and the padding on the top and left are computed as:
 
-    pad_along_height = ((out_height - 1) * strides[1] +
-                        filter_height - in_height)
-    pad_along_width = ((out_width - 1) * strides[2] +
-                       filter_width - in_width)
-    pad_top = pad_along_height / 2
-    pad_left = pad_along_width / 2
+    pad_along_height = max((out_height - 1) * strides[1] +
+                        filter_height - in_height, 0)
+    pad_along_width = max((out_width - 1) * strides[2] +
+                       filter_width - in_width, 0)
+    pad_top = pad_along_height // 2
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/nn.py
```diff
@@ -68,12 +68,15 @@
 
 and the padding on the top and left are computed as:
```
martinwicke: Can you add a ```python block here? It's better than the implicit indent block.

martinwicke (changes requested):

tensorflow/python/ops/nn.py
martinwicke: Actually, never mind.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

PR #6130, opened 2016-12-06
Adding extra options to Kubernetes object generator for distributed example.

elsonrodriguez:
This enables multiple training clusters to be deployed in the same namespace.

### DIFF
--- a/tensorflow/tools/dist_test/scripts/k8s_tensorflow.py
+++ b/tensorflow/tools/dist_test/scripts/k8s_tensorflow.py
@@ -32,6 +32,8 @@ import sys
 
 DEFAULT_DOCKER_IMAGE = 'tensorflow/tf_grpc_test_server'
 DEFAULT_PORT = 2222
+DEFAULT_CLUSTER_NAME = 'cluster1'
+DEFAULT_CREATOR= 'default'
 
 # TODO(cais): Consider adding resource requests/limits to the pods.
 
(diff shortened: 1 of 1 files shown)


tensorflow/tools/dist_test/scripts/k8s_tensorflow.py
```diff
@@ -190,15 +232,19 @@ def main():
...
+                               args.num_workers,
                                args.num_parameter_servers,
                                args.grpc_port,
                                args.request_load_balancer,
                                args.docker_image)
   print(yaml_config)  # pylint: disable=superfluous-parens
 
 
-def GenerateConfig(num_workers,
+def GenerateConfig(cluster_name,
```
caisq: Can you make this change backward compatible by letting cluster_name and creator be keyward arguments that have None values? So if some user doesn't want to explicitly set the arguments, they can just ignore them, and the pods, rcs and services will have the same naming pattern as before, i.e., without the prefixes like "cluster1-"?

tensorflow/tools/dist_test/scripts/k8s_tensorflow.py
```diff
@@ -127,33 +151,51 @@
...
+    ts-cluster-name: "{cluster_name}"
 """)
 
 
 def main():
   """Do arg parsing."""
   parser = argparse.ArgumentParser()
+  parser.add_argument('--cluster_name',
+                      type=str,
+                      default=DEFAULT_CLUSTER_NAME,
```
caisq: Same as below. Can you please make it backward compatible? Let the default be an empty string, which GenerateConfig() can ignore, which leads to the same naming pattern as before. Thanks!

caisq (changes requested):

jhseu:
No opinion about whether to merge, but I think these scripts should be used purely for testing and not a demonstration of how to run distributed TF :)

caisq:
@jhseu Agreed. These scripts should be treated as they are, i.e., tests (they live in the folder called dist_test), not examples, even though some users might find bits and pieces of useful.

jhseu:
Closing for now as these changes aren't needed for testing. @caisq feel free to reopen if you disagree.

elsonrodriguez:
@caisq @jhseu The hardcoding of the objects into the generator script is definitely a shortcoming, and the data is baked into the containers, but aside from that what are the other shortcomings of this test?

I ask because I'm totally new to tensorflow, and was trying to prototype a kubernetes third party resource to help developers quickly create training clusters: https://github.com/elsonrodriguez/tensorsets

What's a better script/test to base this type of automation on?

PR #6138, opened 2016-12-07
[Windows/CMake] Run ctest in parallel

mrry:
Using 32 parallel jobs, as we're using that degree of parallelism for the build too.


gunan (approved):
wow, this was fast :)

mrry:
21 minutes to run the Windows tests!!

PR #6137, opened 2016-12-07
Branch 141241224

andrewharp:
Pushing internal commits.

PR #6145, opened 2016-12-07
[CMake] Add missing Python packages to build scripts

mrry:
Fixes Windows PIP tests after #6137.


mrry:

mrry:
The only failures are in non-CMake code, so this should be good to fix the Windows build now.

PR #6156, opened 2016-12-07
typo in the code for `tf.contrib.learn` tutorial

avloss:
doing tutorial with copy-paste fails with "accessing `test_data` before assignment"


avloss:
I signed it!

gunan:
@sandersk, could you take a look?

sandersk (approved):
Thanks very much for catching and fixing this discrepancy between the sample code and the excerpt in the tutorial, @avloss.

PR #6161, opened 2016-12-07
Improved support for OpenCL

benoitsteiner:


gunan:
GPU failure is just a known flake.
CPU issue I think I have seen before, but I would rerun to make sure.

Windows issue looks like a legit issue, but maybe it is also broken at head?
@andrewharp I remember this windows problem was seen in the push from google. Do we have a fix ready for it?
Who is working on the fix?

gunan (approved):

PR #6176, opened 2016-12-08
Disable all tests depending on *half_plus_two* genrules.

gunan:
This is a measure to stabilize flaky GPU builds.

This is a fix for the CUDA out of memory error issues we have been seeing.


gunan:
The CPU failures are known issues.
So this should fix the GPU issues.

yifeif (approved):

PR #6185, opened 2016-12-08
Add python server stop method

llhe:
Add support for graceful server stop in python.

### DIFF
--- a/tensorflow/python/training/server_lib.py
+++ b/tensorflow/python/training/server_lib.py
@@ -163,6 +163,16 @@ class Server(object):
     with errors.raise_exception_on_not_ok_status() as status:
       pywrap_tensorflow.PyServer_Start(self._server, status)
 
+  def stop(self):
+    """Stops this server.
+
+    Raises:
+      tf.errors.OpError: Or one of its subclasses if an error occurs while
+        stopping the TensorFlow server.
+    """
+    with errors.raise_exception_on_not_ok_status() as status:
+      pywrap_tensorflow.PyServer_Stop(self._server, status)
+
   def join(self):
     """Blocks until the server has shut down.
 


mrry:
Unfortunately this doesn't actually add the support, since the [implementation](https://github.com/tensorflow/tensorflow/blob/a1d06460f56d7b2f5b7a8317c1aeba6a587b104f/tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc#L224) of `GrpcServer::Stop()` is incomplete. We wouldn't be able to accept this PR until this has been implemented and tested. If I recall correctly, the current shutdown code leaks in-flight buffers, so we would need additional tracking code in the gRPC services to make this work cleanly.

Would you be keen to work on that?

yaroslavvb:
BTW, as a work-around, you can do `queue.dequeue()` in your PS instead of `server.join()`, that'll wait until someone pushes something onto that shared queue queue - [example](http://stackoverflow.com/questions/39810356/shut-down-server-in-tensorflow/40186129#40186129)

llhe:
@mrry Sure, if I can help. Any contexts about the in-flight buffers?

mrry:
IIRC the problem is that the [`Call`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/distributed_runtime/rpc/grpc_call.h#L134) objects used in `GrpcMasterService` and `GrpcWorkerService` have pointers to the underlying gRPC service objects in their `responder_` field. So you need some way to ensure that all of the pending requests have completed/been cancelled before deleting the server itself.

At the time I implemented this, it didn't seem worth the additional complexity to support clean shutdown (which was not a feature anybody was using at the time...). I'd be delighted if there's a simple solution here.

mrry:
IIRC the problem is that the `Call` objects in the servers have non-owned pointers to the service, both directly and through the `responder_` object. A clean shutdown would need some way to wait until all of the pending calls had completed or been cancelled. At the time I implemented this, it didn't seem worth the additional bookkeeping to support this feature (and nobody had explicitly requested it). It would be great if you could come up with a simple solution here!

llhe:
Got it. Yeah, this is not a high priority issue, it's not quite useful in production. May be we can just add an TODO tag in the source:).

Any way, I will have a look about gRPC buffer stuff if no googlers will fix https://github.com/grpc/grpc/issues/8975 because it's a real blocker to us.

martinwicke:
So, @llhe, should I close this PR or did you want to fix the issue?

llhe:
Close it first, I can reopen it when I have free cycles.

mayankchatteron1:
Is it implemented?

PR #6198, opened 2016-12-08
Fully disable all test/build targets that execute tensorflow during build.

gunan:


yifeif (approved):

martinwicke:
Jenkins, test this please?

yifeif:

gunan:
Jenkins, test this please.

gunan:
CPU test failure known issue.

kirilg:
Why do the genrule targets need to be commented out? Shouldn't it be sufficient to just mark the tests manual as was done before? If that's not sufficient, adding `tags = [
        "local",
        "manual",
    ],` will definitely have Jenkins skip them for now until a fix is ready. As is, this breaks TensorFlow Serving and we can't sync to head. Can we revert this change or uncomment the genrules?

gunan:
That makes jenkins skip "test execution", but in this case problem is actually building them.
You can see that prior to this change I tried exactly what you suggested. But jenkins still built them. And the problem in our case is the tensorflow library usage during build of these targets.

kirilg:
The problems mentioned shouldn't be an issue if we just have the genrules though. If no test is linking it in (i.e. keep tests commented out for now), it shouldn't be built. If it's somehow being linked, it will be compiled at most once per run so there should be no path collisions. Do you think that would work?

gunan:
we can try it out, if you have a PR I can try to see if it works.
Another way I see it working is, you can create these genrules within tensorflow serving repository and reference those genrules.
Any reason these specificly have to live within tensorflow repo?

kirilg:
I don't have a PR. We don't do much development in git, I'm more comfortable making this change internally, but I see this is purely in Github. Could someone from the TF team undo the genrule comment change and test? (the usual tests we always run pass so I'm not sure where it failed)

We needed it in the TF repo because the other tests there were commented out here depended on it.

gunan:
It causes a flake, which is much harder to track, and thats why it triggered our response.

What do you mean other tests commented out here, they are commented out, so they should not run. You mean tests not under third_party/tensorflow?

Again, if you have any tests in your repo, you can easily create the same genrule, and modify your tests to depend on those. This genrule is definitely going away from tensorflow, I can guarantee that, so the sooner your tests can avoid depending on them, the better.

PR #6200, opened 2016-12-09
Fix tutorials failures.

gunan:
ptb_word_lm should work, but want to see how cifar10_train will look.

### DIFF
--- a/tensorflow/models/image/cifar10/cifar10_input.py
+++ b/tensorflow/models/image/cifar10/cifar10_input.py
@@ -183,6 +183,10 @@ def distorted_inputs(data_dir, batch_size):
   # Subtract off the mean and divide by the variance of the pixels.
   float_image = tf.image.per_image_standardization(distorted_image)
 
+  # Set the shapes of tensors.
+  float_image.set_shape([height, width, 3])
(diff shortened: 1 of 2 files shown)


caisq:
ptb_word_lm is fixed (great!) 

But looks like cifar10_train is still failing, maybe read_input.label needs to call set_shape() as well?

caisq (approved):

caisq:
Thanks a lot for the fix, @gunan !

jperczel:
Hi, with the new version of read.py I get the following error message:
TypeError: strided_slice() takes at least 4 arguments (3 given)

gunan:
@physicsai are you using compatible versions of this file and the pip package?

If you have tensorflow 0.12.0rc0 installed, you need to use the version of the file in r0.12 branch of this repo. If you have built tensorflow from source, you can use this file.

Also, please reach out to us by either filing an issue, or through stackoverflow. Pull requests, after merging, cannot really address any questions anymore.

physicsai:
@gunan  It was indeed an issue with tensorflow verions. It runs all fine when I clone the r0.11 files. Thanks for your help!

Issue #6146, opened 2016-12-07
Fast Layer Normalization GPU kernel


MycChiu:
Hi, I wrote a [custom CUDA op for layer normalization](https://github.com/MycChiu/fast-LayerNorm-TF), which is about 5-10x faster than the current `tf.contrib.layers.layer_norm`.

After I posted the link on reddit, some people suggested that I should try to merge this kernel into the trunk, but I have never done this before, so I need some guidance on how this custom kernel could be integrated into TensorFlow's current code.

With my current understanding on how TensorFlow's code is organized, here are some of the problems I think should be solved before the custom op can be integrated into the trunk:
1. It doesn't have CPU kernels.
2. Since the kernel uses shuffle instructions, it only supports cards newer than Kepler.
3. Current implementation has a restriction on the size of last dimension (cannot be larger than 5102)

Without solving these problems, the fastest way to merge this op into the trunk is probably by putting it under the `tf.contrib` folder as it is right now. I will fill out the CLA and make a pull request if you guys think this custom kernel is ready to be merged into the `tf.contrib` folder.

michaelisard:
@martinwicke can you comment?

martinwicke:
In principle, there's no problem in registering a kernel only for GPU. Kepler (CC 3.0, I think) is the minimum for TF anyway, so no problem there either. We may even be able to dispatch to the kernel from contrib.layers.layer_norm, but that may be a bit of a hack. More probably, it'll be a separate op that you can call instead of the layer.

I think you should try merging this into contrib, please send a PR!

martinwicke:
I changed the tags, please reference this issue from your PR.

MycChiu:
Thank you @martinwicke! As requested, #6205 .

PR #6205, opened 2016-12-09
Adding fast layer normalization GPU kernels and layer functions to tf.contrib.layers

MycChiu:
from issue #6146 
Instead of putting the code into a new directory under `contrib`, I took some time and merge them into `contrib/layers`, because I thought it fits quite nicely there. 

Due to my rudimentary knowledge of Bazel's build rules, I couldn't figure out a way to put the kernel codes in the `kernels` folder, since there's a  `BUILD` file in the folder already, it outputs ` crosses boundary of subpackage` error during `./configure` if I reference kernel codes from `contrib/layers/BUILD`. As the result, I put them into `ln_kernels` folder as a temporary hacky measure.


ilblackdragon (changes requested):
Can you please run http://clang.llvm.org/docs/ClangFormat.html or similar to reformat according to Google C++ style guide (https://google.github.io/styleguide/cppguide.html).

Also I have a bit high level question - can the 4 different C++ Ops you defined be one Op that possible arguments are nullptr? If not - at least than have one main function that does all the argument checks and then passes into launcher. Because currently it's a lot of duplicating code.

tensorflow/contrib/layers/ln_kernels/layer_norm_fused_grad_op.cc
```diff
@@ -0,0 +1,356 @@
+/* Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
MycChiu: Yes, I tried to put them in layers/kernels at first, but bazel would raise `crosses boundary of subpackages` error during build. After some googling, I found out that it's caused by the existing `BUILD` file in layers/kernels, but I couldn't figure out a way to write the build rules in this case, so I put the kernel codes in a separate folder as a temporary measure. It would be great if you could give me some guidance on how to write the build rules to resolve this.

tensorflow/contrib/layers/ln_kernels/layer_norm_fused_op.cc
```diff
@@ -0,0 +1,344 @@
...
+
+
+  TF_DISALLOW_COPY_AND_ASSIGN(LayerNormFusedOp);
+};
+
+
+
+
+
+REGISTER_KERNEL_BUILDER(
```
MycChiu: Yes! I removed them so I could compile the kernel without using bazel and forgot to add them back before I send the pull request.

MycChiu:
@ilblackdragon Thank you for the reviews, I am quite new to C++, and I wasn't even aware that a useful tool like clang-format exists, sorry for the potential headaches I may have caused. 
As for the code duplication problem, I have some rough ideas on how to improve it, but it's quite late over here, so I will work on this tomorrow.

MycChiu:
I have run clang-format on the c++ codes using `-style=Google`, renamed a few variables for clarity, and fixed the issues pointed out by @ilblackdragon that I didn't have questions about.

I also tried to merge the 3 separate ops into one, but to do that, I would have to pass `None` into the ops from python (for the cases where gamma and beta are not needed). I searched through the core kernels but couldn't find examples doing this, I have thought of two ways of handling this:
1. pass `False` when gamma or beta is `None` and add boolean as supported dtype in `REGISTER_OP`. Then I could just handle different cases in c++ codes.
2. Eliminate `LayerNormCustom` and `LayerNormBiasAddCustom`, then just pass in dummy constant tensors into the op like I am currently doing when gamma is not `None` but beta is.

option 1 is a bit hacky and will probably not reduce too much code, since most code duplication happens in `layer_norm_fused_op_gpu.cu.cc` and I am not sure how to merge the three kernels into one without performance hit.
option 2 will reduce the code by 3-fold, but will increase the run-time during backward pass by about 1.5-2x  when beta and gamma are not needed because of the unnecessary atomic adds. However, this performance hit will be a lot less significant in the context of entire model, since it is still a lot faster than most time-consuming ops like convolution and matmul. 

Both options have some drawbacks, and I would like to know if you guys have any better suggestions to tackle this problem.

ilblackdragon:
@MycChiu Sorry for late response. I want to include people with better cuda understanding.

@zheng-xq Can you please take a look at Cuda part and if you have any suggestions on simplifying TF kernels.

drpngx:
@MycChiu also, could you merge & resolve conflicts? Thanks.

MycChiu:
No problem. Done.

tensorflow/contrib/layers/ln_kernels/layer_norm_fused_grad_op.cc
drpngx: Could you be more specific about the BUILD file problem?

MycChiu: Certainly, if I want to put the kernel codes in `/layers/kernel`, I would have to write build rules in the `/layers/kernel/BUILD` and then link to it with deps in the `tf_custom_op_library` build rules defined in `/layers/BUILD` ([here](https://github.com/tensorflow/tensorflow/pull/6205/files#diff-853b616fe717bd914bdae461d1ea26bfR77)). The problem I ran into is that I don't know how to write the `cc_library` build rules in `/layers/kernel/BUILD` that use gpu_srcs and couldn't find examples in the source code.

drpngx: I see, I'll find an example tomorrow. Basically, there are kernels and ops in  `tensorflow/core/kernels`, you can see how they are built.

drpngx:
Sorry for the delay. So, for the kernels, you just define a `tf_kernel_library`:
```
tf_kernel_library(
    name = "depthwise_conv_op",
    prefix = "depthwise_conv_op",
    deps = [
        ":conv_ops",
        ":ops_util",
        "//third_party/tensorflow/core:core_cpu",
        "//third_party/tensorflow/core:framework",
        "//third_party/tensorflow/core:lib",
        "//third_party/tensorflow/core:nn_ops_op_lib",
    ],
)
```
Then, for the ops, you just need to define a `cc_library`. Remember to set the `copts`:
```
cc_library(
    name = "word2vec_ops",
    copts = tf_copts(),
    srcs = ["ops/word2vec_ops.cc"],
    linkstatic = 1,
    visibility = ["//third_party/tensorflow:internal"],
    deps = ["//third_party/tensorflow/core:framework"],
    alwayslink = 1,
)
```

MycChiu:
@drpngx Thanks for the directions, but by looking at the example, I am still not quite sure where to declare the dependencies for the GPU codes (`layer_norm_fused_op_gpu.cu.cc`).

drpngx:
Oh, the `tf_custom_op_library` accepts a `deps` section like everything else, see for instance `contrib/rnn/BUILD`.

MycChiu:
OK, this most recent commit is as far as I can go, with this, if I run
`bazel build -c opt --config=cuda //tensorflow/contrib/layers:layer_norm_fused_op_kernel`
I would get this error:
```
ERROR: /home/dizizmaname/tensorflow/tensorflow/contrib/layers/kernels/BUILD:36:1: C++ compilation of rule '//tensorflow/contrib/layers/kernels:layer_norm_fused_kernel' failed: crosstool_wrapper_driver_is_not_gcc failed: error executing command external/local_config_cuda/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc -U_FORTIFY_SOURCE '-D_FORTIFY_SOURCE=1' -fstack-protector -fPIE -Wall -Wunused-but-set-parameter ... (remaining 61 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
In file included from ./third_party/eigen3/unsupported/Eigen/CXX11/Tensor:4:0,
                 from tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc:23:
external/eigen_archive/unsupported/Eigen/CXX11/Tensor:84:26: fatal error: cuda_runtime.h: No such file or directory
 #include <cuda_runtime.h>
                          ^
compilation terminated.
Target //tensorflow/contrib/layers:layer_norm_fused_op_kernel failed to build
```
I think I was not clear enough in my last comment; what I meant was I didn't know where to put `layer_norm_fused_op_gpu.cu.cc`, if I put it in cc_library's srcs,like what I am doing in this commit, it would give the above error. If I put it in tf_custom_op_library like this:
```
tf_custom_op_library(
    name = "python/ops/_layer_norm_fused_op.so",
    srcs = [
        "ops/layer_norm_fused_op.cc",
        ],
    gpu_srcs = [
        "kernels/layer_norm_fused_op_gpu.cu.cc",
       ],
    deps = [
        "//tensorflow/contrib/layers/kernels:layer_norm_fused_kernel",
    ]
)
```
it would give me the following error:
```
ERROR: /home/dizizmaname/tensorflow/tensorflow/contrib/layers/BUILD:77:1: Label '//tensorflow/contrib/layers:kernels/layer_norm_fused_op_gpu.cu.cc' crosses boundary of subpackage 'tensorflow/contrib/layers/kernels' (perhaps you meant to put the colon here: '//tensorflow/contrib/layers/kernels:layer_norm_fused_op_gpu.cu.cc'?).
```

drpngx:
Ok, the error that you're getting is because of visibility. You have to to contrib/layers/kernels:layer_norm_fused_kernel rule, and set the visibility to your rule, something like 

```
visibility= ["//tensorflow/contrib/layers/python/ops/..."],
```

MycChiu:
I made the visibility public like this in `/layers/kernels/BUILD`:
```
cc_library(
    name = "layer_norm_fused_kernel",
    copts = tf_copts(),
    srcs = [
        "layer_norm_fused_op.cc",
        "layer_norm_fused_grad_op.cc",
        "layer_norm_fused_op_gpu.cu.cc",
        "layer_norm_fused_op.h",
    ],
    deps = [
        "//tensorflow/core:framework_headers_lib",
        "//third_party/eigen3",
        "@protobuf//:protobuf",
    ],
    alwayslink = 1,
)
```
 but bazel still throws the same error
```
ERROR: /home/***/tensorflow/tensorflow/contrib/layers/kernels/BUILD:36:1: C++ compilation of rule '//tensorflow/contrib/layers/kernels:layer_norm_fused_kernel' failed: crosstool_wrapper_driver_is_not_gcc failed: error executing command external/local_config_cuda/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc -U_FORTIFY_SOURCE '-D_FORTIFY_SOURCE=1' -fstack-protector -fPIE -Wall -Wunused-but-set-parameter ... (remaining 61 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
In file included from ./third_party/eigen3/unsupported/Eigen/CXX11/Tensor:4:0,
                 from tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc:23:
external/eigen_archive/unsupported/Eigen/CXX11/Tensor:84:26: fatal error: cuda_runtime.h: No such file or directory
 #include <cuda_runtime.h>
                          ^
compilation terminated.
Target //tensorflow/contrib/layers:layer_norm_fused_op_kernel failed to build
```

drpngx:
The GPU sources need to be compiled in a special way. You probably want to use `tf_custom_op_library`  with `gpu_srcs` to set up the include path right.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+inline int get_block_size(const int slice_size, int& block_size, int& mult) {
+  const int _warp_size = 32;
+  int _block_size = _warp_size;
+  int _mult = slice_size / _block_size;
+  mult = _mult * _block_size >= slice_size ? _mult : _mult + 1;
+  while (mult > 5) {
+    _block_size += _warp_size;
+    _mult = slice_size / _block_size;
+    mult = _mult * _block_size >= slice_size ? _mult : _mult + 1;
+  }
```
zheng-xq: Isn't this loop O(slice_size)? For larger depth, I feel there is a cleaner way. Also better comments documenting what we are trying to do is helpful.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+  block_size = _block_size;
+  return 0;
+}
+
+template <typename T>
+__global__ void fillZeros(T* __restrict__ output, const int n_inputs) {
```
zheng-xq: Is this function used? If only in a comment, better to remove them both.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+  for (int thread_id = threadIdx.x + blockDim.x * blockIdx.x;
+       thread_id < n_inputs; thread_id += blockDim.x * gridDim.x)
```
zheng-xq: Put the next statement within an enclosing block.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
zheng-xq: Is this function used? If only in a comment, better to remove them both.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+  val1 += __shfl_xor(val1, 4);
+  val2 += __shfl_xor(val2, 4);
+  val1 += __shfl_xor(val1, 2);
+  val2 += __shfl_xor(val2, 2);
+  val1 += __shfl_xor(val1, 1);
+  val2 += __shfl_xor(val2, 1);
+}
+
+template <typename T>
+__device__ __inline__ T get_value(const T* index, const int bound_check,
```
zheng-xq: It is a bit misleading to call this "index". It is a pointer. So "pointer" or "address" are better

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+    _beta[m] = get_value<T>(beta + thread_slice_id + m * blockDim.x,
+                            thread_slice_id + m * blockDim.x, in_depth);
+  }
+
+  for (int bId = blockIdx.x; bId < num_blocks; bId += gridDim.x) {
+    sum = static_cast<T>(0.0f);
+    sqSum = static_cast<T>(0.0f);
+
+    if (thread_slice_id == 0) {
+      mean_cache[0] = static_cast<T>(0.0f);
```
zheng-xq: Since this is not really an array, I feel it is more clear to say "*mean_cache", instead of "mean_cache[0]".

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+      std_cache[0] = static_cast<T>(0.0f);
+    }
+    __syncthreads();
+
+    UNROLL for (int m = 0; m < mult; m++) {
+      thread_id[m] = bId * in_depth + m * blockDim.x + thread_slice_id;
+    }
+
+    UNROLL for (int m = 0; m < mult; m++) {
+      // inp[m] = __ldg(input+thread_id[m]);
```
zheng-xq: Remove the comments for the code review.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+      inp[m] = get_value<T>(input + thread_id[m],
+                            thread_slice_id + m * blockDim.x, in_depth);
+      // if(blockIdx.x==0&&threadIdx.x==0)
+      // printf("m:%d,bId:%d,tId:%d,inp:%f\n",m,bId,thread_id[m],inp[m]);
+    }
+
+    UNROLL for (int m = 0; m < mult; m++) { sum += inp[m] * i_n; }
```
zheng-xq: The multiplication of "i_n" can be done outside the loop.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+    // if(threadIdx.x==0&&blockIdx.x==0)printf("mu:%f\n", mu);
+    UNROLL for (int m = 0; m < mult; m++) {
+      if (thread_slice_id + m * blockDim.x < in_depth)
+        sqSum += (inp[m] - mu) * (inp[m] - mu);
+    }
+    for (int mask = warpSize / 2; mask > 0; mask /= 2) {
+      sqSum += __shfl_xor(sqSum, mask);
+    }
+    if (thread_warp_id == 0) {
+      atomicAdd(&std_cache[0], sqSum);
```
zheng-xq: This adds a lot of pressure to the only element with atomic add. The typical practice is to write to different elements, and have the first warp do the final accumulation.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+}
+
+namespace tensorflow {
+
+namespace {
+
+typedef Eigen::GpuDevice GPUDevice;
+
+template <typename T, int mult>
+__global__ void LayerNormGPUKernel(const LayerNormFusedArgs args,
```
zheng-xq: We also need a python or C++ benchmarks so the efficiency of this kernel for typical sizes.

zheng-xq (commented):
Initial comments

tensorflow/contrib/layers/python/ops/layer_norm_fused_op.py
```diff
@@ -0,0 +1,90 @@
...
+def _LayerNormFusedCustomGrad(op, grad):
+    in_back, gamma_back, beta_back = _layer_norm_fused_op.layer_norm_fused_backprop_custom(
+        op.inputs[0], grad, op.inputs[1],
+        op.get_attr("epsilon"))
+    return [in_back, gamma_back, beta_back]
+
+
+def layer_norm_fused_op(input_tensor, gamma=None, beta=None,
+                        epsilon=1e-12, name=None):
+    """Fast and efficient layer normalization along the last dimension
```
MycChiu: Yes, I also wondered whether I should add data_format as param as well, but then I realized NCHW/NHWC only applies to 4D tensors, and this op actually works on tensor rank >=2, and normalizes the last dimension regardless of tensor rank, so I thought data_format is less applicable here.

tensorflow/contrib/layers/kernels/layer_norm_fused_op.cc
```diff
@@ -0,0 +1,317 @@
...
+            << "depth:" << depth << ", "
+            << "n_slices:" << n_slices;
+
+    LayerNormFusedArgs args;
+    args.depth = depth;
+    args.n_slices = n_slices;
+    args.n_inputs = n_slices * depth;
+    args.epsilon = epsilon_;
+
+    if (depth <= WARP_SIZE) {
```
MycChiu: Sorry, I am not sure what Cuda translation unit means. Do you mean I should move this code to `layer_norm_fused_op_gpu.cu.cc`?

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
MycChiu: Yes... I stuck on this part for a long time, but eventually compromised on this solution, because most of my use cases had slice_size <= 1500, so the loop will not go beyond 10 steps. Also, I thought the  extra time needed for this loop would be many times shorter than the kernel itself, so I was a bit reluctant on optimizing this part.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
```diff
@@ -0,0 +1,1455 @@
...
+
+template <typename T>
+__global__ void fillZeros(T* __restrict__ output, const int n_inputs) {
+  for (int thread_id = threadIdx.x + blockDim.x * blockIdx.x;
+       thread_id < n_inputs; thread_id += blockDim.x * gridDim.x)
+    output[thread_id] = static_cast<T>(0.0f);
+}
+
+template <typename T>
+__device__ __inline__ void warpSum(T& val1, T& val2) {
```
MycChiu: In principle, this allows CUDA to issue two instructions at a time without being blocked by the previous `+=` operation. In reality, I am not really sure how much speed gain this would bring us. If you think this is not necessary, I am fine with two loops as well.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
MycChiu: I have provided the benchmark in the [original repo](https://github.com/MycChiu/fast-LayerNorm-TF). Should I move them in here as well?

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
MycChiu: I did this so CUDA can use fused multiply add (FMA) instructoins instead of adding and then multiply. It seems like this is the recommended approach from [CUDA's documentation](http://docs.nvidia.com/cuda/floating-point/#fused-multiply-add-fma), since it reduces rounding steps. Also according to the [programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#maximize-instruction-throughput) add and multiply-add actually take same amount of time.

tensorflow/contrib/layers/kernels/layer_norm_fused_op_gpu.cu.cc
MycChiu: I think I used this approach originally, but it was actually slower (or similar, I am a bit fuzzy on this one since this is a month or two ago).I am not sure why, but my current hypothesis is that accumulating in the end required an additional `__syncthreads()`, and this `atomicAdd` is accessing shared memory with only one thread per warp accessing the location.

MycChiu:
@drpngx I am taking @zheng-xq's advice and move this code out of `tf.contrib.layers` for now, so this is temporarily resolved, thanks for your help!

drpngx:
Please resolve conflicts & update when ready. Thanks!

alonpal:
Any update on this?
Thanks for the effort.

rmlarsen:
@MycChiu what's the status. Are you planning to move this as stated?

MycChiu:
Sorry for the long delay, I have resolved the conflict now.

rmlarsen:

rmlarsen:

vrv:
It looks like there are a lot of failures, most notably the sanity checking failures.

Can you run clang-format --style=google over these files and make sure the relevant tests are passing?  Thanks!

MycChiu:
Oops, sorry for the long delay, I didn't realized my chrome logged me out of github.
It seemed like the sanity check failed because of tensorflow's indentation width is 2 spaces instead of 4.
I have also fixed __ldg's backward compatibility error in Linux GPU test.
All the other checks seem to be aborted by unknown. Except the buildifier error in sanity check, which I am not really sure how to fix it.

vrv:
For BUILD, I would suggest installing https://github.com/bazelbuild/buildifier and running it over the BUILD file to fix any formatting issues

MycChiu:
@vrv Thank you! Apparently it was just some formatting issue, they are now buildified.

drpngx:
Jenkins, test this please.

drpngx:
You need to update your tests. There is no more `mul` or `concat_v2`. See `tools/compatibility/tf_upgrade.py`.

```
ERROR: testCreateConvWithWD (__main__.ConvolutionTest)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/contrib/layers/python/layers/layers_test.py", line 397, in testCreateConvWithWD
    images, 32, [3, 3], weights_regularizer=regularizer)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/contrib/framework/python/ops/arg_scope.py", line 177, in func_with_args
    return func(*args, **current_args)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/contrib/layers/python/layers/layers.py", line 910, in convolution
    outputs = layer.apply(inputs)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/layers/base.py", line 323, in apply
    return self.__call__(inputs, **kwargs)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/layers/base.py", line 289, in __call__
    self.build(input_shapes[0])
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/layers/convolutional.py", line 138, in build
    dtype=self.dtype)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/ops/variable_scope.py", line 1034, in get_variable
    use_resource=use_resource, custom_getter=custom_getter)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/ops/variable_scope.py", line 933, in get_variable
    use_resource=use_resource, custom_getter=custom_getter)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/ops/variable_scope.py", line 349, in get_variable
    validate_shape=validate_shape)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/layers/base.py", line 278, in variable_getter
    variable_getter=functools.partial(getter, **kwargs))
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/python/layers/base.py", line 247, in _add_variable
    regularization = regularizer(variable)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local_linux-opt/bin/tensorflow/contrib/layers/layers_test.runfiles/org_tensorflow/tensorflow/contrib/layers/python/layers/regularizers.py", line 107, in l2
    return standard_ops.mul(my_scale, nn.l2_loss(weights), name=name)
AttributeError: 'module' object has no attribute 'mul'
```

MycChiu:
@drpngx Hmmm...It seems like those are from the `contrib/layers` folder, and git didn't merge the latest changes from master to this pull request after I moved my codes out of `contrib/layers`. I ended up just replace the folder with the one I have from the latest master branch, hopefully this would not make other things break.

drpngx:
Jenkins, test this please.

drpngx:
No luck: `//bazel_pip/tensorflow/contrib/layer_norm:layer_norm_fused_test`.

```
InvalidArgumentError (see above for traceback): No OpKernel was registered to support Op 'LayerNormCustom' with these attrs.  Registered devices: [CPU], Registered kernels:
  <no registered kernels>

	 [[Node: LN/LayerNormCustom = LayerNormCustom[T=DT_FLOAT, epsilon=1e-12, _device="/device:CPU:0"](Const)]]
```

MycChiu:
@drpngx Ah! That's because it doesn't have a CPU implementation, is there a way to by-pass this?

drpngx:
Could we just register the CPU kernel instead? It looks like you're using eigen so it should work both ways?

MycChiu:
@drpngx Unfortunately, the kernel itself is implemented with pure CUDA. Is there no way to only register the kernel for GPU?

drpngx:
How hard would it be to have a CPU implementation? I think it might be possible to skip them under CPU, but it's such an unusual pattern that I'm concerned that it might trip up people.

MycChiu:
I imagine a CPU kernel won't be too hard, but I have almost no knowledge in the eigen library, so it would take a lot of time for me to write a native CPU layer norm kernel in C++. Is it okay to direct the CPU cases to the current layer_norm implementation in `tf.contrib.layers` library in python?

drpngx:
Got it. I guess we can `try: except:` to bypass the error there, or use `tensorflow.python.client.device_lib.list_local_devices()` to check if a GPU is available.

vrv:
I believe there is a 'is_gpu_available' in test_util.py somewhere...  and you can see its uses.

drpngx:
Jenkins, test this please.

MycChiu:
@vrv Thank you, I have added it to all test cases. @drpngx , Sorry, I forgot to add the `cuda_only` arguments in last commit, please run the checks again, thank you.

drpngx:
Jenkins, test this please.

MycChiu:
Hmmm...It seems like Windows Cmake Tests failed with error `ImportError: cannot import name 'layer_norm'`.

drpngx:
Oh did you modify the cmakefile?

MycChiu:
I don't think so...Could it be that the core code hasn't merged with master branch?

MycChiu:
Ah! Do you mean I have to `add_python_module("tensorflow/contrib/layer_norm")` in the [tf_python.cmake](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/tf_python.cmake) file as well? I guess I will do it now.

MycChiu:
Also modify other files imitating what's done for `contrib.rnn`, I hope this will do. *One question out of curiosity though, from `contrib.cmake`'s readme, it seems like they don't support custom_op libraries and most cotrib, so why are we still adding modules from contrib? Just to avoid the `ImportError` I encountered? or prepare for future support?*

drpngx:
There's "it should work" and "we support". Contrib should work, at least build, but we don't support it like we support core, which has a stable API and should work well at all times. CMake itself is not technically officially supported, since it's in contrib, but we make sure it's always working.

MycChiu:
I see. I think I have added all the necessary codes for cmake, can we ask Jenkins to test again now?

drpngx:
Jenkins, test this please.

drpngx:
Woohoo! Passes. According to the thread, @zheng-xq reviewed with "initial comments".

@zheng-xq first pass done, would you mind taking another look?

MycChiu:
YES! Finally! Thank you so much for your patience @drpngx.

drpngx:
Jenkins, test this please.

dandelionmane:
Looks like the test failure was a flake? I'll try retriggering. In the mean time, @zheng-xq PTAL?

dandelionmane:

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

martinwicke:
I re-ran the tests, @zheng-xq left comments, could you address those?

tensorflow/contrib/layers/__init__.py
```diff
@@ -28,6 +28,7 @@
 @@flatten
 @@fully_connected
 @@layer_norm
+@@layer_norm_fused
```
martinwicke: I think it is fine to put into layers. contrib/layers is highly experimental, adding functions there shouldn't be a problem.

zheng-xq:
Since the changes are in the contribs, they look reasonable to me in general. My only remaining concern is lack of benchmarks to support the performance improvement claim. More benchmarks results will make this CL more self sufficient for future improvement. That is a common requirement for core changes. Since this is only in contribs, it is up to you.

martinwicke:
@MycChiu a benchmark would indeed be good. You can look at the benchmarks for the convolution or matmul functions as examples, can you post results?

AndreasMadsen:
I'm very interested in this getting merged. @martinwicke are the benchmarks in https://github.com/MycChiu/fast-LayerNorm-TF not sufficient?

If not, where are the "convolution or matmul" benchmarks you are talking about? I would like to help.

AndreasMadsen:
I'm very interested in this getting merged. @martinwicke are the benchmarks in https://github.com/MycChiu/fast-LayerNorm-TF not sufficient?

If not, where are the "convolution or matmul" benchmarks you are talking about? I would like to help.

drpngx:
Jenkins, test this please.

drpngx:
@MycChiu did you get a chance to add a benchmark?

martinwicke:
Sorry, tensorflow/core/kernels/BUILD is a good entry point for existing benchmarks. It also has instructions on how to run them. You can look for "benchmark" to find benchmark targets and their sources.

@annarev FYI a good use case for benchmarking instructions.

vrv:

I think we can accept without benchmarks for now.

vrv:
@MycChiu seems to have disappeared :(

I'm going to close this to keep our active PRs low (otherwise things slip through the cracks); I'm confident once they are back we'll get this PR in!

manuelsh:
Was this finally implemented in the latest version of Tensorflow (v1.3)? Is it in tf.contrib.layers.layer_norm ?

Thanks

YuhuaBillChen:
Hi, I just wonder how the layer_norm finally goes? As for now (Jan 2018) I think the layer_norm is still using the batch_norm as the backbone. And for my model training of a small mini-batch size, layer_norm is almost 2x slower than batch normalization, any luck that we can make it faster? 
image

Thanks!

AzizCode92:
Is there any update about this issue?

manuelsh:
I believed we are left with slow normalization in TensorFlow, which surprises me.

My experience shared here: https://medium.com/@mansanher/normalization-in-tensorflow-speed-is-an-issue-b8ae14336685

AzizCode92:
Thank you @manuelsh for pointing this issue out. 
I have one question, did you use the `tf.contrib.rnn.LayerNormBasicLSTMCell` with `layer_norm=True` for your experiments? or have you tried your custom layer normalization implementation? if yes, can you share it with us?
I have tried the @MycChiu  pull request and I got fast layer_normalization in term of time but it required more steps to converge than the model without layer_normalization !!!

manuelsh:
Hi Aziz,

You can find my code [here](https://github.com/manuelsh/normalisation-tensorflow/blob/master/mnist_model_tensorflow.ipynb).

vsuthichai:
Was this ever planned on being completed?  It sounded quite promising and seemed like it was almost done.

martinwicke:
If someone wants to take over this PR, that would be great. You are right, it is almost done, it should just have a microbenchmark to make sure it's performing as expected (and we can continuously monitor the performance).

PR #6206, opened 2016-12-09
accelerate crf_log_norm

wanguanglu:
the base `crf_log_norm` function looks very cool, but it runs very slow.
I run the experiment on cpu
```
model name  : Intel(R) Xeon(R) CPU E5-2650 v3 @ 2.30GHz
```

I use the configure
```python
batch_size=128
max_seq_len=200
```
it runs `0.5s` per batch

when I optimize the realization way
it  runs `0.25s` per batch

### DIFF
--- a/tensorflow/contrib/crf/python/ops/crf.py
+++ b/tensorflow/contrib/crf/python/ops/crf.py
@@ -105,23 +105,37 @@ def crf_log_norm(inputs, sequence_lengths, transition_params):
   Returns:
     log_norm: A [batch_size] vector of normalizers for a CRF.
   """
-  # Split up the first and rest of the inputs in preparation for the forward
-  # algorithm.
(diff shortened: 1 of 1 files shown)


kentonl:
This change assumes that the maximum sequence length is known and fixed, which breaks backwards compatibility. It would make sense to fallback to the `dynamic_rnn` implementation in that scenario.

ebrevdo:
Will take a look in early January

drpngx:
Sounds good.

rmlarsen:
@ebrevdo would you have time to review this now?

ebrevdo:
We cannot accept this PR because it assumes a fixed length when calculating
the log_crf.  The current approach uses dynamic_rnn for the reason that the
max_seq_len is probably *not* known at graph build time, and varies from
step to step.

vrv:
Closing due to most recent response suggesting that this change is not correct.  Feel free to comment if you feel this is incorrect, thanks!

PR #6208, opened 2016-12-09
Improve  documentation for tf.split_v

AnishShah:
Hi,

I was reading through [`tf.split_v`](https://www.tensorflow.org/versions/r0.12/api_docs/python/array_ops.html#split_v) documentation. I got confused because `num_split` is not there in the function arguments. To understand better, I read the examples and found that the argument order is not correct. 

So, this pull request currently fixes the example of `tf.split_v`. If you agree that the documentation is not proper, then I can fix that too. Thanks!


ekelsen:
This is fixed internally and will get released in the next cycle, so it probably doesn't make sense to pull, but thanks for catching this.

PR #6212, opened 2016-12-09
Determine output shape for SplitV Op

### LINKED ISSUE #5977: split_v generates tensors of unspecified size for all dimensions, when one dimension is missing at the original one
The new split_v function introduced in 0.12rc0 does not correctly predict the shapes of the resulting tensors when the original tensor has an unknown dimension, resulting on the split dimension being of unknown size. For instance:

```
tensor = tf.placeholder(tf.float32, [None, 12])
parts = tf.split_v(tensor, [6, 4, 2], split_dim=1)
```

This results in three tensors, all with shapes [?, ?], instead of the expected [?, 6], [?, 4] and [?, 2]. While at runtime the resulting tensors are of the right size, this makes it more difficult to debug the code during graph definition.

EDIT: This also creates problems when combined with scan. Because scan expects fn to return tensors of the same shape as the initializers, it forces the use of a reshape just to clear the 'unknown' dimension.

### LINKED ISSUE #6186: tf.split_v results in unknown shape size
### Environment info
Operating System:
Linux Mint 17.3 3.19.0-32-generic
Tensorflow Version 0.12.0-rc0 (https://storage.googleapis.com/tensorflow/linux/gpu/tensorflow_gpu-0.12.0rc0-cp34-cp34m-linux_x86_64.whl)

### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)
`
tf.split_v(tf.placeholder(tf.float32, shape=[10, 32, 32, 3]), [3,7], 0)
`
The result is of shape (?, ?, ?, ?). 

However, when using split for equal length splits it shows the resulting shape:
`
tf.split(0, 2, tf.placeholder(tf.float32, shape=[10, 32, 32, 3]))
`

AnishShah:
Fixes #6186 
Fixes #5977 

I couldn't find any tests for SplitV in `array_ops_test.cc`. Is there any other file for SplitV test?

### DIFF
--- a/tensorflow/core/ops/array_ops.cc
+++ b/tensorflow/core/ops/array_ops.cc
@@ -416,24 +416,38 @@ REGISTER_OP("SplitV")
     .Attr("T: type")
     .Attr("Tlen: {int32, int64} = DT_INT64")
     .SetShapeFn([](InferenceContext* c) {
-      ShapeHandle unused;
+      DimensionHandle split_dimension;
+      TF_RETURN_IF_ERROR(c->MakeDimForScalarInput(2, &split_dimension));
       int32 num_outputs = c->num_outputs();
-      // Return unknown shapes with the same rank as the input
-      // or unknown rank if input's rank isn't known
-      // can't determine exact shapes until runtime because
-      // we don't know where the tensor containing the split sizes
-      // is located
-      int32 rank = c->Rank(c->input(0));
+      // Return unknown shapes if input's rank, split dimension or tensor
+      // containing the split sizes aren't known. If we know the tensor
+      // containing size splits, then determine the output shape.
+      ShapeHandle input = c->input(0);
+      int32 rank = c->Rank(input);
       ShapeHandle output_shape;
+      const Tensor* size_splits = c->input_tensor(1);
       if (rank == InferenceContext::kUnknownRank) {
         output_shape = c->UnknownShape();
+        for (int i = 0; i < num_outputs; ++i) {
+          c->set_output(i, output_shape);
+        }
       } else if (rank == 0) {
         return errors::InvalidArgument("Can't split scalars");
-      } else {
+      } else if (size_splits == nullptr || !c->ValueKnown(split_dimension)) {
         output_shape = c->UnknownShapeOfRank(rank);
-      }
-      for (int i = 0; i < num_outputs; ++i) {
-        c->set_output(i, output_shape);
+        for (int i =0; i < num_outputs; ++i) {
+          c->set_output(i, output_shape);
+        }
+      } else {
+        int64 split_dim = c->Value(split_dimension);
+        TF_RETURN_IF_ERROR(c->WithRankAtLeast(input, split_dim + 1, &input));
+        auto vec = size_splits->vec<int32>();
+        for (int i = 0; i < num_outputs; ++i) {
+          output_shape = c->UnknownShapeOfRank(rank);
+          TF_RETURN_IF_ERROR(
+              c->ReplaceDim(input, split_dim, c->MakeDim(vec(i)), &output_shape));
+          c->set_output(i, output_shape);
+        }
       }
 
       return Status::OK();
--- a/tensorflow/core/ops/array_ops.cc
+++ b/tensorflow/core/ops/array_ops.cc
@@ -419,33 +419,44 @@ REGISTER_OP("SplitV")
       DimensionHandle split_dimension;
       TF_RETURN_IF_ERROR(c->MakeDimForScalarInput(2, &split_dimension));
       int32 num_outputs = c->num_outputs();
-      // Return unknown shapes if input's rank, split dimension or tensor
-      // containing the split sizes aren't known. If we know the tensor
-      // containing size splits, then determine the output shape.
       ShapeHandle input = c->input(0);
       int32 rank = c->Rank(input);
       ShapeHandle output_shape;
       const Tensor* size_splits = c->input_tensor(1);
       if (rank == InferenceContext::kUnknownRank) {
+        // If the rank of input tensor is unknown, then return unkown shapes.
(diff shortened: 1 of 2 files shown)


AnishShah:
Any review for this? I have used three same for loops to set output shape. Can anyone suggest better way to refactor this? Thanks! :)

tensorflow/core/ops/array_ops.cc
```diff
@@ -416,24 +416,38 @@ REGISTER_OP("SplitV")
...
-      for (int i = 0; i < num_outputs; ++i) {
-        c->set_output(i, output_shape);
+        for (int i =0; i < num_outputs; ++i) {
+          c->set_output(i, output_shape);
+        }
+      } else {
+        int64 split_dim = c->Value(split_dimension);
+        TF_RETURN_IF_ERROR(c->WithRankAtLeast(input, split_dim + 1, &input));
+        auto vec = size_splits->vec<int32>();
+        for (int i = 0; i < num_outputs; ++i) {
```
ekelsen: should check that the size of vec is indeed the number of outputs.

tensorflow/core/ops/array_ops.cc
```diff
@@ -416,24 +416,38 @@ REGISTER_OP("SplitV")
...
+        }
       } else if (rank == 0) {
         return errors::InvalidArgument("Can't split scalars");
-      } else {
+      } else if (size_splits == nullptr || !c->ValueKnown(split_dimension)) {
         output_shape = c->UnknownShapeOfRank(rank);
-      }
-      for (int i = 0; i < num_outputs; ++i) {
-        c->set_output(i, output_shape);
+        for (int i =0; i < num_outputs; ++i) {
```
ekelsen: space after =

ekelsen:
added some comments, please also add a test that verifies the output shapes are indeed correct.

martinwicke:
Good question about tests though -- @ekelsen where do splitV tests live?

ekelsen:
The split_v tests live in kernel_tests/split_op_test.py.

tensorflow/core/ops/array_ops.cc
AnishShah: What should I do if both are not equal? `TF_RETURN_IF_ERROR` or I should not determine the shape?

AnishShah: I think, it would be better to add `TF_RETURN_IF_ERROR`.

AnishShah: Sorry, What I meant was should I throw an error if it is not equal?

AnishShah:
@ekelsen Updated the PR.. Please check.

tensorflow/python/kernel_tests/split_op_test.py
```diff
@@ -153,6 +153,14 @@ def testGradientsAll(self):
     self._testGradientsSimple(use_gpu=False)
     self._testGradientsSimple(use_gpu=True)
 
+  def testOutputShape(self):
```
ekelsen: Looks great, thank you, but the indentation is not consistent here with the rest of the file.  Please change to 2 spaces.

ekelsen (changes requested):
Additional error checking looks good.  Just some formatting issues in the new test.

AnishShah:
@ekelsen Done!

ekelsen (approved):
LGTM, thank you for the contribution!

AnishShah:
We can run the tests now?

martinwicke:
Jenkins, test this please!

martinwicke:
split_op_test failed, please check.

AnishShah:
@martinwicke The tests are failing because `size_splits` in [this test](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/split_op_test.py#L69) is a numpy array. If I cast it into a list, then it works properly. I'm not sure any it is failing because of my changes. Do you have any idea?

AnishShah:
@martinwicke The tests are failing because `size_splits` in [these tests](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/split_op_test.py#L69) are numpy array instead of list. If I type cast them into a list, then all the tests are passing. I'm not sure why my changes are causing this error. Do you have any idea about this? Please help. :)

This is the output of running tests.
```
............F tensorflow/core/framework/tensor.cc:487] Check failed: dtype() == expected_dtype (9 vs. 3)
```

tensorflow/core/ops/array_ops.cc
```diff
@@ -416,24 +416,44 @@ REGISTER_OP("SplitV")
...
-      for (int i = 0; i < num_outputs; ++i) {
-        c->set_output(i, output_shape);
+        for (int i = 0; i < num_outputs; ++i) {
+          c->set_output(i, output_shape);
+        }
+      } else {
+        // Determine the output shape if split dimension and split sizes are known
+        int64 split_dim = c->Value(split_dimension);
+        TF_RETURN_IF_ERROR(c->WithRankAtLeast(input, split_dim + 1, &input));
+        auto vec = size_splits->vec<int32>();
```
ekelsen: The problem is here - size_splits can be either of type int32 or int64 (see the signature), but this line assumes int32, which is what is causing the type mismatch.

AnishShah:
I have made the changes and the tests are passing. Can we run the tests again?

AnishShah:
The tests are failing after I rebased the branch. It is failing because of commit - a46b6d211eac423c72d3a57a177daf2f64db8642
Looking into it.

AnishShah:
Done! Please run the tests now.

martinwicke:
Jenkins, test this please!

AnishShah:
All checks have passed :)

ekelsen:
Thanks for the contribution Anish! Should get merged by someone with permission soon.

PR #6218, opened 2016-12-09
Add missing documentation of Nesterov attribute.

inflation:
The doc did not explain the parameter.

### DIFF
--- a/tensorflow/python/training/momentum.py
+++ b/tensorflow/python/training/momentum.py
@@ -40,6 +40,10 @@ class MomentumOptimizer(optimizer.Optimizer):
       use_locking: If `True` use locks for update operations.
       name: Optional name prefix for the operations created when applying
         gradients.  Defaults to "Momentum".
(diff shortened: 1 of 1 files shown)


tensorflow/python/training/momentum.py
```diff
@@ -31,13 +31,16 @@ class MomentumOptimizer(optimizer.Optimizer):
   """
 
   def __init__(self, learning_rate, momentum,
-               use_locking=False, name="Momentum", use_nesterov=False):
+               use_locking=False, use_nesterov=False, name="Momentum"):
```
vrv: Thanks for the doc fix!  Can you restore the ordering of the argument, in case people are not using keyword arguments?

inflation: Ahh, I forgot about that. Thanks!

vrv (approved):

vrv:

PR #6225, opened 2016-12-09
Don't fetch sample Android app deps in configure.

aj-michael:
`bazel fetch` runs the Bazel loading phase on the given targets. At the
moment, the loading phase of an android_binary succeeds even if an
android_sdk_repository is not set up in the WORKSPACE. However, this is
deceptive as the purpose of the loading phase is to ensure that all
files needed for the build are present. Without an android_sdk_repository
set up, this is not the case. In the future, Bazel will likely not allow
the loading phase to succeed for android_binary without an
android_sdk_repository.

Unfortunately, `bazel fetch` does net support the target substitution
syntax (see https://github.com/bazelbuild/bazel/issues/2220) so I've
added a TODO to remove the necessary nested `bazel query` once that
issue is resolved.

### DIFF
--- a/configure
+++ b/configure
@@ -24,7 +24,8 @@ function bazel_clean_and_fetch() {
   if ! is_windows; then
     bazel clean --expunge
   fi
-  bazel fetch //tensorflow/...
+  # TODO(https://github.com/bazelbuild/bazel/issues/2220) Remove the nested `bazel query`.
+  bazel fetch $(bazel query "//tensorflow/... -//tensorflow/examples/android/...")
 }
 
 ## Set up python-related environment settings


andrewharp:

aj-michael:
I guess I should add some motivation for this. I'm planning to make a change in Bazel shortly that will break Tensorflow's `configure` script without this change.

I tried to submit this change a while ago in https://github.com/bazelbuild/bazel/commit/af8ebca2b88a0ee1129b72b4ce825f6489c31e67, which inadvertently broke this `configure` script (see https://github.com/bazelbuild/bazel/issues/1880).

andrewharp (approved):

aj-michael:
Is there anything else I need to do for this PR, or is "andrewharp approved these changes" sufficient to merge it?

aj-michael:
Can someone merge this please?

aj-michael:
Thanks! :)

PR #6231, opened 2016-12-10
Update installation instructions for tensorflow-gpu.

yifeif:


tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -61,24 +61,23 @@ $ sudo easy_install pip
...
-For Windows users, you can also install the GPU version of the binary with:
 ```bash
 $ pip install tensorflow-gpu
 ```
-Unfortunately, this command is not yet available for Linux or Mac GPU binaries
-due to their sizes exceeding the Pypi limit.
 
-If the above commands do not work on your system or you want to install the GPU version
-of the binary on Linux or Mac, you can follow these instructions:
+Note that you will need pip version 8.1 or later for the above commands to work on Linux.
```
gunan: Let's have this at the top, as one of the requirements.

gunan (approved):

tensorflow/g3doc/get_started/os_setup.md
yifeif: Decided to not add it as a requirement as this is only needed for 'pip install tensorflow' to work. If for some reason users cannot upgrade pip, I dont want them to think they cannot install tensorflow. What do you think?

yifeif: what about now?

gunan: Looks good, thanks!

PR #6234, opened 2016-12-10
Moving example models from github.com/tensorflow/tensorflow to github.com/tensorflow/models

nealwu:
See https://github.com/tensorflow/models/pull/731


nealwu:
Looks like we need internal changes to sync before this can go through.

gunan:
Earlier today there was one sync. 
What time do the changes you need from?

gunan:
Also, do we need to clean up any tutorials or documentation?

nealwu:
Deleted and recreated the branch because that was easier than rebasing. It looks like I'll have to make a new pull request though.

For your questions, it looks like the sync that went in should be good. Tutorials and docs have already been updated in CLs that I've submitted. I don't know how long it takes to update to tensorflow.org (which currently still has the old version), but I just checked and the site appears to successfully link to the r0.12 code, so we should be good to go on this.

Issue #6256, opened 2016-12-11
grpc worker->client fetch tops out at 120 MB/s


yaroslavvb:
Fetching data as numpy arrays is slow for grpc sessions. Evaluating 128MB variable repeatedly from `Session("grpc://localhost..")` only gets me 116 MB/s, compared to 5150MB/s for `Session()`

https://gist.github.com/yaroslavvb/59bc8d3d635e1027306486cc418b26aa
```
python client_transfer_benchmark.py --profile
5150.04 MB per second
116.70 MB per second

```
Attached are CPU profiles for worker and for client. Worker uses 2x CPU and both of them are spending majority of the time in in gRPC code, specifically doing `strcmp` (reported as [__nss_hosts_lookup](https://github.com/gperftools/gperftools/issues/1)). 

[profile.client.pdf](https://github.com/tensorflow/tensorflow/files/644905/profile.client.pdf)
[profile.worker.pdf](https://github.com/tensorflow/tensorflow/files/644906/profile.worker.pdf)

This is related but slightly different from https://github.com/tensorflow/tensorflow/issues/6116 -- that issue looks at worker->worker transfer which hits 520 MB/second for the same message size (degrading below 100 MB/second for larger messages)

https://gist.github.com/yaroslavvb/e196107b5e0afc834652bd3153030c42
```
python benchmark_grpc_recv.py --data_mb=128
Local rate:       15214.63 MB/s
Distributed rate: 569.27 MB/s


```

michaelisard:
@mrry do you want to comment on potential fixes in the future, or workarounds for now?

mrry:
We're aware of the inefficiencies in the client-to-and-from-master path when using the distributed runtime, and have plans to cut down on the serialization overhead, especially in the common case where the client and master (and often worker) are colocated in the same process.

@yaroslavvb Can you share more about what your use case is for fetching large values from the distributed runtime? We haven't spent much time optimizing this until now because the common case in distributed execution involves running a training op that returns nothing (or perhaps some small progress indicator values). It would be good to understand your use case so we can prioritize the various steps we can take to improve performance here!

shiyemin:
@mrry Hi, I'm also using distributed training on one server because it's easier to implement and extend. 

When using a large network (e.g. Inception-resnet-2) or batchsize, the traffic between session will be very heavy.

mrry:
[...]

PR #6257, opened 2016-12-11
replace AsProtoField with AsProtoTensorContent for efficiency

yaroslavvb:
This improves throughput of gRPC worker->client fetch about 25% (116 MB/s -> 151 MB/s) as measured by benchmark in https://github.com/tensorflow/tensorflow/issues/6256


mrry:
I'm not sure why this code exists... it seems safe always to do the `Swap()`, but perhaps there are concerns about aliasing....

Anyway, this change looks safe, so

martinwicke:
mrry@ did you want to review further? If not, can you add an approval?

mrry:
@martinwicke Just waiting to see that all the tests pass.

mrry (approved):

PR #6280, opened 2016-12-13
made contrib flatten layer accept tensors with dynamic shapes

ppries:


drpngx:
@ppries sorry we're taking a long time to review. Could you merge and re-upload?

ppries:
@drpngx Done!

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -1201,16 +1202,23 @@ def flatten(inputs,
...
+    batch_dim = array_ops.slice(inputs_shape, [0], [1])
+    spatial_dims = array_ops.slice(inputs_shape, [1], [inputs_rank - 1])
+
+    flat_spatial_dim = math_ops.reduce_prod(spatial_dims)
+    flat_spatial_dim = array_ops.expand_dims(flat_spatial_dim, 0)
+
+    flat_shape = array_ops.concat(0, [batch_dim, flat_spatial_dim])
+
+    outputs = array_ops.reshape(inputs, flat_shape)
+
```
sguada: Can you propagate the shape info to outputs using set_shape() and get_shape()

sguada (commented):
You would need to add tests for static and dynamic shapes.

ppries:
I've added a test for the case of unknown dimensions – the known dimensions case is already handled in testFlattenBatchSize, and it seems silly to add another identical test only for semantics. Let me know if you disagree.

It is currently failing testInvalidRank, which I can fix by checking the length of get_shape and reverting to the previous way of asserting rank > 2 (it is currently done using TF control_flow ops). This then assumes that the rank is static – I don't know enough about TF to know if that is a problem. 

In any case, I am already using some information from get_shape to decide whether I can propagate shape information to outputs or not, so it might not be an issue. Let me know what you think, then I'll make whatever changes are needed to pass all tests.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -1214,11 +1214,17 @@ def flatten(inputs,
...
 
     outputs = array_ops.reshape(inputs, flat_shape)
 
+    # Attempt to propagate shape information, if it is defined.
+    input_shape = inputs.get_shape().as_list()
+    batch_dim, spatial_dims = input_shape[0], input_shape[1:]
+    if all(spatial_dims):
+      outputs.set_shape([batch_dim,
+                        functools.reduce(lambda x, y: x * y, spatial_dims)])
+
```
sguada: else:
  outputs.set_shape([batch_dim, None])

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -1338,13 +1338,6 @@ def testInvalidRank(self):
                                    'must have a least 2 dimensions'):
         _layers.flatten(inputs)
 
-  def testUnknownLastDim(self):
-    with ops.Graph().as_default() as g, self.test_session(g):
-      inputs = array_ops.placeholder(dtype=dtypes.float32)
-      inputs.set_shape(tensor_shape.TensorShape((5, None)))
-      with self.assertRaisesRegexp(ValueError, '2nd dimension must be defined'):
-        _layers.flatten(inputs)
-
```
sguada: Test that the output shape is (5, None)

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -1386,6 +1379,17 @@ def testFlattenBatchSize(self):
...
+    height = width = depth = 3
+    with self.test_session() as sess:
+      images = random_ops.random_uniform(
+          (5, height, width, depth), seed=1, name='images')
+      inputs = array_ops.placeholder(dtypes.int32, (None, None, None, None))
+      output = _layers.flatten(inputs)
+      output = sess.run(output, {inputs: images.eval()})
+      self.assertEqual(output.size, images.get_shape().num_elements())
+      self.assertEqual(output.shape[0], images.get_shape()[0])
+
```
sguada: Add a test where the dimensions where known at built time, but then changed a running time.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -1201,16 +1202,29 @@ def flatten(inputs,
     ValueError: if inputs.dense_shape is wrong.
   """
   with ops.name_scope(scope, 'Flatten', [inputs]) as sc:
-    inputs = ops.convert_to_tensor(inputs)
-    inputs_shape = inputs.get_shape()
-    inputs_rank = inputs_shape.ndims
-    if (inputs_rank is None) or (inputs_rank < 2):
```
sguada: I would probably keep the check that the inputs have at least 2 dimensions, fail sooner.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -1201,16 +1202,29 @@ def flatten(inputs,
...
-    dims = inputs_shape[1:]
-    if not dims.is_fully_defined():
-      raise ValueError('Inputs 2nd dimension must be defined.')
-    k = dims.num_elements()
-    outputs = array_ops.reshape(inputs, [-1, k])
+
+    inputs_shape = array_ops.shape(inputs)
+    inputs_rank = array_ops.rank(inputs)
+
+    assertion = [control_flow_ops.Assert(inputs_rank > 1,
```
sguada: Better to do a check at build time, that each time it runs.

sguada (commented):
Thanks for the test, looks good, a few extra things.

tensorflow/contrib/layers/python/layers/layers_test.py
ppries: As in define a placeholder with a known shape (e.g. 5 x 3 x 3 x 3) and then feed an input with a different shape, which should fail?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -1199,13 +1198,13 @@ def flatten(inputs,
     ValueError: if inputs.dense_shape is wrong.
```
sguada: ValueError: if inputs rank less than 2.

sguada (approved):
Thanks looks good to me.

ppries:
Do I need to do anything?

@sguada

drpngx:
No you should be fine. Jenkins, test this please.

PR #6291, opened 2016-12-13
Fix tutorials test

caisq:
Temporarily disable the testing of models that have recently been moved to a
separate repository, e.g., cifar10_train.

Add two tutorials in tensorflow/examples/tutorials:
word2vec_basic
estimators/abalone

### DIFF
--- a/tensorflow/tools/ci_build/builds/test_tutorials.sh
+++ b/tensorflow/tools/ci_build/builds/test_tutorials.sh
@@ -33,8 +33,7 @@
 #
 
 # List of all tutorial tests to run, separated by spaces
-TUT_TESTS="mnist_softmax mnist_with_summaries cifar10_train "\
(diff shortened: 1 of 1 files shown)


gunan:
I think we should drop the continuous tests for tutorials and models once we guarantee API stability.
I think at best, we will separate all their tests from core tensorflow. Tutorials will also move out of the core repo, and that is in part the reason for this.

So I think we can completely disable running tutorials instead in jenkins.

caisq:
The tutorial tests are already completely disabled in Jenkins right now. But as a test script, this PR makes sure it runs all right. For releases, it might be useful (we don't want to ship tutorial/model code that doesn't work).

gunan:
That's the thing, we will not be shipping them anymore.

gunan (approved):

PR #5944, opened 2016-11-29
Find correct path of libcudnn.dylib on macOS in configure script.

inflation:
The libcudnn file is stored like `libcudnn.5.dylib` on macOS, but the original script looks for `libcudnn5.dylib`, which would produce an invalid path error. And the issue only affects the auto-detecting part.

PR #6295, opened 2016-12-13
Redo: Find correct path of libcudnn.dylib on macOS.

inflation:
I just found that commit a62c532 overwrote my change in #5944.


martinwicke (approved):
lgtm

PR #6296, opened 2016-12-14
Branch 141962539

martinwicke:


caisq:

caisq (approved):

caisq:
//tensorflow/python/kernel_tests:variable_scope_test failure seems to be legit, reproducible failure. This failure is not seen on CL postsubmit tests. So I guess it is caused by a clash with a external PR. This one looks like a candidate: https://github.com/tensorflow/tensorflow/commit/0958bb705aa124368f7b603151a71386146df2f9

martinwicke:
I'm fixing it, slowly. I'll handle this, you should go sleep.

martinwicke:
There was a large change affecting the behavior if zeros_initializer by @itsmeolivia, which the external PR wasn't aware of. Soon...

martinwicke:
@mrry can you help with the cmake failure? I think it's new kernels in contrib, how should these be added to the cmake build?

mrry:
You need to add a line [here](https://github.com/tensorflow/tensorflow/blob/a2dac9bce595e3d85334c582f690583f5d869ae7/tensorflow/contrib/cmake/tf_python.cmake#L515):

```
GENERATE_PYTHON_OP_LIB("contrib_tensor_forest_ops"
  DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/tf_python/tensorflow/contrib/tensor_forest/python/ops/gen_tensor_forest_ops.py)
```

It looks like the kernels are already built (see [here](https://github.com/tensorflow/tensorflow/blob/a2dac9bce595e3d85334c582f690583f5d869ae7/tensorflow/contrib/cmake/tf_core_kernels.cmake#L45)), but the op registrations are missing, so we'll need another line [here](https://github.com/tensorflow/tensorflow/blob/a2dac9bce595e3d85334c582f690583f5d869ae7/tensorflow/contrib/cmake/tf_core_ops.cmake#L51) like:

```
GENERATE_CONTRIB_OP_LIBRARY(tensor_forest "${tensorflow_source_dir}/tensorflow/contrib/tensor_forest/ops/tensor_forest_ops.cc")
```

martinwicke:
Thank you, that worked.

PR #6300, opened 2016-12-14
Add orthogonal initialization

carlthome:
[Lasagne has orthogonal initialization](https://github.com/Lasagne/Lasagne/blob/master/lasagne/init.py#L327-L367) like in [this reference](https://arxiv.org/abs/1312.6120) and it's pretty useful (and quite intuitive: make sure random weights are orthogonal so they should tend to learn different modes in the input and not converge to being redundant).

# TODO
 - [ ] Add unit test
 - [ ] Verify code style
 - [ ] Verify docstring style
 - [ ] Don't import NumPy if possible (is there a SVD in TensorFlow somewhere?)

### DIFF
--- a/tensorflow/contrib/layers/python/layers/initializers.py
+++ b/tensorflow/contrib/layers/python/layers/initializers.py
@@ -20,12 +20,14 @@ from __future__ import print_function
 
 import math
 
+import numpy as np
+
 from tensorflow.python.framework import dtypes
 from tensorflow.python.ops import random_ops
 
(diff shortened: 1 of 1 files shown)


carlthome:
I signed it!

tillahoffmann:
I think this is a duplicate of #5164.

carlthome:
Very nice! Moving over to https://github.com/tensorflow/tensorflow/pull/5164.

PR #6309, opened 2016-12-14
Expose DataTypeSize as TF_DataTypeSize in C API

tsee:
DataTypeSize is a greatly useful utility function for allocating the
right amount of memory when using the C API for (dynamic) language
wrappers using an FFI.

### DIFF
--- a/tensorflow/c/c_api.cc
+++ b/tensorflow/c/c_api.cc
@@ -81,6 +81,12 @@ extern "C" {
 const char* TF_Version() { return TF_VERSION_STRING; }
 
 // --------------------------------------------------------------------------
+size_t TF_DataTypeSize(TF_DataType dt) {
+  return static_cast<size_t>(
+      tensorflow::DataTypeSize(static_cast<DataType>(dt)));
+}
+
+// --------------------------------------------------------------------------
 struct TF_Status {
   Status status;
 };
--- a/tensorflow/c/c_api.h
+++ b/tensorflow/c/c_api.h
@@ -100,6 +100,11 @@ typedef enum {
   TF_RESOURCE = 20,
 } TF_DataType;
 
+// TF_DataTypeSize returns the sizeof() for the underlying type corresponding
+// to the given TF_DataType enum value. Returns 0 for variable length types
+// (eg. TF_STRING) or on failure.
+extern size_t TF_DataTypeSize(TF_DataType dt);
+
 // --------------------------------------------------------------------------
 // TF_Code holds an error code.  The enum values here are identical to
 // corresponding values in error_codes.proto.
--- a/tensorflow/c/c_api_test.cc
+++ b/tensorflow/c/c_api_test.cc
@@ -203,6 +203,12 @@ TEST(CAPI, DataTypeEnum) {
   EXPECT_EQ(TF_UINT16, static_cast<TF_DataType>(tensorflow::DT_UINT16));
   EXPECT_EQ(TF_COMPLEX128, static_cast<TF_DataType>(tensorflow::DT_COMPLEX128));
   EXPECT_EQ(TF_HALF, static_cast<TF_DataType>(tensorflow::DT_HALF));
+  EXPECT_EQ(TF_DataTypeSize(TF_DOUBLE),
+            tensorflow::DataTypeSize(tensorflow::DT_DOUBLE));
+  EXPECT_EQ(TF_DataTypeSize(TF_STRING),
+            tensorflow::DataTypeSize(tensorflow::DT_STRING));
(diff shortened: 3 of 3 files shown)


tsee:
NB: This is my first pull request to tensorflow. If I missed something in some style guide or otherwise, please let me know. If there's something I can do to make it easier to merge (What's with the various checks? Are this automated?), same thing, feedback welcome and I'm happy to try to do the leg work.

martinwicke:
@josh11b can you take a look or designate someone who can reason sensibly about the C API?

tensorflow/c/c_api.cc
```diff
@@ -81,6 +81,9 @@ extern "C" {
 const char* TF_Version() { return TF_VERSION_STRING; }
 
 // --------------------------------------------------------------------------
+int TF_DataTypeSize(TF_DataType dt) { return tensorflow::DataTypeSize(static_cast<DataType>(dt)); }
```
josh11b: Please format this with the return statement on its own line.  We keep our C++ files to < 80 chars per line.

tensorflow/c/c_api.h
```diff
@@ -101,6 +101,11 @@ typedef enum {
 } TF_DataType;
 
 // --------------------------------------------------------------------------
+// TF_DataTypeSize returns the sizeof() for the underlying type corresponding
+// to the given TF_DataType enum value. Returns 0 on failure.
```
josh11b: Document what it does for variable-length types (currently only string).

tensorflow/c/c_api_test.cc
```diff
@@ -203,6 +203,9 @@ TEST(CAPI, DataTypeEnum) {
   EXPECT_EQ(TF_UINT16, static_cast<TF_DataType>(tensorflow::DT_UINT16));
   EXPECT_EQ(TF_COMPLEX128, static_cast<TF_DataType>(tensorflow::DT_COMPLEX128));
   EXPECT_EQ(TF_HALF, static_cast<TF_DataType>(tensorflow::DT_HALF));
+
+  EXPECT_EQ(TF_DataTypeSize(TF_DOUBLE),
+            tensorflow::DataTypeSize(tensorflow::DT_DOUBLE));
```
josh11b: Could include a few more tests here?  At least TF_STRING and something invalid.

tensorflow/c/c_api.h
```diff
@@ -101,6 +101,11 @@ typedef enum {
 } TF_DataType;
 
 // --------------------------------------------------------------------------
```
josh11b: I wouldn't have a divider here -- this is closely related to the TF_DataType definition.

josh11b (changes requested):
I think this would be a good addition to the C API; just a few things to take care of before we can get this in.

tsee:
I force re-pushed the commit with the edits that @josh11b recommended. Let me know if this is good as is.

For reference, I also have no issues with somebody pushing a modified commit under their name - I'm not in it for the glory and this is as trivial a change as it gets.

tensorflow/c/c_api_test.cc
```diff
@@ -203,6 +203,13 @@ TEST(CAPI, DataTypeEnum) {
   EXPECT_EQ(TF_UINT16, static_cast<TF_DataType>(tensorflow::DT_UINT16));
   EXPECT_EQ(TF_COMPLEX128, static_cast<TF_DataType>(tensorflow::DT_COMPLEX128));
   EXPECT_EQ(TF_HALF, static_cast<TF_DataType>(tensorflow::DT_HALF));
+  ASSERT_EQ(TF_DataTypeSize(TF_STRING), 21312311);
```
josh11b: Why is this an ASSERT instead of an EXPECT, and what is with the strange 21312311 value?

tsee: Oh boy, I'm really sorry. That was a quick debugging statement left in. My son started nagging when I just about finished, so I rushed things more than I should have. 

Re-pushed without the bogus assert. Apologies.

josh11b (approved):

martinwicke:
Jenkins, test this please.

jhseu:
Let's make this a size_t return type? To match the mentioned conventions:
`// * size_t is used to represent byte sizes of objects that are
//   materialized in the address space of the calling process.`

jhseu:
@asimshankar fyi

tsee:
Completely agreed on the size_t change. I'd kept it consistent with the C++ API for an ill-advised sense of consistency and in case I was missing some hidden reason why it would be int. Repushed the branch.

jhseu:
Jenkins, test this please

PR #6323, opened 2016-12-14
Added support for AVX512 to fixed point instructions.

benoitsteiner:


third_party/eigen3/unsupported/Eigen/CXX11/FixedPoint
```diff
@@ -31,8 +31,15 @@
 #include "src/FixedPoint/FixedPointTypes.h"
 
 // Use optimized implementations whenever available
-#ifdef EIGEN_VECTORIZE_AVX2
+#ifdef EIGEN_VECTORIZE_AVX512
+#include "src/Tensor/TensorContractionThreadPool.h"
+#include "src/FixedPoint/PacketMathAVX512.h"
+#include "src/FixedPoint/TypeCastingAVX512.h"
+
+#elif defined EIGEN_VECTORIZE_AVX2
```
rmlarsen: Why elif? It seems so sad to go from AVX2 to AVX-512 only to loose the optimized AVX2 matmul. Can't we fiddle the packet_traits definitions to enable us to use the optimized quantized AVX2 matmul even when AVX512 is enabled for other things like casting and elementwise ops? (Fine to do as a followup I suppose)

third_party/eigen3/unsupported/Eigen/CXX11/src/FixedPoint/PacketMathAVX512.h
```diff
@@ -0,0 +1,545 @@
+#ifndef THIRD_PARTY_EIGEN3_UNSUPPORTED_EIGEN_CXX11_SRC_FIXEDPOINT_PACKETMATHAVX512_H_
```
rmlarsen: It worries me that (AFAICT) there are no direct unit tests in the open source version of TensorFlow for the functions in third_party/eigen3.

benoitsteiner: Unfortunately there is no Jenkins slave that supports AVX512 so we can't test this code automatically yet.

third_party/eigen3/unsupported/Eigen/CXX11/FixedPoint
benoitsteiner: MatMul will come in follow up changes.

rmlarsen (approved):

PR #6328, opened 2016-12-15
Fix: Remove models and core losses op from cmake build.

gunan:
The sources do not exist for these anymore, therefore deleting them.


gunan:
Ah, looking at the wrong thing. This fixes the models issues, but not the losses one.
Closing this and will prepare another PR.

PR #6358, opened 2016-12-16
Fix Java API Typo Errors

seongahjo:
I think it is necessary to use capital letter when it refers to Object just like Tensor.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
@@ -118,14 +118,14 @@ public final class Tensor implements AutoCloseable {
    * Returns the <a href="https://www.tensorflow.org/resources/dims_types.html#shape">shape</a> of
    * the Tensor, i.e., the sizes of each dimension.
    *
-   * @return an array where the i-th element is the size of the i-th dimension of the tensor.
+   * @return an array where the i-th element is the size of the i-th dimension of the Tensor.
    */
   public long[] shape() {
     return shapeCopy;
   }
 
   /**
-   * Returns the value in a scalar {@link DataType#FLOAT} tensor.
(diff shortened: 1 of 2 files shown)


philipperemy:
Lots of typos in Tensorflow but those ones are not typos for me.

seongahjo:
Oh sorry I shoud not use the word 'typo',
Are there any reasons why some of comments are 'Tensor' and others are 'tensor'?

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -170,7 +170,7 @@ public boolean booleanValue() {
   }
 
   /**
-   * Returns the value in a scalar {@link DataType#STRING} tensor.
+   * Returns the value in a scalar {@link DataType#STRING} Tensor.
```
asimshankar: I feel that "tensor" and not "Tensor" is appropriate here and in many of the other places.
The idea being that "Tensor" is used when referring to the Java class or objects that are instances of it. While "tensor" is used for the conceptual N-dimensional array.

One could make the case that the Tensor object itself is not a "scalar" or a "vector", hence "scalar Tensor" isn't meaningful while "scalar tensor" is.

That said, I don't feel too strongly about this and I probably wasn't very consistent in the distinction between 'tensor' and 'Tensor' in comments. 

If you feel particularly strongly, I'd appreciate your thoughts on why.

philipperemy: If it's the Java convention to use capital for classes, then I'm ok with that.

Make sure to be coherent and change it everywhere that is related to Java.

jhseu:
Thanks for the contribution! Closing because the cases are intentional there. If you strongly oppose, please reopen with a comment explaining your reasoning.

PR #6372, opened 2016-12-17
Fix typo on exception message

seongahjo:


caisq:

caisq (approved):

tensorflow/java/src/main/java/org/tensorflow/Session.java
```diff
@@ -225,7 +225,7 @@ public Runner addTarget(String operation) {
       public Reference() {
         synchronized (nativeHandleLock) {
           if (nativeHandle == 0) {
-            throw new IllegalStateException("run() called after the Session was close()d");
+            throw new IllegalStateException("run() called after the Session was closed");
```
asimshankar: Would it be clearer to say:
"run() cannot be called on the Session after close()"
?

seongahjo: Yes, I think it is much clearer than the original message

jhseu:
Jenkins, test this please

asimshankar:
Looks good to me, but I'll be in and out of internet access for the next couple of days, deferring to @jhseu to merge it in.

Issue #6366, opened 2016-12-16
Make VLOG minimum level can be updated via environment


leftnoteasy:
I found inside c++ core implementation there're lots of verbose messages like:
```
    VLOG(1) << "Direct session inter op parallelism threads: " << num_threads;
```
They will be very helpful for developer to understand what happens internally and troubleshoot problems.

However I found in implementation, it doesn't look like works without changing the code. (Credit to @yaroslavvb for the solution, see discussion thread: https://groups.google.com/a/tensorflow.org/forum/#!topic/discuss/D0_3jG5_XRQ)
```
    // TODO(jeff): Define a proper implementation of VLOG_IS_ON
    #define VLOG_IS_ON(lvl) ((lvl) <= 0)

    #define VLOG(lvl)      \
        if (VLOG_IS_ON(lvl)) \
          ::tensorflow::internal::LogMessage(__FILE__, __LINE__, tensorflow::INFO)
```

It will be very helpful if we can update the minimum lvl via passing environment.

leftnoteasy:
And I'm interested to contribute to the feature, if folks think it is a valid feature requirement, I can start working on it.

Thanks!

yaroslavvb:
vlog is kind of useful for debugging out of memory situations, right now the only way to get Tensor allocation/deallocation information is though vlog "__LOG_MEMORY__" messages on stdout

aselle:
The probably reason that VLOG is disabled in code is that it would cause a significant runtime penalty. It might be placed in tight loops where comparing against a variable number would cause performance degradation. As it is now, VLOG blocks will basically be eliminated from compiled code and have no performance penalty. Replacing VLOG_IS_ON(lvl) definition with a call to getenv() would not be ok, because getenv() is relatively expensive. One would probably need to make some static data that is initialized by the environment variable once on translation unit initialization.

Perhaps @vrv or @andydavis1 has more insight?

vrv:
https://github.com/tensorflow/tensorflow/pull/6375 is a reasonable solution that is being worked on: it checks the env var once statically on first use and just checks against the member variable on every call, which shouldn't be that expensive.

leftnoteasy:
Thanks for reviews from @cwhipkey, @vrv, @aselle, very helpful!

For performance concerns from @aselle, I think @vrv's comment should solve the problem:
https://github.com/tensorflow/tensorflow/pull/6375#discussion-diff-93282773R130

I'm going to update the patch according to your feedbacks.

[...]

PR #6375, opened 2016-12-17
Make VLOG minimum level can be updated via environment #6366

leftnoteasy:
Added changes to enable specifying VLOG minimum level. With this patch, user can specify VLOG log level to enable VLOG output, like
```
export TF_CPP_MIN_VLOG_LEVEL=2
```
Tested with fully_connected_feed.py. 

More details see issue #6366 

Thanks,

### DIFF
--- a/tensorflow/core/platform/default/logging.cc
+++ b/tensorflow/core/platform/default/logging.cc
@@ -84,8 +84,8 @@ void LogMessage::GenerateLogMessage() {
 
 namespace {
 
-int64 MinLogLevel() {
-  const char* tf_env_var_val = getenv("TF_CPP_MIN_LOG_LEVEL");
+// Parse log level (int64) from environment variable (char*)
+int64 LogLevelStrToInt(const char* tf_env_var_val) {
   if (tf_env_var_val == nullptr) {
     return 0;
   }
@@ -109,6 +109,16 @@ int64 MinLogLevel() {
   }
 }
 
+int64 MinLogLevel() {
+  const char* tf_env_var_val = getenv("TF_CPP_MIN_LOG_LEVEL");
+  return LogLevelStrToInt(tf_env_var_val);
+}
+
+int64 MinVLogLevel() {
+  const char* tf_env_var_val = getenv("TF_CPP_MIN_VLOG_LEVEL");
+  return LogLevelStrToInt(tf_env_var_val);
+}
+
 }  // namespace
 
 LogMessage::~LogMessage() {
@@ -116,6 +126,8 @@ LogMessage::~LogMessage() {
   static int64 min_log_level = MinLogLevel();
   if (TF_PREDICT_TRUE(severity_ >= min_log_level)) GenerateLogMessage();
 }
+
+int LogMessage::min_lvl_ = MinVLogLevel();
 
 LogMessageFatal::LogMessageFatal(const char* file, int line)
     : LogMessage(file, line, FATAL) {}
--- a/tensorflow/core/platform/default/logging.cc
+++ b/tensorflow/core/platform/default/logging.cc
@@ -92,21 +92,16 @@ int64 LogLevelStrToInt(const char* tf_env_var_val) {
 
   // Ideally we would use env_var / safe_strto64, but it is
   // hard to use here without pulling in a lot of dependencies,
-  // so we do a poor-man's parsing.
+  // so we use std:istringstream instead
   string min_log_level(tf_env_var_val);
-  if (min_log_level == "1") {
-    // Maps to WARNING
-    return 1;
-  } else if (min_log_level == "2") {
-    // Maps to ERROR
-    return 2;
-  } else if (min_log_level == "3") {
-    // Maps to FATAL
-    return 3;
-  } else {
-    // Maps to INFO (the default).
-    return 0;
+  std::istringstream ss(min_log_level);
+  int64 level;
+  if (!(ss >> level)) {
+    // Invalid vlog level setting, set level to default (0)
+    level = 0;
   }
+
+  return level;
 }
 
 int64 MinLogLevelFromEnv() {
--- a/tensorflow/core/platform/default/logging.cc
+++ b/tensorflow/core/platform/default/logging.cc
@@ -109,12 +109,12 @@ int64 LogLevelStrToInt(const char* tf_env_var_val) {
   }
 }
 
-int64 MinLogLevel() {
+int64 MinLogLevelFromEnv() {
   const char* tf_env_var_val = getenv("TF_CPP_MIN_LOG_LEVEL");
   return LogLevelStrToInt(tf_env_var_val);
 }
 
-int64 MinVLogLevel() {
+int64 MinVLogLevelFromEnv() {
   const char* tf_env_var_val = getenv("TF_CPP_MIN_VLOG_LEVEL");
   return LogLevelStrToInt(tf_env_var_val);
 }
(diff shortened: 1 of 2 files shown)


tensorflow/core/platform/default/logging.h
```diff
@@ -71,8 +72,8 @@ class LogMessageFatal : public LogMessage {
 
 #define LOG(severity) _TF_LOG_##severity
 
-// TODO(jeff): Define a proper implementation of VLOG_IS_ON
-#define VLOG_IS_ON(lvl) ((lvl) <= 0)
+// Set TF_CPP_MIN_VLOG_LEVEL environment to update minimum log level of VLOG
```
cwhipkey: for binary size reasons, I think it's preferable on mobile devices to not include vlog support.  So can you include somthing like:

#ifdef IS_MOBILE_PLATFORM
/// old VLOG_IS_ON code
#else
// new VLOG_IS_ON code
#endif

tensorflow/core/platform/default/logging.h
```diff
@@ -40,6 +40,7 @@ class LogMessage : public std::basic_ostringstream<char> {
  public:
   LogMessage(const char* fname, int line, int severity);
   ~LogMessage();
+  static int min_lvl_;
```
cwhipkey: all this min_vlog_lvl_ ?

tensorflow/core/platform/default/logging.h
```diff
@@ -71,8 +72,8 @@ class LogMessageFatal : public LogMessage {
 
 #define LOG(severity) _TF_LOG_##severity
 
-// TODO(jeff): Define a proper implementation of VLOG_IS_ON
-#define VLOG_IS_ON(lvl) ((lvl) <= 0)
+// Set TF_CPP_MIN_VLOG_LEVEL environment to update minimum log level of VLOG
+#define VLOG_IS_ON(lvl) ((lvl) <= ::tensorflow::internal::LogMessage::min_lvl_)
 
 #define VLOG(lvl)      \
   if (VLOG_IS_ON(lvl)) \
```
vrv: similar to the normal log level, can you add

if (TF_PREDICT_FALSE(VLOG_IS_ON(lvl))) ?

(Most of the time, vlogging will be off, so it's nice to provide the compiler hint).

tensorflow/core/platform/default/logging.cc
```diff
@@ -117,6 +127,8 @@ LogMessage::~LogMessage() {
   if (TF_PREDICT_TRUE(severity_ >= min_log_level)) GenerateLogMessage();
 }
 
+int LogMessage::min_lvl_ = MinVLogLevel();
```
vrv: The way this is written, this is called during binary initialization.  It might be nice to have the semantics match the existing normal log behavior, which is that the environment variable is read only the first time you call VLOG.

So what do you think about doing the following: make min_lvl_ a function in LogMessage():

int64 LogMessage::min_vlog_level() {
  static int64 min_vlog_level = MinLogLevel();
  return min_vlog_level;
}

and then call LogMessage::min_vlog_level() in VLOG_IS_ON ?

vrv (changes requested):
Looks generally good, thanks!

tensorflow/core/platform/default/logging.h
leftnoteasy: I would prefer to follow comments from @vrv, adding a method and use a static variable to make logic more consistent to normal LOG behavior.

Thoughts?

tensorflow/core/platform/default/logging.cc
leftnoteasy: Good suggestion, this has better suggestion comparing to static member variable. Will address.

leftnoteasy:
@vrv @cwhipkey, I just updated pull request, mind to take a look again? Thanks!

tensorflow/core/platform/default/logging.h
```diff
@@ -40,6 +40,7 @@ class LogMessage : public std::basic_ostringstream<char> {
  public:
   LogMessage(const char* fname, int line, int severity);
   ~LogMessage();
+  static int64 MinVLogLevel();
```
vrv: Can you add documentation about what this?

// Returns the minimum log level for VLOG statements.
//
// E.g., if MinVLogLevel() is 2, then VLOG(2) statements will produce output, 
// but VLOG(3) will not.  Defaults to 0.

tensorflow/core/platform/default/logging.cc
```diff
@@ -109,14 +109,29 @@ int64 MinLogLevel() {
   }
 }
 
+int64 MinLogLevelFromEnv() {
+  const char* tf_env_var_val = getenv("TF_CPP_MIN_LOG_LEVEL");
+  return LogLevelStrToInt(tf_env_var_val);
+}
+
+int64 MinVLogLevelFromEnv() {
+  const char* tf_env_var_val = getenv("TF_CPP_MIN_VLOG_LEVEL");
```
vrv: Oh, we need to add a TODO here that this currently doesn't work for values not in [1,2,3], so VLOG(4) statements will never print.

Ideally, we'd use safe_strto32 in core/lib/strings/numbers.h, but we don't want to add that dependency into the platform logging library since it would bring in StringPiece and a bunch of other code, so we'd need some simple, sanitizing version of string to integer.  I don't know, maybe std::stoi could be used here safely.

leftnoteasy: @vrv, Yeah agree, I noticed this implementation I thought it was by design.

std::stoi doesn't look safe since it throws exception when input is wrong, see http://en.cppreference.com/w/cpp/string/basic_string/stol.

Can we use atol instead, it returns 0 on error, looks like what we want. 

Thoughts?

leftnoteasy: Use std::istringstream should work, Updated patch to use istringstream and also addressed your above comment: https://github.com/tensorflow/tensorflow/pull/6375#discussion-diff-93358241R43.

Please let me know your thoughts, thanks!

vrv: That looks fine too -- let's run a test to see if it passes on all configurations.

vrv:

tensorflow/core/platform/default/logging.h
```diff
@@ -41,6 +41,11 @@ class LogMessage : public std::basic_ostringstream<char> {
   LogMessage(const char* fname, int line, int severity);
   ~LogMessage();
 
+  // Returns the minimum log level for VLOG statements.
+  // E.g., if MinVLogLevel() is 2, then VLOG(2) statements will produce output,
+  // but VLOG(3) will not. Defaults to 0
```
vrv: add a period after the end of the comment

vrv (approved):

tensorflow/core/platform/default/logging.h
leftnoteasy: Done, thanks!

vrv:

leftnoteasy:
Thanks @jhseu for merge the PR and thanks @vrv / @cwhipkey for reviews/feedbacks!

tensorflow/core/platform/default/logging.h
```diff
@@ -71,11 +76,18 @@ class LogMessageFatal : public LogMessage {
...
+#else
+// Otherwise, Set TF_CPP_MIN_VLOG_LEVEL environment to update minimum log level
+// of VLOG
+#define VLOG_IS_ON(lvl) \
+  ((lvl) <= ::tensorflow::internal::LogMessage::MinVLogLevel())
+#endif
 
 #define VLOG(lvl)      \
-  if (VLOG_IS_ON(lvl)) \
+  if (TF_PREDICT_FALSE(VLOG_IS_ON(lvl))) \
```
ScorpioCPH: @leftnoteasy I am sorry to comment in this `merged` PR, i'm just wondering why predict `FALSE` here?

cwhipkey:
This is to hint that the usual case is when verbose logging is off.  When verbose logging is on, we presumably don't care as much about the performance so don't mind the mis-hint.

PR #6376, opened 2016-12-17
Update configure

bondareko:
b/31910117

correctly finds python3 bin libraries


gunan:
From what I can see, this is already merged into master. am I missing something?

Issue #19, opened 2015-11-09
Swift API


janerivi:
Swift is a very popular and expressive language with a large community of pro developers.

ManuelArno:
Providing APIs for every 'very popular' lang would probably bloat the main project quite quickly.

janerivi:
I understand the sentiment, but Swift, C# and Java are not "every" language, and this is not "every project" either. This is open source and Github so there might be enough hands to man the maintanance of ports to the most popular languages for such an important piece of software. Python is quite popular in academia but a little less popular with the bulk of private sector developers.

jamesliu96:
Community may contribute ports to the project. No need to say here if there is real need of the ports. Actually I think Google may do Go API first.

davidzchen:
Given that TensorFlow is designed to also run on mobile and that Apple is recommending [Swift](https://developer.apple.com/swift/) as a primary language for iOS development, I think having a Swift API for TensorFlow would be a good idea.

I don't think adding support for the most widely-used languages such as Java, C#, and Swift would be a bad idea. The [gRPC](http://www.grpc.io/) project already supports 10 languages, including Java, C#, and JavaScript (via Node.js).

@jamesliu96 Feel free to open an issue for adding a Go API.

jamesliu96:
@davidzchen sure :smiley:

vincentvanhoucke:
A Swift frontend, especially for simply running graphs for inference on mobile would be great. I don't know of current plans to provide one, so feel free to dive in. Since our exact plans on accepting external contributions are still in flux, it would be a good idea to check in with the discuss mailing list with a draft of the code ahead of time to figure out where it should live exactly.

loopbum:
I couldn't agree more with janerivi. Swift has many advanced features that make it one, if not the best choice for full API support for TensorFlow.

This is especially true since Swift is now open source. This language is very pleasant to work with and has all of the great features of top expressive languages like Haskell; Generics, Closures, etc.

rxwei:
I'll take on this.

johndpope:
I have this swift docker file with parsley mcparsefsce / tensorflow + separate swift 3 dev container. - I will plugin swift grpc once google resolve an issue with their just released  grpc library
 
https://github.com/johndpope/DockerParseyMcParsefaceAPI 

I need your help to get it all working. 
Should be able to leverage this environment to get above working.

siilime:
[...]

PR #6423, opened 2016-12-20
Swift 3 wrapper around c interface. 

johndpope:
From comments here @rxwei
https://github.com/tensorflow/tensorflow/issues/19

It's apparent that the c wrapper is limited - and there's no feature parity matching python api. 
In order to progress swift - we need more of the python functionality to be ported to c interface.
Or a path forward using the services layer. 

I wonder if this c porting could be done automagically using some off the shelf library?
http://stackoverflow.com/questions/4650243/convert-python-program-to-c-c-code


johndpope:
I signed it!

johndpope:
@rxwei  - Hi Richard, can you sign the contributor license? 
https://cla.developers.google.com/

rxwei:
Hey @johndpope, my previous commit was highly incomplete and not ready for merge. Would you have time implementing `Session`, `Tensor`, `Graph` to make it at least able to execute a graph on trained params?

jhseu:
Thanks for the contribution! We'd prefer not to commit language bindings to the tensorflow repository that the core TensorFlow team won't yet maintain.

I'm happy to do code review on the Swift bindings and link to them on `tensorflow.org` when they're functional, though.

jhseu:
Also, mind elaborating what you need that's in Python but not available in the C API yet?

The C API is intended to be complete for running inference tasks, such as on mobile phones.

johndpope:
hi @jhseu,

apologies for late reply.
fyi @rxwei  - I have succesfully generated  grpc wrappers for all tensorflow services in swift.
https://github.com/nubbel/swift-tensorflow/
(there's a script to auto build in different target languages too / java /cpp)

@jhseu - When you asked previously about the api and what was missing - the answer would be everything that's in python and not in the c api. There is a growing interest, I imagine one day swift will supercede python for server side development. maybe swift v5.
https://github.com/tensorflow/tensorflow/issues/19

My idea when digging into this was one day to sit down and build graphs in swift instead of using python.  so having a library as deep as the python api is would be the best.
https://github.com/petewarden/tensorflow_makefile/blob/49c08e4d4ff3b6e7d99374dc2fbf8b358150ef9c/tensorflow/g3doc/api_docs/python/index.md

the grpc / from my reading is like how the c++ library currently works
 in that it's mostly for 'control the execution of a graph'
https://github.com/petewarden/tensorflow_makefile/blob/49c08e4d4ff3b6e7d99374dc2fbf8b358150ef9c/tensorflow/g3doc/api_docs/cc/index.md
in that effect / the corresponding swift tensorflow package could do that today (via grpc).

To move forward on original vision - I need to put all the bits of the puzzle together - but alas - it would never completely fulfill my intention of building the graph in swift. From reading into swig / this doesn't have support for swift. https://github.com/swig/swig/issues/981

there's an attempt to auto scaffold swift code from cpp 
https://github.com/sandym/swiftpp
but I'm reluctant to dive down that path as the cpp library I believe has parity with only the grpc services. 

I guess grpc could find a home with swift and end user if it was wired up to behave like sourcekitten / xcode in that as you type in editor it's executing the code and getting responses from the rpc calls.  

I'm enjoying learning about grpc / swift microservices in any case. 

A question to ponder - could you do everything in python today if you were using python via grpc (the old adage eating your own dog food comes to mind - but it describes what I need from google to flesh out my swift project)

I'm off to the pub.

johndpope:
hi @jhseu,

fyi - you may consider linking to this swift port
 https://github.com/PerfectlySoft/Perfect-TensorFlow

tirrorex:
@johndpope one might say that a swift port unusable on iOS is not a swift port.
More like server-side swift tensorflow

johndpope:
@tirrorex  - there's been some developers pushing things forward. 
to do a vanillla swift + tensorflow app - I'm only aware of 
https://github.com/xmartlabs/Bender

there's more updates (swift server side libraries) here - 
https://github.com/tensorflow/tensorflow/issues/19

For my needs / Bender couldn't handle a neural net that was not specific to images. 
https://github.com/xmartlabs/Bender/issues/21

tirrorex:
@johndpope yes i bumped into Bender yesterday after some research, doesn't seem to be anything else unfortunatly.

PR #6425, opened 2016-12-20
Branch 142576906

jhseu:


jhseu:
Jenkins, test this please

jhseu:
Jenkins, test this please

jhseu:
Jenkins, test this please

jhseu:
@mrry fyi, I had to move saved_model deps from `tf_cc_ops.cmake` to `tf_core_cpu.cmake` to fix the windows cmake build. I added saved_model support to the C API in this set of changes, and it's closer to core anyway despite being under `cc/`.

yaroslavvb:
Maybe `tf.norm` should use `tf.norm(m, ord='fro')` instead of `tf.norm(m, order='fro')` for consistency with numpy? @martinwicke

jhseu:
@aselle See the `tf.norm` comment above ^

aselle:
I am ok with order instead of ord, because frankly, numpy's name there is worse. Incompatibility with keyword arguments is less pressing, since those basically will not cause as much confusion as positional argument mismatches (we do match them in some cases, like our change on axis).

jhseu:
@yaroslavvb fyi, it's being renamed to `ord` and should show up during the next sync.

Issue #5972, opened 2016-11-30
optimize variable with dynamic shape not supported


albertz:
TensorFlow 0.11.0.

I want to use a variable where the shape is unknown in advance and it will change from time to time (although ndim is known and fixed).

I declare it like:

```
initializer = tf.random_uniform_initializer()
shape = (s0, s1, s2)  # these are symbolic vars
foo_var = tf.Variable(initializer(shape=shape), name="foo", validate_shape=False)
```
This seems to work when I create the computation graph up to the point where I want to optimize w.r.t. this variable, i.e.:
```
optimizer = tf.train.AdamOptimizer(learning_rate=0.1, epsilon=1e-4)
optim = optimizer.minimize(loss, var_list=[foo_var])
```

That fails in the optimizer in some function `create_zeros_slot` where it seems to depend on the static shape information (it uses `primary.get_shape().as_list()`).

So, using the optimizer works only with variables with static shape? Is that a bug?

My current solution is some hacky monkey patching:
```
def _tf_create_slot_var(primary, val, scope):
  """Helper function for creating a slot variable."""

  from tensorflow.python.ops import variables
  slot = variables.Variable(val, name=scope, trainable=False, validate_shape=primary.get_shape().is_fully_defined())
  # pylint: disable=protected-access
  if isinstance(primary, variables.Variable) and primary._save_slice_info:
    # Primary is a partitioned variable, so we need to also indicate that
    # the slot is a partitioned variable.  Slots have the same partitioning
    # as their primaries.
    real_slot_name = scope[len(primary.op.name + "/"):-1]
    slice_info = primary._save_slice_info
    slot._set_save_slice_info(variables.Variable.SaveSliceInfo(
        slice_info.full_name + "/" + real_slot_name,
        slice_info.full_shape[:],
        slice_info.var_offset[:],
        slice_info.var_shape[:]))
  # pylint: enable=protected-access
  return slot


def _tf_create_zeros_slot(primary, name, dtype=None, colocate_with_primary=True):
  """Create a slot initialized to 0 with same shape as the primary object.

  Args:
    primary: The primary `Variable` or `Tensor`.
    name: Name to use for the slot variable.
    dtype: Type of the slot variable.  Defaults to the type of `primary`.
    colocate_with_primary: Boolean.  If True the slot is located
      on the same device as `primary`.

  Returns:
    A `Variable` object.
  """
  if dtype is None:
    dtype = primary.dtype
[...]

PR #6435, opened 2016-12-21
training.slot_creator: support for non-fully-defined shaped vars

albertz:
This is a possible solution for #5972.

With this patch, the `AdamOptimizer` (and other optimizers) can optimize variables with non-fully-defined shape which would raise an exception without the patch because `primary.get_shape().as_list()` does not work.

### DIFF
--- a/tensorflow/python/training/slot_creator.py
+++ b/tensorflow/python/training/slot_creator.py
@@ -52,7 +52,8 @@ def _create_slot_var(primary, val, scope):
   # scope.
   current_partitioner = variable_scope.get_variable_scope().partitioner
   variable_scope.get_variable_scope().set_partitioner(None)
-  slot = variable_scope.get_variable(scope, initializer=val, trainable=False)
+  slot = variable_scope.get_variable(scope, initializer=val, trainable=False,
+                                     validate_shape=primary.get_shape().is_fully_defined())
   variable_scope.get_variable_scope().set_partitioner(current_partitioner)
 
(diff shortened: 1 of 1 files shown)


drpngx:
@keveman any cycles for this?

keveman (approved):
LGTM.

drpngx:
Jenkins, test this please.

vrv:
Shouldn't we have a test for this?

tensorflow/python/training/slot_creator.py
```diff
@@ -52,7 +52,8 @@ def _create_slot_var(primary, val, scope):
   # scope.
   current_partitioner = variable_scope.get_variable_scope().partitioner
   variable_scope.get_variable_scope().set_partitioner(None)
-  slot = variable_scope.get_variable(scope, initializer=val, trainable=False)
+  slot = variable_scope.get_variable(scope, initializer=val, trainable=False,
```
drpngx: As per @vrv we should add some tests.

drpngx (changes requested):

drpngx:
Good point.

rmlarsen:
@albertz could you add a unit test as requested by the reviewers?

vrv:
Closing due to conflicts + lack of update, to keep our list of open PRs small so they all get attention.  Please feel free to update and we can reopen, or send a brand new PR once you've addressed these comments.

PR #6445, opened 2016-12-21
Fix how error message is generated when running configure

gibiansky:
Very minor fix:

When your TF_CUDA_VERSION does not match the observed CUDA version, an error is raised. Before this patch, the error message generation fails because the `%` operator is applied to only half of the format string. After this patch it correctly creates the error message and displays it as expected.


jhseu (approved):

jhseu:
Jenkins, test this please

PR #6453, opened 2016-12-22
better visualization name for mnist_with_summaries.py

EncodeTS:
**Before the modification**, the name showed in tensorboard:

before

**After the modification**:

after


jhseu:
Thanks for the contribution! As an example, it's probably better to keep the existing name for code understandability, even though it clashes with the scope above it.

PR #6485, opened 2016-12-24
Remove Double from supported types for Conv2D since no double impleme…

yaroslavvb:
…ntation is registered for any device. Came up in http://stackoverflow.com/questions/41310637/no-opkernel-was-registered-to-support-op-conv2d-with-these-attrs


vrv:
The OP spec defines theoretically supported types, not necessarily registered types. 

Perhaps our documentation could avoid specifying types that were not implemented, but that will take some work :(. Sorry about the  confusion this causes.

yaroslavvb:
What is the meaning of theoretically supported? IE, would it make sense to add `int32` to the list of theoretically supported types for `Conv2D` because there may eventually be an integer implementation?

vrv:
If the operation makes sense for a data type, it should be registered in the OP spec. So it would make sense for there to be an int32 kernel. However, for backwards compatibility we can only add data types, not remove them

PR #6494, opened 2016-12-26
Correct the order of arguments of `softmax_loss_function`

boathit:
Correct the order of arguments of `softmax_loss_function` in `seq2seq.sequence_loss_by_example()`,

`softmax_loss_function(target, logit)` => `softmax_loss_function(logit, target)`


boathit:
I signed it.

PR #6522, opened 2016-12-27
Fix build error on Windows.

mrry:
MSVC cannot resolve the type of `nullptr` in `CHECK_EQ(nullptr, ...)`, so we use a static cast to make it explicit.


gunan (approved):

gunan:
Running a build here:
http://ci.tensorflow.org/view/Experimental/job/exp-win-bzl-2/1/

gunan:
OK, windows failure is a test failure I will investigate.
the bazel build works on windows with this change.

PR #6528, opened 2016-12-27
Support reference counting on Java Tensor class

### LINKED ISSUE #6524: Feature request: Support reference counting on Java Tensor class
The `Tensor` class of the new Java library would be more useful and easier to use with built-in reference counting, based on a common interface `RefCounted`.    This would be in lieu of `AutoCloseable`.

I suggest that reference counting follow the Netty approach as described [in the docs](http://netty.io/wiki/reference-counted-objects.html).   A nice aspect is that it simplifies the common case of supplying an input tensor to `Session.Runner::feed()` or to `OperationBuilder::setAttr` (the caller needn't unref the tensor; the callee will unref it).  At the same time, a reference count is of more general utility than `AutoCloseable`.

Please consider applying the idea to `Session` and `Graph` later.

EronWright:
Closes #6524 

- introduces RefCounted and AbstractRefCounted, with tests
- implements reference counting on Tensor (which no longer implements
AutoCloseable)

### DIFF
--- a/tensorflow/java/BUILD
+++ b/tensorflow/java/BUILD
@@ -7,7 +7,7 @@ licenses(["notice"])  # Apache 2.0
 
 java_library(
     name = "tensorflow",
-    srcs = glob(["src/main/java/org/tensorflow/*.java"]),
+    srcs = glob(["src/main/java/org/tensorflow/**/*.java"]),
     data = [":libtensorflow-jni"],
     visibility = ["//visibility:public"],
 )
@@ -17,6 +17,18 @@ java_library(
     testonly = 1,
     srcs = ["src/test/java/org/tensorflow/TestUtil.java"],
     deps = [":tensorflow"],
+)
+
+java_test(
+    name = "AbstractRefCountedTest",
+    size = "small",
+    srcs = ["src/test/java/org/tensorflow/util/AbstractRefCountedTest.java"],
+    test_class = "org.tensorflow.util.AbstractRefCountedTest",
+    deps = [
+        ":tensorflow",
+        ":testutil",
+        "//external:junit",
+    ],
 )
 
 java_test(
--- a/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
@@ -158,8 +158,10 @@ public final class OperationBuilder {
 
   public OperationBuilder setAttr(String name, Tensor value) {
     try (Graph.Reference r = graph.ref()) {
+      // note: setAttrTensor copies the tensor and refs the buffer
       setAttrTensor(unsafeNativeHandle, name, value.getNativeHandle());
     }
+    value.unref();
     return this;
   }
 
@@ -170,7 +172,11 @@ public final class OperationBuilder {
       handles[idx++] = t.getNativeHandle();
     }
     try (Graph.Reference r = graph.ref()) {
+      // note: setAttrTensorList copies the tensor and refs the buffer
       setAttrTensorList(unsafeNativeHandle, name, handles);
+    }
+    for (Tensor t : value) {
+      t.unref();
     }
     return this;
   }
--- a/tensorflow/java/src/main/java/org/tensorflow/Session.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Session.java
@@ -29,18 +29,19 @@ import java.util.List;
  * // for the computation y = 3 * x
  *
  * try (Session s = new Session(graph)) {
- *   try (Tensor x = Tensor.create(2.0f);
- *       Tensor y = s.runner().feed("x", x).fetch("y").run().get(0)) {
- *       System.out.println(y.floatValue());  // Will print 6.0f
- *   }
- *   try (Tensor x = Tensor.create(1.1f);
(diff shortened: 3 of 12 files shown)


EronWright:
@asimshankar please review.

asimshankar:
Thanks for the suggestion and pull request @EronWright .

However, I'm a bit skeptical on this one. I did consider this but felt that the language support for `AutoCloseable` makes it a more appropriate choice at this lower level. Some reservations:

- I fear that it will be much easier for developers to forget to `unref()` than forget to close an `AutoCloseable`, mostly because of language support (and implied un-idiomatic nature of reference counting in Java).
- There is asymmetry between creation and cleanup that is very subtle. For example, if `Session.Runner.run()` isn't called after `Session.Runner.feed()` (perhaps because of some exception thrown in intervening code), then the `Tensor`s are not cleaned up. I see that the PR also uses finalizers, but relying on that is generally unadvisable given that there are no guarantees on garbage collection and that `Tensor` objects can have large memory footprints

A wrapper for reference counting and/or a list of `AutoCloseable`s seems easy enough, but perhaps I'm not fully appreciating the complexity of that. I had thought of something like:

```java
public class AutoCloseableList implements List<AutoCloseable>, AutoCloseable {
   // ...
   public void close() {
      for (AutoCloseable c : list) {
         c.close();
      }
   }
}
```

to make it easier to work with `Session.Runner`, or with reference counting something like:

```java
public class RefCounted<E implements AutoCloseable> {
   public E ref() { ... }
   public boolean unref() {
      if (refcnt-- == 0) {
        object.close();
        return true;
     }
    return false;
  }
}
```

Which I was hoping would be trivial and easy to incorporate adapters into Netty-like settings (and I'd be happy to add those to `org.tensorflow.util`)

Could you perhaps elaborate on why you think such wrappers would be burdensome? Is there some example code you can share that can maybe guide us in finding the right balance? For example, I feel `LabelImage.java` is easier to read without the reference counting. But I am wide open to looking at more use cases.

CC @jhseu

EronWright:
Thanks for the feedback!   I'd still like to convince you that reference counting is superior for this library, especially given its power-over-convenience nature (and hopeful foundation for Scala/Kotlin/Closure libraries, which may not support try-with-resources).   Frankly it is more convenient too.

I've been writing integration code based on this patch, that I'm almost ready to share with folks.   I find it more convenient; input tensors (passed to `feed` or `setAttr`) needn't be dereferenced.  Output tensors must have `unref` or `close` called in any case. 

This style of reference counting (callee must unref) is familiar to Netty developers.  Any scenario involving chaining, asynchrony, or caching will benefit from this; it simply took Netty a few versions to find this solution.

I agree with your concern about the runner/builder leaks.  `finalize` was a catch-all solution but the runner should probably be `AutoCloseable`.

asimshankar:
I still feel that this is making the lower level API more complicated and easier to mess up with. I guess I'm not seeing the convenience benefits that you are, particularly those that cannot be achieved with simple wrapper classes.

I understand we're trying to balance between usability, power and maintainability in any API design. With reference counting here, I feel the latter can suffer quite a bit leading to very subtle bugs if there are `ref()`/`unref()` mismatches and it is easy to end up with them. Without this, the API contract is: If a `Tensor` is materialized, it must be `close()`d, irrespective of how the `Tensor` object is used. With the reference counting, whether a `ref()` or `unref()` is required depends on how the materialized `Tensor` is used later on (e.g. is it passed to `fetch()` or not), which is subtle.

Reference counting is beneficial when there is "shared ownership" of the resource. It isn't clear to me that this should be the default case. I'm thinking of `Tensor` more like a `RandomAccessFile` or `FileChannel` here. I'm not familiar enough with netty, but I'm sure there is a clear pattern to ensure that such objects are created and closed without reference counting support in those classes?

Admittedly, this discussion of convenience and power is in the abstract. If there are real examples of code that we can use to iterate on, that might be more productive. Otherwise, I'm tempted to avoid reference counting here.

jhseu:
I'd favor keeping it as is for the reasons @asimshankar mentioned. I appreciate that the current interface adds difficulty for caches and that it requires a wrapper, but that's true for any `close()`able class (of which many are all applicable to caches).

EronWright:
Thanks for indulging this conversation, these are reasonable objections to be sure.  Go ahead and close the PR if still not convinced after reading the below.

Maybe it is a bit counter-intuitive but reference counting as defined in this PR actually leads to less user code in the common case and fewer leaks.  This is because the callee (`OperationBuilder`, `Session.Runner`) dereferences the object.

Maybe I can strengthen the case by showing how reference counting could simplify the handling of the output tensors.   Imagine that `Session.Runner::run()` returned an object of type `RunResult`.   

```java
class RunResult extends AbstractReferenceCounted {
  private RunMetadata metadata;
  private Tensor[] outputTensors;
  
  public List<Tensor> outputTensors() { ... }

  @Override
  protected void deallocate() {
    for(Tensor t: outputTensors) { t.unref(); }
  }
}
```

Typical user code need only `unref` the result to cleanup everything.  Any tensors that need be retained may simply be `ref`ed.   In the status-quo, the protocol would be more complex IMO.

drpngx:
@asimshankar Can we close this PR?

asimshankar:
@EronWright : Thanks for the comments, but I think we feel that being more idiomatic with language support is more suitable at this "lowest level" of the API. Closing this PR out.

PR #6547, opened 2016-12-29
Branch 143142615

yifeif:


gunan:
Both failures are legitimate issues.
ImportError: cannot import name test_ops, likely a missing all_files rule.

windows_cmake:
ImportError: cannot import name 'nn'  : likely missing include in cmake rules.

yifeif:
Close this and push again once failures are fixed internally.

PR #6557, opened 2016-12-29
Remove manual nvidia-docker instructions.

gunan:


yifeif:

sandersk (approved):
Looks good, thanks.

caisq:
A user added these lines, which seems to indicate that these commands could be useful to some users. If the concern is users asking more questions about vanilla docker not working with GPU, maybe we can just emphasize that this is not supported workflow?

gunan:
Are these instructions not simply manually setting everything nvidia-docker sets?
I tried to write that if nvidia-docker does not work, do this and that, however these kind of things are usually diverting users from simple issues in nvidia-docker to a rather complicated one and returning to us with more questions than reducing questions.

As far as I can tell from the dialogue in the PR, the user was trying to get rid of docker_run_gpu script. Most of the wording here is leftover from when we had that script.
I am still in favor of removing complicated instructions. Nvidia docker is much better than what it was earlier this year, and they are providing very timely support with questions.

caisq (approved):

PR #6562, opened 2016-12-29
More numerically stable gradients for division.

tillahoffmann:
This PR improves the numerical stability of gradients for the division operator. 

Let `f = x / y` such that `df/dy = - x / y ** 2`. But if `y` is of small magnitude, `y ** 2` can underflow, and if `y` is of large magnitude, `y ** 2` can overflow. It is thus better to compute the derivative as `df/dy - (x / y) / y`. The changes are illustrated in the figure below using 32-bit floating point representations.

divgrad

The coverage of the new implementation is strictly greater than the coverage of the old implementation, where coverage is the domain in which the gradient is defined.


tillahoffmann:
Here is a [notebook](https://gist.github.com/tillahoffmann/e2e639335f89f2e5e270bb18edd11ded) to reproduce the figure above.

drpngx (approved):
`-x / (y*y)` as `(-x / y) / y` should be more stable.

drpngx:
Jenkins, test this please.

drpngx:
Ignoring unrelated errors.

PR #6564, opened 2016-12-29
Update tf.nn.conv2d_transpose.md

MadcowD:
This seems wrong.


gunan:
This is an autogenerated file which is generated from this:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_ops.py#L1089

Please edit the inline comments instead of the documentation, it will be picked up in our next website release.

Also, please sign the Google CLA, and use the same email to create the git commits for CLA bot to resolve the CLA.

I am closing this PR. Please address the above issues, then we can create a new PR or reopen this one.

PR #6570, opened 2016-12-30
Update version string to 1.0.0-alpha

gunan:


yifeif (approved):

yifeif:

tensorflow/contrib/cmake/setup.py
```diff
@@ -26,7 +26,7 @@
 from setuptools.command.install import install as InstallCommandBase
 from setuptools.dist import Distribution
 
-_VERSION = '0.12.0-cmake-experimental'
+_VERSION = '1.0.0-alpha-cmake-experimental'
```
caisq: I think it'll give us more room for fixes, if we cal it 1.0.0-alpha1. Does that make sense?

gunan: Discussed offline.
Only one alpha planned, so we decided to keep "alpha"

PR #6577, opened 2016-12-30
[Java] Bulk data transfer for Tensor class

### LINKED ISSUE #6576: [Java] Bulk data transfer for Tensor class
Enhance the `Tensor` class with bulk data-access methods, to support the following scenarios:
- serialization of tensors in managed code (regardless of tensor datatype)
- efficient transfer of large blocks of multidimensional data
- support for direct transfers (i.e. using direct `ByteBuffer`s)
- support for typed `Buffer`s

EronWright:
Closes #6576

Introduces bulk data-access pattern to `Tensor`, with tests.
- `Tensor::create(DataType dt, long[] shape, Buffer data)` method
- `Tensor::getDataByteSize()` method
- `Tensor::readData(Buffer data)` method

Details:
- rename `allocate` to `allocateNDArray` since it is designed for the
NDarray scenario.
- introduce a native `allocate` method that accepts the data size in
bytes, rather than computing the size with shape information (which
isn’t always sufficient, e.g. TF_STRING).
- introduce a private `buffer` accessor to access the TF data as a
direct `ByteBuffer`.
- implement `create` and `readData` methods to read and write data
using a variety of `Buffer` variants.   Supports typed buffers and
direct buffers.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
@@ -16,6 +16,13 @@ limitations under the License.
 package org.tensorflow;
 
 import java.lang.reflect.Array;
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.DoubleBuffer;
+import java.nio.FloatBuffer;
+import java.nio.IntBuffer;
+import java.nio.LongBuffer;
 import java.util.Arrays;
 
 /**
@@ -34,6 +41,115 @@ import java.util.Arrays;
  * }</pre>
  */
 public final class Tensor implements AutoCloseable {
+
+  /**
+   * Create a Tensor with data from the given buffer.
+   *
+   * <p>Provide a {@link ByteBuffer} to create a Tensor of any datatype.  Note that primitive
+   * data must be in native byte order. Provide a {@link FloatBuffer}, {@link IntBuffer}, {@link DoubleBuffer},
+   * or {@link LongBuffer} to create a Tensor with a datatype of {@code FLOAT}, {@code INT32}, {@code DOUBLE},
+   * or {@code INT64} respectively.
+   *
+   * <p> This method copies
+   * <i>n</i>&nbsp;=&nbsp;<tt>src.remaining()</tt> elements from the given
+   * buffer into the tensor, starting at the buffer's current position.
+   * The position of the buffer is then incremented by <i>n</i>.
+   *
+   * @param dataType the tensor datatype.
+   * @param shape the tensor shape.
+   * @param data a buffer containing the tensor data.
+   */
+  public static Tensor create(DataType dataType, long[] shape, Buffer data) {
+    Tensor t = new Tensor();
+    t.dtype = dataType;
+    t.shapeCopy = Arrays.copyOf(shape, shape.length);
+
+    int elemsize = 0;
+    if(data instanceof ByteBuffer) elemsize = 1;
+    else if(data instanceof IntBuffer) elemsize = Integer.SIZE / Byte.SIZE;
+    else if(data instanceof FloatBuffer) elemsize = Float.SIZE / Byte.SIZE;
+    else if(data instanceof DoubleBuffer) elemsize = Double.SIZE / Byte.SIZE;
+    else if(data instanceof LongBuffer) elemsize = Long.SIZE / Byte.SIZE;
+    else throwIncompatibleBuffer();
+
+    t.nativeHandle = allocate(t.dtype.c(), t.shapeCopy, elemsize * data.remaining());
+    try {
+      ByteBuffer dst = getBuffer(t.nativeHandle);
+      if(data instanceof ByteBuffer) {
+        dst.put((ByteBuffer) data);
+      }
+      else if(data instanceof IntBuffer) {
+        if(t.dtype != DataType.INT32) throwIncompatibleBuffer();
+        dst.asIntBuffer().put((IntBuffer) data);
+      }
+      else if(data instanceof FloatBuffer) {
+        if(t.dtype != DataType.FLOAT) throwIncompatibleBuffer();
+        dst.asFloatBuffer().put((FloatBuffer) data);
+      }
+      else if(data instanceof DoubleBuffer) {
+        if(t.dtype != DataType.DOUBLE) throwIncompatibleBuffer();
+        dst.asDoubleBuffer().put((DoubleBuffer) data);
+      }
+      else if(data instanceof LongBuffer) {
+        if(t.dtype != DataType.INT64) throwIncompatibleBuffer();
+        dst.asLongBuffer().put((LongBuffer) data);
+      }
+
+      return t;
+    }
+    catch(RuntimeException e) {
+      delete(t.nativeHandle);
+      throw e;
+    }
+  }
+
+  /**
+   * Gets the size (in bytes) of the tensor data.
+   *
+   * <p>Use this size information when constructing a buffer with which to call {@link #readData(Buffer)}.
+   */
+  public int getDataByteSize() {
+    return getBuffer(nativeHandle).remaining();
+  }
+    
+  /**
+   * Reads the tensor data into the given buffer.
+   *
+   * <p>Provide a {@link ByteBuffer} to read a Tensor of any datatype.  Note that primitive
+   * data is in native byte order. Provide a {@link FloatBuffer}, {@link IntBuffer}, {@link DoubleBuffer},
+   * or {@link LongBuffer} to read a Tensor with a datatype of {@code FLOAT}, {@code INT32}, {@code DOUBLE},
+   * or {@code INT64} respectively.
+   *
+   * @param dst the destination buffer.
+   */
+  public void readData(Buffer dst) {
+    ByteBuffer src = getBuffer(nativeHandle);
+    if(dst instanceof ByteBuffer) {
+      ((ByteBuffer) dst).put(src);
+    }
+    else if(dst instanceof IntBuffer) {
+      if(dtype != DataType.INT32) throwIncompatibleBuffer();
+      ((IntBuffer) dst).put(src.asIntBuffer());
+    }
+    else if(dst instanceof FloatBuffer) {
+      if(dtype != DataType.FLOAT) throwIncompatibleBuffer();
+      ((FloatBuffer) dst).put(src.asFloatBuffer());
+    }
+    else if(dst instanceof DoubleBuffer) {
+      if(dtype != DataType.DOUBLE) throwIncompatibleBuffer();
+      ((DoubleBuffer) dst).put(src.asDoubleBuffer());
+    }
+    else if(dst instanceof LongBuffer) {
+      if(dtype != DataType.INT64) throwIncompatibleBuffer();
+      ((LongBuffer) dst).put(src.asLongBuffer());
+    }
+    else throwIncompatibleBuffer();
+  }
+
+  private static void throwIncompatibleBuffer() {
+    throw new IllegalArgumentException("incompatible buffer for Tensor data");
+  }
+
   /**
    * Create a Tensor from a Java object.
    *
@@ -71,7 +187,7 @@ public final class Tensor implements AutoCloseable {
     t.shapeCopy = new long[numDimensions(obj)];
     fillShape(obj, 0, t.shapeCopy);
     if (t.dtype != DataType.STRING) {
-      t.nativeHandle = allocate(t.dtype.c(), t.shapeCopy);
+      t.nativeHandle = allocateNDArray(t.dtype.c(), t.shapeCopy);
       setValue(t.nativeHandle, obj);
     } else if (t.shapeCopy.length != 0) {
       throw new UnsupportedOperationException(
@@ -317,7 +433,15 @@ public final class Tensor implements AutoCloseable {
     }
   }
 
-  private static native long allocate(int dtype, long[] shape);
+  private static ByteBuffer getBuffer(long nativeHandle) {
+    return buffer(nativeHandle).order(ByteOrder.nativeOrder());
+  }
+
+  private static native ByteBuffer buffer(long handle);
+
+  private static native long allocate(int dtype, long[] shape, long sizeInBytes);
+
+  private static native long allocateNDArray(int dtype, long[] shape);
 
   private static native long allocateScalarBytes(byte[] value);
 
--- a/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
@@ -17,6 +17,7 @@ package org.tensorflow;
 
 import java.lang.reflect.Array;
 import java.nio.BufferOverflowException;
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.DoubleBuffer;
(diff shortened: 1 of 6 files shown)


EronWright:
@asimshankar please have a look, thanks.

asimshankar:
Thanks for the PR! Took a quick look, and at a very high level this looks great. Will make a detailed pass a little later (apologies for the slow responses, I've been traveling but will be back next week and faster to respond thereafter :)

One thing to keep in mind: Since part of the intention here is for serialization, this isn't quite accurate. Turns out that the in-memory representation and serialized form aren't always the same (even if you ignore endian-ness).  Ideally we'd always use what is filled in [`TensorProto`](https://github.com/tensorflow/tensorflow/blob/9c41d5f64efedf44d30e5cc21f900ea07097395c/tensorflow/core/framework/tensor.proto#L36) as the serialized form, which is produced by [`Tensor::AsProtoTensorContent`](https://www.tensorflow.org/api_docs/cc/class/tensorflow/tensor#asprototensorcontent) in C++. 

So I would avoid committing to the `ByteBuffer` over the wire. Do keep that in mind. Though, at this stage I'm mostly interested in getting the API right, not having a comprehensive implementation in the short term is fine (For example, we did that with [Go](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/go/tensor.go#L196)).

EronWright:
I agree that TensorProto would be ideal for serialization, and spent some time on it.  But became discouraged when I realized that the generated classes don't use primitive arrays. For many scenarios, Buffers provide the ideal performance and flexibility.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,115 @@
...
+   * Reads the tensor data into the given buffer.
+   *
+   * <p>Provide a {@link ByteBuffer} to read a Tensor of any datatype.  Note that primitive
+   * data is in native byte order. Provide a {@link FloatBuffer}, {@link IntBuffer}, {@link DoubleBuffer},
+   * or {@link LongBuffer} to read a Tensor with a datatype of {@code FLOAT}, {@code INT32}, {@code DOUBLE},
+   * or {@code INT64} respectively.
+   *
+   * @param dst the destination buffer.
+   */
+  public void readData(Buffer dst) {
```
asimshankar: I'm a bit weary about all these `instanceof` statements and the possibility of `throwIncompatibleBuffer`. How about we make this just:

```java
public void copyTo(ByteBuffer dst);
```

Callers can use `dst.asFloatBuffer()` etc. as desired. This will also be consistent with the fact that we only provide guidance on the argument to `ByteBuffer.allocate()` (`getDataByteSize()`) and no direct guidance on the argument to `FloatBuffer.allocate()` etc.

tensorflow/java/src/main/native/tensor_jni.cc
```diff
@@ -218,6 +218,45 @@ size_t readNDArray(JNIEnv* env, TF_DataType dtype, const char* src,
 JNIEXPORT jlong JNICALL Java_org_tensorflow_Tensor_allocate(JNIEnv* env,
                                                             jclass clazz,
                                                             jint dtype,
+                                                            jlongArray shape,
+                                                            jlong size) {
+  int num_dims = static_cast<int>(env->GetArrayLength(shape));
```
asimshankar: Most of this is duplicated in `allocateNDArray`, perhaps structure this so that the code isn't duplicated (`allocateNDArray` can call `allocate` after determining the byte size). Though, see comment above about `allocate` and `allocateForStrings`.

tensorflow/java/src/main/native/tensor_jni.cc
```diff
@@ -433,3 +472,15 @@ JNIEXPORT void JNICALL Java_org_tensorflow_Tensor_readNDArray(JNIEnv* env,
...
+
+JNIEXPORT jobject JNICALL Java_org_tensorflow_Tensor_buffer(JNIEnv* env,
+                                                              jclass clazz,
+                                                              jlong handle) {
+  TF_Tensor* t = requireHandle(env, handle);
+  if (t == nullptr) return nullptr;
+  void* data = TF_TensorData(t);
+  const size_t sz = TF_TensorByteSize(t);
+
+  jobject dataBuffer = env->NewDirectByteBuffer(data, (jlong) sz);
```
asimshankar: Nits:
- Use `static_cast<jlong>` instead of c-style casts
- Perhaps just `return env->NewDirectByteBuffer(...)`

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -24,9 +24,262 @@
...
+import java.nio.*;
+import java.util.Arrays;
+
 /** Unit tests for {@link org.tensorflow.Tensor}. */
 @RunWith(JUnit4.class)
 public class TensorTest {
+  private static final double EPSILON = 1e-7;
+  private static final float EPSILON_F = 1e-7f;
+
+  // reusable sample data, varying by type and dimension
```
asimshankar: Not all of these are re-used between tests (for example, the shape ones are needed only for the `create` tests).

I'd suggest doing away with class variables and instead defining these within the test methods. That may come at the cost of some duplication, but the resulting tests are more readable if they are self contained.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -24,9 +24,262 @@
...
+      }
+      catch(IllegalArgumentException e) {
+      }
+    }
+  }
+
+  @Test
+  public void readData() {
+
+    Tensor tscalar = Tensor.create(scalar);
```
asimshankar: Let's use the [`try-with-resources`](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) statement here (which can handle multiple resources within the `try`):

```java
try (Tensor tscalar = Tensor.create(scalar);
       Tensor tvector = Tensor.create(vector);
       ....) {
   // Body of the test
}
```

That way you don't need the `finally` clause, don't have to worry about the `finally` clause covering all the tensors and it also takes care of any errors within the `create` calls.

tensorflow/java/src/test/java/org/tensorflow/TestUtil.java
```diff
@@ -49,4 +51,43 @@ public static Output matmul(
...
+        count += 1;
+      }
+      else {
+        count += numElements(e);
+      }
+    }
+    return count;
+  }
+
+  public static Object flatten(Object array, Class<?> elementType) {
```
asimshankar: I feel the two `flatten` methods warrant some comments :)

asimshankar (commented):
Thanks for the PR! Generally looks good, but I have some API comments.

Apologies for the delay, I was traveling, but am back now.

Thanks!

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -317,7 +433,15 @@ private void throwExceptionIfTypeIsIncompatible(Object o) {
...
   }
 
-  private static native long allocate(int dtype, long[] shape);
+  private static ByteBuffer getBuffer(long nativeHandle) {
+    return buffer(nativeHandle).order(ByteOrder.nativeOrder());
+  }
+
+  private static native ByteBuffer buffer(long handle);
+
+  private static native long allocate(int dtype, long[] shape, long sizeInBytes);
```
asimshankar: Or maybe just have a single:

```java
private static native long allocate(int dtype, long[] shape, long sizeInBytes);
```

Since it is a private method, we can make it require that `sizeInBytes` be consistent with `shape` and `dtype`. So we don't need two methods, just one?

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,115 @@
...
+   * <p> This method copies
+   * <i>n</i>&nbsp;=&nbsp;<tt>src.remaining()</tt> elements from the given
+   * buffer into the tensor, starting at the buffer's current position.
+   * The position of the buffer is then incremented by <i>n</i>.
+   *
+   * @param dataType the tensor datatype.
+   * @param shape the tensor shape.
+   * @param data a buffer containing the tensor data.
+   */
+  public static Tensor create(DataType dataType, long[] shape, Buffer data) {
```
EronWright: Believe it or not, my first draft was entirely `ByteBuffer` based.   I found it necessary to support other `Buffer` types to optimize the transfer of wrapped primitive arrays. 

Consider a `FloatBuffer` as produced by `FloatBuffer.wrap(new float[] { ... })`.  It is not convertible to a `ByteBuffer`.   To write it directly to the tensor's native `ByteBuffer`, we call `put` on a view of the latter.   At all times we avoid exposing the native buffer to user code.

I'm fine with moving to a set of overloaded methods to retain the benefits of other buffer types while reducing the possibility of runtime exceptions.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
EronWright: I imagined the term `read` to have `InputStream` semantics, but I hear you.

My preference would be `writeTo(Buffer)`, but can do `copyTo` if you prefer.   I'm personally not thrilled by how open-ended `copyTo` is, that's why I like the former.

EronWright: Keep in mind that `FloatBuffer`s that wrap an array are writable, so `copyTo` applied to a float buffer is more efficient than it would be if applied to a `ByteBuffer` (assuming you ultimately want a `float[]`).

I agree that the lack of guidance on sizing of typed buffers is inconvenient, but some utility methods might be enough.

Keeping with the power-over-convenience assumption, I'm inclined to provide optimized transfers.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
EronWright: I agree with your latter point.  If I recall correctly, it would require that the element counting code within the ND array codepath be moved from C to Java.  Since we have `flattenedNumElements` in the managed test code anyway, I have no objection.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
asimshankar: Okay, let's do the overloaded methods then.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
asimshankar: Let's do `writeTo(Buffer)`

asimshankar: Alright, I guess we keep the `instanceof` checks for now.
One minor thing, instead of `throwIncompatibleBuffer()`, mind make the helper function:

```java
private RuntimeException incompatibleBuffer(String got, String want) {
  throw new IllegalArgumentException(...)
}
```

And then at call sites do:
`throw incompatibleBuffer("Byte", "Float")`
(or something along those lines),

so that the stack trace is shorter and some more detail is provided.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
asimshankar: Doing the flattened counting in Java should be fine. So let's do that.

drpngx:
Any luck with this @EronWright ?

EronWright:
@drpngx ETA tomorrow.

drpngx:
No hurry, just checking. Nice contribution!

drpngx:
@asimshankar Just to be sure - is that good to go?

EronWright:
A comment on why the typed `create` methods take a `DataType` parameter, it is to support unsigned values in the future.   i.e.:
```java
IntBuffer buffer = ...
Tensor t = Tensor.create(DataType.UINT32, shape, buffer);
```

Note the JDK 8 support for unsigned values ([reference](https://blogs.oracle.com/darcy/entry/unsigned_api)).

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,352 @@
  * }</pre>
  */
 public final class Tensor implements AutoCloseable {
+
```
asimshankar: Nit: Can you place all these specialized `create()` methods after the one that creates from an `Object` so that all the `create()` methods are grouped together.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,352 @@
...
+      dst.put(data);
+      return t;
+    }
+    catch(RuntimeException e) {
+      delete(t.nativeHandle);
+      throw e;
+    }
+  }
+
+  private static Tensor create(DataType dataType, long[] shape, long byteSize) {
```
asimshankar: Nit: Perhaps call this `createHelper` or something else just to reinforce the fact that this is slightly different from all the other `create()` methods (it doesn't initialize the contents)

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,352 @@
...
+                + dataType());
+    }
+  }
+
+  /**
+   * Write the tensor data into the given buffer.
+   *
+   * <p>This method copies {@code numElements()} elements to the buffer.
+   *
+   * <p>This method may be used to read tensor data of type {@link DataType#INT32}.
```
asimshankar: Nit: Remove this line.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -237,6 +591,50 @@ long getNativeHandle() {
...
   private Tensor() {}
 
+  // buffer support
+
+  private ByteBuffer buffer() {
+    return buffer(nativeHandle).order(ByteOrder.nativeOrder());
+  }
+
+  private static IllegalArgumentException incompatibleBuffer(DataType dataType) {
+    return new IllegalArgumentException("incompatible buffer for Tensor of DataType " + dataType);
```
asimshankar: Perhaps can make this even more informative with:

```java
private static IllegalArgumentException incompatibleBuffer(Buffer buf, DataType dataType) {
  return new IllegalArgumentException("cannot use " + buf.getClass().getName() + " with Tensor of type " + dataType);
}
```

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -237,6 +591,50 @@ long getNativeHandle() {
...
+  }
+
+  private static IllegalArgumentException incompatibleBuffer(long[] shape) {
+    return new IllegalArgumentException("incompatible buffer for Tensor of shape " + Arrays.toString(shape));
```
asimshankar: Same here:

```java
private static IllegalArgumentException incompatibleBuffer(int numElements, long[] shape) {
  return new IllegalArgumentException("buffer with " + numElements + " elements is not compatible with a Tensor that has a shape " + Arrays.toString(shape));
}
```

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -237,6 +591,50 @@ long getNativeHandle() {
...
+      case INT32:
+        return Integer.SIZE / Byte.SIZE;
+      case INT64:
+        return Long.SIZE / Byte.SIZE;
+      case BOOL:
+        return 1;
+      case STRING:
+        return 0;
+      default:
+        throw new IllegalArgumentException("dataType");
```
asimshankar: Remove the quotes and expand on the message: 
```java
throw new IllegalArgumentException("unrecognized DataType " + dataType);
```

tensorflow/java/src/main/native/tensor_jni.cc
```diff
@@ -250,7 +241,7 @@ JNIEXPORT jlong JNICALL Java_org_tensorflow_Tensor_allocate(JNIEnv* env,
     dims_copy[i] = static_cast<int64_t>(dims[i]);
   }
   TF_Tensor* t = TF_AllocateTensor(static_cast<TF_DataType>(dtype), dims_copy,
-                                   num_dims, elem_size * num_elems);
+                                   num_dims, (size_t) sizeInBytes);
```
asimshankar:
```c++
static_cast<size_t>(sizeInBytes)
```

tensorflow/java/src/main/native/tensor_jni.h
```diff
@@ -24,11 +24,19 @@ extern "C" {
 
 /*
  * Class:     org_tensorflow_Tensor
+ * Method:    buffer
+ * Signature: (J)Ljava/nio/ByteBuffer;
+ */
+JNIEXPORT jobject JNICALL Java_org_tensorflow_Tensor_buffer(JNIEnv *, jclass,
```
asimshankar: Nit: Place this method in the `.cc` and `.h` files after `allocate` and `delete`

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -24,23 +24,278 @@
...
+import java.nio.*;
+
 /** Unit tests for {@link org.tensorflow.Tensor}. */
 @RunWith(JUnit4.class)
 public class TensorTest {
+  private static final double EPSILON = 1e-7;
+  private static final float EPSILON_F = 1e-7f;
+
+  @Test
+  @SuppressWarnings("unused")
```
asimshankar: Why is the `@SuppressWarnings` decorator needed?

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -24,23 +24,278 @@
...
+  public void createWithBuffer() {
+    int[] ints = {1, 2, 3, 4};
+    float[] floats = {1f, 2f, 3f, 4f};
+    double[] doubles = {1d, 2d, 3d, 4d};
+    long[] longs = {1L, 2L, 3L, 4L};
+    boolean[] bools = {true, false, true, false};
+    long[] shape = {4};
+
+    // validate creating a tensor using a direct byte buffer (in host order)
```
asimshankar: Would it make sense to split these into separate tests?

1. "byte" tests - like use of the direct buffer and the endian-ness flips

2. One for all the typed buffers (lines 68-91)

3. One for shape checking (lines 92-128)

4. One for datatype checking

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -24,23 +24,278 @@
...
+        fail("should have failed on incompatible buffer");
+      }
+      catch(IllegalArgumentException e) {
+        // expected
+      }
+    }
+  }
+
+  @Test
+  public void writeTo() {
```
asimshankar: If possible, I'd suggest breaking this test too into one per type.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,352 @@
...
+    t.nativeHandle = allocate(t.dtype.c(), t.shapeCopy, byteSize);
+    return t;
+  }
+
+  /**
+   * Gets the size (in bytes) of the tensor data.
+   *
+   * <p>Use this size information when constructing a byte buffer for {@code #writeTo(...)}.
+   */
+  public int byteSize() {
```
asimshankar: Perhaps call this `numBytes()`, so we will have `numBytes()` and `numElements()`

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,352 @@
...
+   *
+   * @throws  BufferOverflowException
+   *          If there is insufficient space in the given buffer
+   *          for the data in this tensor
+   * @throws  IllegalArgumentException
+   *          If the tensor datatype is not {@link DataType#INT32}
+   */
+  public void writeTo(IntBuffer dst) {
+    ByteBuffer src = buffer();
+    if(dtype != DataType.INT32) {
```
asimshankar: In all the `writeTo` methods, maybe check consistency before calling `buffer()`?

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -237,6 +591,50 @@ long getNativeHandle() {
...
+    for(int i = 0; i < shape.length; i++) {
+      n *= shape[i];
+    }
+    return n;
+  }
+
+  private static int elemByteSize(DataType dataType) {
+    switch(dataType) {
+      case FLOAT:
+        return Float.SIZE / Byte.SIZE;
```
asimshankar: It may just be clearer to return constants here. The byte sizes for TensorFlow data types are fixed, so let's return those?

```java
case FLOAT:
case INT32:
   return 4;
case DOUBLE:
case INT64:
  return 8;
case BOOL:
  return 1;
```

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -317,7 +715,9 @@ private void throwExceptionIfTypeIsIncompatible(Object o) {
     }
   }
 
-  private static native long allocate(int dtype, long[] shape);
+  private static native ByteBuffer buffer(long handle);
```
asimshankar: Place this after `allocate` and `delete`

asimshankar (commented):
Thanks for the update, I still suggest that we do without the `DataType` argument. And then some other minor things.

EronWright:
I plan to work on these changes this weekend.

asimshankar:
@EronWright : Any updates? We're looking forward to this :)

EronWright:
@asimshankar sorry for the delay, I need a few more days.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,352 @@
...
+   *          If the tensor datatype or shape is not compatible with the buffer
+   */
+  public static Tensor create(DataType dataType, long[] shape, IntBuffer data) {
+    if(dataType != DataType.INT32) {
+      throw incompatibleBuffer(dataType);
+    }
+    if(data.remaining() != numElements(shape)) {
+      throw incompatibleBuffer(shape);
+    }
+    int elemSize = Integer.SIZE / Byte.SIZE;
```
EronWright: Was unable to do this, tried to but the `ByteBuffer` variant was a special case.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -34,6 +41,352 @@
...
+    if(data.remaining() != numElements(shape)) {
+      throw incompatibleBuffer(shape);
+    }
+    int elemSize = Float.SIZE / Byte.SIZE;
+    Tensor t = create(dataType, shape, elemSize * data.remaining());
+    try {
+      ByteBuffer dst = t.buffer();
+      dst.asFloatBuffer().put(data);
+      return t;
+    }
```
EronWright: I do want to learn, and fixed the catch blocks accordingly.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
EronWright: I reworked it somewhat, separating out the `ByteBuffer` tests.

EronWright:
@asimshankar ready for you to have another look.  Thanks again.

asimshankar (approved):
Thanks very much for the contribution @EronWright ! This is sincerely appreciated.

Jenkins, test this please.

This should be ready to merge once tests pass.  I think the API is what we want it to be.

There are some minor style tweaks (a space here, a comma there), but not worth holding up this pull request. So let's merge it in and I'll make those cosmetic tweaks later on.

Thanks again!

drpngx:
Jenkins, test this please.

drpngx:
This fails in `//tensorflow/python:coordinator_test`. @rmlarsen, feel free to merge if this is a known failure.

vrv:
@asimshankar I believe GitHub now lets you edit PRs, so you can go ahead and make small typo changes yourself :)

rmlarsen:

rmlarsen:
It appears that the test failures are unrelated. Merging.

PR #6601, opened 2017-01-02
Added absolute path expansion of parent directory check in saver.py, …

PuchatekwSzortach:
…so as to correctly handle saving files in current directory

### DIFF
--- a/tensorflow/python/training/saver.py
+++ b/tensorflow/python/training/saver.py
@@ -1360,7 +1360,7 @@ class Saver(object):
             "'latest_filename' collides with 'save_path': '%s' and '%s'" %
             (latest_filename, save_path))
 
-    if not gfile.IsDirectory(os.path.abspath(os.path.dirname(save_path))):
+    if not gfile.IsDirectory(os.path.dirname(save_path)):
       raise ValueError(
           "Parent directory of {} doesn't exist, can't save.".format(save_path))
 
--- a/tensorflow/python/training/saver.py
+++ b/tensorflow/python/training/saver.py
@@ -1360,7 +1360,7 @@ class Saver(object):
             "'latest_filename' collides with 'save_path': '%s' and '%s'" %
             (latest_filename, save_path))
 
-    if not gfile.IsDirectory(os.path.dirname(save_path)):
+    if not gfile.IsDirectory(os.path.dirname(os.path.abspath(save_path))):
       raise ValueError(
           "Parent directory of {} doesn't exist, can't save.".format(save_path))
 
--- a/tensorflow/python/training/saver.py
+++ b/tensorflow/python/training/saver.py
@@ -1436,7 +1436,7 @@ class Saver(object):
     if self._is_empty:
       return
     sess.run(self.saver_def.restore_op_name,
-             {self.saver_def.filename_tensor_name: save_path})
+             {self.saver_def.filename_tensor_name: os.path.abspath(save_path)})
 
   @staticmethod
   def _add_collection_def(meta_graph_def, key, export_scope=None):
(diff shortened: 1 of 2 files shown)


PuchatekwSzortach:
Pull request in response to:
https://github.com/tensorflow/tensorflow/pull/3695#r94327053

rohan100jain:
Is it possible to add a test for the corner case this is trying to fix?

PuchatekwSzortach:
@rohan100jain Yes, but the test case passes on my OSX without the fix too... I assume it would fail on Windows and pass with the new check added, but don't have a Windows machine to really confirm that...
I can add the test case, but someone else would have to run it to see it really behaves as expected - especially that it fails with current code. If I add the test case here in comments, can someone run it to make sure it really fails on Windows? Or should I maybe first commit the test case alone without fix, make sure it fails when google bot runs tests and then add the fix and check that passes?

gunan:
If you can create the test as a separate commit, I can help you get it tested both with and without your fix.

PuchatekwSzortach:
@gunan 
Ok, this was a bit more complicated than I expected. First of all please ignore commits before 8a44d67, I made silly mistakes there.
Commit 8a44d67 adds a test that saves a graph to a file in current directory, using a path without "./" prepended.
I expect this to fail on Windows with parent not found error as per https://github.com/tensorflow/tensorflow/pull/6601#issuecomment-271063210. But what I didn't expect was that it would fail on OSX too - however not on saving the graph, but on restoring it. Adding a simple path expansion to body of Saver.restore(), commit b968f6, fixed that issue.

And commit 690b4ee (hopefully) fixes issue with not being able to save to current directory on Windows when no "./" is used in front of the path.

On a side note - it takes ~30min of rebuilding and re-running tests on my machine whenever I make a simple change to Tensorflow source code. Am I missing some caching settings, or is that a pain everyone here goes through? I could had solved the whole issue in 15 mins if not such slow rebuild + testing cycles...

drpngx:
Is this fixed?

gunan:
No, I still was not able to verify with the tests. Please assign to me and
block on me.
Sorry for the delay.

drpngx:
Can this move forward? I wonder if it's related to the problems that we've seen with "./" @rohan100jain

gunan:
OK, working on it right now.

gunan:
I just confirmed that the test case added, without the changes in saver.py do fail in windows.
I am now testing if I can get the test passing with saver.py changes.
In the meantime,
Jenkins, test this please.

gunan:
And with the fix the test passes.
Will proceed with the review.

gunan:
The issues should be unrelated, and fixed.
Jenkins, test this please.

gunan:
More flakes,
Jenkins, test this please.

gunan (approved):

jhseu:
This change breaks saving checkpoints to any non-local filesystem. Let's revert?

gunan:
Let's also discuss a plan for addressing the issue referenced by this PR.
Maybe we can add a method to wrap os.path.abspath, which checks if the path is local or remote and selectively converts the path?

jhseu:
I think the following version is reasonable:
1. Modify Dirname() to handle non-absolute paths (i.e., return "." for the Dirname() for cwd paths)
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/io/path.h

The return type would probably have to change to a std::string.
2. Expose Dirname() in gfile:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/platform/gfile.py
3. And call from save.

That should fix both save (#3) and restore (#1).

PR #6608, opened 2017-01-03
Detect and match against full cuda and cudnn versions.

davidzchen:
Currently, if a user specifies the full cuda or cudnn versions (with
major, minor, and patch numbers), cuda_configure will fail because we
only detect the version numbers we use (major and minor for cuda and
major only for cudnn). This change detects full versions and matches the
user-provided versions against the full versions. If the user provides
only partial version numbers (e.g. without patch version), then the
match will still succeed if the provided numbers match the detected
numbers.

This fixes the issue raised in this [StackOverflow post](http://stackoverflow.com/questions/41117980/cant-install-cudnn-5-1-5-with-cuda-8-0-and-tensorflow-0-12-on-g2-2xlarge-instan).

### DIFF
--- a/third_party/gpus/cuda_configure.bzl
+++ b/third_party/gpus/cuda_configure.bzl
@@ -147,6 +147,42 @@ def _cudnn_install_basedir(repository_ctx):
   return cudnn_install_path
 
 
+def _matches_version(environ_version, detected_version):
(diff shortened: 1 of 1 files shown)


petewarden (approved):
Looks good to me, thanks David!

Issue #5745, opened 2016-11-21
Constants are placed on a device when they do nothing but feed off-device nodes


DavidNorman:
I have created a device that supports integer constant tensors (because I would like to be able to support the inputs to reshape etc...), but not the integer maths operators (because I'm not going to accelerate those initially).

The gradient generation logic creates trees of integer maths.  The constants for this maths are placed on the device, while the maths nodes themselves are placed on the CPU (as expected).

This isn't very sensible.  

I think it would be better that constants are not placed on a device, unless they are feeding downstream nodes that are also on the same device.

There may be some classes of constant (ones with a specific memory placement, very large ones), which would end up on a different device to their downstream nodes, but I think that the default should be to keep them together.

prb12:
There is some logic in the constant folder which treats int32 ops differently when assigned to non-CPU devices.  This can be a pain when adding new device types and should probably be improved.

We should also probably take a look at the interaction of constant folding (and other optimizations) with the the two graph partitioning stages.

DavidNorman:
Thanks for the feedback.  For this particular issue, I would like to consider changing the placement algorithm so that constants are kept on the same device as the nodes that they feed.

An example of where this is important is as follows:

I support INT constants on my device, but I do not support the 'Add' operation for integer inputs.  As it stands, the placer will put the constants on my device and the Add op on the CPU.  After elaboration, I will have the constants on the device, feeding _Send ops and nothing else.  This doesn't make sense.

I will create a pull request for your evaluation.

DavidNorman:
Hmm.  I will investigate the behaviour of the simple_placer more closely.  At first glance it appears to contain logic for this constant placement already.

DavidNorman:
Ok.  I can see that you've already considered this.  There is code for doing what I would like, but it is failing to activate due to the simplicity of this 'IsGeneratorNode' function.  There is a comment that describes the solution already.  I will see about producing a pull request that solves this 'TODO'

```
// TODO(vrv): Currently this handles only nodes with one output, but
// this could be extended to handle the case where a node has many
// outputs that are connected to nodes in the same colocation group.
bool IsGeneratorNode(const Node* node) {
  return node->num_inputs() == 0 && node->num_outputs() == 1 &&
         node->out_edges().size() == 1 && !IsRefType(node->output_type(0));
}
```

DavidNorman:
[...]

Issue #5543, opened 2016-11-11
Constant folding doesn't remove control edges


DavidNorman:
I believe that when constant folding takes place, and a section of a graph is replaced by a constant, that only the data output edge of the replaced node is removed.  

I believe that I can see that a graph of nodes ending up in a Div (part of the gradient generation bit) is replaced by a Const.  The output of the Div goes to a Mul, and this is changed to the new const correctly.

However, there is a control output from the Div going to a Const (not sure why, but it is).  This is not changed to the Div replacement.  Consequently the dead node pruning doesn't remove the original Div.  

Here is some trace:

During the constant folding:

Graph Before #nodes 67 #edges 109
Graph Constant graph #nodes 32 #edges 42
Constant foldable 32 : 67

Replacing {name:'gradients/Mean_grad/truediv' id:28 op device:{/job:localhost/replica:0/task:0/device:ipu:0} def:{gradients/Mean_grad/truediv = Div[T=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:ipu:0"](gradients/Mean_grad/Tile, gradients/Mean_grad/Cast)}} :: 0 with a constant
Replacing edge to gradients/Square_grad/mul_1:0

During the post constant folding pruning:

PruneForReverseReachability: gradients/Square_grad/mul_1 <- gradients/Mean_grad/truediv/_0__cf__1
PruneForReverseReachability: gradients/Square_grad/mul/x <- gradients/Mean_grad/truediv

After the pruning:

Graph ConstFolding #nodes 68 #edges 110

When the graph is passed to the device, it contains:

Node: gradients/Square_grad/mul/x = Const[dtype=DT_FLOAT, value=Tensor<type: float shape: [] values: 2>, _device="/job:localhost/replica:0/task:0/device:ipu:0"](^gradients/Mean_grad/truediv)

DavidNorman:
Example:
```
import tensorflow as tf

# Creates a graph.
a = tf.placeholder(tf.float32, [2,2])
b = tf.placeholder(tf.float32, [2,2])
c = tf.constant([[2.0,2.0],[2.0,2.0]])
d = tf.constant([[2.0,2.0],[2.0,2.0]])
e = tf.add(c, d)
with tf.control_dependencies([e]):
  f = tf.mul(a, a)
  g = tf.mul(b, e)
  o = tf.add(f, g)

sess = tf.Session()

fd = {
  a: [[1.0,2.0],[3.0,4.0]],
[...]

PR #6615, opened 2017-01-03
Make placement of constants follow consumers if they are all on the same device

DavidNorman:
A generator node (for instant a constant) will now go to the same device as its consumers, when those consumers are all on the same device. Previously a constant would only follow its consumer when there was only a single consumer.


DavidNorman:
ok - i think we are going to need some help with this.   Is it that the same person who has made all of the commits that constitute a pull request must also make that pull request.  In which case, how do multiple people work on a feature in a fork, which is then included in a single pull request to the master repo?

vrv:
The two 'x's by both commits suggest that neither of the emails are being detected by the CLA.  The CLAbot would give a different message if all authors had signed the CLA (and that would be okay).

If your graphcore.ai is part of the googlegroup, can you also check to see that you've added that email to your GitHub account (public) ?  I think that's the only other requirement.

vrv:
(You should also generally squash your commits to avoid the merge commit in this case -- having a series of individual commits is typically only useful during the course of the review process, before merging, or for big feature additions, which typically should be sent as smaller incremental PRs anyway).

DavidNorman:
Ok.  Will merge in recent head changes and squash all commits. 

I see the problem with the emails.  I guess I am commiting using an email that is not recognized by GitHub as an account.

DavidNorman:
The email thing worked.  

The 2 merge commits are just where I was pulling the main tensorflow master to the repo to verify that the changes still work.  i wasn't expecting the change to take so long to get into the repo, so i didn't isolate the commit onto a branch originally.

tensorflow/core/common_runtime/simple_placer.cc
```diff
@@ -785,9 +785,23 @@ Status SimplePlacer::Run() {
...
+
+      bool all_consumers_on_same_device(true);
+      while (it != node->out_edges().end()) {
+        if ((*it)->dst()->assigned_device_name() != output_device_name) {
+          all_consumers_on_same_device = false;
+          break;
+        }
+        it++;
+      }
+
```
vrv: You might be able to replace this entire block of code with:

```
    bool same_device = std::all_of(node->edges().begin(), node->edges().end(), 
                                   [output_device_name](const Edge& e) { 
                                       return e->dst()->assigned_device_name() == output_device_name;
                                   }) 
```

If it turns out to be more complicated than this, let me know.

tensorflow/core/common_runtime/simple_placer_test.cc
```diff
@@ -1138,5 +1138,98 @@ TEST_F(SimplePlacerTest, TestUnsatisfiableConstraintWithReferenceConnections) {
...
+    Node* input = ops::SourceOp("TestCPUGPUOutput", b.opts().WithName("in"));
+
+    // The assigns are bound to CPU by the reference edge.
+    ops::BinaryOp("TestAssign", var1_cpu, input, b.opts().WithName("assign1"));
+    ops::BinaryOp("TestAssign", var2_cpu, input, b.opts().WithName("assign2"));
+
+    TF_EXPECT_OK(BuildGraph(b, &g));
+
+    GetNodeByName(g, "var1_cpu")
+    ->set_assigned_device_name("/job:a/replica:0/task:0/cpu:1");
```
vrv: can you run clang-format --style=google on this file to get consistency in spacing?

tensorflow/core/common_runtime/simple_placer_test.cc
```diff
@@ -1138,5 +1138,98 @@ TEST_F(SimplePlacerTest, TestUnsatisfiableConstraintWithReferenceConnections) {
...
+    ->set_assigned_device_name("/job:a/replica:0/task:0/cpu:2");
+  }
+
+  TF_EXPECT_OK(Place(&g));
+  EXPECT_COLOCATED(g, "assign1", "var1_cpu");
+  EXPECT_COLOCATED(g, "assign2", "var2_cpu");
+  EXPECT_DEVICE_TYPE(g, "in", DEVICE_GPU);
+}
+
+// Test that a generator node follows its consumer (where there is
```
vrv: Is this test any different than https://github.com/DavidNorman/tensorflow/blob/314957c888cddb183c12b41927b76aed9f26f96b/tensorflow/core/common_runtime/simple_placer_test.cc#L313

vrv (changes requested):
Thanks!  Just a few comments, thanks for the nice tests.

tensorflow/core/common_runtime/simple_placer_test.cc
DavidNorman: You're right - its the same. My test is differentiating from the previous two in terms of number of consumers.  However, the single consumer case is tested by the 'TestHeuristicA' test.  I'll remove this test (although I prefer my name to HeuristicA :)

tensorflow/core/common_runtime/simple_placer.cc
DavidNorman: My code pulls the name from the 1st iterator value, then tests it against all of the others.  Your code captures 'output_device_name', which is the name from the 1st iterator value.  This would need to be read in your version of the code too.

What we need is a:

std::when_calling_with_elements_from_a_collection_every_return_value_of_a_functor_is_the_same()

It might exist.  My in-depth knowledge of the std c++ library is limited to pre 2004.

tensorflow/core/common_runtime/simple_placer_test.cc
```diff
@@ -1138,5 +1138,76 @@ TEST_F(SimplePlacerTest, TestUnsatisfiableConstraintWithReferenceConnections) {
                   .contains("Cannot colocate nodes 'var' and 'assign'"));
 }
 
+// Test that a generator node follows its consumers (where there are several
```
vrv: (Feel free to rename the TestHeuristicA to your name)!

vrv (approved):

vrv:
(We can update the name later, let's get this in)

DavidNorman:
hmm.. I didn't run any of the tests but my own one.  I'll run them all now and see if i can track down these failures.  I guess there is a higher level knock on effect maybe.  Seems odd that the CPU only tests would change though.  I would assume that the placement is always on the same device.

DavidNorman:
the failing test does relate to the placement of constants.  it is strange that it should fail, because it explicitly places the constants, and I was pretty sure that explicitly placed things are not affected by my change.  apparently not

DavidNorman:
I'm not sure i fully understand what is going on in the placer.  I have traced some of the state.

```
Graphdef : versions = producer: 20;
const1 = Const[dtype=DT_FLOAT, value=Tensor<type: float shape: [] values: 1>, _device="/device:CPU:0"]();
const2 = Const[dtype=DT_FLOAT, value=Tensor<type: float shape: [] values: 2>, _device="/device:CPU:1"]();
add = Add[T=DT_FLOAT, _device="/device:CPU:2"](const1, const2);
mul = Mul[T=DT_FLOAT, _device="/device:CPU:2"](const1, const2);
add_1 = Add[T=DT_FLOAT, _device="/device:CPU:2"](add, mul);

Node const2 has assigned name 
Node const1 has assigned name 
Node mul has assigned name 
Node add has assigned name 
Node add_1 has assigned name 
.[u'/job:localhost/replica:0/task:0/cpu:1', u'/job:localhost/replica:0/task:0/cpu:0', u'/job:localhost/replica:0/task:0/cpu:2']
```

the first batch is the graphdef showing that the nodes in the testAnalysisAndAllocations test (of //tensorflow/python:timeline_test) know the explicit placements that have been made for them.  The 'Node const2 has assigned name' is saying that node->assigned_device_name() in simple_placer.cc has no value (is skipped in the test on line ~720).

Why isn't the assigned name that is present in the graphdef present in the placer where it is deliberately skipping over things that are pre-placed?

DavidNorman:
Ok.  Now I have dumped the Node*.  It is:

```
Node {name:'const2' id:2 op device:{} def:{const2 = Const[dtype=DT_FLOAT, value=Tensor<type: float shape: [] values: 2>, _device="/device:CPU:0"]()}}
```

This shows that the 'op device' is {}, meaning there is no assigned device, but the NodeDef has _device set correctly.  I shall see if I can find out where one is assigned to the other (if this is what happens at all)

DavidNorman:
So... I think what is happening is that the placement algorithm goes something like this:

```
pass A
1) for each node check if it has already been placed (not by the user but by the algorithm)
2) defer 'generator nodes' to a later pass
3) place nodes according to possible placements (according to GetDevicesForNodes - which is where the explict user-specified location comes in)

pass B
1) place generator nodes with their consumers
```

so generator nodes are not guided by the user's explicit placement commands.  the test 'timeline_test' is only valid when the definition of generator nodes does not include nodes which feed more than one consumer.  I have changed the definition to include multiple consumer nodes, so the test doesn't work any more.

not sure what to do about this.  I could doctor up the test to avoid using generator nodes. 


do you have any advice?

vrv:
Hm, that might be a bug that the generator nodes are not getting their explicit device placements -- let me test a change to see.

vrv:
Okay, there are likely two issues here:

1) That skip to check for already assigned devices is bogus (it's a no-op right now, must have been due to a change).

2) Heuristics should only apply if there are no device specifications, even partial ones (at least, that is the intention right now).

I'll send a fix and a test for this.

vrv (changes requested):
(Let me send in my fix + test first).

vrv:
Okay, after talking with mrry@, we found out this is wrong.

1) The check for already assigned devices is necessary for stateful placements (they get set in simple_graph_execution_state.cc, and must be skipped).

2) The heuristic applies only after checking that to-be-assigned device is in the list of the devices (potentially determined by a partial or full device spec), so there's no violation going on here.

I'll have to dig in more to figure out what's going n.

DavidNorman:
indeed it did seem like the check for existing device placement was a no-op, but i was expecting there to be an outside case that I wasn't aware of.

it seems that the generator placement doesn't follow the explicit placement though.   the explicit placement comes in through the GetDevicesForNode.  In the case of a generator node, even if there is an explicit placement, the assigned node is determined by the location of the consumer.   This would have applied when a generator node had only one consumer, too.

I could generate a simple python test that demonstrates the problem, if you would like.

DavidNorman:
Actually - with reference to your point 1 - i would be tempted to put a comment against the pre- assigned_name test referencing that.   otherwise it seems a bit opaque when someone is running stuff through the debugger.

vrv:
Indeed -- I think I have a better fix coming (including a test for the need for the stateful placements and a comment).

DavidNorman:
nice one :)

vrv:
Okay, about to submit a fix and some more tests -- should be available later today if our pushing process goes well.

Basically, CanAssignToDevice was only checking that the device type was valid, not that the device was in the list generated by the placer (which uses the user's device spec).  The change I am submitting, makes CanAssignToDevice check that the consumer's device is in the list of the potential devices.

The reason timeline_test failed is that the consumer was on /cpu:1, the producer was on /cpu:2 (explicit), and because the device types were the same, it moved producer to /cpu:1, which causes the test to fail.  With this change, the producer will always remain on /cpu:2.

Lastly, the check for assigned_devices / continuing could be removed without breaking any test, but is an optimization (there is no need to do more work to check for metadata / generators when you know the device has been assigned already), so I just added a comment to make it more clear.

After it gets pushed, we can merge this PR and things should get better.

DavidNorman:
I can see that your commit is in, and I like it. 

Perhaps this will cause the buildbot to check the code:

DavidNorman:
i can also see that a local run of timeline test is ok.   I can also see that XLA is in the repo, which is good.  I think I shall have to spend some time browsing that code base now.  

Hopefully you don't think that this change has been a big time hole for you.  It has been very useful for me, especially to get a feeling for the process of pushing code into the public repo.  I guess I will have to do more of this in the future.

It is possible that XLA eliminates the problem that I was trying to solve with this change, although I suspect that it will be useful in general (until the placer is re-written), and also it helped to expose the thing where an explicitly placed constant would still follow its consumer.

I will, soon, do another merge request for the fix for my other issue (where a tree with a constant value remains on a device even when replaced by a single constant, due to the presence of control edges).

tensorflow/core/common_runtime/simple_placer_test.cc
DavidNorman: as you suggested, i think we can leave the name alone.

vrv:

Thanks!

vrv:
oh wait, you have to merge to master to get my changes.  the tests are going to fail again.

vrv:
and no, no problem at all - you helped us find a subtle bug, and since not all devices will end up wanting to use XLA, these heuristics are still useful (if even as examples of the types of heuristics that we will want to maintain in the future),  so thank you for pushing this change forward! :)

DavidNorman:
it needed a little change to the test to make it compatible with the recent changes to the test device names.

DavidNorman:
hopefully it will be ok now.

if this works, I may push through my one and only other 'fix', which is a fix for:

https://github.com/tensorflow/tensorflow/issues/5543


Incidentally - this change is the fix for:

https://github.com/tensorflow/tensorflow/issues/5745

drpngx:
Jenkins, test this please.

drpngx:
@vrv if this doesn't look right, feel free to comment on this thread, o/w I'll merge.

vrv (approved):

DavidNorman:
cheers.

PR #6617, opened 2017-01-03
Make *args in sv.loop example an iterable

carlthome:


gunan:
Jenkins, test this please.

gunan (approved):

PR #6632, opened 2017-01-04
Add adaptive softmax implemtation in nn_impl.py

SteveSYYang:
Add a implementation of AdaptiveSoftmax, see [Efficient softmax approximation for GPUs](https://arxiv.org/pdf/1609.04309v2.pdf) for detail.

The adaptive softmax is a faster way to train a softmax classifier over a huge number of classes, and can be used for both training and prediction. For example, it can be used for training a language model with a very huge vocabulary, and the trained languaed model can be used in speech recognition, text generation, and machine translation very efficiently .

It has been tested for language modeling on PTB (Penn Treebank) and GBW (Google Billion Word) corpus, and achieved a better result than the [Torch implementation](https://github.com/facebookresearch/adaptive-softmax). 

On GBW corpus, we achived a perplexcity of 43.24 after 5 epochs, taking about two days to train on 2 GPUs with synchronous gradient updates. Detail experiment result and usage demo can be found here: [TencentAILab/tf-adaptive-softmax-lstm-lm](https://github.com/TencentAILab/tf-adaptive-softmax-lstm-lm).

Further more, it has been used in the ASR system developed by Tencent AI Lab, and achieved about **20x speed up** than the full sotfmax in the second pass for rescoing.

### DIFF
--- a/tensorflow/contrib/layers/python/layers/layers.py
+++ b/tensorflow/contrib/layers/python/layers/layers.py
@@ -22,6 +22,7 @@ from __future__ import print_function
 
 import functools
 import six
+import math
 
 from tensorflow.contrib.framework.python.ops import add_arg_scope
 from tensorflow.contrib.framework.python.ops import variables
@@ -75,7 +76,8 @@ __all__ = ['avg_pool2d',
            'unit_norm',
            'legacy_fully_connected',
            'legacy_linear',
-           'legacy_relu']
+           'legacy_relu',
+           'adaptive_softmax_loss']
 
 DATA_FORMAT_NCHW = 'NCHW'
 DATA_FORMAT_NHWC = 'NHWC'
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
 
     return _apply_activation(y, activation_fn, output_collections)
 
+def adaptive_softmax_loss(inputs,
+                          labels,
+                          cutoff,
+                          project_factor=4,
+                          initializer=None,
+                          name=None):
+  """Computes and returns the adaptive softmax loss (a improvement of 
+  hierarchical softmax).
+    
+  See [Efficient softmax approximation for GPUs](https://arxiv.org/pdf/1609.04309v2.pdf).
+        
+  This is a faster way to train a softmax classifier over a huge number of 
+  classes, and can be used for **both training and prediction**. For example, it 
+  can be used for training a Language Model with a very huge vocabulary, and 
+  the trained languaed model can be used in speech recognition, text generation, 
+  and machine translation very efficiently.
+  
+  Args:
+    inputs: A `Tensor` of shape `[batch_size, dim]`.  The forward
+      activations of the input network.
+    labels: `Tensor` of shape `[d_0, d_1, ..., d_{r-2}]` and dtype `int32` or
+      `int64`. Each entry in `labels` must be an index in `[0, num_classes)`.
+    cutoff: A list indicating the limits of the different clusters.
+    project_factor: A floating point value greater or equal to 1.0. The projection 
+      factor between two neighboring clusters.
+    initializer: Initializer for adaptive softmax variables (optional).
+    name: A name for the operation (optional).
+
+  Returns:
+    loss: A `batch_size` 1-D tensor of the adaptive softmax cross entropy loss.
+    training_losses: A list of 1-D tensors of adaptive softmax loss for each 
+      cluster, which can be used for calculating the gradients and back 
+      propagation when training.
+  """
+  input_dim = int(inputs.get_shape()[1])
+  sample_num = int(inputs.get_shape()[0])
+  cluster_num = len(cutoff) - 1
+  with ops.name_scope(name, "AdaptiveSoftmax"):
+    if initializer is None:
+      stdv = math.sqrt(1. / input_dim)
+      initializer = init_ops.random_uniform_initializer(-stdv * 0.8, stdv * 0.8)
+
+    head_dim = cutoff[0] + cluster_num
+    head_w = variable_scope.get_variable("adaptive_softmax_head_w", 
+                             [input_dim, head_dim], initializer=initializer)
+
+    tail_project_factor = project_factor
+    tail_w = []
+    for i in range(cluster_num):
+      project_dim = max(1, input_dim // tail_project_factor)
+      tail_dim = cutoff[i + 1] - cutoff[i]
+      tail_w.append([
+        variable_scope.get_variable("adaptive_softmax_tail{}_proj_w".format(i+1), 
+                        [input_dim, project_dim], initializer=initializer),
+        variable_scope.get_variable("adaptive_softmax_tail{}_w".format(i+1), 
+                        [project_dim, tail_dim], initializer=initializer)
+      ])
+      tail_project_factor *= project_factor
+
+    # Get tail masks and update head labels
+    training_losses = []
+    loss = array_ops.zeros([sample_num], dtype=dtypes.float32)
+    head_labels = labels
+    for i in range(cluster_num):
+      mask = math_ops.logical_and(math_ops.greater_equal(labels, cutoff[i]), 
+                                  math_ops.less(labels, cutoff[i + 1]))
+      
+      # Update head labels
+      head_labels = math_ops.select(mask, array_ops.constant([cutoff[0] + i] * 
+                            sample_num), head_labels)
+
+      # Compute tail loss
+      tail_inputs = array_ops.boolean_mask(inputs, mask)
+      tail_logits = math_ops.matmul(math_ops.matmul(tail_inputs, tail_w[i][0]), 
+                                    tail_w[i][1])
+      tail_labels = array_ops.boolean_mask(labels - cutoff[i], mask)
+      tail_loss = nn.sparse_softmax_cross_entropy_with_logits(tail_logits, 
+                                                                  tail_labels)
+      training_losses.append(tail_loss)
+      aligned_tail_loss = sparse_tensor.SparseTensor(
+        array_ops.squeeze(array_ops.where(mask)), tail_loss, [sample_num])
+      loss += sparse_ops.sparse_tensor_to_dense(aligned_tail_loss)
+
+    # Compute head loss
+    head_logits = math_ops.matmul(inputs, head_w)
+    head_loss = nn.sparse_softmax_cross_entropy_with_logits(head_logits, 
+                                                                head_labels)
+    loss += head_loss
+    training_losses.append(head_loss)
+
+    return loss, training_losses
 
 # TODO(eiderm): Verify and fix autocomplete in colab (also relu6).
 # Simple aliases which remove the activation_fn parameter.
--- a/tensorflow/python/ops/nn_impl.py
+++ b/tensorflow/python/ops/nn_impl.py
@@ -1173,4 +1173,3 @@ def sampled_softmax_loss(weights,
   sampled_losses = nn_ops.softmax_cross_entropy_with_logits(logits, labels)
   # sampled_losses is a [batch_size] tensor.
   return sampled_losses
-  
--- a/tensorflow/python/ops/nn_impl.py
+++ b/tensorflow/python/ops/nn_impl.py
@@ -30,8 +30,6 @@ from tensorflow.python.ops import gen_nn_ops
 from tensorflow.python.ops import math_ops
 from tensorflow.python.ops import nn_ops
 from tensorflow.python.ops import sparse_ops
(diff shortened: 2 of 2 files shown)


SteveSYYang:
I signed a CLA!

drpngx:
Please provide a link to the paper from the code.

drpngx:
@ebrevdo want to take a look?

ebrevdo:
Steve, thanks for the contribution!  We'll check internally that we have
someone who has cycles to maintain this function on our side.  If we do, we
can accept it.

Keep in mind that code like this should be added to tf.contrib first, and
not core.  If we decide that we can accept it, you'll have to move it to
somewhere in contrib (perhaps tf.contrib.losses is the appropriate module)

martinwicke (commented):
Since this function creates variables, I believe this functionality should probably go into layers. 

We're trying to keep tf.nn to pure functions, so that we can more easily see which changes will lead to checkpoint incompatibilities.

ebrevdo:
To follow up:

1. Let's get your patch submitted.
2. Please refactor to move the functionality to tf.contrib.layers; follow
all conventions of the code therein (python cleanliness, op and variable
names, etc).

Ping me when ready for review.

SteveSYYang:
@ebrevdo thanks for your help. I have moved the adaptive_softma_loss to tf.contrib.layers, could you please review the code.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
 
     return _apply_activation(y, activation_fn, output_collections)
 
+def adaptive_softmax_loss(inputs,
+                          labels,
+                          cutoff,
+                          project_factor=4,
+                          initializer=None,
+                          name=None):
+  """Computes and returns the adaptive softmax loss (a improvement of
```
ebrevdo: The initial line should be a complete summary that ends in a period (not two lines like you have here)

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+def adaptive_softmax_loss(inputs,
+                          labels,
+                          cutoff,
+                          project_factor=4,
+                          initializer=None,
+                          name=None):
+  """Computes and returns the adaptive softmax loss (a improvement of 
+  hierarchical softmax).
+    
+  See [Efficient softmax approximation for GPUs](https://arxiv.org/pdf/1609.04309v2.pdf).
```
ebrevdo: can you link to the abs?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+        
+  This is a faster way to train a softmax classifier over a huge number of
```
ebrevdo: Describe what softmax is; try to make this similar to the original softmax cross entropy loss function.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
 
     return _apply_activation(y, activation_fn, output_collections)
 
+def adaptive_softmax_loss(inputs,
```
ebrevdo: is this a softmax cross entropy loss?  it's ok to have a longer name.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+  This is a faster way to train a softmax classifier over a huge number of 
+  classes, and can be used for **both training and prediction**. For example, it 
+  can be used for training a Language Model with a very huge vocabulary, and 
+  the trained languaed model can be used in speech recognition, text generation, 
+  and machine translation very efficiently.
+  
+  Args:
+    inputs: A `Tensor` of shape `[batch_size, dim]`.  The forward
+      activations of the input network.
+    labels: `Tensor` of shape `[d_0, d_1, ..., d_{r-2}]` and dtype `int32` or
```
ebrevdo: What's r?  How is it related to the shape of inputs?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+      `int64`. Each entry in `labels` must be an index in `[0, num_classes)`.
```
ebrevdo: Where is num_clsases?  How is it related to the shape of inputs?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+    cutoff: A list indicating the limits of the different clusters.
```
ebrevdo: Python list?  Tensorflow vector?  What types?  What are the constraints on the values?  What's a cluster in context here?  Document that last part at the top.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+    project_factor: A floating point value greater or equal to 1.0. The projection
```
ebrevdo: Python float scalar, or can this be a tensorflow float16/float32/float64 scalar?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+  
+  Args:
+    inputs: A `Tensor` of shape `[batch_size, dim]`.  The forward
+      activations of the input network.
+    labels: `Tensor` of shape `[d_0, d_1, ..., d_{r-2}]` and dtype `int32` or
+      `int64`. Each entry in `labels` must be an index in `[0, num_classes)`.
+    cutoff: A list indicating the limits of the different clusters.
+    project_factor: A floating point value greater or equal to 1.0. The projection 
+      factor between two neighboring clusters.
+    initializer: Initializer for adaptive softmax variables (optional).
```
ebrevdo: Can this initializer just come from the variable scope?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+    name: A name for the operation (optional).
+
+  Returns:
+    loss: A `batch_size` 1-D tensor of the adaptive softmax cross entropy loss.
```
ebrevdo: Can you backprop through this?

At the top can you give examples of how this layer should be constructed, and used, with backprop operations?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+    training_losses: A list of 1-D tensors of adaptive softmax loss for each 
+      cluster, which can be used for calculating the gradients and back 
+      propagation when training.
+  """
+  input_dim = int(inputs.get_shape()[1])
```
ebrevdo: this will fail if inputs.get_shape().ndims is None or if get_shape()[1] is None.  same for the line below.

is it critical that these values are known at graph build time, or can you use TF ops here?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+  sample_num = int(inputs.get_shape()[0])
+  cluster_num = len(cutoff) - 1
```
ebrevdo: should this match any other values, do you need to add any checks?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+  with ops.name_scope(name, "AdaptiveSoftmax"):
+    if initializer is None:
```
ebrevdo: is this from the paper, or is this a standard initializer?  does the default variable scope initializer do a sufficient job?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+      stdv = math.sqrt(1. / input_dim)
+      initializer = init_ops.random_uniform_initializer(-stdv * 0.8, stdv * 0.8)
+
+    head_dim = cutoff[0] + cluster_num
```
ebrevdo: do you check that all values in cutoff are numeric?

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+    head_w = variable_scope.get_variable("adaptive_softmax_head_w", 
+                             [input_dim, head_dim], initializer=initializer)
```
ebrevdo: this makes me think that users can just create this layer with an outer variable scope that creates an initializer; unless you have a good reason for picking the one you did.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+                        [project_dim, tail_dim], initializer=initializer)
+      ])
+      tail_project_factor *= project_factor
+
+    # Get tail masks and update head labels
+    training_losses = []
+    loss = array_ops.zeros([sample_num], dtype=dtypes.float32)
+    head_labels = labels
+    for i in range(cluster_num):
+      mask = math_ops.logical_and(math_ops.greater_equal(labels, cutoff[i]),
```
ebrevdo: which of the inputs here are tensors and which are python values?  either way you can use >= instead of greater_equal and < instead of math_ops.less

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+    # Get tail masks and update head labels
+    training_losses = []
+    loss = array_ops.zeros([sample_num], dtype=dtypes.float32)
+    head_labels = labels
+    for i in range(cluster_num):
+      mask = math_ops.logical_and(math_ops.greater_equal(labels, cutoff[i]), 
+                                  math_ops.less(labels, cutoff[i + 1]))
+      
+      # Update head labels
+      head_labels = math_ops.select(mask, array_ops.constant([cutoff[0] + i] *
```
ebrevdo: math_ops.where instead of select, now.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+      tail_inputs = array_ops.boolean_mask(inputs, mask)
+      tail_logits = math_ops.matmul(math_ops.matmul(tail_inputs, tail_w[i][0]), 
+                                    tail_w[i][1])
+      tail_labels = array_ops.boolean_mask(labels - cutoff[i], mask)
+      tail_loss = nn.sparse_softmax_cross_entropy_with_logits(tail_logits, 
+                                                                  tail_labels)
+      training_losses.append(tail_loss)
+      aligned_tail_loss = sparse_tensor.SparseTensor(
+        array_ops.squeeze(array_ops.where(mask)), tail_loss, [sample_num])
+      loss += sparse_ops.sparse_tensor_to_dense(aligned_tail_loss)
```
ebrevdo: see if you can use array_ops.scatter_nd here.  that one has a gradient.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -2129,6 +2131,97 @@ def legacy_fully_connected(x,
...
+
+    # Compute head loss
+    head_logits = math_ops.matmul(inputs, head_w)
+    head_loss = nn.sparse_softmax_cross_entropy_with_logits(head_logits, 
+                                                                head_labels)
```
ebrevdo: fix whitespace issues everywhere.

ebrevdo (changes requested):
please add comprehensive unit tests, esp. ones we can verify match the results in the paper (if any)

rmlarsen:
@SteveSYYang could you address the reviewers comments, please?

vrv:
Closing due to lack of update to keep our list of active PRs small -- feel free to update this PR or send a new one once the comments have been addressed.

PR #6641, opened 2017-01-04
configure: Handle existing TF_CUDNN_VERSION correctly on OSX.

pwnall:
The configure script sets TF_CUDNN_EXT based on the desired CuDNN version. When TF_CUDNN_VERSION does not exist, the script correctly extracts the version from CuDNN's library symlink and uses it to populate TF_CUDNN_EXT. When TF_CUDNN_VERSION does exist, the script uses it to derive TF_CUDNN_EXT.

Unfortunately, the logic for the latter case does not cover the special case in OSX naming. This commit adds logic to handle the special case in OSX.

### DIFF
--- a/configure
+++ b/configure
@@ -289,7 +289,11 @@ while true; do
       echo "libcudnn.dylib resolves to libcudnn${TF_CUDNN_EXT}"
     fi
   else
-    TF_CUDNN_EXT=".$TF_CUDNN_VERSION"
+    if [ "$OSNAME" == "Darwin" ]; then
+      TF_CUDNN_EXT=".${TF_CUDNN_VERSION}.dylib"
+    else
+      TF_CUDNN_EXT=".$TF_CUDNN_VERSION"
+    fi
   fi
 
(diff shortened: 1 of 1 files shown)


caisq (approved):

caisq:

caisq:
PR merged. Thanks, @pwnall

pwnall:
Thank you very much for the quick turnaround time!

FWIW, this unblocks my work-in-progress [Homebrew formula](https://github.com/pwnall/homebrew-science/blob/tensorflow/tensorflow.rb).

caisq:
Good to know, @pwnall Thank you!

CC @gunan @yifeif

pwnall:
@caisq Any chance this could be merged into the r1.0 branch? It'd be awesome if it'd be possible to install Tensorflow 1.0 with Homebrew.

gunan:
It will be merged to 1.0 branch before the next binary release.
Thanks for the contribution!

PR #6651, opened 2017-01-05
change tensorboard html data request path to relative

lspvic:
Tensorboard use a  absolute path  for data request like `/data/runs`, `/data/graph`, `/data/logdir`.

If someone want to deploy tensorboard with a **path** `http://host/tensorboard/`, the data xmlhttprequest would be  `http://host/data/runs`,  thus give not found 404 errors and tensorboard doesn't work properly.

To make the data request with `http://host/tensorboard/data/runs`, a relative  xmlhttprequest url for data request should be used.  The default `dataDir = '/data';` will change to `dataDir = 'data';`

The default  dataDir to `dataDir = 'data'` will always work whenever `tf-tensorboard.html` is placed or any context path is set.

### DIFF
--- a/tensorflow/tensorboard/components/tf_backend/router.ts
+++ b/tensorflow/tensorboard/components/tf_backend/router.ts
@@ -35,7 +35,7 @@ module TF.Backend {
    * @param dataDir {string} The base prefix for finding data on server.
    * @param demoMode {boolean} Whether to modify urls for filesystem demo usage.
    */
-  export function router(dataDir = '/data', demoMode = false): Router {
+  export function router(dataDir = 'data', demoMode = false): Router {
     var clean = demoMode ? demoify : (x) => x;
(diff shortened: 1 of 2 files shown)


danmane:
Changing the dist/tf-tensorboard.html file alone isn't enough. Can you please change the Typescript source too (I believe it will be router.ts)

lspvic:
@danmane  I have done it, thanks for your reminding.

rmlarsen:
@dandelionmane Can you take another look, please?

dandelionmane:
Sorry for losing track of this. I already made the change (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorboard/components/tf_backend/router.ts), and we will be re-generating the tf-tensorboard.html file with this change and more this week, so I'm going to close the pull request.

Issue #547, opened 2015-12-18
Tensorflow MaxPool doesn't accept float64


ketabdarhamed:
I am using Tensorflow to train a CNN. I am currently basing my calculations on Float32 which is kind of default at the time of initialising variables.

I guessed that by using float64as my dtype I can get more accurate results so I changed the initilization of my varibales as follows:

`w1 = tf.Variable(tf.random_normal([5, 5, 3, 64], stddev=0.1, dtype=tf.float64))`
But I get following error in the maxpooloperation: I chacked the maxpooldocumentation and it accept value type as follow: `value: A 4-D Tensor with shape [batch, height, width, channels] and type float32, float64, qint8, quint8, qint32.`

But I get the following error. Is this a bug or I'm doing something wrong?

`Input 'input' of 'MaxPool' Op has type float64 that does not match expected type of float32.`

Full trace:

``` python
Traceback (most recent call last):
  File "/Users/hamedketabdar/LearningTensorFlow/CIFAR-Khodam/convolutional_network_batch_2d2c_clean_64f.py", line 213, in <module>
    pred = conv_net(x, weights, biases, keep_prob)
  File "/Users/hamedketabdar/LearningTensorFlow/CIFAR-Khodam/convolutional_network_batch_2d2c_clean_64f.py", line 153, in conv_net
    conv1 = max_pool(conv1, k=2) # Normally K=2
  File "/Users/hamedketabdar/LearningTensorFlow/CIFAR-Khodam/convolutional_network_batch_2d2c_clean_64f.py", line 135, in max_pool
    return tf.nn.max_pool(img, ksize=[1, k, k, 1], strides=[1, k, k, 1], padding='SAME')
  File "/usr/local/lib/python2.7/site-packages/tensorflow/python/ops/nn_ops.py", line 235, in max_pool
    name=name)
  File "/usr/local/lib/python2.7/site-packages/tensorflow/python/ops/gen_nn_ops.py", line 449, in _max_pool
    strides=strides, padding=padding, name=name)
  File "/usr/local/lib/python2.7/site-packages/tensorflow/python/ops/op_def_library.py", line 430, in apply_op
    (prefix, dtypes.as_dtype(input_arg.type).name))
TypeError: Input 'input' of 'MaxPool' Op has type float64 that does not match expected type of float32.
```

stackoverflow link: http://stackoverflow.com/questions/34354126/tensorflow-maxpool-doesnt-accept-float64

mrry:
[...]

PR #6299, opened 2016-12-14
Second order derivaties for Pooling operations

AnishShah:
Fixes #6143
Fixes #6294 

This is still not a complete. I wanted to know whether this is the correct way? 
I ran the following code and it worked without any error.
If it is correct, then I will add the same for `avg_pool` too.
```
In [1]: import tensorflow as tf
I tensorflow/stream_executor/dso_loader.cc:128] successfully opened CUDA library libcublas.so.7.5 locally
I tensorflow/stream_executor/dso_loader.cc:128] successfully opened CUDA library libcudnn.so.4 locally
I tensorflow/stream_executor/dso_loader.cc:128] successfully opened CUDA library libcufft.so.7.5 locally
I tensorflow/stream_executor/dso_loader.cc:128] successfully opened CUDA library libcuda.so.1 locally
I tensorflow/stream_executor/dso_loader.cc:128] successfully opened CUDA library libcurand.so.7.5 locally

In [2]: x = tf.placeholder(tf.float32, shape=[1, 4, 4, 1])

In [3]: y = tf.nn.max_pool(x, ksize=[1, 2, 2, 1], strides=[1, 2, 2, 1], padding='VALID')

In [4]: dy = tf.gradients(y, x)

In [5]: ddy = tf.gradients(dy, x)
```

PR #6664, opened 2017-01-05
Max pool grad grad

aam-at:
Adds gpu and cpu kernels for max pool grad. Supports NHWC and NCHW layouts for gpu kernel and only NHWC for cpu kernel. Additionally registers gradient for python `MaxPoolGradGrad` which is just `MaxPoolGrad`. Registers gradient for python `AvgPoolGrad` which is `AvgPool` evaluated on `top_diff`.

Continuation of #6299


aam-at:
I signed it!

aam-at:
I did some refactoring for `MaxPoolingOp` which now uses functors and templates and registers kernels using `TF_CALL_*`. Since currently there more kernels for gpu then cpu, functors are not generalized to devices. Also adding `double` required a minor tweak to `stream_executor` (merely duplicating methods for `double`). I also added python tests with `double` now. This should fix #547

I also updated registered ops to use more general type - `realnumbertype`. Note, since I am not using default type for `T` ops compatibility test fails. Therefore, I have not updated yet `ops.pbtxt`.

I also would like to add `MaxPool3dGradGrad` (currently I added only forward declaration for it).

As this pull request is getting quite big, I would appreciate some comments.

aam-at:
I updated 3d kernels. Cpu kernel supports only NHWC and gpu kernel supports both. Currently, python tests for 3d pooling check only NHWC because gpu kernels for pooling and pooling grad only work with this format.

Someone can correct me if I am wrong, but cudnn 5 supports both formats of tensors (NHWC and NCHW). However, both pooling 2d and 3d kernels with cudnn backend always convert tensors from NHWC to NCHW format.

vrv:
This is really nice @aam-at: we're waiting for Benoit to do an initial review of this code (he's been on vacation the past week or so).

tensorflow/core/ops/nn_ops.cc
```diff
@@ -1102,45 +1112,84 @@ REGISTER_OP("MaxPool3D")
...
+    .Input("orig_input: T")
+    .Input("orig_output: T")
+    .Input("grad: T")
+    .Output("output: T")
+    .Attr("ksize: list(int) >= 5 ")
+    .Attr("strides: list(int) >= 5")
+    .Attr(GetPaddingAttrString())
+    .Attr(GetConvnetDataFormatAttrString())
+    .Attr("T: realnumbertype")
+    .SetShapeFn(shape_inference::Pool3DShape)
```
vrv: small note: this is doing validation that 'orig_input' and 'output' follow Pool3DShape's shape contract, but doesn't validate orig_output and 'grad'

I would recommend augmenting this registration with something like:

```
  .SetShapeFn([](InferenceContext* c) {
    TF_RETURN_IF_ERROR(shape_inference::Pool3DShape(c));
    ShapeHandle unused;
    // Validate 'grad' is the same shape as 'orig_input'
    TF_RETURN_IF_ERROR(c->Merge(c->input(0), c->input(2), &unused));
    // Validate 'orig_output' is same shape as 'output'
    TF_RETURN_IF_ERROR(c->Merge(c->input(1), c->output(0), &unused));
    return Status::OK();
  })
```

or something like that

tensorflow/core/ops/ops.pbtxt
```diff
@@ -10313,6 +10313,154 @@ op {
   summary: "Computes gradients of the maxpooling function."
 }
 op {
+  name: "MaxPoolGradGrad"
```
vrv: Thanks for making these changes, though you don't need to add them in this PR -- they are automatically generated later from the op registrations.

tensorflow/core/ops/nn_ops.cc
aam-at: Thanks! I will add it in the next commit.

tensorflow/core/ops/ops.pbtxt
aam-at: Okay. I may drop these changes then. Also, as I noted before, the previous declaration of MaxPool3dGrad didn't use template attribute `T` for types of input and output. So ops compatibility tests will fail. I assume that the correct declaration is to use template attribute.

vrv: Oh, you probably need to read this: https://github.com/tensorflow/tensorflow/blob/754048a0453a04a761e112ae5d99c149eb9910dd/tensorflow/g3doc/how_tos/adding_an_op/index.md#backwards-compatibility

My suggestion would be to declare this as a new type Tinput with a default of DT_FLOAT to preserve backwards compatibility as in the suggestion in the link above.  It sucks to have to declare a new type for this, but I can't think of alternative that is backwards compatible.

aam-at:
I dropped commit changing ops.pbtxt which required rebase.

tensorflow/core/ops/ops.pbtxt
aam-at: Yeah, that worked. I assume you meant something like this:
```c++
REGISTER_OP("MaxPool3DGrad")
    .Input("orig_input: Tinput")
    .Input("orig_output: Tinput")
    .Input("grad: T")
    .Output("output: T")
    .Attr("T: numbertype = DT_FLOAT")
    .Attr("Tinput: numbertype = DT_FLOAT");
```
Compatibility test passes now but I assume that original definition contained mistake. Do we want to preserve backward compatibility for this op?

aam-at:
Any update on reviewing this?

ahundt:
@vrv @benoitsteiner max pool grad grad definitely looks like some nice changes, would appreciate the review + merge!

vrv:
I'll kick off a test anyway, but will wait for @benoitsteiner to review.

aam-at:
The build was broken due to the changes in the upstream (191658d54f90ac03c15b339326129cd52d1f56a3). It is a bit confusing that testing is performed after merging with upstream, though, quite reasonable.

I updated pooling with doubles to call new cudnn loading mechanism. So build should be successful now. @vrv Can you rerun tests again?

vrv:

(I don't know the status of our builds right now, so just trying anyway since it's the weekend and the load is low)

vrv:
Looks like the sanity check is failing:

=== Sanity check step 3 of 6: do_buildifier (buildifier check) ===

Running do_buildifier on 157 files

tensorflow/core/kernels/BUILD # listsort unsafesort sort:tf_kernel_library.hdrs

buildifier took 0 s

2648d2647
<         "pooling_ops_common.h",
2649a2649
>         "pooling_ops_common.h",

aam-at:
Buildifier requires sorted list of headers and sources. So I fixed that. It's my
first pull request to tensorflow, so I am a bit confused with the requirements
for various tests. Thanks, @vrv!

vrv:

@gunan one thing that would be cool would be to have the sanity checks run automatically when users push their PRs -- but then we'd control the rest of the testing.

vrv:
@aam-at this is a pretty epic first PR to TensorFlow, so thank you for your patience :)

aam-at:
Tests are passing, except for macos. I don't have access to macos and I am not
sure why the build is failing. Thank you for your patience too, @vrv!

vrv:
I think it's a infrastructure failure, the fact that the other tests pass suggests this is ready for review!

gunan:
I think this can be an issue in build files. But bazel 0.4.3 has a bug in exception propagation, which make us see unhelpful messages. I will upgrade what I can access and rerun tests.

gunan:
Jenkins, test this please

aam-at:
All tests are passing now. Thanks for the help! I will wait for review now.

vrv:
ping for @benoitsteiner for this week.

ahundt:
@vrv @benoitsteiner Bumping this item since last post was 10 days ago. :-)

aam-at:
@vrv @benoitsteiner any update on this?

vrv:
@andydavis1 or @rmlarsen any takers? :)   I am swamped with a lot of other things :(

jhseu:
@rmlarsen, do you have time to look at this change?

rmlarsen:
I'll take a look a this sometime next week.

tensorflow/core/kernels/maxpooling_op.cc
```diff
@@ -403,12 +364,260 @@ class MaxPoolingGradOp<Eigen::GpuDevice, T> : public OpKernel {
...
+                errors::InvalidArgument("tensor_out must be 4-dimensional"));
+    // For maxpooling, out_grad_backprop should have 4 dimensions.
+    OP_REQUIRES(
+        context, out_grad_backprop.dims() == 4,
+        errors::InvalidArgument("out_grad_backprop must be 4-dimensional"));
+
+    TensorShape input_shape = tensor_in.shape();
+    TensorShape output_shape = tensor_out.shape();
+    Tensor* output = nullptr;
+    OP_REQUIRES_OK(context, context->allocate_output(0, output_shape, &output));
```
rmlarsen: Try:

OP_REQUIRES_OK(context, context->forward_input_or_allocate_output(
{2}, 0, tensor_out.shape(), &output));

tensorflow/core/kernels/maxpooling_op.cc
```diff
@@ -403,12 +364,260 @@ class MaxPoolingGradOp<Eigen::GpuDevice, T> : public OpKernel {
...
+    if (!context->status().ok()) {
```
rmlarsen: you don't need this. The OP_REQUIRES_OK macro tests the status and returns if not OK.

tensorflow/core/kernels/maxpooling_op.cc
```diff
@@ -403,12 +364,260 @@ class MaxPoolingGradOp<Eigen::GpuDevice, T> : public OpKernel {
...
+    OP_REQUIRES(
+        context, out_grad_backprop.dims() == 4,
+        errors::InvalidArgument("out_grad_backprop must be 4-dimensional"));
+
+    PoolParameters params{context,  ksize_,      stride_,
+                          padding_, FORMAT_NHWC, tensor_in.shape()};
+    Tensor* output = nullptr;
+    OP_REQUIRES_OK(context,
+                   context->allocate_output(0, tensor_out.shape(), &output));
+    if (!context->status().ok()) {
```
rmlarsen: you don't need this. The OP_REQUIRES_OK macro tests the status and returns if not OK.

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
```diff
@@ -199,32 +199,111 @@ __global__ void MaxPoolBackward(const int nthreads, const dtype* top_diff,
...
 
-  MaxPoolForwardNHWC<<<(output_size + kThreadsPerBlock - 1) / kThreadsPerBlock,
-                       kThreadsPerBlock, 0, d.stream()>>>(
-      output_size, bottom_data, height, width, channels, pooled_height,
-      pooled_width, kernel_h, kernel_w, stride_h, stride_w, pad_t, pad_l,
-      top_data, mask);
-  return d.ok();
+template <typename dtype>
+__global__ void MaxPoolGradBackward(const int nthreads, const dtype* top_diff,
+                                    const int64* mask, const int top_offset,
```
rmlarsen: "mask" seems a little misleading. Isn't it really argmax within the window?

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
```diff
@@ -199,32 +199,111 @@ __global__ void MaxPoolBackward(const int nthreads, const dtype* top_diff,
...
+template <typename dtype>
+__global__ void MaxPoolGradBackwardNoMaskNHWC(
+    const int nthreads, const dtype* bottom_data, const dtype* output_data,
+    const int pooled_height, const int pooled_width, const int channels,
+    const int height, const int width, const int kernel_h, const int kernel_w,
     const int stride_h, const int stride_w, const int pad_t, const int pad_l,
-    float* top_data, int64* mask, const Eigen::GpuDevice& d) {
-  const int kThreadsPerBlock = 1024;
-  const int output_size = batch * channels * pooled_height * pooled_width;
+    const dtype* top_diff, dtype* bottom_diff) {
```
rmlarsen: I realize that you didn't start this, but could you please add a comment clarifying what "top" and "bottom" refer to? (I agree that it is preferably to use those terms compared to further overloading "input" and "output".)

Same everywhere.

rmlarsen (changes requested):
First round of comments.

tensorflow/core/ops/nn_ops.cc
```diff
@@ -1102,45 +1112,93 @@ REGISTER_OP("MaxPool3D")
...
+    Alternatively, the format could be "NCHW", the data storage order of:
+        [batch, in_channels, in_planes, in_height, in_width].
 input: Shape `[batch, depth, rows, cols, channels]` tensor to pool over.
 output: The max pooled output tensor.
 )doc");
 
 REGISTER_OP("MaxPool3DGrad")
-    .Input("orig_input: float")
-    .Input("orig_output: float")
+    .Input("orig_input: TInput")
```
rmlarsen: Why allow separate types for input and output? The only existing registrations set them both to float:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/pooling_ops_3d.cc#L287
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/pooling_ops_3d.cc#L490

tensorflow/core/ops/nn_ops.cc
```diff
@@ -1340,6 +1435,39 @@ argmax: The indices of the maximum values chosen for each output of `max_pool`.
...
+      TF_RETURN_IF_ERROR(shape_inference::MaxPoolShape(c));
+      ShapeHandle unused;
+      // Validate 'orig_input' is the same shape as 'grad'
+      TF_RETURN_IF_ERROR(c->Merge(c->input(0), c->input(1), &unused));
+      // Validate 'argmax' is same shape as 'output'
+      TF_RETURN_IF_ERROR(c->Merge(c->input(2), c->output(0), &unused));
+      return Status::OK();
+    })
+    .Doc(R"doc(
+Computes gradients of the maxpooling backward function.
```
rmlarsen: Update comment. Should be "gradients of gradients" or  "second-order" gradients".

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -597,6 +598,38 @@ def _CompareMaxPoolingBk(self, input_shape, output_shape, ksize, strides,
...
+      with self.test_session(use_gpu=False):
+        t = constant_op.constant(tensor_input, shape=input_shape)
+        out_op = nn_ops.max_pool(t, ksize, strides, padding)
+        orig_out = out_op.eval()
+        grad_in = constant_op.constant(tensor_input, shape=input_shape)
+        out_op = gen_nn_ops._max_pool_grad_grad(t, orig_out, grad_in, ksize, strides,
+                                                padding)
+        cpu_val = out_op.eval()
+        self.assertShapeEqual(cpu_val, out_op)
+      if dtype == np.float16:
```
rmlarsen: You can simplify this by using assertAllCloseAccordingToType().

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -715,6 +771,83 @@ def _ConstructAndTestGradient(self,
...
+    for s in input_sizes:
+      total_size *= s
+    # Initializes the input tensor with array containing incrementing
+    # numbers from 1.
+    x = [f * 1.0 for f in range(1, total_size + 1)]
+    with self.test_session(use_gpu=use_gpu):
+      input_tensor = constant_op.constant(x, shape=input_sizes, name="input")
+      if pool_func == nn_ops.avg_pool:
+        func_name = "avg_pool"
+        err_margin = 1e-3
```
rmlarsen: s/margin/tolerance/

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -715,6 +771,83 @@ def _ConstructAndTestGradient(self,
...
+                                      window_cols,
+                                      row_stride,
+                                      col_stride,
+                                      padding,
+                                      data_format,
+                                      use_gpu,
+                                      x_init_value=None):
+    """Verifies the gradients of gradients of the pooling function.
+
+    Args:
```
rmlarsen: cut and paste error?

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -1269,6 +1538,17 @@ def Test(self):
...
+    if not test.is_gpu_available(cuda_only=True):
+      return
+    self._CompareMaxPoolingGradBk(input_size, output_size, filter_size, strides,
+                                  padding)
+
+  return Test
+
 
 if __name__ == "__main__":
   for (name_, input_size_, filter_size_, output_size_, stride_,
```
rmlarsen: This is broken. The test names are not unique and therefore only one of the possible tests will get run. This should set a test name that is a concatenation of all the test params. I also recommend using a function like 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/matmul_op_test.py#L35

to add the tests so this will not happen in the future.

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -3069,6 +3069,41 @@ bool CudnnSupport::DoActivate(Stream* stream,
...
+  mutex_lock lock{dnn_handle_mutex_};
+  auto status = wrap::cudnnSetStream(parent_, ToHandle(dnn_handle_),
+                                        AsCUDAStreamValue(stream));
+  if (status != CUDNN_STATUS_SUCCESS) {
+    LOG(ERROR) << "failed to set stream for cudnn handle: " << ToString(status);
+    return false;
+  }
+
+  // Alpha is the scaling factor for input.
+  float alpha = 1.0;
```
rmlarsen: double?

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -3138,6 +3173,44 @@ bool CudnnSupport::DoPoolForward(
...
```
rmlarsen: double?

rmlarsen (changes requested):

tensorflow/stream_executor/cuda/cuda_dnn.cc
rmlarsen: The cuDNN documentation states that:

"storage data type for alpha[0], beta[0] is float for HALF and SINGLE tensors, and double
for DOUBLE tensors. These parameters are passed using a host memory pointer"

tensorflow/core/kernels/maxpooling_op.cc
```diff
@@ -403,12 +364,260 @@ class MaxPoolingGradOp<Eigen::GpuDevice, T> : public OpKernel {
...
+                errors::InvalidArgument("tensor_out must be 4-dimensional"));
+    // For maxpooling, out_grad_backprop should have 4 dimensions.
+    OP_REQUIRES(
+        context, out_grad_backprop.dims() == 4,
+        errors::InvalidArgument("out_grad_backprop must be 4-dimensional"));
+
+    PoolParameters params{context,  ksize_,      stride_,
+                          padding_, FORMAT_NHWC, tensor_in.shape()};
+    Tensor* output = nullptr;
+    OP_REQUIRES_OK(context,
```
aam-at: It seems that it was added only recently in the upstream. I will do rebase for that.

tensorflow/core/kernels/maxpooling_op.cc
aam-at: Okay. I suspected that but previous implementation of GradOp had an explicit checks. I will remove unnecessary checks throughout `maxpooling` file then.

tensorflow/core/ops/nn_ops.cc
aam-at: It is related to compatibility tests. Previous registration uses `orig_input: float` and `grad: T`. If I try to generalize and set `orig_input: T`, the compatibility test will fail. It was previously discussed here - https://github.com/tensorflow/tensorflow/pull/6664#discussion_r96332231 and I followed @vrv suggestion.

tensorflow/core/ops/nn_ops.cc
aam-at: I updated doc to use your wording "second-order gradients". However, note that C++ code does not compute "final" gradients. GradGradOp will return gradient w.r.t. max pooling gradient. Second-order gradient will be correctly handled in backend gradient implementation, such as python.

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
aam-at: It has the same size as the output `N*C*Hout*Wout` and contains the corresponding index in the input for each element in the output.

tensorflow/python/kernel_tests/pooling_ops_test.py
aam-at: `GetShrunkInceptionMaxPoolShapes` returns unique name for each parameters test combination, so it should be okay.

martinwicke:
@aam-at what's the status of this?

rmlarsen:
@jlebar would you please review the stream executor changes in this PR?

aam-at:
@martinwicke It implements MaxPoolingGradGrad kernels for CPU and GPU. Tests are passing on my machine.

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
rmlarsen: Exactly, so I would not name it mask, but "indices" or "argmax" or something like that.

rmlarsen:
@aam-at Thanks for the updates. Could you please rebase to resolve the conflicts and address the remaining comments.

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
aam-at: Some kernels were borrowed from caffe, so we should rename `mask` there too. In addition, how should we rename functions like `MaxPoolBackwardNoMask`? In my opinion, the naming may be not the most intuitive (like `max_to_indices`) but it does not introduce too much confusion.

aam-at:
@rmlarsen I have finished rebase. One commit was dropped, since `TensorFormat` handles now 5d tensors. Regarding `mask` parameter, I am not sure if we should rename this parameter because it can introduce further confusion:
1) some kernels were borrowed from caffe
2) not clear how to rename `NoMask` methods.

martinwicke:
There are still (or again) conflicts. @rmlarsen if you can resolve the naming issues, we can move this PR along.

rmlarsen:
@aam-at OK, let's stick with mask then.

martinwicke:
(if this is the last sticking point, make sure to update your review, and I'll switch to just caring about the tests)

rmlarsen (approved):

rmlarsen:
@aam-at please resolve the conflict so we can test this more thoroughly.

rmlarsen (changes requested):
I would like to run our internal testing before approving this. This requires a clean rebase.

aam-at:
@rmlarsen I did rebase again. Currently, there are no conflicts, so you can run the tests.

rmlarsen:
@aam-at Thanks!

rmlarsen:

rmlarsen:
@aam-at it looks like there are some broken tests (backwards compatibility and XLA pooling tests) that need attention.

tensorflow/core/ops/nn_ops.cc
```diff
@@ -1303,7 +1341,7 @@ output: The gradients for LRN.
 // --------------------------------------------------------------------------
 
 REGISTER_OP("MaxPool")
-    .Attr("T: {float, half} = DT_FLOAT")
+    .Attr("T: realnumbertype")
```
rmlarsen: Do you need to keep the default value here?

tensorflow/core/ops/nn_ops.cc
```diff
@@ -1336,7 +1374,7 @@ REGISTER_OP("MaxPoolGrad")
     .Input("orig_output: T")
     .Input("grad: T")
     .Output("output: T")
-    .Attr("T: {float, half} = DT_FLOAT")
+    .Attr("T: realnumbertype")
```
rmlarsen: ditto?

tensorflow/core/ops/nn_ops.cc
aam-at: The problem was with `MaxPool3dGrad`. We discussed why it is necessary to have two types for input and output, so compatibility tests can pass. I guess during rebase I dropped those changes by accident.

tensorflow/core/ops/nn_ops.cc
aam-at: Fixed.

aam-at:
@rmlarsen I fixed backward compatibility tests. Xla pooling ops test should be also fixed.

vrv:

rmlarsen (approved):

rmlarsen:
@aam-at Excellent. Thank you for this large contribution!

aam-at:
@vrv @rmlarsen Thank you for the help!

rmlarsen:
@aam-at it appears that the two tests:

_testMaxPoolGradGradSamePadding2_1(data_format, use_gpu)

and 

_testMaxPoolGradGradSamePadding3_1(data_format, use_gpu)

are flaky when use_gpu is True, e.g.:

  File "/usr/grte/v4/k8-linux/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/build/work/0bfa20861d2fc543528291b5513e5ecb/google3/runfiles/google3/third_party/tensorflow/python/kernel_tests/pooling_ops_test.py", line 1293, in testMaxPoolGradGrad
    self._testMaxPoolGradGradSamePadding2_1(data_format, use_gpu)
  File "/build/work/0bfa20861d2fc543528291b5513e5ecb/google3/runfiles/google3/third_party/tensorflow/python/kernel_tests/pooling_ops_test.py", line 1258, in _testMaxPoolGradGradSamePadding2_1
    use_gpu=use_gpu)
  File "/build/work/0bfa20861d2fc543528291b5513e5ecb/google3/runfiles/google3/third_party/tensorflow/python/kernel_tests/pooling_ops_test.py", line 813, in _ConstructAndTestSecondGradient
    self.assertLess(err, err_tolerance)
  File "/usr/grte/v4/k8-linux/lib/python2.7/unittest/case.py", line 932, in assertLess
    self.fail(self._formatMessage(msg, standardMsg))
  File "/build/work/0bfa20861d2fc543528291b5513e5ecb/google3/runfiles/google3/testing/pybase/basetest.py", line 1205, in fail
    return super(TestCase, self).fail(self._formatMessage(prefix, msg))
  File "/usr/grte/v4/k8-linux/lib/python2.7/unittest/case.py", line 412, in fail
    raise self.failureException(msg)
AssertionError: 1.99932861328125 not less than 0.01

Perhaps there's a bug for padding is SAME? Do you have any idea what the underlying cause might be?

rmlarsen:
The problem was caused by a race due to using buffer forwarding instead of allocation in MaxPoolingGradGradOp(). I'll submit a fix.

aam-at:
@rmlarsen It was night in my timezone. Thanks for the fix. I added `forward_input_or_allocated_output` in c1b6aa4b6bbb2e80e228b3ecb9f2877f4e5dc26a. Shouldn't we replace buffer forwarding to allocating output for 3d pooling grad grad too? I am actually not sure why `forward_input_or_allocated_output` creates race conditions.

PR #6665, opened 2017-01-05
Changes in Interactive Usage Section 

shivam-kotwalia:
Changed x from Variable to constant and commented the x.initializer.run()


shivam-kotwalia:
I signed it!

caisq:
What is the rationale for this change?

gunan:
I think having Variable there shows how to create variables. So it adds value.
I also verified that it works on 0.12 as is right now.

So I will close this PR. If you really think we should make this change, please reopen or create a new PR with more explanation.

PR #6684, opened 2017-01-06
TensorFlow API compatibility test.

gunan:


tensorflow/tools/api/golden/tensorflow.AggregationMethod.pbtxt
```diff
@@ -0,0 +1,22 @@
+path: "tensorflow.AggregationMethod"
+tf_class {
+  name: "AggregationMethod"
+  is_instance: "<class \'tensorflow.python.ops.gradients_impl.AggregationMethod\'>"
```
martinwicke: Is this class name relevant? The important thing is that it `isinstance` `tensorflow.AggregationMethod`, not so much that it's in gradients_impl...

martinwicke: This seems to only affect the class itself, its base classes appear fine?

tensorflow/tools/api/golden/tensorflow.layers.pbtxt
```diff
@@ -0,0 +1,517 @@
+path: "tensorflow.layers"
+tf_module {
+  name: "tensorflow.python.layers.layers"
```
martinwicke: So the test will fail if we move a file and keep it the same name in python. Is this intentional? It's definitely the conservative choice, so it's probably for the best.

tensorflow/tools/api/golden/tensorflow.layers.pbtxt
```diff
@@ -0,0 +1,517 @@
...
+      arg_name: "padding"
+    }
+    args {
+      arg_name: "data_format"
+    }
+    args {
+      arg_name: "name"
+    }
+  }
+  member_methods {
```
martinwicke: Should it be `member_method`?

tensorflow/tools/api/golden/tensorflow.layers.pbtxt
```diff
@@ -0,0 +1,517 @@
...
+      arg_name: "inputs"
+    }
+    args {
+      arg_name: "pool_size"
+    }
+    args {
+      arg_name: "strides"
+    }
+    args {
+      arg_name: "padding"
```
martinwicke: Are you not recording the value of defaults?

tensorflow/tools/api/golden/tensorflow.pbtxt
```diff
@@ -0,0 +1,5147 @@
+path: "tensorflow"
+tf_module {
+  name: "tensorflow"
+  members {
+    name: "AggregationMethod"
+    mtype: "<type \'type\'>"
+  }
+  members {
```
martinwicke: Similar to elsehwere: should it be `member`?

tensorflow/tools/common/public_api.py
```diff
@@ -0,0 +1,69 @@
+"""TODO(wicke): DO NOT SUBMIT without one-line documentation for public_api.
```
martinwicke: Mine is now submitted, you should be able to use it.

martinwicke (changes requested):

PR #6686, opened 2017-01-06
Adding support for Linux s390x in sparse_split_op

namrata-ibm:
Changing the data type of split_dim to resolve sparse_split_op_test failure on Big Endian System.

### DIFF
--- a/tensorflow/core/kernels/sparse_split_op.cc
+++ b/tensorflow/core/kernels/sparse_split_op.cc
@@ -30,7 +30,7 @@ class SparseSplitOp : public OpKernel {
   }
 
   void Compute(OpKernelContext* context) override {
-    const int32 split_dim = context->input(0).scalar<int>()();
+    const int64 split_dim = context->input(0).scalar<int64>()();
     const Tensor& input_indices = context->input(1);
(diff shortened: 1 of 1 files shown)


namrata-ibm:
The split_dim in sparse_split_op.cc is int32 type which causes the column wise tests(where split_dim should be 1) to fail on Big Endian system as the "context->input(0).scalar<int>()();" returns 0 all the time.

Could you review the change please?

rohan100jain (approved):

rohan100jain:
Jenkins, test this please

gunan:
Jenkins, test this please.

drpngx:
Could we add a test for this?

namrata-ibm:
The int64 data type change in this Op file is related to the `//tensorflow/python/kernel_tests:sparse_split_op_test`.  With above change test case is passing on big endian as well as little endian machine. @drpngx , Are you referring to any other test here?

drpngx:
I guess I'm looking for something that would test that specifically, on little endian, if possible.

drpngx:
Hmm, thinking about it, it's hard to create a test for this.

PR #6699, opened 2017-01-06
Branch 143784180

rohan100jain:


gunan:
The failure in PR tests is the same as the failure in internal continuous tests.
We should fix that failure first and create a new push.

gunan:
Closing in favor of #6719

PR #6707, opened 2017-01-07
Fix: Check bazel version before trying to setup the workspace.

gunan:
I could not prevent all the other errors from showing up, but this way we at least get the bazel version error message at the end:

```
...
ERROR: /tensorflow/tensorflow/workspace.bzl:351:3: //external:zlib_archive: no such attribute 'urls' in 'new_http_archive' rule.
ERROR: /tensorflow/tensorflow/workspace.bzl:351:3: //external:zlib_archive: missing value for mandatory attribute 'url' in 'new_http_archive' rule.
ERROR: /tensorflow/WORKSPACE:6:1: Traceback (most recent call last):
	File "/tensorflow/WORKSPACE", line 6
		check_version("0.4.2")
	File "/tensorflow/tensorflow/workspace.bzl", line 32, in check_version
		fail("
Current Bazel version is {}, e...))

Current Bazel version is 0.3.2, expected at least 0.4.2
.
ERROR: Error evaluating WORKSPACE file.
ERROR: error loading package 'external': Package 'external' contains errors.
ERROR: missing fetch expression. Type 'bazel help fetch' for syntax and help.
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,12 @@
 workspace(name = "org_tensorflow")
 
+load("//tensorflow:workspace.bzl", "check_version", "tf_workspace")
+
+# We must check the bazel version before trying to parse any other BUILD files,
+# in case the parsing of those build files depends on the bazel version we
+# require here.
+check_version("0.4.2")
+
 # Uncomment and update the paths in these entries to build the Android demo.
 #android_sdk_repository(
 #    name = "androidsdk",
@@ -15,12 +22,7 @@ workspace(name = "org_tensorflow")
 #    api_level=21)
 
 # Please add all new TensorFlow dependencies in workspace.bzl.
-load("//tensorflow:workspace.bzl", "tf_workspace")
 tf_workspace()
-
-# Specify the minimum required bazel version.
-load("//tensorflow:tensorflow.bzl", "check_version")
-check_version("0.4.2")
 
 new_http_archive(
   name = "inception5h",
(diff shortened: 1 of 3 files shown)


vrv:
Nice, do you know why the order matters?

gunan:
I think before we were blocked by tf_workspace() call below.
When loading the functions, I think they are non-fatal error messages, but just like what check_version does, tf_workspace did break up all evaluation. The errors stem from addition of mirrors to all the dependency rules we have.

WORKSPACE
```diff
@@ -1,5 +1,11 @@
 workspace(name = "org_tensorflow")
 
+load("//tensorflow:workspace.bzl", "check_version", "tf_workspace")
+
+# DO NOT ADD ANYTHING ABOVE THIS LINE!!!
```
vrv: Can we be a bit more careful about what is said here?  Maybe we can say "We must check the bazel version before trying to parse any other BUILD files, in case the parsing of those build files depends on the bazel version we require here."

gunan: Done. Made the change as a separate commit but I can squash all after approval, if you like.

vrv (approved):

martinwicke (approved):

gunan:
Jenkins, test this please.

alexeagle:
I was copying this for rules_nodejs and rules_typescript and noticed a bug:

`Current Bazel version is 1.12.3, expected at least 1.2.1`

/cc @mprobst

Mistobaan:
I also think that the check_version should be inside the `tf_workspace` function. Is too easy to update tensorflow without updating the check_version constraints.

alexeagle:
FYI, I added unit tests in my clone of this feature:
https://github.com/bazelbuild/rules_nodejs/pull/19/commits/f401ccb6fa79b4b46efe9cfd8cb6a5ae07d6aaa9

PR #6710, opened 2017-01-07
Update jpeg.BUILD

### LINKED ISSUE #6706: Bazel problem when using Tensorflow as an external dependency.
When using the HEAD version of Tensorflow as an external Bazel dependency (like tensorflow_serving does), we run into an issue in the line:

`load("@//third_party:common.bzl", "template_rule")`

inside

`tensorflow/third_party/jpeg/jpeg.BUILD`

That line assumes that Tensorflow is the main repository. But when Tensorflow is included in another project the main repository is the main project so that common.bzl file is not found. I think that line needs to read:

`load("@org_tensorflow//third_party:common.bzl", "template_rule")`

yaroslavvb:
Fixes #6706

### DIFF
--- a/third_party/jpeg/jpeg.BUILD
+++ b/third_party/jpeg/jpeg.BUILD
@@ -5,7 +5,7 @@ licenses(["notice"])  # custom notice-style license, see LICENSE.md
 
 exports_files(["LICENSE.md"])
 
-load("@//third_party:common.bzl", "template_rule")
+load("@org_tensorflow//third_party:common.bzl", "template_rule")
 
 libjpegturbo_nocopts = "-[W]error"
 


gunan:
Jenkins, test this please.

gunan (approved):

gunan:
It looks like the issue is caused by the change here:
```
ERROR: /workspace/tensorflow/core/platform/default/build_config/BUILD:108:1: error loading package '@jpeg//': Encountered error while reading extension file 'third_party/common.bzl': no such package '@org_tensorflow//third_party': error loading package 'external': The repository named 'org_tensorflow' could not be resolved and referenced by '//tensorflow/core/platform/default/build_config:jpeg'.
```
@jart @damienmg Could you take a look at this change?

gunan:
@jart should I remove the `load` statement, and simply redefine the routine in jpeg.BUILD and llvm.BUILD for now? I can add a TODO to fix this later, and we will unblock @yaroslavvb

jart:
Yes that's probably the wisest course of action right now, per discussion in related issue.

drpngx:
Uh oh, it seems to failing the build.

gunan:
I have contacted bazel team to see how we can work around the problem.
In the meantime, I think we should close this PR as we will not be able to accept it.

Issue #6558, opened 2016-12-29
`TensorContractionThreadPool.h' file not found` when using `--copt=-march=native`


zafartahirov:
When trying to run 

```
$ ./configure
........
$ bazel build --copt=-march=native -c opt //tensorflow/tools/pip_package:build_pip_package --verbose_failures
```

it fails with the `./third_party/eigen3/unsupported/Eigen/CXX11/FixedPoint:42:10: fatal error: 'src/Tensor/TensorContractionThreadPool.h' file not found` error.

Similar problem was seen in #4680, #580, 

I am on Mac OS X Sierra 10.12.2 with no CUDA and cuDNN installed. I am using virtual environment on Python 2.7

Here is the checked commit and `bazel` version
```
$ git rev-parse HEAD
ac28ae043df4bc6f112f964f6df22845b8a05390
$ bazel version
Build label: 0.4.3-homebrew
Build target: bazel-out/local-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 22 15:20:15 2016 (1482420015)
Build timestamp: 1482420015
Build timestamp as int: 1482420015
```

This works fine:
```
$ git checkout remotes/origin/r0.12
$ ./configure # With defaults
.........
$ bazel build --copt=-march=native -c opt //tensorflow/tools/pip_package:build_pip_package --verbose_failures
# This finishes correctly
```

Here is the full terminal log when `bazel` is run on the `master` branch:
```
$ git checkout master
$ ./configure # With defaults
.........
$ bazel build --copt=-march=native -c opt //tensorflow/tools/pip_package:build_pip_package --verbose_failures
INFO: Found 1 target...
ERROR: /Users/zafar/GitHub/tensorflow/tensorflow/core/kernels/BUILD:856:1: C++ compilation of rule '//tensorflow/core/kernels:gather_functor' failed: cc_wrapper.sh failed: error executing command
  (cd /private/var/tmp/_bazel_zafar/c40dcca6773e4515fd11c724e0ebae51/execroot/tensorflow && \
  exec env - \
[...]

PR #6723, opened 2017-01-08
Fix build errors on AVX2+ hosts with -march=native.

pwnall:
third_party/eigen3/unsupported/Eigen/CXX/FixedPoint and third_party/eigen3/unsupported/Eigen/CXX/Tensor use different paths to reach into unsupported/Eigen/CXX/src/Tensor. The former's paths cause the build errors documented in #6558. This commit converges on the path in third_party/eigen3/unsupported/Eigen/CXX/Tensor, which fixes the build error.

I tested the PR with the following command, which fails without the patch.

```bash
bazel build -c opt --copt=-march=native //tensorflow:libtensorflow.so
```

### DIFF
--- a/third_party/eigen3/unsupported/Eigen/CXX11/FixedPoint
+++ b/third_party/eigen3/unsupported/Eigen/CXX11/FixedPoint
@@ -32,14 +32,22 @@
 
 // Use optimized implementations whenever available
 #ifdef EIGEN_VECTORIZE_AVX512
-#include "src/Tensor/TensorContractionThreadPool.h"
+
+// This include reaches into the original eigen3 files. The full relative path
+// is needed to avoid https://github.com/tensorflow/tensorflow/issues/6558
+#include "unsupported/Eigen/CXX11/src/Tensor/TensorContractionThreadPool.h"
+
 #include "src/FixedPoint/PacketMathAVX512.h"
 #include "src/FixedPoint/TypeCastingAVX512.h"
 
 #elif defined EIGEN_VECTORIZE_AVX2
 #define EIGEN_USE_OPTIMIZED_INT8_UINT8_MAT_MAT_PRODUCT
 #define EIGEN_USE_OPTIMIZED_INT16_INT16_MAT_MAT_PRODUCT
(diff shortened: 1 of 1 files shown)


drpngx:
Could you add a comment immediately preceding the change explaining the problem succintly?

pwnall:
@drpngx I hope I have addressed your feedback. PTAL?

I also submitted #6759, which may be a better fix. I don't have access to all the tensorflow buildbots though, so I'm not sure it doesn't break anything. Either that PR or this one should take care of #6558.

Thank you very much for your review!

drpngx:
Jenkins, test this please.

Please ignore the estimator_test error.

drpngx:
If the ThreadPool include is genuinely used in that file, then this is a better fix. 

It sounds like we should try to push upstream fixes to eigen as @fnl suggested, then undo this change once propagated.

pwnall:
@drpngx Are you referring to [this comment](https://github.com/tensorflow/tensorflow/issues/6558#issuecomment-269643996) or another comment from @fnl?

FWIW, I was able to get a full build using `bazel build -c opt --copt=-march=native //...` on an AVX2 machine, so I'm somewhat skeptical that the Tensor include is needed in the FixedNum file. I'm working on getting a Skylake build going (I'm on OSX, and Apple's clang crashes on `--copt=-march=skylake`).

drpngx:
Yes, it sounds like it's something with Eigen that should be fixed there, ultimately.

pwnall:
My understanding is that @fnl is saying that if TensorContractionThreadPool.h is copied from eigen3 into third_party/eigen3/ then the build succeeds. To me, this seems an indication that the path to reach it is incorrect.

Or, are you saying that we shouldn't have to ship a modified FixedPoint header at all?

drpngx:
OK, it looks like #6759 is the better fix. Could you put a comment in that file to explain that we can't add the include in this place?

kunthar:
@drpngx since you've triggered one of our remote sensors with the word "Jenkins" + "please", your message arrived to our space time. Please stand by in your coordinates, we will beam you up to our star system soon. Thank you.

PR #6751, opened 2017-01-10
Branch 144020788

drpngx:


drpngx:
Will re-open with py3 fix.

Issue #136, opened 2015-11-11
ResourceExhaustedError in CNN/MNIST example (with GPU)


ywatanabex:
(I'm using GPU(GTX 980) with CUDA-7.0&cuDNNv2, on Ubuntu 14.04)
I have gone through MNIST tutorial: http://tensorflow.org/tutorials/mnist/pros/index.md

Everything was going well except for the last two lines:

```python
print "test accuracy %g"%accuracy.eval(feed_dict={
    x: mnist.test.images, y_: mnist.test.labels, keep_prob: 1.0})
```

Executing these lines, I got an error: 

```python
ResourceExhaustedError: OOM when allocating tensor with shapedim { size: 10000 } dim { size: 18 } dim { size: 18 } dim { size: 32 }
```

I think basic reason for this error is that test data can not be allocated to GPU device.
Is this a bug or not? Are there good way to avoid this issue?

avitale:
I am having the same issue on OS X Yosemite with Docker container.
In my case it should be using the CPU, I tried bumping Docker memory limit to 8GB but still getting OOM  .
Looking at Docker Stats, during the final accuracy.eval() memory goes from 300MB to 900MB then it crashes.

vincentvanhoucke:
If you don't have enough memory on your GPU to fit the whole test data, you could feed it in small batches to the eval graph using feed_dict like the example does with the training data.

avitale:
I solved my case. Being on OS X, I use Docker via the docker-machine VM.
Both Docker and docker-machine VM have a default memory limit of 1GB, increasing both memory limits I was able to get the results without OOM errors. 
@ywatanabex ,  if it may help you, using Docker stats I saw memory peak around 1.9GB

EigenFace:
I have this same problem on this same example with Ubuntu 14.04 with Nvidia GTX 970. Hits 3.35 GB usage and then crashes.

VedAustin:
Same problem when running from a Dockers terminal on a Windows machine. I ran the example convolutional.py as suggested in their tutorials and for six hours it was still running chugging out the gradually decreasing error rates before crashing out!
Anything unusual about it taking 6 hours or is it normal since I am using CPU and not the recommended GPU for DNNs?

mtourne:
Submitted a PR to that effect : https://github.com/tensorflow/tensorflow/pull/157

Did just as @vincentvanhoucke suggested and transform the test set to use batches.

gyoshi:
[...]

PR #6764, opened 2017-01-10
MNIST batch test

rmekdma:
The test code uses test data as whole. It requires a lot of memory.
I modified it to test every 50 samples and average the results.

### DIFF
--- a/tensorflow/g3doc/tutorials/mnist/pros/index.md
+++ b/tensorflow/g3doc/tutorials/mnist/pros/index.md
@@ -400,6 +400,7 @@ cross_entropy = tf.reduce_mean(
 train_step = tf.train.AdamOptimizer(1e-4).minimize(cross_entropy)
 correct_prediction = tf.equal(tf.argmax(y_conv,1), tf.argmax(y_,1))
 accuracy = tf.reduce_mean(tf.cast(correct_prediction, tf.float32))
+
 sess.run(tf.global_variables_initializer())
 for i in range(20000):
   batch = mnist.train.next_batch(50)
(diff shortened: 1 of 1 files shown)


rmekdma:
I signed it!

drpngx:
The indentation is wrong? What do you mean to implement here?

tensorflow/g3doc/tutorials/mnist/pros/index.md
```diff
@@ -409,8 +410,14 @@ for i in range(20000):
...
-print("test accuracy %g"%accuracy.eval(feed_dict={
-    x: mnist.test.images, y_: mnist.test.labels, keep_prob: 1.0}))
+acc = 0 
+num_batch = mnist.test.num_examples/50
+for i in range(num_batch):
+  batch = mnist.test.next_batch(50)
+  acc += accuracy.eval(feed_dict={x: batch[0], y_: batch[1], keep_prob: 1.0})
+
+acc /= num_batch
+print("test accuracy %g"%acc)
```
kdbanman: @rmekdma you are splitting the mnist test set into 50 batches and testing accuracy of each batch individually.  Then you average the each of those batch accuracy values together into an overall "test accuracy".  But:

- If all batches are of identical size, then this will be exactly equivalent to the existing unbatched accuracy calculation, just slower.
- And if the batches are not all identically sized, the overall test accuracy value will be incorrect.

What is the improvement you are trying to make?

rmekdma: This code is for mnist tutorial (with 10000 test examples) and not for general process. Also, to process unbatched test data, one needs >4GB gpu memory. This modification is for relieving gpu memory usage.

kdbanman: Ahh memory efficiency.  I hadn't thought of that.  I'm curious how different it is, so I'll run some tests and compare.

drpngx:
We're trying to keep the documentation short and sweet. This makes it more complex. I think that most cards available today come with 4GB+ of memory, right?

rmekdma:
@drpngx Experts have expensive GPUs with large memory. I think tutorials are for most people not only for experts. But if you think this modification is unnecessary, please close this issue.

drpngx:
OK, closing for now. If we see some activity in stackoverflow or our github issues around that, we'll revive the thread. Thanks!

wolverine2k:
It would be great to have this patch either as a part of the existing samples or properly in the documentation. As @rmekdma writes, most of the people do not have access to >4GB of GPU memory. Batch processing allows for running the samples of the lower range of graphic cards. Also as I see it, a lot of people have been facing the OOM issue as can be seen by the references to #136. Thanks.

SamuelYG:
agree with “most cards available today come with 4GB+ of memory”, but i do not think all people have this "cards", especially on many laptop which is not newer.

PR #6772, opened 2017-01-10
Branch 144094196

drpngx:


gunan (approved):

drpngx:
Build timed out after 90min.

Jenkins, test this please.

gunan:
@hawkinsp added LLVM dependency is causing builds to time out.
We should change builds to avoid building XLA&LLVM.

drpngx:
How come we are building XLA? I thought configure used `TF_ENABLE_XLA=0`.

hawkinsp:
TF_ENABLE_XLA=1 is currently set in the builder configuration.

But simply disabling that isn't enough, we also need to blacklist
//tensorflow/compiler/... from the set of targets to test.

drpngx:
We could use `$(bazel query "tensorflow/... except tensorflow/compiler/...")` for `TF_BUILD_BAZEL_TARGET` `ci_parameterized_build.sh`... Is there a more elegant way?

drpngx:
move all tensorflow/compiler/ BUILD files, and load them conditionally?

@jart any idea?

gunan:
Working on a fix, will test and see how it works.

jart:
If LLVM can't be fixed, the bazel query proposal in https://github.com/tensorflow/tensorflow/pull/6772#issuecomment-271717205 sounds good. I'm just not positive that query will work. Maybe we need something like `tensorflow/... except rdeps(@llvm//..., tensorflow/...)`.

PR #6805, opened 2017-01-12
Fix a bug in windows_file_system.cc for reading file

meteorcloudy:
When read_result is TRUE, result should be set to bytes_read.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms686358(v=vs.85).aspx
"When a function is called to perform an overlapped operation, the operation might be completed before the function returns. When this happens, the results are handled as if the operation had been performed synchronously. If the operation was not completed, however, the function's return value is FALSE, and the GetLastError function returns ERROR_IO_PENDING."

### DIFF
--- a/tensorflow/core/platform/windows/windows_file_system.cc
+++ b/tensorflow/core/platform/windows/windows_file_system.cc
@@ -72,7 +72,9 @@ SSIZE_T pread(HANDLE hfile, char* src, size_t num_bytes, uint64_t offset) {
 
   BOOL read_result = ::ReadFile(hfile, src, static_cast<DWORD>(num_bytes),
                                 &bytes_read, &overlapped);
-  if ((FALSE == read_result) &&
+  if (TRUE == read_result) {
+    result = bytes_read;
+  } else if ((FALSE == read_result) &&
       ((last_error = GetLastError()) != ERROR_IO_PENDING)) {
     result = (last_error == ERROR_HANDLE_EOF) ? 0 : -1;
   } else {


meteorcloudy:
I found this bug when I am setting a TF job for windows on Bazel ci where `python_op_gen_main` failed to read `hidden_ops.txt`.

drpngx:
Is there another function that won't have a race condition?

meteorcloudy:
I am not sure. @mrry Do we use `::ReadFile` for a reason? The problem is the code didn't deal with the situation when the function is performed as if it's synchronously. Is there a better solution?

mrry:
@meteorcloudy I believe @vit-stepanovs contributed that particular piece of code, and I assume he used overlapped I/O here because it's the easiest way to read from a particular offset **without** a race condition if multiple threads are accessing the file. (@drpngx: I'm not sure what race condition you mean here. The return value is non-deterministic, sure, but as long as we handle it as Yun does here, I think it's threadsafe.)

The fix looks correct to me... I've had to fix similar issues with Windows API calls that way in the past, and I can't think of a more elegant solution.

mrry (approved):
Thanks for fixing this!

drpngx:
I was thinking of GetLastError. Not a huge problem, but if other code thrashes errno then it could become a problem.

mrry:
Ah, I see what you mean. That does sound ominous, but according to the docs, `GetLastError()` returns errno for the calling thread, so we should be alright:

> Retrieves the calling thread's last-error code value. The last-error code is maintained on a per-thread basis. Multiple threads do not overwrite each other's last-error code.

(How nice of the doc writer to state the property in triplicate!)

https://msdn.microsoft.com/en-us/library/windows/desktop/ms679360(v=vs.85).aspx

drpngx:
Oh nice, good to know. Thanks!

Issue #6817, opened 2017-01-13
Add CUDA implementation of fused hue adjustment


mkolod:
Currently, the fused (rgb2hsv->hue_adjust->hsv2rgb) hue adjustment kernel exists in TF [for the CPU](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/adjust_hue_op.cc), but not for the GPU. This is also why the CPU path can only be enabled "secretly" (via [setting the TF_ADJUST_HUE_FUSED env var](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/image_ops_impl.py#L1113) to true). Adding the CUDA kernel would make this op available for both CPUs and GPUs.

mkolod:
Mentioning @zheng-xq since he implemented the CPU version. I will be submitting a PR for this issue soon so feel free to assign me.

mkolod:
BTW, by the [following test](XXX), run on a machine with 32 GB RAM, a 6-core Intel Core i7-5930K CPU @ 3.50GHz and an NVIDIA GTX 1080 GPU produced a GPU-based speedup factor of 17.3, given both the use of a fused CPU and GPU kernel. This was based on a test with a 4D random image tensor with a batch size of 64 and "image" size of 256x256x3. The GTX 1080 throughput 31,950 images per second, while the Core i7-5930K throughput was 1,845 images per second. Longer running times, e.g. running tf.while_loop 10k times, can bring the speedup factor to about 30x on this reference hardware. Both 3D and 4D tensors work, but of course 4D tensors perform better due to batching. Note that for uint8 inputs (e.g. from tf.image_jpeg_decode), the results of the CPU and GPU kernel are identical, pixel for pixel. For inputs of int32, with a numeric range of [0, 256], a small fraction of pixels differ by up to 0.3% in intensity (1/255), which is caused by the fact that the Python wrapper for adjust_hue() does scaling, and scaling a range of [0, 255] by 1E32 causes precision problems. That's a very small error for that unusual case. Numbers in range of [0, 255] represented as floats, or as uint8 (scaled by 255 for float conversion by convert_image_dtype() inside tf.image_adjust_hue()) return exactly the same results on the CPU and GPU. Note that this is just an implementation of adjust_hue, and doesn't necessarily work for tf.image.random_hue, because the hue delta taken by the CUDA kernel is a scalar rather than a whole tensor. This can be addressed it need be to cover both adjust_hue() and random_hue().

[...]

PR #6818, opened 2017-01-13
GPU impl. of hue adjustment op

mkolod:
This is related to issue #6817. Mentioning @zheng-xq since he implemented the fused CPU hue adjustment kernel.

### DIFF
--- a/tensorflow/core/kernels/adjust_hue_op.cc
+++ b/tensorflow/core/kernels/adjust_hue_op.cc
@@ -1,5 +1,4 @@
 /* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -12,16 +11,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+#define EIGEN_USE_THREADS
+
+#if GOOGLE_CUDA
+#define EIGEN_USE_GPU
+#endif
+
 #include <memory>
-#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
 #include "tensorflow/core/framework/op_kernel.h"
 #include "tensorflow/core/framework/register_types.h"
 #include "tensorflow/core/framework/tensor.h"
 #include "tensorflow/core/framework/tensor_shape.h"
+#include "tensorflow/core/framework/tensor_types.h"
 #include "tensorflow/core/framework/types.h"
+#include "tensorflow/core/kernels/adjust_hue_op.h"
 #include "tensorflow/core/lib/core/status.h"
 #include "tensorflow/core/platform/logging.h"
 #include "tensorflow/core/util/work_sharder.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
 
 namespace tensorflow {
 
@@ -77,6 +84,7 @@ template <class Device>
 class AdjustHueOp;
 
 namespace internal {
+
 // Helper function to convert a RGB color to H-and-V-range. H is in the range
 // of [0, 6] instead of the normal [0, 1]
 static void rgb_to_hv_range(float r, float g, float b, float* h, float* v_min,
@@ -185,6 +193,7 @@ static void hv_range_to_rgb(float h, float v_min, float v_max, float* r,
 }
 }  // namespace internal
 
+
 template <>
 class AdjustHueOp<CPUDevice> : public AdjustHueOpBase {
  public:
@@ -237,4 +246,34 @@ class AdjustHueOp<CPUDevice> : public AdjustHueOpBase {
 REGISTER_KERNEL_BUILDER(Name("AdjustHue").Device(DEVICE_CPU),
                         AdjustHueOp<CPUDevice>);
 
+#if GOOGLE_CUDA
+template <>
+class AdjustHueOp<GPUDevice> : public AdjustHueOpBase {
+ public:
+  explicit AdjustHueOp(OpKernelConstruction* context)
+      : AdjustHueOpBase(context) {}
+
+  virtual void DoCompute(OpKernelContext* context, const ComputeOptions& options) override {
+    const Tensor* input = options.input;
+    const Tensor* delta = options.delta;
+    Tensor* output = options.output;
+    const int64 number_of_elements = input->NumElements();
+    GPUDevice device = context->eigen_gpu_device();
+    const auto stream = device.stream();
+    OP_REQUIRES(context, stream, errors::Internal("No GPU stream available."));
+    if (number_of_elements > 0) {
+      const float* input_data = input->flat<float>().data();
+      const float* delta_h = delta->flat<float>().data();
+      float* const output_data = output->flat<float>().data();
+      functor::AdjustHueGPU()(&device, number_of_elements, input_data, delta_h,
+                              output_data);
+    }
+  }
+};
+
+REGISTER_KERNEL_BUILDER(Name("AdjustHue").Device(DEVICE_GPU), AdjustHueOp<GPUDevice>);
+
+#endif
+
+//} // namespace functor
 }  // namespace tensorflow
--- a/tensorflow/core/kernels/adjust_hue_op.h
+++ b/tensorflow/core/kernels/adjust_hue_op.h
@@ -1,0 +1,42 @@
+/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+#ifndef _TENSORFLOW_CORE_KERNELS_ADJUST_HUE_OP_H
+#define _TENSORFLOW_CORE_KERNELS_ADJUST_HUE_OP_H
+
+#if GOOGLE_CUDA
+#define EIGEN_USE_GPU
+
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
+namespace tensorflow {
+
+typedef Eigen::GpuDevice GPUDevice;
+
+namespace functor {
+
+struct AdjustHueGPU {
+  void operator()(
+      GPUDevice* device,
+      const int64 number_of_elements,
+      const float* const input,
+      const float* const delta,
+      float* const output
+  );
+};
+
+} // namespace functor
+} // namespace tensorflow
+
+#endif // GOOGLE_CUDA
+#endif // _TENSORFLOW_CORE_KERNELS_ADJUST_HUE_OP_H
--- a/tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
@@ -1,0 +1,141 @@
+/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+
+#if GOOGLE_CUDA
+
+#define EIGEN_USE_GPU
+
+#include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/kernels/adjust_hue_op.h"
+#include "tensorflow/core/util/cuda_kernel_helper.h"
+
+namespace tensorflow {
+namespace internal {
+
+namespace {
+  typedef struct RgbTuple {
+    float r;
+    float g;
+    float b;
+  } RgbTuple;
+
+  typedef struct HsvTuple {
+    float h;
+    float s;
+    float v;
+  } HsvTuple;
+}  // anon namespace
+
+__device__ HsvTuple rgb2hsv_cuda(const float r, const float g, const float b)
+{
+  HsvTuple tuple;
+  const float M = fmaxf(r, fmaxf(g, b));
+  const float m = fminf(r, fminf(g, b));
+  const float chroma = M - m;
+  float h = 0.0f, s = 0.0f;
+  // hue
+  if (chroma > 0.0f) {
+    if (M == r) {
+      const float num = (g - b) / chroma;
+      const float sign = copysignf(1.0f, num);
+      h = ((sign < 0.0f) * 6.0f + sign * fmodf(sign * num, 6.0f)) / 6.0f;
+    } else if (M == g) {
+      h = ((b - r) / chroma + 2.0f) / 6.0f;
+    } else {
+      h = ((r - g) / chroma + 4.0f) / 6.0f;
+    }
+  } else {
+    h = 0.0f;
+  }
+  // saturation
+  if (M > 0.0) {
+    s = chroma / M;
+  } else {
+    s = 0.0f;
+  }
+  tuple.h = h;
+  tuple.s = s;
+  tuple.v = M;
+  return tuple;
+}
+
+__device__ RgbTuple hsv2rgb_cuda(const float h, const float s, const float v)
+{
+  RgbTuple tuple;
+  const float new_h = h * 6.0f;
+  const float chroma = v * s;
+  const float x = chroma * (1.0f - fabsf(fmodf(new_h, 2.0f) - 1.0f));
+  const float new_m = v - chroma;
+  const bool between_0_and_1 = new_h >= 0.0f && new_h < 1.0f;
+  const bool between_1_and_2 = new_h >= 1.0f && new_h < 2.0f;
+  const bool between_2_and_3 = new_h >= 2.0f && new_h < 3.0f;
+  const bool between_3_and_4 = new_h >= 3.0f && new_h < 4.0f;
+  const bool between_4_and_5 = new_h >= 4.0f && new_h < 5.0f;
+  const bool between_5_and_6 = new_h >= 5.0f && new_h < 6.0f;
+  tuple.r = chroma * (between_0_and_1 || between_5_and_6) +
+      x * (between_1_and_2 || between_4_and_5) + new_m;
+  tuple.g = chroma * (between_1_and_2 || between_2_and_3) +
+      x * (between_0_and_1 || between_3_and_4) + new_m;
+  tuple.b = chroma * (between_3_and_4 || between_4_and_5) +
+      x * (between_2_and_3 || between_5_and_6) + new_m;
+  return tuple;
+}
+
+__global__ void adjust_hue_nhwc(const int64 number_elements,
+                                const float * const __restrict__ input,
+                                float * const output,
+                                const float * const hue_delta)
+{
+  // multiply by 3 since we're dealing with contiguous RGB bytes for each pixel (NHWC)
+  const int64 idx = (blockDim.x * blockIdx.x + threadIdx.x) * 3;
+  // bounds check
+  if (idx > number_elements - 1) {
+    return;
+  }
+  const float delta = hue_delta[0];
+  const HsvTuple hsv = rgb2hsv_cuda(input[idx], input[idx + 1], input[idx + 2]);
+  // hue adjustment
+  float new_h = fmodf(hsv.h + delta, 1.0f);
+  if (new_h < 0.0f) {
+    new_h = fmodf(1.0f + new_h, 1.0f);
+  }
+  const RgbTuple rgb = hsv2rgb_cuda(new_h, hsv.s, hsv.v);
+  output[idx] = rgb.r;
+  output[idx + 1] = rgb.g;
+  output[idx + 2] = rgb.b;
+}
+} // namespace internal
+
+
+namespace functor {
+
+void AdjustHueGPU::operator()(
+  GPUDevice* device,
+  const int64 number_of_elements,
+  const float* const input,
+  const float* const delta,
+  float* const output
+) {
+  const auto stream = device->stream();
+  const CudaLaunchConfig config = GetCudaLaunchConfig(number_of_elements, *device);
+  const int threads_per_block = config.thread_per_block;
+  const int block_count = (number_of_elements + threads_per_block - 1) / threads_per_block;
+  internal::adjust_hue_nhwc<<<block_count, threads_per_block, 0, stream>>>(
+    number_of_elements, input, output, delta
+  );
+}
+} // namespace functor
+}  // namespace tensorflow
+#endif // GOOGLE_CUDA
--- a/tensorflow/core/platform/default/build_config.bzl
+++ b/tensorflow/core/platform/default/build_config.bzl
@@ -7,7 +7,7 @@ load("//tensorflow:tensorflow.bzl", "if_not_mobile")
 # configure may change the following lines
 WITH_GCP_SUPPORT = False
 WITH_HDFS_SUPPORT = False
-WITH_JEMALLOC = True
+WITH_JEMALLOC = False
 
 # Appends a suffix to a list of deps.
 def tf_deps(deps, suffix):
--- a/tensorflow/python/ops/image_ops_test.py
+++ b/tensorflow/python/ops/image_ops_test.py
@@ -299,7 +299,7 @@ class AdjustHueTest(test_util.TensorFlowTestCase):
     return y_v.reshape(x_np.shape)
 
   def _adjustHueTf(self, x_np, delta_h):
-    with self.test_session(use_gpu=False):
+    with self.test_session(use_gpu=True):
       x = constant_op.constant(x_np)
       y = image_ops.adjust_hue(x, delta_h)
       y_tf = y.eval()


tensorflow/core/kernels/adjust_hue_op.h
```diff
@@ -0,0 +1,64 @@
+#include "tensorflow/core/framework/op_kernel.h"
```
drpngx: You need the LICENSE text.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,142 @@
+#if GOOGLE_CUDA
```
drpngx: You need the LICENSE text.

tensorflow/core/kernels/adjust_hue_op.h
mkolod: Will do that, thanks!

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: Will fix that, thanks!

zheng-xq:
Thanks a lot for the contribution! I'll look into it.

mkolod:
BTW, by the [following test](https://gist.github.com/mkolod/71d899687e21f91e1e870832d4de5569), run on a machine with 32 GB RAM, a 6-core Intel Core i7-5930K CPU @ 3.50GHz and an NVIDIA GTX 1080 GPU produced a GPU-based speedup factor of 17.3, given both the use of a fused CPU and GPU kernel. This was based on a test with a 4D random image tensor with a batch size of 64 and "image" size of 256x256x3. The GTX 1080 throughput 31,950 images per second, while the Core i7-5930K throughput was 1,845 images per second. Longer running times, e.g. running tf.while_loop 100k times, can bring the speedup factor to about 30x on this reference hardware. Both 3D and 4D tensors work, but of course 4D tensors perform better due to batching. Note that for uint8 inputs (e.g. from tf.image_jpeg_decode), the results of the CPU and GPU kernel are identical, pixel for pixel. For inputs of int32, with a numeric range of [0, 256], a small fraction of pixels differ by up to 0.3% in intensity (1/255), which is caused by the fact that the Python wrapper for adjust_hue() does scaling, and scaling a range of [0, 255] by 1E32 causes precision problems. That's a very small error for that unusual case. Numbers in range of [0, 255] represented as floats, or as uint8 (scaled by 255 for float conversion by convert_image_dtype() inside tf.image_adjust_hue()) return exactly the same results on the CPU and GPU. Note that this is just an implementation of adjust_hue, and doesn't necessarily work for tf.image.random_hue, because the hue delta taken by the CUDA kernel is a scalar rather than a whole tensor. This can be addressed if need be to cover both adjust_hue() and random_hue().

tensorflow/core/kernels/adjust_hue_op.h
```diff
@@ -0,0 +1,75 @@
...
+    const Tensor* input;
+    const Tensor* delta;
+    Tensor* output;
+    int64 channel_count;
+  };
+
+  virtual void DoCompute(OpKernelContext* context,
+                         const ComputeOptions& options) = 0;
+
+  void Compute(OpKernelContext* context) override {
```
zheng-xq: I think that is too much code in the header file. You can leave the declaration here, but move the definition back to the .cc file.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+
+namespace tensorflow {
+
+typedef Eigen::GpuDevice GPUDevice;
+
+namespace internal {
+
+typedef struct RgbTuple {
+
+  float r, g, b;
```
zheng-xq: Most tensorflow code goes without the empty line above and below, and

float r;
float g;
float b;

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+  const float M = fmaxf(r, fmaxf(g, b));
+  const float m = fminf(r, fminf(g, b));
+  const float chroma = M - m;
+
+  float h = 0.0f, s = 0.0f;
+
+    // hue
+    if (chroma > 0.0f) {
+      if (M == r) {
+
```
zheng-xq: Consistently remove empty lines at the beginning and end of each block. 

In general, TensorFlow wants to fit more content within one page. Adding new line only between important functional blocks. There is no need for cosmetic new lines.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+  tuple.b =  chroma * (between_3_and_4 || between_4_and_5) +
+                      x * (between_2_and_3 || between_5_and_6) + new_m;
+
+  return tuple;
+}
+
+__global__ void adjust_hue_nhwc(const int number_elements,
+                                const float * const input, float * const output, const float * const hue_delta) {
+
+        const float delta = hue_delta[0];
```
zheng-xq: Set the indentation to two spaces consistently.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+	// hue adjustment
+	const float new_h = fmodf(hsv.h + delta, 1.0f);
+
+        const RgbTuple rgb = hsv2rgb_cuda(new_h, hsv.s, hsv.v);
+        output[idx] = rgb.r;
+        output[idx + 1] = rgb.g;
+        output[idx + 2] = rgb.b;
+
+}
+}
```
zheng-xq: Leave a comment marking the end of the namespace.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+__global__ void adjust_hue_nhwc(const int number_elements,
+                                const float * const input, float * const output, const float * const hue_delta) {
+
+        const float delta = hue_delta[0];
+
+	// multiply by 3 since we're dealing with contiguous RGB bytes for each pixel
+	const int idx = (blockDim.x * blockIdx.x + threadIdx.x) * 3;
+
+	// bounds check
+	if (idx > number_elements) {
```
zheng-xq: Out of curiosity, have you tried to do this with either variable number of blocks, or fixed number of blocks and each handling variable number of elements? Does it make a difference in performance?

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+		OP_REQUIRES(context, stream, errors::Internal("No GPU stream available."));
+
+		if (number_elements > 0) {
+
+
+			const CudaLaunchConfig config = GetCudaLaunchConfig(number_elements, d);
+			const float * const input_data = input->flat<float>().data();
+			const float * const delta_h = delta->flat<float>().data();
+			float * const output_data = output->flat<float>().data();
+                        const int threads_per_block = config.thread_per_block;
```
zheng-xq: Wrong indentation.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+                        const int block_count = (number_elements + threads_per_block - 1) / threads_per_block;
```
zheng-xq: This or Eigen::divup. Really up to you.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,182 @@
...
+
+	// bounds check
+	if (idx > number_elements) {
+		return;
+	}
+
+        const HsvTuple hsv = rgb2hsv_cuda(input[idx], input[idx+1], input[idx+2]);
+
+	// hue adjustment
+	const float new_h = fmodf(hsv.h + delta, 1.0f);
```
zheng-xq: Does fmodf handle negative delta?

tensorflow/python/ops/image_ops_impl.py
```diff
@@ -1107,27 +1107,7 @@ def adjust_hue(image, delta, name=None):
     # Remember original dtype to so we can convert back if needed
     orig_dtype = image.dtype
     flt_image = convert_image_dtype(image, dtypes.float32)
-
-    # TODO(zhengxq): we will switch to the fused version after we add a GPU
-    # kernel for that.
-    fused = os.environ.get('TF_ADJUST_HUE_FUSED', '')
```
zheng-xq: For safety concern, leave the default as "fused". If the user specified "0", or "false", use the old code path. 

After a few weeks, if we don't see anything broken, we will remove the old path.

tensorflow/core/kernels/adjust_hue_op_test.cc
```diff
@@ -0,0 +1,56 @@
...
+TEST_F(AdjustHueOpTest, Simple_1113) {
+
+  TF_EXPECT_OK(NodeDefBuilder("adjust_hue_op", "AdjustHue")
+      .Input(FakeInput(DT_FLOAT))
+      .Input(FakeInput(DT_FLOAT))
+      .Finalize(node_def()));
+
+  TF_EXPECT_OK(InitOp());
+
+  AddInputFromArray<float>(TensorShape({1, 1, 1, 3}), {127, 137, 225});
```
zheng-xq: There are a lof of python tests under

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/image_ops_test.py

Does this test do something special? 

Please make sure all the tests will use the new GPU kernel, and pass after this change.

tensorflow/core/kernels/adjust_hue_op.h
mkolod: Will do, thanks!

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: Sounds good!

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: Will do!

tensorflow/python/ops/image_ops_impl.py
mkolod: Sounds good!

tensorflow/core/kernels/adjust_hue_op_test.cc
mkolod: Ah OK, in that case, I'll just use the Python one, that should be enough.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @zheng-xq I tried blocks from size 32 to size 1,024. The difference was very small. The only bad performance was for block sizes less than 32.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: Thanks!

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @zheng-xq It does, plus since the fmodf is around 1.0f, the delta can be even a multiple of 1 or -1. But at a minimum, it does support the range [-1, 1].

mkolod:
@zheng-xq I made all the changes based on your requests. I think it's in a state that's acceptable based on your first review. The tests pass, as do my own experiments, for positive and negative delta. The code was refactored according to your spec. Please let me know if this is good enough to run the CI process.

drpngx:
Jenkins, test this please.

mkolod:
@drpngx @zheng-xq Everything that depends on the sanity checks step fails, because sanity checks don't pass. Looking at the log, I get the following message: 

```FAIL: buildifier found errors and/or warnings in above BUILD files.
1530a1531
>     linkstatic = 1,
1537d1537
<     linkstatic = 1,
1542d1541
<     prefix = "adjust_hue_op",
1543a1543
>     prefix = "adjust_hue_op",
Please fix manually or run buildifier <file> to auto-fix.```

I never encountered this problem while building locally. I can make that change, but I'm wondering if that won't trigger other problems. Is there a way to run this sanity check step locally, without the entire CI infra, to see if that resolves the problem? Even in the absence of this change, I can build the _pywrap_tensorflow.so and pass  //tensorflow/python:image_ops_test, etc.

Please let me know how you would like me to proceed.

**Comment by mkolod** [comment]
@drpngx @zheng-xq While running ./tensorflow/tools/ci_build/ci_sanity.sh locally, I get

> ./tensorflow/tools/ci_build/ci_sanity.sh: line 285: buildifier: command not found

So, I can't iterate locally to address this. 

Moreover, I seem to be getting another set of messages locally that the execution of the sanity check script doesn't seem to produce for you on Jenkins:

> FAIL: mismatch in packaged licenses and external dependencies
> Please remove the licenses for the following external dependencies:
> //third_party/hadoop
> @boringssl//
> @curl//
> @jemalloc//

These don't seem to show up in your CI build.

Please let me know what I can do to make it work.

**Comment by drpngx** [comment]
You can do the link check with our script.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/ci_sanity.sh#L84

It's basically asking you to sort.

**Comment by mkolod** [comment]
@drpngx Ah ok, great! Thank you! I'll check this ASAP.

File: tensorflow/core/kernels/BUILD [hunk 14]
```diff
@@ -1524,10 +1524,24 @@ tf_kernel_library(
...
+    name = "adjust_hsv_op_common",
+    srcs = ["adjust_hsv_op_common.cc"],
+    hdrs = ["adjust_hsv_op_common.h"],
+    visibility = ["//visibility:private"],
+    deps = [
+        "//tensorflow/core:framework",
+        "//tensorflow/core:lib",
+        "//third_party/eigen3",
+    ],
+    linkstatic = 1,
```
zheng-xq: Is "linkstatic = 1" necessary here?

tensorflow/core/kernels/adjust_hsv_op_common.h
```diff
@@ -0,0 +1,44 @@
...
+#include "tensorflow/core/framework/tensor.h"
+
+#ifndef TENSORFLOW_CORE_KERNELS_ADJUST_HSV_OP_COMMON_H_
+#define TENSORFLOW_CORE_KERNELS_ADJUST_HSV_OP_COMMON_H_
+
+namespace tensorflow {
+
+class AdjustHueOpBase : public OpKernel {
+
+protected:
```
zheng-xq: The indentation within this class is lost. Please refer to the style guide as a reference. 

https://google.github.io/styleguide/cppguide.html#Class_Format

tensorflow/core/kernels/adjust_hsv_op_common.h
```diff
@@ -0,0 +1,44 @@
...
+};
+
+virtual void DoCompute(OpKernelContext* context,
+                       const ComputeOptions& options) = 0;
+
+virtual void Compute(OpKernelContext* context);
+};
+
+template <class Device>
+class AdjustHueOp {};
```
zheng-xq: I would prefer not to have a class definition at all, if possible.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,147 @@
...
+  const int idx = (blockDim.x * blockIdx.x + threadIdx.x) * 3;
+  // bounds check
+  if (idx > number_elements) {
+    return;
+  }
+  const HsvTuple hsv = rgb2hsv_cuda(input[idx], input[idx+1], input[idx+2]);
+  // hue adjustment
+  float new_h = fmodf(hsv.h + delta, 1.0f);
+  if (new_h < 0.0f) {
+    new_h = fmodf(1.0f + new_h, 1.0f);
```
zheng-xq: I like this way much better. :)

Just to confirm, a conditional fmodf is actually faster than a unconditional fmodf? Thanks.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,147 @@
...
+  const RgbTuple rgb = hsv2rgb_cuda(new_h, hsv.s, hsv.v);
+  output[idx] = rgb.r;
+  output[idx + 1] = rgb.g;
+  output[idx + 2] = rgb.b;
+}
+} // namespace internal
+
+template <>
+class AdjustHueOp<GPUDevice> : public AdjustHueOpBase {
+public:
```
zheng-xq: Same here. The indentation within the class is lost. 

https://google.github.io/styleguide/cppguide.html#Class_Format

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,147 @@
...
+    const int threads_per_block = config.thread_per_block;
+    const int block_count = (number_elements + threads_per_block - 1) / threads_per_block;
+    internal::adjust_hue_nhwc<<<block_count, threads_per_block, 0, stream>>>(
+      number_elements, input_data, output_data, delta_h
+    );
+  }
+}
+};
+
+REGISTER_KERNEL_BUILDER(Name("AdjustHue").Device(DEVICE_GPU),
```
zheng-xq: Ah! I am a bit surprised that this worked at all. 

Older nvcc cannot handle OpKernel and its registration process. So we had to do all the OpKernel in the CPU translation unit, and only call out an launch function into the Cuda translation unit. 

It might be safer to follow that convention for the multiple Cuda compilers we are supporting. Here is an example, there are many in the kernels directory. 

1. GPU kernel registration in the CPU translation unit: 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/segment_reduction_ops.cc#L315
2. The kernel does all the OpKernel related things in this file, and call out to its GPU functor at. Only simpler data structure crosses the file boundary. 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/segment_reduction_ops.cc#L291
3. The GPU functor is defined here: 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc#L59
4. And instantiated here: 
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc#L95

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,147 @@
...
+__global__ void adjust_hue_nhwc(const int number_elements,
+                                const float * const input, float * const output, const float * const hue_delta) {
+  const float delta = hue_delta[0];
+  // multiply by 3 since we're dealing with contiguous RGB bytes for each pixel (NHWC)
+  const int idx = (blockDim.x * blockIdx.x + threadIdx.x) * 3;
+  // bounds check
+  if (idx > number_elements) {
+    return;
+  }
+  const HsvTuple hsv = rgb2hsv_cuda(input[idx], input[idx+1], input[idx+2]);
```
zheng-xq: Out of curiosity, is it faster to have uncoalesced memory access, or staging through shared memory?

zheng-xq (commented):
Thanks a lot for the new change! It looks much better now!

A few more comments enclosed.

zheng-xq:
In addition, did you include the change to flip the default? It's okay to leave it as a different CL if you want. 

Also we need to enable more tests on GPU by setting this to True: 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/image_ops_test.py#L302

You can test the GPU version with --config=cuda, and the CPU version without --config=cuda.

tensorflow/core/kernels/BUILD
mkolod: @zheng-xq It's not, I just checked, I'll remove it.

tensorflow/core/kernels/adjust_hsv_op_common.h
mkolod: I'll fix that, thanks! I just discovered the [Eclipse formatting style template](https://github.com/google/styleguide/blob/gh-pages/eclipse-cpp-google-style.xml) to make sure the code follows Google's indentation rules now. I hope that's enough to ensure following style.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @zheng-xq I actually haven't tested the shared memory case. It might be slightly better that way, let me know if you'd like me to try. If it's OK with you, I can do it while working on the next kernel I'm about to PR (saturation). Let me know what you think.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @zheng-xq I'll fix that, thanks!

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @zheng-xq Actually I haven't checked, but I will before re-submitting. It definitely doesn't hurt the math for it to be unconditional. The lack of warp divergence may help, though if some threads are running an instruction that's unnecessary seems to be about the same as stalling them. I'll check the perf of both cases and will let you know.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @zheng-xq It worked with nvcc from CUDA 8, but yes, I haven't checked with older compilers. I can try doing it the way you suggested if that helps generalize the build.

mkolod:
@zheng-xq Yes, I changed the default back to the unfused kernels, so now the env var is needed again, like it used to be. I'll make the change in the test as well. Thanks!

tensorflow/core/kernels/adjust_hsv_op_common.h
mkolod: @zheng-xq I think it will be possible to remove it from this file when the op part of the GPU impl gets moved to the same compilation unit as the CPU code, and the GPU file only has a functor, as you pointed out below.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @zheng-xq It seems that shared memory didn't help that much, but the load issue gets addressed quite well by the texture cache by using __restrict__ on the input pointer (see [here](https://devblogs.nvidia.com/parallelforall/cuda-pro-tip-optimize-pointer-aliasing/) for more details).

mkolod:
@zheng-xq @drpngx I'm running CI now and making a few last changes (specifically [this one](https://github.com/tensorflow/tensorflow/pull/6818#discussion_r97458403)), so please don't review or merge yet. I'll make a comment here when the code is ready for final review, build and merge.

mkolod:
@zheng-xq Could you review this? Also, seems like Linux tests have failed due to some networking problem (see [here](https://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/3603/console)). Could you have a look? Thanks!

vrv:

vrv:
@mkolod we're swamped because of the dev summit tomorrow -- XQ will have more time afterwards :)

mkolod:
@vrv No worries! See you at the summit! :)

mkolod:
@zheng-xq Could you review it? Thanks! :)

tensorflow/core/kernels/adjust_hue_op.h
```diff
@@ -0,0 +1,69 @@
...
+#define _TENSORFLOW_CORE_KERNELS_ADJUST_HUE_OP_H
+
+#include <memory>
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+#include "tensorflow/core/framework/device_base.h"
+#include "tensorflow/core/framework/op_kernel.h"
+#include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/framework/tensor.h"
+#include "tensorflow/core/framework/tensor_shape.h"
+#include "tensorflow/core/framework/types.h"
```
zheng-xq: Sorry that I didn't make this clear the first time. In general, we cannot make any major classes from "tensorflow/core/framework" visible to .cu.cc. That means no Tensor, no OpKernelContext there. When we pull the code inside, they might not work.

The common practice is to only pass either Eigen tensor, or just raw pointers between the .cc and .cu.cc files.

zheng-xq (commented):
Thanks a lot for the changes! We are getting close now. Only a few minor comments.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,145 @@
...
+  const float M = fmaxf(r, fmaxf(g, b));
+  const float m = fminf(r, fminf(g, b));
+  const float chroma = M - m;
+  float h = 0.0f, s = 0.0f;
+  // hue
+  if (chroma > 0.0f) {
+    if (M == r) {
+      const float num = (g - b) / chroma;
+      const float sgn = num < 0.0f;
+      const float sign = powf(-1.0f, sgn);
```
mkolod: @jlebar nvcc doesn't offer copysign for __device__ functions. There is signbit, but that only addresses line 48, not 49. One can of course instead do "sgn ? -1 : 1" on line 49, but conditionals cause warp divergence. Comments?

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,145 @@
...
+#if GOOGLE_CUDA
+
+#define EIGEN_USE_GPU
+
+#include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/kernels/adjust_hue_op.h"
+#include "tensorflow/core/util/cuda_kernel_helper.h"
+
+namespace tensorflow {
+namespace internal {
```
mkolod: Will fix that, thanks!

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,145 @@
...
+  tuple.v = M;
+  return tuple;
+}
+
+__device__ RgbTuple hsv2rgb_cuda(const float h, const float s, const float v)
+{
+  RgbTuple tuple;
+  const float new_h = h * 6.0f;
+  const float chroma = v * s;
+  const float x = chroma * (1.0 - fabsf(fmodf(new_h, 2.0f) - 1.0f));
```
mkolod: Yes, good point.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
mkolod: @jlebar Of course one can also implement an ugly branchless solution in the absence of copysign on device, for example

template <typename T> int sgn(T val) {
    return (T(0) < val) - (val < T(0));
}

mkolod: or 
2 * signbit(val) - 1
because signbit is implemented in the CUDA math lib, but copysign is not.

tensorflow/core/kernels/adjust_hue_op.cc
```diff
@@ -185,56 +136,100 @@ static void hv_range_to_rgb(float h, float v_min, float v_max, float* r,
 }
 }  // namespace internal
 
-template <>
-class AdjustHueOp<CPUDevice> : public AdjustHueOpBase {
+template<typename Device>
+class AdjustHueOp : public OpKernel {
```
mkolod: @zheng-xq Somehow I thought that you didn't want the subclassing, in addition to suggesting the functor pattern to address the GPU compilation units. If it's OK to leave for now, I can fix that when I do the saturation PR later. Hopefully it's not too much template bloat relative to Eigen. :)

tensorflow/core/kernels/adjust_hue_op.h
mkolod: Oh OK, I wish I'd known that earlier! :)

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
```diff
@@ -0,0 +1,145 @@
...
+    const float * const delta_h = delta->flat<float>().data();
+    float * const output_data = output->flat<float>().data();
+    const int threads_per_block = config.thread_per_block;
+    const int block_count = (number_elements + threads_per_block - 1) / threads_per_block;
+    internal::adjust_hue_nhwc<<<block_count, threads_per_block, 0, stream>>>(
+        number_elements, input_data, output_data, delta_h
+    );
+  }
+}
+} // namespace functor
```
mkolod: Sounds good.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
jlebar: FWIW `sgn ? -1 : 1` should be lowered using PTX select, which does not cause divergent.  But any of these options seems much better than powf.

jlebar: (FTR, since I checked, clang has no problem with copysign, because we implement all of C++11 math.

    $ echo '__global__ void f(float a, float* b) { *b = copysign(1, a); }' |\
      clang++ -O2 -x cuda -c -S --cuda-device-only - -o -
)

mkolod: Maybe not in clang, but I'm having issues with gcc (4.8.4) interacting with nvcc (CUDA 8.0.53).

jlebar: Yes, I totally believe that nvcc doesn't handle this correctly.  Sorry, didn't want to suggest that you needed to use copysign, just that we *will* be able to use copysign once we've rid ourselves of the nvcc dependency.

mkolod: Actually it turns out that the reason was the issue of a conflict between copysign in gcc and nvcc. Since in nvcc, copysign delegates to the device function copysignf, and one of the imported headers must have transitively imported std::copysign from cmath (I assume removing std via using), there was a conflict. Spelling out the name as copysignf makes things work just fine with gcc + nvcc in the mix. The same goes for using, say, fmodf rather than fmod - it ensures no ambiguity even when cmath functions are "used."

jlebar: (FTR that's another problem that I believe we've fixed in clang. :)

In the absence of a benchmark, I don't care if we do copysign or a select on signbit, although having looked at the ptx, I expect that the signbit select will be faster.  fmodf is already quite involved, so it might be lost in the noise.

jlebar:
Please ping me if and when you'd like me to look at this again.

dandelionmane:
@mkolod Ping, it looks like this is really close to merging. Also, there's a conflict to resolve now.

mkolod:
Thanks @dandelionmane, will make changes, time permitting.

mkolod:
@zheng-xq Any remaining concerns? Regarding formatting, I ended up using `clang-format --fallback-style="Google" < input.cc` so I hope that is enough to resolve formatting. I think I addressed all the other concerns for the last review as well.

tensorflow/core/kernels/adjust_hue_op_gpu.cu.cc
jlebar: You can place the functions in an anon ns too.

namespace tensorflow { namespace internal { namespace { ... }}}

jlebar (approved):

jlebar:
Hm, didn't mean to approve globally.  Definitely still need XQ's sign-off.

jhseu:
@zheng-xq friendly ping

jlebar:
FYI as a general rule, we shouldn't expect reviewers to review code with less latency than it takes reviewees to respond to comments.  If nothing else, this encourages people to respond to comments quickly, which saves reviewer time, letting them keep reviews in cache.

zheng-xq:
LGTM.

Thanks for the contribution!

PR #6821, opened 2017-01-13
Make tfdbg tests pass in Windows Bazel build

caisq:


gunan (approved):

caisq:
Tomorrow, I'll test internally to make sure that moving debug_ops out of array_not_windows is okay.

drpngx:
The upgrade test should be fixed by previous push, and the `seq2seq` and `rnn_core` tests seem to timeout frequently. The `sync_replicas_optimizer_test` seems unrelated:

```
InvalidArgumentError (see above for traceback): Cannot assign a device to node 'save/RestoreV2_3': Could not satisfy explicit device specification '/job:ps/task:1/device:CPU:0' because no devices matching that specification are registered in this process; available devices: /job:ps/replica:0/task:0/cpu:0, /job:worker/replica:0/task:0/cpu:0, /job:worker/replica:0/task:1/cpu:0, /job:worker/replica:0/task:2/cpu:0
```

Issue #6509, opened 2016-12-26
Issue with tf.one_hot() in 0.12.0 in GPU mode (CUDA_ERROR_ILLEGAL_ADDRESS)


SergeyBykov1:
I'm using a LeNet-5 mnist example from Udacity's course. Link to the source code is below.
Training works ok on a CPU (config = tf.ConfigProto(device_count = {'GPU': 0})),
but fails in a GPU mode with the following 'CUDA_ERROR_ILLEGAL_ADDRESS' error:

> I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\core\common_runtime\gpu\gpu_device.cc:885] Found device 0 with properties:
name: GeForce GTX 1060 6GB
major: 6 minor: 1 memoryClockRate (GHz) 1.7845
pciBusID 0000:01:00.0
Total memory: 6.00GiB
Free memory: 5.01GiB
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\core\common_runtime\gpu\gpu_device.cc:906] DMA: 0
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\core\common_runtime\gpu\gpu_device.cc:916] 0:   Y
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\core\common_runtime\gpu\gpu_device.cc:975] Creating TensorFlow device (/gpu:0) -> (device: 0, name: GeForce GTX 1060 6GB, pci bus id: 0000:01:00.0)
Training...
E c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\cuda\cuda_event.cc:49] Error polling for event status: failed to query event: **CUDA_ERROR_ILLEGAL_ADDRESS**
F c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\core\common_runtime\gpu\gpu_event_mgr.cc:198] Unexpected Event status: 1

I have reproduced same error on two setups.
Environment 1 (Home PC):
- windows 10; 
- latest anaconda 4.2.0, python 3.5; 
- cuda 8
- cudnn 5.1 (for win10)
- tensorflow 0.12.0 gpu (https://storage.googleapis.com/tensorflow/windows/gpu/tensorflow_gpu-0.12.0-cp35-cp35m-win_amd64.whl)
- GeForce GTX 1060 6Gb

[...]

PR #6822, opened 2017-01-13
Disable building the MatrixDiag and OneHot kernels for GPU on Windows.

mrry:
This is a temporary workaround for issue #6509, while the root cause of the runtime errors is under investigation.

### DIFF
--- a/tensorflow/contrib/cmake/tf_core_kernels.cmake
+++ b/tensorflow/contrib/cmake/tf_core_kernels.cmake
@@ -112,7 +112,11 @@ if(WIN32 AND tensorflow_ENABLE_GPU)
   add_dependencies(tf_core_kernels_cpu_only tf_core_cpu)
   # Undefine GOOGLE_CUDA to avoid registering unsupported GPU kernel symbols.
   get_target_property(target_compile_flags tf_core_kernels_cpu_only COMPILE_FLAGS)
-  set(target_compile_flags "${target_compile_flags} /UGOOGLE_CUDA")
+  if(target_compile_flags STREQUAL "target_compile_flags-NOTFOUND")
+    set(target_compile_flags "/UGOOGLE_CUDA")
+  else()
+    set(target_compile_flags "${target_compile_flags} /UGOOGLE_CUDA")
+  endif()
   set_target_properties(tf_core_kernels_cpu_only PROPERTIES COMPILE_FLAGS ${target_compile_flags})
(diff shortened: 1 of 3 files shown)


mrry:
@gunan I didn't manage to get the GPU PR build started... here's the error: http://ci.tensorflow.org/job/tf-pr-win-cmake-gpu/5/console

```
21:59:28 c:\tf_jenkins\home\workspace\tf-pr-win-cmake-gpu>call tensorflow/tools/ci_build/windows/gpu/cmake/run_py.bat 
21:59:28 'tensorflow' is not recognized as an internal or external command,
21:59:28 operable program or batch file.
21:59:28 
21:59:28 c:\tf_jenkins\home\workspace\tf-pr-win-cmake-gpu>exit 1 
```

Any ideas? I'll test it on my workstation tomorrow if not....

gunan:
Got it to run now,
http://ci.tensorflow.org/job/tf-pr-win-cmake-gpu/7/
Apparently we need to provide PR number slightly differently.

mrry:
Thanks for looking into that! Trying a new run, which should fix the failure: http://ci.tensorflow.org/job/tf-pr-win-cmake-gpu/8/

mrry:
Once more without a silly typo this time: http://ci.tensorflow.org/job/tf-pr-win-cmake-gpu/9/

mrry:
Well the build timed out before reaching the `one_hot_op_test.py`. :(

@gunan Any ideas for how to make it run longer? I also made a similar change to the MatrixDiag ops and `diag_op_test.py` has failed, but when I look in the workspace at the testing log, I don't see any errors.... Is it maybe failing because it's not using the GPU (e.g. because of some smarts put in to ensure that we're actually testing GPU implementations)?

gunan:
ah, I have upped the timeout to 5 hours and restarted the build here:
http://ci.tensorflow.org/job/tf-pr-win-cmake-gpu/10/

mrry:
Five hours ought to be enough for **any** test run! :P

mrry:
@gunan Looks like `one_hot_op_test.py` and `trace_op_test.py` succeeded, but `diag_op_test.py` failed for mysterious reasons. 2 out of 3 isn't bad:

http://ci.tensorflow.org/job/tf-pr-win-cmake-gpu/10/console

Shall we merge this, since it should at least prevent ugly crashes?

gunan:
Sure, sgtm.
The remaining issues are known to me so the tests are all OK.

PR #6829, opened 2017-01-13
./configure: A better way to exclude fetching //tensorflow/examples/android/...

meteorcloudy:
Bazel at HEAD currently [breaks TF Windows build](http://ci.bazel.io/job/TensorFlow/BAZEL_VERSION=HEAD,PLATFORM_NAME=windows-x86_64/601/console).

After https://github.com/bazelbuild/bazel/commit/0b9ebfee8d7f3bd5cf68f4dba4790bf3fed9349d, `//tensorflow/examples/android/...` has to be excluded during bazel fetch.

Using the workaround mentioned in https://github.com/bazelbuild/bazel/issues/2220, so that it works on both Linux and Windows.

@gunan

### DIFF
--- a/configure
+++ b/configure
@@ -23,13 +23,8 @@ function bazel_clean_and_fetch() {
   # TODO(pcloudy): Re-enable it after bazel clean --expunge is fixed.
   if ! is_windows; then
     bazel clean --expunge
-    # TODO(https://github.com/bazelbuild/bazel/issues/2220) Remove the nested `bazel query`.
-    bazel fetch $(bazel query "//tensorflow/... -//tensorflow/examples/android/...")
-  else
-    # TODO(pcloudy): Also filter out //tensorflow/examples/android/... on Windows after
-    # https://github.com/bazelbuild/bazel/issues/2248 is fixed.
-    bazel fetch //tensorflow/...
   fi
+  bazel fetch "//tensorflow/... -//tensorflow/examples/android/..."
 }
 
 ## Set up python-related environment settings


configure
```diff
@@ -23,13 +23,8 @@ function bazel_clean_and_fetch() {
...
   if ! is_windows; then
     bazel clean --expunge
-    # TODO(https://github.com/bazelbuild/bazel/issues/2220) Remove the nested `bazel query`.
-    bazel fetch $(bazel query "//tensorflow/... -//tensorflow/examples/android/...")
-  else
-    # TODO(pcloudy): Also filter out //tensorflow/examples/android/... on Windows after
-    # https://github.com/bazelbuild/bazel/issues/2248 is fixed.
-    bazel fetch //tensorflow/...
   fi
+  bazel fetch "//tensorflow/... -//tensorflow/examples/android/..."
```
drpngx: That removes the fetch for examples/android on non-windows too?

gunan: @drpngx is right.
We already comment out most android fetch targets by default. If anything, I would lean towards fetching everything even in windows.

meteorcloudy: @aj-michael Since you are the author of https://github.com/bazelbuild/bazel/commit/0b9ebfee8d7f3bd5cf68f4dba4790bf3fed9349d, can you comment on this? Should we exclude `//tensorflow/examples/android/...` or not? What I observed is that `bazel fetch //tensorflow/...` is fine with bazel 0.4.3, but failing with Bazel@HEAD

aj-michael: Yes, bazel fetch on an android_binary will intentionally fail if no android_sdk_repository is set up starting with the next Bazel release. I discussed the reasoning a bit in the Bazel issue associated with that commit.

meteorcloudy: @gunan Looks like we need this PR on both Linux and Windows for the next Bazel Release.

aj-michael: @meteorcloudy why do you say that? It should work on Linux with the next Bazel release. The reason that I added the `bazel fetch $(bazel query "//tensorflow/... -//tensorflow/examples/android/...")` line in https://github.com/tensorflow/tensorflow/pull/6225 was to ensure that configure would work with Bazel from HEAD.

meteorcloudy: Since `bazel fetch $(bazel query "//tensorflow/... -//tensorflow/examples/android/..."` doesn't work on Windows due to https://github.com/bazelbuild/bazel/issues/2248, I replaced it with `bazel fetch "//tensorflow/... -//tensorflow/examples/android/..."` in this change so that `./configure` will work both on Linux and Windows. Do you think it's a reasonable change?

BTW, I am trying to resolve this failure:
http://ci.bazel.io/job/TensorFlow/BAZEL_VERSION=HEAD,PLATFORM_NAME=windows-x86_64/606/console

aj-michael: This change will work for now. The reason that I opted for the nested query instead of what you are proposing now is that _if_ someone from Bazel team gets around to fixing https://github.com/bazelbuild/bazel/issues/2220, I suspect that the fix for it would break this workaround.

And just to confirm, yes that failure is due to https://github.com/bazelbuild/bazel/commit/0b9ebfee8d7f3bd5cf68f4dba4790bf3fed9349d which I thought I resolved in #6225, but did not test on Windows. Sorry about that.

drpngx: So, to summarize, we need to exclude fetch of `//tensorflow/examples/android/...` from all targets, not just windows?

aj-michael:
That is correct.

drpngx:
Jenkins, test this please.

drpngx:
Thanks!

drpngx:
Ignore test failure. I think that division returns a float, but also we're going to remove that test in later pushes.

configure
gunan: @andrewharp Would android build be happy with this change?

PR #6830, opened 2017-01-13
Update einsum to check whether uncompacted_shape has multiple None values

### LINKED ISSUE #6824: tf.einsum fails on evaluation in some cases
Hello,
I was a user of 'reshape-matmul-reshape' approach, however it felt me a bit messy, so I am trying some alternatives that do this automatically.
I had tried `tf.tensordot` before, however it did not work well when shapes are partially known (#6682).
As an alternative, I tried to use `tf.einsum`.
However, similar to `tf.tensordot`, it cannot handle some variables with partially-known-shape, for instance, the below codes fails on evaluation.

```
a = tf.placeholder('float32', shape=[None, None, 100])
b = tf.placeholder('float32', shape=[100, 300])
result_einsum = tf.einsum('ijk,kl->ijl', a, b)
a_value = np.random.randn(10, 20, 100)
b_value = np.random.randn(100, 300)
sess.run(result_einsum, {a: a_value, b: b_value})  # failure
```

This is the error message:
```
InvalidArgumentError (see above for traceback): only one input size may be -1, not both 0 and 1
         [[Node: Reshape_5 = Reshape[T=DT_FLOAT, Tshape=DT_INT32, _device="/job:localhost/replica:0/task:0/cpu:0"](MatMul_3, Reshape_5/shape)]]
```

jihunchoi:
Fixes #6824.
It checks `uncompacted_shape`, which contains the 'final' shape of `matmul` of two tensors, if it contains two or more `None` values.
If `uncompacted_shape` contains two or more `None` values, it means that it cannot be directly given to `reshape` as parameter.
This PR resolves this issue, by querying the dimension value using `tf.shape` function for that case.

### DIFF
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -340,6 +340,22 @@ def _einsum_reduction(t0, t0_axis_labels, t1, t1_axis_labels, axes_to_sum):
         t0_shape[:len(preserved_axes)+len(broadcast_axes[0])] +
         t1_shape[len(t1_shape)-len(broadcast_axes[1]):]
     )
+
+    # Check the number of None values and replace them with Tensors containing
+    # corresponding dimensions if there exist two or more None values
+    num_none_dims = sum(1 for d in uncompacted_shape if d is None)
+    if num_none_dims > 1:
+      uncompacted_shape = list(uncompacted_shape)
+      for i in xrange(len(uncompacted_shape)):
+        if uncompacted_shape[i] is None:
+          if i < len(preserved_axes) + len(broadcast_axes[0]):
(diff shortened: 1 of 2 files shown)


drpngx (approved):

drpngx:
Jenkins, test this please.

drpngx:
@jihunchoi that's consistent with numpy behavior right?

@aselle FYI

jihunchoi:
If you mean the test results, yes:
```
In [2]: m0 = np.array([[[1, 2]]])

In [3]: m1 = np.array([[3], [2]])

In [4]: np.einsum('ijk,kl->ijl', m0, m1)
Out[4]: array([[[7]]])
```

PR #6835, opened 2017-01-13
update Defun doc to clarify that definitions are frozen at first .run call

### LINKED ISSUE #6804: Defun broken: errors_impl.NotFoundError: Op type not registered
This code fails:

```
import tensorflow as tf
from tensorflow.python.framework import function

@function.Defun(tf.float32)
def custom_op(x):
    return x

x = tf.Variable(1, dtype=tf.float32)

sess = tf.Session()
print(sess.run(custom_op(x)))

sess.close()
```

Exception:

    tensorflow.python.framework.errors_impl.NotFoundError: Op type not registered 'custom_op_da39a3ee'

yaroslavvb:
Fixes #6804

### DIFF
--- a/tensorflow/python/framework/function.py
+++ b/tensorflow/python/framework/function.py
@@ -748,6 +748,9 @@ class Defun(object):
   default graph and adds the definition of the function into the
   default graph. Because the addition of the function into the graph
   is deferred, the decorator can be used anywhere in the program.
+  
+  Definitions of function is frozen for the session after first session.run
+  call, so you must create node using given function before it.
 
   Example, but also see the [How To on functions](link_needed).
 
--- a/tensorflow/python/framework/function.py
+++ b/tensorflow/python/framework/function.py
@@ -749,8 +749,9 @@ class Defun(object):
   default graph. Because the addition of the function into the graph
   is deferred, the decorator can be used anywhere in the program.
   
-  Definitions of function is frozen for the session after first session.run
-  call, so you must create node using given function before it.
+  Definitions of functions are frozen in a graph as soon as the graph is used to
+  create a session. Therefore, nodes using the function must be created in the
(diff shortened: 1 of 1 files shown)


drpngx (approved):

drpngx:
Jenkins, test this please.

tensorflow/python/framework/function.py
```diff
@@ -748,6 +748,9 @@ def foo(x, y):
   default graph and adds the definition of the function into the
   default graph. Because the addition of the function into the graph
   is deferred, the decorator can be used anywhere in the program.
+  
+  Definitions of function is frozen for the session after first session.run
```
zffchen78: How about

Definitions of functions are frozen in a graph as soon as the graph is used to create a session.
Therefore, nodes using the function must be created in the graph before the corresponding session
is created.  

Maybe then add an example.

zffchen (changes requested):
Thank you for the update, Yaralsav.

tensorflow/python/framework/function.py
yaroslavvb: That doesn't seem to match current behavior -- I can create a session, then create function node, and then issue first session.run and it works, still want me to change it to that?

drpngx:
We need to wait two weeks on that one, sorry about that.

rmlarsen:
@zffchen78 @drpngx Do you have additional comments? Can we merge this or are we still in the two week waiting period mentioned above?

yaroslavvb:
changed to the wording suggested by zf

rmlarsen:
Thanks @yaroslavvb. @zffchen78 can you take another look, please?

zffchen (approved):

drpngx:
Jenkins, test this please.

PR #6840, opened 2017-01-13
Set NDK_ROOT in android_full.sh to fix Makefile build

andrewharp:
tensorflow/contrib/makefile/build_all_android.sh disagrees about the environment variable to use. This seems the easiest fix with the least chance of breaking anything else.


andrewharp:
Didn't catch this when I tested the previous change locally because NDK_ROOT is set on my system, and the presubmits don't run android_full.sh. Can we have Jenkins make an exception and test it before or immediately after merging so that we know it works?

drpngx (approved):

PR #6843, opened 2017-01-13
Branch 144490912

drpngx:
Includes:

144470919 (compiler opts)
144485165, and 144478254 (parallel concat / parallel stack)
144490912 i/o namespace in C++ API


drpngx:
@petewarden `cc1plus: error: unrecognized argument in option '-march=native'` is that new?

drpngx:
@petewarden NM, I think it's 144470919.

gunan:
@martinwicke could you take a look?
@drpngx Would you like to fix internally first, or should we merge and then fix forward?

drpngx:
I prefer to fix internally, but this adds latency maybe an hour or two of latency. I'm fine working late, up to you. Let's see if we have other failures while we are thinking of a fix.

gunan:
however works best for you I can adjust, then let's close this, prepare a fix and prepare a new push?

drpngx:
OK, closing. Not clear yet what the fix is.

Issue #6762, opened 2017-01-10
configure assumes ldconfig can be called and that CuDNN is installed in a system library location


jowagner:
The configure script calls "ldconfig" to locate the CuDNN library but this makes 2 assumptions that can be wrong: 
1. It assumes that "ldconfig" can be called. However, on some GNU/Linux systems such as openSUSE Leap 42.1, only root has it in their path. (On Leap 42.1, normal users get the desired output from "/sbin/ldconfig -p".)
2. It assumes that the CuDNN library is installed in a system library location that is searched by ldconfig. However, users may install it in their home directory and use LD_LIBRARY_PATH to load it for various reasons. I, for example, was not fully satisfied from reading the licence terms that I am allowed to share the library with other users of the system and therefore did not install it system-wide as root.

Severity is low as normally the CuDNN library location should be auto-detected from the CuDNN library path before "ldconfig" is tried.

How to reproduce:
1. git clone this repo on a system with ldconfig not in your PATH
2. cd tensorflow and run ./configure
3. when asked for the CuDNN path, enter a valid and readable path (to avoid other errors), for example /usr/bin, that does not contain CuDNN
4. Error message: ./configure: line 337: ldconfig: command not found

Workaround options:
- First double check the CuDNN library path (the default path suggests that the last component "lib64" or "lib" should be omitted) and verify that standard folder names are used under this location for library and header files, for example "lib64" and "include".
- If you can call /sbin/ldconfig and if CuDNN is in a standard system library location the user can temporarily add /sbin to their PATH before running "./configure"
- Edit the "configure" script to provide the location manually, for example CUDNN_PATH_FROM_LDCONFIG="$HOME/local/cuda/lib64/libcudnn.so"

A fix should test whether "ldconfig" can be called before trying to call it. Also try /sbin/ldocnfig. Given that the library should normally be found already earlier in the code using the user-provided CuDNN path, a simple solution may be to set CUDNN_PATH_FROM_LDCONFIG to the empty string if ldconfig cannot be called.

If you like to automate the locating of the library as much as possible, you could furthermore check all locations in LD_LIBRARY_PATH.

jart:
Thanks for bringing this to our attention. It would be nice if the configure script was smarter about discovering this library. As for ldconfig, wouldn't it work if it was just changed to `/sbin/ldconfig`? Where else could it be?

[...]

PR #6848, opened 2017-01-14
temporarily add /sbin to PATH for ldconfig call

jowagner:
For issue #6762 (keeping the PATH changes local to where it is used instead of making it global as initially suggested).

### DIFF
--- a/configure
+++ b/configure
@@ -352,7 +352,12 @@ while true; do
   fi
 
   if [ "$OSNAME" == "Linux" ]; then
+    PATH_BEFORE_LDCONFIG="$PATH"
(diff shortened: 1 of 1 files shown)


jowagner:
This is a trivial code change addressing a minor problem. It does not have anywhere near the depth to be protected by copyright or requiring a licence to be copied or used.

If you insist on your CLA, I'm sorry I don't have time to review it at this time. You can easily implement an alternative solution such as the one paraphrased in issue #6762 .

Good luck!

drpngx:
CLA is required for all contributions. Feel free to reopen if this is solved.

PR #6851, opened 2017-01-14
Have `write_graph` return the output path of file

samjabrahams:
Since the function is performing `os.path.join()` for the user, it makes sense to return the location of the final file, or else the user will have to call `os.path.join()` again on their own.

### DIFF
--- a/tensorflow/python/framework/graph_io.py
+++ b/tensorflow/python/framework/graph_io.py
@@ -50,6 +50,9 @@ def write_graph(graph_or_graph_def, logdir, name, as_text=True):
       filesystems, such as Google Cloud Storage (GCS).
     name: Filename for the graph.
     as_text: If `True`, writes the graph as an ASCII proto.
+
+  Returns:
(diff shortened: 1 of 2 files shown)


drpngx:
Could we modify the test to check for that?

samjabrahams:
Done- I think I've added them in the right place.

samjabrahams:
Should I also assert `os.path.exists` is True?

drpngx:
Hopefully that'll work under windows. Yes, `os.path.exists` is probably a good idea. Once you push that, I'll trigger a test.

samjabrahams:
How's this look? Added in the path existence test, and changed `"/".join()` to `os.path.join()` (both in existing code and newly added code) for Windows compatibility.

drpngx:
Jenkins, test this please.

Looks good. It would be nice to specify if it's binary/text proto

drpngx:
NM the comment about text/binary, it's in the arguments.

PR #6853, opened 2017-01-14
Propagate seed in parallel_read to readers. Fixes #6735

### LINKED ISSUE #6735: slim parallel_read should pass seed to string_input_producer
Currently, the seed passed to `parallel_read` is only passed to the RandomShuffleQueue, but not to the string_input_producer. This should be fixed. Furthermore, it should also be documented that the output will never be deterministic if `num_readers` is greater than 1.

https://github.com/tensorflow/tensorflow/blob/e121667dc609de978a223c56ee906368d2c4ceef/tensorflow/contrib/slim/python/slim/data/parallel_reader.py#L212

yaroslavvb:


drpngx:
Jenkins, test this please.

drpngx (approved):
Thanks!

PR #6856, opened 2017-01-15
Remove unused variable in reading data examples

Andyccs:


Andyccs:
This is close because I commit with wrong email address

PR #6859, opened 2017-01-15
The type of `seq_len_max` should be int64

leomao:
If the `sequence_length` is not given, `max_seq_len` will be assigned to `time_len`.
By default, the type of `time_len` is `tf.int32`. But we need `tf.int64`.
So we need to cast it to `int64`, too. (Just like line 640)


drpngx (approved):

drpngx:
Jenkins, test this please.

PR #6868, opened 2017-01-16
Remove overly restrictive check on gradient types.

### LINKED ISSUE #6858: Support for mixed precision gradients
To save memory while keeping precision it's useful to be allow functions that store activations as fp16, but compute gradients in fp32. We are working on custom ops that compute gradients in mixed mode, but it requires hacks on Python gradient type checks side to working. The feature request is to relax these checks.

Here's an example computation graph with mixed precision gradient. Forward propagation is done in fp16, and backprop is in fp32 as in the graph below 
mixed-grads

Here's a toy example that creates such a g

Graph:

```
import tensorflow as tf
from tensorflow.python.framework import function

@function.Defun(tf.float16, tf.float32)
def custom_grad(x, grad):
    return 2*tf.cast(x, tf.float32)*grad

@function.Defun(tf.float16, grad_func=custom_grad)
def custom(x):
    return tf.square(x)

x = tf.Variable(1., dtype=tf.float16)

# override cast to keep first backprop in fp32 
with tf.get_default_graph().gradient_override_map({"Cast": "Identity"}):
    loss = tf.cast(custom(x, name="mycustom_apply"), tf.float32)
    
gradient = tf.gradients(loss, x)

sess = tf.Session()
sess.run(gradient)
```

Currently the example above crashes because various places in code assumes that grad function output matches type of incoming activation.

in particular:

- [tensorflow/python/ops/gradients_impl.py](https://github.com/tensorflow/tensorflow/blob/cb17d1b0e2b581b5da1d9597b7929ba764749d38/tensorflow/python/ops/gradients_impl.py#L260), line 264, in _VerifyGeneratedGradients: enforces that gradient output type is the same as activation input type

yaroslavvb:
Fixes #6858 

Currently gradient computation `input.is_compatible_with(backprop)` for every input. This fails if we have float16 activations and float32 backprops (ie `tf.float32.is_compatible_with(tf.float16)` returns False), this PR removes this check.

### DIFF
--- a/tensorflow/python/ops/gradients_impl.py
+++ b/tensorflow/python/ops/gradients_impl.py
@@ -225,12 +225,6 @@ def _DefaultGradYs(grad_ys, ys, colocate_gradients_with_ops):
         grad_ys[i] = array_ops.fill(
             array_ops.shape(y), constant_op.constant(
                 1, dtype=y.dtype))
-    else:
-      if grad_y.dtype != y.dtype:
-        raise ValueError("Y and ys_grad must be of the same type, "
-                         "not y: %s, ys_grad: %s " %
-                         (dtypes.as_dtype(y.dtype).name,
-                          dtypes.as_dtype(grad_y.dtype).name))
   return grad_ys
 
 
--- a/tensorflow/python/ops/gradients_impl.py
+++ b/tensorflow/python/ops/gradients_impl.py
@@ -241,7 +241,7 @@ def _IsTrainable(tensor):
 
 
 def _VerifyGeneratedGradients(grads, op):
-  """Verify that gradients are valid in number and type.
+  """Verify that gradients are valid in number.
 
   Args:
     grads: List of generated gradients.
@@ -253,15 +253,6 @@ def _VerifyGeneratedGradients(grads, op):
   if len(grads) != len(op.inputs):
     raise ValueError("Num gradients %d generated for op %s do not match num "
                      "inputs %d" % (len(grads), op.node_def, len(op.inputs)))
-  for i in xrange(len(grads)):
-    grad = grads[i]
-    inp = op.inputs[i]
-    if grad is not None:
-      if not grad.dtype.is_compatible_with(inp.dtype):
-        raise ValueError("Gradient type %s generated for op %s does "
-                         "not match input type %s" %
-                         (dtypes.as_dtype(grad.dtype).name, op.node_def,
-                          dtypes.as_dtype(inp.dtype).name))
 
 
 def _StopOps(from_ops, pending_count):
--- a/tensorflow/python/ops/gradients_impl.py
+++ b/tensorflow/python/ops/gradients_impl.py
@@ -225,6 +225,24 @@ def _DefaultGradYs(grad_ys, ys, colocate_gradients_with_ops):
         grad_ys[i] = array_ops.fill(
             array_ops.shape(y), constant_op.constant(
                 1, dtype=y.dtype))
+      continue
+    if y.dtype.is_real or y.dtype.is_integer:
+      if not grad_y.dtype.is_real and not grad_y.dtype.is_integer:
+          raise ValueError("Gradient type %s generated for real or "
+                           "integer-valued tensor %s with type %s must be "
+                           "real or integer" %
+                           (dtypes.as_dtype(grad_y.dtype).name, y,
+                            dtypes.as_dtype(y.dtype).name))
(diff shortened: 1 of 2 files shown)


drpngx:
Should we check that we have at least numeric types?

Jenkins, test this please.

yaroslavvb:
Makes sense, the backprops should be numeric (inputs could still be non-numeric)

yaroslavvb:
Added another place which enforces that gradient type matches activation type. With these changes we can run scott's mixed precision backprop kernels

girving:
It would also be nice to check that the gradient is complex iff the value is.  It might be safer to rephase the check using whitelists of similar ops to avoid missing cases like this.

yaroslavvb:
So checking if backprops are real-valued for real-valued inputs and complex valued for complex valued inputs is more restrictive without ruling out mixed precision. I added `is_real` method to mirror `is_complex`, since the existing `is_floating` doesn't seem to cover variables or quantized types. What did you mean by whitelists of similar ops? What whitelist should I add?

drpngx:
It's plausible to have quantized gradients for floating point parameters, I
think.

yaroslavvb:
added checks that real-valued inputs must have real-valued backprops, allowing for quantized representation, same restriction for complex-valued.

while testing, found that some tensors (TopKV2, TensorArrayReadV3) produce integer valued gradients, so relaxed checks to allow for integer values, PTAL

girving (approved):

girving:
Jenkins, test this please.

tensorflow/python/ops/gradients_impl.py
```diff
@@ -225,6 +225,24 @@ def _DefaultGradYs(grad_ys, ys, colocate_gradients_with_ops):
         grad_ys[i] = array_ops.fill(
             array_ops.shape(y), constant_op.constant(
                 1, dtype=y.dtype))
+      continue
+    if y.dtype.is_real or y.dtype.is_integer:
+      if not grad_y.dtype.is_real and not grad_y.dtype.is_integer:
+          raise ValueError("Gradient type %s generated for real or "
```
vrv: should these be TypeErrors ?

yaroslavvb: That does sound more logical, changed

girving (approved):

girving:
Jenkins, test this please.

girving:
Jenkins, test this please.

girving:
Jenkins, test this please.  Maybe this time it won't abort partway through?

gunan:
Tensorflow-sanity job is failing. Once it fails, it aborts all other jobs.

gunan:
```
FAIL: Found 2 non-whitelited pylint errors:
tensorflow/python/ops/gradients_impl.py:232: [W0311(bad-indentation), ] Bad indentation. Found 10 spaces, expected 8

tensorflow/python/ops/gradients_impl.py:239: [W0311(bad-indentation), ] Bad indentation. Found 10 spaces, expected 8
```

yaroslavvb:
Ooops, that's what I get for doing edits in Github edit window (bazel crashed our NFS so waiting for that to get fixed)

drpngx:
Jenkins, test this please.

martinwicke:
Ok, so we have an issue here. We may have to revert this as is, or maybe we can patch it.

@yaroslavvb, why did you introduce a `is_real`, instead of using the `is_floating` property? More specifically, `bfloat16` is really an internal type, is it necessary to expose it?

martinwicke:
Since there was discussion on this earlier, the bfloat16 quantization is (should be) an implementation detail for network transfer only (@josh11b, @vrv do you agree?). I don't understand the comment about Variables: in what way does is_floating not work for variables?

vrv:
bfloat16 was indeed supposed to be a hidden optimization.  Whether that should remain that way is up to others.

yaroslavvb:
@martinwicke I can send a PR that changes this to `is_floating`. The reason I added `is_real` is because future work may go a step further and use float32 activation, quantized backprop, which makes sense mathematically since both represent real-valued number

martinwicke:
That's true, but let's not get ahead of ourselves, especially not with something as hacky as bfloat16.  I doubt that we'd want that type in particular for quantized backprop. A proper fix-point or short float type (half) is much more likely. 

Can you make a PR changing this to use is_floating, and remove the is_real function? Mention me so we can get this in quickly.

Thanks!

PR #6872, opened 2017-01-16
Added possible responses for uname on Windows, fixes #6871

### LINKED ISSUE #6871: Windows `uname -s` might return different values
[`configure#L14`](https://github.com/tensorflow/tensorflow/blob/73bc42890/configure#L14) checks if the system is Windows, but doesn't check `uname -s` values correctly according to the [wiki page](https://en.wikipedia.org/wiki/Uname#Examples).

Unless not supported, I think other Windows SDE's should be included in the check.

z-a-f:
According to https://en.wikipedia.org/wiki/Uname#Examples, other responses
are possible on Windows, added mingw, cygwin, and uwin

### DIFF
--- a/configure
+++ b/configure
@@ -11,7 +11,7 @@ popd > /dev/null
 PLATFORM="$(uname -s | tr 'A-Z' 'a-z')"
 function is_windows() {
   # On windows, the shell script is actually running in msys
-  if [[ "${PLATFORM}" =~ msys_nt* ]]; then
+  if [[ "${PLATFORM}" =~ msys_nt*|mingw*|cygwin*|uwin* ]]; then
     true
   else
     false


drpngx:
Jenkins, test this please.

drpngx (approved):

drpngx:
If you put "Fixes #6871" in the commit description, the it'll auto-close the issue.

zafartahirov:
Oh, got it -- I thought it was sufficient to put it in the merge description -- will keep in mind next time, thanks!

zafartahirov:
Oh, I don't think I modified those files:

```
62 - C:/tf_jenkins/home/workspace/tensorflow-pr-win-cmake-py/tensorflow/python/kernel_tests/inplace_ops_test.py (Failed)
106 - C:/tf_jenkins/home/workspace/tensorflow-pr-win-cmake-py/tensorflow/python/kernel_tests/scatter_nd_ops_test.py (Failed)`
```

drpngx:
Yeah, it's an unrelated failure.

drpngx:
I wonder if it's transient.

Jenkins, test this please.

zafartahirov:
OK, I looked into the code, and I think there is a potential bug (I didn't introduce it though).

In line [`inplace_ops_test.py#L102`](https://github.com/tensorflow/tensorflow/blob/76d596002/tensorflow/python/kernel_tests/inplace_ops_test.py#L102), there is a `d0 / 10`, which potentially might return a float, while `numpy.random.choice` expects an integer. This is supposed to be `d0 // 10`, given that there is a `from __future__ import division`. The same story is on line [`inplace_ops_test.py#L122`](https://github.com/tensorflow/tensorflow/blob/76d596002/tensorflow/python/kernel_tests/inplace_ops_test.py#L122). I am not really sure about the other error -- need to read the code more 😄

drpngx:
Yeah, we're removing that test. Thanks for looking into it. Don't worry
about that.

zafartahirov:
Oh OK... 

Well, anyway, for future reference --  I think I found what causes the second error. In line [`scatter_nd_ops_test.py#L55](https://github.com/tensorflow/tensorflow/blob/76d596002/tensorflow/python/kernel_tests/scatter_nd_ops_test.py#L55), there is a `num_updates = indices.size / ixdim`. The problem is that, because the header of the file has `from __future__ import division`, this line will return a float. This in return will cause [`scatter_nd_ops_test.py#L61](https://github.com/tensorflow/tensorflow/blob/76d596002/tensorflow/python/kernel_tests/scatter_nd_ops_test.py#L61) to fail, because `numpy.reshape` expects an integer or tuple of ints.

Just thought it might be useful, because this bug will eventually resurface in other builds as well

gunan:
Jenkins, test this please.

caisq:

PR #6874, opened 2017-01-16
Support relative path head with no `./`.

vv1133:
`tf.gfile.Glob` will fail if the argument is a relative path with no `./` in the beginning in current version.


vv1133:
I signed it!

drpngx:
Why do we need this? This won't work on windows.

vv1133:
I have posted an issue [#6893](https://github.com/tensorflow/tensorflow/issues/6893). I closed this pull request.

PR #6881, opened 2017-01-16
Add custom operator code generator script

sjperkins:
This script uses jinja2 templates to generate the following:
* C++ Header file that defines the operator class, templated on Device.
* C++ Header file that defines the CPU implementation of the operator.
* C++ Source file with Shape Function, REGISTER_OP and  REGISTER_KERNEL_BUILDER constructs.
* Cuda Header that defines the GPU implementation of the operator,  including a CUDA kernel.
* Cuda Source file with GPU REGISTER_KERNEL_BUILDER's for the operator.
* python unit test case, which constructs random input data, and calls   the operator.
* Makefile for compiling the operator into a shared library, using g++   and nvcc.

Operator inputs, outputs, polymorphic type attributes, other attributes and documentation can be specified.

### DIFF
--- a/tensorflow/tools/op_generator/README.md
+++ b/tensorflow/tools/op_generator/README.md
@@ -1,0 +1,105 @@
+# Tensorflow Custom Operator Code Outline Generator
+
+Writing a tensorflow operator requires writing fair amounts of boilerplate C++ and CUDA code.
+This script generates code for the CPU and GPU version of a tensorflow operator.
+More specifically, given tensorflow `inputs`, `outputs` and `attribute`s, it generates:
+
+* C++ Header file that defines the operator class, templated on Device.
+* C++ Header file that defines the CPU implementation of the operator.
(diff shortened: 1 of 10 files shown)


sjperkins:
I felt the content in the README.md was too large to be included in the [Adding a New op howto](https://www.tensorflow.org/how_tos/adding_an_op/), but it may be appropriate to add a link to this tool in that file.

sjperkins:
I signed it!

drpngx:
Thanks! While it's nice, it'll take some effort to keep this in sync. I would suggest trying for contrib instead.

vrv:
Indeed, this is cool, but probably would be better as an external repo for now. GitHub doesn't provide great community tools for managing and supporting external code that isn't core to the codebase.  Can this love on your own repo on GitHub for now?

sjperkins:
Apologies for not getting back sooner -- other commitments required my attention

I've packaged this up at https://github.com/sjperkins/tfopgen.

Its also available on pypi -- `pip install tfopgen`

PR #6888, opened 2017-01-16
add Erlang distribution

Mistobaan:
This is in the family of the Gamma distributions but with a=2.0.

### DIFF
--- a/tensorflow/contrib/distributions/python/ops/erlang.py
+++ b/tensorflow/contrib/distributions/python/ops/erlang.py
@@ -1,0 +1,120 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


drpngx:
This looks like the exponential distribution? It's really a special case of Erlang.

Mistobaan:
Yes, is another special case of the Gamma distribution, but instead of alpha=1 is alpha=2.0. I am referencing the Machine Learning through a Probabilistic Perspective book.

drpngx:
What's the difference with [exp](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/exponential.py#L37) then?

Mistobaan:
On a second thought, the Gamma is a generalization of the Erlang, where instead of using the factorial it uses the Gamma function. The only advantage of adding the Erlang is to constrain the scale parameter to just integers and optimize the kernel to just compute the factorial instead of the gamma function.

PR #6909, opened 2017-01-17
Use local variable for match_input_once to avoid saving it in checkpoint

### LINKED ISSUE #6786: Variable "matching_filenames" @  tf.train.match_filenames_once() should be local?
This var holds the matched file names for the input stream pipeline.
Since it is added to the graph, when using the contractor of tf.train.Saver() it is saved
as part of the checkpoint state.
Hence loading saved models from checkpoints for inference, the training file names are loaded instead of the val / inference files.

Shouldn't this Var become a local variable? will it not be saved by default treated as local?

yaroslavvb:
Fixes #6786

### DIFF
--- a/tensorflow/python/training/input.py
+++ b/tensorflow/python/training/input.py
@@ -63,7 +63,8 @@ def match_filenames_once(pattern, name=None):
   """
   with ops.name_scope(name, "matching_filenames", [pattern]) as name:
     return variables.Variable(io_ops.matching_files(pattern), trainable=False,
-                              name=name, validate_shape=False)
+                              name=name, validate_shape=False,
+                              collections=[ops.GraphKeys.LOCAL_VARIABLES])
 
 
 def limit_epochs(tensor, num_epochs=None, name=None):


josh11b:
Are we certain that users don't want the additional reproducibilty of saving this list of filenames in the checkpoint?  If it isn't saved, you might run this again and get a different answer and be surprised that suddenly your training is behaving differently.

yaroslavvb:
In the original issue, the surprise was you run `tf.train.match_filenames_once` was run to generate list of testing filenames. Then pre-trained model was loaded from checkpoint, which overwrote this list with training filenames. 100% test set accuracy, woohoo!

Saved filenames are useful to restore while training a model, but are not useful to restore when loading this model for testing. It seems there needs to be some kind of distinction between model parameters and checkpointed things that are not model parameters.

josh11b:
Jenkins, test this please.

josh11b (approved):

yaroslavvb:
The failure is due to timeout
```
//tensorflow/contrib/layers:feature_column_ops_test                     TIMEOUT in 65.0s
  /var/lib/jenkins/workspace/tensorflow-pull-requests-cpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local-opt/testlogs/tensorflow/contrib/layers/feature_column_ops_test/test.log
```

yaroslavvb:
btw, unrelated timeout issue has been fixed

rmlarsen:

rmlarsen:
@yaroslavvb I'm afraid that this broke an internal test. I'll be reverting this in the next push, while I investigate.

yaroslavvb:
@rmlarsen my first guess would be that the test is trying to restore "matching_filenames" variable from checkpoint

rmlarsen:
@yaroslavvb thanks for the hint. Will take a look later.

PR #6921, opened 2017-01-18
Added Intel MKL graph optimization code and the ability to enable MKL when running configure.

mahmoud-abuzaina:
- Added Intel MKL graph optimization code.

- Modified configure file to build with MKL support.

- Included the MKL binary in Tensorflow wheel.


tensorflow/core/BUILD
```diff
@@ -1762,6 +1762,9 @@ tf_cc_tests(
         "graph/graph_test.cc",
         "graph/node_builder_test.cc",
         "graph/optimizer_cse_test.cc",
+# INTEL_MKL start
```
andydavis1: Can you change this to conditionally include this file, only if MKL is enabled? Example:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/BUILD#L796

mahmoud-abuzaina: I am not able to make that conditional through the select() function because there is a macro which iterates over all srcs and is not able to iterate over select():
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorflow.bzl#L329
it throws error: "type 'select' is not iterable".
Is there any way to add conditional value beside select()?

vrv: I think what you'll want to do is: 

srcs = [
  ... current list of files ..
  ...
] + if_mkl(["graph/optimizer_merge_test.cc"]),

mahmoud-abuzaina: yes this is what I did, but if_mkl() itself has a select function and it complains because of that:

def if_mkl(if_true, if_false = []):
    return select({
        "//third_party/mkl:using_mkl": if_true,
        "//conditions:default": if_false
    })

mahmoud-abuzaina: I can use if_mkl() in different places where is no for loop iteration, but not here.

vrv: Oh, got it.  In that case, you can try putting this in a standalone rule like the "cuda_libdevice_path_test" test below.

configure
```diff
@@ -49,6 +49,76 @@ while true; do
   # Retry
 done
 
+## Set up MKL related environment settings
+
+while [ "$TF_NEED_MKL" == "" ]; do
+	fromuser=""
+  read -p "Do you wish to build TensorFlow with MKL support? [y/N] " INPUT
```
andydavis1: Will this poll the user by default if they want to use MKL support?  I had thought that we wanted to leave this disabled until everything was submitted and verified to be working fine.

tensorflow/core/graph/mkl_optimizer_merge.h
```diff
@@ -0,0 +1,217 @@
...
+#include "tensorflow/core/common_runtime/optimization_registry.h"
+
+namespace tensorflow {
+
+// How many hops do we search for matching node in the backward dataflow graph?
+// We use maxhop of 10 based on empirical observations. Also, these are
+// maxhops in backward data-flow graph. Since input of forward nodes (Conv2D)
+// directly goes to backward nodes, we do not expect the hop-distance
+// would be more than few nodes.
+#define NODEMERGE_CONTEXT_MAXDEPTH 10
```
andydavis1: can this be a static int, which google style constant name "kNodeMergeContextMaxDepth".

tensorflow/core/graph/mkl_optimizer_merge.h
```diff
@@ -0,0 +1,217 @@
...
+    csinfo_.biasadd                    = "BiasAdd";
+    csinfo_.matmul                     = "MatMul";
+    csinfo_.matmulmkl                  = "MatMulMkl";
+    csinfo_.add                        = "Add";
+    csinfo_.biasaddgrad                = "BiasAddGrad";
+
+    minfo_.push_back({csinfo_.conv2d, csinfo_.biasadd, 0,
+                      csinfo_.conv2dwithbias});
+    // NOTE: We no longer use this optimization now.
+    // minfo_.push_back({csinfo_.matmul, csinfo_.add, 0,
```
andydavis1: Please remove code that has been commented out.

tensorflow/core/graph/mkl_optimizer_merge.h
```diff
@@ -0,0 +1,217 @@
...
+    //                csinfo_.matmulmkl});
+
+    // We use maxhop of 10 based on emperical observations. Also, these are
```
andydavis1: Can you explain why you need 'maxhop' at all?

tensorflow/core/graph/mkl_optimizer_merge.h
```diff
@@ -0,0 +1,217 @@
...
+
+/*
+ * Interface to invoke the pass for unit test
+ */
+extern bool OptimizeNodeMerge(std::unique_ptr<Graph>* g);
+
+/* We register merge optimizer for phase 1 and MKLToTF insertion
+ * for phase 2.
+ */
+REGISTER_OPTIMIZATION(OptimizationPassRegistry::PRE_PLACEMENT, 1,
```
andydavis1: Move this registration to the '.cc' file.

configure
mahmoud-abuzaina: right, is it ok to put this in "if false" block for now, or should I comment this code?

tensorflow/core/graph/mkl_optimizer_merge.cc
```diff
@@ -0,0 +1,561 @@
...
+  return result;
+}
+
+///////////////////////////////////////////////////////////////////////////////
+//              Run function for the pass
+///////////////////////////////////////////////////////////////////////////////
+
+bool OptimizeNodeMerge(std::unique_ptr<Graph>* g) {
+  // Get the ownership of the graph.
+  NodeMergeRewritePass *pass = new NodeMergeRewritePass();
```
andydavis1: Does this need to be allocated on the heap?

tensorflow/core/graph/mkl_optimizer_merge.h
```diff
@@ -0,0 +1,217 @@
...
+// Rewriting nodes in the graph:
+// This is neded in order to optimize gradient ops of
+// Conv2D+AddBias and MatMul+AddBias. Gradient op of
+// both the Conv2D and MatMul is BiasAddGrad, and we
+// need to rewrite BiasAddGrad into Conv2D-specific BiasAddGrad,
+// and MatMul-specific BiasAddGrad. This is context-specific
+// optimization, where the context is the forward operator
+// that the BiasAddGrad corresponds to.
+//
+class NodeMergeRewritePass : public GraphOptimizationPass {
```
andydavis1: At a high-level, its not clear to me that all of the code in this header file is needed elsewhere?  If that is not the case, could you move all of this code to the .cc file, and just make a thin wrapper class here in the .h file that inherits from GraphOptimizationPass?  In the .cc file, the thin wrapper classes Run method could create an instance of NodeMergeRewriter and invoke it (so all NodeMergeRewrite class definition could be in .cc file).

tensorflow/core/graph/mkl_optimizer_merge.h
nhasabni: Maxhop is to restrict search for the forward operator in the dataflow graph. Need for maxhop can be eliminated, but then it would search the complete graph till it finds forward operator. And this would increase runtime for the nodemerge optimization pass.

nhasabni: Maxhop is to restrict search for the forward operator in the dataflow graph. Need for maxhop can be eliminated, but then it would search the complete graph till it finds forward operator. And this would increase runtime for the nodemerge optimization pass.

configure
andydavis1: I think you want to make sure that you do not prompt the user until all your PRs are in and things are looking good.
I think its OK to leave in a switch for Intel developers to flip, but that switch should be off by default.

tensorflow/core/graph/mkl_optimizer_merge.h
andydavis1: Ok, so its not a correctness issue then, correct? i.e. it will always terminate, it just might take longer to terminate if you did not use maxhop?

I would re-work your comment about maxhop to start with the description, then add your comment about "we chose a maxhop of 10 based on emperical observations".

tensorflow/core/graph/mkl_optimizer_merge.cc
```diff
@@ -170,7 +306,7 @@ bool NodeMergeRewritePass::MergeNode(std::unique_ptr<Graph>* g,
 
     // 2. Get inputs from both the nodes.
     // Find the 2 inputs from the conv and the bias from the add Bias.
-    Node *oper1 = NULL, *oper2 = NULL, *oper3 = NULL;
+    Node *oper1 = nullptr, *oper2 = nullptr, *oper3 = nullptr;
```
andydavis1: Please put these on separate lines:
Node* oper1 = nullptr;
Node* oper2 = nullptr;
...

tensorflow/core/graph/mkl_optimizer_merge.cc
```diff
@@ -474,46 +521,46 @@ bool NodeMergeRewritePass::IsApplicableRewriteNode(const Node *n) const {
...
+  for (std::pair < Node*, Node* > i : nodes_to_be_merged) {
     // Even if MergeNode merges single pair of nodes, we
     // need to return true.
-    string n1name = i.first->name();
-    string n2name = i.second->name();
+    string n1_name = i.first->name();
+    string n2_name = i.second->name();
     if (MergeNode(g, i.first, i.second)) {
-      VLOG(1) << "Merged nodes " << n1name << " and " << n2name;
+      VLOG(1) << "Merged nodes " << n1_name << " and " << n2_name;
```
andydavis1: Maybe make the log level VLOG(1) for these statements that are in a loop?

tensorflow/core/graph/mkl_optimizer_merge.cc
```diff
@@ -36,6 +36,161 @@ limitations under the License.
...
+  // Standard interface to run optimization pass
+  Status Run(const GraphOptimizationPassOptions& options);
+
+  // Helper function which does most of heavy lifting for node merge
+  //
+  // Extracts common functionality between Run public interface and
+  // test interface.
+  //
+  // @return true, if and only if graph is mutated; false otherwise.
+  bool RunNodeMergeRewritePass(std::unique_ptr<Graph>* g);
```
andydavis1: nit-pick: I think this method name doesn't need to repeat the name of the class. Maybe just "Run" or "RunOnGraph".

tensorflow/core/graph/mkl_optimizer_merge.h
```diff
@@ -30,188 +30,13 @@ limitations under the License.
...
-   * @return Matching rewriteinfo in case a match is found; null otherwise.
-   */
-  const RewriteInfo* FindMatchingRewriteInfo(const Node *n,
-                                             const Node **fwdn) const;
-};
-
-/*
- * Interface to invoke the pass for unit test
- */
+// Returns true if and only if 'g' is mutated.
```
andydavis1: You could put a little wrapper class in the header and still leave most of the implementation in the .cc file. If you choose to keep things as they are, then please prefix the method below with Test_ if it is to be used only by unit tests.

tensorflow/core/graph/mkl_optimizer_merge.cc
```diff
@@ -117,18 +261,9 @@ Node* NodeMergeRewritePass::FindNodeForMerge(const Node* a) const {
...
-
-/*
-  Merge predecessor node with its successor.
-  Currently, we support merging Conv2D with AddBias, and
-  MatMul with Add node.
-
-  Returns true, if merging is successful and supported.
-  Returns false otherwise.
-*/
 bool NodeMergeRewritePass::MergeNode(std::unique_ptr<Graph>* g,
```
andydavis1: Can you change these functions to return a Status instead of a bool?

Then, you should change the TF_CHECK_OK calls below (and elsewhere) to return an error Status instead of doing a check fail.

Also, you should change any checks like the one below where you check "data_format_pred != data_format_succ" to return a descriptive error message, rather than just false.

tensorflow/core/graph/mkl_optimizer_merge.cc
nhasabni: Do you mean VLOG(2)?

tensorflow/core/graph/mkl_optimizer_merge.h
nhasabni: I did not find Test_if function. I found TEST_F. I have used TEST_F in mkl_optimizer_merge_test.cc, but not sure how to use it in mkl_optimizer_merge.h file.

On the other hand, I am using same interface as that is used in optimizer_cse.h file. I did not find any macros used in optimizer_cse.h. Yes, I want to leave .h and .cc files as they are now.

tensorflow/core/graph/mkl_optimizer_merge.h
nhasabni: Yes, it is not a correctness issue. OK, I will modify the comment.

tensorflow/tools/pip_package/BUILD
```diff
@@ -128,6 +128,6 @@ sh_binary(
             "//tensorflow/python/saved_model:all_files",
             "//tensorflow/python/tools:all_files",
             "//tensorflow/tensorboard",
-        ],
-    }),
+        ] 
+				}) + if_mkl(["//third_party/mkl:intel_binary_blob"])
```
vrv: there's a lot of weird whitespacing here that's going to fail our tests, can you try your best to fix these issues, or use buildifier to fix these whitespace problems?

tensorflow/tools/pip_package/build_pip_package.sh
```diff
@@ -89,6 +89,8 @@ function main() {
       bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/external \
       "${TMPDIR}/external"
     RUNFILES=bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles
+		cp -R \
+			bazel-bin/_solib_k8 "${TMPDIR}"
```
vrv: I don't know what this is doing -- this deserves a comment, and it's unclear why this is the right command to use -- should the path be more specific than just bazel-bin/_solib_k8 ?

configure
```diff
@@ -49,6 +49,72 @@ while true; do
   # Retry
 done
 
+## Set up MKL related environment settings
+if false; then # Disable building with MKL for now
```
vrv: please try to match the style of whitespacing as in the rest of this file

tensorflow/core/graph/mkl_optimizer_merge_test.cc
```diff
@@ -0,0 +1,398 @@
...
+#include "tensorflow/core/platform/test_benchmark.h"
+
+namespace tensorflow {
+namespace {
+
+static void InitGraph(const string& s, Graph* graph) {
+  GraphDef graph_def;
+
+  auto parser = protobuf::TextFormat::Parser();
+  //  parser.AllowRelaxedWhitespace(true);
```
vrv: (remove commented lines)

tensorflow/core/graph/mkl_optimizer_merge_test.cc
```diff
@@ -0,0 +1,398 @@
...
+  GraphConstructorOptions opts;
+  TF_CHECK_OK(ConvertGraphDefToGraph(opts, graph_def, graph));
+}
+
+class OptimizerMergeTest : public ::testing::Test {
+ public:
+  OptimizerMergeTest() : graph_(OpRegistry::Global()) {}
+
+  void InitGraph(const string& s) {
+    ::tensorflow::InitGraph(s, &graph_);
```
vrv: nit: this namespacing is unnecessary since we're in the tensorflow namespace already

vrv (commented):
Just a bunch of small nits, I'm letting Andy do the bigger ones :)

vrv:
style changes lgtm (need to resolve a conflict it looks like)

vivek-rane:
Merging is done. Awaiting review completion.

andydavis1 (approved):

andydavis1:

andydavis1:

rmlarsen:

rmlarsen:
@mahmoud-abuzaina please reformat the build file by running buildifier or apply the suggested fixes here: https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/2931/console

rmlarsen:

rmlarsen:

rmlarsen:

rmlarsen:

rmlarsen:
@mahmoud-abuzaina thanks for the contribution. Merging now.

PR #6936, opened 2017-01-18
Change combiner default for crossed_column

terrytangyuan:


caisq:
There are test failures. Can you please take a look, @terrytangyuan

martinwicke:
Sorry, it appears we got ahead of ourselves with that deprecation. It's not clear the new combiner is superior, and changing it does cause a lot of trouble for people with existing models. @ispirmustafa, this is fallout from 31439963. We should remove the deprecation notice.

terrytangyuan:
That's fine. Thanks for letting me know. Just randomly came across it and edited on Github.

PR #6938, opened 2017-01-18
Make XLA command-line option to mnist_softmax_xla.py actually work.

ndronen:
The --xla command-line argument to mnist_softmax_xla.py is effectively always true, so the example on https://www.tensorflow.org/versions/master/experimental/xla/jit (i.e. `python mnist_softmax_xla.py --xla=false`) still runs XLA.  This PR gives the script --xla and --no-xla optional arguments, with --xla as the default.

### DIFF
--- a/tensorflow/examples/tutorials/mnist/mnist_softmax_xla.py
+++ b/tensorflow/examples/tutorials/mnist/mnist_softmax_xla.py
@@ -102,6 +102,14 @@ if __name__ == '__main__':
       default='/tmp/tensorflow/mnist/input_data',
       help='Directory for storing input data')
   parser.add_argument(
-      '--xla', type=bool, default=True, help='Turn xla via JIT on')
+      '--no-xla',
+      dest='xla', action='store_false',
+      help='Turn xla via JIT off (default is on)')
+  parser.add_argument(
+      '--xla',
+      dest='xla', action='store_true',
+      help='Turn xla via JIT on (default is on)')
+  parser.set_defaults(xla=True)
+
   FLAGS, unparsed = parser.parse_known_args()
   tf.app.run(main=main, argv=[sys.argv[0]] + unparsed)


tensorflow/examples/tutorials/mnist/mnist_softmax_xla.py
```diff
@@ -102,6 +102,14 @@ def main(_):
       default='/tmp/tensorflow/mnist/input_data',
       help='Directory for storing input data')
   parser.add_argument(
-      '--xla', type=bool, default=True, help='Turn xla via JIT on')
+      '--no-xla',
+      dest='xla', action='store_false',
```
vrv: I've preferred this alternative version of specifying boolean flags in argparse:

https://github.com/tensorflow/tensorflow/blob/ec7929b878926c39255254e9aea992f0bc65aa68/tensorflow/python/debug/examples/debug_errors.py#L72

This allows you to do --xla=True or --xla and have it work.  What do you think?

ndronen: I'm OK with that. I'd just like this tutorial to work as documented. :-)

It might be a while before I can get a committers agreement signed, so anyone else is welcome to submit an equivalent PR.

vrv:
cc @hawkinsp in case he wants to fix this in another way

drpngx:
@ndronen did you sign the CLA? If so, please reply to the thread as advised. Thanks.

ndronen:
Hi, @drpngx, it's going to be a while before the legal department will get around to approving it.  Not sure if they're even going to act on it.  My apologies!

drpngx:
Got it, thanks for letting us know!

vrv:
I'm going to close this PR so we keep the number of active PRs low  -- feel free to send this to us again when you can sign the CLA -- we really appreciate it!

PR #6940, opened 2017-01-18
use context to cleanly close the file descriptor

Mistobaan:


caisq:

caisq (approved):

yaroslavvb:
By the way, the reason `open('a','w').write('x')` works is because `close` is run by `__del__`, which is called immediately on dereference

Issue #6914, opened 2017-01-17
Binaries should target CUDA architectures 6.0, 6.1 to avoid long delays on startup


benbarsdell:
The public docker images and pip packages are not compiled against CUDA architecture 6.0 or 6.1, which can cause a delay of more than a minute during initialization when executed on corresponding hardware (P100, GTX 10 series etc.) as the CUDA driver runs JIT compilation on all of the internal kernels.

Given that the public binaries are now built against CUDA 8.0, it would make sense to have them target archs 6.0 and 6.1.

Minimal reproducer (run with CUDA_CACHE_DISABLE=1):

```
import tensorflow as tf
sess = tf.Session()
rand_init = tf.random_uniform([32, 8], -1.0, 1.0)
rand_var  = tf.Variable(rand_init)
init = tf.global_variables_initializer()
sess.run(init)
```

### Related issues
https://github.com/tensorflow/tensorflow/issues/3651

### Environment info
Operating System: Linux

vrv:
Gunhan, how hard would it be to do for our pip releases ?

vrv:
@martinwicke

vrv:
cc @yifeif too

gunan:
Not that difficult, but this came up before and the result was to limit our pip packages to 2 cuda compute capabilities. 3.0, we wont drop.
if we do 6.1, we will drop 5.2
We will see a worse degradation for more people IMO once we do that.

vrv:
Let's chat more, I think we have to figure out a way to distribute (potentially large) binaries that works for more people.

@benbarsdell : what do other frameworks do?  Do they release binaries that contain the code for every possible compute capability under the sun too?

cliffwoolley:
My recommendation would be to include, minimally, 3.5, 5.2, 6.0, and 6.1.  Since you also want 3.0, then that makes five.

Compute capability 3.7 is ISA-identical to 3.5; only the on-chip resources increased, so it's only in specific circumstances that building 3.7 separately is worthwhile, and I'd say it's fair to skip it here.  Compute capability 5.0 was first-gen Maxwell (GM10x), and the only chips in that line were on the smaller end of the spectrum -- most people doing compute went straight from Kepler (3.x) to the 2nd-gen Maxwell (GM20x, aka compute capability 5.2).

The only other ones that exist that I've omitted are the ones for mobile GPUs.

vrv:
[...]

PR #6941, opened 2017-01-18
Add pre-compilation of Pascal binaries (CC 6.0/6.1)

mkolod:
See issue #6914.

Mentioning @vrv since he gave the go-ahead for this.

### DIFF
--- a/tensorflow/tools/docker/Dockerfile.devel-gpu
+++ b/tensorflow/tools/docker/Dockerfile.devel-gpu
@@ -89,7 +89,7 @@ WORKDIR /tensorflow
 ENV CI_BUILD_PYTHON python
 ENV LD_LIBRARY_PATH /usr/local/cuda/extras/CUPTI/lib64:$LD_LIBRARY_PATH
 ENV TF_NEED_CUDA 1
-ENV TF_CUDA_COMPUTE_CAPABILITIES=3.0,3.5,5.2
+ENV TF_CUDA_COMPUTE_CAPABILITIES=3.0,3.5,5.2,6.0,6.1
 
 RUN tensorflow/tools/ci_build/builds/configured GPU \
     bazel build -c opt --config=cuda tensorflow/tools/pip_package:build_pip_package && \


vrv (approved):
LGTM!

vrv:
@gunan let us know if this is okay, shouldn't need to run a bunch of tests...

mkolod:
@vrv @gunan This should be fine as long as you have CUDA 8.0 (or an equivalent version of Google-CUDA) available on the CI and Docker build server. Since we're not changing sass-based codes here (same CUBLAS and CUDNN versions), this should be fine, IMHO.

vrv:
@mkolod agreed, just want to make sure there aren't any unexpected configs we have to update first :)

mkolod:
@vrv @gunan Also, since this is a PR to master (rather than 0.12.1 that gcr.io/tensorflow/tensorflow:latest-devel-gpu is currently based on), I assume this won't affect containers until at least 1.0, unless an update is made to 0.12.1, which I assume is probably not advised at this point, but of course it's your decision :)

caisq:
There is another placs where capability versions are specified in the codebase:
third_party/tensorflow/contrib/cmake/CMakeLists.txt
Line 166

Is it necessary to change that as well?

vrv:
not in this change: the goal is to only fix the development gpu docker images.  expanding the releases to do more will involve changing a lot of other things.

gunan (approved):

gunan:
Change looks good to me, @caisq, any objections?

caisq:
LGTM

mkolod:
@vrv @caisq @gunan I was also wondering if this change to master could be cherry-picked for the final 1.0 release, or maybe even now. From what I understand, the concern is not the Pascal cubins in the Docker containers, but the pip package limitations. So, maybe it would be OK to add this change to r1.0 as well? I know of several public users who would appreciate that.

vrv:
It doesn't change the binary, just the docker at the branch, so it seems okay to me.  @martinwicke  wdyt?

mkolod:
@vrv Ah yes, you're right, [the GPU devel Dockerfile](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docker/Dockerfile.devel-gpu) does already point to r1.0 for Git checkout, so the only issue is to ship with r1.0, rather than merging the Dockerfile into r1.0. Of course I understand that r1.0 isn't yet shippable and so "latest" means r0.12.1.

mkolod:
@vrv @martinwicke For the r0.12 users though, would it be possible to also update the shippable compute capabilities so that users get the Pascal cubins in the nightlies even now?

vrv:
we're probably going to release 1.0 imminently, and then the default pip install / nightlies will use 1.0, so I'd rather just wait for that.

mkolod:
Sounds good.

mkolod:
@vrv Also, I think I have a solution to the pip package limitation problem. There is an nvcc flag (not sure about Google's CUDA compiler) `-Xfatbin -compress-all` which compresses the cubins for storage, but uncompresses at runtime. This would enable pip packages to ship with Pascal too, while using less storage. Unfortunately, while I know how to tweak gcc flags in Bazel, I'm not sure how to tweak the nvcc ones - otherwise I would have issued a PR for that as well.

mkolod:
@vrv @martinwicke @gunan @zheng-xq The nightly is still shipping without Pascal support, so this change hasn't been cherry-picked into the 1.0 branch, and hence the 1.0 gpu-devel Docker image. The way I tested it was via cuobjdump, but it's clear from the 1.0 branch as well (see [here](https://github.com/tensorflow/tensorflow/blob/r1.0/tensorflow/tools/docker/Dockerfile.devel-gpu#L92)).

```
$ nvidia-docker run --rm -it gcr.io/tensorflow/tensorflow:latest-devel-gpu /bin/bash
root@ee4171f66f39:~# cuobjdump /usr/local/lib/python2.7/dist-packages/tensorflow/python/_pywrap_tensorflow.so | grep sm_ | sort | uniq
arch = sm_30
arch = sm_35
arch = sm_52

```

PR #6944, opened 2017-01-18
Fix: Fix timing flakiness by mocking time object in

gunan:
basic_session_run_hooks_test.py


tensorflow/python/training/basic_session_run_hooks_test.py
```diff
@@ -19,6 +19,7 @@
 from __future__ import division
 from __future__ import print_function
 
+import mock
```
martinwicke: Use from ...platform import test, then use test.mock, which is an import that's py2/3 safe.

martinwicke (approved):

Issue #6950, opened 2017-01-19
Error downloading zlib 1.2.8 in TensorFlow r0.10


Nayana-ibm:
While building TensorFlow r0.10, we faced below error:
`zlib_archive: Error downloading http://zlib.net/zlib-1.2.8.tar.gz`

It looks like the issue has been fixed in r0.12 through [commit](https://github.com/tensorflow/tensorflow/commit/1e317b1f7dc5ccf04fc51ac96d97f5bdaefa9af9).

Will it be possible to include same change for r0.10?

jart:
Oh my goodness. It looks like I should just cherry pick https://github.com/tensorflow/tensorflow/commit/96e3e6f70ad9c441be46fcf70449e810a4b36896 on all the branches.

jart:
You're going to be so happy to hear that this is never going to happen again in the future, thanks to the incredible improvements we made to the Bazel downloader. Policy now dictates we have redundant mirrors for everything. It's going to do wonders for preserving the archival integrity of this software. https://github.com/bazelbuild/bazel/commit/ed7ced0018dc5c5ebd6fc8afc7158037ac1df00d

trsaunders:
It looks like redundant mirrors are needed for nasm:

```
ERROR: /home/tom/.cache/bazel/_bazel_tom/0eaf2b6b7a8618527eb96256b130e46f/external/jpeg/BUILD:162:1: no such package '@nasm//': Error downloading [http://www.nasm.us/pub/nasm/releasebuilds/2.12.02/nasm-2.12.02.tar.bz2] to /home/tom/.cache/bazel/_bazel_tom/0eaf2b6b7a8618527eb96256b130e46f/external/nasm/nasm-2.12.02.tar.bz2: Connection refused (Connection refused) and referenced by '@jpeg//:simd_x86_64_assemblage23'.
ERROR: /home/tom/.cache/bazel/_bazel_tom/0eaf2b6b7a8618527eb96256b130e46f/external/jpeg/BUILD:162:1: no such package '@nasm//': Error downloading [http://www.nasm.us/pub/nasm/releasebuilds/2.12.02/nasm-2.12.02.tar.bz2] to /home/tom/.cache/bazel/_bazel_tom/0eaf2b6b7a8618527eb96256b130e46f/external/nasm/nasm-2.12.02.tar.bz2: Connection refused (Connection refused) and referenced by '@jpeg//:simd_x86_64_assemblage23'.
[...]

Issue #6865, opened 2017-01-15
zlib removed from repository url mentioned in build


prats226:
I was building docker container for tensorflow r0.11 branch and it throws error because it repository url for package zlib mentioned in build file is no longer available on that url. Here is snippet for error.

```ERROR: /tensorflow/tensorflow/core/BUILD:903:1: no such package '@zlib_archive//': Error downloading 
from http://zlib.net/zlib-1.2.8.tar.gz to /root/.cache/bazel/_bazel_root/68a62076e91007a7908bc42a32e4
cff9/external/zlib_archive: Error downloading http://zlib.net/zlib-1.2.8.tar.gz to /root/.cache/bazel
/_bazel_root/68a62076e91007a7908bc42a32e4cff9/external/zlib_archive/zlib-1.2.8.tar.gz: 404 Not Found:
 <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">                                                  
<html><head>                                                                                         
<title>404 Not Found</title>                                                                         
</head><body>                                                                                        
<h1>Not Found</h1>                                                                                   
<p>The requested URL /zlib-1.2.8.tar.gz was not found on this server.</p>                            
<p>Additionally, a 404 Not Found                                                                     
error was encountered while trying to use an ErrorDocument to handle the request.</p>                
</body></html>                                                                                       
 and referenced by '//tensorflow/core:lib_internal'.                                                 
ERROR: Analysis of target '//tensorflow/tools/pip_package:build_pip_package' failed; build aborted.  
____Elapsed time: 3.516s                                                                             
```

gojira:
I just hit the same thing in r0.11 running ./configure

jart:
Thanks for bringing this to our attention. This was fixed in the r0.12 branch in https://github.com/tensorflow/tensorflow/commit/1e317b1f7dc5ccf04fc51ac96d97f5bdaefa9af9 but not the r0.11 branch, which I'll do now. In the meantime you can change the URL in tensorflow/workspace.bzl to http://zlib.net/fossils/zlib-1.2.8.tar.gz.

I'll also mention that at HEAD we now using redundant mirroring when downloading external dependencies. So problems like these should hopefully never arise again.

jart:
Should be resolved now.

gojira:
Thanks!

PR #6952, opened 2017-01-19
Update zlib URL in r0.9 branch

jart:
I don't know if anyone still uses 0.9 from source, but if they do, this PR is going to preemptively make them happy <('.'<)

See also: #6950 and #6865

### DIFF
--- a/tensorflow/workspace.bzl
+++ b/tensorflow/workspace.bzl
@@ -187,7 +187,7 @@ def tf_workspace(path_prefix = "", tf_repo_name = ""):
 
   native.new_http_archive(
     name = "zlib_archive",
-    url = "http://zlib.net/zlib-1.2.8.tar.gz",
+    url = "http://zlib.net/fossils/zlib-1.2.8.tar.gz",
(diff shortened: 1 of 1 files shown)


martinwicke (approved):

martinwicke:
Some of our tests didn't exist back then, so failures of e.g. Makefile are expected and harmless.

gunan:
I see this error in sanity build:
```
ERROR: no such package '@protobuf//': Error cloning repository: https://github.com/google/protobuf: cannot open git-upload-pack caused by https://github.com/google/protobuf: cannot open git-upload-pack caused by https://github.com/google/protobuf: cannot open git-upload-pack caused by https://github.com/google/protobuf: cannot open git-upload-pack caused by https://github.com/google/protobuf:
```

Are we ok with leaving this one in?
I think this change is OK. But if the build is in a broken state, it wont help, I think.

gunan (approved):

jart:
`git_repository` is evil and jgit is buggy, so maybe that's just a weird flake. Also I'm not sure if we officially maintain this branch anymore, so the way I see it, this PR makes it less obviously broken than before.

PR #6959, opened 2017-01-19
Passing config to MonitoredSession when using Estimator Class

jhaux:
While implementing my own Estimator class I noticed that although I supplied a RunConfig with `gpu_memory_fraction < 1`  all memory was allocated. After some investigation I think this is because the ConfigProto contained in the RunConfig is not passed to the MonitoredTrainingSession. 
Changing the argument in line 981 from `config=None` to `config=self.config.tf_config` did the trick for me.


martinwicke (approved):

martinwicke:
Jenkins, test this please.

gunan:
Please in the future merge all such fixes to the master branch first, then merge to release branch.

PR #6960, opened 2017-01-19
Improved support for libxsmm

benoitsteiner:


tensorflow/workspace.bzl
```diff
@@ -76,7 +76,6 @@ def tf_workspace(path_prefix = "", tf_repo_name = ""):
   native.new_http_archive(
       name = "libxsmm_archive",
       urls = [
-          # "http://bazel-mirror.storage.googleapis.com/github.com/hfp/libxsmm/archive/1.6.1.tar.gz",
```
gunan: Let's also add the bazel-mirror one. The link should look like:
http://bazel-mirror.storage.googleapis.com/github.com/hfp/libxsmm/archive/1.6.5.tar.gz

@jart once we make this change, is there a mechanism to automatically mirror the file under bazel-mirror?

jart: There's no consequences to having this here and the URL not being active, aside from 500ms extra latency, which is amortized out by the eight downloader threads.

My process right now is I run that script you reviewed every few days or so, and whatever GCS URLs exist in the file that aren't currently mirrored, get mirrored.

jart: Now mirrored: http://bazel-mirror.storage.googleapis.com/github.com/hfp/libxsmm/archive/1.6.5.tar.gz

tensorflow/core/kernels/BUILD
```diff
@@ -2128,7 +2128,10 @@ tf_kernel_library(
 tf_kernel_library(
     name = "matmul_op",
     defines = select({
```
jart: Is there any reason why we wouldn't want to always enable libxsmm for CPUs that support it?

tensorflow/workspace.bzl
benoitsteiner: Thanks for updating the mirror. It active again.

tensorflow/core/kernels/BUILD
benoitsteiner: We want to do a little more testing before turning it on by default. I'm hoping we'll be able to have it enabled in time for the 1.1 release.

gunan (approved):

jart (approved):

PR #4051, opened 2016-08-26
549 unsorted segment max

nikste:
added unsorted segment max. Fixes #549.

PR #6975, opened 2017-01-20
TF-549 Adds unsorted segment max Op

### LINKED ISSUE #549: Add a function unsorted_segment_max
We have unsorted_segment_sum and segment_max , so it seems natural to also provide unsorted_segment_max, which I happen to need.

nikste:
Adds unsorted segment max. Fixes #549, Continuation of #4051.
@drpngx, rebased to current master and squashed the commits.
Hope we'll manage this time!

### DIFF
--- a/tensorflow/core/kernels/segment_reduction_ops.cc
+++ b/tensorflow/core/kernels/segment_reduction_ops.cc
@@ -281,7 +281,7 @@ class UnsortedSegmentBaseOp : public OpKernel {
   explicit UnsortedSegmentBaseOp(
       OpKernelConstruction* context,
       functor::UnsortedSegmentBaseFunctor<Device, T, Index>& functor)
-      : OpKernel(context), reductionFunctor(functor) {}
+      : OpKernel(context), reduction_functor_(functor) {}
 
   void Compute(OpKernelContext* context) override {
     const Tensor& data = context->input(0);
@@ -317,11 +317,12 @@ class UnsortedSegmentBaseOp : public OpKernel {
     auto output_flat = output->flat_outer_dims<T>();
 
     auto data_ptr = data.template flat<T>().data();
-    reductionFunctor(context, context->template eigen_device<Device>(),
+    reduction_functor_(context, context->template eigen_device<Device>(),
                      output_rows, segment_ids.shape(), segment_flat,
                      data.NumElements(), data_ptr, output_flat);
   }
-  functor::UnsortedSegmentBaseFunctor<Device, T, Index>& reductionFunctor;
+ private:
+  functor::UnsortedSegmentBaseFunctor<Device, T, Index>& reduction_functor_;
 };
 
 template <typename Device, class T, class Index>
@@ -330,8 +331,9 @@ class UnsortedSegmentSumOp : public UnsortedSegmentBaseOp<Device, T, Index> {
   explicit UnsortedSegmentSumOp(OpKernelConstruction* context)
       : UnsortedSegmentBaseOp<Device, T, Index>(
             context,
-            sumFunctor) {}
-    functor::UnsortedSegmentSumFunctor<Device, T, Index> sumFunctor;
+            sum_functor_) {}
+ private:
+    functor::UnsortedSegmentSumFunctor<Device, T, Index> sum_functor_;
 };
 
 template <typename Device, class T, class Index>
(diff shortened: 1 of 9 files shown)


caisq:

nikste:
Fixed the Unit-tests, they should run through now, does jenkins need to be retriggered?

caisq:

tensorflow/core/kernels/segment_reduction_ops.cc
```diff
@@ -288,27 +317,67 @@ class UnsortedSegmentSumOp : public OpKernel {
...
+};
+
+template <typename Device, class T, class Index>
+class UnsortedSegmentMaxOp : public UnsortedSegmentBaseOp<Device, T, Index> {
+ public:
+  explicit UnsortedSegmentMaxOp(OpKernelConstruction* context)
+      : UnsortedSegmentBaseOp<Device, T, Index>(
+            context,
+            maxFunctor) {}
+    functor::UnsortedSegmentMaxFunctor<Device, T, Index> maxFunctor;
```
andydavis1: Can you make these private members and following the naming convention: 'max_functor_'.  Same for 'sumFunctor' and 'reductionFunctor' above?

tensorflow/core/kernels/segment_reduction_ops.cc
```diff
@@ -243,16 +244,44 @@ struct UnsortedSegmentSumFunctor<CPUDevice, T, Index> {
...
+// UnsortedSegmentMaxFunctor implementation for CPUDevice.
+template <typename T, typename Index>
+struct UnsortedSegmentMaxFunctor<CPUDevice, T, Index>
+    : UnsortedSegmentBaseFunctor<CPUDevice, T, Index> {
+  void operator()(OpKernelContext* ctx, const CPUDevice& d,
+                  const Index output_rows, const TensorShape& segment_ids_shape,
+                  typename TTypes<Index>::ConstFlat segment_ids,
+                  const Index data_size, const T* data,
+                  typename TTypes<T, 2>::Tensor output) override {
+    output.setConstant(std::numeric_limits<T>::min());
```
andydavis1: This doesn't share any code with UnsortedSegmentSumFunctor, yet there is lots of code copied between the two (I think the only differences are the initializer on line 256 and the output update on line 268).

Is it possible to have the base UnsortedSegmentBaseFunctor contain this shared code, and pass in "initializer" and "reducer" functors into the base implementation which can be called at these two call sites?

nikste: So I've been looking at this problem for a while now and I don't really see any straight forward solution to this.
One thing that makes it tricky is the GPU implementation of UnsortedSegmentSum, which does not share the code of the CPU implementation.
So that would be different from baseFunctor, but would need the same signature, since its used as well in BaseOp.
Thus would contain the init functors and reduction functors but would not use them.

Second I can't seem to initialize a functor as a member in e.g. UnsortedSegmentSumOp with member init functors and reduction functors as arguments.
If you see another way I'd be happy if you let me know.

nikste:
Changed the names of member functors now

tensorflow/core/kernels/segment_reduction_ops.cc
andydavis1: I wouldn't expect the GPU and CPU to share implementations, I would expect the CPU implementations of UnsortedSegmentSum/Max to (possibly) share code. Maybe this could be captured as a TODO? (I realize this PR has been pending for quite some time).

andydavis1 (approved):

rmlarsen:

vrv:
@rmlarsen there are still pending changes to be made

rmlarsen:
@vrv OK, thanks for clarifying.

nikste:
Addressed the second concern now

tensorflow/core/kernels/segment_reduction_ops.cc
nikste: Yes, I've understood the concern :) I don't see a straight forward solution to it, though.
Added a todo comment now, however.

rmlarsen:
@vrv is this good to go?

vrv:
@andydavis1 I assume this is okay now?

andydavis1:
yes, changes were approved.

rmlarsen:

rmlarsen:

rmlarsen:
The test failures are unrelated, I believe. Merging.

vrv:
Thank you for your incredible patience @nikste !! Glad to have finally got this in :)

mehmetbasbug:
Are there any plans for the GPU implementation? @nikste

PR #6987, opened 2017-01-20
enable hdfs support for windows

guschmue:


caisq:

mrry:
Thanks for adding this, Guenther! The change looks good to me, assuming the tests pass.

I assume none of our automated tests use this because of the Hadoop distribution dependency. Have you confirmed that it works manually as expected?

mrry (approved):

guschmue:
yes, I tested this against hadoop 2.6.4 running on a linux box. 
Tried to read files for training as well as save() / restore() - worked all ok (actually cool that gfile works).
The biggest hurdle is that you need the hadoop binaries on the windows side and there is no good binary distribution.

tensorflow/core/platform/hadoop/hadoop_file_system.cc
```diff
@@ -27,6 +27,12 @@ limitations under the License.
 #include "tensorflow/core/platform/posix/error.h"
 #include "third_party/hadoop/hdfs.h"
 
+#if defined(PLATFORM_WINDOWS)
+#define LIBHDFS_DSO "hdfs.dll"
```
jhseu: Nit: mind changing these to `const char* kLibHdfsDso = "..."`?

jhseu (changes requested):

tensorflow/core/platform/hadoop/hadoop_file_system.cc
guschmue: sure, will do

jhseu (approved):

caisq:

caisq:
The test timeout in "Linux CPU Tests (Python 3)" seems to be an unrelated flake. Merging the PR now.

caisq:
PR merged. Thanks, @guschmue and reviewers.

snnn:
Some notes for windows users:
1. jvm.dll need be in PATH dirs.
set PATH=%PATH%;%JAVA_HOME%\jre\bin\server

2.  Wildcards in Classpath do not work. This is a limitation of JNI. Please set classpath in this way:
hadoop classpath --jar c.jar
set CLASSPATH=c.jar

guschmue:
I think this is very helpful for people trying to setup hdfs on windows. Let me see if I can add this to the hdfs howto.

snnn:
@guschmue Sure. Please do it.

PR #7003, opened 2017-01-22
Fix breakages in Python 3.5 tests by using int type in indices

caisq:
Previously, float and np.float64 types were used in these places,
causing a certain version of Python 3.5 + NumPy to error out.


caisq:
Also TESTED: Experimental Linux CPU Python3.5 PIP build (Jenkins login required to view log):
http://ci.tensorflow.org/view/Experimental/job/experimental-cais-linux-python35-pip/1/console

gunan (approved):

PR #7021, opened 2017-01-23
Remove errorneous squeeze in `tf.contrib.losses.sparse_softmax_cross_entropy_loss`

### LINKED ISSUE #6846: tf.contrib.losses.sparse_softmax_cross_entropy not normal
When I tried to use tf.contrib.losses.sparse_softmax_cross_entropy with weights for losses, I got the error:

ValueError: weight.get_shape().ndims cannot be None

Later, I used tf.contrib.losses.softmax_cross_entropy and onehot labels,  everything was ok.  Is there anything wrong with tf.contrib.losses.sparse_softmax_cross_entropy?

Thanks a lot!

foxik:
The `tf.contrib.losses.sparse_softmax_cross_entropy_loss` has an weights parameter which can be used to weight the individual batch elements. The weights parameter can have various shape, which are all taken care of in `compute_weighted_loss`.

However, the `sparse_softmax_cross_entropy_loss` methods contains an errorneous squeeze of the weights. If some dimension of weights is not known, the number of dimensions after the squeeze is unknown, which causes the `compute_weighted_loss` method to throw an exception. However, if the squeeze in `sparse_softmax_cross_entropy_loss` is not called, `compute_weighted_loss` can deal even with unknown dimensions of weights.

Note that `tf.contrib.losses.softmax_cross_entropy_loss` does not contain the squeeze of the weights, even if it uses the weights argument in an equal way.

Should fix #6846.

### DIFF
--- a/tensorflow/contrib/losses/python/losses/loss_ops.py
+++ b/tensorflow/contrib/losses/python/losses/loss_ops.py
@@ -421,7 +421,6 @@ def sparse_softmax_cross_entropy(logits, labels, weights=1.0, scope=None):
   with ops.name_scope(scope, "sparse_softmax_cross_entropy_loss",
                       [logits, labels, weights]) as scope:
     labels = array_ops.reshape(labels, shape=[array_ops.shape(labels)[0]])
-    weights = array_ops.squeeze(weights)
 
     losses = nn.sparse_softmax_cross_entropy_with_logits(labels=labels,
                                                          logits=logits,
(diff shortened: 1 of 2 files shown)


drpngx:
Could we add a test for that perhaps?

rmlarsen:
@foxik could you add a test to your PR, please?

foxik:
I implemented the tests, they fail without the fix applied.

I added the tests also for `contrib.losses.softmax_cross_entropy` -- they are not strictly needed there (there was not the same error in `softmax_cross_entropy`), but the API offers the same guarantees for the `weights` in both calls.

rmlarsen:

rmlarsen (approved):

rmlarsen:
Thanks for the fix.

PR #7022, opened 2017-01-23
Remove concat_v2 from r1.0 branch

annarev:
Removing concat_v2 and references. tf.concat now has the same signature as tf.concat_v2. So, tf.concat should be used instead of tf.concat_v2.

### DIFF
--- a/tensorflow/g3doc/api_docs/python/array_ops.md
+++ b/tensorflow/g3doc/api_docs/python/array_ops.md
@@ -3135,8 +3135,6 @@ Compute gradients for a FakeQuantWithMinMaxVarsPerChannel operation.
 ## Other Functions and Classes
 - - -
 
-### `tf.concat_v2(values, axis, name='concat_v2')` {#concat_v2}
-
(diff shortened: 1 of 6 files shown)


martinwicke:
This is a planned and approved API change that should happen pre 1.0.

tensorflow/g3doc/api_docs/python/array_ops.md
```diff
@@ -3135,8 +3135,6 @@ Compute gradients for a FakeQuantWithMinMaxVarsPerChannel operation.
 ## Other Functions and Classes
 - - -
 
-### `tf.concat_v2(values, axis, name='concat_v2')` {#concat_v2}
```
gunan: This is an autogenerated file.
Please revert, I will rerun doc generation after cherrypicks.

annarev: I reverted that file now.

gunan (approved):

PR #7040, opened 2017-01-24
Branch 145423129

gunan:


yifeif:
Pushing again with more fixes.

PR #7073, opened 2017-01-25
FreeBSD compatibility

blodan:
Two commits

One with non breaking changes which should be able to be applied straight away
One with breaking changes as FreeBSD has libdl integrated into libc (so the -ldl linker option should probably be handled in another way but I have no idea how)


blodan:
I've signed the CLA.

blodan:
I've added the e-mail I commited with as alternate e-mail and verified it for my account as described in the contact info section so the CLA should be good now.

hholst80:
@blodan also GPU will work with this patch?

gunan:
By breaking change, do you mean breaking for existing configurations?
Instead of those, we can conditionally set flags with the help of such macros:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorflow.bzl#L79

blodan:
@hholst80 i have not tried GPU mode, i know there is a binary nvidia driver for freebsd, but i don't know if the cudatoolkit is available on bsd

blodan:
@gunan yes, as the "breaking change" commit removes the "-ldl" link option everywhere compilation would probably break for linux

How would i go about writing such a macro for this line for example?
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/BUILD#L1171

If you could give me an example for that one I can fix it for the rest of the -ldl places.

gunan:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorflow.bzl#L94
Here is how most of our copts should be set. As you see, based on different OS and toolchains, we select different options.
The config settings that pick what to build are defined here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/BUILD#L47

We probably need a new config setting for freebsd.
@damienmg @meteorcloudy How can we setup a new config setting for freeBSD in tensorflow BUILD files?

gunan:
OK, after reading a bit more:
once we add a new config value there, then we will need to set it from the command line when compiling for freeBSD.
So let's do this:

in tensorflow/BUILD, we need a new config setting.
```
config_setting(
  name = "freeBSD",
  values = {
    "define": "os=freeBSD"
  },
)
```
Then in this file https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorflow.bzl#L94
We can conditionally unset the linker flags.

Then, when running bazel, you will need to add the flag `--define os=freeBSD`. 

@blodan Could you try the above out?

tensorflow/core/BUILD
```diff
@@ -1150,7 +1150,7 @@ cc_library(
     ],
     copts = tf_copts(),
     defines = tf_additional_lib_defines(),
-    linkopts = ["-ldl"],
```
jart: Would it be possible to have a config_setting() rule and a select() function to selectively apply this change to only freebsd? This might work, but I'm not sure.

```python
config_setting(
    name = "freebsd",
    values = {"cpu": "freebsd"},
)
```

See also: https://github.com/tensorflow/tensorflow/blob/master/third_party/jpeg/jpeg.BUILD#L463

jart (changes requested):

third_party/gpus/cuda_configure.bzl
```diff
@@ -374,6 +374,13 @@ def _lib_name(lib, cpu_value, version="", static=False):
       if version:
         version = ".%s" % version
       return "lib%s.so%s" % (lib, version)
+  elif cpu_value == "FreeBSD":
```
jart: This appears to be the same as the previous clause. Couldn't you say `cpu_value in ("Linux", "FreeBSD")`?

tensorflow/stream_executor/BUILD
```diff
@@ -25,7 +25,6 @@ cc_library(
         "platform/**/*.h",
     ]),
     linkopts = [
-        "-ldl",
```
jart: Have you investigated why this was put here in the first place? We want to preserve support for other platforms.

jart (changes requested):

blodan:
@gunan @jart I'm trying to get this to work with your pointers but it doens't pick up my settings I'm trying to implement.

First of I've defined 
`config_setting(
    name = "FreeBSD",
    values = {"cpu": "FreeBSD"},
)`
in tensorflow/BULD

Then I tried this on line 1153 in tensorflow/core/BUILD:
`linkopts = select({
        "//tensorflow:FreeBSD": [],
        "//conditions:default": ["-ldl"],
})`

However, when trying to build it still adds -ldl to the compile options, should the above "clear" the settings for FreeBSD?

blodan:
Figured out that 
> values = {"cpu": "freebsd"},

had to be in lowercase, just going to build this on a fresh freebsd system to see that it actually works before commiting the changes.

blodan:
Okay so I've commited the changes now and this should be good to go if you don't want me to do any other changes @gunan @jart

rmlarsen:

damienmg:
Sorry I was OOO. /cc @aehlig who made Bazel FreeBSD port.

I had only one comment but I see that @jart already gave the same :)

rmlarsen:

rmlarsen:
@blodan it looks like the build files are still not quite there. Can you try to fix the problem?

blodan:
@rmlarsen from reading the output of jenkins it seems there was a failure during download/fetching some kind of repository? I don't think thats related to what I've changed.

https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/2936/console

Or am I looking at the wrong output?

gunan:
Jenkins, test this please.

gunan:
https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/2967/console

It failed again.
Bazel is not great at rreporting errors, but this usually means there is an error in BUILD of .bzl files.
Could you test it out locally (or on docker)? Simply running `configure` should trigger the issue.

damienmg:
Hi, I am interested to know what generate that stacktrace, Bazel should definitely never print a stack trace (means a bug to us)

damienmg:
Which version of Bazel TensorFlow CI is using? 0.4.3? I  cannot repro the stack trace. However the error message I get is not really helpful either, it tries to load @local_config_cuda//cuda:cudart_static but cannot find @local_config_cuda//cuda build file. Looking at `$(bazel help output_base)/external/local_config_cuda/cuda` I can find a build file...

third_party/gpus/cuda/BUILD.tpl
```diff
@@ -49,8 +49,10 @@ cc_library(
     name = "cudart_static",
     srcs = ["lib/%{cudart_static_lib}"],
     includes = ["include/"],
-    linkopts = [
-        "-ldl",
+    linkopts = select({
+            "//tensorflow:FreeBSD": [],
```
damienmg: This won't find //tensorflow:FreeBSD...

Either the TF team decide to switch to bazel 0.4.4 and you can just add @org_tensorflow in front of the label or you have to work around either by injecting the workspace name from the template generation.

damienmg (changes requested):
This is the only issue on my side.

damienmg:
Ok the problem is that it is looking for `@local_config_cuda//tensorflow:FreeBSD` which does not exists. The error message I get with Bazel 0.4.4 is definitely misleading:

```
.........
ERROR: /usr/local/google/home/dmarting/.cache/bazel/_bazel_dmarting/e5cce820cc082410b4fcc604db349066/external/local_config_cuda/cuda/BUILD:48:1: no such package '@local_config_cuda//tensorflow': BUILD file not found on package path and referenced by '@local_config_cuda//cuda:cudart_static'
ERROR: /usr/local/google/home/dmarting/.cache/bazel/_bazel_dmarting/e5cce820cc082410b4fcc604db349066/external/local_config_cuda/cuda/BUILD:48:1: no such package '@local_config_cuda//tensorflow': BUILD file not found on package path and referenced by '@local_config_cuda//cuda:cudart_static'
ERROR: Evaluation of query "deps((//tensorflow/... - //tensorflow/examples/android/...))" failed: errors were encountered while computing transitive closure
```

The stacktrace is probably coming from one of those case we fixed in recent release.

blodan:
Thanks for the more descriptive error @damienmg, that saved me a lot of time.

I've now changed to a local config for the cuda BUILD.tpl as you seem to have done with darwin previously, it should build fine in jenkins now.

rmlarsen:

rmlarsen:
@blodan it appears there are still formatting errors in the build files : 

1310,1313c1310,1313
<             "//tensorflow:FreeBSD": [],
<             "//conditions:default": ["-ldl"],
<         }) + [
<         "-lm"
---
>         "//tensorflow:FreeBSD": [],
>         "//conditions:default": ["-ldl"],
>     }) + [
>         "-lm",

blodan:
I've now removed one indentation as specified, hopefully it goes through now! :)

gunan:
Jenkins, test this please.

tensorflow/BUILD
```diff
@@ -86,6 +86,12 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+config_setting(
+    name = "FreeBSD",
```
jart: Nit: Can we make this lowercase like the others?

jart (changes requested):

blodan:
I've now changed the config option to lowercase and fixed the last buildifer warning

gunan:
Jenkins, test this please.

gunan:
Looks like all tests are passing.
@jart @damienmg could you take another look?

damienmg (approved):

vrv:
@jart does this all look good to you?

jart (approved):

PR #7078, opened 2017-01-25
Replace all gate_gradients=1 with gate_gradients=GATE_OP in documenta…

slacy:
Replace all "gate_gradients=1" in documentation with "gate_gradients=GATE_OP" to make 
it very clear that GATE_OP is the default value. This matches what exists in the source code as well.

### DIFF
--- a/tensorflow/g3doc/api_docs/python/contrib.opt.md
+++ b/tensorflow/g3doc/api_docs/python/contrib.opt.md
@@ -103,7 +103,7 @@ Construct a new MovingAverageOptimizer.
 
 - - -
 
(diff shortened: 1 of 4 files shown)


slacy:
I signed it!

slacy:
Is there something that I need to do to run the required checks above, or will they eventually run automatically?

rmlarsen (approved):

rmlarsen (changes requested):
The .md files are auto-generated from the docstring in the Python code, so this change would be quickly overwritten.

rmlarsen:
@slacy I'm not sure there is a way to preserve "GATE_OP" when auto-generating the docs. Assigning @xmbrst for advice.

vrv:
All the docs changed in the 1.0 release -- going to close this -- it's possible this got fixed with the new doc generators.  In any case, editing the .md files wouldn't be the right thing to do anyway :(

PR #7081, opened 2017-01-25
Updates to install py3.5 packages. No more need for python 3.4, and new

gunan:
dependency, werkzeug.


caisq:
Do the mac and windows scripts need to updated accordingly as well?

gunan:
They have to be updated internally.

caisq (approved):

Issue #6700, opened 2017-01-06
tensorflow docs have sigmoid_cross_entropy_with_logits args flipped


cpdurham:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/api_docs/python/functions_and_classes/shard5/tf.nn.sigmoid_cross_entropy_with_logits.md 

labels comes before logits, please excuse if i'm way off base, not sure where these docs are generated from

yaroslavvb:
Arg order change was introduced in https://github.com/tensorflow/tensorflow/commit/333dc32ff79af21484695157f3d141dc776f7c02 by @martinwicke ... I guess docs need to be regenerated?

cpdurham:
wow, isn't this going to break upgraded code in very hard to debug ways?

yaroslavvb:
This is one of a number of "hard to debug breaking changes" coming up in preparation to 1.0 release, documented [here](https://github.com/tensorflow/tensorflow/blob/c72d891418c326794a913cc13382c104255fe1f2/RELEASE.md#breaking-changes-to-the-api) . The recommendation for backward compatibility is to not use any positional arguments but specify keywords explicitly

martinwicke:
And in fact, the new interface disallows positional arguments for that
exact reason.

ppwwyyxx:
Some other function, e.g. sparse_softmax_cross_entropy_with_logits also have this breaking change. These need to be documented in RELEASE.md.

But in one function: [tf.nn.weighted_cross_entropy_with_logits](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/api_docs/python/functions_and_classes/shard8/tf.nn.weighted_cross_entropy_with_logits.md) , positional arguments are not disallowed. This may lead to silent failure.

martinwicke:
That's a good point. We should enforce names args for that function as
well.

cpdurham:
argument order was addressed in 1.0 with scripts for compatibility to add named fields (for google searches)

PR #7086, opened 2017-01-26
make weighted_cross_entropy consistent with sigmoid_cross_entropy.

ppwwyyxx:
Some relevant discussions in #6700.
Since Python doesn't allow positional arguments to come after keyword arguments, this PR will enforce `pos_weight` to be passed as named arguments as well.
But maybe it's still better to throw error than to fail silently.


alextp:
While this is an improvement over what is currently in the repository it would break a lot of existing code, and we're sufficiently far along in the 1.0 release process that this kind of API-breaking change is probably not a great idea.

I'm closing this for now, but it's a good idea to revisit it once things have stabilized more.

PR #7090, opened 2017-01-26
Fix wrong description for tf.nn.max_pool_with_argmax

ivankreso:
The description for indices is wrong. The real format is (y * width + x) * channels + c, batch index is ignored which can be seen [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc#L78).

New pull request: https://github.com/tensorflow/tensorflow/pull/7161

### DIFF
--- a/tensorflow/contrib/cmake/tf_core_cpu.cmake
+++ b/tensorflow/contrib/cmake/tf_core_cpu.cmake
@@ -6,6 +6,8 @@ file(GLOB_RECURSE tf_core_cpu_srcs
     "${tensorflow_source_dir}/tensorflow/cc/saved_model/*.cc"
     "${tensorflow_source_dir}/tensorflow/core/common_runtime/*.h"
     "${tensorflow_source_dir}/tensorflow/core/common_runtime/*.cc"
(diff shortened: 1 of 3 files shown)


tensorflow/g3doc/api_docs/python/nn.md
```diff
@@ -1209,8 +1209,7 @@ Performs the max pooling on the input.
 Performs max pooling on the input and outputs both max values and indices.
 
 The indices in `argmax` are flattened, so that a maximum value at position
-`[b, y, x, c]` becomes flattened index
-`((b * height + y) * width + x) * channels + c`.
+`[b, y, x, c]` becomes flattened index `(y * width + x) * channels + c`.
```
rmlarsen: The correct place to update the docstring would be here: 
￼https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/nn_ops.cc#L1305

ivankreso: Ok I created a new pull request here:
https://github.com/tensorflow/tensorflow/pull/7161

PR #7099, opened 2017-01-26
Fixed error due to api changes

ashutoshdas96:
Fixed TypeError in tf.nn.sampled_softmax_loss() function due to api changes.

This error was also reported in Udacity Discussion Forum:
https://discussions.udacity.com/t/assignment-5-ive-encountered-problem-when-execute-the-codes-provided-in-5-word2vec-ipynb/214239

PR #7105, opened 2017-01-27
Fixed a bug

clockzhong:
Tensorflow API has changed, so we also need to change the codes.
The detailed discussion is here:
https://discussions.udacity.com/t/assignment-5-ive-encountered-problem-when-execute-the-codes-provided-in-5-word2vec-ipynb


clockzhong:
I signed it!

vincentvanhoucke:
The diff appears to be truncated somehow. Your patch doesn't yield a parseable notebook. Can you check?

vincentvanhoucke:
Duplicate of #7099.

PR #7114, opened 2017-01-27
cross_validation is deprecated

hprovenza:
And should be replaced with model_selection.

### DIFF
--- a/tensorflow/examples/learn/boston.py
+++ b/tensorflow/examples/learn/boston.py
@@ -16,7 +16,7 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-from sklearn import cross_validation
+from sklearn import model_selection
 from sklearn import metrics
 from sklearn import preprocessing
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


hprovenza:
I signed it!

rmlarsen:
@martinwicke This would break the internal Google build, since our version of sklearn dores not have model_selection. What's your advice?

martinwicke:
We should update sklearn then. @ilblackdragon, wdyt?

hprovenza:
Just a note that pull request #7115 is an identical change in a different tutorial file; if you decide to make the internal changes necessary to merge this one I think it will be pretty trivial to take care of that one also.

dandelionmane:
friendly ping @martinwicke

martinwicke:
We have not updated sklearn internally so we cannot make this change yet.

hprovenza:
The matching change on #7114 is passing tests now; if this one is tested I would expect it to pass also

jhseu (approved):

jhseu:
Jenkins, test this please

Issue #2296, opened 2016-05-09
Issue with retrain.py, "could not convert string to float" when creating bottlenecks


oweingrod:
### Environment info
Operating System:
Ubuntu 14.04 
Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):
CUDA 7.5, cuDNN v2 (6.5)
`lrwxrwxrwx 1 root root 19 May  9 09:11 /usr/local/cuda -> /usr/local/cuda-7.5`

Installed the Nightly pip package from April 12th, with GPU support
Tensorflow version: 0.7.1

If installed from sources, provide the commit hash:

### Steps to reproduce
1. Running the retrain with a very large image directory (around 600k files) causes the training to fail around 3/4 the way through, during the bottleneck creation process. It always seems to fail while within the same label folder, but it does not appear that any file names within that label folder are corrupted or named incorrectly.

### What have you tried?
1. Changing the number of training steps
2. Checking to see if any files are less than 30k (theoretically, very small files are likely to be corrupted jpeg data)

### Logs or other output that would be helpful
For 610,000 files, the process can't seem to get through more than 415k. Am I simply just using too many files? Or am I missing a very, very subtle naming convention issue?

## Full stack trace:
```File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 30, in run
    sys.exit(main(sys.argv))
  File "retrain_reshape_error.py", line 695, in main
    cache_bottlenecks(sess, image_lists, FLAGS.image_dir, FLAGS.bottleneck_dir, jpeg_data_tensor, bottleneck_tensor)
  File "retrain_reshape_error.py", line 400, in cache_bottlenecks
    image_dir, category, bottleneck_dir, jpeg_data_tensor, bottleneck_tensor)
  File "retrain_reshape_error.py", line 372, in get_or_create_bottleneck
    bottleneck_values = [float(x) for x in bottleneck_string.split(',')] 
ValueErrror: could not convert string to float```

I should also note that I have been using this training process for the last few months with absolutely no failure.

Thanks!

Oren

yaroslavvb:
I don't think you are using too many files, however using large amounts of data makes it more likely to expose underlying bugs. The error indicates that "bottleneck_string" contains some non-numeric characters. Not familiar with that part of code, maybe @petewarden  has an idea how non-digits could've snuck into `bottleneck_string`?

oweingrod:
[...]

PR #7131, opened 2017-01-29
Adding checks for broken bottleneck files

rizasif:
An exception is raised when a broken cached bottleneck files is read. Such a file can be created as a result of sudden system failure. 

This commit refers the issue #2296 (https://github.com/tensorflow/tensorflow/issues/2296). Although closed but the problem still persists. Please view comments by @rizasif92 at the end. The changes contain nothing but error handling.

### DIFF
--- a/tensorflow/examples/image_retraining/retrain.py
+++ b/tensorflow/examples/image_retraining/retrain.py
@@ -392,9 +392,12 @@ def get_or_create_bottleneck(sess, image_lists, label_name, index, image_dir,
     with open(bottleneck_path, 'w') as bottleneck_file:
       bottleneck_file.write(bottleneck_string)
 
-  with open(bottleneck_path, 'r') as bottleneck_file:
-    bottleneck_string = bottleneck_file.read()
-  bottleneck_values = [float(x) for x in bottleneck_string.split(',')]
+  try:
+    bottleneck_values = [float(x) for x in bottleneck_string.split(',')]
+  except:
+    print("Invalid float found, sending None instead")
+    return None
+
   return bottleneck_values
 
 
@@ -427,12 +430,13 @@ def cache_bottlenecks(sess, image_lists, image_dir, bottleneck_dir,
     for category in ['training', 'testing', 'validation']:
       category_list = label_lists[category]
       for index, unused_base_name in enumerate(category_list):
-        get_or_create_bottleneck(sess, image_lists, label_name, index,
+        bNeck = get_or_create_bottleneck(sess, image_lists, label_name, index,
                                  image_dir, category, bottleneck_dir,
                                  jpeg_data_tensor, bottleneck_tensor)
-        how_many_bottlenecks += 1
-        if how_many_bottlenecks % 100 == 0:
-          print(str(how_many_bottlenecks) + ' bottleneck files created.')
+        if bNeck is not None:
+          how_many_bottlenecks += 1
+          if how_many_bottlenecks % 100 == 0:
+            print(str(how_many_bottlenecks) + ' bottleneck files created.')
 
 
 def get_random_cached_bottlenecks(sess, image_lists, how_many, category,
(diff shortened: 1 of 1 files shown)


rizasif:
I signed it!

petewarden (changes requested):
Thanks @rizasif , I really appreciate you putting this fix together! Could I request a slightly different approach though? I would prefer to keep the get_or_create_bottleneck() function always returning a bottleneck.

To ensure that, we could move the section nested inside the `if not os.path.exists(bottleneck_path):` condition into a new function, `create_bottleneck_file`. Then we can call it from that condition, but also have a flag that catches float exceptions when reading and calls it in that case too. Here's some untested code to try and show what I mean:

```
  if not os.path.exists(bottleneck_path):
    create_bottleneck_file(...)
  did_hit_error = False
  try:
    bottleneck_values = [float(x) for x in bottleneck_string.split(',')]
  except:
    print("Invalid float found, recreating bottleneck")
    did_hit_error = True
  if did_hit_error:
    create_bottleneck_file(...)
    # Allow exceptions to propagate here, since they shouldn't happen after a fresh creation
    bottleneck_values = [float(x) for x in bottleneck_string.split(',')]
```
I'm hoping with this approach none of the other changes outside this function should be needed since the contract to always return a list of bottleneck values is preserved.

Does that make sense?

rizasif:
@petewarden I agree with your solution and was thinking to implement it. I have made the changes as suggested, and tested them as well; I am attaching a screenshot of the output. Kindly review and let me know if anything else is required.
Thank you :)

capture

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -481,7 +498,6 @@ def get_random_cached_bottlenecks(sess, image_lists, how_many, category,
       ground_truth[label_index] = 1.0
       bottlenecks.append(bottleneck)
       ground_truths.append(ground_truth)
-      filenames.append(image_name)
```
petewarden: I think we should put this line back?

tensorflow/examples/image_retraining/retrain.py
```diff
@@ -493,11 +509,12 @@ def get_random_cached_bottlenecks(sess, image_lists, how_many, category,
                                               image_index, image_dir, category,
                                               bottleneck_dir, jpeg_data_tensor,
                                               bottleneck_tensor)
-        ground_truth = np.zeros(class_count, dtype=np.float32)
```
petewarden: Can we revert this change, since now the bottlenecks will never be None?

petewarden (changes requested):
This looking great, thanks for the updates! I have a couple of minor comments, but once those are addressed this looks good to go in.

rizasif:
@petewarden done

rmlarsen:

rmlarsen:
@rizasif it looks like you added some 4 character indents, which breaks the sanity check. TF uses 2 character indents. Please fix.

rizasif:
Indentation changes made, kindly re-run the tests. Thanks :)

rmlarsen:

rmlarsen:
Thanks @rizasif. @petewarden can you take another look? If the logics right to you, approve and I'll get it merged.

petewarden (approved):
This looks great, thanks!

rizasif:
Did the tests pass? :/

rmlarsen:
@rizasif it looks like you still have some bad indents, e.g.

FAIL: Found 1 non-whitelited pylint errors:
tensorflow/examples/image_retraining/retrain.py:353: [E0001(syntax-error), ] unexpected indent

rizasif:
@rmlarsen the code seems to be working at my end; so we might have to do this a few more times to resolve these indentation issues. I hope this commit works fine. @tensorflow-jenkins please test :)

rmlarsen:
@rizasif Thanks :) 

rmlarsen:
@rizasif thanks for the contribution!

rizasif:
Pleasure's all mine @rmlarsen 
Thank you so much for all the support. Hope to see you guys again. Take Care :)

PR #7135, opened 2017-01-29
Java: Do no strip symbols from C library when compiling in debug mode

karllessard:
When building the Java client, the linker should not strip symbols from the C library if we are compiling in debug mode, otherwise we won't be able to attach GDB to the java process for core debugging.

I thought that knowing the compilation mode might be something useful in general so I declared the "debug" and "optimize" config settings in the parent BUILD file.


klessard:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> :memo: Please visit https://cla.developers.google.com/ to sign.
> 
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

I signed it

klessard:
I signed it

klessard:
I've commited with the wrong email address... This should be fixed by now, since I replaced all commiter/author address with the right one (i.e. with the signed CLA).

PR #7137, opened 2017-01-29
[Java] Don't strip symbols from C library when compiling in debug mode

karllessard:
When building the Java client with "compilation_mode=dbg", the symbols in the C++ library should not be stripped otherwise we won't be able to attach a GDB to the java process for core debugging.

Also, I thought that knowing the active compilation mode might be something useful in general so I declared the "debug" and "optimize" config settings in the parent BUILD file.

### DIFF
--- a/tensorflow/BUILD
+++ b/tensorflow/BUILD
@@ -70,6 +70,22 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+config_setting(
+    name = "debug",
+    values = {
+        "compilation_mode": "dbg",
+    },
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "optimize",
+    values = {
+        "compilation_mode": "opt",
+    },
+    visibility = ["//visibility:public"],
+)
+
 package_group(
     name = "internal",
     packages = ["//tensorflow/..."],
--- a/tensorflow/BUILD
+++ b/tensorflow/BUILD
@@ -79,7 +79,7 @@ config_setting(
 )
 
 config_setting(
-    name = "optimize",
+    name = "optimized",
     values = {
         "compilation_mode": "opt",
     },
--- a/tensorflow/java/BUILD
+++ b/tensorflow/java/BUILD
@@ -111,6 +111,11 @@ cc_binary(
             LINKER_EXPORTED_SYMBOLS,
         ],
         "//tensorflow:windows": [],
+        "//tensorflow:debug" : [
+           	"-z defs",
+           	"-Wl,--version-script",  #  This line must be directly followed by LINKER_VERSION_SCRIPT
+           	LINKER_VERSION_SCRIPT,
+       	],
         "//conditions:default": [
             "-z defs",
             "-s",
--- a/tensorflow/java/BUILD
+++ b/tensorflow/java/BUILD
@@ -106,16 +106,12 @@ cc_binary(
     # symbols from the library. This reduces the size of the library
     # considerably (~50% as of January 2017).
     linkopts = select({
+        "//tensorflow:debug" : [], # Disable all custom linker options in debug mode
         "//tensorflow:darwin": [
             "-Wl,-exported_symbols_list",  # This line must be directly followed by LINKER_EXPORTED_SYMBOLS
             LINKER_EXPORTED_SYMBOLS,
         ],
         "//tensorflow:windows": [],
-        "//tensorflow:debug" : [
-           	"-z defs",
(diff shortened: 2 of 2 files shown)


tensorflow/java/BUILD
```diff
@@ -111,6 +111,11 @@ cc_binary(
             LINKER_EXPORTED_SYMBOLS,
         ],
         "//tensorflow:windows": [],
+        "//tensorflow:debug" : [
+           	"-z defs",
```
jart: Please fix indentation.

tensorflow/java/BUILD
```diff
@@ -111,6 +111,11 @@ cc_binary(
             LINKER_EXPORTED_SYMBOLS,
         ],
         "//tensorflow:windows": [],
+        "//tensorflow:debug" : [
```
jart: You might want to put these into a separate select block and + it with this one. This existing select block appears to be used for operating system detection.

jart (changes requested):

jart:
Assigning @gunan who probably knows more about this area of the build than me.

gunan:
@asimshankar for Java related changes.
I have no objection to the new config_settings defined in tensorflow/BUILD. But I am not sure about the modified default flags.

tensorflow/BUILD
```diff
@@ -70,6 +70,22 @@ config_setting(
...
+config_setting(
+    name = "debug",
+    values = {
+        "compilation_mode": "dbg",
+    },
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "optimize",
```
asimshankar: Nit: Mind using "optimized" instead of "optimize"?

tensorflow/java/BUILD
asimshankar: I agree that it would have been nice, but unfortunately nesting `select`s doesn't quite work out well (https://github.com/bazelbuild/bazel/issues/1623)

@karllessard : `-z defs` doesn't work on all platforms (the linker in OS X for example). How about we leave the set empty for debug builds. So, add `"//tensorflow:debug": [],` as the first condition in the `select` statement here so that for debug builds we don't do any custom linker configuration.

asimshankar (commented):
Thanks for the contribution!

Some minor changes suggested, but overall seems fine.

Out of curiosity: Were you trying to step through native code in GDB or did you see some crashes? Ideally, it wouldn't be possible to have the native code crash by calls to the Java API, so if that is happening and indicates some holes, I'd like to plug those.

Thanks!

tensorflow/BUILD
karllessard: Haha, I hesitated myself too long on this one, that 'd' probably appeared/disappeared more than once in my editor... no problem, I'll do all recommended fixes tonight.

For your previous question, no I did not notice a crash, I was just not able to break in the C++ code before I realized that all symbols were missing.

asimshankar (approved):
Looks good, thanks

gunan:
Jenkins, test this please.

karllessard:
> Jenkins, test this please.

Eeh, I'll be really surprised if my 2-lines patch really broke all sanity checks, looking at the Jenkins console it seems that the process has been aborted or something else went wrong. @gunan or anyone, can you please have a look at it?
```
Build was aborted
Aborted by unknown
Setting status of 5d48a946f02184a6f82d556238e6c2b3db1c5fdb to FAILURE with url https://ci.tensorflow.org/job/tensorflow-pull-requests-android/3425/ and message: 'FAILURE
...
```

gunan:
The failure is in "Sanity" build, which runs pylint and buildifier.
Looks like there is a buildifier error:
https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/2895/console

tensorflow/java/BUILD
```diff
@@ -106,6 +106,7 @@ cc_binary(
     # symbols from the library. This reduces the size of the library
     # considerably (~50% as of January 2017).
     linkopts = select({
+        "//tensorflow:debug" : [], # Disable all custom linker options in debug mode
```
gunan:
```
110c110
<         "//tensorflow:debug" : [], # Disable all custom linker options in debug mode
---
>         "//tensorflow:debug": [],  # Disable all custom linker options in debug mode
```
Two spacing errors here. No space between `...debug"` and `:`. Then exactly two spaces between `[],` and `#`

gunan:
Jenkins, test this please.

rmlarsen:
It looks like this is ready to merge, @gunan can you confirm?

gunan (approved):

gunan:
I think all of the comments from @jart are addressed, too.

PR #7154, opened 2017-01-31
Update inception_v3.py

ymswhu:
the operation array_ops.concat has been changed, which the parameters should be concat_dim followed by values. fixed this problem which could lead a bug in fine-tune and training


samjabrahams:
I believe the original code is correct- check out the note for `tf.concat` for the [v1.0.0rc0](https://github.com/tensorflow/tensorflow/releases/tag/v1.0.0-rc0) release. The changes you're making are actually changing them _back_ to the old function signature.

rmlarsen:
@ymswhu, @samjabrahams is correct.

PR #7161, opened 2017-01-31
Fix wrong description for tf.nn.max_pool_with_argmax

ivankreso:
The description for indices is wrong. The real format is (y * width + x) * channels + c, batch index is ignored which can be seen [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc#L78).

Fixed pull request from https://github.com/tensorflow/tensorflow/pull/7090

### DIFF
--- a/tensorflow/core/ops/nn_ops.cc
+++ b/tensorflow/core/ops/nn_ops.cc
@@ -1301,8 +1301,7 @@ REGISTER_OP("MaxPoolWithArgmax")
 Performs max pooling on the input and outputs both max values and indices.
 
 The indices in `argmax` are flattened, so that a maximum value at position
-`[b, y, x, c]` becomes flattened index
-`((b * height + y) * width + x) * channels + c`.
+`[b, y, x, c]` becomes flattened index `(y * width + x) * channels + c`.
 
 ksize: The size of the window for each dimension of the input tensor.
 strides: The stride of the sliding window for each dimension of the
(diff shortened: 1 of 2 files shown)


rmlarsen:
I'm not sure your change is correct. I think the batch index is used at the top level to shard the computation across threads here:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc#L212

The line you pointed to calculates the index within such a chunk, I think.

@bsteiner, care to comment?

benoitsteiner:
The cuda loop iterates over batch * channels * pooled_height * pooled_width indices. The batch dimension is recovered in the 'n' variable computed by dividing the cuda loop index by pooled_width * pooled_height * channels. So I believe the description is correct.

ivankreso:
'n' variable is not used to compute the index because the data pointer is local for each example.
```
const dtype* bottom_data_n = bottom_data + n * channels * height * width;
...
int idx = c * height * width + h * width + w;
...
mask[index] = maxidx;
```
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc#L75

renmengye:
Why is this pull request still not being merged? I see the old wrong doc with batch dimension on tensorflow official website as of today.

PR #7163, opened 2017-01-31
[WIP] Add standlone Poplar plugin scaffolding

DavidNorman:
Summary:
1) Add poplar and popnn third_party library support
2) Add a poplar XLA device, plugin python wrapper, and appropriate test
3) Adjust XLA unit tests to allow for a plugin, and add IPU as a plugin type


vrv:
I asked David to send this PR so we could see how we could get an XLA device added as a purely external library via tf.load_library.  These are the changes he had to make to get something to load, but we should figure out what changes to TF need to be made to allow this support.

I'm going to be OOO most of this week, so if someone wants to jump in, great!  I'll return this when I am back

DavidNorman:
Some notes:

for the external library (third_party/ipus/...), I have used 2 environment variables to enable the module.  This differs from Cuda / OpenCL which have stuff in the ./configure script.   I avoided copying Cuda / OpenCL because the ./configure script seems messy, and I imagine that it will be cleaned up at some point.

To build our device (it is just a stub copy of the xla cpu device at the moment), you could set the 2 environment variables (TF_POPLAR_BASE and TF_POPNN_BASE) to any empty directory, comment out the two included headers in the xla_ipu_device.cc file.  then ./configure, including XLA, and build something like:

bazel test --config=opt //tensorflow/compiler/poplar:device_test

this should demonstrate the device .so being built, and the .so and .py files working together to load the device into tensorflow, and the test checking that the XLA_IPU device has been registered.

You could also try:

bazel test --config=opt //tensorflow/compiler/tests:xla_device_test

you should see that the CPU device passes, while the IPU device doesn't pass (because it doesn't register any of the ops that are required at the moment).


Finally - I've checked that without the TF_POPLAR_BASE and TF_POPNN_BASE env vars set, the system builds with XLA and the unit tests run successfully.    i.e. I would like this merge request to be actually merged after any discussion is done.

rmlarsen:
@vrv would you like to assign some possible reviewers for this?

DavidNorman:
Hi,

Further to this, I've been working on added a 'platform' and a 'compiler' to the driver.  I have discovered the file tensorflow/tf_exported_symbols.lds requires something adding to it to ensure that XLA can be accessed by a plugin.  I've added *xla*, which works, but generates some warnings on link.

Consequently I have managed to add a platform, compiler and executable (stubs) for my hardware.  I could add these to the pull request if you like.

I have had to add a constant to the stream_executor library that holds my hardware platform ID.  Would it be a good idea to assume that nobody will ever need to load plugins for more than one type of exotic hardware, and just have a single 'kThirdPartyPluginPlatformId' for third party hardware?

DavidNorman:
sorry - that was some bad Git management.  sigh :(  - tried to merge one commit onto the top of the pile and sucked in every merge and old bit of work on the branch.

I think I will try this whole thing again with another pull request.

vrv:
No worries David, and thanks for purusing this. 

1) From a brief look at this code, it appears there isn't much that needs to actually be added to TensorFlow core code (most of the poplar-specific code could live in its own repository, using bazel or not) -- this is great, and I'm glad you went through the effort of helping to identify where it works and where it doesn't.

2) There are some exceptions like you mention (such as the id registration) and several changes to extend stream executor.  @hawkinsp do you have an idea of what could be changed to core TF so that this custom device could be completely separate from the core repo?  @henline too, since stream executor will eventually be a separate project in llvm, and it would be good to understand what needs to be easily extendable without commits to core code.

DavidNorman:
I have replaced this one with https://github.com/tensorflow/tensorflow/pull/7191

PR #7180, opened 2017-02-01
Bump up size for estimator_test

terrytangyuan:
Avoid estimator_test timeout as seen in other PR builds, such as https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu/3437/console

### DIFF
--- a/tensorflow/contrib/learn/BUILD
+++ b/tensorflow/contrib/learn/BUILD
@@ -477,7 +477,7 @@ py_test(
 
 py_test(
     name = "estimator_test",
-    size = "medium",
+    size = "large",
     srcs = ["python/learn/estimators/estimator_test.py"],
     srcs_version = "PY2AND3",
     deps = [
--- a/tensorflow/contrib/learn/BUILD
+++ b/tensorflow/contrib/learn/BUILD
@@ -478,8 +478,8 @@ py_test(
 py_test(
     name = "estimator_test",
     size = "medium",
-    shard_count = 4,
     srcs = ["python/learn/estimators/estimator_test.py"],
+    shard_count = 4,
     srcs_version = "PY2AND3",
     deps = [
         ":learn",
--- a/tensorflow/contrib/learn/BUILD
+++ b/tensorflow/contrib/learn/BUILD
@@ -477,7 +477,8 @@ py_test(
 
 py_test(
     name = "estimator_test",
-    size = "large",
+    size = "medium",
+    shard_count = 4,
     srcs = ["python/learn/estimators/estimator_test.py"],
     srcs_version = "PY2AND3",
     deps = [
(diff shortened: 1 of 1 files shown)


gunan:
@martinwicke This change will disable this test in some of our CI builds.
I think the change is warranted right now. But we will eventually need to resolve it.

caisq:
@gunan @martinwicke @terrytangyuan estimator_test is currently not sharded. Is it possible to shard it, like `shard_count = 4`?

terrytangyuan:
@caisq Good idea. I've changed the commit to add shard count, assuming all tests are independent.

terrytangyuan:
@gunan Just for my curiosity. Large tests are disabled internally? Under what environments/devices?

rmlarsen:

tensorflow/contrib/learn/BUILD
```diff
@@ -478,6 +478,7 @@ py_test(
 py_test(
     name = "estimator_test",
     size = "medium",
+    shard_count = 4,
```
rmlarsen: this breaks the build. Should this be a tf_py_test to support shard_count?

rmlarsen (dismissed):

tensorflow/contrib/learn/BUILD
caisq: +1

terrytangyuan: Why is that? All other tests that use shard_count seem to be py_test

caisq:
@terrytangyuan You are right. The test failures may be due to a BUILD file format requirement. Can you move the `shard_count` line after `srcs` and before `srcs_version`.

gunan:
No, please do not use sharding, they can cause even worse complications
with testinfra. Id rather we just disable this test.

Large tests have 15 minute timeouts, and build phase added on top of that
takes them out of our goals for presubmit time limits. Therefore we
conciously disable large tests in our testinfra.

terrytangyuan:
Okay...should I change this to manual? Let me know what's the correct way
to do it. I think it's better to document the best practices somewhere to
avoid this back and forth conversation.

martinwicke:
The best thing to do would be to split this test into more manageable
pieces, or even better, make it faster (less iterations, etc., if such a
thing is still possible).

On Wed, Feb 1, 2017 at 10:56 AM, Yuan (Terry) Tang <notifications@github.com
> wrote:

> Okay...should I change this to manual? Let me know what's the correct way
> to do it. I think it's better to document the best practices somewhere to
> avoid this back and forth conversation.
>

gunan:
Id prefer either Martin's suggestion, or disabling the test. I am OK with
either right now, as this is becoming a blocker for many contributions.

gunan (approved):

terrytangyuan:

rmlarsen:
The failing tests are unrelated. Merging.

PR #7182, opened 2017-02-01
Fix possible flake sources in saver_test and supervisor_test.

gunan:


tensorflow/python/training/saver_test.py
```diff
@@ -696,7 +684,10 @@ def _testPartitionedVariables(self, use_resource):
     var_full_shape = [10, 3]
     # Allows save/restore mechanism to work w/ different slicings.
     var_name = "my_var"
-    saved_path = os.path.join(_TestDir("partitioned_variables"), "ckpt")
+    saved_dir = os.path.join(self.get_temp_dir(), "partitioned_variables")
```
caisq: These two lines are very repetitive in this test class. Can't we just create a temporary directory for each 

def _test_dir(self, test_name):
  test_dir = os.path.join(self.get_temp_dir(), test_name)
  gfile.MakeDirs(test_dir)
  return test_dir

caisq (changes requested):

tensorflow/python/training/saver_test.py
gunan: unlike supervisor_test, this test file has a lot of test classes.
So even when we do that, it will still have to be repeated 15 times.

One possible solution might be to create a new test base class in this file that implements the method, and base all others on top of that.
Do you think it is worth it?

caisq: I think it's still worthwhile to reduce the number of repeated lines by 50%. It's going to make it easier to maintain and less error prone.

gunan: Done.

caisq (approved):

rmlarsen:
Test failures are unrelated. Merging.

PR #7191, opened 2017-02-01
[WIP] Add standlone Poplar plugin scaffolding

DavidNorman:
This pull request replaces https://github.com/tensorflow/tensorflow/pull/7163

Summary:
1) Add poplar and popnn third_party library support
2) Add a poplar XLA device, plugin python wrapper, and appropriate test
3) Adding copyright messages, and 'poplar configured' function
4) Adding platform, compiler and executable stubs for IPU/Poplar
5) Add XLA symbols to the external symbols in tensorflow main dynamic library
6) Add a transfer manager for Poplar


DavidNorman:
Replying to the comment by @vrv at the end of the closed pull request:

i think it would be great if the poplar plugin could be entirely separate from the main repo.  I think that what is required is some exporting of XLA headers, much like what is done for plugin user ops.

It would certainly please my lords and masters if we could place the whole thing in a different repo.  As it stands, I think that the poplar stuff needs to be in the bazel build system, because there isn't really a pretty way of getting at all of the headers for a compilation otherwise.

It would be excellent, though, if I were able to download a 'TF compiled with XLA' binary from the web, and it came with exported headers for all necessary XLA interfaces.  Or even if I were able to build one locally, and then compile my driver against XLA headers installed in the python site lib.

vrv:
@DavidNorman Take a look at this rule: 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/BUILD#L22   -- it is used by the pip packager to copy over all relevant headers needed to build user ops into the pip-installed directories (Ideally, we'd be able to provide a development header library that doesn't involve python, but pip works in the short term).

It is entirely probable that the transitive headers required for XLA are not packaged properly -- one thing that might be worth doing is looking at https://www.tensorflow.org/how_tos/adding_an_op/#with_tensorflow_binary_installation to figure out how you would compile user ops, but doing the same for the XLA device -- for every header that is missing, we would add it to that 'included_headers' rule somehow, so you can simply use gcc or clang with the appropriate includes to build a custom device shared object.  If that solves your bazel dependency problem, let us know!

@hawkinsp @eliben  I don't know know too much about what xla headers need to be public -- is it possible for you to work out what should be exported for people who want to develop a custom device in XLA, so we can add it to the appropriate header libraries?

DavidNorman:
i think that if the pip installation contained the appropriate XLA external interface headers then would be great.  in the longer term, like you say, a non-pip based export would be good.  You don't provide a binary XLA w/o GPU download at the moment.  Is there an intention to provide this in the longer term?

my only concern then would be that I really like your XLA unit tests.  It would be a shame to be unable to use them to test the plugin.

vrv:
Yeah, putting all XLA dev headers in pip or non-pip is a great goal.  Not sure about the binary XLA, maybe the others can speak to that.

Also, it indeed would be nice to have an "XLA test validation suite" that was built in a way that you didn't need to have a custom device specified in our code, so feature request noted :)

DavidNorman:
The binary thing isn't too much trouble anyway. We can always build it locally. 

Today I realised that my cc_library target is generating a .so file without any dependent .so files in it (as expected now I think about it). 

This is quite inconvenient, and is another reason why having an XLA external headers target is important. 

I can work around it with another genrule that calls a utility to insert the relevant dependencies. 

I'll start an OOB conversation about a couple of other queries I have.

vrv:
I believe progress is being made out-of-band, so marking as 'stalled' for now.

DavidNorman:
Indeed - i don't believe that this merge request should ever actually be merged.  Apart from cherry picking any little bits in the core code base.

Incidentally, I discovered that *protobuf* should probably be added to the exports file, along with *xla*

vrv:
Okay thanks @DavidNorman for sending this -- I think we just need to work a bit more on XLA packaging of headers and we should be good -- feel free to send us PRs for what you think needs to be exposed if we're a bit overwhelmed to get it to it immediately.

PR #7211, opened 2017-02-02
Docfix: Document usage of --config=opt for SIMD instruction sets.

gunan:


tensorflow/core/BUILD
```diff
@@ -811,7 +811,7 @@ filegroup(
 # Compiles to a trivial library on non-Android to prevent irrelevant
 # build errors. If not building this as part of an android_binary,
 # a command such as the following must be used:
-# bazel build -c opt tensorflow/core:android_tensorflow_lib \
+# bazel build --config opt tensorflow/core:android_tensorflow_lib \
```
martinwicke: Wait, this is an android build -- didn't we see that here it won't work? So this should stay -c opt

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -856,13 +856,13 @@ default and if you want to limit RAM usage you can add `--local_resources
...
-$ bazel build -c opt //tensorflow/tools/pip_package:build_pip_package
+$ bazel build --config opt //tensorflow/tools/pip_package:build_pip_package
 
 # To build with support for CUDA:
-$ bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package
+$ bazel build --config opt --config=cuda //tensorflow/tools/pip_package:build_pip_package
 
-# Alternatively, to build with support for OpenCL:
-$ bazel build -c opt --config=sycl //tensorflow/tools/pip_package:build_pip_package
+# Alternatively, to build with support for OpenCL(Experimental):
```
martinwicke: a space before (experimental)?

tensorflow/g3doc/get_started/os_setup.md
```diff
@@ -873,20 +873,21 @@ $ sudo pip install /tmp/tensorflow_pkg/tensorflow-0.12.1-py2-none-any.whl
 ## Optimizing CPU performance
 
 To be compatible with as wide a range of machines as possible, TensorFlow
-defaults to only using SSE4.1 SIMD instructions on x86 machines. Most modern PCs
+defaults to only using SSE4 SIMD instructions on x86 machines. Most modern PCs
```
martinwicke: Maybe instead: "...as possible, TensorFlow binaries only use SSE4 SIMD instructions. Most modern computers support mode advanced instructions, ..."

And, are we building with SSE4? Or only SSE3?

tensorflow/tools/benchmark/BUILD
```diff
@@ -59,7 +59,7 @@ tf_cc_test(
 
 # This binary may be built for either desktop or Android.
 # A typical Android build command will look like the following:
-# bazel build -c opt tensorflow/core:android_tensorflow_lib \
+# bazel build --config opt tensorflow/core:android_tensorflow_lib \
```
martinwicke: Again, android can't use --config opt, right?

martinwicke (changes requested):

tensorflow/g3doc/get_started/os_setup.md
gunan: when I think about it, we only build with -c opt for now.
What does that bring by default?

tensorflow/g3doc/get_started/os_setup.md
gunan: done

tensorflow/core/BUILD
gunan: Yes, it is. reverting this change.

tensorflow/tools/benchmark/BUILD
gunan: done.

tensorflow/g3doc/get_started/os_setup.md
martinwicke: @rmlarsen investigated once, I think nothing at all. Only machine independent optimizations (-O2 or something?)

tensorflow/tools/benchmark/BUILD
gunan: reverted

martinwicke (approved):

yaroslavvb:
btw, --march=native doesn't seem to enable all optimizations, so there's some confusion as to what flags need to be set (someone's getting 30% speedup by adding fma/avx/avx2/mpmath flags in addition to march=native) http://stackoverflow.com/questions/41996006/should-i-enable-simd-extensions-together-with-cuda-support-when-building-tensorf)

martinwicke:
See comment on SO thread -- It's all good.

PR #7229, opened 2017-02-02
Improve upgrade script to handle list comprehensions as arguments.

aselle:
python's ast module does not return the correct location, so we
have to do our best to scan backwards to find where the [ token
that trully started the list comprehension occurs.

### DIFF
--- a/tensorflow/tools/compatibility/tf_upgrade.py
+++ b/tensorflow/tools/compatibility/tf_upgrade.py
@@ -347,6 +347,62 @@ class TensorFlowCallVisitor(ast.NodeVisitor):
     items.append(curr.id)
     return ".".join(reversed(items))
 
+  def _find_true_position(self, node):
+    """Return correct line number and column offset for a given node.
+
+    This is necessary mainly because ListComp's location reporting reports
+    the next token after the list comprehension list opening.
+
+    Args:
+      node: Node for which we wish to know the lineno and col_offset
+    """
+    import re
+    find_open = re.compile("^\s*(\\[).*$")
+    find_string_chars = re.compile("['\"]")
+
+    if isinstance(node, ast.ListComp):
+      # Strangely, ast.ListComp returns the col_offset of the first token
+      # after the '[' token which appears to be a bug. Workaround by
+      # explicitly finding the real start of the list comprehension.
+      line = node.lineno
+      col = node.col_offset
+      # loop over lines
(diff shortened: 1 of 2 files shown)


tensorflow/tools/compatibility/tf_upgrade.py
```diff
@@ -347,6 +347,60 @@ def _get_attribute_full_path(self, node):
...
+            # cases where the comment does not contain string literal starting/
+            # ending characters. If ast gave us start and end locations of the
+            # ast nodes rather than just start, we could use string literal
+            # node ranges to filter out spurious #'s that appear in string
+            # literals.
+            comment_start = prev_line.find("#")
+            if comment_start ==  -1:
+              col = len(prev_line)
+            elif find_string_chars.search(prev_line[comment_start:]) is None:
+              col = comment_start
```
annarev: Should the "col" be one of the return values?

annarev (commented):
Can you add some tests with '[', spaces, and comments?

tensorflow/tools/compatibility/tf_upgrade.py
aselle: Tests are forthcoming (I used some to develop this that I haven't codified yet, but work), but I wanted to get this version up asap so that people wouldn't be hit by these problems. I'll put them in now.

annarev: "col" here still looks unused to me. Is it expected?

tensorflow/tools/compatibility/tf_upgrade_test.py
```diff
@@ -134,3 +146,31 @@ def testInplace(self):
 
 if __name__ == "__main__":
   test_lib.main()
+
+"""tf.concat(0, [tf.concat(0, tiles[y]) for y in range(4)])
```
annarev: Do you plan to use this comment later on? I am just making sure it is not a typo that you planned to remove.

annarev (commented):
Thank you for adding tests! The change looks good to approve only pending the question about "col" variable.

tensorflow/tools/compatibility/tf_upgrade_test.py
aselle: removed

aselle:
I've added tests now.

annarev (approved):

PR #7233, opened 2017-02-02
Fix by Josh Levenberg for cmake build

rmlarsen:


gunan (approved):

gunan:
Should have no effect on py3 build. Merging.

PR #7237, opened 2017-02-03
Fix cudnn filename in configure script.

gunan:


tensorflow/core/platform/default/build_config.bzl
```diff
@@ -7,7 +7,7 @@ load("//tensorflow:tensorflow.bzl", "if_not_mobile")
 # configure may change the following lines
 WITH_GCP_SUPPORT = False
 WITH_HDFS_SUPPORT = False
-WITH_JEMALLOC = True
```
rmlarsen: You didn't mention this change in the description. Is it intentional?

gunan: It is not. It is an artifact of trying to locally test the change.
Removed.

rmlarsen (approved):

rmlarsen:
@gunan looks like agent died again for the mac build. Want to just kick off that one?

rmlarsen:

gunan:
Sorry for the delay, either way, I will investigate why agents are going offline this much today.

PR #7240, opened 2017-02-03
Add graph parameter to tf.import_graph_def()

samjabrahams:
Allows user to indicate a specific `Graph` to place `GraphDef` Operations. This keeps users from needing to use a `with g.as_default()` statement in their code. The documentation for `import_graph_def()` has also been improved to inform users that the default graph is used.. by default. For tests, I've simply done an altered version of the basic test thus far- let me know what further testing you'd like to see.

This should maintain backwards compatibility.

### DIFF
--- a/tensorflow/python/framework/importer.py
+++ b/tensorflow/python/framework/importer.py
@@ -149,9 +149,9 @@ def _FindAttrInOpDef(attr_name, op_def):
   return None
 
 
-def import_graph_def(graph_def, input_map=None, return_elements=None,
-                     name=None, op_dict=None, producer_op_list=None):
-  """Imports the TensorFlow graph in `graph_def` into the Python `Graph`.
+def import_graph_def(graph_def, input_map=None, return_elements=None, name=None,
+                     op_dict=None, producer_op_list=None, graph=None):
+  """Imports the TensorFlow graph in `graph_def` into a Python `Graph`.
 
   This function provides a way to import a serialized TensorFlow
   [`GraphDef`](https://www.tensorflow.org/code/tensorflow/core/framework/graph.proto)
@@ -159,6 +159,8 @@ def import_graph_def(graph_def, input_map=None, return_elements=None,
   [`Tensor`](#Tensor) and [`Operation`](#Operation) objects. See
   [`Graph.as_graph_def()`](#Graph.as_graph_def) for a way to create a
   `GraphDef` proto.
+
+  By default, imports to the graph returned by `get_default_graph()`.
 
   Args:
     graph_def: A `GraphDef` proto containing operations to be imported into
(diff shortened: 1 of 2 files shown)


tensorflow/python/framework/importer.py
```diff
@@ -232,232 +237,237 @@ def import_graph_def(graph_def, input_map=None, return_elements=None,
...
-
-          if input_name in input_map:
-            # (b) Input should be replaced by a tensor from the caller.
-            source_tensor = input_map[input_name]
-            used_input_keys.add(input_name)
-
-          else:
-            # (c) Input should be taken from an op in `graph_def`.
-            operation_name, output_index = _ParseTensorName(input_name)
+  g = ops.get_default_graph() if graph is None else graph
```
samjabrahams: Actual new piece of code here, pretty much everything below this line (aside from `with g.as_default():`) is a diff caused by indentation and tweaks to keep the column width below 80.

tensorflow/python/framework/importer.py
```diff
@@ -232,232 +237,237 @@ def import_graph_def(graph_def, input_map=None, return_elements=None,
...
+  with g.as_default():
+    with ops.name_scope(name, 'import', input_map.values()) as scope:
```
samjabrahams: BEGIN THE WHITE SPACE DIFF

tensorflow/python/framework/importer.py
```diff
@@ -232,232 +237,237 @@ def import_graph_def(graph_def, input_map=None, return_elements=None,
...
+              raise ValueError(
+                  'Requested return_element %r not found in graph_def.' % name)
+          else:
+            try:
+              ret.append(name_to_op[name])
+            except KeyError:
+              raise ValueError(
+                  'Requested return_element %r not found in graph_def.' % name)
+        return ret
+    # LINT.ThenChange(//tensorflow/core/graph/graph_constructor.cc)
```
samjabrahams: END THE WHITE SPACE DIFF

tensorflow/python/framework/importer_test.py
```diff
@@ -219,6 +219,72 @@ def testBasic(self):
       # Check that the op_def is still available.
       self.assertNotEqual(None, a.op_def)
 
+      # Check that the Ops were placed in the graph
```
samjabrahams: Added some tests to `testBasic` to make sure the added Ops were in the default graph.

tensorflow/python/framework/importer_test.py
```diff
@@ -219,6 +219,72 @@ def testBasic(self):
...
+      self.assertEqual(a, g.get_operation_by_name("import/A"))
+      self.assertEqual(b, g.get_operation_by_name("import/B"))
+      self.assertEqual(c, g.get_operation_by_name("import/C"))
+      self.assertEqual(d, g.get_operation_by_name("import/D"))
+
+  def testBasicGraphParam(self):
```
samjabrahams: Basically a copy/paste of `testBasic`, without a call to `as_default()`

martinwicke:
Ok, we looked at this -- since we don't take graph arguments on most functions, and since this is the same behavior as 

```
with g.as_default():
  tf.import_graph_def(...)
```

We probably don't want to add the extra argument here.

samjabrahams:
@martinwicke what are your thoughts on tweaking the documentation to indicate that the imported Ops get placed into the default graph? Something like making the first line read "Imports the TensorFlow graph in `graph_def` into the _current default_ Python `Graph`".

martinwicke:
I think that would be a good idea.

samjabrahams:
Submitted the documentation change as a PR in #7325

PR #7252, opened 2017-02-03
Prevent Python unit test files from going into PIP package

caisq:
See also experimental PIP build at (Jenkins login required to view):
http://ci.tensorflow.org/job/experimental-cais-linux-python35-pip/6/console


tensorflow/tools/pip_package/build_pip_package.sh
```diff
@@ -134,13 +134,17 @@ function main() {
...
+  rm -f $(find ./ -name '*_test.py')
+
   rm -f MANIFEST
   echo $(date) : "=== Building wheel"
   "${PYTHON_BIN_PATH:-python}" setup.py bdist_wheel ${GPU_FLAG} >/dev/null
   mkdir -p ${DEST}
   cp dist/* ${DEST}
   popd
-  rm -rf ${TMPDIR}
+  # rm -rf ${TMPDIR}
```
rmlarsen: Cruft?

gunan:
It is possible due to recent issues pip tests are disabled in presubmits. Could you setup an experimental build and run any pip tests, @caisq ?

tensorflow/tools/pip_package/build_pip_package.sh
caisq: Fixed now.

caisq:
@gunan, I (kind of) did that already. See the original PR comment. But that particular build timed out at 90 minutes. I increased the timeout to 120 minutes and kicked off another one:

http://ci.tensorflow.org/view/Experimental/job/experimental-cais-linux-python35-pip/7/console

caisq:
To quantify the reduction in the PIP wheel file size (in bytes, for Python 3.5):
Before: 43691027
After: 43507664
About 180 kB (0.5% reduction).

Not as much as one would expect :)  But it gets rid of 175 unnecessary files and about 2.5 MB disk space when installed.

gunan:
Change mostly looks good.
Would this lead to any changes in public API?

martinwicke:
I am not a fan of how this is done -- can we instead put this exception into the build_pip_package rule? Otherwise we have another point at which to control what goes into the pip package -- we already have too many.

tensorflow/tools/pip_package/build_pip_package.sh
```diff
@@ -134,6 +134,10 @@ function main() {
   source tools/python_bin_path.sh
 
   pushd ${TMPDIR}
+
+  # No Python unit test files should go into the PIP package.
+  rm -f $(find ./ -name '*_test.py')
```
martinwicke: I would much rather we adjust the dependencies of the build_pip_package rule instead. We already have too many point at which we control what goes into that package.

martinwicke (changes requested):

tensorflow/tools/pip_package/build_pip_package.sh
caisq: I agree this is not the most elegant. Let me work on this a little more.

caisq:
I'll close this for now. There is a related issue of unnecessary include header files in the pip package. I'll reopen the PR again when I have a more elegant fix for both issues.

PR #7275, opened 2017-02-05
Fixed extraction step when using data_root

mlucool:
This fixes an issues introduced in a81c968d7484b04ef6bffe6aef235d1d00c8a81a


benoitsteiner:
@mlucool I believe this is conflicting with https://github.com/tensorflow/tensorflow/pull/7366.

mlucool:
Ah someone else fixed it before me 👍

PR #7277, opened 2017-02-06
Improve docs for incompatible protobuf error

rkaplan:
On Mac OS X, importing tensorflow can result in the error: "TypeError: `__init__()` got an unexpected keyword argument 'syntax'". As explained in the doc, this is caused by an incompatible protobuf version. The suggested fix, "pip install --upgrade protobuf", doesn't resolve the issue if the user has installed protobuf via Homebrew. This commit adds an example command to fix the error in this case.

### DIFF
--- a/tensorflow/g3doc/get_started/os_setup.md
+++ b/tensorflow/g3doc/get_started/os_setup.md
@@ -1212,6 +1212,12 @@ installed, such as:
 $ pip install --upgrade protobuf
 ```
 
+Or (if you have protobuf installed with Homebrew):
+
+```bash
+$ brew upgrade protobuf
+```
+
 ### Mac OS X: Segmentation Fault when import tensorflow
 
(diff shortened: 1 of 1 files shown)


rkaplan:
Signed the CLA

gunan:
Looks like homebrew only has protobuf up to 3.0
http://brewformulas.org/Protobuf
Our pip packages require protobuf 3.1, and we are upgrading to 3.2.
Maybe we should recommend against using homebrew installed protobuf?

@jhseu wdyt?

jhseu:
Homebrew actually has protobuf 3.2.0. That page is outdated. Tested on my mac:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/protobuf.rb

This change seems reasonable to me.

jhseu (approved):

gunan:
Great, thanks for verifying!
Merging.

PR #7279, opened 2017-02-06
resolve #6762 on ldconfig only available on root PATH

### LINKED ISSUE #6762: configure assumes ldconfig can be called and that CuDNN is installed in a system library location

byronyi:
See Linux man page for [ldconfig(8)](https://linux.die.net/man/8/ldconfig), synopsis section:

```bash
/sbin/ldconfig [ -nNvXV ] [ -f conf ] [ -C cache ] [ -r root ] directory ...
/sbin/ldconfig -l [ -v ] library ...
/sbin/ldconfig -p
```

Former pull request #6848 has been closed as @jowagner has not signed a CLA.

### DIFF
--- a/configure
+++ b/configure
@@ -458,7 +458,7 @@ while true; do
   if is_linux; then
     if ! type ldconfig > /dev/null; then
         LDCONFIG_BIN=/sbin/ldconfig
-      else
+    else
         LDCONFIG_BIN=ldconfig
     fi
     CUDNN_PATH_FROM_LDCONFIG="$($LDCONFIG_BIN -p | sed -n 's/.*libcudnn.so .* => \(.*\)/\1/p')"
--- a/configure
+++ b/configure
@@ -456,7 +456,7 @@ while true; do
   fi
 
   if is_linux; then
-    CUDNN_PATH_FROM_LDCONFIG="$(ldconfig -p | sed -n 's/.*libcudnn.so .* => \(.*\)/\1/p')"
+    CUDNN_PATH_FROM_LDCONFIG="$(/sbin/ldconfig -p | sed -n 's/.*libcudnn.so .* => \(.*\)/\1/p')"
     if [ -e "${CUDNN_PATH_FROM_LDCONFIG}${TF_CUDNN_EXT}" ]; then
       export TF_CUDNN_VERSION
       export CUDNN_INSTALL_PATH="$(dirname ${CUDNN_PATH_FROM_LDCONFIG})"
--- a/configure
+++ b/configure
@@ -456,7 +456,7 @@ while true; do
   fi
 
   if is_linux; then
-    if ! type ldconfig > /dev/null; then
+    if ! type ldconfig > /dev/null 2>&1; then
         LDCONFIG_BIN=/sbin/ldconfig
     else
         LDCONFIG_BIN=ldconfig
--- a/configure
+++ b/configure
@@ -456,7 +456,12 @@ while true; do
   fi
 
   if is_linux; then
-    CUDNN_PATH_FROM_LDCONFIG="$(/sbin/ldconfig -p | sed -n 's/.*libcudnn.so .* => \(.*\)/\1/p')"
+    if ! type ldconfig > /dev/null; then
+        LDCONFIG_BIN=/sbin/ldconfig
+      else
+        LDCONFIG_BIN=ldconfig
+    fi
+    CUDNN_PATH_FROM_LDCONFIG="$($LDCONFIG_BIN -p | sed -n 's/.*libcudnn.so .* => \(.*\)/\1/p')"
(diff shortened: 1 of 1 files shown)


jowagner:
This patch may break on some platforms. The FHS states that ldconfig is "usually located in /usr/sbin". The alternative location "/sbin" is described as optional.

jowagner:
openSUSE and Debian: ok (/sbin)

byronyi:
Can someone review this PR? Current master with CUDA still breaks on latest Debian stable.

vrv:

If this passes all of our tested / supported platforms, it seems okay, but will let Gunhan review when he is free (he is out of the office for a bit though).  Or maybe @caisq  if he feels trongly

configure
```diff
@@ -456,7 +456,7 @@ while true; do
   fi
 
   if is_linux; then
-    CUDNN_PATH_FROM_LDCONFIG="$(ldconfig -p | sed -n 's/.*libcudnn.so .* => \(.*\)/\1/p')"
+    CUDNN_PATH_FROM_LDCONFIG="$(/sbin/ldconfig -p | sed -n 's/.*libcudnn.so .* => \(.*\)/\1/p')"
```
gunan: I feel like we need to be compatible with the previous case, where ldconfig is not under `/sbin`, but is in `PATH`.
For example, I have a centos system which is happy the way things are right now, but will break with this change as ldconfig resides in `/usr/sbin/ldconfig`
Can we modify these lines to say:
```
if ! type ldconfig > /dev/null; then
  LDCONFIG_BIN=/sbin/ldconfig
else
  LDCONFIG_BIN=ldconfig
fi
CUDNN_PATH_FROM_LDCONFIG="$($LDCONFIG_BIN -p | ......
```

byronyi:
@gunan Thanks for your help. I've changed the lines as suggested.

vrv:

jowagner:
An error message "configure: line xyz: type: ldconfig: not found" is printed on stderr if the command is not found in the line starting with "if ! type ldconfig".

There's a simple solution but as I cannot sign the CLA at this time, I cannot submit it. Hint: 5 characters.

I can confirm, however, that the variable LDCONFIG_BIN is initialised as intended to "/sbin/ldconfig" for my login without ldconfig in PATH.

Tested on openSUSE Leap 42.2.

byronyi:
@jowagner Thanks for the heads up. Not sure why some checks failed though.

gunan:
Jenkins, test this please.

vrv:
gpu build passed, so I'm calling this a win.

gunan:
Jenkins, test this please.

byronyi:
That's my first PR merged into TF! Thank you guys a lot @vrv @gunan @jowagner for your guide and suggestion. 

Really remarkable moment to me and I hope I could submit more patches in the future.

jowagner:
Thank you for incrementally coming up with a solution that is substantially different to my CLA-less PR and for taking the time to bring this over the finishing line.

PR #7291, opened 2017-02-06
Fix gradle task dependencies

jinqian:
Without the `afterEvaluate` closure, gradle can’t find the `assembleDebug` task.

See: https://code.google.com/p/android/issues/detail?id=219732#c32

### DIFF
--- a/tensorflow/examples/android/build.gradle
+++ b/tensorflow/examples/android/build.gradle
@@ -61,12 +61,13 @@ task buildNative(type:Exec) {
        '--host_crosstool_top=@bazel_tools//tools/cpp:toolchain'
 }
 
-task copyNativeLibs(type: Copy) {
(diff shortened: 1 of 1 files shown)


jinqian:
I signed it!

jinqian:
Should have to be submitted with the update of gradle tool version (to 2.2.x) at the same time!

ArtsiomCh:
Why that Issue is Closed? Code is not committed and still missed in master brunch: tensorflow/tensorflow/examples/android/build.gradle

@andrewharp have a look, please.

PS Got the same problem and found the same solution.

andrewharp:
@ArtsiomCh We actually have a CL in the pipeline internally that fixes this and a couple of other things, should be public in a day or two.

jinqian:
@ArtsiomCh @andrewharp Actually it's not a real issue for previous gradle tool version (pre-2.2), so I withdrew it.

PR #7298, opened 2017-02-06
Branch 146677928

benoitsteiner:


gunan (approved):

jhseu:
Hmm, was there some manual edits done on this change? `tensorflow/go/wrappers.go` which was added in cl/146572093.

That file shows up when I run movechanges.py manually, so I don't think it's part of the script. @asimshankar fyi

PR #7301, opened 2017-02-06
Bump the protobuf version to 3.2.0

jhseu:
Removes the 64 MB default limit.


gunan:
This is great!
We still need to build and upload our own protobuf packages for cpp implementations right?

jhseu:
Yep, after this is submitted, I plan to:

- Build protobuf 3.2.0 and upload it
- Update the documentation to point to the new versions on the website
- Update the version we depend on in the pip package for TF 1.1.

gunan (approved):

Issue #7289, opened 2017-02-06
tensorflow/core/platform/setround.cc:27:4: error: 'fesetround' is not a member of 'std'


carlorosati:
```
ERROR: /local/tmp/tensorflow/tensorflow/core/BUILD:1036:1: C++ compilation of rule '//tensorflow/core:lib_internal' failed: gcc failed: error executing command /local2/tmp/brew/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -B/local2/tmp/brew/bin -B/usr/bin -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 -O2 ... (remaining 104 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
tensorflow/core/platform/setround.cc: In constructor 'tensorflow::port::ScopedSetRound::ScopedSetRound()':
tensorflow/core/platform/setround.cc:27:4: error: 'fesetround' is not a member of 'std'
    std::fesetround(FE_TONEAREST);
    ^
tensorflow/core/platform/setround.cc:27:4: note: suggested alternative:
In file included from tensorflow/core/platform/setround.cc:19:0:
/local2/tmp/brew/include/fenv.h:88:12: note:   'fesetround'
 extern int fesetround (int __rounding_direction) __THROW;
            ^
Target //tensorflow/tools/pip_package:build_pip_package failed to build
Use --verbose_failures to see the command lines of failed build steps.
```

Looks like the issue is that this file is being compiled with gcc instead of g++

girving:
gcc vs. g++ is highly unlikely to be the reason.  What platform are you on, and what compiler are you using?

rmlarsen:
@girving FYI: I'm fixing this internally and will get it pushed to github. port::ScopedSetRound was kind of broken.

rmlarsen:
@carlorosati FYI: The fix was submitted internally at Google and should appear on Github in a day or two.

PR #7331, opened 2017-02-07
fesetround is not part of the C++ std library

carlorosati:
fesetround() is not part of the C++ std library. It is defined in the [fenv.h header in the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fesetround.html).

Addresses #7289

### DIFF
--- a/tensorflow/core/platform/setround.cc
+++ b/tensorflow/core/platform/setround.cc
@@ -24,7 +24,7 @@ namespace port {
 
 ScopedSetRound::ScopedSetRound() {
 #ifdef __STDC_IEC_559__
-   std::fesetround(FE_TONEAREST);
+   fesetround(FE_TONEAREST);
 #endif
 }
 


carlorosati:
i signed the Google CLA

rmlarsen:
@carlorosati Since C++11 it is defined in the std namespace via the header <cfenv>.

http://en.cppreference.com/w/cpp/numeric/fenv/feround

Can we change the code to use that header instead and get rid of the direct references to 
__STDC_IEC_559__? If not, I'm OK with your change.

carlorosati:
@rmlarsen Do you want me to remove the preprocessor statements? Do you know why they were needed?

rmlarsen:
Hm, ScopedSetRound looks a bit broken in general. It doesn't reset the rounding mode as stated in the header. This doesn't doesn't matter at the moment, but I'll fix and test it internally and push to github. Thanks for making me aware of this.

carlorosati:
@rmlarsen Be sure to close out #7289 with your commits.

rmlarsen:
@carlorosati will do, thanks. The change is running through automated testing. Should be pushed tomorrow or the day after.

PR #7334, opened 2017-02-07
Refer users to the RedBaron fork of this script.

aselle:
@machrisaa, has created a fork of tf_upgrade.py that uses RedBaron. For minimal dependency reasons, testing reasons and RedBaron's not full Python3 support, we aren't changing the officially supported version yet. In the mean time, we will refer users README.md that they can try it if our script finds a case it cannot handle.


aselle:
@gunan, could we cherry pick this into the next rc when it hits master?

gunan:
Sure, since it is only a readme change, I am merging it right away.
COuld you create the cherrypick PR?

gunan (approved):

aselle:
sure @gunan

PR #7165, opened 2017-01-31
Make curl forward/backward compatible

kchodorow:
I'm working on fixing https://github.com/bazelbuild/bazel/issues/1262, which will change the way Bazel organizes the execution root.  The curl library needs to use ../curl as an include prefix in new versions of Bazel and external/curl in old versions.

This patch should be forward/backward compatible.

PR #7338, opened 2017-02-07
Update our version of protocol buffers. Use a new numpy and pandas in Jenkins.

jhseu:


gunan:
Looks like at least the android breakage above is related.
Could you take a look?

jhseu:
(Hold off on reviewing/merging. Failures look genuine. Will investigate a bit later.)

jhseu:
@gunan PTAL. Turns out the pypi numpy causes subtle bugs with Ubuntu 14.04. pip install compiles from source.

jhseu:
Was testing upgrading numpy and pandas as part of this. Left them upgraded because they fix several unrelated issues.

jhseu:
Also upgraded to a protocol buffer version past 3.2.0 to support https://github.com/tensorflow/tensorflow/pull/7165. Only needed in bazel rules and not as a runtime dep.

gunan (approved):

gunan:
All test are happy, so I am happy.
Should we document which bugs numpy and pandas upgrades resolve?

PR #7342, opened 2017-02-07
Float indexing test fix for Numpy >= 1.12+

mkolod:
This fixes test failure for NumPy 1.12 that does not allow float indexing of arrays. Of course, it makes sense to index arrays using ints anyway, regardless of NumPy support. I assume this wasn't causing problems to people with NumPy up to 1.11.3, but it's better not to assume a specific NumPy version.


mkolod:
Do not merge yet - looks like there are other tests that depend on pre-NumPy 1.12 behavior. I'll fix those too.

gunan (approved):

PR #7347, opened 2017-02-08
Fixing example strided_slice

avloss:
Fixing broken example of tf.strided_slice
NEGATIVE strides work differently (should they?!)

tf.strided_slice(input, [1, 1, 0], [2, -1, 3], [1, -1, 1])


aselle:
The behavior is correct (matches python range semantics and numpy). The example was wrong. Thanks for the fix!

aselle (approved):

vrv:
skipping test since this is a simple doc change.

PR #7348, opened 2017-02-08
Do not build with sse4.1 by default.

### LINKED ISSUE #6809: TensorFlow 1.0.0 alpha binaries require SSE4.1
The TensorFlow 1.0.0 alpha binaries require SSE4.1, which was not necessary to run the binaries of earlier versions (e.g. 0.12.1). SSE4.1 is only supported by a few AMD architectures. In particular, it is not supported by the  "Barcelona" and the "Bobcat" architecture. They support SSE4a, which is not supported by Intel processors.

Therefore, running the 1.0.0 binaries on our workstations with "AMD Opteron(tm) Processor 6174" fails with the following error:
`The TensorFlow library was compiled to use SSE4.1 instructions, but these aren't available on your machine.`

Is it possible to make the official TensorFlow 1.0.0 binaries not dependent on vendor specific SSE4.1 instructions? I am aware that these AMD processors are not the newest ones and that it's possible to compile TensorFlow without the SSE4.1 requirement (at least, I assume so), but I wanted to raise this issue to see if you think it might be possible to make the official 1.0.0 binaries work without SSE4.1. Alternatively, would you consider offering alternative binaries without SSE4.1 similar to how you offer different binaries for different python versions?

gunan:
Fixes #6809

### DIFF
--- a/tensorflow/tensorflow.bzl
+++ b/tensorflow/tensorflow.bzl
@@ -97,7 +97,6 @@ def tf_copts():
            "-fno-exceptions",] +
           if_cuda(["-DGOOGLE_CUDA=1"]) +
           if_android_arm(["-mfpu=neon"]) +
-          if_x86(["-msse4.1"]) +
           select({
               "//tensorflow:android": [
                   "-std=c++11",


martinwicke (approved):
Sneaky Pete.

PR #7349, opened 2017-02-08
Do not build with sse4.1 by default.

### LINKED ISSUE #6809: TensorFlow 1.0.0 alpha binaries require SSE4.1

gunan:
Fixes #6809

### DIFF
--- a/tensorflow/tensorflow.bzl
+++ b/tensorflow/tensorflow.bzl
@@ -104,7 +104,7 @@ def tf_copts():
           if_cuda(["-DGOOGLE_CUDA=1"]) +
           if_mkl(["-DINTEL_MKL=1"]) +
           if_android_arm(["-mfpu=neon"]) +
-          if_x86(["-msse4.1"]) +
+          if_x86(["-msse3"]) +
           select({
               "//tensorflow:android": [
                   "-std=c++11",
@@ -118,7 +118,7 @@ def tf_copts():
                 "/DPLATFORM_WINDOWS",
                 "/DEIGEN_HAS_C99_MATH",
                 "/DTENSORFLOW_USE_EIGEN_THREADPOOL",
-		"/DEIGEN_VECTORIZE_SSE3",  # To flush denormals without __SSE3__ set.
+                "/DEIGEN_VECTORIZE_SSE3",  # To flush denormals without __SSE3__ set.
               ],
               "//tensorflow:ios": ["-std=c++11"],
               "//conditions:default": ["-pthread"]}))


petewarden (approved):

gunan:
@girving Disabling denormal_test again here, happy to help debug it if you need.

girving:
You're disabling it inside Google as well, which is bad.  If it needs to be disabled in open source, we should disable it *only* in open source.  What is the motivation for dropping SSE3?

gunan:
It should only be disabled for manual runs of `//tensorflow/...`
All internal automated systems should still run this test.

The motivation to drop SSE4.1 is #6809, AMD CPUs do not have SSE4.1.
But I think by dropping SSE4.1, we are dropping all previous versions of SSE too, and thus breaking this test?

girving:
It looks like all the architectures mentioned in that bug do supposed SSE3.  Could we leave SSE3 on even as we turn off SSE4?  It would be nice to have consistent denormal semantics, since it avoids truly horrific surprises for anyone for whom denormals matter.

martinwicke:
We can enable sse3. I suggest we do that not hidden away in an
unoverridable file, but in bazel rc. Ideally by setting the optimization
flags for our builds to -msse3.

gunan:
My concern with that is, someone can just `bazel build
tensorflow/something/to/build` and then use it.
Then denormals will fail.
Would it be possible to do something like this for denormals:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/hash/crc32c_accelerate.cc#L23

girving:
@gunan What's the equivalent of `#ifdef __SSE3__` in Python?  That aside, I really would like default usage of TensorFlow to flush denormals, including if someone downloads and builds from source.  It would terrify me if a source-built version of TensorFlow was by default different from the standard binaries.  It's not likely to hurt many people, but the people it does hurt will be very unhappy and confused.

gunan:
Modified the PR to replace sse4.1 with sse3.
PTAL

bhack:
Have we platform instructions set discovering at build time?

gunan:
I don't think we do at the moment. Bazel can only tell apart which platform we are targeting to build.
We might be able to write some skylark magic to do the detection.
But until then, this will quickly unblock AMD CPU owners.

bhack:
@gunan What do you think of Openblas [dynamic approach](https://github.com/xianyi/OpenBLAS/blob/develop/driver/others/dynamic.c)?

martinwicke:
We're looking at a (more) dynamic solution. @petewarden FYI

bhack:
Is xla involved? ;)

martinwicke:
XLA WILL SOLVE ALL OF THE PROBLEMS!

no, not this time. :)

bhack:
@martinwicke Please respect the XLA Graal ;)

sromano:
Is it the expected behavior to not support SSE4.1, SSE4.2, AVX, AVX2 and FMA instructions when installing tensorflow with pip through the official build in Mac? (Following the instructions for virutalenv at https://www.tensorflow.org/install/install_mac)

I'm using Mac Os. I was using tensorflow version r0.12 (installed through pip). Today I installed version r1.0 through pip using virtualenv instructions and I get these warnings:

`W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.1 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.2 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX2 instructions, but these are available on your machine and could speed up CPU computations.
W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use FMA instructions, but these are available on your machine and could speed up CPU computations.`

I understand I can build it from source to support them, but I'm reporting it because it might be a bug introduced after this change.

petewarden:
Thanks Sergio! We did initially try to default to at least SSE4.1 for the PIP builds, but we had reports that this caused problems for some users on older AMD machines, so we now default to the lower spec. This does mean that these installs are missing out on some CPU performance unfortunately, and we're still looking at better solutions around dynamically choosing code paths based on the current architecture. This will be a longer-term project though, unfortunately.

PR #7351, opened 2017-02-08
Add BUILD file and test case for "fact" user op

tntnatbry:
Current example op defined in tensorflow/core/user_ops/fact.cc is causing a "AttributeError: 'module' object has no attribute 'fact'" error using the following BUILD file and python code under tensorflow/core/user_ops:

BUILD file:
load("//tensorflow:tensorflow.bzl", "tf_custom_op_library")

tf_custom_op_library(
    name = "fact.so",
    srcs = ["fact.cc"],
)

Python code:
import tensorflow as tf
fact_module = tf.load_op_library('fact.so')
with tf.Session():
  print(fact_module.fact().eval())

Is this due to "fact" possibly being a reserved keyword in tensorflow? If yes, this PR will fix this issue by calling the op as my_fact instead of fact.

If it makes sense, I can also add the BUILD file and this python driver code under tensorflow/core/user_ops


vrv:
It is entirely possible!  Please go ahead and play around with what works -- this is supposed to just be an example of how to define custom ops, so the name isn't that important  :)

vrv:
@jart should we have this test be written importing the module dependencies to avoid the hourglass dependency?   As an 'example' of how to load a custom op, using the tf namespace makes more sense to copy-paste, but I'd like to get your opinion first.

tntnatbry:
@vrv thanks!
Seems like I missed some steps while building this op yesterday, the default implementation is working fine. I have however added a python module file and a test case for this op, along with a BUILD file (following the style guide)

vrv:
Also adding @keveman as reviewer since I'm not sure whether it's intentional to have not had a BUILD file (do we want to encourage BUILD files to build custom ops, or recommend just using gcc/clang instead like we also mention on our 'adding an op' guide?)

keveman:

keveman (approved):
Looks good. Will merge after tests pass.

vrv:

vrv:
sadly: ERROR: /workspace/tensorflow/BUILD:139:1: Target '//tensorflow/core/user_ops:all_files' is not visible from target '//tensorflow:all_opensource_files'. Check the visibility declaration of the former target if you think the dependency is legitimate.


@keveman what should they do?

tensorflow/core/user_ops/BUILD
```diff
@@ -0,0 +1,48 @@
...
+filegroup(
+    name = "all_files",
+    srcs = glob(
+        ["**/*"],
+        exclude = [
+            "**/METADATA",
+            "**/OWNERS",
+        ],
+    ),
+    visibility = ["//third_party/tensorflow:__subpackages__"],
```
keveman: This should be `visibility = ["//tensorflow:__subpackages__"],` ?

tntnatbry: @keveman Makes sense, updated the visibility.

tntnatbry:
Friendly ping! Can this be tested again?

tensorflow/core/user_ops/fact_op_test.py
```diff
@@ -0,0 +1,18 @@
+"""Test for fact op"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os.path
```
vrv: You can remove this I think

tensorflow/core/user_ops/fact_op.py
```diff
@@ -0,0 +1,12 @@
+"""Fact user op Python library."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os.path
```
vrv: import os, not os.path

(we prefer to import modules, not methods)

tensorflow/core/user_ops/fact_op_test.py
```diff
@@ -0,0 +1,18 @@
...
+
+import tensorflow as tf
+from tensorflow.core.user_ops import fact_op
+
+class FactTest(tf.test.TestCase):
+  def testFact(self):
+    with self.test_session():
+      self.assertAllEqual(fact_op.fact().eval(), "0! == 1")
+
```
vrv: can you add a linespace here?  (our internal linter likes to have two lines between top-level blocks)

tensorflow/core/user_ops/fact_op_test.py
```diff
@@ -0,0 +1,18 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os.path
+
+import tensorflow as tf
+from tensorflow.core.user_ops import fact_op
+
```
vrv: (same here: a line space)

vrv (changes requested):
Just a few minor things first.

tensorflow/core/user_ops/fact_op_test.py
tntnatbry: added.

tensorflow/core/user_ops/fact_op_test.py
tntnatbry: added.

tensorflow/core/user_ops/fact_op_test.py
tntnatbry: thanks, this was redundant.

tensorflow/core/user_ops/fact_op.py
tntnatbry: done.

tntnatbry:
@vrv I have made the changes based on your review.

vrv:

vrv:

dandelionmane:
@tntnatbry looks the following are broken in py3:
//tensorflow/core/ops/compat:backwards_compatibility_test
//tensorflow/core/user_ops:fact_op_test  
/tensorflow/examples/adding_an_op:fact_test  

https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu-python3/3619/console

Please fix

tntnatbry:
@dandelionmane @vrv Looks like this op is already a registered op in the tensorflow core op library, as it is listed [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/compat/ops_history.v1.pbtxt#L6396). Defining this op as a custom user op instead is therefore failing the following tests, as tensorflow is unable to find it as a registered op anymore:

//tensorflow/core/ops/compat:backwards_compatibility_test
//tensorflow/examples/adding_an_op:fact_test

Also since a test case already exists for this op here: /tensorflow/examples/adding_an_op/fact_test.py, there is no need to add a second test case.

Should we therefore close this PR?

vrv:
That's one option -- the other option is to rename this example op to a different name so it doesn't clash, but I guess then the adding an op documentation would go stale.  I'm going to let @keveman decide what he wants to do here.

martinwicke:
I'm going to close this. If @keveman decides it's better to make the changes, we can.

PR #7366, opened 2017-02-08
set correct target directory for tar extractall()

otzy:
fixes a wrong path when data_root is other than '.'


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

PR #7369, opened 2017-02-09
resize_image_with_crop_or_pad can work with batch of images

### LINKED ISSUE #7359: resize_image_with_crop_or_pad should work with batch of images
This is a repost of https://github.com/tensorflow/tensorflow/issues/2284 [per comment](https://github.com/tensorflow/tensorflow/issues/2284#issuecomment-277339057) from @mrry (CC @ziky90).

`resize_image_with_crop_or_pad` should be modified to take either a single image or a batch of images. Right now the signature of the function specifies:
```
    image: 3-D tensor of shape `[height, width, channels]`
```

For a reference of what the api should look like, see [`resize_images(images,....)`](https://github.com/tensorflow/tensorflow/blob/4c192f060cf9ff897911d240c140299d6db257b6/tensorflow/python/ops/image_ops_impl.py#L605-L608):
```
    images: 4-D Tensor of shape `[batch, height, width, channels]` or
            3-D Tensor of shape `[height, width, channels]`.
```

Some commentary:
* docs for `resize_images` say: " Resized images will be distorted if their original aspect ratio is not the same as `size`.  To avoid distortions see [`resize_image_with_crop_or_pad`]." However the method referenced _cannot_ be currently used in some cases for which `resize_images` works due to the above mentioned limitations.
* the entire operation can be implemented using slice operations on the tensor batch. This is what `resize_image_with_crop_or_pad` does but it has the following logic which would be nice not to copy and paste (there is some subtlety with round down vs up):
```python
  width_diff = target_width - width
  offset_crop_width = max_(-width_diff // 2, 0)
  offset_pad_width = max_(width_diff // 2, 0)

  height_diff = target_height - height
  offset_crop_height = max_(-height_diff // 2, 0)
  offset_pad_height = max_(height_diff // 2, 0)
```
* the fact this one image op method does not handle batches leads to confusion and various [SO](http://stackoverflow.com/questions/33944683/tensorflow-map-operation-for-tensor) posts. People even resort to solving the problem using all sorts of crazy solutions (e.g. `tf.while_loop`).
* this method is useful in combination with concatenation operations. For example see the[ Lasagne `ConcatLayer`](http://lasagne.readthedocs.io/en/latest/modules/layers/merge.html#lasagne.layers.ConcatLayer).

cancan101:
Closes #7359

### DIFF
--- a/tensorflow/python/ops/image_ops_impl.py
+++ b/tensorflow/python/ops/image_ops_impl.py
@@ -90,42 +90,23 @@ def _is_tensor(x):
   return isinstance(x, (ops.Tensor, variables.Variable))
 
 
-def _ImageDimensions(image):
+def _ImageDimensions(image, rank):
   """Returns the dimensions of an image tensor.
 
   Args:
-    image: A 3-D Tensor of shape `[height, width, channels]`.
+    image: A rank-D Tensor. For 3-D  of shape: `[height, width, channels]`.
+    rank: The expected rank of the image
 
   Returns:
-    A list of `[height, width, channels]` corresponding to the dimensions of the
+    A list of corresponding to the dimensions of the
     input image.  Dimensions that are statically known are python integers,
     otherwise they are integer scalar tensors.
   """
   if image.get_shape().is_fully_defined():
     return image.get_shape().as_list()
   else:
-    static_shape = image.get_shape().with_rank(3).as_list()
-    dynamic_shape = array_ops.unstack(array_ops.shape(image), 3)
-    return [s if s is not None else d
-            for s, d in zip(static_shape, dynamic_shape)]
-
-
-def _BatchImageDimensions(image):
-  """Returns the dimensions of an batch of images tensor.
-
-  Args:
-    image: A 4-D Tensor of shape `[batch, height, width, channels]`.
-
-  Returns:
-    A list of `[batch, height, width, channels]` corresponding to the dimensions of the
-    input image.  Dimensions that are statically known are python integers,
-    otherwise they are integer scalar tensors.
-  """
-  if image.get_shape().is_fully_defined():
-    return image.get_shape().as_list()
-  else:
-    static_shape = image.get_shape().with_rank(4).as_list()
-    dynamic_shape = array_ops.unpack(array_ops.shape(image), 4)
+    static_shape = image.get_shape().with_rank(rank).as_list()
+    dynamic_shape = array_ops.unstack(array_ops.shape(image), rank)
     return [s if s is not None else d
             for s, d in zip(static_shape, dynamic_shape)]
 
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/image_ops_impl.py
```diff
@@ -412,9 +416,17 @@ def pad_to_bounding_box(image, offset_height, offset_width, target_height,
...
+  assert_ops += _CheckAtLeast3DImage(image, require_static=False)
+
+  is_batch = True
+  if image.get_shape().ndims == 3:
+    is_batch = False
+    image = array_ops.expand_dims(image, 0)
+  elif image.get_shape().ndims != 4:
+    raise ValueError('\'image\' must have either 3 or 4 dimensions.')
+
+  num_channels, height, width, depth = image.get_shape().as_list()
```
cancan101: this seems like it might be problematic for dynamic sizes. Perhaps I should add a `_ImageDimensions` that works in higher dimensions?

tensorflow/python/ops/image_ops_impl.py
```diff
@@ -110,6 +110,26 @@ def _ImageDimensions(image):
             for s, d in zip(static_shape, dynamic_shape)]
 
 
+def _BatchImageDimensions(image):
```
gpapan: There is a lot of code duplication between this function and `_ImageDimensions(image)` above.
Can you please unify them by adding a `rank` argument to `_ImageDimensions()`?
Then you can call it with rank=3 or rank=4.

tensorflow/python/ops/image_ops_impl.py
```diff
@@ -110,6 +110,26 @@ def _ImageDimensions(image):
...
+  if image.get_shape().is_fully_defined():
+    return image.get_shape().as_list()
+  else:
+    static_shape = image.get_shape().with_rank(4).as_list()
+    dynamic_shape = array_ops.unpack(array_ops.shape(image), 4)
+    return [s if s is not None else d
+            for s, d in zip(static_shape, dynamic_shape)]
+
+
 def _Check3DImage(image, require_static=True):
```
gpapan: Similarly here, can you please rename `_Check3DImage()`  to `_GetImageShapeWithRankMaybeRequireStatic(image, rank, require_static=True)`.

gpapan (changes requested):

tensorflow/python/ops/image_ops_impl.py
cancan101: does this name make sense? This function doesn't "get an image shape" rather it:
>  Returns An empty list, if `image` has fully defined dimensions. Otherwise, a list containing an assert op is returned.

cancan101: @gpapan ping on calling it `_GetImageShapeWithRankMaybeRequireStatic`?

cancan101:
@gpapan ping on calling it _GetImageShapeWithRankMaybeRequireStatic?

dandelionmane:
@gpapan PTAL?

gpapan (approved):

dandelionmane:

cancan101:
will tests run again?

jhseu:
Jenkins, test this please

jhseu:
Mind fixing the test failure? Search for the word "failed" here:
https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu/4077/consoleFull

jhseu:
Jenkins, test this please

PR #7371, opened 2017-02-09
docs: add  a new 0.7.0 release with TF 0.12.1

Kjeanclaude:
docs: add  a new 0.7.0 release with TF 0.12.1 in the History part.

### DIFF
--- a/tensorflow/examples/udacity/README.md
+++ b/tensorflow/examples/udacity/README.md
@@ -109,3 +109,4 @@ History
 * 0.4.0: Move notMMNIST data for Google Cloud.
 * 0.5.0: Actually use 0.7.1 release.
 * 0.6.0: Update to TF 0.10.0, add libjpeg (for Pillow).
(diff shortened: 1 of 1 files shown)


vincentvanhoucke:
The container that I have tagged 0.7.0 right now does not work, because it picks up the notebooks at the head of the tree which are not compatible with 0.12.1. I am waiting for a new TF container to be created.

Kjeanclaude:
Hi !
The 0.7.0 tag version is what I actually use inside docker **with the udacity project** and it is compatible with the version 0.12.1 of tensorflow [**print(tf.__version__)** in the notebook].
With this version the last update of initializer : **tf.global_variables_initializer()** is applied but not with the 0.6.0 tag.
So if you release another version, please udpate the readme so that everyone be informed, the new initializer is faster.

vincentvanhoucke:
@Kjeanclaude assignment 6 will not work under current 0.7.0

PR #7387, opened 2017-02-09
Update shape checking logic in einsum

jihunchoi:
In #6830, I modified einsum function to check whether new shape has two or more unknown dimensions.
However, according to the comment on #6824, it is proven to be not sufficient to check only the last reshape call.
This PR fixes the issue.

### DIFF
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -370,7 +370,7 @@ def _einsum_reduction(t0, t0_axis_labels, t1, t1_axis_labels, axes_to_sum):
         t0_shape[:len(preserved_axes)+len(broadcast_axes[0])] +
         t1_shape[len(t1_shape)-len(broadcast_axes[1]):]
     )
-    # Make new_shape valid, if invalid to be used in reshape
+    # Make uncompacted_shape valid, if invalid to be used in reshape
     none_dims = [i for i, d in enumerate(uncompacted_shape) if d is None]
     if len(none_dims) > 1:
       uncompacted_shape = list(uncompacted_shape)
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -320,7 +320,7 @@ def _einsum_reduction(t0, t0_axis_labels, t1, t1_axis_labels, axes_to_sum):
 
     t0_shape = _get_shape(t0)
     num_broadcast_elements_t0 = _total_size(
-      t0_shape[len(preserved_axes):-len(axes_to_sum)])
+        t0_shape[len(preserved_axes):-len(axes_to_sum)])
     num_summed_elements = _total_size(t0_shape[-len(axes_to_sum):])
     new_shape = (t0_shape[:len(preserved_axes)]
                  + [num_broadcast_elements_t0, num_summed_elements])
@@ -328,7 +328,7 @@ def _einsum_reduction(t0, t0_axis_labels, t1, t1_axis_labels, axes_to_sum):
 
     t1_shape = _get_shape(t1)
     num_broadcast_elements_t1 = _total_size(
-      t1_shape[len(preserved_axes)+len(axes_to_sum):])
+        t1_shape[len(preserved_axes)+len(axes_to_sum):])
     new_shape = (t1_shape[:len(preserved_axes)]
                  + [num_summed_elements, num_broadcast_elements_t1])
     t1 = _reshape_if_necessary(t1, new_shape)
@@ -337,8 +337,8 @@ def _einsum_reduction(t0, t0_axis_labels, t1, t1_axis_labels, axes_to_sum):
 
     # Undo compaction of broadcast axes
     uncompacted_shape = (
-      t0_shape[:len(preserved_axes)+len(broadcast_axes[0])]
-      + t1_shape[len(t1_shape)-len(broadcast_axes[1]):]
+        t0_shape[:len(preserved_axes)+len(broadcast_axes[0])]
+        + t1_shape[len(t1_shape)-len(broadcast_axes[1]):]
     )
     product = _reshape_if_necessary(product, uncompacted_shape)
 
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -318,70 +318,28 @@ def _einsum_reduction(t0, t0_axis_labels, t1, t1_axis_labels, axes_to_sum):
     # into a single axis, and combine multiple summed axes into a
     # single axis.
 
-    t0_shape = tuple(x.value for x in t0.get_shape())
+    t0_shape = _get_shape(t0)
     num_broadcast_elements_t0 = _total_size(
-        t0_shape[len(preserved_axes):-len(axes_to_sum)])
-    num_summed_elements = _total_size(t0_shape[-len(axes_to_sum):])
-    new_shape = t0_shape[:len(preserved_axes)] + (num_broadcast_elements_t0,
-                                                  num_summed_elements)
-    # Make new_shape valid, if invalid to be used in reshape
-    none_dims = [i for i, d in enumerate(new_shape)
-                 if d is None or d == -1]
-    if len(none_dims) > 1:
-      new_shape = list(new_shape)
-      for i in none_dims[:-1]:
-        if i < len(preserved_axes):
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/special_math_ops.py
```diff
@@ -324,13 +324,43 @@ def sort_key(input_index, a):
     num_summed_elements = _total_size(t0_shape[-len(axes_to_sum):])
     new_shape = t0_shape[:len(preserved_axes)] + (num_broadcast_elements_t0,
                                                   num_summed_elements)
+    # Make new_shape valid, if invalid to be used in reshape
+    none_dims = [i for i, d in enumerate(new_shape)
```
vrv: there is quite a bit of duplicated code here and in the block below -- is it possible to factor it out into a helper function?

tensorflow/python/ops/special_math_ops.py
```diff
@@ -324,13 +324,43 @@ def sort_key(input_index, a):
...
 
     t1_shape = tuple(x.value for x in t1.get_shape())
     num_broadcast_elements_t1 = _total_size(
         t1_shape[len(preserved_axes)+len(axes_to_sum):])
     new_shape = t1_shape[:len(preserved_axes)] + (num_summed_elements,
                                                   num_broadcast_elements_t1)
+    # Make new_shape valid, if invalid to be used in reshape
+    none_dims = [i for i, d in enumerate(new_shape)
+                 if d is None or d == -1]
+    if len(none_dims) > 1:
```
vrv: it would be nice to have more comments about what it means to 'make new_shape valid' in order to guide the code reader about what is being done here.

tensorflow/python/ops/special_math_ops.py
jihunchoi: I've tried to, however I found that it is not simple task to abstract the functionalities, since the newly added components require information about different variables.
In line 327-341, it queries only the shape of `t0`, however in line 349-363, it queries both the shapes of `t0` and `t1`.
Finally, in line 373-383, it needs neither `t0` or `t1`, but the original shapes of them (before reshape operations).
If you have an idea to factor the codes out, please let me know.

tensorflow/python/ops/special_math_ops.py
jihunchoi: Okay. I will add more comments soon!

tensorflow/python/ops/special_math_ops.py
mstreeter: Rather than defining new_shape in terms of get_shape() values (on line 325) and then fixing it here if necessary, could we just define new_shape in terms of array_ops.shape initially?  In other words, something like:

t0_shape = array_ops.shape(t0)
t1_shape = array_ops.shape(t1)
num_summed_elements = math_ops.reduce_prod(t0_shape[len(preserved_axes):-len(axes_to_sum)])
num_broadcast_elements_t1 = math.ops.reduce_prod(t0_shape[-len(axes_to_sum):])
new_shape = t1_shape[:len(preserved_axes)] + (num_summed_elements, num_broadcast_elements_t1)

You could do the same for lines 344-348, and then remove the _total_size() method.  It seems like this would also let you remove the logic for fixing uncompacted_shape below.

mstreeter (changes requested):

tensorflow/python/ops/special_math_ops.py
jihunchoi: That would be the simplest way, however the reason I tried to avoid fixing like that way is that it might cause some performance overhead.
As far as I understood, if one can determine the shape by just using `get_shape`, then node creation caused by `array_ops.shape` does not occur, thus there exists a possibility to gain performance a bit.
I don't know how large the overhead of adding `array_ops.shape` node to the graph is. If it is negligible, then replacing `get_shape` with `array_ops.shape` would be the simplest and the cleanest way to fix this issue.

mstreeter: Good point.  I suspect the overhead is negligible, but I haven't verified this.

I would be fine with always using array_ops.shape to keep the code simple.  Alternatively, to keep the number of ops small, you could create a _get_shape helper that either returns array_ops.shape or returns a tuple of integers when possible.  You could also modify _total_size so that it returns an integer if given a tuple of integers, and uses math_ops.reduce_prod otherwise.

It seems like this would get the best of both worlds, but let me know if I've missed something.

vrv: TF implements constant folding, which means that if a shape() node that exists in the graph can be computed statically from constants, it will be done once during the first run and never again for future computations, so concerns about performance here may not matter.

jihunchoi: That sounds great. Then I will update the codes in that way!

jihunchoi:
I updated the codes as you suggested, but there are some differences to keep its coverage on shape checks before evaluated.
Specifically, I modified the codes in the below ways:
1) Instead of calling `get_shape()`, I added `_get_shape()` helper function that returns a result of `get_shape().to_list()` if the result does not contain `None` values, and else the helper function explicitly queries the dimension value using `array_ops.shape` function.
2) `_total_size` helper function is not deleted but modified a bit. Previously it returned `-1` if the shape contains `None` dimensions. I modified the function not to return `-1` even when some of dimensions are unknown.
If the shape is fully known, it returns an integer, and if the shape contains unknown dimensions, it returns a scalar tensor.
The reason I didn't obtain the total size by using just `math_ops.reduce_prod` is to keep its capability to check the shape before actually evaluated if available. (See `test_dim_mismatch` test.)

tensorflow/python/ops/special_math_ops.py
```diff
@@ -318,44 +318,28 @@ def sort_key(input_index, a):
...
     # single axis.
 
-    t0_shape = tuple(x.value for x in t0.get_shape())
+    t0_shape = _get_shape(t0)
     num_broadcast_elements_t0 = _total_size(
-        t0_shape[len(preserved_axes):-len(axes_to_sum)])
-    num_summed_elements = _total_size(t0_shape[-len(axes_to_sum):])
-    new_shape = t0_shape[:len(preserved_axes)] + (num_broadcast_elements_t0,
-                                                  num_summed_elements)
+      t0_shape[len(preserved_axes):-len(axes_to_sum)] + [1])
```
mstreeter: Minor: is adding [1] necessary here?  _total_size returns 1 if given an empty list.  If for some reason this is still necessary, I think it would be cleaner to do result *= 1 at the end of _total_size.

Same question on line 331.

mstreeter (approved):

mstreeter:
Thanks for fixing this!

tensorflow/python/ops/special_math_ops.py
jihunchoi: That's surely my oversight, I forgot to remove them.

tensorflow/python/ops/special_math_ops.py
```diff
@@ -320,15 +320,15 @@ def sort_key(input_index, a):
...
+      t0_shape[len(preserved_axes):-len(axes_to_sum)])
+    num_summed_elements = _total_size(t0_shape[-len(axes_to_sum):])
     new_shape = (t0_shape[:len(preserved_axes)]
                  + [num_broadcast_elements_t0, num_summed_elements])
     t0 = _reshape_if_necessary(t0, new_shape)
 
     t1_shape = _get_shape(t1)
     num_broadcast_elements_t1 = _total_size(
-        t1_shape[len(preserved_axes)+len(axes_to_sum):] + [1])
+      t1_shape[len(preserved_axes)+len(axes_to_sum):])
```
vrv: p.s. I think your indention was right before (4 space continuation)

jihunchoi: I fixed my codes to use 4 spaces in line continuation.
By the way, the TensorFlow indentation guideline is always confusing to me; is there any handy way to check code style locally?

vrv: not that I know of @yifeif might know more about possible pylint options.

We use two-space indent but 4-space continuations, that's about all I know :)

Thanks for doing this

vrv:

tensorflow/python/ops/special_math_ops.py
yifeif: Our pylintrc file is located at [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/pylintrc). You can pass this file to python with --rcfile option.

PR #7395, opened 2017-02-09
Remove leading slash from data directory

geromueller:
To allow embedding of TensorBoard, the data directories need to be relative.


vrv:
@geromueller  are you able to sign the CLA?

geromueller:
I consent to contribute to this project with my commits

vrv:
I think you have to sign the CLA using "post@geromueller.de" and then comment here, since that's the github email you used.

dandelionmane:
I already made this change internally on the src file (router.ts), so it will already be fixed when we make the next update of the generated tf-tensorboard.html (should be this week).

PR #7396, opened 2017-02-09
Branch 147051664

benoitsteiner:


tensorflow/third_party/mkl/BUILD
```diff
@@ -0,0 +1,9 @@
+# This file is replaced on github with a real build rule.
```
cwhipkey: hmm, it looks like this is one level too low - it should be third_party/mkl/BUILD ?  same for third_party/mkl/build_defs.bzl

third_party/eigen3/BUILD
```diff
@@ -28,6 +30,7 @@ cc_library(
         "unsupported/Eigen/CXX11/Tensor",
         "unsupported/Eigen/CXX11/FixedPoint",
     ],
+    includes = if_mkl(["./mkl_include"]),
     visibility = ["//visibility:public"],
 # INTEL_MKL start
 		includes = if_mkl(["./mkl_include"]),
```
cwhipkey: the includes= is repeated again here.  I had removed the INTEL_MKL_start/end comments.

third_party/mkl/build_defs.bzl
```diff
@@ -1,13 +1,13 @@
...
-
-    Returns a select statement which evaluates to if_true if we're building
-    with MKL enabled.  Otherwise, the select statement evaluates to if_false.
-
-    """
-    return select({
-        "//third_party/mkl:using_mkl": if_true,
-        "//conditions:default": if_false
-    })
+# Macros for building MKL code.
```
cwhipkey: I can't tell what this diff is.  what is it saysing?

tensorflow/third_party/mkl/BUILD
benoitsteiner: Indeed. I kept the ones that reside in third_party/mkl, even though they won't work since the .so files are missing.

third_party/mkl/build_defs.bzl
benoitsteiner: Beats me. The code is the same before and after !

gunan (approved):

gunan:
Makefile issue is a transient failure.
python3 test is a known issue which is rolled back internally.

PR #7400, opened 2017-02-09
Better Variable.__repr__

### LINKED ISSUE #6539: better Variable.__repr__
Right now, I get sth like:

    <tensorflow.python.ops.variables.Variable object at 0x2b9c67daae10>

It would be nice if it would show sth like

    <Variable name="foo" shape=(100,200) dtype="float32">

or so.

irapha:
Closes #6539


tensorflow/python/ops/variables.py
```diff
@@ -226,7 +226,12 @@ def __init__(self,
           expected_shape=expected_shape)
 
   def __str__(self):
-    return str(self._snapshot)
+    return "<Variable name='%s' shape=%s dtype=%s>" % (
+            self.name(), self.get_shape(), self.dtype().name)
```
mrry: For consistency with the repr for `tf.Tensor`, can you please change the format string to:

`<tf.Variable '%s' shape=%s dtype=%s>`

Also, I think it should be `self.dtype` rather than `self.dtype()`.

tensorflow/python/ops/variables.py
```diff
@@ -226,7 +226,12 @@ def __init__(self,
...
+
+  def __repr__(self):
+    return "<Variable name='%s' shape=%s dtype=%s>" % (
+            self.name(), self.get_shape(), self.dtype().name)
```
mrry: Return `self.__str__()` here (or vice versa) to reduce duplicated code.

mrry (changes requested):

tensorflow/python/ops/variables.py
iRapha: Done

tensorflow/python/ops/variables.py
iRapha: Done, except for the second part. The reason I did `self.dtype().name` is because of this method in Variable: https://github.com/iRapha/tensorflow/blob/ab88819372cd83cb3db7fa5419946e3e4846ba29/tensorflow/python/ops/variables.py#L745

mrry: `Variable.dtype` is a property, not a method though, so I think `self.dtype` is required.

Can you also remove `name=` from the repr for consistency with the other reprs?

iRapha: Ah okay, I overlooked that. Done! :)

tensorflow/python/ops/variables.py
```diff
@@ -226,7 +226,11 @@ def __init__(self,
           expected_shape=expected_shape)
 
   def __str__(self):
-    return str(self._snapshot)
+    return self.__repr_()
+
+  def __repr__(self):
+    return "<tf.Variable '%s' shape=%s dtype=%s>" % (
+            self.name(), self.get_shape(), self.dtype.name)
```
mrry: Thanks for making that change. I just noticed that it should be `self.name` instead of `self.name()` (for the same reasons as `self.dtype`...). Can you please make that change as well?

This little piece of code has proven surprisingly tricky for such a small change! I think it might be a good idea to add a unit test, which would have caught these issues. Can you please add a test of `str(tf.Variable(...))` and `repr(tf.Variable(...))` to `variables_test.py`? You can base it on the similar tests for `tf.Operation` [here](https://github.com/tensorflow/tensorflow/blob/286a5361df528dd5d5f1c5b319a69a0029e245bc/tensorflow/python/framework/ops_test.py#L344).

tensorflow/python/ops/variables.py
```diff
@@ -226,7 +226,11 @@ def __init__(self,
           expected_shape=expected_shape)
 
   def __str__(self):
-    return str(self._snapshot)
+    return self.__repr_()
```
shoyer: You actually don't need a `__str__` method at all if it should be the same as  `__repr__`.

mrry: That's good to know! @iRapha, could you please remove the unnecessary `__str__()` method? Thanks!

iRapha: Done!

As a note, I've read elsewhere that the difference between them is that `__repr__` is supposed to be more descriptive, and often times copy-pastable as python code, whereas `__str__` is meant to be more readable. In this case, it means that we could do:

```python
>>> variable.__str__()
<tf.Variable 'name' shape=(10, 2) dtype=float32>
>>>variable.__repr__()
tf.Variable()
```

The reason I didn't try doing this in the original version of the PR is because
a) I assume this would be a much bigger change than what the issue was asking for and
b) I found that this wasn't consistent within tf:

```python
>>> tf.float32.__str__()
<dtype: 'float32'>
>>> tf.float32.__repr__()
tf.float32

>>> tf.constant(0.1).__str__()
Tensor("Const:0", shape=(), dtype=float32)
>>> tf.constant(0.1).__repr__()
<tf.Tensor 'Const:0' shape=() dtype=float32>
```

tensorflow/python/ops/variables.py
iRapha: Done! Sorry about the delay; bazel run super slow when your other computer isn't a datacenter :P

One quick thing about the test: I couldn't figure out why the dtype name has a `_ref` postfix when testing

mrry (approved):

mrry:
Looks good to me, thanks!

PR #7421, opened 2017-02-10
Change mismatched ref input error message

### LINKED ISSUE #202: Better error message for tf.assign
When I try to use tf.assign I'm always running into the error message

TypeError: Input 'ref' of 'Assign' Op requires l-value input

I have no idea what this means. It's a pretty unhelpful error message.

Some context ...

```
a = tf.random_normal([4,4]) # I tried this with tf.constant and tf.Variable and got the same result

tf.assign(tf.slice(a, [1, 0], [-1, -1]), tf.slice(a, [0, 0], [3, -1]))

tf.assign(a, 5)

tf.assign(a, a)
```

All three give the same error message.

irapha:
Closes #202

### DIFF
--- a/tensorflow/python/framework/op_def_library.py
+++ b/tensorflow/python/framework/op_def_library.py
@@ -618,8 +618,8 @@ class OpDefLibrary(object):
         if input_arg.is_ref:
           if not all(x._is_ref_dtype for x in types):  # pylint: disable=protected-access
             raise TypeError(
-                "'%s' Op requires that input '%s' be tf.Variable" %
-                (op_type_name, input_name))
+                ("'%s' Op requires that input '%s' be a mutable tensor " +
+                "(e.g.: a tf.Variable)") % (op_type_name, input_name))
           input_types.extend(types)
         else:
           input_types.extend(base_types)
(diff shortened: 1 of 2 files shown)


tensorflow/python/framework/op_def_library.py
```diff
@@ -618,8 +618,8 @@ def apply_op(self, op_type_name, name=None, **keywords):
         if input_arg.is_ref:
           if not all(x._is_ref_dtype for x in types):  # pylint: disable=protected-access
             raise TypeError(
-                "Input '%s' of '%s' Op requires l-value input" %
-                (input_name, op_type_name))
+                "'%s' Op requires that input '%s' be tf.Variable" %
+                (op_type_name, input_name))
```
mrry: I think this error message could be misleading. Sure, `tf.Variable` is the most common "l-value input", but you can also pass the result of `tf.assign()` to another `tf.assign()` as the l-value, in order to chain updates.

How about instead making the error message:

"'%s' op requires that input '%s' be a mutable tensor (e.g a tf.Variable)"

mrry (changes requested):

tensorflow/python/framework/op_def_library.py
iRapha: Done!

mrry (approved):
LGTM!

mrry:
@tensorflow-jenkins  test this please.

PR #7426, opened 2017-02-11
CP 1.0

yifeif:
Fix double evaluation of macro argument that was causing duplicate CUDA batched GEMM calls.

Change: 147025110

(cherry picked from commit f439016532a8b310bb4cc984eed1c331b4d3be8a)


martinwicke (approved):
This is still dangerous (thanks for pointing that out, @girving), but it's better than before.

girving:
In particular, it's dangerous for future usage but (probably) isn't causing any bugs right now.

Issue #7375, opened 2017-02-09
tensorboard show empty page (master)


chihyaoma:
### What related GitHub issues or StackOverflow threads have you found by searching the web for your problem?

[#5341](https://github.com/tensorflow/tensorflow/issues/5341)

### Environment info
Ubuntu 14.04 LTS. 
git clone https://github.com/tensorflow/tensorflow
./configure
bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package
bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
pip install /tmp/tensorflow_pkg/tensorflow-1.0.0rc1-cp35-cp35m-linux_x86_64.whl 

The output from `python -c "import tensorflow; print(tensorflow.__version__)"`.
1.0.0-rc1

If installed from source, provide 

1. The commit hash (`git rev-parse HEAD`)
bd4a1c58d28734c77ecb17d719e72c45c6c33077
2. The output of `bazel version`
Build label: 0.4.4

### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)
I use a tensorboard example that I think is quite easy to reproduce the issue. 
```shell
git clone https://github.com/normanheckscher/mnist-tensorboard-embeddings
cd mnist-tensorboard-embeddings/

# modify the path at line 97 and 99 in mnist_t-sne.py
python mnist_t-sne.py
# start tensorboard
tensorboard --logdir=/home/chih-yao/Documents/mnist-tensorboard-embeddings/logs/
```
### What other attempted solutions have you tried?
- [x] Make sure my `tensorboard --logdir=/home/chih-yao/Documents/mnist-tensorboard-embeddings/logs/` path is correct
- [x] Reinstall tensorflow from source
- [x] Make sure the metadata and event files exist in the `logdir`

### Logs or other output that would be helpful
```
tensorboard --logdir=/home/chih-yao/Documents/mnist-tensorboard-embeddings-master/logs
Starting TensorBoard b'46' on port 6006
(You can navigate to http://127.0.1.1:6006)
 * Running on http://0.0.0.0:6006/ (Press CTRL+C to quit)
127.0.0.1 - - [08/Feb/2017 21:06:07] "GET /webcomponentsjs/webcomponents-lite.min.js HTTP/1.1" 200 -
127.0.0.1 - - [08/Feb/2017 21:06:08] "GET /lib/css/global.css HTTP/1.1" 200 -
[...]

PR #7431, opened 2017-02-11
Fix regression in TensorBoard

jart:
TensorBoard is currently broken. This fixes it. See #7375.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -480,7 +480,7 @@ new_http_archive(
 )
 
 new_http_archive(
-  name = "polymer_archive",
+  name = "polymer",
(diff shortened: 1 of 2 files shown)


gunan (approved):

gunan:
Approved and merging as users tested this and it is reported to be working.
Just for my curiosity, what was the problem with the old name of the repo? was there a collision in name?

jart:
In a CL I changed the name to that thing when I shouldn't have, because I ended up changing the new webfiles repository to be org_polymer instead of what I was originally going to name it: polymer.

gunan:
Thanks for the clarification, and the fix!

PR #7447, opened 2017-02-12
update

jsonkey:


gunan:
Is this PR created by mistake, by any chance?
I do not think we can accept this change, but just to make sure,
cc/ @josh11b @asimshankar @skye

asimshankar:
I'm guessing this was accidentally created too.
If I'm wrong, @jsonkey , would appreciate some discussion on what you're trying to achieve here and the best way to do so (for example, we'd want to avoid enhancements to `TF_DeprecatedSession` and instead have changes focused on `TF_Session`).

PR #7471, opened 2017-02-13
Branch 147365276

vrv:


jhseu:
Sorry, mind redoing the push? We need at least cl/147369186 so that we don't break the build on g++ 5/6

jhseu (changes requested):
.

vrv:
yeah there are two failures.  I'll repush after they are fixed.

PR #7475, opened 2017-02-13
Replace deprecated tf.op_scope by tf.name_scope

JarnoRFB:
Since tf.op_scope is deprecated (see https://www.tensorflow.org/api_docs/python/framework/defining_new_operations#op_scope) it should probably be removed from the style guide.


JarnoRFB:
I signed it!

PR #7481, opened 2017-02-14
Hide platform/logging.h from core/public

### LINKED ISSUE #7480: preprocessor definition clash with glog
CHECK macros from `platform/logging.h` leak out into `core/public` headers which clash with users of glog.

One path is through `core/platform/allocator.h`:
```
In file included from external/org_tensorflow/tensorflow/core/platform/logging.h:25:0,
                 from external/org_tensorflow/tensorflow/core/framework/allocator.h:26,
                 from external/org_tensorflow/tensorflow/core/framework/tensor.h:21,
                 from external/org_tensorflow/tensorflow/core/public/session.h:23,
(snip)
external/org_tensorflow/tensorflow/core/platform/default/logging.h:224:0: note: this is the location of the previous definition
 #define CHECK_OP_LOG(name, op, val1, val2)                            \
 ^
```
This one is easy to fix by moving method implementation to allocator.cc.

Another is through `core/lib/core/status.h`.
```
In file included from external/org_tensorflow/tensorflow/core/platform/logging.h:25:0,
                 from external/org_tensorflow/tensorflow/core/lib/core/status.h:24,
                 from external/org_tensorflow/tensorflow/core/lib/core/errors.h:19,
                 from external/org_tensorflow/tensorflow/core/framework/tensor_shape.h:24,
                 from external/org_tensorflow/tensorflow/core/framework/tensor.h:24,
                 from external/org_tensorflow/tensorflow/core/public/session.h:23,
```

This one is more work to fix because `TF_CHECK_OK` is used all over the code, but it does not seem to be necessary for `core/public`.

szym:
Fixes #7480

### DIFF
--- a/tensorflow/core/common_runtime/bfc_allocator.h
+++ b/tensorflow/core/common_runtime/bfc_allocator.h
@@ -30,6 +30,7 @@ limitations under the License.
 #include "tensorflow/core/platform/thread_annotations.h"
 #include "tensorflow/core/platform/types.h"
 #include "tensorflow/core/protobuf/config.pb.h"
+#include "tensorflow/core/platform/logging.h"
 
 namespace tensorflow {
 
--- a/tensorflow/core/framework/allocator.cc
+++ b/tensorflow/core/framework/allocator.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include "tensorflow/core/framework/log_memory.h"
 #include "tensorflow/core/framework/tracking_allocator.h"
 #include "tensorflow/core/lib/strings/stringprintf.h"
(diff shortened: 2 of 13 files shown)


szym:
It looks like it's going to be difficult to do this due to use of DCHECKs in array_slice_internal.h : `ArraySliceImplBase` which is a template class.

vrv:
Thanks, closing this for now -- please feel free to send another PR when you can sign the CLA.

I'm also uncertain what our API guarantees are for this function -- it's possible we can't remove this or we'll break our public API.  Perhaps we should follow up on the issue thread more.

PR #7485, opened 2017-02-14
Branch 147412093

vrv:


vrv:
(makefile breakage is known and should be fixed in the next push, so I'll push without it if this PR fixes everything else)

asimshankar (approved):

vrv:
cc @martinwicke since he wants to know when this push goes in ?

vrv:
shape_test is still failing, but is not reproducible internally, so I'm still going to push if the makefile and the shape_test are the only tests failing.

Issue #7397, opened 2017-02-09
`tf.dynamic_stitch` gradient is incorrect


drasmuss:
### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)
``` python
import tensorflow as tf

x = tf.zeros((1, 3))
y = tf.dynamic_stitch([[0], [0]], [x, tf.ones((1, 3))])

with tf.Session() as sess:
    print("y")
    print(sess.run(y))

    analytic, numeric = tf.test.compute_gradient(x, (1, 3), y, (1, 3))
    print("analytic")
    print(analytic)
    print("numeric")
    print(numeric)
```

gives output
```
y
[[ 1.  1.  1.]]
analytic
[[ 1.  0.  0.]
 [ 0.  1.  0.]
 [ 0.  0.  1.]]
numeric
[[ 0.  0.  0.]
 [ 0.  0.  0.]
 [ 0.  0.  0.]]
```

The numeric gradient correctly shows that `x` has no impact on `y` (since the value of `x` is completely overwritten by a constant in the `dynamic_stitch`).  The analytic gradient is incorrect; it seems like the gradient calculation in `dynamic_stitch` does not handle the case where there are duplicate indices being merged.

girving:
Ug.  You're correct that the gradients are wrong, but I don't see how to fix it without a dramatic performance hit.  Do you have any suggestions?

aselle:
Automatically closing due to lack of recent activity. Please update the issue when new information becomes available, and we will reopen the issue. Thanks!

drasmuss:
The bug still exists, meaning that the `tf.dynamic_stitch` gradients are incorrect.  Is there any other information I can provide that would be helpful?

girving:
Let's leave this open.  Anyone interested should refer to the comments in https://github.com/tensorflow/tensorflow/pull/7487.  The next step would have been to add a new C++ kernel to speed up the bookkeeping required by accurate gradients.

bhack:
Can we close this?

drasmuss:
The gradient implementation is still incorrect, as of TF 2.5.0rc.  Here is an updated example showing the same error
``` python
import tensorflow as tf

x = tf.zeros((1, 3))

analytic, numeric = tf.test.compute_gradient(
    lambda x: tf.dynamic_stitch([[0], [0]], [x, tf.ones((1, 3))]), [x]
)
print("analytic")
print(analytic)
print("numeric")
[...]

PR #7487, opened 2017-02-14
Fix dynamic_stitch gradient implementation

drasmuss:
This modifies the gradient calculation for the `dynamic_stitch` op so that it gives the correct gradient when the input indices have duplicate values (issue is described in more detail in #7397).

I also expanded the unit tests so that the gradient calculation is tested with a wider range of inputs (since it is now more complicated/error-prone), and added a test specifically for the duplicate indices case.

### DIFF
--- a/tensorflow/python/kernel_tests/dynamic_stitch_op_test.py
+++ b/tensorflow/python/kernel_tests/dynamic_stitch_op_test.py
@@ -21,8 +21,10 @@ from __future__ import print_function
 import numpy as np
 
 from tensorflow.python.framework import constant_op
+from tensorflow.python.framework import dtypes
+from tensorflow.python.ops import array_ops
 from tensorflow.python.ops import data_flow_ops
-from tensorflow.python.ops import gradients_impl
+from tensorflow.python.ops import gradient_checker
 import tensorflow.python.ops.data_flow_grad  # pylint: disable=unused-import
 from tensorflow.python.platform import test
 
@@ -32,7 +34,7 @@ class DynamicStitchTest(test.TestCase):
   def testScalar(self):
     with self.test_session():
       indices = [constant_op.constant(0), constant_op.constant(1)]
-      data = [constant_op.constant(40), constant_op.constant(60)]
+      data = [constant_op.constant(40.), constant_op.constant(60.)]
       for step in -1, 1:
         stitched_t = data_flow_ops.dynamic_stitch(indices[::step], data)
         stitched_val = stitched_t.eval()
@@ -42,14 +44,21 @@ class DynamicStitchTest(test.TestCase):
         # length.
         self.assertEqual([None], stitched_t.get_shape().as_list())
 
+        # Test gradients
+        self.assertLess(
+            gradient_checker.compute_gradient_error(
+                data, [x.get_shape().as_list() for x in data],
+                stitched_t, array_ops.shape(stitched_t).eval()),
+            1e-4)
+
   def testSimpleOneDimensional(self):
     with self.test_session():
       indices = [
           constant_op.constant([0, 4, 7]), constant_op.constant([1, 6, 2, 3, 5])
       ]
       data = [
-          constant_op.constant([0, 40, 70]),
-          constant_op.constant([10, 60, 20, 30, 50])
+          constant_op.constant([0, 40, 70], dtypes.float32),
+          constant_op.constant([10, 60, 20, 30, 50], dtypes.float32)
       ]
       stitched_t = data_flow_ops.dynamic_stitch(indices, data)
       stitched_val = stitched_t.eval()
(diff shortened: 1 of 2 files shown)


drasmuss:
Actually, hold off on reviewing this; I had an idea last night for a different implementation that might be more efficient.

drasmuss:
The new implementation does seem to be an improvement.  I left both up there in case there are some internal benchmarks you want to run.

girving:
It looks like your implementation does `n-1` dynamic stitches if there are n inputs.  It seems like this could be reduce to a single dynamic stitch by dynamically stitching from constant tensors 0, 1, ... to get a `source` tensor that says where each destination came from.  This source tensor could then be used to mask out the inputs.  If that makes sense, could you see how it compares speed-wise?

Also, please rewrite the gradient tests to use `tf.test.compute_gradient_error`.  Comparing against numerical gradients avoids more possibilities for error, so it's good to do it unless there's a compelling reason.

drasmuss:
Do you want me to change the existing tests in `testHigherRank` to use `compute_gradient_error` as well?

girving:
@drasmuss Nah, just yours is fine (unless you feel inspired).

girving:
@drasmuss Can you actually squash this time to make it easier to read?  The commits say "squash!" but are not squashed, which is confusing.

drasmuss:
I did some speed tests, and it seems like implementation 2 (with the `n-1` `dynamic_stitches`) is the fastest.  That kind of surprised me, but I guess the masking operation is more expensive than the dynamic stitching.  Exactly where that tradeoff lies will probably depend on the size of the gradient versus the number of inputs.  But in general I think we expect the number of inputs to be fairly small, so the implementation 2 approach scales fairly well.

Let me know if that sounds good, and I'll squash them all and just keep implementation 2.

girving:
@drasmuss Could I see the numbers from the speed tests?  What was `n`?  Do you have a speed comparison with the (incorrect) baseline?

girving:
It is also be possible to massively accelerate the third implementation by writing a new C++ op that does

    merged[indices[m][i, ..., j], ...] = m

It would be pretty easy to make this by copying the current `DynamicStitchOp` kernel and (essentially) constant folding.

Also: Thank you for working on this!  I realize I'm asking for a lot, but it's important to minimize performance regressions to existing functionality.

drasmuss:
Here's the quick script I put together to test things out, if you want to play with the numbers.  I didn't do any rigorous parameter sweeps.
``` python
import time

import numpy as np
import tensorflow as tf

stitch_size = 1000
n_inputs = 10
input_shape = (100, 50, 50)
reps = 10

with tf.device("/cpu:0"):
    idxs = [tf.constant(np.random.randint(stitch_size, size=input_shape[0]))
            for _ in range(n_inputs)]
    vals = [tf.constant(np.random.uniform(-1, 1, size=input_shape))
            for _ in range(n_inputs)]
    y = tf.dynamic_stitch(idxs, vals)
    grad = tf.gradients(y, vals)

total_time = 0.0
for _ in range(reps):
    with tf.Session() as sess:
        start = time.time()
        sess.run(grad)
        total_time += time.time() - start

print(total_time / reps)
```
on my computer this gives
```
baseline: 0.17385516166687012
implementation 1: 0.45461173057556153
implementation 2: 0.3225952863693237
implementation 3: 0.4177032709121704
```

girving:
I think a 2x slowdown is too frightening without doing some more work to reduce, since dynamic_stitch shows up in the embedding layers of a lot of models.  Would you be interested in trying the new C++ op to see if that reduces the gap?  I'm afraid it wouldn't reduce it all that much, though, since as is the gradients are very simple.

@ebrevdo Do you have any thoughts here?  I think the embedding usage of dynamic stitch doesn't have this problem since the indices are all unique there.  Unfortunately, the op doesn't know that.

drasmuss:
One option would be to just change the definition of `dynamic_stitch` to not allow duplicate indices, since that is how it seems to be used a lot of the time.

However, one of the important uses of `dynamic_stitch` (from my perspective anyway) is that it's the only way I can find to do sparse updates on tensors that both runs on the GPU and has a gradient implementation (as suggested by @yaroslavvb here http://stackoverflow.com/questions/37061808/how-does-tensorflow-indexing-work). And that use case would be made less efficient if we don't allow duplicate indices (for the same reason that the gradient is less efficient here, since we'd need to manually compute the non-duplicate indices).

So if we're going to be adding new ops anyway, perhaps the best solution would be to add an op that does those sparse updates (the equivalent of `scatter_update`, but for tensors) to cover that use case, and then restrict the definition of `dynamic_stitch` to unique indices to handle the other use cases (without sacrificing performance).

girving:
@drasmuss Unfortunately I don't see how to restrict dynamic_stitch without sacrificing performance.  We're also bound by backwards compatibility to keep the current spec, which explicitly allows duplicates and says how they're deduped.

drasmuss:
I probably won't have time to add a new kernel, so if anyone else wants to take that on feel free.  In the meantime I squashed all the commits for readability as requested (just keeping implementation 2).  The alternate implementations are preserved at https://github.com/drasmuss/tensorflow/tree/dynamic_stitch_grad_copy if anyone wants to look at them in the future.

dandelionmane:
Sounds like this isn't going to make it in due to the performance regression, so I'm closing this PR. Thanks for the hard work @drasmuss, we appreciate it. Not everything lands, but such is life.

PR #7505, opened 2017-02-14
Cherrypicks

av8ramit:
Cherrypicked the following hashes:
a4281af8626ba07b0fd4235a48ee9a909fb21519
55db877eef7265ec1a48aeaa3ca22d568aa81812
bad13d7a316813d5df0d6dff576d4f071e99f830
4514c71b3b9fbd828c2d29c43da771aede804c9d
69d028435d3b10809f5bf34708e493233485e626
d065a5d984794a0e59bc1787010ef4b911d4ef89
e62619519b788a73393ca64651b8a081ec9c63c1
f953a314426e661ec59da004cfa5d0e5847cd69f
6cfa696b455c34f501bce784b92aaf16d0f14f7c
78c491caaf34f308ce77a61f34770af5f04f8c2a
d85616e623028464027cf67683be696ec666a79c
07deeacf33bd60b776ad35394f5767c42feef138
3aca494186c94090cdac7796694558907c3e0b5e
dd61d018f1ab0ad51bf68204bfc7b58abfbea801
b2607d7c599c8d0d3eed3f712a02dc19eb7d2087

Had some trouble with 78c491caaf34f308ce77a61f34770af5f04f8c2a. Please review again.


vrv:

tensorflow/cc/framework/cc_op_gen.cc
```diff
@@ -57,6 +57,16 @@ string GetPath(const string& dot_h_fname) {
   return result;
 }
 
+// Converts: some/path/to/file.xx
+// to: file
+// (note that suffix is removed)
+string GetFilename(const string& path) {
```
gunan: I thought we had no code changes among these?
This will be misleading, as these wont be in the binary, but it will be in the code of the release.

gunan (dismissed):

gunan:
The failure is caused by the doc generator being missing from the release branch.

tensorflow/tools/docs/BUILD
```diff
@@ -16,6 +16,100 @@ py_binary(
...
+)
+
+py_library(
+    name = "parser",
+    srcs = [
+        "parser.py",
+    ],
+    srcs_version = "PY2AND3",
+)
+
```
yifeif: do we need all the doc gen changes here?

tensorflow/cc/framework/cc_op_gen.cc
martinwicke: These are code changes to the part of the code which writes the comments for autogenerated files, which in turn show up in the docs. @josh11b, @skye for context.

skye: Correct, this affects the comments of the generated header files in the C++ API. So the binary's header files won't have the updated comments, but the docs and manual builds will.

Are the header files even exposed in the binary? (I'm not exactly sure what kind of binary it is... @keveman told me that the headers files aren't included in the pip install, so does that mean they're not exposed?)

yifeif:

yifeif:

tensorflow/tools/docs/generate.py
```diff
@@ -1,408 +0,0 @@
-# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
av8ramit: Ah I see, I should have removed this file from the conflict.

tensorflow/cc/framework/cc_op_gen.cc
martinwicke: That's right. This shouldn't affect the binary distribution, assuming it is stripped properly.

martinwicke (approved):

yifeif:

PR #7310, opened 2017-02-07
change tf.contrib.learn.datasets.load_dataset('boston') to sklearn.da…

hevensun:
change tf.contrib.learn.datasets.load_dataset('boston') to sklearn.datasets.load_boston()

change tf.contrib.learn.datasets.load_dataset('iris') to sklearn.datasets.load_iris()

because when run examples/learn/iris.py and examples/learn/boston.py  with python35, errors like followings occurs:
UnicodeEncodeError: 'utf-16-le' codec can't encode character '\udcds' in position 166: surrogates not allowed
so it's better to change sklearn's method.

PR #7512, opened 2017-02-15
examples change to sklearn dataset load method

hevensun:
The same to #7310 but change the commit account.

change tf.contrib.learn.datasets.load_dataset('boston') to sklearn.datasets.load_boston()
change tf.contrib.learn.datasets.load_dataset('iris') to sklearn.datasets.load_iris()
because when run examples/learn/iris.py and examples/learn/boston.py with python35, errors like followings occurs:
UnicodeEncodeError: 'utf-16-le' codec can't encode character '\udcds' in position 166: surrogates not allowed
so it's better to change sklearn's method.


vrv:
Great, CLA looks good, thanks.  We're quite busy through tomorrow, but we'll look at this soon.

dandelionmane:
Based on the discussion at https://github.com/tensorflow/tensorflow/pull/7310, it sounds like we should fix the unicode compat bug in tf.contrib.learn.datasets.load_dataset rather than adding more dependencies on SKLearn. @hevensun can you take that on?

hevensun:
@dandelionmane  first I want to fix this unicode error using sklearn's datasets load  method for other users to use, then I will deep into the unicode compat bug in tf.contrib.learn.datasets.load_dataset to if I can fix it.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

Issue #6906, opened 2017-01-17
Relaxing inferenced shape


gaohuazuo:
Inferenced shape will put restriction on feeding, so sometimes it is desirable to relax inferenced shape. Currently this can be done by `lambda x: tf.cond(tf.constant(True), lambda: x, lambda: tf.placeholder(x.dtype))`, or `tf.placeholder_with_default` if gradient is not required (`placeholder_with_default` has no gradient registered).

I also read a comment in `identity_op.cc`: 

>PlaceholderWithDefault does the same thing as Identity, but has a
>different shape function (and constant value function) registered.

Possibly register a gradient for `placeholder_with_default`?

asimshankar:
@vrv : Any thoughts here? I'm not quite sure what the position is on feeding arbitrary nodes in the graph as it can can mess with shape inference, gradient computations (see #6903 that @gaohuazuo also filed).

vrv:
I have no thoughts because I don't really understand the problem.  Specifying a placeholder without shape allows feeding any shape data.

I don't think we register gradients for placeholders either.  I think providing more context into what you're trying to do will help us out.

yaroslavvb:
I don't understand the problem either, but you can specify the gradient for placeholder_with_default using `override_gradient_map` (it'll return `None` instead of crashing with error on gradient computation)

gaohuazuo:
Thanks for the replies. What I'm doing is to explore the behaviour of the discriminator in GAN. The graph is roughly the following:

![untitled diagram](https://cloud.githubusercontent.com/assets/10446514/22094828/4b0cfaf0-de4b-11e6-97e3-f3c6b263b04d.png)

During training, gradients have to propagate through `x` to `G`. On the other hand, I want to feed some arbitrary `x` to `D` for testing. I feel that this should be suitable for `placeholder_with_default`, but it has no gradient and thereby prevents training using the same graph. Note that I either feed `x` or use the gradient, but never both, so this should be a good use case of `placeholder_with_default`.

I guess it is a protective measure that `placeholder_with_default` doesn't have a gradient. However, this isn't quite meaningful, as other ops don't have the same protection, and `placeholder_with_default` has nothing special to do with feeding, except its name.

In short, I would suggest using `identity` as the gradient of `placeholder_with_default`, because it does no harm and some good.

vrv:
[...]

PR #7517, opened 2017-02-15
Add gradient for `placeholder_with_default`

gaohuazuo:
patch for #6906

### DIFF
--- a/tensorflow/python/kernel_tests/constant_op_test.py
+++ b/tensorflow/python/kernel_tests/constant_op_test.py
@@ -768,6 +768,12 @@ class PlaceholderWithDefaultTest(test.TestCase):
       self.assertAllEqual(
           [[3, 3], [3, 3]], a.eval(feed_dict={p: [[3, 3], [3, 3]]}))
 
+  def testGradient(self):
+    with self.test_session():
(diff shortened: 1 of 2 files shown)


tensorflow/python/kernel_tests/constant_op_test.py
```diff
@@ -768,6 +768,12 @@ def testNoShape(self):
       self.assertAllEqual(
           [[3, 3], [3, 3]], a.eval(feed_dict={p: [[3, 3], [3, 3]]}))
 
+  def testGradient(self):
+    with self.test_session():
+      x = array_ops.placeholder(dtypes_lib.float32, [5, 7])
+      y = array_ops.placeholder_with_default(x, None)
+      err = gradient_checker.compute_gradient_error(x, [5, 7], y, [5, 7])
+      self.assertLess(err, 1e-3)
```
vrv: given that this should be passing the identity, should we be even more aggressive and expect fewer differnces?

gaohuazuo: The error is 1.3e-5 in my environment.

vrv:

vrv (approved):

PR #7527, opened 2017-02-15
Fix #7328 by checking if socket address is valid

### LINKED ISSUE #7328: Add network address check in clusterSpec
Today, I got an error when starting a distributed TensorFlow computation. I debugged for hours and finally found the problem is that TensorFlow doesn't have network address check in `ClusterSpec` method. And it endlessly connect the illegal network address `100.xxx.xxx.232.:2223`, because there is a redundant `.` after `232`.

byronyi:
Not sure if it's the most standard compliant way for both platforms, but I've tried my best.

Some references: [POSIX](http://www.unix.com/man-page/POSIX/3posix/getaddrinfo/) and [Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/ms738520(v=vs.85).aspx).

### DIFF
--- a/tensorflow/compiler/xla/tests/literal_test_util.cc
+++ b/tensorflow/compiler/xla/tests/literal_test_util.cc
@@ -36,6 +36,7 @@ limitations under the License.
 #include "tensorflow/core/platform/protobuf.h"
 #include "tensorflow/core/platform/test.h"
 #include "tensorflow/core/platform/types.h"
+#include "tensorflow/core/platform/host_info.h"
 
 namespace xla {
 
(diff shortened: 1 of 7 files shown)


byronyi:
An interesting typo in the ``BUILD`` file seems to exclude several tests, including the one I added for this patch.

byronyi:
Can one of the committer check with [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/BUILD#L1650)? Seems several tests are excluded because of a typo, and some of them failed in my build when they are enabled.

byronyi:
Since the original submitter of #7328 @DjangoPeng is working on this, I'm closing this PR.

PR #7532, opened 2017-02-15
[#issue7328]Validate ip address in distributed computation

DjangoPeng:
[#issue7328](https://github.com/tensorflow/tensorflow/issues/7328)

`inet_pton()` is a recommended and general way to validate ip address(including IPv4 and IPv6) both in [linux](https://linux.die.net/man/3/inet_pton) and [Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/cc805844(v=vs.85).aspx). 

@mrry  Do I add the validation in the right place? I would add some unit tests tomorrow morning.

BTW:Does TensorFlow support IPv6 now?

### DIFF
--- a/tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc
+++ b/tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc
@@ -133,7 +133,7 @@ Status GrpcServer::Init() {
         return errors::InvalidArgument(
             "Could not parse port for local server from \"", iter->second,
             "\"");
-      } else if (!IsValidIpAddr(hostname_port[0])) {
+      } else if (!port::IsValidIpAddr(hostname_port[0])) {
         return errors::InvalidArgument(
             "Invalid ip address for local server from \"", iter->second,
             "\"");
--- a/tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc
+++ b/tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc
@@ -42,6 +42,7 @@ limitations under the License.
 #include "tensorflow/core/framework/op.h"
 #include "tensorflow/core/lib/strings/strcat.h"
 #include "tensorflow/core/platform/env.h"
+#include "tensorflow/core/platform/host_info.h"
 #include "tensorflow/core/platform/mem.h"
 #include "tensorflow/core/public/session_options.h"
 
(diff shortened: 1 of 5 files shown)


byronyi:
Since this is a WIP, please take a look at #7527, which is aimed to give a portable solution via ``getaddrinfo``.

DjangoPeng:
@mrry please review.

tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc
```diff
@@ -132,6 +133,10 @@ Status GrpcServer::Init() {
         return errors::InvalidArgument(
             "Could not parse port for local server from \"", iter->second,
             "\"");
+      } else if (!port::IsValidIpAddr(hostname_port[0])) {
+        return errors::InvalidArgument(
+            "Invalid ip address for local server from \"", iter->second,
+            "\"");
```
mrry: I don't think this solves issue #7328. This is only checking the address for the local task (which we then overwrite with `0.0.0.0` or `localhost` later in this method). Since the process running this code will (or at least should) never attempt to connect to itself, I very much doubt that it's the source of the hanging.

It'd probably be more valuable to add validation for the addresses that get added to the `GrpcChannelSpec` (on L180). These are the addresses that will be used for contacting other workers, which is probably where your code was blocking.

The more important place for validation would be when you build the `GrpcChannelSpec`

tensorflow/core/platform/posix/port.cc
```diff
@@ -47,6 +48,11 @@ string Hostname() {
   return string(hostname);
 }
 
+bool IsValidIpAddr(const string& ip) {
+  struct sockaddr_in sa;
+  return inet_pton(AF_INET, ip.c_str(), &(sa.sin_addr))!=0;
```
mrry: `inet_pton()` only accepts textual IP addresses, so this will prevent users from providing hostnames in their `ClusterSpec`, and break backwards compatibility.

mrry (changes requested):

tensorflow/core/platform/posix/port.cc
byronyi: @DjangoPeng If you have time please take a look at #7527, which attempts to use ``getaddrinfo`` that supports hostnames.

dandelionmane:
This PR has been waiting for response for two weeks, and based on @mrry's review it sounds like it won't fix the issue it's trying to address. So I'm going to close this for now. @DjangoPeng thanks for sending it in, and feel free to open a new PR that addresses mrry's concerns.

Issue #4742, opened 2016-10-04
Atrous convolution does not preserve tensor shape


ethereon:
For an input with an undefined batch size, `atrous_conv2d` emits tensors where all except the final dimension are undefined:

``` python
input = tf.placeholder(tf.float32, (None, 256, 256, 3))

conv = tf.nn.conv2d(input, tf.zeros((3, 3, 3, 16)), strides=[1, 1, 1, 1], padding='SAME')
print(conv.get_shape()) # Correctly displays (?, 256, 256, 16)

dilated = tf.nn.atrous_conv2d(input, tf.zeros((3, 3, 3, 16)), rate=2, padding='SAME')
print(dilated.get_shape()) # Displays (?, ?, ?, 16)
```

(For concrete batch sizes, everything works as expected.)

Tested on `0.10.0rc0`

tatatodd:
Indeed I have reproduced the problem.  I believe @gpapan implemented atrous_conv2d, and might have thoughts on how easy this would be to fix.

jkiske:
:+1:

gerbenvanveenendaal:
:+1:

AnishShah:
Can I work on this? I think, I can solve this.

vrv:
That would be great @AnishShah, can you describe what the problem is?

AnishShah:
@vrv The problem is in this [line](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_ops.py#L974). It is using `ShapeOp` to estimate `paddings` for `ShapeToBatchOp`. That is why it is not able to predict the output shape. I tried few things, but I was unsuccessful. What do you suggest?

aselle:
fdsa

aselle:
@AnishShah, do you have any updates?

AnishShah:
I tried, but not able to solve it. Sorry.

ahundt:
I believe the source of confusion here may be due to varying definitions of atrous convolution depending on the paper being read. Basically, some papers defined atrous convolutions incorrectly when they really meant dilated convolutions. This is explained in [Multi-Scale Context Aggregation by Dilated Convolutions](https://arxiv.org/abs/1511.07122) with the authors' implementation in https://github.com/fyu/dilation.

Also see the related issue https://github.com/tensorflow/tensorflow/issues/3492.

[...]

PR #7545, opened 2017-02-16
Added atrous_conv1d and 3d. Refactored 2d.

nmiculinic:
This commit makes following changes:
* deleted most atrous_conv2d code to reuse existing tf.nn.convolution function
* added atrous_conv1d and atrous_conv3d with similar API as atrous_conv2d
* Added support for variable rate per dimension, e.g. for atrous_conv2d
  rate=2, or rate=[2,1] does different things. Former is equal to
  rate=[2,2]. rate_i determines dilation_rate in dimension i.
* added strides support with same API as in tf.nn.convolution function

This commit makes more code deletions then additions. However
documentation per function makes it appear large.

Test Plan:

Some simple tests to verify I haven't screwed something up:

```
A = np.array([1, 2, 3, 4, 5, 6], dtype=np.float32).reshape(1, 6, 1)
print(A)

kernel = np.array([100, 10, 1], dtype=np.float32).reshape(3, 1, 1)

with tf.Session() as sess:
    print(sess.run(tf.nn.atrous_conv1d(A, kernel, padding='SAME', rate=[2])))
```

```
B = np.arange(16, dtype=np.float32).reshape(1, 4, 4, 1)
kernel = np.array([1000, 100, 10, 1.0], dtype=np.float32).reshape(2, 2, 1, 1)

with tf.Session() as sess:
    a = sess.run(tf.nn.convolution(B, kernel, padding='SAME', dilation_rate=np.array([2, 2])))
    b = sess.run(tf.nn.atrous_conv2d(B, kernel, rate=2, padding='SAME'))
    print(np.allclose(a, b))
    print(a)
    print(b)
```

```
C = np.arange(4**3, dtype=np.float32).reshape(1, 4, 4, 4, 1)
kernel = (10**np.arange(8, 0, -1, dtype=np.float32)).reshape(2, 2, 2, 1, 1)

with tf.Session() as sess:
    a = sess.run(tf.nn.conv3d(C, kernel, strides=[1, 1,1,1,1], padding='SAME'))
    b = sess.run(tf.nn.atrous_conv3d(C, kernel, rate=1, padding='SAME'))
    print(np.allclose(a, b))
    print(a)
    print(b)
```

Also running atrous_conv2d unit tests to verify backward compatibility.

### DIFF
--- a/tensorflow/python/ops/nn.py
+++ b/tensorflow/python/ops/nn.py
@@ -31,7 +31,9 @@
 @@depthwise_conv2d
 @@depthwise_conv2d_native
 @@separable_conv2d
+@@atrous_conv1d
 @@atrous_conv2d
+@@atrous_conv3d
 @@atrous_conv2d_transpose
 @@conv2d_transpose
 @@conv1d
--- a/tensorflow/python/ops/nn.py
+++ b/tensorflow/python/ops/nn.py
@@ -31,9 +31,7 @@
 @@depthwise_conv2d
 @@depthwise_conv2d_native
 @@separable_conv2d
-@@atrous_conv1d
 @@atrous_conv2d
-@@atrous_conv3d
 @@atrous_conv2d_transpose
 @@conv2d_transpose
 @@conv1d
--- a/tensorflow/python/ops/nn_ops.py
+++ b/tensorflow/python/ops/nn_ops.py
@@ -805,7 +805,52 @@ def pool(input,  # pylint: disable=redefined-builtin
         filter_shape=window_shape)
 
 
-def atrous_conv2d(value, filters, rate, padding, name=None):
+def atrous_conv1d(input, filter, rate, padding, strides=None, name=None):
+  """Atrous convolution (a.k.a. convolution with holes or dilated convolution).
+
+  Computes a 1-D atrous convolution, also known as convolution with holes or
+  dilated convolution, given 3-D `input` and `filter` tensors. If the `rate`
+  parameter is equal to one, it performs regular 1-D convolution. If the `rate`
+  parameter is greater than one, it performs convolution with holes, sampling
+  the input values every `rate` points.
+  This is equivalent to convolving the input with a set of upsampled filters,
+  produced by inserting `rate - 1` zeros between two consecutive values of the
+  filters along the `height` dimensions, hence the name atrous
+  convolution or convolution with holes (the French word trous means holes in
+  English).
+
+  See @{tf.nn.atrous_conv2d}
+
+  Args:
+  input: A 3-D `Tensor` of type `float`. It needs to be in the default "NHWC"
+    format. Its shape is `[batch, in_height, in_channels]`.
+  filter: A 3-D `Tensor` with the same type as `input` and shape
+    `[filter_height, in_channels, out_channels]`. `filter`'
+    `in_channels` dimension must match that of `input`. Atrous convolution is
+    equivalent to standard convolution with upsampled filters with effective
+    height `filter_height + (filter_height - 1) * (rate - 1)`, produced by
+    inserting `rate - 1` zeros along consecutive elements across the
+    `filter`' spatial dimensions.
+  rate: A positive int32. The stride with which we sample input values across
+    the `height` and `width` dimensions. Equivalently, the rate by which we
+    upsample the filter values by inserting zeros across the `height` and
+    `width` dimensions. In the literature, the same parameter is sometimes
+    called `input stride` or `dilation`.
+  padding: A string, either `'VALID'` or `'SAME'`. The padding algorithm.
+  strides: Optional.  Sequence of 1 ints >= 1.  Specifies the output stride.
+    Defaults to [1].  If any value of strides is > 1, then all values of
+    dilation_rate must be 1.
+  name: Optional name for the returned tensor.
+  Returns:
+    A `Tensor` with the same type as `value`.
+  Raises:
+    ValueError: If input/output depth does not match `filters`' shape, or if
+      padding is other than `'VALID'` or `'SAME'`.
+
+  """
+  return convolution(input=input, filter=filter, padding=padding, dilation_rate=np.broadcast_to(rate, (1, )), strides=strides, name=name)
+
+def atrous_conv2d(value, filters, rate, padding, strides=None, name=None):
   """Atrous convolution (a.k.a. convolution with holes or dilated convolution).
 
   Computes a 2-D atrous convolution, also known as convolution with holes or
@@ -903,7 +948,7 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
       the `height` and `width` dimensions. Equivalently, the rate by which we
       upsample the filter values by inserting zeros across the `height` and
       `width` dimensions. In the literature, the same parameter is sometimes
-      called `input stride` or `dilation`.
+      called `input stride` or `dilation`. Altrenatively it could be 3 element sequence denoting dilation rate in each dimension.
     padding: A string, either `'VALID'` or `'SAME'`. The padding algorithm.
     name: Optional name for the returned tensor.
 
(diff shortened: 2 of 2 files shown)


dandelionmane:
@gpapan PTAL

tensorflow/python/ops/nn_ops.py
```diff
@@ -903,7 +948,7 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
       the `height` and `width` dimensions. Equivalently, the rate by which we
       upsample the filter values by inserting zeros across the `height` and
       `width` dimensions. In the literature, the same parameter is sometimes
-      called `input stride` or `dilation`.
+      called `input stride` or `dilation`. Altrenatively it could be 3 element sequence denoting dilation rate in each dimension.
```
gpapan: s/Altrenatively/Alternatively

gpapan: All lines must have length at most 80 pixels.

tensorflow/python/ops/nn_ops.py
```diff
@@ -914,94 +959,53 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
...
-    else:
-      value_shape = array_ops.shape(value)
-
-    in_height = value_shape[1] + pad_top + pad_bottom
-    in_width = value_shape[2] + pad_left + pad_right
-
-    # More padding so that rate divides the height and width of the input.
-    pad_bottom_extra = (rate - in_height % rate) % rate
-    pad_right_extra = (rate - in_width % rate) % rate
+  return convolution(input=value, filter=filters, padding=padding, dilation_rate=np.broadcast_to(rate, (2, )), strides=strides, name=name)
```
gpapan: If I recall correctly, we had kept the legacy atrous_conv2 as a separate op for backwards compatibility with already exported graphs. Can one of the TF admins comment if this is needed or not?

tensorflow/python/ops/nn_ops.py
```diff
@@ -914,94 +959,53 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
...
-                                     paddings=space_to_batch_pad,
-                                     block_size=rate)
+  Computes a 3-D atrous convolution, also known as convolution with holes or
+  dilated convolution, given 5-D `input` and `filter` tensors. If the `rate`
+  parameter is equal to one, it performs regular 3-D convolution. If the `rate`
+  parameter is greater than one, it performs convolution with holes, sampling
+  the input values every `rate` points.
+  This is equivalent to convolving the input with a set of upsampled filters,
+  produced by inserting `rate - 1` zeros between two consecutive values of the
+  filters along the `height`, `width` and `depth` dimensions, hence the name atrous convolution or convolution with holes (the French word trous means holes in
```
gpapan: Line >80 chars long

tensorflow/python/ops/nn_ops.py
```diff
@@ -914,94 +959,53 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
...
+      `in_channels` dimension must match that of `input`. Atrous convolution is
+      equivalent to standard convolution with upsampled filters with effective
+      height `filter_height + (filter_height - 1) * (rate - 1)`, produced by
+      inserting `rate - 1` zeros along consecutive elements across the
+      `filter`' spatial dimensions.
+    rate: A positive int32. The stride with which we sample input values across
+      the `height` and `width` dimensions. Equivalently, the rate by which we
+      upsample the filter values by inserting zeros across the `height` and
+      `width` dimensions. In the literature, the same parameter is sometimes
+      called `input stride` or `dilation`. Altrenatively it could be 3 element sequence denoting dilation rate in each dimension.
```
gpapan: Line >80 chars long

tensorflow/python/ops/nn_ops.py
```diff
@@ -914,94 +959,53 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
...
+  Returns:
+    A `Tensor` with the same type as `value`.
 
-    return value
+  Raises:
+    ValueError: If input/output depth does not match `filters`' shape, or if
+      padding is other than `'VALID'` or `'SAME'`.
 
+  """
+  return convolution(input=input, filter=filter, padding=padding, dilation_rate=np.broadcast_to(rate, (3, )), strides=strides, name=name)
```
gpapan: Line >80 chars long

tensorflow/python/ops/nn_ops.py
gpapan: Line >80 chars long

tensorflow/python/ops/nn_ops.py
```diff
@@ -805,7 +805,52 @@ def op(converted_input, _, converted_padding):  # pylint: disable=missing-docstr
...
+    dilation_rate must be 1.
+  name: Optional name for the returned tensor.
+  Returns:
+    A `Tensor` with the same type as `value`.
+  Raises:
+    ValueError: If input/output depth does not match `filters`' shape, or if
+      padding is other than `'VALID'` or `'SAME'`.
+
+  """
+  return convolution(input=input, filter=filter, padding=padding, dilation_rate=np.broadcast_to(rate, (1, )), strides=strides, name=name)
```
gpapan: Line >80 chars long

gpapan (commented):
Can you please address some minor comments?

gpapan:
@jbms

jbms:
Removing the duplicate implementation of atrous convolution in the atrous_conv2d function is a desirable change.

As far as adding atrous_conv1d and atrous_conv3d, I don't see the advantage --- better to just use the convolution function directly.  atrous_conv2d existed prior to the unified tf.nn.convolution function, and is preserved for backwards compatibility.

tensorflow/python/ops/nn_ops.py
jbms: The underlying BatchToSpace and SpaceToBatch ops are needed for backwards compatibility with exported graphs.  Changing the Python wrappers (which change how new graphs are constructed) should not impact that.

nmiculinic:
> As far as adding atrous_conv1d and atrous_conv3d, I don't see the advantage --- better to just use the convolution function directly. atrous_conv2d existed prior to the unified tf.nn.convolution function, and is preserved for backwards compatibility.

For both completeness and ease of use. When I was searching for atrous_conv1d I only found atrous_conv2d and assumed only that one is implemented (and I implemented one myself mimicking atrous_conv1d implementation before realizing there's general convolution which handles all cases). Having atrous_conv2d as standalone increase its salience and confuses users (me included).

jbms:
I think this would be better addressed by adding a note to the documentation of atrous_conv2d explaining that it is a legacy interface and that tf.nn.convolution can be used for atrous convolutions of any number of dimensions, rather than adding additional redundant functions.

ahundt:
Some related discussion is going on in #4742, part of which I'll repost here bc it is pertinent.

[conv2d_same](https://github.com/tensorflow/models/blob/master/slim/nets/resnet_utils.py#L77) in `tensorflow/models` provides both input tensor size and output tensor size:
```
def conv2d_same(inputs, num_outputs, kernel_size, stride, rate=1, scope=None):
  """[snip...]
  Args:
    inputs: A 4-D tensor of size [batch, height_in, width_in, channels].
  [snip...]
  Returns:
    output: A 4-D tensor of size [batch, height_out, width_out, channels] with
      the convolution output.
  """
```

I think it would be very productive to add a note in each appropriate function explaining what the output dimensions would be relative to given input dimensions in as they vary by configuration.

nmiculinic:
>  think this would be better addressed by adding a note to the documentation of atrous_conv2d explaining that it is a legacy interface and that tf.nn.convolution can be used for atrous convolutions of any number of dimensions, rather than adding additional redundant functions.

What is the protocol for function deprecation? Is there some annotation, some specific string in documentation or what?

If I understand correctly, you'd like deleting atrous 1d and 3d, and editing 2d with documentation changes

tensorflow/python/ops/nn_ops.py
```diff
@@ -903,7 +950,8 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
       the `height` and `width` dimensions. Equivalently, the rate by which we
       upsample the filter values by inserting zeros across the `height` and
       `width` dimensions. In the literature, the same parameter is sometimes
-      called `input stride` or `dilation`.
+      called `input stride` or `dilation`. Alternatively it could be 3 element
```
jbms: This should be "2 element".

jbms:
Yes, remove atrous 1d and 3d.

There is the @deprecated decorator in tensorflow/python/util/deprecation.py.  However, since there isn't yet a specific plan/date for removing the atrous_conv2d function, you shouldn't use that in this case.

Instead, you could just add a note to the top of the atrous_conv2d, such as:

This function is a simpler wrapper around the more general @{tf.nn.convolution}, and exists only for backwards compatibility.  You can use @{tf.nn.convolution} to perform 1-D, 2-D, or 3-D atrous convolution.

tensorflow/python/ops/nn.py
```diff
@@ -31,7 +31,9 @@
 @@depthwise_conv2d
 @@depthwise_conv2d_native
 @@separable_conv2d
+@@atrous_conv1d
```
jbms: These should be removed.

tensorflow/python/ops/nn_ops.py
```diff
@@ -805,9 +805,14 @@ def op(converted_input, _, converted_padding):  # pylint: disable=missing-docstr
         filter_shape=window_shape)
 
 
-def atrous_conv2d(value, filters, rate, padding, name=None):
+def atrous_conv2d(value, filters, rate, padding, strides=None, name=None):
```
jbms: Not clear if it makes sense to add strides, since it is supported by the more generic convolution anyway, but if it is added, it should be documented in the Args section.

drpngx:
Jenkins, test this please.

drpngx:
@jbms @gpapan good to go?

martinwicke:
@jbms @gpapan Ping?

martinwicke:
@nmiculinic there are test failures, can you address those?

tensorflow/python/ops/nn_ops.py
```diff
@@ -914,93 +919,9 @@ def atrous_conv2d(value, filters, rate, padding, name=None):
...
-    batch_to_space_crop = [[0, pad_bottom_extra], [0, pad_right_extra]]
-
-    value = array_ops.batch_to_space(input=value,
-                                     crops=batch_to_space_crop,
-                                     block_size=rate)
-
-    return value
+  return convolution(input=value, filter=filters, padding=padding,
+                     dilation_rate=np.broadcast_to(rate, (2, )),
+                     strides=strides, name=name)
```
jbms: Test failure is caused by reference to undefined variable strides.

vrv:
@jbms assuming tests pass, is this approved?  Thanks!

jbms:
Yes looks good once tests pass.

vrv:

vrv:
I made the fix for the @nmiculinic, let me know if this looks good (hit approve).  Thanks!

jbms (approved):

vrv:
I'm going to assume the build timeout is unrelated since no tests were added.  Merigng.

PR #7555, opened 2017-02-16
TFLearn Estimator Summary Writer Fix

### LINKED ISSUE #7554: TFLearn Estimator Summary Writer Fails.
### Issue:
TFLearn estimator summary writer fails because _write_dict_to_summary() in [/tensorflow/contrib/learn/python/learn/estimators/estimator.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L324) dose not allow int values, which naturally stops the np.int64 valued 'global_step' parameter from being written, thus causing the summary writer to fail.

### How to reproduce:
Try and run [wide_n_deep_tutorial.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/learn/wide_n_deep_tutorial.py), and you will get:

`INFO:tensorflow:Saving dict for global step 202: accuracy = 0.818254, accuracy/baseline_label_mean = 0.236226, accuracy/threshold_0.500000_mean = 0.818254, auc = 0.742517, global_step = 202, labels/actual_label_mean = 0.236226, labels/prediction_mean = 0.182356, loss = 0.680463, precision/positive_threshold_0.500000_mean = 0.764145, recall/positive_threshold_0.500000_mean = 0.333593
WARNING:tensorflow:Skipping summary for global_step, must be a float or np.float32.`

### Pull Request:
#7555

thesuperzapper:
Fixes #7554


tensorflow/contrib/learn/python/learn/estimators/estimator.py
```diff
@@ -321,11 +321,15 @@ def _write_dict_to_summary(output_dir,
       continue
     value = summary_proto.value.add()
     value.tag = key
-    if (isinstance(dictionary[key], np.float32) or
-        isinstance(dictionary[key], float)):
+    if (isinstance(dictionary[key], np.float32) or 
+            isinstance(dictionary[key], float)):
```
martinwicke: Can you remove the extra whitespace here?

tensorflow/contrib/learn/python/learn/estimators/estimator.py
```diff
@@ -321,11 +321,15 @@ def _write_dict_to_summary(output_dir,
...
       value.simple_value = float(dictionary[key])
+    elif (isinstance(dictionary[key], np.int64) or
+              isinstance(dictionary[key], np.int32) or
+              isinstance(dictionary[key], int)):
```
martinwicke: Is int even different from both of these?

tensorflow/contrib/learn/python/learn/estimators/estimator.py
```diff
@@ -321,11 +321,15 @@ def _write_dict_to_summary(output_dir,
...
+      value.simple_value = int(dictionary[key])
     else:
-      logging.warn('Skipping summary for %s, must be a float or np.float32.',
+      logging.warn('Skipping summary for %s, must be a float, np.float32, np.int64, np.int32 or int.',
```
martinwicke: Here also, I don't think int can ever be neither int32, nor int64.

martinwicke (changes requested):
Thank you for this! Can you please add a test that exercises this new code?

thesuperzapper:
Sorry for disappearing for a while, been super busy, but we really need to get this fixed. We need the int type because people can write their own MetricSpec's.

tensorflow/contrib/learn/python/learn/estimators/estimator.py
thesuperzapper: Yep

dandelionmane:
@thesuperzapper Please add a test as martinwicke requested.

drpngx:
@thesuperzapper feel free to add tests. Thanks!

thesuperzapper:
I am not 100% sure how to write test cases.

martinwicke:
You can look at the estimator_test.py file to see other tests.
We should have test cases that exercise these new code paths.

I'm a bit unclear what failure this addresses. I think the change is fine in general though.

Can you make the same change (and test) to tensorflow/python/estimators/estimator.py as well? The _write_dict_to_summary code looks the same.

martinwicke:
@thesuperzapper Ping?

thesuperzapper:
_Note: I added the same fix to `tensorflow/python/estimators/estimator.py`._

This pull request simply allows us to write int-type values in the summary file.

A side effect of this pull request will be **writing the value of 'global_step' at 'global_step'** -- a useless piece of information -- this is because 'global_step' is inside the ['eval_results'](https://github.com/tensorflow/tensorflow/blob/1a1527ab1fcffd94f692b1301fdbe87903ce7bdb/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L841) & ['dictionary'](https://github.com/tensorflow/tensorflow/blob/1a1527ab1fcffd94f692b1301fdbe87903ce7bdb/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L330) objects, (Which would not previously have been written by [`_write_dict_to_summary`](https://github.com/tensorflow/tensorflow/blob/1a1527ab1fcffd94f692b1301fdbe87903ce7bdb/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L316) because 'global_step' is an int type, as seen in #7554)

Therefore, we probably should remove the 'global_step' key from the 'eval_results' dictionary. (Or filter it in `_write_dict_to_summary`), otherwise 'global_step' by 'global_step' is displayed in TensorBoard. 

But also note that we return the 'eval_results' dictionary in [ `_evaluate_model`](https://github.com/tensorflow/tensorflow/blob/1a1527ab1fcffd94f692b1301fdbe87903ce7bdb/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L853). Therefore, I would like some advice on the best way of removing/filtering the 'global_step' key, so that we don't break this return, or refactor stuff unnecessarily.

martinwicke:
I think the right solution here is to filter global_step out inside _write_dict_to_summary. Nobody needs to see that graph, it's boring.

I still want a test. Without one, I guarantee that this will break sooner or later.

vrv:
another ping for @thesuperzapper !

thesuperzapper:
@vrv 
Sorry guys, I have been stupidly busy! I am still not sure how the testing framwork works.

Do I just add a function to the [estimator_test.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/estimators/estimator_test.py) file?

Also, do you think hard-coded filtering the `global_step` key inside [_write_dict_to_summary](https://github.com/tensorflow/tensorflow/blob/1a1527ab1fcffd94f692b1301fdbe87903ce7bdb/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L331) is good enough?

vrv:
Yes, a test to estimator_test.py would be the right place.

(I don't know about the 'global step' key filtering, punting to @martinwicke and @ispirmustafa )

martinwicke:
You can use the `tf.train.get_global_step` method to check whether a value you're about to write is the global step tensor.

ispirmustafa:
+1 for "filter global_step out inside _write_dict_to_summary"
You can do if before calling the _write_dict_to_summary.

ispirmustafa (changes requested):
global step should be filtered out before calling this function.

drpngx:
@thesuperzapper any luck on this?

thesuperzapper:
I will look at this later today, but it's pretty much just making the current changes, and then stopping the writer from writing the global step.

rmlarsen:
@ispirmustafa @martinwicke could you take another look, please?

ispirmustafa (approved):

rmlarsen:

PR #7580, opened 2017-02-16
Upgrade docker images for distributed testing to unbutu:16.04

caisq:


jhseu (approved):

vrv:
https://ci.tensorflow.org/job/tensorflow-pull-requests-mac/3805/console shows an interestng new error I haven't seen before -- do you think it's related?

jhseu:
Can't possibly be related since we don't even have Docker installed on the macs. Looking at the errors, it seems that someone changed the numpy version on our macs.

caisq:
@vrv. I don't think it is related. The files in tools/dist_test is not depended on by the failing tests. 

Looking closer at the log, I noticed that most of failures have the following error: 
```
TypeError: norm() got an unexpected keyword argument 'keepdims'
```

as well as:
```
TypeError: linspace() got an unexpected keyword argument 'dtype'
```

These seem to be some sort of numpy version inconsistency in the functions `np.linalg.norm` and `np.linspace` on that particular Mac machine (mac0)

vrv:
@yifeif another instance of inconsistency ...

caisq:
@yifeif  can you check the version of numpy and scipy on mac0? 
On mac1, they are 1.11.0 and 0.15.1, respectively, consistent with our slave install script.

yifeif:
>>> numpy.version.version
'1.11.0'
>>> scipy.version.version
'0.15.1'

but looks like builds on mac0 is running with different Python library path compared to other macs(/Library/Python/2.7/site-packages vs /usr/local/Cellar/python/2.7.12_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages)??

I'll take it offline and investigate further.

Issue #7374, opened 2017-02-09
Issues with Eigen when building tf_tutorials_example_trainer on Windows 10


Katuali:
I'm attempting to build Tensorflow C++ on Windows 10 following the instructions of the cmake readme. However, I'm getting issues with compiling at the Eigen step.

```Compiling the Fortran compiler identification source file "CMakeFortranCompilerId.F" failed.
Compiler:  
Build flags: 
Id flags: 

The output was:
1

Microsoft Visual Studio 2015 Version 14.0.25420.1.
Copyright (C) Microsoft Corp. All rights reserved.

The license for Visual Studio has expired.

The evaluation period for this product has ended.

Performing C++ SOURCE FILE Test COMPILER_SUPPORT_FASTMATH failed with the following output:
Change Dir: C:/Users/Irene/Documents/tensorflow/tensorflow/contrib/cmake/build/eigen/src/eigen-build/CMakeFiles/CMakeTmp

Run Build Command:"C:/Program Files (x86)/MSBuild/14.0/bin/MSBuild.exe" "cmTC_70ffc.vcxproj" "/p:Configuration=Debug" "/p:VisualStudioVersion=14.0"
Microsoft (R) Build Engine version 14.0.25420.1

Copyright (C) Microsoft Corporation. All rights reserved.



Build started 9/02/2017 3:13:30 PM.

Project "C:\Users\Irene\Documents\tensorflow\tensorflow\contrib\cmake\build\eigen\src\eigen-build\CMakeFiles\CMakeTmp\cmTC_70ffc.vcxproj" on node 1 (default targets).

PrepareForBuild:

  Creating directory "cmTC_70ffc.dir\Debug\".

  Creating directory "C:\Users\Irene\Documents\tensorflow\tensorflow\contrib\cmake\build\eigen\src\eigen-build\CMakeFiles\CMakeTmp\Debug\".

  Creating directory "cmTC_70ffc.dir\Debug\cmTC_70ffc.tlog\".

InitializeBuildStatus:

  Creating "cmTC_70ffc.dir\Debug\cmTC_70ffc.tlog\unsuccessfulbuild" because "AlwaysCreate" was specified.

ClCompile:

[...]

PR #7586, opened 2017-02-16
[CMake] Disable BUILD_TESTING for Eigen.

mrry:
We don't actually run these tests, so building them is a waste of time. They can also cause surprising interactions with the environment (e.g. see issue #7374), in cases where some of the test dependencies are available.


mrry:
Looks like the build is a bit hungover this morning....

mrry:

PR #7592, opened 2017-02-16
Branch 147741833

vrv:

PR #7605, opened 2017-02-17
Add missing typename to fix Windows build

snnn:
#7592


snnn:
It has been fixed internally

PR #7606, opened 2017-02-17
Update random_forest_mnist.py

gautam1858:
Calling x and y outside once, it will be clean and helpful for the programmers to understand quickly, x = mnist.train.images and y = mnist.train.labels has been called multiple times instead calling those once with the variable x and y assigned would be helpful for programmers


gautam1858:
Sorry made a mistake

PR #7667, opened 2017-02-19
[GO] Make utility for ops public

Lewuathe:
Ops in util_test cannot be used because these are _test.go files. Since the usage
of these ops utils are writen in API docs, it is better to make it public.

### DIFF
--- a/tensorflow/go/session_test.go
+++ b/tensorflow/go/session_test.go
@@ -187,7 +187,7 @@ func ExamplePartialRun() {
 		// Create a graph: a + 2 + 3 + b.
 		//
 		// Skipping error handling for brevity of this example.
-		// The 'op' package can be used to make graph construction code
+		// The package provides utilities used to make graph construction code
 		// with error handling more succinct.
 		g        = NewGraph()
 		a, _     = Placeholder(g, "a", Int32)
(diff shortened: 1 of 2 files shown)


asimshankar:
Thanks for the change @Lewuathe, but I'd actually prefer to _not_ export these functions.

Users of the API should use the ['op' package](https://godoc.org/github.com/tensorflow/tensorflow/tensorflow/go/op) to construct the graphs. The functions in `util_test.go` are there only for unittests of the `tensorflow` package to avoid any circular dependencies (the 'op' package depends on the 'tensorflow' package and I didn't want the reverse dependency just for unittests).

Let me know if I misunderstood something.

Thanks!

Lewuathe:
@asimshankar Thanks for comment. I understand.
But [example of partial run](https://godoc.org/github.com/tensorflow/tensorflow/tensorflow/go#example-PartialRun) in godoc uses these util test APIs. It was confusing to me because interface is different. So how about rewriting this example?

asimshankar:
Ah, I can see how that is confusing. Feel free to suggest rewrites to the example, otherwise I can take a stab at it later  (hard to balance between keeping the example focused and not having to depend on the 'op' package :).

asimshankar:
But I'll go ahead and close this Pull Request for now, since it seems we've agreed to try something else.

Lewuathe:
@asimshankar Thanks. I'll update example for partial run later.

Issue #7615, opened 2017-02-17
tfdbg Dump root directory does not exist


XiaolongMeng:
I run 
python -m tensorflow.python.debug.examples.debug_mnist --debug
on my laptop.which has windows 10 and gtx1070.but after I type first run,it return an exception like this:
Traceback (most recent call last):
  File "d:\Anaconda3\lib\runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "d:\Anaconda3\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "d:\Anaconda3\lib\site-packages\tensorflow\python\debug\examples\debug_mnist.py", line 138, in <module>
    tf.app.run()
  File "d:\Anaconda3\lib\site-packages\tensorflow\python\platform\app.py", line 43, in run
    sys.exit(main(sys.argv[:1] + flags_passthrough))
  File "d:\Anaconda3\lib\site-packages\tensorflow\python\debug\examples\debug_mnist.py", line 131, in main
    acc = sess.run(accuracy, feed_dict=feed_dict(False))
  File "d:\Anaconda3\lib\site-packages\tensorflow\python\debug\wrappers\framework.py", line 419, in run
    run_end_resp = self.on_run_end(run_end_req)
  File "d:\Anaconda3\lib\site-packages\tensorflow\python\debug\wrappers\local_cli_wrapper.py", line 262, in on_run_end
    self._dump_root, partition_graphs=partition_graphs)
  File "d:\Anaconda3\lib\site-packages\tensorflow\python\debug\debug_data.py", line 328, in __init__
    raise IOError("Dump root directory %s does not exist" % dump_root)
OSError: Dump root directory C:\Users\tony8\AppData\Local\Temp\tfdbg_cvtcdnhl does not exist

how can I fix it?Thank you

caisq:
Thanks for reporting this issue. I can reproduce it in 1.0.0 on Windows. It appears that this is because the core/debug module was not linked for the Windows cmake build (despite the fact that it is built for the Windows bazel build, but cmake is the one used in building the release binary). I'll fix it and the fix will be reflected in nightly builds and the next release. Sorry about the inconvenience it has caused.

cc @mrry, @gunan

caisq:
This is fixed by https://github.com/tensorflow/tensorflow/commit/d82e42ca3a0b06d768a439d84ff74741de7e092d

[...]

PR #7670, opened 2017-02-19
Fix tfdbg in windows cmake build

caisq:
Add tfdbg to cmake build
    
    Guard grpc:// debug URL scheme implementation out from open-source.
    The service side of the grpc:// debugging is not available in
    open-source anyway, due to the lack of an open-source Python gRPC
    iibrary genrule. See b/23796275.
    
    Added C++ core/debug library, Python modules and unit tests of tfdbg.
    chaser


caisq:
Fixes issue https://github.com/tensorflow/tensorflow/issues/7615

caisq:
Tested with experimental builds (Jenkins login required to view logs):
Windows cmake CPU build:
http://ci.tensorflow.org/view/Experimental/job/experimental-cais-windows-cmake/16/

Windows cmake GPU build:
http://ci.tensorflow.org/view/Experimental/job/experimental-cais-windows-cmake-gpu/6/

caisq:
cc @gunan @yifeif

tensorflow/core/debug/debug_io_utils.cc
```diff
@@ -147,6 +146,7 @@ Status DebugIO::PublishDebugMetadata(
   Status status;
   for (const string& url : debug_urls) {
     if (str_util::Lowercase(url).find(kGrpcURLScheme) == 0) {
+#if defined(PLATFORM_GOOGLE)
```
mrry: Doesn't this break existing open-source code, or was this path inaccessible before?

mrry (changes requested):

tensorflow/core/debug/debug_io_utils.cc
```diff
@@ -163,6 +163,11 @@ Status DebugIO::PublishDebugMetadata(
 
       status.Update(
           DebugGrpcIO::SendEventProtoThroughGrpcStream(grpc_event, url));
+#else
+      return Status(error::UNIMPLEMENTED,
```
caisq: Done.

tensorflow/core/debug/debug_io_utils.cc
```diff
@@ -213,13 +218,19 @@ Status DebugIO::PublishDebugTensor(const string& tensor_name,
...
+#if defined(PLATFORM_GOOGLE)
       Status s = DebugGrpcIO::SendTensorThroughGrpcStream(
           node_name, output_slot, debug_op, tensor, wall_time_us, url);
 
       if (!s.ok()) {
         num_failed_urls++;
         fail_statuses.push_back(s);
       }
+#else
+      return Status(error::UNIMPLEMENTED,
```
caisq: Done.

tensorflow/core/debug/debug_io_utils.cc
```diff
@@ -264,8 +275,14 @@ Status DebugIO::PublishGraph(const Graph& graph,
       status.Update(
           DebugFileIO::DumpEventProtoToFile(event, dump_root_dir, file_name));
     } else if (debug_url.find(kGrpcURLScheme) == 0) {
+#if defined(PLATFORM_GOOGLE)
       status.Update(
           DebugGrpcIO::SendEventProtoThroughGrpcStream(event, debug_url));
+#else
+      return Status(error::UNIMPLEMENTED,
```
caisq: Done.

tensorflow/core/debug/debug_io_utils.cc
```diff
@@ -275,7 +292,13 @@ Status DebugIO::PublishGraph(const Graph& graph,
 // static
 Status DebugIO::CloseDebugURL(const string& debug_url) {
   if (debug_url.find(DebugIO::kGrpcURLScheme) == 0) {
+#if defined(PLATFORM_GOOGLE)
     return DebugGrpcIO::CloseGrpcStream(debug_url);
+#else
+    return Status(error::UNIMPLEMENTED,
```
caisq: Done.

tensorflow/core/debug/debug_grpc_io_utils_test.cc
```diff
@@ -25,6 +25,9 @@ limitations under the License.
 #include "tensorflow/core/platform/env.h"
 #include "tensorflow/core/util/event.pb.h"
 
+// TODO(cais): Support grpc:// debug URLs in open source once Python grpc
+//   genrule becomes available. See b/23796275.
+#if defined(PLATFORM_GOOGLE)
```
caisq: The problem is that there doesn't seem to be a macro that I can use for cmake builds specifically. So if I just excluded this test from tf_tests.cmake, the test would still be run in Linux cmake builds and fail because of the PLATFORM_GOOGLE macro I applied in debug_io_utils.cc. Do you know of a macro defined specifically for cmake builds? If not, I think the approach here is the cleanest. This test used to run and pass in open-source bazel tests (e.g., on Linux and Mac), but since the server side of grpc:// debug URLs is not defined in open source anyway, open-source users aren't able to use the feature yet anyway. I don't think it's going it's going to cause any problems to exclude it for now.

tensorflow/core/debug/debug_io_utils.cc
caisq: grpc:// debug URL implementation has two parts: client (debug ops) and server. The server part, which is defined as a Python grpc library, is currently not available in open source. See b/23796275 cited below. So even though the client side has been available in OSS for a while, OSS users haven't been able to use the feature yet to the lack of a working server side. As such, the change here shouldn't affect them.

tensorflow/core/debug/debug_grpc_io_utils_test.cc
mrry: There's no `#define` for CMake builds, but I don't see why you'd need one.

Excluding this test in `tf_tests.cmake` (e.g. by adding its filename as another line [here](https://github.com/tensorflow/tensorflow/blob/43c71a03380d8de18202cc399563814b2f438cd2/tensorflow/contrib/cmake/tf_tests.cmake#L241)) would prevent it being run in the CMake on *any* platform. (Are you instead concerned about the Linux/Mac *Bazel* tests? I guess these would start failing without this `#if`....)

I still think it would be cleaner to exclude the test explicitly in the CMake/Bazel build files, rather than building a test that "passes" but doesn't do anything.

caisq: Yeah - the concern about excluding it using tf_tests.cmake is Bazel OSS builds, because the non-test code in debug_io_util.cc are guarded with PLATFORM_GOOGLE. I guess it's better to submit this change from internal remove this test file from OSS altogether for now. WDYT?

mrry: Yep, that sounds like the best plan! (I don't know if we have a policy on `#if PLATFORM_GOOGLE` code in OSS, but it seems like it'd be best to minimize it.)

caisq:
OK. I'll close this PR and do it from internal.

PR #7676, opened 2017-02-19
Add .shape property to Variable object

yaroslavvb:
also adjust docstrings for consistency with Tensor.shape/Tensor.get_shape


vrv:
If this acceptable, I'd suggest adding a test of the new property, at the very least.

yaroslavvb:
I'll add tests if the API change goes through. The motivation for this is that Variables/Tensors are largely exchangeable and we had users confused to why "a.shape" fails for some a

yifeif (approved):
I dont know what's the formal procedures for PR API change review, but the change itself LGTM.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

PR #7691, opened 2017-02-20
Update stat_summarizer_test.py

gautam1858:
Optimizing and easy to understand the code, I have set "sa = self.assertRegexpMatches" and have called the variable sa later instead of "self.assertRegexpMatches" many times


gautam1858:
Anyone ?

maciekcc:
This style: 'self.assertRegexpMatches' is used throughout the whole repository where test.TestCase us used. Other assertions, e.g. 'assertEqual', 'assertFalse' etc... also follow this kind of explicit calling, let's keep it this way.

Issue #7225, opened 2017-02-02
Feature request: equivalent of np.clip


yaroslavvb:
`tf.clip_by_value` is missing an important property of `np.clip` which lets upper/lower bounds to be tensors, whereas `tf.clip_by_value` only takes scalars

A work-around is `tf.minimum(upper, tf.maximum(lower, x))` but that uses 2x memory for gradients. Another potential work-around is to use `tf.map_fn` + `tf.clip_by_value`, but it is orders of magnitude slower. Some profiling: https://github.com/yaroslavvb/notebooks/blob/master/clipping-profile.ipynb

goodfeli:
+1

jmchen-g:
Assigned to @vrv mainly because you wrote the original clip_by_value(). Feel free to re-assign. Thanks.

vrv:
@yaroslavvb if you look at the implementation of clip_by_value, it is exactly doing ```tf.minimum(upper, tf.maximum(lower, x))```, so presumably we could change the documentation to not mention it's only scalar arguments.

Your gradient argument seems reasonable, I'm going to assume that XLA will help address these issues instead of writing custom kernels.

However, if someone wants to write a custom CPU and GPU kernel for clip_by_value that fuses it in the short term, that would be a nice intermediate contribution.

(p.s., I actually didn't write this op).

yongtang:
Added a PR #13998 to add the customized kernel for `tf.clip_by_value`. Please take a look.

PR #7692, opened 2017-02-20
Update doc string to indicate clip_by_value accepts tensors as min and max arguments too

ankeshanand:
`tf.clip_by_value`'s current documentation mentions that it accepts only scalars as min and max arguments, but it can accept tensors too. This PR fixes that, and resolves the confusion raised in #7225


ankeshanand:
cc @yaroslavvb @vrv

vrv (approved):

PR #7710, opened 2017-02-20
[WIP] MPI Support for Tensor data communication

jbedorf:
This adds a second communication path to the distributed TensorFlow implementation, next to the gRPC communication path. This new communication path uses the Message Passing Interface (MPI) API to handle communication between processes allowing TensorFlow to take advantage of modern high performance networks such as Infiniband.
This pull request is used to start a discussion about this implementation, any draw backs, alternatives and requests for feedback and participation from the open source community.

**The current pull request makes the following changes**
- A new communication path has been added for sending and receiving Tensors between distinct processes.
- This new communication path uses the MPI API to handle setting up the connections and the data transfer.
- When using a CUDA-Aware (and/or GPUDirect RDMA) MPI implementation this path works for both CPU and GPU based Tensor data. 

This new path is implemented by modifying the current 'send' and 'receive' operations. The original gRPC code stays in place and is responsible for setting up connections and all non-tensordata communications. 
Although MPI supports one sided RDMA operations, this example implementation uses plain blocking send/receive operations. This because the current TensorFlow (memory) model makes it difficult to efficiently implement the communication using MPI_Get/MPI_Put (see below). 

**Usage**
In order to setup the required MPI environment the Python instances have to be launched using 'mpirun'. An example launch script, which does not modify the original TF script, is supplied in: tensorflow/tools/dist_test/mpi/start-openmpi.sh

The launching can be done differently if one handles the task type/id selection inside the TF python script, by either reading the environment variables as set by mpirun or by using the mpi4py Python library to retrieve process IDs and hostnames. 

The MPI execution path can be disabled by setting the environment variable 'MPI_PATH_DISABLED' to 1, like:
export MPI_PATH_DISABLED=1

**MPI Requirements**
- The MPI implementation should be built with support for MPI_THREAD_MULTIPLE
- To enable this path for GPU data, the MPI implementation should be built with CUDA support (CUDA-Aware MPI)
- Tested using OpenMPI-2.0.1 
- Currently the path to the OpenMPI library is hard-coded in the "third_party/gpus/crosstool/CROSSTOOL.tpl" file, this should be made an option in the configure code. 

**Implementation details**

**MPI process ID mapping to gRPC names**
To handle communication between processes MPI identifies processes using a unique process-ID. TensorFlow uses gRPC which uses names based on things like 'worker', 'gpu', 'task'. To enable the MPI path in co-existence with the gRPC communication stack the names of the gRPC 'server' are mapped to the unique MPI IDs during the initialization of the gRPC stack. This conversion is then available once the tensor-data will be communicated.

**Send details**
The original 'send' operation places a tensor in a table which is then picked up by the gRPC thread once a request for that data arrives. In the MPI path we place an MPI_Send call which will block until the matching MPI_Recv call is made on the receiving side. By making a hash of the tensor description we create a unique combination of sending-process/tensor-id which will be matched by the receiving process. This enables multiple tensors to be in flight from a single process using different send threads. Sending is a two-step process, the first message contains the properties of the Tensor (data type & shape) followed by a message containing the actual data. 
The send call is intercept in the file: 'tensorflow/core/distributed_runtime/base_rendezvous_mgr.cc'
In this function: BaseRemoteRendezvous::Send
The code verifies that the destination is a different process from the sending process, if this is the case (and the implementation is enabled at run-time) then it will enter the new 'SendToRemote' functions.  Otherwise it will progress through the original code.
To keep the changes as organized as possible, the actual implementation of the sending-functions are in:  tensorflow/core/distributed_runtime/rpc/grpc_remote_worker.cc
The other file modifications are required to get the path enabled through out the calling stack.

**Receive details**
The original 'receive' operation requests a named-tensor from a remote process by placing a request through the gRPC stack. In the MPI path this request has been replaced by a set of receive calls that match the send operations. If there is no matching send operation yet, then the path will block until the sending process has arrived at the same point. The send and receive operations are matched using a hash of the requested tensor name. 
The first message will describe the tensor, this description is passed on to the memory allocation functions after which the actual tensor data flows directly into the newly allocated memory buffer.
All the modified receive functionality is in: tensorflow/core/distributed_runtime/rpc/grpc_remote_worker.cc
By inheriting the 'GrpcRemoteWorker' and replacing the 'RecvTensorAsync' function. 

Note: It is possible that a tensor with the same name is send (received) multiple times between the same set of processes. This is not a problem as messages are handled as unique units and have unique follow up IDs (based on thread IDs).

**TensorFlow limitations**
- Memory is constantly being (de)allocated in between iterations. This hinders the performance when using Infiniband connections as memory buffers must be pinned/mapped before they can be used. This can impact the effective bandwidth by a factor 3 as this pinning is a relatively costly operation. 
- It would be more efficient if memory buffers for the same tensors would be reused/retained in between iterations, this makes the pinning of memory buffers a one-time occurrence. It also allows for the removal of the allocation step in the 'receive' functions.

Minds.ai
Jeroen Bédorf


jbedorf:
I signed it!

martinwicke:
What's the status of this? Is it still WIP? @poxvoculi are you reviewing it?

poxvoculi:
Yes, turns out I'm reading it right now.  Note that there's overlap in functionality with 2 other PRs, so I'm looking at them all together.

byronyi:
Which are two other PRs being reviewed? Appreciate if you could point me there.

poxvoculi:
@byronyi 

One is limited to AllReduce over MPI, but has a different approach to MPI integration: https://github.com/baidu-research/tensorflow-allreduce/pull/2

The other one is not yet a PR, but probably will generate one.  It bypasses MPI and gets TF to work directly on verbs:  https://github.com/yahoo/tensorflow/tree/22fd3aad2dac96d521b5224519f443cef5269efb/tensorflow

tensorflow/core/distributed_runtime/rpc/grpc_remote_worker.cc
```diff
@@ -260,10 +288,206 @@ class GrpcRemoteWorker : public WorkerInterface {
...
+
+  void RecvTensorAsync(WorkerEnv* env, CallOptions* call_opts,
+                       const RecvTensorRequest* request,
+                       TensorResponse* response, StatusCallback done) override {
+    if (disabledAtRuntime) {
+      GrpcRemoteWorker::RecvTensorAsync(env, call_opts, request, response,
+                                        done);
+    } else {
+      // TODO: Figure out how to get the size of the requested Tensor
+      // Is this known by looking up the key/request?
```
byronyi: I guess you could keep the gRPC path, and send partial RPC reply with no actual tensor content, which includes tensor shape and data type information (so size could be computed).

tensorflow/core/distributed_runtime/rpc/grpc_remote_worker.cc
```diff
@@ -260,10 +288,206 @@ class GrpcRemoteWorker : public WorkerInterface {
...
+      // TODO is it possible that a tensor has been pre-allocated?
+      // so we can reuse the same (mapped) pointer?
```
byronyi: I am actually wondering if we should use a customised `Allocator` here, just to keep track of tensors from/to `SendOp` and `RecvOp`.

tensorflow/core/distributed_runtime/rpc/grpc_remote_worker.cc
jbedorf: Yes that's how it's currently implemented. Knowing that info at forehand would save a communication message which saves on latency.

byronyi: Considering that the shape of the same tensor (with same name, for example) may not change across different iterations, it might be helpful to cache the size information to reduce latency after the 1st epoch.

I am wondering if the size info could be obtained by inspecting the computational graph as well.

jbedorf: Yes that might work but it should not be stored inside the communication library. If, for example, the tensors are modified between iteration runs than the library will not catch that change which leads  to undefined behaviour. 
I assume the compute graph has this info but it is not passed down to the communication routines, so for now I rather do the extra step. For future versions it could be an optimization to cache that info if there is a good way to verify that it has not changed since previous iterations

poxvoculi: This situation illustrates an advantage of programming to a lower-level interface like IB/verbs.  The client can send a request for the tensor not needing to know its size, then the response comes in two parts: meta-data including the size, followed by data bytes only when the client has ack'd that a receive buffer is ready.  Does MPI not have a way of doing this?

byronyi: I believe that's the same with current gRPC/out-of-band approach, with 2 round trips: 1st round is for setting up the buffer and 2nd round is for actual data exchange. The latency comes not only from those round trips though; the buffer needs to be pinned after allocation, both of which cannot be done until after 1st round. If we could know the size beforehand, we could eliminate such overhead.

jeffhammond: You can do this without a round trip if you can pre-post a buffer that is larger than any you will receive, which is basically the eager protocol `MPI_Send` uses internally.  Or you can do the rendezvous approach mentioned already, which is of course also used by `MPI_Send` internally, although there are many different variants of rendezvous depending on what the hardware can do.  Depending on the usage, you can use both, as MPI does internally.

In addition to the obvious rendezvous protocol that uses two send-recv pairs, one can do a variant that uses `MPI_Get` at the target to pull data from the origin, which is how e.g. Blue Gene does rendezvous.  This assumes that `MPI_Get` is asynchronous, which is not always true.  I can work up examples for a bunch of different ways to implement this if people are interested.  Pseudo-code is easy, but I have on occasion written educational code that actually works 😃  ([example](https://github.com/jeffhammond/HPCInfo/blob/master/mpi/active-messages/armci.c))

In the end, what is needed here is an active-message interface, which MPI doesn't expose but often uses internally.  The MPICH team [prototyped](http://www.mcs.anl.gov/~thakur/papers/mpi-active-messages.pdf) an implementation of active-messages that extends the RMA interface, but it is obviously not portable to other MPI implementations.

byronyi: “You can do this without a round trip if you can pre-post a buffer that is larger than any you will receive” I don't think that approach would be appropriate here. If the buffer size does not match the exact size of the tensor, there is always gonna be one extra copy to pass it to the downstream of the pipeline. The exception would be when you decide to go ahead and implement your own memory allocator that is shared among transport and application...which is probably too intrusive for TF.

Posting a buffer that is of unlimited size also has the problem of DoS.

jbedorf: It is as Byronyi says, right now it's a two step process:
- After the initial request the first information to arrive will the tensor configuration (type,size) which in turn allows the buffer to be allocated on the receiving side.
- Followed by a message that contains the actual data.
The latency comes from the receiving side having to allocate the buffer, depending on from which pre-allocated buffer this comes you might have some pinning overhead.

On the sending side this all can be handled without too much latency by using non-blocking communication and possibly keeping a local copy of the data that is being send (done by default for some implementations of MPI_Send).

Then again all this latency is of minimal impact for large networks/buffers where other things than data sends form the majority of the execution time.

byronyi: One low-hanging fruit for optimisation would be decide when to send the tensor inline in the first round given a certain configurable size limit, when memory pinning and other overhead is significant compared to data transmission. This could reduce latency for the common case when the tensor is small, and we can always fall back to 2 rounds when it gets large enough.

jeffhammond: @byronyi `MPI_Send` automatically handles the latency trade-off associated with memory registration.  That's the whole point of eager.  If page-registration costs more than copying the buffer, MPI copies the buffer into a temporary owned by the MPI runtime that is already registered and fires that.  In this case, `MPI_Send` returns after the copy, in contrast to full rendezvous, where it waits for the acknowledgement that the data has been transferred to the network (which on an end-to-end network means it is buffered at the target).

From a communication runtime standpoint, MPI is high-level, which means that it makes a lot of protocol choices and optimizations for the user, and exposes only the primitive that the application needs (e.g. `MPI_Send`).  On the other hand, IB Verbs and other low-level protocols require the user to implement the protocol choices and optimizations.  Which one is more appropriate obviously depends on how much effort the user has to optimize such things for their workload.

byronyi: Thanks for your comments! Really helpful as I am not that familiar with MPI (I started from the verbs API and tend to think from that perspective).

One more question: I am wondering which verbs MPI_send/recv is implemented for RDMA transport, send/recv, read, or write?

third_party/mpi/BUILD
```diff
@@ -19,4 +19,5 @@ cc_library(
         "//conditions:default": ["libmpi.so"],
     }),
     hdrs = ["mpi.h", "mpi_portable_platform.h"],
```
wscullin: mpi_portable_platform.h is a header specific to OpenMPI

tensorflow/core/distributed_runtime/rpc/grpc_remote_worker.cc
jbedorf: In a future version it will have the option to send the tensor in a single step, not for latency per se (although this can be configured if needed), but for handling certain tensor types that can not be trivially copied.

The fact that MPI takes care of all these things in MPI_Send/MPI_Recv is the main reason that I chose for the MPI API. TensorFlow is likely be used on a variety of machines and configurations and it would not be doable to implement it with verbs such that it will be optimal on all these networks. Better to let MPI implementations take care of that as they have been doing this for decades.

byronyi: It might not make real difference performance-wise, but what is the rationale for you to not use `MPI_Probe` to receive message of unknown size?

jbedorf: This is already done, see this lines:
```
MPICheck(MPI_Mprobe(src, hash, MPI_COMM_WORLD, &msg, &status));
MPICheck(MPI_Get_count(&status, MPI_CHAR, &incSize));
std::vector<char> sta(incSize);
MPICheck(MPI_Mrecv(&sta[0], incSize, MPI_CHAR, &msg, &status));

third_party/mpi/BUILD
jbedorf: Correct, I have not enough experience with Bazel to let it determine at runtime/configuration time if the MPI implementation pointed to is OpenMPI or MPICH based. Therefore I just added it for the OpenMPI case which is what I used internally.

jeffhammond: There is no guarentee that this header will be available in the future, so it represents a portability problem on multiple levels.  I don't see why you need this header at all, but if you like it, include it TF.  The license is permissive and it is standalone (it is originally from GASNet, not Open-MPI).

drpngx:
@jbedorf could you merge and update?
@poxvoculi any comments?

jbedorf:
I'm working on a refactoring that is less intrusive to TensorFlow and solves some problems that surfaced with the current implementation. Once that is working I'll update this pull-request.

qiaohaijun:
this is a great work

vrv:
@jbedorf should we keep this PR open or do you think it would make sense to send a new PR?

poxvoculi:
@jbedorf told me he will be sending a new PR soon, so this one can be closed.

PR #7721, opened 2017-02-21
Remove outdated link to nightly-matrix-android builds

rbspark:
This has already been replaced with the nightly-android links.


andrewharp:
@rbspark The commit you're referencing here was already merged into master, so I think this will have no effect? (Just trying to make sure I understand what the intent here is)

maciekcc (approved):

PR #7730, opened 2017-02-21
[Java] [Feature] Tensor clone

EronWright:
Allows a tensor to be cloned (shallow), resulting in a new tensor
sharing the same underlying buffer.

The cloned tensor is still usable after the original is closed.  This
is possible because the buffer supports reference counting.

- Added `TF_Clone` to the C API.
- Added `Tensor::clone` to the Java API.

### DIFF
--- a/tensorflow/c/c_api.cc
+++ b/tensorflow/c/c_api.cc
@@ -217,6 +217,11 @@ TF_Tensor* TF_NewTensor(TF_DataType dtype, const int64_t* dims, int num_dims,
   return new TF_Tensor{dtype, TensorShape(dimvec), buf};
 }
 
+TF_Tensor* TF_CloneTensor(TF_Tensor* t) {
+  t->buffer->Ref();
+  return new TF_Tensor{t->dtype, TensorShape(t->shape), t->buffer};
+}
+
 void TF_DeleteTensor(TF_Tensor* t) {
   t->buffer->Unref();
   delete t;
--- a/tensorflow/c/c_api.h
+++ b/tensorflow/c/c_api.h
@@ -220,6 +220,9 @@ extern TF_Tensor* TF_NewTensor(TF_DataType, const int64_t* dims, int num_dims,
 extern TF_Tensor* TF_AllocateTensor(TF_DataType, const int64_t* dims,
                                     int num_dims, size_t len);
 
+// Clone a tensor.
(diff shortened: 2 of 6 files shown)


asimshankar:
Thanks for the change. Though, looking at the [API contract for `Object.clone()`](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#clone()), I'm not sure if this is appropriate. Particularly, the line:

By convention, the object returned by this method should be independent of this object (which is being cloned).

I realize that currently Tensor objects are essentially immutable, so it may not matter yet. But I suspect that this will change in the future (folks wanting to write directly to portions of the buffer) and then the fact that two clones share the same underlying buffer may become confusing.

IIRC, your motivation to do this stems from making scala bindings more idiomatic, where there are no `try-with-resources` blocks? Would a lightweight reference counting class over Tensor be something you could use in your code instead?

(FYI @jhseu)

EronWright:
I agree that tensor mutability would be incompatible with this, unless a copy-on-write approach were used under the hood.  Can you think of other cases of shared tensor buffers in the core?  

If impractical, we can close this PR and I'll try something else.

asimshankar:
I don't want to rule this out outright yet, but I would encourage you to see if there are other options that work well for your use case.

An alternative would be to add a finalizer to the `Tensor` class that does the cleanup, allowing users to avoid using `close()` and let the resources be reclaimed on GC, or introduce the reference counts as you had proposed earlier. I'm not sure there is any perfect solution.

EronWright:
I rely on a wrapper at this time, no worries you can close this.

It can't hurt to have a finalizer as a safety net.   But as you know it's dangerous to leave the cleanup of large unmanaged blocks of memory to the GC.  Apparently direct ByteBuffers create a similar situation (and [hacks abound](http://stackoverflow.com/questions/3496508/deallocating-direct-buffer-native-memory-in-java-for-jogl/26777380#26777380)).

I do like how Netty solved this problem ([ref](http://netty.io/wiki/using-as-a-generic-library.html#buffer-api)), and it inspired the approach in #6528.  Netty-style ref counting is the most powerful yet most convenient alternative that I know of.   Netty resembles a dataflow system in some ways, with its channel pipeline of buffers, further inspiring my earlier comments.

asimshankar:
Thanks for the comments. I'm going to close this for now, but perhaps we will revisit this or your ref change in the near future. 

Thanks!

PR #7736, opened 2017-02-21
Convert some tf2xla ops to use elementwise tensor XLA ops, not Map

DavidNorman:
Some binary operations were implemented as maps of a scalar computation over the input tensors.  Since the computations consist of operations that can be element-wise operations, they could also be implemented in this manner.

In the case of the CPU, it is highly likely that it is improved, as the element-wise tensor operations are highly optimized. I do not know about the GPU operations, but I suspect that they will also be optimized in the element-wise tensor form.

This change re-implements all of the ops in this class.


hawkinsp:
Jenkins, test this please.

hawkinsp (approved):
Looks good to me.

maciekcc:
Jenkins, test this please.

DavidNorman:
cheers.

PR #7752, opened 2017-02-21
Fix compile warnings

conqerAtapple:
Fixes:
  - Warnings related to comparison of signed vs unsigned types.
  - Initialized variables.

### DIFF
--- a/tensorflow/compiler/xla/service/cpu/ir_emitter.cc
+++ b/tensorflow/compiler/xla/service/cpu/ir_emitter.cc
@@ -506,7 +506,7 @@ Status IrEmitter::HandleReduceWindow(HloInstruction* reduce_window,
 
         llvm_ir::IrArray::Index input_index(index.size());
         llvm::Value* in_bounds_condition = nullptr;
-        for (int64 i = 0; i < index.size(); ++i) {
+        for (auto i = 0; i < index.size(); ++i) {
           llvm::Value* strided_index = ir_builder_.CreateNSWMul(
               index[i], ir_builder_.getInt64(window.dimensions(i).stride()));
           input_index[i] = ir_builder_.CreateNSWSub(
@@ -1111,7 +1111,7 @@ Status IrEmitter::HandleReduce(HloInstruction* reduce, HloInstruction* arg,
         llvm_ir::IrArray::Index input_index = reduced_dims_index;
         llvm_ir::IrArray::Index::const_iterator it = index.begin();
 
-        for (int64 i = 0; i < input_index.size(); ++i) {
+        for (auto i = 0; i < input_index.size(); ++i) {
           if (input_index[i] == nullptr) {
             input_index[i] = *it++;
           }
@@ -1180,7 +1180,7 @@ Status IrEmitter::HandlePad(HloInstruction* pad) {
   // output_index := edge_padding_low + operand_index * (interior_padding + 1)
   const PaddingConfig& padding_config = pad->padding_config();
   llvm_ir::IrArray::Index output_index;
-  for (int64 i = 0; i < operand_index.size(); ++i) {
+  for (auto i = 0; i < operand_index.size(); ++i) {
     llvm::Value* offset = ir_builder_.CreateMul(
         operand_index[i],
         ir_builder_.getInt64(padding_config.dimensions(i).interior_padding() +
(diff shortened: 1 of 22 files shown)


conqer:
@martinwicke looks like the jenkins checks are not being run.

martinwicke:
Chad can you look at this? I think issues with it may be mostly cosmetic.

martinwicke:
jenkins, test this please.

conqer:
@martinwicke please review

tensorflow/compiler/xla/service/cpu/ir_emitter.cc
```diff
@@ -506,7 +506,7 @@ Status IrEmitter::HandleReduceWindow(HloInstruction* reduce_window,
 
         llvm_ir::IrArray::Index input_index(index.size());
         llvm::Value* in_bounds_condition = nullptr;
-        for (int64 i = 0; i < index.size(); ++i) {
+        for (auto i = 0; i < index.size(); ++i) {
```
cwhipkey: curious - I was going to ask earlier if "auto" would work for some earlier cases, but didn't think it would (I assumed it would derive the type from the i=0 assignment).  Do you know what the auto type here is?  I guess "int"?

tensorflow/contrib/ffmpeg/default/ffmpeg_lib.cc
```diff
@@ -69,7 +69,10 @@ bool IsBinaryInstalled(const string& binary_name) {
   for (const string& dir : str_util::Split(path, ':')) {
     const string binary_path = io::JoinPath(dir, binary_name);
     char absolute_path[PATH_MAX + 1];
-    ::realpath(binary_path.c_str(), absolute_path);
+    if (::realpath(binary_path.c_str(), absolute_path)) {
```
cwhipkey: should this be comparing to nullptr?  It seems here that we will reject valid paths.

tensorflow/stream_executor/dnn.cc
```diff
@@ -122,7 +122,11 @@ string PadAlignmentString(PadAlignment alignment) {
       return "cuDNN padding";
     case PadAlignment::kTensorFlowPadding:
       return "TensorFlow padding";
+    default:
+      LOG(FATAL) << "Unknown padding alignment "
```
cwhipkey: nit: maybe just fall through instead of log fatal (e.g., in stream.cc change below, you don't LOG(FATAL) in the default case).  Having a default: case will make it so that a new PadAlignment value won't trigger a warning here for not handling all cases...

conqer: Good point. I was trying to follow the style in the file. For example: in the methods 'FilterLayoutString' and 'ShortPoolingModeString', the default case is LOG(FATAL).

tensorflow/contrib/ffmpeg/default/ffmpeg_lib.cc
conqer: Good catch. I intended to check for NULL (not nullptr).

tensorflow/compiler/xla/service/cpu/ir_emitter.cc
conqer: Will fix it to be explicitly 'size_t'

conqer:
@cwhipkey Thanks for reviewing. Addressed review comments.

cwhipkey (approved):

cwhipkey:
jenkins, test this please

conqer:
@cwhipkey merge conflicts resolved

cwhipkey:
Jenkins, test this please

cwhipkey:
@andrewharp - can you merge this in?

tatatodd:
FYI this PR has a typo in it, which breaks XLA builds.  The typo is `size_typ` rather than `size_type`:
https://github.com/tensorflow/tensorflow/blame/master/tensorflow/compiler/tf2xla/xla_compiler.cc#L228

I think we need a better process for testing such changes.  I'm suspecting that "jenkins, test this please" isn't actually compiling / testing XLA?  @gunan might have an opinion here.

gunan:
Correct, XLA is not tested in standard PR tests, we should add that.

PR #7758, opened 2017-02-22
Make the comment explaining arg_scope clearer

nealwu:
I found the use of "..." in a comment providing an example of the functionality of arg_scope to be very confusing, so I expanded it out to be less ambiguous.

### DIFF
--- a/tensorflow/contrib/framework/python/ops/arg_scope.py
+++ b/tensorflow/contrib/framework/python/ops/arg_scope.py
@@ -53,7 +53,7 @@
     net = layers.conv2d(net, 256, [5, 5], scope='conv2')
   ```
 
-  Example of how to use tf.contrib.framework.add_arg_scope:
+  Example of how to use tf.contrib.framework.add_arg_scope to enable your function to be called within an arg_scope later:
 
   @tf.contrib.framework.add_arg_scope
   def conv2d(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


maciekcc:
Sorry that I didn't reply earlier. I think this change will lead to more problems, you will need to keep the inlined default values synced with the place where they are defined - and this is a thing people for sure will forget.

nealwu:
Hi @maciekcc, I'm not sure I understand what you mean. You're saying that with future changes to this explanatory comment, people will modify one part and forget to modify the other affected parts? That problem should be handled by future code reviews.

On top of that, the current explanation had little value to me, as the "..." omitted the very part I was trying to understand. If the explanation had included these changes when I first looked at it, it would have been much easier to understand, and I expect that to be true for others as well.

martinwicke:
The change doesn't make it much worse: the `padding=` argument in the docstring already depends on the value in the code, and would have to be changed along with any changes in the function.

You could also omit the ... entirely, since it is irrelevant -- default args are used anyway, so not additional args need to be set.

nealwu:
@martinwicke Right, I'm assuming the docstring is meant to explain how the functionality works, in which case expanding out the full functionality makes it clearer. As someone who didn't understand how `arg_scope` works (and maybe still doesn't), I found the existing docstring to be very unhelpful for figuring it out.

nealwu:
@martinwicke @maciekcc Any thoughts on this?

jhseu:
(On sync rotation, clearing out old pull requests). Seems strictly better than current docs, so merging.

jhseu (approved):

PR #7761, opened 2017-02-22
Use Ubuntu 14.04 for the 1.0 branch GPU build.

jhseu:
nvidia/cuda:8.0-cudnn5-devel currently points to Ubuntu 16.04 which causes it to fail on Jenkins. This change pins to the same Ubuntu version we used for the 1.0 release.

Tested with docker build.


jhseu:
Looks like that image installs CUDA in a non-standard location. Abandoning for now.

jhseu:
Jenkins, test this please

jhseu:
@caisq ok, looks like this fix worked. PTAL.

maciekcc (approved):

Issue #7251, opened 2017-02-03
Native GPU version of `tf.dynamic_stitch`


drasmuss:
### Environment info
Operating System: Windows 10

Installed version of CUDA and cuDNN: 8.0, 5105
tensorflow release 0.12.1

### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)
``` python
import tensorflow as tf
from tensorflow.python.client.timeline import Timeline

with tf.device("/gpu:0"):
    x = tf.ones(100)
    idxs = tf.range(100)

    for _ in range(10):
        y = tf.identity(x)
        x = tf.dynamic_stitch([idxs, idxs], [x, y])
        # x = tf.gather(y, idxs)

with tf.Session(config=tf.ConfigProto(log_device_placement=True)) as sess:
    metadata = tf.RunMetadata()
    sess.run(x, options=tf.RunOptions(trace_level=tf.RunOptions.FULL_TRACE),
             run_metadata=metadata)

timeline = Timeline(metadata.step_stats)
with open("profile.json", "w") as f:
    f.write(timeline.generate_chrome_trace_format())
```

The `log_device_placement` output shows that everything is assigned to the GPU, as expected.  However, inspecting the trace output shows that data is being copied on and off the GPU for each call to `dynamic_stitch`.  This is something specific to the `dynamic_stitch` implementation, because using `tf.gather` (a similar indexed read operation, and functionally equivalent in this case), doesn't show this behaviour.

Is this intended behaviour for `dynamic_stitch` (i.e., the copying to and from the GPU is necessary)?  Or is this a bug?  If it isn't a bug, is there some equivalent solution that doesn't require the data to be copied back and forth?

yaroslavvb:
This is probably due to the following `HostMemory` annotations in definition of DynamicStitch in [dynamic_stitch_op.cc](https://github.com/tensorflow/tensorflow/blob/27711108b5fce2e1692f9440631a183b3808fa01/tensorflow/core/kernels/dynamic_stitch_op.cc#L184)

```
#define REGISTER_DYNAMIC_STITCH_GPU(type)                \
  REGISTER_KERNEL_BUILDER(Name("DynamicStitch")          \
                              .Device(DEVICE_GPU)        \
                              .TypeConstraint<type>("T") \
                              .HostMemory("indices")     \
                              .HostMemory("data")        \
                              .HostMemory("merged"),     \
                          DynamicStitchOp<type>)
```

[...]

PR #7764, opened 2017-02-22
native Cuda kernel for tf.dynamic_stitch op

MycChiu:
Correspond to issue #7251 
Here is the comparison between 
my implementation:
dynamic_stitch_gpu
and current implementation:
dynamic_stitch_cpu
generated by Nvidia visual profiler with [this code](https://gist.github.com/MycChiu/c353fd3c98c47dceae866ed30f40a5a8). *Since the kernel uses `CudaDeviceArrayStruct`, which depends on code from `//tensorflow/core:lib`, I can't generate custom_op library and use the custom_op for performance comparison. I ended up manually switch the tensorflow directory python uses between runs*
Currently, I still have two problems:
1. For some reason, tensorflow won't use the GPU kernel automatically, I had to force it with `tf.device`. What's even worse, tensorflow seems to refuse loading the GPU kernel in `test_session()` even if I tried to force it with `tf.device`, so this code is currently not tested with `dynamic_stitch_test.py`.
~~2. As @yaroslavvb suggested, Cuda kernels need to be real-valued, so I ended up calling `TF_CALL_GPU_NUMBER_TYPES_NO_HALF(REGISTER_DYNAMIC_STITCH_GPU);` to register the kernel. However, I feel like this will break some people's current code if their input data is from the unsupported data type.~~ *fixed in the latest commit.*

Finally, this implementation is really under-optimized for cases where the size of each input data slice is small, because the indices are still processed by the CPU (plus CUDA reads data in warps of 32 threads). However, I still don't think it will be worse than processing everything on CPU then copy them back to GPU though.


dandelionmane:
Not sure who the best reviewers are - added @girving and @yuanbyu, feel free to hot potato away if you're not the right reviewer. But we've let this sit for 14 days without any attention so we really should take a look promptly.

girving:
I can review this.  Can you squash into a single CL before I start, since it looks like you undid some mistakes?

MycChiu:
@girving Yes, I redirected inputs with datatypes not supported by current CUDA kernel implementation to CPU in the latest commit, but I think the latest commit reflects all the changes I made though? Or perhaps you and I are talking about different things?

girving:
Squashing won't change the final diff, but it makes it easier to read all the changes together as I review them.  So please squash as I requested.

MycChiu:
@girving Welp... Sorry, I think I F'ed up the rebasing part...I think it will be easier if I close this pull request and start a new one. I will tag you when I am done.

yaroslavvb:
@MycChiu I believe you did `git pull` into your PR branch. Never do a `git pull`, only rebase :)

MycChiu:
@yaroslavvb Haha, yes, that's indeed what I did. I will definitely keep this in mind next time, thank you!

PR #7777, opened 2017-02-22
Make ExamplePartialRun hidden

Lewuathe:
This is follow up of https://github.com/tensorflow/tensorflow/pull/7667

Since using op package in tf package causes cyclic dependencies,
using wrapper ops in `util_test.go`  is required. But the interface of
these ops are different to public one.
In order to avoid confusing of API usage, it is better to hide ExamplePartialRun
and add note of reference to `op` document.

### DIFF
--- a/tensorflow/go/session_test.go
+++ b/tensorflow/go/session_test.go
@@ -182,13 +182,16 @@ func TestConcurrency(t *testing.T) {
 	}
 }
 
-func ExamplePartialRun() {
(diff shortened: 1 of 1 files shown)


jhseu:
I'd prefer to leave it as an example perhaps with a comment explaining that you should use the ops from the ops package instead.

jhseu:
Closing for now, but feel free to make a new pull request if you'd like to add a comment for clarification.

PR #7797, opened 2017-02-22
Change pip.sh to take Bazel flags directly so AVX packages are built properly.

jhseu:
Also fix EXTRA_ARGS typo.

Haven't tested yet, will try on Jenkins.


jhseu:
Also tested locally

caisq (approved):

PR #7798, opened 2017-02-22
Fix nccl when referenced from a different repo

jart:
This fixes the following error:

ERROR: external/nccl_archive/BUILD.bazel:33:1: no such package 'tensorflow': Package crosses into repository `@org_tensorflow` and referenced by `'@nccl_archive//:nccl'`.

Also mirror URL for great justice.

CC @cwhipkey


girving (approved):

cwhipkey:
Thanks for fixing this!

PR #7799, opened 2017-02-22
Assert Bazel version from tf_workspace()

jart:
This ensures that projects depending on TensorFlow, e.g. TFServing, will
benefit from the version check.

The mandatory version has also been bumped up to Bazel 0.4.4 which has been out
for 21 days. It includes numerous very important fixes. TensorFlow also appears
to not work without it.

CC: @martinwicke


girving:
Leaving the test failures aside, the details of the change look reasonable.  However, @martinwicke's a better person to answer whether we want to force people to upgrade.  It may be that Bazel 0.4.4 is only required by folks using TensorFlow as a submodule or via `http_archive`.

One cludgy option if forcing everyone to upgrade is bad: we could require 0.4.4 for anyone calling `tf_workspace` *except* for `tensorflow/WORKSPACE` itself, via an extra boolean flag to `tf_workspace`.

martinwicke:
We've traditionally been tracking bazel close to head, so I have no problem requiring fresh bazel versions. Historically, the threshold has been "Is it in homebrew yet?", which has usually happened a few days after release. 

So I like this change, but it does mean that we need to upgrade our test machines to have the new bazel (i.e. change the bazel version in ci_build/install/install_bazel.sh, and then wait while we update all the other machines that don't use docker before we merge this.

girving:
0.4.4 is in Homebrew.  I just updated to it on my laptop with `brew update bazel`.

girving (approved):
LGTM once the test machines are updated.

gunan:
0.4.4 breaks TF build, so we cannot use it. (There is a post mortem for the release).
We have to wait for 0.4.5 to bump our bazel dependency.

jart:
What happened?

girving:
If Bazel 0.4.4 is broken, this is problematic for deepmath including tensorflow as a submodule.  Using it as a submodule does not work with 0.4.3.  Is there a workaround that will make 0.4.4 work?  Otherwise I'll end up blocked for a long time.

martinwicke:
0.4.5 RC came out yesterday, so it may not be a long time.

maciekcc:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

martinwicke:
Updated and testing.

PR #7806, opened 2017-02-23
Change pip.sh to take Bazel flags directly so AVX packages are built properly

jhseu:
Cherry-pick for r1.0 branch


jhseu:
Abandoning since it's not needed anymore. Built off of my own github branch since I needed another change there.

PR #7833, opened 2017-02-24
Add a non-debug avx2 fma build

jhseu:


caisq:
FYI, I may be a little slow in reviewing PRs in the next couple of days.

gunan:
I have some concerns about this cl.
1-ci_parameterized_build script is already trying to do too much. It is getting more and more confusing with each addition.
2- Im not sure if all our machines have avx2 (Some old machines on GCE, and we cannot really select CPUs.) I just checked, and we have a lot of builds running on Ivy bridge machines.

So I think, we should 
1-Get dedicated machines to run these builds.
2-Also, we can have a dedicated script under `ci_build/linux/optimized`?

jhseu:
1. Yeah, I agree. There's a lot of code that can be deleted in ci_parameterized_build right now (e.g., TF_BUILD_IS_OPT, TF_BUILD_MAVX now that TF_BUILD_OPTIONS works with pip.sh). I'd favor migrating this script to Python (all machines we use have Python preinstalled) so it's more maintainable, and then this list can just be a Python dictionary. I think cleaning up the script should be separated from this change, though.
2. Yeah, we should get AVX2/FMA-supported machines, but it's not an immediate priority because:
- You can build and just choose not to run tests on Jenkins. That's what I'm doing right now.
- Some tests fail with AVX2/FMA anyway because people are hardcoding epsilons in tests that are too small. We need an ongoing AVX2/FMA test somewhere.
- Testing manually doesn't take much time.

tensorflow/tools/ci_build/ci_parameterized_build.sh
```diff
@@ -83,8 +83,8 @@
 #                      support for Google Cloud Platform (GCP), which is
 #                      enabled by default.
 #   TF_BUILD_OPTIONS:
-#                     (FASTBUILD | OPT | OPTDBG | MAVX | MAVX2 | MAVX_DBG |
-#                      MAVX2_FMA_DBG)
+#                     (FASTBUILD | OPT | OPTDBG | MAVX | MAVX2 | MAVX2_FMA |
```
gunan: Let's instead change MAVX2 build to also enable FMA, they come together anyway.
To make the naming more understandable you can remove MAVX2

jhseu: Done.

gunan (approved):

PR #7882, opened 2017-02-25
Update DetectorActivity.java

qzw331:
Set default detector in DetectorActivity.java back to MultiBox


andrewharp:
@qzw331 be518926cd4bd5535a6b44c5542b605c5470e1a6 was committed to master back on Jan 11, I just hadn't deleted the andrewharp-patch-1 branch. I think we can close this, unless I'm missing something?

PR #7891, opened 2017-02-26
Improve tensorboard smoothing

jyegerlehner:
The smoothing of training loss in Tensorboard was frustrating; it would seemingly select a random set of unfiltered points near the end of the loss history leaving one unable to see the most recent trend.

I see that at least one other person was annoyed at this, as a change has been committed to master that changes the behavior. I think the change in this PR is superior. The previous one appears to have just truncated the smoothed curve, whereas this one will smooth all the way up until the end. When I'm 400K training iterations in, I find myself really wanting to see how the trend has gone in those last 20K iterations.

I also notice that the previous commit didn't update `dist/tf-tensorboard.html`, which I think is the only thing that affects most people running latest master.

### DIFF
--- a/tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
+++ b/tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
@@ -441,15 +441,14 @@ module VZ {
       let data = dataset.data();
       let kernelRadius = Math.floor(data.length * factor / 2);
       data.forEach(function (d, i) {
-          var start = Math.max(0, i - kernelRadius);
-          var end = Math.min(data.length-1, i + kernelRadius);
-          // Only smooth finite numbers.
-          if (!_.isFinite(d.scalar)) {
-              d.smoothed = d.scalar;
-          }
-          else {
-              d.smoothed = d3.mean(data.slice(start, end).filter(function (d) { return _.isFinite(d.scalar); }), function (d) { return d.scalar; });
-          }
+        var start = Math.max(0, i - kernelRadius);
+        var end = Math.min(data.length - 1, i + kernelRadius);
+        // Only smooth finite numbers.
+        if (!_.isFinite(d.scalar)) {
+          d.smoothed = d.scalar;
+        } else {
+          d.smoothed = d3.mean(data.slice(start, end).filter(function (d) { return _.isFinite(d.scalar); }), function (d) { return d.scalar; });
+        }
       });
 
     }
--- a/tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
+++ b/tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
@@ -441,8 +441,8 @@ module VZ {
       let data = dataset.data();
       let kernelRadius = Math.floor(data.length * factor / 2);
       data.forEach(function (d, i) {
-        var actualKernelRadius = Math.min(kernelRadius, i)
-        var start = Math.max(0, i - actualKernelRadius);
+        var actualKernelRadius = Math.min(kernelRadius, i);
+        var start = i - actualKernelRadius;
         var end = Math.min(data.length - 1, i + actualKernelRadius);
         // Only smooth finite numbers.
         if (!_.isFinite(d.scalar)) {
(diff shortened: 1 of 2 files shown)


jyegerlehner:
Whoopsiedaisy, apparently the change to vz-line-chart.ts and tf-tensorboard.html aren't identical. Maybe not updating tf-tensorboard.html was a better idea. Should it be updated in sync with the .ts, or should I leave it unchanged?

jyegerlehner:
And just to make things more concrete, here are images showing the different behaviors on the same data.

The previous behavior with the "noisy" end section was like this:
p_old

The behavior currently in master (or what you would get in master if tf-tensorboard.html were updated) is this:
p_prev

It does remove the noisy bit at the end, but leaves one unable to see the most recent trend.

And the behavior with the change in this PR preserves the trend all the way to the end:
p_thispr

tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
```diff
@@ -440,26 +440,18 @@ module VZ {
...
-          // In the beginning, it's OK for the smoothing window to be small,
-          // but this is not desirable towards the end. Rather than shrinking
-          // the window, or extrapolating data to fill the gap, we're simply
-          // not going to display the smoothed line towards the end.
-          d.smoothed = Infinity;
-        } else if (!_.isFinite(d.scalar)) {
-          // Only smooth finite numbers.
+      data.forEach(function (d, i) {
+        var actualKernelRadius = Math.min(kernelRadius, i)
+        var start = Math.max(0, i - actualKernelRadius);
```
dandelionmane: Isn't the Math.max(0, i - actualKernelRadius) unnecessary?
Because when (i <= kernelRadius), actualKernelRadius = min(kernelRadius, i) = i
so i - actaulKernelRadius == 0 in that case.

conversely, if (i > kernelRadius), then (i - actualKernelRadius > 0)

so the Max(0) never gets invoked

tensorflow/tensorboard/dist/tf-tensorboard.html
```diff
@@ -3856,9 +3856,9 @@ <h3 id="tooltip-help" class="tooltip-container">
             var data = dataset.data();
             var kernelRadius = Math.floor(data.length * factor / 2);
             data.forEach(function (d, i) {
-                var actualKernelRadius = Math.min(kernelRadius, i, data.length - i - 1);
```
dandelionmane: There's no need to update tf-tensorboard by hand, it will get automatically regenerated in time for the next release.
We're working on removing this layer of indirection by directly compiling TensorBoard in bazel... it should come Soon(tm)

teamdandelion (changes requested):
Thanks for this change!
We've had a lot of discussion about how to do smoothing best internally, I'm going to summarize some considerations here:

Previously (the code in the current release of TB) the window was always time-netural (i.e. included past and future data equally) but the window shrinks near the edges, resulting in noisy data at the end. This was undesirable.

With jart@'s changes (the version currently in Typescript) the smoothing just ends as soon as we don't have enough data to fill the window, which is really consistent (after the window has filled up, the smoothing algorithm never changes) but throws away some data.

With your changes, the smoothing ceases being time-neutral near the end of the sequence. Instead, it becomes backwards-biased because the right side of the kernel closes off but the left side stays the same size.

Personally, I like your solution the most: I think having a backwards-biased smoothing at the end of the sequence seems reasonable, and it's nice to show smooth data through the end of the plot. We might want to document this somewhere (e.g. in the README for the vz-line-chart or vz-scalar-dashboard?)

I'm going to let other people (e.g. jart@) chime in too.

tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
jyegerlehner: Good point. I removed max().

tensorflow/tensorboard/dist/tf-tensorboard.html
jyegerlehner: OK. tf-tensorboard.html change rolled back.

martinwicke:
@dandelionmane Can you check in on the changes? Thanks!

dandelionmane:
Let's take #8363 instead.

omasoud:
The new smoothing is nice. But we lost the ability of being able to peek at values in the non-smoothed region. It would be good to let the tooltip show up when mousing in that region (and perhaps show blank under the "smoothed" column).

PR #7893, opened 2017-02-26
Add MNIST dataset loader in Go binding

Lewuathe:
As well as python examples, MNIST dataset is good resource for tutorial and examples in Go API. 

In addition to this, it might be better to write examples using MNIST like logistic regression or CNN but Go binding API (wrappers.go) looks so different from python API. Could you let me know if there is good resource to write MNIST example in Go?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ node_modules
 __pycache__
 *.swp
 .vscode/
+
+tensorflow/go/data/*.gz
--- a/tensorflow/go/data/mnist.go
+++ b/tensorflow/go/data/mnist.go
@@ -1,0 +1,230 @@
+// Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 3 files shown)


asimshankar:
@Lewuathe : This a wonderful, though I'd suggest that you keep this example in your repository rather than putting it in the TensorFlow repository for now. 

Regarding construction of the CNN: `wrappers.go` provides only the low level TensorFlow ops, while Python has richer APIs in `tf.layers` (other other frameworks like Keras) for constructing models. To reproduce that in Go, for now, the choices would be to, either:

- Construct the model in Python and then export/import into Go (which is what I did in this [word2vec example](https://github.com/asimshankar/go-tensorflow/tree/master/word2vec), OR
- Follow the codepath for model construction in python to figure out the primitive TensorFlow ops being use and reconstruct with that (perhaps adding a "layers" like API for Go).

Hope that helps.

Lewuathe:
@asimshankar Thanks for letting me know. 
Just one question. So the main purpose of Go binding is not training but embedding each respective applications? For example, we train model with Python API and export. Go binding is used in Go application to read and run the model. Is this assumption correct?

asimshankar:
Thanks for understanding, I'll close this pull request for now.

To answer your question: As of today, the Go API is well suited for importing and executing a trained model. It is also suited, but a little less so, for importing models constructed in Python and training them from within Go (e.g. the word2vec example I alluded to above) as it lacks the more convenient higher level APIs (like [Estimator](https://www.youtube.com/watch?v=t64ortpgS-E&index=6&list=PLOU2XLYxmsIKGc_NBoIhTn2Qhraji53cv)) that are implemented in Python. A little work is needed to duplicate that convenience. For model construction, it is possible in Go since it implements the lowest level of graph construction, but this is where you're likely to miss the higher level constructs like [`tf.layers`](https://www.tensorflow.org/api_docs/python/tf/layers) and a little more work is needed to develop those.

If there are folks in the community who want to develop their own APIs on top of the primitives of this `tensorflow` package in Go, we encourage that.

PR #7937, opened 2017-02-28
Adding Intel Conv2D kernel implementation alongwith required Graph pa…

vivek-rane:
…sses

This commit contains 4 main components:

1) Intel-optimized kernel implementation for Conv2D op
   Implementation in kernels/mkl_conv_ops.*

2) Graph passes required to enable Conv2D optimized implementation
   Implementation in graph/mkl_*. We also need a new op, MklToTf op.
   Its implementation is in kernels/mkl_tfconv_op.cc.

3) Utility functions used in kernel implementation
   Implementation is in common_runtime/mkl_layer_registry* and util/mkl_util.h

4) BUILD changes for Conv2D, graph passes and utility functions


nhasabni:
Hi,

This contribution is under Intel CLA Submissions. I've setup my email on github account, not sure what is wrong here.

nhasabni:
Hi,

I've signed new CLA.

tensorflow/core/common_runtime/mkl_layer_registry.h
```diff
@@ -0,0 +1,107 @@
...
+  }
+};
+
+///////////////////////////////////////////////////////////
+//           Public interface for registration
+///////////////////////////////////////////////////////////
+
+// We do not need any types other than half, float, and double for MKL for now.
+#define REGISTER_MKL_LAYER_half(opname)                                 \
+  REGISTER_MKL_LAYER_UNIQ_HELPER(__COUNTER__, std::string(opname), DT_DOUBLE)
```
andydavis1: Did you intend to put DT_HALF here?

tensorflow/core/common_runtime/mkl_layer_registry.h
```diff
@@ -0,0 +1,107 @@
...
+#include <string>
+
+#include "tensorflow/core/framework/types.pb.h"
+
+namespace tensorflow {
+
+// A global MKLLayerRegistry is used to hold information about all MKL-compliant
+// layers. Registry maintains information about <opname,T> pair supported
+// by MKL API.
+class MklLayerRegistry {
```
andydavis1: I'm not understanding the need for another static registry here. Could you explain why the current TF op or kernel registries were not sufficient?

tensorflow/core/kernels/mkl_conv_ops.cc
```diff
@@ -0,0 +1,420 @@
...
+    const int64 stride_n = GetTensorDim(strides_, data_format_, 'N');
+    const int64 stride_c = GetTensorDim(strides_, data_format_, 'C');
+    OP_REQUIRES(
+        context, stride_n == 1 && stride_c == 1,
+        errors::InvalidArgument("Current implementation does not yet support "
+                                "strides in the batch and depth dimensions."));
+    OP_REQUIRES_OK(context, context->GetAttr("padding", &padding_));
+  }
+
+  void Compute(OpKernelContext* context) override {
```
andydavis1: This function is kind of long. Any chance of breaking it up into 3 or 4 member function calls that focus on particular things: format conversion sub-routine, resource clean-up sub-routine, etc...

tensorflow/core/common_runtime/mkl_layer_registry.h
nhasabni: Yes, my bad.

tensorflow/core/common_runtime/mkl_layer_registry.h
nhasabni: We need a mechanism to track which layers are supported by MKL and the input datatypes that are supported by the MKL layer. An example of this registration is in mkl_conv_ops.cc: 416, where call is: REGISTER_MKL_LAYER_float("MklConv2D").

tensorflow/core/kernels/mkl_conv_ops.cc
jbobba: We could. Are you OK with us promoting the MKL variables to class variables? That will help us keep the argument lists short.

andydavis1: Class variables for the MklConv2DOp class? That seems fine.

jbobba: Yes. We'll break it up then.

tensorflow/core/common_runtime/mkl_layer_registry.h
andydavis1: Right, but could just just use the OpKernel registry that we already have? It can be queried in this way:

https://github.com/tensorflow/tensorflow/blob/99fe61a8a8f3dd41b4e1e4dedfc53b45f67e88a7/tensorflow/core/framework/op_kernel.h#L1062

tensorflow/core/kernels/mkl_conv_ops.cc
```diff
@@ -70,11 +70,161 @@ class MklConv2DOp : public OpKernel {
...
+    size_t filter_strides[4];
+    size_t bias_sizes[1];
+    size_t bias_strides[1];
+    int input_offset[2];
+    size_t conv_stride[2];
+    MklShape input_shape;
+  } MklConv2DOpParams;
+
+  // Create MKL dnnLayout_t objects for tensors coming into the layer
+  void MklCreateInputLayouts(OpKernelContext* context) {
```
andydavis1: These functions look good. Thanks for doing this. Can you move them into the "private" section of this class defination at line 429?

jbobba: Done.

andydavis1: ok. that looks good.

tensorflow/core/common_runtime/mkl_layer_registry.h
andydavis1: Curious about the response here (i.e. could we just leverage our OpKernel registry here)?  Outside of this question, I think most other PR issues are resolved...

nhasabni: Yes, I think it can be used. I will upload changes soon.

andydavis1 (approved):

andydavis1:

dandelionmane:
Failing buildifier. Please run buildifier and then we'll rerun tests.
https://github.com/bazelbuild/buildifier

jbobba:
Fixed the buildifier error but not sure what we need to do for author consent.

andydavis1:

andydavis1:
I asked around and the "basic_session_run_hooks_test" failure is occuring on other builds so we are ok to merge here.

tensorflow/core/kernels/mkl_conv_ops.cc
jbobba: @andydavis1 Promoting MKL state to class members seems to conflict with this advice. https://www.tensorflow.org/extend/adding_an_op "Important: Instances of your OpKernel may be accessed concurrently. Your Compute method must be thread-safe. Guard any access to class members with a mutex. Or better yet, don't share state via class members! Consider using a ResourceMgr to keep track of op state." We are thinking of moving this back into the Compute function and explicitly pass state between MKL sub-routines. thoughts?

zhangyaobit: Which variables are you considering making to class members? If they are static (don't change across training/inference runs), I think it is ok to make them class members, just like what you already have for strides_, padding_, data_format_. My understanding is that each node will have its own instance of OpKernel, but this instance is shared across different runs.

andydavis1: If the variables can change per-invocation of the Compute method, then you'll need to pass them into the sub-routines (instead of keep them as class variables). Sorry about that, I assumed the state you were adding as class variables was thread safe.

jbobba: Thanks for the feedback. The MKL variables are not all thread-safe. So we will have to either pass them through sub-routines or revert to a single monolithic function. Do you have a preference?

zhangyaobit: I think Andy suggested to "then you'll need to pass them into the sub-routines". Please let us know if this is too cumbersome to do or if you prefer a monolithic function.

jbobba: Its more of a style issue. Monolithic makes the function too long while sub-routines have some ugliness in that they dont always localize Mkl state. Effort is minimal in either case. We will go with your preference.

PR #7947, opened 2017-02-28
Allows C++ API to be built on Windows within Visual Studio and Debug configurations.

markslwong:
Problem: TensorFlow does not build directly from Visual Studio.

**Cause 1**: The CMake scripts were written to assume building from CMake directly due to it's usage of the CMAKE_BUILD_TYPE variable.  If you generate a Visual Studio solution and build from directly within that solution, the build will fail because Visual Studio has multiple-configurations, rendering this variable blank when generating a solution.
**Solution 1**: Replace CMAKE_BUILD_TYPE with $(Configuration) which is a Visual Studio macro.  During build time, Visual Studio will replace $(Configuration) with the currently active configuration being built.

**Cause 2**: The CMake build scripts were written to assume 'Release' builds would only be built.  When building in Debug, some external libraries generate 'lib' files with a 'd' fixed to the end of the name.  However, the reference to these 'lib' files do not account for the 'd' at the end of the name.  During debug builds, linker errors occur when the file is not found.
**Solution 2**: use the CMake command 'debug' and 'optimized' to specify references to difference file names upon 'Debug' and 'Release' builds respectively.

**Cause 3**: The CMake build scripts were written to assume 'Release' builds would only be built.  When building in Debug, linker errors occur because _ITERATOR_DEBUG_LEVEL is defined and set to zero for all configurations.  External libraries (like Eigen) build with it's own CMake scripts do not define this, causing a different code path to be built.
**Solution 3**: Only define _ITERATOR_DEBUG_LEVEL in Debug builds.

**Cause 4**: Visual Studio compile errors occur.  Functions must always have a return value.
**Solution 4**: For function stubs without proper implementation, we return NULL.

**Cause 5**: Visual Studio compile errors occur.  We define a "double" code path for Eigen that is not available for CPU builds using SSE only.
**Solution 5**: We remove the "double" code path if AVX and CUDA are not available on Windows builds.  This error only shows up on Debug builds.  I'm assuming because the compiler doesn't strip out unused symbols and tries to link against it.

### DIFF
--- a/tensorflow/examples/tutorials/estimators/abalone.py
+++ b/tensorflow/examples/tutorials/estimators/abalone.py
@@ -134,12 +134,22 @@ def main(unused_argv):
 
   # Instantiate Estimator
   nn = tf.contrib.learn.Estimator(model_fn=model_fn, params=model_params)
-
+  
+  def get_train_inputs():
+    x = tf.constant(training_set.data)
+    y = tf.constant(training_set.target)
+    return x, y
+  
   # Fit
-  nn.fit(x=training_set.data, y=training_set.target, steps=5000)
+  nn.fit(input_fn=get_train_inputs, steps=5000)
 
   # Score accuracy
-  ev = nn.evaluate(x=test_set.data, y=test_set.target, steps=1)
+  def get_test_inputs():
+    x = tf.constant(test_set.data)
+    y = tf.constant(test_set.target)
+    return x, y
+  
+  ev = nn.evaluate(input_fn=get_test_inputs, steps=1)
   print("Loss: %s" % ev["loss"])
   print("Root Mean Squared Error: %s" % ev["rmse"])
 


caisq:
@markslwong, can you please sign the CLA before the review starts. Thanks.

markslwong:
I signed it!

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -58,14 +58,14 @@ add_definitions(-DEIGEN_AVOID_STL_ARRAY)
...
-  add_definitions(-DTENSORFLOW_USE_EIGEN_THREADPOOL -DEIGEN_HAS_C99_MATH -D_ITERATOR_DEBUG_LEVEL=0)
-  add_definitions(-DNDEBUG /O2)  # Equivalent of -c opt in Bazel.
+  add_definitions(-DTENSORFLOW_USE_EIGEN_THREADPOOL -DEIGEN_HAS_C99_MATH)
   add_definitions(/bigobj /nologo /EHsc /GF /FC /MP /Gm-)
   # Suppress warnings to reduce build log size.
   add_definitions(/wd4267 /wd4244 /wd4800 /wd4503 /wd4554 /wd4996 /wd4348 /wd4018)
   add_definitions(/wd4099 /wd4146 /wd4267 /wd4305 /wd4307)
   add_definitions(/wd4715 /wd4722 /wd4723 /wd4838 /wd4309 /wd4334)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /D_ITERATOR_DEBUG_LEVEL=0 /DNDEBUG /O2")
```
mrry: I suppose `/O2` is redundant in Release builds. Should we remove it?

tensorflow/core/framework/register_types.h
```diff
@@ -58,7 +58,11 @@ limitations under the License.
 // Note: macros are defined in same order as types in types.proto, for
 // readability.
 #define TF_CALL_float(m) m(float)
+#if defined(_WINDOWS) && !defined(GOOGLE_CUDA) && !defined(__AVX__) && !defined(__AVX2__) // Not supported on Eigen for SSE
```
mrry: I'm a little concerned about this line. How many things failed without it? If the number of kernels to disable is not overwhelming, we might be able to do this at the CMake or kernel registration level... disabling an entire type seems drastic!

Would the predicate here be equivalent to `#if defined(PLATFORM_WINDOWS) && defined(DEBUG)`? If so, the intent might be clearer.

mrry (commented):
Thanks for this contribution! I have a couple of questions about the configuration changes.

tensorflow/contrib/cmake/CMakeLists.txt
markslwong: I think it's a good idea.  Will remove.

tensorflow/core/framework/register_types.h
markslwong: Hmm... I guess you're right.  It is a little extreme.

I'll look for a more precise way of doing this.

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -58,14 +58,14 @@ add_definitions(-DEIGEN_AVOID_STL_ARRAY)
...
-  add_definitions(-DTENSORFLOW_USE_EIGEN_THREADPOOL -DEIGEN_HAS_C99_MATH -D_ITERATOR_DEBUG_LEVEL=0)
-  add_definitions(-DNDEBUG /O2)  # Equivalent of -c opt in Bazel.
+  add_definitions(-DTENSORFLOW_USE_EIGEN_THREADPOOL -DEIGEN_HAS_C99_MATH)
   add_definitions(/bigobj /nologo /EHsc /GF /FC /MP /Gm-)
   # Suppress warnings to reduce build log size.
   add_definitions(/wd4267 /wd4244 /wd4800 /wd4503 /wd4554 /wd4996 /wd4348 /wd4018)
   add_definitions(/wd4099 /wd4146 /wd4267 /wd4305 /wd4307)
   add_definitions(/wd4715 /wd4722 /wd4723 /wd4838 /wd4309 /wd4334)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /D_ITERATOR_DEBUG_LEVEL=0 /DNDEBUG")
```
snnn: Does it also work for RelWithDbgInfo Build?

snnn: Is "/DNDEBUG" necessary?

martinwicke:
Ping! What's the status of this? If it is abandoned, I will close it soon.

markslwong:
I've been chipping away at it.  I've had to setup a new environment at home to be able to work at it.  I'm leaving my current job next week... so haven't had time at work.  Sorry it's been a bit slow.  :)  But I have worked been putting in some time this weekend and yesterday night.

martinwicke:
Thanks @markslwong, comment when you're done so @mrry knows to take another look.

markslwong:
I've disabled the problematic kernel directly as suggested by @mrry 

I've placed the D_ITERATOR_DEBUG_LEVEL in RelWithDebug and MinSizeRel to address @snnn   However, RelWithDebug never worked for me as I discuss later in a bigger issue below.

I don't know why GitHub diff shows the whole file as different.  My first assumption was it was different line endings... but after a binary diff it showed that it had not changed.

A new change that I want to get in is to not have to play with the Debug flags for a Debug build.  Visual Studio runs out of memory when building (even when using the 64-bit version of the compiler on a machine with 8gb physical memory and 126gb page file available).  This can be resolved by removing the /Zi /Od and /RTC1 flags... which effectively build a release build with _DEBUG defined.  Since our company statically links to Tensorflow... we cannot create a Debug build because the _DEBUG flag is defined for our builds... and when we build the headers, it runs a different code path which runs into linker errors.  Doing this allows us to get a lib files that we make a Debug build of our application and Debug our own code.  However, we cannot Debug the Tensorflow code as a consequence.  So I'm not sure whether you want this change or not.

Then there is the bigger question of RELWITHDEBINFO.  I did not make the same changes to this build.  I just let it fail naturally because it should always fail on Windows.  I could apply the same changes as I did with Debug.  However, I feel it's entirely contradictory to it's name... so I think failure is an acceptable outcome in this circumstance.

Let me know your thoughts.

snnn:
Hi @markslwong could you tell me why it fails? At least it works for me. Have you tried to compile a RELWITHDEBINFO build from master HEAD ?

markslwong:
I have not.  I going to assume the difference might be command line build single configuration build vs Visual Studio IDE build.

markslwong:
I'll test my repo for RelWithDebugInfo and report the outcome.  I'm assuming it would work.

guschmue:
RelWithDebInfo works for sure - I have been using those ever since windows is supported.

markslwong:
RelWithDebug works for me on both Tensorflow HEAD and my fork with the command line.  It is the good idea to keep things as is.  However this brings the bigger question of what to do with the Debug build.  I can remove those flags.  However, the consequence of doing so is that it can be only be built via MSBuild.  I've setup my previous company to build directly from Visual Studio so I have a personal preference to keep it in there so that they can continue with the existing workflow.  I also think building a solution within Visual Studio seems like a fairly reasonable thing to do.  As most people are likely users of the Tensorflow libraries and not necessarily going to need to be debugging it... I would guess that this would be the most common path.  However, I can only take a guess from my own bias lens.  I don't have any true statistics as to the most typical way to build.

drpngx:
@markslwong could you pull rebase and push again?

markslwong:
@drpngx will do.

guschmue:
Maybe it would simplify to split this PR into 2:
There is the issue that debug builds don't compile - should be a nice, clean PR to fix this.

There is the other part in this PR to replace CMAKE_BUILD_TYPE with $(Configuration) which worries me a little because  CMAKE_BUILD_TYPE is defined by cmake where $(Configuration) is only known by Visual Studio and should not be in cmake ... the cmake generator for Visual Studio is supposed to deal with this. I  normally don't switch the configuration in VS but just tried and it seems to be doing the right thing (cmake 3.8). CMAKE_BUILD_TYPE is used consistently in the cmake files, including the external projects (like grpc), CI, ... so if it is changed the review and test are much more involved.

markslwong:
@guschmue I have been running CMake 3.7.2, so I was curious to see if something had changed to make it work.  I upgraded to CMake 3.8.0 (official release) and tried to compile with Visual Studio 2017 Community Edition on the tensorflow official repository at commit 97c6203.  92 projects have failed to build, with the same error that I had seen and fixed with this pull request.

I think there might be a misunderstanding as to where the errors are introduced.  If you use CMake on the command line and generate a single configuration solution, regardless what version of CMake (3.7.2 or 3.8.0), Tensorflow will successfully build.  The problem is introduced when you generate a multi-configuration solution.  This is done by default if you use the CMake GUI.

I understand your concerns about not relying on CMAKE_BUILD_TYPE.  I was equally as concerned as I was trying to find a solution.  I thoroughly went through the CMake documentation and could not find an equivalent option for multi-configuration solutions.  

You are correct external projects like grpc do use the CMAKE_BUILD_TYPE flag.  But that is okay because Visual Studio will call CMake during the build for the active configuration.  I have not needed to modify the CMake scripts of these projects.  Most of what I have changed is when the build references a build output.  (eg: a pre-build step that calls proto.exe will actually look like this: protobuf\src\protobuf\\protoc.exe in the generated solution for multiple configurations.  This is wrong since the \\ is missing the build configuration in between.)

The changes that I have made are in cmake scripts are nested within "Windows" only build configurations.  I have tested this change both against the single-configuration solutions with the command line and multi-configuration with the CMake GUI and building within Visual Studio.

Given this new information... are you still concerned about the change?  Would you still like me to separate this into two CL's?

guschmue:
I guess if $(Configuration) is under if(WIN32) it should not break something and I should not be worried about it.
Splitting the pr:  smaller pr addressing one topic is easier but I with this one already there ...

martinwicke:
@markslwong is this ready for @mrry to take another look?

vrv:
friendly ping, is there progress being made here behind the scenes that we should wait for?

markslwong:
Not yet.  Sorry... after the new job I moved and purchased a home.

Things are starting to get back to normal.  I'll rebase this before Monday.

adennie:
I was following this PR with interest... @markslwong are you planning to open a new one (or multiple)?

markslwong:
Hi @adennie.  GitHub was having trouble with the diff... so I reapplied the changes on head revision and force pushed.  I thought that GitHub would treat it as the same pull request but it did not.  A new pull request has been made here https://github.com/tensorflow/tensorflow/pull/9666

markslwong:
@mrry The changes have been reapplied here https://github.com/tensorflow/tensorflow/pull/9666

PR #7948, opened 2017-02-28
Second derivative computation in softmax xent op

persiyanov:
PR for issue #7403

@girving I rewrited second derivative with reshapes and one `matmul`, no more `diag_part` or `reduce_sum`.

* Also, I writed test for checking that no extra computation performed when only first gradient is requested.

* I am not sure whether I implemented hessian correctness test properly, but if I did it poorly, please guide me into right way.

* Is such fix suites trivially for `_SparseSoftmaxCrossEntropyWithLogitsGrad`? If so, I will modify it too.

* I think that I can fix `SoftmaxGrad` op, because it uses `reduce_sum(grad * softmax)` approach to compute the derivative. I could do it with `matmul` and reshapes.

### DIFF
--- a/tensorflow/python/kernel_tests/xent_op_test.py
+++ b/tensorflow/python/kernel_tests/xent_op_test.py
@@ -195,17 +195,17 @@ class XentTest(test.TestCase):
                                                    name="xent")
       loss = math_ops.reduce_sum(x)
 
-      trivia_loss = -math_ops.reduce_sum(l * math_ops.log(nn_ops.softmax(f)))
+      hessians = gradients_impl.hessians(loss, [f])[0]
 
-      hessians = sess.run(gradients_impl.hessians(loss, [f]))
-      trivia_hessians = sess.run(gradients_impl.hessians(trivia_loss, [f]))
-
-      self.assertAllClose(hessians, trivia_hessians)
+      err = gradient_checker.compute_gradient_error(f, [12], hessians, [12, 12])
 
       # Check that second derivative is calculated.
       # (it is equivalent to being `BatchMatMul` op in the graph because of implementation of xentropy grad)
       op_names = [op.op_def.name for op in sess.graph.get_operations() if op.op_def]
       self.assertIn('BatchMatMul', op_names)
+
+    print("cross entropy hessian err = ", err)
+    self.assertLess(err, 5e-8)
 
   def testWrapper(self):
     features = np.array(
--- a/tensorflow/python/kernel_tests/xent_op_test.py
+++ b/tensorflow/python/kernel_tests/xent_op_test.py
@@ -195,9 +195,9 @@ class XentTest(test.TestCase):
                                                    name="xent")
       loss = math_ops.reduce_sum(x)
 
-      hessians = gradients_impl.hessians(loss, [f])[0]
+      gradients = gradients_impl.gradients(loss, [f])[0]
 
-      err = gradient_checker.compute_gradient_error(f, [12], hessians, [12, 12])
+      err = gradient_checker.compute_gradient_error(f, [12], gradients, [12])
 
       # Check that second derivative is calculated.
       # (it is equivalent to being `BatchMatMul` op in the graph because of implementation of xentropy grad)
--- a/tensorflow/python/kernel_tests/xent_op_test.py
+++ b/tensorflow/python/kernel_tests/xent_op_test.py
@@ -157,7 +157,7 @@ class XentTest(test.TestCase):
         np.array([[0., 0., 0., 1.], [0., .5, .5, 0.]]).astype(np.float64))
 
   def testGradient(self):
-    with self.test_session():
+    with self.test_session() as sess:
       l = constant_op.constant(
           [0.0, 0.0, 1.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.5, 0.0, 0.5],
           shape=[3, 4],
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/nn_grad.py
```diff
@@ -321,21 +321,24 @@ def _BroadcastMul(vec, mat):
 
 
 @ops.RegisterGradient("SoftmaxCrossEntropyWithLogits")
-def _SoftmaxCrossEntropyWithLogitsGrad(op, grad_0, _):
+def _SoftmaxCrossEntropyWithLogitsGrad(op, grad_0, grad_1):
```
girving: Change the arguments from `grad_0`, `grad_1` to `grad_loss`, `grad_grad`.

tensorflow/python/ops/nn_grad.py
```diff
@@ -321,21 +321,24 @@ def _BroadcastMul(vec, mat):
...
-  softmax_grad_without_gradient = array_ops.prevent_gradient(
-      op.outputs[1], message="Currently there is no way to take the second "
-      "derivative of softmax_cross_entropy_with_logits due to the fused "
-      " implementation's interaction with tf.gradients()")
-  return _BroadcastMul(grad_0, softmax_grad_without_gradient), None
+  # Second derivative is just softmax derivative w.r.t. logits.
+  softmax_grad = op.outputs[1]
+  grad = _BroadcastMul(grad_0, softmax_grad)
+
+  if grad_1.op.type not in ('ZerosLike', 'Zeros'):  # or `in ('Zeros', 'ZerosLike')`
```
girving: Remove the obsolete comment.

tensorflow/python/ops/nn_grad.py
```diff
@@ -321,21 +321,24 @@ def _BroadcastMul(vec, mat):
...
+      logits = op.inputs[0]
+      softmax = nn_ops.softmax(logits)
+
+      grad += ((grad_1 - math_ops.matmul(grad_1[:, None, :], softmax[:, :, None])[:, 0]) * softmax)
```
girving: Instead of `[:, 0]`, use `tf.squeeze(..., axis=1)` to make it clear no work is getting wasted.

tensorflow/python/kernel_tests/xent_op_test.py
```diff
@@ -171,28 +171,41 @@ def testGradient(self):
...
                                                    name="xent")
-      loss = math_ops.reduce_mean(x)
+      loss = math_ops.reduce_sum(x)
+
+      trivia_loss = -math_ops.reduce_sum(l * math_ops.log(nn_ops.softmax(f)))
+
+      hessians = sess.run(gradients_impl.hessians(loss, [f]))
+      trivia_hessians = sess.run(gradients_impl.hessians(trivia_loss, [f]))
+
+      self.assertAllClose(hessians, trivia_hessians)
```
girving: Use `gradient_checker.compute_gradient_error` instead of writing out the gradient manually a second time.

girving (changes requested):
Thank you for the contribution!  Looks good with a few minor comments.

persiyanov:
@girving fixed

tensorflow/python/kernel_tests/xent_op_test.py
```diff
@@ -195,18 +195,18 @@ def testSecondGradient(self):
...
       loss = math_ops.reduce_sum(x)
 
-      trivia_loss = -math_ops.reduce_sum(l * math_ops.log(nn_ops.softmax(f)))
+      hessians = gradients_impl.hessians(loss, [f])[0]
 
-      hessians = sess.run(gradients_impl.hessians(loss, [f]))
-      trivia_hessians = sess.run(gradients_impl.hessians(trivia_loss, [f]))
-
-      self.assertAllClose(hessians, trivia_hessians)
+      err = gradient_checker.compute_gradient_error(f, [12], hessians, [12, 12])
```
girving: Won't this fail, because you're asking Hessians to be the same as gradients?  In order to use `compute_gradient_error`, you have to call `gradients` yourself twice.

Also, please run this unit test to make sure it passes before asking for review.

girving (changes requested):

tensorflow/python/kernel_tests/xent_op_test.py
persiyanov: Its strange, but it did not fail. But I got it, the correct way is something like that:

```python
gradients = gradients_impl.gradients(loss, [f])[0]

err = gradient_checker.compute_gradient_error(f, [12], gradients, [12])
```

Yes? This does not fail.

girving: Ah, right: it wouldn't have failed but it also wouldn't have checked the right thing.  It would have checked that the gradient of the Hessian was consistent.   Your correct way is indeed correct.

persiyanov: Yeah, I have realized it too=)

girving (approved):

girving:
Jenkins, test this please.

girving:
Looks like we're having infrastructure issues, so there will be a delay before the tests can be run.

girving:
Jenkins, test this please.

girving:
@gunan Are Jenkins builds still broken?  I fired off the last build after you said they were good, but it still failed with a timeout.

gunan:
They are fixed internally, but externally we do not have the workaround.
We can separately make the same change to fix the external builds
@yifeif could you send a PR?

yifeif:
Fix at #7953

girving:
Jenkins, test this please.

gunan:
Jenkins, test this please.

persiyanov:
@girving tests looks okay. What about `SparseSoftmaxCrossEntropyGrad`? Also, maybe `SoftmaxGrad` needs to be rewritten? For now, the implementation of it is this:
```python
softmax = op.outputs[0]
  grad_x = ((grad_softmax - array_ops.reshape(
      math_ops.reduce_sum(grad_softmax * softmax, [1]), [-1, 1])) * softmax)
  return grad_x
```

girving:
@persiyanov Thank you for the contribution!  Looks like `SparseSoftmaxCrossEntropyGrad` has the same issue as you say.  For `SoftmaxGrad`, and you referring to the fact that it doesn't use `MatMul`?

persiyanov:
@girving Yes, I say that I also could fix the sparse xentropy grad (in the same way) and rewrite softmax grad with `matmul`, not reshape followed by `reduce_sum`. What do you think?

girving:
That would be good, but it's worth checking that I'm actually right that it's a performance win.  It's always possible that I've mislead you. :)

persiyanov:
@girving Okay, I will measure the performance before and after and if its okay I will make a PR.

PR #7953, opened 2017-02-28
Replace broken junit download link with maven.

yifeif:


tensorflow/workspace.bzl
```diff
@@ -418,15 +418,15 @@ def tf_workspace(path_prefix = "", tf_repo_name = ""):
...
 
   # Make junit-4.12 available as //external:junit
-  native.http_jar(
-      name = "junit_jar",
-      url = "https://github.com/junit-team/junit4/releases/download/r4.12/junit-4.12.jar",
-      sha256 = "59721f0805e223d84b90677887d9ff567dc534d7c502ca903c0c2b17f05c116a",
+  native.maven_jar(
+      name = "junit_junit",
+      artifact = "junit:junit:4.12",
+      sha1 = "2973d150c0dc1fefe998f834810d68f278ea58ec",
```
girving: Should we have a TODO to remove sha1 dependencies? :)

yifeif: I followed the example on bazel doc https://bazel.build/versions/master/docs/be/workspace.html#maven_jar
Do we not need the sha1 here?

girving:
Don't worry about it for this patch, but in the long run we should switch
over to hashes which aren't broken (sha256, for instance).

yifeif:
ah totally behind on news :). Good to know. Don't think this change should affect input_test in cpu build. @tensorflow-jenkins test this please.

yifeif:
cpu build passed, merging.

Issue #7719, opened 2017-02-21
Feature request: convenience functions for getting variables from scopes


cooijmanstim:
There does not currently seem to be a good way to get a collection of variables that belong to a given scope. I've seen issue #7295, but the answer given by Yaroslav (at http://stackoverflow.com/questions/42073239/tf-get-collection-to-extract-variables-of-one-scope) does not nest:

```python
import tensorflow as tf

def bar():
  with tf.variable_scope("bar") as scope:
    x = tf.Variable(0, name="x")
    print list(map(str, tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES, scope=scope.name)))
    print list(map(str, tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES, scope=scope.name + "/")))
  print list(map(str, tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES, scope="bar")))

def foo():
  with tf.variable_scope("foo") as scope:
    with tf.variable_scope("bar_") as scope:
      y = tf.Variable(0, name="y")
    bar()

with tf.Graph().as_default():
  bar()
print
with tf.Graph().as_default():
  foo()
```

The snippet above outputs (on TF 1.0):
```
['Tensor("bar/x/read:0", shape=(), dtype=int32)']
['Tensor("bar/x/read:0", shape=(), dtype=int32)']
['Tensor("bar/x/read:0", shape=(), dtype=int32)']

['Tensor("foo/bar_/y/read:0", shape=(), dtype=int32)', 'Tensor("foo/bar/x/read:0", shape=(), dtype=int32)']
['Tensor("foo/bar/x/read:0", shape=(), dtype=int32)']
[]
```

The output shows that wrapping the scope "bar" in another scope ("foo") breaks Yaroslav's proposal of using `tf.get_collection(..., "bar")`. It also breaks `tf.get_collection(..., scope.name)` by including variables from outside the scope. The only version that works regardless of context is `tf.get_collection(..., scope.name + "/")`, which is ugly and too informed.

It would be great to have `scope.trainable_variables()`, or `scope.get_collection(...)`, or even `tf.get_collection(..., scope)`.

suiyuan2009:
@cooijmanstim the function `tf.get_collection(key, scope)` does a prefix match in all variables belonging to the `key` collection. The Variable `foo/bar_/y/read:0` and the scope `foo/bar` have same prefix.

cooijmanstim:
[...]

PR #7977, opened 2017-03-01
add function get_collection for scope

suiyuan2009:
#7719

add get_collection(key) to class VariableScope.

example:
```
with tf.variable_scope("foo") as scope:
  print scope.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES)
```

### DIFF
--- a/tensorflow/python/kernel_tests/variable_scope_test.py
+++ b/tensorflow/python/kernel_tests/variable_scope_test.py
@@ -725,18 +725,26 @@ class VariableScopeTest(test.TestCase):
     with self.test_session():
       a = variable_scope.get_variable("a", [])
       b = variable_scope.get_variable("b", [], trainable=False)
-      with variable_scope.variable_scope("foo_") as scope:
-        a = variable_scope.get_variable("a", [])
-        b = variable_scope.get_variable("b", [], trainable=False)
-      with variable_scope.variable_scope("foo") as scope:
+      with variable_scope.variable_scope("foo_") as scope1:
         a = variable_scope.get_variable("a", [])
         b = variable_scope.get_variable("b", [], trainable=False)
         self.assertEqual(
-            [v.name for v in scope.get_collection(
+            [v.name for v in scope1.get_collection(
+              ops.GraphKeys.TRAINABLE_VARIABLES)],
+            ["foo_/a:0"])
+        self.assertEqual(
+            [v.name for v in scope1.get_collection(
+              ops.GraphKeys.GLOBAL_VARIABLES)],
+            ["foo_/a:0", "foo_/b:0"])
+      with variable_scope.variable_scope("foo") as scope2:
+        a = variable_scope.get_variable("a", [])
+        b = variable_scope.get_variable("b", [], trainable=False)
+        self.assertEqual(
+            [v.name for v in scope2.get_collection(
               ops.GraphKeys.TRAINABLE_VARIABLES)],
             ["foo/a:0"])
         self.assertEqual(
-            [v.name for v in scope.get_collection(
+            [v.name for v in scope2.get_collection(
               ops.GraphKeys.GLOBAL_VARIABLES)],
             ["foo/a:0", "foo/b:0"])
       scope = variable_scope.get_variable_scope()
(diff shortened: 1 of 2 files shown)


caisq:

tensorflow/python/ops/variable_scope.py
```diff
@@ -882,6 +882,11 @@ def set_custom_getter(self, custom_getter):
     """Set custom getter for this scope."""
     self._custom_getter = custom_getter
 
+  def get_collection(self, name):
```
caisq: Since this adds a new method to the VariableScope class, it will require API change review. 

Also, assuming that this change is all right, it will require some unit tests.

lukaszkaiser:
Thanks for the PR, this is a method we should have. But yes, it will need a test. Also, instead of the generic "get_collection" I'd suggest 2 methods: scope.trainable_variables and scope.global_variables -- to mimic the TF methods.

Is this PR still in active development?

lukaszkaiser:
Or we could have all 3: get_collection, trainable_variables, and global_variables.

tensorflow/python/ops/variable_scope.py
suiyuan2009: add a new test file? I didn't find the test file to variable_scope.py.

suiyuan2009:
@lukaszkaiser I'll add trainable_variables and global_variables.

lukaszkaiser:
The test file is in ../kernel_tests/variable_scope_test (not in ops).

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -721,6 +721,48 @@ def device_func(op):
     self.assertEqual(varname_type[0], ("x", dtypes.float32))
     self.assertEqual(varname_type[1], ("y", dtypes.int64))
 
+  def testGetCollection(self):
+    with self.test_session():
+      a = variable_scope.get_variable("a", [])
+      b = variable_scope.get_variable("b", [], trainable=False)
+      with variable_scope.variable_scope("foo_") as scope:
```
lukaszkaiser: Could this be "scope1" and the next "scope2" and we test variables in both?

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -721,6 +721,48 @@ def device_func(op):
...
+            ops.GraphKeys.GLOBAL_VARIABLES)],
+          ["a:0", "b:0", "foo_/a:0", "foo_/b:0", "foo/a:0", "foo/b:0"])
+      self.assertEqual(
+          [v.name for v in scope.get_collection(
+            ops.GraphKeys.TRAINABLE_VARIABLES)],
+          ["a:0", "foo_/a:0", "foo/a:0"])
+
+  def testGetTrainableVariables(self):
+    with self.test_session():
+      a = variable_scope.get_variable("a", [])
```
lukaszkaiser: Please first add a variable (say "a"), then make a scope("foo") as scope, add "b" and non-trainable "c", and get trainable variables from scope and check that it's only b.

tensorflow/python/kernel_tests/variable_scope_test.py
```diff
@@ -721,6 +721,48 @@ def device_func(op):
...
+      b = variable_scope.get_variable("b", [], trainable=False)
+      scope = variable_scope.get_variable_scope()
+      self.assertEqual(scope.trainable_variables()[0].name, "a:0")
+
+  def testGetGlobalVariables(self):
+    with self.test_session():
+      a = variable_scope.get_variable("a", [])
+      b = variable_scope.get_variable("b", [], trainable=False)
```
lukaszkaiser: Again, please add a variable and then open a scope, and check that the first one isn't in the ones we get.

tensorflow/python/ops/variable_scope.py
```diff
@@ -887,6 +887,14 @@ def get_collection(self, name):
     scope = self._name + '/' if self._name else self._name
```
lukaszkaiser: In the else-branch, is self._name an empty string? None? Could we just use a constant after else here?

lukaszkaiser (changes requested):
It's almost ready to go in, but a few small style and test changes would be good. Thanks!

tensorflow/python/ops/variable_scope.py
suiyuan2009: empty string in the constructor.

lukaszkaiser (approved):
Thanks!

lukaszkaiser:

caisq:
@martinwicke please let us know whether it is okay to merge this PR, which involved additive API changes.

suiyuan2009:
the failed test target is passed on my computer.

caisq:
@suiyuan2009 The failure you saw on our Jenkins is from a flaky test. You can ignore that.

dandelionmane:
@tensorflow-jenkins test this please (trying to deflake)
gentle ping @martinwicke

martinwicke:
Ok, LGTM. Go ahead and merge.

PR #7978, opened 2017-03-01
Ignore generated bazel_config while configure

Lewuathe:
`./configure` with default setting generates untracked files of bazel configs.

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	tensorflow/core/platform/default/build_config.bzl-e
	tensorflow/core/platform/default/build_config_root.bzl-e
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ node_modules
 __pycache__
 *.swp
 .vscode/
+build_config.bzl-e
+build_config_root.bzl-e


jart:
Do you know what exactly is generating those files? I haven't seen them before.

jart:
I see what the problem is. It's a cross platform sed issue. I think the solution is to not have sed generate those files on Mac. Will send out PR soon.

PR #7992, opened 2017-03-02
Update MNIST source url, original url is dead

MostafaGazar:
http://yann.lecun.com/exdb/mnist/ is not accessible

### DIFF
--- a/tensorflow/contrib/learn/python/learn/datasets/mnist.py
+++ b/tensorflow/contrib/learn/python/learn/datasets/mnist.py
@@ -27,7 +27,7 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 from tensorflow.contrib.learn.python.learn.datasets import base
 from tensorflow.python.framework import dtypes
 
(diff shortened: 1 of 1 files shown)


gunan:
@ilblackdragon @martinwicke 
I don't think I can accept this PR.
I am not sure we would like to have a reference to a personal repository. We can think about alternate storage options for this dataset. Ilia, Martin, should we host this file instead? Are there any license implications we need to consider?

MostafaGazar:
@gunan That would be the ideal solution. I think you should also consider hosting the files in https://github.com/tensorflow/tensorflow/commit/d2a5b7c8f9bd789423566c56c01d1f92cc94d74d in some place you can manage if license allows

martinwicke:
Yeah, apparently there's been a change and it now requires a signup?

Anyway, we have asked before and we've always been turned down for re-hosting. Possibly we'll have to remove mnist altogether. Let me ask again.

MostafaGazar:
Looks like this http://yann.lecun.com/exdb/mnist/ is working fine now!

Issue #7966, opened 2017-03-01
Compile TensorFlow 1.0 on Windows 10 for Python 3.6 (Anaconda) cause compile issues


pavelbulanov:
As per agreement in #6999 [specific comment](https://github.com/tensorflow/tensorflow/issues/6999#issuecomment-280749568), I'm creating a new issue for compilation errors for TF 1.0 under Windows for Python 3.6

**Pre-requisites:**
Windows 10 v.14986
Python 3.6 64 bit (Anaconda 4.3)
git version 2.11.0.windows.1
Visual Studio Build Tools (called "Build Tools for Visual Studio 2017 RC" in the [download](https://www.visualstudio.com/downloads/#build-tools-for-visual-studio-2017-rc) section)
cmake-3.8.0-rc1 and swigwin-3.0.12 (just downloaded latest version at that moment)
TensorFlow 1.0.0 from source (master branch from github)

**Reproduction**. I followed [instruction](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/README.md) for CMake just running two commands - CMake first and MSBuild second.

To compile, I made two changes in sources. I found them based on compilation errors.
One was adding `#include <intrin.h>` to `tensorflow\core\platform\windows\cpu_info.h` inside of `#ifndef` block because of the following error:
> error C3861: '__cpuidex': identifier not found 

I found [commit](https://github.com/tensorflow/tensorflow/commit/a672cb166dae93ae955c1d38f3de8903dd242373) introduced usage of `__cpuidex` function. Based on [this](https://msdn.microsoft.com/en-us/library/hskdteyh.aspx) MSDN link, I discovered that Header file `<intrin.h>` has to be included.  That's why I put it inside `cpu_info.h` (not sure if it's exactly perfect place).

**Second issue** was commenting out procedures `_mm256_extract_epi32` and `_mm256_insert_epi32` in `tensorflow\core\platform\windows\intrinsics_port.h` due to error

> error C2169: '_mm256_extract_epi32': intrinsic function, cannot be defined 

The second issue with `_mm256_extract_epi32` may be connected with the same reason, I don't know. 
[...]

PR #8008, opened 2017-03-02
fix compiler issues with intrinsics under Windows for Build Tools for VS2017

pavelbulanov:
Based on discussion in Issue #7966 

two changes made:
1) add `#include <intrin.h>` to `tensorflow\core\platform\windows\cpu_info.h ` for availability of `__cpuidex` function based on [this](https://msdn.microsoft.com/en-us/library/hskdteyh.aspx)

2) do not define functions `_mm256_extract_epi32 `and `_mm256_insert_epi32 `in `tensorflow/core/platform/windows/intrinsics_port.h` because they are already declared in `immintrin.h` in VS2017 (as extern as they are intrinsics now)

Quote from `immintrin.h`
```
// Insert integer into 256-bit packed integer array at element selected by index
extern __m256i __cdecl _mm256_insert_epi8 (__m256i /* dst */, int /* src */, const int /* index */);
extern __m256i __cdecl _mm256_insert_epi16(__m256i /* dst */, int /* src */, const int /* index */);
extern __m256i __cdecl _mm256_insert_epi32(__m256i /* dst */, int /* src */, const int /* index */);
#if defined(_M_X64)
extern __m256i __cdecl _mm256_insert_epi64(__m256i /* dst */, __int64 /* src */, const int /* index */);
#endif  // defined (_M_X64)

// Extract integer element selected by index from 256-bit packed integer array
extern int __cdecl _mm256_extract_epi8 (__m256i /* src */, const int /* index */);
extern int __cdecl _mm256_extract_epi16(__m256i /* src */, const int /* index */);
extern int __cdecl _mm256_extract_epi32(__m256i /* src */, const int /* index */);
#if defined(_M_X64)
extern __int64 __cdecl _mm256_extract_epi64(__m256i /* src */, const int /* index */);
#endif  // defined (_M_X64)
```

### DIFF
--- a/tensorflow/core/platform/windows/intrinsics_port.h
+++ b/tensorflow/core/platform/windows/intrinsics_port.h
@@ -25,8 +25,8 @@ limitations under the License.
 
 #define _mm_load_pd1 _mm_load1_pd
 
-// only define these intrinsics if imminitrin.h by Intel wasn't included
-#ifndef _INCLUDED_IMM
+// only define these intrinsics if immintrin.h doesn't have them (VS2015 and earlier)
+#if _MSC_VER < 1910
 static inline int
 _mm256_extract_epi32(__m256i a, const int i)
 {
--- a/tensorflow/core/platform/windows/intrinsics_port.h
+++ b/tensorflow/core/platform/windows/intrinsics_port.h
@@ -24,6 +24,9 @@ limitations under the License.
 #include "tensorflow/core/platform/types.h"
 
 #define _mm_load_pd1 _mm_load1_pd
+
+// only define these intrinsics if imminitrin.h by Intel wasn't included
+#ifndef _INCLUDED_IMM
 static inline int
 _mm256_extract_epi32(__m256i a, const int i)
 {
@@ -37,5 +40,6 @@ _mm256_insert_epi32(__m256i a, int b, const int i)
   c.m256i_i32[i & 7] = b;
   return c;
 }
+#endif
 #endif
 #endif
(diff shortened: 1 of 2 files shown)


gunan:
Jenkins, test this please.

tensorflow/core/platform/windows/intrinsics_port.h
```diff
@@ -24,6 +24,9 @@ limitations under the License.
 #include "tensorflow/core/platform/types.h"
 
 #define _mm_load_pd1 _mm_load1_pd
+
+// only define these intrinsics if imminitrin.h by Intel wasn't included
+#ifndef _INCLUDED_IMM
```
guschmue: I think this will break vs2015 builds if compiled with /arch:avx2.
maybe instead of 
#ifndef _INCLUDED_IMM
use
#if _MSC_VER < 1910
to preserve vs2015 since this is what is out there.

CI will not see this because it does not build for avx2 but out does.

gunan:
Jenkins, test this please.

mrry:
I think the test failure was a flake....

mrry (approved):

guschmue:
Since there is no CI for vs2017 yet I ran the python unit tests for CPU manually - looks good.
  ctest -C RelWithDebInfo
  100% tests passed, 0 tests failed out of 226
  Total Test time (real) = 4348.83 sec

Tried a GPU build but there is no CUDA SDK for vs2017 yet.

mrry:
Thanks for confirming that, Guenther!

PR #8028, opened 2017-03-02
Fix local_test.sh to respect --num_workers and --num_parameter_servers

cberner:
Previously, 2 workers and 2 parameters servers were always used,
regardless of the value of these flags.


caisq (approved):

caisq:

PR #8032, opened 2017-03-02
Add support for the matmul (@) operator.

### LINKED ISSUE #1062: python3.5 support and @ operator with  __matmul__  method
Are there any plans to include Python3.5 support (incl. binaries) and [dedicated matmul `@` operator](https://www.python.org/dev/peps/pep-0465/)?

ageron:
This PR closes #1062.
Assuming `a` and `b` are tensors, `a @ b`  is equivalent to `tf.matmul(a, b)`.

Note that either `a` or `b` can be a NumPy array or a regular python array, for example the following expressions are valid:
* `a @ [[1.], [2.]]`
* `[[1., 2.]] @ a`
* `a @ np.array([[1.], [2.]])`
* `np.array([[1., 2.]]) @ a`

### DIFF
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -265,7 +265,9 @@ class Tensor(_TensorLike):
       # Unary.
       "__invert__",
       "__neg__",
-      "__abs__"
+      "__abs__",
+      "__matmul__",
+      "__rmatmul__"
   }
 
   def __init__(self, op, value_index, dtype):
--- a/tensorflow/python/kernel_tests/matmul_op_test.py
+++ b/tensorflow/python/kernel_tests/matmul_op_test.py
@@ -19,7 +19,7 @@ from __future__ import division
 from __future__ import print_function
 
 import numpy as np
-import sys
+import operator
 
 from tensorflow.python.framework import constant_op
 from tensorflow.python.framework import dtypes
@@ -442,11 +442,10 @@ class MatMulStatsTest(test.TestCase):
           self.assertEqual(7200, flops)
 
 
-# @ operator supported since python 3.5.
-if sys.version_info >= (3, 5):
-  # Use eval() to avoid a compilation error on earlier versions.
-  infix_matmul = lambda x, y: eval("x @ y")
-else:
+try:
+    # @ operator supported since python 3.5.
+    infix_matmul = operator.matmul
+except AttributeError:
   # For earlier versions of python, emulate regular behavior.
   # Useful to build and test for 3.5+ on earlier versions.
   def infix_matmul(x, y):
@@ -484,10 +483,18 @@ class MatMulInfixOperatorTest(test.TestCase):
         ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
 
   def testInfixMatmulIsTfMatmul(self):
-      a = ops.convert_to_tensor([[10.0, 20.0, 30.0]])
-      b = ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0], [80.0, 90.0]])
-      c = infix_matmul(a, b)
-      self.assertEqual(c.op.type, 'MatMul')
+    a = ops.convert_to_tensor([[10.0, 20.0, 30.0]])
+    b = ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0], [80.0, 90.0]])
+    c = infix_matmul(a, b)
+    self.assertEqual(c.op.type, 'MatMul')
+
+  def testInfixMatmulDoesDotProduct(self):
+    a = ops.convert_to_tensor([[10.0, 20.0, 30.0]])
+    b = ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0], [80.0, 90.0]])
+    c = infix_matmul(a, b)
+    d = math_ops.matmul(a, b)
+    with self.test_session():
+      self.assertAllEqual(c.eval(), d.eval())
 
 
 if __name__ == "__main__":
--- a/tensorflow/python/kernel_tests/matmul_op_test.py
+++ b/tensorflow/python/kernel_tests/matmul_op_test.py
@@ -483,6 +483,12 @@ class MatMulInfixOperatorTest(test.TestCase):
         ops.convert_to_tensor([[10.0, 20.0, 30.0]]),
         ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
 
+  def testInfixMatmulIsTfMatmul(self):
+      a = ops.convert_to_tensor([[10.0, 20.0, 30.0]])
+      b = ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0], [80.0, 90.0]])
+      c = infix_matmul(a, b)
+      self.assertEqual(c.op.type, 'MatMul')
+
 
 if __name__ == "__main__":
   test.main()
--- a/tensorflow/python/kernel_tests/matmul_op_test.py
+++ b/tensorflow/python/kernel_tests/matmul_op_test.py
@@ -445,11 +445,11 @@ class MatMulStatsTest(test.TestCase):
 # @ operator supported since python 3.5.
 if sys.version_info >= (3, 5):
   # Use eval() to avoid a compilation error on earlier versions.
-  _MATMUL = lambda x, y: eval("x @ y")
+  infix_matmul = lambda x, y: eval("x @ y")
 else:
   # For earlier versions of python, emulate regular behavior.
   # Useful to build and test for 3.5+ on earlier versions.
-  def _MATMUL(x, y):
+  def infix_matmul(x, y):
     try:
       r = type(x).__matmul__(x, y)
     except AttributeError:
@@ -472,14 +472,14 @@ class MatMulInfixOperatorTest(test.TestCase):
   def testMismatchedShape(self):
     with self.assertRaisesWithPredicateMatch(
         ValueError, lambda e: "Shape must" in str(e)):
-      _MATMUL(
+      infix_matmul(
         ops.convert_to_tensor([10.0, 20.0, 30.0]),
         ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
 
   def testMismatchedDimensions(self):
     with self.assertRaisesWithPredicateMatch(
         ValueError, lambda e: "Dimensions must" in str(e)):
-      _MATMUL(
+      infix_matmul(
         ops.convert_to_tensor([[10.0, 20.0, 30.0]]),
         ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
 
--- a/tensorflow/python/kernel_tests/matmul_op_test.py
+++ b/tensorflow/python/kernel_tests/matmul_op_test.py
@@ -466,7 +466,7 @@ except AttributeError:
     return r
 
 
-class MatMulInfixOperatorTest(test.TestCase):
+class MatMulInfixOperatorTest(test_lib.TestCase):
 
   def testMismatchedShape(self):
     with self.assertRaisesWithPredicateMatch(
--- a/tensorflow/python/kernel_tests/matmul_op_test.py
+++ b/tensorflow/python/kernel_tests/matmul_op_test.py
@@ -19,6 +19,7 @@ from __future__ import division
 from __future__ import print_function
 
 import numpy as np
+import sys
 
 from tensorflow.python.framework import constant_op
 from tensorflow.python.framework import dtypes
@@ -439,6 +440,48 @@ class MatMulStatsTest(test.TestCase):
         flops = ops.get_stats_for_node_def(g, op.node_def, "flops").value
         if op.name == "MatMul":
           self.assertEqual(7200, flops)
+
+
+# @ operator supported since python 3.5.
+if sys.version_info >= (3, 5):
+  # Use eval() to avoid a compilation error on earlier versions.
+  _MATMUL = lambda x, y: eval("x @ y")
+else:
+  # For earlier versions of python, emulate regular behavior.
+  # Useful to build and test for 3.5+ on earlier versions.
+  def _MATMUL(x, y):
+    try:
+      r = type(x).__matmul__(x, y)
+    except AttributeError:
+      r = NotImplemented
+    if r is NotImplemented and type(x) is not type(y):
+      try:
+        r = type(y).__rmatmul__(y, x)
+      except AttributeError:
+        r = NotImplemented
+    if r is NotImplemented:
+      raise TypeError(
+        "unsupported operand type(s) for @: '{}' and '{}'"
+        .format(type(x).__name__, type(y).__name__)
+        )
+    return r
+
+
+class MatMulInfixOperatorTest(test.TestCase):
+
+  def testMismatchedShape(self):
+    with self.assertRaisesWithPredicateMatch(
+        ValueError, lambda e: "Shape must" in str(e)):
+      _MATMUL(
+        ops.convert_to_tensor([10.0, 20.0, 30.0]),
+        ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
+
+  def testMismatchedDimensions(self):
+    with self.assertRaisesWithPredicateMatch(
+        ValueError, lambda e: "Dimensions must" in str(e)):
+      _MATMUL(
+        ops.convert_to_tensor([[10.0, 20.0, 30.0]]),
+        ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
 
 
 if __name__ == "__main__":
--- a/tensorflow/python/kernel_tests/matmul_op_test.py
+++ b/tensorflow/python/kernel_tests/matmul_op_test.py
@@ -443,8 +443,8 @@ class MatMulStatsTest(test.TestCase):
 
 
 try:
-    # @ operator supported since python 3.5.
-    infix_matmul = operator.matmul
+  # @ operator supported since python 3.5.
+  infix_matmul = operator.matmul
 except AttributeError:
   # For earlier versions of python, emulate regular behavior.
   # Useful to build and test for 3.5+ on earlier versions.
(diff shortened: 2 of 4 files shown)


tensorflow/python/ops/math_ops.py
```diff
@@ -1771,6 +1776,8 @@ def matmul(a,
           a, b, transpose_a=transpose_a, transpose_b=transpose_b, name=name)
 
 
+_OverrideBinaryOperatorHelper(matmul, "matmul")
```
ageron: I did not add `if sys.version_info >= (3, 5)` because it probably does not hurt to add `__matmul__()` and `__rmatmul__()`, and it simplifies the code and the tests.

tensorflow/python/framework/ops.py
```diff
@@ -265,7 +265,9 @@ class Tensor(_TensorLike):
       # Unary.
       "__invert__",
       "__neg__",
-      "__abs__"
+      "__abs__",
+      "__matmul__",
```
ageron: This is needed to allow `_OverrideBinaryOperatorHelper(matmul, "matmul")` in `math_ops.py`.

tensorflow/python/kernel_tests/cwise_ops_test.py
```diff
@@ -55,6 +56,30 @@
...
 _INV = lambda x: ~x
 
+# @ operator supported since python 3.5.
+if sys.version_info >= (3, 5):
+  # Use eval() to avoid a compilation error on earlier versions.
+  _MATMUL = lambda x, y: eval("x @ y")
+else:
+  # For earlier versions of python, emulate regular behavior.
+  # Useful to build and test for 3.5+ on earlier versions.
+  def _MATMUL(x, y):
```
ageron: I did my best to emulate the @ operator properly so the tests would work whether on python 3.5+ or with earlier versions.  Simply calling `type(x).__matmul__(x, y)` obviously does not work when `x` is a python array (since it is not defined), but it does not work either when `x` is a NumPy array, because it returns `NotImplemented` when `y` is not a NumPy array and it implements `__rmatmul__()`.  Lines 74 to 80 are unused by the current tests, but this code is simple enough and allows adding more tests without bad surprises, so I think it shouldn't be shortened.

tensorflow/python/kernel_tests/cwise_ops_test.py
```diff
@@ -587,7 +612,8 @@ def _compareBoth(self, x, y, np_func, tf_func, also_compare_variables=False):
     self._compareCpu(x, y, np_func, tf_func, also_compare_variables)
     if x.dtype in (np.float16, np.float32, np.float64):
       if tf_func not in (_FLOORDIV, math_ops.floordiv, math_ops.igamma,
-                         math_ops.igammac, math_ops.zeta, math_ops.polygamma):
+                         math_ops.igammac, math_ops.zeta, math_ops.polygamma,
+                         _MATMUL):
```
ageron: I added `_MATMUL` to this list because I was getting an error in `testFloatBasic()` due to a lack of precision in `_compareGradientX()` (which calls `assertAllClose()`).  This might be worth investigating, but it does not seem related to this PR.  Here is the output of the error:

    Not equal to tolerance rtol=0.001, atol=0.001
    
    (mismatch 11.111111111111114%)
     x: array([[ 22.      ,  20.035713,  18.07143 ,   0.      ,   0.      ,
              0.      ,   0.      ,   0.      ,   0.      ],
           [ 16.107143,  14.142857,  12.178572,   0.      ,   0.      ,...
     y: array([[ 22.003172,  20.03479 ,  18.074036,   0.      ,   0.      ,
              0.      ,   0.      ,   0.      ,   0.      ],
           [ 16.098022,  14.144897,  12.176513,   0.      ,   0.      ,...

caisq:
cc @ibab

caisq:

ageron:
@caisq That's odd, 2 tests fail with a weird error: "Unable to get pull request builder trigger".  I'm guessing this is a temporary error? Perhaps just retry? How can I help?

caisq:
@ageron. I don't see the errors now. Maybe those were just transient. The //tensorflow/python:localhost_cluster_performance_test failure in

tensorflow/python/kernel_tests/cwise_ops_test.py
girving: The matmul operator doesn't really belong in the *componentwise* operator test file.  I think it would be better to make a new test file.  I'll hold off reviewing this file past this point, since I expect you'll have to make a lot of strange workarounds for the fact that matmul is the furthest thing from componentwise.

In particular, you shouldn't need to run tests of `@` if you're not in Python >= 3.5, since that doesn't make any sense.  So `_MATMUL` should be superfluous.

tensorflow/python/kernel_tests/cwise_ops_test.py
girving: You shouldn't need to test gradients for `@`, since it's just a wrapper around `matmul`.  Gradient tests are for new ops.  This complexity will go away once you make a new file with targeted tests of `@`.

tensorflow/python/ops/math_ops.py
```diff
@@ -1685,6 +1685,11 @@ def matmul(a,
                            [229 244]],
                           [[508 532]
                            [697 730]]]
+
+  # Since python >= 3.5 and TensorFlow >= 1.1, the @ operator is supported.
```
girving: Don't mention TensorFlow 1.1.  This comment is unlikely to be read by someone reading the source code for an earlier version of TensorFlow.

tensorflow/python/ops/math_ops.py
girving: Agreed.  In particular, it means that downstream libraries are free to call `__matmul__` without worrying about version grudginess.

Which makes me understand your `_MATMUL` function above. :)  Okay, that should stay.

tensorflow/python/kernel_tests/cwise_ops_test.py
girving: Actually, after reading further, it's reasonable to test that `__matmul__` works as expected in lower versions of Python.  So keep `_MATMUL`, but still split out into a separate test file.

Alternately, you could put it in `matmul_op_test.py`.

girving (changes requested):
Please don't merge other branches into this one during a review.  Either remove the merge commit from the pull request branch or rebase so that it goes away (but then we lose review comment locations).

ageron:
@girving Thanks for your review. Sorry for the merge commit during review, I'll remove it.  I'll also add the separate test file for the @ operator.

ageron:
@girving As you requested, I moved the test code from cwise_ops_test.py to  matmul_op_test.py, and I removed the mention of TensorFlow 1.1 from the documentation.

tensorflow/python/kernel_tests/matmul_op_test.py
```diff
@@ -441,5 +442,47 @@ def testTransposedStatistics(self):
           self.assertEqual(7200, flops)
 
 
+# @ operator supported since python 3.5.
+if sys.version_info >= (3, 5):
+  # Use eval() to avoid a compilation error on earlier versions.
+  _MATMUL = lambda x, y: eval("x @ y")
+else:
```
ageron: Lines 460 to 466 are unused for now, but I think it's better to keep them here just in case someone adds more tests: they make _MATMUL work very much the same way whatever the version of python used to run the tests.

tensorflow/python/kernel_tests/matmul_op_test.py
```diff
@@ -441,5 +442,47 @@ def testTransposedStatistics(self):
           self.assertEqual(7200, flops)
 
 
+# @ operator supported since python 3.5.
+if sys.version_info >= (3, 5):
+  # Use eval() to avoid a compilation error on earlier versions.
+  _MATMUL = lambda x, y: eval("x @ y")
```
girving: Rename `_MATMUL` -> `infix_matmul`.

tensorflow/python/kernel_tests/matmul_op_test.py
```diff
@@ -441,5 +442,47 @@ def testTransposedStatistics(self):
...
+        ops.convert_to_tensor([10.0, 20.0, 30.0]),
+        ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
+
+  def testMismatchedDimensions(self):
+    with self.assertRaisesWithPredicateMatch(
+        ValueError, lambda e: "Dimensions must" in str(e)):
+      _MATMUL(
+        ops.convert_to_tensor([[10.0, 20.0, 30.0]]),
+        ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
+
```
girving: You should also check that `infix_matmul` is the same as `matmul`.

girving (changes requested):

tensorflow/python/kernel_tests/matmul_op_test.py
```diff
@@ -441,5 +442,53 @@ def testTransposedStatistics(self):
           self.assertEqual(7200, flops)
 
 
+# @ operator supported since python 3.5.
+if sys.version_info >= (3, 5):
+  # Use eval() to avoid a compilation error on earlier versions.
+  infix_matmul = lambda x, y: eval("x @ y")
```
shoyer: Instead, try:
```
try:
    infix_matmul = operator.matmul
except AttributeError:
    def infix_matmul(x, y):
        ...
```

It avoids eval and explicit sys.version_info checks

tensorflow/python/kernel_tests/matmul_op_test.py
```diff
@@ -483,6 +483,12 @@ def testMismatchedDimensions(self):
         ops.convert_to_tensor([[10.0, 20.0, 30.0]]),
         ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0]]))
 
+  def testInfixMatmulIsTfMatmul(self):
+      a = ops.convert_to_tensor([[10.0, 20.0, 30.0]])
+      b = ops.convert_to_tensor([[40.0, 50.0], [60.0, 70.0], [80.0, 90.0]])
+      c = infix_matmul(a, b)
+      self.assertEqual(c.op.type, 'MatMul')
```
girving: I may be paranoid, but please also do the check that

    self.assertAllEqual(c.eval(), tf.matmul(a, b).eval())

girving (changes requested):

ageron:
@girving I added the test you requested.

ageron:
@shoyer Great point, thanks. I replaced `eval("x @ y")` by `operator.matmul`.

girving (approved):

girving:
Jenkins, test this please.

girving:
@gunan What does this sort of build abort mean?

    Build was aborted
    Aborted by unknown
    Setting status of 0c19220e483aabfd25e5453fed9f7ef896c9b4aa to FAILURE with url https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu/3951/ and message: 'FAILURE
 '
    Using context: Linux CPU Tests
    Finished: ABORTED

girving:
Ah, right.  Apologies for forgetting to look at the sanity checks again.

@ageron Can you fix these indentation errors?  @caisq Can @ageron check such things locally?

caisq:
@girving, @ageron can use the following commands to run the sanity build locally (requires docker):
```
tensorflow/tools/ci_build/ci_build.sh CPU tensorflow/tools/ci_build/ci_sanity.sh
```

ageron:
@girving @caisq Sorry about that, I'm on it.

ageron:
@girving @caisq Your tool worked, thanks. I fixed the indentation, everything should be green now.

caisq:

ageron:
@caisq Oh noooo, what's wrong now? Apparently a performance test failed, correct? Perhaps the performance tests need to be adjusted to be la bit less strict. Wdyt?

girving (approved):

girving:
The errors look like

        class MatMulInfixOperatorTest(test.TestCase):
    NameError: name 'test' is not defined

girving:
I think it should be `test_lib.TestCase`.

girving:
Did you try running the tests locally?

caisq:
@ageron , there are some remaining problems. 

See https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu/3955/consoleFull

```
    class MatMulInfixOperatorTest(test.TestCase):
NameError: name 'test' is not defined
```

Please use test_util.TensorFlowTestCase instead. There are plenty of examples in other py test files.

For the Linux CPU test, you can run the tests locally by using

```
export TF_BUILD_CONTAINER_TYPE=CPU
export TF_BUILD_PYTHON_VERSION=PYTHON2
export TF_BUILD_IS_OPT=OPT
export TF_BUILD_IS_PIP=PIP
export TF_BUILD_APPEND_CI_DOCKER_EXTRA_PARAMS="-e TF_BUILD_FILTER_INSTALL_TESTS_BY_TAG=-manual"
tensorflow/tools/ci_build/ci_parameterized_build.sh
```

ageron:
@caisq @girving Yes, I fully tested on my machine using the command you provided, and everything worked fine.  The problem is the test file was updated in the meantime by other people (among other things, test was renamed to test_lib), and I need to merge their changes.  But earlier, Geoffrey told me *not* to merge during a code review, so I'm a bit stuck.  What should I do? Merge master despite the ongoing code review, test locally and commit? Or...? I see no other option.

ageron:
@caisq @girving Ok, I replaced "test.TestCase" with "test_lib.TestCase". Please try again... I hope this will be it.  I ran the tests on my machine, everything worked fine:

```
[...]
//bazel_pip/tensorflow/tensorboard/lib/python:http_util_test             PASSED in 3.3s
//bazel_pip/tensorflow/tensorboard/lib/python:json_util_test             PASSED in 3.3s
//bazel_pip/tensorflow/tensorboard/plugins/debugger:debugger_plugin_test PASSED in 3.7s
//bazel_pip/tensorflow/tensorboard/plugins/projector:projector_plugin_test PASSED in 5.1s

Executed 570 out of 570 tests: 570 tests pass.

=== Testing user ops ===
PYTHON_BIN_PATH: /workspace/pip_test/venv/bin/python
/tmp/tmp.UxvZtfODLd /workspace
TensorFlow include path: /workspace/pip_test/venv/local/lib/python2.7/site-packages/tensorflow/include

g++ version:
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/5/lto-wrapper
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 5.4.0-6ubuntu1~16.04.4' --with-bugurl=file:///usr/share/doc/gcc-5/README.Bugs --enable-languages=c,ada,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-5 --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-5-amd64/jre --enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-5-amd64 --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-5-amd64 --with-arch-directory=amd64 --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --enable-objc-gc --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) 

Extra GPP flag: 
Testing user ops in CPU environment
Compiling user op C++ source file zero_out_op_kernel_1.cc
Invoking user op zero_out defined in file zero_out.so via pip installation
2017-03-07 10:07:46.865252: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.1 instructions, but these are available on your machine and could speed up CPU computations.
2017-03-07 10:07:46.865286: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.2 instructions, but these are available on your machine and could speed up CPU computations.
2017-03-07 10:07:46.865298: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX instructions, but these are available on your machine and could speed up CPU computations.
2017-03-07 10:07:46.865309: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use AVX2 instructions, but these are available on your machine and could speed up CPU computations.
2017-03-07 10:07:46.865321: W tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use FMA instructions, but these are available on your machine and could speed up CPU computations.
Output from user op ([42,0,0]) matches expected output
/workspace

SUCCESS: Testing of user ops PASSED

Parameterized build ends with SUCCESS at: mardi 7 mars 2017, 11:07:47 (UTC+0100) (Elapsed time: 2794 s)

```

ageron:

caisq:

caisq:

caisq:
@ageron All tests passed. Yay! 
@girving Should be okay to merge unless @shoyer has more comments?

ageron:
Joy! Happiness! The tests pass at last, thanks @caisq.

girving (approved):

girving:
Merged.  Thank you for the contribution!

ageron:
Awesome, thanks Geoffrey and Shanqing.

PR #8046, opened 2017-03-03
Windows: Fix error in release script for C library

asimshankar:
Bug in 8777d9a15614b65271d25401f488b70b6b3dca1e

### DIFF
--- a/tensorflow/tools/ci_build/windows/libtensorflow_cpu.sh
+++ b/tensorflow/tools/ci_build/windows/libtensorflow_cpu.sh
@@ -76,14 +76,12 @@ zip -j ${DIR}/libtensorflow_jni-cpu-windows-$(uname -m).zip \
 rm -f ${DIR}/tensorflow_jni.dll
 
 # Zip up the .dll, LICENSE and include files for the C library.
+mkdir -p ${DIR}/include/tensorflow/c
+mkdir -p ${DIR}/lib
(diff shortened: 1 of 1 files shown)


tensorflow/tools/ci_build/windows/libtensorflow_cpu.sh
```diff
@@ -76,14 +76,12 @@ zip -j ${DIR}/libtensorflow_jni-cpu-windows-$(uname -m).zip \
 rm -f ${DIR}/tensorflow_jni.dll
 
 # Zip up the .dll, LICENSE and include files for the C library.
+mkdir -p ${DIR}/include/tensorflow/c
+mkdir -p ${DIR}/lib
+cp bazel-bin/tensorflow/libtensorflow.so ${DIR}/lib/tensorflow.dll
```
gunan: we have another relative address in the source here. Do we also need to fix this?

asimshankar: I think it's fine. Until the time we create the .Zip for the C library, we're in the root of the git repo, so at that time bazel-bin is fine. Or did I miss something?

gunan: AFAICT, no :)

gunan (approved):

PR #8058, opened 2017-03-03
Made sequence_length optional in bidirectional dynamic rnn

### LINKED ISSUE #5588: bidirectional_dynamic_rnn: make sequence length optional
Qualifier: very new to TF, so I may be missing something very obvious.  If so, apologies.

DOCUMENTATION: 

Is a little conflicting:

"sequence_length: An int32/int64 vector, size [batch_size], containing the actual lengths for each of the sequences." <-- implies sequence_length is required

However...

"The initial state for both directions is zero by default (but can be set optionally) and no intermediate states are ever returned -- the network is fully unrolled for the given (passed in) length(s) of the sequence(s) or completely unrolled if length(s) is not given." <-- implies sequence_length is not required.

FUNCTIONALITY"

Base on the function itself ("_dynamic_"), I would assume that allowing sequence_length=None is intended (although, in either case, the documentation should be straightened out).

Assuming we do want to allow seq_length=None, the offending portion within bidirectional_dynamic_rnn would appear to be:

inputs_reverse = array_ops.reverse_sequence(...seq_lengths=sequence_length...)

reverse_sequence appears to require seq_lengths != None

Oddly enough, bidirectional_rnn may(?) be robust to this problem, as it uses _reverse_seq, which handles lengths=None.  

That said, I haven't tested this extensively, as I'm of course not 100% sure I've interpreted the above correctly and as to what the intended functionality is.

Given guidance, I could take a stab at a pull request to address the above.  Or maybe this is a quick fix/known issue (or non-issue...).

persiyanov:
Closes #5588

### DIFF
--- a/tensorflow/python/ops/rnn.py
+++ b/tensorflow/python/ops/rnn.py
@@ -266,6 +266,10 @@ def bidirectional_dynamic_rnn(cell_fw, cell_bw, inputs, sequence_length=None,
   length(s) of the sequence(s) or completely unrolled if length(s) is not
   given.
 
+  When not passing `sequence_length`, make sure that you do not use padding
+  (i.e. your sequences really have the same lengths), because it can lead you
+  to wrong final cell states.
+
   Args:
     cell_fw: An instance of RNNCell, to be used for forward direction.
     cell_bw: An instance of RNNCell, to be used for backward direction.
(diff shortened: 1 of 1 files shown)


caisq:

caisq:
@persiyanov FYI, you need to resolve conflicts in rnn.py first.

ebrevdo:
I just pushed this fix internally?

caisq:
In that case I will close this PR. Thank you for sending this PR all the same, @persiyanov

PR #8070, opened 2017-03-04
Cherry-pick: Add the graphdef version to InferenceContext and to ShapeRefiner::Add…

cwhipkey:
…Node.

Use this to allow loading reductions saved with older graphdefs.

Change GraphConstructor to not increase the version when importing, but instead take the min of all versions.
Change: 149152437


cwhipkey:
Jenkins, test this please

jhseu (approved):

PR #8075, opened 2017-03-04
Update the github docs link in the error msg upon import failure

rasbt:
Hi,
since the docs have moved from `g3doc` to `docs_src` (and have been refactored), I thought it would be good to update the link in the error message that comes up upon import failures.

### DIFF
--- a/tensorflow/python/pywrap_tensorflow.py
+++ b/tensorflow/python/pywrap_tensorflow.py
@@ -46,8 +46,12 @@ try:
     sys.setdlopenflags(_default_dlopen_flags)
 except ImportError:
   msg = """%s\n\nFailed to load the native TensorFlow runtime.\n
-See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#import_error\n
-for some common reasons and solutions.  Include the entire stack trace
+Visit https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/install/index.md\n
+and click on the link that corresponds to your operating system to read about\n
+common reasons and solutions in the "Common installation problems"\n
+subsection.
+
+Include the entire stack trace
 above this error message when asking for help.""" % traceback.format_exc()
   raise ImportError(msg)
 
--- a/tensorflow/python/pywrap_tensorflow.py
+++ b/tensorflow/python/pywrap_tensorflow.py
@@ -46,7 +46,7 @@ try:
     sys.setdlopenflags(_default_dlopen_flags)
 except ImportError:
   msg = """%s\n\nFailed to load the native TensorFlow runtime.\n
-Visit https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/install/index.md
+Visit https://www.tensorflow.org/install/
 and click on the link that corresponds to your operating system to read about
 common reasons and solutions in the "Common installation problems"
 subsection.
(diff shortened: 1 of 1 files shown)


tensorflow/python/pywrap_tensorflow.py
```diff
@@ -46,8 +46,12 @@
     sys.setdlopenflags(_default_dlopen_flags)
 except ImportError:
   msg = """%s\n\nFailed to load the native TensorFlow runtime.\n
-See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#import_error\n
-for some common reasons and solutions.  Include the entire stack trace
+Visit https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/install/index.md\n
+and click on the link that corresponds to your operating system to read about\n
```
caisq: There is no need for this line break here, because it is the middle of a sentence. With this line break, the sentence would be broken into two lines separated by an empty line.
Same below.

caisq (changes requested):

tensorflow/python/pywrap_tensorflow.py
caisq: Thanks for sending the fix, @rasbt . Please address the comment above, then we can merge this PR.

tensorflow/python/pywrap_tensorflow.py
```diff
@@ -46,8 +46,12 @@
     sys.setdlopenflags(_default_dlopen_flags)
 except ImportError:
   msg = """%s\n\nFailed to load the native TensorFlow runtime.\n
-See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#import_error\n
-for some common reasons and solutions.  Include the entire stack trace
+Visit https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/install/index.md\n
```
caisq: Use https://www.tensorflow.org/install/.

rasbt:
good call @caisq . I prob. had python's original pep-8 and users with 80-char terminals in mind, but yeah, it's 2017 :).

tensorflow/python/pywrap_tensorflow.py
```diff
@@ -46,8 +46,12 @@
     sys.setdlopenflags(_default_dlopen_flags)
 except ImportError:
   msg = """%s\n\nFailed to load the native TensorFlow runtime.\n
-See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#import_error\n
-for some common reasons and solutions.  Include the entire stack trace
+Visit https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/install/index.md
```
caisq: Can you update the URL to https://www.tensorflow.org/install/, which is prettier and pushed per release, so more stable?

caisq (changes requested):

tensorflow/python/pywrap_tensorflow.py
rasbt: sure, np, give me a sec

josh11b (approved):

caisq:

caisq (approved):

caisq:
PR merged. Thanks for the contribution, @rasbt !

PR #8079, opened 2017-03-04
Support read data from ViewFs.

wangyum:
Support read data from [ViewFs](http://hadoop.apache.org/docs/r2.6.5/hadoop-project-dist/hadoop-hdfs/ViewFs.html).

### DIFF
--- a/tensorflow/core/platform/hadoop/hadoop_file_system.cc
+++ b/tensorflow/core/platform/hadoop/hadoop_file_system.cc
@@ -58,6 +58,7 @@ class LibHDFS {
   std::function<hdfsFS(hdfsBuilder*)> hdfsBuilderConnect;
   std::function<hdfsBuilder*()> hdfsNewBuilder;
   std::function<void(hdfsBuilder*, const char*)> hdfsBuilderSetNameNode;
+  std::function<int(const char*, char**)> hdfsConfGetStr;
   std::function<void(hdfsBuilder*, const char* kerbTicketCachePath)>
       hdfsBuilderSetKerbTicketCachePath;
   std::function<int(hdfsFS, hdfsFile)> hdfsCloseFile;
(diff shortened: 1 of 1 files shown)


wangyum:
@gunan @jhseu Our Hadoop cluster need this feature. thanks.

jhseu:
How is the viewfs cluster name being picked up? The change seems to ignore that component.

jhseu:
For example, how would you distinguish between:
`viewfs://clusterx/path/to/foo` vs. `viewfs://clustery/path/to/foo`

tensorflow/core/platform/hadoop/hadoop_file_system.cc
```diff
@@ -147,6 +147,10 @@ Status HadoopFileSystem::Connect(StringPiece fname, hdfsFS* fs) {
   hdfsBuilder* builder = hdfs_->hdfsNewBuilder();
   if (scheme == "file") {
     hdfs_->hdfsBuilderSetNameNode(builder, nullptr);
+  } else if (scheme == "viewfs") {
+    // The default NameNode configuration will be used (from the XML configuration files). See:
+    // https://github.com/tensorflow/tensorflow/blob/v1.0.0/third_party/hadoop/hdfs.h#L259
+    hdfs_->hdfsBuilderSetNameNode(builder, "default");
```
jhseu: Right, but that only works if `fs.defaultFS` is set, right? There's no way to refer to another cluster here, and it'll always use the default one.

Suppose `fs.defaultFS` is set to `viewfs://cluster4`. If someone refers to a file: `viewfs:///foo/bar`, that'll point to `viewfs://cluster4/foo/bar`. With this change, if someone uses `viewfs://cluster5/foo/bar`, that'll also point to `viewfs://cluster4/foo/bar` because cluster4 is the defaultFS.

I couldn't figure out how to pass the viewfs cluster to the HDFS C API. If we can't, how about checking that nn is an empty string, otherwise return an error? That way viewfs is only supported as a defaultFS.

jhseu (approved):

jhseu:
Jenkins, test this please

PR #8106, opened 2017-03-05
Fix a bug of initializing

Jiaming-Liu:
Prevent the exceptional illegal call of __initializer.\_\_init\_\_(shape, dtype, partition_info)__ appears when the param "initializer" is an init_ops class instead of its instance.

### DIFF
--- a/tensorflow/python/ops/variable_scope.py
+++ b/tensorflow/python/ops/variable_scope.py
@@ -685,7 +685,7 @@ class _VariableStore(object):
       else:
         # Instantiate initializer if provided initializer is a type object.
         if isinstance(initializer, type(init_ops.Initializer)):
-          initializer=initializer(dtype)
+          initializer=initializer(dtype=dtype)
         init_val = lambda: initializer(  # pylint: disable=g-long-lambda
             shape.as_list(), dtype=dtype, partition_info=partition_info)
         variable_dtype = dtype.base_dtype
--- a/tensorflow/python/ops/variable_scope.py
+++ b/tensorflow/python/ops/variable_scope.py
@@ -683,6 +683,8 @@ class _VariableStore(object):
         init_val = initializer
         variable_dtype = None
       else:
+        if type(initializer)==type:
+          initializer=initializer(dtype)
         init_val = lambda: initializer(  # pylint: disable=g-long-lambda
             shape.as_list(), dtype=dtype, partition_info=partition_info)
         variable_dtype = dtype.base_dtype
--- a/tensorflow/python/ops/variable_scope.py
+++ b/tensorflow/python/ops/variable_scope.py
@@ -683,7 +683,8 @@ class _VariableStore(object):
         init_val = initializer
         variable_dtype = None
       else:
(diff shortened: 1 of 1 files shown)


Jiaming-Liu:
I signed it!

This bug is exposed when using tf/model/slim/models/slim/train_image_classifier.py to run vgg-16.

tensorflow/python/ops/variable_scope.py
```diff
@@ -683,6 +683,8 @@ def _get_single_variable(self,
         init_val = initializer
         variable_dtype = None
       else:
+        if type(initializer)==type:
```
lukaszkaiser: I don't understand what this line does, nor why it's needed. Please add an extensive comment explaining what's happening here.

lukaszkaiser (changes requested):
Please add a comment. It's also very strange to see "==type" anywhere in python -- did you mean to check if it's isinstance(..., some-class)?

tensorflow/python/ops/variable_scope.py
Jiaming-Liu: It is for preventing the exceptional illegal call of "initializer.\_\_init\_\_(shape, dtype, partition_info)" appears when the param "initializer" is an init_ops class instead of its instance.
For example, when I was using "tf/models/slim/train_image_classifier.py" to run VGG-16, the script actually passed the **type object** of class "tensorflow.python.ops.init_ops.Zeros" instead of an  **instance** of class Zeros. Hence, type(initializer) would resulted in <type 'type'>. 
This might be an out-dated usage or a minor exception. If so, this commit would not be necessary.

Jiaming-Liu: And the line actually means 

> isinstance(initializer, type(Initializer))

lukaszkaiser:
> And the line actually means
> isinstance(initializer, type(tensorflow.python.ops.init_ops.Initializer))

If that's what it means, then please code it like that. And add a clarifying comment in code so nobody has to trace it back to this discussion (in case someone wants to read or change this code even again).

tensorflow/python/ops/variable_scope.py
```diff
@@ -683,7 +683,8 @@ def _get_single_variable(self,
         init_val = initializer
         variable_dtype = None
       else:
-        if type(initializer)==type:
+        # Instantiate initializer if provided initializer is a type object.
+        if isinstance(initializer, type(init_ops.Initializer)):
           initializer=initializer(dtype)
```
lukaszkaiser: Thanks, I think it's much clearer now! Would it be possible to call with a named argument, i.e., initializer = initializer(dtype=dtype)?

lukaszkaiser (changes requested):
One more small thing, looks good otherwise. Thanks!

lukaszkaiser (approved):
Thanks!

lukaszkaiser:

lukaszkaiser:
Guys, can we get this in?

dandelionmane:

PR #8115, opened 2017-03-06
Use vector constructor to initialize Array2D.

conqerAtapple:
Current constructor of Array2D calls resize to initialize the inner vector of
  Array2D and follows up with a Fill call. This can be replaced with a simple
  call to vector constructor. Vector constructor might also be more efficient.


conqer:
@hawkinsp please review.

hawkinsp (approved):
Thanks!

conqer:
@hawkinsp Thanks for the review!

conqer:
@hawkinsp will create a PR for `Array3d` to address the same issue.

dandelionmane:

PR #8125, opened 2017-03-06
Fix broken links to docs_src/get_started/get_started.md

kswapd:
Fix broken links to tensorflow/tensorflow/docs_src/get_started/get_started.md

### DIFF
--- a/tensorflow/docs_src/get_started/get_started.md
+++ b/tensorflow/docs_src/get_started/get_started.md
@@ -2,7 +2,7 @@
 # Getting Started With TensorFlow
 
 This guide gets you started programming in TensorFlow. Before using this guide,
-@{$install$install TensorFlow}. To get the most out of
+see [Install TensorFlow](../install/index.md). To get the most out of
 this guide, you should know the following:
 
 *   How to program in Python.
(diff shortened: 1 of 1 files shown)


kswapd:
I signed it!

dandelionmane:
Can you repro that the links are broken? They seem to be working on tensorflow.org. If there is a bug, we would want to fix our custom link generator rather than short-circuiting it and using raw markdown.

kswapd:
I don't know why I cannot access tensorflow.org, so I can only learn TensorFlow from github, Is there anyway that the links work well in both link generator and markdown environment?

dandelionmane:
It sounds like the root issue to debug is that you can't access tensorflow.org. I don't think we're aiming to maintain the doc links in pure-markdown, but I'll let @josh11b comment on that.

josh11b:
I agree with @dandelionmane , our strategy is described in https://github.com/tensorflow/tensorflow/tree/master/tensorflow/g3doc

Maybe @martinwicke can offer suggestions for accessing these docs without tensorflow.org (a case we would like to support, for e.g. people who know they will be offline).

josh11b:
Actually, you can follow the instructions in that README, to generate an offline copy of the docs with the links fixed.

dandelionmane:
Closing this PR since the site is working as intended.

kswapd:
I got it, thank you.

PR #7688, opened 2017-02-20
[Java] [Feature] Load from SavedModel

EronWright:
Added `SavedModelBundle` class to represent a loaded saved model.

Closes #7134

PR #8148, opened 2017-03-07
[Java] Ability to load from SavedModel

### LINKED ISSUE #7134: [Java] [Feature] Load from SavedModel
Make it possible for the Java library to load exported TF models in SavedModel format.   

Outline of the proposed design:
1. Define a new class `SavedModelBundle` to represent the loaded saved model.  Properties include the graph, session, and metagraphdef.
2. Define a static method on `SavedModelBundle` to load from disk, given an export path and some tags.

```
class SavedModelBundle implements AutoCloseable { 
  public Graph graph() { ... }
  public Session session() { ... }
  public byte[] metaGraphDef() { ... }
  public void close() { /* close the session and graph */ }
  public static SavedModelBundle loadSavedModel(String exportDir, Set<String> tags) { ... }
}
```

3. Write a JNI method to support loading, based on `TF_LoadSessionFromSavedModel` from the C API.

asimshankar:
This PR is exactly the same as #7688 with updates to the BUILD file to make tests pass so that we can merge the changes in.

Closes #7134 and #7688 

@EronWright : Not sure if you saw updates to #7688 asking for tweaks to the PR to make the tests pass. If you want to update that PR instead, I'm happy to abandon this one. Otherwise, will merge this one in and abandon the other.

### DIFF
--- a/tensorflow/java/BUILD
+++ b/tensorflow/java/BUILD
@@ -55,6 +55,19 @@ java_test(
 )
 
 java_test(
+    name = "SavedModelBundleTest",
+    size = "small",
+    srcs = ["src/test/java/org/tensorflow/SavedModelBundleTest.java"],
+    test_class = "org.tensorflow.SavedModelBundleTest",
+    data = ["//tensorflow/cc/saved_model:saved_model_half_plus_two"],
+    deps = [
+        ":tensorflow",
+        ":testutil",
+        "//external:junit",
+    ],
+)
+
+java_test(
     name = "SessionTest",
     size = "small",
     srcs = ["src/test/java/org/tensorflow/SessionTest.java"],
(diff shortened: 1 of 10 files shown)


asimshankar:
Setting "cla:yes" since the commits authored by @EronWright are exactly the same as in PR #7688 which he proposed.

jhseu (approved):

jhseu:
Jenkins, test this please

EronWright:
Thanks @asimshankar for the tweaks and for driving this through.

jhseu:
Ah, I probably shouldn't have squashed so we could preserve the correct `git blame` history.

EronWright:
@jhseu yes that saddens me a little but don't worry about it

PR #8163, opened 2017-03-07
Go: Handle nil values for list-valued attributes gracefully.

asimshankar:
Without this change, setting a list-valued attribute to 'nil' would
result in a crash (&list[0] is malformed).


jhseu (approved):

jhseu:
Jenkins, test this please

Issue #8165, opened 2017-03-07
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md 404


CanoeFZH:
rt, 404 error of link from readme, Download and Setup

prb12:
Should be fixed by https://github.com/tensorflow/tensorflow/pull/8169

PR #8166, opened 2017-03-07
Fix broken link due to recent docs move

chrisvarns:
The link was broken when the docs moved from g3doc to their new place. This fixes by linking to the "Installing TensorFlow" page on the website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ tracking requests and bugs, but please see
 and discussion.**
 
 ## Installation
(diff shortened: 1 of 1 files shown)


MicaelCarvalho:
There are other places in the repository with the broken URL. I didn't see this PR before submitting another one, otherwise I would post a comment here instead. Anyway, it's basically the same thing, but fixing other places as well: #8165

dandelionmane:
Closing this as it is redundant with https://github.com/tensorflow/tensorflow/pull/8169/files which also fixes some other files.

PR #8169, opened 2017-03-07
Closes #8165 -- Removing traces of os_setup.md

### LINKED ISSUE #8165: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md 404

MicaelCarvalho:


caisq:

caisq (approved):

caisq:
@josh11b, i'm not entirely sure whether it makes more sense to link to the website or files under docs_src. Adding you as a reviewer.

dandelionmane:
I don't think the link should point to tensorflow.org directly - I think it should use a relative path. @josh11b can you comment?

MicaelCarvalho:
I agree it should have relative links. Hold on, I'm updating files and will let you know as soon as it's OK (the last commit has a broken link).

josh11b:
@dandelionmane We don't want links to docs_src -- they are source files and are not necessarily readable.  tensorflow.org is the source of truth for docs and should be the target of links.  The main exception is if you are modifying something in docs_src or in a comment that will end up in the API docs, then you should use the new @{...} syntax (described in https://github.com/tensorflow/tensorflow/tree/master/tensorflow/g3doc ) for links when possible.

MicaelCarvalho:
@josh11b I believe anyone seeing the README file would have a Markdown interpreter, no? Unless they're reading in plain text, then it's safe to assume they will be able to follow links by themselves.

Relative links will alow users who have an offline copy to continue working, reading their copies of the docs, instead of going online to find the same resource.

josh11b:
The files in docs_src/ will have lots of problems being rendered by a Markdown interpreter, since we have our own syntax (@{...}) for internal links that requires a preprocessing step to be made useful.

MicaelCarvalho:
I see... On a side note, even after this PR there are still 38 references in 16 files to g3doc, they should eventually be fixed as well.

Should I retreat the last commit with relative links?

josh11b:
Yes please, retreat the last commit

MicaelCarvalho:
@josh11b done. ;)

dandelionmane:
Cool, merging it then. Thanks for the fix.

Issue #8134, opened 2017-03-06
Download of test data not working


CmasterG:
The get started part has a section MNIST for ML Beginners. In this tutorial you should download the test data via:

from tensorflow.examples.tutorials.mnist import input_data
mnist = input_data.read_data_sets("MNIST_data/", one_hot=True)

Unfortunately that's not working because of an connection error.

prb12:
The MNIST example downloads its dataset from `http://yann.lecun.com/exdb/mnist`.
If this site is temporarily down or unreachable then the download will fail. 

Please can you try accessing the URL directly in a web browser?  (it is currently unreachable for me) 
Once the page is accessible in a browser, then please try the tutorial again.

CmasterG:
Hi!
Thanks for the fast reply. No, it is not reachable in the browser. Is there an alternative way to get the relevant data?

prb12:
Duplicate of #8126

amita-kapoor:
I am getting " HTTP Error 403: Forbidden" while trying to download MNIST dataset using ```input_data.read_data_sets("MNIST_data/", one_hot=True)``` comand, though the data is downloadable from Yann LeCun website. 

Tensorflow version 1.4

Issue #8126, opened 2017-03-06
yann.lecun.com/exdb/mnist/ down


jegli:
The website of Yann LeCun is down at the moment. 
Therefore the MNIST-Script for downloading the files doesn't work either. (-> learn/python/learn/datasets/mnist.py won't work)
Are there any mirrors?

prb12:
Apparently (according to @kakawait) you can download dataset via archive machine, e.g. https://web.archive.org/web/20160828233817/http://yann.lecun.com/exdb/mnist/index.html 

You will need to download and copy the files to your training data directory manually.

Issue #8116, opened 2017-03-06
Errno 110 occured when running all python files in /examples/tutorials/mnist/ directory.


yw346090249:
davidtest@CaffeVM:~/tensorflow/tensorflow/tensorflow/examples/tutorials/mnist$ python fully_connected_feed.py 
Traceback (most recent call last):
  File "fully_connected_feed.py", line 277, in <module>
    tf.app.run(main=main, argv=[sys.argv[0]] + unparsed)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 44, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "fully_connected_feed.py", line 222, in main
    run_training()
  File "fully_connected_feed.py", line 120, in run_training
    data_sets = input_data.read_data_sets(FLAGS.input_data_dir, FLAGS.fake_data)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/datasets/mnist.py", line 211, in read_data_sets
    SOURCE_URL + TRAIN_IMAGES)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/datasets/base.py", line 208, in maybe_download
    temp_file_name, _ = urlretrieve_with_retry(source_url)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/datasets/base.py", line 165, in wrapped_fn
    return fn(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/datasets/base.py", line 190, in urlretrieve_with_retry
    return urllib.request.urlretrieve(url, filename)
  File "/usr/lib/python2.7/urllib.py", line 94, in urlretrieve
    return _urlopener.retrieve(url, filename, reporthook, data)
  File "/usr/lib/python2.7/urllib.py", line 240, in retrieve
    fp = self.open(url, data)
  File "/usr/lib/python2.7/urllib.py", line 208, in open
    return getattr(self, name)(url)
  File "/usr/lib/python2.7/urllib.py", line 345, in open_http
    h.endheaders(data)
  File "/usr/lib/python2.7/httplib.py", line 1013, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python2.7/httplib.py", line 864, in _send_output
    self.send(msg)
[...]

Issue #6742, opened 2017-01-09
`tensorflow.examples.tutorials.mnist` not currently downloadable


asross:
### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)

```python
from tensorflow.examples.tutorials.mnist import input_data
mnist = input_data.read_data_sets("/tmp/data/", one_hot=True)
```

raises `http.client.RemoteDisconnected: Remote end closed connection without response`, if you don't already have the data cached. I believe this is because the underlying data [SOURCE_URL](http://yann.lecun.com/exdb/mnist/) is currently [down](http://downforeveryoneorjustme.com/http://yann.lecun.com/exdb/mnist/).

asross:
...It's back up again now. Looks like it was just a temporary disruption.

joethompson1961:
I encountered the same problem just now.  how do I fix it?

asross:
> Using the same code:
> 
> from tensorflow.examples.tutorials.mnist import input_data
> mnist = input_data.read_data_sets("/tmp/data/", one_hot=True)
> 
> raises a timeout error: TimeoutError: [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
> 
> What would cause this?

The issue is that the website where MNIST is canonically hosted sometimes goes down. I think an ideal solution would be to find or set up a mirror, then fall back to it? Or just find a more stable source.

gdfishhannah:
I also have the 10060 problem and it always happened. This seems like not the host problem. Do you know what caused this?

asross:
Reopening because others seem to have this problem.

asross:
I would suggest that if you encounter this problem, check http://downforeveryoneorjustme.com/http://yann.lecun.com/exdb/mnist/ just to make sure.

gdfishhannah:
I cannot reach this website

awilliamson:
Confirmed. I'm also having MNIST issues. Is there any alternative URL we can use for this which will still be compatible with existing MNIST tutorials. I'm relying on these as baselines for experiments, I'd rather keep the data and TF format the same.

kakawait:
Still down seems few days... Other related issue #8116

auserdude:
Same project here, solved by downloading file from a different source. I think we should put in place a mirror somewhere.

jeffheaton:
Also cannot access.

prb12:
[...]

PR #8171, opened 2017-03-07
add backup URL for downloading MNIST

asross:
A lot of folks (https://github.com/tensorflow/tensorflow/issues/6742, https://github.com/tensorflow/tensorflow/issues/8126, https://github.com/tensorflow/tensorflow/issues/8134,
https://github.com/tensorflow/tensorflow/issues/8116) have been having sporadic trouble downloading MNIST, because the site where it is canonically hosted sometimes goes down. This PR just adds a backup URL for convenience (though with the increased traffic the backup will get when normal MNIST is down, the backup might fail too. Someone might want to put this up on S3 or something Google-hosted.)

### DIFF
--- a/tensorflow/contrib/learn/python/learn/datasets/mnist.py
+++ b/tensorflow/contrib/learn/python/learn/datasets/mnist.py
@@ -28,7 +28,7 @@ from tensorflow.contrib.learn.python.learn.datasets import base
 from tensorflow.python.framework import dtypes
 
 SOURCE_URL = 'http://yann.lecun.com/exdb/mnist/'
-
+BACKUP_URL = 'https://web.archive.org/web/20160828233817/http://yann.lecun.com/exdb/mnist/'
 
 def _read32(bytestream):
   dt = numpy.dtype(numpy.uint32).newbyteorder('>')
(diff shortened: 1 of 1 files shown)


asross:
I signed it!

tensorflow/contrib/learn/python/learn/datasets/mnist.py
```diff
@@ -28,7 +28,7 @@
 from tensorflow.python.framework import dtypes
 
 SOURCE_URL = 'http://yann.lecun.com/exdb/mnist/'
-
+BACKUP_URL = 'https://web.archive.org/web/20160828233817/http://yann.lecun.com/exdb/mnist/'
```
caisq: Instead of this, can we use a more general approach, namely a list of URLs, such as 

```
SOURCE_URLS = ['http://yann.lecun.com/exdb/mnist/', 'https://web.archive.org/web/20160828233817/http://yann.lecun.com/exdb/mnist/']
```

and let the code try them one by one until the first successful one? If all failed, throw an exception.

tensorflow/contrib/learn/python/learn/datasets/mnist.py
```diff
@@ -223,23 +223,25 @@ def fake():
   TEST_IMAGES = 't10k-images-idx3-ubyte.gz'
   TEST_LABELS = 't10k-labels-idx1-ubyte.gz'
 
-  local_file = base.maybe_download(TRAIN_IMAGES, train_dir,
-                                   SOURCE_URL + TRAIN_IMAGES)
+  def download_local_file(fname):
+    try:
+      return base.maybe_download(fname, train_dir, SOURCE_URL + fname)
```
caisq: The retry logic should be put into maybe_download() to avoid bloating code like here.

caisq (changes requested):

tensorflow/contrib/learn/python/learn/datasets/mnist.py
asross: Hmm, `maybe_download` is defined in a different file that I think is used by many other datasets, and its current method signature doesn't really allow for the concept of a backup URL. Would you suggest an optional `backup` keyword argument?

tensorflow/contrib/learn/python/learn/datasets/mnist.py
asross: Sure, that would be a simple modification. If anyone has a better canonical backup URL than `web.archive.com`, feel free to suggest it, too. I don't want that site to get overwhelmed...

tensorflow/contrib/learn/python/learn/datasets/mnist.py
asross: I guess if we're expanding to a list of backup URLs, I feel like this change would require significant changes to other files I haven't looked at!

dandelionmane: We could make maybe_download take a filename, a directory, and either a URL or a list of URLs. if it gets a string (single url) behavior is same as present. if it gets an array of strings, then it tries them in sequence until successful. that way we can generalize the concept of multiple sources across the codebase in a backwards compatible way.

asross: Cool, that sounds like a great way of resolving the issue. I am a bit busy for the rest of today but I will try to take a crack at it in the next day or so (unless someone else wants to jump in).

gunan:
We do not have the permission from Yann Lecun to mirror these files.
I am guessing you also do not have either.

Therefore, we cannot accept this PR. We are looking into modifying our tutorials to not depend on mnist dataset.

asross:
Cool, no problem. Maybe someone can contact him and see if he can improve how MNIST is hosted, since it seems like tons and tons of people are downloading it on a daily basis now.

erolrecep:
I found a basic solution. I manually downloaded all dataset from http://yann.lecun.com/exdb/mnist/ to my local and made them reachable by tensorflow. And it's working. I know it's a dummy solution but it's working :)

zaikuzailei123:
It works by manually downloading all dataset. Save them in the format of *.gz in the MINST folder

augch:
which folder to should I download .gz files?

yurap:
MNIST_DATA, next to the corresponding python script
it should be created by previous failed calls

sizbrother:
Was this issue solved?

sizbrother:
I tried doing that, and I now get a new error:
ValueError: Cannot feed value of shape (100, 10) for Tensor u'Placeholder_1:0', which has shape '(?,)'

sizbrother:
the folder MNIST_data was created, and I added the data into the folder; yet I still get the error.

PR #8200, opened 2017-03-08
Add cmake build for benchmark_model

snnn:


tensorflow/tools/benchmark/benchmark_model.cc
```diff
@@ -272,7 +272,11 @@ Status TimeMultipleRuns(double sleep_seconds, int num_runs,
     // This can be helpful to determine the effect of mobile processor
     // scaling and thermal throttling.
     if (sleep_seconds > 0.0) {
+#ifdef _WIN32
```
petewarden: Minor nit, but can you add an include to [tensorflow/core/platform/platform.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/platform.h) and use the PLATFORM_WINDOWS macro instead? That is a bit more idiomatic.

petewarden (commented):
Thanks, this is looking great overall! Just one minor comment.

tensorflow/tools/benchmark/benchmark_model.cc
snnn: ok.

snnn: Done.

petewarden (approved):
Thanks for the update, looks good!

caisq:

caisq:

Issue #6379, opened 2016-12-17
Inconsistency in parameter naming


nmiculinic:
```
tf.nn.conv1d(value, filters, stride, padding, use_cudnn_on_gpu=None, data_format=None, name=None)
tf.nn.conv2d(input, filter, strides, padding, use_cudnn_on_gpu=None, data_format=None, name=None)
tf.nn.conv3d(input, filter, strides, padding, name=None)
```

There's discrepancy between parameter naming between conv1d and 2,3d counterparts. Firstly value vs input, and then singular vs plural filter(s) (singular would be correct as in 2d/3s case since function accepts only single filter if I understood documentation correctly.)

vrv:
Thanks for pointing out the inconsistency.  To fix it, we'll probably have to support both for a while but at least make the correct ones documented.

There's a generic 'convolution' parameter here that is more agnostic about the dimensionality, for what it's worth: https://github.com/tensorflow/tensorflow/blob/a276e0999ab4223ac36d75221028d3e8835c60ae/tensorflow/python/ops/nn_ops.py#L472

wt-huang:
Closing as it is resolved

PR #7883, opened 2017-02-25
Changed inconsistent conv1d docs to use 'input' param instead of 'value'

teldridge11:
Fixed inconsistency in documentation for conv1d.  Partial fix for issue #6379

PR #8203, opened 2017-03-08
Changed inconsistent conv1d to use 'filter' param instead of 'filters'

JulienSiems:
Updated the docs and all occurrences of 'filters' when used as input to conv1d. Partial fix for #6379.

### DIFF
--- a/tensorflow/python/ops/nn_ops.py
+++ b/tensorflow/python/ops/nn_ops.py
@@ -27,6 +27,7 @@ from tensorflow.python.framework import graph_util
 from tensorflow.python.framework import ops
 from tensorflow.python.framework import tensor_shape
 from tensorflow.python.framework import tensor_util
+from tensorflow.python.util.deprecation import deprecated_args
 from tensorflow.python.ops import array_ops
 from tensorflow.python.ops import gen_nn_ops
 from tensorflow.python.ops import math_ops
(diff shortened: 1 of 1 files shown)


dandelionmane:
@martinwicke This is technically a backwards-incompatible API change (b.c. ppl may use keyword params) and therefore inadmissible, yes?

martinwicke:
Indeed, that is sad. We can accept a second parameter called filter, which
takes filters' spot, and we can deprecate filters, which would move to the
end. That would be backward compatible. This change we cannot make because
it may well break current users.

JulienSiems:
I can change it to be a second parameter. Should there be a deprecation message when using filters?
Why was is it ok to do this in #7883 ?  Didn't that break the API as well?

terrytangyuan:
@JulienSiems The PR you referred was only documentation change. The first argument name change is not as big as this one. A longer warning period before removal would be more appropriate, e.g. remove when it gets to next major release.

JulienSiems:
Should I close this pull request? If it's better solved for the next major release?

dandelionmane:
You could close the PR, or make the change in a backwards-compatible way as martin suggested.

JulienSiems:
I used the deprecated_args decorator to deprecate filters. Is that the way you want to be done?

PR #8204, opened 2017-03-08
CMake: use copy_if_different to replace copy

snnn:
For speedup edit-compile-run loop


mrry (approved):
Thanks for this change! Looks good to me... we'll just have to run the tests.

mrry:

PR #8218, opened 2017-03-08
Use .bazelrc to configure optional dependencies

### LINKED ISSUE #8202: build_config.bzl file modified by configure
`configure` modifies `build_config.bzl` to set the `WITH_JEMALLOC` flag.  This is inconvenient when using git, since it means I can't pull after doing configure without manually clearing the file.

Can the setup be modified so that `configure` makes a file not under git control?

jart:
Rather than having ./configure mutate one of our .bzl files, which dirties the git repository, this change has ./configure put --define options in .bazelrc for jemalloc, GCS, HDFS, and XLA inside a .bazelrc at the root of the TensorFlow repository. This file is listed in .gitignore.

Therefore, running ./configure will no longer cause the git repository to be in a modified state.

Fixes #8202 
CC: @girving


gunan:
I like how the change looks. I will wait on the approval until Damien and Martin takes a look, and raise any concerns if there is any.

configure
```diff
@@ -180,13 +183,12 @@ else
...
-  sed_hyphen_i -e "s/WITH_JEMALLOC = False/WITH_JEMALLOC = True/" tensorflow/core/platform/default/build_config.bzl
-else
-  sed_hyphen_i -e "s/WITH_JEMALLOC = True/WITH_JEMALLOC = False/" tensorflow/core/platform/default/build_config.bzl
+sed_hyphen_i -e "/with_jemalloc/d" .bazelrc
+if [[ "$TF_NEED_JEMALLOC" == "1" ]]; then
+  echo 'build --define with_jemalloc=true' >>.bazelrc
 fi
 
-while [ "$TF_NEED_GCP" == "" ]; do
+while [[ "$TF_NEED_GCP" == "" ]]; do
```
girving: Why the `[ ... ] -> [[ ... ]]` changes?  Was it wrong before?

jart: `[[` is bash builtin syntax whereas `[` is a literal command line program that needs to be spawned.

martinwicke: Yes, this is unnecessary. Basic `test` is enough to evaluate `==`. That said, it's never wrong to use `[[...]]`.

martinwicke:
Shiny! The only thing left now is python config, right?

damienmg (approved):
w00h00 +1000 for putting those in .bazelrc

damienmg:
After re-reading the change: all those define will have to go in an external repository to make the import TF as a workspace works nicely, unfortunately there is no vardef so you probably wants to write a bzl file with those variable defined with an external repository (either as a follow-up or as part of this change).

dandelionmane:
@jart Do you want to merge this, or add in the change damien suggested?

jhseu:
Jenkins, test this please

jart:
@dandelionmane I wrote this change under the assumption that dependent projects might choose to write their own ./configure scripts, that define the .bazelrc in their root directory.

@damienmg How robust can we make Bazel's `.bazelrc` discovery? Assume we have three .bazelrc files:

- ~/.bazelrc
- ~/tf_serving/.bazelrc
- ~/tf_serving/tensorflow/.bazelrc

If we build something inside ~/tf_serving/tensorflow/... then should all three bazelrc's be loaded? And if so, having the latter ones take priority?

damienmg:
~/.bazelrc will be loaded when building tf_serving or tensorflow, .bazelrc
from tf_serving can import the .bazelrc from tensorflow it should just work.

There is no way to load ~/tf_serving/.bazelrc when building from
~/tf_serving/tensorflow.

As said previously, to make that use case works nicely, it would be better
to write a remote repository that write those configuration in a BUILD file
/ skylark file.

On Wed, Mar 15, 2017 at 1:59 AM Justine Tunney <notifications@github.com>
wrote:

@dandelionmane <https://github.com/dandelionmane> I wrote this change under
the assumption that dependent projects might choose to write their own
./configure scripts, that define the .bazelrc in their root directory.

@damienmg <https://github.com/damienmg> How robust can we make Bazel's
.bazelrc discovery? Assume we have three .bazelrc files:

   - ~/.bazelrc
   - ~/tf_serving/.bazelrc
   - ~/tf_serving/tensorflow/.bazelrc

If we build something inside ~/tf_serving/tensorflow/... then should all
three bazelrc's be loaded? And if so, having the latter ones take priority?

—


<https://github.com/tensorflow/tensorflow/pull/8218#issuecomment-286609051>,
or mute the thread

PR #8219, opened 2017-03-08
Make binary ops delegate to rhs when tensor conversion fails

### LINKED ISSUE #8051: Make tf.Tensor's magic functions (e.g., `__add__()`) play nice with unknown types
`tf.Tensor`'s magic functions should return `NotImplemented` rather than raising an exception when they are given an object of an unsupported type (i.e., an object that TensorFlow cannot `convert_to_tensor()`). This would give that object a chance to perform the operation with its reverse magic function (e.g., `__radd__()`). See the code example below. Note that this is what NumPy arrays do.

### Environment info
Operating System: **Ubuntu 16.10**
Installed version of CUDA and cuDNN:  **N/A**
Installed pip package:
https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.0.0-cp35-cp35m-linux_x86_64.whl
TensorFlow version: **1.0.0**

### Reproducible example

```
>>> class AmazingMatrix(object):
...   def __add__(self, b):
...     print("I know how to handle TensorFlow Tensors!")
...   def __radd__(self, a):
...     print("I know how to handle TensorFlow Tensors!")
... 
>>> am = AmazingMatrix()
>>> t = tf.constant([[0.]])
>>> am + t
I know how to handle TensorFlow Tensors!
>>> t + am
[...]/python/ops/math_ops.py", line 799, in binary_op_wrapper
    y = ops.convert_to_tensor(y, dtype=x.dtype.base_dtype, name="y")
[...]
TypeError: Expected float32, got <__main__.AmazingMatrix object at 0x7f5999a73518> of type 'AmazingMatrix' instead.
```

### Related GitHub issues or StackOverflow threads
http://stackoverflow.com/questions/38599637/overloading-tensorflow-tensor-magic-method-priority-similar-to-numpys-array

### What other attempted solutions have you tried?
None. I guess one could create Tensor wrappers, with the appropriate behavior, but it would be quite cumbersome to use.

None

ageron:
Fixes #8051.

The goal is to allow users to choose whether `tensor + something` should call `tensor`'s `__add__()` method or `something`'s `__radd__()` method.

Example usage:

```
>>> import tensorflow as tf
>>> tensor = tf.constant(3.0)
>>> class MagicMatrix(object):
...   def __radd__(self, lhs):
...     print("MagicMatrix.__radd__", lhs)
... 
>>> tensor + MagicMatrix()
MagicMatrix.__radd__ Tensor("Const:0", shape=(), dtype=float32)
```

If the RHS's type does not define the reverse method (in this example `__radd__()`) then the usual behavior applies:

```
>>> class MagicMatrix(object):
...   pass
... 
>>> tensor + MagicMatrix()
[...]
TypeError: Expected float32, got <__main__.MagicMatrix object at 0x7f14e32aa978> of type 'MagicMatrix' instead.
```

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -799,6 +799,9 @@ def _OverrideBinaryOperatorHelper(func, op_name, clazz_object=ops.Tensor):
         try:
           y = ops.convert_to_tensor(y, dtype=x.dtype.base_dtype, name="y")
         except TypeError:
+          # If the RHS is not a tensor, it might be a tensor aware object
+          # that can implement the operator with knowledge of itself
+          # and the tensor.
           if hasattr(type(y), "__r%s__" % op_name):
             return NotImplemented
           else:
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -796,7 +796,13 @@ def _OverrideBinaryOperatorHelper(func, op_name, clazz_object=ops.Tensor):
   def binary_op_wrapper(x, y):
     with ops.name_scope(None, op_name, [x, y]) as name:
       if not isinstance(y, sparse_tensor.SparseTensor):
-        y = ops.convert_to_tensor(y, dtype=x.dtype.base_dtype, name="y")
+        try:
+          y = ops.convert_to_tensor(y, dtype=x.dtype.base_dtype, name="y")
+        except TypeError:
+          if hasattr(type(y), "__r%s__" % op_name):
+            return NotImplemented
+          else:
+            raise
       return func(x, y, name=name)
 
   def binary_op_wrapper_sparse(sp_x, y):
--- a/tensorflow/python/kernel_tests/tensor_priority_test.py
+++ b/tensorflow/python/kernel_tests/tensor_priority_test.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"""Tests for the __tensor_priority__ mechanism."""
+"""Tests for the binary ops priority mechanism."""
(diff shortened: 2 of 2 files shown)


ageron:
I still need to add some tests and documentation. Is there an existing file I should add them to?

dandelionmane:
@tensorflow-jenkins TEst this please

martinwicke:
Jenkins, test this please.

martinwicke:
@ebrevdo @aselle ping for a review?

drpngx:
@ebrevdo @aselle any chance to look at this?

tensorflow/python/ops/math_ops.py
```diff
@@ -796,7 +796,13 @@ def _OverrideBinaryOperatorHelper(func, op_name, clazz_object=ops.Tensor):
   def binary_op_wrapper(x, y):
     with ops.name_scope(None, op_name, [x, y]) as name:
       if not isinstance(y, sparse_tensor.SparseTensor):
-        y = ops.convert_to_tensor(y, dtype=x.dtype.base_dtype, name="y")
+        try:
+          y = ops.convert_to_tensor(y, dtype=x.dtype.base_dtype, name="y")
+        except TypeError:
```
aselle: Please add a comment here about why this change has been done i.e.
```
# If the RHS is not a tensor, it might be a tensor aware object
# that can implement the operator with knowledge of itself
# and the tensor.
```

aselle (approved):
I am happy with this change. Sorry for the delay in reviewing.

tensorflow/python/ops/math_ops.py
martinwicke: done

martinwicke:
Jenkins, test this please.

PR #8232, opened 2017-03-09
Optimize Array initializer_list constructor.

conqerAtapple:
std::vector reserve + emplace could be more efficient than
  - resize + copy
      or
  - construct default + copy

### DIFF
--- a/tensorflow/compiler/xla/array2d.h
+++ b/tensorflow/compiler/xla/array2d.h
@@ -55,12 +55,12 @@ class Array2D {
   // initializer list is the first dimension; the inner is the second dimension.
   // For example, {{1, 2, 3}, {4, 5, 6}} results in an array with n1=2 and n2=3.
   Array2D(std::initializer_list<std::initializer_list<T>> values)
-      : Array2D(values.size(), values.begin()->size()) {
-    int64 n1 = 0;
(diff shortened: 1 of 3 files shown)


conqer:
@hawkinsp please review.

hawkinsp (commented):
I'm not convinced this is actually faster.

I would have assumed that default initialization followed by writing each entry was cheaper than calling reserve() followed by repeated emplace_back() calls, if nothing else because the emplace_back() must update the integer representing the size of the vector each time. (Perhaps the compiler is smart enough to optimize this.)

I would reconsider if there is a benchmark that proves me wrong...

I should also add: the speed of the various ArrayXD classes doesn't matter a whole lot, since they are mainly used for compile-time constants, and most Tensors flowing into and out of XLA aren't constant.

conqer:
@hawkinsp Thanks for the review. When we "reserve" a capacity for vector, we avoid creating the elements. On the other hand, the "resize" option or the "vector constructor + copy " option, the elements are created in the vector. So when we call the "[]"  operator to get a reference to the object at that index, we again copy the element to the already default constructed element.  So we have a creation + copy cost. In the reserve case, we

martinwicke:
I think there are two ways forward here: I can close this PR, or we can look at benchmarks. You can find other micro-benchmarks throughout the code, and writing one of those for this piece of code would be a nice thing to have anyway.

Let me know.

PR #8248, opened 2017-03-09
Fix: Join checked thread in random_shuffle_queue_test.

gunan:


gunan:
closing pr, fix sent internally.

PR #8249, opened 2017-03-09
use sed separator compatible with busybox sed

dimpavloff:
busybox's `sed` errors with '[' as separator. This becomes an issue when
trying to run `tensorflow/tools/ci_build/builds/configured` on Alpine
Linux.


dimpavloff:
An easy way to confirm:
```
$ docker run -it --rm busybox /bin/sh -c "echo abc | sed -e 's[abc[xyz[g'"
sed: unmatched '['
$ docker run -it --rm busybox /bin/sh -c "echo abc | sed -e 's|abc|xyz|g'"
xyz
```

dandelionmane:

teamdandelion (approved):

PR #8257, opened 2017-03-10
Fix docs for pool and convolution functions

nghiattran:
On https://www.tensorflow.org/api_docs/python/tf/nn/pool 

* 2 code blocks are not highlighted correctly because of semicolons come before them 
* Missing new line for padding explanation 

On https://www.tensorflow.org/api_docs/python/tf/nn/convolution

* 1 code block is not highlighted correctly because of semicolons come right before it.


teamdandelion (approved):

dandelionmane:

PR #8260, opened 2017-03-10
native Cuda kernel for tf.dynamic_stitch op (reopened PR)

MycChiu:
I screwed up when squashing commits in #7764, so I started a clean pull request. @girving Please take a look at this, thank you. *(and I sincerely apologize for my horrible git-fu)*

### DIFF
--- a/tensorflow/core/kernels/dynamic_stitch_op.cc
+++ b/tensorflow/core/kernels/dynamic_stitch_op.cc
@@ -20,9 +20,119 @@ limitations under the License.
 #include "tensorflow/core/framework/tensor.h"
 #include "tensorflow/core/kernels/bounds_check.h"
 
+#if GOOGLE_CUDA
+#include "tensorflow/core/kernels/cuda_device_array.h"
+#endif  // GOOGLE_CUDA
+
 namespace tensorflow {
 
-template <class T>
+typedef Eigen::ThreadPoolDevice CPUDevice;
+typedef Eigen::GpuDevice GPUDevice;
+
+template <typename Device, typename T>
+struct LaunchDynamicStitchOp;
+
+template <typename T>
+struct LaunchDynamicStitchOp<CPUDevice, T> {
+  static void launch(OpKernelContext* c, const int slice_size,
+                     const int first_dim_size, OpInputList indices_inputs,
+                     OpInputList data_inputs, Tensor* merged) {
+    auto merged_flat = merged->flat_outer_dims<T>();
+    for (int input_num = 0; input_num < indices_inputs.size(); input_num++) {
+      const Tensor& indices = indices_inputs[input_num];
+      auto indices_vec = indices.flat<int32>();
+      const Tensor& data = data_inputs[input_num];
+      auto data_flat =
+          data.shaped<T, 2>({indices_vec.dimension(0), slice_size});
+
+      if (DataTypeCanUseMemcpy(DataTypeToEnum<T>::v())) {
+        T* merged_base = &merged_flat(0, 0);
+        const T* data_base = &data_flat(0, 0);
+        const size_t slice_bytes = slice_size * sizeof(T);
+        for (int i = 0; i < indices_vec.size(); i++) {
+          int32 index = internal::SubtleMustCopy(indices_vec(i));
+          OP_REQUIRES(
+              c, FastBoundsCheck(index, first_dim_size),
+              errors::InvalidArgument("indices[", i, "] is out of range"));
+          memcpy(merged_base + index * slice_size, data_base + i * slice_size,
+                 slice_bytes);
+        }
+      } else {
+        Eigen::DSizes<Eigen::DenseIndex, 2> sizes(1, slice_size);
+        for (int i = 0; i < indices_vec.size(); i++) {
+          // Copy slice data[i] to merged[indices[i]]
+          Eigen::DSizes<Eigen::DenseIndex, 2> data_indices(i, 0);
+          int32 index = internal::SubtleMustCopy(indices_vec(i));
+          OP_REQUIRES(
+              c, FastBoundsCheck(index, first_dim_size),
+              errors::InvalidArgument("indices[", i, "] is out of range"));
+          Eigen::DSizes<Eigen::DenseIndex, 2> merged_indices(index, 0);
+          merged_flat.slice(merged_indices, sizes) =
+              data_flat.slice(data_indices, sizes);
+        }
+      }
+    }
+  }
+};
+
+#if GOOGLE_CUDA
+
+template <typename T>
+struct DynamicStitchGPULaunch {
+  static void Run(const GPUDevice& d, const int slice_size,
+                  const int first_dim_size,
+                  const CudaDeviceArrayStruct<int32>& indices_flat,
+                  const CudaDeviceArrayStruct<const T*>& data_slice_ptrs,
+                  T* output);
+};
+
+template <typename T>
+struct LaunchDynamicStitchOp<GPUDevice, T> {
+  static void launch(OpKernelContext* c, const int slice_size,
+                     const int first_dim_size, OpInputList indices_inputs,
+                     OpInputList data_inputs, Tensor* merged) {
+    // createt two arrays that will be sent to CUDA device
+    // one used as pointers to output's row id
+    CudaDeviceArrayOnHost<int32> indices_flat(c, first_dim_size);
+    // another used as pointers to the pointers of the head of each data_slice
+    CudaDeviceArrayOnHost<const T*> data_slice_ptrs(c, first_dim_size);
+    OP_REQUIRES_OK(c, indices_flat.Init());
+    OP_REQUIRES_OK(c, data_slice_ptrs.Init());
+
+    int ptr_num = 0;
+    for (int input_num = 0; input_num < indices_inputs.size(); input_num++) {
+      auto indices_vec = indices_inputs[input_num].flat<int32>();
+      auto base_ptr = data_inputs[input_num].template flat<T>().data();
+      for (int ind_num = 0; ind_num < indices_vec.size(); ind_num++) {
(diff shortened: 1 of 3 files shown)


tensorflow/core/kernels/dynamic_stitch_op.cc
```diff
@@ -95,46 +205,13 @@ class DynamicStitchOp : public OpKernel {
     }
     Tensor* merged = nullptr;
     OP_REQUIRES_OK(c, c->allocate_output(0, result_shape, &merged));
-
     // TODO(jeff): Currently we leave uninitialized any portions of
     // merged that aren't covered by an index in indices.  What should we do?
     if (first_dim_size > 0) {
```
girving: Can't this be `result_shape.num_elements() > 0` to exit early if other dimensions are zero too?

tensorflow/core/kernels/dynamic_stitch_op.cc
```diff
@@ -20,9 +20,119 @@ limitations under the License.
...
+  static void launch(OpKernelContext* c, const int slice_size,
+                     const int first_dim_size, OpInputList indices_inputs,
+                     OpInputList data_inputs, Tensor* merged) {
+    // createt two arrays that will be sent to CUDA device
+    // one used as pointers to output's row id
+    CudaDeviceArrayOnHost<int32> indices_flat(c, first_dim_size);
+    // another used as pointers to the pointers of the head of each data_slice
+    CudaDeviceArrayOnHost<const T*> data_slice_ptrs(c, first_dim_size);
+    OP_REQUIRES_OK(c, indices_flat.Init());
+    OP_REQUIRES_OK(c, data_slice_ptrs.Init());
```
girving: Presumably these get zero initialized, but I can't find where.  Can you confirm?

tensorflow/core/kernels/dynamic_stitch_op.cc
```diff
@@ -20,9 +20,119 @@ limitations under the License.
...
+
+    int ptr_num = 0;
+    for (int input_num = 0; input_num < indices_inputs.size(); input_num++) {
+      auto indices_vec = indices_inputs[input_num].flat<int32>();
+      auto base_ptr = data_inputs[input_num].template flat<T>().data();
+      for (int ind_num = 0; ind_num < indices_vec.size(); ind_num++) {
+        indices_flat.Set(ptr_num, indices_vec(ind_num));
```
girving: Needs the same error checking as in the CPU case.

tensorflow/core/kernels/dynamic_stitch_op_gpu.cu.cc
```diff
@@ -0,0 +1,68 @@
...
+
+#include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/framework/tensor_types.h"
+#include "tensorflow/core/kernels/cuda_device_array_gpu.h"
+#include "tensorflow/core/util/cuda_kernel_helper.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
+namespace tensorflow {
+typedef Eigen::GpuDevice GPUDevice;
+namespace {
```
girving: Newline before `namespace {`.

tensorflow/core/kernels/dynamic_stitch_op_gpu.cu.cc
```diff
@@ -0,0 +1,68 @@
...
+  int32* output_indices = GetCudaDeviceArrayOnDevice(&indices_flat);
+  CUDA_1D_KERNEL_LOOP(thread_id, output_size) {
+    const int slice_id = thread_id / slice_size;
+    const int slice_offset = thread_id % slice_size;
+    int32 output_id = output_indices[slice_id] * slice_size + slice_offset;
+    output[output_id] = ldg(data_ptrs[slice_id] + slice_offset);
+  }
+}
+}  // namespace
+template <typename T>
```
girving: Newline after `}  // namespace`.

tensorflow/core/kernels/dynamic_stitch_op.cc
girving: Also needs to use `SubtleMustCopy` appropriately.

girving: Wait, how is this correct at all?  What guarantee is there that `ptr_num < first_dim_size`?

tensorflow/core/kernels/dynamic_stitch_op_gpu.cu.cc
```diff
@@ -0,0 +1,68 @@
...
+    const int slice_size, const int output_size,
+    CudaDeviceArrayStruct<int32> indices_flat,
+    CudaDeviceArrayStruct<const T*> data_slice_ptrs, T* output) {
+  const T** data_ptrs = GetCudaDeviceArrayOnDevice(&data_slice_ptrs);
+  int32* output_indices = GetCudaDeviceArrayOnDevice(&indices_flat);
+  CUDA_1D_KERNEL_LOOP(thread_id, output_size) {
+    const int slice_id = thread_id / slice_size;
+    const int slice_offset = thread_id % slice_size;
+    int32 output_id = output_indices[slice_id] * slice_size + slice_offset;
+    output[output_id] = ldg(data_ptrs[slice_id] + slice_offset);
```
girving: Should you use `ldg` for `output_indices[slice_id]` and `data_ptrs[slice_id]`?

tensorflow/core/kernels/dynamic_stitch_op_gpu.cu.cc
```diff
@@ -0,0 +1,68 @@
...
+  static void Run(const GPUDevice& d, const int slice_size,
+                  const int first_dim_size,
+                  const CudaDeviceArrayStruct<int32>& indices_flat,
+                  const CudaDeviceArrayStruct<const T*>& data_slice_ptrs,
+                  T* output) {
+    // TODO(MycChiu): Currently these launch settings are hard-coded and
+    // roughly optimized for my local system, is there a better way to
+    // automatically determine the optimal launch settings?
+    const int block_size = 256;
+    const int grid_size = 60;
```
girving: @zheng-xq What's the right thing to do here?

tensorflow/core/kernels/dynamic_stitch_op.cc
```diff
@@ -20,9 +20,119 @@ limitations under the License.
...
+struct DynamicStitchGPULaunch {
+  static void Run(const GPUDevice& d, const int slice_size,
+                  const int first_dim_size,
+                  const CudaDeviceArrayStruct<int32>& indices_flat,
+                  const CudaDeviceArrayStruct<const T*>& data_slice_ptrs,
+                  T* output);
+};
+
+template <typename T>
+struct LaunchDynamicStitchOp<GPUDevice, T> {
```
girving: Upon further thought: the code doesn't seem to handle collisions correctly at all.  Am I reading this right?

girving (changes requested):
Please use a real commit message.

girving:
I took a look.  For incremental changes, please leave the current commit as is and push new commits on top, *without* pulling in from master.

tensorflow/core/kernels/dynamic_stitch_op.cc
MycChiu: Yes, I should actually just replace `ptr_num` with the value of `indices_vec(ind_num)`, so when collision happens the new index value will naturally overwrite the old one, which would be the same as how current implementation handles collisions, is my understanding correct? 

However, if we end up taking this approach, we would also need to account for missing indices on `indices_flat`. To do this, I plan on adding 1 to all the values of output index then subtract it back on the GPU kernel, so if the threads on CUDA kernel encounter a 0 on `output_id` they can just skip the read and write. 

Unfortunately, I think `CudaDeviceArrayOnHost` is not zero initialized, so I would need to do it manually. Is there a standard way to zero-initialize a temporary tensor? or do I just use normal for-loop to achieve this?

MycChiu:
@girving Thank you for the review! After reading your feedback, I realized that I totally omitted the cases where there are collisions in the output indices when I wrote the code, and since collisions never happened in my own use cases, I never caught this major flaw in my mind.

At first, I thought it would be hard to let GPU mimic how current CPU implementation handles collisions, but then I realized that some rather simple modifications to my current code would achieve the same effect.

I wrote the details under the relevant part of the code review, please take a look. I want to make sure the approach is right before I start modifying the code. Thank you very much!

tensorflow/core/kernels/dynamic_stitch_op.cc
girving: No, that is a bad way to handle collisions.  It doesn't respect the documented semantics at all: the order of writes would be entirely random.  Also, it doesn't work if each output index corresponds to a slice with more than one element, and would require very slow atomic operations.

A better way would be to have a single array indexed by `first_dim_size` that's computed on the CPU with correct collision resolution.  In the case of no uninitialized entries in the output array (the common case), this array will be no larger than current arrays.

However, this whole approach of doing a large chunk of the work on CPU seems low quality at best.  I don't see any way around it given the collision requirements, though, and the current fake GPU implementation is probably strictly worse.  Unpleasant!

drpngx:
@girving what do you think of @MycChiu 's proposed approch in the last comment?

girving:
Apologies for missing the notifications earlier.  Looking now.

girving:
Oh, it looks like I already commented that the proposed adjustment didn't work.  @drpngx Do I have the flow right, or did I miss something new?

drpngx:
Oh, OK thanks @girving ! @MycChiu could you propose an alternative?

martinwicke:
ping @MycChiu

vrv:
Another ping for @MycChiu (it looks like they are not around lately).

girving:
I think we should close for now.

ahundt:
for those that come across this see https://github.com/tensorflow/tensorflow/pull/11940 for a merged implementation

PR #8270, opened 2017-03-10
Explain that to restore without a graph one can use import_meta_graph

ariejdl:
Explain that to restore without a graph one can use import_meta_graph


martinwicke (approved):

martinwicke:
Jenkins, test this please.

PR #7872, opened 2017-02-25
updated word2vec basic example with current initialize_all_variables(…

tbonza:
…) from issue #5514

Basic example wasn't running so I updated it with change recommended in [Issue 5514](https://github.com/tensorflow/tensorflow/issues/5514).

PR #8272, opened 2017-03-10
Fix the initializer and make global variable specified in word2vector example

hsiaoyi0504:
- Fix the initializer
- Make global variable explicitly specified
Original modification in #7872 is not for master.
```
WARNING:tensorflow:From word2vec_basic.py: initialize_all_variables (from tensorflow.python.ops.variables) is deprecated and will be removed after 2017-03-02.
Original version works for me. The referenced issue is not for current version.
```
The modification will cause the example arise this warning.

### DIFF
--- a/tensorflow/examples/tutorials/word2vec/word2vec_basic.py
+++ b/tensorflow/examples/tutorials/word2vec/word2vec_basic.py
@@ -62,8 +62,7 @@ print('Data size', len(words))
 vocabulary_size = 50000
 
 
-def build_dataset(words):
-  global vocabulary_size
+def build_dataset(words, vocabulary_size):
   count = [['UNK', -1]]
   count.extend(collections.Counter(words).most_common(vocabulary_size - 1))
   dictionary = dict()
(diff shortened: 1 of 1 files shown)


hsiaoyi0504:
I signed it!

tensorflow/examples/tutorials/word2vec/word2vec_basic.py
```diff
@@ -63,6 +63,7 @@ def read_data(filename):
 
 
 def build_dataset(words):
+  global vocabulary_size
```
caisq: Avoid using global variables. Prefer making it into an argument of build_dataset.

caisq (changes requested):

tensorflow/examples/tutorials/word2vec/word2vec_basic.py
hsiaoyi0504: Ok, I will fix it, but the original purpose of me is just to make the original implementation more readable with minimal change.

caisq (approved):

caisq:

dandelionmane:
Looks like the cmake issue is a tool failure. Since it's a py-only change I feel comfortable merging it. Thanks for the fix!

PR #8292, opened 2017-03-11
Update documentation for BN

Conchylicultor:
The [documentation code](https://www.tensorflow.org/api_docs/python/tf/contrib/layers/batch_norm) to update `moving_mean` and `moving_variance` for BN raise an error with TF 1.0:
 
```
AttributeError: module 'tensorflow' has no attribute 'control_flow_ops'
``` 

It seems the new way to add dependencies between ops is to use the `tf.control_dependencies`.

I also attached the `updates_op` to the `train_op` instead of `total_loss` to be more coherent with the associated warning: `[...] they need to be added as a dependency to the train_op, example:`. It also make more sense as we don't want to update `moving_mean` when computing the loss on the testing set.


zhangyaobit (approved):
Thank you!

zhangyaobit:
Jenkins, test this please.

PR #8308, opened 2017-03-11
Indentation in word2vec_basic.py

brchristian:
Running the raw word2vec_basic.py in a python terminal threw a number of `IndentationError: unexpected indent` exceptions, and it looks like the cause is the indentation of the blank lines.

I’ve added the expected spacing into the blank lines here, and can confirm that the script now runs correctly.


brchristian:
I signed it!

caisq:
@brchristian , thank you for sending the PR. However, these changes are not consistent with Google Python style. Therefore I will close this PR. With regard to the issue you are experiencing, maybe you can do do `run <file_name>` instead of copy-pasting the code.

PR #8331, opened 2017-03-12
Fix `go generate` error in case multiple directories in GOPATH exist

### LINKED ISSUE #7136: ../genop/main.go:15: running "sh": exit status 1
I am unable to get the Go binding for TensorFlow working on OS X 10.12. I followed the installation instructions at https://github.com/tensorflow/tensorflow/tree/master/tensorflow/go.

Things go wrong when generating the wrapper functions for TensorFlow ops:
go generate github.com/tensorflow/tensorflow/tensorflow/go/op

The output from the above command is:
../genop/main.go:15: running "sh": exit status 1
op/generate.go:15: running "go": exit status 1

When browsing the sources i see an imported package
"github.com/tensorflow/tensorflow/tensorflow/go/genop/internal/proto/tensorflow/core/framework"
which isn't there. Maybe that is causing the issue?

yongtang:
This fix tries to address the issue raised in #7136 where `go generate` will produce an error when multiple directories in GOPATH exists.

The issue is because, in case multiple directories exist in GOPATH, the following line in `generate.sh` will not work:
```
cd $(dirname $0)
TF_DIR=${GOPATH}/src/github.com/tensorflow/tensorflow
PROTOC="${TF_DIR}/bazel-out/host/bin/external/protobuf/protoc"
```

This fix address the issue by iterating througgh each directory in `GOPATH` so that `protoc` (`${PROTOC}`) could be correctly located:
```
for g in $(echo $GOPATH | sed "s/:/ /g"); do
    TF_DIR="${g}/src/github.com/tensorflow/tensorflow"
    PROTOC="${TF_DIR}/bazel-out/host/bin/external/protobuf/protoc"
    if [ -x "${PROTOC}" ]; then
        break
    fi
done
```

This fix fixes #7136.

### DIFF
--- a/tensorflow/go/genop/generate.sh
+++ b/tensorflow/go/genop/generate.sh
@@ -20,8 +20,13 @@ go get github.com/golang/protobuf/proto
 go get github.com/golang/protobuf/protoc-gen-go
 
 cd $(dirname $0)
-TF_DIR=${GOPATH}/src/github.com/tensorflow/tensorflow
-PROTOC="${TF_DIR}/bazel-out/host/bin/external/protobuf/protoc"
+for g in $(echo $GOPATH | sed "s/:/ /g"); do
+    TF_DIR="${g}/src/github.com/tensorflow/tensorflow"
+    PROTOC="${TF_DIR}/bazel-out/host/bin/external/protobuf/protoc"
+    if [ -x "${PROTOC}" ]; then
+        break
+    fi
+done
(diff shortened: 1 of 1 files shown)


notnot:
i'm not a shell script expert, but i think i get the idea and it looks good to me

caisq:

jhseu (approved):

Issue #8238, opened 2017-03-09
OS/X compile error with tensorflow/compiler/xla/service/allocation_tracker.cc


DavidNorman:
There is a length/signed/unsigned mismatch in an inline vector initialization.

```
tensorflow/compiler/xla/service/allocation_tracker.cc:178:54: error: non-constant-expression cannot be narrowed from type 'std::vector<se::DeviceMemoryBase>::size_type' (aka 'unsigned long') to 'long long' in initializer list [-Wc++11-narrowing]
        ShapeUtil::GetSubshape(allocation->shape(), {i}),
                                                     ^
tensorflow/compiler/xla/service/allocation_tracker.cc:178:54: note: insert an explicit cast to silence this issue
        ShapeUtil::GetSubshape(allocation->shape(), {i}),
                                                     ^
                                                     static_cast<long long>( )
1 error generated.
```

diff:

```
--- a/tensorflow/compiler/xla/service/allocation_tracker.cc
+++ b/tensorflow/compiler/xla/service/allocation_tracker.cc
@@ -175,7 +175,7 @@ StatusOr<std::vector<GlobalDataHandle>> AllocationTracker::DeconstructTuple(
        i < element_bases.size(); ++i) {
     element_handles.push_back(RegisterInternal(
         allocation->backend(), allocation->device_ordinal(), element_bases[i],
-        ShapeUtil::GetSubshape(allocation->shape(), {i}),
+        ShapeUtil::GetSubshape(allocation->shape(), {static_cast<long long>(i)}),
         tensorflow::strings::StrCat(allocation->tag(), ".element_", i),
         /*initial_ref_count=*/2));
   }
```

inflation:
Can confirm.

prb12:
@DavidNorman Please feel free to submit a pull request!

DavidNorman:
will be fixed by https://github.com/tensorflow/tensorflow/pull/8276

PR #8276, opened 2017-03-10
Fix XLA build breakage on Mac OS X.

hawkinsp:
Fixes two compilation problems on Mac OS X:
* call_graph.*: `error: incomplete type 'xla::CallGraph' used in type trait expression`
* allocation_tracker.cc: `error: non-constant-expression cannot be narrowed from type 'std::vector<se::DeviceMemoryBase>::size_type' (aka 'unsigned long') to 'long long' in initializer list [-Wc++11-narrowing]`

PR #8357, opened 2017-03-13
Fix for compilation error in OS/X, due to size mismatch

DavidNorman:
See https://github.com/tensorflow/tensorflow/issues/8238

Error was:

```
tensorflow/compiler/xla/service/allocation_tracker.cc:178:54: error: non-constant-expression cannot be narrowed from type 'std::vector<se::DeviceMemoryBase>::size_type' (aka 'unsigned long') to 'long long' in initializer list [-Wc++11-narrowing]
        ShapeUtil::GetSubshape(allocation->shape(), {i}),
                                                     ^
tensorflow/compiler/xla/service/allocation_tracker.cc:178:54: note: insert an explicit cast to silence this issue
        ShapeUtil::GetSubshape(allocation->shape(), {i}),
                                                     ^
                                                     static_cast<long long>( )
1 error generated.
```


hawkinsp (commented):
Is this still needed after #8276 ?

DavidNorman:
No = your change includes this one.

PR #8363, opened 2017-03-13
Tensorboard smoothing 1st order lag filter

jyegerlehner:
@dandelionmane  This is an alternative to [PR7891](https://github.com/tensorflow/tensorflow/pull/7891), since you mentioned you are considering alternative ways forward. 

It replaces the use of a moving average window with a simple first-order lag filter. It has a few advantages: 1) A bit less code. 2) It doesn't require visiting a potentially large number of values in the window to be averaged when computing each smoothed value. 3) The degree of smoothing doesn't change when you get near the end of the curve.

I left the changes in the html file so you could just run it without having to rebuild.

Here's what it does on the same data shown in the other PR.
p_alt

### DIFF
--- a/tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
+++ b/tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
@@ -430,28 +430,25 @@ module VZ {
     }
 
     private resmoothDataset(dataset: Plottable.Dataset) {
-      // When increasing the smoothing window, it smoothes a lot with the first
-      // few points and then starts to gradually smooth slower, so using an
-      // exponential function makes the slider more consistent. 1000^x has a
-      // range of [1, 1000], so subtracting 1 and dividing by 999 results in a
-      // range of [0, 1], which can be used as the percentage of the data, so
-      // that the kernel size can be specified as a percentage instead of a
-      // hardcoded number, what would be bad with multiple series.
-      let factor = (Math.pow(1000, this.smoothingWeight) - 1) / 999;
-      let data = dataset.data();
-      let kernelRadius = Math.floor(data.length * factor / 2);
+      var data = dataset.data();
+      var smoothingWeight = this.smoothingWeight;
       data.forEach(function (d, i) {
-        var actualKernelRadius = Math.min(kernelRadius, i);
-        var start = i - actualKernelRadius;
-        var end = Math.min(data.length - 1, i + actualKernelRadius);
-        // Only smooth finite numbers.
         if (!_.isFinite(d.scalar)) {
-          d.smoothed = d.scalar;
-        } else {
-          d.smoothed = d3.mean(data.slice(start, end).filter(function (d) { return _.isFinite(d.scalar); }), function (d) { return d.scalar; });
+            d.smoothed = d.scalar;
+        }
+        else if (i == 0) {
+            d.smoothed = d.scalar;
+        }
+        else {
+          var prev_index = i > 0 ? i - 1 : 0;
+          if (!_.isFinite(data[prev_index].smoothed)) {
+              d.smoothed = d.scalar
+          }
+          else {
+            d.smoothed = data[prev_index].smoothed * smoothingWeight + (1.0 - smoothingWeight) * d.scalar;
+          }
         }
       });
-
     }
 
     private getDataset(name: string) {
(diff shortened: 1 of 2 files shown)


tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
```diff
@@ -430,34 +430,23 @@ module VZ {
...
+          d.smoothed = d.scalar;
+        }
+        else {
+          var prev_index = i - 1;
+          if (!_.isFinite(data[prev_index].smoothed)) {
+            d.smoothed = d.scalar
+          }
+          else {
+            d.smoothed = data[prev_index].smoothed * smoothingWeight + (1.0 - smoothingWeight) * d.scalar;
+          }
```
dandelionmane: I think it is cleaner as:

```
let last = data.length > 0 ? data[0].scalar : NaN;
data.forEach((d) => {
  if (!_.isFinite(last)) {
    d.smoothed = d.scalar;
  } else {
    d.smoothed = last * smoothingWeight + (1 - smoothingWeight) * d.scalar;
  }

  last = d.smoothed;
}

teamdandelion (changes requested):
Overall, looks good! We actually used to have this smoothing (also known as exponentially weighted moving average), but we switched away from it because it was too biased near the start of the sequence (the first point implicitly has a very high weight). This was unfortunate because often with loss curves the first point is also an extreme outlier.

However, I'm working on another change that will filter out outlier data, when computing the y-scale domain, so the outlier won't matter as much visually, and that makes EMA a nice approach to smoothing.

tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts
jyegerlehner: Thanks; updated.

jyegerlehner:
There is another reason I think this kind of filtering is superior. The moving-average window filter currently implemented averages over a window centered on the current point, and so in general the "kernel radius" stretches forward to include future data, making it non-causal. In the common use case where you've got a training loss curve, you make a hyper-param or other change, and restart training, and then you want to compare how training progresses relative to the previous iteration. The previous iteration's loss curve you are comparing to stretches out into the future, which is reflected in its curve. However for the currently-executing training, the future loss values don't exist yet so you can't directly compare the latest smoothed curve values of the currently-executing training loss to that of the previous iterations. 

That problem doesn't apply with this kind of smoothing. In the case of this PR, the smoothed curves are always causal, so you can compare previous iterations' curves to the current iteration's curve at every point, right up to the most recent data. Sure, it has some lag, but the lag is the same for all the curves.

Hope that makes sense.

martinwicke:
@dandelionmane what's the plan with this and #7891?

teamdandelion (approved):

dandelionmane:
LGTM. Please fix the conflicts (you can just port the changes into the new tf-tensorboard.html) and then we'll merge.

dandelionmane:
Or you can just revert the conflicting file and wait for us to update it.

jyegerlehner:
OK thanks; I reverted html file.

martinwicke:
Jenkins, test this please.

cancan101:
Not sure if this "issue" was introduced by this PR, but this seems odd to me:
image
The smoothed line (dark purple is not defined for the last N steps). If the implication is that smoothing cannot be calculated for these values, then I would argue the same should apply for the first N steps as well. 

Rather I would prefer smoothing to be defined all the way to the left and right with a reduced window (which is what I assume happens initially) on both sides.

jyegerlehner:
@cancan101 I checked and tf-tensorboard.html in 1.1 does not yet reflect the changes made in this PR (as of this writing), and that's the file that determines the behaviour one sees when running tensorboard. It would show up in the `resmoothDataset` method in tf-tensorboard.html which needs to be regenerated from the underlying .ts file that was changed as a part of this PR. There are instructions on how to do that somewhere under the tensorboard source subtree. 

The problem you show (I think) is one of the things this PR was meant to correct. 

A simple hack that may be easier than regenerating the `tf-tensorboard.html` properly, is to replace the `resmoothDataset` function in `tf-tensorboard.html` with the version of the function that's in  `tensorflow/tensorboard/components/vz_line_chart/vz-line-chart.ts `. At least I think that worked for me.

PR #8372, opened 2017-03-13
Fix CameraActivity typo

seonho:
Fix crash when activity paused

### DIFF
--- a/tensorflow/examples/android/src/org/tensorflow/demo/CameraActivity.java
+++ b/tensorflow/examples/android/src/org/tensorflow/demo/CameraActivity.java
@@ -82,7 +82,7 @@ public abstract class CameraActivity extends Activity implements OnImageAvailabl
   public synchronized void onPause() {
     LOGGER.d("onPause " + this);
 
-    if (!isFinishing()) {
+    if (isFinishing()) {
       LOGGER.d("Requesting finish");
       finish();
     }


seonho:
I signed it!

seonho:
When I open popup activity (e.g., PreferenceActivity), onPause is called and "Activity has leaked window that was originally added ..." error occurs.
This is because the code wrote as !isFinishing(), activity call finish() even thouch it is not actually finishing.
New activity works, but cannot go back to previous activity since it's already finished.

martinwicke:
ping @andrewharp

andrewharp:
So to clarify, it's when you modify the demo to create new windows that you have issues? Does this finish() line actually ever get called for you with this change?

seonho:
@andrewharp Exactly. To be able to return to the Camera Activity, finish() should not be called when creating a new window.

andrewharp:
@seano314 Ok, so what I am trying to understand is why it would ever make sense to call finish() if the app is already finishing. It seems that there is no typo here that I can see, just a use-case that the demos don't currently need to account for.

If we remove the explicit finish() call, the I worry that confusing behavior may occur when users switch quickly between different demo activities. Adding finish() here was a quick stop-gap solution that worked, but if you would like to add a real fix that does things the Android way that would be appreciated. Otherwise I think we should keep it as-is, or find another way to make your modifications work with the version of TF in the repo that doesn't break the existing demos..

seano314:
@seonho Mind taking a look at @andrewharp comment above? Looks like his question is for you? :)

andrewharp:
@seano314 Closing as this is intended behavior. I think I understand the issue you're seeing, but it needs to be handled differently, as simply reversing the check actually introduces a bug.

PR #8373, opened 2017-03-14
Switch back to Ubuntu 14.04 for our releases.

jhseu:
Haven't tested this yet. Letting Jenkins do the work.

@caisq fyi. Note that we should keep our Docker Hub images on Ubuntu 16.04 because that fixes the bugs users encounter.

Note for others looking at this change: we're downgrading to Ubuntu 14.04 for official releases so that it links against the older version of glibc and allows for wider compatibility.

### DIFF
--- a/tensorflow/tools/ci_build/Dockerfile.android
+++ b/tensorflow/tools/ci_build/Dockerfile.android
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:14.04
 
 MAINTAINER Jan Prach <jendap@google.com>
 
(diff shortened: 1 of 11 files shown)


jhseu:
@gunan this is mostly a revert plus some minor version changes and edits.

yaroslavvb:
I'm expecting you'll see some jemalloc caused SEGFAULTS on process shutdown, since that happens in 14.04 but not 16.04

jhseu:
We're using the other fix here: reinstalling numpy without OpenBLAS support.

caisq:
@jhseu Thanks for cc'ing me. +1 for keeping the Docker Hub and gcr.io images at 16.04.

gunan (approved):

gunan:
Thank you very much for the change!

PR #8400, opened 2017-03-14
Update Java API test for size and missing asserts

craykg:
Corrected issues where unit tests tested too many things in a single
test, and where they didn't perform any effective tests. Its best
practice to keep the size of test to as close to one assert as
possible, and to not have tests that never test anything.


martinwicke:
ping @asimshankar

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -105,55 +116,90 @@ public void createFromBufferWithNonNativeByteOrder() {
...
+    final double[] expected = {1D, 2D, 3D, 4D};
+
+    try(Tensor t = Tensor.create(new long[expected.length + 1], DoubleBuffer.wrap(expected))) {
+    }
+  }
+
+  @Test
+  public void whenCreateUsingFloatBufferWithIncorrectShapeThenThrowExpectedException() {
+    expectedException.expect(IllegalArgumentException.class);
+    expectedException.expectMessage("buffer with 4 elements is not compatible with a Tensor with shape [0, 0, 0, 0, 0]");
```
asimshankar: Now that the tests are broken up and we're not re-using the same testdata, perhaps these can be simplified (e.g., using: `expected = {1F};` instead of a longer array)

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -256,82 +302,114 @@ public void writeTo() {
...
     }
+  }
 
-    try (Tensor t = Tensor.create(8589934592L)) {
-      assertEquals(DataType.INT64, t.dataType());
-      assertEquals(0, t.numDimensions());
-      assertEquals(0, t.shape().length);
-      assertEquals(8589934592L, t.longValue());
+  @Test
+  public void whenCreateUsingDoubleThenReturnDoubleScalarWithExpectedValue() {
```
asimshankar: Can we shorten these and other test names where possible. For example, this can be just
```java
@Test
public void createScalarDoubleTensor()
```

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -401,71 +479,74 @@ public void failCreateOnMismatchedDimensions() {
   }
 
   @Test
-  public void failCopyToOnIncompatibleDestination() {
+  public void whenCopyToWithDimensionMismatchThenThrowExpectedException() {
```
asimshankar: Many of these names seem more verbose than they need to be. Is there some well known convention around naming all tests as "whenXThenY" that you're following? I'd be tempted to stick with more succinct names.

tensorflow/java/src/test/java/org/tensorflow/TensorMatcher.java
```diff
@@ -0,0 +1,86 @@
+/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
asimshankar: It seems to me that we get useful information without adding more test convenience code. I'd be tempted to drop this file altogether

asimshankar (commented):
Thanks for the PR. Breaking the tests into smaller units sounds great.

That said though, I do have some reservations about the verbosity of the names of the tests and the value of the additional matchers (e.g., `assertEquals(t.shape().numDimensions(), 0)` vs `assertThat(is(scalar(t))`) .

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
craykg: Certainly could. My general aim, given how invasive the changes were to the structure of the files, was to leave the test data intact in case it was like it was for a reason.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -105,55 +116,90 @@ public void createFromBufferWithNonNativeByteOrder() {
   }
 
   @Test
-  public void createWithTypedBuffer() {
-    int[] ints = {1, 2, 3, 4};
-    float[] floats = {1f, 2f, 3f, 4f};
-    double[] doubles = {1d, 2d, 3d, 4d};
-    long[] longs = {1L, 2L, 3L, 4L};
-    long[] shape = {4};
+  public void whenCreateUsingDoubleBufferThenExpectedMatrixIsStoredInTensor() {
```
craykg: My mistake, saw matrix sprinkled throughout a lot of tests (for matrix data) and put the wrong term in there. Will fix that up.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -105,55 +116,90 @@ public void createFromBufferWithNonNativeByteOrder() {
...
+    final double[] expected = {1d, 2d, 3d, 4d};
+    final long[] shape = {4};
+
+    try (final Tensor t = Tensor.create(shape, DoubleBuffer.wrap(expected))) {
+      final double[] actual = new double[expected.length];
+      t.copyTo(actual);
+      assertArrayEquals(expected, actual, EPSILON);
```
craykg: My reasoning behind the three lines was:

- I prefer to break tests up so that the arrange/act/assert statements are separate where possible (as it improves readability which is the most important aspect of tests)
- Most junit asserts, like this one, are backwards compared to assertThat, and I wanted to be absolutely sure I got them in the right order.

It can change if that is preferred, but I don't think it should.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
craykg: Generally I aim to write test method names so that when they show up in a failure report, you don't have to look at the test code to work out what went wrong. The whenXThenY is a shorted version of behaviour driven development's Given When Then format.

In the case of failCopyToOnIncompatibleDestination() you know it should fail, but it doesn't tell you how its should fail, nor does it tell you what specific thing you were expecting to cause the failure. Anyone who has dug into the Tensor class will know that you can fail on a type mismatch, a dimension mismatch, or a shape mismatch, and all of these throw a IllegalArgumentException with different exception messages (or at least should do). 

We could change it to failCopyToOnIncompatibleDestinationShape() although that still doesn't tell us how its supposed to fail.

tensorflow/java/src/test/java/org/tensorflow/TensorMatcher.java
craykg: The benefit comes from readability in the tests, and improved failure messages, and in the case of Scalar Tensors a reduction in the number of asserts needed per test. Generally, I advocate for replacing all assert* with assertThat and hamcrest matchers because of readability of code and failure messages. There are also type safety and portability (between junit and testng) benefits as well, but thats not usually why I'm advocating for them.

**Readability**
`assertThat(tensor, is(scalar());` 
versus
`assertEquals(0, tensor.numDimensions());
assertEquals(0, tensor.shape().length);`

The first when read left to right is literally _'assert that tensor is scalar'_. Using the older junit assertEquals we need two tightly coupled asserts statements to prove a tensor is scalar, which read like _'assert equals 0 the tensor.numDimensions and assert equals 0 the tensor.shape length'_

**Failure messages**
If we have a tensor with a dimension of 1 and a shape length of 1, the assertThat failure message will read:
"Expected: Tensor with shape length [0] and number of dimensions [0]
but: was Tensor with shape length [1] and number of dimensions [1]"

The assertEquals failure messages will read:
Expected: 0 was 1.

**Type safety**
assertThat(tensor, is(scalar()); will not compile if tensor is not actually a tensor. This is less of a problem for assertEquals(0, tensor.numDimension()) and assertEquals(0, tensor.shape().length) because we are calling specific function numDimension() and shape() which are non-standard java getters names anyway (so chance of accidentally having a class with the same method signatures is probably low). Still, something to think about.

Ultimately I was on the fence about adding the tensor matcher class, primarily because of the additional maintenance burden all code come with. What pushed me over to the side of adding one is that the tensor seems to be the building block of the program, so it is likely to be used all over the place to test things.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
asimshankar: Readability is in the eye of the reader :). I find the single line more readable, but I don't feel too strongly and appreciate that others may legitimately differ, so you're welcome to keep the three lines if you wish.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
asimshankar: `ThenThrowExpectedException` seems like a verbose way of saying `fail` :). I suppose there are other ways to fail. Anyway, the "when", "then", and "expected" words in all these tests do seem a bit verbose to me. I'd prefer keeping the names concise. For example, in this case `failCopyToOnDimensionMismatch` vs. `whenCopyToWithDimensionMismatchThenThrowExpectedException` - both of them seem like they convey the same information when shown up in a failure report.

tensorflow/java/src/test/java/org/tensorflow/TensorMatcher.java
asimshankar: I appreciate the value of matchers in general and how their use is a judgement call in any given situations. In this case, I would like to avoid additional indirection and maintenance burden.

```java
assertEquals(0, tensor.numDimensions());
assertEquals(0, tensor.shape().length());
```
very explicitly tells the reader what properties of the `Tensor` object are being validated.
On the other hand `assertThat(tensor, is(scalar()))` doesn't. Is it looking at all relevant properties? Which ones are relevant? I think on balance, I'd prefer the explicit calls here.

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
RealTimeDeployment: Shortening names sounds good to me

tensorflow/java/src/test/java/org/tensorflow/TensorTest.java
RealTimeDeployment: It is verbose more figurative than literal. I did not know what language would be used. If,if else

drpngx:
@craykg any chance to address the comments?

craykg:
@drpngx Haven't had a chance yet - planning on making the changes on Friday/Saturday.

martinwicke:
ping @craykg

vrv:
Another ping for @craykg

vrv:
(Closing this for now to keep our backlog low since it's been a while since we've heard back -- once you are back and capable of moving forward we can make progress again!)

PR #8402, opened 2017-03-14
Branch 150085723

jhseu:


gunan (approved):

yaroslavvb:
@gunan btw, one idea of making some of these flaky numeric tests less flaky is to always run nodes in the same order. We improved determinism of our stuff by adding control deps that force computation to use deterministic execution order (https://github.com/yaroslavvb/stuff/tree/master/linearize)

gunan:
@yaroslavvb Thanks for the suggestion.
When we previously looked into this there were some issues we ran into when making the run order deterministic.
@aselle I think you took a look at this, right?
Maybe we should reconsider this using the code @yaroslavvb shared as a starting point?

PR #8420, opened 2017-03-15
set _interactive in logging utilities

### LINKED ISSUE #8401: Tensorflow not working in Zeppelin 0.7.0
There seems to be some issue using Tensorflow in Zeppelin 0.7.0 and it throws this error:
NameError: name `_interactive` is not defined

Another user reported the [same issue](http://stackoverflow.com/questions/42757433/tensorflow-can-not-work-with-zeppelin) on SO. The fix suggested there seems to be really hacky.

The issue seems to be that in tf_logging.py file [.\Anaconda3\Lib\site-packages\tensorflow\python\platform\tf_logging.py] the _interactive variable is not triggering for Zeppelin.

```
# Determine whether we are in an interactive environment
try:
  # This is only defined in interactive shells
  if _sys.ps1: _interactive = True
except AttributeError:
  # Even now, we may be in an interactive shell with `python -i`.
  _interactive = _sys.flags.interactive
```

nsriram13:
this update fixes #8401

### DIFF
--- a/tensorflow/python/platform/tf_logging.py
+++ b/tensorflow/python/platform/tf_logging.py
@@ -37,6 +37,7 @@ from tensorflow.python.util.all_util import remove_undocumented
 
 
 # Determine whether we are in an interactive environment
+_interactive = False
 try:
   # This is only defined in interactive shells
   if _sys.ps1: _interactive = True


nsriram13:
I just signed the CLA!

jhseu (approved):

jhseu:
Jenkins, test this please

PR #8450, opened 2017-03-15
Move bazelrc to the master location so that it concatenates both configurations.

jhseu:


gunan (approved):

jhseu:
Mac timeouts are unrelated, but we do need to move this file on the macs. They're still colorless.

gunan:
Could you file a bug for the bazelrc changes for macs?
We will take care of those too as soon as possible.

Issue #6682, opened 2017-01-06
Weird behavior of tf.tensordot when shapes are partly known


jihunchoi:
I frequently use ```tf.tensordot```, for its flexibility on axes.
However, as I multiply two tensors whose shapes are partly known, it cannot infer the shape of the result tensor, i.e. shape is `<unknown>`, while other alternative functions that do the same operation infer shapes well.

For example, let `a` and `b` be
```
a = tf.placeholder('float32', shape=[None, 100])
b = tf.placeholder('float32', shape=[100, 300])
```

We can validate that the below functions infer shapes well.
```
result_matmul = tf.matmul(a, b)
result_matmul.get_shape().as_list()  # [None, 300]

result_einsum = tf.einsum('ij,jk->ik', a, b)
result_einsum.get_shape().as_list()  # [None, 300]
```

However, when `tf.tensordot` is used, the weird result occurs:
```
result_tensordot = tf.tensordot(a, b, axes=[[1], [0]])
result_tensordot.get_shape()  # TensorShape(None)
result_tensordot.get_shape().as_list()  # Error
```
Thus, I have to call `set_shape(shape)` function explicitly to make the output be able to be used for further operations.

There also exists an issue that it still uses `concat` function, where the warning that it would be deprecated after 2016-12-14 occurs, instead of `concat_v2`.

michaelisard:
@rmlarsen here's a customer for partial shape inference in tensordot. Any plans on working on it, or want to throw it over to contributions welcome?

rmlarsen:
Despite having a TODO for myself in the code, I'll add contributions welcome, in case somebody can beat me to it. It is not on the top of my list.

jihunchoi:
Okay. I will look forward it if there's a point I can contribute.
By the way, is it somewhat "uncommon" to use `tensordot` function?
I was a user of "reshape (or transpose) - matmul - reshape (or transpose)" approach previously, and am finding methods for doing this boring series of operations.
I think `einsum` function is an attractive alternative, indeed more powerful alternative. A single question that makes me hesitate to move to `einsum` is: is there no performance issue?

jihunchoi:
Actually, for some cases, `einsum` also fails!
For instance, the below codes fails on evaluation.

```
a = tf.placeholder('float32', shape=[None, None, 100])
b = tf.placeholder('float32', shape=[100, 300])
result_einsum = tf.einsum('ijk,kl->ijl', a, b)
[...]

PR #8452, opened 2017-03-16
Tensordot partial shape inference

person594:
Addresses #6682.  Tensordot now correctly infers partial shapes when possible.  In some cases, such inference is still impossible, such as when axes are specified as a tensor or when either of the operand tensors has an unknown number of dimensions.

The tests for tensordot were updated as well in order to test the new behavior


jhseu:
Jenkins, test this please

rmlarsen (approved):

rmlarsen:

rmlarsen:
@person594 Thanks for the fix!

PR #8468, opened 2017-03-16
Fix minor error-handling bug

critiqjo:


critiqjo:
I signed it!

tensorflow/contrib/seq2seq/python/ops/helper.py
```diff
@@ -443,7 +443,7 @@ def sample(self, time, outputs, state, name=None):
     # Outputs are logits, use argmax to get the most probable id
     if not isinstance(outputs, ops.Tensor):
       raise TypeError("Expected outputs to be a single Tensor, got: %s" %
```
caisq: it would be clearer to say "got type: ".

caisq (changes requested):

tensorflow/contrib/seq2seq/python/ops/helper.py
critiqjo: Sure, I was actually borrowing the pattern from [right above it](https://github.com/critiqjo/tensorflow/blob/84f0cce/tensorflow/contrib/seq2seq/python/ops/helper.py#L337-L338). Should I change that too?

caisq: OK. In that case never mind. On second thought, the alternative I suggested was not 100% clear, either. What you have is fine and shorter.

caisq (approved):

caisq:

PR #8488, opened 2017-03-17
Checkpoints default values for RunConfig

Immexxx:
With save_checkpoints_secs having a default value (of 600), the call to RunConfig fails if only save_checkpoints_steps is passed. To address this, the value is defaulted only if both the parameters are None. In this new way, the call to RunConfig will not fail if only save_checkpoints_steps is passed

### DIFF
--- a/tensorflow/contrib/learn/python/learn/estimators/run_config.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/run_config.py
@@ -259,7 +259,7 @@ class RunConfig(ClusterConfig):
     self._tf_random_seed = tf_random_seed
     self._save_summary_steps = save_summary_steps
     self._save_checkpoints_secs = save_checkpoints_secs
-    if save_checkpoints_secs == _USE_DEFAULT:
+    if save_checkpoints_secs == RunConfig._USE_DEFAULT:
       if save_checkpoints_steps is None:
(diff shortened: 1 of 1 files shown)


ispirmustafa (commented):
If both of them None, that means user disables the saving. This change prevents that. Instead you can use a constant _USE_DEFAULT. if it's _USE_DEFAULT and not steps then set it to 600. if it's _USE_DEFAULT and steps then set it to None.

ispirmustafa (changes requested):
If both of them None, that means user disables the saving. This change prevents that. Instead you can use a constant _USE_DEFAULT. if it's _USE_DEFAULT and not steps then set it to 600. if it's _USE_DEFAULT and steps then set it to None.

ispirmustafa (approved):
Thanks

Immexxx:
@ispirmustafa  - Thanks for the review

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

martinwicke:
Please fix the test failures. `_USE_DEFAULT` must be qualified (as `RunConfig._USE_DEFAULT`).

Immexxx:
@martinwicke Thanks. Will do

martinwicke:
Jenkins, test this please.

PR #8522, opened 2017-03-18
Disable FILE_FLAG_RANDOM_ACCESS in WindowsFileSystem

snnn:
You will run out of physical memory if you need to read a large file on Windows.
https://support.microsoft.com/en-us/help/2549369
The behavior looks very like a memory leak, however, the process's memory usage doesn't grow.

### DIFF
--- a/tensorflow/core/platform/windows/windows_file_system.cc
+++ b/tensorflow/core/platform/windows/windows_file_system.cc
@@ -230,11 +230,9 @@ Status WindowsFileSystem::NewRandomAccessFile(
   result->reset();
 
   // Open the file for read-only random access
-  // Random access is to disable read-ahead as the system reads too much data
   // Open in async mode which makes Windows allow more parallelism even
   // if we need to do sync I/O on top of it.
-  DWORD file_flags = FILE_ATTRIBUTE_READONLY | FILE_FLAG_RANDOM_ACCESS |
-      FILE_FLAG_OVERLAPPED;
+  DWORD file_flags = FILE_ATTRIBUTE_READONLY | FILE_FLAG_OVERLAPPED;
   // Shared access is necessary for tests to pass
   // almost all tests would work with a possible exception of fault_injection.
   DWORD share_mode = FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE;
@@ -306,8 +304,8 @@ Status WindowsFileSystem::NewReadOnlyMemoryRegionFromFile(
   result->reset();
   Status s = Status::OK();
 
-  // Open the file for read-only random access
-  DWORD file_flags = FILE_ATTRIBUTE_READONLY | FILE_FLAG_RANDOM_ACCESS;
+  // Open the file for read-only
+  DWORD file_flags = FILE_ATTRIBUTE_READONLY;
 
   // Open in async mode which makes Windows allow more parallelism even
   // if we need to do sync I/O on top of it.


caisq:

snnn:
Do not merge, WIP.
Should we revert it to the original fopen/fread/fseek version?

snnn:
Some performance test result after this change:

input data: A text file in libsvm format, 10906667862 bytes, 164540416 lines, with CRLF line terminators

test code:
```
#include <mutex>
#include <Windows.h>
#include <stdio.h>

#include "tensorflow/core/lib/core/errors.h"
#include "tensorflow/core/lib/io/inputbuffer.h"
using namespace tensorflow;

int test1(const char* filename) {
	LARGE_INTEGER li;
	QueryPerformanceFrequency(&li);
	Env* env = Env::Default();
	std::unique_ptr<RandomAccessFile> file_;
	const ::tensorflow::Status status = env->NewRandomAccessFile(filename, &file_);
	if (!status.ok()) {
		return -1;
	}
	LARGE_INTEGER startTime, endTime;
	int kBufferSize = 2 * 1024 * 1024;
	io::InputBuffer* input_buffer = new io::InputBuffer(file_.get(), kBufferSize);
	QueryPerformanceCounter(&startTime);
	string line_contents;
	while (true) {
		Status status = input_buffer->ReadLine(&line_contents);
		if (errors::IsOutOfRange(status)) {
			break;
		}
		if (!status.ok()) {
			return -1;
		}
	}
	QueryPerformanceCounter(&endTime);
	const double seconds = ((endTime.QuadPart - startTime.QuadPart) / (double)li.QuadPart);
	printf("dur = %g\n", seconds);
	return 0;
}

int test2(const char* filename) {
	LARGE_INTEGER li;
	QueryPerformanceFrequency(&li);
	int kBufferSize = 2 * 1024 * 1024;
	std::unique_ptr<char[]> buf(new char[kBufferSize]);
	FILE* fd = fopen(filename, "rb");
	if (fd == nullptr) return -1;
	LARGE_INTEGER startTime, endTime;
	QueryPerformanceCounter(&startTime);
	while (fgets(buf.get(), kBufferSize, fd) != nullptr) {

	}
	QueryPerformanceCounter(&endTime);
	const double seconds = ((endTime.QuadPart - startTime.QuadPart) / (double)li.QuadPart);
	printf("read by fgets, dur = %g\n", seconds);
	return 0;
}

int main(int argc,char* argv[]) {
	const char* filename = argv[1];
	const char* method = argv[2];
	if (_stricmp(method, "fgets")==0) {
		return test2(filename);
	}
	return test1(filename);
}

```

Result:
file on SSD, read by fgets, dur = 35 seconds
file on SSD, read by InputBuffer, dur = 19 seconds
file on HardDisk, read by fgets, dur = 104 seconds
file on HardDisk, read by InputBuffer, dur = 114 seconds

So, I think it's ok to simply remove FILE_FLAG_RANDOM_ACCESS flag.

mrry:
I don't have a strong opinion about the change, but it would be great to get an opinion from @vit-stepanovs or @guschmue, who worked on this part of the Windows patch.

guschmue:
@snnn - can you provide performance numbers before and after your change ?
Basically how much perf hit is it removing FILE_FLAG_RANDOM_ACCESS?
What at what filesize do you start seeing issues ?

snnn:
Hi @guschmue , Could you tell me why do you need this flag?

guschmue:
Didn't say we need the flag. I think this code was used for some storage project where random access is common (aka FILE_FLAG_RANDOM_ACCESS is used to not evict behind. I don't think this is the common access pattern for RandomAccessFile in tensorflow and removing FILE_FLAG_RANDOM_ACCESS might be a good thing for sequential reads.
Since you already measure perf after, why not measure before as well to confirm that this also helps performance.

snnn:
Hi @guschmue  

1. FILE_FLAG_RANDOM_ACCESS flag is a hint for Cache Manager to keep mapped views(read blocks) of the file in memory as long as possible, until Memory Manager doesn’t signal low memory condition.  So, if we run tensorflow standalone, without sharing resources with other programs, the performance is almost the same. However, if we run it on yarn, the situation is quite different.  And it's hard to say how much it faster/slower because it number varies, and there is no way to control tensorflow's memory usage.

2. At the same time, this flag instructs Cache Manager to disable prefetching of file data, which could reduce unnecessary disk io and increase read latency.  However, we already have IoBuffer as the user mode cache. For sequential reads, I can't see any difference.

@mrry  It would be great if there is a new file type like "SeqAccessFile". Actually, FILE_FLAG_RANDOM_ACCESS  is good for random accesses(e.g. KV DBs, sparse model serving). It is just not good for sequential reads.

guschmue:
Agree, it should be better to turn FILE_FLAG_RANDOM_ACCESS off ... will be more common to have sequential io.

mrry:
I'm happy with this change if @guschmue is :). I agree that a specific sequential-access file interface would make sense for most input file processing; I think the main case for random access is when restoring a checkpoint. It's possible that (with this change) restoring a checkpoint will get slower but reading input files will get faster, and we should keep an eye on the former.

mrry (approved):

PR #8534, opened 2017-03-19
Performance improvement for io::inputbuffer

snnn:
Do not add chars to string one by one.
In most cases, buffer size >= max line length. 
Let's do it in one function call.

### DIFF
--- a/tensorflow/core/lib/io/inputbuffer.cc
+++ b/tensorflow/core/lib/io/inputbuffer.cc
@@ -15,6 +15,7 @@ limitations under the License.
 
 #include "tensorflow/core/lib/io/inputbuffer.h"
 #include "tensorflow/core/lib/core/errors.h"
+#include "tensorflow/core/platform/logging.h"
 
 namespace tensorflow {
 namespace io {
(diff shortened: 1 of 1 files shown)


snnn:
For input like:
"aaa\rbbb\r\n"

The old code will output: "aaabbb"
This new code will output: "aaa\rbbb"

I think this behavior change is acceptable. Do you?

mrry:
Have you measured the performance improvement?

It certainly *looks* more efficient to me, but I'd like to weigh this against the cost of making a (probably irrelevant, but you never know...) backwards incompatible change.

snnn:
Working. I'll post the result later.

snnn:
input data: A text file in libsvm format, 10906667862 bytes, 164540416 lines, with CRLF line terminators

test code 1:
```cpp
#include <mutex>
#include <Windows.h>
#include <stdio.h>

#include "tensorflow/core/lib/core/errors.h"
#include "tensorflow/core/lib/io/inputbuffer.h"
using namespace tensorflow;

int test1(const char* filename) {
	LARGE_INTEGER li;
	QueryPerformanceFrequency(&li);
	Env* env = Env::Default();
	std::unique_ptr<RandomAccessFile> file_;
	const ::tensorflow::Status status = env->NewRandomAccessFile(filename, &file_);
	if (!status.ok()) {
		return -1;
	}
	LARGE_INTEGER startTime, endTime;
	int kBufferSize = 2 * 1024 * 1024;
	io::InputBuffer* input_buffer = new io::InputBuffer(file_.get(), kBufferSize);
	QueryPerformanceCounter(&startTime);
	string line_contents;
	while (true) {
		Status status = input_buffer->ReadLine(&line_contents);
		if (errors::IsOutOfRange(status)) {
			break;
		}
		if (!status.ok()) {
			return -1;
		}
	}
	QueryPerformanceCounter(&endTime);
	const double seconds = ((endTime.QuadPart - startTime.QuadPart) / (double)li.QuadPart);
	printf("dur = %g\n", seconds);
	return 0;
}
```
test code 2:
Almost the same as test code 1, with only one line change:
```
while (true) {
    string line_contents;  // this line is moved from the line before "while(true)"
    Status status = input_buffer->ReadLine(&line_contents);
}
```
Result:

<table>
<thead>
<tr>
<td>&nbsp;</td>
<td>Before</td>
<td>After</td>
<td>Changes</td>
</tr>
</thead>
<tbody>
<tr>
<td>test1</td>
<td>67 seconds</td>
<td>26 seconds</td>
<td> - 61%</td>
</tr>
<tr>
<td>test2(cache line_contents)</td>
<td>34 seconds</td>
<td>19 seconds</td>
<td> - 44%</td>
</tr>
</tbody>
</table>

mrry:
Great, thanks for confirming that. I'm satisfied with the change, so let's run the tests and merge if they pass.

mrry (approved):

tensorflow/core/lib/io/inputbuffer.cc
```diff
@@ -43,25 +43,26 @@ Status InputBuffer::FillBuffer() {
...
       return Status::OK();
     }
-    // We don't append '\r' to *result
-    if (c != '\r') {
-      *result += c;
-    }
+    if (buf_remain > 0) result->append(pos_, buf_remain);
+    // Get more data into buffer
+    s = FillBuffer();
+    assert(pos_ == buf_);
```
mrry: One minor change, sorry: can you replace this line with `DCHECK_EQ(pos_, buf_);` to match the style of the rest of the codebase? Thanks!

mrry (changes requested):
Just noticed one minor change... should be good to merge after changing.

tensorflow/core/lib/io/inputbuffer.cc
snnn: Sure.

snnn: Done.

mrry (approved):
Great, thanks! Looks good to me.

mrry:

PR #8537, opened 2017-03-19
Fixed `needs_broadcasting` expression in tf.layers.normalization

sugyan:
in Python3, `range()` function doesn't create lists, so `[0, 1, 2] != range(4)[:-1]` always returns `True`.

### DIFF
--- a/tensorflow/python/layers/normalization.py
+++ b/tensorflow/python/layers/normalization.py
@@ -162,7 +162,7 @@ class BatchNormalization(base._Layer):  # pylint: disable=protected-access
     broadcast_shape[self.axis] = input_shape[self.axis].value
 
     # Determines whether broadcasting is needed.
-    needs_broadcasting = (sorted(reduction_axes) != range(ndim)[:-1])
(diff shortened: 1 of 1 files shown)


martinwicke:
Thanks for this fix. Can you add a test that would have failed? Clearly this code isn't tested well.

martinwicke (changes requested):
Please add a test for this behavior.

sugyan:
I am sorry for a miscellaneous pull-request with no test code.
I just pointed out a programming mistake.

I read the code and checked its behavior, but in `normalization.py`,` axis` in initializer is a single integer and `len(reduction_axes)` is always equal to `len(input_shape) - 1`, so the calculation results will be exactly the same regardless of the value of `needs_broadcast`...
I'm wondering if this variable is really necessary.

martinwicke:
Jenkins, test this please.

martinwicke (approved):

martinwicke:
Fine. This is a clear improvement. Thanks!

PR #8541, opened 2017-03-19
Modified mnist.py to utilize random seed settings.

td2014:
Modified the file mnist.py to allow for users to set the random seed, either based on the graph-level seed via tf.set_random_seed or an op-level seed via a call like:
mnist = input_data.read_data_sets(FLAGS.data_dir, one_hot=True, seed=12345).

This should allow users to get results that are repeatable during development and testing, especially for those becoming familiar with Tensorflow via the MNIST tutorials.


td2014:
Trying to fix issue with submission names and email addresses to match CLA.

PR #8542, opened 2017-03-19
Modified the file mnist.py to allow for users to set the random seed.

td2014:
Modified the file mnist.py to allow for users to set the random seed, either based on the graph-level seed via tf.set_random_seed or an op-level seed via a call like:
mnist = input_data.read_data_sets(FLAGS.data_dir, one_hot=True, seed=12345).

This should allow users to get results that are repeatable during development and testing, especially for those becoming familiar with Tensorflow via the MNIST tutorials.


td2014:
I am the only committer (who also signed the CLA).  Not sure how to fix the blocking by the googlebot.

td2014:
Created an updated pull request with CLA issue fixed.

PR #8553, opened 2017-03-20
Add header only build rules for XLA external plugins

DavidNorman:
This change includes the exporting of XLA symbols in the main python plugin.

This allows external shared-object plugins to be built against the XLA source header files
and to be loaded into a running process.

Note:  I have only exposed some protobuf symbols.  Exposing *protobuf* causes tensorboard to fail to load.

### DIFF
--- a/tensorflow/compiler/jit/BUILD
+++ b/tensorflow/compiler/jit/BUILD
@@ -18,7 +18,16 @@ package(
     default_visibility = [":internal"],
 )
 
+load("//tensorflow:tensorflow.bzl", "cc_header_only_library")
 load("//tensorflow:tensorflow.bzl", "tf_kernel_library")
+
+cc_header_only_library(
+    name = "xla_jit_headers_lib",
+    visibility = ["//visibility:public"],
+    deps = [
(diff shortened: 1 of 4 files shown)


DavidNorman:
hello.   i've been looking into my problems with the startup error when Variables are included in the graph (not always - but it fails for all of the variable type unit tests).

https://github.com/google/protobuf/issues/2894

I suspect, after reading a lot of protobuf stuff, that it might be hard/impossible to have a plugin due to protobuf global state / static initializers.  i note that it seems to be ok for kernel_op plugins, but they might sidestep some global state issue.

nevertheless, i think that a public headers interface for XLA is a good thing, so that I can have my driver code elsewhere in the repo.  At the moment only things in //tensorflow/compiler can access enough of the headers to compile.

DavidNorman:
ok - i'm abandoning this.  i am going to build a 'tf with poplar' integrated release along the lines of the other backends.

DavidNorman:
at some point i will still upstream some header only targets, just to keep my stuff as much in line with the public repo as possible.

PR #8563, opened 2017-03-20
Remove unused assets/code from iOS Camera Example

joeymeyer:
After going through the iOS camera example this weekend I noticed several assets are unused, and there are several methods that only call `super`. Let me know if I missed something and they're there for a reason. Feel free to squash commits (or I can as well) if preferred.


petewarden (approved):

yifeif:

yifeif:

Kangbababa:
I have tested for camer sample.  It 's different result  from load local bmp  with the same modle,when I use moible phone against the same bmp.  Anyone else have this problem?
capture frame data is different  from loading bmp picture data?  how to make the same?

PR #8565, opened 2017-03-20
Type conversion for non-fused, default batch norm #8535

jakelee8:
CLA signed and submitted.

### DIFF
--- a/tensorflow/contrib/layers/python/layers/layers_test.py
+++ b/tensorflow/contrib/layers/python/layers/layers_test.py
@@ -1664,94 +1664,56 @@ class FCTest(test.TestCase):
 
 class BatchNormTest(test.TestCase):
 
-  dtype = None
-
-  @property
-  def _dtype_default(self):
-    return self.dtype or dtypes.float32
-
-  @property
-  def _kw_dtype_default(self):
-    return {'dtype': self._dtype_default}
-
-  @property
-  def _kw_dtype(self):
-    return {'dtype': self.dtype} if self.dtype else {}
-
-  @property
-  def _numpy_dtype_default(self):
-    if self.dtype is None:
-      return np.float32
-    return np.dtype(self.dtype.name)
-
-  @property
-  def _kw_numpy_dtype(self):
-    return {'dtype': self._numpy_dtype_default}
-
   def _addBesselsCorrection(self, sample_size, expected_var):
     correction_factor = sample_size / (sample_size - 1)
     expected_var *= correction_factor
     return expected_var, correction_factor
 
-  def assertAllClose(self, *args, **kwargs):
-    if self.dtype == dtypes.float16:
-      kwargs.setdefault('atol', 1e-3)
-    return super(BatchNormTest, self).assertAllClose(*args, **kwargs)
-
   def testUnknownShape(self):
     with ops.Graph().as_default() as g, self.test_session(g):
-      inputs = array_ops.placeholder(dtype=self._dtype_default)
+      inputs = array_ops.placeholder(dtype=dtypes.float32)
       with self.assertRaisesRegexp(ValueError, 'undefined rank'):
         _layers.batch_norm(inputs)
 
   def testInvalidDataFormat(self):
     with ops.Graph().as_default() as g, self.test_session(g):
-      inputs = array_ops.placeholder(dtype=self._dtype_default)
+      inputs = array_ops.placeholder(dtype=dtypes.float32)
       with self.assertRaisesRegexp(
           ValueError, 'data_format has to be either NCHW or NHWC.'):
         _layers.batch_norm(inputs, data_format='CHWN')
 
   def testUnknownChannelsDimNHWC(self):
     with ops.Graph().as_default() as g, self.test_session(g):
-      inputs = array_ops.placeholder(dtype=self._dtype_default)
+      inputs = array_ops.placeholder(dtype=dtypes.float32)
       inputs.set_shape(tensor_shape.TensorShape((5, 3, 3, None)))
       with self.assertRaisesRegexp(ValueError, 'undefined'):
         _layers.batch_norm(inputs, data_format='NHWC')
 
   def testUnknownChannelsDimNCHW(self):
     with ops.Graph().as_default() as g, self.test_session(g):
-      inputs = array_ops.placeholder(dtype=self._dtype_default)
+      inputs = array_ops.placeholder(dtype=dtypes.float32)
       inputs.set_shape(tensor_shape.TensorShape((5, None, 3, 3)))
       with self.assertRaisesRegexp(ValueError, 'undefined'):
         _layers.batch_norm(inputs, data_format='NCHW')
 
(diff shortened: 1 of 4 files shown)


alextp:
Jenkins, test this please.

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -2553,6 +2641,18 @@ def testNHWCAndNCHWTrainingProduceSameOutput(self):
...
+
+class BatchNormFloat32Test(BatchNormTest):
+  dtype = dtypes.float32
+
+
+class BatchNormFloat64Test(BatchNormTest):
+  dtype = dtypes.float64
+
+
 class LayerNormTest(test.TestCase):
```
alextp: This will make the tests substantially more expensive to run (3x the cost they have now).

I think it's best to triplicate a single representative test to cover all dtypes, and assume that because there's no dtype-specific code in the layers this is enough coverage.

What do you think?

tensorflow/python/layers/normalization_test.py
```diff
@@ -501,18 +515,32 @@ def testNoScale(self):
...
-    inputs = random_ops.random_uniform((5, 4, 3), seed=1)
+    bn = normalization_layers.BatchNormalization(
+      axis=1, gamma_regularizer=reg, **self.kw_dtype)
+    inputs = random_ops.random_uniform((5, 4, 3), seed=1, **self.kw_dtype)
     training = array_ops.placeholder(dtype='bool')
     _ = bn.apply(inputs, training=training)
     self.assertEqual(len(bn.losses), 1)
 
 
+class BNFloat16Test(test.TestCase):
```
alextp: Because these classes inherit from test.TestCase and not from the test above this does nothing.

That said, see my other comment about not triplicating all tests.

alextp (changes requested):
Thanks for the contribution!

The change itself looks good, but I think the tests add much computational cost at little benefit. See my comment about triplicating a single test instead of the entire suite.

tensorflow/contrib/layers/python/layers/layers_test.py
jakelee8: I originally updated the tests as you suggested. However, when I ran all the test cases for float16, one important test failed. I think float16 is not supported due to lack of precision? Description of the issue from #8535 reproduced below.

I do agree that 3x the cost is quite substantial, especially since most people use float32. Though compared to how long it took to build Tensorflow, the cost seem negligible, but that's not to say I advocate always increasing testing costs.

Would you think that only enabling full combinatorial testing before general releases to a more reasonable approach? Given how many changes I needed to parameterize BatchNormTest, it would likely take a good chunk of time to support optional combinatorial testing with types.

---

There seems to be an issue for supporting `float16` concerning moment calculation. Non-fused BatchNorm uses `moments()` internally and `moments()` has this caveat. Except from source code follows.

```py
    # The dynamic range of fp16 is too limited to support the collection of
    # sufficient statistics. As a workaround we simply perform the operations
    # on 32-bit floats before converting the mean and variance back to fp16
    y = math_ops.cast(x, dtypes.float32) if x.dtype == dtypes.float16 else x
```

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_impl.py#L631-L634


Running `layers_test.py` with support for `float16` and `float32` BatchNorm resulted in the following numerical errors. The `float16` moving mean values are incorrect.

alextp: Let's take that test itself and run for float16, then, leaving the others untouched?

jakelee8: Sounds good.

alextp: Cool. Please ping when you've updated the code :-)

jakelee8:
Updated

alextp:
Jenkins, test this please.

alextp:
Can you addess the test failures?

jakelee8:
I believe the failures are due to lack of precision for float16.

```python
    # The dynamic range of fp16 is too limited to support the collection of
    # sufficient statistics. As a workaround we simply perform the operations
    # on 32-bit floats before converting the mean and variance back to fp16
    y = math_ops.cast(x, dtypes.float32) if x.dtype == dtypes.float16 else x
```

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_impl.py#L631-L634

Fix options:

- Lower numerical tolerance when testing for float16; hides practical limits of float16 for batch norm. 👎 
- Use float32 for running stats; converting float32 to float16 to apply batch norm still has problems with precision. 👎 
- ???

Suggestions?

P.S. Low priority pull request: I stopped using float16 in favor of smaller models/batches w/ multi-gpu setup. Works 💯 for fully automated lung segmentation!

alextp:
I think lowering the numeric tolerance for float16 tests is fine here; I think we need to experiment more to find out how to make best use of batch norm for float16 but we don't need to block this PR on it.

jakelee8:
Updated, but normalization_test still fails, this time with nan errors for the float16 gradient. Normalization error is also nearly 1 with float16, quite substantial an error. Thoughts?

alextp:
This to me sounds like getting this to work well with float16 will be more subtle than simply updating the dtypes. Maybe it's worth going back to float32 and changing each individual tensor to float16 to see which ones lead to breakage, and seeing then if it's possible to go around that somehow?

jakelee8:
Unfortunately, simply changing a subset of the variable types is not a simple fix either.

The original issue was that when the batch norm variables did not match the input tensor type, e.g. `float16` inputs vs `float32` variables, Tensorflow would throw a type error. We've seen that simply using `float16` runs into numerical precision issues.

Casting `float32` to `float16` has issues with numerical precision as well. I don't know how truncating `float32` to `float16` would work out, seems like a rabbit hole I'd rather avoid...

For reference, 4 variables are defined for `batch_norm`:

- `beta`
- `gamma`
- `moving_mean`
- `moving_variance`

P.S. Implementing evolutionary learning as described by Salimans et al. (2017) sounds more promising in terms of the original goal to make more GPU memory available to build larger models. Salimans et al. estimated in their [talk](https://www.technologyreview.com/s/603921/elon-musks-openai-unveils-a-simpler-way-for-machines-to-learn/) that not calculating gradients means models take up to 1/3 the memory. This is 33% better memory efficiency that a conversion to float16 would offer, assuming float16 models take 50% less memory than float32 (`1. - .33/.5 = .33`).

In terms of performance, the Salisman et al. implementation scales linearly with additional nodes, something Tensorflow already support.

martinwicke:
Ok, @jakelee8 @alextp, what do we do about this PR?

alextp:
I think as it is we should close it. It seems like simply changing the dtypes in batchnorm to float16 is not the right solution to make it work well with float16, so it might require more thought.

jakelee8:
Closing as per discussion.

PR #8580, opened 2017-03-21
modify annotations of data_format of average_pooling2d in file python/layers/pooling.py 

lyuwenyu:
Modified wrong annotation of data_format in average_pooling2d in file python/layers/pooling.py

from (batch, height, channels, width) to (batch, height, width, channels) when 'channels_last'


lyuwenyu:
I signed it!

martinwicke (approved):

martinwicke:
Jenkins, test this please.

PR #8629, opened 2017-03-22
explicitly assign to dtype for tf.Variable code

zhaojunz:
dtype is not the second parameter in tf.Variable, though the type is inferred correctly by the initial values.


martinwicke (approved):

martinwicke:
Thank you!

Issue #3638, opened 2016-08-04
tf.gather produces zeros for invalid indices on GPU


hholst80:
### Environment info

Operating System: Ubuntu 16.04 LTS (64 bit)

```
$ dpkg -l | grep cuda | grep ^ii
ii  libcuda1-361                                361.42-0ubuntu2                                             amd64        NVIDIA CUDA runtime library
ii  libcudart7.5:amd64                          7.5.18-0ubuntu1                                             amd64        NVIDIA CUDA Runtime Library
ii  libcudnn5                                   5.0.5-1+cuda7.5                                             amd64        cuDNN runtime libraries
ii  libcudnn5-dev                               5.0.5-1+cuda7.5                                             amd64        cuDNN development libraries and headers
ii  libcudnn5-doc                               5.0.5-1+cuda7.5                                             amd64        cuDNN documents and samples
ii  nvidia-cuda-dev                             7.5.18-0ubuntu1                                             amd64        NVIDIA CUDA development files
ii  nvidia-cuda-doc                             7.5.18-0ubuntu1                                             all          NVIDIA CUDA and OpenCL documentation
ii  nvidia-cuda-gdb                             7.5.18-0ubuntu1                                             amd64        NVIDIA CUDA Debugger (GDB)
ii  nvidia-cuda-toolkit                         7.5.18-0ubuntu1                                             amd64        NVIDIA CUDA development toolkit
```

```
$ find /usr/lib -name libcud\*
/usr/lib/i386-linux-gnu/libcuda.so.1
/usr/lib/i386-linux-gnu/libcuda.so.361.42
/usr/lib/i386-linux-gnu/libcuda.so
/usr/lib/x86_64-linux-gnu/libcudnn_static.a
/usr/lib/x86_64-linux-gnu/libcudnn_static_v5.a
/usr/lib/x86_64-linux-gnu/libcuda.so.1
/usr/lib/x86_64-linux-gnu/libcudart.so.7.5.18
/usr/lib/x86_64-linux-gnu/libcudnn.so
/usr/lib/x86_64-linux-gnu/libcuda.so.361.42
/usr/lib/x86_64-linux-gnu/libcudnn.so.5.0.5
/usr/lib/x86_64-linux-gnu/libcudart.so
/usr/lib/x86_64-linux-gnu/libcudart.so.7.5
/usr/lib/x86_64-linux-gnu/libcudadevrt.a
/usr/lib/x86_64-linux-gnu/libcudnn.so.5
[...]

PR #8632, opened 2017-03-22
Doc addition to tf.gather describing validate_indices option behavior

Fenugreek:
See Issue #3638 . This helps people debugging their code, as otherwise they may assume `tf.gather` throws an error when invalid indices are present (it currently does so only when executed on CPU).


Fenugreek:
I've been trying to get Googlebot to recognize me as the one who signed the CLA. Been juggling email addresses both at github and Google; apparently so far not working. Will try again later. Any tips appreciated.

PR #8636, opened 2017-03-22
[CMake] Fix for building tools when GPU support is enabled

vit-stepanovs:
Before this change, benchmark_model would not build with GPU support enabled.

### DIFF
--- a/tensorflow/contrib/cmake/CMakeLists.txt
+++ b/tensorflow/contrib/cmake/CMakeLists.txt
@@ -202,7 +202,6 @@ endif()
 
 # Let's get to work!
 include(tf_core_framework.cmake)
-include(tf_tools.cmake)
 # NOTE: Disabled until issue #3996 is fixed.
 # include(tf_stream_executor.cmake)
 if (tensorflow_ENABLE_GPU)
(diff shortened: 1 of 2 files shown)


gunan:
Jenkins, test this please.
@vit-stepanovs Would you like me to test this on a windows-gpu machine, or have you already taken care of that?

gunan:
mac tests failed due to a bad machine.
Just to verify,
Jenkins, test this please.

vit-stepanovs:
@gunan I tested this locally with GPU support enabled and it worked... It would not hurt to still test it as part of the check in (note that most tools are probably not built by CI, so it would only help ensure the core is not broken after the change).

gunan:
I had to take most of our windows machines offline due to required updates to base images.
I will merge this then reach out to you if I see any failures.

gunan (commented):
Should we also add tests under tensorflow/tools in tf_tests.cmake?
Looks like tf_test_src_py rule there does not look into tools directory at all.

gunan:
http://ci.tensorflow.org/job/tf-pr-win-cmake-gpu/13/

gunan (approved):

gunan:
windows cmake issue was a missing package we forgot to reinstall during maintenance.
Reinstalled now.

PR #8638, opened 2017-03-22
Support pandas Series and DataFrame as input for tf.constant

### LINKED ISSUE #2318: tf.constant should support pandas Series and DataFrame as input
Currently, it does not.

I would suggest implementing this by checking for a `__array__` method, which is the [standard API](http://docs.scipy.org/doc/numpy/reference/arrays.classes.html#numpy.class.__array__) used for indicating that an object is coercible into NumPy arrays, and is implemented by nearly every library that implements "array like" objects in the Python ecosystem. This includes pandas, as well as many other widely used libraries, including xarray and dask.array.

yongtang:
This fix tries to address the proposal raised in #2318 to support pandas Series and DataFrame as input for tf.constant.

As was explained in #2318, Series and DataFrame has `__array__` method, which is the standard API used for indicating that an object is coercible into NumPy arrays.

This fix adds the support for objects with `__array__` to be served as input for tf.constant.

This fix fixes #2318.

### DIFF
--- a/tensorflow/python/framework/tensor_util_test.py
+++ b/tensorflow/python/framework/tensor_util_test.py
@@ -663,6 +663,32 @@ class TensorUtilTest(test.TestCase):
     self.assertFalse(tensor_util.ShapeEquals(t, [1, 4]))
     self.assertFalse(tensor_util.ShapeEquals(t, [4]))
 
+  def testMockArray(self):
+    class MockArray(object):
+      def __init__(self, array):
+        self.array = array
+
+      def __array__(self, dtype=None):
+        return np.asarray(self.array, dtype)
+
+    ma = MockArray(np.array([10, 20, 30]))
+    t = tensor_util.make_tensor_proto(ma)
+    if sys.byteorder == "big":  
+      self.assertProtoEquals("""  
+        dtype: DT_INT64  
+        tensor_shape { dim { size: 3 } }  
+        tensor_content: "\000\000\000\000\000\000\000\\n\000\000\000\000\000\000\000\024\000\000\000\000\000\000\000\036"  
+        """, t)  
+    else:  
+      self.assertProtoEquals("""
+        dtype: DT_INT64
+        tensor_shape { dim { size: 3 } }
+        tensor_content: "\\n\000\000\000\000\000\000\000\024\000\000\000\000\000\000\000\036\000\000\000\000\000\000\000"
+        """, t)
+    a = tensor_util.MakeNdarray(t)
+    self.assertEquals(np.int64, a.dtype)
+    self.assertAllClose(np.array([10, 20, 30], dtype=np.int64), a)
+
   def testSeries(self):
     df = pd.Series([10, 20, 30])
     t = tensor_util.make_tensor_proto(df)
--- a/tensorflow/python/framework/tensor_util_test.py
+++ b/tensorflow/python/framework/tensor_util_test.py
@@ -19,11 +19,11 @@ from __future__ import division
 from __future__ import print_function
 
 import numpy as np
-import pandas as pd
 import sys
 
 from tensorflow.python.framework import constant_op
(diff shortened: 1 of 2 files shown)


tensorflow/python/framework/tensor_util_test.py
```diff
@@ -662,6 +663,43 @@ def testShapeEquals(self):
     self.assertFalse(tensor_util.ShapeEquals(t, [1, 4]))
     self.assertFalse(tensor_util.ShapeEquals(t, [4]))
 
+  def testSeries(self):
+    df = pd.Series([10, 20, 30])
```
shoyer: I would suggest testing this with a dummy class that implements `__array__` rather than actually using pandas.

yongtang:
@shoyer Thanks for the review. The PR has been updated with added test case of `testMockArray` which has the dummy class that implements `__array__`. Please take a look.

shoyer:
Just to be clear, I'm Googler and TensorFlow contributor, but not an owner for this part of the code. Somebody else will triage this PR and assign it for review.

tensorflow/python/framework/tensor_util_test.py
```diff
@@ -19,6 +19,7 @@
 from __future__ import print_function
 
 import numpy as np
+import pandas as pd
```
martinwicke: We can't have unguarded pandas imports -- you can try to import it -- see pandas imports in contrib/learn

martinwicke: Actually, please see @shoyer's comment below. Please remove pandas from this.

tensorflow/python/framework/tensor_util_test.py
```diff
@@ -662,6 +663,69 @@ def testShapeEquals(self):
...
+        dtype: DT_INT64
+        tensor_shape { dim { size: 3 } }
+        tensor_content: "\\n\000\000\000\000\000\000\000\024\000\000\000\000\000\000\000\036\000\000\000\000\000\000\000"
+        """, t)
+    a = tensor_util.MakeNdarray(t)
+    self.assertEquals(np.int64, a.dtype)
+    self.assertAllClose(np.array([10, 20, 30], dtype=np.int64), a)
+
+  def testDataFrame(self):
+    df = pd.DataFrame({'A': [10], 'B': [20], 'C': [30]})
```
alextp: Let's test only a generic object with the __array__ method and comment that this supports pandas.

tensorflow/python/framework/tensor_util_test.py
```diff
@@ -662,6 +663,69 @@ def testShapeEquals(self):
...
 
+  def testMockArray(self):
+    class MockArray(object):
+      def __init__(self, array):
+        self.array = array
+
+      def __array__(self, dtype=None):
+        return np.asarray(self.array, dtype)
+
+    ma = MockArray(np.array([10, 20, 30]))
```
alextp: Can you make these arrays smaller (maybe single-element) so the proto will fit in the test?

tensorflow/python/framework/tensor_util_test.py
```diff
@@ -662,6 +663,69 @@ def testShapeEquals(self):
...
+    t = tensor_util.make_tensor_proto(ma)
+    if sys.byteorder == "big":
```
alextp: Actually, instead of using tensor proto, which relies on big vs little endian, how about just session.run the return value of ops.convert_to_tensor and comparing with the original .__array__ result?

alextp (changes requested):
Thanks for the contribution!

I like the added support; I just have concerns about the tests which should be easy to address.

yongtang:
Thanks @martinwicke @alextp for the review. The PR has been updated as suggested. Please take a look.

martinwicke:
Jenkins, test this please.

alextp (approved):

PR #8646, opened 2017-03-23
Doc addition to tf.gather describing validate_indices option behavior

Fenugreek:
See Issue #3638 . This helps people debugging their code, as otherwise they may assume `tf.gather` throws an error when invalid indices are present (it currently does so only when executed on CPU).


girving (approved):

girving:
Jenkins, test this please.

girving:
@Fenugreek Thank you for the fix!

PR #8653, opened 2017-03-23
Add the sched_getaffinity support for Android.

qjivy:
Now the NumSchedulableCPUs() always chooses to  return 4 cores for Android, other than try the Bionic supported sched_getaffinity to get the actual available cores. Here I remove the "!defined(__ANDROID__)" macro and just keep the " __linux__" macro to work for both pure-linux and android-linux platform. 
The reason why we can drop the "__ANDROID__" marco is that the GNU toolchain would always has the system-specific predefined macros "__linux__"  setting to 1.

### DIFF
--- a/tensorflow/core/platform/posix/port.cc
+++ b/tensorflow/core/platform/posix/port.cc
@@ -21,7 +21,7 @@ limitations under the License.
 #include "tensorflow/core/platform/logging.h"
 #include "tensorflow/core/platform/mem.h"
 #include "tensorflow/core/platform/types.h"
-#if defined(__linux__) && !defined(__ANDROID__)
+#if defined(__linux__)
 #include <sched.h>
 #endif
 #include <stdio.h>
@@ -48,7 +48,7 @@ string Hostname() {
 }
 
 int NumSchedulableCPUs() {
-#if defined(__linux__) && !defined(__ANDROID__)
+#if defined(__linux__) 
   cpu_set_t cpuset;
   if (sched_getaffinity(0, sizeof(cpu_set_t), &cpuset) == 0) {
     return CPU_COUNT(&cpuset);


qjivy:
1. Information about the GNU toolchain's "system-specific predefined macros" can be get from:\
[https://gcc.gnu.org/onlinedocs/cpp/System-specific-Predefined-Macros.html](url)

2. How can we infer that the __linux__ is always set to 1 by the GNU cpp?

(1) gcc-4.9/gcc/c-family/c-cppbuiltin.c: in c_cpp_builtins()
`1041   TARGET_OS_CPP_BUILTINS ();`

(2) gcc-4.9/gcc/config/aarch64/aarch64-linux-android.h: 
`#undef TARGET_OS_CPP_BUILTINS
#define TARGET_OS_CPP_BUILTINS()                \
  do                                            \
    {                                           \
        GNU_USER_TARGET_OS_CPP_BUILTINS();      \
        ANDROID_TARGET_OS_CPP_BUILTINS();       \
    }                                           \
  while (0)
`

(3) gcc-4.9/gcc/config/linux.h:
`#define GNU_USER_TARGET_OS_CPP_BUILTINS()                       \
    do {                                                        \
        if (OPTION_GLIBC)                                       \
          builtin_define ("__gnu_linux__");                     \
        builtin_define_std ("linux");                           \
        builtin_define_std ("unix");                            \
        builtin_assert ("system=linux");                        \
        builtin_assert ("system=unix");                         \
        builtin_assert ("system=posix");                        \
    } while (0)
`

(4) gcc-4.9/gcc/c-family/c-cppbuiltin.c: in builtin_define_std() routine, the "linux" macro will be converted to "__linux__"

qjivy:
Maybe pete@petewarden.com<mailto:pete@petewarden.com> can help to review.
Thanks!


-------------------------------------------------------------------------------------------------------------------------------------
THIS E-MAIL INCLUDING ANY ATTACHMENTS CONTAINS SPREADTRUM’S PROPRIETARY CONFIDENTIAL INFORMATION THAT IS HIGHLY CONFIDENTIAL AND PRIVILEGED.

andrewharp (approved):

andrewharp:

andrewharp:
Please don't merge this yet; while testing this internally I discovered a build issue. Looking into it now.

drpngx:
@andrewharp working now?

andrewharp:
@drpngx Not yet. We need to replicate sched_getaffinity() as the function isn't present in the NDK/API combination we use internally for building.

martinwicke:
@andrewharp is this good to go now?

andrewharp:
@martinwicke No, it can't go in as-is since we need a workaround internally. sched_getaffinity() isn't provided by our Android NDK so we need to replicate the system calls it uses. I'll update once I get a chance to properly test the fix.

vrv:
I talked to Andrew offline and it sounds like due to internal API differences, this change sort of has to be completely done differently, and Andrew has decided to take this on, so I'm going to close this PR.

I wish I had a better answer for you, but the TF team doesn't control the NDK versions that are built inside all of Google, so the best we can do is do the work required to solve this problem for you.

PR #8662, opened 2017-03-23
Support self_adjoint_eig to complex64 and complex128

### LINKED ISSUE #5190: tf.self_adjoint_eig does not work for complex matrices
The documentation of tf.self_adjoint_eig (and the likes) does not state that which data types are accepted by this function. Turns out, only float32 and float64. But the name is misleading: self-adjointness is a property of complex matrices, and one could expect that this function would work on complex64 and complex128 as well.

For me it would be very useful if tf.self_adjoint_eig could work for complex matrices.

Please amend the documentation with the list of accepted dtypes (this goes for SVD and other similar functions as well). And of course it would be great if tf.self_adjoint_eig could accept complex data. :)

yongtang:
This fix adds the complex64 and complex128 support for `self_adjoint_eig`, as was raised in #5190.

This fix fixes #5190.

### DIFF
--- a/tensorflow/python/kernel_tests/self_adjoint_eig_op_test.py
+++ b/tensorflow/python/kernel_tests/self_adjoint_eig_op_test.py
@@ -78,7 +78,7 @@ def _GetSelfAdjointEigTest(dtype_, shape_):
         low=-1.0, high=1.0, size=n * n).reshape([n, n]).astype(dtype_)
     a += a.T
     a = np.tile(a, batch_shape + (1, 1))
-    if dtype_ == np.float32:
+    if dtype_ == np.float32 or dtype_ == np.complex64:
       atol = 1e-4
     else:
       atol = 1e-12
@@ -150,13 +150,14 @@ def _GetSelfAdjointEigGradTest(dtype_, shape_):
 
 
 if __name__ == '__main__':
-  for dtype in np.float32, np.float64:
+  for dtype in np.float32, np.float64, np.complex64, np.complex128:
     for size in 1, 2, 5, 10:
       for batch_dims in [(), (3,)] + [(3, 2)] * (max(size, size) < 10):
         shape = batch_dims + (size, size)
         name = '%s_%s' % (dtype.__name__, '_'.join(map(str, shape)))
         setattr(SelfAdjointEigTest, 'testSelfAdjointEig_' + name,
                 _GetSelfAdjointEigTest(dtype, shape))
-        setattr(SelfAdjointEigGradTest, 'testSelfAdjointEigGrad_' + name,
-                _GetSelfAdjointEigGradTest(dtype, shape))
+        if dtype in [np.float32, np.float64]:
+          setattr(SelfAdjointEigGradTest, 'testSelfAdjointEigGrad_' + name,
+                  _GetSelfAdjointEigGradTest(dtype, shape))
   test.main()
--- a/tensorflow/core/kernels/self_adjoint_eig_v2_op.cc
+++ b/tensorflow/core/kernels/self_adjoint_eig_v2_op.cc
@@ -81,7 +81,15 @@ class SelfAdjointEigV2Op : public LinearAlgebraOp<Scalar> {
 
 REGISTER_LINALG_OP("SelfAdjointEigV2", (SelfAdjointEigV2Op<float>), float);
 REGISTER_LINALG_OP("SelfAdjointEigV2", (SelfAdjointEigV2Op<double>), double);
+REGISTER_LINALG_OP("SelfAdjointEigV2", (SelfAdjointEigV2Op<complex64>),
+                   complex64);
+REGISTER_LINALG_OP("SelfAdjointEigV2", (SelfAdjointEigV2Op<complex128>),
+                   complex128);
 REGISTER_LINALG_OP("BatchSelfAdjointEigV2", (SelfAdjointEigV2Op<float>), float);
 REGISTER_LINALG_OP("BatchSelfAdjointEigV2", (SelfAdjointEigV2Op<double>),
                    double);
+REGISTER_LINALG_OP("BatchSelfAdjointEigV2", (SelfAdjointEigV2Op<complex64>),
+                   complex64);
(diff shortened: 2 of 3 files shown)


drpngx:
Jenkins, test this please.

tensorflow/python/kernel_tests/self_adjoint_eig_op_test.py
```diff
@@ -78,7 +78,7 @@ def Test(self):
         low=-1.0, high=1.0, size=n * n).reshape([n, n]).astype(dtype_)
     a += a.T
     a = np.tile(a, batch_shape + (1, 1))
-    if dtype_ == np.float32:
+    if dtype_ != np.float64:
```
martinwicke: complex128 too?

rmlarsen: Yup, I'd do

if dtype_ == np.float32 or dtype_ == np.complex64:

yongtang: Thanks @martinwicke @rmlarsen for the review. The PR has been updated. Please take a look.

martinwicke:
Jenkins, test this please.

Itaydal:
Hi @martinwicke, great contribution! just ran into this issue myself few days ago.
I've compiled fork with you'r fix, and I think I've found some strange problem,
When i try to use tf.self_adjoint_eig over a large tensor of type tf.float64 (large is [2e7, 4, 4]), the work gets distributed between my 8 CPU cores. 
But when do the same with tensor of type tf.complex64 (with the same values) tf uses only single CPU core.
What could be the cause of this problem?

Test code:

```
def eig_test(H, w):
    H_tens = tf.constant(H[:, :, np.newaxis], dtype=tf.float64) 
    w_tens = tf.constant(w, dtype=tf.float64)
    H_tens = tf.multiply(w_tens, H_tens)
    H_tens = tf.transpose(H_tens, perm=[2, 0, 1])
    
    (e_tens, v_tens) = tf.self_adjoint_eig(H_tens)
    
    sess = tf.Session()
    return sess.run(e_tens)

H = np.array([[1, 2, 3, 4], [2, 1, 5, 6], [3, 5, 1, 7], [4, 6, 7, 1]])
w = np.random.rand(20000000)
    
start_time = time()
res = eig_test(H, w)
print 'duration', time() - start_time
```

Test results:
with type tf.complex64 - test duration was 45 sec
with type tf.float64 - test duration was 10 sec

tensorflow/python/kernel_tests/self_adjoint_eig_op_test.py
```diff
@@ -150,13 +150,14 @@ def Test(self):
...
+  for dtype in np.float32, np.float64, np.complex64, np.complex128:
     for size in 1, 2, 5, 10:
       for batch_dims in [(), (3,)] + [(3, 2)] * (max(size, size) < 10):
         shape = batch_dims + (size, size)
         name = '%s_%s' % (dtype.__name__, '_'.join(map(str, shape)))
         setattr(SelfAdjointEigTest, 'testSelfAdjointEig_' + name,
                 _GetSelfAdjointEigTest(dtype, shape))
-        setattr(SelfAdjointEigGradTest, 'testSelfAdjointEigGrad_' + name,
-                _GetSelfAdjointEigGradTest(dtype, shape))
+        if dtype is [np.float32, np.float64]:
```
jheymann85: This effectively disables all gradient checks. I guess it should be `if dtype in [np.float32, np.float64]:` ?
But despite that, shouldn't the complex gradients also be tested?

rmlarsen: Please fix is -> in. It would be preferable to test all gradients. You might need to fix the gradient computation for complex types. We've had a few bugs like that. I'm guessing that the transpose here:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/linalg_grad.py#L215

should be a conjugate transpose, but I'm not sure.

rmlarsen (changes requested):

tensorflow/python/kernel_tests/self_adjoint_eig_op_test.py
jheymann85: Yes, that should fix the gradient.
It might be worth discussing if the gradient should be hermitian instead of the lower triangular matrix, since the user provides a hermitian matrix as the input and the fact that only the lower triangular is used is an implementation detail of the OP. E.g. if the input was a variable it would not be hermitian anymore after applying the gradient.

rmlarsen: @jheymann85 I agree that we should consider changing the gradients of the functions manipulating Hermitian matrices to return the a Hermitian gradient. But that should be done separately from this change, and for all such method in a single update.

rmlarsen:
@yongtang would you be able to make the requested changes to fix the tests?

yongtang:
Thanks @rmlarsen for the reminder. I will address the issue shortly.

rmlarsen:

drpngx:
@rmlarsen good to go?

rmlarsen:
@drpngx @yongtang I thought @yongtang was going to fix/enable the gradient tests for complex? I guess we can merge this and address it in a followup.

rmlarsen (approved):
We can address complex gradients in a followup.

drpngx:
@rmlarsen you are right. @yongtang any chance you can address this?

rmlarsen:
@drpngx Ooops, I missed your update. I agree that it would be better to address before merging.

yongtang:
Thanks @rmlarsen @drpngx for the review. I am actively looking to fixing the complex issue. I tried to fix the https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/linalg_grad.py#L215
as mentioned in the review comment.
However, it seems there are other places that needs to be fixed as well and I am still working on it.

For that I think it probably make sense to merge this one first. I can create another follow up PR to address the complex gradients issue once I figure out.

rmlarsen:
@yongtang thanks for looking into this. It makes sense to do it in a followup. 
@drpngx feel free to squash & merge.

drpngx:
SG, thanks guys!

PR #8668, opened 2017-03-23
get_attr returns dtype objects instead of raw ints

### LINKED ISSUE #447: Python get_attr on DT_FLOAT returns proto type, not python DType
If my graph has a Sub operation, and I get the "T" key for the op, it always returns an integer type incorrectly.

```
op: "Sub"
input: "ResizeBilinear"
input: "Sub/y"
attr {
  key: "T"
  value {
    type: DT_FLOAT
  }
}
```

```
print(ops[8].get_attr("T"))
1

print(type(ops[8].get_attr("T")))
int
```

Mycosynth:
Fixes #447.

get_attr is documented to return Python types.  It was returning
raw integers (enumeration values) for 'type' attributes.  Now, those
are converted to dtype objects.


martinwicke:
@vrv there may be API to consider, but it may be ok given conversion logic.

vrv:
Adding josh, who has thought more about types in TF in general.

josh11b:
I think this is a good idea in principle (looks like it was an oversight that we didn't do that earlier when we introduced the dtype classes) as long as we have some confidence that the automatic conversion logic for dtypes means this won't break users. My expectation is that not many users are using this API (main exception would be a few gradient functions, but those are being moved to C++ anyway), so there is likely very little breakage introduced.

tensorflow/python/framework/ops.py
```diff
@@ -1522,12 +1522,18 @@ def get_attr(self, name):
     if x.HasField("list"):
       for f in fields:
         if getattr(x.list, f):
-          return list(getattr(x.list, f))
+          if f == "type":
```
vrv: Can you add a note to RELEASE.md indicating the change in behavior?

Something like:

"In python, "Operation.get_attr" on type attributes now returns the Python DType version of the type rather than the protobuf enum."

vrv (dismissed):
Approval after making the one comment in RELEASE.md, this seems like a useful change, thanks!

Mycosynth:
Noobie at github. I made the requested changes to RELEASE.md , but a merge conflict was noted on that file in this pull request. I pulled the current tensorflow master branch and addressed the conflict, and pushed the change to my forked branch.  I'm not certain that I've done that right, though. Any feedback would be appreciated.

martinwicke:
I resolved the conflict. Thank you for the contribution!

martinwicke:
Jenkins, test this please.

martinwicke (approved):

drpngx:
Jenkins, test this please.

drpngx:
Jenkins, test this please.

martinwicke:
So, those test failures are a bad omen. Clearly we cannot assign this return type to an int, specifically, not an int which is a field of a proto. I think this sinks this PR, since that's too common a thing to currently do to be able to break this behavior.

Maybe it's possible to repair by improving automatic conversion. @Mycosynth do you want to give it a try?

vrv:
I think this PR could be saved if we added the following to DType:

def __int__(self):
    return self.as_datatype_enum

So that implicit conversion to int would happen.

This seems safe to do for a few reasons:

- Callers invoking implicit conversion are naturally expecting to understand the 'enum' of the DType.
- We validate the correctness of the DType enum value during construction, so the return value of the int is always a true DType.

The only potential ambiguity is whether we want to ever return ref types, or their base enum, but since we're moving away from ref types towards resource types, I think it doesn't matter.

@Mycosynth can you try making the change to add an __int__ method to DType?  If you don't want to, let us know and we can take a look first.

vrv:
@Mycosynth do you think you will be able to do this soon?  (Trying to close old PRs that haven't had any progress in a while)

Mycosynth:
No. I will not be able to get to it soon. Feel free to take a look at it.
Apologies on the late response.

vrv:
Okay, I'll close this then, and maybe I'll try the __int__ approach (and give you credit on this if we end up submitting it).

Issue #3001, opened 2016-06-22
Use Basic neural network subroutines (BNNS) on iOS


cancan101:
Take advantage of the [BNNS library that Apple announced for iOS](https://developer.apple.com/reference/accelerate/1912851-bnns).

aselle:
Thanks for suggesting this. At this time we probably do not have the bandwidth to look into it immediately, so I am marking it as a contributions welcome.

khurram18:
I want to work and enhance using BNNS on iOS. Can anybody suggest from where to start.

damienstanton:
I'd like to help here. I have a well-defined use case in a personal project. 

When I have time to look into it, I'll track on the `bnns` branch on my fork, [here](https://github.com/damienstanton/tensorflow/tree/bnns).

hanleyweng:
These sample projects might help: 
https://developer.apple.com/library/content/samplecode/MPSCNNHelloWorld/Introduction/Intro.html#//apple_ref/doc/uid/TP40017482
https://developer.apple.com/library/content/samplecode/MetalImageRecognition/Introduction/Intro.html#//apple_ref/doc/uid/TP40017385
https://github.com/shu223/iOS-10-Sampler

For custom weights / own models, perhaps this might help: https://www.bignerdranch.com/blog/use-tensorflow-and-bnns-to-add-machine-learning-to-your-mac-or-ios-app/

damienstanton:
I had to jump away from this for a while due to other obligations  but would love to take a look again. Thanks for the links @hanleyweng

paiv:
I demoed MNIST on BNNS with TensorFlow model here: https://github.com/paiv/mnist-bnns

BNNS is somewhat limited in features. One should make performance comparison before committing.

cancan101:
@paiv Wondering if you saw any issues like this: http://stackoverflow.com/questions/42641668/using-basic-neural-network-subroutines-bnn-accelerate

cancan101:
Core ML might be an interesting abstraction to both this and BNNS: https://github.com/tensorflow/tensorflow/issues/10468.

tensorflowbutler:
It has been 14 days with no activity and this issue has an assignee.Please update the label and/or status accordingly.

tensorflowbutler:
Please remove the assignee, as this issue is inviting external contributions. Otherwise, remove the `contributions welcome` label. Thank you.

bhack:
[...]

PR #8673, opened 2017-03-23
BNNS Based Implementations for MaxPool and Conv2d

cancan101:
Fixes https://github.com/tensorflow/tensorflow/issues/3001

I wanted to get this up for feedback, comments, etc. I know the files are not in the right places.
I called it `Conv2DBNNS` but perhaps `BNNSConv2D` is a better name?

I also have another PR on top of this that adds an attr `filter_layout` to allow pre-transposing the weights.

### DIFF
--- a/tensorflow/core/user_ops/BUILD
+++ b/tensorflow/core/user_ops/BUILD
@@ -1,0 +1,32 @@
+load("//tensorflow:tensorflow.bzl", "tf_custom_op_library", "cc_header_only_library")
+
+cc_header_only_library(
+    name = "conv_2d",
+    deps = [
+        "//tensorflow/core/kernels:conv_2d",
+    ],
+)
+
+cc_header_only_library(
+    name = "pooling_ops_hdrs",
+    deps = [
+        "//tensorflow/core/kernels:pooling_ops_hdrs",
+    ],
+)
+
+cc_header_only_library(
+    name = "cast_ops_hdrs",
+    deps = [
+        "//tensorflow/core/kernels:cast_op",
+    ],
+)
+
+tf_custom_op_library(
+    name = "bnns.so",
+    srcs = ["bnns_conv.cc", "bnns_maxpool.cc"],
+    deps = [
+        ":conv_2d",
+        ":pooling_ops_hdrs",
+        ":cast_ops_hdrs",
+    ],
+)
--- a/tensorflow/core/user_ops/bnns_conv.cc
+++ b/tensorflow/core/user_ops/bnns_conv.cc
@@ -258,15 +258,16 @@ public:
         Tensor transformed_filter;
         {
             OP_REQUIRES_OK(context, context->allocate_temp(
-                                                           DataTypeToEnum<T>::value,
-                                                           TensorShape({out_depth, in_depth,
-                filter_rows,
-                filter_cols}),
-                                                           &transformed_filter));
+                DataTypeToEnum<T>::value,
+                TensorShape({out_depth, in_depth,
+                    filter_rows,
+                    filter_cols}),
+                &transformed_filter));
 
             functor::TransformFilter<Device, T, int, 4>()(
-                                                          context->eigen_device<Device>(), To32Bit(filter.tensor<T, 4>()),
-                                                          To32Bit(transformed_filter.tensor<T, 4>()));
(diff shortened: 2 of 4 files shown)


rmlarsen:
Sorry for not responding sooner.
@petewarden would you or somebody in your team have time to look at this? Thanks.

petewarden (commented):
Thanks for putting this together, it is very interesting. It looks like it's still in an experimental phase right now? Here are a few suggestions for the direction to head in:

 - It would be preferable not to add new ops, but instead reuse the existing Conv2D and MaxPool ops (though I know this won't use the activation functions). You can look at how we use the Accelerate framework in conv_ops_using_gemm.cc to get an idea of what I mean.

 - To understand how this improves performance compared to the current implementation, we'd need some benchmark numbers. To generate these, you could use some of our pretrained models like Inception v1 and v3 within the iOS benchmark app.

Thanks again for creating this, I'm hoping we can work together to get something ready to check in!

cancan101:
Thanks for the review and comments. A few thoughts:

As far as the (new) fused `Conv2D` Op that I am using, I used the [MKL Conv2D Op](https://github.com/tensorflow/tensorflow/blob/904edee4456a61d50d5b1ffe9858a7772acc423e/tensorflow/core/ops/nn_ops.cc#L2632-L2645) as a predicate. In that case Intel has defined a new Op that includes both the convolution Op and the BiasAdd Op. My Conv Op actually isn't pulling in just the activation function, rather, my Op and the MKL both have the bias add has been fused as well.

I understand that the primitive `Conv2D` op in TF does not currently handle activations or bias addition, but this op may be "too primitive". In fact cuDNN is now moving in the direction of fusing these three into one kernel (see: https://github.com/tensorflow/tensorflow/issues/8828). Perhaps, a different FusedConv2D Op should be added that can be used by not just BNNS but also cuDNN v6 and any other underlying platform implementations that can take advantage of doing all three ops together. This would be  along the lines of batch norm, which now offers a `fused` version of the Op to reduce the number of primitive Ops used.

The secondary benefit of this Op is it allows for BNNS specific optimization flags, for example the level of precision (float16 vs float32) and control over the `filter_layout` (I already have code that can avoid having to transpose the weights on each call).

As far as testing, I have tested locally with some models, the one thing to keep in mind is that these two new Ops work with NCHW so the models do have to work with that channel layout.

drpngx:
@petewarden what do you think?

martinwicke:
@petewarden ping?

petewarden:
Sorry for the slow response on this one, that's my fault. We've been having some internal debate about the best way to support these kind of fused ops, since this is coming up more often. I like your proposal to create a general fused conv op that can have multiple implementations. I ended up doing something similar for fusing ResizeBilinear and MirrorPad into Conv2D:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/nn_ops.cc#L732

Ideally I'd love to see a version where the channel order was an attribute, even if the standard order was unimplemented for these cases. It would also be great if we could add a transform to automatically spot this pattern and replace it in graphs, the way I did here for the resize fusion:
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms#fuse_convolutions

As a start though, would you be willing to put together a FusedConv2DWithBiasAndActivation op proposal? Apologies again for taking so long to get back on this.

cancan101:
What do you mean by channel order, the order of the input / output tensors dimensions or the order of the filter dimensions? Or something else?

vrv:
@cancan101 I believe he means the "data_format" (e.g., NCHW).

I believe the concrete suggestion is:

- Add an op that has data_format as an optional attribute, whose default is "NHWC". 
- In the op, check the data_format and do an OP_REQUIRES if the format isn't NCHW.  All tests would explicitly pass in NCHW as the data_format.

vrv:
(Any updates @cancan101 ?)

drpngx:
@cancan101 any luck with this?

cancan101:
I haven't had much time to look at generalizing the approach.

rmlarsen:
@cancan101 marking this as stalled for now. Please ping this thread when you have a chance to get back to it.

drpngx:
@cancan101 please re-open when you get cycles on this. Thanks!

PR #8677, opened 2017-03-24
Added Windows JNI download to README

Rabrg:
The Java README is missing the download link for the Windows JNI download


asimshankar:
@Rabrg : Thanks very much for the PR. 

However, if you don't mind, I'd like to abandon this PR as I have another change that is updating the same file with both Windows and Maven instructions. There are a few issues with Maven and RC0 on Windows that I've discovered that I'd like to fix.

FWIW though, that URL is correct for now so feel free to use it.

PR #8714, opened 2017-03-25
Fix typo at description of tf.nn.conv1d

tae-jun:
As you see [description of tf.nn.conv1d](https://www.tensorflow.org/versions/master/api_docs/python/tf/nn/conv1d), `batch, out_width, out_channels` is linked, because markdown recognizes it as link. [batch, out_width, out_channels]\(...\)

So, add escape character to parentheses!


yifeif (approved):
Thanks for sending this fix @tae-jun!

yifeif:

PR #8716, opened 2017-03-25
Fix tf.nn.softmax

### LINKED ISSUE #8667: tf.nn.softmax -- the "name" argument is passed to the internal flattened tensor
Hi, if I use tf.nn.softmax with the `name` argument, I expect the output tensor to be given this name.
However, the output tensor is always a Reshape op named `Reshape_...`, and the passed name refers to an internal (flattened) tensor.

Steps to reproduce (TF v1.0.1):
```
In [1]: import tensorflow as tf
In [2]: a = tf.nn.softmax(tf.ones([2, 8, 20]), name="softmax_custom_name")
In [3]: print a
Tensor("Reshape_1:0", shape=(2, 8, 20), dtype=float32)
In [4]: print tf.get_default_graph().get_tensor_by_name("softmax_custom_name:0")
Tensor("softmax_custom_name:0", shape=(16, 20), dtype=float32)
```

Here is the relevant code from the master branch:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_ops.py#L1501
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_ops.py#L1519

Is this the expected behaviour? In this case it would probably make sense to mention it in the documentation?

Otherwise I'd submit a fix.

ilya-eder:
Fixes #8667

### DIFF
--- a/tensorflow/python/ops/nn_ops.py
+++ b/tensorflow/python/ops/nn_ops.py
@@ -1474,14 +1474,14 @@ def _softmax(logits, compute_op, dim=-1, name=None):
     InvalidArgumentError: if `logits` is empty or `dim` is beyond the last
       dimension of `logits`.
   """
-  def _swap_axis(logits, dim_index, last_index):
+  def _swap_axis(logits, dim_index, last_index, name=None):
     """Swaps logits's dim_index and last_index."""
     return array_ops.transpose(logits,
                                array_ops.concat([
                                    math_ops.range(dim_index), [last_index],
                                    math_ops.range(dim_index + 1, last_index),
                                    [dim_index]
-                               ], 0))
+                               ], 0), name=name)
 
   logits = ops.convert_to_tensor(logits)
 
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/nn_ops.py
```diff
@@ -1523,7 +1523,7 @@ def _swap_axis(logits, dim_index, last_index):
   # shape.
   output.set_shape(shape)
 
-  return output
+  return tf.identity(output, name=name)
```
skye: Instead of adding an identity op, how about plumbing name through _swap_axis() to array_ops.tranpose() above?

drpngx:
Jenkins, test this please.

@ilya-edrenkin any chance to take a look at the review comments?

tensorflow/python/ops/nn_ops.py
ilya-edrenkin: sure!

skye (approved):

drpngx:
Jenkins, test this please.

PR #8728, opened 2017-03-26
PollZmqOp for network communication

### LINKED ISSUE #4836: ZeroMQ Operator
Currently (it seems like) the encouraged way to read data is either through the provided reader/producer ops which are not quite flexible, or through python feed_dict which is slower.

Is there a plan to support something like a ZeroMQReaderOp, which keeps reading and deserializing tensors from a zmq socket and produce them? This way we can use any comfortable way to process data perhaps in an independent process, and send them through zmq for training.

tillahoffmann:
This PR adds `PollZmqOp` for retrieving data over the network (fixes #4836, ref #7951) . A few points for discussion:

* ZeroMQ is LGPL but has some exceptions which seem to allow static linking. Is that a problem?
* The `Compute` method of the Op is ensured to be thread safe by blocking. This could be improved by having a dictionary of sockets keyed by thread id or a connection pool. However, I have never needed concurrent calls to `Compute` in my work.
* I'm not very familiar with `bazel` so there may be some problems in the dependencies.

This op allows for distributed training data generation to minimize the CPU workload on the training machine. For example, one or more machines can generate training data which can be retrieved by a training machine using the op. An example can be found in the test.


yifeif:
Thanks @tillahoffmann! @mrry assign to you because you are working on the new input pipeline. Feel free to reassign as you see fit. Thank you :)

tillahoffmann:
Here is a small example on how to learn the mean of a distribution by generating data in two separate processes and consuming the data in a single process.

```python
# consumer.py
import tensorflow as tf

# Define parameters
addresses = ["tcp://localhost:5555", "tcp://localhost:5556"]
batch_size = 100
min_after_dequeue = 1000
queue_capacity = min_after_dequeue + 3 * batch_size

with tf.Graph().as_default():
    # Create a queue to hold the data
    queue = tf.RandomShuffleQueue(
        queue_capacity,
        min_after_dequeue,
        [tf.float32], 
        [[]],
        ['x']
    )
    
    # Add zmq ops
    enqueue_ops = []
    for address in addresses:
        raw = tf.poll_zmq(b"", address, 1000)
        decoded = tf.decode_raw(raw, tf.float32)
        enqueue_ops.append(queue.enqueue_many({'x': decoded}))
        
    # Add a queue runner for the enqueue ops
    qr = tf.train.QueueRunner(queue, enqueue_ops)
    tf.train.add_queue_runner(qr)
    
    # Get the data
    data = queue.dequeue_many(batch_size)
    # Compute the loss
    mean = tf.Variable(0, dtype=tf.float32)
    loss = tf.reduce_mean((data['x'] - mean) ** 2)
    
    # Add optimization op
    train_op = tf.train.AdamOptimizer(1).minimize(loss)
    
    # Setup
    sess = tf.Session()
    tf.train.start_queue_runners(sess)
    
# Run the training
losses = []
for _ in range(100):
    _, loss_val = sess.run([train_op, loss])
    losses.append(loss_val)
    
print(sess.run(mean))
```

```python
# producer.py
import zmq
import sys
import numpy as np

# Get the port to listen on and define the ground truth mean
port = int(sys.argv[1])
mean = 15

# Create a socket and listen
context = zmq.Context()
socket = context.socket(zmq.REP)
socket.bind("tcp://*:%d" % port)

# Respond to requests
while True:
    message = socket.recv()
    socket.send(np.random.normal(mean, 1, 1000).astype(np.float32).tostring())
```

RealTimeDeployment:
Help someone is hacking me and played me for the idiot I am. I was trying to learn how to write code. I thought I was doing something good I didn't know it did this. I was trying to do some good. My reference was to thinking it had been done wrong. I was wrong. I pulled the cla when I realized what it did immediately and they are driving me crazy. I reset my phone to give off a warning and hoped they would leave me alone but they still won't stop. I was only trying to figure out an equation. I never wrote any code. I thought I was talking to Nvidia and tensor flow. There was a device listed on my account that isn't mine.

mrry:
@tillahoffmann Thanks for submitting this PR! I have some thoughts about the interface, but many of these could be postponed by moving the implementation to `tf.contrib` while we iterate on the design.

* For most new stateful ops, we try to separate the state from the op-kernels that act on it. This would entail splitting the op into two: a "constructor" op that sets up the ZMQ socket/context and adds an object to the `ResourceMgr`, and an "accessor" op that retrieves the object from the `ResourceMgr` and invokes a method on it. (For examples, see the implementations of the readers, queues, variables, etc.) The benefits of doing this are clearer when there are multiple implementations of the same state (e.g. a similar RabbitMQ or Kafka resource? Or maybe various RPC protocols?) or multiple "accessor" ops that access the same state, and it would be good to understand if we'd like to support such things before adding this to the core library.

* I assume that the `zmq_msg_recv()` call can block. This would block one of the (finitely many) TF thread pool threads, and could lead to deadlock. Typically we would implement such an operation as an `AsyncOpKernel`. I'm not familiar enough with ZMQ to know the best way to receive a message asynchronously, but hopefully there is a standard pattern that we could use here.

(Regarding the licensing question, I don't know what the consequences are, but @martinwicke probably does (or knows someone who does). Regarding the post from @RealTimeDeployment, I don't know what we can do to help on this thread, but I'd suggest opening a new issue if you continue to have problems.)

tillahoffmann:
@mrry, thanks for the feedback. I'm happy to leave this PR open and iterate on the design here unless you have a strong preference to merge into `tf.contrib` first.

1. I'll have a look into splitting the op into a state and an accessor as suggested (do you have a particular state/op combination in mind that is most closely related?). For now, I will hold off until we've heard from @martinwicke regarding the legal aspects.
2. Yes, `zmq_msg_recv` does indeed block. Once 1. is addressed, I will move the accessor to an async op.

martinwicke:
Yeah, let's wait until we have clarity on the licensing issues. 

It would be our preference to merge into contrib first if at all possible.

tillahoffmann:
Sure, I'll make the appropriate changes for merging into contrib once we know more regarding licensing.

martinwicke:
The ZMQ dependency is fine. 

Let's move this to contrib, and we can iterate on details there.

Thanks!

tillahoffmann:
The original PR is now migrated to `contrib`. Note that I haven't really got my head around `bazel` yet so the `BUILD` files are copied together from different modules in `contrib`. They can likely be condensed.

@mrry do you want to merge the existing interface into `contrib` first and then move to the constructor/accessor design in a subsequent PR?

drpngx:
(OK, just FYI I checked internally and ZMQ licensing is fine.)

tensorflow/contrib/io/kernels/zmq_op.cc
```diff
@@ -0,0 +1,108 @@
...
+  ~PollZmqOp() {
+    if (socket_ != nullptr) {
+      zmq_close(socket_);
+    }
+    if (context_ != nullptr) {
+      zmq_ctx_destroy(context_);
+    }
+  }
+
+  using AsyncOpKernel::AsyncOpKernel;
```
mrry: I *think* this `using` statement is a no-op and can be deleted.

tensorflow/contrib/io/kernels/zmq_op.cc
```diff
@@ -0,0 +1,108 @@
...
+                errors::Internal("Failed to initialize context."));
+    socket_ = zmq_socket(context_, ZMQ_REQ);
+    OP_REQUIRES(context, socket_ != nullptr,
+                errors::Internal("Failed to initialize socket."));
+    OP_REQUIRES(context,
+                zmq_setsockopt(socket_, ZMQ_RCVTIMEO,
+                               &timeout_, sizeof(timeout_)) == 0,
+                errors::Internal("Failed to set timeout."));
+    OP_REQUIRES(context, zmq_connect(socket_, &address_[0]) == 0,
+                errors::Internal("Failed to connect to ", address_, "."));
```
mrry: Can this block? Should it happen in `ComputeAsync()` on the first use instead?

tensorflow/contrib/io/kernels/zmq_op.cc
```diff
@@ -0,0 +1,108 @@
...
+    OP_REQUIRES(context, zmq_msg_init(&reply_msg) == 0,
+                errors::Internal("Failed to initialize reply."));
+    OP_REQUIRES(context, zmq_msg_recv(&reply_msg, socket_, 0) != -1,
+                errors::Internal("Failed to receive reply within ", timeout_, "ms."));
+    // Copy the data and clean up
+    reply.resize(zmq_msg_size(&reply_msg));
+    memmove(&reply[0], zmq_msg_data(&reply_msg), zmq_msg_size(&reply_msg));
+    OP_REQUIRES(context, zmq_msg_close(&reply_msg) == 0,
+                errors::Internal("Failed to close reply message."));
+    done();
```
mrry: Since the `done()` callback is always* called synchronous, this is effectively a synchronous `OpKernel`, and the `AsyncOpKernel` is just adding complexity for no gain.

However, since this is a polling method and there's no loop, I assume that `zmq_msg_recv()` can block. We discourage writing kernels that block a thread-pool thread because there is only a finite number of them and blocking operations leads straightforwardly to deadlock.

Are there asynchronous variants of the `zmq_*()` methods that could call the `done()` callback on completion? Or would it make sense for this kernel to own a thread that performs the blocking iterations on behalf of the op thread-pool, and executes `done()` when a message arrives?

(* In fact, if any of the `OP_REQUIRES()` macros fails, `done()` will never be called and the execution will hang. It's important (though not well documented) that you use `OP_REQUIRES_ASYNC()` and `OP_REQUIRES_OK_ASYNC()` in a `ComputeAsync()` method.)

tensorflow/contrib/io/kernels/zmq_op.cc
```diff
@@ -0,0 +1,108 @@
...
+                errors::InvalidArgument(
+                    "Input message tensor must be scalar, but had shape: ",
+                    input->shape().DebugString()));
+
+    Tensor* output = nullptr;
+    OP_REQUIRES_OK(context, context->allocate_output("reply",
+                                                      TensorShape({}), &output));
+
+    // Get the message and response as string scalars
+    auto request = input->scalar<string>()();
```
mrry: Can this be `const auto&`?

tensorflow/contrib/io/ops/io_ops.cc
```diff
@@ -0,0 +1,52 @@
...
+  return Status::OK();
+}
+
+}  // namespace
+
+REGISTER_OP("PollZmq")
+    .Input("request: string")
+    .Output("reply: string")
+    .Attr("address: string")
+    .Attr("timeout: int = -1")
```
mrry: I assume this defaults to an infinite timeout, but it would be good to add that to the docs.

tensorflow/contrib/io/ops/io_ops.cc
```diff
@@ -0,0 +1,52 @@
...
+  }
+  return Status::OK();
+}
+
+}  // namespace
+
+REGISTER_OP("PollZmq")
+    .Input("request: string")
+    .Output("reply: string")
+    .Attr("address: string")
```
mrry: What format is the address argument? Can you add descriptions of this and the other arguments to the `.Doc()` string?

tensorflow/contrib/io/python/kernel_tests/io_ops_test.py
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for tensorflow.contrib.io."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import zmq
```
mrry: I think this might require a new dependency on our test infrastructure. @yifeif Do you know how to make this happen? (I'm assuming `pip install zmq` will do it.)

tensorflow/contrib/io/ops/io_ops.cc
```diff
@@ -0,0 +1,52 @@
+/* Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
mrry: Copyright 2017.

third_party/libzmq.BUILD
```diff
@@ -0,0 +1,241 @@
+package(default_visibility = ["//visibility:public"])
+
+licenses(["notice"])  # LGPL license with exception for static linking (for libzmq)
+
+genrule(
+    name = "configure",
+    outs = ["platform.hpp"],
+    cmd = "cd external/libzmq && ./configure && cd ../..; cp external/libzmq/src/platform.hpp $@",
+)
```
mrry: This might be problematic for the Windows/Bazel build... I wonder if this file can be used on Windows:

https://github.com/zeromq/libzmq/blob/master/builds/msvc/platform.hpp

(using the `if_windows()` macro to change the command to copy that file to `$@`)

tensorflow/contrib/io/kernels/zmq_op.cc
mrry: It might be fine to delay refactoring this until the constructor/accessor is in place (since e.g. a resource for the ZMQ socket would be the logical owner for a background thread), but please either switch back to a synchronous kernel or fix the `OP_REQUIRES()` invocations.

mrry (changes requested):
Sorry for the delay on this! Thanks for the `tf.contrib` move. A few comments based on the latest code....

tensorflow/contrib/io/kernels/zmq_op.cc
tillahoffmann: Unfortunately, letting ZMQ handle the callback doesn't seem to be easy. Let's postpone that to a separate PR if you're happy with it.

tensorflow/contrib/io/kernels/zmq_op.cc
tillahoffmann: [`zmq_connect`](http://api.zeromq.org/4-2:zmq-connect) does not usually block. From the docs:

> [F]or most transports and socket types the connection is not performed immediately but as needed by ØMQ. [...] The first exception is when using the inproc:// transport: you must call zmq_bind() before calling zmq_connect(). The second exception are ZMQ_PAIR sockets, which do not automatically reconnect to endpoints.

third_party/libzmq.BUILD
tillahoffmann: Not sure what the best thing to do here is. Do you build on windows using MSVC? The [`configure`](https://github.com/zeromq/libzmq/blob/3f80657c247bf14ba29d6de3cf8aeea8d0f1bc87/configure.ac) script of libzmq seems pretty sophisticated and I don't know whether just copying the file is sufficient. I'm afraid I don't have a windows machine to test it.

tillahoffmann:
@mrry, long-term it may be worthwhile thinking about mirroring python's object-oriented ZMQ interface to give users maximum flexibility. Thoughts?

drpngx:
Jenkins, test this please.

drpngx:
@tillahoffmann there is a problem in the build file. Could you fix (and maybe try `bazel build -c opt tensorflow/...`)?

```
ERROR: /workspace/tensorflow/contrib/io/BUILD:36:1: no such package 'tensorflow/contrib/io/kernels': BUILD file not found on package path and referenced by '//tensorflow/contrib/io:io_ops_kernels'.
ERROR: Analysis of target '//tensorflow/contrib/io:io_ops_kernels' failed; build aborted.
```

vrv:
@tillahoffmann I think given the extra zmq dependency, and the lack of clarity for exactly how the core system would use this, I'd prefer to keep the core code slim.

Because this is a nicely self-contained op, this is actually a great candidate to release / deploy as a custom, external operation via load_library.  If this op becomes popular, we can then assess the tradeoff of adding it to the core vs. the additional dependency / portability issues it creates.  It'll hopefully also be more clear why this op is even useful (and for example, why we wouldn't want an alternative to zeromq instead).  Without more context / use patterns, it's hard to know :).

I'm going to close this for now, but feel free to ping me or open an issue to discuss this further.

tillahoffmann:
@vrv, thanks for the input. I think there is already a desire for network communication as part of tensorflow (cf. #7951, #4836). In particular, network io ops

* enable decentralised data preprocessing in arbitrary languages and technologies.
* provide an interface with external systems (in particular for reinforcement learning) such as game emulators, robots, etc.

As you mentioned, the dependency is a bit of an issue. Do you have other basic communication libraries in mind that would provide similar functionality?

Given @mrry's comments regarding proper resource management, I'm happy to leave this PR closed. Let me know whether there is appetite for proper ZMQ integration. Otherwise, we will manage network communication externally as you suggested.

Issue #8724, opened 2017-03-26
Error occurred in Tensorflow version 1.0.1 date: 26th March 2017.


mrnameless123:
I just reinstalled Tensorflow on my laptop this morning and got this problem. 
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "BestSplits" device_type: "CPU"') for unknown op: BestSplits
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "CountExtremelyRandomStats" device_type: "CPU"') for unknown op: CountExtremelyRandomStats
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "FinishedNodes" device_type: "CPU"') for unknown op: FinishedNodes
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "GrowTree" device_type: "CPU"') for unknown op: GrowTree
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "ReinterpretStringToFloat" device_type: "CPU"') for unknown op: ReinterpretStringToFloat
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "SampleInputs" device_type: "CPU"') for unknown op: SampleInputs
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "ScatterAddNdim" device_type: "CPU"') for unknown op: ScatterAddNdim
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "TopNInsert" device_type: "CPU"') for unknown op: TopNInsert
E c:\tf_jenkins\home\workspace\release-win\device\cpu\os\windows\tensorflow\core\framework\op_kernel.cc:943] OpKernel ('op: "TopNRemove" device_type: "CPU"') for unknown op: TopNRemove
[...]

PR #8737, opened 2017-03-27
[CMake] Enabling CPU optimization with MSVC

Carmezim:
Updating CMake docs to add clarity about how to enable advanced CPU instructions 
with MSVC.

Reference: [comment and following ones](https://github.com/tensorflow/tensorflow/issues/8724#issuecomment-289313811)

cc @mrry


yifeif:
@mrry Derek could you take a look? Thank you!

tensorflow/contrib/cmake/README.md
```diff
@@ -165,6 +165,29 @@ Step-by-step Windows build
 
 3. Invoke CMake to create Visual Studio solution and project files.
 
+   **NOTE**: MSVC does not support [the flag `-march=native` defaulted on `CMakeLists.txt`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/CMakeLists.txt#L78) 
+   so in order to build TensorFlow with MSVC using advanced CPU instructions
+   you will need to edit `CMakeLists.txt` and explicitly define the options
+   as in `/arch:[ARG1|ARG2|ETC]`, e.g.:
```
gunan: The instruction is confusing I think.
What if we did this to avoid the need for users to edit the cmakefile:

In Cmakefiles.txt, add a new option name can be something like "tensorflow_WIN_CPU_SIMD_OPTIONS", default value to this can be empty.
Then we append this to to the MSVC compilation commands if the build is running on windows.

With the above change, our instructions become this: Just set `-Dtensorflow_WIN_CPU_SIMD_OPTIONS=/arch:AVX`

What do you think?

Carmezim: You are right, it turned out way too complex. That would be ideal, great suggestion

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -81,6 +82,18 @@ if (tensorflow_OPTIMIZE_FOR_NATIVE_ARCH)
   endif()
 endif()
 
+# MSVC SIMD instructions
```
Carmezim: Do you rather add it in an `else()` in the same closure of the `arch=native` or leave it separate like it is?

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -81,6 +82,18 @@ if (tensorflow_OPTIMIZE_FOR_NATIVE_ARCH)
...
+if (tensorflow_WIN_CPU_SIMD_OPTIONS)
+  if (WIN32)
+    CHECK_CXX_COMPILER_FLAG("${tensorflow_WIN_CPU_SIMD_OPTIONS}" COMPILER_OPT_WIN_CPU_SIMD_SUPPORTED)
+    if(COMPILER_OPT_WIN_CPU_SIMD_SUPPORTED)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${tensorflow_WIN_CPU_SIMD_OPTIONS}")
+    else()
+      message("${tensorflow_WIN_CPU_SIMD_OPTIONS} not supported")
```
Carmezim: remove message?

mrry: Perhaps it should be a `message(FATAL_ERROR ...)`?

tensorflow/contrib/cmake/README.md
```diff
@@ -181,7 +181,16 @@ Step-by-step Windows build
...
    More? -DCUDNN_HOME="D:\...\cudnn"
    ```
-    
+   To enable SIMD instructions with MSVC, as AVX and SSE, define them as follows:
+   ```
+   More? -Dtensorflow_WIN_CPU_SIMD_OPTIONS=/arch:AVX
+   ```     
+   or if more than one as in:
+   ```
+   More? -Dtensorflow_WIN_CPU_SIMD_OPTIONS=/arhc:[AVX2|SSE2]
```
mrry: "arhc" should be "arch" here :).

The syntax reads weirdly to me... `[AVX2|SSE2]` implies (to me) that you can specify `/arch:AVX2` *or* `/arch:SSE2`, but not both together. (It's weird to see a pipe symbol in the middle of a flag, for one thing....) Does it actually work to put `/arch:[AVX2|SSE2]`?

mrry (changes requested):
Thanks for adding the new configuration option! I had a few nits and clarifications....

tensorflow/contrib/cmake/README.md
Carmezim: Oops, it was very late, will fix the typo :)

I had the same reading of this syntax. When looking at [1] it seems that is how you define the set of instructions but apparently you are right. I tested on x64 which only supports AVX and AVX2 and using both it yelled at me. Would you please confirm it does not work?

I will change it anyway leaving only the first option `/arch:ARG`, what do you think?

[1] https://msdn.microsoft.com/en-us/library/7t5yh4fd.aspx

tensorflow/contrib/cmake/CMakeLists.txt
Carmezim: Good catch

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ if (tensorflow_WIN_CPU_SIMD_OPTIONS)
     if(COMPILER_OPT_WIN_CPU_SIMD_SUPPORTED)
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${tensorflow_WIN_CPU_SIMD_OPTIONS}")
     else()
-      message("${tensorflow_WIN_CPU_SIMD_OPTIONS} not supported")
+      message(FATAL_ERROR "${tensorflow_WIN_CPU_SIMD_OPTIONS} not supported")
```
Carmezim: I noticed if you you insert a wrong value as an unsupported instruction to this option when building it will always give you an error on further build attempts regardless if the argument was the supported instruction requiring to delete the build folder and starting again.

tensorflow/contrib/cmake/README.md
```diff
@@ -181,15 +181,10 @@ Step-by-step Windows build
    More? -Dtensorflow_ENABLE_GPU=ON ^
    More? -DCUDNN_HOME="D:\...\cudnn"
    ```
-   To enable SIMD instructions with MSVC, as AVX and SSE, define them as follows:
+   To enable SIMD instructions with MSVC, as AVX and SSE, define it as follows:
    ```
    More? -Dtensorflow_WIN_CPU_SIMD_OPTIONS=/arch:AVX
```
Carmezim: Any changes here?

mrry (approved):
Looks good to me, thanks!

gunan:
Jenkins, test this please

Issue #8742, opened 2017-03-27
Broken Link in README.md


seventhmoon:
In https://github.com/tensorflow/tensorflow/tree/master/tensorflow/examples/android/README.md,
the link of _Get the recommended Bazel version listed in os_setup.html_ which is pointing to https://www.tensorflow.org/versions/master/get_started/os_setup#source is broken.

PR #8747, opened 2017-03-27
Fix broken link in Android example README

jamesmishra:
Issue  #8742 identified that the link to os_setup.html does not work for the master version.

os_setup.html is still available for old versions ( https://www.tensorflow.org/versions/r0.11/get_started/os_setup#prepare_environment_for_linux ), but it makes more sense to link directly to the Bazel installation instructions.

### DIFF
--- a/tensorflow/examples/android/README.md
+++ b/tensorflow/examples/android/README.md
@@ -83,7 +83,7 @@ instead.
 Bazel is the primary build system for TensorFlow. To build with Bazel,
 it and the Android NDK and SDK must be installed on your system.
 
-1. Get the recommended Bazel version listed in [os_setup.html](https://www.tensorflow.org/versions/master/get_started/os_setup.html#source)
+1. Install the latest version of Bazel as per the instructions [on the Bazel website](https://bazel.build/versions/master/docs/install.html).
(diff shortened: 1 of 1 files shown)


jamesmishra:
I just signed the CLA.

yifeif:
@andrewharp do we always recommend the latest bazel version for android?

andrewharp:
@yifeif for Android in particular I'm unaware of a specific reason currently not to use the latest version. That's not to say there couldn't be a regression in the future, though.

As long as we verify that the known good version of Bazel works for Android maybe we should prefer that. Is there a new location for that info?

BrianOn99:
Actually I compiled the android sample with bazel 0.4.4 from homebrew last week and failed (I forget the error message).  Upgrade to 0.4.5 solved the problem.

yifeif:
Thanks for the confirmation @andrewharp, and thanks for the info @BrianOn99. One place I can think of is the android nightly build, which we do usually run with the latest bazel. Sounds like recommending the latest bazel works for now. We can update if there is regression in the future.

gunan (approved):

gunan:
Jenkins, test this please.

PR #8748, opened 2017-03-27
Semi sync optimizer

renyi533:
Async SGD is faster than sync SGD but inferior in AUC. 

This optimizer wrapper would like to provide a balance between speed and AUC. Each worker mainly works in async mode. But after certain number of batches (usually n_of_batch> worker_count but not necessarily), workers should sync together to mitigate the stale gradient issue.

Based on our test, this is useful for our CTR prediction tasks.

### DIFF
--- a/tensorflow/python/training/semi_sync_replicas_optimizer.py
+++ b/tensorflow/python/training/semi_sync_replicas_optimizer.py
@@ -1,0 +1,311 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


renyi533:
Just signed the CLA.

yifeif:
Thanks for sending the PR@renyi533! Before we can review the change, it looks like the commits were pushed with a different email than the email signed CLA. Could you check? Also is this PR supposed to be for r1.1 branch?

gunan:
We cannot accept this directly into the r1.1 branch.
Please do the following:
1-Make sure your commits use the same email address as the one you signed your Google CLA
2-Apply your changes on top of master branch, and send a pr against master.

I will close this now. Please create a new PR with the above instructions.

PR #8765, opened 2017-03-28
Update README.md for graph transform

raingo:
Do not recommend to use`remove_nodes(op=Identity, op=CheckNumerics)` in the README. It breaks the control flow based operations.

### DIFF
--- a/tensorflow/tools/graph_transforms/README.md
+++ b/tensorflow/tools/graph_transforms/README.md
@@ -94,7 +94,8 @@ transforms to modify the graph with. The transforms are given as a list of
 names, and can each have arguments themselves. These transforms define the
 pipeline of modifications that are applied in order to produce the output.
 Sometimes you need some transforms to happen before others, and the ordering
-within the list lets you specify which happen first. Note that the optimization 
+within the list lets you specify which happen first. 
+Note that the optimization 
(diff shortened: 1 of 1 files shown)


yifeif:
Thanks for sending the PR @raingo. @petewarden mind taking a look? Thanks!

petewarden:
Thanks for the PR! Can you give some more background on the specific issues you're hitting? Control flow is generally only needed for Variable assignment, which generally doesn't happen during inference, so it would be good to know more about what's going wrong in your case.

raingo:
@petewarden For example, for object detection, there is a `tf.image.non_max_suppression` step per each object class. I used `tf.map_fn`to iterate through each class. This iteration is necessary in the deploy time.

drpngx:
@petewarden any modifications we would need here?

petewarden:
Sorry for the slow response on this. Can we address this by leaving the code example but adding documentation along the lines of "this may cause problems if you have a model with control-flow based operations"? I'd hate people to miss out on the optimizations in other cases.

I think the real solution would be to warn or otherwise avoid removing ops that are significant for control flows, but I'm not sure how to do that yet.

vrv:
ping for @raingo !

raingo:
@petewarden PTAL. Just added a note about the control flow ops.

petewarden (approved):
That looks good, thanks!

Issue #8342, opened 2017-03-13
[TensorBoard] load data from relative path with the projector plugin


lspvic:
This is a  problem similar with issue #7382 but it has been closed.

@dandelionmane 
Another place that still has the similar problem in #7382 is the request of project plugin data in Embeddings panel, e.g.  /data/plugin/projector/runs still requests in a absolute path.
I found the code  [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorboard/components/tf_tensorboard/tf-tensorboard.html#L134) , there is a leading slash for the `route-prefix` property in `vz-projector-dashboard`.
If that is correct, I will submit a PR. Thanks!

Another question is when will the tensorboard be recompiled. As now in version 1.0 these features are not included.

dandelionmane:
I migrated this to the new TensorBoard repository. https://github.com/tensorflow/tensorboard/issues/37

PR #8791, opened 2017-03-29
[tensorboard] make projector plugin request data path relative

lspvic:
this PR solves #8342 
As in version 1.1rc0, the changes are made:
>TensorBoard uses a relative data directory, for easier embedding.

but the request of project plugin data in Embeddings panel, e.g. `/data/plugin/projector/runs` still requests in a absolute path.  the code 
```html
<vz-projector-dashboard
            id="projector"
            route-prefix="/data/plugin/projector">
```
with the leading slash leads the  absolutely path request.
this PR remove the leading slash.

UPDATED:
new version of code is:
```new TF.Dashboard.VzProjectorDashboard('/data/plugin/projector'),```
still has a leading slash.


yifeif:
Thanks for the PR @lspvic. @dandelionmane could you take a look?

drpngx:
@dandelionmane how does that look?

teamdandelion (approved):
Looks good to me.

dandelionmane:

PR #8816, opened 2017-03-29
Thread-safe implementations of MKL kernels and rerolling changes lost in previous merge

jbobba:
- Changes to mkl_matmul BUILD rule so it is built as part of the matmul kernel library
- Bug fixes to MKL kernels for thread-safety
- Adding a TF to MKL dimension ordering map to track dimension order of MKL tensors.
- Rerolling changes to MKL layout pass, mkl_util.h, nn_ops.cc and BUILD files that got lost with parallel merges


yifeif:
@jbobba thanks for the PR. Could you resolve the conflicts? @rmlarsen mind taking a look?

martinwicke:
I moved the if_mkl into tf_cc_test_mkl. So that part of the change shouldn't be necessary.

jbobba:
@martinwicke do you have that change coming in from a different branch or should I remove the if_mkl in this PR?

jbobba:
@martinwicke never mind. I checked the definition of tf_cc_test_mkl

andydavis1 (approved):

jbobba:
Can we start the tests on this?

martinwicke:
Jenkins, test this please.

tensorflow/core/kernels/BUILD
```diff
@@ -4403,19 +4410,26 @@ tf_cc_test(
 
 if_mkl(
     tf_kernel_library(
```
martinwicke: Could you make a tf_mkl_kernel_library and pull the if_mkl inside it? Buildozer has a bug which prevents us from inspecting/changing anything inside a top-level if, so pulling the if inside the new rule would make it a lot easier on our sync infra.

martinwicke (changes requested):

martinwicke:
One more thing related to the build.

tensorflow/core/kernels/BUILD
jbobba: Will do.

jbobba: Can you rerun the tests?

martinwicke:
Jenkins, test this please!

martinwicke (approved):

jbobba:
Can we merge this in?

PR #8831, opened 2017-03-30
Add CMake v3.8 as working version

KonstantinKhabarlak:
While CMake 3.7 was causing build errors with GPU ON. I was able to compile Tensorflow (master) with CMake 3.8 with GPU enabled perfectly fine

### DIFF
--- a/tensorflow/contrib/cmake/README.md
+++ b/tensorflow/contrib/cmake/README.md
@@ -45,7 +45,7 @@ bindings.
 
 ### Pre-requisites
 
-* CMake version 3.5 up to 3.6 or 3.8
+* CMake version 3.5 or later.
 
 * [Git](http://git-scm.com)
 
--- a/tensorflow/contrib/cmake/README.md
+++ b/tensorflow/contrib/cmake/README.md
@@ -45,7 +45,7 @@ bindings.
 
 ### Pre-requisites
 
-* CMake version 3.5 up to 3.6
+* CMake version 3.5 up to 3.6 or 3.8
 
 * [Git](http://git-scm.com)
 


KhabarlakKonstantin:
I signed it!

yifeif:
Thanks @KhabarlakKonstantin!

fyi @gunan.  Also looks like cmake 3.8 is still in RC4?

KhabarlakKonstantin:
@yifeif Yes, CMake 3.8 is in RC4
It looks like it has started compiling with CMake 3.7.2. Probably, someone has fixed the issue
Let me recheck it once more and I'll update PR, if it works

gunan:
If the build also works with 3.7.2, let's edit the message to 3.5 or later, but leave a note saying 3.7.0 and 3.7.1 have a bug that breaks our GPU build.

@mrry @guschmue FYI.

guschmue:
I made a change in CMakeLists.txt some time ago and the define that was breaking gpu builds should be gone (there was a define passed to nvcc which broken nvcc but that define was only used to generate version.cc) so I moved it and its no longer passed to nvcc.
All cmake versions better than 3.6 should work now. I use 3.8 on my box and its fine.

tensorflow/contrib/cmake/README.md
```diff
@@ -45,7 +45,7 @@ bindings.
 
 ### Pre-requisites
 
-* CMake version 3.5 up to 3.6
+* CMake version 3.5 up to 3.6 or 3.8
```
gunan: Based on the latest information by @guschmue Let's go with  "Cmake version 3.6 or later"

gunan:
Jenkins, test this please.

gunan (approved):

PR #8637, opened 2017-03-22
Add the corresponding --action_env to .bazelrc from ./configure

damienmg:
This will make the fetch phase reproducible so the refetch does
not happens unless people re-run configure (the environment variable
are stored in the .bazelrc file).

Fixes #8619.

PR #8837, opened 2017-03-30
Windows: Fix cuda_configure.bzl for Bazel build

meteorcloudy:
1. Remove corresponding action_env before writing it
2. Add quote to environment variables
3. Return a dummpy value for gcc detection on Windows

@gunan @damienmg

### DIFF
--- a/configure
+++ b/configure
@@ -46,6 +46,11 @@ function sed_hyphen_i() {
   fi
 }
 
+function write_env_to_bazelrc() {
+  sed_hyphen_i "/$1/d" .bazelrc
+  echo "build --action_env $1=\"$2\"" >> .bazelrc
+}
+
 # Delete any leftover BUILD files from the Makefile build, which would interfere
 # with Bazel parsing.
 MAKEFILE_DOWNLOAD_DIR=tensorflow/contrib/makefile/downloads
@@ -310,7 +315,7 @@ while [[ "$TF_CUDA_CLANG" != "1" ]] && ! is_windows && true; do
   fi
   if [ -e "$GCC_HOST_COMPILER_PATH" ]; then
     export GCC_HOST_COMPILER_PATH
-    echo "build --action_env GCC_HOST_COMPILER_PATH=$GCC_HOST_COMPILER_PATH" >>.bazelrc
+    write_env_to_bazelrc "GCC_HOST_COMPILER_PATH" "$GCC_HOST_COMPILER_PATH"
     break
   fi
   echo "Invalid gcc path. ${GCC_HOST_COMPILER_PATH} cannot be found" 1>&2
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -9,6 +9,7 @@ SOURCE_BASE_DIR=`pwd -P`
 popd > /dev/null
 
 # This file contains customized config settings.
+rm -f .bazelrc
```
damienmg: Why remove it?

damienmg (changes requested):

meteorcloudy:
This will help solve https://github.com/bazelbuild/bazel/issues/2753

configure
meteorcloudy: Otherwise, when you rerun ./configure, .bazelrc file will still contain the content produced by your last run, possibly with different env value.

meteorcloudy: I am assuming ./configure is the only one who touchs .bazelrc, is that ok?

damienmg:
Since you have the sed_hyphen_i lines lower I think the assumption is that
it is not. We strip down the --action_env flag from the file instead of
deleting the file.

The user might do custom change to it.

meteorcloudy:
@damienmg I see, but we didn't actually use `sed_hyphen_i` on `action_env`, now I added the code to strip the flags. Please take a look again.

damienmg:
https://github.com/tensorflow/tensorflow/pull/8637/files#diff-e2d5a00791bce9a01f99bc6fd613a39dR282

Apparently I missed that part when I rolled the change forward. You actually want to remove the action_env that you might set anywhere in that script unconditionally because if the user reconfigure he doesn't want to have stale entry in that file.

Maybe we should just do an "%import .action_env.bazelrc" and wipe that file out and put all the action env in it. That would simplify the sed code.

meteorcloudy:
Sounds good, where should I add the `%import .action_env.bazelrc` ?

damienmg:
top of the .bazelrc?

meteorcloudy:
@damienmg Thanks for the hint, I pushed another change for this.

gunan:
Jenkins, test this please.

third_party/gpus/cuda_configure.bzl
```diff
@@ -35,6 +35,10 @@ _DEFAULT_CUDA_COMPUTE_CAPABILITIES = ["3.5", "5.2"]
 # BEGIN cc_configure common functions.
 def find_cc(repository_ctx):
   """Find the C++ compiler."""
+  # On Windows, we use Bazel's MSVC CROSSTOOL for GPU build
+  # Return a dummy value for GCC detection here to avoid error
+  if _cpu_value(repository_ctx) == "Windows":
+    return "/usr/bin/dummycompiler"
```
gunan: Can we make this name like this, so users can understand what is going on if they use the wrong config to compile on windows?

Something like: `/use/--config x64_msvc_windows/instead`

meteorcloudy: Good idea, done!

gunan:
Change LGTM, Damien, any remaining comments?

gunan (approved):

damienmg (approved):

gunan:
I just added back bazel clean to configure script, so testing one more time to make sure.
Jenkins, test this please.

PR #8844, opened 2017-03-30
Fix doc for svd

ebigelow:
Error where "left" and "right" were reversed


yifeif (approved):
Thanks for the PR @ebigelow!

PR #8862, opened 2017-03-31
MSVC 2015 port.

hughsando:
My version of MSVC 2015 has some issues with some newer c++ constructs.
There are 3 type of fixes:
1. Remove optional 'typename' specifications outside template definitions.  They are optional on newer compilers, but msvc does not thinks so.
2. Remove brackets from initializer lists, `perm({0,1})` becomes `perm{0,1}`. I am not sure why this is different on msvc.
3. Explicitly convert from `Eigen::TensorMap` to `gtl::ArraySlice` via .data() and .size().  Again, I'm not sure why this is required for msvc, but this solves the remaining compiler issues.

### DIFF
--- a/tensorflow/core/framework/tensor.cc
+++ b/tensorflow/core/framework/tensor.cc
@@ -235,7 +235,7 @@ struct ProtoHelper {};
       return proto.N##_val().size();                                   \
     }                                                                  \
     static void Fill(const T* data, size_t n, TensorProto* proto) {    \
-      typename ProtoHelper<T>::FieldType copy(data, data + n);         \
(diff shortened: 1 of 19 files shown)


hughsando:
I signed it!

vit-stepanovs:
@hughsando TensorFlow currently builds with Visual Studio 2015 Update 3 using CMake... Are you using earlier version of Visual Studio 2015 to build?

hughsando:
Yes, I was running Update 1, so that was the problem.
I will close this one - I'm not sure what the best practice use of `typename` is, but there seems to be only a small number of optional uses, and can you can always compiler with Update1 of you want to find them.

PR #8880, opened 2017-03-31
Add back bazel clean into configure script.

gunan:


av8ramit (approved):

gunan:
Jenkins, test this please.

gunan:
Look like there is an issue we missed due to missing bazel clean.
I am looking into root causing this, and solving the problem

Androbin:
Could you please explain briefly, what kind of issue this is?
I would like to know under which circumstances cleaning is actually necessary to recompiling successfully.
This would be good to know because this comes undeniably with some speed penalty.

> Bazel's design is such that these problems are fixable; we consider such bugs a high priority, and will do our best fix them. If you ever find an incorrect incremental build, please file a bug report. We encourage developers to get out of the habit of using `clean` and into that of reporting bugs in the tools.

gunan:
https://github.com/bazelbuild/bazel/issues/2759

Androbin:
@gunan from what the referenced issue says, this should only be an issue on macOS?

gunan:
We first observed this on macos, but then reproduced the same problem on
different platforms as well.

Androbin:
As far as I know, the according fix was already available and now introduced in the recent release of bazel.

PR #8901, opened 2017-04-02
Add gzip and zlib support for FixedLengthRecordReader

### LINKED ISSUE #8865: [FeatureRequest] Decode gzipped files in tf.FixedLengthRecordReader
My dataset is stored in binary format that is easily compressible (from 100 MB down to ~0.5 MB), but I currently cannot make use of this in TF, unless I write a TFRecords file, which is way more cumbersome and leads to a lot of duplication -- I need to work with those files in other programs as well, so the binary format is the easiest solution for me. It would be nice if  tf.FixedLengthRecordReader would support on-the-fly decompression.

yongtang:
This fix adds gzip and zlib support for FixedLengthRecordReader,
as was discussed in #8865.

When FixedLengthRecordReader is used, it will check for
compression_type flag and use ZlibInputStream as needed.
The usage of InputBuffer in FixedLengthRecordReader has
also been changed to BufferedInputStream to match ZlibInputStream.

This fix fixes #8865.

### DIFF
--- a/tensorflow/core/kernels/fixed_length_record_reader_op.cc
+++ b/tensorflow/core/kernels/fixed_length_record_reader_op.cc
@@ -19,7 +19,7 @@ limitations under the License.
 #include "tensorflow/core/framework/reader_base.h"
 #include "tensorflow/core/framework/reader_op_kernel.h"
 #include "tensorflow/core/lib/core/errors.h"
-#include "tensorflow/core/lib/io/inputbuffer.h"
+#include "tensorflow/core/lib/io/buffered_inputstream.h"
 #include "tensorflow/core/lib/strings/strcat.h"
 #include "tensorflow/core/platform/env.h"
 
@@ -41,8 +41,8 @@ class FixedLengthRecordReader : public ReaderBase {
         record_number_(0) {}
 
   // On success:
-  // * input_buffer_ != nullptr,
-  // * input_buffer_->Tell() == header_bytes_
+  // * buffered_inputstream_ != nullptr,
+  // * buffered_inputstream_->Tell() == header_bytes_
   // * file_pos_limit_ == file size - footer_bytes_
   Status OnWorkStartedLocked() override {
     record_number_ = 0;
@@ -50,43 +50,80 @@ class FixedLengthRecordReader : public ReaderBase {
     TF_RETURN_IF_ERROR(env_->GetFileSize(current_work(), &file_size));
     file_pos_limit_ = file_size - footer_bytes_;
 
+    lookahead_cache_.clear();
+
     TF_RETURN_IF_ERROR(env_->NewRandomAccessFile(current_work(), &file_));
 
-    input_buffer_.reset(new io::InputBuffer(file_.get(), kBufferSize));
-    TF_RETURN_IF_ERROR(input_buffer_->SkipNBytes(header_bytes_));
+    buffered_inputstream_.reset(
+        new io::BufferedInputStream(file_.get(), kBufferSize));
+    TF_RETURN_IF_ERROR(buffered_inputstream_->SkipNBytes(header_bytes_));
+    // In case hop_bytes_ is in between 0 and record_bytes_,
+    // we will need to hold a cache so that later on we could prefix the cache
+    // with the remaining data to read:
+    // For example, assume record_bytes is 2 and hop_bytes is 1,
+    // For file "H123",
+    // We will process the data in the following way:
+    // 1. Read hope_bytes of 1 ("1") and save it in hop cache.
+    // 2. Read record_bytes - hop_bytes = 1 ("2")
+    // 3. Prefix data in 2. with hop cache in one, we have "12",
+    //    this is used for the record.
+    // 4. Shift hop_bytes of 1 from the record, and we put "2" to hop cache.
+    // 5. Continue step 2.
+    // In order to acheive the above, we will peek in "record_bytes_ -
+    // hop_bytes_"
+    // before we read the first record.
+    if (0 < hop_bytes_ && hop_bytes_ < record_bytes_) {
+      TF_RETURN_IF_ERROR(buffered_inputstream_->ReadNBytes(
+          record_bytes_ - hop_bytes_, &lookahead_cache_));
+    }
     return Status::OK();
   }
 
   Status OnWorkFinishedLocked() override {
-    input_buffer_.reset(nullptr);
+    buffered_inputstream_.reset(nullptr);
     return Status::OK();
   }
 
   Status ReadLocked(string* key, string* value, bool* produced,
                     bool* at_end) override {
-    // The condition `input_buffer_->Tell() + record_bytes_ > file_pos_limit_`
-    // is to confirm that none of record bytes is out of the range of
-    // file_pos_limit_.
-    // This is necessary for the condition `hop_bytes > 0`. For example.
-    // File: "0123456"
-    // Reader setting: `record_bytes=3`, `hop_bytes=2`, `footer_bytes=0`,
-    //     `header_bytes=0`
-    // Without this checking condition, the forth time the reader will at
-    // this position: "012345|6" and the reading operation will result in
-    // an error.
-    if (input_buffer_->Tell() >= file_pos_limit_ ||
-        input_buffer_->Tell() + record_bytes_ > file_pos_limit_) {
+    // In case hop_bytes_ is in between 0 and record_bytes_,
+    // we will need to hold a cache so that later on we could prefix the cache
+    // with the remaining data to read:
+    // For example, assume record_bytes is 2 and hop_bytes is 1,
+    // For file "H123",
+    // We will process the data in the following way:
+    // 1. Read hope_bytes of 1 ("1") and save it in hop cache.
+    // 2. Read record_bytes - hop_bytes = 1 ("2")
+    // 3. Prefix data in 2. with hop cache in one, we have "12",
+    //    this is used for the record.
+    // 4. Shift hop_bytes of 1 from the record, and we put "2" to hop cache.
+    // 5. Continue step 2.
+    // In order to acheive the above, in the following only
+    // 'record_bytes_ - lookahead_cache_.size()' needs to be read. The
+    // lookahead_cache_
+    // is then prefixed to piece together the whole record.
+
+    int bytes_to_read = record_bytes_;
+    if (0 < hop_bytes_ && hop_bytes_ < record_bytes_) {
+      bytes_to_read = hop_bytes_;
+    }
+    if (buffered_inputstream_->Tell() >= file_pos_limit_ ||
+        buffered_inputstream_->Tell() + bytes_to_read > file_pos_limit_) {
       *at_end = true;
       return Status::OK();
     }
-    const int64 pos_before_read = input_buffer_->Tell();
-    TF_RETURN_IF_ERROR(input_buffer_->ReadNBytes(record_bytes_, value));
+    TF_RETURN_IF_ERROR(buffered_inputstream_->ReadNBytes(bytes_to_read, value));
+    if (0 < hop_bytes_ && hop_bytes_ < record_bytes_) {
+      lookahead_cache_.append(*value, 0, bytes_to_read);
+      *value = lookahead_cache_;
+      lookahead_cache_ = lookahead_cache_.substr(bytes_to_read);
+    }
     *key = strings::StrCat(current_work(), ":", record_number_);
     *produced = true;
     ++record_number_;
 
-    if (hop_bytes_ > 0) {
-      input_buffer_->Seek(pos_before_read + hop_bytes_).IgnoreError();
+    if (hop_bytes_ > record_bytes_) {
+      buffered_inputstream_->SkipNBytes(hop_bytes_ - record_bytes_);
     }
 
     return Status::OK();
@@ -95,7 +132,8 @@ class FixedLengthRecordReader : public ReaderBase {
   Status ResetLocked() override {
     file_pos_limit_ = -1;
     record_number_ = 0;
-    input_buffer_.reset(nullptr);
+    buffered_inputstream_.reset(nullptr);
+    lookahead_cache_.clear();
     return ReaderBase::ResetLocked();
   }
 
@@ -107,11 +145,13 @@ class FixedLengthRecordReader : public ReaderBase {
   const int64 record_bytes_;
   const int64 footer_bytes_;
   const int64 hop_bytes_;
+  string lookahead_cache_;
   Env* const env_;
   int64 file_pos_limit_;
   int64 record_number_;
-  std::unique_ptr<RandomAccessFile> file_;  // must outlive input_buffer_
-  std::unique_ptr<io::InputBuffer> input_buffer_;
+  // must outlive buffered_inputstream_
+  std::unique_ptr<RandomAccessFile> file_;
+  std::unique_ptr<io::InputStreamInterface> buffered_inputstream_;
 };
 
 class FixedLengthRecordReaderOp : public ReaderOpKernel {
--- a/tensorflow/core/kernels/fixed_length_record_reader_op.cc
+++ b/tensorflow/core/kernels/fixed_length_record_reader_op.cc
@@ -37,18 +37,13 @@ class FixedLengthRecordReader : public ReaderBase {
         footer_bytes_(footer_bytes),
         hop_bytes_(hop_bytes),
         env_(env),
-        file_pos_limit_(-1),
         record_number_(0) {}
 
   // On success:
   // * buffered_inputstream_ != nullptr,
   // * buffered_inputstream_->Tell() == header_bytes_
-  // * file_pos_limit_ == file size - footer_bytes_
   Status OnWorkStartedLocked() override {
     record_number_ = 0;
-    uint64 file_size = 0;
-    TF_RETURN_IF_ERROR(env_->GetFileSize(current_work(), &file_size));
-    file_pos_limit_ = file_size - footer_bytes_;
 
     lookahead_cache_.clear();
 
@@ -57,6 +52,9 @@ class FixedLengthRecordReader : public ReaderBase {
     buffered_inputstream_.reset(
         new io::BufferedInputStream(file_.get(), kBufferSize));
     TF_RETURN_IF_ERROR(buffered_inputstream_->SkipNBytes(header_bytes_));
+
+    int bytes_to_read = footer_bytes_;
+
     // In case hop_bytes_ is in between 0 and record_bytes_,
     // we will need to hold a cache so that later on we could prefix the cache
     // with the remaining data to read:
@@ -73,9 +71,10 @@ class FixedLengthRecordReader : public ReaderBase {
     // hop_bytes_"
     // before we read the first record.
     if (0 < hop_bytes_ && hop_bytes_ < record_bytes_) {
-      TF_RETURN_IF_ERROR(buffered_inputstream_->ReadNBytes(
-          record_bytes_ - hop_bytes_, &lookahead_cache_));
+      bytes_to_read += record_bytes_ - hop_bytes_;
     }
+    TF_RETURN_IF_ERROR(
+        buffered_inputstream_->ReadNBytes(bytes_to_read, &lookahead_cache_));
     return Status::OK();
   }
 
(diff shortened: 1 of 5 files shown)


tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -28,45 +31,92 @@ namespace tensorflow {
...
     return Status::OK();
   }
 
   Status ReadLocked(string* key, string* value, bool* produced,
                     bool* at_end) override {
-    if (input_buffer_->Tell() >= file_pos_limit_) {
-      *at_end = true;
-      return Status::OK();
+    if (file_pos_limit_ < 0) {
+      // In case of compressed file (non-seekable), attampt to read
```
Mistobaan: *attempt

tensorflow/core/ops/io_ops.cc
```diff
@@ -442,6 +442,7 @@ REGISTER_OP("FixedLengthRecordReader")
     .Attr("footer_bytes: int = 0")
     .Attr("container: string = ''")
     .Attr("shared_name: string = ''")
+    .Attr("compression_type: string = ''")
```
Mistobaan: I would suggest to use the word "encoding" instead of "compression_type"

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: @Mistobaan Fixed. Thanks.

tensorflow/core/ops/io_ops.cc
yongtang: Done.

yongtang:
@Mistobaan Thanks for the review. The PR has been updated. Please take a look.

drpngx:
@Mistobaan what do you think?

Mistobaan:
@drpngx looks fine to me

drpngx:
@saxenasaurabh please take a look

drpngx:
@saxenasaurabh LMK if you have cycles or we can find another reviewer.

tensorflow/python/kernel_tests/reader_ops_test.py
```diff
@@ -367,14 +367,43 @@ def _CreateFiles(self):
...
+        for j in range(self._num_records):
+          f.write(self._Record(i, j))
+        f.write(b"F" * self._footer_bytes)
+      with open(fn+".tmp", "rb") as f:
+        cdata = zlib.compress(f.read())
+        with open(fn, "wb") as zf:
+          zf.write(cdata)
+    return filenames
+
+  def _TestOneEpoch(self, files, encoding=None):
```
saxenasaurabh: Is the `files` param used? It gets overwritten on the next line.

tensorflow/python/kernel_tests/reader_ops_test.py
```diff
@@ -390,6 +419,18 @@ def testOneEpoch(self):
                                     "\\(requested 1, current size 0\\)"):
         k, v = sess.run([key, value])
 
+  def testOneEpoch(self):
+    files = self._CreateFiles()
+    self._TestOneEpoch(files)
+
+  def testGzipOneEpoch(self, encoding="GZIP"):
```
saxenasaurabh: Change to
```
def testGzipOneEpoch(self):
  files = self._CreateGzipFiles()
  self._TestOneEpoch(files, encoding="GZIP")
```

saxenasaurabh (changes requested):
Thanks for doing this and sorry for the slow review.

yongtang:
Thanks @saxenasaurabh for the review. The reason that `footer_cache_` is needed is that `zlib/gzip` are essentially non-seekable. In other words, once a byte is read, we could not go backward. We can only start from the beginning again to find the previous location. In addition, `zlib/gzip` does not give us the uncompressed file size before hand. (we could certainly do a two-pass but that is likely inefficient if the file is large.)

In case of FixedlengthRecord, a footer might be possible. However, at the time we try to read the next bytes, we don't know if the next bytes belongs to a record, or a footer. Assume the record is `1` bytes and the footer is `2` bytes:
```
HRFF
```
When we read the first `R`, we don't know if it is record, or footer. We only knows it is a record when we read two `FF` + EOF. At this point, we knows that `FF` is the footer, thus `R` can only be record.

For that we maintains a `footer_cache_` so that we could find out the record `R` without seek back the file. (gzip/zlib is non-seekable and we could not find the length before hand).

I will update the PR to add a detailed explanation.

yongtang:
Also, it looks like the format of the `FixedLengthRecord` now accepts `hop_bytes` (added in another PR). This may complicate the processing for non-seekable files (like gzip/zlib).

I will take a closer look and see what we could do to handle `hop_bytes` as well.

vrv:
@yongtang let us know what you find.

yongtang:
Thanks @saxenasaurabh @vrv. I managed to update the PR and the `hop_bytes_` should have been covered as well.

It use similar an approach of caching a buffer for `hop_bytes`.

Basically the purpose is to avoid the need to randomly seek to a position of the file stream. Now the file stream (possibly gzip or zlib) could be "read through" without moving backward. (gzip or zlib does not support moving backward as they are not "seekable").

The comment has been expanded to explain the logic but let me know if more explanation is needed.

yongtang:
I splited the PR into two commits. The first commit switches the InputBuffer to BufferedInputStream only (and drop the need of `Seek()` call). The second commit adds the gzip and zlib process. That may make it easy to follow.

vrv:
@yongtang it looks like somments from @saxenasaurabh haven't been addressed -- can you make sure or at least comment on them?  Thanks!

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -28,45 +31,92 @@ namespace tensorflow {
...
+      // record, and append record to cached footer, then pop head
+      // of cached footer as the record
+      Status s = buffered_input_stream_->ReadNBytes(record_bytes_, value);
+      if (!s.ok() && !errors::IsOutOfRange(s)) {
+        return s;
+      }
+
+      // EOF reached
+      if (value->size() == 0) {
+        file_pos_limit_ = buffered_input_stream_->Tell() - footer_bytes_;
```
yongtang: Thanks @saxenasaurabh. The updated PR includes additional comments for the explanation.

tensorflow/python/kernel_tests/reader_ops_test.py
yongtang: Thanks @saxenasaurabh. That was a typo. The updated PR has fixed this issue by removing the  next line.

yongtang:
@vrv @saxenasaurabh Thanks for the review. The review comments should have been covered. Please take a look.

drpngx:
Jenkins, test this please.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
saxenasaurabh: I still don't understand why you need to set file_pos_limit_ here?

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -38,55 +42,178 @@ class FixedLengthRecordReader : public ReaderBase {
...
+      // Basically, each read only advances hop_bytes_. However, we have the data
+      // that is pieced together by
+      // hop_cache_ + footer_cache_ + read(hop_bytes_)
+      // Then we use the following to cut data for use:
+      // value:                     0 --> record_bytes_
+      // hop_cache_:       hop_bytes_ --> record_bytes_
+      // footer_cache_: record_bytes_ --> end(record_bytes_ + footer_bytes_)
+      // Next time we read, we advance hop_bytes_ and append to the end, and
+      // repeat the same process.
+      if (0 < hop_bytes_ && hop_bytes_ < record_bytes_) {
```
saxenasaurabh: Thanks for intricately handling hop_bytes and footer_bytes. It may be cleaner to combine the two caches into a `lookahead_cache_` which could be of size `record_bytes_ + footer_bytes`. You could use the same algorithm as you do now i.e. return record_bytes_ from lookahead_cache_ -> shift cache forward.

saxenasaurabh (changes requested):

josh11b:
API review is fine with this change in principle, as long as it is backwards compatible (defaults for all new attrs, and those defaults give old behavior).

tensorflow/core/ops/io_ops.cc
```diff
@@ -454,6 +455,7 @@ record_bytes: Number of bytes in the record.
 footer_bytes: Number of bytes in the footer, defaults to 0.
 hop_bytes: Number of bytes to hop before each read. Default of 0 means using
         record_bytes.
+encoding: The type of encoding for the file. Defaults to none.
```
asimshankar: Might be a good idea to elaborate on what valid values of this attribute can be.

drpngx:
@yongtang any progress?

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -38,55 +42,178 @@ class FixedLengthRecordReader : public ReaderBase {
...
-    *produced = true;
-    ++record_number_;
-
-    if (hop_bytes_ > 0) {
-      input_buffer_->Seek(pos_before_read + hop_bytes_).IgnoreError();
+    if (file_pos_limit_ < 0) {
+int bytes_to_read = record_bytes_;
+if (0 < hop_bytes_ && hop_bytes_ < record_bytes_) {
+bytes_to_read = hop_bytes_;
+}
```
yongtang: Thanks @saxenasaurabh. The indent has been fixed.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Thanks @saxenasaurabh. The PR has been updated with footer_cache and hop_cache combined. Also, now the `file_pos_limit_` seems to be not needed so I just removed the `file_pos_limit_` completely.

tensorflow/core/ops/io_ops.cc
yongtang: @asimshankar Thanks. The PR has been updated with valid values (`GZIP` and `ZLIB`) covered.

yongtang:
Thanks all for the review and help and sorry for the delay. The PR has been updated. Please take a look and let me know if there are any issues.

rmlarsen:

rmlarsen:
@saxenasaurabh can you take another look and approve?

martinwicke:
@saxenasaurabh Can you check this please? In particular make sure that all ops are backwards compatible.

I'm assuming this needs to be rebased.

martinwicke:
Jenkins, test this please.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -29,73 +32,137 @@ class FixedLengthRecordReader : public ReaderBase {
...
+
+    int bytes_to_read = footer_bytes_;
+
+    // In case hop_bytes_ is in between 0 and record_bytes_,
+    // we will need to hold a cache so that later on we could prefix the cache
+    // with the remaining data to read:
+    // For example, assume record_bytes is 2 and hop_bytes is 1,
+    // For file "H123",
+    // We will process the data in the following way:
+    // 1. Read hope_bytes of 1 ("1") and save it in hop cache.
```
saxenasaurabh: nit typo: hope_bytes -> hop_bytes

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -29,73 +32,137 @@ class FixedLengthRecordReader : public ReaderBase {
...
+          new io::BufferedInputStream(file_.get(), kBufferSize));
+    }
+    TF_RETURN_IF_ERROR(buffered_input_stream_->SkipNBytes(header_bytes_));
+
+    int bytes_to_read = footer_bytes_;
+
+    // In case hop_bytes_ is in between 0 and record_bytes_,
+    // we will need to hold a cache so that later on we could prefix the cache
+    // with the remaining data to read:
+    // For example, assume record_bytes is 2 and hop_bytes is 1,
```
saxenasaurabh: and header_bytes is 1?

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -107,11 +174,15 @@ class FixedLengthRecordReader : public ReaderBase {
...
-  int64 file_pos_limit_;
   int64 record_number_;
-  std::unique_ptr<RandomAccessFile> file_;  // must outlive input_buffer_
-  std::unique_ptr<io::InputBuffer> input_buffer_;
+  string encoding_;
+  // must outlive buffered_input_stream_
+  std::unique_ptr<RandomAccessFile> file_;
+  // must outlive buffered_input_stream_
+  std::unique_ptr<io::RandomAccessInputStream> file_stream_;
+  std::unique_ptr<io::InputStreamInterface> buffered_input_stream_;
```
saxenasaurabh: May be call this `buffered_inputstream_` to be consistent with the file naming convention(buffered_inputstream.h etc).

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -29,73 +32,137 @@ class FixedLengthRecordReader : public ReaderBase {
...
+    // 3. Prefix data in 2. with hop cache in one, we have "12",
+    //    this is used for the record.
+    // 4. Shift hop_bytes of 1 from the record, and we put "2" to hop cache.
+    // 5. Continue step 2.
+    // In order to acheive the above, we will peek in "record_bytes_ - hop_bytes_"
+    // before we read the first record. 
+    if (0 < hop_bytes_ && hop_bytes_ < record_bytes_) {
+      bytes_to_read += record_bytes_ - hop_bytes_;
+    }
+    TF_RETURN_IF_ERROR(buffered_input_stream_->ReadNBytes(bytes_to_read, &lookahead_cache_));
```
saxenasaurabh: What happens if the file is empty, i.e., has no records? ZlibInputStream would throw a OutOfRangeError here. May be defer filling cache till first call to ReadLocked?
It would be super awesome if you could add a test for this!

drpngx:
It looks like the test fails because of an API change. @martinwicke could you confirm that it's OK?

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -29,73 +32,137 @@ class FixedLengthRecordReader : public ReaderBase {
...
+    lookahead_cache_ = lookahead_cache_.substr(bytes_to_read);
+
     *key = strings::StrCat(current_work(), ":", record_number_);
     *produced = true;
     ++record_number_;
 
-    if (hop_bytes_ > 0) {
-      input_buffer_->Seek(pos_before_read + hop_bytes_).IgnoreError();
+    if (hop_bytes_ > record_bytes_) {
+      buffered_input_stream_->SkipNBytes(hop_bytes_ - record_bytes_);
```
saxenasaurabh: Here you would need to skip past `hop_bytes_ - record_bytes_` in the `lookahead_cache_` and not the `buffered_input_stream_` right? Then you would probably need to try to fill up the `lookahead_cache_` to `footer_bytes_`. If that is not possible you can set `*at_end = true`.
Could you add a test for this case as well please?

saxenasaurabh (changes requested):
Really sorry for the delay in review.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Thanks @saxenasaurabh. Done.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Done.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Done.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Thanks @saxenasaurabh. Yes the skip should happen at the beginning of the `lookahead_cache_` and `buffered_input_stream_` should read and append buffer to `lookahead_cache_`. I have updated the implementation and added the test cases for it.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Thanks @saxenasaurabh. This part has also been addressed and additional test cases have been added.

yongtang:
Thanks @saxenasaurabh for the review. The PR has been updated. Please take a look.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -29,73 +32,155 @@ class FixedLengthRecordReader : public ReaderBase {
...
+          return Status::OK();
+        }
+        // Now
+        lookahead_cache_.append(*value, 0, bytes_to_read);
+        value->clear();
+      }
+    }
+
+    int bytes_to_read = record_bytes_;
+    if (0 < hop_bytes_) {
```
saxenasaurabh: I know this has a lot of corner cases. Here's one suggestion for a simpler mental model:

In the constructor, set `hop_bytes_ = record_bytes_` if `hop_bytes_ == 0`. So now `hop_bytes_` is always > 0.

Remove the `buffered_inputstream_->ReadNBytes(footer_bytes_, &lookahead_cache_)` from `OnWorkStarted`.

In `ReadLocked`

1. Skip past `hop_bytes_` if not reading first record.
```cpp
if (record_number_ != 0) {
  if (hop_bytes_ <= lookahead_cache_.size()) {
    lookahead_cache_ = lookahead_cache_.substr(hop_bytes_);
  } else {
    int64 cache_size = lookahead_cache_.size();
    lookahead_cache_.clear();
    Status s = buffered_inputstream_->SkipNBytes(hop_bytes_-cache_size);
    if (!s.ok()) {
      if (!errors::IsOutOfRange(s)) {
        return s;
      }
      *at_end = true;
      return Status::OK();
    }
  }
}
```

2. Fill up `lookahead_cache_` to `record_bytes_ + footer_bytes_`.
```cpp
bytes_to_read = record_bytes_ + footer_bytes_ - lookahead_cache_.size();
Status s = buffered_inputstream_->ReadNBytes(bytes_to_read, value);
if (!s.ok()) {
  value.clear();
  if (!errors::IsOutOfRange(s)) {
    return s;
  }
  *at_end = true;
  return Status::OK();
}
lookahead_cache_.append(*value, 0, bytes_to_read);
value->clear();
```

3. Copy first `record_bytes_` from cache to `value`
```cpp
*value = lookahead_cache_.substr(0, record_bytes_);
```

saxenasaurabh (changes requested):

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Thanks @saxenasaurabh for the review. Yes that would be better to understand. The PR has been updated.

yongtang:
Thanks @saxenasaurabh for the review. The PR has been updated. Please take a look.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -107,11 +142,15 @@ class FixedLengthRecordReader : public ReaderBase {
   const int64 record_bytes_;
   const int64 footer_bytes_;
   const int64 hop_bytes_;
+  string lookahead_cache_;
```
saxenasaurabh: May be add a comment explaining
1. Why this is needed.
2. What the maximum size of this buffer will be.

saxenasaurabh (approved):

saxenasaurabh:
Jenkins, test this please.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
yongtang: Thanks @saxenasaurabh. The PR has been updated with comments added.

yongtang:
Thanks @saxenasaurabh for the review. The PR has been updated. I also updated the api golden so that api compatibility could pass. Please take a look.

drpngx:
Jenkins, test this please.

tensorflow/core/ops/io_ops.cc
```diff
@@ -448,6 +448,7 @@ REGISTER_OP("FixedLengthRecordReader")
     .Attr("record_bytes: int")
     .Attr("footer_bytes: int = 0")
     .Attr("hop_bytes: int = 0")
+    .Attr("encoding: string = ''")
```
josh11b: There should be no new uses of the V1 op, so there should be no reason to change it.

tensorflow/core/ops/io_ops.cc
```diff
@@ -473,6 +476,7 @@ REGISTER_OP("FixedLengthRecordReaderV2")
...
```
josh11b: Please move this attr to the end, to match the order in the Python API.

tensorflow/python/ops/io_ops.py
```diff
@@ -396,6 +396,7 @@ def __init__(self,
                header_bytes=None,
                footer_bytes=None,
                hop_bytes=None,
+               encoding=None,
```
josh11b: For backwards-compatibility reasons, please put this new parameter last.

josh11b (commented):
(Comments from API review)

drpngx (changes requested):
Marking changes request to avoid accidentally merging.

tensorflow/core/ops/io_ops.cc
yongtang: Thanks @josh11b, done.

tensorflow/core/ops/io_ops.cc
yongtang: Thanks @josh11b, done.

tensorflow/python/ops/io_ops.py
yongtang: Done.

yongtang:
Thanks @josh11b @drpngx for the review. The PR has been updated. Please take a look.

drpngx:
Jenkins, test this please.

drpngx (approved):
Verified that the latest version addressed @josh11b comments.

Issue #8852, opened 2017-03-30
Allow SavedModelBuilder to overwrite existing directory


rajhans:
Currently, SavedModelBuilder throws an exception when called with a directory that already exists. It will be helpful to add a flag `overwrite` when initializing SavedModelBuilder, which allows it to overwrite the contents of the directory when .save() is called. 

When training a model every couple of hours on fresh data, it will be easier to overwrite existing models than having to implement housekeeping code around cleaning old models or writing code to move around new model after training.

terryKing1992:
much more agree with you

aselle:
We currently don't have anybody working on this. It would be great if you could help us by working on this and submitting a PR. Let us know if you need further clarification. Thanks!

Mistobaan:
I was looking into this, and I think that is cleaner to have one single export dir per trained graph. You can add a time signature to the export dir '<model-id>-<YYYYHHMM>'  and link it to a  "last model" directory. Changing the behaviour of the SaveModelBuilder to account for overwrite cases is not worth it.

martinwicke:
We had some internal discussions about this and we vehemently agree with @Mistobaan. We would very much avoid giving users a very sharp stick with which to poke themselves in the eye. We have lost lots of work to setups like this, let's not make this easier.

If you must, you can remove the directory before making the SavedModelBuilder.

rajhans:
Yeah, already following single-export-dir per trained graph. Thanks for the discussion.

tbrodbeck:
Would still be nice to have an optional overwrite flag though. (Sorry for warming up this old pizza with my 5 cents, btw)

PR #8902, opened 2017-04-02
Allow SavedModelBuilder to overwrite existing folders

just-a-jazz:
This adds the feature #8852 requested

### DIFF
--- a/tensorflow/python/saved_model/builder_impl.py
+++ b/tensorflow/python/saved_model/builder_impl.py
@@ -81,7 +81,7 @@ class SavedModelBuilder(object):
 
     Args:
       export_dir: The directory to export to
-      overwrite: If `True` overwrites if a directory already exists 
+      overwrite: If `True` overwrites if a directory already exists
     """
     self._saved_model = saved_model_pb2.SavedModel()
     self._saved_model.saved_model_schema_version = (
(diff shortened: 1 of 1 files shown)


rohan100jain:
Thanks for doing this!

rajhans:
thanks for adding this!

alisidd:
No problem! :)

rohan100jain (changes requested):

tensorflow/python/saved_model/builder_impl.py
```diff
@@ -76,16 +76,25 @@ class SavedModelBuilder(object):
   ```
   """
 
-  def __init__(self, export_dir):
+  def __init__(self, export_dir, overwrite=False):
+    """Construct a SavedModelBuilder.
+
+    Args:
+      export_dir: The directory to export to
+      overwrite: If `True` overwrites if a directory already exists
```
alisidd: Done 👍

rohan100jain (approved):

rohan100jain:

martinwicke:
Sorry, we have to revert this. It is breaking some tests we require (which were sadly not run yet).

alisidd:
@martinwicke Could you let me know what the problem was, so I could fix it?

martinwicke:
I'm sorry the initial advice was the other way around, but we had some more discussions about this and we're very certain that we do not want to make this easy. It's a bad idea (tm), take it from people who have lost lots of work to easy overwriting. 

You can always emulate the behavior by doing the delete_recursively manually, but that I would advise against. Ideally, timestamp your directories and symlink the latest.

PR #8907, opened 2017-04-02
corrected which graph_def to validate

Androbin:
After each transformation comes a check for validity.
merged_graph_def was already tested a few lines above.
output_graph_def should be tested directly afterwards.

### DIFF
--- a/tensorflow/tools/graph_transforms/quantize_nodes.cc
+++ b/tensorflow/tools/graph_transforms/quantize_nodes.cc
@@ -941,7 +941,7 @@ Status QuantizeNodes(const GraphDef& input_graph_def,
   // keep interoperability with float ops.
   TF_RETURN_IF_ERROR(RemoveRedundantQuantizations(deduped_graph_def, context,
                                                   output_graph_def));
-  TF_RETURN_IF_ERROR(IsGraphValid(merged_graph_def));
+  TF_RETURN_IF_ERROR(IsGraphValid(output_graph_def));
 
   return Status::OK();
 }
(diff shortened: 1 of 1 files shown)


rohan100jain:
Pete... this does look okay to me, but just wanted to double check.

petewarden (approved):
Thanks, good catch!

rohan100jain:

tensorflow/tools/graph_transforms/quantize_nodes.cc
```diff
@@ -941,7 +941,7 @@ Status QuantizeNodes(const GraphDef& input_graph_def,
   // keep interoperability with float ops.
   TF_RETURN_IF_ERROR(RemoveRedundantQuantizations(deduped_graph_def, context,
                                                   output_graph_def));
-  TF_RETURN_IF_ERROR(IsGraphValid(merged_graph_def));
+  TF_RETURN_IF_ERROR(IsGraphValid(output_graph_def));
```
rohan100jain: I think this needs to be TF_RETURN_IF_ERROR(IsGraphValid(*output_graph_def));

Androbin: Would you like to make the edit?

rohan100jain: I could, but I would prefer if you could push a commit with this minor change.

rohan100jain (approved):

rohan100jain:

PR #8931, opened 2017-04-03
Remove --no-deps flag from pip install

caisq:
of tensorflow whl in pip.sh.
Fixes ongoing breakages related to the new dependencies of the
tensorflow wheel (e.g., bleach and markdown).


yifeif (commented):
Do we still want --force-reinstall?

caisq:
@yifeif I believe the "--force-reinstall" is a vestige from the very old days when we were not doing the install and test in virtualenv. I don't think it is necessary anymore.

caisq:
Also tested: GPU Python3.4 and 3.5 pip builds:
http://ci.tensorflow.org/view/Experimental/job/experimental-cais-matrix-linux-gpu/2/

gunan (approved):

yifeif (approved):

PR #8935, opened 2017-04-03
Fix broken ./configure on mac because of non-portable sed syntax

### LINKED ISSUE #9553: configure does not work on macOS if sed is GNU sed
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 10.12.4
- **TensorFlow version (use command below)**: Current head: 3ce228e

### Describe the problem
If sed is GNU sed on macOS, configure fails with `sed: can't read : No such file or directory`.

JVillella:
Running `./configure` on mac complains with the following error:

```shell
sed: can't read /tf_configure/d: No such file or directory
```

According to [this SO post](http://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux) the syntax to avoid creating backup files with sed on 10.9+ is without a space between `-i ''`.

Fixes #9553.

### DIFF
--- a/configure
+++ b/configure
@@ -35,12 +35,9 @@ function is_windows() {
   fi
 }
 
-function sed_hyphen_i() {
-  if is_macos; then
-    sed -i '' "$@"
-  else
-    sed -i "$@"
-  fi
+function sed_in_place() {
+  sed -e $1 $2 > "$2.bak"
+  mv "$2.bak" $2
 }
 
 function write_to_bazelrc() {
@@ -170,7 +167,7 @@ function setup_python {
 rm -f .tf_configure.bazelrc
 touch .tf_configure.bazelrc
 touch .bazelrc
-sed_hyphen_i "/tf_configure/d" .bazelrc
+sed_in_place "/tf_configure/d" .bazelrc
 echo "import %workspace%/.tf_configure.bazelrc" >> .bazelrc
 
 # Delete any leftover BUILD files from the Makefile build, which would interfere


JVillella:
I signed it!

gunan:
Jenkins, test this please.

gunan (approved):

JVillella:
@gunan what version of OSX are the mac slaves running? If it's pre-10.9 that would explain why it's failing (and worked before). I can try to make the code more portable for this.

gunan:
@JVillella our executors should be el capitan and sierra only, (all 10.11+).

drpngx:
Error seems invalid.

Jenkins, test this please.

gunan:
The failure in mac cpu tests is caused by the changes in this PR.

@JVillella do you have any insight about the failure, or how to fix it?
If not, I recommend we close this PR

JVillella:
@gunan looking again now.

JVillella:
So I seem to be running `gnu-sed 4.2.2` which was likely installed from `coreutils` at some point in time. It has a different syntax. MacOS's built-in `sed` wants the empty string, `gsed` doesn't which is why there is a check (though for platform, assuming if you're on mac you're not using gsed):

```shell
function sed_hyphen_i() {
  if is_macos; then
    sed -i '' "$@"
  else
    sed -i "$@"
  fi
}
```

Using backups seems to be more portable (e.x. `sed -i '.bak' ...`). No platform check is required either. We could do this and just remove the backup. If you guys are ok with this solution I can make the change. `awk` (not `gawk`) doesn't let us do it in place either.

This issue only affects someone running `gsed` in place of `sed` on mac. If this makes it not-a-concern we can close the ticket.

martinwicke:
Sorry for leaving this hanging. I think using .bak is a fine solution, especially if you delete them afterwards. I'd like to avoid the if, which seems brittle.

vrv:
@JVillella any updates?

JVillella:
Hey @martinwicke, @vrv. I've updated the PR with a more portable solution. The build is passing now as well. Let me know if this is okay with you guys.

martinwicke:
Looks good to me. 

Jenkins, test this please.

PR #8940, opened 2017-04-04
Updating update version script

av8ramit:
Updating the install source files to be rc1 compliant. Updating the update_version script as well to handle these changes.

### DIFF
--- a/tensorflow/tools/ci_build/update_version.sh
+++ b/tensorflow/tools/ci_build/update_version.sh
@@ -73,6 +73,7 @@ OLD_MINOR=$(cat ${VERSION_H} | grep -E "^#define TF_MINOR_VERSION [0-9]+" | \
 cut -d ' ' -f 3)
 OLD_PATCH=$(cat ${VERSION_H} | grep -E "^#define TF_PATCH_VERSION [[:alnum:]-]+" | \
 cut -d ' ' -f 3)
+OLD_PIP_PATCH="${OLD_PATCH//-}"
 
 sed -i -e "s/^#define TF_MAJOR_VERSION ${OLD_MAJOR}/#define TF_MAJOR_VERSION ${MAJOR}/g" ${VERSION_H}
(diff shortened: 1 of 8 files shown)


tensorflow/tools/ci_build/update_version.sh
```diff
@@ -92,6 +93,16 @@ check_existence file "${README_MD}"
 
 sed -i -r -e "s/${OLD_MAJOR}\.${OLD_MINOR}\.([[:alnum:]]+)-/${MAJOR}.${MINOR}.${PIP_PATCH}-/g" "${README_MD}"
 
+# Update the install md files
+NEW_PIP_TAG=$MAJOR.$MINOR.$PIP_PATCH
+OLD_PIP_TAG=$OLD_MAJOR.$OLD_MINOR.$OLD_PIP_PATCH
+
+for file in ${TF_SRC_DIR}/docs_src/install/install_{linux,mac,windows,sources}.md
```
gunan: How about install_java, install_c and install_go files?

av8ramit: I believe those are maintained manually. I can add them, but I was just worried about unintended conversions.

gunan: For this release, they may be maintained manually.
But eventually they will be a part of the release
@asimshankar for verification.

As they already have 1.1.0 in their versions, until our "old" version becomes 1.1.0, they wont be updated by the script's sed commands. So I think it is safe to incorporate them here now.

asimshankar: Yup, since this was the first release for some of those docs (and the second release for the Java and C APIs in general), they were manual. But going forward, yes, they should be updated concurrently with everything else (the release artifacts are built together).

Thanks for checking!

av8ramit: Done. All install files are now updated with update_version.

gunan (approved):

PR #8955, opened 2017-04-04
Fix gradient implementation documentation

neilmb:
The docs for implementing gradients for a custom op are inconsistent. It seems that a gradient op for an op that takes multiple *inputs* should return a list of tensors with the gradients with respect to the individual inputs. The next paragraph matches this interpretation.

### DIFF
--- a/tensorflow/docs_src/extend/adding_an_op.md
+++ b/tensorflow/docs_src/extend/adding_an_op.md
@@ -1118,8 +1118,8 @@ Details about registering gradient functions with
   about any attrs can be found via
   @{tf.Operation.get_attr}.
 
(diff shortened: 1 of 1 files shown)


neilmb:
I signed it!

vrv:
I think the documentation is correct:

- An op that produces multiple outputs may receive gradients from each of them.  Hence, in the gradient function for an op, 'grads' will be the gradients it receives for each output, and so 'grads' will contain as many Tensors as there are outputs. 

- The gradient function must *produce* / return Tensors (or None) for each of its inputs, since the ops that produce the input would also expect to receive a Tensor as their gradients. 

Is that clear?  Feel free to comment if you think further clarification is needed..

PR #8970, opened 2017-04-05
Added "hop_bytes" option to FixedLengthRecordReader

t13m:
A "hop_bytes" option added to FixedLengthRecordReader.

With this option, one can easily read fixed-length record from file with a sliding window. This is useful for reading audio waveform or features with contexts.

Example:

file:
`"1234567"`

reader:
`record_bytes=3, hop_bytes=2`

result:
`"123", "345", "567"`


t13m:
I signed it!

concretevitamin:
@t13m please ping this PR when you've got a chance to add some unit tests.  Thanks!

t13m:
@concretevitamin @rohan100jain Sure I'm working on unit tests. Should the unit test be added to `tensorflow/python/test_kernels/reader_ops_test.py`? Should I modify the origin `FixedLengthRecordReaderTest` or write a new test just for this new option?

Thank you very much.

t13m:
@concretevitamin Hi, I modified the original unittest of `FixedLengthRecordReader` to make it test the new argument. The unittest is passed. I'm not sure if it is enough, if not, please let me know. Thank you for reviewing.

tensorflow/python/kernel_tests/reader_ops_test.py
```diff
@@ -374,23 +384,23 @@ def testOneEpoch(self):
           header_bytes=self._header_bytes,
           record_bytes=self._record_bytes,
           footer_bytes=self._footer_bytes,
+          hop_bytes=self._hop_bytes,
```
concretevitamin: Please add another test method, e.g. `testOneEpochWithHopBytes()` to test the new behavior.  We need to restore this unit test to its original coverage, namely `hop_bytes == 0` by default and use `_Record()`, `_num_records`, etc.

Furthermore, I'd highly suggest to simplify the record setup by explicitly spelling out the whole content for a simple case.  Your example in the pull request description is much clearer.

tensorflow/python/kernel_tests/reader_ops_test.py
```diff
@@ -23,6 +23,7 @@
 import os
 import threading
 import zlib
+import math
```
concretevitamin: nit: please sort by alphabetical order

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -62,14 +64,20 @@ class FixedLengthRecordReader : public ReaderBase {
 
   Status ReadLocked(string* key, string* value, bool* produced,
                     bool* at_end) override {
-    if (input_buffer_->Tell() >= file_pos_limit_) {
+    if (input_buffer_->Tell() >= file_pos_limit_ ||
+        input_buffer_->Tell() + record_bytes_ > file_pos_limit_) {
```
concretevitamin: IIUC, you probably don't need this new condition?  This error case will be signaled by Line 73's `ReadNBytes()` call.  If this is right, for clarity let's remove it here.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -62,14 +64,20 @@ class FixedLengthRecordReader : public ReaderBase {
...
       *at_end = true;
       return Status::OK();
     }
+    auto current_position = input_buffer_->Tell();
```
concretevitamin: 1. Spell out `const int64`.  It's our style guide to not excessively `auto` (for some definition of excessiveness :)).

2. It's clearer to call it something like `pos_before_read`.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
```diff
@@ -62,14 +64,20 @@ class FixedLengthRecordReader : public ReaderBase {
...
     TF_RETURN_IF_ERROR(input_buffer_->ReadNBytes(record_bytes_, value));
     *key = strings::StrCat(current_work(), ":", record_number_);
     *produced = true;
     ++record_number_;
+
+    if (hop_bytes_ > 0)
```
concretevitamin: Add a { } block.

concretevitamin (changes requested):

tensorflow/core/kernels/fixed_length_record_reader_op.cc
t13m: The error case will be signaled by `ReadNBytes()`, when the reading position is out of the range of file. With the original condition, it is possible to read a record containing some bytes of footer. Is this the expected behavior?

t13m: When `hop_bytes_ < record_bytes_`, which is true in most cases, the last record will always read some bytes out of `file_pos_limit_` if we only check `input_buffer_->Tell() >= file_pos_limit_`.

For example, assume `hop_bytes_ = 2`, when reading a file with content "0123456FF" ("FF" is "footer"), the last record will be "6FF". If the file is "0123456", an error will be signaled when reading the last record. I thought behaviors like this is not what is expected for a reader with `hop_bytes` argument. It maybe better to have the all bytes of the last record in range of `file_pos_limit_`. What do you think? Thank you for reviewing.

drpngx:
@t13m feel free update the code and push, or ask if you have any questiosn.

tensorflow/core/kernels/fixed_length_record_reader_op.cc
concretevitamin: It seems to me the 2 cases you described are expected behaviors.  I believe the caller of this reader class is responsible for setting up the reader configuration so that footer bytes are disallowed if desired (and vice versa).  Do you agree?

tensorflow/python/kernel_tests/reader_ops_test.py
```diff
@@ -21,6 +21,7 @@
 import collections
 import gzip
 import os
+import math
```
t13m: Oh, yes!

tensorflow/core/kernels/fixed_length_record_reader_op.cc
t13m: For the original `FixedLengthRecordReader`, I agree with you that the caller is possible and is responsible for setting up everything for the reader. 
But I don't think it's true for the condition `hop_bytes>0`. Under this situation, it maybe not possible for caller to setting up everything. Consider following condition.

File:
`0123456`

Reader setting:
`record_bytes=3`, `hop_bytes=2`, `footer_bytes=0`, `header_bytes=0`

What we expected is:
`012`, `234`, `456`

If I get it right, with the original checking condition(`input_buffer_->Tell() >= file_pos_limit_`), what we actually get will be:
`012`, `234`, `456`, `Error`.
There seems not possible to avoid the last error, because for the last record, the remaining bytes will always be less than `record_bytes`: `01234|56`("|" for the start position of the last record)

Or should errors like this be catched by the caller?

concretevitamin: Ah, thanks for the explanation.  I think you meant that after the first three successful reads, the `Tell()` cursor will be at `|` in `012345|6`.  Please add this explanation for `hop_bytes > 0` and this example in a comment.

concretevitamin (commented):
Looks good modulo adding a comment.  Thanks for your explanation!

t13m:
Hi, sorry for reply so late. I added the comment requested, thanks for reviewing!

drpngx:
Jenkins, test this please.

concretevitamin (approved):
Thanks for your patience!

caisq:
cc @martinwicke @josh11b 
This PR involves an API change (in io_ops.py / FixedLengthRecordReader). I can fix the golden during the upcoming pull. Let me know if you think we should proceed in other ways.

t13m:
@caisq Hi, sorry but I didn't get it. Could you please explain what does "fix the golden" and "upcoming pull" mean?

caisq:
@t13m These changes are being pulled to Google's internal SCM as per our routine sync, where it is subject to compatibility checks not present on GitHub.

PR #8981, opened 2017-04-05
fix summary writer thread leak

suiyuan2009:
#4820 
I assume there is only one daemon thread consumes event_queue.

### DIFF
--- a/tensorflow/python/summary/writer/event_file_writer.py
+++ b/tensorflow/python/summary/writer/event_file_writer.py
@@ -64,9 +64,10 @@ class EventFileWriter(object):
     self._event_queue = six.moves.queue.Queue(max_queue)
     self._ev_writer = pywrap_tensorflow.EventsWriter(
         compat.as_bytes(os.path.join(self._logdir, "events")))
+    self._flush_secs = flush_secs
     self._closed = False
     self._worker = _EventLoggerThread(self._event_queue, self._ev_writer,
-                                      flush_secs)
+                                      self._flush_secs)
 
     self._worker.start()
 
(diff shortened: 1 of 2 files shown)


tensorflow/python/summary/writer/event_file_writer.py
```diff
@@ -108,7 +112,9 @@ def close(self):
 
     Call this method when you do not need the summary writer anymore.
     """
+    self._worker.stop()
     self.flush()
+    self._worker.join()
```
mrry: Won't this block forever if `self._worker` is blocked in `self._queue.get()` on L151?

suiyuan2009: I'll check queue's size first, I assume there is only one daemon thread.

suiyuan2009: @mrry I should not use while True: continue because this will consume much cpu resource.

mrry: I think the following test will almost always block forever:

```python
e = EventFileWriter(...)
time.sleep(1.0)
e.close()
```

The standard approach here would be to send a sentinel element.

(I think I made a similar comment on #4860, which unfortunately was dropped. It would be good to fix this problem, so I'd encourage you to revise this and add some tests that would catch this misbehavior.)

suiyuan2009: @mrry , I've tested, before queue get, I'll check queue size, so in my approach, the daemon thread will never be blocked by queue get, but in a busy waiting state(execute while True and continue again and again).

sentinel element is a better approach.

mrry (approved):
Looks good to me. Thanks for the fix!

mrry:

PR #8987, opened 2017-04-05
Adding a hint to the XLA op metadata, set when the op is updating a variable

DavidNorman:
The hint can be used for ensuring that the buffer allocated to a variable remains at the same place.  For a static graph compiler, this prevents an extra copy at the end of the execution.


DavidNorman:
ok - i've run all the XLA unit tests and there is something wrong - will look into that immediately

PR #8990, opened 2017-04-05
wipe out null string

zjj2wry:
wipe out null string. markdown will not influence


rohan100jain (approved):

rohan100jain:

PR #8996, opened 2017-04-05
clean separation

zjj2wry:
better differentiate license and annotation.


rohan100jain:
Thanks for the change but in general, we don't have any separation between code and copyright. Closing PR.

PR #8999, opened 2017-04-05
Fix KeyError in quantize_graph round, quantize opt

samjabrahams:
The round/quantize paths check for a boolean value in a dictionary before it is set, leading to a KeyError. This change allows the input graph to be quantized using the 'round' and 'quantize' options. The quantized graph can't actually be used (for the 'round' option, at least) due to a lack of a `RoundToSteps` Op, but I figured I'd submit this anyway.

Pinging @petewarden for thoughts on whether this is worth incorporating now.

### DIFF
--- a/tensorflow/tools/quantization/quantize_graph.py
+++ b/tensorflow/tools/quantization/quantize_graph.py
@@ -485,7 +485,7 @@ class GraphRewriter(object):
 
   def quantize_nodes_recursively(self, current_node):
     """The entry point for quantizing nodes to eight bit and back."""
-    if current_node.name in self.already_visited:
+    if (current_node.name in self.already_visited) and self.already_visited[current_node.name]:
       return
     self.already_visited[current_node.name] = True
     for input_node_name in current_node.input:
@@ -507,7 +507,7 @@ class GraphRewriter(object):
   def quantize_node(self, input_node):
     """Handles quantizing a single node."""
     input_name = input_node.name
-    if input_name in self.already_quantized:
+    if (current_node.name in self.already_visited) and self.already_visited[current_node.name]:
       return
     self.already_quantized[input_name] = True
     original_input_name = input_name + "_original"
--- a/tensorflow/tools/quantization/quantize_graph.py
+++ b/tensorflow/tools/quantization/quantize_graph.py
@@ -485,7 +485,7 @@ class GraphRewriter(object):
 
   def quantize_nodes_recursively(self, current_node):
     """The entry point for quantizing nodes to eight bit and back."""
-    if self.already_visited[current_node.name]:
+    if self.already_visited.get(current_node.name):
       return
     self.already_visited[current_node.name] = True
     for input_node_name in current_node.input:
--- a/tensorflow/tools/quantization/quantize_graph.py
+++ b/tensorflow/tools/quantization/quantize_graph.py
@@ -453,7 +453,7 @@ class GraphRewriter(object):
 
   def round_nodes_recursively(self, current_node):
     """The entry point for simple rounding quantization."""
-    if (current_node.name in self.already_visited) and self.already_visited[current_node.name]:
+    if self.already_visited.get(current_node.name, False):
       return
     self.already_visited[current_node.name] = True
     for input_node_name in current_node.input:
@@ -485,7 +485,7 @@ class GraphRewriter(object):
 
   def quantize_nodes_recursively(self, current_node):
     """The entry point for quantizing nodes to eight bit and back."""
-    if (current_node.name in self.already_visited) and self.already_visited[current_node.name]:
+    if self.already_visited.get(current_node.name, False):
       return
     self.already_visited[current_node.name] = True
     for input_node_name in current_node.input:
@@ -494,10 +494,6 @@ class GraphRewriter(object):
       self.quantize_nodes_recursively(input_node)
     nodes_to_quantize = ["Conv2D", "BiasAdd", "MatMul"]
     if any(current_node.op in s for s in nodes_to_quantize):
-      for input_name in current_node.input:
-        input_name = node_name_from_input(input_name)
-        input_node = self.nodes_map[input_name]
-        self.quantize_node(input_node)
       self.quantize_node(current_node)
     else:
       new_node = node_def_pb2.NodeDef()
@@ -507,7 +503,7 @@ class GraphRewriter(object):
   def quantize_node(self, input_node):
     """Handles quantizing a single node."""
     input_name = input_node.name
-    if (current_node.name in self.already_visited) and self.already_visited[current_node.name]:
+    if self.already_quantized.get(input_name, False):
       return
     self.already_quantized[input_name] = True
     original_input_name = input_name + "_original"
--- a/tensorflow/tools/quantization/quantize_graph.py
+++ b/tensorflow/tools/quantization/quantize_graph.py
@@ -453,7 +453,7 @@ class GraphRewriter(object):
 
   def round_nodes_recursively(self, current_node):
     """The entry point for simple rounding quantization."""
-    if self.already_visited.get(current_node.name, False):
+    if current_node.name in self.already_visited:
       return
     self.already_visited[current_node.name] = True
     for input_node_name in current_node.input:
@@ -485,7 +485,7 @@ class GraphRewriter(object):
 
   def quantize_nodes_recursively(self, current_node):
     """The entry point for quantizing nodes to eight bit and back."""
-    if self.already_visited.get(current_node.name, False):
+    if current_node.name in self.already_visited:
       return
     self.already_visited[current_node.name] = True
     for input_node_name in current_node.input:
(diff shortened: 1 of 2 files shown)


drpngx:
Can we have a test for this perhaps?

samjabrahams:
@drpngx - agreed on figuring out a way to properly test these. Originally this was just to fix some regressions, but it's better to follow through and try to get the entire tool up-to-spec. I made a few tweaks to the existing `quantize_graph_test.py` file to help get that ball rolling. Notes on the current changes:

* Changed `self.already_visited.get(...)` to `if current_node.name in self.already_visited` to [keep existing pattern consistent](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/quantization/quantize_graph.py#L510-L511)
* As [@petewarden notes](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/quantization/quantize_graph_test.py#L165-L173)- we can't actually use the `'rounded'` mode right now as there is no `RoundToSteps` Op. We can still run the `GraphRewriter` to check that the recursion no longer fails.
* I created a test for the `'quantize'` mode (and set it to use the `QuantizeV2` Op), but I'm getting some failures for several of the tests. I've attached the failure log below.

[Test failure log](https://github.com/tensorflow/tensorflow/files/914756/test.txt)

I'll try to spend more time on this over the next week or two.

(Additionally, I had a rebase hiccup, which caused the CLA issue; should be fixed now).

drpngx:
@samjabrahams that sounds like a good approach to me. Looking forward to updates!

samjabrahams:
Hi @vrv- I'm going to spend some time on this today- hopefully have the PR updated relatively shortly!

samjabrahams:
Making a bit of progress figuring out what's going wrong (at least for this leg of it). It looks like somewhere between `quantize_graph_test.py` and `quantize_graph.py`, a `GraphDef` containing two nodes with the same name is getting passed along to `import_graph_def()`. This causes an error when `import_graph_def()` attempts to find the inputs for each operation, as the inputs being assigned to the "Dequantize" op are those from another operation (in this case, the other operation doesn't have any inputs).

[Relevant section of `import_graph_def()`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/importer.py#L305-L313)

I'll keep poking to see where the hiccup is. In the meantime, would it be a good idea to add a check in `import_graph_def()` to make sure there aren't duplicate operation names in the `GraphDef`?

samjabrahams:
I figured out that some nodes were getting added twice: [once here](https://github.com/tensorflow/tensorflow/blob/d57c31661e8f76d6c3d91f3969d4d4219c9ea97c/tensorflow/tools/quantization/quantize_graph.py#L503-L505), and again [here](https://github.com/tensorflow/tensorflow/blob/d57c31661e8f76d6c3d91f3969d4d4219c9ea97c/tensorflow/tools/quantization/quantize_graph.py#L497-L500). If a node is an input to `Conv2D`, `BiasAdd`, or `MatMul`, both the original and a quantized version of that input is added to the graph, both of which have the same name (leading to the error mentioned above).

For now, I've removed the loop that adds quantized inputs, and the "Dequantize Operation expects three inputs" errors have gone away. However, I'm not sure if this is the correct fix: is the `quantize_graph` tool supposed to quantize the inputs to `Conv2D`, `BiasAdd`, and `MatMul`? If so, we'll need to rework the algorithm to not double-add operations.

Additionally, we're getting some numeric failures from converting to/from quantized land. My guess is that these are due to forcing the quantized range to include the number 0, but I figured I would include this just in case something looks screwy.

```
2.0000 4.0000 6.0000 8.0000 10.0000 12.0000 8.0000 10.0000 12.0000 14.0000 16.0000 18.0000 
3.7647 5.5843 7.3412 9.0980 10.9176 12.6745 9.0980 10.9176 12.6745 14.4314 16.2510 18.0078 
Tensors have 5 different values (41.6666666667%), with mean difference -0.896732529004 and mean absolute difference 0.896732529004

2.0078 4.0157 6.0235 7.9686 9.9765 7.0275 8.9725 10.9804 12.9882 14.9961 
3.7020 5.4588 7.2157 8.9098 10.6667 8.0314 9.7882 11.5451 13.2392 14.9961 
Tensors have 4 different values (40.0%), with mean difference -0.85960791111 and mean absolute difference 0.85960791111

2.0000 4.0000 6.0000 8.0000 10.0000 7.0000 9.0000 11.0000 13.0000 15.0000 
3.7216 5.4549 7.1882 8.9216 10.6549 8.0549 9.7882 11.5216 13.2549 14.9882 
Tensors have 4 different values (40.0%), with mean difference -0.854902219772 and mean absolute difference 0.857255125046

105.0000 150.0000 183.0000 95.0000 235.0000 312.0000 357.0000 178.0000 187.0000 234.0000 261.0000 121.0000 
171.5843 204.4628 229.1216 164.3922 267.1373 323.6471 356.5255 225.0118 232.2039 266.1098 285.6314 182.8863 
Tensors have 11 different values (91.6666666667%), with mean difference -40.8928273519 and mean absolute difference 40.9719085693

1.0000 4.0000 9.0000 16.0000 25.0000 36.0000 49.0000 64.0000 81.0000 100.0000 121.0000 144.0000 
2.2431 5.0471 10.0941 16.8235 25.7961 37.0118 49.9098 64.4902 81.3137 100.3804 121.1294 144.1216 
Tensors have 4 different values (33.3333333333%), with mean difference -0.696733077367 and mean absolute difference 0.696733077367

74.0000 80.0000 86.0000 92.0000 173.0000 188.0000 203.0000 218.0000 
123.1059 127.0588 131.0118 134.9647 188.0471 198.2118 207.8118 217.9765 
Tensors have 7 different values (87.5%), with mean difference -26.7735290527 and mean absolute difference 26.7794113159

348.0000 252.0000 274.0000 175.0000 
348.0353 300.5451 311.4000 261.8745 
Tensors have 3 different values (75.0%), with mean difference -43.2137298584 and mean absolute difference 43.2137298584
```

tensorflow/tools/quantization/quantize_graph.py
```diff
@@ -453,7 +453,7 @@ def rewrite(self, output_node_names):
 
   def round_nodes_recursively(self, current_node):
     """The entry point for simple rounding quantization."""
-    if (current_node.name in self.already_visited) and self.already_visited[current_node.name]:
+    if current_node.name in self.already_visited:
```
samjabrahams: This was actually changed recently in #9393, but this actually brings the column width past 80 characters. This syntax is used in the rest of the file, and since the values in `self.already_visited` are only ever set to `True`, we can stick to this more concise version.

tensorflow/tools/quantization/quantize_graph.py
```diff
@@ -485,7 +485,7 @@ def round_nodes_recursively(self, current_node):
 
   def quantize_nodes_recursively(self, current_node):
     """The entry point for quantizing nodes to eight bit and back."""
-    if self.already_visited[current_node.name]:
+    if current_node.name in self.already_visited:
```
samjabrahams: This is broken logic, fixing using same syntax as above.

vrv:

samjabrahams:
Noting the failures I get on my machine when running [`bazel test tensorflow/tools/quantization:quantize_graph_test`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/quantization/quantize_graph_test.py):

* test_bias_add
* test_bias_add_w_fake_quant_w_min_max_vars
* test_bias_add_w_fallback_min_max_vars
* test_conv
* test_identity
* test_mat_mul_small
* test_odd_padding_problem

These are the ones that complain about numeric inaccuracy. I don't think `tensorflow/tools` is included in the CI tests, so the errors may not pop up here.

vrv:
It looks like it is running in our CI: https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu-python3/4296/consoleFull

I can see the same types of accuracy errors, so good to confirm.  I'll leave it up to @petewarden to figure out what next steps are!

petewarden:
Sorry for commenting so late (the original mention came in while I was on vacation, and I must have missed it when I was catching up). I've been trying to move over to the newer version of the quantize_nodes rewriting rule in the Graph Transform Tool, since this python script became hard to maintain. As a sanity check, are you able to use this alternative approach?

https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms/#eight-bit-calculations

samjabrahams:
@petewarden - apologies, I haven't forgotten about this. Last week was busy- I'll try out the new script and let you know the results this week.

rmlarsen:
@samjabrahams any updates?

rmlarsen:
@petewarden can you answer @samjabrahams's questions regarding future plans? 
@samjabrahams does this mean we can close this PR or do you intend to rework it?

samjabrahams:
@rmlarsen if the tool is going to be deprecated in favor of the GTT, I'm fine closing this out. Otherwise, I'm still not sure whether this fix is the right one ([questions/concerns here](https://github.com/tensorflow/tensorflow/pull/8999#issuecomment-298783900)).

martinwicke:
I believe that is the case. Sorry for the large amounts of semi-deprecated code all around... 
I will close this issue then.

@petewarden will you remove the dead code?

Issue #206, opened 2015-11-13
Generalize slicing and slice assignment ops (including gather and scatter)


girving:
We should make our slicing and assignment ops more general to capture more of the functionality of numpy slicing, and add `__getitem__` sugar for all of it.  Specifically,

1. We should have a 2.5 dimensional set of ops, with dimensions (1) get vs. set, (2) slice type, and for the assignment ops (3) the update op.  Currently we have `slice`, `assign_update`, `assign_add`, `assign_sub`, `gather`, `scatter_update`, `scatter_add`, `scatter_sub`.  We should also have `assign_slice_update`, `assign_slice_add`, `assign_slice_sub`.

2. Both slicing and slice assignment should support strides, with no performance cost if strides aren't used.

3. Ideally, the slice ops should support negative indexing a la Python.  Since the slice parameters are already CPU, this is implementable with near zero cost.   The unfortunate bit is that since we picked the wrong format for specifying ranges (start + length instead of start : end), negative indexing might be awkward.  Thus, it might be best left to a separate bug.

4. Support numpy-like boolean indexing.

5. Generalize `gather` and `scatter_*` to take an array of input index tensors, efficiently broadcast them, and do multidimensional indexing similar to numpy.

6. Make `__getitem__` provide sugar for all of the above.  Ideally we'd have something idiomatically similar at least to `__setitem__`, but this is problematic since the returned assignment op is important to have, `__setitem__` does not return a value, and the nice range sugar is available only inside indexing / assignment calls.

@ebrevdo: I'm assigning this to you for now since you might get to it first, but feel free to grab only the piece of it that you need for now.

girving:
Lasse requests the equivalent of numpy mixed indexing:

    x[:, tensor]

which combines slicing with indexing-by-tensor.

lespeholt:
...where tensor can be either a scalar (which would select all the values in that column) or a vector which can select individual columns in the rows, so:
foo = tf.constant([[1,2,3], [4,5,6]])
foo[:, 1] # [2, 5]
indexes = tf.constant([1, 2])
foo[:, indexes] # [2, 6]

avostryakov:
If I understand correctly the following code is exactly what is needed for cross-entropy loss function
indexes = tf.constant([1, 2])
foo[:, indexes] # [2, 6]

If we will have this kind of indexing we can write:
cost = -tf.reduce_sum(tf.log(network_output[:, targets]))

, where targets is a vector of class's indexes instead of

[...]

PR #9000, opened 2017-04-05
Allow the output of `tf.argmax` as index type

### LINKED ISSUE #8951: Using output of `tf.argmax` as index raises TypeError
### Minimal example
```
a = tf.constant([1, 2, 3], dtype=tf.float32)
b = tf.argmax(a)
tf.Session().run(a[b])
```
This raises a `TypeError`
```
TypeError: Input 'strides' of 'StridedSlice' Op has type int32 that does not match type int64 of argument 'begin'.
```
The solution is to `tf.cast` `b` into a `tf.int32` before using it as an index. The output of `tf.argmax` should be usable as an index immediately. I feel that the current state is not consistent, is there some hidden reason for this?

yongtang:
This fix tries to fix the issue raised in #8951 where the following will raise a `TypeError`:
```
a = tf.constant([1, 2, 3], dtype=tf.float32)
b = tf.argmax(a)
tf.Session().run(a[b])

TypeError: Input 'strides' of 'StridedSlice' Op has type int32 that does not match type int64 of argument 'begin'.
```

The reason for the erorr is that, `strides` is added  as `append(1)` without type while `begin` is appended with type. (See commit diff).

The mismatch of `strides` and `begin` causes the error.

This fix fixes the issue by cast the stride with the same type as `begin` when needed.

This issue was raised in #8951. It was also raised earlier in https://github.com/tensorflow/tensorflow/issues/206#issuecomment-259076860

This fix fixes #8951.

### DIFF
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -470,7 +470,10 @@ def _SliceHelper(tensor, slice_spec, var=None):
     else:
       begin.append(s)
       end.append(s + 1)
-      strides.append(1)
+      if isinstance(s, ops.Tensor):
+        strides.append(constant(1, s.dtype))
+      else:
+        strides.append(np.ones_like(s).dtype.type(1))
       shrink_axis_mask |= (1 << index)
     index += 1
 


tensorflow/python/ops/array_ops.py
```diff
@@ -470,7 +470,7 @@ def _SliceHelper(tensor, slice_spec, var=None):
     else:
       begin.append(s)
       end.append(s + 1)
-      strides.append(1)
+      strides.append(gen_math_ops.cast(1, s.dtype))
```
ebrevdo: please use constant_op.constant(1, s.dtype) or np.array(1, dtype=s.dtype.as_numpy_dtype)  (the latter is probably better!)

yongtang:
Thanks @ebrevdo for the review. The PR has been updated. Please take a look.

tensorflow/python/ops/array_ops.py
```diff
@@ -470,7 +470,7 @@ def _SliceHelper(tensor, slice_spec, var=None):
     else:
       begin.append(s)
       end.append(s + 1)
-      strides.append(1)
+      strides.append(constant_op.constant(1, s.dtype))
```
ebrevdo: hm.  is s assumed to be a numpy element here?  or could it be a regular integer?

should this be `np.ones_like(s)` instead?

yongtang:
Thanks @ebrevdo. The `s` (`slice_spec`) seems to be Tensor. I updated the PR and used `ones_like` (`strides.append(ones_like(s))`). Please take a look.

ebrevdo:
Lgtm if it passes tests.

caisq:

caisq:
There are quite a few test failures that seem to be related to this PR.

drpngx:
@yongtang could you check the tests?

yongtang:
@caisq @drpngx Thanks for the review and sorry for the late reply. The PR has been updated and the failed tests should have been fixed. Please take a look and let me know if there are any issues.

caisq:

martinwicke:
@ebrevdo, are your comments addressed?

ebrevdo:
if s is not a scalar this will do the wrong thing.  use s.dtype.as_numpy_dtype(1) and similar in the numpy array case

yongtang:
Thanks @ebrevdo. The PR has been updated.

ebrevdo:
LGTM so long as tests pass.

vrv:

vrv:
Sadly this did not work

ERROR: test_test_tensordot_complex128_4_4_2_True (__main__.TensordotTest)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-cpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local-opt/bin/bazel_pip/tensorflow/python/kernel_tests/tensordot_op_test.runfiles/org_tensorflow/bazel_pip/tensorflow/python/kernel_tests/tensordot_op_test.py", line 153, in test_tensordot
    c = math_ops.tensordot(a, b, axes)
  File "/workspace/pip_test/venv/local/lib/python2.7/site-packages/tensorflow/python/ops/math_ops.py", line 2413, in tensordot
    a_axes, b_axes = _tensordot_axes(a, axes)
  File "/workspace/pip_test/venv/local/lib/python2.7/site-packages/tensorflow/python/ops/math_ops.py", line 2408, in _tensordot_axes
    return axes[0], axes[1]
  File "/workspace/pip_test/venv/local/lib/python2.7/site-packages/tensorflow/python/ops/array_ops.py", line 476, in _SliceHelper
    strides.append(s.dtype.type(1))
AttributeError: 'int' object has no attribute 'dtype

:(

yongtang:
Thanks @vrv for the review. The PR has been updated to address the Jenkins failures. Please take a look.

vrv:

vrv:
The only remaining thing that would be nice is a test of this behavior.  Since I've already run the tests and it's passing, I'll merge this, but a follow up PR with a test of this behavior would be swell.

PR #9005, opened 2017-04-06
fix windows gpu build by disabling nccl which currently will not compile/work on windows

guschmue:
Windows gpu builds got broken because nccl was include in cmake (https://github.com/tensorflow/tensorflow/commit/08a3e36c97a644377c07d39d6c707d1abfb2c394). 
nccl currently does not compile/work on windows (unfortunate). To make nccl compile/work is more work so I disable the nccl op from windows for now to make the gpu build happy. There is a PR (https://github.com/NVIDIA/nccl/pull/31) for nccl that makes it mostly work on windows ... watching that and if merged we can try to make this work.

### DIFF
--- a/tensorflow/contrib/cmake/tf_core_kernels.cmake
+++ b/tensorflow/contrib/cmake/tf_core_kernels.cmake
@@ -102,6 +102,10 @@ if(WIN32)
       "${tensorflow_source_dir}/tensorflow/contrib/rnn/kernels/lstm_ops.cc"
       "${tensorflow_source_dir}/tensorflow/contrib/rnn/ops/gru_ops.cc"
       "${tensorflow_source_dir}/tensorflow/contrib/rnn/ops/lstm_ops.cc"
+      # temporarily disable nccl (nccl itself needs to be ported to windows first)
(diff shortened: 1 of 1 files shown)


GitBubble:
what happend, this bug should be fixed earlier

guschmue:
sorry, jenkins reported it yesterday but we did not get to it until today.

gunan:
CC @cwhipkey

cwhipkey:
Does accessing tf.contrib on windows still work with this change?

(in the change that enabled nccl in cmake, it was also imported by contrib/__init__.py - and that import may fail now if the op library can't be loaded).

guschmue:
yes, you can even import nccl ... doesn't throw until you call into the .so.
>>> import tensorflow.contrib.nccl as nccl
>>> dir(nccl)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'all_max', 'all_min', 'all_prod', 'all_sum', 'broadcast']

All unit tests are passing for gpu as well.

cwhipkey (approved):

cwhipkey:
Thansk for the fix!

gunan:
Jenkins, test this please.

PR #9013, opened 2017-04-06
Add hint to distinguish between assignment ops and non-assignment ops

DavidNorman:
This allows XLA backends to decide whether an op is doing an in-place
modification of a tensor. a backend can use this to ensure assignment
operations do not move data.


caisq:
gentle ping @hawkinsp

hawkinsp:
There is a discussion thread about this PR on the xla-dev Google group:
https://groups.google.com/forum/#!forum/xla-dev
and that seems like a better forum for the wider design discussion.

(It seems unlikely we will merge this PR as is.)

drpngx:
Closing then. Feel free to reopen or push a new PR that addresses the main concerns.

Issue #386, opened 2015-12-01
Implement Fast Fourier Transform ops


NickShahML:
Hey TF,

Trying to integrate the new Unitary RNN's into tensorflow. Unfortunately, they require Fast Fourier Transform and Inverse Fast Fourier Transforms. 

For efficiency this needs to be done on the GPU. On the author's code (theano), they do this by making a Theano Op, and inserting scikit's Fast Fourier and Inverse Fast Fourier Here:

https://github.com/amarshah/complex_RNN/blob/master/fftconv.py

How can this be done in TF? Upon completing the Unitary RNN, I plan to share it to everyone! I asked on the google groups, but didn't get a reply.

vrv:
We'll need to add a few FFT ops to get this done -- probably using cufft or fbfft for GPU.  Assigning to @zffchen78 since he has experience in ffts and adding ops :)

LeavesBreathe:
Thank you @vrv and @zffchen78 . Basically, these new types of RNN's are very promising and I hope to integrate them into TensorFlow over the next week if possible. I'm starting to work on it here:
https://github.com/LeavesBreathe/Seq2Seq_Upgrade_TensorFlow

However, it will not be possible to do without FFT and IFFT support so I really appreciate your help!

adder:
So would this imply that the FFT ops only work for GPU and not CPU? Or just very very slow on CPU?

LeavesBreathe:
Well the problem is that the actual unitary RNN uses FFT and IFFT within its cell. Ideally you want the cell's computation done in parallel on the GPU. If its on the CPU, its going to take forever.

Here's the paper: Look at their main equation http://arxiv.org/pdf/1511.06464v1.pdf #6

adder:
With 'taking forever', you mean compared to the uRNN GPU version or also compared to a 'classical' RNN or a LSTM with the same number of parameters? I would at least hope the speed to convergence demonstrated in the paper would outweigh any extra computation time per iteration. :)

Anyways, I would expect that all ops defined in Tensorflow would be available for both CPU and GPU (and any future device). Me and I guess others don't have day to day access to a suitable GPU. Or is that not one of the design goals?

vrv:
We'll try to provide an implementation for both.

LeavesBreathe:
Perfect! Thanks!

zffchen78:
@adder , can you elaborate a bit more of what kind of FFT/IFFT you need? Specifically,
[...]

PR #9029, opened 2017-04-06
CPU kernels for FFT (WIP)

tillahoffmann:
This PR adds CPU kernels for FFTs (cf #386). A few points to note:

* This PR does not yet support RFFTs. These can in principle be implemented by slicing such that the negative frequency components of the FFT are removed. Maybe @benoitsteiner has some better ideas though.
* Does it make sense to split up the code into multiple files to avoid having `#if GOOGLE_CUDA` half way?
* @rryan, because the tensor FFT in Eigen is templated, I couldn't avoid bloating the binary a bit as discussed via e-mail.

### DIFF
--- a/tensorflow/core/kernels/fft_ops.cc
+++ b/tensorflow/core/kernels/fft_ops.cc
@@ -15,7 +15,7 @@ limitations under the License.
 
 #define EIGEN_USE_THREADS
 
-// See docs in ../ops/fft_ops.cc.
+// See docs in ../ops/spectral_ops.cc.
 
 #include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
 #include "tensorflow/core/framework/op.h"
@@ -97,10 +97,10 @@ struct FFTFunctor {
   void operator()(const Device& d,
                   typename TTypes<TOutput, FFTRank + 1>::Tensor output,
                   typename TTypes<TInput, FFTRank + 1>::Tensor input) {
-      // Create the axes (which are always trailing).
-      auto axes = Eigen::ArrayXi::LinSpaced(FFTRank, 1, FFTRank);
-      // Evaluate the fft on the specified device.
-      output.device(d) = input.template fft<FFTResultType, FFTDir>(axes);
+    // Create the axes (which are always trailing).
+    auto axes = Eigen::ArrayXi::LinSpaced(FFTRank, 1, FFTRank);
+    // Evaluate the fft on the specified device.
+    output.device(d) = input.template fft<FFTResultType, FFTDir>(axes);
   }
 };
 
@@ -115,17 +115,63 @@ class FFTCPU : public FFTBase {
 
   void DoFFT(OpKernelContext* ctx, const Tensor& in, uint64* fft_shape,
              Tensor* out) override {
-    // TODO(tillahoffmann): Support RFFTs by slicing away the negative
-    // frequency components.
-    OP_REQUIRES(ctx, !IsReal(), errors::Internal("Real FFT not supported."));
-    auto input = ((Tensor) in).flat_inner_dims<complex64, FFTRank + 1>();
+    if (!IsReal()) {
+      auto input = ((Tensor) in).flat_inner_dims<complex64, FFTRank + 1>();
 
-    // Apply the functor.
-    FFTFunctor<CPUDevice, complex64, complex64, Eigen::BothParts,
-               Forward ? Eigen::FFT_FORWARD : Eigen::FFT_REVERSE,
-               FFTRank> functor;
-    functor(ctx->eigen_device<CPUDevice>(),
-            out->flat_inner_dims<complex64, FFTRank + 1>(), input);
+      // Apply the functor.
+      FFTFunctor<CPUDevice, complex64, complex64, Eigen::BothParts,
+                Forward ? Eigen::FFT_FORWARD : Eigen::FFT_REVERSE,
+                FFTRank> functor;
+      functor(ctx->eigen_device<CPUDevice>(),
+              out->flat_inner_dims<complex64, FFTRank + 1>(), input);
+    }
+    else {
+      if (IsForward()) {
+        auto input = ((Tensor) in).flat_inner_dims<float, FFTRank + 1>();
+        // Create a temporary placeholder for the full FFT.
+        Tensor temp;
+        OP_REQUIRES_OK(ctx,ctx->allocate_temp(
+          DataTypeToEnum<complex64>::v(), in.shape(), &temp
+        ));
+        auto full_fft = temp.flat_inner_dims<complex64, FFTRank + 1>();
+        // Apply the functor.
+        FFTFunctor<CPUDevice, float, complex64, Eigen::BothParts,
+                  Eigen::FFT_FORWARD, FFTRank> functor;
+        functor(ctx->eigen_device<CPUDevice>(), full_fft, input);
+        // Create zero indices for slicing.
+        Eigen::DSizes<Eigen::DenseIndex, FFTRank + 1> startIndices;
+        // Convert output to tensor and get tensor size for slicing.
+        auto output = out->flat_inner_dims<complex64, FFTRank + 1>();
+        auto sizes = output.dimensions();
+        // Slice the full FFT to get the non-negative frequency components only.
+        output.slice(startIndices, sizes) =
+          full_fft.slice(startIndices, sizes);
+      }
+      else {
+        auto input = ((Tensor) in).flat_inner_dims<complex64, FFTRank + 1>();
+        // The first dimension contains the zero-frequency component which we
+        // do not want to duplicate. So we reconstruct the complex signal by
+        // (1) slicing from the second element, (2) reversing the order,
+        // (3) taking the complex conjugate, (4) concatenating with the original
+        // input. Note that for an even input length, the last element is the
+        // Nyquist frequency which we also do not want to duplicate.
+        Eigen::DSizes<Eigen::DenseIndex, FFTRank + 1> startIndices;
+        startIndices[FFTRank] = 1;
+        auto sizes = input.dimensions();
+        if (sizes[FFTRank] % 2 == 0) {
+          sizes[FFTRank] -= 1;
+        }
+        auto cc = input.slice(startIndices, sizes).conjugate()
+          .reverse(FFTRank);
+        auto full_fft = input.concatenate(cc, FFTRank);
+
+        // Evaluate the IFFT
+        auto output = out->flat_inner_dims<float, FFTRank + 1>();
+        FFTFunctor<CPUDevice, complex64, float, Eigen::RealPart,
+                  Eigen::FFT_REVERSE, FFTRank> functor;
+        functor(ctx->eigen_device<CPUDevice>(), output, full_fft);
+      }
+    }
   }
 };
 
@@ -139,6 +185,18 @@ REGISTER_KERNEL_BUILDER(Name("IFFT2D").Device(DEVICE_CPU),
 REGISTER_KERNEL_BUILDER(Name("FFT3D").Device(DEVICE_CPU),
                         FFTCPU<true, false, 3>);
 REGISTER_KERNEL_BUILDER(Name("IFFT3D").Device(DEVICE_CPU),
+                        FFTCPU<false, false, 3>);
+
+REGISTER_KERNEL_BUILDER(Name("RFFT").Device(DEVICE_CPU), FFTCPU<true, true, 1>);
+REGISTER_KERNEL_BUILDER(Name("IRFFT").Device(DEVICE_CPU),
+                        FFTCPU<false, false, 1>);
+REGISTER_KERNEL_BUILDER(Name("RFFT2D").Device(DEVICE_CPU),
+                        FFTCPU<true, true, 2>);
+REGISTER_KERNEL_BUILDER(Name("IRFFT2D").Device(DEVICE_CPU),
+                        FFTCPU<false, false, 2>);
+REGISTER_KERNEL_BUILDER(Name("RFFT3D").Device(DEVICE_CPU),
+                        FFTCPU<true, true, 3>);
+REGISTER_KERNEL_BUILDER(Name("IRFFT3D").Device(DEVICE_CPU),
                         FFTCPU<false, false, 3>);
 
 #if GOOGLE_CUDA
--- a/tensorflow/core/kernels/fft_ops.cc
+++ b/tensorflow/core/kernels/fft_ops.cc
@@ -84,7 +84,7 @@ class FFTBase : public OpKernel {
   virtual bool IsForward() const = 0;
   virtual bool IsReal() const = 0;
 
-  // The function that actually computes the FFT
+  // The function that actually computes the FFT.
   virtual void DoFFT(OpKernelContext* ctx, const Tensor& in, uint64* fft_shape,
                      Tensor* out) = 0;
 };
@@ -96,17 +96,11 @@ template <typename Device, typename TInput, typename TOutput,
 struct FFTFunctor {
   void operator()(const Device& d,
                   typename TTypes<TOutput, FFTRank + 1>::Tensor output,
-                  typename TTypes<TInput, FFTRank + 1>::ConstTensor input) {
-      // Cast to non-const data type to ensure we can call the template fft
-      // function
-      // TODO(tillahoffmann): Remove reinterpret_cast once Eigen supports const
-      // arguments for tensor fft
-      auto casted_input =
-       *reinterpret_cast<typename TTypes<TInput, FFTRank + 1>::Tensor*>(&input);
-      // Create the axes (which are always trailing)
+                  typename TTypes<TInput, FFTRank + 1>::Tensor input) {
+      // Create the axes (which are always trailing).
       auto axes = Eigen::ArrayXi::LinSpaced(FFTRank, 1, FFTRank);
-      // Evaluate the fft on the specified device
-      output.device(d) = casted_input.template fft<FFTResultType, FFTDir>(axes);
+      // Evaluate the fft on the specified device.
+      output.device(d) = input.template fft<FFTResultType, FFTDir>(axes);
   }
 };
 
@@ -122,11 +116,11 @@ class FFTCPU : public FFTBase {
   void DoFFT(OpKernelContext* ctx, const Tensor& in, uint64* fft_shape,
              Tensor* out) override {
     // TODO(tillahoffmann): Support RFFTs by slicing away the negative
-    // frequency components
+    // frequency components.
     OP_REQUIRES(ctx, !IsReal(), errors::Internal("Real FFT not supported."));
-    auto input = in.flat_inner_dims<complex64, FFTRank + 1>();
+    auto input = ((Tensor) in).flat_inner_dims<complex64, FFTRank + 1>();
 
-    // Apply the functor
+    // Apply the functor.
     FFTFunctor<CPUDevice, complex64, complex64, Eigen::BothParts,
                Forward ? Eigen::FFT_FORWARD : Eigen::FFT_REVERSE,
                FFTRank> functor;
--- a/tensorflow/core/kernels/fft_ops.cc
+++ b/tensorflow/core/kernels/fft_ops.cc
@@ -27,24 +27,11 @@ limitations under the License.
 #include "tensorflow/core/platform/types.h"
 #include "tensorflow/core/util/work_sharder.h"
 
-#if GOOGLE_CUDA
-#include "tensorflow/core/platform/stream_executor.h"
-
 namespace tensorflow {
 
-namespace {
-// TODO(vrv/zhifengc): Refactor AsDeviceMemory() into GPUUtil.
-template <typename T>
-perftools::gputools::DeviceMemory<T> AsDeviceMemory(const T* cuda_memory) {
-  perftools::gputools::DeviceMemoryBase wrapped(const_cast<T*>(cuda_memory));
-  perftools::gputools::DeviceMemory<T> typed(wrapped);
-  return typed;
-}
-}  // end namespace
-
-class FFTGPUBase : public OpKernel {
+class FFTBase : public OpKernel {
  public:
-  explicit FFTGPUBase(OpKernelConstruction* ctx) : OpKernel(ctx) {}
+  explicit FFTBase(OpKernelConstruction* ctx) : OpKernel(ctx) {}
 
   void Compute(OpKernelContext* ctx) override {
     const Tensor& in = ctx->input(0);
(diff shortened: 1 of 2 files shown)


tensorflow/core/kernels/fft_ops.cc
```diff
@@ -97,9 +84,89 @@ class FFTGPUBase : public OpKernel {
...
+  bool IsForward() const override { return Forward; }
+  bool IsReal() const override { return _Real; }
+
   void DoFFT(OpKernelContext* ctx, const Tensor& in, uint64* fft_shape,
-             Tensor* out) {
+             Tensor* out) override {
+    // TODO(tillahoffmann): Support RFFTs by slicing away the negative
+    // frequency components
+    OP_REQUIRES(ctx, !IsReal(), errors::Internal("Real FFT not supported."));
+    auto input = in.flat_inner_dims<complex64, FFTRank + 1>();
```
rryan: I'd slightly prefer to avoid the above reinterpret_cast since I'm not sure converting a `TensorMap<const T, ...>*` to a `TensorMap<T, ...>*` will always be undefined-behavior free. Maybe someone more expert than me at interpreting the standard or Eigen can chime in :).

Since Tensor types are lightweight wrappers around reference-counted buffers, you can [copy](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.h#L161
) away the const-ness cheaply.
```c++
Tensor nonconst_in = in;
auto input = nonconst_in.flat_inner_dims<complex64, FFTRank + 1>();
```

tensorflow/core/kernels/fft_ops.cc
```diff
@@ -27,24 +27,11 @@ limitations under the License.
...
-template <typename T>
-perftools::gputools::DeviceMemory<T> AsDeviceMemory(const T* cuda_memory) {
-  perftools::gputools::DeviceMemoryBase wrapped(const_cast<T*>(cuda_memory));
-  perftools::gputools::DeviceMemory<T> typed(wrapped);
-  return typed;
-}
-}  // end namespace
-
-class FFTGPUBase : public OpKernel {
+class FFTBase : public OpKernel {
```
rryan: Sorry for nitpicking, can you go through the comments here and add periods? Our style guide prefers punctuation:
https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar

rryan (changes requested):
Thanks! Looks really good.

> This PR does not yet support RFFTs. These can in principle be implemented by slicing such that the negative frequency components of the FFT are removed. Maybe @benoitsteiner has some better ideas though.

Can I get you to tackle that in this PR? ;) Or were you thinking of leaving it as C2C only?

> Does it make sense to split up the code into multiple files to avoid having #if GOOGLE_CUDA half way?

I'll let a TF person chime in on code organization. It'd be nice for FFT-functor to live in it's own file (as is common in core/kernels) and the GPU stream-executor code to also be pushed into an FFTFunctor specialization for DEVICE_GPU. That would allow more code sharing between the CPU and GPU kernels. I'm not sure that needs to happen in this PR though but I don't feel it's my call.

> @rryan, because the tensor FFT in Eigen is templated, I couldn't avoid bloating the binary a bit as discussed via e-mail.

It looks fine to me as is.

tensorflow/core/kernels/fft_ops.cc
```diff
@@ -97,9 +84,89 @@ class FFTGPUBase : public OpKernel {
...
+    // Apply the functor
+    FFTFunctor<CPUDevice, complex64, complex64, Eigen::BothParts,
+               Forward ? Eigen::FFT_FORWARD : Eigen::FFT_REVERSE,
+               FFTRank> functor;
+    functor(ctx->eigen_device<CPUDevice>(),
+            out->flat_inner_dims<complex64, FFTRank + 1>(), input);
+  }
+};
+
+REGISTER_KERNEL_BUILDER(Name("FFT").Device(DEVICE_CPU), FFTCPU<true, false, 1>);
```
rmlarsen: Not sure if this would be the right solution, but one way might be to just name these kernels FFT_V2 etc. and have internal Python wrappers calling the old versions while the third_party/tensorflow code would call the new ones?

tensorflow/core/kernels/fft_ops.cc
tillahoffmann: Yes, much better.

tillahoffmann:
Alright, RFFT is added but I can't quite get the IRFFT to compile. In particular, the compiler isn't happy about passing the reconstructed full FFT (with negative frequencies) to the functor. Help from someone with more Eigen-knowledge would be great. Here's the compiler message.

```
tensorflow/core/kernels/fft_ops.cc:172:9: error: no matching function for call to object of type 'FFTFunctor<CPUDevice, complex64, float, Eigen::RealPart, Eigen::FFT_REVERSE, 3>' (aka 'FFTFunctor<Eigen::ThreadPoolDevice, complex<float>, float, Eigen::RealPart, Eigen::FFT_REVERSE, 3>')
        functor(ctx->eigen_device<CPUDevice>(), output, full_fft);
        ^~~~~~~
tensorflow/core/kernels/fft_ops.cc:198:25: note: in instantiation of member function 'tensorflow::FFTCPU<true, true, 3>::DoFFT' requested here
                        FFTCPU<true, true, 3>);
                        ^
tensorflow/core/kernels/fft_ops.cc:97:8: note: candidate function not viable: no known conversion from 'Eigen::TensorConcatenationOp<const int, Eigen::TensorMap<Eigen::Tensor<std::__1::complex<float>, 4, 1, long>, 16, MakePointer>, Eigen::TensorReverseOp<const int, const Eigen::TensorCwiseUnaryOp<Eigen::internal::scalar_conjugate_op<std::__1::complex<float> >, const Eigen::TensorSlicingOp<const Eigen::DSizes<long, 4>, const Eigen::DSizes<long, 4>, Eigen::TensorMap<Eigen::Tensor<std::__1::complex<float>, 4, 1, long>, 16, MakePointer> > > > >' to 'typename TTypes<complex<float>, 3 + 1>::Tensor' (aka 'TensorMap<Eigen::Tensor<complex<float>, 4, Eigen::RowMajor, long>, Eigen::Aligned>') for 3rd argument
  void operator()(const Device& d,
```

I've tried casting to `typename TTypes<complex64, FFTRank + 1>` to force an evaluation of the intermediate state but it wasn't happy with that either.

tensorflow/core/kernels/fft_ops.cc
```diff
@@ -97,9 +84,141 @@ class FFTGPUBase : public OpKernel {
...
+                        FFTCPU<true, true, 2>);
+REGISTER_KERNEL_BUILDER(Name("IRFFT2D").Device(DEVICE_CPU),
+                        FFTCPU<false, false, 2>);
+REGISTER_KERNEL_BUILDER(Name("RFFT3D").Device(DEVICE_CPU),
+                        FFTCPU<true, true, 3>);
+REGISTER_KERNEL_BUILDER(Name("IRFFT3D").Device(DEVICE_CPU),
+                        FFTCPU<false, false, 3>);
+
+#if GOOGLE_CUDA
+#include "tensorflow/core/platform/stream_executor.h"
```
benoitsteiner: Shouldn't we move this at the top of the file to avoid including stream_executor.h in the tensorflow namespace ?

tensorflow/python/kernel_tests/fft_ops_test.py
```diff
@@ -39,30 +39,28 @@ def _Compare(self, x, rank, fft_length=None, use_placeholder=False):
...
+    if use_placeholder:
+      x_ph = array_ops.placeholder(dtype=dtypes.as_dtype(x.dtype))
+      x_tf = self._tfFFT(x_ph, rank, fft_length, use_gpu=True,
+                          feed_dict={x_ph: x})
+    else:
+      x_tf = self._tfFFT(x, rank, fft_length, use_gpu=True)
 
-      # GPU/Forward
-      self.assertAllClose(x_np, x_tf, rtol=1e-4, atol=1e-4)
+    # GPU/Forward
```
benoitsteiner: Please update this comment, the code isn't GPU any anymore.

tensorflow/python/kernel_tests/fft_ops_test.py
```diff
@@ -39,30 +39,28 @@ def _Compare(self, x, rank, fft_length=None, use_placeholder=False):
...
+    if use_placeholder:
+      x_ph = array_ops.placeholder(dtype=dtypes.as_dtype(x.dtype))
+      x_tf = self._tfIFFT(x_ph, rank, fft_length, use_gpu=True,
+                          feed_dict={x_ph: x})
+    else:
+      x_tf = self._tfIFFT(x, rank, fft_length, use_gpu=True)
 
-      # GPU/Backward
-      self.assertAllClose(x_np, x_tf, rtol=1e-4, atol=1e-4)
+    # GPU/Backward
```
benoitsteiner: Same as above, this isn't limited to GPUs anymore.

benoitsteiner:
Jenkins, test this please.

tensorflow/core/kernels/fft_ops.cc
```diff
@@ -97,9 +84,141 @@ class FFTGPUBase : public OpKernel {
...
+        }
+        auto cc = input.slice(startIndices, sizes).conjugate()
+          .reverse(FFTRank);
+        auto full_fft = input.concatenate(cc, FFTRank);
+
+        // Evaluate the IFFT
+        auto output = out->flat_inner_dims<float, FFTRank + 1>();
+        FFTFunctor<CPUDevice, complex64, float, Eigen::RealPart,
+                  Eigen::FFT_REVERSE, FFTRank> functor;
+        functor(ctx->eigen_device<CPUDevice>(), output, full_fft);
```
rryan: Based on the compiler errors I'm seeing here:
https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu/4430/console
they all seem to be from the IRFFT here (line 172):
```
tensorflow/core/kernels/fft_ops.cc:172:65: error: no match for call to

'(tensorflow::FFTFunctor<Eigen::ThreadPoolDevice, std::complex<float>, float, 0, 1, 3>) (const Eigen::ThreadPoolDevice&, Eigen::TensorMap<Eigen::Tensor<float, 4, 1, long int>, 16, Eigen::MakePointer>&, Eigen::TensorConcatenationOp<const int, Eigen::TensorMap<Eigen::Tensor<std::complex<float>, 4, 1, long int>, 16, Eigen::MakePointer>, Eigen::TensorReverseOp<const int, const Eigen::TensorCwiseUnaryOp<Eigen::internal::scalar_conjugate_op<std::complex<float> >, const Eigen::TensorSlicingOp<const Eigen::DSizes<long int, 4>, const Eigen::DSizes<long int, 4>, Eigen::TensorMap<Eigen::Tensor<std::complex<float>, 4, 1, long int>, 16, Eigen::MakePointer> > > > >&)'
         functor(ctx->eigen_device<CPUDevice>(), output, full_fft);

no known conversion for argument 3 from 

'Eigen::TensorConcatenationOp<const int, Eigen::TensorMap<Eigen::Tensor<std::complex<float>, 4, 1, long int>, 16, Eigen::MakePointer>, Eigen::TensorReverseOp<const int, const Eigen::TensorCwiseUnaryOp<Eigen::internal::scalar_conjugate_op<std::complex<float> >, const Eigen::TensorSlicingOp<const Eigen::DSizes<long int, 4>, const Eigen::DSizes<long int, 4>, Eigen::TensorMap<Eigen::Tensor<std::complex<float>, 4, 1, long int>, 16, Eigen::MakePointer> > > > >'

to 

'tensorflow::TTypes<std::complex<float>, 4, long int>::Tensor {aka Eigen::TensorMap<Eigen::Tensor<std::complex<float>, 4, 1, long int>, 16, Eigen::MakePointer>}'
```

I think one problem is that `Eigen::TensorConcatenationOp<const int, Eigen::TensorMap<Eigen::Tensor<std::complex<float>, 4, 1, long int>, 16, Eigen::MakePointer>, Eigen::TensorReverseOp<const int, const Eigen::TensorCwiseUnaryOp<Eigen::internal::scalar_conjugate_op<std::complex<float> >, const Eigen::TensorSlicingOp<const Eigen::DSizes<long int, 4>, const Eigen::DSizes<long int, 4>, Eigen::TensorMap<Eigen::Tensor<std::complex<float>, 4, 1, long int>, 16, Eigen::MakePointer> > > > >` is not convertible to `typename TTypes<TInput, FFTRank + 1>::Tensor`. 

Maybe FFTFunctor needs to be changed to take expressions as arguments instead of TensorMaps?

I think it would also work to push the Eigen expressions here (which aren't convertible to TensorMap without an `.eval()`?) into a template specialization of FFTFunctor for TInput=float TOutput=complex64 and TInput=complex64 TOutput=float ?

tillahoffmann: @rryan, thank you for the comments. Yes, all the compiler issues arise from the IRFFT implementation. I tried forcing an evaluation using `.eval()` but got similar messages. I was wondering whether it makes sense to stick to the functor paradigm in this case because the functors are just a templated wrapper around Eigen which are not used anywhere else. What are your thoughts on the suggested implementation in https://github.com/tensorflow/tensorflow/pull/9029#issuecomment-292505411. And do you have any thoughts on the resulting [compiler error](http://stackoverflow.com/questions/43364086/struggling-to-evaluate-operation-in-the-tensor-module-of-eigen)?

rryan: > I was wondering whether it makes sense to stick to the functor paradigm in this case because the functors are just a templated wrapper around Eigen which are not used anywhere else.

I was thinking it would cut down on instantiations of the Eigen expression in the .o, but I just noticed that forward/backward is part of the TensorFFT template arguments, so we can't reuse the same C2C code path for forward and backward (unless we did the constant scaling for the backward outside of FFTFunctor and always did C2C forward inside of FFTFunctor). Dropping the functor is fine w/ me, but I'll defer to @benoitsteiner.

> What are your thoughts on the suggested implementation in https://github.com/tensorflow/tensorflow/pull/9029#issuecomment-292505411. And do you have any thoughts on the resulting [compiler error](http://stackoverflow.com/questions/43364086/struggling-to-evaluate-operation-in-the-tensor-module-of-eigen)? 

Argh, sorry -- not sure how I missed your comment, you were already a step ahead of me. 

Off hand without trying to compile your example, that smells like TensorFFT doesn't like having a float tensor as its input. 

Though from skimming [TensorFFT](https://bitbucket.org/eigen/eigen/src/f3a22f35b0444e03a1f65941800b9a2283de1398/unsupported/Eigen/CXX11/src/Tensor/TensorFFT.h?at=default&fileviewer=file-view-default#TensorFFT.h-30), it kind of looks like it's meant to convert `T` to `std::complex<T>` on the fly.

Have you already tried casting?

```c++
 auto op = input
  .cast<std::complex<float>>()
  .template fft<Eigen::BothParts, Eigen::FFT_FORWARD>(axes)
  .slice(startIndices, sizes);
```

rryan: Hm, after reading TensorFFT a bit, I'm not sure if its [TensorTraits](https://bitbucket.org/eigen/eigen/src/f3a22f35b0444e03a1f65941800b9a2283de1398/unsupported/Eigen/CXX11/src/Tensor/TensorFFT.h?fileviewer=file-view-default#TensorFFT.h-62) are defined properly. It does not define `Scalar`, only `RealScalar`, `ComplexScalar`, `InputScalar` and `OutputScalar`. So the inherited traits `traits<XprType>` defines `Scalar`. `XprType` is `Tensor<float>` in this case, so `Scalar` is `float`. Using [TensorConversion](https://bitbucket.org/eigen/eigen/src/f3a22f35b0444e03a1f65941800b9a2283de1398/unsupported/Eigen/CXX11/src/Tensor/TensorConversion.h?at=default&fileviewer=file-view-default#TensorConversion.h-27) as an example (since I'm not familiar enough with Eigen to know what this is supposed to be), it looks like it defines `Scalar` as the type casted to.

@benoitsteiner / @rmlarsen?

tillahoffmann: Great, thanks for the insight. I think I tried casting without any luck before but might give it another go. 

Another option could be to allocate a temporary tensor, do all the preprocessing, assign to the temporary tensor to evaluate, compute the IFFT of the temporary tensor. It would lose some of the elegance of Eigen but might be an acceptable workaround.

vrv:
@tillahoffmann any updates on this PR?

tillahoffmann:
@vrv, there seems to be an issue with the types declared in the FFT operator of Eigen which means that compiling the (I)RFFT fails (see @rryan's comment above). Unfortunately, I'm not sufficiently familiar with Eigen to be able to fix it/find a work around. 

We could either see whether @benoitsteiner or @rmlarsen have some ideas or we could change this PR to only provide an implementation of the complex FFT.

vrv:
Okay, @benoitsteiner or @rmlarsen let us know if you have any ideas on next steps here.

drpngx:
@b11094 are you suggesting a fix for the build failures?

ebrevdo:
@rmlarsen @benoitsteiner can you provide some guidance here re: @rryan's findings?  Would love to get this unblocked and into TF.

ebrevdo:
We could also consider having the kernels be complex-valued, and insert casting operations on the input and output as appropriate.

tensorflow/core/kernels/fft_ops.cc
rmlarsen: Scalar should denote the output type. Does it help if you rename OutputScalar to Scalar? (or add another typedef for it?

tillahoffmann:
Reconstructing the full FFT from the RFFT coefficients in Eigen is proving to be a bit tricky. The current PR implements the complex FFT, IFFT and the real FFT. The real IFFT is not implemented. I imagine someone with better Eigen knowledge than me would be better suited to implement it.

rmlarsen:
@tillahoffmann Thanks a lot for this contribution! I think this is a great step forward. If @rryan approves, I think we can merge this and add the IRFFT in a followup.

rmlarsen:

tillahoffmann:
@rmlarsen, do you know why the [GPU builds are failing](https://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/4815/consoleFull)? It's complaining that some symbols are not declared:

```
In file included from ./tensorflow/stream_executor/kernel.h:80:0,
                 from ./tensorflow/stream_executor/trace_listener.h:22,
                 from ./tensorflow/stream_executor/platform.h:30,
                 from ./tensorflow/stream_executor/cuda/cuda_platform_id.h:19,
                 from ./tensorflow/core/platform/default/stream_executor.h:23,
                 from ./tensorflow/core/platform/stream_executor.h:24,
                 from tensorflow/core/kernels/fft_ops.cc:176:
./tensorflow/stream_executor/lib/array_slice.h:25:19: error: 'tensorflow::tensorflow::gtl' has not been declared
 using tensorflow::gtl::ArraySlice;
```

tensorflow/core/kernels/fft_ops.cc
```diff
@@ -97,9 +84,114 @@ class FFTGPUBase : public OpKernel {
...
+REGISTER_KERNEL_BUILDER(Name("IFFT3D").Device(DEVICE_CPU),
+                        FFTCPU<false, false, 3>);
+
+REGISTER_KERNEL_BUILDER(Name("RFFT").Device(DEVICE_CPU), FFTCPU<true, true, 1>);
+REGISTER_KERNEL_BUILDER(Name("RFFT2D").Device(DEVICE_CPU),
+                        FFTCPU<true, true, 2>);
+REGISTER_KERNEL_BUILDER(Name("RFFT3D").Device(DEVICE_CPU),
+                        FFTCPU<true, true, 3>);
+
+#if GOOGLE_CUDA
```
rmlarsen: this conditional include should go at the top of the file, and it should be

#include "third_party/tensorflow/core/platform/stream_executor.h"

rmlarsen (changes requested):

tensorflow/core/kernels/fft_ops.cc
rmlarsen: (this fails because you are including this inside namespace tensorflow.

tillahoffmann: Ok, given that there is both CPU and GPU code, would the following work?

```cpp
namespace tensorflow {
// CPU stuff
}

#if GOOGLE_CUDA
#include "tensorflow/core/platform/stream_executor.h"

namespace tensorflow {
// GPU stuff
}
#endif
```

rmlarsen: It would, but you still need to put the include at the top of the file.

```
// regular includes ...
#if GOOGLE_CUDA
#include "tensorflow/core/platform/stream_executor.h"
#endif  // GOOGLE_CUDA

namespace tensorflow {
// CPU stuff

#if GOOGLE_CUDA
// GPU stuff
#endif  // GOOGLE_CUDA

}  // namespace tensorflow```

tillahoffmann:
@rmlarsen, let's hope this compiles. Btw, is there a way to get testing privileges on PRs?

tensorflow/core/kernels/fft_ops.cc
rryan: Hm, yea that's an option, though it'd be nice for internal graphs with FFTs to be compatible with open source TF.

rryan (approved):

rryan:
LGTM, thanks! Though I'm still curious about how to handle the internal/external issue. I think merging this will break internal code unless we ifdef the kernel to be opensource-only somehow.

drpngx:
Jenkins, test this please.

tillahoffmann:
@rryan, do you think it would be possible to make the internal code available?

rryan:
> @rryan, do you think it would be possible to make the internal code available?

Unfortunately we can't, and I'm not qualified to explain why since I'm not a lawyer. :(

rmlarsen:
@tillahoffmann thanks! Our test resources are somewhat limited, so we keep the privilege to trigger it to people in the team.

rmlarsen (approved):

rmlarsen:
@tillahoffmann thanks again for the contribution!

rryan:
thanks @tillahoffmann !

Dref360:
Could this get merged into r1.2? Seems like a feature that people are really excited about.

beniroquai:
Maybe supid question, but along which axis is the fft performed? 
Assume a 3D tensor with 2D-images along the 3rd dimension, now I'm doing a fft2d. How would I do it if I want to do it for each 2D slice seperatly? 

Most other libraries have something like `np.fft2d(image, s=None, axes=2) `or so. 
Would be nice to have this feature also in TensorFlow :)

Androbin:
@beniroquai It's common to have to batch dimension first and not last.

beniroquai:
Ahh! Nice! That's my error. So it transforms along 1st dimension? 
Imagine you have an Image-stack of 128x128x45, where z=45; I should better permute it to 45x128x128? 
Thank you! :)

tillahoffmann:
@beniroquai, yes the batch dimension (45 in your case) should be the first dimension. The n-dimensional fft will transform the *last* n dimension. I don't expect to see axis support in tensorflow any time soon because the GPU kernel implementation seems to rely on the transform being along the last dimensions--but I'm happy to stand corrected.

rryan:
> I don't expect to see axis support in tensorflow any time soon because the GPU kernel implementation seems to rely on the transform being along the last dimensions--but I'm happy to stand corrected.

I'd love to add an axis parameter. If someone from the community would like to add it I think it'd be welcomed by the team. I doubt it's high on anybody's TODO list at the moment though.

I think the code is already generic enough to support it for GPU since we already specify the "advanced data layout" parameters to configure it to do the last N dimensions.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/fft_ops.cc#L308
http://docs.nvidia.com/cuda/cufft/index.html#advanced-data-layout

In the worst case, we could perform multiple serial 1D FFTs.

PR #9040, opened 2017-04-07
set save_model_secs=60 in mnist_replica.py

lvjiangzhao:
This would make restoring from a checkpoint more easily. The default value for save_model_secs is 600 which is relatively too long for this tutorial example model.

### DIFF
--- a/tensorflow/tools/dist_test/python/mnist_replica.py
+++ b/tensorflow/tools/dist_test/python/mnist_replica.py
@@ -206,14 +206,16 @@ def main(unused_argv):
           init_op=init_op,
           local_init_op=local_init_op,
           ready_for_local_init_op=ready_for_local_init_op,
-          recovery_wait_secs=1,
+          save_model_secs=60,
+          recovery_wait_secs=10,
           global_step=global_step)
     else:
       sv = tf.train.Supervisor(
           is_chief=is_chief,
           logdir=train_dir,
           init_op=init_op,
-          recovery_wait_secs=1,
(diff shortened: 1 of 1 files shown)


lvjiangzhao:
I signed it!

tensorflow/tools/dist_test/python/mnist_replica.py
```diff
@@ -206,14 +206,16 @@ def main(unused_argv):
           init_op=init_op,
           local_init_op=local_init_op,
           ready_for_local_init_op=ready_for_local_init_op,
-          recovery_wait_secs=1,
+          save_model_secs=60,
```
caisq: This code here is for testing gRPC (distributed) sessions only. Can you explain why you need this change?

caisq (changes requested):

tensorflow/tools/dist_test/python/mnist_replica.py
caisq: Sorry - more clarification: I saw and understand your explanation above. But I don't understand why you need to change this testing code under tools/dist_test.

caisq:
Closing PR due to no author response. Feel free to re-open it if you'd like to work on it further.

tensorflow/tools/dist_test/python/mnist_replica.py
lvjiangzhao: Sorry for the delay.  It's OK to close this because I was just trying to test some functions with this script out of it. :)

PR #9043, opened 2017-04-07
flatten inputs in tf.layers.dense

mtjuney:
In the document of `tf.layers.dense` is writen that: 

> Note: if the inputs tensor has a rank greater than 2, then it is flattened prior to the initial matrix multiply by kernel.

but inputs is not flattened.

I modified that to flatten inputs.

### DIFF
--- a/tensorflow/python/layers/core.py
+++ b/tensorflow/python/layers/core.py
@@ -108,15 +108,17 @@ class Dense(base._Layer):  # pylint: disable=protected-access
       raise ValueError('Inputs to `Dense` should have known rank.')
     if len(input_shape) < 2:
       raise ValueError('Inputs to `Dense` should have rank >= 2.')
-    if input_shape[-1].value is None:
-      raise ValueError('The last dimension of the inputs to `Dense` '
+    if None in [s.value for s in input_shape[1:]]:
+      raise ValueError('The dimensions other than first of the inputs to `Dense` '
                        'should be defined. Found `None`.')
     # Note that we set `trainable=True` because this is a trainable
     # weight of the layer. If the layer is not trainable
     # (self.trainable = False), the variable will not be added to
     # tf.trainable_variables(), and self.trainable_weights will be empty.
+    kernel_input_shape = six.moves.reduce(lambda x,y: x*y, input_shape.as_list()[1:])
+
     self.kernel = vs.get_variable('kernel',
-                                  shape=[input_shape[-1].value, self.units],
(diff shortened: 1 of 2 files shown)


drpngx:
@fchollet could you take a look?

ebrevdo:
Not sure if that the original docstring is correct.  Nor that this PR fixes
anything; more likely that it restricts the behavior of tf.layers.dense.

the current behavior is:

input is [a, b, c, ..., m], and num_units = 10

output is: [a, b, c, ..., 10]

(here a, b, c, ..., need not be known at graph build time)

this PR seems to break that guarantee.

fchollet:
Eugene is right, the goal with the `Dense` layer is that we apply a densely-connected layer on the last axis of the input. This allows e.g. time-distributed densely-connected transformations, where the same transformation is applied at every timestep of an input. 

So we don't want to flatten the output. And besides this overarching conceptual issue, there are also many technical problems with this PR.

If the current docstring is ambiguous, then we should update it to make the layer behavior clearer. But the behavior as currently implemented is the proper one.

joyousrabbit:
Hi, all, I searched everywhere but cannot find the difference between ft.layers.dense and tf.contrib.layers.fully_connected. 

Could you please point me out their difference?

And if dense only applies to the last axis, how can we do a real full connections of elements from all dims like in Cafffe? In Caffe, input is [N, b, c, ..., m], and num_units = 10, then output is: [N,10]

fchollet:
> And if dense only applies to the last axis, how can we do a real full connections of elements from all dims like in Cafffe?

You could either flatten your input then apply a `Dense` layer, or write a custom layer that does the same.

> Could you please point me out their difference?

They are meant to be the same. They do have a slightly different API and a different implementation. In code going forward, you should use `tf.layers` rather than `tf.contrib.layers`, since `tf.layers` is guaranteed to be API-stable (as part of the 1.0 API) and to stay there for the long term.

joyousrabbit:
@fchollet Thanks. Can we add a parameter to Dense (full_connected) layer to make it caffe mode for all axis but the first one? You see it's a very often meeted feature as the last layer.

fchollet:
@joyousrabbit it's really simple to just flatten your inputs... 

In Keras:

```python
x = keras.layers.Flatten()(x)
x = keras.layers.Dense(units)(x)
```

joyousrabbit:
@fchollet Yes, I see in tensorflow, there is a same layer do the flatten. Thanks a lot.
tf.contrib.layers.flatten

Issue #7662, opened 2017-02-19
Add a dynamic_partial_sum operator to tensorflow?


zasdfgbnm:
Hi,

In my application I need to do operations that dynamically sum some rows of a matrix to get a new matrix.  There will be an input tensor named "index" that guides which part of the tensor to be summed. 

An example is, if the input matrix is
```python
[ [   1,   1,   1],
  [  10,  10,  10],
  [ 100, 100, 100] ]
```
And the index is
```python
[ [ 0, 2 ],
  [ 2, 3 ]]
```
which simply says the output tensor will have two rows (because the "index" have two rows), the first row is the sum of rows with row number i that satisfies `0 <= i < 2` in the input, and the second row is the sum of rows with row number 2. So the result should be
```python
[ [  11,  11,  11],
  [ 100, 100, 100] ]
```

I don't find any existing operation that does this job, so I implement it (support only 2d matrix and GPU) by myself.  Since I already have an implementation, I'm not requesting a new feature here. But I do want to know that if the tensorflow team is interested in adding this operation as part of tensorflow. If the answer is yes, I will add CPU support (maybe also xla? I have no idea on how to add xla support yet), and then create a pull request for that.

aselle:
@ebrevdo, @rmlarsen, what do you two think? What would we call such an operation. It does seem useful. I've wanted to do this type of thing once or twice, even.

ebrevdo:
Is there a numpy equivalent function or composition of functions that do
this?

zasdfgbnm:
There is something similar:
https://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.ufunc.reduceat.html

zasdfgbnm:
To be honest, numpy's `numpy.ufunc.reduceat` is so confusing and unnatural to me. I prefer the behavior I define here at this issue.

aselle:
This functionality seems useful to me. Marking as contributions welcome.

zasdfgbnm:
This is the branch I'm working on:
https://github.com/zasdfgbnm/tensorflow/tree/dynamic_sum

I'm still struggling on the organization of codes at tensorflow, I will explore myself, but it will take time.

Do I put the source files to the correct directory, i.e. `tensorflow/contrib/framework/kernels`?

girving:
Isn't this a scan, a gather, and a subtraction?

zasdfgbnm:
[...]

PR #8954, opened 2017-04-04
Add TensorFlow equivalent to np.repeat

zycdragonball:
This is requested in #8246 

I'll add the gradient op soon.

This is my first TensorFlow PR. Please let me know if I made any mistake. Thanks!

PR #9063, opened 2017-04-08
Ops and kernels for reduce_slice_ops

zasdfgbnm:
This is a PR for https://github.com/tensorflow/tensorflow/issues/7662

This PR implements `reduce_slice_sum`, `reduce_slice_prod`, `reduce_slice_max`, `reduce_slice_min`.

There are also tests for these new ops.

No gradients are implemented now, but I will start working on that after this PR is merged.


girving:
This is implementable with current ops by doing a scan followed by a gather.  Is that not sufficient for your purposes?

drpngx:
I guess @zasdfgbnm has a point on the #7662. @aselle @girving what do you think?

girving:
@drpngx Yep, @zasdfgbnm and @aselle convinced me the op is reasonable, though I switched the reviewer to Andy.

caisq:
Gentle ping @aselle ?

aselle:
I'm sorry I'll take a thorough look today.

aselle:
Before merging this, can we think about if this will ever be useful to extend:
1. to higher order tensors than matrices. 
2. to axes that are not 0 and also if there are multiple axes?
If those things make sense and we can define them well. I'd like to make this function have a signature in the final form of that concept, so that we do not need to change the API to add that functionality.

It seems simple that we could define partial products of axis=1 right now where we sum over the columns rather than rows at least. This suggests we should have an axis parameter.

tensorflow/core/ops/math_ops.cc
```diff
@@ -2365,4 +2365,132 @@ output_max: the computed max output.
...
+REGISTER_OP("PartialProd")
+    .Input("data: T")
+    .Input("indices: Tindices")
+    .Output("output: T")
+    .Attr("T: numbertype")
+    .Attr("Tindices: {int32,int64}")
+    .SetShapeFn(ParticalReductionShapeFn)
+    .Doc(R"doc(
+Dynamically compute the product over the first dimension of a tensor according
+to start and end indices specified at "index". For example if input is
```
aselle: <pre>
"index" -> `indices`
</pre>

tensorflow/core/ops/math_ops.cc
```diff
@@ -2365,4 +2365,132 @@ output_max: the computed max output.
...
+[[1,2,3],[10,20,30],[100,200,300],[10,10,10]] and index is [[0,1],[1,1],[0,2]],
```
aselle: Please use an example with different values in each column to make the example more illustrative.

tensorflow/core/ops/math_ops.cc
```diff
@@ -2365,4 +2365,132 @@ output_max: the computed max output.
...
+    .Input("data: T")
+    .Input("indices: Tindices")
+    .Output("output: T")
+    .Attr("T: numbertype")
+    .Attr("Tindices: {int32,int64}")
+    .SetShapeFn(ParticalReductionShapeFn)
+    .Doc(R"doc(
+Dynamically compute the maximum over the first dimension of a tensor according
+to start and end indices specified at "index". For example if input is
+[[1,2,3],[10,20,30],[100,200,300],[10,10,10]] and index is [[0,1],[1,1],[0,2]],
```
aselle: It might be worth using pretty print to make the matrices multi line so they are more readable see array_ops.cc:884 "MatrixBandPart"

tensorflow/core/ops/math_ops.cc
```diff
@@ -2365,4 +2365,132 @@ output_max: the computed max output.
...
+  TF_RETURN_IF_ERROR(c->WithRankAtLeast(c->input(0), 1, &handle));
+  // shape of output tensor
+  TF_RETURN_IF_ERROR(c->ReplaceDim(handle, 0, c->Dim(c->input(1),0), &handle));
+  c->set_output(0, handle);
+  return Status::OK();
+}
+
+} // namespace
+
+REGISTER_OP("PartialSum")
```
aselle: Perhaps we should call this family of ops 
PartialReduceSum, PartialReduceProd, PartialReduceMin, PartialReduceMax so it is analogous to reduce_min, reduce_max, ...

tensorflow/python/ops/math_ops_test.py
```diff
@@ -423,5 +423,72 @@ def testConsistent(self):
...
+
+  def testPartialMin(self):
+    x = np.array([[1, 2, 3], [4, 5, 6], [7,8,9]], dtype=np.int32)
+    indices = np.array([[0, 1], [0, 3], [1, 2], [1, 3], [0,2]], dtype=np.int32)
+    result = np.array([[1, 2, 3], [1, 2, 3], [4, 5, 6], [4, 5, 6], [1, 2, 3]], dtype=np.int32)
+    with self.test_session(use_gpu=True):
+      y_tf = math_ops.partial_min(x,indices).eval()
+      self.assertAllEqual(y_tf, result)
+
+  def testPartialReduceEmptyDataRows(self):
```
aselle: thanks for testing this case.

tensorflow/python/ops/math_ops_test.py
```diff
@@ -423,5 +423,72 @@ def testConsistent(self):
...
+      self.assertAllEqual(y_tf, result)
+
+  def testPartialReduceEmptyIndicesRows(self):
+    x = np.array([[1, 2, 3], [4, 5, 6], [7,8,9]], dtype=np.int32)
+    indices = np.empty((0,2), dtype=np.int32)
+    result = np.empty((0,3), dtype=np.int32)
+    with self.test_session(use_gpu=True):
+      y_tf = math_ops.partial_sum(x,indices).eval()
+      self.assertAllEqual(y_tf, result)
+
```
aselle: test scalars for indices and values?

aselle (commented):
Not quite done reviewing yet...

aselle:

zasdfgbnm:
Response to @aselle 
1. Higher order tensor is actually already supported now.
2. This is a great point, I think I will change the code to support this. It won't require too much work to do so.

tensorflow/python/ops/math_ops_test.py
zasdfgbnm: Do you mean the case where "data" and "indices" are scalars? In my shape function, I explicitly forbid this case. See math_ops_test.cc: 493-495 in this PR

tensorflow/core/ops/math_ops.cc
zasdfgbnm: Yes, I didn't think too much when naming this. Do you think naming them as RangedReduceSum, RangedReduceProd, RangedReduceMin, RangedReduceMax would be even better? The word "partial" is too general and the ops I'm adding can not do all types of "partial" reduction (for example adding all elements with odd index)

aselle: It almost is a ReduceSliceSum becaue it's like you are taking a slice in one dimension and reducing along it? no?

zasdfgbnm: Yes, "Slice" sounds better. Let's call it ReduceSliceSum for now before we can think of any better name.

zasdfgbnm:
When implementing the reduction along any axis, I find it helpful to flat both inner and outer dims in the shape. So I create a new method `flat_inner_outer_dims` for class `Tensor` in a separate PR https://github.com/tensorflow/tensorflow/pull/9452

drpngx:
Jenkins, test this please.

zasdfgbnm:
@drpngx this hasn't been done (actually still a lot to do). I just merge with master in order to use merged changes I made.

zasdfgbnm:
Before changing this to add support for user specified axis, I would like to add a `GetCuda3DLaunchConfig` to `tensorflow/core/util/cuda_kernel_helper.h`, in a separate PR. The newly added function will be used here.

drpngx:
Sounds good.

rmlarsen:
@zasdfgbnm any update on your progress with GetCuda3DLaunchConfig?

zasdfgbnm:
@rmlarsen There is some part of the existing code that makes me not feeling well.
See https://github.com/tensorflow/tensorflow/issues/9937
Once everything is clear, I think it won't take too long to get a `GetCuda3DLaunchConfig`

rmlarsen:
@zasdfgbnm OK, thanks for the update! I'll mark this as stalled for now.

zasdfgbnm:
@aselle I think all my changes are done for this PR, please continue the review process. This op now allows users to specify the axis. I also added one test case to test when `axis != 0` and one test case to test when `rank > 2`. All these tests pass on my machine.

This PR relies on https://github.com/tensorflow/tensorflow/pull/10032 and https://github.com/tensorflow/tensorflow/pull/10068 that has not been merged into master branch yet. I manually merge them into this branch to make this work. So you don't have to worry about changes in `shape_inference.cc` and `cuda_kernel_helper.h` because they are reviewed separately. You may not want to merge this before these two dependencies are merged.

zasdfgbnm:
By the way, this PR include a fix in `tensor.h` related to const tensors

shoyer:
Just to note, #8954 adds `tf.repeat`, which provides gradients for `partial_sum` (and vice-versa). See discussion at https://github.com/tensorflow/tensorflow/pull/8954#discussion_r118302537.

zasdfgbnm:
@shoyer The name of `partial_sum` has been suggested to change to `reduce_slice_sum`, see https://github.com/tensorflow/tensorflow/pull/9063#discussion_r113259298
The `tf.repeat` and `tf.reduce_slice_sum` are gradients of each other regardless of the API difference. But the current API of `tf.repeat` and `tf.reduce_slice_sum` makes it a little bit tricky to use directly:

- In order to use `tf.reduce_slice_sum` as gradient of `tf.repeat`, there must be another op to compute the `indices` that `tf.reduce_slice_sum` takes.
- In order to use `tf.repeat` as gradient of `tf.reduce_slice_sum`, there must be another op to reorder and merge some gradients, and compute the `repeats` argument that `tf.repeat` takes.

zasdfgbnm:
Currently, `indices` can only be a `n*2` matrix where the first column is the starting index and the second column is the ending index. It would be nice to extend the `indices` argument of `tf.reduce_slice_***` here to support a vector input. In the case of vector input, the second column of row `i` will be treated as the first column of row  `i+1`. For example:
```python
indices = [0,5,10,15]
``` 

is equivalent to

```python
indices = [ [0,5],
            [5,10],
            [10,15],
            [15,+infinity]]
```
This extension would make `tf.reduce_slice_***` more useful. And with this extension, the gradient of `tf.repeat` would become a very simple like
```python
tf.reduce_slice_sum(grad, tf.scan(lambda a, x: a + x, repeats)))
```

@zycdragonball I think in your pull request, you can submit a `tf.repeat` without implementation of gradient and wait for this PR to be merged. What do you think?

zasdfgbnm:
The extend is done

zycdragonball:
@zasdfgbnm I'll leave the gradient of `tf.repeat` as it is now, and change it to `tf.reduce_slice_sum` once it is merged.

drpngx:
Jenkins, test this please.

drpngx:
@zasdfgbnm this doesn't seem to build. Could you fix and push again?

zasdfgbnm:
@drpngx It should be fixed now.

drpngx:
Jenkins, test this please.

drpngx:
@aselle WDYT? If OK, then we can trigger an API review, or recommend it go to contrib.

aselle:
I'd like to put these into contrib for now. The reason being that we have repeat going into contrib, and this is needed for the gradients of repeat() and repeat() is needed for gradient here. Once they are both in contrib,  the gradients should be added, and once things have stabilized we should do api review and bring them in.

drpngx:
Jenkins, test this please.

zasdfgbnm:
Not done yet

zasdfgbnm:
done now. should be working

_Sent from my OnePlus ONEPLUS A3000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

vrv:

zasdfgbnm:
seems to be unrelated fails?

rmlarsen:

rmlarsen:
@zasdfgbnm It looks like you need to add your op to the Cmake build. 
https://ci.tensorflow.org/job/tensorflow-pr-win-cmake-py/3768/consoleFull

zasdfgbnm:
@rmlarsen I added the cmake build rules. Hopefully it will work. I'm unable to test it on my machine because I'm running into this issue: https://github.com/tensorflow/tensorflow/issues/12018

drpngx:
Jenkins, test this please.

rmlarsen:
@zasdfgbnm Thanks!

rmlarsen (approved):

rmlarsen:
@zasdfgbnm Thanks a lot for this nice contribution!

tensorflow/contrib/reduce_slice_ops/python/ops/reduce_slice_ops.py
```diff
@@ -0,0 +1,30 @@
...
+from __future__ import division
+from __future__ import print_function
+
+from tensorflow.contrib.util import loader
+from tensorflow.python.platform import resource_loader
+
+_reduce_slice_ops = loader.load_op_library(
+    resource_loader.get_path_to_datafile("_reduce_slice_ops.so"))
+
+reduce_slice_sum = _reduce_slice_ops.reduce_slice_sum
```
andrewharp: This is giving me a problem with an internal build:
```
    reduce_slice_sum = _reduce_slice_ops.reduce_slice_sum
AttributeError: 'module' object has no attribute 'reduce_slice_sum'
```
Is there an extra step necessary somewhere to expose the ops completely?

zasdfgbnm: Hi @andrewharp, I didn't see any problem on this line on my computer. But I do see another problem on somewhere else. I added some more commits to my branch. See the file changes in:
https://github.com/tensorflow/tensorflow/compare/master...zasdfgbnm:partial_reduces

Do you think this will fix the problem you are talking about?

PR #9070, opened 2017-04-08
fix example typo (too many brackets)

ajschumacher:
The example was like `json.dumps({{}})`, which fails because `{{}}` is not a valid data structure.

Also, just below this, there is `assert config.master == 'host4:2222'`, which I believe would fail since that value comes back as `'grpc://host4:2222'`, but I'm less sure about recommending this change.


caisq (approved):

caisq:

PR #9079, opened 2017-04-09
Add output_shape parameter to transposed2d_convolution layer

marcociccone:
I just added this parameter that was missing in the `conv2d_transpose` layer.
If `None`, then the layer acts as usual returning the smallest shape possible


marcociccone:
I signed it!

PR #9093, opened 2017-04-10
Update input.py

MartinMoon:
When the input tensor of the tf.train.string_input_producer() is not fully defined (as a tf.placeholder(tf.string, ...)), then the element_shape is set as '[ ]'. There will raise a ValueError in line 150 "element_shape = input_tensor.get_shape()[1:].merge_with(element_shape)", since the input of the element is '[ ]', shapes will be not compatible. So I suggest the default 'element_shape' value should be set as 'None'.

### DIFF
--- a/tensorflow/python/training/input.py
+++ b/tensorflow/python/training/input.py
@@ -221,7 +221,7 @@ def string_input_producer(string_tensor,
       string_tensor = array_ops.identity(string_tensor)
     return input_producer(
         input_tensor=string_tensor,
-        element_shape=[],
+        element_shape=None,
         num_epochs=num_epochs,
         shuffle=shuffle,
         seed=seed,


MartinMoon:
I signed it!

drpngx:
It looks like this was added intentionally at 117583214. @mrry care to comment?

mrry:
If you look at the docstring for `tf.train.string_input_producer()`, `string_tensor` is specified as being 1-D, so the element shape is always `[]`. Switching it to `None` would inhibit later shape inference, for no apparent benefit.

I'm not sure what the original error message is, so it's possible that there's something we could improve, but this isn't the way to do it. Can you share more details or point to a relevant issue?

MartinMoon:
Hi mrry, thanks for your reply. I got the original error message when I try to feed a *string tensor* to a *tf.train.string_input_producer()*:

```python
input_placeholder = tf.placeholder(tf.string, shape=[None])
name_queue = tf.train.string_input_producer(input_placeholder)
...

with tf.Session() as sess:
        
        coord = tf.train.Coordinator()
        threads = tf.train.start_queue_runners(coord=coord)
        ....
        images = ['1.jpg', '2.jpg', ....]
        sess.run(predict,  feed_dict={input_placeholder:images})

```
The program will got the *InvalidArgumentError*, which seems the input images can not be passed to initialized the queue of file names.

At first, I think it is because of the shapes are not compatible, so I suggest the default 'element_shape' value should be set as 'None'. Finally I found this not the root of the problem. The non-compatible problem is due to my silly code which set the wrong placeholder shape as *tf.placeholder(tf.string, shape=[None,1])* .....

Come back to the *InvalidArgumentError* problem, I found it is because when *tf.train.string_input_producer* is called, a queue is created and return back, and implicitly, a *enqueue_many* operation is added into the queue_runner, combined with the queue, as follow:

```python
    enq = q.enqueue_many([input_tensor])
    queue_runner.add_queue_runner(
            queue_runner.QueueRunner(
            q, [enq], cancel_op=cancel_op))
```
So in the session module, when tf.train.start_queue_runners(coord=coord) is called, the string_name queue is run as a independent thread, when this thread excute *enqueue_many* operation, there will be a *InvalidArgumentError* problem, since there is no feed_dict.

So, as a summary, I think if the  *tf.train.string_input_producer()* is designed to only accept constant tensors (or return from *tf.train.match_filenames_once()*), this is OK. You can just close this pull request :sweat_smile:. If it is also want to accept input from *tf.placeholder*, there maybe need some change...

Thanks for your patience :smile:

mrry:
If I understood your comment correctly, @MartinMoon, I think the next action here is to close this PR without merging. The queue-based input producer methods and placeholders are not really compatible, because the queue runners that get set up in the background do not know how to feed the placeholders, and it would require major API changes to reconcile them.

PR #9122, opened 2017-04-11
Mofify seq2seq.py

oxwsds:
Modify the tf.contrib.legacy_seq2seq.embedding_attention_seq2seq so that it allows user to explicitly provide a second cell for the encoding step, and if it's None then to fallback on the deepcopy.


yifeif:

PR #9124, opened 2017-04-11
[CMake] Optionally support building TF as a shared lib

vit-stepanovs:
Currently, TF is already built as a shared library that is included in the Python package. However, that library:

1. Is implicitly linked against Python libs, and thus expects Python to be present on the machine wherever it is used. It is undesirable for scenarios not requiring Python (e.g. native application that need TensorFlow).
2.  Does not include the TF C++ API.

This PR allows optionally building TF as a stand-alone DLL that does not have the above issues. I am also working on allowing CMake to link all C++ tests against that DLL, and will submit such changes in a separate PR.

As a bonus, this PR also fixes a build break for tf_tools.cmake when GPU is enabled.


mrlzla:
It seems like you save my life ;) Thank you!

tomjaguarpaw:
@vit-stepanovs I should run cmake with `-DBUILD_SHARED_LIB=ON` but then what build target do I run?

vit-stepanovs:
@tomjaguarpw, the actual option name is tensorflow_BUILD_SHARED_LIB, not BUILD_SHARED_LIB. If you use that, cmake generates a project tensorflow.vcxproj (not tensorflow.sln!) which you can use to build the DLL.

drpngx:
Nice!

LGTM, @mrry what do you think?

drpngx:
Jenkins, test this please.

mrry (approved):
Looks good to me too!

drpngx:
Thanks!

ivan-aksamentov:
It works great on both Linux and Windows for me. Why is this PR removed from release 1.1.0? (it was there in 1.1.0-rc2)

drpngx:
It's probably because it didn't make it before the cutoff point when we forked the branch.

/CC: @av8ramit

av8ramit:
I don't believe this was in 1.1.0-rc2, but yes we decided to push it back to 1.2 since it was merged too late for 1.1 to test properly.

mrlzla:
Is there some way to build tensorflow as static lib?

adennie:
I tried this in conjunction with https://github.com/tensorflow/tensorflow/pull/9666 to try to build a debug DLL,, but the generated .def file had over 120K symbols, so it exceeded the 65535 limit.  Any way to filter more aggressively?

drpngx:
@girving another reason to restrict symbols #9525

On May 12, 2017 1:24 PM, "Andy Dennie" <notifications@github.com> wrote:

I tried this in conjunction with #9666
<https://github.com/tensorflow/tensorflow/pull/9666> to try to build a
debug DLL,, but the generated .def file had over 120K symbols, so it
exceeded the 65535 limit. Any way to filter more aggressively?

—
You are receiving this because you modified the open/close state.
<https://github.com/tensorflow/tensorflow/pull/9124#issuecomment-301176827>,
or mute the thread

girving:
Nice.  Is there a thread I should notify once we make progress here?  This one seems merged.

PR #9140, opened 2017-04-11
Update issue template (thanks jart@) and add env collection script.

aselle:


ISSUE_TEMPLATE.md
```diff
@@ -1,14 +1,31 @@
...
+- *TensorFlow version* (use command below):
 - *Bazel version (if compiling from source)*:
 - *CUDA/cuDNN version*:
 - *GPU Model and Memory*:
 - *Exact command to reproduce*:
 
+You can collect some of this information using our environment capture script https://github.com/tensorflow/tensorflow/blob/master/tools/
+You can collect the TensorFlow version with
+```sh
+python -c "import tensorflow as tf; print (tf.__git_version__, tf.__version__)"
```
drpngx: Could we use GIT_VERSION and VERSION instead?

tools/tf_env_collect.sh
```diff
@@ -0,0 +1,83 @@
...
+
+echo >> $OUTPUT_FILE
+echo '== check for virtualenv =========================================' >> $OUTPUT_FILE
+python -c "import sys;print(hasattr(sys, \"real_prefix\"))" >> $OUTPUT_FILE
+
+echo >> $OUTPUT_FILE
+echo '== tensorflow import ============================================' >> $OUTPUT_FILE
+cat <<EOF > /tmp/check_tf.py
+import tensorflow as tf;
+print("tf.__version__ = %s" % tf.__version__)
```
drpngx: same here

tools/tf_env_collect.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
```
drpngx: bash -u

drpngx (changes requested):

tools/tf_env_collect.sh
aselle: That does not seem to work. What is it supposed to do?

ISSUE_TEMPLATE.md
aselle: ok

tools/tf_env_collect.sh
aselle: ok

drpngx:
It generates an error if a variable is undefined. That's good for spotting
typos. You can set -u in the script.

The style guide mandates -eu, but -e which causes the script to end on any
non success command is too harsh in that case

aselle:
It seems I can do
```
#!/bin/bash -u
```
but I don't think I can use env (which gives robustness to bash not installed in /bin)
```
#!/usr/bin/env bash -u
```
doesn't work because the -u doesn't get sent to bash.

drpngx:
OK. From the bash script you can just do set -u first thing.

aselle:
@drpngx, your changes are now addressed I believe.

drpngx (approved):
Sorry I missed this

drpngx:
Jenkins, test this please.

PR #9145, opened 2017-04-11
Fix loading metadata in tensorboard embedding projector.

darrengarvey:
Fixes loading metadata (eg. vocab files, sprites) for display in
tensorboard's embedding projector plugin.

Since [1], `project_plugin.pbtxt` is looked for under the directory
`$LOGDIR/plugins/org_tensorflow_tensorboard_projector`, different to the
docs and where `projector.visualise_embeddings()` stores the file.
Revert to the previous location.

[1] See #9059, commit 52dcb259

### DIFF
--- a/tensorflow/tensorboard/plugins/projector/projector_plugin.py
+++ b/tensorflow/tensorboard/plugins/projector/projector_plugin.py
@@ -272,9 +272,10 @@ def _assets_dir_to_logdir(assets_dir):
 def _latest_checkpoints_changed(configs, run_path_pairs):
   """Returns true if the latest checkpoint has changed in any of the runs."""
   for run_name, assets_dir in run_path_pairs:
+    logdir = _assets_dir_to_logdir(assets_dir)
     if run_name not in configs:
       config = projector_config_pb2.ProjectorConfig()
-      config_fpath = os.path.join(assets_dir, PROJECTOR_FILENAME)
+      config_fpath = os.path.join(log_dir, PROJECTOR_FILENAME)
       if file_io.file_exists(config_fpath):
         file_content = file_io.read_file_to_string(config_fpath)
         text_format.Merge(file_content, config)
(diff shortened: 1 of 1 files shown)


drpngx:
OK, checking with author of that commit.

dsmilkov:
Hi @darrengarvey,

The change we made is about an oncoming higher-level API that should be backwards compatible. Did it stop working for you? Can you share with us the directory structure of your `LOGDIR` (no contents, just file/dir names will suffice)  so we can reproduce the error?

Thanks!

darrengarvey:
Thanks @dsmilkov. I can get the exact structure later, but `LOGDIR` is the directory given to a `SummaryWriter` and the same passed to `tensorboard --logdir $LOGDIR`. It contains `events-*`, `checkpoint`, the run meta stuff as normal.

I've added a `metadata.tsv` in the correct format and a `projector_config.pbtxt` that looks like:

``
embeddings {
  tensor_name: 'word_embedding'
  metadata_path: '/full/path/to/LOGDIR/metadata.tsv'
}
``

(as per. https://www.tensorflow.org/get_started/embedding_viz)
(also as per. [`projector.visualize_embeddings()`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/tensorboard/plugins/projector/__init__.py#L39))

This doesn't work for me (ie. the metadata isn't returned from the API call that tensorboard frontend makes). As in, the metadata isn't shown, even though the embedding projector otherwise does work.

If I move the `projector_config.pbtxt` to `$LOGDIR/plugins/org_tensorflow_tensorboard_projector` then the metadata shows up, but that's a very strange path and not what the docs suggest.

If I move [this line](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tensorboard/plugins/projector/projector_plugin.py#L364) - which was added in the referenced commit - to the bottom of the function then it also works, but presumably breaks on the next time `_latest_checkpoints_changed()` is called. That seemed like a hack, hence the suggested "fix".

dsmilkov:
Thanks for helping identify the problem. We can't accept your fix unfortunately since it won't work for the oncoming new API, but we have a fix internally that should surface on Github as a commit within couple of days.

Issue #7715, opened 2017-02-20
Feature: nadam optimizer 


cancan101:
Adam with nesterov momentum. 

Keras has it: https://keras.io/optimizers/#nadam
TF implementation (old and doesn't quite work) : https://github.com/tdozat/Optimization/blob/master/tensorflow/nadam.py

inflation:
I'm thinking about the same approach like `MomentumOptimizer` with a `use_nesterov` flag. But I don't know if it is still preferable to change the interface now.

poxvoculi:
Is there a reason this shouldn't simply be a community supplied option, like the cited version?   Are you suggesting a specific modification to TF?  Otherwise perhaps it would be better to discuss such an extension on stack overflow.

cancan101:
@poxvoculi Do you think it makes sense to add a `use_nesterov` flag into Adam (in the tf repos)?

poxvoculi:
@ebrevdo: who would have an informed opinion on this?

inflation:
@poxvoculi A C++ implementation would be faster than python hacks like above. Also, maybe an awaiting googler label?

ebrevdo:
Paul: probably vincent would know how popular / useful this optimizer is
for the general case.

vincentvanhoucke:
Plausible. I have no experience with it. If someone implements it, it sounds like something we would incorporate.

girving:
Automatically closing due to lack of recent activity. Please update the issue when new information becomes available, and we will reopen the issue. Thanks!

PR #8405, opened 2017-03-14
NADAM Optimizer

louiehelm:
Implements Nadam - a published variant of Adam Optimizer with Nestov momentum

Same basic algorithm used by the Nadam Optimizer in Keras (https://keras.io/optimizers/#nadam) but implemented from scratch by me for TF's dense / gpu kernels

Cites academic references for source algorithms

PR #9175, opened 2017-04-12
Nadam optimizer

jitindua:
It has the implementation of Nadam optimizer by using a use_nesterov flag in Adam optimizer (as done in Momentum class)

Issue link: https://github.com/tensorflow/tensorflow/issues/7715

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,9 @@
+# Release 1.0.1
+
+## Bug Fixes and Other Changes
+* Change GraphConstructor to not increase the version when importing, but instead take the min of all versions.
+* Google Cloud Storage fixes
+
 # Release 1.0.0
 
 ## Major Features and Improvements
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 
 ## Bug Fixes and Other Changes
 * Change GraphConstructor to not increase the version when importing, but instead take the min of all versions.
-* Google Cloud Storage fixes
+* Google Cloud Storage fixes.
 * Removed `tf.core` and `tf.python` modules from the API. These were never intended to be exposed. Please use the same objects through top-level `tf` module instead.
 
 # Release 1.0.0
@@ -94,7 +94,7 @@ To help you upgrade your existing TensorFlow Python code to match the API change
 * Change arg order for `{softmax,sparse_softmax,sigmoid}_cross_entropy_with_logits` to be (labels, predictions), and force use of named args.
 * tf.nn.rnn_cell.* and most functions in tf.nn.rnn.* (with the exception of dynamic_rnn and raw_rnn) are temporarily in tf.contrib.rnn.  They will be moved back into core for TF 1.1.
 * `tf.nn.sampled_softmax_loss` and `tf.nn.nce_loss` have both changed their API such that you need to switch the `inputs, labels` to `labels, inputs` parameters.
-* The shape keyword argument of the SparseTensor constructor changes its name to dense_shape between Tensorflow 0.12 and Tensorflow 1.0.
+* The shape keyword argument of the `SparseTensor` constructor changes its name to `dense_shape` between Tensorflow 0.12 and Tensorflow 1.0.
 
 ## Bug Fixes and Other Changes
 * Numerous C++ API updates.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,6 +3,7 @@
 ## Bug Fixes and Other Changes
 * Change GraphConstructor to not increase the version when importing, but instead take the min of all versions.
 * Google Cloud Storage fixes
+* Removed `tf.core` and `tf.python` modules from the API. These were never intended to be exposed. Please use the same objects through top-level `tf` module instead.
 
 # Release 1.0.0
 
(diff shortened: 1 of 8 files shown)


drpngx:
Could you:
1. Sign the CLA
2. git pull rebase and push again

Take a look at the recent merge:
https://github.com/tensorflow/tensorflow/pull/8405

@lukaszkaiser it looks like this implementation might be more compact and easier to maintain? Rather than splitting the class into two, it modifies the current class.

lukaszkaiser:
I fully agree. But the other CL also had RAdam. I'm not sure how to do it best to assure these are easy to maintain. This one should also have the advantage of working with LazyAdam.

louiehelm:
@lukaszkaiser @drpngx I agree @jitindua's code looks more compact and has testing and maintenance advantages.

One of my calculations may be more efficient in the CPU version of Nadam because of how I'm using Eigen. Perhaps @jitindua could try to adapt this efficiency if you're gonna go with his PR? See code near https://github.com/tensorflow/tensorflow/pull/8405/files#diff-fee33779233fb6ff133bbb93c7830880R301

One of the subtle / critical issues I was fixing with my Nadam interface (which this can't fix) was setting saner default decay factors.

Kingma & Ba's paper on Adam only sets beta1=0.9 and beta2=0.999 to promote their pet initialization scheme (that no one uses). And if you don't use it, those factors are absurdly large + bad. I know in theory that developers could set good decay factors on their own. But in practice, 100 out of 100 developers incorrectly compensate for Adam's poorly set default decay factors by setting depressingly low learning rates to prevent it from diverging instantly.

In some sense, Radam itself is a research "discovery" about how wrong the most widely-used learning rate / decay factors in Adam are. Its gamma factor is just a more convenient knob to help change beta1 and beta2 in the correct proportions so you can properly decrease them in tandem.

How about you pull this PR, and I'll submit two thin, python-only, init-only front-ends into core (like my current Radam one) so that there's an explicit Radam call and a Nadam optimizer that doesn't have unconscionably bad default decay factors?

lukaszkaiser:
@louiehelm Just to give you some context: I was fighting for different defaults in TF Adam since before the first release of TF. Unluckily, people protest when things become incomparable to other frameworks and papers, so here we are with bad defaults (and we probably have to stick to them in the core).

How about this: we merge @jitindua's code first (please, take a good look at it too and comment and help to make it efficient) -- and this goes into core (it's just a slight change of core API, I hope Martin will be ok with that). Then, using his changes, we merge your NAdam and RAdam into contrib/opt, also ensuring they handle sparse updates fast, same as LazyAdam which is also in contrib/opt (as you saw, Martin is not ok to have them in core API yet). Later, when everyone starts using them instead of the core Adam, we'll migrate them into core. Does that sound ok?

ahundt:
@lukaszkaiser sorry to hijack the conversation a bit, but where can I find out about better defaults?

lukaszkaiser:
I guess @louiehelm can tell you more. I was using large epsilon (1e-4) to stabilize for a while, then I lowered beta1 to 0.85 and beta2 to 0.997 and brought epsilon back down to 1e-7. But I doubt this is the best. @louiehelm what do you use with pure Adam?

ahundt:
Cool thanks, also what NADAM parameters do you recommend? Perhaps that could be included in the docstring if they won't be defaults?

I know performance will vary by application but maybe there is a good place to provide reference values to reduce the searching (via hyperopt or google) required.

louiehelm:
I've had good performance with beta1 = 0.825, beta2 = 0.99685, lr = double.

If you have a well-tuned model using Adam's stock decay params, this beta2 lets you simply double the learning rate, while still landing back in the same favorable gradient landscape you were navigating into before (but with more stability).

I didn't do the explicit math, so there might be a more precise set of beta1 / beta2 that give this benefit. Given how many people are getting comfortable with well-tuned models that hum along with Adam, maybe this is a desirable target for future default decay factors, so that we can easily migrate researchers / developers to a new standard. At very least, figuring out decay factors that do this and then mentioning them in Adam's documentation like @ahundt suggests seems wise.

I share @lukaszkaiser's intuition that epsilon should be higher than 1e-8 but never benefited from raising it myself in practice. However, that's a parameter you wouldn't see harming your results until it bites you... so I maybe just got lucky. I haven't done the proper data interrogation to see the warning signs if 1e-8 is concerningly low or not.

drpngx:
@jitindua how much bandwidth do you have for this? Could you sign the CLA?

lukaszkaiser:
I only have problems with epsilon when training on tasks where the accuracy gets very high, and gradients very low, sometimes close to 0. Some tasks are like this, you want to get to 99% accuracy. But then with low epsilon Adam makes this crazy large step somewhere and gets bad. I think that a good way would be instead of having `(sqrt(v_t) + epsilon)` to have something smarter, like `max(sqrt(v_t), epsilon)` but maybe a little smoothed. So one could set epsilon higher without effecting all the normal gradients. Not sure if it's possible though, but I'd love not to have to care about that...

martinwicke:
@lukaszkaiser @jitindua What's the verdict here?

lukaszkaiser:
Again, here is my suggestion:

How about this: we merge @jitindua's code first (please, take a good look at it too and comment and help to make it efficient) -- and this goes into core (it's just a slight change of core API, I hope Martin will be ok with that). Then, using his changes, we merge your NAdam and RAdam into contrib/opt, also ensuring they handle sparse updates fast, same as LazyAdam which is also in contrib/opt (as you saw, Martin is not ok to have them in core API yet). Later, when everyone starts using them instead of the core Adam, we'll migrate them into core.

Do we all agree on that general plan?

louiehelm:
@jitindua 's code is cool and all. But let's not forget that my PR still has a signed CLA, is properly based, linted, and ready to merge.

Remember those 12 hours or so it was live? Good times!

@martinwicke -- why don't we just add Nadam and Radam to golden and go back to those good dope days?

I was all set to merge down the repeated C++ for another PR when my Nadam code got unpulled  💔
https://github.com/tensorflow/tensorflow/pull/8405#issuecomment-293402921

Still game to clean up TF's C++ optimizers or even migrate some of their interfaces towards Python eventually. Can I have Nadam and Radam in core in exchange for properly refactoring all the Optimizer interfaces?

drpngx:
Sorry @jitindua but if you don't sign the CLA there's nothing we can do. Also, if you don't have cycles to respond then I'm not sure we can really help.

@louiehelm we have made the decision that the code should live in contrib first, as we usually do. Contrib is our canary. Code that lives there has time mature with a lot of freedom.
As long as the API doesn't change, we can refactor code and make it cleaner.
As I understand, this PR has the advantage that there is less replicated code. If you think you can resubmit your PR, in contrib, with more shared code, then I think that might be a good avenue and we can drop this PR.

vrv:
It looks like @jitindua has not responded since the original PR, and it's been 3 weeks.  So I'm going to close this PR and hope that it gets opened by someone (anyone!) who can sign the CLA, preferably under 'contrib', and with the suggestions / improvements posted here!

PR #9181, opened 2017-04-13
fix filter position for Im2Col

phomes:
cac52fdebb292bc7706383d07c7987be66882085 fixed the filter position for
the reference class but the same problem exists in the Im2Col class.

This patch makes the calculation of the filter position consistent
with the reference quantized class and the gemm/fused conv_ops
classes.

@petewarden you reviewed the reference class fix. Would you have the time to review this too?


drpngx:
Jenkins, test this please.

caisq:
@petewarden Can you give an official approval or comments for this PR? Thanks.

petewarden (approved):

Issue #9188, opened 2017-04-13
Improving tf.cond


gaohuazuo:
Currently `tf.cond` requires a function for each branch, which makes no difference semantically but serves as an optimization hint. Such requirement becomes quite inconvenient in some cases, for example,

```
def build_graph(p1, p2):
    a = a_fn() # Redundant when p1 and p2 are both false.
    b1 = tf.cond(p1, lambda: a, b1_fn)
    b2 = tf.cond(p2, lambda: a, b2_fn)
    return b1, b2

# An alternative
def build_graph(p1, p2):
    # Redundant computation when p1 and p2 are both true.
    b1 = tf.cond(p1, a_fn, b1_fn)
    b2 = tf.cond(p2, a_fn, b2_fn)
    return b1, b2
```

Though possible, it is not so straightforward to transform the above graph into an efficient one. In fact, we don't really need the optimization hint. Once target nodes are known, the exact execution condition for each node can be readily calculated. Furthermore, these conditions can be implemented using existing `Switch` and `Merge` ops. I have made a prototype to illustrate the idea at https://github.com/tensorflow/tensorflow/pull/9189.

itsmeolivia:
Automatically closing due to lack of recent activity. Since this issue is old at this point, please reopen the issue if it still occurs when tried with the latest version of Tensorflow. Thank you.

PR #9189, opened 2017-04-13
A replacement for tf.cond

gaohuazuo:
This is a prototype for https://github.com/tensorflow/tensorflow/issues/9188.

Two new ops, `Mux` and `Demux` are introduced by this patch. `Mux` op takes `[i, input_0, input_1, ...]` as input, and produces `input_i` as output. It is meant to replace `tf.case` and `tf.cond`. `Demux` op is a generalized version of `Switch`. It takes `[i, input]` as input, and produces `[output_0, ..., output_N]` where `output_i` is set to `input` and other outputs are dead.

This patch also contains an algorithm that rewrites the graph before execution so that `Mux` ops are efficient just like `tf.cond`.

Example:

```
i = tf.placeholder(tf.int32, [])
a = tf.placeholder(tf.int32, [])
b = tf.placeholder(tf.int32, [])
c = tf.select(i, [a+1, b*2]) # Mux op is renamed to select in python api.

sess.run(c, feed_dict={i: 0, a: 1, b: 2}) # returns 2, trace to see that "Mul" node is not executed.
sess.run(c, feed_dict={i: 1, a: 1, b: 2}) # returns 4, trace to see that "Add" node is not executed.
```

I wouldn't be surprised if this patch causes problem somewhere. In particular, the graph rewriting algorithm may produce valid but cyclic graph. It also lacks proper handling of special nodes like `Enter` `NextIteration` etc. However, I would like to know if this approach is promising. I can work on this if there is no fundamental difficulty.


drpngx:
@sherrym @yuanbyu any thoughts on this?

@gaohuazuo we should probably try this in contrib first.

danijar:
This seems really cool, I'd be happy to see something like this in TensorFlow.

drpngx:
Jenkins, test this please.

yuanbyu:
The approach looks interesting. One area that you probably need to pay some attention to is stateful ops. One good test is to replace all the conds in third_party/tensorflow/python/kernel_tests/control_flow_ops_py_test.py with yours and make sure all the tests still pass.

vrv:
@gaohuazuo friendly ping!  This is indeed a very interesting change that we'd love to see more work on!

vrv:
(My suggestion would be to leave the C++ op in core, but have the 'select' API in contrib)

gaohuazuo:
I'm working on it. I will push to github once I have checked all tests in `control_flow_ops_py_test`.

Progress:
The graph rewriting algorithm now play nice with while loops. However, optimization does not cross loop or iteration boundary.

TODOs:
Check all tests in `control_flow_ops_py_test`.
Implement gradient.
Python api improvements (move to contrib, add a drop-in replacement for `tf.cond`, docs).

gaohuazuo:
`control_flow_ops_py_test` result:

`bazel test -c dbg --config cuda //tensorflow/python/kernel_tests:control_flow_ops_py_test`

* testCondGrad_1
    `LookupError: No gradient defined for operation 'select/Mux' (op type: Mux)`

* testCondGrad_2
    `LookupError: No gradient defined for operation 'select/Mux' (op type: Mux)`

* testCondGrad_Gather
    `LookupError: No gradient defined for operation 'select/Mux' (op type: Mux)`

* testCondIndexedSlices
    `ValueError: Tensor conversion requested for IndexedSlices without dense_shape: IndexedSlices(indices=Tensor("Const_1:0", shape=(), dtype=int32, device=/device:CPU:0), values=Tensor("Add:0", shape=(), dtype=int32, device=/device:CPU:0))`

* testCondIndexedSlicesDifferentTypes
    `ValueError: Tensor conversion requested for IndexedSlices without dense_shape: IndexedSlices(indices=Tensor("one:0", shape=(), dtype=int32, device=/device:CPU:0), values=Tensor("Add:0", shape=(), dtype=int32, device=/device:CPU:0))`

* testCondRef
    `ValueError: Input tensors must have the same dtype.`

* testCondSparseTensor
    `TypeError: Expected binary or unicode string, got <tensorflow.python.framework.sparse_tensor.SparseTensor object at 0x7fa52815a898>`

* testCondWhile_3
    `LookupError: No gradient defined for operation 'select/Mux' (op type: Mux)`

* testNestedCond_Simple
    `LookupError: No gradient defined for operation 'select_1/Mux' (op type: Mux)`

* testNestedWhileCondWhileGrad
    `LookupError: No gradient defined for operation 'while/select/Mux' (op type: Mux)`

* testWhileCondGrad_Simple
    `LookupError: No gradient defined for operation 'while/select/Mux' (op type: Mux)`

* testWhileCondGrad_UnknownShape
    `LookupError: No gradient defined for operation 'while/select/Mux' (op type: Mux)`

* testWhileGrad_Square
    `LookupError: No gradient defined for operation 'select/Mux' (op type: Mux)`

* testCond_4
    ```
    AssertionError: Tuples differ: () != (0,)

    Second tuple contains 1 additional elements.
    First extra element 0:
    0

    - ()
    + (0,) : Shape mismatch: expected (), got (0,).
    ```

### Summary

* `tf.select` and `tf.cond_v2` does not handle `IndexedSlice` and `SparseTensor` properly.
    Accounts for `testCondIndexedSlices`,`testCondIndexedSlicesDifferentTypes` and `testCondSparseTensor`.
   I am not very familiar with `IndexedSlice` and `SparseTensor`, so I need to study further to provide a fix.

* `tf.select` does not handle ref dtypes properly.
   Accounts for `testCondRef`.
   Possible fix would be adding a `RefMux` op and some handling in python `select` function.

* When an input is an `tf.Operation`, `tf.cond` produces a scalar bool tensor whereas `tf.cond_v2` produces an empty float tensor. This is just a minor change and any sane application should not rely on it.
    Accounts for `testCond_4`.

* `Mux` op has no gradient.
    Accounts for other failed tests.
   Possible fix would be adding a gradient function or modifying `tf.gradients`. Simply registering a gradient function should work, but is less efficient.

### Some questions

* Some op kernels, `Merge` for example, require that all int32 inputs reside in host memory. Why? Should I do the same?

* Which contrib module should I add `select` and `cond` to? Or should I create a new contrib module?

danijar:
@gaohuazuo I can't answer your questions, but thanks a lot for putting in all this effort!

gaohuazuo:
@yuanbyu  @vrv `cond_v2` has passed almost all tests in `ControlFlowTest` (see previous comment and https://github.com/tensorflow/tensorflow/pull/9189/commits/2d882220c3ccaf161e9d243a382121b030b762cf). I will complete the docs and code comments soon. What else has to be done before I may ask for review?

Notes on gradient:
The current naive approach is to register a gradient function for `Mux`, as if it was an ordinary op. It works, but the optimization that is enjoyed by `tf.cond` is missing. Even so, it already provides more efficiency in some scenarios, for example `y = do_something(tf.cond(tf.constant(False), lambda: x1, lambda: x2)); tf.gradients(y, x1)[0].eval()`.

rmlarsen:
@yuanbyu could you provide some guidance for the author here?

cwhipkey:
Regarding questions from before:

1. "What else has to be done before I may ask for review?"

I think as long as the functionality is complete enough, with the tests, then we can start a review of the code.  Some things may come up during that review.

2. "Additionally, which contrib module should I use?"

Does tf.contrib.control_flow seem okay?

3. "Some op kernels, Merge for example, require that all int32 inputs reside in host memory. Why? Should I do the same?"

Some background: Some operations are commonly used for operations on shape tensors, which tend to be int32 (although int64 is possible).  If those operations have int32 GPU kernels registered which put the memory on device, then placement of ops to devices will put them on GPU, but it's better to avoid the bouncing back-and-forth from device-to-host for those operations.  Currently, this is avoided by having the int32 GPU kernels request HostMemory for the inputs and outputs; iiuc, this mainly works because int32 is not used often for non-shape tensors on GPU.  So I would guess that your ops should match their switch/merge analogs with this behavior.

ahundt:
Looks interesting!

> Mux op is renamed to select in python api.

I like `select`, it is a much better name! Could select or an incremented equivalent be considered for the C++ side as well?

cwhipkey:
@gaohuazuo - Please let me know when I should review the code (I can review now, or wait for soem more moves and updates)

gaohuazuo:
@cwhipkey Thanks for the reply! I am planning to complete the comments and clean up the code. I don't think this PR in its current state is ready for merge. However, I am on a full time intern now, so it is going to take some time. I will update this thread when it is done.

gaohuazuo:
@ahundt There is a `RefSelect` op in C++. Although `RefSelect` do the same thing, I think it would be safer to use another name.

ahundt:
> @ahundt There is a RefSelect op in C++. Although RefSelect do the same thing, I think it would be safer to use another name.

Can the existing RefSelect be extended to work in both cases in a clear and concise manner that also has backwards compatibility?

Consistent and clear naming is one of the most important things for users to make sense of and use the API, so please consider giving some additional thought on what naming approach will be clear between python, and C++, and RefSelect. :-)

drpngx:
@gaohuazuo could you resolve conflicts and respond to comments?

Thanks

vrv:
Any updates?

gaohuazuo:
@vrv I think I won't have the spare time until October.

vrv:
:( I'm sorry to hear that but perhaps someone else will pick this up sooner, or it will be waiting for you otherwise.  Thanks for your effort, and I hope you learned something in the process too :)

yaroslavvb:
I'm curious as to the motivation for this change. Referenced issue talks about `tf.cond` awkwardness -- things need to be wrapped into a function. But this awkwardness can be fixed on the client level using existing Merge/Switch primitives. IE, a smarter `tf.cond` replacement could instead take existing TensorFlow tensors and ensure laziness by inserting Merge/Switch nodes into the graph like [here](https://gist.github.com/yaroslavvb/d67410e240369736fc4ba0267250ef27)

gaohuazuo:
@yaroslavvb I feel that the current `tf.cond` adds an unnecessary mechanism to the computational graph formulation. This pr tries to make `tf.cond` an ordinary op.

yaroslavvb:
Ah so it's for simplification, makes sense. Yeah, the amount of Python logic in existing tf.cond/tf.case ops is a bit scary

PR #9193, opened 2017-04-13
Support int32 idxs in sparse_tensor_dense_matmul

drasmuss:
The `sparse_tensor_dense_mat_mul` operator requires indices to be stored in 64bit precision, which can end up consuming a lot of memory for large arrays.  This adds support for 32bit indices.

Note that I only changed `gen_sparse_ops._sparse_tensor_dense_mat_mul`, not `sparse_ops.sparse_tensor_dense_matmul`.  This makes this change fairly invisible, only accessible to those who know to look for it.  Changing the public facing function would require changing `SparseTensor` to accept 32bit indices, which might be a good thing to do but would be a much larger change.

### DIFF
--- a/tensorflow/core/kernels/sparse_tensor_dense_matmul_op.cc
+++ b/tensorflow/core/kernels/sparse_tensor_dense_matmul_op.cc
@@ -29,7 +29,7 @@ namespace tensorflow {
 typedef Eigen::ThreadPoolDevice CPUDevice;
 typedef Eigen::GpuDevice GPUDevice;
 
-template <typename Device, typename T>
+template <typename Device, typename T, typename Tindices>
 class SparseTensorDenseMatMulOp : public OpKernel {
  public:
   explicit SparseTensorDenseMatMulOp(OpKernelConstruction* ctx)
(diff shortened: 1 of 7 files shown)


drpngx (dismissed):
Nice!

drpngx:
Jenkins, test this please.

drasmuss:
Pushed an update that I think should fix that backwards compatibility error.

drpngx:
Wait what broken the backwards compat?

Jenkins, test this please.

drasmuss:
It was because the new attribute I added (Tindices) didn't have a default value (this line https://github.com/tensorflow/tensorflow/pull/9193/commits/7adcd0d2dba35e7802591a64e8d025e19f951668#diff-ac07c69dd7c613a69fab85a90cae89f7R137)

drpngx:
SG.

@vrv checking if there are any concerns with this change.

vrv:
seems reasonable to me since it's backwards compatible and doesn't create the potential mess of SparseTensor casting :)

drpngx (approved):

PR #9195, opened 2017-04-13
Enable protobuf's --with-pic option

FloopCZ:
This PR only applies to **contrib/makefile** project.

`libtensorflow-core.a` static library is built with position independent code (i.e., -fPIC option), however, `libprotobuf.a` static library is not. This MR enables the protobuf's `--with-pic` option in order to be able to use `libprotobuf.a` when building a shared library.


drpngx (approved):

drpngx:
Jenkins, test this please.

PR #9211, opened 2017-04-14
Enabled LLVM to be built on Linux ppc64le when XLA is enable…

npanpaliya:
Added ppc64le's support while building LLVM when it is enabled through configure option.


drpngx (approved):

drpngx:
Jenkins, test this please.

drpngx:
Ignoring mac failure from XLA graph.

PR #9214, opened 2017-04-14
Fixed jemalloc build on linux ppc64le when it is enabled while configure

npanpaliya:


tensorflow/core/platform/default/build_config.bzl
```diff
@@ -194,12 +194,14 @@ def tf_kernel_tests_linkstatic():
 def tf_additional_lib_defines():
   return select({
       "//tensorflow:with_jemalloc": ["TENSORFLOW_USE_JEMALLOC"],
+      "//tensorflow:linux_ppc64le":["TENSORFLOW_USE_JEMALLOC"],
```
npanpaliya: Here, we know that this change would always enable jemalloc to build irrespective of whether it is configured. If we don't do this, then code in tensorflow/BUILD that enables jemalloc only for k8 cpu will never allow jemalloc to be built on non-x86 arch. Kindly suggest appropriate way to handle ppc64le case too.

third_party/jemalloc.BUILD
```diff
@@ -89,6 +89,13 @@ cc_library(
         "-D_REENTRANT",
     ],
     includes = ["include"],
+    linkopts = select({
+        "@%ws%//tensorflow:linux_ppc64le": [
+            "-lpthread",
+        ],
+        "//conditions:default": [
+        ],
+    }),
```
npanpaliya: This change is needed to fix bazel tests in tensorflow/compiler when jemalloc is enabled. Jemalloc code uses pthread_atfork function in jemalloc.c which gets executed in case Power but not on x86. To resolve this function, we need jemalloc to be built linking libpthread.

third_party/jemalloc.BUILD
```diff
@@ -183,12 +190,12 @@ sh_binary(
...
 )
 
-# Size classes for Linux x86_64. Update if adding builds for other
+# Size classes for Linux x86_64 and ppc64le. Update if adding builds for other
 # architectures. See size_classes.sh for details on the arguments.
 genrule(
     name = "size_classes_h",
     outs = ["include/jemalloc/internal/size_classes.h"],
-    cmd = "$(location :size_classes_sh) \"3 4\" 3 12 2 >$@",
+    cmd = "if [ $$(uname -p) = ppc64le ]; then $(location :size_classes_sh) \"3 4\" 3 16 2 >$@; else $(location :size_classes_sh) \"3 4\" 3 12 2 >$@; fi",
```
npanpaliya: I tried avoiding uname here as it won't work on windows. Tried to use $(TARGET_CPU) which gives correct value on command line "ppc" but not when this file is executed. That time, it takes "local" as the cpu value defined in crosstool.tpl file. I'll update this code if anyone has any better suggestions other than using uname to detect architecture.

drpngx:
@jhseu @jart any suggestion?

tensorflow/BUILD
```diff
@@ -89,6 +89,12 @@ config_setting(
 )
 
 config_setting(
+    name = "linux_ppc64le",
```
jhseu: Can you make this into a separate rule that also checks for jemalloc?
"with_jemalloc_ppc64le"

jart:
This PR can not be merged as is. @npanpaliya are these select() expression changes meant to enable jemalloc when `n` is answered to the `./configure` prompt asking if jemalloc should be enabled?

hartb:
@jart @jhseu I'm another member of @npanpaliya 's team.  We don't want to force jemalloc to always build on ppc, but we couldn't think of a good way to work around the fact that the current with_jemalloc config_setting in tensorflow/BUILD is limited to k8.

Looking at the suggestions in the thread now...  Would it be a good solution to:

- modify existing K8-specific config_setting rule to be "with_jemalloc_k8"
- add new ppc64le-specific config_setting rule "with_jemalloc_ppc64le"
- update the selects in build_config.bzl / tf_additional_lib_defines and tf_additional_lib_deps to recognize both with_jemalloc_k8 and with_jemalloc_ppc64le

jart:
I see what you're saying. In that case could we say:

```py
config_setting(
    name = "with_jemalloc",
    values = {"define": "with_jemalloc=true"},
)
```

And then:

```py
select({
    "//tensorflow:with_jemalloc": select({
        "//tensorflow:linux_x86_64": [...],
        "//tensorflow:linux_ppc64le": [...],
        "//conditions:default": [],
```

jhseu:
@hartb Yeah, that works.

@jart Nesting selects doesn't work in Bazel, unfortunately:
https://github.com/bazelbuild/bazel/issues/1623

jart:
In that case I'd recommend one of the following:

1. Keep current approach and rename `with_jemalloc` to `with_jemalloc_on_linux_x86_64` and rename `linux_ppc64le` to `with_jemalloc_on_linux_ppc64le`.

2. Investigate if jemalloc can just be enabled for all platforms if `with_jemalloc` is `true`. When I coded it this way IIRC I was erring on the side of caution and inheriting previous decisions. I haven't investigated what platforms jemalloc does and does not support.

I would prefer (2) but it's probably more work.

jhseu:
#2 is a decent amount of work, so let's stick with #1

hartb:
@jart @jhseu Thank you both!  @npanpaliya will re-work the PR.

We'll go with approach 1; we just don't have access to most of the other platforms to test.

Do you have any suggestions for the change in third_party/jemalloc.BUILD?  There we want to substitute in a page size indicator depending on the architecture (and really, OS and arch).  Is there a way we could set an intermediate variable based on linux_ppc64le / otherwise, and the substitute that into 'cmd'?

jart:
We can add select() expressions inside jemalloc.BUILD using the "@%ws%//tensorflow:FOO" config_setting label hack.

hartb:
@jart Thank you; we'll take a look at that. 

I also see some cases of select() used for genrule / cmd in third_party/jpeg/jpeg.BUILD.  Maybe we can do something like that.

npanpaliya:
Thank you all for your suggestions. I've implemented the suggested approach. Kindly review and let me know your feedback.

drpngx:
Jenkins, test this please.

jhseu (approved):

jhseu:
Jenkins, test this please

PR #9219, opened 2017-04-14
R0.11

jitendramaan:
need to build tensorflow model


martinwicke:
This looks like a mistake to me.

PR #9229, opened 2017-04-15
An ugly hack to compile on Tegra X1 /w Jetpack 2.3.1 release.

wang007987:
* Applied patch from https://github.com/jetsonhacks/installTensorFlowTX1
* Revert Eigen to revision used in Tensorflow r0.11 to avoid cuda compile error
* Remove expm1 op that was added with new additions to Eigen


caisq:
@benoitsteiner do you have time to take a look at this PR?

martinwicke:
We certainly don't want to accept ugly hacks. I'll close this PR. I think we would welcome a principled solution to making this work on Tegra X1. @petewarden FYI.

petewarden:
Sorry we couldn't take this one! I am very keen to see better TX1 support, would you be able to at least write up a quick doc about the process you went through, and maybe link to a fork containing this patch?

PR #9236, opened 2017-04-15
add "frames" op to "contrib/signal"

Androbin:
Added op to frame a signal into overlapping frames.
Inspired by [`python_speech_features.sigproc.framesig`](https://github.com/jameslyons/python_speech_features/blob/master/python_speech_features/sigproc.py#L13)

May be used in front of [spectral functions](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/api_guides/python/spectral_ops.md).

```
pcm = tf.placeholder(tf.float32, [None, 9152])
frames = tf.contrib.signal.frames(pcm, 512, 180)
magspec = tf.abs(tf.spectral.rfft(frames, [512]))
image = tf.expand_dims(magspec, 3)
```


Androbin:
I am unsure about the difference between `tf.parallel_stack` and `tf.stack`, so I added the flag `parallel` defaulting to `False`

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1275,51 @@ def transpose(a, perm=None, name="transpose"):
...
+    frame_step: The step between frames.
+    parallel: Whether to use `tf.parallel_stack` or `tf.stack`
+    name: A name for the operation (optional).
+  
+  Returns:
+    A `Tensor` of frames with shape [batch_size, num_frames, frame_len].
+  """
+  sig_len = int(sig.get_shape()[1])
+  
+  num_frames = 1 + int(math.ceil((1.*sig_len-frame_len)/frame_step))
```
caisq: You need spaces between the tokens. See https://google.github.io/styleguide/pyguide.html

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1275,51 @@ def transpose(a, perm=None, name="transpose"):
     return ret
 
 
+def framesig(sig, frame_len, frame_step, parallel=False, name="framesig"):
```
caisq: Please add unit tests.

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1275,51 @@ def transpose(a, perm=None, name="transpose"):
...
+  """Frame a signal into overlapping frames.
+  May be used in front of spectral functions.
+  
+  For example:
+  
+  ```python
+  pcm = tf.placeholder(tf.float32, [None, 9152])
```
caisq: Add error checking for cases in which the input tensor is not 2D?

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1275,51 @@ def transpose(a, perm=None, name="transpose"):
...
+  
+  with ops.name_scope(name, "framesig", [sig]) as name:
+    if pad_len != sig_len:
+      pad_sig = ops.pad(sig, [[0, 0], [0, pad_len-sig_len]], name=name)
+    else:
+      pad_sig = sig
+    
+    frames = []
+    
+    for i in range(num_frames):
```
caisq: Better to use xrange (six.moves.xrange for Python 2-3 compatibility).

caisq (changes requested):

tensorflow/python/ops/array_ops.py
Androbin: Sure, where should they go?

tensorflow/python/ops/array_ops.py
Androbin: Sure, but how could higher ranks be supported?

tensorflow/python/ops/array_ops.py
Androbin: Sure, sorry for the hurry.

tensorflow/python/ops/array_ops.py
Androbin: Sure, but why isn't xrange used in all the other ops?

tensorflow/python/ops/array_ops.py
```diff
@@ -49,6 +49,7 @@
 @@reverse
 @@reverse_v2
 @@transpose
+@@framesig
```
ebrevdo: this should not go into TF core, but into contrib/spectral/

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1277,57 @@ def transpose(a, perm=None, name="transpose"):
     return ret
 
 
+def framesig(sig, frame_len, frame_step, parallel=False, name="framesig"):
```
ebrevdo: let's not allow parallel; at least to start with. h i believe parallel_stack does not support backprop.

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1277,57 @@ def transpose(a, perm=None, name="transpose"):
...
+  ```python
+  pcm = tf.placeholder(tf.float32, [None, 9152])
+  frames = tf.framesig(pcm, 512, 180)
+  magspec = tf.abs(tf.spectral.rfft(frames, tf.constant(512, shape=[1])))
+  image = tf.reshape(magspec, [-1, 49, 257, 1])
+  ```
+  
+  Args:
+    sig: A `Tensor` with `rank = 2`.
+    frame_len: The length of each frame.
```
ebrevdo: can these be tensors?  if so, what dtype and shapes?  does this broadcast with sig?

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1277,57 @@ def transpose(a, perm=None, name="transpose"):
...
+  
+  ```python
+  pcm = tf.placeholder(tf.float32, [None, 9152])
+  frames = tf.framesig(pcm, 512, 180)
+  magspec = tf.abs(tf.spectral.rfft(frames, tf.constant(512, shape=[1])))
+  image = tf.reshape(magspec, [-1, 49, 257, 1])
+  ```
+  
+  Args:
+    sig: A `Tensor` with `rank = 2`.
```
ebrevdo: signal, not sig.  whats dtypes and what are the names of the axes?

tensorflow/python/ops/array_ops.py
ebrevdo: frame_length

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1277,57 @@ def transpose(a, perm=None, name="transpose"):
...
+    frame_len: The length of each frame.
+    frame_step: The step between frames.
+    parallel: Whether to use `tf.parallel_stack` or `tf.stack`
+    name: A name for the operation (optional).
+  
+  Returns:
+    A `Tensor` of frames with shape [batch_size, num_frames, frame_len].
+  """
+  sig_shape = sig.get_shape()
```
ebrevdo: this may be TensorShape(None) - in which case, neither the rank nor the batch size, nor the length are known.  in that case you have to do all your calculations using tensorflow math_ops.

you could require that sig_shape.ndims == 2, which is fine; but len(sig_shape) will not work.

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1277,57 @@ def transpose(a, perm=None, name="transpose"):
...
+  sig_rank = len(sig_shape)
+  
+  if sig_rank != 2:
+    raise ValueError("expected sig to have rank 2 but was " + sig_rank)
+  
+  sig_len = int(sig_shape[1])
```
ebrevdo: this will fail if sig_shape[1].value is None.  i suggest just doing all the math using TF ops.

tensorflow/python/ops/array_ops.py
```diff
@@ -1273,6 +1277,57 @@ def transpose(a, perm=None, name="transpose"):
...
+  
+  with ops.name_scope(name, "framesig", [sig]) as name:
+    if pad_len != sig_len:
+      pad_sig = ops.pad(sig, [[0, 0], [0, pad_len - sig_len]], name=name)
+    else:
+      pad_sig = sig
+    
+    frames = []
+    
+    for i in xrange(num_frames):
```
ebrevdo: can you do this without a python for loop?  again, this requires the num_frames be known at graph build time which you should not assume.

ebrevdo (changes requested):

tensorflow/python/ops/array_ops.py
Androbin: I understand your concerns. Unfortunately I don't know how to do this dynamically...

tensorflow/python/ops/array_ops.py
Androbin: Alright, but I am unsure about the directory hierarchy and the need for additional files.

tensorflow/python/ops/array_ops.py
Androbin: Okay, but why wouldn't it?

tensorflow/python/ops/array_ops.py
Androbin: I have rewritten it to use Tensor objects, but the loop breaks it all.

tensorflow/contrib/spectral/python/ops/shape_ops.py
```diff
@@ -0,0 +1,52 @@
...
+  signal_length = int(signal_shape[1])
+  
+  with ops.name_scope(name, "frames", [signal]) as name:
+    num_frames = ops.add(1, ops.cast(ops.ceil(ops.divide(ops.subtract(signal_length, frame_length), frame_step)), dtypes.int32))
+    pad_length = ops.add(ops.multiply(ops.subtract(num_frames, 1), frame_step), frame_length)
+    pad_signal = ops.pad(signal, [[0, 0], [0, ops.subtract(pad_length, signal_length)]])
+    
+    frames = []
+    
+    for index in xrange(49): # TODO(Androbin) use num_frames dynamically
```
ebrevdo: ok so this is a really good question.  how can you vectorize copying out slices into one operation?
perhaps this can be done with a array_ops.gather or a 2d convolution?  instead of writing it as a for loop?  the way that'll work but is the slowest is probably to use functional_ops.scan possibly followed by a reshape.  only do it that way if you can't find a more efficient way via convolution / sparse matmul / gather / gather_nd.

also, it looks like you're explicitly padding the signal at the end with zeros.  is this how the scipy version does it?

tensorflow/contrib/spectral/python/ops/shape_ops.py
```diff
@@ -0,0 +1,52 @@
...
+      frames.append(ops.slice(pad_signal, [0, ops.mutiply(index, frame_step)], [-1, frame_length]))
+    
+    # TODO(Androbin) consider using tf.gather(...)
+    return ops.stack(frames, axis=1, name=name)
```
ebrevdo: you'll probably want to remove this once you write it using dynamic sized primitives.

tensorflow/contrib/spectral/python/ops/shape_ops.py
```diff
@@ -0,0 +1,52 @@
+import math
```
ebrevdo: add the copyright header here

see other contrib/ packages for examples of boilerplate you need:

1) a BUILD file
2) an __init__.py file
3) at least one file in python/ops/ (this one)
4) at least one file in python/kernel_tests/ (e.g. shape_ops_test.py) containing unit tests.

ebrevdo (changes requested):

tensorflow/contrib/spectral/python/ops/shape_ops.py
Androbin: Unfortunately, the other packages are not that consistent, but I will try my best!

tensorflow/contrib/spectral/python/ops/shape_ops.py
```diff
@@ -0,0 +1,52 @@
...
+  
+  signal_rank = signal.shape.ndims
+  
+  if signal_rank != 2:
+    raise ValueError("expected signal to have rank 2 but was " + signal_rank)
+  
+  signal_length = int(signal_shape[1])
+  
+  with ops.name_scope(name, "frames", [signal]) as name:
+    num_frames = ops.add(1, ops.cast(ops.ceil(ops.divide(ops.subtract(signal_length, frame_length), frame_step)), dtypes.int32))
```
Androbin: Thanks, I should have read myself into this!

tensorflow/contrib/spectral/python/ops/shape_ops.py
Androbin: Thanks for the elaboration! I have actually thought of some of that myself.
The code was inspired by [python_speech_features.sigproc.framesig](https://github.com/jameslyons/python_speech_features/blob/master/python_speech_features/sigproc.py#L13)
which does add padding via `numpy.concatenate`.
What do you mean by "explicitly"?
How would this even work if the signal was not of the right length?

ebrevdo: Their approach looks right. Transpose so the columns are axis 0, Create indices into the new rows the way they do, and use gather() to index into the transposed array.  then transpose the result.

Androbin: Alright, I have already thought about how to do it.
I will figure this out, commit and log any insights here.

Androbin: I think this might be due to how numpy handle broadcasting. Why doesn't TensorFlow support this for tf.tile just like np.tile?

Androbin: Okay, I think to have fixed it.
So, `frames ` should basically work now.
But how would one do `deframe` like `python_speech_features.sigproc.deframesig`?

Androbin:
It should work now using `tf.gather`, but I am unsure about how batches are handled.

Androbin:
The graph is constructed as expected.
But there is an issue with the initialization of the AdamOptimizer.

```
E tensorflow/core/common_runtime/executor.cc:644] Executor failed to create kernel. Internal: Dst tensor is not initialized.
	 [[Node: dens/w/Adam/Initializer/Const = Const[_class=["loc:@dens/w"], dtype=DT_FLOAT, value=Tensor<type: float shape: [131072,1024] values: [0 0 0]...>, _device="/job:localhost/replica:0/task:0/gpu:0"]()]]
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 1136, in _do_call
    return fn(*args)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 1118, in _run_fn
    status, run_metadata)
  File "/usr/lib/python3.5/contextlib.py", line 66, in __exit__
    next(self.gen)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/framework/errors_impl.py", line 466, in raise_exception_on_not_ok_status
    pywrap_tensorflow.TF_GetCode(status))
tensorflow.python.framework.errors_impl.InternalError: Dst tensor is not initialized.
	 [[Node: dens/w/Adam/Initializer/Const = Const[_class=["loc:@dens/w"], dtype=DT_FLOAT, value=Tensor<type: float shape: [131072,1024] values: [0 0 0]...>, _device="/job:localhost/replica:0/task:0/gpu:0"]()]]

During handling of the above exception, another exception occurred:

  File "conv_test.py", line 445, in <module>
    train(name, data_x, data_y, n_epoch, n_batch, load=load, save=save, log=log)
  File "conv_test.py", line 321, in train
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 786, in run
    run_metadata_ptr)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 994, in _run
    feed_dict_string, options, run_metadata)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 1129, in _do_run
    target_list, options, run_metadata)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 1149, in _do_call
    raise type(e)(node_def, op, message)
tensorflow.python.framework.errors_impl.InternalError: Dst tensor is not initialized.
	 [[Node: dens/w/Adam/Initializer/Const = Const[_class=["loc:@dens/w"], dtype=DT_FLOAT, value=Tensor<type: float shape: [131072,1024] values: [0 0 0]...>, _device="/job:localhost/replica:0/task:0/gpu:0"]()]]

Caused by op 'dens/w/Adam/Initializer/Const', defined at:
  File "conv_test.py", line 445, in <module>
    train(name, data_x, data_y, n_epoch, n_batch, load=load, save=save, log=log)
  File "conv_test.py", line 319, in train
    train_op = tf.train.AdamOptimizer().minimize(xent, name="train_op")
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/training/optimizer.py", line 325, in minimize
    name=name)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/training/optimizer.py", line 446, in apply_gradients
    self._create_slots([_get_variable_for(v) for v in var_list])
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/training/adam.py", line 128, in _create_slots
    self._zeros_slot(v, "m", self._name)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/training/optimizer.py", line 766, in _zeros_slot
    named_slots[_var_key(var)] = slot_creator.create_zeros_slot(var, op_name)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/training/slot_creator.py", line 174, in create_zeros_slot
    colocate_with_primary=colocate_with_primary)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/training/slot_creator.py", line 146, in create_slot_with_initializer
    dtype)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/training/slot_creator.py", line 66, in _create_slot_var
    validate_shape=validate_shape)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variable_scope.py", line 1054, in get_variable
    use_resource=use_resource, custom_getter=custom_getter)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variable_scope.py", line 951, in get_variable
    use_resource=use_resource, custom_getter=custom_getter)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variable_scope.py", line 356, in get_variable
    validate_shape=validate_shape, use_resource=use_resource)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variable_scope.py", line 341, in _true_getter
    use_resource=use_resource)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variable_scope.py", line 714, in _get_single_variable
    validate_shape=validate_shape)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variables.py", line 197, in __init__
    expected_shape=expected_shape)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variables.py", line 275, in _init_from_args
    initial_value(), name="initial_value", dtype=dtype)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/variable_scope.py", line 690, in <lambda>
    shape.as_list(), dtype=dtype, partition_info=partition_info)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/ops/init_ops.py", line 63, in __call__
    dtype=dtype, shape=shape)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/framework/constant_op.py", line 106, in constant
    attrs={"value": tensor_value, "dtype": dtype_value}, name=name).outputs[0]
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/framework/ops.py", line 2336, in create_op
    original_op=self._default_original_op, op_def=op_def)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/framework/ops.py", line 1228, in __init__
    self._traceback = _extract_stack()

InternalError (see above for traceback): Dst tensor is not initialized.
	 [[Node: dens/w/Adam/Initializer/Const = Const[_class=["loc:@dens/w"], dtype=DT_FLOAT, value=Tensor<type: float shape: [131072,1024] values: [0 0 0]...>, _device="/job:localhost/replica:0/task:0/gpu:0"]()]]
```

tensorflow/python/ops/array_ops.py
Androbin: I added `tf.convert_to_tensor` for each of them.
For no, only one-dimensional (except `batch_size`) signals are supported.

Androbin:
Which of the two would be faster or more readable or whatsoever?

This is how it's done in `python_speech_features.sigproc.framesig`:
`indices_steps = array_ops.transpose(array_ops.tile(array_ops.expand_dims(math_ops.range(0, num_frames * frame_step, frame_step), 0), [frame_length, 1]))`

But I figured out that, because `expand_dims` is already inserted explicitly:
`indices_steps = array_ops.tile(array_ops.expand_dims(math_ops.range(0, num_frames * frame_step, frame_step), 1), [1, frame_length])`

ebrevdo:
yeah try to avoid a transpose if you can. they're not free.

Androbin:
Would it make a difference to use `tf.zeros` and `tf.concat` instead of `tf.pad` just like they did with numpy?

ebrevdo:
pad is probably a bit faster

Androbin:
What could it possibly be that it has issues inferring `num_frames` as dimension?
`num_frames = 1 + math_ops.cast(math_ops.ceil((signal_length - frame_length) / frame_step), dtypes.int32)`

Androbin:
I think that one issue might be that `tf.gather` seems to apply `indices` to the batch dimension.

Androbin:
At least the first dimension of `indices_steps` is resolved when replacing
`math_ops.range(0, num_frames * frame_step, frame_step)`
by the equivalent of
`math_ops.range(num_frames) * frame_step`

Androbin:
How could `tf.gather_nd` be utilized here? #206
I don't want to `tf.transpose` the `signal` two times ...

Androbin:
When the batch dimension is stripped out entirely, then the only thing still missing would be inferring `num_frames` as dimension correctly.

rryan:
@ebrevdo WDYT about tf.contrib.signal instead? This isn't really a spectral operation per se.

rryan:
Other signal processing ops a la scipy.signal could live there too eventually.

ebrevdo:
SGTM for tf.contrib.signal.

Androbin:
Okay, I moved this to contrib/signal and adapted all references accordingly.

Androbin:
@ebrevdo 
Referring to the issue with `num_frames`:
Could this be an issue with `tf.cast` or even `tf.ceil`?

Androbin:
@ebrevdo 
I have tried a few variations for `num_frames`:
`num_frames = tf.cast(48.0, tf.int32)`
at least results in a proper value.
Any modification to that seems to break it, including:
`num_frames = 1 + tf.cast(48.0, tf.int32)`
`num_frames = tf.cast(tf.ceil(48.0), tf.int32)`

ebrevdo:
Not sure I understand the problem.  Is the function failing?

Androbin:
@ebrevdo 
As described one comment above:
The shape of `tf.range` or `tf.tile` is (partially) undefined depending on the calculation of `num_frames`

Androbin:
And even with `num_frames` hardcoded, then `pad_signal = tf.pad(signal, [[0, 0], [0, pad_length - signal_length]])` is still of undefined shape `(?, ?)` (first is of course the batch dimension)

Androbin:
@ebrevdo 
So, the remaining problem is basically the missing support in `tf.gather` for specifying an `axis` to apply the indices to.
The current workaround is to apply `tf.transpose` before and after.
It would be great if this could be removed.

Androbin:
In the meantime, I will write a basic `TestCase` for this!

Androbin:
Apart from the workaround concerning `tf.gather` using two `tf.transpose`,
what is currently left to do for this PR to get merged?

ebrevdo:
will look tomorrow

tensorflow/contrib/signal/BUILD
```diff
@@ -0,0 +1,30 @@
...
+
+exports_files(["LICENSE"])
+
+package(default_visibility = ["//visibility:public"])
+
+py_library(
+    name = "signal_py",
+    srcs = [
+        "__init__.py",
+        "python/__init__.py",
```
ebrevdo: i think just:
`srcs = ["__init__.py"] + glob(["python/ops/*.py"]),`

tensorflow/contrib/signal/BUILD
```diff
@@ -0,0 +1,30 @@
...
+        "python/ops/__init__.py",
+        "python/ops/shape_ops.py",
+    ],
+    srcs_version = "PY2AND3",
+    visibility = ["//visibility:public"],
+    deps = ["//tensorflow/python:signal"],
+)
+
+filegroup(
```
ebrevdo: needs a cuda_py_test target for your unit test.

tensorflow/contrib/signal/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
```
ebrevdo: from tensorflow.contrib.signal.python.ops.shape_ops import frames

tensorflow/contrib/signal/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
ebrevdo: add:

"""
\@\@frames
"""

tensorflow/contrib/signal/python/kernel_tests/shape_ops_test.py
```diff
@@ -0,0 +1,52 @@
...
+from tensorflow.contrib.signal.python.ops import shape_ops
+from tensorflow.python.framework import constant_op
+from tensorflow.python.framework import dtypes
+from tensorflow.python.ops import array_ops
+from tensorflow.python.platform import test
+
+class FramesTest(test.TestCase):
+  def test_mapping_of_indices_without_padding(self):
+    with self.test_session():
+      tensor = array_ops.expand_dims(constant_op.constant(np.arange(9152), dtypes.int32), 0)
```
ebrevdo: keep within 80 character limit for all code.

tensorflow/contrib/signal/python/kernel_tests/shape_ops_test.py
```diff
@@ -0,0 +1,52 @@
...
+  def test_mapping_of_indices_with_padding(self):
+    with self.test_session():
+      tensor = array_ops.expand_dims(constant_op.constant(np.arange(10000), dtypes.int32), 0)
+      result = shape_ops.frames(tensor, 512, 192)
+      
+      expected = np.tile(np.arange(512), (51, 1)) + np.tile(np.arange(51) * 192, (512, 1)).T
+      expected[expected >= 10000] = 0
+      expected = np.expand_dims(expected, axis=0)
+      
+      np.testing.assert_array_equal(expected, result.eval())
```
ebrevdo: self.assertAllEqual

tensorflow/contrib/signal/python/kernel_tests/shape_ops_test.py
```diff
@@ -0,0 +1,52 @@
...
+class FramesTest(test.TestCase):
+  def test_mapping_of_indices_without_padding(self):
+    with self.test_session():
+      tensor = array_ops.expand_dims(constant_op.constant(np.arange(9152), dtypes.int32), 0)
+      result = shape_ops.frames(tensor, 512, 180)
+      
+      expected = np.tile(np.arange(512), (49, 1)) + np.tile(np.arange(49) * 180, (512, 1)).T
+      expected = np.expand_dims(expected, axis=0)
+      
+      np.testing.assert_array_equal(expected, result.eval())
```
ebrevdo: self.assertAllEqual

tensorflow/contrib/signal/python/kernel_tests/shape_ops_test.py
```diff
@@ -0,0 +1,52 @@
...
+import numpy as np
+
+from tensorflow.contrib.signal.python.ops import shape_ops
+from tensorflow.python.framework import constant_op
+from tensorflow.python.framework import dtypes
+from tensorflow.python.ops import array_ops
+from tensorflow.python.platform import test
+
+class FramesTest(test.TestCase):
+  def test_mapping_of_indices_without_padding(self):
```
ebrevdo: add a simple test where you can write the output exactly as a np.array

ebrevdo (changes requested):

tensorflow/contrib/signal/BUILD
Androbin: So, what about `python/__init__.py` then?

tensorflow/contrib/signal/BUILD
Androbin: Okay, but unfortunately I am somewhat unsure about exactly which `additional_deps` are needed here...

tensorflow/contrib/signal/python/kernel_tests/shape_ops_test.py
Androbin: Okay, I hope to have made things more cleanly.

Androbin:
@ebrevdo 
Most of this should work now, but I can't figure out all of `BUILD > cuda_py_tests > additional_deps`

josh11b:
API review not needed for contrib/ changes.

ebrevdo:
What failures are you getting with your existing additional_deps?

Androbin:
Sorry for the hiccup, I had to figure out how some parts of the framework were organized.
I have now successfully compiled the `py_library` as well as the `cuda_py_tests` and `Ran 3 tests in 0.105s`. But I can only remember writing 2 tests instead of 3?

tensorflow/contrib/signal/python/__init__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
ebrevdo: i *think* this file can be empty (but check other contrib/../python/ directories)

ebrevdo (approved):
LGTM; one small nit.

tensorflow/contrib/signal/python/ops/shape_ops.py
```diff
@@ -0,0 +1,68 @@
...
+    if signal_rank != 2:
+      raise ValueError("expected signal to have rank 2 but was " + signal_rank)
+    
+    signal_length = array_ops.shape(signal)[1]
+    
+    num_frames = 1 + math_ops.cast(math_ops.ceil((signal_length - frame_length) / frame_step), dtypes.int32)
+    pad_length = (num_frames - 1) * frame_step + frame_length
+    pad_signal = array_ops.pad(signal, [[0, 0], [0, pad_length - signal_length]])
+    
+    indices_frames = array_ops.tile(array_ops.expand_dims(math_ops.range(frame_length), 0), [num_frames, 1])
```
ebrevdo: can you wrap these so they're within the 80-character line limit?

tensorflow/contrib/signal/python/ops/shape_ops.py
```diff
@@ -0,0 +1,68 @@
...
+  
+  ```python
+  pcm = tf.placeholder(tf.float32, [None, 9152])
+  frames = tf.contrib.signal.frames(pcm, 512, 180)
+  magspec = tf.abs(tf.spectral.rfft(frames, [512]))
+  image = tf.expand_dims(magspec, 3)
+  ```
+  
+  Args:
+    signal: A `Tensor` of shape [batch_size, signal_length].
```
ebrevdo: ... of shape `[batch_size, signal_length]`.

ebrevdo (commented):
couple more requests.

martinwicke:
Jenkins, test this please.

martinwicke:
Can you add a dependency for the pip package as well? It's in pip_package/BUILD.

tensorflow/contrib/signal/python/ops/shape_ops.py
Androbin: I am sorry, I already noted that. Where do you want me to add that?

tensorflow/contrib/signal/python/ops/shape_ops.py
```diff
@@ -0,0 +1,68 @@
...
+    frame_length: An `int32` or `int64` `Tensor`. The length of each frame.
+    frame_step: An `int32` or `int64` `Tensor`. The step between frames.
+    name: A name for the operation (optional).
+  
+  Returns:
+    A `Tensor` of frames with shape [batch_size, num_frames, frame_length].
+  """
+  with ops.name_scope(name, "frames", [signal, frame_length, frame_step]) as name:
```
Androbin: The length of this line is 82 characters. How would I reasonably wrap this?

tensorflow/contrib/signal/python/ops/shape_ops.py
```diff
@@ -0,0 +1,79 @@
...
+    if signal_rank != 2:
+      raise ValueError("expected signal to have rank 2 but was " + signal_rank)
+    
+    signal_length = array_ops.shape(signal)[1]
+    
+    num_frames = math_ops.ceil((signal_length - frame_length) / frame_step)
+    num_frames = 1 + math_ops.cast(num_frames, dtypes.int32)
+    
+    pad_length = (num_frames - 1) * frame_step + frame_length
+    pad_signal = array_ops.pad(signal, [[0, 0], [0, pad_length - signal_length]])
```
Androbin: The length of this line is 81 characters. How would I reasonably wrap this?

tensorflow/contrib/signal/python/ops/shape_ops.py
```diff
@@ -0,0 +1,79 @@
...
+    
+    indices_frame = array_ops.expand_dims(math_ops.range(frame_length), 0)
+    indices_frames = array_ops.tile(indices_frame, [num_frames, 1])
+    
+    indices_step = array_ops.expand_dims(math_ops.range(num_frames) * frame_step, 1)
```
Androbin: The length of this line is 84 characters. How would I reasonably wrap this?

tensorflow/contrib/signal/python/ops/shape_ops.py
martinwicke: break before [, align the new line to name

tensorflow/contrib/signal/python/ops/shape_ops.py
martinwicke: break before [, align to signal.

tensorflow/contrib/signal/python/ops/shape_ops.py
martinwicke: break before math_ops, indent +4 (align to the c in indices_step)

tensorflow/contrib/signal/python/ops/shape_ops.py
Androbin: Would you typically break before the first, second or third opening brace?

tensorflow/contrib/signal/python/__init__.py
Androbin: Well, this file by @martinwicke isn't either:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/nn/python/__init__.py
Is there any special intent behind this?

martinwicke: Empty __init__.py files can either be completely empty, or they can contain license headers, py3 compatibility imports, and nothing else. I would prefer completely empty, but it doesn't matter much either way.

tensorflow/contrib/signal/python/ops/shape_ops.py
martinwicke: backticks around that.

tensorflow/contrib/signal/python/ops/shape_ops.py
Androbin: I mean, what would usually be considered more readable?

ebrevdo: I usually break on the first paren, and pad 4 chars on the next line:

```
pad_signal = ....pad(
    signal, ....)
```
this is valid and pylint is happy with this form of break.

Androbin:
@martinwicke I added the dependency for the pip package and fixed the formatting.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

tensorflow/contrib/signal/python/ops/shape_ops.py
```diff
@@ -0,0 +1,82 @@
...
+    indices_step = array_ops.expand_dims(
+        math_ops.range(num_frames) * frame_step, 1)
+    indices_steps = array_ops.tile(indices_step, [1, frame_length])
+    
+    indices = indices_frames + indices_steps
+    
+    # TODO(Androbin) remove `transpose` when `gather` gets `axis` support
+    pad_signal = array_ops.transpose(pad_signal)
+    frames = array_ops.gather(pad_signal, indices)
+    frames = array_ops.transpose(frames, perm=[2, 0, 1], name=name)
```
ebrevdo: remove the name=name parameter.

ebrevdo (approved):

tensorflow/python/ops/array_ops.py
ebrevdo: seems ok to punt on that support for now.

tensorflow/python/ops/array_ops.py
ebrevdo: it makes little difference in practice, to use range vs. six.moves.xrange, unless num_frames is large and you're on python2.

tensorflow/contrib/signal/python/ops/shape_ops.py
Androbin: I thought this should be attached to the final op? Should this be moved somewhere else? May the name_scope alias be omitted?

ebrevdo:
When you do things within a name scope context, all ops inside that scope
have the name-scope string associated with their op names.  you can change
your code from:

with ops.name_scope(...) as name:
   ....

to just:

with ops.name_scope(...):
  ....

since you don't need the "name" variable.

Androbin:
@ebrevdo Okay, I have dropped passing the name scope.
@martinwicke Now, there is the issue with the CLA again.

Androbin:
I am not sure about the correctness of the `visibility` values. How is this supposed to be broken down?

vrv:
I have made some fixes myself (ignore CLA bot, all commits are from people who have signed the CLA).

vrv:

vrv:
Seems like an unrelated change, merging.

PR #9244, opened 2017-04-16
fix warning about comparing signed/unsigned types in auto_parallel.cc

eklitzke:
This fixes a harmless (but annoying) GCC warning about comparing signed/unsigned types in auto_parallel.cc.


caisq:

caisq (approved):

Issue #6095, opened 2016-12-05
Add atan2 trigonometry function


jphalip:
Tensorflow has a number of built-in trigonometry functions like `atan` but it doesn't have `atan2` yet.

This can be achieved with some custom code as pointed in this comment: https://github.com/tensorflow/tensorflow/issues/3624#issuecomment-242085609.

There's actually a small mistake on the first line, where `np.pi` should not be added. The corrected code is presented below:

```python
def atan2(y, x):
    angle = tf.select(tf.greater(x,0.0), tf.atan(y/x), tf.zeros_like(x))
    angle = tf.select(tf.logical_and(tf.less(x,0.0),  tf.greater_equal(y,0.0)), tf.atan(y/x) + np.pi, angle)
    angle = tf.select(tf.logical_and(tf.less(x,0.0),  tf.less(y,0.0)), tf.atan(y/x) - np.pi, angle)
    angle = tf.select(tf.logical_and(tf.equal(x,0.0), tf.greater(y,0.0)), 0.5*np.pi * tf.ones_like(x), angle)
    angle = tf.select(tf.logical_and(tf.equal(x,0.0), tf.less(y,0.0)), -0.5*np.pi * tf.ones_like(x), angle)
    angle = tf.select(tf.logical_and(tf.equal(x,0.0), tf.equal(y,0.0)), np.nan * tf.zeros_like(x), angle)
    return angle
```

While this custom code performs reasonably well, it'd be both faster and more convenient if Tensorflow had it built-in like the other existing functions.

michaelisard:
Thanks, please feel free to submit a PR and reference this issue.

tesfaldet:
Is there any way to convert the range to between 0 and 2*pi? I'm trying to mimic opencv's cartToPolar method this way, however, I'm having trouble finding a way to modify elements in a tensor (on a placeholder, so I can't use scatter update) based off of a condition i.e. where angle < 0, add 2*pi.

tesfaldet:
For what it's worth, I managed to implement two versions of atan2. One that follows the opencv style (returns degrees), and another that uses tensorflow's atan function (returns rads).

`def atan2(y, x):
    angle = tf.select(tf.greater(x, 0.0), tf.atan(y / x), tf.zeros_like(x))
    angle = tf.select(tf.greater(y, 0.0), 0.5 * np.pi - tf.atan(x / y), angle)
[...]

PR #9276, opened 2017-04-17
Implement atan2

admcrae:
This addresses #6095.

Since `atan2` is not implemented in Eigen, this is loosely based on the `rint` implementation. Suggestions are welcome.

### DIFF
--- a/tensorflow/core/kernels/cwise_ops.h
+++ b/tensorflow/core/kernels/cwise_ops.h
@@ -664,7 +664,7 @@ struct scalar_atan2_op {
   EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE Scalar
   operator()(const Scalar& y, const Scalar& x) const {
 #if defined(__CUDACC__)
-    return ::atan2(a);
+    return ::atan2(y, x);
 #else
     return std::atan2(y, x);
 #endif
--- a/tensorflow/core/kernels/cwise_ops.h
+++ b/tensorflow/core/kernels/cwise_ops.h
@@ -663,7 +663,7 @@ struct scalar_atan2_op {
   EIGEN_EMPTY_STRUCT_CTOR(scalar_atan2_op)
   EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE Scalar
   operator()(const Scalar& y, const Scalar& x) const {
-#if defined(__CUDACC__)
+#if GOOGLE_CUDA
     return ::atan2(y, x);
 #else
     return std::atan2(y, x);
--- a/tensorflow/core/kernels/cwise_ops.h
+++ b/tensorflow/core/kernels/cwise_ops.h
@@ -658,6 +658,22 @@ struct zeta : base<T, Eigen::internal::scalar_zeta_op<T>> {};
 template <typename T>
 struct polygamma : base<T, Eigen::internal::scalar_polygamma_op<T>> {};
 
+template <typename Scalar>
+struct scalar_atan2_op {
+  EIGEN_EMPTY_STRUCT_CTOR(scalar_atan2_op)
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE Scalar
+  operator()(const Scalar& y, const Scalar& x) const {
+#if defined(__CUDACC__)
+    return ::atan2(a);
+#else
+    return std::atan2(y, x);
+#endif
+  }
+};
+
+template <typename T>
+struct atan2 : base<T, scalar_atan2_op<T>> {};
+
 template <typename T>
 struct squared_difference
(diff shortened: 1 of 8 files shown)


tensorflow/core/ops/math_ops.cc
```diff
@@ -732,6 +732,18 @@ The polygamma function is defined as:
 where \\(\psi(x)\\) is the digamma function.
 )doc");
 
+REGISTER_OP("Atan2")
+    .Input("y: T")
```
drpngx: numpy uses x1, x2. It's ugly but we probably want to stay consistent.

tensorflow/core/kernels/cwise_ops.h
```diff
@@ -658,6 +658,22 @@ struct zeta : base<T, Eigen::internal::scalar_zeta_op<T>> {};
 template <typename T>
 struct polygamma : base<T, Eigen::internal::scalar_polygamma_op<T>> {};
 
+template <typename Scalar>
+struct scalar_atan2_op {
+  EIGEN_EMPTY_STRUCT_CTOR(scalar_atan2_op)
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE Scalar
+  operator()(const Scalar& y, const Scalar& x) const {
+#if defined(__CUDACC__)
```
drpngx: GOOGLE_CUDA

drpngx (changes requested):

drpngx:
That will require and API review.

admcrae:
I've addressed your code comments (although I won't get a chance to test the GPU code myself until tomorrow).

drpngx:
Jenkins, test this please.

drpngx:
Could we add a test that checks random values, and also all special values listed here?
https://docs.scipy.org/doc/numpy/reference/generated/numpy.arctan2.html

Thanks

admcrae:
Done. Gradients for random arguments are rather flaky, unless I clip the inputs to be well away from zero. Perhaps this is because the gradients blow up near the origin. Curiously, I have to clip them more severely for float64 inputs, due to the stricter tolerance on the check.

drpngx (approved):

drpngx:
Jenkins, test this please.

admcrae:
Apparently `math.inf` doesn't exist in python 2. I've fixed that, but it doesn't explain the python 3 failure. How can I see the detailed log of the failure? I can't find it on the CI web interface.

admcrae:
FWIW, running bazel test --config=opt //tensorflow/python/kernel_tests:cwise_ops_test passes now (with my previous commit) for both python 2 and 3 on my local machine. Perhaps you could test it again?

drpngx:
You'll need to add an entry [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/api/golden/tensorflow.pbtxt#L635) to extend the public API.

Jenkins, test this please.

admcrae:
Done.

drpngx:
Jenkins, test this please.

drpngx:
@martinwicke this is an API change, WDYT?

martinwicke (changes requested):
We looked at this and feel that we should follow the saner convention of y, x. Detailed comments inline.

tensorflow/core/ops/math_ops.cc
```diff
@@ -732,6 +732,17 @@ The polygamma function is defined as:
...
+REGISTER_OP("Atan2")
+    .Input("x1: T")
+    .Input("x2: T")
+    .Output("z: T")
+    .Attr("T: {float, double}")
+    .SetShapeFn(shape_inference::BroadcastBinaryOpShapeFn)
+    .Doc(R"doc(
+Computes arctangent of x1/x2 element-wise, respecting signs of the arguments.
+This is the angle of x2 + x1 * i.
+)doc");
```
martinwicke: After looking at the numpy docs and shivering, we decided it'll be best to stick with the time tested atan2(y,x) convention. Could you revise the docstring and arg names?

For reference: 
Python's math.atan2: https://docs.python.org/2/library/math.html
Numpy's arctan2: https://docs.scipy.org/doc/numpy/reference/generated/numpy.arctan2.html

martinwicke (changes requested):

martinwicke:
This is good from an API perspective after reverting to the `(y, x)` signature.

admcrae:
Done.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

drpngx:
Fetch errors.

Jenkins, test this please.

PR #9279, opened 2017-04-17
Add BestSplits op to common windows installation problems.

### LINKED ISSUE #9273: [Docs] Update wheels URLs to match latest TensorFlow release (1.1.0)
The wheels available on the website are from the previous release (1.0.1) which still contain OpKernels errors that are solved at head (as well as on 1.1.0) and can be ignored but are confusing users prompting repeated reports of this issue.

yifeif:
Fix #9273

### DIFF
--- a/tensorflow/docs_src/install/install_windows.md
+++ b/tensorflow/docs_src/install/install_windows.md
@@ -193,5 +193,11 @@ ImportError: cannot import name 'descriptor'</pre>
   <td><pre>No module named "pywrap_tensorflow"</pre></td>
 </tr>
 
+<tr>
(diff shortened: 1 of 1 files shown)


mrry (commented):
This will cease to be a problem in r1.1 though, right?

yifeif:
yea, only for 1.0.1 and its RCs.

mrry:
Here's a link to the other FAQ (about SSE instructions): http://stackoverflow.com/q/43134753

yifeif:
Added to the table as well!

mrry (approved):
Looks great, thanks!

PR #9281, opened 2017-04-18
Close FileWriterCache before moving training directory in estimator_test

### LINKED ISSUE #9185: Windows: //tensorflow/python/estimator:estimator_test failing in Bazel build
http://ci.tensorflow.org/job/tf-master-win-bzl/751/consoleFull

It has been failing on ci for a while with:
```
22:36:33 ======================================================================
22:36:33 ERROR: test_train_save_copy_reload (__main__.EstimatorTrainTest)
22:36:33 ----------------------------------------------------------------------
22:36:33 Traceback (most recent call last):
22:36:33   File "\\?\c:\tmp\Bazel.runfiles_r2z2r52c\runfiles\org_tensorflow\py_test_dir\tensorflow\python\estimator\estimator_test.py", line 267, in test_train_save_copy_reload
22:36:33     os.renames(model_dir1, model_dir2)
22:36:33   File "C:\Program Files\Anaconda3\lib\os.py", line 288, in renames
22:36:33     rename(old, new)
22:36:33 PermissionError: [WinError 5] Access is denied: 'c:\\tmp\\tmp8f7qnomv\\model_dir1' -> 'c:\\tmp\\tmp8f7qnomv\\model_dir2'
```
@gunan Can we fix this test case on Windows? Otherwise we'd better disable this test on Windows.

martinwicke:
Hopefully this fixes #9185 by releasing some stray file handle.

### DIFF
--- a/tensorflow/contrib/testing/python/framework/fake_summary_writer.py
+++ b/tensorflow/contrib/testing/python/framework/fake_summary_writer.py
@@ -127,3 +127,6 @@ class FakeSummaryWriter(object):
 
   def reopen(self):
     pass
+  
+  def close(self):
+    pass
--- a/tensorflow/python/estimator/estimator_test.py
+++ b/tensorflow/python/estimator/estimator_test.py
@@ -262,7 +262,8 @@ class EstimatorTrainTest(test.TestCase):
         model_dir=model_dir1,
         model_fn=model_fn_global_step_incrementer)
     est1.train(dummy_input_fn, steps=5)
-
+    del est1
+    
     model_dir2 = os.path.join(tmpdir, 'model_dir2')
     os.renames(model_dir1, model_dir2)
     est2 = estimator.Estimator(
--- a/tensorflow/python/estimator/estimator_test.py
+++ b/tensorflow/python/estimator/estimator_test.py
@@ -51,6 +51,7 @@ from tensorflow.python.platform import test
 from tensorflow.python.platform import tf_logging as logging
 from tensorflow.python.saved_model import loader
 from tensorflow.python.saved_model import tag_constants
+from tensorflow.python.summary.writer import writer_cache
 from tensorflow.python.training import checkpoint_state_pb2
 from tensorflow.python.training import saver
 from tensorflow.python.training import saver_test_utils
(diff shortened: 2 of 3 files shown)


martinwicke:
I triggered http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/3/console

I hope the setup is the same as tf-master-win-bzl.

@meteorcloudy FYI.

meteorcloudy:
@martinwicke I tested on my local machine as well, it doesn't seem to fix the problem.

But I found `shutil.copytree(model_dir1, model_dir2)` works, can we use this instead of `os.rename`?

martinwicke:
The problem with that would be that this does not delete the old tree. This test is to ensure that no absolute paths creep in anywhere, so this would defang the test.

I cannot see what hangs on to a file handle.

mrry:
Perhaps the easiest approach would be to install the current nightly on one of our Windows test VMs, run a modified version of `estimator_test.py` that blocks before the rename, and apply https://technet.microsoft.com/en-us/sysinternals/processexplorer.aspx to the Python interpreter to see what handles are open?

meteorcloudy:
@mrry It shows this file is in use by python
`c:/tmp/tmp3y2vhvzi/model_dir1/events.out.tfevents.1492671112.TENSORFLOW-JENK`


image

mrry:
Thanks for tracking that down, Yun! Then I think we have a smoking gun: the `tf.summary.FileWriter` is holding open a file handle after `est1.train()` returns.

It looks like this happens deep in the `MonitoredTrainingSession`... and there seems to be a cache(!!) involved:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/summary/writer/writer_cache.py#L27

As a hack, you might try invoking `FileWriterCache.clear()` before the rename, but I'm not sure that that's sufficient because deleting a `FileWriter` doesn't appear to `close()` it, so you might need to add a `writer.close()` call somewhere (or maybe add it to the destructor?).

meteorcloudy:
@mrry Thanks for the hint, Derek! 
I tested adding `writer_cache.FileWriterCache._cache[model_dir1].close()` before rename also solves the problem. Does this sounds good to you?

martinwicke:
That'll fix the test, but I'm wondering whether we can get put that line (or some equivalent) somewhere were it will be called automatically when the Estimator owning the model_dir is deleted?

martinwicke:
I'm even wondering whether we should even keep the cache live when any of train(), eval(), predict() exit. They're all meant to be long running and ideally shouldn't leave open file handles hanging around.

meteorcloudy:
@martinwicke Yes, I totally agree! But I am lack of knowledge how these function works. Can someone with more experience look into this?

martinwicke:
Jenkins, test this please.

martinwicke:
Triggered another Windows build: http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/4/

martinwicke:
I edited FileWriterCache to `close()` all caches it has, and call that in the test. Cleaner than closing the cache directly.

mrry (approved):
LGTM (assuming the test passes)!

martinwicke:
I'll merge to fix build.

PR #9286, opened 2017-04-18
Multiplicative Integration Recurrent Neural Networks

cydonia99:
I implemented Multiplicative Integration variants of recurrent neural networks
(RNN, GRU and LSTM) proposed in 

  Yuhuai Wu, Saizheng Zhang, Ying Zhang, Yoshua Bengio, Ruslan Salakhutdinov,
  On Multiplicative Integration with Recurrent Neural Networks. NIPS, 2016.
  https://arxiv.org/abs/1606.06630

These RNNs are: 
   - MultiplicativeIntegrationRNNCell
   - MultiplicativeIntegrationGRUCell
   - MultiplicativeIntegrationLSTMCell
   - _multiplicative_integration as a helper function
in
tensorflow/contrib/rnn/python/ops/rnn_cell.py

Test codes are in
tensorflow/contrib/rnn/python/kernel_tests/rnn_cell_test.py


cydonia99:
As googlebot said, I changed my email address of github just before I signed the CLA using the changed email address.
# I guess googlebot added cla: no, because I did not log out before signing CLA.

What should I do?

martinwicke:
You need to change the email address in your commits. You can do that with git commit --amend, but depending on how good your git-fu is, that may be too much trouble. If it is, I would recommend making a new PR with a single, clean (new email) commit.

vrv:
There's also conflicts here too.  I'd suggest making a new PR with the correct email after syncing to HEAD.  Sorry about this!

PR #9290, opened 2017-04-18
Update adding_an_op.md

kumasento:
The variable name for the output should better be consistent with the name below.


caisq:
@kumasento, can you explain what "below" refers to? I can't find any instances of "output_flat" or "flat" in this markdown file. So I'm confused about what is inconsistent.

kumasento:
@caisq Thanks for your comments. Maybe you could try to search with `output\_flat`? Or you could find it in the section "Op registration > Attrs", the code snippet in that section contains an example that uses `output_flat`.

Cheers!

kumasento:
Just above this section: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/extend/adding_an_op.md#attr-types-attr-types

caisq:
@kumasento ok - I overlooked that. Thanks.

caisq (approved):

kumasento:
Cheers!

PR #9291, opened 2017-04-18
Docs: Fix #9074 in release branch

### LINKED ISSUE #9074: Documentation: broken links and image in documentation
The URL:  https://www.tensorflow.org/versions/master/api_docs/python/tf/segment_sum has a broken link labelled "the section on Segmentation".  Also, there is a broken image on that page as well.

Same problems here too (except not all have an image to be broken):
   https://www.tensorflow.org/versions/master/api_docs/python/tf/segment_prod
   https://www.tensorflow.org/versions/master/api_docs/python/tf/segment_min
   https://www.tensorflow.org/versions/master/api_docs/python/tf/segment_max
   https://www.tensorflow.org/versions/master/api_docs/python/tf/segment_mean
   https://www.tensorflow.org/versions/master/api_docs/python/tf/unsorted_segment_sum
   https://www.tensorflow.org/versions/master/api_docs/python/tf/sparse_segment_sum
   https://www.tensorflow.org/versions/master/api_docs/python/tf/sparse_segment_mean
   https://www.tensorflow.org/versions/master/api_docs/python/tf/sparse_segment_sqrt_n

The page I believe they should link back to is here: https://www.tensorflow.org/versions/master/api_guides/python/math_ops#Segmentation

I came here because I haven't grokked just what segmented operations are yet.  So I would be much obliged if the explanation of segmentation were more thoroughly be explained there, perhaps with a couple of examples.

I found [a few places] in the code where this could be address.  And I would have attempted a pull request to fix these, but I'm not up on how to generate the documentation to verify that it was fixed.

asimshankar:
Change: 153414133


av8ramit:
Not 100% certain about documentation changes, over to @wolffg for a quick glance.

av8ramit:
Not 100% certain about documentation changes, over to @MarkDaoust  for a quick glance.

MarkDaoust (approved):
LGTM

Note: This only fixes the python docs. The link translator does not currently run on the cc api docs, so I think the "@{$...}" will end up visible on those pages. 

This was thoroughly discussed and accepted in the initial submission. So shouldn't block it here either.

PR #9305, opened 2017-04-19
Add reqeueue operations.

tillahoffmann:
This PR adds `requeue` and `requeue_many` operations to `QueueBase`. The operations dequeue an element/elements from the queue, add the element(s) to the queue again, and return an op that evaluates to the dequeued value.

The motivation behind these operations is to be able to reuse training data that is hard to come by after using them in training.

### DIFF
--- a/tensorflow/python/kernel_tests/fifo_queue_test.py
+++ b/tensorflow/python/kernel_tests/fifo_queue_test.py
@@ -1409,6 +1409,31 @@ class FIFOQueueTest(test.TestCase):
           lambda e: "Before enqueue" in str(e)):
         session.run([a, c])
 
+  def testRequeue(self):
+    with self.test_session() as sess:
+      q = data_flow_ops.FIFOQueue(10, dtypes_lib.float32)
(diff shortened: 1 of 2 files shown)


caisq:

vrv:
I think this is a simple enough change (without enough context) that we should let clients write this code.

The sniff test for adding an API to a core library, whose implementation only has a few lines of code is:

1) How often do people do this?  If rare, then let clients write the code.  I haven't seen this be used very often.

2) Is it easy to write those few lines wrong?  I'd argue in this case, the only challenge is remembering to add the control dependency.  But by that definition, all code that uses control dependencies should be in TF core, which would be unmaintainable.  (We definitely made a mistake in exposing control dependencies to users by default, but that needs to be fixed in a more general way). 

So I'm going to err on the side of saying "this is a thing you can do" but not have the burden of maintenance be transferred to the TF core team, at least until more context is available.

Hope that makes sense, and that I'm being reasonable in your opinion!

tillahoffmann:
Sounds good. We'll keep it in our code for the time being.

PR #9335, opened 2017-04-20
tf.train.Saver: be verbose about ckpt load errors

vmarkovtsev:
I've just had an AbortedError which resulted in the following log:
```
WARNING:tensorflow:/path/to/checkpoint
WARNING:tensorflow:/path/to/checkpoint: Checkpoint ignored
```

With this addition, I get

```
WARNING:tensorflow:tensorflow.python.framework.errors_impl.AbortedError: /path/to/checkpoint
WARNING:tensorflow:/path/to/checkpoint: Checkpoint ignored
```

which is a bit nicer.

### DIFF
--- a/tensorflow/python/training/saver.py
+++ b/tensorflow/python/training/saver.py
@@ -935,11 +935,11 @@ def get_checkpoint_state(checkpoint_dir, latest_filename=None):
           ckpt.all_model_checkpoint_paths[i] = os.path.join(checkpoint_dir, p)
   except errors.OpError as e:
     # It's ok if the file cannot be read
-    logging.warning(str(e))
+    logging.warning("%s: %s" % (type(e).__name__, e))
     logging.warning("%s: Checkpoint ignored", coord_checkpoint_filename)
     return None
   except text_format.ParseError as e:
-    logging.warning(str(e))
+    logging.warning("%s: %s" % (type(e).__name__, e))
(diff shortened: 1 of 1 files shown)


caisq:

martinwicke:
I think it would be better to fill the OpError with better information (or fix its str() to provide proper information) -- just the checkpoint path is a bit silly.

vmarkovtsev:
Disagree. The problem is that there are many inherited errors. Combined exc type and message is the standard way of reporting (see how python reports uncaught exceptions). This allows the exact diagnosis of the problem. Besides, we should not modify all inherited exc types - this goes out of the scope of this PR.

martinwicke:
Even `AbortedError: /path/to/checkpoint` is a pretty useless error description. This change is fine, but it does not address the real failure: The error should contain a useful message. What was the problem with the path contained in it? File not found? Path not found? Permission denied? I haven't tracked down where this error is emitted, but wherever it is, it should set a proper message.

vmarkovtsev:
I agree that errors should contain useful messages. Again, this goes out of the scope of this PR. Let me put it straight: I am not going to code anything extra here. If you feel this change is useless, please reject it.

vrv:
I agree this is an incremental step towards understanding what's going on, so this looks fine to me for now.

Hopefully this allows someone to figure out what is going wrong (I did a search for an empty Aborted error message being thrown in C++ and couldn't find any...)

PR #9350, opened 2017-04-20
Fix issue where TensorBoard run selector can become non-responsive.

teamdandelion:
Various dashboards in TensorBoard store the same runSelector state via the url bar. However, the text dashboard may have a different number of runs, because it only shows runs that have text data to display. Some logic that I added to try to keep the runSelector state short actually causes state updates to be lost when the text dashboard and scalar dashboard have very different numbers of runs. I removed the clever update logic in this commit.


av8ramit (approved):

av8ramit:
core_rnn_cell_test unrelated

PR #9361, opened 2017-04-21
Fix a typo.

cxxgtxy:
Fix typo

### DIFF
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -2240,7 +2240,7 @@ class Graph(object):
 
   @property
   def building_function(self):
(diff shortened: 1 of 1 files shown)


tensorflow/python/framework/ops.py
```diff
@@ -2240,7 +2240,7 @@ def _add_function(self, function):
 
   @property
   def building_function(self):
-    """Returns True iff this graph represents a function."""
+    """Returns True if this graph represents a function."""
```
caisq: This doesn't look like a typo.  "iff" stands for "if and only if" here, I think.

cxxgtxy: thanks!

PR #9393, opened 2017-04-23
Quantize Graph - check whether the key exists before using it

kumasento:
It is a simple fix for a hidden bug in `tensorflow/tools/quantize_graph`, the existence of the node key wasn't checked before using it.

Tests in `//tensorflow/tools/quantization:quantize_graph_test` are all passed.


martinwicke:
Jenkins, test this please.

martinwicke (approved):

PR #9395, opened 2017-04-23
Use AppendShape() instead of looping over and adding the dims of output_matrix_shape individually

ghost:


rmlarsen (approved):

rmlarsen:

PR #9408, opened 2017-04-24
Allocation tracking of pre-tracked buffer returned in a tuple should increment the buffer by 2

DavidNorman:
A buffer not pre-tracked has its ref-count set to 2 when it is part of a tuple. A buffer
pre-tracked has its value increased by 1, whether it is part of a tuple or not.  This is
an error.

This fix increments a pre-tracked buffer by the amount of the 'initial_ref_count', which
is really the number of references that a buffer acquires as part of being returned (1 for
a single buffer, but 2 when it is part of a tuple).

### DIFF
--- a/tensorflow/compiler/xla/service/allocation_tracker.h
+++ b/tensorflow/compiler/xla/service/allocation_tracker.h
@@ -65,7 +65,7 @@ class Allocation {
   }
   void increment_ref_count(int inc) {
     CHECK_GT(ref_count_, 0);
-    CHECK_GT(ref_count_+inc, 0);
+    CHECK_LT(ref_count_, INT_MAX-inc-1);
     ref_count_ += inc;
   }
   void decrement_ref_count() {
--- a/tensorflow/compiler/xla/service/allocation_tracker.h
+++ b/tensorflow/compiler/xla/service/allocation_tracker.h
@@ -65,7 +65,7 @@ class Allocation {
   }
   void increment_ref_count(int inc) {
     CHECK_GT(ref_count_, 0);
-    CHECK_LT(ref_count_, INT_MAX-inc-1);
+    CHECK_LE(ref_count_, INT_MAX - inc);
     ref_count_ += inc;
   }
   void decrement_ref_count() {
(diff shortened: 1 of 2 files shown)


tensorflow/compiler/xla/service/allocation_tracker.h
```diff
@@ -63,10 +63,10 @@ class Allocation {
     CHECK_GE(ref_count_, 0);
     return ref_count_;
   }
-  void increment_ref_count() {
+  void increment_ref_count(int inc) {
     CHECK_GT(ref_count_, 0);
-    CHECK_LT(ref_count_, INT_MAX);
-    ++ref_count_;
+    CHECK_GT(ref_count_+inc, 0);
```
meheffernan: Assuming ref_count and inc are both non-negative, this CHECK_GT will only fail in the event of signed overflow which is undefined behavior so this CHECK can't be relied upon. How about replacing it with:

CHECK_GT(inc, 0);
CHECK_LT(ref_count_, INT_MAX - inc + 1);

meheffernan (changes requested):
Thanks for the fix!

tensorflow/compiler/xla/service/allocation_tracker.h
DavidNorman: ok.  i was assuming that an overflow would result in a negative number. however your solution is nicer.  in practice i wouldn't expect a refcount to get anywhere close to MAX_INT

tensorflow/compiler/xla/service/allocation_tracker.h
```diff
@@ -63,10 +63,10 @@ class Allocation {
     CHECK_GE(ref_count_, 0);
     return ref_count_;
   }
-  void increment_ref_count() {
+  void increment_ref_count(int inc) {
     CHECK_GT(ref_count_, 0);
-    CHECK_LT(ref_count_, INT_MAX);
-    ++ref_count_;
+    CHECK_LT(ref_count_, INT_MAX-inc-1);
```
meheffernan: I think this should be:

CHECK_LT(ref_count_, INT_MAX - inc + 1);

or perhaps more clearly:

CHECK_LE(ref_count_, INT_MAX - inc);

Agreed that this is largely of theoretical interest as reference counts should never get close to INT_MAX :-)

DavidNorman: you're absolutely right - silly error

meheffernan (approved):
Looks good. Thanks!

vrv:

PR #9409, opened 2017-04-24
Add scaffolding for running XLA python tests on a plugin backend.

DavidNorman:
Copy the plugin_config.tpl file to plugin_config.py and configure
the two lines to name the plugin device and types that the device
supports.  The device can the be referred to as 'plugin', eg there
will be test targets called:

tensorflow/compiler/tests:plugin_tests
tensorflow/compiler/test:binary_ops_test_plugin

etc..


tensorflow/compiler/tests/plugin_config.tpl
```diff
@@ -0,0 +1,14 @@
+# Configuration for a plugin
```
vrv: (we may need copyright notices on all files, so if you can add one, it would be good!)

DavidNorman: sure thing... will do

tensorflow/compiler/tests/xla_test.py
```diff
@@ -44,18 +44,28 @@
...
 class XLATestCase(test.TestCase):
   """XLA test cases are parameterized test cases."""
 
   def __init__(self, method_name='runTest'):
     super(XLATestCase, self).__init__(method_name)
-    self.device = FLAGS.test_device
+    test_device = FLAGS.test_device
+    test_types = FLAGS.types
+
+    if FLAGS.load_plugin:
```
hawkinsp: I think we could simplify this change. The design here is that we already take the type and device inputs via flags passed from the BUILD file. I don't think we need to do anything different for the plugin case --- we can still pass these parameters via flags; all the magic can live in the build file.

I think you can revert the changes to xla_test.py and remove plugin_config.tpl.

tensorflow/compiler/tests/build_defs.bzl
```diff
@@ -39,6 +39,12 @@ def tf_xla_py_test(name, srcs=[], deps=[], tags=[], data=[], main=None,
     disabled_backends = []
 
   enabled_backends = [b for b in all_backends() if b not in disabled_backends]
+
+  load_plugin = False
+  if "xla_test" in native.existing_rules():
+    if ":plugin_config.py" in native.existing_rules()["xla_test"]["srcs"]:
+      load_plugin = True
+
```
hawkinsp: Why not make the pluggable thing a fragment of build rules, instead?

DavidNorman:
I can see how the device name and type list can be pulled out of some bazel rule (using the 'native.' extensions).

I'm not sure how to set up a bazel rule that depends on another rule that may not exist.  I wouldn't like to have to modify a file in the repo in order to enable plugin support.  if I was going to do that then I might as well just add my device in the same way as the GPU one is.

I will investigate optional dependencies in bazel to see if that is a thing.

DavidNorman:
ok - i see that the Skylark load function is more interesting than I thought.  I will have a go with that

DavidNorman:
no - the load statement isn't acceptable within a skylark function.  it is something that is processed during the loading of the files, before the building of the rules begins.

there are no ways of loading configuration information before rule generation apart from through .bzl/BUILD files.  These can only be included unconditionally, and therefore have to be part of the repository, or generated by the ./configure command.  I don't think that either of those is nice.

If the bazel could accept rule generation time config options which have value (instead of being on/off), then it could be done.

I think it is simpler to inject the config file into the runtime environment of the test and let the python script load it.

hawkinsp:
I was thinking along the lines of unconditionally loading an dummy plugin .bzl file, and you can replace the contents of that file in your tree with something more meaningful. That way you don't need to conditionally load anything. Would something like that work?

(I think the right "real" solution is to set this up at configure time, but that's probably a bit annoying to do.)

DavidNorman:
I can see that would work.  The reason I don't like that is that both of its side effects are unpleasant:

1) either that a dummy file has to be created before the repo will build.  if configure created it then that would be ok, but configure isn't very well designed or extensible at the moment.
2) or that there will be a file in the repo that will be necessary to edit, but not to check in.  its messy and prone to error.

I've never checked this out, but maybe you can check a file into the repo, and also include it in .gitignore.  if that works, then it is a good solution.

DavidNorman:
No - unfortunately that isn't how git seems to work.  you can't check a file in, then include it in .gitignore so that any future changes are ignored.  however - apparently you can update the index to mark it as assume-unchanged.  this works ok.

i have submitted a new diff

DavidNorman:
Ok - i've squashed the commits -- now at:

https://github.com/tensorflow/tensorflow/pull/9759

DavidNorman:
Ok - i've squashed the commits -- now at:

https://github.com/tensorflow/tensorflow/pull/9759

PR #9426, opened 2017-04-25
Add link for Xavier initializer paper and update example for zero_faction function

nghiattran:
On https://www.tensorflow.org/api_docs/python/tf/nn/zero_fraction:
* The example for `zero_faction` uses an artificial function `tf.Relu` and deprecated one ` tf.contrib.deprecated.scalar_summary`.

On, https://www.tensorflow.org/api_docs/python/tf/contrib/layers/xavier_initializer:
* Added a link to Xavier Glorot and Yoshua Bengio's paper for quick reference.
* Fixed broken link to https://www.tensorflow.org/api_guides/python/constant_op#random-tensors.

### DIFF
--- a/tensorflow/contrib/layers/python/layers/initializers.py
+++ b/tensorflow/contrib/layers/python/layers/initializers.py
@@ -34,9 +34,9 @@ def xavier_initializer(uniform=True, seed=None, dtype=dtypes.float32):
   This function implements the weight initialization from:
 
   Xavier Glorot and Yoshua Bengio (2010):
-           Understanding the difficulty of training deep feedforward neural
+           [Understanding the difficulty of training deep feedforward neural
            networks. International conference on artificial intelligence and
-           statistics.
+           statistics.](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.207.2059&rep=rep1&type=pdf)
 
   This initializer is designed to keep the scale of the gradients roughly the
   same in all layers. In uniform distribution this ends up being the range:
--- a/tensorflow/contrib/layers/python/layers/initializers.py
+++ b/tensorflow/contrib/layers/python/layers/initializers.py
@@ -46,7 +46,7 @@ def xavier_initializer(uniform=True, seed=None, dtype=dtypes.float32):
   Args:
     uniform: Whether to use uniform or normal distributed random initialization.
     seed: A Python integer. Used to create random seeds. See
-      [`set_random_seed`](../../api_docs/python/constant_op.md#set_random_seed)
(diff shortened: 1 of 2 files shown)


sguada (approved):

tensorflow/contrib/layers/python/layers/initializers.py
```diff
@@ -46,7 +46,7 @@ def xavier_initializer(uniform=True, seed=None, dtype=dtypes.float32):
   Args:
     uniform: Whether to use uniform or normal distributed random initialization.
     seed: A Python integer. Used to create random seeds. See
-      [`set_random_seed`](../../api_docs/python/constant_op.md#set_random_seed)
+      [`set_random_seed`](../../../../../api_guides/python/constant_op#random-tensors)
```
martinwicke: Can you use `${$constant_op#random-tensors}` instead of the path?

martinwicke (changes requested):

tensorflow/contrib/layers/python/layers/initializers.py
nghiattran: Sorry, I am not fully sure why this is the case. One of the tool you guys are using will interpolate that token with actual values when building tensorflow.org?

If so, you want me to change be like below, right?
```py
[`set_random_seed`](${$constant_op#random-tensors})
```

martinwicke: Yes, we have a doc generator which fills in that token with a link. And I lied to you, you should write:

```prettyprint
@{$constant_op#random-tensors$set_random_seed}
```

This will produce the whole markdown link, including the [] part.

You could also simply write @{set_random_seed}, which will link to that function's docs, which in turn link to the guide. I think that's preferred.

Which should turn into the link you want. We will have public docs on how to write docs soon, I can't wait.

tensorflow/contrib/layers/python/layers/initializers.py
```diff
@@ -46,8 +46,7 @@ def xavier_initializer(uniform=True, seed=None, dtype=dtypes.float32):
   Args:
     uniform: Whether to use uniform or normal distributed random initialization.
     seed: A Python integer. Used to create random seeds. See
-      [`set_random_seed`](../../../../../api_guides/python/constant_op#random-tensors)
-      for behavior.
+      @{set_random_seed} for behavior.
```
nghiattran: Just got the link fixed. I also notice that there is another link to `set_random_seed` so I fixed it too,

martinwicke (approved):

martinwicke:
Jenkins, test this please.

PR #9429, opened 2017-04-25
Make orthogonal initializer use qr factorization (instead of svd)

GBLin5566:
Use QR factorization in orthogonal initializer to get better efficiency when computing orthogonal matrix.

### DIFF
--- a/tensorflow/contrib/keras/python/keras/initializers_test.py
+++ b/tensorflow/contrib/keras/python/keras/initializers_test.py
@@ -120,7 +120,7 @@ class KerasInitializersTest(test.TestCase):
                    target_mean=0., target_std=None, target_max=2 * scale)
 
   def test_orthogonal(self):
-    tensor_shape = (10, 10)
+    tensor_shape = (20, 20)
     with self.test_session():
       self._runner(keras.initializers.orthogonal(seed=123), tensor_shape,
                    target_mean=0.)
--- a/tensorflow/python/ops/init_ops.py
+++ b/tensorflow/python/ops/init_ops.py
@@ -40,6 +40,7 @@ from tensorflow.python.framework import dtypes
 from tensorflow.python.ops import array_ops
 from tensorflow.python.ops import linalg_ops
 from tensorflow.python.ops import random_ops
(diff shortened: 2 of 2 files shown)


yaroslavvb:
BTW, that doesn't give uniform distribution over orthogonal matrices. Search for "wrong distribution" in https://arxiv.org/pdf/math-ph/0609050.pdf

To do uniformly at random, need to divide by diagonal entries of the qr term. IE, something like this in numpy
```
from scipy import linalg
import numpy as np
def orthogonal_measure(n):
    z = np.random.randn(n,n)
    q, r = linalg.qr(z)
    d = np.diagonal(r)
    ph = d/np.absolute(d)
    return q*ph

```

before normalization
screenshot 2017-04-25 13 09 13

after normalization
screenshot 2017-04-25 13 09 27

[Notebook with sanity checks](https://wolfr.am/lbWEtaFF)

GBLin5566:
@yaroslavvb thanks for pointing it out !
Gonna fix it soon.

keveman:
Not super familiar with this code. Lobbing over to @rmlarsen

rmlarsen (approved):

rmlarsen:
Thanks for the change!

rmlarsen:

martinwicke:
There is an issue with the tests in a ominous location -- @GBLin5566 can you look into this?

GBLin5566:
I checked the error message and found that it was caused due to the mean of Q was bigger than 0. over the threshold. (In this case, mean of Q is  0.04, bigger than +- 0.03).
However, if I change the shape of array or change the seed (123 to 124 for example), the test will be fine.
So I guess the error is from minor instability and it's permissible ?

martinwicke:
It is probably ok, but it does cast some doubt on the quality of this test. Increasing the threshold (or changing the seed) are acceptable, assuming we have made extra certain that our new scheme does not introduce some bias.

martinwicke:
Oops.

GBLin5566:
@martinwicke Thanks for the feedback !

PR #9432, opened 2017-04-25
Add ibverbs support for RoCEv2 networks

li-ch:
Current support for ibverbs cannot run on RoCE networks, as RoCE requires the global identifier (gid) to connect. This patch adds gid information for RDMA connection setup.

### DIFF
--- a/tensorflow/contrib/verbs/grpc_verbs_service.cc
+++ b/tensorflow/contrib/verbs/grpc_verbs_service.cc
@@ -117,6 +117,7 @@ Status GrpcVerbsService::GetRemoteAddressSync(
   ra.lid = request->channel().lid();
   ra.qpn = request->channel().qpn();
   ra.psn = request->channel().psn();
+  ra.gid = request->channel().gid();
   rc->SetRemoteAddress(ra, false);
   rc->Connect();
   int i = 0;
@@ -146,6 +147,7 @@ Status GrpcVerbsService::GetRemoteAddressSync(
   channel_info->set_lid(rc->self().lid);
   channel_info->set_qpn(rc->self().qpn);
   channel_info->set_psn(rc->self().psn);
+  channel_info->set_psn(rc->self().gid);
   for (int i = 0; i < RdmaChannel::kNumMessageBuffers; i++) {
     MemoryRegion* mr = response->add_mr();
     mr->set_remote_addr(reinterpret_cast<uint64>(mb[i]->buffer()));
--- a/tensorflow/contrib/verbs/grpc_verbs_service.cc
+++ b/tensorflow/contrib/verbs/grpc_verbs_service.cc
@@ -117,7 +117,8 @@ Status GrpcVerbsService::GetRemoteAddressSync(
   ra.lid = request->channel().lid();
   ra.qpn = request->channel().qpn();
   ra.psn = request->channel().psn();
-  ra.gid = request->channel().gid();
+  ra.snp = request->channel().snp();
+  ra.iid = request->channel().iid();
   rc->SetRemoteAddress(ra, false);
   rc->Connect();
   int i = 0;
@@ -147,7 +148,8 @@ Status GrpcVerbsService::GetRemoteAddressSync(
   channel_info->set_lid(rc->self().lid);
   channel_info->set_qpn(rc->self().qpn);
   channel_info->set_psn(rc->self().psn);
-  channel_info->set_psn(rc->self().gid);
+  channel_info->set_snp(rc->self().snp);
+  channel_info->set_iid(rc->self().iid);
   for (int i = 0; i < RdmaChannel::kNumMessageBuffers; i++) {
     MemoryRegion* mr = response->add_mr();
     mr->set_remote_addr(reinterpret_cast<uint64>(mb[i]->buffer()));
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -477,7 +477,7 @@ void RdmaChannel::Connect(const RdmaAddress& remoteAddr) {
     attr.dest_qp_num = remoteAddr.qpn;
     attr.rq_psn = remoteAddr.psn;
     attr.max_dest_rd_atomic = 1;
-    attr.min_rnr_timer = 12;+    
+    attr.min_rnr_timer = 12;   
     attr.ah_attr.is_global = 1;
     attr.ah_attr.grh.dgid.global.subnet_prefix = remoteAddr.snp;
     attr.ah_attr.grh.dgid.global.interface_id = remoteAddr.iid;
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -271,6 +271,10 @@ RdmaChannel::RdmaChannel(const RdmaAdapter* adapter, const string local_name,
     self_.lid = attr.lid;
     self_.qpn = qp_->qp_num;
     self_.psn = static_cast<uint32_t>(random::New64()) & 0xffffff;
+    union ibv_gid gid;
+    CHECK(!ibv_query_gid(adapter_->context_, (uint8_t) 1, 0, &gid)) << "Query gid";
+    self_.snp = gid.global.subnet_prefix;
+    self_.iid = gid.global.interface_id;
   }
 
   // create message and ack buffers, then initialize the tables.
@@ -320,11 +324,13 @@ void RdmaChannel::SetRemoteAddress(const RdmaAddress& ra, bool override) {
     remote_.lid = ra.lid;
     remote_.qpn = ra.qpn;
     remote_.psn = ra.psn;
+    remote_.gid = ra.gid;
     remote_set_ = true;
   } else {
     CHECK(remote_.lid == ra.lid);
     CHECK(remote_.qpn == ra.qpn);
     CHECK(remote_.psn == ra.psn);
+    CHECK(remote_.gid == ra.gid);
   }
 }
 
@@ -471,8 +477,12 @@ void RdmaChannel::Connect(const RdmaAddress& remoteAddr) {
     attr.dest_qp_num = remoteAddr.qpn;
     attr.rq_psn = remoteAddr.psn;
     attr.max_dest_rd_atomic = 1;
-    attr.min_rnr_timer = 12;
-    attr.ah_attr.is_global = 0;
+    attr.min_rnr_timer = 12;+    
+    attr.ah_attr.is_global = 1;
+    attr.ah_attr.grh.dgid.global.subnet_prefix = remoteAddr.snp;
+    attr.ah_attr.grh.dgid.global.interface_id = remoteAddr.iid;
+    attr.ah_attr.grh.flow_label = 0;
+    attr.ah_attr.grh.hop_limit = 3; 
     attr.ah_attr.dlid = remoteAddr.lid;
     attr.ah_attr.sl = 0;
     attr.ah_attr.src_path_bits = 0;
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -324,13 +324,15 @@ void RdmaChannel::SetRemoteAddress(const RdmaAddress& ra, bool override) {
     remote_.lid = ra.lid;
     remote_.qpn = ra.qpn;
     remote_.psn = ra.psn;
-    remote_.gid = ra.gid;
+    remote_.snp = ra.snp;
+    remote_.iid = ra.iid;
     remote_set_ = true;
   } else {
     CHECK(remote_.lid == ra.lid);
     CHECK(remote_.qpn == ra.qpn);
     CHECK(remote_.psn == ra.psn);
-    CHECK(remote_.gid == ra.gid);
+    CHECK(remote_.snp == ra.snp);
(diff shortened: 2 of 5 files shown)


li-ch:
I signed it!

jhseu:
@junshi15 mind also taking a look?

junshi15:
taking a look now..

tensorflow/contrib/verbs/rdma_mgr.cc
```diff
@@ -69,6 +69,7 @@ void RdmaMgr::SetupChannels() {
     channel_info->set_lid(rc->self_.lid);
     channel_info->set_qpn(rc->self_.qpn);
     channel_info->set_psn(rc->self_.psn);
+    channel_info->set_psn(rc->self_.gid);
```
junshi15: set_gid?
where is self_.gid defined?

tensorflow/contrib/verbs/rdma_mgr.cc
```diff
@@ -85,6 +86,7 @@ void RdmaMgr::SetupChannels() {
       ra.lid = resp.channel().lid();
       ra.qpn = resp.channel().qpn();
       ra.psn = resp.channel().psn();
+      ra.gid = resp.channel().gid();
```
junshi15: where is gid defined?
I only see snp and iid in verbs_service.proto

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -471,8 +477,12 @@ void RdmaChannel::Connect(const RdmaAddress& remoteAddr) {
...
     attr.max_dest_rd_atomic = 1;
-    attr.min_rnr_timer = 12;
-    attr.ah_attr.is_global = 0;
+    attr.min_rnr_timer = 12;   
+    attr.ah_attr.is_global = 1;
+    attr.ah_attr.grh.dgid.global.subnet_prefix = remoteAddr.snp;
+    attr.ah_attr.grh.dgid.global.interface_id = remoteAddr.iid;
+    attr.ah_attr.grh.flow_label = 0;
+    attr.ah_attr.grh.hop_limit = 3; 
     attr.ah_attr.dlid = remoteAddr.lid;
```
junshi15: Can you verify the original RDMA (without RoCE) still work? I am wondering if some of the ah_attr settings will affect the raw RDMA functionalities (especially attr.ah_attr.is_global = 0 --> 1).

tensorflow/contrib/verbs/grpc_verbs_service.cc
```diff
@@ -117,6 +117,7 @@ Status GrpcVerbsService::GetRemoteAddressSync(
   ra.lid = request->channel().lid();
   ra.qpn = request->channel().qpn();
   ra.psn = request->channel().psn();
+  ra.gid = request->channel().gid();
```
junshi15: gid is not defined

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -320,11 +324,13 @@ void RdmaChannel::SetRemoteAddress(const RdmaAddress& ra, bool override) {
     remote_.lid = ra.lid;
     remote_.qpn = ra.qpn;
     remote_.psn = ra.psn;
+    remote_.gid = ra.gid;
     remote_set_ = true;
   } else {
     CHECK(remote_.lid == ra.lid);
     CHECK(remote_.qpn == ra.qpn);
     CHECK(remote_.psn == ra.psn);
+    CHECK(remote_.gid == ra.gid);
```
junshi15: gid is not defined.

junshi15 (commented):
@li-ch 
Thanks for your patch. Below are my comments.

1) Can you make sure this patch compiles? I don't see where "gid" is defined? Besides changes in the proto file, you also need to modify rdma.h accordingly. 

2) Please make sure your changes in ah_attr do not affect original RDMA operations. We need to support both modes, not just RoCE.

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -271,6 +271,10 @@ RdmaChannel::RdmaChannel(const RdmaAdapter* adapter, const string local_name,
     self_.lid = attr.lid;
     self_.qpn = qp_->qp_num;
     self_.psn = static_cast<uint32_t>(random::New64()) & 0xffffff;
+    union ibv_gid gid;
+    CHECK(!ibv_query_gid(adapter_->context_, (uint8_t) 1, 0, &gid)) << "Query gid";
```
shamoya: I think we must have from the user the port number.
There are a lot of rdma devices with 2 ports and we can't assume the first is the one configured.
We can (optionally) get the port number from the user in server creation and expose it here.
@junshi15 - we need it also for the device selection (today it take ibv_devices_list[0]). 
I can do that tomorrow with pleasure, unless you already have something.

tensorflow/contrib/verbs/rdma.cc
shamoya: @junshi15 - you can use ah_attr.is_global=1 in IB, it's not a must since since you have LID (Assuming L2 IB fabric)
In RoCE it's a must since the MAC is calculated from the GID. 

@il-ch - Thanks for the patch. Just curious about the hop_limit=3.
Usually it 1. Do you have L3 (router) Infiniband fabric?

tensorflow/contrib/verbs/rdma.cc
junshi15: @shamoya - Thanks for taking a look at this patch. 
Regarding device selection, @bkovalev raised this issue to me. However, I have not had time to work on it. Feel free to send a PR.

tensorflow/contrib/verbs/rdma_mgr.cc
li-ch: Thanks for the comments, GID definition is added to rdma.h.

tensorflow/contrib/verbs/rdma.cc
li-ch: GID definition is added to rdma.h. 61d985a

tensorflow/contrib/verbs/rdma.cc
li-ch: @junshi15 I currently only have RoCE network. I can try to find a Infiniband network to test on.

@shamoya Since RoCEv2 packets have IP/UDP headers, they can be routed in multi-hop IP networks. The testbed I'm using have 2-tier Ethernet network (leaf-spine topology) with support for 802.1Qbb (data center bridging). I choose hop-limit=3 so that the RoCEv2 packets can be routed in a 2-tier network (leaf-switch --> spine-switch --> leaf-switch). I think it could be set higher to accommodate large-scale networks with more switching tiers.

tensorflow/contrib/verbs/grpc_verbs_service.cc
li-ch: GID definition is added to rdma.h. 61d985a

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -320,11 +324,13 @@ void RdmaChannel::SetRemoteAddress(const RdmaAddress& ra, bool override) {
     remote_.lid = ra.lid;
     remote_.qpn = ra.qpn;
     remote_.psn = ra.psn;
+    remote_.gid = ra.gid;
```
junshi15: there is no member called "gid" in RdmaAddress.

tensorflow/contrib/verbs/grpc_verbs_service.cc
```diff
@@ -146,6 +147,7 @@ Status GrpcVerbsService::GetRemoteAddressSync(
   channel_info->set_lid(rc->self().lid);
   channel_info->set_qpn(rc->self().qpn);
   channel_info->set_psn(rc->self().psn);
+  channel_info->set_psn(rc->self().gid);
```
junshi15: set_psn? typo?

junshi15 (commented):
@li-ch - you add "snp" and "iid" to struct RdmaAddress and proto definition, but you use .gid everywhere else. Does the code compile on your local box (with verbs turned on)?

tensorflow/contrib/verbs/rdma.cc
shamoya: Thanks @il-ch, sorry I was mistaken (not related to IB as the fabric is UDP-IP-ETH).
So I recommend we set the highest number from hop_limit (0xFF).
I think this the case when you set hop_limit=1 in RoCE (I think this code is in the verbs).

junshi15:
The patch looks good to me. I compiled it successfully on Ubuntu 14.04. I did not have a chance to run tests as our IB networks were not available at this moment. 

@shamoya - how does it look to you? especially the settings for attr.ah_attr?

shamoya:
I can verify all is working on IB and RoCE on our cluster soon .. I'll run and update

jhseu:
Jenkins, test this please

li-ch:
@junshi15 @shamoya Thank you so much for your help!

On my RoCE testbed, it can compile and run successfully.

shamoya:
Hi @junshi15  @li-ch @jhseu @poxvoculi @bkovalev,

After thinking about it in the last day, we must move to rdmacm for the connection management part !
This makes much more sense and the the correct way to code verbs applications.

This way rdmacm will configure the correct rdma device and port, according to the gRPC channels given by the user in the cluster definition (the rdma device+port which belongs to the IP Address of each server in cluster specification will be chosen and configured correctly by a code that specialized in it).

This will also eliminate the need for this code, as rdmacm will configure anything needed for the connection according to the link detected (IB or Ethernet) on the corresponding interface of the specified gRPC channel, and if Ethernet it will also take care of  RoCE V0/V1/V2 configuration according to user desired choice (The user configures the GID tables specifying which RoCE mode to configure for each gid index).

rdmacm knows the best how to configure everything in the connection (you don't need to set all the fields in ah_attr or even pass the QPs between its states) and it's a well maintained code by the rdma community - https://github.com/linux-rdma/rdma-core/tree/master/librdmacm

Also this way I don't need the user the specify the rdma/port from the Server definition in the python.
This is the way to go also when you use the same cluster for Ethernet and IB (different adapters).

I can start working on it immediately and shouldn't take long (few days).
So what do you think ?

junshi15:
@shamoya 
I totally support the idea of using rdma_cm to establish the cluster. I compared rdma_cm and out-of-band communication (gRPC in this case) during the early stage of my design. I chose the latter mostly out of  convenience: I had working code from CaffeOnSpark, which uses Spark to establish the cluster. In addition, we only use IB fabric in house, no need to worry about the diversity of the family of IB networks. Rdma_cm should make it easy to support most of them with minimal effort. 

Note we still need specify host:port in python server definition as we use gRPC for sending graphs, etc.
 
If this patch looks good, let's approve it. We can start another PR for rdma_cm. I am happy to work with you.

shamoya:
No problem @junshi15, Thanks.
I Checked the code on our setup and it's working good.

jhseu (approved):

llhe:
@shamoya I tested this in our RoCE environment, and the session run will get stuck. Here is the test scripts:
https://gist.github.com/llhe/da4e6ac1be65291d30de08a90ecc01e2

byronyi:
I saw (line 85) 
```sess = tf.Session("grpc://" + cluster['worker'][0], config=default_config())```
 in your code. 

Shouldn't it be 
```sess = tf.Session("grpc+verbs://" + cluster['worker'][0], config=default_config())```
?

llhe:
@byronyi I think the verbs are supported at Server level instead of Session level. Only 'grpc' session is registered in the session factory.

junshi15:
@llhe, you code looks ok to me. Did you try it on an IB network or you only have RoCE? 

@li-ch @shamoya Mind taking a look?

shamoya:
Very cool script @llhe , thanks.

Now I rechecked again and indeed the code doesn't work on Ethernet (RoCE) setup.
It does work on IB.
Also rechecked with our scripts (Inception and VGG) and it doesn't work as well on Ethernet/RoCE.
We probably checked IB last week and made some error when checking the RoCE case.

I see the following issue in my case:
One of the worker (usually task 1) will report "retry count exceeded".
Usually there's an issue with the connection (we connected QPs but for some reason, the packets one QP send doesn't reach to other side QP).
All the GIDs configuration looks ok, I'm debugging it now.
@Ilhe - is this what you see as well ? 

I hope to send the rdmacm code soon and then we won't fall (hopefully) on those connection bugs.

shamoya:
OK @llhe @junshi15 @li-ch - I see the issue now.

It comes from:
(line 478 rdma.cc) attr.path_mtu = IBV_MTU_4096; 

This won't work on standard 1500 MTU Ethernet network of course.
In this case (1500 MTU Ethernet network) one should use IBV_MTU_1024 (RoCE MTU are always power of 2).

I guess whoever didn't noticed it, is working with MTU 9000 (or above 4K) on his Ethernet network.

Also in IB network this is wrong (since the MTU configured can also different, depends on how SM was configured) but will most likely work as 4K as the usually the MTU.

Another good reason to use rdmacm, as it deals with it in the best way (queries the correct MTU configured on the device, in both IB/Ethernet cases).

Until we'll have the rdmacm code (I hope soon, having some delays here) I think we should change to attr.path_mtu = IBV_MTU_1024 - which will work for 99.99% of the cases and the performance will not be felt.

shamoya:
Verified there isn't performance impact for 1K MTU and sent PR #9667

PR #9438, opened 2017-04-25
Delete an unnecessary include and define

ghost:


vrv:
Thanks for the PR!  I'm not entirely sure this is "safe" to do given the way Eigen works -- that define needs to be before the first include of the Tensor class so that the threaded implementations in Eigen are used.  Looking at some of the includers of this file, they are not defining EIGEN_USE_THREADS at the very beginning.

cc @rmlarsen 

I have two suggestions:

1. Find all cc files that include linalg_ops_common and add the #define at the top, so you can remove the need for the define in the header file.  This will take some work because you need to add the EIGEN_USE_THREADS in a bunch of places.

2. Leave the code as is -- less ideal since I think it's bad practice to put this define in a header file, but certainly less work :)

vrv:
(For example, other ops that use Eigen typically define this field at the top of the implementation, like https://github.com/tensorflow/tensorflow/blob/754048a0453a04a761e112ae5d99c149eb9910dd/tensorflow/core/kernels/bias_op.cc#L18).

dtrebbien:
Hello Vijay,

I agree with you that defining `EIGEN_USE_THREADS` before including "third_party/eigen3/unsupported/Eigen/CXX11/Tensor" is so that the multi-threaded implementations of Eigen-unsupported Tensor ops will be enabled. In this case, it's a little confusing to have that preprocessor define in there, and an include of the Eigen-unsupported Tensor library, because the linalg ops don't actually use the Tensor library; they are using the Eigen Matrix library. So, effectively the presence of the define and include is unnecessarily slowing down compilation of linalg ops.

Daniel

vrv:
Thanks for the clarification!  Since @rmlarsen has written most of this code, I'll let him take over :)

rmlarsen:
@dtrebbien This goes back to before I touched this code and must have been copied from some generic op template back when the world was young. Thanks for the cleanup. You are correct that these are not needed for the linear algebra ops.

rmlarsen (approved):

rmlarsen:

Issue #7543, opened 2017-02-15
Error on building with non-system GCC: "gcc: error trying to exec 'as': execvp: No such file or directory"


dchirikov:
# Description
Hi,

I am trying to compile TensorFlow with non-default GCC. 5.4.0, and get follwing error:
```
gcc: error trying to exec 'as': execvp: No such file or directory
ERROR: /local/cvsupport/_bazel_cvsupport/c61d2ac558d6d30ef2694b9af72e4144/external/nccl_archive/BUILD.bazel:33:1: output 'external/nccl_archive/_objs/nccl/external/nccl_archive/src/libwrap.cu.pic.o' was not created.
ERROR: /local/cvsupport/_bazel_cvsupport/c61d2ac558d6d30ef2694b9af72e4144/external/nccl_archive/BUILD.bazel:33:1: not all outputs were created or valid.
Target //tensorflow/tools/pip_package:build_pip_package failed to build
```
# Environment
```
$ cat /etc/redhat-release 
CentOS Linux release 7.3.1611 (AltArch)

$ uname -a
Linux power004.cluster 3.10.0-514.6.1.el7.ppc64le #1 SMP Thu Jan 19 14:34:54 GMT 2017 ppc64le ppc64le ppc64le GNU/Linux

$ ls -l /trinity/shared/apps/cv-ppc64le/nvidia/cuda/8.0/lib64/libcud*
-rw-r--r-- 1 root root 559800 Oct 29 10:22 /trinity/shared/apps/cv-ppc64le/nvidia/cuda/8.0/lib64/libcudadevrt.a
lrwxrwxrwx 1 root root     16 Feb 14 23:26 /trinity/shared/apps/cv-ppc64le/nvidia/cuda/8.0/lib64/libcudart.so -> libcudart.so.8.0
lrwxrwxrwx 1 root root     19 Feb 14 23:26 /trinity/shared/apps/cv-ppc64le/nvidia/cuda/8.0/lib64/libcudart.so.8.0 -> libcudart.so.8.0.54
-rwxr-xr-x 1 root root 476024 Oct 29 10:22 /trinity/shared/apps/cv-ppc64le/nvidia/cuda/8.0/lib64/libcudart.so.8.0.54
-rw-r--r-- 1 root root 966166 Oct 29 10:22 /trinity/shared/apps/cv-ppc64le/nvidia/cuda/8.0/lib64/libcudart_static.a

$ git rev-parse HEAD
[...]

PR #9482, opened 2017-04-27
Fix PATH error with non-system GCC (#7543).

kmhofmann:
I have been getting errors from `gcc` not finding `as`, due to my user-set PATH not being respected. See issue #7543.
For example, the following setting on Ubuntu 17.04 failed:

    export PATH=$HOME/compilers/gcc540/install/bin:$PATH
    export LD_LIBRARY_PATH=$HOME/compilers/gcc540/install/lib64:$LD_LIBRARY_PATH
    ./configure
        [...]
    bazel build --config=opt --config=cuda //tensorflow/tools/pip_package:build_pip_package

A fix seems to be rather simple. By just adding `:$PATH` to the respective command, TensorFlow now builds correctly. See diff.
However, I did not try to understand the full Python script, so I can't judge potential side effects.

### DIFF
--- a/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc.tpl
+++ b/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc.tpl
@@ -228,7 +228,7 @@ def InvokeNvcc(argv, log=False):
 
   # TODO(zhengxq): for some reason, 'gcc' needs this help to find 'as'.
   # Need to investigate and fix.
-  cmd = 'PATH=' + PREFIX_DIR + ' ' + cmd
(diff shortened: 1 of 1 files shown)


vrv:
@caisq do you know who might know more about whether this is the right thing to do?  I no longer understand our complicated build environments :)

caisq:
@vrv This particular file has a lot of contribution from @keveman and @zheng-xq.

But this change is simple enough that it makes sense to me: it just ensures that whatever is already in `PATH` stays in `PATH` when `cmd` is invoked, which is necessary, e.g., if your `gcc` is not the system default. This is good practice in general.

caisq (approved):

caisq:

PR #9497, opened 2017-04-27
small change to allow multiple wrappings with EmbeddingWrapper

JKurland:
Hi,
I tried wrapping an LSTM in an attention wrapper followed by an embedding wrapper, however I got an error here as state[0] is still a tuple so i wrote this small change to fix the problem

### DIFF
--- a/tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
+++ b/tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
@@ -866,11 +866,11 @@ class EmbeddingWrapper(RNNCell):
           # Default initializer for embeddings should have variance=1.
           sqrt3 = math.sqrt(3)  # Uniform(-sqrt(3), sqrt(3)) has variance=1.
           initializer = init_ops.random_uniform_initializer(-sqrt3, sqrt3)
-
-        if type(state) is tuple:
-          data_type = state[0].dtype
-        else:
-          data_type = state.dtype
+          
+        s = state
+        while type(s) is tuple:
+          s = s[0]
+        data_type = s.dtype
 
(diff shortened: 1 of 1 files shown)


JKurland:
signed it

tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
```diff
@@ -866,11 +866,11 @@ def __call__(self, inputs, state, scope=None):
           # Default initializer for embeddings should have variance=1.
           sqrt3 = math.sqrt(3)  # Uniform(-sqrt(3), sqrt(3)) has variance=1.
           initializer = init_ops.random_uniform_initializer(-sqrt3, sqrt3)
-
-        if type(state) is tuple:
-          data_type = state[0].dtype
-        else:
-          data_type = state.dtype
+          
+        s = state
```
ebrevdo: can you try using:  nest.flatten(state)[0].dtype

?

JKurland: that would work too, nest.flatten(state)[0] would probably do more work than this for loop but it takes fewer lines so I'm not too fussed

JKurland: because of how wrappers tend to add their states at the end it might also be slightly faster to use s = s[-1] than s = s[0] but i don't think the state object is ever going to be deep enough for this to matter

rmlarsen:
@ebrevdo can you approve or request further changes?

gunan:
this PR is to a release branch.
We do not accept contributions directly into release branches.
Please recreate the PR against TF master branch.

PR #9504, opened 2017-04-28
README.md: make the settings.gradle and build.gradle example code match

johansen:
As currently written, the example code that README.md recommends adding to settings.gradle and build.gradle doesn't match. Adding the code as-is results in a Gradle error.

This change fixes the discrepancy, making it easier to get started using the TensorFlow Android interface.


vrv:
(If you can sign the CLA, please feel free to ping this change and/or reopen a new one with the signed CLA -- we can't take a look at this without it).

PR #9514, opened 2017-04-28
Dynamic ksize and strides with MaxPool and AvgPool

### LINKED ISSUE #4746: ksize in max_pool
I am trying to build a convolution (followed by max_pool) for variable length input size. Since the length of the input in a batch should be the same, I set the batch size to 1. However, for some reason ksize in tf.nn.max_pool is a list attribute required at run-time. This would prevent any neural network that uses variable input size to apply max pool. Is there a workaround for this? Should ksize be an input tensor otherwise?

yongtang:
This fix tries to fix the issue raised in #4746 where ksize and strides is static (attr) with max_pool (and avg_pool).

This fix changes ksize and strides to input tensor with `MaxPoolV2` and `AvgPoolV2` so that it is dynamic now. This fix also deprecates `MaxPool` and `AvgPool`.

This fix fixes #4746.

### DIFF
--- a/tensorflow/core/kernels/avgpooling_op.cc
+++ b/tensorflow/core/kernels/avgpooling_op.cc
@@ -48,9 +48,9 @@ typedef Eigen::ThreadPoolDevice CPUDevice;
 typedef Eigen::GpuDevice GPUDevice;
 
 template <typename Device, typename T>
-class AvgPoolingOp : public UnaryOp<T> {
+class AvgPoolingOp : public OpKernel {
  public:
-  explicit AvgPoolingOp(OpKernelConstruction* context) : UnaryOp<T>(context) {
+  explicit AvgPoolingOp(OpKernelConstruction* context) : OpKernel(context) {
     string data_format;
     OP_REQUIRES_OK(context, context->GetAttr("data_format", &data_format));
     OP_REQUIRES(context, FormatFromString(data_format, &data_format_),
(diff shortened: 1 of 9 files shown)


josh11b:
This is fine for API review, but we'd like the Python changes to be separated into a follow-on change after 3 weeks, where you also bump the GraphDef version in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/public/version.h . This is to accommodate our 3-week limited forward compatibility goal mentioned here:
https://www.tensorflow.org/programmers_guide/data_versions#evolving_graphdef_versions and to support eventually deleting the first versions of these ops.

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -1452,17 +1452,6 @@ def testOpEdgeCases(self):
       if test.is_gpu_available():
         pool_funcs.append(nn_ops.max_pool_with_argmax)
       for pool_func in pool_funcs:
-        # Illegal strides.
-        with self.assertRaisesRegexp(
```
yzhwang: Is there any specific reason to exclude this test?

tensorflow/python/ops/nn_grad.py
```diff
@@ -491,6 +491,19 @@ def _AvgPoolGrad(op, grad):
       data_format=op.get_attr("data_format"))
 
 
+@ops.RegisterGradient("AvgPoolV2")
+def _AvgPoolGradV2(op, grad):
```
yzhwang: We should add tests that actually take in ksizes and strides as input tensors during session running, but not like the current tests where ksizes and strides are still provided when building the graph? So that we can test whether the change works. For example, we could define ksizes and strides as variables and feed values only when we run the session.

This in general holds for all the operators you changed: MaxPool, MaxPoolGrad, MaxPoolGradGrad, AvgPool, AvgPoolGrad, AvgPoolGradGrad.

yzhwang (commented):
In general a nice and clean change. Thanks! I've put some comments in your 3 commits. The major issue is to make sure that current tests in pooling_ops_test.py cover your new changes and pass.

rmlarsen:
@yongtang any update on this? Could you please rebase to resolve conflicts?

tensorflow/core/kernels/avgpooling_op.cc
```diff
@@ -257,12 +295,36 @@ class AvgPoolingGradOp : public OpKernel {
     OP_REQUIRES_OK(context, context->allocate_output(0, output_shape, &output));
     output->flat<T>().setZero();
 
-    const int window_rows = ksize_[1];
-    const int window_cols = ksize_[2];
-    const int depth_window = ksize_[3];
+    std::vector<int32> ksize = ksize_;
+    std::vector<int32> stride = stride_;
+    if (context->num_inputs() != 2) {
```
yongtang: Thanks @yzhwang. The PR has been updated with all related issue fixed.

tensorflow/python/ops/nn_grad.py
yongtang: Thanks @yzhwang The PR has been updated with tests using placeholder now. Please take a look.

yongtang:
@rmlarsen @yzhwang The PR has been rebased and updated. Please take a look.

rmlarsen:

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -103,17 +103,18 @@ def _VerifyOneType(self, pool_func, input_sizes, ksize, strides, padding,
         t = test_util.NHWCToNCHW(t)
         ksize = test_util.NHWCToNCHW(ksize)
         strides = test_util.NHWCToNCHW(strides)
+      k = array_ops.placeholder(dtypes.int32, shape=[4])
```
rmlarsen: Could you test with both placeholders and and numpy arrays?

rmlarsen:
@yongtang There are a number of test failures due to the changes, please address those.

martinwicke:
I'll close this, as it appears inactive. Please create a new PR if you would like to pick this back up.

hgaiser:
@yongtang is this still being worked on? It appears to be exactly what I need.

yongtang:
@hgaiser A new PR #11875 has been created to try to address the Jenkins test failures.

Issue #7664, opened 2017-02-19
tf.nn has no attribute rnn_cell in version 1.0.1


xiatian122:
### Environment info
Operating System: Ubuntu 16.04

Installed version of CUDA and cuDNN: 
(please attach the output of `ls -l /path/to/cuda/lib/libcud*`):

If installed from binary pip package, provide:

1.0.0

### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)

`<ipython-input-66-e322aca5289e> in makeGRUCells()
      7 
      8         def makeGRUCells():
----> 9             base_cell = tf.nn.rnn_cell.GRUCell(num_units=RNN_HIDDEN_SIZE,)
     10             layered_cell = tf.nn.rnn_cell.MultiRNNCell([base_cell] * NUM_LAYERS,state_is_tuple=False)
     11             attn_cell =tf.contrib.rnn.AttentionCellWrapper(cell=layered_cell,attn_length=ATTN_LENGTH,state_is_tuple=False)

AttributeError: 'module' object has no attribute 'rnn_cell'`

### Logs or other output that would be helpful
(If logs are large, please upload as attachment or provide link).

standy66:
It was moved to contrib. Take a look [here](https://www.tensorflow.org/api_docs/python/tf/contrib/rnn)

gaika:
The change is not mentioned in release notes, conversion script doesn't notice it too.

aselle:
@gaika, sorry about that. Unfortunately it appears a few things slipped through. I'll update the conversion script. @yifeif, can you update the release notes, and @ebrevdo are there any rnn things that need to be added to the release notes?

aselle:
##TODO

- [ ] fix converter script (aselle)
- [ ] fix release notes (yifeif)
- [ ] document anything else that is needed on release notes (ebrevdo)

gaika:
Why did you close it? A lot of people will run into this issue when migrating to 1.0.

ebrevdo:
Andrew, it does indeed seem that the move of many of the ops and cells to
tf.contrib.rnn was not mentioned in release.md.  mea culpa; i was hoping to
get them all back into core by 1.0.  Unfortunately looks like it won't
happen until TF 1.0.

So in the meantime, we should update the release notes to say that
tf.nn.rnn_cell.* and most functions in tf.nn.rnn.* (with the exception of
[...]

PR #9523, opened 2017-04-28
Indicate that tf.contrib.rnn functions will be moved back to core for TF 1.2, not 1.1

nealwu:
See https://github.com/tensorflow/tensorflow/issues/7664


yifeif (approved):
Thanks!

PR #7442, opened 2017-02-11
Add a bit more descriptions on adding an op?

zasdfgbnm:
Hi,

I'm trying to add my custom op and feeling that the document doesn't give enough information on how it works. I struggled a lot and finally I'm able to figure things out by reading the code of some header files.

Can anyone take a look if the changes to the document in this PR will make it better for new people?

PR #9529, opened 2017-04-29
Unveil type check part in _VerifyGeneratedGradients that was not actu…

YuMS:
…ally used

### DIFF
--- a/tensorflow/python/ops/gradients_impl.py
+++ b/tensorflow/python/ops/gradients_impl.py
@@ -273,28 +273,28 @@ def _VerifyGeneratedGradients(grads, op):
   if len(grads) != len(op.inputs):
     raise ValueError("Num gradients %d generated for op %s do not match num "
                      "inputs %d" % (len(grads), op.node_def, len(op.inputs)))
-    for i in xrange(len(grads)):
-      grad = grads[i]
-      inp = op.inputs[i]
-      if grad is None:
-        continue
-      if grad.dtype.is_floating:
-        if not inp.dtype.is_floating:
-          raise TypeError("Gradient type %s generated for real-valued op %s "
-                           "with type %s must be real" %
-                           (dtypes.as_dtype(grad.dtype).name, op.node_def,
-                            dtypes.as_dtype(inp.dtype).name))
-      elif grad.dtype.is_complex:
-        if not inp.dtype.is_complex:
-          raise TypeError("Gradient type %s generated for complex-valued op %s"
-                           " with type %s must be complex" %
-                           (dtypes.as_dtype(grad.dtype).name, op.node_def,
-                            dtypes.as_dtype(inp.dtype).name))
-      else:
-        raise TypeError("Gradient type %s generated for op %s "
-                         "with type %s must be either real or complex" %
-                         (dtypes.as_dtype(grad.dtype).name, op.node_def,
-                          dtypes.as_dtype(inp.dtype).name))
+  for i in xrange(len(grads)):
+    grad = grads[i]
+    inp = op.inputs[i]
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/gradients_impl.py
```diff
@@ -273,28 +273,28 @@ def _VerifyGeneratedGradients(grads, op):
...
-          raise TypeError("Gradient type %s generated for complex-valued op %s"
-                           " with type %s must be complex" %
-                           (dtypes.as_dtype(grad.dtype).name, op.node_def,
-                            dtypes.as_dtype(inp.dtype).name))
-      else:
-        raise TypeError("Gradient type %s generated for op %s "
-                         "with type %s must be either real or complex" %
-                         (dtypes.as_dtype(grad.dtype).name, op.node_def,
-                          dtypes.as_dtype(inp.dtype).name))
+  for i in xrange(len(grads)):
```
vrv: Whoa, this was bad, thank you for finding this.

The fact that no tests failed means that there's no test that exercises this.  Can you conceive of a test in the gradients_test.py file that would have exercised this code?  Thank you!

vrv:
cc @yaroslavvb in case he has an idea of how he wanted this code to be checked in the first place

yaroslavvb:
rats! Python indendation!

I added the check because @girving asked me to add a check that gradients must be complex when inputs are complex in https://github.com/tensorflow/tensorflow/pull/6868

But, I'm not sure of a way to exercise complex gradients, @girving any idea? Since nobody complained, maybe we can just drop that check completely?

The original reason for PR is that it enforced that activation value and backprop value have the same type, but we wanted to mix types (float32, float16). What if someone decides to mix real/complex types in the future?

tensorflow/python/ops/gradients_impl.py
YuMS: I'd like to, when your final call is to keep these line.

girving:
It's hard for me to imagine a use of replacing real with complex or vice versa during backprop.  You'd be throwing away half the information, or adding zeros for some reason.

Exercising complex gradients is pretty easy: just make a complex constant and square it, for instance.

yaroslavvb:
I think something extending "synthetic-gradients" idea could mix complex real for gradient -- they use a different function for forward and backward, so you throw away half the information, but gain some regularization/efficiency.

I don't have any strong opinion for what to do here, I think the number of people doing complex gradients is pretty small

vrv:
Let's just get rid of the check -- we can add it back later, it's clear no one was depending on it.

yaroslavvb:
@YuMS could you just delete that whole block? Thanks for the find

vrv:
@yaroslavvb GitHub now allows you to make edits like these in their PR (click the pencil button in the review tab) -- if you think this is the way to go.

yaroslavvb:
Yes, that's super convenient, although last time I tried it, it broke the CLA bot, lets see what happens this time

yaroslavvb:
OK, looks like CLA bot is still confused by github inline edits, same as in https://github.com/tensorflow/tensorflow/pull/7442#issuecomment-279272947 , it needs someone to manually add "cla: yes" label @martinwicke

vrv:
It's possible that it doesn't like it when the person making the edit isn't an admin (it allows me to do it all the time without complaining...).  IN any case, this is fine, so I'll test and merge.

martinwicke:
It just doesn't like it period. It's fine to ignore the CLAbot if it
complains about the consent thing and all the commits are from the
submitter or maintainers.

vrv:
I've been making edits all week and it never complained :)

PR #9531, opened 2017-04-29
Avoid extra TensorReference allocation in gRPC

byronyi:
The original TODO comments pretty much speak for themselves.

### DIFF
--- a/tensorflow/core/distributed_runtime/rpc/grpc_tensor_coding.cc
+++ b/tensorflow/core/distributed_runtime/rpc/grpc_tensor_coding.cc
@@ -16,6 +16,7 @@ limitations under the License.
 #include "tensorflow/core/distributed_runtime/rpc/grpc_tensor_coding.h"
 #include "grpc++/support/byte_buffer.h"
 #include "grpc++/support/slice.h"
+#include "tensorflow/core/common_runtime/dma_helper.h"
 #include "tensorflow/core/framework/tensor.h"
 #include "tensorflow/core/framework/tensor_reference.h"
 #include "tensorflow/core/lib/gtl/inlined_vector.h"
@@ -27,10 +28,9 @@ namespace tensorflow {
 namespace grpc {
 
 static void do_nothing(void* raw) {}
-static void unref_tensorreference(void* raw) {
-  TensorReference* ref = static_cast<TensorReference*>(raw);
-  ref->Unref();
-  delete ref;
+static void unref_tensorbuffer(void* raw) {
+  TensorBuffer* buf = static_cast<TensorBuffer*>(raw);
+  buf->Unref();
 }
 
 void EncodeRecvTensorResponseToByteBuffer(const RecvTensorResponse& proto,
@@ -219,8 +219,8 @@ void EncodeTensorToByteBuffer(bool is_dead, const Tensor& val,
 
     if (tensor_data_is_large) {
       // Encode the actual tensor data by pointing to the backing store,
-      // and add a special zero-length slice that is really a TensorReference
-      // object that we will destroy when we are done.
+      // and add a special zero-length slice that is really a TensorBuffer
+      // reference that we will unref when we are done.
       //
       // TODO(jeff): Note that this approach relies on the fact that
       // slices are destroyed in the order in which they are added to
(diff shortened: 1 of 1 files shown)


tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc
```diff
@@ -337,32 +337,25 @@ void GrpcWorker::RecvTensorAsync(CallOptions* opts,
...
+              alloc_attrs.set_on_host(true);
+              Tensor* tmp = new Tensor(src_dev->GetAllocator(alloc_attrs),
+                                       val.dtype(), val.shape());
               CHECK(send_dev_context)
                   << "send dev name: " << src_dev->name()
                   << " gpu_info: " << src_dev->tensorflow_gpu_device_info();
-              // "val" is on a GPU. Uses GPUUtil to fill the response proto.
-              StatusCallback response_ready = [response, done,
+              // "val" is on a GPU. Uses GPUUtil to fill the tensor.
+              StatusCallback response_ready = [is_dead, response, done,
```
jhseu: We don't need the is_dead capture here, right?

tensorflow/core/distributed_runtime/rpc/grpc_tensor_coding.cc
```diff
@@ -241,17 +241,13 @@ void EncodeTensorToByteBuffer(bool is_dead, const Tensor& val,
 
       // (E) Encode tensor data, but by sharing backing store
 
-      // TODO(jeff,sanjay): It'd be nice to avoid this TensorReference
-      // allocation, and instead get our hands on the underlying
-      // TensorBuffer object and just directly ref it here and unref
-      // it in unref_tensorreference.
-      TensorReference* ref = new TensorReference(val);
+      TensorBuffer* buf = DMAHelper::buffer(&val);
```
jhseu: Shouldn't we `->Ref()` the TensorBuffer here?

tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc
```diff
@@ -337,32 +337,25 @@ void GrpcWorker::RecvTensorAsync(CallOptions* opts,
             if (src_dev->tensorflow_gpu_device_info() && (!on_host)) {
 #if GOOGLE_CUDA
               const DeviceContext* send_dev_context = send_args.device_context;
-              RecvTensorResponse* tmp = new RecvTensorResponse;
-              tmp->set_is_dead(is_dead);
+              AllocatorAttributes alloc_attrs;
```
jhseu: We don't have any benchmarks to measure the effect of this particular change. Do you want to take a stab at modifying rpcbench_test.cc to (optionally) allocate tensors on GPU? If not, I'll find time later on.

It's hard to know what impact this will have unless we measure it :)

jhseu (changes requested):

tensorflow/core/distributed_runtime/rpc/grpc_tensor_coding.cc
jhseu: Also, seems unlikely this will compile without `-fpermissive` because of the const conversions. Make buf const and do the const_cast below?

jhseu:
By the way, I measured the first change, and it improved RPC performance by 2%, so yay! That's actually pretty significant for us.

jhseu:
Jenkins, test this please

tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc
byronyi: That's true. Fixed now.

tensorflow/core/distributed_runtime/rpc/grpc_tensor_coding.cc
byronyi: Sorry there should be a `buf->Ref()`. It's fixed now.

`DMAHelper::buffer(&val)` does return a `const TensorBuffer*` so I will just make `buf` that type. It should be fine.

tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc
byronyi: This should be interesting and I will take some time to look into the benchmarks. Thanks!

jhseu: Ok, thanks! To run the benchmark:
1. Run `./configure` and enable CUDA.
2. `bazel build -c opt tensorflow/core/distributed_runtime:rpcbench_test`
3. `bazel-bin/tensorflow/core/distributed_runtime/rpcbench_test --benchmarks=all > /tmp/benchmark_output`

Read and compare the before and after files for benchmark output.

byronyi: Should `--config cuda` be set when building test?

jhseu: Ah, yes, it should be.

jhseu (approved):

jhseu:
Jenkins, test this please

byronyi:
Can we merge this now? Thanks!

jhseu:
Yeah, @vrv is handling merges this week and will do that when he gets the chance. Thanks!

PR #9535, opened 2017-04-29
Update address in Readme

EddieOne:
side note, git > svn


EddieOne:
I signed it! :taco:

vrv:
Thanks, I'll accept this, though we don't plan on releasing new binaries or website for r0.7, since it is over a year old.

PR #9537, opened 2017-04-29
Cifar10 classification with Resnet

kovalera:
This tutorial is a basic one regarding classification of a more advanced kind than the MNIST data, that includes some basic information preprocessing, data storage (hdf5) and a simplified Residual Network model that is easily understood for entry into the field of Deep Learning.


vrv:
Thanks for writing this tutorial, the community I'm sure will find it useful?

We've been trying to avoid having tutorials checked into our code (it's another thing for our team to maintain), so we've been encouraging people to host their tutorials on their own GitHub accounts, and we have pages like https://www.tensorflow.org/community/welcome where we can link content that becomes popular.  Thanks!

Issue #8744, opened 2017-03-27
DC-ASGD(Delay Compensated Asynchronous Stochastic Gradient Descent)?


liufei1656:
DC-ASGD is Microsoft's very useful algorithm for distributed asynchronous training. Compared with the ordinary ASGD algorithm, DC-ASGD has no significant loss in speed, but can get almost the same effect as Sequential SGD. As far as I know, other mainstream deep learning open source tools have implemented this algorithm, such as: CNTK, Mxnet, Paddle and so on. But in Tensorflow I have not found similar modules.

### What related GitHub issues or StackOverflow threads have you found by searching the web for your problem?
[https://github.com/Microsoft/CNTK/issues/1295](https://github.com/Microsoft/CNTK/issues/1295)
[https://github.com/PaddlePaddle/Paddle/issues/185](https://github.com/PaddlePaddle/Paddle/issues/185)
[https://github.com/dmlc/mxnet/pull/3614](https://github.com/dmlc/mxnet/pull/3614)

### What other attempted solutions have you tried?
I tried to implement this algorithm in Tensorflow by myself.  I do not have enough ability to do this now.

### The link address of the paper
[Asynchronous Stochastic Gradient Descent with Delay Compensation for Distributed Deep Learning](https://arxiv.org/abs/1609.08326)

renyi533:
I am also very interested with this. Any update?

alisidd:
I can do this

alisidd:
@aselle Tensorflow's documentation on distributed computing leads me to believe that the end user is responsible for assigning workers and parameter servers to different devices. Should I then only implement Gradient Descent with Delay Compensation, which the user will have to implement asynchronously?

alisidd:
@aselle Tensorflow's documentation on distributed computing leads me to believe that the end user is responsible for assigning workers and parameter servers to different devices. Should I then only implement Gradient Descent with Delay Compensation, which the user will have to implement asynchronously?

aselle:
@mrry, might be able to give you more direction in this vein (also @tfboyd )

mrry:
I haven't read the paper in depth, but from a quick skim of the paper you could implement the update rule as a `tf.train.Optimizer` subclass, and the asynchronous execution would be provided by whatever training loop the user used. (It would presumably "work" in a synchronous case as well, because *w<sub>t</sub>* &ndash; *w<sub>bak</sub>* would be zero, and hence it would devolve into classic SGD.)

[...]

Issue #7443, opened 2017-02-11
Add Continuous Integration + Review Requirements to CONTRIBUTING.md


ahundt:
Could some basic instructions and suggestions be added to [CONTRIBUTING.md](
https://github.com/tensorflow/tensorflow/blob/master/CONTRIBUTING.md)?

Example notes to add:
 - correctly installing and configuring a compliant python/C++ linter
 - links and explanation of other guidelines that must be followed
 - how to run tests on your own machine

yaroslavvb:
I agree that it would be useful to community. I think perhaps someone who has recently gone through the process might be motivated to put the important things together into a nice guide.

Some snippets from my development notes:

making a PR: https://gist.github.com/yaroslavvb/bdb0a92f2a516ebbe2d386c48f5e2c45

pylint:

```
wget -O /tmp/pylintrc https://raw.githubusercontent.com/tensorflow/tensorflow/master/tensorflow/tools/ci_build/pylintrc
pylint tensorflow/get_string_ops.py --rcfile=/tmp/pylintrc
```

Running tests:

You can see requirements in tensorflow/tools/ci_build/install/install_pip_packages.sh

Most recently this meant doing this

```
pip install -I --upgrade setuptools
pip install portpicker
pip install mock
pip install pep8
pip install pylint
pip install py-cpuinfo
pip install pandas==0.18.1
pip install wheel
pip install --upgrade six==1.10.0
pip install --upgrade werkzeug==0.11.10
pip install --upgrade protobuf==3.0.0

```

Then on Ubuntu:
```
export LD_LIBRARY_PATH="/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:$LD_LIBRARY_PATH"

export flags="--config=opt --config=cuda -k"
bazel test $flags //tensorflow/python/...
```

On MacOS: (bazel needs to be explicitly told to pass env vars for some reason)
```
export CUDA_HOME=/usr/local/cuda
export DYLD_LIBRARY_PATH=/usr/local/cuda/lib:/usr/local/cuda/extras/CUPTI/lib
export LD_LIBRARY_PATH=$DYLD_LIBRARY_PATH
export PATH=$DYLD_LIBRARY_PATH:$PATH

export flags="--config=cuda --config=opt -k --action_env PATH --action_env DYLD_LIBRARY_PATH --action_env LD_LIBRARY_PATH -k"
bazel test $flags //tensorflow/python/...
```

For changes to core framework
[...]

PR #9551, opened 2017-04-30
Delay Compensated Asynchronous Stochastic Gradient Descent

just-a-jazz:
This adds the feature requested by #8744


vrv:
Thanks for this change!

- There are very few comments explaining what's going on.  Because this is a variation of a standard algorithm, it probably needs some more comments explaining what the equations are.

- This is going to fail our linter because the lines are too long, indentation is not correct, etc.  See https://github.com/tensorflow/tensorflow/issues/7443#issuecomment-279182613 for what needs to be done!

- There are no tests; we need tests to make sure this is working correctly from a unit-testing perspective.

- People would want proof that this implementation can be used to validate the results of the paper -- have you done any validation yourself?  If so, can you provide some reproduction steps?  For example, having something like in the mxnet pull request linked from the issue will be useful to show: 1) that it is working, 2) that it actually helps.

Thanks again!

vrv:
ping for @alisidd !

alisidd:
@vrv Hi, I'm still working on it, I should hopefully have a working version by tomorrow.

alisidd:
@vrv I've made the requested changes, and I found quite a few problems with my implementations that required a few changes that I wasn't sure if I should be making, like an addition of an optional parameter to a method in the Optimizer class. It allows for multithreaded processing using one optimizer object. Regardless, I'm sure you can see what else I had to change. 

Furthermore, I tested and validated the implementation using a modified version of mnist_softmax.py with 12 asynchronous workers, and a variance parameter of 5. I found there to be an improvement of around 1.5% in test accuracy over asynchronous stochastic gradient descent, similar to improvements found in the research paper.

vrv:
Thanks for making these changes -- though I think changing the public API of optimizers just for this one case suggests that the Optimizer interface is now doing too much.

I'm not too familiar with the optimizer interface personally, but one thought: the idea of delay compensated updates could theoretically apply to any gradient descent algorithm; instead of having this version descend from Optimizer, does it make sense to:

1) Subclass Optimizer to be a DelayCompensatedOptimizer, where you add your specific modifications based on worker_index 

2) Make GradientDescentDC descend from it?

Let me know what you think!

(Also, I'd probably be verbose on the names here: DelayCompensatedGradientDescent; the term "DC" isn't standard enough for it to be abbreviated).

vrv:
Also there aren't any tests that actually exercise more than 1 worker, so it's hard to tell the context in which it's needed.

@jhseu in case he has any thoughts on optimizer interface

alisidd:
@vrv A whole class for adding the ability to address updates by the worker index would have usage if more algorithms were added similar to this one that depended on having that kind of information. I need that information because I need to store a separate var for each worker, and keep them separate throughout the optimization. I don't think most gradient descent algorithms need that kind of information.

vrv:
(I've assigned this the API review label so that the folks who look at API changes can assess whether this is something to accept.)

alisidd:
@vrv I can also copy all relevant code to my new file and override the relevant methods if this is too drastic a change to the Optimizer class.

vrv:
It would certainly avoid an API review if you did this.  In addition, it might also make sense for this to be in contrib/opt like some other optimizers, so that we can evolve the delay compensated optimizers further.  All code in core has API stability guarantees that will far outlast the initial maintainers :)

asimshankar:
In terms of API review, +1 to the last two suggestions:

- Put this optimizer in contrib
- Avoid changing the Optimizer interface and override the relevant methods in the subclass

alisidd:
@vrv I've made the required changes.

tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py
```diff
@@ -0,0 +1,254 @@
+# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
vrv: 2017

tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py
```diff
@@ -0,0 +1,254 @@
...
+    """Processor for Variable."""
+    def update_op_asynchronous(self, optimizer, g, index):
+        if isinstance(g, ops.Tensor):
+            return optimizer._apply_dense(g, self._v, index)
+        else:
+            assert isinstance(g, ops.IndexedSlices), ("Gradient ", g, " is neither a "
+                                                      "tensor nor IndexedSlices.")
+            # pylint: disable=protected-access
+            return optimizer._apply_sparse_duplicate_indices(g, self._v, index)
+
```
vrv: two whitespace lines between every top line function

tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py
```diff
@@ -0,0 +1,254 @@
...
+class _DenseResourceVariableAsynchronousProcessor(optimizer._DenseResourceVariableProcessor):
+    """Processor for dense ResourceVariables."""
```
vrv: two space indent in all python files

see https://github.com/tensorflow/tensorflow/blob/master/CONTRIBUTING.md#python-coding-style in general

tensorflow/python/training/optimizer.py
```diff
@@ -101,6 +101,14 @@ def update_op(self, optimizer, g):
       # pylint: disable=protected-access
       return optimizer._apply_sparse_duplicate_indices(g, self._v)
 
+  def update_op_asynchronous(self, optimizer, g, index):
```
vrv: are the changes in this file still required?

tensorflow/core/ops/training_ops.cc
```diff
@@ -102,6 +102,30 @@ use_locking: If `True`, the subtraction will be protected by a lock;
   otherwise the behavior is undefined, but may exhibit less contention.
 )doc");
 
+REGISTER_OP("ApplyDelayCompensatedGradientDescent")
```
vrv: Sorry for not seeing this before: can you update this to use the "resource type" like in ResourceApplyGradientDescent instead?  We're moving to "Resource variables" which have meaningful type semantics (refs don't).

tensorflow/core/ops/training_ops.cc
```diff
@@ -102,6 +102,30 @@ use_locking: If `True`, the subtraction will be protected by a lock;
...
+    .SetShapeFn(ApplyGradientDescentShapeFn)
+    .Doc(R"doc(
+var -= alpha * (delta + lambda * delta * (var - var_bak))
+Update '*var_bak' by changing it to the new value of 'var'
+
+var: Should be from a Variable().
+alpha: Scaling factor. Must be a scalar.
+delta: The change.
+lambda: The variance paramter.
+var_bak: Same as "var".
```
vrv: Is there a better name that 'var_bak' we can use?  maybe like 'shadow' or something?

tensorflow/core/ops/training_ops.cc
```diff
@@ -102,6 +102,30 @@ use_locking: If `True`, the subtraction will be protected by a lock;
...
+    .Attr("use_locking: bool = false")
+    .SetShapeFn(ApplyGradientDescentShapeFn)
+    .Doc(R"doc(
+var -= alpha * (delta + lambda * delta * (var - var_bak))
+Update '*var_bak' by changing it to the new value of 'var'
+
+var: Should be from a Variable().
+alpha: Scaling factor. Must be a scalar.
+delta: The change.
+lambda: The variance paramter.
```
vrv: parameter

vrv (changes requested):

rmlarsen:

rmlarsen:
@alisidd sorry it looks like you need to rebase again.

alisidd:
@rmlarsen Should be good now

tensorflow/core/ops/training_ops.cc
alisidd: Did you also want me to rename the op to "ResourceApplyDelayCompensatedGradientDescent"?

vrv:
No need to change the name!  @tensorflow-jenkins test this please

alisidd:
#10055 seems to have had the same problem when it was approved, not sure if I'm supposed to do something.

vrv (approved):

vrv:

(it looks like no GPU kernels are registered, but perhaps that can be added as a future PR if someone wants.  it's pretty straightforward to copy the existing code for other optimizers).

fangy:
@alisidd  would you like share your modified version of mnist_softmax.py ? I wanna have a try. thx .

zhuhyc:
@alisidd @vrv 
I add supporting GPU kernels base your code, when I execute a test,it runs on CPU normally, but it occurs an error  on GPU

     python tensorflow/tensorflow/examples/tutorials/mnist/mnist_softmax_dc_asgd.py

the error is:

      **_2017-06-28 10:58:29.296340: E tensorflow/core/common_runtime/executor.cc:641] Executor failed to create kernel. Invalid argument: AttrValue must not have reference type value of float_ref
	 for attr 'tensor_type'
	; NodeDef: w/DelayCompensatedGradientDescent/_23 = _HostRecv[_start_time=0, client_terminated=false, recv_device="/job:localhost/replica:0/task:0/gpu:0", send_device="/job:localhost/replica:0/task:0/gpu:0", send_device_incarnation=1, tensor_name="edge_103_w/DelayCompensatedGradientDescent", tensor_type=DT_FLOAT_REF, _device="/job:localhost/replica:0/task:0/gpu:0"](^w/DelayCompensatedGradientDescent/_22, ^DelayCompensatedGradientDescent/learning_rate, ^gradients/MatMul_grad/tuple/control_dependency_1, ^DelayCompensatedGradientDescent/lambda); Op<name=_HostRecv; signature= -> tensor:tensor_type; attr=tensor_type:type; attr=tensor_name:string; attr=send_device:string; attr=send_device_incarnation:int; attr=recv_device:string; attr=client_terminated:bool,default=false; is_stateful=true>
	 [[Node: w/DelayCompensatedGradientDescent/_23 = _HostRecv[_start_time=0, client_terminated=false, recv_device="/job:localhost/replica:0/task:0/gpu:0", send_device="/job:localhost/replica:0/task:0/gpu:0", send_device_incarnation=1, tensor_name="edge_103_w/DelayCompensatedGradientDescent", tensor_type=DT_FLOAT_REF, _device="/job:localhost/replica:0/task:0/gpu:0"](^w/DelayCompensatedGradientDescent/_22, ^DelayCompensatedGradientDescent/learning_rate, ^gradients/MatMul_grad/tuple/control_dependency_1, ^DelayCompensatedGradientDescent/lambda)]]_**


the test code is here : 
     https://github.com/zhuhyc/tensorflow/blob/dc-asgd/tensorflow/examples/tutorials/mnist/mnist_softmax_dc_asgd.py

the patch is here: 
     https://github.com/zhuhyc/tensorflow/blob/dc-asgd/dc-asgd.patch
can you give me some suggestions to reslove this error?

sesamesub:
I'm trying to learn more dcsgd and external optimizers for tf, I'm unsure how to implement this feature with the tf learn library examples. 

@zhuhyc 
Does your testcode work for CPUs? When I run the softmax example on docker nightly CPU build, I get the following result:

 $ python mnist_softmax_dc_asgd.py

> Traceback (most recent call last):
  File "mnist_softmax_dc_asgd.py", line 84, in <module>
    tf.app.run(main=main, argv=[sys.argv[0]] + unparsed)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "mnist_softmax_dc_asgd.py", line 64, in main
    train_step = optimizer.minimize(cross_entropy,worker_index=0)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py", line 144, in minimize
    name=name, worker_index=worker_index)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py", line 217, in apply_gradients
    worker_index))
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py", line 34, in update_op_asynchronous
    return optimizer._apply_dense(g, self._v, index)
TypeError: _apply_dense() takes exactly 3 arguments (4 given)

>

zhuhyc:
@sesamesub  
    Yes, it works for on CPUs correctly. My tf version is master branch and have applyed dc-asgd.patch
    the patch is here:
        https://github.com/zhuhyc/tensorflow/blob/dc-asgd/dc-asgd.patch
    It looks like you don't apply the patch, please apply it and test again

fangy:
ping to check status

sesamesub:
I had some issues on my machine which I fixed, your code/patch is fine for
CPU tensorflow and works for per your example code.

zhuhyc:
Ping to check status @alisidd

tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py
```diff
@@ -0,0 +1,256 @@
...
+      """
+    if num_workers <= 0:
+      raise ValueError("num_workers must be positive: %s" % num_workers)
+    super(DelayCompensatedGradientDescentOptimizer, self).__init__(
+          use_locking, name)
+    self._learning_rate = learning_rate
+    self._lambda = variance_parameter
+    self._num_workers = num_workers
+
+  def minimize(self, loss, global_step=None, var_list=None,
```
ry: Hello -

ry: Why not use the `minimize` from the parent class? It seems this is copied exactly?

tensorflow/contrib/opt/python/training/delay_compensated_gradient_descent.py
```diff
@@ -0,0 +1,256 @@
...
+          ([str(v) for _, v in grads_and_vars], loss))
+
+    return self.apply_gradients(grads_and_vars, global_step=global_step,
+                                name=name, worker_index=worker_index)
+
+  def apply_gradients(self,
+                      grads_and_vars,
+                      global_step=None,
+                      name=None,
+                      worker_index=None):
```
ry: Also seems to be copied from parent class?

ry: @alisidd question for you above regarding subclassing. 
cc @vrv

vrv: My preference was that delay-compensated-specific bits should not leak out to the parent class, which the original implementation did by adding worker_index to the public APIs.  Where possible, it would be indeed nice to just reuse existing funtionality like in minimize(), but this function relies on extra state that is not available in the base APIs.

MattX:
It looks like this PR was removed from master (commit 47949be1aae1d7c97f75712e0655da96fab90208), with the message "Remove Delay Compensated Asynchronous Stochastic Gradient Descent (DCASGD). The implementation is broken in its current form."

What was wrong with the implementation? The message only references this PR and issue, none of which seem to be raising issues.

vrv:
IIRC, it was lacking an _apply_dense() implementation, there were typos that caused failures when used, etc.

The commit is obviously in the history for somebody to re-use and fix, if they wanted, but they'd need to write more tests and probably refactor it to make it something that the team could accept to TensorFlow.

PR #9555, opened 2017-04-30
Added output information with path to  wrong JPEG file

AsterAI:


vrv:

vrv:
I switched this to raising a RuntimeError instead of killing the program.  If lint passes, this seems okay to me.

PR #9563, opened 2017-05-01
Add a simple BMP decoder and enable it on Android

freedomtan:
Add a simple BMP decoder and enable it on Android so that we can test simple image classification in command line on Android platform. On Android, image decoding is not directly available to NDK C/C++ 
programs. 

### Why bmp?
.bmp can be decoded easily without relying on external libraries.


tensorflow/core/kernels/BUILD
```diff
@@ -4079,6 +4087,7 @@ filegroup(
         "fake_quant_ops.cc",
         "fifo_queue.cc",
         "fused_batch_norm_op.cc",
+        "whole_file_read_ops.cc",
```
vrv: This seems like a separate change, can we send this as a separate PR?

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,66 @@
...
+    OP_REQUIRES(context, TensorShapeUtils::IsScalar(contents.shape()),
+                errors::InvalidArgument("contents must be scalar, got shape ",
+                                        contents.shape().DebugString()));
+
+    // Start decoding image to get shape details
+    const StringPiece input = contents.scalar<string>()();
+
+    // Decode image, allocating tensor once the image size is known
+    Tensor* output = nullptr;
+    OP_REQUIRES(
```
vrv: Given how simple bmp's implementation is, I think we can avoid the need for a bmp_io library; can we just inline the Decode implementation into this OpKernel?  There'll be much fewer files and changes to make if we do that!

tensorflow/examples/label_image/BUILD
```diff
@@ -12,12 +12,32 @@ cc_binary(
...
         "main.cc",
     ],
-    linkopts = ["-lm"],
-    deps = [
-        "//tensorflow/cc:cc_ops",
-        "//tensorflow/core:framework_internal",
-        "//tensorflow/core:tensorflow",
-    ],
+    linkopts = select({
+        "//tensorflow:android": [
```
vrv: @andrewharp for understanding this, if you know

tensorflow/python/ops/image_ops_impl.py
```diff
@@ -1324,26 +1324,32 @@ def decode_image(contents, channels=None, name=None):
       raise ValueError('channels must be in (None, 0, 1, 3)')
     substr = string_ops.substr(contents, 0, 4)
 
+    def _bmp():
```
vrv: You should update the documentation for this function to indicate it also supports BMP in the appropriate places

tensorflow/python/ops/image_ops_impl.py
```diff
@@ -1324,26 +1324,32 @@ def decode_image(contents, channels=None, name=None):
...
+      is_bmp = math_ops.equal(substr, 'BM', name='is_bmp')
+      decode_msg = 'Unable to decode bytes as JPEG, PNG, GIF, or BMP'
+      assert_decode = control_flow_ops.Assert(is_bmp, [decode_msg])
```
vrv: This is unused, you probably want to use a control dependency like in the other functions

vrv (changes requested):
Thanks!

josh11b:
I'm assuming that after @vrv 's suggested changes you won't need to modify tensorflow/core/BUILD. If you back out the tensorflow/core/BUILD changes, this is fine for API review.

tensorflow/core/kernels/BUILD
freedomtan: Thanks for the review. 

Do you mean the whole android part or the "whole_file_read_ops.cc"?
For the whole Android part, yes, I'll make it another PR.
For "whole_file_read_ops.cc", label_image need it.

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: Sound good. Will do it.

tensorflow/core/kernels/BUILD
vrv: I was just thinking the whole_file_read_ops.cc, but if it's related to this PR, it's fine to keep it.

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: done. please check it

tensorflow/python/ops/image_ops_impl.py
freedomtan: done

tensorflow/python/kernel_tests/decode_image_op_test.py
```diff
@@ -33,6 +33,17 @@
 
 class DecodeImageOpTest(test.TestCase):
 
+  def testBmp(self):
+    # Read a real bmp and verify shape
+    path = os.path.join(prefix_path, "bmp", "testdata", "grace_hopper.bmp")
```
vrv: What's the smallest image you can use to validate correctness?

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,125 @@
...
+        context,
+        Decode(input.data(), input.size(),
```
vrv: I was thinking you can do:

Tensor* output = nullptr;
TF_RETURN_IF_ERROR(context->allocate_output(0, TensorShape({height, width, channels}));

Decode(input, output);

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,125 @@
...
+
+  // if height is negative, data layout is top down
+  // otherwise, it's bottom up
+  if (height < 0)
+    top_down = true;
+  uint8* dstdata = allocate_output(width, abs(height), 3);
+  uint8* const bmp_pixels = &img_bytes[header_size];
+  int row_size = (8 * width * 3 + 31) / 32 * 4;
+
+#ifndef NDEBUG
```
vrv: Can you remove this debugging?

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,125 @@
...
+
+  bool top_down = false;
+  uint8* const img_bytes = (uint8 *) srcdata;
+  const int header_size = *(reinterpret_cast<int*>(img_bytes + 10));
+  const int width = *(reinterpret_cast<int*>(img_bytes + 18));
+  const int height = *(reinterpret_cast<int*>(img_bytes + 22));
+
+  // if height is negative, data layout is top down
+  // otherwise, it's bottom up
+  if (height < 0)
```
vrv: Can you use clang-format --style=google to format all of the C++ code here?

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,125 @@
...
+#include "tensorflow/core/framework/types.pb.h"
+#include "tensorflow/core/lib/core/status.h"
+#include "tensorflow/core/platform/logging.h"
+
+namespace tensorflow {
+
+// Decode the contents of a BMP file
+class DecodeBmpOp : public OpKernel {
+ public:
+  explicit DecodeBmpOp(OpKernelConstruction* context) : OpKernel(context) {}
```
vrv: You might want to check that channels = 3 in the constructor -- it looks like your implementation only works for channels = 3 anyway.

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,125 @@
...
+  const int data_size = *(reinterpret_cast<int*>(img_bytes + 34));
+  LOG(INFO) << "header size = " << header_size;
+  LOG(INFO) << "width = " << width;
+  LOG(INFO) << "height = " << height;
+  LOG(INFO) << "data size = " << data_size;
+  LOG(INFO) << "row size = " << row_size;
+  LOG(INFO) << "top down = " << top_down;
+#endif
+
+  for (int i = 0; i < abs(height); i++) {
```
vrv: I should have looked at this earlier on.  this looks like it's going to be super slow -- it's not using any threading, nor vector intrinsics to make this work quickly.  I'm a bit worried about accepting this because then people are going to complain about TensorFlow being slow, and if they are reading bmp files, it'll be because this op is slow.

Without making this competitive with optimized bmp libraries, I am not sure we'd want to accept this into core tensorflow, and it may be nice as an external op library.

What do you think -- do you have time to compare against other bmp libraries and see how fast this is?  If not, would you be okay with making it an external op library?

vrv (changes requested):

tensorflow/examples/label_image/BUILD
```diff
@@ -12,12 +12,32 @@ cc_binary(
...
+            "-lm",
+            "-z defs",
+            "-s",
+            "-Wl,--exclude-libs,ALL",
+        ],
+        "//conditions:default": ["-lm",],
+    }),
+    deps = select({
+        "//tensorflow:android": [
+            "//tensorflow/cc:cc_ops",
```
andrewharp: android_tensorflow_lib should be all you need to depend on. And in fact, we explicitly don't want to depend on any lib that doesn't start with "android_"

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: fair enough. I am not proud of this simple implementation :-) 

I can try to compare with other libraries and try to improve this. Any particular libraries to compare you have in mind?

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: sure

tensorflow/core/kernels/decode_bmp_op.cc
vrv: I am not familiar at all with bmp, so I don't know :(.

Since BMP is a windows format, I would assume Microsoft's decoder would probably be a start ;) https://msdn.microsoft.com/en-us/library/system.windows.media.imaging.bmpbitmapdecoder(v=vs.110).aspx

My general feeling is that this function could be optimized in two related ways:

1) Using multiple threads, mapping each thread to a unique portion of the input.

2) Within each thread, handing it enough work such that it could maybe share some of the indexing computation for many output pixels.

It *might* make sense for you to play around with this on your own first, and to do some benchmark comparisons to other libraries.  At a point when you think this implementation would be somewhat competitive with other libraries, it might be a good candidate for inclusion in core TensorFlow.  How does that sound?

freedomtan: Sounds reasonable. Don't have windows programming environment for many years though

I know SIMD definitely will help, but I don't know if multithreading will help when the code is vectorized with SIMD. I think I can

1. find some good graphics/vision libraries to check and compare bmp decoding speed
2. if simple C/C++ copy like this is too slow, use SIMD intrinsics (ARM NEON, x86 SSE/AVX) and maybe threads to improve memory copy efficiency.

tensorflow/examples/label_image/BUILD
freedomtan: I included "//tensorflow/cc:cc_ops" for image ops (needed by label_image). Dunno how to include image ops into android_tensorflow_lib. Any suggestion?

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: Step 0. I wrote a script to test image load + decode time. It seems bmp decoding is not slower than jpg and png decodings on my desktop machine (a 2014 iMac). Mostly because decoding bmp is simply memory copy without decompression involved. Will check on ARMv8 devices later.

```python
    file_reader = tf.read_file(file_name, 'file_reader')

    if file_name.endswith(".bmp"):
        image_reader = tf.image.decode_bmp(file_reader, name='bmp_reader')
    elif file_name.endswith(".png"):
        image_reader = tf.image.decode_png(file_reader, name='png_reader')
    else:
        image_reader = tf.image.decode_jpeg(file_reader, channels = 3, name='jpeg_reader')

    graph_def = graph.as_graph_def()
    stat = pywrap_tensorflow.NewStatSummarizer(graph_def.SerializeToString())

    with tf.Session() as sess:
        for i in range(0, 100):
            run_metadata = tf.RunMetadata()
            run_options = tf.RunOptions(trace_level=tf.RunOptions.FULL_TRACE)
            result = sess.run(image_reader, {}, run_options, run_metadata)
            step_stats = run_metadata.step_stats
            stat.ProcessStepStatsStr(step_stats.SerializeToString())

    stat.PrintStepStats()
```

tensorflow/examples/label_image/BUILD
andrewharp: Does this bring in the jpeg dependencies too?

Since this is limited to this specific binary it could be ok, but we should probably have a comment here about it since it's a nonstandard dep.

tensorflow/core/kernels/decode_bmp_op.cc
vrv: Awesome, this is probably sufficient to me to move forward -- it's possible it could be made faster, but as long as it isn't a computational bottleneck for real models, it seems okay.  

Do you have the real numbers to share of running this?

tensorflow/examples/label_image/BUILD
freedomtan: Nope, gif/jpeg/png dependencies are not linked in, because decode_{gif/jpeg/png}_op are not included in android_tensorflow_lib.

I'll add a comment.

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: I put some numbers and their descriptions [here](https://github.com/freedomtan/tf-image-decoder-test)

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: done

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: done

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: Modified the code a bit. Please check

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: done

tensorflow/python/kernel_tests/decode_image_op_test.py
freedomtan: replaced this with a small lena.bmp converted from the lena.png in png data

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,104 @@
...
+  void Compute(OpKernelContext* context) override {
+    const Tensor& contents = context->input(0);
+    OP_REQUIRES(context, TensorShapeUtils::IsScalar(contents.shape()),
+                errors::InvalidArgument("contents must be scalar, got shape ",
+                                        contents.shape().DebugString()));
+
+    // Start decoding image to get shape details
+    const StringPiece input = contents.scalar<string>()();
+
+    bool top_down = false;
```
vrv: You can do:

 bool top_down = (height < 0);

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,104 @@
...
+    uint8* const img_bytes = (uint8*)input.data();
```
vrv: we prefer to avoid c-style casts by doing:

static_cast<uint8*>(input.data());

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,104 @@
...
+    const int width = *(reinterpret_cast<int*>(img_bytes + 18));
+    const int height = *(reinterpret_cast<int*>(img_bytes + 22));
+
+    // if height is negative, data layout is top down
+    // otherwise, it's bottom up
+    top_down = (height < 0) ? true : false;
+
+    // Decode image, allocating tensor once the image size is known
+    Tensor* output = nullptr;
+    const ::tensorflow::Status status = context->allocate_output(
```
vrv: I think you can do
TF_RETURN_IF_ERROR(context->allocate_output(...)); 

and get rid of the separate status and OP_REQUIRES_OK commands.

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,104 @@
...
+
+    Decode(bmp_pixels, output->flat<uint8>().data(), width, abs(height),
+           top_down);
+  }
+
+  uint8* Decode(uint8* const input, uint8* const output, const int width,
+                const int height, bool top_down);
+
+ private:
+  int channels;
```
vrv: private member variables should end with _ so: channels_.

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,104 @@
...
+    const int height = *(reinterpret_cast<int*>(img_bytes + 22));
+
+    // if height is negative, data layout is top down
+    // otherwise, it's bottom up
+    top_down = (height < 0) ? true : false;
+
+    // Decode image, allocating tensor once the image size is known
+    Tensor* output = nullptr;
+    const ::tensorflow::Status status = context->allocate_output(
+        0, TensorShape({abs(height), width, 3}), &output);
```
vrv: So channels is never used -- can you explain whether channels is ever something that people are going to pass for this op?  If not, maybe we can get of the channels attribute.

vrv (changes requested):
Thanks for making the changes!

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: Well, I don't think putting `TF_RETURN_IF_ERROR()` here is the right way to do it. This is in `void Compute(OpKernelContext* context)`. As far as I know, `TF_RETURN_IF_ERROR()` will be expanded into something with `return a_status_variable`, so it will be rejected by gcc or clang.  I cannot find a way to use TF_RETURN_IF_ERROR() without more modifications to either this class or `class OpKernel`,  its base class. Any suggestion?

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: Yes, we can get rid of it for now. The channel attribute is there because BMP actually has more different bit-per-pixel formats and other decoders have this attribute (for consistency).

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: done

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: done

tensorflow/core/kernels/decode_bmp_op.cc
freedomtan: Done. Thanks for the review.

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,110 @@
...
+  uint8* Decode(const uint8* input, uint8* const output, const int width,
+                const int height, bool top_down);
+
+ private:
+  int channels_;
+};
+REGISTER_KERNEL_BUILDER(Name("DecodeBmp").Device(DEVICE_CPU), DecodeBmpOp);
+
+uint8* DecodeBmpOp::Decode(const uint8* input, uint8* const output,
+                           const int width, const int height, bool top_down) {
```
vrv: maybe pass 'channels' here and use it in the calculations, so it's clearer from the name what the calculations are doing.

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -0,0 +1,110 @@
...
+  // there may be padding bytes when the width is not a multiple of 4 bytes
+  int row_size = (8 * width * 3 + 31) / 32 * 4;
```
vrv: There are some magic numbers here (8 and 4) that are probably unclear to most readers -- can you add a comment?

vrv (changes requested):
Just a few more comments, otherwise this looks good.

vrv:

tensorflow/core/kernels/decode_bmp_op.cc
```diff
@@ -67,7 +67,7 @@ class DecodeBmpOp : public OpKernel {
     // Decode image, allocating tensor once the image size is known
     Tensor* output = nullptr;
     OP_REQUIRES_OK(context, context->allocate_output(
-        0, TensorShape({abs(height), width, 3}), &output));
+        0, TensorShape({abs(height), width, channels_}), &output));
```
vrv: This won't work if channels_ is 0.

I think my suggestion would be as follows:

If the user provides channels_ != 0, validate that channels_ = bpp / 8 as you do above.

But in the rest of the code, always use bpp / 8 as the channels you use to allocate the tensor and pass into the Decode function.  Does that make sense?

freedomtan: argh, yes, thanks. Forgot to add let channels_ = bpp / 8 when it's 0.

vrv:

vrv:
(Made a few changes to do validation of channels)

vrv:
(Also added use of internal::SubtleMustCopy).

vrv:

tensorflow/core/ops/image_ops.cc
```diff
@@ -558,6 +558,19 @@ contents: 0-D. PNG-encoded image.
 )doc");
 
 // --------------------------------------------------------------------------
+REGISTER_OP("DecodeBmp")
+    .Input("contents: string")
+    .Output("image: uint8")
```
vrv: Two thoughts here:

1) uint8 isn't the only possible output: if the bpp is 32 bits, we'd want to be able to produce uint32 tensors, right?  If so, perhaps we can do:

.Output("image: T")
.Attr("T: {uint8, uint16, uint32} = DT_UINT8")

I'm not sure what the default should be: are most bmp's only bpp == 8 ?

You then would need to templatize the OpKernel by output type, and register different kernels using TypeConstraint on REGISTER_KERNEL; let me know if you'd like some more pointers about how to do this!

tensorflow/core/ops/image_ops.cc
```diff
@@ -558,6 +558,19 @@ contents: 0-D. PNG-encoded image.
...
+    .Attr("channels: int = 3")
```
vrv: For consistency with the other operations, let's default this to 0

tensorflow/core/ops/image_ops.cc
```diff
@@ -558,6 +558,19 @@ contents: 0-D. PNG-encoded image.
...
+    .SetShapeFn(DecodeImageShapeFn)
+    .Doc(R"doc(
+Decode the first frame of a BMP-encoded image to a uint8 tensor.
+
+contents: 0-D.  The BMP-encoded image.
+image: 3-D with shape `[height, width, 3]`. RGB order
```
vrv: Technically this should now be 'channels', in case bpp == 32?

vrv (changes requested):
Sorry, just a few more comments (I'm looking at the API aspects now).

tensorflow/core/ops/image_ops.cc
freedomtan: ` Unrecognized type string 'uint32' from Attr("T: {uint8, uint16, uint32} = DT_UINT8") for Op DecodeBmp`
It seems uint32 is a valid type supported by `tensorflow/core/framework/op_def_builder.cc`. Is int32 acceptable? Or you really want to have uint32.

vrv: Ah, I just realized, because the bpp value sort of implies 8-bits per channel, it probably doesn't make sense to have anything but uint8 for now, so maybe we keep it as is.  Sorry about this.

vrv:

vrv:
Can you run buildifier on the BUILD files so it passes the sanity checker?
https://github.com/bazelbuild/buildtools

vrv:

tensorflow/python/kernel_tests/decode_image_op_test.py
```diff
@@ -33,6 +33,17 @@
 
 class DecodeImageOpTest(test.TestCase):
 
+  def testBmp(self):
+    # Read a real bmp and verify shape
+    path = os.path.join(prefix_path, "bmp", "testdata", "lena.bmp")
```
vrv: NotFoundError: tensorflow/core/lib/bmp/testdata/lena.bmp
	 [[Node: ReadFile = ReadFile[_device="/job:localhost/replica:0/task:0/cpu:0"](ReadFile/filename)]]


It looks like you need to add this file as testdata in the BUILD rule.

vrv:

hopefully this works!

vrv (approved):

vrv:
This looks good, @rmlarsen for merging.

rmlarsen:
Sorry for not merging this earlier. @freedomtan can you please rebase to resolve conflicts?

vrv:
:(.  Sorry @freedomtan.  There was an internal refactoring and I was hoping to get this change in first so that we'd have to do the refactoring, but looks like it didn't happen.

The context of the merge conflict is that all of the decode C++ implementations now are a single kernel (DecodeImage), because it was the case that too many people used decode_jpeg on images that were not jpegs; so now we have a single kernel that can decode all of them.  However, the python code still has separate python functions.  Let me know if you need any help making progress.

girving:
@vrv @freedomtan Actually, please just resolve the merge conflicts in a simple way (they should be just a few lines in BUILD files?) and submit.  I can merge into the large op as a second pass.

freedomtan:
@vrv Thank you for lots of review. Is it possible to still keep bmp as a separate kernel. My original goal was to be able to have a simple image format for Android so that I can run something like label_image on Android without relying on external image libraries. Surely I can merge the bmp decoder into the single kernel if that's not possible.

girving:
@freedomtan Yes, please leave it as a separate kernel.  One of us can do the further merge as a second pass.

freedomtan:
@girving OK. Thanks. Will do.

rmlarsen:

freedomtan:
rebased to resolve conflicts and test failures

rmlarsen:

rmlarsen:
@freedomtan thanks for the contribution!

rmlarsen:
FYI: We have to temporarily revert the changes to tensorflow/core/BUILD to avoid breaking several internal Android builds. 

I will ask somebody from @petewarden's team to take a look at how to re-enable it.

rmlarsen:
@freedomtan Clarification: What I had to revert was the addition of whole_file_reader_op on Android, which I assume was just needed for a one-off test, and should not have been part of the original PR?

freedomtan:
@rmlarsen So what am I supposed to do? The whole_file_reader_op is used by label_image, that's why it has to be added to make label_image work on Android. Surely you can remove whole_file_reader_op and I think label_image will still build without problem. Just label_image won't be able to find the whole_file_reader_op at run-time on Android.

rmlarsen:
Yes, we need to enable some file reader on Android and have label_image use that, but as far as I know whole_file_reader is not recommended. @josh11b @petewarden can you advice on what file reader to use?

rmlarsen:
@freedomtan feel free to open an issue and assign it to petewarden@ for triage.

cwhipkey:
Instead of using a file reader, the example could use a placeholder in the graph instead of the file reader.  Then, pass the bytes of the image into the session.run call as a feed to that placeholder.   Those bytes can be obtained directly from available APIs (file loading or memmapping), or through the tensorflow API Env::Default).

freedomtan:
@cwhipkey Thanks. Will fiddle with that.

freedomtan:
@rmlarsen I created a patch to remove use of whole_file_reader_op, https://github.com/tensorflow/tensorflow/pull/10034. Follows what suggested by @cwhipkey.

Issue #9328, opened 2017-04-20
Error in `python`: free(): invalid pointer


DjangoPeng:
TensorFlow Version: 1.0 and 1.0.1 (whl package)
Kubernetes Version: 1.5.1
Docker Version: 1.12.5
Container OS: ubuntu 14.04

I run tensorflow applications, say mnist_replica.py, in the kubernetes cluster and this error occured everytime when finishing the whole train step.

I attached the log of mnist_replica.py as below:
```
1492616963.574406: Worker 0: training step 5045 done (global step: 9999)
1492616963.583257: Worker 0: training step 5046 done (global step: 10001)
Training ends @ 1492616963.583282
Training elapsed time: 54.435048 s
After 10000 training step(s), validation cross entropy = 1170.07
*** Error in `/usr/bin/python': free(): invalid pointer: 0x0000000001e16c50 ***
```
And everything goes well when I run the mnist_replica.py in a physical machine with RHEL 7.0. I Guess the problem is that your google's environment is different with me. So the whl package that you compiled can not run well in the ubuntu 14.04 container, especially when it comes to the C program calling Python program

drpngx:
It could be wrong even on your own machine, just going undetected. Could you run a version with more instrumentation? Maybe a debug version with asan? Or maybe valgrind on the last step?

/CC @yifeif

DjangoPeng:
It's not only an unique case, but also happened in other ubuntu 14.04 in physical machine with TensorFlow v1.0, but it has different error log`Error in `python': double free or corruption (!prev): 0x00000000016b43c0`.

drpngx:
Can you reproduce with a build from source on a recent version? If so, can you try asan?

DjangoPeng:
@drpngx I pulled the TensorFlow r1.0 branch and built the whl package from source.  Getting the same error and then I would like to try official docker image.

DjangoPeng:
@drpngx I pulled the official docker image tagged as `tensorflow/tensorflow:1.0.1`, and it works well. I do not yet know what's wrong with my self-built image. I attached my [Dockerfile](https://gist.github.com/DjangoPeng/ce00add09714f247fe211c1dcff708aa).

girving:
@DjangoPeng Let us know if you figure out additional information.  My only wild guess is something to do with incompatible versions of Python.

DjangoPeng:
[...]

PR #9573, opened 2017-05-01
[issue#9328]fix invalid pointer when free()

DjangoPeng:
Fix the [issue#9328](https://github.com/tensorflow/tensorflow/issues/9328) of TensorFlow v1.0 images.


DjangoPeng:
@caisq @girving Please check.

caisq:
@DjangoPeng Can you send the PR to r1.0 instead of master? Master (and r1.1 and future branches) are on ubuntu:16.04. Only r1.0 is on ubuntu:14.04, which requires this patch, if I understand correctly.

vrv:
(Closing since this needs to go into r1.0 branch)

PR #9575, opened 2017-05-01
Catch exception on tf_session.TF_NewStatus()

### LINKED ISSUE #8652: AttributeError: 'NoneType' object has no attribute 'TF_NewStatus
In the latest sources the issue: https://github.com/tensorflow/tensorflow/issues/3388 is fixed in the thensorflow/python/client/session.py, But I now receive the following error instead:
```
Exception ignored in: <bound method BaseSession.__del__ of <tensorflow.python.client.session.Session object at 0x7fb2e08cee48>>
  File "$HOME/.local/lib/python3.5/site-packages/tensorflow/python/client/session.py", line 595, in __del__
AttributeError: 'NoneType' object has no attribute 'TF_NewStatus'

```
I compiled the model manually using the latest version sources and using Bazel 0.4.5. The same seems to be true as with the issue 3388: The error is not always popping up and the run seems to have ended normally. However, now and then to above error message is displayed.

sffc:
"try" and "finally" without "except" is not sufficient to catch the exception.  If the exception gets thrown, it can stop output from other threads from being printed, and the exception makes TF appear to be broken.  Fixes #8652

### DIFF
--- a/tensorflow/python/client/session.py
+++ b/tensorflow/python/client/session.py
@@ -594,7 +594,8 @@ class BaseSession(SessionInterface):
       try:
         status = tf_session.TF_NewStatus()
         tf_session.TF_DeleteDeprecatedSession(self._session, status)
-      except:
+      except AttributeError:
+        # 'NoneType' object has no attribute 'TF_NewStatus'
         pass
       finally:
         if status is not None:
--- a/tensorflow/python/client/session.py
+++ b/tensorflow/python/client/session.py
@@ -594,6 +594,8 @@ class BaseSession(SessionInterface):
       try:
         status = tf_session.TF_NewStatus()
         tf_session.TF_DeleteDeprecatedSession(self._session, status)
+      except:
+        pass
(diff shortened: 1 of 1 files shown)


sffc:
I am covered by the Google Corporate CLA.  (ldap: sffc)

sffc:
Alternative: the exception that gets thrown in this case is a *AttributeError*:

```
Exception ignored in: <bound method BaseSession.__del__ of <tensorflow.python.client.session.Session object at 0x112402c50>>
  File "/usr/local/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 587, in __del__
AttributeError: 'NoneType' object has no attribute 'TF_NewStatus'
```

I can change the `except` clause to catch only an AttributeError instead.

vrv:
@mrry can you take a look to figure out what exception type / strategy we would want to employ here?  Thanks!

vrv:

(this seems fine to me)

PR #9577, opened 2017-05-01
Fix Community link in README

pvaneck:
Linking to the rendered community page is more useful compared to the current link which seems to be out-of-date.


vrv:
Thanks!

PR #9584, opened 2017-05-02
Remove comments about an unused argument

tae-jun:
Remove comments about an unused argument in `tf.train.suffle_batch` and `tf.train.batch`. There are commented lines which exactly is:

> Note: if `num_epochs` is not `None`, this function creates local counter
  `epochs`. Use `local_variables_initializer()` to initialize local variables.

But, there is no argument called `num_epochs`. So, it is unnecessary comments! 😄


tae-jun:
I already signed it!

tae-jun:
I signed it!

PR #9592, opened 2017-05-02
Added 10 comments that illustrate each tests in EventAccumulator.

chris-chris:
I added 10 comments for event_accumulator module in Tensorboard.

Let me fix it if you find any misunderstanding toward the test codes.


chris-chris:
@vrv Thank you! :)

Issue #9467, opened 2017-04-26
Link Error for the deprecated/__init__.py


kemaswill:
It seems that the link "look '[here](https://www.tensorflow.org/code/tensorflow/contrib/deprecated/__init__.py)'" can not point to the right page in the [doc of histogram_summary](https://www.tensorflow.org/api_docs/python/tf/contrib/deprecated/histogram_summary), should we just change it to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/deprecated/__init__.py?

Please go to Stack Overflow for help and support:

http://stackoverflow.com/questions/tagged/tensorflow

If you open a GitHub issue, here is our policy:

1. It must be a bug or a feature request.
2. The form below must be filled out.

**Here's why we have that policy**: TensorFlow developers respond to issues. We want to focus on work that benefits the whole community, e.g., fixing bugs and adding features. Support only helps individuals. GitHub also notifies thousands of people when issues are filed. We want them to see you communicating an interesting problem, rather than being redirected to Stack Overflow.

------------------------

### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
- **TensorFlow installed from (source or binary)**:
- **TensorFlow version (use command below)**:
- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**:
- **GPU model and memory**:
- **Exact command to reproduce**:

You can collect some of this information using our environment capture script:

https://github.com/tensorflow/tensorflow/tree/master/tools/tf_env_collect.sh

You can obtain the TensorFlow version with

python -c "import tensorflow as tf; print(tf.GIT_VERSION, tf.VERSION)"

### Describe the problem
Describe the problem clearly here. Be sure to convey here why it's a bug in TensorFlow or a feature request.

### Source code / logs
Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Try to provide a reproducible test case that is the bare minimum necessary to generate the problem.

aselle:
@wolffg, can we add this to the fix it... I don't know if it is covered by another more general fix approach.

souravsingh:
@aselle I would like to work on the issue. How do I start?

kemaswill:
[...]

PR #9598, opened 2017-05-02
Fix a link error in the doc of histogram_summary

junwei-pan:
As mentioned in [issue 9467](https://github.com/tensorflow/tensorflow/issues/9467), the link "look '[here](https://www.tensorflow.org/code/tensorflow/contrib/deprecated/__init__.py)'" can not point to the right page in the [doc of histogram_summary](https://www.tensorflow.org/api_docs/python/tf/contrib/deprecated/histogram_summary), so I change the link to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/deprecated/__init__.py.


kemaswill:
I signed it!

PR #9605, opened 2017-05-03
Add check for duplicate names in import_graph_def

samjabrahams:
Found when working on #8999. Currently, `import_graph_def` naively goes through all the Ops in the GraphDef without asserting that all names are unique. This causes bugs when [it loops through the ops a second time in order to update their input values](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/importer.py#L310-L313), as the input values for one op may be assigned to another op.

Generally, people don't need to worry about unique names, as TensorFlow provides that functionality automatically. However, this will be a useful debugging check to have for tools that modify GraphDefs directly (such as the [`quantize_graph`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/quantization/quantize_graph.py) tool).

### DIFF
--- a/tensorflow/python/framework/importer.py
+++ b/tensorflow/python/framework/importer.py
@@ -275,6 +275,9 @@ def import_graph_def(graph_def, input_map=None, return_elements=None,
 
     # 1. Add operations without their inputs.
     for node in graph_def.node:
+      # Check to see if this op's name matches a previously seen op
+      if node.name in name_to_op:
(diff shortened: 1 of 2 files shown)


vrv:
I think this is good, but want mrry@ to double-check in case there's any sneaky name_scope issues to deal with.

mrry (approved):
Looks good to me! Thanks for contributing this fix.

vrv:

vrv:
Looks like a build failure, so I'll merge since the other scenarios are passing.

PR #9617, opened 2017-05-03
Add input function for training and testing

HuaizhengZhang:
Estimator is decoupled from Scikit Learn interface by moving into separate class SKCompat. Arguments x, y and batch_size are only available in the SKCompat class, Estimator will only accept input_fn

### DIFF
--- a/tensorflow/examples/tutorials/estimators/abalone.py
+++ b/tensorflow/examples/tutorials/estimators/abalone.py
@@ -141,7 +141,7 @@ def main(unused_argv):
     return x, y
   
   # Fit
-  nn.fit(input_fn=get_train_inputs,, steps=5000)
+  nn.fit(input_fn=get_train_inputs, steps=5000)
 
   # Score accuracy
   def get_test_inputs():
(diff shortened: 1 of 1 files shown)


HuaizhengZhang:
I signed it!

tensorflow/examples/tutorials/estimators/abalone.py
```diff
@@ -134,12 +134,22 @@ def main(unused_argv):
...
-
+  
+  def get_train_inputs():
+    x = tf.constant(training_set.data)
+    y = tf.constant(training_set.target)
+    return x, y
+  
   # Fit
-  nn.fit(x=training_set.data, y=training_set.target, steps=5000)
+  nn.fit(input_fn=get_train_inputs,, steps=5000)
```
sandersk: Delete extra comma after "get_train_inputs"

sandersk (commented):
Thanks, looks good! Just see note about extra comma that needs to be deleted.

tensorflow/examples/tutorials/estimators/abalone.py
vrv: @sandersk you can actually edit small changes like this on their behalf (i'll do it in this case)

vrv:

vrv:
Seems like an unrelated failure, so merging.

vrv:
I should have checked but this got pushed to r1.1 -- you may want to send the PR to the master branch if you want this to persist in the future.

sandersk:
Thanks for going ahead and also merging into master, @vrv!

PR #9640, opened 2017-05-03
Java API to get the size of output list operations

kbsriram:
This opens up access to the TF_OperationOutputListLength C API from java, for operations that return output lists.

### DIFF
--- a/tensorflow/java/src/test/java/org/tensorflow/OperationTest.java
+++ b/tensorflow/java/src/test/java/org/tensorflow/OperationTest.java
@@ -1,0 +1,58 @@
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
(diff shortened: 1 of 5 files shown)


tensorflow/java/src/test/java/org/tensorflow/OperationTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+/** Unit tests for {@link org.tensorflow.Operation}. */
+@RunWith(JUnit4.class)
+public class OperationTest {
+
+  @Test
+  public void outputListLength() {
+    try (Graph g = new Graph()) {
+
+      checkSplit(g, "t1", new int[] {0, 1}, 1);
```
asimshankar: It might be better to have a helper function return the integer value and then the `assert` call can be in the test, that way the line number rendered on test failures is more useful and you don't need to use the string names to locate the failure. Something like:

```java
public void outputListLength() {
  assertEquals(1, split(new int[]{0, 1}, 1));
  assertEquals(2, split(new int[]{0, 1}, 2));
  assertEquals(3, split(new int[]{0,1,2}, 3));
}

@Test
public void  outputListLengthFailsOnInvalidName() {
  ...
}

private int split(int[] values, int num_split) {
  try (Graph g = new Graph()) {
    return g.opBuilder("Split", "Split")
      .addInput(TestUtil.constant(g, "split_dim", 0))
      .addInput(TestUtil.constant(g, "values", values)
      .setAttr("num_split", num_split)
      .build()
      .outputListLength("output");
  }
}
```

asimshankar (commented):
Thanks very much for the PR! Much appreciated.
A couple of minor things, otherwise looks great.

kbsriram:
great suggestions @asimshankar - thanks, and I've incorporated them wholesale.

asimshankar (approved):
Thanks for the PR!

vrv:

Issue #9593, opened 2017-05-02
libtensorflow.so: undefined reference to `__cpu_model'


smistad:
When linking with the tensorflow C++ library, the following error occurs: libtensorflow.so: undefined reference to `__cpu_model'

Seems to be related to issue #7223

I manage to work around this error with: ( as done by https://github.com/tensorflow/tensorflow/pull/7241/commits/d7955a66088567ade60213448bbab861de9055cd)
`#ifndef USE_SSE_CRC32C`
in the file tensorflow/core/lib/hash/crc32c_accelerate.cc

Doing this removed the error.

System information:
Ubuntu 16.04 64 bit
Intel i5 CPU
NVIDIA GTX 980 GPU
Built using CMake with BUILD_SHARED_LIB enabled which creates the libtensorflow.so file

skye:
This does seem to be the same problem as #7223. Can you post what compiler you're using to build TensorFlow? I wonder if we need to expand the check in https://github.com/tensorflow/tensorflow/pull/7241/commits/d7955a66088567ade60213448bbab861de9055cd.

smistad:
The compiler used is gcc 5.4:
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

skye:
Looks like gcc 5 has a similar bug to the one referenced in #7223: https://bugs.launchpad.net/ubuntu/+source/gcc-5/+bug/1568899

According to the bug report, a workaround is to "add -lgcc_s -lgcc at the end of the g++ command that links the shared library".

@smistad do you think you could wade into the CMake build and try the workaround? (Even if you don't plumb it all the way through maybe you can at least confirm it works by printing and manually modifying the compiler commands.)

smistad:
Sure, I will try it and if it works I can add an "if GCC version 5" in cmake, add the libraries mentioned to the linker and create a pull request for it.

HFTrader:
I am using clang 4.0.0 and also having the same problem. However as the compiler is not GCC then the line with target_link_libraries() is not activated. Commenting out the IF() block solves the problem.

PR #9642, opened 2017-05-03
Link to gcc_s and gcc if compiler is GCC version 5. Fixes issue #9593

smistad:
Link to gcc_s and gcc if compiler is GCC version 5 to fix issue #9593

### DIFF
--- a/tensorflow/contrib/cmake/tf_shared_lib.cmake
+++ b/tensorflow/contrib/cmake/tf_shared_lib.cmake
@@ -82,6 +82,13 @@ target_link_libraries(tensorflow PRIVATE
     tf_protos_cc
 )
 
+# There is a bug in GCC 5 resulting in undefined reference to a __cpu_model function when
+# linking to the tensorflow library. Adding the following libraries fixes it.
+# See issue on github: https://github.com/tensorflow/tensorflow/issues/9593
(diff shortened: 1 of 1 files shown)


tensorflow/contrib/cmake/tf_shared_lib.cmake
```diff
@@ -76,10 +76,18 @@ add_library(tensorflow SHARED
     ${tensorflow_deffile}
 )
 
+# There is a bug in GCC 5 resulting in undefined reference to a __cpu_model function when
+# linking to the tensorflow library. Adding the following libraries fixes it.
+# See issue on github: https://github.com/tensorflow/tensorflow/issues/9593
+if(CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.0)
+    set(tensorflow_ADDITIONAL_LIBRARIES gcc_s gcc)
```
skye: nit: instead of creating a new tensorflow_ADDITIONAL_LIBRARIES variable, it might be cleaner to add this check below L90 and call target_link_libraries again (according to the cmake docs, it looks like multiple calls append).

smistad: Good idea, commit added

skye (approved):
Thanks!

vrv:

PR #9660, opened 2017-05-04
Fix error of version_info.cc not being generated on windows

smistad:
If tensorflow_BUILD_PYTHON_BINDINGS is disabled in CMake, the Python interpreter is never resolved in CMake, but it is needed in tf_core_framework.cmake for generating the version_info.cc file with the script tensorflow/tools/git/gen_git_source.py. This seems to only be a problem on windows, resulting in a failed build complaining about version_info.cc missing. This is not an issue on linux, probably because of the shebang #!/usr/bin/env python in the script. 

I suggest moving the part in tf_python.cmake that sets up the python interpreter to the CMakeLists.txt as done in this PR so that it is always executed.

### DIFF
--- a/tensorflow/contrib/cmake/CMakeLists.txt
+++ b/tensorflow/contrib/cmake/CMakeLists.txt
@@ -229,6 +229,12 @@ if (tensorflow_ENABLE_GPU)
   endif()
 endif()
 
+# Find python executable
+include(FindPythonInterp)
+if(NOT ${PYTHONINTERP_FOUND})
+    message(FATAL_ERROR "CMake was unable to find a python interpreter.")
+endif()
+
 # Let's get to work!
 include(tf_core_framework.cmake)
 # NOTE: Disabled until issue #3996 is fixed.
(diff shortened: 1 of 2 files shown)


tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -229,6 +229,21 @@ if (tensorflow_ENABLE_GPU)
   endif()
 endif()
 
+#  Resolve the installed version of Python (for Python.h and python).
+include(FindPythonInterp)
```
mrry: Would it be enough to just add this line to `CMakeLists.txt`, and maybe add a test to see if `${PYTHON_EXECUTABLE}` has been set?

(It seems like it would be fine for the for `PYTHON_INCLUDE_DIR` to fail, and still build the non-Python parts of the repository.)

mrry (changes requested):

tensorflow/contrib/cmake/CMakeLists.txt
smistad: You're right, it should be enough. Then we can check if(${PYTHONINTERP_FOUND}) (https://cmake.org/cmake/help/v3.0/module/FindPythonInterp.html). I'll draft up a new commit.

vrv:
@smistad it helps to let us know when the code is ready to review again -- let us know!

smistad:
Ok, the commit is updated and ready for review

vrv:

mrry (approved):
Looks good to me, thanks!

PR #9666, opened 2017-05-04
Allows direct builds from a Visual Studio solution & Enables Debug builds

markslwong:
**Problem1**: When using CMake GUI, it by default generates a Visual Studio solution.  Visual Studio solutions cannot be built within Visual Studio because the `${CMAKE_BUILD_TYPE}` symbol is not defined for multi-configuration environments like Visual Studio.
**Solution1**: Replace the `${CMAKE_BUILD_TYPE}` symbol with `$(Configuration)`.  The latter symbol works on both command line building and building within Visual Studio.

**Problem2**: The build scripts were written for Windows to assume only optimized builds.  There are references to libraries (eg: "_zlibstatic.lib_").  However on a debug build these references have a different name (eg: "_zlibstaticd.lib_").  There are also some warnings as errors that needed to be fixed.  _(A function that asserts and does not return any value.)_  Lastly one of the kernels did not compile because Eigen does not define a double code path on a Debug build.
**Solution2**: Use the CMake 'debug' and 'optimized' keywords to specify different libraries depending on the build.  Adds `return NULL` to functions that have not return values.  Wraps the declaration of kernels that have double precision definitions in pre-processor statements that will prevent it from being build on Windows.


markslwong:
**NOTE**: This is an identical pull request to one earlier.  That has been approved but I was asked to rebase.  GitHub was having trouble with the diffs after a rebase, so I took HEAD revision and manually reapplied all changes again and force pushed.  I though GitHub would still consider it valid for the pull request.  However, it seems I was sorely wrong.

markslwong:
**NOTE**: After reapplying the changes... I have not run a test build on this.  Let me run a test build first.  I cannot run a test build at home because my home machine has 8gb of system memory, which is not enough to successfully run a Visual Studio build.  I will run a build on another machine tonight.

vrv:
ok let us know when this is ready to look at!

markslwong:
Verified the build.  It's ready to be looked at.

vrv:
Okay, great!  Testing this ahead of time before review.

mrry (approved):
Looks good to me, thanks!

rmlarsen:

PR #9667, opened 2017-05-04
verbs: Configure IB/RoCE QPs MTU according to the port MTU

shamoya:
For the verbs RDMA code to work on standard Ethernet networks,
where ports are configured with MTU 1500, need to set the path
MTU to 1K, as the previous 4K value is typical for IB networks.

@junshi15  @llhe @jhseu

### DIFF
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -475,7 +475,11 @@ void RdmaChannel::Connect(const RdmaAddress& remoteAddr) {
     struct ibv_qp_attr attr;
     memset(&attr, 0, sizeof(ibv_qp_attr));
     attr.qp_state = IBV_QPS_RTR;
-    attr.path_mtu = IBV_MTU_4096;
+    struct ibv_port_attr port_attr;
+    CHECK(!ibv_query_port(adapter_->context_, (uint8_t)1, &port_attr))
+            << "Query port failed";
+    // This assumes both QP's ports are configured with the same MTU
+    attr.path_mtu = port_attr.active_mtu;
     attr.dest_qp_num = remoteAddr.qpn;
     attr.rq_psn = remoteAddr.psn;
     attr.max_dest_rd_atomic = 1;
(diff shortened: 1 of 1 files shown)


poxvoculi:
Can you make this a compile time or load time option, instead of hard-coding the MTU value?  I would guess there's no one value that's going to be optimal for everyone.

jhseu:
Up to @poxvoculi, but I'd favor defaulting to a lower MTU for better compatibility and proving that a higher MTU matters for performance before offering a configuration option. @shamoya did some basic tests and it didn't seem to matter.

poxvoculi:
A colleague has a better suggestion: Why not just use the library ibv_port_attr.active_mtu value at runtime, instead of compiling in a fixed value?

shamoya:
Thanks @poxvoculi @jhseu,
You are right of course and it's better to query the port (ibv_qeury_port) attributes for the MTU.
The attention was to create a (very) temporary fix just to get things working in Roce for the time being, as 
I am working on a patch that adds rdmacm connection management support to the verbs code.

This is the best to go, and removes all the manual connection configurations the code is doing today.
It will also save the need to get all configurations from the user in Server creation (device, port, gid index, roce version, mtu, etc..) since all the connection configuration will be set automatically accordingly to the grpc network provided by the cluster spec (user controls the verbs network according the GRPC network interfaces). It means they are binded (grpc and verbs on the same hw interface ) but this is the way to go according the grpc+verbs protocol @junshi15 created.
What do you think ?

Anyway, I will change for selecting the MTU from the port query.

junshi15:
Yes, I think this is a temporary fix for the RoCE folks to be able to use Infiniband.

@shamoya is working on a better version with rdmacm, which is smart enough to set the network parameters automatically and properly. I assume that end users, especially those unfamiliar with hardware, would rather not tweak the network parameters, unless the goal is benchmarking.

shamoya:
@poxvoculi - Changed to take active_mtu from port attribures

vrv:

vrv:
(A spurious failure)

poxvoculi (approved):

martinwicke:
Jenkins, test this please.

bkovalev:
What the PR a status?

bkovalev:
Why it still not merged?

rmlarsen:

rmlarsen:
@bkovalev It seems that the tests failed on the last try. Let's give it another try.

bkovalev:
Thanks

bkovalev:
Hi,
ROCE is working, but with poor performance. In inception v3 arround 1/8 from Ethernet speed.

junshi15:
@bkovalev did you by any chance checked IB as well? It should not affect IB performance. Verbs supports RoCE, not sure why the performance penalty.

PR #9671, opened 2017-05-04
fix symbol count in def file for windows

guschmue:
This fixes some issue introduced in the previous version where RTTI
was removed from the exclusion list. Because of this the number of
symbols in the def file was close to 64K for gpu builds and yesterday
a few added symbols pushed us over the 64K limit for the windows linker, breaking the windows gpu build
Adding RTTI back to the exclusion list.


vrv:

mrry (approved):

PR #9674, opened 2017-05-05
Register variable's proto function with key 'LOCAL_VARIABLES'

austinzh:
When export/input meta graph, user expect all collections
been rebuilded on target graph, but it turns out local variables
on target collection is not Variable but Tensor because local variable
do not have it's proto fn registered


alextp:
Sadly we can't accept this as-is because it breaks the capability of loading any meta graphs which were saved before this change.

Instead for now it's safer to put the registration in your model core.

austinzh:
Well then, Let's change it into a issue, cause it look like one.

Issue #9634, opened 2017-05-03
cuda_configure.bzl makes bad symlink for: cuda/include/cudnn.h --> cuda/include/include/cudnn.h


momeara:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: NO
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: CentOS release 6.7 (Final) x86_64
- **TensorFlow installed from (source or binary)**:  source
- **TensorFlow version (use command below)**: master: git version 550df413158b32645ca5df4dcaabc67f1a48964d (checked out May 3rd 2017)
- **Bazel version (if compiling from source)**: bazel-0.4.5
- **CUDA/cuDNN version**: 7.5/5.1.3
- **GPU model and memory**: Quadro K600 1GB DDR3
- **Exact command to reproduce**:

```
setenv CC '/mnt/nfs/home/momeara/opt/bin/gcc'
setenv CXX '/mnt/nfs/home/momeara/opt/bin/g++'
setenv EXTRA_BAZEL_ARGS '--verbose_failures --jobs=1'
setenv CPLUS_INCLUDE_PATH '/mnt/nfs/home/momeara/opt/include'
setenv C_INCLUDE_PATH '/mnt/nfs/home/momeara/opt/include'
setenv LIBRARY_PATH '/mnt/nfs/home/momeara/opt/lib:/mnt/nfs/home/momeara/opt/lib64'
setenv LD_LIBRARY_PATH /mnt/nfs/work/momeara/sea/DeepSEA/cuda/lib64:/usr/local/cuda-7.5/lib64:/mnt/nfs/home/momeara/opt/lib:/mnt/nfs/home/momeara/opt/lib64:/usr/local/cuda-7.5/extras/CUPTI/lib64:$LD_LIBR\
ARY_PATH
setenv PATH /usr/local/cuda-7.5/bin:/mnt/nfs/work/momeara/sea/DeepSEA/tensorflow/bazel-0.4.5/output:$PATH

./configure
Please specify the location of python. [Default is /mnt/nfs/work/momeara/tools/anaconda2/envs/sea16/bin/python]:
Found possible Python library paths:
  /mnt/nfs/work/momeara/tools/anaconda2/envs/sea16/lib/python2.7/site-packages
Please input the desired Python library path to use.  Default is [/mnt/nfs/work/momeara/tools/anaconda2/envs/sea16/lib/python2.7/site-packages]

[...]

Issue #6845, opened 2017-01-14
input pipeline spends all time in QueueDequeueMany


ericyue:
My training process use tfrecord format for train&eval dataset .
I test the benchmark of reader , only 8000records/second.  and io speed(see from iotop command) just 400KB-500KB/s. 
I'm using the cpp version of protobuf here https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/get_started/os_setup.md#protobuf-library-related-issues

### What related GitHub issues or StackOverflow threads have you found by searching the web for your problem?
https://github.com/tensorflow/tensorflow/issues/4467
https://github.com/tensorflow/tensorflow/issues/4732
### If possible, provide a minimal reproducible example (We usually don't have time to read hundreds of lines of your code)
```
def read_and_decode(filename_queue):
     reader = tf.TFRecordReader()
    _, serialized_example = reader.read(filename_queue)
    return serialized_example
```

```
  serialized_example = read_and_decode(filename_queue)
  batch_serialized_example = tf.train.shuffle_batch(
      [serialized_example],
      batch_size=batch_size,
      num_threads=thread_number,
      capacity=capacity,
      min_after_dequeue=min_after_dequeue)
  features = tf.parse_example(
      batch_serialized_example,
      features={
          "label": tf.FixedLenFeature([], tf.float32),
          "ids": tf.VarLenFeature(tf.int64),
          "values": tf.VarLenFeature(tf.float32),
      })
```

### What other attempted solutions have you tried?

I try to set num_threads in  tf.train.shuffle_batch but not working. It seems that when set to 2 threads, it work at 8000records/s, when enlarge the thread number, it get slower. (I remove all ops that cost cpus. Just read data.) 
My sever are 24 core cpus.

ericyue:
here is my benchmark script and timeline result (timeline.json original file inlcude)
https://gist.github.com/ericyue/7705407a88e643f7ab380c6658f641e8

ericyue:
https://github.com/tensorflow/tensorflow/issues/4732
I read this issue that you guys talk about, seems the same problem. I'am using the latest version of tensorflow , but it's still unsolved.
@gunan  @ebrevdo  @yaroslavvb

ericyue:
![screen shot 2017-01-14 at 11 15 56 pm](https://cloud.githubusercontent.com/assets/918889/21956027/ec463fbe-dab1-11e6-9670-9f75f1b21e6d.png)

[...]

Issue #5907, opened 2016-11-28
FIFOQueue speed problem 


gowithqi:
Hi I am testing the speed of FIFOQueue. I create a queue of Tensor with type of tf.float32 and shape of (576, 3, 220, 220). Then I push a tensor to the queue, followed by pop the tensor from the queue.
The speed of single node version( sess = tf.Session())  and distributed version of tensorflow (even in single machine single process scenario, sess = tf.Session(server.target) ) differs much. 
Pushing and poping take about 0.2s in single node version while 4s in  distributed version.
I know that using distributed version of tensorflow will encounter some proto serialize overhead, but the overhead seems too much.

The main code is like this (I also give a link to full source code below, which can be used to reproduce the problem)
```
    raw_shape = [576, 3, 220, 220]
    shape = tf.TensorShape(raw_shape)
    if FLAGS.cluster:    
        server = tf.train.Server.create_local_server()
        sess = tf.Session(server.target)
    else:
        sess = tf.Session()
    with tf.device('/cpu:0'): 
        q = tf.FIFOQueue(10, tf.float32, shape)
        rand_data = tf.zeros(shape)
        init_op = tf.initialize_local_variables()
        sess.run(init_op)
        result = q.dequeue()
        x = tf.placeholder(tf.float32, shape, 'data')
        enqueue_op = q.enqueue(x)
        while True:
            log('pushing')
            sess.run(enqueue_op, feed_dict = {x: np.zeros(raw_shape)})
            log('push done')
            sess.run(result)
            log('pop done')
```

### What related GitHub issues or StackOverflow threads have you found by searching the web for your problem?

https://github.com/tensorflow/tensorflow/issues/3009 
But this issue is more related to threading, not same as mine case.

### Environment info
Operating System:
Ubuntu 14.04
CPU E5-2643 v3 @ 3.40GHz

Installed version of CUDA and cuDNN: 
None (I used CUDA_VISIBLE_DEVICES='')

If installed from source, provide 
1. The commit hash (`git rev-parse HEAD`)
d6b25985ac219a6e58d186a2beb74d5e8d9e4533
2. The output of `bazel version`
Build label: 0.4.0
Build target: bazel-out/local-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Nov 2 17:54:14 2016 (1478109254)
Build timestamp: 1478109254
Build timestamp as int: 1478109254

[...]

Issue #3009, opened 2016-06-23
FIFOQueue: dequeue many operation very slow?


tomrunia:
When training a relatively simple model (1-layer LSTM, 256 units) my Titan X GPU keeps spiking from 0% to 30% GPU utilization. Conclusion: somewhere in the pipeline there is a bottleneck which limits the GPU to be processing the training batches continuously. I use a FIFOQueue to which examples are being fed in one or more separate threads:

```python
queue = tf.FIFOQueue(
     capacity=self.config.queue_capacity,
     dtypes=[tf.float32, tf.float32],
     shapes=[[30, 49, 512], [30]],
     name="FIFOQueue"
)
```

For the training operation I use `queue.dequeue_many` to get examples from the queue. As you can see the batch size is 64 examples. So in the end the input tensor is `64x30x49x512` of type `tf.float32`:

```python
# Model inputs, either use the queue (training) or feed_dict (evaluation)
inputs, targets = queue.dequeue_many(64)
```

To find out why my code is running "slow" (i.e. spiking GPU allocation and no temperature increase) I use the `Timeline` object ([see here](http://stackoverflow.com/questions/34293714/tensorflow-can-i-measure-the-execution-time-of-individual-operations)) to measure execution times of individual operations. The results displayed below show the measurements for one training iteration at which point the queue was filled with more than 1000 examples. I have included screenshots for both GPU and CPU-only runs (forced with `export CUDA_VISIBLE_DEVICES=-1`. 

What strikes me from these results is that it takes a really long time to dequeue examples from the FIFOQueue. What is happening here...something wrong or is the dequeuing operation just very slow? Overall the dequeuing operation and sending the data to the GPU takes up half of the time of a training iteration. No wonder that the GPU utilization is spiking. Any help is welcome optimizing my training pipeline! As I understand correctly the examples are all queued in RAM, is there also a way to queue them ahead on GPU memory so when they are needed they do not have to be moved CPU => GPU?

This is tested on TensorFlow v9.0 build from sources about 1.5 week ago.

**GPU running on Titan X**
![gpu](http://i.imgur.com/eTZNPDM.png?1) 

**CPU running on Xeon CPU E5-2640**
![cpu](http://i.imgur.com/mQkLH8h.png?1)

prb12:
Thanks for including the timelines!  Nice to see people using that ;-)

[...]

PR #9686, opened 2017-05-05
Map Staging Area

sjperkins:
I liked the idea of StagingArea but I wanted more functionality out of it. I've implemented an Associative StagingArea backed by C++ `std::map` and `std::unordered_map`.

Functionality:
- Keys are int64, Values are just like StagingArea Tensors or list of Tensors.
- put(key, value), get(key),  pop(key), popitem(), size(), clear() operations
- Can choose ordered `std::map` (Get a basic priority queue with popitem()) or unordered `std::unordered`.
- Can specify a capacity which inserts will wait on, if not, it will be unbounded.

Have tested that this works on CPU + GPU in a custom op library. I haven't tested the compile on my laptop due to #9634, but I've cut and pasted the relevant sections into the tensorflow code base for your perusal.

Let me know if you're keen to merge this and I can clean it up a bit further.


ekelsen:
Hi Simon,

Thanks for this nice patch.  We had a couple of thoughts for a re-organization to avoid creating multiple classes with slightly different functionality and naming conventions:

1) Rather than creating new python high level ops, could we slightly modify the existing StagingArea interface by adding an optional key argument to get and put?  If the key is present, then your C++ ops would be used, otherwise the current C++ ops would be used.  The constructor could take an argument that decides on ordering.

2) Your pop method would map to get and if you want to retain the ability to inspect the StagingArea without removing an element, then add a new method 'peek'.

3) You can choose to add fixed capacity to the existing StagingArea or remove it from the new classes (so only unbounded).

Thanks,
Erich

sjperkins:
> Rather than creating new python high level ops, could we slightly modify the existing StagingArea interface by adding an optional key argument to get and put? If the key is present, then your C++ ops would be used, otherwise the current C++ ops would be used. The constructor could take an argument that decides on ordering.

I disagree with this approach because two very different data structures are being exposed and I think it would be a mistake to combine their interface. They offer different [affordances](https://en.wikipedia.org/wiki/Affordance) (see Norman's [The Design of Everyday Things](https://en.wikipedia.org/wiki/The_Design_of_Everyday_Things)) to the user. Thus:

1. Mixing the interfaces will be confusing for the end user. StagingArea behaves like a list while MapStagingArea behaves like a dict. 
2. I actually think it will be more difficult to maintain since the combination of interfaces is confusing.
3. Does the data for the list exist in a different container from the dict?
4. What happens if the user puts without a key and then then tries to get with a key?

I think the following would be better:

1. Keep the separate list (StagingArea) vs map (MapStagingArea) python classes
2. Move common functionality into a Base Class (coloc op, dtype, shape, name, return value handling etc.)
3. Rename `get(key)` to `peek(key)` in the dict. I dislike losing the name of the python dict method, but it probably too late to change `get()` to `pop()` in the list.
4. Add `peek(index)`, `size` and `clear` methods to the list (Already got a size op for the list).

> You can choose to add fixed capacity to the existing StagingArea or remove it from the new classes (so only unbounded).

OK I can do this.

ekelsen:
That works, thanks again.

Re: 3) Make sure to also change pop in the dict to get.

sjperkins:
> Re: 3) Make sure to also change pop in the dict to get.

Just realised it may make more sense `popitem()` changed to `get()`, since `popitem`  currently takes no `key` argument (it returns a random element if unordered or the max element if ordered), while `pop(key)` returns an element based on a key. Do you agree?

I also want to extend the dict methods as follows:

- `peek(key, default=None)`
- `pop(key, default=None)`
- setdefault(key, value, default=None)

where the specified default is returned if no element associated with the key is found.

Also I remember someone (maybe @girving) mentioning somewhere that there's a way of placing an int32 input for a GPU op on the CPU (not through HostMemory). IIRC its not really supported but forcing the int64 key to the CPU is a good case for it. Could someone point me to code that does this?

ekelsen:
Maybe if popitem(key=None) == pop() just popitem -> get and "get" rid of pop ?

I don't think is what you want, since it does use HostMemory: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/split_v_op.cc#L383
but I'm not sure what mechanism @girving was thinking of.

Why not do it this way?  You could force the int64 case only to the CPU.

yaroslavvb:
Hm, are integer types even supported on GPU? There's [this line](https://github.com/tensorflow/tensorflow/blob/a304537954a865752ad1b18461e6bd67b36082db/tensorflow/core/framework/register_types.h#L171) in register_types

```
// Call "m" on all types supported on GPU.
#define TF_CALL_GPU_NUMBER_TYPES(m) \
  TF_CALL_half(m) TF_CALL_float(m) TF_CALL_double(m)

```

sjperkins:
OK I think this is ready for review.

Please take note of https://github.com/tensorflow/tensorflow/pull/9686/commits/92d4d732cab8c6b8db83a38166298ed9a4e8961c. I've made StagingArea.peek(index), MapStagingArea.peek(key) and MapStagingArea.get(key) block until the requested index/key is present in the staging area. This means that bounds checks/key lookups may frequently occur in waiting getters when inserters insert their data and notify the getters.

tensorflow/core/kernels/stage_op.cc
```diff
@@ -30,26 +30,82 @@ namespace {
...
+
+    notify_inserters_if_bounded(l);
+  }
+
+  // Return tuple at index
+  Status Peek(int index, Tuple* tuple) {
+    mutex_lock l(mu_);
+
+    // Wait if the requested index is not available
+    non_empty_cond_var_.wait(l, [&, this]() {
```
ekelsen: [&, this] might redundant

"[this] captures the current object (*this) by reference
[&] captures all automatic variables used in the body of the lambda by reference and current object by reference if exists"

tensorflow/core/kernels/map_stage_op.cc
```diff
@@ -0,0 +1,497 @@
...
+  int capacity_;
+  mutex mu_;
+  condition_variable not_empty_;
+  condition_variable full_;
+  map_type map_ GUARDED_BY(mu_);
+
+public:
+  // public methods
+  explicit StagingMap(int capacity)
+    { capacity_ = capacity; }
```
ekelsen: initialization list

tensorflow/core/kernels/stage_op.cc
```diff
@@ -30,26 +30,82 @@ namespace {
...
+  explicit Buffer(int capacity) {
+    capacity_ = capacity;
+  }
+
+  bool HasBoundedCapacity() {
+    return capacity_ > 0;
+  }
+
+  bool IsFull() {
+    return buf_.size() >= capacity_;
```
ekelsen: If Size grabs the mutex, should IsFull?

tensorflow/core/kernels/map_stage_op.cc
```diff
@@ -0,0 +1,497 @@
...
+#if GOOGLE_CUDA
+REGISTER_KERNEL_BUILDER(Name("MapStage").HostMemory("key")
+                      .Device(DEVICE_GPU), MapStageOp<false>);
+REGISTER_KERNEL_BUILDER(Name("OrderedMapStage").HostMemory("key")
+                      .Device(DEVICE_GPU), MapStageOp<true>);
+#endif
+#ifdef TENSORFLOW_USE_SYCL
+REGISTER_KERNEL_BUILDER(Name("MapStage").HostMemory("key")
+                      .Device(DEVICE_SYCL), MapStageOp<false>);
+REGISTER_KERNEL_BUILDER(Name("MapStage").HostMemory("key")
```
ekelsen: OrderedMapStage

ekelsen (commented):
In general, looks very nice, thank you.

tensorflow/core/ops/data_flow_ops.cc
```diff
@@ -1963,6 +1965,39 @@ dequeue with many fewer capabilities and options.  This Op is optimized for
...
+    .Input("index: int32")
+    .Output("values: dtypes")
+    .Attr("capacity: int = 0")
+    .Attr("dtypes: list(type)")
+    .Attr("container: string = ''")
+    .Attr("shared_name: string = ''")
+    .SetShapeFn(shape_inference::UnknownShape)
+    .SetIsStateful()
+    .Doc(R"doc(
+Op is similar to a lightweight Dequeue.  The basic funtionality is similar to
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/a9a6ccb43f069cda06f43976be55d35ab95a6d09

tensorflow/core/ops/data_flow_ops.cc
```diff
@@ -1963,6 +1965,39 @@ dequeue with many fewer capabilities and options.  This Op is optimized for
...
+
+
+REGISTER_OP("StageSize")
+    .Output("size: int32")
+    .Attr("capacity: int = 0")
+    .Attr("container: string = ''")
+    .Attr("shared_name: string = ''")
+    .SetShapeFn(shape_inference::ScalarShape)
+    .SetIsStateful()
+    .Doc(R"doc(Staging Area Size)doc");
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/a9a6ccb43f069cda06f43976be55d35ab95a6d09

tensorflow/core/ops/data_flow_ops.cc
```diff
@@ -1963,6 +1965,39 @@ dequeue with many fewer capabilities and options.  This Op is optimized for
...
+
+REGISTER_OP("StageClear")
+    .Attr("capacity: int = 0")
+    .Attr("container: string = ''")
+    .Attr("shared_name: string = ''")
+    .SetShapeFn(shape_inference::UnknownShape)
+    .SetIsStateful()
+    .Doc(R"doc(Staging Area Clear)doc");
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/a9a6ccb43f069cda06f43976be55d35ab95a6d09

tensorflow/core/ops/data_flow_ops.cc
```diff
@@ -1986,4 +2021,129 @@ file_parallelism: How many sstables are opened and concurrently iterated over.
...
+REGISTER_OP("MapStage")
+    .Input("key: int64")
+    .Input("values: dtypes")
+    .Attr("capacity: int = 0")
+    .Attr("dtypes: list(type)")
+    .Attr("container: string = ''")
+    .Attr("shared_name: string = ''")
+    .SetShapeFn(tensorflow::shape_inference::NoOutputs)
+    .SetIsStateful()
+    .Doc(R"doc(MapStage)doc");
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/a9a6ccb43f069cda06f43976be55d35ab95a6d09

tensorflow/python/kernel_tests/map_stage_op_test.py
```diff
@@ -0,0 +1,244 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/f3e0ed97f79fbcb8ac1123761d75295230ae177e

tensorflow/python/kernel_tests/map_stage_op_test.py
```diff
@@ -0,0 +1,244 @@
...
+
+      self.assertTrue(sess.run(size) == 10)
+
+  def testSizeAndClear(self):
+    with ops.device('/cpu:0'):
+      x = array_ops.placeholder(dtypes.float32, name='x')
+      pi = array_ops.placeholder(dtypes.int64)
+      gi = array_ops.placeholder(dtypes.int64)
+      v = 2. * (array_ops.zeros([128, 128]) + x)
+    with ops.device(test.gpu_device_name()):
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/0693e69213e560be8b790ab29dd201769adc7b09

tensorflow/python/kernel_tests/map_stage_op_test.py
```diff
@@ -0,0 +1,244 @@
...
+
+    n = 10
+
+    with self.test_session(use_gpu=True) as sess:
+      # Keys 0..n-1
+      keys = list(six.moves.range(n))
+
+      # Shuffle them for random insert
+      shuffle_keys = [i for i in keys]
+      random.shuffle(shuffle_keys)
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/ff0c599ed31977df722808d36531548944e35fb6

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -1584,19 +1642,270 @@ def get(self, name=None):
...
+    """
+    if name is None:
+        name = "%s_size" % self._name
+
+    return gen_data_flow_ops.stage_size(shared_name=self._name,
+                        name=name,
+                        capacity=self._capacity)
+
+  def clear(self, name=None):
+    """Empties the staging area."""
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/1cd58f2c241676ce933a15beb680cfc0f53a795c

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -1584,19 +1642,270 @@ def get(self, name=None):
...
+      name = "%s_clear" % self._name
+
+    return gen_data_flow_ops.stage_clear(shared_name=self._name,
+                        name=name,
+                        capacity=self._capacity)
+
+class MapStagingArea(BaseStagingArea):
+    """
+    Class for staging inputs. Similar to `StagingArea` but behaves
+    like a hashtable with support for puts, gets, pops, popitems,
```
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/ce326f6aa03e4375331a8ddbcec25e0dc97c681b

tensorflow/core/kernels/stage_op.cc
sjperkins: Original intention was a private helper method. Made it private.

https://github.com/tensorflow/tensorflow/pull/9686/commits/5ad80b2d5013cb521dd5fc176852748d8c6f8e89

tensorflow/core/kernels/map_stage_op.cc
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/4f8b0a6a6c27c0005c3b34760606cdb2fa81cdd3 and https://github.com/tensorflow/tensorflow/pull/9686/commits/d97bce1c3d100f4f6070890202f2a427cb04616e

tensorflow/core/kernels/map_stage_op.cc
sjperkins: https://github.com/tensorflow/tensorflow/pull/9686/commits/895be638df52a50de74e6f99dbabb5c866df5b95

tensorflow/core/kernels/stage_op.cc
sjperkins: It needs to capture both `index ` and `this`. I've changed the capture of `index` to capture by value as its just an int.

https://github.com/tensorflow/tensorflow/pull/9686/commits/a90c2e59a37acd2dd54d838261831aab34de3474

sjperkins:
Just had a thought that it might be good to move the C++ StagingArea and MapStagingArea classes in separate header and source files so that they can be reused as primitives in other ops. I'll try do this soon.

sjperkins:
One question. I see that structures like FIFOQueue use PersistentTensors, presumably for reasons suggested [here](https://github.com/tensorflow/tensorflow/blob/554b57f74a99ea18baf616ae5e3cff8b137430e6/tensorflow/core/framework/op_kernel.h#L193-L198):

```c++
// Wraps a tensor that is held by an Op across calls to Compute(). For
// memory safety when using asynchronous devices like GPUs, the system
// must be notified when a Tensor is used inside an Op execution. The
// wrapper ensures that all uses of the Tensor are tracked, because in
// order to retrieve the Tensor the caller must use AccessTensor which
// notifies the context.
class PersistentTensor { ... };
```

Should the tensors in the Staging Area be wrapped in PersistentTensors?

rmlarsen:
@ekelsen could you please take another look and  address @sjperkins's questions?

tensorflow/python/kernel_tests/map_stage_op_test.py
ekelsen: I think that this still has a small but finite probability of ending up with an unsorted list.  The shuffle could undo the reverse.  Just reverse.  Then you guarantee that the list is not sorted.

tensorflow/core/kernels/map_stage_op.cc
```diff
@@ -65,6 +65,8 @@ class StagingMap : public ResourceBase
 private:
   // Private variables
   int capacity_;
+  int memory_limit_;
```
ekelsen: When dealing with memory sizes, it's probably best to use size_t.  A signed 32-bit integer could easily be exceed.

ekelsen:
Regarding using PersistentTensor, I think PersistentTensor is poorly named and that that comment is a bit out-of-date.  Just using Tensors should be fine.  The PersistentTensor existed to deal with synchronization between multiple GPU streams.

ekelsen (approved):
LGTM

rmlarsen:

rmlarsen:
@sjperkins Can you address the lint and BUILD file formatting errors here:

https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/4336/consoleFull

sjperkins:
Fixed

rmlarsen:
@sjperkins Thanks!

rmlarsen:
@sjperkins it looks like you missed a few.

sjperkins:
@rmlarsen I ran buildifier on the source tree, but it different extra line (2306) than that mentioned in the log (2355 below). 

```bash
=== Sanity check step 3 of 8: do_buildifier (buildifier check) ===

Running do_buildifier on 287 files

tensorflow/python/kernel_tests/BUILD # reformat 

buildifier took 0 s

2355d2354
< 
```

Would you test again please?

ekelsen:
BTW, can you expand on your use case for the map version of the staging area?  Just out of curiosity...

rmlarsen:

sjperkins:
Taking another look

sjperkins:
> BTW, can you expand on your use case for the map version of the staging area? Just out of curiosity...

1. Partial inserts. Feeding one Staging Area (key, tuple) from multiple (possibly, distributed) locations. i.e. scatter portions of the problem from master/chief and feed other portions of the problem on a local worker. I've a [data source](https://github.com/ska-sa/montblanc/blob/75e47e5a4ee8e7f599b6e5805333a39754beadd8/montblanc/impl/rime/tensorflow/sources/ms_source_provider.py#L84-L213) interface to my computation and I want to allow my users to feed data in multiple places.
2. Capacity and Memory Budget. Feeding a potentially very large set of Tensors onto the GPU asynchronously and making sure that only a restricted portion of the set is on the Staging Area at one time.
3. Peek functionality. Caching Tensors on GPU for the duration of other operations (i.e. point 2).
4. Ordering. I'm grouping data (and computation) by a integer key defined by data tile extents. There's a partial ordering here thats useful to me. Ultimately I want to use this key for [DHT](https://en.wikipedia.org/wiki/Distributed_hash_table) functionality, but a priority queue is always useful.
5. Holding ref-counted Tensors, rather than memcpying tensors slices around.

In general, having a stateful hash table in the Graph, logically grouping ref-counted Tensor objects together on either CPUs or GPUs is pretty useful. The Queues always seemed heavyweight (https://github.com/tensorflow/tensorflow/issues/3009, https://github.com/tensorflow/tensorflow/issues/5907, https://github.com/tensorflow/tensorflow/issues/6845) so I wanted to avoid memcpys. I saw some tensor slicing memcpys in Barrier, and it's not always clear what's happening lower down in the PriorityQueue/FIFOQueue level.

rmlarsen:

rmlarsen:

rmlarsen:

rmlarsen:
Thanks for the contribution. This is pretty cool stuff!

sjperkins:
@rmlarsen @ekelsen Thanks for the reviews and accepting the PR.

ekelsen:
Ditto on the thanks - this is a great contribution.

sjperkins:
Will this be included in the 1.2.0 release?

ekelsen:
1.2 seems likely, but I can't say for sure.

But one minor request - the staging_op_test now takes a bit too long for a "small" test.  Can you submit a minor PR that bumps the test from small -> medium?  I'll review and approve quickly.  Thanks!

ekelsen:
Could also consider making the test have a shard count of 2. "shard_count = 2,"

sjperkins:
> 1.2 seems likely, but I can't say for sure.

It would be helpful if it made it in, as I'm developing with the Map Staging Area as a dependency.

ekelsen:
It seems that this PR has caused the staging_op_test to become flaky.  Sometimes it passes in 30s, sometimes it fails and times out in 300s.  My original intuition to increase the test time was incorrect.  Can you look into a possible race condition?

ekelsen:
(This is blocking merging now, so it is rather important to fix the root cause of the hanging test.)

jhseu:
We'll likely revert this change for now to unblock merging. Feel free to resend the pull request with the fix.

sjperkins:
@jhseu I suspect the fix is in #10276 as I've done 5000 runs on `stage_op_test` and have seen no errors. `map_stage_op_test` is on 572/5000 runs on  now.

PR #9693, opened 2017-05-05
Added 2 unit tests for BasicLSTMCell to check ValueError.

chris-chris:
- Tests that dimension 0 in both(x and m) shape must be equal.
- Tests that state_size must be num_units * 2.

### DIFF
--- a/tensorflow/contrib/rnn/python/kernel_tests/core_rnn_cell_test.py
+++ b/tensorflow/contrib/rnn/python/kernel_tests/core_rnn_cell_test.py
@@ -206,8 +206,7 @@ class RNNCellTest(test.TestCase):
         x = array_ops.zeros(
           [batch_size, input_size])
         m = array_ops.zeros([batch_size - 1, state_size])
-        with self.assertRaisesRegexp(ValueError,
-            r"Dimension 0 in both shapes must be equal"):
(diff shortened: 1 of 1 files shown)


tensorflow/contrib/rnn/python/kernel_tests/core_rnn_cell_test.py
```diff
@@ -194,6 +194,50 @@ def testBasicLSTMCell(self):
...
+          "root", initializer=init_ops.constant_initializer(0.5)):
+        num_units = 2
+        state_size = num_units * 3 # state_size must be num_units * 2
+        batch_size = 3
+        input_size = 4
+        x = array_ops.zeros(
+          [batch_size, input_size])
+        m = array_ops.zeros([batch_size, state_size])
+        with self.assertRaisesRegexp(ValueError,
+            r"Dimensions must be equal, but are 3 and 2"):
```
vrv: It becomes difficult to change error messages when tests check for exact contents, can we just do self.assertRaises instead?

chris-chris: @vrv Okay, I'll fix it! Thanks!

vrv:

vrv:

PR #9698, opened 2017-05-05
Fix build scripts for MKL

caisq:


vrv (approved):
I assume a TODO is to have a build for testing the MKL config though.

andydavis1 (approved):

caisq:
@vrv, yes, that's a TODO item. But it's part of the build system config, not part of the code.

Issue #9600, opened 2017-05-02
Why the basic function of different rnn cells are different?


rylanchiu:
I'd like to tackle two issues. The first one is about the initializer. I can define initializer in `LSTMCell ` but not in `GRU` cell. The second one is `layer_normalization`. There is `LayerNormBasicLSTMCell` available for using layer normalization on LSTM cell. But there is not other kinds of cell used in that way. I wonder why not make this general functions available for all kinds of cells?

ebrevdo:
PRs are welcome.

chris-chris:
If there is no volunteer yet, can I try this issue?
I reviewed source codes of GRUCell, LSTMCell, and LayerNormBasicLSTMCell.
I think I can give it a shot.

ebrevdo:
If making API changes I encourage sooner rather than later. They should be
backwards compatible. I'll move rnncells back to core soon and it will be
harder to make changes after that.

chris-chris:
@ebrevdo Okay, I'll make it quick.

There are two tasks.
1. Add initializer parameter to GRUCell, MultiRNNCell
I've done this issue. I'll make a PR for this one first.

2. Add layer_norm, norm_gain, norm_shift parameters
 to TimeFreqLSTMCell, CoupledInputForgetGateLSTMCell, GridLSTMCell, BidirectionalGridLSTMCell ( all LSTMCells in rnn_cell.py )

```
    Args:
      layer_norm: If `True`, layer normalization will be applied.
      norm_gain: float, The layer normalization gain initial value. If
        `layer_norm` has been set to `False`, this argument will be ignored.
      norm_shift: float, The layer normalization shift initial value. If
        `layer_norm` has been set to `False`, this argument will be ignored.
```

This issue will take some more time to apply because I couldn't find a way to make general wrapper like `LayerNormWrapper`(I named it).

I think I need to add layer_norm parameters to all those LSTMCells, and implement it separately.
If you have any good idea, please tell me.

chris-chris:
https://github.com/chris-chris/tensorflow/commit/a9ddcbddddb3d8a375c1e128d3c21b78e8bef2f6

I applied layer normalization to CoupledInputForgetGateLSTMCell. 
But I'm not sure that this is the right application. So I need some reviews.
I didn't make a PR yet.

Please leave a comment if you have any feedback.

I'm now reading the references for layer normalization now...

[Reference]
https://arxiv.org/abs/1607.06450
"Layer Normalization"
[...]

PR #9700, opened 2017-05-05
Add initializer to GRUCell, MultiRNNCell #9600

chris-chris:
Add initializer to GRUCell, MultiRNNCell #9600

### DIFF
--- a/tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
+++ b/tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
@@ -84,13 +84,14 @@ class GRUCell(RNNCell):
   """Gated Recurrent Unit cell (cf. http://arxiv.org/abs/1406.1078)."""
 
   def __init__(self, num_units, input_size=None, activation=tanh, reuse=None,
-               initializer=None):
+               kernel_initializer=None, bias_initializer=None):
     super(GRUCell, self).__init__(_reuse=reuse)
     if input_size is not None:
       logging.warn("%s: The input_size parameter is deprecated.", self)
     self._num_units = num_units
     self._activation = activation
-    self._initializer = initializer
+    self._kernel_initializer = kernel_initializer
+    self._bias_initializer = bias_initializer
 
   @property
   def state_size(self):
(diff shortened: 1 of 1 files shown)


ebrevdo:
LGTM so long as tests continue to pass.

tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
```diff
@@ -100,11 +102,11 @@ def output_size(self):
 
   def call(self, inputs, state):
     """Gated recurrent unit (GRU) with nunits cells."""
-    with vs.variable_scope("gates"):  # Reset gate and update gate.
+    with vs.variable_scope("gates", initializer=self._initializer):  # Reset gate and update gate.
```
vrv: @ebrevdo are you concerned that eventually someone might want to pass different initializers to these different variables?

ebrevdo:
We should consult with Lukasz about what's common, and try to be consistent
with keras. The initializer arg is standard and we can always add others to
override specific layer behavior.  I'd check the keras code but am ooo
today. Will let you make the final decision.

tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
```diff
@@ -83,12 +83,14 @@ def call(self, inputs, state):
...
 
-  def __init__(self, num_units, input_size=None, activation=tanh, reuse=None):
+  def __init__(self, num_units, input_size=None, activation=tanh, reuse=None,
+               initializer=None):
     super(GRUCell, self).__init__(_reuse=reuse)
     if input_size is not None:
       logging.warn("%s: The input_size parameter is deprecated.", self)
     self._num_units = num_units
     self._activation = activation
+    self._initializer = initializer
```
lukaszkaiser: Please add 2 initializer arguments: kernel_initializer and bias_initializer (which will need to be passed to linear). That'll be compatible with current tf.layers. (We might want more kinds of initializers layer, but at least as a start this looks good.)

lukaszkaiser (changes requested):
I think we should start with 2 initializers: kernel_initializer and bias_initializer, as in most current tf.layers. We might add more later, but I think having these 2 is a good idea (esp. as some cells already tweak the biases).

tensorflow/contrib/rnn/python/ops/core_rnn_cell_impl.py
fchollet: For reference, Keras RNNs use the following initializers:

`kernel_initializer` -> for `kernel` (which parametrizes the input transform)
`recurrent_initializer` -> for `recurrent_kernel` (which parametrizes the recurrent transform)
`bias_initializer` -> for `bias`

chris-chris: @lukaszkaiser @fchollet @ebrevdo 

Yeap, I'm checking the keras documents for this issue.

To the point, I have a question.
Is 'kernel_initializer' and 'bias_initializer' a string parameter like 'zeros' 'ones' 
or it can just directly insert the parameters into initial values?

I found some strings for initializer like this

'ones'
'zeros'
'random_uniform'
'glorot_normal'
https://keras.io/initializers/#usage-of-initializers

but in tensorflow, the initializer only directly assigned the initializer parameters to the initial values.

lukaszkaiser: In TF we use initializer functions/objects, not strings, so I think that'd be preferable.

chris-chris: @lukaszkaiser 

Thanks for your reply.
I have one more question.

According to the keras documentation, it seems `kernel_initializer` is same with the `initializer` parameter I added.

```
kernel is a weights matrix created by the layer, and bias is a bias vector created by the layer (only applicable if use_bias is True).
```

If `initializer` and `kernel_initializer` are same, is it okay to add `bias_initializer` only?

fchollet: As much as possible, you should use the same keywords (for the same features) to avoid API fragmentation. It is preferable not to have 2 redundant APIs where the same things are configured via similar-but-not-quite-the-same keywords.

In TF layers, initializers should simply be functions (this is also the case in Keras, but we also allow function names as strings for convenience).

chris-chris: https://github.com/chris-chris/tensorflow/commit/faa8fef734fab591cd05c1a3e110203e32ada0db

I added kernel_initializer and bias_initializer to GRUCell.
I used kernel_initializer to initialize weights and used bias_initializer to initialize bias.

Could you check this out?

lukaszkaiser (approved):
Looks good to me, thanks.

rmlarsen:

PR #9606, opened 2017-05-03
Group LSTM cell

okuchaiev:
Implementing (in contrib/rnn) group LSTM cell from "Factorization Tricks for LSTM Networks" ICLR 2017 Workshop.
https://openreview.net/forum?id=ByxWXyNFg&noteId=ByxWXyNFg

PR #9704, opened 2017-05-05
Making sure GLSTMCell is visible through tf.contrib.rnn.GLSTMCell

okuchaiev:
This is a followup to #9606. Just adding it to 

>  contrib/rnn/\_\_init\_\_.py.

One question, though. Am I supposed  to add something to:
> /contrib/rnn/python/tools/checkpoint_convert.py 

If yes, what exactly? The challenge here is that GLSTM has different number of variables/groups - hence number of "kernels" is determined at runtime.

### DIFF
--- a/tensorflow/contrib/rnn/__init__.py
+++ b/tensorflow/contrib/rnn/__init__.py
@@ -46,6 +46,7 @@ See @{$python/contrib.rnn} guide.
 @@IntersectionRNNCell
 @@PhasedLSTMCell
 @@HighwayWrapper
+@@GLSTMCell
 
 ### RNNCell wrappers
 @@AttentionCellWrapper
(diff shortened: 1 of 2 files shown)


ebrevdo:
LGTM

vrv:

okuchaiev:
also, fixed an issue in 08f91565e97067041a4d2f12093e533b51344cbe 
Verified this works on models/tutorials/ptb and on https://github.com/google/seq2seq  (older version didn't work on seq2seq).

vrv:

ebrevdo:
No need to use checkpoint converter because no one has created checkpoints
with your cell yet, and we plan to change the banking system very soon so
they won't have a chance to create them before we start using the new
variable names.

PR #9723, opened 2017-05-06
suggest is_training not known at construction time

Androbin:


Androbin:
@vrv My concern is, that the phrase "being **built** for training" suggests knowledge of `is_training` at **construction** time. This is why I proposed replacing "**build**" with "**run**".

vrv:
Oh, sorry, I didn't exactly parse the commit message properly, and the existing
documentation was already pretty confusing -- how about:

* Layers that behave differently during training should take
   an `is_training` boolean Tensor as input to conditionally
   choose different computations paths (e.g., using `tf.cond`).

Androbin:
@vrv **SGTM** but it is good style to accept both Python `Booleans` and `bool` `Tensor`s.

vrv:
A python boolean would make it a static graph-build-time decision, not a dynamic one though, but I guess that's up to the caller to decide.  (Will you send the PR?)

Androbin:
Continued at #9730

PR #9759, opened 2017-05-08
[XLA] Add ability to run the XLA unit tests against a different device

DavidNorman:
This is a squashed version of https://github.com/tensorflow/tensorflow/pull/9409

### DIFF
--- a/tensorflow/compiler/tests/build_defs.bzl
+++ b/tensorflow/compiler/tests/build_defs.bzl
@@ -1,12 +1,13 @@
 """Build rules for Tensorflow/XLA testing."""
 
 load("@local_config_cuda//cuda:build_defs.bzl", "cuda_is_configured")
+load("//tensorflow/compiler/tests:plugin.bzl", "plugin_device", "plugin_types")
 
 def all_backends():
+  b = ["cpu", "plugin"]
   if cuda_is_configured():
-    return ["cpu", "gpu"]
-  else:
-    return ["cpu"]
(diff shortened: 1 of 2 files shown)


tensorflow/compiler/tests/build_defs.bzl
```diff
@@ -1,12 +1,13 @@
 """Build rules for Tensorflow/XLA testing."""
 
 load("@local_config_cuda//cuda:build_defs.bzl", "cuda_is_configured")
+load("//tensorflow/compiler/tests:plugin.bzl", "plugin_device", "plugin_types")
 
 def all_backends():
+  b = ["cpu", "plugin"]
```
hawkinsp: I think this is already an improvement, thanks!
I would prefer that we avoid explicitly mentioning a "plugin" backend by name here. It isn't especially intuitive and we might want more than one such backend.

Could we instead have plugin.bzl contain a dictionary of (backend name, flags) entries, rather than hardcoding one plugin named "plugin"?

DavidNorman:
Hi all.  I wonder if you might consider this patch again?  Just trying to clear the many diffs between my version of TF and the public one.

rmlarsen:
@hawkinsp could you give this another look? Thanks.

DavidNorman:
@hawkinsp hi.  I wonder if this can be merged now?

hawkinsp (approved):

jhseu:
Jenkins, test this please

PR #9762, opened 2017-05-08
Bazel CI / TensorFlow: fix more errors

laszlocsomor:
In this commit:

1. Keep just the first character of the python
version checker script's output. This is a simple
way to strip \n and \r and we only need the major
version anyway. The string is later used in
writing the .bazelrc file where these characters
would cause problems.

2. Export environment variables in
run_configure_for_cpu_build, use their default
values

3. Allow a user-defined value for BAZEL_SH

4. Fix python_configure.bzl:

4.a. Fix the symlink creation on Windows: create a
single junction pointing to the src_dir. This is a
trick to get around the problem that on Windows we
can't create symlinks. By creating this symlink
and declaring the same outputs as on Linux/Darwin,
we make it look like that the genrule successfully
produced these files. This only works in a
single-machine execution environment.

4.b. Get rid of the "select" in
@local_config_python//:numpy_include and
python_include. The results of `find` and `dir`
only make sense on Linux/Darwin and Windows
respectively, so there's no point in generating
the non-Windows genrule on Windows and the
Windows-specific one on non-Windows, because all
the paths would be wrong. So instead of that,
generate just the genrule corresponding to the
platform.

4.c. Write the genrule.cmd's commands to scripts
and execute those instead. This avoids the
difficulties with escaping backslashes and quotes.
Backslashes are necessary on Windows because paths
with forward slashes aren't generally accepted,
e.g. "cmd.exe /c foo/bar" doesn't work because
Windows believes foo is an executable and /bar is
its switch:

    c:\tempdir>type bar.cmd
    @echo this is bar.cmd, %%1=(%1)

    c:\tempdir>cmd.exe /c bar/baz
    this is bar.cmd, %1=(/baz)

5. simplify the "is_windows" function in
./configure

Fixes https://github.com/bazelbuild/bazel/issues/2892


caisq:

tensorflow/tools/ci_build/windows/bazel/bazel_test_lib.sh
```diff
@@ -175,6 +175,11 @@ function run_configure_for_cpu_build {
   if [ -z "$CC_OPT_FLAGS" ]; then
     export CC_OPT_FLAGS="-march=native"
   fi
+  export TF_NEED_VERBS=0
```
meteorcloudy: Can you also add this in function run_configure_for_gpu_build ?

laszlocsomor: Done.

rmlarsen:

rmlarsen:
@meteorcloudy could you take another look?

meteorcloudy:
@laszlocsomor I tested your change on my local machine and found a few bugs.
First, dest_dir is not the complete output path, we need to add `$(@D)` as prefix.
Second, we cannot just create a junction to the include directory. Because Bazel cleans up all the output files which means all the files in the original directory will be deleted. So I think the safest way is to copy all the headers on Windows. Can you apply the following patch?

```
$ git diff third_party/py/python_configure.bzl
diff --git a/third_party/py/python_configure.bzl b/third_party/py/python_configure.bzl
index ed576a25d..525f4acc6 100644
--- a/third_party/py/python_configure.bzl
+++ b/third_party/py/python_configure.bzl
@@ -122,8 +122,10 @@ def _symlink_genrule_for_dir(repository_ctx, src_dir, dest_dir, genrule_name):
         executable = True,
         content = "\n".join([
             '@rem Auto-generated by //third_party/py/python_configure.bzl',
-            '@mklink /J "%s" "%s"' % (dest_dir, src_dir)]))
-    cmd = "cmd.exe /c $$(echo $(location %s) | sed 's,/,\\\\,g')" % cmd_script
+           'set DEST_DIR=%%1\\%s\\' % dest_dir,
+           'rmdir /s /q %DEST_DIR%',
+            'xcopy /e /q /y "%s" "%%DEST_DIR%%"' % src_dir]))
+    cmd = "cmd.exe /c \\\"$$(echo $(location %s) $(@D) | sed 's/\//\\\\\\\\/g')\\\"" % cmd_script
   else:
     cmd_script = "%s-cmd.sh" % genrule_name
     repository_ctx.file(
```

laszlocsomor:
> Second, we cannot just create a junction to the include directory. Because Bazel cleans up all the output files which means all the files in the original directory will be deleted. 

Ouch. Heavy facepalm. Great catch Yun, thanks, also for the patch; I'll apply it now.

laszlocsomor:
Applied your fix, plus added "/s" to xcopy, as advised by the docs: https://technet.microsoft.com/en-us/library/cc771254(v=ws.11).aspx

meteorcloudy:
@laszlocsomor Thanks, the change looks good to me!

rmlarsen:
@mrry would you please review this?

mrry:
It looks fine to me, but I'm not familiar enough with Skylark to know if it's correct. Can we somehow kick off Bazel/Windows and Bazel/Windows/GPU presubmits to make sure that things work?

meteorcloudy:
I think that's a great idea! @gunan Do we have plan to enable it?

meteorcloudy:
@mrry For this change, I tested on my local machine, it should fix the local_python_config breakage.
I also found some other issues, will send PR to fix them.

gunan:
This is a matter of capacity, and test times.
I do not want to slow down development for everyone at the moment, but it is always possible to manually run bazel presubmits.

Jenkins, test this please.

gunan:
Running bazel windows presubmit here:
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/5

rmlarsen:
@gunan @mrry who can review / approve this?

gunan:
In our CI this PR is failing.
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/6/console

@meteorcloudy @laszlocsomor could you take a look?

meteorcloudy:
```
13:00:07 INFO: From Executing genrule @local_config_python//:numpy_include:
13:00:07 26 File(s) copied
13:00:07 INFO: From Executing genrule @local_config_python//:python_include:
13:00:07 100 File(s) copied
```
This indicates the local_python_config problem is fixed.

The new error we got:
```
13:00:10 ERROR: missing input file '@local_jdk//:bin/jar'
```
is because on Windows it should be `@local_jdk//:bin/jar.exe` which is already fixed at https://github.com/bazelbuild/bazel/commit/c77eac95ffdee4f127e20266c41cb92cec2a8956.
But the fix is not in 0.4.5. We need to update Bazel to 0.5.0 once it's released.

laszlocsomor:
Thanks @meteorcloudy for investigating!

gunan (approved):

PR #9765, opened 2017-05-08
Add skeleton/example XLA plugin device

DavidNorman:
I wonder if this would be a useful thing to have in the repo?


eliben:
Hi David - thanks for the contribution!

Could you elaborate a little bit on the motivation for landing this? Is it much different from, say, the existing CPU backend? If we just want to make new backends easier to write, would it not be simpler to beef up the docs a bit for what it takes to write a new backend?

All these backends naturally share quite a bit of boilerplate, as your PR demonstrates. Keeping this boilerplate working in the face of occasional API changes is burdensome, and any in-tree backend adds to the burden. This is why, although I'm not *opposed* to having such an example backend in-tree, I do want to hear some more about the motivation and potential advantages of having one.

DavidNorman:
The CPU backend isn't really a skeleton, and depending on your jumping off point it is too complex of an example.

I was really just sharing the minimal example that I generated when starting my own backend.  Someone on the google group was having trouble with a minimal working example.

but I am happy to abandon it.  it was really just a point thing.

from a purely selfish point of view, having a working example that closely matches our own backend implementation will tend to prevent any API changes that cause significant work for us.  for instance, it would be a massive pain if someone decided to have tf2xla produce an llvm structure instead of Hlo objects.

DavidNorman:
incidentally - although this is the wrong place to ask this question... i have an Hlo optimization stage that looks for 'big' ops (conv, dot) and outlines them, and downstream elementwise ops (up to a max computation parameter size).  would you like this upstreamed in the main service directory, or shall I keep it private?

eliben:
@DavidNorman actually we're in the process of creating a much simpler backend, based on the [HLO evaluator](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/hlo_evaluator.h) -- the evaluator isn't 100% feature complete yet, but it will eventually be. Having the evaluator as an XLA backend could serve both has as an example for writing a new simple backend, and be useful for testing XLA and the evaluator itself. @kayzhu has more details.

We've been busy adding more features to the evaluator but hadn't started wrapping it in an actual backend yet. If you'd like to contribute, would you consider sending a PR that does that?

DavidNorman:
yeah - i don't think it would be too hard to drop into my example.  if the evaluator will, in itself,  handle calls and maps, then i suspect it is as simple as copying the hlo module for the 'compile' stage, and calling the evaluator at execution time.  copying buffers in and out from the stream is the same as my existing code.

kayzhu:
@DavidNorman Thanks for your interest and contribution!

As @eliben mentioned, we are working on getting the elevator more feature-complete and wrap it up as a reference backend. Support for a few more simple ops are on the way, but I haven't worked on calls and maps yet.

If you are interested in contributing to the reference backend, I'm more than happy to review incoming PRs.

DavidNorman:
Oops - forgot to squash the merge.  Now opened a new one:

https://github.com/tensorflow/tensorflow/pull/10031

PR #9768, opened 2017-05-08
download if not exists, extract never overwrite

codingyourlife:
Removed the comment for wget and added arguments so that download and extract happens just once


rmlarsen:
@shlens would you mind taking a look?

rmlarsen:

rmlarsen:
@shlens you seemed to have an opinion about how should review this. Can you please assign a reviewer?

shlens (approved):

Issue #9462, opened 2017-04-26
numpy prod overflow during creating tensor


fesun:
I am trying to allocate super large tensor using tensorflow, but failed.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/tensor_util.py#L417
Above code uses numpy.prod to calculate shape size, and for numpy everything is typed, say if the shape is [500000000, 5], then numpy.prod returns -1794967296, it's very easy to reproduce it.
So how about use int64 instead? int64 should be large enough for any tensor.
Changing shape_size = np.prod(shape) to shape_size = np.prod(shape, dtype=np.int64) should fix it. Also about 100 lines of code using np.prod, could we change them all to int64?

girving:
@fesun Are you on a 32 bit machine?  I get

    >>> np.prod([500000000, 5])
    2500000000
    >>> np.prod([500000000, 5]).dtype
    dtype('int64')

Explicitly setting `dtype=np.int64` does seem like the right solution, but I'd like to understand why it's breaking for you first.

mrry:
Maybe it's a Windows issue? https://docs.scipy.org/doc/numpy/user/basics.types.html#id3 says that the default integer type is the same as a C `long`, which is 32 bits in MSVC.

fesun:
@girving I am on 64 bit windows machine, I think @mrry is right, numpy default int type inside windows is 32 bit.

girving:
@fesun Would you be up for a pull request?  Adding `dtype=np.int64` sounds good.

fesun:
@girving sure, thanks.

PR #9780, opened 2017-05-09
fix numpy.prod and GatherGrad overflow for large tensors inside windows

fesun:
#9462

### DIFF
--- a/tensorflow/python/framework/tensor_shape.py
+++ b/tensorflow/python/framework/tensor_shape.py
@@ -66,6 +66,9 @@ class Dimension(object):
 
   def __int__(self):
     return self._value
+  
+  def __long__(self):
+    return self._value
 
   def __index__(self):
     # Allow use in Python 3 range
--- a/tensorflow/python/framework/tensor_util.py
+++ b/tensorflow/python/framework/tensor_util.py
@@ -368,9 +368,9 @@ def make_tensor_proto(values, dtype=None, shape=None, verify_shape=False):
       np_dt = dtype.as_numpy_dtype
     else:
       np_dt = None
-    # if shape is None, numpy.prod returns None when dtype is not set, but raise
+    # If shape is None, numpy.prod returns None when dtype is not set, but raises
     # exception when dtype is set to np.int64
-    if not(shape is None) and np.prod(shape, dtype=np.int64) == 0:
+    if shape is not None and np.prod(shape, dtype=np.int64) == 0:
       nparray = np.empty(shape, dtype=np_dt)
     else:
       _AssertCompatible(values, dtype)
--- a/tensorflow/python/framework/tensor_util.py
+++ b/tensorflow/python/framework/tensor_util.py
@@ -368,7 +368,9 @@ def make_tensor_proto(values, dtype=None, shape=None, verify_shape=False):
       np_dt = dtype.as_numpy_dtype
     else:
       np_dt = None
-    if np.prod(shape) == 0:
+    # if shape is None, numpy.prod returns None when dtype is not set, but raise
+    # exception when dtype is set to np.int64
+    if not(shape is None) and np.prod(shape, dtype=np.int64) == 0:
       nparray = np.empty(shape, dtype=np_dt)
     else:
       _AssertCompatible(values, dtype)
@@ -414,7 +416,7 @@ def make_tensor_proto(values, dtype=None, shape=None, verify_shape=False):
     shape_size = nparray.size
   else:
     shape = [int(dim) for dim in shape]
-    shape_size = np.prod(shape)
+    shape_size = np.prod(shape, dtype=np.int64)
     is_same_size = shape_size == nparray.size
 
     if verify_shape:
@@ -491,7 +493,7 @@ def MakeNdarray(tensor):
 
   """
   shape = [d.size for d in tensor.tensor_shape.dim]
-  num_elements = np.prod(shape)
+  num_elements = np.prod(shape, dtype=np.int64)
   tensor_dtype = dtypes.as_dtype(tensor.dtype)
   dtype = tensor_dtype.as_numpy_dtype
 
--- a/tensorflow/python/ops/array_grad.py
+++ b/tensorflow/python/ops/array_grad.py
@@ -355,10 +355,10 @@ def _GatherGrad(op, grad):
   """Gradient for Gather op."""
   # params can be large, so colocate the shape calculation with it.
   #
-  # params can be super-large for sparse model, array_ops.shape raises
-  # exception inside Windows when any dimension is larger than int32,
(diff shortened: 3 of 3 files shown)


fesun:
I signed it!

fesun:
Sadly, this commit can't solve all overflow issues, see #9781 for tf.shape tf.size etc.

fesun:
fixed #9781

tensorflow/python/framework/tensor_util.py
```diff
@@ -368,7 +368,9 @@ def make_tensor_proto(values, dtype=None, shape=None, verify_shape=False):
       np_dt = dtype.as_numpy_dtype
     else:
       np_dt = None
-    if np.prod(shape) == 0:
+    # if shape is None, numpy.prod returns None when dtype is not set, but raise
```
rmlarsen: s/if/If/
s/raise/raises/

rmlarsen (changes requested):

rmlarsen:

tensorflow/python/framework/tensor_util.py
```diff
@@ -368,7 +368,9 @@ def make_tensor_proto(values, dtype=None, shape=None, verify_shape=False):
...
+    # exception when dtype is set to np.int64
+    if not(shape is None) and np.prod(shape, dtype=np.int64) == 0:
```
rmlarsen: I would write this as 

if shape is not None and np.prod(shape, dtype=np.int64) == 0:

rmlarsen:
@fesun This appears to break several tests. Can you take a look?

fesun:
@rmlarsen fixed it. Now too many code depends on int32, just suspended windows exception of tf.shape for large tensor.

tensorflow/python/ops/array_grad.py
```diff
@@ -354,9 +354,15 @@ def _PreventGradientGrad(op, _):
 def _GatherGrad(op, grad):
   """Gradient for Gather op."""
   # params can be large, so colocate the shape calculation with it.
+  #
+  # params can be super-large for sparse model, array_ops.shape raises
+  # exception inside Windows when any dimension is larger than int32,
```
rmlarsen: Nits:
s/super-large/very large/
s/inside Windows/on the Windows platform/

Period after int32.

Maybe the last sentence should be

"params_shape is not used in apply_sparse gradients, so it's fine to convert it back to int32 regardless of truncation."

tensorflow/python/ops/array_grad.py
```diff
@@ -354,9 +354,15 @@ def _PreventGradientGrad(op, _):
 def _GatherGrad(op, grad):
   """Gradient for Gather op."""
   # params can be large, so colocate the shape calculation with it.
+  #
+  # params can be super-large for sparse model, array_ops.shape raises
```
rmlarsen: Could we fix the underlying issue in array_ops.shape instead?

fesun: Here is the exception:
```
>>> import tensorflow as tf
>>> a = tf.get_variable("W", [3000000000], dtype=tf.bool)
>>> b = tf.shape(a)
  File "D:\tensorflow\copyable\python3.5\lib\site-packages\tensorflow\python\ops\array_ops.py", line 245, in shape
    return shape_internal(input, name, optimize=True, out_type=out_type)
  File "D:\tensorflow\copyable\python3.5\lib\site-packages\tensorflow\python\ops\array_ops.py", line 271, in shape_i
nternal
    return constant(input_shape.as_list(), out_type, name=name)
  File "D:\tensorflow\copyable\python3.5\lib\site-packages\tensorflow\python\framework\constant_op.py", line 102, in
 constant
    tensor_util.make_tensor_proto(value, dtype=dtype, shape=shape, verify_shape=verify_shape))
  File "D:\tensorflow\copyable\python3.5\lib\site-packages\tensorflow\python\framework\tensor_util.py", line 377, in
 make_tensor_proto
    nparray = np.array(values, dtype=np_dt)
OverflowError: Python int too large to convert to C long
```
Windows numpy throws exception when converting large int to C long while linux will truncate it silently.
```
Python 3.4.3 (default, Nov 17 2016, 01:08:31)
[GCC 4.8.4] on linux
>>> import numpy as np
>>> a = np.array(3000000000, np.int32)
>>> a
array(-1294967296, dtype=int32)
>>>
``` 
Discussed in #9781 , it's hard to change tf.shape default int32 to int64, also can't change make_tensor_proto, it will make hundreds of dtype inference.

tensorflow/python/ops/array_grad.py
fesun: @rmlarsen will change comments as you suggested

rmlarsen: @fesun OK, sounds good. If you update the comment, this should be ready to merge.

rmlarsen:

fesun:
@rmlarsen  There are still some errors, how to run those jenkins test locally? So I don't need to submit change and wait for you to invoke jenkins test.

tensorflow/python/framework/tensor_shape.py
```diff
@@ -66,6 +66,9 @@ def __ne__(self, other):
 
   def __int__(self):
     return self._value
+  
+  def __long__(self):
+    return self._value
```
fesun: numpy.prod raises exception when dtype is set to numpy.int64 in python2
```
TypeError: long() argument must be a string or a number, not 'Dimension'
```
So, add __long__ for Dimension.

tensorflow/python/ops/array_grad.py
```diff
@@ -355,10 +355,10 @@ def _GatherGrad(op, grad):
...
   # params can be large, so colocate the shape calculation with it.
   #
-  # params can be super-large for sparse model, array_ops.shape raises
-  # exception inside Windows when any dimension is larger than int32,
-  # params_shape is not used by optimizer apply_sparse gradients, it's
-  # fine to convert it back to int32 regardless of truncation
+  # params can be very large for sparse model, array_ops.shape raises
+  # exception on the Windows platform when any dimension is larger than
+  # int32. params_shape is not used in optimizer apply_sparse gradients,
+  # so it's fine to convert it back to int32 regardless of truncation.
```
fesun: modified comments according to review suggestion

rmlarsen:

rmlarsen (approved):

rmlarsen:
Thanks for the contribution!

asimshankar:
@fesun: Could you elaborate on the exception thrown that prompted the change to `array_grad.py`? Is there a unittest/code snippet you could share that causes the exception?

(`array_ops.shape` would throw an exception on all platforms if the provided tensor had a dimension that doesn't fit in 32-bits. So I'm trying to understand what the platform specific part is. Thanks!)

Issue #9703, opened 2017-05-05
pkg-config file generation


arrufat:
Hi! I was wondering if a pkg-config file for the tensorflow library (`libtensorflow.so`) could be added so that depending projects could use it more easily.

For the current stable version (`1.1.0`), if we installed it into `/usr` (which should be configurable), the generated pkg-config file `tensorflow.pc` should look something like:
```
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
modules=1

Name: tensorflow
Version: 1.1.0
Description: Library for computation using data flow graphs for scalable machine learning
Requires:
Libs: -L${libdir} -ltensorflow -lstdc++
Cflags: -I${includedir}/tensorflow
```
Thanks in advance :)

arrufat:
I'd like to make a pull request, but I don't know how you would prefer this to be implemented.
I've just written a small shell script that generates this pkconfig file for a given prefix.
It coud be run as `./generate-pc.sh /usr`:
```bash
#!/usr/bin/env bash

TF_VERSION=`grep Release RELEASE.md | head -1 | sed -e 's/# Release //g'`

if [[ $# == 0 ]]
then
	echo "Usage:"
	echo -e "\t$0 install_prefix"
	exit
fi

TF_PREFIX=$1

echo "Generating pkgconfig file for TensorFlow $TF_VERSION in $TF_PREFIX"

cat << EOF > tensorflow.pc
prefix=${TF_PREFIX}
exec_prefix=\${prefix}
libdir=\${exec_prefix}/lib
includedir=\${prefix}/include
modules=1

Name: TensorFlow
Version: ${TF_VERSION}
Description: Library for computation using data flow graphs for scalable machine learning
Requires:
Libs: -L\${libdir} -ltensorflow -lstdc++
Cflags: -I\${includedir}/tensorflow
EOF
```
Yes, I know, the way I get the version is not the best...

drpngx:
@yifeif is that something we'd like to have or should it be in a separate repo?

yifeif:
@asimshankar who knows more about C API distribution :).

asimshankar:
[...]

PR #9784, opened 2017-05-09
add pkg-config generation script

arrufat:
Opening a pull request as discussed in https://github.com/tensorflow/tensorflow/issues/9703

### DIFF
--- a/tensorflow/c/generate-pc.sh
+++ b/tensorflow/c/generate-pc.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+TF_PREFIX='/usr/local'
+
+usage() {
+    echo "Usage: $0 OPTIONS"
+    echo -e "-p, --prefix\tset installation prefix (default: /usr/local)"
+    echo -e "-v, --version\tset TensorFlow version"
+    echo -e "-h, --help\tdisplay this message"
+}
+
+# read the options
+ARGS=`getopt -o p:v:h --long prefix:,version:,help -n $0 -- "$@"`
+eval set -- "$ARGS"
+
+# extract options and their arguments into variables.
+while true ; do
+    case "$1" in
+        -h|--help) usage ; exit ;;
+        -p|--prefix)
+            case "$2" in
+                "") shift 2 ;;
+                *) TF_PREFIX=$2 ; shift 2 ;;
+            esac ;;
+        -v|--version)
+            case "$2" in
+                "") shift 2 ;;
+                *) TF_VERSION=$2 ; shift 2 ;;
+            esac ;;
+        --) shift ; echo "Try '$0 --help' for more information."; exit 1 ;;
+        *) echo "Internal error! Try '$0 --help' for more information." ; exit 1 ;;
+    esac
+done
+
+echo "Generating pkgconfig file for TensorFlow $TF_VERSION in $TF_PREFIX"
+
+cat << EOF > tensorflow.pc
+prefix=${TF_PREFIX}
+exec_prefix=\${prefix}
+libdir=\${exec_prefix}/lib
+includedir=\${prefix}/include
+modules=1
+
+Name: TensorFlow
+Version: ${TF_VERSION}
+Description: Library for computation using data flow graphs for scalable machine learning
+Requires:
+Libs: -L\${libdir} -ltensorflow
+Cflags: -I\${includedir}/tensorflow
+EOF
--- a/tensorflow/c/generate-pc.sh
+++ b/tensorflow/c/generate-pc.sh
@@ -39,12 +39,11 @@ prefix=${TF_PREFIX}
 exec_prefix=\${prefix}
 libdir=\${exec_prefix}/lib
 includedir=\${prefix}/include
-modules=1
 
 Name: TensorFlow
 Version: ${TF_VERSION}
 Description: Library for computation using data flow graphs for scalable machine learning
 Requires:
 Libs: -L\${libdir} -ltensorflow
-Cflags: -I\${includedir}/tensorflow
+Cflags: -I\${includedir}
 EOF


arrufat:
I've already signed it.

tensorflow/c/generate-pc.sh
```diff
@@ -0,0 +1,50 @@
...
+done
+
+echo "Generating pkgconfig file for TensorFlow $TF_VERSION in $TF_PREFIX"
+
+cat << EOF > tensorflow.pc
+prefix=${TF_PREFIX}
+exec_prefix=\${prefix}
+libdir=\${exec_prefix}/lib
+includedir=\${prefix}/include
+modules=1
```
asimshankar: What does `modules=1` do?
(Wondering why it is needed)

tensorflow/c/generate-pc.sh
```diff
@@ -0,0 +1,50 @@
...
+    esac
+done
+
+echo "Generating pkgconfig file for TensorFlow $TF_VERSION in $TF_PREFIX"
+
+cat << EOF > tensorflow.pc
+prefix=${TF_PREFIX}
+exec_prefix=\${prefix}
+libdir=\${exec_prefix}/lib
+includedir=\${prefix}/include
```
asimshankar: Why define all these variables? Since the pkgconfig file is being generated, can't it just be:

```sh
cat <<EOF >tensorflow.pc
Name: TensorFlow
Version: ${TF_VERSION}
Description: ...
Libs: -L\${TF_PREFIX}/lib -ltensorflow
Cflags: -I\${TF_PREFIX}/include
```
?

tensorflow/c/generate-pc.sh
```diff
@@ -0,0 +1,50 @@
...
+modules=1
+
+Name: TensorFlow
+Version: ${TF_VERSION}
+Description: Library for computation using data flow graphs for scalable machine learning
+Requires:
+Libs: -L\${libdir} -ltensorflow
+Cflags: -I\${includedir}/tensorflow
```
asimshankar: Shouldn't this be: `Cflags: -I\${includedir}` (i.e., no `/tensorflow`) since we want users to be able to use:

```c
#include <tensorflow/c/c_api.h>
```

instead of
```c
#include <c/c_api.h>
```
?

asimshankar (changes requested):
Thanks for the change. It should help with distributions going forward.

That said, I don't have much experience with creating pkg-config files, so some of the questions here might be silly. Please humor me.

(Also, I'm taking inspiration from how https://github.com/Homebrew/homebrew-core/blob/master/Formula/libtensorflow.rb generates the same file. Once this goes it, might make sense to update the homebrew rules to use this)

tensorflow/c/generate-pc.sh
arrufat: You're right, this line is not needed

tensorflow/c/generate-pc.sh
arrufat: I was just looking at some generated pkgconfig files such as OpenCV:
```
# Package Information for pkg-config

prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir_old=${prefix}/include/opencv
includedir_new=${prefix}/include

Name: OpenCV
Description: Open Source Computer Vision Library
Version: 3.1.0
Libs: -L${exec_prefix}/lib -lopencv_stitching -lopencv_superres -lopencv_videostab -lopencv_aruco -lopencv_bgsegm -lopencv_bioinspired -lopencv_ccalib -lopencv_dnn -lopencv_dpm -lopencv_fuzzy -lopencv_hdf -lopencv_line_descriptor -lopencv_optflow -lopencv_plot -lopencv_reg -lopencv_saliency -lopencv_stereo -lopencv_structured_light -lopencv_rgbd -lopencv_surface_matching -lopencv_tracking -lopencv_datasets -lopencv_text -lopencv_face -lopencv_xfeatures2d -lopencv_shape -lopencv_video -lopencv_ximgproc -lopencv_calib3d -lopencv_features2d -lopencv_flann -lopencv_xobjdetect -lopencv_objdetect -lopencv_ml -lopencv_xphoto -lopencv_highgui -lopencv_videoio -lopencv_imgcodecs -lopencv_photo -lopencv_imgproc -lopencv_core
Libs.private: -L/lib64 -lwebp -ljasper -lgtk-3 -lgdk-3 -lpangocairo-1.0 -lpango-1.0 -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lgobject-2.0 -lgthread-2.0 -lglib-2.0 -ldc1394 -lxine -lv4l1 -lv4l2 -lgphoto2 -lgphoto2_port -lprotobuf -lhdf5 -lz -ldl -lm -lpthread -lrt -ltbb
Cflags: -I${includedir_old} -I${includedir_new}
```
And they define all variables, even if they don't use them at all.

tensorflow/c/generate-pc.sh
arrufat: This choice was completely arbitrary on my side

rmlarsen:
@asimshankar could you take another look, please?

asimshankar (approved):
Thanks

rmlarsen:

rmlarsen:
@asimshankar thanks.

arrufat:
Just one quick question. Arch Linux now packages the tensorflow library with the pkgconfig file.
However, they place the header file in 
```
/usr/include/tensorflow/c_api.h
```
instead of
```
/usr/include/tensorflow/c/c_api.h
```
Is that a problem? which one should be the standard? Thanks in advance!

asimshankar:
@arrufat : I'd encourage use of `tensorflow/c/c_api.h`. Either can be made to work, but for consistency with the examples we've been putting out and other source code (e.g., how we distribute the tarballs and where other language bindings expect it to be), I'd lean towards `tensorflow/c/c_api.h`.

Hope that helps.

arrufat:
@asimshankar thanks for the quick responde. I'll suggest that to the packagers.

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #9812, opened 2017-05-10
Implement strided slice (stride > 2) in tf2xla

DavidNorman:
Before, strides are not acceptable in strided_slice where the stride > 1 (or < -1).

This uses a concatenation of normal slices to implement strided_slice.

For backends where all tensor slicing/padding/concating/striding are just index manipulation, this implementation is as efficient as any other.

### DIFF
--- a/tensorflow/compiler/tests/slice_ops_test.py
+++ b/tensorflow/compiler/tests/slice_ops_test.py
@@ -117,7 +117,7 @@ class StridedSliceTest(XLATestCase):
         i = array_ops.placeholder(dtype, shape=[3, 4, 10])
         with self.test_scope():
           o = array_ops.strided_slice(i, [2, 2, 6], [0, 0, 2], [-1, -1, -2])
-      params = {
+        params = {
           i: [[[0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
                [9, 8, 7, 6, 5, 4, 3, 2, 1, 0],
                [5, 3, 1, 7, 9, 2, 4, 6, 8, 0],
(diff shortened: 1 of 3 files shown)


tensorflow/compiler/tf2xla/kernels/strided_slice_op.cc
```diff
@@ -99,6 +97,30 @@ class StridedSliceOp : public XlaOpKernel {
       slice = ctx->builder()->Rev(slice, dimensions_to_reverse);
     }
 
+    if (!simple_strides) {
```
hawkinsp: It took me a while to work out what's going on here --- perhaps add a brief comment explaining the general approach?

tensorflow/compiler/tests/slice_ops_test.py
```diff
@@ -0,0 +1,136 @@
...
+from tensorflow.compiler.tests.xla_test import XLATestCase
+from tensorflow.python.framework import dtypes
+from tensorflow.python.ops import array_ops
+from tensorflow.python.platform import googletest
+
+class SliceTest(XLATestCase):
+
+  def test1D(self):
+    with self.test_session():
+      i = array_ops.placeholder(dtypes.float32, shape=[10])
```
hawkinsp: Since none of these test cases are specific to a particular type, the test coverage would be better if you wrapped these tests in a loop of the form:
```
for dtype in self.numeric_types:
  ...
```
and used `dtype` as the dtype where applicable.

hawkinsp (changes requested):
This is great! Thanks for the contribution.

I think eventually we should add direct support for strides to the XLA Slice() operator. (Not least since Slice() is the inverse of Pad() and Pad() supports dilation.)

That said this is definitely a welcome improvement in functionality!

I have a few minor comments.

hawkinsp (approved):

DavidNorman:
agreed that Slice should probably have strides as a parameter in the long run

hawkinsp (approved):

yifeif:

PR #9815, opened 2017-05-10
Add control input dependency to Java bindings

kbsriram:
Add a builder method to OperationBuilder to allow adding
control inputs to Operations.

### DIFF
--- a/tensorflow/java/src/test/java/org/tensorflow/OperationBuilderTest.java
+++ b/tensorflow/java/src/test/java/org/tensorflow/OperationBuilderTest.java
@@ -24,8 +24,6 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
-import java.util.List;
-
 /** Unit tests for {@link org.tensorflow.OperationBuilder}. */
 @RunWith(JUnit4.class)
 public class OperationBuilderTest {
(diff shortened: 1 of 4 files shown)


tensorflow/java/src/test/java/org/tensorflow/OperationBuilderTest.java
```diff
@@ -149,6 +151,36 @@ public void setAttrShape() {
     }
   }
 
+  @Test
+  public void addControlInput() {
+    assertEquals(0, lessEqualControl(2, 3).size());
+    try {
+      lessEqualControl(3, 2);
```
asimshankar: There is a lot happening in lessEqualControl to be able to tie the exception to the specific line we want.
It might be better to create a simpler graph with a placeholder and have the `try{}` block around just the session run call. Something like:

```java
public void addControlInput() {
  try (Graph g = new Graph();
         Session s = new Session(g);
         Tensor yes = Tensor.create(true);
         Tensor no = Tensor.create(false)) {
    Output placeholder = TestUtil.placeholder(g, "boolean", DataType.BOOL);
    Output assert = g.opBuilder("Assert", "assert")
       .addInput(placeholder)
       .addInputList(new Output[]{})
       .build();
      Output noop = g.opBuilder("NoOp", "noop")
          .addControlInput(assert)
          .build();

     // No problems when the assert succeeds
     s.runner().feed(placeholder, yes).addTarget(noop).run();

     // Exception thrown by the execution of the Assert node
     try {
        s.runner().feed(placeholder, no).addTarget(noop).run();
        fail(...);
      } catch (IllegalArgumentException e) {
        // expected
      }
  }
}
```

tensorflow/java/src/test/java/org/tensorflow/OperationBuilderTest.java
```diff
@@ -24,6 +24,8 @@
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
+import java.util.List;
```
asimshankar: Why was this import needed?

asimshankar (commented):
Thanks very much for the change.

Some minor comments, otherwise looks great.

asimshankar (approved):
Thanks!

tensorflow/java/src/test/java/org/tensorflow/OperationBuilderTest.java
kbsriram: Nice! Thanks @asimshankar for the pointers to simplify the test, and the other suggestions. I've incorporated all of them in the next update of this PR.

caisq:

asimshankar (approved):

kbsriram:
Is there perhaps a flaky test in the windows CMake build?

The only time the ci build ran it reported an error in a seemingly [unrelated area](https://ci.tensorflow.org/job/tensorflow-pr-win-cmake-py/2249/console) of the code; and I'm not able to connect the dots between the reported failure and this change.

caisq:

PR #9828, opened 2017-05-11
Add input function for training and testing (#9617)

kiat-tree:
* Add input function for training and testing

Estimator is decoupled from Scikit Learn interface by moving into separate class SKCompat. Arguments x, y and batch_size are only available in the SKCompat class, Estimator will only accept input_fn

* remove extra comma


vrv:
Didn't I already do this in https://github.com/tensorflow/tensorflow/pull/9658 ?  Closing but feel free to reopen if somehow we should merge this again.

Issue #9855, opened 2017-05-12
Unexpected error at contrib.seq2seq's BeamSearchDecoder


shuuki4:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS
- **TensorFlow installed from (source or binary)**: Source
- **TensorFlow version (use command below)**:  v1.1.0-rc2-773-g7fa0cf39f
- **Bazel version (if compiling from source)**:  0.4.5
- **CUDA/cuDNN version**:  None
- **GPU model and memory**: CPU

### Describe the problem

I encountered an unexpected error at tf.contrib.seq2seq's BeamSearchDecoder, *when the beam width is smaller than number of vocabs*.
This is a part of `_beam_search_step` operation in `beam_search_decoder.py`:

```
  scores = _get_scores(
      log_probs=total_probs,
      sequence_lengths=new_prediction_lengths,
      length_penalty_weight=length_penalty_weight)

  time = ops.convert_to_tensor(time, name="time")
  # During the first time step we only consider the initial beam
  scores_flat = control_flow_ops.cond(
      time > 0,
      lambda: array_ops.reshape(scores, [batch_size, -1]),
      lambda: scores[:, 0])

  # Pick the next beams according to the specified successors function
  next_beam_scores, word_indices = nn_ops.top_k(scores_flat, k=beam_width)
  next_beam_scores.set_shape([static_batch_size, beam_width])
  word_indices.set_shape([static_batch_size, beam_width])
```

Since the shape of `scores` is `[batch_size, beam_width, vocab_size]`,  the shape of `scores_flat` is`[batch_size, vocab_size]` at time step 0. However, if `k < shape[-1]` in `nn.top_k` operation, it just throws `InvalidArgumentError: input must have at least k columns`. Thus the code just throws an error and dies.
I think code should be modified to handle cases where `vocab_size ** n < beam_width`, or at least throw an appropriate error message when the input is `vocab_size < beam_width`.

girving:
Nice catch!  Seems like the `top_k` operation should be shielded with a `min`.  Would be up for submitting a small PR?

shuuki4:
OK, I fixed the issue and made PR #9875

ttrouill:
I am currently interested in a case where `beam_width > vocab_size`, and run into the following error:

```
tensorflow.python.framework.errors_impl.InvalidArgumentError: Input to reshape is a tensor with 224 values, but the requested shape has 256
[...]

PR #9875, opened 2017-05-13
Fix contrib.seq2seq.BeamSearchDecoder #9855

shuuki4:
Fixed Issue 9855 by shielding the `top_k` op's input with `min`.
Also fixed wrong `log_probs `shape at `_get_scores` method docstring.

### DIFF
--- a/tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
+++ b/tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
@@ -497,17 +497,21 @@ def _beam_search_step(time, logits, beam_state, batch_size, beam_width,
 
   time = ops.convert_to_tensor(time, name="time")
   # During the first time step we only consider the initial beam
+  scores_shape = array_ops.shape(scores)
   scores_flat = control_flow_ops.cond(
       time > 0,
       lambda: array_ops.reshape(scores, [batch_size, -1]),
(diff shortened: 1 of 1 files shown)


chenghuige:
scores_flat.shape[-1].value) might be None for dynamic batch size ?

shuuki4:
@chenghuige That is true. It will be better to treat two values as tensor since top_k operation can handle `k` as tensor.
```
from tensorflow.python.ops import gen_math_ops

...

next_beam_size = gen_math_ops.minimum(
  ops.convert_to_tensor(
    beam_width, dtype=tf.int32, name="beam_width"),
  ops.convert_to_tensor(
    scores_flat.shape[-1], dtype=tf.int32, name="num_available"))
```
Will this code be OK?

rmlarsen:
@shuuki4 

Yes, that should work, but use math_ops.minimum, not gen_math_ops.minimum.

rmlarsen:

rmlarsen (approved):

rmlarsen:
@shuuki4 the unit test is failing, can you take a look?

shuuki4:
@rmlarsen Error was due to the unknown shape after cond operation. Fixed the problem.

rmlarsen:

rmlarsen:
Thanks for the contribution!

Issue #9858, opened 2017-05-12
Function decode_raw ignoring parameter little_endian.


jiefangxuanyan:
`little_endian` parameter in `decode_raw` takes no effect. I had a look at the C++ source code (`tensorflow/tensorflow/core/kernels/decode_raw_op.cc`), you are just using `reinterpret_cast` without any awareness of this parameter.

drpngx:
Would you be willing to send a PR?

jiefangxuanyan:
I hope PR #9876 could be good.

tensorflowbutler:
It has been 14 days with no activity and the `awaiting tensorflower` label was assigned. Please update the label and/or status accordingly.

jiefangxuanyan:
Fixed in #9876.

PR #9876, opened 2017-05-13
Fix issue #9858.

jiefangxuanyan:
Added some byte swapping code to solve endianness problem.

### DIFF
--- a/tensorflow/core/kernels/decode_raw_op.cc
+++ b/tensorflow/core/kernels/decode_raw_op.cc
@@ -70,7 +70,7 @@ class DecodeRawOp : public OpKernel {
     auto out = output_tensor->flat_inner_dims<T>();
     DCHECK_EQ(flat_in.size(), out.dimensions()[0]);
     T* out_data = out.data();
-    if (port::kLittleEndian == little_endian_) {
+    if (port::kLittleEndian == little_endian_ || sizeof(T) == 1) {
       for (int64 i = 0; i < flat_in.size(); ++i) {
         const T* in_data = reinterpret_cast<const T*>(flat_in(i).data());
         memcpy(out_data, in_data, str_size);
--- a/tensorflow/core/kernels/decode_raw_op.cc
+++ b/tensorflow/core/kernels/decode_raw_op.cc
@@ -70,17 +70,24 @@ class DecodeRawOp : public OpKernel {
     auto out = output_tensor->flat_inner_dims<T>();
     DCHECK_EQ(flat_in.size(), out.dimensions()[0]);
     T* out_data = out.data();
-    for (int64 i = 0; i < flat_in.size(); ++i) {
-      const T* in_data = reinterpret_cast<const T*>(flat_in(i).data());
-      memcpy(out_data, in_data, str_size);
-      if (port::kLittleEndian != little_endian_) {
-        char* out_data_bytes = reinterpret_cast<char*>(out_data);
-        for (char* p = out_data_bytes; p < out_data_bytes + str_size;
-             p += sizeof(T)) {
-          std::reverse(p, p + sizeof(T));
-        }
+    if (port::kLittleEndian == little_endian_) {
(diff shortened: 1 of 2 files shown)


jiefangxuanyan:
I signed it!

tensorflow/core/kernels/decode_raw_op.cc
```diff
@@ -73,6 +73,13 @@ class DecodeRawOp : public OpKernel {
     for (int64 i = 0; i < flat_in.size(); ++i) {
       const T* in_data = reinterpret_cast<const T*>(flat_in(i).data());
       memcpy(out_data, in_data, str_size);
+      if (port::kLittleEndian != little_endian_) {
```
rmlarsen: Instead if memcpy + byteswap, just use a custom copy loop for the big endian case, so you don't have to touch the data twice.

rmlarsen (changes requested):

jiefangxuanyan:
@rmlarsen like this?

rmlarsen (approved):

tensorflow/core/kernels/decode_raw_op.cc
```diff
@@ -70,10 +70,24 @@ class DecodeRawOp : public OpKernel {
...
-      const T* in_data = reinterpret_cast<const T*>(flat_in(i).data());
-      memcpy(out_data, in_data, str_size);
-      out_data += added_dim;
+    if (port::kLittleEndian == little_endian_) {
+      for (int64 i = 0; i < flat_in.size(); ++i) {
+        const T* in_data = reinterpret_cast<const T*>(flat_in(i).data());
+        memcpy(out_data, in_data, str_size);
+        out_data += added_dim;
+      }
+    } else {
```
rmlarsen: Please add unit test(s) such that both branches are tested.

rmlarsen (changes requested):
Please add unit test(s).

rmlarsen (changes requested):
Please add a unit test.

jiefangxuanyan:
I added a test case for endianness. By the way, according to the documentation (https://www.tensorflow.org/versions/master/api_docs/python/tf/decode_raw), I think this op should behave consistently on platforms with different endianness, so I removed the endianness test in an existing test case. If the original behavior in that test case is expected, maybe you should make changes on the documentation.

tensorflow/python/kernel_tests/decode_raw_op_test.py
```diff
@@ -54,18 +54,26 @@ def testToInt16(self):
...
         decode.eval(feed_dict={in_bytes: ["123", "456"]})
 
+  def testEndianness(self):
+    with self.test_session():
+      in_bytes = array_ops.placeholder(dtypes.string, shape=[None])
+      decode_le = parsing_ops.decode_raw(
+          in_bytes, out_type=dtypes.int32, little_endian=True)
+      decode_be = parsing_ops.decode_raw(
+          in_bytes, out_type=dtypes.int32, little_endian=False)
+      result = decode_le.eval(feed_dict={in_bytes: ["\x01\x02\x03\x04"]})
```
rmlarsen: Won't this depend on the endianness of the machine where this test is run? You can test the endianness of the machine here using the Python sys module, e.g.,
```
>>> import sys
>>> sys.byteorder
'little'
```

rmlarsen (changes requested):

tensorflow/python/kernel_tests/decode_raw_op_test.py
jiefangxuanyan: It won't. This decode op gets same result on machines with different endianness, just like the way these machines receive data from the network. In fact, platform endianness is considered in the C++ code (`if (port::kLittleEndian == little_endian_)`) to ensure consistency. As I said before, the old platform-dependent behavior in this test is contradictionary with the documentation.

rmlarsen (approved):

rmlarsen:

Issue #8572, opened 2017-03-21
Image Labeler Fails with Gifs?


tomasreimers:
The image_label example assumes that decode_gif (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/label_image/main.cc#L105) will return a 3-d tensor ([height, width, channels]) in the same way that decode_jpeg or decode_png would. However decode_dif returns a 4-d shape ([frames, height, width, channels], https://www.tensorflow.org/api_docs/python/tf/image/decode_gif). This will cause ResizeBilinear to fail.

This could be fixed by pulling the first frame from the gif, or by choosing to not expand dimensions on the gif (and classifying each frame).

drpngx:
I think you typically just decode a single channel by passing the channel argument to the `tf.image.decode_*` functions?

freedomtan:
Since only the single frame gif will be used in this label_image, frames should be 1. So we can use Squeeze() to remove it. See my PR, https://github.com/tensorflow/tensorflow/pull/9877

itsmeolivia:
Automatically closing due to lack of recent activity. Since this issue is old at this point, please reopen the issue if it still occurs when tried with the latest version of Tensorflow. Thank you.

PR #9877, opened 2017-05-13
gif decoder returns 4-D tensor, remove the first dim

freedomtan:
What returned by the gif decoder returns is a 4-D tensor [frames, height, width, channels]. Since only the single frame gif will be used in this label_image, frames should be 1. So we can use Squeeze() to remove it. 

The patch solve https://github.com/tensorflow/tensorflow/issues/8572

### DIFF
--- a/tensorflow/examples/label_image/main.cc
+++ b/tensorflow/examples/label_image/main.cc
@@ -30,6 +30,9 @@ limitations under the License.
 // the top of the main() function.
 //
 // The googlenet_graph.pb file included by default is created from Inception.
+//
+// Note that, for GIF inputs, to reuse existing code, only single-frame ones
+// are supported.
 
 #include <fstream>
 #include <utility>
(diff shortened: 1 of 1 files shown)


rmlarsen:

tensorflow/examples/label_image/main.cc
```diff
@@ -103,7 +103,10 @@ Status ReadTensorFromImageFile(const string& file_name, const int input_height,
     image_reader = DecodePng(root.WithOpName("png_reader"), file_reader,
                              DecodePng::Channels(wanted_channels));
   } else if (tensorflow::StringPiece(file_name).ends_with(".gif")) {
-    image_reader = DecodeGif(root.WithOpName("gif_reader"), file_reader);
+    // gif decoder returns 4-D tensor, remove the first dim
+    image_reader = Squeeze(root.WithOpName("squeeze_first_dim"),
```
rmlarsen: This would still fail if a multi-frame (animated) GIF image is input, but perhaps that's OK?

rmlarsen: @freedomtan Perhaps you can add a line to the documentation at the top to mention that animated GIFs are not expected/supported.

freedomtan: done

rmlarsen (approved):

rmlarsen:

Issue #9130, opened 2017-04-11
r.0.11 avro missing files 


arturovivas:
Hi

I am trying to install tf r0.11 in a Centos 7 without GPU support from sources. I have a similar problem like the https://github.com/tensorflow/tensorflow/issues/4312. I followed the recommended actions, yet I get the same error. BESIDES, I get the following (not described in the issue 4312) error:

`Error downloading from http://www-us.apache.org/dist/avro/avro-1.8.0/cpp/avro-cpp-1.8.0.tar.gz` 

I tryed to find this file using the web browser and I coudn't find it. The online repository doesn't have it anymore. Just the avro-1.8.1 and avro-1.7.7 are available.

Thanks
Regards
A

asimshankar:
0.11 is a very old release (prior to stability guarantees of 1.0). Is it possible for you to use 1.0?

That said, **please provide all information asked for in the new issue template**. Without that it is hard to be able to help (in particular, do provide the exact sequence of commands to reproduce the problem).

Thanks!

arturovivas:
I run my applications in serveral clusters. One of them, has only tf.11 installed, and since I don't have admin rights, I would like to keep the other servers with the tf.11. 

The secuence of commands was:

```
./configure
~/tensorflow ~/tensorflow
Please specify the location of python. [Default is /home/AVV/rocm/anaconda2/bin/python]: 
Do you wish to build TensorFlow with Google Cloud Platform support? [y/N] N
No Google Cloud Platform support will be enabled for TensorFlow
Do you wish to build TensorFlow with Hadoop File System support? [y/N] N
No Hadoop File System support will be enabled for TensorFlow
Found possible Python library paths:
  /home/AVV/rocm/anaconda2/lib/python2.7/site-packages
Please input the desired Python library path to use.  Default is [/home/AVV/rocm/anaconda2/lib/python2.7/site-packages]

/home/AVV/rocm/anaconda2/lib/python2.7/site-packages
Do you wish to build TensorFlow with GPU support? [y/N] N
No GPU support will be enabled for TensorFlow
Configuration finished
...................
INFO: Starting clean (this may take a while). Consider using --expunge_async if the clean takes more than several minutes.
..........
[...]

Issue #5143, opened 2016-10-23
Error building from source


mihail911:
Hi all, 

I'm trying to build TF from source so that I can run it in development mode and be able to expand the functionality of some of the modules, but am running into some snags with the bazel building. 
### Environment info

Operating System:

Mac OSX Yosemite

If installed from source, provide 
1. The commit hash (`git rev-parse HEAD`)
   
   ec7f37e40fedb23435bfb7e28668e5fa63ff52f3
2. The output of `bazel version`
   
   Build label: 0.3.2-2016-10-22 (@088bbc6)
   Build target: bazel-out/local-    opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
   Build time: Sun Oct 23 00:04:18 2016 (1477181058)
   Build timestamp: 1477181058
   Build timestamp as int: 1477181058

Running: `bazel build -c opt //tensorflow/tools/pip_package:build_pip_package --verbose_failures`
gives

```
WARNING: Sandboxed execution is not supported on your system and thus hermeticity of actions cannot be guaranteed. See http://bazel.io/docs/bazel-user-manual.html#sandboxing for more information. You can turn off this warning via --ignore_unsupported_sandboxing.
ERROR: /Users/mihaileric/Documents/Research/tensorflow/tensorflow/python/BUILD:1806:1: in cc_library rule //tensorflow/python:tf_session_helper: non-test target '//tensorflow/python:tf_session_helper' depends on testonly target '//tensorflow/python:construction_fails_op' and doesn't have testonly attribute set.
ERROR: Analysis of target '//tensorflow/tools/pip_package:build_pip_package' failed; build aborted.
INFO: Elapsed time: 0.848s
```

I'm having a lot of difficulty interpreting these error messages  and am not finding too much on Stack Overflow or other TF git threads. Any help would be greatly appreciated!

asimshankar:
@mihail911 : Thanks for the error report. There indeed seems to be an error in the BUILD file, tehre the `//tensorflow/python:tf_session_helper` target has a dependency on a library meant only for tests. That dependency shouldn't exist and I'll try to clear that up.

Amusingly though, this was not an error till very very recently. Since you're using bazel built from source, it includes the change https://github.com/bazelbuild/bazel/commit/5f026f96118e76c8a6357c61e5710e0bd1bf0f54 , which makes this an error.

[...]

PR #9879, opened 2017-05-13
Fix building r0.11 from source

vladfi1:
Fixes a bunch of issues I had building from source on r0.11, including #9130 and #5143. With these changes I was able to successfully build tensorflow with cuda support with bazel 0.4.5.

### DIFF
--- a/tensorflow/python/BUILD
+++ b/tensorflow/python/BUILD
@@ -1730,7 +1730,6 @@ tf_cuda_library(
     srcs = ["client/tf_session_helper.cc"],
     hdrs = ["client/tf_session_helper.h"],
     deps = [
-        ":construction_fails_op",
         ":numpy_lib",
         ":test_ops_kernels",
         "//tensorflow/c:c_api",
--- a/configure
+++ b/configure
@@ -7,7 +7,8 @@ popd > /dev/null
 
 function bazel_clean_and_fetch() {
   bazel clean --expunge
-  bazel fetch //tensorflow/...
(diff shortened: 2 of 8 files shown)


vladfi1:
@kashif @aj-michael are you ok with my cherry-picking your commits here?

tensorflow/python/BUILD
```diff
@@ -1730,7 +1730,6 @@ tf_cuda_library(
     srcs = ["client/tf_session_helper.cc"],
     hdrs = ["client/tf_session_helper.h"],
     deps = [
-        ":construction_fails_op",
```
jart: Why this?

jart:
This looks good, and thank you for the contribution, but help me understand why it's necessary? This branch is quite old. I'm wondering why no one else has run into problems with it.

vladfi1:
I suspect people are simply using the pre-built pip wheels (this included me until recently). These changes were only necessary to build from source.

tensorflow/python/BUILD
vladfi1: It breaks building with bazel 0.4.5. I believe the error is due to `:construction_fails_op` being marked as "test only".

gunan:
I am opposed to accepting PRs to old release branches.
The biggest reason is, bazel had a lot of backwards incompatible changes.
When fixing the build for latest bazel version included in the dockerfiles in this branch.

Unless anyone else has any objections, I will close this PR.

PR #9884, opened 2017-05-14
No longer use openjdk8 PPA in Dockerfile.devel

staticfloat:
Ubuntu 16.04 has an openjdk8 package, so use that instead of a PPA with a backported version, which was a holdover from when the base image was ubuntu 14.04

### DIFF
--- a/tensorflow/tools/docker/Dockerfile.devel
+++ b/tensorflow/tools/docker/Dockerfile.devel
@@ -17,6 +17,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         unzip \
         zip \
         zlib1g-dev \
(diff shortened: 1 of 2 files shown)


caisq:
@staticfloat Thank you for the PR. The changes look good to me. But can you make the same changes in this file as well?
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docker/Dockerfile.devel-gpu

staticfloat:
@caisq I've done so.  Feel free to squash the two commits together when you merge, or if you prefer, I can do that when I get back to a computer with command-line `git`.

caisq (approved):

PR #9887, opened 2017-05-14
Allow tensor as iou_threshold parameter to tf.image.non_max_suppression.

xodus7:
I have a use case where dynamically changing the iou_threshold parameter on a frozen graph is required. I'm not 100% positive this doesn't violate the backwards compatibility rules, and it does make the iou_threshold a required parameter. Feedback from the maintainers would be appreciated.

### DIFF
--- a/tensorflow/core/kernels/non_max_suppression_op.cc
+++ b/tensorflow/core/kernels/non_max_suppression_op.cc
@@ -33,6 +33,7 @@ limitations under the License.
 #include "tensorflow/core/platform/logging.h"
 
 namespace tensorflow {
+namespace {
 
 typedef Eigen::ThreadPoolDevice CPUDevice;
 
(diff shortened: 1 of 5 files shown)


rmlarsen:
Since this changes the functionality of the op, I believe you have to do this by 

1) adding a new kernel named NonMaxSuppressionOpV2 
2) migrate all producers (e.g. Python wrappers) to use that kernel instead.

See this for more details:
https://www.tensorflow.org/programmers_guide/data_versions

vrv:
(Though you have to wait 3 weeks after checking in the new Op/OpKernel before changing the python code).

xodus7:
Ok, I moved my changes into a new NonMaxSuppressionV2Op as suggested.

tensorflow/core/kernels/non_max_suppression_op.cc
```diff
@@ -89,6 +89,61 @@ static inline float ComputeIOU(typename TTypes<float, 2>::ConstTensor boxes,
   return intersection_area / (area_i + area_j - intersection_area);
 }
 
+static void DoNonMaxSuppressionOp(OpKernelContext* context,
```
rmlarsen: Nit: Could you put this in anonymous namespace instead of making it static?

rmlarsen (changes requested):
Looks good! Just one nit.

tensorflow/core/kernels/non_max_suppression_op.cc
xodus7: Done!

rmlarsen (approved):

rmlarsen:

Issue #8310, opened 2017-03-11
XLA/AOT on Windows build?


neale:
I know windows is low on the priority list (I don't blame you), but will there be support in the future for AOT compilation for Windows, or is it possible now in theory?

### What related GitHub issues or StackOverflow threads have you found by searching the web for your problem?
There aren't a lot of questions anywhere about this yet :/

### Environment info
Operating System:
Windows 10

### Installed version of CUDA and cuDNN: 
None

### Installation 
I installed from cmake using the tutorial here and it worked fine, the stock mnist example ran and I couldn't find a fault with the installation. 

I also tried building with the windows Bazel build. I didn't have much success with this (lots of errors on configure) but if I know it can work than I can put for effort here. 

### What other attempted solutions have you tried?
I attemped the JIT example [here](https://www.tensorflow.org/versions/master/experimental/xla/jit) but I got the same output using XLA and not using XLA. Checking chrome://tracing revealed no XLA startup. This is where I assumed that the cmake build does not include tensorflow/compiler/xla, indeed I found no corresponding MSVC build files. 

Attempting to build AOT binaries with bazel didn't work because configure had not been run. I know you can build these binaries with alternate build systems (in the works?) but there's no docs on how to do that. 

I just want to know if I should even bother trying while this part of the project is so young. I should add that the JIT works very well for me on ubuntu, in places where it should work. 

If I can compile tf_library and get the tfcompile tool then I would appreciate some pointers on how to get there. I don't know how productive it is to spend time decoding bazel build files and try to construct my own alternate build the replicates it.

### Logs or other output that would be helpful
Log of JIT example output
```
$ TF_XLA_FLAGS=--xla_generate_hlo_graph=.* python mnist_softmax_xla.py
Extracting /tmp/tensorflow/mnist/input_data\train-images-idx3-ubyte.gz
Extracting /tmp/tensorflow/mnist/input_data\train-labels-idx1-ubyte.gz
Extracting /tmp/tensorflow/mnist/input_data\t10k-images-idx3-ubyte.gz
Extracting /tmp/tensorflow/mnist/input_data\t10k-labels-idx1-ubyte.gz
0.9179
```
`bazel ./configure` output
```
$ ./configure
Please specify the location of python. [Default is /c/Users//AppData/Local/Programs/Python/Python35/python]:
Please specify optimization flags to use during compilation when bazel option "--config=opt" is specified [Default is -march=native]:
[...]

PR #9908, opened 2017-05-15
Make possible to use static libraries generated by tfcompile in MSVC.

leandro-gracia-gil:
Currently it is possible to generate MSVC-compatible static libraries from a
Linux build of tfcompile using the target triple "x86_64-pc-windows-msvc".

However, just linking such library does not work out-of-the-box because of a
trivial class -> struct issue that becomes a compiler error in MSVC and because
multiple other symbols are missing.

The files defining these symbols must be manually built into the MSVC project,
but many have minor incompatibilities that prevent them to be built.

This patch makes a few minor modifications to these files, making it possible
to link and run Tensorflow graphs generated with tfcompile in MSVC in Windows.

The files that might need to be built into the MSVC project are as follows:

1. For symbols required by the generated headers.
 - tensorflow/compiler/aot/runtime.cc
 - tensorflow/compiler/xla/executable_run_options.cc

2. For symbols that provide custom XLA operations.
 - tensorflow/compiler/tf2xla/kernels/gather_op_kernel_float_int32.cc
 - tensorflow/compiler/tf2xla/kernels/gather_op_kernel_float_int64.cc
 - tensorflow/compiler/tf2xla/kernels/index_ops_kernel_argmax_float_1d.cc
 - tensorflow/compiler/tf2xla/kernels/index_ops_kernel_argmax_float_2d.cc

3. For symbols required by SimpleResolver in xla/service/cpu/simple_orc_jit.cc.
 - tensorflow/compiler/xla/service/cpu/runtime_matmul.cc
 - tensorflow/compiler/xla/service/cpu/runtime_single_threaded_matmul.cc
 - tensorflow/compiler/xla/service/cpu/runtime_conv2d.cc
 - tensorflow/compiler/xla/service/cpu/runtime_single_threaded_conv2d.cc
 - tensorflow/compiler/xla/service/cpu/cpu_runtime.cc

Additionally, the following files might also required by SimpleResolver, but are
not included in this patch because there is no simple alternative to
__attribute__((vector_size(x))) and __attribute__((weak)) in MSVC.

 - tensorflow/compiler/xla/service/cpu/cpu_runtime_sse4_1.cc
 - tensorflow/compiler/xla/service/cpu/cpu_runtime_avx.cc

See https://github.com/tensorflow/tensorflow/issues/8310 for more information.

### DIFF
--- a/tensorflow/compiler/aot/runtime.cc
+++ b/tensorflow/compiler/aot/runtime.cc
@@ -31,6 +31,8 @@ namespace {
 inline void* aligned_malloc(size_t size, int minimum_alignment) {
 #if defined(__ANDROID__) || defined(OS_ANDROID) || defined(OS_CYGWIN)
   return memalign(minimum_alignment, size);
+#elif defined(COMPILER_MSVC)
+  return _aligned_malloc(size, minimum_alignment);
 #else  // !__ANDROID__ && !OS_ANDROID && !OS_CYGWIN
   void* ptr = nullptr;
   // posix_memalign requires that the requested alignment be at least
@@ -45,7 +47,13 @@ inline void* aligned_malloc(size_t size, int minimum_alignment) {
 #endif
 }
 
-inline void aligned_free(void* aligned_memory) { free(aligned_memory); }
+inline void aligned_free(void* aligned_memory) {
+#if defined(COMPILER_MSVC)
+  _aligned_free(aligned_memory);
+#else
+  free(aligned_memory);
+#endif
+}
 
 size_t align_to(size_t n, size_t align) {
   return (((n - 1) / align) + 1) * align;
--- a/tensorflow/compiler/aot/codegen.cc
+++ b/tensorflow/compiler/aot/codegen.cc
@@ -365,7 +365,7 @@ Status GenerateHeader(const HeaderOpts& opts, const Config& config,
 #include "tensorflow/core/platform/macros.h"
 #include "tensorflow/core/platform/types.h"
 
-namespace Eigen { class ThreadPoolDevice; }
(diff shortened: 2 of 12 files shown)


tensorflow/compiler/aot/runtime.cc
```diff
@@ -45,7 +47,11 @@ inline void* aligned_malloc(size_t size, int minimum_alignment) {
 #endif
 }
 
+#if defined(COMPILER_MSVC)
+inline void aligned_free(void* aligned_memory) { _aligned_free(aligned_memory); }
+#else
 inline void aligned_free(void* aligned_memory) { free(aligned_memory); }
```
tatatodd: Small style nit: can you write the code this way instead:

```c++
inline void aligned_free(void* aligned_memory) {
#if defined(COMPILER_MSVC)
  _aligned_free(aligned_memory);
#else
  free(aligned_memory);
#endif
}
```

tatatodd (changes requested):
Looks great, just a small style nit.  Thanks for the collection of fixes!  :)

leandro-gracia-gil:
Changes done. Sorry, I force-pushed them without knowing it would affect the review.

leandro-gracia-gil:
Also fixed some tabs in this pull request that MSVC put into a couple of the files. Whitespace should be ok now.

tatatodd:
Jenkins, test this please.

rmlarsen:
@tatatodd can you approve this?

tatatodd (approved):

tatatodd:
Thanks again for the fixes, and details like the whitespace!  :)

PR #9909, opened 2017-05-15
Fixed _linear(.) to use *batch* matrix multiplication.

KristianHolsheimer:
In the `_linear` function (from the `tensorflow.contrib.rnn.python.ops.core_rnn_cell_impl` module), there's relatively subtle bug. The reason that this bug is subtle is that it only affects tensors used by `_linear` internally.

For future reference, the signature and docstring of `_linear` are:
```python
def _linear(args, output_size, bias, bias_start=0.0):
  """Linear map: sum_i(args[i] * W[i]), where W[i] is a variable.

  Args:
    args: a 2D Tensor or a list of 2D, batch x n, Tensors.
    output_size: int, second dimension of W[i].
    bias: boolean, whether to add a bias term or not.
    bias_start: starting value to initialize the bias; 0 by default.

  Returns:
    A 2D Tensor with shape [batch x output_size] equal to
    sum_i(args[i] * W[i]), where W[i]s are newly created matrices.

  Raises:
    ValueError: if some of the arguments has unspecified or wrong shape.
  """
```

**The Problem.**
From the docsting of `_linear`, the expected behavior of the function is that it should return an output `y` such that `y = sum_i(args[i] * W[i])`, indicating a batch-wise `matmul` (formerly known as `batch_matmul`). This wasn't actually how `_linear` was implemented, though. Instead, all `args` were concatenated and the weights were chosen to have rank 2, i.e.
```python
x = concat(args, axis=1)  # x.shape == (batch_size, total_arg_size)
w = Variable(...)         # w.shape == (total_arg_size, output_size)
y = matmul(x, w)          # y.shape == (batch_size, output_size)
```
Now, the main problem here is that **`w` is not block-diagonal**, which means that it contains more entries than it should. There are two problems with this:
1. The **number of weights is too large** by a factor of `num_args`.
2. The extra "off-diagonal" weights yield **spurious connections** in the net architecture of e.g. LSTMCell.

**The Solution.**
The solution is to use batch-wise `matmul` instead. In order to do this, we need `x` and `w` to be rank-3 (rather than rank-2) tensors, e.g.
```python
x = stack(args, axis=0)                    # x.shape == (num_args, batch_size, input_size)
w = Variable(...)                          # w.shape == (num_args, input_size, output_size / num_args)
y = matmul(x, w)                           # y.shape == (num_args, batch_size, output_size / num_args)
y = transpose(y, [1, 0, 2])                # y.shape == (batch_size, num_args, output_size / num_args)
y = reshape(y, [batch_size, output_size])  # y.shape == (batch_size, output_size)
```

**Tests.**
I ran `tensorflow/contrib/rnn/python/kernel_tests/core_rnn_cell_impl_test.py`, but it was failing all over the place. I expected this, because the `_linear` is used in many of the rnn cell classes. The tests were failing in more places than just the place that covered `_linear`. Let me know if I need to update the unit tests too. Also, the change in the PR put somewhat more stringent constraints on the shapes of the input `args`, i.e. they must all be the same. In the previous implementation, only the axis-0 sizes (`batch_size`) needed to be the same. We could use padding and slicing if we need to be able to handle different axis-1 sizes (`input_size`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ node_modules
 __pycache__
 *.swp
 .vscode/
+.sublime-*
--- a/tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
+++ b/tensorflow/contrib/grid_rnn/python/kernel_tests/grid_rnn_test.py
@@ -63,10 +63,10 @@ class GridRNNCellTest(test.TestCase):
         self.assertEqual(res_s[1].c.shape, (1, 2))
         self.assertEqual(res_s[1].h.shape, (1, 2))
 
-        self.assertAllClose(res_g, ([[0.36617181, 0.36617181]],))
+        self.assertAllClose(res_g, ([[0.27813572, 0.27813572]],))
         self.assertAllClose(
-            res_s, (([[0.71053141, 0.71053141]], [[0.36617181, 0.36617181]]),
-                    ([[0.72320831, 0.80555487]], [[0.39102408, 0.42150158]])))
+            res_s, (([[0.58389962, 0.58389962]], [[0.27813572, 0.27813572]]),
+                    ([[0.52472770, 0.60331118]], [[0.27650252, 0.30985516]])))
 
         # emulate a loop through the input sequence,
         # where we call cell() multiple times
@@ -86,10 +86,10 @@ class GridRNNCellTest(test.TestCase):
         self.assertEqual(res_s2[0].h.shape, (1, 2))
         self.assertEqual(res_s2[1].c.shape, (1, 2))
         self.assertEqual(res_s2[1].h.shape, (1, 2))
-        self.assertAllClose(res_g2[0], [[0.58847463, 0.58847463]])
+        self.assertAllClose(res_g2[0], [[0.44053382, 0.44053382]])
         self.assertAllClose(
-            res_s2, (([[1.40469193, 1.40469193]], [[0.58847463, 0.58847463]]),
-                     ([[0.97726452, 1.04626071]], [[0.4927212, 0.51137757]])))
+            res_s2, (([[1.1822226, 1.18222260]], [[0.44053382, 0.44053382]]),
+                     ([[0.6710391, 0.73025036]], [[0.30998221, 0.32985979]])))
 
   def testGrid2BasicLSTMCellTied(self):
     with self.test_session(use_gpu=False) as sess:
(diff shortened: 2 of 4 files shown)


drpngx:
@KristianHolsheimer could you rebase and push again?

KristianHolsheimer:
Hi @drpngx, I just rebased

drpngx:
Jenkins, test this please.

ebrevdo:
the documentation of `_linear` is not correct; but the implementation *is*.  we want to concatenate args along the columns, create a variable that's [num_columns, output_size], and do a matmul against this variable.

if you'd like to clean this up; please update the documentation of `_linear` in a separate PR.  anyway, it's a hidden function so the documentation shouldn't really break anyone (no one outside of tensorflow core is supposed to use it).

KristianHolsheimer:
@ebrevdo The docstring isn't the thing that's wrong here. The reason is that it's used by e.g. `LSTMCell`, which requires the behavior mentioned in the docsting, *not* the implementation that you suggest. If we keep the implementation like this, we are diverting from the standard LSTM architecture in two ways:

1. We use `len(args)` more weights
2. We allow for mixing between input and hidden state at each time step

I suspect that the real reason we're doing this concatenation thing is (time) optimization. So, if we're willing to have a non-standard definition of LSTM in tensorflow, we could decide to leave it like this, but *we need to be honest and make this choice consciously*, not as an unintended by-product of a low-level optimization trick.

For the record, my choice would be to use the definition of LSTM that is used by the rest of the community (as per this pull request).

KristianHolsheimer:
Also, I'm currently updating the unit tests that are failing because of this. As you can imagine, it touches quite a bit of code, so give a little bit of time.

ebrevdo:
My reading of the equations is that the input and hidden state should be
mixed in this way at each time step

https://en.m.wikipedia.org/wiki/Long_short-term_memory

We also compare to cudnn implementation here and ensure we have identical
values within machine precision:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cudnn_rnn/python/kernel_tests/cudnn_rnn_ops_test.py

Function testCudnnCompatibleRnnCells (the lstm compatible cell is just a
subclass of LSTMCell with forget_bias=0).

ebrevdo:
Perhaps it's confusing because we are right multiplying by the parameters,
unlike the typical format?  Tf data is row major and batch is the major
dimension, which is why everything is transposed.

KristianHolsheimer:
Okay thanks, I'll have another look at it and come back to you.

KristianHolsheimer:
Wow, you're right, the same happens in `cudnn_rnn_ops`.  Okay, this is a bit bigger than I initially thought. I'll have to dig into it later this week.

Thanks again @ebrevdo

KristianHolsheimer:
Okay, I did look into it in more detail now. I agree with your earlier comment. Now that I see it, I don't understand how I was confused about it...

Thanks @ebrevdo

PR #9913, opened 2017-05-15
[XLA] Add F16 support to the Literal protobuf and LiteralUtils class.

DavidNorman:
No support has been added to any public back-end, however the unit tests demonstrate that the literals can store and retrieve data correctly.

[Note: Resize needed to be moved so it could be used in a subsequent function]

### DIFF
--- a/tensorflow/compiler/xla/literal_util.cc
+++ b/tensorflow/compiler/xla/literal_util.cc
@@ -154,6 +154,9 @@ template <typename T>
     case F64:
       return CopyRange<double>(src_literal, src_base, dest_literal, dest_base,
                                copy_size);
+    case F16:
+      return CopyRange<half>(src_literal, src_base, dest_literal, dest_base,
+                             copy_size);
     case PRED:
       return CopyRange<bool>(src_literal, src_base, dest_literal, dest_base,
                              copy_size);
@@ -188,7 +191,7 @@ template <typename T>
     case U16:
       LOG(FATAL) << "u16/s16 literals not yet implemented";
     case F16:
-      LOG(FATAL) << "f16 literals not yet implemented";
+      return *LiteralUtil::CreateR0<half>((half)0.0f);
     case TUPLE:
       LOG(FATAL) << "tuple element type cannot take on value of 0";
     case OPAQUE:
@@ -222,7 +225,7 @@ template <typename T>
     case U16:
       LOG(FATAL) << "u16/s16 literals not yet implemented";
     case F16:
-      LOG(FATAL) << "f16 literals not yet implemented";
+      return *LiteralUtil::CreateR0<half>((half)1.0f);
     case TUPLE:
       LOG(FATAL) << "tuple element type cannot take on value of 1";
     case OPAQUE:
@@ -258,7 +261,8 @@ template <typename T>
     case U16:
       LOG(FATAL) << "u16/s16 literals not yet implemented";
     case F16:
-      LOG(FATAL) << "f16 literals not yet implemented";
+      return *LiteralUtil::CreateR0<half>((half)
+              -std::numeric_limits<float>::infinity());
     case TUPLE:
       LOG(FATAL) << "tuple element type has no minimum value";
     case OPAQUE:
@@ -294,7 +298,8 @@ template <typename T>
     case U16:
       LOG(FATAL) << "u16/s16 literals not yet implemented";
     case F16:
-      LOG(FATAL) << "f16 literals not yet implemented";
+      return *LiteralUtil::CreateR0<half>((half)
+              std::numeric_limits<float>::infinity());
     case TUPLE:
       LOG(FATAL) << "tuple element type has no maximum value";
     case OPAQUE:
@@ -498,6 +503,8 @@ template <typename T>
       return tensorflow::strings::StrCat(Get<float>(literal, multi_index));
     case F64:
       return tensorflow::strings::StrCat(Get<double>(literal, multi_index));
+    case F16:
+      return tensorflow::strings::StrCat(Get<half>(literal, multi_index));
     default:
       return tensorflow::strings::StrCat(
           "[", PrimitiveType_Name(literal.shape().element_type()), "]");
@@ -652,6 +659,8 @@ template <typename T>
       return reinterpret_cast<const void*>(literal.f32s().data());
     case F64:
       return reinterpret_cast<const void*>(literal.f64s().data());
+    case F16:
+      return reinterpret_cast<const void*>(literal.f16s().data());
     default:
       LOG(FATAL) << "primitive type not supported in literals: "
                  << PrimitiveType_Name(literal.shape().element_type());
@@ -691,6 +700,8 @@ template <typename T>
       break;
     case F64:
       Resize<double>(num_elements, 0, literal);
+    case F16:
+      Resize<half>(num_elements, (half)0.0f, literal);
       break;
     default:
       LOG(FATAL) << "primitive type not supported in literals: "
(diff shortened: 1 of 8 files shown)


tensorflow/compiler/xla/literal_util_test.cc
```diff
@@ -730,6 +766,38 @@ TEST_F(LiteralUtilTest, CopyScalars) {
...
+  EXPECT_EQ(d1[7], 0);
+  EXPECT_EQ(LiteralUtil::InternalData(*l1),
+            LiteralUtil::MutableInternalData(l1));
+
+  half h1 = (half)1.0f;
+  half h2 = (half)2.0f;
+  auto m2 = LiteralUtil::CreateR2<half>({{h1, h2}, {h2, h1}});
+  Literal* l2 = m2.get();
+  const char* d2 = (const char*)LiteralUtil::InternalData(*l2);
+  EXPECT_EQ(d2[0], 0);
```
hawkinsp: Is the intent here to check the endian-ness of the encoding? (Sounds good, but probably deserves a comment.)

tensorflow/compiler/xla/literal_util.cc
```diff
@@ -917,6 +933,16 @@ LiteralUtil::GetMutableArraySlice(Literal* literal) {
 }
 
 template <>
+/* static */ tensorflow::gtl::MutableArraySlice<half>
```
hawkinsp: It occurs to me that there is a latent endian-ness bug here. If you were to use this code on a big-endian architecture, the layout of the encoded bytes would not match the byte layout in memory.

Not sure what to do here. I suspect for now, leave a // TODO: add big-endian support and move on.

I suspect to fix this right, we should add logic to perform byteswapping on serialization or deserialization of a half literal.

tensorflow/compiler/xla/xla_data.proto
```diff
@@ -286,6 +286,7 @@ message Literal {
   repeated float f32s = 8;
   repeated double f64s = 9;
   repeated Literal tuple_literals = 10;
+  bytes f16s = 11;
```
hawkinsp: I would add a comment stating that the encoding is in little-endian order.

hawkinsp (changes requested):
Thanks for the contribution! Looks pretty good on the whole, but I have a few minor comments.

tensorflow/compiler/xla/literal_util.cc
DavidNorman: or better still, add a 16 bit uint support in protobuf (or even a 16 bit float support in protobuf).  I decided to open an Issue with Protobuf:
https://github.com/google/protobuf/issues/3091

DavidNorman: it's only a problem if the literals are sent over the network of course.  Is that ever true?

tensorflow/compiler/xla/literal_util.cc
```diff
@@ -1015,6 +1050,8 @@ static bool AllElementsEqualValue(const Literal& literal, NativeT value) {
       return AllElementsEqualValue<float>(literal, value);
     case F64:
       return AllElementsEqualValue<double>(literal, value);
+    case F16:
+      return AllElementsEqualValue<half>(literal, (half)value);
```
DavidNorman: strangely this function is checking that they are all equal to an int8.  all the native types support transparent casting between int8 and themselves, but Eigen::half does not.  Shame - it would look prettier.

tensorflow/compiler/xla/literal_util_test.cc
DavidNorman: Just a check that getting data via the InternalData and MutableInternalData work as expected.    It does veryify the endianess, but also that there are no weird things going on with 'bytes'.  I don't trust the protobuf 'bytes' as it is implemented as a std::string, which seems like a weird container to choose when std::vector<uint8> would work.

tensorflow/compiler/xla/xla_data.proto
DavidNorman: will do

tensorflow/compiler/xla/literal_util_test.cc
```diff
@@ -373,6 +376,15 @@ TEST_F(LiteralUtilTest, IsAll) {
   EXPECT_FALSE(
       LiteralUtil::IsAll(*LiteralUtil::CreateR2<uint64>({{9, 8}, {8, 8}}), 8));
 
+  half h8 = (half)8.0f;
+  half h9 = (half)9.0f;
+  EXPECT_TRUE(
```
DavidNorman: oops - missed this one - stand by....

DavidNorman:
@hawkinsp  It's all yours.

tensorflow/compiler/xla/literal_util.cc
hawkinsp: Literals are certainly stored on disk. I don't know whether you have discovered them, but you can dump computations to disk (TF_XLA_FLAGS=--xla_dump_computations_to=/tmp/somewhere) and there are utilities to visualize and replay dumped computations in tensorflow/compiler/xla/tools. So the potential for the problem is there, at least --- you can copy the dumped computations between machines.

I agree that it would be nice if protobuf added a 16-bit type, but it won't happen immediately even if they agree.

tensorflow/compiler/xla/literal_util.cc
```diff
@@ -222,7 +225,7 @@ template <typename T>
     case U16:
       LOG(FATAL) << "u16/s16 literals not yet implemented";
     case F16:
-      LOG(FATAL) << "f16 literals not yet implemented";
+      return *LiteralUtil::CreateR0<half>((half)1.0f);
```
hawkinsp: static_cast<half>

tensorflow/compiler/xla/literal_util.cc
```diff
@@ -258,7 +261,8 @@ template <typename T>
     case U16:
       LOG(FATAL) << "u16/s16 literals not yet implemented";
     case F16:
-      LOG(FATAL) << "f16 literals not yet implemented";
+      return *LiteralUtil::CreateR0<half>((half)
```
hawkinsp: static_cast<half>

tensorflow/compiler/xla/literal_util.cc
```diff
@@ -294,7 +298,8 @@ template <typename T>
...
```
hawkinsp: static_cast<half>

tensorflow/compiler/xla/literal_util.cc
```diff
@@ -691,6 +700,8 @@ template <typename T>
       break;
     case F64:
       Resize<double>(num_elements, 0, literal);
+    case F16:
+      Resize<half>(num_elements, (half)0.0f, literal);
```
hawkinsp: static_cast<half>

tensorflow/compiler/xla/literal_util.cc
```diff
@@ -1128,4 +1172,15 @@ template <>
   literal->mutable_f64s()->Resize(num_elements, value);
 }
 
+template <>
+/* static */ void LiteralUtil::Resize<half>(int64 num_elements, half value,
+                                            Literal* literal) {
+  CHECK_EQ(ShapeUtil::ElementsIn(literal->shape()), num_elements);
+  literal->mutable_f16s()->resize(num_elements * 2);
```
hawkinsp: sizeof(half)

hawkinsp (changes requested):
A couple more casts and constants you missed. Otherwise, it looks good!

hawkinsp (approved):
Looks great! Thanks for the contribution!

Jenkins, test this please.

rmlarsen:

PR #9941, opened 2017-05-16
Replaced deprecated op_scope call with name_scope

krivard:
This was flooding our experiment logs, so I fixed it. Should be pretty straightforward?


rmlarsen (approved):

rmlarsen:

PR #9942, opened 2017-05-16
Merged performance docs.

tfboyd:


tensorflow/docs_src/performance/leftnav_files
```diff
@@ -1,9 +1,12 @@
 performance_guide.md
+performance_models.md
+benchmarks.md
+quantization.md
+>>>
```
rmlarsen: is this intentional?

tfboyd: Yes, it puts a space.  It will look just like the link below.  XLA ends up broken out a little.  I would rather have bold sections or something but this is better, in my opinion, than having it all smashed together.  XLA is a section not really individual pages and this is a step back in the direction.  The step actually already took place weeks ago.  But if you do not merge the website changes back into the last release there is a possibility that they will get erased if the site is regenerated.  Wolff wants to regen the site so i needed to push these changes back into the release branch.  I gambled it would not need to be done and last.  The best approach is write something and wait a couple months for the site to get updated or possibly just edit in github vs. google.  My aggravation is being told how easy this process is but it does not seem easy to me, which is ok and I do not mind I just do not like the ruse.

https://www.tensorflow.org/performance/

rmlarsen (approved):

rmlarsen:

PR #9950, opened 2017-05-17
New reader for LMDB databases

bowang:
Implement a reader op to access records stored in an LMDB database.
This facilitates Caffe users to switch to TensorFlow without rebuilding their (potentially very large) LMDB datasets.


tensorflow/core/kernels/lmdb_reader_op.cc
```diff
@@ -0,0 +1,133 @@
...
+        return Status::OK();
+      }
+    }
+    else {
+      if (Seek(MDB_NEXT) == false) {
+        *at_end = true;
+        return Status::OK();
+      }
+    }
+    *key = string(static_cast<const char*>(mdb_key_.mv_data), mdb_key_.mv_size);
```
rmlarsen: Does LMDB handle endianness conversions automatically?

rmlarsen (changes requested):
Looks nice! A few comments and questions.

tensorflow/core/kernels/lmdb_reader_op.cc
```diff
@@ -0,0 +1,133 @@
...
+  Status OnWorkStartedLocked() override {
+    MDB_CHECK(mdb_env_create(&mdb_env_));
+    int flags = MDB_RDONLY | MDB_NOTLS;
+
+    // Check if the LMDB filename is actually a file instead of a directory.
+    // If so, set appropriate flags so we can open it.
+    struct stat source_stat;
+    if (stat(current_work().c_str(), &source_stat) == 0 &&
+        (source_stat.st_mode & S_IFREG)) {
+      flags |= MDB_NOSUBDIR;
```
bowang: Good point. A test case for reading from the folder is added. When specifying a directory, LMDB expects a file with the name `data.mdb` in it. I updated the test database file accordingly.

tensorflow/core/kernels/lmdb_reader_op.cc
bowang: Nope. The format of both key and value are opaque from LMDB's perspective. Users serialize key and value into strings before passing them to LMDB. Since byte stream is endian agnostic, LMDB just stores whatever is given. It is up to users' deserializer to handle endianness.

tensorflow/python/kernel_tests/reader_ops_test.py
```diff
@@ -858,5 +858,30 @@ def _RunSessionAndSave(sess, args, output):
...
+      queue.close().run()
+      for i in range(10):
+        k, v = sess.run([key, value])
+        self.assertEqual(k, str(i))
+        self.assertEqual(v, str(chr(ord('a') + i)))
+
+      with self.assertRaisesOpError("is closed and has insufficient elements "
+                                    "\\(requested 1, current size 0\\)"):
+        k, v = sess.run([key, value])
+
```
bowang: When opening a database file, LMDB creates a lock file `lock.mdb` in the same folder to prevent that database being opened more than once. This unfortunately prevents us to enqueue the database file multiple times.

bowang (commented):
Thanks for the prompt review and detailed comments! I added a testcase to load LMDB from a folder and replied on the comments.

tensorflow/python/kernel_tests/reader_ops_test.py
```diff
@@ -858,5 +858,40 @@ def _RunSessionAndSave(sess, args, output):
...
+      queue = data_flow_ops.FIFOQueue(99, [dtypes.string], shapes=())
+      key, value = reader.read(queue)
+
+      queue.enqueue([path]).run()
+      queue.close().run()
+      for i in range(10):
+        k, v = sess.run([key, value])
+        self.assertEqual(k, str(i))
+        self.assertEqual(v, str(chr(ord('a') + i)))
+
```
rmlarsen: Please test that you get an assertion when reading past the end, see, e.g., https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/reader_ops_test.py#L483

bowang: Thanks for pointing it out! Just added.

rmlarsen:

bowang:

rmlarsen:

rmlarsen:

tensorflow/workspace.bzl
```diff
@@ -508,6 +508,14 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
   )
 
   native.new_http_archive(
+    name = "lmdb",
```
rmlarsen: Can you add a license here?

rmlarsen (changes requested):

tensorflow/workspace.bzl
bowang: I was looking at Bazel manual and `new_http_archive` doesn't seem to have a field called `licenses`. Shall I switch to use `web_library_external` which has the license field? I've added a license in `third_party/lmdb.BUILD` though

rmlarsen: @jart is web_library_external the right thing to use here?

bowang: Any update on this?

third_party/lmdb.BUILD
```diff
@@ -0,0 +1,29 @@
...
+    srcs = [
+        "libraries/liblmdb/mdb.c",
+        "libraries/liblmdb/midl.c",
+    ],
+    hdrs = [
+        "libraries/liblmdb/lmdb.h",
+        "libraries/liblmdb/midl.h",
+    ],
+    includes = ["libraries/liblmdb/"],
+    copts = [
```
jart: [optional] Consider `copts = ["-w"],`

tensorflow/workspace.bzl
jart: What has been done appears to be correct. The `licenses(["notice"])` statement is in the associated lmdb.BUILD file. Please note that `web_library_external` basically uses its licenses attribute to generate that statement, when it creates a BUILD file.

jart (changes requested):

tensorflow/workspace.bzl
```diff
@@ -508,6 +508,14 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
...
+    url = "https://github.com/LMDB/lmdb/archive/LMDB_0.9.19.tar.gz",
+    sha256 = "108532fb94c6f227558d45be3f3347b52539f0f58290a7bb31ec06c462d05326",
+    strip_prefix = "lmdb-LMDB_0.9.19",
```
bowang: Fixed

tensorflow/workspace.bzl
```diff
@@ -508,6 +508,14 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
   )
 
   native.new_http_archive(
+    name = "lmdb",
+    url = "https://github.com/LMDB/lmdb/archive/LMDB_0.9.19.tar.gz",
```
bowang: Fixed

third_party/lmdb.BUILD
bowang: Fixed

tensorflow/workspace.bzl
bowang: Thanks for looking into this, @jart! Not sure why Jenkins complains about license.
```
FAIL: mismatch in packaged licenses and external dependencies
Missing the licenses for the following external dependencies:
@lmdb//
```

rmlarsen:

tensorflow/workspace.bzl
rmlarsen: @bowang You need to add a line "@lmdb//:LICENSE" here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/BUILD

(I'll make sure it gets added in the appropriate place internally to make it stick.)

rmlarsen: @bowang It looks like you also need it to add in a couple of build rules here:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/lib_package/BUILD

bowang: Thanks for pointing it out @rmlarsen ! Just added.

rmlarsen:

rmlarsen:

rmlarsen:
@bowang Your tests are filing because the key is a bytearray, not a string. PErhaps you can re-use the _ExpectRead method here: https://github.com/bowang/tensorflow/blob/89297b6c00bd00181c1522cf84e84c5e7ec10baa/tensorflow/python/kernel_tests/reader_ops_test.py#L237

bowang:
@rmlarsen Thanks for the help! Do you have any idea why cmake cannot find `lmdb.h` while others can?
https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu-cmake/2960/console

third_party/lmdb.BUILD
```diff
@@ -0,0 +1,29 @@
...
+        "midl.h",
+    ],
+    copts = [
+        "-Wbad-function-cast",
+        "-Wno-unused-but-set-variable",
+        "-Wno-unused-parameter",
+        "-Wuninitialized",
+    ],
+    linkopts = [
+        "-w",
```
jart: Isn't `-w` a copt that means ignore all warnings?

bowang: Good catch. Just fixed it.

tensorflow/core/BUILD
```diff
@@ -2879,6 +2879,19 @@ filegroup(
 )
 
 filegroup(
+    name = "lmdb_testdata",
```
jart: [Optional] Please add `#TODO(jart): Add testonly = 1 to this filegroup.`

tensorflow/core/kernels/BUILD
```diff
@@ -1922,6 +1923,14 @@ tf_kernel_library(
 )
 
 tf_kernel_library(
+    name = "lmdb_reader_op",
+    prefix = "lmdb_reader_op",
+    deps = IO_DEPS + [
+        "@lmdb//:lmdb",
```
jart: This can be shortened to `"@lmdb"`.

jart (approved):
Thank you for the contribution. Please consider making the few minor changes mentioned below.

rmlarsen (approved):
Thanks for the contribution.

rmlarsen:

bowang:
Thank you very much for the help and guiding me through this process! @rmlarsen @jart 
In cmake build, it still complains failing to find `lmdb.h`. Any idea how I can fix that?

maciekcc:
I think you might need to add external dependencies to third_party somewhere here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/CMakeLists.txt
and possibly a .cmake for lmdb in external/ to keep it clean.

rmlarsen:
@bowang FYI: We are waiting for confirmation that code covered by the OpenLDAP license is allowed in TensorFlow.

rmlarsen:
@bowang @maciekcc license is approved.

maciekcc:
Jenkins test this please.

bowang:

bowang:
@rmlarsen Thanks a lot for checking the license!
@maciekcc Thank you for the suggestion. It really helps! I added cmake files for lmdb and succeeded in a local cmake build.

maciekcc:
Jenkins test this please.

rmlarsen:
@bowang can you take a look at the test failures, please?

rmlarsen:

rmlarsen:
@bowang I also saw some bytearray vs. string literal errors  in the last set of tests (sorry, lost the link). Mind taking a look?

bowang:
@rmlarsen Definitely. Thanks for looking into the log. It's a pain that I can't reproduce the bug locally on Ubuntu while the Linux cpu test fails on CI.

bowang:

rmlarsen:

rmlarsen:

bowang:
Finally all checks have passed! Thanks for the help and guidance! @rmlarsen

bowang:
Can we merge this PR?

rmlarsen:
Thanks for the contribution! Merging now.

tensorflow/python/kernel_tests/reader_ops_test.py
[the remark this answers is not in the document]
Enigma-li: If we cannot enqueue the database file multiple times, how we can read data over one epoch using LMDB in tensorflow?

Issue #9827, opened 2017-05-11
tf.abs() isn't documented to handle complex, but it does appear to work as tf.complex_abs() used to.


ddurham2:
Issue #7405 was a bug filed that tf.complex_abs() was removed in 1.0.  At the bottom it says that tf.abs() now does that work, but the docs for tf.abs() only mention float.  I confirmed that tf.abs() does in fact do as the comment on the issue describes (https://github.com/tensorflow/tensorflow/blob/r1.1/tensorflow/python/ops/math_ops.py#L225)

Please incorporate the info from https://www.tensorflow.org/versions/r0.11/api_docs/python/math_ops/complex_number_functions#complex_abs 

into
https://www.tensorflow.org/api_docs/python/tf/abs

... specifically the parts about it computing sqrt(a^2 + b^2) for complex numbers.

girving:
@rmlarsen Do you mind fixing this?

ddurham2:
BTW- I would be happy to regularly make PRs for various documentation fixes.  However, I'm a little confused by the code.  I can grep for some of the descriptions to find current docs in the code, but I find that there is quite a bit of duplicate documentation.   Is this because there is separate documentation for the C++ and python code?  

Since I assume you want the same behavior in both, could this be changed to use \copydoc from doxygen to avoid having to put the docs in two places?

Otherwise, If I were to make a change, do I change all the places or is some of it generated and I only need to change one?

Generally, what is the rule for where to change things when fixing documentation?

girving:
@ddurham2 (cc @martinwicke): I don't think we have a good story for avoid code duplication between C++ and Python for ops with wrappers.  In this case (and many others) I don't think there's a good substitute for a human merging the docs, since they're different ops at the C++ level (though they probably shouldn't be).

@ddurham2 Did I interpret correctly that you're volunteering to send a PR?  If so, thank you!

ddurham2:
sure.. I'll give it a whirl. Could you give me a pointer on how to build the documentation (HTML or otherwise) to make sure my changes would look presentable?

martinwicke:
The documentation on documentation is here. If you have questions or if something doesn't work as expected, please ask. Note that the doc generator currently only works for Python 2.7.

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/community/documentation.md

PR #9954, opened 2017-05-17
Fixing documentation for tf.abs to include missing details about complex types.

ddurham2:
Adding lost documentation to tf.abs from the old tf.complex_abs when it learned how to work on complex data.


ddurham2:
I don't see a way to associate it, but this PR goes with https://github.com/tensorflow/tensorflow/issues/9827

rmlarsen (approved):

rmlarsen:
Thanks for fixing that!

nealwu:
We should merge this into master as well -- it was only merged into r1.1.

PR #9957, opened 2017-05-17
Branch 155393864

av8ramit:


yifeif (approved):

av8ramit:
Jenkins, test this please.

annarev (approved):

av8ramit:
Jenkins, test this please.

av8ramit:
Jenkins, test this please.

av8ramit:
Jenkins, test this please.

caisq:
cc @mrry. There seems to be a test failure in Windows CMAKE related to dataset_construct_op:
https://ci.tensorflow.org/job/tensorflow-pr-win-cmake-py/2372/console

mrry:
@caisq That's mysterious, because there's no platform-specific code in that test (as far as I can tell). To unblock the process while I investigate, can you add the following line:

```
      "${tensorflow_source_dir}/tensorflow/contrib/data/python/kernel_tests/dataset_constructor_op_test.py"
```

...to the list of tests to disable on Windows, here:

https://github.com/av8ramit/tensorflow/blob/5abe2a643c591f91d73abd53502ab2ac344298e6/tensorflow/contrib/cmake/tf_tests.cmake#L217

av8ramit:
@mrry @caisq done, rerunning tests now.

snnn:
Hi @mrry  I have a question: why these new ops (like RangeDatasetOp) are not inherited from ResourceOpKernel ?

mrry:
@snnn AFAIK `ResourceOpKernel<T>` doesn't give you enough control over how the resource is created. In particular, the Dataset constructor ops need to create an ephemeral resource in the step-local container, whereas—if I'm reading the code correctly—`ResourceOpKernel<T>` only creates session-scoped (or more widely shared) resources.

snnn:
@mrry  Thanks for the detailed explanation

snnn:
Hi @mrry , can you use MakePerStepResourceHandle<T> instead? 
How about add a template arg to ResourceOpKernel class, like

	template <typename T, bool per_step=false>
	class ResourceOpKernel : public OpKernel {
	 ...
	 
	 if (context->expected_output_dtype(0) == DT_RESOURCE) {    
		   if(per_step){
			   ...
			} else {
			....
		   }
		} 
	}

I want to create a new Dataset OP for a proprietary file format, now it is not easy to do so.

mrry:
It looks like `MakePerStepResourceHandle` would work, yes. As for adding a template parameter to `ResourceOpKernel<T>`, I don't have a strong feeling either way... but it wouldn't be useful for creating per-step `DatasetBase` resources, because the `CreateResource()` method does not have access to the `OpKernelContext*`, and this is usually needed for getting the arguments to the `Dataset` constructor.

PR #9965, opened 2017-05-17
fix a bug in using record_host_persistent_memory_allocation

snnn:


snnn:
Indeed, this bug only affects 32 bits builds. Is there any one using 32-bits tensorflow?

yaroslavvb:
@samjabrahams are Raspberry Pi TF builds 32 bit by any chance?

samjabrahams:
@yaroslavvb yes- though unfortunately I won't be able to play around with the effect of this code for a week or two (traveling).

Is this the gist of the problem: casting between `size_t` and `int64` on 32-bit systems causes memory to not be freed/tracked properly? Just trying to make sure I'm following.

snnn:
The following code:

	#include <iostream>
	#include <stdint.h>

	void record_host_persistent_memory_allocation(int64_t size) {
		std::cout << size << std::endl;
	}
	int main() {
		size_t a = 3;
		record_host_persistent_memory_allocation(-a);
		return 0;
	}

output is "4294967293" on 32 bits build and "-3" on 64 bits build.

samjabrahams:
I'll take a looksy when I am near some hardware. I also need to double check exactly which flags are set when TensorFlow compiles.

tensorflow/core/kernels/reduction_ops_common.h
```diff
@@ -238,9 +238,9 @@ class ReductionOp : public OpKernel {
     if (ctx->track_allocations()) {
       // The temporary memory becomes the output memory.
       if (ctx->allocate_on_host(alloc_attr)) {
-        ctx->record_host_temp_memory_size(-out.AllocatedBytes());
+        ctx->record_host_temp_memory_size(-(int64)out.AllocatedBytes());
```
rmlarsen: Please use C++ style casts, i.e. static_cast<int64>(foo).

rmlarsen (changes requested):

tensorflow/core/kernels/reduction_ops_common.h
snnn: @rmlarsen fixed.

rmlarsen:

rmlarsen (approved):

PR #9975, opened 2017-05-17
Fix a typo in export_output.py

raviqqe:


rmlarsen (approved):

nealwu:
We should merge this into master as well -- it was only merged into r1.1.

PR #9993, opened 2017-05-18
Add to tensor definition

redwrasse:
Add a small snippet to the tensor definition for the sake of correctness. People always ask about how these are related to tensors as physicists and mathematicians use.

### DIFF
--- a/tensorflow/docs_src/get_started/get_started.md
+++ b/tensorflow/docs_src/get_started/get_started.md
@@ -29,7 +29,9 @@ working internally when you use the more compact higher level API.
 
 # Tensors
 
-The central unit of data in TensorFlow is the **tensor**. A tensor consists of a
+The central unit of data in TensorFlow is the **tensor**. The formal physicist's and 
+mathematician's definition of a tensor is a multilinear map; the tensors here are in fact the same 
+thing and are simply values in a particular basis. A tensor in Tensorflow consists of a
 set of primitive values shaped into an array of any number of dimensions. A
 tensor's **rank** is its number of dimensions. Here are some examples of
 tensors:


4d55397500:
I signed it!

tensorflow/docs_src/get_started/get_started.md
```diff
@@ -29,7 +29,9 @@ working internally when you use the more compact higher level API.
 
 # Tensors
 
-The central unit of data in TensorFlow is the **tensor**. A tensor consists of a
+The central unit of data in TensorFlow is the **tensor**. The formal physicist's and 
+mathematician's definition of a tensor is a multilinear map; the tensors here are in fact the same
```
caisq: Do you mean "not the same thing"?

caisq: Please wrap line at 80 characters.

tensorflow/docs_src/get_started/get_started.md
```diff
@@ -29,7 +29,9 @@ working internally when you use the more compact higher level API.
 
 # Tensors
 
-The central unit of data in TensorFlow is the **tensor**. A tensor consists of a
+The central unit of data in TensorFlow is the **tensor**. The formal physicist's and 
+mathematician's definition of a tensor is a multilinear map; the tensors here are in fact the same 
+thing and are simply values in a particular basis. A tensor in Tensorflow consists of a
```
caisq: Maybe mention numpy.ndarray as very close analogy, as it is familiar to a lot of users:
https://docs.scipy.org/doc/numpy-1.12.0/reference/generated/numpy.ndarray.html

tensorflow/docs_src/get_started/get_started.md
caisq: You can add a link to the Wikipedia page for tensor or something like that for clarity.

caisq (changes requested):

tensorflow/docs_src/get_started/get_started.md
```diff
@@ -29,7 +29,9 @@ working internally when you use the more compact higher level API.
 
 # Tensors
 
-The central unit of data in TensorFlow is the **tensor**. A tensor consists of a
+The central unit of data in TensorFlow is the **tensor**. The formal physicist's and
```
rmlarsen: Some physicists can be very formal, but I don't think we need to go into this :-)

Jokes aside, I am not sure this clarification is particularly useful for most users of tensorflow, and I think that introducing more formal concepts like "map" and "basis" here will confuse more users than it will help. I think the users who know what those terms mean will also know that the concrete "tensors" in TensorFlow would be a representation of an abstract tnesor in a particular basis.

rmlarsen:
Closing.

PR #9994, opened 2017-05-18
Make ResourceHandle private to tensorflow library

snnn:
tensorflow::ResourceHandle is a protobuf object. The official protobuf document suggests:

"If your project is itself a DLL intended for use by third-party software, we recommend
that you do NOT expose protocol buffer objects in your library's public interface, and
that you statically link protocol buffers into your library"

Without this change, it is not possible to implement a ResourceOpKernel in DLL on Windows.

Error message(before this change):

```some.obj : error LNK2001: unresolved external symbol "class google::protobuf::internal::ExplicitlyConstructed<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > > google::protobuf::internal::fixed_address_empty_string" (?fixed_address_empty_string@internal@protobuf@google@@3V?$ExplicitlyConstructed@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@123@A)
some.obj : error LNK2019: unresolved external symbol "private: void __cdecl google::protobuf::Arena::AddListNode(void *,void (__cdecl*)(void *))" (?AddListNode@Arena@protobuf@google@@AEAAXPEAXP6AX0@Z@Z) referenced in function "public: void __cdecl google::protobuf::Arena::Own<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > *)" (??$Own@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@Arena@protobuf@google@@QEAAXPEAV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@Z)```

### DIFF
--- a/tensorflow/core/framework/resource_mgr.cc
+++ b/tensorflow/core/framework/resource_mgr.cc
@@ -24,6 +24,34 @@ limitations under the License.
 #include "tensorflow/core/platform/demangle.h"
 
 namespace tensorflow {
+ResourceHandle MakeResourceHandle(OpKernelContext* ctx, const string& container,
+                                  const string& name,
+                                  const TypeIndex& type_index) {
+  ResourceHandle result;
+  result.set_device(ctx->device()->attributes().name());
+  string actual_container;
+  if (!container.empty()) {
+    actual_container = container;
+  } else {
+    actual_container = ctx->resource_manager()->default_container();
+  }
+  result.set_container(actual_container);
+  result.set_name(name);
+  result.set_hash_code(type_index.hash_code());
+  result.set_maybe_type_name(type_index.name());
+  return result;
+}
+
+Status MakeResourceHandleToOutput(OpKernelContext* context, int output_index,
+                                  const string& container, const string& name,
+                                  const TypeIndex& type_index) {
+  Tensor* handle;
+  TF_RETURN_IF_ERROR(
+      context->allocate_output(output_index, TensorShape({}), &handle));
+  handle->scalar<ResourceHandle>()() =
+      MakeResourceHandle(context, container, name, type_index);
+  return Status::OK();
+}
 
 namespace internal {
 
--- a/tensorflow/core/framework/resource_mgr.h
+++ b/tensorflow/core/framework/resource_mgr.h
@@ -202,9 +202,20 @@ class ResourceMgr {
 
 // Makes a resource handle with the specified type for a given container /
 // name.
+ResourceHandle MakeResourceHandle(OpKernelContext* ctx, const string& container,
+                                  const string& name,
+                                  const TypeIndex& type_index);
+
 template <typename T>
 ResourceHandle MakeResourceHandle(OpKernelContext* ctx, const string& container,
-                                  const string& name);
+                                  const string& name) {
+  return MakeResourceHandle(ctx, container, name, MakeTypeIndex<T>());
+}
+
+Status MakeResourceHandleToOutput(OpKernelContext* context, int output_index,
+                                  const string& container, const string& name,
+                                  const TypeIndex& type_index);
+
 template <typename T>
 ResourceHandle MakePerStepResourceHandle(OpKernelContext* ctx,
                                          const string& name);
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by keveman** [comment]
Looks good to me. Adding @alextp just to have one more pair of eyes look at this.

File: tensorflow/core/framework/resource_mgr.cc [hunk 1]
```diff
@@ -25,6 +25,26 @@ limitations under the License.
 
 namespace tensorflow {
 
+Status setOutputResourceHandle(OpKernelContext* context, int output_index,
```
alextp: Can you name this SetOutputResourceHandle for more consistency with the surrounding code?

tensorflow/core/framework/resource_mgr.cc
```diff
@@ -25,6 +25,26 @@ limitations under the License.
...
+                               const string& container, const string& name,
+                               tensorflow::TypeIndex& type_index) {
+  Tensor* handle;
+  TF_RETURN_IF_ERROR(context->allocate_output(0, TensorShape({}), &handle));
```
alextp: I believe you meant to say output_index instead of 0 here.

tensorflow/core/framework/resource_mgr.cc
```diff
@@ -25,6 +25,26 @@ limitations under the License.
...
+  ResourceHandle& result = handle->scalar<ResourceHandle>()();
+  result.set_device(context->device()->attributes().name());
```
alextp: Use MakeResourceHandle here? Then we only have one place in the codebase which knows how to set the internal fields.

alextp (commented):
I do not understand how this change is supposed to make ResourceHandle private to the tensorflow dll. Can you explain?

tensorflow/core/framework/resource_mgr.cc
arena: Congrats on your pull request. I think you want to style your code snippets so that anybody that has the same username as one of your local variables doesn't get notifications.

rmlarsen: I added triple tickmarks around the code snipped. Sorry for the noise.

tensorflow/core/framework/resource_mgr.cc
snnn: @alextp any suggestion?

alextp: I meant capitalize the first letter, so it's Set and not set. And maybe make this a part of the public API too?

snnn:
Hi @alextp  

Assume we have a custom OP like this:

	template <typename T>
	class NumericResource : public tensorflow::ResourceBase
	{
	private:
		std::atomic<T> value;

	public:
		NumericResource() : value(0) {}
		T get() const { return value; }
		void inc() { ++value; }
		virtual std::string DebugString()
		{
			std::ostringstream oss;
			oss << value;
			return oss.str();
		}
	};
	typedef NumericResource<int> IntResource;

	class IntVarOp : public tensorflow::ResourceOpKernel<IntResource>
	{
	public:
		explicit IntVarOp(OpKernelConstruction* context) : tensorflow::ResourceOpKernel<IntResource>(context) {}
	private:
		virtual tensorflow::Status CreateResource(IntResource** resource)
		{
			*resource = new IntResource();
			return tensorflow::Status::OK();
		}
	};

	REGISTER_KERNEL_BUILDER(Name("IntVar").Device("CPU"), IntVarOp);

And we want to compile this OP into a dll or so.  And because ResourceOpKernel  is a template class,  the code in that class is generated in user ops' dll, not tensorflow's dll. So it shouldn't call any internal function.

Before this change, user's dll will contain a call to MakeResourceHandle<T> function, which will generate protobuf related code, and allocate a protobuf object in user's dll, pass it to tensorflow's dll. This is dangerous because each dll may have its own protobuf memory pool. By any chance, if you allocate an object in one pool but free it in another pool, it will crash. That's why protobuf's document prohibit you expose protocol buffer objects in your library's public interface.

After this change, the protobuf object's construction is done in tensorflow's dll (MakeResourceHandleToOutput function). 


 


ResourceOpKernel is a template class, so its code is generated in user ops' dll, not tensorflow's dll.

rmlarsen:
@alextp can you approve or provide further feedback?

rmlarsen:

snnn:
Sorry, I didn't test it under Linux. Will fix it.

tensorflow/core/framework/resource_mgr.cc
snnn: "public API" means what?

alextp:
Ah, I see your issue now. Can you respond to my review comments?

alextp (approved):

alextp:
Jenkins, test this please.

tensorflow/core/framework/resource_mgr.cc
snnn: fixed.

tensorflow/core/framework/resource_mgr.cc
snnn: fixed.

alextp:
Jenkins, test this please.

rmlarsen:

PR #9995, opened 2017-05-18
Update rnn.py

carefree0910:
Support returning all states generated in dynamic_rnn if the states are rather 'simple'
Unit test could be found [here](https://github.com/carefree0910/MachineLearning/blob/master/RNN/Test/UnitTest.py)
p.s. I finished this implementation under API r1.1 and I found the latest code of tensorflow.python.ops.rnn.py is quite different from that on my Windows machine with API r1.1. Since I cannot build the source code, though I've tested them under API r1.1, I haven't actually tested the committed codes in the latest environment

### DIFF
--- a/tensorflow/python/ops/rnn.py
+++ b/tensorflow/python/ops/rnn.py
@@ -404,7 +404,7 @@ def bidirectional_dynamic_rnn(cell_fw, cell_bw, inputs, sequence_length=None,
 
 def dynamic_rnn(cell, inputs, sequence_length=None, initial_state=None,
                 dtype=None, parallel_iterations=None, swap_memory=False,
-                time_major=False, scope=None):
+                time_major=False, scope=None, return_all_states=False):
   """Creates a recurrent neural network specified by RNNCell `cell`.
 
   Performs fully dynamic unrolling of `inputs`.
@@ -468,6 +468,7 @@ def dynamic_rnn(cell, inputs, sequence_length=None, initial_state=None,
       most TensorFlow data is batch-major, so by default this function
       accepts input and emits output in batch-major form.
     scope: VariableScope for the created subgraph; defaults to "rnn".
+    return_all_states: Whether return all states or not
 
   Returns:
     A pair (outputs, state) where:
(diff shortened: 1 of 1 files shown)


carefree0910:
In fact I was just wondering whether it is customary to return the final state only... Since the implementation is quite straightforward (Under 'simple' condition)

vrv:
@carefree0910 Can you provide context for this change?  Why does it need to be made?  This is adding to the API so will require API review before a lot of work should be done on it, and they will need more context as to why this is necessary.

ebrevdo:
The easiest way to pass states out to the output is to write a wrapper RNNCell that calls the underlying cell, propagates its state through, but emits a tuple output: the underlying cell's output, and the state as the second part of the tuple.

marcociccone:
Hi @ebrevdo , could you please elaborate more on the solution you provide here? I can't figure out how to implement such a thing. Thanks!

ebrevdo:
```python
class Wrapper(tf.nn.rnn_cell.RNNCell):
  def __init__(self, inner_cell):
     super(self, Wrapper).__init__()
     self._inner_cell = inner_cell
  @property
  def state_size(self):
     return self._inner_cell.state_size
  @property
  def output_size(self):
    return (self._inner_cell.state_size, self._inner_cell.output_size)
  def call(self, input, state)
    output, next_state = self._inner_cell(input, state)
    emit_output = (next_state, output)
    return emit_output, next_state
```

something like this

marcociccone:
Thanks! I didn't know you could pass a tuple as output too! In my case this wasn't necessary because I wrote my custom cell, but this is useful if you are using a cell already implemented in the api.

AleksTk:
@ebrevdo Is your wrapper solution applicable for lstml cell too? Please check my question on [stackoverflow](https://stackoverflow.com/questions/47745027/tensorflow-how-to-obtain-intermediate-cell-states-c-from-lstmcell-using-dynam). Thanks!

arvoelke:
@ebrevdo this approach doesn't work if using a [bidirectional wrapper](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Bidirectional). This seems to be because the bidirectional wrapper tries to stack the outputs from the cell, and doesn't like that the shape of `next_state` doesn't match `output`. Specifically,
```
...
   1653   try:
-> 1654     c_op = pywrap_tf_session.TF_FinishOperation(op_desc)
   1655   except errors.InvalidArgumentError as e:

InvalidArgumentError: Dimension 2 in both shapes must be equal, but are 128 and 640. Shapes are [?,1024,128] and [?,1024,640].
	From merging shape 2 with other shapes. for '{{node bidirectional/ReverseV2/tensor}} = Pack[N=4, T=DT_FLOAT, axis=0](bidirectional/backward_rnn/Identity, bidirectional/backward_rnn/Identity_1, bidirectional/backward_rnn/Identity_2, bidirectional/backward_rnn/Identity_3)' with input shapes: [?,1024,128], [?,1024,128], [?,1024,128], [?,1024,640].
```

I realize this thread is three years old, but is there any way to get all of the hidden states from a bidirectional RNN cell? Using `return_sequences=True` and `return_state=True` only returns the final state. I haven't been able to find another approach in my searches.

PR #10008, opened 2017-05-18
fix a bug in generate-pc.sh

arrufat:
I don't know how line 30 in the script managed to make its way into the last pull request.
I corrected the bug and forced the user to provide a version to generate the pkg-config file.

Sorry about that...


rmlarsen:
@arrufat Thanks for the quick fix.

rmlarsen (approved):

PR #10030, opened 2017-05-19
Fix comments for dataset ops

snnn:
@mrry


mrry (approved):
Thanks for fixing that! It was a late-breaking move during the release, and I didn't get round to fixing all of the comments :).

mrry:

PR #10032, opened 2017-05-19
add Cuda{2D,3D}LaunchConfig that maximizes occupancy

zasdfgbnm:
Add `Cuda2DLaunchConfig` and `Cuda3DLaunchConfig` that uses `cudaOccupancyMaxPotentialBlockSize` to calculate the best kernel launch parameters.


tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -68,15 +70,14 @@ inline CudaLaunchConfig GetCudaLaunchConfig(int work_element_count,
 template <typename DeviceFunc>
 inline CudaLaunchConfig GetCudaLaunchConfig(int work_element_count,
                                             const GPUDevice& d, DeviceFunc func,
-                                            size_t dynamic_shared_memory_size) {
+                                            size_t dynamic_shared_memory_size = 0) {
```
rmlarsen: Default values are discouraged. Please remove.

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -109,14 +110,52 @@ inline Cuda2DLaunchConfig GetCuda2DLaunchConfig(int xdim, int ydim,
...
 }
 
+// Calculate the Cuda 2D and 3D launch config we should use for a kernel launch.
+// This variant takes the resource limits of func into account to maximize occupancy.
+using Cuda3DLaunchConfig = Cuda2DLaunchConfig;
+
+template <typename DeviceFunc>
+inline Cuda3DLaunchConfig GetCuda3DLaunchConfig(int xdim, int ydim, int zdim,
+                                                const GPUDevice& d, DeviceFunc func,
+                                                size_t dynamic_shared_memory_size = 0) {
```
rmlarsen: Don't use default value.

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -109,14 +110,52 @@ inline Cuda2DLaunchConfig GetCuda2DLaunchConfig(int xdim, int ydim,
...
+  config.thread_per_block = dim3(threadsx, threadsy, threadsz);
+  config.block_count = dim3(blocksx, blocksy, blocksz);
+
+  return config;
+}
+
+template <typename DeviceFunc>
+inline Cuda2DLaunchConfig GetCuda2DLaunchConfig(int xdim, int ydim,
+                                                const GPUDevice& d, DeviceFunc func,
+                                                size_t dynamic_shared_memory_size = 0) {
```
rmlarsen: No default value.

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -109,14 +110,52 @@ inline Cuda2DLaunchConfig GetCuda2DLaunchConfig(int xdim, int ydim,
...
+  int block_count = 0;
+  int thread_per_block = 0;
+  cudaOccupancyMaxPotentialBlockSize(&block_count, &thread_per_block, func,
+                                     dynamic_shared_memory_size,
+                                     xdim * ydim * zdim);
+
+  int threadsx = std::min(xdim, thread_per_block);
+  int threadsy = std::min(ydim, std::max(thread_per_block / threadsx, 1));
+  int threadsz = std::min(zdim, std::max(thread_per_block / (threadsx * threadsy), 1));
```
rmlarsen: This can cause division by zero, e.g., if xdim is zero.  Please add handling for one or mode dims being <= 0.

rmlarsen (changes requested):

tensorflow/core/util/cuda_kernel_helper.h
zasdfgbnm: `cudaOccupancyMaxPotentialBlockSize` has a default value for this parameter, so I add the same value here. I will remove it if doing so will make it fits better with the code style of tensorflow.

tensorflow/core/util/cuda_kernel_helper.h
zasdfgbnm: fixed

rmlarsen:

rmlarsen (approved):

zasdfgbnm:
I make some changes because I think it would also be necessary to check the physical limitation of each dimension:
```c++
int dev;
cudaGetDevice(&dev);
cudaDeviceProp deviceProp;
cudaGetDeviceProperties(&deviceProp, dev);
int xthreadlimit = deviceProp.maxThreadsDim[0];
int ythreadlimit = deviceProp.maxThreadsDim[1];
int zthreadlimit = deviceProp.maxThreadsDim[2];
int xgridlimit = deviceProp.maxGridSize[0];
int ygridlimit = deviceProp.maxGridSize[1];
int zgridlimit = deviceProp.maxGridSize[2];
```

rmlarsen:
@zasdfgbnm Could you add a test for this?

zasdfgbnm:
@rmlarsen I add and pass these tests. I also add some explanation of these `Cuda{2D,3D,}LaunchConfig` into `cuda_kernel_helper.h`. 

The test is a `.cu.cc` file. To build it, I have to add something to `tensorflow.bzl`. I'm not sure if I'm doing the correct thing for this change.

rmlarsen:
@zasdfgbnm thanks for the thorough test!

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -23,10 +23,82 @@ limitations under the License.
 #include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
 #include "tensorflow/core/platform/types.h"
 
+// Usage of GetCudaLaunchConfig, GetCuda2DLaunchConfig, and GetCuda3DLaunchConfig:
+//
+// There are two versions of GetCudaLaunchConfig and GetCuda2DLaunchConfig, one
```
rmlarsen: It would be less confusing if the two versions had different names, instead of overloading the same symbol.

rmlarsen (changes requested):

rmlarsen:
@zasdfgbnm can you fix the build file formatting errors reported here:


https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/4433/consoleFull

tensorflow/core/util/cuda_kernel_helper.h
zasdfgbnm: Yes, I agree, but there is another thing to consider: The 1D maximum occupancy version is introduced in https://github.com/tensorflow/tensorflow/commit/3c02d1100788789b04e04feb93761f0ad898ea77 and given the same symbol `GetCudaLaunchConfig` half a month ago. I'm not sure how to deal with this. Three options are:

- Keep `GetCudaLaunchConfig` unchanged, and rename the maximize occupancy version of 2D and 3D to `Get2DCudaMaxOccupancyLaunchConfig` and `...3D...`. This might make things even more confusing.
- Rename all these functions, and change the kernel launches affected. This makes things clearer, but might break somebody's code. Are functions in this file considered API that should avoid change?
- Leave these functions as it is (i.e. overload the same symbol)

Which one do you prefer?

rmlarsen:

zasdfgbnm:
Check return value of `cudaOccupancyMaxPotentialBlockSize`
Plus some style fixes.

Although github mark it as outdated, the issue https://github.com/tensorflow/tensorflow/pull/10032#discussion_r118041775 and my question https://github.com/tensorflow/tensorflow/pull/10032#discussion_r118129926  is still there.

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -47,16 +129,22 @@ struct CudaLaunchConfig {
 // memory-limited.
 inline CudaLaunchConfig GetCudaLaunchConfig(int work_element_count,
                                             const GPUDevice& d) {
+  CudaLaunchConfig config;
+
+  // in case of invalid input, return the default value config, which has all -1
+  if (work_element_count <= 0) {
+    return config;
+  }
```
zasdfgbnm: Which is better here? Log and exit using `CHECK_GE`, or explicitly return an invalid output as in above?

rmlarsen: The original code ends up creating a config with virtual_thread_count=0, I think. I'm guessing that the original author thought that returning such a config is fine in this case, since presumably the caller will subsequently call CUDA_1D_KERNEL_LOOP(i, n)   with n==0 which becomes a noop. @zheng-xq can you advice?

rmlarsen:

rmlarsen:

rmlarsen:

rmlarsen (approved):

rmlarsen:
@zasdfgbnm can you please rebase to resolve conflicts?

zasdfgbnm:
@rmlarsen done

rmlarsen:

rmlarsen:
@zasdfgbnm The test failure looks like it's unrelated to your change, so I'll merge it as is. Thanks for the contribution.

PR #10034, opened 2017-05-19
Replace use of tensorflow::ops::ReadFile in label_image

freedomtan:
Use tensorflow::Env instead of tensorflow::ops::ReadFile so that
we can avoid including whole_file_read_ops for Android.


rmlarsen (approved):
Cool, thanks for the quick fix!

rmlarsen:

PR #10037, opened 2017-05-19
Fix capitalization typo in release notes

daj:
`TensorFLowInferenceInterface` had an incorrectly capitalized "L" in it.

While doing it, I formatted a couple of package names as code (since others seem to be).


daj:
D'oh, the commit was made while logged into the wrong account, retrying. :-(

PR #10079, opened 2017-05-21
Fix classification dataset example in readme

see--:
Using python lists results in an IndexError for me.

PR #10082, opened 2017-05-21
Automatically convert inputs to tensors in Dataset.from_tensor_slices

omoindrot:
Allows usage like:
```python
tf.contrib.data.Dataset.from_tensor_slices([['img1.jpg', 'img2.jpg'], [1, 2]])
```
Otherwise, the two input lists (2 of size 2 here) were flattened into 1 list (of size 4), which resulted in a "list index out of range" error.

My correction was inspired by how the `tf.train.slice_input_producer` is defined: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/input.py#L302

Also see pull request #10079.

### DIFF
--- a/tensorflow/contrib/data/python/ops/dataset_ops.py
+++ b/tensorflow/contrib/data/python/ops/dataset_ops.py
@@ -887,6 +887,10 @@ class TensorSliceDataset(Dataset):
     """See `Dataset.from_tensor_slices()` for details."""
     super(TensorSliceDataset, self).__init__()
     with ops.name_scope("tensors"):
+      tensors = ops.convert_n_to_tensor_or_indexed_slices(tensors)
+      if not tensors:
+        raise ValueError(
+            "Expected at least one tensor in Dataset.from_tensor_slices().")
       flat_tensors = [
           ops.convert_to_tensor(t, name="component_%d" % i)
           for i, t in enumerate(nest.flatten(tensors))


maciekcc:
Jenkins test this please.

maciekcc:
Jenkins test this please.

tensorflow/contrib/data/python/ops/dataset_ops.py
```diff
@@ -887,6 +887,10 @@ def __init__(self, tensors):
     """See `Dataset.from_tensor_slices()` for details."""
     super(TensorSliceDataset, self).__init__()
     with ops.name_scope("tensors"):
+      tensors = ops.convert_n_to_tensor_or_indexed_slices(tensors)
```
mrry: I think this raises a `ValueError` if the `tensors` argument is a single `tf.Tensor`, which is a valid "nested structure" to pass here. (`tf.train.slice_input_producer()` is more restrictive on the types that it accepts.) It would also fail on more complex nested structures, which are handled by the `nest.flatten()` on the lines below.

Perhaps the right thing to do here is to raise a better error than the `IndexError` when the constructor flattens these arguments into four scalars? (The out-of-range is because `Dataset.from_tensor_slices()` requires each of its tensor arguments to have rank > 1, but there's currently no error checking code for that case.) The message could guide users to specify an explicit `tf.constant()` or `tf.convert_to_tensor()`.

mrry (changes requested):

tensorflow/contrib/data/python/ops/dataset_ops.py
omoindrot: Hi @mrry,
I really like the idea of not having to specify `tf.constant(filenames)` and `tf.constant(labels)` when feeding them into the function to create a dataset.
Even in the `tf.contrib.data` README, you use as example:
```python
filenames = ["/var/data/file1.txt", "/var/data/file2.txt"]
dataset = tf.contrib.data.Dataset.from_tensor_slices(filenames)
```
And
```python
filenames = tf.constant(["/var/data/image1.jpg", "/var/data/image2.jpg", ...])
labels = tf.constant([0, 37, 29, 1, ...])

dataset = tf.contrib.data.Dataset.from_tensor_slices((filenames, labels))
```

Both examples don't work with the current version.

I think the easiest fix would be to force users to pass in a list, as in `tf.train.slice_input_producer`? We would have to rewrite some examples and kernel tests, but I feel it's more natural.

drpngx:
@omoindrot ping?

omoindrot:
@drpngx I answered @mrry in the comments of the change, but I am waiting for his answer.

I can make the simple change he requested (just modifying the error raised), but I'd like to find a better solution.

frankchn:
ping @mrry

tensorflow/contrib/data/python/ops/dataset_ops.py
mrry: Sorry for dropping this PR on the floor... https://github.com/tensorflow/tensorflow/commit/f3f53e8b394bdcaddc707f7bde8dcc98a73531e7 modifies the way nested structures are interpreted to add support for dicts and remove support for lists, so as a result a list will be interpreted as Tensor-like argument and converted to a `tf.Tensor` by the existing code. This fixes the broken doc examples, and I think it satisfies your use cases as well. 

Correct me if I'm wrong, but I think that change supersedes the need for this PR. Is that right?

omoindrot: You're right, this does seem like a good compromise to force users to use tuples (which actually make more sense in this case).
I'll close the PR.

PR #10084, opened 2017-05-21
Adding NewScopeWithGraph func to Scope struct. This enables the construction of an in-memory graph from a previously serialized form + new additions.

ctava:
Found a use case that wasn't covered while developing with tensorflow in go.

Since working with multiple-graphs is an anti-pattern:
http://stackoverflow.com/questions/35955144/working-with-multiple-graphs-in-tensorflow

needed to reuse a previously trained GraphDef file (binary) and then make additions. 

What do you think?

### DIFF
--- a/tensorflow/go/op/scope_test.go
+++ b/tensorflow/go/op/scope_test.go
@@ -95,6 +95,40 @@ func TestMultipleGeneratedOps(t *testing.T) {
 	}
 }
 
+func TestScopeWithGraph(t *testing.T) {
+	s1 := NewScope()
+
+	var hello interface{} = "hello"
+	tensor1, _ := tf.NewTensor(hello)
+	s1.AddOperation(tf.OpSpec{
+		Type: "Const",
+		Name: hello.(string),
+		Attrs: map[string]interface{}{
+			"dtype": tensor1.DataType(),
+			"value": tensor1,
+		},
+	})
+	graph, err := s1.Finalize()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	s2 := NewScopeWithGraph(graph)
+	var world interface{} = "world"
+	tensor2, _ := tf.NewTensor(world)
+	s2.AddOperation(tf.OpSpec{
+		Type: "Const",
+		Name: world.(string),
+		Attrs: map[string]interface{}{
+			"dtype": tensor2.DataType(),
+			"value": tensor2,
+		},
+	})
+	if _, err := s2.Finalize(); err != nil {
+		t.Fatal(err)
+	}
+}
+
 func Example() {
 	// This example creates a Graph that multiplies a constant matrix with
 	// a matrix to be provided during graph execution (via
--- a/tensorflow/go/op/scope_test.go
+++ b/tensorflow/go/op/scope_test.go
@@ -97,34 +97,15 @@ func TestMultipleGeneratedOps(t *testing.T) {
 
 func TestScopeWithGraph(t *testing.T) {
 	s1 := NewScope()
-
-	var hello interface{} = "hello"
-	tensor1, _ := tf.NewTensor(hello)
-	s1.AddOperation(tf.OpSpec{
-		Type: "Const",
-		Name: hello.(string),
-		Attrs: map[string]interface{}{
-			"dtype": tensor1.DataType(),
-			"value": tensor1,
-		},
-	})
+	Const(s1, "hello")
 	graph, err := s1.Finalize()
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	s2 := NewScopeWithGraph(graph)
-	var world interface{} = "world"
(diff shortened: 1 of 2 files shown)


asimshankar:
Thanks for the PR @ctava 

Working with multiple-graphs is not an anti-pattern in the Go API (or C++, or Java, or C for that matter) as these APIs do not have an implicit global graph like the Python API does. You should not hesitate from using multiple Graph objects in the same Go program.

If that is the case, do you still want to pursue this PR?

If you really need to extend an existing graph, then something like this might be necessary. I'm a bit torn between adding a `Scope.ImportGraphDef` vs. this approach. Currently, the `Scope` API makes for a clear separation between construction and use of the graph which will break if you can create a `Scope` with an arbitrary graph. It may be okay to forgo that property, but still, it is something that we're giving up :). (Adding a `Scope.ImportGraphDef` won't do that).

Do let us know if you need this functionality to extend a graph (as opposed to creating a different one) and we can take the PR further.

Thanks!

FYI @jhseu

ctava:
@asimshankar it says, we make assumption that one session has one graph:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api.h#L59

https://github.com/tensorflow/tensorflow/blob/08cb0ba5431ef2508c6ca63c498023a84f7b4a18/tensorflow/go/op/scope.go#L55
when calling finalized on a scope - it outputs 1 graph. 

NewSession: accepts a graph:
https://github.com/tensorflow/tensorflow/blob/08cb0ba5431ef2508c6ca63c498023a84f7b4a18/tensorflow/go/session.go#L51

yea. i'd like to build a base Tensorflow graph, persist it and have the ability to hydrate it and extend it on a case-by-case basis

asimshankar:
@ctava : Ah, sorry, from the discussion in the stackoverflow thread and the use of the phrase "anti-pattern" you alluded to, I thought you were referring to multiple graphs in a single process. Multiple graphs in a session isn't an "anti-pattern" (in that it is possible but discouraged), it simply isn't possible :)

Thanks for the follow up, and the use case of wanting to extend an existing graph is quite reasonable. Just wanted to clarify your original comments. Will take a look at the PR now.

tensorflow/go/op/scope_test.go
```diff
@@ -95,6 +95,40 @@ func TestMultipleGeneratedOps(t *testing.T) {
 	}
 }
 
+func TestScopeWithGraph(t *testing.T) {
+	s1 := NewScope()
+
+	var hello interface{} = "hello"
```
asimshankar: Why define `hello` as an `interface{}` instead of just `string`?

This would work just as well, no?
```go
hello := "hello"
```

tensorflow/go/op/scope_test.go
```diff
@@ -95,6 +95,40 @@ func TestMultipleGeneratedOps(t *testing.T) {
...
+	tensor1, _ := tf.NewTensor(hello)
+	s1.AddOperation(tf.OpSpec{
```
asimshankar: Any particular reason to avoid the the `Const` method used in other tests in this file?
So this could be as simple as:

```go
s1 := NewScope()
Const(s1, "hello")
graph, err := s1.Finalize()
if err != nil {
  t.Fatal(err)
}

s2 := NewScopeWithGraph(graph)
Const(s2, "world")
if err := s2.Err(); err != nil {
  t.Fatal(err)
}
```

asimshankar (changes requested):

tensorflow/go/op/scope_test.go
ctava: using the `Const` method leads to:
 scope_test.go:109: failed to add operation "Const": Duplicate node name in graph: 'Const'

tensorflow/go/op/scope_test.go
ctava:
```hello := hello
	tensor1, _ := tf.NewTensor(hello) ```

leads to `op/scope_test.go:101: undefined: hello`

Comment (asimshankar) [inline tensorflow/go/op/scope_test.go]: Sorry, that should be:

```go
hello := "hello"
tensor1, _ := tf.NewTensor(hello)
```

(The quotes were missing)

tensorflow/go/op/scope_test.go
asimshankar: That's because of the name collission. Perhaps you can replace the second `Const` with:

```go
s2 := NewScopeWithGraph(graph)
Const(s2.SubScope("addition"), "world")
...
```

ctava:
@asimshankar changes have been made to the TestScopeWithGraph unit test. its much more concise now.

asimshankar (approved):
Thanks for the contribution!

maciekcc:
Jenkins test this please.

maciekcc:
Jenkins test this please.

PR #10108, opened 2017-05-22
Reverting the manual tags added to problem tests.

av8ramit:
Hopefully we can fix them by rc1.


gunan:
Are these fixed?
If they are not fixed yet, let's wait on this until we have a fix.

av8ramit:
Is it okay to merge 1.2 back into master then? This would disable these tests from nightly as well.

PR #10115, opened 2017-05-22
Fix TF windows bazel tests.

gunan:
- Ignore errors in shutil.rmtree until we figure out why rmtree fails.
 - Do not run neon_depthwise_conv_op, as we already omit linking in
 the op.


martinwicke:
The rmtree fails because the SummaryWriterCache keeps open file handles around which causes Windows to disallow removing the directory.

Look at estimator_test: test_train_save_copy_reload for an example of how to fix.

@ispirmustafa should we call SummaryWriterCache.clear() after each train()/eval() call? There's little we gain from this cache between invocations.

ispirmustafa:
we need to keep cache for some edge cases. For example let's say there are two threads one for training and one for evaluation.

martinwicke:
But that doesn't prevent us from clearing the cache, right? All that would
do is force us to reopen the writer, which is a performance hit but should
otherwise be harmless.

meteorcloudy:
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/10/consoleFull
Looks like adding `ignore_errors=True` didn't fix the shutil.rmtree error.

gunan:
I can try, but for the said edge cases, multiple threads can also be causing the failures to remove files.

gunan:
Running windows presubmits here:
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/11/

martinwicke:
Well that didn't work -- There shouldn't be multiple thread unless several methods inside DNNRegressorEvaluateTest run in multiple threads, and even then, the directory is unique to that test case.

gunan:
It did not work because there were >1 `shutil.rmtree` calls, I now updated them all.
Trying again.
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/12/

gunan:
Looks like some machines had some missing libraries?
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/13/

Retrying tests.

gunan:
There were some new tests added to this file.
Also modified them, and rerunning tests:
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/14/

gunan:
OK, now I got all the tests. Looks like while I was fixing testcases, more cases were added to dnn_test.

http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/16/console

martinwicke (approved):

PR #10126, opened 2017-05-23
cudnn: Fix symlink includes handling

jthestness:
Following a change to symlink creation for cuda_configure.bzl, the cudnn
include path symlink would try to link to a path ending with
'includeinclude/cudnn.h' rather than the correct path, 'include/cudnn.h'.
It appears this is caused by the way _symlink_genrule_for_dir handles
particular source header files. Since the cudnn library only exports a single
include file, cuda_configure.bzl can just use _symlink_genrule_for_dir
without explicitly specifying the header files to symlink.

### DIFF
--- a/third_party/gpus/cuda_configure.bzl
+++ b/third_party/gpus/cuda_configure.bzl
@@ -875,8 +875,8 @@ def _create_cuda_repository(repository_ctx):
   included_files = _read_dir(repository_ctx, cuda_include_path).replace(
       cuda_include_path, '').splitlines()
   if '/cudnn.h' not in included_files:
-    genrules.append(_symlink_genrule_for_dir(repository_ctx, None, "",
-        "cudnn-include", [cudnn_header_dir + "/cudnn.h"], ["include/cudnn.h"]))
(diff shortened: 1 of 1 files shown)


maciekcc:
Jenkins test this please.

martinwicke:
@jart @caisq can you comment on this?

@jthestness There's a conflict, likely because someone else solved the same problem, can you rebase and see whether this is still needed?

jart:
@jthestness Bazel recommends against a genrule specifying a directory as an input or output. It's best to have individual files be listed. I read your commit message, but I don't 100% understand yet what's going on here.

One thing that might help is if you run the following command:

```sh
less $(bazel info output_base)/external/local_config_cuda/cuda/BUILD
```

That will show you the generated BUILD file so we can get a better idea of what this genrule looks like, and why it's behaving poorly.

jthestness:
I was able to rebuild current mainline without this PR, so it does seem that someone else fixed this issue. This PR can be closed.

caisq:
Thanks for the info, @jthestness

PR #10143, opened 2017-05-23
RecordInput mini batches for dividing processing among multiple devices.

ahundt:
Adds one component of the [high performance models tensorflow benchmark](https://www.tensorflow.org/performance/performance_models) upstream so it is more readily available. This is based on the minibatch function in [preprocessing.py](https://github.com/tensorflow/benchmarks/blob/master/scripts/tf_cnn_benchmarks/preprocessing.py) in the example benchmark.

I need to write a test, but wasn't sure if this is the right place to add this functionality to the API.

### DIFF
--- a/tensorflow/python/ops/data_flow_ops.py
+++ b/tensorflow/python/ops/data_flow_ops.py
@@ -2155,7 +2155,8 @@ class RecordInput(object):
                parallelism=1,
                shift_ratio=0,
                seed=0,
-               name=None):
+               name=None,
+               batches=None):
     """Constructs a RecordInput Op.
 
     Args:
@@ -2169,12 +2170,18 @@ class RecordInput(object):
       seed: Specify the random number seed used by generator that randomizes
         records.
       name: Optional name for the operation.
+      batches: None by default, creating a single batch op. Otherwise specifies
+        how many batches to create, which are returned as a list when
+        `get_yield_op()` is called. An example use case is to split processing
+        between devices on one computer.
 
     Raises:
       ValueError: If one of the arguments is invalid.
     """
-
     self._batch_size = batch_size
+    if batches is not None:
+      self._batch_size *= batches
+    self._batches = batches
     self._file_pattern = file_pattern
     self._buffer_size = buffer_size
     self._parallelism = parallelism
@@ -2183,8 +2190,11 @@ class RecordInput(object):
     self._name = name
 
   def get_yield_op(self):
-    """Add a node that yields a minibatch every time it is executed."""
-    return gen_data_flow_ops.record_input(
+    """Adds a node that yields a group of records every time it is executed.
+    If RecordInput `batches` parameter is not None, it yields a list of
+    record batches with the specified `batch_size`.
+    """
+    records = gen_data_flow_ops.record_input(
         file_pattern=self._file_pattern,
         file_buffer_size=self._buffer_size,
         file_parallelism=self._parallelism,
@@ -2192,3 +2202,14 @@ class RecordInput(object):
         batch_size=self._batch_size,
         file_random_seed=self._seed,
         name=self._name)
+    if self._batches is None:
+      return records
+    else:
+      with ops.name_scope(self._name):
+        batch_list = [[] for i in six.moves.range(self._batches)]
+        records = array_ops.split(records, self._batch_size, 0)
+        records = [array_ops.reshape(record, []) for record in records]
+        for index, protobuf in zip(six.moves.range(len(records)), records):
+          batch_index = index % self._batches
+          batch_list[batch_index].append(protobuf)
+        return batch_list
--- a/tensorflow/python/kernel_tests/record_input_test.py
+++ b/tensorflow/python/kernel_tests/record_input_test.py
@@ -53,6 +53,7 @@ class RecordInputOpTest(test.TestCase):
   def testRecordInputEpochs(self):
     files = 100
     records_per_file = 100
+    batches = 2
     with self.test_session() as sess:
       self.generateTestData("basic", files, records_per_file)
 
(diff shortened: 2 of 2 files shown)


mrry:
Reassigning to @ekelsen, who added this code and is best placed to rule on API changes!

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -2088,6 +2088,7 @@ class RecordInput(object):
   def __init__(self,
                file_pattern,
                batch_size=1,
+               minibatches=None,
```
ekelsen: I think it's best to add new optional parameters to the end of the list, otherwise you break callers not using named parameters.

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -2099,6 +2100,10 @@ def __init__(self,
       file_pattern: File path to the dataset, possibly containing wildcards.
         All matching files will be iterated over each epoch.
       batch_size: How many records to return at a time.
+      minibatches: None by default, otherwise specifies
```
ekelsen: I personally find it more natural to think of minibatches as a multiplier of batch_size, rather than a divisor.  batch_size specifies the size of each mini_batch and minibatches tell you how many you get. 

So if you want 4 batches of 32, (batch_size = 32 and minibatches = 4) rather than (batch_size = 128 and minibatches=4).

Thoughts?

ahundt: I was debating between the two and just picked one, I was thinking you choose a batch size because that is what will fit on your machine(s), then subdivide it.

tensorflow/python/ops/data_flow_ops.py
ahundt: sure, can do

tensorflow/python/ops/data_flow_ops.py
ahundt: shouldn't it be called a macrobatch if it multiplies? :-)

ekelsen: There could still be a larger concept of batch size assuming there are multiple machines involved, which is why I think the least ambiguous option is to use batch size to describe either the largest or smallest grouping.

Perhaps we can slip the issue by avoiding ML terminology all together, since really we're just trying to ask for N groups of size K.

records_per_group
num_groups

ahundt:
@ekelsen tried incorporating some of those changes you suggested, guess there are also some conflicts to resolve

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -2108,12 +2109,17 @@ def __init__(self,
       seed: Specify the random number seed used by generator that randomizes
         records.
       name: Optional name for the operation.
+      num_groups: None by default, otherwise specifies
+        how many groups of records of size records_per_group to create.
+        An example use case is to split processing between devices on one computer.
+        _Must_ be a multiple of batch_size.
```
ekelsen: last line is out of date

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -2108,12 +2109,17 @@ def __init__(self,
       seed: Specify the random number seed used by generator that randomizes
         records.
       name: Optional name for the operation.
+      num_groups: None by default, otherwise specifies
+        how many groups of records of size records_per_group to create.
```
ekelsen: "how many groups of size records_per_group to create." seems more clear to me.

ekelsen (approved):
LGTM, thanks for bearing with me caring too much about the names.

ekelsen (changes requested):
Approved too soon, tests still need to be added.

ahundt:
@ekelsen where should a test go? There doesn't seem to be a `data_flow_ops_test.py`.

I also realized that changing `batch_size` would break backwards compatibility, so I decided to just name the second parameter `batches`.

ahundt:
found it https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/record_input_test.py

ekelsen:
sgtm

ahundt:
@ekelsen ready for another review

ekelsen (approved):

ekelsen:
LGTM, thanks!

ahundt:
Great! Do we need to @ mention a tensorflower for the final steps?

ekelsen:
Jenkins, test this please

gunan:
Jenkins, test this please.

ahundt:
tests pass locally for me now

ahundt:
@ekelsen could you re-run the test?

ekelsen:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

martinwicke:
Broken Mac build rebuilt here: http://ci.tensorflow.org/job/tensorflow-pull-requests-mac/5319/console

ahundt:
So you're saying the test passed? My local test was actually on OS X

martinwicke:
Yes, the test did pass.

ahundt:
Awesome my first real code merged in TF :-) thanks!

ahundt:
Oh wait I guess some code took the back door earlier via keras, but still good haha.

martinwicke:
Thanks!​

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -2183,12 +2190,26 @@ def __init__(self,
...
+    if self._batches is None:
+      return records
+    else:
+      with ops.name_scope(self._name):
+        batch_list = [[] for i in six.moves.range(self._batches)]
+        records = array_ops.split(records, self._batch_size, 0)
+        records = [array_ops.reshape(record, []) for record in records]
+        for index, protobuf in zip(six.moves.range(len(records)), records):
+          batch_index = index % self._batches
+          batch_list[batch_index].append(protobuf)
```
ahundt: @ekelsen I just realized I forgot to include the equivalent of the lines that [parallel_stack batches](https://github.com/tensorflow/benchmarks/blob/b9221114d1261ec28f103455178abcfbeeb9b498/scripts/tf_cnn_benchmarks/preprocessing.py#L378) together, do you think that belongs here as well or should it stay in user code? 

Leaving things as they are may make sense in this case because this code contains full records they'll need to reorganize according to their features.

```
      for device_index in xrange(self.device_count):
        images[device_index] = tf.parallel_stack(images[device_index])
        label_index_batch[device_index] = tf.concat(labels[device_index], 0)
```

ekelsen:
I agree that stacking the full records probably doesn't make sense.

Could maybe consider passing an optional function that turns records into something that can be stacked, and then stacking them?

ahundt:
Seems like a technically sound strategy, but maybe there is an alternative that is more straightforward for users?

I don't know the internals of `StagingArea` but perhaps feeding full records directly to that, pulling out the result, then returning might be enough so loading will be fast and users can do as they please with the returned results?

Perhaps the easiest thing to do is I could add a comment recommending users just read the [high performance models guide](https://www.tensorflow.org/performance/performance_models) before using. :-)

PR #10158, opened 2017-05-24
Fix Cuda configuration on Windows

meteorcloudy:
When creating a genrule for a set of files, we should use copy instead of
symlink on Windows.

This addresses an error reported at stackoverflow:
https://stackoverflow.com/questions/44038367/genrules-without-outputs-dont-make-sense-bazel-windows-10-build

Applied the same approach to python_config.bzl to simplify the code.


tensorflow/tools/ci_build/windows/gpu/pip/build_tf_windows.sh
```diff
@@ -67,4 +67,4 @@ passing_tests=$(bazel query "kind(py_test,  //${PY_TEST_DIR}/tensorflow/python/.
 # Define no_tensorflow_py_deps=true so that every py_test has no deps anymore,
 # which will result testing system installed tensorflow
 # GPU tests are very flaky when running concurently, so set local_test_jobs=5
```
maciekcc: nit: please update comment to reflect the change in script.

meteorcloudy: Thanks! Done.

maciekcc:
Jenkins test this please.

third_party/gpus/cuda_configure.bzl
```diff
@@ -769,14 +773,44 @@ def _create_dummy_repository(repository_ctx):
...
     src_files = [], dest_files = []):
-  """Returns a genrule to symlink a set of files.
+  """Returns a genrule to symlink(or copy if on Windows) a set of files.
 
   If src_dir is passed, files will be read from the given directory; otherwise
   we assume files are in src_files and dest_files
   """
   if src_dir != None:
+    src_dir = src_dir.replace("\\", "/")
+    dest_dir = dest_dir.replace("\\", "/")
```
maciekcc: One more nit: can you please extract a method to perform this path slash replacement and trimming, and then use it in both the files.

meteorcloudy: No problem. Done.

maciekcc (approved):

PR #10161, opened 2017-05-24
DOC modify install_mac and install_linux

shitian-ni:
Originally in [install_mac](https://www.tensorflow.org/install/install_mac) it says 

> installing from sources, which is for experts and is documented in a separate guide.

But it didn't say what that separate guide is. I think that is wield, so I added a link to that.
Also, [install_linux](https://www.tensorflow.org/install/install_linux) didn't mention install from source option, so I added one to that too

### DIFF
--- a/tensorflow/docs_src/install/install_linux.md
+++ b/tensorflow/docs_src/install/install_linux.md
@@ -69,6 +69,8 @@ supported choices are as follows:
   * ["native" pip](#InstallingNativePip)
   * [Docker](#InstallingDocker)
   * [Anaconda](#InstallingAnaconda)
+  * installing from sources, which is for experts and is documented in
+    [a separate guide](https://www.tensorflow.org/install/install_sources).
 
 **We recommend the virtualenv installation.**
 [Virtualenv](https://virtualenv.pypa.io/en/stable/)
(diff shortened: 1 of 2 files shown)


tensorflow/docs_src/install/install_linux.md
```diff
@@ -69,6 +69,8 @@ supported choices are as follows:
   * ["native" pip](#InstallingNativePip)
   * [Docker](#InstallingDocker)
   * [Anaconda](#InstallingAnaconda)
+  * installing from sources, which is for experts and is documented in
```
maciekcc: I would remove the 'is for experts' part... it is not that complicated :)

tensorflow/docs_src/install/install_mac.md
```diff
@@ -12,7 +12,7 @@ You must pick the mechanism by which you install TensorFlow. The supported choic
   * "native" pip
   * Docker
   * installing from sources, which is for experts and is documented in
```
maciekcc: same here

tensorflow/docs_src/install/install_linux.md
shitian-ni: I see, sounds reasonable :)
I will remove them

maciekcc (approved):

maciekcc:
Jenkins test this please.

PR #10164, opened 2017-05-24
[XLA] Move some useful Literal conversion code into LiteralUtil

DavidNorman:
This code looks like it should be in LiteralUtil.  It is useful to have it there for our backend, and probably also for allowing literals to be created with explicit types to the constant data type.


tensorflow/compiler/xla/literal_util_test.cc
```diff
@@ -857,5 +857,88 @@ TEST_F(LiteralUtilTest, ConvertR4) {
...
+  conv = LiteralUtil::ConvertIfSrcTypeMatches(*f32, S32).ConsumeValueOrDie();
+  EXPECT_TRUE(LiteralUtil::Equal(*conv, *s32));
+
+  conv = LiteralUtil::ConvertIfSrcTypeMatches(*f64, S32).ConsumeValueOrDie();
+  EXPECT_TRUE(LiteralUtil::Equal(*conv, *s32));
+
+  conv = LiteralUtil::ConvertIfSrcTypeMatches(*s32, F32).ConsumeValueOrDie();
+  EXPECT_TRUE(LiteralUtil::Equal(*conv, *f32));
+
+  EXPECT_FALSE(LiteralUtil::ConvertIfSrcTypeMatches(*s32, TUPLE).ok());
```
dlibenzi: I see an EXPECT_NOT_OK() macro as being useful to test negative cases.

dlibenzi: Did these test pass (the negative ones)?

DavidNorman: they did pass.  i'll check again and change the test macro.

DavidNorman: Oh.  I can only find this in the source base:
```
// There are no EXPECT_NOT_OK/ASSERT_NOT_OK macros since they would not
// provide much value (when they fail, they would just print the OK status
// which conveys no more information than EXPECT_FALSE(status.ok());
```

dlibenzi: Doing EXPECT_FALSE() on ok() is not going to generate a better message, but the comment you found has a point.
Maybe EXPECT_EQ(func(), util::OkStatus())?

DavidNorman: Ok.  Couldn't find util::OkStatus() in the code, but I changed it to check for the type of error explicitly.

dlibenzi: Sorry, my bad, Status::OK(), but checking the error code is good as well.

dlibenzi (approved):

kayzhu:

kayzhu (approved):

kayzhu:
Passing this back to @hawkinsp to merge.

hawkinsp (approved):

hawkinsp:

DavidNorman:
I notice dnovillo has made a significant change to the LiteralUtil class recently.  he/she should probably be aware of this change.

DavidNorman:
ok - this one seems good to merge

DavidNorman:
it would be really useful to merge this one today.  this evening i will try to fix the F16 support in LiteralUtil, and we might end up with two different merges which both modify the same files.

kayzhu:
@hawkinsp this looks good to merge. I've pinged dnovillo to take a look as well.

dnovillo:
So, this change is actually going in the opposite direction it should've gone.  LiteralUtil is on the way out.  I'm working on a patch that is getting rid of it.

It's not a big deal that this change is in, but note that I will be moving the new code inside xla::Literal.

DavidNorman:
awesome, thanks guys.  will submit the f16 fix as soon as I can (later tonight as i have to be a taxi driver now)

martinwicke:
I am sorry -- I did merge this. I made a revert but it sounds like that's probably not worth it, so I'll get rid of that.

dnovillo:
Yup, no worry about this one.  I'll roll it up with my patch to get rid of
LiteralUtil.

PR #10174, opened 2017-05-24
Register GPU RefExit kernel

drasmuss:
It looks like the code was all set up to support a GPU kernel for `RefExit` (as was done for similar kernels, e.g. `RefEnter`), but `RefExit` was never actually registered for the GPU.  I'm not sure if it was intentionally left out for some reason, or if this was just an oversight.  All I did was add the registration calls and it appears to work.


drasmuss:
The more general motivation for this change is that currently all loop Variables are pinned to the CPU, because they are colocated with the CPU-only `RefExit`.  This means that the Variables need to be transferred on and off the GPU every loop iteration.  Adding a GPU kernel for `RefExit` allows the whole loop to run on the GPU, which significantly improves performance.

drasmuss:
Any updates on the review?  Normally I wouldn't push, but it'd be really nice to get this into the 1.2 release if possible, since it's a pretty significant performance bottleneck.  @ebrevdo might be able to quickly comment, since he wrote the GPU RefExit code.

ebrevdo (dismissed):
this *seems* fine; though i think we'll have to run some internal tests to confirm no breakages.

you'll probably want to also register a GPU_HOST_REF_KERNEL for the RefExit.

drasmuss:
I believe the GPU_HOST_REF_KERNEL is already being registered [here](https://github.com/drasmuss/tensorflow/blob/51a7b7be792139f0f35877f464b88c50f11190c8/tensorflow/core/kernels/control_flow_ops.cc#L445)

drasmuss:
Are there any other changes you'd suggest @ebrevdo?

ebrevdo:
No, I'm away this week so will ask someone else to review.

On Jun 13, 2017 7:29 AM, "Daniel Rasmussen" <notifications@github.com>
wrote:

Are there any other changes you'd suggest @ebrevdo
<https://github.com/ebrevdo>?

—

<https://github.com/tensorflow/tensorflow/pull/10174#issuecomment-308134751>,
or mute the thread

martinwicke:
@keveman can you take a look?

keveman (approved):

keveman:

gunan:
I have just ran all internal tests, and looks like this change does not cause any breakages internally.
It is safe to merge this PR.

PR #10176, opened 2017-05-24
Make PYTHON_BIN_PATH error go away

### LINKED ISSUE #9436: Python Configuration Error: 'PYTHON_BIN_PATH' environment variable is not set
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: N/A
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OS X Sierra
- **TensorFlow installed from (source or binary)**: N/A (compiling from HEAD)
- **TensorFlow version (use command below)**: N/A (see above)
- **Bazel version (if compiling from source)**: 0.4.5-homebrew
- **CUDA/cuDNN version**: none (AMD GPU)
- **GPU model and memory**: Radeon Pro 460

`sudo bazel build --config opt --copt=-msse4.1 --copt=-msse4.1 --copt=-mavx --copt=-mavx2 --copt=-mfma //tensorflow/tools/pip_package:build_pip_package`

### Describe the problem

Trying to build tensorflow from source (version installed via pip does not use some optimised CPU instructions), get the following error:

```bash
 ~/workspace/tensorflow   master  sudo bazel build --config opt --copt=-msse4.1 --copt=-msse4.1 --copt=-mavx --copt=-mavx2 --copt=-mfma //tensorflow/tools/pip_package:build_pip_package
WARNING: Config values are not defined in any .rc file: opt
ERROR: /Users/kachkach/workspace/tensorflow/third_party/py/numpy/BUILD:11:1: no such package '@local_config_python//': Traceback (most recent call last):
	File "/Users/kachkach/workspace/tensorflow/third_party/py/python_configure.bzl", line 180
		_create_python_repository(repository_ctx)
	File "/Users/kachkach/workspace/tensorflow/third_party/py/python_configure.bzl", line 157, in _create_python_repository
		_get_env_var(repository_ctx, _PYTHON_BIN_PATH)
	File "/Users/kachkach/workspace/tensorflow/third_party/py/python_configure.bzl", line 48, in _get_env_var
		_python_configure_fail("'%s' environment variable is no...)
	File "/Users/kachkach/workspace/tensorflow/third_party/py/python_configure.bzl", line 36, in _python_configure_fail
		fail("
%sPython Configuration Error:%...))

Python Configuration Error: 'PYTHON_BIN_PATH' environment variable is not set
 and referenced by '//third_party/py/numpy:headers'.
ERROR: Analysis of target '//tensorflow/tools/pip_package:build_pip_package' failed; build aborted.
```
Note that I used `which python3` to find where my Python binary is located, and exported that as PYTHON_BIN_PATH, but it does not seem to help.

jart:
This also makes the logging a little less aggressive.

Fixes #9436

CC: @nlopezgi

### DIFF
--- a/third_party/py/python_configure.bzl
+++ b/third_party/py/python_configure.bzl
@@ -28,14 +28,14 @@ def _python_configure_warning(msg):
   """Output warning message during auto configuration."""
   yellow = "\033[1;33m"
   no_color = "\033[0m"
-  print("\n%sPython Configuration Warning:%s %s\n" % (yellow, no_color, msg))
+  print("%sPython Configuration Warning:%s %s" % (yellow, no_color, msg))
 
 
 def _python_configure_fail(msg):
   """Output failure message when auto configuration fails."""
   red = "\033[0;31m"
   no_color = "\033[0m"
-  fail("\n%sPython Configuration Error:%s %s\n" % (red, no_color, msg))
+  fail("%sPython Configuration Error:%s %s\n" % (red, no_color, msg))
 
 
 def _get_env_var(repository_ctx, name, default = None, enable_warning = True):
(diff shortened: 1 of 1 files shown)


dandelionmane:
So if I understand correctly, this defaults to /usr/bin/python when we can't find it from explicit configuration. But we don't check whether this actually points to a good python. So this seems more like a bandaid fix rather than getting at the root issue. Am I correct in that analysis? If so, should we at least put a comment in the code saying that this is a hack? I notice in the linked issue you said the root cause may be an upstream issue in bazel.

jart:
Yes this is a bandaid fix, because Bazel is currently broken.

@nlopezgi wrote this Python configuration tool, so I hope he will take ownership of creating a more optimal solution.

dandelionmane:
Can we open an issue against bazel and have a link in our code saying essentially "this is a hack, remove when bazelbuild/issues/4242 is resolved"?

jart:
Done.

maciekcc (approved):

nlopezgi:
Looking into a permanent fix for the issue. Apologies for the late reply.

nlopezgi:
Tried to repo this error by using standard Dockerfile.devel in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docker/Dockerfile.devel (comment on https://github.com/tensorflow/tensorflow/issues/9436 indicates they could repro the issue with this setup, iiuc?). Was able to run the build successfully even after removing the default value set in https://github.com/tensorflow/tensorflow/blob/master/third_party/py/python_configure.bzl#L246. Ran build successfuly with both bazel 0.4.5 and 0.5.1. Do you have any pointers as to how to reliably repro this error so I can diagnose better. Thanks.

PR #10191, opened 2017-05-25
[XLA] Ensure constants conform to the specified index type

DavidNorman:
The strided slice update introduces constants into the HLO graph which are not of the type specified by the core op.

This adds methods to the computation builder to allow the creation of literals with types determined at runtime, and makes the strided slice update op use them.

it depends on https://github.com/tensorflow/tensorflow/pull/10164


DavidNorman:
I left this line in the elemental_ir_emitter:

```
llvm_ir::IrArray::Index dim_index(1, ir_builder_->getInt64(i));
```
I believe it is an index into the vector of indices, and that the CreateZExtOrBitCast function will successfully extend the int32 to whatever the native size of the target IR is.

DavidNorman:
Actually - despite the potential for more type-specific constants in the future, I'm abandoning this one.

PR #10201, opened 2017-05-25
Enable more robust Java lint and errorprone checks

kbsriram:
1. Turn on stricter Xlint and errorprone checks in javac to converge
   on a more consistent coding style.

2. Minor changes to existing code to conform to the new checks.

3. google-java-format corrections on any java file modified as part of
   the fix.


maciekcc:
Jenkins test this please.

asimshankar:
Thanks @kbsriram . This is pretty neat.

Though, I'm a bit confused about why we need to explicitly list out all the `-Xep` flags. Is there some documentation you can point to? http://errorprone.info/docs/installation suggests that "Error Prone works out of the box with bazel".

Ideally we wouldn't have to list out 125 lines of flags :), so I'd like to understand the configuration here a bit better.

asimshankar:
(FYI @jhseu )

kbsriram:
Ah - sorry! Out of an abundance of caution, my goal was to reduce divergence between new java code coming in via github vs critique before we queue up our java-related changes.

Many of these flags are "hand-curated" within g3 linting tools, and not enabled as errors in the out-of-box errorprone configuration (I'll send you links separately.) Happy to defer to your judgement of course; I've also grouped them into sections in case a smaller set seems to be preferable.

asimshankar (approved):
Thanks a bunch. Looking forward to seeing it in action.

maciekcc:
Jenkins test this please.

PR #10209, opened 2017-05-26
Suppress windows build warnings to reduce win build log size.

gunan:
Can't add as reviewers, but also happy to hear comments from @guschmue @vit-stepanovs 
The suppressed warnings are:
 - 4003 - not enough actual parameters for macro 'identifier'
 - 4244 - 'conversion_type': conversion from 'type1' to 'type2', possible loss of data
 - 4267 - 'variable': conversion from 'size_t' to 'type', possible loss of data
 - 4503 - 'identifier': decorated name length exceeded, name was truncated
 - 4506 - no definition for inline function 'function'
 - 4800 - 'type': forcing value to bool 'true' or 'false' (performance warning)
 - 4996 - 'function' or 'variable' marked as deprecated

The numbers of times we encounter these:
```
   1374 C4003:
   1106 C4244:
    850 C4267:
   1104 C4503:
    156 C4506:
    102 C4800:
    116 C4996:
```


gunan:
I also have the following linker warnings:
```
    444 LNK4049:
      2 LNK4070:
    246 LNK4197:
   1332 LNK4217:
    126 LNK4221:
```
do I need to change somewhere else for linker suppressions, or should I just add `/ignore:<id>` same way I add compiler warning suppressions.

mrry:
Change looks good to me; I don't know the answer about linker warnings.

mrry (approved):

guschmue:
The linker errors: should do no harm to /ignore: them.
Most of them seem to be coming from the external png lib. I can take a look why we see them.
If you are looking at disk usage for the builds in general: I noticed that some unit tests are collecting a good amount of disk space in \windows\temp\

PR #10031, opened 2017-05-19
[XLA] HLO Executor based backend example

DavidNorman:
Adding an example XLA device that uses the hlo evaluator to perform the execution.

PR #10217, opened 2017-05-26
[XLA] Add a strides parameter to the XLA slice operation

DavidNorman:
This looks quite large but really it is just threading a strides parameter for the XLA Slice op through the stack.

The elemental emitter has actually changed, as it now performs the striding
The slice shape inference has changed to take into account striding
the strided slice tf2xla no longer uses concat to construct a strided slice


tensorflow/compiler/xla/service/shape_inference.cc
```diff
@@ -1028,32 +1029,41 @@ ShapeInference::InferDegenerateDimensionBroadcastShape(
...
                                            start_index);
     VLOG(2) << tensorflow::strings::Printf("limits[%lld] = %lld", dimension,
                                            limit_index);
-
-    sizes.push_back(limits[dimension] - starts[dimension]);
+    if (stride > 0) {
+      if (start_index > limit_index) {
+        return InvalidArgument(
+                "limit index (%lld) must be greater or equal to "
+                        "start index (%lld) in slice with positive stride",
```
hawkinsp: Strange indent.

tensorflow/compiler/xla/service/shape_inference.cc
```diff
@@ -1028,32 +1029,41 @@ ShapeInference::InferDegenerateDimensionBroadcastShape(
...
+                limit_index, start_index);
+      }
+      sizes.push_back((limit_index - start_index + stride - 1) / stride);
+    } else {
+      if (start_index < limit_index) {
+        return InvalidArgument(
+                "limit index (%lld) must be less than or equal to "
+                        "start index (%lld) in slice with negative stride",
```
hawkinsp: Strange indent.

hawkinsp (commented):
A couple of small comments.

There's also an API design question I sent to the xla-dev list.

tensorflow/compiler/xla/client/computation_builder.h
```diff
@@ -211,9 +211,11 @@ class ComputationBuilder {
   //
   // Note that "limit" means up-to-but-not-including; i.e. [start, limit) in 1D
   // range notation.
+  // The stride parameter determines the stride over the slice
```
DavidNorman: Personally I would rather stick to positive only strides (on the grounds of similarity with our backend API).   I suppose that if the 'rev plus slice from end+1 -> begin+1'  code goes into tf2xla then that makes for simpler code in all backends.

I put in negative strides because that's what TF core does.  It does it (probably) because of numpy.

tensorflow/compiler/tf2xla/kernels/split_op.cc
```diff
@@ -79,12 +79,14 @@ class SplitOp : public XlaOpKernel {
     // split dimensions varies for each output.
     std::vector<int64> begin;
     std::vector<int64> limits;
+    std::vector<int64> strides;
```
DavidNorman: indeed. i was maintaining coding style, but agreed it would be better to be more efficient.

tensorflow/compiler/xla/service/shape_inference.cc
DavidNorman: yeah.  my editor is weird sometimes.

DavidNorman:
i suspect no-one cares enough to discuss it on the mailing list.  If you think it should be positive strides only then I'm happy to change it.  Otherwise I think it should be merged sooner rather than later because it is quite a large change in terms of the number of files touched.

tensorflow/compiler/xla/client/computation_builder.h
hawkinsp: I talked about it with some folks and the consensus seemed to be to prefer positive strides only for now. The reasoning is:
a) we want to keep XLA minimal where possible, since it makes life easier for backend authors --- fewer cases to implement,
b) it doesn't add any expressive power, since you can use Rev() to reverse dimensions, and it doesn't add any efficiency because we can fuse Rev() and Slice().
c) it's easier to add features later rather than to remove them, so we can always add support for negative strides if we think of a good reason why.

DavidNorman: ok.  i'm on holiday next week, but I will try to get it sorted out this weekend.  shouldn't be hard to add the rev+offset to the tf2xla and remove from the elemental emitter.

DavidNorman:
@hawkinsp This is now positive strides only.

hawkinsp (approved):

hawkinsp:

DavidNorman:
Hi.  See the comments on https://github.com/tensorflow/tensorflow/pull/10031 regarding the test failures.   I can't see a fix in the changes that I pulled a minute ago, but I filed an Issue with the main TF repo and hopefully someone is on it.  I imagine it is holding all public contributors up in the same way.

DavidNorman:
oops - i jumped the gun by looking at only that there were 3 failures.  will look closely at the XLA one

DavidNorman:
maybe the Python3 one is something to do with not having the latest tensorboard included (?) will assume that for the moment.

DavidNorman:
and the Linux XLA one is down to bad initializers.  will fix ASAP

DavidNorman:
@hawkinsp hi - i see that the XLA test failure is because there are tests in tensorflow/compiler/xla/service that i am not currently running (only tensorflow/compiler/xla and tensorflow/compiler/tests at the moment).

is there a test suite runner for the ones in .../xla/service ?  I can't find one

hawkinsp:
You should just be able to run something like this:

```
$ bazel test //tensorflow/compiler/xla/...
```

I don't think there's an explicit `test_suite` for those tests but you can just use the `...` pattern to test everything in a subdirectory.

martinwicke:
Do you need help running the tests, or are you working on these failures?

DavidNorman:
working on failures.    have to pull recent changes - and also fix up one of our own internal problems.... getting there :)

DavidNorman:
ps - @hawkinsp thanks for the tip - didn't know about '...'

DavidNorman:
Ok - i think that the XLA tests are probably ok now.   ran all the tests as suggested.

DavidNorman:

DavidNorman:
maybe it doesn't work if i do it

hawkinsp:

DavidNorman:
phew

DavidNorman:
the irony is that I only did that to see how easy it would be to add another Op called SliceUpdate (the non-dynamic version of DynamicUpdateSlice).  Since then I've decided to implement a cheat version of int64 in my device, because I suspect it is a losing battle to keep XLA using only types that the core Ops indicate.

DavidNorman:
the nice thing is that this merge will mean that the difference between my repo and the public one is back to a few lines in 3 files. happy days.

hawkinsp:
Thanks for all the contributions!

PR #10219, opened 2017-05-26
Fixed crf.py to force log norm of zero length sequences to zero

chqiwang:
The original log_norm function in crf.py will not output zero when sequence length is zero. This new version force it to be zero.


chqiwang:
I signed it!

PR #10231, opened 2017-05-26
Disable flaky tests that have been breaking nightly builds.

gunan:
FYI @caisq @mrry @ekelsen @alisidd

### DIFF
--- a/tensorflow/python/debug/BUILD
+++ b/tensorflow/python/debug/BUILD
@@ -727,7 +727,6 @@ sh_test(
     name = "examples_test",
     size = "medium",
     srcs = ["examples/examples_test.sh"],
-    tags = ["manual"],
     data = [
         ":debug_errors",
         ":debug_fibonacci",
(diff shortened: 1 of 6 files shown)


ekelsen:
Did my CL that went in yesterday not fix the stage_op_test ?

tensorflow/python/debug/BUILD
```diff
@@ -727,6 +727,7 @@ sh_test(
     name = "examples_test",
     size = "medium",
     srcs = ["examples/examples_test.sh"],
+    tags = ["manual"],
```
caisq: I think the flaky timeout should have been fixed by CL/156945796, which is already pushed out to GitHub and increases the test size to medium to prevent flaky timeouts. So this change is not necessary.

caisq (changes requested):

gunan:
@ekelsen @caisq I probably looked into test results before your changes were merged. reverting changes to stabe_op_test and example_test

caisq (approved):

PR #10235, opened 2017-05-27
fix the return value of Tensor::flat_inner_outer_dims

zasdfgbnm:
The const version of `flat_inner_outer_dims` should return `ConstTensor` instead of `Tensor`


maciekcc:
Jenkins test this please.

maciekcc (approved):

PR #10242, opened 2017-05-27
Fixed tf.contrib.crf.crf_log_norm to handle zero sequence length.

chqiwang:
The original version of tf.contrib.crf.crf_log_norm can not produce zero output when the sequence length is zero or a negtive number, which will leads tf.contrib.crf_log_likelihood to produce non-zero output, so that the training procedure can be disturbed.
This patch force the output to be zero when the sequence length is equal or less than zero.


alextp:
Can you add a test with zero-sized sequences?

alextp:
Just checking in; did you have a chance to add the test I asked?

drpngx:
@chqiwang feel free to re-open once you have the test in.

PR #10246, opened 2017-05-27
Undo modification to `x` dict by deleting `_TARGET_KEY` after features dequeue

md-mq:


tensorflow/python/estimator/inputs/numpy_io.py
```diff
@@ -132,6 +132,8 @@ def input_fn():
     features = dict(zip(ordered_dict_x.keys(), features))
     if y is not None:
       target = features.pop(unique_target_key)
+      # undo modification to dictionary
+      del x[unique_target_key]
```
xiejw: Do you really need this line? The modified dict is the ordered_dict_x (a shadow copy of x). Input x is not modified.

tensorflow/python/estimator/inputs/pandas_io.py
```diff
@@ -116,6 +116,8 @@ def input_fn():
     features = dict(zip(list(x.columns), features))
     if y is not None:
       target = features.pop(target_column)
+      # undo modification to dictionary
+      del x[target_column]
```
xiejw: Line 79 of this file has x=x.copy(). So, same question: Do you really need this line?

xiejw (commented):
As commented in the changes, I cannot approve this change. Do you have an example that the input x has been polluted?

tensorflow/python/estimator/inputs/numpy_io.py
mouradmourafiq: Ok, this PR is then obsolete, there was no dictionary copy at the time, and the modification was basically done directly on the `x` instead of the `ordered_dict_x`. I failed to check the code after rebasing, my modification was based on this [version](https://github.com/tensorflow/tensorflow/blob/r1.1/tensorflow/python/estimator/inputs/numpy_io.py#L102)

tensorflow/python/estimator/inputs/pandas_io.py
mouradmourafiq: Same as my other comment. I will close the PR, as it is not necessary anymore.

Issue #7149, opened 2017-01-30
[Java][Feature] Generating operation methods used to build a graph


karllessard:
As suggested in https://www.tensorflow.org/versions/r0.11/how_tos/language_bindings, the list of operation methods for building a graph should be generated dynamically from the list exposed by the core, using preferably protobuf. 

Is anyone already working on that feature? If not, I'm tempted to try it. I was thinking of generating a builder hierarchy that allows to optionally set an argument after adding an operation. For example
```java
GraphBuider
    .matMul(a, b)
        .withTransposeB(true)
    .softmax(logits)
    ...
```
and the GraphBuilder classes would be generated at build time by Bazel and would make use of the already existing OperationBuilder. Since I'm new to Tensorflow, please tell me if that doesn't make any sense to you

Thanks

(p.s. as suggested by @drpngx, I've started this discussion as a seperate issue to avoid continuing the #5 saga)

drpngx:
Thanks @karllessard! @asimshankar should know if that's in the plans or if it would be a welcome contribution.

asimshankar:
Having generated code for ops in Java is indeed a next step for the Java API, but we haven't gotten around to it, or figuring out what an appropriate API would be.

We'd be happy to hear any suggestions/proposals you may have. As for the suggestion above, I don't think we want to put all operations as methods in a single `GraphBuilder` class for a few reasons such as (a) Per-op attribute handling seems a bit tricky, (b) What about custom operations defined by other external libraries (e.g. as per the [Adding a new op HOWTO](https://www.tensorflow.org/how_tos/adding_an_op/)).

We can take inspiration from the generated C++ API (see the [label_image example](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/label_image/main.cc#L102) and the Go API, but it does require more thought.

karllessard:
I agree that exposing operations in their own particular object would create a richer API than declaring everything into a single builder class, I'll also get my mind into it.

Now, without insisting, I'll clarify how my suggestion would have handled point a), maybe some of those ideas can morph into something else:

1. The generated builder class implements each operations by returning a specialized builder that inherits this same class (e.g. GraphBuilder.matMul() returns a MatMulGraphBuilder that inherits GraphBuilder).

2. The specialized builder exposes all other operations directly from its superclass (for going on with graph building) + methods for setting optional attributes of the current operation, prefixed by "with" (e.g. withTransposeA() and withTransposeB()).

[...]

PR #10255, opened 2017-05-27
Initial scope implementation.

kbsriram:
This is an initial implementation to support scopes for the
Java language bindings. It currently provides only hierarchical
name scopes.

Please see https://github.com/tensorflow/tensorflow/issues/7149
to track the master issue.


kbsriram:
@Moriadry thanks! I've started [an informal group](https://groups.google.com/forum/#!forum/tensorflow-java-dev-unofficial) to find out who's working on what in this particular area, and consolidate efforts from interested people - we can continue the conversation there.

Moriadry:
@kbsriram Got it, I already submit a request to join, many thanks!

tensorflow/java/src/main/java/org/tensorflow/Scope.java
```diff
@@ -0,0 +1,198 @@
...
+      graph = g;
+    }
+
+    private Builder nameScope(NameScope ns) {
+      nameScope = ns;
+      return this;
+    }
+
+    private Scope build() {
+      return new Scope(this);
```
asimshankar: I'm a little confused by this Scope, Scope.Builder, NameScope, NameScope.Builder relationships. It seems like a lot of abstraction that can be simplified, but perhaps I'm mistaken.

I was imagining a single class, `Scope`, with something like this:

```
public final class Scope {
   ...
   // All the state needed for a Scope
   private final Graph graph;
   private final String name;  // "fully qualified name", i.e., path from root to here with slashes
   private final Map<String, int> opNames;  // Map from op-name (ignoring scope name prefix) to count
   private final Map<String, int> subNames;  // Map from sub-scope name to count
  
   // Unfortunate additional state needed for withOpName.
   // null to use the operation name as the node name
   private String opName;
}
```

And then:

- `new Scope` would set all fields (`name` to `""`, `opName` to `null`, the maps to newly allocated maps)
- `withSubScope` would do the name as `new Scope` except set `name` to `this.name + "/" + name`
- `withOpName` would copy all the fields of the "parent" `Scope` (so that the count maps are shared), but set `opName` differently

I _think_ doing so will make the scope code easier to follow as there will be fewer layers to navigate through.
Thoughts?

tensorflow/java/src/test/java/org/tensorflow/ScopeTest.java
```diff
@@ -0,0 +1,184 @@
...
+  @Test
+  public void basic() {
+    try (Graph g = new Graph()) {
+      Scope s = Scope.create(g);
+      Const c1 = Const.create(s, 42);
+      assertEquals("Const", c1.input().op().name());
+      Const c2 = Const.create(s, 7);
+      assertEquals("Const_1", c2.input().op().name());
+      Const c3 = Const.create(s.withOpName("four"), 4);
+      assertEquals("four", c3.input().op().name());
```
asimshankar: Perhaps also add, for documentation purposes, test for the following scenario (a second constant with the same name):

```java
Const c4 = Const.create(s.withOpName("four"), 4);
```

tensorflow/java/src/test/java/org/tensorflow/ScopeTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+  @Test
+  public void basic() {
+    try (Graph g = new Graph()) {
+      Scope s = Scope.create(g);
+      Const c1 = Const.create(s, 42);
+      assertEquals("Const", c1.input().op().name());
+      Const c2 = Const.create(s, 7);
+      assertEquals("Const_1", c2.input().op().name());
+      Const c3 = Const.create(s.withOpName("four"), 4);
```
asimshankar: What do you think about `withName` instead of `withOpName`? So, a client will deal with `s.withSubScope("foo")` and `s.withName("bar")`?

tensorflow/java/src/test/java/org/tensorflow/ScopeTest.java
```diff
@@ -0,0 +1,184 @@
...
+      Tensor result = sess.runner().fetch(var1.input()).run().get(0);
+      assertEquals(21704, result.intValue());
+      result = sess.runner().fetch(var2.input()).run().get(0);
+      assertEquals(21704, result.intValue());
+    }
+  }
+
+  // Convenience interface
+  // TODO: replace when standardized
+  interface Input {
```
asimshankar: For now, I know we might undo this, but might I recommend getting rid of this interface and the `input()` method in the "handwritten" classes? Without the complete context, things like `public Output input() { return output; }` and the use in `Const(...).input().op()...` seem confusing. Better to keep this patch focused on the classes it adds?

asimshankar (commented):
Thanks for this PR and sorry for the late reply! Somehow I missed that I hadn't sent out these comments and just noticed. 

My main concern is about the implementation complexity of `Scope` with many helper classes. I was wondering if we can simplify the implementation down to a single class with a small number of "state" member variables.

tensorflow/java/src/main/java/org/tensorflow/Scope.java
kbsriram: Thank you for the thoughtful and detailed comments @asimshankar - much appreciated! I'll update this PR to reflect your suggestions.

May we sound you out a bit further on `Scope` and `NameScope`? No problem collapsing them in one class, though this what we were thinking.

There are two classes now, with `Scope` delegating to `NameScope`. We organized it as a main class delegating to a specialized class so we can add other self-contained groups of properties like "control dependencies" without putting everything in one implementation.

In this approach, the public `Scope` would be augmented to hold package private `NameScope` and `ControlDependency` objects, delegating to each as appropriate.

A `Scope` carrying a certain name hierarchy would then implement `withControlDependencies()` by reusing the existing `NameScope` and just creating a new `ControlDependency` object. `withName()` and `withOpName()` would do the opposite, etc. Having specialized classes to manage each group of properties would also make it a bit easier to test each in isolation.

What do you think?

[Re: Builders to construct classes rather than raw constructors; we were following [https://engdoc.corp.google.com/eng/doc/devguide/java/practices/builders.md?cl=head](https://engdoc.corp.google.com/eng/doc/devguide/java/practices/builders.md?cl=head) style that  prefers Builders or a static factory method to raw constructors. We chose Builders here as `NameScope` manages 3/4 constructor parameters of interest, and `Scope` currently has just one but we anticipate at least two after adding `ControlDependency`.  That said, we're happy to switch to raw constructors if you prefer this style for Tensorflow, we can switch 
 in this and subsequent PRs, please let us know.]

asimshankar: Appreciate the thoughtful reply. A few replies:

- RE: Builders: I took those guidelines to mean that that builders are great as a replacement for factory methods when there are many parameters, or optional parameters or when calling code is less self-documenting that it should be. However, in this particular case, the proposal is to have only a single constructor (`new Scope(graph)`) and then named methods for child scopes (`withFoo()`), hence I don't think the preconditions for those guidelines are met. The builder in this PR is private and used for the implementation and IMHO in this case they are making the implementation harder to follow. So, I'd like to lean against the use of the private builders here. Sound fair?

- RE: Helper scopes: I see. IIUC, you're trying to encapsulate all the state needed for "names" in one class, all the state needed for "control dependencies" in another class etc. So, instead of having each `withFoo()` method worry about how it handles every private state member of the `Scope` class, it can deal with groups? I can buy that - let's see how things look once the builders are gone (I suspect the implementation of both classes will become much smaller) and see how it goes then.

kbsriram: That sounds great :-) Your summary about helper scopes is just how we're thinking about it as well. I'll update the PR with all the changes you suggested (but keep the helper scopes as-is) and go from there. Thanks!

tensorflow/java/BUILD
```diff
@@ -15,6 +15,16 @@ java_library(
     visibility = ["//visibility:public"],
 )
 
+# Operator convenience wrappers for the core tensorflow library.
+java_library(
+    name = "tensorflow_op",
```
asimshankar: Instead of a separate `java_library`, how about including `:java_op_sources` in the `tensorflow` target? That way we end up with one `.jar`. Otherwise we have to worry about distribution of this jar and will have to deal with more complicated instructions on www.tensorflow.org/install/install_java

(The Android builds use the `java_sources` filegroup so they won't include these files).

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+ * separated by a forward slash {@code '/'}. For instance, {@code nn/Const_72} or {@code
+ * nn/gradient/assign/init}. Each scope is a subtree in this hierarchy.
+ *
+ * <p>Use {@code NameScope} to group related operations within a hierarchy, which for example lets
+ * tensorboard coalesce nodes for better graph visualizations.
+ *
+ * <p>This class is package private, user code creates {@link Scope} which internally delegates
+ * calls to an underlying {@code NameScope}.
+ *
+ * <p>This class is thread-safe.
```
asimshankar: Does it need to be? Do we want to encourage graph construction in multiple threads?
I'm tempted to suggest that we:

- Change this to `<p>Objects of this class are NOT thread-save`
- Remove the use of `synchronized` in the implementation

It will be easier to make this thread-safe in the future if needed, but harder to remove the locks if they are a performance bottleneck. Or perhaps I'm overthinking this - it's just a single lock :) Thoughts?

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+ */
+final class NameScope {
+
+  NameScope withSubScope(String scopeName) {
+    if (baseName == null) {
+      checkPattern(OP_NAME_REGEX, scopeName);
```
asimshankar: Optional: I find this easier to read:

```
checkPattern(baseName == null ? OP_NAME_REGEX : SUBSCOPE_NAME_REGEX, scopeName);
```

but maybe it's just me. Your call.

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+  // noted in the proto definition core/framework/node_def.proto for NodeDef and actually
+  // implemented in core/framework/node_def_util.cc [Note that the proto comment does not include
+  // dash (-) in names, while the actual implementation permits it. These regexs follow the actual
+  // implementation.]
+  //
+  // These two patterns are used to ensure fully qualified names always start with a
+  // LETTER_DIGIT_DOT, followed by zero or more LETTER_DIGIT_DASH_DOT_SLASH_UNDERSCORE. SLASH is not
+  // permitted in actual user-supplied names to NameScope - it is used as a reserved character to
+  // separate subcomponents within fully qualified names.
+  private static final Pattern OP_NAME_REGEX = Pattern.compile("[A-Za-z0-9.][A-Za-z0-9_.\\-]*");
```
asimshankar: Isn't this supposed to be a slash (`/`) and not a backslash (`\`)? So: `/` instead of `\\`?

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+  private static final Pattern SUBSCOPE_NAME_REGEX = Pattern.compile("[A-Za-z0-9_.\\-]+");
```
asimshankar: Doesn't this regex make a single slash a valid subscope name. Should that be the case?
Actually, I'm not sure I understand why we need two regexps

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+
+  // The constraints for operator and scope names originate from restrictions on node names
+  // noted in the proto definition core/framework/node_def.proto for NodeDef and actually
+  // implemented in core/framework/node_def_util.cc [Note that the proto comment does not include
+  // dash (-) in names, while the actual implementation permits it. These regexs follow the actual
+  // implementation.]
+  //
+  // These two patterns are used to ensure fully qualified names always start with a
+  // LETTER_DIGIT_DOT, followed by zero or more LETTER_DIGIT_DASH_DOT_SLASH_UNDERSCORE. SLASH is not
+  // permitted in actual user-supplied names to NameScope - it is used as a reserved character to
```
asimshankar: IIUC, Python allows slashes in user-supplied names and this regexp also makes that possible. Why the comment?

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+    this(null, null, null);
+  }
+
+  private NameScope(String baseName, String baseOpName, Map<String, Integer> ids) {
+    this.baseName = baseName;
+    this.baseOpName = baseOpName;
+    if (ids != null) {
+      this.ids = ids;
+    } else {
+      this.ids = new HashMap<String, Integer>();
```
asimshankar: I just realized that if we want to allow users to name ops with slashes in them (as is possible in Python), then we'd probably need the name-map attached to the root scope and not create new hash-maps per sub-scope - i.e., a single map that covers all the nodes in the graph. I believe that is the case in the C++ scope as well.

Should we replicate that here?

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
asimshankar: Doh! Sorry, nevermind - misread the escaping :)

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
asimshankar: Doh! Sorry, nevermind - misread the escaping :)

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+  NameScope withSubScope(String scopeName) {
+    if (baseName == null) {
+      checkPattern(OP_NAME_REGEX, scopeName);
+    } else {
+      checkPattern(SUBSCOPE_NAME_REGEX, scopeName);
+    }
+
+    if (baseOpName != null) {
+      // Use the base name instead to derive the subscope.
+      scopeName = baseOpName;
```
asimshankar: This is a bit confusing in that `withSubScope` ignores the argument if `baseOpName` is `null`.
One alternative would be that `withSubScope` ignores the `baseOpName` and just creates a new sub-scope.
So something like:

```java
Const.create(s.withName("foo"), ...) // Results in "foo"
Const.create(s.withName("foo").withSubScope("foo"), ...)  // Results in "foo/Const" - as if the withName was ignored
```

That way it is clearer than `withName` only applies to the name (suffix) of the op and doesn't affect the scoping prefixes ever.

What do you think?

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  // If baseName is non-null, it is a prefix applied to all names
+  // created by this instance.
+  private final String baseName;
+
+  // If baseOpName is non-null, it is used to derive the unique name
+  // for operators rather than the provided default name.
+  private final String baseOpName;
```
asimshankar: How about `scopeName` and `opName` instead of `baseName` and `baseOpName`? Or perhaps `opPrefix` and `opName` - since one is intended to be a prefix for any ops created?

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
asimshankar: This makes `"-"` s valid subscope name (`withSubScope("-")`). Is that intentional? (Wondering if we can use a single regexp for now.

tensorflow/java/src/main/java/org/tensorflow/op/Scope.java
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * <p>An Operation building class acquires a Scope, and uses it to set properties on the underlying
+ * Tensorflow ops. For example:
+ *
+ * <pre>{@code
+ * // An operator class that adds a constant.
+ * public class Constant {
+ *   public static Constant create(Scope scope, ...) {
+ *      scope.graph().opBuilder(
+ *        "Constant", scope.makeOpName("Constant"))
```
asimshankar: To be faithful to the real op name, `Const` instead of `Constant`?

tensorflow/java/src/test/java/org/tensorflow/op/ScopeTest.java
```diff
@@ -0,0 +1,265 @@
...
+                .build()
+                .output(0));
+      }
+    }
+
+    private Const(Output o) {
+      output = o;
+    }
+
+    public Output output() {
```
asimshankar: Should we remove these `private` and `public` method qualifiers from the sample ops? (They are only needed for the tests and since they are inner classes, the test code is able to call them anyway (e.g., `Const.create` is private, but still callable - so the distinction between `private` and `public` here is immaterial anyway?)

Alternatively, make the `create` factory function `public` to be closer to where we want to head anyway?

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
kbsriram: If graph construction time seems a non-issue, we can close off the option entirely as you suggest.

One thought: should you be considering more of a wait-and-see approach - it may be less disruptive to improve a less performant but correct implementation than change it from "not thread-safe" to "thread-safe." The reason is that user-side code will likely need to restructure itself to take advantage in the latter case; but in the former case it will improve directly as the framework code improves.

For what it's worth, the locks here are per scope-instance; so to the extent different threads use different scopes there should be no contention. Aiui, possibly the place to watch out may be  [the per-graph global lock](https://github.com/tensorflow/tensorflow/blob/ce02c770fb269e5e607da459bde4f580ef108137/tensorflow/c/c_api_internal.h#L67) in the underlying c_api abstraction that guards operation creation itself.

asimshankar: I was thinking that making the API not promise thread-safety and bolting it on later will be less disruptive (as any correct usage of the API will not run into trouble) than the other way around.

Compounded with the fact that I don't quite see a pressing need for multi-threaded graph construction. Do you see any?

(I agree that the lock overhead itself may not be a big deal, but overall I'm trying to err on the side of simplicity instead of a feature for a use-case that we don't foresee)

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,148 @@
...
+ * <p>This class is package private, user code creates {@link Scope} which internally delegates
+ * calls to an underlying {@code NameScope}.
+ *
+ * <p>This class is <b>not</b> thread-safe.
+ */
+final class NameScope {
+
+  NameScope withSubScope(String scopeName) {
+    checkPattern(NAME_REGEX, scopeName);
+
```
asimshankar: Out of curiosity, why all the blank lines in these short functions?
Maybe this is my C/Go background talking, but compact is better in these cases? :)

(Feel free to leave as is or change, your call - I was just curious)

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,148 @@
...
+  // needed. This is a map containing names already created by this
+  // instance mapped to the next available numeric suffix for it.
+  private final Map<String, Integer> ids;
+
+  private static void checkPattern(Pattern pattern, String name) {
+    if (name == null) {
+      throw new IllegalArgumentException("Names cannot be null");
+    }
+    if (!pattern.matcher(name).matches()) {
+      throw new IllegalArgumentException(String.format("Invalid name '%s'", name));
```
asimshankar: Let's add some more detail to this error, perhaps something like:

"invalid name: '%s' does not match the regular expression NAME_REGEX"

asimshankar (changes requested):
Thanks for your patience, excited to have this in.

One minor comment about the error message - other than that I think this is good to go.

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
kbsriram: Wasn't a conscious choice :-) realize I habitually set off single-line comments with a blank line, and here probably that combined with removing unneeded blank lines. Out they go.

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
```diff
@@ -0,0 +1,147 @@
...
+
+  private static void checkPattern(Pattern pattern, String name) {
+    if (name == null) {
+      throw new IllegalArgumentException("Names cannot be null");
+    }
+    if (!pattern.matcher(name).matches()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "invalid name: '%s' does not match the regular expression %s",
+              name, NAME_REGEX_STRING));
```
asimshankar: I don't think you need NAME_REGEX_STRING, doesn't `NAME_REGEX.toString()` work just as well?

asimshankar (approved):
One comment about NAME_REGEX_STRING, otherwise LGTM

tensorflow/java/src/main/java/org/tensorflow/op/NameScope.java
kbsriram: Sure, that works too - ok to use pattern() instead, just to ensure it's clear what's being returned without referring to the documentation?

asimshankar: Sure.

asimshankar (approved):

kbsriram:
Gentle ping - could a member wake up the test this please ci bot?

martinwicke:
Jenkins, test this please!

asimshankar:
The single test failure (`//tensorflow/python:basic_session_run_hooks_test`) looks unrelated.

martinwicke:
Jenkins, test this please.

martinwicke:
Broken Mac test rebuilt here: http://ci.tensorflow.org/job/tensorflow-pull-requests-mac/5318/console

martinwicke:
Broken Makefile test rebuilt here: http://ci.tensorflow.org/job/tensorflow-pull-requests-makefile/9172/console

martinwicke:
Broken Android test rebuilt here: http://ci.tensorflow.org/job/tensorflow-pull-requests-android/5321/console

PR #10264, opened 2017-05-28
add missing import for `signal` package

Androbin:


drpngx (approved):

drpngx:
Jenkins, test this please.

drpngx:
I guess we need to change the cmakelists as well.

Androbin:
@drpngx Unfortunately, I don't know enough about CMake or the code organization to do this myself yet.

drpngx:
You'll need to add the module:
https://www.github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/tf_python.cmake#L467

Androbin:
@drpngx Thanks for the hint!

I noticed that the entry `session_bundle/testdata/saved_model_half_plus_two` doesn't exist (anymore).
This seems to be some kind of a consistency issue.

Do you think these entries could be put into a file separate from the code?
And would it be possible to programatically check them for validity?

drpngx:
Could you send a separate PR fixing the saved model issue?

/CC: @gunan @yifeif for the consistency check. @Androbin I don't have a great answer to that, but if you can propose something I'm all ears. Right now, CMake has been adding support piecemeal as needed, but as you found out, it's not perfect.

drpngx:
Thanks for the PR!

gunan:
I agree with @drpngx 's comment that cmake support is not perfect. It is also under contrib, so any contributions there are welcome.
Im not sure about separating those out of the cmake file though. Especially contrib is under very heavy development, and as things are added and removed people just forget updating all references to those.

Androbin:
@drpngx @gunan I get your point. It's just that even documentation changes are run through the entire build process and all the unit tests. And although it's `contrib`, installing proper testing allows for a more dynamic work flow. And since they are there to grow in a safe environment and maybe eventually get upgraded to `core` they should not be treated as second-class citizens when it comes to testing.

gunan:
For multiple reasons, bazel will stay as our primary build system.
cmake wont be promoted to core.
I am not against the test itself, I am happy to accept a test as you described as a contribution, and run it under contrib/cmake. But it is not a priority for us.

PR #10271, opened 2017-05-29
Remove r in docstrings that do not have backslashes

taehoonlee:


jhseu:
Thanks for the pull request! Not sure it hurts to have r there, since backslashes might easily be added later. Closing.

PR #10276, opened 2017-05-29
Support partial gets in MapStagingArea

sjperkins:
- size from `small` to `medium`
- introduce 2 shards

### DIFF
--- a/tensorflow/python/kernel_tests/BUILD
+++ b/tensorflow/python/kernel_tests/BUILD
@@ -2350,7 +2350,7 @@ cuda_py_test(
 
 cuda_py_test(
     name = "map_stage_op_test",
-    size = "small",
+    size = "large",
     srcs = ["map_stage_op_test.py"],
     additional_deps = [
         "//tensorflow/python:array_ops",
--- a/tensorflow/python/kernel_tests/BUILD
+++ b/tensorflow/python/kernel_tests/BUILD
@@ -2336,7 +2336,7 @@ cuda_py_test(
 
 cuda_py_test(
     name = "stage_op_test",
-    size = "small",
+    size = "medium",
     srcs = ["stage_op_test.py"],
     additional_deps = [
         "//tensorflow/python:array_ops",
@@ -2350,7 +2350,7 @@ cuda_py_test(
 
 cuda_py_test(
     name = "map_stage_op_test",
-    size = "large",
+    size = "medium",
     srcs = ["map_stage_op_test.py"],
     additional_deps = [
         "//tensorflow/python:array_ops",
--- a/tensorflow/python/kernel_tests/BUILD
+++ b/tensorflow/python/kernel_tests/BUILD
@@ -2336,7 +2336,7 @@ cuda_py_test(
 
 cuda_py_test(
     name = "stage_op_test",
-    size = "medium",
+    size = "small",
     srcs = ["stage_op_test.py"],
     additional_deps = [
         "//tensorflow/python:array_ops",
@@ -2346,12 +2346,11 @@ cuda_py_test(
         "//tensorflow/python:util",
         "//tensorflow/python:data_flow_ops",
     ],
-    shard_count = 2,
 )
 
 cuda_py_test(
     name = "map_stage_op_test",
-    size = "medium",
+    size = "small",
     srcs = ["map_stage_op_test.py"],
     additional_deps = [
         "//tensorflow/python:array_ops",
@@ -2361,7 +2360,6 @@ cuda_py_test(
         "//tensorflow/python:util",
         "//tensorflow/python:data_flow_ops",
     ],
-    shard_count = 2,
 )
 
 cuda_py_test(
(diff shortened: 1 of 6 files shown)


sjperkins:
@ekelsen I've updated this PR to support partial gets in the MapStagingArea. It was a use case I'd forgotten and it now mirrors the partial puts which is nicely symmetrical.

tensorflow/python/kernel_tests/BUILD
```diff
@@ -2351,7 +2351,7 @@ cuda_py_test(
 
 cuda_py_test(
     name = "map_stage_op_test",
-    size = "small",
+    size = "medium",
```
ekelsen: This is actually unneeded, but the bug causing the stage_op_test to occasionally hang does need to be fixed.  And verify if it affects the MapStagingArea as well.

tensorflow/python/ops/data_flow_ops.py
```diff
@@ -1919,16 +1961,19 @@ def peek(self, key, name=None):
...
+                      dtypes=dtypes,
                       name=name,
                       capacity=self._capacity,
                       memory_limit=self._memory_limit)
 
-    return self._get_return_value(result)
+    return self._get_return_value(result, indices)
 
-  def get(self, key=None, name=None):
+  def get(self, key=None, indices=None, name=None):
```
ekelsen: I think it's fine in this case since I don't think there are likely any users of MapStagingArea just yet (and definitely none internally), but generally adding new optional parameters should go at the end of the list, otherwise you break all callers that haven't been using named parameters.

sjperkins: Agreed. In this case I wanted to change the signature before before it made it into the final release.

sjperkins:
I ran `stage_op_test.py` 1000 times and saw no failures. One thing that came to immediately came to mind was that [test_util.test_session](https://github.com/tensorflow/tensorflow/blob/2eac53d5ea540b0b09326ba9330b6051d742532d/tensorflow/python/framework/test_util.py#L368-L377) states that test sessions are re-used unless an explicit graph is provided:

```python
Returns a TensorFlow Session for use in executing tests.
    This method should be used for all functional tests.
    This method behaves different than session.Session: for performance reasons
    `test_session` will by default (if `graph` is None) reuse the same session
    across tests. This means you may want to either call the function
    `reset_default_graph()` before tests, or if creating an explicit new graph,
    pass it here (simply setting it with `as_default()` won't do it), which will
    trigger the creation of a new session.
```

I've:

1. modified all the tests cases to explicitly create and finalize the compute graph (and hence recreate the Session in each case).
2. Separated some test cases that involved multiple graphs and sessions into separate tests.

I'll do some more test runs overnight (GMT +2) and see if I can reproduce any hangs.

sjperkins:
```
$ bazel test --test_output=all --jobs=1 --runs_per_test=1000 --test_sharding_strategy=disabled -c opt --config=cuda //tensorflow/python/kernel_tests:map_stage_op_test //tensorflow/python/kernel_tests:stage_op_test

...

INFO: Elapsed time: 3483.632s, Critical Path: 10.51s
//tensorflow/python/kernel_tests:map_stage_op_test                       PASSED in 2.1s
  Stats over 1000 runs: max = 2.1s, min = 1.6s, avg = 1.8s, dev = 0.1s
//tensorflow/python/kernel_tests:stage_op_test                           PASSED in 3.0s
  Stats over 1000 runs: max = 3.0s, min = 1.5s, avg = 1.7s, dev = 0.1s
```

I don't see anything in 1000 runs on either `stage_op_test` or `map_stage_op_test`. With what frequency are you seeing hangs?

ekelsen:
The stage_op_test hangs at least 10% of the time on the jenkins test server (map_stage_op_test seems fine).

sjperkins:
@ekelsen 5000 runs on `stage_op_test` have not revealed any problems thus far. `map_stage_op_test` is on 572/5000 runs on  now. Note this includes the changes I made to the Graph construction in https://github.com/tensorflow/tensorflow/pull/10276/commits/62948388bb350d0a136e333f93370e5e1cb1ae4b

ekelsen (approved):
LGTM, thanks!

sjperkins:
Thanks, I've attached the test output if you wish to double check: [bazel_test_output.zip](https://github.com/tensorflow/tensorflow/files/1056440/bazel_test_output.zip).

I've grepped on `fail` and `error` and seen nothing.

gunan:
Jenkins, test this please.

jhseu:
Jenkins, test this please

sjperkins:
I put in a fix for the python 2/3 range list/class change. Its very late here, I'll check back in in the morning.

jhseu:
map_stage_op_test seems to time out. Trying again.

Jenkins, test this please

jhseu:
Chatted with @ekelsen, and this timeout is reproducible with:
`CUDA_CACHE_DISABLE=1 bazel test --test_env=CUDA_CACHE_DISABLE --nocache_test_results --config=cuda //tensorflow/python/kernel_tests:map_stage_op_test`

Setting the test size to large and merging for now. If you have cycles, might be worth reducing the JIT compilation time in the test.

jhseu:
Jenkins, test this please

jhseu:
Actually, the test times should be fine. They're ~60s on my desktop with CUDA_CACHE_DISABLE set.

Jenkins, test this please

PR #10298, opened 2017-05-30
PhiloxRandom: Fix race in GPU fill function

jthestness:
The PhiloxRandom fill kernel for the GPU had race conditions that caused the
outputs to be non-deterministic. In particular, the code previously executed
with N GPU threads (# thread contexts per GPU), but it would only advance the
fill addresses by N-1 stride in each step. This incorrect stride caused the
0th and N-1st threads to write to the same memory locations, racing for which
was last to write their common locations. Make the stride equal to the number
of threads to eliminate the race.

BONUS: By fixing this race, PhiloxRandom constant-sized GPU initializers now
match CPU initializers.

### DIFF
--- a/tensorflow/core/kernels/random_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/random_op_gpu.cu.cc
@@ -141,7 +141,7 @@ struct FillPhiloxRandomKernel<Distribution, false> {
       const typename Distribution::ResultType samples = dist(&gen);
       copier(&data[offset], samples);
 
-      offset += (total_thread_count - 1) * kGroupSize;
+      offset += total_thread_count * kGroupSize;
       gen.Skip(total_thread_count - 1);
     }
 
--- a/tensorflow/python/kernel_tests/random_ops_test.py
+++ b/tensorflow/python/kernel_tests/random_ops_test.py
@@ -66,7 +66,8 @@ class RandomNormalTest(test.TestCase):
     for dt in dtypes.float16, dtypes.float32, dtypes.float64:
       results = {}
       for use_gpu in [False, True]:
-        sampler = self._Sampler(1000, 0.0, 1.0, dt, use_gpu=use_gpu, seed=12345)
+        sampler = self._Sampler(
+            1000000, 0.0, 1.0, dt, use_gpu=use_gpu, seed=12345)
         results[use_gpu] = sampler()
       if dt == dtypes.float16:
         self.assertAllClose(results[False], results[True], rtol=1e-3, atol=1e-3)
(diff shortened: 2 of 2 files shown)


tensorflow/core/kernels/random_op_gpu.cu.cc
```diff
@@ -141,7 +141,7 @@ struct FillPhiloxRandomKernel<Distribution, false> {
       const typename Distribution::ResultType samples = dist(&gen);
       copier(&data[offset], samples);
 
-      offset += (total_thread_count - 1) * kGroupSize;
+      offset += total_thread_count * kGroupSize;
       gen.Skip(total_thread_count - 1);
```
ekelsen: I think you also want to skip total_thread_count as well.

ekelsen:
Would be great if you could also add a test that would fail before your change and works after.

tensorflow/core/kernels/random_op_gpu.cu.cc
jthestness: I thought so too, but when I changed this, it gave different results than the CPU version of the initializer. Somehow, the generator does not generate repeated values even though this code appears to only skip `total_thread_count - 1` indices in the randomized sequence (i.e. seems like we should expect that the 0th and `total_thread_count - 1`st threads would produce the same results in successive steps). Perhaps there is an off-by-one issue inside the generator code that is consistent between CPU and GPU versions?

Do you have thoughts/recommendations for how to proceed here?

ekelsen: Hmm...you've verified that the 0th thread on the 2nd pass through the loop doesn't generate the same result as the last thread on the first pass?  It seems like that must be the case.  I'd like to understand how that's possible.

jthestness: Yes, I verified that. The PhiloxRandom::Skip function appears correct, so I'm also not sure how there can be an off-by-one in there.

ekelsen: Can you examine the state of the kernel for those two threads using cuda-gdb or printfs?  I'd like to make sure we understand what is actually going on - there's probably something we're missing.

jthestness:
@ekelsen: I updated the existing `random_ops_test.py` to manifest the race condition

tensorflow/core/kernels/random_op_gpu.cu.cc
jthestness: Hah! Silly: The generator automatically advances one position when it is read. Thus, when you skip after that, you only need to skip `total_thread_count - 1`

ekelsen:
Haha, good catch!  LGTM

ekelsen (approved):

jhseu:
Jenkins, test this please

PR #10304, opened 2017-05-30
Batch norm docs -> r1.1

nealwu:
Fixed up the documentation for tf.contrib.layers.batch_norm and tf.layers.batch_normalization explaining how to include the update ops in the train_op. Also simplified the existing tf.layers.batch_normalization documentation to the least error-prone option (adding to the train_op).

PiperOrigin-RevId: 156609483

See https://github.com/tensorflow/tensorflow/commit/675f0f9073e9e6ac6df8018e972f2b38f2c4c23c for the same change in master.

### DIFF
--- a/tensorflow/contrib/layers/python/layers/layers.py
+++ b/tensorflow/contrib/layers/python/layers/layers.py
@@ -387,16 +387,18 @@ def batch_norm(inputs,
 
   Can be used as a normalizer function for conv2d and fully_connected.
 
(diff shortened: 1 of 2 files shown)


gunan:
We probably wont touch this branch anymore, as 1.2 is around the corner.
Is it OK if we only cherrypick this into r1.2?

jhseu:
Yes, please recreate pulling this change into r1.2.

PR #10307, opened 2017-05-30
Update docker to cudnn6.

yifeif:


gunan:
I think python 3.4 is causing issues in your builds.
Could you check what is going on?

av8ramit (approved):

yifeif:
@damienmg @davidzchen  we are getting the following, but we don't see any cuda configuration being hard-coded to cudnn 5. Any idea why we are seeing the error? Thanks.
ERROR: /var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/local_config_cuda/cuda/BUILD:1302:1: declared output 'external/local_config_cuda/cuda/lib/libcudnn.so.5' is a dangling symbolic link.

yifeif:

yifeif:

configure
```diff
@@ -522,6 +522,12 @@ while true; do
   CUDA_TOOLKIT_PATH=""
 done
 
+# Set default CUDA version if not set
+if [ -z "$TF_CUDA_VERSION" ]; then
+  TF_CUDA_VERSION="8.0"
+  export TF_CUDA_VERSION
```
gunan: I think I know what is going on.
Simply exporting it wont help. You need the value written to bazelrc, with this command:
`write_action_env_to_bazelrc "TF_CUDA_VERSION" "$TF_CUDA_VERSION"`
However, you need to make sure that we do not have two conflicting values are not written, so it may complicate things a little.

configure
```diff
@@ -626,6 +632,12 @@ while true; do
   CUDNN_INSTALL_PATH=""
 done
 
+# Set default CUDNN version if not set
+if [ -z "$TF_CUDNN_VERSION" ]; then
```
gunan: same as above, this has to be written to bazelrc.

gunan (changes requested):

damienmg:
The variable that is used in cuda_configure.bzl starts with an underscore
so that would be _TF_CUDNN_VERSION.

configure
yifeif: Fixed accordingly.

yifeif:
we seems to also set _TF_CUDNN_VERSION to "TF_CUDNN_VERSION" :)

configure
```diff
@@ -522,6 +521,13 @@ while true; do
   CUDA_TOOLKIT_PATH=""
 done
 
+# Set default CUDA version if not set
+if [ -z "$TF_CUDA_VERSION" ]; then
+  TF_CUDA_VERSION="8.0"
+  export TF_CUDA_VERSION 
+fi
+write_action_env_to_bazelrc "TF_CUDA_VERSION" "$TF_CUDA_VERSION"
```
gunan: Let's also reflect this behavioral change in line 469. The part in the parenthesis can say (use empty to default to CUDA 8.0)

configure
```diff
@@ -626,6 +631,13 @@ while true; do
   CUDNN_INSTALL_PATH=""
 done
 
+# Set default CUDNN version if not set
+if [ -z "$TF_CUDNN_VERSION" ]; then
+  TF_CUDNN_VERSION="6"
+  export TF_CUDNN_VERSION
+fi
+write_action_env_to_bazelrc "TF_CUDNN_VERSION" "$TF_CUDNN_VERSION"
```
gunan: Ditto above, line 564 this time. Default to cuDNN 6.0

gunan (changes requested):

configure
yifeif: Good point. Updated.

gunan (approved):

PR #10312, opened 2017-05-31
Fix comments error in mnist_replica.py where only one ps is used with…

cxxgtxy:
… two works by default.


caisq (changes requested):

cxxgtxy:
I have changed the format.

caisq:
Thanks, @cxxgtxy !

caisq (approved):

Issue #10236, opened 2017-05-27
Remove "bazel clean" after new release


Androbin:
In the new release of bazel, #8880 was fixed which required `bazel clean` at the end of `configure`.
```
# TODO(gunan): Remove once bazel correctly handles changes in remote repositories.
bazel clean
```
Could @gunan look if this is save to remove now?

PR #10318, opened 2017-05-31
Remove "bazel clean"

Androbin:
Reverting #8880 @gunan (see #10236)
unnecessary since bazelbuild/bazel#2759 was merged

### DIFF
--- a/configure
+++ b/configure
@@ -844,6 +844,4 @@ if [ "$TF_NEED_MPI" == "1" ]; then
 fi
 
 
-# TODO(gunan): Remove once bazel correctly handles changes in remote repositories.
-bazel clean
 echo "Configuration finished"


damienmg (approved):

gunan:
Jenkins, test this please

gunan:
During upgrade to cudnn 6, we discovered another issue.
https://github.com/bazelbuild/bazel/issues/3062
Due to this, we may need to hold on to bazel clean a little while longer.
And even expand it by running `bazel clean --expunge`

damienmg:
Bazel clean --expunge at each reconfigure is really extreme I would
recommend against it. Especially since there is an easy​ workaround if you
really want it to force a refetch of cuda configure: add an environment
variable to the list of variable to invalidate on, say
cuda_configure_version, that you increase when you want a refetch of that
repository. Cuda configure is relatively cheap so even if you make it that
all the time, it shouldn't be a performance issue unless the output of cuda
configure change.

gunan:
@damienmg I understand the concern around the cost of clean --expunge.
however, in #10307 we are trying to use the existing environment variables to invalidate the cuda repo and get it to be rebuilt, with no results.
We tried to edit configure script to Provide defaults for `TF_CUDNN_VERSION`, where it was "" before, now we provided "6", but local_config_cuda still did not get rebuilt.

Could you chime in on that other pull request to direct us towards a working solution?

gunan (approved):

PR #10333, opened 2017-05-31
[OpenCL] Cleans variable op

lukeiwanski:


tensorflow/core/kernels/variable_ops.cc
```diff
@@ -32,36 +32,6 @@ REGISTER_KERNEL_BUILDER(Name("DestroyTemporaryVariable").Device(DEVICE_CPU),
 REGISTER_KERNEL_BUILDER(Name("IsVariableInitialized").Device(DEVICE_CPU),
                         IsVariableInitializedOp);
 
-#if TENSORFLOW_USE_SYCL
```
gunan: Any reason for moving the block?
This change looks like a noop too.

lukeiwanski: It was fixing the order of registrations: CPU, GPU, SYCL to be consistent with the rest of TF.
I am happy to put it back where it was.

gunan (approved):

gunan:
Jenkins, test this please.

PR #10344, opened 2017-05-31
[OpenCL] Cleans sequence_ops.cc

lukeiwanski:


tensorflow/core/kernels/sequence_ops.cc
```diff
@@ -112,6 +107,14 @@ TF_CALL_int64(REGISTER_GPU_KERNEL);
 
 #endif  // GOOGLE_CUDA
 
+#ifdef TENSORFLOW_USE_SYCL
```
gunan: Why do we need to move these? I feel like this change was a noop?

lukeiwanski: Move mos intended to keep CPU, GPU, SYCL order of registration. It's not essential.

gunan: Then in that case, I will close this PR.

Issue #10299, opened 2017-05-30
document how to use selective_registration and the use of __ANDROID_FULL_TYPE__


Lakedaemon:
Every developer wants to use his own model. 
Most app built for Mobile will crash at runtime, because those model won't just use the sparse subset of ops shipped with the aar available from jcenter 

Which means, that mobile developers need a way to easily and painlessly cross compile tenserflow with the right ops for their custom models. 

The print_header_for_selecrtive_registration.py  script is a good step in the right direction for this but it is not documented. 

The documentation is completely lacking. Please document how to cross-compile tenserflow for mobile with the right types, and the right ops. With want command ? what files should we modify, where ? 

I spent days looking at the tensorflow code/build files trying things and I still could not build a binary that would make that annoying "Op wasn't registered issue" away

Fix this ! This is not an individual problem, look at the amount of questions and issues about "tenserflow and Op wasn't registered" on the internet.

btw, the tensorflow aar from jcenter is completely useless as it can only be used to build the demo app.
It would be better to put the demo apk on Google play, it would not mislead developers into thinking they can easily build apps with it

asimshankar:
CC/ @petewarden @andrewharp

Androbin:
Couldn't agree more! At least the following should work:
`bazel build -c opt --copt="-DSELECTIVE_REGISTRATION" //tensorflow/contrib/android:libtensorflow_inference.so --config=android_arm`

Androbin:
current reference for selective registration: [core/framework/selective_registration.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/selective_registration.h)

andrewharp:
@cwhipkey re: selective registration docs.

@Lakedaemon If there are useful ops that you think should be included in the default set please feel free to send a PR adding them. It's our intention for the AAR to be useful for the majority of common models and a stepping stone before developers build their own size-optimized TF library.

tensorflowbutler:
It has been 14 days with no activity and the `awaiting tensorflower` label was assigned. Please update the label and/or status accordingly.

tensorflowbutler:
It has been 14 days with no activity and the `awaiting tensorflower` label was assigned. Please update the label and/or status accordingly.

Androbin:
@andrewharp @cwhipkey Any updates regarding documentation for selective registration yet?

tensorflowbutler:
[...]

PR #10351, opened 2017-05-31
Improve docs for selective registration headers

Androbin:
progressing #10299 @andrewharp

### DIFF
--- a/tensorflow/python/tools/print_selective_registration_header.py
+++ b/tensorflow/python/tools/print_selective_registration_header.py
@@ -25,7 +25,9 @@ core/framework/selective_registration.h for more details.
 
 When compiling for Android:
   bazel build -c opt --copt="-DSELECTIVE_REGISTRATION" \
+    --copt="-DSUPPORT_SELECTIVE_REGISTRATION" \
     //tensorflow/contrib/android:libtensorflow_inference.so \
+    --host_crosstool_top=@bazel_tools//tools/cpp:toolchain \
     --config=android_arm
 """
 
--- a/tensorflow/python/tools/print_selective_registration_header.py
+++ b/tensorflow/python/tools/print_selective_registration_header.py
@@ -14,13 +14,19 @@
 # ==============================================================================
 r"""Prints a header file to be used with SELECTIVE_REGISTRATION.
 
-Example usage:
-  print_selective_registration_header \
-      --graphs=path/to/graph.pb > ops_to_register.h
+An example of command-line usage is:
+  bazel build tensorflow/python/tools:print_selective_registration_header && \
+  bazel-bin/tensorflow/python/tools:print_selective_registration_header \
+    --graphs=path/to/graph.pb > ops_to_register.h
 
-  Then when compiling tensorflow, include ops_to_register.h in the include
-  search path and pass -DSELECTIVE_REGISTRATION  - see
-  core/framework/selective_registration.h for more details.
+Then when compiling tensorflow, include ops_to_register.h in the include
+search path and pass -DSELECTIVE_REGISTRATION  - see
+core/framework/selective_registration.h for more details.
+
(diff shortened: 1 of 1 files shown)


tensorflow/python/tools/print_selective_registration_header.py
```diff
@@ -14,13 +14,21 @@
...
+An example of command-line usage is:
+  bazel build tensorflow/python/tools:print_selective_registration_header && \
+  bazel-bin/tensorflow/python/tools:print_selective_registration_header \
+    --graphs=path/to/graph.pb > ops_to_register.h
 
-  Then when compiling tensorflow, include ops_to_register.h in the include
-  search path and pass -DSELECTIVE_REGISTRATION  - see
-  core/framework/selective_registration.h for more details.
+Then when compiling tensorflow, include ops_to_register.h in the include
+search path and pass -DSELECTIVE_REGISTRATION  - see
```
cwhipkey: -DSELECTIVE_REGISTRATION and -DSUPPORT_SELECTIVE_REGISTRATION here ?

Androbin: Of course (my fault), which brings up the question (again), why there even is this disconnect? Since the original docs only mention `-DSELECTIVE_REGISTRATION`, it seems that `-DSUPPORT_SELECTIVE_REGISTRATION` came up later in the process.

cwhipkey (approved):

cwhipkey:
Regarding why there are two macros, some of it is historical:
- originally, on mobile platforms, a reduced set of types were supported (see register_types.h, for example).
- then we added selective registration support via ops_to_register.h
- but then, when using selective registration, you might as well include all kernel types, since unused kernels are stripped.  So SUPPORT_SELECTIVE_REGISTRATION was added.

Probably we could make SELECTIVE_REGISTRATION imply SUPPORT_SELECTIVE_REGISTRATION...

When building the binary in pieces (like bazel would), it may be useful to have SUPPORT_SELECTIVE_REGISTRATION available as a standalone option - so that the framework+core kernels can all be built with that macro, without requiring passing SELECTIVE_REGISTRATION+including a header.  Although that can probably be eliminated as well, and just use SELECTIVE_REGISTRATION there.

jhseu:
Jenkins, test this please

gunan:
Jenkins, test this please.

nimatter:
Is SUPPORT_SELECTIVE_REGISTRATION supposed to make all ops available in the build, or just all the types needed to support all ops?  When using SELECTIVE_REGISTRATION and a model that includes ops not built by default, I still have to modify tensorflow/core/kernels/BUILD to add files under android_extended_ops_group1.  I would have expected SUPPORT_SELECTIVE_REGISTRATION to eliminate the need to edit any BUILD files in order to include an op such as Sin or Cos.

Androbin:
@nimatter Did you create and include `ops_to_register.h`?

nimatter:
@Androbin yes, I created and included `ops_to_register.h`.  For my procedure, please see my response [here](https://stackoverflow.com/questions/51105847/how-to-include-contrib-op-in-tensorflow-bazel-build/52170008#52170008) on StackOverflow.  I was searching github one more time while considering submitting an issue.

Androbin:
AFAIK, you need to pass both `-DSELECTIVE_REGISTRATION` and `-DSUPPORT_SELECTIVE_REGISTRATION` (for historical reasons).

Androbin:
Would you mind sharing (a minimal version of) your graph in `.pbtxt` format (via Pastebin)?
What transformations do you apply to the graph? (Freezing, Quantizing, Stripping)
Where do you use `tf.asin`? (sometimes there are issues towards the end of the graph, or with certain ops like `tf.case` or `tf.identity`)

Issue #9906, opened 2017-05-15
tf.import_graph_def() restricts the order of nodes in graph proto.


sj6077:
-----------------------
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Linux Ubuntu 14.04
- **TensorFlow installed from (source or binary)**:
source
- **TensorFlow version (use command below)**:
commit 3bee923c9
- **Bazel version (if compiling from source)**:
0.4.5
- **CUDA/cuDNN version**:
cuda 8.0/cudnn 5.1.5
- **GPU model and memory**:
Tesla P40 
- **Exact command to reproduce**:

### Describe the problem
I used auto parallel optimizer in grappler for cifar10 example, but grappler changes the order of nodes in graph proto. It causes the failure of import_graph_def, because ops.set_shapes_for_outputs(op), which is line 407 of tensorflow/tensorflow/python/framework/importer.py is failed. This is because all of the input nodes must be defined before to define a node. importer restricts the order of nodes for graph definition proto, but I think importer shoud be flexible to the order of nodes. 

### Source code / logs
  ```
with tf.Graph().as_default() as graph:
    #global_step = tf.contrib.framework.get_or_create_global_step()
    global_step = tf.get_variable(
        'global_step', [], dtype=tf.int32,
        initializer=tf.constant_initializer(0), trainable=False)

    # Get images and labels for CIFAR-10.
    images, labels = cifar10.distorted_inputs()

    # Build a Graph that computes the logits predictions from the
    # inference model.
    logits = cifar10.inference(images)

    # Calculate loss.
    loss = cifar10.loss(logits, labels)

    # Build a Graph that trains the model with one batch of examples and
    # updates the model parameters.
    train_op = cifar10.train(loss, global_step)

    init_op = tf.global_variables_initializer()

    queue_runners = []
    for qr in ops.get_collection(ops.GraphKeys.QUEUE_RUNNERS):
      queue_runners.append(qr.to_proto())

    mg = meta_graph.create_meta_graph_def(graph=graph)

  #Auto-parallel
  rewriter_config = rewriter_config_pb2.RewriterConfig()
  rewriter_config.optimizers.append('autoparallel')
  rewriter_config.auto_parallel.num_replicas = FLAGS.num_gpus

  graph_def = tf_optimizer.OptimizeGraph(rewriter_config, mg)

[...]

PR #10368, opened 2017-06-01
Fixes of AutoParallel bug

sj6077:
This PR fixes the bug that AutoParallel sets variables as don't_replicate_nodes, but not for variable initializers and snapshot nodes. This can break variable collections in meta graph because variable initializers and snapshot nodes could be replicated with different names.

This issue is already discussed in #9906.

### DIFF
--- a/tensorflow/core/grappler/grappler_item.h
+++ b/tensorflow/core/grappler/grappler_item.h
@@ -50,7 +50,7 @@ struct GrapplerItem {
   std::vector<QueueRunnerDef> queue_runners;
 
   // Variables of this model.
-  std::vector<VariableDef> variables;
+  std::vector<VariableDef> variable_def;
 
   // Return the set of node evaluated during a regular train/inference step.
   std::vector<const NodeDef*> MainOpsFanin() const;
--- a/tensorflow/core/grappler/grappler_item.h
+++ b/tensorflow/core/grappler/grappler_item.h
@@ -49,9 +49,6 @@ struct GrapplerItem {
   // Queue runner(s) required to run the queue(s) of this model.
   std::vector<QueueRunnerDef> queue_runners;
 
-  // Variables of this model.
-  std::vector<VariableDef> variable_def;
-
   // Return the set of node evaluated during a regular train/inference step.
   std::vector<const NodeDef*> MainOpsFanin() const;
   // Return the set nodes used by TensorFlow to initialize the graph.
--- a/tensorflow/core/grappler/grappler_item.h
+++ b/tensorflow/core/grappler/grappler_item.h
@@ -23,6 +23,7 @@ limitations under the License.
 
 #include "tensorflow/core/framework/graph.pb.h"
 #include "tensorflow/core/framework/tensor.h"
+#include "tensorflow/core/framework/variable.pb.h"
 #include "tensorflow/core/protobuf/queue_runner.pb.h"
 
 namespace tensorflow {
@@ -47,6 +48,9 @@ struct GrapplerItem {
 
   // Queue runner(s) required to run the queue(s) of this model.
   std::vector<QueueRunnerDef> queue_runners;
+
+  // Variables of this model.
+  std::vector<VariableDef> variables;
 
   // Return the set of node evaluated during a regular train/inference step.
   std::vector<const NodeDef*> MainOpsFanin() const;
--- a/tensorflow/core/grappler/optimizers/auto_parallel.cc
+++ b/tensorflow/core/grappler/optimizers/auto_parallel.cc
@@ -97,7 +97,7 @@ Status AutoParallel::Initialize(const GrapplerItem& item) {
     VLOG(2) << "Variable: " << var->name();
   }
 
-  for (const auto& var : item.variables) {
+  for (const auto& var : item.variable_def) {
     variables_.insert(std::make_pair(var.variable_name(), var));
   }
 
@@ -172,12 +172,10 @@ Status AutoParallel::Initialize(const GrapplerItem& item) {
     dont_replicate_nodes.insert(variable->name());
     VariableDef var = variables_[strings::StrCat(variable->name(), ":0")];
     if (!var.initializer_name().empty()) {
-      dont_replicate_nodes.insert(
-        var.initializer_name().substr(0, var.initializer_name().size() - 2));
+      dont_replicate_nodes.insert(NodeName(var.initializer_name()));
     }
     if (!var.snapshot_name().empty()) {
-      dont_replicate_nodes.insert(
-        var.snapshot_name().substr(0, var.snapshot_name().size() - 2));
+      dont_replicate_nodes.insert(NodeName(var.initializer_name()));
     }
   }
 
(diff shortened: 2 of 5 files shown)


martinwicke:
Benoit, are you reviewing this?

zhangyaobit:
Thanks Martin. I'm taking a look.

zhangyaobit:
@sj6077, Thanks for the contribution! I'm not sure how updating initializers and snapshot nodes inside C++ auto_parallel.cc would help, as the C++ code doesn't return the modified metagraph back. Currently they have to be on the python side, for example, like [this](https://gist.github.com/zhangyaobit/08c0b776e4b7a4f786714d46e723ee8a).

We do plan to come up with a better solution that avoids importing and exporting graph and thus directly uses session.run with ConfigProto to specify the optimizers (currently this is not possible because metagraph is not passed to the optimizer, that's why the ptb code has to use the tf_optimizer.OptimizeGraph interface). 

However, once we solved the problem of passing metagraph to the optimizer, we don't need to use import/export or tf_optimizer any more and I think it may make sense to update initializers and snapshot nodes inside C++ auto_parallel.cc and we can revisit this issue.

I don't know about your use case. Could you provide some skeleton code/usage pattern to show how this change would help? Thanks.

sj6077:
```
class UpdateCollection(object):
  """Update collection info in MetaGraphDef for AutoParallel optimizer."""

  def __init__(self, metagraph, model):
    self._metagraph = metagraph
    self.replicate_states(model.initial_state_name)
    self.replicate_states(model.final_state_name)
    self.update_snapshot_name("variables")
    self.update_snapshot_name("trainable_variables")

  def update_snapshot_name(self, var_coll_name):
    var_list = self._metagraph.collection_def[var_coll_name]
    for i, value in enumerate(var_list.bytes_list.value):
      var_def = variable_pb2.VariableDef()
      var_def.ParseFromString(value)
      # Somehow node Model/global_step/read doesn't have any fanout and seems to
      # be only used for snapshot; this is different from all other variables.
      if var_def.snapshot_name != "Model/global_step/read:0":
        var_def.snapshot_name = with_autoparallel_prefix(
            0, var_def.snapshot_name)
      value = var_def.SerializeToString()
      var_list.bytes_list.value[i] = value

  def replicate_states(self, state_coll_name):
    state_list = self._metagraph.collection_def[state_coll_name]
    num_states = len(state_list.node_list.value)
    for replica_id in range(1, FLAGS.num_gpus):
      for i in range(num_states):
        state_list.node_list.value.append(state_list.node_list.value[i])
    for replica_id in range(FLAGS.num_gpus):
      for i in range(num_states):
        index = replica_id * num_states + i
        state_list.node_list.value[index] = with_autoparallel_prefix(
            replica_id, state_list.node_list.value[index])
```
According to the example that you mentione, there is update_snapshot_name for metagraph collection. This is because variable, variable_snapshot, variable_initializer are all together defined for a variable, but snapshot and initializer are differently managed in auto_parallel. With this PR, the update_snapshot_name could be removed.

tensorflow/core/grappler/optimizers/auto_parallel.cc
```diff
@@ -166,7 +170,17 @@ Status AutoParallel::Initialize(const GrapplerItem& item) {
   std::set<string> dont_replicate_nodes;
   for (const auto& variable : item.MainVariables()) {
     dont_replicate_nodes.insert(variable->name());
+    VariableDef var = variables_[strings::StrCat(variable->name(), ":0")];
+    if (!var.initializer_name().empty()) {
+      dont_replicate_nodes.insert(
+        var.initializer_name().substr(0, var.initializer_name().size() - 2));
```
zhangyaobit: Could you use this function here? https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/grappler/utils.cc#L109

tensorflow/core/grappler/optimizers/auto_parallel.cc
```diff
@@ -166,7 +170,17 @@ Status AutoParallel::Initialize(const GrapplerItem& item) {
...
+    }
+    if (!var.snapshot_name().empty()) {
+      dont_replicate_nodes.insert(
+        var.snapshot_name().substr(0, var.snapshot_name().size() - 2));
```
zhangyaobit: Same here.

tensorflow/core/grappler/grappler_item.h
```diff
@@ -48,6 +49,9 @@ struct GrapplerItem {
   // Queue runner(s) required to run the queue(s) of this model.
   std::vector<QueueRunnerDef> queue_runners;
 
+  // Variables of this model.
+  std::vector<VariableDef> variables;
```
zhangyaobit: Can you use GrapplerItem:MainVariables()?

zhangyaobit (commented):
Ah, I see. Thanks for the clarification! I think this change is nice. Look forward to your more contributions :)

tensorflow/core/grappler/grappler_item.h
zhangyaobit: I see. This is to store variable_def; could you rename it to variables_def?

zhangyaobit: Actually, to "variable_def"

sj6077:
@zhangyaobit I addressed your comments, thank you for the review.

zhangyaobit (approved):

zhangyaobit:
Jenkins, test this please.

tensorflow/core/grappler/optimizers/auto_parallel.cc
```diff
@@ -166,7 +170,15 @@ Status AutoParallel::Initialize(const GrapplerItem& item) {
   std::set<string> dont_replicate_nodes;
   for (const auto& variable : item.MainVariables()) {
     dont_replicate_nodes.insert(variable->name());
+    VariableDef var = variables_[strings::StrCat(variable->name(), ":0")];
+    if (!var.initializer_name().empty()) {
+      dont_replicate_nodes.insert(NodeName(var.initializer_name()));
+    }
+    if (!var.snapshot_name().empty()) {
+      dont_replicate_nodes.insert(NodeName(var.initializer_name()));
```
benoitsteiner: You probably meant to at the snapshot_name to the list of nodes to avoid replicating here.

Snapshots are used as a way to prefetch the content of a variable onto a device. If you start creating multiple replicas, you probably want to prefetch the variables onto the devices where each replica is going to be executed. So you probably should replicate the variable snapshots.

zhangyaobit (changes requested):
Thanks for reviewing, Benoit. @sj6077, please try address Benoit's comments, thanks.

sj6077:
@benoitsteiner I addressed your comments, I didn't know about the usage of snapshot nodes.

benoitsteiner:
Jenkins, please test this.

benoitsteiner (approved):

zhangyaobit (approved):

benoitsteiner:
Jenkins, please test this.

benoitsteiner:
Jenkins, test this please.

PR #10372, opened 2017-06-01
Enable figures in the tfprof README.md

cNikolaou:
I don't know if excluding the figures from the tfprof README.md was intentional or not and I could not find any reasoning why they are excluded. 

If there is a reason, then ignore the PR.


Androbin:
The links should be "tensorflow/g3doc/..." instead of "g3doc/..."

panyx0718 (approved):

panyx0718:
LGTM. It's a syntax difference between internal/external doc.

jhseu:
Jenkins, test this please

PR #10381, opened 2017-06-01
Clarify tf.matmul documentation

orome:
Maybe I'm confused about what "inner" and "outer" tensor dimensions are, but the documentation for `tf.matmul` puzzles me:

> The inputs must be matrices (or tensors of rank > 2, representing
> batches of matrices), with matching inner dimensions, possibly after
> transposition.

Isn't it the case that R-rank arguments need to have matching (or no) R-2 outer dimensions, and that (as in normal matrix multiplication) the Rth, inner dimension of the first argument must match the R-1st dimension of the second. That is, in

    A = tf.constant(..., shape=[a, ..., z, p, x])
    B = tf.constant(..., shape=[a', ..., z', x', q]) 
    C = tf.matmul(A, B)

The outer dimensions `a, ..., z` must be identical to `a', ..., z'` (or not exist), and `x` and `x'` must match (while `p` and `q` can be anything).

Or put another way, shouldn't the docs say: 

> The inputs must, following any transpositions, be tensors of rank ≥ 2 where the inner 2 dimensions specify valid matrix multiplication arguments, and any further outer dimensions match.

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -1686,8 +1686,9 @@ def matmul(a,
            name=None):
   """Multiplies matrix `a` by matrix `b`, producing `a` * `b`.
 
-  The inputs must be matrices (or tensors of rank > 2, representing batches of
-  matrices), with matching inner dimensions, possibly after transposition.
+  The inputs must, following any transpositions, be tensors of rank ≥ 2 
+  where the inner 2 dimensions specify valid matrix multiplication arguments, 
+  and any further outer dimensions match.
 
   Both matrices must be of the same type. The supported types are:
   `float16`, `float32`, `float64`, `int32`, `complex64`, `complex128`.
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -1686,7 +1686,7 @@ def matmul(a,
            name=None):
   """Multiplies matrix `a` by matrix `b`, producing `a` * `b`.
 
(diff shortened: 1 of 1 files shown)


orome:
I signed it!

jhseu:
I agree the existing language is confusing, but the updated language is also difficult to parse. I think the example you gave is much clearer. Should we put that there instead? Something like:

For rank > 2:
```
A = tf.constant(..., shape=[a, ..., z, p, x])
B = tf.constant(..., shape=[a', ..., z', x', q]) 
C = tf.matmul(A, B)
```
The outer dimensions a, ..., z must be identical to a', ..., z', and x and x' must match.

jhseu:
Adding Andrew in case he has an opinion.

orome:
@jhseu : That could be added but would need to also make clear what transposed arguments also work, and that "matching" for  a, ..., z and a', ..., z' means they can both be absent.

I'm partial to using the proposed

> The inputs must, following any transpositions, be tensors of rank ≥ 2 where the inner 2 dimensions specify valid matrix multiplication arguments, and any further outer dimensions match.

Since even if it's a bit hard to parse at first, the example (which I agree should be added) clears it up.

jhseu (approved):
Coming back to this, the new wording is much clearer than the old wording.

jhseu:
Jenkins, test this please

jhseu:
Jenkins, test this please

jhseu:
I think the tests don't like the non-ascii `≥` character in there. Mind switching it to >=?

jhseu:
Actually, I just did it instead.

Jenkins, test this please

PR #10384, opened 2017-06-02
moved code generated by protoc-gen-go out of internal pkg so it can be referenced/used

ctava:
would like to call String() method on GraphDef. can't do so if its in the internal pkg. there are likely other use cases here. 

what do you think?


ctava:
@asimshankar can this be done in the sort term until the Go code generator to uses the same configuration file as the code generator for Python and C++ (and others)?

tensorflow/go/.gitignore
```diff
@@ -1,2 +1,2 @@
 # .pb.go files generated by generate.sh
-internal/proto/*
+pb/proto/*
```
asimshankar: IIUC, the `pb` directory is one level up, so should the `.gitignore` of the parent dir be changed instead?
Though see comment later on.

tensorflow/go/genop/generate.sh
```diff
@@ -46,8 +46,8 @@ fi
 # Ensure that protoc-gen-go is available in $PATH
 # Since ${PROTOC} will require it.
 export PATH=$PATH:${GOPATH}/bin
-mkdir -p ./internal/proto
+mkdir -p ../pb
```
asimshankar: It might be confusing to have `genop/generate.sh` generate files in `../pb`. Either we should have a `pb` directory that contains the script to generate the protos, or we keep the `pb` directory inside `genop`.

asimshankar (commented):
@ctava: Thanks for the PR.

We've tried to keep the API dependencies minimal and have avoided requiring users of the Go API depending on the protobufs (currently, the protobufs are only generated and used to create  the op wrappers - in normal course of `go get`, no protobufs are fetched).

That said, I understand that there are legitimate uses of the protobufs. Though, would you mind elaborating on your use - is it aimed at debugging (like `GraphDef.String()`), or is your intent to manipulate protobufs?

Do you want the `pb` directory to be part of the public API (in which case, I'd like to think about the organization of the code and the awkwardness of the import path - `github.com/tensorflow/tensorflow/tensorflow/go/pb/tensorflow/...` :)?

Alternatively, for your own purposes, you could import from `genop/internal/proto` directly, right?

tensorflow/go/genop/generate.sh
ctava: i agree that this could be confusing

ctava:
@asimshankar my goal is more than debugging its training / teaching people the go APIs for tensorflow. as such having a human readable non-binary export of tensorflow graph is very useful. e.g.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/graph_io.py#L69

drpngx:
Jenkins, test this please.

@asimshankar what do you think?

ctava:
docker workaround:

```
#Begin: install protoc
ENV PROTOC_VERSION 3.3.0
ENV PROTOC_LIB_ZIP protoc-$PROTOC_VERSION-linux-x86_64.zip
ENV TARGET_DIRECTORY /usr/local
RUN  curl -fsSL "https://github.com/google/protobuf/releases/download/v$PROTOC_VERSION/$PROTOC_LIB_ZIP" -o $PROTOC_LIB_ZIP && \
     sudo unzip $PROTOC_LIB_ZIP -d $TARGET_DIRECTORY && \
     rm -Rf $TPROTOC_LIB_ZIP
#End: install protoc

#Begin: generate go files from protobufs
RUN go get github.com/golang/protobuf/proto
RUN go get github.com/golang/protobuf/protoc-gen-go
ENV TF_DIR /go/src/github.com/tensorflow/tensorflow
ENV TF_PB_DIR /go/src/github.com/tensorflow/tensorflow/tensorflow/go/pb
RUN mkdir -p $TF_PB_DIR
RUN /usr/local/bin/protoc -I $TF_DIR \
  --go_out=$TF_PB_DIR \
  $TF_DIR/tensorflow/core/framework/*.proto
#End: generate go files from protobufs
```

ctava:
cgo program to list out ops (using docker workaround):
https://github.com/ctava/tensorflow-go-opslist/blob/master/main.go#L29

PR #10387, opened 2017-06-02
Update the instruction for building a minimal xla benchmark

freedomtan:
1. the instruction for building lstm_layer_inference_benchmark doesn't work.
2. the "cc_target_os" in "android_armeabi" is an unknown option

### DIFF
--- a/tensorflow/compiler/tests/BUILD
+++ b/tensorflow/compiler/tests/BUILD
@@ -523,8 +523,12 @@ cuda_py_test(
 # --dump_graph_dir, and the config file was written by hand.
 #
 # Run the following to build a minimal benchmark of the computation on Android:
-# $ bazel build -c opt --config=android_arm \
-#       third_party/tensorflow/compiler/tests:lstm_layer_inference_benchmark
+# $ bazel build -c opt --cxxopt='-std=c++11' --linkopt='-lm' \
+#   --cpu=armeabi-v7a \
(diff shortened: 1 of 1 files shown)


tensorflow/BUILD
```diff
@@ -39,7 +39,7 @@ config_setting(
 config_setting(
     name = "android_armeabi",
     values = {
-        "cc_target_os": "android",
+        "crosstool_top": "//external:android/crosstool",
```
tatatodd: While this is indeed what we want the BUILD file to look like, this isn't the correct way to make this change.

You haven't done anything wrong; we just have a separate (internal) mechanism to deal with BUILD file differences between internal and open-source builds.

The easiest way forward is for you to revert the change in this file, but keep the change in the other file.  I'll fix up the internal mechanism.  Thanks!

tatatodd (changes requested):
Thanks for the PR @freedomtan !

tensorflow/BUILD
freedomtan: Thanks for the review. Rebased and removed this.

tatatodd:
Jenkins, test this please.

tatatodd (approved):
Great, thanks @freedomtan !

tatatodd:
Jenkins, test this please.

tatatodd:
@martinwicke Can you merge this PR?  Thanks!

PR #10414, opened 2017-06-03
Solution to non context words due to wrap around

Humanity123:
Due to wrap around non-context words also appear in the window for skip gram which degrades the quality for word vectors.
Simple patch to avoid appearance of non-context words

### DIFF
--- a/tensorflow/examples/tutorials/word2vec/word2vec_basic.py
+++ b/tensorflow/examples/tutorials/word2vec/word2vec_basic.py
@@ -91,7 +91,6 @@ print('Sample data', data[:10], [reverse_dictionary[i] for i in data[:10]])
 
 data_index = 0
 
-
 # Step 3: Function to generate a training batch for the skip-gram model.
 def generate_batch(batch_size, num_skips, skip_window):
   global data_index
@@ -101,9 +100,10 @@ def generate_batch(batch_size, num_skips, skip_window):
   labels = np.ndarray(shape=(batch_size, 1), dtype=np.int32)
   span = 2 * skip_window + 1  # [ skip_window target skip_window ]
   buffer = collections.deque(maxlen=span)
-  for _ in range(span):
-    buffer.append(data[data_index])
-    data_index = (data_index + 1) % len(data)
+  if data_index + span > len(data):
+    data_index = 0
+  buffer.extend(data[data_index:data_index + span])
+  data_index += span
   for i in range(batch_size // num_skips):
     target = skip_window  # target label at the center of the buffer
     targets_to_avoid = [skip_window]
@@ -113,8 +113,12 @@ def generate_batch(batch_size, num_skips, skip_window):
       targets_to_avoid.append(target)
       batch[i * num_skips + j] = buffer[skip_window]
       labels[i * num_skips + j, 0] = buffer[target]
-    buffer.append(data[data_index])
-    data_index = (data_index + 1) % len(data)
+    if data_index == len(data):
+      buffer[:] = data[:span]
+      data_index = span
+    else:
+      buffer.append(data[data_index])
+      data_index += 1
   # Backtrack a little bit to avoid skipping words in the end of a batch
   data_index = (data_index + len(data) - span) % len(data)
   return batch, labels


Humanity123:
I have signed it

jhseu:
Geoffrey, mind taking a look?

girving:
@Humanity123 Can you explain what this does?  I don't follow the code.

Humanity123:
Since the original code didn't pay attention to wrap around. Non Context Words used to creep in. For Example if there are 100 words from 0 - 99 and the span of the window is 3 then there will be a case when the buffer contains 99 , 0 , 1
So to prevent this there is a simple check, if the buffer is not empty that is this is not the first word to be filled in buffer and  the next word to be filled is zero (0th index word) which means that there is a wrap around. We flush the items in the buffer and start from the beginning of the document that is we refill the buffer with the words from the starting of the document. This thus prevents  non context words to ever occur together.

tensorflow/examples/tutorials/word2vec/word2vec_basic.py
```diff
@@ -102,6 +110,9 @@ def generate_batch(batch_size, num_skips, skip_window):
   span = 2 * skip_window + 1  # [ skip_window target skip_window ]
   buffer = collections.deque(maxlen=span)
   for _ in range(span):
+    if len(buffer)!=0 and data_index == 0:
+      skip_to_start(buffer, skip_window)
+      break
     buffer.append(data[data_index])
     data_index = (data_index + 1) % len(data)
```
girving: I'm still not sure I follow the code, since you don't clear the buffer when you call `skip_to_start`.  If I'm correct that that's a bug, here's a much cleaner version of the whole loop:

    if data_index + span > len(data):
      data_index = 0
    buffer.extend(data[data_index:data_index + span])

girving: Note that unlike your version, which still loops around if the whole buffer is smaller than `span`, the smaller version would crash.  I think that's a reasonable policy.

tensorflow/examples/tutorials/word2vec/word2vec_basic.py
```diff
@@ -93,6 +93,14 @@ def build_dataset(words, n_words):
 
 
 # Step 3: Function to generate a training batch for the skip-gram model.
```
girving: Please don't move comments attached to one function to another function that you've just created.

tensorflow/examples/tutorials/word2vec/word2vec_basic.py
```diff
@@ -113,6 +124,9 @@ def generate_batch(batch_size, num_skips, skip_window):
       targets_to_avoid.append(target)
       batch[i * num_skips + j] = buffer[skip_window]
       labels[i * num_skips + j, 0] = buffer[target]
+    if len(buffer) != 0 and data_index == 0:
+      skip_to_start(buffer, skip_window)
+      continue
```
girving: How does this prevent wraparound?  It doesn't seem like you clear the buffer.

girving (changes requested):

tensorflow/examples/tutorials/word2vec/word2vec_basic.py
Humanity123: Yes it was a bug. It has been fixed now. The buffer is cleared if there is a wraparound

Humanity123: Yes you're shorter versions achieves the same effect, you just need to clear the buffer before extending it.

girving: Please use the shorter and clearer version.  Note that you don't need to clear the buffer for the first use, since it was created immediately before.  However, note that my shorter clearer version is slightly wrong.  It should be

    if data_index + span > len(data):
      data_index = 0
    buffer.extend(data[data_index:data_index + span])
    data_index += span

tensorflow/examples/tutorials/word2vec/word2vec_basic.py
Humanity123: Yes there was a bug. The buffer is now cleared if there is a wrap-around

girving: Replace this with

    if data_index == len(data):
      buffer[:] = data[:span]
      data_index = span
    else:
      buffer.append(data[data_index])
      data_index += 1

Then you should be able to delete your `skip_to_start` function.

girving (changes requested):

Humanity123:
Recommended Changes have been made.

girving (approved):

girving:
Jenkins, test this please.

girving:
Windows failure looks unrelated.  @jhseu: Go for merging.

girving:
@Humanity123 Thank you for the contribution!

PR #10418, opened 2017-06-03
Expose canned estimators in core

terrytangyuan:
I know this is still WIP but it would be good to expose this for adventurous external users.


martinwicke:
We're not quite ready for that, and we do not want to muddle the API compatibility story. 

Can you import directly, like

from tensorflow.python.estimators.canned import ...

?

terrytangyuan:
Apologies. I tried that earlier but didn't work. Turned out that I was using a different python lib. I am closing this. Totally agree with the importance of compatibility history.

PR #10444, opened 2017-06-05
Fix CMD in Dockerfile

inc0:
Currently Notebook fails execution because default user for this container is root, and unless explicitly allowed, jupyter notebook will not start.

### DIFF
--- a/tensorflow/examples/udacity/Dockerfile
+++ b/tensorflow/examples/udacity/Dockerfile
@@ -12,4 +12,4 @@ RUN pip install scikit-learn pyreadline Pillow
 RUN rm -rf /notebooks/*
 ADD *.ipynb /notebooks/
 WORKDIR /notebooks
-CMD ["/run_jupyter.sh"]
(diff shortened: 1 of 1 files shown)


vincentvanhoucke:
What changed that this no longer works?

vincentvanhoucke:
And does this mean we need to rebuild the existing containers?

inc0:
I'm not sure when it changed, however this file already has this change https://github.com/tensorflow/tensorflow/blob/bc4101eb4d1d088faaadb68991e4c177bdf249bd/tensorflow/tools/docker/Dockerfile#L69

Even if registry has new container that this one inherits FROM, CMD is being overridden so we need to change it here too.

Existing containers should be fine, if they worked they'll keep working.

vincentvanhoucke (approved):

jhseu:
Jenkins, test this please

inc0:
I'll need your help guys debugging this gate failure...can't see anything related to my change

jhseu:
Looks unrelated (a github failure).

Jenkins, test this please

Issue #10435, opened 2017-06-05
build optimize_for_inference fail


JcmeLs:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
    No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
    Windows 7 x64
- **TensorFlow installed from (source or binary)**:
    binary (pip3 install....)
- **TensorFlow version (use command below)**:
    tensorflow 1.1.0 CPU only
- **Bazel version (if compiling from source)**:
    Build label: 0.5.0
    Build target: bazel-out/msys_x64-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
    Build time: Fri May 26 12:12:09 2017 (1495800729)
    Build timestamp: 1495800729
    Build timestamp as int: 1495800729
  
    $ protoc --version
    libprotoc 3.2.0
- **CUDA/cuDNN version**:
    CPU Only

### Describe the problem
I want to build optimize_for_inference and optimize and would like to optimize my graph by the optimizer.But when I run ` bazel build tensorflow/python/tools:optimize_for_inference `.It will get me a problem--` Encountered error while reading extension file 'protobuf.bzl': no such package '@protobuf//'`.The problem occurs when i try to build anything with tensorflow as dependancy,like:`convert_graphdef_memmapped_format`.  
I've seen all the issues from github or stack overflow.And I tried all the fixes from issues or stackoverflow.But the problem still exists.
### Source code / logs
Source code:
``` 
bazel build tensorflow/python/tools:optimize_for_inference && bazel-bin/tensorflow/python/tools/optimize_for_inference --input=frozen_inception_graph.pb --output=optimized_inception_graph.pb --frozen_graph=True --input_names=Mul --output_names=softmax
```
logs:
``` 
D:\tensorflow-r1.2>bazel build tensorflow/python/tools:optimize_for_inference
[32mINFO: [0mLoading complete.  Analyzing...
[1A[K[32mINFO: [0mDownloading http://mirror.bazel.build/github.com/google/protobuf/archive/2b7430d96aeff2bb624c8d52182ff5e4b9f7f18a.tar.gz: 257,202 bytes
[1A[K[32mINFO: [0mDownloading http://mirror.bazel.build/github.com/google/protobuf/archive/2b7430d96aeff2bb624c8d52182ff5e4b9f7f18a.tar.gz: 501,196 bytes
[...]

PR #10452, opened 2017-06-06
Fix patching issue on Windows

meteorcloudy:
Fix the issue in https://github.com/tensorflow/tensorflow/issues/10435
@gunan


gunan (approved):

gunan:
I think I was too hasty merging this PR, now we have the merge issue on master.

meteorcloudy:
Thanks for taking care of this, looks like this commit is the wrong solution.

Issue #7541, opened 2017-02-15
Fatal messages mixing C libtensorflow with python tensorflow


pwaller:
I'm trying to write mixed C tensorflow code with python tensorflow code by
embedding the CPython interpreter in my application.

I'm mainly doing this because defining the model is only really possible in
Python at the moment due to the lack of gradients (#6268), and I want to define
new models from the C side at speed without needing to invoke or
communicate to an external python process to get a new model.

To reproduce the problem is quite straightforward, simply `import tensorflow`
in python after the libtensorflow library has already been dynamically linked.
Here is a quick reproducer in pure python which will not run:

```python
import ctypes

tf_dll = ctypes.CDLL("/usr/local/lib/libtensorflow.so")

import tensorflow
```

libtensorflow can be obtained like so:

```
TF_TYPE=cpu # Set to gpu for GPU support
TF_OS=linux
curl -L \
  "https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-${TF_TYPE}-${TF_OS}-x86_64-1.0.0.tar.gz" |
sudo tar -C /usr/local -xz
```

Here are two fatal messages I have encountered (the first from the Python reproducer above, the second from a C program):

```
F tensorflow/stream_executor/cuda/cuda_platform.cc:180] Check failed: ::perftools::gputools::port::Status::OK() == (MultiPlatformManager::RegisterPlatform(std::move(platform))) (OK vs. Internal: platform is already registered with name: "CUDA")
```

```
F tensorflow/core/lib/monitoring/collection_registry.cc:77] Cannot register 2 metrics with the same name: /tensorflow/cc/saved_model/load_attempt_count
```

I assume the problem is that the `_pywrap_tensorflow.so` has tensorflow
statically linked into them, so they don't use libtensorflow. Then you have
two shared libraries conflicting with one another.

Is there a way to avoid this conflict?

aselle:
@jhseu, do you have any insight on this. I think @pwaller's analysis is correct. There is clearly some singleton that is getting double initialized.

jhseu:
Yeah, I think this happens because we import pywrap_tensorflow with RTLD_GLOBAL. Unfortunately I don't think there's an easy way to make this work at the moment. It's possible we can go the other way and make the C API accessible from `import tensorflow`.

Keeping this open, but it's uncommon and difficult, so we may not work on it anytime soon. A possible workaround is to mangle the symbols (namespace?) in the TensorFlow shared library.

[...]

PR #10469, opened 2017-06-06
Export C API symbols in _pywrap_tensorflow_internal.so

fritzo:
This PR exports C API symbols so that python extension libraries can use the C API.

As noted in #7541, there is currently a conflict between `_pywrap_tensorflow_internal.so` and `libtensorflow.so`. This conflict prevents use of the tensorflow C API and the python API in the same process. This PR attempts to implement a workaround suggested by @jhseu by exporting the C API symbols in the python library, so that the Python and C APIs are both available from the single `_pywrap_tensorflow_internal.so`.

### DIFF
--- a/tensorflow/tf_exported_symbols.lds
+++ b/tensorflow/tf_exported_symbols.lds
@@ -1,3 +1,4 @@
 *tensorflow*
 *perftools*gputools*
 *tf_*
+TF_*
--- a/tensorflow/tf_version_script.lds
+++ b/tensorflow/tf_version_script.lds
@@ -2,6 +2,7 @@ tensorflow {
   global:
     *tensorflow*;
     *perftools*gputools*;
+    TF_*;
   local:
     *;
 };


tensorflow/tf_exported_symbols.lds
```diff
@@ -1,3 +1,4 @@
 *tensorflow*
 *perftools*gputools*
 *tf_*
+TF_*
```
jhseu: This shouldn't be necessary due to the `*tf_*` above.

tensorflow/tf_version_script.lds
```diff
@@ -2,6 +2,7 @@ tensorflow {
   global:
     *tensorflow*;
     *perftools*gputools*;
+    TF_*;
```
jhseu: *tf_* instead?

fritzo: This appears to be case sensitive, and the C API functions are all prefixed with `TF_`.

fritzo: Actually, I don't know whether it's case sensitive. I'll look around for documentation.

fritzo: I confirmed that `--exclude_symbols_list` is case sensitive by replacing `TF_` with `*tf_*` and observing that the `TF_New*` symbols were present with `TF_` but absent with `*tf_*`.

tensorflow/tf_exported_symbols.lds
fritzo: `--exclude_symbols_list` is case sensitive; see comment below.

jhseu:
Seems fine to me. Adding Asim to see if he has any objections.

asimshankar (approved):
Looks good to me, thanks!

jhseu:
Jenkins, test this please

fritzo:
Thanks for reviewing, @jhseu!

PR #10470, opened 2017-06-06
Fix test failures on windows.

gunan:


av8ramit (approved):

meteorcloudy:
@gunan Thanks for fixing these tests!
As for `tensorflow/python/kernel_tests:reader_ops_test`, it failed because having problem [reading data file](https://github.com/tensorflow/tensorflow/commit/e6f58186363279496c46563e6f065ce7ea16c501#diff-5f4bbe7d0a58188f2ab3766df90062aaR888). This is a bug in Bazel, I've filed https://github.com/bazelbuild/bazel/issues/3134. I think we'll have to disable this test temporarily on Windows until the bug is fixed in a new release.

PR #10490, opened 2017-06-07
Recognize CPU core count in FreeBSD

blodan:
At the moment FreeBSD can't recognize the CPU count, however FreeBSD uses the same method as OS X does so provided PR fixes that.

If this could be included in the 1.2.0 release that would be great.


jhseu (approved):
It'll be too late to make it into 1.2, though.

jhseu:
Jenkins, test this please

gunan:
Makefile failure unrelated. Merging

PR #10502, opened 2017-06-07
A few changes to kernel_tests.

gunan:
@meteorcloudy FYI


gunan:
Jenkins, test this please.

yifeif (approved):

jhseu:
stage_op_test is still timing out too. Set that to large?

gunan:
I think it just gets stuck.
I can set it to large, but I think we will need to disable it.

PR #10511, opened 2017-06-08
Printdhruv deep cnn.md

printdhruv:
Typo


jhseu:
Thanks for the pull request! Both versions are correct, but since the change is unnecessary, closing.

Issue #9360, opened 2017-04-21
Understanding cast() in tensorflow


sandipmgiri:
### System Information

- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Ubuntu 16.04 (ppc64le)
- **TensorFlow installed from (source or binary):**
Installed from source (v1.0.1)
- **TensorFlow version (use command below):**
('v1.0.1-0-ge895d5c-dirty', '1.0.1')
- **Bazel version (if compiling from source)**:
 bazel release 0.4.4-2017-04-10 (@80a07b5)
- **CUDA/cuDNN version:**
In disable mode
- **Exact command to reproduce**:
bazel test //tensorflow/python/kernel_tests:cast_op_test

### Describe the problem clearly 
This is regarding failure of test case testInfNan in cast_op_test.py file.While executing this test case on ppc64le, it was observed that following line returns unexpected results:

`self._compare(np.inf, np.int32, i4.min, False)`

i4.min value on x86 as well as on ppc64le is -2147483648. However "np.inf" on x86 is "signed" by default whereas on ppc64le it is "unsigned" by default. To make the results compatible with x86, somehow np.inf should be cast as "signed" on ppc64le. In my opinion there could be two ways of doing this.

1. Use a proper cast (python equivalent of "(signed int) var" in C) which    would always interprete np.inf as of type "signed" on ppc64le
-- or --
2. if we are on ppc64le platform, when dealing with np.inf, convert it explicitly to signed as "-np.inf" and then perform subsequent operations

Though I have not yet decided on which one to implement, I am trying to find a right place first to put this fix in tensorflow code. I guess the right place would be somewhere in the code related to following 2 lines in cast_op_test.py (line# 57 and 58, in function _cast):

```
val = constant_op.constant(x, self._toDataType(np.array([x]).dtype))
return math_ops.cast(val, self._toDataType(dtype), name="cast").eval()

```
However I am unable to grasp code details about constant() in python/framework/constant_op.py and cast() in python/ops/math_ops.py, similarly there is REGISTER_OP("Cast") in core/ops/math_ops.cc which I guess is the heart of cast functionality. Is my understanding correct?

So if I have to implement the changes for ppc64le, which could be the right place to do so?

### Source Code / Logs
```
`$ bazel test //tensorflow/python/kernel_tests:cast_op_test


exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
[...]

PR #10522, opened 2017-06-08
np.float64("np.inf").astype(np.int32) is negative on x86 but positive on ppc64le

sandipmgiri:
- np.float64("np.inf").astype(np.int32) is negative on x86 but positive on ppc64le, that is the reason we 
  added this special case for ppc64le

- Discussed with numpy community and they mentioned as follows -
1) Doing this cast incurs undefined behaviour, so it's hard to say which platform is incorrect here.
    I'd argue that we should just issue a warning and leave the platform-specific behaviour.
2) This is also true of any C program that tries to cast float to int. If the author of a python package 
    included a C extension that does this cast, they still have the problem. 
   The same argument you use for "we should fix this is numpy, not in the user package" can be 
    applied to "we should fix this in C, not in numpy". Presumably for speed reasons, that argument was 
    rejected during the  design of C.
    Having said that, we already diverge from C behaviour in some place (eg, integer promotion for 
    arithmetic and comparison), so it wouldn't be unreasonable  to add special casing here.
    
Numpy link to relevant discussion - https://github.com/numpy/numpy/issues/9040
Also discussed this on Tensorflow , link to relevant discussion - https://github.com/tensorflow/tensorflow/issues/9360

### DIFF
--- a/tensorflow/core/platform/profile_utils/cpu_utils.cc
+++ b/tensorflow/core/platform/profile_utils/cpu_utils.cc
@@ -28,8 +28,8 @@ namespace profile_utils {
 
 static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
 
-/* static */ int64 CpuUtils::GetCycleCounterFrequency() {
-  static const int64 cpu_frequency = GetCycleCounterFrequencyImpl();
+/* static */ uint64 CpuUtils::GetCycleCounterFrequency() {
+  static const uint64 cpu_frequency = GetCycleCounterFrequencyImpl();
   return cpu_frequency;
 }
 
@@ -53,7 +53,7 @@ static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
                                        GetCycleCounterFrequency());
 }
 
-/* static */ int64 CpuUtils::GetCycleCounterFrequencyImpl() {
+/* static */ uint64 CpuUtils::GetCycleCounterFrequencyImpl() {
 // TODO(satok): do not switch by macro here
 #if defined(__ANDROID__)
   return GetCpuUtilsHelperSingletonInstance().CalculateCpuFrequency();
--- a/tensorflow/core/platform/profile_utils/cpu_utils.cc
+++ b/tensorflow/core/platform/profile_utils/cpu_utils.cc
@@ -60,7 +60,7 @@ static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
                                        GetCycleCounterFrequency());
 }
 
-/* static */ uint64 CpuUtils::GetCycleCounterFrequencyImpl() {
+/* static */ int64 CpuUtils::GetCycleCounterFrequencyImpl() {
 // TODO(satok): do not switch by macro here
 #if defined(__ANDROID__)
   return GetCpuUtilsHelperSingletonInstance().CalculateCpuFrequency();
--- a/tensorflow/core/platform/profile_utils/cpu_utils.cc
+++ b/tensorflow/core/platform/profile_utils/cpu_utils.cc
@@ -28,7 +28,7 @@ namespace profile_utils {
 
 static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
 
-#if defined(__powerpc__) || defined(__ppc__)
+#if defined(__powerpc__) || defined(__ppc__) && ( __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
    /* static */ uint64 CpuUtils::GetCycleCounterFrequency() {
      static const uint64 cpu_frequency = GetCycleCounterFrequencyImpl();
      return cpu_frequency;
--- a/tensorflow/core/platform/profile_utils/cpu_utils.cc
+++ b/tensorflow/core/platform/profile_utils/cpu_utils.cc
@@ -28,10 +28,17 @@ namespace profile_utils {
 
 static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
 
-/* static */ uint64 CpuUtils::GetCycleCounterFrequency() {
-  static const uint64 cpu_frequency = GetCycleCounterFrequencyImpl();
-  return cpu_frequency;
+#if defined(__powerpc__) || defined(__ppc__)
+   /* static */ uint64 CpuUtils::GetCycleCounterFrequency() {
+     static const uint64 cpu_frequency = GetCycleCounterFrequencyImpl();
+     return cpu_frequency;
 }
+#else
+   /* static */ int64 CpuUtils::GetCycleCounterFrequency() {
+     static const int64 cpu_frequency = GetCycleCounterFrequencyImpl();
+     return cpu_frequency;
+}
+#endif
 
 /* static */ double CpuUtils::GetMicroSecPerClock() {
   static const double micro_sec_per_clock =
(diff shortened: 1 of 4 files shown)


sandipmgiri:
Hi  @martinwicke, as discussed on thread https://github.com/tensorflow/tensorflow/issues/9360, I have created this PR.

sandipmgiri:
For second (https://github.com/tensorflow/tensorflow/pull/10522/commits/c63cebb80a26cb7eeaf0afe095886488f1a8ac8a) commit details are as follows :

This is regarding failure of test case //tensorflow/core:platform_profile_utils_cpu_utils_test in tensorflow/core/platform/profile_utils/cpu_utils_test.cc file.
On ppc64le this test was failing for following two checks i.e. CHECK_GT(cpu_frequency, 0) and CHECK_NE(cpu_frequency, CpuUtils::INVALID_FREQUENCY) :

https://github.com/tensorflow/tensorflow/blob/v1.0.1/tensorflow/core/platform/profile_utils/cpu_utils_test.cc#L56-L57

```
TEST_F(CpuUtilsTest, CheckCycleCounterFrequency) {
  const int64 cpu_frequency = CpuUtils::GetCycleCounterFrequency();
  CHECK_GT(cpu_frequency, 0);
  CHECK_NE(cpu_frequency, CpuUtils::INVALID_FREQUENCY);
  if (DBG) {
    LOG(INFO) << "Cpu frequency = " << cpu_frequency;
  }
}
```

After invetigation found that - "cpu_frequency" value on x86 is "usigned" by default whereas on ppc64le it is "signed" by default, that is why this test was failing on ppc64le, PFA for log details 
[platform_profile_utils_cpu_utils_test failure_log.txt](https://github.com/tensorflow/tensorflow/files/1071075/platform_profile_utils_cpu_utils_test.failure_log.txt)


To make the results compatible with x86, some code changes were required on ppc64le. 
After this code changes, I ran this test and it is passing on ppc64le as well as on X86.

Please review this changes as well?
Thanks!

martinwicke:
@prb12 friendly nudge?

tensorflow/core/platform/profile_utils/cpu_utils.cc
```diff
@@ -53,7 +53,7 @@ static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
                                        GetCycleCounterFrequency());
 }
 
-/* static */ int64 CpuUtils::GetCycleCounterFrequencyImpl() {
+/* static */ uint64 CpuUtils::GetCycleCounterFrequencyImpl() {
```
prb12: What is the reason for this change?

Several of the return paths for the function return negative numbers (INVALID_FREQUENCY), some now with implicit casts to unsigned.  

The only use in the rest of this patch adds the opposide cast of INVALID_FREQUENCY back to unsigned.  

What am I missing here?

tensorflow/python/kernel_tests/cast_op_test.py
```diff
@@ -146,9 +147,16 @@ def testInfNan(self):
     if sys.byteorder == "big":  
       self._compare(np.inf, np.int32, i4.max, False)  
       self._compare(np.inf, np.int64, i8.max, False)  
-    else:  
-      self._compare(np.inf, np.int32, i4.min, False)  
-      self._compare(np.inf, np.int64, i8.min, False)  
+    else:
+      #np.float64("np.inf").astype(np.int32) is negative on x86 but positive on ppc64le
```
prb12: Nit: Comments should have a space after the #

tensorflow/python/kernel_tests/cast_op_test.py
```diff
@@ -146,9 +147,16 @@ def testInfNan(self):
...
+      #Numpy link to relevant discussion - https://github.com/numpy/numpy/issues/9040
+      #Tensorflow link to relevant discussion - https://github.com/tensorflow/tensorflow/issues/9360
+      if platform.machine() == "ppc64le":
+        self._compare(-np.inf, np.int32, i4.min, False)
```
prb12: As I read the posts linked to above, -np.inf is technically undefined?  
I think this is a little ugly, but fine as a temporary fix, and it doesn't affect other architectures.

prb12 (commented):
Could you please provide some justification for the change to GetCycleCounterFrequency?

sandipmgiri:
**Details for the change to GetCycleCounterFrequency :** 

This is regarding test case` //tensorflow/core:platform_profile_utils_cpu_utils_test`. While running this test on `ppc64le` platform `GetCycleCounterFrequency` function returning  unexpected  value (i.e. `-1` vs expected `greater than 1`)  ,which is ` "int64 cpu_frequency"` value
As per analysis here we found that  :  "`int64`"  value on `x86` is "unsigned" by default whereas on `ppc64le` it is "signed" by default. Because of that we were getting incorrect value on power.

To make the result compatible with `x86`, I have done this changes on `ppc64le`.
After code changes, I ran this test and now it is passing on` ppc64le` as well as on `X86`.

In first commit (https://github.com/tensorflow/tensorflow/commit/c63cebb80a26cb7eeaf0afe095886488f1a8ac8a) I had made generic changes . After that I realized that this test is passing successfully on` X86` without any change. Hence, now I have made power specific changes only in latest commit (i.e.  https://github.com/tensorflow/tensorflow/pull/10522/commits/c42d2f52a8237e7e73e29a5658bfe0c83baee414)

Please provide your comments on this and  let me know if you need any other details. Thanks!

tensorflow/core/platform/profile_utils/cpu_utils.cc
drpngx: @sandipmgiri could you address this?

tensorflow/python/kernel_tests/cast_op_test.py
drpngx: @sandipmgiri could you fix the comment style?

sandipmgiri:
@drpngx fixed comment style and made required changes. Please review and let me know if you need any additional details. Thanks!

drpngx:
Jenkins, test this please.

drpngx:
Thanks!

@prb12 good to go?

prb12 (approved):

eric-wieser:
> "int64" value on x86 is "unsigned" by default

What exactly do you mean by this? `int64_t` is defined as a _"signed integer type with width of exactly 8, 16, 32 and 64 bits respectively"_. So I still have no understanding of why that change to `GetCycleCounterFrequency` is necessary

tensorflow/core/platform/profile_utils/cpu_utils_test.cc
```diff
@@ -53,9 +53,15 @@ TEST_F(CpuUtilsTest, CheckGetCurrentClockCycle) {
 }
 
 TEST_F(CpuUtilsTest, CheckCycleCounterFrequency) {
-  const int64 cpu_frequency = CpuUtils::GetCycleCounterFrequency();
-  CHECK_GT(cpu_frequency, 0);
-  CHECK_NE(cpu_frequency, CpuUtils::INVALID_FREQUENCY);
+  #if defined(__powerpc__) || defined(__ppc__) && ( __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
+     const uint64 cpu_frequency = CpuUtils::GetCycleCounterFrequency();
+     CHECK_GT(cpu_frequency, 0);
+     CHECK_NE(cpu_frequency, unsigned(CpuUtils::INVALID_FREQUENCY));
```
eric-wieser: Why are you casting to `unsigned`, which means `unsigned int`? Shouldn't you be casting to `int64`?

PR #10525, opened 2017-06-08
Try both python and python3

Androbin:
Used `configure` script for reference


aselle:
Did you mean to close this?

Androbin:
@aselle, This was intended be to taken together with another PR, but I seam to have missed this one.

PR #10528, opened 2017-06-08
Make stage tests less sensitive to timeouts

sjperkins:
Previously, staging tests used the number of failed token dequeues (with 50 ms timeouts) on
a signalling queue to indicate blocked puts. This was probably too sensitive on highly contended testing systems. Basically, there's no good way to distinguish between a put that took too long and a blocking put.

Staging tests now simply count the number of iterations before a timeout of 1s occurs on the signalling queue. This should mean that:

1. The test case is less brittle around what is considered a block vs  the putting thread not being able to keep up.
2. Both stage_op_test and map_stage_op_test take slightly longer, ~2.5s vs ~1s, to run on an uncontended system..

Also simplified the tests cases in general.

/cc @ekelsen @jhseu I saw #10516 and the duration of test runs (~30s on your test systems vs ~1s on my laptop) made me reconsider the duration on the blocking timeout. I've upped it 20X from 50ms to 1s.

In an ideal world, the test cases wouldn't depend on a timeout when running -- if you can suggest a different method of testing blocking puts and gets, please let me know.

I also relooked at the multi-threaded code in the StagingArea, but I think the standard producer/consumer queue has been implemented correctly.

I hope this helps.

### DIFF
--- a/tensorflow/python/kernel_tests/BUILD
+++ b/tensorflow/python/kernel_tests/BUILD
@@ -2382,11 +2382,6 @@ cuda_py_test(
         "//tensorflow/python:util",
         "//tensorflow/python:data_flow_ops",
     ],
-    tags = [
-        "manual",  # http://b/62429636
-        "noguitar",
-        "notap",
-    ],
 )
 
 cuda_py_test(
@@ -2400,11 +2395,6 @@ cuda_py_test(
         "//tensorflow/python:math_ops",
         "//tensorflow/python:util",
         "//tensorflow/python:data_flow_ops",
-    ],
-    tags = [
-        "manual",  # http://b/62429636
-        "noguitar",
-        "notap",
     ],
 )
 
--- a/tensorflow/python/kernel_tests/map_stage_op_test.py
+++ b/tensorflow/python/kernel_tests/map_stage_op_test.py
@@ -24,6 +24,7 @@ from tensorflow.python.ops import data_flow_ops
 from tensorflow.python.ops import math_ops
 from tensorflow.python.platform import test
 
+TIMEOUT = 1
 
 class MapStageTest(test.TestCase):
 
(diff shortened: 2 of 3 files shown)


ekelsen:
Jenkins, test this please.

ekelsen:
Jenkins, test this please

ekelsen:
Simon - thanks for looking into this.  I know it's annoying to deal with test failures that only happen on the test machines and not locally.

We had disabled the tests, so to be able to test them on the test server you'll need to re-enable them.

sjperkins:
> Simon - thanks for looking into this. I know it's annoying to deal with test failures that only happen on the test machines and not locally.

Likewise, thanks for your patience with this. I've re-enabled the tests.

gunan:
Jenkins, test this please.

sjperkins:
Test failures look unrelated. Note that stage_op_test.py never managed to run on the `Linux CPU Test`, while both map_stage_op_test.py and stage_op_test.py completed successfully on the `Windows Cmake Tests`

ekelsen:
I agree, this looks good.  It would be nice to get that api compatibility test fixed so we can get a run on linux.

ekelsen (approved):

sjperkins:
I misread the logs -- `stage_op_test.py` passed successfully in both of the failed test suites.

martinwicke:
Could you fix the conflicts?

martinwicke:
@jhseu does this change address http://b/62429636? If so, we can re-enable the tests (they don't run right now, so this is an untested change, and I don't want to merge it like this).

sjperkins:
Removed all tags in `stage_op_test.py` and `map_stage_op_test.py`. The `noguitar` and `notap` tags aren't documented so I'm not sure what their removal implies...

jhseu:
@martinwicke I'm not certain, but it seems likely to fix that bug.

jhseu:
Jenkins, test this please

sjperkins:
Staging Tests look good. The Windows Cmake Tests failed on an unrelated test case.

gunan:
Confirming that windows cmake issues are unrelated. Should be fixed at master in the last hour.

PR #10531, opened 2017-06-08
Improve RDMA rendezvous speed

### LINKED ISSUE #10530: Protobuf serialization slows down RDMA
According to a test similar to https://github.com/tensorflow/tensorflow/issues/6116, the current RDMA implementation has performance issues.

This tensor copy throughput according to [this test](https://gist.github.com/yaroslavvb/1124bb02a9fd4abce3d86caf2f950cb2) (change `assign_add` to `assign` in L53, and test with tensor size 100MB):
```
Distributed rate: 536.72 MB per second
```
Which is just slightly higher than the gRPC implementation (the gRPC fixed version: https://github.com/tensorflow/tensorflow/pull/7466)

Here is the profiling result:
[device1 worker profiling report](https://github.com/tensorflow/tensorflow/files/1060989/dev1.pdf)
[device2 worker profiling report](https://github.com/tensorflow/tensorflow/files/1060990/dev2.pdf)

There are some obvious issue, for example, the [string resize issue](https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/stl_util.h#L67), the unnecessary memory copy and the single threaded serialization/deserialization.

llhe:
This PR fixes https://github.com/tensorflow/tensorflow/issues/10530.

Tested with RoCE setting: Mellanox Technologies MT27620 Family

Before this PR:
```
Distributed rate: 536.72 MB per second
```
After this PR:
```
Distributed rate: 1346.06 MB per second
```

TODO: memcpy is still slow and need to improved.

### DIFF
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -21,6 +21,7 @@ limitations under the License.
 #include "tensorflow/core/common_runtime/device_mgr.h"
 #include "tensorflow/core/common_runtime/dma_helper.h"
 #include "tensorflow/core/common_runtime/gpu/gpu_util.h"
+#include "tensorflow/core/common_runtime/gpu/process_state.h"
 #include "tensorflow/core/distributed_runtime/rendezvous_mgr_interface.h"
 #include "tensorflow/core/distributed_runtime/session_mgr.h"
 #include "tensorflow/core/framework/rendezvous.h"
@@ -683,7 +684,6 @@ void RdmaTensorBuffer::SendNextItem() {
                          << " error message: " << status.error_message();
       size_t buffer_size = RdmaMessage::kMessageTotalBytes;
       size_t tensor_bytes = 0;
-      TensorProto proto;
       // Figures out which device the tensor is hosted on.
       Device* src_dev = nullptr;
       Status s = channel_->adapter_->worker_env_->device_mgr->LookupDevice(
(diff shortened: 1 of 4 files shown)


byronyi:
Seems a reasonable patch. 

By the way, how do you measure the rates? The MT27620 ConnectX-4 NIC should support 100 GbE fabrics. These rates seem a little bit too small compared to the hardware capabilities.

llhe:
It's not pure bandwidth test but a tensor copy test like this: https://gist.github.com/yaroslavvb/1124bb02a9fd4abce3d86caf2f950cb2. (change `assign_add` to `assign` in L53).

byronyi:
Thanks for pointing me to the script. I am actually wondering if the loopback RDMA performance could reflect the actual in-fabric performance. Honestly I am not sure if hardware is involved at all when using loopback; it might use a completely different mechanism than NIC doing DMA.

Nevertheless this patch will still help in the case of local worker-ps communication.

llhe:
@junshi15 @anfeng @shamoya  Any comments on this?

2sin18:
It doesn't converge in GPU device if I patch this PR in tensorflow master branch. It's OK in CPU device, so I guess there're some bugs in `SetProtoFromGPUSync` related code.

byronyi:
@2sin18 Note that the original implementation uses ``SetProtoFromGPUSync`` even before this patch. If there is anything susceptible it should be ``CopyGPUTensorToCPUSync`` and ``CopyCPUTensorToGPUSync `` which is introduced in this patch.

tensorflow/contrib/verbs/rdma_rendezvous_mgr.cc
```diff
@@ -99,12 +100,41 @@ void RdmaRemoteRendezvous::RecvFromRemoteAsync(
...
+          Allocator* alloc = ProcessState::singleton()->GetCUDAHostAllocator(0);
+          Tensor copy(alloc, rm.data_type_, rm.tensor_shape_);
+          memcpy(DMAHelper::base(&copy), input, rm.tensor_bytes_);
+
+          Allocator* dst_alloc = dst_dev->GetAllocator(host_alloc_attrs);
+          Tensor gpu_copy(dst_alloc, rm.data_type_, rm.tensor_shape_);
+          s = VerbsUtil::CopyCPUTensorToGPUSync(&copy, recv_args.device_context,
+                                                dst_dev, &gpu_copy);
+          CHECK(s.ok()) << "copy tensor to gpu sync";
+          val = std::move(gpu_copy);
```
junshi15: Does std::move work properly for GPUTensor?

junshi15: The chain of copying can be further simplified. Currently, you copy twice here (assuming move works as expected):
input --> (cpu) copy --> gpu_copy -->(move)--> val,
If you want to be aggressive, you can do input-->gnu_copy directly, skipping the cpu_copy. In one of our early versions (not released), we had that. But it required low-level functions similar to cudaMemcpy. It was too complex, so in the final public version, we chose the protobuf route for simplicity, but sacrificed performance. Ditto on the sender side (rdma.cc).

junshi15: @llhe Thanks for your contribution. The patch looks good except that I am not sure whether std::move works for GPU tensors as expected. Besides speed, can you verify the result is correct? @2sin18 had issues about the convergence while running with GPU. I suspect it has something to do either "copy" or "move" operations.

llhe:
@2sin18 Did you tested with RDMA? Can you provide a repro about the convergence issue? Thanks.

tensorflow/contrib/verbs/rdma_rendezvous_mgr.cc
llhe: std::move should be ok for GPU tensor, see the code [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.h#L692). Actually the [standard assignment](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.h#L161) also just an pointer assignment.

tensorflow/contrib/verbs/rdma_rendezvous_mgr.cc
```diff
@@ -99,12 +100,41 @@ void RdmaRemoteRendezvous::RecvFromRemoteAsync(
...
+        if (dst_dev->tensorflow_gpu_device_info() &&
+            (!recv_args.alloc_attrs.on_host())) {
+          CHECK(recv_args.device_context)
+            << "send dev name: " << src_dev->name()
+            << " gpu_info: " << src_dev->tensorflow_gpu_device_info();
+          Allocator* alloc = ProcessState::singleton()->GetCUDAHostAllocator(0);
+          Tensor copy(alloc, rm.data_type_, rm.tensor_shape_);
+          memcpy(DMAHelper::base(&copy), input, rm.tensor_bytes_);
+
+          Allocator* dst_alloc = dst_dev->GetAllocator(host_alloc_attrs);
```
llhe: @2sin18 @junshi15 Thanks for pointing this out. This should be changed to `recv_args.alloc_attrs`?

junshi15: yes, I think "recv_args.alloc_attrs" is the right attributes.

junshi15:
@2sin18 Does @llhe 's new patch fix the convergence problem?

2sin18:
@junshi15 Thanks to @llhe 's patch, it works now.

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -703,21 +703,47 @@ void RdmaTensorBuffer::SendNextItem() {
...
-        // "val" is on a GPU. Uses GPUUtil to fill the proto.
-        s = VerbsUtil::SetProtoFromGPUSync(
-            in, src_dev, send_args.device_context, &proto, is_dead);
-        CHECK(s.ok()) << "set proto from gpu sync";
+          << "send dev name: " << src_dev->name()
+          << " gpu_info: " << src_dev->tensorflow_gpu_device_info();
+
+        if (can_memcpy) {
+          AllocatorAttributes host_alloc_attrs;
+          host_alloc_attrs.set_gpu_compatible(true);
```
poxvoculi: Add
host_alloc_attrs.set_nic_compatible(true);
This indicates that the memory should be allocated from a pool registered for DMA by the NIC.   Not effective yet, but probably soon.

poxvoculi:
Jenkins test this please

poxvoculi (approved):

PR #10544, opened 2017-06-08
[Bash] Use grep -q instead of [ -n .. ]

Androbin:
As proposed by static analysis tool:
https://github.com/koalaman/shellcheck/wiki/SC2143


jhseu:
Closing these all out. Mind making all these changes in one pull request? It's easier to run through tests that way.

PR #10574, opened 2017-06-08
Changed default `evalution_master` to also use cluster information

jhaux:
When using the standard Estimator during evaluation in distributed setting (a `cluster_spec` was supplied) I encountered the following error message:

```
InvalidArgumentError (see above for traceback): Cannot assign a device for operation 
'save/RestoreV2_28': Operation was explicitly assigned to 
/job:ps/replica:0/task:0/device:CPU:0 but available devices are
[ /job:localhost/replica:0/task:0/cpu:0, /job:localhost/replica:0/task:0/gpu:0 ]. 
Make sure the device specification refers to a valid device.
```
```
[[Node: save/RestoreV2_28 = RestoreV2[dtypes=[DT_FLOAT], 
_device="/job:ps/replica:0/task:0/device:CPU:0"](save/Const, save/RestoreV2_28/tensor_names, 
save/ResyoreV2_28/shape_and_slices)]]
```
In short:
The evaluation graph is looking for Variables on ` /job:localhost` but they are on ` /job:ps`, as is intended. 

To fix this I changed the default settings of placing the `evaluation_master` on `localhost` to the same as for `master`. This means that when instancing a RunConfig with no `evaluation_master` given it will check if there is a `cluster_spec`. If so it will get the `evaluation_master` based on the `cluster_spec`. 

This fixes the problem for me and leads to correct behaviour.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/estimators/run_config.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/run_config.py
@@ -152,7 +152,9 @@ class ClusterConfig(object):
       self._is_chief = (self._task_type == TaskType.WORKER and
                         self._task_id == 0)
 
-    self._evaluation_master = evaluation_master or ''
+    self._evaluation_master = (evaluation_master if evaluation_master is not None else
+                    _get_master(self._cluster_spec, self._task_type,
+                                self._task_id) or '')
 
   @property
   def cluster_spec(self):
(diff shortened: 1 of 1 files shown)


xiejw:
Sorry, I cannot approve this pull request. 

By design, the evaluation will be executed based on the model parameters from checkpoint. This ensures that the whole evaluation, which could be multiple steps, has a consistent snapshot/view of the model parameters and will not change (and also avoid the race condition) if there is a training happening in parallel (so, ps gets updated). 

This is why evaluation_master is '' by default. And in most of the use cases, this is a only correct behavior. If there is any cutting edge research exploring new evaluation approach or user strongly wants the customized evaluation_master to be the same as master (though I am not fully convinced this is correct), user can set the evaluation_master explicitly. The default setting should not be changed.

drpngx:
Jenkins, test this please.

jhaux:
All right, thanks! I did not think about it this way.

PR #10591, opened 2017-06-09
Fix incorrect documentation in _SliceHelper #10494

### LINKED ISSUE #10494: incorrect documentation in _SliceHelper
The outputs shown for several of the examples in the `_SliceHelper` docstring are incorrect (based on some previous examples?).

See, for example, this line: https://github.com/tensorflow/tensorflow/blob/359d6f9716c0bb9bd8201ce600da98b0481a8049/tensorflow/python/ops/array_ops.py#L412

Androbin:
Fixes #10494.

### DIFF
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -407,9 +407,10 @@ def _SliceHelper(tensor, slice_spec, var=None):
   print(foo[:, :, tf.newaxis].eval()) # => [[[1],[2],[3]], [[4],[5],[6]], [[7],[8],[9]]]
 
   # Ellipses (3 equivalent operations)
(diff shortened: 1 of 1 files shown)


martinwicke:
Jenkins, test this please.

Androbin:
@martinwicke Why do documentation updates even need a full CI test?

martinwicke:
They don't, a partial one would do, but we haven't set it up with two separate builds. We can do that, it's just that more interesting things keep happening. But the documentation is built as well, and doc changes can break for a variety of reasons, some of them subtle. 

That said, I do at times merge without testing. Don't tell anyone.

aselle (approved):
Thanks for the fix!

PR #10595, opened 2017-06-09
Remove deprecated functions

Androbin:
Remove deprecated classes and functions marked for removal until May 2017.
This PR doesn't touch deprecated parameters nor deprecations for June 15.


martinwicke:
Thanks! I really want to accept this, but I cannot (it would save a lot of headache).

I should never have put dates on these, and probably should remove them. We'd be breaking too many people and we won't actually remove these functions, probably until we reorganize contrib as a whole. 

I'm sorry about this. I should remove the dates, since they're misleading.

PR #10598, opened 2017-06-09
Implements inverse hyperbolic operations (fix for #7531)

### LINKED ISSUE #7531: Hyperbolic functions
I'd like to request support for all hyperbolic ops and their inverse:
sinh, cosh, asinh, acosh, atanh

Especially asinh is interesting, as it can be used as an activation function.
This would also allow to address problems with their calculation centrally, e.g.:
`asinh(x) = log(x + sqrt(x² + 1))` will yield -∞/NaN (and break gradients) for small `x`. This is caused by the fact that the `log` argument `x + sqrt(x² + 1)` will get rounded to 0 due to machine precision.

lakshayg:
fixes #7531.


caisq:

lakshayg:
I ran the tests which failed here on my machine and found that the only cases which were failing are the ones in which `tf.acosh` evaluates to `inf/-inf` and `np.arccosh` evaluates to `nan/-nan`. Should I change the test cases or do something else? Any ideas on how to proceed?

rmlarsen (approved):

rmlarsen:

lakshayg:
The failures in tests/build seem unrelated to the changes made by this pull request.

martinwicke:
Jenkins, test this please.

martinwicke:
The cwise ops test failure (only on Windows, ugh) is probably caused by this CL -- the tolerances appear to be too tight.

lakshayg:
Any hints on how to proceed?

martinwicke:
I would look at which test fails, exactly, and adjust the tolerance for that test in the test code. I think the values are still close so this is not a real problem, just numerics and a particularly unfriendly function.

lakshayg:
If that is the case then this PR would have to wait for a few days since I do not have access to my machine and would be able to fix it only after the 16th.

martinwicke:
No problem.

martinwicke:
This test failure is not due to this PR. We'll review API changes on Monday, should be able to merge then.

lakshayg:
Great, thanks!

lakshayg:
Since the tests for `acosh` and `asinh` were passing, I have also added a commit which implements `atanh` thus completely resolving #7531. Please review.

rmlarsen (approved):

rmlarsen:

lakshayg:
@rmlarsen `//tensorflow/cc:gradients_math_grad_test` is failing in Linux CPU Tests and MacOS CPU Tests but passes in Linux CPU Tests (Python 3) and Linux GPU. Are these failures due to some bug in the code?

lakshayg:
I ran the test locally with python 2 and found that the test was failing due to some case for which numpy gave `nan` and tf gave `inf`. I have changed the test cases so that this does not happen. The test passes on my machine with python2 now.

rmlarsen:
@lakshayg is inf the mathematically correct answer? We should strive for compatibility with numpy whenever possible, even for arguments that give nan/in, unless it is clearly wrongf. Does this depend on the version of numpy? I'm OK with the PR as is, but perhaps add a TODO in the test code to investigate this edge case?

lakshayg:
@rmlarsen Oh, I just checked and found that numpy gave inf and tf gave nan. WolframAlpha says that the value should be `inf` (http://www.wolframalpha.com/input/?i=atanh(1%2B0i)). Any ideas on how to fix this?

rmlarsen:

rmlarsen:
@lakshayg could it be that in the gradient computation, TF ends up computing inf/inf or 0*inf, while you compute the numpy result more directly in the test?

lakshayg:
The gradient for atanh(x) is 1 / (1 - x^2). Could that be causing the problem for x = 1?

rmlarsen:
@lakshayg that should give 1/0 = inf. I'm not sure why it produces NaN instead of inf, but that is unrelated to your change.

lakshayg:
@rmlarsen I just compiled the binary and checked what it gave for x=1 and the answer is `inf`. I guess that everything is working fine and this PR can be merged.

screenshot from 2017-06-21 08-49-33

rmlarsen:
@martinwicke This is ready to be merged. You added the API review tag. What needs to be done?

rmlarsen:
@lakshayg could you please rebase to resolve conflicts?

rmlarsen:
@martinwicke I missed you earlier comment about API review last monday. I'll assign you on this PR and you can merge when it's approved.

lakshayg:
@rmlarsen rebase done!

rmlarsen:
@lakshayg Thanks!

jhseu:
How about renaming to arcsinh, arccosh, arctanh to match the numpy names?

lakshayg:
@jhseu I guess then we would have to rename asin, acos and atan as well.

jhseu:
Yes, but perhaps wait for API review for the final word.

rmlarsen:
@jhseu We already have atan, atan2, asin etc. so it doesn't make sense unless we change the existing interface.

drpngx:
@jhseu or @martinwicke : to confirm, good to go?

martinwicke:
Good for API review, but watch out, this will require Eigen changes to pull in.

gunan:
I see failures in windows CPU and GPU build due to additions from this change:
http://ci.tensorflow.org/job/tensorflow-master-win-cmake-py/1101/console
```
cwise_ops_test.py", line 411, in testComplex64Basic
cwise_ops_test.py", line 302, in testDoubleBasic
cwise_ops_test.py", line 208, in testFloatBasic
```

Looks like there is some flakiness in arccosh implementation on windows?

drpngx:
@rmlarsen is out next week.

It was green when I merged. Looking at the code, there doesn't seem to be anything obviously special about cosh.

@lakshayg could you take a look?

tensorflow/core/kernels/cwise_op_acosh.cc
```diff
@@ -0,0 +1,38 @@
...
+#define REGISTER_SYCL_KERNEL(TYPE)                                    \
+  REGISTER_KERNEL_BUILDER(                                            \
+                          Name("Acosh")                               \
+                          .Device(DEVICE_SYCL)                        \
+                          .TypeConstraint<TYPE>("T"),                 \
+                          UnaryOp<SYCLDevice, functor::acosh<TYPE>>);
+REGISTER_SYCL_KERNEL(float);
+REGISTER_SYCL_KERNEL(double);
+#undef REGISTER_SYCL_KERNEL
+#endif // TENSORFLOW_USE_SYC
```
drpngx: We should fix this typo in another PR.

lakshayg: Sent in a PR for the typo SYC -> SYCL  #11204

lakshayg:
@gunan @drpngx: @martinwicke earlier suggested (https://github.com/tensorflow/tensorflow/pull/10598#issuecomment-30832473) that I change the tolerances for those but since the tests passed on the next build, I did not change the tolerances. Should I go ahead and change them?

drpngx:
It's not just the tolerance, check this out:

```
21:10:19 <class 'numpy.ndarray'> <class 'numpy.ndarray'> <class 'tensorflow.python.ops.variables.Variable'> <class 'tensorflow.python.ops.variables.Variable'>
21:10:19 <class 'tensorflow.python.framework.ops.Tensor'> <class 'tensorflow.python.framework.ops.Tensor'>
21:10:19 not close where =  (array([7, 7], dtype=int64), array([6, 7], dtype=int64))
21:10:19 not close lhs =  [ 1.  0.]
21:10:19 not close rhs =  [    0.          1570.79632679]
21:10:19 not close dif =  [  1.00000000e+00   1.57079633e+03]
21:10:19 not close tol =  [  1.00000000e-05   1.57179633e-02]
```

tensorflow/core/kernels/cwise_ops.h
```diff
@@ -43,6 +43,39 @@ struct functor_traits<scalar_fmod2_op<T>> {
...
+};
+template<typename T>
+struct functor_traits<scalar_asinh_op<T>> {
+  enum { Cost = 5 * NumTraits<T>::MulCost, PacketAccess = false };
+};
+
+template<typename T> struct scalar_acosh_op {
+  EIGEN_EMPTY_STRUCT_CTOR(scalar_acosh_op)
+  EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE const T operator()(const T& a) const {
+    return std::acosh(a);
```
lakshayg: It's surprising that everything works fine on Linux and fails on Windows. I don't think this is the case but could it have something to do with the standard library implementation of acosh, since that is what is used in the background for computation.

PR #10600, opened 2017-06-09
Fixes python doc for tanh (Resolves #10376)

### LINKED ISSUE #10376: Lack support of qint32 in tf.nn.tanh
According to the [doc](https://www.tensorflow.org/versions/master/api_docs/python/tf/tanh), the `tanh` operation supports floating point inputs as well as fixed point inputs of type qint32. However, in the latest master, a `TypeError` raised when running following code:
```
import tensorflow as tf
x = tf.constant([1.,2.,3.], dtype=tf.float32)

from tensorflow.python.ops.gen_array_ops import quantize_v2
x_quant = quantize_v2(x, min_range=0., max_range=4., T=tf.qint32)
y_quant = tf.nn.tanh(x_quant[0])
```
The complete error message is 
```
TypeError: Value passed to parameter 'x' has DataType qint32 not in list of allowed values: float16, float32, float64, complex64, complex128
```

According to the backend function `_tanh` in `gen_math_ops.py`:
```
def _tanh(x, name=None):
  r"""Computes hyperbolic tangent of `x` element-wise.

  Args:
    x: A `Tensor`. Must be one of the following types: `half`, `float32`, `float64`, `complex64`, `complex128`.
    name: A name for the operation (optional).
```
It shows that it doesn't support qint32.

lakshayg:


aselle (approved):
Thanks for the patch.

martinwicke:
Jenkins, test this please.

PR #10623, opened 2017-06-11
Change np.array to tf.constant

magixsno:
np.array fails with no 'module' get_shape().


magixsno:
I signed it!

caisq (approved):

caisq:
Thanks, @magixsno !

Issue #5986, opened 2016-11-30
TensorFlow on Windows logs an error when using GPUs: "Could not identify NUMA node"


nobsu:
I use one GPU card now,  tensorflow(0.12) reports follow error, but the process run correctly.   How to suport  if we use multi-gpu on windows ?

E c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\core\common_runtime\gpu\gpu_device.cc:586] Could not identify NUMA node of /job:localhost/replica:0/task:0/gpu:0, defaulting to 0.  Your kernel may not have been built with NUMA support.

mrry:
This error should be a warning. Multiple GPUs work today on Windows, and they should work with a NUMA setup (although we haven't tested this). However, the stream executor code will not take advantage of NUMA topology if it is available.

mrry:
@poxvoculi I think you know most about the more complex bus topologies we use. What are we missing if we don't have GPU-to-NUMA-node mapping information in stream executor?

poxvoculi:
The code and comments in gpu_device.cc clarify what's going on.  We use the StreamExecutor interface to get NUMA affinity for a GPU, then use that to set a (PCI) bus identifier for the GPU.  In the generic open-source version of TensorFlow, the bus id is of no consequence because there are no adaptations for multi-bus architectures.    It would be possible to utilize this feature to customize TF for higher performance in some kinds of multi-bus architecture.

mrry:
Should we remove this log statement from `tensorflow/stream_executor` then? Or should we even remove the code for querying the NUMA affinity, since the bus ID is never used?

poxvoculi:
That should probably be a LOG(INFO), not an error.

PR #10629, opened 2017-06-11
[WIP] Multi-bus support with NUMA-aware allocator

byronyi:
As the title suggests, this is a WIP to partially address issues brought by @poxvoculi in #5986. 

In our testbed, we have a multi-bus topology where 4 K40m GPUs are connected by different PCI-e root, as shown with the following command:

```
$ nvidia-smi topo -m
	GPU0	GPU1	GPU2	GPU3	mlx4_0	CPU Affinity
GPU0	 X 	PHB	SOC	SOC	SOC	0-5
GPU1	PHB	 X 	SOC	SOC	SOC	0-5
GPU2	SOC	SOC	 X 	PHB	PHB	6-11
GPU3	SOC	SOC	PHB	 X 	PHB	6-11
mlx4_0	SOC	SOC	PHB	PHB	 X

Legend:

  X   = Self
  SOC  = Connection traversing PCIe as well as the SMP link between CPU sockets(e.g. QPI)
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe switches (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing a single PCIe switch
  NV#  = Connection traversing a bonded set of # NVLinks
```

Current TF always allocates the host memory used to transfer from/to GPU on NUMA node 0, which is sub-optimal for GPUs with other CPU affinity. This could be further validated by adjusting `CUDA_VISIBLE_DEVICES` and `numactl -m <numa_node> -N <numa_node>` prepended to TF process.

The basic idea is simple: use ``numa_alloc_onnode`` and other related functions available in [libnuma](https://linux.die.net/man/3/numa) to allocate memory for CUDA host allocator. There are some pieces of code, e.g. [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L669), [there](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/stream_executor/cuda/cuda_gpu_executor.cc#L951), and [there](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/stream_executor/cuda/cuda_gpu_executor.cc#L855) to identify bus id and numa node for each device, and I plan to reuse those as much as possible.

I am wondering if anyone in the TF team could give me some advice so I can start to implement it.

### DIFF
--- a/tensorflow/core/common_runtime/gpu/gpu_device_factory.cc
+++ b/tensorflow/core/common_runtime/gpu/gpu_device_factory.cc
@@ -42,7 +42,8 @@ class GPUDevice : public BaseGPUDevice {
     if (attr.on_host()) {
       if (attr.gpu_compatible() || force_gpu_compatible_) {
         ProcessState* ps = ProcessState::singleton();
-        return ps->GetCUDAHostAllocator(0);
+        int numa_node = executor()->GetDeviceDescription().numa_node();
+        return ps->GetCUDAHostAllocator(numa_node);
       } else {
         return cpu_allocator_;
       }
--- a/tensorflow/core/common_runtime/gpu/process_state.cc
+++ b/tensorflow/core/common_runtime/gpu/process_state.cc
@@ -152,11 +152,7 @@ Allocator* ProcessState::GetGPUAllocator(const GPUOptions& options, int gpu_id,
 }
 
 Allocator* ProcessState::GetCPUAllocator(int numa_node) {
-  // Although we're temporarily ignoring numa_node, check for legality.
   CHECK_GE(numa_node, 0);
-  // TODO(tucker): actually maintain separate CPUAllocators for
-  // different numa_nodes.  For now, just one.
-  numa_node = 0;
   mutex_lock lock(mu_);
   while (cpu_allocators_.size() <= static_cast<size_t>(numa_node)) {
     bool use_bfc_allocator = false;
@@ -196,18 +192,14 @@ Allocator* ProcessState::GetCPUAllocator(int numa_node) {
     }
     cpu_allocators_.push_back(allocator);
   }
-  return cpu_allocators_[0];
+  return cpu_allocators_[numa_node];
 }
 
 Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
   if (!HasGPUDevice() || !FLAGS_brain_mem_reg_cuda_dma) {
     return cpu_allocator();
   }
-  // Although we're temporarily ignoring numa_node, check for legality.
   CHECK_GE(numa_node, 0);
-  // TODO(tucker): actually maintain separate CPUAllocators for
-  // different numa_nodes.  For now, just one.
-  numa_node = 0;
   mutex_lock lock(mu_);
 
   // Find the first valid StreamExecutor to request CUDA host memory
--- a/tensorflow/core/common_runtime/gpu/process_state.cc
+++ b/tensorflow/core/common_runtime/gpu/process_state.cc
@@ -249,8 +249,8 @@ Allocator* ProcessState::GetCUDAHostAllocator(int numa_node) {
           &mem_desc_map_, cuda_host_allocators_.back(), md, &mu_));
     }
   }
-  if (FLAGS_brain_gpu_record_mem_types) return cuda_al_[0];
-  return cuda_host_allocators_[0];
+  if (FLAGS_brain_gpu_record_mem_types) return cuda_al_[numa_node];
+  return cuda_host_allocators_[numa_node];
 }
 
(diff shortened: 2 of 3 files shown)


poxvoculi:
I'm out of the office this week with poor internet connectivity.  I hope
this can wait till next week, or someone else can review it.

byronyi:
Sure. I can start researching myself in the mean time. Hope that I could make some progress before your first review.

drpngx:
@poxvoculi can this be merged? I wonder if we should try this internally first to see if there are perf regressions.

poxvoculi:
@byronyi: do you see any significant performance improvements with this change?  We experimented with per-numa-node buffer allocation a long time ago, and didn't see any significant effect on our platforms.  I suppose we should revisit the question.

byronyi:
@poxvoculi This is a primer patch to a larger one that tries to implement NIC-compatible host allocator and GPU allocator for RDMA (w/ GPU Direct). We had a similar design to current contrib/verbs implementation but we would like to avoid memory copies at all. Our initial prototype shows that the overhead of ad-hoc memory registration is noticeable, and the best solution seems to be allocating memory exclusively from a RDMA capable memory pool (with memory window registration).

byronyi:
I am wondering if #4185 is related, as I am not familiar with code in the stream executor part.

poxvoculi:
@byronyi 

If you look at https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/process_state.h#L43
you'll see some indication of how these issues are handled in the Google internal version, which is specific to our hardware platform.  (Handling of networking related issues is nearly the only difference between internal and open-source versions.)   Memory is either GPU RAM or CPU RAM.  CPU RAM may optionally be gpu_registered which means that it was allocated via CUDAHostAllocator and both page locked and known to the CUDA driver as such.  gpu_registered memory is used for CPU resident tensors which are going to be transferred to/from GPU RAM via a copy operation.  nic_registered memory has been preregistered for RDMA.  GPU RAM is pre-allocated in large regions by the allocator (BFCAllocator), and when we're planning to do RDMA and GPUDirect is available, we go ahead and preregister all of those GPU memory regions with the bus-adjacent NIC.  If we plan on doing RDMA with CPU RAM as well, we preregister all memory in the pool obtained from CUDAHostAllocator with all NICs bus-adjacent to CPU RAM.

The numa-related bits that you're touching in this change were written into the code base a long time ago on the assumption that someday we'd want do numa-specific memory allocation and registration.  In fact, we don't.  The page locking and NIC registration issues are important, with significant performance gain.  But we have not been able to get any gain out of maintaining multiple numa-specific memory pools, and having multiple pools will probably incur greater memory overhead. 

This lack of benefit probably has something to do with characteristics of our platforms and execution environment, but I would encourage you to address the other issues first, and only try to change the numa-0 default here when doing so realizes demonstrable benefit.

byronyi:
@poxvoculi Thanks a lot to that bit of information. It clearly points out the direction for my next step, so I won't waste too much time on NUMA-specific issues. In fact, I think your feedback confirmed many observations I had, which is worth my effort digging into the internals of memory management. At least I know I am on the right track.

Personally I found it quite interesting that GPU DMA does share a lot of similarities with RDMA, and there is plenty of GPU DMA code in the open sourced TF that could be reused. So I will just go ahead and implement it.

If you don't mind, I have two more questions: 

1. As we do have verbs support now, would you mind to accept another PR doing RDMA? It is designed to handle many issues in a better way, i.e. an interface that utilises out-of-band ``transport_options`` in current gRPC, connection management using `librdmacm` (I already implemented these), zero-copy of tensor buffer and no extra memory allocation (as we discussed above), and GPU direct when available. However, the performance gain may not be substantial than what we have now, which, after several patches since it's merged, is close to linear scaling on a small testbed. Additionally, it will probably touch a lot different places outside `core/distributed_runtime` and adds more code complexity.
2. You mentioned there are multiple NICs bus-adjacent to different PCI-e devices in Google's internal platform, and we are also working on hardware spec that fits TF running on RoCE. I understand it might be confidential, but would it be a performance boost good enough to justify the extra NICs? I will probably consider that in my patch as well, just to deal with different pairings of GPU and NIC, or to decide if we should do GPU direct at all.

Thanks again for your comments!

reedwm:
@byronyi 6448f41d0b0ee45fea482ee14952f006245feb56 allows a BFC allocator to be used instead of a pool allocator for the CPUAllocator, because pool allocators can run out of memory quickly if many allocs and frees are made with many different sizes. I know almost nothing about NUMA and RDMA, but since 6448f41d0b0ee45fea482ee14952f006245feb56 did not change the host allocator, I don't think it will have an affect on RDMA.

PR #10643, opened 2017-06-12
Implements tf.arg for complex numbers (Closes #483)

### LINKED ISSUE #483: Request for angle
Getting the angle of a complex number is another mathematical operation that would be useful / basic to have, along the lines of [np.angle](http://docs.scipy.org/doc/numpy/reference/generated/numpy.angle.html).

lakshayg:
The following code runs as expected so I believe that the gradients are working fine
```python
import tensorflow as tf

x0 = tf.constant(1.0)
y0 = tf.constant(2.0)
c0 = tf.complex(x0, y0)

y = tf.Variable(0.4)
c = tf.complex(x0, y)

err = tf.abs(c - c0)
optimize = tf.train.AdagradOptimizer(1e-1).minimize(err)

for i in range(10000):
    sess.run(optimize)
    if i % 10 == 0:
        print(sess.run(c)) # should eventually converge to (1 + 2j)

```

### DIFF
--- a/tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
+++ b/tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
@@ -13,7 +13,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-#if GOOGLE_CUDA
+// TODO: Enable GPU support for arg op after resolving
+// build failures on GPU (See #10643 for context).
+#if 0 && GOOGLE_CUDA
 
 #include "tensorflow/core/kernels/cwise_ops_gpu_common.cu.h"
 
--- a/tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
+++ b/tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
@@ -1,0 +1,26 @@
+/* Copyright 2015 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#if GOOGLE_CUDA
+
+#include "tensorflow/core/kernels/cwise_ops_gpu_common.cu.h"
+
+namespace tensorflow {
+namespace functor {
+DEFINE_UNARY2(get_arg, complex64, complex128);
+}  // namespace functor
+}  // namespace tensorflow
+
+#endif  // GOOGLE_CUDA
--- a/tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
+++ b/tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-// TODO: Enable GPU support for arg op after resolving
+// TODO: Enable GPU support for angle op after resolving
 // build failures on GPU (See #10643 for context).
 #if 0 && GOOGLE_CUDA
 
@@ -21,7 +21,7 @@ limitations under the License.
 
 namespace tensorflow {
 namespace functor {
-DEFINE_UNARY2(get_arg, complex64, complex128);
+DEFINE_UNARY2(get_angle, complex64, complex128);
 }  // namespace functor
 }  // namespace tensorflow
 
--- a/tensorflow/core/ops/math_ops.cc
+++ b/tensorflow/core/ops/math_ops.cc
@@ -2065,7 +2065,7 @@ tf.imag(input) ==> [4.75, 5.75]
 ```
 )doc");
 
-REGISTER_OP("Arg")
+REGISTER_OP("Angle")
     .Input("input: T")
     .Output("output: Tout")
     .Attr("T: {complex64, complex128} = DT_COMPLEX64")
@@ -2085,8 +2085,12 @@ For example:
 
 ```
 # tensor 'input' is [-2.25 + 4.75j, 3.25 + 5.75j]
-tf.arg(input) ==> [2.0132, 1.056]
+tf.angle(input) ==> [2.0132, 1.056]
 ```
+
+@compatibility(numpy)
+Equivalent to np.angle.
+@end_compatibility
 )doc");
 
 REGISTER_OP("Conj")
--- a/tensorflow/core/ops/math_ops.cc
+++ b/tensorflow/core/ops/math_ops.cc
@@ -2065,6 +2065,30 @@ tf.imag(input) ==> [4.75, 5.75]
 ```
 )doc");
 
+REGISTER_OP("Arg")
+    .Input("input: T")
+    .Output("output: Tout")
+    .Attr("T: {complex64, complex128} = DT_COMPLEX64")
+    .Attr("Tout: {float, double} = DT_FLOAT")
+    .SetShapeFn(shape_inference::UnchangedShape)
+    .Doc(R"doc(
+Returns the argument of a complex number.
+
+Given a tensor `input` of complex numbers, this operation returns a tensor of
+type `float` that is the argument of each element in `input`. All elements in
+`input` must be complex numbers of the form \\(a + bj\\), where *a*
+is the real part and *b* is the imaginary part.
+
+The argument returned by this operation is of the form \\(atan2(b, a)\\).
+
+For example:
+
+```
+# tensor 'input' is [-2.25 + 4.75j, 3.25 + 5.75j]
+tf.arg(input) ==> [2.0132, 1.056]
+```
+)doc");
+
 REGISTER_OP("Conj")
     .Input("input: T")
     .Output("output: T")
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -100,6 +100,7 @@ See the @{$python/math_ops} guide.
 @@complex
 @@conj
 @@imag
+@@arg
 @@real
 @@fft
 @@ifft
@@ -622,10 +623,9 @@ def imag(input, name=None):
   r"""Returns the imaginary part of a complex number.
 
   Given a tensor `input` of complex numbers, this operation returns a tensor of
-  type `float32` or `float64` that is the imaginary part of each element in
-  `input`. All elements in `input` must be complex numbers of the form \\(a +
-  bj\\), where *a* is the real part and *b* is the imaginary part returned by
-  this operation.
+  type `float` that is the argument of each element in `input`. All elements in
+  `input` must be complex numbers of the form \\(a + bj\\), where *a*
+  is the real part and *b* is the imaginary part returned by the operation.
 
   For example:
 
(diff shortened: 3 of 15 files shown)


caisq:
@lakshayg i don't see the new `tf.arg` used anywhere in the example code you provided above.

lakshayg:
@caisq Thanks for pointing that out. All this time I was testing on the incorrect code. I just tested on that actual code I was supposed to test on and it seems I have stumbled upon a bug. I will be revising the commits and resubmitting shortly. Here is the correct code

```python

theta0 = tf.atan2(1.0, 1.0) # pi/4
x0 = tf.constant(1.0)
y  = tf.Variable(0.4)
c  = tf.complex(x0, y)
theta = tf.arg(c)
err = (theta - theta0)**2
optimize = tf.train.GradientDescentOptimizer(1e-1).minimize(err)

for i in range(10000):
    sess.run(optimize)
    if i % 10 == 0:
        print(sess.run(c)) # should go to 1 + 1j
```

lakshayg:
@caisq I have just amended the commit. The posted code works on my machine. Please review.

drpngx:
@lakshayg sorry, it looks like there are merge conflicts now. Feel free to merge and rebase.
CC @asimshankar for go

lakshayg:
@drpngx Conflicts resolved!

drpngx:
Jenkins, test this please.

lakshayg:
Linux GPU test shows NO STATUS for all the tests. Seems to be a problem with the build infrastructure.

drpngx:
So, the XLA is more of a problem. This is a new one, so it could be unrelated to this change.

We need to re-trigger the Linux GPU test to see what's going on.

Jenkins, test this please.

lakshayg:
@drpngx Linux GPU does the same again
```
Executed 0 out of 319 tests: 1 test passes and 318 were skipped.

Parameterized build ends with FAILURE at: Tue Jun 27 16:59:38 UTC 2017 (Elapsed time: 104 s)
Build step 'Execute shell' marked build as failure
[Set GitHub commit status (universal)] ERROR on repos [] (sha:5dcc47c) with context:tensorflow-pull-requests-gpu
Setting status of 65e5a2cfe96dba87ebbf26e8e264acaa6c009067 to FAILURE with url https://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/5513/ and message: 'FAILURE
```

drpngx:
@gunan Linux GPU and XLA are both new, right?

drpngx:
@gunan never mind, it seems to succeed on other builds, so it's probably specific to this one.

drpngx:
BTW, `tf.arg` is probably not a good name. I would suggest `angle`, like `numpy`.

lakshayg:
@drpngx There is some discussion about the name for this operation here: https://github.com/tensorflow/tensorflow/issues/483. `tf.arg` seems to be the one that was finally chosen.

drpngx:
Jenkins, test this please.

lakshayg:
@drpngx these tests (XLA & GPU) are failing again. Do you know of a reason which might be causing this failure?

drpngx:
The code fails to build:

```
external/eigen_archive/unsupported/Eigen/CXX11/src/Tensor/TensorAssign.h(136): error: calling a __host__ function("__builtin_cargf") from a __device__ function("Eigen::internal::EigenMetaKernelEval< ::Eigen::TensorEvaluator<const  ::Eigen::TensorAssignOp< ::Eigen::TensorMap< ::Eigen::Tensor<float, (int)1, (int)1, int> , (int)16,  ::Eigen::MakePointer> , const  ::Eigen::TensorCwiseUnaryOp< ::Eigen::internal::scalar_arg_op< ::std::complex<float> > , const  ::Eigen::TensorMap< ::Eigen::Tensor<const  ::std::complex<float> , (int)1, (int)1, int> , (int)16,  ::Eigen::MakePointer> > > ,  ::Eigen::GpuDevice> , int, (bool)0> ::run") is not allowed
```

lakshayg:
@drpngx Could you please run the tests again. I don't have a GPU machine to run them locally.

lakshayg:
Jenkins, test this please.

gunan:
Jenkins, test this please.

drpngx:
Here's the error:

```
NFO: From Compiling tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc:
external/eigen_archive/unsupported/Eigen/CXX11/src/Tensor/TensorAssign.h(136): error: calling a __host__ function("__builtin_cargf") from a __device__ function("Eigen::internal::EigenMetaKernelEval< ::Eigen::TensorEvaluator<const  ::Eigen::TensorAssignOp< ::Eigen::TensorMap< ::Eigen::Tensor<float, (int)1, (int)1, int> , (int)16,  ::Eigen::MakePointer> , const  ::Eigen::TensorCwiseUnaryOp< ::Eigen::internal::scalar_arg_op< ::std::complex<float> > , const  ::Eigen::TensorMap< ::Eigen::Tensor<const  ::std::complex<float> , (int)1, (int)1, int> , (int)16,  ::Eigen::MakePointer> > > ,  ::Eigen::GpuDevice> , int, (bool)0> ::run") is not allowed

external/eigen_archive/unsupported/Eigen/CXX11/src/Tensor/TensorAssign.h(136): error: calling a __host__ function("__builtin_carg") from a __device__ function("Eigen::internal::EigenMetaKernelEval< ::Eigen::TensorEvaluator<const  ::Eigen::TensorAssignOp< ::Eigen::TensorMap< ::Eigen::Tensor<double, (int)1, (int)1, int> , (int)16,  ::Eigen::MakePointer> , const  ::Eigen::TensorCwiseUnaryOp< ::Eigen::internal::scalar_arg_op< ::std::complex<double> > , const  ::Eigen::TensorMap< ::Eigen::Tensor<const  ::std::complex<double> , (int)1, (int)1, int> , (int)16,  ::Eigen::MakePointer> > > ,  ::Eigen::GpuDevice> , int, (bool)0> ::run") is not allowed

2 errors detected in the compilation of "/tmp/tmpxft_00006593_00000000-7_cwise_op_gpu_arg.cu.cpp1.ii".
ERROR: /workspace/tensorflow/core/kernels/BUILD:2358:1: output 'tensorflow/core/kernels/_objs/cwise_op_gpu/tensorflow/core/kernels/cwise_op_gpu_arg.cu.o' was not created.
ERROR: /workspace/tensorflow/core/kernels/BUILD:2358:1: not all outputs were created or valid.
```

lakshayg:
Another attempt at solving this. Please test again.

drpngx:
Jenkins, test this please.

lakshayg:
I have reimplemented the op from scratch just to make sure that I am not missing something. Please test this.

drpngx:
Jenkins, test this please.

lakshayg:
Can't seem to get my head aroud the problem. Should I just remove the GPU kernels at the moment, open an issue for writing tf.arg gpu kernels and send in a PR for that later?

lakshayg:
I have just removed the GPU support for arg. Please test again.

vrv:

lakshayg:
```
ERROR: No test targets were found, yet testing was requested.
```

This is the error message in the log. Any hints on how to proceed? @vrv

gunan:
We identified the culprit for that problem, and created a rollback for it.
Once that is pushed, the test should be back to normal.
Later today we should be able to test your PR again.

vrv:

lakshayg:
This is ready to be merged.

vrv:
(Adding final API review tag before we can merge).

tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
```diff
@@ -0,0 +1,28 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+// TODO: Enable GPU support for arg op after resolving
+// build failures on GPU (See #10643 for context).
+#if 0 && GOOGLE_CUDA
```
aselle: Have the build failures been resolved? I'm concerned about putting a new op without gpu support.

tensorflow/python/ops/math_ops.py
```diff
@@ -646,6 +646,35 @@ def imag(input, name=None):
     return gen_math_ops.imag(input, Tout=input.dtype.real_dtype, name=name)
 
 
+def arg(input, name=None):
```
aselle: I don't believe the wrapper is necessary, as the default wrapper will have the exact same functionality. Are you just doing this to provide a python example and to use a name_scope wrapper?

tensorflow/python/ops/math_ops.py
```diff
@@ -100,6 +100,7 @@
 @@complex
 @@conj
 @@imag
+@@arg
```
aselle: I'd argue we should perhaps only put this in tf.math.arg rather than in the root namespace. It doesn't seem that commonly used.

lakshayg: But then we would have some functions in the root namespace and others in math. Other than being less frequently used is there a reason to move to a namespace separate from other related functions like abs, conj etc?

tensorflow/python/ops/math_ops.py
lakshayg: Is this not the case with `imag`? I am doing this only to maintain consistency with the way other ops are defined.

tensorflow/python/ops/math_ops.py
aselle: We have a more extensive namespacing plan, where all the operators will eventually be in more appropriate sub-namespaces. So conj, abs, etc. will be in math. This is similar to say numpy where not everything is in the root numpy. I know in numpy we have numpy.angle, but we have already diverged on name.

tensorflow/core/kernels/cwise_op_gpu_arg.cu.cc
lakshayg: No they haven't been resolved. Since I don't have access to a machine with GPU, I am unable to debug the build failures. Could you suggest a way around this problem?

tensorflow/python/ops/math_ops.py
lakshayg: So should I move it or leave it here and then it will be moved when all others are moved.

asimshankar:
(For API review): Deferring to @aselle. A few notes:

- Would it make sense for this to be a composite op in Python instead of defining a TensorFlow op? Something like:
```python
def arg(value, name=None):
  return tf.atan2(tf.imag(value), tf.real(value))
```
Or am I missing something?

- You don't need to update the `.go` files, it will be updated by a cron job that keeps the generated wrappers for Go in sync.

- And what's the story about `numpy.angle` vs. `tf.arg`. If this is a new op, should we call it `angle`? @aselle ?

rmlarsen:
@asimshankar It would be vastly slower to implement this as a composite arg, and might not even handle corner cases correctly. Since Eigen now has an arg functor, we should use it. I'm OK with deferring the GPU versions until the issues have been resolved.

rmlarsen (approved):

rmlarsen:
@lakshayg  As @asimshankar  suggests, we should call the op angle for Numpy compatibility. 
@aselle do you agree?

tensorflow/core/ops/math_ops.cc
```diff
@@ -2065,6 +2065,30 @@ tf.imag(input) ==> [4.75, 5.75]
 ```
 )doc");
 
+REGISTER_OP("Arg")
```
rmlarsen: Change to angle for Numpy compatibility.

tensorflow/core/ops/math_ops.cc
```diff
@@ -2065,6 +2065,30 @@ tf.imag(input) ==> [4.75, 5.75]
...
+
+The argument returned by this operation is of the form \\(atan2(b, a)\\).
+
+For example:
+
+```
+# tensor 'input' is [-2.25 + 4.75j, 3.25 + 5.75j]
+tf.arg(input) ==> [2.0132, 1.056]
+```
+)doc");
```
rmlarsen: Add

@compatibility(numpy)
Equivalent to numpy.angle.
@end_compatibility

tensorflow/python/ops/math_ops.py
rmlarsen: I think you can leave it here for now, and we will move everything together.

rmlarsen (changes requested):

lakshayg:
Just renamed arg to angle. Please review.

tensorflow/core/ops/math_ops.cc
```diff
@@ -2065,6 +2065,30 @@ tf.imag(input) ==> [4.75, 5.75]
...
+is the real part and *b* is the imaginary part.
+
+The argument returned by this operation is of the form \\(atan2(b, a)\\).
+
+For example:
+
+```
+# tensor 'input' is [-2.25 + 4.75j, 3.25 + 5.75j]
+tf.angle(input) ==> [2.0132, 1.056]
+```
```
rmlarsen: Could you please the @compatibility decorator here, as requested?

rmlarsen (changes requested):

tensorflow/core/ops/math_ops.cc
lakshayg: done!

tensorflow/core/ops/math_ops.cc
lakshayg: done!

rmlarsen: Thanks!

rmlarsen: Hmm I dobn't see it. Did you forget to merge it?

lakshayg: I amended the commit. Maybe you need to look for the changes here: [a7c7907](https://github.com/tensorflow/tensorflow/pull/10643/commits/a7c7907b9d0b539ec5cf9bdc3c98a387b84f8935)

rmlarsen: Thanks.

rmlarsen (approved):

asimshankar:
(Looks good for API change)

rmlarsen:
@asimshankar Thanks!

PR #10657, opened 2017-06-12
Lazily configure TensorFlow logger

### LINKED ISSUE #10498: TensorFlow logging configuration is non-standard
Ref https://github.com/tensorflow/tensorflow/issues/8023

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04
- **TensorFlow installed from (source or binary)**: Docker image
- **TensorFlow version (use command below)**: 1.1.0
- **Bazel version (if compiling from source)**: N/A
- **CUDA/cuDNN version**: Per Docker image
- **GPU model and memory**: K80
- **Exact command to reproduce**: N/A

### Describe the problem
The logging configuration in `tf_logging` is odd and inconsistent with how things are generally done in Python.

In general, it's typical to attach handlers to the root logger in Python: https://docs.python.org/2/library/logging.html#logging.Logger.propagate

However, because `tf_logging` sets up its own stream handler, and does not disable the propagate flag, anybody who does follow the standard Python convention of attaching handlers to the root logger gets duplicated log output from TensorFlow.

Per https://github.com/tensorflow/tensorflow/issues/8023, while this isn't a bug per se, it's still wrong and undesirable.

The cleanest way forward is probably to just set `propagate=False` on the TF logger.

taion:
Fixes #10498

### DIFF
--- a/tensorflow/python/platform/tf_logging.py
+++ b/tensorflow/python/platform/tf_logging.py
@@ -30,64 +30,92 @@ from logging import ERROR
 from logging import FATAL
 from logging import INFO
 from logging import WARN
+import threading
 
 import six
 
 from tensorflow.python.util.all_util import remove_undocumented
 
 
-# Determine whether we are in an interactive environment
-_interactive = False
-try:
-  # This is only defined in interactive shells
-  if _sys.ps1: _interactive = True
-except AttributeError:
-  # Even now, we may be in an interactive shell with `python -i`.
-  _interactive = _sys.flags.interactive
+# Don't use this directly. Use _get_logger() instead.
+_logger = None
+_logger_lock = threading.Lock()
 
-# Scope the tensorflow logger to not conflict with users' loggers
-_logger = _logging.getLogger('tensorflow')
 
-# If we are in an interactive environment (like jupyter), set loglevel to info
-# and pipe the output to stdout
-if _interactive:
-  _logger.setLevel(INFO)
-  _logging_target = _sys.stdout
-else:
-  _logging_target = _sys.stderr
+def _get_logger():
+  global _logger
 
-# Add the output handler
-_handler = _logging.StreamHandler(_logging_target)
-_handler.setFormatter(_logging.Formatter(_logging.BASIC_FORMAT, None))
-_logger.addHandler(_handler)
+  # Use double-checked locking to avoid taking lock unnecessarily.
+  if _logger:
+    return _logger
+
+  _logger_lock.acquire()
+
+  try:
+    if _logger:
+      return _logger
+
+    # Scope the TensorFlow logger to not conflict with users' loggers.
+    logger = _logging.getLogger('tensorflow')
+
+    # Don't further configure the TensorFlow logger if the root logger is
+    # already configured. This prevents double logging in those cases.
+    if not _logging.getLogger().handlers:
+      # Determine whether we are in an interactive environment
+      _interactive = False
+      try:
+        # This is only defined in interactive shells.
+        if _sys.ps1: _interactive = True
+      except AttributeError:
+        # Even now, we may be in an interactive shell with `python -i`.
+        _interactive = _sys.flags.interactive
+
+      # If we are in an interactive environment (like Jupyter), set loglevel
+      # to INFO and pipe the output to stdout.
+      if _interactive:
+        logger.setLevel(INFO)
+        _logging_target = _sys.stdout
+      else:
+        _logging_target = _sys.stderr
+
+      # Add the output handler.
+      _handler = _logging.StreamHandler(_logging_target)
+      _handler.setFormatter(_logging.Formatter(_logging.BASIC_FORMAT, None))
+      logger.addHandler(_handler)
+
+    _logger = logger
+    return _logger
+
+  finally:
+    _logger_lock.release()
 
 
 def log(level, msg, *args, **kwargs):
-  _logger.log(level, msg, *args, **kwargs)
+  _get_logger().log(level, msg, *args, **kwargs)
 
 
 def debug(msg, *args, **kwargs):
-  _logger.debug(msg, *args, **kwargs)
+  _get_logger().debug(msg, *args, **kwargs)
 
 
 def error(msg, *args, **kwargs):
-  _logger.error(msg, *args, **kwargs)
+  _get_logger().error(msg, *args, **kwargs)
 
 
 def fatal(msg, *args, **kwargs):
-  _logger.fatal(msg, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


taion:
cc @tatatodd @mrry

taion:
More correct, I think, would be to drop automatically attaching the handler, and instead just use the `warning` library for deprecation warnings.

It's very odd to automatically opt-in on INFO-level logging. That should usually be a user-level concern.

mrry:
Reassigning this to @dave-andersen, since he wrote the affected code (and I have no sense of what consequences this change will have).

taion:
To be concrete, the problem is that, currently:

```
>>> import logging, tensorflow as tf; logging.basicConfig(); tf.logging.info('foo')
INFO:tensorflow:foo
INFO:tensorflow:foo
```

This is not really what a library should be doing.

drpngx:
@dave-andersen any thoughts on that?

dave-andersen:
Sorry for not seeing things.  Now aware and paying attention.

@taion - do you see a solution that results in two properties:
 (a)  Users out-of-the-box see info warnings if they configure:
  `tf.logging.set_verbosity(tf.logging.INFO)

and
(b)  We don't get this duplicate behavior if the user manually imports logging?  (@mrry, I didn't actually write all the code in question, I was just the last fool to touch it when trying to fix an earlier manifestation of this same problem. :)

vrv:
Also cc'ing @yilei who might care about this.

taion:
@dave-andersen 

What about this bit? https://github.com/tensorflow/tensorflow/blob/v1.3.0-rc0/tensorflow/python/platform/tf_logging.py#L51-L57

In what cases is it desirable to you to automatically log things to the console, without explicit user setup?

dave-andersen:
@taion - I'm not quite sure how to interpret your comment. :)  Are you saying that what's in the platform for _interactive is bad and we should change it, or are you pointing it out as a case in which it's desirable to default to verbose output?

In my view, there's a tricky thing here that TensorFlow is a library for some people, and kind of more an application for others, at least, when they're doing things like testing DNNs in Jupyter, etc.

taion:
There's a few things going on here.

- Using `logging` for _usage_ warning from a library is not ideal. It's much better to use the [`warnings`](https://docs.python.org/3.6/library/warnings.html) package from the stdlib. This is what e.g. numpy does for things like dividing by zero, and it sidesteps all the (sort of bizarre) Python logging configuration. `warnings` also exposes a bunch of hooks to users to do stuff like assert that code does not throw warnings, or to throw on hitting warnings, or to only log each warning once.
- Setting up a log handler for a non-root logger is sort of unidiomatic; normally, it'd be possible to do things like set up my own logging format and handlers, and it's unusual to have to e.g. reach into a library's logging config to disable what that library does

In terms of options, I see:

1. Use `warnings` for usage warnings, and drop the automatic logging setup. This is probably the most correct (assuming the things you want to show are usage warnings), and e.g. matches what numpy does, but it's a lot of work, and might require rethinking log levels.
2. Disable propagation for the TF logger **(this PR)**. It makes it still be more of a hassle than I'd like, if I want to customize my log message format, but it's enough for me to be reasonably satisfied. This stops the double-logging if there are user-supplied root log handlers, and otherwise keeps working as things do currently.
3. Defer the log handler setup until the first time logging runs. At that point, check to see if there are any configured root log handlers. If not, set up the `tensorflow` log handler. This is a little hacky, but it's sort of a best-of-both – by deferring logging setup to first use (given that TF doesn't log anything when you import it), users get a chance to configure their own logging in an idiomatic manner, and have things otherwise "just work".

rmlarsen:
@vrv @dave-andersen can we come to a resolution about this. Reject this change?

vrv:
@rmlarsen I'm not really a decision maker here, so I will defer to others.  I like 3 as I think it's better hygiene to wait until users can set things up before making decisions about the loggers.

@craigcitro might have some idea of whether this impacts stuff like Jupyter notebook logging in interesting ways.

jhseu:
@dave-andersen Friendly ping. Any thoughts on what we should do with this change?

vrv:
I think proposal 3 from @taion's latest comment was the ideal since it requires the least amount of work and has the fewest implications for going wrong.  @taion would you like to give this a try?

craigcitro:
I realize I also failed to chime in: 

* logging doesn't show up by default in Jupyter unless a user calls something like `logging.basicConfig(level=logging.INFO)` themselves. This is the same "users won't see info logs unless they opt in" that matches @taion's expectations above.

* @taion you're right that "I see INFO logs by default" is an unusual default in the outside world. As you can probably guess, it's the common default *inside* Google.

I'm with @vrv that (3) is probably the best choice. I'm generally against (1) since I find the `warnings` module pretty cumbersome.

taion:
I'll take a look later this week or early next week.

We actually do mostly set log level to `INFO` as well, but we want _our_ log format, not yours :p (And we definitely don't want to see all the logging twice.)

drpngx:
@taion Any chance to take a look?

@vrv, @dave-andersen, @craigcitro . To recap, the recommendation is:

Defer the log handler setup until the first time logging runs. At that point, check to see if there are any configured root log handlers. If not, set up the tensorflow log handler. This is a little hacky, but it's sort of a best-of-both – by deferring logging setup to first use (given that TF doesn't log anything when you import it), users get a chance to configure their own logging in an idiomatic manner, and have things otherwise "just work".

taion:
Sorry, I've been a bit swamped lately. I'll try to look this week.

dave-andersen:
(fwiw, i concur about #3, though I do worry that it may be fragile in a way we haven't expected -- but the examples that come to mind are weird, such as someone configuring logging *after* tf starts loading, which is already buggy, given that it would lose logging messages).

sb2nov:
@taion PTAL

taion:
Updated.

The explicit lock is with the double-checked locking is meh, but the lock is unavoidable and is consistent with what upstream Python logging does.

The 100% semantically correct solution would be to use the same lock as `basicConfig`, but it's not exposed, and anyway if a user is racing `logging.basicConfig` against TensorFlow's logging on multiple threads, they deserve whatever weird stuff they end up getting.

martinwicke:
@dave-andersen triage ping? What should happen here?

taion:
Any update here? I have implemented (3) as discussed above.

martinwicke (approved):
I think this is great.

martinwicke:
Jenkins, test this please.

martinwicke:
@sb2nov This is ready to merge assuming it passes tests.

PR #10660, opened 2017-06-12
Upgrade to protobuf 3.3.1

jhseu:
(Haven't tested, letting Jenkins try it out)

### DIFF
--- a/tensorflow/java/maven/proto/pom.xml
+++ b/tensorflow/java/maven/proto/pom.xml
@@ -16,7 +16,7 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
-      <version>3.2.0</version>
+      <version>3.3.1</version>
     </dependency>
   </dependencies>
 
--- a/tensorflow/java/maven/run_inside_container.sh
+++ b/tensorflow/java/maven/run_inside_container.sh
@@ -23,7 +23,7 @@ IS_SNAPSHOT="false"
 if [[ "${TF_VERSION}" == *"-SNAPSHOT" ]]; then
   IS_SNAPSHOT="true"
 fi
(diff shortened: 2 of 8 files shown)


jhseu:
Jenkins, test this please

jart (approved):

martinwicke:
Does this have implications on our installation instructions, the pip package dependencies, or other documentation?

jhseu:
It should be ok to update the build dependency without updating the pip) runtime dependency. They're used for two different purposes right now.

pip tests would also catch any issues.

martinwicke:
different purposes? So using two different protobufs in different parts of the code isn't a problem?

jhseu:
Yeah, the interaction points only test wire-compatibility, which is guaranteed by protobuf. The different versions only occur when serialization happens in the Python runtime (the pip version) and it's deserialized in TF core (bazel version in pywrap_tensorflow).

jart:
Good catch @martinwicke. OOC is there any reason why we _wouldn't_ want to bump our setup.py protobuf dependency from 3.2.0 -> 3.3.1? Purely for the sake of consistency?

Our Python code probably depends on `@com_google_protobuf//:python_srcs` somehow, for things like `from google.protobuf import json_format`. There's been changes to their Python API since 3.2.0 and probably other subtleties we haven't considered.

martinwicke:
Let's update the pip dependency as well. I think Jonathan is right and it
won't be a problem. But consistency is nice.

jhseu:
Bumped all protobuf versions throughout. pypi doesn't have 3.3.1 yet, so depended on 3.3.0 instead.

jart:
LGTM. Thanks @jhseu.

jhseu:
Also had to downgrade some other spots to 3.3.0: only source is available for 3.3.1. The prebuilt protoc binary is not available.

martinwicke:
Jenkins, test this please.

Issue #8531, opened 2017-03-19
Add support for non-scalar string tensors in java


ophiry:
Currently, string tensors are not supported in java:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/java/src/main/java/org/tensorflow/Tensor.java#L90

Implementing this  would be very helpful in scenarios of training a model in one language, and serving from a different one, as it allows moving some of the logic to the tensor flow model.

asimshankar:
I'm under the impression that someone is working on a contribution for this. I'll check in with them to see if they expect to send a PR anytime soon.

erranli:
I completed the implementation of the feature and will contribute it back soon.

jan-apptornado:
Thanks a lot for making this possible. Any idea when "soon" is?

quaeler:
@erranli could you push your fix, or cut this issue loose?

asimshankar:
@quaeler : I had an earlier attempt at this than I can resurrect. I'll do that and hope to have something in the next few days.

quaeler:
@asimshankar ok sounds good; if you want to pass the baton, i'm happy to do the resurrection as well.

quaeler:
Thanks @ebrevdo - any reason not make this into a PR for tensorflow master?

ebrevdo:
Please ignore my commit; this was written by @asimshankar and will show up on master soon.

myanything:
@caisq thank you for making the change. Curious when will the change be published to maven central?

asimshankar:
This is part of 1.4, so is available on Maven Central.

PR #10678, opened 2017-06-13
Add jni methods to support creating string Tensors from string bytes

Moriadry-zz:
Here are native methods to support creating string Tensors by string bytes. The implement of Tensor class and test will be soon available.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
@@ -618,6 +618,12 @@ public final class Tensor implements AutoCloseable {
 
   private static native byte[] scalarBytes(long handle);
 
+  private static native int getEncodedStringSize(int numStringBytes);
+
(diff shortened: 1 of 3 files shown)


asimshankar:
Thanks for the PR @Moriadry , but I think I'm lacking some context here.
What's the plan for these methods? Could you elaborate on what you intend to do with these methods?

Moriadry:
@asimshankar I'm trying to add this feature [string support](https://github.com/tensorflow/tensorflow/issues/8531) . And here is part of job I have done.

I thought the problem was when calling Tensor.create(Object o), for example,  with byte[][]

`byte[][] bytes = {{“x”, 1}, {“y”, “z"}, {“a”, “b”}, {“m”, 2}};`

we need to allocate the string in off-heap memory, and nativeHandle returns its address.

Do I learn it correctly? Is there any thing different between string and primitive? Thank you for your help :)

Moriadry:
I will create a new pr when all code is finished, close it now, thanks

PR #10690, opened 2017-06-13
Fix the f16 implementation of Literal and LiteralUtil

DavidNorman:
Adjust the implementation of Literal to do the right thing for F16.   The new Literal class is nice, in that it is just a std::vector of half types, which is nice and simple.

The Proto version still has to have a bit of munging around due to protobuf limitations.

### DIFF
--- a/tensorflow/compiler/xla/literal_util_test.cc
+++ b/tensorflow/compiler/xla/literal_util_test.cc
@@ -966,5 +966,62 @@ TEST_F(LiteralUtilTest, CopyFromProto_Bool) {
   }
 }
 
+// Note that f16 is currently stored in a byte array in little endian byte order
+TEST_F(LiteralUtilTest, ToProto_f16) {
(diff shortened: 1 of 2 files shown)


tensorflow/compiler/xla/literal_util_test.cc
```diff
@@ -966,5 +966,62 @@ TEST_F(LiteralUtilTest, CopyFromProto_Bool) {
   }
 }
 
+// Note that f16 is currently stored in a byte array in little endian byte order
```
hawkinsp: Perhaps say "serialized" rather than "stored". "stored" is a bit ambiguous because "stored" could mean in memory and in memory the floats are now in the host byte order. Same below.

(I guess we need to add some byte swapping code to ToProto but not in this CL.)

hawkinsp (approved):

hawkinsp:

PR #10696, opened 2017-06-14
Add input interfaces required by the Java Ops API.

karllessard:
Those interfaces are at the base of the upcoming Java Ops API, please consult https://github.com/tensorflow/tensorflow/issues/7149 for more details.


tensorflow/java/src/main/java/org/tensorflow/Input.java
```diff
@@ -0,0 +1,61 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+package org.tensorflow;
+
+/**
+ * A tensor that can be used as an operand for operation wrappers.
```
asimshankar: How about:

Interface implemented by operands of a TensorFlow operation.

?

tensorflow/java/src/main/java/org/tensorflow/Input.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * <p>The {@code Input} interface is at the base of the Java Ops API by allowing any object that
```
asimshankar: Nit: Don't need to say "Java Ops API", Java is implicit :). Maybe just remove this paragraph and jump into:

Example usage:

tensorflow/java/src/main/java/org/tensorflow/Input.java
```diff
@@ -0,0 +1,61 @@
...
+ * wraps an input tensor to be passed as an operand to an operation.
+ *
+ * <pre>{@code
+ * // Output implements Input.
```
asimshankar: This by itself might raise questions, how about saying:

// The Output of the "decodeJpeg" operation can be used as input to the "cast" operation?

tensorflow/java/src/main/java/org/tensorflow/Input.java
```diff
@@ -0,0 +1,61 @@
...
+ * Output image = ops.image().decodeJpeg(...).image();
```
asimshankar: For this snippet, let's keep it simpler and just do:

```java
Output image = ops.image().decodeJpeg(...);
ops.math().cast(image, DataType.FLOAT);
```

instead of examples both with and without `.image()` (especially since `DecodeJPEG` has only one output - showing both examples I think raises a couple of questions that we don't need to raise right now)

tensorflow/java/src/main/java/org/tensorflow/Input.java
```diff
@@ -0,0 +1,61 @@
...
+ * // DecodeJpeg operation also implements Input, returning the image as its output.
+ * ops.math().cast(ops.image().decodeJpeg(...), DataType.FLOAT);
+ * }</pre>
+ *
+ * <p>Additionally, objects implementing {@code Iterable<Input>} can be passed in input to
+ * operations handling a list of tensors as one of its operand.
+ *
+ * <pre>{@code
+ * // List<Output> extends from Iterable<Input>
+ * List<Output> outputs = ops.array().split(...).output();
```
asimshankar: I see you prefer to use the named outputs, but we also plan to support the operations directly.
If so, I feel it's better to use just the op name (instead of naming the output) when there is a single output. So `split(...)` instead of `split(...).output()`. What do you think?

tensorflow/java/src/main/java/org/tensorflow/Input.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Returns the symbolic handle of a tensor.
+   *
+   * <p>This method is called by the Java Ops API to retrieve the symbolic handle of a tensor added
+   * in input to an operation.
+   *
+   * @see {@link OperationBuilder#addInput(Output)}.
+   */
+  Output asOutput();
```
asimshankar: I know this `Input` and `Output` type naming is confusing, sorry about that :)
How about:

```java
/**
 * Returns the symbolic handle of the tensor.
 *
 * <p>Inputs to TensorFlow operations are outputs of another TensorFlow operation. This method is used to obtain a symbolic handle that represents the computation of the input.
 ...
```
?

tensorflow/java/src/main/java/org/tensorflow/Output.java
```diff
@@ -20,8 +20,11 @@
  *
  * <p>An Output is a symbolic handle to a tensor. The value of the Tensor is computed by executing
  * the {@link Operation} in a {@link Session}.
+ *
+ * <p>By implementing the {@link Input} interface, instances of this class could also be passed
+ * directly in input to an operation wrapper when using the Java Ops API.
```
asimshankar: Again, I'd suggest avoiding the term "Java Ops API", instead just say "directly in input to an operation"?

tensorflow/java/src/main/java/org/tensorflow/VariableInput.java
```diff
@@ -0,0 +1,35 @@
+/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
asimshankar: Let's remove this file for now - since we don't have a Variable class, having this in the code is a bit confusing. Perhaps we can add it when there is a Variable class introduced?

asimshankar (changes requested):
Thanks for the PR @karllessard . I sincerely appreciate sending this smaller "API" PR first as it helps focus on the plan (and the detailed implementation will come later :)

Overall looks great, some minor suggestions on commentary, but otherwise looks nice.

Thanks!

tensorflow/java/src/main/java/org/tensorflow/Input.java
karllessard: In this case, it won't be true, since the DecodeJpeg operation cannot be cast to Output directly. I will change the type of the "image" variable for "Input" and add an additional example where an Output must be selected from multiple-output operation.

tensorflow/java/src/main/java/org/tensorflow/Input.java
karllessard: It was principally to demonstrate that both operations and outputs could be passed directly in input. But from my comment above, the mutliple-output example will show clearly that so I'll remove that variant in here, thanks

tensorflow/java/src/main/java/org/tensorflow/Input.java
karllessard: No problem

tensorflow/java/src/main/java/org/tensorflow/VariableInput.java
karllessard: Just to clarify quickly, VariableInputs are to be used/instantiated by any operations dealing with variable references, not only by Variable. But I can plan a separate PR for everything related to variables if you prefer, that's ok with me too.

asimshankar: Yeah, let's separate those out. Thanks.

asimshankar (approved):

asimshankar:
Jenkins, test this please

PR #10700, opened 2017-06-14
Allow variable sized tensor for RDMA transport

### LINKED ISSUE #10699: RDMA transport should support variable sized tensor
RDMA transport disabled variable size tensor https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/verbs/rdma.cc#L738 and https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/verbs/rdma.cc#L759.

However, in the embedding lookup model, the tensor has variable size. It's a common model parallelism case.

llhe:
This PR fixes https://github.com/tensorflow/tensorflow/issues/10699.

Q: Should we make aggressive buffer extension?

### DIFF
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -735,8 +735,8 @@ void RdmaTensorBuffer::SendNextItem() {
           (buffer_size > size_ && local_status_ == idle &&
            remote_status_ == idle)) {
         if ((local_status_ != none) && (buffer_size > size_)) {
-          CHECK(rm.data_type_ == DT_STRING)
-              << "Only string tensor allows to change size";
+          VLOG(2) << "Extend RDMA buffer from " << size_ << " to "
+                  << buffer_size;
         }
         CreateCPUBuffer(buffer_size, false);
         mu_.unlock();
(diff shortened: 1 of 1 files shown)


junshi15:
The code is fine. The question is whether we should allow buffers to grow infinitely. Ideally, a buffer should shrink if the current size is unnecessarily large. But I could not think a case where the tensor size is temporary, i.e.  one will likely see the tensor of the same size in the future training steps.

drpngx:
@michaelisard what do you think?

drpngx:
Jenkins, test this please.

drpngx:
@poxvoculi if you have suggestions.

poxvoculi:
We have seen models where there is a very large number of different tensor sizes that don't necessarily repeat.  If there's only one buffer and it is allocated from CPU RAM, it probably doesn't matter.

junshi15:
There is one buffer per tensor, allocated in CPU memory. 

The original assumption was that tensor sizes (except string tensors) did not change over the course of training. It was valid for the early models, e.g. InceptionV3, VGG16, etc. As new models appear, the assumption is not valid anymore. I think we should let buffer sizes vary. 

We could have some kind of policy to reclaim the memory if the tensor sizes have been smaller than the buffer size for some period of time. Note the buffers are pinned and registered for RDMA. Frequent allocation and registration can slow down things. The patch is good for now due to its simplicity.

poxvoculi:
Use of verbs/rdma is optional.  If the model won't run with this constraint, user can just drop back to gRPC.  I'm ok with this.  @drpngx do you want me to review, or wait for Michael?

drpngx:
@poxvoculi I'm not sure how much @michaelisard knows about this, maybe you have the most context on this one.

poxvoculi (approved):

drpngx:
Thanks!

PR #10701, opened 2017-06-14
Try both python and python3

Androbin:
Used `configure` script for reference
Reopens mistakenly closed #10525

### DIFF
--- a/tools/tf_env_collect.sh
+++ b/tools/tf_env_collect.sh
@@ -19,6 +19,7 @@ set -u  # Check for undefined variables
 echo "Collecting system information..."
 
 OUTPUT_FILE=tf_env.txt
+python_bin_path=$(which python || which python3 || true)
 
 {
   echo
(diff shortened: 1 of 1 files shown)


tools/tf_env_collect.sh
```diff
@@ -19,6 +19,7 @@ set -u  # Check for undefined variables
 echo "Collecting system information..."
 
 OUTPUT_FILE=tf_env.txt
+python_bin_path=$(which python || which python3 || true)
```
martinwicke: If we don't find python or python3, is it reasonable to not error out?

Androbin: Good point, had just taken from [`configure#L76`](https://github.com/tensorflow/tensorflow/blob/master/configure#L76)
So, should we display an appropriate error message?

martinwicke: It's probably a super-rare case, but that would be nicest.

martinwicke (approved):

martinwicke:
skipping tests since this is untested.

PR #10721, opened 2017-06-15
DataFeeder fails on default random_state

mdymczyk:
Fixes `len() of unsized object` error in [DataFeeder](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py#L367) due to incorrect object type.

Tested with Python 3.5 and latest Tensorflow (1.1.0). Failed on

```
import tensorflow as tf
from tensorflow.contrib.learn import KMeansClustering
from tensorflow.contrib.learn import SKCompat

def my_input_fn(df):
    return {k: tf.constant(df[k].values) for k in df.columns}

def train_input_fn():
    return my_input_fn(pd.DataFrame(train_data)) # train_data is anything that can be passed to pd.DataFrame()

model = SKCompat(KMeansClustering(num_clusters=k))
model.fit(x = train_input_fn(), y = None)
```

Resulted in:

```
tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py in __init__(self, x, y, n_classes, batch_size, shuffle, random_state, epochs)
    365         0] if x_is_dict else self._x.shape[0]
    366     if self._shuffle:
--> 367       self.indices = self.random_state.permutation(num_samples)
    368     else:
    369       self.indices = np.array(range(num_samples))

mtrand.pyx in mtrand.RandomState.permutation (numpy/random/mtrand/mtrand.c:39959)()

mtrand.pyx in mtrand.RandomState.shuffle (numpy/random/mtrand/mtrand.c:38843)()

TypeError: len() of unsized object
```

Couldn't find any myself, but is there any workaround for this?


mdymczyk:
I signed it!

yifeif:

mdymczyk:
@yifeif my bad, I see that either a NumPy array (like in the failing tests) or a Tensor (like in [this](https://www.tensorflow.org/get_started/input_fn) example) can be passed to the feeder.

mdymczyk:

drpngx:
Jenkins, test this please.

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
```diff
@@ -363,6 +363,11 @@ def __init__(self,
 
     num_samples = list(self._x.values())[0].shape[
```
drpngx: Can you refactor the logic above so that we have all three cases clearly checked?

frankchn:
ping @mdymczyk

vrv:
one more ping for @mdymczyk -- can you address the review comments?

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
mdymczyk: what do you exactly have in mind? I thought there are only 2 cases (NumPy array or Tensor).

rmlarsen:
@mdymczyk Could you please address the comments?

mdymczyk:
@rmlarsen I asked for clarification 2 weeks ago and got no reply. Thought the way it is now covers all the cases already so I could use a pointer what I'm missing.

rmlarsen:
@mdymczyk sorry, missed that.
@drpngx Could you respond, please?

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
drpngx: Sorry for the delay. I think it's a bit harder to read this way. If `x_is_dict` then you have a list, which never has a `value` attr. It's only in the latter case that you need to check for value. So I think the cases are:

* `x_is_dict`
* else
    * shape has a value attr
    * shape doesn't

In general, I think you want to use `is_tensor` from `tensor_utils.py` rather than checking for the attr.

jhseu:
@mdymczyk friendly ping if there's any update.

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
drpngx: @mdymczyk can you use `is_tensor` as recommended?

mdymczyk: @drpngx ah yes sorry, this has totally slipped my mind - busy prepping for StrataNY.

sb2nov:
@mdymczyk ping

tensorflow/contrib/learn/python/learn/learn_io/data_feeder.py
```diff
@@ -361,8 +361,13 @@ def __init__(self,
...
 
-    num_samples = list(self._x.values())[0].shape[
-        0] if x_is_dict else self._x.shape[0]
+    if x_is_dict:
+      num_samples = list(self._x.values())[0].shape[0]
+    elif is_tensor(self._x):
+      num_samples = self._x.shape[0].value  # shape will be a Dimension, extract an int
+    else:
+      num_samples = self._x.shape[0]
+
```
martinwicke: @drpngx Is this what you had in mind?

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

martinwicke:
Ignoring CLA, it's confused because of my commits.

martinwicke:
Jenkins, test this please.

martinwicke:
multinomial_test is flaky (and unrelated), ignoring.

PR #10743, opened 2017-06-15
don't lock markdown version and bring CI deps install script in sync with setup.py

chenliu0831:
This markdown version lock could cause versioning conflict for downstream (e.g. if requirements are compiled and resolved by `pip-compile`). Wondering If there's a reason to lock this version 

Seems someone pushed `1.2.0` to PYPI and it cause some dependency conflict.

### DIFF
--- a/tensorflow/tools/ci_build/install/install_pip_packages.sh
+++ b/tensorflow/tools/ci_build/install/install_pip_packages.sh
@@ -28,24 +28,24 @@ pip2 install wheel
 pip3 install wheel
 
 # Install six.
-pip2 install --upgrade six==1.10.0
-pip3 install --upgrade six==1.10.0
+pip2 install --upgrade six>=1.10.0
+pip3 install --upgrade six>=1.10.0
 
 # Install werkzeug.
-pip2 install --upgrade werkzeug==0.11.10
-pip3 install --upgrade werkzeug==0.11.10
+pip2 install --upgrade werkzeug>=0.11.10
+pip3 install --upgrade werkzeug>=0.11.10
 
 # Install bleach. html5lib will be picked up as a dependency.
 pip2 install --upgrade bleach==1.5.0
 pip3 install --upgrade bleach==1.5.0
 
 # Install markdown.
-pip2 install --upgrade markdown==2.6.8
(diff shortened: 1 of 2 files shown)


tensorflow/tools/ci_build/install/install_pip_packages.sh
```diff
@@ -28,33 +28,33 @@ pip2 install wheel
...
 # Remove obsolete version of six, which can sometimes confuse virtualenv.
 rm -rf /usr/lib/python3/dist-packages/six*
 
 # numpy needs to be installed from source to fix segfaults. See:
 # https://github.com/tensorflow/tensorflow/issues/6968
 # This workaround isn't needed for Ubuntu 16.04 or later.
-pip2 install --no-binary=:all: --upgrade numpy==1.12.0
-pip3 install --no-binary=:all: --upgrade numpy==1.12.0
+pip2 install --no-binary=:all: --upgrade numpy>=1.11.0
+pip3 install --no-binary=:all: --upgrade numpy>=1.11.0
```
chenliu0831: I think those should be at least in sync with the [setup.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py)

gunan:
I am inclined to reject the changes in install_pip_packages.sh file.
First, that file is for our continuous integration system, and end users should not be using/depending on it.
Second, in our CI I would like all our dependencies to be tightly controlled.
For example, your changes will cause numpy to be upgraded to 1.13, which causes a lot of test failures due to a bug in the new numpy version. Without a version lock on numpy version, our CI will start breaking and debugging such issues would be very difficult.

For the changes to setup.py, I think that is a tensorboard dependency.
Please send a pull request with this change to our brand new tensorboard repository for discussion.

chenliu0831:
OK. I don't think lock versions are a great thing to do especially it's not in sync with the actual `setup.py` which breaks the parity between CI and end user. @gunan

gunan:
Thanks for your input, however we will favor reproducibility in our CI to help us with debugging. `setup.py` is designed to be more flexible to enable a wider variety of user environments, but our CI has to be reproducible so that we can clearly pinpoint any issues in our build history.

Issue #781, opened 2016-01-15
rotate an image for data augmentation


mostafachatillon:
I would like to rotate an image from a random angle, for data augmentation. But I don't find this transformation in the tf.image module.

see: http://stackoverflow.com/questions/34801342/tensorflow-how-to-rotate-an-image-for-data-augmentation

bhack:
I will extend this request to more transformation like https://github.com/BVLC/caffe/pull/2252

ducha-aiki:
@bhack thanks! Actually, I could think about implementing it...but need to get time for at least installing TF.

jkschin:
@ducha-aiki are you going to implement it? I'm interested to contribute this functionality. 

Should I implement it first then send a pull request? Or is someone going to be assigned this. Sorry for the questions. I'm new to how contributions work.

ducha-aiki:
@jkschin not now, so you can try first :)
>Should I implement it first then send a pull request? 
Yes, usually it is done like this.

jkschin:
Alright. What I intend to do would be to first implement some form of smart padding, so not much information is lost. If we look at python PIL, when we rotate the image, some parts are left black. We just have to pad the image by a scale of sqrt(2), do a rotation, find the center and extract the same sized square.

This is how I intend to do it, suggestions and improvements are welcome!

ducha-aiki:
@jkschin There is a little discussion about it in https://github.com/BVLC/caffe/pull/2252
The following ones are implemented in different libraries:
- pad with zeros
- pad with random noise
- pad with mean pixel (nice for images)
- pad with border mirroring.

jkschin:
@ducha-aiki thanks!

Another question. Is it practice to assume a square image? Or do we have to factor in rectangular use cases as well?

ducha-aiki:
@jkschin TensorFlow is very flexible instrument, so rectangular as well.

bhack:
Probably we'll need also GPU ops version at some point like https://github.com/tensorflow/tensorflow/pull/588

cesarsalgado:
What about the case when we get, for example, an 256x256 image and we want to extract a random rotated 224x224 crop from WITHIN the image? In this case we don't need padding. Should we have in mind this use case?

sdemyanov:
[...]

PR #10748, opened 2017-06-16
Add a tf.contrib.image.translate function

sampepose:
Progress on this issue: https://github.com/tensorflow/tensorflow/issues/781

### DIFF
--- a/tensorflow/contrib/image/python/ops/image_ops.py
+++ b/tensorflow/contrib/image/python/ops/image_ops.py
@@ -81,6 +81,48 @@ def rotate(images, angles, interpolation="NEAREST"):
     return output
 
 
+def translate(images, translations, interpolation="NEAREST"):
+  """Translate image(s) by the passed vectors(s).
+
+  Args:
+    images: A tensor of shape (num_images, num_rows, num_columns, num_channels)
+       (NHWC), (num_rows, num_columns, num_channels) (HWC), or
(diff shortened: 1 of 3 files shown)


tensorflow/contrib/image/python/ops/image_ops.py
```diff
@@ -81,6 +81,48 @@ def rotate(images, angles, interpolation="NEAREST"):
...
+  if len(image_or_images.get_shape()) == 2:
+    images = image_or_images[None, :, :, None]
+  elif len(image_or_images.get_shape()) == 3:
+    images = image_or_images[None, :, :, :]
+  elif len(image_or_images.get_shape()) == 4:
+    images = image_or_images
+  else:
+    raise TypeError("Images should have rank between 2 and 4.")
+
+  output = transform(
```
ringw: You should be able to remove everything but this statement because `transform` does the same checks, and we don't do anything else with `images` like we do in `rotate`.

ringw (changes requested):

drpngx:
@sampepose how about what @ringw suggested?

vrv:
Ping for @sampepose !

ringw:
Closing in favor of #12306, where I made a small simplification. @sampepose can you please confirm there that it's ok to merge in your commit? Thanks!

PR #10698, opened 2017-06-14
[OpenCL] Fixes grpc test failure for SYCL devices (#87)

lukeiwanski:
The test constructs a graph containing an IdentityOp on strings,
which doesn't exist on CUDA and SYCL devices. The test expects to fail
for CUDA devices, so add the same expectation for SYCL devices.

PR #10758, opened 2017-06-16
[OpenCL] Fixes grpc test failure for SYCL devices

jwlawson:
The test constructs a graph containing an IdentityOp on strings, which doesn't exist on CUDA and SYCL devices. The test expects to fail for CUDA devices, so add the same expectation for SYCL devices.

Resubmission of #10698


caisq (approved):

caisq:

PR #10778, opened 2017-06-16
Merging r1.2 back into master.

av8ramit:


martinwicke (approved):

jhseu (changes requested):
This change deletes the non_max_suppression_v2 op. Let's keep it in master and just resolve it as part of b/62486586?

jhseu:
In other words, omit 323d40fce604e9e31975714197ce4a7064b55dc2

martinwicke:
yes please.​

allenlavoie (approved):
Either way works for the RewriterConfig PR; sync it or don't sync it, whatever's easiest. There's a CL coming the other direction which does about the same thing but includes a bit more explanation.

tensorflow/core/ops/image_ops.cc
```diff
@@ -1030,50 +1030,11 @@ method: A string specifying the interpolation method. Only 'bilinear' is
 // --------------------------------------------------------------------------
 
 REGISTER_OP("NonMaxSuppression")
-  .Input("boxes: float")
```
jhseu: This probably shouldn't be deleted. Not sure where it's coming from.

jhseu (approved):

PR #10797, opened 2017-06-17
Tutorial: Pass train_file.name instead of train_file

sahildua2305:
Change read_csv argument from train_file to train_file.name in tutorial
for Linear model.

Passing train_file to read_csv doesn't load any data into dataframe
and the dataframe remains empty. It works when changed to file name.


yifeif (approved):

yifeif:
Thanks @sahildua2305!

PR #10831, opened 2017-06-19
Documentation fix: Referencing correct pip in installation examples

chirag64:
The examples are supposed to be for Python2.7 so it makes sense that they refer to the correct pip (pip instead of pip3) to avoid any potential confusion.

### DIFF
--- a/tensorflow/docs_src/install/install_linux.md
+++ b/tensorflow/docs_src/install/install_linux.md
@@ -171,7 +171,7 @@ Take the following steps to install TensorFlow with Virtualenv:
      issue the following command to install TensorFlow in the active
      virtualenv environment:
 
-     <pre>(tensorflow)$ <b>pip3 install --upgrade \
+     <pre>(tensorflow)$ <b>pip install --upgrade \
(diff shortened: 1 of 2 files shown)


tensorflow/docs_src/install/install_linux.md
```diff
@@ -171,7 +171,7 @@ Take the following steps to install TensorFlow with Virtualenv:
      issue the following command to install TensorFlow in the active
      virtualenv environment:
 
-     <pre>(tensorflow)$ <b>pip3 install --upgrade \
```
gunan: In this file, the above version needs to be updated.
The pip package version below is 3.4 so chancing this wont fix the documentation.

gunan (changes requested):

tensorflow/docs_src/install/install_linux.md
chirag64: I found multiple instances of where this is happening and fixed it 👍🏼

gunan (approved):

gunan:
Jenkins, test this please.

Issue #10715, opened 2017-06-14
Feature Request: Update tf.image.sample_distorted_bounding_box to use Scalar Tensor Inputs


blake-varden:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Ubuntu 14.04
- **TensorFlow installed from (source or binary)**:
binary
- **TensorFlow version (use command below)**:
1.1
- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**:
8.0/5.1
- **GPU model and memory**:
Titan X
- **Exact command to reproduce**:

Can you please update the scalar inputs to tf.image.sample_distorted_bounding_box to also be accepted as Tensors.  I am dynamically creating min_object_covered values (random values) but currently can not input them to the function

aselle:
This looks like it could be done w/o breaking shape inference, but it would need to be done w/o breaking graphdef compatibility by creating a duplicate of the sample bounding box op as a V2. I am labeling as contributions welcome since this is not likely going to be worked on soon.

One work around would be to make a few discrete values that would be acceptible and use a switch statement to select between several bounding box ops.

yongtang:
Created a PR #10840 for this issue.

PR #10840, opened 2017-06-20
Update sample_distorted_bounding_box to use Tensor Input for min_object_covered

yongtang:
This fix tries to address the issue raised in #10715 where it was not possible
to use dynamic values for min_object_covered in sample_distorted_bounding_box.

This fix created an Op of SampleDistortedBoundingBoxV2 so that min_object_covered
takes input. SampleDistortedBoundingBox remains the same (min_object_covered as attr).

Additional test cases have been added.

### DIFF
--- a/tensorflow/core/kernels/sample_distorted_bounding_box_op.cc
+++ b/tensorflow/core/kernels/sample_distorted_bounding_box_op.cc
@@ -192,18 +192,20 @@ bool GenerateRandomCrop(int original_width, int original_height,
 }  // namespace
 
 template <typename T>
-class SampleDistortedBoundingBoxOp : public OpKernel {
+class SampleDistortedBoundingBoxV2Op : public OpKernel {
  public:
-  explicit SampleDistortedBoundingBoxOp(OpKernelConstruction* context)
+  explicit SampleDistortedBoundingBoxV2Op(OpKernelConstruction* context)
       : OpKernel(context) {
     OP_REQUIRES_OK(context, generator_.Init(context));
 
-    OP_REQUIRES_OK(
-        context, context->GetAttr("min_object_covered", &min_object_covered_));
-    OP_REQUIRES(
-        context, min_object_covered_ >= 0,
-        errors::InvalidArgument("Min object covered must be non-negative: ",
-                                min_object_covered_));
+    if (context->num_inputs() == 2) {
+      OP_REQUIRES_OK(context, context->GetAttr("min_object_covered",
+                                               &min_object_covered_));
+      OP_REQUIRES(
+          context, min_object_covered_ >= 0,
+          errors::InvalidArgument("Min object covered must be non-negative: ",
+                                  min_object_covered_));
+    }
 
     OP_REQUIRES_OK(context, context->GetAttr("use_image_if_no_bounding_boxes",
                                              &use_image_if_no_bounding_boxes_));
@@ -275,6 +277,25 @@ class SampleDistortedBoundingBoxOp : public OpKernel {
                     "bounding boxes must have shape [4] or [*, 4], got ",
                     input_boxes.shape().DebugString()));
 
+    float min_object_covered_val = 0.0;
+    if (context->num_inputs() == 3) {
+      const Tensor& min_object_covered = context->input(2);
+
+      OP_REQUIRES(
+          context, TensorShapeUtils::IsScalar(min_object_covered.shape()),
+          errors::InvalidArgument("min_object_covered must be 0-D, got shape ",
+                                  min_object_covered.shape().DebugString()));
+
+      min_object_covered_val = min_object_covered.scalar<float>()();
+
+      OP_REQUIRES(
+          context, min_object_covered_val >= 0,
+          errors::InvalidArgument("Min object covered must be non-negative: ",
+                                  min_object_covered_val));
+    } else {
+      min_object_covered_val = min_object_covered_;
+    }
+
     std::vector<Rectangle> bounding_boxes;
     if (input_boxes.NumElements() > 0) {
       TTypes<float>::ConstMatrix boxes = input_boxes.flat_inner_dims<float>();
@@ -325,7 +346,7 @@ class SampleDistortedBoundingBoxOp : public OpKernel {
 
       if (GenerateRandomCrop(width, height, min_sample_area, max_sample_area,
                              sample_aspect_ratio, &random, &crop_rect)) {
-        if (SatisfiesOverlapConstraints(crop_rect, min_object_covered_,
+        if (SatisfiesOverlapConstraints(crop_rect, min_object_covered_val,
                                         bounding_boxes)) {
           sample_generated = true;
           break;
@@ -399,11 +420,15 @@ class SampleDistortedBoundingBoxOp : public OpKernel {
   bool use_image_if_no_bounding_boxes_;
 };
 
-#define REGISTER_KERNELS(type)                               \
-  REGISTER_KERNEL_BUILDER(                                   \
-      Name("SampleDistortedBoundingBox").Device(DEVICE_CPU)   \
-           .TypeConstraint<type>("T"),                       \
-      SampleDistortedBoundingBoxOp<type>)
+#define REGISTER_KERNELS(type)                                  \
+  REGISTER_KERNEL_BUILDER(Name("SampleDistortedBoundingBox")    \
+                              .Device(DEVICE_CPU)               \
+                              .TypeConstraint<type>("T"),       \
+                          SampleDistortedBoundingBoxV2Op<type>) \
+  REGISTER_KERNEL_BUILDER(Name("SampleDistortedBoundingBoxV2")  \
+                              .Device(DEVICE_CPU)               \
+                              .TypeConstraint<type>("T"),       \
+                          SampleDistortedBoundingBoxV2Op<type>)
 
 TF_CALL_INTEGRAL_TYPES(REGISTER_KERNELS);
 #undef REGISTER_KERNELS
--- a/tensorflow/core/ops/image_ops.cc
+++ b/tensorflow/core/ops/image_ops.cc
@@ -82,9 +82,8 @@ Status DecodeImageShapeFn(InferenceContext* c) {
     channels_dim = c->MakeDim(channels);
   }
 
-  c->set_output(0,
-                c->MakeShape({InferenceContext::kUnknownDim,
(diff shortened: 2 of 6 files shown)


yifeif:
@vrv could you take a look at this change? Thanks!

tensorflow/core/kernels/sample_distorted_bounding_box_op.cc
```diff
@@ -275,6 +277,22 @@ class SampleDistortedBoundingBoxOp : public OpKernel {
...
 
+    if (context->num_inputs() == 3) {
+      const Tensor& min_object_covered = context->input(2);
+
+      OP_REQUIRES(
+          context, TensorShapeUtils::IsScalar(min_object_covered.shape()),
+          errors::InvalidArgument("min_object_covered must be 0-D, got shape ",
+                                  min_object_covered.shape().DebugString()));
+
+      min_object_covered_ = min_object_covered.scalar<float>()();
```
vrv: Compute() can be run by multiple threads, so this is going to result in a potential memory race.  I would suggest doing something like:

float min_object_recovered_val = 0.0;
if (context->num_inputs() == 3) {
  ...
  min_object_recovered_val = min_object_recovered.scalar<float>()();
} else {
  min_object_recovered_val = min_object_recovered_;
}

and then use that float below.

vrv (changes requested):
Looks generally good!  I suspect in the future we can make the default python version of sample_distorted_bounding_box to use the v2 version, but we have to wait a few weeks to do that (for forward compatibility).

yongtang:
@vrv Thanks for the review. The PR has been updated. Please take a look and let me know if there are any issues.

vrv (approved):

vrv:

drpngx:
I think this adds an op and will need API review. Could you run:

```
   $ bazel-bin/tensorflow/tools/api/tests/api_compatibility_test \
          --update_goldens True
```

and push again? It'll be clear what the API change is.

yongtang:
Thanks @drpngx. The PR has been updated to address the api compatibility test failure. Please take a look and let me know if there are any issues.

drpngx:
Ugh, somehow all the commits ended up on the CL. Could you rebase and push again?

drpngx:
Adding @martinwicke for API review.

yongtang:
Thanks @drpngx . The PR has been rebased and pushed.

drpngx:
Jenkins, test this please.

tensorflow/tools/api/golden/tensorflow.image.pbtxt
```diff
@@ -173,6 +173,10 @@ tf_module {
     argspec: "args=[\'image_size\', \'bounding_boxes\', \'seed\', \'seed2\', \'min_object_covered\', \'aspect_ratio_range\', \'area_range\', \'max_attempts\', \'use_image_if_no_bounding_boxes\', \'name\'], varargs=None, keywords=None, defaults=[\'None\', \'None\', \'None\', \'None\', \'None\', \'None\', \'None\', \'None\'], "
   }
   member_method {
+    name: "sample_distorted_bounding_box_v2"
+    argspec: "args=[\'image_size\', \'bounding_boxes\', \'min_object_covered\', \'seed\', \'seed2\', \'aspect_ratio_range\', \'area_range\', \'max_attempts\', \'use_image_if_no_bounding_boxes\', \'name\'], varargs=None, keywords=None, defaults=[\'None\', \'None\', \'None\', \'None\', \'None\', \'None\', \'None\'], "
+  }
```
martinwicke: Please hide this op, as well as the original, and write a python wrapper which calls only the V2 version.

martinwicke (changes requested):

tensorflow/tools/api/golden/tensorflow.image.pbtxt
vrv: Forwards compatibility requirements suggests we need to wait several weeks before changing python to use V2.

drpngx: Aww, right. We could still hide the op, then switch to V2 in a couple of weeks. @yongtang if you're up for it, you'd need to change https://www.github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/hidden_ops.txt#L159 and write a wrapper like this: http://google3/third_party/tensorflow/python/ops/image_ops_impl.py?l=925

yongtang:
@martinwicke @vrv @drpngx The PR has been updated as suggested. Please take a look.

tensorflow/python/ops/image_ops.py
```diff
@@ -62,6 +62,7 @@
 @@non_max_suppression
 @@non_max_suppression_v2
 @@sample_distorted_bounding_box
+@@sample_distorted_bounding_box_v2
```
martinwicke: Please don't expose the _v2 op directly. We need to register both kernels, but then only expose a single python op, which only generates the _v2 op. To do that, you currently have to mark both the v1 and v2 ops "hidden" in the BUILD file, and then write a python wrapper which calls the generated_image_ops._..._v2() directly (ignoring the linter that will complain about a protected symbol access).

We're working on improving this workflow but that's what it is for now.

martinwicke (dismissed):

tensorflow/python/ops/image_ops.py
martinwicke: In fact, it's ever worse: Because we're guaranteeing forward compatibility, this change has to be split in two: One that introduces the new _v2 kernel (hidden), and no less than three weeks later, one that switches the behavior of the python op from v1 to _v2 by hiding the v1 op and adding the python wrapper.

yongtang:
@martinwicke Thanks for the review. There was some confusion from my side before about the workflow. Now the PR has been updated with the following changes:
This PR:
1. New kernel (`SampleDistortedBoundingBoxV2Op`) has been registered.
2. `SampleDistortedBoundingBoxV2` is hidden in python (tensorflow/python/ops/hidden_ops.txt)

Next PR (I will submit it after 3 weeks):
1. `SampleDistortedBoundingBox` will be hidden in python (tensorflow/python/ops/hidden_ops.txt)
2. An wrapper `sample_distorted_bounding_box` will be added in `image_ops_impl.py` and calls `sample_distorted_bounding_box_v2`.

Please take a look at the update PR and let me know if the above changes matches the API workflow.

yongtang:
@martinwicke @drpngx Thanks. The PR has been updated with both `SampleDistortedBoundingBox` and `SampleDistortedBoundingBoxV2` hidden. A wrapper has been added in image_ops_impl.py.

I will create another PR after 3 weeks. 

Please take a look.

tensorflow/python/ops/image_ops.py
drpngx: remove

tensorflow/python/ops/image_ops_impl.py
```diff
@@ -1472,3 +1473,104 @@ def total_variation(images, name=None):
...
+    begin: A `Tensor`. Has the same type as `image_size`. 1-D, containing `[offset_height, offset_width, 0]`. Provide as input to
+      `tf.slice`.
+    size: A `Tensor`. Has the same type as `image_size`. 1-D, containing `[target_height, target_width, -1]`. Provide as input to
+      `tf.slice`.
+    bboxes: A `Tensor` of type `float32`. 3-D with shape `[1, 1, 4]` containing the distorted bounding box.
+      Provide as input to `tf.image.draw_bounding_boxes`.
+  """
+
+  with ops.name_scope(name, 'sample_distorted_bounding_box'):
+    return gen_image_ops._sample_distorted_bounding_box_v2(image_size,
```
drpngx: You need the v1 version, then we'll switch this one to V2 in three weeks.

tensorflow/core/kernels/sample_distorted_bounding_box_op.cc
```diff
@@ -399,11 +420,15 @@ class SampleDistortedBoundingBoxOp : public OpKernel {
...
-#define REGISTER_KERNELS(type)                               \
-  REGISTER_KERNEL_BUILDER(                                   \
-      Name("SampleDistortedBoundingBox").Device(DEVICE_CPU)   \
-           .TypeConstraint<type>("T"),                       \
-      SampleDistortedBoundingBoxOp<type>)
+#define REGISTER_KERNELS(type)                                  \
+  REGISTER_KERNEL_BUILDER(Name("SampleDistortedBoundingBox")    \
+                              .Device(DEVICE_CPU)               \
+                              .TypeConstraint<type>("T"),       \
+                          SampleDistortedBoundingBoxV2Op<type>) \
```
drpngx: I would favor keeping the old code around (the whole class).

tensorflow/core/ops/image_ops.cc
```diff
@@ -1030,16 +1120,16 @@ method: A string specifying the interpolation method. Only 'bilinear' is
 // --------------------------------------------------------------------------
 
 REGISTER_OP("NonMaxSuppression")
-  .Input("boxes: float")
-  .Input("scores: float")
-  .Input("max_output_size: int32")
-  .Output("selected_indices: int32")
-  .Attr("iou_threshold: float = 0.5")
-  .SetShapeFn([](InferenceContext* c) {
+    .Input("boxes: float")
```
drpngx: keep old 2-spaces identation, as per our style guide.

drpngx (changes requested):
@martinwicke or @vrv can chime in about keeping the old code around.

tensorflow/core/kernels/sample_distorted_bounding_box_op.cc
vrv: As long as the v1 API contract is held, I feel like it doesn't matter how the OpKernel is written, and less code is better.  But my risk tolerance might be higher than everyone's :)

tensorflow/python/ops/image_ops.py
yongtang: Thanks. Done.

tensorflow/python/ops/image_ops_impl.py
yongtang: Done.

tensorflow/core/ops/image_ops.cc
yongtang: Done.

drpngx (approved):

drpngx:
Jenkins, test this please.

PR #10849, opened 2017-06-20
Java: equality semantics for Operation and Output

kbsriram:
* Add .equals(), .hashCode() and .toString() for value objects Operation
  and Output.
* Annotate these value objects @Immutable for compile-time validation.


yifeif:
@asimshankar mind taking a look? Thanks!

tensorflow/java/src/main/java/org/tensorflow/Operation.java
```diff
@@ -97,6 +99,46 @@ public Output output(int idx) {
     return new Output(this, idx);
   }
 
+  @Override
+  public int hashCode() {
+    Graph.Reference r = graph.ref();
+    try {
+      // xor the first and last 32 bits of the underlying 64-bit pointer value.
+      return (int) (unsafeNativeHandle ^ (unsafeNativeHandle >>> 32));
```
asimshankar: Do we need to grab a reference on the graph? Seems like the contract for `hashCode` is such that it would be okay to write this as:

```java
public int hashCode() { return Long.valuOf(unsafeNativeHandle).hashCode(); }
```
?

tensorflow/java/src/main/java/org/tensorflow/Output.java
```diff
@@ -57,6 +60,30 @@ public Output asOutput() {
     return this;
   }
 
+  @Override
+  public int hashCode() {
+    return operation.hashCode() + 37 * index + 17;
```
asimshankar: Why this choice?
Actually, would it make sense to use [`Objects.hash(operation, index)`](https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#hash(java.lang.Object...)) instead?

tensorflow/java/src/test/java/org/tensorflow/OperationTest.java
```diff
@@ -46,6 +52,96 @@ public void outputListLengthFailsOnInvalidName() {
...
+      assertTrue(ops.contains(op3));
+      assertTrue(ops.contains(op4));
+    }
+  }
+
+  @Test
+  public void operationToString() {
+    try (Graph g = new Graph()) {
+      Operation op = TestUtil.constant(g, "c", new int[] {1}).op();
+      assertEquals("<Const 'c'>", op.toString());
```
asimshankar: Do we want to specify a format, or just make sure it doesn't crash?
Since the `toString()` is more of a logging/debugging aid, I'd suggest that we don't test for a specific format and just make sure that the function returns something?

tensorflow/java/src/test/java/org/tensorflow/OperationTest.java
```diff
@@ -46,6 +52,96 @@ public void outputListLengthFailsOnInvalidName() {
...
+    }
+  }
+
+  @Test
+  public void outputEquality() {
+    try (Graph g = new Graph()) {
+      Output output = TestUtil.constant(g, "c", 1);
+      Output output1 = output.op().output(0);
+      Output output2 = g.operation("c").output(0);
+      assertNotSame(output, output1);
```
asimshankar: I understand why you have this assertion, but this isn't something we want to promise, right? As in, it should be fine if `Operation.output(index)` were to cache and return the same objects?

I'd suggest that the assertions be for behavior that we want to guarantee from the public API and not be testing the details of the specific implementation (https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html). The same goes for the `operationEquality` test, where it should be fine for `Graph.operation()` to return the same `Operation` object when possible.

asimshankar (changes requested):
Thanks for the PR and sorry for the delayed response. Some suggestions.

tensorflow/contrib/android/BUILD
```diff
@@ -60,6 +60,7 @@ android_library(
         "manual",
         "notap",
     ],
+    deps = ["@errorprone_annotations"],
```
kbsriram: Sure, I'll move it to a separate PR; and followup there.

tensorflow/java/src/main/java/org/tensorflow/Operation.java
kbsriram: This works too - just two heads-up:
- A new object is created each time .hashCode() is called.
- .equals() can throw exceptions when the underlying graph is deleted, but .hashCode() will never throw exceptions - pragmatically don't see any issues here other than exactly when adding/fetching items from a cache/collection with stale objects trigger exceptions.

tensorflow/java/src/main/java/org/tensorflow/Output.java
kbsriram: This is a [suggested approach](https://stackoverflow.com/questions/113511/best-implementation-for-hashcode-method) via Josh Bloch and avoids the implicit array object creation for the varargs, but either way is good enough for me; I'll change to using the utils.

tensorflow/java/src/test/java/org/tensorflow/OperationTest.java
kbsriram: Makes sense - will do, thanks.

tensorflow/java/src/main/java/org/tensorflow/Output.java
asimshankar: According to the stackoverflow thread you linked to `Objects.hash` implements the same approach (though, looking at the sources I'm not sure that that's true).

Anyway, I figured `Objects.hash` should encode the best practice :)

tensorflow/java/src/main/java/org/tensorflow/Operation.java
asimshankar: - A new object is created regardless, no? (`graph.ref()` is creating a new object), so the same for both cases?
- The exception: That seems fine. IIUC, the only contract is that if `a.equals(b)` then `a.hashCode() == b.hashCode()`, right? So, throwing an exception in `hashCode()` doesn't really help anyway?

kbsriram: One sequence is `hashset.add(stale_object)` -- it can succeed if `stale_object.hashCode()` happens to select an empty bucket, but cause a later `hashset.add(good_object)` to fail if `good_object.hashCode()` picks the same bucket and so triggers a `.equals()` comparison.

Pragmatically, raising exceptions on both hashCode() and equals() will generally fail quicker for stale objects within collections. But this is also an internal detail that can be changed if needed - happy to start the way you suggest and revisit if needed.

(The new object comment was just about the difference between Long.valueOf() and primitive longs, separate from whether or not we use Graph.ref(). I think here and in the other hashcode implementation trading-off this micro-optimization in favor of core hashcode calculations works for me as well.)

asimshankar (approved):

asimshankar:
Jenkins, test this please

PR #10865, opened 2017-06-20
Java API to get the size of specified input list of operations.

Moriadry-zz:
This opens up access to the TF_OperationInputListLength C API from java, for operations that return specified input lists.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Operation.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Operation.java
@@ -102,9 +102,7 @@ public final class Operation {
    *
    * <p>An Operation has multiple named inputs, each of which contains either
    * a single tensor or a list of tensors. This method returns the size of
-   * the list of tensors for a specific named input of the operation.The length
-   * of the list should be equal to the number of inputs specified by this
-   * operation's op def.
+   * the list of tensors for a specific named input of the operation.
    *
    * @param name identifier of the list of tensors (of which there may
    *        be many) inputs to this operation.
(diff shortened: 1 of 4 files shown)


yifeif:
@asimshankar could you take a look? Thanks!

tensorflow/java/src/main/java/org/tensorflow/Operation.java
```diff
@@ -97,6 +97,30 @@ public Output output(int idx) {
...
   }
 
+  /**
+   * Returns the size of the given inputs list of Tensors for this operation.
+   *
+   * <p>An Operation has multiple named inputs, each of which contains either
+   * a single tensor or a list of tensors. This method returns the size of
+   * the list of tensors for a specific named input of the operation.The length
+   * of the list should be equal to the number of inputs specified by this
+   * operation's op def.
```
asimshankar: I don't believe we explain what an "op def" is in the API documentation, and we don't really need to for most users. Mind rephrasing this to avoid introducing a new term to the reader? Perhaps you can do something very similar to the Javadoc for `outputListLength`? Which I think just means removing this last sentence.

asimshankar (changes requested):
Thanks for the change. One minor suggestion, otherwise looks great.

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please.

Moriadry:
@asimshankar  Thanks for your review,  and your suggestion does make sense. Wish you
 a good day:)

Issue #78, opened 2015-11-10
Scalar Equation is incorrect ? Documentation


ArEnSc:
```
http://tensorflow.org/tutorials/mnist/beginners/index.md#the_mnist_data

https://upload.wikimedia.org/math/7/3/f/73fc7ef1bf6f6822115c41cff58535e1.png

http://api.tensorflow.org/system/image/body/1704/softmax-regression-scalargraph.png

http://api.tensorflow.org/system/image/body/1707/softmax-regression-scalarequation.png

http://api.tensorflow.org/system/image/body/1705/softmax-regression-vectorequation.png

scalar equation should be:
[W11 * X1 + W12 * X2 +  W13 * X3 + b1]
[W21 * X1 +W22 * X2 + W23 * X3 + b2]
[W31 * X1 + W32 * X2 + W33 * X3 + b3] 
```

vrv:
Thanks for the report -- we'll update the image soon and let you know when this is fixed!

vrv:
Fixed in f6d5a2caedd3982a1e6dcaa99bfafe0539527144 by colah.  Thanks!

vrv:
(will take some time to propagate to the website though)

PR #10871, opened 2017-06-21
Add the shared libraries in the quickstart

superryanguo:
When following the quickstart steps of the golang README.md, below issue is found: 

$ uname -a
Linux ubuntu-16 4.4.0-57-generic #78-Ubuntu SMP Fri Dec 9 23:50:32 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
$ go test github.com/tensorflow/tensorflow/tensorflow/go
/tmp/go-build828413616/github.com/tensorflow/tensorflow/tensorflow/go/_test/go.test: error while loading shared libraries: libtensorflow.so: cannot open shared object file: No such file or directory
FAIL	github.com/tensorflow/tensorflow/tensorflow/go	0.001s

It's due to the env variable, after adding the LIBRARY_PATH and LD_LIBRARY_PATH, it became ok.

$ go test github.com/tensorflow/tensorflow/tensorflow/go
ok  	github.com/tensorflow/tensorflow/tensorflow/go	0.246s

so I updated the README.md file.

### DIFF
--- a/tensorflow/go/README.md
+++ b/tensorflow/go/README.md
@@ -9,37 +9,6 @@ Construct and execute TensorFlow graphs in Go.
 > (`github.com/tensorflow/tensorflow/tensorflow/go`).
 
 ## Quickstart
-1.  Download and extract the TensorFlow C library, preferably into `/usr/local`.
-    GPU-enabled versions require CUDA 8.0 and cuDNN 5.1. For other versions, the
-    TensorFlow C library will have to be built from source (see below).
-
-    -   Linux:
-        [CPU-only](https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-linux-x86_64-1.1.0.tar.gz),
-        [GPU-enabled](https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-gpu-linux-x86_64-1.1.0.tar.gz)
-    -   OS X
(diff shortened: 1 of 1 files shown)


superryanguo:
I signed it!  cla

tensorflow/go/README.md
```diff
@@ -26,6 +26,8 @@ Construct and execute TensorFlow graphs in Go.
     curl -L \
       "https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-${TF_TYPE}-$(go env GOOS)-x86_64-1.1.0.tar.gz" |
     sudo tar -C /usr/local -xz
+    export LIBRARY_PATH=$LIBRARY_PATH:/usr/local/lib
+    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
```
asimshankar: Instead of setting the environment variables, perhaps we can just run `ldconfig`?

Actually, better still, perhaps it would be better to replace the "Quickstart" section with a pointer to the installation instructions on the website: https://www.tensorflow.org/install/install_go ?

asimshankar (changes requested):
Thanks for the PR. Some suggestions.

tensorflow/go/README.md
superryanguo: Sounds good! We can totally replace the "Quickstart" with the web link since there's also a section "Installing into locations other than `/usr/local` " under it, which can make the whole file terse.

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please.

PR #10948, opened 2017-06-21
fix error string format

dgboy2000:
I just got this helpful error message, and I'm submitting a fix for the format.

ValueError: ('num_outputs should be int or long, got %s.', <tf.Tensor 'mul:0' shape=() dtype=int32>)

becomes

ValueError: num_outputs should be int or long, got Tensor("mul:0", shape=(), dtype=int32).

### DIFF
--- a/tensorflow/contrib/layers/python/layers/layers.py
+++ b/tensorflow/contrib/layers/python/layers/layers.py
@@ -1445,7 +1445,8 @@ def fully_connected(inputs,
     ValueError: If x has rank less than 2 or if its last dimension is not set.
   """
   if not isinstance(num_outputs, six.integer_types):
-    raise ValueError('num_outputs should be int or long, got %s.', num_outputs)
+    raise ValueError(
+        'num_outputs should be int or long, got %s.' % (num_outputs,))
 
   layer_variable_getter = _build_variable_getter({'bias': 'biases',
                                                   'kernel': 'weights'})


tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -1445,7 +1445,8 @@ def fully_connected(inputs,
     ValueError: If x has rank less than 2 or if its last dimension is not set.
   """
   if not isinstance(num_outputs, six.integer_types):
-    raise ValueError('num_outputs should be int or long, got %s.', num_outputs)
+    raise ValueError(
+        'num_outputs should be int or long, got %s.' % (num_outputs,))
```
caisq: There is no need for the brackets and comma here, since there is only one item.

caisq (changes requested):

dgboy2000:
What if num_outputs is a tuple? Then you'd get an obscure TypeError: not
all arguments converted during string formatting. This ValueError is only
raised if the type is wrong in the first place (as I experienced), so in my
opinion it's helpful to be robust to this case.

What do you think?

tensorflow/contrib/layers/python/layers/layers.py
dgboy2000: What if num_outputs is a tuple? Then you'd get an obscure TypeError: not
all arguments converted during string formatting. This ValueError is only
raised if the type is wrong in the first place (as I experienced), so in my
opinion it's helpful to be robust to this case.

What do you think?

caisq: OK. That makes sense.

caisq (approved):

caisq:

caisq:
Test failure is unrelated. Merging PR.

Thanks, @dgboy2000 !

PR #10956, opened 2017-06-21
Improve docs for tf.nn.depthwise_conv2d_native

AnishShah:
Hi,

In [`depthwise_conv2d_native` docs](https://www.tensorflow.org/api_docs/python/tf/nn/depthwise_conv2d_native), the formula to calculate the output tensor is not written inside "code" tag for markdown.

Screenshot of the docs -

screenshot from 2017-06-21 21-38-21


yifeif (approved):

yifeif:
Looks like it is not supposed to be code snippet, but this does make it look better. Thanks!

Issue #10867, opened 2017-06-20
C++ api in Debug x64 mode not building for VS2015


pabloxrl:
Hi,

I'm not sure if this is expected behaviour or a problem with the solution file generated by CMake, so my apologies if I'm writing in the wrong place.

I'm trying to build Tensorflow C++ API on Windows using VS2015, and so far I didn't have problems to build Release x64 binaries, which just required me to ensure that the toolset for 64 bits is in use when creating the def file (dump bin.exe and undname.exe in particular).

I could verify that the build works as expected, being able to load models saved in Python and perform inference successfully.

However, when building the debug x64 version of the .dll, I'm being blocked by the following error:

```
104>  tensorflow_static.vcxproj -> D:\out\Debug\tensorflow_static.lib
104>  symbols=1016688, taken=140782, dupes=3472
105>------ Build started: Project: tensorflow, Configuration: Debug x64 ------
105>  Building Custom Rule D:/tensorflow/tensorflow/contrib/cmake/CMakeLists.txt
105>  CMake does not need to re-run because D:/out/CMakeFiles/generate.stamp is up-to-date.
105>LINK : fatal error LNK1189: library limit of 65535 objects exceeded
========== Build: 104 succeeded, 1 failed, 1 up-to-date, 0 skipped ==========
```

Is there any workaround for this issue? Has anyone managed to build the debug version of the library? Is it possible at all?

From what I could google around, this linker issue is not easy to overcome, so feedback is most welcome.

tatatodd:
Perhaps @mrry or @asimshankar might have some ideas here.

pabloxrl:
As a sidenote, I'm trying to build other configurations available in the solution like: MinSizeRel or RelWithDebug and get hit by compiler errors of the form:

`d:\out\external\eigen_archive\eigen\src/Core/products/GeneralMatrixVector.h(351): fatal error C1060: compiler is out of heap space (compiling source file D:\tensorflow\tensorflow\core\kernels\matmul_op.cc)`

pabloxrl:
After digging for a while I found the following (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/README.md):

> Note that the -DCMAKE_BUILD_TYPE=Release flag must match the build configuration that you choose when invoking msbuild. The known-good values are Release and RelWithDebInfo. The Debug build type is not currently supported, because it relies on a Debug library for Python (python35d.lib) that is not distributed by default.
> 
[...]

PR #10962, opened 2017-06-21
expand inline for debug builds to limit number of symbols

guschmue:
https://github.com/tensorflow/tensorflow/issues/10867
debug build does not expand inlines which will generate lots of unused symbols that result in too many entries in the def file for python wrapper dll.
Change debug build to expand inlines.


yifeif (approved):

yifeif:

yifeif:
The failure is due to jenkins tool issue, irrelevant to this PR. Merging.

Issue #97, opened 2015-11-10
How to extract predictions


vkuznet:
Hi, can someone either point to code example or documentation how to extract final predictions after the training the model. For example, it would be nice to complement existing tutorials, e.g. mnist, and show additional (final) step to get prediction out of the trained model.

pannous:
# Test trained model
correct_prediction = tf.equal(tf.argmax(y,1), tf.argmax(y_,1))

abunsen:
+1

vkuznet:
Hi,
well I see that step, but if I print the correct_prediction I got
Tensor("Equal:0", shape=TensorShape([Dimension(None)]), dtype=bool) which is
<class 'tensorflow.python.framework.ops.Tensor'>

and I want to get array of numbers (probabilities). My question is how to get
it from Tensor then?

rafaljozefowicz:
Check this response on stackoverflow: https://stackoverflow.com/questions/33633370/how-to-check-the-contents-of-a-tensor-object-in-tensorflow

pannous:
the array of numbers is returned by eval / run method:
```
prediction=tf.argmax(y,1)
print prediction.eval(feed_dict={x: mnist.test.images})
```

vkuznet:
Thanks, this works if I pass session to eval too.
Here is working example for those who are interested:
```
import input_data
import tensorflow as tf
mnist = input_data.read_data_sets('MNIST_data', one_hot=True)
x = tf.placeholder("float", shape=[None, 784])
W = tf.Variable(tf.zeros([784,10]))
b = tf.Variable(tf.zeros([10]))
y = tf.nn.softmax(tf.matmul(x,W) + b)
y_ = tf.placeholder("float", shape=[None, 10])
cross_entropy = -tf.reduce_sum(y_*tf.log(y))
train_step = tf.train.GradientDescentOptimizer(0.01).minimize(cross_entropy)
# train data and get results for batches
init = tf.initialize_all_variables()
sess = tf.Session()
sess.run(init)
# train the data
for i in range(10):
    batch_xs, batch_ys = mnist.train.next_batch(100)
    sess.run(train_step, feed_dict={x: batch_xs, y_: batch_ys})
correct_prediction = tf.equal(tf.argmax(y,1), tf.argmax(y_,1))
accuracy = tf.reduce_mean(tf.cast(correct_prediction, "float"))
[...]

PR #10978, opened 2017-06-22
[OpenCL] Provides SYCL kernels for 3D pooling (#97)

lukeiwanski:
* [OpenCL] Adds SYCL kernels for 3D pooling

Uses simple SYCL kernels to provide implementations for all 3D pooling
ops currently in use. These kernels pass the tests, but haven't really
been optimized.

These need benchmarking to compare with Eigen and CPU kernels.

* [OpenCL] Refactors SYCL kernels to use parameter struct

Moves a lot of the functor parameters into a separate data struct, with
the aim of simplifying the functor code.

* [OpenCL] Removes extra fetching of tensor dimensions

We already had the tensor dimensions passed into
LaunchMaxPooling3dGradOP, so don't need to fetch them from the
tensor.

* [OpenCL] Renames SYCL 3D pooling kernels

Adds '3D' to kernel names.

* [OpenCL] Adds 3D pooling SYCL kernel documentation

* [OpenCL] Adds guards around SYCLDevice typedef

* [OpenCL] Use forward input for SYCL MaxPool3DGradGrad

When we had a mix of SYCL and CPU kernels the forward_input would break
and cause computation problems. Now that we have SYCL kernels for all 3D
pooling operations, this is not a problem.

* [OpenCL] Reformats SYCL 3D pooling code

* [OpenCL] Moves SYCL utils into separate header

* [OpenCL] Simplifies SYCL Pool param contructors

Instead of each constructor initialising the data, simplifies the
constructors to call the first constructor.


drpngx:
Jenkins, test this please.

tensorflow/core/kernels/pooling_ops_3d.cc
```diff
@@ -837,6 +845,698 @@ TF_CALL_float(REGISTER_GPU_KERNELS) TF_CALL_half(REGISTER_GPU_KERNELS)
 
 #endif  // GOOGLE_CUDA
 
+#ifdef TENSORFLOW_USE_SYCL
```
drpngx: This file is starting to get long. I wonder if we should split it. @benoitsteiner WDYT?

tensorflow/core/util/sycl_util.h
```diff
@@ -0,0 +1,98 @@
...
+// Use the OpenCL atomic uint operations to provide a floating point atomic add.
+// For the device we use the atomic compare-exchange builtin to keep trying to
+// add to the memory in a thread safe way. The union is needed as these
+// builtins are not availble for floating point types, only integer types, so
+// we do the addition on the float and the memory update on the uint.
+//
+// TODO(jwlawson): Remove once we have different type accessors for SYCL buffers
+// Providing a way to cast the types of buffers or accessors has been proposed
+// as a SYCL extension, so once this is available we can use an atomic
+// accessor and remove this.
```
benoitsteiner: The sycl spec states that "The cl::sycl::atomic<T> must be used for elements of a buffer to provide safe atomic access to the buffer from device code." This seems to violate the spec. Why not use cl::sycl::atomic<T> in the max pooling code instead ?

benoitsteiner (changes requested):

tensorflow/core/util/sycl_util.h
DuncanMcBain: Hi Benoit,
unfortunately there is no way to access elements of an accessor atomically unless all elements are atomic, which is expressed by using a cl::sycl::accessor<T, dims, access::mode::atomic>. Because of this restriction, John has instead used the OpenCL function directly, which is exactly what the SYCL implementation should be doing. It is perfectly valid to call OpenCL functions from SYCL code provided that the host compiler doesn't see them as they will be undefined, which I am told is the case here. cl::sycl::atomic is not user-constructible, and so there's no way to use that directly. Instead it is easiest to use the SPIR-mangled function as John has here.

lukeiwanski:
@mehdi-goli should be now in CLA

tensorflow/core/util/sycl_util.h
benoitsteiner: Access to the underlying buffers is requested as follow:
  auto input_access = input_buffer.template get_access<cl::sycl::access::mode::read>(cgh);
  auto output_access = output_buffer.template get_access<cl::sycl::access::mode::write>(cgh);
Can you explain why this can't be modifed into:
  auto input_access = input_buffer.template get_access<cl::sycl::access::mode::atomic>(cgh);
  auto output_access = output_buffer.template get_access<cl::sycl::access::mode::atomic>(cgh);

tensorflow/core/kernels/pooling_ops_3d.cc
benoitsteiner: It would make sense to split the file.

lukeiwanski: How about a separate PR for SYCL3DPoolParams?

tensorflow/core/util/sycl_util.h
DuncanMcBain: In particular, you can't turn those into atomic accessors because there is no atomic add for floats in OpenCL 1.2. Even OpenCL 2.0, which expanded the atomic operations available, doesn't have them.

In addition, the underlying buffer type is of int8_t, which has no atomic support either. Similarly, the cl::sycl::atomic type is not user-constructible, but even if it were, it would effectively be doing this anyway (or rather would call this SPIR-mangled function on device, and the C++11 version on host). That said, we'll take note of this feedback - seeing real-world usage is always useful.

Personally, I'm more of a fan of the less-abstract form of atomics where you simply operate atomically on addresses. though that's somewhat lower-level than SYCL aims for. I liked the atomic_view (later atomic_ref) proposals for C++, but I don't believe they made it into C++17.

lukeiwanski:
Closing. We will open new PR with implementation that doesn't use atomics.

tensorflow/core/util/sycl_util.h
keryell: @DuncanMcBain this is useful feedback you should send to SYCL committee about what would be helpful to add for this.

lukeiwanski:
@benoitsteiner could you have a look at this updated version that does not contain atomics

vrv:
ping for @benoitsteiner

benoitsteiner (changes requested):
Please move the code to a new file as previously requested to keep it manageable.

lukeiwanski:
@benoitsteiner is that better?

tensorflow/core/kernels/pooling_ops_3d.cc
lukeiwanski: Fixed in 4e2e36b

benoitsteiner:
Jenkins, test this please.

benoitsteiner (approved):

PR #10980, opened 2017-06-22
python_configure.bzl: Add python import library on Windows

meteorcloudy:
When building a Python C/C++ extension on Windows, we need to link to the python [import library](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682592(v=vs.85).aspx) pythonXY.lib (eg. `C:\Program Files\Anaconda3\libs\python35.lib`)
See https://docs.python.org/3/extending/windows.html

Previously, TensorFlow relies on Bazel [adding ](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/windows_cc_configure.bzl#L324)the python lib path `C:\Program Files\Anaconda3\libs` into `LIB` environment variable, the build will break after Bazel stop doing so.

To solve this problem, we generate a python_import_lib rule. Ideally, it should be a `cc_library`, but when building a dll, Bazel add /WHOLEARCHIVE to `python35.lib` because it cannot tell it's a static library or import library(they have the same extension `.lib`). This will cause a linking error. As a workaround, we put it into `linkopts` if `python_headers`.
See more details at https://github.com/bazelbuild/bazel/issues/3237

### DIFF
--- a/third_party/gpus/cuda_configure.bzl
+++ b/third_party/gpus/cuda_configure.bzl
@@ -847,10 +847,10 @@ def _genrule(src_dir, genrule_name, command, outs):
       genrule_name + '",\n' +
       '    outs = [\n' +
       outs +
-      '    ],\n' +
+      '\n    ],\n' +
       '    cmd = """\n' +
       command +
-      '    """,\n' +
+      '\n   """,\n' +
       ')\n\n'
   )
 
--- a/third_party/py/BUILD.tpl
+++ b/third_party/py/BUILD.tpl
@@ -6,6 +6,16 @@ cc_library(
     name = "python_headers",
     hdrs = [":python_include"],
     includes = ["python_include"],
+    data = select({
+      ":windows" : [":python_import_lib"],
+      "//conditions:default": [],
+    }),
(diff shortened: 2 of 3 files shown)


meteorcloudy:
http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/27/console

meteorcloudy:
It's broken all the tests, let me investigate...

gunan:
+CC @nlopezgi to make sure we are not undoing any of his changes.

meteorcloudy:
Turns out there are other lib files under `C:\Program Files\Anaconda3\libs` that cannot be passed to linker, I've updated the change to only use `pythonXY.lib` as output of the genrule.
Testing at http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/29/console

meteorcloudy:
@nlopezgi Does this change look good to you?

gunan (approved):
The changes look safe to me. Let's move forward with this, and then look into the merge issues.

nlopezgi:
sorry for the late reply, lgtm

PR #10985, opened 2017-06-22
Update initializers.py

zhongzyd:
According to [Understanding the difficulty of training deep feedforward neural networks](http://jmlr.org/proceedings/papers/v9/glorot10a/glorot10a.pdf) and the codes, it is sqrt(2. / (in + out))

### DIFF
--- a/tensorflow/contrib/layers/python/layers/initializers.py
+++ b/tensorflow/contrib/layers/python/layers/initializers.py
@@ -42,7 +42,7 @@ def xavier_initializer(uniform=True, seed=None, dtype=dtypes.float32):
   This initializer is designed to keep the scale of the gradients roughly the
   same in all layers. In uniform distribution this ends up being the range:
   `x = sqrt(6. / (in + out)); [-x, x]` and for normal distribution a standard
-  deviation of `sqrt(3. / (in + out))` is used.
+  deviation of `sqrt(2. / (in + out))` is used.
 
   Args:
     uniform: Whether to use uniform or normal distributed random initialization.


drpngx:
@sguada what do you think?

sguada:
Not clear to me the paper mentioned that.

zhongzyd:
@sguada The paper mentioned that in (eq.12). 
The variance of the weights is  2. / (in + out) , so the standard deviation should be sqrt(2. / (in + out)).

frankchn:
ping @sguada

sguada:
The normalized initialization is defined in eq. (16) 
`sqrt(6) / sqrt(n_in + n_out)`
which equivalent to 
`sqrt(3 / ((n_in + n_out)/2))`

zhongzyd:
@sguada 
Eq. (16) uses uniform distribution.  
If **W** is sampled from a uniform distribution in the range **[-sqrt(6) / sqrt(n_in + n_out), sqrt(6) / sqrt(n_in + n_out)]**, then it has a standard deviation of **sqrt(2) / sqrt(n_in + n_out)**.
So when **W** is sampled from a **normal distribution**, the standard deviation should be **sqrt(2) / sqrt(n_in + n_out)**. 

You could check the code in **https://github.com/tensorflow/tensorflow/blob/r1.2/tensorflow/contrib/layers/python/layers/initializers.py** .
If **mode == 'FAN_AVG'**, it uses a standard deviation of **sqrt(2) / sqrt(n_in + n_out)** for normal distribution. Because it is sampled from a **truncated normal distribution**, so it multiply the factor with 1.3 to keep the standard deviation at **sqrt(2) / sqrt(n_in + n_out)**.

sguada:
@zhongzyd Thanks for the clarification, I miss understood your previous comment.

sguada (commented):
Looks good

sguada (approved):

zhongzyd:
@sguada Thanks.

frankchn:
Jenkins, test this please.

Issue #56, opened 2015-11-10
pip error: No such file or directory: '/tmp/pip-...-build/setup.py'


markusdr:
I'm getting the following error when trying to install on a GPU desktop using pip as described in the [tensorflow setup page](http://tensorflow.org/get_started/os_setup.md):

    $ sudo pip install https://storage.googleapis.com/tensorflow/linux/gpu/tensorflow-0.5.0-cp27-none-linux_x86_64.whl
    Downloading/unpacking https://storage.googleapis.com/tensorflow/linux/gpu/tensorflow-0.5.0-cp27-none-linux_x86_64.whl
      Downloading tensorflow-0.5.0-cp27-none-linux_x86_64.whl (50.5Mb): 50.5Mb downloaded
      Running setup.py egg_info for package from https://storage.googleapis.com/tensorflow/linux/gpu/tensorflow-0.5.0-cp27-none-linux_x86_64.whl
        Traceback (most recent call last):
          File "<string>", line 14, in <module>
        IOError: [Errno 2] No such file or directory: '/tmp/pip-kYFqUa-build/setup.py'
        Complete output from command python setup.py egg_info:
        Traceback (most recent call last):
      File "<string>", line 14, in <module>
    IOError: [Errno 2] No such file or directory: '/tmp/pip-kYFqUa-build/setup.py'

See [this Stackoverflow page](http://stackoverflow.com/questions/2204811/pip-install-a-python-package-without-a-setup-py-file) that discusses the problem and indicates that a `setup.py` file is missing in the package.

markusdr:
Upgrading pip to the latest version has solved this issue.

vrv:
@markusdr: do you happen to know which version you had and which version you upgraded to?  We'd love to update our docs to indicate a minimum pip version.

markusdr:
It was pip 1.0 on Ubuntu 12.04. I upgraded to 7.1.2 and it worked.

mrob27:
The command to type is "pip install --upgrade pip", and this should be added to the [http://tensorflow.org/get_started/os_setup.md](instructions) right after where they tell the user to "source bin/activate"

fancyerii:
after pip install --upgrade pip
it still not work. I am using ubuntu 12.04 and pip version is 1.0

nishasraj:
me too after pip upgrade its not working

nishasraj:
how u corrected the error..my pip got upgraded correctly

aolang:
[...]

PR #10988, opened 2017-06-22
[OpenCL] Fixes CUDA specific test run on SYCL (#56)

lukeiwanski:
The testBadParentValuesOnGPU should only be run on CUDA devices, as the
test checks for particular CUDA behaviour. We don't actually provide a
SYCL kernel for GatherTree and so it's not a problem that the tests
don't target SYCL.


benoitsteiner (approved):

yifeif:

PR #10995, opened 2017-06-22
Use native compute capabilities as default

zasdfgbnm:
Instead of using 3.5 and 5.2, would it be better to use compute capabilities of native GPUs?

### DIFF
--- a/configure
+++ b/configure
@@ -658,8 +658,8 @@ write_action_env_to_bazelrc "TF_CUDNN_VERSION" "$TF_CUDNN_VERSION"
 # Since Cuda toolkit is not backward-compatible, this is not guaranteed to work.
 function get_native_cuda_compute_capabilities {
   device_query_bin="$CUDA_TOOLKIT_PATH/extras/demo_suite/deviceQuery"
-  device_query_bin_windows="${device_query_bin}.exe"
-  (is_windows && "$device_query_bin_windows" || "$device_query_bin") | grep 'Capability' | grep -o '[0-9]*\.[0-9]*' | sed ':a;{N;s/\n/,/};ba'
+  "$device_query_bin" | grep 'Capability' | grep -o '[0-9]*\.[0-9]*' | sed ':a;{N;s/\n/,/};ba'
+  exit 0
 }
 while true; do
   fromuser=""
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -656,16 +656,26 @@ write_action_env_to_bazelrc "TF_CUDNN_VERSION" "$TF_CUDNN_VERSION"
 
 # Configure the compute capabilities that TensorFlow builds for.
 # Since Cuda toolkit is not backward-compatible, this is not guaranteed to work.
+function get_native_cuda_compute_capabilities {
```
gunan: Sure, this can be helpful, but we need a script that works with linux and windows.
What if we did this on linux, and on windows we simply defaulted to 3.5 and 5.2?
On windows, this script runs under msys

zasdfgbnm: Yes, you are right. And I think this is done automatically (Not tested on Windows yet. I don't have a Windows machine available now for me to test).

In case of errors, for example if on windows, or the samples are not installed with cuda, etc. The function `get_native_cuda_compute_capabilities` will exit without outputting anything (and the configure script will keep running). In this case, `native_cuda_compute_capabilities` will be empty and
```bash
default_cuda_compute_capabilities=${native_cuda_compute_capabilities:-"3.5,5.2"}
```

will automatically gives `3.5,5.2`

gunan:
@meteorcloudy Could you rake a look?
Jenkins, test this please.

meteorcloudy (changes requested):
This change won't break cuda configuration on Windows.
But the script won't work on Windows, either. Because there is no `samples/1_Utilities/deviceQuery/deviceQuery.cpp` on Windows.

But! There is a binary called `extras/demo_suite/deviceQuery.exe`. I believe you can use it to detect native cuda capabilities.

Try use this function on Windows:
```
function get_native_cuda_compute_capabilities {
  device_query_bin="$CUDA_TOOLKIT_PATH/extras/demo_suite/deviceQuery.exe"
  "$device_query_bin" |grep 'Capability'|grep -o '[0-9]*\.[0-9]*'|sed ':a;{N;s/\n/,/};ba'
}
```

meteorcloudy:
Just realized `$CUDA_TOOLKIT_PATH/extras/demo_suite/deviceQuery` also exists on Linux, can you use this binary directly instead of compiling it as well? Like this:
```
function get_native_cuda_compute_capabilities {
  device_query_bin="$CUDA_TOOLKIT_PATH/extras/demo_suite/deviceQuery" # Also works on Windows without .exe
  "$device_query_bin" |grep 'Capability'|grep -o '[0-9]*\.[0-9]*'|sed ':a;{N;s/\n/,/};ba'
}
```

zasdfgbnm:
not done yet

zasdfgbnm:
@meteorcloudy I think I'm done now. I test this configure on my Linux machine with two GPU and a Windows machine without any GPU under msys2. Both works well.

meteorcloudy:
I tests on a GPU machine, it also worked! Thanks!

meteorcloudy (approved):

gunan:
Jenkins, test this please.

gunan (approved):

PR #11010, opened 2017-06-23
Sort shuffled indices before indexing NumpyFeed

swook:
It is possible to use `h5py` datasets with `tf.estimator.inputs.numpy_input_fn` as there is a high level of compatibility.

However, with `shuffle=True`, one can no longer do so due to the following error:

    TypeError: Indexing elements must be in increasing order

This small change allows for one to use large HDF5 datasets via `h5py` with shuffled batch indices.

### DIFF
--- a/tensorflow/python/estimator/inputs/queues/feeding_functions.py
+++ b/tensorflow/python/estimator/inputs/queues/feeding_functions.py
@@ -171,9 +171,10 @@ class _OrderedDictNumpyFeedFn(object):
 
     # Get sorted indices and remember how to reorder output.
     # This allows for the use of h5py databases which require ordered indexing.
-    sorted_integer_indexes, reorder_indexes = zip(
-        *sorted([(value, index)
-                 for index, value in enumerate(integer_indexes)]))
+    original_indexes, sorted_integer_indexes = \
+        zip(*sorted(enumerate(integer_indexes), key=lambda x: x[1]))
+    reorder_indexes, _ = \
+        zip(*sorted(enumerate(original_indexes), key=lambda x: x[1]))
     sorted_integer_indexes = list(sorted_integer_indexes)
     reorder_indexes = list(reorder_indexes)
 
--- a/tensorflow/python/estimator/inputs/queues/feeding_functions.py
+++ b/tensorflow/python/estimator/inputs/queues/feeding_functions.py
@@ -170,6 +170,7 @@ class _OrderedDictNumpyFeedFn(object):
         total_epochs=self._num_epochs)
 
     # Get sorted indices and remember how to reorder output.
+    # This allows for the use of h5py databases which require ordered indexing.
     sorted_integer_indexes, reorder_indexes = zip(
         *sorted([(value, index)
                  for index, value in enumerate(integer_indexes)]))
--- a/tensorflow/python/estimator/inputs/queues/feeding_functions.py
+++ b/tensorflow/python/estimator/inputs/queues/feeding_functions.py
@@ -169,7 +169,7 @@ class _OrderedDictNumpyFeedFn(object):
         current_epoch=self._epoch,
         total_epochs=self._num_epochs)
 
-    # Get sorted indices and remember how to reorder output
+    # Get sorted indices and remember how to reorder output.
     sorted_integer_indexes, reorder_indexes = zip(
         *sorted([(value, index)
                  for index, value in enumerate(integer_indexes)]))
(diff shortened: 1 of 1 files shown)


swook:
I signed it!

drpngx:
Jenkins, test this please.

drpngx:
That seems to make a lot of tests fail. Could you take a look?

swook:
@drpngx I believe this now allows tests to pass.

However, I'm unsure if the tests should be adjusted instead.

drpngx:
Jenkins, test this please.

drpngx:
Now passes the tests.

@xiejw would have cycles to take a look?

frankchn:
ping @xiejw

tensorflow/python/estimator/inputs/queues/feeding_functions.py
```diff
@@ -169,10 +169,17 @@ def __call__(self):
         current_epoch=self._epoch,
         total_epochs=self._num_epochs)
 
+    # Get sorted indices and remember how to reorder output.
```
xiejw: The code looks good. But can you add a comment to briefly explain why this is required?

xiejw:
Sorry for the delay. Just a nit.

swook:
@xiejw thanks for reviewing this PR. I've added a line to explain why sorting is necessary.

xiejw (approved):

xiejw:
Thanks for the contribution!

jhseu:
The timeseries failures look related. Allen, thoughts?

allenlavoie:
Just so I'm clear, this change should ideally only be affecting the order in which the Numpy arrays are indexed, but should be returning the same values as before down the line in TensorFlow land?

I can dig into them to figure out how exactly, but I think those test failures are indicating that with shuffle=False this PR is changing the output of the op (i.e. it's no longer returned in the input order). Probably indicates a bug?

tensorflow/python/estimator/inputs/queues/feeding_functions.py
```diff
@@ -169,10 +169,18 @@ def __call__(self):
...
+        *sorted([(value, index)
+                 for index, value in enumerate(integer_indexes)]))
+    sorted_integer_indexes = list(sorted_integer_indexes)
+    reorder_indexes = list(reorder_indexes)
+
     self._trav = (integer_indexes[-1] + 1) % self._max
     feed_dict = {self._index_placeholder: integer_indexes}
     cols = [
-        column[integer_indexes]
+        column[sorted_integer_indexes][reorder_indexes]
```
xiejw: After digging into details, I think this line has a bug. You meant to put the elements into position reorder_indexes, but instead, here it is getting the elements from position reorder_indexes. I should catch this earlier. Sorry.

(I was also worrying about performance if this line does the expected assigning). 

One mini example could be:

>>> a = np.array([10, 20, 30, 40])
>>> i = [0, 1, 2, 3, 0, 1, 2, 3]                                                                                                                                                                                                                                                                                              
>>> a[i]
array([10, 20, 30, 40, 10, 20, 30, 40])
>>> sorted_integer_indexes, reorder_indexes = zip(*sorted([(value, index) for index, value in enumerate(i)]))
>>> sorted_integer_indexes = list(sorted_integer_indexes)
>>> reorder_indexes = list(reorder_indexes)
>>> reorder_indexes
[0, 4, 1, 5, 2, 6, 3, 7]
>>> a[sorted_integer_indexes][reorder_indexes]
array([10, 30, 10, 30, 20, 40, 20, 40])

xiejw (changes requested):
I should catch this bug earlier. Sorry for that.

swook:
@xiejw thanks for catching the bug out. I have committed a fix to make the tests pass. I have confirmed that tests pass on my end by rebasing on upstream.

Your concern on performance is something I agree with however...

If I understand right.. this issue occurs when wrapping around datasets to return to index 0. Perhaps it's better to split the indexing operations to allow for this.

Or dismiss this PR as it could be argued that it's a bug in `h5py`.

I leave the decision to you, the maintainers.

xiejw:
For now, I think you can convert the h5py as numpy which is one time cost. If this has more cost than expected or it is extremely inconvenient, we can revisit this PR and propose some new ideas. 

With that said, I will close this PR and please reopen if anything changed in future. 

Thanks for your contribution!

swook:
I understand. Not to re-open the issue, but I'd just like to mention that I'm using `h5py` for large input data which cannot be held in memory. While one could write their own pre-processing routine using tensorflow queues, using the numpy feeding functions with hdf5 structures is very convenient.

I attach the explanation for future reference.

PR #11024, opened 2017-06-23
Adjust doc MNIST learning rate

pvaneck:
The gradient descent learning rate specified in the MNIST beginner
tutorial should be 0.5 instead of 0.05 to stay consistent with the
actual code in the repo.

A learning rate of .05 will actually decrease accuracy.

Code in repo: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/mnist/mnist_softmax.py#L59

### DIFF
--- a/tensorflow/docs_src/get_started/mnist/beginners.md
+++ b/tensorflow/docs_src/get_started/mnist/beginners.md
@@ -362,12 +362,12 @@ minimize. Then it can apply your choice of optimization algorithm to modify the
 variables and reduce the loss.
 
 ```python
-train_step = tf.train.GradientDescentOptimizer(0.05).minimize(cross_entropy)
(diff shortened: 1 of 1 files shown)


yifeif (approved):
Nice catch, thanks!

foxish:
The documentation on the [website](https://www.tensorflow.org/get_started/mnist/beginners) still says 0.05. Some generation script needs to be run to update that?

yifeif:
Looks like the code block on the website was somehow not updated with this change, but the second instance that is not in the code block got updated. @MarkDaoust have you run into this before?

MarkDaoust:
It looks like there's an intermediate version of this page where that was the actual content, then someone updated the comment in the doc to match the code in the doc...

The root docs (that you link to) are only updated with releases.

But this fix will be in [versions/master](https://www.tensorflow.org/versions/master/get_started/mnist/beginners) next time it gets published.

PR #11030, opened 2017-06-24
test_session: fix doc formatting

siimphh:
This is a blind fix following create_local_cluster example below. See broken doc formatting here:

https://www.tensorflow.org/api_docs/python/tf/test/TestCase#test_session


caisq:

caisq (approved):

PR #11075, opened 2017-06-27
doc: update the conv1 shape.

lazyparser:
Update the conv1 shape from [5, 5, 32, 32] to [5, 5, 1, 32]. (`1` means gray image.)

Although this is irrelevant with the key content in the chapter, one might be confused when looking at this detail.

### DIFF
--- a/tensorflow/docs_src/programmers_guide/variables.md
+++ b/tensorflow/docs_src/programmers_guide/variables.md
@@ -410,7 +410,7 @@ calling this function repeatedly would not work:
 ``` python
 input1 = tf.random_normal([1,10,10,32])
 input2 = tf.random_normal([1,20,20,32])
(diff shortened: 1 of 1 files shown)


drpngx:
@MarkDaoust WDYT?

MarkDaoust:
It's reasonable for the example model to start with a 1 channel image.

But this `variable_scope.md` doc was just merged into `variables.md`, so accepting this now would just set us up for a merge conflict when the sync happens.

If you want to update `variables.md` after the sync, that's fine with me.

lazyparser:
Got it. Thank you. I'll update this PR after sync.

lazyparser:
rebased.

drpngx:
Jenkins, test this please.

drpngx (approved):
As per Mark's comment.

PR #11084, opened 2017-06-27
Fix import in documentation documentation

martinwicke:


tensorflow/docs_src/community/documentation.md
```diff
@@ -276,7 +276,7 @@ __init__.py:
     # Otherwise import symbols directly
     from tensorflow.some_module.some_other_file import some_symbol
 
-    from tensorflow.platform.all_util import remove_undocumented
+    from tensorflow.platform.util.all_util import remove_undocumented
```
MarkDaoust: This still gives a:

`ImportError: No module named platform.util.all_util`

looks like that "platform" should be replaced with "python"

`from tensorflow.python.util.all_util import remove_undocumented`

drpngx (approved):

drpngx:
Ignoring unrelated failure.

Issue #10616, opened 2017-06-10
[feature] Support S3 Filesystem for Tensorflow


yongtang:
At the moment Tensorflow supports distributed file system implementations such as HDFS or GCS which has been very useful for handling cases in a distributed environment. Both of them utilize Tensorflow's FileSystem C++ interface.

In case Tensorflow is used in AWS, it would be great to have a S3 Filesystem for Tensorflow as well.

We have a preliminary implementation of S3 Filesystem on Tensorflow, based on AWS's C++ SDK:
https://github.com/tensorflow/tensorflow/compare/master...yongtang:s3

The code is placed under the directory of tensorflow/contrib/s3.

As AWS is widely used, I am just wondering if it make sense to have a pull request to add S3 File system on Tensorflow?

It will be better to have feedback from community so that we could improve our implementation and help those deploying Tensorflow on AWS.

aselle:
@jhseu, it would be great if you could comment on this and provide guidance. @yongtang prepared a patch linked above.

jhseu:
@yongtang Yes, please make a pull request with your changes! This would be a very useful feature for a lot of people.

yongtang:
Closed through #11089

PR #11089, opened 2017-06-27
Support S3 Filesystem for Tensorflow

yongtang:
This is for S3 Filesystem support in Tensorflow. 

It utilizes Tensorflow's FileSystem C++ interface, similar to currently available HDFS or GCS support. The code depends on AWS's C++ SDK (Apache License) https://github.com/aws/aws-sdk-cpp
The code is placed under the directory of tensorflow/contrib/s3

TODO List:

- [X] Add the implementation of S3 file system
- [X] Work out the BUILD configuration (currently build a .so, not sure the best way in bazel)
- [X] Add tests similar to `gcs_file_system_test.cc`

This fix is related to #10616


drpngx:
@jhseu FYI

sswv:
@yongtang Thanks for your contribution! Now the code can only connect to AWS S3. Is it right? One suggestion: it is better to make it compatible with other 3rd-party S3 implementations (e.g. Eucalyptus Walrus, Scality S3 server, Minio). I think it can be done by setting `ClientConfiguration::endpointOverride` and `ClientConfiguration::scheme`.  Some custom options may be added.

yongtang:
Thanks @sswv. I will take a look at the configurations you mentioned.

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,546 @@
...
+
+  Status Sync() override {
+    if (!outfile_) {
+      return errors::FailedPrecondition(
+          "The internal temporary file is not writable.");
+    }
+    if (!sync_needed_) {
+      return Status::OK();
+    }
+    Aws::Client::ClientConfiguration clientConfig;
```
drpngx: We need to be able to change region etc

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,546 @@
...
+  Status GetChildren(const string& dir, std::vector<string>* result) override {
+    string bucket, prefix;
+    TF_RETURN_IF_ERROR(ParseS3Path(dir, false, &bucket, &prefix));
+
+    if (prefix.back() != '/') {
+      prefix.push_back('/');
+    }
+
+    Aws::S3::S3Client s3Client;
+    Aws::S3::Model::ListObjectsV2Request listObjectsV2Request;
```
sswv: I noticed that you used `listObjectV2`. In some compatible S3 implementations from 3rd-party, the new V2 interface is not supported. I think it is better to provide an option for using `listObject` (V1). With the V1 inferface, TensorFlow-S3 can work with more 3rd-party S3 implementations.

I tried to remove all `V2`, and change `ContinuationToken` into `Marker`. It seems working well.

Thanks!

sswv:
Hi, @yongtang,

I wrote some small patches to allow running the S3 component with 3rd-party S3 implementations, and I also fixed a potential bug. Please see [my branch](https://github.com/sswv/tensorflow/commits/s3-patch). Note that these patches are all simple but not mature solutions, so I did not make them as pull requests. You can refer to them or pick up code from them, and provide more complete solutions based on my patches.

* [Patch 1](https://github.com/sswv/tensorflow/commit/d3eb70e56fd1c590245414f628e168168339ec56):

Get S3 configurations from environment variables to support 3rd-party S3 implementations. Do you think it is a good way to get configurations from environment variables? If not, we can choose a better way. More configuration items of `Aws::Client::ClientConfiguration` might be added.

* [Patch 2](https://github.com/sswv/tensorflow/commit/a89294358662d0f82a1f51530f016ae3e353b18e):

Use old `ListObject` (V1) instead of `ListObjectV2` to support some 3rd-party S3 implementations, especially legacy systems. A better way is adding a switch to allow users choosing from V1 and V2.

* [Patch 3](https://github.com/sswv/tensorflow/commit/c3f28c9cd233ecf1923dfaa435b069acf8937dc1):

A workaround for the "Invalid Range Error" when reading ImageNet data set. When I train the `inception_v3` network from `tensorflow/models/slim` by using the ImageNet data set, I see an "Invalid Range Error" after 120 steps and the program crashes immediately. It happens only with the S3 file system, but not happen with local file system and HDFS. I can reproduce it with 2 different kinds of 3rd-party S3 implementations. After some debugging, I found it is because of `offset == size of file` in `S3RandomAccessFile::Read`. In this case, `s3Client.GetObject` will report the "Invalid Range Error".

Refer to the `HDFSRandomAccessFile::Read` function in `hadoop_file_system.cc`. It has a more elegant behavior when out-of-range error happens. For the HDFS case, A non-fatal `error::OUT_OF_RANGE` status will be returned to `TFRecordReader::ReadLocked` when out-of-range error happens, and the program will not crash.

Thus, I made a workaround patch. It works well in my simple tests. However, I am not a S3 expert, so I am not sure whether my solution is correct and good enough. I think you can find a more reasonable way to fix the issue.

jhseu:
@yongtang please let me know when this pull request is ready for review.

vrv:
@yongtang is this ready to review?  Also, can you fix the CLA issue?  Probably by fixing your git commit email on some of the newer commits?

vrv:
Oh, @sswv you probably need to sign the CLA for us to accept these patches.

sswv:
@vrv I had signed CLA. I think the check will pass when the bot run next time.

vrv:
Did the email you used for CLA match the email used in https://patch-diff.githubusercontent.com/raw/tensorflow/tensorflow/pull/11089.patch ?  The bot checks the CLA every time a new commit is added.

sswv:
@vrv Yes. I checked the patch and the CLA page. I am using the same email address in the two places.

TK-blost:
I try this branch but when setup there is an error:
tensorflow.python.framework.errors_impl.UnimplementedError: File system scheme s3 not implemented
I install this as these steps:
1.  git clone
2. ./configure
3. bazel build 
4. bazel-bin/**
5. pip install 

after that i find s3_file_system.so  is not exist? is that right ? 
any help will be vary thanks.

drpngx:
You may have to answer some question in the configure part that enables it

TK-blost:
@drpngx  thank you for your attention,  I check again not find any option about s3 configuration?

sswv:
@TK-blost Did you add `tf.load_file_system_library('/path/to/XXX_file_system.so')` in your code? Refer to: <https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/extend/add_filesys.md#registering-and-loading-the-filesystem>

TK-blost:
@sswv  wow it worked as you said.

neuzxy:
Thanks for your contribution! I'm very interested in the S3 feature and I hope it can release soon. I notice that there is no Unit Test for the S3 feature. I think I can contribute a group of Unit Tests. Please let me know whether it is OK. @yongtang

yongtang:
@neuzxy Thanks. That would be great! You can add commits on top of the PR.

neuzxy:
Hi, @yongtang ,

I wrote some small patches to do unit test for S3 file system, and I also found and fixed a bug through unit test. Please see [my branch](https://github.com/neuzxy/tensorflow/tree/s3-ut). These patches are simple, you can refer them or pick up useful code.

- [Patch 1](https://github.com/neuzxy/tensorflow/commit/d4a297518c8027f3b365f5e3af2a759b8e48eacb)

Split the header and source of s3_file_system.cc. To test s3_file_system.cc, I should create a S3FileSystem Object, but class S3FileSystem was defined in s3_file_system.cc. So I splited the header and source of s3_file_system.cc. Due to the indent problem, you may see the diff is confusion, you may infer [`this patch`](https://gist.github.com/neuzxy/0ea925c053d25d54b4b722ab9d764616) to see the change in an easy way.

- [Patch 2](https://github.com/neuzxy/tensorflow/commit/7b35e561d45727056eb7af11e6ae94cc04792ed6)

Fix a return status issue in RandomAccessFile::Read. When I tested the `NewAppendableFile` method and couldn't pass the test. I have contacted to the code contributor @sswv , he has confirmed the bug.

- [Patch 3](https://github.com/neuzxy/tensorflow/commit/d4a297518c8027f3b365f5e3af2a759b8e48eacb)

Add the unit test for S3 file system. I have created unit tests for all methods of class S3FileSystem and passed all tests in my environment. Before the test, you should configure `TFS3_ENDPOINT`, `TFS3_REGION`, `TFS3_USE_HTTPS`, `TFS3_VERIFY_SSL` and the tmp dir `S3_TMPDIR`. If you execute the test for more than once, you should empty `S3_TMPDIR` first.

```
bazel test \
    //tensorflow/contrib/s3:s3_file_system_test \
    --test_env=TFS3_ENDPOINT=$TFS3_ENDPOINT \
    --test_env=TFS3_REGION=$STFS3_REGION \
    --test_env=TFS3_USE_HTTPS=$TFS3_USE_HTTPS \
    --test_env=TFS3_VERIFY_SSL=$TFS3_VERIFY_SSL \
    --test_env=S3_TMPDIR=S3://test 
```

It works well in my simple tests. However, I am a junior engineer, I wrote the unit test by referring hadoop_file_system_test.cc and gcs_file_system_test.cc. It would be great if the author or other contributors provide more solid unit test.

sswv:
Thanks, @neuzxy . As we discussed in email, I agree with your Patch 2. Returning a non-fatal `error::OUT_OF_RANGE` in this case is similar to the implementation of HDFS plugin. I tested your patch with the ImageNet dataset in local Minio storage, and it worked well.

BTW, I noticed the @googlebot said: *"We need to confirm that they're okay with their commits being contributed to this project."* It is good for me that my commits being contributed. Thanks!

yongtang:
@neuzxy Thanks. You could actually create a pull request to the branch in my forked repo: "yongtang:s3". In that way once it is merged in my forked repo then your commits will atomically update this PR and show up. I just did that for you.

yongtang:
Ping @jhseu I think the PR is ready for review. Please take a look.

drpngx:
Jenkins, test this please

yongtang:
@drpngx The Jenkins build failure was caused by the change from other PR's `@protobuf//` -> `@protobuf_archive//`. I rebased and updated the PR. Now it the Jenkins should work.

drpngx:
Did you push?

yongtang:
@drpngx Yes the change has been pushed to fix `@protobuf` -> `@protobuf_archive` for bazel.

drpngx:
Jenkins, test this please.

rmlarsen:

jhseu:
@sb2nov @frankchn @saeta FYI

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+#include <mutex>
+
+namespace tensorflow {
+
+static const char* S3FileSystemAllocationTag = "S3FileSystemAllocation";
+static const size_t S3ReadAppendableFileBufferSize = 1024 * 1024;
+static const int S3GetChildrenMaxKeys = 100;
+
+Aws::Client::ClientConfiguration& GetDefaultClientConfig() {
+  static std::mutex cfg_lock;
```
jhseu: Can you switch to the tensorflow mutex? tensorflow/core/platform/mutex.h

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+  static bool init(false);
+  static Aws::Client::ClientConfiguration cfg;
+
+  std::lock_guard<std::mutex> lock(cfg_lock);
+
+  if (!init) {
+    const char* endpoint = getenv("TFS3_ENDPOINT");
```
jhseu: Here and for the other environment variables: can we use just S3_ENDPOINT to match the conventions in tensorflow/core/platform/cloud?

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+ public:
+  S3RandomAccessFile(const string& bucket, const string& object)
+      : bucket_(bucket), object_(object) {}
+
+  Status Read(uint64 offset, size_t n, StringPiece* result,
+              char* scratch) const override {
+    Aws::S3::S3Client s3Client(GetDefaultClientConfig());
+    Aws::S3::Model::GetObjectRequest getObjectRequest;
+    getObjectRequest.WithBucket(bucket_.c_str()).WithKey(object_.c_str());
+    char buffer[50];
```
jhseu: Use a string instead, and convert to Aws::String below? Many implementations will inline small strings on the stack.

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+    memset(buffer, 0x00, sizeof(buffer));
+    snprintf(buffer, sizeof(buffer) - 1, "bytes=%lld-%lld", offset,
+             offset + n - 1);
+    getObjectRequest.SetRange(buffer);
+    getObjectRequest.SetResponseStreamFactory([]() {
+      return Aws::New<Aws::StringStream>(S3FileSystemAllocationTag);
+    });
+    auto getObjectOutcome = s3Client.GetObject(getObjectRequest);
+    if (!getObjectOutcome.IsSuccess()) {
+      n = 0;
```
jhseu: Does this work for short reads? We still want to return the bytes read even with OUT_OF_RANGE.

For example, if we requested 10 bytes, but only read 5 of them, then we'll return with OUT_OF_RANGE but return the 5 bytes.

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+    long offset = outfile_->tellp();
+    outfile_->seekg(0);
+    putObjectRequest.SetBody(outfile_);
+    putObjectRequest.SetContentLength(offset);
+    auto putObjectOutcome = s3Client.PutObject(putObjectRequest);
+    outfile_->clear();
+    outfile_->seekp(offset);
+    if (!putObjectOutcome.IsSuccess()) {
+      std::stringstream ss;
+      ss << putObjectOutcome.GetError().GetExceptionName() << ": "
```
jhseu: Use strings::StrCat instead

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+#include <aws/s3/model/ListObjectsRequest.h>
+#include <aws/s3/model/PutObjectRequest.h>
+
+#include <cstdlib>
+#include <mutex>
+
+namespace tensorflow {
+
+static const char* S3FileSystemAllocationTag = "S3FileSystemAllocation";
+static const size_t S3ReadAppendableFileBufferSize = 1024 * 1024;
```
jhseu: Rename to kS3ReadAppendableFileBufferSize

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+static const int S3GetChildrenMaxKeys = 100;
```
jhseu: kS3GetChildrenMaxKeys

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+#include <aws/s3/model/HeadObjectRequest.h>
+#include <aws/s3/model/ListObjectsRequest.h>
+#include <aws/s3/model/PutObjectRequest.h>
+
+#include <cstdlib>
+#include <mutex>
+
+namespace tensorflow {
+
+static const char* S3FileSystemAllocationTag = "S3FileSystemAllocation";
```
jhseu: kS3FileSystemAllocationTag

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+      .WithMaxKeys(S3GetChildrenMaxKeys)
+      .WithDelimiter("/");
+  listObjectsRequest.SetResponseStreamFactory(
+      []() { return Aws::New<Aws::StringStream>(S3FileSystemAllocationTag); });
+
+  Aws::S3::Model::ListObjectsResult listObjectsResult;
+  do {
+    auto listObjectsOutcome = s3Client.ListObjects(listObjectsRequest);
+    if (!listObjectsOutcome.IsSuccess()) {
+      std::stringstream ss;
```
jhseu: Use strings::StrCat

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+  TF_RETURN_IF_ERROR(ParseS3Path(fname, true, &bucket, &object));
+
+  Aws::S3::S3Client s3Client(GetDefaultClientConfig());
+  if (object.empty()) {
+    Aws::S3::Model::HeadBucketRequest headBucketRequest;
+    headBucketRequest.WithBucket(bucket.c_str());
+    auto headBucketOutcome = s3Client.HeadBucket(headBucketRequest);
+    if (!headBucketOutcome.IsSuccess()) {
+      std::stringstream ss;
+      ss << headBucketOutcome.GetError().GetExceptionName() << ": "
```
jhseu: strings::StrCat

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+  string bucket, object;
+  TF_RETURN_IF_ERROR(ParseS3Path(fname, false, &bucket, &object));
+
+  Aws::S3::S3Client s3Client(GetDefaultClientConfig());
+  Aws::S3::Model::DeleteObjectRequest deleteObjectRequest;
+  deleteObjectRequest.WithBucket(bucket.c_str()).WithKey(object.c_str());
+
+  auto deleteObjectOutcome = s3Client.DeleteObject(deleteObjectRequest);
+  if (!deleteObjectOutcome.IsSuccess()) {
+    std::stringstream ss;
```
jhseu: strings::StrCat

tensorflow/contrib/s3/s3_file_system.cc
```diff
@@ -0,0 +1,578 @@
...
+      target_object.pop_back();
+    }
+  }
+
+  Aws::S3::S3Client s3Client(GetDefaultClientConfig());
+
+  Aws::S3::Model::CopyObjectRequest copyObjectRequest;
+  Aws::S3::Model::DeleteObjectRequest deleteObjectRequest;
+
+  Aws::S3::Model::ListObjectsRequest listObjectsRequest;
```
jhseu: Not sure I understand this part: RenameFile is only supposed to rename a single file and not be made to rename directories. Shouldn't it just be a single CopyObjectRequest and DeleteObjectRequest?

jhseu (changes requested):
Thanks! Looks good. Mostly minor changes requested.

Please run buildifier on the BUILD files to fix the sanity check.

jhseu:
@neuzxy can you confirm here that you're ok with this pull request being committed? Thanks!

neuzxy:
@jhseu Yes, I'm ok with this pull request.

tensorflow/contrib/s3/s3_file_system.cc
yongtang: Thanks @jhseu  In the GCS file systems the RenameFile also does the directory rename:
https://github.com/tensorflow/tensorflow/blob/0440c77935915fce0d9286ce5e7407b31c6f4975/tensorflow/core/platform/cloud/gcs_file_system.cc#L1154-L1168

yongtang:
Thanks @jhseu for the review. The PR has been rebased and updated. Please take a look.

jhseu (approved):

jhseu:
Jenkins, test this please

martinwicke:
Jenkins, test this please.

yongtang:
@martinwicke The Jenkins build error on Mac OS X was caused by no target specification on Bazel aws.BUILD file. I have updated the PR and the issue should have been fixed now. Sorry for the inconvenience.

martinwicke:
No problem at all, thank you!

Jenkins, test this please.

martinwicke:
Re-running for infra failures.

Jenkins, test this please.

drpngx:
Trying again.

Jenkins, test this please.

drpngx:
It failed the sanity check:

```
ERROR: /workspace/tensorflow/contrib/s3/BUILD:39:1: no such package '@aws//': java.io.IOException: Error downloading [http://bazel-mirror.storage.googleapis.com/github.com/aws/aws-sdk-cpp/archive/1.0.90.tar.gz, https://github.com/aws/aws-sdk-cpp/archive/1.0.90.tar.gz] to /var/lib/jenkins/workspace/tensorflow-pull-requests-sanity/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/aws/1.0.90.tar.gz: Checksum was f599b57aec4f03ad696044dd430b2d201864113937353adc346f53ad47991319 but wanted ffa501fa00e18d789cd10c768b55a2907c314816f09bcaf96e6d816a5b67f6e7 and referenced by '//tensorflow/contrib/s3:s3_crypto'.
ERROR: /workspace/tensorflow/contrib/s3/BUILD:56:1: no such package '@aws//': java.io.IOException: Error downloading [http://bazel-mirror.storage.googleapis.com/github.com/aws/aws-sdk-cpp/archive/1.0.90.tar.gz, https://github.com/aws/aws-sdk-cpp/archive/1.0.90.tar.gz] to /var/lib/jenkins/workspace/tensorflow-pull-requests-sanity/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/aws/1.0.90.tar.gz: Checksum was f599b57aec4f03ad696044dd430b2d201864113937353adc346f53ad47991319 but wanted ffa501fa00e18d789cd10c768b55a2907c314816f09bcaf96e6d816a5b67f6e7 and referenced by '//tensorflow/contrib/s3:s3_file_system'.
ERROR: Analysis of target '//tensorflow/contrib/s3:s3_file_system' failed; build aborted.
```

yongtang:
@drpngx The error was caused by hash mismatch, likely the same issue encounter related to GitHub changing the download content recetly. I have updated the PR. Please take a look.

drpngx:
Would you mind pull rebase and push agaian? It is confusing the CLA bot.

yongtang:
@drpngx I think all commits have signed CLA. The `cla: no` flag is likely caused by the fact that this PR consists of commits from @sswv and @neuzxy.

Both of them have confirmed (in previous comments) they are OK with the commits in the PR.

drpngx:
Jenkins, test this please.

It used to be that we could override the CLA bot, but it looks like this no longer works.

yongtang:
The error is
```
missing input file '@androidsdk//:build-tools/25.0.2/aapt'.
```

Not sure where it comes from, though I assume a rerun might be able to fix it?

drpngx:
Oh, that's a transient failure.

Jenkins, test this please.

drpngx:
@sswv and @neuzxy could you confirm that the email address is correct on the commits? Thanks.

sswv:
@drpngx I checked the git log and confirmed that my email is correct. It is the same as the email I used for CLA.

neuzxy:
@drpngx I also checked the git log and confirmed that my email address is correct. I signed CLA with the same email.

drpngx:
Thanks guys for checking!

drpngx:
@caisq it's stuck on the CLA. If you know how to override this, LMK, otherwise I can look it up.

drpngx:
Talked to @gunan, he will change the config.

gunan:
Merging the PR.

yongtang:
Thanks all for the help! 👍 🍺 💥

sandyskies:
does this support Python AI tf.train.string_input_producer ?

DjangoPeng:
@sswv @drpngx What released version of TensorFlow would includes the S3 feature? And I also have an  interest in the supports of  `tf.train.string_input_producer`.

martinwicke:
The next version will include it: 1.4. 

`string_input_producer` just produces strings, if you make it produce strings with a "s3://" prefix it will now use S3.

DjangoPeng:
@martinwicke Got it. Thanks for your reply!

sandyskies:
@martinwicke   @yongtang  I have complied tensorflow in master branch, and I use string_input_producer with a "s3://" prefix ,but It won't woker. Also ,how do I provide access key and secret key for this API .

yongtang:
@sandyskies To build s3 you could invoke `./configure` then:
```
bazel build -s --config=opt --verbose_failures //tensorflow/contrib/s3:s3_file_system.so
```
After that, you should see an s3_file_system.so in the binary directory `bazel-bin/tensorflow/contrib/s3/s3_file_system.so`.

To use S3 plugin with tensorflow, you will need to load the generated dynamic .so file:
```python
import tensorflow as tf
tf.load_file_system_library('/path/to/s3_file_system.so')

with gfile.Open("s3://path/to/file/in/s3.txt") as w:
  w.write("hi") 
```

You could config the access and secret key the same way you config for you system and it will be picked up automically.

caisq:
@martinwicke @yifeif Can we consider adding a question to https://github.com/tensorflow/tensorflow/blob/master/configure.py to enable s3 support at build time, so there is no need to build and load the .so file separately.

gunan:
We are working on enabling s3 support by default.

sandyskies:
@yongtang  Can't this patch do just like hdfs， just use filename_queue = tf.train.string_input_producer(["s3://path/to/file/in/s3.txt"]) ?

martinwicke:
@gunan @case540 I guess since we'll enable it by default it would make sense to simply move it to core.

sandyskies:
import tensorflow as tf
>>> tf.load_file_system_library('/data/tensorflow/bazel-bin/tensorflow/contrib/s3/libs3_file_system.so')
  File "/data/anaconda3/lib/python3.6/site-packages/tensorflow/python/framework/load_library.py", line 100, in load_file_system_library
    lib_handle = py_tf.TF_LoadLibrary(library_filename, status)
  File "/data/anaconda3/lib/python3.6/contextlib.py", line 89, in __exit__
    next(self.gen)
  File "/data/anaconda3/lib/python3.6/site-packages/tensorflow/python/framework/errors_impl.py", line 467, in raise_exception_on_not_ok_status
    c_api.TF_GetCode(status.status))
tensorflow.python.framework.errors_impl.NotFoundError: /data/tensorflow/bazel-bin/tensorflow/contrib/s3/libs3_file_system.so: undefined symbol: _ZTVN3Aws2S35Model16GetObjectRequestE

sandyskies:
@yongtang @martinwicke can u look at the problem?

case540:
I have an internal CL to fix this. I think you can just change s3_file_system.so from a cc_binary target to a tf_cc_binary target (this fixed it for me)

drpngx:
Thank you @case540! Please submit and reference this `#11089` in the public description.

yongtang:
Thanks @case540 a lot!

@sandyskies Sorry for the late reply. I am out of the town for the past week and couldn't get a quick fix done. Will try to help if there are any further issues once I am back.

sandyskies:
@case540   Can u show me how to change the BUILD file?

sandyskies:
I have use the workaround that @case540 provide, and it works fine for me. Should I create a pr for it , or wait for @yongtang to create himself.

quantumlicht:
If I want to use this feature (I'd really want to try it out in tensorflow-serving), Do I need to compile tensorflow with special flags? I am not sure how I need to configure it

I tried with settings these environment variables,
ENV S3_REGION=us-east-2
ENV S3_VERIFY_SSL=0
ENV S3_USE_HTTPS=1

The tensorflow serving uses `Env::Default()->FileExists` from tensorflow/core/platform/env.h for checking if a file exists. From what I understand, The s3 filesystem is registered with
`REGISTER_FILE_SYSTEM("s3", S3FileSystem);`, but I am not sure if it's actually working because the error gets swallowed in the tf-serving code ( this error https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/env.cc#L91 gets swallowed herehttps://github.com/tensorflow/serving/blob/master/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc#L221).

So what I am asking is: Do I need to pass additional options at build to
a - be sure the s3 filesystem is usable
b - get additional debug information to know

yongtang:
@sandyskies I think @case540 will create an internal PR for the fix. The internal PR will show up in GitHub once it is merged internally and merged back to public GitHub.

yongtang:
@quantumlicht I would assume you may get more help from StackOverflow as there are more users there. Beside this PR has been merged so less attention here.

At the moment s3 support is not part of the 1.3 release so you could only build from the master branch of the TensorFlow.

See comment: https://github.com/tensorflow/tensorflow/pull/11089#issuecomment-332589113 for steps of building and using s3 with TensorFlow. 

@case540 has been generous enough to provide a fix so that loading `s3_file_system.so` is not needed anymore. As far as I understand the fix will be pushed internally so it will show up later in the GitHub.

coldsheephot:
Hello! I want to use model files on S3 for TF Serving. Can I use an S3 URI for `--model_base_path` when starting `tensorflow_model_server`? Is there any code changes to TF Serving required? (According to the code comments, TF Serving has already supported GCS.) Thanks!

quantumlicht:
@coldsheephot  let me refer you to this issue I opened https://github.com/tensorflow/serving/issues/615#issuecomment-335875164

If you could try it and report if it works for you that would be lovely. I will do the same thing

mas-dse-greina:
Is there a python example on how to use this? I'd like to be able to access an HDF5 file from my S3 bucket within TensorFlow.

Thanks.

yongtang:
@mas-dse-greina I added a very basic doc for s3 usage with TensorFlow in PR #16923.

mas-dse-greina:
Thanks.

I'd like to be able to access an HDF5 file on an S3 bucket. This way I
don't have to download the entire file at once.

I've managed to accomplish this with goofys mounting the S3 bucket as a
local directory. Do you think your code could be used in the same way with
h5py?

Thanks.
-Tony

PR #11104, opened 2017-06-28
Bug fixes to input function tutorial

LennyBoyatzis:
I tried running this example today and hit a few issues. This PR fixes the issues I found:

- DNNRegressor is located within `tf.contrib.learn.DNNRegressor` not `tf.estimator.DNNRegressor`
- the method `train` does not exist, it's `fit`
- returning `p` instead of `p['predictions']`

With the above fixes I was able to get the example running as expected.

### DIFF
--- a/tensorflow/examples/tutorials/input_fn/boston.py
+++ b/tensorflow/examples/tutorials/input_fn/boston.py
@@ -54,12 +54,12 @@ def main(unused_argv):
   feature_cols = [tf.feature_column.numeric_column(k) for k in FEATURES]
 
   # Build 2 layer fully connected DNN with 10, 10 units respectively.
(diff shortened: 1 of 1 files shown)


LennyBoyatzis:
I signed it!

drpngx:
Jenkins, test this please.

MarkDaoust:
Hi,

Thanks for the patch, but I don't think we can take this one.

This is part of the effort to get the docs updated *ahead* of moving the canned estimators out of `tf.contrib.learn` into `tf.estimators`. 

The current version is consistent with the behavior of `tf.estimator.Estimator`.

The 1.2 docs and examples work, but `master` is always a work in progress.

Thanks Again.

PR #11106, opened 2017-06-28
Feature: implement setAttrStringList method in Operation Builder with java api

Moriadry-zz:
Implement setAttrStringList method in Operation Builder with java api.


Moriadry:
@asimshankar please help me check this, many thanks!

Moriadry:
Will open another pr.

PR #11107, opened 2017-06-28
[Java] Add support for list(string) attributes in OperationBuilder

Moriadry-zz:
Feature: implement setAttrStringList method in Operation Builder with Java api

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
@@ -256,7 +256,7 @@ public final class OperationBuilder {
     return this;
   }
 
-  public OperationBuilder setAttrList(String name,  String... value) {
+  public OperationBuilder setAttr(String name,  String[] value) {
     Graph.Reference r = graph.ref();
     try {
       Object[] objects = new Object[value.length];
@@ -289,8 +289,6 @@ public final class OperationBuilder {
   // Java library types. Roughly, setAttrFoo calls the TensorFlow C library function: TF_SetAttrFoo.
   //
   // TODO(ashankar):
-  // - setAttrStringList: Which would take in an array of byte[] (java Strings will need to be UTF-8
-  //   encoded?)
   // - setAttrShapeList: Which would take in a long[][]
 
   private static native void setAttrString(long handle, String name, byte[] value);
--- a/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
@@ -257,12 +257,13 @@ public final class OperationBuilder {
   }
 
   public OperationBuilder setAttr(String name,  String[] value) {
+    Charset utf8 = Charset.forName("UTF-8");
+    Object[] objects = new Object[value.length];
+    for (int i = 0; i < value.length; ++i) {
+      objects[i] = value[i].getBytes(utf8);
+    }
     Graph.Reference r = graph.ref();
     try {
-      Object[] objects = new Object[value.length];
-      for (int i = 0; i < value.length; i++) {
-        objects[i] = value[i].getBytes(Charset.forName("UTF-8"));
-      }
       setAttrStringList(unsafeNativeHandle, name, objects);
     } finally {
       r.close();
--- a/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
@@ -256,6 +256,20 @@ public final class OperationBuilder {
     return this;
   }
 
+  public OperationBuilder setAttrList(String name,  String... value) {
+    Graph.Reference r = graph.ref();
+    try {
+      Object[] objects = new Object[value.length];
+      for (int i = 0; i < value.length; i++) {
+        objects[i] = value[i].getBytes(Charset.forName("UTF-8"));
+      }
+      setAttrStringList(unsafeNativeHandle, name, objects);
+    } finally {
+      r.close();
+    }
+    return this;
+  }
+
   private long unsafeNativeHandle;
   private Graph graph;
 
@@ -302,4 +316,7 @@ public final class OperationBuilder {
   private static native void setAttrTensorList(long handle, String name, long[] tensorHandle);
 
   private static native void setAttrShape(long handle, String name, long[] shape, int numDims);
+
+  private static native void setAttrStringList(long handle, String name, Object[] value);
+
 }
--- a/tensorflow/java/src/main/native/operation_builder_jni.cc
+++ b/tensorflow/java/src/main/native/operation_builder_jni.cc
@@ -262,7 +262,7 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrStringList(
         JNIEnv* env, jclass object, jlong handle, jstring name, jobjectArray values) {
   TF_OperationDescription* d = requireHandle(env, handle);
   if (d == nullptr) return;
-  const char *cname = env->GetStringUTFChars(name, nullptr);
+  const char* cname = env->GetStringUTFChars(name, nullptr);
   int num_values = env->GetArrayLength(values);
   static_assert(sizeof(jbyte) == 1,
                 "Require Java byte to be represented as a single byte");
--- a/tensorflow/java/src/main/native/operation_builder_jni.cc
+++ b/tensorflow/java/src/main/native/operation_builder_jni.cc
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
   TF_SetAttrShape(d, cname, cvalue.get(), static_cast<int>(num_dims));
   env->ReleaseStringUTFChars(name, cname);
 }
+
+JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrStringList(
+        JNIEnv* env, jclass object, jlong handle, jstring name, jobjectArray values) {
+ static_assert(sizeof(jbyte) == 1,
+                "Require Java byte to be represented as a single byte");
+  TF_OperationDescription* d = requireHandle(env, handle);
+  if (d == nullptr) return;
+  const char *cname = env->GetStringUTFChars(name, nullptr);
+  int num_strings = env->GetArrayLength(values);
+  size_t *lengths = new size_t[num_strings];
+  jbyte **strings = new jbyte *[num_strings];
+  for (int i = 0; i < num_strings; i++) {
+    jbyteArray value = (jbyteArray) env->GetObjectArrayElement(values, i);
+    lengths[i] = static_cast<size_t>(env->GetArrayLength(value));
+    strings[i] = env->GetByteArrayElements(value, nullptr);
+  }
+  if (num_strings > 0) {
+    TF_SetAttrStringList(d, cname, reinterpret_cast<const void *const *>(strings), lengths, num_strings);
+  } else {
+    TF_SetAttrStringList(d, cname, nullptr, nullptr, 0);
+  }
+  for (int i = 0; i < num_strings; i++) {
+    jbyteArray value = (jbyteArray) env->GetObjectArrayElement(values, i);
+    strings[i] = env->GetByteArrayElements(value, nullptr);
+    env->ReleaseByteArrayElements(value, strings[i], JNI_ABORT);
+    env->DeleteLocalRef(value);
+  }
+  env->ReleaseStringUTFChars(name, cname);
+}
--- a/tensorflow/java/src/main/native/operation_builder_jni.cc
+++ b/tensorflow/java/src/main/native/operation_builder_jni.cc
@@ -260,29 +260,27 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
 
 JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrStringList(
         JNIEnv* env, jclass object, jlong handle, jstring name, jobjectArray values) {
- static_assert(sizeof(jbyte) == 1,
-                "Require Java byte to be represented as a single byte");
   TF_OperationDescription* d = requireHandle(env, handle);
   if (d == nullptr) return;
   const char *cname = env->GetStringUTFChars(name, nullptr);
-  int num_strings = env->GetArrayLength(values);
-  size_t *lengths = new size_t[num_strings];
-  jbyte **strings = new jbyte *[num_strings];
-  for (int i = 0; i < num_strings; i++) {
-    jbyteArray value = (jbyteArray) env->GetObjectArrayElement(values, i);
-    lengths[i] = static_cast<size_t>(env->GetArrayLength(value));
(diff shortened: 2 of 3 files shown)


Moriadry:
Hi, @asimshankar  please help me check this pr, many thanks!

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
...
   env->ReleaseStringUTFChars(name, cname);
 }
+
+JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrStringList(
+        JNIEnv* env, jclass object, jlong handle, jstring name, jobjectArray values) {
+ static_assert(sizeof(jbyte) == 1,
+                "Require Java byte to be represented as a single byte");
+  TF_OperationDescription* d = requireHandle(env, handle);
+  if (d == nullptr) return;
+  const char *cname = env->GetStringUTFChars(name, nullptr);
```
asimshankar: Style nit: To be consistent with the styling used in the rest of this file, `const char* cname` instead of `const char *cname` etc.

Would you mind running [`clang-format`](http://clang.llvm.org/docs/ClangFormat.html) on C/C++ source code and [`google-java-format`](https://github.com/google/google-java-format) on Java source code to ensure consistent formatting?

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
   TF_SetAttrShape(d, cname, cvalue.get(), static_cast<int>(num_dims));
   env->ReleaseStringUTFChars(name, cname);
 }
+
+JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrStringList(
+        JNIEnv* env, jclass object, jlong handle, jstring name, jobjectArray values) {
+ static_assert(sizeof(jbyte) == 1,
+                "Require Java byte to be represented as a single byte");
```
asimshankar: Nit: Let's move this to closer to where the assumption is made (line 277?)

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
...
+  const char *cname = env->GetStringUTFChars(name, nullptr);
+  int num_strings = env->GetArrayLength(values);
+  size_t *lengths = new size_t[num_strings];
+  jbyte **strings = new jbyte *[num_strings];
+  for (int i = 0; i < num_strings; i++) {
+    jbyteArray value = (jbyteArray) env->GetObjectArrayElement(values, i);
+    lengths[i] = static_cast<size_t>(env->GetArrayLength(value));
+    strings[i] = env->GetByteArrayElements(value, nullptr);
+  }
+  if (num_strings > 0) {
```
asimshankar: Is this condition needed?
Seems like the statement in the "true" block will work in both cases?

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
...
+JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrStringList(
+        JNIEnv* env, jclass object, jlong handle, jstring name, jobjectArray values) {
+ static_assert(sizeof(jbyte) == 1,
+                "Require Java byte to be represented as a single byte");
+  TF_OperationDescription* d = requireHandle(env, handle);
+  if (d == nullptr) return;
+  const char *cname = env->GetStringUTFChars(name, nullptr);
+  int num_strings = env->GetArrayLength(values);
+  size_t *lengths = new size_t[num_strings];
+  jbyte **strings = new jbyte *[num_strings];
```
asimshankar: Where are these being deallocated? Is there a missing:

```
delete[] lengths;
delete[] strings;
```
in this function?

Better still, use `std::unique_ptr` as is used in other places in this file to avoid worrying about manual deletes (e.g., see definition of `DEFINE_SET_ATTR_LIST`), so this would change to:

```
std::unique_ptr<size_t[]> lengths(new size_t[num_strings]);
std::unique_ptr<jbyte*[]> strings(new jbyte*[num_string]));
```

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
...
+  for (int i = 0; i < num_strings; i++) {
+    jbyteArray value = (jbyteArray) env->GetObjectArrayElement(values, i);
```
asimshankar: Avoid using a C-style cast here (`static_cast` instead?)
(See https://google.github.io/styleguide/cppguide.html#Casting)

tensorflow/java/src/main/native/operation_builder_jni.h
```diff
@@ -169,6 +169,14 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrTensorList(
 JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
     JNIEnv *, jclass, jlong, jstring, jlongArray, jint);
 
+/*
+ * Class:     org_platanios_tensorflow_jni_Op__
```
asimshankar: This class seems incorrect

drpngx:
Jenkins, test this please.

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
...
+  }
+  if (num_strings > 0) {
+    TF_SetAttrStringList(d, cname, reinterpret_cast<const void *const *>(strings), lengths, num_strings);
+  } else {
+    TF_SetAttrStringList(d, cname, nullptr, nullptr, 0);
+  }
+  for (int i = 0; i < num_strings; i++) {
+    jbyteArray value = (jbyteArray) env->GetObjectArrayElement(values, i);
+    strings[i] = env->GetByteArrayElements(value, nullptr);
+    env->ReleaseByteArrayElements(value, strings[i], JNI_ABORT);
```
asimshankar: I don't believe this is correct, you want to provide `ReleaseByteArrayElements` the same pointer value that you got from a call to `GetByteArrayElements` above, not from the new call to `GetByteArrayElements` made here (which is allowed to return a different pointer value).

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
...
+    env->DeleteLocalRef(value);
```
asimshankar: I believe this is unnecessary, since we're doing it in the end (just before returning), it isn't buying us much. We could just remove this call, I think (see the `GetObjectArrayElement` example in https://www.ibm.com/developerworks/library/j-jni/index.html)

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,32 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
...
+  int num_strings = env->GetArrayLength(values);
+  size_t *lengths = new size_t[num_strings];
+  jbyte **strings = new jbyte *[num_strings];
+  for (int i = 0; i < num_strings; i++) {
+    jbyteArray value = (jbyteArray) env->GetObjectArrayElement(values, i);
+    lengths[i] = static_cast<size_t>(env->GetArrayLength(value));
+    strings[i] = env->GetByteArrayElements(value, nullptr);
+  }
+  if (num_strings > 0) {
+    TF_SetAttrStringList(d, cname, reinterpret_cast<const void *const *>(strings), lengths, num_strings);
```
asimshankar: The presence of this `reinterpret_cast`  and what I think is incorrect cleanup (see comments later on) makes me think that we should make this implementation more explicit and simpler. How about something like the following (do validate that this is correct :):

```
int num_values = env->GetArrayLength(values);
std::unique_ptr<jbyteArray[]> jarrays(new jbyteArray[num_values]);
std::unique_ptr<jbyte*[]> jvalues(new jbyte*[num_values]);
std::unique_ptr<void*[]> cvalues(new void*[num_values]);
std::unique_ptr<size_t[]> lengths(new size_t[num_values]);

for (int i = 0; i < num_values; ++i) {
  jByteArray v = static_cast<jbyteArray>(env->GetObjectArrayElement(values, i));
  jarrays[i] = v;
  jvalues[i] = env->GetByteArrayElements(v, nullptr);
  cvalues[i] = jvalues[i];
  lengths[i] = static_cast<size_t>(env->GetArrayLength(v));
}
TF_SetAttrStringList(d, cname, cvalues.get(), lengths.get(), num_values);
for (int i = 0; i < num_values; ++i) {
  env->ReleaseByteArrayElements(jarrays[i], jvalues[i], JNI_ABORT);
}
env->ReleaseStringUTFChars(name, cname);
```

I realize that this has additional allocations and overheads that could be optimized away. However, I don't think that this function is going to be common enough to be a performance bottleneck so the explicit clarity is well worth it. If, in the future the allocations are of a concern, we can optimize it, but for now we don't need to.

tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
```diff
@@ -256,6 +256,20 @@ public OperationBuilder setAttr(String name, Shape value) {
     return this;
   }
 
+  public OperationBuilder setAttrList(String name,  String... value) {
```
asimshankar: The TODO in line 278 can be removed now?

asimshankar (changes requested):
Thanks for the PR! A bunch of comments though :)

Moriadry:
@asimshankar ok, will be fixed, thank you.

tensorflow/java/src/main/native/operation_builder_jni.cc
Moriadry: what does that mean?

Moriadry: I got it.

Moriadry: @asimshankar in Java_org_tensorflow_OperationBuilder_setAttrString method, line 141. static_assert at the very beginning. Should I move this line? If you want to, I will fix it.

tensorflow/java/src/main/native/operation_builder_jni.cc
Moriadry: if num_strings==0, "i<num_strings" will be false, and I think the statement in true block will not work?

tensorflow/java/src/main/native/operation_builder_jni.cc
asimshankar: That's for catching that. Though, let's keep this PR focused on what it's adding for now (so leave the other one as is)

tensorflow/java/src/main/native/operation_builder_jni.cc
asimshankar: Not quite sure I follow. When `num_strings` is 0, then:

```
TF_SetAttrStringList(d, cname, reinterpret_cast<const void* const*>(strings), lengths, num_strings);
```

amounts to:

```
TF_SetAttrStringList(d, cname, <some pointer>, <some pointer>, 0);
```

which is fine, right?

Moriadry: Copy that.

tensorflow/java/src/main/native/operation_builder_jni.cc
Moriadry: Yeah, your solution looks more simpler and explicit. I'm not that familiar with C++, and trying to become more comfortable with it. I will change it to your version to make sure everything is ok. Thanks for your effort on tf and open source, learn a lot from you 👍

tensorflow/java/src/main/native/operation_builder_jni.cc
Moriadry: this may take a little time

tensorflow/java/src/main/java/org/tensorflow/OperationBuilder.java
```diff
@@ -256,6 +256,20 @@ public OperationBuilder setAttr(String name, Shape value) {
...
   }
 
+  public OperationBuilder setAttr(String name,  String[] value) {
+    Graph.Reference r = graph.ref();
+    try {
+      Object[] objects = new Object[value.length];
+      for (int i = 0; i < value.length; i++) {
+        objects[i] = value[i].getBytes(Charset.forName("UTF-8"));
+      }
+      setAttrStringList(unsafeNativeHandle, name, objects);
```
asimshankar: Two things:
(1) Let's tighten the `try` block
(2) Let's avoid repeated `Charset` lookups.

So, something like:

```java
Charset utf8 = Charset.forName("UTF-8");
Object[] objects = new Object[value.length];
for (int i = 0; i < value.length; ++i) {
  objects[i] = value[i].getBytes(utf8);
}
Graph.Reference r = graph.ref();
try {
  setAttrStringList(unsafeNativeHandle, name, objects);
} finally {
  r.close();
}
```

tensorflow/java/src/main/native/operation_builder_jni.cc
```diff
@@ -257,3 +257,30 @@ JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrShape(
   TF_SetAttrShape(d, cname, cvalue.get(), static_cast<int>(num_dims));
   env->ReleaseStringUTFChars(name, cname);
 }
+
+JNIEXPORT void JNICALL Java_org_tensorflow_OperationBuilder_setAttrStringList(
+        JNIEnv* env, jclass object, jlong handle, jstring name, jobjectArray values) {
+  TF_OperationDescription* d = requireHandle(env, handle);
+  if (d == nullptr) return;
+  const char *cname = env->GetStringUTFChars(name, nullptr);
```
asimshankar: Nit: `const char* cname` (not `const char *cname`)

asimshankar (changes requested):
Two minor changes, otherwise looks good.

I would have also encouraged a unittest but I realized that almost none of the core operations have a `list(string)` attribute. Which got me thinking about what prompted you to add support for list-of-string attributes? :)

It's fine, but unless I'm missing something, I don't think any core ops use it. So was curious about the impetus to add this :)

Moriadry:
@asimshankar Indeed I didn't find a core operations which contains the list(string) attribute by now, however I saw it was implemented in go. Also I found a todo in OperationBuilder.java. So I'm wondering perhaps It will be needed in some certain operation if created in future. So do you think it is unnecessary?

asimshankar (approved):
@Moriadry : Now that you have this, it's fine :).

asimshankar:
Jenkins, test this please.

PR #11110, opened 2017-06-28
change an error in annotation`

sunpro:
there is an error in annotation of ops.py in tensorflow\tensorflow\python\ops. I think it is just a spelling mistake


caisq:
"iff" means "if and only if"

Issue #7128, opened 2017-01-28
QR decomposition is slow


yaroslavvb:
We are doing a bunch of of QR decompositions in numpy. I did preliminary investigation of moving them to TF, but TF version is slow compared to numpy.

Below is a benchmark script that runs QR decomposition of 4096x4096 matrix. It took 7.3 seconds in TF and 1.93 in numpy MKL. Numpy MKL is the default numpy that comes when installing Anaconda.

version: HEAD from last week, built with `--config=cuda --config=opt`
cpu: 32 core Intel(R) Xeon(R) CPU E5-2630 v3 @ 2.40GHz
https://github.com/yaroslavvb/stuff/blob/master/tiny_runs/qr_test.py

Note that `pip install --upgrade $TF_BINARY_URL` will overwrite MKL numpy with OpenBLAS numpy that is actually slower than TF version. The way to check is to look at `np.__config__.show()` and look for strings like `mkl_intel_lp64`. You can get MKL version back by uninstalling numpy and doing `conda install numpy`

@rmlarsen

yaroslavvb:
@rmlarsen is there some planned work, or should I mark this as contributions welcome?

rmlarsen:
@yaroslavvb The current OP is using the single-threaded (CPU) Householder QR in Eigen. Does the MKL version run multi-threaded?

yaroslavvb:
Yes, MKL version seems to be using 800-900 CPU

suiyuan2009:
I'll working on this!

suiyuan2009:
I find MKL is already supported in the latest tf code...

suiyuan2009:
@yaroslavvb , I both installed mkl_dnn 2018 and full mkl 2018, add
`#ifdef INTEL_MKL

#define EIGEN_USE_MKL_ALL

#endif // INTEL_MKL` to `tensorflow/core/kernels/qr_op_impl.h`, run `bazel build --config=opt --config=cuda --define=using_mkl=true //tensorflow/tools/pip_package:build_pip_package` to build, got
`TF QR on 4096 by 4096 matrix in 2.59 seconds

numpy QR on 4096 by 4096 matrix in 2.64 seconds`.

yaroslavvb:
Oh interesting! PS, I thought latest MKL is 2017 update 2. At least that's the version I get when I download latest official Intel for Python distribution from Intel.

You can get MKL version using this snippet

```
import ctypes
import numpy as np

def mklVersion():
[...]

PR #11116, opened 2017-06-28
enable mkl in eigen for qr_op

suiyuan2009:
#7128

### DIFF
--- a/tensorflow/core/kernels/qr_op_impl.h
+++ b/tensorflow/core/kernels/qr_op_impl.h
@@ -20,6 +20,10 @@ limitations under the License.
 // improve compilation times.
 #include <algorithm>
 
+#ifdef INTEL_MKL
+#define EIGEN_USE_MKL_ALL
+#endif // INTEL_MKL
+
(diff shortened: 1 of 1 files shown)


suiyuan2009:
`--define=using_mkl=true` must be added when using bazel to build, sometimes you need to clean `/path/to/tensorflow/third_party/mkl` first(remove `.so` file),  I added `/path/to/tensorflow/third_party/mkl` to `LD_LIBRARY_PATH` in `~/.bash_profile`.

drpngx:
Jenkins, test this please.

drpngx:
I'm a bit wary of having special instructions out. Could you add them in the documentation?

@benoitsteiner WDYT?

suiyuan2009:
it seems mkl is not enabled even if I configure to use mkl when I build from source, so maybe it's still under developing?

gunan (approved):

gunan:
We made a recent change about how we build with MKL.
I removed the configure option for MKL, but now you can just add `--config=mkl` to your build command to build with MKL support.

PR #11120, opened 2017-06-28
fix link in README.md in Using TensorFlow via Docker

chunjy92:
correctly link parameterized_docker_build.sh to its actual location, instead of README.md


chunjy92:
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

Signed CLA!

caisq (approved):

caisq:
Thanks for the fix, @chunjy92 !

PR #11130, opened 2017-06-29
Add bazel build configs for cpu type 'x64_windows_msvc'

snnn:
Without this fix, the final generated whl file will contain a '_pywrap_tensorflow_internal.so' instead of '_pywrap_tensorflow_internal.pyd'


meteorcloudy:
@snnn This change looks good to me, but after https://github.com/tensorflow/tensorflow/commit/08ed32dbb9e8f67eec9efce3807b5bdb3933eb2f, with Bazel 0.5.0 or later version you don't need `--cpu=x64_windows_msvc` to build TensorFlow anymore.
And it will be deprecated in future.

snnn:
got it.

PR #11136, opened 2017-06-29
Windows: Add missing source files declaration for Bazel build

meteorcloudy:
Fix http://ci.tensorflow.org/job/tf-master-win-bzl/1169/
@gunan

### DIFF
--- a/tensorflow/core/BUILD
+++ b/tensorflow/core/BUILD
@@ -1216,6 +1216,8 @@ LIB_INTERNAL_WINDOWS_DEPS = glob(
         "platform/*.cc",
         "platform/profile_utils/**/*.h",
         "platform/profile_utils/**/*.cc",
+        "framework/resource_handle.h",
+        "framework/resource_handle.cc",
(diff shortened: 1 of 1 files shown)


meteorcloudy:
Testing http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/30/console

tensorflow/core/BUILD
```diff
@@ -1216,6 +1216,9 @@ LIB_INTERNAL_WINDOWS_DEPS = glob(
         "platform/*.cc",
         "platform/profile_utils/**/*.h",
         "platform/profile_utils/**/*.cc",
+    ] + [
```
gunan: Do we need this line?
Maybe we can have a comment to explain these, but not sure about two separate lists.

meteorcloudy: I copied from somewhere else, I think we can remove the extra line. PTAL.

gunan (approved):

Issue #8873, opened 2017-03-31
Arbitrary dimension support for tf.tile and binary operators


girving:
`tf.transpose` uses template specialization for dimensions <= 5, then falls back to a slightly slower generic implementation which works for any dimension:

https://github.com/tensorflow/tensorflow/blob/504b91de8fa9a9cc4a4e17e59ed753ab677a1410/tensorflow/core/kernels/transpose_functor_cpu.cc#L23

It would be nice to use the same mechanism for `tf.tile` and the various binary operators.

girving:
If anyone wants to take this on, I'd recommend trying `tf.tile` first: it's closer to `tf.transpose`.  Happy to answer questions.

yanchen036:
Hi @girving ,
I'd like to work on this issue. I've written sketch code for it:
https://github.com/yanchen036/tensorflow/blob/arbitrary_dim_for_tile/tensorflow/core/kernels/tile_ops_impl.h#L55
https://github.com/yanchen036/tensorflow/blob/arbitrary_dim_for_tile/tensorflow/core/kernels/tile_ops.cc#L187

Are these the same as what you think? If I'm wrong pls correct me!

girving:
@yanchen036 Those links don't make it easy to see diffs.  Am I right that you just added a stub for TileSimple but didn't write it?  If so, the existence of TileSimple is on the right path.  I'm not sure if you made other changes.

yanchen036:
@girving I did 2 things. 
First, I wrote a TileSimple stub(now I know that's not you need) and a TileUsingEigen function
```c++
void TileUsingEigen(const Device& d, const Tensor& in, Tensor* out,
                    const gtl::ArraySlice<int32> broadcast_array)
```
which implements the old Tile::operator(). See https://github.com/yanchen036/tensorflow/blob/arbitrary_dim_for_tile/tensorflow/core/kernels/tile_ops_impl.h#L32

Second, I changed the Tile::operator() function from
```c++
template <typename Device, typename T, int NDIM>
void operator()(const Device& d, typename TTypes<T, NDIM>::Tensor out,
                typename TTypes<T, NDIM>::ConstTensor in,
                const Eigen::array<int32, NDIM>& broadcast_array) const
```
To
```c++
template <typename Device, typename T, int NDIM>
void operator()(const Device& d, const Tensor& in, Tensor* out,
                const gtl::ArraySlice<int32> broadcast_array) const {
    switch (in.dims()) {
        case 0:
[...]

PR #11140, opened 2017-06-29
Arbitrary dim for slice

yanchen036:
Add arbitrary dim support for slice op(https://github.com/tensorflow/tensorflow/issues/8873)
There is still a compiling problem. When build //tensorflow/core/kernel:slice_op_test，I got the following errors:
```
ERROR: /home/parallels/Desktop/github/tensorflow/tensorflow/core/kernels/BUILD:1153:1: Linking of rule '//tensorflow/core/kernels:slice_op_test' failed: gcc failed: error executing command /usr/bin/gcc -o bazel-out/local-opt/bin/tensorflow/core/kernels/slice_op_test '-Wl,-rpath,$ORIGIN/../../../_solib_k8/' -Lbazel-out/local-opt/bin/_solib_k8 -pthread '-fuse-ld=gold' -Wl,-no-as-needed ... (remaining 6 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
bazel-out/local-opt/bin/tensorflow/core/kernels/_objs/slice_op/tensorflow/core/kernels/slice_op.o:slice_op.cc:function tensorflow::SliceOp<Eigen::ThreadPoolDevice, Eigen::QInt8>::Compute(tensorflow::OpKernelContext*): error: undefined reference to 'void tensorflow::internal::SliceSimple<Eigen::ThreadPoolDevice, Eigen::QInt8>(Eigen::ThreadPoolDevice const&, tensorflow::Tensor*, tensorflow::Tensor const&, tensorflow::gtl::ArraySlice<long long> const&)'
bazel-out/local-opt/bin/tensorflow/core/kernels/_objs/slice_op/tensorflow/core/kernels/slice_op.o:slice_op.cc:function tensorflow::SliceOp<Eigen::ThreadPoolDevice, Eigen::QUInt8>::Compute(tensorflow::OpKernelContext*): error: undefined reference to 'void tensorflow::internal::SliceSimple<Eigen::ThreadPoolDevice, Eigen::QUInt8>(Eigen::ThreadPoolDevice const&, tensorflow::Tensor*, tensorflow::Tensor const&, tensorflow::gtl::ArraySlice<long long> const&)'
bazel-out/local-opt/bin/tensorflow/core/kernels/_objs/slice_op/tensorflow/core/kernels/slice_op.o:slice_op.cc:function tensorflow::SliceOp<Eigen::ThreadPoolDevice, Eigen::QInt32>::Compute(tensorflow::OpKernelContext*): error: undefined reference to 'void tensorflow::internal::SliceSimple<Eigen::ThreadPoolDevice, Eigen::QInt32>(Eigen::ThreadPoolDevice const&, tensorflow::Tensor*, tensorflow::Tensor const&, tensorflow::gtl::ArraySlice<long long> const&)'
```
It looks like there are some types which haven't be instantiated. However, I can't find that where are these type used. And simply add these instantiation is not a good idea. Could you give some help? @girving


girving:
I think the `SliceSimple` problem is that you are declaring instantiations for `Slice` in a file that defines `SliceSimple`.  But the place that uses `Slice` doesn't know that you've already instantiated it, so it instantiates it again (wasting code).  And the file that declares `SliceSimple` doesn't know that `SliceSimple` needs to be used outside of that file, so it does a locally visible instantiation (or perhaps inlines everything.

Bottom line: You should declare instantiations of stuff you define in that file, not other stuff.

girving:
Jenkins, test this please.

drpngx:
Why are we merging the `cpu_impl` files?  As I understand it, they take a long time so that's why we split them.

tensorflow/core/kernels/slice_op_cpu_impl_6.cc
```diff
@@ -1,18 +0,0 @@
-/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
drpngx: These files are here to minimize the compile time.

drpngx (changes requested):

tensorflow/core/kernels/slice_op_cpu_impl_6.cc
yanchen036: I see, but I delete the template parameter "NDIM", how could I bring these files back? There is similar implementation in transpose_op https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/transpose_functor_cpu.cc#L55

yanchen036:
fixed compiling problem and pass python unit test

yanchen036:
@girving could you test this code please?

girving:
Unfortunately I no longer work at Google, and am currently on vacation. 
@drpngx: Can you take over?

drpngx:
Jenkins, test this please.

@girving yes, on it!

yanchen036:
```
ERROR: /workspace/tensorflow/stream_executor/BUILD:39:1: undeclared inclusion(s) in rule '//tensorflow/stream_executor:cuda_platform':
this rule is missing dependency declarations for the following files included by 'tensorflow/stream_executor/cuda/cuda_event.cc':
  'bazel-out/local_linux-opt/genfiles/external/local_config_cuda/cuda/include/cuda.h'.
```
@drpngx seems the error is about latest changes in 'tensorflow/stream_executor/cuda/cuda_event.cc'. I update my code to master, Could you test it again? Thank you.

drpngx:
Jenkins, test this please.

tensorflow/core/kernels/slice_op_cpu_impl_6.cc
drpngx: The static part can be compiled into multiple files. Each branch of the switch statement goes into a separate file.

yanchen036: Do you means these?
file: tensorflow/core/kernels/slice_op.h
```C++
template <typename Device, typename T, int NDIMS>
struct Slice {
  void operator()(const Device& d, Tensor* out, const Tensor& in,
                  const gtl::ArraySlice<int64>& slice_indices,
                  const gtl::ArraySlice<int64>& slice_sizes) {
    switch (in.dims()) {
      case NDIMS:
        internal::SliceUsingEigen<Device, T, NDIMS>(d, out, in, slice_indices, slice_sizes);
        break;
      default:
        if (Eigen::internal::is_same<Device, Eigen::GpuDevice>::value) {
          internal::SliceSimpleGpu<Device, T>(d, out, in, slice_indices);
        } else {
          internal::SliceSimple<Device, T>(d, out, in, slice_indices);
        }
        break;
    }
  }
};
```
and file: tensorflow/core/kernels/slice_op_cpu_impl_1.cc
```C++
#define CPU_PROVIDED_IXDIM 1
#include "tensorflow/core/kernels/slice_op_cpu_impl.h"
#undef CPU_PROVIDED_IXDIM
```

drpngx: Yes something like that.

yanchen036:
@drpngx add impl_*.cc files back, pls review and run test

drpngx:
Jenkins, test this please.

tensorflow/core/kernels/slice_op.cc
```diff
@@ -192,39 +192,23 @@ class SliceOp : public OpKernel {
...
     return;                                         \
   }
-
       HANDLE_DIM(1);
       HANDLE_DIM(2);
       HANDLE_DIM(3);
       HANDLE_DIM(4);
       HANDLE_DIM(5);
       HANDLE_DIM(6);
-      HANDLE_DIM(7);
```
drpngx: Why not 7?

tensorflow/core/kernels/slice_op.cc
```diff
@@ -264,24 +248,13 @@ class MklSliceOp : public OpKernel {
...
-      HANDLE_DIM(4);
-      HANDLE_DIM(5);
-      HANDLE_DIM(6);
-      HANDLE_DIM(7);
-
-#undef HANDLE_DIM
-
-      OP_REQUIRES(context, false, errors::Unimplemented(
-                                      "SliceOp : Unhandled input dimensions"));
+      // special case for handling 4-D tensor slice
```
drpngx: "Special case for handling 4-d tensor slice."

tensorflow/core/kernels/slice_op.h
```diff
@@ -19,31 +19,84 @@ limitations under the License.
...
+template <typename Device, typename T>
+void SliceSimpleGpu(const Device& d, Tensor* out, const Tensor& in,
+                 const gtl::ArraySlice<int64>& slice_indices);
+
+template <typename Device, typename T>
+void SliceSimple(const Device& d, Tensor* out, const Tensor& in,
+                 const gtl::ArraySlice<int64>& slice_indices) {
+  const int ndims = in.dims();
+  const int64 nelem = out->NumElements();
+  gtl::InlinedVector<int64, 8> in_strides = ComputeStride<int64>(in.shape());
```
drpngx: const

tensorflow/core/kernels/slice_op.h
```diff
@@ -19,31 +19,84 @@ limitations under the License.
...
+  gtl::InlinedVector<int64, 8> out_strides = ComputeStride<int64>(out->shape());
```
drpngx: const

tensorflow/core/kernels/slice_op.h
```diff
@@ -19,31 +19,84 @@ limitations under the License.
...
+  auto output = out->tensor<T, NDIMS>();
+  Eigen::DSizes<int, NDIMS> indices;
+  for (int i = 0; i < NDIMS; ++i) {
+    indices[i] = slice_indices[i];
+  }
+  Eigen::DSizes<int, NDIMS> sizes;
+  for (int i = 0; i < NDIMS; ++i) {
+    sizes[i] = slice_sizes[i];
+  }
+  bool use_64bit = (input.size() > Eigen::NumTraits<int>::highest());
```
drpngx: const. No need for parens.

drpngx (changes requested):

tensorflow/core/kernels/slice_op.cc
yanchen036: case 7 is in line 208. It handle all the cases which dim >= 7

tensorflow/core/kernels/slice_op.cc
yanchen036: handled in line 304

tensorflow/core/kernels/slice_op.h
yanchen036: ok

tensorflow/core/kernels/slice_op.h
yanchen036: ok

tensorflow/core/kernels/slice_op.h
yanchen036: ok

yanchen036:
Is there anything I need to do?

tensorflow/core/kernels/slice_op.h
```diff
@@ -19,31 +19,84 @@ limitations under the License.
...
+void SliceSimple(const Device& d, Tensor* out, const Tensor& in,
+                 const gtl::ArraySlice<int64>& slice_indices) {
+  const int ndims = in.dims();
+  const int64 nelem = out->NumElements();
+  const gtl::InlinedVector<int64, 8> in_strides = ComputeStride<int64>(in.shape());
+  const gtl::InlinedVector<int64, 8> out_strides = ComputeStride<int64>(out->shape());
+  const T* p = in.flat<T>().data();
+  T* q = out->flat<T>().data();
+
+  for (int64 o_idx = 0; o_idx < nelem; ++o_idx) {
```
aselle: This implementation is correct but relatively slow. It doesn't use threading and uses a lot of divides and mods. I would add  TODO to indicate that this could be improved if performance of higher order tensor becomes a problem.

aselle: Specifically, I think if you incremented the input index in non-collapsed form it might be somewhat faster.

aselle (commented):
LGTM generally, but I would add a benchmark for this in the vein of the one I wrote for strided_slice_op_test.cc 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/strided_slice_op_test.cc

It would be interesting to know how much slower the higher dimension version is than the main version i.e. try two slices one with say a d slice and one with an 8d slice.

yanchen036:
@aselle I've added some benchmark, Please take a look.

tensorflow/core/kernels/slice_op.h
yanchen036: Yes, you're right, but I've no idea how to uncollapse the for loop. Do you have any suggestion?

rmlarsen:
@aselle @drpngx can you please take another look?

jhseu:
@aselle @drpngx friendly ping

tensorflow/core/kernels/slice_op.cc
drpngx: I mean, please capitalize and punctuate comment.

yanchen036: Done.

tensorflow/core/kernels/slice_op.h
drpngx: You can always unroll loops with Duff's device which batches up the divisions by 8, not if that's what Andy meant. It feels like the inner loop has lots of computation repeated for every element, so if you can work out a version where you have ndims on the outer and nelems on the inner loop then it would be faster.

It would be nice to measure the improvement on a benchmark.

yanchen036: How about use a buffer to record all outer's i_idx and t, in this way, I can switch the inner loop and outer loop. It's space-time tradeoff.

drpngx: Sounds like a reasonable trade.

martinwicke:
@aselle @drpngx ping?

martinwicke:
Jenkins, test this please.

tensorflow/core/kernels/slice_op.h
yanchen036: @drpngx uncollapse the inner loop, pls review it.

yanchen036:
@drpngx uncollapse the for loop by Duff's device, pls take a look. Thank you.

drpngx:
Jenkins, test this please.

drpngx:
If you had some benchmarks for before/after, please report them here. That would help.

yanchen036:
Yes, but I can't find the output of benchmark result. I read the rest log, there isn't benchmark result there. Could you tell me where to get the benchmark result? thanks.

sb2nov:
@drpngx ping

drpngx (approved):

drpngx:
We don't have a benchmark for slice. Can you create a program to run on a large number of random values, exercising both cases when the outer and inner loop have different sizes?

sb2nov:
Jenkins, test this please.

yanchen036:
Jenkins, test this please.

gunan:
Could you rebase this PR on top of master?
Also, was the comment by @drpngx addressed?

tensorflow/python/kernel_tests/slice_op_test.py
```diff
@@ -227,6 +251,7 @@ def testPartialShapeInference(self):
     m2 = array_ops.slice(z, [0, 0, 0], [constant_op.constant(1) + 0, 2, -1])
     self.assertAllEqual(m2.get_shape().as_list(), [None, 2, None])
 
+>>>>>>> upstream/master
```
gunan: Forgotten from the merge?

yanchen036:
sorry, removed line ">>>>>>>"

gunan:
Jenkins, test this please.

PR #11153, opened 2017-06-29
Fix sparse_placeholder error when int is passed in shape argument

### LINKED ISSUE #6749: sparse_placeholder no longer accepts python ints in shape argument
Hi,

After recently updating tensorflow, sparse_placeholder stopped working correctly.  It appears that the shape argument must now be int64 in order for tensorflow to convert the shape to a tensor, so the following fails:
```
ph = tf.sparse_placeholder(dtype=tf.float32, shape=(50, 10000))
```
with error message:
```
ValueError: Tensor conversion requested dtype int64 for Tensor with dtype int32: 'Tensor("Const:0", shape=(2,), dtype=int32)'
```
This is inconsistent with the behavior of tf.placeholder, for which: 
```
ph = tf.placeholder(dtype=tf.float32, shape=(50, 10000))
```
succeeds. 

Thanks,
Shawn

yongtang:
This fix tries to address the issue raised in #6749 where and error
```
ValueError: Tensor conversion requested dtype int64 for Tensor with dtype int32: 'Tensor("Const:0", shape=(2,), dtype=int32)'
```
is thrown out in case int is passed in the shape argument.

This fix fixes the issue and adds test cases for it.

This fix fixes #6749.

### DIFF
--- a/tensorflow/python/kernel_tests/sparse_ops_test.py
+++ b/tensorflow/python/kernel_tests/sparse_ops_test.py
@@ -909,5 +909,16 @@ class SparseTransposeTest(test.TestCase):
           self.assertAllEqual(dn_trans, expected_trans)
 
 
+class SparsePlaceholderTest(test.TestCase):
+
+  def testPlaceholder(self):
+    with self.test_session(use_gpu=False):
+      foo = array_ops.sparse_placeholder(dtypes.float32, shape=(10, 47))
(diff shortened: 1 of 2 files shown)


drpngx:
Jenkins, test this please.

mrry (approved):
Looks reasonable to me.

Issue #11121, opened 2017-06-28
reset_default_graph awkwardly breaks graph nesting.


Thenerdstation:
### System information
- **Linux Ubuntu 16.04**:
- **TF version 1.1.0, though it should still be present in master**:

### Bug
Running this:
```python
import tensorflow as tf

g = tf.Graph()
with g.as_default():
  tf.reset_default_graph()
  # Build something...
  a = tf.constant(1.0)
  b = tf.constant(1.0)
  c = a + b
  # Bug happens on exit of the with statement
```
Causes this

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-1-af7c5ed96704> in <module>()
      7   a = tf.constant(1.0)
      8   b = tf.constant(1.0)
----> 9   c = a + b

/usr/lib/python3.5/contextlib.py in __exit__(self, type, value, traceback)
     64         if type is None:
     65             try:
---> 66                 next(self.gen)
     67             except StopIteration:
     68                 return

/usr/local/lib/python3.5/dist-packages/tensorflow/python/framework/ops.py in get_controller(self, default)
   3626     finally:
   3627       if self._enforce_nesting:
-> 3628         if self.stack[-1] is not default:
   3629           raise AssertionError(
   3630               "Nesting violated for default stack of %s objects"

IndexError: list index out of range
```

This bug causes a confusing error message. The stack should either include the new default graph created by reset_default_graph or a warning/error message should happen when you reset the default within nested graphs.

printdhruv:
The default graph is a property of the current thread. `tf.reset_default_graph()` function applies only to the current thread. Calling this function while a tf.Session or tf.InteractiveSession is active will result in undefined behavior. Using any previously created tf.Operation or tf.Tensor objects after calling this function will result in undefined behavior.

```
import tensorflow as tf
g = tf.Graph()
with g.as_default():
  a = tf.constant(1.0)
  b = tf.constant(1.0)
  c = a + b
```

Thenerdstation:
Right, but you don't need an active session or any previously created ops/tensors to cause the error. Just calling reset_default_graph within nested graphs causes a confusing error.

Just this:
```python
import tensorflow as tf

g = tf.Graph()
with g.as_default():
  tf.reset_default_graph()
```

Also breaks with the same error. 
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-1-8377d035bd16> in <module>()
[...]

PR #11158, opened 2017-06-30
Added assertion error in reset_default_graph()

chaserileyroberts:
See #11121

### DIFF
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -3873,7 +3873,7 @@ class _DefaultStack(threading.local):
   def reset(self):
     self.stack = []
 
-  def is_cleared():
+  def is_cleared(self):
     return self.stack == []
 
   @property
--- a/tensorflow/python/framework/ops.py
+++ b/tensorflow/python/framework/ops.py
@@ -3873,6 +3873,9 @@ class _DefaultStack(threading.local):
   def reset(self):
     self.stack = []
 
+  def is_cleared():
+    return self.stack == []
+
   @property
   def enforce_nesting(self):
     return self._enforce_nesting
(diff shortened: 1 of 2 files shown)


drpngx:
That might trigger an API change.

Jenkins, test this please.

Thenerdstation:
Tests should pass now unless I did something very wrong on my end.

drpngx:
Jenkins, test this please.

drpngx (approved):

caisq:
@Thenerdstation @drpngx  FYI, it seems that the PR is causing a test failure. The failure seems to be specific to Mac , for reasons I don't fully understand. See example log from Jenkins nightly Mac build:
http://ci.tensorflow.org/view/Tensorflow%20Jenkins%20Monitored%20builds/job/nightly-matrix-cpu/TF_BUILD_IS_OPT=OPT,TF_BUILD_IS_PIP=NO_PIP,TF_BUILD_PYTHON_VERSION=PYTHON3,label=mac-slave/537/consoleFull

If you have thoughts, please let me know.

drpngx:
That is very odd. It passed when I merged.

The `exporter_test` looks like a genuine error, except that it's Mac only, and the `session_test` points to some other part of the code where `self.stack` is used. The `generate_lib_test` is probably some other error :(

Checking the submit logs, it looks like `exporter_test` did not exist on Mac at the time, but `session_test` passed. `session_test` has been there for a while and it's not clear what changed.

Given the risk that the extra strict check would blow up when merging into google code, we might want to revert this CL and push it from the inside with more internal test coverage.

Thenerdstation:
Yeah, I don't own a mac so I wasn't able to test for that, sorry.

PR #11167, opened 2017-06-30
Improve docstrings involving package structure

taehoonlee:
This PR improves docstrings involving `tf.learn`. In the case of `tensorflow_dataframe.py`, I think it is better to put just `DataFrame` in order to keep consistency within one file. Except for the case, all the `tf.learn`s are replaced by `tf.contrib.learn`s.

### DIFF
--- a/tensorflow/contrib/layers/python/layers/feature_column.py
+++ b/tensorflow/contrib/layers/python/layers/feature_column.py
@@ -15,13 +15,13 @@
 """This API defines FeatureColumn abstraction.
 
 FeatureColumns provide a high level abstraction for ingesting and representing
-features in tf.learn Estimator models.
+features in `Estimator` models.
 
 FeatureColumns are the primary way of encoding features for pre-canned
-tf.learn Estimators.
+`Estimator` models.
 
-When using FeatureColumns with tf.learn models, the type of feature column you
-should choose depends on (1) the feature type and (2) the model type.
+When using FeatureColumns with `Estimator` models, the type of feature column
+you should choose depends on (1) the feature type and (2) the model type.
 
 (1) Feature type:
 
@@ -74,7 +74,7 @@ should choose depends on (1) the feature type and (2) the model type.
       columns=[department_column, bucketized_age_column],
       hash_bucket_size=1000)
 
-Example of building tf.learn model using FeatureColumns:
+Example of building an `Estimator` model using FeatureColumns:
 
   # Define features and transformations
   deep_feature_columns = [age_column, embedded_dept_column]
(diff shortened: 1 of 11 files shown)


drpngx:
Thank you for putting this together. I think that the original intent was to have "tf.learn" as a project name, not reflecting the actual package name. In that case we would want to leave it as is.

@martinwicke WDYT?

martinwicke:
The tf.learn name is confusing and we want to basically remove it completely. Using the package name is fine, but there's usually no need to. Ideally, we'd just get rid of the qualifier and talk about e.g., Estimator directly, or if clarification is required, tf.estimators.Estimator.

taehoonlee:
I agree with the feedback, thank you @drpngx and @martinwicke! This PR has been revised as follows:
- `tf.learn` -> `Estimator` under `learn.estimators` and `layers.feature_column`.
- `tf.learn.DataFrame` -> `DataFrame` under `learn.dataframe`.
- `tf.learn/inflow` -> `inflow` under installation shell scripts.
I would like you to review again this version.

tensorflow/contrib/layers/python/layers/feature_column.py
```diff
@@ -74,7 +74,7 @@
       columns=[department_column, bucketized_age_column],
       hash_bucket_size=1000)
 
-Example of building tf.learn model using FeatureColumns:
+Example of building `Estimator` model using FeatureColumns:
```
drpngx: Can you check the grammar RE plurals, "A Estimator" etc?

taehoonlee: Yes, I have been checked the grammar as follows:
```
$ grep -r "\`A Estimator\`" . | wc -l
      0
$ grep -r "\`Estimator\`" . | wc -l
      100
$ grep -r "\`Estimators\`" . | wc -l
      1
```

The "Estimator" is the most frequently used word among the above.

drpngx: It doesn't sound good grammatically. I'm thinking that this should be an Example of building *an* Estimator, for instance.

taehoonlee: Thank you! I put "an" or plural forms on usages of `Estimator`.

vrv:
ping for @drpngx (or @martinwicke )

drpngx:
Sorry I'm out this week. Martin, feel free to reply if you get to it sooner

drpngx:
Jenkins, test this please.

drpngx (approved):
Thanks!

PR #11170, opened 2017-06-30
Adding support for s390x for boringssl 

Nayana-ibm:
Adding support for s390x for boringssl through patched_http_archive rule.

### DIFF
--- a/tensorflow/workspace.bzl
+++ b/tensorflow/workspace.bzl
@@ -520,7 +520,7 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
       actual = "@jsoncpp_git//:jsoncpp",
   )
 
-  native.http_archive(
+  patched_http_archive(
       name = "boringssl",
       urls = [
           "http://mirror.bazel.build/github.com/google/boringssl/archive/bbcaa15b0647816b9a1a9b9e0d209cd6712f0105.tar.gz",
@@ -528,6 +528,9 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
       ],
       sha256 = "025264d6e9a7ad371f2f66d17a28b6627de0c9592dc2eb54afd062f68f1f9aa3",
       strip_prefix = "boringssl-bbcaa15b0647816b9a1a9b9e0d209cd6712f0105",
+
+      # Add patch to boringssl code to support s390x
+      patch_file = str(Label("//third_party/boringssl:add_boringssl_s390x.patch")),
   )
 
   native.new_http_archive(
--- a/third_party/boringssl/add_boringssl_s390x.patch
+++ b/third_party/boringssl/add_boringssl_s390x.patch
@@ -1,0 +1,14 @@
+diff --git a/src/include/openssl/base.h b/src/include/openssl/base.h
+index 7a3adfb..88012ad 100644
+--- a/src/include/openssl/base.h
++++ b/src/include/openssl/base.h
+@@ -94,6 +94,8 @@ extern "C" {
+ #elif defined(__pnacl__)
+ #define OPENSSL_32_BIT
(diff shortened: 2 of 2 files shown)


gunan:
Jenkins, test this please.

drpngx:
@Nayana-ibm did you try to push this upstream?

gunan:
We tried, but boringssl was not willing to maintain this.
We discussed this with @martinwicke and we decided it is OK to patch boringssl here as long as s390x support is maintained by ibm, which @Nayana-ibm 

Does this capture the whole story, Nayana?

gunan (approved):

drpngx:
Thank you @gunan for the context!

Nayana-ibm:
@gunan Yes..correct. Boringssl was not willing to maintain this s390x support.
@drpngx  Hence this PR.

agl:
Please note that while it's correct that BoringSSL declined to carry this patch, that's because we don't support big-endian systems and _will produce the wrong results on big-endian systems_. The unittests will catch this all over the place and that's why it's not possible to compile BoringSSL on big-endian systems.

In keeping with most Google software, we have no plans to support big-endian systems and recommend that this patch be reverted. It'll just mislead people.

martinwicke:
So basically, anything that uses boringssl will likely be broken on s390x. 

In fact, the only part of TensorFlow that uses boringssl is the core/platform/cloud:oauth_client target. 

@Nayana-ibm since that target is not going to work anyway, could you look into removing this patch, and simply disabling the oauth support on big endian instead?

Nayana-ibm:
@martinwicke I will prepare one more PR to remove the patch added for s390x support in boringssl.
However I'm not sure how to disable the oauth support. Could you please let me know.

gunan:
Could you try disabling GCP support whenever we are on a big endian system?
I think that should do the trick. You can then disable the tests under `tensorflow/core/platform/cloud` with a tag like `no_big_endian`, then we need to edit the s390x testing bazel command with `--test_tag_filters=-no_big_endian`

Nayana-ibm:
Hi @gunan Sorry for late reply on this. 
I could see that the PR changes made for s390x are now reverted in Tensorflow master.

About disabling GCP support on a big-endian systems,  I followed following steps:
~/tensorflow# ./configure

> You have bazel 0.6.1- (@non-git) installed.
> Please specify the location of python. [Default is /usr/bin/python]:
> Found possible Python library paths:
>   /usr/local/lib/python2.7/dist-packages
>   /usr/lib/python2.7/dist-packages
> Please input the desired Python library path to use.  Default is [/usr/local/lib/python2.7/dist-packages]
> 
> Do you wish to build TensorFlow with jemalloc as malloc support? [Y/n]:
> jemalloc as malloc support will be enabled for TensorFlow.
> 
> Do you wish to build TensorFlow with Google Cloud Platform support? [Y/n]: n
> No Google Cloud Platform support will be enabled for TensorFlow.

~/tensorflow# bazel build -c opt //tensorflow/tools/pip_package:build_pip_package
```

Even after disabling GCP support, build fails with an error:
ERROR: /root/.cache/bazel/_bazel_root/30b0ed2738c25ecefa4b108f71649fc1/external/boringssl/BUILD:115:1: C++ compilation of rule '@boringssl//:crypto' failed (Exit 1).
In file included from external/boringssl/src/include/openssl/bio.h:60:0,
                 from external/boringssl/src/crypto/bio/printf.c:57:
external/boringssl/src/include/openssl/base.h:114:2: error: #error "Unknown target CPU"
 #error "Unknown target CPU"
  ^

```
Need more changes here?

On test cases part, I have added `tags = ["no_big_endian"] ` in `tensorflow/core/platform/cloud/BUILD `file. 
Yet to verify this changes.

PR #11172, opened 2017-06-30
[XLA] Ensure that the 2 data types chosen by the test are both valid for the device

DavidNorman:
Simple change.  The test previously assumed that all devices support float64, which isn't true for our device.

Now the test will find 2 different floating point types and use those.  Doesn't execute if there are not 2 floating point types to choose from.


tensorflow/compiler/tests/tensor_array_ops_test.py
```diff
@@ -325,21 +325,24 @@ def testTensorArrayWriteWrongIndexOrDataTypeFails(self):
...
-      r0_bad = gen_data_flow_ops._tensor_array_read_v3(
-          handle=w0.handle, index=0, dtype=dtypes.float64, flow_in=w0.flow)
-      with self.assertRaisesOpError(
-          "TensorArray dtype is float but op has dtype double."):
-        r0_bad.eval()
-
-      # Test reading from a different index than the one we wrote to
-      w0.read(1)
+    if len(self.float_types) > 1:
+      dtype1 = self.float_types[0]
```
drpngx: Please add a comment like `'# tf.float32 if supported'`, and `tf.float64 if supported, tf.float32 otherwise`.

DavidNorman: What it is really doing is picking 2 float types which are not the same. In the case of a CPU or GPU it will probably pick f32 and f64. However for our device it is f16 and f32, since we don't have f64 support (yet). 

I suspect the types selected depend on the order that the test framework specifies them. 

Perhaps a good comment would be to describe the above (in a concise way)

DavidNorman: I suppose the ideal thing to do is to use every pair of dissimilar floating point types (or maybe every pair of dissimilar types).  I could code that up if you would like.

DavidNorman:
hi.  Is this one ok now?

drpngx:
Jenkins, test this please.

drpngx (approved):

drpngx:
Jenkins, test this please.

DavidNorman:
thanks

drpngx:
Ah, there is a problem.

```
InvalidArgumentError (see above for traceback): TensorArray dtype is float but op has dtype double.
```

caisq:
@DavidNorman, any thoughts on how to fix the test failure that @drpngx pointed out?

DavidNorman:
thinking about it now...  shouldn't be too hard.

DavidNorman:
hmm... that error is what it is meant to do.  it is checking to see if that exception is thrown when the types don't match.   perhaps the regular expression for matching the types isn't quite right.  it is ok on my Mac - but it could be down to different python versions.  i'll check on a linux machine.

DavidNorman:
ok - it really is that the exception regexp match doesn't match on the test machine, but does on my two machines.  am thinking of using [a-z] instead of the character class.  or maybe just match:

"TensorArray dtype is "

DavidNorman:
@drpngx  ok - lets try testing that again.  I've tried all 3 of our various machines and none of them fail.  it must be some sort of python weirdness.

caisq:

DavidNorman:
hmm..  still failing.  It doesn't really make sense.   I think I will have to do a clean build from a fresh checkout to make sure I am not missing something.

I'll get back to you once I have found something

DavidNorman:
@caisq 

oops - i just noticed that my last change didn't actually merge the RE change that I was hoping for.   Can we try once again please?

hopefully the problem is that the CI system python doesn't like '\c' as a regexp thing.

drpngx:
Jenkins, test this please.

DavidNorman:
phew - at least this time it was a completely different build problem.   presumably it has gone away by now...

DavidNorman:
and thankfully the XLA test passed.   It must have been something about regexp support in python (perhaps not accepting \c as a character class) - or maybe a failure in my escaping (?)

anyway - hope that the GPU tests go through on the next try.

drpngx:
Jenkins, test this please.

DavidNorman:
@drpngx hooray - thanks goodness :)

drpngx:
Yay!

Issue #11082, opened 2017-06-27
Only use weakref.finalize from backports in Python < 3.4


mdeff:
### System information
- Arch Linux (up-to-date)
- TensorFlow installed from binary (with pacman, Arch package manager)
- TensorFlow version: 1.2.0
- No GPU or CUDA

### Describe the problem
[tensorflow/python/util/tf_should_use.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/tf_should_use.py) is doing `from backports import weakref` (introduced by    cf238e1f2f68309822e1adb3f86dd439c0b87441), though I guess this is only useful for Python 2. In Python 3, we could simply `import weakref`. This bug has been observed by other people and reported on the [Arch Linux bug tracker](https://bugs.archlinux.org/task/54606).

### Traceback
```
python -c "import tensorflow"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/tensorflow/__init__.py", line 24, in <module>
    from tensorflow.python import *
  File "/usr/lib/python3.6/site-packages/tensorflow/python/__init__.py", line 63, in <module>
    from tensorflow.python.framework.framework_lib import *
  File "/usr/lib/python3.6/site-packages/tensorflow/python/framework/framework_lib.py", line 100, in <module>
    from tensorflow.python.framework.subscribe import subscribe
  File "/usr/lib/python3.6/site-packages/tensorflow/python/framework/subscribe.py", line 26, in <module>
    from tensorflow.python.ops import variables
  File "/usr/lib/python3.6/site-packages/tensorflow/python/ops/variables.py", line 26, in <module>
    from tensorflow.python.ops import control_flow_ops
  File "/usr/lib/python3.6/site-packages/tensorflow/python/ops/control_flow_ops.py", line 70, in <module>
    from tensorflow.python.ops import tensor_array_ops
  File "/usr/lib/python3.6/site-packages/tensorflow/python/ops/tensor_array_ops.py", line 33, in <module>
    from tensorflow.python.util import tf_should_use
  File "/usr/lib/python3.6/site-packages/tensorflow/python/util/tf_should_use.py", line 28, in <module>
    from backports import weakref  # pylint: disable=g-bad-import-order
ModuleNotFoundError: No module named 'backports'
```

spinnau:
[...]

PR #11174, opened 2017-06-30
Only use weakref.finalize from backports in Python < 3.4

mdeff:
The `backports` module should not be forced as a dependency for Python >= 3.4 as `weakref.finalize` has been introduced in Python 3.4. This solves #11082 and an [Arch Linux bug](https://bugs.archlinux.org/task/54606).

### DIFF
--- a/tensorflow/python/util/tf_should_use.py
+++ b/tensorflow/python/util/tf_should_use.py
@@ -25,7 +25,12 @@ import types
 
 import six  # pylint: disable=unused-import
 
-from backports import weakref  # pylint: disable=g-bad-import-order
+# pylint: disable=g-bad-import-order
+try:
+  from weakref import finalize
+except ImportError:
+  from backports.weakref import finalize
+# pylint: enable=g-bad-import-order
 
 from tensorflow.python.platform import tf_logging
 from tensorflow.python.util import tf_decorator
@@ -107,7 +112,7 @@ def _add_should_use_warning(x, fatal_error=False):
       # garbage collected.  Can't add self as the args because the
       # loop will break garbage collection.  We keep track of
       # ourselves via python ids.
-      weakref.finalize(self, _deleted, self._tf_ref_id, fatal_error)
+      finalize(self, _deleted, self._tf_ref_id, fatal_error)
 
     # Not sure why this pylint warning is being used; this is not an
     # old class form.
--- a/tensorflow/tools/pip_package/setup.py
+++ b/tensorflow/tools/pip_package/setup.py
@@ -35,7 +35,6 @@ REQUIRED_PACKAGES = [
     'numpy >= 1.11.0',
     'six >= 1.10.0',
     'protobuf >= 3.2.0',
-    'backports.weakref == 1.0rc1',
     'tensorflow-tensorboard',
 ]
 
@@ -53,6 +52,10 @@ else:
   REQUIRED_PACKAGES.append('wheel')
   # mock comes with unittest.mock for python3, need to install for python2
   REQUIRED_PACKAGES.append('mock >= 2.0.0')
+
+# weakref.finalize was introduced in Python 3.4
+if sys.version_info < (3, 4):
+  REQUIRED_PACKAGES.append('backports.weakref >= 1.0rc1')
 
 # pylint: disable=line-too-long
 CONSOLE_SCRIPTS = [


ebrevdo:
LGTM but do we need to disable the third_party build rules for python3 to fix the arch Linux issue?  Either way this PR should go in if all tests pass.

ebrevdo (approved):

mdeff:
I don't know what you mean by "the third_party build rules for python3" but this change fixes the Arch Linux issue (which uses Python 3.6). The problem is that the `python-tensorflow` package depends on the `backports-weakref` package which is not listed in the requirements. BTW shall we declare `backports-weakref` as a requirement for Python < 3.4 somewhere?

drpngx:
Jenkins, test this please.

ebrevdo:
I thought it's in the pip setup.py as a requirement?

polyzen:
@ebrevdo, https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py#L38

Seems like [this](https://github.com/sphinx-doc/sphinx/pull/3789#issuecomment-303943521) would work.

spinnau:
In [arch linux PKGBUILD](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/tensorflow#n104) the `pip install` command will be executed with `--no-dependencies` option. All the dependencies will be packaged separately and handled/installed by the systems package manager pacman. 

Nevertheless, it makes sense to restrict the `backports.weakref` requirement to Python < 3.4 in the setup.py file, to avoid unneccessary installation of this package also on other platforms.

ebrevdo:
Feel free to add it to this PR.

mdeff:
Indeed most package managers store the dependencies on their own. Let's conditionally depend on `backports.weakref` in the `setup.py` (avoiding clutter for `pip` users). I'll update the PR. Thanks for your comments!

mdeff:
I've updated `setup.py` to only depend on `backports.weakref` for Python < 3.4.

Should we do something about 4ac7cc54c81984e38da7a17992a56f62560e2870? I don't know how Bazel works.

ebrevdo (commented):
Please confirm that the >= version check works as well as the == check did.

mdeff:
There's only one version of `backports.weakref` so yes. I think it's less restrictive for the users in case they release an update.

ebrevdo:
The versioning on that package is funny.  pip won't find it if you don't
put the version number after the package name. That is why I was asking.

spinnau:
@ebrevdo: For me pip finds and installs backports.weakref (version 1.0rc1) just fine without giving any version number:
 
```bash
$ pip --version
pip 9.0.1 from /usr/lib/python3.6/site-packages (python 3.6)

$ pip install --user backports.weakref
Collecting backports.weakref
  Using cached backports.weakref-1.0rc1-py3-none-any.whl
Installing collected packages: backports.weakref
Successfully installed backports.weakref-1.0rc1
```

The `>=1.0rc` version check from the setup.py creates the entry `Requires-Dist: backports.weakref (>=1.0rc1)` inside the wheels METADATA file. Installing the package and pulling the dependency also works fine for me with this using pip.

According to this section in [PEP440](https://www.python.org/dev/peps/pep-0440/#inclusive-ordered-comparison) the inclusive ordered comparison `>=` can be used with the standard [version scheme](https://www.python.org/dev/peps/pep-0440/#version-scheme) and thus, also includes pre-release specifiers like `rc`.

ebrevdo:
Interesting. That command did not work for me on a standard Ubuntu install
with python2.7.

mdeff:
I just checked on Ubuntu 16.04 with Python 2.7 and it worked fine without giving any version number either:

```
$ pip --version
pip 8.1.1 from /usr/lib/python2.7/dist-packages (python 2.7)

$ pip install --user backports.weakref
Collecting backports.weakref
  Downloading backports.weakref-1.0rc1-py2-none-any.whl
Installing collected packages: backports.weakref
Successfully installed backports.weakref
```

Shall we get rid of the version number altogether?

ebrevdo:
$ cat /etc/lsb-release
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04 LTS"

$ pip --version
pip 1.5.4 from /usr/lib/python2.7/dist-packages (python 2.7)

$ pip install backports.weakref
Downloading/unpacking backports.weakref
  Could not find a version that satisfies the requirement backports.weakref
(from versions: 1.0rc1, 1.0rc1)

$ pip install backports.weakref>=1.0rc1
(does NOTHING - just exits)

$ pip install backports.weakref==1.0rc1
(works!)

so, please keep the string "backports.weakref==1.0rc1", not
"backports.weakref>=1.0rc1" and do not remove the version number.

On Tue, Jul 4, 2017 at 2:06 AM, Michaël Defferrard <notifications@github.com

mdeff:
Okay. Maybe it's due to the old pip 1.5.4... I updated the PR.

spinnau:
> $ pip install backports.weakref>=1.0rc1
(does NOTHING - just exits)

@ebrevdo: The package name and the version should be covered in quotation marks here, as otherwise in the shell the `>` sign redirects the output to file named `=1.0rc1`.

```bash
$ pip --version
pip 1.5.4 from /usr/local/lib/python2.7/dist-packages (python 2.7)

$ pip install --user "backports.weakref>=1.0rc1"
Downloading/unpacking backports.weakref>=1.0rc1
  Downloading backports.weakref-1.0rc1-py2-none-any.whl
Installing collected packages: backports.weakref
Successfully installed backports.weakref
Cleaning up...
```

As I mentioned before, installing the tensorflow wheel package works fine for me if I change the version specifier inside the wheel from `==1.0rc1` to `>=1.0rc1` (unzip wheel file; edit METADATA file in dist-info folder; zip changed package again).

I have also tested this on Ubuntu 14.04.05 LTS with pip 1.5.4, which works fine and downloads the rc1 package as expected:

```bash
$ pip install --user tensorflow-1.2.1-cp27-none-linux_x86_64.whl 
Unpacking ./tensorflow-1.2.1-cp27-none-linux_x86_64.whl
Downloading/unpacking mock>=2.0.0 (from tensorflow==1.2.1)
Downloading/unpacking bleach==1.5.0 (from tensorflow==1.2.1)
  Downloading bleach-1.5.0-py2.py3-none-any.whl
Downloading/unpacking numpy>=1.11.0 (from tensorflow==1.2.1)
  Running setup.py (path:/tmp/pip_build_ubuntu/numpy/setup.py) egg_info for package numpy
    Running from numpy source directory.
    warning: no previously-included files matching '*.pyc' found anywhere in distribution
    warning: no previously-included files matching '*.pyo' found anywhere in distribution
    warning: no previously-included files matching '*.pyd' found anywhere in distribution
Downloading/unpacking backports.weakref>=1.0rc1 (from tensorflow==1.2.1)
  Downloading backports.weakref-1.0rc1-py2-none-any.whl
Downloading/unpacking markdown>=2.6.8 (from tensorflow==1.2.1)
[...]
```

Your suggested change that was already implemented by @mdeff should be ok, but according to PEP404 *"The use of == (without at least the wildcard suffix) when defining dependencies for published distributions is strongly discouraged as it greatly complicates the deployment of security fixes."*

mdeff:
I agree about PEP404: that is why I changed `==` to `>=` in the first place.

@spinnau do you know why `pip install backports.weakref` does not work with pip 1.5.4?

spinnau:
@mdeff: According to [this section of PEP404](https://www.python.org/dev/peps/pep-0440/#handling-of-pre-releases) *"Pre-releases of any kind, including developmental releases, are implicitly excluded from all version specifiers, unless they are already present on the system, explicitly requested by the user, or if the only available version that satisfies the version specifier is a pre-release."* 

Maybe the functionality regarding the last part of that sentence was implemented after pip-1.5.4. You can also get this behavior by explicitly adding the `--pre` option to the pip install command.

But as the `>=` comparison combined with pre-release version specifiers already works in pip-1.5.4, this should not be a problem here.

mdeff:
Thanks @spinnau for your explanations. Back to `>=` then. I hope we can merge now. :)

caisq:
@ebrevdo, @spinnau does this look good to you now?

spinnau:
LGTM

* bazel build of pip-package (wheel) for Python-3.6 works fine on archlinux
* the created Python-3.6 wheel installs and works fine on archlinux
* verified that Python-2.7 wheel with changed dependency in METADATA file (`backports.weakref >= 1.0rc1` instead of `backports.weakref == 1.0rc1`) finds and downloads the backports.weakref package just fine on Ubuntu 14.04 LTS

ebrevdo:
LGTM

caisq:

PR #11177, opened 2017-06-30
[OpenCL] Fixes warning caused by half type

lukeiwanski:
- half requires opencl extension that might not be avaiable on all platforms
  - extends cast operation to cover int8, int16, uint8, uint16


drpngx:
Jenkins, test this please.

benoitsteiner (approved):

PR #11188, opened 2017-06-30
[Java] Add base classes and utilities for operation wrappers.

karllessard:
This pull requests includes a set of basic classes, interfaces and utilities useful for the automatic generation of operations by the upcoming C++ module. They are meant to be used more as internal tools than being part of the public API.


drpngx:
Jenkins, test this please.

tensorflow/java/src/main/java/org/tensorflow/op/Op.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.tensorflow.op;
+
+/**
+ * A marker interface for all operation wrappers.
+ *
+ * <p>Operation wrappers provide strongly typed interfaces for building operations and linking them
+ * into a graph without the use of literals and indexes required by the core classes.
+ *
+ * <p>This interface basically allows keeping references to any operation wrapper using a common type.
```
asimshankar: Nit: Remove "basically"

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
```diff
@@ -0,0 +1,147 @@
...
+ * A helper for wrapping a new operation.
+ *
+ * <p>It provides utilities for wrappers to build a single operation and collect its output(s)
+ * afterward.
+ *
+ * <pre>{@code
+ * public final class Unique extends AbstractSingleOp {
+ *    ...
+ *    public static Unique create(Scope s, Input x) {
+ *        OperationHelper unique = OperationHelper.create(s, "Unique");
```
asimshankar: I'm probably missing something, but do you think we can do away with this `OperationHelper` class and have the generated code invoke the `OperationBuilder` directly? This seems like a relatively thin wrapper over `OperationBuilder` and I'm wondering if reducing the number of similar/related concepts would be beneficial.

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
```diff
@@ -0,0 +1,147 @@
...
+ *        unique.builder().input(x.asOutput());
+ *        return new Unique(unique);
+ *    }
+ *    ...
+ *    private Unique(OperationHelper unique) {
+ *        super(unique.operation());
+ *        y = unique.nextOutput();
```
asimshankar: Perhaps flesh this example out in more detail (what are `y` and `idx` for example).

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * <p>All outputs collected with this method or with {@link #nextOutputList(String)} are retrieved
+   * sequentially, like an iterator would do.
+   *
+   * <p>This method invokes {@link #operation()} internally to build the operation if it has not
+   * been done yet.
+   *
+   * @return an output
+   */
+  public Output nextOutput() {
```
asimshankar: If we do keep `OperationHelper` (see above, I'm tempted to avoid it altogether) - then I'd suggest against a `nextOutput`. I think it will be clearer and more explicit to not rely on the ordering of calls and do something like:

```
y = unique.operation().output(0);
idx = unique.operation().output(1);
```

tensorflow/java/src/main/java/org/tensorflow/op/AbstractSingleOp.java
```diff
@@ -0,0 +1,60 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+package org.tensorflow.op;
+
+import org.tensorflow.Operation;
+
+/** A base class for an {@link Op} that wraps a single operation. */
+public abstract class AbstractSingleOp implements Op {
```
asimshankar: I'm not quite clear on the utility of this abstract class. It seems like a layer of abstraction that can be done away with - the generated `Op` subclasses can use `Operation`, `OperationBuilder`, `Output` etc. directly?

asimshankar (changes requested):
Thanks for the PR.
My primary observation is that I _think_ we can do away with the `OperationHelper` and `AbstractSingleOp` classes as I feel that they add an unnecessary layer of abstraction, and the more layers one has the harder the code is to follow.

But perhaps I'm missing the benefits that will become clearer in future PRs? So happy to discuss.

tensorflow/java/src/main/java/org/tensorflow/op/Inputs.java
```diff
@@ -0,0 +1,47 @@
...
+package org.tensorflow.op;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.tensorflow.Input;
+import org.tensorflow.OperationBuilder;
+import org.tensorflow.Output;
+
+/** Utilities for manipulating input related types and lists. */
```
karllessard: Just to make sure I understood, are you just suggesting to rename `Inputs` utility class to `Operands` or to rename the `Input` interface to `Operand ` as well?

asimshankar: `Input` interface to `Operand`, but this was more seeking your opinion on the suggestion than necessarily saying we must do so :)

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
karllessard: The reason of this helper is that it simplifies the code required to retrieve multiple outputs from some operation. The trivial case is having multiple single outputs, where like you suggested below, this simple code will do the job:
```
y = unique.operation().output(0);
idx = unique.operation().output(1);
```
But it gets complicated when output list are involved, where the length of the list is unknown at compile time. For example, collecting outputs from the `ops.parsing().ParseExample` operation ends up with code way more complex than the previous example.

Now, since the code will be generated, writing complex code might not been an issue. But the intention here is to allow users writing custom ops to benefit from those helper classes, so they can avoid this hassle.

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
karllessard: See previous comment on this, once again the `OperationHelper` can be a useful tool for those who write custom ops.

tensorflow/java/src/main/java/org/tensorflow/op/AbstractSingleOp.java
karllessard: Pretty much the same thing here, this class basically ensures a common equals/hashcode/toString behaviour among single operations, including custom ones.

tensorflow/java/src/main/java/org/tensorflow/op/Inputs.java
karllessard: Ok, got it ;) Well, my opinion on this is pretty much the same as yours: all this `Output/Input` conversion is confusing as both terms sound contradictory but are used to represent the exact same thing, one being a concrete class and the other an interface.

`Operand` is more "neutral" so might effectively result in less brain glitches, I like it. But I think it would be nice to exchange and think more about this, can we address this after as a separate issue?

asimshankar: Seems like both of us are leaning towards `Operand` for now. I'd be tempted to name this `Operand' now and possibly rename to `Input` later rather than the other way around.

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
asimshankar: The vast majority of operations will be simpler ones (i.e, not involve lists). Furthermore, even for the operations that produce lists of tensors as named outputs, it appears the contents of `nextOutputList()` is simple enough that unrolling it will make the custom `Op`s code easier to follow too.

I understand that this is judgement on aesthetics and thus there is no right or wrong, just differing opinions.  I'm of the opinion that optimizing for readability of the code for the common case is worth the few extra lines for list-producing operation wrappers. Furthermore, if a helper is really needed, perhaps it is better placed as a method on the `Operation` class (e.g., `Operation.outputListStart(String name)` or something, that returns the first output index at which the named output's elements begin), instead of having this helper class that is essentially wrapping an `Operation` object and a counter.

Also, the fact that we'll have a public class with a name `OperationHelper` makes me a bit uneasy. Something about the name isn't descriptive enough, which makes me feel that it shouldn't be a public API class.

tensorflow/java/src/main/java/org/tensorflow/op/AbstractSingleOp.java
asimshankar: I see. Makes sense. Though, perhaps a different name. `SingleOp` is a bit confusing, because aren't all `Op` implementations a "single" "op"? :)

I'm guessing you meant to convey the difference between some composite operations and wrappers of "primitive" operations, i.e., operations that have a 1:1 correspondence with a node in the graph? What examples of composite operations did you have in mind? That said, how about `PrimitiveOp` instead? (`AbstractSingleOp` or `SingleOp` is also fine, just proposing an alternative to see if it appeals to you)

Let's also elaborate on the documentation about what "single"/"primitive" mean?

```java
/** Base class for {@link Op} implementations that are backed by a single {@link org.tensorflow.Operation}. */
```

asimshankar (changes requested):

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
asimshankar: The vast majority of operations will be simpler ones (i.e, not involve lists). Furthermore, even for the operations that produce lists of tensors as named outputs, it appears the contents of `nextOutputList()` is simple enough that unrolling it will make the custom `Op`s code easier to follow too.

I understand that this is judgement on aesthetics and thus there is no right or wrong, just differing opinions.  I'm of the opinion that optimizing for readability of the code for the common case is worth the few extra lines for list-producing operation wrappers. Furthermore, if a helper is really needed, perhaps it is better placed as a method on the `Operation` class (e.g., `Operation.outputListStart(String name)` or something, that returns the first output index at which the named output's elements begin), instead of having this helper class that is essentially wrapping an `Operation` object and a counter.

Also, the fact that we'll have a public class with a name `OperationHelper` makes me a bit uneasy. Something about the name isn't descriptive enough, which makes me feel that it shouldn't be a public API class.

tensorflow/java/src/main/java/org/tensorflow/op/Inputs.java
karllessard: Ok perfect, let's do this!

tensorflow/java/src/main/java/org/tensorflow/op/AbstractSingleOp.java
karllessard: Yes, 'primitive' sounds good to me too (and yes, that was the meaning of it)

karllessard: Yes, 'primitive' sounds good to me too (and yes, that was the meaning of it)

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
karllessard: Not a big fan of `OperationHelper` neither... it is kind of a "enhanced operation" that does different things around one operation. Or an "output collector" if we stick to the nextOutput* part.

But I'm not opposed to drop it for now and maybe reintroduce something when custom/default ops will be written. 

May I change your suggestion of `Operation.outputListStart(String name)` to something like `Operation.outputs(int start, int length)` returning a `Output[]`? I think it leaves the `Operation` namespace clean and avoid at least the need to iterate at many places.

asimshankar: That sounds like a good idea. Maybe even `Output[] Operation.outputList(String name)`. Anyway, we can table this for now and return to it in an independent PR later when needed.

karllessard: `Operation.outputList(String)` would require again an `outputListStart` equivalent to know from where to start collecting the outputs of a list. I took a chance and pushed a version of `Operation.outputList(idx, length)` as it will help us to go forward writing the operation generator. I guess it could evolve later if needed. Please tell me if that works with you too.

asimshankar (changes requested):
Thanks for your patience. I think this is almost good to go, just some minor comments about and some chatter about testing the public interfaces instead of the implementation details.

tensorflow/java/src/main/java/org/tensorflow/op/OperationHelper.java
asimshankar: Ah, had missed this. Sounds good.

tensorflow/java/src/main/java/org/tensorflow/Operation.java
```diff
@@ -91,6 +91,21 @@ public int outputListLength(final String name) {
...
   }
 
+  /**
+   * Returns symbolic handles to a list of tensors produced by this operation.
+   *
+   * @param idx index of the first tensor of the list
+   * @param length number of tensors in the list
+   * @return array of {@code Output}
+   */
+  public Output[] outputList(int idx, int length) {
```
karllessard: Ok good. Just to explain my thoughts around this a little more, I agree that `outputList(String name)` is a most intuitive API for getting hold on an output list but in fact, since we collect iteratively the outputs of an operation, the use of the index is more useful. For example, if you have three outputs where the last one is a list, you automatically know that it starts from index 3, and so on.

tensorflow/java/src/main/java/org/tensorflow/Output.java
```diff
@@ -22,10 +22,10 @@
  * <p>An Output is a symbolic handle to a tensor. The value of the Tensor is computed by executing
  * the {@link Operation} in a {@link Session}.
  *
- * <p>By implementing the {@link Input} interface, instances of this class could also be passed
+ * <p>By implementing the {@link Operand} interface, instances of this class could also be passed
```
karllessard: Exactly. And since I'm not even sure that the `.op` package will be included into the Android build, it might be better not to refer one of its class in the `Output` javadoc, what do you think? 
What about "also act as operands to TensorFlow operations" to use the same terminology of the `Operand` documentation?

tensorflow/java/src/main/java/org/tensorflow/op/PrimitiveOp.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * A base class for {@link Op} implementations that are backed by a single {@link Operation}.
+ *
+ * <p>Each operation registered in the TensorFlow core is a primitive and is provided as a {@code
+ * PrimitiveOp}. Custom operations working with only one primitive may also derive from this class.
+ */
+public abstract class PrimitiveOp implements Op {
+
+  @Override
+  public int hashCode() {
```
karllessard: Sounds good to me

tensorflow/java/src/test/java/org/tensorflow/op/PrimitiveOpTest.java
```diff
@@ -0,0 +1,177 @@
...
+import org.tensorflow.Graph;
+import org.tensorflow.Operand;
+import org.tensorflow.Operation;
+import org.tensorflow.Output;
+import org.tensorflow.TestUtil;
+
+public class PrimitiveOpTest {
+
+  @Test
+  public void createOp() {
```
karllessard: You are right, those tests are now relics of the previous implementation (i.e. when the `OperationHelper` was present) and don't make any sense now, I'll remove them. But I admit that I used this test class to prove that the generated primitive ops will work flawlessly, which is not the goal of a unit test so I'll clean that up...

tensorflow/java/src/test/java/org/tensorflow/op/PrimitiveOpTest.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private static class TestOp extends PrimitiveOp implements Operand {
+
+    @Override
+    public Output asOutput() {
+      return output;
+    }
+
+    Output output() {
```
karllessard: Like I mentioned in a previous comment, It is to represent what will be generated automatically. Even if an operation implements `Operand`, it exposes its output with a named method. But that test class might go away anyway...

tensorflow/java/src/test/java/org/tensorflow/op/PrimitiveOpTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private Output output;
+
+    private TestOp(Operation shape) {
+      super(shape);
+      output = shape.output(0);
+    }
+  }
+
+  private static class TestListOp extends PrimitiveOp implements Iterable<Output> {
```
karllessard: This is a BIG question for me lately: should the operation wrappers returns their outputs as `Output` or as `Operand`s? The first is easier to implement and seems more useful while the second seems more logic and consistent among all cases (it gets even trickier with the upcoming `VariableRef`).

Again, we might not decide this now since those test classes might go away but I would like to discuss it with you after this PR, if you're available.

tensorflow/java/src/main/java/org/tensorflow/Output.java
asimshankar: I think it's okay to refer to the `op` package, since that is where this will be used. If you don't care about the `op` package (like in Android distributions perhaps), then you don't need to care about the fact that "it can also act as something you don't care about " :)

tensorflow/java/src/test/java/org/tensorflow/op/PrimitiveOpTest.java
asimshankar: Yup, sounds fine to finalize this later, for this PR we're good.

tensorflow/java/src/main/java/org/tensorflow/Operation.java
asimshankar: Sounds good.

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please.

asimshankar:
(Confirming that it's okay to merge the one commit authored by me in addition to the rest of the commits by the PR author)

frankchn:
Jenkins, test this please.

karllessard:
IMHO, test failures are not related to what was committed by this PR

frankchn:
@karllessard There are some conflicts with `master` right now -- can you rebase?

karllessard:
Thanks @frankchn , rebasing completed.

frankchn:
Jenkins, test this please.

PR #11198, opened 2017-06-30
Enable bitcode compilation

dcirne:
In order to enable bitcode compilation it is necessary more than adding `-fembed-bitcode`. The `MACOSX_DEPLOYMENT_TARGET` variable must be defined and exported as well.

In this commit the variable is defined grabbing the version of the underlying running macOS.

### DIFF
--- a/tensorflow/contrib/makefile/compile_ios_protobuf.sh
+++ b/tensorflow/contrib/makefile/compile_ios_protobuf.sh
@@ -15,8 +15,11 @@
 # ==============================================================================
 # Builds protobuf 3 for iOS.
 
-set -x
 set -e
+
+if [[ -n MACOSX_DEPLOYMENT_TARGET ]]; then
+    export MACOSX_DEPLOYMENT_TARGET=$(sw_vers -productVersion)
+fi
 
 SCRIPT_DIR=$(dirname $0)
 source "${SCRIPT_DIR}/build_helper.subr"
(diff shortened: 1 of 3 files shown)


tensorflow/contrib/makefile/compile_ios_protobuf.sh
```diff
@@ -15,9 +15,10 @@
 # ==============================================================================
 # Builds protobuf 3 for iOS.
 
-set -x
 set -e
 
+export MACOSX_DEPLOYMENT_TARGET=$(sw_vers | grep ProductVersion | awk '{print $NF}')
```
drpngx: Is there a better way to get this? That could be brittle.

dcirne: This can be a string literal like "10.12", or, as written above, it gets the version of the underlying macOS, making it the most compatible solution and future proofing the script.

As an example, running `sw_vers` on my computer I get:
```bash
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.5
BuildVersion:	16F73
```

drpngx: So, we could use `sw_vers -productversion` ?

dcirne:
In hindsight `sw_vers -productVersion` seems obvious. Thanks for the suggestion @drpngx . I have updated the PR with the new code.

drpngx:
Jenkins, test this please.

dcirne:
I updated the commit with a small modification. The script first checks whether `MACOSX_DEPLOYMENT_TARGET` has been defined and exported, if not, then defines and exports it.

This way someone would have the option to populate it with a custom value prior to invoking the script.

drpngx (approved):

drpngx:
Jenkins, test this please.

PR #11207, opened 2017-07-01
Branch 160731226

caisq:


drpngx:
Jenkins, test this please.

drpngx (approved):

drpngx:
Not sure what that is:

```
ERROR: /var/lib/jenkins/workspace/tensorflow-pull-requests-gpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/external/local_config_cuda/cuda/BUILD:1305:1: declared output 'external/local_config_cuda/cuda/lib/libcublas.so.8.0' is a dangling symbolic link.
```

caisq:
@gunan @annarev Any idea where the libcublas-related error quoted by @drpngx above comes from?

gunan:
Not sure what it is, but looks bad.
Internal and external builds are both healthy, so I am not sure what is happening with this one.

drpngx:
Maybe some cache corruption?

Jenkins, test this please.

caisq:
@drpngx I was thinking of the same thing. I wiped the cache of the "tensorflow-pull-requests-gpu" build on all Jenkins GPU instances. Let's see how the new builds go.

caisq:
Looks like wiping the cache gets rid of the previous error. But it appears that the cache of "tensorflow-pull-requests-xla" also needs to be wiped. Doing that right now.

drpngx:
/CC: @jart @davidzchen @rmlarsen in case they know something.

caisq:
hopefully this time all tests will pass

caisq:

caisq:
Both the GPU and XLA builds passed after clearing the caches. So it does look like a cache problem related to libcublas. @davidzchen @meteorcloudy , if you have thoughts, please let us know. 

I'm merging this PR now.

gunan:
Looks like another bazel bug related to caching of external build rules?
@damienmg fyi

damienmg:
That's because we do not track the binary to which we point the symlink too, so if you change the install of cuda, this will fail and needs a reconfigure. We are discussing how to do it correctly in the future already. FTR the same fix that the previous case is feasible (force a reconfiguration of the repo).

gunan:
Thank you very much for the clarification Damien!

Issue #11029, opened 2017-06-24
install with MKL and OpenCL without locate command


shitian-ni:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:Description:	SUSE Linux Enterprise Server 11 (x86_64)
- **TensorFlow installed from (source or binary)**:source
- **TensorFlow version (use command below)**:master branch
- **Bazel version (if compiling from source)**:0.5.1
- **CUDA/cuDNN version**:7.5
- **GPU model and memory**:Tesla K20Xm
- **Exact command to reproduce**:./configure

### Describe the problem
I am trying to install tensorflow on Tokyo Institute of Technology Supercomputer TSUBAME.
I want to install with MKL support, but locate command is required.
The university says that because TSUBAME is a public service and `locate` can be used to see files from other users, the command is not going to be installed. So it means MKL can't be used even it is installed in the supercomputer.
I suggest a feature to install tensorflow with the supports without locate command.
I guess tensorflow is used in a lot of supercomputers, the same problem may occur elsewhere.

### Source code / logs
`Do you wish to build TensorFlow with MKL support? [y/N] y
MKL support will be enabled for TensorFlow
Do you wish to download MKL LIB from the web? [Y/n] n
Please specify the location where MKL is installed. [Default is /opt/intel/mklml]: /usr/apps.sp3/isv/intel/xe2013.1.046/composer_xe_2013_sp1.2.144/mkl
./configure: line 279: locate: command not found`

joshua-xia:
can you use whereis command? or find? In TSUBAME

hotchkiss87:
Locate is not installed by default on some systems, like Gentoo. It's a nice tool to have though (sometimes it's called 'mlocate' though, and it doesn't have a standard, so it's generally not a great idea to use it in scripts.) If you don't want locate though, you can manually change the line to this:
```bash
loc="$(find / -mount -name 'libdl.so.2' -print 2>/dev/null | head -n1)"
```
It looks like that section is going to change a lot in the future, so it might be better just edit the config instead of merging a change someone else will have to remove.

gunan:
MKL build should now not depend on locate command, if you are building from master branch.

PR #9580, opened 2017-05-01
Enable MKL in configure and various bug fixes

vivek-rane:
This pull request enables MKL in configure. Building with bazel using `--config=mkl` will utilize all our ops for existing frameworks, replaced via the graph pass. 

Various bug fixes include:
- Performance fix for relu/maxpool (grad)
- Graph layout pass changes
- LRN defaults to eigen when depth radius is not 2
- Various style fixes
- Unit test build failure fixes
- Fixed missing _ in name that was missed during previous search & replace by Google

... and other minor changes.

PR #11209, opened 2017-07-01
fallback to find if locate is not used

hotchkiss87:
Related to #11029 and discussion in #9580

Locate is non-standard, and not installed on every system. It is much faster than 'find', but 'find' should be on every system. Find is even in busybox. On extremely large systems, it can potentially run for a long period of time.

This fix will first attempt to use the locate command, piping any error messages to /dev/null. If locate did not generate output, the script will warn that it will attempt to use 'find', and then use 'find'.

When using find, piping stderr to /dev/null is required since 'find' will generate errors when it cannot access a file (for instance if permissions are enforced). Piping the output from 'find' into head -n1 will complete the line of code, and thus stop 'find' from running longer.

### DIFF
--- a/configure
+++ b/configure
@@ -264,19 +264,27 @@ if [ "$TF_NEED_MKL" == "1" ]; then # TF_NEED_MKL
       exit 1
   fi
 
+  loc="$(locate -e libdl.so.2 2>/dev/null | sed -n 1p)"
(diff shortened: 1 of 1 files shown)


hotchkiss87:
Closing due to other work on the section for now. It looks like this won't be needed in the future.

PR #11212, opened 2017-07-02
Tidy up MKL configuration.

gunan:
Also only use "--config=mkl" to build with MKL support.

### DIFF
--- a/tensorflow/core/kernels/mkl_tfconv_op.cc
+++ b/tensorflow/core/kernels/mkl_tfconv_op.cc
@@ -29,8 +29,8 @@ limitations under the License.
 #include "tensorflow/core/util/tensor_format.h"
 
 #include "tensorflow/core/util/mkl_util.h"
-#include "third_party/mkl/include/mkl_dnn.h"
-#include "third_party/mkl/include/mkl_dnn_types.h"
+#include "mkl_dnn.h"
+#include "mkl_dnn_types.h"
 
 namespace tensorflow {
 typedef Eigen::ThreadPoolDevice CPUDevice;
--- a/tensorflow/tools/pip_package/build_pip_package.sh
+++ b/tensorflow/tools/pip_package/build_pip_package.sh
@@ -17,6 +17,10 @@
 
 set -e
 
+function real_path() {
+  [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
+}
+
 function cp_external() {
   local src_dir=$1
   local dest_dir=$2
@@ -41,7 +45,7 @@ function main() {
     exit 1
   fi
 
-  DEST=$(realpath $1)
+  DEST=$(real_path $1)
   TMPDIR=$(mktemp -d -t tmp.XXXXXXXXXX)
 
   GPU_FLAG=""
--- a/tensorflow/tools/pip_package/build_pip_package.sh
+++ b/tensorflow/tools/pip_package/build_pip_package.sh
@@ -41,7 +41,7 @@ function main() {
     exit 1
   fi
 
-  DEST=$1
+  DEST=$(realpath $1)
   TMPDIR=$(mktemp -d -t tmp.XXXXXXXXXX)
 
   GPU_FLAG=""
@@ -90,11 +90,11 @@ function main() {
       "${TMPDIR}/external"
     RUNFILES=bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles
     # Copy MKL libs over so they can be loaded at runtime
-    if [ -d bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/org_tensorflow/_solib_k8/_U_S_Sthird_Uparty_Smkl_Cintel_Ubinary_Ublob___Uthird_Uparty_Smkl ]; then
+    so_lib_dir="bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/org_tensorflow/_solib_k8"
+    mkl_so_dir=$(ls ${so_lib_dir} | grep mkl)
+    if [ $? -eq 0 ]; then
       mkdir "${TMPDIR}/_solib_k8"
-  		cp -R \
-  			bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/org_tensorflow/_solib_k8/_U_S_Sthird_Uparty_Smkl_Cintel_Ubinary_Ublob___Uthird_Uparty_Smkl \
-        "${TMPDIR}/_solib_k8"
+      cp -R ${so_lib_dir}/${mkl_so_dir} "${TMPDIR}/_solib_k8"
     fi
   else
     if [ -d bazel-bin/tensorflow/tools/pip_package/build_pip_package.runfiles/org_tensorflow/external ]; then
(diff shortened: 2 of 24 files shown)


gunan:
@tfboyd Could you get reviewers from Intel to look at this PR? Also, if we have any documentation for mkl, it will need to be updated..

@allenlavoie @jart Looks like my bazel configs were correct, but build_pip_package script had some hardcoded references to mkl libraries.

gunan:
@tfboyd, this PR blocks my further work on having MKL on windows and macos.
However, this PR should be sufficient to build TF with both MKL support and cuda.

gunan:
windows cmake tests have an issue, but they are known issues due to a bug in `tf.acosh`

tensorflow/core/kernels/mkl_tfconv_op.cc
```diff
@@ -29,8 +29,8 @@ limitations under the License.
 #include "tensorflow/core/util/tensor_format.h"
 
 #include "tensorflow/core/util/mkl_util.h"
-#include "third_party/mkl/include/mkl_dnn.h"
-#include "third_party/mkl/include/mkl_dnn_types.h"
+#include "mkl_dnn.h"
```
drpngx: Why this change?

gunan: previously all the headers were manually copied to third_paarty/mkl/include.
However, now we use "includes" field of the cc_library rule "intel_binary_blob", which adds the include folder for mkl to the gcc commands as a "-iquote" flag. Then all these libraries become available to include without the need for a relative path.

Wang-Yue: Don't do that. This will cause various issues. Always use full include path in tensorflow code please.

gunan: Could you elaborate on 'various issues'?
It worked as is before, because of all the non-standard things that were done in configure. This is what we do for any other external dependency headers, libjpeg, libpng, etc. The path here was just a hand fabricated path due to the way things were copied in `confiure`

Wang-Yue: In Google projects, always use full include path for C++/ObjC programs if you can. That's the convention. 
If you have no other options, there's a couple of things you can do
- guard the full path inclusion for google use (PLATFORM_GOOGLE), and partial path for others. 
- in addition, you should use <lib.h> rather than "lib.h" whenever possible.

gunan: That convention is for internal includes. Anything that is a part of tensorflow. If something is an external dependency, we point our build rules to their include folder:
https://github.com/tensorflow/tensorflow/blob/master/third_party/png.BUILD#L31

Then include them directly in our files. This is what I am doing here. Getting rid of brittle downloading through configure script mechanism end up moving us to here.
Unless we check in the headers into our third_party/mkl folder, we cannot make them accessible through the `third_party/mkl/include`path.

Moving everything into a single header is something I had in mind, but that will come in a later PR.

tensorflow/tools/pip_package/build_pip_package.sh
```diff
@@ -17,6 +17,10 @@
 
 set -e
 
+function real_path() {
```
drpngx: I think `readlink -f` should work on modern systems.

tensorflow/core/kernels/mkl_tfconv_op.cc
drpngx: Gotcha.

third_party/mkl/mkl.BUILD
```diff
@@ -0,0 +1,22 @@
+licenses(["restricted"])  # MPL2, portions GPL v3, LGPL v3, BSD-like     TODO
```
drpngx: The web site says Apache... Could you check or get someone to check?

agramesh1:
One reason we had added MKL download in the config file was to enable the user to be able to point to point to a local copy instead of downloading it. Is it possible to accomplish the same if the download is  through bazel? 

Other than this the changes look fine.

gunan:
@jart, how can we enable users to use their own copies of MKL with this setup?

@agramesh1 Can you provide input on the license question @drpngx added on the mkl.BUILD file?

third_party/mkl/mkl.BUILD
agramesh1: MKL ML is distributed under 3-clause BSD.

drpngx (approved):

drpngx:
```
FAIL: mismatch in packaged licenses and external dependencies
Missing the licenses for the following external dependencies:
//third_party/mkl
@mkl//
```

drpngx:
(Removing "awaiting testing then merge", since @jart might have some input)

tensorflow/tools/pip_package/build_pip_package.sh
gunan: When I try that on macOS, I see this:
```
gunan$ readlink -f .
readlink: illegal option -- f
usage: readlink [-n] [file ...]
```

readlink -f seems to be OK with GNU coreutils, but it does not seem to work with BSD coreutils.

gunan:
Resolved license issue. 
I will tidy things up further when adding macOS support by removing third_party/mkl directory.

@agramesh1 looks like as @drpngx pointed out the license in MKL girhub repository is apache.
You might want to update the license there.

gunan:
@jart do you have any comments on BUILD files.

gunan:
Ping! Any objections to this PR?

agramesh1:
@gunan  I will get MKL team to update license on github repository.  Regarding the PR,  we want be sure there is a way for users to use their own copy of MKL. Awaiting response from @jart

gunan:
Current PR removes straightforward support for working with your local copy of MKL.
If that is a vital feature you need, how about I add that with a subsequent PR?

tensorflow/tools/pip_package/build_pip_package.sh
jart: Why do we need a realpath function?

tensorflow/workspace.bzl
```diff
@@ -172,6 +172,17 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
   )
 
   native.new_http_archive(
+      name = "mkl",
+      urls = [
+          "https://github.com/01org/mkl-dnn/releases/download/v0.7/mklml_lnx_2018.0.20170425.tgz",
```
jart: Mirrored.

tensorflow/workspace.bzl
```diff
@@ -172,6 +172,17 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
...
+          "http://mirror.bazel.build/github.com/01org/mkl-dnn/releases/download/v0.7/mklml_lnx_2018.0.20170425.tgz",
```
jart: The downloader is designed to *always* try the first URL in the list, and then begins to fail over in parallel to the second URL after like 2000ms. So the fastest URL should always be the first URL.

gunan: Done

tensorflow/workspace.bzl
gunan: thank you very much!

gunan:
real_path() addition is sort of irrelevant to this change, but when experimenting I discovered a bug in build pip package. If I ran build_pip_package script with a relative path, the whl file just got lost. I wanted to fix that in this PR.
Would you like me to move it to a new PR?

gunan:
I have prepared the local mkl configuration here:  https://github.com/gunan/tensorflow/tree/local_mkl
Once this PR is merged, I will create a new pull request.
@agramesh1 the model I have will be this:
-configure script won't ask anything to the users about MKL.
-using `--config=mkl` will be sufficient to build a pip package with MKL support.
-If you use the environment variable `TF_MKL_ROOT`, `--config=mkl` will use that mkl instead of downloading it from the internet.
Does the above sound good to you?

agramesh1:
@gunan Yes, that will work. Thanks.

jhseu:
As discussed with @gunan, will submit internally to make this easier to merge. Closing.

PR #11213, opened 2017-07-02
Disable flaky cwise_ops_test due to broken acosh on windows.

gunan:


drpngx (approved):

guschmue:
I think this started  6/28 and I see a few other tests failing on windows around the same time. 
Can take a look if we can fix the root cause.

drpngx:
Thanks! The cosh was introduced this week. I can find the pointer tomorrow

On Jul 2, 2017 10:37 PM, "Guenther Schmuelling" <notifications@github.com>
wrote:

I think this started 6/28 and I see a few other tests failing on windows
around the same time.
Can take a look if we can fix the root cause.

—
You are receiving this because you modified the open/close state.
<https://github.com/tensorflow/tensorflow/pull/11213#issuecomment-312553315>,
or mute the thread

gunan:
I have root caused the problem to https://github.com/tensorflow/tensorflow/pull/10598
However, as this was blocking most submissions, I have elected to disable the test.
I am happy to quickly reenable it once we have a fix for acosh.
Also, @guschmue any input on the discussion about what maybe wrong with acosh is welcome!

guschmue:
Ah, makes perfect sense.

PR #11220, opened 2017-07-02
updated for Estimators and input_fn

yufengg:


yufengg:
made changes on my branch that it turns out were already done by the time I got around to merging... So it goes.

PR #11234, opened 2017-07-03
add a new config option sycl_nodouble for SYCL build

guoyejun:
When TF is built with SYCL enabled, the SYCL device code is generated
at build time. Currently, all the data types such as float and double
are registered to generate the device code.

The SYCL device code is compiled into SPIR at build time, and then
passed to OpenCL implemenation at runtime. Since double precision is
an optional feature in the OpenCL spec, it is possible that an OpenCL
implemenation does not support double.

To make some platforms without double support work, this new config
option disables double register for SYCL device code.

This patch just changes the cwise_add operation as an example, and
other operations will be changed in future small patches one by one.

### DIFF
--- a/tensorflow/core/kernels/cwise_op_add_1.cc
+++ b/tensorflow/core/kernels/cwise_op_add_1.cc
@@ -36,11 +36,7 @@ REGISTER_KERNEL_BUILDER(Name("Add")
 
 
 #if TENSORFLOW_USE_SYCL
-REGISTER(BinaryOp, SYCL, "Add", functor::add, float);
-
-#ifndef SYCL_NO_DOUBLE
-REGISTER(BinaryOp, SYCL, "Add", functor::add, double);
-#endif
+REGISTER_FLOATING(BinaryOp, SYCL, "Add", functor::add);
 
 REGISTER_KERNEL_BUILDER(Name("Add")
                             .Device(DEVICE_SYCL)
--- a/tensorflow/core/kernels/cwise_op_add_1.cc
+++ b/tensorflow/core/kernels/cwise_op_add_1.cc
@@ -36,7 +36,8 @@ REGISTER_KERNEL_BUILDER(Name("Add")
 
 
 #if TENSORFLOW_USE_SYCL
-REGISTER_FLOATING(BinaryOp, SYCL, "Add", functor::add);
+#define REGISTER_KERNEL(type) REGISTER(BinaryOp, SYCL, "Add", functor::add, type);
+TF_CALL_SYCL_NUMBER_TYPES(REGISTER_KERNEL);
 
 REGISTER_KERNEL_BUILDER(Name("Add")
                             .Device(DEVICE_SYCL)
--- a/tensorflow/core/kernels/cwise_op_add_1.cc
+++ b/tensorflow/core/kernels/cwise_op_add_1.cc
@@ -36,7 +36,12 @@ REGISTER_KERNEL_BUILDER(Name("Add")
 
 
 #if TENSORFLOW_USE_SYCL
-REGISTER2(BinaryOp, SYCL, "Add", functor::add, float, double);
+REGISTER(BinaryOp, SYCL, "Add", functor::add, float);
+
+#ifndef SYCL_NO_DOUBLE
+REGISTER(BinaryOp, SYCL, "Add", functor::add, double);
+#endif
+
 REGISTER_KERNEL_BUILDER(Name("Add")
                             .Device(DEVICE_SYCL)
                             .HostMemory("x")
--- a/tensorflow/core/kernels/cwise_ops_common.h
+++ b/tensorflow/core/kernels/cwise_ops_common.h
@@ -487,7 +487,6 @@ struct ApproximateEqual<CPUDevice, T> {
   REGISTER(OP, D, N, F, T0)
 #define REGISTER9(OP, D, N, F, T0, T1, T2, T3, T4, T5, T6, T7, T8) \
   REGISTER(OP, D, N, F, T0)
-#define REGISTER_FLOATING(OP, D, N, F) REGISTER(OP, D, N, F, float)
 #else  // !defined(__ANDROID_TYPES_SLIM__)
 #define REGISTER2(OP, D, N, F, T0, T1) \
   REGISTER(OP, D, N, F, T0)            \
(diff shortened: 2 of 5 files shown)


guoyejun:
I signed it!

keryell:
That makes sense.
Is there a place were the compilation flags are summarized?
If so, this new `SYCL_NO_DOUBLE` flag should be documented there.

guoyejun:
yes, it would be better to summarize all the compilation flags in one place. Looks that the most like place for sycl_nodouble would be at https://www.tensorflow.org/install/install_sources. Once sycl build is introduced here, sycl_nodouble can also be introduced here. Anyway, i think it can be done in another patch.

guoyejun:
just curious why the sanity checks are still waiting for status to be reported after 2 days, how could I see the detail of the sanity check to see what is happening. thanks.

lukeiwanski:
Thanks @guoyejun !
@keryell that is an excelent point! In the near future we will be adding something simillar for half ( as it requires an OpenCL extension and might be not supported on all OpenCL platforms ). Indeed https://www.tensorflow.org/install/install_sources seems like good place for that. 

@benoitsteiner / @gunan / @drpngx what do you think? Is this approach acceptable?

third_party/sycl/crosstool/computecpp.tpl
```diff
@@ -12,6 +12,7 @@ CURRENT_DIR = os.path.dirname(sys.argv[0])
 COMPUTECPP_ROOT = CURRENT_DIR + '/../sycl/'
 COMPUTECPP_DRIVER= COMPUTECPP_ROOT + 'bin/compute++'
 COMPUTECPP_INCLUDE = COMPUTECPP_ROOT + 'include'
+SYCL_NO_DOUBLE = os.getenv('SYCL_NO_DOUBLE')
```
drpngx: It's better not to use environment variables since it makes it harder to debug. Prefer to use the bazel.rc below.

guoyejun: good point, i'll use --cxxopt instead of --action_env in tools/bazel.rc, thanks.

caisq:
@drpngx Please take another look?

drpngx:
Jenkins, test this please.

drpngx:
So, there are a lot more double sycl kernels in the code. We'll need the ifdef guard everywhere?

What is the nature of the error? Is it a runtime error or a build error?

guoyejun:
yes, there will be many ifdef to disable double sycl kernels (within the scope of TENSORFLOW_USE_SYCL).

It is a runtime error.  At build time, sycl device compiler (/usr/local/computecpp/bin/compute++) compiles the sycl kernels into SPIR. While at runtime, the SPIR (including double operations) is passed to OpenCL implementation. There will be an issue if the OpenCL implementation does not support double, which is allowed by OpenCL spec.

drpngx:
Another question, is SYCL_NO_DOUBLE used anywhere else but tensorflow?

guoyejun:
SYCL_NO_DOUBLE is only used in tensorflow.
(do you prefer another naming? thanks)

jwlawson:
A lot of the SYCL kernels (though not all of them) are registered using `TF_CALL_GPU_NUMBER_TYPES_NO_HALF` from [register_types.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/register_types.h). This could be changed to a more SYCL specific define (something like `TF_CALL_SYCL_NUMBER_TYPES`?), which can be changed according to whether we want to support double, half etc. This would minimise the changes needed when we want to change the types registered for SYCL.
This doesn't help for the cwise ops though.

drpngx:
The convention seems to be using `TENSORFLOW_SYSCL_NO_DOUBLE` for defines passed from the outside (e.g. `TENSORFLOW_USE_VERBS`). For inside variables, I would recommend a single place that has the `#ifdef` switch and use a macro in the spirit of what @jwlawson suggested. Could you draft that?

guoyejun:
The SYCL kernels are registered with different methods. Some sycl kernels are registered with TF_CALL_float/double, TF_CALL_GPU_NUMBER_TYPES_NO_HALF and TF_CALL_NUMBER_TYPES_NO_INT32 defined in register_types.h. Some are registered with REGISTER* defined in cwise_ops_common.h. And some are registered specially with the macro defined within its own .cc file, for example, fill_functor.cc.  

It's hard to find only one place to handle all the cases (for example, as @jwlawson mentioned, cwise ops cannot benifit from register_types.h ). With spirit of what @jwlawson suggested, we can try to reduce the places of #ifdef as less as possible. 

For all the cwise ops, we can use one #ifdef in cwise_ops_common.h to handle double, and future half float usage. I'll upload the patch. For other sycl kernel categories, we can use the same spirit to reduce the number of #ifdef.

tensorflow/core/kernels/cwise_ops_common.h
```diff
@@ -516,6 +517,15 @@ struct ApproximateEqual<CPUDevice, T> {
 // Instead of adding REGISTER10, etc., shard the .cc files - see
 // cwise_op_equal_to_*.cc for an example.
 
+// add half float when necessary
+#ifdef TENSORFLOW_SYCL_NO_DOUBLE
```
lukeiwanski: In that case it feels like better place for that will be: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/register_types.h#L183

tensorflow/core/kernels/cwise_op_add_1.cc
```diff
@@ -36,11 +36,7 @@ REGISTER_KERNEL_BUILDER(Name("Add")
 
 
 #if TENSORFLOW_USE_SYCL
-REGISTER(BinaryOp, SYCL, "Add", functor::add, float);
-
-#ifndef SYCL_NO_DOUBLE
-REGISTER(BinaryOp, SYCL, "Add", functor::add, double);
-#endif
+REGISTER_FLOATING(BinaryOp, SYCL, "Add", functor::add);
```
lukeiwanski: Assuming that we added macros to (https://github.com/tensorflow/tensorflow/pull/11234/commits/f9845d72f605ae83b7511d7dd39d2c79221f5bc4#r126394733)
like  `TF_CALL_SYCL_NUMBER_TYPES` - that could be somehow composed - depending on the compiler flag - to include / exclude some types ( eg double or half )

we should be able to replace REGISTERx macro with something like:

```
#define REGISTER_KERNEL(type)                                  \
REGISTER(BinaryOp, SYCL, "Add", functor::add, type);
TF_CALL_SYCL_NUMBER_TYPES(REGISTER_KERNEL);
```
Suggested (https://github.com/tensorflow/tensorflow/pull/11234/commits/f9845d72f605ae83b7511d7dd39d2c79221f5bc4#diff-0c5a090f03366bec631c68d29f852ceeR490) will work for cwise ops but other ops (eg. conv, pooling ) would benefit from adding `TF_CALL_SYCL_NUMBER_TYPES` to `register_types.h`

What do you think?

tensorflow/core/kernels/cwise_ops_common.h
guoyejun: sure, no problem

tensorflow/core/kernels/cwise_op_add_1.cc
guoyejun: good point

guoyejun:
any other comments? thanks.

drpngx (approved):

drpngx:
Jenkins, test this please.

Looking forward to the CL that does it on all operators.

PR #11243, opened 2017-07-03
Use custom BUILD file for protobuf

meteorcloudy:
This change won't be needed after updating protobuf to a version
containing google/protobuf@0b059a3


jhseu (commented):
Can we point the rule to that git hash instead?

meteorcloudy:
Yeah, that would be better, but the last time I did, it broke the macOS build.
Anyway, let's try again https://github.com/tensorflow/tensorflow/pull/11317

meteorcloudy:
Closing this because we have #11317

PR #11246, opened 2017-07-03
install cub under external/cub_archive to fix windows gpu build

guschmue:
Fix cmake/windows gpu build. 
install cub under external/cub_archive so  where_op_gpu.cu.cc will find it for cmake and bazel in the same place.
https://github.com/tensorflow/tensorflow/commit/8280e0ae9083a65b23608b34723f07e028a56dc8


drpngx (approved):

drpngx:
Jenkins, test this please.

PR #11251, opened 2017-07-03
Adding generics to the Java API - Phase 1

andrewcmyers:
Here is an initial cut at enhancing the Java API with generics.


drpngx:
Thanks @andrewcmyers ! Please check the CLA.

In the future, if you have large changes, it's best to first file an issue and discuss the implementation plan beforehand.

andrewcmyers:
Thanks, Asim and others are already aware of this effort and there has been a fair amount of discussion about the implementation strategy and interaction with other ongoing efforts. Which is not to say that more work isn't needed.

Technically I don't think I should need the CLA because I am a temp at Google, but I have signed it anyway!

drpngx:
Jenkins, test this please.

asimshankar:
Thanks @andrewcmyers !

(FYI @karllessard @kbsriram @jhseu )

Will take a more detailed look at this, but can we avoid the dependency on `perl` and having to generate source files? In this particular case, I feel that manually repeating some lines for each type in the `Tensor` and related classes is worth the benefit of having the `.java` files be checked in and avoiding a complication in the build step. Or do you disagree? :)

andrewcmyers:
Unfortunately it ends up being more than a few lines repeated, and will get worse once more TF types are supported. It may be quite annoying to maintain without some automation. I know that Karl is aiming to generate the code for ops too.

An alternative strategy that still seems better than hand-editing these four files (DataType.java should also be included, but I have not touched it) is to semi-automate it: scripts are used to generate the files manually, but the generated files are manually copied back to the source tree and checked in as regular source files. Then Perl would not be required to build.

karllessard:
What I was telling to Andrew previously is that the op generator is going to be written in C++ and will have its own templating engine but it would be great to share the same techonology since the final goals are pretty similar, I don't know Asim if you have any good suggestion for this too... 

Also, about copying the generated files into source trees, current solution for the op generator is to generate those files into an external source jar (.srcjar) that will be included into the library build rule.

kbsriram:
Regarding typed helper constructors - if you decide to support them (either as part of this CL or as a breakout change) then in addition to the classes above, at least the Constant and Variable op-wrappers would want to include such blocks to avoid the caller needing to pass in a Tensor that must then be closed. With these we'd have at least four "special-case" classes, so it seems useful to automate their generation as part of the standard build.

As @karllessard notes, we'll be using some form of templating in both C++ and Java. An option might be to re-purpose one or the other to generate these - neither would eliminate the extra build-step of course, but it might minimize the number of template systems in use and avoid a dependency on perl.

andrewcmyers:
I signed yet another CLA.

andrewcmyers:
I signed it! The CLA. Trying to trigger the bot here.

andrewcmyers:
I signed the CLA yet again, after adding all email addresses to my GitHub account. Hopefully that will do the trick.

andrewcmyers:
I signed the CLA under a *third* email address.

drpngx:
Jenkins, test this please.

andrewcmyers:
After discussing with Asim, I have broken the update into 3 phases. The first phase seems to be ready for consideration. It mainly just adds things that are not yet used.

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/perl
```
asimshankar: Having this in perl is a bit worrying, given that most of the TensorFlow maintainers are not perl users :)

So this script will have to be rewritten at some point in Python, but I guess in the mean time having the generated files checked in works out well.

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
```diff
@@ -0,0 +1,37 @@
...
+
+  static final Map<Class<?>, Integer> typeCodes = new HashMap<>();
+  /** Convert to the equivalent DataType. */
+
+  public static DataType dataType(Class<?> c) {
+    Integer code = typeCodes.get(c);
+    if (code == null) {
+      throw new IllegalArgumentException("" + c + " is not a TensorFlow type.");
+    }
+    return DataType.fromC(code.intValue());
```
asimshankar: Wouldn't this require `fromC` to be public? (Since the package for `DataType` is `org.tensorflow`, which is different from the package for this class - `org.tensorflow.types`)?

Alternatively, would it make sense for the map to be:

```
static final Map<Class<?>, DataType> typeMap = new HashMap();
```

instead? (And the `.csv` file can refer to enums in the `DataType`  instead of integers).
So the call in the generated files would look like:

```
static {
  Types.typeCodes.put(TFUint8.class, DataType.UINT8);
}
```

Or perhaps I'm missing something?

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
```diff
@@ -0,0 +1,37 @@
+package org.tensorflow.types;
```
asimshankar: Can we add a `package-info.java` to this package as well? In particular, I'm hoping that anyone who may be confused about why there are seemingly two ways to specify TensorFlow types (`DataType` enum and `Types.TFType` implementations) would read it and come our clearly satisfied. (Or perhaps this documentation belongs with the `Types.TFType` interface, but a `package-info.java` would still be nice)

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
```diff
@@ -0,0 +1,37 @@
...
+ * and a corresponding immutable object (e.g., TFInt32.T) that represents it at run time.
+ */
+public class Types {
+
+  private Types() {} // not instantiable
+
+  static final Map<Class<?>, Integer> typeCodes = new HashMap<>();
+  /** Convert to the equivalent DataType. */
+
+  public static DataType dataType(Class<?> c) {
```
asimshankar: Mind adding Javadoc comments for each of the public methods?

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
```diff
@@ -0,0 +1,37 @@
...
+  static final Map<Class<?>, Object> scalars = new HashMap<>();
+
+  public static Object defaultScalar(Class<?> c) {
+    return scalars.get(c);
+  }
+
+  /**
+   * A marker interface for classes representing Tensorflow types.
+   */
+  public interface TFType {}
```
asimshankar: Would it make sense for this to be in the package instead of being an inner class?
(So `org.tensorflow.types.Type` instead of `org.tensorflow.types.Types.TFType`?)

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -0,0 +1,138 @@
...
+                        ."  }\n";
+        if ($default ne '') {
+            print CLASSFILE
+                         "  static {\n"
+                        ."    Types.scalars.put($tfname.T, $default);\n"
+                        ."  }\n";
+        }
+        print CLASSFILE  "}\n";
+        close(CLASSFILE);
+    } elsif ($option eq '-d') {
```
asimshankar: Remove this block? (We're not generating `DataType.java`, or are you planning to change that?)

At this point, I think the only thing we gain from generating `DataType.java` is the ability to share the short description fragment for each type. I think that the gains from not having this generation machinery for it outweigh the benefits of sharing that comment, so I'd be tempted to avoid generating it.

Also, see comment below about doing away with the integer value in the `.csv` and having `Types.typeCodes` map from `TFType` to `DataType` directly, instead of redirecting through an `Integer`

tensorflow/java/src/main/java/org/tensorflow/types/TFDouble.java
```diff
@@ -0,0 +1,14 @@
...
+
+/** The class TFDouble represents 64-bit double precision floating point. */
+public class TFDouble implements Types.TFType {
+  /** Represents the type TFDouble at run time. */
+  public static final Class<TFDouble> T = TFDouble.class;
+  static {
+    Types.typeCodes.put(TFDouble.T, 2);
+  }
+  static {
+    Types.scalars.put(TFDouble.T, 0);
```
asimshankar: How does the compiler know to interpret this `0` as a double and not an integer?

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
asimshankar: Should this be `Class<? extends TFType>`?
(In which case it won't be possible to have a runtime exception).

Or is this because we were (are?) thinking of `Tensor<Integer>` and thus the needed to use `Types.dataType(Integer.class)` as well?

tensorflow/java/src/gen/perl/tftypes-runall.pl
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/perl
```
asimshankar: My recommendation would be to do away with this script and instead have `tftypes.pl` just generate all the files, always. No flags, no options - it generates all files in place. Thoughts?

asimshankar (changes requested):
Thanks for the breaking this down!

My main gripe is with the use of perl :), but I think we live with that in the short term.

Some other comments/questions.

tensorflow/java/src/gen/perl/tftypes.pl
andrewcmyers: Python would probably look a bit cleaner. I'm not Pythonic enough to do the conversion easily, though.

tensorflow/java/src/gen/perl/tftypes-runall.pl
andrewcmyers: I think if you're actually changing the types, it's very handy to be able to generate one file at a time as you debug it. That functionality could be supported but at the cost of making the script more complex; I think it's useful to keep tftypes.pl as simple as possible.

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -0,0 +1,138 @@
...
+        @{$info[$i]};
+    my $tfname = "TF".$name;
+    my $ucname = uc $name;
+
+    if ($option eq '-t') {
+        if ($jtype eq '') { next }
+        # Generate class declarations
+        # print STDERR "Creating $dirname/$tfname.java\n";
+        open (CLASSFILE, ">$dirname/$tfname.java") || die "Can't open $tfname.java";
+        print CLASSFILE "// GENERATED FILE. Edit tftypes.pl instead.\n";
```
andrewcmyers: Thanks, good point.

tensorflow/java/src/gen/resources/tftypes.csv
```diff
@@ -0,0 +1,21 @@
+TF type,datatype code,Java type,Java boxed type,Creator?,default,Description
```
andrewcmyers: I generally prefer to keep code and data resources separate. I think it *is* less readable if combined with the script, and we lose the functionality of viewing and editing it as a spreadsheet if they are combined.

tensorflow/java/src/gen/resources/Tensors.java.tmpl
```diff
@@ -0,0 +1,19 @@
+package org.tensorflow.op;
```
andrewcmyers: True, good point.

tensorflow/java/src/gen/resources/tftypes.csv
```diff
@@ -0,0 +1,21 @@
...
+Float,1,float,Float,y,0f,32-bit single precision floating point
+Double,2,double,Double,y,0,64-bit double precision floating point
+Int32,3,int,Integer,y,0,32-bit signed integer
+UInt8,4,byte,Byte,n,,8-bit unsigned integer
+Int16,5,,Short,n,(short)0,16-bit signed integer
+Int8,6,,Byte,n,(byte)0,8-bit signed integer
+String,7,byte,Byte,n,,An arbitrary sequence of bytes
+Complex64,8,,,n,,single-precision complex
+Int64,9,long,Long,y,0L,64-bit signed integer
+Bool,10,boolean,Boolean,y,false,Boolean
```
andrewcmyers: Fine. That was just duplicating the way it looked in DataType.

tensorflow/java/src/main/java/org/tensorflow/types/TFBool.java
```diff
@@ -0,0 +1,14 @@
+// GENERATED FILE. Edit tftypes.pl instead.
+package org.tensorflow.types;
+
+/** The class TFBool represents Boolean. */
+public class TFBool implements Types.TFType {
+  /** Represents the type TFBool at run time. */
+  public static final Class<TFBool> T = TFBool.class;
```
andrewcmyers: I wonder if there is value in keeping a level of abstraction here. We might later decide to use something other than the class object as the representative of the type. If we call it "class" then we lose that flexibility.

tensorflow/java/src/main/java/org/tensorflow/types/TFDouble.java
andrewcmyers: Because of the type of TFDouble.

tensorflow/java/src/main/java/org/tensorflow/types/TFBool.java
asimshankar: What kinds of such uses do you foresee? My inclination is to cut down on layers of abstraction where the value isn't clear. In this case, my understanding is that `TFBool.T` is meant to be fed into methods that require a `Class<TFBool>` parameter. Wouldn't that imply that the class object must be the representative of the type? 

Or am I missing something?

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
andrewcmyers: very nice simplifying idea.

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
andrewcmyers: Sure, that seems good.

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
```diff
@@ -0,0 +1,37 @@
...
+    Integer code = typeCodes.get(c);
+    if (code == null) {
+      throw new IllegalArgumentException("" + c + " is not a TensorFlow type.");
+    }
+    return DataType.fromC(code.intValue());
+  }
+
+  static final Map<Class<?>, Object> scalars = new HashMap<>();
+
+  public static Object defaultScalar(Class<?> c) {
```
andrewcmyers: Some ops need a zero value.

tensorflow/java/src/gen/resources/tftypes.csv
asimshankar: Personally, I feel that including it in the script as a datastructure would be easier to follow as it would allow both the definition and its use in the script to use named fields, but I don't feel too strongly right now (and perhaps it's prettier in Python than in Perl), so okay for now. I guess

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
andrewcmyers: Good idea, thanks.

tensorflow/java/src/gen/resources/tftypes.csv
```diff
@@ -0,0 +1,21 @@
+TF type,datatype code,Java type,Java boxed type,Creator?,default,Description
+Float,1,float,Float,y,0f,32-bit single precision floating point
+Double,2,double,Double,y,0,64-bit double precision floating point
+Int32,3,int,Integer,y,0,32-bit signed integer
+UInt8,4,byte,Byte,n,,8-bit unsigned integer
+Int16,5,,Short,n,(short)0,16-bit signed integer
+Int8,6,,Byte,n,(byte)0,8-bit signed integer
+String,7,byte,Byte,n,,An arbitrary sequence of bytes
+Complex64,8,,,n,,single-precision complex
+Int64,9,long,Long,y,0L,64-bit signed integer
```
asimshankar: Is there a use case for the boxed type to be anything other than the Java type with the first letter in upper-case?
(Wondering if we could cut down on the level of redundancy in the `.csv`)

tensorflow/java/src/gen/perl/tftypes-runall.pl
asimshankar: My concern is that the code generation process is more layers for someone who needs to make a change to sift through. Removing option handling form `tftypes.pl` would make it simpler for both a reader and an editor IMHO. One could always use source control to restore unexpected changes to generated files. 

So I believe we agree in keeping `tftypes.pl` as simple as possible, we just have a slight disagreement on what makes it so :). That said, happy to defer to you for now (and will possibly revisit when redoing this in Python)

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -0,0 +1,138 @@
...
+               ."  }\n";
+        }
+      }
+      if ($text =~ m/\b$tfname\b/ || $creat eq 'y') {
+            $imports .= "import org.tensorflow.types.$tfname;\n";
+      }
+      #if ($text =~ m/\b$ucname\b/ || $creat eq 'y') {
+      #  $imports .= "import static org.tensorflow.Types.$ucname;\n";
+      #}
+    } elsif ($option eq '-T') { # Tensor.java
```
asimshankar: I believe we agreed to not generate `Tensor.java`, right? Remove?

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
```diff
@@ -0,0 +1,37 @@
+package org.tensorflow.types;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.tensorflow.DataType;
+
+/**
+ * Utility class for representing TF types as Java types. For each TF type (e.g., int32),
```
andrewcmyers: Done.

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
andrewcmyers: Done.

tensorflow/java/src/main/java/org/tensorflow/types/TFBool.java
andrewcmyers: Agreed, T does not serve enough purpose here.

tensorflow/java/src/gen/resources/tftypes.csv
```diff
@@ -0,0 +1,21 @@
+TF type,datatype code,Java type,Java boxed type,Creator?,default,Description
+Float,1,float,Float,y,0f,32-bit single precision floating point
+Double,2,double,Double,y,0,64-bit double precision floating point
+Int32,3,int,Integer,y,0,32-bit signed integer
+UInt8,4,byte,Byte,n,,8-bit unsigned integer
+Int16,5,,Short,n,(short)0,16-bit signed integer
+Int8,6,,Byte,n,(byte)0,8-bit signed integer
+String,7,byte,Byte,n,,An arbitrary sequence of bytes
```
andrewcmyers: Done.

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
andrewcmyers: Done.

tensorflow/java/src/gen/perl/tftypes-runall.pl
andrewcmyers: I like this idea of rolling this together when it's moved to Python.

tensorflow/java/src/gen/perl/tftypes.pl
andrewcmyers: Done.

tensorflow/java/src/gen/resources/tftypes.csv
andrewcmyers: Simpifying the CSV is appealing, but In general they will be different types. In particular, some types like String do not have a primitive type equivalent. I guess I'd also like to avoid making semantics associations via string hacking.

tensorflow/java/src/gen/perl/tftypes.pl
andrewcmyers: good catch.

andrewcmyers:
@asimshankar thanks for the very thorough and thoughtful review.

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -76,42 +91,51 @@ sub usage {
...
+';
+        my $fulldesc = $desc;
+        if (substr($desc, 0, 1) =~ m/^[aeoiu8]$/i) {
+            $fulldesc = "an $desc"
+        } else {
+            $fulldesc = "a $desc"
+        }
+        print CLASSFILE  "package org.tensorflow.types;\n\n";
+        print CLASSFILE  "/** The class $tfname represents $fulldesc. */\n"
+                        ."public class $tfname implements TFType {\n"
```
asimshankar: Should we also prevent instantiation of these classes?  i.e., add a

```
private $tfname() {}
```
?

tensorflow/java/src/gen/resources/tftypes.csv
```diff
@@ -1,21 +1,21 @@
 TF type,datatype code,Java type,Java boxed type,Creator?,default,Description
-Float,1,float,Float,y,0f,32-bit single precision floating point
-Double,2,double,Double,y,0,64-bit double precision floating point
+Float,1,float,Float,y,0f,32-bit single precision floating point number
+Double,2,double,Double,y,0,64-bit double precision floating point number
```
asimshankar: The datatype code is no longer needed, right? If so, better to remove them - as it removes a possibly unnecessary source of inconsistency?

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -76,42 +91,51 @@ sub usage {
...
+
+';
+        my $fulldesc = $desc;
+        if (substr($desc, 0, 1) =~ m/^[aeoiu8]$/i) {
+            $fulldesc = "an $desc"
+        } else {
+            $fulldesc = "a $desc"
+        }
+        print CLASSFILE  "package org.tensorflow.types;\n\n";
+        print CLASSFILE  "/** The class $tfname represents $fulldesc. */\n"
```
asimshankar: Nit: How about just:

```
/** Represents $fulldesc. **/
```

That makes the index tables in Javadoc more readable without losing information
(https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment)

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -76,42 +91,51 @@ sub usage {
...
+                        ."public class $tfname implements TFType {\n"
                         ."  static {\n"
-                        ."    Types.typeCodes.put($tfname.T, $index);\n"
+                        ."    Types.typeCodes.put($tfname.class, org.tensorflow.DataType.$ucname);\n"
```
asimshankar: Out of curiosity: Why the fully qualified name here instead of an import + `DataType.$ucname` here?
(I don't have a strong opinion either way, I just thought the import would be more idiomatic, but either is fine)

tensorflow/java/src/main/java/org/tensorflow/types/package-info.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Defines classes that represent TensorFlow data types. For each possible data type
+ * that can be used in a tensor, there is a corresponding class in this package that
+ * is used to represent it. For example, the TensorFlow int32 type is represented by
+ * the type TFInt32 and by the class object TFInt32.class. The former is used to
+ * support compile-time checking of tensor data types and the latter is used for
+ * run-time checking of data types. All such classes implement the TFType interface.
+ *`
+ * <p><b>WARNING</b>: The API is currently experimental and is not covered by TensorFlow <a
```
asimshankar: You can remove this, but perhaps talk about the distinction between the classes here and the `{@link org.tensorflow.DataType} enum`?

tensorflow/java/src/main/java/org/tensorflow/types/TFDouble.java
asimshankar: How so? IIUC, `Types.scalars` is a `Map<Class<T>, Object>`, so where is the association between `TFDouble` and the `Object` argument to `put` being made?

tensorflow/java/src/gen/perl/tftypes.pl
```diff
@@ -0,0 +1,158 @@
...
+    print "Usage: tftypes [-ctdT] <type desc file> <tmpl file>\n\n"
+         ."This script generates parts of various .java files that depend on which"
+         ."TensorFlow types are supported by the Java API and how much. For each"
+         ."such .java file, there is a .tmpl file in the same source directory in"
+         ."which the strings \@TYPEINFO\@ and \@IMPORTS\@ are replaced with"
+         ."appropriate Java code. Output code is sent to standard output.\n\n";
+
+    print "Modulo putting in the correct directory names, it can be invoked as follows:\n";
+    print "tftypes -c tftypes.csv Tensors.java.tmpl > Tensors.java\n";
+    print "tftypes -d tftypes.csv DataType.java.tmpl > DataType.java\n";
```
asimshankar: Remove?

tensorflow/java/src/main/java/org/tensorflow/types/TFUInt8.java
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+// GENERATED FILE. To update, edit tftypes.pl instead.
+
+package org.tensorflow.types;
+
+/** The class TFUInt8 represents an 8-bit unsigned integer. */
+public class TFUInt8 implements TFType {
```
asimshankar: Is `TFType.java` missing from the PR?

tensorflow/java/src/gen/perl/tftypes.pl
andrewcmyers: Good point!

tensorflow/java/src/gen/resources/tftypes.csv
```diff
@@ -0,0 +1,21 @@
+TF type,datatype code,Java type,Java boxed type,Creator?,default,Description
+Float,1,float,Float,y,0f,32-bit single precision floating point number
+Double,2,double,Double,y,0,64-bit double precision floating point number
+Int32,3,int,Integer,y,0,32-bit signed integer
+UInt8,4,byte,Byte,n,,8-bit unsigned integer
+Int16,5,,Short,n,(short)0,16-bit signed integer
```
asimshankar: I'm pretty sure that the 'datatype code' is no longer needed, so shall we remove that.
I'm also unclear about whether the boxed type is needed. Is it? 
And perhaps use 'zero value' instead of 'default' for the header line?

andrewcmyers: We could definitely get rid of the boxed type. I thought maybe the datatype code served some documentation function and helped ensure consistency with DataType.

asimshankar: I'd be tempted to remove it - we map from `TFType` to `DataType` and from `DataType` to the integer, so presence in the `.csv` is unnecessary and at worst is a potential source of inconsistency?

andrewcmyers: I am happy to remove it. I guess we were seeing the role of this file a little differently. You are looking at it as something you would like to minimize as much as possible. I was viewing it as a kind of very-lightweight spec, analogous to the proto buf specs, that summarizes how the Java API handles TF data types. A little redundancy in a spec can be good because it allows tools and humans to check for errors, though I still actually think DataType ought to be generated from it so there aren't two specs.

By the way, DataType is currently only visible in the API because Tensor lets you create a Tensor by giving an array of longs and a DataType, because you can ask what DataType a Tensor has, and because the API talks about types in terms of it. Once the ops are expressed in terms of the types package, DataType will become almost completely an implementation detail that is present mainly for backward compatibility. Probably all the javadoc specs should be written in terms of the types package once that happens.

tensorflow/java/src/gen/perl/tftypes.pl
andrewcmyers: You're right, the import is more idiomatic.

tensorflow/java/src/main/java/org/tensorflow/types/TFDouble.java
andrewcmyers: You're right. This was originally a 0.0, but the spreadsheet got rid of the decimal point. Fixed.

tensorflow/java/src/gen/resources/tftypes.csv
andrewcmyers: Done.

tensorflow/java/src/gen/perl/tftypes.pl
andrewcmyers: @kbsriram also suggested this might be done in Java. If there is going to be a master program that generates a lot of Java source code, rather than just a simple largely straight-line script like the one that exists now, then writing it in a statically typed language makes a lot of sense to me. I am not yet clear on how this would interact with the Java code that he is working on for ops.

tensorflow/java/src/gen/perl/tftypes-runall.pl
```diff
@@ -0,0 +1,40 @@
...
+        exit 1;
+    }
+}
+&locchk("$gen");
+&locchk("$tfjavasrc/gen");
+&locchk("$dir/tftypes.pl");
+&locchk("$rsrc/tftypes.csv");
+
+system("perl $dir/tftypes.pl -t $rsrc/tftypes.csv $pkg/types");
+# system("perl $dir/tftypes.pl -c $rsrc/tftypes.csv $rsrc/Tensors.java.tmpl > $pkg/op/Tensors.java");
```
asimshankar: I'm not going to try and convince you that it's okay to not generate `Tensors.java` :). IIUC, these will be simple one-line wrappers and the generation provides little value (other than automating the rank of the array).

But anyway, we can talk about that later :)

asimshankar (commented):
I believe `TFType.java` is still missing from the PR (which would mean that this can't compile). 

Other than that, looks great.
Thanks very much for your patience!

tensorflow/java/src/gen/perl/tftypes-runall.pl
andrewcmyers: I have checked in TFType.java.

andrewcmyers: The value of Tensors (which doesn't really exist in this PR) is that it provides real static checking that Tensor does not. Tensor.create leads to run-time checking. Unfortunately statically checking this requires duplicating all creators for all ranks we want to support, because Java does not support rank polymorphism. Note that Constant and Variable ops will want to work similarly.

asimshankar: I meant it will be okay to not _generate_ `Tensors.java`, but instead have it written by hand. We'll talk :).

asimshankar (approved):
Thanks for your patience. Looking forward to phase 2 :)

asimshankar:
Jenkins, test this please

asimshankar:
Jenkins, test this please

PR #11260, opened 2017-07-04
Add Edward and GPflow to community page

dustinvtran:
Edward is a library for probabilistic programming with a few thousand active users. GPflow is a library for Gaussian processes which also has an active user base.


drpngx (approved):

drpngx:
Jenkins, test this please.

Issue #6116, opened 2016-12-06
grpc RecvTensor is slow


llhe:
I made benchmark tests for distributed setup with loopback network, profiling it and found there is excessive memory copying in the client side of RecvTensor call, which is actually one of the bottleneck.

Here is the code, which mainly stolen from @yaroslavvb [here](https://gist.github.com/yaroslavvb/1124bb02a9fd4abce3d86caf2f950cb2),
```python
  with tf.device(device1):                                                      
    params = tf.get_variable("params", shape=[params_size], dtype=dtype,        
                             initializer=tf.zeros_initializer)                  
  with tf.device(device2):                                                      
    # constant node gets placed on device1 because of simple_placer             
    #    update = tf.constant(1, shape=[params_size], dtype=dtype)              
    update = tf.get_variable("update", shape=[params_size], dtype=dtype,        
                             initializer=tf.ones_initializer())                 
    add_op = params.assign(update)
```

Here is the profiling result (google perftools) with tensor size 100MB (one fact is, the throughput will degrade with the increasing of tensor size):
* [device1 worker profiling report](https://github.com/tensorflow/tensorflow/files/633183/device1-prof.pdf)
* [device2 worker profiling report](https://github.com/tensorflow/tensorflow/files/633184/device2-prof.pdf)

From the result, the sending side (device2) look fine, but the receiving side (device1, the grpc client) consumes too many CPU cycles for the data transfer.

By the way, I made rough stats for this [memmove](https://github.com/grpc/grpc/blob/d7ff4ff40071d2b486a052183e3e9f9382afb745/src/core/lib/support/slice_buffer.c#L278) call. For one round of 100MB tensor assignment, there are roughly 2GB data moved (actually, including the copy inside memmove, it should be 4GB copied with a naive memmove), which is 20+ times RAM bandwidth amplification (the result is an average of 100 round run, which may not precise but the scale should be ok).

llhe:
To make things more clear, I collected more detailed data for [memmove](https://github.com/grpc/grpc/blob/d7ff4ff40071d2b486a052183e3e9f9382afb745/src/core/lib/support/slice_buffer.c#L278):
```
  // int move_size = (sb->count - 1) * sizeof(gpr_slice);
  memmove(&sb->slices[0], &sb->slices[1], (sb->count - 1) * sizeof(gpr_slice));
  // int data_size = GPR_SLICE_LENGTH(slice);
```
A typical `move_size`, `slice_size` sequence,
[...]

PR #11261, opened 2017-07-04
Fix TODO avoiding serialization in gRPC/GPU path

byronyi:
See #10530 and #10531 for the rationale.

### DIFF
--- a/tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc
+++ b/tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc
@@ -347,32 +347,25 @@ void GrpcWorker::RecvTensorAsync(CallOptions* opts,
             if (src_dev->tensorflow_gpu_device_info() && (!on_host)) {
 #if GOOGLE_CUDA
               const DeviceContext* send_dev_context = send_args.device_context;
-              RecvTensorResponse* tmp = new RecvTensorResponse;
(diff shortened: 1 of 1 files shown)


byronyi:
Using a [modified version](https://gist.github.com/byronyi/cb4ba0f6c5815ce78924ebcdd56989b9) of local recv_tensor [benchmark script](https://gist.github.com/yaroslavvb/e196107b5e0afc834652bd3153030c42) (thanks to @yaroslavvb), the performance improvement over current master is shown below:

Current master:
```
Local rate:       23185.76 MB/s
Distributed rate: 196.11 MB/s
```

With this patch:
```
Local rate:       26588.30 MB/s
Distributed rate: 256.12 MB/s
```

yaroslavvb:
why closed? Seems like an improvement

byronyi:
Further investigation shows the l benchmark difference seems to come from different build options, e.g. AVX, SSE, etc. I suspect the difference is only visible after #6116 is fixed, so I'll revisit this PR then.

byronyi:
This should be relevant now as we have the gRPC upgrade finished recently. With this change, it should do less work (mainly serialisation) in principle, but I have not checked.

rmlarsen:

mrry (approved):
Change looks good to me. Thanks for the improvement!

byronyi:
Thank you Derek @mrry :)

PR #11262, opened 2017-07-04
[XLA] Add scaffolding to allow XLA unit tests to run for other devices

DavidNorman:
This is an indentical change to the one in the compiler/tests directory.  It allows devices other than the CPU and GPU to be available targets when running the tests in compiler/xla/tests.

### DIFF
--- a/tensorflow/compiler/xla/tests/plugin.bzl
+++ b/tensorflow/compiler/xla/tests/plugin.bzl
@@ -1,0 +1,28 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Additional XLA devices to be included in the unit test suite."""
+
+plugins = {
+  "poplar": {
+    "deps": [
+      "//tensorflow/compiler/plugin/poplar:poplar_lib",
+      "//tensorflow/compiler/xla/tests:test_macros_cpu",
+    ],
+    "copts": [],
+    "tags": [],
+    "args": []
+  },
+}
+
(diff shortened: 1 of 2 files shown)


drpngx:
Jenkins, test this please.

drpngx:
@tatatodd WDYT?

DavidNorman:
Hi,

I'd like to think that the failure in a python3 front end test isn't anything to do with my change. 

Perhaps we can test again, or see if other public pull requests are suffering from the same problem?

DavidNorman:
So, in terms of the effectiveness of this change, I think it is quite important to have a way for non-CPU/GPU devices to be tested against all of XLA.    My device actually fails quite a few of the tests because they are written with the assumption that devices support the full intersection of CPU/GPU types (which out device currently does not).  I will extend type support on our device to unsigned integer types, eliminating most of the problems, but I will probably have to submit another pull request in the future that allows devices to opt out of double precision float tests.  While our compiler supports doubles, it isn't on the timeline to support them in the graph framework.

drpngx:
Yes, it seems unrelated to your change.

tensorflow/compiler/xla/tests/plugin.bzl
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Additional XLA devices to be included in the unit test suite."""
+
+# Example:
+#
+#plugins = {
+#  "poplar": {
```
tatatodd: Minor suggestion: Perhaps change `poplar` to `foo`, to make it more clear this is the thing you're supposed to fill in.

tensorflow/compiler/xla/tests/plugin.bzl
```diff
@@ -0,0 +1,32 @@
...
+#    "deps": [
+#      "//tensorflow/compiler/plugin/poplar:poplar_lib",
```
tatatodd: Ditto, if you change the above, change this to:
```
  "//tensorflow/compiler/plugin/foo:foo_lib",
```

tatatodd (changes requested):

tatatodd:
@DavidNorman Looks good to me, thanks for the PR!

Just some minor comments; I'll approve and kick off new tests once those are in.

DavidNorman:
thanks - will do ASAP

DavidNorman:
@tatatodd Done - cheers

drpngx:
Jenkins, test this please.

tatatodd (approved):
@DavidNorman Looks great!

Issue #11181, opened 2017-06-30
The 2015 Inception checkpoint gives incorrect results


korrawat:
### Problems
- The 2015 Inception checkpoint file called `classify_image_graph_def.pb`, which can be found in [`inception-2015-12-05.tgz`](http://download.tensorflow.org/models/image/imagenet/inception-2015-12-05.tgz), gives weird results.
- The [tutorial on quantization](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/performance/quantization.md#how-can-you-quantize-your-models) (together with [this one](https://www.tensorflow.org/performance/quantization)) needs corrections.

### Explanation
I originally wanted to try [this tutorial on quantization](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/performance/quantization.md#how-can-you-quantize-your-models) to quantize the model and then check the result with `label_image`.

Using the same name convention as in the tutorial, the quantize step will quantize the original graph, called `classify_image_graph_def.pb`, into a quantized one, called `quantized_graph.pb`. So before checking the latter, I wanted to first `label_image` using this `classify_image_graph_def.pb` and [this sample image](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/label_image/data/grace_hopper.jpg).

The result should be the same as in [this `label_image`'s README](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/label_image/), as shown below. This example, however, uses a newer pre-trained graph, `inception_v3_2016_08_28_frozen.pb`, from [this compressed file](https://storage.googleapis.com/download.tensorflow.org/models/inception_v3_2016_08_28_frozen.pb.tar.gz). In addition to the pb file, the compressed tarball also contains `imagenet_slim_labels.txt`, containing 1,001 lines of label names, e.g. "military uniform".

```bash
# expected result
military uniform (653): 0.834306
mortarboard (668): 0.0218692
academic gown (401): 0.0103579
pickelhaube (716): 0.00800814
bulletproof vest (466): 0.00535088
```

[...]

Issue #10463, opened 2017-06-06
Where are quantization dependencies?


fferroni:
In https://www.tensorflow.org/performance/quantization tutorial, there is //tensorflow/contrib/quantization and //tensorflow/contrib/quantization/kernels required to use a quantized graph.

I cannot find this in https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/quantization 

Has the location been updated?

```
# Note: You need to add the dependencies of the quantization operation to the
#       cc_binary in the BUILD file of the label_image program:
#
#     //tensorflow/contrib/quantization:cc_ops
#     //tensorflow/contrib/quantization/kernels:quantized_ops

bazel build tensorflow/examples/label_image:label_image
bazel-bin/tensorflow/examples/label_image/label_image \
--image=<input-image> \
--graph=/tmp/quantized_graph.pb \
--labels=/tmp/imagenet_synset_to_human_label_map.txt \
--input_width=299 \
--input_height=299 \
--input_mean=128 \
--input_std=128 \
--input_layer="Mul:0" \
--output_layer="softmax:0"
```

michaelisard:
The quantization ops all moved into core tensorflow so you shouldn't need a separate contrib dependency for those ops now. However, I can't vouch for what is needed to make that tutorial work. @petewarden  is anyone maintaining this tutorial?

Androbin:
The tutorial references the outdated `quantize_graph` tool which is a bit buggy now.
This should be updated to make use of the new `transform_graph` tool.

petewarden:
Thanks @androbin! I'll close this bug since that fixes the documentation.

korrawat:
@petewarden So should the code comments about dependencies be removed as well?
```bash
# Note: You need to add the dependencies of the quantization operation to the
#       cc_binary in the BUILD file of the label_image program:
#
#     //tensorflow/contrib/quantization:cc_ops
#     //tensorflow/contrib/quantization/kernels:quantized_ops
```

PR #11285, opened 2017-07-05
Fix quantization tutorial

korrawat:
Fix [this tutorial on quantization](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/performance/quantization.md).

1. Remove outdated dependencies, as pointed out in #10463.

1. The change made in #10592 was incomplete. The [example](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/performance/quantization.md#how-can-you-quantize-your-models) on how to run `transform_graph` to quantize a graph should follow [this tutorial](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms#eight-bit-calculations) about `graph_transforms`, but the argument `--inputs='Mul'` is missing.

@petewarden I'm also wondering when/how will [the tutorial on tensorflow.org](https://www.tensorflow.org/performance/quantization) get updated? I'm assuming it should be the same as `quantization.md` that I'm fixing?

Related to #11181.

Thanks!

### DIFF
--- a/tensorflow/docs_src/performance/quantization.md
+++ b/tensorflow/docs_src/performance/quantization.md
@@ -108,12 +108,6 @@ versus 91MB). You can still run this model using exactly the same inputs and
 outputs though, and you should get equivalent results. Here's an example:
 
 ```sh
-# Note: You need to add the dependencies of the quantization operation to the
-#       cc_binary in the BUILD file of the label_image program:
-#
-#     //tensorflow/contrib/quantization:cc_ops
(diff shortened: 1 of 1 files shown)


drpngx:
@saxenasaurabh does that look right?

Jenkins, test this please.

vrv (approved):

vrv:
This looks correct, and if it's wrong I'm sure we'll get another PR to fix it :P

PR #11292, opened 2017-07-05
Return output_graph_def in tools/freeze_graph.py

carlthome:
It makes sense that `def freeze_graph` works with files for both the input graph and frozen output graph.

It does not make sense that `def freeze_graph_with_def_protos` only uses a tf.GraphDef for the input, but doesn't actually return the frozen graph (and instead only writes it to a file).

This pull request changes `freeze_graph_with_def_protos` to optionally write the output to file (for backwards-compatibility) when `output_graph` is set as usual, but also returns the GraphDef protobuf to make the API more symmetric.

Otherwise you have to write and read from a tempfile just to use freeze_graph, which is a hassle.

### DIFF
--- a/tensorflow/python/tools/freeze_graph.py
+++ b/tensorflow/python/tools/freeze_graph.py
@@ -115,9 +115,14 @@ def freeze_graph_with_def_protos(
         output_node_names.split(","),
         variable_names_blacklist=variable_names_blacklist)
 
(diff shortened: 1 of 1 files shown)


drpngx:
Jenkins, test this please.

petewarden (approved):

Issue #11229, opened 2017-07-02
32-bit build failure in tensorflow/contrib/tensor_forest/kernels/stats_ops.cc


alistairlow:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 14.04.5 i386
- **TensorFlow installed from (source or binary)**: source
- **TensorFlow version (use command below)**: Git revision 744120fd8a0c3be0a90cca5a971459894c90b859
- **Python version**: 2.7.6
- **Bazel version (if compiling from source)**: 0.4.5
- **CUDA/cuDNN version**: N/A
- **GPU model and memory**: N/A
- **Exact command to reproduce**: See log below

### Describe the problem
Building //tensorflow/contrib/tensor_forest:stats_ops_lib for a 32-bit architecture fails,

### Source code / logs
git checkout 744120fd8a0c3be0a90cca5a971459894c90b859
export PYTHON_BIN_PATH=/usr/bin/python
export USE_DEFAULT_PYTHON_LIB_PATH=1
export CC_OPT_FLAGS="-march=native"
export TF_NEED_MKL=0
export TF_NEED_JEMALLOC=1
export TF_NEED_GCP=0
export TF_NEED_HDFS=0
export TF_ENABLE_XLA=0
export TF_NEED_OPENCL=0
export TF_NEED_CUDA=0
export TF_NEED_VERBS=0
export TF_NEED_MPI=0
export COMPUTE=:0
./configure
bazel build -c opt --verbose_failures //tensorflow/contrib/tensor_forest:stats_ops_lib

(removed unimportant log output)

ERROR: /home/codeplay/tensorflow/tensorflow/contrib/tensor_forest/BUILD:277:1: C++ compilation of rule '//tensorflow/contrib/tensor_forest:stats_ops_lib' failed: gcc failed: error executing command 
  (cd /home/codeplay/.cache/bazel/_bazel_codeplay/552bd1ca300856cf615cf243a9219401/execroot/tensorflow && \
  exec env - \
    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games \
    PYTHON_BIN_PATH=/usr/bin/python \
    PYTHON_LIB_PATH=/usr/local/lib/python2.7/dist-packages \
[...]

PR #11302, opened 2017-07-05
Fixed 32-bit build error in tensor_forest.

alistair-low:
This patch fixes the build error described in #11229.

### DIFF
--- a/tensorflow/contrib/tensor_forest/kernels/v4/input_target.h
+++ b/tensorflow/contrib/tensor_forest/kernels/v4/input_target.h
@@ -21,7 +21,7 @@ namespace tensorflow {
 namespace tensorforest {
 
 typedef Eigen::TensorMap<
-    Eigen::Tensor<const float, 1, 1, long>, 0>  // NOLINT(runtime/int)
+    Eigen::Tensor<const float, 1, 1>, 0>  // NOLINT(runtime/int)
     SingleDimStorageType;
 
 // Base class for classes that hold labels and weights. Mostly for testing


alistairlow:
I signed it!

alistairlow:
Fixed the CLA to include my work email in addition to my personal one. I signed it!

gilberthendry:
Thanks for finding this!  The change actually isn't the best solution, so I'll fix this internally.  Should be committed in a day or so.

aloisklink:
Hi @gilberthendry,

I'm not sure if you've fixed this yet, but it seems to me that 
``` c++
typedef Eigen::TensorMap<
    Eigen::Tensor<const float, 1, 1, long>, 0>  // NOLINT(runtime/int)
    SingleDimStorageType;
```
at `tensorflow/contrib/tensor_forest/kernels/v4/input_target.h:23`
should be replaced with:
``` c++
typedef Eigen::TensorMap<
    Eigen::Tensor<const float, 1, 1, Eigen::DenseIndex>, 0>  // NOLINT(runtime/int)
    SingleDimStorageType;
```
to match
``` c++
template <typename T, int NDIMS = 1, typename IndexType = Eigen::DenseIndex>
struct TTypes {
// Rank-<NDIMS> tensor of scalar type T.
typedef Eigen::TensorMap<Eigen::Tensor<T, NDIMS, Eigen::RowMajor, IndexType>,
Eigen::Aligned> Tensor;
```
at `tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor_types.h:24`

Be aware `Eigen::DenseIndex` has been depreciated as of [Eigen 3.3.0](http://eigen.tuxfamily.org/index.php?title=3.3) and should be replaced by `Eigen::Index`, so it might be worth sticking `Eigen::Index` in there instead.

gilberthendry:
Actually I replaced it with tensorflow::TTypes<float, 1>::ConstTensor, which is more portable I think.

aloisklink:
4de7361c43adc65c442359104a17c449a02eee7a and 14101a8ba173b44179b2a9317781f140eb61b0a1 seems to have fixed this issue for me, I managed to successfully compile and run Tensorflow on a 32-bit Raspberry Pi 3. Cheers @gilberthendry for fixing this bug!

alistairlow:
Yes, thanks @gilberthendry! I'm closing my pull request now

Issue #117, opened 2015-11-11
TensorBoard logdir path, if relative, is relative to $HOME


technopolymath:
It took me a while to figure this out and I'm not entirely sure that my interpretation is correct, but even if I'm not in $HOME, running tensorboard --logdir=. will give a list of directories in my home, and passing an absolute path appears to pick things up correctly. It is somewhat counterintuitive to me that relative paths are not relative to the current directory, so maybe that's something to document, disallow, or change?

Thanks, this library is amazing!

danmane:
Thanks for reporting this, I'm looking into it.

pirate:
@danmane Filed a PR for this, should be a simple fix.  I don't have access to a Windows machine to run the test suite at the moment, but I doubt it'll break anything.

pirate:
Thanks @vrv for reviewing, we should be able to close this out now.

vrv:
Indeed, should be fixed by https://github.com/tensorflow/tensorflow/pull/337

PR #11322, opened 2017-07-06
[OpenCL] Fixes SYCL registration for MapStageOp (#117)

lukeiwanski:
* [OpenCL] Fixes SYCL registration for MapStageOp

As the MapStageOp is actually run on the host, we need to ensure that
both the 'key' and 'indices' tensors are in host memory. This now
mirrors what CUDA is doing.

* [OpenCL] Fixes device name comparison map_stage_op_test

Changes the expected device name string, as the device name could be
'/device:SYCL:0' or '/device:GPU:0'.

The CUDA device name reported by test_util is in the form '/gpu:0',
while the device name used in the nodes looks like '/device:GPU:0'. When
running on CUDA we need to change the expected device name accordingly,
however this isn't a problem on SYCL.

### DIFF
--- a/tensorflow/core/kernels/map_stage_op.cc
+++ b/tensorflow/core/kernels/map_stage_op.cc
@@ -550,12 +550,17 @@ REGISTER_KERNEL_BUILDER(Name("OrderedMapStage")
 #endif // GOOGLE_CUDA
 
 #ifdef TENSORFLOW_USE_SYCL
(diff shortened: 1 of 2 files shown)


caisq:

caisq (approved):

caisq:
Thanks, @lukeiwanski

PR #11323, opened 2017-07-06
Minor fix typo

ScorpioCPH:
Minor fix comments typo.

### DIFF
--- a/tensorflow/core/lib/gtl/array_slice.h
+++ b/tensorflow/core/lib/gtl/array_slice.h
@@ -191,7 +191,7 @@ class ArraySlice {
   void pop_front() { remove_prefix(1); }
 
   // These relational operators have the same semantics as the
-  // std::vector<T> relational operators: they do deep (elementwise)
+  // std::vector<T> relational operators: they do deep (element-wise)
   // comparisons.  Array slices are equal iff their size is the same
   // and all their elements are equal.
   bool operator==(ArraySlice<T> other) const { return impl_ == other.impl_; }
--- a/tensorflow/core/protobuf/worker.proto
+++ b/tensorflow/core/protobuf/worker.proto
@@ -141,7 +141,7 @@ message DeregisterGraphResponse {
 message CleanupAllRequest {
   // A list of container names.
   //
-  // If 'container' is not empty, releases resoures in the given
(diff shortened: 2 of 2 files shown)


tensorflow/core/lib/gtl/array_slice.h
```diff
@@ -191,8 +191,8 @@ class ArraySlice {
   void pop_front() { remove_prefix(1); }
 
   // These relational operators have the same semantics as the
-  // std::vector<T> relational operators: they do deep (elementwise)
-  // comparisons.  Array slices are equal iff their size is the same
```
caisq: This is not a typo. "Iff" means "if and only if".

tensorflow/core/protobuf/worker.proto
```diff
@@ -86,7 +86,7 @@ message RegisterGraphRequest {
   // having its device_name filled in.
   GraphDef graph_def = 2;
 
-  // True iff the graph (before partitioning) contains control flow nodes.
```
caisq: Same as above.

caisq (changes requested):

tensorflow/core/lib/gtl/array_slice.h
ScorpioCPH: ok, i will revert it.

ScorpioCPH:
@caisq ping.

tensorflow/core/lib/gtl/array_slice.h
```diff
@@ -191,8 +191,8 @@ class ArraySlice {
   void pop_front() { remove_prefix(1); }
 
   // These relational operators have the same semantics as the
-  // std::vector<T> relational operators: they do deep (elementwise)
   // comparisons.  Array slices are equal iff their size is the same
+  // std::vector<T> relational operators: they do deep (element-wise)
```
caisq: I think this line is moved to the wrong place.

caisq (changes requested):

tensorflow/core/lib/gtl/array_slice.h
ScorpioCPH: You are right, fix it by amend last commit :)

caisq (approved):

PR #11340, opened 2017-07-07
Adapt TensorFlowTestCase.setUp() to new reset_default_graph() semantics

caisq:
Avoid calling reset_default_graph() directly to prevent exceptions in
cases where test methods error out from within nested graph contexts,
which can leave _default_graph_stack non-empty in certain Python
versions.

This should address the test failures (tensorflow/python:session_test and tensorflow/contrib/session_bundle:exporter_test) introduced by #11158 in certain Python versions (e.g., Mac Python 3.5).
cc @Thenerdstation

### DIFF
--- a/tensorflow/python/framework/test_util.py
+++ b/tensorflow/python/framework/test_util.py
@@ -266,7 +266,12 @@ class TensorFlowTestCase(googletest.TestCase):
     self._ClearCachedSession()
     random.seed(random_seed.DEFAULT_GRAPH_SEED)
     np.random.seed(random_seed.DEFAULT_GRAPH_SEED)
-    ops.reset_default_graph()
+    # Note: we avoid calling ops.reset_default_graph() here due to the fact that
+    # under certain Python versions, test methods that error out from within
+    # nested graph contexts may leave ops._default_graph_stack non-empty,
+    # which would cause ops.reset_default_graph() to throw an exception if it
+    # were used in the following line.
+    ops._default_graph_stack.reset()  # pylint: disable=protected-access
     ops.get_default_graph().seed = random_seed.DEFAULT_GRAPH_SEED
 
   def tearDown(self):
--- a/tensorflow/python/framework/test_util.py
+++ b/tensorflow/python/framework/test_util.py
@@ -266,12 +266,14 @@ class TensorFlowTestCase(googletest.TestCase):
     self._ClearCachedSession()
     random.seed(random_seed.DEFAULT_GRAPH_SEED)
     np.random.seed(random_seed.DEFAULT_GRAPH_SEED)
-    # Note: we avoid calling ops.reset_default_graph() here due to the fact that
-    # under certain Python versions, test methods that error out from within
-    # nested graph contexts may leave ops._default_graph_stack non-empty,
-    # which would cause ops.reset_default_graph() to throw an exception if it
-    # were used in the following line.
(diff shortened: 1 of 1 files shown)


caisq:
This should address the test failures (tensorflow/python:session_test and tensorflow/contrib/session_bundle:exporter_test) introduced by #11158 in certain Python versions (e.g., Mac Python 3.5). 
cc @Thenerdstation

drpngx:
Thank you for figuring this out! I wonder why it was Mac only.

Can we call reset default graph after resetting the stack?

caisq:
@drpngx This is not only "mac-only". This seems to be "mac + python 3.5 only", because python 2.7 and 3.6 don't cause those tests to fail on mac.

Calling reset_default_graph() after calling reset() on the stack seems redundant?

drpngx:
Yeah I was just thinking that 😋

drpngx (approved):

caisq:
@drpngx Actually on second thought, I think calling reset_default_graph() after reset() is a good idea. reset_default_graph() may do extra things in the future. Calling it gets rid of the need to track those future changes here. I made some further revision to add that and to make the comments clearer.

tensorflow/python/framework/test_util.py
```diff
@@ -266,6 +266,13 @@ def setUp(self):
     self._ClearCachedSession()
     random.seed(random_seed.DEFAULT_GRAPH_SEED)
     np.random.seed(random_seed.DEFAULT_GRAPH_SEED)
+    # Note: The following line is necessary because some test methods may error
+    # out from within nested graph contexts (e.g., via assertRaises and
+    # assertRaisesRegexp), which may leave ops._default_graph_stack non-empty
+    # under certain versions of Python. That would cause
+    # ops.reset_default_graph() to throw an exception if the stack were not
+    # cleared first.
```
mrry: I'm still confused on why `ops._default_graph_stack` can be non-empty after an exception, since the `finally:` block in the default graph context manager should still execute, right?

If not, is it possible we have a bug in that context manager?

caisq: Good question! This seems strange to me, too. But this is what happens in Mac+Python3.5: see code at: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/client/session_test.py#L1556

This code calls "with sess" twice. By using print debugging, I can see that during the first call, which is legitimate, a Graph object is pushed to the default graph stack and a Session object is pushed to the default session stack. But during the second call, which is illegal and triggers an exception about re-entry, the finally block of _only_ the default session stack is called. That of the default graph stack is not called. This is unique to Mac+Python3.5. In other systems and python versions, the finally blocks of both the default session stack and the default graph stack get called.

Any thoughts?

av8ramit:
Merging to fix nightlies.

caisq:
@mrry We can continue investigating whether there is a bug in the stack finally blocks.

mrry:
Ack, merging was the right way to go... I've looked in the implementation of `contextlib.py` for Python [3.5](https://github.com/python/cpython/blob/3.5/Lib/contextlib.py) and [3.6](https://github.com/python/cpython/blob/3.6/Lib/contextlib.py). While there are some small diffs, I can't see anything that looked relevant. My hunch is that there's something special about `RuntimeError` (since it gets used internally in the `contextlib` implementation), but I didn't find a smoking gun.

PR #11386, opened 2017-07-08
[XLA] Add DT_HALF to the list of floating point types

DavidNorman:
This should not affect the existing CPU and GPU devices, as they do not claim support for D_HALF, and the final registration for each device is the intersection of their advertised list and the global lists.

This merely adds DT_HALF as a kind of floating point number, which will allow MatMul to work with DT_HALF. 

Lots of other arithmetic ops don't restrict themselves to any specific types, so they accept DT_HALF by default.

Why isn't Convolution restricted to kFloatTypes?  I don't see why convolution and matmul are not very similar in backend implementation?

[I can't imagine why I had not spotted this before]


DavidNorman:
@frankchn Would be good to get this one run through the unit tests.  I suspect it will be good, but it is a bit of a risk.

frankchn:
Jenkins, test this please.

DavidNorman:
thanks.  i suspect that the failure to build is something to do with a dependency management error by someone else.

the only file that i have changed is xla_op_registry.h and this is not mentioned.  xla_op_registry.cc is mentioned, but only because of the old favourite 'tensorflow/compiler/xla/service/hlo_instruction.h' producing a signed/unsigned comparison warning.

hopefully if I merge in master then this error will go away.

frankchn:
@DavidNorman The tests have been flaky lately, but can you rebase to the latest master and push again? I am happy to retest it.

@hawkinsp ping on review.

DavidNorman:
@frankchn done :)

frankchn:
Jenkins, test this please.

hawkinsp (approved):

PR #5991, opened 2016-11-30
[Windows] Don't try to read numa node from SysFS

mrry:
This makes the behavior on Windows consistent with Mac OS X, and prevents a needless error message when using GPUs.

Partially addresses #5986.

PR #11392, opened 2017-07-09
GPUDirect RDMA Out-of-Band Tensor Transport

byronyi:
Introduction
===

This PR implements GDR out-of-band transport for TensorFlow distributed runtime, complementary to current gRPC transport. It uses gRPC as control plane to setup rendezvous for each tensor transmission, and utilizes [GPU Direct RDMA](https://developer.nvidia.com/gpudirect) whenever possible to transmit tensors in remote GPU memory through network interface card (NIC), bypassing host memory and CPU entirely. It gracefully falls back to ordinary RDMA or even gRPC when GDR is not available.

Design
===

The GDR out-of-band transport is designed to avoid any unnecessary memory copies, especially for large tensors (>100MB). That typically requires registration of tensor buffers to NIC on the fly, which is rather slow as described in the design trade-off of the verbs runtime. The verbs runtime thus chooses to manage its own NIC-registered buffers and copy the tensors from/to those buffers for every single tensor transfer.

We show that, however, such design trade-off is not always relevant. In this patch, we manage both computation and communication buffers in a unified manner. By pre-registration of large buffers to NIC and allocating small tensors from the buffer pool using a BFC allocator, it is possible to avoid both buffer registration on the fly and memory copies all together.

For the actual tensor transport, we rely on gRPC to transmit the remote buffer information. This greatly simplifies our design, and there are only 2 types of RDMA messages: a single READ to retrieve the tensor data (bypassing remote CPU), and another invalidate using WRITE with IMM to release the tensor buffer on the remote side. The remote side will only be polling the invalidate message and `Unref` the tensor buffers that read by its peer.

Environment
===

To fully utilize GDR, the target environment has to meet 3 conditions:

1. There is an RDMA capable device with corresponding [OFED package](https://www.openfabrics.org/index.php/overview.html) installed (detailed information is available from your [Infiniband/RoCE](http://www.mellanox.com/page/products_dyn?product_family=116)/[iWarp](http://www.chelsio.com/gpudirect-rdma/) vendor), which could be verified through `ibv_devinfo`, e.g.

```
$ ibv_devinfo
hca_id:	mlx4_0
	transport:			InfiniBand (0)
	fw_ver:				2.40.7000
	node_guid:			248a:0703:00f6:3370
	sys_image_guid:			248a:0703:00f6:3370
	vendor_id:			0x02c9
	vendor_part_id:			4099
	hw_ver:				0x1
	board_id:			MT_1090110023
	phys_port_cnt:			2
	Device ports:
		port:	1
			state:			PORT_ACTIVE (4)
			max_mtu:		4096 (5)
			active_mtu:		1024 (3)
			sm_lid:			0
			port_lid:		0
			port_lmc:		0x00
			link_layer:		Ethernet

		port:	2
			state:			PORT_ACTIVE (4)
			max_mtu:		4096 (5)
			active_mtu:		1024 (3)
			sm_lid:			0
			port_lid:		0
			port_lmc:		0x00
			link_layer:		Ethernet
```

2. There is a GDR capable GPU, i.e. of Fermi, Kepler or later architecture with [corresponding driver](http://docs.nvidia.com/cuda/gpudirect-rdma/index.html) installed. The PCI-e topology could be confirmed by `nvidia-smi topo -m`. For example, in the following topology, `GPU2` and `GPU3` are adjacent to `mlx4_0`, and tensors on these devices could benefit from GDR in current implementation.

```
$ nvidia-smi topo -m
	GPU0	GPU1	GPU2	GPU3	mlx4_0	CPU Affinity
GPU0	 X 	PHB	SOC	SOC	SOC	0-5
GPU1	PHB	 X 	SOC	SOC	SOC	0-5
GPU2	SOC	SOC	 X 	PHB	PHB	6-11
GPU3	SOC	SOC	PHB	 X 	PHB	6-11
mlx4_0	SOC	SOC	PHB	PHB	 X

Legend:

  X   = Self
  SOC  = Connection traversing PCIe as well as the SMP link between CPU sockets(e.g. QPI)
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe switches (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing a single PCIe switch
  NV#  = Connection traversing a bonded set of # NVLinks
```

3. The [`nv_peer_mem`](https://github.com/Mellanox/nv_peer_memory) kernel module is installed.

How to build and run in GDR mode
===

To test it out on a GDR capable environment, choose to enable GDR in your configure script.

```
Do you wish to build TensorFlow with GDR support? [y/N]: y
GDR support will be enabled for TensorFlow.
```

Change your `protocol` to `grpc+gdr` to enable GDR in your deployment.

```
server = tf.train.Server(cluster, job_name="local", task_index=0, protocol='grpc+gdr') # default protocol is 'grpc'
```

Currently the out-of-band transport service listens to the same IP and port address as specified in gRPC.

A successful initialization looks like this:

```
2017-08-05 19:10:38.601718: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1045] Creating TensorFlow device (/gpu:0) -> (device: 0, name: Tesla K40m, pci bus id: 0000:02:00.0)
2017-08-05 19:10:38.601728: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1045] Creating TensorFlow device (/gpu:1) -> (device: 1, name: Tesla K40m, pci bus id: 0000:03:00.0)
2017-08-05 19:10:38.601736: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1045] Creating TensorFlow device (/gpu:2) -> (device: 2, name: Tesla K40m, pci bus id: 0000:82:00.0)
2017-08-05 19:10:38.601742: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1045] Creating TensorFlow device (/gpu:3) -> (device: 3, name: Tesla K40m, pci bus id: 0000:83:00.0)
2017-08-05 19:10:39.591026: I tensorflow/contrib/gdr/gdr_memory_manager.cc:235] RDMA server is listening on 10.40.2.200:5001
2017-08-05 19:10:39.591071: I tensorflow/contrib/gdr/gdr_memory_manager.cc:285] Instrumenting CPU allocator cuda_host_bfc
2017-08-05 19:10:39.591083: I tensorflow/contrib/gdr/gdr_memory_manager.cc:285] Instrumenting CPU allocator cpu_pool
2017-08-05 19:10:39.591095: I tensorflow/contrib/gdr/gdr_memory_manager.cc:285] Instrumenting CPU allocator cpu_rdma_bfc
2017-08-05 19:10:39.591278: I tensorflow/contrib/gdr/gdr_memory_manager.cc:78] NUMA node for device: mlx4_0 is 1
2017-08-05 19:10:39.740253: I tensorflow/contrib/gdr/gdr_memory_manager.cc:296] Instrumenting GPU allocator with bus_id 2
```

The last line suggests that the GPUs with bus id 2 (mapped to pci bus id prefixed 0000:8) will benefit from GDR and host memory bypass, which is `/gpu:2` and `/gpu:3` in this case.

Caveats
===

In current implementation, only tensors that reside in host memory or in GPU memory such that the GPU is adjacent to an RDMA capable NIC will use direct RDMA as its transport. When RDMA is available but not GDR, a temporary tensor copy on host memory will be used as RDMA source/destination (and copied from/to the target device). When there is no RDMA device present, it can even fallback to the original gRPC runtime. While it is theoretically possible to mix GDR enabled TF with non-GDR deployments in the same job, make sure the environment is properly setup so the GDR mode is enabled whenever possible (i.e. do not fall back to gRPC when it is not absolutely necessary).


drpngx:
Jenkins, test this please.

drpngx:
Are you depending on a host RDMA library to be installed?

```
```
tensorflow/core/distributed_runtime/rpc/rdma.cc:13:27: fatal error: rdma/rdma_cma.h: No such file or directory
 #include <rdma/rdma_cma.h>
```

**Comment by byronyi** [comment]
@drpngx Yes, as in case of CUDA, we assume an OFED package is installed with all the drivers and user space libraries for to RDMA to properly function. `libibverbs` and `librdmacm` are included in an OFED package.

I agree that we need some sort of compile switch, but I haven't figure out the best way to do it.

**Comment by byronyi** [comment]
I have done some investigations and would like to purpose adding [`librdmacm`](https://github.com/linux-rdma/rdma-core/tree/master/librdmacm), [`libibverbs`](https://github.com/linux-rdma/rdma-core/tree/master/libibverbs), and [`libnl`](https://github.com/thom311/libnl) as third party dependencies (`librdmacm` depends on `libibverbs` which in turn depends on `libnl`, and `libnl` only depends on `libc`). All of these compile on latest Linux without any hardware requirements. For the runtime, one could simply call `ibv_get_device_list(3)` and check if the target environment requirements are met.

Now of course all of these are Linux specific, so we need an extra compatibility layer against non-Linux platforms. I do know there are [RDMA APIs](https://docs.microsoft.com/en-us/windows-hardware/drivers/network/overview-of-network-direct-kernel-provider-interface--ndkpi-) available on Windows, but I don't have a Windows box, so this has to be left to future contributors.

And for the real testing, we could use [SoftRXE](https://github.com/torvalds/linux/blob/master/drivers/infiniband/sw/rxe/Kconfig) as a software emulated RDMA layer that runs on any platform. It does require upstream Linux kernel version [at least 4.8](https://github.com/torvalds/linux/commit/8700e3e7c4857d28ebaa824509934556da0b3e76) to enable such feature, so we do need to setup a separated CI environment for that purposes.

**Comment by byronyi** [comment]
@shamoya Any interests on this patch? I would certainly appreciate any help from Mellanox on testing and setting up CI environment with GDR (and possibly with RoCE).

**Comment by drpngx** [comment]
@gunan

**Comment by gunan** [comment]
buldifier checks are failing here. Thus all the other tests are blocked.

```
Running do_buildifier on 204 files

tensorflow/core/distributed_runtime/rpc/BUILD # reformat callsort listsort unsafesort sort:tf_cuda_library.deps

buildifier took 1 s

248a249,254
>     linkopts = select({
>         "//conditions:default": [
>             "-libverbs",
>             "-lrdmacm",
>         ],
>     }),
250d255
<         "//tensorflow/core/distributed_runtime:rdma",
254a260
>         "//tensorflow/core/distributed_runtime:rdma",
256,258d261
<     linkopts = select({
<         "//conditions:default": ["-libverbs", "-lrdmacm"],
<     }),
```

Could you look into this failure, so that the rest of our CI tests can run?

**Comment by byronyi** [comment]
@gunan Yes, I am looking into it. Will ping you after it's resolved.

**Comment by byronyi** [comment]
I am working on introducing the [RDMA user space libraries](https://github.com/linux-rdma/rdma-core) as third party dependencies, but I am not familiar with converting CMake rules into bazel BUILD files. A similar case #5349 of integrating cURL seems to be a much larger effort than I expected. Would appreciate if @jart or @gunan could give some suggestions on this.

**Comment by drpngx** [comment]
Jenkins, test this please.

**Comment by byronyi** [comment]
@drpngx Sorry I was pushing the wrong commit. Could you ask Jenkins to test it again?

**Comment by drpngx** [comment]
Jenkins, test this please.

**Comment by byronyi** [comment]
Seems a rebase is needed.

**Comment by shamoya** [comment]
Thanks @byronyi 
Let me take a look at this patch and get back to you soon.
We thought of GDR in the context of the verbs code till now, so this is interesting.
Adding the RDMA libraries as third party is a good idea, good also for the grpc+verbs protocol flow.

**Comment by byronyi** [comment]
@shamoya Glad to have your prompt reply. The last commit turns out to be a unsuccessful attempt to introduce RDMA dependency. If you would like to try out the patch on a machine with OFED installed, you could revert the last commit and build it from source.

**Comment by shamoya** [comment]
The libraries in third_party are linked statically to the code right ?
In general, how can one provide a different library version then the one in third_party (grpc for example) ?

**Comment by byronyi** [comment]
My initial attempt was to include only RDMA headers, which apparently is not runnable as it isn't linked against the actual library. I haven't figured it out how to use bazel to build rdma-core; they are using CMake and it's hard to port it to bazel in general. I guess we need to consult someone in the TF team on platform and portability issues.

Regarding to the version of specific third party library, I think it is hardcoded into the compilation directives. Google seems to compile everything from source for their projects, and I guess this isn't a issue for them.

**Comment by drpngx** [comment]
Jenkins, test this please.

**Comment by byronyi** [comment]
@drpngx I am still working on resolving the build issues...maybe we should wait for review from @poxvoculi first?

**Comment by drpngx** [comment]
OK. I usually prefer to review when everything builds and tests properly.
Can you summarize what your questions are?

**Comment by byronyi** [comment]
Current CI doesn't have GDR environment so we are specifically testing the "fallback to gRPC" behaviour.

I've already implemented such logic in my code ([here](https://github.com/tensorflow/tensorflow/pull/11392/files#diff-3dd79508fcea22184e8086c24dae0deaR349) and [there](https://github.com/tensorflow/tensorflow/pull/11392/files#diff-c404e67a9906d0e4749fd1f718ea3c73R137)); but GDR environment requires RDMA user space libraries, which are currently not linked into the TF core library. Currently I am stuck by adding them as a third-party dependency (porting CMake based project to bazel).

Alternatively I could make it a separated runtime (like grpc+verbs and grpc+mpi), but that will defeat the whole point of being able to fallback to gRPC.

**Comment by byronyi** [comment]
Or I could just add a compiler directive in `configure` and disable everything that's useful.

**Comment by byronyi** [comment]
@drpngx I've tested my non-RDMA box and everything works just fine. It does require the box to have `librdmacm` and `libibverbs` installed, though. It falls back to gRPC correctly as suggested.

Is there anyway to setup a CI environment with those two libraries? They are available on all Linux distros I know of. After we confirm everything builds and tests properly, we can arrange a review on my patch.

**Comment by byronyi** [comment]
@gunan The buildifier complaint is fixed.

**Comment by byronyi** [comment]
As discussed in linux-rdma/rdma-core#163, I have to give up adding the RDMA libraries as third party dependencies, and rely on target environment for building GDR. This means an extra compile switch in the `configure` file, and also a macro that shadow all the RDMA api calls.

I will start to migrate my patch in the next few days.

**Comment by byronyi** [comment]
@drpngx I tested locally with my non-RDMA box and it works. Would appreciate if you could let Jenkins test it again.

**Comment by drpngx** [comment]
Jenkins, test this please.

On Jul 12, 2017 9:22 PM, "Bairen Yi" <notifications@github.com> wrote:

@drpngx <https://github.com/drpngx> I tested locally with my non-RDMA box
and it works. Would appreciate if you could let Jenkins test it again.

—
<https://github.com/tensorflow/tensorflow/pull/11392#issuecomment-314966557>,
or mute the thread

**Comment by byronyi** [comment]
@drpngx Tests are blocked by sanity checks. I've auto-fixed all of them using clang-format and buildifier. Mind to let Jenkins test again?

**Comment by gunan** [comment]
Jenkins, test this please.

**Comment by jart** [comment]
@byronyi We haven't written a Bazel config for rdma-core internally. What I would probably do if I were you is see you can run its standard cmake/ninja build and get it to output the actual gcc/etc. commands it's running (in Bazel that's `bazel build -s //foo`) Then follow loosely the same pattern we did with our third party build files. You can add me as a reviewer once you come up with a PR that does this. Thank you, by the way.

**Comment by byronyi** [comment]
@drpngx While I will be working to resolve the failing Windows/MacOS/Py3 tests, could we arrange a review as Linux CPU/GPU has already builds and tests well?

**Comment by drpngx** [comment]
Yes, I think @poxvoculi is OK with it in principle, I will let him approve.

**Comment by byronyi** [comment]
@drpngx That would be great. I would definitely appreciate if there are any comments on the overall design or styling issue from @poxvoculi.

**Comment by byronyi** [comment]
Seems the patch could be further refactored using `DeviceAttributes` that present in `RecvTensorRequest`. See the commit 79228c74e64a639aeb5692b442522d4aa279f885.

**Comment by drpngx** [comment]
Jenkins, test this please.

**Comment by poxvoculi** [comment]
My main initial review question is why this PR modifies so many files in core.   We're trying to keep alternate RPC implementations in contrib for the time being.  Are there any special difficulties with putting this stuff into contrib?

**Comment by byronyi** [comment]
No, at least I couldn't think of any specific difficulties preventing us from moving it to contrib. Let's just say we register a separated server protocol `grpc+gdr`, and override specific behaviors like what's currently done for verbs and MPI.

Do you have any other comments before I start migrating this patch?

**Comment by poxvoculi** [comment]
No, other than the disruption to core it looks pretty good.  Our long term design goal is to enable a diversity of distributed I/O solutions to be used in a modular fashion with minimal impact on binaries that don't need them.  While we're still gaining experience with alternate solutions, we want to keep them in contrib.

**Comment by vrv** [comment]
@byronyi I assume you're still working on this?  Just want to keep tabs :)

**Comment by byronyi** [comment]
@vrv 

Yes, I was busy on something else for the past week. I will try to finish the migration and hopefully get this PR into rc1 or rc2. We don't want to miss the 1.3 release :)

**Comment by suiyuan2009** [comment]
I got this error more than once when using this patch to run benchmarks code.
```
2017-07-28 11:22:43.361608: I tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc:330] Started server with target: grpc://localhost:39799
2017-07-28 11:23:28.395109: I tensorflow/core/distributed_runtime/master_session.cc:998] Start master session ec37ba189b6f7f05 with config: intra_op_para
llelism_threads: 1 gpu_options { force_gpu_compatible: true } allow_soft_placement: true
2017-07-28 11:23:34.103612: I tensorflow/core/distributed_runtime/rpc/rdma.cc:435] Accepted new RDMA connection
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 1285, in _do_call
    return fn(*args)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/client/session.py", line 1264, in _run_fn
    status, run_metadata)
  File "/usr/lib/python3.5/contextlib.py", line 66, in __exit__
    next(self.gen)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/framework/errors_impl.py", line 466, in raise_exception_on_not_ok_status
    pywrap_tensorflow.TF_GetCode(status))
tensorflow.python.framework.errors_impl.InternalError: WhereOp: Could not launch cub::DeviceReduce::Sum to count number of true indices, status: invalid 
resource handle
         [[Node: report_uninitialized_variables/boolean_mask/Where = Where[_device="/job:worker/replica:0/task:0/gpu:0"](report_uninitialized_variables/b
oolean_mask/Reshape_1_S4219)]]

```

**Comment by suiyuan2009** [comment]
The above error is fixed by recent commits on master branch.

**Comment by suiyuan2009** [comment]
@byronyi , I did some test on benchmarks code using `distributed_replicated` mode and `gpu` as `local_param_device`.
I find sometimes I got ``` 2017-07-31 17:09:05.880231: W tensorflow/core/distributed_runtime/rpc/rdma.cc:129] No RDMA device found ```, maybe my ib driver has some problem.
I also got ```*** Error in `python3': free(): invalid pointer: 0x00007f47740066d0 ***``` sometimes, detail is on [gist](https://gist.github.com/suiyuan2009/434b8abaf98ad9f81e39905df0a98db2).
And my 1 worker to 4 workers speedup ratio is a little strange(small).

**Comment by byronyi** [comment]
@poxvoculi While the migration to contrib is still ongoing, I have removed most modifications to core. Those left are the ones may lead to substantial duplicate code if removed. I'd like to seek for your opinion on whether these changes are acceptable for core, so I will have a solid ground for the contrib part.

**Comment by byronyi** [comment]
@suiyuan2009 Thanks for raising those issues. 

When `No RDMA device found` prints [here](https://github.com/red-bird/tensorflow/blob/0cb5019de7329b10d60a8d74f6c4665169578400/tensorflow/core/distributed_runtime/rpc/rdma.cc#L129) it probably means there is no available IB device present. It is a fairly simple API call (`rdma_get_devices`) without any dependency, so it itself is unlikely to go wrong. Let me know if this issue persists; I could take a closer look to your environment.

We have seen the invalid free during our development process, and while we haven't figured out the reasons behind, some suggested it might because of incompatible build/runtime environment, such as python or malloc version (#9328). I suspect it might because the virtualenv that you used to build TF is different from the runtime (as once in our case), but I am not sure. I will come back to this one after tidying up the current mess.

About the slow speedup, could you elaborate?

**Comment by byronyi** [comment]
@drpngx I've finished migration to contrib. Mind to kick off Jenkins for test?

**Comment by byronyi** [comment]
@suiyuan2009 I've finished the migration, and several potential inconsistencies in `rdmacm` API calls have been fixed. Would you mind to test your code again and see if your issue persists?

**Comment by gunan** [comment]
Jenkins, test this please.

**Comment by byronyi** [comment]
Seems to be the same issue mentioned [here](https://github.com/tensorflow/tensorflow/pull/8943#issuecomment-292676004). Fixed by a macro wrapping up the external rdma headers.

**Comment by suiyuan2009** [comment]
@byronyi , I meet `tensorflow.python.framework.errors_impl.UnavailableError: Cannot bind to rdma://10.9.38.144:50543` error, [gist log](https://gist.github.com/suiyuan2009/1c7f54908b121227d01914548ffe3bdb#file-gdr-patch-test-L16). I also meet `tensorflow.python.framework.errors_impl.UnavailableError: Cannot find pinned memory region` error, [gist log](https://gist.github.com/suiyuan2009/1c7f54908b121227d01914548ffe3bdb#file-gdr-patch-test-L249).

**Comment by byronyi** [comment]
@suiyuan2009 I am trying to fix this problem, thank you for your log! It helps a lot to locate the bug.

I've refactored the code a lot and going through all these manual test cycles every single time. Maybe I should write some unit tests for the patch to save all these efforts, but I have not figured out the best way to organize it. Any ideas are more than welcomed :)

**Comment by byronyi** [comment]
@poxvoculi After several rounds of refactoring, this patch is finally ready for review. Please let me know if there is any issue.

**Comment by suiyuan2009** [comment]
I still got `tensorflow.python.framework.errors_impl.UnavailableError: Cannot bind to rdma://10.9.0.196:41368` sometimes.

**Comment by byronyi** [comment]
@suiyuan2009 I've pushed a new commit that improves error handling by including `errno` in the reporting error status. Mind to check again?

**Comment by byronyi** [comment]
I've been communicating with @suiyuan2009 via another channel, and he confirms the `cannot bind` is an environment issue of their IB devices.

**Comment by byronyi** [comment]
@vrv Kindly ping you for kicking off the testing again.

**Comment by vrv** [comment]

File: tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+  void Start(std::function<void()> recv_done) override {
+    req_.set_dma_ok(true);
+    resp_.InitAlloc(dst_device_, recv_args_.alloc_attrs);
+    StatusCallback cb = [this, recv_done](const Status& s) {
+      bool dma_ok = resp_.metadata().has_transport_options();
+      if (s.ok() && tensor().TotalBytes() > 0 && (!is_dead()) && dma_ok) {
+        auto transport_options = resp_.metadata().transport_options();
+        Status s = remote_memory_manager_->TensorFromTransportOptions(
+            const_cast<Tensor*>(&tensor()), transport_options, dst_device_,
+            recv_args_.device_context, recv_args_.alloc_attrs.on_host());
```
poxvoculi: I believe that alloc_attrs.on_host is not guaranteed to be true for all CPU RAM resident tensors, just for those that are parameters of a non-CPU device Op that are also CPU RAM resident.   How does this code handle ordinary CPU Tensors?

tensorflow/contrib/gdr/gdr_memory_manager.cc
```diff
@@ -0,0 +1,608 @@
...
+    LOG(WARNING) << "Cannot register memory region";
+  }
+}
+
+void GdrMemoryManager::EvictMemoryRegion(void* addr, size_t length) {
+  if (length == 0) return;
+  mutex_lock l(alloc_mu_);
+  auto iter = std::upper_bound(mrs_.begin(), mrs_.end(), addr, &Comparator);
+  if (iter != std::end(mrs_) && iter->get()->addr == addr) {
+    mrs_.erase(iter);
```
poxvoculi: There's no deregistration, right?  In practice we never shrink our memory pools until process termination, so this is probably ok, but might be better to clean up properly.

tensorflow/contrib/gdr/gdr_memory_manager.h
```diff
@@ -0,0 +1,38 @@
...
+  virtual Status TransportOptionsFromTensor(
+      ::google::protobuf::Any* mutable_transport_options, const Tensor& tensor,
+      Device* device, DeviceContext* device_context, bool on_host) = 0;
+
+  virtual Status TensorFromTransportOptions(
+      Tensor* tensor, const ::google::protobuf::Any& transport_options,
+      Device* device, DeviceContext* device_context, bool on_host) = 0;
+};
+
+RemoteMemoryManager* CreateRemoteMemoryManager(const string& host,
```
poxvoculi: What happens if a second GdrMemoryManager is created on the same host:port without deleting the first?

tensorflow/contrib/gdr/gdr_worker.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "tensorflow/core/distributed_runtime/rpc/grpc_worker_service.h"
+
+namespace tensorflow {
+
+class GdrWorker : public GrpcWorker {
+ public:
+  GdrWorker(WorkerEnv* env, RemoteMemoryManager* remote_memory_manager);
+
+  virtual void RecvTensorAsync(CallOptions* opts,
```
poxvoculi: Document semantics with comment.

poxvoculi (commented):
Thanks for moving this out into contrib.  The changes to core are minimal and look fine.

tensorflow/contrib/gdr/gdr_memory_manager.h
byronyi: It will fail in the `GdrMemoryManager::Init()`, which will break in [`rdma_create_ep`](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L204) with `EADDRNOTAVAIL` (Cannot assign requested address). 

As a side note, the rdmacm library supports set `RDMA_OPTION_ID_REUSEADDR` using `rdma_set_option`, similar to `SO_REUSEADDR`. I could not think of an use case setting that option though, nor have I tried it myself.

tensorflow/contrib/gdr/gdr_memory_manager.cc
```diff
@@ -0,0 +1,608 @@
...
+#include "tensorflow/core/lib/core/status.h"
+#include "tensorflow/core/platform/macros.h"
+#include "tensorflow/core/platform/mutex.h"
+
+namespace tensorflow {
+
+namespace {
+
+bool IsGDRAvailable() {
+  std::ifstream ifs("/proc/modules");
```
byronyi: There are 3 Linux specific dependencies:

1. the [RDMA user space libraries](https://github.com/linux-rdma/rdma-core), co-evolving with all [kernel drivers](https://github.com/torvalds/linux/tree/master/drivers/infiniband) in Linux source tree
2. the [nv_peer_mem](https://github.com/Mellanox/nv_peer_memory) Linux kernel module, developed by Menallox and NVIDIA using their [kernel APIs](http://docs.nvidia.com/cuda/gpudirect-rdma/index.html#kernel-api)
3. detection of numa node of IB device using `/proc`, as similar to [what is being done for CUDA device](https://github.com/tensorflow/tensorflow/blob/45254d9d7f790f97746775967637773d137682a9/tensorflow/stream_executor/cuda/cuda_gpu_executor.cc#L899)

Being said that, I agree with you that a partial fix could be introduced in the same way as it is handled in #5991.

tensorflow/contrib/gdr/gdr_memory_manager.cc
```diff
@@ -0,0 +1,608 @@
...
+class BFCRdmaAllocator : public BFCAllocator {
+ public:
+  BFCRdmaAllocator()
+      : BFCAllocator(new BasicCPUAllocator(), 1LL << 36, true, "cpu_rdma_bfc") {
+  }
+};
+
+REGISTER_MEM_ALLOCATOR("BFCRdmaAllocator", 300, BFCRdmaAllocator);
+
+GdrMemoryManager::GdrMemoryManager(const string& host, const string& port)
```
byronyi: Yes, this is designed to be consistent with the `cluster_def` [passed into gRPC server construction](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/core/distributed_runtime/rpc/grpc_server_lib.cc#L137) (each process listen to a particular TCP/IP address, which [propagates to here](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_server_lib.cc#L14).

As a side note, I have not figured out a viable design to accommodate multiple IB interfaces yet. For the NICs I know of, there could actually be multiple ports on the same NIC (connected to different switches, as in our case),  so there are some potentials doing load-balancing or fault-tolerance. And then of course there could be multiple NICs (as in DGX-1) to serve on different buses, which leaves interesting potentials doing request routing using device localities in the future.

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: The deregistration logic is handled by [`EvictMemoryRegion`](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L593), which calls `rdma_dereg_mr` in [`MRDeleter`](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L83). 

`EvictMemoryRegion` is registered as a [free visitor](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L250) for CPU allocators, and I explained why it is not handled for GPU device allocator [in the comment](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L270).

tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc
byronyi: The `on_host` parameter is only used([here](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L390) and [there](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L448)) when the passed tensor cannot be found in the NIC-registered regions. And if we register ordinary CPU tensors to NIC, they could be transferred by RDMA directly.

Being said that, I agree that we could use a more accurate flag than `alloc_attrs.on_host`. Any suggestions?

byronyi:
Apart from docs, the only remaining issue I am not sure about are the CPU-only allocators. The way it is handled now is a bunch of hacks:

1. it overrides the default CPU allocator, which allocates 100% of all CPU RAM tensors 
2. it re-introduces a `BasicCPUAllocator` which is exactly the same with the one in `gpu/pool_allocator.h` (I actually copied directly from there)
3. it could not co-exists with `MKLCPUAllocator` as it precedes the latter one with a higher priority

The main rationale of these hacks is that the current [default cpu allocator](https://github.com/tensorflow/tensorflow/blob/e073322452e41e76754314aa75d543d071003fc5/tensorflow/core/framework/allocator.cc#L71) is not visitable, so it will fail at [this line](https://github.com/red-bird/tensorflow/blob/955764316dd9c796db874ae2034e336a48b1fb46/tensorflow/contrib/gdr/gdr_memory_manager.cc#L257). The entire `BFCRdmaAllocator` is introduced to avoid this failure.

Note that MKL suffers the exact same problem, and they uses the exactly same hacks: 

1. it overrides the default CPU allocator
2. it re-introduces a `MklSubAllocator` that is equivalent to the `BasicCPUAllocator`
3. it could not co-exists with other CPU allocators

The tentative fix I will propose is just to make the current default cpu allocator subclasses from `VisitableAllocator`, which save all my hacks to re-introduce a visitable default cpu allocator.

tensorflow/contrib/gdr/gdr_memory_manager.cc
```diff
@@ -0,0 +1,608 @@
...
+      std::bind(&GdrMemoryManager::InsertMemoryRegion, this, _1, _2);
+  VisitableAllocator::Visitor free_visitor =
+      std::bind(&GdrMemoryManager::EvictMemoryRegion, this, _1, _2);
+
+  // Host memory allocators
+  for (Allocator* allocator : allocators) {
+    CHECK(allocator);
+    auto* visitable_allocator = dynamic_cast<VisitableAllocator*>(allocator);
+    if (!visitable_allocator) {
+      LOG(WARNING) << "Cannot instrument non-visitable CPU allocator "
```
byronyi: See my comments in the main thread; these hacks are related with each other.

tensorflow/contrib/gdr/gdr_memory_manager.cc
```diff
@@ -0,0 +1,608 @@
...
+};
+
+class BFCRdmaAllocator : public BFCAllocator {
+ public:
+  BFCRdmaAllocator()
+      : BFCAllocator(new BasicCPUAllocator(), 1LL << 36, true, "cpu_rdma_bfc") {
+  }
+};
+
+REGISTER_MEM_ALLOCATOR("BFCRdmaAllocator", 300, BFCRdmaAllocator);
```
byronyi: See my comments in the main thread; these hacks are related with each other.

tensorflow/contrib/gdr/gdr_memory_manager.cc
```diff
@@ -0,0 +1,608 @@
...
+      GUARDED_BY(cient_mu_);
+
+  // Managed memory regions
+  mutex alloc_mu_;
+  std::vector<MemoryRegionPtr> mrs_ GUARDED_BY(alloc_mu_);
+
+  TF_DISALLOW_COPY_AND_ASSIGN(GdrMemoryManager);
+};
+
+class BasicCPUAllocator : public SubAllocator {
```
byronyi: See my comments in the main thread; these hacks are related with each other.

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: I've pushed a commit af28527; let me know if you have any comments on that.

tensorflow/contrib/gdr/gdr_worker.h
byronyi: See commit `48824184e`. Feel free do modify it if you see as fit.

tensorflow/contrib/gdr/gdr_memory_manager.h
```diff
@@ -0,0 +1,38 @@
...
+class DeviceContext;
+class Tensor;
+
+class RemoteMemoryManager {
+ public:
+  virtual Status Init() = 0;
+  virtual void Run() = 0;
+  virtual void Stop() = 0;
+
+  virtual Status TransportOptionsFromTensor(
```
byronyi: See commit `48824184e`. Feel free do modify it if you see as fit.

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: BTW, is `common/runtime/gpu/pool_allocator.h` available when building without GPU runtime?

byronyi:
It just came to my mind that I don't need to precede the MKL allocator; it's visitable (BFCAllocator) by default. I just need to precede the non-MKL default allocator, which is not visitable as in current stage.

I haven't tested it yet, but presumably the interference issue should be resolved with a lower allocator registration priority than that of MKL (higher than default).

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: @shamoya Is GDR available on Windows? I am actually pretty interested in porting my patch to Win32, as there are already some APIs like [NetDMA](https://docs.microsoft.com/en-us/windows-hardware/drivers/network/netdma-drivers) or [Network Direct Kernel Provider Interface](https://docs.microsoft.com/en-us/windows-hardware/drivers/network/network-direct-kernel-programming-interface--ndkpi-). In a recent talk from MSR people, I heard that they could achieve pretty good improvement using RDMA to train RNN on CNTK, and I think it would be worth doing that on TF as well.

tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc
poxvoculi: It's a bit of a hack, but the current standard test for CPU RAM is here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc#L347
Of course, it requires the device ptr.

poxvoculi:
I think it makes sense to try to make the default CPUAllocator visitable, but maybe not in this PR.  I may not understand the difficulties of doing so.  I'll open up a discussion with @vrv and @zheng-xq about it.

poxvoculi:

tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc
byronyi: It's fixed in c20e8f4c3.

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: I've added a `CHECK(visitable_allocator)` in e841bc80f.

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: I've added a TODO to remind me remove this when the default allocator is visitable; see e841bc8.

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: I've added a TODO to remind me remove this when it is accessible in common_runtime; see e841bc8.

byronyi:
Seems the Windows build is broken perhaps because of a crashed build earlier:
```
11:51:15 Another git process seems to be running in this repository, e.g.
11:51:15 an editor opened by 'git commit'. Please make sure all processes
11:51:15 are terminated then try again. If it still fails, a git process
11:51:15 may have crashed in this repository earlier:
11:51:15 remove the file manually to continue.
11:51:15 From https://github.com/tensorflow/tensorflow
11:51:15  ! [new ref]               refs/pull/5992/head -> origin/pr/5992/head  (unable to update local ref)
11:51:15 error: cannot lock ref 'refs/remotes/origin/pr/915/head': Unable to create 'c:/tf_jenkins/home/workspace/tensorflow-pr-win-cmake-py/.git/refs/remotes/origin/pr/915/head.lock': File exists.
```

poxvoculi (approved):
This change looks fine to me.  I'm not an expert on the OS config & BUILD changes but they look consistent with existing practice.

drpngx:
Jenkins, test this please.

byronyi:
Thanks @poxvoculi. I have learned a lot as it is my first patch to OSS implementing an end to end feature. I really appreciate your taking time reviewing my code. 

I am glad we could finally catch up with the 1.3 release :)

By the way, this patch is an effort I have worked on for a long time. As a grad student, it would not be possible without the funding from my lab (System Networking Lab in HKUST). I personally spent quite some time convincing my supervisor to open source it, given the condition that our credit may be acknowledged in the release notes. That would be immensely helpful to us.

poxvoculi:
@byronyi This is a nice contribution and I'm glad you were able to take the time to do it.  I look forward to hearing from appreciative users.  I don't know anything about how OS contributors may be acknowledged in the release notes.  Do you wish to inquire about that?

Mazecreator:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  mazecreator@gmail.com
    Domain mazecreator.com has exceeded the max emails per hour (26/25 (104%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from o10.sgmail.github.com ([167.89.101.201]:21026)
	by server2.lowesthostingrates.com with esmtps (TLSv1.2:ECDHE-RSA-AES128-GCM-SHA256:128)
	(Exim 4.89)
	(envelope-from <bounces+848413-e6d9-mazecreator=mazecreator.com@sgmail.github.com>)
	id 1ddkcm-0000Km-IE
	for mazecreator@mazecreator.com; Fri, 04 Aug 2017 16:59:46 -0500
DKIM-Signature: v=1; a=rsa-sha1; c=relaxed/relaxed; d=github.com; 
	h=from:reply-to:to:cc:in-reply-to:references:subject:mime-version:content-type:content-transfer-encoding:list-id:list-archive:list-post:list-unsubscribe; 
	s=s20150108; bh=I+XY5aaYnuMYWTkL8rLESL+Ai/o=; b=QWInvodHLYIV/qbF
	h0C5AqJM6BqWPNHh0SI8YuK47PGgmls6iQqgTbtnkrUwe83GllX/QhCZ44CQpVo7
	qTiWgT4UZl3756f65pVDtEKrQsleWcMSpjrODHwrN3N8/UH++xSBzVd4MBQT57kE
	og+6KQnjpyNfqPFjQGbCPUSNWo8=
Received: by filter0454p1mdw1.sendgrid.net with SMTP id filter0454p1mdw1-1327-5984F0B4-49
        2017-08-04 22:09:56.68207495 +0000 UTC
Received: from github-smtp2b-ext-cp1-prd.iad.github.net (github-smtp2b-ext-cp1-prd.iad.github.net [192.30.253.17])
	by ismtpd0024p1mdw1.sendgrid.net (SG) with ESMTP id okXsfIqdRkmaL3TXzA__Zg
	for <mazecreator@mazecreator.com>; Fri, 04 Aug 2017 22:09:56.621 +0000 (UTC)
Date: Fri, 04 Aug 2017 22:09:56 +0000 (UTC)

suiyuan2009:
I meet stuck problem on 1080ti today twice, first time I got warning says `tensorflow/core/common_runtime/bfc_allocator.cc:217] Allocator (GPU_2_bfc) ran out of memory trying to allocate 2.29GiB. The call`, this time I got nothing, I see cpu and memory are still being used but gpu utility is 0%, this tf benchmarks program has run 3 hours...

byronyi:
@poxvoculi Would you mind to review the last commit again? I have fixed several issues under VLOG, some of the rationale could be found [here](https://github.com/tensorflow/tensorflow/commit/ec1403e7dc2b919531e527d36d28659f60621c9e#commitcomment-23432313).

tensorflow/contrib/gdr/README.md
```diff
@@ -0,0 +1,122 @@
+Introduction
+===
+
+This is an implementation of GDR out-of-band transport for TensorFlow distributed runtime, complementary to current gRPC transport. It uses gRPC as control plane to setup rendezvous for each tensor transmission, and utilizes [GPU Direct RDMA](https://developer.nvidia.com/gpudirect) whenever possible to transmit tensors in remote GPU memory through network interface card (NIC), bypassing host memory and CPU entirely. It gracefully falls back to ordinary RDMA or even gRPC when GDR is not available.
+
+Design
+===
+
+The GDR out-of-band transport is designed to avoid any unnecessary memory copies, especially for large tensors (>100MB). That typically requires registration of tensor buffers to NIC in an ad-hoc manner, which is rather slow as described in the design trade-off of the verbs runtime. The verbs runtime thus chooses to manager its own NIC-registered buffers and copy the tensors from/to those buffers for every single tensor transfer.
```
poxvoculi: s/to manager/to manage/

tensorflow/contrib/gdr/README.md
```diff
@@ -0,0 +1,122 @@
...
+			state:			PORT_ACTIVE (4)
+			max_mtu:		4096 (5)
+			active_mtu:		1024 (3)
+			sm_lid:			0
+			port_lid:		0
+			port_lmc:		0x00
+			link_layer:		Ethernet
+```
+
+2. There is a GDR capable GPU, i.e. of Fermi, Kepler or later architecture with [corresponding driver](http://docs.nvidia.com/cuda/gpudirect-rdma/index.html) installed. The PCI-e topology could be confirmed by `nvidia-smi topo -m`. Fore example, in the following topology, `GPU2` and `GPU3` are adjacent to `mlx4_0`, and tensors on these devices could benefit from GDR in current implementation.
```
poxvoculi: s/Fore/For/

byronyi:
After adding checksum support in VLOG, I found GDR will result in increasing loss for GPU when checksum is disabled. It seems that some synchronisation is required for truly direct DMA between NIC and GPU, and the stream executor seems to be place for me to look into.

@poxvoculi Any comments?

byronyi:
Seems a merge is needed. Mind to kickoff testing again?

byronyi:
@poxvoculi We discussed a little bit internally and agree that getting GDR shipped on 1.3 is our priority. Would you mind putting GDR on the release notes and get this PR merged? That would be great for potential users out there.

martinwicke:
1.3 has been cut a while ago, and we won't put additional features in this
late (we're already at RC2). This will have to wait until 1.4.

byronyi:
So this PR will not be merged until 1.3 is released? 

I'm fine with it per se, but would it be possible for you guys to let us know when will be the merge window cutdown in the future releases? 

It would be great for us (and all other contributors) to make better planning ahead, instead of struggling for missed deadlines.

martinwicke:
No, we can merge this PR any time, but it won't be in the 1.3 branch, and
if anyone installs the 1.3 binary they won't see it's effects.

The branches are cut every 6 weeks or so. I don't want to encourage aiming
features at specific releases. The deadline is usually not good for
quality. But if you say loud and early that you think a feature should be
in the next release (and if it is ready), then we can plan for that. This
feature wasn't ready in time for 1.3 (~2 weeks ago) anyway, so not much
point in hurrying too much.

byronyi:
You mentioned about the binary shipped, but this PR is in contrib and not built by default. The users have to turn on the related compile switch and build it from source themselves. Does the release policy apply to source release as well?

martinwicke:
There is no source release other than getting the source code from GitHub.
So once we merge this it's available for people to use.

However if someone specifically asks to build the 1.3 branch, they will get
the source code which builds the 1.3 binary, which would not include this
change.

byronyi:
Thanks for clarification! I actually just realized that :)

rmlarsen:

tensorflow/contrib/gdr/gdr_memory_manager.cc
```diff
@@ -285,12 +290,12 @@ Status GdrMemoryManager::Init() {
   }
 
 #if GOOGLE_CUDA
+  VisitableAllocator::Visitor cuda_alloc_visitor =
+      std::bind(&GdrMemoryManager::InsertCUDAMemoryRegion, this, _1, _2);
```
poxvoculi: If I understand what you're doing correctly, you're registering the entire GPU memory to have the CU_POINTER_ATTRIBUTE_SYNC_MEMOPS property.  This seems to be a technique that NVIDIA provides because it makes concurrent programming easier, but I would avoid it because it likely will also damage performance considerably.

It sounds like you suspect a race condition on either the RDMA read, or some use of the final value.   For background, here's how we generally handle this issue.

TensorFlow tries hard to use as few synchronous CUDA calls as possible, and to introduce as few sync points as possible.   The techniques used rely on careful use of streams. The general discipline is that 4 streams are created for each GPU, by the [StreamGroupFactory](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L182).  All compute Ops (i.e. CUDA kernels) are launched on the single compute stream, and the other three are used for memcpys: one for H2D, one for D2H, and one for D2D copies.  In the normal case where an op relies only on the completion of prior ops in the same stream, it can be launched async without danger or any other temporal dependency. In the case where it relies on prior completion of an op on another stream, we need to introduce a sync dependency, which can be done in more than one way.  If we need to wait for an ordinary compute op prior to e.g. a D2H copy, we can just introduce a wait on the current compute stream prior to the copy, as [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_util.cc#L214).   This will cause the i/o stream to wait for completion of all ops _pending at the time of call_ to complete before launching the next op added to itself.  

In the case where we need to wait for a memcpy to terminate on one stream before launching a compute on another, the technique used is [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_util.cc#L168).  We don't want the compute stream to stall until the i/o is done, so we wait for the i/o to complete before queuing the compute op.  The wait is accomplished by a call the the EventMgr.

In our internal networking which uses GPUDirect, we use the EventMgr to ensure that we don't start a send before the op writing the buffer to be read has completed.  In the other direction, our RPC system provides a callback that executes only after the target buffer area has been written, and we use that callback to trigger subsequent ops that want to read that area.

rmlarsen:

tensorflow/contrib/gdr/gdr_memory_manager.cc
byronyi: I noticed the stream executor and I think that it is a good design for both pipelining and parallelism. It is just that I have not figured out a viable path to accommodate my current design with it, and I need to make a temporary fix to the race condition issue.

Let me take a closer look and see if I can do better. Thanks for the comment! It is very helpful.

byronyi:
I thought I will make some further improvements, but @rmlarsen seems to be eager to merge this. Alright then, I think I will leave them to the future 😆 

Good job guys, thank you all for all the feedbacks in this thread. Hope we could have some users soon so we could hear back from them.

shamoya:
Thank you @byronyi for the contribution!
I hope we'll start testing it soon.
J

suiyuan2009:
I find first gpu's forward propagation is much slower than other gpus on same machine when training resnet101 using `grpc+gdr` on a 4 gpus worker machine (`grpc` may have same problem, but `grpc` is much slower so it may hide the problem.), I have run many times, trace file is in [here](https://drive.google.com/file/d/0B6phHPK0iUvdSk9hV2NIY1hkZ3M/view?usp=sharing).

avolkov1:
@byronyi I keep getting an error: `UnavailableError: No such device: cannot bind to rdma://hsw215:2302`

No matter what port I've tried I get that error. The hsw215 is a compute node. This is my cluster spec:
```
CLUSTER_SPEC_DICT: {'ps': ['hsw215:2300'], 'worker': ['hsw215:2301', 'hsw215:2302', 'hsw215:2303', 'hsw215:2304']}
```

I compiled TF 1.4 with GDR, and I have nv_peer_mem running:
```
$ ibv_devinfo
hca_id: mlx5_0
        transport:                      InfiniBand (0)
        fw_ver:                         10.16.1006
        node_guid:                      e41d:2d03:0006:ae40
        sys_image_guid:                 e41d:2d03:0006:ae40
        vendor_id:                      0x02c9
        vendor_part_id:                 4113
        hw_ver:                         0x0
        board_id:                       MT_1210110019
        phys_port_cnt:                  2
        Device ports:
                port:   1
                        state:                  PORT_ACTIVE (4)
                        max_mtu:                4096 (5)
                        active_mtu:             4096 (5)
                        sm_lid:                 1
                        port_lid:               16
                        port_lmc:               0x00
                        link_layer:             InfiniBand

                port:   2
                        state:                  PORT_DOWN (1)
                        max_mtu:                4096 (5)
                        active_mtu:             4096 (5)
                        sm_lid:                 0
                        port_lid:               65535
                        port_lmc:               0x00
                        link_layer:             InfiniBand


$ service nv_peer_mem status
nv_peer_mem module is loaded.
```

My code runs fine with `protocol='grpc+verbs'`. I get messages like:
```
I tensorflow/contrib/verbs/rdma_mgr.cc:56] connecting to remote node /job:worker/replica:0/task:0
```
Rdma runs well, but I can't get the gdr working. Maybe I'm doing something stupid, but it's not clear from the README what else needs to be done. Thanks.

byronyi:
@avolkov1 The GDR runtime binds to whatever IP your assigned, even if it is not an IB device. Please check with IP address first. Sometimes it resolves to different IP for the same host on different machines.

avolkov1:
@byronyi Thank you. Made some progress. I'm getting `Unavailable: Cannot find pinned memory region` errors. Have you come across that? I'm trying to run multigpu training. Again, this worked with `grpc+verbs`. Are there some gotchas with parameter servers on CPU.

I get the RDMA GDR endpoints working:
```
1: 2017-10-19 11:17:07.804151: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
2: 2017-10-19 11:17:07.803164: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw215.ib.cluster:2301
0: 2017-10-19 11:17:08.033571: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw215.ib.cluster:2301
1: 2017-10-19 11:17:08.033677: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
1: 2017-10-19 11:17:21.449746: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw216.ib.cluster:2300
2: 2017-10-19 11:17:21.448892: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
3: 2017-10-19 11:17:21.455124: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw216.ib.cluster:2300
2: 2017-10-19 11:17:21.455214: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
1: 2017-10-19 11:17:22.995750: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw215.ib.cluster:2300
0: 2017-10-19 11:17:22.995812: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
2: 2017-10-19 11:17:29.572063: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw216.ib.cluster:2301
3: 2017-10-19 11:17:29.572255: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
3: 2017-10-19 11:17:29.572637: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw215.ib.cluster:2300
0: 2017-10-19 11:17:29.573710: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
0: 2017-10-19 11:17:38.169948: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw216.ib.cluster:2301
3: 2017-10-19 11:17:38.169127: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
1: 2017-10-19 11:17:49.647909: I tensorflow/contrib/gdr/gdr_memory_manager.cc:333] Accepted new RDMA connection
3: 2017-10-19 11:17:49.646921: I tensorflow/contrib/gdr/gdr_memory_manager.cc:678] RDMA endpoint connected to rdma://hsw215.ib.cluster:2301
```
I'm running this on slurm. The first number above is the task id. Here's my cluster setup:
```
CLUSTER_SPEC_DICT: {'ps': ['hsw215.ib.cluster:2300', 'hsw216.ib.cluster:2300'], 'worker': ['hsw215.ib.cluster:2301', 'hsw216.ib.cluster:2301']}
# slurm task 0 - hsw215.ib.cluster:2300 - ps 0
# slurm task 1 - hsw215.ib.cluster:2301 - worker 0
# slurm task 2 - hsw216.ib.cluster:2300 - ps 1
# slurm task 3 - hsw216.ib.cluster:2301 - worker 1
```
The parameter servers run on cpu and each worker has 4 GPUs. Devices list:
```
['/job:ps/task:0', '/job:ps/task:1']
['/job:worker/task:0/device:GPU:0', '/job:worker/task:0/device:GPU:1', '/job:worker/task:0/device:GPU:2', '/job:worker/task:0/device:GPU:3', '/job:worker/task:1/device:GPU:0', '/job:worker
/task:1/device:GPU:1', '/job:worker/task:1/device:GPU:2', '/job:worker/task:1/device:GPU:3']
```
ERRORS:
```
1:       [[Node: _recv_concatenate_1_target_0_S581 = _Recv[client_terminated=false, recv_device="/job:ps/replica:0/task:0/device:GPU:0", send_device="/job:worker/replica:0/task:0/device:CPU:0", send_device_in
carnation=1867630968541887687, tensor_name="edge_3354__recv_concatenate_1_target_0", tensor_type=DT_FLOAT, _device="/job:ps/replica:0/task:0/device:GPU:0"]()]]
1:       [[Node: training/RMSprop/gradients/concatenate_1/concat_grad/Slice_2_S699 = _Recv[client_terminated=false, recv_device="/job:worker/replica:0/task:0/device:GPU:2", send_device="/job:ps/replica:0/task
:0/device:CPU:0", send_device_incarnation=-1707922374272032114, tensor_name="edge_1755_training/RMSprop/gradients/concatenate_1/concat_grad/Slice_2", tensor_type=DT_FLOAT, _device="/job:worker/replica:0/task:
0/device:GPU:2"]()]]
1: 2017-10-19 11:17:49.641150: W tensorflow/core/framework/op_kernel.cc:1192] Unavailable: Cannot find pinned memor
1: y region
1:       [[Node: _recv_concatenate_1_target_0_S581 = _Recv[client_terminated=false, recv_device="/job:ps/replica:0/task:0/device:GPU:0", send_device="/job:worker/replica:0/task:0/device:CPU:0", send_device_in
carnation=1867630968541887687, tensor_name="edge_3354__recv_concatenate_1_target_0", tensor_type=DT_FLOAT, _device="/job:ps/replica:0/task:0/device:GPU:0"]()]]
1:       [[Node: training/RMSprop/gradients/concatenate_1/concat_grad/Slice_2_S699 = _Recv[client_terminated=false, recv_device="/job:worker/replica:0/task:0/device:GPU:2", send_device="/job:ps/replica:0/task
:0/device:CPU:0", send_device_incarnation=-1707922374272032114, tensor_name="edge_1755_training/RMSprop/gradients/concatenate_1/concat_grad/Slice_2", tensor_type=DT_FLOAT, _device="/job:worker/replica:0/task:
0/device:GPU:2"]()]]
1: 2017-10-19 11:17:49.641155: W tensorflow/core/framework/op_kernel.cc:1192] Unavailable: Cannot find pinned memory region
1:       [[Node: _recv_concatenate_1_target_0_S581 = _Recv[client_terminated=false, recv_device="/job:ps/replica:0/task:0/device:GPU:0", send_device="/job:worker/replica:0/task:0/device:CPU:0", send_device_in
carnation=1867630968541887687, tensor_name="edge_3354__recv_concatenate_1_target_0", tensor_type=DT_FLOAT, _device="/job:ps/replica:0/task:0/device:GPU:0"]()]]
1:       [[Node: training/RMSprop/gradients/concatenate_1/concat_grad/Slice_2_S699 = _Recv[client_terminated=false, recv_device="/job:worker/replica:0/task:0/device:GPU:2", send_device="/job:ps/replica:0/task
:0/device:CPU:0", send_device_incarnation=-1707922374272032114, tensor_name="edge_1755_training/RMSprop/gradients/concatenate_1/concat_grad/Slice_2", tensor_type=DT_FLOAT, _device="/job:worker/replica:0/task:
0/device:GPU:2"]()]]

3: 2017-10-19 11:18:25.395597: E tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc:71] Cannot find pinned memory region from allocator cpu_pool
3: 2017-10-19 11:18:25.395852: E tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc:71] Cannot find pinned memory region from allocator cuda_host_bfc
3: 2017-10-19 11:18:25.395913: E tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc:71] Cannot find pinned memory region from allocator cpu_pool
3: 2017-10-19 11:18:25.396150: E tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc:71] Cannot find pinned memory region from allocator cpu_pool
3: 2017-10-19 11:18:25.396200: E tensorflow/contrib/gdr/gdr_rendezvous_mgr.cc:71] Cannot find pinned memory region from allocator cpu_pool
```

byronyi:
@avolkov1 I suspect it is because of the limit on page locked memory size. Check with `ulimit -l` or `ulimit -a` to see the limit of your current user.

avolkov1:
@byronyi Thanks. I looked at it. It seems to be unlimited.
```
hsw228$ ulimit -l
unlimited
hsw228$ ulimit -a
core file size          (blocks, -c) 0
data seg size           (kbytes, -d) unlimited
scheduling priority             (-e) 0
file size               (blocks, -f) unlimited
pending signals                 (-i) 1031351
max locked memory       (kbytes, -l) unlimited
max memory size         (kbytes, -m) unlimited
open files                      (-n) 4096
pipe size            (512 bytes, -p) 8
POSIX message queues     (bytes, -q) 819200
real-time priority              (-r) 0
stack size              (kbytes, -s) unlimited
cpu time               (seconds, -t) unlimited
max user processes              (-u) 4096
virtual memory          (kbytes, -v) unlimited
file locks                      (-x) unlimited
```

byronyi:
@avolkov1 I failed to reproduce your issue on a fresh installation of vendor provided OFED on my testbed. Would you mind to paste the full log?

avolkov1:
@byronyi Will do. Thank you very much. Give me a day or two (busy with work related stuff). I'll post the error log and the code I use to run.

PR #11397, opened 2017-07-09
Third-party Scala API Links in the Documentation

eaplatanios:
I added a couple links to my Scala API repository in the documentation pages. I'm not sure if it's in the expected format, but feel free to rephrase if needed.

It would be great if we could also add a link in the following page (I could not find it in the documentation sources): https://www.tensorflow.org/api_docs/.

Thanks!

EDIT: @asimshankar will take care of adding a link to https://www.tensorflow.org/api_docs/, so we're good with that one. Thanks! :)


tensorflow/docs_src/extend/index.md
```diff
@@ -20,8 +20,9 @@ TensorFlow:
 
 Python is currently the only language supported by TensorFlow's API stability
 promises.  However, TensorFlow also provides functionality in C++, Java, and Go,
-plus community support for [Haskell](https://github.com/tensorflow/haskell)
-and [Rust](https://github.com/tensorflow/rust).  If you'd like to create or
+plus community support for [Haskell](https://github.com/tensorflow/haskell), 
+[Scala](https://github.com/eaplatanios/tensorflow_scala), and
```
asimshankar: Let's revert the change in this file. The languages listed here are maintained by members of the TensorFlow github organization, just don't get the same level of support. I'd like for documentation to be clear between bindings maintained by (and for which support can be expected from) the TensorFlow organization and those owned, maintained and supported by others.

tensorflow/docs_src/programmers_guide/faq.md
```diff
@@ -149,6 +149,8 @@ TensorFlow also has a
 to help build support for more client languages.  We invite contributions of new
 language bindings.
 
+An experimental Scala API that replicates a lot of the Python API functionality using the C API is located at [eaplatanios/tensorflow_scala](https://github.com/eaplatanios/tensorflow_scala).
```
asimshankar: Here too I'd like to ensure that the reader is clear about the distinction between bindings that are maintained by (and for which they can expect support for) by the TensorFlow organization and those owned, maintained and supported by others.

How about the following instead:

Bindings for various other languages (such as [C#](https://github.com/migueldeicaza/TensorFlowSharp), [Julia](https://github.com/malmaud/TensorFlow.jl), [Ruby](https://github.com/somaticio/tensorflow.rb) and [Scala](https://github.com/eaplatanios/tensorflow_scala)) created and supported by the opensource community build on top of the [C API](https://www.tensorflow.org/code/tensorflow/c/c_api.h) supported by the TensorFlow maintainers.

asimshankar (changes requested):

eaplatanios:
@asimshankar I made the changes you requested. Thanks for clarifying which part of the documentation refers to officially maintained libraries and third-party ones. :)

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please

PR #11406, opened 2017-07-10
modify SaverDef default version with v2

horance-liu:
SaverDef V1 has been deprecated, so modify default version with V2.


tensorflow/core/protobuf/saver.proto
```diff
@@ -37,9 +37,9 @@ message SaverDef {
   enum CheckpointFormatVersion {
     // Internal legacy format.
     LEGACY = 0;
-    // Current format: tf.Saver() which works with tensorflow::table::Table.
+    // Deprecated format: tf.Saver() which works with tensorflow::table::Table.
     V1 = 1;
-    // Experimental format under development.
+    //Current format: more efficient.
```
frankchn: Nit, a space between `//` and `Current`

frankchn (changes requested):

tensorflow/core/protobuf/saver.proto
horance-liu: @frankchn Sorry, I have modified to commit it, please review again.

frankchn:
Can you rebase to master and then apply your changes rather than merging from master? Right now it shows that you are merging in a bunch of changes that are irrelevant to you.

PR #11408, opened 2017-07-10
[XLA] Add filtering of tests for 3rd party devices on the python suite

DavidNorman:
Currently the tests are marked with specific exclusions for devices.

For 3rd party devices where we don't want their names and configs in the public repo, this change allows the devices to specify which tests they want to exclude.

No change for the CPU/GPU builtin devices.

### DIFF
--- a/tensorflow/compiler/tests/build_defs.bzl
+++ b/tensorflow/compiler/tests/build_defs.bzl
@@ -40,6 +40,10 @@ def tf_xla_py_test(name, srcs=[], deps=[], tags=[], data=[], main=None,
   if disabled_backends == None:
     disabled_backends = []
 
+  for p in plugins:
+    if name in plugins[p]["disabled"]:
+      disabled_backends += [p]
+
   enabled_backends = [b for b in all_backends() if b not in disabled_backends]
   test_names = []
   for backend in enabled_backends:
--- a/tensorflow/compiler/tests/plugin.bzl
+++ b/tensorflow/compiler/tests/plugin.bzl
@@ -18,6 +18,11 @@
 #   git update-index --assume-unchanged tensorflow/compiler/tests/plugin.bzl
 
 plugins = {
-  #"poplar": {"device":"XLA_IPU", "types":"DT_FLOAT,DT_INT32", "tags":[]},
+  #"example": {
+  #  "device":"XLA_EXAMPLE",
+  #  "types":"DT_FLOAT,DT_INT32,DT_INT64",
+  #  "tags":[],
+  #  "disabled": ["a_test","a_n_other_test"],
+  #},
 }
 


frankchn:
Jenkins, test this please

hawkinsp (commented):
It turns out we actually already have a mechanism for disabling tests on a plugin backend via a manifest file:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tests/xla_test.py#L76

Could you use that mechanism instead?

You create a file with a list of lines like this:
DisabledTest.ATestCase  # Does not work because XYZ

and pass its details to the test via a flag (which we normally configure using the .bzl logic).

That said I don't think we are particularly wedded to the current mechanism and we could look into specifying the disabled tests in a .bzl file instead.

DavidNorman:
no problem - will check this method out.  better to use an existing scheme where one exists

DavidNorman:
There's a few things I don't get:

how do you pass the flag that names the manifest file into the build?  do you do a point edit to the build_defs.bzl file?

it looks to me (without actually trying it though) that the manifest file is consumed at test run time, which means that it needs to get into the runfiles for each test.  how is this achieved?


cheers

DavidNorman:
@hawkinsp 

I think I have answered my questions.  It is possible that my scheme is a bit cleaner than the existing one.   Could they both co-exist?

vrv:
I'm on PR rotation this week, but as a passerby I think it makes sense to have one mechanism that works for everyone, so if this version is cleaner, it is worth porting that mechanism to use this one too.

hawkinsp:
I would like to do something to consolidate the two schemes rather than having two ways to disable tests.

The main feature missing from the approach in the PR is the ability to disable individual tests. We use this fairly frequently if there is just one test in a suite that is broken for some reason. This does require some runtime reasoning --- it's not sufficient to disable at the build rule granularity.

To inject the extra manifest file into the tests, we have a local modification to the tf_xla_py_test() build rule. It's nothing fancy --- it just adds an extra argument to the py_test() targets it generates. Is there any reason we couldn't do something similar for your plugin tests --- have the plugin description point to a manifest file instead?

DavidNorman:
ok - i'll have another, harder, look at it.

hawkinsp:
For reference here's what we do for an experimental internal backend.

The BUILD file in the directory containing the manifest has:

```
exports_files([
    "disabled_manifest.txt",
])
```

The build_defs.bzl contains the following logic in `tf_xla_py_test` for the additional backend:
```
      backend_args += [
          ...
          "--disabled_manifest=path/to/my/disabled_manifest.txt",
      ]
      backend_data += ["//path/to/my:disabled_manifest.txt"]
```

This only works for Python tests --- we have a separate mechanism for C++ tests. The C++ mechanism is also based around a manifest file, but we never bothered to opensource that code with the rest of XLA since it didn't seem useful to anyone else. If you would like, I can opensource that code.

DavidNorman:
ok.  i didn't realise that I could dynamically add a dependency to a target in any dir (using backend_data).   i will change my code to look very similar to that (where the contents of the backend_args and backend_data are added from the plugin config.

don't worry about the C++ tests. i don't run them for my own backend right now because too many are failing.  I looked,briefly, at some of the failures.  I saw some 64 bit assumptions and some things that we don't support currently.  i'm not too worried at the moment.

DavidNorman:
@vrv @hawkinsp 

I've created a different PR for a method that uses the disabled-manifest

https://github.com/tensorflow/tensorflow/pull/11826

Issue #11339, opened 2017-07-07
Implement Audio Ops for Python Client


Abhipray:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OS X 10.10.5
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: 1.2.1
- **Python version**: 3.5.2

There are some very useful audio operations defined in tensorflow/core/ops/audio_ops.cc. It would be great if python client users could access these by default.

asimshankar:
@petewarden would know about any plans for the state of these operations and whether they can be made available in the Python API

andykernahan:
@Abhipray @petewarden FYI I have recently begun work on [this](../compare/master...andykernahan:wav-op-fixes). There are a couple of outstanding tasks:
* Add tests for the `audio_ops.py`
* Add tests cases for multi-channel inputs.
* Determine why [`AudioSpectrogramOp`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/spectrogram_op.cc) fails with `Spectrogram size calculation failed` for multi-channel inputs.

rryan:
Ahh hm, so there's a reason that those ops don't have Python wrappers yet -- we don't actually want to make them public / supported yet. I'm sorry this decision wasn't better documented, but could you please hold off on this work?

andykernahan:
Ah, that's a shame; may I inquire as to the reason why? Presumably the ops & kernels will remain in core, so it's safe to take a dependency on them in a saved model?

petewarden:
We are planning on making them public soon, I have a pending change that I hope to get in over the next couple of weeks.

andykernahan:
Excellent, thanks.

In the mean while, I identified a small [issue](/andykernahan/tensorflow/commit/0c0152d31a5e02f0025b62d1c715dc3cecb7c876) with the `AudioSpectrogram` op shape function; should I submit a PR for this, or leave it?

cchandak:
@petewarden @rryan Is the python wrapper for mfcc available in the 1.3 release of tensorflow ? Thanks

Joseph1977:
Trying the "Simple Audio Recognition", but get the " ImportError: cannot import name 'audio_ops'", Any way or plans to make it work?

petewarden:
Are you using the latest development version of TensorFlow? https://hub.docker.com/r/tensorflow/tensorflow/

[...]

PR #11412, opened 2017-07-10
WAV encoding/decoding fixes.

andykernahan:
Hi,

This PR contains two WAV related fixes.

Fixes WAV encoding/decoding for multi-channel data:
* `DecodeLin16WaveAsFloatVector` was failing for multi-channel data with “Bad bytes per second” due to it expecting the byte rate to be that for a single channel, whereas the header specifies the byte rate for _all_ channels.
* `EncodeAudioAsS16LEWav` was incorrectly encoding the byte rate as that for a single channel instead of for _all_ channels.

Fixes shape function for the `DecodeWav` op:
* The function was returning a non-ok status when either `desired_{samples,channels}` attributes were default; now, we use the unknown-dim instead.

Cheers.

### DIFF
--- a/tensorflow/core/kernels/decode_wav_op_test.cc
+++ b/tensorflow/core/kernels/decode_wav_op_test.cc
@@ -23,6 +23,7 @@ limitations under the License.
 #include "tensorflow/cc/ops/audio_ops.h"
 #include "tensorflow/cc/ops/const_op.h"
 #include "tensorflow/cc/ops/math_ops.h"
+#include "tensorflow/core/framework/shape_inference_testutil.h"
 #include "tensorflow/core/framework/tensor_testutil.h"
 #include "tensorflow/core/framework/types.h"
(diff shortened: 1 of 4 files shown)


frankchn:
Jenkins, test this please

rryan (approved):
LGTM -- thank you!

I think this fix is technically backwards incompatible but IMHO it's fine since:
* It doesn't affect anyone doing mono (encoding or decoding).
* We're in the wrong here (producing incorrect WAV files). 
* EncodeWav/DecodeWav aren't really public ops since they have no Python wrapper even though they're part of core.

andykernahan:
@rryan Thanks!

I'm slowly working towards adding Python wrappers for the audio_ops (mentioned in #11339); these issues were identified as part of that work and I thought it best to fix them in isolation.

rryan:
Ahh sorry I haven't seen that issue yet. I'll comment on there.

PR #11425, opened 2017-07-11
Imply default argument in docstring properly

carlthome:
Since the default is `as_text=True` the current docstring is a real gotcha IMO. Clearer to flip the explanation around.


frankchn:
Jenkins, test this please.

frankchn (approved):

PR #11429, opened 2017-07-11
modify SaverDef default version with v2

horance-liu:
SaverDef V1 has been deprecated, so modify default version with V2.


frankchn:
Jenkins, test this please.

frankchn (approved):

PR #11432, opened 2017-07-11
Enable building grpc+verbs runtime on any Linux box

byronyi:
I don't have a Windows machine so it is probably broken on Windows. Currently I don't have an idea on how to achieve cross-platform portability. Any suggestions?


junshi15:
Is there any advantage to combine the system-level library? If one wants to compile verbs, hopefully he has the hardware, or at least install the verbs library first. For those boxes that has verbs library pre-installed, if the existing verbs version is different from the compiled one listed in your PR, will that be a problem. My other concern is that tensor flow compile time is already long, this just adds to that.

PR #11437, opened 2017-07-11
Fix unpickling/copying tf.app.flags.FLAGS

vmarkovtsev:
pickle.load() and copy.copy() check for the presence of  __setstate__().
The problem is that this check is made in the freshly allocated instance which
has not been __init__()-ed. Thus it's __dict__ is completely empty and
__getattr__() fail with KeyError.

The fix is to check if __parsed is in the __dict__ and raise
AttributeError if it isn't.

### DIFF
--- a/tensorflow/python/platform/flags.py
+++ b/tensorflow/python/platform/flags.py
@@ -44,7 +44,12 @@ class _FlagValues(object):
 
   def __getattr__(self, name):
     """Retrieves the 'value' attribute of the flag --name."""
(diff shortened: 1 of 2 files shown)


frankchn:
Jenkins, test this please.

vmarkovtsev:
Looks like `//tensorflow/python/kernel_tests:barrier_ops_test` is flacky on macOS or just too little timeout - no problems apart from it.

frankchn (approved):

PR #11465, opened 2017-07-13
Use Windows Threadpool for SchedClosure

snnn:
Before this change, the distributed runtime will create a new thread for each mini-batch. Quite wasteful.


mrry (approved):
This change looks great! Thanks for the fix... and I wish we had a similar feature for the POSIX version :).

mrry:

PR #11477, opened 2017-07-13
dequeue or dequeue_many will fail immediately after close the queue

horance-liu:
After close the queue, subsequent `dequeue` and `dequeue_many` operations that won't block waiting for more elements to be enqueued and will fail immediately.

### DIFF
--- a/tensorflow/python/ops/data_flow_ops.py
+++ b/tensorflow/python/ops/data_flow_ops.py
@@ -513,7 +513,8 @@ class QueueBase(object):
     operations will fail. Subsequent `dequeue` and `dequeue_many`
     operations will continue to succeed if sufficient elements remain
     in the queue. Subsequent `dequeue` and `dequeue_many` operations
-    that would block will fail immediately.
+    that won't block waiting for more elements to be enqueued and 
+    will fail immediately.
 
     If `cancel_pending_enqueues` is `True`, all pending requests will also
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/data_flow_ops.py
```diff
@@ -513,7 +513,8 @@ def close(self, cancel_pending_enqueues=False, name=None):
     operations will fail. Subsequent `dequeue` and `dequeue_many`
     operations will continue to succeed if sufficient elements remain
     in the queue. Subsequent `dequeue` and `dequeue_many` operations
-    that would block will fail immediately.
+    that won't block waiting for more elements to be enqueued and 
+    will fail immediately.
```
frankchn: Perhaps something like:

Subsequently `dequeue` and `dequeue_many` operations that would otherwise block waiting for more elements (if `close` hadn't been called) will now fail immediately.

will make it clearer.

frankchn (changes requested):

tensorflow/python/ops/data_flow_ops.py
horance-liu: the problem is more clear after adding condition that indicates that the queue is closed.

frankchn (approved):

frankchn:
Jenkins, test this please.

PR #11481, opened 2017-07-13
Fix on 'Use mcpu instead of march for ppc64le'

elefthei:
Use of [] around 'if is_ppc64le; then' makes it always return true, 
causing "deprecated: -mcpu" gcc warnings on every other arch.
This fixes commit 9f57dc8

### DIFF
--- a/configure
+++ b/configure
@@ -11,6 +11,7 @@ SOURCE_BASE_DIR=`pwd -P`
 popd > /dev/null
 
 PLATFORM="$(uname -s | tr 'A-Z' 'a-z')"
+CPU="$(uname -m)"
 
 function is_linux() {
   [[ "${PLATFORM}" == "linux" ]]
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -26,7 +27,7 @@ function is_windows() {
 }
 
 function is_ppc64le() {
-  [[ "${uname -m}" == "ppc64le" ]]
+  [[ "${CPU}" == "ppc64le" ]]
```
caisq: Instead of defining another global variable, which has the risk of clashing at some point in the future, it's safer to do:
[[ "$(uname -m)" == "ppc64le" ]]

configure
```diff
@@ -11,6 +11,7 @@ SOURCE_BASE_DIR=`pwd -P`
 popd > /dev/null
 
 PLATFORM="$(uname -s | tr 'A-Z' 'a-z')"
+CPU="$(uname -m)"
```
caisq: Remove this line. See comment below.

caisq (changes requested):

caisq (approved):

caisq:

caisq:
PR merged. Thanks, @elefthei

PR #11455, opened 2017-07-12
update word2vec_basic.py:	Fix TSNE invocation in Udacity word2vec assignment

Shuolongbj:
Update 5_word2vec.ipnb-

Fix TSNE invocation in Udacity word2vec assignment

#add method='exact' .fix ValueError

num_points = 400
tsne = TSNE(perplexity=30, n_components=2, init='pca',n_iter=5000,**method='exact'**)
two_d_embeddings = tsne.fit_transform(final_embeddings[1:num_points+1, :])

PR #11482, opened 2017-07-13
Fix TSNE invocation in Udacity word2vec assignment

vincentvanhoucke:
Default TSNE method recently changed to the approximate `method='barnes_hut'` which fails to converge per https://github.com/tensorflow/tensorflow/pull/11455


Shuolongbj:
thank you!

frankchn (approved):

PR #11492, opened 2017-07-14
R1.2

pipigenius:


frankchn:
I think this is an erroneous pull request. The 1.2.1 documentations has already been updated.

PR #11503, opened 2017-07-14
Branch 161949540

caisq:


frankchn (approved):

ebrevdo (approved):

agarwal-ashish (approved):
Looks good for contrib/factorization changes.

tensorflow/core/ops/ops.pbtxt
```diff
@@ -28468,33 +28468,6 @@ op {
   is_stateful: true
 }
 op {
-  name: "LMDBReader"
```
alextp: Is it safe to remove an op?

caisq:
@alextp these deletions to ops.pbtxt are auto-generated.

Looks like @jhseu should know about the underlying change.

caisq:
@allenlavoie FYI, tensorflow/contrib/timeseries/python/timeseries:state_management_test is segfaulting in the PIP build:
https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu-python3/5400/consoleFull

testing again to see whether it is reproducible.

tensorflow/core/ops/ops.pbtxt
jhseu: Hmm, yeah, this is a surprising deletion. Let's track down why it happened?

jhseu: Ah, it looks like it just removed a duplicate entry, so this seems fine.

jhseu (approved):

alextp (approved):

tensorflow/core/ops/ops.pbtxt
caisq: Thanks for investigating this, @jhseu

caisq:
@allenlavoie Looks the failure in state_management_test is random. There may be other tests in contrib/timeseries that have the same problem. We will keep an eye on them. I'm merging this PR now.

allenlavoie:
Yeah, I can reproduce the same kind of flakiness for that test locally (passes, but segfaults after) without this PR. Odd since we're not using any timeseries-specific custom ops and it's not at all flaky internally. I'll try to track it down.

PR #11507, opened 2017-07-14
Fix syntax error in sample code

CatalinVoss:
This would have otherwise caused a non-keyword arg after keyword arg


MyHumbleSelf:
There are actually a bunch more issues in this sample code. I can fix some of these (or the author can take a look) later

frankchn (approved):

frankchn:
@MyHumbleSelf definitely feel free to fix anything that you think is wrong.

Jenkins, test this please.

PR #11510, opened 2017-07-15
Add "name" variable support for tf.einsum

### LINKED ISSUE #11501: tf.einsum does not have "name" variable
I noticed there is no ability to name a tensor generated from from tf.einsum(). Is this on purpose?

yongtang:
This fix adds "name" variable support for tf.einsum

This fix fixes #11501

### DIFF
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -156,7 +156,7 @@ def einsum(equation, *inputs, **kwargs):
   if kwargs:
     raise TypeError("invalid keyword arguments for this function: " +
                     ", ".join([format(key)
-                               for key in reversed(list(kwargs.keys()))]))
+                               for key in sorted(list(kwargs.keys()))]))
   with ops.name_scope(name, "einsum", [equation, inputs]) as name:
     if '...' in equation:
       raise ValueError('Subscripts with ellipses are not yet supported.')
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -154,8 +154,9 @@ def einsum(equation, *inputs, **kwargs):
   """
   name = kwargs.pop("name", None)
   if kwargs:
-    raise TypeError("invalid keyword arguments for this function: "
-        +", ".join([format(key) for key in reversed(list(kwargs.keys()))]))
+    raise TypeError("invalid keyword arguments for this function: " +
+                    ", ".join([format(key)
+                               for key in reversed(list(kwargs.keys()))]))
   with ops.name_scope(name, "einsum", [equation, inputs]) as name:
     if '...' in equation:
       raise ValueError('Subscripts with ellipses are not yet supported.')
--- a/tensorflow/python/ops/special_math_ops.py
+++ b/tensorflow/python/ops/special_math_ops.py
@@ -82,7 +82,7 @@ def lbeta(x, name='lbeta'):
     return result
 
 
-def einsum(equation, *inputs):
+def einsum(equation, *inputs, **kwargs):
   """A generalized contraction between tensors of arbitrary dimension.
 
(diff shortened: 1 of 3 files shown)


tensorflow/python/ops/special_math_ops.py
```diff
@@ -82,7 +82,7 @@ def lbeta(x, name='lbeta'):
     return result
 
 
-def einsum(equation, *inputs):
+def einsum(equation, *inputs, **kwargs):
```
vrv: Is it possible to avoid the use of the **kwargs and just add name=None instead?

yongtang:
@vrv Thanks for the review. If we use `name=None` at the end then it will thrown out a syntax error:
```
  File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/ad1e09741bb4109fbc70ef8216b59ee2/execroot/tensorflow/bazel-out/local-opt/bin/tensorflow/python/special_math_ops_test.runfiles/org_tensorflow/tensorflow/python/ops/special_math_ops.py", line 85
    def einsum(equation, *inputs, name=None):
                                     ^
SyntaxError: invalid syntax
```

So I used `**kwargs` to get around it.

tensorflow/python/ops/special_math_ops.py
```diff
@@ -151,70 +152,76 @@ def einsum(equation, *inputs):
...
-  inputs = list(inputs)
-  input_axis_labels = match.group(1).split(',')
-
-  if len(inputs) != len(input_axis_labels):
-    raise ValueError('Got %d arguments for equation "%s", expecting %d' % (
-        len(inputs), equation, len(input_axis_labels)))
+  name = kwargs.pop("name", None)
+  if kwargs:
+    key = list(kwargs.keys())[0]
+    raise TypeError("'{}' is an invalid keyword argument "
```
vrv: Should we raise an error for all the keys passed, not just the first one?

vrv:

yongtang:
Thanks @vrv for the review. The PR has been updated so that all invalid argument keys would be printed in case of error. A test case has been added as well.

I also updated with
```
   $ bazel-bin/tensorflow/tools/api/tests/api_compatibility_test \
          --update_goldens True
```
so that api_compatibility_test could pass on Jenkins.

vrv:
Thanks!  This looks good.

I'm not sure if this qualifies as something that needs API review -- presumably it does and it should be fine.

tensorflow/python/ops/special_math_ops.py
josh11b: It appears to be unavoidable, see https://stackoverflow.com/questions/20961100/passing-default-parameters-with-varargs-in-function-in-python

josh11b:
Fine for API review.

rmlarsen:

rmlarsen:
@yongtang It looks like you need to fix the expected error message in your test:

  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-cpu-python3/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local-py3-opt/bin/bazel_pip/tensorflow/python/special_math_ops_test.runfiles/org_tensorflow/bazel_pip/tensorflow/python/ops/special_math_ops_test.py", line 251, in test_invalid_keyword_arguments
    invalid1="value1", invalid2="value2")
**AssertionError: "invalid keyword arguments for this function: invalid1, invalid2" does not match "invalid keyword arguments for this function: invalid2, invalid1"**

yongtang:
@rmlarsen Thanks for the review. The Jenkins error for was because dict in python is not ordered when iterate through. I have sorted the error message so that the output is repeatable. The PR has been updated now.

rmlarsen:
@yongtang thanks for the quick fix. Yeah, iterating over and unordered container is asking for trouble :-)

rmlarsen (approved):

PR #11511, opened 2017-07-15
Remove RTLD_GLOBAL when loading pywrap_tensorflow

allenlavoie:
Note: I'm making this a pull request to test it. I don't propose merging as-is, and this will likely need to be split up to be reviewed sensibly.

Splits TensorFlow into //tensorflow/core:framework and //tensorflow/core:lib shared objects. Custom ops will be able to rely on these shared objects (e.g. for op registration) rather than relying on the global symbol table. Should enable custom ops in languages other than Python, among other things.


allenlavoie:
Jenkins, test this please.

allenlavoie:
Ugh, I'll figure out how to amend the email for those first few commits on Monday.

Issue #11519, opened 2017-07-15
import_pb_to_tensorboard.py fails with TypeError


daj:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No.
- **OS Platform and Distribution**: Mac OSX Sierra 10.12.5
- **TensorFlow installed from**: Source.
- **TensorFlow version**: ('v0.10.0-1727-g484ca8a-dirty', '0.11.0rc0')
- **Python version**: 2.7.6
- **Bazel version**: 0.4.4-homebrew
- **CUDA/cuDNN version**: N/A
- **GPU model and memory**: N/A
- **Exact command to reproduce**:
```
python import_pb_to_tensorboard.py --model_dir ~/Code/AndroidTensorFlowMNISTExample/app/src/main/assets/mnist_model_graph.pb --log_dir /tmp/tensorboard/
```

### Describe the problem
I'm trying to use `import_pb_to_tensorboard.py` to import an existing TensorFlow model into TensorBoard.  It appears to always return a TypeError due to line 74:

```
python import_pb_to_tensorboard.py --model_dir ~/Code/AndroidTensorFlowMNISTExample/app/src/main/assets/mnist_model_graph.pb --log_dir /tmp/tensorboard/
Traceback (most recent call last):
  File "import_pb_to_tensorboard.py", line 74, in <module>
    app.run(main=main, argv=[sys.argv[0]] + unparsed)
TypeError: run() got an unexpected keyword argument 'argv'
```

I discovered `import_pb_to_tensorboard.py` in [this Stackoverflow answer](https://stackoverflow.com/a/44955005/112705).

asimshankar:
Marking this as "Contributions Welcome" since it seems you already have a PR to improve the tool.

iskenderunlu:
Is there anyone who can assign the problem to me ? Thanks

AswinBlue:
I recommend you to reference these methods.
It must be run 0.12 tensorflow.

https://github.com/tensorflow/tensorflow/blob/r0.12/tensorflow/examples/learn/text_classification.py

https://github.com/tensorflow/tensorflow/blob/r0.12/tensorflow/python/platform/app.py

AswinBlue:
I recommend you to reference these methods.
It must be run 0.12 tensorflow.

https://github.com/tensorflow/tensorflow/blob/r0.12/tensorflow/examples/learn/text_classification.py
[...]

PR #11520, opened 2017-07-15
Show usage when no arguments passed to import_pb_to_tensorboard.py

daj:
It seems like `--model_dir` and `--log_dir` should be required arguments.  After adding at least one required argument, then `parseargs` handles no arguments cleanly.

TESTING

Running it before the fix:
```
$ python import_pb_to_tensorboard.py 
  File "import_pb_to_tensorboard.py", line 74, in <module>
    app.run(main=main, argv=[sys.argv[0]] + unparsed)
TypeError: run() got an unexpected keyword argument 'argv'
```

Running it after the fix:
```
$ python import_pb_to_tensorboard.py 
usage: import_pb_to_tensorboard.py [-h] --model_dir MODEL_DIR --log_dir
                                   LOG_DIR
import_pb_to_tensorboard.py: error: argument --model_dir is required
```

I was unable to test the success path due to this issue: https://github.com/tensorflow/tensorflow/issues/11519

Testing done on Mac OSX Sierra 10.12.5 using Python 2.7.6.

ALTERNATIVE SUGGESTION

Let me know if you prefer that no arguments are required.  If that is the case, I was considering this solution (from [this Stackoverflow answer](https://stackoverflow.com/a/29293080/112705)):
```
  try:
    app.run(main=main, argv=[sys.argv[0]] + unparsed)
  except:
    # Print the usage if the user provides no arguments
    parser.print_help()
    sys.exit(0)
```

OTHER THOUGHTS

It seems like `model_dir` should maybe be renamed to `model_path` (assuming that it should link the file, not the directory which it is in)?

### DIFF
--- a/tensorflow/python/tools/import_pb_to_tensorboard.py
+++ b/tensorflow/python/tools/import_pb_to_tensorboard.py
@@ -64,11 +64,13 @@ if __name__ == "__main__":
       "--model_dir",
       type=str,
       default="",
+      required=True,
       help="The location of the protobuf (\'pb\') model to visualize.")
   parser.add_argument(
       "--log_dir",
       type=str,
       default="",
+      required=True,
       help="The location for the Tensorboard log to begin visualization from.")
   FLAGS, unparsed = parser.parse_known_args()
   app.run(main=main, argv=[sys.argv[0]] + unparsed)


frankchn:
Jenkins, test this please.

petewarden (approved):
Thanks, looks good!

PR #11530, opened 2017-07-16
Add SquaredDifference gradient

DimanNe:
I added gradient function for the [SquaredDifference](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/squared-difference) and test case.


frankchn:
Jenkins, test this please.

tensorflow/cc/gradients/math_grad_test.cc
```diff
@@ -737,6 +737,108 @@ TEST_F(CWiseUnaryComplexGradTest, Conj) {
...
+    const Output x = Const(scope, GetRandTensor(tensor_shape));
+    const Output y = Const(scope, GetRandTensor(tensor_shape));
+    const Output df = Const(scope, GetRandTensor(tensor_shape));
+    const Output expected_df_dx = Const(
+        scope, GetExpectedGradient(scope, tensor_shape, x, y, df, calc_df_dx));
+    const Output expected_df_dy = Const(
+        scope, GetExpectedGradient(scope, tensor_shape, x, y, df, calc_df_dy));
+    const Output f = get_function_to_test(scope, x, y);
+
+    // All data & info have been collected => run test itself:
```
drpngx: remove. This should be straightforward from the code and is vulnerable to getting stale after refactorings.

tensorflow/cc/gradients/math_grad_test.cc
```diff
@@ -737,6 +737,108 @@ TEST_F(CWiseUnaryComplexGradTest, Conj) {
...
+
+    std::vector<Output> actual_gradients;
+    TF_ASSERT_OK(test::CallGradFunction(scope, Operation(f.node()), {df},
+                                        &actual_gradients));
+
+    auto check = [&scope](const Output& actual_gradient,
```
drpngx: const auto tensor_expect_close

tensorflow/cc/gradients/math_grad.cc
```diff
@@ -463,6 +463,30 @@ Status BatchMatMulGrad(const Scope& scope, const Operation& op,
 }
 REGISTER_GRADIENT_OP("BatchMatMul", BatchMatMulGrad);
 
+Status SquaredDifferenceGrad(const Scope& scope, const Operation& op,
```
drpngx: While this seems right, this is not what the [python](https://www.github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py#L826) does, right?

The C++ and python bindings have to match.

drpngx (changes requested):
Thanks! The main issue is py/c++ compatibility.

DimanNe:
I noticed that they in [python](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py#L826) have over a dozen(!) of instances of copy-pasted lines for reshaping gradients due to broadcasting feature, and decided to make a helper class `ReshapeGradientsWRTBroadcasting` that will make everything for me.

tensorflow/cc/gradients/math_grad.cc
```diff
@@ -463,6 +464,53 @@ Status BatchMatMulGrad(const Scope& scope, const Operation& op,
 }
 REGISTER_GRADIENT_OP("BatchMatMul", BatchMatMulGrad);
 
+class ReshapeGradientsWRTBroadcasting {
```
drpngx: Probably ReshapGradientsForBroadcast is a better name.

tensorflow/cc/gradients/math_grad.cc
```diff
@@ -463,6 +464,53 @@ Status BatchMatMulGrad(const Scope& scope, const Operation& op,
...
+ public:
+  ReshapeGradientsWRTBroadcasting(const ::tensorflow::Scope& scope,
```
drpngx: It's fine to use just `const Scope& scope`.

tensorflow/core/lib/random/random.cc
```diff
@@ -22,17 +22,28 @@ limitations under the License.
 namespace tensorflow {
 namespace random {
 
-std::mt19937_64* InitRng() {
+namespace {
```
drpngx: This should be split into another PR

drpngx (changes requested):

vrv:
@DimanNe friendly ping!  Thanks

DimanNe:
I forgot to squash commits while merging, sorry.

@vrv  As far as I understood this PR was blocked by [this PR](https://github.com/tensorflow/tensorflow/pull/11634), so I could not do anything here.

DimanNe:
Did something go wrong with Sanity Checks? Or have I done something wrong?

rmlarsen:

rmlarsen:
@drpngx Did @DimanNe address your comments? 
@vrv Don't we need to also define the second derivative as well when moving something like this to a C++ kernel?

vrv:
Yes, if you add a new op that fuses existing python ops, you need to add the gradient for the fused op.

rmlarsen:
@DimanNe Can you please add a gradient op for the new (gradient) op?

DimanNe:
@rmlarsen Yes, sure, I will do that.

rmlarsen:
@DimanNe thanks!

DimanNe:
@rmlarsen Sorry for bothering, but I realized that I had misunderstood something. Can you, please, clarify where I should add gradient operation for the new gradient operation(s) `+-2(x-y)`?

For example, in order to make gradient op for the **named** operation `"SquaredDifference"`, I registered a C++ function using macro `REGISTER_GRADIENT_OP`, but how am I supposed to register function computing gradient for two **unnamed** functions `+-2(x-y)`? Or perhaps there is any example of doing it?

If it is not difficult, could you please take a step backward and elucidate why do we need at all gradient op for gradient? What is the rationale behind having hand-written second-order derivatives/gradients? Would not it be enough to have only first-order gradients and do everything else with automatic differentiation?

rmlarsen:
@DimanNe My apologies. You are indeed just creating a composite op using the C++ interface. What I was referring to what cases in the past where people would add a specialized C++ kernel for computing a gradient, in which case the automatic differentiation would have no way of acting on it.

I think what you have here is fine. @skye can you confirm this?

DimanNe:
@rmlarsen , @skye I understand that you are busy, and probably do not want at all to make C++ interface of TensorFlow great again. So, can I do anything to help you and/or to expedite merging the PR?

skye:
Hey sorry, I meant to ping @suharshs but forgot!

DimanNe:
@suharshs any news/thoughts?)

suharshs:
Ah, unfortunately it looks like someone recently just submitted this gradient :(
https://github.com/tensorflow/tensorflow/commit/293020e83df4fdaa8bfb59275cf05028609b92cd
Thanks so much for implementing this though, and let us know if you see anything wrong with what was submitted.

PR #11532, opened 2017-07-16
support "required" for tf.app.flags

ericyue:
support "required" for tf.app.flags


frankchn:
Jenkins, test this please.

vrv:
Thanks for the contribution @ericyue, this is indeed a useful feature for a flags library.

Some background is probably warranted though: TF's flags library is an API port of an internal flags library at Google that I believe predated argparse.  The flags library's existence is solely to make it easier to opensource internal code, since flags.py's API is used throughout Google internally.  But argparse is more than sufficient for most opensource use cases, so I would like to avoid changing the existing flags API, even to enhance it.

Furthermore, I believe there are efforts to opensource the internal flags library, which already has this feature of marking flags as required, and rather than introduce a new API that is incompatible with the internal one, I think it would be a good idea to wait for that library to be released.  If it's the case that several months go by and that library is not released, let's revisit this PR and perhaps plan to make it API compatible with that library.

Let me know if you feel strongly about this right now though, and we can find a way to get this feature in.  In the meantime, I strongly encourage you to use argparse directly.

ericyue:
@vrv  ok, I think you means the  google's python-gflag?   in python-gflag make a flag required by using `gflags.MarkFlagAsRequired` . but tensorflow's flag not implement this api.  is there a chance to replace `require` parameter with `MarkFlagAsRequired` ?

vrv:
@ericyue It's similar to python-gflags, but I'll have to double check that the API is the exact same... give me some time to contact the authors internally and see what the plans are.

PR #11535, opened 2017-07-16
Java API Generics Phase 2

andrewcmyers:
This is the big change for generics. The main classes like Tensor and Output acquire type parameters. The test programs have been updated accordingly. Existing code written against the non-generic API should still mostly work but will generate "raw types" warnings. The setup of the tables in the Types class from phase 1 has been centralized into Types itself to ensure it happens early enough.


frankchn:
Jenkins, test this please.

andrewcmyers:
The strategy of providing backward compatibility through raw types runs into one snag: Java is not willing to let a `List<Tensor<?>>` be used where a `List<Tensor>` is permitted. It's <5-line fix to the Android demo program to make this work, and I expect a similarly small change to other existing code.

The type `Tensor<?>` is admittedly slightly annoying, though accurate. It is possible to avoid having to write it, at the price of adding another class, and backward compatibility would not be regained in any case.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -96,43 +145,43 @@ public static Tensor create(Object obj) {
...
   }
 
   /**
-   * Create a {@link DataType#FLOAT} Tensor with data from the given buffer.
+   * Create a {@link TFFloat} Tensor with data from the given buffer.
    *
    * <p>Creates a Tensor with the given shape by copying elements from the buffer (starting from its
    * current position) into the tensor. For example, if {@code shape = {2,3} } (which represents a
-   * 2x3 matrix) then the buffer must have 6 elements remaining, which will be consumed by this
+   * 2×3 matrix) then the buffer must have 6 elements remaining, which will be consumed by this
```
KrisRoofe: I think `2x3` may be better than `2×3`, Or can we use `2*3` here.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -348,21 +433,21 @@ public boolean booleanValue() {
    * @throws IllegalArgumentException if the tensor is a scalar or if {@code dst} is not compatible
    *     with the tensor (for example, mismatched data types or shapes).
    */
-  public <T> T copyTo(T dst) {
+  public <U> U copyTo(U dst) {
```
KrisRoofe: Why change the `T` to `U`?

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: Sure, I'll change to 2x3.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: The type parameter U here is not the type parameter of the `Tensor` type itself, but rather a parameter of the method `copyTo`. A different name is used to avoid confusing these two types.

LF-DevJourney (approved):

rmlarsen:

rmlarsen:
@andrewcmyers it looks like there is some bugs in the GPU build, e.g.

https://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/6226/consoleFull
```
ERROR: /workspace/tensorflow/java/BUILD:194:1: Building tensorflow/java/ConstantTest.jar (1 source file) failed: Worker process sent response with exit code: 1.
tensorflow/java/src/test/java/org/tensorflow/op/core/ConstantTest.java:51: warning: [rawtypes] found raw type: Tensor
      Tensor result = sess.runner().fetch(op.asOutput()).run().get(0);
      ^
  missing type arguments for generic class Tensor<T>
  where T is a type-variable:
    T extends Object declared in class Tensor
tensorflow/java/src/test/java/org/tensorflow/op/core/ConstantTest.java:53: warning: [unchecked] unchecked call to <U>copyTo(U) as a member of the raw type Tensor
      assertArrayEquals(ints, result.copyTo(actual));
                                           ^
...
```

andrewcmyers:
Thanks for the diagnosis, this looks like a (semantic) conflict with other recent commits that have happened in parallel. I'll resolve it.

andrewcmyers:
@rmlarsen This should build and test successfully now.

rmlarsen:
@andrewcmyers Thanks!

rmlarsen:
@asimshankar could you please review/approve this?

andrewcmyers:
@rmlarsen I think maybe this PR doesn't have quite the right tags attached to it. Probably needs an "awaiting review"?

asimshankar:
Sorry for the delay, between vacation and some other things this dropped out of my view.
I will plan on taking a look at this next week. Thanks for your patience Andrew!

martinwicke:
@quaeler can you resolve the conflicts? Sorry, it's our bad for leaving this simmer so long.

@asimshankar ping

andrewcmyers:
I should probably do this merge; there are actually some small design decisions.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -614,10 +644,17 @@ private static int elemByteSize(DataType dataType) {
     throw new IllegalArgumentException("DataType " + dataType + " is not supported yet");
   }
 
+  private static void throwExceptionIfNotByteOfByteArrays(Object array) {
```
quaeler: I think there's something amiss with the way this method is being used, at least at [line 667](https://github.com/andrewcmyers/tensorflow/blob/46981fb61c393a1b32b6e4569ac2e3e5c47b240f/tensorflow/java/src/main/java/org/tensorflow/Tensor.java#L667) -- the intent of that codes with the exception of this method appears to be to label a byte[] as a STRING datatype, however how this method is being used will throw an exception before returning the STRING datatype for a non-0-length primitive byte array.

(Also, it would be nice if we used class methods giving component types and array checks instead of seeing whether its toString() value is an arcane shorthand for arrays and bytes.)

quaeler: I realize this isn't even your code change - that it is a merge down due to the code which came in with https://github.com/tensorflow/tensorflow/issues/8531 - so i'll avoid harping on it here. Sorry.

quaeler (approved):

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: @quaeler No problem. Can/should you fire off another build to make sure it all works?

martinwicke:
Jenkins test this please.

quaeler:
Cloning `phase2` locally and running the test which failed also fails for me.
```
JUnit4 Test Runner
...............E
Time: 0.083
There was 1 failure:
1) testUInt8Tensor(org.tensorflow.TensorTest)
java.lang.IllegalArgumentException: class org.tensorflow.types.TFUInt8 is not a TensorFlow type.
        at org.tensorflow.types.Types.dataType(Types.java:40)
        at org.tensorflow.Tensor.create(Tensor.java:117)
        at org.tensorflow.TensorTest.testUInt8Tensor(TensorTest.java:423)
...
```
It ain't lying... the static block in `org.tensorflow.types.Types` does not register the `org.tensorflow.types.TFUInt8` class in its `zeroes` and `typeCodes` maps.

quaeler (approved):

quaeler (approved):
I'm unbelievably loathe to drag this whole PR out any longer; it would have been nice were the unit test commented out as opposed to nuked.

... And there's an angels-on-the-head-of-pin-esque debate as to what impact a lower layer not properly supporting this data type has on whether this unit test should exist as long as it passes here.

That being said, it seems definitely sufficient - let's get this into master so we can move ahead.

quaeler:

quaeler (approved):

quaeler (approved):

andrewcmyers:
Can someone with the privileges kick off Jenkins to test this, please?

There is a lot of history accumulating on this PR. I am happy to create a "fresh" PR with the same changes if that helps move things along.

martinwicke:
Jenkins, test this please.

quaeler:
@martinwicke Is there potentially something wrong with Jenkins - XLA failure cites docker device space issues:
```
Successfully built e0b6e2dbd179
Running 'tensorflow/tools/ci_build/xla/linux/gpu/run_py3.sh' inside jenkins-tensorflow-pull-requests-xla-2627.gpu...
docker: Error response from daemon: devmapper: Thin Pool has 977073 free data blocks which is less than minimum required 983040 free data blocks. Create more free space in thin pool or use dm.min_free_space option to change behavior.
See 'docker run --help'.
Build step 'Execute shell' marked build as failure
```
and the mac failure is what looks like all of the Python-based tests; WRT the latter, i suppose @andrewcmyers could have merged down from master when master was broken, but given that so many tests are failing, it seems like something more Jenkins-systemic is afoot.

martinwicke:
Jenkins, test this please.

Both issues ought to be resolved now. Let's hope for the best.

andrewcmyers:
Hi @asimshankar, I created a smaller PR (#12928) that factors out some of this from the addition of generics.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -26,89 +26,130 @@
...
+   * @param type The class object representing the type T.
    * @throws IllegalArgumentException if {@code obj} is not compatible with the TensorFlow type
-   *     system, or if obj does not disambiguate between multiple DataTypes. In that case, consider
-   *     using {@link #create(DataType, long[], ByteBuffer)} instead.
+   *     system.
+   * @see org.tensorflow.op.Tensors
    */
-  public static Tensor create(Object obj) {
-    return create(obj, dataTypeOf(obj));
+  public static <T extends TFType> Tensor<T> create(Object obj, Class<T> type) {
```
asimshankar: IIUC, this is the factory function to use to get type-checking while the `create(Object)` one will generate runtime errors.

This distinction between when types are specified as `DataType` and when as `Class<T>` is a bit confusing, particularly since this `Tensor` class is providing factory functions for both. What do you think about the following:

- `Tensor.java` will NOT provide compile-time type checking. All the methods do runtime checks and functions use the `DataType` enum for type information
- All the "type-safe" construction is with methods in `Tensors.java`

Subtle, but that way we perhaps more clearly layer the use of `DataType` vs. `Class<T>`?
And then over time we can train folks to use `Tensors` for creating `Tensor` objects.

tensorflow/java/src/main/java/org/tensorflow/Operand.java
```diff
@@ -22,19 +22,19 @@
...
  * // The "decodeJpeg" operation can be used as an operand to the "cast" operation
- * Operand decodeJpeg = ops.image().decodeJpeg(...);
+ * Operand<TFUint8> decodeJpeg = ops.image().decodeJpeg(...);
  * ops.math().cast(decodeJpeg, DataType.FLOAT);
  *
  * // The output "y" of the "unique" operation can be used as an operand to the "cast" operation
- * Output y = ops.array().unique(...).y();
- * ops.math().cast(y, DataType.FLOAT);
+ * Output<TFInt32> y = ops.array().unique(...).y();
+ * ops.math().cast(y, TFFloat);
```
asimshankar: This line doesn't change, I think.
(Or at least let's leave it for now, since the code generation isn't even in place yet, so I'm not quite sure how it will evolve)

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -300,7 +380,7 @@ public int numElements() {
   }
 
   /**
-   * Returns the value in a scalar {@link DataType#FLOAT} tensor.
+   * Returns the value in a scalar {@link TFFloat} tensor.
```
asimshankar: Perhaps leave these comments (and the ones above) unchanged?
I'm thinking that the type-system of a Tensor should always refer to the `DataType` (i.e., the return value of `Tensor.dataType()`). So, this does return the value in a scalar `DataType#FLOAT` tensor, and something like `Tensor<Float>` or `Tensor<TFFloat>` has a "type" of `DataType.FLOAT`.

Does that sound sensible? Separating the "type of the tensor" (from the TensorFlow type system) from the "type parameter" of the class.

tensorflow/java/src/test/java/org/tensorflow/op/ScopeTest.java
```diff
@@ -148,122 +151,137 @@ public void composite() {
...
-    private final Output output;
+  private static final class Variance<T> {
+    private final Output<T> output;
 
-    static Variance create(Scope base, Output x) {
+    static <T extends TFType> Variance<T> create(Scope base, Output<T> x, Class<T> type) {
       Scope s = base.withSubScope("variance");
-      Output zero = Const.create(s.withName("zero"), new int[] {0}).output();
-      Output sqdiff =
+      Output<T> zero = Const.create(base, Types.zeroValue(type), type).output();
```
asimshankar: Is this the only place `zeroValue` is invoked?
Could we do away with `zeroValue` from `Types.java` completely if we made this:

```
static Variance create(Scope base, Output<TFInt32> x, Class<TFInt32> type) {
   Scope s = base.withSubScope("variance");
   Output<TFInt32> zero = Const.create(base, ....)
   ...
```
?

asimshankar (commented):
Thank you very much for your patience!

A few comments. Main points:

- Whether to use Java classes (e.g., `Float` vs. `TFFloat`) where possible (let's discuss this in person)

- Define the "type" of a Tensor. I think we want to refer to the DataType enum for the "type of a tensor". So all comments (particularly in Tensor.java) should talk about the Tensor being a `DataType.INT32` etc. Tensor (and not a `TFUint32` tensor).

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -26,89 +26,130 @@
...
+              + ")");
+    }
+    return create(obj, dt2);
   }
 
   /**
-   * Create a Tensor of data type {@code dtype} from a Java object.
+   * Creates a tensor from an object whose class is inspected to figure out what the underlying data
+   * type should be. The parameter {@code T} must match the data type of the object, but this is
+   * <em>not</em> checked. Not all types {@code T} can be distinguished based on the run-time class
```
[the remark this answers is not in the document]
andrewcmyers: Yes, that could be clearer.

tensorflow/java/src/test/java/org/tensorflow/op/ScopeTest.java
andrewcmyers: Unfortunately not. Variance is a parameterized class -- it's Variance<T>. So Const.create needs to infer from its argument what the type of output is that's being created. I suppose that the (generic) computation of zeroValue, and the table from Types, could be moved into this test if we think it's not generally useful.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: With the design I've proposed, the type DataType is actually mostly invisible and programmers can just use class objects to represent the type of tensors at run time -- rather than having two different notions of type floating around the system. That seems like a win to me.

Now, you might think that we should keep DataType as a very visible entity in the API because of existing code, but as far as I can tell, existing code tends not to talk about DataType because there is no need to: the existing creators in Tensor already automatically decide on datatypes using the object provided at construction time. It seems to me that the stratification you propose will actually make DataType *more* prominent than it currently is, resulting in programmers having two separate ways to think about type.

tensorflow/java/src/test/java/org/tensorflow/TestUtil.java
```diff
@@ -17,35 +17,54 @@
...
           .setAttr("dtype", t.dataType())
           .setAttr("value", t)
           .build()
-          .output(0);
+          .<TFInt32>output(0);
     }
   }
 
-  public static Output placeholder(Graph g, String name, DataType dtype) {
+  public static Output<?> placeholder(Graph g, String name, DataType dtype) {
```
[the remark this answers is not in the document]
andrewcmyers: Yes. This particular declaration was here for backward compatibility but doesn't seem to be getting used any more. Oops.

tensorflow/java/src/main/java/org/tensorflow/examples/LabelImage.java
```diff
@@ -161,46 +168,97 @@ private static int maxIndex(float[] probabilities) {
...
+            .setAttr("dtype", Types.dataType(type))
+            .setAttr("value", t)
+            .build()
+            .<T>output(0);
+      }
     }
 
-    Output constant(String name, Object value) {
-      try (Tensor t = Tensor.create(value)) {
+    Output<TFString> stringConstant(String name, byte[] value) {
```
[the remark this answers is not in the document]
andrewcmyers: You're absolutely right. I've rewritten various similar methods similarly.

tensorflow/java/src/main/java/org/tensorflow/types/Types.java
```diff
@@ -19,20 +19,23 @@
...
- * class object (e.g., TFInt32.class) that represents it at run time. There is
- * also an enumeration value in DataType that can be used to represent the
- * type, though that should rarely be required.
+ * Utility class for managing the representation of TensorFlow types as Java types. For each
+ * TensorFlow type (e.g., int32), there is a corresponding Java type (e.g., TFInt32) that represents
+ * it at compile time and a corresponding class object (e.g., TFInt32.class) that represents it at
+ * run time. There is also an enumeration value in DataType that can be used to represent the type,
+ * though that should rarely be required.
  */
 public class Types {
```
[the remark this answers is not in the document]
andrewcmyers: This sounds okay. I was trying to avoid having DataType depend on this layer of stuff, but I don't have a strong preference.

tensorflow/java/src/test/java/org/tensorflow/TestUtil.java
```diff
@@ -17,35 +17,54 @@
...
+  public static <T extends TFType> Output<T> constant(Graph g, String name, Object value) {
+    try (Tensor<T> t = Tensor.create(value)) {
+      return g.opBuilder("Const", name)
+          .setAttr("dtype", t.dataType())
+          .setAttr("value", t)
+          .build()
+          .<T>output(0);
+    }
+  }
+  public static Output<TFInt32> constant(Graph g, String name, int[][] value) {
```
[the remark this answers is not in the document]
andrewcmyers: Yes, good catch.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: I think we need to discuss this. See my comments about Tensor.java:383.

I worry that the unsafe Tensor interface for creation will always "work", so there will be no impetus for people to switch to the safe Tensors interface.

andrewcmyers: It occurs to me that if the Tensor create operation returns a `Tensor<?>` instead of a `Tensor<T>`, that would give programmers the impetus to use the `Tensors` factory instead. I haven't worked through the implications for all the code that currently uses `Tensor.create`, though.

andrewcmyers: I've explored what it would mean to do this last thing. It turns out there is a lot of code in the test cases that is currently using `Tensor.create` to magically get a `Tensor⟨T⟩` of the right T for the surrounding context (regardless of what is actually in the created tensors, since there is no actual compile-time checking that the object being passed is the right type). If the return type of Tensor.create is changed to Tensor⟨?⟩, all this code has to be fixed by inserting a cast to the appropriate type, e.g., `Tensor⟨TFInt32⟩`. That certainly creates an incentive to switch over to type-safe `Tensors` creators but will be annoying for all the existing tests. Maybe that's the right thing -- people will not switch unless `Tensor.create` becomes painful.

The other option I can think of to incentivize use of the type-safe interface is to deprecate `Tensor.create`.

drpngx:
Jenkins, test this please.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -140,15 +129,14 @@
...
-   * org.tensorflow.Tensors} to create Tensors in a fully type-safe way.
+   * to match {@code type}, but this condition is checked only at run time.
    *
    * @param obj the object supplying the tensor data.
    * @param dtype the data type of the tensor to create. It must be compatible with the run-time
    *     type of the object.
    * @return the new tensor
    */
-  public static Tensor<?> create(Object obj, DataType dtype) {
+  private static Tensor<?> create(Object obj, DataType dtype) {
```
andrewcmyers: This is a new method as of this PR, so nothing is really lost by making it private.

tensorflow/java/src/test/java/org/tensorflow/OperationBuilderTest.java
```diff
@@ -51,7 +49,7 @@ public void failWhenMixingOperationsOnDifferentGraphs() {
   public void failOnUseAfterBuild() {
     try (Graph g = new Graph();
         @SuppressWarnings("unchecked")
-            Tensor<TFInt32> t = (Tensor<TFInt32>) Tensor.create(1)) {
+            Tensor<Integer> t = (Tensor<Integer>) Tensor.create(1)) {
```
andrewcmyers: I've left the test cases creating tensors in various ways that are currently supported, though user code should probably use the `Tensors` class that doesn't exist yet.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -190,47 +224,85 @@ public static Tensor create(long[] shape, DoubleBuffer data) {
...
+   * href="https://www.tensorflow.org/code/tensorflow/c/c_api.h">C API</a>.
+   *
+   * @param <T> The tensor element type
+   * @param type the tensor element type, specified as a DataType. This must agree with T.
+   * @param shape the tensor shape.
+   * @param data a buffer containing the tensor data.
+   * @throws IllegalArgumentException If the tensor datatype or shape is not compatible with the
+   *     buffer
+   */
+  public static Tensor<?> create(DataType dtype, long[] shape, ByteBuffer data) {
```
asimshankar: Let's make this private (ensuring that all creation methods require a `Class` and not a `DataType`)?

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -190,47 +224,85 @@ public static Tensor create(long[] shape, DoubleBuffer data) {
...
       nremaining = data.remaining();
     }
-    Tensor t = allocateForBuffer(dataType, shape, nremaining);
+    Tensor<?> t = allocateForBuffer(dtype, shape, nremaining);
     t.buffer().put(data);
     return t;
   }
 
+  /**
+   * Returns this Tensor object with the type {@code Tensor<U>}. An exception is thrown if the
```
asimshankar: Use the `@throws' clause instead of describing in the main paragraph?

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -577,11 +655,16 @@ private static DataType dataTypeOf(Object o) {
...
+   * whose datatype is {@code dtype}. Normally this is the same as the number of dimensions of o
+   * itself, but is one smaller for tensors of strings.
+   *
+   * @param o The object to inspect. It must be a valid representation of the given data type.
+   * @param dtype The expected data type of the tensor.
    */
   private static int numDimensions(Object o, DataType dtype) {
     int ret = numArrayDimensions(o);
-    if (dtype == DataType.STRING && ret > 0) {
+    if (dtype == DataType.STRING) {
```
asimshankar: Why remove the `&& ret > 0` from here?

tensorflow/java/src/test/java/org/tensorflow/SessionTest.java
```diff
@@ -35,9 +35,10 @@ public void runUsingOperationNames() {
     try (Graph g = new Graph();
         Session s = new Session(g)) {
       TestUtil.transpose_A_times_X(g, new int[][] {{2}, {3}});
-      try (Tensor x = Tensor.create(new int[][] {{5}, {7}});
-          AutoCloseableList<Tensor> outputs =
-              new AutoCloseableList<Tensor>(s.runner().feed("X", x).fetch("Y").run())) {
+      try (@SuppressWarnings("unchecked")
+              Tensor<Integer> x = (Tensor<Integer>) Tensor.create(new int[][] {{5}, {7}});
```
asimshankar: Let's avoide the use of `@SuppressWarnings` in all tests, and use `.expect()` instead

asimshankar (changes requested):
Thanks! This is shaping up well I think. Two high level comments:

- Let's avoid `@SuppressWarnings` in all test/example code and instead use `Tensor.expect()` where needed
- I'm still iffy on the use if `String` as a type parameter - it's the only one for which the type of the argument to `create()` does not match the type parameters. Perhaps introduce an `org.tensorflow.types.Bytes` instead?

@kbsriram @karllessard @ctrueden @dietzc would also probably be interested in the change here.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -28,89 +28,123 @@
...
+    DataType dtype = DataType.fromClass(type);
+    if (!objectCompatWithType(obj, dtype)) {
+      throw new IllegalArgumentException(
+          "Data type of object does not match T (expected "
+              + dtype
+              + ", got "
+              + dataTypeOf(obj)
+              + ")");
+    }
+    return (Tensor<T>) create(obj, dtype);
```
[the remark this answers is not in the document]
andrewcmyers: That would introduce inefficiency because the type gets checked twice. This is the one place where I think we really want to suppress the warning.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: This is a preexisting method so I was leaving it around for backward compatibility. How about we deprecate it?

asimshankar: I believe the only real user of this was `TensorFlowInferenceInterface`, which you've updated. So I'd be tempted to mark this private.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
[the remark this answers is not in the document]
asimshankar: I might be missing something, but it is already being checked twice right now - once in line 108 and then again in line 143. So there would be no added inefficiency?

tensorflow/java/src/main/java/org/tensorflow/DataType.java
```diff
@@ -53,16 +61,40 @@
...
+
   // Cached to avoid copying it
-  final private static DataType[] values = values();
+  private static final DataType[] values = values();
 
   static DataType fromC(int c) {
     for (DataType t : values) {
-      if (t.value == c)
-        return t;
+      if (t.value == c) return t;
```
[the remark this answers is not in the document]
andrewcmyers: Good point. Actually, I did run google-java-format on all files and it doesn't fix this one.

tensorflow/java/src/main/java/org/tensorflow/Operation.java
```diff
@@ -98,16 +98,25 @@ public int outputListLength(final String name) {
...
     for (int i = 0; i < length; ++i) {
       outputs[i] = output(idx + i);
     }
     return outputs;
   }
 
-  /** Returns a symbolic handle to one of the tensors produced by this operation. */
-  public Output output(int idx) {
+  /**
+   * Returns a symbolic handle to one of the tensors produced by this operation. Warning: Does not
```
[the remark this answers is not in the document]
andrewcmyers: Done. For what it's worth, google-format tosses out the self-closing tag and turns it into `<p>`

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -28,89 +28,123 @@
...
-   * {@link DataType#STRING} typed Tensors are multi-dimensionary arrays of arbitrary byte sequences
-   * and thus have {@code byte[]} and not {@code String}-valued elements. For example:
+   * {@link types.String} typed Tensors are multi-dimensional arrays of arbitrary byte sequences, so
+   * can be initialized from arrays of {@code byte[]} elements. For example:
    *
    * <pre>{@code
-   * // Valid: A DataType.STRING tensor.
-   * Tensor s = Tensor.create(new byte[]{1, 2, 3});
+   * // Valid: A String tensor.
+   * Tensor<String> s = Tensor.create(new byte[]{1, 2, 3}, String.class);
```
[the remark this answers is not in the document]
andrewcmyers: I see what you mean about potential confusion, but I think programmers will be more comfortable with using String than a new Bytes class. I think programmers will be mostly creating string Tensors using the Tensors class, which for reasons of both convenience and efficiency ought to allow creating tensors from (arrays of) Java Strings with a specified encoding.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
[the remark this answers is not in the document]
andrewcmyers: Actually, the check on 143 is superfluous now. Thanks for catching that; I've removed it. The call to expect() adds an unnecessary run-time check either way.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -28,89 +28,123 @@
...
-    t.dtype = dtype;
+  private static Tensor<?> create(Object obj, DataType dtype) {
+    @SuppressWarnings("rawtypes")
+    Tensor<?> t = new Tensor(dtype);
     t.shapeCopy = new long[numDimensions(obj, dtype)];
-    assert objectCompatWithType(obj, dtype);
+    if (!objectCompatWithType(obj, dtype)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Expected an object compatible with %s, got %s instead", dtype, dataTypeOf(obj)));
```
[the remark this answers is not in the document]
andrewcmyers: This was a gratuitous run-time check anyway, removed.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: This was something I added in order to support initializing scalar UInt8 tensors from scalar byte values. But it's not clear we want to support that. Maybe should go into a separate PR?

andrewcmyers: I guess a scalar byte value can still initialize a single-byte string, so I've restored the `ret > 0`

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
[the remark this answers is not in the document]
andrewcmyers: Also, calling it a Tensor<String> is actually more compatible with the way tensors are presented in the other language APIs.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -244,9 +237,10 @@
    * @throws IllegalArgumentException If the tensor datatype or shape is not compatible with the
    *     buffer
    */
-  @SuppressWarnings("unchecked")
   public static <T> Tensor<T> create(Class<T> type, long[] shape, ByteBuffer data) {
-    return (Tensor<T>) create(DataType.fromClass(type), shape, data);
+    @SuppressWarnings("unchecked")
```
andrewcmyers: Local variable ret must be introduced to allow @SuppressWarnings to be moved inside the method defn.

andrewcmyers:
Ready for review.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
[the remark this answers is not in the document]
andrewcmyers: I agree that it's "really" a bunch of bytes, but programmers will think of it as containing strings. The fact that there is an encoding to and from bytes will be largely a detail, since people are mostly not going to write code manipulating these strings at the byte level anyway. I would even suggest picking a default encoding -- UTF-8 seems to have become
 dominant -- and allowing alternate encodings to be provided.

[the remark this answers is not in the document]
asimshankar: Many of the TensorFlow operations that accept TF_STRING tensors truly expect a sequence of bytes, such as an encoded protocol buffer.  A popular one is [`tf.parse_example`](https://www.tensorflow.org/api_docs/python/tf/parse_example). The serialized protocol buffers cannot be represented as a Java string.

So I guess I'm disputing the claim that "mostly not going to write code manipulating these strings at the byte level anyway"  :).  

Hence, I'm tempted to introduce a `Bytes` type. If later, we decide to add a `String` which implies UTF-8 encoding to `Bytes` as a convenience, maybe that will be okay. But I feel we should be true to what the type really represents?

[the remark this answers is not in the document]
andrewcmyers: I guess I'm looking at the Python API code and seeing that it's really nice to be able to write things like `tf.constant("filename.txt")` rather than some awkward thing that explicitly encodes the string into an array of bytes. The Python programmer gets to treat Python strings as being the things in tensors, so it seems like a shame to give the Java programmer something clunkier and more foreign when Java strings are arguably more capable than Python strings.

Not to pick nits, but it seems to me that serialized protocol buffers actually can be represented as a Java string, precisely because Java characters are bigger than 8 bits. If encoded or decoded using ISO-8859-1, there will be a one-to-one correspondence between the Java characters and the bytes in the serialized protobuf. Yes, Java strings can have characters that can't be represented as bytes, but I guess I'm not seeing the harm that comes from nevertheless saying these are tensors of String.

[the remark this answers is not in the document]
andrewcmyers: Of course, there is no real requirement that the type parameter actually correspond to what is stored in the tensor. But the thrust of the recent set of changes was to move away from inventing new types like TFString/Bytes and try to use existing, familiar Java types. I thought you had a good idea there!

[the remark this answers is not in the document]
asimshankar: Taking a step back, I think we agree on wanting to make it convenient to convert Java `String`s into `Tensor`s. Mostly via methods in the (yet to come) `Tensors` class with factory functions.

The point of debate here is what these functions should return, i.e.:

```java
public class Tensors {
  public static Tensor<String> create(String value);  // UTF-8 encoded
  public static Tensor<String> create(String value, Charset charset);
  public static Tensor<String> create(byte[] value);
}
```

*or*

```java
public class Tensors {
  public static Tensor<Bytes> create(String value);  // UTF-8 encoded
  public static Tensor<Bytes> create(String value, Charset charset);
  public static Tensor<Bytes> create(byte[] value);
}
```

My contention was that the latter is preferable. It makes it clear that the element type of the tensor is a sequence of bytes, not a `String`. I feel that this is more accurate and if the developer blinks, it's because this is a distinction worth knowing about.  For example, [`tf.equal`](https://www.tensorflow.org/api_docs/python/tf/equal) on two scalar `Tensor<String>` objects will return `false` even if they are the same "strings" because the encoding use to create the `Tensor<String>` might have been different.

Point taken about the serialized protocol buffers and ISO-8859-1 encoding. But it seems to be more natural to represent the serialized buffer as a `byte[]` rather than a `String`, which is why presumably the [protocol buffer APIs use them](https://developers.google.com/protocol-buffers/docs/javatutorial#parsing-and-serialization).

You example on `tf.constant("filename.txt")` remains equally convenient in both, no? It's just whether the returned value is a `Tensor<Bytes>` or a `Tensor<String>`.

Thoughts? :)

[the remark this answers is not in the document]
andrewcmyers: You're absolutely right, we can offer exactly the same API whether we call it `Tensor<String>` or `Tensor<Bytes>`. If we go the second route, I would prefer to call it `Tensor<TFString>` rather than `Tensor<Bytes>` because 1) that makes the correspondence to the other language APIs (e.g., `tf.string`) more direct and obvious, 2) it communicates that this is the way you solve the common problem of putting strings into tensors, and 3) still adds the developer eye-blink you're after.

Personally, I think that people will, in the odd case that it matters, still make the same silly encoding/decoding mistakes regardless of what we call the type parameter, so I feel the familiarity of `String` is more valuable than the warning conveyed by the unexpected names `Bytes`/`TFString`.

tensorflow/java/src/test/java/org/tensorflow/op/ScopeTest.java
```diff
@@ -148,122 +151,166 @@ public void composite() {
...
   // "handwritten" sample operator classes
-  private static final class Const {
-    private final Output output;
+  private static final class Const<T> {
+    private final Output<T> output;
 
-    static Const create(Scope s, Object v) {
-      try (Tensor value = Tensor.create(v)) {
-        return new Const(
+    @SuppressWarnings("unchecked")
```
asimshankar: Avoid the `@SuppressWarnings` by making this (and the next function) be:

```java
static Const<Integer> create(Scope s, int v) {
  return create(s, Tensor.create(Integer.class, v));
}
```

?

(Basically, I want to avoid `@SuppressWarnings` from any test/example code, since those often act as inspiration to programmers)

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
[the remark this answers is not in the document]
asimshankar: Let's agree to disagree here (regarding familiarity with `String`) :). My fear is that developers will _think_ they're comfortable with `String` and 95% of the time it will work out fine, but those 5% where it will differ from the other types (e.g., `Integer.equals()` and `tf.equals(Tensor<Integer>, Tensor<Integer>)` will be consistent, while `String.equals()` and `tf.equals(Tensor<String>, Tensor<String>)` will not) will be annoying.

That said, other than this `String` business we do agree and there is much to this PR and I'd like to get this in before 1.4 is cut, so let's proceed with the PR as is and we can consider changing `String` to `Bytes` or `TFString` in a follow up (I'll try to find some time to gather more opinions).

asimshankar (approved):

asimshankar:
Jenkins, test this please

tensorflow/java/src/test/java/org/tensorflow/op/ScopeTest.java
andrewcmyers: Oops, I missed that one. Thanks

asimshankar:
Jenkins, test thsi please

asimshankar (approved):

asimshankar:
Thanks @andrewcmyers !

asimshankar:
Jenkins, test this please

PR #11551, opened 2017-07-17
Fixes ArgMax test deprecation warnings

jwlawson:
The recent commit cbe1ef0 marks the `dimension` argument of argmax and argmin as deprecated, which are used in the corresponding tests and so this brings up warnings when the tests are run. The tests should be using the `axis` argument instead.


jhseu (approved):

jhseu:
Jenkins, test this please

PR #11556, opened 2017-07-17
Correct learning rate in code snippet.

ZeroCool2u:
Correct the learning rate in code snippet, such that it reflects the expected outcome, ~92%, instead of ~90%.

If this discrepancy is intentional, feel free to close this request. It's such a small fix I figured I'd submit the request just in case the change was accidental. 

I would note that the change is fixed in the master brach, however the problem is currently reflected in the production docs, which is confusing to beginners, like myself.


ZeroCool2u:
@googlebot I signed it!

jhseu:
Thanks for the contribution! We're not currently updating 1.2 anymore.

ZeroCool2u:
@jhseu Awesome, thanks for the heads up!

PR #11558, opened 2017-07-17
Add 'axis' option for 'tf.boolean_mask()'

### LINKED ISSUE #9721: Feature request: Add 'axis' option for 'tf.boolean_mask()'
`tf.boolean_mask()` is extremely useful when it comes to training.

However, the present `tf.boolean_mask()` does not have an `axis` option, thus masking some specific dimension of an array could be troublesome. 

Appreciate your consideration.

yongtang:
This fix tries to address #9721 where it was not possible to pass an
'axis' option for 'tf.boolean_mask()'.

This fix fixes #9721.

### DIFF
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -1113,12 +1113,17 @@ def boolean_mask(tensor, mask, name="boolean_mask", axis=None):
   the first K dimensions of `tensor`'s shape.  We then have:
     `boolean_mask(tensor, mask)[i, j1,...,jd] = tensor[i1,...,iK,j1,...,jd]`
   where `(i1,...,iK)` is the ith `True` entry of `mask` (row-major order).
+  The `axis` could be used with `mask` to indicate the axis to mask from.
+  In that case, `axis + dim(mask) <= dim(tensor)` and `mask`'s shape must match
+  the first `axis + dim(mask)` dimensions of `tensor`'s shape.
 
   Args:
     tensor:  N-D tensor.
     mask:  K-D boolean tensor, K <= N and K must be known statically.
     name:  A name for this operation (optional).
-    axis: The axis in `tensor` to mask from. Defaults to the first dimension.
+    axis:  A 0-D int Tensor representing the axis in `tensor` to mask from.
+      By default, axis is 0 which will mask from the first dimension. Otherwise
+      K + axis <= N.
 
   Returns:
     (N-K+1)-dimensional tensor populated by entries in `tensor` corresponding
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -1099,7 +1099,7 @@ def concat(values, axis, name="concat"):
   return gen_array_ops._concat_v2(values=values, axis=axis, name=name)
 
 
-def boolean_mask(tensor, mask, axis=None, name="boolean_mask"):
+def boolean_mask(tensor, mask, name="boolean_mask", axis=None):
   """Apply boolean mask to tensor.  Numpy equivalent is `tensor[mask]`.
 
   ```python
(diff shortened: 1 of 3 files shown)


tensorflow/python/ops/array_ops.py
```diff
@@ -1084,6 +1084,7 @@ def boolean_mask(tensor, mask, name="boolean_mask"):
   Args:
     tensor:  N-D tensor.
     mask:  K-D boolean tensor, K <= N and K must be known statically.
+    axis: The axis in `tensor` to mask from. Defaults to the first dimension.
```
josh11b: What if mask isn't a 1-D tensor?

tensorflow/python/ops/array_ops.py
```diff
@@ -1066,7 +1066,7 @@ def concat(values, axis, name="concat"):
                                   name=name)
 
 
-def boolean_mask(tensor, mask, name="boolean_mask"):
+def boolean_mask(tensor, mask, axis=None, name="boolean_mask"):
```
josh11b: For backwards compatibility, all new arguments must be last (in addition to having defaults).

josh11b (commented):
(API review.)

tensorflow/python/ops/array_ops.py
yongtang: Thanks @josh11b. The PR has been updated with the suggested changes.

tensorflow/python/ops/array_ops.py
yongtang: @josh11b If mask isn't 1-D then it should work as well. I added an additional test case for this.

yongtang:
@josh11b Thanks for the review. The PR has been updated. Please take a look.

tensorflow/python/ops/array_ops.py
```diff
@@ -1085,6 +1085,7 @@ def boolean_mask(tensor, mask, name="boolean_mask"):
     tensor:  N-D tensor.
     mask:  K-D boolean tensor, K <= N and K must be known statically.
     name:  A name for this operation (optional).
+    axis: The axis in `tensor` to mask from. Defaults to the first dimension.
```
josh11b: You should clarify the documentation. In particular you should mention that it must be an integer, and please update lines 1080-1082. Ideally in addition `axis` would also be allowed to be a list of length K to be consistent with other `axis` arguments in the API, e.g. https://www.tensorflow.org/api_docs/python/tf/reduce_sum .

josh11b (dismissed):
(for API review)

tensorflow/python/ops/array_ops.py
langmore: `boolean_mask` already can work as an n-dimensional mask.  Currently if `mask.shape.ndims = K`, it masks over the first `K` dimensions of `tensor`.  So the docstring should be changed to reflect the new behavior if e.g. `axis != None` and `mask` is multi-dimensional.

langmore: `boolean_mask(tensor, mask)` is supposed to act like numpy's `my_array[my_mask]`, which it currently does.  What you're proposing here looks like [`numpy.compress`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.compress.html#numpy.compress)

So it seems like this functionality should be part of an `array_ops.compress`.

josh11b: So numpy.compress only allows 1-D boolean masks, which makes the choice of only using an integer for the axis straightforward.  This op allows K-D boolean masks, which raises the question of whether the axis should be (allowed to be) a length-K vector.

jhseu:
@yongtang Any updates? Thanks!

yongtang:
@josh11b @langmore Sorry for missing the GitHub notification and late reply. I have rebased and updated PR with the docstrings addressed based on the comments.

drpngx:
@langmore @josh11b any additional thoughts?

Jenkins, test this please.

sb2nov:
@langmore @josh11b any additional thoughts?

josh11b:
This is good for api review.

yongtang:
Thanks @josh11b for the review. The API goldens have been updated so that `//tensorflow/tools/api/tests:api_compatibility_test` could be passed.

drpngx:
Jenkins, test this please.

martinwicke (approved):

martinwicke:
Ignoring Kokoro -- looks like tool failure to me.

PR #11559, opened 2017-07-17
Add the Constant operator class

kbsriram:
Create a custom operator class to create constants in the Graph,
and introduce the Operator marker annotation to identify
operator classes.

Please see #7149 for the master tracking issue.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
@@ -1,0 +1,112 @@
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+package org.tensorflow.op.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used by classes to make TensorFlow operations conveniently accessible via {@code
+ * org.tensorflow.op.Ops}.
+ *
+ * <p>An annotation processor (TODO: not yet implemented) builds the {@code Ops} class by
+ * aggregating all classes annotated as {@code @Operator}s. Each annotated class <b>must</b> have at
+ * least one public static factory method named {@code create} that accepts a {@link
+ * org.tensorflow.op.Scope} as its first argument. The processor then adds a convenience method in
+ * the {@code Ops} class. For example:
+ *
+ * <pre>{@code
+ * @Operator
+ * public final class MyOp implements Op {
+ *   public static MyOp create(Scope scope, Operand operand) {
+ *     ...
+ *   }
+ * }
+ * }</pre>
+ *
+ * <p>results in a method in the {@code Ops} class
+ *
+ * <pre>{@code
+ * import org.tensorflow.op.Ops;
+ * ...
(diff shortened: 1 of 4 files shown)


tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
```diff
@@ -0,0 +1,98 @@
...
+   * }</pre>
+   *
+   * <p>results in the {@code add} method placed within a {@code math} group within the {@code Ops}
+   * API.
+   *
+   * <pre>{@code
+   * ops.math().add(...);
+   * }</pre>
+   */
+  String group() default "";
```
asimshankar: Couple of questions:

1. Do we want to support multiple groups (e.g. `math().tanh()` and `nn.tanh()`?)
2. Do we want to support nested namespaces, or just one level below `ops`?

tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
```diff
@@ -0,0 +1,98 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used by classes that should be added to the Ops convenience API.
+ *
+ * <p>An annotation processor builds the {@code Ops} API by aggregating all classes annotated as
```
asimshankar: The annotation processor doesn't exist yet, right? It will come in a future PR? (Or am I missing something? :)

tensorflow/java/src/main/java/org/tensorflow/op/core/Constant.java
```diff
@@ -0,0 +1,177 @@
...
+      return createWithTensor(scope, value);
+    }
+  }
+
+  private static Constant createWithTensor(Scope scope, Tensor value) {
+    return new Constant(
+        scope
+            .graph()
+            .opBuilder(OP, scope.makeOpName(OP))
+            .setAttr(VALUE, value)
```
asimshankar: Since this is the one and only place where the string constants are used, might as well inline them

asimshankar (commented):
Apologies for the delay. Some clarification questions, but otherwise looks good. Thanks!

tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
kbsriram: Yes, in a future PR :-) I couldn't come up with a good way of documenting the intent of the annotation without referring to a yet-to-come processor.  Suggestions? I'll insert a TODO in the javadoc to make it clear it doesn't exist yet.

tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
kbsriram: It seems like at least the generated wrapper classes and the few core custom ones would use single groups and one level. The second item seems potentially useful, but hadn't considered the first - that's an interesting idea.

In the interests of spacing out CLs - what do you think about beginning with supporting one group/one level, and adding support for these over time?

As an annotation definition, potentially we can change it incrementally (i.e. converting to an array to support multiple groups shouldn't change the syntax for existing single-group ops, though they would need to be recompiled. Supporting a naming convention like `@group("contrib.mnist")` might be the approach to provide nested namespaces, so that can work incrementally.)

asimshankar: Yup, only adding it later if and when we need it sounds good. Was just raising the question to think about (and also perhaps suggest that the `AnnotationProcessor` would want to ensure that there are no `.`s in the `String` returned by `group()` etc.)

tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
asimshankar: TODO for now suffices :)

tensorflow/java/src/main/java/org/tensorflow/op/annotation/Operator.java
kbsriram: Ack - will do. I've corrected the other nits, and added a note in the annotation Javadoc that the names must be valid Java identifiers.

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please

PR #11565, opened 2017-07-18
update:tf_inspect.py

Shuolongbj:
DeprecationWarning: inspect.getargspec() is deprecated, use
inspect.signature() or inspect.getfullargspec()
if d.decorator_argspec is not None), _inspect.getargspec(target))


jhseu:
Thanks for the contribution! We don't use inspect.signature because it isn't available in Python 2.

Issue #10399, opened 2017-06-02
export step of Experiment fails in python 3.5 (likely due to string-type mismatch)


yufengg:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yep
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OSX
- **TensorFlow installed from (source or binary)**: binary 1.1.0
- **TensorFlow version (use command below)**: 1.1.0
- **Bazel version (if compiling from source)**: n/a
- **CUDA/cuDNN version**: n/a
- **GPU model and memory**: n/a
- **Exact command to reproduce**:
git clone https://github.com/amygdala/tensorflow-workshop
python tensorflow-workshop/workshop_sections/wide_n_deep/widendeep/model.py

### Describe the problem

When running Experiment stuff, it works fine in 2.7 but breaks in 3.5. From what I can tell, it's happening in gc.py, which is being called _after_ the model has been exported and everything is done. 

My best guess from looking at similar bugs on SO is that there's a mismatch somewhere between 'byte' strings and python strings, possibly because the base folder path of the export is generated (e.g. /exports/Servo/123456789/saved_model.py), so the somehow the concat isn't happy with this.

### Source code / logs
Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached. Try to provide a reproducible test case that is the bare minimum necessary to generate the problem.

File "/Users/yufengg/.pyenv/versions/anaconda3-4.1.1/lib/python3.5/site-packages/tensorflow/contrib/learn/python/learn/utils/gc.py", line 202, in get_paths
    p = parser(Path(os.path.join(base_dir, r), None))
  File "/Users/yufengg/.pyenv/versions/anaconda3-4.1.1/lib/python3.5/posixpath.py", line 89, in join
    genericpath._check_arg_types('join', a, *p)
  File "/Users/yufengg/.pyenv/versions/anaconda3-4.1.1/lib/python3.5/genericpath.py", line 145, in _check_arg_types
    raise TypeError("Can't mix strings and bytes in path components") from None
TypeError: Can't mix strings and bytes in path components

asimshankar:
Thanks for the report @yufengg . Can you include the full stracktrace and ideally a reproducible test case?

martin-gorner:
Same problem here. All you need in your code is to include an export strategy and the checkpoints garbage collector will raise an exception:

[...]

PR #9719, opened 2017-05-06
Fix mixed use of bytes and unicode string in gc.py

raviqqe:

PR #11593, opened 2017-07-18
Fix issue #10399

ssaavedra:
Because of the bytes compatibility this function is returning bytes
when it should be returning strings.

Fixes issue "TypeError: Can't mix strings and bytes in path
components" when using `tf.contrib.learn.learn_runner` with
experiments with an `export_strategy`.

See issue #10399 for a code example reference.

I didn't find obvious mistakes, but this is my first contribution and
I may have missed something.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/experiment.py
+++ b/tensorflow/contrib/learn/python/learn/experiment.py
@@ -594,8 +594,8 @@ class Experiment(object):
   def _maybe_export(self, eval_result, checkpoint_path=None):
     """Export the Estimator using export_fn, if defined."""
     export_dir_base = os.path.join(
-        compat.as_bytes(self._estimator.model_dir),
-        compat.as_bytes("export"))
+        self._estimator.model_dir,
+        "export")
 
     export_results = []
     for strategy in self._export_strategies:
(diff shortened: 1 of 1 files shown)


xiejw:
Sorry, I do not understand this fix. Do you have any particular error case you want to fix?

as_bytes might not be the best choice here. But dropping it seems very bad, as it will trigger the mix strings and bytes error. 

Please reopen if I missed something here.

ssaavedra:
On Python 3, the referred bug happens due to the modfiied function returning paths as bytes, while other functions, namely list_directory in [file_io.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/lib/io/file_io.py) expect and deal with strings. If you are not comfortable dropping as_bytes, as per the poor type safety that Python gives us, I can resubmit the patch changing the conversion to `compat.as_str_any` instead, if that seems more reasonable. The rationale behind bug #10399 is that a `os..path.join` is happening between a something returned by this function (bytes) and something returned (down the stacktrace) by list_directory in file_io.py.

On Python 2, I haven't seen any regression and all tests in //tensorflow/python/... seem to be passing on my computer.

xiejw:
No, not exactly. Unless I missed something, the underlying issue is all args passed to os.path.join should be same. It can be both bytes or strs, but not mix. 

You fix simply dropped the as_bytes for both components in this file, which would make the mix happen  in this file not the gc.py (as the log shown in #10399). And #10399 seems be fixed by  https://github.com/tensorflow/tensorflow/pull/9719/files already. 

I am going to ping #10399 to see whether this got fixed first.

PR #11606, opened 2017-07-19
replace supervised session to monitored session

horance-liu:
The author should forget to modify doc for `MonitoredSession._is_closed` when refactoring supervisor to monitored session.

### DIFF
--- a/tensorflow/python/training/monitored_session.py
+++ b/tensorflow/python/training/monitored_session.py
@@ -575,7 +575,7 @@ class _MonitoredSession(object):
           ops.get_default_graph()._unsafe_unfinalize()  # pylint: disable=protected-access
 
   def _is_closed(self):
-    """Return True if the supervised session is closed.  For tests only.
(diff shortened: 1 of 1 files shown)


tensorflow/python/training/monitored_session.py
```diff
@@ -575,7 +575,7 @@ def _close_internal(self, exception_type=None):
           ops.get_default_graph()._unsafe_unfinalize()  # pylint: disable=protected-access
 
   def _is_closed(self):
-    """Return True if the supervised session is closed.  For tests only.
+    """Return True if the monitored session is closed.  For tests only.
```
vrv: I think the existing documentation is correct: it's the underlying supervised session in _coordinated_creator.tf_sess which is being checked, not self.

vrv: That being said, it's not clear the documentation would be improved by keeping it the same, so i'll just merge this.

PR #11622, opened 2017-07-19
Enable passing scope to variable lookups

darrengarvey:
NB: Changes the public API, albeit in a backwards-compatible way.

Especially when copying variables between scopes it's handy to be able to grab sets of variables by scope. This allows the much more succinct:

``tf.trainable_variables("myscope")``

which is equivalent to:

``tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES, "myscope")``

- Also fixup a minor unrelated documentation bug.

### DIFF
--- a/tensorflow/python/kernel_tests/variables_test.py
+++ b/tensorflow/python/kernel_tests/variables_test.py
@@ -290,6 +290,21 @@ class VariablesTestCase(test.TestCase):
                        variables.global_variables())
       self.assertEqual([var_x, var_z, var_t], variables.trainable_variables())
 
+  def testCollectionsWithScope(self):
+    with self.test_session():
+        with ops.name_scope("scope_1"):
+          var_x = variables.Variable(2.0)
+        with ops.name_scope("scope_2"):
+          var_y = variables.Variable(2.0)
+
+        self.assertEqual([var_x, var_y], variables.global_variables())
+        self.assertEqual([var_x], variables.global_variables("scope_1"))
+        self.assertEqual([var_y], variables.global_variables("scope_2"))
+
+        self.assertEqual([var_x, var_y], variables.trainable_variables())
+        self.assertEqual([var_x], variables.trainable_variables("scope_1"))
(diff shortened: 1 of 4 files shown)


jhseu:
Josh, thoughts on this API change?

josh11b:
Alex, you are more familiar with this stuff than I am.  What do you think?

alextp:
You can get similar functionality now with "scope.global_variables()": https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/variable_scope.py#L907 (there is scope.trainable_variables as well) on the variable scope objects.

That said I'm not against this change, as it is backwards-compatible

darrengarvey:
Thanks for the reviews. I've added documentation for the new args too, as well as including `_all_saveable_objects()` to the list as that will be used similarly once it's public.

alextp:
Jenkins, test this please.

alextp (approved):
Approving that I think this is an improvement to the API; will let @josh11b flip the tag on whether to merge this as API owner

darrengarvey:
Gah, I've missed something with pylint. I'll check that when I'm home and update.

darrengarvey:
Fixed the indentation.

josh11b:
Approval for API review

alextp:
Jenkins, test this please.

darrengarvey:
The Windows cmake test failed because of a checkout issue. The Linux CPU test failed the api compatibility test. That passes here with (albeit with this PR on top of master), with:

    bazel test --verbose_failures --config=cuda --config=opt //tensorflow/tools/api/tests:api_compatibility_test

I'll see if it was broken on master when I made the PR, but I'm away for a few days now.

alextp:
Jenkins, test this please.

vrv:
One needs to update the API golden files for the test to pass.  See the logs and search for api_compatiblity_test to see how to update the golden files.

darrengarvey:
I was building with python 3 so the api checks looked like they passed, but were being skipped. Should have spotted that.

Updated.

alextp:
Jenkins, test this please.

PR #11595, opened 2017-07-18
[XLA] Fix plugin example buffer address for tuples

yxlao:
Currently the returned `se::DeviceMemoryBase(buf, size)` uses the incremented `buf`, which may cause deallocation error.

PR #11625, opened 2017-07-19
[XLA] Fix plugin example buffer address for tuples

yxlao:
Currently the returned `se::DeviceMemoryBase(buf, size)` uses the incremented `buf`, which may cause deallocation error.

(Resubmit of #11595 after CLA fix)


jhseu (approved):

jhseu:
Jenkins, test this please

Issue #11228, opened 2017-07-02
GPU kernel for segment_sum?


yycho0108:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**: Binary
- **TensorFlow version (use command below)**: ('v1.2.0-5-g435cdfc', '1.2.1')
- **Python version**: 2.7.12
- **CUDA/cuDNN version**: 6.0
- **GPU model and memory**: GTK1070/ 8105MB

^^ I don't think system information is hugely relevant in this case, but writing them down anyways.

### Describe the problem
Describe the problem clearly here. Be sure to convey here why it's a bug in TensorFlow or a feature request.

This is a feature request. I was running *tf.unsorted_segment_sum* on an unsorted list of segments, and thought that I might be able to get a performance gain if the segment ids are sorted (as that is usually the case).
However, to my surprise, I discovered that there are no supported GPU kernels for vanilla segment_sum, forcing tensorflow to copy memory over to the CPU and thereby slowing down the operation greatly.
Would it be possible to support an optimized GPU version of segment_sum that takes advantage of the fact that the segments are sorted?

### Source code / logs

(Device Placement)
UnsortedSegmentSum: (UnsortedSegmentSum)/job:localhost/replica:0/task:0/gpu:0
softmax/ops/SegmentSum: (SegmentSum)/job:localhost/replica:0/task:0/cpu:0

(Profiling)
[Unsorted Segment Sum] : Took 1.979 Seconds
[Segment Sum] : Took 2.704 Seconds

tjingrant:
I'm working on this one.

nolanliou:
@tjingrant How is the PR going?

tjingrant:
Hi, I got some feedback about a week ago but since then was caught up with some other things. I will likely be able to submit an update within 2 days. Sorry for the delay.

yongtang:
It looks like this issue has been fixed in PR #11630. I will close this one. Thanks for the contribution!

PR #11630, opened 2017-07-20
Add GPU implementation for tf.segment_sum.

tjingrant:
As per #11228, a GPU version of segment sum has been created. 
## Performance 
On a Tesla K40c GPU, the following performance tests have been performed. 
- The test case sizes are represented using 3-tuples of integers. The first two integers correspond to the outer and inner dimension of the input data and the last integer corresponds to the outer dimension of the output data, which is also the total number of segments. 
- The first three rows of data are obtained by running reductions on float32 input type whereas the last three rows of data are obtained by running reductions on float64 input type. 
- We use the GPU version of unsorted segment reduction as the baseline. 
- During experiments, they share the same input data and outputs are compared for consistency.

|  | (1024, 1024, 128)            |   (2048, 2048, 256)          |    (4096, 4096, 512)         | 
|-------------------------------------------------------------|-------------|-------------|-------------| 
| t_unsorted                                                  | 84.942us    | 332.12us    | 1.3170ms    | 
| t_sorted                                                    | 71.047us    | 264.16us    | 1.0391ms    | 
| speedup                                                     | 1.20  | 1.26 | 1.27  | 
|                                                             |             |             |             | 
| t_unsorted                                                  | 160.69us    | 594.93us    | 2.2895ms    | 
| t_sorted                                                    | 106.94us    | 395.19us    | 1.5662ms    | 
| speedup                                                     | 1.50 | 1.51 | 1.46 | 

## Known Limitations:
- We do not check segment_id to make sure that they are within [0, num_outputs-1].
- We do not verify that segment_id is increasing.

### DIFF
--- a/tensorflow/core/kernels/segment_reduction_ops.cc
+++ b/tensorflow/core/kernels/segment_reduction_ops.cc
@@ -35,6 +35,15 @@ limitations under the License.
 #include "tensorflow/core/platform/logging.h"
 #include "tensorflow/core/util/util.h"
 
+
+#if GOOGLE_CUDA
+#include "tensorflow/core/common_runtime/gpu/gpu_event_mgr.h"
+#include "tensorflow/core/kernels/cuda_solvers.h"
+#include "tensorflow/core/platform/cuda.h"
+
+using ::perftools::gputools::cuda::ScopedActivateExecutorContext;
+#endif  // GOOGLE_CUDA
+
 namespace tensorflow {
 
 typedef Eigen::ThreadPoolDevice CPUDevice;
@@ -189,11 +198,12 @@ class SegmentReductionOp : public OpKernel {
 #ifdef GOOGLE_CUDA
 //  SegmentSumGPUOp is a segment sum operator implemented for GPU only.
 template <class T, class Index>
-class SegmentSumGPUOp : public OpKernel {
+class SegmentSumGPUOp : public AsyncOpKernel {
  public:
-  explicit SegmentSumGPUOp(OpKernelConstruction* context) : OpKernel(context) {}
+  explicit SegmentSumGPUOp(OpKernelConstruction* context)
+      : AsyncOpKernel(context) {}
 
-  void Compute(OpKernelContext* context) override {
+  void ComputeAsync(OpKernelContext* context, DoneCallback done) override {
     const Tensor& input = context->input(0);
     const Tensor& segment_ids = context->input(1);
 
@@ -202,35 +212,50 @@ class SegmentSumGPUOp : public OpKernel {
     }
 
     const int64 num_indices = segment_ids.NumElements();
+    perftools::gputools::DeviceMemoryBase output_rows_device(
+        (void*)(segment_ids.template flat<Index>().data() + (num_indices - 1)));
+    ScratchSpace<Index> output_rows_host(context, 1, /* on_host */ true);
 
-    Index output_rows;
-    context->eigen_device<GPUDevice>().memcpyDeviceToHost(
-        (void*)&output_rows,
-        (void*)(segment_ids.template flat<Index>().data() + (num_indices - 1)),
-        sizeof(Index));
-    output_rows++;
+    auto stream = context->op_device_context()->stream();
+    OP_REQUIRES_ASYNC(
+        context, stream
+                     ->ThenMemcpy(output_rows_host.mutable_data(),
+                                  output_rows_device, sizeof(Index))
+                     .ok(),
+        errors::Internal(
+            "SegmentSumGPUOp: failed to copy num_true from device"),
+        done);
 
-    OP_REQUIRES(context, output_rows > 0,
-                errors::InvalidArgument("segment ids must be >= 0"));
+    functor::SegmentSumFunctor<T, Index> functor_;
+    auto create_and_check_output = [context, output_rows_host, &input,
+                                    &segment_ids, &functor_, done]() {
+      // Ensure that within the callback, the proper GPU settings are
+      // configured.
+      auto stream = context->op_device_context()->stream();
+      ScopedActivateExecutorContext scoped_activation{stream->parent()};
 
-    TensorShape output_shape = input.shape();
-    output_shape.set_dim(0, output_rows);
+      Index output_rows = *output_rows_host.data();
+      output_rows++;
+      TensorShape output_shape = input.shape();
+      output_shape.set_dim(0, output_rows);
 
-    Tensor* output = nullptr;
-    OP_REQUIRES_OK(context, context->allocate_output(0, output_shape, &output));
+      Tensor* output = nullptr;
+      OP_REQUIRES_OK(context,
+                     context->allocate_output(0, output_shape, &output));
 
-    auto output_flat = output->flat_outer_dims<T>();
-    auto data_ptr = input.template flat<T>().data();
-    auto segment_flat = segment_ids.flat<Index>();
+      auto output_flat = output->flat_outer_dims<T>();
+      auto data_ptr = input.template flat<T>().data();
+      auto segment_flat = segment_ids.flat<Index>();
+      functor_(context, context->eigen_device<GPUDevice>(), output_rows,
+               segment_ids.shape(), segment_flat, input.NumElements(), data_ptr,
+               output_flat);
 
-    functor_(context, context->eigen_device<GPUDevice>(), output_rows,
-             segment_ids.shape(), segment_flat, input.NumElements(), data_ptr,
-             output_flat);
+      done();
+    };
 
+    context->device()->tensorflow_gpu_device_info()->event_mgr->ThenExecute(
+        stream, create_and_check_output);
   }
-
- private:
-  functor::SegmentSumFunctor<T, Index> functor_;
 };
 #endif  // GOOGLE_CUDA
 
--- a/tensorflow/core/kernels/segment_reduction_ops.cc
+++ b/tensorflow/core/kernels/segment_reduction_ops.cc
@@ -241,7 +241,7 @@ class SegmentSumGPUOp : public AsyncOpKernel {
                                   output_rows_device, sizeof(Index))
                      .ok(),
         errors::Internal(
-            "SegmentSumGPUOp: failed to copy num_true from device"),
+            "SegmentSumGPUOp: failed to copy output_rows from device"),
         done);
 
     functor::SegmentSumFunctor<T, Index> functor_;
@@ -254,6 +254,10 @@ class SegmentSumGPUOp : public AsyncOpKernel {
 
       Index output_rows = *output_rows_host.data();
       output_rows++;
+      OP_REQUIRES_ASYNC(context, output_rows > 0,
+                        errors::InvalidArgument("segment ids must be >= 0"),
+                        done);
+
       TensorShape output_shape = input.shape();
       output_shape.set_dim(0, output_rows);
 
--- a/tensorflow/core/kernels/segment_reduction_ops.cc
+++ b/tensorflow/core/kernels/segment_reduction_ops.cc
@@ -197,6 +197,19 @@ class SegmentReductionOp : public OpKernel {
 
 #ifdef GOOGLE_CUDA
 //  SegmentSumGPUOp is a segment sum operator implemented for GPU only.
+//  TODO: This implementation of SegmentSumGPUOp is sometimes slower than
+//  its unsorted counterpart (mostly when problem size is small).
+//  This is due to the following two main reasons and a cost-effective way
+//  to resolve these problems is desirable.
+//  1. Sorted segment sum requires a memory transfer from device to host in
+//     order to know the size of the output dimension whereas unsorted segment
+//     sum receives the size of the output dimension as an input parameter.
+//  2. Sorted segment sum is essentially a tiled version of unsorted segment
+//     sum and therefore such optimization comes at an inherent cost. However
+//     such cost may not be justified when the problem size is small. When to
+//     use the tiled version or the untiled version depends on many factors
+//     including data alignments, ratio of calculation to memory traffic and
+//     obviously, the problem sizes.
 template <class T, class Index>
 class SegmentSumGPUOp : public AsyncOpKernel {
  public:
--- a/tensorflow/core/kernels/segment_reduction_ops.cc
+++ b/tensorflow/core/kernels/segment_reduction_ops.cc
@@ -16,6 +16,9 @@ limitations under the License.
 // See docs in ../ops/math_ops.cc.
 
 #define EIGEN_USE_THREADS
+#if GOOGLE_CUDA
+#define EIGEN_USE_GPU
+#endif  // GOOGLE_CUDA
 
 #include "tensorflow/core/kernels/segment_reduction_ops.h"
 #include <vector>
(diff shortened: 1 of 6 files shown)


tjingrant:
I signed it!

tensorflow/core/kernels/segment_reduction_ops.cc
```diff
@@ -183,6 +186,54 @@ class SegmentReductionOp : public OpKernel {
...
+    const Tensor& segment_ids = context->input(1);
+
+    if (!SegmentReductionDoValidation(context, input, segment_ids)) {
+      return;
+    }
+
+    const int64 num_indices = segment_ids.NumElements();
+
+    Index output_rows;
+    context->eigen_device<GPUDevice>().memcpyDeviceToHost(
```
zheng-xq: This introduces a device/host sync. TensorFlow tries very hard to avoid that. 

An example on how to address such issue: 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/where_op.cc#L253

Another common practice is to check the indices on GPU, and do nothing with invalid indicies.

tensorflow/core/kernels/segment_reduction_ops.cc
```diff
@@ -183,6 +186,54 @@ class SegmentReductionOp : public OpKernel {
   }
 };
 
+#ifdef GOOGLE_CUDA
+//  SegmentSumGPUOp is a segment sum operator implemented for GPU only.
+template <class T, class Index>
+class SegmentSumGPUOp : public OpKernel {
```
zheng-xq: The convention is to overload "SegmentReductionOp" for GPUDevice

tensorflow/core/kernels/segment_reduction_ops.cc
```diff
@@ -183,6 +186,54 @@ class SegmentReductionOp : public OpKernel {
...
+    output_shape.set_dim(0, output_rows);
+
+    Tensor* output = nullptr;
+    OP_REQUIRES_OK(context, context->allocate_output(0, output_shape, &output));
+
+    auto output_flat = output->flat_outer_dims<T>();
+    auto data_ptr = input.template flat<T>().data();
+    auto segment_flat = segment_ids.flat<Index>();
+
+    functor_(context, context->eigen_device<GPUDevice>(), output_rows,
```
zheng-xq: Since the functor is trivial, we don't have to cache it.

tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc
```diff
@@ -54,6 +54,74 @@ __device__ __forceinline__ void AccumulateInto(
...
+// to an output element because of the increasing nature of segment
+// ids. In such cases, we do not need to use atomic operations
+// to write results to global memory.
+// In the flattened view of input data (with only outer and inner
+// dimension), every thread processes a strip of input data of
+// size OUTER_DIM_TILE_SIZE x 1. This strip runs across multiple
+// rows of input data and all reduction elements share one inner
+// dimension index.
+#define OUTER_DIM_TILE_SIZE 8
+#define CEIL_DIV(x, y) (1 + (((x)-1) / (y)))
```
zheng-xq: This produces the wrong answer when x == 0. The typical way is to write (x + y - 1) / y. 

In TF, the convention is to use Eigen::divup

tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc
```diff
@@ -80,6 +148,44 @@ __global__ void UnsortedSegmentSumCustomKernel(
...
+  // Notes:
+  // *) 'input_total_size' is the total number of elements to process.
+  // *) 'segment_ids.shape' is a prefix of data's shape.
+  // *) 'input_outer_dim_size' is the total number of segments to process.
+  const Index input_total_size = data_size;
+  const Index input_outer_dim_size = segment_ids.dimension(0);
+  const Index input_inner_dim_size = input_total_size / input_outer_dim_size;
+
+  const Index input_outer_dim_num_strip =
+      1 + (input_outer_dim_size - 1) / OUTER_DIM_TILE_SIZE;
```
zheng-xq: Same here. Use div up

zheng-xq (commented):
First wave of comments. I'll take a closer look at the details later.

Overall, please add functional tests and benchmarks to make sure it is actually faster across a range of different shapes with different aspect ratios.

Also we need to add more data shapes and randomly generated data to make sure there is no hidden race conditions in the logic.

tensorflow/core/kernels/segment_reduction_ops.cc
tjingrant: Hi, the current implementation of SegmentReductionOps is actually rather specialized for non-GPU devices. I'm not sure if the usual different-functor-for-different-device approach is the best way to go here. I actually started this implementation by porting existing code to be reused on GPU but later gave up because it's very inefficient for GPU device; in addition, the original implementation takes data input as non-primitive data types (i.e., TensorMap) which means if I were to use GPU specialized functors, I may have to introduce some extra data types into the .cu file and risk contaminating the .cu file with unnecessary header files; again, with these considerations, I'm not sure overloading SegmentReductionOp is the best approach. Nevertheless, I'm more than happy to consult with your thoughts on this.

tensorflow/core/kernels/segment_reduction_ops.cc
tjingrant: Thanks! Will update that.

tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc
tjingrant: Thanks for picking up this mistake.

tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc
tjingrant: Will do.

tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc
```diff
@@ -54,6 +54,74 @@ __device__ __forceinline__ void AccumulateInto(
...
+      // decide whether to write result to global memory
+      if (current_output_segment_id > last_output_segment_id) {
+        const Index output_index =
+            last_output_segment_id * inner_dim_size + segment_offset;
+        // decide whether to write result to global memory using atomic
+        // operations
+        if (last_output_segment_id == first_segment_id) {
+          AccumulateInto<T>(output + output_index, sum);
+        } else {
+          *(output + output_index) = sum;
```
zheng-xq: I took a closer look at this code. I think we should definitely add more tests to make sure there is no hidden race condition. Also adding a benchmark showing there is no performance regression against unsorted_segment_sum is good. 

Please make sure try a few different tensor shape aspect ratio in the benchmarks.

tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc
```diff
@@ -54,6 +54,74 @@ __device__ __forceinline__ void AccumulateInto(
...
+
+  CUDA_1D_KERNEL_LOOP(strip_index, total_stripe_count) {
+    const Index segment_offset = strip_index % inner_dim_size;
+    const Index input_outer_dim_index_base =
+        strip_index / inner_dim_size * OUTER_DIM_TILE_SIZE;
+
+    T sum = T(0);
+    Index first_segment_id = segment_ids[input_outer_dim_index_base];
+    Index last_output_segment_id = output_outer_dim_size;
+    const Index actual_stripe_height = MIN(
```
zheng-xq: As for "MIN", tf convention is "min"

tensorflow/core/kernels/segment_reduction_ops_gpu.cu.cc
```diff
@@ -54,6 +54,74 @@ __device__ __forceinline__ void AccumulateInto(
...
+#define OUTER_DIM_TILE_SIZE 8
+#define CEIL_DIV(x, y) (1 + (((x)-1) / (y)))
+template <typename T, typename Index>
+__global__ void SortedSegmentSumCustomKernel(const Index input_outer_dim_size,
+                                             const Index inner_dim_size,
+                                             const Index output_outer_dim_size,
+                                             const Index* segment_ids,
+                                             const T* input, T* output) {
+  const Index input_outer_dim_num_strip =
+      CEIL_DIV(input_outer_dim_size, OUTER_DIM_TILE_SIZE);
```
zheng-xq: Integer division is quite expensive, maybe pass them in.

tjingrant:
I have updated the pertinent code. Performance tests with more aspect ratios are conducted and recorded in this [gist](https://gist.github.com/tjingrant/3113562d6e9fc8738487ca91847ffd16). The results of both segment_sum and unsorted_segment_sum are always compared for consistency and therefore this also serves as a correctness test for potential race conditions.

In general, the execution time of sorted segment sum cuda kernel is uniformly shorter than that of unsorted segment sum cuda kernel when the problem size is medium or large. It is clear that tiling comes with an inherent cost and therefore when the problem size is small, the performance penalty imposed by tiling will be more noticeable but hopefully nevertheless small. Of all the problem sizes tested (small and large problem sizes alike), the average speed up factor is 1.17 for fp32 and 1.38 for fp64. In general, for fp32 typed segment summation, the break even point is around 512x512 and for any problem sizes beyond this point, using sorted segment sum will be profitable. For fp64, this point is very low and we actually did not see executions of any problem sizes during which sorted segment sum is slower.

It is also worth noting that sorted segment sum requires stream synchronization in order to retrieve output size (a scalar) from device memory. Therefore the overall break even point is somewhat higher but since this cost is constant, it will not impact the execution time of kernels doing more significant work. Please let me know if there is any way to mitigate this problem. Specifically, 

- is it possible to access the input array on the host before it is moved to the device? 
- Or slightly worse, is it possible to have duplicate copies of array at both host and device memory?

rmlarsen:
@zheng-xq can you please take another look?

rmlarsen:

tjingrant:
Hi, judging from the console output, I'm not sure if the test failure is caused by my changes.

`//bazel_pip/tensorflow/contrib/timeseries/python/timeseries:state_management_test FAILED in 13.3s`

This does not look like anything I've touched.

rmlarsen:
@tjingrant indeed, that appears unrelated to your change.

rmlarsen:
@zheng-xq XQ, this appears to pass (except for known flake). Can you approve it?

zheng-xq:
LGTM

@tjingrant, I think it is a good idea to check in your benchmark so the changes are self-sufficient. But I will leave it up to you whether you want to make it into a follow-up CL.

tjingrant:
Updated with benchmarks included.

tjingrant:
ping @zheng-xq , @rmlarsen

nolanliou:
I do a quick test about your code with the [script](https://gist.github.com/nolanliou/a1dc1df3d7d6267a0177b6749ccdf696). the result show there is not as good as unsorted_segment_sum with larger scale of segment_ids.
```
unsorted_segment_sum: 5.616264s
sorted_segment_sum: 7.558630s
unique ids: 6287
```

tjingrant:
Hi @nolanliou , thanks for checking out my code! As for your observation, as you probably know, there are a lot of factors that could affect the performance and after all, your ratio of reductions performed to your total data footprint is quite low and therefore in this case I do not expect my code to perform well. I encourage you to explore more problem sizes and reduction to data footprint ratio. Apart from that, the **main reason** you are seeing such a dramatic performance discrepancy is that you are not using the unsorted segment reductions correctly and probably it's not producing the correct results. Try placing it on a CPU device and you will see the error messages.

tensorflow/python/kernel_tests/segment_reduction_ops_test.py
```diff
@@ -641,6 +646,56 @@ def testGradientSegmentsInvalid4(self):
         with self.assertRaisesOpError(r"Segment id 0 out of range \[0, 0\)"):
           s.eval()
 
+def type_to_str(t):
```
zheng-xq: Minor nitpicking: maybe np_type_to_str?

tensorflow/python/kernel_tests/segment_reduction_ops_test.py
```diff
@@ -641,6 +646,56 @@ def testGradientSegmentsInvalid4(self):
...
+  if t == np.float32:
+    return "fp32"
+  if t == np.float64:
+    return "fp64"
+
+class SegmentReductionOpBenchmark(test.Benchmark):
+
+  def benchmarkSegmentSumGPU(self):
```
zheng-xq: Check "is_gpu_available"?

tensorflow/python/kernel_tests/segment_reduction_ops_test.py
```diff
@@ -641,6 +646,56 @@ def testGradientSegmentsInvalid4(self):
...
+    for outer_dim, ratio, inner_dim, dtype in \
+      itertools.product(outer_dim_options,
+                        ratio_options, inner_dim_options, dtype_options):
+      output_outer_dim = int(outer_dim/ratio)
+
+      const = np.random.randint(5, size=(outer_dim, inner_dim))
+      seg_ids = np.sort(np.random.randint(
+          output_outer_dim, size=outer_dim))
+
+      op_functors = [lambda vc, vs:
```
zheng-xq: Most benchmarks would split them into two different sub-benchmarks. They can share the underlying implementation

tensorflow/python/kernel_tests/segment_reduction_ops_test.py
```diff
@@ -641,6 +646,56 @@ def testGradientSegmentsInvalid4(self):
...
+  if t == np.float32:
+    return "fp32"
+  if t == np.float64:
+    return "fp64"
+
+class SegmentReductionOpBenchmark(test.Benchmark):
+
+  def benchmarkSegmentSumGPU(self):
+    repeat = 10
+    outer_dim_options = [2**x for x in range(7, 14, 2)]
```
zheng-xq: Please check large size with less alignments as well.

tensorflow/python/kernel_tests/segment_reduction_ops_test.py
```diff
@@ -641,6 +646,56 @@ def testGradientSegmentsInvalid4(self):
...
+                                      name="_".join(map(str,
+                                                        [name,
+                                                         outer_dim,
+                                                         ratio,
+                                                         inner_dim,
+                                                         type_to_str(dtype)])))
+            t.append(r["wall_time"])
+
+      # print out the speed up factor for each test
+      print("{:d}\t{:d}\t{:d}\t{:f}".format(
```
zheng-xq: We really care about the speedup before and after your CL. Please include the comparison before and after in your commit description. Thanks!

zheng-xq (changes requested):

tensorflow/python/kernel_tests/segment_reduction_ops_test.py
tjingrant: Hi @zheng-xq, there's no (GPU) implementation before my commit. This commit enables segment_sum on GPU. The speedup to the base-line here is another thing. Can you double-check this (the speed up against baseline, not impl before my commit) is what you want?

zheng-xq: Fair point. In that case, the comparison report in this issue is good enough. If you can include in the commit description, that is helpful. The benchmarks should be separate though.

tjingrant:
Hi @zheng-xq , I think I resolved all your requested changes.

tensorflow/python/kernel_tests/segment_reduction_ops_test.py
```diff
@@ -646,56 +646,59 @@ def testGradientSegmentsInvalid4(self):
...
+      for op_functor in self.op_functors:
+        with ops.Graph().as_default():
+          name, time = self._runGraph(op_functor, outer_dim, ratio, inner_dim, dtype)
+          t.append(time)
 
       # print out the speed up factor for each test
-      print("{:d}\t{:d}\t{:d}\t{:f}".format(
-          outer_dim, output_outer_dim, inner_dim, t[1]/t[0]))
+      print("__bench__ {} {:d}\t{:d}\t{:d}\t{:f}".format(
+          self._npTypeToStr(dtype), outer_dim, output_outer_dim, inner_dim, t[1]/t[0]))
```
zheng-xq: Most benchmark comparison uses percentage in the difference. But it's up to you. Also mention "positive means the new implementation is faster".

Given some kernels do get slower. Add a TODO to investigate why that is the case. 

Thanks for your contribution!

zheng-xq:
LGTM, after minor changes.

tjingrant:
I have modified the commit message to use percentage differences and a todo note is added to the file segment_reduction_ops.cc.

zheng-xq:
LGTM

zheng-xq (approved):

tjingrant:
Hi @jhseu , any plan to move forward?

martinwicke:
Jenkins, test this please.

tjingrant:
Hi, the error message seems unrelated to my changes:
```
00:36:53 The following tests FAILED:
00:36:53 	252 - C:/tf_jenkins/home/workspace/tensorflow-pr-win-cmake-py/tensorflow/python/training/server_lib_test.py (Timeout)
00:36:53 Errors while running CTest
```

PR #11631, opened 2017-07-20
Add logical_or and less_equal to tf_op_files.txt

resec:
Add below ops implementations to the list
tensorflow/core/kernels/cwise_op_logical_or.cc
tensorflow/core/kernels/cwise_op_less_equal.cc

When I am using the Android and iOS lib, above ops are not registered while in full version they are available. I think these are very basic ops and should add to the list by default.


petewarden (approved):

petewarden:
Jenkins, test this please.

PR #11632, opened 2017-07-20
Add interface for long(int64) datatype

resec:
Added below methods for TensorFlowInferenceInterface

```Java
public void feed(String inputName, long[] src, long... dims)
public void feed(String inputName, LongBuffer src, long... dims)
public void fetch(String outputName, long[] dst)
public void fetch(String outputName, LongBuffer dst)
```

They are very useful in case when we are handling int64 Tensor (e.g. output of tf.argmax)

I do not see the unit test file for TensorFlowInferenceInterface.java, so I did not add any test case for the change. But this change is very straight forward. Please let me know if I missed any. Thanks.


andrewharp:

andrewharp (approved):

PR #11634, opened 2017-07-20
Add a function for generating random values with predefined seed

DimanNe:
I added a seeded random function, as it was requested in [this PR](https://github.com/tensorflow/tensorflow/pull/11530#issuecomment-315590857):

> Please use seeded rands to make the tests deterministic and easier to debug.

### DIFF
--- a/tensorflow/core/lib/random/random.cc
+++ b/tensorflow/core/lib/random/random.cc
@@ -22,16 +22,27 @@ limitations under the License.
 namespace tensorflow {
 namespace random {
 
-std::mt19937_64* InitRng() {
+namespace {
+std::mt19937_64* InitRngWithRandomSeed() {
   std::random_device device("/dev/urandom");
   return new std::mt19937_64(device());
 }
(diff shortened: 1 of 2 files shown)


DimanNe:
@drpngx , any suggestions?

vrv:
@drpngx is out this week, unfortunately; can this wait until he gets back?

drpngx (approved):

drpngx:
Jenkins, test this please.

drpngx:
OK, i just did this one.

vrv:
Do these need API review the same way the python API does?

drpngx:
There is no requirement enforced by the build system. I think the equivalent would be `tensorflow/cc` for API. I think of this more as an internal lib change.

vrv:
Oh it's also only in lib_internal, not a public library, so yeah this is probably fine.

vrv:
(Though a test would have been nice :)

PR #11659, opened 2017-07-21
Update normalisation.py

gautam1858:
The self for mean and vaiance has to be added as we need to retain the mean value as it is


jhseu:
The existing code is correct. Closing as intended behavior.

PR #11661, opened 2017-07-21
Definitions about @protobuf between the workspace.bzl and grpc

Shuolongbj:
grpc expects //external:protobuf_clib and //external:protobuf_compiler to point to the protobuf's compiler library.

Definitions about @protobuf between the workspace.bzl and grpc(https://github.com/grpc/grpc/blob/master/WORKSPACE)

### DIFF
--- a/tensorflow/examples/tutorials/word2vec/word2vec_basic.py
+++ b/tensorflow/examples/tutorials/word2vec/word2vec_basic.py
@@ -253,7 +253,7 @@ try:
   from sklearn.manifold import TSNE
   import matplotlib.pyplot as plt
 
-  tsne = TSNE(perplexity=30, n_components=2, init='pca', n_iter=5000)
+  tsne = TSNE(perplexity=30, n_components=2, init='pca', n_iter=5000, method='exact')
   plot_only = 500
   low_dim_embs = tsne.fit_transform(final_embeddings[:plot_only, :])
   labels = [reverse_dictionary[i] for i in xrange(plot_only)]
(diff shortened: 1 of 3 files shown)


jhseu:
(Adding Justine because I don't understand our protobuf Bazel config). Some unrelated changes made it in, please revert.

Shuolongbj:
@jhseu @jart thanks !
Definitions about @protobuf between the tensorflow.bzl and [grpc](https://github.com/grpc/grpc/blob/master/WORKSPACE)

jart:
Mr. Jenkins: test this please

jart:
I don't understand this change. Why change the name from `@protobuf` to `@com_google_protobuf`? The only difference between the two is the noinline patch we have.

Shuolongbj:
@jart 
thanks! 
tensorflow binds @protubuf to the protobuf repository at workspace.bzl, while https://github.com/grpc/grpc/blob/master/WORKSPACE binds @protobuf to the build target (not the repo).  the build may fail due to name clash, the error message is something like external/protobuf does not have attribute testonly, which basically says Bazel is trying to use @protobuf as a target but in fact the name is pulled in as a repo by tensorflow, if you have a third repository and pulling in both tensorflow and grpc as external repositories.

snnn:
Why not replace all @protobuf occurrences and remove @protobuf from WORKSPACE?
like this: https://github.com/snnn/tensorflow/commit/341bb1e207e299a942695a2db1bc823457549a29

Shuolongbj:
thank you ! @jhseu @jart @snnn 
Upgrade gRPC (#11768)

Issue #8953, opened 2017-04-04
Enable full message with tf.Print


MicaelCarvalho:
**Operating System:** Debian 4.8.15-2
**Installed version of CUDA and cuDNN:** CUDA 8, cuDNN 5
**python3 -c "import tensorflow; print(tensorflow.__version__)":** 1.0.1

**Minimal reproducible example**
```
import tensorflow as tf
ph = tf.placeholder(tf.float32, [3,4,5,6])
ts = tf.shape(ph)
tp = tf.Print(ts, [ts])
tm = tf.reduce_mean(tp)
sess = tf.Session()
res = sess.run(tm)
```

**Output**
`I tensorflow/core/kernels/logging_ops.cc:79] [3 4 5...]`

Maybe there is some obvious way around this, but I couldn't find any on the [docs](https://www.tensorflow.org/api_docs/python/tf/Print). To state the obvious : I'm trying to see the full shape of the tensor, and not only the first 3 positions. In this example its size is 4, but I would like to be able to see whatever I want with whatever size it has.

Furthermore, debugging into TensorFlow did became a problem when I got to line [62 of logging_ops.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/logging_ops.py#L62), where `gen_logging_ops` is used but it doesn't seem to be declared anywhere — and older versions already had this, I'm probably missing some link, because `grep` and `find` couldn't find anything useful:

**Output of `grep`**
```
tensorflow$ grep -R "gen_logging_ops" .
./tensorflow/python/framework/function_test.py:from tensorflow.python.ops import gen_logging_ops
./tensorflow/python/framework/function_test.py:      check = gen_logging_ops._assert(math_ops.greater(x, 0), [x])
./tensorflow/python/ops/control_flow_ops.py:from tensorflow.python.ops import gen_logging_ops
./tensorflow/python/ops/control_flow_ops.py:      return gen_logging_ops._assert(
./tensorflow/python/ops/control_flow_ops.py:        return gen_logging_ops._assert(
./tensorflow/python/ops/summary_ops.py:from tensorflow.python.ops import gen_logging_ops
./tensorflow/python/ops/summary_ops.py:from tensorflow.python.ops.gen_logging_ops import *
./tensorflow/python/ops/summary_ops.py:    val = gen_logging_ops._tensor_summary(
./tensorflow/python/ops/logging_ops.py:from tensorflow.python.ops import gen_logging_ops
[...]

PR #11663, opened 2017-07-21
Add name of C++ source file to generated Python files for ops.

frreiss:
While tracing back through the Python logging APIs to find their implementation, I ran into a dead end at the generated file `gen_logging_ops.py`. This generated file contains no information about where it came from. It took a fair amount of time poking through Bazel build files to track down the rule that generates this file from the output of a program that is linked against an object file that is produced by compiling `logging_ops.cc`.

Other developers don't seem to have gotten that far; for example the reporter https://github.com/tensorflow/tensorflow/issues/8953 , who gave up without finding `logging_ops.cc`.

This PR adds the name of the original C++ source file to the generated Python files for built-in operators. I've modified `python_op_gen_main.cc` to check whether it is running from one of the special executables that the Bazel build creates for these internal operators. If the program is running from inside such a rule, it generates a header comment with the name of the original C++ source file:
```
"""Python wrappers around TensorFlow ops.

This file is MACHINE GENERATED! Do not edit.
Original C++ source file: logging_ops.cc
"""
```
Note the last line, "Original C++ source file...". This should be enough information for someone new to the project to track down the implementation of a built-in operator.

I would have preferred to modify the `REGISTER_OP()` macro to incorporate information about the original source file in the `OpDef` itself, but doing so would have required modifying `op_def.proto`. Hence this approach of checking the executable name in `python_op_gen_main.cc`.


asimshankar:
Adding @josh11b who has been looking at the python generator lately.

tensorflow/python/framework/python_op_gen.cc
```diff
@@ -710,14 +710,24 @@ string GetPythonOp(const OpDef& op_def, const string& function_name) {
 }
 
 string GetPythonOps(const OpList& ops, const std::vector<string>& hidden_ops,
-                    bool require_shapes) {
+                    const string source_file_name, bool require_shapes) {
```
josh11b: In general, please use `const string&` instead of `const string` in function arguments.

josh11b (changes requested):

josh11b:
Thanks for the PR! I think we may eventually need to add the filename to the OpDef itself, but this is a quicker fix.

tensorflow/python/framework/python_op_gen.cc
frreiss: Sure, no problem. Thanks for the feedback! Updated the PR accordingly.

tensorflow/python/framework/python_op_gen_main.cc
```diff
@@ -78,10 +79,33 @@ Status ParseHiddenOpsCommandLine(const char* arg,
 }
 
 void PrintAllPythonOps(const std::vector<string>& hidden_ops,
+                       const string& source_file_name,
                        bool require_shapes) {
   OpList ops;
   OpRegistry::Global()->Export(false, &ops);
-  PrintPythonOps(ops, hidden_ops, require_shapes);
+  PrintPythonOps(ops, hidden_ops, source_file_name,require_shapes);
```
josh11b: Space after comma.  I think there is some program to automatically format the code, but I don't have instructions.

josh11b (changes requested):

tensorflow/python/framework/python_op_gen_main.cc
frreiss: Found that program -- `clang-format -style=Google` -- and reformatted the file properly.

rmlarsen:
@josh11b can you take another look, please?

tensorflow/python/framework/python_op_gen.cc
```diff
@@ -713,14 +713,24 @@ string GetPythonOp(const OpDef& op_def, const string& function_name) {
...
   // TODO(josh11b): Mention the library for which wrappers are being generated.
   strings::StrAppend(&result, R"("""Python wrappers around TensorFlow ops.
 
 This file is MACHINE GENERATED! Do not edit.
-"""
+)");
+ 
+  // Mention the original source file so someone tracing back through generated
+  // Python code will know where to look next.
+  if (source_file_name.size() > 0) {
```
josh11b: The idiom we use through this code base is to say
if (!source_file_name.empty())

josh11b (dismissed):
This is good with that one change is made.  Thanks!

jhseu:
@frreiss Friendly ping for an update. Thanks!

tensorflow/python/framework/python_op_gen.cc
frreiss: Sorry, didn't notice this review comment the last time I checked up on the PR. Thanks for the reminder! I've fixed my local copy. Will push the changes once I've merged updates from the trunk and run tests.

frreiss:
Update: Made the remaining requested change from the previous review, then merged with the latest changes from main. On merging, I found that `python_op_gen_main.cc` now uses the routines in `python_eager_op_gen.cc` instead of using `python_op_gen.cc`. So I moved the code that adds the "Original C++ source" notice over to `python_eager_op_gen.cc`. I ran the build, verified that the generated files still get "original c++ source" notices, and ran all the unit tests under `//tensorflow/python/...`.

On a side note, the code in `python_op_gen.cc` and `python_op_gen.h` doesn't appear to be called from anywhere any more. Is that intentional?

martinwicke (approved):

martinwicke:
Approving since the changes @josh11b were made.

martinwicke:
Jenkins, test this please.

Issue #8581, opened 2017-03-21
tf.read_file doesn't support non ASCII characters in filename


wangxianliang:
Windows 2003 server R2, if input file names contain non ASCII characters (such as Chinese), exception will raise.

     [[Node: ReadFile_17 = ReadFile[_device="/job:localhost/replica:0/task:0/cpu:0"](unstack_5:2)]]

Caused by op 'ReadFile_17', defined at:
  File "train.py", line 660, in <module>
    tf.app.run()
  File "C:\Anaconda3\lib\site-packages\tensorflow\python\platform\app.py", line 44, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "train.py", line 657, in main
    train()
  File "train.py", line 430, in train
    num_preprocess_threads=num_preprocess_threads)
  File "input.py", line 216, in distorted_inputs
    num_readers = FLAGS.num_readers)
  File "input.py", line 256, in batch_inputs
    image_buffer = tf.read_file(filename)
  File "C:\Anaconda3\lib\site-packages\tensorflow\python\ops\gen_io_ops.py", line 203, in read_file
    result = _op_def_lib.apply_op("ReadFile", filename=filename, name=name)
  File "C:\Anaconda3\lib\site-packages\tensorflow\python\framework\op_def_library.py", line 768, in apply_op
    op_def=op_def)
  File "C:\Anaconda3\lib\site-packages\tensorflow\python\framework\ops.py", line 2334, in create_op
    original_op=self._default_original_op, op_def=op_def)
  File "C:\Anaconda3\lib\site-packages\tensorflow\python\framework\ops.py", line 1226, in __init__
    self._traceback = _extract_stack()

NotFoundError (see above for traceback): Can not get size for: 01_瑜_057_14734860
91829.jpg : \u03f5\u0373\udcd5\u04b2\udcbb\udcb5\udcbd\u05b8\udcb6\udca8\udcb5\udcc4\udcce\u013c\udcfe\udca1\udca3

         [[Node: ReadFile_17 = ReadFile[_device="/job:localhost/replica:0/task:0/cpu:0"](unstack_5:2)]]

drpngx:
Right. We need to call the right function on windows. It would be nice if you could submit a PR.

skye:
[...]

PR #11562, opened 2017-07-17
https://github.com/tensorflow/tensorflow/issues/9481

HectorSVC:
Descrition:
Fix the bug that Tensorflow on Windows running into issues when there's UTF8 encoded characters in the file path.
Solution:
Switch to use WideChar API calls for Windows, like the CreateFileW,FindFirstFileW,FindNextFileW, LoadLibraryExW.
Test:
Install Tensorflow on Windows with path have Chinese characers in it. No issue.
Run command "from tensorflow.contrib.rnn.python.ops.gru_ops import *" No issue.

PR #11664, opened 2017-07-21
make tf.read_file support non ASCII characters in filename

wangxianliang:
fixes the bug in issue #8581 "tf.read_file doesn't support non ASCII characters in filename".


wangxianliang:
I signed it!

mrry:
Thanks for this contribution. As it happens, we've received a concurrent PR (#11562) that fixes the same bug, and it is going to be merged soon. I'm going to close this PR, as I think the other one covers more cases, but thanks again!

PR #11672, opened 2017-07-21
Update monitors.py

alanyee:
Renaming PrintTensor to Logging Tensor

### DIFF
--- a/tensorflow/contrib/learn/python/learn/monitors.py
+++ b/tensorflow/contrib/learn/python/learn/monitors.py
@@ -24,7 +24,7 @@
 @@GraphDump
 @@LoggingTrainable
 @@NanLoss
-@@PrintTensor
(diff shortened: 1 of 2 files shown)


jhseu:
Up to Jianwei, but I'm assuming we're not taking renames here and will do any necessary renames during the move to core.

alanyee:
Ah, I just saw a TODO on renaming

jhseu:
Yeah, I think the blocker is doing the mass rename internally.

terrytangyuan:
@jhseu Only renaming the ones in contrib, right?

jhseu:
@terrytangyuan I meant also updating internal Google code that might refer to it.

Given that there's a TF core equivalent called LoggingTensorHook, I'll close this out.

Issue #3469, opened 2016-07-22
Raspberry Pi Bazel Build "Define appropriate PLATFORM macro" error


samjabrahams:
When building TensorFlow on Raspberry Pi with Bazel, the pre-compiler end up throwing this error:

```
Define the appropriate PLATFORM_<foo> macro for this platform
```

From what I've been able to gather, here's the cause of the issue:
1. Because Raspberry Pi uses an ARM CPU, the `IS_MOBILE_PLATFORM` macro is defined [in `tensorflow/core/platform/platform.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/platform.h#L45)
2. Later, when trying to compile [png.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/png.h#L23-L26), [gif.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/gif.h), or [jpeg.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/jpeg.h), the if/else statements fall through, leaving with the default option to throw an error.
### Environment info

Operating System: Raspbian GNU/Linux, version 8.0

Installed version of CUDA and cuDNN: N/A

Installed from source.
1. The commit hash : e95f4e760c6b6713b6b686ebeff9a1586a5831dd
2. The output of `bazel version` : 

```
Build label: 0.2.1-2016-06-13 (@447f7f3)
Build target: bazel-out/local_linux-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Mon Jun 13 19:05:24 2016 (1465844724)
Build timestamp: 1465844724
Build timestamp as int: 1465844724
```
### Steps to reproduce
1. Build TensorFlow from source on Raspberry Pi [as described in this guide](https://github.com/samjabrahams/tensorflow-on-raspberry-pi/blob/master/GUIDE.md) _without_ deleting [this line from `tensorflow/core/platform/platform.h`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/platform.h#L45), which defines `IS_MOBILE_PLATFORM` when compiling on ARM devices.
### What have you tried?
1. Deleted the line mentioned above, which fixes the problem for local builds
### Logs or other output that would be helpful

[define_macro_error.txt](https://github.com/tensorflow/tensorflow/files/378789/define_macro_error.txt)

[...]

PR #11675, opened 2017-07-21
Add Raspberry PI build

ebrevdo:
Does what it says on the box; follow instructions in third_party/toolchains/cpus/arm/build_raspberry_pi.sh


tensorflow/core/platform/platform.h
```diff
@@ -43,9 +43,10 @@ limitations under the License.
 #elif defined(__arm__)
 #define PLATFORM_POSIX
 
-// Since there's no macro for the Raspberry Pi, assume we're on a mobile
-// platform if we're compiling for the ARM CPU.
+// Require an outside macro to tell us if we're building for Raspberry Pi.
+#if !defined(RASPBERRY_PI)
```
petewarden: Can you add this define to the Makefile for the Pi too? Hopefully that will go away eventually, but for now it would be good to keep them in sync.

petewarden (approved):
One minor comment.

tensorflow/core/platform/platform.h
ebrevdo: Done.

tools/arm_compiler/build_raspberry_pi.sh
```diff
@@ -0,0 +1,43 @@
...
+# you have a python directory in /usr/include/arm-linux-gnueabihf/.
+
+# To install the architecture includes for python on ubuntu trusty;
+# run:
+#  sudo dpkg --add-architecture armhf
+#  echo "deb [arch=armhf] http://ports.ubuntu.com/ trusty main universe" \
+#    sudo tee -a /etc/apt/sources.list.d/armhf.list
+#  # Ignore errors about missing armhf packages in other repos.
+#  sudo apt-get update
+#  sudo apt-get install libpython-all-dev:armhf
```
petewarden: I had to use aptitude instead of apt-get on my Goobuntu machine, since otherwise I saw:

Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libpython-all-dev:armhf : Depends: libpython-dev:armhf (= 2.7.5-5ubuntu3) but it is not going to be installed
                           Depends: libpython2.7-dev:armhf but it is not going to be installed
E: Unable to correct problems, you have held broken packages.

ebrevdo: Interesting.  I updated the comment to suggest using aptitude instead.

It did work for you in the end though?

tools/arm_compiler/build_raspberry_pi.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+set -e
+
+# To be able to build any of the python packages, it is assumed that
+# you have a python directory in /usr/include/arm-linux-gnueabihf/.
+
+# To install the architecture includes for python on ubuntu trusty;
+# run:
+#  sudo dpkg --add-architecture armhf
+#  echo "deb [arch=armhf] http://ports.ubuntu.com/ trusty main universe" \
```
petewarden: I think there should be a pipe '|' in here?

tensorflow/workspace.bzl
```diff
@@ -83,6 +84,49 @@ temp_workaround_http_archive = repository_rule(
...
+def _unsafe_unzip_http_archive_impl(repo_ctx):
+  repo_ctx.symlink(repo_ctx.attr.build_file, "BUILD")
+  repo_ctx.download(repo_ctx.attr.urls, "_archive.zip", repo_ctx.attr.sha256)
+  _unzip(repo_ctx, "_archive.zip", repo_ctx.attr.strip_prefix)
+  _run_cmd(repo_ctx, ["rm", repo_ctx.path("_archive.zip")])
+  repo_ctx.symlink("/usr/include", "local_include/")
+
+
+# This performs an unsafe unzip operation - since zip files may have
+# relative and absolute paths, it should only be executed on trusted files.
```
gunan: @meteorcloudy Can bazel handle zip files with symlinks in windows now?
Would this break windows bazel build (not currently run in presubmits)

tools/arm_compiler/BUILD
```diff
@@ -0,0 +1,49 @@
+package(default_visibility = ["//visibility:public"])
```
gunan: Should we move this and its crosstool file under third_party/cpus/cc ?

tensorflow/workspace.bzl
```diff
@@ -83,6 +84,49 @@ temp_workaround_http_archive = repository_rule(
...
+      fail("BAZEL_SH environment variable is not set")
+    cmd = [bazel_sh, "-c", " ".join(cmd)]
+  if check:
+    _execute_and_check_ret_code(repo_ctx, cmd)
+  else:
+    return repo_ctx.execute(cmd, timeout=10)
+
+def _unzip(repo_ctx, zip_file, strip_prefix):
+  _run_cmd(repo_ctx,
+           ["unzip", repo_ctx.path(zip_file),
```
gunan: should we protect this with an OS check?
How platform independent is unzip command?
macos should be OK, but what would this do on windows @meteorcloudy

arm_compiler.BUILD
```diff
@@ -0,0 +1,81 @@
+package(default_visibility = ['//visibility:public'])
```
gunan: Should this file also be under third_party/ ?

tensorflow/workspace.bzl
ebrevdo: isn't this going to be running inside a sh shell within windows?  i.e., if mkdir and cp work, unzip should too...

ebrevdo: (anyway; i'm pretty sure that the crosstool compiler won't run on windows anyway...)

tensorflow/workspace.bzl
ebrevdo: no; this part is only executed if you run with the --crosstool argument in the .sh script.  otherwise none of these commands are executed.

arm_compiler.BUILD
ebrevdo: moved.

tools/arm_compiler/BUILD
ebrevdo: moved to third_party/toolchains/cpus/arm.

petewarden:
I've got a gist showing how to build this on a clean docker image, and I've verified the resulting wheel works on my Pi!
https://gist.github.com/petewarden/6aa9650fe72d88889d24efd76598f8cb

Adding in @samjabrahams to the thread too, since he's the expert on Pi Python building.

samjabrahams:
Very cool; I'm really happy to see this getting added in. I won't be able to test it personally until August (due to all my RPis being in storage), but I'm excited to give it a shot as soon as I can.

Couple questions/comments:

* [We're targeting the `armv6` instruction set](https://github.com/tensorflow/tensorflow/pull/11675/files#diff-baf419669ca66edf686a287c3dc3f23bR20)- is the plan to have Raspberry Pi 1/Zero support?
* We're not using NEON in the current build (I'm assuming because of the ARMv6 target), but I think it's provides pretty significant optimizations (@petewarden knows more about the details of this than myself). Hopefully, the difference in speed between this and a build with NEON is negligible, but if not, would it be possible to consider adding an option to the build to use it? Granted, this might cause some headaches in the various scripts and would require using GCC 4.8 (I think we could swap in [this gnueabi](https://github.com/raspberrypi/tools/tree/master/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian/lib/gcc/arm-linux-gnueabihf/4.8.3), but it's older and also not my area of expertise).

Finally, it's satisfying that #3469 is getting addressed almost exactly a year to the day after it was opened :D

ebrevdo:
Anything else left to fix for this?  @petewarden should i move any of your docker comments into my PR; or do you want to have a separate .sh file after this one is pushed?  I'll be OOO for 2 weeks starting Wednesday, so any changes should happen by tomorrow.

ebrevdo:
@samjabrahams happy to help.  i am happy to add an argument or some comments in the .sh file for making an optimized build for arm7; or alternatively building for arm7 + neon and showing how to make an arm6 build (probably leaning towards this latter approach).

petewarden:
@ebrevdo I've got an outstanding PR to add ARM7/NEON support and update the comments here, can you take a look? https://github.com/ebrevdo/tensorflow/pull/3

gunan (approved):

gunan:
Looks good after a rebase.
Test failures do not look related to me, we can rerun them after the rebase.

ebrevdo:
Added @petewarden's changes to make default build armv7/NEON.  This required upgrading the version of Eigen we use.  @benoitsteiner is this OK?

petewarden:
@benoitsteiner I believe this addresses issue https://github.com/tensorflow/tensorflow/issues/9697

vrv:

petewarden:
The remaining MacOS failures look like timeouts or flakes.

samjabrahams:
Awesome! Once I find some time, I'll note in my repo that official support is available and link to the relevant instructions.

ebrevdo:
I think we need to wait for pete to update the version of eigen and fix any
breakages before we can make it official.

petewarden:
That's correct. We're blocked on #9697 for the Pi 3 build, and once that's done I also need to create a Jenkins CI build to run nightly, so we'll have automatically generated binaries for the wheel. I hope to get to that soon.

ptone:
Once done, will these wheels be available on PyPI for pip installation?

https://pypi.python.org/pypi/tensorflow

ebrevdo:
if we can figure out how to fake the architecture to bdist_wheel after the
cross compile, maybe.

ptone:
I believe the grpc team sorted this out somehow - I can forward you an internal thread.

https://pypi.python.org/pypi/grpcio/1.4.0

ptone:
I'm not sure what happened, (git-gone-funky?), but the update to the Eigen version did not make it into master:
https://github.com/tensorflow/tensorflow/pull/11675/commits/c8485830c2fb5add5e6dcf5fad7d3920d59622c0#diff-455a4c7f8e22d7c514e8c2caa27506c5R162

As of the current [master commit](https://github.com/tensorflow/tensorflow/blob/91d0f2eaa8e8bf340c48a36f32140ef0cdbda0a8/tensorflow/workspace.bzl) the eigen archive is still set to rev f3a22f35b044

petewarden:
@ptone Sorry, I should have been more clear in the comments. We're blocked on updating Eigen (#9697) because the newer Eigen version breaks on MacOS. We'll need to see if other Eigen versions work on both, but I'm not working on that at the moment.

If anyone wants, they can update the Eigen hash to the Pi 3-capable one locally and build the wheel, but we won't be able to support it officially until the MacOS issue is resolved.

AnchoretBaladev:
Hello @petewarden I'm trying to cross compile tensorflow for Raspberry PI 3 from Ubuntu 16.04LTS using the "build_raspberry_pi.sh" present inside "/tensorflow/tools/ci_build/pi/" with the following steps:-
1:- Installing Bazel
2:- Cloning the Tensorflow and runnining the "build_raspberry_pi.sh" 
but not getting success!!
Can you please guide me what are the needed steps to follow to cross compile!!! Thanks in Advance!!

Issue #11241, opened 2017-07-03
[Feature request] add checkpoint_convert.py script to pip package


albertz:
In the TensorFlow 1.2.0 pip package (at least for Linux), [checkpoint_convert.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/rnn/python/tools/checkpoint_convert.py) is not included. Because of #11168, 92da8abfd35b93488ed7a55308b8f589ee23b622, 157370e5916b85c65958ed8383ae31d727228ed7, it might be useful. Esp., in our framework, I would like to add some automatic handling at runtime when a checkpoint file is loaded and some variables are not found, to automatically try to load variables under different names according to `_RNN_NAME_REPLACEMENTS`. Thus, extending to adding `checkpoint_convert.py` to the pip package, it would also be nice to make `_RNN_NAME_REPLACEMENTS` public (remove the leading underscore).

cy89:
@yifeif can you please comment on whether adding checkpoint_convert.py would be OK?

frreiss:
I'm looking into this.

frreiss:
Created PR https://github.com/tensorflow/tensorflow/pull/11691

PR #11691, opened 2017-07-23
[Issue 11241] Add checkpoint_convert.py script to package and make _RNN_NAME_REPLACEMENTS public.

frreiss:
As requested in https://github.com/tensorflow/tensorflow/issues/11241 , this PR adds the script https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/rnn/python/tools/checkpoint_convert.py to the TensorFlow pip package and removes the leading underscore from the name of `_RNN_NAME_REPLACEMENTS` to indicate that this constant can be used by user code.

Testing done:
* Verified that `checkpoint_convert.py` is not included in the package prior to this change.
* Verified that `checkpoint_convert.py` is included in the package after this change.
* Verified that `checkpoint_convert.py` does not complain about missing imports when invoked from the command line after installing the pip package.
* Ran unit tests in `contrib/rnn` after the change.


ebrevdo (approved):

vrv:

tensorflow/contrib/rnn/BUILD
```diff
@@ -26,7 +26,8 @@ load(
 
 tf_custom_op_py_library(
     name = "rnn_py",
-    srcs = ["__init__.py"] + glob(["python/ops/*.py"]),
+    srcs = ["__init__.py"] + glob(["python/ops/*.py"]) 
+                           + glob(["python/tools/*.py"]),
```
josh11b: How about this instead:
  srcs = ["__init__.py", "python/tools/checkpoint_convert.py"] + glob(["python/ops/*.py"])
This excludes the test by not using the glob, instead of moving the test into its own directory (creating two directories with one file each).

josh11b (changes requested):
For API review.

tensorflow/contrib/rnn/BUILD
frreiss: Works for me. I actually did it that way at first but changed my mind. I've updated this PR.

tensorflow/contrib/rnn/BUILD
```diff
@@ -27,7 +27,7 @@ load(
 tf_custom_op_py_library(
     name = "rnn_py",
     srcs = ["__init__.py"] + glob(["python/ops/*.py"]) 
-                           + glob(["python/tools/*.py"]),
+                           + glob(["python/tools/.py"]),
```
josh11b: This doesn't look right.

frreiss: Well, that was embarrassing. Changed it to the right filename and verified that the script actually gets included in the pip package this time around.

vrv:

josh11b (approved):

asimshankar:
(Tests are failing because of the BUILD file formatting (see [logs](https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/5418/consoleFull)).  

Please fix those and you should be good to go. Thanks!
Details:

```
=== Sanity check step 3 of 10: do_buildifier (buildifier check) ===

Running do_buildifier on 208 files

tensorflow/contrib/rnn/BUILD # reformat 

buildifier took 0 s

29,30c29,31
<     srcs = ["__init__.py"] + glob(["python/ops/*.py"]) 
<                            + ["python/tools/checkpoint_convert.py"],
---
>     srcs = ["__init__.py"] + glob(["python/ops/*.py"]) + [
>         "python/tools/checkpoint_convert.py",
>     ],
```

vrv:

PR #11694, opened 2017-07-23
Add `as_default()` to `MonitoredSession`, `MonitoredTrainingSession` and `SingularMonitoredSession`.

darrengarvey:
Equivalent to `tf.Session.as_default()`, allows refetching the session from elsewhere in code using `tf.get_default_session()`.

Notes:
- (additive) public API change.
- If there is a better way to get the current session when using these `tf.Session` wrappers, this may be unnecessary. I just couldn't find it.

### DIFF
--- a/tensorflow/python/training/monitored_session.py
+++ b/tensorflow/python/training/monitored_session.py
@@ -525,6 +525,16 @@ class _MonitoredSession(object):
   def close(self):
     self._close_internal()
 
+  def as_default(self):
+    """Returns a context manager that makes this object the default session.
+
+    Use with the `with` keyword to specify that calls to @{tf.Operation.run}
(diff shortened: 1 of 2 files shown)


darrengarvey:
Hmm, just noticed [this comment](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/monitored_session.py#L649) saying a `MonitoredTrainingSession` cannot be set as the default session. No mention of _why_ though. Perhaps there's some unspecified assumption that this PR would break?

vrv:
@darrengarvey I believe that this is because MonitoredTrainingSession isn't a Session, it's something that wraps a Session, so does not implement the interface that as_default() would require.  If it was a subclass of Session, this would be possible, but I think by design it is not.  :(

PR #11697, opened 2017-07-23
[OpenCL] Extends matmul_benchmark.py to cover SYCL

lukeiwanski:


vrv:

lukeiwanski:
`googletest.gpu_device_name()` returns _/gpu:0_ but `self._StripGraph(gd)` returns proto that contains _/device:GPU:0_ 
There is a lot of inconsistency across the tests - both /gpu:0 and /device:GPU:0 are used. 
Is there intention to unifie them in the future?

vrv:
If you want to send a CL to change gpu_device_name to return /device:GPU:0, that would be great!

yzhwang:
Thanks for the pull request @lukeiwanski !
Given that you have started working on this, would you also consider resolving the check failure as @vrv suggested? Thank you!

lukeiwanski:
I am looking into this - It might be quite big changeset as everything point to gpu device creation (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L591) as _gpu_device_name_ (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/test_util.py#L61) returns an element from _list_local_devices()_
that trails back to the _ListDevices_  (a list of the devices registered in the session)

The dev branch is here: https://github.com/lukeiwanski/tensorflow/tree/dev/gpu_device

vrv:
You might be able to get away with renaming that to '/device:GPU:0' -- I can try this today too and see what breaks.  We've been trying to make sure we canonicalize the device name to the 'new style names' whenever needed, but it's possible there are some places that aren't fixed.

lukeiwanski:
@vrv had to make changes in few places :)

Could you guys have a look if the changes are reasonable?

Btw. something odd happened in the upstream as after merge I see: 

`ERROR: No test targets were found, yet testing was requested.` 
When running: 
```bazel test --config=cuda -c opt --local_test_jobs=1 --run_under=//tensorflow/tools/ci_build/gpu_build:parallel_gpu_execute --distinct_host_configuration=false --test_tag_filters=local,-no_gpu,-no_oss,-oss_serial,-benchmark-test -- //tensorflow/... -//tensorflow/compiler/...```

**Comment by rmlarsen** [comment]

**Comment by rmlarsen** [comment]

**Comment by lukeiwanski** [comment]
There are few tests failing eg: 
```
//tensorflow/core/profiler/internal:tfprof_show_test
//tensorflow/core/profiler/internal:tfprof_stats_test
//tensorflow/core/profiler/internal:tfprof_timeline_test
```
They all reading graph data from ```tensorflow/core/profiler/internal/testdata``` that have gpu in the file instead device:GPU not sure how to approach this.

**Comment by rmlarsen** [comment]
@lukeiwanski thanks for updating the datafile! 

**Comment by rmlarsen** [comment]
@lukeiwanski it looks like there are still some changes in the profiler tests.
@panyx0718 could you help Luke sort this out?

**Comment by panyx0718** [comment]
quote "googletest.gpu_device_name() returns /gpu:0 but self._StripGraph(gd) returns proto that contains /device:GPU:0"
what's the reason to faover "/device:GPU:0" over "/gpu:0"? It seems the "device:" is redundant and avoiding it allows to keep backward compatibility?

**Comment by panyx0718** [comment]
If you want to fix the tfprof tests, see here.
https://github.com/tensorflow/tensorflow/blob/92111fdd1a9baa113264149413105f700f9ee310/tensorflow/core/profiler/internal/tfprof_node.cc#L28

There might be one or more places to fix. profiler depends on the device name format to correctly analyze metrics of each operation. Hence, those tests should be kept correct. One solution to keep backward compatibility with old RunMetadata is to support all device naming schemes, instead of just replacing them.

**Comment by lukeiwanski** [comment]
@panyx0718 the reason for 
> what's the reason to faover "/device:GPU:0" over "/gpu:0"? It seems the "device:" is redundant and avoiding it allows to keep backward compatibility?

see @vrv comment here: https://github.com/tensorflow/tensorflow/pull/11697#issuecomment-317800870 
It seems that /device:XXX:x is desired way to report supported devices.. in our effort to add SYCL we have been advised to use /device:SYCL:x 
That inconsitency in the way how devices are reported makes testing generalisation painful.
The backward compatiblity is intact as you still can register /gpu:0 I believe that the device gets "canonicalized to new style names" (https://github.com/tensorflow/tensorflow/pull/11697#issuecomment-318111065)

**Comment by vrv** [comment]
Yes, the problem with the old style names like '/gpu:0' is that you don't know ahead of time when parsing a device name whether it even is a device name.  For example, we know the valid names are '/job:foo', '/task:N', etc.  But if you allow '/anything:here', the parsing logic basically has to assume that anything other than 'job', 'task', and 'replica' must be a device, which is error prone.

So having 'device' in the name allows us to unambiguously know that the entry is a 'device type'.  We had to backwards-compat the old cpu and gpu names because they were too widely used.

**Comment by vrv** [comment]
For internal folks, the related bug about replacing old device names is b/37868888

File: tensorflow/core/profiler/README.md [hunk 1]
```diff
@@ -127,7 +127,7 @@ tfprof> advise
 Not running under xxxx. Skip JobChecker.
 
 AcceleratorUtilizationChecker:
-device: /job:worker/replica:0/task:0/gpu:0 low utilization: 0.03
+device: /job:worker/replica:0/task:0/device:GPU:0 low utilization: 0.03
 device: /job:worker/replica:0/task:0/gpu:1 low utilization: 0.08
```
panyx0718: I assume gpu:x are all changed?

tensorflow/core/profiler/g3doc/advise.md
```diff
@@ -31,7 +31,7 @@ tfprof --graph_path=graph.pbtxt \
 
 tfprof> advise
 AcceleratorUtilizationChecker:
-device: /job:worker/replica:0/task:0/gpu:0 low utilization: 0.03
+device: /job:worker/replica:0/task:0/device:GPU:0 low utilization: 0.03
 device: /job:worker/replica:0/task:0/gpu:1 low utilization: 0.08
```
panyx0718: same here

tensorflow/core/profiler/g3doc/profile_time.md
```diff
@@ -134,7 +134,7 @@ AddN                            50.10ms (17.33%, 1.34%),       5481
 tfprof> op -select micros,device -order_by micros
 node name | execution time | assigned devices
 SoftmaxCrossEntropyWithLogits     1.37sec (100.00%, 36.44%), /job:worker/replica:0/task:0/cpu:0
-MatMul                        618.97ms (63.56%, 16.51%), |/job:worker/replica:0/task:0/cpu:0|/job:worker/replica:0/task:0/gpu:0|/job:worker/replica:0/task:0/gpu:1|/job:worker/replica:0/task:0/gpu:2|/job:worker/replica:0/task:0/gpu:3
+MatMul                        618.97ms (63.56%, 16.51%), |/job:worker/replica:0/task:0/cpu:0|/job:worker/replica:0/task:0/device:GPU:0|/job:worker/replica:0/task:0/gpu:1|/job:worker/replica:0/task:0/gpu:2|/job:worker/replica:0/task:0/gpu:3
```
panyx0718: same here

tensorflow/core/profiler/internal/tfprof_node.cc
```diff
@@ -25,7 +25,7 @@ bool CountAsAcceleratorTime(const string& device) {
 }
 
 bool CountAsCPUTime(const string& device) {
-  return RE2::FullMatch(device, ".*/(gpu|cpu|device:sycl):\\d+");
+  return RE2::FullMatch(device, ".*/(device:gpu|cpu|device:sycl):\\d+");
```
panyx0718: does cpu changed as well?

tensorflow/core/profiler/internal/tfprof_node.cc
```diff
@@ -158,7 +158,7 @@ void TFGraphNode::AddStepStat(int64 step, const string& device,
       if (dev.find("sycl") != dev.npos) {
         host_device_ = StringReplace(dev, "device:sycl:\\d+", "cpu:0");
       } else {
-        host_device_ = StringReplace(dev, "gpu:\\d+", "cpu:0");
+        host_device_ = StringReplace(dev, "device:gpu:\\d+", "cpu:0");
```
panyx0718: If cpu has changed as well, should we add "device:" here?

tensorflow/core/profiler/internal/tfprof_stats_test.cc
```diff
@@ -84,12 +84,12 @@ TEST_F(TFProfStatsTest, CustomOpType) {
       "name: \"_TFProfRoot\"\ntotal_exec_micros: 13\ntotal_parameters: "
       "451\nchildren {\n  name: \"DW\"\n  exec_micros: 2\n  parameters: 162\n  "
       "total_exec_micros: 2\n  total_parameters: 162\n  devices: "
-      "\"/job:localhost/replica:0/task:0/gpu:0\"\n  cpu_exec_micros: 2\n  "
+      "\"/job:localhost/replica:0/task:0/device:GPU:0\"\n  cpu_exec_micros: 2\n  "
```
panyx0718: these test use data from testdata. So I guess it might still be /gpu:0. It good to keep it pass for backward compatibility

tensorflow/core/protobuf/config.proto
```diff
@@ -42,7 +42,7 @@ message GPUOptions {
   // A comma-separated list of GPU ids that determines the 'visible'
   // to 'virtual' mapping of GPU devices.  For example, if TensorFlow
   // can see 8 GPU devices in the process, and one wanted to map
-  // visible GPU devices 5 and 3 as "/gpu:0", and "/gpu:1", then one
```
panyx0718: why not gpu:1?

tensorflow/core/profiler/internal/tfprof_node.cc
panyx0718: this might break backward compatibility. maybe (device:gpu|gpu)

panyx0718 (commented):
Some general issues:
1. only gpu:0 is updated, while gpu:1, etc are not updated.
2. need to be compatible with /gpu:0
3. tests use static test data, so it might not be affected by your changes. (i.e. still named /gpu:0)

lukeiwanski:
@vrv should I do the same for /cpu ?

tensorflow/core/protobuf/config.proto
lukeiwanski: Fixed in fb7ff72

rmlarsen:

lukeiwanski:
@panyx0718 / @rmlarsen  / @vrv thanks for the review! :)

rmlarsen:

rmlarsen:
@yzhwang @vrv can you approve this PR?

yzhwang:
coordinator_test.py failed in Windows Cmake Tests, but it is irrelevant to this PR?

vrv (approved):
Seems like for the most part, it's all tests that have switched to /device:GPU:0, which is great.  cpu:0 change could happen in another PR.  I guess this doesn't change any behavior, it just changes our codebase examples to be more explicit.

The longer-term solution would be to figure out where device names are being first interpreted and canonicalizing them as early as possible, so that the parts of code that need to handle device strings can more easily assume that it's been canonicalized before parsing them.

yzhwang (approved):

rmlarsen:
@lukeiwanski thanks for the contribution. 
Merging. The failing coordinator test is unrelated.

PR #11698, opened 2017-07-23
Exposed more graph io methods

terrytangyuan:
I think these should probably be exposed since some of these are being used in examples, such as parsing utilities [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/estimator/canned/parsing_utils.py#L86).


alextp:
Jenkins, test this please.

vrv:

rohan100jain (approved):

PR #11719, opened 2017-07-24
Add a custom tag for use by fusion ops

DavidNorman:
Currently when replacing an op with a fusion of many, the information which is carried by that fusion op is limited to a small fixed enumeration.

I have previously added a 'custom' entry to that enumeration, however it isn't really enough to label an arbitrary and non-fixed set of fuse results.

This adds a general integer that can be used by any back end to label fused ops arbitrarily.

There are other ways to achieve this:
- replacing the FusionKind enumeration with an integer field
- allowing read-write access to the metadata structure, and replacing it with something like a map
- some other thing I can't think of right now.

I'm happy to go with a different scheme, but I do think that the HloInstruction needs an annotation interface of some kind.

### DIFF
--- a/tensorflow/compiler/xla/service/hlo_instruction.h
+++ b/tensorflow/compiler/xla/service/hlo_instruction.h
@@ -627,6 +627,16 @@ class HloInstruction {
     return fusion_kind_;
   }
 
+  unsigned int fusion_custom_tag() const {
+    CHECK_EQ(HloOpcode::kFusion, opcode_);
+    return fusion_custom_tag_;
+  }
+
+  void set_fusion_custom_tag(unsigned int tag) {
+    CHECK_EQ(HloOpcode::kFusion, opcode_);
+    fusion_custom_tag_ = tag;
+  }
+
   // Merges the fused instructions from 'instruction_to_merge' into the
   // fused instruction set of 'this', updating operands as necessary.
   //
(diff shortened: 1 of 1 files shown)


DavidNorman:
actually - i've decided to abandon using fusing as a method for representing custom blocks of code - i will use sub-computations and Call ops instead.   this problem is no longer an issue.

DavidNorman:
actually - maybe this is a thing anyway.

DavidNorman:
no - sorry.  changed my mind again.  the code which looked promising re using the fusion op is the same as the code which I write - it cannot correctly fuse a set of nodes where one is referred to by 2 others.

DavidNorman:
once again sorry - i have found a way of getting fuse to behave the way I want,  so I would like this change (or something similar to allow a bit of meta-data to be attached to a fusion instruction or the internal computation).

another idea - perhaps - is to allow the API to set the name of the internal instruction.  I don't like using strings as a place to store enumerations, but it is an alternative.

yet another is to allow the API to set the name of the instruction itself - same problem with using a string as an enumeration.

DavidNorman:
@vrv I may have confused things by opening and closing this one over and over.  I am now sure that I would like some sort of annotation mechanism for fusion instructions (and ideally also computations).

vrv:
(I have very little knowledge about XLA at this point, so assigning to others!)

majnemer:
@DavidNorman Can you please explain the motivation behind FusionKind::kCustom and this custom tag? I'd like to better understand the limitations in the current scheme.

DavidNorman:
sure thing.

currently, when a fusion instruction is created it can get one of the enumeration values to indicate its type.  I added 'kCustom' because there are ops in our device which are not really fusions that fit the current set of enumerations.

I really need lots of kCustom enumeration values though, because I have more than one specialized ops. 

I added another 'sub_type' kind of field for this.   However, I think I could static_cast<> in and out of an int - which perhaps would be nicer.    A sub_type field makes the code neater. Or maybe it doesn't.   I'll make the casting change and see if it feels ok.

DavidNorman:
ok - i will just do it with the existing enum and some type casting.   closing this off :)

PR #11722, opened 2017-07-24
Update wide_n_deep_tutorial.py

alanyee:
-mkdtemp is low-level, and creates a temporary file somewhere out of sight that requires users to delete the file themselves. For the use case of a tutorial, it is best to use a method that cleanups the file when finished
-TemporaryFile does such as described above, but not available in Python 2


vrv:

vrv (approved):

PR #11731, opened 2017-07-25
Fix wrong template type name in QuantizedAddUsingEigen

junluan:
input type is T1 and smaller input type is T2


junluan:
I signed it!

vrv (approved):

vrv:

(in practice, T1 and T2 were the same type, but this matches expectation).

PR #11739, opened 2017-07-25
Add less_equal and logical_or to android_core_ops

resec:
Added below two ops to android_core_ops, as similar to #11631
tensorflow/core/kernels/cwise_op_logical_or.cc
tensorflow/core/kernels/cwise_op_less_equal.cc

I think these are very basic ops and should add to the list by default.


vrv:

vrv (approved):

PR #11747, opened 2017-07-25
typo in the beginning comment

tigercosmos:
There is typo "r" before the beginning comment.


tigercosmos:
I signed it!

caisq:
@tigercosmos: thanks for the PR. However, this is not a typo. The doc string contains special characters, which make it necessary to use the r prefix.

PR #11748, opened 2017-07-25
Ensure that the multi-instruction fuse can take shared inputs

DavidNorman:
Note that the fuse action only works when the shared input / constant
appears after all of its consumers in the list of instructions.

there isn't (currently) a test to verify that this behaviour will continue to function in the future


vrv:

DavidNorman:
I do not think that the failure in the windows test is down to my change.

vrv:
Thanks @DavidNorman, I'll wait for the review from an XLA person to move forward.

tensorflow/compiler/xla/tests/fusion_test.cc
```diff
@@ -522,6 +522,39 @@ XLA_TEST_F(FusionTest, DISABLED_ON_CPU(ReduceWindow)) {
       *ExecuteAndTransfer(std::move(hlo_module), {}));
 }
 
+XLA_TEST_F(FusionTest, SharedConstant) {
+  auto hlo_module = CreateNewModule();
```
eliben: Can you add a comment here explaining what this test is testing, on a high-level?

eliben (changes requested):
Please double check the build failures too

tensorflow/compiler/xla/tests/fusion_test.cc
DavidNorman: sure thing

DavidNorman: done.  

as it happens, the CreateFusionInstruction does the right thing at the moment (provided the list of operations is in the right order).  but nowhere it is asserted that the contents of the fusion will be the same in this case. 

I came upon it because I had the constant at a different place in the list at first, and in that case you get duplication - the CreateFusionInstruction is iterative, and can make quite a mess when partially complete.

DavidNorman: I'm also going to merge the master head, to eliminate the failure.  it's clearly nothing to do with adding an XLA unit test - but it would be good to have a clean test bill.

eliben (approved):

vrv:

vrv:
Thanks for the PR and for the review!

DavidNorman:
thanks @vrv and @eliben

DavidNorman:
This failure is due to a disk space failure on the OS/X build machine I think.

DavidNorman:
Actually - that's why the other change has failed.  this one has failed due to a test called:

//tensorflow/contrib/signal:spectral_ops_test 

I don't think that adding an XLA test can affect this test though.  I'm confused about why it failed for OS/X and not Linux.  Perhaps it is a flakey test?   I will double check the logs though.

vrv:
It's a flaky test on mac only for an unrelated op, (not sure why the numerics are bad there, the author of the test knows about it).

Issue #11113, opened 2017-06-28
Add cosine annealing for learning rate decay


adler-j:
[SGDR: Stochastic Gradient Descent With Warm Restarts](https://openreview.net/pdf?id=Skq89Scxx), proposes decaying the learning rate according to

![image](https://user-images.githubusercontent.com/2202312/27641761-1bf302c8-5c1d-11e7-8d4b-15988701ff3f.png)

where ![image](https://user-images.githubusercontent.com/2202312/27641775-28b4e0a8-5c1d-11e7-8adc-1dcaea55d77c.png) is the minimum step length, ![image](https://user-images.githubusercontent.com/2202312/27641798-3be6a80a-5c1d-11e7-9b41-d3c25b0b2b96.png) is the maximum step length, ![image](https://user-images.githubusercontent.com/2202312/27641818-4686daf0-5c1d-11e7-8991-533ce8710e8f.png) is the global step and ![image](https://user-images.githubusercontent.com/2202312/27641841-5594f2ac-5c1d-11e7-9f54-e3cc8ccd1566.png) is the maximum number of iterations.

I've personally found this strategy to  be easy to use given that the number of hyperparameters is relatively small and results are good.

Is this something we want added to tensorflow? Would you accept submissions?

shin-migami:
Can i get assigned to this ?

skye:
@ebrevdo @martinwicke can you comment?

martinwicke:
@shin-migami Consider yourself assigned!

The right spot for this is in contrib.

I cannot formally assign you, but everybody who wants to work on this should check this issue first, and find that you are already working on it. Please reference this issue in any PR you create, and update the issue when you are making progress.

michaelisard:
Adding contributions welcome tag just so it's clear that this issue has been triaged: this doesn't contradict the fact that @shin-migami is working on it!

PatrykChrabaszcz:
Hi @shin-migami 
I was asked by Ilya and Frank (authors of the paper) to implement this learning rate decay algorithm in tensorflow. Now I found out you also work on it. 

We would like to include more general version where you can also use restarts.
Did you make any progress already?

adler-j:
I guess this can be closed now?

PR #11749, opened 2017-07-25
SGDR Learning Rate Decay Algorithm

PatrykChrabaszcz:
Implementation of 
"SGDR: Stochastic Gradient Descent with Warm Restarts" by Ilya Loshchilov & Frank Hutter.
https://arxiv.org/pdf/1608.03983.pdf
Developed together with the author @loshchil
Adds cosine annealing for learning rate decay #11113.

mmul


PatrykChrabaszcz:
Recheck CLA

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
```diff
@@ -0,0 +1,146 @@
...
+  as defined in the paper.
+
+  ```
+  Args:
+    learning_rate: A scalar `float32` or `float64` `Tensor` or a
+      Python number.  The initial learning rate.
+    global_step: A scalar `int32` or `int64` `Tensor` or a Python number.
+      Global step to use for the decay computation.  Must not be negative.
+    t_0: A scalar `int32` or `int64` `Tensor` or a Python number.
+      Must be positive.  Number of iterations in the first restart,
```
vrv: Can you clarify what scale 'iterations' is?  iterations is the number of epochs used before the first restart, correct? 

So with default t_0 and t_mul, you restart after every epoch of training, correct?

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
```diff
@@ -0,0 +1,146 @@
...
+      it can be set to a multiplicative of the number of batches per epoch.
+      Defaults to 1
+    t_mul: A scalar `int32` or `int64` `Tensor` or a Python number.
+      Must be positive.
+      Used to derive the number of iterations in the i-th restart:
+      `t_0 * (t_mul^i)`. Defaults to 1
```
vrv: add a period at the end of every statement

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
```diff
@@ -0,0 +1,146 @@
...
+    m_mul: A scalar `float32` or `float64` `Tensor` or a Python number.
+      Must be positive.
+      Used to derive the initial learning of the i-th restart:
+      `learning_rate * (m_mul^i)`. Defaults to 1
+  Returns:
```
vrv: Would it be useful to provide some concrete examples here?  E.g., 

{lr = 0.01, t_0 = 1, t_mul = 1, m_mul = 0.5}:
```
Epoch | 0    | 1     |  2
LR    | 0.01 | 0.005 | 0.0025
```

{lr = 0.01, t_0 = 2, t_mul = 2, m_mul = 0.1}:
```
Epoch | 0    | 2     |  6     |  14
LR    | 0.01 | 0.001 | 0.0001 | 0.00001
```

I feel like that will give people a bit more intuition for how m_mul and t_mul affect the LR schedule, but it's up to you.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
```diff
@@ -0,0 +1,146 @@
...
+  ```
+  Args:
+    learning_rate: A scalar `float32` or `float64` `Tensor` or a
+      Python number.  The initial learning rate.
+    global_step: A scalar `int32` or `int64` `Tensor` or a Python number.
+      Global step to use for the decay computation.  Must not be negative.
+    t_0: A scalar `int32` or `int64` `Tensor` or a Python number.
+      Must be positive.  Number of iterations in the first restart,
+      it can be set to a multiplicative of the number of batches per epoch.
+      Defaults to 1
```
vrv: It currently defaults to 1000, is that a bug in the doc or the impl?

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
```diff
@@ -0,0 +1,146 @@
...
+from __future__ import print_function
+
+import math
+
+from tensorflow.python.framework import constant_op
+from tensorflow.python.framework import ops
+from tensorflow.python.ops import math_ops, control_flow_ops
+
+
+def sgdr_decay(learning_rate, global_step, t_0=1000, t_mul=1, m_mul=1,
```
vrv: Consider using 1.0 for m_mul to make it clear it's supposed to be a floating point type.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
```diff
@@ -0,0 +1,146 @@
...
+  Raises:
+    ValueError: if `global_step` is not supplied.
+  """
+
+  if global_step is None:
+    raise ValueError("global_step is required for sgdr_decay.")
+  with ops.name_scope(name, "SGDRDecay",
+                      [learning_rate, global_step,
+                       t_0, t_mul, m_mul]) as name:
+    learning_rate = ops.convert_to_tensor(learning_rate, name="learning_rate")
```
vrv: name="initial_learning_rate" ?

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class SGDRDecayTest(test_util.TensorFlowTestCase):
+
+  # Original iterative approach
+  # https://github.com/loshchil/SGDR/blob/master/SGDR_WRNs.py
+  def get_original_values(self, lr, Te, multFactor, iter_per_epoch, epochs):
+    t0 = math.pi / 2.0
+    tt = 0
+    TeNext = Te
```
vrv: our variable names usually start with lower case and are not CamelCase so maybe te_next.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
```diff
@@ -0,0 +1,136 @@
...
+      if (epoch + 1) == TeNext:  # time to restart
+        sh_lr = lr
+        tt = 0                # by setting to 0 we set lr to lr_max, see above
+        Te = Te * multFactor  # change the period of restarts
+        TeNext = TeNext + Te  # note the next restart's epoch
+
+    return lr_values
+
+  def get_sgdr_values(self, lr, Te, multFactor, iters):
+    step = gen_state_ops._variable(shape=[], dtype=dtypes.int32,
```
vrv: For the purposes of testing, is there any reason why step needs to be a variable?  You could make it a placeholder and then just feed the value of the step, and that would get rid of all of the assignment statements.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
```diff
@@ -0,0 +1,136 @@
...
+      iters = 10
+      epochs = 50
+
+      org_lr = self.get_original_values(lr, t_e, t_mul, iters, epochs)
+      sgdr_lr = self.get_sgdr_values(lr, t_e*iters, t_mul, iters*epochs)
+
+      for org, sgdr in zip(org_lr, sgdr_lr):
+        self.assertAllClose(org, sgdr)
+
+  def testMDecay(self):
```
vrv: These test names aren't very descriptive, can you add comments to each test explaining what scenario they are trying to test?  For example, my guess is in this one you are validating that an m_mul < 1 is properly computed according to the LR equation.

vrv (changes requested):
This is a very cool learning rate schedule, excited to see this in TF!  I have a lot of minor comments, but overall this looks very nice.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: We observe that all learning rate decay algorithms implemented in TensorFlow compute proper learning rate with respect to the global step.
This global step starts at 0 and is increased by one after each optimizer step (If we provide it as an argument when constructing optimizer).
So it would mean that t_0 is the duration of the first cycle measured as the number of minibatch updates (Iteration = minibatch update).
If one wants to use epochs, one should compute the number of iterations required for an epoch.
Example:
Minibatch size: 100
Training dataset size: 10 000
User wants the first decay cycle to span across 5 epochs: t_0 = 5 * 10000/100

We did it to be consistent with the other learning rate decay implementations available in TensorFlow. Original approach presented in the paper
however defines t_0 as the number of epochs. In both cases here and in the original implementation it is important to notice that the learning rate is adjusted after each minibatch update, so there is no sudden drop in the learning rate at the end of epoch.

We might need to clarify it inside the description.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: Ok

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: Sounds like a good idea. We will provide examples.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: Bug in the doc. However it might be better to remove the default value for this argument. In the most common scenario user will want compute this value
such that number of iterations is N * iterations_per_epoch.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: Yes, you are right. That will be fixed.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: That would be better indeed. I'll change that.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
PatrykChrabaszcz: I tried to make as little changes compared to the original implementation as possible. I'll fix variable naming to match TensorFlow style.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
PatrykChrabaszcz: That's right, we could use a placeholder, but this tests implementation is inspired by test cases for other learning rate decay algorithms:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/learning_rate_decay_test.py

So we use assignment ops as other tests do. Should we change it?

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
PatrykChrabaszcz: You are right, we will add comments.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
vrv: I think it probably makes more sense to use placeholders for testing.  The use of variables and assignment ops is not so a great idea for a few reasons, and given that your implementation doesn't rely on global_step being a Variable, I think it would be better for it to be a placeholder for testing.

For example, each call to .assign() adds a new node to the graph, which requires 'recompiling' the graph for each step, and generally should be avoided in current TF, so I feel like demonstrating good practice is also useful here, even though it deviates from the existing tests.  In fact, I can then point to this example as the shining example for others to copy :)

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
vrv: Given that the implementation in the paper doesn't match the interpretation here, do you think we should consider renaming the variables to avoid confusion?  Something like

t_0 would become 'initial_period_steps', so it's clear that it indicates something about 'periods', and also that it is measured in 'steps' to match global_step.  Similarly with the other variables, maybe, but I'm not sure.

Basically, I'm worried that people will think there are bugs in the implementation because the variable names are the same but the 'types' are different :)

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: We changed m_mul default value to 1.0 as you suggested, we also changed t_mul default value to 2.0,  We think that this should also be a float not an integer, sometimes some people might want to increase the budget for the next period by 20%, 50% etc. so they would use t_mul=1.2, t_mul=1.5

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: We changed the name of 't_0' to 'initial_period_steps' . We removed default value for it, it depends on the specific application and we think it's better if user provides it. We did not change names of other variables, those keep the same function as in the original implementation.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay.py
PatrykChrabaszcz: We provided one example that shows how learning rate changes between steps. It's a bit longer than your examples but should be easy to follow.

tensorflow/contrib/training/python/training/sgdr_learning_rate_decay_test.py
PatrykChrabaszcz: I fixed tests to use placeholder. It looks better I think :)

rmlarsen:
@vrv could you take another look, please?

vrv:
Yes, sorry, looking now.  GitHub doesn't notify me when people push changes. :(

vrv (approved):
Approval, assuming tests pass.

vrv:

martin-gorner:
This new function needs to be referenced in the documentation here:
https://www.tensorflow.org/api_guides/python/train#Decaying_the_learning_rate

Only tf.train.cosine_decay is referenced there. cosine_decay just decreases the learning rate town to its minimum. It does not restart.

By the way the documentation of cosine_decay should be corrected too. It mentions the SGDR article "with warm restarts" but it never states that it is NOT the cosine decay with restarts mentioned in the article.

loshchil:
@martin-gorner To my best understanding, the original function mentioned here was removed in https://github.com/tensorflow/tensorflow/pull/14500/commits/65a11ef710083c106fb5479145b3b8e133f79b1d after @PatrykChrabaszcz introduced (please see https://github.com/tensorflow/tensorflow/pull/14500 ) cosine_decay function and cosine_decay_restarts which implements cosine annealing with restarts.

The paper is referenced correctly because this is where cosine decay was proposed. Cosine decay  does not automatically assume restarts, it defines how learning rate decays / the shape of its decrease. SGDR represents SGD with restarts employing cosine annealing for each of its cycles by default (SGDR will remain SGDR if you replace cosine annealing for each cycle with an alternative decay scheme such as  piecewise_constant).  

@PatrykChrabaszcz could you please double-check my assessment and (if you agree with it) add cosine_decay_restarts to https://www.tensorflow.org/api_guides/python/train#Decaying_the_learning_rate
once you have some spare time?

PatrykChrabaszcz:
Yes, **sgdr_decay** from the contrib package is not available anymore. 
Instead, one can use **cosine_decay_restarts** function, which is implemented here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/learning_rate_decay.py#L600

The documentation is available here:
https://www.tensorflow.org/api_docs/python/tf/train/cosine_decay_restarts

I do not know why the link was not added to the api_guides section

@martin-gorner Do you want us to fix this? If it is the case then can you point me to the files which are used to generate api_guides part of the documentation?

martin-gorner:
I do not know which files are used. @martinwicke should know.
Where are the files that generate this list in the documentation:
https://www.tensorflow.org/api_guides/python/train#Decaying_the_learning_rate
The function cosine_decay_restarts should be added to the list.

martinwicke:
Ah. @MarkDaoust knows for sure, I believe the file is here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/api_guides/python/train.md

PR #11773, opened 2017-07-26
Enable Android lib bazel build for mips and mips64

resec:
Currently the Android .so lib bazel build is failing for `--cpu=mips` and `--cpu=mips64`. This PR is fixing this.

The changes are:

- Add "-latomic" linkopt for `--cpu=mips`,  this is to avoid failing at the end for `undefined reference to '__atomic_*'` alike errors.
- Remove `-Os` optimization flag for `--cpu=mips` and `--cpu=mips64`, this is to avoid failing in compiling `tensorflow/core/lib/core/threadpool.o` for `failure memory model cannot be stronger than success memory model for '__atomic_compare_exchange'` errors.
- Added some hints in `tensorflow/workspace.bzl` for API level selection when compiling 64-bit lib.

By checking online, seems there is a bug in gcc compiler (at least the gcc inside NDK r12b, I did try using newer NDK version, however since r13b, the default compiler is changed to clang, and bazel is only working with clang with NDK r13b or higher, as well as currently, compiling Tensorflow with clang is failing for some strange errors to me, so it did not work), that in some cases, `-Os` flag would crash the compiling. Removing `-Os` for `--cpu=mips` and `--cpu=mips64` is only a sub-optimal solution as the generated .so lib will be relatively bigger (but not much).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -32,6 +32,9 @@ load("//tensorflow:workspace.bzl", "tf_workspace")
 #    name="androidndk",
 #    path="<PATH_TO_NDK>",
 #    # This needs to be 14 or higher to compile TensorFlow.
+#    # Please specify API level to >= 21 to build for 64-bit
+#    # archtectures or the Android NDK will automatically select biggest
+#    # API level that it supports without notice.
 #    # Note that the NDK version is not the API level.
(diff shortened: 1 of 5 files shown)


vrv:

PR #11779, opened 2017-07-26
Fix GradientDescentOptimizer argument

clemens-tolboom:
Both the text (line 370) and python code example example code https://github.com/tensorflow/tensorflow/blob/r1.2/tensorflow/examples/tutorials/mnist/mnist_softmax.py#L59 uses 0.5. 

Using 0.05 keeps success percentage around 90% instead of the mentioned 92%


clemens-tolboom:
I checked the wrong version branch ... this is fixed in master but not yet deployed by 8c868fa

PR #11768, opened 2017-07-26
Upgrade gRPC

jhseu:
Just Jenkins testing for now.

PR #11801, opened 2017-07-27
rename target protobuf, closes #11599

### LINKED ISSUE #11599: Would you mind not calling the protobuf repository protobuf?

http://stackoverflow.com/questions/tagged/tensorflow

------------------------

### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:

Yes

Ubuntu 16.04

Source

1.1.0

- **Python version**: 

2.7.12

0.5.2
- **CUDA/cuDNN version**:
- **GPU model and memory**:

### Describe the problem

Currently the protobuf repo in workspace.bzl is called @protobuf. Unfortunately, in github.com/grpc/grpc, the protobuf is bind to @com_github_google_protobuf//:protobuf....So if I have anything that uses the GRPC repository it cannot be built together with Bazel.

I tried rename all @protobuf// to @com_github_google_protobuf//, but it does not work, presumably because the patch file for protobuf get in the way.

### Source code / logs

lakshayg:


vrv:
This looks like it's already done in the tested and pending merge of https://github.com/tensorflow/tensorflow/pull/11768 during the upgrade of gRPC -- would you mind if we take that one instead?  It also requires bazel cleaning for our test infra, so that one is probably more complete.  Sorry we did not know of the work that was in parallel :(

PR #11811, opened 2017-07-27
fix memory leak, and remove duplicated implements

horance-liu:
fix memory leak, and remove duplicated implements.


vrv:
The previous code was added by https://github.com/tensorflow/tensorflow/commit/00b79b47939beda85c22d5f603b2904596c624d2 with the explicit purpose of reducing code size, so you would have to verify that the benchmark and the code size are not affected by this change.  Can you verify this?  Thanks!

horance-liu:
How should I run `status_test.cc` alone and execute benchmark? Please give me some advise. 

In addition, I want to run all test cases or dependent test cases every time before submit the code, but the compile time is too long, Do you have any good suggestions?

vrv:
I believe if you do

```
bazel build -c opt tensorflow/core:status_test  (or maybe it's core:lib_core_status_test)
```

then you can run:

```
bazel-bin/tensorflow/core/status_test --benchmarks=all 
```

to run the benchmark, and you can compare before and after your change to see the difference in performance.

My belief is that it is very intentional to avoid constructing empty string objects, and to instead leak memory, since the program is about to crash anyway.  So I'm inclined to want to leave the code as is, unless we can prove that constructing the empty strings is not harmful.

As for all test cases: it's hard to test all dependencies without recompiling everything :(.  That's why our test infrastructure can be useful.

horance-liu:
OK, I give up this commit, and thanks for you help. @vrv

PR #11814, opened 2017-07-27
Fix error with default python path selection

DavidNorman:
When the env var USE_DEFAULT_PYTHON_LIB_PATH is set to 1, and there isn't PYTHON_LIB_PATH set, then the configure used to select the first entry that python itself returned.

this was broken recently. I suspect that it was a mistake rather than a deliberate choice.


vrv:
I made one small change to move the setting of the env var outside of the if/else so the contract of the if/else is more clear. 

DavidNorman:
the MacOS CPU test is an out of disk space problem on the build machine by the looks of things.

PR #11819, opened 2017-07-27
Cmake support snappy

alex-petrenko:
Windows CMake builds do not include SNAPPY compression library support, thus the resulting applications are not able to parse TensorFlow models that were compressed with SNAPPY.
This pull request adds SNAPPY as ExternalProject, just as it's done with all other 3rd party libraries.

Unfortunately I cannot verify this on a Linux machine right now, so option is disabled for Linux by default.

There is a relevant Stackoverflow question: https://stackoverflow.com/questions/45304986/cant-read-saved-tensorflow-model-failed-to-seek-to-header-entry

This is my first contribution, so I apologize in advance for any guidelines I might have violated. Feedback is appreciated 👍


alex-petrenko:
Seems like I used different emails for my commits and for the Contributor License Agreement. Let me close this pull request and create another one, it'll be easier.

PR #11820, opened 2017-07-27
Added SNAPPY support in CMake scripts

alex-petrenko:
Windows CMake builds do not include SNAPPY compression library support, thus the resulting applications are not able to parse TensorFlow models that were compressed with SNAPPY.
This pull request adds SNAPPY as ExternalProject, just as it's done with all other 3rd party libraries.

Unfortunately I cannot verify this on a Linux machine right now, so option is disabled for Linux by default.

There is a relevant Stackoverflow question: https://stackoverflow.com/questions/45304986/cant-read-saved-tensorflow-model-failed-to-seek-to-header-entry

Example command line to build snappy as a part of TF:
```
cmake .. -G"Visual Studio 15 2017 Win64" -Dtensorflow_BUILD_PYTHON_BINDINGS=OFF -Dtensorflow_ENABLE_GRPC_SUPPORT=OFF -Dtensorflow_BUILD_SHARED_LIB=ON -Dtensorflow_ENABLE_SNAPPY_SUPPORT=ON
MSBuild /p:PreferredToolArchitecture=x64 /p:Configuration=Release snappy.vcxproj
```

This is my first contribution, so I apologize in advance for any guidelines I might have violated. Feedback is appreciated 👍


rmlarsen:
@saxenasaurabh Saurabh, could you take a look at this, please?

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(tensorflow_BUILD_PYTHON_TESTS "Build python unit tests " OFF)
 option(tensorflow_BUILD_SHARED_LIB "Build TensorFlow as a shared library" OFF)
 option(tensorflow_OPTIMIZE_FOR_NATIVE_ARCH "Enable compiler optimizations for the native processor architecture (if available)" ON)
 option(tensorflow_WIN_CPU_SIMD_OPTIONS "Enables CPU SIMD instructions")
+option(tensorflow_ENABLE_SNAPPY_SUPPORT "Enable SNAPPY compression support" OFF)
```
mrry: Can you set this option to default `ON`? It seems reasonable to me that we should build in Snappy support as standard, just like the Bazel build does :).

tensorflow/contrib/cmake/external/snappy.cmake
```diff
@@ -0,0 +1,47 @@
...
+include (ExternalProject)
+
+set(snappy_URL https://github.com/google/snappy.git)
+set(snappy_TAG "1.1.6")
+set(snappy_BUILD ${CMAKE_CURRENT_BINARY_DIR}/snappy/src/snappy)
+set(snappy_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/snappy/src/snappy)
+
+if(WIN32)
+    set(snappy_STATIC_LIBRARIES ${snappy_BUILD}/$(Configuration)/snappy.lib)
+endif()
```
mrry: It would be nice if we could make this work on Linux as well... at a guess, would 

```cmake
else()
    set(snappy_STATIC_LIBRARIES ${snappy_BUILD}/libsnappy.a)
endif()
```

...work?

mrry (changes requested):
Thanks for this contribution... just a couple of comments!

alex-petrenko:
Google bot, please test CLA once again.

mrry:

mrry (approved):

tensorflow/contrib/cmake/CMakeLists.txt
alex-petrenko: Ok, enabled by default.

tensorflow/contrib/cmake/external/snappy.cmake
alex-petrenko: Apparently, SNAPPY CMake build does not offer static library option, so I used the shared library instead. Should work, I guess.
I could not verify it completely on Linux because in the latest master CMake build fails somewhere in the middle for a completely unrelated reason. But the change should be alright.

mrry: If you set `BUILD_SHARED_LIBS=OFF` in the Snappy CMake build, does it not build static libraries? For distributing binaries, we'd strongly prefer a static library, because otherwise you end up with a further source of DLL path hell (see for example the failing Windows tests, which I'd guess are failing because they can't find the Snappy DLL).

alex-petrenko: I used the same version of SNAPPY that was used in Bazel builds - 1.1.6. In that version Snappy's CMakeLists.txt explicitly builds shared library:

```
# Define the main library.
ADD_LIBRARY(snappy SHARED
```

After 1.1.6 release in master branch this behavior was changed:
`add_library(snappy "")`

I suppose it'd be right to use the recent commit on master instead of 1.1.6 tag and build Snappy as a static library. What do you think?

mrry: Thanks for looking into that! I'd be in favor of using the newer version of Snappy, and building as a static library.

jhseu:
Jenkins, test this please

jhseu:
This breaks the windows cmake build. Mind taking a look at the failure?

alex-petrenko:
Changed a few things as per discussion above with @mrry

alex-petrenko:
Jenkins, test this please (can I do that? :) )

mrry:
Looks good to me!

alex-petrenko:
@mrry Looks like the tests were terminated abruptly. I hope it's a glitch of a build system?

mrry:
Looks like it, yes. Annoyingly, it seemed to be doing well on the Windows build. Let's try again!

mrry:

PR #11828, opened 2017-07-27
Add missing grpc dependency to include the proto_util.h header

jhseu:


jhseu:
@nlopezgi

vrv:
windows cmake test looks a bit concerning, seems related to protobuf change, but is that related?

jhseu:
Doesn't seem like it could be related since it's a different build system.

Clicking the Rebuild button in Jenkins to try again.

vrv (approved):

jhseu:
Looks like Windows passes now. Merging this to fix the issue.

PR #11830, opened 2017-07-27
Fixes #11829: Faster to import tensorflow.contrib on Python 3

### LINKED ISSUE #11829: Slow to import tensorflow.contrib with Python 3 because inspect.stack is slow
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Sierra 10.12.5
- **TensorFlow version (use command below)**: v1.2.0-2420-g2b4a0f9a4 1.3.0-rc0
- **Python version**: 3.6.2
- **Bazel version (if compiling from source)**: 0.5.2-homebrew
- **CUDA/cuDNN version**: CPU only build
- **GPU model and memory**: CPU only build
- **Exact command to reproduce**: time python3 -c "import tensorflow.contrib"

### The problem
Doing `import tensorflow.contrib` take 7.5 seconds on my machine when doing it with Python 3.6.2. With Python 2.7.13 it takes 3.2 seconds.

Investigating this revealed that a lot of time is spent in `_inspect.stack()` in the function `make_decorator` in  `python/util/tf_decorator.py`. The stack is inspected to find the name of the caller of the function. With Python2 `inspect.stack()` is fast, but with Python 3 each call to `inspect.stack()` take approximately 0.2 seconds and there are 23 calls made, which account for the difference in time between Python 2 and 3.

### References
Keras by default imports tensorflow.contrib when the Tensorflow backend is used. Therefore Keras is slow to import when using Python 3: https://github.com/fchollet/keras/issues/7408

There is a stackoverflow question referencing this issue: https://stackoverflow.com/questions/45093653/import-tensorflow-contrib-module-is-slow-in-tensorflow-1-2-1

patrikerdes:
As described in #11829, a lot of time is spent doing `inspect.stack()` when importing `import tensorflow.contrib` with Python 3. (There does not seem to be any issue with Python 2.)

By replacing `_inspect.stack()` with only getting the current and previous frame and then getting the frame info for just the previous frame, this unnecessary overhead is removed.

The time to import tensorflow.contrib, as reported by `time python2/3 -c "import tensorflow.contrib"` before and after my commit:

When|Python 2.7.13|Python 3.6.2|
------|-------------|-------------|
before|3.2|7.5|
after|3.1|3.3|

There are unit tests that would catch if the `decorator_name` would be incorrect after my change, specifically these tests:
* testSetsDecoratorNameToFunctionThatCallsMakeDecoratorIfAbsent
* testUnwrapReturnsDecoratorListFromOutermostToInnermost
* testUnwrapBoundMethods

All Python unit tests passed (in Docker). I didn't find any issues when linting the changed file.

### DIFF
--- a/tensorflow/python/util/tf_decorator.py
+++ b/tensorflow/python/util/tf_decorator.py
@@ -83,7 +83,8 @@ def make_decorator(target,
     The `decorator_func` argument with new metadata attached.
   """
   if decorator_name is None:
-    decorator_name = _inspect.stack()[1][3]  # Caller's name.
+    prev_frame = _inspect.currentframe().f_back
+    decorator_name = _inspect.getframeinfo(prev_frame)[2]  # Caller's name.
   decorator = TFDecorator(decorator_name, target, decorator_doc,
                           decorator_argspec)
   setattr(decorator_func, '_tf_decorator', decorator)


charlesnicholson:
I'm not an admin but I wrote the original code. When I wrote it, I saw both my approach and what @patrikerdes, but chose mine because it was simpler and not (that I observed!) a performance problem.

If this patch fixes real performance problems, and the tests pass, we should prefer it. 

Thanks for the patch!

benoitsteiner:
@jart, can you take a look since you're also working on speeding up the tensorflow imports.

rmlarsen:
@jart would you have time to look at this?

patrikerdes:
This is still a problem in tensorflow 1.3.0. Could someone look at this patch? If more information is needed, I would be happy to supply it.

drpngx:
@charlesnicholson could you take a look? That seems quite a bit faster.

charlesnicholson:
@drpngx I'm not authorized to merge the request, but I gave it my whole-hearted approval on July 27. @jart seems to not be writing back, so if she's busy or uninterested I'd suggest finding another approver: there's really no reason to leave this unmerged.

drpngx:
Thanks @charlesnicholson , we can take it from here!

Jenkins, test this please.

drpngx (approved):
As per Charles' comment.

PR #11834, opened 2017-07-27
Ensure that TensorBoard is still available when pip installed. (r1.3)

teamdandelion:
Context:

After r1.2, TensorBoard moved out of the TensorFlow repository, into its own
repository and its own pip package (presently tensorflow-tensorboard, will later
switch to just tensorboard). The new pip package specifies the `tensorboard`
command, so I removed it from the list of console scripts forTensorFlow. I also
added tensorflow-tensorboard as a pip dependency.

However, it turns out that the pip order of operations is:
- install pip dependencies (thus getting tensorflow-tensorboard and the new
  tensorboard command)
- remove deprecated console scripts (thus erroneously removing the new pointer
  to tensorboard)

To fix this, I returned the `tensorboard` console script to tensorflow's
setup.py, except it now references the tensorboard package rather than the
tensorflow package. Thus, the console script declaration in tensorflow and
tensorboard are identical. We can be confident that the tensorboard package is
available, because it is specified by the pip dependency.

Test Plan:

- Create a clean virtualenv.
- pip install tensorflow < 1.3.
- verify that the tensorboard command works properly
- pip install tensorflow 1.3 using a pip package generated with this change
- verify that the tensorboard command still works

This is the fix for r1.3. I will also send this same commit in a PR against master.


tensorflow/tools/pip_package/setup.py
```diff
@@ -60,6 +60,11 @@
 # pylint: disable=line-too-long
 CONSOLE_SCRIPTS = [
     'saved_model_cli = tensorflow.python.tools.saved_model_cli:main',
+    # We need to keep the TensorBoard command, even though the console script
+    # is now declared by the tensorboard pip package. If we remove the
```
wchargin: nit: I'd write either "the TensorBoard pip package" or "the tensorflow-tensorboard pip package" (because `tensorflow-tensorboard` is the pip package for the TensorBoard application).

wchargin:
LGTM, but I'd wait for a Real TensorFlow Reviewer.

MarkDaoust (approved):
I don't know much about the internal workings of setup.py

But I have also build this branch and have confirmed that this fixes the problem.

vrv:
You have to update unfortunately.

vrv:
Amit for 1.3 merge

av8ramit (approved):

Issue #11384, opened 2017-07-08
Upgrading to 1.2 Causes Issue with CUDA


RylanSchaeffer:
I don't know how to describe my problem any better than that, unfortunately. Previously, I was running TensorFlow 1.0, and I could successfully use CUDA:

```
ubuntu@ip-10-0-2-117:~$ /usr/bin/python2.7 /home/ubuntu/src/tensorflow/tensorflow/examples/tutorials/mnist/fully_connected_feed.py
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcublas.so.7.5 locally
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcudnn.so.5 locally
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcufft.so.7.5 locally
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcuda.so.1 locally
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcurand.so.7.5 locally
Successfully downloaded train-images-idx3-ubyte.gz 9912422 bytes.
Extracting /tmp/tensorflow/mnist/input_data/train-images-idx3-ubyte.gz
Successfully downloaded train-labels-idx1-ubyte.gz 28881 bytes.
Extracting /tmp/tensorflow/mnist/input_data/train-labels-idx1-ubyte.gz
Successfully downloaded t10k-images-idx3-ubyte.gz 1648877 bytes.
Extracting /tmp/tensorflow/mnist/input_data/t10k-images-idx3-ubyte.gz
Successfully downloaded t10k-labels-idx1-ubyte.gz 4542 bytes.
Extracting /tmp/tensorflow/mnist/input_data/t10k-labels-idx1-ubyte.gz
I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:910] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero
I tensorflow/core/common_runtime/gpu/gpu_device.cc:885] Found device 0 with properties: 
name: Tesla K80
major: 3 minor: 7 memoryClockRate (GHz) 0.8235
pciBusID 0000:00:1e.0
Total memory: 11.17GiB
Free memory: 11.11GiB
I tensorflow/core/common_runtime/gpu/gpu_device.cc:906] DMA: 0 
[...]

PR #11835, opened 2017-07-27
Ensure that TensorBoard is still available when pip installed.

teamdandelion:
Context:

After r1.2, TensorBoard moved out of the TensorFlow repository, into its own
repository and its own pip package (presently tensorflow-tensorboard, will later
switch to just tensorboard). The new pip package specifies the `tensorboard`
command, so I removed it from the list of console scripts forTensorFlow. I also
added tensorflow-tensorboard as a pip dependency.

However, it turns out that the pip order of operations is:
- install pip dependencies (thus getting tensorflow-tensorboard and the new
  tensorboard command)
- remove deprecated console scripts (thus erroneously removing the new pointer
  to tensorboard)

To fix this, I returned the `tensorboard` console script to tensorflow's
setup.py, except it now references the tensorboard package rather than the
tensorflow package. Thus, the console script declaration in tensorflow and
tensorboard are identical. We can be confident that the tensorboard package is
available, because it is specified by the pip dependency.

Test Plan:

- Create a clean virtualenv.
- pip install tensorflow < 1.3.
- verify that the tensorboard command works properly
- pip install tensorflow 1.3 using a pip package generated with this change
- verify that the tensorboard command still works

This is the fix for master. I sent a separate PR to move this fix into r1.3

### DIFF
--- a/tensorflow/tools/pip_package/setup.py
+++ b/tensorflow/tools/pip_package/setup.py
@@ -60,6 +60,11 @@ if sys.version_info < (3, 4):
 # pylint: disable=line-too-long
 CONSOLE_SCRIPTS = [
     'saved_model_cli = tensorflow.python.tools.saved_model_cli:main',
+    # We need to keep the TensorBoard command, even though the console script
+    # is now declared by the tensorboard pip package. If we remove the
+    # TensorBoard command, pip will inappropriately remove it during install,
+    # even though the command is not removed, just moved to a different wheel.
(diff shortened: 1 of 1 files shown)


wchargin:
(Same comment as on #11384.)

MarkDaoust (approved):
I don't know much about the internal workings of `setup.py`

But I have build this branch and have confirmed that this fixes the problem.

vrv:

PR #11849, opened 2017-07-28
[XLA] RNG root instruction can be fused as it effectively has one user

DavidNorman:
The existing test checks for a single user when trying to fuse a random number generator instruction.   This is because if it is fused but has multiple users, the value may be generated an incorrect number of times.  (multiple users should receive the same random number, not different ones).

The test checks for users() length == 1.  This ignores the case there the instruction is the root, where users() length == 0, but there is effectively one user.

### DIFF
--- a/tensorflow/compiler/xla/service/hlo_instruction.cc
+++ b/tensorflow/compiler/xla/service/hlo_instruction.cc
@@ -1759,7 +1759,7 @@ bool HloInstruction::IsFusable() const {
     // Only fuse Rng if it is used once, otherwise the random numbers generated
     // will be different in each fusion.
     case HloOpcode::kRng:
-      return users_.size() == 1;
+      return users_.size() <= 1;
     default:
       return true;
   }
--- a/tensorflow/compiler/xla/service/hlo_instruction.cc
+++ b/tensorflow/compiler/xla/service/hlo_instruction.cc
@@ -1757,7 +1757,8 @@ bool HloInstruction::IsFusable() const {
     case HloOpcode::kRecv:
       return false;
     // Only fuse Rng if it is used once, otherwise the random numbers generated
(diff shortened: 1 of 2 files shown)


blakehechtman (changes requested):
I do not think this will break anything, but it it would be nice to see a test where Rng is both the root of a computation and is fused to make sure nothing breaks.

DavidNorman:
@blakehechtman done.

i tried adding a death test for the 2+ case, but the testing framework complains that death tests are unstable with multiple threads.

blakehechtman (approved):

DavidNorman:
cheers :)

DavidNorman:
@vrv @blakehechtman 

I think you need to do the 'test this' command to get the tests to run.

vrv:

DavidNorman:
Thanks.  The failure in the windows build is this:

```
10:23:57 error: cannot lock ref 'refs/remotes/origin/pr/5992/head': Unable to create 'c:/tf_jenkins/home/workspace/tensorflow-pr-win-cmake-py/.git/refs/remotes/origin/pr/5992/head.lock': File exists.
```

Think it is a different problem - maybe even a CI system issue.

vrv:
Yup, seems like a flaky failure that can be ignored!  This is ready to merge (I'm not on merge rotation this week so I don't want to screw anything up)

DavidNorman:
:)  sure thing

DavidNorman:
@blakehechtman @vrv 

Hi gues.  Do you think we could get this merged?  I have quite a few outstanding merge requests and it is making me nervous that our repo is drifting away from the public one.

vrv:
ping @rmlarsen

rmlarsen:

PR #11852, opened 2017-07-28
fixed tf.pow edge case

galvk:
The tf.pow() function has an edge case which causes it to hang with no error message.

If you try to evaluate tf.pow(x,y), when x is an integer (and thus the output tensor is also an integer), while y is a negative value, tensorflow hangs trying to cast the fraction as an integer.

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -162,6 +162,8 @@ from tensorflow.python.ops import gen_sparse_ops
 from tensorflow.python.ops import gen_spectral_ops
 from tensorflow.python.ops import gen_state_ops
 from tensorflow.python.ops import state_ops
+from tensorflow.python.ops.gen_array_ops import identity
+from tensorflow.python.ops.check_ops import assert_greater_equal
 # go/tf-wildcard-import
 # pylint: disable=wildcard-import
 from tensorflow.python.ops.gen_math_ops import *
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/math_ops.py
```diff
@@ -541,6 +541,9 @@ def pow(x, y, name=None):
     A `Tensor`.
   """
   with ops.name_scope(name, "Pow", [x]) as name:
+    if 'int' in str(x.dtype):
+      tf.assert_negative(y,
```
aselle: don't you mean tf.assert_positive here?
also, I don't think this strategy will work. You need to make sure that there is not any negative exponents for example.
for example this will hang but doesn't have a consistent tf.assert_positive() or negative()

I'd start by writing a set of unit tests you want to prevent from haing i.e.
```
sess.run(tf.pow([5,2],[-2,3]))
sess.run(tf.pow([5,2],[2,-3]))
sess.run(tf.pow([5,2],-3))
```

aselle (changes requested):

vrv:
We also need the CLA signed if we are to accept this PR, thanks!

galvk:
I have already signed the CLA.

vrv:
Looks like you used two different commit emails -- can you use the one you used for the CLA and squash your commits ?

galvk (commented):
I've edited the code so that a negative y (be it an int, float, or list) casts x to a float, this way the operation works, instead of throwing an error. However, I'm not sure this is possible when x and y are tensors. So I tried to force an error when x is an integer tensor, combo'd with assert_greater_equal(y, 0). This doesn't work though, so any help would be much appreciated.

The last case to prevent from hanging;

tf.pow(tf.constant(2), tf.constant(-2))

rmlarsen:
@galvk Thanks for the contribution, but I don't think introducing this level of complexity in the Python wrapper is the right solution. We should fix the underlying C++ functor. I'd recommend opening an issue instead.

jhseu:
Closing this out due to @rmlarsen 's comment. Feel free to reopen with a fix in C++ or file an issue instead. Thanks!

PR #11858, opened 2017-07-28
Bug in builds/pip.sh causes failure if NO_TEST_ON_INSTALL=1

sumitgouthaman:
do_virtualenv_pip_test() uses ${CLEAN_VENV_DIR} which is defined in
do_clean_virtualenv_smoke_test(). But if NO_TEST_ON_INSTALL is set to 1
${CLEAN_VENV_DIR} would never be defined, and pip.sh will fail.

It is very likely that we actually meant to use ${VENV_DIR} instead.


caisq (approved):

caisq:

caisq:
Thanks for the fix, @sumitgouthaman

vrv:

PR #11861, opened 2017-07-28
grpc changed to only build Release flavor

guschmue:
cmake builds are broken for Debug/RelWithDebInfo builds because grpc now only builds Release binaries. Always point to the Release bits to fix it.

### DIFF
--- a/tensorflow/contrib/cmake/external/grpc.cmake
+++ b/tensorflow/contrib/cmake/external/grpc.cmake
@@ -21,9 +21,9 @@ set(GRPC_TAG 781fd6f6ea03645a520cd5c675da67ab61f87e4b)
 
 if(WIN32)
   set(grpc_STATIC_LIBRARIES
-      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/$(Configuration)/grpc++_unsecure.lib
-      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/$(Configuration)/grpc_unsecure.lib
-      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/$(Configuration)/gpr.lib)
+      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/Release/grpc++_unsecure.lib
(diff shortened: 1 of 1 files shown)


vrv:

Thanks!

tensorflow/contrib/cmake/external/grpc.cmake
```diff
@@ -21,9 +21,9 @@ set(GRPC_TAG 781fd6f6ea03645a520cd5c675da67ab61f87e4b)
 
 if(WIN32)
   set(grpc_STATIC_LIBRARIES
-      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/$(Configuration)/grpc++_unsecure.lib
-      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/$(Configuration)/grpc_unsecure.lib
-      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/$(Configuration)/gpr.lib)
+      ${CMAKE_CURRENT_BINARY_DIR}/grpc/src/grpc/Release/grpc++_unsecure.lib
```
jhseu: Would it make more sense to use $(Configuration) below for the BUILD_COMMAND instead? I'm not as familiar with cmake, so I don't know the best practices here.

guschmue:
I guess the change that broke it is:
BUILD_COMMAND ${CMAKE_COMMAND} --build . --config **Release** --target grpc++_unsecure
and we could use  $(Configuration) .
There is also -DCMAKE_BUILD_TYPE:STRING=Release passed to the grpc cmake but assume that is never looked at,, else this this would have been broken in past.
Either way is fine with me - your pick.
@mrry  - any preference ?

A little upside for rpc Release builds only: for gpu builds there is an issue that the total size of the symbols info of the python_wrapper.dll is > 4GB which hits some limit in link.exe (it fails) ... less symbols is a good thing.

jhseu (approved):
@mrry will be away for a bit. Approving for now and we can improve later if needed.

guschmue:
perfect - Thanks!

PR #11878, opened 2017-07-30
SVD-operation on the GPU

shamanDevel:
This merge request implements the SVD on the GPU using cuSolver.

We are building a sparse Gaussian Mixture Model using Tensorflow and for that we need the SVD and determinant computation (see another merge request). Since these operations were not implemented on the GPU yet and the communication resulted in a large overhead, I implemented them on the GPU.

The SVD is implemented using cuSolver. 
cuSolver, however, has the downside of only supporting matrices with m>=n. The CPU-Version of SVD supports also matrices with m<n. So this kernel throws an error when matrices with m<n are passed and notifies the user to explicitly switch to the CPU kernel.


tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+    const int ndims = input.dims();
+    const int64 m = input.dim_size(ndims - 2);
+    const int64 n = input.dim_size(ndims - 1);
+    const int64 p = std::min(m, n);
+
+    // This is a limitation of cuSolver
+    OP_REQUIRES_ASYNC(
+        context, m >= n,
+        errors::InvalidArgument(
+            "GPU implementation of SVD only supports m>=n, but m=", m, ", n=",
```
rmlarsen: Could you support all cases by transposing the input and swapping and transposing U and V on output, as needed?

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+    if (n == 0 || m == 0) {
+      // If X is an empty matrix (0 rows, 0 col), X * X' == X.
+      // Therefore, we return X.
+      done();
+      return;
+    }
+
+    // Copy and reshape input tensor
+    // SVD modifies the input, so I need to copy it.
+    Tensor inputCopy;
```
rmlarsen: Please use lower case variable names instead of camel case.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+    OP_REQUIRES_OK_ASYNC(
+        context,
+        context->allocate_temp(input.dtype(), input.shape(), &inputCopy), done);
```
rmlarsen: You could use buffer forwarding here, like so:
```
    // Reuse the input buffer or make a copy for the SVD depending on whether this ops owns the 
    // input buffer exclusively.
    Tensor input_copy;
    OP_REQUIRES_OK_ASYNC(
        context,
        context->forward_input_or_allocate_temp(
            {0}, DataTypeToEnum<Scalar>::value, input.shape(), &input_copy),
        done);

    if (!input.SharesBufferWith(input_copy)) {
      const GPUDevice& d = context->eigen_device<GPUDevice>();
      d.memcpy(input_copy.flat<Scalar>().data(), input.flat<Scalar>().data(),
               input.NumElements() * sizeof(Scalar));
    }
```

Please do not call cudaMemcpy directly as it synchronizes the stream.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+    Scalar* outputU_reshaped_ptr = NULL;
+    Scalar* outputVT_reshaped_ptr = NULL;
+    if (compute_uv_) {
+      auto outputU_reshaped = outputU->template flat_inner_dims<Scalar, 3>();
+      auto outputVT_reshaped = outputVT.template flat_inner_dims<Scalar, 3>();
+      outputU_reshaped_ptr = outputU_reshaped.data();
+      outputVT_reshaped_ptr = outputVT_reshaped.data();
+    }
+
+    // Launch a SVD kernel for each matrix in the batch.
```
rmlarsen: nit: s/n/an/

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+      int* dev_info_ptr = dev_info.mutable_data() + i;
+      OP_REQUIRES_OK_ASYNC(
+          context,
+          solver.Gesvd(jobu, jobvt, m, n, input_ptr, lda, outputS_ptr,
+                       outputU_ptr, ldu, outputVT_ptr, ldvt, dev_info_ptr),
+          done);
+    }
+
+    // Test if it was successfull.
+    // I'm not using solver.CopyLapackInfoToHostAsync
```
rmlarsen: You need to use solver.CopyLapackInfoToHostAsync. If you the launch the followup operation inside the closure after checking info, you should be OK. Make sure to capture everything needed in the closure so it does not get freed prematurely when the asynchronous kernel returns from CompureAsync.

The idea it to not use synchronous cudaMemcpy, which blocks other kernels in the stream ready to be run.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+        status = errors::InvalidArgument(
+            "Got info = ", info_value, " for batch index ", i,
+            ", expected info = 0. Debug_info =", host_info.debug_info());
+      }
+    }
+    if (!status.ok()) {
+      status.Update(errors::InvalidArgument(kErrMsg));
+    }
+    OP_REQUIRES_OK_ASYNC(context, status, done);
+    // TODO: Maybe switch to solver.CopyLapackInfoToHostAsync again
```
rmlarsen: Sort of, calling DoTranspose inside the closure is the trick you are looking for.

tensorflow/core/kernels/cuda_solvers.cc
```diff
@@ -202,6 +202,7 @@ Status CudaSolver::CopyLapackInfoToHostAsync(
 // numeric types.
 #define TF_CALL_LAPACK_TYPES(m) \
   m(float, S) m(double, D) m(std::complex<float>, C) m(std::complex<double>, Z)
+#define TF_CALL_LAPACK_TYPES_NO_COMPLEX(m) m(float, S) m(double, D)
```
rmlarsen: Can you add support for complex types, while we're at it?

rmlarsen (commented):
Thanks a lot for this contribution!

rmlarsen (changes requested):
I hit the wrong button :-) Please address my previous comments.

tensorflow/core/kernels/svd_op_gpu.cu.cc
rmlarsen: Alternatively, you can just launch the SVD and the subsequent transpose kernel opportunistically up front, and just use solver.CopyLapackInfoToHostAsync to check for the status at the end. The SVD and transpose kernels will get serialized on the stream.

rmlarsen (changes requested):

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+#include "tensorflow/core/platform/logging.h"
+#include "tensorflow/core/platform/types.h"
+
+#include "tensorflow/core/kernels/cuda_solvers.h"
+#include "tensorflow/core/platform/stream_executor.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
+// I need to transpose V afterwards
+#include <vector>
+#include "transpose_functor.h"
```
rmlarsen: transpose_functor.h -> tensorflow/core/kernels/transpose_functor.h

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,249 @@
...
+#include "tensorflow/core/lib/core/errors.h"
+#include "tensorflow/core/platform/logging.h"
+#include "tensorflow/core/platform/types.h"
+
+#include "tensorflow/core/kernels/cuda_solvers.h"
+#include "tensorflow/core/platform/stream_executor.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
+// I need to transpose V afterwards
+#include <vector>
```
rmlarsen: Standard includes at the top, please

https://google.github.io/styleguide/cppguide.html

rmlarsen (changes requested):

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: What do you mean by that?

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: I think I'll go for the second approach

tensorflow/core/kernels/cuda_solvers.cc
shamanDevel: That's actually a bit tricky: The SVD of a complex matrix gives real singular values. The implementation of cuSolver also provides the singular values in a real vector. The tensorflow operation, however, requires the singular values to be complex values as well. So I would need to store the singular values in a temporal array and convert them to complex values for the output (I can use CastFunctor defined in cast_op.h for that).
What is the reason that Tensorflow requires the singular values to be stored in a complex tensor?

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: That's a good idea. I'll implement that.

tensorflow/core/kernels/svd_op_gpu.cu.cc
rmlarsen: I meant s/a/an/. (Unix sed syntax for replace "a" by "an".)

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -231,7 +216,22 @@ class SvdOpGpu : public AsyncOpKernel {
       DoTranspose(device, outputVT, permAS, outputV);
     }
 
-    done();
+    // Now we check if the SVD operation succeeded or not
+    auto info_checker = [context, dev_info, done](
```
rmlarsen: Also capture input_copy by value here, since otherwise it might get deallocated before the SVD actually runs, giving you memory corruption. The cost is negligible since "Tensor" is just a light-weight wrapper and copying it only copies a pointer and updates the refcount on the underlying buffer.

rmlarsen (changes requested):
@shamanDevel  Almost there.

tensorflow/core/kernels/cuda_solvers.cc
rmlarsen: There is no good reason. It was done this way because it made the code slight simpler when using Eigen. But you'll notice that the Python wrapper immediately casts it back to real again:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/linalg_ops.py#L281

We should probably fix the kernels to return real singular values in both cases. But since it requires changing the signature of the kernel, we would have to introduce a new SvdOpV2 to avoid backwards compatibility problems. Then we could get rid of the wasteful casting in the Python wrapper. I'll leave it up to you whether you are up for it in this PR. We can also do that and add complex support in a follow up.

shamanDevel:
It took me a while to find a way how to deal with all the needed matrix transpositions when also supporting matrices of shape m<n. The result included a lot of refactoring, but I think that it is now also clearer to understand.

Note: don't merge yet!
Not all python tests pass. 
First, the GPU version is numerically much less precise. Even after explicitly increasing the tolerance by a factor of 1000 when the GPU version is running, I sometimes get errors that 0.014% (or so) of the matrices don't match.
Second: The GPU version fails if one dimension of the input matrix is one. Then the SVD returns zeros for U or V instead of 1. I have to investigate that more.

rmlarsen:
@shamanDevel OK, thanks for the update. I have noticed the GPU version being less accurate on occasion, so you might have to increase the tolerance in the tests. Please ping the PR when you are ready to proceed.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,343 @@
...
+    // Transpose M, because cuSolver expects it to be column-major
+    TensorShape input_shape = shapeRaw;
+    input_shape.AddDim(n);
+    input_shape.AddDim(m);
+    Tensor input_copy;
+    OP_REQUIRES_OK_ASYNC(
+        context, context->allocate_temp(M.dtype(), input_shape, &input_copy),
+        done);
+    auto device = context->eigen_device<GPUDevice>();
+    DoTranspose(device, M, perm, &input_copy);
```
rmlarsen: You need to check the return value from DoTranpose. Wrap in OP_REQUIRES_OK macro.

rmlarsen (changes requested):

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -93,6 +103,50 @@ class SvdOpGpu : public AsyncOpKernel {
...
           done);
+
+      // This is a bug in cuSolver:
+      // If n is one, then outputVT only contains zeros instead of ones.
+      // Hence, I need to fill outputVT manually
+      if (compute_uv_ && n==1) {
+        int64 count =  n * (full_matrices_ ? n : p);
+        const GPUDevice& d = context->eigen_device<GPUDevice>();
+        CudaLaunchConfig cfg = GetCudaLaunchConfig(count, d);
+        FillWithOnesKernel <<<cfg.block_count,
```
rmlarsen: Vt could also be -1. Are you sure this is always correct? (Trying it on a number of random matrices should suffice.)

rmlarsen: We already have a functor like this: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/fill_functor.h#L63

rmlarsen (changes requested):

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: No, it is not correct, it was only a first attempt. But I think I found a way how to determine the sign.
I'll implement it the next days, I was on a short holiday.

shamanDevel:
Now it should also work for matrices with shape of m==1 or n==1.
The unit tests succeed now.

Can you have a look at the changes?
I'll fix the merge conflicts to master now

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,455 @@
...
+      //printf("V'[%d] = %f\n", (int)i, (float)V[i]);
+      V[i] = (Scalar) copysign(1.0, V[i]);
+    }
+  }
+}
+
+// Scalar: The input scalar type (can be complex)
+// SScalar: The output type for the singular value,
+//   same as Scalar if real, or the real version if Scalar is complex
+template <class Scalar, class SScalar>
```
rmlarsen: Instead of adding another template parameters, you can get the real type corresponding to a complex (or real) type like this:
```
using RealScalar = typename Eigen::NumTraits<Scalar>::Real;
```

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,455 @@
...
+      int ldvt = n;
+      Scalar* input = input_ptr + i * m * n;
+      SScalar* outputS = outputS_ptr + i * p;
+      Scalar* outputU = NULL;
+      Scalar* outputVT = NULL;
+      signed char jobu = 'N';
+      signed char jobvt = 'N';
+
+      // Save input matrix for n=1 fix
+      if (compute_uv_ && n==1) {
```
rmlarsen: Since the kernel launch overhead is pretty high and the matrix is small in the n==1 case, perhaps you can just create a single temp tensor and copy the entire input outside the batch loop.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,455 @@
...
+    // Needed for the n=1 fix, see below, since SVD destroys the input
+    Tensor input_copy;
+    if (compute_uv_ && n==1) {
+      OP_REQUIRES_OK_ASYNC(
+        context,
+        context->allocate_temp(DataTypeToEnum<Scalar>::v(), TensorShape({m}), &input_copy),
+        done);
+    }
+
+    for (int64 i = 0; i < batch_size; ++i) {
```
rmlarsen: s/i/batch/

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,455 @@
...
+      // Hence, I need to fill outputVT manually
+      // The question is: +1 or -1?
+      // -> Compute U*S and compare sign against M
+      // But because S is zero except for the first entry, the multiplication simplifies a lot
+      // However, what happens if M contains zeros? At these indices, it is impossible
+      // to determine the value of V
+      // -> Compute V for all rows in M to cope for zeros.
+      // 1. V' = sum_i (M_i * U_i,1 * S_i)
+      // 2. V = {1, V'>=0, -1, V'<0}
+      // TODO: what is with complex values?
```
rmlarsen: For complex singular vectors are only unique up to a complex phase factor (rather than just sign for real). You can see how I normalize them in the unit test here: 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/svd_op_test.py#L73

so you can probably use the same idea of computing the mean (complex) product.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,455 @@
...
+        const GPUDevice& d = context->eigen_device<GPUDevice>();
+        d.memset(outputVT, 0, sizeof(Scalar));
+        CudaLaunchConfig cfg = GetCudaLaunchConfig(m, d);
+        ComputeV1Kernel <<<cfg.block_count, 
+                              cfg.thread_per_block, 0, d.stream()>>>
+            (cfg, m, input_copy.flat<Scalar>().data(), outputU, outputS, outputVT);
+      }
+      // For the second part, see below
+
+#if 0
```
rmlarsen: Get rid of cruftf.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,455 @@
...
+          DoTranspose(device, u_copy, perm, U),
+          done);
+    }
+
+    // now check if the SVD operation succeeded or not
+    CheckResult(context, done, dev_info, solver, input_copy, u_copy);
+  }
+
+  // The SVD if m < n
+  void PerformSVD_MlessN(OpKernelContext* context, DoneCallback done, int64 m,
```
rmlarsen: It would seem that you could avoid a lot of duplicated boilerplate code, e.g., by introducing variables "small_dim" and "large_dim" and combining  PerformSVD_MlessN and  PerformSVD_MgeqN.

rmlarsen (changes requested):
@shamanDevel Thanks for persevering in the face of the n==1 bug, and handling the n>m case! Here are some more comments. Mostly suggestions on how to simplify the code.

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: Oh, I was looking for exactly that but couldn't find it. Thanks.

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: Yes, I know that there is quite some duplication. But because I have to handle the matrices differently in terms of transposition each time I need this. It could have been combined into one function with if-statements on which matrix should be transposed.
There's a tradeoff between two functions for the two cases and one function for both with many if-statements for which matrix to transpose (and whether we need an additional copy of the input matrix).
I doubt that it is worth it.

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: Yes, I guess the case of a 1000000x1-matrix is pretty rare ;)
But then I could go further and directly call ComputeV1Kernel over multiple batches.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -71,7 +71,7 @@ namespace {
   __global__ void ExtractSignOfVKernel(CudaLaunchConfig config, Scalar* V)
   {
     CUDA_1D_KERNEL_LOOP(i, config.virtual_thread_count) {
-      V[i] = V[i]>=0 ? (Scalar)1 : (Scalar)0
+      V[i] = V[i]>=0 ? (Scalar)1 : (Scalar)0;
```
rmlarsen: Formatting and use of C-style casts does not agree with style guide. It should be 

V[i] = V[i]>=0 ? Scalar(1) : Scalar(0);

tensorflow/core/kernels/svd_op_gpu.cu.cc
rmlarsen: My concern is batch_size >> 1, but I suppose we can address it in a followup.

rmlarsen:

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,427 @@
...
+    Tensor input_copy;
+    if (compute_uv_ && n==1) {
+      OP_REQUIRES_OK_ASYNC(
+        context,
+        context->allocate_temp(DataTypeToEnum<Scalar>::v(), TensorShape({m}), &input_copy),
+        done);
+    }
+
+    for (int64 batch = 0; batch < batch_size; ++batch) {
+      int lda = m;
```
rmlarsen: You don't really need these temporary variables for leading dimensions, since they are passed to solver.Gesvd(...) by value. (This is not Fortran ;-) )

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,427 @@
...
+      int ldu = m;
+      int ldvt = n;
+      Scalar* input = input_ptr + batch * m * n;
+      SScalar* outputS = outputS_ptr + batch * p;
+      Scalar* outputU = NULL;
+      Scalar* outputVT = NULL;
+      signed char jobu = 'N';
```
rmlarsen: s/signed//

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,427 @@
...
+      V[i] = V[i]>=0 ? (Scalar)1 : (Scalar)0;
+    }
+  }
+}
+
+// Scalar: The input scalar type (can be complex)
+template <class Scalar>
+class SvdOpGpu : public AsyncOpKernel {
+ public:
+  using SScalar = typename Eigen::NumTraits<Scalar>::Real;
```
rmlarsen: s/SScalar/RealScalar/ This is the name used elsewhere in TF.

tensorflow/core/kernels/svd_op_gpu.cu.cc
rmlarsen: Perhaps put a TODO.

rmlarsen (changes requested):

rmlarsen:
@shamanDevel Thanks for the updates!

tensorflow/python/kernel_tests/svd_op_test.py
```diff
@@ -152,8 +157,16 @@ def Test(self):
           # dimension that is equal to one
           s_np = np.reshape(s_np, s_tf_val.shape)
 
+          #print("expected S:", s_np)
```
rmlarsen: get rid of cruft

tensorflow/python/kernel_tests/svd_op_test.py
```diff
@@ -152,8 +157,16 @@ def Test(self):
...
+          #print("actual S:", s_tf_val)
+
           CompareSingularValues(self, s_np, s_tf_val)
           if compute_uv:
+            #print("expected U:", u_np)
```
rmlarsen: get rid of cruft

tensorflow/python/kernel_tests/svd_op_test.py
```diff
@@ -165,16 +178,22 @@ def Test(self):
 
   return Test
 
+# TODO
```
rmlarsen: no longer relevant?

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,414 @@
...
+#include "tensorflow/core/platform/logging.h"
+#include "tensorflow/core/platform/stream_executor.h"
+#include "tensorflow/core/platform/types.h"
+#include "tensorflow/core/util/cuda_kernel_helper.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
+// I need to transpose V afterwards
+#include "tensorflow/core/kernels/transpose_functor.h"
+
+// Logging
```
rmlarsen: no longer needed

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,414 @@
...
+#include "tensorflow/core/kernels/linalg_ops_common.h"
+#include "tensorflow/core/lib/core/errors.h"
+#include "tensorflow/core/platform/logging.h"
+#include "tensorflow/core/platform/stream_executor.h"
+#include "tensorflow/core/platform/types.h"
+#include "tensorflow/core/util/cuda_kernel_helper.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
+// I need to transpose V afterwards
+#include "tensorflow/core/kernels/transpose_functor.h"
```
rmlarsen: Put with other includes and sort.

tensorflow/core/kernels/svd_op_gpu.cu.cc
```diff
@@ -0,0 +1,414 @@
...
+limitations under the License.
+==============================================================================*/
+
+// See docs in ../ops/linalg_ops.cc.
+// TODO(shamanDevel): Enable complex inputs. This will require additional tests
+//                    and OP_REQUIRES.
+#if GOOGLE_CUDA
+#define EIGEN_USE_GPU
+
+#include <algorithm>
```
rmlarsen: do you still need <algorithm> ?

rmlarsen (changes requested):
Looks like you need to clean up a few things left behind, otherwise I think this is good.

rmlarsen:
It looks like there are a few test failures because tol=1e-14 is too tight for double precision. You can just loosen the tolerance a little bit.

tensorflow/core/kernels/svd_op_gpu.cu.cc
shamanDevel: Do you mean EIGEN_USE_GPU or <algorithm>? I need both. EIGEN_USE_GPU for the GPU_Device, and algorithm for p=std::min(m, n)

shamanDevel:
@rmlarsen all cleanup tasks should now be implemented

martinwicke:
I think I made conflicts for you by merging a bunch of stuff -- can you resolve?

shamanDevel:
@martinwicke the conflict was a single blank line with different whitespaces ^^. The conflict is fixed now.

rmlarsen:
@shamanDevel Thanks!

rmlarsen (approved):

rmlarsen:
@shamanDevel It looks like you have a compilation error (bad merge?):

....out/local-opt/bin/tensorflow/core/kernels/_objs/svd_op_gpu/tensorflow/core/kernels/svd_op_gpu.cu.o): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
tensorflow/core/kernels/svd_op_gpu.cu.cc:413:1: error: expected declaration before '}' token
 }  // namespace tensorflow
 ^

shamanDevel:
@rmlarsen The error is fixed now.

Oh man, I learned a lot during this pull request. Like this one:
If you have a fallback in your code (here: no CUDA support), then continue testing this fallback as well!
This was what caused the error: a stupid typo that only appears if GOOGLE_CUDA is not defined.

For the support of complex values, what do you think of the following procedure:
Once this pull request is merged, I open an issue to discuss how to implement the complex value support. This is needed because cuSolver returns the singular values directly as reals while Eigen returns them as complex values and hence they are casted to real in the python wrapper. So the kernel definition and the python wrapper have to be adapted as well.
Then I'll implement this as a new pull request.

rmlarsen:
@shamanDevel Happy to hear it was a "learning experience" :-) Thanks for all your hard work! I like you suggestion for adding complex support. I think we should probably clean up the existing interfaces to return real singular values instead of the ugly casting. This might require renaming the kernels (adding V2 suffix for the new version) for backwards compatibility. 

shamanDevel:
Yeah, all checks have passed!

rmlarsen:
Woohoo! Thanks a lot for the contribution @shamanDevel!

hzhangxyz:
Emmm, It seems in r1.4(4196d6d), SVD on CPU is faster than SVD on GPU?

SVD size: 1024 * 1024 float32
GPU : one GeForce GTX 950M
CPU : Intel(R) Core(TM) i7-4720HQ CPU @ 2.60GHz, (4 core 8 thread)

hzhangxyz:
I don't know why in my PC, SVD on GPU is slower than SVD on CPU... version r1.4(4196d6d)
main.py
```
import tensorflow as tf
import numpy as np
import sys

D = 1024
dA = np.random.normal(size=(D,D))

dev = "/gpu:0" if len(sys.argv)==1 else "/cpu:0"

with tf.device(dev):
    A = tf.placeholder(shape=(D,D),dtype=tf.float32)
    S, U, V = tf.svd(A)

config = tf.ConfigProto()
config.log_device_placement = True
config.graph_options.optimizer_options.global_jit_level=tf.OptimizerOptions.ON_1
sess = tf.Session(config=config)

for _ in xrange(10):
    dS, dU, dV = sess.run((S, U, V), feed_dict={A:dA})
```

## run on GPU
time python main.py
```
2017-10-10 16:28:49.047703: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:892] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero
2017-10-10 16:28:49.048176: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1030] Found device 0 with properties:
name: GeForce GTX 950M major: 5 minor: 0 memoryClockRate(GHz): 1.124
pciBusID: 0000:0a:00.0
totalMemory: 3.95GiB freeMemory: 3.91GiB
2017-10-10 16:28:49.048205: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1120] Creating TensorFlow device (/device:GPU:0) -> (device: 0, name: GeForce GTX 950M, pci bus id: 0000:0a:00.0, compute capability: 5.0)
Device mapping:
/job:localhost/replica:0/task:0/device:XLA_GPU:0 -> device: XLA_GPU device
/job:localhost/replica:0/task:0/device:XLA_CPU:0 -> device: XLA_CPU device
/job:localhost/replica:0/task:0/device:GPU:0 -> device: 0, name: GeForce GTX 950M, pci bus id: 0000:0a:00.0, compute capability: 5.0
2017-10-10 16:28:49.064960: I tensorflow/core/common_runtime/direct_session.cc:299] Device mapping:
/job:localhost/replica:0/task:0/device:XLA_GPU:0 -> device: XLA_GPU device
/job:localhost/replica:0/task:0/device:XLA_CPU:0 -> device: XLA_CPU device
/job:localhost/replica:0/task:0/device:GPU:0 -> device: 0, name: GeForce GTX 950M, pci bus id: 0000:0a:00.0, compute capability: 5.0

Svd: (Svd): /job:localhost/replica:0/task:0/device:GPU:0
2017-10-10 16:28:49.067234: I tensorflow/core/common_runtime/placer.cc:874] Svd: (Svd)/job:localhost/replica:0/task:0/device:GPU:0
Placeholder: (Placeholder): /job:localhost/replica:0/task:0/device:GPU:0
2017-10-10 16:28:49.067302: I tensorflow/core/common_runtime/placer.cc:874] Placeholder: (Placeholder)/job:localhost/replica:0/task:0/device:GPU:0
2017-10-10 16:28:49.074053: I tensorflow/core/kernels/cuda_solvers.cc:159] Creating CudaSolver handles for stream 0x488e860
(1024,) (1024, 1024) (1024, 1024)
python main.py  27.50s user 2.30s system 100% cpu 29.658 total
```

## run on CPU
time python main.py -
```
2017-10-10 16:29:53.252138: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:892] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero
2017-10-10 16:29:53.252572: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1030] Found device 0 with properties:
name: GeForce GTX 950M major: 5 minor: 0 memoryClockRate(GHz): 1.124
pciBusID: 0000:0a:00.0
totalMemory: 3.95GiB freeMemory: 3.91GiB
2017-10-10 16:29:53.252600: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1120] Creating TensorFlow device (/device:GPU:0) -> (device: 0, name: GeForce GTX 950M, pci bus id: 0000:0a:00.0, compute capability: 5.0)
Device mapping:
/job:localhost/replica:0/task:0/device:XLA_GPU:0 -> device: XLA_GPU device
/job:localhost/replica:0/task:0/device:XLA_CPU:0 -> device: XLA_CPU device
/job:localhost/replica:0/task:0/device:GPU:0 -> device: 0, name: GeForce GTX 950M, pci bus id: 0000:0a:00.0, compute capability: 5.0
2017-10-10 16:29:53.269242: I tensorflow/core/common_runtime/direct_session.cc:299] Device mapping:
/job:localhost/replica:0/task:0/device:XLA_GPU:0 -> device: XLA_GPU device
/job:localhost/replica:0/task:0/device:XLA_CPU:0 -> device: XLA_CPU device
/job:localhost/replica:0/task:0/device:GPU:0 -> device: 0, name: GeForce GTX 950M, pci bus id: 0000:0a:00.0, compute capability: 5.0

Svd: (Svd): /job:localhost/replica:0/task:0/device:CPU:0
2017-10-10 16:29:53.271505: I tensorflow/core/common_runtime/placer.cc:874] Svd: (Svd)/job:localhost/replica:0/task:0/device:CPU:0
Placeholder: (Placeholder): /job:localhost/replica:0/task:0/device:CPU:0
2017-10-10 16:29:53.271544: I tensorflow/core/common_runtime/placer.cc:874] Placeholder: (Placeholder)/job:localhost/replica:0/task:0/device:CPU:0
(1024,) (1024, 1024) (1024, 1024)
python main.py -  34.33s user 10.68s system 621% cpu 7.241 total
```

shamanDevel:
@hzhangxyz I'm confused.
Your first log shows that the ops are placed on the GPU while the second log shows the CPU placement. This is exactly opposite to what the execution command 'time python main.py' indicates.
If I just look at the logs, then the GPU version takes 2.3sec (100% CPU load) and the CPU version takes 10.68sec (621% CPU load).
Based on that, the SVD is clearly faster on the GPU than the CPU version. Even if the CPU version uses multiple cores, indicated by a CPU load of 621%.

hzhangxyz:
@shamanDevel Hi

> Your first log shows that the ops are placed on the GPU while the second log shows the CPU placement. This is exactly opposite to what the execution command 'time python main.py' indicates.
yeah, in `time python main.py` len of `sys.argc` is `1` so it is running on GPU.

> If I just look at the logs, then the GPU version takes 2.3sec (100% CPU load) and the CPU version takes 10.68sec (621% CPU load).
but 2.3sec and 10.68sec are system time, which isn't the exact time cost. the exact time is 29.658sec(gpu) and 7.241sec(cpu). For example `time sleep 1` outputs `sleep 1  0.00s user 0.00s system 0% cpu 1.001 total`, in which 1.001 is the exact timecost, although cpu calculates nothing.

So, the problem is that SVD on GPU is slower than SVD on CPU sometimes, which is very strange.

PR #11879, opened 2017-07-30
Determinant-operation on the GPU

shamanDevel:
This pull request implements the computation of the matrix determinant on the GPU using LU-factorization with cuSolver.

This is the second part of the use-case presented in pull request #11878 .

It uses the LU factorization instead of QR-factorization, because this allows to compute the determinant also for non-symmetric matrices. This is required by the definition of the operation in Tensorflow. Hence, I also had to implement the LU-interface to cuSolver, QR would have already been implemented.

Two further notes: 
First, the precision of this algorithm to compute the determinant is less precise than the Eigen-implementation. Hence I had to relax the allowed error in the unit test. This is especially important for singular or near-singular matrices.
Second, if this pull request and #11878 will both be accepted, they will probably clash in cuda_solvers.h and/or cuda_solvers.cc. I'll resolve the conflicts if it happens to be so.

### DIFF
--- a/tensorflow/core/kernels/BUILD
+++ b/tensorflow/core/kernels/BUILD
@@ -2159,7 +2159,7 @@ tf_kernel_library(
 tf_kernel_library(
     name = "determinant_op",
     prefix = "determinant_op",
-    deps = LINALG_DEPS,
+    deps = LINALG_DEPS + if_cuda([":cuda_solvers"]),
 )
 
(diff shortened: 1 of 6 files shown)


rmlarsen:
@shamanDevel Thanks for your submission, but I just implemented this kernel myself. It should be available in the opensource tree later this week.

PR #11886, opened 2017-07-30
Inconsistent normalization formula

jacobyeager:
Hi there,

In the Input Normalization video, you normalize the pixels by dividing by 1/2 of the pixel depth; however, in the code you divide by the pixel depth. This confused me at first, and you may want to consider making the formula the same in both video and code.

Thanks!

All the best,
Jake


vincentvanhoucke:
Yes, it was an oversight. The point is that a factor 2 won't matter. I'd rather not change the behavior at this point since the class has been running for a while and I worry about unnecessarily changing people's baselines that depend on this notebook, and introducing further delta between the documentation produced around the assignments and the assignment itself. Thanks for flagging it though.

PR #11889, opened 2017-07-30
correct description of bounding box points

jeremy-rutman:
bbox points are upperleft and bottomright not bottomleft and upperright. Also coordinates were described in idiosyncratic (y,x) instead of (x,y) form. And no, those two corrections don't cancel.


jeremy-rutman:
signed CLA

benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

PR #11898, opened 2017-07-31
Update the graph used in quantization tutorial

korrawat:
_Another fix to quantization tutorial after #11285_

Currently, following the steps in the [quantization tutorial](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/performance/quantization.md#how-can-you-quantize-your-models) will lead to weird results as pointed out in #11181. This is because the graph file and the evaluation script are unmatched.

From this [image recognition tutorial](https://www.tensorflow.org/tutorials/image_recognition), there are two pretrained graphs with two evaluation scripts.
1. `classify_image_graph_def.pb` in [`inception-2015-12-05.tgz`](http://download.tensorflow.org/models/image/imagenet/inception-2015-12-05.tgz) is to be used with the (older) [`classify_image.py`](https://github.com/tensorflow/models/blob/master/tutorials/image/imagenet/classify_image.py).
1. `inception_v3_2016_08_28_frozen.pb` in [`inception_v3_2016_08_28_frozen.pb.tar.gz`](https://storage.googleapis.com/download.tensorflow.org/models/inception_v3_2016_08_28_frozen.pb.tar.gz) is to be used with [`label_image`](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/examples/label_image).

The weird results came from using `classify_image_graph_def.pb` with `label_image`. The graphs are different (see also [models#1314](https://github.com/tensorflow/models/issues/1314)), especially in the last layer. The two scripts extract information from the last layer in different ways, so it works with its corresponding graph but not the other.

### Further changes?
1. In the [image recognition tutorial](https://www.tensorflow.org/tutorials/image_recognition) ([source](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/docs_src/tutorials/image_recognition.md)), the usage of Python API points to option 1 (above) but the C++ API points to option 2. This might cause confusion as the two graphs/scripts are not equivalent. Now that there's a [python implementation](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/label_image/label_image.py) of the newer `label_image` by @freedomtan, should we update the tutorial to use this python implementation instead?
1. Should this [graph transform README](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms/README.md) be updated as well? Currently it uses the 2015 pre-trained graph, hence `--inputs='Mul' --outputs='softmax'`. If it is updated to the 2016 graph, these should instead say `--inputs=input --outputs=InceptionV3/Predictions/Reshape_1`.

### DIFF
--- a/tensorflow/docs_src/performance/quantization.md
+++ b/tensorflow/docs_src/performance/quantization.md
@@ -89,12 +89,14 @@ here's how you can translate the latest GoogLeNet model into a version that uses
 eight-bit computations:
 
 ```sh
-curl http://download.tensorflow.org/models/image/imagenet/inception-2015-12-05.tgz -o /tmp/inceptionv3.tgz
-tar xzf /tmp/inceptionv3.tgz -C /tmp/
+curl -L "https://storage.googleapis.com/download.tensorflow.org/models/inception_v3_2016_08_28_frozen.pb.tar.gz" |
+  tar -C tensorflow/examples/label_image/data -xz
 bazel build tensorflow/tools/graph_transforms:transform_graph
 bazel-bin/tensorflow/tools/graph_transforms/transform_graph \
-  --inputs="Mul" --in_graph=/tmp/classify_image_graph_def.pb \
-  --outputs="softmax" --out_graph=/tmp/quantized_graph.pb \
+  --in_graph=tensorflow/examples/label_image/data/inception_v3_2016_08_28_frozen.pb \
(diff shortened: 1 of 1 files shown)


suharshs (approved):
This looks fine to me. Pete would know more about the two label_image scripts and what the GTT tutorial uses.

petewarden (approved):

rmlarsen:

PR #11901, opened 2017-07-31
Make Windows Bazel GPU build work again

meteorcloudy:
This change mainly fixed the Windows GPU build.

Besides it also has some other fixes and improvments.

1. print installed Bazel version in configure.py

2. Remove slice_op_gpu.cu.cc from strided_slice_op because it's already in slice_op

3. Mark some failing GPU tests as no_windows_gpu

4. Add /DNOGDI flag for some targets to fix GPU build

5. Update TF_CUDNN_VERSION to 6.0

6. add test filters -no_gpu, -no_pip_gpu

@gunan

### DIFF
--- a/tensorflow/core/kernels/BUILD
+++ b/tensorflow/core/kernels/BUILD
@@ -103,7 +103,6 @@ tf_kernel_library(
         "strided_slice_op.h",
         "strided_slice_op_impl.h",
         "strided_slice_op_gpu.cu.cc",
-        "slice_op_gpu.cu.cc",
     ],
     deps = [
         ":bounds_check",
--- a/tensorflow/python/BUILD
+++ b/tensorflow/python/BUILD
@@ -2351,6 +2351,7 @@ cuda_py_test(
         ":math_ops",
         "//third_party/py/numpy",
     ],
+    tags = ["no_windows_gpu"],
 )
 
 cuda_py_test(
@@ -2368,6 +2369,7 @@ cuda_py_test(
         ":variables",
         "//third_party/py/numpy",
     ],
+    tags = ["no_windows_gpu"],
 )
 
 cuda_py_test(
@@ -2448,6 +2450,7 @@ cuda_py_test(
         ":special_math_ops",
         "//third_party/py/numpy",
     ],
+    tags = ["no_windows_gpu"],
 )
 
 py_library(
--- a/tensorflow/tools/ci_build/windows/gpu/pip/build_tf_windows.sh
+++ b/tensorflow/tools/ci_build/windows/gpu/pip/build_tf_windows.sh
@@ -46,7 +46,7 @@ run_configure_for_gpu_build
 
 clean_output_base
 
-bazel build -c opt --config=win-cuda $BUILD_OPTS tensorflow/tools/pip_package:build_pip_package || exit $?
+bazel build -c opt $BUILD_OPTS tensorflow/tools/pip_package:build_pip_package || exit $?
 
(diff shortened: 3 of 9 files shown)


tensorflow/core/kernels/BUILD
```diff
@@ -103,7 +103,6 @@ tf_kernel_library(
         "strided_slice_op.h",
         "strided_slice_op_impl.h",
         "strided_slice_op_gpu.cu.cc",
-        "slice_op_gpu.cu.cc",
```
gunan: Is this an intended change?

tensorflow/python/BUILD
```diff
@@ -2313,6 +2313,7 @@ cuda_py_test(
         ":math_ops",
         "//third_party/py/numpy",
     ],
+    tags = ["no_windows_gpu"],
```
gunan: Do these match with the test exclusion list in cmake tests?

tensorflow/tools/ci_build/windows/gpu/pip/build_tf_windows.sh
```diff
@@ -46,7 +46,7 @@ run_configure_for_gpu_build
 
 clean_output_base
 
-bazel build -c opt --config=win-cuda $BUILD_OPTS tensorflow/tools/pip_package:build_pip_package || exit $?
+bazel build -c opt $BUILD_OPTS tensorflow/tools/pip_package:build_pip_package || exit $?
```
gunan: Without the "--config" flag, where/how do we select GPU config for the build?
Are we sure that the built binaries still use GPU?

gunan:
Jenkins, test this please.

tensorflow/core/kernels/BUILD
meteorcloudy: Yes, I think slice_op_gpu.cu.cc should not be here.
It's already the source of slice_op during gpu build, it'll be compiled twice in different targets.
And then I got this linking error:

```
slice_op_gpu.cu.o : error LNK2005: "unsigned __int64 __cdecl Eigen::internal::_GLOBAL__N__34_slice_op_gpu_cu_compute_52_cpp1_ii_b2f7d5b8::get_random_seed(void)" (?get_random_seed@_GLOBAL__N__34_slice_op_gpu_cu_compute_52_cpp1_ii_b2f7d5b8@internal@Eigen@@YA_KXZ) already defined in slice_op_gpu.cu.o
```

tensorflow/tools/ci_build/windows/gpu/pip/build_tf_windows.sh
meteorcloudy: Yes, because `--config=win-cuda` is added into `.tf_configure.bazelrc` file when running ./configure with gpu support.
See https://github.com/tensorflow/tensorflow/blob/a4a3a3335bd273e6224d16a934b59531ef239893/configure.py#L726

tensorflow/python/BUILD
meteorcloudy: I don't see any of the tests under `tensorflow/python/ops` running in cmake build, maybe we missed them?
But all these tests used to pass in GPU build, we definitely need to investigate the reasons.
Filed issue to track this https://github.com/tensorflow/tensorflow/issues/11962

gunan:
Jenkins, test this please.

gunan (approved):

gunan:
@meteorcloudy could you sync your branch? then the change is ok to merge.

meteorcloudy:
Great, It's done!

PR #11907, opened 2017-07-31
[XLA] Any unrecognised fusion type should be described as kCustom

DavidNorman:
@majnemer @vrv 

Hi,  

this is the knock on from cancelling having a fusion sub-type.  if I use (kCustom+N) as types of fusion then I need the printing routine to not fail when given a type that is not recognised.

The cancelled PR was https://github.com/tensorflow/tensorflow/pull/11719

### DIFF
--- a/tensorflow/compiler/xla/service/hlo_instruction.cc
+++ b/tensorflow/compiler/xla/service/hlo_instruction.cc
@@ -2381,7 +2381,7 @@ string ToString(HloInstruction::FusionKind kind) {
       return "kConvBackwardFilter";
     case HloInstruction::FusionKind::kConvBackwardInput:
       return "kConvBackwardInput";
-    case HloInstruction::FusionKind::kCustom:
+    default:
       return "kCustom";
   }
 }


majnemer:
@DavidNorman Can you please explain, in more detail, the motivation behind this custom tag? I'd like to better understand the limitations in the current scheme.

DavidNorman:
Sure.  The limitations of the current scheme are that there are only 6 explicit types.  One of them is  'kConvBackwardFilter', which maps to a special backwards conv operation specifically for the GPU hardware.

In our hardware we have, for instance, an operation is which like parts of an average_pool.  It is like the reduce_window<add> followed by the divide.  I am using the fusion mechanism to fuse these two things into one thing, just like the kConvBackwardFilter does for the GPU hardware.

To label this fusion, I could introduce a new value into the enumeration that describes the operation.  However, I have been asked not to put things into the public repo that are specific to our hardware.  Notwithstanding the fact that other companies are doing exactly that, I think it is a good idea to separate out 3rd party devices.

Which means that the public repo must provide a mechanism for extending all enumerations cleanly.   Adding a 'custom' upper bound is one way of doing this - probably the simplest way.

Hope this helps.

Cheers

DavidNorman:
@majnemer 

Hi.

I wonder if you have had a chance to think about this?

Cheers

David

majnemer:
Hi,

I believe I can make your example of ReduceWindow followed by Divide work within the existing framework of fusion.

You could represent it as a kOutput fusion node where the kDivide follows the kReduceWindow.

DavidNorman:
Thanks.  Nevertheless I think that there should be the option to extend the enumeration.  I do not want to have several of my fusions labelled with the same token.

For instance, I have the reduce_window+divide, and also a convolution+add.  If I fuse them both to something called 'kOutput' then when I am lowering the graph later, how do I tell one from the other?

Also - if this is the case, why does GPU backward convolution get to have a specially named op?  It is merely a conv followed by an optional transpose...

```
  //        activations  gradients
  //              \         /
  //               v       v
  //              Convolution
  //                 conv
  //                   |
  //                   v
  //               Transpose (optional if identity transposition)
```

majnemer:
Here is one approach for identifying what sort of operation you are looking at for a kOutput node: do a reverse post order traversal on the instructions in the fused computation. The first operation which is not a parameter is the operation of interest. In the case of reduce_window+divide, it would give you the reduce_window. In the case of convolution+add, it would give you the add.

DavidNorman:
Again, I wonder why the GPU reverse convolution is allowed a special label, but 3rd party devices are not?

Having gone to the trouble of identifying large combinations of ops, I do not really want to have to go back through them again at elaboration time, in order to figure out what I knew before.

```
  // Sigmoid
  {{HloOpcode::kAdd, true, nullptr, {4, 1}},
   {HloOpcode::kMultiply, true, nullptr, {4, 2}},
   {HloOpcode::kTanh, true, nullptr, {3}},
   {HloOpcode::kMultiply, true, nullptr, {4, -1}},
   {HloOpcode::kConstant, true, IsConstantHalf, {}}},

  // BiasAdd on convolution (explicit broadcast)
  {{HloOpcode::kAdd, true, nullptr, {1, 2}},
   {HloOpcode::kCall, false, IsPoplarConvolution, {-1, -1}},
   {HloOpcode::kBroadcast, true, nullptr, {-1}}},

  // Random truncated normal with post scale and add
  {{HloOpcode::kAdd, true, nullptr, {2, 1}},
   {HloOpcode::kConstant, true, nullptr, {}},
   {HloOpcode::kMultiply, true, nullptr, {4, 3}},
   {HloOpcode::kConstant, true, nullptr, {}},
   {HloOpcode::kWhile, true, IsTruncatedNormalWhile, {5}},
   {HloOpcode::kRng, true, nullptr, {6, 7}},
   {HloOpcode::kConstant, true, nullptr, {}},
   {HloOpcode::kConstant, true, nullptr, {}}},
```

majnemer:
The GPU convolution fusions are things we would remove given the time to do so, I see them as warts on fusion.

Fusion nodes are mostly intended for rather generic computations. If your hardware only has limited support for particular fusions, it makes sense to use something like CustomCall to represent the specific thing you have already pattern matched.

DavidNorman:
Thanks for the suggestion.  I think I will shelve this for the time being and discuss this in the XLA forum to get some clarity.

Cheers

DavidNorman:
Ok.  It's all good.  Thanks for the input.  Closing this now, and using kCall until I can see what happens to the GPU fusion.

Thanks

:)

PR #11922, opened 2017-07-31
Modify implementation of tf.nn.batch_normalization

liuzjmike:
Give a slightly different implementation of tf.nn.batch_normalization so that per-example gradients needed by differentially private stochastic gradient descent can be calculated for gamma and beta in batch norm layers.

### DIFF
--- a/tensorflow/python/ops/nn_impl.py
+++ b/tensorflow/python/ops/nn_impl.py
@@ -752,12 +752,13 @@ def batch_normalization(x,
   Returns:
     the normalized, scaled, offset tensor.
   """
-  with ops.name_scope(name, "batchnorm", [x, mean, variance, scale, offset]):
-    inv = math_ops.rsqrt(variance + variance_epsilon)
+  with tf.name_scope(name, "batchnorm", [x, mean, variance, scale, offset]):
+    normalized = (x - mean) * tf.rsqrt(variance + variance_epsilon)
     if scale is not None:
-      inv *= scale
-    return x * inv + (offset - mean * inv
-                      if offset is not None else -mean * inv)
+      normalized *= scale
+    if offset is not None:
+      normalized += offset
+    return normalized
 
 
 def fused_batch_norm(
--- a/tensorflow/python/ops/nn_impl.py
+++ b/tensorflow/python/ops/nn_impl.py
@@ -752,8 +752,8 @@ def batch_normalization(x,
   Returns:
     the normalized, scaled, offset tensor.
   """
-  with tf.name_scope(name, "batchnorm", [x, mean, variance, scale, offset]):
-    normalized = (x - mean) * tf.rsqrt(variance + variance_epsilon)
+  with ops.name_scope(name, "batchnorm", [x, mean, variance, scale, offset]):
+    normalized = (x - mean) * math_ops.rsqrt(variance + variance_epsilon)
     if scale is not None:
       normalized *= scale
     if offset is not None:


rmlarsen:

sguada (commented):
As first step, can you run batch_norm_benchmark.py to measure the impact in performance of the change.

jhseu:
@liuzjmike Friendly ping for an update. Thanks!

liuzjmike:
Sorry for the late reply. I'm travelling right now. Will run the benchmark as soon as I get back.

liuzjmike:
Following are the outputs of two consecutive run of batch_norm_benchmark.py on the original implementation and this new implementation, respectively. 

Original implementation:
```
> bazel test //tensorflow/python:batch_norm_benchmark
WARNING: Output base '/home/home5/zl99/.cache/bazel/_bazel_zl99/7128647531e7ce469ec086a63edb81ef' is on NFS. This may lead to surprising failures and undetermined behavior.
INFO: Found 1 test target...
Target //tensorflow/python:batch_norm_benchmark up-to-date:
  bazel-bin/tensorflow/python/batch_norm_benchmark
INFO: Elapsed time: 39.019s, Critical Path: 5.24s
//tensorflow/python:batch_norm_benchmark                                 PASSED in 5.1s

Executed 1 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

New implementation:
```
> bazel test //tensorflow/python:batch_norm_benchmark
WARNING: Output base '/home/home5/zl99/.cache/bazel/_bazel_zl99/7128647531e7ce469ec086a63edb81ef' is on NFS. This may lead to surprising failures and undetermined behavior.
INFO: Found 1 test target...
Target //tensorflow/python:batch_norm_benchmark up-to-date:
  bazel-bin/tensorflow/python/batch_norm_benchmark
INFO: Elapsed time: 35.650s, Critical Path: 5.22s
//tensorflow/python:batch_norm_benchmark                                 PASSED in 5.1s

Executed 1 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

I'm not sure if the benchmarks were run completely and correctly because of the last line of the output. But as far as the result shows, the two implementation should have similar performance.

drpngx:
Sorry to be dense, but if it's the same speed and returns the same results, why change?

liuzjmike:
This is not necessary an improvement to TensorFlow itself. But in order to apply differentially private stochastic gradient descent described in [this paper](https://arxiv.org/abs/1607.00133) by Abadi et al. on a network with batch normalization layers, per-example gradients need to be calculated for `gamma` and `beta`. The original implementation makes this calculation very hard in the TensorFlow framework while the new implementation significantly simplifies it.

sguada:
You need to do blaze run not blaze test so it would print the actual timings.

sb2nov:
@liuzjmike ping

liuzjmike:
When I do bazel run I get this
```
> bazel run //tensorflow/python:batch_norm_benchmark
INFO: Found 1 target...
Target //tensorflow/python:batch_norm_benchmark up-to-date:
  bazel-bin/tensorflow/python/batch_norm_benchmark
INFO: Elapsed time: 389.044s, Critical Path: 289.23s

INFO: Running command line: bazel-bin/tensorflow/python/batch_norm_benchmark

Ran 0 tests in 0.000s

OK
```
There are other jobs running on the cluster so reading the target was slow. But according to the output I think no test was run. Any idea why?

I will also try again when the cluster is less busy.

drpngx:
You'll likely want to add the following to your command line:
```
-- --benchmarks=BatchNormBenchmark.benchmark_batch_norm
``

**Comment by liuzjmike** [comment]
```
bazel run //tensorflow/python:batch_norm_benchmark --benchmarks=BatchNormBenchmark.benchmark_batch_norm
```
gives `Unrecognized option: --benchmarks=BatchNormBenchmark.benchmark_batch_norm`.
```
bazel-bin/tensorflow/python/batch_norm_benchmark --benchmarks=BatchNormBenchmark.benchmark_batch_norm
```
gives the following output:

On original impl:
```
Forward convolution (lower layers).
2017-09-27 18:27:39.204075: W tensorflow/core/framework/op_def_util.cc:333] Op BatchNormWithGlobalNormalization is deprecated. It will cease to work in GraphDef version 9. Use tf.nn.batch_normalization().
cpu shape:4/3 #layers:10 mode:op scale:True train:False - 0.049103 secs
entry {
  name: "BatchNormBenchmark.batch_norm_cpu_input_shape_[8,128,128,32]_axes_[0,1,2]_mode_op_layers_10_scale_True_train_False"
  iters: 5
  wall_time: 0.0491034030914
}

cpu shape:4/3 #layers:10 mode:py scale:True train:False - 0.109263 secs
entry {
  name: "BatchNormBenchmark.batch_norm_cpu_input_shape_[8,128,128,32]_axes_[0,1,2]_mode_py_layers_10_scale_True_train_False"
  iters: 5
  wall_time: 0.109263372421
}

cpu shape:4/3 #layers:10 mode:slow scale:True train:False - 0.182931 secs
entry {
  name: "BatchNormBenchmark.batch_norm_cpu_input_shape_[8,128,128,32]_axes_[0,1,2]_mode_slow_layers_10_scale_True_train_False"
  iters: 5
  wall_time: 0.182931375504
}

=== op vs py: 122.5% ===
=== py vs slow: 67.4% ===
2017-09-27 18:27:42.277512: F tensorflow/core/common_runtime/gpu/gpu_device.cc:131] Check failed: err == cudaSuccess (2 vs. 0)
Abort (core dumped)
```

On proposed impl:
```
Forward convolution (lower layers).
2017-09-27 18:30:13.324701: W tensorflow/core/framework/op_def_util.cc:333] Op BatchNormWithGlobalNormalization is deprecated. It will cease to work in GraphDef version 9. Use tf.nn.batch_normalization().
cpu shape:4/3 #layers:10 mode:op scale:True train:False - 0.047454 secs
entry {
  name: "BatchNormBenchmark.batch_norm_cpu_input_shape_[8,128,128,32]_axes_[0,1,2]_mode_op_layers_10_scale_True_train_False"
  iters: 5
  wall_time: 0.0474540233612
}

cpu shape:4/3 #layers:10 mode:py scale:True train:False - 0.191871 secs
entry {
  name: "BatchNormBenchmark.batch_norm_cpu_input_shape_[8,128,128,32]_axes_[0,1,2]_mode_py_layers_10_scale_True_train_False"
  iters: 5
  wall_time: 0.191870594025
}

cpu shape:4/3 #layers:10 mode:slow scale:True train:False - 0.180249 secs
entry {
  name: "BatchNormBenchmark.batch_norm_cpu_input_shape_[8,128,128,32]_axes_[0,1,2]_mode_slow_layers_10_scale_True_train_False"
  iters: 5
  wall_time: 0.180248832703
}

=== op vs py: 304.3% ===
=== py vs slow: -6.1% ===
2017-09-27 18:30:33.873145: F tensorflow/core/common_runtime/gpu/gpu_device.cc:131] Check failed: err == cudaSuccess (2 vs. 0)
Abort (core dumped)

drpngx:
Yes, if you use `bazel run` you need the extra leading `--` to indicate that it isn't a bazel option, but an option passed to the test.

The core dump is a bit worrisome, though. Could you look into it?

drpngx:
Finally, the proposed implementation looks 2x slower?

liuzjmike:
Thank you for the tip. I tried running with
```
bazel run //tensorflow/python:batch_norm_benchmark -- --benchmarks=BatchNormBenchmark.benchmark_batch_norm
```
and the core dump did not happen. It might be due to some mechanism of bazel.

Yes it looks like the proposed impl is sadly slower than the original one, maybe due to the original impl optimizing on the calculation dimension. Maybe we can give users an option in tf.contrib.layer.batch_norm to use the proposed impl if they want to do the per-example gradient operation. If you don't want to make that change, I'm ok with closing this PR.

drpngx:
@sguada what's the right thing to do? Close or provide this implementation as an additional option?

sguada:
Given the slowness I'm not inclined to include it, also adding extra options requires extra maintenance. So unless convinced otherwise I would close it.

martinwicke:
Closing. It can always be reopened, or a new PR can be made.

PR #11925, opened 2017-08-01
Update tf.nn.max_pool docs for ksize/strides == 4

### LINKED ISSUE #10729: tf.nn.max_pool wrong docs?
### System information
Not Applicable

### Describe the problem
[API](https://www.tensorflow.org/api_docs/python/tf/nn/max_pool) states that ksize has length >= 4, the size of window for each dimension of the input tensor. However, value is a 4-D Tensor so doesn't this mean that ksize should be length == 4? Same for strides.

Digging into maxpooling_op.cc shows that there's some check that does `==`. Line 212:

```
    OP_REQUIRES(context, ksize_.size() == 4,
                errors::InvalidArgument("Sliding window ksize field must "
                                        "specify 4 dimensions"));
```

yongtang:
Update tf.nn.max_pool docs for `ksize/strides == 4`, instead of '>= 4' in the existing docs.

As both `max_pool` and `avg_pool` are wrapped in the python code, this fix only changes the docstring in `tensorflow/python/ops/nn_ops.py`.

This fix fixes #10729.

Note: I also tried using `:=` syntax, as was suggested in https://github.com/tensorflow/tensorflow/issues/10729#issuecomment-309133461 However, it seems that the `:=` only works with `Input`, not with `Attr`.

### DIFF
--- a/tensorflow/python/ops/nn_ops.py
+++ b/tensorflow/python/ops/nn_ops.py
@@ -1721,9 +1721,9 @@ def avg_pool(value, ksize, strides, padding, data_format="NHWC", name=None):
   Args:
     value: A 4-D `Tensor` of shape `[batch, height, width, channels]` and type
       `float32`, `float64`, `qint8`, `quint8`, or `qint32`.
-    ksize: A list of ints that has length >= 4.
+    ksize: A 1-D int Tensor of 4 elements.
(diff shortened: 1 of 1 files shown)


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

PR #11932, opened 2017-08-01
Increase docstring consistency

taehoonlee:
This PR considers the following six main items:
- Adjusting indents of markdown code blocks
- Adding language identifiers on markdown code blocks
- Replacing `Return` with `Returns` (The frequencies over all python codes: 2 vs. 3369)
- Replacing `Arguments` with `Args`
(`Args` is a consistent word over all python codes except `tensorflow/contrib/keras` and `tensorflow/python/layers`)
- In `tensorflow/python/framework/tensor_shape.py`, removing ` ``` ` for consistency on the file
- In `tensorflow/python/ops/array_ops.py` and `math_ops.py`, revising python code blocks to make them more concrete
  - I think it is better to replace `==>` with `#` because `==>` is not a python keyword.
  - For each operation, two styles (with and without `tf.`) are mixed. I think it is better to make them as `tf.operation` not `operation`.
  - For more concrete examples, I think it is better to write down real tensors not pseudo. (e.g.,`# 'x' is [1, 4]` is replaced by `x = tf.constant([1, 4])`.)

### DIFF
--- a/tensorflow/python/layers/base.py
+++ b/tensorflow/python/layers/base.py
@@ -696,11 +696,9 @@ def _unique_layer_name(name):
 
   Example:
 
-  ```
-    >>> _unique_layer_name('dense')
-    dense_1
-    >>> _unique_layer_name('dense')
-    dense_2
+  ```python
+  _unique_layer_name('dense')  # dense_1
+  _unique_layer_name('dense')  # dense_2
   ```
   """
   graph = ops.get_default_graph()
--- a/tensorflow/contrib/batching/python/ops/batch_ops.py
+++ b/tensorflow/contrib/batching/python/ops/batch_ops.py
@@ -67,7 +67,7 @@ def batch_function(num_batch_threads, max_batch_size, batch_timeout_micros,
 
   So, for example, in the following code
 
-  ```
+  ```python
   @batch_function(1, 2, 3)
   def layer(a):
     return tf.matmul(a, a)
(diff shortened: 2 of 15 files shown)


tensorflow/python/layers/base.py
```diff
@@ -696,7 +696,7 @@ def _unique_layer_name(name):
 
   Example:
 
-  ```
+  ```python
     >>> _unique_layer_name('dense')
```
MarkDaoust: In the rest of the examples we have no carets, and the result in an inline comment.

You could do the same thing here.

MarkDaoust (changes requested):
I like these changes. Making things more consistent, and making examples more directly runnable is very nice.

Most of my attached comments are just farther consistency improvements.

But please revert the two spots where you changed the code at the same time.

Thanks.

taehoonlee:
I agree with the feedback, thank you @MarkDaoust! I reflected all the comments.

MarkDaoust:
I am happy with this.

It will also fixes a bunch of inline examples that weren't rendering on the website.  

I've re-checked the PR. It is now a pure docs-only change. 
All my comments are resolved.
I've added a couple of tiny fixes to some broken examples.

(It could have been a little easier to review the update if you had just added the commit instead of squashing them together.)

Thanks!

MarkDaoust (approved):

rmlarsen:

PR #11945, opened 2017-08-01
Fix #11803: Force CUPTI to flush GPU activity buffers when stopping tracing.

### LINKED ISSUE #11803: /gpu:0/stream doesn't appear on Timeline
### System information
Ubuntu 16.04
source
('v1.2.0-2479-g88abddb', '1.3.0-rc0')
- **Python version**: 
2.7.12
bazel release 0.4.5
- **CUDA/cuDNN version**:
CUDA 8.0
cuDNN 5.1.5
- **GPU driver version**
375.66 
- **GPU model and memory**:
GeForce GTX 1080 
8GB
python convolution.py (attached)

### The Problem
Making a timeline.Timeline object in my environment, "/gpu:0/stream:xx" rows don't appear. 
timeline_no_gpu

I found the event collector function [BufferCompleted](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/default/gpu_tracer.cc#L149) were not called at all during measuring performance. 

Though BufferCompleted is registered as a callback function to CUPTI for processing GPU events, CUPTI doesn't call it in my environment. This comes from failing to flush GPU events.
GPU events are flushed by [ActivityFlushAll](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/default/gpu_tracer.cc#L206) with flag [CUPTI_ACTIVITY_FLAG_NONE](http://docs.nvidia.com/cuda/cupti/group__CUPTI__ACTIVITY__API.html#group__CUPTI__ACTIVITY__API) in tensorflow. But this flag doesn't seem to cover necessary GPU events, so GPU events were not flushed and collected.

I modified gpu_tracer.cc to call ActivityFlushAll with flag CUPTI_ACTIVITY_FLAG_FORCE_INT, then I got "/gpu:0/stream:xx" rows in Timeline. 
timeline_with_gpu

Is there any reason to use CUPTI_ACTIVITY_FLAG_NONE?

### Source code / logs
Attached source code comes from [tensorflow/models](https://github.com/tensorflow/models/blob/master/tutorials/image/mnist/convolutional.py) .
The following modification was added to measure performance.

        from tensorflow.python.client import timeline
        run_options = tf.RunOptions(trace_level=tf.RunOptions.FULL_TRACE)
        run_metadata = tf.RunMetadata()
        l, lr, predictions = sess.run([loss, learning_rate, train_prediction],
                                       feed_dict=feed_dict, options=run_options, run_metadata=run_metadata)

        tl = timeline.Timeline(run_metadata.step_stats)
        ctf = tl.generate_chrome_trace_format(show_memory=True,
                                            show_dataflow=True)
        with open("timeline.json", "w") as f:
          f.write(ctf)

[convolutional.py.zip](https://github.com/tensorflow/tensorflow/files/1178943/convolutional.py.zip)

laket:
As described #11803, stopping GPU tracing, GPU events are not flushed. Then we cannot get GPU activities in performance statistics.

Calling  ActivityFlushAll() with flag CUPTI_ACTIVITY_FLAG_NONE causes this problem, so this commit changes the flag to CUPTI_ACTIVITY_FLAG_FLUSH_FORCED.

### DIFF
--- a/tensorflow/core/platform/default/gpu_tracer.cc
+++ b/tensorflow/core/platform/default/gpu_tracer.cc
@@ -205,7 +205,7 @@ Status CUPTIManager::DisableTrace() {
   CUPTI_CALL(ActivityDisable(CUPTI_ACTIVITY_KIND_MEMCPY));
   CUPTI_CALL(ActivityDisable(CUPTI_ACTIVITY_KIND_MEMCPY2));
   CUPTI_CALL(ActivityDisable(CUPTI_ACTIVITY_KIND_MEMSET));
-  CUPTI_CALL(ActivityFlushAll(0));
+  CUPTI_CALL(ActivityFlushAll(CUPTI_ACTIVITY_FLAG_FLUSH_FORCED));
   {
     // Don't acquire this lock until Flush returns, since Flush
     // will potentially cause callbacks into BufferCompleted.


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

Issue #11871, opened 2017-07-29
Error during compilation of tensorflow-GPU using bazel 0.5.3


feranick:
Using Bazel 0.5.3 (from installer, both sh and deb) on Ubuntu 16.04 to compile tensorflow-gpu, I get the following error that fails compilation:

` Encountered error while reading extension file 'cuda/build_defs.bzl': no such package '@local_config_cuda//cuda': Traceback (most recent call last):
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 1039
		_create_local_cuda_repository(repository_ctx)
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 976, in _create_local_cuda_repository
		_host_compiler_includes(repository_ctx, cc)
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 145, in _host_compiler_includes
		get_cxx_inc_directories(repository_ctx, cc)
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 120, in get_cxx_inc_directories
		set(includes_cpp)
depsets cannot contain mutable items
WARNING: Target pattern parsing failed.
ERROR: error loading package 'tensorflow/tools/pip_package': Encountered error while reading extension file 'cuda/build_defs.bzl': no such package '@local_config_cuda//cuda': Traceback (most recent call last):
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 1039
		_create_local_cuda_repository(repository_ctx)
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 976, in _create_local_cuda_repository
		_host_compiler_includes(repository_ctx, cc)
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 145, in _host_compiler_includes
		get_cxx_inc_directories(repository_ctx, cc)
	File "/home/nicola/Software/tensorflow/gpu/tensorflow/third_party/gpus/cuda_configure.bzl", line 120, in get_cxx_inc_directories
		set(includes_cpp)`

[...]

PR #11949, opened 2017-08-01
Fix "depsets cannot contain mutable items" error when using Bazel 0.5.3

allenlavoie:
Allows CUDA builds with more recent versions of Bazel. Fixes https://github.com/tensorflow/tensorflow/issues/11871. Perhaps we should also file a Bazel bug, although it would be nice to have the workaround in the meantime.

### DIFF
--- a/third_party/gpus/cuda_configure.bzl
+++ b/third_party/gpus/cuda_configure.bzl
@@ -106,7 +106,7 @@ def _get_cxx_inc_directories_impl(repository_ctx, cc, lang_is_cpp):
   else:
     inc_dirs = result.stderr[index1 + 1:index2].strip()
 
-  return [repository_ctx.path(_cxx_inc_convert(p))
+  return [str(repository_ctx.path(_cxx_inc_convert(p)))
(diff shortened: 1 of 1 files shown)


martinwicke:
We know that building with bazel 0.5.3 doesn't work -- bazel knows as well. Thanks for the fix though!

martinwicke (approved):

yifeif:
Thanks Allen! cmake failure is an infra issue, and this shouldn't affect cmake anyway. Merging.

martinwicke:
@av8ramit is this on the 1.3 branch? If not, can you cherry-pick it? It's build only so no need for another RC because of this.

ahundt:
@martinwicke @av8ramit I've created #12248 where I cherry picked the fix and added a pull request to r1.3 as detailed above. I needed the change anyway. :-)

av8ramit:
Sure, I'll add it to 1.3

av8ramit:
Thanks @ahundt ! I'll merge your change into the 1.3 branch.

PR #11951, opened 2017-08-01
Make plugin_data optional and (not repeated)

chihuahua:
Make plugin_data optional and (not repeated) …Every summary op writes data for a single plugin to process. Hence, each SummaryMetadata proto should have a single PluginData optional field (instead of a repeated one). This removes much complexity from TensorBoard logic that loops over the plugin data. It also simplifies the SQL schema - it can now enforce a one-to-one relationship between summary op and plugin.


chihuahua:
I am closing this specific PR. Please review https://github.com/tensorflow/tensorflow/pull/11952 instead.

Issue #10089, opened 2017-05-22
Scala API


eaplatanios:
Hi,

I have open sourced a Scala API for TensorFlow that contains much more complete functionality than the Java API, [here](https://github.com/eaplatanios/tensorflow_scala). The README file in the repository contains information on what features are supported etc. I would really appreciate some feedback from interested parties in the community on the library.

One main limitation is that the library does not yet support while loops in the graph. This is due to being unable to implement gradient backpropagation using the current C API. @skye do you have any suggestions on how to proceed with respect to that? It would be really useful for implementing RNNs.

Other cool stuff (such as fetching arbitrarily structured data from sessions -- e.g., lists of tensors, or maps of tensors, etc.) are supported and are for the most part strongly typed.

Cheers,
Anthony

skye:
Super cool Anthony! This looks really great. FYI @asimshankar

Regarding while loops, I still need to reimplement the backprop logic in the C API as we discussed in #9150. Given that we're both depending on this (I'm currently switching the Python API to use the C API), I'll start work on this ASAP. I can't make any promises as to when it will land, but hopefully within the next month depending on the complexity and what else comes up (I'll also be on vacation next week).

eaplatanios:
@skye Thanks! :)

Regarding the gradients, I'm currently not using the provided C API gradients, because it's missing a lot of them (my library still supports it but the default one is my implementation). I have instead implemented back-propagation in the Scala API (essentially porting the Python API code). In this case it would be great for your implementation to allow me to implement back-propagation over while loops in the Scala side, or to allow registering gradient functions through the C API (although I feel this is not easy to do). Do you have any thoughts on this?

P.S. I'm actually also going on vacation in a couple weeks! Have a great time wherever you're going! :))

jart:
Thank you @eaplatanios for bringing your work to the attention of the community. We appreciate what you're doing. We hope the Scala community will find TensorFlow helpful.

Here's one way we might be able to support you. You could send us a pull request updating our documentation to point to your project. For example, the [Which client languages are supported in TensorFlow?](https://www.tensorflow.org/programmers_guide/faq) section in our Programmer's Guide. You could add a sentence that says something like, "There is also an independently maintained interface for [Scala](https://github.com/eaplatanios/tensorflow_scala)." 

Any thoughts @martinwicke?

CC: @jdegoes

asimshankar:
[...]

PR #11960, opened 2017-08-02
Added an "UpdateInput" method to the Python C API

eaplatanios:
This is the only current limitation I see for allowing the implementation of gradients through the C API for while loops. @skye this relates to our discussion in #10089 and could provide a temporary solution until we have a more elegant solution. Since it is in the `python_api.h` interface, similar to `AddControlInput`, I hope accepting this doesn't cause any issues.


eaplatanios:
I also added a "ClearControlInputs" function which turns out to be necessary to implement the "_RemoveExternalControlEdges" method of control flow contexts in the Python API.

skye:
Hi Anthony, I'd prefer not to add the ClearControlInputs function, and have you wait for the C API implementation. (We actually need the UpdateInput function to support parts of the Python API.) Have you actually started implementing while loop backprop in Scala? I believe even with these functions it'll be tricky without having the C API provide something like WhileContext (although maybe possible, if you traverse the new while loop and reconstruct the state needed).

eaplatanios:
@skye I have actually started implementing it by mirroring what's done in the Python API. I am creating and managing control flow contexts in the Scala side, similar to how it's done on the Python side. However, that requires the ClearControlInputs function since that's used in Python. Once I finish the implementation and have something working, I can look into whether I can remove the calls to ClearControlInputs. This is unclear at this point though.

One problem I have in my mind that is also not resolved in the Python API is what happens when you deserialize a graph from a MetaGraphDef file. This is generally handled very loosely in the Python API (e.g., with respect to prevent feeding and fetching, and contexts). I have done some work to improve the Scala implementation such that after loading a graph from a file, appending ops to that graph and manipulating it mimicks the behavior you'd get had you created it in the Scala side in the first place. However, with respect to control flow contexts, I don't yet have a clear picture at to when they come in and out and I can also not see them being serialized with the graph in the Python side. That's something to think about I guess, but I'll work on it after I have a basic implementation working for while loops.

Also, note that I am implementing conditionals (i.e., tf.cond) in the same way as done for the Python API. That is, using a CondContext.

skye:
Ah I see. Unfortunately we really do not want every language binding to reimplement this code (it's extremely easy to introduce bugs, and forces us to allow dangerous graph modifications like creating cycles, as this patch does). It looks like you're familiar with C/++; would you be willing to help me with the C API while loop implementation? I'm working on getting the major gradient pieces in, but there are still many TODOs to resolve in the forward-pass loop implementation. I can help you get started with some of these if you like, and once I get the basic gradient code checked in we can possibly divide up that work as well. What do you think?

Re: loading MetaGraphDefs, I too have noticed that loading while loops doesn't really work if you wanna continue graph construction. I haven't thought about this much either, since my immediate goal is to reimplement the current Python behavior. I agree it would be good to properly handle this though.

eaplatanios:
@skye I would be more than willing to help with C API while loop implementation, as I do agree that repetition of this code in other language APIs is not a good idea. :)

I have a couple of questions though to get started with this:
1. Why do we also have a CondContext in the Python API? If we do not have that, how would we handle nested conds? In general, I would like to be clear on the functionality related to cond, before we go into while loops.
2. Intuitively, it seems to me that we would know the control flow context of every op, at creation time, since they would be part of a cond, for example. So, wouldn't it better to make the control flow context of each op immutable and serialize it together with the op? We can perform the serialization efficiently by storing all control flow contexts of a graph, alongside the graph and then storing an ID for each op's control flow context.
3. Are redundant cycles currently being introduced in the Python API implementation?

I'm not sure if my questions make sense the way I phrased them, but I could explain further if needed. Probably the most helpful thing would be if you gave me an overview of what needs to be done as you said.

In any case, it would also be good to think of potential ways for other language APIs to provide gradient implementations that are not yet supported in the C API, so that we can have some plan in mind for how to integrate well with them. We could follow a similar approach to how while loops are defined and create a new graph for each gradient op that we then merge into the main target graph, although I'm not sure how efficient that would be.

skye:
Great, thanks Anthony! And apologies for not suggesting we collaborate earlier, I didn't mean to send you down a rabbit hole. I'll answer your questions shortly, but do you mind shooting me an email in the meantime? That'll probably be better than communicating in this PR :)

rmlarsen:
@skye what's the status of this?

skye:
We don't actually need this patch, closing.

PR #11966, opened 2017-08-02
[XLA] Add a scheme for disabling tests based on type support by backends

DavidNorman:
This change allows CPP tests to be marked as requiring certain type support.  As it stands, I have updated only the scalar math test with the annotations.  I will do the rest once this change is debated.

The plugin config has something like this added to the 'copts' option:

```
    "copts": [
      "-DXLA_TEST_DISABLE_F64",
    ],
```


DavidNorman:
after discussion on the xla dev forum, cancelling this.

PR #11968, opened 2017-08-02
[OpenCL] Fix allocator destruction race condition (#136)

lukeiwanski:
* [OpenCL] Changes SYCL Interface construction

Uses C++11 static initialisation to provide singleton instance, rather
than a mutex and pointer.

* [OpenCL] Adds const to SYCL Interface methods

* [OpenCL] Fix allocator destruction race condition

A Tensor's allocator must outlive it, however there is no easy way to
determine whether an Allocator has any Tensors still alive, and so we
cannot know when it is safe to destroy an allocator. The CPU allocator
gets round this by being deleted, so we adopt this convention here.

* [OpenCL] Reformats SYCL code

* [OpenCL] Fixes SYCL comments

* [OpenCL] Tidies SYCL device description

Adds check for whether the QueueInterface pointer is valid, as this
may not always be the case.

* [OpenCL] Adds nullptr checking to SYCL allocator

* [OpenCL] Adds const specifier to SYCL Interface

### DIFF
--- a/tensorflow/core/common_runtime/sycl/sycl_allocator.cc
+++ b/tensorflow/core/common_runtime/sycl/sycl_allocator.cc
@@ -36,6 +36,7 @@ SYCLAllocator::~SYCLAllocator() {
 string SYCLAllocator::Name() { return "device:SYCL"; }
 
 void* SYCLAllocator::AllocateRaw(size_t alignment, size_t num_bytes) {
+  mutex_lock lock(mu_);
   assert(sycl_device_);
   if (num_bytes == 0) {
     // Cannot allocate no bytes in SYCL, so instead allocate a single byte
@@ -45,7 +46,6 @@ void* SYCLAllocator::AllocateRaw(size_t alignment, size_t num_bytes) {
   const auto& allocated_buffer = sycl_device_->get_sycl_buffer(p);
   const std::size_t bytes_allocated = allocated_buffer.get_range().size();
 
-  mutex_lock lock(mu_);
   ++stats_.num_allocs;
   stats_.bytes_in_use += bytes_allocated;
   stats_.max_bytes_in_use =
@@ -57,10 +57,10 @@ void* SYCLAllocator::AllocateRaw(size_t alignment, size_t num_bytes) {
 }
 
 void SYCLAllocator::DeallocateRaw(void* ptr) {
+  mutex_lock lock(mu_);
   if (sycl_device_) {
     const auto& buffer_to_delete = sycl_device_->get_sycl_buffer(ptr);
     const std::size_t dealloc_size = buffer_to_delete.get_range().size();
-    mutex_lock lock(mu_);
     stats_.bytes_in_use -= dealloc_size;
     sycl_device_->deallocate(ptr);
   }
@@ -72,6 +72,7 @@ void SYCLAllocator::GetStats(AllocatorStats* stats) {
 }
 
 size_t SYCLAllocator::RequestedSize(void* ptr) {
+  mutex_lock lock(mu_);
   if(!sycl_device_) {
     return 0;
   }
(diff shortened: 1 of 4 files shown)


tensorflow/core/common_runtime/sycl/sycl_allocator.h
```diff
@@ -46,8 +50,14 @@ class SYCLAllocator : public Allocator {
   // AllocatedSize(void* ptr) by default.
   size_t RequestedSize(void* ptr) override;
   Eigen::SyclDevice* getSyclDevice() { return sycl_device_; }
+  // Clear the SYCL device used by the Allocator
+  void ClearSYCLDevice() {
+    delete sycl_device_;
```
rmlarsen: Should this be protected by a mutex. Might it now race with some of the if (sycl_device_) usage above?

tensorflow/core/common_runtime/sycl/sycl_allocator.cc
```diff
@@ -19,7 +19,7 @@ limitations under the License.
 
 namespace tensorflow {
 
-SYCLAllocator::SYCLAllocator(Eigen::QueueInterface *queue)
+SYCLAllocator::SYCLAllocator(Eigen::QueueInterface* queue)
```
rmlarsen: In the future, please keep large formating changes and functional changes in separate PRs.

lukeiwanski: Ok sorry :)

tensorflow/core/common_runtime/sycl/sycl_allocator.h
lukeiwanski: Fixed in d75b94f3e42eb88c4f0428499be11cce13abe87a

tensorflow/core/common_runtime/sycl/sycl_allocator.h
```diff
@@ -46,8 +50,17 @@ class SYCLAllocator : public Allocator {
...
+    if(sycl_device_) {
+      mutex_lock lock(mu_);
+      delete sycl_device_;
+      sycl_device_ = nullptr;
+    }
+  }
+
  private:
-  Eigen::SyclDevice *sycl_device_;  // owned
+  Eigen::SyclDevice* sycl_device_;  // owned
```
rmlarsen: Move down and decorate with GUARADED_BY(mu_) ?

tensorflow/core/common_runtime/sycl/sycl_allocator.h
```diff
@@ -46,8 +50,17 @@ class SYCLAllocator : public Allocator {
   // AllocatedSize(void* ptr) by default.
   size_t RequestedSize(void* ptr) override;
   Eigen::SyclDevice* getSyclDevice() { return sycl_device_; }
+  // Clear the SYCL device used by the Allocator
+  void ClearSYCLDevice() {
+    if(sycl_device_) {
+      mutex_lock lock(mu_);
```
rmlarsen: This only helps if all the reads and uses of sycl_device_ are guarded by the mutex. For example 

Thread 1 calls, e.g., Synchronize, reads non-zero sycl_device_ in  line 40.
Thread 2 calls ClearSYCLDevice, deletes the object, sets sycl_device_ = nullptr;
Thread 1 Tries to call sycl_device_->synchronize()
Boom!

I'm not sure if this is a scenario that we need to guard against?

rmlarsen (changes requested):

lukeiwanski:
Thanks for the feedback we will apply it ASAP

tensorflow/core/common_runtime/sycl/sycl_allocator.h
lukeiwanski: Fixed in: 8070cad

tensorflow/core/common_runtime/sycl/sycl_allocator.h
lukeiwanski: Fixed in 8070cad

rmlarsen (approved):

rmlarsen:

PR #11973, opened 2017-08-02
[Java] Build rules and skeleton for operation wrappers generator

karllessard:
In this PR, Bazel rules to generate operation wrappers in Java at build-time and a very minimal implementation of the generator (in C++) are added.

For now, the generator only create empty packages (one package per op-library) and combine them into a source archive (.srcjar) that is added as a source dependency of the tensorflow Java library target.

The goal of this PR is to make sure that everything is setup properly and agreed on before starting to generate for real the operation wrappers.


tensorflow/java/src/gen/cc/ops/main.cc
```diff
@@ -0,0 +1,33 @@
...
+
+#include <string>
+
+#include "tensorflow/core/platform/init_main.h"
+#include "tensorflow/core/platform/logging.h"
+#include "tensorflow/java/src/gen/cc/ops/op_generator.h"
+
+int main(int argc, char* argv[]) {
+  tensorflow::port::InitMain(argv[0], &argc, &argv);
+  const std::string& output_dir = argv[1];
```
asimshankar: These will segfault (as opposed to providing a useful error message) if both arguments aren't provided.
Perhaps add something like:

`QCHECK_GT(argc, 2) << (usage string)`

tensorflow/java/src/gen/cc/ops/op_generator.h
```diff
@@ -0,0 +1,37 @@
...
+#include <string>
+
+namespace tensorflow {
+
+class OpGenerator {
+ public:
+  explicit OpGenerator(const std::string& ops_lib);
+  virtual ~OpGenerator();
+
+  int Run(const std::string& output_path);
```
asimshankar: Out of curiosity: How did you decide on the partition between a constructor argument (`ops_lib`) and an argument to `Run`?

tensorflow/java/src/gen/cc/ops/op_generator.h
```diff
@@ -0,0 +1,37 @@
...
+
+ private:
+  const std::string ops_lib;
+  std::string package_name;
```
asimshankar: Should this be a `const` too? (Can be set in the initializer list)

tensorflow/java/src/gen/cc/ops/op_generator.h
```diff
@@ -0,0 +1,37 @@
...
+==============================================================================*/
+
+#ifndef TENSORFLOW_JAVA_SRC_GEN_CC_OPS_OP_GENERATOR_H_
+#define TENSORFLOW_JAVA_SRC_GEN_CC_OPS_OP_GENERATOR_H_
+
+#include <string>
+
+namespace tensorflow {
+
+class OpGenerator {
```
asimshankar: Some commentary about the class, its purpose and usage would be great.

tensorflow/java/src/gen/cc/ops/op_generator.cc
```diff
@@ -0,0 +1,52 @@
...
+  tensorflow::Env* env = tensorflow::Env::Default();
+  std::string package_path(output_path);
+  package_path += gen_util::ReplaceChar(package_name, '.', '/');
+
+  if (!env->FileExists(package_path).ok()) {
+    TF_CHECK_OK(env->RecursivelyCreateDir(package_path));
+  }
+
+  OpList ops;
+  OpRegistry::Global()->Export(true, &ops);
```
asimshankar: My recommendation would be for `OpGenerator` to be provided with an `OpList` and the use of the global registry can be moved to `main()`. That would make `OpGenerator` more of a library that doesn't depend on global variables (`OpRegistry::Global`) and keep all the global variables in `main()`, which will be nice.

tensorflow/java/src/gen/cc/ops/op_generator.h
asimshankar: Instead of returning an `int`, `Run` should return a `tensorflow::Status` as that can provide more detailed error information. And then, you'd avoid `TF_CHECK_OK` in the implementation and have `main()` print out the error string (if any)

asimshankar (changes requested):
Thanks for the PR and in particular separating out this skeleton from the "meat" that will come later :)

tensorflow/java/src/gen/cc/ops/op_generator.cc
```diff
@@ -0,0 +1,52 @@
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
asimshankar: Oh, also, I think we can simplify the directory structure, don't need all these subdirectories.
How about placing all these files in `tensorflow/javasrc/gen`? (So, `tensorflow/java/src/gen/op_generator{.h,.cc,_main.cc}` etc. instead of the `cc/ops` subdirectories)

tensorflow/tensorflow.bzl
```diff
@@ -403,6 +403,57 @@ def tf_gen_op_wrapper_py(name,
       ],)
 
 
+# Given a list of "ops_libs" (a list of files in the core/ops directory
```
karllessard: I've put it there only because similar functions for C++ and Python wrappers are already in this file, even if there is no apparent reason for having them as such a high level. I'm pleased to move it back down like you propose.

tensorflow/java/src/gen/cc/ops/main.cc
```diff
@@ -0,0 +1,33 @@
...
+ ==============================================================================*/
+
+#include <string>
+
+#include "tensorflow/core/platform/init_main.h"
+#include "tensorflow/core/platform/logging.h"
+#include "tensorflow/java/src/gen/cc/ops/op_generator.h"
+
+int main(int argc, char* argv[]) {
+  tensorflow::port::InitMain(argv[0], &argc, &argv);
```
karllessard: Sounds good

tensorflow/java/src/gen/cc/ops/main.cc
karllessard: Sounds good too

tensorflow/java/src/gen/cc/ops/op_generator.h
karllessard: Probably because I knew I'll need `ops_lib` again later on but `output_path` could be discarded once I have my `package_path`... But let's keep them together in the constructor.

tensorflow/java/src/gen/cc/ops/op_generator.h
karllessard: All right

tensorflow/java/src/gen/cc/gen_util.h
```diff
@@ -0,0 +1,46 @@
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
karllessard: It probably can, I'll take a look, thanks!

tensorflow/java/src/gen/cc/ops/op_generator.h
karllessard: Ok

tensorflow/java/src/gen/cc/ops/op_generator.h
karllessard: Sorry, that's the result of moving some stuff around without connecting all the dots at the end, I'll mark it `const`.

tensorflow/java/src/gen/cc/ops/op_generator.cc
karllessard: Good, I'll pass the `OpList` as a argument to the `Run` method. I think it could make sense now to have the `ops_lib` name also as an argument to `Run` with the list of ops since they go together... (that may help you understand what is my mind set when I'm trying to partition what goes in the constructor and what doesn't ;) )

tensorflow/java/src/gen/cc/ops/op_generator.cc
karllessard: @kbsriram  is about to send another PR where an annotation processor written in Java will generate the Ops API file (from the generated operations and others found in the classpath). We planned to put this generator under src/gen/java, the reason why having a src/gen/cc here. Does it make more sense to you?

For the /ops subfolder, I was planning to split the source code of the generator in a few files so it would make sense to remove it only if you are confident that we won't have other generators written in C++ under the Java package.

tensorflow/tensorflow.bzl
asimshankar: Thanks, yeah, we should start moving them down and makes sense to start the Java one in the right place :)

tensorflow/java/src/gen/cc/ops/op_generator.cc
asimshankar: My mental model (and please correct me if I'm wrong) is that no code in `src/gen` will be part of the `.jar` distributions. The annotation processor will have to be part of it, so it makes sense for that to be in something like `src/main/java/org/tensorflow/annotations/Processor.java` or something.

If that is correct, then I don't think we need the subdirectories here now. I don't yet see the need for any other generators, but if and when they do come about, we can partition the files differently then. For now, I'd be tempted to stick with a simpler/flatter directory structure

karllessard: The annotation processor won't be part of the final jar, it will only be invoked while building it, as a Bazel `java_plugin`, and won't be accessible publicly after. The Ops file that it generates, though, will be included in the library.

So the need of keeping CC and Java generators separated is still present but we could drop off the /ops subfolder as you suggested?

asimshankar: Won't the annotations that require it appear in source code? Anyway, let's drop just the `/ops` for now. When I see @kbsriram's PR, I'll get a better idea of the annotation processor.

kbsriram: @asimshankar - in case it helps (this or the upcoming PR :-) [this is the draft of the change](https://github.com/tensorflow/tensorflow/compare/master...kbsriram:java-ops-3a-processor-bazel-draft) to kick off the annotation processor work.

tensorflow/java/src/gen/cc/op_gen_main.cc
```diff
@@ -0,0 +1,61 @@
...
+    "their appropriate package and using a maven-style directory layout.\n\n";
+
+}  // namespace op_gen
+}  // namespace tensorflow
+
+int main(int argc, char* argv[]) {
+  tensorflow::string ops_file;
+  tensorflow::string output_dir;
+  std::vector<tensorflow::Flag> flag_list = {
+    tensorflow::Flag("file", &ops_file,
```
asimshankar: The use of the word "file" here seems a bit iffy since the notion of a "file" doesn't quite exist here. It is generating code for all the operations linked in.

Instead of various magic constants embedded in the code, how about we have this tool take in the fully qualified package name as a flag instead? So something like:

- `--output`: Root directory into which output files are generated
- `--package`: Name of the package in which generated ops will be placed

Then all the magic constants can be confined to the genrule implementation, which:
(1) Adds `//tensorflow/core:` prefix to the `op_libs`
(2) Strips `_ops` from `op_libs` and slaps on `org.tensorflow.op` as a prefix for the package name

tensorflow/java/src/gen/cc/op_generator.h
```diff
@@ -0,0 +1,59 @@
...
+ public:
+  /// \brief Create a new generator, giving an environment and an
+  /// output directory path.
+  explicit OpGenerator(Env* env, const string& output_dir);
+  virtual ~OpGenerator();
+
+  /// \brief Generates wrappers for the given list of 'ops'.
+  ///
+  /// The list of operations should be issued from the library whose
+  /// file name starts with 'ops_file' (see /core/ops/*.cc).
```
asimshankar: As mentioned above, let's make this:

```
Status Run(const string& package_name, const OpList& ops);
```

and remove any notion of "files" or (`/core/ops/*.cc`) from the implementation/API/documentation of this lower level library. Doing so will make this library more amenable to inevitable future use of generating Java sources for custom operations (which may be generated in different packages as well).

asimshankar (changes requested):
Sorry for the delayed response, I didn't notice that you had made changes and the PR was ready to look at again. Next time feel free to add a comment to the conversation on the PR (since that generates a notification)

Overall, looks great, thanks for doing this and slogging through the fancy bazel rule :). My main comment is to avoid "magic" in the libraries and move all convention handling to the bazel rule (and out of the C++ libraries).

Thanks!

tensorflow/java/src/gen/gen_ops.bzl
```diff
@@ -0,0 +1,53 @@
+# -*- Python -*-
+
+load("//tensorflow:tensorflow.bzl", "tf_copts")
+
+# Given a list of "ops_libs" (a list of files in the core/ops directory
+# without their .cc extensions), generate Java wrapper code for all operations
+# found in the ops files.
+# Then, combine all those source files into a single archive (.srcjar).
```
karllessard: Exactly. It is not allowed to use wildcards (e.g. `ops/array/*.java`) in the `outs` argument of a `genrule` and since we don't know at the Bazel level which files are to be generated, we cannot list them statically neither.

Also, reading at the Java rules documentation site, it seems to be the right way doing it: 

> Source files of type .srcjar are unpacked and compiled. (This is useful if you need to generate a set of .java files with a genrule.)

tensorflow/java/src/gen/cc/op_gen_main.cc
karllessard: This sounds like a good idea. But it might lead to some other difficulties in other way. Let me explain.

The `ops_file` (or let rename it `ops_lib` or `ops_lib_name`) is not only used to produce a package name but also to extract the group name (`@Operator(group="...")`) in a near future, so it could be accessible from the `Ops` API (`ops.groupName().opName()`).

But the package and group names uses different naming conventions, the former being flat and the later in pascal/camel case. So by transforming, for instance, the `control_flow_ops` library name to `org.tensorflow.op.controlflow` in the Bazel rule, we won't be able to extract the `controlFlow` group name from it.

A solution could be to pass the `org.tensorflow.op` root package as an argument but also the group name as a different argument (`--group`), which could then be flatten to a fully qualified package name, what do you think? 

If we do so, then in addition to remove the `_op[s]` suffix to the lib name, I need to find out an easy way to transform it to camel case from its original underscore form.

asimshankar: Ah. To make sure I understand, what would the ops defined in `//tensorflow/core:control_flow_ops` result in? I believe:

- Package name: `org.tensorflow.ops.control_flow`
- Group name: `controlFlow`

Is that right?
Hmm...I can see this evolving as we go along. I guess, in the mean time perhaps  change from `--file` to something like `--name` and make it clear in the documentation that this is a snake-case name of the group of operations and is used to compute the package name?

As we make more progress, we can decide whether we want to move the package prefix out as a separate flag and/or provide additional flags for groups etc.

karllessard: I think it would be better to name the package `org.tensorflow.ops.controlflow`. This would also be in line with [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5-naming) for naming a package.

NIT: What about '--libname' for the name of the argument? And yes, I'll document it better and we can make adjustments later as we go.

asimshankar (approved):
Looks great, thanks. Looking forward to the "meat" :)

(Took the liberty to make a trivial change in the `.cc` file)

asimshankar:
Jenkins, test this please

kbsriram:

asimshankar:
This is good to merge. All tests passed (and I, the "other author" in the PR consent :)

karllessard:
Hi @asimshankar , just to make sure everything has been approved correctly for merging this PR, the previous checks are still complaining about the missing CLA. Thanks!

asimshankar:
Jenkins, test this please.

karllessard:
Do my last 2 changes really caused the tests to fail? Anything I can do to fix it?

asimshankar:
No, it shouldn't have. I believe there are some flaky tests that are being looked into. In the mean time, will try again

asimshankar:
Jenkins, test this please

asimshankar:
Jenkins, test this please

PR #11989, opened 2017-08-03
Upgrade protobuf version to fix "A protocol message was rejected because it was too big (more than 67108864 bytes)." error

yjmade:
custom package 3.1.0 still get this "A protocol message was rejected because it was too big (more than 67108864 bytes)." error, but 3.2.0 get it fixed

### DIFF
--- a/tensorflow/docs_src/install/install_linux.md
+++ b/tensorflow/docs_src/install/install_linux.md
@@ -727,13 +727,13 @@ the custom binary protobuf pip package, invoke one of the following commands:
 
   <pre>
   $ <b>pip install --upgrade \
-  https://storage.googleapis.com/tensorflow/linux/cpu/protobuf-3.1.0-cp27-none-linux_x86_64.whl</b></pre>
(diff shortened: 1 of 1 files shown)


benoitsteiner:
@martinwicke Do we support the 3.2 version of the protobuf library ? If yes we should merge this.

martinwicke:
I think we upgraded, but just missed the install doc. 

@jhseu @jart can you confirm?

jhseu:
We should delete these instructions now that the protobuf team supplies optimized builds on pypi.

jhseu:
These were for when we were building our own protobuf packages and asking users to install them for performance reasons.

yjmade:
@jhseu the problem of using official c protobuf library is it has the limitation of max size of a message is 64MB. For a exported graph, it's very easy to bigger than that size. The package provide by Tensorflow  has extend this limitation.

jhseu:
@yjmade Don't install this package. Use the pip installed protobuf package. It removes that limit and is optimized.

yjmade:
Ok , just close this.

PR #11992, opened 2017-08-03
Add SerializeTensor operator

mavenlin:
The opposite of ParseTensor, this operator takes a tensor as input and outputs a serialized TensorProto.

### DIFF
--- a/tensorflow/core/kernels/parse_tensor_op.cc
+++ b/tensorflow/core/kernels/parse_tensor_op.cc
@@ -70,8 +70,7 @@ REGISTER_KERNEL_BUILDER(Name("ParseTensor").Device(DEVICE_CPU), ParseTensorOp);
 template <typename T>
 class SerializeTensorOp : public OpKernel {
  public:
-  explicit SerializeTensorOp(OpKernelConstruction* context)
-      : OpKernel(context) {}
+  using OpKernel::OpKernel;
 
   void Compute(OpKernelContext* context) override {
     const Tensor& tensor = context->input(0);
--- a/tensorflow/core/kernels/parse_tensor_op.cc
+++ b/tensorflow/core/kernels/parse_tensor_op.cc
@@ -21,6 +21,7 @@ limitations under the License.
 #include "tensorflow/core/framework/tensor_shape.h"
 #include "tensorflow/core/framework/types.h"
 #include "tensorflow/core/lib/core/errors.h"
+#include "tensorflow/core/framework/register_types.h"
 
 namespace tensorflow {
 
@@ -64,5 +65,34 @@ class ParseTensorOp : public OpKernel {
 };
 
 REGISTER_KERNEL_BUILDER(Name("ParseTensor").Device(DEVICE_CPU), ParseTensorOp);
+
+
+template <typename T>
+class SerializeTensorOp : public OpKernel {
+ public:
+  explicit SerializeTensorOp(OpKernelConstruction* context)
+      : OpKernel(context) {}
+
+  void Compute(OpKernelContext* context) override {
+    const Tensor& tensor = context->input(0);
+    TensorProto proto;
+    if (tensor.dtype() == DT_STRING) {
+      tensor.AsProtoField(&proto);
+    } else {
+      tensor.AsProtoTensorContent(&proto);
+    }
+    Tensor* proto_string = nullptr;
+    OP_REQUIRES_OK(
+        context, context->allocate_output(0, TensorShape({}), &proto_string));
+    CHECK(proto.SerializeToString(&proto_string->scalar<string>()()));
+  }
+};
+
+#define REGISTER(T) \
+  REGISTER_KERNEL_BUILDER( \
+      Name("SerializeTensor").Device(DEVICE_CPU).TypeConstraint<T>("T"), \
+      SerializeTensorOp<T>);
+TF_CALL_ALL_TYPES(REGISTER)
+#undef REGISTER
 
 }  // namespace tensorflow
--- a/tensorflow/core/kernels/parse_tensor_test.cc
+++ b/tensorflow/core/kernels/parse_tensor_test.cc
@@ -170,7 +170,8 @@ TEST_F(SerializeTensorOpTest, SerializeTensorOpTest_uint8) {
 
 TEST_F(SerializeTensorOpTest, SerializeTensorOpTest_complex64) {
   MakeOp<complex64>(TensorShape({}), [](int x) -> complex64 {
-      return complex64{ x / 8., x / 2. };
+      return complex64{ static_cast<float>(x / 8.),
+            static_cast<float>(x / 2.) };
     });
   TF_ASSERT_OK(RunOpKernel());
   Tensor parse_output;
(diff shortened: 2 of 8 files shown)


mavenlin:
I signed it.

vrv:
This needs API-review first, and it passes API-review, then you should add tests to validate its correctness.

tensorflow/core/kernels/parse_tensor_op.cc
```diff
@@ -65,4 +66,33 @@ class ParseTensorOp : public OpKernel {
...
+template <typename T>
+class SerializeTensorOp : public OpKernel {
+ public:
+  explicit SerializeTensorOp(OpKernelConstruction* context)
+      : OpKernel(context) {}
+
+  void Compute(OpKernelContext* context) override {
+    const Tensor& tensor = context->input(0);
+    TensorProto proto;
+    if (tensor.dtype() == DT_STRING) {
```
asimshankar: Why this special casing for `DT_STRING`?

asimshankar (commented):
This seems fine to add from an API perspective. Please do add tests and the reviewer will pick it up.

Thanks!

tensorflow/core/kernels/parse_tensor_op.cc
vrv: There are comments in the codebase where people say:

// Treat DT_STRING specially, so that tensor_util.MakeNdarray in Python can
// convert the TensorProto to string-type numpy array. MakeNdarray does not
// work with strings encoded by AsProtoTensorContent() in tensor_content.

So presumably the same issue is here, but it would be good to validate this is really necessary.

tensorflow/core/kernels/parse_tensor_op.cc
```diff
@@ -65,4 +66,33 @@ class ParseTensorOp : public OpKernel {
 
 REGISTER_KERNEL_BUILDER(Name("ParseTensor").Device(DEVICE_CPU), ParseTensorOp);
 
+
+template <typename T>
+class SerializeTensorOp : public OpKernel {
+ public:
+  explicit SerializeTensorOp(OpKernelConstruction* context)
```
vrv: You might be able to do:

using OpKernel::OpKernel;

And avoid the need for the explicit constructor.

vrv (changes requested):
Looks good modulo the comment about the need to special case String.

tensorflow/core/kernels/parse_tensor_op.cc
mavenlin: This is meant to keep the serialize and parse op compatible to each other.

mavenlin:
Test code added.

tensorflow/core/kernels/parse_tensor_test.cc
```diff
@@ -0,0 +1,118 @@
...
+  void MakeOp() {
+    TF_ASSERT_OK(
+        NodeDefBuilder("myop", "SerializeTensor")
+            .Input(FakeInput(DataTypeToEnum<T>::value))
+            .Finalize(node_def()));
+    TF_ASSERT_OK(InitOp());
+  }
+};
+
+#define REGISTER_TEST(T, Shape, Mapping)                                 \
```
vrv: Thanks for adding the test!

I think it would be preferable not to use a huge macro for the entire test though:

- It's harder to know the name of the test to run because the full name is hidden behind a macro expansion
- There are two levels of macros here, which makes it harder to read.

Instead, can you do the following?

* Define each TEST_F manually with a name that is fixed
* define helper functions templated by T that does things like MakeOp, AddInput, NodeDefBuilder, etc, but keep parts that are not templated by T outside.
* define helper functions that do all of the other boiler plate not templated by T
* have each test call those helper functions with the appropriate fixed type.

mavenlin:
Would this work?

tensorflow/core/kernels/parse_tensor_test.cc
```diff
@@ -0,0 +1,236 @@
...
+    OpKernelContext ctx(&params);
+    op->Compute(&ctx);
+    TF_EXPECT_OK(status);
+    Tensor* output = ctx.mutable_output(0);
+    return *output;
+  }
+};
+
+TEST_F(SerializeTensorOpTest, SerializeTensorOpTest_half) {
+  MakeOp<Eigen::half>();
```
vrv: Thanks for doing this!  I like that you can see the names of the tests, which to me was the most important thing to have.

Now that I see the repetition more clearly, I think it might be fine to do the following to reduce it a little bit:

Add a function:

```
template <typename T>
void ConstructOp(std::function<T(int)> functor) {
  MakeOp<T>();
  AddInput<T>(TensorShape({10}), functor);
}
```
and another function called ParseSerializedOutput<T>() which does:

NodeDef parse;
code in BuildParseNodeDef
return ParseSerialized(parse, GetOutput(0));

So that each test looks like:

```
ConstructOp<float>([](int x) -> float { ... };);
TF_ASSERT_OK(RunOpKernel());
Tensor parse_output = ParseSerializedOutput<float>();
test::ExpectTensorEqual<float>(parse_putput, GetInput(0));
```

There's still some duplication, but I think it makes the code a bit more readable; the first statement indicates that you are building the op, the second says you are running it.  The third is all the code required to produce the Tensor that is the inverse of the output, and the last statement shows that you expect the expression to be reversible.

Let me know what you think of this!

mavenlin: Thanks for the suggestion! Sure, that make it more succinct, but I would like to also keep the `TensorShape` as an input parameter for `ConstructOp`, that it is possible to test with different shapes.

vrv: That sounds good too!

rmlarsen:

vrv (approved):

jhseu:
Jenkins, test this please

jhseu:
Can you update the goldens as part of the API compatibility test? See the Linux CPU Tests failure.

mavenlin:
Goldens updated.

jhseu:
Jenkins, test this please

martinwicke:
Jenkins, test this please.

martinwicke:
There seem to be build failures and I cannot rule out that this is the cause on the face of it.

mavenlin:
The failure information is quite strange, and I couldn't see the relationship to my modification.

vrv:
It looks like some clang compile error, take a look at:


bazel-out/darwin_x86_64-opt/bin/tensorflow/core/kernels/_objs/parse_tensor_test/tensorflow/core/kernels/parse_tensor_test.o): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
tensorflow/core/kernels/parse_tensor_test.cc:173:25: error: non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list [-Wc++11-narrowing]
      return complex64{ x / 8., x / 2. };
                        ^~~~~~
tensorflow/core/kernels/parse_tensor_test.cc:173:25: note: insert an explicit cast to silence this issue
      return complex64{ x / 8., x / 2. };
                        ^~~~~~
                        static_cast<float>( )
tensorflow/core/kernels/parse_tensor_test.cc:173:33: error: non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list [-Wc++11-narrowing]
      return complex64{ x / 8., x / 2. };
                                ^~~~~~
tensorflow/core/kernels/parse_tensor_test.cc:173:33: note: insert an explicit cast to silence this issue
      return complex64{ x / 8., x / 2. };
                                ^~~~~~
                                static_cast<float>( )

martinwicke:
Jenkins, test this please.

mavenlin:
Thanks for all the help!

PR #12006, opened 2017-08-03
Data type support for seq2seq attention mechanisms

ceteke:
Got an exception like the following when I was trying to use tf.float16 in my seq2seq model with Bahdanau Attention to allocate less memory on GPU:

`ValueError: Tensor conversion requested dtype float32 for Tensor with dtype float16: 'Tensor("decoder_1/BahdanauAttention/mul:0", shape=(?, ?, 2048), dtype=float16)'`

Inserted `dtype` argument to attention mechanisms that is used in Dense layer of both query layer and memory layer. Default is `tf.float32`


ceteke:
Sorry about the mistake 😄

PR #12010, opened 2017-08-03
[OpenCL] Uses Eigen implementation of asinh, acosh and atanh

lukeiwanski:
- [Eigen] Bumps to the version that supports asinh, acosh and atanh
- std:: -> numext
- simplifies SYCL kernel registration


benoitsteiner:
Jenkins, test this please.

benoitsteiner (approved):

benoitsteiner:
The code fails to compile with the following error:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/limits:546:64: error: no member named 'max_digits10' in 'std::__1::numeric_limits<Eigen::half>'
    static _LIBCPP_CONSTEXPR const int  max_digits10 = __base::max_digits10;
                                                       ~~~~~~~~^
external/eigen_archive/unsupported/Eigen/CXX11/../../../Eigen/src/Core/NumTraits.h:91:22: note: in instantiation of template class 'std::__1::numeric_limits<const Eigen::half>' requested here
    IsInteger = std::numeric_limits<T>::is_integer,
                     ^
external/eigen_archive/unsupported/Eigen/CXX11/../../../Eigen/src/Core/NumTraits.h:150:41: note: in instantiation of template class 'Eigen::GenericNumTraits<const Eigen::half>' requested here
template<typename T> struct NumTraits : GenericNumTraits<T>
                                        ^
external/eigen_archive/unsupported/Eigen/CXX11/../../../Eigen/src/Core/functors/UnaryFunctors.h:787:38: note: in instantiation of template class 'Eigen::NumTraits<const Eigen::half>' requested here
template<typename Scalar,bool iscpx=(NumTraits<Scalar>::IsComplex!=0) > struct scalar_sign_op;
                                     ^
external/eigen_archive/unsupported/Eigen/CXX11/src/Tensor/TensorBase.h:89:60: note: in instantiation of default argument for 'scalar_sign_op<const Eigen::half>' required here
    EIGEN_STRONG_INLINE const TensorCwiseUnaryOp<internal::scalar_sign_op<Scalar>, const Derived>
                                                           ^~~~~~~~~~~~~~~~~~~~~~
external/eigen_archive/unsupported/Eigen/CXX11/src/Tensor/TensorMap.h:27:112: note: in instantiation of template class 'Eigen::TensorBase<Eigen::TensorMap<Eigen::Tensor<const Eigen::half, 1, 1, long>, 16, MakePointer>, 0>' requested here
template<typename PlainObjectType, int Options_, template <class> class MakePointer_> class TensorMap : public TensorBase<TensorMap<PlainObjectType, Options_, MakePointer_> >
                                                                                                               ^
tensorflow/examples/adding_an_op/zero_out_op_kernel_2.cc:66:31: note: in instantiation of template class 'Eigen::TensorMap<Eigen::Tensor<const Eigen::half, 1, 1, long>, 16, MakePointer>' requested here
    auto input = input_tensor.flat<T>();
                              ^
tensorflow/examples/adding_an_op/zero_out_op_kernel_2.cc:61:12: note: in instantiation of member function 'ZeroOutOp<Eigen::half>::Compute' requested here
  explicit ZeroOutOp(OpKernelConstruction* context) : OpKernel(context) {}
           ^
tensorflow/examples/adding_an_op/zero_out_op_kernel_2.cc:114:27: note: in instantiation of member function 'ZeroOutOp<Eigen::half>::ZeroOutOp' requested here
TF_CALL_REAL_NUMBER_TYPES(REGISTER_KERNEL);

lukeiwanski:
Interestingly that fails to compile `scalar_sign_op` - that we did not touch. I am going to investigate more

lukeiwanski:
Further investigation points to: https://bitbucket.org/eigen/eigen/src/8d1ccfd9c5a0bf1a662bd0fe35858f4e527d2bac/Eigen/src/Core/arch/CUDA/Half.h?at=default&fileviewer=file-view-default#Half.h-516
```//static const int max_digits10 = ;``` Mac is missing the definition for it - hence the compilation error.
@ggael / @benoitsteiner why is that commented out? Could we simply set `max_digits10` to `5`? (Based on [IEEE 754](http://half.sourceforge.net/structstd_1_1numeric__limits_3_01half__float_1_1half_01_4.html#a1561594738198528e5b3c227a8ba9337) )

**Comment by rmlarsen** [comment]
@benoitsteiner any feedback for Luke?

**Comment by martinwicke** [comment]
@benoitsteiner ping

**Comment by ggael** [comment]
I added max_digits10: https://bitbucket.org/eigen/eigen/commits/da11a2b1da3f/

**Comment by lukeiwanski** [comment]
Thanks @ggael ! I will bump Eigen to the tip and give it a sping :)

**Comment by drpngx** [comment]
Jenkins, test this please.

**Comment by drpngx** [comment]
Apparently stuck somewhere. Jenkins, test this please.

**Comment by drpngx** [comment]
Here's the android error:

```
20:36:21 In file included from tensorflow/core/kernels/cwise_op_log.cc:16:
20:36:21 In file included from ./tensorflow/core/kernels/cwise_ops_common.h:27:
20:36:21 ./tensorflow/core/kernels/cwise_ops.h:50:12: error: no member named 'asinh' in namespace 'Eigen::numext'; did you mean simply 'asinh'?
20:36:21     return numext::asinh(a);
20:36:21            ^~~~~~~~~~~~~
20:36:21            asinh
20:36:21 external/androidndk/ndk/platforms/android-24/arch-x86_64/usr/include/math.h:224:8: note: 'asinh' declared here
20:36:21 double  asinh(double);
20:36:21         ^
20:36:21 In file included from tensorflow/core/kernels/cwise_op_log.cc:16:
20:36:21 In file included from ./tensorflow/core/kernels/cwise_ops_common.h:27:
20:36:21 ./tensorflow/core/kernels/cwise_ops.h:62:12: error: no member named 'acosh' in namespace 'Eigen::numext'; did you mean simply 'acosh'?
20:36:21     return numext::acosh(a);
20:36:21            ^~~~~~~~~~~~~
20:36:21            acosh
20:36:21 external/androidndk/ndk/platforms/android-24/arch-x86_64/usr/include/math.h:223:8: note: 'acosh' declared here
20:36:21 double  acosh(double);
20:36:21         ^
20:36:21 In file included from tensorflow/core/kernels/cwise_op_log.cc:16:
20:36:21 In file included from ./tensorflow/core/kernels/cwise_ops_common.h:27:
20:36:21 ./tensorflow/core/kernels/cwise_ops.h:74:12: error: no member named 'atanh' in namespace 'Eigen::numext'; did you mean simply 'atanh'?
20:36:21     return numext::atanh(a);
20:36:21            ^~~~~~~~~~~~~
20:36:21            atanh
20:36:21 external/androidndk/ndk/platforms/android-24/arch-x86_64/usr/include/math.h:225:8: note: 'atanh' declared here
20:36:21 double  atanh(double);
20:36:21         ^
```

**Comment by mehdi-goli** [comment]
Eigen implementation of asinh, acosh and atanh in Eigen can be found in  Eigen/src/Core/MathFunctions.h .
To Solve the android error take the following steps:
Firstly, all of the hyperbolic functions will be enabled if the "EIGEN_HAS_CXX11_MATH" flag is 1. So if you want to use eigen hyperbolic math, either you need to define your own macro flag for android and  change the #ifdef macros at lines 1412, 1381, and 1443 in Eigen/src/Core/MathFunctions.h to 
#if EIGEN_HAS_CXX11_MATH || [YOUR_FLAG_FOR_ANDROID].
Secondly, in the Eigen/Core file at lines 74-78 you need to add EIGEN_USING_STD_MATH(FUNC) for your own platform. See the #elif added below:
#if defined(__CUDA_ARCH__) && defined(__NVCC__)
  #define EIGEN_USING_STD_MATH(FUNC) using ::FUNC;
#elif defined([YOUR_ANDROID_PLATFORM]) 
 #define EIGEN_USING_STD_MATH(FUNC) using ::FUNC;
#else
  #define EIGEN_USING_STD_MATH(FUNC) using std::FUNC;
#endif

**Comment by lukeiwanski** [comment]
@benoitsteiner / @rmlarsen / @drpngx we are currently It seems like additional development for android is required - and at present we are not able to test that target. If that is OK with you I am tempted to do something like: 
```
#ifndef __ANDROID__
    return numext::atanh(a);
#else
    return std::atanh(a);
#endif  // __ANDROID__
```
What do you think?

**Comment by benoitsteiner** [comment]
Using #ifdef EIGEN_HAS_CXX11_MATH would make the code much more portable since atanh, asinh nd acosth are available in Eigen iff EIGEN_HAS_CXX11_MATH is defined

**Comment by lukeiwanski** [comment]
@benoitsteiner done in 8e42815

**Comment by sb2nov** [comment]
Jenkins, test this please.

**Comment by sb2nov** [comment]
Previous run had tests timeout.

Jenkins, test this please.

**Comment by lukeiwanski** [comment]
Windows fails seems to be unrelated.

@benoitsteiner / @rmlarsen / @drpngx it seems that changes suggested by @mehdi-goli on Eigen side are needed. As mentioned earlier we are not able of testing this target at present.
How should we proceed with this?

**Review by benoitsteiner — APPROVED** [review]

**Comment by benoitsteiner** [comment]
Jenkins, test this please.

File: tensorflow/core/kernels/cwise_ops.h [hunk 1]
```diff
@@ -47,7 +47,11 @@ template <typename T>
 struct scalar_asinh_op {
   EIGEN_EMPTY_STRUCT_CTOR(scalar_asinh_op)
   EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE const T operator()(const T& a) const {
+#ifdef EIGEN_HAS_CXX11_MATH
```
benoitsteiner: This should be #if EIGEN_HAS_CXX11_MATH to be consistent with Eigen and allow the code to compile on Windows and Android platforms.

benoitsteiner (dismissed):

tensorflow/core/kernels/cwise_ops.h
drpngx: Done

drpngx:
Jenkins, test this please.

martinwicke (approved):
@drpngx has addressed @benoitsteiner's comments.

martinwicke:
The Py3 pip failure is unrelated to this.

PR #11952, opened 2017-08-01
Make plugin_data optional instead of repeated

chihuahua:
Every summary op writes data for a single plugin to process. Hence, each
SummaryMetadata proto should have a single PluginData optional field
(instead of a repeated one). This removes much complexity from
TensorBoard logic that loops over the plugin data. It also simplifies
the SQL schema - it can now enforce a one-to-one relationship between
summary op and plugin.

PR #12012, opened 2017-08-03
Make plugin_data optional instead of repeated

chihuahua:
This patches tensorflow/tensorflow#11952 into master. Here is that PR's description:

Every summary op writes data for a single plugin to process. Hence, each
SummaryMetadata proto should have a single PluginData optional field
(instead of a repeated one). This removes much complexity from
TensorBoard logic that loops over the plugin data. It also simplifies
the SQL schema - it can now enforce a one-to-one relationship between
summary op and plugin.

### DIFF
--- a/tensorflow/core/debug/debug_grpc_testlib.cc
+++ b/tensorflow/core/debug/debug_grpc_testlib.cc
@@ -56,17 +56,15 @@ namespace test {
 
       // Obtain the device name, which is encoded in JSON.
       third_party::tensorflow::core::debug::DebuggerEventMetadata metadata;
-      for (int i = 0; i < val.metadata().plugin_data_size(); i++) {
-        if (val.metadata().plugin_data(i).plugin_name() != "debugger") {
-          // This plugin data was meant for another plugin.
-          continue;
-        }
-        auto status = tensorflow::protobuf::util::JsonStringToMessage(
-            val.metadata().plugin_data(i).content(), &metadata);
-        if (status.ok()) {
-          // The device name has been determined.
-          break;
-        }
+      if (val.metadata().plugin_data().plugin_name() != "debugger") {
+        // This plugin data was meant for another plugin.
+        continue;
+      }
+      auto status = tensorflow::protobuf::util::JsonStringToMessage(
+          val.metadata().plugin_data().content(), &metadata);
+      if (!status.ok()) {
+        // The device name could not be determined.
+        continue;
       }
 
       device_names.push_back(metadata.device());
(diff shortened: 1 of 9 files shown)


tensorflow/core/debug/debug_grpc_testlib.cc
```diff
@@ -56,17 +56,15 @@ ::grpc::Status TestEventListenerImpl::SendEvents(
...
-        }
-        auto status = tensorflow::protobuf::util::JsonStringToMessage(
-            val.metadata().plugin_data(i).content(), &metadata);
-        if (status.ok()) {
-          // The device name has been determined.
-          break;
-        }
+      if (val.metadata().plugin_data().plugin_name() != "debugger") {
+        // This plugin data was meant for another plugin.
+        continue;
```
chihuahua: After creating a CL and running internal tests, I realized that this logic should continue instead of returning a GRPC canceled status. Hence, this line differs from its counterpart in tensorflow/tensorflow#11952.

tensorflow/core/debug/debug_grpc_testlib.cc
```diff
@@ -56,17 +56,15 @@ ::grpc::Status TestEventListenerImpl::SendEvents(
...
+      }
+      auto status = tensorflow::protobuf::util::JsonStringToMessage(
+          val.metadata().plugin_data().content(), &metadata);
+      if (!status.ok()) {
+        // The device name has been determined.
+        break;
```
chihuahua: Same here.

caisq:

chihuahua:

chihuahua:

tensorflow/core/debug/debug_grpc_testlib.cc
```diff
@@ -56,17 +56,15 @@ ::grpc::Status TestEventListenerImpl::SendEvents(
...
-          break;
-        }
+      if (val.metadata().plugin_data().plugin_name() != "debugger") {
+        // This plugin data was meant for another plugin.
+        continue;
+      }
+      auto status = tensorflow::protobuf::util::JsonStringToMessage(
+          val.metadata().plugin_data().content(), &metadata);
+      if (!status.ok()) {
+        // The device name could not be determined.
```
jart: @caisq We might want to log a warning here here. However we shouldn't make that change in this PR.

jart (approved):

tensorflow/core/debug/debug_grpc_testlib.cc
chihuahua: +1, thank you, Justine!

caisq:

chihuahua:
I am canceling this PR and will make these changes internally instead.

PR #12026, opened 2017-08-04
Add loss to summary during estimator.train()

terrytangyuan:
This would be useful for visualizing the losses during training phase.


ispirmustafa (changes requested):
Thank you Yuan!
writting summary every step is slower. We expect user will add a summary in their model-fn related to loss.

terrytangyuan:
@ispirmustafa Thanks! What about the canned estimators? There doesn't seem to be an easy way to access the state/graph/session of the estimator via run hooks either.

terrytangyuan:
@ispirmustafa Thanks! I am going to close this one since we can do this via a run hook.

PR #12050, opened 2017-08-04
Fix debugger logic in r1.3

chihuahua:
In tensorflow/tensorflow#11952, I had set made some logic within debug_grpc_testlib return too early, breaking some debugger-related behavior. This PR fixes that.


chihuahua:

jart (approved):

jart:
Jenkins test this please

caisq:
The Linux GPU build is failing in an obscure way. Testing again

caisq (approved):

caisq:
Ah - the GPU build failure can be ignored. It is due to an unrelated test tag change.

caisq:
@chiahuahua, does this change need to be commit to the master branch as well?

chihuahua:
@caisq, tensorflow/tensorflow#12012 already makes this change in master. Thanks for starting tests. Let me try again.

PR #12075, opened 2017-08-07
add args of TimeReversedFusedRNN

halleywj:
I want to use TimeReversedFusedRNN together with LSTMBlockFusedCell , but the batch_axis of LSTMBlockFusedCell's input is 1 , but not 0 . But the reverse function of TimeReversedFusedRNN can't decide which is batch_axis , just do array_ops.reverse_sequence(t, lengths, 0, 1) . So , I think it's useful that users can tell it which is batch_axis and seq_axis .

### DIFF
--- a/tensorflow/contrib/rnn/python/ops/fused_rnn_cell.py
+++ b/tensorflow/contrib/rnn/python/ops/fused_rnn_cell.py
@@ -146,7 +146,7 @@ class TimeReversedFusedRNN(FusedRNNCell):
   def __init__(self, cell):
     self._cell = cell
 
-  def _reverse(self, t, lengths):
+  def _reverse(self, t, lengths, time_major):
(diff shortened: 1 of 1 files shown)


halleywj:
I signed it!

rmlarsen:
@ebrevdo feel free to delegate

ebrevdo:
Isn't seq_axis just 1-batch_axis?

halleywj:
yes, that's true.So should I recommit?

halleywj:
I think user can decide which is seq_axis and batch_axis is useful.

ebrevdo:
Please remove these two arguments and instead provide the single bool
argument time_major, just like tf.nn.dynamic_rnn does.  The default value
should be such that the existing behavior does not change.

halleywj:
Thanks ， I've recommited.

PR #12087, opened 2017-08-07
Update variables.py

alanyee:
Added deprecated message


alanyee:
I have been looking at the core training framework. It seems that the other functions such as `get_global_step` are also deprecated. Should I add the deprecated comments for them too?

rmlarsen:
@martinwicke should we add deprecation statements for all these function?

martinwicke:
Yes. For every function for which a direct equivalent exists outside of contrib, we should deprecate the contrib version.

Please (also in this PR) set the date to `None`. Timelines confuse people, especially if they're in the past.

This may lead to spurious warnings, which we'll have to fight with the `with deprecation.silence` decorator, but we can do that in a follow-up.

rmlarsen (approved):

rmlarsen:

Issue #11713, opened 2017-07-24
decode_csv is extremely slow


sgirkin:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
I tested on both MacOS 10, Centos 7, https://pastebin.com/LsADQ89s

- **TensorFlow installed from (source or binary)**:
Tensorflow 1.2.1, also tried with tensorflow compiled from sources

- **TensorFlow version (use command below)**:
('v1.2.0-5-g435cdfc', '1.2.1'), also compiled and tested from master

- **Python version**: 
2.7.13

- **CUDA/cuDNN version**:
no

- **GPU model and memory**:
no

- **Exact command to reproduce**:
see above

### The problem
I noticed that a huge amount of training time CPU is idle. I isolated the problem to the code snippet bellow, where I create a string tensor using tf.constant with N rows each of them contains comma separated list of randomly generated numberical values, then I decode these string tensor to rank2 tensor of floats using 3 different approaches. 

I discovered following:
decoding tfrecords with *tf.parse_example* is quite fast
decoding csv with *decode_csv* is ~20 (4 cores) 60 (16 cores) times slower than parse_example and it utilises only 1 core on my machine
decoding csv with *tf.string_split* + *tf.string_to_number()* is ~20-60 times slower than parse_example. 
It seems *tf.string_to_number()* makes it slow, if I comment out the line with *tf.string_to_number* it is 2-3 times slower than parse example

The code I used for that:
https://pastebin.com/ZR152y5M

### Source code / logs
Running parse_example
Total time: 2.01925897598

Running string_split
Total time: 5.52524709702

Running decode_csv
Total time: 22.6595089436

reedwm:
It makes sense for `tf.decode_csv` to be slower than `tf.parse_example`, since `tf.train.Example`s are designed to be decoded fast while CSVs are not. When training, I would recommend first converting your data into `tf.train.Example`s.

If you find that TensorFlow decodes CSVs significantly slower than some other CSV decoder implementation, then reopen this issue.

sgirkin:
Hi,

I implemented decoding csv using _tf.py_func_ and numpy (test_numpy_csv function in the code below). It seems even reading with numpy is faster than decode_csv. 

https://pastebin.com/rTZjCwGH

So far parsing csv using py_func+numpy is ~5 times faster than using decode_csv on this particular synthetic data (1000 examples with 100 float columns)

reedwm:
[...]

PR #12102, opened 2017-08-08
Speed up safe_strtod and safe_strtof functions by using double-conversion library

akindyakov:
#11713


AKindyakov:
@rmlarsen, performance improvement is so attractive :)
There is benchmark: https://gist.github.com/AKindyakov/2c528beccb11a24c1e6ec02ec12df47e
DC is significantly faster.

rmlarsen:
@AKindyakov thanks. That is indeed a nice speedup. 
@martinwicke @vrv what are your opinions? Should we introduce this new dependency or try to speed up decode_csv in another way?

rmlarsen:
We will investigate and see what direction to take.

drpngx:
@rmlarsen any decision on this?

I think we might be open sourcing the google library, which is probably faster than the TF re-implementation.

rmlarsen:
@drpngx I have not had time to pursue this.

AKindyakov:
JFYI patch about case insensitivity for special values was accepted and merged into the master branch.
https://github.com/google/double-conversion/pull/47

drpngx:
Oh, I hadn't realized this was from google. @martinwicke wdyt about taking on this additional dependency?

martinwicke:
I have no problem with it. Seems lightweight enough. We should double-check whether it supports all platforms, but I see MSVC files there, so I presume the answer is yes.

third_party/double-conversion/add_case_insensitive_nan.patch
```diff
@@ -0,0 +1,188 @@
+diff --git a/double-conversion/double-conversion.cc b/double-conversion/double-conversion.cc
+index 6f21a01..835e426 100644
+--- a/double-conversion/double-conversion.cc
++++ b/double-conversion/double-conversion.cc
+@@ -26,6 +26,7 @@
+ // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
drpngx: Please undo, as per comment, case insensitivity should already be there.

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -384,5 +384,13 @@ bool SplitAndParseAsFloats(StringPiece text, char delim,
                                     result);
 }
 
+size_t Strnlen(const char* str, size_t strsz) {
```
drpngx: `const size_t string_max_len` instead of `strsz`

drpngx (dismissed):

sb2nov:
@AKindyakov can you look at the merge conflict and the PR comments

AKindyakov:
Yes, sure, I'll do it this night.

rmlarsen:
I honestly don't have time to look at this right now, unassigning myself. @sb2nov could you re-assign, please?

sb2nov:
@AKindyakov there is a merge conflict. Can you take a look.

AKindyakov:
@sb2nov, it's done

sb2nov:
Jenkins, test this please.

AKindyakov:
Jenkins, test this, please.

AKindyakov:
@sb2nov, I've fixed build and test problems. May I ask you to run full tests set for it?

drpngx:
Jenkins, test this please.

AKindyakov:
I've fixed makefile build and have fixed up dependencies lists in ``tensorflow/core``.
May I ask you to run full tests on it?

AKindyakov:
Fixed problems with sanity check

AKindyakov:
Can I ask someone to run the full test set for it?

cwhipkey:
jenkins, test this please

tensorflow/core/BUILD
```diff
@@ -1092,6 +1096,7 @@ cc_library(
     deps = [
         ":protos_all_cc_impl",
         "//third_party/eigen3",
+        "@double_conversion//:double-conversion",
```
cwhipkey: It's a little unfortunate that mobile builds will end up using it, when I assume they don't need the extra speed of parsing (since they will only be parsing a few attribute values, not large files of text).  Do you know how large the double-conversion library is when compiled for android?

(one option if the net change in binary size would be too large is to have the double conversion library only used by the op that needs it for speed).

tensorflow/core/lib/strings/numbers.cc
```diff
@@ -165,8 +112,8 @@ char* DoubleToBuffer(double value, char* buffer) {
     // larger than the precision we asked for.
     DCHECK(snprintf_result > 0 && snprintf_result < kFastToBufferSize);
 
-    full_precision_needed =
-        locale_independent_strtonum<double>(buffer, nullptr) != value;
+    auto parsed_value = double{};
```
cwhipkey: why this change?  (also, we could possibly use DoubleToStringConverter to implement all of DoubleToBuffer?)  (but see the question about android code size as well).

tensorflow/core/lib/strings/numbers_test.cc
```diff
@@ -278,6 +279,30 @@ TEST(safe_strtof, Float) {
   EXPECT_EQ(-42.0f, result);
```
cwhipkey: maybe test case of leading and trailing whitespace as well

tensorflow/core/lib/strings/numbers.cc
```diff
@@ -32,72 +35,16 @@ namespace tensorflow {
...
-        (s.fail() ? static_cast<std::iostream::pos_type>(0)
-                  : (s.eof() ? static_cast<std::iostream::pos_type>(strlen(str))
-                             : s.tellg()));
-  }
-  return result;
+static inline const double_conversion::StringToDoubleConverter& StringToFloatConverter() {
+    const static double_conversion::StringToDoubleConverter converter(
+        double_conversion::StringToDoubleConverter::ALLOW_LEADING_SPACES
+        | double_conversion::StringToDoubleConverter::ALLOW_HEX
+        | double_conversion::StringToDoubleConverter::ALLOW_OCTALS
```
cwhipkey: do we need to support OCTALS?

tensorflow/core/BUILD
AKindyakov: It's not so large actually. The size of the compiled static library is 486KiB on my laptop.

tensorflow/core/lib/strings/numbers.cc
AKindyakov: I don't know. Should I remove it for the sake of nice little interface?

cwhipkey: sgtm, since we didn't support before I'd rather not add support if we don't really need it.

tensorflow/core/lib/strings/numbers.cc
AKindyakov: The previous implementation of DoubleToBuffer function was using old locale_independent_strtonum<double> which is replaced by double_conversion::StringToDoubleConverter with slightly differen interface.

But it is a good point to use only one function to convert str to double everywhere here. Consider double_conversion::StringToDoubleConverter only as hidden implementation details of safe_strto{d,f}.

This is a good idea to use double-conversion for {Double,Float}ToBuffer functions. I can do it here or in another pull request.

tensorflow/core/lib/strings/numbers_test.cc
AKindyakov: Sure, I'll do it. Thank you.

tensorflow/core/BUILD
AKindyakov: Stripped binary build with optimisation level 3 has size 77KiB. By the way, I can figure out what can we throw out from the library in the android build.

tensorflow/core/lib/strings/numbers.cc
AKindyakov: ok, I'll remove it.

AKindyakov:
@cwhipkey I fixed up ```CMake``` build, remove octal number support and add test cases with trailing and leading characters. Could you please run tests on it?

I was looking for a code in double conversion library which we can throw out in the android build. But this library is very solid and there is nothing to cut off.

cwhipkey:
+ @petewarden  - pete, this will increase binary size some on  mobile (maybe not the full 77k, but it will probably be an increase).  What do you think we should do?   Some options are:

1. use ifdef to use the old code on mobile.  This clearly would cause no binary increase on mobile, but makes another code path that will be less tested and it's a bit unfortunate to have 2 code paths.

2. similar to 1., but do it in a way that can be tested off android.

3. add in the performant library and code only use by the kernels, but not in the string library.  This seems not great, since you would need to call the fast library from the kernels, instead of just using the string library.

4. test more fully and maybe it won't be a big change, or we can find savings elsewhere.

samtaylor159:
Is there any update on this? This issue seems to be a current bottleneck in my code, it adds a huge overhead when reading my data.

cwhipkey:
There is a pull request for it:
https://github.com/tensorflow/tensorflow/pull/12102

petewarden (approved):
Sorry for the delay! We may end up figuring out how to avoid the extra dependency on mobile as an option after this goes in, but I don't want that to block this change, so approving from my side.

cwhipkey (approved):

martinwicke:
Jenkins, test this please.

martinwicke:
There are still problems with the Windows build, relevant error:

```
17:22:02 LINK : fatal error LNK1181: cannot open input file 'double_conversion\src\double_conversion\double-conversion\double-conversion.lib' [C:\tf_jenkins\home\workspace\tensorflow-pr-win-cmake-py\cmake_build\proto_text.vcxproj]
```

martinwicke:
@mrry, do you know how to fix this?

mrry:
At a guess, maybe the static library has a different filename on Windows? Maybe there's a missing `lib/` in the definition of `double_conversion_STATIC_LIBRARIES`?

AKindyakov:
Is it me suppose to fix this?
I don't have any WIN machine for the test.

drpngx:
@AKindyakov yes usually the committer ensures that everything builds, with our help. For this one, I pushed a commit that I think should fix it.

drpngx:
@AKindyakov yes usually the committer ensures that everything builds, with our help. For this one, I pushed a commit that I think should fix it.

drpngx:
@AKindyakov yes usually the committer ensures that everything builds, with our help. For this one, I pushed a commit that I think should fix it.

AKindyakov:
@drpngx thanks a lot!

drpngx:
Looks like there's a sycl breakage at `master`. Let's wait.

martinwicke:
Jenkins, test this please.

AKindyakov:
@martinwicke, @drpngx have you any idea what is wrong with the build. There are no errors in build log only warnings which are not related to my changes at all.

drpngx:
The mac test failed with a timeout.

The CMake failed with the same error as before: 
```
22:31:05 LINK : fatal error LNK1181: cannot open input file 'double_conversion\src\double_conversion\double-conversion\lib\double-conversion.lib' [C:\tf_jenkins\home\workspace\tensorflow-pr-win-cmake-py\cmake_build\proto_text.vcxproj]
```
The other errors don't appear to be related. Let's try again.

AKindyakov:
@martinwicke, @drpngx, I believe I fixed up windows build. May I ask you to test it, please? (sorry for the delay - busy time)

martinwicke:
Jenkins. Test this please.

sb2nov:
@martinwicke is there a way to fix the CLA bot here. It was approved yes but the check seems to be pending.

martinwicke:
It looks to me like it is stuck. I will override. I guess the XLA tests are known to be broken?

@willnorris FYI, not sure how to get it out of its funk.

willnorris:
There's no obvious reason that I can see that would have caused this to get stuck, but I've manually approved it and will look into the cause.

AKindyakov:
Thanks!

caisq:
Just as a heads-up: This contribution is causing difficulties when the code is sync'ed to Google internal. We are considering reverting it first to unblock the code sync'ing, and maybe finding a way to make a similar change in internal.

cc @gunan

PR #12104, opened 2017-08-08
Ensure that bazel versions of the form X.Y.Z-mmmmm work correctly

DavidNorman:
The code in configure.py doesn't allow for bazel versions which identify themselves as X.Y.Z-mmmmm.  This is true for bazel on OS/X installed through homebrew.

Recently some code that checks the version number has appeared, and this is causing configurations built with OS/X homebrew bazel to fail.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -384,12 +384,16 @@ def set_action_env_var(environ_cp,
 def convert_version_to_int(version):
   """Convert a version number to a integer that can be used to compare.
 
+  Version strings of the form X.YZ and X.Y.Z-xxxxx are supported. The
+  'xxxxx' part, for instance 'homebrew' on OS/X, is ignored.
+
   Args:
-    version: a version to be covnerted
+    version: a version to be converted
 
   Returns:
     An integer if converted successfully, otherwise return None.
   """
+  version = version.split('-')[0]
   version_segments = version.split('.')
   for seg in version_segments:
     if not seg.isdigit():


configure.py
```diff
@@ -384,12 +384,16 @@ def set_action_env_var(environ_cp,
 def convert_version_to_int(version):
   """Convert a version number to a integer that can be used to compare.
 
+  Version strings of the form X.YZ and X.Y.Z-xxxxx are supported. The
```
rmlarsen: Should we check that the leading parts are not zero? Otherwise the version number will not be unique, e.g. 0.23 and 0.0.23 would results in the same integer.

rmlarsen (changes requested):

configure.py
DavidNorman: The current versions of bazel do have a leading zero - mine is '0.5.2-homebrew'.

I know what you mean though - In the past I have solved this problem by converting X.Y.Z into X.YZ, where Y and Z are formatted %04u.  

I'm not really trying to solve that problem though.  I could raise an issue for that if you would like.

DavidNorman: perhaps we could just stick a '1' onto the front before converting the digits to an int.  The numbers which are generated are never displayed so it wouldn't be an issue.

rmlarsen: SGTM

rmlarsen: SGTM. Feel free to file an issue separately, if you don't want to add it in this PR.

rmlarsen:

configure.py
DavidNorman: Indeed - I think this should just be merged.

If the bazel rev number is always 3 fields, then there isn't a problem, and i've not seen any examples where it isn't.

rmlarsen (approved):

PR #12111, opened 2017-08-08
Added a bias term in Bahdanau attention alignments

rubenvereecken:
The original paper by Bahdanau (https://arxiv.org/abs/1409.0473) lists a bias term (Eq. 18) and Tensorflow's normalized Bahdanau implementation uses one, so I don't see why we shouldn't use one in the original implementation.

Edit: Err, I was wrong, staring at papers too much all day, and quoted an equation from the paper that introduces Bahdanau energy normalization (http://arxiv.org/abs/1704.00784). Still, the original may have used it too as they state they "omit bias terms to make the equations less cluttered". My bad - the reviewer decides what to do with this I guess.

### DIFF
--- a/tensorflow/contrib/seq2seq/python/ops/attention_wrapper.py
+++ b/tensorflow/contrib/seq2seq/python/ops/attention_wrapper.py
@@ -429,22 +429,23 @@ def _bahdanau_score(processed_query, keys, normalize):
   processed_query = array_ops.expand_dims(processed_query, 1)
   v = variable_scope.get_variable(
       "attention_v", [num_units], dtype=dtype)
+  # Bias added prior to the nonlinearity
+  b = variable_scope.get_variable(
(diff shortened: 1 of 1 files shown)


jhseu:
@lmthang Mind taking a look?

drpngx:
@lmthang any luck with this?

sb2nov:
@lmthang ping

lmthang:
Hi guys,

Sorry for a delay on this. 

There was no bias term in the original paper by Baudanau https://arxiv.org/pdf/1409.0473.pdf (see page 14). This change will break existing models that uses "bahdanau" attention, so I will oppose this change.
(The author of this PR also acknowledged that he/she misread the paper).

If people would like to have bias in the attention, I'd recommend using normed_bahdanau instead. See https://github.com/tensorflow/nmt#hyperparameters.

-Thang

ebrevdo:
Closing...

PR #12122, opened 2017-08-08
Add name field for tf.contrib.data.Iterator.make_initializer operation

### LINKED ISSUE #12083: tf.contrib.data.Iterator.make_initializer operation lacks name
I'm using tf.contrib.data for my input pipeline. Since it feeds the model without placeholders, I need to be able to reload both the model and input pipeline in order to resume training later. However, the make_initializer method of tf.contrib.data.Iterator does not have a name argument like most (all?) other Tensorflow operations, making it impossible (or at best, error-prone) to find these ops in the reloaded graph. Unless I'm mistaken, this seems like a standard use-case and the missing name argument is inconsistent with the rest of Tensorflow.

yongtang:
This fix adds name fields for tf.contrib.data.Iterator.make_initializer operation

This fix fixes #12083.

### DIFF
--- a/tensorflow/contrib/data/python/ops/dataset_ops.py
+++ b/tensorflow/contrib/data/python/ops/dataset_ops.py
@@ -257,11 +257,12 @@ class Iterator(object):
       # initializers that simply reset their state to the beginning.
       raise ValueError("Iterator does not have an initializer.")
 
-  def make_initializer(self, dataset):
+  def make_initializer(self, dataset, name=None):
     """Returns a `tf.Operation` that initializes this iterator on `dataset`.
 
     Args:
       dataset: A `Dataset` with compatible structure to this iterator.
+      name: (Optional.) A name for the created operation.
 
     Returns:
       A `tf.Operation` that can be run to initialize this iterator on the given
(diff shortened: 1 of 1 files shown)


tensorflow/contrib/data/python/ops/dataset_ops.py
```diff
@@ -271,22 +272,25 @@ def make_initializer(self, dataset):
...
-            (self._output_types, dataset.output_types))
-    for iterator_shape, dataset_shape in zip(
-        nest.flatten(self._output_shapes), nest.flatten(dataset.output_shapes)):
-      if not iterator_shape.is_compatible_with(dataset_shape):
-        raise TypeError("Expected output shapes compatible with %r but got "
-                        "dataset with output shapes %r." %
-                        (self._output_shapes, dataset.output_shapes))
-    return gen_dataset_ops.make_iterator(dataset.make_dataset_resource(),
-                                         self._iterator_resource)
+    with ops.name_scope(name, "make_initializer", [dataset]) as name:
```
mrry: The `[dataset]` argument has no effect here (because it isn't a `tf.Operation` or `tf.Tensor`, so you should remove it.

mrry (changes requested):
Thanks for the change! Just one minor nit...

tensorflow/contrib/data/python/ops/dataset_ops.py
yongtang: Thanks. Done.

yongtang:
Thanks @mrry for the review. The PR has been updated.

mrry (approved):
Looks good to me!

mrry:

PR #12125, opened 2017-08-09
Add `skip_empty` option to `tf.string_split`

### LINKED ISSUE #12108: tf.string_split didn't behave the same as that split in python
In python, if we split the following string `'a#'.split('#')`, we will get a list of two elements: `['a', '']`.
While in tensorflow, if we use `tf.string_split`, we will get an sparsetensor with only one elements. 
For reproduction, you can run the following piece of code
```
a = tf.constant("#2")
b = tf.string_split([a], "#")
with tf.Session() as ss:
    ss.run(tf.local_variables_initializer())
    print(ss.run(b))
```
As a result, when we use `tf.string_split`, we don't know whether the `2` is before or after the `#`. I think such behavior is undesirable since we usually have missing values when decoding csv files.

yongtang:
This fix add `skip_empty` to `tf.string_split` so that it is possible to have `tf.string_split` behaving similiarly with python.

This fix fixes #12108.

### DIFF
--- a/tensorflow/core/kernels/string_split_op.cc
+++ b/tensorflow/core/kernels/string_split_op.cc
@@ -28,9 +28,13 @@ namespace tensorflow {
 
 namespace {
 
-std::vector<string> Split(const string& str, const string& delimiter) {
+std::vector<string> Split(const string& str, const string& delimiter,
+                          const bool skipEmpty) {
   if (!delimiter.empty()) {
-    return str_util::Split(str, delimiter, str_util::SkipEmpty());
+    if (skipEmpty) {
+      return str_util::Split(str, delimiter, str_util::SkipEmpty());
+    }
+    return str_util::Split(str, delimiter);
   }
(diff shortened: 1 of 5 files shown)


rmlarsen:

asimshankar:
(Looks good for API change, removing the label)

rmlarsen:
@asimshankar Thanks!

rmlarsen:

yongtang:
@rmlarsen @asimshankar Thanks for the review. All Jenkins tests passed except `//tensorflow/tools/api/tests:api_compatibility_test`

The PR has now been updated with 
```
bazel-bin/tensorflow/tools/api/tests/api_compatibility_test           --update_goldens True
```

Please take a look.

rmlarsen (approved):

rmlarsen:
@yongtang thanks for the contribution!

PR #12127, opened 2017-08-09
tf.reshape accepts Dimension objects for the shape parameter

### LINKED ISSUE #11974: tf.reshape does not accept Dimension objects for the shape parameter

------------------------

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10 x64
- **TensorFlow version (use command below)**:  1.2.1
- **Python version**: 3.6.1 (Anaconda 4.4.0 64-bit)
- **CUDA/cuDNN version**: 
- **GPU model and memory**: GTX 780
- **Exact command to reproduce**: tf.reshape()

### Describe the problem
Describe the problem clearly here. Be sure to convey here why it's a bug in TensorFlow or a feature request.

tf.reshape does not accept a list with mixed Integer and Dimension() as elements for the shape parameter. It should accept shapes that have Dimension as elements since tensor shapes consist of dimensions. Specifically, tf.tensor.shape returns a list of Dimensions, therefore using a similar object to specify a shape in tf.reshape should not cause an error.

### Source code / logs

Example code:

```
X = tf.placeholder("float", [1,784])
X = tf.reshape(X, [1, X.shape[-1]])
```

results in

```
TypeError: Failed to convert object of type <class 'list'> to Tensor. Contents: [1, Dimension(784)]. Consider casting elements to a supported type.
```

This is rectified by casting X.shape[-1] to int before passing to tf.reshape.

A similar unrelated issue is that X.shape returns (784,) while it should return (1,784). This requires reshaping to turn the placeholder back into a 2D tensor. I haven't determined if this is a bug but it occurs when the tensor is explicitly specified as a 2D tensor so it is probably worth changing.

facaiy:
The PR aims to fix #11974.

Because it's my first contribution, so the PR is opened early to get feedback from community.

Cast tf.Dimension to int for shape argument.

+ [x] add a doctest.

### DIFF
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -2461,8 +2461,12 @@ def reshape(tensor, shape, name=None):
   Returns:
     A `Tensor`. Has the same type as `tensor`.
   """
-  shape = tensor_shape.as_shape(shape)
-  return gen_array_ops.reshape(tensor, shape, name)
+  if hasattr(shape, "__len__") and len(shape) <= 8:
+    for i in range(len(shape)):
+      dim = shape[i]
+      shape[i] = int(dim) if isinstance(dim, tensor_shape.Dimension) else dim
+
+  return gen_array_ops._reshape(tensor, shape, name)
 # pylint: enable=redefined-builtin,protected-access
 
 
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -2463,10 +2463,8 @@ def reshape(tensor, shape, name=None):
     A `Tensor`. Has the same type as `tensor`.
   """
   if hasattr(shape, "__len__") and len(shape) <= 8:
-    for i in range(len(shape)):
-      dim = shape[i]
-      shape[i] = int(dim) if isinstance(dim, tensor_shape.Dimension) else dim
-
+    shape = [int(dim) if isinstance(dim, tensor_shape.Dimension) else dim
+                 for dim in shape]
   return gen_array_ops._reshape(tensor, shape, name)
 # pylint: enable=redefined-builtin,protected-access
 
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -2389,7 +2389,7 @@ def where(condition, x=None, y=None, name=None):
 
 # pylint: disable=redefined-builtin,protected-access
 def reshape(tensor, shape, name=None):
-  r"""Reshapes a tensor.
+  """Reshapes a tensor.
 
   Given `tensor`, this operation returns a tensor that has the same values
   as `tensor` with shape `shape`.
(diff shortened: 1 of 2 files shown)


yaroslavvb:
Sorry I don't understand the motivation
compat.as_bytes(x.__str__()) instead of compat.as_bytes(x)

what does this improve?

facaiy:
@yaroslavvb , the PR is modified as suggested by @pmccarter . But I am not sure whether it is implemented correctly, could you give more feedback? Thanks.

rmlarsen:

rmlarsen:
@facaiy As you can see here: https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu/6320/consoleFull
this breaks a large number of tests. I'll let @martinwicke give detailed comments.

facaiy:
@rmlarsen @martinwicke  It'll be really useful for if I can check test.log of CI. Because my laptop is left on office, hence I'll check it later. Thanks very much!

facaiy:
As `reshape` is migrated from gen_array_ops to array_ops, I guess that many unit tests will be broken because of import error.

tensorflow/python/ops/array_ops.py
```diff
@@ -2386,6 +2387,90 @@ def where(condition, x=None, y=None, name=None):
     raise ValueError("x and y must both be non-None or both be None.")
 
 
+# pylint: disable=redefined-builtin,protected-access
```
facaiy: Is `protected-access` right?

facaiy:

facaiy:
Hi, @rmlarsen . I submit a fix, and could you help retest the PR? Thanks.

tensorflow/python/ops/array_ops.py
```diff
@@ -2386,6 +2387,88 @@ def where(condition, x=None, y=None, name=None):
...
+  Args:
+    tensor: A `Tensor`.
+    shape: A `Tensor`. Must be one of the following types: `int32`, `int64`.
+      Defines the shape of the output tensor.
+    name: A name for the operation (optional).
+
+  Returns:
+    A `Tensor`. Has the same type as `tensor`.
+  """
+  if hasattr(shape, "__len__") and len(shape) <= 8:
```
facaiy: Should we raise an exception when `len(shape) > 8`?

facaiy:
By the way, since `reshape` is migrated from `_gen_array_op` to `array_op`, should we remove the duplicate document of `_gen_array_op._reshape`? If I understand correctly, the description is in the [tensorflow/core/ops/ops.pbtxt](https://github.com/tensorflow/tensorflow/blob/ca1f9846a1fa6056c28b6512e3dd1037132ef7af/tensorflow/core/ops/ops.pbtxt#L19593)

yaroslavvb:
@facaiy The docs are inside cc file which defines reshape op.

To think of it, this approach would trigger an API review. I wonder if there's a simple solution which would only touch a couple of lines of Python code, better chance of being accepted without a lot of work on maintaining team. BTW, the best way to test things is to run all relevant tests locally first, using "bazel test"

facaiy:
Thanks for your explanation, @yaroslavvb . 

I agree that migrating `reshape` will introduce duplicate code and increase maintain cost, so I'd like to try to fix it in `tensor_util.make_tensor_proto` as well. If success, another PR will be opened.

Running all test locally is time-consuming, hence only several unit tests are selected by me to be tested at first. By the way, could you tell me how to run doctest? Thanks.

Issue #12117, opened 2017-08-08
Bazel Windows Build: '//tensorflow/tools/proto_text:gen_proto_text_functions' failed to link


meteorcloudy:
This was first caught by Bazel CI: http://ci.bazel.io/blue/organizations/jenkins/TensorFlow/detail/TensorFlow/1030/pipeline
Reported at https://github.com/bazelbuild/bazel/issues/3524
```
  C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/amd64/link.exe /nologo /OUT:bazel-out/msvc_x64-py3-opt/bin/tensorflow/tools/proto_text/gen_proto_text_functions.exe /MACHINE:X64 /SUBSYSTEM:CONSOLE @bazel-out/msvc_x64-py3-opt/bin/tensorflow/tools/proto_text/gen_proto_text_functions.exe-2.params /DEFAULTLIB:libcmt.lib.
[...]

PR #12151, opened 2017-08-09
Fix Windows Linking error

meteorcloudy:
Fix https://github.com/tensorflow/tensorflow/issues/12117

`stringpiece.cc` is only the src of `lib_internal`, so we have to add it as a dependency.
@gunan @learyg

### DIFF
--- a/tensorflow/tools/proto_text/BUILD
+++ b/tensorflow/tools/proto_text/BUILD
@@ -34,7 +34,7 @@ cc_binary(
     visibility = ["//tensorflow:internal"],
     deps = [
         ":gen_proto_text_functions_lib",
-        "//tensorflow/core:lib_proto_parsing",
+        "//tensorflow/core:lib_internal",
     ],
 )
 


meteorcloudy:
Testing at http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/35/console

tatatodd (approved):

tatatodd:
Jenkins, test this please.

gunan:
the failure is due to sanity checks failing.
Here is the log:

```
Running do_buildifier on 213 files

tensorflow/tools/proto_text/BUILD # listsort unsafesort sort:cc_binary.deps

buildifier took 0 s

37d36
<         "//tensorflow/core:lib_proto_parsing",
38a38
>         "//tensorflow/core:lib_proto_parsing",
```

tensorflow/tools/proto_text/BUILD
```diff
@@ -35,6 +35,7 @@ cc_binary(
     deps = [
         ":gen_proto_text_functions_lib",
         "//tensorflow/core:lib_proto_parsing",
+        "//tensorflow/core:lib_internal",
```
tatatodd: Oops, you'll need to swap this line with the previous one.  Otherwise our BUILD-file lint checker fails.

tatatodd (dismissed):

tensorflow/tools/proto_text/BUILD
snnn: Remove lib_proto_parsing?

snnn:
@meteorcloudy  You may remove core:lib_proto_parsing

```diff
--- a/tensorflow/tools/proto_text/BUILD
+++ b/tensorflow/tools/proto_text/BUILD
@@ -34,7 +34,7 @@ cc_binary(
	 visibility = ["//tensorflow:internal"],
	 deps = [
		 ":gen_proto_text_functions_lib",
-        "//tensorflow/core:lib_proto_parsing",
+        "//tensorflow/core:lib_internal",
	 ],
 )

@@ -60,7 +60,7 @@ cc_library(
		 ],
	 }),
	 deps = [
-        "//tensorflow/core:lib_proto_parsing",
+        "//tensorflow/core:lib_internal",
	 ],
 )
```

tensorflow/tools/proto_text/BUILD
meteorcloudy: Thanks, will do it.

meteorcloudy:
Looks like the build is fixed!
The test failures at http://ci.tensorflow.org/job/tensorflow-pr-win-bazel/35/console is caused by a bug in Bazel.
Already filed https://github.com/bazelbuild/bazel/issues/3531, I'll make sure the fix cherry-picked for 0.5.4

meteorcloudy:
Lint error is fixed. \o/

rmlarsen (approved):

learyg:
Thanks for the fix, and sorry for the breakage!

rmlarsen:
@meteorcloudy I'm afraid that this creates a layering check violation internally. I'll have to roll it back.

meteorcloudy:
@rmlarsen I see, I'm fine with the rollback.

@learyg Is it possible to implement vmodule without introducing `stringpiece` in [`logging.cc`](https://github.com/tensorflow/tensorflow/commit/072b0c9#diff-9c3ca4a8ed42c1c586aff410128f2ab0R17)?

Related change: https://github.com/tensorflow/tensorflow/commit/072b0c9

learyg:
@meteorcloudy Yes, we could define a hasher for use with the unordered map, that was the primary motivation for grabbing StringPiece (it already had one defined). I'm unfortunately AFK for the next few days, understand if you need to roll back in the meantime, might not be too hard to dupe a custom hasher into that translation unit though.

meteorcloudy:
@rmlarsen Looks like this change is not getting rollbacked? Is it OK to keep it like this way?
Otherwise, @learyg do you mind implementing the hasher for unordered map to avoid include stringpiece?

snnn:
Hi,

Now we get the same error for gen_proto_text_functions.exe.

PR #12171, opened 2017-08-10
Allow disabling of selected tests in vector ops

DavidNorman:
Swap over the test instantiate macro to allow selective disabling of these tests


rmlarsen (approved):

rmlarsen:

Issue #151, opened 2015-11-11
Typo in reshape documentation


EigenFace:
Small typo at 
http://tensorflow.org/api_docs/python/array_ops.md#AUTOGENERATED-shapes-and-shaping 
Under the reshape example it says: 
> tensor 't' is [[[1, 1], [2, 2]]
>                [[3, 3], [4, 4]]]
> tensor 't' has shape [2, 2]

But I think t initially has the shape [2, 2, 2] not [2, 2].

PR #12173, opened 2017-08-10
[OpenCL] Fixes double memcpy bug (#151)

lukeiwanski:
As the debg CopyOp is called on a Tensor without type, we need to use
the DataType enum to get type information, and use this to pass the type
on to Eigen. This is a workaround Eigen's need to have a type when
calling memcpy. If the Eigen memcpy can be provided without a type
requirement, then the memcpy in sycl_util is unnecessary.

### DIFF
--- a/tensorflow/core/common_runtime/sycl/sycl_util.h
+++ b/tensorflow/core/common_runtime/sycl/sycl_util.h
@@ -32,7 +32,7 @@ inline void* GetBase(Tensor* dst) { return DMAHelper::base(dst); }
 
 inline void SYCLmemcpy(Eigen::SyclDevice const& device,
                        Tensor const& src_tensor, Tensor* dst_tensor) {
-  size_t size = src_tensor.TotalBytes();
+  const size_t size = src_tensor.TotalBytes();
   void* dst_ptr = GetBase(dst_tensor);
   void const* src_ptr = GetBase(&src_tensor);
 
@@ -67,8 +67,10 @@ inline void SYCLmemcpy(Eigen::SyclDevice const& device,
     case DT_INT8:
     case DT_QINT8:
     case DT_QUINT8:
-    default:
       COPY_WITH_TYPE(cl::sycl::cl_uchar);
+      break;
+    default:
+      LOG(FATAL) << "Unknown data type " << src_tensor.dtype();
       break;
   }
 #undef COPY_WITH_TYPE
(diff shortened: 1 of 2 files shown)


rmlarsen (changes requested):

tensorflow/core/common_runtime/sycl/sycl_util.h
```diff
@@ -21,17 +21,58 @@ limitations under the License.
...
-  inline void* GetBase(const Tensor* src) {
-    return const_cast<void*>(DMAHelper::base(src));
-  }
+inline void const* GetBase(const Tensor* src) { return DMAHelper::base(src); }
+inline void* GetBase(Tensor* dst) { return DMAHelper::base(dst); }
 
-  inline void* GetBase(Tensor* dst) { return DMAHelper::base(dst); }
+inline void SYCLmemcpy(Eigen::SyclDevice const& device,
+                       Tensor const& src_tensor, Tensor* dst_tensor) {
+  size_t size = src_tensor.TotalBytes();
```
lukeiwanski: I will fix `size` to be const in the next commit ( waiting on the testing to finish ) 
the `src_ptr` is already const 
in order to make `dst_ptr` const we would have to change the interfance of device.memcpy ( it won't compile with const at present )

tensorflow/core/common_runtime/sycl/sycl_util.h
```diff
@@ -21,17 +21,58 @@ limitations under the License.
...
+      COPY_WITH_TYPE(cl::sycl::cl_ushort);
+      break;
+    case DT_BOOL:
+      COPY_WITH_TYPE(bool);
+      break;
+    case DT_UINT8:
+    case DT_INT8:
+    case DT_QINT8:
+    case DT_QUINT8:
+    default:
```
lukeiwanski: Good point! Thanks for pointing this out!

lukeiwanski: Fixed in bcfa192

tensorflow/core/common_runtime/sycl/sycl_util.h
lukeiwanski: `size_t size` -> `const size_t size` in bcfa192

rmlarsen (approved):

rmlarsen:

Issue #10450, opened 2017-06-06
//tensorflow/core:platform_profile_utils_cpu_utils_test is failing on ppc64le with "Check failed: cpu_frequency > 0 (-1 vs. 0)"


sandipmgiri:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
    Ubuntu 16.04 (ppc64le)
- **TensorFlow installed from (source or binary)**:
    Installed from source (v1.0.1)
- **TensorFlow version (use command below)**:
   ('v1.0.1-0-ge895d5c-dirty', '1.0.1')
- **Bazel version (if compiling from source)**:
    0.4.4-2017-05-26 (@80a07b5)
- **CUDA/cuDNN version**:
    CUDA = 8.0 and cuDNN = 5.1
- **GPU model and memory**:
    GPU 0: Tesla P100-SXM2-16GB
    GPU 1: Tesla P100-SXM2-16GB
- **Exact command to reproduce**:
   bazel test --config=opt --config=cuda  //tensorflow/core:platform_profile_utils_cpu_utils_test

### Describe the problem
This test is passing successfully on X86, however getting failure on ppc64le with following errors:

### Source code / logs
```
$ bazel test --config=opt --config=cuda  //tensorflow/core:platform_profile_utils_cpu_utils_test

exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
Running main() from test_main.cc
[==========] Running 5 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 5 tests from CpuUtilsTest
[ RUN      ] CpuUtilsTest.SetUpTestCase
[       OK ] CpuUtilsTest.SetUpTestCase (0 ms)
[ RUN      ] CpuUtilsTest.TearDownTestCase
[       OK ] CpuUtilsTest.TearDownTestCase (0 ms)
[ RUN      ] CpuUtilsTest.CheckGetCurrentClockCycle
[       OK ] CpuUtilsTest.CheckGetCurrentClockCycle (0 ms)
[ RUN      ] CpuUtilsTest.CheckCycleCounterFrequency
F tensorflow/core/platform/profile_utils/cpu_utils_test.cc:56] Check failed: cpu_frequency > 0 (-1 vs. 0)
external/bazel_tools/tools/test/test-setup.sh: line 123:  7285 Aborted                 (core dumped) "${TEST_PATH}" "$@"
```

[...]

PR #12180, opened 2017-08-10
Adding support for s390x in calculation of cpu_frequency

namrata-ibm:
PR for changing the GetCycleCounterFrequency for s390x to fix test `//tensorflow/core:platform_profile_utils_cpu_utils_test`.

Facing issue similar to power reported [here](https://github.com/tensorflow/tensorflow/issues/10450#issuecomment-307097621) .

On s390x too `cpu_frequency = -1` and hence below test fails:
```
[ RUN      ] CpuUtilsTest.CheckCycleCounterFrequency
2017-07-17 04:43:02.050579: F tensorflow/core/platform/profile_utils/cpu_utils_test.cc:57] Check failed: cpu_frequency > 0 (-1 vs. 0)
```

Extending the fix added for PPC raised via [PR ](https://github.com/tensorflow/tensorflow/pull/10522#issuecomment-311064869)

### DIFF
--- a/tensorflow/core/platform/profile_utils/cpu_utils.cc
+++ b/tensorflow/core/platform/profile_utils/cpu_utils.cc
@@ -33,6 +33,11 @@ static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
      static const uint64 cpu_frequency = GetCycleCounterFrequencyImpl();
      return cpu_frequency;
 }
+#elif defined(__s390x__)
+   /* static */ uint64 CpuUtils::GetCycleCounterFrequency() {
+     static const uint64 cpu_frequency = GetCycleCounterFrequencyImpl();
+     return cpu_frequency;
+}
 #else
    /* static */ int64 CpuUtils::GetCycleCounterFrequency() {
      static const int64 cpu_frequency = GetCycleCounterFrequencyImpl();
--- a/tensorflow/core/platform/profile_utils/cpu_utils.h
+++ b/tensorflow/core/platform/profile_utils/cpu_utils.h
@@ -99,6 +99,8 @@ class CpuUtils {
   // the first call will incur overhead, but not subsequent calls.
   #if defined(__powerpc__) || defined(__ppc__) && ( __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
      static uint64 GetCycleCounterFrequency();
+  #elif defined(__s390x__)
+     static uint64 GetCycleCounterFrequency();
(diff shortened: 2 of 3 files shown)


tensorflow/core/platform/profile_utils/cpu_utils.cc
```diff
@@ -33,6 +33,11 @@ static ICpuUtilsHelper* cpu_utils_helper_instance_ = nullptr;
      static const uint64 cpu_frequency = GetCycleCounterFrequencyImpl();
      return cpu_frequency;
 }
+#elif defined(__s390x__)
```
gunan: The block here looks exactly same as the block above. Shouldn't we just add `|| defined(__s390x__)` to the end of the `if` statement above?

tensorflow/core/platform/profile_utils/cpu_utils.h
```diff
@@ -99,6 +99,8 @@ class CpuUtils {
   // the first call will incur overhead, but not subsequent calls.
   #if defined(__powerpc__) || defined(__ppc__) && ( __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
      static uint64 GetCycleCounterFrequency();
+  #elif defined(__s390x__)
```
gunan: same as above. Why not merge this with the `if` block?

tensorflow/core/platform/profile_utils/cpu_utils_test.cc
```diff
@@ -57,6 +57,10 @@ TEST_F(CpuUtilsTest, CheckCycleCounterFrequency) {
      const uint64 cpu_frequency = CpuUtils::GetCycleCounterFrequency();
      CHECK_GT(cpu_frequency, 0);
      CHECK_NE(cpu_frequency, unsigned(CpuUtils::INVALID_FREQUENCY));
+  #elif defined(__s390x__)
```
gunan: same as above.

gunan (changes requested):

namrata-ibm:
@gunan, I saw a check for LE in the ppc commit and since s390x is big endian, thought of keeping the condition separate for s390x for avoiding any future conflicts. 
Would it be better to merge both in same if condition?

gunan:
I would still merge these for the sake of keeping code cleaner.
Then we can worry about the added complexity (if needed) when adding ppc LE support.

namrata-ibm:
@gunan, thank you for  your feedback.
Closing this PR; opened #12201

Issue #12070, opened 2017-08-07
[XLA] Should we disable `REGISTER_XLA_BACKEND(GPU)` if there is no GPU installed on the machine? 


ScorpioCPH:
I install TensorFlow from source with `XLA` and `TF_CPP_MIN_VLOG_LEVEL` enabled (without `GPU`), after trying `XLA` example, I got these logs:

```shell
...
XLA op registration: device: XLA_CPU_JIT op: AssignAddVariableOp
XLA op registration: device: XLA_GPU_JIT op: AssignAddVariableOp
XLA op registration: device: XLA_CPU_JIT op: _UnsafeReadVariable
XLA op registration: device: XLA_GPU_JIT op: _UnsafeReadVariable
XLA op registration: device: XLA_CPU_JIT op: Unpack
XLA op registration: device: XLA_GPU_JIT op: Unpack
...
```

But there is no `GPU` installed on my machine, so I don't enable `CUDA` option, but it still do

```cpp
REGISTER_XLA_BACKEND(DEVICE_GPU_XLA_JIT, kGpuAllTypes, GpuOpFilter);
```

Should we disable `REGISTER_XLA_BACKEND(GPU)` if there is no `GPU` installed on the machine?

@tatatodd @hawkinsp WDYT?

------------------------

### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
  No.
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
  Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**:
  Source
- **TensorFlow version (use command below)**:
  v1.2.1
- **Python version**: 
  2.7
- **Bazel version (if compiling from source)**:
  0.4.5
- **CUDA/cuDNN version**:
  N/A
- **GPU model and memory**:
  N/A
- **Exact command to reproduce**:

ScorpioCPH:
ping @tatatodd @hawkinsp

hawkinsp:
Thanks for the bug report!

These messages are benign — I would just ignore them. They won't hurt anything and they only show up if you turn on verbose logging; at worst they could slow down startup a very tiny amount.

We could maybe try to avoid creating op registrations if not building with CUDA, but if it isn't causing a problem I don't think it's a high priority.

However if you want to send a PR to fix it I wouldn't object :-)

ScorpioCPH:
@hawkinsp Thanks for your reply!

I will send a PR to fix it later :)

PR #12182, opened 2017-08-10
[XLA] Register XLA_GPU iff cuda enable

ScorpioCPH:
Fix https://github.com/tensorflow/tensorflow/issues/12070.

@hawkinsp PTAL, Thanks!

### DIFF
--- a/tensorflow/compiler/jit/BUILD
+++ b/tensorflow/compiler/jit/BUILD
@@ -24,6 +24,7 @@ package(
 load("//tensorflow:tensorflow.bzl", "cc_header_only_library")
 load("//tensorflow:tensorflow.bzl", "tf_kernel_library")
 load("@local_config_cuda//cuda:build_defs.bzl", "if_cuda")
+load("@local_config_cuda//cuda:build_defs.bzl", "if_cuda_is_configured")
 
 # Target that bundles up the XLA CPU and GPU JIT devices.
 cc_library(
@@ -32,10 +33,11 @@ cc_library(
     deps = [
         ":xla_cpu_device",
         ":xla_cpu_jit",
+        "//tensorflow/compiler/plugin",
+    ] + if_cuda_is_configured([
         ":xla_gpu_device",
         ":xla_gpu_jit",
-        "//tensorflow/compiler/plugin",
-    ],
+    ]),
     alwayslink = 1,
 )
 
--- a/tensorflow/compiler/tf2xla/BUILD
+++ b/tensorflow/compiler/tf2xla/BUILD
@@ -20,6 +20,8 @@ package(
     default_visibility = [":internal"],
 )
 
+load("@local_config_cuda//cuda:build_defs.bzl", "if_cuda_is_configured")
+
 cc_library(
     name = "xla_compiler",
     srcs = [
@@ -29,7 +31,10 @@ cc_library(
         "xla_helpers.cc",
         "xla_op_kernel.cc",
         "xla_op_registry.cc",
-    ],
+        "xla_cpu_backend.cc",
+    ] + if_cuda_is_configured([
+        "xla_gpu_backend.cc",
+    ]),
     hdrs = [
         "xla_compilation_device.h",
         "xla_compiler.h",
--- a/tensorflow/compiler/tf2xla/xla_cpu_backend.cc
+++ b/tensorflow/compiler/tf2xla/xla_cpu_backend.cc
@@ -1,0 +1,37 @@
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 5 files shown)


tensorflow/compiler/jit/xla_gpu_device.cc
```diff
@@ -51,14 +51,18 @@ Status XlaGpuDeviceFactory::CreateDevices(const SessionOptions& options,
   return Status::OK();
 }
 
+#if GOOGLE_CUDA
```
hawkinsp: Now I think about this a bit more closely, there's a problem here.

There are two separate things going on here:
a) whether you have the CUDA libraries and enabled CUDA at "configure" time.
b) whether you are building with a CUDA compiler (by passing --config=cuda to Bazel).

For most parts of TensorFlow, these are identical. However for XLA, we need only (a) but not (b) --- XLA generates its own GPU code without needing nvcc.

The PR as written compiles the GPU registrations only in case (b). However we do support (and indeed have continuous builds that test) that you can run the XLA tests if (a) is true but (b) is not.

The correct way to do this would be to use the "if_cuda_is_configured()" bazel macro to conditionally compile this code.

You can see an example of how this works here:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/stream_executor/BUILD#L41
The if_cuda_is_configured macro expands to the empty list if (a) is false, or its argument, if (a) is true. So we can use it to add or remove files from the build if CUDA is configured.

For xla_gpu_device.cc, you can probably achieve this by wrapping the two GPU dependencies of this build target in a "if_cuda_is_configured" block. That target is the one linked in by the Tensorflow Python runtime if you enable XLA.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/jit/BUILD#L35
(You'll need to add if_cuda_is_configured to the "load" statements, as in the stream_executor BUILD file.)

For xla_op_registry.cc, you'll need to move the REGISTER_XLA_BACKEND and GpuOpFilter into a separate file (say xla_gpu_backend.cc) and compile that file conditionally as well.

Make sense?

hawkinsp (commented):
Thanks for the PR!

tensorflow/compiler/jit/xla_gpu_device.cc
ScorpioCPH: @hawkinsp Thank you for your suggestion!

 I will surely follow this to update this PR.

ScorpioCPH:
@hawkinsp PTAL :)

rmlarsen:
@hawkinsp could you take another look, please?

ScorpioCPH:
@hawkinsp PR updated. Could you take another look, please?

:-)

hawkinsp (approved):
Looks good. Sorry for the delayed response; I'm on vacation this week.

hawkinsp:

jhseu:
Jenkins, test this please

jhseu:
Jenkins, test this please

hawkinsp:
Jenkins, test this please.

caisq:
FYI, this CL has caused compilation error in tensorflow's nightly builds, e.g.,

tensorflow/compiler/tf2xla/xla_cpu_backend.cc: In function 'bool tensorflow::CpuOpFilter(tensorflow::KernelDef*)':
tensorflow/compiler/tf2xla/xla_cpu_backend.cc:23:11: error: invalid use of incomplete type 'class tensorflow::KernelDef'
   if (kdef->op() == "RandomStandardNormal") {
           ^
In file included from ./tensorflow/core/framework/op_kernel.h:27:0,
                 from ./tensorflow/core/common_runtime/device.h:41,
                 from ./tensorflow/core/common_runtime/local_device.h:19,
                 from ./tensorflow/compiler/tf2xla/xla_op_registry.h:25,
                 from tensorflow/compiler/tf2xla/xla_cpu_backend.cc:16:
./tensorflow/core/framework/kernel_def_builder.h:27:7: error: forward declaration of 'class tensorflow::KernelDef'
 class KernelDef;
       ^
tensorflow/compiler/tf2xla/xla_cpu_backend.cc:24:9: error: invalid use of incomplete type 'class tensorflow::KernelDef'
     kdef->clear_constraint();

See full example logs [here](http://ci.tensorflow.org/view/Tensorflow%20Jenkins%20Monitored%20builds/job/nightly-matrix-cpu/TF_BUILD_IS_OPT=OPT,TF_BUILD_IS_PIP=NO_PIP,TF_BUILD_PYTHON_VERSION=PYTHON2,label=cpu-slave/605/consoleFull).

@ScorpioCPH Can you come up with a quick fix?

ScorpioCPH:
@caisq ok, I will fix it today.

jhseu:
@ScorpioCPH already fixed it here: https://github.com/tensorflow/tensorflow/pull/12565

ScorpioCPH:
@jhseu Thanks!

PR #12189, opened 2017-08-10
CI script for building Raspberry Pi wheels

petewarden:
This CL expands on @ebrevdo's original PR adding Pi support by integrating it into the CI build scripts. This is a step towards nightly builds of Pi wheels and distribution of prebuilt binaries.

Issue #9697 is still blocking us from switching to the latest Eigen version, which is required for ARM compilation, so the script makes a local change to work around this for now. There's also one header added to oauth_client.cc, because it's required by boringssl on this platform.


petewarden:
Fixed commit emails.

PR #12190, opened 2017-08-10
Create CI build script for Raspberry Pi

petewarden:
This integrates the work by @ebrevdo to create a build script for the Raspberry Pi with the CI system. This script creates a Python wheel for TensorFlow on the Pi, and is a step towards creating automatic redistributable binaries for the platform.

I had to update oauth_client.cc for a missing header on this platform, and work around the eigen updating issue tracked in #9697.

### DIFF
--- a/third_party/toolchains/cpus/arm/build_raspberry_pi.sh
+++ b/third_party/toolchains/cpus/arm/build_raspberry_pi.sh
@@ -36,6 +36,17 @@ set -e
 
 yes '' | ./configure
 
+# We need to update the Eigen version, because of compiler failures on ARM when
+# using the version currently (Aug 10th 2017) pulled by mainline TensorFlow. We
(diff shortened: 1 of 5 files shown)


third_party/toolchains/cpus/arm/build_raspberry_pi.sh
```diff
@@ -36,6 +36,17 @@ set -e
 
 yes '' | ./configure
```
gunan: I think I previously missed this, but the location to this file is wrong. This folder is for toolchains bazel uses to build.
Could you move this under `tensorflow/tools/ci_build/pi` ?

petewarden: Good idea, thanks! Done.

gunan (approved):

PR #12191, opened 2017-08-10
Fix bugs of rounding in converting image from float to int

ychfan:


ychfan:
@googlebot I signed it!

drpngx:
@ychfan I think it's a potentially dangerous change that breaks existing behavior. It is probably better to add a new option for this functionality. Feel free to amend the PR if you think that's worth the trouble.

sb2nov:
@ychfan ping

ychfan:
@drpngx but the current implementation is not working as it supposed, in the "else" branch, image is converted from float(between 0 to 1) to unsigned int(for example 0 to 255), then dtype.max is 255. The origin code uses dtype.max + 0.5, makes the scale non-uniform.

drpngx:
@ychfan Right, I see. Can you add a test for that?

ychfan:
@drpngx I added a very simple test.

drpngx:
Jenkins, test this please.

drpngx (approved):

drpngx:
OK, thanks. We'll need to be careful when merging into our internal repo since this is a bugfix that might break existing behavior.

/CC: @martinwicke FYI

drpngx:
@gunan @yifeif kokoro is wedged? On this PR and #12504?

yifeif:
Looks like we need to ping to a newer version of https://github.com/tensorflow/tensorflow/blob/master/WORKSPACE#L3?
@jart

jart:
I'm not sure I'd describe the old behavior a problem or a bug. This just appears to be round instead of floor. I'm assuming there's a performance advantage to using `+ 0.5` instead of `tf.round`?

@yifeif What's wrong with the Closure Rules dependency? How can I help?

yifeif:
Thanks for the comment @jart. Sorry should have provided more info. The kokoro build is failing with `/tmpfs/tmp/bazel/external/io_bazel_rules_closure/closure/private/defs.bzl:27:16: The `set` constructor for depsets is deprecated and will be removed. Please use the `depset` constructor instead. You can temporarily enable the deprecated `set` constructor by passing the flag --incompatible_disallow_set_constructor=false`. 

But I just realized this is due to the "io_bazel_rules_closure" dependency in the PR branch is too old.

@ychfan do you mind do a rebase?

drpngx:
Thank you for figuring this out!

ychfan:
@jart `+ 0.5` can do the round things, but it has to be applied to the image data not scaler

ychfan:
rebased

drpngx:
Jenkins, test this please.

drpngx:
@andrewharp is the `split` problem at `master`?

gunan:
Looks like these failures are directly related to this cl:
https://source.cloud.google.com/results/invocation/543774f5-bfbb-4362-806e-63e4e3b31004/targets/%2F%2Ftensorflow%2Fpython:image_ops_test?page=log

Could you fix them?

ychfan:
The test case is for `rgb_to_grayscale` and it calls `convert_image_dtype` to convert image from `float32` to `uint8`. I guess the test case may be wrong

martinwicke:
I am still not convinced this actually is a needed bugfix:

The old behavior is
```
new = old * (max+0.5)
```
with the assumption that values are between 0 and 1 before. The desired behavior is a clean floor, i.e., if `old == n/max`, then `new = n`. The old way does this, I think, while the new could produce either `new = n` or `new = n+1` depending on how the numerics dice fall.

But it's been a bit, I may have messed up too.

tensorflow/python/ops/image_ops_impl.py
```diff
@@ -1086,8 +1086,9 @@ def convert_image_dtype(image, dtype, saturate=False, name=None):
         return math_ops.multiply(cast, scale, name=name)
       else:
         # Converting from float: first scale, then cast
-        scale = dtype.max + 0.5  # avoid rounding problems in the cast
+        scale = dtype.max
         scaled = math_ops.multiply(image, scale)
+        scaled = scaled + 0.5  # avoid rounding problems in the cast
```
martinwicke: See my comment, I am not sure this is correct.

martinwicke (changes requested):

tensorflow/python/ops/image_ops_impl.py
ychfan: Think about the case, when you converting a `uint8` image from RGB to grayscale. You first convert the image to `float32`, then do the conversion, you may get the value like `0.9/255`. when converting back to `uint8`, it supposed to be rounded to `1`. I guess this conversion bug will cause mistakes in many other image operations, like resize.

martinwicke: I suppose whether 0.9/255 should be 0 or 1 depends on whether you want your int domain to consist of equal-sized buckets. The current behavior implements a floor. I believe this is the correct behavior, since it maps 0..1 into 256 equal-sized buckets. 

If you use round behavior as you suggest, [0..0.5) is mapped to 0, then you have 254 buckets of size 1, and finally another bucket of size 0.5. 

I believe floor is the better solution here. However, I recognize this could be considered  a matter of taste. We cannot change behavior though, so we should leave this alone.

martinwicke:
I will close this PR for now; I don't believe this is a bug. @aselle, if you want to change my mind, please review. 

I would be happy to consider a PR which clarifies the definition of the intensity representation using integers, since it appears to be confusing.

tensorflow/python/ops/image_ops_impl.py
ychfan: But the old code doesn't have equal-sized buckets either.
For example, when converting from `244.9/255` in `float32` to `uint8`, first multiply with `255 + 0.5` and get `245.4`, then floor to `245`. However, it supposed to be `244`

martinwicke: You are right. 

0.5 is a very crude epsilon. We could reduce it to a very small number. The point of adding it in the first place was to avoid 1/255 + 3/255 = 3.9999/255 ==> 3/255. But it will never be perfect. I'm happy to reduce it, or if you have a fundamentally better approach, I'd be excited about that too.

PR #12196, opened 2017-08-11
Add regression tests  and fix some minor bugs

frreiss:
This pull request adds some additional unit tests for the `add_n`  and `accumulate_n` operators. I'm adding these tests as preparation for working on https://github.com/tensorflow/tensorflow/issues/10607 . Specifically, I added tests of the argument validation code in `accumulate_n` and a test of gradient computation for `add_n`.

This pull request also fixes two minor bugs that I encountered in the process of adding the unit tests:
* `accumulate_n` ignores its `tensor_dtype` argument when the operator has only one input. I've added additional error-handling logic to `accumulate_n` that rejects values of `tensor_dtype` that are different from the input type.
* The  test`UnaryOpTest.testComplex64Basic` was failing on my Mac because the `acosh` and `asinh` functions for single-precision complex floating point are only accurate to 6 decimal places. Under the assumption that 6 significant figures is good enough, I increased the threshold that the test cases use when comparing gradients. 

Note that if 6 significant figures is *not* good enough for `acosh` and `asinh`, a more involved fix will be necessary for the second problem.


rmlarsen:

rmlarsen (approved):

rmlarsen:
@frreiss please fix the Python linter errors.

https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/5801/consoleFull

PR #12206, opened 2017-08-11
BUG: Estimator.eval() runs feature_engineering_fn more than once.

### LINKED ISSUE #12205: BUG: TypeError in DNNClassifier.eval() when using same name for feature in feature_engineering_fn
### Describe the problem

If we use the  same key to replace a feature, tensorflow might throw TypeError when evaluating:

eg:
```python
def feature_engineering_fn(features, label):
  features["x"] = some_func(features["x"])
```

When `features` is `dict`, it is a mutable object. Hence the bug is caused by `evaluate` method which runs `feature_engineering_fn` again, see [code](https://github.com/facaiy/tensorflow/blob/c7b80d51da4fb6d51ea54a0bdf2601afa379d60c/tensorflow/contrib/learn/python/learn/estimators/estimator.py#L1165).

I'll open a PR later.

facaiy:
The PR is aimed to fix #12205.

eliminate the second call of `feature_engineering_fn` in `evaluate` method.

+ [x] add an unit test.

### DIFF
--- a/tensorflow/contrib/learn/python/learn/estimators/estimator.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/estimator.py
@@ -1167,7 +1167,6 @@ class Estimator(BaseEstimator):
 
     return model_fn_ops
 
-
   def _get_train_ops(self, features, labels):
     """Method that builds model graph and returns trainer ops.
 
--- a/tensorflow/contrib/learn/python/learn/estimators/estimator.py
+++ b/tensorflow/contrib/learn/python/learn/estimators/estimator.py
@@ -1118,13 +1118,14 @@ class Estimator(BaseEstimator):
     self._feature_engineering_fn = (
         feature_engineering_fn or _identity_feature_engineering_fn)
 
-  def _call_model_fn(self, features, labels, mode):
+  def _call_model_fn(self, features, labels, mode, metrics=None):
     """Calls model function with support of 2, 3 or 4 arguments.
 
     Args:
       features: features dict.
       labels: labels dict.
       mode: ModeKeys
+      metrics: Dict of metrics.
 
     Returns:
       A `ModelFnOps` object. If model_fn returns a tuple, wraps them up in a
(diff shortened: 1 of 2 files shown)


tensorflow/contrib/learn/python/learn/estimators/estimator.py
```diff
@@ -1203,12 +1211,6 @@ def _get_eval_ops(self, features, labels, metrics):
     model_fn_ops = self._call_model_fn(
         features, labels, model_fn_lib.ModeKeys.EVAL)
```
ispirmustafa: you need to send metrics to call_model_fn, right?
could you please add a test which catches this error?

ispirmustafa (changes requested):
Thank you Yan!

tensorflow/contrib/learn/python/learn/estimators/estimator.py
facaiy: Thanks for your review. 

Now `feature_engineerng_fn(features, label)` is called by `Estimator.eval()` twice.
Because `features` can be a `dict` which is passed by pointer for function, hence `features` will be converted by `feature_engineering_fn` twice. It's dangerous and might get in trouble.

So we should assure that `feature_engineering_fn` only be called once for `train`, `eval` and `predict`. That's why I send metrics to `call_model_fn`. 

In fact, I have added an unit test below to catch the error.

tensorflow/contrib/learn/python/learn/estimators/estimators_test.py
```diff
@@ -77,6 +78,44 @@ def model_fn(features, labels):
...
+             }, {
+               "y": constant_op.constant(["99."])
+             }
+
+    def feature_engineering_fn(features, labels):
+      # Github #12205: raise a TypeError if called twice.
+      _ = string_ops.string_split(features["x"])
+      features["x"] = constant_op.constant([9.])
+      label["y"] = constant_op.constant([99.])
+      return features, labels
```
facaiy: Hi, @ispirmustafa . In the `feature_engineering_fn`, it converts string tensor to float tensor. So if it's be called twice, TypeError will be raised because `features["x"]` has been converted to float tensor, which is unsupported by string_op.

tensorflow/contrib/learn/python/learn/estimators/estimator.py
ispirmustafa: Hi Yan,
I understand the reason of change. 
It seems there is a bug in current code: you forgot to send metrics to call_model_fn in Eval mode. Could you please check that?
otherwise LGTM

facaiy: Oh, My fault! I have corrected it. Thanks for your comment, ispirmustafa!

facaiy:
By the way, @ispirmustafa . I'm a newbie. If I want to run only one unit test, say the test case I added, could you tell me how to do it? Thanks.

facaiy:
I have found it: `bazel test -c opt //tensorflow/contrib/learn:estimators_test`

Could you let @tensorflow-jenkins  test the PR? Thanks.

facaiy:
Hi, @ispirmustafa . The patch has been modified as suggested, and could you review it? Thanks.

ispirmustafa (approved):

jhseu:
Jenkins, test this please

facaiy:
Thanks, @ispirmustafa and @jhseu .

PR #12209, opened 2017-08-11
Make LayerNormBasicLSTMCell compatible with datatypes other than float32

muxamilian:
`LayerNormBasicLSTMCell` only supported float32 so far. With this patch other datatypes such as float64 are available too. The datatype doesn't have to be specified explicitly but it is deducted from the input data as it already happens for example for BasicLSTMCell.

### DIFF
--- a/tensorflow/contrib/rnn/python/ops/rnn_cell.py
+++ b/tensorflow/contrib/rnn/python/ops/rnn_cell.py
@@ -1303,24 +1303,25 @@ class LayerNormBasicLSTMCell(rnn_cell_impl.RNNCell):
   def output_size(self):
     return self._num_units
 
-  def _norm(self, inp, scope):
+  def _norm(self, inp, scope, dtype=dtypes.float32):
     shape = inp.get_shape()[-1:]
     gamma_init = init_ops.constant_initializer(self._g)
     beta_init = init_ops.constant_initializer(self._b)
     with vs.variable_scope(scope):
       # Initialize beta and gamma for use by layer_norm.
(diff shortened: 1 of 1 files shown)


drpngx:
@ebrevdo any cycles for this?

tensorflow/contrib/rnn/python/ops/rnn_cell.py
```diff
@@ -1303,24 +1303,25 @@ def state_size(self):
   def output_size(self):
     return self._num_units
 
-  def _norm(self, inp, scope):
+  def _norm(self, inp, scope, dtype=dtypes.float32):
```
ebrevdo: Do you need a default value here?

muxamilian: No, you don't need it but in case some legacy code calls ```_norm``` directly – for whatever reason – the default argument prevents this code from breaking.

ebrevdo:
No one else should call _norm from outside this module... Does anyone?

muxamilian:
Not that I know of anyone doing that... I thought the default value wouldn't hurt anyone either.

ebrevdo:
sounds fine.

On Sun, Sep 17, 2017 at 11:37 AM, Maximilian Bachl <notifications@github.com

muxamilian:
Any updates on this?

martinwicke (approved):
Looks like Eugene is fine with this.

martinwicke:
Jenkins, test this please.

martinwicke:
Kokoro isn't reporting status atm.

PR #12216, opened 2017-08-11
Added the EMNIST dataset

JoshuaBradbury:
This PR adds EMNIST to the datasets. EMNIST is an extension of the MNIST dataset but it also adds Characters, and more numbers. It can be found in the paper here: https://arxiv.org/abs/1702.05373v1

### DIFF
--- a/tensorflow/contrib/learn/python/learn/datasets/emnist.py
+++ b/tensorflow/contrib/learn/python/learn/datasets/emnist.py
@@ -34,10 +34,10 @@ from tensorflow.python.framework import dtypes
 from tensorflow.python.framework import random_seed
 
 def enum(*sequential, **named):
-    enums = dict(zip(sequential, range(len(sequential))), **named)
-    reverse = dict((value, key) for key, value in enums.iteritems())
-    enums['name'] = reverse
-    return type('Enum', (), enums)
+  enums = dict(zip(sequential, range(len(sequential))), **named)
+  reverse = dict((value, key) for key, value in enums.iteritems())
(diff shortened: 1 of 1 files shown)


av8ramit:
Jenkins, test this please.

JoshuaBradbury:
The problem in the initial test was down to it taking too long, but this is mainly caused by where the data is pulled from, as well as the size of the data. This problem wasn't really a concern for the MNIST set as it only had at most 70,000 images and labels, where as every set in the EMNIST is bigger, and can be up to 800,000 images and labels.

JoshuaBradbury:

tensorflow/contrib/learn/python/learn/datasets/emnist.py
```diff
@@ -0,0 +1,309 @@
+# Copyright 2017 Joshua Bradbury. All Rights Reserved.
```
drpngx: We usually have "The TensorFlow Authors" here. Any to change it to that?

nealwu:
@vincentvanhoucke is this a dataset we are interested in adding? I hadn't heard of it before this.

vincentvanhoucke:
@nealwu I don't know what's our policy in regards to accepting new datasets into TF.
This one only has 2 citations thus far: https://scholar.google.com/scholar?oi=bibs&hl=en&cites=8838276906904533917&as_sdt=5

nealwu:
Doesn't seem noteworthy enough to include then. Going to close this. If others think it is worth including, feel free to reopen.

PR #12226, opened 2017-08-12
Fix issues due to old numpy versions in dist_test and gcs_test

caisq:
Previously numpy was installed with apt-get, leading to a version too
old to be compatible with the recently-added autograd dependency of
tensorflow. This change set fixes that.


caisq:
Test failures are all unrelated.

gunan (approved):

gunan:
Jenkins, test this please.

PR #12229, opened 2017-08-12
Fix error caused in MonitoredSession.close()

### LINKED ISSUE #12224: MonitoredSession has no `close`
------------------------

### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Minimum custom code
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: mac os 10.12.6
- **TensorFlow version (use command below)**: 1.2.1
- **Python version**: 2.7.13
- **Bazel version (if compiling from source)**: NA
- **CUDA/cuDNN version**: NA
- **GPU model and memory**: NA

### Describe the problem
Bug: MonitoredSession has no `close` but it should have according to https://www.tensorflow.org/api_docs/python/tf/train/MonitoredSession#close

I think in some cases, self._sess is not initialized.

### Source code / logs
minimum code to reproduce:
with tf.train.MonitoredTrainingSession() as sess:
  sess.close()

compared to working code using simple session:
with tf.Session() as sess:
  sess.close()

yongtang:
This fix tries to address the error raised in #12224:
```
>>> import tensorflow as tf
>>> with tf.train.MonitoredTrainingSession() as sess:
...   sess.close()
...
  File "<stdin>", line 2, in <module>
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/monitored_session.py", line 534, in __exit__
    self._close_internal(exception_type)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/training/monitored_session.py", line 569, in _close_internal
    self._sess.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

The error was caused by double call on `_internal_close()`.

This fix fixes the issue by skipping the error at the end of the `close()`.

This fix fixes #12224

### DIFF
--- a/tensorflow/python/training/monitored_session.py
+++ b/tensorflow/python/training/monitored_session.py
@@ -566,6 +566,8 @@ class _MonitoredSession(object):
           h.end(self._coordinated_creator.tf_sess)
     finally:
       try:
+        if self._sess is None:
+          raise RuntimeError('Session is already closed.')
         self._sess.close()
       finally:
         self._sess = None
--- a/tensorflow/python/training/monitored_session_test.py
+++ b/tensorflow/python/training/monitored_session_test.py
@@ -1414,6 +1414,12 @@ class MonitoredSessionTest(test.TestCase):
             isinstance(hook.run_metadata_list[0], config_pb2.RunMetadata))
         self.assertGreater(len(hook.run_metadata_list[0].partition_graphs), 0)
 
+  def test_with_statement_and_close(self):
+    # Test case for https://github.com/tensorflow/tensorflow/issues/12224
+    # where close() inside the with should have a better error message.
+    with self.assertRaisesRegexp(RuntimeError, 'Session is already closed'):
+      with monitored_session.MonitoredSession() as session:
+        session.close()
 
 class SingularMonitoredSessionTest(test.TestCase):
   """Tests SingularMonitoredSession."""


tensorflow/python/training/monitored_session.py
```diff
@@ -567,6 +567,8 @@ def _close_internal(self, exception_type=None):
     finally:
       try:
         self._sess.close()
+      except Exception:
```
caisq: Instead of this, would it be safer and cleaner to do:
```python
try:
  if self._sess:
    self._sess.close()
#...
```

caisq:
Also, it would be good to understand under what circumstances `self._sess` can be None. Maybe there is a bug elsewhere that should be fixed. If not, a unit test is needed to cover the case.

yongtang:
Thanks @caisq for the review. The PR has been updated. A test case has also been added.

tensorflow/python/training/monitored_session_test.py
```diff
@@ -1414,6 +1414,11 @@ def test_merge_caller_and_hook_run_options(self):
             isinstance(hook.run_metadata_list[0], config_pb2.RunMetadata))
         self.assertGreater(len(hook.run_metadata_list[0].partition_graphs), 0)
 
+  def test_with_statement_and_close(self):
+    # Test case for 12224
+    # No expection should be raised
+    with monitored_session.MonitoredSession() as session:
```
ispirmustafa: context manager exit calls the close. sess.close should not be called by user if context manager is used.

ispirmustafa (commented):
Thank you Yong. 
Instead of ignoring the wrong usage, could you please improve the error message? you can check if self._sess and return error saying 'it's already closed' if it's none.

yongtang:
@ispirmustafa Thanks for the review. The PR has been updated with error message outputted. Please take a look.

tensorflow/python/training/monitored_session_test.py
```diff
@@ -1414,6 +1414,12 @@ def test_merge_caller_and_hook_run_options(self):
             isinstance(hook.run_metadata_list[0], config_pb2.RunMetadata))
         self.assertGreater(len(hook.run_metadata_list[0].partition_graphs), 0)
 
+  def test_with_statement_and_close(self):
+    # Test case for 12224
```
caisq: nit: end comments with periods per google python style. 
https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

Also, the number "12224" doesn't mean much. Include the full github issue link instead.

yongtang: @caisq The PR has been updated. Thanks.

ispirmustafa (approved):

Issue #10767, opened 2017-06-16
[Feature Request] Exclude ties in function in_top_k


gfolego:
This is a feature request related to #10489 (tie handling in function in_top_k)

Would it be possible to add an option/argument specifying whether ties should be included or excluded?
In other words, make it possible for ties to return False instead of True.

Thanks!

aselle:
@gfolego, could you give some motivation for why this is needed. It seems you worry that there is a trivial null-space to the loss function where you return constant 1's as your logits which causes zero loss since every class is in top_k. However, with random weight initialization this doesn't seem likely to be found with sgd since the gradient will not reliably target that.

gfolego:
@aselle Sure! I'll use ImageNet as an example, but my case is quite similar.
Let's say I have the ILSVRC2017 dataset, and I'm working on image classification only, so my main objective is to get a high top 5 accuracy.
Now, I'm analyzing a CNN's performance given a specific number of training images, which will vary from very few to the complete training set. The motivation behind this analysis is to determine the best trade-off between size and accuracy, so we know how many images we need when collecting a new dataset.

In particular, ILSVRC2017 training dataset (for classification/localization) has between 732 and 1300 images for each class. If I use as few as 5 from each class for training a (large) CNN from scratch, this will most likely result in a CNN with random predictions or maybe predicting always the same class. So the top 5 accuracy for this model will be artificially (and incorrectly, IMHO) high.

So, if I'm early stopping or choosing a model based on this metric, I'll get a CNN that doesn't work in practice. Also, based on this approach, training with 5 samples will most likely achieve better top 5 accuracy than training with the complete set.

Please note that I'm not the first one experiencing this, as indicated in #10489 and in this issue on Keras (using TF's in_top_k) https://github.com/fchollet/keras/issues/6167

TL;DR: If a CNN predicts the same probability for every class, its top 5 accuracy in ILSVRC2017 classification is 100%, which is naturally incorrect.

Thanks!

aselle:
@fchollet, @martinwicke, what is your feeling on this feature? Should we add an argument to in_top_k? It seems like the conceptional conflict is that it is hard to make a top_k that ignores ties in a consistent way but is also fixed size.

martinwicke:
It seems it's hard either way -- if you include ties, you may end up with a larger than wanted set, if you exclude them, you may end up with a smaller than wanted set. If you want to enforce a constant size, you have to pick from the set of ties.

[...]

PR #12230, opened 2017-08-12
(Reopened PR) Add handle_ties arg to in_top_k Op to deal with ties.(#10767)

nolanliou:
Reopened PR for [10767](https://github.com/tensorflow/tensorflow/issues/10767)
Add handle_ties argument to in_top_k Op to make the Op more clear.


nolanliou:
@martinwicke please review it if you have time. thanks.

martinwicke:
Few questions: Is the shape function using the new information? Was the shape function wrong before?

The new default changes behavior (although it has strictly stronger guarantees, so I think it's fine). But we should call this out very explcitly in relnotes. Can you add relnotes for this change?

Should we make a similar change for topk so the two are aligned again?

nolanliou:
@martinwicke 
1. The shape function is right before and this change don't affect the input and output shape.
2. where is the relnotes?  sorry, I cannot find it.
3.  `If two elements are equal, the lower-index element appears first.`, The topk strategy is different from InTopK, Is align necessary?

cwhipkey:
I had a couple questions about the SAMPLE method and the intended use:

1. would it be useful to have a mode that matches top_k exactly, so that any class returning true here would be included in the equivalent top_k call?

2. what is SAMPLE used for?  My understanding of the code is that it doesn't guarantee exactly k classes return true.  Is that right?

nolanliou:
@cwhipkey 
1. So it is necessary to align with topk. 
2. SAMPLE means we choose some elements from the ties for guarantee exactly k classes. the result k classes may contain the object or not.

Actually, there is not a perfect way to deal with the tie situation. you could have a look at the [issue](https://github.com/tensorflow/tensorflow/issues/10767).

martinwicke:
I think aligning with top_k would be good. How hard would that be?

Issue #12133, opened 2017-08-09
[Feature] [Java]: Inspect tensors in a Graph


borissmidt:
Dear Tensorflow maintainers,

First of all i really like it that tensorflow allows easy interop from keras to tensorflow and then allows for interference to Java.

I have noticed that when you load an invalid graph, for example an empty byte array, then it will just accept it without giving an error. An exception will be thrown when we try to run the graph.

Thus i would propose two improvements for the tensorflow java API,
* throw an exception when loading an invalide graph.
* add api functionality to inspect the tensors contained in the graph

I also noticed in the example for java that you wrapped the graph builder in a convenience java class and it seems that this could be implemented in a really clean way in scala, I would gladly offer you my assistance to implement a scala wrapper. 

Kind regards, Boris

tatatodd:
@asimshankar Can you comment on the suggestions from @borissmidt ?  Thanks!

asimshankar:
@borissmidt : Thanks for the notes. Some thoughts:

- In general, `importGraphDef` will raise an exception when provided an invalid serialized `GraphDef` protocol message. It's just that the empty byte array is a valid serialization of an empty graph :). So I think it's working correctly for an empty graph. Are there invalid graphs for which you do not see an exception raised?

- For the other two, yeah, that seems fine to add. Though, what kinds of inspection were you interested in? You can get tensors by name using something like [`Graph.operation(String name)`](https://www.tensorflow.org/api_docs/java/reference/org/tensorflow/Graph.html#operation(java.lang.String)). If you'd want some form of graph traversal, that will require some implementation work, but the underlying C API does have the functions you need to build on (e.g., [`TF_OperationOutputConsumers`](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/c/c_api.h#L601))

We (TensorFlow maintainters) will probably not get to it in the near future, but if anyone would like to contribute a PR, we'd be more than happy to take a look. Just please do outline the proposed additions to the Java API first :)

Thanks!

borissmidt:
@asimshankar 
Thanks for noting me that there is a function named `Graph.operation(String name)`, it is really useful and it could be further improved with a function like `Graph.getOperations()` which returns a list of operations in the graph. This would allow us the inspect the graph directly and see if the correct one is loaded. 

I'm thinking about the empty graph behavior, since it seems like an odd thing to load an empty graph on purpose?

[...]

PR #12243, opened 2017-08-13
Feature add provide Java Graph class getter for all operations

quaeler:
c_api code and tests, Java code and tests, and JNI code to support exposing and using the cardinality of operations contained in a graph and an aggregate of them. This is to support https://github.com/tensorflow/tensorflow/issues/12133

As is the stuff of Git sleepover folklore, i stand in the bathroom mirror and softly say @martinwicke three times.


quaeler:
`I signed it!`

tensorflow/c/c_api.cc
```diff
@@ -1667,6 +1667,25 @@ void TF_DeleteGraph(TF_Graph* g) {
   if (del) delete g;
 }
 
+size_t TF_GraphOperationsCount(TF_Graph* graph) {
```
asimshankar: I'd recommend against adding these functions. It seems that [`TF_GraphNextOperation`](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/c/c_api.h#L837) provides the required functionality. So, unless this separate API provides some strong benefit (which I don't yet think it does), I'd advise against adding it.

tensorflow/java/src/main/java/org/tensorflow/Graph.java
```diff
@@ -77,6 +80,22 @@ public Operation operation(String name) {
   }
 
   /**
+   * Returns all operations that exist in the Graph.
+   */
+  public List<Operation> getOperations() {
```
asimshankar: Let's make this return an `Iterator<Operation>` instead of a `List<Operation>`. The C API for iteration predates my involvement with the project, so I'm not entirely sure, but it seems that if there was a reason to prefer iteration there (perhaps memory footprint of when traversing very large graphs), that same reason would apply here. 

I'd imagine something like:

```java
private final class OperationIterator implements Iterator<Operation> {
  OperationIterator(Graph g) {
    this.graph = g;
    advance();
  }

  private void advance() {
    Graph.Reference ref = graph.ref();
    long opHandle = 0;
    try {
       // nextOperation would be a method in graph_jni.cc
       opHandle = Graph.nextOperation(ref.nativeHandle(), pos++);
    } finally {
      ref.close();
    }
    op = opHandle == 0 ? new Operation(this.graph, opHandle) : null;
  }
  public boolean hasNext() { return op != null; }
  public Operation next() { 
    Operation ret = op;
    advance();
    return ret;
  }

  private Graph graph;
  private Operation op = null;
  private int pos = 0;
}

public Iterator<Operation> operations() {
  return new OperationIterator(this);
}
```

asimshankar: Let's call this `operations()` instead of `getOperations()` - keeping with the convention of avoiding the `get` prefix on other methods in this package.

tensorflow/java/src/main/java/org/tensorflow/Graph.java
```diff
@@ -77,6 +80,22 @@ public Operation operation(String name) {
   }
 
   /**
+   * Returns all operations that exist in the Graph.
```
asimshankar: Let's clarify that there is no significance to the order of the returned operations. Perhaps something like:

```java
* The {@link Operation} objects are returned in an underfined order.
```

tensorflow/java/src/test/java/org/tensorflow/GraphTest.java
```diff
@@ -71,6 +73,21 @@ private static void validateImportedGraph(Graph g, String prefix) {
   }
 
   @Test
+  public void getAllOperations() {
+    try (Graph g = new Graph()) {
+      List<Operation> operations = g.getOperations();
+
+      assertEquals(0, operations.size());
+
+      TestUtil.transpose_A_times_X(g, new int[2][2]);
```
asimshankar: To keep the test tighter (i.e., not requiring the test reader to know what graph `transpose_A_times_X` will produce), how about changing this to add a single operation?
Something like:

```java
assertFalse(g.operations().hasNext());
Operation op = TestUtil.placeholder(g, "myPlaceholder", DataType.FLOAT).op();
Iterator<Operation> iter = g.operations();
assertTrue(iter.hasNext());
assertEquals(iter.next(), op);
assertFalse(iter.hasNext());
```

asimshankar (changes requested):
Thanks for the PR! Some suggestions below.

tensorflow/java/src/main/java/org/tensorflow/Graph.java
quaeler: Returning as an Iterator opens up an issue with a potential concurrent modification; implementing an assurance (or preferably throwing) against this would be an extra pot of stew. Barring objection, i will note the lack of observance in the javadocs for the ~getter and implement it as spec'd.

asimshankar: The underlying C API is safe to concurrent modification, so the Java iterator would be too without any additional machinery. Perhaps you have a concern I'm not catching on to, and if so, please do elaborate.

quaeler: Traditionally (with quotes around that), some Iterator implementations will throw `ConcurrentModificationException` should the underlying collection change between construction time and a given invocation of `next()`. To have this iterator do that would be a bit of a hugger-ma-rug; so i'd vote for noting its absence in javadocs and moving on. Should this functionality become a demand for prime-time, we could go back and add functionality to allow such detection and a subsequent throw.

asimshankar: Ah, I see. Yes, some iterator implementations do that but as far as I understand the [interface specification](https://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html) does not require it (or even mention that they _could_ return this error).

So, yes, I agree that this iterator implementation shouldn't try to worry about throwing `ConcurrentModificationException`s. I'd even be okay with not talking about this detail in the javadoc at all - since it isn't part of the interface specification.

Thanks!

quaeler: Ya, the specification doesn't mention it, but it is basically ubiquitous in the implementation of the collections in the JRE and so often i find people expect it.

Thanks for the feedback - l.

tensorflow/c/c_api.cc
quaeler: I've newly arrived to TF, so i'm not sure i totally grok the graph representation. Looking at the implementation of `TF_GraphNextOperation`, it is both (potentially) returning the next operation, and also (potentially) altering the position value by greater than +1 per invocation (for what conditions, i won't understand until i understand the graph representation.)

Given that, the suggested iteration implementation where we rely on a JNI implementation that receives a position argument that we continually increment at the Java layer, seems like it might fall afoul.

That being written, i'd obviously cede to you if you told me that this usage of position won't fail us.

asimshankar: Ah, right. Good catch!
I suppose we could have the native method operate on the `OperationIterator` and edit the fields in-place?  Or perhaps return an object representing a pair of integers - the operation handle and the new value of pos?

quaeler: Ya - sure, it's not a grave thing; i just wanted to make sure i understood the salient actors of the method before i went ahead with the implementation.
Ok - thanks for the sanity check.

quaeler:
(Sorry for the bad cut and paste in my commit message.)

quaeler:
I don't believe i have the privs to affect any labels on this PR, so i'm textually passing the baton back to @asimshankar here.

tensorflow/java/src/main/native/graph_jni.cc
```diff
@@ -54,6 +54,30 @@ JNIEXPORT jlong JNICALL Java_org_tensorflow_Graph_operation(JNIEnv* env,
...
+JNIEXPORT jlongArray JNICALL Java_org_tensorflow_Graph_nextOperation(JNIEnv* env,
+                                                                     jclass clazz,
+                                                                     jlong handle,
+                                                                     jint position) {
+  TF_Graph* g = requireHandle(env, handle);
+  jlongArray rhett = nullptr;
+
+  if (g == nullptr) return rhett;
+
+  size_t* pos = reinterpret_cast<size_t*>(&position);
```
asimshankar: I don't think we want `reinterpret_cast` here. How about the following instead:

```c
size_t pos = static_cast<size_t>(position);
TF_Operation* operation = TF_GraphNextOperation(g, &pos);
```

tensorflow/java/src/main/native/graph_jni.cc
```diff
@@ -54,6 +54,30 @@ JNIEXPORT jlong JNICALL Java_org_tensorflow_Graph_operation(JNIEnv* env,
   return reinterpret_cast<jlong>(op);
 }
 
+JNIEXPORT jlongArray JNICALL Java_org_tensorflow_Graph_nextOperation(JNIEnv* env,
+                                                                     jclass clazz,
+                                                                     jlong handle,
+                                                                     jint position) {
+  TF_Graph* g = requireHandle(env, handle);
+  jlongArray rhett = nullptr;
```
asimshankar: Might I suggest moving this much lower to where it is needed? So this would be something like:

```c
  TF_Graph* g = requireHandle(env, handle);
  if (g == nullptr) return nullptr;
  ...
  if (operation == nullptr) return nullptr;
  ...
  jlongArray ret = env->NewLongArray(2);
  env->SetLongArrayRegion(ret, 0, 2, handle_and_position);
  return ret;
}
```

asimshankar (changes requested):
Thanks for the change!

I took the liberty of making some trivial formatting changes (in the future, if you could run "[google-java-format](https://github.com/google/google-java-format)" on the Java files, that would be great - helps us keep formatting consistent with the [style guide](https://google.github.io/styleguide/javaguide.html) we try to follow).

Other than one question about the use of `reinterpret_cast`, I think we're good to go.

tensorflow/java/src/main/native/graph_jni.cc
quaeler: shoaley - sounds fine & will do.

tensorflow/java/src/main/native/graph_jni.cc
quaeler: better indeed - will do.

quaeler:
Sorry about the formatting issue - will run the formatter on code prior to commit in the future.

asimshankar (approved):

asimshankar:
Jenkins, test this please

quaeler:
Are these build timeouts a Jenkins issue or an issue that got fixed in master since i branched and that i need bring into this branch to make the builds copacetic?

martinwicke:
Jenkins, test this please.

quaeler:
It appears that there is not consistency in these build failures - this time it's a Linux Python 3 failure and Linux CPU failure ( //tensorflow/contrib/rnn:gru_ops_test ); last time Linux CPU failed with a different earlier test (in time out) so perhaps it would have failed gru_ops_test, but Python 3 passed and it was Windows CMake that had to be killed after nearly two hours.

What's the path forward in this situation?

quaeler:
FWIW `//tensorflow/contrib/rnn:gru_ops_test` passes locally against this branch (though i had to dance with pip to get it to install python-mock in a non-virtualenv since mock wants a modern version of six and macOS these days SIP-protects the six python libraries for whatever reason.)

martinwicke:
You can check ci.tensorflow.org and see whether there's an existing failure
at head. If so, failures may be unrelated. Otherwise the build should be
consistent.

quaeler:
The status balls at https://ci.tensorflow.org/job/tensorflow-pull-requests-multijob/ were scene extras in that elevator scene from The Shining. 

With that plus random tests failing and timing out that touch nothing to do with the meager Java related toe nail clippings of files that i touched, can we assume that there is a larger problem here?

martinwicke:
The py3 and cpu failures are unrelated. This is fine to merge.

quaeler:
I don't have permission to merge this; could one of you fine gents do the deed?

jhseu:
Jenkins, test this please

quaeler:
tensorflow-pull-request-multijob last had a successful build `6 days 23 hr` ago.

Should i try to merge down from master to hopefully pick up any recent changes that might produce a better build result, in the theory that many of the branches in this week of continual PR build fails stem from a master at a non-building point?

jhseu:
Yeah. Keep in mind that I'm going through and bulk merging everything that passes, so it's easier for me to just rerun everything that's stale :)

quaeler:
Ok - your work is much appreciated - thanks :- )

quaeler:
Huzzah!

PR #12276, opened 2017-08-14
Add annotations support for tf.estimator.Estimator

### LINKED ISSUE #12249: tf.estimator.Estimator breaks when using python 3.5 type annotations
Minimal example:
```
import tensorflow as tf

def model_fn(features: dict, labels: tf.Tensor, mode: str):
    pass

estimator = tf.estimator.Estimator(model_fn)
```

results in 
```
File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/estimator/estimator.py", line 173, in __init__
    _verify_model_fn_args(model_fn, params)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/estimator/estimator.py", line 742, in _verify_model_fn_args
    args = set(_model_fn_args(model_fn))
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/estimator/estimator.py", line 737, in _model_fn_args
    return tuple(tf_inspect.getargspec(fn).args)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/util/tf_inspect.py", line 45, in getargspec
    if d.decorator_argspec is not None), _inspect.getargspec(target))
  File "/usr/lib/python3.5/inspect.py", line 1045, in getargspec
    raise ValueError("Function has keyword-only arguments or annotations"
ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them
```

### System information
- **OS Platform and Distribution**: Linux Mint 17.2
- **TensorFlow version**: v1.2.0-0-g12f033d 1.2.0
- **Python version**: 3.5

yongtang:
This fix adds annotations support for tf.estimator.Estimator so that the following works in python 3:
```
import tensorflow as tf

def model_fn(features: dict, labels: tf.Tensor, mode: str):
    pass

estimator = tf.estimator.Estimator(model_fn)
```

This fix fixes #12249.

### DIFF
--- a/tensorflow/python/util/tf_inspect.py
+++ b/tensorflow/python/util/tf_inspect.py
@@ -45,6 +45,26 @@ def getargspec(object):  # pylint: disable=redefined-builtin
                if d.decorator_argspec is not None), _inspect.getargspec(target))
 
 
+def getfullargspec(obj):  # pylint: disable=redefined-builtin
+  """TFDecorator-aware replacement for inspect.getfullargspec and fallback to
+  inspect.getargspec in Python 2.
+
+  Args:
+    obj: A callable, possibly decorated.
+
+  Returns:
+    The `FullArgSpec` (`ArgSpec` in Python 2) that describes the signature of
+    the outermost decorator that changes the callable's signature. If the
+    callable is not decorated, `inspect.getfullargspec()`
+    (`inspect.getargspec()` in Python 2) will be called directly on the
+    callable.
+  """
+  spec_fn = getattr(_inspect, 'getfullargspec', getattr(_inspect, 'getargspec'))
+  decorators, target = tf_decorator.unwrap(obj)
+  return next((d.decorator_argspec for d in decorators
+               if d.decorator_argspec is not None), spec_fn(target))
+
+
 def getcallargs(func, *positional, **named):
   """TFDecorator-aware replacement for inspect.getcallargs.
 
--- a/tensorflow/python/estimator/estimator.py
+++ b/tensorflow/python/estimator/estimator.py
@@ -20,6 +20,7 @@ from __future__ import division
 from __future__ import print_function
 
 import copy
+import inspect
 import os
 import tempfile
 
@@ -57,6 +58,30 @@ from tensorflow.python.util import nest
 
 _VALID_MODEL_FN_ARGS = set(
     ['features', 'labels', 'mode', 'params', 'self', 'config'])
+
+
+def _model_fn_args(fn):
+  # The _model_fn is called with args directly so no unwrap like other places
+
+  spec_fn = getattr(inspect, 'getfullargspec', getattr(inspect, 'getargspec'))
+  # Handle callables.
+  if hasattr(fn, '__call__') and inspect.ismethod(fn.__call__):
+    return tuple(spec_fn(fn.__call__).args)
+
+  # Handle functools.partial and similar objects.
+  if hasattr(fn, 'func') and hasattr(fn, 'keywords') and hasattr(fn, 'args'):
+    # Handle nested partial.
+    original_args = _model_fn_args(fn.func)
+    if not original_args:
+      return tuple()
+
+    return tuple([
+        arg for arg in original_args[len(fn.args):]
+        if arg not in set((fn.keywords or {}).keys())
+    ])
+
+  # Handle function.
+  return tuple(spec_fn(fn).args)
 
 
 class Estimator(object):
@@ -690,7 +715,7 @@ class Estimator(object):
     Raises:
       ValueError: if model_fn returns invalid objects.
     """
-    model_fn_args = util.fn_args(self._model_fn)
+    model_fn_args = _model_fn_args(self._model_fn)
     kwargs = {}
     if 'labels' in model_fn_args:
       kwargs['labels'] = labels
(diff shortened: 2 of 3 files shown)


tensorflow/python/util/tf_inspect.py
```diff
@@ -45,6 +45,27 @@ def getargspec(object):  # pylint: disable=redefined-builtin
...
+
+  Returns:
+    The `FullArgSpec` (Python 3) or `ArgSpec` (Python 2)
+    that describes the signature of the outermost decorator that
+    changes the callable's signature. If the callable is not decorated,
+    `inspect.getfullargspec()/inspect.getargspec()`
+    will be called directly on the callable.
+  """
+  # The getfullargspec is called in estimator.Estimator (See issue 12249)
+  decorators, target = tf_decorator.unwrap(object)
```
martinwicke: If you unwrap it, it won't be the outermost object. I.e. if I have a decorator modifying the arguments, we should return the arguments directly, without unwrapping.

tensorflow/python/util/tf_inspect.py
```diff
@@ -45,6 +45,27 @@ def getargspec(object):  # pylint: disable=redefined-builtin
                if d.decorator_argspec is not None), _inspect.getargspec(target))
 
 
+def getfullargspec(object):  # pylint: disable=redefined-builtin
```
martinwicke: Please name this something else, not "object", which is a reserved word.

tensorflow/python/util/tf_inspect.py
```diff
@@ -45,6 +45,27 @@ def getargspec(object):  # pylint: disable=redefined-builtin
...
+  """TFDecorator-aware replacement for inspect.getargspec.
+
+  Args:
+    object: A callable, possibly decorated.
+
+  Returns:
+    The `FullArgSpec` (Python 3) or `ArgSpec` (Python 2)
```
martinwicke: Are these classes compatible enough that all downstream code can use them interchangeably? I.e. is FullArgSpec backwards compatible with ArgSpec?

martinwicke (changes requested):

tensorflow/python/util/tf_inspect.py
yongtang: @martinwicke They are not truly compatible as FullArgSpec has `varkw` while ArgSpec has `keywords`. For example, if `getfullargspec` is used, then in `tensorflow/python/util/deprecation.py`:

https://github.com/tensorflow/tensorflow/blob/325fc2330f4e527b394456ce1d821f2c73b1e7a2/tensorflow/python/util/deprecation.py#L224-L229

The following has to be used:
```
# Cannot use arg_spec.keywords directory
getattr(arg_spec, "keywords", getattr(arg_spec, "varkw"))
```

For that I try to use `util.getfullargspec` only for `estimator.Estimator`. For everything else, the original `util.getargspec` would be used.

tensorflow/python/util/tf_inspect.py
yongtang: @martinwicke The unwrap was done previously so the implementation before PR might have already been problematic. Let me take a look and update the PR accordingly.

yongtang:
@martinwicke I traced through the model_fn and the call with args. The model_fn is actually called with the args directly so as you pointed out the unwrapping may not be appropriate here.

I updated the PR and now the getfullargspec is called to the function directly.

drpngx:
@martinwicke take another look?

sb2nov:
@martinwicke ping

martinwicke:
So this works. I'm sorry I wasn't clear before, it is good to use tf_inspect instead of inspect, so it can use TFDecorator decorated functions as model_fn. 

To do that, use tf_inspect. getargspec, and ideally make that function compatible with py3 by adding the switch for getfullargspec as you did inline.

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/tf_inspect.py#L32

Wee7:
Hi, this also happens to me when I use tf.contrib.variance_scaling_initializer()

`initializer = tf.contrib.variance_scaling_initializer()

  File "/Users/WeeYin/anaconda/lib/python3.5/inspect.py", line 1050, in getargspec
    raise ValueError("Function has keyword-only arguments or annotations"

ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them`

yongtang:
@Wee7 I am not able to reproduce with the current head:
```python
ubuntu@ubuntu:~$ python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
>>> import tensorflow as tf
>>> initializer = tf.contrib.layers.variance_scaling_initializer()
```

```python
ubuntu@ubuntu:~$ python
>>> import tensorflow as tf
>>> initializer = tf.contrib.layers.variance_scaling_initializer()
>>>
```

In case you are able to reproduce, could you open a separate issue and list the details steps to reach the error?

yongtang:
@martinwicke Thanks for the review. The PR has been updated. Please take a look.

martinwicke:
Sorry for the delay, can you update the PR and resolve the conflicts?

martinwicke:
Also, there's one unresolved comment (use of "object" as a arg name)

yongtang:
Thanks @martinwicke. The PR has been rebased and `object` has been changed to `obj` as well. Please take a look.

martinwicke:
Jenkins, test this please.

martinwicke (approved):

caisq:

martinwicke:
ignoring stale Kokoro.

PR #12279, opened 2017-08-14
R1.2

ronendar:


av8ramit:
We have merged all the applicable bugfixes from 1.2 back into master already.

Issue #11636, opened 2017-07-20
stack bidirectional RNNs: swap_memory and time_major args


dieka13:
### Feature Request
adding `swap_memory` and `time_major` arguments to `stack_bidirectional_rnn` and `stack_bidirectional_dynamic_rnn`

from the source i see that both stack_bidirectional function just wraps bidirectional_rnn so it use concatenated output as next bidirectional layer's input. is it intentional that `swap_memory` and `time_major` argument not added?

asimshankar:
@ebrevdo : Any comments?

tensorflowbutler:
It has been 14 days with no activity and the `awaiting tensorflower` label was assigned. Please update the label and/or status accordingly.

ebrevdo:
Looks like it was fixed in #12297.

PR #12297, opened 2017-08-15
fix bugs in `stack_bidirectional_rnn` 

SwordYork:
The condition of the original codes is not related to the constraint.
```
  if initial_states_bw is not None and (not isinstance(cells_bw, list) or
                                        len(cells_bw) != len(cells_bw)):
     raise ValueError(
         "initial_states_bw must be a list of state tensors (one per layer).")
```
I think `initial_states_bw` is written as `cells_bw` by mistake.


ebrevdo (approved):
thank you!

jhseu:
Jenkins, test this please

SwordYork:
Thanks! By the way, `swap_memory` and `time_major` arguments are missing, I think it should be added as suggested in #11636.

Issue #9775, opened 2017-05-09
Feature Request: "training" argument for contrib.rnn.DropoutWrapper like the one in tf.layers.dropout


nelson-liu:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.0.4
- **TensorFlow installed from (source or binary)**: Binary
- **TensorFlow version (use command below)**: 1.1.0

### Feature Request: "training" argument for contrib.rnn.DropoutWrapper for applying dropout depending on train/inference phase.

In [`tf.layers.dropout`](https://www.tensorflow.org/api_docs/python/tf/layers/dropout), the `training` parameter is a handy setting that lets you apply dropout depending on whether the model is training or doing inference. It's very convenient to be able to pass a boolean to the model placeholder and have it automatically do the right thing when it comes to dropout.

Unfortunately, [`tf.contrib.rnn.DropoutWrapper`](https://www.tensorflow.org/api_docs/python/tf/contrib/rnn/DropoutWrapper) does not have this same parameter, and I think it would greatly benefit from it. This is a feature request for it.

I tried implementing it myself with `tf.cond` and either returning the dropped-out outputs/states or the untouched ones, but I couldn't figure out how to share the variables between them in the cond.

girving:
@ebrevdo This seems like a reasonable request.  Can you comment on the best way to add it?

sogangori:
keep_prob = tf.cond(dropOut,lambda:tf.constant(0.9), lambda:tf.constant(1.0))
cells = rnn.DropoutWrapper(cells, output_keep_prob=keep_prob)

tensorflowbutler:
It has been 14 days with no activity and the `awaiting tensorflower` label was assigned. Please update the label and/or status accordingly.

tensorflowbutler:
Nagging Awaiting TensorFlower: It has been 14 days with no activityand the `awaiting tensorflower` label was assigned. Please update the label and/or status accordingly.

ebrevdo:
Does "training" need to be a tensor, or can it be a regular Python bool always?  If so, you don't need the cond() call.

songmeixu:
I understand your confusion. Use a python bool for is_training, it cannot be in graph; while use a tensor.bool for is_training, you need something like tf.is_true(is_training) , but tf without it.

The problem can be solve by:
`
[...]

PR #12316, opened 2017-08-16
training argument for tf.nn.rnn_cell.DropoutWrapper

resec:
This PR resolves issue #9775 

Added a `training` argument to  `tf.nn.rnn_cell.DropoutWrapper`, identical to `tf.layers.dropout`.

Also added to testcase to `tf.contrib.rnn.python.kernel_tests/core_rnn_cell_test.py` for this new argument.


ebrevdo:
We're in the process of fixing a bug in dropout which makes some changes to these lines of code.  Please bear with us a day or two and you may need to resync once that bug is fixed.

resec:
@ebrevdo I merge your bugfix code for state dropout, please let me know if I break anything. My main idea is if `training` is marked `True`, then all dropout intention should be suppressed, so I just move those bugfix lines into the state dropout sub function. Thanks.

drpngx:
@ebrevdo what are you thoughts on this?

sb2nov:
@ebrevdo ping

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -649,7 +650,7 @@ class DropoutWrapper(RNNCell):
 
   def __init__(self, cell, input_keep_prob=1.0, output_keep_prob=1.0,
```
ebrevdo: the `training` argument should be passed to `call`, not `__init__`.  See, e.g., the `Dropout` [layer impl](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/python/layers/core.py#L257).

ebrevdo (changes requested):

tensorflow/python/ops/rnn_cell_impl.py
resec: 👌  let me update it. thanks.

resec: sorry, second thought, is it not easier for user define or control the cell action via a `training` in `init`?

martinwicke:
@resec any update?

resec:
@martinwicke sorry, I did not get what @ebrevdo means, I will check more on the implementation but just too busy for the next few days. Will update soon. Thanks.

resec:
@ebrevdo do you mean that I should create a wrapping function which returns the output tensor, to class `tf.nn.rnn_cell.DropoutWrapper`, just like function `tf.layers.dropout` to class `tf.layers.Dropout`?

But normally we pass the created RNN Cell object to other wrapper methods like `tf.nn.dynamic_rnn`. Maybe we should define the training in Cell construction and adjust its actions when is applied?

Please correct me if I am misunderstanding anything.

drpngx:
@resec could you pull rebase and push again?

@ebrevdo any more comment?

tensorflowbutler:

tensorflowbutler:

ebrevdo:
I believe you're right.

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -1036,23 +1042,35 @@ def _should_dropout(p):
...
+        # which ones to keep.
+        shallow_filtered_substructure = nest.get_traverse_shallow_structure(
+            self._dropout_state_filter, new_state)
+        return self._dropout(new_state, "state",
+                             self._recurrent_state_noise,
+                             self._state_keep_prob,
+                             shallow_filtered_substructure)
+      new_state = utils.smart_cond(self._training,
+                                   dropped_new_state,
+                                   lambda: array_ops.identity(new_state))
```
ebrevdo: do you need the `array_ops.identity` ?

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -1036,23 +1042,35 @@ def _should_dropout(p):
...
     if _should_dropout(self._output_keep_prob):
-      output = self._dropout(output, "output",
+      def dropped_output():
+        return self._dropout(output, "output",
                              self._recurrent_output_noise,
                              self._output_keep_prob)
+      output = utils.smart_cond(self._training,
+                                dropped_output,
+                                lambda: array_ops.identity(output))
```
ebrevdo: ditto

ebrevdo (changes requested):

protoget:
@resec, will you try addressing the comments?

tensorflow/python/ops/rnn_cell_impl.py
resec: In `tf.layers.dropout`, it applies `tf.identity` to `inputs` before returning, but I haven't able to fully understand why it has to be this way. I think (guess) it is to make sure the `input` tensor to be in the same device as the `dropped input (if applied)`.

In here, If I am guessing above correctly, `tf.identity` can be skipped as the `scope` or `device` etc. is decided only when the Cell object is applied?

I think `tf.identity` is a little bit difficult to understand, please correct me if I am wrong, thanks.

resec: @ebrevdo any comment is very appreciated.

tensorflowbutler:
Nagging Assignee @ebrevdo: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

tensorflowbutler:

tensorflowbutler:
It has been 43 days that this pull-request has stalled. Please create a new pull-request with the requested changes.

PR #12317, opened 2017-08-16
remove duplicated code for find master session by handle

horance-liu:
remove all duplicated code for finding master session by session handle.

### DIFF
--- a/tensorflow/core/distributed_runtime/master.cc
+++ b/tensorflow/core/distributed_runtime/master.cc
@@ -116,6 +116,21 @@ void Master::GC() {
   }
 }
 
+MasterSession* Master::FindMasterSession(const string& handle) {
+  MasterSession* session = nullptr;
+  {
+    mutex_lock l(mu_);
+    session = gtl::FindPtrOrNull(sessions_, handle);
+    if (session != nullptr) {
+      session->Ref();
+    }
+  }
+  if (session == nullptr) {
+    done(errors::Aborted("Session ", handle, " is not found."));
+  }
+  return session;
+}
+
 class DeviceFinder {
  public:
   static Status GetRemoteDevices(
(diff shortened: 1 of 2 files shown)


tensorflow/core/distributed_runtime/master.cc
```diff
@@ -116,6 +116,21 @@ void Master::GC() {
...
+  MasterSession* session = nullptr;
+  {
+    mutex_lock l(mu_);
+    session = gtl::FindPtrOrNull(sessions_, handle);
+    if (session != nullptr) {
+      session->Ref();
+    }
+  }
+  if (session == nullptr) {
+    done(errors::Aborted("Session ", handle, " is not found."));
```
mrry: `done` isn't defined in the body of this function, so this won't compile. (Also, it will continue to run the request handler after calling the callback, whereas the request handler should terminate immediately after calling `done` and doing any necessary cleanup.)

tensorflow/core/distributed_runtime/master.cc
```diff
@@ -526,20 +514,7 @@ void Master::ListDevices(const ListDevicesRequest* req,
...
-        mutex_lock l(mu_);
-        session = gtl::FindPtrOrNull(sessions_, req->session_handle());
-        if (session != nullptr) {
-          session->Ref();
-        }
-      }
-      if (session == nullptr) {
-        done(errors::InvalidArgument(
-            "Session ", req->session_handle(),
-            " is not found. Possibly, this master has restarted."));
```
mrry: This change would remove the informative "Possibly, this master has restarted." from the error message. It might be better to have `FindMasterSession()` return `nullptr` in the event that the session isn't found, and allow the caller to send the appropriate error response.

tensorflow/core/distributed_runtime/master.h
```diff
@@ -94,6 +94,9 @@ class Master {
   // Cleanup unused session.
   void GC();
 
+  // Find master session.
+  MasterSession* FindMasterSession(const string& handle);
```
mrry: This method increments the refcount on the returned `MasterSession*`, so the comment should indicate that.

mrry (changes requested):
Thanks for the contribution. Unfortunately, it doesn't look like the code will compile, and there are some issues with the design. I've included comments that should help you get this PR back on track!

horance-liu:
@mrry I'm sorry for my mistake, I have fixed it.

mrry (approved):
Looks good to me!

jhseu:
Jenkins, test this please

jhseu:
Jenkins, test this please

Issue #12309, opened 2017-08-16
Tensorflow installation fails on python 3.3.6


jeevandev:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: See below
- **TensorFlow installed from (source or binary)**: Tried with pip3
- **TensorFlow version (use command below)**: 
- **Python version**: 3.3.6
- **Bazel version (if compiling from source)**:
- **CUDA/cuDNN version**:
- **GPU model and memory**:
- **Exact command to reproduce**:

You can collect some of this information using our environment capture script:
== cat /etc/issue ===============================================
Linux 11bb6e68305e 4.9.36-moby #1 SMP Wed Jul 12 15:29:07 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
VERSION="14.04.5 LTS, Trusty Tahr"
VERSION_ID="14.04"

== are we in docker =============================================
Yes

== compiler =====================================================
c++ (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4
Copyright (C) 2013 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

== uname -a =====================================================
Linux 11bb6e68305e 4.9.36-moby #1 SMP Wed Jul 12 15:29:07 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

== check pips ===================================================

== check for virtualenv =========================================
True

== tensorflow import ============================================
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named 'tensorflow'

== env ==========================================================
LD_LIBRARY_PATH is unset
DYLD_LIBRARY_PATH is unset

== nvidia-smi ===================================================
./tf_env_collect.sh: line 105: nvidia-smi: command not found

== cuda libs  ===================================================

### Describe the problem
[...]

PR #12335, opened 2017-08-16
Correct minimum Python 3 version in install instruction

carlthome:
https://github.com/tensorflow/tensorflow/issues/12309

### DIFF
--- a/tensorflow/docs_src/install/install_linux.md
+++ b/tensorflow/docs_src/install/install_linux.md
@@ -231,7 +231,7 @@ Python is automatically installed on Ubuntu.  Take a moment to confirm
 versions is already installed on your system:
 
   * Python 2.7
-  * Python 3.3+
+  * Python 3.4+
 
 The pip or pip3 package manager is *usually* installed on Ubuntu.  Take a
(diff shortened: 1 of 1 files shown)


caisq:
We do have python 3.3 wheels available on PyPI: https://pypi.python.org/pypi/tensorflow

If we make this change, we should remove the wheels from pypi in the next release.

carlthome:
It's actually NumPy that made 3.4 the minimum, but I imagine that will cause some confusion for people that are careless with their `pip install` commands.

Another alternative would be to change `numpy>=1.11.0` to `numpy==1.11.0`

gunan:
Actually we need to go higher than 1.12 in numpy due to our new autograd dependency.
So we wont be able to pin it to 1.11.

I am ok with this change and also having 3.3 wheel on pypi for a while.
We can use it as a transition period.

gunan (approved):

hugovk:
This page https://www.tensorflow.org/install/install_linux still says "Python 3.3+". 

Please could you deploy these updates to avoid further confusion?

Thank you.

PR #12341, opened 2017-08-17
Constructor with InputStream only for TensorFlowInferenceInterface

resec:
Added a InputStream Constructor for `TensorFlowInferenceInterface`, which support loading model from network or many other situations that we can only have a `InputStream` of the model.

One thing worth mentioning is that the private `modelName` member is not set this new constructor, but the field is private and only use in the original constructor (which I prefer this member should be deleted), I think it should be fine.


andrewharp (approved):

jhseu:
Jenkins, test this please

jhseu:
The tests fail with a syntax error. Mind taking a look?

jhseu:
Jenkins, test this please

jhseu:
Jenkins, test this please

resec:
@andrewharp @jhseu thanks for the fix

jhseu:
Jenkins, test this please

alexdennis:
Was just thinking about adding something like this. Glad to see someone beat me to the punch :)

PR #12354, opened 2017-08-17
[XLA] Break the Slice test into different tests for each data type

DavidNorman:
This allows our backend (and others) to disable types which they do not support.  Previously all types were tested by the same test function.

### DIFF
--- a/tensorflow/compiler/xla/tests/slice_test.cc
+++ b/tensorflow/compiler/xla/tests/slice_test.cc
@@ -193,13 +193,28 @@ class SliceR1Test : public ClientLibraryTestBase,
   }
 };
 
-XLA_TEST_P(SliceR1Test, DoIt) {
-  Run<float>(GetParam());
-  Run<double>(GetParam());
-  Run<uint32>(GetParam());
-  Run<int32>(GetParam());
-  Run<uint64>(GetParam());
-  Run<int64>(GetParam());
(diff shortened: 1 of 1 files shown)


tensorflow/compiler/xla/tests/slice_test.cc
```diff
@@ -193,13 +193,28 @@ class SliceR1Test : public ClientLibraryTestBase,
...
 };
 
-XLA_TEST_P(SliceR1Test, DoIt) {
-  Run<float>(GetParam());
-  Run<double>(GetParam());
-  Run<uint32>(GetParam());
-  Run<int32>(GetParam());
-  Run<uint64>(GetParam());
-  Run<int64>(GetParam());
+XLA_TEST_P(SliceR1Test, DoIt_float) {
```
hawkinsp: The convention in other tests is to use the XLA names for types (F32, F64, ...).

Other than that, looks good.

tensorflow/compiler/xla/tests/slice_test.cc
```diff
@@ -193,13 +193,28 @@ class SliceR1Test : public ClientLibraryTestBase,
...
+Run<float>(GetParam());
```
hawkinsp: Please indent the function body two spaces. Same everywhere.

hawkinsp (approved):

hawkinsp:
Jenkins, test this please.

Issue #11725, opened 2017-07-24
tf_cnn_benchmarks.py stuck when running with multiple GPUs and ImageNet data with protocol grpc+verbs


shamoya:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No, running tf_cnn_benchmarks.py from benchmarks repo
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04.2 LTS
- **TensorFlow installed from (source or binary)**: Unmodified source with RDMA Verbs enabled
- **TensorFlow version (use command below)**: 1.3.0-rc0
- **Python version**: 2.7.12
- **Bazel version (if compiling from source)**: 0.5.1
- **CUDA/cuDNN version**: 8.0/6
- **GPU model and memory**: NVIDIA Tesla P100 PCIe 16GB (8 per node)
- **Exact command to reproduce**: 

PS: CUDA_VISIBLE_DEVICES='' python tf_cnn_benchmarks.py --ps_hosts 12.12.12.43:20000 --worker_hosts 12.12.12.44:20000,12.12.12.41:20000 --batch_size=64 --model=inception3 --variable_update=parameter_server --local_parameter_device=cpu --job_name=ps --task_index=0 --server_protocol grpc+verbs

Worker0: CUDA_VISIBLE_DEVICES='0,1,2,3,4,5,6,7' python tf_cnn_benchmarks.py --ps_hosts 12.12.12.43:20000 --worker_hosts 12.12.12.44:20000,12.12.12.41:20000 --batch_size=64 --model=inception3 --variable_update=parameter_server --local_parameter_device=cpu --job_name=worker --task_index=0 --num_gpus=8 --data_dir=/data/imagenet_data/ --train_dir=/data/imagenet_train/ --server_protocol grpc+verbs

Worker1: CUDA_VISIBLE_DEVICES='0,1,2,3,4,5,6,7' python tf_cnn_benchmarks.py --ps_hosts 12.12.12.43:20000 --worker_hosts 12.12.12.44:20000,12.12.12.41:20000 --batch_size=64 --model=inception3 --variable_update=parameter_server --local_parameter_device=cpu --job_name=worker --task_index=1 --num_gpus=8 --data_dir=/data/imagenet_data/ --train_dir=/data/imagenet_train/ --server_protocol grpc+verbs

- **RDMA driver version**: MLNX_OFED_LINUX-4.1-1.0.2.0

[...]

Issue #11825, opened 2017-07-27
Why remove tolerate_dup_recv from LocaRendezvous.


junshi15:
I did not find a PR for this [patch](https://github.com/tensorflow/tensorflow/commit/cbfd50ff0f01e1825922230a8bc6e5766da98dd7). Maybe this discussion belongs to that particular commit page. If so, I will move it there.

Can somebody explain the rationale behind this patch? Current contrib/verbs uses exactly this flag to transfer tensors. Note it is set true [here](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/contrib/verbs/rdma_rendezvous_mgr.cc#L34). Removing it totally breaks verbs. I am asking if this patch is absolutely needed since there is no simple workaround for contrib/verbs without this flag.

shamoya:
@junshi15 @reedwm @poxvoculi - How do we proceed here ? 
The tolerate_dup_recv is needed by the verbs code.

junshi15:
Can we revert [this patch](https://github.com/tensorflow/tensorflow/commit/cbfd50ff0f01e1825922230a8bc6e5766da98dd7)? Just looking at the patch alone, I feel it is cosmetic (cleaning up a parameter that is not used in the core code). Or maybe I missed something important.

junshi15:
While we are looking for the motivation behind this change, we will also investigate how to adapt verbs code to accommodate this change.

poxvoculi:
It was removed because we no longer needed it for our internal networking, where it was needed to deal with hardware errors in an earlier generation of machines.  Turning this feature on comes at a cost of higher memory use, since tensors backing sent values cannot be garbage collected until the end of the step.  Also, there may be some upcoming feature whose implementation was made more difficult by allowing this option.

I should have noticed you were using this, and pinged you: sorry!  How important is this feature to verbs support?  It is a convenience, or is it not possible to make the service reliable without it?

junshi15:
@poxvoculi This is a corner stone of the current verbs design, where the receiving side could make multiple Recv attempts. Since the pinned memory is shared, a single Recv attempt may not be successful if the shared memory is used by other transmission.

But no worries, let us see if we can do without it.

byronyi:
[...]

PR #12361, opened 2017-08-17
Verbs hang fix

yanivbl6:
Replaces the sync wrappers for device to device operations in verbs_util, with a call to the async function with a callback function. Resolves Issue #11725 

As mentioned in the issue:
I think the problem rises because the Sync deviceToDevice operation blocks the thread, preventing the earlier Async Device to Device operation from finishing- which, for some reason, blocks the later operation.

I've tested this fix in origin/r1.3 since origin/master is currently broken.

2nd commit only removes the unused functions from verbs_utils. It is not required for validity. 

@junshi15 , @shamoya , @byronyi , can you please go over this?

### DIFF
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -707,7 +707,6 @@ void RdmaTensorBuffer::SendNextItem() {
       bool can_memcpy = DataTypeCanUseMemcpy(in.dtype());
       // string tensor needs to be serialized
       Tensor copy;
-      StringPiece copy_buf;
       TensorProto proto;
       if (src_dev->tensorflow_gpu_device_info() &&
           (!send_args.alloc_attrs.on_host())) {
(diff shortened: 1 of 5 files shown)


junshi15:
Thanks for the fix. Is there any reason you did not get rid of CopyCPUTensorToGPUSync in verbs_util.h?

yanivbl6:
CopyCPUTensorToGPUSync is still being used by RdmaRemoteRendezvous [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/verbs/rdma_rendezvous_mgr.cc#L116)
I am still studying it, so I avoided changing this call. But if my understanding of the bug was correct it is definitely a risk.

byronyi:
Nice work!

I think I am using the same sync wrappers (yes all of us are lazy) in my GDR patch. I might as well take a closer look on this before someone raises another issue :)

junshi15:
@yanivbl6 `CopyCPUTensorToGPUSync` can be an issue if not fixed. Can we fix the one in RdmaRemoteRendezvous as well? thanks.

yanivbl6:
Sure, I am on it.

junshi15:
I only have access to two boxes with fairly old OS (RHEL6.5) and CUDA 7.5, which limited me to TF1.1. I modify the patch to fit TF1.1, but got the following error during benchmark test.
`error message: Cannot parse tensor from proto`
Not sure if this is due to my adaptation or it is a problem with this patch. 
@shamoya , @bkovalev can you test this patch on TF1.2 or 1.3, if you have access to newer OS/CUDA. Thanks.

shamoya:
Very good work @yanivbl6 , patch looks good.
Plz just update when u finish checking this patch on TF1.2 and TF1.3.

yanivbl6:
I ran successful tf_cnn_benchmarks with r1.3 ( 4 hosts (worker + ps), 8X4 GPUs, Imagenet data, Resnet-50 model).

shamoya:
Hi @poxvoculi @junshi15, 

Can we merge this ?
And when it's merged to master, can we cherry-pick it to r1.3 (since master verbs code is broken) ? 

Ido

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -721,109 +720,133 @@ void RdmaTensorBuffer::SendNextItem() {
...
+              });
         } else {
-          // "val" is on a GPU. Uses GPUUtil to fill the proto.
-          s = VerbsUtil::SetProtoFromGPUSync(
-              in, src_dev, send_args.device_context, &proto, is_dead);
-          CHECK(s.ok()) << "set proto from gpu sync";
+          // "val" is on a GPU. No longer uses GPUUtil to fill the proto, use
+          // aync instead
+          GPUUtil::SetProtoFromGPU(
+              in, src_dev, send_args.device_context, &proto, is_dead,
```
poxvoculi: Someday you should return an error status instead of check failing, especially if there's a chance this is a transient error.  (Applies to the CHECK a couple lines below.)

poxvoculi (approved):

tensorflow/contrib/verbs/rdma.cc
junshi15: Agreed. The error handling can be improved. Maybe we should send error status back to the caller (like [here](https://github.com/petewarden/tensorflow_makefile/blob/49c08e4d4ff3b6e7d99374dc2fbf8b358150ef9c/tensorflow/core/distributed_runtime/rpc/grpc_worker_service.cc#L410-L413)) and let the caller decide what to do, either re-try or abort. 

However, for this patch, let's not address it as returning the error status to the caller/requester may require an extra signal path.

This patch looks good to me otherwise.

junshi15:
@poxvoculi What's the policy for patching a branch? Given the fix for verbs on master will take time, it will be useful to patch TF 1.3 (and maybe 1.2).

jhseu:
Jenkins, test this please

jhseu:
We're not making anymore changes to v1.3. This will appear in TF 1.4, though.

shamoya:
Thanks @jhseu 
The problem is that the verbs code is broken in master for now due to [this](https://github.com/tensorflow/tensorflow/issues/11825) issue.
We are working on a fix for, which may require some time.
For now, people who want to work with the RDMA Verbs need to patch r1.3 manually with this commit.
It's not so crazy, but less comfortabale.
I thought v1.3.0 is already out, so cherry-picking this to r1.3 doesn't mean a lot.
or are you updating the v1.3 binary version with r1.3 from time to time ?

PR #12369, opened 2017-08-17
Add compression support to RecordInput

### LINKED ISSUE #12344: Feature request: RecordInput to support gzipped tfrecord files.
Currently [RecordInput](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/record-input) doesn't support gzipped tfrecord files, would be great to have this supported.

yongtang:
This fix tries to fix the request raised in #12344 so that it is possible to process RecordInput with compressions.

An attr of `compression_type` has been added.

Additional tests have been created to cover the changes.

This fix fixes #12344.

### DIFF
--- a/tensorflow/core/api_def/base_api/api_def_RecordInput.pbtxt
+++ b/tensorflow/core/api_def/base_api/api_def_RecordInput.pbtxt
@@ -43,5 +43,12 @@ END
 The batch size.
 END
   }
+  attr {
+    name: "compression_type"
+    description: <<END
+The type of compression for the file. Currently ZLIB and
+GZIP are supported. Defaults to none.
+END
+  }
   summary: "Emits randomized records."
 }
--- a/tensorflow/core/kernels/record_input_op.cc
+++ b/tensorflow/core/kernels/record_input_op.cc
@@ -37,6 +37,8 @@ class RecordInputOp : public OpKernel {
     GETATTR(int64, file_parallelism);
     GETATTR(int64, batch_size);
 #undef GETATTR
+    string compression_type;
+    ctx->GetAttr("compression_type", &compression_type);
 
     RecordYielder::Options yopts;
     yopts.file_pattern = file_pattern;
@@ -44,6 +46,7 @@ class RecordInputOp : public OpKernel {
     yopts.bufsize = file_buffer_size;
     yopts.file_shuffle_shift_ratio = file_shuffle_shift_ratio;
     yopts.parallelism = file_parallelism;
+    yopts.compression_type = compression_type;
     yielder_ = std::unique_ptr<RecordYielder>(new RecordYielder(ctx, yopts));
 
     batch_size_ = batch_size;
--- a/tensorflow/core/kernels/record_yielder.cc
+++ b/tensorflow/core/kernels/record_yielder.cc
@@ -206,7 +206,10 @@ void RecordYielder::ShardLoop(Shard* shard) {
       shard->status = errors::InvalidArgument("Can't open ", filename);
       break;
     }
-    io::RecordReader rdr(file.get());
+    io::RecordReaderOptions options =
+        io::RecordReaderOptions::CreateRecordReaderOptions(
+            opts_.compression_type);
+    io::RecordReader rdr(file.get(), options);
     uint64 offset = 0;
     string record;
     while (true) {
--- a/tensorflow/core/kernels/record_yielder.h
+++ b/tensorflow/core/kernels/record_yielder.h
@@ -78,6 +78,8 @@ class RecordYielder {
     // Uses these many concurrent tfrecord iterators to iterate through
     // tfrecords.
     int32 parallelism = 1;
+
+    string compression_type;
   };
 
   explicit RecordYielder(OpKernelConstruction* context,
--- a/tensorflow/core/ops/data_flow_ops.cc
+++ b/tensorflow/core/ops/data_flow_ops.cc
@@ -2497,6 +2497,7 @@ REGISTER_OP("RecordInput")
     .Attr("file_buffer_size: int = 10000")
     .Attr("file_parallelism: int = 16")
     .Attr("batch_size: int = 32")
+    .Attr("compression_type: string = ''")
     .SetIsStateful()
     .SetShapeFn(shape_inference::UnknownShape)
     .Doc(R"doc(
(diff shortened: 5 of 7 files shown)


drpngx (commented):
Nice, thanks!

Am I correct to understand that this compresses the entire file, not just the payload? Any thoughts on how to make that compress only the payload and leave the headers alone?

yongtang:
Thanks @drpngx for the review. The current PR utilizes the existing implementation of `io::RecordReader`which allows a compression type as the flag.

To add compression to the payload but expose the header might be possible. It may require additional implementations on `io::RecordReader`.

Let me play with it and see if I could create another PR for that.

drpngx:
OK, I think that's the same as other behavior in TF, but I'm curious to see what you'll find.

/CC @skye

martinwicke:
@yongtang any update? What should we do with this PR?

yongtang:
@martinwicke Given that the original issue #12344 was to support gzipped tfrecord files, I think it probably make sense to limit the scope of this PR and merge if there is no outstanding issues.

In the review @drpngx raised an interesting point of having payload compressed while exposing the header. I haven't spend much time on that though I think it is worth at least exploring.

Maybe we could move forward to merge with this PR (assuming no other outstanding issue) so that the original issue #12344 could be closed? I could open a separate issue to track the alternative approach. You can assign the new issue to me and I will spend time on it.

yongtang:
Added a new issue #14342 to cover the payload only compression.

drpngx (approved):
Still need API review for the new option.

martinwicke:
I see no API change in this PR which is surprising. Are you sure it does what it's intended to? Or do you just need to update the golden files for the API test?

drpngx:
Strange, there should be a `compression_type`. Maybe there's some indirection of some kind.

drpngx:
It looks like the op is only available through [here](https://www.github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/data_flow_ops.py#L2267), @yongtang can you check?

yongtang:
@martinwicke @drpngx The C++ kernel ops update is through:

https://github.com/tensorflow/tensorflow/pull/12369/files#diff-9dd26225c3129302e6f60e2b119f48d2R2478

and for python, it is through the `__init__()` method of the class `RecordInput` and `TFRecordInput`:

https://github.com/tensorflow/tensorflow/pull/12369/files#diff-156604dc2cdee8c1c0cd03edb1e8a7fdR2155

https://github.com/tensorflow/tensorflow/pull/12369/files#diff-156604dc2cdee8c1c0cd03edb1e8a7fdR2178

As it is through class constructor, it is not a python ops like `decode_raw`.

Maybe that is the reason there is no api compatibility issue?

yongtang:
Not able to see the build error. Maybe it is an infrastructure failure?

drpngx:
Ah, there is an API change detected, good.

```
2017-12-15 18:38:23.754138: I tensorflow/core/api_def/api_test.cc:233] Skipping GenerateBigQueryReaderPartitions
tensorflow/core/api_def/api_test.cc:298: Failure
      Expected: golden_api_defs_str
      Which is: "op {\n  graph_op_name: \"RecordInput\"\n  out_arg {\n    name: \"records\"\n    description: <<END\nA tensor of shape [batch_size].\nEND\n  }\n  attr {\n    name: \"file_pattern\"\n    description: <<END\nGlob pattern for the data files.\nEND\n  }\n  attr {\n    name: \"file_random_seed\"\n    description: <<END\nRandom seeds used to produce randomized records.\nEND\n  }\n  attr {\n    name: \"file_shuffle_shift_ratio\"\n    description: <<END\nShifts the list of files after the list is randomly\nshuffled.\nEND\n  }\n  attr {\n    name: \"file_buffer_size\"\n    description: <<END\nThe randomization shuffling buffer.\nEND\n  }\n  attr {\n    name: \"file_parallelism\"\n    description: <<END\nHow many sstables are opened and concurrently iterated over.\nEND\n  }\n  attr {\n    name: \"batch_size\"\n    description: <<END\nThe batch size.\nEND\n  }\n  summary: \"Emits randomized records.\"\n}\n"
To be equal to: new_api_defs_str
      Which is: "op {\n  graph_op_name: \"RecordInput\"\n  out_arg {\n    name: \"records\"\n    description: <<END\nA tensor of shape [batch_size].\nEND\n  }\n  attr {\n    name: \"file_pattern\"\n    description: <<END\nGlob pattern for the data files.\nEND\n  }\n  attr {\n    name: \"file_random_seed\"\n    description: <<END\nRandom seeds used to produce randomized records.\nEND\n  }\n  attr {\n    name: \"file_shuffle_shift_ratio\"\n    description: <<END\nShifts the list of files after the list is randomly\nshuffled.\nEND\n  }\n  attr {\n    name: \"file_buffer_size\"\n    description: <<END\nThe randomization shuffling buffer.\nEND\n  }\n  attr {\n    name: \"file_parallelism\"\n    description: <<END\nHow many sstables are opened and concurrently iterated over.\nEND\n  }\n  attr {\n    name: \"batch_size\"\n    description: <<END\nThe batch size.\nEND\n  }\n  attr {\n    name: \"compression_type\"\n    description: <<END\nThe type of compression for the file. Currently ZLIB and\nGZIP are supported. Defaults to none.\nEND\n  }\n  summary: \"Emits randomized records.\"\n}\n"
With diff:
@@ +42,11 @@
     description: <<END
 The batch size.
+END
+  }
+  attr {
+    name: \"compression_type\"
+    description: <<END
+The type of compression for the file. Currently ZLIB and
+GZIP are supported. Defaults to none.
 END
   }

To update golden API files, run tensorflow/core/api_def/update_api_def.sh.
```

yongtang:
Thanks @drpngx. The PR has been updated and pushed with changes from `tensorflow/core/api_def/update_api_def.sh`.

Most of the tests passed except `Linux CPU Tests (Python 3)`, which seems to be an infrastructure issue:
```
FATAL: command execution failed
java.nio.channels.ClosedChannelException
	at org.jenkinsci.remoting.protocol.impl.ChannelApplicationLayer.onReadClosed(ChannelApplicationLayer.java:208)
```

Please take a look.

drpngx:
Jenkins, test this please.

drpngx:
Double-checking with @martinwicke for API if it's OK to merge.

martinwicke:
Yes, these classes are not part of the public API at all, so this is good to go.

PR #12399, opened 2017-08-18
Eager API inclusion to the dynamic library build target

eaplatanios:


alextp (approved):

alextp:
Jenkins, test this please.

caisq:
There seems to be a buildifier error:
https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/6134/consoleFull

> 472d471
> <         "//tensorflow/c/eager:c_api",
> 474a474
> >         "//tensorflow/c/eager:c_api",
> Please fix manually or run buildifier <file> to auto-fix.

Tabs?

eaplatanios:
@caisq I'm not sure what the problem is exactly. Should I be using tabs for the BUILD files? Currently they are all indented with spaces (4 spaces per indentation level).

caisq:
@eaplatanios I'm not sure either. The diff in the error message seems to show two exactly identical lines, which is confusing to me, too. Yes, you should be using 4 spaces, instead of tabs. I'll let you know when I come up other thoughts.

tensorflow/BUILD
```diff
@@ -469,6 +469,7 @@ cc_binary(
     linkshared = 1,
     deps = [
         "//tensorflow/c:c_api",
+        "//tensorflow/c/eager:c_api",
```
caisq: I see what the problem is: this line should be added immediately before `//tensorflow/core:tensorflow`, instead of here. Buildifier is complaining about the sort order.

caisq (changes requested):

tensorflow/BUILD
eaplatanios: Thanks for this! I had no idea what the problem was. Is there a way for me to run the buildifier tool locally or is it not public?

caisq: Please see: https://github.com/tensorflow/tensorflow/blob/master/CONTRIBUTING.md#running-sanity-check

caisq:

caisq (approved):

PR #12407, opened 2017-08-18
Added a library to convert TensorFlow GraphDef to XLA SessionModule.

keveman:
I have only added the minimal code without lot of comments. I wanted to make sure this approach is acceptable. I have made the xlagen library public, so a python library can be built externally by using TensorFlow as a submodule. Please let me know if this is acceptable and I will spruce up the PR with more documentation, comments, and tests.

### DIFF
--- a/tensorflow/compiler/xla/BUILD
+++ b/tensorflow/compiler/xla/BUILD
@@ -8,6 +8,7 @@ package_group(
         "//tensorflow/compiler/...",
         "//tensorflow/contrib/tpu/...",
         "//tensorflow/contrib/xla_tf_graph/...",
+        "//tensorflow/contrib/xlagen/...",
     ],
 )
 
--- a/tensorflow/contrib/xlagen/BUILD
+++ b/tensorflow/contrib/xlagen/BUILD
@@ -1,0 +1,57 @@
+# Description:
+#   contains parts of TensorFlow that are experimental or unstable and which are not supported.
+
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+licenses(["notice"])  # Apache 2.0
+
+exports_files(["LICENSE"])
+
+filegroup(
+    name = "all_files",
+    srcs = glob(
+        ["**/*"],
+        exclude = [
+            "**/METADATA",
+            "**/OWNERS",
+        ],
+    ),
+)
+
+cc_library(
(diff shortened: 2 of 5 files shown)


keveman:
This approach may not be as tenable as I originally anticipated. Even with this in contrib, it is still hard to make use of it from another project that uses TensorFlow as a (git) submodule, and swig wrap it. This is mostly because of the build system, and swig wrapping this causes symbol collisions with tensorflow (can't import tensorflow and this library together). Please let me know what the best way forward for making it easy to extract XLA graphs from GraphDefs in Python.

jmchen-g:
Adding myself :)

tatatodd:
Hi Manjunath!

FYI you might be interested in the following code in contrib, which looks like it does what you want:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/xla_tf_graph/xla_tf_graph_util.h#L62

I don't think I understand the Python / swig-wrapping issues.  Perhaps you can open a bug to discuss that further, if the problem still exists with the existing contrib code?

keveman:
Thanks @tatatodd for the response. I did notice that code before I embarked on this. However, we wanted to be able to call this from Python, so I needed going from GraphDef -> SessionModule. Actually, the better piece of code that I can reuse is a combination of `InitGraph` and `CompileGraph` in your https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/aot/compile.h. Also, the BUILD visibility of the various components is not suitable for making tensorflow a git submodule under a different project. Here is what would be ideal for our (Cerebras) use case:

    with tf.Graph().as_default() as g:
      a = tf.placeholder(tf.int32, shape=[2, 2])
      b = tf.placeholder(tf.int32, shape=[2])
      c = tf.add(a, b, name="c")
    # xla_graph would be a xla::SessionModule
    xla_graph = tf2xla.GraphDefToXla(g.as_graph_def(), ["c:0"])

Now, `tf2xla` can be something that we build on our own using the stock `tensorflow` repo as a git submodule, or if you think this is going to be more generally useful, this can be in `tf` module itself. The former would require changing the visibility of some of the libraries to `//visibility:public`. Either way, I am willing to do the necessary work. But please let me know what you think the best approach is.

keveman:
Closing this, will make do with a fork for now.

Issue #8100, opened 2017-03-05
XLA Design :: Incorporate Polyhedral Compilation (through LLVM Polly).


annanay25:
Polyhedral Compilation is a method of modeling iterations of loop nests into points on a multidimensional space (Determined by the loop nest depth). A detailed description can be found on the website as given [here](http://polyhedral.info/).

Is it possible to pass the deeply nested kernels (for-loops) modeled in TensorFlow via XLA-JIT (or independently) through [Polly](http://polly.llvm.org/) that will extract dependence information and perform scheduling using Polyhedral compilation? Tensorflow Ops can be reproduced and modeled to basic math operators and can be applied to points on the Integer Polyhedra. 

This can lead to significant speedup in program execution. However, since Polyhedral compilation can be expensive, it can lead to increase in compile time. (Trade-off in compile time)

I am a student working on Polyhedral Compilation and would love to contribute to this if it spans out!

prb12:
We have open-sourced the XLA compiler precisely so that it is possible for external contributors and researchers to experiment with this sort of idea.   

@eliben @learyg May be able to comment further.

eliben:
@annanay25 XLA is exactly the kind of compiler that could greatly benefit form polyhedral optimizations. I would say certainly try it! You could fork the project and try applying Polly in your fork and see the results you get. @jpienaar for more insights

annanay25:
@eliben @prb12 Thanks for the replies! 

I was trying to put together a compilation workflow to incorporate Polly in XLA. The glitch being that we need the Graph composed by XLA (the entire computation graph of the program and not a subgraph) and pass this to Polly - in a way that it can extract the SCoPs properly. The reason for this is because Polly can then be used to schedule the entire program at one go, which could otherwise be expensive if performed on each subgraph.

It was not very obvious to me as to how I could get this done. Any help on this would be great!

Couldn't be happier to see XLA Open Source!

jpienaar:
Definitely would be good to try XLA with Polly. Forking and modifying the CPU/GPU compiler backend might be the easiest. The [CG]PuCompiler's Compile function takes as input the HloModule with all the HloComputations. From there you'd have access to all the graphs translated from TF to XLA and can perform the optimizations you wish. You could look at entire graphs or just operations within the graph (for example, look at IrEmitter). The requirement of the entire graph is more of an optimization (large input graph, better return on optimization) requirement than a correctness one (correct me if wrong).

[...]

PR #12408, opened 2017-08-19
[XLA][WIP] Add support for Polyhedral compilation through Polly

annanay25:
Added Bazel build files for compiling Polly with LLVM to enable Polyhedral compilation, as discussed in #8100


annanay25:
cc @BrianRetford @meinersbur @tobig

annanay25:
@googlebot `I signed it!`

annanay25:
cc @hawkinsp 

Could you please look into this?

tensorflow/compiler/xla/service/cpu/compiler_functor.cc
```diff
@@ -211,6 +214,11 @@ void CompilerFunctor::AddOptimizationPasses(
   builder.SLPVectorize = opt_level_ > 1;
 
   builder.populateFunctionPassManager(*function_passes);
+
+  llvm::PassRegistry &Registry = *llvm::PassRegistry::getPassRegistry();
+  ::polly::initializePollyPasses(Registry);
+  ::polly::registerPollyPasses(*module_passes);
```
sanjoy: This needs to be guarded by an off by default flag for now.

Google C++ style is `llvm::PassRegistry &Registry = *llvm::PassRegistry::getPassRegistry();`

tensorflow/compiler/xla/service/cpu/BUILD
```diff
@@ -295,6 +295,7 @@ cc_library(
         "@llvm//:object",
         "@llvm//:support",
         "@llvm//:target",
+        "@llvm//:polly",
```
sanjoy: Can we only conditionally pull in this dependency?  Not everyone will use polly, so it makes sense to avoid pulling in the dependency for folks who don't need it.

sanjoy (changes requested):

tensorflow/compiler/xla/service/cpu/BUILD
annanay25: Thanks for reviewing :)

How can I add a conditional in the cc_library() rule?

I could add a flag to check if the user wants to add Polly's passes in the `tensorflow/compiler/xla/service/cpu/compiler_functor.cc` file.

tensorflow/compiler/xla/service/cpu/compiler_functor.cc
annanay25: I read this 4 times but I do not see the difference. Could you point it out?

sanjoy: Sorry, my bad.  I meant to say `llvm::PassRegistry& registry = *llvm::PassRegistry::getPassRegistry();`.

annanay25: Alright sure, will make the change. :)

tensorflow/compiler/xla/service/cpu/BUILD
sanjoy: To add a conditional take a look at how if_mkl works as an example.  Some links to help you get started:

 - Definition for if_mkl: https://github.com/tensorflow/tensorflow/blob/master/third_party/mkl/build_defs.bzl
 - The corresponding config_setting: https://github.com/tensorflow/tensorflow/blob/master/third_party/mkl/BUILD
 - Documentation for config_setting: https://docs.bazel.build/versions/master/be/general.html#config_setting

However, before you go ahead and do a bunch of work, I think we should take a step back and reach agreement on whether Polly constitutes a supported configuration.  By "supported configuration" I mean that it is linked into TensorFlow by default and breaking a Polly test case is sufficient cause to revert a patch or to hold a push.  My perspective on this is that it isn't fair to ask Google and other TF developers to pay the cost of supporting Polly till we've shown a clear benefit (better runtime performance, basically) from Polly; and we should start off by designating the TF-with-Polly build as "experimental" -- it won't be built by default and the onus of keeping it working will fall on you.  Once we are able to demonstrate value from Polly, and have community support to address breakages and regressions in Polly, we can discuss enabling it by default as a supported configuration.

tobig: @sanjoy this is a very good comment. I think having experimental optimizations, such as in LLVM having we have experimental backends is a very good idea. Polly certainly is not a -enabled-and-supported-by-default feature.

annanay25: @sanjoy @tobig Thank you for this discussion. I am working on getting the benchmarking complete ASAP.

bhack:
Seems that there is also a recent effort to mainline [Polly in LLVM](http://lists.llvm.org/pipermail/llvm-dev/2017-September/117063.html) /cc @hfinkel

drpngx:
@annanay25 any chance to pull rebase and address comments? Thank you.

annanay25:
@drpngx, yes, will do it shortly.

martinwicke:
@annanay25 any update?

martinwicke:
@sanjoy did the new commits resolve your comments?

martinwicke:
@annanay25 @arpith-jacob could you resolve the conflicts?

tensorflow/compiler/xla/service/cpu/build_defs.bzl
```diff
@@ -10,3 +10,14 @@ def runtime_copts():
       "//tensorflow:android": ["-O2"],
       "//conditions:default": []
   }))
+
+def if_polly(basic_deps):
```
sanjoy: This should not be called `if_polly`.  Given what it does today, it should be called `append_polly_dep_if_enabled`, though I think it is better to have a `polly_component_if_available` function that returns `["@llvm//:polly"]` if polly is enabled else returns an empty list; and explicitly appending `polly_component_if_available()` in the `deps` for various rules.

tensorflow/compiler/xla/service/cpu/compiler_functor.cc
```diff
@@ -22,6 +22,9 @@ limitations under the License.
 #include <utility>
 #include <vector>
 
+#include "external/llvm/tools/polly/include/polly/LinkAllPasses.h"
```
sanjoy: These need to be conditional on whether polly is enabled right?

tensorflow/compiler/xla/service/cpu/compiler_functor.cc
sanjoy: This hasn't been addressed.

tensorflow/workspace.bzl
```diff
@@ -537,6 +537,7 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
 
   # TODO(phawkins): currently, this rule uses an unofficial LLVM mirror.
   # Switch to an official source of snapshots if/when possible.
+  # FIXME: The LLVM Mirror should now have polly checked-out as well.
```
sanjoy: Not clear what this FIXME is getting at -- is it that the LLVM mirror has polly checked out and we should figure out some way to avoid that?  Or the converse?

third_party/llvm/expand_isl_cmake_vars.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+#
```
sanjoy: How is this different from tensorflow/third_party/llvm/expand_cmake_vars.py ?

third_party/llvm/llvm.BUILD
```diff
@@ -18,7 +19,7 @@ load(
     "template_rule",
 )
 
-package(default_visibility = ["@%ws%//tensorflow/compiler/xla:internal"])
+package(default_visibility = ["@%ws%//tensorflow:internal"])
```
sanjoy: Why do you need to increase the visibility here?

third_party/llvm/llvm.BUILD
```diff
@@ -128,6 +129,125 @@ cmake_vars = {
...
+    "HAVE_SYS_PARAM_H": 1,
+    "HAVE_SYS_RESOURCE_H": 1,
+    "HAVE_SYS_STAT_H": 1,
+    "HAVE_SYS_TIME_H": 1,
+    "HAVE_SYS_TYPES_H": 1,
+    "HAVE_TERMIOS_H": 1,
+    "HAVE_UNISTD_H": 1,
+    "HAVE_ZLIB_H": 1,
+
+    # Features
```
sanjoy: How did you generate these?  If there are subsets (like `# Features`) that have to be common with `cmake_vars` by construction, we should common them up.

third_party/llvm/llvm.bzl
```diff
@@ -85,6 +85,26 @@ def cmake_var_string(cmake_vars):
   return " ".join([_quote("{}={}".format(k, str(v)))
                    for (k, v) in cmake_vars.items()])
 
+def expand_isl_cmake_vars(name, src, dst, cmake_vars):
```
sanjoy: Is this the same as `expand_cmake_vars`?  If not, please add a comment on how they're different; otherwise please just use `expand_cmake_vars`.

sanjoy (changes requested):

drpngx:
@annanay25 any luck with addressing the comments? Could you pull rebase and push again?

tensorflowbutler:

third_party/llvm/expand_isl_cmake_vars.py
annanay25: The `_CMAKE_VAR_REGEX` for parsing isl source files is not the same as the one for LLVM.

```
$ diff expand_cmake_vars.py expand_isl_cmake_vars.py
27c27
< _CMAKE_VAR_REGEX = re.compile(r"\${([A-Za-z_0-9]*)}")
---
> _CMAKE_VAR_REGEX = re.compile(r"@([A-Za-z_0-9]*)@")
```

rmlarsen:
@annanay25 any progress on this? Please rebase.

tensorflow/workspace.bzl
annanay25: No, I meant to say that once the Polly build file integration is complete, the Polly source code should be checked-out in the LLVM mirror used for TensorFlow.

third_party/llvm/llvm.BUILD
annanay25: Copied the LLVM features from `cmake_vars` but added isl-specific features to the second one.

third_party/llvm/llvm.bzl
annanay25: The parsing tool used would be `expand_isl_cmake_vars` in this function (which has a different regex for parsing isl files as mentioned above).

tensorflow/compiler/xla/service/cpu/build_defs.bzl
annanay25: Polly would be added as a dependency in just one component. Do you still suggest we take the second approach?

tensorflow/compiler/xla/service/cpu/compiler_functor.cc
annanay25: This has not been addressed.

I've been looking for a way to define a macro like so - `ENABLE_POLLY` with bazel, in which case I could wrap this block of code with -

    #ifdef ENABLE_POLLY
    #include "external/llvm/tools/polly/include/polly/LinkAllPasses.h"
    .
    .
    #endif

Suggestions?

third_party/llvm/expand_isl_cmake_vars.py
sanjoy: In that case can you please extract out a helper function to a common .py file and call that from both `expand_cmake_vars.py` and this one, passing `_CMAKE_VAR_REGEX` as an argument?  This duplicated code will otherwise become a maintenance issue.

I'd also be fine with a "master" `expand_cmake_vars.py` that takes `_CMAKE_VAR_REGEX` and as a command line argument (which will be passed in via the bazel rule).

Why is `_CMAKE_VAR_REGEX` different here btw?  Can we do something to make them the same?

+CC @hawkinsp just in case I'm misleading you

tensorflow/compiler/xla/service/cpu/compiler_functor.cc
sanjoy: Just like you have `append_polly_dep_if_enabled` you should be able to write a `POLLY_DEFINES_IF_ENABLED` which will expand to `"ENABLE_POLLY"` if polly is enabled (and empty string otherwise).  You can then use `defines` (https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library.defines) to pass this in to the `"compiler_functor"` build rule.

tensorflow/compiler/xla/service/cpu/build_defs.bzl
```diff
@@ -10,3 +10,14 @@ def runtime_copts():
       "//tensorflow:android": ["-O2"],
       "//conditions:default": []
   }))
+
+def append_polly_dep_if_enabled(basic_deps):
```
sanjoy: Might be better to use the same pattern as in `runtime_copts` -- return `["@llvm//:polly"]` or not, and then append the result of `polly_deps()` explicitly in the calling rule.

sanjoy (changes requested):

tensorflowbutler:

tensorflowbutler:
Nagging Assignee @jpienaar: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

tensorflowbutler:
Nagging Assignee @jpienaar: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

bhack:
Can @tensorflowbutler count real days? This second one was probably 28 days so it doesn't make sense to count bot only posts.

martinwicke:
Good point. @av8ramit how hard would it be to ignore bot posts for the purposes of counting days only (the 2 weeks interval in nagging is still good).

av8ramit:
Sure I'll add this to our upcoming feature set. Thanks @bhack

tensorflowbutler:

PR #12415, opened 2017-08-19
Refactor create worker sessions

horance-liu:
Refactoring to OO style for worker group for create worker sessions, and it is more clear for implements.


mrry:
This PR replaces a single function and 58 lines of code with two structs and 122 lines of code to do the same thing. In its current form, I wouldn't be inclined to accept it. Can you explain why it is an improvement?

(The original function is quite dense, admittedly, but I think a better solution would be to add one or two comments at the appropriate points.)

/cc @saeta as the original author.

horance-liu:
I‘m not familiar with google code style before, but I like "One Level of Abstraction per Function/Class". After modify it, every class and function will do one thing, and I will read code quickly and discover potential problems easily.

 I pull request it to make sure the commit criterion, and to decide more commits for this class in the future. I believe that the number of lines will decrease, because it will generate more reusable classes and functions.

horance-liu:
@mrry I give up this PR, it's not a successful case, and I don't satisfy the result. I will continue to commit to tensorflow. 

Thanks.

jhseu:
@horance-liu Closing as per your comment. Thank you for your contributions to TF :)

PR #12447, opened 2017-08-21
reuse RefCountIsOne function member

horance-liu:
reuse RefCountIsOne function member, and replace the same implement.


jhseu (approved):

jhseu:
Jenkins, test this please

PR #12448, opened 2017-08-21
Implement BiasAddGradHelper.

bpiel:
Registers the BiasAddGrad op as the gradient for BiasAdd by implementing BiasAddGradHelper.

Unlike the other tested ops in nn_grad_tests.cc, BiasAdd takes two arguments. The test I added does test both arguments, but not simultaneously. Not sure if that's acceptable. Guidance appreciated.

This is only my second PR to TensorFlow. When giving feedback, please assume that I barely know what I'm doing.


asimshankar:
Thanks for the PR
@suharshs : Mind taking a look?

tensorflow/cc/gradients/nn_grad_test.cc
```diff
@@ -113,5 +113,23 @@ TEST_F(NNGradTest, SeluGrad) {
...
+    {-0.9f, -0.5f, 0.1f, 0.3f, 0.5f},
+    {5});
+  auto y1 = BiasAdd(scope_, x1, bias1);
+  auto x2  = test::AsTensor<float>(
+    {-0.9f, -0.7f, -0.5f, -0.3f, -0.1f, 0.1f, 0.3f, 0.5f, 0.7f, 0.9f,
+     -0.9f, -0.7f, -0.5f, -0.3f, -0.1f, 0.1f, 0.3f, 0.5f, 0.7f, 0.9f},
+    {4, 5});
+  auto bias2 = Placeholder(scope_, DT_FLOAT, Placeholder::Shape(bias_shape));
+  auto y2 = BiasAdd(scope_, x2, bias2);
+  RunTest(x1, shape, y1, shape);
```
suharshs: You can create another overload for RunTest that takes multiple "x_shapes" like here in array_grad_test: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/cc/gradients/array_grad_test.cc#L44

This should allow you to just specify shapes for the inputs and run one RunTest. Here is an example of a test that does that: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/cc/gradients/array_grad_test.cc#L56

tensorflow/cc/gradients/nn_grad.cc
```diff
@@ -95,6 +95,20 @@ Status SeluGradHelper(const Scope& scope, const Operation& op,
 }
 REGISTER_GRADIENT_OP("Selu", SeluGradHelper);
 
+Status BiasAddGradHelper(const Scope& scope, const Operation& op,
+                         const std::vector<Output>& grad_inputs,
+                         std::vector<Output>* grad_outputs) {
+  string data_format;
+  BiasAddGrad::Attrs input_attrs;
+  auto dx_1 = BiasAddGrad(scope, grad_inputs[0], input_attrs);
```
suharshs: I think the BiasAddGrad should happen after line 105

suharshs (changes requested):
Thanks for the PR! Just a couple of changes.

bpiel:
@suharshs back to you

tensorflow/cc/gradients/nn_grad_test.cc
```diff
@@ -47,6 +47,15 @@ class NNGradTest : public ::testing::Test {
     EXPECT_LT(max_error, 1e-4);
   }
 
+  void RunTest(const OutputList& xs, const std::vector<TensorShape>& x_shapes,
+               const OutputList& ys, const std::vector<TensorShape>& y_shapes) {
+    TF_ASSERT_OK(scope_.status());
+    float max_error;
+    TF_ASSERT_OK(
+        ComputeGradientError(scope_, xs, x_shapes, ys, y_shapes, &max_error));
+    EXPECT_LT(max_error, 1e-3);
```
suharshs: Super minor: Lets make this 1e-4 to match the rest of the file. The array grads were 1e-3 because of some issue IIRC.

suharshs (changes requested):
Looks good after this minor change.

suharshs (approved):

jhseu:
Jenkins, test this please

PR #12459, opened 2017-08-21
Respect `optimize` parameter in `zeros_like`

### LINKED ISSUE #12436: zeros_like doesn't fully respect the optimize argument
[The definition of `zeros_like`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/array_ops.py#L1463) is:

```python
def zeros_like(tensor, dtype=None, name=None, optimize=True):
  with ops.name_scope(name, "zeros_like", [tensor]) as name:
    tensor = ops.convert_to_tensor(tensor, name="tensor")

    if tensor.shape.is_fully_defined():
      # We can produce a zeros tensor independent of the value of 'tensor',
      # since the shape is known statically.
      return zeros(tensor.shape, dtype=dtype or tensor.dtype, name=name)

    if dtype is not None and dtype != tensor.dtype:
      return zeros(
          shape_internal(tensor, optimize=optimize), dtype=dtype, name=name)
    else:
      return gen_array_ops._zeros_like(tensor, name=name)
```

We can see that if the shape of `tensor` is already known, the `optimize` parameter is ignored, which is inconsistent with the [documented behavior](https://www.tensorflow.org/api_docs/python/tf/zeros_like).

yongtang:
This fix address the issue in #12436 where `optimize` flag was not respected when shape is fully defined for `zeros_like`.

This fix fixes #12436.

### DIFF
--- a/tensorflow/python/ops/array_ops.py
+++ b/tensorflow/python/ops/array_ops.py
@@ -1481,7 +1481,8 @@ def zeros_like(tensor, dtype=None, name=None, optimize=True):
     # For now, variant types must be created via zeros_like; as we need to
     # pass the input variant object to the proper zeros callback.
 
-    if tensor.shape.is_fully_defined() and tensor.dtype != dtypes.variant:
+    if optimize and tensor.shape.is_fully_defined() and \
+        tensor.dtype != dtypes.variant:
       # We can produce a zeros tensor independent of the value of 'tensor',
       # since the shape is known statically.
       return zeros(tensor.shape, dtype=dtype or tensor.dtype, name=name)


yongtang:
The PR has been rebased to fix the merging conflict and pushed. Please take a look.

benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please

drpngx:
Jenkins, test this please.

drpngx:
Error on MacOS was a lot larger than on other platforms:

```
[ RUN      ] NNGradTest.MaxPoolGradHelper
tensorflow/cc/gradients/nn_grad_test.cc:39: Failure
Expected: (max_error) < (2e-4), actual: 0.134928 vs 0.0002
```

drpngx:
Yay! All green.

PR #12469, opened 2017-08-22
replace auto deduction for master reset member function

horance-liu:


jhseu:
(Closing for the same reason as the other change)

PR #12471, opened 2017-08-22
replace prune subgraph algorithm with auto deduction for ranged-for

horance-liu:


jhseu:
(Closing for the same reason as the other change).

PR #12476, opened 2017-08-22
replace explicit type decl with auto deduction for device manager

horance-liu:


jhseu (commented):
I'm not sure we want to do this more generally. Only in cases where it helps readability.

Explicit types are more common throughout the codebase.

jhseu:
Thanks for the pull request, though!

PR #12477, opened 2017-08-22
replace loop with ranged-for from worker's get_status

horance-liu:


jhseu (approved):

jhseu:
Jenkins, test this please

PR #12479, opened 2017-08-22
find worker session by handle for worker

horance-liu:
extract common template method to find worker session by handle, and remove duplicated implements.

### DIFF
--- a/tensorflow/core/distributed_runtime/worker.cc
+++ b/tensorflow/core/distributed_runtime/worker.cc
@@ -25,6 +25,12 @@ limitations under the License.
 
 namespace tensorflow {
 
+// Find worker session by session handle. It's will be not null.
+template<typename Request>
+WorkerSession* Worker::FindWorkerSession(const Request* request) const {
+  return env_->session_mgr->WorkerSessionForSession(request->session_handle());
+}
+
 Worker::Worker(WorkerEnv* env)
     : env_(env), cancellation_manager_(new CancellationManager) {}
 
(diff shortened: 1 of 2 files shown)


tensorflow/core/distributed_runtime/worker.cc
```diff
@@ -25,6 +25,12 @@ limitations under the License.
 
 namespace tensorflow {
 
+// Find worker session by session handle. It's will be not null.
```
mrry: Replace "It's will be not null." with "REQUIRES: `request` is not null. The return value is guaranteed not to be null."

tensorflow/core/distributed_runtime/worker.cc
```diff
@@ -25,6 +25,12 @@ limitations under the License.
...
+template<typename Request>
+WorkerSession* Worker::FindWorkerSession(const Request* request) const {
```
mrry: I'm not convinced that the `template` method is more readable here. Perhaps this could just take a `const string&` and the various callsites could pass `request->session_handle()`?

mrry (changes requested):

tensorflow/core/distributed_runtime/worker.cc
horance-liu: OK, I fixed it.

tensorflow/core/distributed_runtime/worker.cc
horance-liu: But it will repeat many times for 'request->session_handle()', so I rename `FindWorkerSession(req)` to `FindWorkerSessionBy(req)`, does this improve readability?

mrry: I think repeating a (self explanatory) accessor method call four times is still more readable than a template method. A casual reader might be confused into thinking that `RegisterGraphRequest`, `DeregisterGraphRequest`, `RunGraphRequest` etc. have a common superclass, whereas this templating only works because they share a coincidentally-named member.

Please switch `FindWorkerSession()` to taking a string.

drpngx:
@horance-liu did you address all comments?

sb2nov:
@mrry ping for review

tensorflow/core/distributed_runtime/worker.cc
gunan: Looks like this comment is still not addressed?
Please correct me if I am wrong.

martinwicke:
@mrry can you check the changes please?

mrry:
@martinwicke My [comment from 8/27](https://github.com/tensorflow/tensorflow/pull/12479#discussion_r135431498) is still unresolved.

martinwicke:
Closing as stalled.

PR #12481, opened 2017-08-22
add lvalue ref for auto deduction, and avoid to copy object.

horance-liu:


mrry (approved):

mrry:

PR #12495, opened 2017-08-22
Fix ProfileContext location in README

bryant1410:


panyx0718:
Are you sure you can call it like: tf.python.profiler.profile_context.ProfileContext?
I only expose it via contrib on purpose.

bryant1410:
> Are you sure you can call it like: tf.python.profiler.profile_context.ProfileContext?
> I only expose it via contrib on purpose.

Yeah, after 2b51e0ba27af69c914a7523d9aae232de09e3206 it's accessible that way, I have tested it. Furthermore, the README way like it is now doesn't work.

By looking at af23ae65db2585f4a18d0bc5f21f15e94805aa4f, it seems to be migrated to core.

panyx0718:
@bryant1410 
I just build from source and install the latest package

Python 3.4.3 (default, Nov 17 2016, 01:08:31) 
[GCC 4.8.4] on linux
>>> import tensorflow as tf
>>> tf.contrib.tfprof.ProfileContext
<class 'tensorflow.python.profiler.profile_context.ProfileContext'>
>>> tf.python.profiler.profile_context.ProfileContext
AttributeError: 'module' object has no attribute 'python'

drpngx:
@bryant1410 please fix, pull rebase and push again. Thanks!

sb2nov:
@bryant1410 ping

gunan:
I have just confirmed that the original readme is correct.
Maybe we had a bad release which you saw the issue?
Or is it possible maybe on windows our API is messed up somehow?
```
>>> tf.python.profiler.profile_context.ProfileContext
AttributeError: 'module' object has no attribute 'python'
>>> tf.contrib.tfprof.ProfileContext
<class 'tensorflow.python.profiler.profile_context.ProfileContext'>
>>> tf.__version__
'1.4.0'
```

Therefore, I will close this PR. Thank you very much for the proposed fix, and if you can find the root cause of the issue, I am happy to work with you to fix the issue on our end.

Issue #12489, opened 2017-08-22
error: token ""__CUDACC_VER__ is no longer supported.  Use __CUDACC_VER_MAJOR__, __CUDACC_VER_MINOR__, and __CUDACC_VER_BUILD__ instead."" is not valid in preprocessor expressions


lunkyguy:
bazel build -c opt --config=cuda //tensorflow/tools/pip_package:build_pip_package

Linux Ubuntu 16.04
Bazel version (if compiling from source):0.5.2
CUDA/cuDNN version:9.0/7.0
GPU model and memory:GTX 965m

ERROR: /home/zhouzl/tensorflow-master/tensorflow/contrib/seq2seq/BUILD:51:1: error while parsing .d file: /home/zhouzl/.cache/bazel/_bazel_zhouzl/418dff85d676df3fe7a9d3de8f68c1df/execroot/org_tensorflow/bazel-out/local_linux-opt/bin/tensorflow/contrib/seq2seq/_objs/python/ops/_beam_search_ops_gpu/tensorflow/contrib/seq2seq/kernels/beam_search_ops_gpu.cu.pic.d (No such file or directory).
In file included from /usr/local/cuda-9.0/bin/../targets/x86_64-linux/include/common_functions.h:50:0,
                 from /usr/local/cuda-9.0/bin/../targets/x86_64-linux/include/cuda_runtime.h:115,
                 from <command-line>:0:
/usr/local/cuda-9.0/bin/../targets/x86_64-linux/include/crt/common_functions.h:64:24: error: token ""__CUDACC_VER__ is no longer supported.  Use __CUDACC_VER_MAJOR__, __CUDACC_VER_MINOR__, and __CUDACC_VER_BUILD__ instead."" is not valid in preprocessor expressions
 #define __CUDACC_VER__ "__CUDACC_VER__ is no longer supported.  Use __CUDACC_VER_MAJOR__, __CUDACC_VER_MINOR__, and __CUDACC_VER_BUILD__ instead."
                        ^
/usr/local/cuda-9.0/bin/../targets/x86_64-linux/include/crt/common_functions.h:64:24: error: token ""__CUDACC_VER__ is no longer supported.  Use __CUDACC_VER_MAJOR__, __CUDACC_VER_MINOR__, and __CUDACC_VER_BUILD__ instead."" is not valid in preprocessor expressions
[...]

PR #12502, opened 2017-08-22
Support for CUDA 9.0

nluehr:
For review by @zheng-xq.

Add explicit __syncwarp to bias_op
 - Makes warp-synchronous code safe on Volta
Add sync mask to __shfl intrinsics
Add libdevice bytecode paths for CUDA 9
 - In CUDA 9, all supported architectures are merged into a single file
Update code gating for CUDA 9
Add sm_70 to the lookup table used by XLA
Change the default sm arch from 20 to 30.
Fix for NVPTX not yet supporting sm_70
Remove unnecessary cuda decorators from defaulted constructors
Use updated NCCL for CUDA 9 fp16 support

### DIFF
--- a/tensorflow/compiler/xla/service/gpu/ir_emitter_unnested.cc
+++ b/tensorflow/compiler/xla/service/gpu/ir_emitter_unnested.cc
@@ -996,7 +996,7 @@ Status IrEmitterUnnested::EmitRowReduction(
   //   for (shuffle_distance = 16; shuffle_distance > 0; shuffle_distance /= 2)
   //     partial_result = Reducer(
   //         partial_result,
-  //         __shfl_down(partial_result, shuffle_distance));
+  //         __shfl_down_sync(CUDA_WARP_ALL, partial_result, shuffle_distance));
   //   if (lane_id == 0)
   //     AtomicReducer(&output[y], partial_result);
   // }
--- a/tensorflow/compiler/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
+++ b/tensorflow/compiler/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
@@ -71,7 +71,17 @@ const int kDefaultInlineThreshold = 1100;
 // Gets the libdevice filename for a particular compute capability.  When
 // presented with a GPU we don't recognize, we just return the libdevice from
 // compute_20.
-static string GetLibdeviceFilename(std::pair<int, int> compute_capability) {
+static string GetLibdeviceFilename(const string& libdevice_dir_path,
+                                   std::pair<int, int> compute_capability) {
+  // Since CUDA 9.0, all GPU versions are included in a single file
+  const char* unified_libdevice_filename = "libdevice.10.bc";
+  std::vector<string> unified_libdevice_files;
+  tensorflow::Env::Default()->GetMatchingPaths(
+      tensorflow::io::JoinPath(libdevice_dir_path, unified_libdevice_filename),
+      &unified_libdevice_files);
+  if( unified_libdevice_files.size() == 1 ) {
+    return unified_libdevice_filename;
+  }
   // There are only four libdevice files: compute_{20,30,35,50}.  Each GPU
   // version gets mapped to one of these.  Note in particular that sm_60 and
   // sm_61 map to libdevice.compute_30.
@@ -101,7 +111,7 @@ static string GetLibdeviceFilename(std::pair<int, int> compute_capability) {
 }
 
 // Gets the GPU name as it's known to LLVM for a given compute capability.  If
-// we see an unrecognized compute capability, we return "sm_20".
+// we see an unrecognized compute capability, we return "sm_30".
 static string GetSmName(std::pair<int, int> compute_capability) {
   static auto* m = new std::map<std::pair<int, int>, int>({{{2, 0}, 20},
                                                            {{2, 1}, 21},
@@ -114,8 +124,10 @@ static string GetSmName(std::pair<int, int> compute_capability) {
                                                            {{5, 3}, 53},
                                                            {{6, 0}, 60},
                                                            {{6, 1}, 61},
-                                                           {{6, 2}, 62}});
-  int sm_version = 20;
+                                                           {{6, 2}, 62},
+                    // TODO: Change this to 70 once LLVM NVPTX supports it
+                                                           {{7, 0}, 60}});
+  int sm_version = 30;
   auto it = m->find(compute_capability);
   if (it != m->end()) {
     sm_version = it->second;
@@ -306,7 +318,8 @@ tensorflow::Status LinkLibdeviceIfNecessary(
 
   llvm::Linker linker(*module);
   string libdevice_path = tensorflow::io::JoinPath(
-      libdevice_dir_path, GetLibdeviceFilename(compute_capability));
+      libdevice_dir_path, GetLibdeviceFilename(libdevice_dir_path,
+                                               compute_capability));
   TF_RETURN_IF_ERROR(tensorflow::Env::Default()->FileExists(libdevice_path));
   VLOG(1) << "Linking with libdevice from: " << libdevice_path;
   std::unique_ptr<llvm::Module> libdevice_module =
--- a/tensorflow/core/kernels/bias_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/bias_op_gpu.cu.cc
@@ -173,15 +173,20 @@ __global__ void BiasGradNCHW_SharedAtomics(const T* output_backprop,
   // Accumulate the results in the shared memory into the first element.
   // No syncthreads is needed since this is only in the same warp.
   int32 thread_index = threadIdx.x;
-  if (thread_index < 16) s_data[thread_index] += s_data[thread_index + 16];
-  if (thread_index < 8) s_data[thread_index] += s_data[thread_index + 8];
-  if (thread_index < 4) s_data[thread_index] += s_data[thread_index + 4];
-  if (thread_index < 2) s_data[thread_index] += s_data[thread_index + 2];
-  if (thread_index < 1) s_data[thread_index] += s_data[thread_index + 1];
-
-  // The first thread writes out the accumulated result to the global location.
-  if (thread_index == 0) {
-    CudaAtomicAdd(bias_backprop + bias_index, T(s_data[0]));
+  if (thread_index < 16) {
+    s_data[thread_index] += s_data[thread_index + 16];
+    __syncwarp(0xFFFF);
+    if (thread_index < 8) s_data[thread_index] += s_data[thread_index + 8];
+    __syncwarp(0xFF);
+    if (thread_index < 4) s_data[thread_index] += s_data[thread_index + 4];
+    __syncwarp(0xF);
+    if (thread_index < 2) s_data[thread_index] += s_data[thread_index + 2];
+    __syncwarp(0x3);
+    if (thread_index == 0) {
+      T val = T(s_data[0] + s_data[1]);
+      // The first thread writes out the accumulated result to global location.
+      CudaAtomicAdd(bias_backprop + bias_index, val);
+    }
   }
 }
 
--- a/tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
@@ -22,6 +22,7 @@ limitations under the License.
 #include "tensorflow/core/platform/types.h"
 #include "tensorflow/core/util/cuda_kernel_helper.h"
 #include "tensorflow/core/util/tensor_format.h"
+#include "external/cub_archive/cub/util_ptx.cuh"
 
 #if !defined(_MSC_VER)
 #define UNROLL _Pragma("unroll")
@@ -1015,6 +1016,21 @@ __global__ void __launch_bounds__(640, 2)
   }
 }
 
+// Device function to compute sub-warp sum reduction for a power-of-two group of
+// neighboring threads.
+template<int kWidth, typename T>
+__device__ __forceinline__ T WarpSumReduce(T val) {
+  // support only power-of-two widths.
+  assert(__popc(kWidth) == 1);
+  int sub_warp = cub::LaneId() / kWidth;
+  int zeros = sub_warp * kWidth;
+  unsigned mask = ((1U << kWidth) - 1) << zeros;
+  for (int delta = kWidth / 2; delta > 0; delta /= 2) {
+    val += CudaShuffleXor(mask, val, delta);
+  }
+  return val;
+}
+
 // CUDA kernel to compute the depthwise convolution backward w.r.t. filter in
 // NHWC format, tailored for small images up to 32x32. Stride and depth
 // multiplier must be 1. Padding must be 'SAME'. Only use this kernel if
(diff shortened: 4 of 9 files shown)


jhseu:

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
```diff
@@ -1153,8 +1154,13 @@ __launch_bounds__(1024, 2) void DepthwiseConv2dBackpropFilterGPUKernelNHWCSmall(
       if (filter_depth < in_depth) {
         T val = accum_data[i];
         // Warp-accumulate the pixels of the same depth from the accumulator.
+        int lane_id;
+        asm volatile ("mov.u32 %0, %laneid;" : "=r"(lane_id));
```
jlebar: In the past we've used the threadIdx % 32.  Does that no longer work?

It's kind of a layering violation to put inline assembly here.  If there's no good way to get the lane ID other than inline asm, it should be abstracted away in a function.

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
```diff
@@ -1116,6 +1116,7 @@ __launch_bounds__(1024, 2) void DepthwiseConv2dBackpropFilterGPUKernelNHWCSmall(
 
     // Note: the condition to reach this is uniform across the entire block.
     __syncthreads();
+    unsigned active_threads = CudaBallot(0xFFFFFFFF, depth_in_range);
```
jlebar: Can we make 0xFFFFFFFF a constant?  It seems pretty bad if someone were to leave off one of the F's.  Same elsewhere.

Or perhaps the order of args to CudaBallot should be reversed, and 0xFFFFFFFF should be the default for the (now) second argument?

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
```diff
@@ -1153,8 +1154,13 @@ __launch_bounds__(1024, 2) void DepthwiseConv2dBackpropFilterGPUKernelNHWCSmall(
...
         // Warp-accumulate the pixels of the same depth from the accumulator.
+        int lane_id;
+        asm volatile ("mov.u32 %0, %laneid;" : "=r"(lane_id));
+        int sub_warp = lane_id / kAccumPixels;
+        int zeros = sub_warp * kAccumPixels;
+        unsigned mask = ((1U << kAccumPixels) - 1) << zeros;
         for (int delta = kAccumPixels / 2; delta > 0; delta /= 2) {
-          val += CudaShuffleDown(val, delta);
+          val += CudaShuffleXor(mask, val, delta);
         }
```
jlebar: This whole chunk of code is now a lot more complicated than it was, and is repeated elsewhere.  Can we abstract it into a function?

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -26,6 +26,16 @@ limitations under the License.
...
 
+#if defined(__CUDACC_VER_MAJOR__) && __CUDACC_VER_MAJOR__ < 9
+__device__ inline void __syncwarp(unsigned mask=0xFFFFFFFF) {}
+
+#define __ballot_sync(mask, predicate)              __ballot(predicate)
+#define __shfl_sync(mask, val, srcLane, width)      __shfl(val, srcLane, width)
+#define __shfl_down_sync(mask, val, delta, width)   __shfl_down(val, delta, width)
+#define __shfl_up_sync(mask, val, delta, width)     __shfl_up(val, delta, width)
+#define __shfl_xor_sync(mask, val, laneMask, width) __shfl_xor(val, laneMask, width)
+#endif
```
jlebar: Please add comments explaining what's going on here, and in particular what "sync" means.

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -26,6 +26,16 @@ limitations under the License.
 #include "tensorflow/core/platform/stream_executor.h"
 #include "tensorflow/core/platform/types.h"
 
+#if defined(__CUDACC_VER_MAJOR__) && __CUDACC_VER_MAJOR__ < 9
```
jlebar: TensorFlow defines a CUDA_VERSION macro; please use that so this code will be portable to other compilers.

jlebar (dismissed):
Thank you for doing this; it will be a big help.

What tools did you use to figure out what needed to be changed and to verify that the code now works?

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
zheng-xq: +1 about abstracting away in a function

sjperkins: `cub::LaneId()` does this and I believe cub is a dependency now.

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
sjperkins: Available in cub I believe

tensorflow/core/util/cuda_kernel_helper.h
nluehr: TF_CUDA_VERSION is a string containing a decimal value (e.g. "8.0") rather than an INT. This makes it unwieldy for inequality comparisons.

jlebar: What about the CUDA_VERSION macro?  https://github.com/tensorflow/tensorflow/search?utf8=%E2%9C%93&q=cuda_version&type=

nluehr: Perhaps I'm being dense, but I don't see where that macro is defined, and it doesn't seem to be available in CUDA compilation units.

jlebar: CUDA_VERSION is defined by nvidia's headers, in cuda.h.

(FYI I'm going to be OOO for a week; hopefully you can get assistance from the TF team in the meantime.)

tfboyd:
Asking team about CLA for associate corporation.  Should not take long to sort out.  Thank you for starting the review.

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
nluehr: cuda_kernel_helper.h now defines CUDA_WARP_ALL for use in the warp-function also defined there.

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
nluehr: I wasn't initially sure of the ordering of treads in thread_idx, but on closer inspection, thread_idx % 32 would give the lane correctly. However, after abstracting the warp reduction into a separate function, it is preferable to not pass a separate thread index / lane id anyway.

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
nluehr: Done.

tensorflow/core/util/cuda_kernel_helper.h
nluehr: Done

jthestness:
Nice work! Thanks for these changes. I have these cuda-9 and cudnn-v7 (#12503) patches built and working.

Notes for others trying this out:
1) This patch alone either requires prior versions of cuDNN (up to v6), or cuDNN must be disabled (either option does not support new mixed-precision fp16 NVIDIA Volta V100). Another patch is required to also support cuDNN v7 (and Volta). @nluehr has posted a patch for cuDNN 7 support here: #12503

2) Eigen must also be updated to work with cuda-9 per this thread: #12489. For now, @nluehr has a modified Eigen archive available with these changes to `tensorflow/workspace.bzl`:
```
diff --git a/tensorflow/workspace.bzl b/tensorflow/workspace.bzl
index e5c2ab5..1b3314a 100644
--- a/tensorflow/workspace.bzl
+++ b/tensorflow/workspace.bzl
@@ -169,12 +169,14 @@ def tf_workspace(path_prefix="", tf_repo_name=""):
   native.new_http_archive(
       name = "eigen_archive",
       urls = [
-          "http://mirror.bazel.build/bitbucket.org/eigen/eigen/get/f3a22f35b044.tar.gz",
-          "https://bitbucket.org/eigen/eigen/get/f3a22f35b044.tar.gz",
+          "http://bazel-mirror.storage.googleapis.com/bitbucket.org/ncluehr/eigen/get/6cc72014d888.tar.gz",
+          "https://bitbucket.org/ncluehr/eigen/get/6cc72014d888.tar.gz",
       ],
-      sha256 = "ca7beac153d4059c02c8fc59816c82d54ea47fe58365e8aded4082ded0b820c4",
-      strip_prefix = "eigen-eigen-f3a22f35b044",
+      sha256 = "549fc01af565d185123016c12baf997eeb462b088c1885a4c3dfaf2a5c82d677",
+      strip_prefix = "ncluehr-eigen-6cc72014d888",
       build_file = str(Label("//third_party:eigen.BUILD")),
+
+
   )
 
   native.new_http_archive(
```

nluehr:
I signed it!

zheng-xq:
LGTM

zheng-xq (commented):
LGTM

tensorflow/core/kernels/bias_op_gpu.cu.cc
```diff
@@ -173,10 +173,10 @@ __global__ void BiasGradNCHW_SharedAtomics(const T* output_backprop,
   // Accumulate the results in the shared memory into the first element.
   // No syncthreads is needed since this is only in the same warp.
   int32 thread_index = threadIdx.x;
-  if (thread_index < 16) s_data[thread_index] += s_data[thread_index + 16];
-  if (thread_index < 8) s_data[thread_index] += s_data[thread_index + 8];
-  if (thread_index < 4) s_data[thread_index] += s_data[thread_index + 4];
-  if (thread_index < 2) s_data[thread_index] += s_data[thread_index + 2];
+  if (thread_index < 16) s_data[thread_index] += s_data[thread_index + 16]; __syncwarp(0xFFFF);
```
jlebar: Please put the __syncwarp calls on a separate line.  As written it looks like the if statement applies to both statements, but it doesn't.

tensorflow/compiler/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
```diff
@@ -114,8 +124,10 @@ static string GetSmName(std::pair<int, int> compute_capability) {
                                                            {{5, 3}, 53},
                                                            {{6, 0}, 60},
                                                            {{6, 1}, 61},
-                                                           {{6, 2}, 62}});
-  int sm_version = 20;
+                                                           {{6, 2}, 62},
+                                                           // TODO: Change this to 70 once LLVM NVPTX supports it
```
jlebar: Please keep lines <80 characters.

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
```diff
@@ -1015,6 +1016,21 @@ __global__ void __launch_bounds__(640, 2)
...
+// Device function to compute sub-warp reductions for a power-of-two group of
+// neighboring threads.
+template<int kWidth, typename T>
+__device__ __forceinline__ T WarpReduce(T val) {
+  // support only power-of-two widths.
+  assert(__popc(kWidth) == 1);
+  int sub_warp = cub::LaneId() / kWidth;
+  int zeros = sub_warp * kWidth;
+  unsigned mask = ((1U << kWidth) - 1) << zeros;
+  for (int delta = kWidth/2; delta>0; delta/=2) {
```
jlebar: TensorFlow style adds spaces around operators.

tensorflow/compiler/xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -996,7 +996,7 @@ Status IrEmitterUnnested::EmitRowReduction(
   //   for (shuffle_distance = 16; shuffle_distance > 0; shuffle_distance /= 2)
   //     partial_result = Reducer(
   //         partial_result,
-  //         __shfl_down(partial_result, shuffle_distance));
+  //         __shfl_down_sync(0xFFFFFFFF, partial_result, shuffle_distance));
```
jlebar: CUDA_WARP_ALL?

tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc
```diff
@@ -1015,6 +1016,21 @@ __global__ void __launch_bounds__(640, 2)
   }
 }
 
+// Device function to compute sub-warp reductions for a power-of-two group of
```
jlebar: "reduction" is a generic term -- what is being done here is a *sum* reduction.  I think we should change the comment, but also the name of the function.

tensorflow/core/util/cuda_kernel_helper.h
```diff
@@ -25,6 +25,28 @@ limitations under the License.
 #include "tensorflow/core/platform/logging.h"
 #include "tensorflow/core/platform/stream_executor.h"
 #include "tensorflow/core/platform/types.h"
+#include "cuda/include/cuda.h"
+
+#define CUDA_WARP_ALL 0xFFFFFFFF
```
jlebar: Please add a comment.

jlebar: Please add a comment.

tensorflow/compiler/xla/service/gpu/ir_emitter_unnested.cc
nluehr: Switched to CUDA_WARP_ALL. But if this is ever un-commented, a dependency on cuda_kernel_helper.h will need to be added.

jlebar:
What tools did you use to figure out what needed to be changed and to verify that the code now works?

jlebar (approved):

jlebar:
Approval, but I would really like to know what tools were used and so on so we can (ideally) add tests that this continues to be clean wrt whatever sanitizers you used

Artem-B (commented):
Issue with __*_sync macros.

pieroit:
So being this PR merged into master, should we compile by hand to use it?
How much time before it goes into the version distributed by `pip`?

Thank you :)

byronyi:
I think tf-nightly is specifically for the impatient.

pieroit:
thank you @byronyi, unfortunately tf-nightly is not yet (but soon will) suport GPUs. From the readme:

```We are pleased to announce that TensorFlow now offers nightly pip packages under the tf-nightly project on pypi. Simply run pip install tf-nightly in a clean environment to install the nightly tensorflow build. We currently only support CPU packages on Linux, Mac, and Windows. GPU packages on all platforms will arrive soon!```

lefnire:
FWIW I'm seeing a tf-gpu-nightly link in https://github.com/tensorflow/tensorflow/commit/4270c4a00bb90d97418df5b0e9b6b3e148a72a1e, in case they just haven't updated the README? Worth a shot trying to install it with pip, just in case

pieroit:
@lefnire cannot find it on pip. I guess it's a matter of days ;)

tfboyd:
It will not be in the nightly as far as linking to CUDA 9.  TF 1.4 is next, which will **most likely** stay CUDA 8 + cuDNN 6 and then 1.5 would move to cuDNN 7 and CUDA 9 if all goes well.  cuDNN 7 + Cuda 9 gives you Volta support along with other internal changes like batch_norm supporting FP16 optimally.  There may be other benefits but Volta is the big deal.

pieroit:
I'm gonna roll back to cuDNN 6 + CUDA 8 and wait some time. GPU computing is quite a mess :D 
Thank you guys

PR #12503, opened 2017-08-22
Add cudnn7 support

nluehr:
Attention @zheng-xq 

- Account for cudnnSetRNNDescriptor API change
- Add support for CUDNN_TENSOR_OP_MATH in cudnn v7
  - Applies to forward and backward convolutions that have fp16
  input/output. Computations will fall back to pseudo-fp16
  if tensor op math is disabled or not supported.
  - Enabled by default, but can be disabled using the environment
  variable TF_ENABLE_TENSOR_OP_MATH=0.
  - The choice of whether to use tensor op math is included in the
  autotuning of convolutions.

### DIFF
--- a/tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
+++ b/tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
@@ -33,6 +33,7 @@ using se::dnn::ConvolutionDescriptor;
 using se::dnn::DataLayout;
 using se::dnn::FilterDescriptor;
 using se::dnn::FilterLayout;
+using se::dnn::AlgorithmDesc;
 
 ConvolveScratchAllocator::ConvolveScratchAllocator(
     int device_ordinal, DeviceMemoryAllocator* memory_allocator)
@@ -251,12 +252,13 @@ tensorflow::Status ConvolutionThunk::Convolve(
       "Unable to launch convolution for thunk %p with type %s and algorithm "
       "(%lld, %lld)",
       this, ConvolutionKindToString(convolution_kind_).c_str(),
-      algorithm_config.algorithm(), algorithm_config.algorithm_no_scratch());
+      algorithm_config.algorithm().algo_id(),
+      algorithm_config.algorithm_no_scratch().algo_id());
 }
 
-std::vector<se::dnn::AlgorithmType> ConvolutionThunk::GetAlgorithms(
+std::vector<AlgorithmDesc::Index> ConvolutionThunk::GetAlgorithms(
     se::StreamExecutor* stream_exec) const {
-  std::vector<se::dnn::AlgorithmType> algorithms;
+  std::vector<AlgorithmDesc::Index> algorithms;
   // TODO(yangzihao): Currently disable the use of winograd nonfused in XLA
   // by default. Should send in conv parameters and enable it when
   // ShouldIncludeWinogradNonfusedAlgo() returns true.
@@ -286,7 +288,7 @@ tensorflow::Status ConvolutionThunk::ConvolveWithTune(
     const ConvolutionDescriptor& convolution_descriptor,
     const BufferAllocations& buffer_allocations, se::Stream* stream) {
   // TODO(b/29126320): Try cudnn v5's new auto-tuner when it's rolled out.
-  if (best_algorithm_.algorithm() == se::dnn::kDefaultAlgorithm) {
+  if (best_algorithm_.algorithm().is_default()) {
     // Auto-tuning either is disabled or only happens in the first run of this
     // function.
     VLOG(2) << "Profiling for best convolution algorithm used for "
@@ -295,26 +297,32 @@ tensorflow::Status ConvolutionThunk::ConvolveWithTune(
 
     se::dnn::ProfileResult best_result;
     se::dnn::ProfileResult best_result_without_scratch;
-    for (se::dnn::AlgorithmType algorithm : GetAlgorithms(stream->parent())) {
-      ConvolveScratchAllocator scratch_allocator(
-          buffer_allocations.device_ordinal(),
-          buffer_allocations.memory_allocator());
-      se::dnn::ProfileResult profile_result;
-      bool launch_ok =
-          Convolve(input_descriptor, input_data, filter_descriptor, filter_data,
-                   output_descriptor, output_data, convolution_descriptor,
-                   se::dnn::AlgorithmConfig(algorithm, algorithm), stream,
-                   &scratch_allocator, &profile_result)
-              .ok();
-      if (launch_ok && profile_result.is_valid()) {
-        if (profile_result.elapsed_time_in_ms() <
-            best_result.elapsed_time_in_ms()) {
-          best_result = profile_result;
-        }
-        if (scratch_allocator.TotalAllocatedBytes() == 0 &&
-            profile_result.elapsed_time_in_ms() <
-                best_result_without_scratch.elapsed_time_in_ms()) {
-          best_result_without_scratch = profile_result;
+    std::vector<AlgorithmDesc::Index> algorithms =
+        GetAlgorithms(stream->parent());
+    for (bool use_tensor_ops : {false, true}) {
+      for (auto algo_index : algorithms) {
+        AlgorithmDesc algorithm(algo_index, use_tensor_ops);
+        ConvolveScratchAllocator scratch_allocator(
+            buffer_allocations.device_ordinal(),
+            buffer_allocations.memory_allocator());
+        se::dnn::ProfileResult profile_result;
+        bool launch_ok =
+            Convolve(input_descriptor, input_data, filter_descriptor,
+                     filter_data, output_descriptor, output_data,
+                     convolution_descriptor,
+                     se::dnn::AlgorithmConfig(algorithm, algorithm), stream,
+                     &scratch_allocator, &profile_result)
+                .ok();
+        if (launch_ok && profile_result.is_valid()) {
+          if (profile_result.elapsed_time_in_ms() <
+              best_result.elapsed_time_in_ms()) {
+            best_result = profile_result;
+          }
+          if (scratch_allocator.TotalAllocatedBytes() == 0 &&
+              profile_result.elapsed_time_in_ms() <
+                  best_result_without_scratch.elapsed_time_in_ms()) {
+            best_result_without_scratch = profile_result;
+          }
         }
       }
     }
@@ -324,7 +332,7 @@ tensorflow::Status ConvolutionThunk::ConvolveWithTune(
     } else {
       LOG(ERROR) << "No convolution algorithm works with profiling. Fall back "
                     "to the default algorithm.";
-      best_algorithm_.set_algorithm(se::dnn::kDefaultAlgorithm);
+      best_algorithm_.set_algorithm(AlgorithmDesc());
     }
 
     if (best_result_without_scratch.is_valid()) {
@@ -334,13 +342,14 @@ tensorflow::Status ConvolutionThunk::ConvolveWithTune(
       LOG(ERROR) << "No convolution algorithm without scratch works with "
                     "profiling. Fall back "
                     "to the default algorithm.";
-      best_algorithm_.set_algorithm_no_scratch(se::dnn::kDefaultAlgorithm);
+      best_algorithm_.set_algorithm_no_scratch(AlgorithmDesc());
     }
   }
 
   {
-    VLOG(2) << "Using convolution algorithm (" << best_algorithm_.algorithm()
-            << ", " << best_algorithm_.algorithm_no_scratch()
+    VLOG(2) << "Using convolution algorithm ("
+            << best_algorithm_.algorithm().algo_id() << ", "
+            << best_algorithm_.algorithm_no_scratch().algo_id()
             << ") for ConvolutionThunk: " << this;
     ConvolveScratchAllocator scratch_allocator(
         buffer_allocations.device_ordinal(),
--- a/tensorflow/compiler/xla/service/gpu/convolution_thunk.h
+++ b/tensorflow/compiler/xla/service/gpu/convolution_thunk.h
@@ -115,7 +115,9 @@ class ConvolutionThunk : public Thunk {
       perftools::gputools::dnn::ProfileResult* profile_result);
 
   // Returns the convolve algorithms that can be used for this ConvolutionThunk.
-  std::vector<perftools::gputools::dnn::AlgorithmType> GetAlgorithms(
+  // TODO(nluehr) GetAlgorithms should return AlgorithmDesc including both
+  // tensor-op and non-tensor-op variants.
+  std::vector<perftools::gputools::dnn::AlgorithmDesc::Index> GetAlgorithms(
       perftools::gputools::StreamExecutor* stream_exec) const;
 
   // Fastest cuDNN convolution algorithm for this thunk learned from
--- a/tensorflow/contrib/fused_conv/kernels/fused_conv2d_bias_activation_op.cc
+++ b/tensorflow/contrib/fused_conv/kernels/fused_conv2d_bias_activation_op.cc
@@ -493,37 +493,42 @@ void LaunchFusedConv2DBiasActivationOp<GPUDevice, T, BiasType, ScaleType>::
   dnn::AlgorithmConfig algorithm_config;
   if (cudnn_use_autotune && !AutoTuneConvBiasActivation::GetInstance()->Find(
                                 fused_conv_parameters, &algorithm_config)) {
-    std::vector<dnn::AlgorithmType> algorithms;
+    std::vector<dnn::AlgorithmDesc::Index> algorithms;
     CHECK(stream->parent()->GetConvolveAlgorithms(
         fused_conv_parameters.ShouldIncludeWinogradNonfusedAlgo<T>(),
         &algorithms));
     dnn::ProfileResult best_result;
     dnn::ProfileResult best_result_no_scratch;
-    for (auto profile_algorithm : algorithms) {
-      // TODO(zhengxq): profile each algorithm multiple times to better
-      // accuracy.
-      CudnnScratchAllocator scratch_allocator(ConvolveScratchSize, ctx);
-      dnn::ProfileResult profile_result;
-      bool cudnn_launch_status =
-          stream
-              ->ThenFusedConvolveWithAlgorithm(
-                  conv_input_desc, conv_input_ptr, conv_input_scale,
-                  filter_desc, filter_ptr, conv_desc, side_input_ptr,
-                  side_input_scale, bias_desc, bias_ptr,
-                  dnn::ActivationMode::kRelu, output_desc, &output_ptr,
-                  &scratch_allocator, dnn::AlgorithmConfig(profile_algorithm),
-                  &profile_result)
-              .ok();
-      if (cudnn_launch_status) {
-        if (profile_result.is_valid()) {
-          if (profile_result.elapsed_time_in_ms() <
-              best_result.elapsed_time_in_ms()) {
-            best_result = profile_result;
-          }
-          if (scratch_allocator.TotalByteSize() == 0 &&
-              profile_result.elapsed_time_in_ms() <
-                  best_result_no_scratch.elapsed_time_in_ms()) {
-            best_result_no_scratch = profile_result;
+    // TODO(benbarsdell): Ideally this should not attempt using tensor op math
+    // if it's not enabled.
+    for (bool use_tensor_ops : {false, true}) {
+      for (auto algo_index : algorithms) {
+        // TODO(zhengxq): profile each algorithm multiple times to better
+        // accuracy.
+        dnn::AlgorithmDesc profile_algorithm(algo_index, use_tensor_ops);
+        CudnnScratchAllocator scratch_allocator(ConvolveScratchSize, ctx);
+        dnn::ProfileResult profile_result;
+        bool cudnn_launch_status =
+            stream
+                ->ThenFusedConvolveWithAlgorithm(
+                    conv_input_desc, conv_input_ptr, conv_input_scale,
+                    filter_desc, filter_ptr, conv_desc, side_input_ptr,
+                    side_input_scale, bias_desc, bias_ptr,
+                    dnn::ActivationMode::kRelu, output_desc, &output_ptr,
+                    &scratch_allocator, dnn::AlgorithmConfig(profile_algorithm),
+                    &profile_result)
+                .ok();
+        if (cudnn_launch_status) {
+          if (profile_result.is_valid()) {
+            if (profile_result.elapsed_time_in_ms() <
+                best_result.elapsed_time_in_ms()) {
+              best_result = profile_result;
+            }
+            if (scratch_allocator.TotalByteSize() == 0 &&
+                profile_result.elapsed_time_in_ms() <
+                    best_result_no_scratch.elapsed_time_in_ms()) {
+              best_result_no_scratch = profile_result;
+            }
           }
         }
       }
--- a/tensorflow/core/kernels/conv_grad_input_ops.cc
+++ b/tensorflow/core/kernels/conv_grad_input_ops.cc
@@ -630,7 +630,7 @@ void LaunchConv2DBackpropInputOp<GPUDevice, T>::operator()(
     int col_stride, const Padding& padding, Tensor* in_backprop,
     TensorFormat data_format) {
   using perftools::gputools::dnn::AlgorithmConfig;
-  using perftools::gputools::dnn::AlgorithmType;
+  using perftools::gputools::dnn::AlgorithmDesc;
   using perftools::gputools::dnn::ProfileResult;
 
   std::vector<int32> strides(4, 1);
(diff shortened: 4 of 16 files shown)


jhseu:

tfboyd:
Asking team about CLA for associate corporation.   @zheng-xq  Please review or assign others before CLA so we can merge as soon as we figure that out.

tensorflow/core/kernels/conv_grad_input_ops.cc
```diff
@@ -892,10 +897,13 @@ class Conv2DSlowBackpropInputOp : public OpKernel {
                   errors::NotFound("No algorithm worked!"));
       if (best_result.is_valid()) {
         algorithm_config.set_algorithm(best_result.algorithm());
+        algorithm_config.set_use_tensor_op_math(best_result.use_tensor_op_math());
```
zheng-xq: Could you add a TODO to merge AlgorithmConfig into PerfResult? We shouldn't have to copy individual field here.

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -541,6 +553,29 @@ class ScopedFilterDescriptor {
...
+class TensorOpMath {
+ public:
+  static bool IsEnabled() {
+    static bool is_enabled = IsEnabledImpl();
+    return is_enabled;
+  }
+
+ private:
+  static bool IsEnabledImpl() {
+    const char* tf_env_var_val = getenv("TF_ENABLE_TENSOR_OP_MATH");
```
zheng-xq: Add a TODO to refactor read from env-var into its own helper function, similar to what TF does. Or reuse the one from TF.

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -583,6 +618,27 @@ class ScopedConvolutionDescriptor {
...
+    this->set_use_tensor_op_math(true);
+  }
+
+  void set_use_tensor_op_math(bool use_tensor_op_math) {
+#if CUDNN_VERSION >= 7000
+    cudnnMathType_t math_type =
+        (use_tensor_op_math ?
+         CUDNN_TENSOR_OP_MATH :
+         CUDNN_DEFAULT_MATH);
+    if (TensorOpMath<true>::IsEnabled()) {
```
zheng-xq: Out of curiosity, is there a legitimate reason why people don't want to use tensor_op when it is available? Since we have autotuning, it will never get slower. 

So if the problem is functional, we might need to consider whether we should make the control op/subgraph level or globally?

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -1010,11 +1066,21 @@ class CudnnRnnDescriptor : public CudnnDescriptorCommon<dnn::RnnDescriptor> {
     // Create the RNN handle
     cudnnStatus_t status = wrap::cudnnCreateRNNDescriptor(parent_, &rnn_desc_);
     CUDNN_RETURN_IF_FAIL(status, "Unable to create RNN descriptor");
+    #if CUDNN_VERSION >= 6000
+    cudnnRNNAlgo_t rnn_algo = CUDNN_RNN_ALGO_STANDARD;
```
zheng-xq: Add a TODO to autotune this

tensorflow/stream_executor/dnn.h
```diff
@@ -723,7 +741,9 @@ class AlgorithmConfig {
 
  private:
   AlgorithmType algorithm_;
+  bool          use_tensor_op_math_;
```
zheng-xq: We should consolite AlgorithmType and bool into a struct that represents the real algorithm to use. Either modify AlgorithmType, or add a new type. Either way is fine with me.

tensorflow/stream_executor/dnn.h
```diff
@@ -701,20 +704,35 @@ class AlgorithmConfig {
...
-  AlgorithmConfig(AlgorithmType algorithm, AlgorithmType algorithm_no_scratch)
-      : algorithm_(algorithm), algorithm_no_scratch_(algorithm_no_scratch) {}
+        use_tensor_op_math_(false),
+        algorithm_no_scratch_(kDefaultAlgorithm),
+        use_tensor_op_math_no_scratch_(false) {}
+  explicit AlgorithmConfig(AlgorithmType algorithm, bool use_tensor_op_math=false)
+      : algorithm_(algorithm), use_tensor_op_math_(use_tensor_op_math),
+        algorithm_no_scratch_(kDefaultAlgorithm), use_tensor_op_math_no_scratch_(false) {}
+  AlgorithmConfig(AlgorithmType algorithm, AlgorithmType algorithm_no_scratch,
+                 bool use_tensor_op_math=false, bool use_tensor_op_math_no_scratch=false)
```
zheng-xq: nit: even without merge the two types into a single type, I think the order "algorithm, bool, algorithm, bool" works better.

Also I'm not a big fan of default arguments in C++. Overloaded functions tend to work better. But it's up to you.

zheng-xq (changes requested):

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: We expect tensor ops to be sufficient wherever FP16 inputs are numerically feasible. The environment variable just provides a sanity check to quickly rule out tensor ops as a culprit when issues are encountered (e.g., during some conversion from fp32 to fp16). In this context, a global option is sufficient. Once the numerics of fp16 training are well understood, the opt-out can likely be removed entirely.

tensorflow/core/kernels/conv_grad_input_ops.cc
nluehr: The copy of separate parameters is avoided by merging the bool field into AlgorithmType.

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: Refactored.

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: Done

tensorflow/stream_executor/dnn.h
nluehr: Types merged.

tensorflow/stream_executor/dnn.h
nluehr: Expanded AlgorithmType into a struct.

nluehr:
I signed it!

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,33 @@ class PoolingDescriptor {
   std::vector<int64> strides_;
 };
 
-typedef int64 AlgorithmType;
-constexpr AlgorithmType kDefaultAlgorithm = -1;
+class AlgorithmType {
+  private:
```
zheng-xq: The convention is that the data fields are at the end.

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,33 @@ class PoolingDescriptor {
...
+  AlgorithmType(int64 a) : algo_(a), tensor_ops_enabled_(false) { }
+  bool is_default() const { return algo_ == kDefaultAlgorithm; }
+  bool operator==(const AlgorithmType& other) const {
+    return this->algo_ == other.algo_ &&
+           this->tensor_ops_enabled_ == other.tensor_ops_enabled_;
+  }
+  bool tensor_ops_enabled() const {
+    return tensor_ops_enabled_;
+  }
+  void set_tensor_ops(bool b) {
```
zheng-xq: Why do we have this function instead of a constructor that takes both arguments?

zheng-xq (commented):
Only minor nits. Please let me know if you want to fix it in this CL, or with a follow-up.

tensorflow/stream_executor/dnn.h
nluehr: Replaced setters with constructor.

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,31 @@ class PoolingDescriptor {
   std::vector<int64> strides_;
 };
 
-typedef int64 AlgorithmType;
-constexpr AlgorithmType kDefaultAlgorithm = -1;
+class AlgorithmType {
```
jlebar: Please add a comment explaining what this is.

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,31 @@ class PoolingDescriptor {
...
+  }
+  Index get_algo_id() const {
+    return algo_;
+  }
+
+  private:
+  Index algo_;
+  bool tensor_ops_enabled_;
+  enum { kDefaultAlgorithm = -1 };
+
```
jlebar: Please remove blank line.

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,31 @@ class PoolingDescriptor {
   std::vector<int64> strides_;
 };
 
-typedef int64 AlgorithmType;
-constexpr AlgorithmType kDefaultAlgorithm = -1;
+class AlgorithmType {
+  public:
```
jlebar: "public:" and "private:" should be indented one space.

Recommend running clang-format over this whole patch, it will make other changes as well.

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,31 @@ class PoolingDescriptor {
...
+      : algo_(a), tensor_ops_enabled_(use_tensor_ops) {}
+  bool is_default() const { return algo_ == kDefaultAlgorithm; }
+  bool operator==(const AlgorithmType& other) const {
+    return this->algo_ == other.algo_ &&
+           this->tensor_ops_enabled_ == other.tensor_ops_enabled_;
+  }
+  bool tensor_ops_enabled() const {
+    return tensor_ops_enabled_;
+  }
+  Index get_algo_id() const {
```
jlebar: Suggest simply algo_id() -- that would match the other functions.

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,31 @@ class PoolingDescriptor {
...
-constexpr AlgorithmType kDefaultAlgorithm = -1;
+class AlgorithmType {
+  public:
+  typedef int64 Index;
+  AlgorithmType()
+      : algo_(kDefaultAlgorithm), tensor_ops_enabled_(false) {}
+  AlgorithmType(Index a, bool use_tensor_ops)
+      : algo_(a), tensor_ops_enabled_(use_tensor_ops) {}
+  bool is_default() const { return algo_ == kDefaultAlgorithm; }
+  bool operator==(const AlgorithmType& other) const {
```
jlebar: Suggest reordering member functions so that operator== doesn't interrupt the getters.

tensorflow/stream_executor/dnn.h
```diff
@@ -667,8 +667,31 @@ class PoolingDescriptor {
...
+    return tensor_ops_enabled_;
+  }
+  Index get_algo_id() const {
+    return algo_;
+  }
+
+  private:
+  Index algo_;
+  bool tensor_ops_enabled_;
+  enum { kDefaultAlgorithm = -1 };
```
jlebar: Please put this first, since it's a constant.

tensorflow/stream_executor/dnn.h
jlebar: In particular, it seems that "type" here does not refer to "C++ type", but instead "type of algorithm".  That was confusing to me; probably worth calling that out explicitly in a comment.

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -1010,11 +1064,22 @@ class CudnnRnnDescriptor : public CudnnDescriptorCommon<dnn::RnnDescriptor> {
     // Create the RNN handle
     cudnnStatus_t status = wrap::cudnnCreateRNNDescriptor(parent_, &rnn_desc_);
     CUDNN_RETURN_IF_FAIL(status, "Unable to create RNN descriptor");
+    #if CUDNN_VERSION >= 6000
+    // TODO: autotune algo selection.
```
jlebar: I don't think StreamExecutor is the right place to do auto-tuning; it's too low-level.  XLA and TF both implement auto-tuning atop SE for gemm, and I think we should do the same thing for convs.

Perhaps a better TODO for here is something like "allow the user to choose an algorithm".

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -541,6 +556,24 @@ class ScopedFilterDescriptor {
...
+  static bool IsEnabled() {
+    static bool is_enabled = IsEnabledImpl();
+    return is_enabled;
+  }
+
+ private:
+  static bool IsEnabledImpl() {
+    using tensorflow::ReadBoolFromEnvVar;
+    bool ret;
+    ReadBoolFromEnvVar("TF_ENABLE_TENSOR_OP_MATH", DefaultFlag, &ret);
```
jlebar: Since this flag is an opt-out, maybe it would make more sense to call it TF_DISABLE_TENSOR_OP_MATH?  If I had to guess as a user, I'd expect that TF_ENABLE_TENSOR_OP_MATH=0 would be the default, but it's not...

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -541,6 +556,24 @@ class ScopedFilterDescriptor {
   SE_DISALLOW_COPY_AND_ASSIGN(ScopedFilterDescriptor);
 };
 
+// A helper class to decide whether to enable the TENSOR_OP_MATH math type
+template <bool DefaultFlag>
```
jlebar: This could be simply a function instead of a templated class:

```
  static bool TensorOpMathEnabled() {
    static bool is_enabled = [] {
      bool ret;
      tensorflow::ReadBoolFromEnvVar("TF_ENABLE_TENSOR_OP_MATH", /*default=*/true, &ret);
      return ret;
    }();
    return is_enabled;
  }
```

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -541,6 +556,24 @@ class ScopedFilterDescriptor {
...
+class TensorOpMath {
+ public:
+  static bool IsEnabled() {
+    static bool is_enabled = IsEnabledImpl();
+    return is_enabled;
+  }
+
+ private:
+  static bool IsEnabledImpl() {
+    using tensorflow::ReadBoolFromEnvVar;
```
jlebar: This "using" statement doesn't really save us anything.  Recommend not doing it this way, especially because this looks like an ADL idiom, https://stackoverflow.com/questions/28130671/how-does-using-stdswap-enable-adl.

tensorflow/core/kernels/conv_ops_3d.cc
```diff
@@ -385,39 +385,43 @@ struct LaunchConvOp<GPUDevice, T> {
...
-            if (profile_result.elapsed_time_in_ms() <
-                best_result.elapsed_time_in_ms()) {
-              best_result = profile_result;
-            }
-            if (scratch_allocator.TotalByteSize() == 0 &&
-                profile_result.elapsed_time_in_ms() <
-                    best_result_no_scratch.elapsed_time_in_ms()) {
-              best_result_no_scratch = profile_result;
+      // TODO(benbarsdell): Ideally this should not attempt using tensor op math
+      //                      if it's not enabled.
```
jlebar: We don't indent comments this way; please just flow this second line all the way to the LHS.

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
```diff
@@ -295,26 +297,32 @@ tensorflow::Status ConvolutionThunk::ConvolveWithTune(
...
-            best_result.elapsed_time_in_ms()) {
-          best_result = profile_result;
-        }
-        if (scratch_allocator.TotalAllocatedBytes() == 0 &&
-            profile_result.elapsed_time_in_ms() <
-                best_result_without_scratch.elapsed_time_in_ms()) {
-          best_result_without_scratch = profile_result;
+    std::vector<se::dnn::AlgorithmType::Index> algorithms
+        = GetAlgorithms(stream->parent());
+    for (int use_tensor_ops = 0; use_tensor_ops < 2; ++use_tensor_ops) {
```
jlebar: Much prefer to write this loop as

  for (bool use_tensor_ops : {false, true}) { ... }

But even better would be for GetAlgorithms to return a vector<se::dnn::AlgorithmType> -- then you wouldn't need to do this at all.

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
```diff
@@ -295,26 +297,32 @@ tensorflow::Status ConvolutionThunk::ConvolveWithTune(
...
-        if (profile_result.elapsed_time_in_ms() <
-            best_result.elapsed_time_in_ms()) {
-          best_result = profile_result;
-        }
-        if (scratch_allocator.TotalAllocatedBytes() == 0 &&
-            profile_result.elapsed_time_in_ms() <
-                best_result_without_scratch.elapsed_time_in_ms()) {
-          best_result_without_scratch = profile_result;
+    std::vector<se::dnn::AlgorithmType::Index> algorithms
+        = GetAlgorithms(stream->parent());
```
jlebar: Don't ask how I know this, but `=` should go on the line above.  :)

clang-format is our friend.

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
```diff
@@ -33,6 +33,7 @@ using se::dnn::ConvolutionDescriptor;
 using se::dnn::DataLayout;
 using se::dnn::FilterDescriptor;
 using se::dnn::FilterLayout;
+using se::dnn::AlgorithmType;
```
jlebar: We added this "using" statement, but then we fully-qualify the type below.  We should do only one or the other.

tensorflow/contrib/fused_conv/kernels/fused_conv2d_bias_activation_op.cc
```diff
@@ -412,34 +411,39 @@ void LaunchFusedConv2DBiasActivationOp<GPUDevice, T>::launch(
...
-          if (profile_result.elapsed_time_in_ms() <
-              best_result.elapsed_time_in_ms()) {
-            best_result = profile_result;
-          }
-          if (scratch_allocator.TotalByteSize() == 0 &&
-              profile_result.elapsed_time_in_ms() <
-                  best_result_no_scratch.elapsed_time_in_ms()) {
-            best_result_no_scratch = profile_result;
+    // TODO(benbarsdell): Ideally this should not attempt using tensor op math
+    //                      if it's not enabled.
```
jlebar: ...having GetConvolveAlgorithms return a vector<AlgorithmType> instead of a vector<AlgorithmType::Index> would also resolve this TODO, I think?

jlebar (dismissed):
Thank you for doing this -- we are excited to have these changes.

tensorflow/stream_executor/dnn.h
nluehr: Renamed dnn::AlgorithmType to dnn::AlgorithmDesc and added explanatory comment.

tensorflow/stream_executor/dnn.h
nluehr: Thanks. Fixed all clang-format issues.

tensorflow/stream_executor/dnn.h
nluehr: Done

tensorflow/stream_executor/dnn.h
nluehr: Done

tensorflow/stream_executor/dnn.h
nluehr: Done

tensorflow/stream_executor/dnn.h
nluehr: Done

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: Done

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: Fair enough.

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: Done

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: OK.

tensorflow/core/kernels/conv_ops_3d.cc
nluehr: OK

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
nluehr: Fixed - now using unqualified name elsewhere in file.

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
nluehr: fixed

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
nluehr: Improved loop syntax.

tensorflow/contrib/fused_conv/kernels/fused_conv2d_bias_activation_op.cc
nluehr: That may be the best option. But the TODO would still move inside the GetConvolveAlgorithms function. The work required is figuring out how to determine where tensor ops are applicable without brute forcing the search.

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
jlebar: Did you look into making GetAlgorithms return a `vector<se::dnn::AlgorithmDesc>`?  I think that would greatly simplify things...

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
```diff
@@ -295,26 +297,32 @@ tensorflow::Status ConvolutionThunk::ConvolveWithTune(
...
-            best_result.elapsed_time_in_ms()) {
-          best_result = profile_result;
-        }
-        if (scratch_allocator.TotalAllocatedBytes() == 0 &&
-            profile_result.elapsed_time_in_ms() <
-                best_result_without_scratch.elapsed_time_in_ms()) {
-          best_result_without_scratch = profile_result;
+    std::vector<AlgorithmDesc::Index> algorithms =
+        GetAlgorithms(stream->parent());
+    for (bool use_tensor_ops : {false, true}) {
```
zheng-xq: This is quite a bit slower, if I remember correct. I admit it looks nicer.

Please confirm if you don't see a performance problem with 8-GPUs. Then we can keep it in.

tensorflow/compiler/xla/service/gpu/convolution_thunk.h
```diff
@@ -115,7 +115,7 @@ class ConvolutionThunk : public Thunk {
       perftools::gputools::dnn::ProfileResult* profile_result);
 
   // Returns the convolve algorithms that can be used for this ConvolutionThunk.
-  std::vector<perftools::gputools::dnn::AlgorithmType> GetAlgorithms(
+  std::vector<perftools::gputools::dnn::AlgorithmDesc::Index> GetAlgorithms(
```
zheng-xq: Add a TODO to return AlgorithmDesc. The GetAlgorithms itself can go through the loop.

zheng-xq (approved):

zheng-xq:
In general, I'm fine with this change. We can keep improving it afterwards. 

Nathan, please confirm that the performance still stays the same with Cuda9+Cudnn7 on Volta. And nothing changes for Cuda8+Cudnn6 on P100. I'll grant my LGTM afterwards. Thanks!

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
jlebar: It generates the same code: https://godbolt.org/g/9mZHTL.

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
nluehr: This would require fixing the forward, backward filter and backward data GetAlgorithm variants. There is a fair bit of cleanup needed around this section of code in general. I think it would be better to tackle it as a separate issue.

tensorflow/compiler/xla/service/gpu/convolution_thunk.h
nluehr: Added TODO. Hopefully this completes the PR.

drpngx:
@nluehr Can you confirm that the performance is the same on Volta, and that it also works on cuda8+cudnn6 on P100?

Also, did you address @jlebar comments?

tensorflow/stream_executor/dnn.h
nluehr: Done.

tensorflow/compiler/xla/service/gpu/convolution_thunk.cc
nluehr: Added TODO to track this.

nluehr:
I believe all comments from @jlebar have been addressed.
Code still builds on cuda 8.0 / cudnn 6.
Perf remains unchanged: Resnet 50 (batch size = 64) on a P100 gets 224.5 img/sec before and 224.7 after.

jlebar:
> I believe all comments from @jlebar have been addressed.

Sorry to be a broken record, but to be totally clear, the GetAlgorithms issue (i.e. requiring a loop over {true, false} every time we call the function) is still outstanding.  I believe XQ is OK waiting for you to fix that in a later PR because of the time-criticality of this change, and that's fine with me.

I don't immediately see any clang-format issues, thank you for fixing those.  (Or, if you didn't, blame it on my eyes not being as sharp this time of day. :)

tensorflow/compiler/xla/service/gpu/convolution_thunk.h
```diff
@@ -115,7 +115,9 @@ class ConvolutionThunk : public Thunk {
       perftools::gputools::dnn::ProfileResult* profile_result);
 
   // Returns the convolve algorithms that can be used for this ConvolutionThunk.
-  std::vector<perftools::gputools::dnn::AlgorithmType> GetAlgorithms(
+  // TODO(nluehr) GetAlgorithms should return AlgorithmDesc including both
```
nluehr: @jlebar. Sorry for not noting more clearly. In a future PR I will re-implement the loop over tensor ops inside GetAlgorithms, as noted in by this TODO.

drpngx:
Jenkins, test this please.

drpngx:
Looking forward to follow up PR ;-)

drpngx:
Windows build reported as failure but I don't see an obvious place: https://ci.tensorflow.org/job/tensorflow-pr-win-cmake-py/4420/consoleFull

Jenkins, test this please.

PR #12504, opened 2017-08-22
NaN propagation for GPU pooling ops

nluehr:
Attention @zheng-xq

- Changes the custom fwd maxpooling kernel to propagate NaNs. This
  makes it match the behavior of CUDNN, and ensures that CUDNN's
  bwd maxpooling kernel behaves as expected (propagating NaNs).
- Previous behavior can be restored with environment variable
  TF_ENABLE_MAXPOOL_NANPROP=0.
- Changes the GPU bwd maxpool op tests to expect propagated NaNs
  (matching the behavior of the CPU path).

### DIFF
--- a/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
+++ b/tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
@@ -29,6 +29,15 @@ limitations under the License.
 
 namespace tensorflow {
 namespace {
+template <bool propagate_nans, typename dtype>
+EIGEN_DEVICE_FUNC EIGEN_ALWAYS_INLINE bool IsGreaterThan(dtype a, dtype b) {
+  if (propagate_nans) {
+    return !(a <= b);
+  } else {
+    return a > b;
+  }
+}
+
 // This is Yangqing's custom kernel for the maxpooling operation. There are
 // three functions: MaxPoolForwardNCHW and MaxPoolForwardNHWC are the two
 // forward functions, dealing with the forward case. MaxPoolBackward is the
@@ -51,7 +60,7 @@ namespace {
 // const int output_size = batch * channels * pooled_height * pooled_width;
 // MaxPoolForwardNCHW<<<(output_size + kThreadsPerBlock - 1) / kThreadsPerBlock,
 //                      kThreadsPerBlock, 0, cuda_stream>>>(...);
-template <typename dtype>
+template <bool propagate_nans, typename dtype>
 __global__ void MaxPoolForwardNCHW(const int nthreads, const dtype* bottom_data,
                                    const int channels, const int height,
                                    const int width, const int pooled_height,
@@ -77,7 +86,7 @@ __global__ void MaxPoolForwardNCHW(const int nthreads, const dtype* bottom_data,
     for (int h = hstart; h < hend; ++h) {
       for (int w = wstart; w < wend; ++w) {
         int idx = c * height * width + h * width + w;
-        if (bottom_data_n[idx] > maxval) {
+        if (IsGreaterThan<propagate_nans>(bottom_data_n[idx], maxval)) {
           maxidx = idx;
           maxval = bottom_data_n[idx];
         }
@@ -126,7 +135,7 @@ __global__ void MaxPoolForwardNoMaskKernel_NCHW_VECT_C(
   }
 }
 
-template <typename dtype>
+template <bool propagate_nans, typename dtype>
 __global__ void MaxPoolForwardNHWC(const int nthreads, const dtype* bottom_data,
                                    const int height, const int width,
                                    const int channels, const int pooled_height,
@@ -153,7 +162,7 @@ __global__ void MaxPoolForwardNHWC(const int nthreads, const dtype* bottom_data,
     for (int h = hstart; h < hend; ++h) {
       for (int w = wstart; w < wend; ++w) {
         int idx = (h * width + w) * channels + c;
-        if (bottom_data_n[idx] > maxval) {
+        if (IsGreaterThan<propagate_nans>(bottom_data_n[idx], maxval)) {
           maxidx = idx;
           maxval = bottom_data_n[idx];
         }
@@ -390,15 +399,24 @@ bool MaxPoolForwardWithOptionalArgmax<T>::operator()(
     const int channels, const int pooled_height, const int pooled_width,
     const int kernel_h, const int kernel_w, const int stride_h,
     const int stride_w, const int pad_t, const int pad_l, T* top_data,
-    int64* mask, const Eigen::GpuDevice& d) {
+    int64* mask, const Eigen::GpuDevice& d, bool propagate_nans) {
   const int kThreadsPerBlock = 1024;
   const int output_size = batch * channels * pooled_height * pooled_width;
-
-  MaxPoolForwardNHWC<<<(output_size + kThreadsPerBlock - 1) / kThreadsPerBlock,
-                       kThreadsPerBlock, 0, d.stream()>>>(
-      output_size, bottom_data, height, width, channels, pooled_height,
-      pooled_width, kernel_h, kernel_w, stride_h, stride_w, pad_t, pad_l,
-      top_data, mask);
+  if (propagate_nans) {
+    MaxPoolForwardNHWC<true>
+        <<<(output_size + kThreadsPerBlock - 1) / kThreadsPerBlock,
+           kThreadsPerBlock, 0, d.stream()>>>
+        (output_size, bottom_data, height, width, channels, pooled_height,
+         pooled_width, kernel_h, kernel_w, stride_h, stride_w, pad_t, pad_l,
+         top_data, mask);
+  } else {
+    MaxPoolForwardNHWC<false>
+        <<<(output_size + kThreadsPerBlock - 1) / kThreadsPerBlock,
+           kThreadsPerBlock, 0, d.stream()>>>
+        (output_size, bottom_data, height, width, channels, pooled_height,
+         pooled_width, kernel_h, kernel_w, stride_h, stride_w, pad_t, pad_l,
+         top_data, mask);
+  }
   return d.ok();
 }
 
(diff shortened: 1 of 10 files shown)


jhseu:

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
```diff
@@ -28,6 +28,21 @@ limitations under the License.
 
 namespace tensorflow {
 namespace {
+// Checks env var to determine whether maxpool functions should propagate NaN
+// values. NaNs are propagated unless TF_ENABLE_MAXPOOL_NANPROP=0.
+bool QueryMaxPoolNanPropIsEnabled() {
+  bool default_value = true;
+  const char* tf_env_var = getenv("TF_ENABLE_MAXPOOL_NANPROP");
```
zheng-xq: Please use "ReadBoolFromEnvVar" instead.

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
```diff
@@ -76,7 +91,10 @@ __global__ void MaxPoolForwardNCHW(const int nthreads, const dtype* bottom_data,
     for (int h = hstart; h < hend; ++h) {
       for (int w = wstart; w < wend; ++w) {
         int idx = c * height * width + h * width + w;
-        if (bottom_data_n[idx] > maxval) {
+        // Note: The first inequality below ensures NaNs are propagated.
+        // Template argument used to avoid conditional eval in inner loop.
+        if (   (  propagate_nans && !(bottom_data_n[idx] <= maxval) )
```
zheng-xq: Put this into a common shared device inline function.

zheng-xq: Also please run the benchmarks and put in performance comparison before/after your change in the commit description. Thanks!

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
```diff
@@ -28,6 +28,21 @@ limitations under the License.
 
 namespace tensorflow {
 namespace {
+// Checks env var to determine whether maxpool functions should propagate NaN
+// values. NaNs are propagated unless TF_ENABLE_MAXPOOL_NANPROP=0.
+bool QueryMaxPoolNanPropIsEnabled() {
+  bool default_value = true;
```
zheng-xq: Check the default behavior is a high risk. You can potentially break models that are already working. 

It is safe to add a flag that works both way, and have the default do the old way. We can have a separate CL trying to flip the default. The default is to announce it, have people try it for a few weeks. If nobody complains, we will then switch. We need to be a lot more careful changing default behavior for existing ops, than to add new ops.

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -1271,11 +1271,8 @@ def _testMaxPoolGradDirectWithNans2_1(self):
       return
 
     # Test the GPU implementation that uses cudnn for now.
-    # It does not propagate the diff in cases of NaNs
-    expected_input_backprop_cudnn = [
-        0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
-        0.0, 0.0
-    ]
+    # It also propagates the diff in cases of NaNs
+    expected_input_backprop_cudnn = expected_input_backprop_tf_cpu
```
zheng-xq: Both ways need to be tested.

zheng-xq (changes requested):

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
```diff
@@ -28,6 +28,21 @@ limitations under the License.
 
 namespace tensorflow {
 namespace {
+// Checks env var to determine whether maxpool functions should propagate NaN
+// values. NaNs are propagated unless TF_ENABLE_MAXPOOL_NANPROP=0.
+bool QueryMaxPoolNanPropIsEnabled() {
```
zheng-xq: Thinking about this. If we were to add this feature, we will want to test it both ways. We also cannot leave it as a global/static variable, as it won't change from test it test. 

An alternative is to leave it part of the OpKernelConstruction. So it's initialized the first as the graph constructs. Then we can easily give different settings per test.

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
nluehr: Is there a targeted benchmark for this operation that you have in mind?

tensorflow/python/kernel_tests/pooling_ops_test.py
nluehr: Is it sufficient to make the test respect the environment variable? Currently, toggling NaN propagation off and on in a single test is not possible.

tensorflow/core/kernels/maxpooling_op_gpu.cu.cc
zheng-xq: How about the benchmarks in tensorflow/core/kernels/nn_ops_test.cc?

tensorflow/python/kernel_tests/pooling_ops_test.py
zheng-xq: Please look at my other comment for this. It's better to move the setting into the graph, meaning in the OpKernel constructors.

tfboyd:
Asking team about CLA for associate corporation.

drpngx:
@nluehr did you address @zheng-xq comments?

Jenkins, test this please.

nluehr:
Yes, all comments have been addressed.

drpngx:
Thanks! @zheng-xq PTAL?

tensorflow/python/kernel_tests/pooling_ops_test.py
```diff
@@ -1341,11 +1342,32 @@ def _testMaxPoolGradDirectWithNans2_1(self):
...
+          window_rows=2,
+          window_cols=2,
+          row_stride=1,
+          col_stride=1,
+          padding="VALID",
+          use_gpu=True,
+          v2=v2)
+
+    # Propagate the diff in cases of NaNs
+    os.environ["TF_ENABLE_MAXPOOL_NANPROP"] = "1"
```
zheng-xq: In both cases, remove this env-var setting at the end of the test so it has no effect on other tests.

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -616,7 +616,9 @@ class ScopedPoolingDescriptor {
       LOG(FATAL) << "could not create cudnn pooling descriptor: "
                  << ToString(status);
     }
-
+    bool propagate_nans;
+    tensorflow::ReadBoolFromEnvVar("TF_ENABLE_MAXPOOL_NANPROP",
```
zheng-xq: If I remember correct, this function is called very time the op is called. Can we cache the value on the first reading? There are a number of examples like that in this file.

zheng-xq (changes requested):

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: Most other examples of caching envars are static, but we need to be able to toggle the value in order to test both propagating and non-propagating nans conveniently. This will require plumbing the propagation flag through from the MaxPoolingGradOp<> class in maxpooling_op.cc. I'll also also add a nan-prop field to the dnn::PoolingDescriptor wrapper class.

zheng-xq: In that case, how about having the caller should pass it in. TensorFlow is better equipped to cache the value.

tensorflow/python/kernel_tests/pooling_ops_test.py
nluehr: Done

tensorflow/stream_executor/cuda/cuda_dnn.cc
nluehr: How is this?

nluehr:
@zheng-xq, are further changes needed here?

reedwm:
Unassigning myself for review since I'm not very familiar with the pooling code. Let's wait for @zheng-xq to get back.

nluehr:
@zheng-xq can this be merged?

zheng-xq:
LGTM

Please move ahead with merging.

zheng-xq (approved):

drpngx:
Jenkins, test this please.

drpngx:
Some kind of infra error (CC @gunan FYI)

```
Analyzing: 367 targets (111 packages loaded)
ERROR: no such package '@bazel_toolchains//configs/debian8_clang/0.1.0/bazel_0.6.0': BUILD file not found on package path
```

yifeif:
@nlopezgi any idea on the bazel_toolchains error?

nlopezgi:
The only reason I can think of that would manifest with this error is if the repo is behind commit 6425dbd10e9bc5a765807c25d3da109230840096 ( https://github.com/tensorflow/tensorflow/commit/6425dbd10e9bc5a765807c25d3da109230840096#diff-455a4c7f8e22d7c514e8c2caa27506c5), as those configs should be available after that commit (https://github.com/bazelbuild/bazel-toolchains/tree/b2b4b38433bf2d1159360855ea4004378308711b/configs/debian8_clang/0.1.0). Other than that, it might be a transient error? So I'd recommend checking the repo is sync'd and retrying. Let me know if the issue persists (and please, ideally, send a link with a full log of the command executed and the error). Sorry for the issues.

yifeif:
You are right @nlopezgi , the version in this PR branch is old. But we should really be testing against master. I will check with kokoro. Meanwhile to unblock this PR, we might need to do a rebase.

yifeif:
The branch infra issue should be fixed now.

PR #12490, opened 2017-08-22
Expose new CRF classes and function.

guillaumekln:
#12056 introduced a new tensor-based CRF decoding. This PR exposes the new names at the module level.

PR #12506, opened 2017-08-22
Update doc and imports for crf_decode

cBournhonesque:
A new method crf_decode was added recently (does the equivalent of viterbi_decode but with tf.Tensors as input).
However it is still not in the docs and it is not importable in tf.contrib.crf; hence the proposed changes.


cBournhonesque:
I've signed it!

guillaumekln:
See #12490.

PR #12509, opened 2017-08-23
remove unused code, it does not happen

horance-liu:
I remove some implements for `GetDeviceLocalityAsync`, because it does not happen.

### DIFF
--- a/tensorflow/core/distributed_runtime/worker_cache_partial.cc
+++ b/tensorflow/core/distributed_runtime/worker_cache_partial.cc
@@ -29,7 +29,7 @@ namespace tensorflow {
 bool WorkerCachePartial::GetDeviceLocalityNonBlocking(
     const string& device_name, DeviceLocality* locality) {
   mutex_lock lock(mu_);  // could use reader lock
-  const auto& iter = device_status_cache_.find(device_name);
+  auto iter = device_status_cache_.find(device_name);
   if (iter != device_status_cache_.end()) {
     *locality = iter->second.locality();
     return true;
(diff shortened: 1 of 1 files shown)


mrry:
Transfering this to @poxvoculi, because I think he wrote the code that relies on this check.

byronyi:
We might need this in the future. See https://github.com/tensorflow/tensorflow/pull/11392#issuecomment-315818751

poxvoculi:
@byronyi I think this PR is essentially just removing the error case
   errors::Internal("Failed to find locality for ", device_name);
which cannot be hit, due to the immediately prior call to GetDeviceLocalityNonBlocking.
(Now a default value for DeviceAttributes.locality is returned, but perhaps an earlier version of GetDeviceLocalityNonBlocking could have returned an error if the table entry existed but had no locality value.)

How do you anticipate hitting this case in the future?

byronyi:
Sorry, I shall be fine with the changes here. As a side note, there could be three different cases here:

1. There's a valid table entry (the happy path)
2. There's no _valid_ table entry, because the corresponding device attributes can't be found
3. There's no table entry at all, as the RPC has not returned from the remote

Correct me if I'm wrong, but now it seems the 2nd and 3rd cases could not be differentiated. Like you said, an early implementation could mark the 2nd case with an _invalid_ table entry, and return false from GetDeviceLocalityNonBlocking (and thus we might need what will be removed in this PR). And for the current implementation, would there be any chances for the caller to keep refreshing the table for nonexistent device attributes, i.e. mistaken case 2 as case 3?

poxvoculi:
@byronyi In an earlier version of protobufs it was possible to distinguish a missing member from a default value.  That's no longer the case, so [this line ](https://github.com/petewarden/tensorflow_makefile/blob/master/tensorflow/core/distributed_runtime/worker_cache_partial.cc#L34) should either return a default value or a good value set by RefreshDeviceStatus, and we can't tell which one.  In the future, if useful, we could set the default value to some distinguishable invalid value.

poxvoculi (approved):

byronyi:
Thanks for the explanation! Very much helpful as always :)

martinwicke:
Jenkins, test this please.

PR #12513, opened 2017-08-23
make Dimension be compatible with integer

### LINKED ISSUE #11974: tf.reshape does not accept Dimension objects for the shape parameter

facaiy:
The PR is another solution to fix #11974 .  I believe that it is better than #12127 .

Make Dimension to be compatible with integer, so `[1, Dimension(2)]` will be casted to `[1, 2]` automatically.

+ [x] add a doctest. 
+ [x] add an unit test.
+ [x] pass all unit tests.


facaiy:
cc @yaroslavvb @pmccarter @martinwicke , who I believe are interested in this PR. Could you take a look please?

tensorflow/python/framework/tensor_shape.py
```diff
@@ -428,6 +429,8 @@ def __init__(self, dims):
       TypeError: If dims cannot be converted to a list of dimensions.
     """
     # TODO(irving): Eliminate the single integer special case.
+    from tensorflow.python.util import compat
```
alextp: imports should be at the top of the file in tensorflow

tensorflow/python/util/compat.py
```diff
@@ -40,6 +40,7 @@
 import numpy as _np
 import six as _six
 
+from tensorflow.python.framework import tensor_shape
```
alextp: I don't like this dependency. util should be small utilities which do not depend on any of the concrete tensorflow stuff.

I think the other fix is better.

tensorflow/python/framework/tensor_shape.py
facaiy: `tensor_shape` and `compat` are depended by each other if imports are at the top. I don't know how to deal with the circular imports.

tensorflow/python/util/compat.py
facaiy: Thanks for comment. Could you please give me a bit more detail about where to fix?

How about hack [`def _FilterInt(v):`](https://github.com/tensorflow/tensorflow/blob/c48aaadbe9c5802316d73e40bbcde34221277b81/tensorflow/python/framework/tensor_util.py#L238) function? 
or add a `if` branch at [`_AssertCompatible(values, dtype)`](https://github.com/tensorflow/tensorflow/blob/c48aaadbe9c5802316d73e40bbcde34221277b81/tensorflow/python/framework/tensor_util.py#L383)?

alextp: Essentially, I do not think that this approach is the right way to fix the problem. compat.as_int shouldn't know anything about dimension. I think something like the change made in  #12558  is the right way to go.

I'll close this pull request now.

facaiy: Hi, @alextp . Perhaps the [new commit](https://github.com/facaiy/tensorflow/commits/BUG/tf_reshape_mix_int_with_dim) will be interesting for you. Unfortunately, the PR has been closed.

facaiy: How about #12127 ? It is more straightforward.

tensorflow/core/ops/ops.pbtxt
```diff
@@ -19893,7 +19893,7 @@ op {
...
...\'t\' is [[[1, 1], [2, 2]],\n#                [[3, 3], [4, 4]]]\n# tensor \'t\' has shape [2, 2, 2]\nreshape(t, [2, 4]) ==> [[1, 1, 2, 2],\n                        [3, 3, 4, 4]]\n\n# tensor \'t\' is [[[1, 1, 1],\n#                 [2, 2, 2]],\n#                [[3, 3, 3],\n#                 [4, 4, 4]],\n#                [[5, 5, 5],\n#                 [6, 6, 6]]]\n# tensor \'t\' has shape [3, 2, 3]\n# pass \'[-1]\' to flatten \'t\'\nreshape(t, [-1]) ==> [1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6]\n\n# -1 can also be used to infer the shape\n\n# -1 is inferred to be 9:\nreshape(t, [2, -1]) ==> [[1, 1, 1, 2, 2, 2, 3, 3, 3],\n                         [4, 4, 4, 5, 5, 5, 6, 6, 6]]\n# -1 is inferred to be 2:\nreshape(t, [-1, 9]) ==> [[1, 1, 1, 2, 2, 2, 3, 3, 3],\n                         [4, 4, 4, 5, 5, 5, 6, 6, 6]]\n# -1 is inferred to be 3:\nreshape(t, [ 2, -1, 3]) ==> [[[1, 1, 1],\n                              [2, 2, 2],\n                              [3, 3, 3]],\n                             [[4, 4, 4],\n                              [5, 5, 5],\n                              [6, 6, 6]]]\n\n# tensor \'t\' is [7]\n# shape `[]` reshapes to a scalar\nreshape(t, []) ==> 7\n```"
```
alextp: Revert this change? This is a generated file.

alextp:
I am ready to approve this when the change to ops.pbtxt gets reverted.

alextp:
Jenkins, test this please.

alextp (approved):

alextp:
Jenkins, test this please.

facaiy:
Thanks. The new commits are modified on my personal laptop. It is no surprise if some unit tests are broken. If so, I'll fix them on next Monday.

facaiy:
It seems a trivial mistake, and has been fixed. Could you retest the PR? Thanks.

alextp:
Jenkins, test this please.

facaiy:
Thanks, @alextp .  All tests pass.

PR #12516, opened 2017-08-23
Update head.py

alanyee:
-Use tf.metrics instead of contrib version
-Use core lookup instead of contrib version (Nvm, another PR)
-Remove TODO in tensorflow/contrib/lookup/__init__.py since already done


ispirmustafa (approved):
Thank you Alan

jhseu:
Jenkins, test this please

PR #12531, opened 2017-08-23
Enable int16 and uint16 for reverse op

### LINKED ISSUE #12528: Data Augmentation of uint16 images (such as PNG)
How we can do data augmentation (rotation, flip,...) of uint16 images?

Here is a piece of my code:

import os
import tensorflow as tf

filQ=['ex4497_fbp_13.png','ex4497_fbp_12.png']
filQQ=tf.train.string_input_producer(filQ)

reader =tf.WholeFileReader()
key, value = reader.read(filQQ)

myimg = tf.image.decode_png(value, dtype=tf.uint16) 
myimg1 = tf.image.rot90(myimg,k=1,name=None)

I get an error that:

TypeError: Value passed to parameter 'tensor' has DataType uint16 not in list of allowed values: uint8, int8, int32, int64, bool, float16, float32, float64, complex64, complex128, string

So basically unit16 doesn't exist!
unit8 is not a good choice because the pixel values in my image are NOT in the range [0 255],
and we can only use data type uint8, uint16 for PNG in tensorflow!!

So again the question becomes:
**How we can do data augmentation (rotation, flip,...) of uint16 images?**

yongtang:
This fix tries to address the issue in #12528 where it was not possible to run uint16 on rot90 (reverse) op

The int16 and uint16 has been implemented in CPU though they were not enabled yet.

This fix enables int16 and uint16.

This fix fixes #12528.

### DIFF
--- a/tensorflow/core/ops/array_ops.cc
+++ b/tensorflow/core/ops/array_ops.cc
@@ -1006,7 +1006,8 @@ REGISTER_OP("Reverse")
     .Input("dims: bool")
     .Output("output: T")
     .Attr(
-        "T: {uint8, int8, int32, int64, bool, half, float, double, complex64, "
+        "T: {uint8, int8, uint16, int16, int32, int64, bool, half, float, "
+        "double, complex64, "
         "complex128, string}")
     .SetShapeFn([](InferenceContext* c) {
       ShapeHandle input = c->input(0);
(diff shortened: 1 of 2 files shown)


jhseu (approved):

jhseu:
Jenkins, test this please

PR #12535, opened 2017-08-23
Added shape information to the while loop input placeholders for the C API.

eaplatanios:
As discussed with @skye :)

### DIFF
--- a/tensorflow/python/saved_model/builder_impl.py
+++ b/tensorflow/python/saved_model/builder_impl.py
@@ -76,25 +76,16 @@ class SavedModelBuilder(object):
   ```
   """
 
-  def __init__(self, export_dir, overwrite=False):
-    """Construct a SavedModelBuilder.
-
-    Args:
-      export_dir: The directory to export to
-      overwrite: If `True` deletes export_dir if it already exists
-    """
+  def __init__(self, export_dir):
     self._saved_model = saved_model_pb2.SavedModel()
     self._saved_model.saved_model_schema_version = (
         constants.SAVED_MODEL_SCHEMA_VERSION)
 
     self._export_dir = export_dir
     if file_io.file_exists(export_dir):
-      if overwrite:
-        file_io.delete_recursively(export_dir)
-      else:
-        raise AssertionError(
-            "Export directory already exists. Please specify a different export "
-            "directory: %s or change overwrite flag to `True`" % export_dir)
+      raise AssertionError(
+          "Export directory already exists. Please specify a different export "
+          "directory: %s" % export_dir)
 
     file_io.recursive_create_dir(self._export_dir)
 


tensorflow/c/c_api.cc
```diff
@@ -1812,11 +1812,19 @@ void TF_GraphImportGraphDef(TF_Graph* graph, const TF_Buffer* graph_def,
...
-bool CreateInput(const TF_Output& parent_input, TF_Graph* g, const char* name,
-                 TF_Output* input, TF_Status* status) {
+bool CreateInput(TF_Graph* parent_graph, const TF_Output& parent_input, TF_Graph* g,
+                 const char* name, TF_Output* input, TF_Status* status) {
   TF_OperationDescription* desc = TF_NewOperation(g, "Placeholder", name);
   TF_SetAttrType(desc, "dtype", TF_OperationOutputType(parent_input));
-  // TODO(skyewm): set placeholder shape
+  // Set placeholder shape.
+  int num_dims = TF_GraphGetTensorNumDims(parent_graph, parent_input, status);
+  if (status->status.ok() && num_dims >= 0) {
```
skye: Return if !status.ok() (here and below)

tensorflow/c/c_api.cc
```diff
@@ -1812,11 +1812,19 @@ void TF_GraphImportGraphDef(TF_Graph* graph, const TF_Buffer* graph_def,
...
+    int64_t* dims = new int64_t[num_dims];
```
skye: Use a vector instead of new (std::vector<int64_t> dims(num_dims), and then use dims.data() and dims.size()). I believe you are leaking dims, and using raw memory allocations like this is in general very dangerous.

tensorflow/c/c_test_util.cc
```diff
@@ -36,17 +36,31 @@ TF_Tensor* Int8Tensor(const int64_t* dims, int num_dims, const char* values) {
...
-  values[0] = v;
-  return TF_NewTensor(TF_INT32, nullptr, 0, values, num_bytes,
-                      &Int32Deallocator, nullptr);
+TF_Tensor* Int32Tensor(int32_t v, const tensorflow::TensorShape shape) {
+  int num_dims = shape.dims();
+  int64_t* dims = new int64_t[num_dims];
+  for (int i = 0; i < num_dims; ++i)
+    dims[i] = reinterpret_cast<int64_t>(shape.dim_size(i));
+  int64_t num_values = 1;
+  for (int i = 0; i < num_dims; ++i)
```
skye: The style guide says braces are optional for single-statement loops, but I prefer braces :) (or put the whole thing on one line without braces)

tensorflow/c/c_test_util.cc
```diff
@@ -36,17 +36,31 @@ TF_Tensor* Int8Tensor(const int64_t* dims, int num_dims, const char* values) {
...
 
-TF_Tensor* Int32Tensor(int32_t v) {
-  const int num_bytes = sizeof(int32_t);
-  int32_t* values = new int32_t[1];
-  values[0] = v;
-  return TF_NewTensor(TF_INT32, nullptr, 0, values, num_bytes,
-                      &Int32Deallocator, nullptr);
+TF_Tensor* Int32Tensor(int32_t v, const tensorflow::TensorShape shape) {
+  int num_dims = shape.dims();
+  int64_t* dims = new int64_t[num_dims];
```
skye: I think you can use shape.dim_sizes() directly instead of creating your own array (here and below).

skye (changes requested):

eaplatanios:
@skye Thanks for pointing out those things. I made all the requested changes. :)

tensorflow/c/while_loop_test.cc
```diff
@@ -33,13 +33,13 @@ class CApiWhileLoopTest : public ::testing::Test {
     TF_DeleteStatus(s_);
   }
 
-  void Init(int ninputs) {
+  void Init(int ninputs, const tensorflow::TensorShape shape = {}) {
```
skye: Can you actually make this a PartialTensorShape? That's the only way we can get coverage of unknown-rank inputs (TensorShape = {} corresponds to a scalar, but PartialTensorShape() is unknown rank).

Once you do that, manually specify a scalar shape for BasicLoop or one of the other tests so we get coverage of known shapes working too.

skye (changes requested):
Looking good

eaplatanios:
Done :)

tensorflow/c/c_test_util.h
```diff
@@ -33,16 +35,18 @@ typedef std::unique_ptr<TF_Tensor, decltype(&TF_DeleteTensor)>
 // Create a tensor with values of type TF_INT8 provided by `values`.
 TF_Tensor* Int8Tensor(const int64_t* dims, int num_dims, const char* values);
 
-TF_Tensor* Int32Tensor(int32_t v);
+TF_Tensor* Int32Tensor(int32_t v, const tensorflow::TensorShape& shape = {});
 
 TF_Operation* Placeholder(TF_Graph* graph, TF_Status* s,
-                          const char* name = "feed");
+                          const char* name = "feed",
+                          const tensorflow::PartialTensorShape& shape = tensorflow::PartialTensorShape());
```
skye: Long line?

tensorflow/c/while_loop_test.cc
```diff
@@ -33,13 +33,13 @@ class CApiWhileLoopTest : public ::testing::Test {
     TF_DeleteStatus(s_);
   }
 
-  void Init(int ninputs) {
+  void Init(int ninputs, const tensorflow::PartialTensorShape shape = {}) {
```
skye: Make the default PartialTensorShape() like above, I think this is a scalar

tensorflow/c/c_test_util.h
eaplatanios: What is your convention for breaking these lines? In my code I usually break after the opening parenthesis of the argument list.

skye: Here's what clang-format gives me:
```
TF_Operation* Placeholder(TF_Graph* graph, TF_Status* s,
                          const char* name = "feed",
                          const tensorflow::PartialTensorShape& shape =
                              tensorflow::PartialTensorShape());
```

FYI I highly recommend using clang-format (set to Google style). There's an emacs plugin I use, and maybe there are plugins for other editors too. I can provide more info if necessary.

eaplatanios: Thanks! I'll look into that. I'm used to IntelliJ for coding in Java, Scala, or Python and I don't really have a preferred editor for C++ as I don't use it as much. I've started using CLion recently, which I like, but I feel it's not very mature yet.

eaplatanios:
I hope it all looks good now! :)

tensorflow/c/c_test_util.h
skye: It looks like you might able to use it from clion: https://stackoverflow.com/questions/34648255/using-clang-format-in-clion. I think this requires installing the standalone version of clang-format, which you can possibly use directly to at least see its formatting suggestions and manually apply them. I'm not sure though. Anyway, maybe worth looking into depending on how hard manually formatting is.

skye (approved):
Thanks Anthony!!

PR #12544, opened 2017-08-24
remove temp variable

horance-liu:
1. I remove temp variable, and return it directly. 
2. using `auto` deduction.

### DIFF
--- a/tensorflow/core/distributed_runtime/rpc/grpc_worker_cache.cc
+++ b/tensorflow/core/distributed_runtime/rpc/grpc_worker_cache.cc
@@ -43,7 +43,7 @@ class GrpcWorkerCache : public WorkerCachePartial {
           void* tag;
           bool ok;
           while (completion_queue_.Next(&tag, &ok)) {
-            GrpcClientCQTag* callback_tag = static_cast<GrpcClientCQTag*>(tag);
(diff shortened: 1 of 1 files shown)


tensorflow/core/distributed_runtime/rpc/grpc_worker_cache.cc
```diff
@@ -43,7 +43,7 @@ class GrpcWorkerCache : public WorkerCachePartial {
           void* tag;
           bool ok;
           while (completion_queue_.Next(&tag, &ok)) {
-            GrpcClientCQTag* callback_tag = static_cast<GrpcClientCQTag*>(tag);
+            auto callback_tag = static_cast<GrpcClientCQTag*>(tag);
```
jhseu: Mind excluding these auto changes? It doesn't match the majority of TF style.

horance-liu: I have seen many examples for using `auto`, and here there is no need to write `GrpcClientCQTag*` with two times. So, I apply this change not only to use `auto` deduction, but to eliminate duplicate type declaration.

horance-liu:
I have rollback modification, and please review again.

jhseu (approved):

martinwicke:
Jenkins, test this please.

Issue #12537, opened 2017-08-23
Printing lines without logging prefix using tf.Print()


danijar:
Right now, all messages printed via `tf.Print()` are prefixed by:

```
<timestamp>: I tensorflow/core/kernels/logging_ops.cc:79]
```

It would be useful to have a way to print user-friendly output from within the graph. Therefore, suggest adding a parameter for the logging format to `tf.Print()`, or at least adding a flag that disables the logging prefix.

vrv:
The prefix comes from LOG(INFO) inside the logging_ops kernel.  The fix needs to be made at the C++ level, though I don't have a good recommendation as to how.

PR #12559, opened 2017-08-24
#12537 issue solution

yoon-hyoung:
* printing prefix message by parameter value

### DIFF
--- a/tensorflow/python/ops/logging_ops.py
+++ b/tensorflow/python/ops/logging_ops.py
@@ -36,15 +36,11 @@ from tensorflow.python.util.deprecation import deprecated
 # Assert and Print are special symbols in python, so we must
 # use an upper-case version of them.
 def Print(input_, data, message=None, first_n=None, summarize=None,
-          name=None):
+          name=None,flag=False):
   """Prints a list of tensors.
 
   This is an identity op with the side effect of printing `data` when
   evaluating.
-
-  Note: This op prints to the standard error. It is not currently compatible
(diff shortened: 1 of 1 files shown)


tensorflow/python/ops/logging_ops.py
```diff
@@ -124,7 +125,7 @@ def image_summary(tag, tensor, max_images=3, collections=None, name=None):
   """Outputs a `Summary` protocol buffer with images.
 
   For an explanation of why this op was deprecated, and information on how to
-  migrate, look ['here'](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/deprecated/__init__.py)
+  migrate, look ['here'](https://www.tensorflow.org/code/tensorflow/contrib/deprecated/__init__.py)
```
byronyi: This link appears to be broken. I do not understand why you would like to change the links here.

byronyi: This link appears to be broken. I do not understand why you would like to change the links here.

vrv (commented):
I don't think this works, message defaults to None and so the flag doesn't do anything different.

The prefix comes from the C++ LOG(INFO) statement.  One would have to find a way to prevent the prefix from being printed there, probably via an environment variable.

There are also unrelated changes in this PR, which you would want to remove.

drpngx:
Closing, since it's not clear that this achieves the desired result. Feel free to cleanup and reopen with comments. Thanks!

Issue #11847, opened 2017-07-28
Bug: Op type not registered 'BlockLSTM' in binary


ShapingView:
### Describe the problem
I want to load and run a single tensorflow model within another C++ project. To do this, I defined the `tensorflow_all` library in [`tensorflow/BUILD`] which should include all the necessary dependencies. (Details below)

When loading the tensorflow model via the C++ API ([LoadSavedModel]), the following runtime error occurs:

[LoadSavedModel]: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/cc/saved_model/loader.h#L50

```
2017-07-28 13:36:41.875493: I tensorflow/cc/saved_model/loader.cc:284] Loading SavedModel: fail. Took 43550 microseconds.
terminate called after throwing an instance of 'std::runtime_error'
  what():  Not found: Op type not registered 'BlockLSTM' in binary running on myMachine. 
  Make sure the Op and Kernel are registered in the binary running in this process.
```

I found that the BlockLSTM Op is registered in [`tensorflow/contrib/rnn/ops/lstm_ops.cc`].
However, I could not find a way to include the operation BlockLSTM in my C++ library `tensorflow_all`.

[`tensorflow/contrib/rnn/ops/lstm_ops.cc`]: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/rnn/ops/lstm_ops.cc#L179

Bear with me if this is actually not a bug (or even a feature request). I am still getting started with tensorflow. 

Since my implementation is part of another larger project, it is currently difficult for me to give an easy to reproduce example. However, if this is necessary, I will do it as soon as I find the time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:

	Yes, a colleague of mine trained a tensorflow graph which uses the BlockLSTM Op from `tensorflow/contrib/rnn/` using the python API. 

	As mentioned before, I added code to the [`tensorflow/BUILD`] to create a library `tensorflow_all`. This library is used to load the graph externally.

	[`tensorflow/BUILD`]: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/BUILD

	```bazel
	cc_binary(
		name = "libtensorflow_all.so",
		linkshared = 1,
		linkopts = ["-Wl,--version-script=tensorflow/tf_version_script.lds"],
		deps = [
			"//tensorflow/cc:cc_ops",
			"//tensorflow/cc:client_session",
[...]

PR #12566, opened 2017-08-24
enable use of transform_graph tool with contrib/rnn

pks:
This is a simple fix for using contrib/rnn with the transform_graph (and possibly other tools).

In particular, it fixes the following error:
```
2017-08-24 16:07:41.015752: E tensorflow/tools/graph_transforms/transform_graph.cc:210] Op type not registered 'BlockLSTM' in binary running on ip-172-31-30-181. Make sure the Op and Kernel are registered in the binary running in this process.
```

It's also possibly a remedy for issue #11847.

### DIFF
--- a/tensorflow/contrib/rnn/BUILD
+++ b/tensorflow/contrib/rnn/BUILD
@@ -328,6 +328,8 @@ tf_kernel_library(
     srcs = [
         "kernels/blas_gemm.cc",
         "kernels/blas_gemm.h",
+        "kernels/gru_ops.h",
+        "ops/gru_ops.cc",
     ],
     gpu_srcs = [
         "kernels/blas_gemm.h",
@@ -346,6 +348,8 @@ tf_kernel_library(
     srcs = [
         "kernels/blas_gemm.cc",
         "kernels/blas_gemm.h",
+        "kernels/lstm_ops.h",
+        "ops/lstm_ops.cc",
     ],
     gpu_srcs = [
         "kernels/blas_gemm.h",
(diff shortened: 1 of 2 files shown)


jhseu:
Please address the CLA before we look at the change. Thanks!

dkurt:
@pks, unfortunately, it doesn't solve a `ValueError: No op named BlockLSTM in defined operations.` problem mentioned in the issue. But `from tensorflow.contrib.rnn import *` does. Thank you!

pks:
@dkurt the proposed fix here fixes the transform_graph tool, not freeze_graph.py -- but glad that the import works :-)

pks:
I signed it!

tensorflow/contrib/rnn/BUILD
```diff
@@ -346,6 +348,8 @@ tf_kernel_library(
     srcs = [
         "kernels/blas_gemm.cc",
         "kernels/blas_gemm.h",
+        "kernels/lstm_ops.h",
```
jhseu: You shouldn't add these here. Instead add the dependencies
`"gru_ops_op_lib"` and `"lstm_ops_op_lib"` to the target

pks:
That works as well. Thanks!

drpngx:
Jenkins, test this please.

@petewarden any opinion on this PR?

sb2nov:
@petewarden ping?

petewarden (approved):
Sorry for the slow response, this looks good to me.

martinwicke:
MacOS failure is unrelated.

Issue #12298, opened 2017-08-15
iOS: thread pools appear to spin forever, even after session close


sadlerjw:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No. Built for iOS using provided scripts. Defined ops to be included in build as documented (using `ops_to_register.h`).
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Built on macOS 10.12.6 for distribution on iOS
- **TensorFlow installed from (source or binary)**: source
- **TensorFlow version (use command below)**: Unsure, sorry, my python environment is all messed up right now. My HEAD is at 0d2f6918322c7bf29d1de3075b0d4ed3b1b72919
- **Python version**: 2.7.12, but I believe that's irrelevant for this issue
- **Bazel version (if compiling from source)**: 0.5.1-homebrew, but I'm using `build_all_ios.sh` instead of Bazel to build
- **CUDA/cuDNN version**: Unknown
- **GPU model and memory**: iPhone 7 Plus, but AFAIK GPU is unavailable on iOS
- **Exact command to reproduce**: No particular command. Please see description of issue

Also [asked about this](https://stackoverflow.com/questions/45682837/tensorflow-thread-pools-appear-to-spin-forever-even-after-session-is-closed) on Stack Overflow, where it was suggested I file an issue

### Describe the problem
I believe this is a bug in Tensorflow. I am running Tensorflow on iOS, using the C++ API. I'm doing some image classification. I have a long-lived session, and I call `Run` many times on it, to evaluate different images from a backlog. Once I'm done, the `RunQueue`s (via `NonBlockingThreadPool`s) continue to pin the CPU at near max usage. They appear to be stuck in the `Steal` loop, presumably with no work to do.

### Source code / logs
I tried `Close()`ing and then `delete`ing the session, and having read some of the C++ source, this *should* have shut down the thread pools that belong to the session, but this didn't change the situation:
```objc++
auto status = session->Close();
delete session;
session = nil;
```
I tried setting some specific configuration options so I could be sure that the session did in fact own its thread pools instead of using a global thread pool, but this didn't help either:
```objc++
tensorflow::SessionOptions options;
options.config.clear_session_inter_op_thread_pool();
options.config.set_use_per_session_threads(true);
auto status = tensorflow::NewSession(options, &session);
```

[...]

PR #12573, opened 2017-08-24
Updated iOS build to handle threading correctly, requires iOS 9 or later

petewarden:
There was a long-standing issue with Apple's version of clang not supporting thread local attributes, which are used inside Eigen in some places. As a 'temporary' hack, I defined these attributes out of existence, but this lingered for a lot longer than it should have, and I believe is the cause of issues like #12298. From comments there testing this fix, it looks like it does help.

iOS 9 and later support `thread_local`, so as an improvement this change moves to that. It will block building on older versions, but now that v11 is almost out I think supporting two versions back should be ok for most developers (though feedback is welcome if I'm wrong).


jhseu (approved):

jhseu:
Jenkins, test this please

PR #8153, opened 2017-03-07
Mkl bfc allocator and static memory allocator registry

jbobba:
Pull request for
1) Static Registry of CPU Memory Allocators that can be queried to return the highest priority allocator
2) A BFC-based memory allocator that MKL backends can use to request memory from

PR #12576, opened 2017-08-25
remove unused cpu allocator factory

horance-liu:
Because `MakeCpuAllocator` was replaced with `cpu_allocator`,  so I remove unused `MakeCpuAllocator` factory method.

### DIFF
--- a/tensorflow/core/framework/allocator.cc
+++ b/tensorflow/core/framework/allocator.cc
@@ -117,16 +117,6 @@ class CPUAllocator : public Allocator {
   TF_DISALLOW_COPY_AND_ASSIGN(CPUAllocator);
 };
 
-namespace {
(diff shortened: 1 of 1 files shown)


byronyi:
I do not see any benefits brought by this patch.

Note that there is some bits of TF not open sourced (i.e. internal to Google, or other private branches), it is not always a good idea to judge if a piece of code is unused solely by what you see here.

See #8153 for a little bit of background why it is introduced in the first place.

yaroslavvb:
cc @reedwm who knows about memory allocators

jhseu (approved):

jhseu:
Jenkins, test this please

martinwicke:
GPU failure is unrelated.

Issue #10857, opened 2017-06-20
No way to freeze fused BN stats


DeepLearningJohnDoe:
When fine-tuning networks trained with BN sometimes we want to freeze and use the accumulated moving averages while allowing the gradients to be backpropagated through the BN layer, but currently there is no way of doing so with fused BN, since when is_training = False the layer gives erroneous gradients. Of course, we could use the batch statistics from the new task to accumulate the stats, but it isn't possible in the case of batch_size = 1.

I understand that due to the nature of the CuDNN kernel it might be hard to implement such feature, but a fused Batch Renorm layer could be a decent compromise, as it uses the moving averages when training as well as during inference.

tatatodd:
@zhangyaobit can you comment on this?  Thanks!

zhangyaobit:
Setting is_training to False and at the same time doing gradient computation is expected to get erroneous gradients, because the second and third outputs of FusedBatchNorm are only applicable/valid when is_training is True. 

Could you try train with fused batch norm, and fine-tuning with non-fused batch norm?

DeepLearningJohnDoe:
In object detection and semantic segmentation sometimes the batch_size is set to 1 for the original network, and unfortunately our use case fall under this category. Non-fused batch norm for NCHW is around 6x slower https://github.com/tensorflow/tensorflow/issues/7551#issuecomment-280421351.

zhangyaobit:
Not sure if I understand correctly: are you saying fine tune is done with a batch size of 1 and non-fused batch norm doesn't work with a batch size of 1? And even if it works, it is 6x slower?

How long do you expect fine tune currently takes (on the scale of hours or days? say on GPU)?

DeepLearningJohnDoe:
Yes, fine tune is done with a batch size of 1 (on each GPU). I am training on a 4x 1080 machine, and fine tuning would take around ~20 hours. Non-fused batch norm does work (gives correct gradients), but it is way too slow and will push the fine tuning time beyond 100 hours. It would be faster to simply retrain the original network in NHWC and fine tune there with non-fused batch norm, which is ironic, considering NCHW is the CuDNN canonic format.

zhangyaobit:
It may make sense to simply forward the population mean and variance to the second and third outputs of FusedBatchNorm when fused=Tue. There is no performance penalty for this, since it is a forwarding instead of a copy.

[...]

PR #12580, opened 2017-08-25
Add kernels for FusedBatchNormGrad when is_training=False

ppwwyyxx:
#10857

### DIFF
--- a/tensorflow/core/kernels/fused_batch_norm_op.cc
+++ b/tensorflow/core/kernels/fused_batch_norm_op.cc
@@ -17,12 +17,12 @@ limitations under the License.
 
 #if GOOGLE_CUDA
 #define EIGEN_USE_GPU
-#include "tensorflow/core/kernels/fused_batch_norm_op.h"
 #include "tensorflow/core/kernels/conv_2d.h"
 #include "tensorflow/core/kernels/conv_ops_gpu.h"
 #include "tensorflow/core/util/stream_executor_util.h"
 #endif
 
+#include "tensorflow/core/kernels/fused_batch_norm_op.h"
 #include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
 #include "tensorflow/core/framework/op_kernel.h"
 #include "tensorflow/core/framework/register_types.h"
--- a/tensorflow/core/kernels/fused_batch_norm_op.cc
+++ b/tensorflow/core/kernels/fused_batch_norm_op.cc
@@ -352,7 +352,7 @@ template <typename T>
 struct FusedBatchNormGrad<GPUDevice, T> {
   void operator()(OpKernelContext* context, const Tensor& y_backprop,
                   const Tensor& x, const Tensor& scale, const Tensor& mean,
-                  const Tensor& variance, T epsilon, Tensor* x_backprop,
+                  const Tensor& inv_variance, T epsilon, Tensor* x_backprop,
                   Tensor* scale_backprop, Tensor* offset_backprop,
                   TensorFormat tensor_format) {
     auto* stream = context->op_device_context()->stream();
@@ -441,16 +441,17 @@ struct FusedBatchNormGrad<GPUDevice, T> {
     auto x_ptr = StreamExecutorUtil::AsDeviceMemory<T>(x_maybe_transformed);
     auto scale_ptr = StreamExecutorUtil::AsDeviceMemory<T>(scale);
     auto mean_ptr = StreamExecutorUtil::AsDeviceMemory<T>(mean);
-    auto variance_ptr = StreamExecutorUtil::AsDeviceMemory<T>(variance);
+    auto inv_variance_ptr = StreamExecutorUtil::AsDeviceMemory<T>(inv_variance);
     auto scale_backprop_ptr =
         StreamExecutorUtil::AsDeviceMemory<T>(*scale_backprop);
     auto offset_backprop_ptr =
         StreamExecutorUtil::AsDeviceMemory<T>(*offset_backprop);
 
+    // the cudnn kernel outputs inverse variance in forward and reuse it in backward
     bool cudnn_launch_status =
         stream
             ->ThenBatchNormalizationBackward(
-                y_backprop_ptr, x_ptr, scale_ptr, mean_ptr, variance_ptr,
+                y_backprop_ptr, x_ptr, scale_ptr, mean_ptr, inv_variance_ptr,
                 x_desc, scale_offset_desc, static_cast<double>(epsilon),
                 &x_backprop_ptr, &scale_backprop_ptr, &offset_backprop_ptr)
             .ok();
@@ -546,13 +547,13 @@ class FusedBatchNormOp : public OpKernel {
     Tensor* saved_mean = nullptr;
     OP_REQUIRES_OK(context,
                    context->allocate_output(3, scale.shape(), &saved_mean));
-    Tensor* saved_inv_var = nullptr;
+    Tensor* saved_maybe_inv_var = nullptr;
     OP_REQUIRES_OK(context,
-                   context->allocate_output(4, scale.shape(), &saved_inv_var));
+                   context->allocate_output(4, scale.shape(), &saved_maybe_inv_var));
 
     functor::FusedBatchNorm<Device, T>()(
         context, x, scale, offset, estimated_mean, estimated_variance, epsilon_,
-        y, batch_mean, batch_var, saved_mean, saved_inv_var, tensor_format_,
+        y, batch_mean, batch_var, saved_mean, saved_maybe_inv_var, tensor_format_,
         is_training_);
   }
 
@@ -581,10 +582,11 @@ class FusedBatchNormGradOp : public OpKernel {
     const Tensor& y_backprop = context->input(0);
     const Tensor& x = context->input(1);
     const Tensor& scale = context->input(2);
-    const Tensor& saved_mean = context->input(3);
-    // The Eigen implementation saves variance in the forward pass,  while cuDNN
-    // saves inverted variance.
-    const Tensor& saved_maybe_inv_var = context->input(4);
+    // When is_training=True, batch mean and variance/inverted variance are saved in the forward pass to be reused here.
+    // When is_training=False, population mean and variance need to be forwarded here to compute the gradients.
+    const Tensor& saved_mean_or_pop_mean = context->input(3);
+    // The Eigen implementation saves variance in the forward pass, while cuDNN saves inverted variance.
+    const Tensor& saved_maybe_inv_var_or_pop_var = context->input(4);
 
     OP_REQUIRES(context, y_backprop.dims() == 4,
                 errors::InvalidArgument("input must be 4-dimensional",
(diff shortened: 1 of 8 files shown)


tensorflow/core/kernels/fused_batch_norm_op.h
```diff
@@ -43,6 +45,66 @@ struct InvVarianceToVariance {
...
+// Functor used by FusedBatchNormGradOp to do the computations when is_training=False.
+// Both CPU and GPU will use this functor.
+template <typename Device, typename T>
+struct FusedBatchNormFreezeGrad {
+  void operator()(const Device& d, const Tensor& y_backprop_input,
+                  const Tensor& x_input, const Tensor& scale_input,
+                  const Tensor& mean_input, const Tensor& variance_input,
+                  T epsilon, Tensor* x_backprop_output,
+                  Tensor* scale_backprop_output, Tensor* offset_backprop_output,
+                  typename TTypes<T>::Vec scratch1, typename TTypes<T>::Vec scratch2) {
```
zhangyaobit: Are these two scratch allocation needed? Could you follow the Eigen implementation of FusedBatchNorm, where  no temp allocation is needed (you can still use something like "Eigen::Tensor<T, 1, Eigen::RowMajor> mean(depth)" though).

tensorflow/core/kernels/fused_batch_norm_op.h
```diff
@@ -43,6 +45,66 @@ struct InvVarianceToVariance {
   void operator()(const Eigen::GpuDevice& d, double epsilon, int sample_size,
                   int channels, T* variance);
 };
+
+
+// Functor used by FusedBatchNormGradOp to do the computations when is_training=False.
+// Both CPU and GPU will use this functor.
+template <typename Device, typename T>
+struct FusedBatchNormFreezeGrad {
```
zhangyaobit: Could you move this function to fused_batch_norm_op.cc?

tensorflow/python/ops/nn_grad.py
```diff
@@ -809,14 +830,17 @@ def _FusedBatchNormGradGrad(op, *grad):
   """
   data_format = op.get_attr("data_format")
   epsilon = op.get_attr("epsilon")
+  is_training = op.get_attr("is_training")
   grad_y = op.inputs[0]
   x = op.inputs[1]
   scale = op.inputs[2]
+  pop_mean = op.inputs[3]
```
zhangyaobit: I think here you will need the input 3 and 4 of op FusedBatchNorm instead of FusedBatchNormGrad. Could you forward the pop mean and pop variance to the output 3 and 4 of FusedBatchNorm in the C++ code? This way you can also unified the two branches in the "_FusedBatchNormGrad(op, *grad)".

zhangyaobit (commented):
Thanks for the nice contribution, Yuxin!

tensorflow/core/kernels/fused_batch_norm_op.h
ppwwyyxx: If I understand the build process correctly, this functor needs to be instantiated in both fused_batch_norm_op.cc and fused_batch_norm_op.cu.cc, to be compiled to two kernels by nvcc and gcc respectively.  Therefore it has to be in a header file to be included by both .cc and .cu.cc. This seems like what's been done for other kernels as well.

tensorflow/core/kernels/fused_batch_norm_op.h
ppwwyyxx: Seems like `Eigen::Tensor<T, 1, Eigen::RowMajor> scratch1(depth)` only allocate memory on CPUs? In a GPU kernel, using this ends up with `CUDA_ERROR_ILLEGAL_ADDRESS`.

tensorflow/core/kernels/fused_batch_norm_op.h
zhangyaobit: Sounds good.

tensorflow/core/kernels/fused_batch_norm_op.h
zhangyaobit: Thanks, sounds good to me.

tensorflow/core/kernels/fused_batch_norm_op.h
```diff
@@ -43,6 +45,66 @@ struct InvVarianceToVariance {
...
+  void operator()(const Device& d, const Tensor& y_backprop_input,
+                  const Tensor& x_input, const Tensor& scale_input,
+                  const Tensor& mean_input, const Tensor& variance_input,
+                  T epsilon, Tensor* x_backprop_output,
+                  Tensor* scale_backprop_output, Tensor* offset_backprop_output,
+                  typename TTypes<T>::Vec scratch1, typename TTypes<T>::Vec scratch2) {
+    typename TTypes<T, 4>::ConstTensor out_backprop(y_backprop_input.tensor<T, 4>());
```
zhangyaobit: Rename to y_backprop>

tensorflow/core/kernels/fused_batch_norm_op.h
```diff
@@ -43,6 +45,66 @@ struct InvVarianceToVariance {
...
+    Eigen::IndexList<Eigen::type2index<1>, Eigen::Index> one_by_depth;
+    one_by_depth.set(1, depth);
+    Eigen::IndexList<Eigen::type2index<0> > reduction_axis;
+    Eigen::IndexList<Eigen::Index, Eigen::type2index<1> > rest_by_one;
+    rest_by_one.set(0, rest_size);
+#endif
+
+    // db = out_backprop
+    // dg = out_backprop * ((x - m) * rsqrt(v + epsilon))
+    // dx = out_backprop * (gamma * rsqrt(v + epsilon))
```
zhangyaobit: Rename all names to be consistent of what is used in the program?

tensorflow/core/kernels/fused_batch_norm_op.h
```diff
@@ -43,6 +45,66 @@ struct InvVarianceToVariance {
...
+                          (input.reshape(rest_by_depth) -
+                           mean.reshape(one_by_depth).broadcast(rest_by_one)))
+                           .sum(reduction_axis);
+
+    x_backprop.reshape(rest_by_depth).device(d) =
+        out_backprop.reshape(rest_by_depth) * ((scratch1 * scale)
+                                               .eval()
+                                               .reshape(one_by_depth)
+                                               .broadcast(rest_by_one));
+    scale_backprop.device(d) = scratch2 * scratch1;
```
zhangyaobit: Are what implemented above equivalent to python implementation below? 
grad_offset = reduce_sum(grad_y)
grad_scale = reduce_sum(grad_y*(x-pop_mean)*var_rsqrt)
grad_x = grad_y * scale * var_rsqrt

ppwwyyxx: That looks equivalent to me

tensorflow/python/ops/nn_grad.py
ppwwyyxx: `pop_mean` and `pop_var` are inputs to `FusedBatchNormGrad` as well, what's the reason to not use them directly like this?

zhangyaobit: Note that input 3 and 4 are reserve_space_1 and reserve_space_2, which are not pop mean and pop var, but 
reserve_space_1: A 1D Tensor for the computed batch mean, to be reused
                 in the gradient computation.
reserve_space_2: A 1D Tensor for the computed batch variance (inverted variance
                 in the cuDNN case), to be used in the gradient computation.

REGISTER_OP("FusedBatchNormGrad")
    .Input("y_backprop: T")
    .Input("x: T")
    .Input("scale: T")
    .Input("reserve_space_1: T")
    .Input("reserve_space_2: T")
    .Output("x_backprop: T")
    .Output("scale_backprop: T")
    .Output("offset_backprop: T")
    .Output("reserve_space_3: T")
    .Output("reserve_space_4: T")
    .Attr("T: {float}")
    .Attr("epsilon: float = 0.0001")
    .Attr("data_format: string = 'NHWC'")
    .Attr("is_training: bool = true")
......

zhangyaobit: Ok, you did the forwarding in _FusedBatchNormGrad, instead of on the C++ side. I think that is fine too. If you go with this way, could you update the comments of reserve_space_1 and reserve_space_2 saying they are pop mean and variance when is_training is False?

ppwwyyxx: Comments were updated.

tensorflow/core/ops/nn_ops.cc
```diff
@@ -409,10 +409,12 @@ The size of 1D Tensors matches the dimension C of the 4D Tensors.
 y_backprop: A 4D Tensor for the gradient with respect to y.
 x: A 4D Tensor for input data.
 scale: A 1D Tensor for scaling factor, to scale the normalized x.
-reserve_space_1: A 1D Tensor for the computed batch mean, to be reused
+reserve_space_1: A 1D Tensor for the computed batch mean when is_training is True,
```
zhangyaobit: How about something like this?

A 1D Tensor for the computed batch mean when is_training is True, to be reused in the gradient computation; or the population mean when is_training is False, to be used in the second-order gradient computation.

And the same for reserve_space_2.

ppwwyyxx: When is_training is False, pop_mean/pop_variance is needed for first-order gradient computation as well.

zhangyaobit: Ah, ok, then "to be used in the first-order and second-order gradient computation."

tensorflow/core/kernels/fused_batch_norm_op.cc
```diff
@@ -607,14 +623,34 @@ class FusedBatchNormGradOp : public OpKernel {
     OP_REQUIRES_OK(
         context, context->allocate_output(4, TensorShape({}), &placeholder_2));
 
-    functor::FusedBatchNormGrad<Device, T>()(
-        context, y_backprop, x, scale, saved_mean, saved_maybe_inv_var,
-        epsilon_, x_backprop, scale_backprop, offset_backprop, tensor_format_);
+    if (is_training_) {
+      functor::FusedBatchNormGrad<Device, T>()(
+          context, y_backprop, x, scale, saved_mean, saved_maybe_inv_var,
```
zhangyaobit: This is a bit confusing. Rename to saved_mean_or_pop_mean and saved_maybe_inv_var_or_pop_var?

ppwwyyxx: Did some rename (on one existing kernel as well) to improve clarity.

zhangyaobit (commented):
Nice. Thanks!

zhangyaobit:
Let's wait a bit to see if zheng-xq has any comments (This PR may affect lots of users, let's be extra careful :) ). Thanks!

ppwwyyxx:
Any updates?

zhangyaobit:
Thanks for the patience, Yuxin! zheng-xq will respond soon.

drpngx:
/CC @zheng-xq with the `@` sign to trigger notification.

drpngx:
Jenkins, test this please.

tensorflow/core/kernels/fused_batch_norm_op.h
zhangyaobit: Should this be inside functor namespace? Note this test failure: tensorflow/core/kernels/fused_batch_norm_op.cc:645:7: error: 'FusedBatchNormFreezeGrad' is not a member of 'tensorflow::functor'

at https://ci.tensorflow.org/job/tensorflow-pull-requests-cpu-python3/6452/console

ppwwyyxx: Looks like this header is only included when GPU is enabled. I'll fix it.

zhangyaobit:
Jenkins, test this please.

tensorflow/core/kernels/fused_batch_norm_op.h
ppwwyyxx: Ah this is more complicated than I thought. Looks like I need to change BUILD file somehow.

drpngx:
Jenkins, test this please.

zhangyaobit (approved):

zhangyaobit:
Please merge this change. Thanks!

PR #12347, opened 2017-08-17
Clean up 'shell=True' in using subprocess module

wangqr:
Currently, `configure.py` always use the running python interpreter to find the site-packages path. When user has specified a different python in the previous step, the specified one should be used.

PR #12582, opened 2017-08-25
Double check after regex search for cudnn_path in configure.py

ScorpioCPH:
If chose wrong `cuDNN` version on `linux`, `configure.py` will throw an exception as shown below:

```shell
Please specify the location where cuDNN 6 library is installed. Refer to README.md for more details. [Default is /usr/local/cuda]:

  File "configure.py", line 1021, in <module>
    main()
  File "configure.py", line 997, in main
    set_tf_cunn_version(environ_cp)
  File "configure.py", line 688, in set_tf_cunn_version
    cudnn_path_from_ldconfig).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```

This is because these code return `None`:

```python
cudnn_path_from_ldconfig = re.search('.*libcudnn.so .* => (.*)',
                                     cudnn_path_from_ldconfig)
```

This PR add double check after regex search.

Related PR: https://github.com/tensorflow/tensorflow/pull/12347 

@wangqr @yifeif PTAL.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -685,7 +685,9 @@ def set_tf_cunn_version(environ_cp):
       ldconfig_bin = which('ldconfig') or '/sbin/ldconfig'
       cudnn_path_from_ldconfig = run_shell([ldconfig_bin, '-p'])
       cudnn_path_from_ldconfig = re.search('.*libcudnn.so .* => (.*)',
-                                           cudnn_path_from_ldconfig).group(1)
+                                           cudnn_path_from_ldconfig)
+      if cudnn_path_from_ldconfig != None:
+        cudnn_path_from_ldconfig = cudnn_path_from_ldconfig.group(1)
       if os.path.exists('%s.%s' % (cudnn_path_from_ldconfig, tf_cudnn_version)):
         cudnn_install_path = os.path.dirname(cudnn_path_from_ldconfig)
         break
--- a/configure.py
+++ b/configure.py
@@ -686,11 +686,11 @@ def set_tf_cunn_version(environ_cp):
       cudnn_path_from_ldconfig = run_shell([ldconfig_bin, '-p'])
       cudnn_path_from_ldconfig = re.search('.*libcudnn.so .* => (.*)',
                                            cudnn_path_from_ldconfig)
-      if cudnn_path_from_ldconfig != None:
+      if cudnn_path_from_ldconfig:
         cudnn_path_from_ldconfig = cudnn_path_from_ldconfig.group(1)
-      if os.path.exists('%s.%s' % (cudnn_path_from_ldconfig, tf_cudnn_version)):
-        cudnn_install_path = os.path.dirname(cudnn_path_from_ldconfig)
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -685,7 +685,9 @@ def set_tf_cunn_version(environ_cp):
       ldconfig_bin = which('ldconfig') or '/sbin/ldconfig'
       cudnn_path_from_ldconfig = run_shell([ldconfig_bin, '-p'])
       cudnn_path_from_ldconfig = re.search('.*libcudnn.so .* => (.*)',
-                                           cudnn_path_from_ldconfig).group(1)
+                                           cudnn_path_from_ldconfig)
+      if cudnn_path_from_ldconfig != None:
```
yifeif: [nit]: remove " != None"

ScorpioCPH: @yifeif Sorry, I can't see the comments, could you have a check please?

wangqr:
Maybe the next [if block](https://github.com/ScorpioCPH/tensorflow/blob/ab635bcafaa42f599e49ceb6ed324e32996d8b45/configure.py#L691-L693) should only run when `cudnn_path_from_ldconfig` is not None, since some file named `None.6.0` is not what we want

ScorpioCPH:
@wangqr Yes, but before this block, exception has been thrown out on calling `.group(1)`

wangqr:
Yes I cam confirm that we should check for the None value returned by `search`, and your commit does the check. I just suggest that maybe moving the following if block into the if block you've created is more proper.

wangqr:
From
```python
      if cudnn_path_from_ldconfig != None:
        cudnn_path_from_ldconfig = cudnn_path_from_ldconfig.group(1)
      if os.path.exists('%s.%s' % (cudnn_path_from_ldconfig, tf_cudnn_version)):
        cudnn_install_path = os.path.dirname(cudnn_path_from_ldconfig)
        break
```
to
```python
      if cudnn_path_from_ldconfig != None:
        cudnn_path_from_ldconfig = cudnn_path_from_ldconfig.group(1)
        if os.path.exists('%s.%s' % (cudnn_path_from_ldconfig, tf_cudnn_version)):
          cudnn_install_path = os.path.dirname(cudnn_path_from_ldconfig)
          break
```

ScorpioCPH:
@wangqr PR updated, could you take another look, please?

wangqr (approved):

wangqr:
@ScorpioCPH I don't have write access to this repo...

configure.py
yifeif: Could you remove the "!= None"?

yifeif:
@ScorpioCPH let me know if you still cannot see the review comment.

ScorpioCPH:
@yifeif I see your comment now, thanks! And fix it by amending last commit. PTAL.

yifeif (approved):

yifeif:

yifeif:

PR #12594, opened 2017-08-25
fix: profiler bazel build usage

csarron:


tensorflow/core/profiler/g3doc/command_line.md
```diff
@@ -57,7 +57,7 @@ Note: this feature is not well maintained now.
 
 ```shell
 # Build the tool.
-bazel build --config opt third_party/tensorflow/core/profiler/...
+bazel build --config opt tensorflow/core/profiler:profiler
```
jhseu: Leave the ... part?

csarron: Not sure about what the `...` mean, maybe should be kept.

jhseu (approved):

PR #12602, opened 2017-08-25
Updating install_linux.md for cuDNN 6.

av8ramit:
@gunan does line 59 need to be updated as well?


wolffg (approved):

gunan:
Not that line, but let's convert line 58 to say CUDA 8.0 or greater

Issue #10838, opened 2017-06-19
Compiling TensorFlow gives 3k lines of warnings!


Iolaum:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04.2 (4.8 kernel)
- **TensorFlow installed from (source or binary)**: Compiling TF from Source
- **TensorFlow version (use command below)**: 1.2.0
- **Bazel version (if compiling from source)**: 0.5.1
- **CUDA/cuDNN version**: CUDA 8.0, CuDNN 5.1
- **GPU model and memory**: 1050Ti 4GB (Notebook version) (Intel i5-7300hq CPU)
- **Exact command to reproduce**: Compiling from source [following documentation](https://www.tensorflow.org/install/install_sources).

### Describe the problem
I compiled tensorflow from source. The process finished successfully and the binary managed to install and run successfully. What seems strange is that during the process I got ~3k lines of warnings. I am linking to them at the end of the issue. I am wondering if that's expected behavior or indication of a (small or _not_so_small_?) problem.

One thing that may affect this is bazel installation. I followed [Bazel Installation Instructions](https://bazel.build/versions/master/docs/install.html) and used the recommended apt method. This led to me running into [this](https://github.com/tensorflow/tensorflow/issues/8092) issue. Installing openjdk-8-jdk on top of the ibm-java80-jdk as suggested in a [comment](https://github.com/tensorflow/tensorflow/issues/8092#issuecomment-304957009) solves the problem (although I am not sure how much technical debt this solution caries, which may have manifested in some of the warnings produced during compilation).

### Source code / logs
**Configuration Script options:**
```{shell}
$ ./configure
Please specify the location of python. [Default is /usr/bin/python]: /usr/bin/python3
Found possible Python library paths:
  /usr/local/lib/python3.5/dist-packages
  /usr/lib/python3/dist-packages
Please input the desired Python library path to use.  Default is [/usr/local/lib/python3.5/dist-packages]
/usr/local/lib/python3.5/dist-packages
Do you wish to build TensorFlow with MKL support? [y/N] y
MKL support will be enabled for TensorFlow
Do you wish to download MKL LIB from the web? [Y/n] y
Please specify optimization flags to use during compilation when bazel option "--config=opt" is specified [Default is -march=native]: 
[...]

PR #12606, opened 2017-08-25
Add `-Wno-write-strings` to remove the warning in external protobuf compilation

yongtang:
This fix is an effort based on the discussion https://github.com/tensorflow/tensorflow/issues/10838#issuecomment-311420644

Currently a big chunk (200+) of the warnings when compiling tensorflow are caused by `-Wwrite-strings` in external protobuf package:
```
external/protobuf_archive/python/google/protobuf/pyext/message_factory.cc:78:37: warning: ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]
...
```

As protobuf is external, it makes sense to add the flag of `-Wno-write-strings` to reduce the number of warnings in the output.

This fix reduces 200+ warnings.


jhseu:
Please suggest this change directly to the protobuf repository instead. We want to minimize the amount of patching we have to do to external repos.

PR #12607, opened 2017-08-26
First base version of snapcraft.yaml. Work in progress. [DO NOT MERGE]

av8ramit:
I'm following the instructions from https://www.tensorflow.org/install/install_sources

I'm using this as a tutorial. https://snapcraft.io/docs/build-snaps/python

### DIFF
--- a/tensorflow/tools/ci_build/snap/snapcraft.yaml
+++ b/tensorflow/tools/ci_build/snap/snapcraft.yaml
@@ -1,0 +1,34 @@
+name: tensorflow
+version: 'v1.3.0'
+summary: An open-source software library for Machine Intelligence
+description: |
+  TensorFlow™ is an open source software library or numerical computation
+  using data flow graphs. Nodes in the graph represent mathematical operations,
+  while the graph edges represent the multidimensional data arrays (tensors)
+  communicated between them. The flexible architecture allows you to deploy
+  computation to one or more CPUs or GPUs in a desktop, server, or mobile device
+  with a single API. TensorFlow was originally developed by researchers and
+  engineers working on the Google Brain Team within Google's Machine Intelligence
+  research organization for the purposes of conducting machine learning and deep
+  neural networks research, but the system is general enough to be applicable in
+  a wide variety of other domains as well.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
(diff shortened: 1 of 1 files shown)


tensorflow/tools/ci_build/snap/snapcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+    build-packages:
+      - openjdk-8-jdk
+      - oracle-java8-installer
+      - bazel
+    build:
+      pip install six numpy wheel 
+      cd tensorflow
+      ./configure
+      bazel build --config=opt //tensorflow/tools/pip_package:build_pip_package
+      pip install /tmp/tensorflow_pkg/tensorflow-1.3.0-py2-none-any.whl
```
saeta: Newline nit.

Note: is there any way to avoid burning in the version into this file?

saeta (approved):
This looks good to me, although I'm probably not the best reviewer.

Thanks for doing this!!

tensorflow/tools/ci_build/snap/snapcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+confinement: devmode # use 'strict' once you have the right plugs and slots
+
+parts:
+  tensorflow:
+    source: https://github.com/tensorflow/tensorflow/archive/$SNAPCRAFT_PROJECT_VERSION.tar.gz
+    plugin: python
+    python-version: python2
+
+    build-packages:
+      - openjdk-8-jdk
```
gunan: Also, needed here. Swig, gcc, python-pip, python

gunan: Also add these:
python-dev (instead of python)
git (needed during configure, https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/git/gen_git_source.py#L155)

tensorflow/tools/ci_build/snap/snapcraft.yaml
av8ramit: So unfortunately no, I'll have to hardcode the version, but what I can do is modify the update_version script to update it during releases.

av8ramit:
https://github.com/av8ramit/tensorflow-snap

PR #12615, opened 2017-08-26
Update head.py

alanyee:
-Replace contrib framework with core framework
-Replace contrib model_fn with core model_fn


ispirmustafa (changes requested):
core model_fn and contrib ones are incompatible. We'll not update the contrib ones to not to break existing users.
Let's keep them as it is.

PR #12620, opened 2017-08-26
[XLA] Add a check to the HLO verifier for badly formatted Broadcasts.

DavidNorman:
This adds a check to the HloVerifier which ensures that the Hlo Broadcast instructions are formatted correctly.  Since they are different from the UserComputation broadcast instructions (and therefore the XLA docs), examples of invalid instructions were appearing in the repo.

Also fix up the two cases where the code was creating invalid broadcasts.


meheffernan (approved):
Thanks for the fix!

jpienaar:
Jenkins, test this please

DavidNorman:
@jpienaar hi - seems like the code review thing has not run.  is that an automated thing?

martinwicke:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

DavidNorman:
Ah - this Linux XLA error seems to be a memory/disk space issue with the docker builder.

jpienaar:
Jenkins, test this please

DavidNorman:
Hmm.  I'm not really sure I understand the problem here.  it appears to be another build system error, but two in a row seems suspicious.

is this happening to other CI builds too, or just this one?

should I be merging in the head of master?

DavidNorman:
The 'Linux CPU Tests (Python 3)' failure is definitely not this PR - it contains only XLA changes, and the failure is something about a gRPC server.

the XLA one is harder to judge - is it a git merge problem?  or just an infrastructure problem?

jpienaar:
It is weird, I've tried prodding it too. And I do see passing cases post this. I've set off another round of building. And will ask around more if that doesn't work.

jpienaar:
Jenkins, test this please.

jpienaar (approved):
Don't know why it says awaiting a review as Mark already did, but re-approving.

martinwicke:
@gunan, can we change "TF Code Review" to "TF Test Suite" to avoid confusion?

Jenkins, test this please.

gunan:
Renaming of the build complete.

Issue #12431, opened 2017-08-20
[android demo] UnsatisfiedLinkError exception on app start if build with nativeBuildSystem = 'none'


ArtsiomCh:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
NO
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
app build at Android Studio 2.3.3 on Linux Ubuntu 16.04
app tested on Meizu M2 Note (Android 5.1, API 22) 
- **TensorFlow installed from (source or binary)**:
Source
- **TensorFlow version (use command below)**:
Master branch from GitHub
- **Python version**: 
2.7 (not used in that case)
- **Bazel version (if compiling from source)**:
0.5.3 (not used in that case) 
- **CUDA/cuDNN version**:
none
- **GPU model and memory**:
none
- **Exact command to reproduce**:
Android Studio/Run (build.gradle)

### Describe the problem
With `nativeBuildSystem = 'none'` app exit with UnsatisfiedLinkError exception (see log below).
Quick investigation shows call of `env.ImageUtils.convertYUV420SPToARGB8888(Native Method)` native method from `CameraActivity.onPreviewFrame(CameraActivity.java:113)`. 
`convertYUV420SPToARGB8888` suppose to be implemented in `libtensorflow_demo.so` wich is not created in case of `nativeBuildSystem = 'none'`. (With `nativeBuildSystem = 'cmake'` `libtensorflow_demo.so` created and `convertYUV420SPToARGB8888` works fine)
Quick fix may be obtained by including `libtensorflow_demo.so` into JCenter and download it like the TensorFlow Inference Interface package.
Another solution would be Java implementation of `convertYUV420SPToARGB8888`, the same as it been done for `convertYUV420ToARGB8888` in commit #003deb8 at PR #10771 "Refactor and implementation of the camera API 1, it fixes #8736 #10771".
I start working with that Java implementation of `convertYUV420SPToARGB8888` but I don't think I have enough skills to make it. At the moment I create Java wrapper for native convertYUV420SPToARGB8888 and make all native methods `private` to isolate the rest of the Java code from occasional use them:
[https://github.com/ArtsiomCh/tensorflow/blob/master/tensorflow/examples/android/src/org/tensorflow/demo/CameraActivity.java](url)
[...]

PR #12622, opened 2017-08-26
[android demo] fix issue #12431 Java implementation of YUV420SP to ARGB8888 converting

ArtsiomCh:
This PR fix issue #12431
Java implementation of `ImageUtils.convertYUV420SPToARGB8888` made and refactoring for Java implementation of `ImageUtils.convertYUV420ToARGB8888` made as well.

Despite the fact that code looks a bit 'nuddle-style' that was made for efficiency reasons. Breaking down it into few, more readable functions (like it was made before at `ImageUtils.convertYUV420ToARGB8888`) lead into 3-time function execution increase:
Previous implementation of `ImageUtils.convertYUV420ToARGB8888` cause invocation inside double nested circles 307200 times (`640x480`) of `YUV2RGB` function and three times more `convertByteToInt` with `Math.min` and `Math.max` functions. 

Total costs of classification at TF Classify was around 3000 ms on my test device (Meizu N2, API 22). Proposed implementation drops it down to 1000-1100 ms (which is comparable with native C implementation from `libtensorflow_demo.so`  ~700 ms).

### DIFF
--- a/tensorflow/examples/android/src/org/tensorflow/demo/env/ImageUtils.java
+++ b/tensorflow/examples/android/src/org/tensorflow/demo/env/ImageUtils.java
@@ -98,6 +98,60 @@ public class ImageUtils {
   // Always prefer the native implementation if available.
   private static boolean useNativeConversion = true;
 
+  public static void convertYUV420SPToARGB8888(
+      byte[] input,
+      int[] output,
+      int width,
+      int height) {
+
+    if (useNativeConversion) {
+      try {
+        ImageUtils.convertYUV420SPToARGB8888(input, output, width, height, false);
+        return;
+      } catch (UnsatisfiedLinkError e) {
+        LOGGER.w("Native YUV420SP -> RGB implementation not found, falling back to Java implementation");
+        useNativeConversion = false;
+      }
+    }
+
+    // Java implementation of YUV420SP to ARGB8888 converting
+    final int frameSize = width * height;
+    for (int j = 0, yp = 0; j < height; j++) {
+      int uvp = frameSize + (j >> 1) * width, u = 0, v = 0;
+      for (int i = 0; i < width; i++, yp++) {
+        int y = (0xff & ((int) input[yp])) - 16;
+        if (y < 0)
+          y = 0;
+        if ((i & 1) == 0) {
+          v = (0xff & input[uvp++]) - 128;
+          u = (0xff & input[uvp++]) - 128;
+        }
+
+        int y1192 = 1192 * y;
+        int r = (y1192 + 1634 * v);
+        int g = (y1192 - 833 * v - 400 * u);
+        int b = (y1192 + 2066 * u);
+
+        if (r < 0)
+          r = 0;
+        else if (r > kMaxChannelValue)
+          r = kMaxChannelValue;
+        if (g < 0)
+          g = 0;
+        else if (g > kMaxChannelValue)
+          g = kMaxChannelValue;
+        if (b < 0)
+          b = 0;
+        else if (b > kMaxChannelValue)
+          b = kMaxChannelValue;
+
+        output[yp] = 0xff000000 | ((r << 6) & 0xff0000)
+                | ((g >> 2) & 0xff00) | ((b >> 10) & 0xff);
+
+      }
+    }
+  }
+
   public static void convertYUV420ToARGB8888(
       byte[] yData,
       byte[] uData,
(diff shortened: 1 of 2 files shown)


terrylewis60:
Not interested forward to police

tensorflow/examples/android/src/org/tensorflow/demo/env/ImageUtils.java
```diff
@@ -98,6 +98,60 @@ public static void saveBitmap(final Bitmap bitmap, final String filename) {
   // Always prefer the native implementation if available.
   private static boolean useNativeConversion = true;
 
+  public static void convertYUV420SPToARGB8888(
+      byte[] input,
+      int[] output,
```
andrewharp: for maximum parallelism with `convertYUV420ToARGB8888`, `output` should be the last argument

tensorflow/examples/android/src/org/tensorflow/demo/env/ImageUtils.java
```diff
@@ -122,48 +176,47 @@ public static void convertYUV420ToARGB8888(
...
+        if (nY < 0)
+          nY = 0;
+
+        // This is the floating point equivalent. We do the conversion in integer
+        // because some Android devices do not have floating point in hardware.
+        // nR = (int)(1.164 * nY + 2.018 * nU);
+        // nG = (int)(1.164 * nY - 0.813 * nV - 0.391 * nU);
+        // nB = (int)(1.164 * nY + 1.596 * nV);
+
+        final int foo = 1192 * nY;
```
andrewharp: s/foo/y1192/

tensorflow/examples/android/src/org/tensorflow/demo/env/ImageUtils.java
```diff
@@ -98,6 +98,60 @@ public static void saveBitmap(final Bitmap bitmap, final String filename) {
...
+      for (int i = 0; i < width; i++, yp++) {
+        int y = (0xff & ((int) input[yp])) - 16;
+        if (y < 0)
+          y = 0;
+        if ((i & 1) == 0) {
+          v = (0xff & input[uvp++]) - 128;
+          u = (0xff & input[uvp++]) - 128;
+        }
+
+        int y1192 = 1192 * y;
```
andrewharp: It seems like everything from here to the bottom of the loop can be shared between both conversion functions in a common helper method. The only difference between the two conversion types should be how it initializes the y,u,v variables per output pixel.

tensorflow/examples/android/src/org/tensorflow/demo/env/ImageUtils.java
```diff
@@ -98,6 +98,60 @@ public static void saveBitmap(final Bitmap bitmap, final String filename) {
...
+        int r = (y1192 + 1634 * v);
+        int g = (y1192 - 833 * v - 400 * u);
+        int b = (y1192 + 2066 * u);
+
+        if (r < 0)
```
andrewharp: Assuming the speed is equivalent, would prefer ternary assignment here, e.g.:

r = r > kMaxChannelValue ? kMaxChannelValue : (r < 0 ? 0 : r);

or possibly:

r = Math.max(0, Math.min(r, kMaxChannelValue));

just for readability purposes.

Otherwise please add {} blocks to each if/else clause

ArtsiomCh:
@andrewharp Please check the corrected version of ImageUtils.java
I tried to find a compromise between readability of the code and performance. 
Unfortunately, there is no way to declare an in-line function in java. So using any additional function inside our double nested loops will create the overhead of making a function call. 
As I mentioned before the original java implementation of `convertYUV420ToARGB8888` function took 3000 ms on my test device to show a result. After making Method Tracing I've found that majority of the time being spent on calling helping functions: `convertByteToInt` with `Math.min` and `Math.max`. And the actual classification work by TF took just ~500ms.
My original "all-in-one" implementation decrease the total time between showing result from 3000ms to 1000ms. Even adding YUV2RGB function (for readability and avoiding duplicating the code) already increase the total time from ~1000 ms to ~1100. I think this is acceptable, as it just 10%.
Please feel free to ask for any additional correctness.

PS Looks like GitHub shows a bit mess in `Split` mode. It replaced `convertYUV420ToARGB8888` with `convertYUV420SPToARGB8888` code. And create new `convertYUV420ToARGB8888` instead of correct existing implementation. Any case resulting code looks clean enough.

andrewharp:
Thanks, I think this is a good compromise. IIRC on some devices Java already performed nearly equivalently to the native conversion, so I think you might be seeing one of the more extreme situations.

andrewharp (approved):

andrewharp (approved):

Issue #12612, opened 2017-08-26
Eager execution with string tensors.


eaplatanios:
@alextp String tensors seem to cause lots of errors for me, when using the eager execution API, that do not occur when using normal ops and executing them using a session.

The way I'm creating string tensors is by allocating a `TF_Tensor` with the right data type, shape, and size (where I use `TF_StringEncodedSize` to get the byte sizes of each element). Then, I encode the strings using `TF_StringEncode` and write them in the tensor buffer along with the offset table, as described [here](https://github.com/tensorflow/tensorflow/blob/752dcb61ef7a8fd6555909dc37c1f2a2e5792227/tensorflow/c/c_api.h#L211). Afterwards, I call `TFE_NewTensorHandle` to get an eager tensor handle and I delete the `TF_Tensor` object. This works fine for normal op construction and execution using sessions, but it fails for eager execution. Often I get a `check failed: IsAligned()` error from `tensor.cc`, but other times the whole application crashes, such as in the example that follows:

```scala
val t = Tensor.fill(STRING, Shape())("test")  // The generated tensor is good. 
                                              // I can access its elements and use it 
                                              // with session execution.
val tt = tfe.stack(Seq(t, t))  // I can no longer access the tensor elements correctly 
                               // and when I try to slice it I get either errors (crashing) 
                               // or empty strings.
tt.unstack(2)                  // Crashes with the error shown below.
```

This is run using my Scala API but I hope that the error I get may help resolve it. In the error stacktrace I get this:

```
Stack: [0x0000700011571000,0x0000700011671000],  sp=0x000070001166cbf0,  free space=1006k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libsystem_platform.dylib+0x15da]  _platform_memmove$VARIANT$Nehalem+0x1da
C  [libc++.1.dylib+0x3af29]  _ZNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEC2ERKS5_+0x8d
[...]

PR #12626, opened 2017-08-27
Eager API String Tensors (and others that need to be copied) Fix

eaplatanios:
This PR partially fixes the issue related to string tensors, described in #12612. Note that there are still some issues remaining related to string tensors:
1. There is a small inconsistency when dealing with TFE tensors now in that for the string data type, a copy of the provided `TF_Tensor` is created, whereas for others types that's not the case. This can be problematic when users of the API attempt to use the underlying tensor buffer directly (e.g., to read values of specific elements). The problem could potentially be alleviated if the same memory layout is used for string `TF_Tensor`s and `tensorflow::Tensor`s. I'm not sure why that's not the case and I'm also not sure what the memory layout for string tensors is internally (could someone provide some insight into that please?).
2. The above-mentioned problem could be partially alleviated by providing a TFE function for obtaining a byte array representation of the `i`th element in the flattened (row-major) tensor. That is currently the only use I have found for directly accessing the underlying buffer.
3. This fix resolves the issues specific to string tensors and the eager execution API, but one issue related to #12612 still remains. When I try to use some ops, such as the `Unpack` op, I get the following error (**no matter what the data type of the tensor is**): `C  [libtensorflow.so+0x20f0]  TFE_TensorHandleDeviceName+0x0`. I plan to look more into this, but please let me know if you can see a reason why this might be happening.

@alextp could you please give me some information on what the memory layout is for string tensors internally and on why we cannot use the same representation for the C API `TF_Tensor`s? That would resolve points 1 and 2 above. I think that a function for obtaining the `i`th element would still be very useful though, in either case. :)


eaplatanios:
For some reason the commits here are mixed with some I made for another PR. Sorry about that. I'll look into resolving it.

eaplatanios:
I'll just make another PR.

PR #12629, opened 2017-08-27
Update feature_column_ops.py

alanyee:
Using get_variables over model_variable


ispirmustafa (changes requested):
model_variable adds variable into collection ops.GraphKeys.MODEL_VARIABLES. THere may be some users depend on it. let's keep them as it is.

alanyee:
Ah yeah, I have noticed that. I was going to check if there was a way to add them into both graphs.

yifeif:
I'm closing this for now. @alanyee feel free to reopen when you have an update. Thanks~

PR #12631, opened 2017-08-27
Make Dataset.map not unpack namedtuple

adaitche:
The map method of tf.contrib.data.Dataset unpacks iterables before passing them to the mapped function. This is quite convenient. However, currently this also happens for namedtuples, which IMHO is not what one would expect. Consider, for example, the following code:

```python
import tensorflow as tf
from tensorflow.contrib.data import Dataset
from collections import namedtuple

Example = namedtuple("Example", ["label", "attributes", "bounding_box", "image"])


def read_example(serialized_example):
    # parse example
    label = ...
    attribute = ...
    bounding_box = ...
    image = ...
    
    return Example(label, attribute, bounding_box, image)


def preprocess_example(example):
    image = some_function(example.image, example.bounding_box)
    return example._replace(image=image)


dataset = tf.contrib.data.TFRecordDataset(filename)
dataset = dataset.map(read_example)
dataset = dataset.map(preprocess_example)
```

This does not work because the namedtuple instance is unpacked and its fields are passed as separate arguments to `preprocess_example`. So currently one has to reassemble the namedtuple manually. This is not a big deal, but it would be more convenient if namedtuples were not unpacked, as is also the case for dictionaries.

This can be easily achieved by making `_should_unpack_args` return False for namedtuples, as is done in this pull request.

### DIFF
--- a/tensorflow/contrib/data/python/ops/dataset_ops.py
+++ b/tensorflow/contrib/data/python/ops/dataset_ops.py
@@ -1879,8 +1879,7 @@ class DenseToSparseBatchDataset(Dataset):
 
 def _should_unpack_args(args):
   """Returns `True` if `args` should be `*args` when passed to a callable."""
-  is_namedtuple = isinstance(args, tuple) and hasattr(args, "_fields")
-  return nest.is_sequence(args) and not isinstance(args, dict) and not is_namedtuple
+  return type(args) is tuple  # pylint: disable=unidiomatic-typecheck
 
 
 class _ResourceDataset(Dataset):
--- a/tensorflow/contrib/data/python/ops/dataset_ops.py
+++ b/tensorflow/contrib/data/python/ops/dataset_ops.py
@@ -2126,7 +2126,7 @@ class InterleaveDataset(Dataset):
 
       nested_args = nest.pack_sequence_as(input_dataset.output_types, args)
 
-      if nest.is_sequence(nested_args):
+      if _should_unpack_args(nested_args):
         dataset = map_func(*nested_args)
       else:
         dataset = map_func(nested_args)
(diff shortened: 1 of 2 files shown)


adaitche:
I signed it!

tensorflow/contrib/data/python/ops/dataset_ops.py
```diff
@@ -1879,7 +1879,8 @@ def output_types(self):
 
 def _should_unpack_args(args):
   """Returns `True` if `args` should be `*args` when passed to a callable."""
-  return nest.is_sequence(args) and not isinstance(args, dict)
+  is_namedtuple = isinstance(args, tuple) and hasattr(args, "_fields")
+  return nest.is_sequence(args) and not isinstance(args, dict) and not is_namedtuple
```
mrry: Perhaps it would be simpler to replace this with `return type(args) is tuple`? (We might need a pylint exclusion to make the sanity check presubmit happy, but I think it would be justified in this case.)

mrry (changes requested):
Thanks for your contribution, @adaitche! I'm persuaded by your reasoning, and while this code is still in `tf.contrib` it's a bit easier to break existing users. (Also, I've done a quick search and haven't been able to find any code that depends on this behavior.)

Can you please add some tests to `tensorflow/contrib/data/python/kernel_tests/map_dataset_op_test.py` to ensure that your example program from the PR description has the right behavior?

adaitche:
Added a test case and simplified as suggested to `return type(args) is tuple`. This is indeed simpler. And also more explicit, i.e. now it is clear that only tuples will be unpacked. Before it was everything what is a Sequence but not a list or dict, which for example included a deque.

@mrry I am wondering why in InterleaveDataset on line 2129 `nest.is_sequence` is used in contrast to `_should_unpack_args`. This means that e.g. dictionaries will be unpacked. Is this the intended behaviour?

mrry:
Thanks for adding the test! Code looks great now.

As for `InterleaveDataset`, I think that's a bug (probably my oversight when I added it...): we should be consistent across the various function-taking transformations, so it would be great if you could change it to use `_should_unpack_args()` as well.

adaitche:
Changed InterleaveDataset to use _should_unpack_args.

mrry (approved):

mrry:
Awesome, thanks!

PR #12651, opened 2017-08-28
R1.3 - Delete monitors.md

MarkDaoust:
The monitors in this doc are not compatible with the core `Estimators` in the other get started docs.

This will likely only cause confusion. Since AFAIK monitors are deprecated I see no advantage to preserving this doc in `api_guides`.

If we delete it here I can regenerate the root version of the website without this file.

I'll mirror these changes into master.


ispirmustafa (approved):
Thank you Mark

scotthuang1989:
I think this is why this [webpage](https://www.tensorflow.org/get_started/monitors) show 404 error?
I just wrote a program that user "ValidationMonitor " to implement early stopping, even though this class is in contrib  but it still works fine, why delete the documentation?

martinwicke:
The class is deprecated. We would like to not encourage its use.

scotthuang1989:
@martinwicke 
 I didn't see any sign or comment said that the class is deprecated(both master and r1.3 branch).  Is there some place that I can check which API is going to deprecated? 
Another question, What's the new API to support early stopping?

martinwicke:
Whenever you instantiate a ValidationMonitor, a deprecation warning is
logged, with the text ​

"Monitors are deprecated. Please use tf.train.SessionRunHook."

You can use SessionRunHooks to achieve the same thing, although we have
none that does it. The reason for that is that this implementation has
issues, and we want to provide a better one, which we're working on.

PR #12652, opened 2017-08-28
delete `get_started/monitors.md`

MarkDaoust:
This is a copy of  https://github.com/tensorflow/tensorflow/pull/12651 cherrypicked onto master. 

The monitors in this doc are not compatible with the core Estimators in the other get started docs.

This will likely only cause confusion. Since AFAIK monitors are deprecated I see no advantage to preserving this doc in api_guides.

If we delete it here I can regenerate the root version of the website without this file.

I'll mirror these changes into master.


ispirmustafa (approved):
Thank you Mark

martinwicke:
Jenkins, test this please.

PR #12658, opened 2017-08-28
Support CopyFile with streaming

### LINKED ISSUE #12641: Improve all-in-memory file copy architecture (Python at least)
Current file copy (at least via Python `tf.gfile.Copy` ([gfile.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/platform/gfile.py#L22)→ [file_io.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/lib/io/file_io.py#L371) → [file_io.i](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/lib/io/file_io.i#L113))) involves copying the source contents into memory, and then writing memory to the destination. For scenarios like https://github.com/tensorflow/tensorflow/issues/12630 which is working with an 11GB asset, this is unacceptable design.

`file_system.h`'s `WritableFile` is not stubbed to allow anything like a streaming, though its `RandomAccessFile` is.

To cull the Python low hanging fruit, at least, please implement [file_io.i](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/lib/io/file_io.i#L113) using a regular streaming design instead of the above described current design.

yongtang:
This fix tries to address the issue raised in #12641 where it was not possible to have CopyFile with streaming. The original implementation copies the whole content of the file to a string
buffer and write to the file. This could be an issue if the file size is too large (than the memory of the host).

This fix streams the CopyFile operation.

*Also, sendfile is used if the file system is posix*

This fix fixes #12641.

### DIFF
--- a/tensorflow/core/platform/env.cc
+++ b/tensorflow/core/platform/env.cc
@@ -40,6 +40,8 @@ limitations under the License.
 
 namespace tensorflow {
 
+constexpr size_t kCopyFileBufferSize = 1024 * 1024;
+
 class FileSystemRegistryImpl : public FileSystemRegistry {
  public:
   Status Register(const string& scheme, Factory factory) override;
@@ -359,6 +361,57 @@ Status WriteStringToFile(Env* env, const string& fname,
     s = file->Close();
   }
   return s;
+}
+
+Status CopyFile(Env* env, const string& oldpath, const string& newpath,
+                bool overwrite) {
+  // If overwrite is false and the newpath file exists then it's an error.
+  if (!overwrite && env->FileExists(newpath).ok()) {
+    return errors::AlreadyExists("file already exists");
+  }
+  uint64 file_size;
+  Status s = env->GetFileSize(oldpath, &file_size);
+  if (!s.ok()) {
+    return s;
+  }
+  std::unique_ptr<RandomAccessFile> oldfile;
+  s = env->NewRandomAccessFile(oldpath, &oldfile);
+  if (!s.ok()) {
+    return s;
+  }
+  std::unique_ptr<WritableFile> newfile;
+  s = env->NewWritableFile(newpath, &newfile);
+  if (!s.ok()) {
+    return s;
+  }
+
+  uint64 offset = 0;
+  while (offset < file_size) {
+    char scratch[kCopyFileBufferSize];
+    StringPiece result;
+    size_t chunk = offset + sizeof(scratch) < file_size ? sizeof(scratch)
+                                                        : file_size - offset;
+    s = oldfile->Read(offset, chunk, &result, scratch);
+    if (!s.ok()) {
+      return s;
+    }
+
+    s = newfile->Append(result);
+    if (!s.ok()) {
+      return s;
+    }
+
+    offset += chunk;
+  }
+  s = env->GetFileSize(oldpath, &file_size);
+  if (!s.ok()) {
+    return s;
+  }
+  if (offset != file_size) {
+    return errors::Aborted("File ", oldpath, " changed while reading: ",
+                           file_size, " vs. ", offset);
+  }
+  return newfile->Close();
 }
 
 // A ZeroCopyInputStream on a RandomAccessFile.
--- a/tensorflow/core/platform/env.cc
+++ b/tensorflow/core/platform/env.cc
@@ -374,6 +374,32 @@ Status WriteStringToFile(Env* env, const string& fname,
   return s;
 }
 
+Status CopyFile(FileSystem* src_fs, const string& src, FileSystem* target_fs,
+                const string& target) {
+  uint64 size;
+  TF_RETURN_IF_ERROR(src_fs->GetFileSize(src, &size));
+
+  std::unique_ptr<RandomAccessFile> src_file;
+  TF_RETURN_IF_ERROR(src_fs->NewRandomAccessFile(src, &src_file));
+
+  std::unique_ptr<WritableFile> target_file;
+  TF_RETURN_IF_ERROR(target_fs->NewWritableFile(target, &target_file));
+
+  uint64 offset = 0;
+  while (offset < size) {
+    char scratch[kCopyFileBufferSize];
+    StringPiece result;
+    size_t bytes_to_read =
+        offset + sizeof(scratch) < size ? sizeof(scratch) : size - offset;
+    TF_RETURN_IF_ERROR(src_file->Read(offset, bytes_to_read, &result, scratch));
+
+    TF_RETURN_IF_ERROR(target_file->Append(result));
+
+    offset += bytes_to_read;
+  }
+  return target_file->Close();
+}
+
 // A ZeroCopyInputStream on a RandomAccessFile.
 namespace {
 class FileStream : public ::tensorflow::protobuf::io::ZeroCopyInputStream {
@@ -411,32 +437,6 @@ class FileStream : public ::tensorflow::protobuf::io::ZeroCopyInputStream {
 };
 
 }  // namespace
-
-Status CopyFile(FileSystem* src_fs, const string& src, FileSystem* target_fs,
-                const string& target) {
-  uint64 size;
-  TF_RETURN_IF_ERROR(src_fs->GetFileSize(src, &size));
-
-  std::unique_ptr<RandomAccessFile> src_file;
-  TF_RETURN_IF_ERROR(src_fs->NewRandomAccessFile(src, &src_file));
-
-  std::unique_ptr<WritableFile> target_file;
-  TF_RETURN_IF_ERROR(target_fs->NewWritableFile(target, &target_file));
-
-  uint64 offset = 0;
-  while (offset < size) {
-    char scratch[kCopyFileBufferSize];
-    StringPiece result;
-    size_t bytes_to_read =
-        offset + sizeof(scratch) < size ? sizeof(scratch) : size - offset;
-    TF_RETURN_IF_ERROR(src_file->Read(offset, bytes_to_read, &result, scratch));
-
-    TF_RETURN_IF_ERROR(target_file->Append(result));
-
-    offset += bytes_to_read;
-  }
-  return target_file->Close();
-}
 
 Status WriteBinaryProto(Env* env, const string& fname,
                         const ::tensorflow::protobuf::MessageLite& proto) {
--- a/tensorflow/core/platform/env.cc
+++ b/tensorflow/core/platform/env.cc
@@ -260,6 +260,60 @@ Status Env::RenameFile(const string& src, const string& target) {
   return src_fs->RenameFile(src, target);
 }
 
+Status Env::CopyFile(const string& src, const string& target) {
+  FileSystem* src_fs;
+  FileSystem* target_fs;
+  TF_RETURN_IF_ERROR(GetFileSystemForFile(src, &src_fs));
+  TF_RETURN_IF_ERROR(GetFileSystemForFile(target, &target_fs));
+  if (src_fs == target_fs) {
+    return src_fs->CopyFile(src, target);
+  }
+
+  uint64 size;
+  Status s = GetFileSize(src, &size);
+  if (!s.ok()) {
+    return s;
+  }
+  std::unique_ptr<RandomAccessFile> src_file;
+  s = src_fs->NewRandomAccessFile(src, &src_file);
+  if (!s.ok()) {
+    return s;
+  }
+  std::unique_ptr<WritableFile> target_file;
+  s = target_fs->NewWritableFile(target, &target_file);
+  if (!s.ok()) {
+    return s;
+  }
+
+  uint64 offset = 0;
+  while (offset < size) {
+    char scratch[kCopyFileBufferSize];
+    StringPiece result;
+    size_t chunk =
+        offset + sizeof(scratch) < size ? sizeof(scratch) : size - offset;
+    s = src_file->Read(offset, chunk, &result, scratch);
+    if (!s.ok()) {
+      return s;
+    }
+
+    s = target_file->Append(result);
+    if (!s.ok()) {
+      return s;
+    }
+
+    offset += chunk;
+  }
+  s = GetFileSize(src, &size);
+  if (!s.ok()) {
+    return s;
+  }
+  if (offset != size) {
+    return errors::Aborted("File ", src, " changed while reading: ", size,
+                           " vs. ", offset);
+  }
+  return target_file->Close();
+}
+
 string Env::GetExecutablePath() {
   char exe_path[PATH_MAX] = {0};
 #ifdef __APPLE__
@@ -361,57 +415,6 @@ Status WriteStringToFile(Env* env, const string& fname,
     s = file->Close();
   }
   return s;
-}
-
-Status CopyFile(Env* env, const string& oldpath, const string& newpath,
-                bool overwrite) {
-  // If overwrite is false and the newpath file exists then it's an error.
-  if (!overwrite && env->FileExists(newpath).ok()) {
-    return errors::AlreadyExists("file already exists");
-  }
-  uint64 file_size;
-  Status s = env->GetFileSize(oldpath, &file_size);
-  if (!s.ok()) {
-    return s;
-  }
-  std::unique_ptr<RandomAccessFile> oldfile;
-  s = env->NewRandomAccessFile(oldpath, &oldfile);
-  if (!s.ok()) {
-    return s;
-  }
-  std::unique_ptr<WritableFile> newfile;
-  s = env->NewWritableFile(newpath, &newfile);
-  if (!s.ok()) {
-    return s;
-  }
-
-  uint64 offset = 0;
-  while (offset < file_size) {
-    char scratch[kCopyFileBufferSize];
-    StringPiece result;
-    size_t chunk = offset + sizeof(scratch) < file_size ? sizeof(scratch)
-                                                        : file_size - offset;
-    s = oldfile->Read(offset, chunk, &result, scratch);
-    if (!s.ok()) {
-      return s;
-    }
-
-    s = newfile->Append(result);
-    if (!s.ok()) {
-      return s;
-    }
-
-    offset += chunk;
-  }
-  s = env->GetFileSize(oldpath, &file_size);
-  if (!s.ok()) {
-    return s;
-  }
-  if (offset != file_size) {
-    return errors::Aborted("File ", oldpath, " changed while reading: ",
-                           file_size, " vs. ", offset);
-  }
-  return newfile->Close();
 }
 
 // A ZeroCopyInputStream on a RandomAccessFile.
--- a/tensorflow/core/platform/env.cc
+++ b/tensorflow/core/platform/env.cc
@@ -269,7 +269,7 @@ Status Env::CopyFile(const string& src, const string& target) {
   if (src_fs == target_fs) {
     return src_fs->CopyFile(src, target);
   }
-  return tensorflow::CopyFile(src_fs, src, target_fs, target);
+  return FileSystemCopyFile(src_fs, src, target_fs, target);
 }
 
 string Env::GetExecutablePath() {
@@ -375,8 +375,8 @@ Status WriteStringToFile(Env* env, const string& fname,
   return s;
 }
 
-Status CopyFile(FileSystem* src_fs, const string& src, FileSystem* target_fs,
-                const string& target) {
+Status FileSystemCopyFile(FileSystem* src_fs, const string& src,
+                          FileSystem* target_fs, const string& target) {
   std::unique_ptr<RandomAccessFile> src_file;
   TF_RETURN_IF_ERROR(src_fs->NewRandomAccessFile(src, &src_file));
 
--- a/tensorflow/core/platform/env.cc
+++ b/tensorflow/core/platform/env.cc
@@ -40,7 +40,8 @@ limitations under the License.
 
 namespace tensorflow {
 
-constexpr size_t kCopyFileBufferSize = 1024 * 1024;
+// 128KB copy buffer
+constexpr size_t kCopyFileBufferSize = 128 * 1024;
 
 class FileSystemRegistryImpl : public FileSystemRegistry {
  public:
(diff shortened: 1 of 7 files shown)


tensorflow/core/platform/file_system.cc
```diff
@@ -259,4 +261,50 @@ Status FileSystem::RecursivelyCreateDir(const string& dirname) {
   return Status::OK();
 }
 
+Status FileSystem::CopyFile(const string& src, const string& target) {
+  uint64 size;
+  Status s = GetFileSize(src, &size);
```
rohan100jain: Use TF_RETURN_IF_ERROR(GetFileSize(src, &size)); instead. Here and elsewhere...

tensorflow/core/platform/file_system.cc
```diff
@@ -259,4 +261,50 @@ Status FileSystem::RecursivelyCreateDir(const string& dirname) {
...
+  s = NewWritableFile(target, &target_file);
+  if (!s.ok()) {
+    return s;
+  }
+
+  uint64 offset = 0;
+  while (offset < size) {
+    char scratch[kCopyFileBufferSize];
+    StringPiece result;
+    size_t chunk =
```
rohan100jain: Name this bytes_to_read?

tensorflow/core/platform/file_system.cc
```diff
@@ -259,4 +261,50 @@ Status FileSystem::RecursivelyCreateDir(const string& dirname) {
...
+    }
+
+    offset += chunk;
+  }
+  s = GetFileSize(src, &size);
+  if (!s.ok()) {
+    return s;
+  }
+  if (offset != size) {
+    return errors::Aborted("File ", src, " changed while reading: ", size,
```
rohan100jain: In this case, we'll end up copying something into target_file. Do those semantics make sense? Or should we delete target_file? What are the semantics in Unix for instance? Either way, can we document this in the header file?

tensorflow/core/platform/env.cc
```diff
@@ -258,6 +260,60 @@ Status Env::RenameFile(const string& src, const string& target) {
   return src_fs->RenameFile(src, target);
 }
 
+Status Env::CopyFile(const string& src, const string& target) {
```
rohan100jain: the logic for the same FS and different FS seem very similar... would it make sense to create a utility method that takes in FileSystem* src_fs and FileSystem* target_fs and does the copy. For the Filesystem::CopyFile method pass in this, this to that method?

tensorflow/core/platform/env.h
```diff
@@ -381,6 +385,10 @@ Status ReadFileToString(Env* env, const string& fname, string* data);
 Status WriteStringToFile(Env* env, const string& fname,
                          const StringPiece& data);
 
+/// A utility routine: copy contents of `oldpath` to `newpath`
+Status CopyFile(Env* env, const string& oldpath, const string& newpath,
```
rohan100jain: I don't see where this is implemented?

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +277,46 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
   return result;
 }
 
+Status PosixFileSystem::CopyFile(const string& src, const string& target) {
+#if defined(PLATFORM_WINDOWS)
```
rohan100jain: Why do you need to handle this case? From what I understand this case should be handled by platform/windows/windows_file_system

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +277,46 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
...
+  if (stat(translated_src.c_str(), &sbuf) != 0) {
+    return IOError(src, errno);
+  }
+  int src_fd = open64(translated_src.c_str(), O_RDONLY);
+  if (src_fd < 0) {
+    return IOError(src, errno);
+  }
+  string translated_target = TranslateName(target);
+  int target_fd =
+      open64(translated_target.c_str(), O_WRONLY | O_CREAT,
```
rohan100jain: Add a comment here explaining all the flags turned on.

rohan100jain (changes requested):
Thanks for the change! Please add tests for the new functionality added!

quaeler:
@rohan100jain will this make the 1.4 release?

tensorflow/core/platform/file_system.cc
yongtang: Thanks @rohan100jain, done.

tensorflow/core/platform/file_system.cc
yongtang: Done.

tensorflow/core/platform/file_system.cc
yongtang: @rohan100jain That was to follow the old semantics (where no partial copy in memory). Now as we use streaming, it is not possible to rollback any way. The `cp` in Linux is not impacted by the file change so I just removed the check at the end.

tensorflow/core/platform/env.cc
yongtang: Done.

tensorflow/core/platform/env.h
yongtang: Sorry, that was left over when check in. It has been cleaned.

tensorflow/core/platform/posix/posix_file_system.cc
yongtang: Removed. Thanks.

tensorflow/core/platform/posix/posix_file_system.cc
yongtang: Done. Thanks.

yongtang:
Thanks @rohan100jain for the review. The PR has been updated. Please take a look.

drpngx:
Jenkins, test this please.

sb2nov:
@yongtang can you look into the failing tests

yongtang:
On Mac sendfile is not defined in `<sys/sendfile.h>`. Though it looks like there is a Darwin version of `sendfile`:
https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man2/sendfile.2.html

Will take a look and update the PR.

quaeler:
@yongtang It seems like there is also an issue with the Linux ARM build:

```c++
11:44:48 /android/ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-g++ --std=c++11 -DIS_SLIM_BUILD -fno-exceptions -DNDEBUG -O2 --sysroot /android/ndk/platforms/android-21/arch-arm -Wno-narrowing -fomit-frame-pointer -march=armv7-a -mfloat-abi=softfp -mfpu=neon -fPIE -MT /workspace/tensorflow/contrib/makefile/gen/obj/tensorflow/core/platform/posix/posix_file_system.o -MMD -MP -MF /workspace/tensorflow/contrib/makefile/gen/dep//tensorflow/core/platform/posix/posix_file_system.Td -I/android/ndk/sources/android/support/include -I/android/ndk/sources/cxx-stl/gnu-libstdc++/4.9/include -I/android/ndk/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi/include -I. -I/workspace/tensorflow/contrib/makefile/downloads/ -I/workspace/tensorflow/contrib/makefile/downloads/eigen -I/workspace/tensorflow/contrib/makefile/downloads/gemmlowp -I/workspace/tensorflow/contrib/makefile/downloads/nsync/public -I/workspace/tensorflow/contrib/makefile/downloads/fft2d -I/workspace/tensorflow/contrib/makefile/gen/protobuf/include -I/workspace/tensorflow/contrib/makefile/gen/proto/ -I/workspace/tensorflow/contrib/makefile/gen/proto_text/ -c tensorflow/core/platform/posix/posix_file_system.cc -o 
11:44:50 tensorflow/core/platform/posix/posix_file_system.cc: In member function 'virtual tensorflow::Status tensorflow::PosixFileSystem::CopyFile(const string&, const string&)':
11:44:50 tensorflow/core/platform/posix/posix_file_system.cc:307:54: error: cannot convert 'off_t* {aka long int*}' to 'off64_t* {aka long long int*}' for argument '3' to 'ssize_t sendfile64(int, int, off64_t*, size_t)'
11:44:50      rc = sendfile64(target_fd, src_fd, &offset, chunk);
11:44:50                                                       ^
```

yongtang:
@quaeler Thanks and sorry for the late reply (was out of town). The Android failure could be fixed with `sendfile64` -> `sendfile`. I also updated the Mac OS X implementation so that appropriate sendfile API could be used.

quaeler:
@drpngx could you kick off a build please

drpngx:
Jenkins, test this please.

quaeler:
Jenkins seems troubled again - https://ci.tensorflow.org/job/tensorflow-pull-requests-mac/6766/console

(Thanks for the kickoff, @drpngx!)

yongtang:
Thanks @quaeler @drpngx, the PR has been updated with Darwin OS X fix.

quaeler:
Thanks @yongtang - your branch now builds the libtensorflow.so target fine for me on OS X 10.12.6.

@drpngx could you pass this on to Jenkins please?

martinwicke:
Jenkins, test this please.

martinwicke:
@rohan100jain are you good with the changes?

Checking again... Jenkins, test this please.

martinwicke:
@yongtang The file_io_test failure on MacOS does look a little suspicipous. I reran the tests, but could you take a look?

Thanks!

rohan100jain (approved):

quaeler:
@yongtang The [Darwin man page for sendfile](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man2/sendfile.2.html) reads differently than the Linux one (aside from method signature) and seems to imply the destination descriptor is not allowed to be a file descriptor, only a socket descriptor; this would jive with the failing tests that complain like:
```
InvalidArgumentError: /var/folders/l_/bvssjjhd2s10ymxdfx1xy12c0000gn/T/exporter_testI_uRih/export_no_shard/00000222-tmp/assets/hello42.txt; Socket operation on non-socket
```

drpngx:
@quaeler thanks for looking into this. Any recommendation on a fix?

yongtang:
@quaeler @drpngx Thank for the help. I think it makes sense to disable sendfile in MacOS and use fallback implementation. I don't have a Mac dev env setup at the moment but plan to setup one today to take a look.

quaeler:
@yongtang @drpngx I concur - disable sendfile usage for MacOS and fallback to read()/write()-isms.

drpngx:
Sounds good, waiting for the patch. Thanks!

yongtang:
The PR has been updated. Tested on my Mac machine and it works. Please take a look.

quaeler:
Passes for me as well. Thanks @yongtang !

martinwicke:
Jenkins, test this please.

yongtang:
MacOSX issue has been fixed now though Jenkins build on Windows still fails with linkage issue. I will take a look.

tensorflow/core/platform/env.cc
```diff
@@ -361,6 +374,32 @@ Status WriteStringToFile(Env* env, const string& fname,
   return s;
 }
 
+Status CopyFile(FileSystem* src_fs, const string& src, FileSystem* target_fs,
+                const string& target) {
+  uint64 size;
+  TF_RETURN_IF_ERROR(src_fs->GetFileSize(src, &size));
```
jart: From `RandomAccessFile`:

```c
  /// Returns `OUT_OF_RANGE` if fewer than n bytes were stored in `*result`
  /// because of EOF.
```

So you could possibly be able to avoid the round trip of fetching the length, if you check the status returned by `Read`.

tensorflow/core/platform/env.cc
```diff
@@ -361,6 +374,32 @@ Status WriteStringToFile(Env* env, const string& fname,
...
+
+  std::unique_ptr<RandomAccessFile> src_file;
+  TF_RETURN_IF_ERROR(src_fs->NewRandomAccessFile(src, &src_file));
```
jart: Do we want copies to be atomic? I just noticed our GCS RandomAccessFile impl doesn't lock-in the [generation](https://cloud.google.com/storage/docs/generations-preconditions). Can we do that @dbcode? If that can happen soon, I don't see any reason to block this PR.

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +281,57 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
...
+  }
+  int src_fd = open(translated_src.c_str(), O_RDONLY);
+  if (src_fd < 0) {
+    return IOError(src, errno);
+  }
+  string translated_target = TranslateName(target);
+  // O_WRONLY | O_CREAT:
+  //   Open file for write and if file does not exist, create the file.
+  // S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH:
+  //   Create the file with permission of 0666
```
jart: How about 0644?

tensorflow/core/platform/env.cc
```diff
@@ -361,6 +374,32 @@ Status WriteStringToFile(Env* env, const string& fname,
...
+
+  std::unique_ptr<WritableFile> target_file;
+  TF_RETURN_IF_ERROR(target_fs->NewWritableFile(target, &target_file));
+
+  uint64 offset = 0;
+  while (offset < size) {
+    char scratch[kCopyFileBufferSize];
+    StringPiece result;
+    size_t bytes_to_read =
+        offset + sizeof(scratch) < size ? sizeof(scratch) : size - offset;
```
jart: This is fine, but can we have an if statement instead? I always get scared when I have to search Google to make sure unsafe code is correct. Same goes for that other ternary.

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -34,6 +37,8 @@ limitations under the License.
 
 namespace tensorflow {
 
+constexpr size_t kPosixCopyFileBufferSize = 1024 * 1024;
```
jart: 128kB is the optimal according to this guy: https://eklitzke.org/efficient-file-copying-on-linux

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +281,57 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
...
+      break;
+    }
+    rc = write(target_fd, buffer, chunk);
+    offset += chunk;
+#endif
+    if (rc <= 0) {
+      break;
+    }
+  }
+  close(target_fd);
```
jart: Please check for `close` errors.

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +281,57 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
...
+    close(src_fd);
+    return IOError(target, errno);
+  }
+  int rc = 0;
+  off_t offset = 0;
+  while (offset < sbuf.st_size) {
+    size_t chunk = (sbuf.st_size - offset) < SSIZE_MAX ? (sbuf.st_size - offset)
+                                                       : SSIZE_MAX;
+#if !defined(__APPLE__)
+    rc = sendfile(target_fd, src_fd, &offset, chunk);
```
jart: kernelspace! :D

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +281,57 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
...
+#else
+    char buffer[kPosixCopyFileBufferSize];
```
jart: Ditto on heap memory.

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +281,57 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
...
+      open(translated_target.c_str(), O_WRONLY | O_CREAT,
+           S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
+  if (target_fd < 0) {
+    close(src_fd);
+    return IOError(target, errno);
+  }
+  int rc = 0;
+  off_t offset = 0;
+  while (offset < sbuf.st_size) {
+    size_t chunk = (sbuf.st_size - offset) < SSIZE_MAX ? (sbuf.st_size - offset)
```
jart: Mixing signed and unsigned with integer limits like this scares me a little bit because I can't tell if the code is correct. Would it be possible to write this in a way there the code is obviously correct? Even if it results in a ton of static_cast<> boilerplate? I'll also note that I would personally feel completely fine if sendfile() context-switched back into userspace every 128kB. What matters IIRC is that the memory itself isn't being copied between the two.

jart (changes requested):
I am *so* excited about this change. I really can't thank you enough for taking the initiative on this suggestion.

tensorflow/core/platform/posix/posix_file_system.cc
```diff
@@ -276,4 +281,57 @@ Status PosixFileSystem::RenameFile(const string& src, const string& target) {
...
+                                                       : SSIZE_MAX;
+#if !defined(__APPLE__)
```
jart: Having done more homework, I would say change this to `#if defined(__linux__) && !defined(__ANDROID__)` because it won't work on FreeBSD and you never know with Android. I've confirmed it will work on Linux kernels ≥2.6.33 (published in 2010.) TensorFlow supports as far back as CentOS 7, which uses Linux 3.10.0, so this should be fine. If we want to be even more conservative, we could use the [splice](https://en.wikipedia.org/wiki/Splice_(system_call)) system call instead, which became available in 2.6.17 in 2006. It's Linux-only, but we've constrained ourselves to Linux anyway.

yongtang: Thanks @jart. Done.

tensorflow/core/platform/env.cc
yongtang: Thanks. The PR has been updated and size has been removed.

tensorflow/core/platform/env.cc
```diff
@@ -361,6 +374,32 @@ Status WriteStringToFile(Env* env, const string& fname,
...
+
+  std::unique_ptr<RandomAccessFile> src_file;
+  TF_RETURN_IF_ERROR(src_fs->NewRandomAccessFile(src, &src_file));
+
+  std::unique_ptr<WritableFile> target_file;
+  TF_RETURN_IF_ERROR(target_fs->NewWritableFile(target, &target_file));
+
+  uint64 offset = 0;
+  while (offset < size) {
+    char scratch[kCopyFileBufferSize];
```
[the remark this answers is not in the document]
yongtang: Updated. Thanks.

tensorflow/core/platform/env.cc
yongtang: Done.

tensorflow/core/platform/posix/posix_file_system.cc
yongtang: Updated.

tensorflow/core/platform/posix/posix_file_system.cc
yongtang: Updated. Thanks.

tensorflow/core/platform/posix/posix_file_system.cc
yongtang: Done.

tensorflow/core/platform/posix/posix_file_system.cc
yongtang: Done.

tensorflow/core/platform/posix/posix_file_system.cc
yongtang: Done.

yongtang:
Thanks @jart for the review. The PR has been updated.

The Windows build might fail as there was a linkage error before and I haven't been able to find out the reason. Unfortunately I don't have a Windows dev machine available at the moment. I will see what I could do to resolve the Windows build error.

tensorflow/core/platform/env.h
```diff
@@ -373,6 +377,11 @@ struct ThreadOptions {
   size_t guard_size = 0;  // 0: use system default value
 };
 
+/// A utility routine: copy contents of `src` in file system `src_fs`
+/// to `target` in file system `target_fs`.
+Status CopyFile(FileSystem* src_fs, const string& src, FileSystem* target_fs,
```
jart: It might confuse users if env.h provides two APIs for CopyFile. The CMake error also points to this helper being the linker error. I'm not entirely sure why, but if the helper is removed it might fix the problem.

jart (changes requested):

tensorflow/core/platform/env.h
yongtang: Thanks @jart for the help and sorry for the late reply. I made a change to the name `CopyFile` -> `FileSystemCopyFile` and it seems that Windows build is successful. Please take a look.

yongtang:
@jart The PR has been updated. Now all build passes (I think `Ubuntu contrib` failure is unrelated). Please take a look and let me know if there are any issues.

drpngx:
@jart good to go?

tensorflowbutler:

tensorflowbutler:

yongtang:
@jart Can you take a look and see if the PR is ready to go? I think the build `Ubuntu contrib` is unrelated.

jart (approved):
Oh I didn't realize I was the blocker. Thanks for the ping.

This change is both fabulous and difficult. Something nice was accomplished here. I think it'll have impact, yet poses a tiny risk for bugs. If we keep our eyes peeled, and maybe attract a few more looks at this impl after we submit it, I think we're going to be fine.

LGTM

martinwicke:
Yup, it's unrelated.

PR #12662, opened 2017-08-28
Adding a decode to support Python 3.5 on Windows.

av8ramit:
Fix this error: python 3.5: TypeError: a bytes-like object is required, not 'str'


gunan:
Let's move this change internally to discuss the full details.

gunan (approved):

PR #12665, opened 2017-08-28
Conv2DGrad & MaxPoolGradHelper

bpiel:
Gradients for the nn ops, Conv2D and MaxPool. Ported from python:

https://github.com/tensorflow/tensorflow/blob/86d9171b006a2f4a65055228aa13d83d60916052/tensorflow/python/ops/nn_grad.py#L457

These are simple enough that I grouped them into a single PR. I can split into two, if necessary.

cc @suharshs @dguerra

### DIFF
--- a/tensorflow/cc/gradients/nn_grad.cc
+++ b/tensorflow/cc/gradients/nn_grad.cc
@@ -118,6 +118,57 @@ Status BiasAddGradHelper(const Scope& scope, const Operation& op,
 }
 REGISTER_GRADIENT_OP("BiasAdd", BiasAddGradHelper);
 
+Status Conv2DGrad(const Scope& scope, const Operation& op,
+                  const std::vector<Output>& grad_inputs,
+                  std::vector<Output>* grad_outputs) {
+  string data_format;
+  string padding;
+  std::vector<int32> strides;
+  bool use_cudnn_on_gpu;
+  auto attrs = op.output(0).node()->attrs();
+  GetNodeAttr(attrs, "data_format", &data_format);
+  GetNodeAttr(attrs, "padding", &padding);
+  GetNodeAttr(attrs, "strides", &strides);
+  GetNodeAttr(attrs, "use_cudnn_on_gpu", &use_cudnn_on_gpu);
+  Conv2DBackpropInput::Attrs input_attrs;
+  input_attrs.DataFormat(data_format);
+  input_attrs.UseCudnnOnGpu(use_cudnn_on_gpu);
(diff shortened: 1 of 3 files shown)


tensorflow/cc/gradients/nn_grad.cc
```diff
@@ -118,6 +118,57 @@ Status BiasAddGradHelper(const Scope& scope, const Operation& op,
...
+  filter_attrs.UseCudnnOnGpu(use_cudnn_on_gpu);
+  auto dx_2 = Conv2DBackpropFilter(scope, op.input(0),
+                                   Shape(scope, op.input(1)), grad_inputs[0],
+                                   strides, padding, filter_attrs);
+  grad_outputs->push_back(dx_2);
+  return scope.status();
+}
+REGISTER_GRADIENT_OP("Conv2D", Conv2DGrad);
+
+Status MaxPoolGradHelper(const Scope& scope, const Operation& op,
```
suharshs: I am not sure what the status of MaxPool and MaxPoolV2 is, but my guess is that MaxPoolV2 is the one more commonly used now. You may want to also add that gradient https://github.com/tensorflow/tensorflow/blob/86d9171b006a2f4a65055228aa13d83d60916052/tensorflow/python/ops/nn_grad.py#L545

If you feel like it you can add MaxPoolGradGrad and MaxPoolGradGradV2, but I doubt these are blocking your models.

suharshs (changes requested):

tensorflow/cc/gradients/nn_grad.cc
bpiel: Ah, didn't know about MaxPoolV2. Added it.

suharshs (approved):
Thanks!

bpiel:
should this get merged? Just worried about the branch becoming in conflict with master

martinwicke:
Jenkins, test this please.

martinwicke:
GPU build rerun at http://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/6635

martinwicke:
XLA rerun at http://ci.tensorflow.org/job/tensorflow-pull-requests-xla/2636

dguerra:
What's the difference between MaxPool and MaxPoolV2? Should it be used one over the other?

martinwicke:
We (almost) never change ops, whenever we have to, we instead introduce a new V2 (or V3, or V4) op. That makes sure that old serialized graphdefs keep working.

So in your code, use V2, it's the newest version.

PR #12666, opened 2017-08-29
c++ gradients for mean and sum

bpiel:
Gradients for the math ops, Mean and Sum. Ported from python:

https://github.com/tensorflow/tensorflow/blob/86d9171b006a2f4a65055228aa13d83d60916052/tensorflow/python/ops/math_grad.py#L95

I grouped these together because they share helper functions.

cc @suharshs

### DIFF
--- a/tensorflow/cc/BUILD
+++ b/tensorflow/cc/BUILD
@@ -276,6 +276,7 @@ cc_library(
         ":cc_ops",
         ":cc_ops_internal",
         ":grad_op_registry",
+        ":gradients",
     ],
     alwayslink = 1,
 )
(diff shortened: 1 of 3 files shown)


suharshs:
Ah unfortunately someone internally just submitted this change yesterday.  Should sync soon. Apologies :(

From now on can you create an issue with your in progress gradients with "c++ gradient" in the title. I will ask internal folks to do the same, this way we can avoid stepping on toes due to sync latency.

bpiel:
@suharshs No problem at all. Getting the functionality in there is the important thing.

PR #12668, opened 2017-08-29
Make Graph(long) Constructor Public

resec:
It would be very useful when user wants to do custom model loading in C/C++ via JNI then return the handle to Java and use this constructor to wrap/maintain the graph object.

Changed Graph(long) from `Graph(long)` To `public Graph(long)`.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Graph.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Graph.java
@@ -33,7 +33,7 @@ public final class Graph implements AutoCloseable {
   }
 
   /** Create a Graph from an existing handle (takes ownership). */
-  Graph(long nativeHandle) {
+  public Graph(long nativeHandle) {
     this.nativeHandle = nativeHandle;
   }
 


asimshankar:
Thanks for the PR, but  I am a little weary of this change. It makes a very unsafe call part of the public API which makes it very easy to incorrectly use or misuse. For example, something like `new Graph(1)` is going to crash with a segmentation fault bringing the JVM down. This also opens the door to security holes.

The use case you've described above does seem a bit esoteric. Can't you create the graph in Java? Or create the graph in native code but provide a serialized form to Java? There must be a better alternative to exposing unsafe calls, particularly those that can fail in mysterious ways to a Java programmer (for example, an invalid value will not raise a Java exception, just crash the JVM).

resec:
@asimshankar First of all thanks for your quick response.

I am not sure how this change would opens the door to security holes, but I am not a expert in this and actually, the reason I want this change is for security and performance in my case. 

I want to deploy my model to mobile device. and the final model size is around 100MB, and have the model zipped and encrypted and splitted into multiple files as assert before putting into the final APK. So I need to merge and decrypt and unzip the model in C/C++ first. 

As you can see, moving/copying 100MB across C/C++ and Java runtime for multiple times is not a good choice and leaves much memory footprint.

Note: in iOS case, I really got things done much more smoothly.......

Besides, I really not like the way that have the plain serialized graph existed in JVM, which is kind of easy for someone to just copy it. I understand that C/C++ is not perfect safe, but we all want more security especially the case that it would even bring better performance?

I have been struggling in this for few days, and this is the simplest change I came up to achieve what I want, by opening a kind of low-level API.

Just to demonstrate the situation I am facing, any advice towards a better design/solution will be really very appreciated.

Besides, I believe its user's responsibility to choice to what API to and take the consequence? in the case anyone wants to be evil, a public or not descritor can hardly stop it. So maybe add some more warning doc for this constructor?

Waiting for your feedback, great thanks.

asimshankar:
Regarding your specific use case, it seems that you should be able to provide a `ByteBuffer` (see Java NIO support for `DirectByteBuffers`) from native code to Java, and then use that in `Graph.importGraphDef` instead of creating the `TF_Graph` object in native code. Doing so would also mean that you don't copy 100MB across the boundary. Basically, if you need some data manipulation in native code, then make the boundary between Java and native code the data manipulation instead of making the boundary be TensorFlow objects.

If that doesn't work, I'd suggest posting a more detailed description of what you're trying to do on StackOverflow and seek opinions there - there is a larger community on StackOverflow and good ideas are likely to come up.

For the PR though, I'm going to close it out. I don't feel comfortable providing a public API that is very easy to use incorrectly and which manifests itself with a JVM segmentation fault instead of a clear error message/exception. It also breaks an abstraction boundary - the fact that the graphs are backed by native long handles is an implementation detail that I do not think should be exposed to the user.

Thanks for your understanding.

resec:
@asimshankar Thanks for your patience, I think for the moment I would stick with reflection to use this constructor. And I will really go trying the `DirectByteBuffers`.

Once again, here is just sharing my use case, and actually I was trying my luck with this PR(to see if people can accept the idea). I think "big model" is a relatively common case, and Tensorflow Android API is now relying on Java API, "model security" problem is also be an important concern in mobile deployment. 

So I do hope in Android/Java API, Tensorflow will have better support for this in the future.

I am fine with closing this PR. Once again, thanks.

PR #12670, opened 2017-08-29
add name param to ctc_ops

afpro:


tensorflow/python/ops/ctc_ops.py
```diff
@@ -31,7 +31,8 @@
 def ctc_loss(labels, inputs, sequence_length,
              preprocess_collapse_repeated=False,
              ctc_merge_repeated=True,
-             ignore_longer_outputs_than_inputs=False, time_major=True):
+             ignore_longer_outputs_than_inputs=False, time_major=True,
+             name=None):
```
caisq: The added `name` kwarg needs to be documented in the Args section of the doc string below.

tensorflow/python/ops/ctc_ops.py
```diff
@@ -149,7 +150,8 @@ def ctc_loss(labels, inputs, sequence_length,
       sequence_length,
       preprocess_collapse_repeated=preprocess_collapse_repeated,
       ctc_merge_repeated=ctc_merge_repeated,
-      ignore_longer_outputs_than_inputs=ignore_longer_outputs_than_inputs)
+      ignore_longer_outputs_than_inputs=ignore_longer_outputs_than_inputs,
+      name=name)
```
caisq: Can you add a unit test to test that using `ctc_loss` with the new `name` kwarg works.

afpro: ok, i will do this as soon as possible

afpro: done

drpngx:
Jenkins, test this please.

@caisq @ebrevdo how does the new version look?

drpngx:
Jenkins, test this please.

drpngx:
Once we like this change, we should trigger an API review.

sb2nov:
@afpro can you take a look at the failing tests. so that we can move the review forward

afpro:
@sb2nov i'll check this after work, cause i'm using a mac for work, linux at home.

ebrevdo:
Looks like the API has now changed; so LGTM but we'll need API approvers to describe how you should update the API metadata.

ebrevdo:
@martinwicke ptal - do we need to run the api regeneration?

martinwicke:
Yes, the instructions are with the failed test.

martinwicke:
@afpro, could you run the API golden regeneration (must be on Linux, py2), and add the changed files to this PR?

afpro:
'name' param is not necessary, close this PR.

PR #12682, opened 2017-08-29
modify lambda capture list, and prohibit copying objects.

horance-liu:
1. capture `work_name, done` by reference to avoid copying object.
2. capture pointer by value, and remove reference.

### DIFF
--- a/tensorflow/core/distributed_runtime/remote_device.cc
+++ b/tensorflow/core/distributed_runtime/remote_device.cc
@@ -67,12 +67,12 @@ void NewRemoteDevices(Env* env, WorkerCacheInterface* worker_cache,
     GetStatusResponse resp;
   };
   Call* call = new Call;
-  auto cb = [env, worker_cache, worker_name, done, wi,
+  auto cb = [env, worker_cache, &worker_name, &done, wi,
              call](const Status& status) {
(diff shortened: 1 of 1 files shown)


martinwicke:
Jenkins, test this please.

tensorflow/core/distributed_runtime/remote_device.cc
```diff
@@ -67,12 +67,12 @@ void NewRemoteDevices(Env* env, WorkerCacheInterface* worker_cache,
     GetStatusResponse resp;
   };
   Call* call = new Call;
-  auto cb = [env, worker_cache, worker_name, done, wi,
+  auto cb = [env, worker_cache, &worker_name, &done, wi,
```
mrry: Is this safe? `worker_name` is a `const string&` and `cb` might be called after `NewRemoteDevices()` returns and `worker_name` no longer refers to the same object. (Similarly, it seems as if `done` will be destroyed before the version captured by reference in this change is called in the callback.)

mrry:
Looking closer at the test failures, this pull request breaks the distributed runtime. Since it's not clear that there's a fix other than maintaining the status quo, I'm going to close this PR.

PR #12705, opened 2017-08-30
Verbs fix: Removed Dependency on Duplicate Recv Flag

yanivbl6:
This PR is a response for #11825 

I've added a new Queue and Table for the **RdmaTensorBuffer**, where the relevant information from the send-recv waiter table is stored in case the actual rdma-write does not occur (If the allocated buffer is too small, or busy). When we get a RDMA_MESSAGE_BUFFER_RESPONSE, or RDMA_MESSAGE_BUFFER_IDLE, we check the new queue/table instead of checking the waiter table. (This happens in the new ReSendNextItem() function).

Also added a new function that generates the callback lambda function, to avoid code redundancy.
The result is that the code is less clean- I am open to suggestions on making it more tidy.

@junshi15 @shamoya

### DIFF
--- a/tensorflow/contrib/verbs/rdma.cc
+++ b/tensorflow/contrib/verbs/rdma.cc
@@ -165,9 +165,9 @@ void RdmaAdapter::Process_CQ() {
           RdmaBuffer* ab = rc->tx_ack_buffer_;
           ab->SendNextItem();
           // find buffer
-          RdmaBuffer* tb = rc->FindBuffer(rm.name_);
+          RdmaTensorBuffer* tb = reinterpret_cast<RdmaTensorBuffer*> (rc->FindBuffer(rm.name_));
           tb->SetBufferStatus(remote, idle);
-          worker_env_->compute_pool->Schedule([tb]() { tb->SendNextItem(); });
+          worker_env_->compute_pool->Schedule([tb]() { tb->ReSendNextItem(); });
         } else if (rm.type_ == RDMA_MESSAGE_BUFFER_REQUEST) {
           // remote host requests to create a tensor buffer;
           // send ack to release remote tx message buffer
@@ -198,7 +198,7 @@ void RdmaAdapter::Process_CQ() {
           RdmaBuffer* ab = rc->tx_ack_buffer_;
           ab->SendNextItem();
           // find buffer
-          RdmaBuffer* tb = rc->FindBuffer(rm.name_);
+          RdmaTensorBuffer* tb = reinterpret_cast<RdmaTensorBuffer*> (rc->FindBuffer(rm.name_));
           CHECK(rm.buffer_size_ == tb->size_)
               << "rm.buffer_size = " << rm.buffer_size_
               << "tb->size_ = " << tb->size_ << "rm.name_ = " << rm.name_;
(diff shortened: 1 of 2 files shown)


shamoya:
+ @poxvoculi

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -793,9 +841,16 @@ void RdmaTensorBuffer::PostCopyOperations(
       VLOG(2) << "Extend RDMA buffer from " << size_ << " to " << buffer_size;
     }
     CreateCPUBuffer(buffer_size, false);
+    //Need to be recieved again, put into the re-recv queue and the table
+    requeue.push(key_with_step_id);
+    ReItem* item = new ReItem();
+    item->send_args = send_args;
+    item->recv_args = recv_args;
+    item->in = in;
```
junshi15: Do we need to increase reference count on [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/rendezvous.cc#L217-L219) or [that](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/rendezvous.cc#L175-L177)?
```
      if (item->recv_args.device_context) {
        item->recv_args.device_context->Ref();
      }
```

junshi15:
@yanivbl6, thanks for the quick response. It looks mostly good. I have a comment regarding the reference count.

yanivbl6:
Thanks @junshi15,
I've added a constructor / destructor to the ReItem struct, to take care of the referencing and the de-referencing.  Also removed the unused field (Status).

junshi15:
It looks good to me. @shamoya any comment?

yanivbl6:
@shamoya participated in the code writing, and reviewed it prior to the PR.

tensorflow/contrib/verbs/rdma.h
```diff
@@ -224,14 +225,52 @@ class RdmaMessageBuffer : public RdmaBuffer {
...
+      const string& key_with_step_id, const string& key,
+      int64 step_id, const Rendezvous::ParsedKey& parsed);
+
+  struct ReItem {
+    Rendezvous::Args send_args;
+    Rendezvous::Args recv_args;
+    Tensor in;
+    bool is_dead;
+
+    ReItem(const Rendezvous::Args& send_args_, const Rendezvous::Args& recv_args_, const Tensor& in_, bool is_dead_ ): send_args(send_args_), recv_args(recv_args_), in(in_), is_dead(is_dead_){
```
poxvoculi: too long line, break by col 80

tensorflow/contrib/verbs/rdma.cc
```diff
@@ -793,9 +840,12 @@ void RdmaTensorBuffer::PostCopyOperations(
       VLOG(2) << "Extend RDMA buffer from " << size_ << " to " << buffer_size;
     }
     CreateCPUBuffer(buffer_size, false);
+    //Need to be recieved again, put into the re-recv queue and the table
```
poxvoculi: s/recieved/received/

poxvoculi (approved):

martinwicke:
Jenkins, test this please.

yifeif:

PR #12708, opened 2017-08-30
Remove initialize_all_tables

alanyee:
Finishing TODO(yleon): Remove this function.


martinwicke:
Thank you for doing some cleanup work. We're very careful with removing functions from the API though. In this case, even though we're technically allowed to according to our deprecation and versioning policy, we'll err on the side of caution to avoid breaking users.

If you'd like, you can remove the TODO, and set the date in the deprecation notice to None instead. Best to open a new PR for that.

Issue #12294, opened 2017-08-15
Computation of Mean IoU


Sylvus:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
No.
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
Windows 7
- **TensorFlow installed from (source or binary)**:
Source.
- **TensorFlow version (use command below)**:
r1.2
- **Python version**: 
3.5

### Describe the problem

I believe that the calculation of the IoU metric in tf.metrics.mean_iou is incorrect. Consider an image which can have 3 classes (sky, road, building). In one image there is no building and everything has been perfectly classified as either sky or road so the tp + fn + fp for buildings is zero. To avoid divisions by zero in the IoU calculation (tp / (tp + fn + fp)) the current implementation sets the denominator to 1. This causes a result of 0 because tp=0 and we are computing 0=0/1. 

Both the sky and road class get a score of 1 so the overall image gets a mean_iou score of 2/3 even though every single pixel was correctly classified. I propose to change the metric so that when the denominator is 0 the result is 1. I can send a pull request if this is needed. 

### Source code / logs

Complete example as a unit test:

    class TestMetrics(tf.test.TestCase):

        def get_data(self):
            label_int64 = tf.constant([
                [[0, 1, 2, 2, 1, 0],
                 [2, 1, 0, 0, 1, 2]],
                [[2, 2, 2, 2, 2, 2],
                 [1, 1, 1, 1, 1, 1]]], dtype=tf.int64)

            predicted_perfect_int64 = tf.constant([
                [[1, 1, 2, 2, 1, 2],
                 [2, 1, 9, 3, 1, 2]],
                [[2, 2, 2, 2, 2, 2],
                 [1, 1, 1, 1, 1, 1]]], dtype=tf.int64)
            return label_int64, predicted_perfect_int64

        def test_iou_metric(self):
            with self.test_session():
                l, p = self.get_data()
                num_classes = 9
                weights = tf.cast(tf.not_equal(l, 0), tf.int32)

                iou, update_op = tf.metrics.mean_iou(
                    l - 1, p - 1, num_classes, weights=weights)
                tf.global_variables_initializer().run()
                tf.local_variables_initializer().run()
                update_op.eval()
                self.assertAlmostEqual(iou.eval(), 1)

Currently this test fails with: 0.222 is not equal to 1.

aquariusjay:
[...]

PR #12709, opened 2017-08-30
Fix for the IOU metric

Sylvus:
Previously the IOU metric under estimated the true value as described [here](https://github.com/tensorflow/tensorflow/issues/12294)

This pull request proposes a fix following the suggestion of @aquariusjay. The score is now only taking into account classes that actually appear in the sample. 

The new code works like this:
a) We compute how many classes appear in the sample.
b) We compute the score for every class. If the denominator is 0 then the nominator will be 0 as well! To avoid a zero division we set the denominator to 1 so the result will be 0/1=0. 
c) Instead of taking the mean over all scores we sum up all scores and divide by the previously computed number of classes in the sample.

### DIFF
--- a/tensorflow/python/kernel_tests/metrics_test.py
+++ b/tensorflow/python/kernel_tests/metrics_test.py
@@ -3331,7 +3331,7 @@ class MeanIOUTest(test.TestCase):
       sess.run(variables.local_variables_initializer())
       for _ in range(5):
         sess.run(update_op)
-      desired_output = np.mean([1.0 / 3.0, 2.0 / 4.0, 0.])
+      desired_output = np.mean([1.0 / 3.0, 2.0 / 4.0])
       self.assertAlmostEqual(desired_output, miou.eval())
 
   def testUpdateOpEvalIsAccumulatedConfusionMatrix(self):
@@ -3409,6 +3409,55 @@ class MeanIOUTest(test.TestCase):
       self.assertAllEqual([[2, 0], [2, 4]], update_op.eval())
       desired_miou = np.mean([2. / 4., 4. / 6.])
       self.assertAlmostEqual(desired_miou, miou.eval())
+
+  def testMissingClassInLabels(self):
+    labels = constant_op.constant([
+      [[0, 0, 1, 1, 0, 0],
+       [1, 0, 0, 0, 0, 1]],
+      [[1, 1, 1, 1, 1, 1],
+       [0, 0, 0, 0, 0, 0]]])
+    predictions = constant_op.constant([
+      [[0, 0, 2, 1, 1, 0],
+       [0, 1, 2, 2, 0, 1]],
+      [[0, 0, 2, 1, 1, 1],
+       [1, 1, 2, 0, 0, 0]]])
+    num_classes = 3
+    with self.test_session() as sess:
+      miou, update_op = metrics.mean_iou(labels, predictions, num_classes)
+      sess.run(variables.local_variables_initializer())
+      self.assertAllEqual([[7, 4, 3], [3, 5, 2], [0, 0, 0]], update_op.eval())
+      self.assertAlmostEqual(
+        1 / 3 * (7 / (7 + 3 + 7) + 5 / (5 + 4 + 5) + 0 / (0 + 5 + 0)),
+        miou.eval())
+
+  def testMissingClassOverallSmall(self):
+    labels = constant_op.constant([0])
+    predictions = constant_op.constant([0])
+    num_classes = 2
+    with self.test_session() as sess:
+      miou, update_op = metrics.mean_iou(labels, predictions, num_classes)
+      sess.run(variables.local_variables_initializer())
+      self.assertAllEqual([[1, 0], [0, 0]], update_op.eval())
+      self.assertAlmostEqual(1, miou.eval())
+
+  def testMissingClassOverallLarge(self):
+    labels = constant_op.constant([
+      [[0, 0, 1, 1, 0, 0],
+       [1, 0, 0, 0, 0, 1]],
+      [[1, 1, 1, 1, 1, 1],
+       [0, 0, 0, 0, 0, 0]]])
+    predictions = constant_op.constant([
+      [[0, 0, 1, 1, 0, 0],
+       [1, 1, 0, 0, 1, 1]],
+      [[0, 0, 0, 1, 1, 1],
+       [1, 1, 1, 0, 0, 0]]])
+    num_classes = 3
+    with self.test_session() as sess:
+      miou, update_op = metrics.mean_iou(labels, predictions, num_classes)
+      sess.run(variables.local_variables_initializer())
+      self.assertAllEqual([[9, 5, 0], [3, 7, 0], [0, 0, 0]], update_op.eval())
+      self.assertAlmostEqual(
+        1 / 2 * (9 / (9 + 3 + 5) + 7 / (7 + 5 + 3)), miou.eval())
 
 
 class MeanPerClassAccuracyTest(test.TestCase):
--- a/tensorflow/python/ops/metrics_impl.py
+++ b/tensorflow/python/ops/metrics_impl.py
@@ -929,6 +929,12 @@ def mean_iou(labels,
       cm_diag = math_ops.to_float(array_ops.diag_part(total_cm))
       denominator = sum_over_row + sum_over_col - cm_diag
 
+      # The mean is only computed over classes that appear in the
+      # label or prediction tensor. If the denominator is 0, we need to
+      # ignore the class.
+      num_valid_entries = math_ops.reduce_sum(math_ops.cast(
+          math_ops.not_equal(denominator, 0), dtype=dtypes.float32))
+
       # If the value of the denominator is 0, set it to 1 to avoid
       # zero division.
       denominator = array_ops.where(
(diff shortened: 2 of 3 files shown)


Sylvus:
I wanted to add a test but couldn't find the correct unit test file. Maybe it is missing? I posted on SO but got no reply, maybe somebody want to give me some advice. The SO post is here:

https://stackoverflow.com/questions/45937936/writing-tests-for-tensorflow

alextp:
The correct test file is tensorflow/python/kernel_tests/metrics_test.py

alextp (commented):
Happy to approve this once a test is added.

drpngx:
@Sylvus any luck on adding the test?

Sylvus:
Sorry for the delay. Had to finish an other project.

Yes, I found the test files and added 3 tests on my own. One is just ensuring that the behaviour in the normal case stays the same, one is a minimal example of the problem and the third one is a proper test case showing the new behaviour. Let me know if you would like to see any changes. 

Also I accidentally pushed from my AWS machine (without a user email) so I am not sure if the cla/google check will pass. I tried to rebase it but I am not sure that helps.

Thank you!

drpngx:
You can change the email on git with an option to commit

drpngx:
Jenkins, test this please.

drpngx:
Here is the error;

```
FAIL: //tensorflow/contrib/metrics:metric_ops_test (shard 2 of 3) (see /var/lib/jenkins/workspace/tensorflow-pull-requests-cpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local-opt/testlogs/tensorflow/contrib/metrics/metric_ops_test/shard_2_of_3/test.log).
INFO: From Testing //tensorflow/contrib/metrics:metric_ops_test (shard 2 of 3):
==================== Test output for //tensorflow/contrib/metrics:metric_ops_test (shard 2 of 3):
.2017-09-18 16:50:49.363948: I tensorflow/core/platform/cpu_feature_guard.cc:137] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX AVX2 FMA
...................................F.............../var/lib/jenkins/workspace/tensorflow-pull-requests-cpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local-opt/bin/tensorflow/contrib/metrics/metric_ops_test.runfiles/org_tensorflow/tensorflow/contrib/metrics/python/ops/metric_ops_test.py:4576: RuntimeWarning: Degrees of freedom <= 0 for slice
  fweights=weights[:stride * (i + 1)])
/usr/local/lib/python2.7/dist-packages/numpy/lib/function_base.py:2929: RuntimeWarning: divide by zero encountered in double_scalars
  c *= 1. / np.float64(fact)
/usr/local/lib/python2.7/dist-packages/numpy/lib/function_base.py:2929: RuntimeWarning: invalid value encountered in multiply
  c *= 1. / np.float64(fact)
............WARNING:tensorflow:From /var/lib/jenkins/workspace/tensorflow-pull-requests-cpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local-opt/bin/tensorflow/contrib/metrics/metric_ops_test.runfiles/org_tensorflow/tensorflow/contrib/metrics/python/ops/metric_ops_test.py:2304: streaming_recall_at_k (from tensorflow.contrib.metrics.python.ops.metric_ops) is deprecated and will be removed after 2016-11-08.
Instructions for updating:
Please use `streaming_sparse_recall_at_k`, and reshape labels from [batch_size] to [batch_size, 1].
................................
FAIL: testMultipleUpdatesWithMissingClass (__main__.StreamingMeanIOUTest)
  File "/var/lib/jenkins/workspace/tensorflow-pull-requests-cpu/bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/eab0d61a99b6696edb3d2aff87b585e8/execroot/workspace/bazel-out/local-opt/bin/tensorflow/contrib/metrics/metric_ops_test.runfiles/org_tensorflow/tensorflow/contrib/metrics/python/ops/metric_ops_test.py", line 4982, in testMultipleUpdatesWithMissingClass
    self.assertAlmostEqual(desired_output, miou.eval())
AssertionError: 0.27777777777777773 != 0.41666669 within 7 places
```

Sylvus:
Thank you! Send a fix. I wasn't aware of the fact that I also needed to change the test file for the streaming metrics iou which was located somewhere else. I added my test cases there as well. This seems like a lot of code duplication!

Let me try: Jenkins, test this please.

drpngx:
Jenkins, test this please.

Sylvus:
All good. Can somebody review and let me know how to change the already pushed commits to get approved by cla/google (or just override the tag)?

drpngx:
You need to change commits: `9207da9e9f9653c84481eba6853f20785c2b58a4`, `9207da9e9f9653c84481eba6853f20785c2b58a4`, and `8921c7b58f17045b58fb8e5a3700c9deff88bcfe` which are committed from the wrong author. Assuming that they are yours, you can simply squash them. You can use `git rebase --interactive`, then mark these commits as `fixup` or `squash` to merge them with previous commits.

drpngx:
OK, now it looks like there are conflicts. Could you pull rebase and push again?

Sylvus:
Yep, my squash did not work because of the merge. I rebased and pushed again.

drpngx:
Great, CLA issue solved. Did you add the test as @alextp requested?

Sylvus:
Yes, added 3 tests in total. Happy to delete one or two again if they are not needed. 
1. Test verifies that my change does not break the normal behaviour (when the class is present)
2. Test is a minimal example that fails without my changes
3. Test is a normal use case that would also fail without my changes.

drpngx:
Thanks! @alextp PTAL?

sb2nov:
@alextp ping

tensorflow/python/kernel_tests/metrics_test.py
```diff
@@ -3331,7 +3331,7 @@ def testMultipleUpdatesWithMissingClass(self):
       sess.run(variables.local_variables_initializer())
       for _ in range(5):
         sess.run(update_op)
-      desired_output = np.mean([1.0 / 3.0, 2.0 / 4.0, 0.])
```
alextp: Why does this change?

Sylvus: This test is using 3 classes but there are only predictions/labels for the first 2 classes. This means the last row and col of the matrix will be zero. Previously this would cause the result to be 0 but now those results are simply ignored.

tensorflow/python/ops/metrics_impl.py
```diff
@@ -929,14 +929,20 @@ def compute_mean_iou(name):
...
+
       # If the value of the denominator is 0, set it to 1 to avoid
       # zero division.
       denominator = array_ops.where(
           math_ops.greater(denominator, 0),
           denominator,
           array_ops.ones_like(denominator))
       iou = math_ops.div(cm_diag, denominator)
-      return math_ops.reduce_mean(iou, name=name)
+      return math_ops.reduce_sum(iou, name=name)/num_valid_entries
```
Sylvus: We could divide by zero here if the function is called without any predictions (or with all predictions masked out). Should we add a check for this case and raise an exception ourselves?

tensorflow/python/kernel_tests/metrics_test.py
alextp: Should those results still be zero? Otherwise it's an API compatibility change and we need to make a new v2 version of this metric and initially put it in contrib.

tensorflow/python/ops/metrics_impl.py
alextp: I think we should add a cond check for zero valid entries and return 0.0 in that case (which is I think what reduce_mean does)

tensorflow/python/kernel_tests/metrics_test.py
Sylvus: As far as I understand this, the matrix stays the same (the values are still 0). All that changes is that we count how many classes are actually in the input and divide by that number instead of the matrix size. 
In this test case the output changes because it is no longer:
1/3 * (1.0 / 3.0 +  2.0 / 4.0 + 0.) but instead
1/2 * (1.0 / 3.0 +  2.0 / 4.0 + 0.)
Even though the matrix dimensions do not change.

tensorflow/python/ops/metrics_impl.py
Sylvus: I added a check that would return 0. If you would like, I can also add a test to verify the correct behaviour. Let me know.

drpngx:
Jenkins, test this please.

tensorflow/python/kernel_tests/metrics_test.py
alextp: Let's add an iou v2 metric in contrib then.

Sylvus: Are you sure? Sorry if I wasn't clear in my previous comment:
All that changes is the final result of the function in a corner case where one (or more) of the classes are not present. The matrix stays the same.

I can move it to contrib but I assumed that such a small change would not require a v2.

alextp: Ok, I see.

alextp (approved):

PR #12711, opened 2017-08-30
Update CUB version in the cmake build.

gunan:


mrry (approved):
Looks good to me, although I thought this change had already been made (perhaps on the internal version)?

gunan:
yeah, I think a push will be better, there are more fixes we want externally.

PR #12714, opened 2017-08-30
Fix dtype for streaming confusion matrix

taion:
Currently, dependents of `_streaming_confusion_matrix` don't actually work correctly with floating-point weights.

### DIFF
--- a/tensorflow/python/kernel_tests/metrics_test.py
+++ b/tensorflow/python/kernel_tests/metrics_test.py
@@ -3541,7 +3541,7 @@ class MeanPerClassAccuracyTest(test.TestCase):
       _enqueue_vector(sess, weights_queue, [1.0])
       _enqueue_vector(sess, weights_queue, [1.0])
       _enqueue_vector(sess, weights_queue, [0.0])
-      _enqueue_vector(sess, weights_queue, [1.0])
+      _enqueue_vector(sess, weights_queue, [0.5])
       _enqueue_vector(sess, weights_queue, [0.0])
       weights = weights_queue.dequeue()
 
@@ -3551,7 +3551,7 @@ class MeanPerClassAccuracyTest(test.TestCase):
       variables.local_variables_initializer().run()
       for _ in range(6):
         sess.run(update_op)
-      desired_output = np.mean([2.0 / 2.0, 1.0 / 2.0])
+      desired_output = np.mean([2.0 / 2.0, 0.5 / 1.5])
       self.assertAlmostEqual(desired_output, mean_accuracy.eval())
 
   def testMultipleUpdatesWithMissingClass(self):
--- a/tensorflow/python/kernel_tests/metrics_test.py
+++ b/tensorflow/python/kernel_tests/metrics_test.py
@@ -3538,10 +3538,10 @@ class MeanPerClassAccuracyTest(test.TestCase):
       weights_queue = data_flow_ops.FIFOQueue(
           6, dtypes=dtypes_lib.float32, shapes=(1, 1))
       _enqueue_vector(sess, weights_queue, [1.0])
-      _enqueue_vector(sess, weights_queue, [1.0])
+      _enqueue_vector(sess, weights_queue, [0.5])
       _enqueue_vector(sess, weights_queue, [1.0])
       _enqueue_vector(sess, weights_queue, [0.0])
-      _enqueue_vector(sess, weights_queue, [0.5])
+      _enqueue_vector(sess, weights_queue, [1.0])
       _enqueue_vector(sess, weights_queue, [0.0])
       weights = weights_queue.dequeue()
 
--- a/tensorflow/python/ops/metrics_impl.py
+++ b/tensorflow/python/ops/metrics_impl.py
@@ -258,8 +258,11 @@ def _streaming_confusion_matrix(labels, predictions, num_classes, weights=None):
     total_cm: A `Tensor` representing the confusion matrix.
     update_op: An operation that increments the confusion matrix.
   """
+  # TODO: This should be int64 when weights is None, but legacy bug-
+  # compatibility requires keeping this as float64 in that case.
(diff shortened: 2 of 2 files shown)


tensorflow/python/ops/metrics_impl.py
```diff
@@ -258,8 +258,11 @@ def _streaming_confusion_matrix(labels, predictions, num_classes, weights=None):
     total_cm: A `Tensor` representing the confusion matrix.
     update_op: An operation that increments the confusion matrix.
   """
+  # TODO: This should be int64 when weights is None, but legacy bug-
+  # compatibility requires keeping this as float64 in that case.
+  cm_dtype = dtypes.float64
```
taion: The more straightforward fix would be to just invert the branches.

However, if anybody is importing this function from here, and is e.g. dividing this by whatever in Python 2 without doing a float cast, this would then change the numerical results that they see.

Given that this bug has been around for a while, I think it'd perhaps be justifiable to keep the old behavior for cases where weights aren't specified.

I personally wouldn't be affected by a correct fix of just inverting the branches, though, and would be happy to amend this PR if such a fix would be better.

martinwicke:
@poxvoculi your name was all over blame, I was confused too. Probably a merge you did.

@alextp can you take a look?

tensorflow/python/kernel_tests/metrics_test.py
```diff
@@ -3541,7 +3541,7 @@ def testMultipleUpdatesWithWeights(self):
       _enqueue_vector(sess, weights_queue, [1.0])
       _enqueue_vector(sess, weights_queue, [1.0])
       _enqueue_vector(sess, weights_queue, [0.0])
-      _enqueue_vector(sess, weights_queue, [1.0])
```
alextp: Why this change?

tensorflow/python/ops/metrics_impl.py
alextp: I think inverting the branches is better here.

alextp (changes requested):

tensorflow/python/kernel_tests/metrics_test.py
taion: It's the simplest way to exercise the new logic – the old code would fail with this case, since the weight rounds to 0.

tensorflow/python/ops/metrics_impl.py
taion: Sure, will do.

taion:
Updated.

tensorflow/python/kernel_tests/metrics_test.py
taion: Truncate to 0, rather.

alextp (approved):

alextp:
Jenkins, test this please.

alextp:
Please address the test failure as it is relevant.

taion:
Oops, my bad. I updated the wrong weight for the test. Fixed now.

taion:
Anyway, be aware that if you have any internal consumers of `_streaming_confusion_matrix` without weights in Python 2 that divide by integers (e.g. for building a streaming normalized confusion matrix metric), this would constitute a bit of a breaking change in practice, as in the no-weights case, the confusion matrix tensors will now be of type `int64` rather than `float64` as before.

alextp:
Jenkins, test this please.

alextp:
Jenkins, test this please.

tensorflow/python/ops/metrics_impl.py
```diff
@@ -259,7 +259,7 @@ def _streaming_confusion_matrix(labels, predictions, num_classes, weights=None):
     update_op: An operation that increments the confusion matrix.
   """
   # Local variable to accumulate the predictions in the confusion matrix.
-  cm_dtype = dtypes.int64 if weights is not None else dtypes.float64
+  cm_dtype = dtypes.float64 if weights is not None else dtypes.int64
```
martinwicke: Can you make this float64 unconditionally? That is a slightly less breaking change, since it doesn't change the return type in the case that actually works (without weights). Changing the type in the case that was broken is a plain bugfix.

martinwicke (changes requested):

martinwicke:
One change request from API review.

tensorflow/python/ops/metrics_impl.py
taion: This was the initial implementation per https://github.com/tensorflow/tensorflow/pull/12714/commits/4b521d0992fbdd7d9f18a708111e2501689f7994. @alextp explicitly asked for it to be changed to the current implementation per https://github.com/tensorflow/tensorflow/pull/12714/commits/4b521d0992fbdd7d9f18a708111e2501689f7994#r137026663.

I'd be happy to change it back to my original implementation (with the FIXME?), but please confirm that this is as desired.

In some moral sense, `int64` probably is more correct with no weights (or with integer weights), but I share your concerns about API breakage. Just want to confirm this given the specific earlier request to use `int64` in the no-weights case.

taion: There are no semver implications here per se, since `_streaming_confusion_matrix` is not part of the public API, and in fact no public APIs change.

martinwicke: In every place we use it, we convert to float. So it may make more sense to always have it in float so we don't have to do that. 

But you're right, technically it's an implementation detail and nobody would need to ever know.

taion: Yeah. I'm fine either way though. Just let me know which you want.

martinwicke: Let's go with all float.

taion: Done.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

Issue #12686, opened 2017-08-29
Feature Request: C++ gradient for SoftmaxCrossEntropyWithLogits


bpiel:
Implement the gradient for SoftmaxCrossEntropyWithLogits in c++ so that it is available for TF_AddGradients.

This is the python code that I believe would need to be ported:
https://github.com/tensorflow/tensorflow/blob/4b2fb49fd7578afe7e289936f347af581b5bdab1/tensorflow/python/ops/nn_grad.py#L409

skye:
@bpiel you're working on this right?

@suharshs FYI

bpiel:
@skye Not currently. If I start, I will post here. If someone else starts first, hopefully they'll post here.

facaiy:
Hi, I'm really interested in the work because it is a little challenge for me. #12391 of @bpiel seems like a perfect example, right? I could be more confident if more detailed guide are given or someone would like shepherd it.

suharshs:
@facaiy Happy to review your PR. :) 
@kbsriram may also be interested in reviewing/guiding.

facaiy:
Thanks, @suharshs .

I have read the code in python side,
```python
@ops.RegisterGradient("SoftmaxCrossEntropyWithLogits")
def _SoftmaxCrossEntropyWithLogitsGrad(op, grad_loss, grad_grad):
```

If I understand correctly, the arguments `grad_loss` and `grad_grad` in python side will be packed in `grad_inputs` in C++ side, right?

Namely, the outline of code in C++ should like:
```c++
Status SoftmaxCrossEntropyWithLogitsGrad(const Scope& scope, const Operation& op,
                                         const std::vector<Output>& grad_inputs,
                                         std::vector<Output>* grad_outputs) {
  auto grad_loss = grad_inputs[0];
  auto grad_grad = grad_inputs[1];

  // migrate the implementation in python to here.
  // code...........
  // auto grad = .....

  grad_outputs->push_back(grad);
  grad_outputs->push_back(null);
  return scope.status();
}
REGISTER_GRADIENT_OP("SoftmaxCrossEntropyWithLogitsGrad", SoftmaxCrossEntropyWithLogitsGrad);
```

Please correct me if I'm wrong.

kbsriram:
@facaiy the outline of the code looks fine, one change would be to use
```c++
grad_outputs->push_back(NoGradient())
```
rather than `null` to indicate gradients should not be propagated.

[...]

PR #12720, opened 2017-08-31
PREP: migrate lgamma gradient to c++ side

facaiy:
It's my first attempt to contribute to C++ side. Hence the PR is opened early to get feedback.

Migrate the implementation of _Lagmma from python side to c++ side, see  #12686.

+ [x] add unit tests.
+ [ ] pass all tests.


tensorflow/cc/gradients/math_grad_test.cc
```diff
@@ -679,6 +682,30 @@ TEST_F(CWiseUnaryGradTest, Atan_Complex) {
   }
 }
 
+TEST_F(CWiseUnaryGradTest, Lgamma) {
+  auto x_fn = [this](const int i) { return RV({0, -1, 1, -2, 2, -3, 3}); };
+  auto dy_fn = [this](const float x) { return x + RV({-2, 2, -3, 3, -4, 4}); };
+  auto dx_fn = [this](const float x, const float dy) {
+    // where is digamma? Eigen::digamma ?
```
facaiy: Hi, @kbsriram and @skye . I get in trouble. Since `digamma` function is missing in std library, is [Eigen::digamma](https://eigen.tuxfamily.org/dox/namespaceEigen.html#af40db84b3db19fe25fe2f77c429420e5) a good candidate? Any help will be greatly appreciated, thanks.

tensorflow/cc/gradients/math_grad.cc
```diff
@@ -687,6 +687,18 @@ Status MeanGrad(const Scope& scope, const Operation& op,
 }
 REGISTER_GRADIENT_OP("Mean", MeanGrad);
 
+Status LgammaGrad(const Scope& scope, const Operation& op,
+                  const std::vector<Output>& grad_inputs,
+                  std::vector<Output>* grad_outputs) {
+  auto grad = grad_inputs[0];
+  Scope grad_scope = scope.WithControlDependencies(grad);
+  auto x = ConjugateHelper(grad_scope, op.output(0));
```
kbsriram: op.input(0)?

tensorflow/cc/gradients/math_grad_test.cc
kbsriram: Hm... don't have any guidance here, though it strikes me all these tests probably should use ComputeGradientError anyway :-) but unfortunately it currently only tests for real-valued inputs, not complex ones.

In the interests of simplicity, maybe one alternative is to use `NaryGradTest` rather than the `CWiseUnaryGradTest`, and add a note that it doesn't test complex values? You'll want to select values away from any poles to avoid numeric estimation errors. Maybe something like
```c++
TEST_F(NaryGradTest, Lgamma) {
// ...
  TensorShape shape({2, 3});
  auto x = PlaceHolder(scope_, DT_FLOAT, Placeholder::Shape(shape));
  auto y = Lgamma(scope_, x);
  Tensor x_init_value = test::AsTensor<float>({ /* six random float values, but not close to any poles for Lgamma */}, {2, 3});
  RunTest(x, x_init_value, y, shape);
}

Comment (facaiy) [reply tensorflow/cc/gradients/math_grad_test.cc]: Thank you very much! I'll modify code later.

File: tensorflow/cc/gradients/math_grad.cc (hunk 2 above)
Comment (facaiy) [reply tensorflow/cc/gradients/math_grad.cc]: fixed

File: tensorflow/cc/gradients/math_grad_test.cc (hunk 1 above)
Comment (facaiy) [reply tensorflow/cc/gradients/math_grad_test.cc]: modified.

File: tensorflow/cc/gradients/math_grad_test.cc [hunk 3]
```diff
@@ -1091,5 +1091,16 @@ TEST_F(NaryGradTest, Minimum) {
...
 }
 
+TEST_F(NaryGradTest, Lgamma) {
+  TensorShape shape({3, 2});
+  auto x = Placeholder(scope_, DT_FLOAT, Placeholder::Shape(shape));
+  auto y = Lgamma(scope_, x);
+  // Select values to avoid instability when computing finite differences.
+  // Ref: https://en.wikipedia.org/wiki/File:Gamma_plot.svg
+  Tensor x_init_value =
+      test::AsTensor<float>({-3.5f, -2.5f, -1.5f, 1.0f, 2.0f, 3.5f}, {3, 2});
```
facaiy: Since `lgamma` is the log of the absolute value of `Gamma(x)`, the values is selected by  :

facaiy:
Hi, @kbsriram @skye . Could you take a look? Thanks.

kbsriram:
Looks ok to me, may be an appropriate time to get an opinion/review from one of the maintainers like @suharshs ?

tensorflow/cc/gradients/math_grad_test.cc
```diff
@@ -1091,5 +1091,17 @@ TEST_F(NaryGradTest, Minimum) {
...
+  RunTest(x, x_init_value, y, shape);
+  // TODO: add test case for complex values
```
suharshs: If you don't plan to add the test case for complex values in this pull request, please assign the todo to me like 'TODO(suharshs): ...' so that someone owns it in the future.

Thanks!

suharshs (changes requested):
Thanks! looks great!

tensorflow/cc/gradients/math_grad_test.cc
facaiy: modified todo as suggested.

suharshs (approved):

yifeif:

facaiy:
Thanks, @suharshs . As select values to avoid instability is a little tricky,  I am curious where is the unit tests of python implementation.

suharshs:
I don't believe testing was done very thoroughly for the python gradients :( if there are these for math_grad they would be here: https://github.com/petewarden/tensorflow_makefile/blob/master/tensorflow/python/ops/math_grad_test.py

But it doesn't look like there is one for this op.

PR #12741, opened 2017-09-01
Export C API symbols of _pywrap_tensorflow_internal.so on OS X

fritzo:
This follows up #10469 by exporting C API symbols on OS X.

## How?

This PR widens the `.lds` matching pattern to include the C API symbols as they are mangled on OS X.

Before this PR the C API symbols are exported in Linux but are hidden in OS X:
```sh
(Linux)$ nm _pywrap_tensorflow_internal.so | grep TF_CloseSession
00000000012d2d40 T TF_CloseSession
00000000010f8e50 t _wrap_TF_CloseSession

(OS X)$ nm _pywrap_tensorflow_internal.so | grep TF_CloseSession
00000000002be1b0 t _TF_CloseSession
000000000001a0a0 t __Z21_wrap_TF_CloseSessionP7_objectS0_
```
After this PR the C API symbols are exported in both Linux and OS X:
```sh
(Linux)$ nm _pywrap_tensorflow_internal.so | grep TF_CloseSession
0000000000fa25f0 T TF_CloseSession
0000000000cf02e0 t _wrap_TF_CloseSession

(OS X)$ nm _pywrap_tensorflow_internal.so | grep TF_CloseSession
0000000000392880 T _TF_CloseSession
000000000001ae50 t __Z21_wrap_TF_CloseSessionP7_objectS0_
```
(notice the change from `t` to `T`, i.e. static to global)

## Why?

This is important for libraries that use multiple client APIs as described in #7541. My use case is [an R package](https://github.com/nimble-dev/nimble) that uses the Python API through [rstudio/tensorflow](https://github.com/rstudio/tensorflow) and also uses the C API library: I build tf graphs using the Python API and run those graphs from R-wrapped C++ code using the C API. This already works in Linux, and should also work in OS X after this PR.

## Tested

Successfully loaded [a shared library](https://github.com/nimble-dev/nimble/tree/devel/packages/nimble/inst/CppCode/tensorflow.cpp) in R that uses symbols from `_pywrap_tensorflow_internal.so`:
- [x] Linux
- [x] OS X


martinwicke:
I believe we are going another way, and will provide a separate DLL you can link against to get to the C API, but @allenlavoie to confirm.

fritzo:
@martinwicke @allenlavoie Can you confirm that #7541 has been resolved, so that both DLLs can be linked? If not, could this PR be merged as a temporary workaround?

allenlavoie:
We are planning to split out a shared object, which precompiled op libraries will depend on, and remove RTLD_GLOBAL when loading pywrap_tensorflow. This shared object will not include the C API symbols (the C API relies on several ops, and we'd like to keep ops and kernels out of this new shared object). Both the C API and pywrap_tensorflow will depend on this new shared object, but trying to load both of them will attempt to register ops and kernels twice and throw an error.

So this PR is probably your best bet medium-term for adding Python API calls to an otherwise C API usage. Removing RTLD_GLOBAL will mean that you'll need to explicitly link against _pywrap_tensorflow_internal.so for the C API symbols (you don't need to do it now, but I'd recommend it to avoid breaking in the future).

allenlavoie (approved):

fritzo:
Thanks, @allenlavoie !

allenlavoie:
Jenkins, test this please

yifeif:
Looks like there are some conflicts. @fritzo do you mind resolve them? Thanks!

yifeif:

fritzo:
@yifeif Thanks, I've resolved the conflict. To resolve the conflict I've also added `*` to the `TFE_` symbols exported by #12785 discussed in #12784, so that change will also work in OS X (this is the same issue I had).

Issue #6815, opened 2017-01-12
keep_dims vs keepdims


yaroslavvb:
There are 15 results for [keepdims](https://github.com/tensorflow/tensorflow/search?utf8=%E2%9C%93&q=keepdims) and 77 results for [keep_dims](https://github.com/tensorflow/tensorflow/search?utf8=%E2%9C%93&q=keep_dims&type=Code) in TF codebase, any chance this can be made consistent in 1.0? Ideally TF would just match numpy API and use `keepdims` @martinwicke @karpathy

yaroslavvb:
@aselle for comment on whether it's a feasible change

aselle:
Yes, it's possible, probably it would be handled like axis where we keep the old argument around, but make keepdims the defacto version. It's too late to remove the old arg  for 1.0, but we can definitely make both work and the document be keepdims.

yaroslavvb:
Cool, having alias for `keepdims` would solve the user annoyance problem

tensorflowbutler:
It has been 14 days with no activity and this issue has an assignee.Please update the label and/or status accordingly.

kho:
Names like "keepdims" is a violation of PEP-8's naming convention [https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables], thus a violation of TensorFlow's style guide

> Method Names and Instance Variables
>
> Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability. 

Now one has to write code in inconsistent style like

```
tf.reduce UNDERSCORE max(..., keep NO_UNDERSCORE dims=True)
```

to avoid a deprecation warning.

Given that keep_dims had been the majority in TF's code base, wouldn't also accepting keepdims (while issuing a warning to suggest "keep_dims" instead) be a better solution?

yaroslavvb:
IMHO compatibility with numpy is more important than compatibility with PEP-8

neilpanchal:
I am not sure why this was modified in the first place? `keep_dims` is perfectly good.

martinwicke:
We modified it to conform to the numpy API, which for most users is much more relevant than PEP8. 

We could talk about supporting both (without deprecation warning), or even pushing the keepdims support into a tf.numpy compatibility module. I'd much rather just have one though. This bug is about that, make sure that one can use a single version throughout.

@saxenasaurabh FYI.

apacha:
[...]

PR #12756, opened 2017-09-01
Use `keepdims` and maintain backward compatible for `keep_dims`

yongtang:
This fix tries to address the issue raised in #6815 where
both `keepdims` and `keep_dims` were used with inconsistency.

This fix changes related api to `keepdims` while at the same
time maintain backward compatible for `keep_dims` so that
use will not be impacted.

### DIFF
--- a/tensorflow/python/ops/linalg_ops.py
+++ b/tensorflow/python/ops/linalg_ops.py
@@ -422,7 +422,7 @@ def svd(tensor, full_matrices=False, compute_uv=True, name=None):
 
 
 # pylint: disable=redefined-builtin
-def norm(tensor, ord='euclidean', axis=None, keep_dims=False, name=None):
+def norm(tensor, ord='euclidean', axis=None, keepdims=None, name=None, keep_dims=None):
   r"""Computes the norm of vectors, matrices, and tensors.
 
   This function can compute several different vector norms (the 1-norm, the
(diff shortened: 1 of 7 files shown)


tensorflow/python/ops/linalg_ops.py
```diff
@@ -422,7 +422,7 @@ def svd(tensor, full_matrices=False, compute_uv=True, name=None):
 
 
 # pylint: disable=redefined-builtin
-def norm(tensor, ord='euclidean', axis=None, keep_dims=False, name=None):
+def norm(tensor, ord='euclidean', axis=None, keepdims=None, name=None, keep_dims=None):
```
martinwicke: please use @deprecated_args

yongtang:
Thanks @martinwicke for the review. The PR has been updated with `@deprecated_args` used.

drpngx:
Any reason to choose `keepdims` instead of `keep_dims`? The latter seems like the better choice to me.

yongtang:
@drpngx Not knowing enough history about the code base though I would assume the reason of having `keepdims` instead of `keep_dims` is to align with `numpy`? e.g., 

https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.amin.html#numpy.amin

At the moment `keras` backend use `keepdims` while the rest of tensorflow mostly use `keep_dims`.

drpngx:
Thank you @yongtang , that makes sense, `keepdims` seems good then.

drpngx:
So, sorry to be dense, but does that mean that current code will issue a warning when `tf.reduce_sum(.., keep_dims=True)` will issue a warning but and behave like `tf.reduce_sum(..., keepdims=False)` (the default)?

Don't we want to keep **both** and override `keep_dims` with `keepdims` if both are specified?

yongtang:
Thanks @drpngx. The current behavior of this PR is to show a warning in doc string (from the annotation `@deprecated_args`), and accept either `keep_dims` or `keepdims` in the call.

Both `keep_dims` or `keepdims` should work.

However, in case both `keep_dims` and `keepdims` are specified in the same call, then an error will be thrown.

drpngx:
Oh, just saw the code. Sounds good.

drpngx:
Jenkins, test this please.

yongtang:
The PR has been updated with 
```
    bazel-bin/tensorflow/tools/api/tests/api_compatibility_test \
              --update_goldens True
```
so that `//tensorflow/tools/api/tests:api_compatibility_test` could pass.

martinwicke:
Jenkins, test this please.

yongtang:
The `Linux CPU Tests Makefile` test might be unrelated I think?

drpngx:
this one seems like a download error

Jenkins, test this please.

martinwicke (approved):

neilpanchal:
Violates PEP guidelines, https://github.com/tensorflow/tensorflow/issues/6815#issuecomment-355110926

I vote for keeping `keep_dims`. There needs to be a strong compelling reason to change. Or support both and get rid of the annoying warnings.

PR #12762, opened 2017-09-02
Removing extra space

printdhruv:


martinwicke:
Yifei, is that space intentional?

yifeif:
Thanks for the PR @printdhruv. The space here is intentional. I will close this PR for now.

PR #12766, opened 2017-09-02
Update monitors.py

alanyee:
Refactor contrib

### DIFF
--- a/tensorflow/contrib/learn/python/learn/monitors.py
+++ b/tensorflow/contrib/learn/python/learn/monitors.py
@@ -51,7 +51,7 @@ from tensorflow.python.summary import summary as core_summary
 from tensorflow.python.training import saver as saver_lib
 from tensorflow.python.training import session_run_hook
 from tensorflow.python.training import training_util
-from tensorflow.python.util.deprecation import deprecated
+from tensorflow.python.util import deprecation
 from tensorflow.python.util import tf_inspect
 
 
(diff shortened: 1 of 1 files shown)


tensorflow/contrib/learn/python/learn/monitors.py
```diff
@@ -54,7 +51,7 @@
 from tensorflow.python.training import saver as saver_lib
 from tensorflow.python.training import session_run_hook
 from tensorflow.python.training import training_util
-from tensorflow.python.util import deprecation
+from tensorflow.python.util.deprecation import deprecated
```
martinwicke: We generally prefer to import the whole file.

tensorflow/contrib/learn/python/learn/monitors.py
```diff
@@ -42,9 +42,6 @@
 import numpy as np
 import six
 
-from tensorflow.contrib.framework import deprecated
-from tensorflow.contrib.framework.python.ops import variables as contrib_variables
-from tensorflow.contrib.learn.python.learn.summary_writer_cache import SummaryWriterCache
```
martinwicke: All those are not needed? Good find!

martinwicke (changes requested):

tensorflow/contrib/learn/python/learn/monitors.py
alanyee: Ah I actually update this code some time ago, and realized that these import statements were no longer needed.

yifeif:

alanyee:
@martinwicke does this look good?

martinwicke (approved):

PR #12768, opened 2017-09-02
Removing /contrib/tensorboard and its references

k4r4koyun:
The only reference to /contrib/tensorboard was in /contrib/__init.py__ and /contrib/keras/python/keras/callbacks.py.

Both of those references did not include and tensorboard plugins that the Tensorboard repo does not have. So I have added multiple if-exists statements inside /keras/callbacks.py and removed /contrib/tensorboard completely.

The reason for this, some guy at Arch Linux repos have added a "rm -rf" to his PKGBUILD and now me and anybody who has the latest version of Tensorflow on Arch Linux is unable to use it. I thought maybe we can clear up the confusion since Kenser will be included with Tensorflow by default.

Link to my ticket: [Arch Linux - Tensorflow Ticket](https://bugs.archlinux.org/task/55477?project=5&status%5B0%5D=open)


goktay:
I signed it!

martinwicke:
I don't think I have fully understood the problem, but I think you want to make this change to master, not 1.3.

goktay:
I have made a new PR to the master branch at https://github.com/tensorflow/tensorflow/pull/12782.

The reason behind the PR is that having another tensorboard in tensorflow repo seems to create confusion here and there. It is not a bug, just a recommendation to clean it up.

PR #12809, opened 2017-09-05
Fix compile_nsync.sh

resec:
compile_nsync.sh is not workable with android x86 arch, here is the fix by replacing arm specific arch option.


yifeif:
@m3bm3b mind taking a look? Thanks!

m3bm3b (approved):
looks good to me.

yifeif:

PR #12814, opened 2017-09-05
Fix a typo

Meinwerk:
Fix a typo in comments


Meinwerk:
I signed it!

caisq:
This is not a typo. This doc string contains special characters (e.g., backslashes). So adding r before the doc string is required by our linter.

PR #12820, opened 2017-09-05
Install name fix for libtensorflow.so

eaplatanios:
There exists a problem with the built shared library for Mac. If ones creates a shared library that depends on libtensorflow.so and distributes it in a folder, along with libtensorflow.so, then the Mac linker will failed to load the library, unless libtensorflow.so is in `DYLD_LIBRARY_PATH`. This is problematic when libraries are distributed together as a package (e.g., in the context of JNI). From what I read online, this can be fixed using `@rpath`. This is what this patch is about and it seems to work in my experiments.


asimshankar:
@allenlavoie : Mind taking a look?

asimshankar:
@eaplatanios : Could you elaborate a bit - in the context of JNI libraries, the Java API explicitly loads the `.so` using [`System.loadLibrary`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/java/src/main/java/org/tensorflow/NativeLibrary.java#L83), and I don't think the `rpath` issue has troubled usage on OS X there. So, I'd like to understand the details of the problem you're running into a bit more.

asimshankar:
Oh, I see, you probably have another shared library that depends on `libtensorflow.so`, whereas in the Java API right now we package a single `.so`. 

Looks good then. Thanks.

asimshankar:
Jenkins, test this please

eaplatanios:
@asimshankar Yes, that's right. What I've done is that I package my cross-compiled JNI bindings along with the main Scala API in one package distributed through Sonatype. Then, for the TensorFlow binaries, I package each pre-compiled binary in its own package. Thus, when you load my API, you don't have to download the pre-compiled binaries. Also, when you do download them via depending on one of my wrapping packages, you only get to download the one corresponding to your platform (done so that the JARs are not very big).

This way, when I load the libraries through the JVM, I extract the JNI bindings library into a temporary directory. If you have also provided one of my wrapping packages for the main native library in your classpath, then I also extract that in the same directory. I then load the libraries from that directory, but in this case, `dlopen` would fail to load the JNI bindings dependency on the main native library due to the absolute path in the dependency name.

I hope this makes sense.

caisq (approved):

PR #12828, opened 2017-09-05
Skeleton code for annotation processor.

kbsriram:
Add build changes and skeleton code (with test harness) for
the Operator annotation processor. This change focuses on build
and test-related changes, and generates an empty Ops class.

Please see #7149 for the master tracking issue.


drpngx:
@asimshankar what do you think?

Jenkins, test this please.

sb2nov:
@asimshankar ping

tensorflow/tools/ci_build/builds/libtensorflow.sh
```diff
@@ -78,9 +78,40 @@ function build_libtensorflow_tarball() {
...
-  cp bazel-bin/tensorflow/java/libtensorflow.jar bazel-bin/tensorflow/java/libtensorflow-src.jar ${DIR}
+  cp bazel-bin/tensorflow/java/libtensorflow.jar ${DIR}
+  cp_normalized_srcjar bazel-bin/tensorflow/java/libtensorflow-src.jar ${DIR}/libtensorflow-src.jar
   cp bazel-genfiles/tensorflow/tools/lib_package/libtensorflow_proto.zip ${DIR}
   chmod -x ${DIR}/*
 }
+
+# Helper function to copy a srcjar after moving any source files
+# directly under the root to the "maven-style" src/main/java layout
+# Arguments:
```
asimshankar: I'm a bit lost on why we end up in this situation (that the src_jar contains anything outside the "maven-style" hierarchy. Perhaps that will dawn on my as I see the rest of the CL, just wondering out aloud for now.

tensorflow/java/src/test/java/org/tensorflow/processor/operator/BasicProcessorTest.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.testing.compile.Compilation;
+import com.google.testing.compile.Compiler;
+import com.google.testing.compile.JavaFileObjects;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Basic tests for {@link org.tensorflow.processor.operator.OperatorProcessor}. */
+@RunWith(JUnit4.class)
+public final class BasicProcessorTest {
```
asimshankar: Should this be `OperatorProcessorTest`?

tensorflow/java/BUILD
```diff
@@ -40,6 +41,31 @@ filegroup(
...
+    name = "operator_plugin",
+    generates_api = True,
+    processor_class = "org.tensorflow.processor.operator.OperatorProcessor",
+    visibility = ["//visibility:public"],
+    deps = [":operator_plugin_library"],
+)
+
+java_library(
+    name = "operator_plugin_library",
+    srcs = [":operator_plugin_sources"],
```
asimshankar: Why bother with the filegroup here instead of say `srcs = glob(["src/gen/java/org/tensorflow/processor/**/*.java"])`?

tensorflow/java/BUILD
```diff
@@ -40,6 +41,31 @@ filegroup(
...
+    javacopts = JAVACOPTS,
+    resources = [":operator_plugin_config"],
```
asimshankar: Same here - inline the resource instead of indirecting through a `filegroup`?

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
+package org.tensorflow.processor.operator;
```
asimshankar: Could we reduce one level of nesting and make this `org.tensorflow.processor`? Is there anything else in the `processor` package?

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.annotation.processing.AbstractProcessor;
+import javax.annotation.processing.Filer;
+import javax.annotation.processing.Messager;
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.annotation.processing.RoundEnvironment;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.TypeElement;
+import javax.tools.Diagnostic.Kind;
+
```
asimshankar: Unrelated question: Why is the `RetentionPolicy` of the annotation set to `CLASS` instead of `SOURCE`?

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+  @Override
+  public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+    // Nothing needs to be done at the end of all rounds.
+    if (roundEnv.processingOver()) {
+      return false;
+    }
+
+    // Collect all classes tagged with our annotation.
+    Set<TypeElement> opClasses = new HashSet<TypeElement>();
+    for (TypeElement annotation : annotations) {
```
asimshankar: Shouldn't there be precisely one `annotations`? If so, perhaps we can assert that `annotations.size()` is 1 and then call `collectOpClasses` once?

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+
+  private boolean collectOpClasses(
+      RoundEnvironment roundEnv, Set<TypeElement> opClasses, TypeElement annotation) {
+    boolean result = true;
+    for (Element e : roundEnv.getElementsAnnotatedWith(annotation)) {
+      // This processor has to aggregate all op classes in one round, as it generates a single Ops
+      // API class which cannot be modified once generated. If we find an annotation after we've
+      // generated our code, flag the location of each such class.
+      if (hasRun) {
+        error(e, "Cannot handle generated @Operators");
```
asimshankar: Perhaps elaborate on the error message - explaining that we already processed Operators and wrote out a file and cannot process additional ones?

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+        result = false;
+        continue;
+      }
+
+      // @Operator can only apply to types, so e must be a TypeElement.
+      if (!(e instanceof TypeElement)) {
+        error(e, "@Operator can only be applied to classes");
```
asimshankar: Can we add more information that will help those that may encounter this error in debugging? For example, `e.getKind()` and `e.getSimpleName()` seem like they have something useful in them?

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+      opClasses.add((TypeElement) e);
+    }
+    return result;
+  }
+
+  private void error(Element e, String message, Object... args) {
+    message(Kind.ERROR, e, message, args);
+  }
+
+  private void message(Kind kind, Element e, String msg, Object... args) {
```
asimshankar: Optional: If `message` is used only for `error`, perhaps inline it in `error`?

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  private boolean collectOpClasses(
+      RoundEnvironment roundEnv, Set<TypeElement> opClasses, TypeElement annotation) {
+    boolean result = true;
+    for (Element e : roundEnv.getElementsAnnotatedWith(annotation)) {
+      // This processor has to aggregate all op classes in one round, as it generates a single Ops
+      // API class which cannot be modified once generated. If we find an annotation after we've
+      // generated our code, flag the location of each such class.
+      if (hasRun) {
```
asimshankar: Instead of doing this in the loop, what about the following:

- Make this method:

```java
private boolean collectOpClasses(Set<? extends Element> elements, Set<TypeElement> opClasses) {
  for (Element e : elements) {
    ...
  }
}
```

- Make the caller (`process`) do something like:

```java
assert annotations.size() == 0;
Set<? extends Element> annotated = roundEnv.getElementsAnnotatedWith(annotation);
if (annotated.size() == 0) {
  return true;
}
if (annotated.size() > 0 && hasRun) {
  // print error
  return false;
}
if (!collectOpClasses(annotated, opClasses)) {
  return false;
}
...
```

It's essentially the same, but I think it makes the expected constraints clearer to follow. Though, it's possible you disagree :)

asimshankar (changes requested):
Thanks for doing this and apologies for the delay. Some comments.

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
kbsriram: This was mostly if anticipating processors for other annotations down the road, but also sgtm to reorganize later if needed. I've moved it up a level.

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
kbsriram: No good reason at this point, and we can change it to a longer retention if we find a need. Have changed it to SOURCE.

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.annotation.processing.Filer;
+import javax.annotation.processing.Messager;
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.annotation.processing.RoundEnvironment;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.TypeElement;
+import javax.tools.Diagnostic.Kind;
+
+public final class OperatorProcessor extends AbstractProcessor {
```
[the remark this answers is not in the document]
kbsriram: Done

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
kbsriram: It can be zero or 1 depending on which phase this runs in. I'll combine this with the rest of the constraint reformulation you suggested. Ptal, and if overall it reads better to you, sgtm.

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
kbsriram: Done, combined with the previous suggestion.

tensorflow/tools/ci_build/builds/libtensorflow.sh
kbsriram: The problem is that the annotation processor doesn't control where its generated sources are written. The processor only provides a package name - the rest of it is determined by the environment.

The environment is first determined by the way bazel calls javac, and then by how bazel packages such generated files into its -srcs.jar. bazel generates files into a bazel-out directory, and then jars them directly in the root of its -srcs.jar file. Is there an option I missed to control all this? I tried adding things like compiler options, but this ends up just confusing bazel.

tensorflow/java/src/test/java/org/tensorflow/processor/operator/BasicProcessorTest.java
kbsriram: Done - if the tests cases gets large large, I'll split it into BasicOperatorProcessorTest if needed.

tensorflow/java/BUILD
kbsriram: done

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
kbsriram: Done

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
kbsriram: Given this shows up as a compiler error message with file/line/character pointing to the incorrectly targeted element, do you think that might be sufficient to locate the problem element itself? I've added e.getKind() to the error message as well.

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
kbsriram: done

asimshankar (approved):
Looks great, thanks. One minor comment

tensorflow/java/src/gen/java/org/tensorflow/processor/operator/OperatorProcessor.java
asimshankar: Sounds great, thanks.

tensorflow/java/BUILD
```diff
@@ -41,6 +42,21 @@ filegroup(
...
 
+java_plugin(
+    name = "operator_plugin",
+    generates_api = True,
+    processor_class = "org.tensorflow.processor.OperatorProcessor",
+    visibility = ["//visibility:public"],
+    deps = [":operator_plugin_library"],
+)
+
+java_library(
```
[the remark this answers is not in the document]
kbsriram: Each java_plugin defines one processor, so I've renamed it to "processor" and "processor_library", wdyt? If we add another processor, we can disambiguate this name later as needed.

tensorflow/tools/ci_build/builds/libtensorflow.sh
kbsriram: Thanks for the review Asim!

Have added a comment here.

kbsriram:

asimshankar:
Jenkins, test this please

asimshankar (approved):

kbsriram:
Could a kindly member look at/restart the testing? Looks like the "Ubuntu CC" test didn't run, or stalled. Thanks!

asimshankar:
Jenkins, test this please.

Issue #11341, opened 2017-07-07
Feature request: Add float16 support for Conv3D, MaxPool3D and AvgPool3D ops


Warvito:
Support for `tf.float16 dtype` was added ([#1300](https://github.com/tensorflow/tensorflow/issues/1300)) to a bunch of ops. Can we add it for conv3d too, please?

conv3d is important to development of videos and medical images systems. Since both consumes a lot of memory, it would be good to have fp16 support to allow deeper models.

cy89:
@sesse and @benoitsteiner could you comment on plans for further fp16 support, please?

sesse:
@cy89 I left Google over a year ago, so I'm not involved in this anymore.

cy89:
Thanks, @sesse; sorry I didn't know.

tensorflowbutler:
It has been 14 days with no activity and the `awaiting tensorflower` label was assigned. Please update the label and/or status accordingly.

tensorflowbutler:
A member of the TensorFlow organization has replied after the stat:awaiting tensorflower label was applied.

tatatodd:
Looks like #12832 was merged, which (I believe) resolves this issue.  If I'm mistaken, just ping this issue and I'll re-open.

Warvito:
Hi @tatatodd 

#12832 did not implemented support for MaxPool3D and AvgPool3D ops with half type. Would you please consider re-open this issue?

RSteendam:
When I set dtype to float16 for a model with Conv3D layers, I get the following errors 
_ValueError: Tensor conversion requested dtype float32 for Tensor with dtype float16: 'Tensor("spatial_stream_Conv3d_1a_7x7_conv/kernel/read:0", shape=(7, 7, 7, 3, 64), dtype=float16)'
TypeError: Input 'filter' of 'Conv3D' Op has type float16 that does not match type float32 of argument 'input'._

So it seems like fp16 is still not supported for Conv3D operations. It would be very welcome if this will be implemented with the current rise of mixed- and low precision GPUs.

PR #12832, opened 2017-09-05
Float16 (half or Eigen::half) for conv3d ops

opensourcemattress:
Registrations of conv3d operations with fp16,  fp16 for batch_norms in tf.layers and tf.contrib.layers.  Related issue: https://github.com/tensorflow/tensorflow/issues/11341

I don't understand how fp16 operations must be implemented in low-level with CUDA and I didn't make any additional optimizations.
There is no implementation for fused batch_norm yet.

I copied part of code for dtypes from conv2d test to conv3d test and seems like all works. 
With fp16 I get `inf` and large absolute errors with large numbers. I suppose that it's fine. At least CPU and GPU  implementations return almost similar values. I skip such cases in test.
Example of such case:

> 
> use_gpu: False
> dtype: <dtype: 'float32'>
> data_format: NDHWC
> expected =  [36564.0, 38022.0, 39480.0, 37824.0, 39354.0, 40884.0, 39084.0, 40686.0, 42288.0, 46644.0, 48678.0, 50712.0, 47904.0, 50010.0, 52116.0, 49164.0, 51342.0, 53520.0, 107124.0, 112614.0, 118104.0, 108384.0, 113946.0, 119508.0, 109644.0, 115278.0, 120912.0, 117204.0, 123270.0, 129336.0, 118464.0, 124602.0, 130740.0, 119724.0, 125934.0, 132144.0]
> actual =  [[[[[  36564.   38022.   39480.]
>     [  37824.   39354.   40884.]
>     [  39084.   40686.   42288.]]
> 
>    [[  46644.   48678.   50712.]
>     [  47904.   50010.   52116.]
>     [  49164.   51342.   53520.]]]
> 
> 
>   [[[ 107124.  112614.  118104.]
>     [ 108384.  113946.  119508.]
>     [ 109644.  115278.  120912.]]
> 
>    [[ 117204.  123270.  129336.]
>     [ 118464.  124602.  130740.]
>     [ 119724.  125934.  132144.]]]]]
> 
> use_gpu: False
> dtype: <dtype: 'float16'>
> data_format: NDHWC
> expected =  [36564.0, 38022.0, 39480.0, 37824.0, 39354.0, 40884.0, 39084.0, 40686.0, 42288.0, 46644.0, 48678.0, 50712.0, 47904.0, 50010.0, 52116.0, 49164.0, 51342.0, 53520.0, 107124.0, 112614.0, 118104.0, 108384.0, 113946.0, 119508.0, 109644.0, 115278.0, 120912.0, 117204.0, 123270.0, 129336.0, 118464.0, 124602.0, 130740.0, 119724.0, 125934.0, 132144.0]
> actual =  [[[[[ 36544.  38016.  39488.]
>     [ 37824.  39360.  40896.]
>     [ 39072.  40704.  42304.]]
> 
>    [[ 46656.  48672.  50688.]
>     [ 47936.  50016.  52128.]
>     [ 49184.  51328.  53536.]]]
> 
> 
>   [[[    inf     inf     inf]
>     [    inf     inf     inf]
>     [    inf     inf     inf]]
> 
>    [[    inf     inf     inf]
>     [    inf     inf     inf]
>     [    inf     inf     inf]]]]]
> fp16 using may result in inf values and large absolute errors when used with large numbers, skipping
> 
> use_gpu: True
> dtype: <dtype: 'float32'>
> data_format: NDHWC
> expected =  [36564.0, 38022.0, 39480.0, 37824.0, 39354.0, 40884.0, 39084.0, 40686.0, 42288.0, 46644.0, 48678.0, 50712.0, 47904.0, 50010.0, 52116.0, 49164.0, 51342.0, 53520.0, 107124.0, 112614.0, 118104.0, 108384.0, 113946.0, 119508.0, 109644.0, 115278.0, 120912.0, 117204.0, 123270.0, 129336.0, 118464.0, 124602.0, 130740.0, 119724.0, 125934.0, 132144.0]
> actual =  [[[[[  36564.   38022.   39480.]
>     [  37824.   39354.   40884.]
>     [  39084.   40686.   42288.]]
> 
>    [[  46644.   48678.   50712.]
>     [  47904.   50010.   52116.]
>     [  49164.   51342.   53520.]]]
> 
> 
>   [[[ 107124.  112614.  118104.]
>     [ 108384.  113946.  119508.]
>     [ 109644.  115278.  120912.]]
> 
>    [[ 117204.  123270.  129336.]
>     [ 118464.  124602.  130740.]
>     [ 119724.  125934.  132144.]]]]]
> 
> use_gpu: True
> dtype: <dtype: 'float16'>
> data_format: NDHWC
> expected =  [36564.0, 38022.0, 39480.0, 37824.0, 39354.0, 40884.0, 39084.0, 40686.0, 42288.0, 46644.0, 48678.0, 50712.0, 47904.0, 50010.0, 52116.0, 49164.0, 51342.0, 53520.0, 107124.0, 112614.0, 118104.0, 108384.0, 113946.0, 119508.0, 109644.0, 115278.0, 120912.0, 117204.0, 123270.0, 129336.0, 118464.0, 124602.0, 130740.0, 119724.0, 125934.0, 132144.0]
> actual =  [[[[[ 36576.  38016.  39488.]
>     [ 37824.  39360.  40896.]
>     [ 39072.  40672.  42304.]]
> 
>    [[ 46656.  48672.  50720.]
>     [ 47904.  50016.  52128.]
>     [ 49152.  51328.  53504.]]]
> 
> 
>   [[[    inf     inf     inf]
>     [    inf     inf     inf]
>     [    inf     inf     inf]]
> 
>    [[    inf     inf     inf]
>     [    inf     inf     inf]
>     [    inf     inf     inf]]]]]
> fp16 using may result in inf values and large absolute errors when used with large numbers, skipping

fp16 is not fully covered by tests because I not sure how to do it.


yifeif:
@yzhwang do you mind taking a look? Thanks!

drpngx:
@yzhwang any cycles for this?

Jenkins, test this please.

sb2nov:
@yzhwang ping

yzhwang:
Sorry I haven't looking at this yet. Will finish it this week.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -80,23 +90,42 @@ def _SetupValuesForDevice(self, tensor_in_sizes, filter_in_sizes, stride,
...
+
           print("expected = ", expected)
           print("actual = ", value)
-          self.assertAllClose(expected, value.flatten(), atol=tolerance,
-                              rtol=1e-6)
+          tol = 1e-5
+          if value.dtype == np.float16:
+            tol = 1e-3
+            # fp16 using may result in inf values and large absolute errors
+            # when used with large numbers
```
yzhwang: Could you be more specific about large numbers? Or if you could change the value of the input tensors to make them smaller, that would be better. I'm worried that the skipped tests may affect our coverage.

tensorflow/python/layers/convolutional.py
```diff
@@ -810,6 +810,7 @@ def conv3d(inputs,
       bias_constraint=bias_constraint,
       trainable=trainable,
       name=name,
+      dtype=inputs.dtype.base_dtype,
```
yzhwang: Check if this breaks backward compatibility please.

tensorflow/python/layers/normalization.py
```diff
@@ -567,6 +567,7 @@ def batch_normalization(inputs,
       fused=fused,
       trainable=trainable,
       name=name,
+      dtype=inputs.dtype.base_dtype,
```
yzhwang: If you haven't tested this as you do for conv3d, I highly recommend you to split this into another pull request.

yzhwang (commented):
Thank you for the pull request. Please address my comments. Also, we are working on enabling half on batch_norm right now. Once the change merges to github, me or my colleague will point you to that change. As for now, I highly recommend you to make this pull request handle conv3d only. It will be much easier to get merged.

tensorflow/python/layers/convolutional.py
opensourcemattress: I have added this string similarly to conv2d methods.
Tests are rewritten similarly to conv2d tests so compatibility is checked by existing tests. There are both fp16 and fp32 in https://github.com/opensourcemattress/tensorflow/blob/e5f6a30d7b9d95e52fad491b6f951eb71b54df9e/tensorflow/python/kernel_tests/conv_ops_3d_test.py#L56 
and they are used here https://github.com/opensourcemattress/tensorflow/blob/e5f6a30d7b9d95e52fad491b6f951eb71b54df9e/tensorflow/python/kernel_tests/conv_ops_3d_test.py#L94

opensourcemattress: I haven't changed this main line https://github.com/opensourcemattress/tensorflow/blob/e5f6a30d7b9d95e52fad491b6f951eb71b54df9e/tensorflow/python/kernel_tests/conv_ops_3d_test.py#L83
It means that there is nothing to do.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: fp16 max value is `65504.0`. So if we try to convolve array and filter which values contain 1e3 it's highly probable that I will got `inf`. Should I write this in code comment?

For now I use test values from old tests for both dtypes and skip only cases with large values. I can try to divide large numbers in fp16 tests. Is it appropriate?

tensorflow/core/kernels/conv_grad_ops_3d.cc
```diff
@@ -1101,6 +1104,23 @@ class Conv3DBackpropFilterOp<GPUDevice, T> : public OpKernel {
 };
 
 REGISTER_KERNEL_BUILDER(
```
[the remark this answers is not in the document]
opensourcemattress: Thanks, I will rewrite it.

tensorflow/python/layers/normalization.py
opensourcemattress: Yes, if you are working on batch norm it's better to remove it from my pull request.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: Sorry, I mean "which contain values greater than 1e4", as checked here
 https://github.com/opensourcemattress/tensorflow/blob/0853d075ceaaaf18f5d8de35656635f9e4cd6845/tensorflow/python/kernel_tests/conv_ops_3d_test.py#L122

yzhwang: Your explanation here is fine. Please remove the additional logging lines (3 prints) in there (111,112,113). Thank you!

tensorflow/python/layers/convolutional.py
yzhwang: LGTM.

tensorflow/python/layers/normalization.py
yzhwang: My colleague is working on that. I saw you have removed this. Thank you!

tensorflow/core/kernels/conv_grad_ops_3d.cc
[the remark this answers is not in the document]
yzhwang: Thank you.

yzhwang (commented):
In general LGTM. Please update the title of your pull request as you've made a few more changes after the initial pull request. Just wait for the test to finish to approve.

tensorflow/contrib/layers/python/layers/layers.py
```diff
@@ -633,6 +633,7 @@ def batch_norm(inputs,
           renorm_clipping=renorm_clipping,
           renorm_momentum=renorm_decay,
           name=sc.name,
+          dtype=inputs.dtype.base_dtype,
```
opensourcemattress: What do you think about this line?
I forgot to remove it in last commit. It adds possibility to use not-fp32-types with tf.contrib.slim.batch_norm and it also doesn't break backward compatibility. 
If you are working on batch norm and it's better to remove it from this pull request, please, add same line in your work.

yzhwang: @reedwm Would you like to comment on this?

reedwm: We should not support fp16 and fp64 without tests. Also, for fused_batch_norm, the only fp16 case we support is where the input is fp16 and the scale and offset are in fp32. So if we support fp16 for non-fused batch norm, we should support that case.

In short, I'd remove the line. I plan on adding support for fp16 in layers shortly. If you want to do so, tell me so I don't repeat the work.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -80,23 +90,42 @@ def _SetupValuesForDevice(self, tensor_in_sizes, filter_in_sizes, stride,
...
-      result = self._SetupValuesForDevice(
-          tensor_in_sizes,
-          filter_in_sizes,
-          stride,
-          padding,
-          data_format,
-          use_gpu=use_gpu)
-      results.append(result)
-      tolerance = 1e-2 if use_gpu else 1e-5
+    for (data_format, use_gpu) in GetTestConfigs():
```
reedwm: Remove the parentheses around data_format, use_gpu.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -80,23 +90,42 @@ def _SetupValuesForDevice(self, tensor_in_sizes, filter_in_sizes, stride,
...
+            tol = 1e-3
+            # fp16 using may result in inf values and large absolute errors
+            # when used with large numbers
+            if np.any(value > 1e4):
+              print("fp16 using may result in inf values and large absolute "
+                    "errors when used with large numbers, skipping")
+              continue
+
+          self.assertAllClose(expected, value.flatten(), atol=tol,
+                              rtol=tol)
```
reedwm: Before, fp32 rtol was 1e-6, but now its 1e-5. It should be changed back to 1e-6 for fp16.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -80,23 +90,42 @@ def _SetupValuesForDevice(self, tensor_in_sizes, filter_in_sizes, stride,
...
```
reedwm: Please add gradient tests. See the ConstructAndTestGradient function in this file.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -80,23 +90,42 @@ def _SetupValuesForDevice(self, tensor_in_sizes, filter_in_sizes, stride,
...
+
       with self.test_session() as sess:
-        values = sess.run(results)
-        for value in values:
+        tensor_results = [result[0] for result in results]
+        test_params = [result[1:] for result in results]
+        values = sess.run(tensor_results)
+        for i in range(len(tensor_results)):
+          value = values[i]
+          print("use_gpu:", test_params[i][0])
```
reedwm: This print statements are unnecessary, so remove them.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
reedwm: I don't think we have enough tests due to the fact we skip large numbers. For example, testConv3D2x2x2FilterStride2, testConv3DStride3, and testConv3D2x2x2FilterStride2Same are not run at all with fp16. Can you add a few more tests, testing the sorts of things that are currently skipped, that are run with fp16?

reedwm (changes requested):

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: Ok. But firstly I need to figure out which absolute errors are appropriate in different cases.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: Ok, thanks for mention.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: There is already normalized initial values => small absolute errors. 
I copied code from conv2d tests. 
There was code with float64 in tests but I doubt that it's actually was used (is there fp64 version of conv3d?) so I removed it.

opensourcemattress: Sorry, fp64 was really declared for CPU.

opensourcemattress:
@reedwm, what should I do next?

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -328,50 +356,64 @@ def _ConstructAndTestGradientForConfig(
...
           filter_data, shape=filter_shape, dtype=data_type, name="filter")
 
-      if data_format == "NCDHW":
-        input_tensor = test_util.NHWCToNCHW(orig_input_tensor)
-        strides = test_util.NHWCToNCHW(strides)
-      else:
-        input_tensor = orig_input_tensor
+        if data_format == "NCDHW":
+          input_tensor = test_util.NHWCToNCHW(orig_input_tensor)
+          strides = test_util.NHWCToNCHW(orig_strides)
```
reedwm: Remove the line "orig_strides = strides[:]" above, and change this to "new_strides = test_util.NHWCToNCHW(strides)". Then use new_strides instead of strides.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -328,50 +356,64 @@ def _ConstructAndTestGradientForConfig(
...
           data_format=data_format, name="conv")
 
-      if data_format == "NCDHW":
-        conv = test_util.NCHWToNHWC(conv)
+        if data_format == "NCDHW":
+          conv = test_util.NCHWToNHWC(conv)
+
+        
+        if test_input:
+          print("test_input")
```
reedwm: Remove this print statement.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -81,22 +94,37 @@ def _VerifyValues(self, tensor_in_sizes, filter_in_sizes, stride, padding,
...
           print("expected = ", expected)
           print("actual = ", value)
-          self.assertAllClose(expected, value.flatten(), atol=tolerance,
-                              rtol=1e-6)
+          tol = 1e-6
+          if value.dtype == np.float16:
+            tol = 1e-3
+            # fp16 using may result in inf values and large absolute errors
+            # when used with large numbers (fp16 max is 65004.)
+            if np.any(value > 1e4):
```
reedwm: As I stated before, this will skip some important tests, causing some things not to be tested. testConv3D2x2x2FilterStride2, testConv3DStride3, and testConv3D2x2x2FilterStride2Same and not run at all with this, so you should add more tests.

@mjanusz, how were the expected outputs calculated for the original tests? It is a pain to add new tests since new expected outputs need to be calculated.

reedwm (changes requested):

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: I made normalization of input and supposed  that current fp64 convs implemented correctly => I took them as reference value. It's not a good idea to test fp64 outputs by themselves but I not sure how to solve this problem in different way. 
It's better just to have reference implementation and just compare outputs but it's additional code and we need more tests for tests, yes? :)

drpngx:
Jenkins, test this please.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -56,11 +69,11 @@ def _SetupValuesForDevice(self, tensor_in_sizes, filter_in_sizes, stride,
 
     # Initializes the input tensor with array containing incrementing
```
reedwm: Can you update this comment? Something like "Initializes the input tensor with array containing numbers from 0 to 1. We keep the input tensor values fairly small to avoid overflowing a float16 tensor during the conv3d."

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -81,27 +94,36 @@ def _VerifyValues(self, tensor_in_sizes, filter_in_sizes, stride, padding,
...
+            stride,
+            padding,
+            data_format,
+            dtype,
+            use_gpu=use_gpu)
+        results.append(result)
+
       with self.test_session() as sess:
-        values = sess.run(results)
+        tensor_results = [result[0] for result in results]
```
reedwm: Why is this line necessary? Why take the first element for each result?

reedwm (changes requested):

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: Done.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: Sorry, probably it was used for manual comparison of results. 
I fixed it.

tensorflow/python/kernel_tests/conv_ops_3d_test.py
```diff
@@ -45,22 +47,34 @@ def GetTestConfigs():
...
     for s in filter_in_sizes:
       total_size_2 *= s
 
-    # Initializes the input tensor with array containing incrementing
-    # numbers from 1.
-    x1 = [f * 1.0 for f in range(1, total_size_1 + 1)]
-    x2 = [f * 1.0 for f in range(1, total_size_2 + 1)]
+    # Initializes the input tensor with array containing numbers from 0 to 1.
+    # We keep the input tensor values fairly small to avoid overflowing a float16 
+    # tensor during the conv3d and to keep absolute errors small
```
reedwm: Why does it matter that the absolute error is small? If the absolute error were larger, you can adjust what you pass to atol in self.assertAllClose.

Also, I just realized that you didn't modify what you pass to atol in self.assertAllClose. Can you make it a bit smaller to take into account the expected output values are smaller?

reedwm (changes requested):

tensorflow/python/kernel_tests/conv_ops_3d_test.py
opensourcemattress: >Why does it matter that the absolute error is small?

Because if I understand correctly we compare results with reference using absolute difference. And absolute error depend on current significant and exponent. Large numbers (huge exponent) and small amount of bit for significant (fp16) => big absolute errors (I seen difference >10 with huge numbers in some fp16 tests).

I can make 'tolerances' smaller but not sure which values are appropriate. It's hard to tell because convs are not 'pure convs' but some tricky methods from computation mathematics. Error depends on particular implementation of convolution and if we want to be really sure it's better to find out what are relative/absolute errors for particular method and particular type. But I think that from practical point of view it's better to take 'small enough' values and continue to work on other problems.

reedwm: Sorry for late response. I see your point about making 'tolerances' smaller, so you can leave atol as is.

I think you should remove the "and to keep absolute errors small" comment. You're right that having a large exponent with fp16 means large absolute errors, but I don't see that as a problem. We could have just raised 'atol' in that case. Differences greater than 10 seem reasonable given how imprecise fp16 is. However, currently, some expected_outputs are above the maximum fp16 value of 65504, which would cause them to overflow. This seems like the only reason to reduce the numbers in the input tensor IMO.

reedwm (approved):

reedwm:
Jenkins, test this please.

opensourcemattress:
Thanks for review!

reedwm:
Jenkins, test this please.

PR #12836, opened 2017-09-06
memory mapped file

cloudbank:
This uses a FileChannel to map the model file for slightly faster IO. Because of the difficulty with assets in android, the files must be copied outside of the apk first and then read in.  I did not include that part, deciding it would be the onus of the parent application.   ctx.getFileStreamPath(model) is called in a static factory alongside what is already there for backward compatibility.  This means files must have been copied to the internal storage per ctx.getFilesDir().   Also it passed in the Application for context, taking care to recreate it as needed for reclaimed memory.
This is a preliminary example of what could be changed.  I am disappointed that the mmap implementation in Java is very limited.  The byte[] of array() is not written to work with 
MappedByteBuffer oddly, and asking for loaded state with load()  can kick off additional IO calls.
It is working in my library but I will also continue to improve on it. I  also plan to add a native jni version as well which makes use of the mmap util to see which runs faster.
I backed out my static changes and will see if that works in another PR.


yifeif:
@cloudbank thanks for the PR. Do you mind resolving the conflicts?

cloudbank:
Sure

Best regards,

Sabine G. Vogel

cloudbank:
@yifeif  I overloaded the constructor and loadGraph method instead of introducing something new.  My static factory version runs slightly faster, but this works pretty well, too.  I may add another PR to introduce the static usage.

PR #12837, opened 2017-09-06
Add SONAME for Makefile-built libtensorflow_inference.so

resec:
Currently if we build our own shared library that depends on Makefile-built Tensorflow shared library, our built library is linking the Tensorflow library with **absolute path**. So added a `SONAME` to Makefile-built Tensorflow shared library to resolve this problem.


yifeif:

yifeif (approved):

PR #12885, opened 2017-09-07
mmap file

cloudbank:
I added a memory mapped file option for the model file.  Since the model files have to be copied out of the apk due to assets directory readonly status and inability to create the FileChannel directly, it is assumed that you have done so to internal storage with the same context that is passed into the overloaded constructor.  Internal storage eliminates problems of SD card not being present or M runtime permissions requests but is also in danger of being full etc.  Expansion files are the only other option, but this requires a download from the play store, which is an extra step.  I therefore did the most simple thing.

### DIFF
--- a/tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
+++ b/tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
@@ -15,21 +15,13 @@ limitations under the License.
 
 package org.tensorflow.contrib.android;
 
+import android.content.Context;
 import android.content.res.AssetManager;
 import android.os.Build.VERSION;
 import android.os.Trace;
 import android.text.TextUtils;
 import android.util.Log;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.nio.ByteBuffer;
-import java.nio.DoubleBuffer;
-import java.nio.FloatBuffer;
-import java.nio.IntBuffer;
-import java.nio.LongBuffer;
-import java.util.ArrayList;
-import java.util.List;
+
 import org.tensorflow.DataType;
 import org.tensorflow.Graph;
 import org.tensorflow.Operation;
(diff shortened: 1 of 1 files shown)


cloudbank:
I could not squash my commits so I created a new request with a clean history.

yifeif:
@andrewharp could you take a look or reassign? Thank you!

cloudbank:
@andrewharp if you want some code that uses it and copies files initially,  I am working on a library  which is using the TensorFlowImageClassifier. I see the comment:
TODO(andrewharp): make this handle non-assets.
If you have ideas I have not mentioned about where to hold the model files, I would be flexible about working on some other solutions too.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -15,40 +15,114 @@
 
 package org.tensorflow.contrib.android;
 
+import android.content.Context;
```
andrewharp: Please revert the import reordering

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -118,6 +192,7 @@ public TensorFlowInferenceInterface(InputStream is) {
    *
    * @param outputNames A list of output nodes which should be filled by the inference pass.
    */
+
```
andrewharp: Can you make a pass and delete errant blank lines like this please? No blank lines are necessary after comments or after opening braces or before closing braces, and a single blank line is preferred after closing braces (depending on context).

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -458,6 +537,39 @@ private void prepareNativeRuntime() {
...
+    try {
+      g.importGraphDef(graphDef);
+    } catch (IllegalArgumentException e) {
+      throw new IOException("Not a valid TensorFlow Graph serialization: " + e.getMessage());
+    }
+
+    if (VERSION.SDK_INT >= 18) {
+      Trace.endSection(); // importGraphDef.
+      Trace.endSection(); // initializeTensorFlow.
```
andrewharp: This seems to match up with "readGraphDef", not "initializeTensorFlow".

In general the tracing is a good idea, but it should probably be added consistently to the entire interface in a separate PR.

andrewharp:
@cloudbank That TODO refers to the label file for the demo specifically. Since that's just example app code (under examples/android) rather than library code (under contrib/android) it felt less urgent to generalize it.

I would not be opposed to having the TFII class do a check and copy the file from assets to a specified tmp dir if not already present, if we can show that this sufficiently speeds up subsequent runs.

cloudbank:
@andrewharp  The worry point is when to delete the file that was copied. Although this does not affect the buffer, obviously the mapped file won't be there forever either. Ideally, this heavy create operation involving the big graph file is done off heap avoiding IO exactly once and the classifier object persisted for further usage. Of course all of this is taken care of outside of the TFII in my work.   I certainly could add copy and delete methods if you insist but would not want to encourage repeated copy, or for that matter repeated create.  I would love to work on adding persistence and feel safe to copy/delete once and then null out the buffer.  
I am fixing the typo and whitespace edits.  Tell me what you think about the delete.

PR #12886, opened 2017-09-07
Fixed #12797: uint8 now has a length

### LINKED ISSUE #12797: UINT8 is used but only partly supported in the Java API.
The datatype uint8 exists in the Java API (see DataType.UINT8) and is used by the LabelImage program. However, it does not seem to be fully supported. In particular, elemByteSize() in tensor_jni.cc does not have a case for TF_UINT8, which prevents uint8 tensors from being created or extracted from TensorFlow.

andrewcmyers:
This small change allows tensors of uint8 to be constructed from the Java API. I am leaving the new test case of it over in the phase2 PR (#11535) because that test case relies on other changes to Tensor.create.


asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please

Issue #12857, opened 2017-09-06
Unhelpful exceptions from tf.truncated_normal with dtype=tf.int32 


slacy:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: NO
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 14.04 LTS 
- **TensorFlow installed from (source or binary)**: pip install tensorflow 
- **TensorFlow version (use command below)**: ('v1.3.0-rc2-20-g0787eee', '1.3.0')
- **Python version**: Python 2.7.6
- **Bazel version (if compiling from source)**: N/A
- **CUDA/cuDNN version**: N/A (CPU)
- **GPU model and memory**: N/A 
- **Exact command to reproduce**:

python -c "import tensorflow as tf ; tf.truncated_normal([1], dtype=tf.int32)"

### Describe the problem
When attempting to initialize a tf.Variable of type tf.int32 using ```tf.truncated_normal()```, invocation with simple args raise a TypeError, but the error message is ambiguous and unhelpful for debugging: 

```
python -c "import tensorflow as tf ; tf.truncated_normal([1], dtype=tf.int32)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/slacy/src/tf_clean/env/local/lib/python2.7/site-packages/tensorflow/python/ops/random_ops.py", line 168, in truncated_normal
    mean_tensor = ops.convert_to_tensor(mean, dtype=dtype, name="mean")
  File "/usr/slacy/src/tf_clean/env/local/lib/python2.7/site-packages/tensorflow/python/framework/ops.py", line 611, in convert_to_tensor
    as_ref=False)
  File "/usr/slacy/src/tf_clean/env/local/lib/python2.7/site-packages/tensorflow/python/framework/ops.py", line 676, in internal_convert_to_tensor
    ret = conversion_func(value, dtype=dtype, name=name, as_ref=as_ref)
  File "/usr/slacy/src/tf_clean/env/local/lib/python2.7/site-packages/tensorflow/python/framework/constant_op.py", line 121, in _constant_tensor_conversion_function
    return constant(v, dtype=dtype, name=name)
  File "/usr/slacy/src/tf_clean/env/local/lib/python2.7/site-packages/tensorflow/python/framework/constant_op.py", line 102, in constant
[...]

PR #12887, opened 2017-09-07
Rearranging statements

printdhruv:
https://github.com/tensorflow/tensorflow/issues/12857

Highlights:-
It generates informative error.

### DIFF
--- a/tensorflow/python/ops/random_ops.py
+++ b/tensorflow/python/ops/random_ops.py
@@ -165,11 +165,11 @@ def truncated_normal(shape,
   """
   with ops.name_scope(name, "truncated_normal", [shape, mean, stddev]) as name:
     shape_tensor = _ShapeTensor(shape)
-    mean_tensor = ops.convert_to_tensor(mean, dtype=dtype, name="mean")
-    stddev_tensor = ops.convert_to_tensor(stddev, dtype=dtype, name="stddev")
     seed1, seed2 = random_seed.get_seed(seed)
     rnd = gen_random_ops._truncated_normal(
         shape_tensor, dtype, seed=seed1, seed2=seed2)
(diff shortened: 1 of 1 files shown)


yifeif (approved):

yifeif:

printdhruv:
@yifeif Any suggestions for `tensorflow/contrib/layers/python/layers/optimizers_test.py`? My local machine is passing all test cases for that.

yifeif:
@honkentuber could you help taking a look? Thanks!

honkentuber:
+wicke,ispir

I'm not on ML any more.

martinwicke:
It appears that you changed the behavior, probably by changing the order of ops. @aselle do you see why that might be? It may be a question if simply updating the tests. The second failure looks a little dangerous though.

printdhruv:
@martinwicke Yes.I am changing the flow.

aselle:
Essentially, the order the graph is constructed will change the random number seeding, even if the number of ops and connectivity is the same. That test should be uneatable in the baseline.

printdhruv:
@aselle what should be your guidelines to achieve tf.int32 error message as expected ?
By default, mean is invoked hence I changed the order.
Should I check it in truncated_normal() function itself and raise error for invalid dtype?

ispirmustafa:
Why does this change generates informative error?

gunan:
Should we close this PR?

printdhruv:
@gunan As,it goes against the random number seeding pointed out by aselle , this fix won't work. The simplest thing would be to brute force checking for this function or to change error message which will be affect all connected components.

martinwicke:
Ok, so close it? Fix the tests? @aselle? @yifeif?

gunan:
I am closing this PR, as now it is clear that this is not the way we would like to fix this.
We can explore the other approach in another PR.

PR #12890, opened 2017-09-07
MKL build broken / Proposed fix here

EdwardDixon:
The CPU version of TensorFlow wasn't building because the MKL was broken; it was broken because a merge conflict somehow made it all the way through to here.  See diff (!).


EdwardDixon:
I signed it

yifeif (approved):

Issue #12895, opened 2017-09-08
dlopen error when using TF_LoadLibrary


eaplatanios:
I'm trying to use `TF_LoadLibrary` on a mac and I keep getting errors such as:
```
Caused by: org.platanios.tensorflow.jni.TensorFlow$NativeException: dlopen(/var/folders/rw/lqrc8nk52kqcc4_zq3c2b6dh0000gn/T/tensorflow_scala_native_libraries7467192138547554485/libtensorflow_ops.dylib, 6): Symbol not found: __ZN10tensorflow15shape_inference12UnknownShapeEPNS0_16InferenceContextE
  Referenced from: /var/folders/rw/lqrc8nk52kqcc4_zq3c2b6dh0000gn/T/tensorflow_scala_native_libraries7467192138547554485/libtensorflow_ops.dylib
  Expected in: flat namespace
 in /var/folders/rw/lqrc8nk52kqcc4_zq3c2b6dh0000gn/T/tensorflow_scala_native_libraries7467192138547554485/libtensorflow_ops.dylib
```
The same thing happens when I try to load the compiled dylib file through Python directly. It was compiled it using `-D_GLIBCXX_USE_CXX11_ABI=0` and `-undefined dynamic_lookup` for the linker. @asimshankar

asimshankar:
Unfortunately, loading custom op libraries from the C API isn't fully functional now
(and worked around using [`RTLD_GLOBAL` in Python](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/python/pywrap_tensorflow.py#L40)).

The good news is that @allenlavoie is working on a fix. No immediate ETA, but long story short - loading custom op libraries might not work outside of Python yet, but we're on it.

(Related issue #10454)

eaplatanios:
@asimshankar That's interesting because I also tried to load it using RTLD_GLOBAL. However, on my mac, even `tf.load_op_library(...)` in Python is not working and is throwing the same error.

eaplatanios:
@asimshankar I found out that it's not just RTLD_GLOBAL that matters. The symbols are actually exported through `_pywrap_tensorflow_internal.so` in the Python API. I was only loading `libtensorflow.so` with RTLD_GLOBAL, but that is not enough to make things work. Is there any easy workaround that you can see, until the fix comes out? Thanks! :)

eaplatanios:
[...]

PR #12915, opened 2017-09-08
Load op library fix

eaplatanios:
This fixes the issue with external bindings to TensorFlow loading custom op libraries. It does so by exporting all the relevant symbols in `libtensorflow.so`. Note that these symbols were there but they were hidden and we could not link to them previously. Currently, with this fix, developers of custom op libraries can link at compile time using `libtensorflow.so`, thus avoiding the need for `-undefined -dynamic_lookup` on Macs (which also currently doesn't work as a solution on its own) and the need for loading with `RTLD_GLOBAL` (which also does not work on its own, given that `pywrap_tensorflow.so` needs to be loaded with that flag, given it's the own exporting the necessary symbols). I've tested this on my mac and it resolves the issue.

@asimshankar I know this is not really a good solution given that you probably don't want to export everything in `libtensorflow.so`, but I think it offers a good temporary workaround to the issue, until  the fix @allenlavoie is working on, is published.

P.S. This issue has been discussed a bit in #12895.


asimshankar:
Thanks @eaplatanios : but I'd really like to avoid this for now.  This restriction was introduced in https://github.com/tensorflow/tensorflow/pull/9059/commits/b0989a0280971034bd838fa176ae3f92b210e8dc for two reasons: (1) binary size and (2) clearly defining the API.

There is a reasonable chance that @allenlavoie 's more principled fix will make it to release 1.5 (if not earlier in 1.4). So paying this temporary technical debt doesn't seem worth it. (The workaround won't be helpful to any users of the release till the next release anyway, and those building from source can do this in their build)

(Though, I'll sync with Allen - if I'm wrong about his fix being enough, then will revisit)

eaplatanios:
@asimshankar That sounds reasonable! I agree that it's better to have a clearly defined API. In general, modularization of TensorFlow and it's binaries would probably be a good idea.

As for a temporary workaround, I only need this for my Scala API for defining an op equivalent to `py_func`. One way to go about this would be to make `pywrap_tensorflow.so` accessible through the nightly CI builds. Is that something that's easy to do? Then, I can simply also package it with `libtensorflow.so` for my current purposes, until a more elegant alternative is available.

P.S. I refer to the CI builds because my packaging/release script right now, pulls the `libtensorflow.so` binaries from there.

asimshankar:
Great, thanks for your understanding (I'll go ahead and close this PR).

You could pull `pywrap_tensorflow.so` from the nightly builds by downloading the `whl` file and extracting it :)

eaplatanios:
No problem! :)

I checked out [http://ci.tensorflow.org/view/Nightly/](http://ci.tensorflow.org/view/Nightly/) for the nightly whl files but I can't find them. I only see one build for linux-python3 but that doesn't seem to include the pywrap shared object. Could you point me to a download link for one of those whl files?

asimshankar:
See: https://github.com/tensorflow/tensorflow#installation

allenlavoie:
And the .so itself should be at something like tensorflow-1.3.0.data/purelib/tensorflow/python/_pywrap_tensorflow_internal.so once you unzip the .whl.

eaplatanios:
Great! Thanks a lot! :)

PR #12918, opened 2017-09-08
Corrected description of what padding SAME does

erlichson:
The description read that the padding=SAME adds zeros to the output tensor. It should say that it pads the input tensor.


hawkinsp (approved):
Looks good, thanks for the pull request!

PR #12919, opened 2017-09-08
Changed number to float in documentation example "Get Started" in order to be consistent

sauercrowd:
Stumbled upon this while going through the "Get Started".
added a dot to the `7` to make clear it's a float (like every other number)


yifeif (approved):

yifeif:
Thanks @sauercrowd!

PR #12920, opened 2017-09-08
Workaround for NVCC 9.0 internal error

nluehr:
Fixes 'Internal Compiler Error (codegen)' encountered when building with nvcc 9.0.

### DIFF
--- a/tensorflow/core/kernels/reduction_gpu_kernels.cu.h
+++ b/tensorflow/core/kernels/reduction_gpu_kernels.cu.h
@@ -266,7 +266,9 @@ __global__ void ColumnReduceMax16ColumnsKernel(
   if (row * num_cols + col < num_rows * num_cols)
     sum = in[row * num_cols + col];
 
-  __shared__ value_type partial_sums[32][33];
+  // 1D array necessary due to bug in CUDA 9 compiler.
+  // TODO(nluehr) revert to 2D array when compiler is ready.
+  __shared__ value_type partial_sums[32 * 33];
 
   row += rows_per_warp * gridDim.y * blockDim.y;
   for (; row < num_rows; row += rows_per_warp * gridDim.y * blockDim.y) {
@@ -283,16 +285,16 @@ __global__ void ColumnReduceMax16ColumnsKernel(
     if (lane < num_cols) sum = op(sum, tmp);
   }
 
-  if (lane < num_cols) partial_sums[lane][threadIdx.y] = sum;
+  if (lane < num_cols) partial_sums[lane * 33 + threadIdx.y] = sum;
 
   __syncthreads();
 
   if (threadIdx.y == 0 && threadIdx.x < num_cols) {
-    value_type s = partial_sums[threadIdx.x][0];
+    value_type s = partial_sums[threadIdx.x * 33];
 
     if (blockDim.y > 1) {
       for (int row = 1; row < blockDim.y; ++row) {
-        s = op(s, partial_sums[threadIdx.x][row]);
+        s = op(s, partial_sums[threadIdx.x * 33 + row]);
       }
     }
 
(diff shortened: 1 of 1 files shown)


tensorflow/core/kernels/reduction_ops_gpu_kernels.h
```diff
@@ -238,7 +238,10 @@ __global__ void ColumnReduceMax16ColumnsKernel(
   if (row * num_cols + col < num_rows * num_cols)
     sum = in[row * num_cols + col];
 
-  __shared__ value_type partial_sums[32][33];
+  // 1D array necessary due to bug in CUDA 9 compiler.
+  //__shared__ value_type partial_sums[32][33];
```
ekelsen: Can we change this to a TODO when the bug is fixed?  (Here and similar places)

How will we know when the bug is fixed?  Not until CUDA 10?  Does this bug break all 2D shared arrays?

Also, I think you need need a space after the // or the linter will complain.

tensorflow/core/kernels/reduction_ops_gpu_kernels.h
```diff
@@ -717,11 +728,11 @@ struct ReduceFunctor<GPUDevice, Eigen::internal::SumReducer<T>> {
   static void Reduce(OpKernelContext* ctx, OUT_T out, IN_T in,
                      const ReductionAxes& reduction_axes,
                      const Eigen::internal::SumReducer<T>& reducer) {
-    ReduceImpl<T, cub::Sum, T*, T*, ReductionAxes>(
+    ReduceImpl<T, Eigen::internal::scalar_sum_op<T>, T*, T*, ReductionAxes>(
```
ekelsen: Are these changes for a different compiler problem?  Are other ops affected, or only sum?

ekelsen (changes requested):

tensorflow/core/kernels/reduction_ops_gpu_kernels.h
nluehr: These correct a separate issue also related to reductions with std::complex types. cub::Sum relies on an operator overload for operator+, this is not available (in device code) for std::complex. This results in a compiler warning about calling "__host__" functions from "__host__ __device__ functions, and incorrect code.

tensorflow/core/kernels/reduction_ops_gpu_kernels.h
nluehr: Not all 2D shared arrays are affected. As far as I can tell, only those whose types rely on constexpr default constructors enabled by NVCC's -expt-relaxed-constexpr option.

tensorflow/core/kernels/reduction_ops_gpu_kernels.h
ekelsen: There are already workarounds for complex product and division; sum worked fine, but I guess is broken in nvcc 9?

Can you either change this fix to be consistent with the existing work arounds, or change the existing work arounds to this style?

nluehr: I made the code consistent with the existing workarounds.

cuda9 does have different behavior than cuda8. This may well be related to the same expt-relaxed-constexpr bug.

tensorflow/core/kernels/reduction_ops_gpu_kernels.h
```diff
@@ -255,16 +287,19 @@ __global__ void ColumnReduceMax16ColumnsKernel(
     if (lane < num_cols) sum = op(sum, tmp);
   }
 
-  if (lane < num_cols) partial_sums[lane][threadIdx.y] = sum;
+  // if (lane < num_cols) partial_sums[lane][threadIdx.y] = sum;
```
ekelsen: please remove the commented lines - just switch back when ready.

ekelsen:
Just one minor comment.  Unfortunately some internal code just changed that will interact with this code.  I'm going to hold off on approval so that the merge can be resolved here rather than when trying to pull internally.

nluehr:
@ekelsen is there an ETA on when the internal changes will be published so that we can complete this PR?

ekelsen:
Sorry the internal changes are done and the fix for the sum compiler bug is already integrated.

nluehr:
Great. I already rebased to incorporate those changes. So hopefully this PR is ready to merge.

ekelsen (approved):

sb2nov:
Jenkins, test this please.

PR #12922, opened 2017-09-08
mmap file

cloudbank:
I added a memory mapped file option for the model file. Since the model files have to be copied out of the apk due to assets directory readonly status and inability to create the FileChannel directly, it is assumed that you have done so to internal storage with the same context that is passed into the overloaded constructor. Internal storage eliminates problems of SD card not being present or M runtime permissions requests but is also in danger of being full etc. Expansion files are the only other option, but this requires a download from the play store, which is an extra step. I therefore did the most simple thing.


cloudbank:
@andrewharp The worry point is when to delete the file that was copied. Although this does not affect the buffer, obviously the mapped file won't be there forever either. Ideally, this heavy create operation involving the big graph file is done off heap avoiding IO exactly once and the classifier object persisted for further usage. Of course all of this is taken care of outside of the TFII in my work. I certainly could add copy and delete methods if you insist but would not want to encourage repeated copy, or for that matter repeated create. I would love to work on adding persistence and feel safe to copy/delete once and then null out the buffer.
I am fixing the typo and whitespace edits. Tell me what you think about the delete.

andrewharp (commented):
Please undo the reformatting here so that diffs are easier to see.

2 spaces for new nesting levels, 4 for continued lines

Generally speaking, no blank lines after opening braces or before closing braces, and only 1 blank line max between code blocks.

cloudbank:
I am using the google standard template to reformat code of 2 indent, 4 continuation, 8 tab in AS.  Line 82 where it adds/deletes a line is an example that I cannot seem to get rid of.  If you know what is happening I would appreciate a suggestion as to get rid of that behavior.

cloudbank:
@andrewharp  Did you review the changes?

cloudbank:
@andrewharp I changed it to copy the file from res/raw to cache temporarily then map it and delete it the copy.  Even with the copy, this version takes the create to **.6 seconds on average,** from 1.6 seconds that I previously benchmarked using the assets/ and IO.  It requires that you pass in the model file without the pb extension as a  resource id and of course store the file in res/raw.

cloudbank:
@yifeif  Hi: I changed my solution but it seems Andrew is busy. Is it possible to re-assign it?

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -46,9 +53,99 @@
  * example usage.
  */
 public class TensorFlowInferenceInterface {
-  private static final String TAG = "TensorFlowInferenceInterface";
```
andrewharp: Please revert

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -79,14 +176,14 @@ public TensorFlowInferenceInterface(AssetManager assetManager, String model) {
...
+
     try {
       if (VERSION.SDK_INT >= 18) {
         Trace.beginSection("initializeTensorFlow");
         Trace.beginSection("readGraphDef");
       }
-    
-      // TODO(ashankar): Can we somehow mmap the contents instead of copying them?
+
+     // TODO(ashankar): Can we somehow mmap the contents instead of copying them?
```
andrewharp: alignment issue

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -46,9 +53,99 @@
...
+        //@todo individual exceptions
+        fileChannel.close();
+      } catch (IOException e) {
+        throw new RuntimeException("Failed to close RAF '" + fileChannel + "'", e);
+      }
+      if (buffer != null) {
+        buffer.clear();
+        buffer = null;
+      }
+
```
andrewharp: extra blank line

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -46,9 +53,99 @@
...
+    this.g = new Graph();
+    this.sess = new Session(g);
+    this.runner = sess.runner();
+    this.modelName = String.valueOf(model);
+    MappedByteBuffer buffer = null;
+    FileChannel fileChannel = null;
+    RandomAccessFile raf = null;
+    File cacheFile;
+    try {
+      cacheFile = createCacheFile(ctx, model, "delete-me-please");
```
andrewharp: Why not use the model's filename?

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -46,9 +53,99 @@
...
+   * Load a TensorFlow model via FileChannel mmap
+   * This requires that assets files are copied to internal storage already
+   * getFileStreamPath() points to that directory; context used must be passed in here
+   *
+   * @param ctx The context to use to load the model file; same as one used to copy files
+   * @param model The filepath to the GraphDef proto representing the model.
+   */
+  public TensorFlowInferenceInterface(Context ctx, int model) {
+    prepareNativeRuntime();
+    File mModelFile = null;
```
andrewharp: TF style is not to use type prefixes on variable names.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -46,9 +53,99 @@
...
+  /**
+   * Copies raw resource to a cache file.
+   *
+   * @return File reference to cache file.
+   */
+  private File createCacheFile(Context context, int resourceId, String filename)
+      throws IOException {
+    File cacheFile = new File(context.getCacheDir(), filename);
+
+    if (cacheFile.createNewFile() == false) {
```
andrewharp: If the file is already present, you could check that size and timestamp of the file matches the one from assets, and if so just avoid copying altogether.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
cloudbank: do you mind if I just delete it? since I am addressing it.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
cloudbank: Remember I copy from res/raw not assets. It is faster that way but still suffers from the fact that aapt compresses it.  To get its original size for comparison I have to copy it again which defeats the purpose of this check.  I delete the file at the end.  It is just for safety's sake that the file check is there in case somehow another file got written with the same name.

andrewharp: What is the advantage of res/raw over assets/?

It's possible to explicitly keep aapt from compressing extensions, e.g. with [noCompress](https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.AaptOptions.html#com.android.build.gradle.internal.dsl.AaptOptions:noCompress) in Gradle. I think passing the --uncompressedExtensions flag to bazel might do the same thing.

A properly trained model will probably be very high entropy anyway, so typically you won't see a lot of compression savings. And in fact I just checked the demo -- found that none of the .pb files in it are compressed at all, simply stored. In that case, can we just mmap it from assets/ directly without the copy?

cloudbank: res/raw is indexed in the R file and claims on SO to be faster.  I did not benchmark it to prove it.
Of course I could not get a map with compression so the assets would *have* to be uncompressed, which might be hard to enforce.
Right now I am trying to test if a filedescriptor can be used on uncompressed asset to make a FileInputStream--the only way I have not yet tried without copy.

cloudbank: res/raw is indexed in the R file and claims on SO to be faster. It does come back faster than the assets.
Of course I could not get a memory map with compression so the assets would have to be uncompressed for a direct map, which unfortunately, creates an OutOfMemoryError, when trying to open from a FileInputStream. The (not retrained) inception file is 50MB which is half your allowance on Android, for instance. It makes little sense to try to do it that way for this reason. People want to use it in apps, not just stand alone and will run out of memory fast.
I will push the file again with the other corrections as per above

cloudbank:
@andrewharp this version works best in terms of average speed.

cloudbank:
@yifeif  Can you give me a hint as to how I work this through with @andrewharp  to get it finished?  I don't know what he wants in addition to what I have changed.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
andrewharp: My major reservation is that this commit only seems to speed up loading from res. It might be nice to support this functionality, but up until this point we've suggested assets/ as the place to place pb files. So it will be a bit more difficult for users to switch to the new method, and possibly confuse things unless well documented. How much faster is it actually to copy from res? Also, why does it give you an OOM error when mapping from the cache dir, but not from assets/?

It should be possible to mmap files directly from assets with the nocompressextensions directive in bazel. In which case the copy is not even necessary. Can you tell if this is the case?

andrewharp:
@cloudbank Apologies for the delay, up against some deadlines internally.

cloudbank:
@andrewharp NP.  I don't mind at all as long as you eventually get back to
me.  : )


Best regards,

Sabine G. Vogel

cloudbank:
@andrewharp Do you have any estimate (next week?) when you will be able to finish this?

andrewharp:
@cloudbank I still have reservations about limiting this to a res/raw implementation, would like the questions mentioned before clarified:

- Does noCompressExtensions not solve the compression problem for assets/?
- Why does it give you an OOM error when mapping from the cache dir, but not from assets/?

cloudbank:
@andrewharp

-no compression either using gradle or bazel leads to OOM.  It is
unreasonable I think to not make every effort to compress such a large file
in a 100MB limit and kind of violates best practices.  I tried to map the
uncompressed file directly and ran out of memory right away.  I am building
the java and native libraries with bazel and then using them in a gradle
built app.  We want people to build stuff with it, right?
-using the res/raw and cache files dir doesn't run out of memory.  It might
run a little faster than using assets and cache since res are indexed.  I
will do a benchmark for both and get back to you.  If it is not a big
difference and you really want assets dir I can move it back there.
-I have not given up on trying to map the file without copying but
  1. every SO thread, and advice about this says it is not possible from
expert level people including Pete Warden and Mark Murphy.
  2. I have tried every reasonable approach without getting into a hack so
it might be best to start with this approach and if I, or someone else
stumbles upon a better solution, it can be improved.  Given the natural way
android is set up right now it seems improbable that copying can be avoided.

Best regards,

Sabine G. Vogel

cloudbank:
@andrewharp I benchmarked it and the res/raw solution comes back on average in just under a second with a best time average around .5 seconds.  The assets solution never comes back any where near that fast on average, with around 1.2-1.3seconds.  Could you live with the idea that if you want to use the overloaded create that uses a memmap, you need to move your files over to res/raw?  Not that big a deal.  You also have to remove the file extension, and I could add a check for that.

cloudbank:
@yifeif Can you give me a hint as to how I work this through with @andrewharp to get it finished? I don't know what he wants in addition to what I have changed.

yifeif:
@cloudbank we always try to look at the PRs as soon as possible, but sometimes we might also have internal deadlines to meet and can have some delay. Thanks again for your patience.

@andrewharp could you let @cloudbank know whenever you get a chance? Thanks a lot!

andrewharp:
@cloudbank I'm not sure I understand why no compression leads to OOM. Where is the 100mb limit coming from? For a properly trained model, the entropy should be high enough that you won't get any significant compression gains anyway. For reference this is the unzip -v output for inception in tensorflow_demo.apk:

```
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
 3771239  Stored  3771239   0% 1980-01-01 00:00 9225f2cc  assets/conv_actions_frozen.pb
29083865  Defl:X 25434578  13% 1980-01-01 00:00 ac57f5c3  assets/ssd_mobilenet_v1_android_export.pb
  563897  Defl:X   257084  54% 1980-01-01 00:00 7d7f4ff3  assets/stylize_quantized.pb
53884595  Stored 53884595   0% 1980-01-01 00:00 6d3b8d15  assets/tensorflow_inception_graph.pb

```
Compression was not deliberately turned off here, zip just decided it wasn't worth it for 2/4 graphs. The savings from the other two are very minimal overall. 

What I think might be ideal would be to have the res/raw path as you have it now, and also the assets/ path if the user has not chosen to compressed the pb file. Then the user automatically gets the improvement if they set things up the default way with assets (e.g. the default classifier demo will be improved), and get even more speed you've found if they take the time to move things to res/raw.

cloudbank:
@andrewharp
I was just thinking to change it to detect compressed or not myself...great
idea. I will change it over the next few days.

Best regards,

Sabine G. Vogel

martinwicke:
@cloudbank any update?

cloudbank:
working on it.....sorry have been busy with work...
I really want to finish it soon, so I will carve out some time ASAP.

Best regards,

Sabine G. Vogel

cloudbank:
@andrewharp prior to Android 2.3, any compressed asset file with an uncompressed size of over 1 MB cannot be read from the assets, but then who uses prior to 2.3?  I am using gradle with compression and both my graphs are compressedly by 25 and 53 % in the apk.  I do not see any way to control whether or not the files are to be compressed---even if I do get into the build programmatically, obviously from your input it may not matter. So, I will check both places, and cross their fingers that mem map will open the file.

cloudbank:
@andrewharp @martinwicke I had to upgrade bazel,ruby, AS 3.o, build tools... I reset my fork because I was so far behind on time and  I wanted a clean slate.  The new PR is  #14351

martinwicke:
Thanks, closing this one.

PR #12928, opened 2017-09-09
Improvements to Java API handling of datatypes and arrays

andrewcmyers:
This is a set of connected changes broken out from #11535 (Generic Java API) and improved
a bit further; still synergistic with that PR  and should allow whittling that PR down toward a digestible size.

* Changed how datatypes are computed from arrays to not crawl over the array but instead just inspect the array's class object. This seems somewhat simpler, more efficient, and easier to maintain going forward.
* Added the notion that a given array component type can correspond to more than one tensor datatype. I think that we already would like this to be able to support both uint8 and string using byte arrays, and the need will become more acute as more TensorFlow types are added.
* Resurrected the test case for creating and extracting uint8 tensors (requires previous change)

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
@@ -25,6 +25,7 @@ import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 import java.nio.LongBuffer;
 import java.util.Arrays;
+import java.util.HashMap;
 
 /**
  * A typed multi-dimensional array.
@@ -97,9 +98,19 @@ public final class Tensor implements AutoCloseable {
    *     using {@link #create(DataType, long[], ByteBuffer)} instead.
    */
   public static Tensor create(Object obj) {
+    return create(obj, dataTypeOf(obj));
+  }
+
+  /**
+   * Create a Tensor of data type {@code dtype} from a Java object.
+   * @param dtype the intended tensor data type. It must match the
+   *        the run-time type of the object.
+   */
+  static Tensor create(Object obj, DataType dtype) {
     Tensor t = new Tensor();
-    t.dtype = dataTypeOf(obj);
-    t.shapeCopy = new long[numDimensions(obj)];
+    t.dtype = dtype;
+    t.shapeCopy = new long[numDimensions(obj, dtype)];
+    assert objectCompatWithType(obj, dtype);
     fillShape(obj, 0, t.shapeCopy);
     if (t.dtype != DataType.STRING) {
       int byteSize = elemByteSize(t.dtype) * numElements(t.shapeCopy);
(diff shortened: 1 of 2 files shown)


andrewcmyers:
The resurrected test case was written for #12797 but wouldn't work without this additional flexibility for tensor creation.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -537,56 +548,72 @@ private static void throwExceptionIfNotByteOfByteArrays(Object array) {
...
+  
+  /** Returns the number of dimensions of the array object o.
+   *  Returns 0 if o is not an array.
+   */
+  private static int numArrayDimensions(Object o) {
+    Class<?> c = o.getClass();
+    if (!c.isArray()) return 0;
+    String s = c.getName();
+    int i = 0;
+    while (s.charAt(i) == '[') {
```
andrewcmyers: This could also be done by walking up the `getComponentType()` path but I'm conjecturing it's a little faster to just walk through the string.

It should also be correct to initialize i to 1 instead of 0 since the first character is guaranteed to be `[`, but starting from zero seemed more obviously correct .

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -537,56 +548,72 @@ private static void throwExceptionIfNotByteOfByteArrays(Object array) {
...
-      Object e = Array.get(o, 0);
-      if (e == null) {
-        throwExceptionIfNotByteOfByteArrays(o);
-        return 1;
-      } else if (Byte.class.isInstance(e) || byte.class.isInstance(e)) {
-        return 0;
-      }
-      return 1 + numDimensions(e);
+  /** Returns the number of dimensions of a tensor of type dtype
+   *  when represented by the object o.
```
asimshankar: Mind running the [source code formatter](https://github.com/google/google-java-format) on the files?

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
asimshankar: I don't have a strong opinion either way (though is there any reason to believe that one would be significantly faster than the other?)

asimshankar (changes requested):
Thanks for splitting this out, looks great, just a few minor nits.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
andrewcmyers: Done.

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -537,56 +548,72 @@ private static void throwExceptionIfNotByteOfByteArrays(Object array) {
     }
   }
 
+  static HashMap<Class<?>, DataType> class_datatypes = new HashMap<>();
```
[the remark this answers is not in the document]
andrewcmyers: Done.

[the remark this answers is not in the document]
andrewcmyers: Done.

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please

drpngx:
Windows should be fixed now.

Jenkins, test this please.

PR #12933, opened 2017-09-09
Reverse change eb75ded6 so that internal tests will pass.

frankchn:
As support for int64 global steps is not ready in TPUs, I am reversing this change so that our internal performance and regression tests will pass.


frankchn:
Jenkins, test this please.

jpienaar (approved):

PR #12934, opened 2017-09-09
Added support for JVM callback ops

eaplatanios:
This PR is a proposed workaround for a limitation related to my Scala API and discussed in #12895 and #12915. Given that we currently cannot load external op libraries through languages other than Python, this PR adds a JVM callback op to the main TensorFlow library. I understand this is not ideal and may not be acceptable but it's just offered as a possible workaround.

On the plus side, this op is not specific to Scala. It allows any JVM language to create ops that call functions in that language. The arguments and outputs of such functions are arrays of pointers to "eager tensors". The only limitation is that the op invocation has access to the original JVM instance that was used when constructing it. This is similar to the current `py_func` op limitation though and so, should be fine for now.

@asimshankar Do you think that this is acceptable for now?

P.S. The replication of a few lines from the C API code is ugly but I'm not sure what's best to do here and I didn't want to introduce more changes until you take a look.


asimshankar:
@eaplatanios : IIUC, this is to allow distribution of your Scala bindings using then nightly builds for the period during which @allenlavoie 's changes to the shared libraries aren't in effect. Given the estimated timeline (O(weeks)), and the fact that in the interim you could build your own libraries instead of using the nightly builds for your distribution, I'm tempted to suggest against adding this operation to the core TensorFlow libraries.

Thanks for for your understanding.

eaplatanios:
@asimshankar No problem, I totally understand and was kind of expecting that. I'll look into using the CI build scripts you provide on one of our servers.

PR #12950, opened 2017-09-10
random_crop option to return the crop offset

ahundt:
I have a segmentation application where I need to know the selected random crop offset to correctly crop the label image in the same manner. This PR adds the `random_crop(get_offset=True)` option to return that additional information to the user.

### DIFF
--- a/tensorflow/python/ops/random_ops.py
+++ b/tensorflow/python/ops/random_ops.py
@@ -274,7 +274,7 @@ def random_shuffle(value, seed=None, name=None):
       value, seed=seed1, seed2=seed2, name=name)
 
 
-def random_crop(value, size, seed=None, name=None):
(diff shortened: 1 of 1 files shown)


drpngx:
That looks a bit backward. Maybe passing in the precomputed random offset is better.

asimshankar:
(For API review: +1 to @drpng's suggestion: A new operation that requires the offset and crops to that offset sounds better)

ahundt:
@drpngx wouldn't that either be redundant with `tf.slice` or be better named `tf.crop` since it is not random?

drpngx:
Right, why not just use `tf.crop` and `tf.random`?

ahundt:
Fair enough, I might as well just write it in user code. Thanks for the feedback!

PR #12971, opened 2017-09-11
Add utf8 support for string_split

### LINKED ISSUE #11399: Let string_split support splitting utf-8 characters
### Describe the problem

Type of issue: feature request

The [string_split](https://www.tensorflow.org/versions/r0.12/api_docs/python/string_ops/splitting) function has (mostly) good behavior when splitting utf-8 strings by single-character delimiter, but fails to do it properly on null-width delimiter because of its documented behavior:

> If delimiter is an empty string, each element of the source is split into individual strings, each containing one byte. (This includes splitting multibyte sequences of UTF-8.)

For models like seq2seq one needs a split function that can split utf-8 strings into individual characters that can be processed by model as units, also embeddings having properly of being easily joined as utf-8 strings.

Could tensorflow provide alternative implementation of string_split that is utf-8 - aware?

yongtang:
This fix is an effort to try to address the request raised in #11399 where it was not possible to split utf8 strings with `tf.string_split`.

This fix adds an additional attr of `encoding` so that `utf8` could be specified for `string_split`.

This fix fixes #11399.


tensorflow/core/kernels/string_split_op.cc
```diff
@@ -28,19 +28,62 @@ namespace tensorflow {
 
 namespace {
 
-std::vector<string> Split(const string& str, const string& delimiter,
-                          const bool skipEmpty) {
+Status Split(std::vector<string>& char_vector, const string& str,
```
drpngx: We don't do references. Instead, pass the output as the last parameter, as a pointer:

```Status Split(const string& str, const string& delimiter, ..., std::vector<string>* char_vector)
```

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -76,6 +127,12 @@ class StringSplitOp : public OpKernel {
     const auto delimiter_vec = delimiter_tensor->flat<string>();
     const string& delimiter = delimiter_vec(0);
     // Empty delimiter means split the input character by character.
+    // Empty delimiter could be combined with a non-empty encoding.
+    if (encoding_ != "") {
+      OP_REQUIRES(ctx, delimiter == "",
```
drpngx: Why not implement this case?

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -28,19 +28,62 @@ namespace tensorflow {
...
     }
-    return str_util::Split(str, delimiter);
+    char_vector = str_util::Split(str, delimiter);
+    return Status::OK();
   }
-  std::vector<string> char_vector(str.size());
-  for (size_t i = 0; i < str.size(); ++i) {
-    char_vector[i] = str[i];
+  if (encoding == "utf8") {
+    // Bytes    Byte 1    Byte 2    Byte 3    Byte 4
```
drpngx: Could we push this upstream to str_util and provide a variant there? It might be useful for the rest of the code. If moved there, it will also have to have a unit test on its own.

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -28,19 +28,62 @@ namespace tensorflow {
...
+    //   1     0xxxxxxx
+    //   2     110xxxxx  10xxxxxx
+    //   3     1110xxxx  10xxxxxx  10xxxxxx
+    //   4     11110xxx  10xxxxxx  10xxxxxx  10xxxxxx
+    char_vector.reserve(str.size());
+    size_t off = 0, len = 0;
+    for (size_t i = 0; i < str.size(); ++i) {
+      if (off == i) {
+        if ((str[i] & 0x80) == 0x00) {
+          char_vector.push_back(str.substr(off, 1));
```
drpngx: emplace_back

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -28,19 +28,62 @@ namespace tensorflow {
...
+          off++;
+        } else if ((str[i] & 0xE0) == 0xC0) {
+          len = 2;
+        } else if ((str[i] & 0xF0) == 0xE0) {
+          len = 3;
+        } else if ((str[i] & 0xF8) == 0xF0) {
+          len = 4;
+        } else {
+          char_vector.clear();
+          return errors::Internal("invalid utf8 encoding");
```
drpngx: maybe report at which character we failed

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -28,19 +28,62 @@ namespace tensorflow {
...
+        }
+        if (off + len == i + 1) {
+          char_vector.push_back(str.substr(off, len));
+          off += len;
+        }
+      }
+    }
+    if (off < str.size()) {
+      // Not enough chars
```
drpngx: Remove comment and put that in the error message.

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -28,19 +28,62 @@ namespace tensorflow {
...
+      return errors::Internal("invalid utf8 encoding");
```
drpngx: I guess these are not really internal errors, right? They could be due to invalid input?

drpngx (changes requested):
After we agree, this will require an API review.

tensorflow/core/kernels/string_split_op.cc
yongtang: Thanks @drpngx. Done.

tensorflow/core/kernels/string_split_op.cc
yongtang: Done.

tensorflow/core/kernels/string_split_op.cc
yongtang: Done.

tensorflow/core/kernels/string_split_op.cc
yongtang: Done.

tensorflow/core/kernels/string_split_op.cc
yongtang: Done.

tensorflow/core/kernels/string_split_op.cc
yongtang: Done.

tensorflow/core/kernels/string_split_op.cc
yongtang: Done. Thanks.

yongtang:
Thanks @drpngx for the review. The PR has been updated with comments addressed. Please take a look.

drpngx:
Jenkins, test this please.

yongtang:
One Jenkins failure is `//tensorflow/tools/api/tests:api_compatibility_test` which is expected (API Change).

Another Jenkins failure is `//tensorflow/python/kernel_tests:fft_ops_test`. I am wondering if it is untreated? (saw this failure in other PRs as well).

drpngx:
Oh you need to rebase, we've fixed that. The `fft_ops_test` is fixed internally, I will push that later.

tensorflow/core/lib/strings/str_util.h
```diff
@@ -131,6 +131,15 @@ std::vector<string> Split(StringPiece text, StringPiece delims);
 template <typename Predicate>
 std::vector<string> Split(StringPiece text, StringPiece delims, Predicate p);
 
+// Split "text" into into a vector of strings where each string
+// in the array is a string of ASII or unicode chars. If the
+// delim is empty, then each string in the array consists of
+// one ASCII or one unicode char. If successful, the result
+// is passed to "*result" and returns true. Otherwise returns false
+// and error message is passed to "*error".
+bool SplitUTF8(StringPiece text, string delim, std::vector<string>* result,
```
drpngx: Can we use the `Status` object to return errors?

Also, you should use `const string& delim`, or `StringPiece delim`. Isn't there a restriction that `delim` is only one character? If so, we should document that, and check for that in the implementation.

yongtang: Thanks @drpngx. The PR has been updated with `Status` returned and `const string& delim`.

For `delim` length, in the kernel of StringSplitOp, there is already a check to guard the length of `delim`.

In the util library (`str_util.h`) the `SplitUTF8` actually allows a UTF8 encoded string (1-4 chars) as the delim. There is a test case in `str_util_test.cc` to show that.

yongtang:
Thanks @drpngx  for the review. The PR has been rebased and updated. Please take a look.

drpngx:
Jenkins, test this please.

drpngx:
Sounds fair. Note that since we have it in `str_util`, it's no longer sufficient to check at the single call site, since there might be a different part of the code calling us now.

Jenkins, test this please.

yongtang:
Thanks @drpngx for the help. I updated the PR with `PREDICT_FALSE` -> `TF_PREDICT_FALSE`.

drpngx:
Thanks! Jenkins, test this please.

tensorflow/core/ops/string_ops.cc
```diff
@@ -203,6 +203,7 @@ REGISTER_OP("StringSplit")
     .Output("values: string")
     .Output("shape: int64")
     .Attr("skip_empty: bool = true")
+    .Attr("encoding: string = ''")
```
asimshankar: Please also include a description of what this is in the documentation string below

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,86 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
   }
 }
 
+Status SplitUTF8(StringPiece text, const string& delim,
```
asimshankar: Isn't this function invoked only when `delim` is the empty string? 
I realize that you're perhaps adding a more general function, but would it make sense to not have this function in `str_util.h` but instead inline a function to split into UTF8 characters when the delimiter is empty in the anonymous namespace in `kernels/string_split_op.cc`?

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -85,7 +98,9 @@ class StringSplitOp : public OpKernel {
     int64 max_num_entries = 0;
     std::vector<int64> num_indices(batch_size);
     for (int64 i = 0; i < batch_size; ++i) {
-      std::vector<string> parts = Split(input_vec(i), delimiter, skip_empty_);
+      std::vector<string> parts;
+      OP_REQUIRES_OK(
+          ctx, Split(input_vec(i), delimiter, skip_empty_, encoding_, &parts));
```
asimshankar: If `encoding_ == "utf8"`, should the kernel fail if the delimiter is not valid UTF8?

asimshankar (commented):
From API review: A few comments. The observation that the `encoding` attribute is only relevant when the delimiter is the empty string is a bit surprising.

Not sure: But would it make sense to instead have a separate op that splits a string into UTF8 characters instead of adding to this op?

tensorflow/core/ops/string_ops.cc
yongtang: Thanks @asimshankar. The docs has been updated.

tensorflow/core/lib/strings/str_util.cc
yongtang: @asimshankar I updated the PR so that it is possible to have a non empty `delim` when encoding is `utf8`. However, because of the limitation of the current ops definition, only one char (instead of 1-4 chars in utf8 case) is allowed.

tensorflow/core/kernels/string_split_op.cc
yongtang: Thanks @asimshankar. The PR has been updated so that delimiter will be checked for properly encoding.

yongtang:
@asimshankar Thanks for the review. The PR has been updated. Currently, the implementation of this PR is to:

In case an UTF8 encoding is used, then either an empty string or a one character UTF8 (actually ASCII as there is only one char) is allowed to be specified as a delimiter. Properly encoding is validated now as well.

The reason for only allowing one char in UTF8 case is that, the `StringSplit` OP was defined such that only one char is allowed. If we expand the ops to take more than one chars, I think backward-compatibility may not be maintained.

For that it might make sense to introduce another OP instead of adding additional `encoding` attribute to `StringSplit`. I can update the PR if the other option is preferred. Please let me know.

sb2nov:
@asimshankar ping

asimshankar:
@yongtang : A separate op sounds preferable as it will be the more complete solution (like wanting to split on a multi-byte character). Let's do that instead.  Thanks!

yongtang:
@asimshankar Thanks. I will update the PR.

yongtang:
Thanks for the review @asimshankar. The PR has been updated with a new OP `string_utf8_split` added, which supports multi-bytes delimiter. Please take a look and let me know if there are any issues.

tensorflow/core/lib/strings/str_util.h
```diff
@@ -131,6 +133,18 @@ std::vector<string> Split(StringPiece text, StringPiece delims);
 template <typename Predicate>
 std::vector<string> Split(StringPiece text, StringPiece delims, Predicate p);
 
+// Split "text" into into a vector of strings where each string
+// in the array is a string of ASII or unicode chars. The `delim`
+// is a string with 0 or 1 chars.
```
asimshankar: Should be stay "`delim` is a single UTF-8-encoded unicode character or the empty string" (to avoid any confusion, since `delim.size()` can be more than 1)?

tensorflow/core/ops/string_ops.cc
```diff
@@ -247,6 +247,57 @@ shape: a length-2 vector of int64 representing the shape of the sparse
...
+
+  indices = [0, 0;
+             0, 1;
+             1, 0;
+             1, 1;
+             1, 2]
+  shape = [2, 3]
+  values = ['hello', 'world', 'a', 'b', 'c']
+
+input: 1-D. Strings to split.
```
asimshankar: UTF-8 encoded strings to split.

tensorflow/core/ops/string_ops.cc
```diff
@@ -247,6 +247,57 @@ shape: a length-2 vector of int64 representing the shape of the sparse
...
+delimiter: 1-D. Delimiter characters (bytes), or empty string.
```
asimshankar: Empty string or a single UTF-8 encoded unicode character

tensorflow/core/ops/string_ops.cc
```diff
@@ -247,6 +247,57 @@ shape: a length-2 vector of int64 representing the shape of the sparse
   of tokens in a single input entry.
 )doc");
 
+REGISTER_OP("StringUTF8Split")
```
asimshankar: Nit: Can we call this `StringSplitUTF8` instead?

tensorflow/core/ops/string_ops.cc
asimshankar: I didn't quite follow why `delimiter` is a 1-D tensor. Shouldn't it be a string scalar?

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -122,6 +122,84 @@ class StringSplitOp : public OpKernel {
   bool skip_empty_;
 };
 
+class StringUTF8SplitOp : public OpKernel {
```
asimshankar: IIUC, the implementation of these two kernels is almost identical, save for the different function call to split (and the vall to validating UTF8 of the delimiter).

Could we refactor the kernel implementation so that a bulk of the code is shared? Something along the lines of:

```c++
class StringSplitOpBase : public OpKernel {
    ...
    Status ValidateDelimiter(const string& delimiter) = 0;
    Status Split(const string& input, const string& delimiter, vector<string>* parts) = 0;
}

class StrignSplitOp : public StringSplitOpBase {
  Status ValidateDelimiter(const string& delimiter) { return Status::OK(); }
  Status Split(const string& input, const string& delimiter, vector<string>* parts) {
    return str_util::Split(...);
  }  
};

class StringSplitUTF8Op : public StringSplitOp {
  Status ValidateDelimiter(const string& delimiter) { return str_util::ValidUTF8(delimiter); }
  Status Split(...) { return str_util::SplitUTF8(...); }
};
```

That way it's easy to keep a bulk of the logic in sync?

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,119 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+        off++;
+      } else if ((text[i] & 0xE0) == 0xC0) {
+        len = 2;
+      } else if ((text[i] & 0xF0) == 0xE0) {
+        len = 3;
+      } else if ((text[i] & 0xF8) == 0xF0) {
+        len = 4;
+      } else {
+        result->clear();
+        return errors::InvalidArgument("Invalid UTF8 encoding at position of ",
```
asimshankar: Replace "position" with "byte" in the error messages to make it clear that the position refers to the byte and not the character?

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,119 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+  size_t off = 0, len = 0;
+  bool advance = true;
+  for (size_t i = 0; i < text.size(); ++i) {
+    if (off == i) {
+      if ((text[i] & 0x80) == 0x00) {
+        string entry = text.substr(off, 1).ToString();
+        if (delim == "") {
+          // If delim is "" then always advance
+          result->emplace_back(entry);
+        } else if (delim != entry) {
```
asimshankar: Could this be made a little more compact by separating out the length determination from the emplacement? Something like:

```c++
size_t char_start = 0;
size_t len = 0;
for (size_t i = 0; i < text.size(); ++i) {
  if (char_start == i) {
    // Determine byte length of next character.
    const char c = text[i];
    if (c & 0x80 == 0x00) {
     len = 1;
    } else if (c & 0xE0 == 0xC0) {
      len = 2;
    } else if (c & 0xF0 == 0xE0) {
      len = 3;
    } else if (c & 0XF8 == 0xF0) {
      len = 4;
   } else {
    result->clear();
    return errors::InvalidArgument("Invalid UTF-8 encoding at byte ", i);
   }
   // Validate all intermediate bytes are valid UTF-8
   if (TF_PREDICT_FALSE((len > 1 && (text[i] & 0xC0) != 0x80)) {
      /*  return error */
   }
   // Reached the end of a character
   if (char_start + len == i + 1) {
     const StringPiece entry = text.substr(char_start, len).ToString();
     char_start += len;
     if (delim == "") {
       result->emplace_back(entry);
     } else if (delim == entry) {
        advance = true;
     } else if (advance) {
        advance = false;
        result->emplace_back(entry);
    } else {
        result->back() = result->back() + entry;
   }
}
```

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,119 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+    }
+  }
+  if (TF_PREDICT_FALSE(off < text.size())) {
+    result->clear();
+    return errors::InvalidArgument("Not enough characters for UTF8 encoding");
+  }
+  return Status::OK();
+}
+
+Status ValidUTF8(StringPiece text) {
```
asimshankar: It seems like this function is determining whether `text` is a valid UTF-8 character (as opposed to a valid string). Perhaps call this function so: `ValidUTF8Character` or something?

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,119 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+  // Bytes    Byte 1    Byte 2    Byte 3    Byte 4
+  //   1     0xxxxxxx
+  //   2     110xxxxx  10xxxxxx
+  //   3     1110xxxx  10xxxxxx  10xxxxxx
+  //   4     11110xxx  10xxxxxx  10xxxxxx  10xxxxxx
+  if (text.size() == 0) {
+    return Status::OK();
+  }
+  size_t len = 0;
+  if ((text[0] & 0x80) == 0x00) {
```
asimshankar: You could factor this out into a helper function that both this and `Split` use:

```c++
static Status UTF8CharNumBytes(StringPiece text, size_t* num_bytes);
```

tensorflow/python/kernel_tests/string_split_op_test.py
```diff
@@ -21,6 +21,7 @@
 import numpy as np
 
 from tensorflow.python.framework import dtypes
+from tensorflow.python.framework import errors
```
asimshankar: Do you need to import `errors` or can you use `errors_impl` like the other tests?

asimshankar (changes requested):
Thanks for the update @yongtang , some suggestions.

tensorflow/core/kernels/string_split_op.cc
yongtang: Thanks @asimshankar. The class has been split as suggested.

tensorflow/core/lib/strings/str_util.cc
yongtang: Thanks. Done.

tensorflow/core/lib/strings/str_util.cc
yongtang: Done.

tensorflow/core/lib/strings/str_util.cc
yongtang: Done.

tensorflow/core/lib/strings/str_util.cc
yongtang: Thanks. Done.

tensorflow/core/lib/strings/str_util.h
yongtang: Done.

tensorflow/core/ops/string_ops.cc
yongtang: Done.

tensorflow/core/ops/string_ops.cc
yongtang: Done.

tensorflow/core/ops/string_ops.cc
yongtang: Done.

yongtang:
Thanks @asimshankar for the detailed review. The PR has been updated. Please take a look and let me know if there are any issues.

tensorflow/core/lib/strings/str_util.cc
asimshankar: Hmm...didn't see any change here. Perhaps a forgotten commit?

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -45,19 +45,16 @@ std::vector<string> Split(const string& str, const string& delimiter,
...
-  }
+  explicit StringSplitOpBase(OpKernelConstruction* c) : OpKernel(c) {}
+
+  virtual Status ValidateDelimiter(const string& delimiter) = 0;
+
+  virtual Status Split(const string& input, const string& delimiter,
+                       std::vector<string>* parts) = 0;
 
-  void Compute(OpKernelContext* ctx) override {
+  void Compute(OpKernelContext* ctx) {
```
asimshankar: Why was the "override" keyword removed?

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -118,10 +117,50 @@ class StringSplitOp : public OpKernel {
...
+    return Status::OK();
+  }
+};
+
+class StringSplitUTF8Op : public StringSplitOpBase {
+ public:
+  explicit StringSplitUTF8Op(OpKernelConstruction* context)
+      : StringSplitOpBase(context) {
+    // The skip_empty is not optional (unlike StringSplitOp)
+    OP_REQUIRES_OK(context, context->GetAttr("skip_empty", &skip_empty_));
```
asimshankar: I'm probably missing something, but I don't see `skip_empty_` being used by `StringSplitUTF8Op`. How is it having an effect?

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -118,10 +117,50 @@ class StringSplitOp : public OpKernel {
     }
   }
 
- private:
+ protected:
```
asimshankar: Let's make `skip_empty_` a private member of the subclasses. Doesn't seem like it us used in the base class and it is only set by the subclass constructors, so it will be cleaner to move it down there?

tensorflow/core/lib/strings/str_util.cc
asimshankar: Hmm...I didn't see much of a change. By switching things around a bit we won't have to treat the `len == 1` case differently from the others repeat the blocks of code that add to `result`. Something like this:

```c++
Status SplitUTF8(StringPiece text, const string& delim, std::vector<string>* result) {
  result->clear();
  size_t char_start = 0, char_len = 0;
  bool advance = true;
  for (size_t i = 0; i < text.size(); ++i) {
    if (char_start == i) {
      Status s = UTF8CharNumBytes(text, &char_len);
      if (!s.ok()) {
        result->clear();
        return errors::InvalidArgument("Invalid UTF8 encoding at byte ", i, ": ", s.error_message());
      }
    }
    // Validate all intermediate bytes are valid UTF-8
    if (TF_PREDICT_FALSE((len > 1 && (text[i] & 0xC0) != 0x80)) {
       return errors::InvalidArgument("Invalid UTF8 encoding at byte ", i);
    }
    // Reached the end of a character
    if (char_start + char_len == i + 1) {
      const StringPiece entry = text.substr(char_start, char_len).ToString();
      char_start += len;
      if (delim == "") {
        result->emplace_back(entry);
      } else if (delim == entry) {
         advance = true;
      } else if (advance) {
         advance = false;
         result->emplace_back(entry);
      } else {
         result->back() = result->back() + entry;
      }
    }
  }
  if (TF_PREDICT_FALSE(char_start < text.size())) {
    return errors::InvalidArgument("Invalid UTF8 encoding, incomplete last character");
  }
  return Status::OK();
}
```

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,118 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+  //   3     1110xxxx  10xxxxxx  10xxxxxx
+  //   4     11110xxx  10xxxxxx  10xxxxxx  10xxxxxx
+  result->clear();
+  result->reserve(text.size());
+  size_t off = 0, len = 0;
+  bool advance = true;
+  for (size_t i = 0; i < text.size(); ++i) {
+    if (off == i) {
+      size_t num_bytes = 0;
+      Status s = UTF8CharNumBytes(text.substr(off, 4), &num_bytes);
```
asimshankar: Why `text.substr(off, 4)` here?
That might overflow `text` no? And we could just as well provide `text` (i.e, without `substr`)?

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,118 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+  if ((text[0] & 0x80) == 0x00) {
+    *num_bytes = 1;
+  } else if ((text[0] & 0xE0) == 0xC0) {
+    *num_bytes = 2;
+  } else if ((text[0] & 0xF0) == 0xE0) {
+    *num_bytes = 3;
+  } else if ((text[0] & 0xF8) == 0xF0) {
+    *num_bytes = 4;
+  } else {
+    return errors::InvalidArgument("Invalid UTF8 encoding at position of 0");
```
asimshankar: Hmm...this "position 0" isn't adding much value. Perhaps make this function return a `bool` instead and fill in the status message with offset information at the caller?

tensorflow/python/kernel_tests/string_split_op_test.py
```diff
@@ -145,6 +145,56 @@ def testStringSplitWithNoSkipEmpty(self):
       self.assertAllEqual(indices, [[0, 0], [1, 0], [2, 0]])
       self.assertAllEqual(shape, [3, 1])
 
+  def testStringSplitWithUtf8AndEmptyDelimiter(self):
+    # utf8 \xE6\x82\xA8 \xE5\xA5\xBD, \xE6\x82\xA8 \xE5\xA5\xBD
+    strings = [b"\xE6\x82\xA8\xE5\xA5\xBD", b"\xE4\xB8\x96\xE7\x95\x8C"]
+
+    with self.test_session() as sess:
+      tokens = string_ops.string_split_utf8(strings, delimiter="")
```
asimshankar: I don't see any tests for `skip_empty=True`. Also, it appears that `skip_empty` isn't handled. If you'd rather get rid of the `skip_empty` op attribute altogether, that's fine. If not, let's make it work :)

asimshankar (changes requested):
Thanks for the update @yongtang - though now I see that `skip_empty` isn't being handled. And some other suggestions...

yongtang:
Thanks @asimshankar for the review. Will updated the PR shortly.

tensorflow/core/kernels/string_split_op.cc
yongtang: Added back. Thanks for notice that.

tensorflow/core/kernels/string_split_op.cc
yongtang: Done.

tensorflow/core/kernels/string_split_op.cc
yongtang: The PR has been updated with `skip_empty_`.

tensorflow/core/lib/strings/str_util.cc
yongtang: Done.

yongtang:
Thanks @asimshankar for the review. The PR has been updated with review comments addressed. Please take a look.

tensorflow/python/kernel_tests/string_split_op_test.py
asimshankar: Ping?

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,99 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+    }
+    // Reached the end of a character
+    if (char_start + char_len == i + 1) {
+      const StringPiece entry = text.substr(char_start, char_len).ToString();
+      if (delim == "") {
+        result->emplace_back(entry.ToString());
+      } else if (delim == entry) {
+        advance = true;
+        if (!skip_empty) {
+          if (TF_PREDICT_FALSE(char_start == 0)) {
```
asimshankar: IIUC, this is handling the case when `skip_empty` is true and the delimiter is the first or last character. But what about cases where there are multiple consecutive delimiters? See comment about testing (and perhaps duplicating/reusing `testStringSplitWithNoSkipEmpty` with this new op)

asimshankar (changes requested):

tensorflow/core/lib/strings/str_util.cc
yongtang: Thanks @asimshankar. My previous interpretation was not right. I have updated the PR so that now it will match python behavior in case `skip_empty=False`, i.e.,
```
>>> "##a##b##c##".split("#")
['', '', 'a', '', 'b', '', 'c', '', '']
>>> 
```

tensorflow/python/kernel_tests/string_split_op_test.py
yongtang: @asimshankar The tests have been updated with test cases covered when skip_empty=False. Ln 160-208.

yongtang:
@asimshankar Thanks for the review. The PR has been updated. Please take a look.

martinwicke:
@asimshankar can you look at the new changes?

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,103 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+        result->emplace_back(entry.ToString());
+      } else if (delim == entry) {
+        advance = true;
+        if (!skip_empty) {
+          // Follow python style, if it is beginning or the end, always add ''
+          // >>> "##a##b##c##".split("#")
+          // ['', '', 'a', '', 'b', '', 'c', '', '']
+          if (result->size() == 0 || result->back() != "") {
+            result->emplace_back("");
+          } else if (result->size() == 1) {
```
asimshankar: I didn't quite follow the logic here (in particulary while `result->size() == 1` is any more special that `result->size() == 2` etc.). For example, let's say the the call is:

```
SplitUTF8("###a#", '#')
```

then I think this will end up with: `{ "", "", "a", ""}` instead of three empty strings before "a".
If that is correct, probably worth adding the unittest that would have caught the bug here.

I think this can be simplified a bit by keeping a boolean indicating whether the last character was a delimiter and using that instead. Something like:

```c++
} else if (delim == entry) {
  if (last_char_was_a_delimiter && !skip_empty) {
    result->emplace_back("");
  }
  last_char_was_a_delimiter = true;
} else if (last_char_was_a_delimiter) {
  last_char_was_a_delimiter = false;
  result->emplace_back(entry.ToString());
} else {
  result->back() += enty.ToString();
}
```

tensorflow/core/lib/strings/str_util_test.cc
```diff
@@ -417,4 +417,33 @@ TEST(StringReplace, EmptyStringReplaceAll) {
...
+      Status::OK(),
+      str_util::SplitUTF8("\xE6\x82\xA8\xE5\xA5\xBD\xE4\xB8\x96\xE7\x95\x8C",
+                          "\xE5\xA5\xBD", &result));
+  EXPECT_EQ(result.size(), 2);
+
+  EXPECT_EQ(errors::InvalidArgument("Invalid UTF8 encoding at byte of 1"),
+            str_util::SplitUTF8("\xE2\x28\xA1", "", &result));
+
+  EXPECT_EQ(errors::InvalidArgument("Not enough characters for UTF8 encoding"),
+            str_util::SplitUTF8("\xE6\x82", "", &result));
```
asimshankar: Perhaps also include the simpler string tests, like the ones in `Split.Basic`?

asimshankar (changes requested):
Sorry, somehow missed my radar.

tensorflow/core/kernels/string_split_op.cc
```diff
@@ -117,11 +116,54 @@ class StringSplitOp : public OpKernel {
...
+};
+
+class StringSplitOp : public StringSplitOpBase {
+ public:
+  explicit StringSplitOp(OpKernelConstruction* context)
+      : StringSplitOpBase(context) {
+    skip_empty_ = true;
+    // By default skip_empty_ is true. We only get the value from attr if it is
+    // available, so that it is backward compatible.
+    context->GetAttr("skip_empty", &skip_empty_);
```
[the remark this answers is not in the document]
yongtang: Thanks @asimshankar. Done.

tensorflow/core/lib/strings/str_util_test.cc
yongtang: Thanks. The tests with `Split.Basic` has been added to test against `SplitUTF8` as well.

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,103 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+  bool advance = true;
+  for (size_t i = 0; i < text.size(); ++i) {
+    if (char_start == i) {
+      if (!UTF8CharNumBytes(text, &char_len)) {
+        result->clear();
+        return errors::InvalidArgument("Invalid UTF8 encoding at byte ", i);
+      }
+    }
+    // Validate all intermediate bytes are valid UTF-8
+    if (TF_PREDICT_FALSE((char_len > 1) && (char_start != i) &&
```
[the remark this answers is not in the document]
yongtang: Thanks. Done.

tensorflow/core/lib/strings/str_util.cc
yongtang: @asimshankar I think some special handling of begin and end may still be needed. For example, in python:
```python
Python 2.7.10 (default, Feb  7 2017, 00:08:15) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
>>> "###a#".split("#")
['', '', '', 'a', '']
>>> ",".split(",")
['', '']
>>> 
```

The first case ` "###a#".split("#")` returns 3 empty chars before `a` and the second case will return 2 chars.

I updated the PR to try to match the above situation. Please take a look and see if it fit.

yongtang:
Thanks @asimshankar for the review and help. The PR has been updated. Please take a look.

tensorflow/core/lib/strings/str_util_test.cc
```diff
@@ -417,4 +417,77 @@ TEST(StringReplace, EmptyStringReplaceAll) {
...
+  EXPECT_EQ(result[3], "");
+  EXPECT_EQ(result[4], "b");
+  EXPECT_EQ(result[5], "");
+  EXPECT_EQ(result[6], "c");
+  EXPECT_EQ(result[7], "");
+  EXPECT_EQ(result[8], "");
+
+  // ASCII should work as well
+  EXPECT_EQ(Status::OK(), str_util::SplitUTF8("", ",", false, &result));
+  EXPECT_EQ(result.size(), 0);
```
asimshankar: Add a test for an empty string split with `skip_empty = true` as well?

tensorflow/core/ops/string_ops.cc
```diff
@@ -247,6 +247,57 @@ shape: a length-2 vector of int64 representing the shape of the sparse
...
+Split elements of `input` based on `delimiter` into a `SparseTensor`.
+
+Let N be the size of source (typically N will be the batch size). Split each
+element of `input` based on `delimiter` and return a `SparseTensor`
+containing the splitted tokens. Empty tokens are ignored.
+
+`delimiter` can be empty, or a string of split characters. If `delimiter` is an
+ empty string, each element of `input` is split into individual single-UTF-8
+ character strings, including splitting of UTF-8 multibyte sequences. Otherwise
+ every character of `delimiter` is a potential split point.
```
asimshankar: Sorry, I should have caught this earlier - the documentation here says that `delimiter` can be a sequence of characters, in which case, it splits on any of them. However, the implementation does not seem to do that.

For the original `StringSplit` op:
- [this test](https://github.com/tensorflow/tensorflow/blob/r1.4/tensorflow/core/lib/strings/str_util_test.cc#L242), and
- [this test](https://github.com/tensorflow/tensorflow/blob/r1.4/tensorflow/python/kernel_tests/string_split_op_test.py#L110)
exercise the case where there are multiple delimiter characters specified.

So, long story short, unless I'm missing something - for `StringSplitUTF8`, delimiter cannot be a string of split characters, it has to be a single split character.

Is that correct? Instead of adding the complexity of handling multiple split characters for now (unless you're up for it), I'd change this to expect a single split character, and if the feature ever comes up then we'd allow for multiple.

Sound reasonable?

asimshankar (changes requested):

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,100 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+    if (char_start + char_len == i + 1) {
+      const StringPiece entry = text.substr(char_start, char_len).ToString();
+      if (delim == "") {
+        result->emplace_back(entry.ToString());
+      } else if (delim == entry) {
+        if (last_char_was_a_delimiter && !skip_empty) {
+          result->emplace_back("");
+        }
+        last_char_was_a_delimiter = true;
+        if (i + 1 == text.size() && !skip_empty) {
```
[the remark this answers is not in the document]
yongtang: Thanks @asimshankar. That works. The PR has been updated.

tensorflow/core/lib/strings/str_util_test.cc
yongtang: Thanks. Done.

tensorflow/core/lib/strings/str_util_test.cc
```diff
@@ -423,21 +423,65 @@ TEST(SplitUTF8, Basic) {
...
+  EXPECT_EQ(errors::InvalidArgument("Invalid UTF8 encoding at byte 1"),
+            str_util::SplitUTF8("\xE2\x28\xA1", "", false, &result));
+
+  EXPECT_EQ(errors::InvalidArgument(
+                "Invalid UTF8 encoding, incomplete last character"),
+            str_util::SplitUTF8("\xE6\x82", "", false, &result));
+
+  // split("###a#", "#") -> { "", "", "", "a", ""}
+  EXPECT_EQ(Status::OK(), str_util::SplitUTF8("###a#", "#", false, &result));
+  EXPECT_EQ(result.size(), 5);
```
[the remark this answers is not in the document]
yongtang: Thanks. The PR has been updated.

tensorflow/core/lib/strings/str_util.cc
```diff
@@ -416,6 +417,100 @@ bool ConsumeNonWhitespace(StringPiece* s, StringPiece* val) {
...
+          // If this is the last we always add an empty one
+          // to handle split(",",",") -> ["", ""]
+          result->emplace_back("");
+        }
+      } else if (last_char_was_a_delimiter) {
+        last_char_was_a_delimiter = false;
+        result->emplace_back(entry.ToString());
+      } else if (result->size() == 0) {
```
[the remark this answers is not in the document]
yongtang: @asimshankar This condition might actually be needed. The reason is that the following line 
```
      } else {
        result->back() = result->back() + entry.ToString();
      }
```
might have result empty. So incase result is empty then we will initialize instead of append to `result->back()`.

tensorflow/core/ops/string_ops.cc
yongtang: Thanks @asimshankar. Given the original scope of #11399 I think that might be the best option for now. We can always add additional processing later if there is a request. The doc string has been updated.

Thanks for pointing this out and really appreciate that.

yongtang:
Thanks @asimshankar for the help and review. The PR has been updated. Please take a look.

yongtang:
There seems to have failed tests. Will update the PR shortly.

yongtang:
The failure is about Windows and the rest of builds seem pass. I don't have a Windows dev machine setup yet, but will try to take a look and see what I could do to fix it.

martinwicke:
@asimshankar can you take another look? I triggered the tests.

asimshankar:
I was under the impression that @yongtang was digging into the test failures on Windows. Looking at the failures I see:

```
FAIL: testStringSplitWithUtf8AndEmptyDelimiter (__main__.StringSplitOpTest)
  File "C:/tf_jenkins/workspace/tensorflow-pr-win-cmake-py/tensorflow/python/kernel_tests/string_split_op_test.py", line 221, in testStringSplitWithUtf8AndEmptyDelimiter
    b"\xE4\xB8\x96", b"\xE7\x95\x8C"])
  File "C:\Program Files\Anaconda3\lib\site-packages\tensorflow\python\framework\test_util.py", line 1148, in assertAllEqual
    np.testing.assert_array_equal(a, b)
  File "C:\Program Files\Anaconda3\lib\site-packages\numpy\testing\utils.py", line 871, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Program Files\Anaconda3\lib\site-packages\numpy\testing\utils.py", line 796, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

(mismatch 100.0%)
 x: array([b'\x00\x82\xa8', b'\x00\xa5\xbd', b'\x00\xb8\x96', b'\x00\x95\x8c'], dtype=object)
 y: array([b'\xe6\x82\xa8', b'\xe5\xa5\xbd', b'\xe4\xb8\x96', b'\xe7\x95\x8c'], 
      dtype='|S3')
```

Suggesting things about the `dtype` of numpy arrays of strings that I am not well versed in.
Will have to look into it. If anyone has ideas, much appreciated. Otherwise will ask around internally as well.

tensorflowbutler:

drpngx:
@yongtang did you investigate this any further?

tensorflowbutler:

tensorflowbutler:

tensorflowbutler:
Nagging Assignee @asimshankar: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

yongtang:
The PR has been rebased to resolve the merge conflict. The Windows issue may still exist.

I finally managed to find a Windows VM image from Microsoft:
https://developer.microsoft.com/en-us/windows/downloads/virtual-machines

I will spend some time to see if I could fix the Windows issue.

tensorflowbutler:
Nagging Assignee @asimshankar: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

yongtang:
@asimshankar Sorry for lack of updates.

I was able to find a Windows VM with Visual Studio 2017 preinstalled license  from https://developer.microsoft.com/en-us/windows/downloads/virtual-machines (license will expire in 05/13).

However, it appears that TensorFlow currently does not work well with Visual Studio 2017 build.

I have tried multiple times but still could not get a successful Windows build first (either with cmake or with bazel). Therefore unable to move forward to debug the issue of utf8 string with Windows.

GIven the current status I think it might make sense to close this PR or label it as `stalled` for now. I will give it another try and if I am able to move forward again, will open the new PR.

Thanks for the help during the process and sorry for the inconvenience.

drpngx:
@guschmue does anything jump at you?

drpngx:
Re-running tests to get the logs out.

guschmue:
I have not build for some time but can give it a try.

tensorflowbutler:
Nagging Assignee @asimshankar: It has been 15 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

tensorflowbutler:
Nagging Assignee @asimshankar: It has been 32 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

yongtang:
I feel really ashamed to not be able to fix the Windows test failure after so long. It shouldn't be very hard.

However, I still couldn't even build the tensorflow from source with a Windows 10 VM machine (and Visual Studio 2017), after another attempt. Without that I just could not debug the issue. (Don't have a Windows machine).

I have created an issue #19583 to list the detailed steps for my failure encountered. Would appreciate any help to make tensorflow build on Windows 10 + VS2017.

drpngx:
No worries! It's always hard to debug if you don't have access to a system. For the record, the failures are:

```
FAIL: testStringSplitWithUtf8AndEmptyDelimiter (__main__.StringSplitOpTest)
  File "T:/src/github/tensorflow/tensorflow/python/kernel_tests/string_split_op_test.py", line 221, in testStringSplitWithUtf8AndEmptyDelimiter
    b"\xE4\xB8\x96", b"\xE7\x95\x8C"])
  File "C:\Python35\lib\site-packages\tensorflow\python\framework\test_util.py", line 1460, in assertAllEqual
    np.testing.assert_array_equal(a, b, err_msg=msg)
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 813, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Arrays are not equal
(mismatch 100.0%)
 x: array([b'\x00\x82\xa8', b'\x00\xa5\xbd', b'\x00\xb8\x96', b'\x00\x95\x8c'], dtype=object)
 y: array([b'\xe6\x82\xa8', b'\xe5\xa5\xbd', b'\xe4\xb8\x96', b'\xe7\x95\x8c'],
      dtype='|S3')
FAIL: testStringSplitWithUtf8AndNonSkipEmpty (__main__.StringSplitOpTest)
  File "T:/src/github/tensorflow/tensorflow/python/kernel_tests/string_split_op_test.py", line 170, in testStringSplitWithUtf8AndNonSkipEmpty
    b"", b"\xE7\x95\x8C"])
  File "C:\Python35\lib\site-packages\tensorflow\python\framework\test_util.py", line 1460, in assertAllEqual
    np.testing.assert_array_equal(a, b, err_msg=msg)
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 813, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Arrays are not equal
(mismatch 50.0%)
 x: array([b'\x00\xa5\xbd', b'', b'', b'\x00\x95\x8c'], dtype=object)
 y: array([b'\xe5\xa5\xbd', b'', b'', b'\xe7\x95\x8c'],
      dtype='|S3')
FAIL: testStringSplitWithUtf8AndSkipEmpty (__main__.StringSplitOpTest)
  File "T:/src/github/tensorflow/tensorflow/python/kernel_tests/string_split_op_test.py", line 157, in testStringSplitWithUtf8AndSkipEmpty
    self.assertAllEqual(values, [b"\xE5\xA5\xBD", b"\xE7\x95\x8C"])
  File "C:\Python35\lib\site-packages\tensorflow\python\framework\test_util.py", line 1460, in assertAllEqual
    np.testing.assert_array_equal(a, b, err_msg=msg)
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 813, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Arrays are not equal
(mismatch 100.0%)
 x: array([b'\x00\xa5\xbd', b'\x00\x95\x8c'], dtype=object)
 y: array([b'\xe5\xa5\xbd', b'\xe7\x95\x8c'],
      dtype='|S3')
FAIL: testStringSplitWithUtf8AndUtf8MultiBytesDelimiter (__main__.StringSplitOpTest)
  File "T:/src/github/tensorflow/tensorflow/python/kernel_tests/string_split_op_test.py", line 235, in testStringSplitWithUtf8AndUtf8MultiBytesDelimiter
    b"\xE4\xB8\x96\xE7\x95\x8C", b"\xE5\xA5\xBD"])
  File "C:\Python35\lib\site-packages\tensorflow\python\framework\test_util.py", line 1460, in assertAllEqual
    np.testing.assert_array_equal(a, b, err_msg=msg)
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 813, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Python35\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Arrays are not equal
(mismatch 100.0%)
 x: array([b'\x00\xa5\xbd', b'\x00\xb8\x96\xe7\x95\x8c',
       b'\x00\xb8\x96\xe7\x95\x8c', b'\x00\xa5\xbd'], dtype=object)
 y: array([b'\xe5\xa5\xbd', b'\xe4\xb8\x96\xe7\x95\x8c',
       b'\xe4\xb8\x96\xe7\x95\x8c', b'\xe5\xa5\xbd'],
      dtype='|S6')
```
and the MacOS errors are:
```
[ RUN      ] SplitUTF8.Basic
tensorflow/core/lib/strings/str_util_test.cc:470: Failure
      Expected: result[3]
      Which is: "\0"
To be equal to: "a"
tensorflow/core/lib/strings/str_util_test.cc:479: Failure
      Expected: result[2]
      Which is: "\0"
To be equal to: "a"
tensorflow/core/lib/strings/str_util_test.cc:481: Failure
      Expected: result[4]
      Which is: "\0"
To be equal to: "b"
tensorflow/core/lib/strings/str_util_test.cc:483: Failure
      Expected: result[6]
      Which is: "\0"
To be equal to: "c"
tensorflow/core/lib/strings/str_util_test.cc:495: Failure
      Expected: str_util::Join(result, "|")
      Which is: "\0"
To be equal to: "a"
tensorflow/core/lib/strings/str_util_test.cc:501: Failure
      Expected: str_util::Join(result, "|")
      Which is: "\0|\0|\0"
To be equal to: "a|b|c"
tensorflow/core/lib/strings/str_util_test.cc:505: Failure
      Expected: str_util::Join(result, "|")
      Which is: "\0|||\0||\0|"
To be equal to: "a|||b||c|"
tensorflow/core/lib/strings/str_util_test.cc:508: Failure
      Expected: str_util::Join(result, "|")
      Which is: "\0|\0|\0"
```

drpngx:
Basically, it looks like there is something wrong with the logic. It seems to return `\x0` instead of the original substrings.

tensorflowbutler:
Nagging Assignee @asimshankar: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

tensorflowbutler:

tensorflowbutler:
It has been 43 days that this pull-request has stalled. Please create a new pull-request with the requested changes.

marhlder:
Is there any progress on this or is it just stalled forever?

Issue #12986, opened 2017-09-12
boringssl sha256 checksum mismatched


mylxiaoyi:
https://github.com/tensorflow/tensorflow/blob/5541ef4fbba56cf8930198373162dd3119e6ee70/tensorflow/workspace.bzl#L583

when download from https://github.com/google/boringssl/archive/e3860009a091cd1bd2bc189cdbc3c6d095abde84.tar.gz, the sha256 checksum is 
```
a9a3673b1f7bd80ef563e9de1d9ccdb5126dc0cce6377977009092148993c4fe
```

 , but expected is
```
02f5950f93c4fd3691771c07c9d04cf2999ab01383ff99da345249e93b0fcfb2 
```
.

Is there something wrong?

thanks.

d4l3k:
Also at https://github.com/tensorflow/tensorflow/issues/12979

kdatta:
Also at #12984

cy89:
@gunan, can you please take a look?

gunan:
Are we still seeing this?
The root cause is changes in github side.
They are just creating archives differently now. We are looking into if we can mirror our dependencies ourselves now.

gunan:
deduping this against #12979 Please watch that issue for progress.

PR #12752, opened 2017-09-01
Avoid gcc7 memcpy build error by updating BoringSSL

louiehelm:
Tensorflow produces this build error with gcc7 due to ambiguous memcpy inlining in all versions of BoringSSL prior to https://github.com/google/boringssl/commit/17cf2cb1d226b0ba2401304242df7ddd3b6f1ff2  on Dec 12, 2016:

```
In file included from /usr/include/string.h:639:0,
                 from external/boringssl/src/crypto/asn1/a_bitstr.c:59:
In function 'memcpy',
    inlined from 'i2c_ASN1_BIT_STRING' at external/boringssl/src/crypto/asn1/a_bitstr.c:118:5:
/usr/include/bits/string3.h:53:10: error: '__builtin_memcpy': specified size between 18446744071562067968 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]
   return __builtin___memcpy_chk (__dest, __src, __len, __bos0 (__dest));
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

Updating BoringSSL snapshot from 7-11-16 to 7-7-17 fixes this.

July 7 snapshot was picked because this is the stable build that Apollo uses:
https://github.com/ApolloAuto/apollo/pull/185/files

Reliance on older snapshots of BoringSSL with this same memcpy bug is also breaking GRPC and Tensorflow Serving on modern gcc toolchains in the same way:

https://github.com/grpc/grpc/issues/10843
https://github.com/grpc/grpc/issues/11765
https://github.com/tensorflow/serving/issues/475

Note: This build error is ubiquitous since bazel's primary workspace.bzl downloads and builds BoringSSL regardless of whether it is required (i.e., even when Google Cloud Storage support is disabled and when SSL support is not enabled).

PR #12995, opened 2017-09-12
boringssl sha256 checksum mismatched

facaiy:
issue #12986 .

### How to test 

+ [ ] confirmed by #12752 @louiehelm @gunan


drpngx:
How can the SHA be off?

gunan:
This also happened for gemmlowp.
Someone mentioned github regenerated all their compressed archive files which can explain all these failures.

Since we also have a mirror, this is not as simple. I would like to triage this internally and push out the change.

facaiy:
So the best way is to stop check SHA for Github, right?

gunan:
No, we will try to find a stable source for our dependencies. We will always enforce sha checks whenever we depend on other repositories.

PR #13021, opened 2017-09-13
Fixes in MKL add_n  to address several regressions

agramesh1:
Fixed failures due to scalar and empty tensors

### DIFF
--- a/tensorflow/core/kernels/mkl_aggregate_ops.cc
+++ b/tensorflow/core/kernels/mkl_aggregate_ops.cc
@@ -82,7 +82,7 @@ class MklAddNOp : public OpKernel {
         Tensor* out_tensor = nullptr;
         mkl_context.output_shape.SetMklTensor(false);
         AllocateOutputSetMklShape(ctx, 0, &out_tensor, o_shape,
-                                mkl_context.output_shape);
+                                 mkl_context.output_shape);
         return;
       }
     }
--- a/tensorflow/core/kernels/mkl_aggregate_ops.cc
+++ b/tensorflow/core/kernels/mkl_aggregate_ops.cc
@@ -55,18 +55,16 @@ class MklAddNOp : public OpKernel {
     bool input2_in_mkl_format = mkl_context.input2_shape.IsMklTensor();
 
     // handle the case of a scalar
-    if (!input1_in_mkl_format && !input0.dims()) {
+    if (!input1_in_mkl_format && input0.dims() == 0) {
       const TensorShape& o_shape = input0.shape();
       Tensor* out_tensor = nullptr;
       mkl_context.output_shape.SetMklTensor(false);
       AllocateOutputSetMklShape(ctx, 0, &out_tensor, o_shape,
                                 mkl_context.output_shape);
-      void* out_o = const_cast<void*>(static_cast<const void*>(
-                                out_tensor->flat<T>().data()));
-      float* user_i1 = const_cast<float*>(input0.flat<T>().data());
-      float* user_i2 = const_cast<float*>(input1.flat<T>().data());
-      (static_cast<T*>(out_o))[0] =
-          std::plus<float>{}(user_i1[0], user_i2[0]);
+      float user_i1 = (input0.scalar<T>()());;
+      float user_i2 = (input1.scalar<T>()());;
+      out_tensor->scalar<T>()() =
+          std::plus<float>{}(user_i1, user_i2);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


drpngx:
@rmlarsen could you take a look?

caisq:
Gentle ping, @rmlarsen

tensorflow/core/kernels/mkl_aggregate_ops.cc
```diff
@@ -54,17 +54,63 @@ class MklAddNOp : public OpKernel {
     GetMklShape(ctx, 1, &(mkl_context.input2_shape));
     bool input2_in_mkl_format = mkl_context.input2_shape.IsMklTensor();
 
+    // handle the case of a scalar
+    if (!input1_in_mkl_format && !input0.dims()) {
```
rmlarsen: prefer input0.dims() == 0.

tensorflow/core/kernels/mkl_aggregate_ops.cc
```diff
@@ -54,17 +54,63 @@ class MklAddNOp : public OpKernel {
...
+      const TensorShape& o_shape = input0.shape();
+      Tensor* out_tensor = nullptr;
+      mkl_context.output_shape.SetMklTensor(false);
+      AllocateOutputSetMklShape(ctx, 0, &out_tensor, o_shape,
+                                mkl_context.output_shape);
+      void* out_o = const_cast<void*>(static_cast<const void*>(
```
rmlarsen: You can use tensor->scalar<T>()() to access the scalar value (both lvalue and rvalue), instead of all the ugly casting.

tensorflow/core/kernels/mkl_aggregate_ops.cc
```diff
@@ -54,17 +54,63 @@ class MklAddNOp : public OpKernel {
...
         : input1.dims();
+
+    // If there is nothing to compute, return.
+    if (!input1_in_mkl_format && !input2_in_mkl_format) {
+      const TensorShape& o_shape = input0.shape();
+      if (o_shape.num_elements() == 0) {
+       Tensor* out_tensor = nullptr;
+       mkl_context.output_shape.SetMklTensor(false);
+       AllocateOutputSetMklShape(ctx, 0, &out_tensor, o_shape,
+                                mkl_context.output_shape);
```
rmlarsen: bad indent

tensorflow/core/kernels/mkl_aggregate_ops.cc
```diff
@@ -246,15 +259,15 @@ class MklAddNOp : public OpKernel {
       bool input1_in_mkl_format = input1_shape.IsMklTensor();
       bool input2_in_mkl_format = input2_shape.IsMklTensor();
       dnnDelete_F32(Eltwise);
+      if (!input1_in_mkl_format || !input2_in_mkl_format) {
```
rmlarsen: This condition should be 

if (input1_in_mkl_format || input2_in_mkl_format) {


To be the opposite of line 81.

tensorflow/core/kernels/mkl_aggregate_ops.cc
```diff
@@ -113,44 +159,11 @@ class MklAddNOp : public OpKernel {
...
-
  private:
   typedef struct {
     int in_dims;
-    size_t* in1_sizes;
-    size_t* in1_strides;
-
-    size_t* in2_sizes;
-    size_t* in2_strides;
+    size_t* in_sizes;
```
rmlarsen: set both to nullptr either here or in initializer list of constructor.

rmlarsen (changes requested):
Please address previous comments.

tensorflow/core/kernels/mkl_aggregate_ops.cc
```diff
@@ -81,11 +79,11 @@ class MklAddNOp : public OpKernel {
     if (!input1_in_mkl_format && !input2_in_mkl_format) {
       const TensorShape& o_shape = input0.shape();
       if (o_shape.num_elements() == 0) {
-       Tensor* out_tensor = nullptr;
-       mkl_context.output_shape.SetMklTensor(false);
-       AllocateOutputSetMklShape(ctx, 0, &out_tensor, o_shape,
+        Tensor* out_tensor = nullptr;
+        mkl_context.output_shape.SetMklTensor(false);
+        AllocateOutputSetMklShape(ctx, 0, &out_tensor, o_shape,
                                 mkl_context.output_shape);
```
rmlarsen: Still missing an extra indent in line 85.

rmlarsen (changes requested):

rmlarsen:

tensorflow/core/kernels/mkl_aggregate_ops.cc
jinghuangintel: @rmlarsen this part is the opposite of line 173 and 182, only in the case that input 1 and input 2 are not in mkl format, we will need to allocate the array. so we will release the buffer only in the case these two are not in mkl format. Thank you ~ please let me know if you have any suggestions. :)

rmlarsen: haha, I guess I didn't look at the fully expanded view :-) Thanks for pointing that out.

rmlarsen (approved):

rmlarsen:

rmlarsen:
@agramesh1 thanks for the improvement.

PR #13022, opened 2017-09-13
Give accumulate_n op a gradient

frreiss:
This pull request addresses issue #10607 by adding a gradient to the existing `accumulate_n` operator. I followed the approach suggested by @alextp: rewrite `accumulate_n` as an atomic op which has a gradient defined for it and which gets rewritten by the runtime into the current implementation. Previously, this op had been implemented in Python as a constellation of lower-level ops, some of which are not differentiable.

**Implementation Details**
I have added a new C++ op, `AccumulateN`, which serves as a placeholder for type inference and gradient computation. A new rewrite, implemented in `accumulate_n_optimizer.cc`, replaces this placeholder with a group of `AssignAdd` ops and some additional ops that create, initialize, and destroy temporary variables.

The original Python code for `accumulate_n` has been replaced by a function that validates its arguments and creates an instance of the `AccumulateN` placeholder op.

**Testing**
I added a more complete set of tests for `accumulate_n` in a previous pull request (https://github.com/tensorflow/tensorflow/pull/12196) to ensure that the op would still be correct after the changes in the current pull request. I also added one additional test to verify that `accumulate_n` now has a gradient. All the tests under `//tensorflow/python/...` currently pass on my MacOS and Linux test machines.

**Things to Note**
The semantics of the new implementation are broadly the same as the original, with the exception of one corner case. The original implementation allowed all the inputs to `accumulate_n` to have an undefined shape. My new code requires that at least one input have a defined shape; or that the user provides a shape using the `shape` argument to the `accumulate_n` function.

While implementing the `AccumulateN` op, I noticed that the code to do the kind of shape initialization I needed was repeated verbatim at several places in the TensorFlow code base. I refactored this code into a new function `shape_inference::ExplicitShape()` and replaced all the existing copies.


tensorflow/core/common_runtime/accumulate_n_optimizer.cc
```diff
@@ -0,0 +1,204 @@
...
+
+  // Conveniently, all numeric data types have 0x0 == zero.  Otherwise we would
+  // need a giant switch statement here.
+  memset(const_cast<char*>(tensor.tensor_data().data()), 0,
+         tensor.tensor_data().size());
+
+  return tensor;
+}
+
+// TEMPORARY debugging code
```
alextp: Remove comment and debugging code

tensorflow/python/ops/math_ops.py
```diff
@@ -2041,48 +2041,38 @@ def accumulate_n(inputs, shape=None, tensor_dtype=None, name=None):
...
+
+  # tensor_dtype is for safety only; operator's output type computed in C++
+  if tensor_dtype is not None and tensor_dtype != inputs[0].dtype:
     raise TypeError("tensor_dtype is {}, but input is of type {}"
                     .format(tensor_dtype, inputs[0].dtype))
-  if len(inputs) == 1:
+
+  if len(inputs) == 1 and name is None:
     return inputs[0]
-  with ops.name_scope(name, "AccumulateN", inputs) as name:
```
alextp: Can we separate the change which adds the C++ infrastructure from the change which switches the default implementation of accumulate_n?

For now, let's add the new op in contrib somewhere.

Tensorflow tries to maintain forward compatibility (new graphs should work on C++ runtimes which are two weeks old) and adding the new op and switching to use it at the same time breaks it.

alextp (changes requested):

tensorflow/core/common_runtime/accumulate_n_optimizer.cc
frreiss: Ugh, forgot about that. Those two lines are removed now.

alextp: Thanks! Still waiting on the other change (related to forward compatibility)

tensorflow/python/ops/math_ops.py
frreiss: Sure, no problem. I can restructure the code so the new implementation is a different op with a different name and the old op remains unchanged.

frreiss:
I will be splitting this PR into two sets of changes: One change set with the refactoring changes around the `ExplicitShape()` function I added to `common_shape_fns.cc`; and a second change set that adds a new `AccumulateN`-like op to `contrib/framework`.

frreiss:
Closing this PR in favor of https://github.com/tensorflow/tensorflow/pull/13325

PR #13026, opened 2017-09-14
Get sysroot path in a robust way for nsync ios build

resec:
The sysroot path in nsync ios build is hardcoded. this PR changed it to use `xcrun`, just like how other build scripts did.


drpngx (approved):
Nice, thanks!

PR #13028, opened 2017-09-14
Update README.md

zjujeremy:


drpngx:
I'm not sure it's something that we want to merge. The comment has no place here.

PR #13036, opened 2017-09-14
delete dummy 'r' in front of annotation

wei-yuan:
I just delete this dummy r character in the front of multi-line annotation with the newest version.


wei-yuan:
I have signed it (CLA)

drpngx:
Hi @wei-yuan, the `r` is fine, it's just telling that the text is a raw string.

wei-yuan:
@drpngx thx for the reply

PR #13042, opened 2017-09-14
docs: Fix Anaconda environment creation step

stuarteberg:
There is a bug in the [install instructions for Anaconda users][1].  Before you can use `pip` to install tensorflow, you must first install `python` (which includes `pip`) to the new environment.  Otherwise, the user will accidentally use the system's `pip`, and tensorflow will be installed to the system python instead of the intended environment.

[1]: https://www.tensorflow.org/install/install_linux#InstallingAnaconda


stuarteberg:
I signed the CLA.

drpngx (approved):

PR #13049, opened 2017-09-14
Optimize batch matrix transposition for narrow matrices.

tjingrant:
This is an improved version of matrix transposition that specializes in the case when the matrices to be transposed are narrow. This is a preliminary implementation and I hope to get some feedback before further optimizing this implementation as it would likely entail great deal of efforts and patience.
- This implementation modifies the general matrix transposition kernel in a minimal way. Essentially this implementation enables the rectangular tile shapes to be used for transposition. Before, only square tiles are allowed.
- This implementation has a couple of specialized tile sizes and a very simple cost function is used to select among them during runtime. The logic behind this cost function is simple and will likely to be effective on a variety of platforms.
- We tested this implementation on problem sizes of {32, 64, 128, 256, 512, 1024} X range(16, 2048, 32) X range(2, 16) as well as {32, 64, 128, 256, 512, 1024} X range(2, 16)  X range(16, 2048, 32) where these problem size dimensions can be interpreted as batch_size, matrix_height, matrix_width correspondingly.  The average speedup is 16.7%. This experiment includes 10752 data points and is plotted using excel to indicate for what sub-space of all problem size space do we see speedups over baseline/existing implementation. See [picture](https://imgur.com/a/MyVOf). In the picture, problem sizes where we do see speedups are colored red and otherwise white. The three dimensional problem size space is collapsed to two by serializing the batch size dimension and matrix height dimension.
- As you can see from the current performance results, existing implementation is very good at dealing with large batches of tiny matrices. This is due to the fact that the baseline implementation uses a brute force (and nonetheless effective) way of dividing up workloads evenly which outperforms this commits. I may consider modify the cost function to use the baseline in these cases in the future.
- Again, this is only a preliminary implementation, I'm happy to make significant changes.


jlebar:
A few high-level thoughts:

* What kind of comments are you looking for on this PR?  I typed up a bunch of stuff below, but I realize now you may be looking for an entirely different kind of feedback.  Sorry if below isn't helpful.

* It seems like you have a benchmark for this.  Can you add it to the PR?  Otherwise it's going to be hard for anyone else to evaluate other changes they might want to make.

* Similarly, can you add the detailed benchmark results (beyond just the picture) to the commit message?

* This adds a *lot* of new kernels to compile, and that can have an effect on compilation speed and binary size, especially when, as is common, we have to compile the GPU code for multiple GPU archs.

I don't want to prematurely optimize or make this a PITA for you, especially if this ends up not being important.  Maybe a sufficient starting place would be: What's the change in compile time for this file?

* It seems like you'd want to add some additional tests to cover new edge cases in this code?

* In terms of formatting, please clang-format over this code so we don't have to argue about whitespace, and you'll want to revert some of the style changes, e.g. changing int template parameters from CamelCase to MACRO_CASE, adding (unnecessary parens) and so on.  A decent starting place is https://google.github.io/styleguide/cppguide.html, although that's unfortunately out of date.  :(

We prefer constants to macros, e.g. TILE_SIZE.  Also, I think you could rewrite the dispatch using templated functions rather than macros, which might be a nice simplification.

```
template <int foo, int bar>
bool DoTranspose(...) {
  if (...) {
    launch_kernel<<<...>>>();
    return true;
  }
  ...
  return false;
}
```

* Please add comments to the code and commit message explaining what this patch does.

tjingrant:
Sorry for this late reply. I was organizing my benchmarks into a presentable form. 
- I'm looking for comments that address whether this PR clears or can be improved to clear all the basic design rationales behind Tensorflow (My last point of the PR message euphemistically states that if this PR fails to clear such criteria then I'm willing to let it go). Some of my concerns are:
    - Method invocation is very very ugly.
    - Maybe TF dev team values maintainability well over performance.
And your comments do provide answers to my concerns and you showed me what metrics you hope to evaluate this PR with apart from performance (compilation time, binary size).
- I have uploaded my standalone benchmark [here](https://github.com/tjingrant/batch_matrix_transpose_benchmark) which includes a README for anyone interested in evaluating. Also a detailed performance evaluation report is included in the repo for data collected on K40 machine and P100 machine. Basically based on our observation, this PR on average gives 12% performance increase on K40 machine and 39% on P100. This benchmark also serves as a test suite as all results are cross-checked.
- I will eventually integrate the standalone benchmark to TF but for now the standalone version seems better suited for evaluating compilation times and binary size. I will add the performance figures to the commit message once implementation stabilizes.
- Compilation time is 5.6 seconds for float. Assuming this holds more or less the same for half and double this amounts to 16.8 seconds on our older 32 core Power8 machine with nvcc 8.0.61. Binary size is 2MB. I hope these figures provide reassurance.

jlebar:
> Sorry for this late reply.

No problem.  I really appreciate what you're doing here.  This is an important kernel, and it hasn't been loved as much as it deserves.

> Compilation time is 5.6 seconds for float. [...] Binary size is 2MB. I hope these figures provide reassurance.

Thanks you.  I think that should be fine.  If the compilation time becomes a problem, it sounds like we could reasonably easily split it up into three files, one for each data type (i.e. float/half/double).  But I also don't think it's necessary.

tjingrant:
@jlebar , can I have your comments on the latest commit? Some improvements include:
- Using recursive template functions to search and launch appropriate kernels.
- I noticed that in `transpose_op_test.py`, there are already tests in place to check the cases I'm trying to optimize. Specifically:
```
  def test3DGPU(self):
    # If no GPU available, skip the test
    if not test.is_gpu_available(cuda_only=True):
      return

    datatypes = [np.int8, np.float16, np.float32, np.float64, np.complex128]
    large_shapes = [[4, 1000, 3], [4, 1000, 8], [4, 1000, 13], [4, 3, 1000],
                    [4, 8, 1000], [4, 13, 1000]] * 3
    perms = [[0, 2, 1]] * 6 + [[2, 1, 0]] * 6 + [[1, 2, 0]] * 3 + [[2, 0, 1]
                                                                  ] * 3
```
The latest commit passes all the test cases present in `transpose_op_test.py`.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
                             input_dims_in_tiles[2];
-    SwapDimension1And2InTensor3UsingTiles<T, TileSize, NumSubTiles><<<
-        total_tiles_count, dim3(TileSize, NumSubTiles), 0, d.stream()>>>(
+
+    // We recusively search for the minimum pre-compiled configuration that
+    // satisfies the requested tile sizes.
+    BatchNarrowMatrixTransposeDispatcher<T, 32, 2>::DoBatchNarrowMatrixTranspose(
+        d, requested_tile_size_i, requested_tile_size_j, total_tiles_count,
         input, input_dims, output);
+
```
jlebar: Can we pull this `if` body out into a separate function?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
+                       input_dims[2] >= kMinDimensionToUseTiles);
+  bool narrow_matrix = (input_dims[1] >= kMinDimensionToUseRectTiles ||
+                      input_dims[2] >= kMinDimensionToUseRectTiles);
+  if (large_matrix) {
     // We get best performance when TileSize is the number of threads in a warp
     // (32 on our GPUs) and NumSubTiles is 8, so our block size is 8 * 32 = 256
     // threads.
     static const int TileSize = 32;
-    static const int NumSubTiles = 8;
+    static const int ThreadNum = 256;
```
jlebar: Hm, `ThreadNum` sounds to me like "this thread's number" rather than "number of threads".  But any reason not to call this `BlockSize`?

jlebar: Style guide says kFoo for compile-time constants.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
+    std::map<int, int> tile_spec_128 = {{32, 15}, {64, 15}, {128, 15},
+                                        {256, 2}};
+    std::map<int, int> tile_spec_64  = {{32, 15},  {64, 15}, {128, 15},
+                                        {256, 8},  {512, 2}};
+    std::map<int, int> tile_spec_32  = {{32, 15},  {64, 15}, {128, 15},
+                                        {256, 10}, {512, 4}, {1024, 2}};
+    std::map<int, int> tile_spec_16  = {{32, 15},  {64, 15}, {128, 15},
+                                        {256, 10}, {512, 4}, {1024, 2}};
+    std::map<int, int> tile_spec_8   = {{32, 15},  {64, 15}, {128, 15},
+                                        {256, 10}, {512, 4}, {1024, 2}};
```
jlebar: I'm concerned constructing all these std::maps may add a nontrivial overhead to kernel launches.  std::map is kind of evil.

As a first change, it seems like you could accomplish basically the same thing with fewer maps by doing something like:

```
  std::map<int, int> tile_spec = []() -> std::map<int, int> {
  switch (sizeof(T)) {
    case 1: return {{32, 15}, ...};
    case 2: ...
  }();
```

But looking closer, it seems that we also don't use the fact that this is a map, as opposed to, say, a `vector<pair<int, int>>`?  We could optimize further, but constructing one small vector is probably fine in terms of overhead.

What do you think of pulling this whole computation out into a separate function?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
+
+    std::map<int, int> tile_spec = tile_spec_map[8 * sizeof(T)];
+
+    int tile_long_side_len = 0;
+    int tile_short_side_len = 0;
+    float lowest_cost = std::numeric_limits<float>::max();
+    int data_long_side = max(input_dims[1], input_dims[2]);
+
+    for (std::map<int, int>::iterator it = tile_spec.begin();
+         it != tile_spec.end(); ++it) {
```
jlebar: Recommend a c++11 for-each loop.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
+      int proposed_tile_long_side_len = it->first;
+
+      // Threads that will not be doing anything useful when reading the matrix
+      // because the thread block size is bigger than the data block size.
+      float wasted_threads = (data_long_side -
+                              data_long_side / proposed_tile_long_side_len *
+                                  proposed_tile_long_side_len);
```
jlebar: All of the operands here appear to be `int`s, so does the LHS need to be a float?

jlebar: There's a math_util::FloorOfRatio function I think you can use.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
+      // launch as many threads
+      // as possible.
+      if (cost <= lowest_cost) {
+        tile_long_side_len = proposed_tile_long_side_len;
+        tile_short_side_len = it->second;
+        lowest_cost = cost;
+      }
+    }
+
+    // Request tile sizes such that the longer side of threadblock align with
```
jlebar: aligns

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
+    // the longer side of input data block to maximize read throughput.
+    // The ideal tile shape to request is one with its length of the shorter
```
jlebar: with the length

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,29 +595,131 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
+    int requested_tile_size_i = input_dims[1] >= kMinDimensionToUseTiles
+                                    ? tile_long_side_len
+                                    : input_dims[1];
+    int requested_tile_size_j = input_dims[1] >= kMinDimensionToUseTiles
+                                    ? input_dims[2]
+                                    : tile_long_side_len;
+
+    // Truncate the shorter size requested according to the manual limit set in
+    // tile_spec to make sure that we do not
+    // launch configurations violating hardware limits.
```
jlebar: Reflow comment

jlebar (commented):
Looking better.  Inspired by your code, I may have figured out a way to do this without macros -- have a look and let me know wyt.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +432,160 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+        SwapDimension1And2InTensor3UsingTiles<                                 \
+            TYPE, ThreadNum, SHORT_SIDE,                                       \
+            LONG_SIDE><<<total_tiles_count, ThreadNum, 0, d.stream()>>>(       \
+            input, input_dims, output);                                        \
+      return;                                                                  \
+    }                                                                          \
+  };
+
+#define BATCH_NARROW_MATRIX_TRANSPOSE_128(TYPE)               \
+  BATCH_NARROW_MATRIX_TRANSPOSE_LIMIT_OVERALL(TYPE, 256, 16); \
```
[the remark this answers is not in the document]
tjingrant: Hi, I think the point is to prevent tile size combinations like (256, 3) or (256, 4) to materialize as legitimate launch parameters. The compiler may not be aware that I always try doubling the first param (long side length) before incrementing the second (short side length) so it may attempt to jump from tile size combinations like (128, 3) to (256, 3) and compile a dispatcher that launches tile size combinations like (256, 3) which would be a nuisance because this breaks the shared memory limits and may result in a compile time error.

Also, I'm not sure if this is a concern: by template partial ordering, kernel<256, 2> is considered more specialized than kernel<256, anything> thus the implementation of kernel<256, 2> will override that of kernel<256, anything> thus a dispatcher for tile size combinations <256, 2> will materialize.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
tjingrant: Because in general the imaginary 2-dimensional thread blocks can change shape for reading and outputting matrices and therefore do not have a fixed size (essentially the effective_thread_num is the size of the imaginary block size, which can be different in the same kernel execution depending on whether you are reading or writing). I'm afraid that calling it BlockSize may induce some incorrect intuition.

jlebar:
(No problem that you're responding incrementally, but please ping me directly when you need my attention again so I don't accidentally leave you hanging.)

tjingrant:
Hi, @jlebar some more updates:
- Compile 1 set of dispatchers for each data size instead of each data type.
- Tried out your trick and got the tile size searching code to a much cleaner form.
- Refactor by singling out the cost analysis for narrow matrices transposition.
- Fixed misc issues raised.
Can I have your comments? Again, thanks for being so responsive!

tjingrant:
And a comment to your WDYT question regarding your suggestion of using enable::if to reduce cases down to 2:

I played around with your neat trick and was able to get a much cleaner presentation. However, fundamentally we can't get below three cases. The "base case" you mentioned is really the boundary case. Visualizing all the tile sizes combinations in a 2D space, we face three constraints that, when considered all together, would define the _valid combinations_ subset: physical limit on thread numbers, physical limit on shared memory and further limit on the tile area (product of 2 tile sizes) to increase parallelism. I conveniently called this boundary tile size possibility frontier, drawing the name from the concept of [production possibility frontier](https://en.wikipedia.org/wiki/Production%E2%80%93possibility_frontier). I think this is a very intuitive name even without the parallel drawn.

Thus with this frontier established, we have three cases:

- tile size combination lands in the interior of the frontier: we try to launch kernel and if request is not satisfied, we search for larger tile size combinations.
- tile size combination lands on the frontier: we launch regardless of whether request is satisfied, no search performed.
- tile size combination lands in the exterior of the frontier: even compiling such kernel would cause us trouble, thus specialize dispatcher to do nothing at all.

tjingrant:
Ping @jlebar , I have also included a performance note for all types on K40 for float type on P100.

jlebar:
Hi, sorry I haven't been able to get to this quickly.  I'll try to do the review by eow, things have just been a little crazy over here.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+
+  for (auto tile_size_pair : tile_spec) {
+    int proposed_tile_long_side_len = tile_size_pair.first;
+
+    // Threads that will not be doing anything useful when reading the matrix
+    // because the thread block size is bigger than the data block size.
+    int wasted_threads = (data_long_side -
+                          MathUtil::FloorOfRatio<int>(
+                              data_long_side, proposed_tile_long_side_len) *
+                              proposed_tile_long_side_len);
```
jlebar: I know it's picky, but I think it would be great to clang-format these commits.  For example, I'm pretty sure clang-format will pull the last line of this statement in by 4 spaces, which makes the computation much easier to follow.  Using a relatively new version of clang-format will also format the `<<<foo>>>` kernel launch nicely.

`git-clang-format` will let you format the commits without reformatting the whole file.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -175,34 +177,31 @@ __global__ void SwapDimension1And2InTensor3Simple(int nthreads, const T* input,
...
-
-  static_assert(TileSize % NumSubTiles == 0,
-                "TileSize must be divisible by NumSubTiles");
-  eigen_assert(blockDim.x == TileSize);
-  eigen_assert(blockDim.y == NumSubTiles);
+// In general, for best performance, you should probably set TileSizeI,
+// TileSizeJ equal to the number of threads in a warp (32 in nvidia GPUs).
+// With a TileSizeI, TileSizeJ of 32, NumThread of 128 or 256 seems to get
+// the best performance on K40 GPUs.
+template <typename T, int NumThread, int TileSizeI, int TileSizeJ>
```
jlebar: Nit, can we call it NumThreads?  `NumThread` sounds to me a lot like `ThreadNum`, which is a different thing.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -211,39 +210,50 @@ __global__ void SwapDimension1And2InTensor3UsingTiles(const T* input,
...
-    for (int y = y_start; y < y_end; y++) {
-      shared_memory_tile[y][x] = input[input_flat_index + y * input_dims[2]];
+  // Calculate effective thread number. This ensures that we use the largest
+  // number of threads available to form a regular thread block with no
+  // trailing incomplete lines.
+  int effective_thread_num = NumThread / TileSizeJ * TileSizeJ;
+
+  if (x < effective_thread_num) {
+    // Orient the logical thread block with respect to the input array.
+    // ie. align the contiguous dimension of thread blocks with contiguous
```
jlebar: with the contiguous dimension?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
   }
 };
 
+// TPF stands for Tile size Posibility Frontier. It denotes the tile size
+// combinations that consume the most computational resources constrained by
+// number of threads per SM limit, shared memory limit and overall degree of
+// parallelism.
```
jlebar: Maybe this would be easier to read if we made it a list:

```
... constrained by the limits on
  - the number of threads per SM
  - shared memory per thread block, and
  - overall degree of parallelism.
```

except (and this is actually the important part) I don't really understand what the last one means.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+// Tile size combinations lying on this frontier would achieve the maximum
+// utilization of available resources and combinations lying outside this
```
jlebar: , and

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+// frontier is not performance-wise profitable or simply not possible.
```
jlebar: s/is not/are not/

jlebar: Suggest rewording to e.g. "are either not possible, or are slower than the alternatives".

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+#define TPF_CHECKER(OP)                                                   \
+  size_of_t == 16 &&    (TileLongSide == 32   && (TileShortSide OP 4)  || \
```
jlebar: You went through all this work to indent this macro nicely, but then this first line is odd.  :)

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+                         TileLongSide == 1024 && (TileShortSide OP 2))
+
+constexpr bool TileSizePossibilityFrontierCheck(int TileLongSide,
+                                                int TileShortSide,
+                                                int size_of_t, int mode) {
+  // When mode is 0, this function checks whether the tile size combination lies
+  // on the tile size possibility frontier. When mode is 1, it checks whether
+  // the combination lies outside the frontier.
+  return mode == 0 ? TPF_CHECKER(==) : TPF_CHECKER(>);
+}
```
jlebar: I am kind of tempted to suggest that we get rid of this macro too:

```
// We want std::equal_to and std::greater, but they're not constexpr until C++14.
struct EqualTo {
  constexpr bool operator()(int a, int b) { return a == b; }
};
struct GreaterThan {
  constexpr bool operator()(int a, int b) { return a > b; }
};
template <typename Op>
constexpr bool TileSizeFrontierCheck(..., Op op) {
  return size_of_t == 16 && ((TileLongSide == 32 && op(TileShortSide, 4)) ||
                             (TileLongSide == 64 && op(TIleShortSide, 4)) || ...
                                            
}
constexpr bool TileSizeOnFrontier(...) { return TileSizeFrontierCheck(..., EqualTo()); }
constexpr bool TileSizePastFrontier(...) { return TileSizeFrontierCheck(..., GreaterThan()); }
```

In addition to getting rid of the macro, this also lets you avoid the sort of ugly `mode` param.

I also renamed it "tile size frontier", which I think gets the point across just as well as "tile size possibility frontier" with one less word.

WDYT?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+  static void DoIt(const GPUDevice& d, int tile_size_i,
+                                           int tile_size_j,
+                                           int total_tiles_count,
+                                           const T* input,
+                                           const Dimension<3>& input_dims,
+                                           T* output) {
+    assert(false &&
+           "BatchNarrowMatrixTransposeDispatcher has requested an unexpected "
+           "launch configuration. ");
+  }
```
jlebar: I get that we need this base case, but do we need to define `DoIt` here as well?  It would be better if we could statically ensure that we never call this `DoIt` instead of relying on a runtime check.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+      SwapDimension1And2InTensor3UsingTiles<
+          T, NumThread, TileLongSide,
+          TileShortSide><<<total_tiles_count, NumThread, 0, d.stream()>>>(
+          input, input_dims, output);
+    else if (tile_size_j <= TileLongSide && tile_size_i <= TileShortSide)
+      SwapDimension1And2InTensor3UsingTiles<
+          T, NumThread, TileShortSide,
+          TileLongSide><<<total_tiles_count, NumThread, 0, d.stream()>>>(
+          input, input_dims, output);
+    return;
```
jlebar: Unnecessary return?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+    if (tile_size_i <= TileLongSide && tile_size_j <= TileShortSide)
+      SwapDimension1And2InTensor3UsingTiles<
+          T, NumThread, TileLongSide,
+          TileShortSide><<<total_tiles_count, NumThread, 0, d.stream()>>>(
+          input, input_dims, output);
+    else if (tile_size_j <= TileLongSide && tile_size_i <= TileShortSide)
+      SwapDimension1And2InTensor3UsingTiles<
+          T, NumThread, TileShortSide,
+          TileLongSide><<<total_tiles_count, NumThread, 0, d.stream()>>>(
+          input, input_dims, output);
```
jlebar: Nit: Even though these are a single statement, we treat them as "multiline" and add braces.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+                                           int tile_size_j,
+                                           int total_tiles_count,
+                                           const T* input,
+                                           const Dimension<3>& input_dims,
+                                           T* output) {
+
+    static_assert((TileLongSide & (TileLongSide - 1)) == 0,
+      "The length of the longer side of the tile is always a power of 2.");
+
+    const int NumThread = TileLongSide;
```
jlebar: NumThreads

jlebar: Huh, I am surprised this works with regular `const` and not constexpr, since you use it as a template param.  Probably better to use constexpr to drive the point home, though.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+      else if (tile_size_j <= TileLongSide && tile_size_i <= TileShortSide) {
+        SwapDimension1And2InTensor3UsingTiles<
+            T, NumThread, TileShortSide,
+            TileLongSide><<<total_tiles_count, NumThread, 0, d.stream()>>>(
+            input, input_dims, output);
+      }
+      return;
+    }
+
+    // Kernel is not launched, meaning the launch configuration is not
```
jlebar: s/is/was/?

jlebar: Maybe we can be more precise in the part of the comment after the comma.  Right now it just repeats the fact that if we got here, `!request_satisfied`.  Fine, but what does that really mean?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+                                           const Dimension<3>& input_dims,
+                                           T* output) {
+
+    static_assert((TileLongSide & (TileLongSide - 1)) == 0,
+      "The length of the longer side of the tile is always a power of 2.");
+    bool request_satisfied = max(tile_size_i, tile_size_j) <= TileLongSide &&
+                             min(tile_size_i, tile_size_j) <= TileShortSide;
+
+    if (request_satisfied) {
+      const int NumThread = TileLongSide;
```
jlebar: Although this works, I think constexpr would be better than plain const, to emphasize the fact that we're going to use this as a template parameter so it must be knowable at compile-time.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+                              data_long_side, proposed_tile_long_side_len) *
+                              proposed_tile_long_side_len);
+
+    int num_full_tiles = MathUtil::FloorOfRatio<int>(
+        data_long_side, proposed_tile_long_side_len);
+
+    float cost = 0;
+
+    // However, if we can execute two or more full tiles, then we gladly
+    // accept any number of wasted thread and ignore its cost.
```
jlebar: threads

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+
+// A helper function to make RunSwapDimension1And2InTensor3 concise. This
+// helper function looks at the data type and input matrix sizes and decides
+// the thread numbers and tile sizes to use.
+template <typename T>
+void SwapDimension1And2InTensor3WithNarrowMatrices(
+    const GPUDevice& d, const T* input, const Dimension<3>& input_dims,
+    T* output, const int kMinDimensionToUseTiles) {
+  // Define available tile sizes here for each size of data type supported:
+  std::vector<std::pair<int, int>> tile_spec =
```
jlebar: nit: `auto tile_spec`, since we can see the type of the lambda.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+      case 2:
+        return {{32, 15}, {64, 15}, {128, 15}, {256, 15}, {512, 4}, {1024, 2}};
+      case 4:
+        return {{32, 15}, {64, 15}, {128, 15}, {256, 15}, {512, 4}, {1024, 2}};
+      case 8:
+        return {{32, 15}, {64, 15}, {128, 8}, {256, 4}, {512, 2}};
+      case 16:
+        return {{32, 4},  {64, 4},  {128, 4}, {256, 2}};
+    }
+    return {};
```
jlebar: Nit: I think this would be clearer as a `default`.  But also...should we have some static assert that checks that `sizeof(T) <= 16`?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+};
+
+// A helper function to make RunSwapDimension1And2InTensor3 concise. This
+// helper function looks at the data type and input matrix sizes and decides
+// the thread numbers and tile sizes to use.
+template <typename T>
+void SwapDimension1And2InTensor3WithNarrowMatrices(
+    const GPUDevice& d, const T* input, const Dimension<3>& input_dims,
+    T* output, const int kMinDimensionToUseTiles) {
+  // Define available tile sizes here for each size of data type supported:
```
jlebar: If this needs to be kept in sync with something else, we should probably say so...

In fact it would be really nice if we could autogenerate it from the frontier function?  It seems like this might actually be feasible, since the search space is small.  One only needs to compute it once; you can cache the result in a function-local static variable.

The trick is that the function-local static variable must be a POD, according to our style rules.  So a vector is not OK, but a vector* is fine.  For example:

```
// (Some meaningful comment...)
template <int SizeOfT>
const std::vector<std::pair<int, int>>& GetTileSizes() {
  static_assert(SizeOfT <= 16);
  static auto* sizes = [] {
    auto* sizes = new std::vector<pair<int, int>>();
    // Expensive work to populate sizes, lazily run in a thread-safe manner the first time GetTileSizes<N> is called.
    return sizes;
  }();
  return *sizes;
}
```

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+
+  int tile_long_side_len = 0;
+  int tile_short_side_len = 0;
+  float lowest_cost = std::numeric_limits<float>::max();
+  int data_long_side = max(input_dims[1], input_dims[2]);
+
+  for (auto tile_size_pair : tile_spec) {
+    int proposed_tile_long_side_len = tile_size_pair.first;
+
+    // Threads that will not be doing anything useful when reading the matrix
```
jlebar: s/Threads/Number of threads/?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+      tile_short_side_len = tile_size_pair.second;
+      lowest_cost = cost;
+    }
+  }
+
+  // Request tile sizes such that the longer side of threadblock aligns with
+  // the longer side of input data block to maximize read throughput.
+  // The ideal tile shape to request is one with its length of the shorter
+  // side of the tile being equal to the length of the shorter side of the
+  // input matrix.
```
jlebar: Suggest: "The ideal tile shape is one where the length of the shorter side of the tile is equal to the length of the shorter side of the input matrix."

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+                              ? tile_long_side_len
+                              : min(requested_tile_size_i, tile_short_side_len);
+  requested_tile_size_j = requested_tile_size_j == tile_long_side_len
+                              ? tile_long_side_len
+                              : min(requested_tile_size_j, tile_short_side_len);
+
+  Dimension<3> input_dims_in_tiles = {
+      input_dims[0],
+      (input_dims[1] + requested_tile_size_i - 1) / requested_tile_size_i,
+      (input_dims[2] + requested_tile_size_j - 1) / requested_tile_size_j,
```
jlebar: CeilOfRatio?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+      break;
+    case 8:
+      DISPATCH_WITH_UINT_TYPE(64);
+      break;
+    case 16:
+      BatchNarrowMatrixTransposeDispatcher<float4, 32, 2>::DoIt(
+          d, requested_tile_size_i, requested_tile_size_j, total_tiles_count,
+          reinterpret_cast<const float4*>(input), input_dims,
+          reinterpret_cast<float4*>(output));
+      break;
```
jlebar: I think we should static-assert that `alignof(T) >= alignof(uintXX / float4)`.  Otherwise it seems we could get into trouble.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,275 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+  };
+
+  int total_tiles_count =
+      input_dims_in_tiles[0] * input_dims_in_tiles[1] * input_dims_in_tiles[2];
+
+#define DISPATCH_WITH_UINT_TYPE(TYPE_SIZE)                                \
+  BatchNarrowMatrixTransposeDispatcher<uint##TYPE_SIZE, 32, 2>::DoIt(     \
+      d, requested_tile_size_i, requested_tile_size_j, total_tiles_count, \
+      reinterpret_cast<const uint##TYPE_SIZE*>(input), input_dims,        \
+      reinterpret_cast<uint##TYPE_SIZE*>(output))
```
jlebar: Perhaps it would be worthwhile to have something like:

```
// TransposeElemType<N>::type is an N-byte type.  We use this to map types of the same size onto a single type -- for example, {int32, uint32, and float} all map to ElemType<4>::type.  This lets us compile just one transpose kernel for all N-byte types.
template <int ElemBytes> struct TransposeElemType;
template <> struct TransposeElemType<1> { using type = uint8; };
template <> struct TransposeElemType<2> { using type = uint16; };
template <> struct TransposeElemType<4> { using type = uint32; };
template <> struct TransposeElemType<8> { using type = uint64; };
template <> struct TransposeElemType<16> { using type = float4; };
```

Then we don't need this macro or switch statement: it becomes simply:

```
using ElemType = TransposeElemType<sizeof(T)>::type;
static_assert(alignof(T) >= alignof(ElemType));
BatchNarrowMatrixTransposeDispatcher<ElemType, 32, 2>::DoIt(...);
```

WDYT?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -422,23 +707,34 @@ void RunSwapDimension1And2InTensor3(const GPUDevice& d, const T* input,
...
-    static const int TileSize = 32;
-    static const int NumSubTiles = 8;
+    static const int kTileSize = 32;
+    static const int kNumThread = 256;
+
     Dimension<3> input_dims_in_tiles = {
-        input_dims[0], (input_dims[1] + TileSize - 1) / TileSize,
-        (input_dims[2] + TileSize - 1) / TileSize,
+        input_dims[0], (input_dims[1] + kTileSize - 1) / kTileSize,
+        (input_dims[2] + kTileSize - 1) / kTileSize,
```
jlebar: CeilOfRatio?

jlebar (commented):
This is getting close, as evinced by the fact that I'm now nit'ing the comments.  :)

tjingrant:
@jlebar can I have your feedback? 

I think previously I was using clang-format from clang-4.0 installation and it was having trouble formatting kernel launches. Switching to 5.0 seems to solve this problem.

The latest commit fixes the issues raised except for the base case problem. I commented below your comment.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,313 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+  static void DoIt(const GPUDevice& d, int tile_size_i, int tile_size_j,
+                   int total_tiles_count, const T* input,
+                   const Dimension<3>& input_dims, T* output) {
+    assert(false &&
+           "BatchNarrowMatrixTransposeDispatcher has requested an unexpected "
+           "launch configuration. ");
+  }
+};
+
+// This function tries to recover, in a brute force way, the frontier defined in
```
jlebar: Nit, as written it sounds like this isn't guaranteed to succeed?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,313 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+
+    assert(false &&
+           "The corresponding short side length of the largest long side "
+           "length has to be 2.");
+    return frontier;
+  }();
+  return *frontier;
+}
+
+template <int ElemBytes>
```
jlebar: Maybe have a brief comment?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -413,6 +429,313 @@ struct PadInput<GPUDevice, T, int, NDIMS> {
...
+
+// A helper function to make RunSwapDimension1And2InTensor3 concise. This
+// helper function looks at the data type and input matrix sizes and decides
+// the thread numbers and tile sizes to use.
+template <typename T>
+void SwapDimension1And2InTensor3WithNarrowMatrices(
+    const GPUDevice& d, const T* input, const Dimension<3>& input_dims,
+    T* output, const int kMinDimensionToUseTiles) {
+  // Define available tile sizes here for each size of data type supported:
+  auto tile_spec = GetTileSizesFrontier<sizeof(T)>();
```
jlebar: Using auto rather than const auto& will copy the vector, no?

jlebar (commented):
This looks really good to me.

(I'm not sure how to respond to this comment inline, since github shows it as as "outdated"?  Tips appreciate if you have any, sorry for starting a new thread.)

> Hi, getting rid of this DoIt is conceptually easy; however I'm unable to arrive at a clean solution.

I see, you're right, sorry for not realizing this earlier.  `if constexpr` is what you want, but it's...post C++11, I forget exactly which spec it made it into.  The only way to do it without `if constexpr` would be to add new specializations for the two edge cases, as you suggest.

Is specializing in that way a big pain?  We already have one specialization that we'd be able to get rid of, so the net is just one additional specialization.  The alternative is changing assert(false) to LOG(FATAL) (which we should do anyway for the other assert(false)) because basically nobody builds TF with asserts enabled.  But LOG(FATAL) is pretty scary, so if the alternative is not really awful, it seems maybe better to avoid the fatal log?

Maybe if you write the code I can stare at it and see if there's a way to simplify it somehow.

Thanks again for your patience here, I think this is going to be a really good change.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
tjingrant: Hi, this indeed has an unusually high chance of failure in the case when someone changes the tile spec and forgets to end the short side len definition with 2.

Although I did not consciously thought this way when writing this comment, I believe the wording did reflect my reservation.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
tjingrant: You are right.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -439,17 +439,22 @@ struct GreaterThan {
   constexpr bool operator()(int a, int b) const { return a > b; }
 };
 
-// Tile size posibility frontier denotes the tile size combinations that consume
-// the most computational resources constrained by
+// The performance subspace refers to a subspace of the tile size space consists
```
jlebar: "that consists of", or "consisting of"

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -490,14 +495,63 @@ constexpr bool TileSizeOnFrontier(int TileLongSide, int TileShortSide,
...
-constexpr bool TileSizePastFrontier(int TileLongSide, int TileShortSide,
-                                    int size_of_t) {
+constexpr bool TileSizeOutsideBoundary(int TileLongSide, int TileShortSide,
+                                       int size_of_t) {
   return TileSizePossibilityFrontierCheck(TileLongSide, TileShortSide,
                                           size_of_t, GreaterThan());
 }
+constexpr bool TileSizeOnBoundaryNotFrontier(int TileLongSide,
+                                             int TileShortSide, int size_of_t) {
+  // For a tile size combination (longside, shortside), lying the boundary
```
jlebar: lying on the boundary

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -439,17 +439,22 @@ struct GreaterThan {
...
-// utilization of available resources, and combinations lying outside this
-// frontier are either not possible, or are slower than the alternatives.
-
+// Tile size posibility frontier denotes the tile size combinations in that
+// consume the most computational resources given a particular long side len.
+//
+// It is worth noting that due to the discrete nature of tile size space, the
+// frontier does not constitute the boundary of the performant subspace. For
+// this reason, there are tile size combinations that lie on the boundary of
+// this performant subspace, but not on the frontier.
```
jlebar: Is it the "performant" or "performance" subspace?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
jlebar: This comment could now use an introductory paragraph, I think, so we don't just jump in with two definitions apropos of nothing.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -439,17 +439,22 @@ struct GreaterThan {
...
 };
 
-// Tile size posibility frontier denotes the tile size combinations that consume
-// the most computational resources constrained by
+// The performance subspace refers to a subspace of the tile size space consists
+// of all tile sizes satisfying the constraint of
 // - number of threads per SM limit,
 // - shared memory limit and
 // - some experimentally determined, type-specific constraint on the product of
 // two side lengths to increase grid-level parallelism.
```
jlebar: Maybe indent this line two spaces so it doesn't run into the line below it?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -439,17 +439,22 @@ struct GreaterThan {
...
+// Tile size combinations lying outside this subspace are either not possible,
+// or are slower than the alternatives
```
jlebar: Missing period

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
jlebar: Ah, I see.  Do you think it would be worthwhile to say that, so as to help out someone who might hit this failure?  Maybe something like: "Builds the frontier defined in TileSizePossibilityFrontierCheck as a vector of tile size combinations.  This will fail if ...".

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -631,6 +690,10 @@ const std::vector<std::pair<int, int>>& GetTileSizesFrontier() {
   return *frontier;
```
jlebar: I think the assert above needs to be a LOG(FATAL), for the reasons in my previous review.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -631,6 +690,10 @@ const std::vector<std::pair<int, int>>& GetTileSizesFrontier() {
...
 }
 
+// Helper structs to help determine which data types to use given the size of
```
jlebar: s/to help/to/

jlebar: s/data types/data type/

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -631,6 +690,10 @@ const std::vector<std::pair<int, int>>& GetTileSizesFrontier() {
...
+// the matrix data type. For each size of the data type, we only compile one
+// kernel of that size and every data type sharing the same size will also share
+// the same kernel (unless data alignment issues forbids so).
```
jlebar: Maybe we can replace the second sentence with: "A transpose of elements of size N will use a kernel which operates on an array of TransposeElemType<N>::type."

jlebar: It may be better not to mention alignment here, because as written it implies that if the alignment is wrong, we'll use some other data type, when in fact we'll fail to compile.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -548,46 +593,60 @@ struct BatchNarrowMatrixTransposeDispatcher {
...
+        (TileLongSide & (TileLongSide - 1)) == 0,
+        "The length of the longer side of the tile is always a power of 2.");
+
+    LaunchBatchNarrowMatrixTransposeKernel<T, TileLongSide, TileShortSide>(
+        d, tile_size_i, tile_size_j, total_tiles_count, input, input_dims,
+        output);
   }
 };
 
+template <typename T, int TileLongSide, int TileShortSide>
```
jlebar: Maybe add

  // Base case -- no `DoIt` function because it's illegal to do transposes when the tile sizes are outside of the boundary.

Although, just to check: Do you need this specialization?  I'd guess you maybe don't, because it should never be instantiated anymore.

tjingrant:
@jlebar Fixed issues raised. I think the comment becomes outdated once the associated code changes in the new commit.

This latest commit does the specialization as discussed.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
jlebar: AIUI the purpose of introducing the notion of the performant subspace is so that we can say "lies on the boundary of the subspace, but not the frontier."

But I think we can simplify this for readers.  What if we said something to the effect of what you had before about the frontier, and then defined the concept of "on the frontier in the short dim but not the long dim"?  (Or is it on the long dim but not the short dim?  Anyway.)

I think that would obviate a lot of the confusion I'm having reading this patch now, where I have to keep straight the "boundary" as separate from the "frontier", even though those words are synonymous in my mind, and even though the two divisions are closely related.

WDYT?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -490,14 +495,63 @@ constexpr bool TileSizeOnFrontier(int TileLongSide, int TileShortSide,
...
                                           size_of_t, EqualTo());
 }
-constexpr bool TileSizePastFrontier(int TileLongSide, int TileShortSide,
-                                    int size_of_t) {
+constexpr bool TileSizeOutsideBoundary(int TileLongSide, int TileShortSide,
+                                       int size_of_t) {
   return TileSizePossibilityFrontierCheck(TileLongSide, TileShortSide,
                                           size_of_t, GreaterThan());
 }
+constexpr bool TileSizeOnBoundaryNotFrontier(int TileLongSide,
```
jlebar: Do you mean the boundary of the performant/performance subspace?  If so I don't think this is clear -- in the definitions above, we define this subspace, but we never use the word "boundary", so here it's not clear to me what "boundary" refers to.

Maybe we can fix this as suggested above by removing the concept of this new subspace.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
tjingrant: I agree this way is much more economical.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
tjingrant: I believe I used the terms `boundary` and `interior` according to their topological definitions to save comments; but your way of phrasing it could indeed save even more.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
tjingrant: Indeed no need!

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -640,15 +641,12 @@ struct BatchNarrowMatrixTransposeDispatcher<
...
-    typename std::enable_if<TileSizeOutsideBoundary(TileLongSide, TileShortSide,
-                                                    sizeof(T)),
-                            void>::type> {};
-
 // This function tries to recover, in a brute force way, the frontier defined in
 // TileSizePossibilityFrontierCheck as a vector of tile size combinations.
+// Note that if one changes the frontier definition as in
+// TileSizePossibilityFrontierCheck and forgets to set the corresponding short
+// side len of the last legal long side len with just one possibility of 2, this
+// function will fail and produce an error.
```
jlebar: Doesn't really produce an error, but rather crashes the program (or "will raise a fatal error", if you prefer).

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -640,15 +641,12 @@ struct BatchNarrowMatrixTransposeDispatcher<
...
-template <typename T, int TileLongSide, int TileShortSide>
-struct BatchNarrowMatrixTransposeDispatcher<
-    T, TileLongSide, TileShortSide,
-    typename std::enable_if<TileSizeOutsideBoundary(TileLongSide, TileShortSide,
-                                                    sizeof(T)),
-                            void>::type> {};
-
 // This function tries to recover, in a brute force way, the frontier defined in
 // TileSizePossibilityFrontierCheck as a vector of tile size combinations.
+// Note that if one changes the frontier definition as in
```
jlebar: s/as in/in/

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -640,15 +641,12 @@ struct BatchNarrowMatrixTransposeDispatcher<
...
+// TileSizePossibilityFrontierCheck and forgets to set the corresponding short
+// side len of the last legal long side len with just one possibility of 2, this
```
jlebar: I don't know what "with just one possibility of 2" means -- can we reword?

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -681,19 +679,17 @@ const std::vector<std::pair<int, int>>& GetTileSizesFrontier() {
...
       }
     }
-
-    assert(false &&
-           "The corresponding short side length of the largest long side "
-           "length has to be 2.");
+    LOG(FATAL)
+        << "The corresponding short side length of the largest long side "
+           "length has to be 2.";
     return frontier;
```
jlebar: LOG(FATAL) is noreturn, so this should no longer be necessary.

tensorflow/core/kernels/conv_ops_gpu_3.cu.cc
```diff
@@ -452,10 +452,17 @@ struct GreaterThan {
...
-// - long side frontier: the union of the biggest tile size combinations for
+// It is instrumental to consider, for each data type, two subsets of the
+// corresponding frontier:
+// - long side frontier: the union of the biggest tile size combination for
 //   each legal long side len.
 // - non long side frontier: the frontier set minus the long side frontier.
+//
+// The following function TileSizePossibilityFrontierCheck will define the
+// frontier using only the long side frontier tile size combinations (since
+// one can easily extrapolate the entire frontier from this subset). It
```
jlebar: Since TileSizePossibilityFrontierCheck is nonrestrictive (meaning, the meaning of the sentence is the same if you leave it out), it should be set off with commas.

But probably a better way to write it would just be to say:

> TileSizePossibilityFrontierCheck defines the frontier ...

jlebar (approved):
\o/

tjingrant:
Hi @jlebar , thanks again for being so responsive. I was actually reviewing my last commit when your comments came in...

tjingrant:
Ping @jlebar , @yzhwang,
Unfortunately I think we've run into a conflict. By examining the code @yzhwang submitted, I think we have very similar ideas but the implementation @jlebar and I worked out here forms a super set of what @yzhwang has in his implementation. Notably from his comment:
```
// Each thread block operates on a single rectangle tile, where its width is
// kTileLength (we currently set it to 64) and its height is small_dim,
// We set the thread block's X dimension to be tile_num_per_block, and its Y
// and Z to be one.
```
and 
```
// When only one of the dimensions is smaller than kMinDimensionToUseTiles,
// we use one block to process a rectangle region with the size of
// kTileLength * small_dim. We found that when set kTileLength to 64 on
// TitanX Maxwell GPU, it achieves the best performance.
```
A few things I would like to point out:
- this PR specializes the kernel launch parameters for each type.
- this PR also does not have a fixed long side len (in fact we have choices of {32, 64, 128...1024}).
- this PR also has a very nicely designed kernel dispatcher (thanks to @jlebar ) that works to select among multiple pre-compiled kernels.
- @yzhwang seems to have written another kernel but I simply generalized the SwapDimension1And2InTensor3UsingTiles kernel so it seems like my approach produces less maintenance overhead.
- this PR performs exhaustive experiments on 2 server-grade GPUs including P100 and K40 and thus the performance results will be more reliable and trust-worthy for performance-critical scenarios.

So I think if I get these facts right, I would like to propose the following to resolve the conflicts:
- Absorb suggestions from @yzhwang. Maybe there are reasons for having another kernel SwapDimension1And2InTensor3UsingTiles.
- Acknowledge in any ways we see fit the contribution of @yzhwang.
- Unfortunately, if the reasons I listed above are factually correct, I think the logical thing to do is to base everything in `conv_ops_gpu_3.cu.cc` on this PR and acknowledge @yzhwang 's contribution by applying his suggestions here.
- @yzhwang seems to have a more comprehensive test suite that should be used instead of mine.
- I hope there could be a mechanism for people inside and outside Google to know what each other is working on... at least for the open source part of Tensorflow.

tjingrant:
Ping @jlebar , @yzhwang,
Unfortunately I think we've run into a conflict. By examining the code @yzhwang submitted, I think we have very similar ideas but the implementation @jlebar and I worked out here forms a super set of what @yzhwang has in his implementation. Notably from his comment:
```
// Each thread block operates on a single rectangle tile, where its width is
// kTileLength (we currently set it to 64) and its height is small_dim,
// We set the thread block's X dimension to be tile_num_per_block, and its Y
// and Z to be one.
```
and 
```
// When only one of the dimensions is smaller than kMinDimensionToUseTiles,
// we use one block to process a rectangle region with the size of
// kTileLength * small_dim. We found that when set kTileLength to 64 on
// TitanX Maxwell GPU, it achieves the best performance.
```
A few things I would like to point out:
- this PR specializes the kernel launch parameters for each type.
- this PR also does not have a fixed long side len (in fact we have choices of {32, 64, 128...1024}).
- this PR also has a very nicely designed kernel dispatcher (thanks to @jlebar ) that works to select among multiple pre-compiled kernels.
- this PR has dynamically determined share memory sizes and therefore will be more shared-memory-efficient.
- @yzhwang seems to have written another kernel but I simply generalized the SwapDimension1And2InTensor3UsingTiles kernel so it seems like my approach produces less maintenance overhead.
- Similarly, can you explain why do you need two `__syncthreads()`?
- this PR performs exhaustive experiments on 2 server-grade GPUs including P100 and K40 and thus the performance results will be more reliable and trust-worthy for performance-critical scenarios.

So I think if I get these facts right, I would like to propose the following to resolve the conflicts:
- Absorb suggestions from @yzhwang. Maybe there are reasons for having another kernel SwapDimension1And2InTensor3UsingTiles.
- Acknowledge in any ways we see fit the contribution of @yzhwang.
- Unfortunately, if the reasons I listed above are factually correct, I think the logical thing to do is to base everything in `conv_ops_gpu_3.cu.cc` on this PR and acknowledge @yzhwang 's contribution by applying his suggestions here and make explicit attributions in comment. 
- @yzhwang seems to have a more comprehensive test suite that should be used instead of mine.
- I hope there could be a mechanism for people inside and outside Google to know what each other is working on... at least for the open source part of Tensorflow.

At the end of the day, I hope that this conflict can be resolved in a way that pays acknowledgement to every one involved in a fair way. It is genuinely awful see anyone's effort made in vain and I hope we can work towards preventing that from happening.

jlebar:
> I hope there could be a mechanism for people inside and outside Google to know what each other is working on... at least for the open source part of Tensorflow.

Yeah, this is disappointing for me, too.

I'm not sure what we should do to avoid this in the future -- @hawkinsp, do you have any thoughts?  Should I announce a big change like this on our internal TF mailing list so people know we're working on it?  I guess I figured this code hadn't been reworked in so long, the chances were small that we'd midair like this...

yzhwang:
Hi @tjingrant Thanks for your changes! I have been working on-and-off on this for a long time. The original submission got roll-backed since I spotted a bug for large batch size with one large dimension. Recently I fixed the bug and resubmitted the code.
I agree we should have some mechanism letting people in the open source know what we are working on. Usually we submit smaller change lists with quicker cycles. I don't think our case is very common.
I will now study your design and implementation and run some tests and benchmark internally. I will keep you updated on this.
If your implementation is a super set of my kernel and it always performs better, I'd be happy to assist you get this in.
Again, thanks for your work on this.

tjingrant:
@yzhwang great to see you here! Now we can get the conversation starting and move forward! If you look into my implementation, beware of a bug introduced by the later commits that causes redundant kernel executions and somehow slipped through the tests because it doesn't interfere with the results.

I also looked into your design and you certainly raise the baseline to a new height! As I put in my edit, I no longer believe that what you did is a subset of my implementation. Your specializations prove useful for later architectures (which I did not specifically target since from comments of your code, I inferred that K40 is what you care about and this implementation is indeed faster on K40 using your benchmarks) and I'm still in the process of figuring out which are the most important factors contributing to the better performance. That being said, it would be nice to take the best from both of our implementation and properly recognize these contributions.

Also, I suspect that you are trying to tackle the possibility that y dimension of the grid may overflow. I'm curious why not just use the x dimension of the grid as the general case does?

yzhwang:
Hi @tjingrant ! Using x dimension of the grid will simplify the implementation (we don't need to handle multiple batches per block any more. I put batch size to y dimension at first because I didn't realize we would have batch number larger than 65536, then when someone reported such bug, I tried to make as few changes as possible to fix that bug. Hence the current implementation. Also, the two _syncthreads() is likely redundant, currently it solves a bug for certain cases. I will root cause this and eventually figure out a better way.
As for benchmark, I will do it on all the platforms we care internally and report back. We can decide then how to proceed. I'm sure we can figure out a way to integrate our work into TensorFlow.

tjingrant:
@yzhwang I think this problem may not matter (batch size > 65536). Whoever having such trouble is not using a device compatible with TF documentation from what I understand. [Here](https://www.tensorflow.org/versions/r0.12/get_started/os_setup) says TF requires compute capability to be >=3.0 and [here](http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities) says that any device with compute capability 3.0 or higher can accommodate a grid x-dim of 2^32-1. This is a limit that is very hard to go beyond. Can you verify this?

yzhwang:
@tjingrant Yes you are right. I know for gridDim.x we can go as far as 2^32-1, switching it though requires more changes to the kernel and I currently don't have too much time to look at. Anyway, let me finish testing and benchmark on your PR first. We can always keep improving the kernel.

tjingrant:
@yzhwang thanks for your feedback, I realized that you did not intentionally put it on the y dim which explains.

yzhwang:
Hi @tjingrant Sorry for the late reply. I spent some time understanding your implementation. I really like how it provides a unified version of doing transpose with tiling and shared memory. Thanks a lot for the work! After some internal discussion with my colleagues, we would like to integrate your implementation.

So I did some benchmarks using: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/transpose_benchmark.py and we found that comparing with our current implementation, the performance of your implementation is not always better. For float64, float32, int8, and float16, they all have some input shapes that showed regression with your implementation. I did the benchmark on a GTX 1080 card. That being said, it is likely due to that you are not tuning the consts in TileSizePossibilityFrontierCheck() for that specific architecture. I will try to benchmark this on more architectures this week and let you know.

Since checking in your code as it is would cause some performance regression, going forward, we should work together to make sure that this unified tiling transpose implementation could provide the best performance on every architecture (or most architectures) we care about. I hope you could keep working on this with us. So my plan is to first figure out a better tuning strategy so that it could work better on both Pascal and K40/K80. Then we can improve it further for Volta in another PR. I know it might be hard for you to work on this given the hardware constraint, I will try my best to assist you with the benchmarks. Right now, it would be really nice if you could share some of your thoughts on tuning for different architectures. Thank you!

tjingrant:
@yzhwang Thanks for getting back to me! Glad to know you are interested in my implementation.

I sure will be working closely with the TF team. But now I'm temporarily busy with other things our team is working on related to TF for this week so likely I'll get back to you with more substance some time next week. From my experience, the frontier indeed needs to be tuned. However the kernel also needs some tweaking. I noticed that on P100, cuda does not realize that I'm reading contiguous memories possibly because of the thread ID relabeling process. With nvprof, the resulting shared memory store transactions are significantly higher. I'm not very sure about the exact cause of it yet because we have a custom cuda driver that may break things. Regarding the hardware constraint, I have access to all the machines you are interested in but the constraint is that I may have a slightly customized driver and therefore it would be useful for you to post your performance results so that we can cross-check.

Again thanks a lot for actively resolving the conflict!

tjingrant:
@yzhwang Hi, sorry for my digression, I'm more or less done with other works and will be working on tuning this patch now. As explained in my previous comment, can you share your performance results on the P100 machine so that I can cross-check whether my driver is interfering with the results? (It has been in the past).

yzhwang:
Hey @tjingrant ! Thank you! I've sent the performance comparison sheet to your gmail.

martinwicke:
@tjingrant @yzhang87 Any progress on this?

tjingrant:
@martinwicke hi, this is not stalled, it just takes time.

yzhwang:
@tjingrant Thanks for the work! Let me know if you need any help from my side too.

drpngx:
@tjingrant any luck with this?

tjingrant:
Yes, I was able to identify the major culprit behind the performance degradation and I'm applying my fix. 

Specifically, by examining the generated assembly code, I think the performance degradation compared to @yzhwang 's implementation comes down to these lines:
```
int ti = x / TileSizeJ;
int tj = x % TileSizeJ;
```
I think due to these computations to obtain a 2D thread ids, the compiler is unable to recognize that I'm actually accessing contiguous memory regions within a warp. I'm trying to make it more obvious for the compiler to optimize for such memory access. I'll try to push some changes between today and tomorrow.

drpngx:
Strange, there's some issue with e4659c4fcbf64b96c5f389519d3d87416bf8bc70. Could you check the email? This is only a two-line change so it can't be the previous comment.

tjingrant:
@drpngx what problems specifically? CI build failures are all Google internal CI failures; they seem to be caused by failure to start the build process.

drpngx:
OK, the bot appears to have approved the change. Let's see.

drpngx:
Jenkins, test this please.

drpngx:
For some reason the Windows build didn't kick off, either transient or we ran out of executors. Trying again.

Jenkins, test this please.

drpngx:
Jenkins, test this please.

PR #13008, opened 2017-09-13
EHN: csv supports missing value

facaiy:
### What changes were proposed in this pull request?

`tf.decode_csv` support NA values, see #13007 

### How was this patch tested?

+ [x] add unit tests.
+ [ ] pass all tests.

PR #13051, opened 2017-09-14
Revert "EHN: csv supports missing value (#13008)"

gunan:
This reverts commit 36c5f218d00c4bdee296974b7a7e4d95da8a88a7.

Original change needs to go through API review.


yifeif (approved):
Thanks for fixing the build!

PR #13055, opened 2017-09-15
Add support of `drop_negatives` for `tf.unsorted_segment_sum`

### LINKED ISSUE #478: Extend tf.unsorted_segment_sum to allow 'rejecting' entries
As already discussed in #466, it would be useful to be able to mark entries which are not to be summed anywhere by assigning negative indexes (or just -1) there. @girving suggested, that this behavior should only be enabled when setting a flag (e. g. `drop_negatives`) to preserve the existing error detection behavior, which I support.

yongtang:
This fix tries to address the issue raised in #478 by adding the support of `drop_negatives` for `tf.unsorted_segment_sum` so that it is possible to skip entries (when index = -1)

This fix fixes #478.


tensorflow/core/kernels/segment_reduction_ops.cc
```diff
@@ -470,9 +498,18 @@ class UnsortedSegmentSumOp : public UnsortedSegmentBaseOp<Device, T, Index> {
   explicit UnsortedSegmentSumOp(OpKernelConstruction* context)
       : UnsortedSegmentBaseOp<Device, T, Index>(
             context,
-            sum_functor_) {}
+            &sum_functor_) {
+    bool drop_negatives = false;
+    OP_REQUIRES_OK(context, context->GetAttr("drop_negatives", &drop_negatives));
+    if (drop_negatives) {
+      this->reduction_functor_ = &this->sum_with_drop_negatives_functor_;
```
drpngx: Do we need `this->`? If not, please remove.

tensorflow/core/kernels/segment_reduction_ops.h
```diff
@@ -78,6 +78,24 @@ struct UnsortedSegmentSumFunctor: public UnsortedSegmentBaseFunctor<Device, T, I
                   typename TTypes<T, 2>::Tensor output);
 };
 
+// Functor for UnsortedSegmentSumWithDropNegativesOp.
+// 'output_rows': the number of output segments (unique segment ids in
```
drpngx: Remove quotes unless they're there for consistency.

drpngx (dismissed):

drpngx:
This will probably require an API review.

At a high level, could we use `tf.where` with the simple `reduce_sum`?

tensorflow/core/kernels/segment_reduction_ops.cc
yongtang: Thanks @drpngx. Done.

tensorflow/core/kernels/segment_reduction_ops.h
yongtang: Thanks, updated.

yongtang:
Thanks @drpngx for the review. I addressed some of the comments. For using `tf.where` with `reduce_sum`, I am still playing with it.

However, I noticed that at the moment, the behavior of CPU and GPU is actually different. On CPU, negative indices will through out an error while on GPU negative indices will be ignored silently.

In other words, on GPU the behavior of `tf.unsorted_segment_sum` is already `drop_negatives=True` even before this PR.

I am wondering if it make sense to just adjust the behavior on CPU so that negatives will be dropped? 

This will bring the behavior of GPU and CPU to the same without introducing another flag of `drop_negatives=True`.

Below is the output for GPU and CPU without this PR:
```python
>>> import tensorflow as tf
>>> print tf.__version__
1.3.0
>>> with tf.device('/gpu:0'):
...   v = tf.unsorted_segment_sum([1,2], [1, -1], 5)
KeyboardInterrupt
>>> 
>>> import numpy as np
>>> with tf.device('/gpu:0'):
... v = tf.unsorted_segment_sum(np.asarray([1,2], dtype=np., [1, -1], 5)
KeyboardInterrupt
>>> 
>>> 
>>> with tf.device('/gpu:0'):
...   v = tf.unsorted_segment_sum([1.0, 2.0], [1, -1], 5)
...   tf.Session().run(v)
... 
....
array([ 0.,  1.,  0.,  0.,  0.], dtype=float32)
>>> with tf.device('/cpu:0'):
...   v = tf.unsorted_segment_sum([1.0, 2.0], [1, -1], 5)
...   tf.Session().run(v)
... 
2017-09-19 21:26:50.961788: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1055] Creating TensorFlow device (/device:GPU:0) -> (device: 0, name: Tesla K80, pci bus id: 0000:00:1e.0, compute capability: 3.7)
...
...
InvalidArgumentError (see above for traceback): segment_ids[1] = -1 is out of range [0, 5)
	 [[Node: UnsortedSegmentSum_1 = UnsortedSegmentSum[T=DT_FLOAT, Tindices=DT_INT32, _device="/job:localhost/replica:0/task:0/cpu:0"](UnsortedSegmentSum_1/data, UnsortedSegmentSum_1/segment_ids, UnsortedSegmentSum_1/num_segments)]]

>>> 
```

drpngx:
Requesting API review. To rephrase, see @yongtang comment: GPU and CPU behave differently, one dropping negatives and the other one doesn't. We can leave the behavior undefined until we use `drop_negatives`, or just decide on a common behavior (error out or silently skip).

martinwicke:
Since the behavior on GPU is `drop_negative=True`, and since (IIRC) the current behavior on CPU is an error, we are within our right to change the behavior on CPU to match GPU (`drop_negative=True`). I think this alone would address #478. 

I do not know whether we would still want the extra argument. Does anyone require `drop_negative=False` (which does not result in an error)?

yongtang:
Thanks @drpngx @martinwicke for the review. I updated the PR so that now negative indices will be dropped silently, bring in behaviors of GPU and CPU closer.

I do notice another discrepancy between GPU and CPU. In GPU the bad index (> num_segments) are ignored as well while in CPU greater than num_segments will results in error. Not sure the best way to handle this situation. Any suggestions about that?

martinwicke:
Can you edit the docs in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc#L1681 to point out the negative-dropping behavior?

yongtang:
Thanks @martinwicke. The PR has been updated with the documentation added in `math_ops.cc`.

martinwicke (approved):

sb2nov:
Jenkins, test this please.

yongtang:
I tried to produce locally for the python 3 failure `//tensorflow/contrib/learn:dnn_linear_combined_test` but it works fine. Maybe it is transient?

drpngx:
Jenkins, test this please.

josh11b:
Approval for API review.

yongtang:
Thanks all for the review. The Windows build failure is likely to be  transient (didn't failure in the previous runs). I think a rerun might fix it.

drpngx:
Jenkins, test this please.

yongtang:
The `Linux CPU Tests Makefile` is the only test failing. It might be unrelated (also saw the same failure on other PRs).

drpngx:
Yes, but I still want to see it run.

Jenkins, test this please.

PR #13056, opened 2017-09-15
Update mnist beginners softmax variables

maspwr:
The usage of variable 'x' when describing the softmax function can be confusing to newcomers since 'x' is used immediately prior to represent the unweighted input. Continuing the usage of 'evidence' helps with continuity.


drpngx:
@nealwu what do you think?

nealwu (approved):
Sure, this seems reasonable.

PR #13070, opened 2017-09-15
Use AllClose instead of AllEqual in layers tests

nluehr:
- While simple implementations of convolution (e.g., gemm-based) will
  produce exact results in these tests, general implementations
  only guarantee floating-point precision.
- (One of these tests was previously observed to fail with a max error
  of 1e-7).


drpngx:
Where are you seeing this error? If we introduce this, I'd like to set the tightest tolerance bounds so that we can track precision changes over time.

sb2nov:
@nluehr gentle ping?

nluehr:
Sorry for the delay. We are no longer able to reproduce the test failures on the master branch using either cuda 9/cudnn 7 or cuda 8/cudnn 6. So I'll Close the PR.

PR #13071, opened 2017-09-15
Unreachable input gradients

theflofly:
This PR checks if all the inputs are reachable from the outputs during AddSymbolicGradients.
@skye would be the best person to review this task as she asked me to add a TODO during a previous PR.

### DIFF
--- a/tensorflow/cc/framework/gradients.cc
+++ b/tensorflow/cc/framework/gradients.cc
@@ -175,8 +175,12 @@ Status SymbolicGradientBuilder::Initialize() {
         "Must specify a gradient input for each output.");
   }
   std::vector<bool> reachable_nodes = GetReachableNodes();
-  // TODO(theflofly) Check that inputs_ are reachable from
-  // outputs_ using reachable_nodes
+  for (const Output& nout : inputs_) {
+    if (!reachable_nodes[nout.node()->id()]) {
+      return errors::InvalidArgument(
+        nout.node()->name() + " is unreachable from the output(s).");
+    }
+  }
   grad_outputs_->clear();
   grad_outputs_->resize(inputs_.size());
   // Populate `output_nodes_` from node ids in `outputs_`.
@@ -198,6 +202,12 @@ Status SymbolicGradientBuilder::Initialize() {
     std::unordered_set<Node*> visited;
     std::deque<Node*> queue;
     for (const Output& nout : inputs_) {
+      LOG(INFO) << "Input: " << nout.node()->id();
+      LOG(INFO) << "Input: " << nout.node()->name();
+      if (!reachable_nodes[nout.node()->id()]) {
+        return errors::InvalidArgument(
+          nout.node()->name() + " is unreachable from the output(s).");
+      }
       if (visited.find(nout.node()) == visited.end()) {
         queue.push_back(nout.node());
         visited.insert(nout.node());
--- a/tensorflow/cc/framework/gradients.cc
+++ b/tensorflow/cc/framework/gradients.cc
@@ -178,9 +178,9 @@ Status SymbolicGradientBuilder::Initialize() {
   for (const Output& input : inputs_) {
     if (!reachable_nodes[input.node()->id()]) {
       return errors::InvalidArgument(
-        "Cannot compute the partial derivative for node '"
-        + input.node()->name()
-        + "' as it's unreachable from the output node(s).");
+        "Cannot compute the partial derivative for node '",
+        input.node()->name(),
+        "' as it's unreachable from the output node(s).");
     }
   }
   grad_outputs_->clear();
--- a/tensorflow/cc/framework/gradients.cc
+++ b/tensorflow/cc/framework/gradients.cc
@@ -175,10 +175,12 @@ Status SymbolicGradientBuilder::Initialize() {
         "Must specify a gradient input for each output.");
   }
   std::vector<bool> reachable_nodes = GetReachableNodes();
-  for (const Output& nout : inputs_) {
-    if (!reachable_nodes[nout.node()->id()]) {
+  for (const Output& input : inputs_) {
+    if (!reachable_nodes[input.node()->id()]) {
       return errors::InvalidArgument(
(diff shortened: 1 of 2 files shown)


tensorflow/cc/framework/gradients_test.cc
```diff
@@ -431,6 +431,39 @@ TEST_F(GradientsTest, UnreachableEdgeGradTwoOutputs) {
       outputs[0], test::AsTensor<double>({17.5, 24.7, 26.8}, {3, 1}), 1e-5);
 }
 
+TEST_F(GradientsTest, UnreachableInput) {
+  auto x = Variable(scope_test_, {2, 3}, DT_DOUBLE);
+  auto x_const = Const(scope_test_, {{1.0, 2.0, 3.0}, {4.0, 5.0, 6.0}});
+  auto x_assign = Assign(scope_test_, x, x_const);
```
skye: Do you need to use the Variable-Const-Assign pattern here, or can you just use Const or Placeholder nodes? I realize this is very similar to the discussion we had in the last PR, but I think in this case it'll be easier to understand and verify what the test is actually testing if you use simple inputs.

tensorflow/cc/framework/gradients_test.cc
```diff
@@ -431,6 +431,39 @@ TEST_F(GradientsTest, UnreachableEdgeGradTwoOutputs) {
...
+
+  auto z = Variable(scope_test_.WithOpName("z"), {1, 3}, DT_DOUBLE);
+  auto z_const = Const(scope_test_, {{9.0, 10.0, 11.0}});
+  auto z_assign = Assign(scope_test_, z, z_const);
+
+  auto m2 = MatMul(scope_test_, y, z);
+
+  auto dm1 = Const(scope_test_, {{0.5}, {0.5}});
+  auto dm2 =
+      Const(scope_test_, {{0.5, 0.5, 0.5}, {0.6, 0.7, 0.8}, {0.6, 0.7, 0.9}});
```
skye: You don't need dm2 right?

tensorflow/cc/framework/gradients_test.cc
```diff
@@ -431,6 +431,39 @@ TEST_F(GradientsTest, UnreachableEdgeGradTwoOutputs) {
...
+
+  std::vector<Output> grad_outputs;
+  // from m1 z is unreachable, an error status should be returned
+  //   m2  m1
+  //   |   |
+  //   *   *
+  //  / \ / \
+  // z   y   x
+  Status status = AddSymbolicGradients(scope_test_, {m1}, {z}, {dm1}, nullptr);
```
skye: Pass in &grad_outputs (we should probably DCHECK that it's not null or something)

tensorflow/cc/framework/gradients.cc
```diff
@@ -175,8 +175,12 @@ Status SymbolicGradientBuilder::Initialize() {
         "Must specify a gradient input for each output.");
   }
   std::vector<bool> reachable_nodes = GetReachableNodes();
-  // TODO(theflofly) Check that inputs_ are reachable from
-  // outputs_ using reachable_nodes
+  for (const Output& nout : inputs_) {
```
skye: nit: nout -> input

tensorflow/cc/framework/gradients_test.cc
```diff
@@ -431,6 +431,39 @@ TEST_F(GradientsTest, UnreachableEdgeGradTwoOutputs) {
...
+  EXPECT_EQ(status.code(), error::INVALID_ARGUMENT);
+  EXPECT_EQ(status.error_message(), "z is unreachable from the output(s).");
```
skye: I think this error message could provide some more context, maybe:
Cannot compute the partial derivative for node 'z' as it's unreachable from the output node(s).

skye (changes requested):
Thanks for picking this up, looks good overall, just a few nits :)

theflofly:
It should be okay.

theflofly:
Actually some lines exceeded 80 chars.

drpngx:
Jenkins, test this please.

tensorflow/cc/framework/gradients_test.cc
drpngx: Verified.

tensorflow/cc/framework/gradients_test.cc
drpngx: Verified. Since m2 was also no longer needed, eliminated that and renamed variables.

tensorflow/cc/framework/gradients_test.cc
drpngx: Verified.

tensorflow/cc/framework/gradients.cc
drpngx: Verified.

tensorflow/cc/framework/gradients_test.cc
drpngx: Verified.

tensorflow/cc/framework/gradients.cc
```diff
@@ -175,8 +175,14 @@ Status SymbolicGradientBuilder::Initialize() {
         "Must specify a gradient input for each output.");
   }
   std::vector<bool> reachable_nodes = GetReachableNodes();
-  // TODO(theflofly) Check that inputs_ are reachable from
-  // outputs_ using reachable_nodes
+  for (const Output& input : inputs_) {
+    if (!reachable_nodes[input.node()->id()]) {
+      return errors::InvalidArgument(
+        "Cannot compute the partial derivative for node '"
+        + input.node()->name()
```
skye: nit: don't use + to concat strings, you can just pass in multiple arguments: return InvalidArgument("foo", "bar")

skye (approved):

tensorflow/cc/framework/gradients.cc
theflofly: Indeed, I made the edit.

caisq:

PR #13105, opened 2017-09-17
Add implementation of `conv1d_transpose`

### LINKED ISSUE #8729: conv1d_transpose not implemented
I would like to perform a conv1d_transpose, but I can’t see any implementation in tensorflow.

I guess that inserting zeros between each values and then applying a regular conv1d would do the job ?

(principle explained hereafter with conv2d_tranpose : https://github.com/tensorflow/tensorflow/issues/4306)

But I can't find a way to implement this kind of operation with tensorflow :

import numpy as np
arr = np.arange(1, 10)                 # array([1, 2, 3, 4, 5, 6, 7, 8, 9])
np.insert(arr, slice(1, None), 0)  # array([1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8, 0, 9])

Any ideas ?

Thank you

yongtang:
This fix tries to address the issue raised in #8729 by providing the implementation of `conv1d_transpose`.

This fix fixes #8729.

### DIFF
--- a/tensorflow/python/ops/nn_ops.py
+++ b/tensorflow/python/ops/nn_ops.py
@@ -23,6 +23,7 @@ import numbers
 import numpy as np
 
 from tensorflow.python.eager import context
+from tensorflow.python.framework import constant_op
 from tensorflow.python.framework import dtypes
 from tensorflow.python.framework import graph_util
 from tensorflow.python.framework import ops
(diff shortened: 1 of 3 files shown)


sb2nov:
Jenkins, test this please.

sb2nov:
@yongtang can you look at the test failure

yongtang:
Thanks @sb2nov. The failure was caused by `xrange` in python 3. The failure has been fixed and the PR has been updated now. Please take a look and let me know if there are any other issues.

sb2nov:
Jenkins, test this please.

yongtang:
From the Jenkins output the Python 3 build has been fixed. The “Linux CPU Test Makefile” might be caused by the network issue. I think a rerun will fix it.

alextp:
Jenkins, test this please.

tensorflow/python/ops/nn_ops.py
```diff
@@ -2043,6 +2044,98 @@ def conv1d(value, filters, stride, padding,
...
+
+  This operation is sometimes called "deconvolution" after [Deconvolutional
+  Networks](http://www.matthewzeiler.com/pubs/cvpr2010/cvpr2010.pdf), but is
+  actually the transpose (gradient) of `conv1d` rather than an actual
+  deconvolution.
+
+  Args:
+    value: A 3-D `Tensor` of type `float` and shape
+      `[batch, in_width, in_channels]` for `NHWC` data format or
+      `[batch, in_channels, in_width]` for `NCHW` data format.
```
alextp: I'm confused; if it's conv1d, why is there h and w?

alextp (changes requested):

yongtang:
Thanks @alextp for the review and sorry for the late reply as I missed the GitHub previously notification. Indeed `NHWC` and `NCHW` is not appropriate for 1d situations. I have rebased and updated the PR so that `NWC` and `NCW` are used instead.

alextp (changes requested):
First, why the diff in fast_tensor_util.cpp? Is that a merger artifact?

Second, let's add an import of this symbol to tf.contrib.nn ; see how it's done in nth_element: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/nn/__init__.py#L35

This way users can access this function through the contrib module for at least one release before we graduate it to tf.nn.

yongtang:
Thanks @alextp for the help.

Not sure why `fast_tensor_util.cpp` was included in the PR as the filename is actually in `.gitignore`. I have removed the file from PR manually.

The conv1d_transpose has also been imported in `tf.contrib.nn` and exposed. Please take a look.

alextp (approved):

alextp:
Jenkins, test this please.

martinwicke:
Jenkins, test this please.

martinwicke:
Failures are unrelated.

Issue #12018, opened 2017-08-03
Missing tensorflow/core/debug/debug_service.grpc.pb.h header file


rlegithub:
Hi,

debug_io_utils.h includes a header file tensorflow/core/debug/debug_service.grpc.pb.h, which is missing from the repo. This causes build on Linux failed. 

`// TODO(cais): Support grpc:// debug URLs in open source once Python grpc
//   genrule becomes available. See b/23796275.

#ifndef PLATFORM_WINDOWS
#include "tensorflow/core/debug/debug_service.grpc.pb.h"
`
Anyone see this issue?

Thanks,
RLE

rlegithub:
@caisq I think this change may relates to the problem: https://github.com/tensorflow/tensorflow/commit/41803db36d4f4a3239bd81e5d460eb0e6e2eea88

That changes #if defined(PLATFORM_GOOGLE) to #ifndef PLATFORM_WINDOWS, which essentially allows it to #include "tensorflow/core/debug/debug_service.grpc.pb.h" on Linux but that file does not exist on the master branch.

Could you help taking a look?

caisq:
`debug_service_grpc.pb.h` is an automatically generated file from this bazel BUILD rule:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/debug/BUILD#L45

So there should be no problem building TF is you are using bazel. What build tool are you using?

rlegithub:
I used Linux Continuous Integration Build steps from this link: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/cmake/README.md

caisq:
So I see that you are running cmake build on Linux. This is our nightly build:
http://ci.tensorflow.org/view/Nightly/job/nightly-cmake-cpu/

It seems to be okay. You can look at the log to see what commands / configuration it uses:
http://ci.tensorflow.org/view/Nightly/job/nightly-cmake-cpu/368/consoleFull

gmaher:
On Ubuntu 16 with anaconda 2.7, when running 

cmake -DCMAKE_BUILD_TYPE=Release ../tensorflow/contrib/cmake
make -j6 all

I get the error: fatal error: tensorflow/core/debug/debug_service.grpc.pb.h: No such file or directory

Any ideas how to resolve this?

yongtang:
Encountered the similar issue. The 
```
tensorflow/core/debug/debug_service.pb.h
```
was generated but not
```
[...]

PR #13128, opened 2017-09-18
Replace `WIN32` -> `PLATFORM_WINDOWS`

yongtang:
I was trying to make cmake build on Linux (#12018/#13061) but noticed that `WIN32` was used in `tensorflow/core/lib/jpeg/jpeg_mem.cc`.

I don't know enough history with TensorFlow's code base though I assume `WIN32` should only be inside `tensorflow/core/platform`?


yongtang:
With #13145 merged this PR is not needed. Closing.

PR #13151, opened 2017-09-19
Improve input tensor structure validation algorithm

anight:
Existing code that does input tensor validation is unable to check slices sizes properly.
Consider this input:
`[][][]float32{{{1, 2}, {3, 4}}, {{1}, {3}}, {{1, 2, 3}, {2, 3, 4}}}`
NewTensor() currently treat this input as valid one.
This patch fixes this behaviour and also simplifies the check by removing unnecessary for cycles.

### DIFF
--- a/tensorflow/go/tensor.go
+++ b/tensorflow/go/tensor.go
@@ -92,7 +92,7 @@ func NewTensor(value interface{}) (*Tensor, error) {
 	raw := tensorData(t.c)
 	buf := bytes.NewBuffer(raw[:0:len(raw)])
 	if dataType != String {
-		if err := encodeTensor(buf, val); err != nil {
+		if err := encodeTensor(buf, val, shape); err != nil {
 			return nil, err
 		}
 		if uintptr(buf.Len()) != nbytes {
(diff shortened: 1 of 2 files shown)


tensorflow/go/tensor.go
```diff
@@ -92,7 +92,7 @@ func NewTensor(value interface{}) (*Tensor, error) {
 	raw := tensorData(t.c)
 	buf := bytes.NewBuffer(raw[:0:len(raw)])
 	if dataType != String {
-		if err := encodeTensor(buf, val); err != nil {
+		if err := encodeTensor(buf, val, shape); err != nil {
```
jhseu: This means that we no longer do the check on string types, right?

anight: You are absolutely right, I missed this point.
Here is a fix for strings as well

jhseu (approved):

gunan:
Jenkins, test this please.

sb2nov:
Jenkins, test this please.

sb2nov:
Jenkins, test this please.

PR #13153, opened 2017-09-19
Fix GRUBlockCell parameter naming inconsistency

### LINKED ISSUE #13137: RNN cells parameter naming inconsistency
In most RNN cells the size is spified by the `num_units` parameter.
This is true for LSTMCell, LSTMBlockCell and GRUCell.
For GRUBlockCell the same parameter is called `cell_size`.
This discrepancy could be problematic in some cases like this code I'm using to create a RNN cell depending on a string parameter passed by the user:

```
def get_cell_fn(cell_type):
    if cell_type == 'rnn':
        cell_fn = tf.nn.rnn_cell.BasicRNNCell
    elif cell_type == 'lstm_basic':
        cell_fn = tf.nn.rnn_cell.BasicLSTMCell
    elif cell_type == 'lstm_basic_ln':
        cell_fn = tf.contrib.rnn.LayerNormBasicLSTMCell
    elif cell_type == 'lstm_block':
        cell_fn = tf.contrib.rnn.LSTMBlockCell
    elif cell_type == 'lstm':
        cell_fn = tf.nn.rnn_cell.LSTMCell
    elif cell_type == 'gru':
        cell_fn = tf.nn.rnn_cell.GRUCell
    elif cell_type == 'gru_block':
        # Faster version of GRU (25% faster in my tests)
        cell_fn = tf.contrib.rnn.GRUBlockCell
    else:
        cell_fn = tf.nn.rnn_cell.BasicRNNCell
    return cell_fn

cell = get_cell_fn(cell_type)(num_units=256)
```

As this will return an error in case of GRUBlockCell.

The solution is just renaming that parameter and possibly add **kwargs to all cell constructors,as right now only RNNCell has **kwargs.

yongtang:
This fix tries to fix the issue in #13137 where parameter `cell_size` is used instead of `num_units`. This is inconsistent with other RNN cells.

This fix adds support of `num_units` while at the same time maintains backward compatiblility for `cell_size`.

This fix fixes #13137.

### DIFF
--- a/tensorflow/contrib/rnn/python/ops/gru_ops.py
+++ b/tensorflow/contrib/rnn/python/ops/gru_ops.py
@@ -136,14 +136,17 @@ class GRUBlockCell(rnn_cell_impl.RNNCell):
     """Initialize the Block GRU cell.
 
     Args:
-      num_units: int, The number of units in the GRU cell
+      num_units: int, The number of units in the GRU cell.
+      cell_size: int, The old (deprecated) name for `num_units`.
+
+    Raises:
+      ValueError: if both cell_size and num_units are not None;
+        or both are None.
     """
-    if cell_size is not None:
-      if num_units is not None:
-        raise ValueError("Cannot specify both 'num_units' and 'cell_size'")
+    if (cell_size is None) == (num_units is None):
+      raise ValueError("Exactly one of num_units or cell_size must be provided.")
+    if num_units is None:
       num_units = cell_size
-    if num_units is None and cell_size is None:
-      raise ValueError("`num_units` and `cell_size` cannot both be None.")
     self._cell_size = num_units
 
   @property
--- a/tensorflow/contrib/rnn/python/ops/gru_ops.py
+++ b/tensorflow/contrib/rnn/python/ops/gru_ops.py
@@ -27,6 +27,7 @@ from tensorflow.python.ops import nn_ops
 from tensorflow.python.ops import rnn_cell_impl
 from tensorflow.python.ops import variable_scope as vs
 from tensorflow.python.platform import resource_loader
+from tensorflow.python.util.deprecation import deprecated_args
 
 _gru_ops_so = loader.load_op_library(
     resource_loader.get_path_to_datafile("_gru_ops.so"))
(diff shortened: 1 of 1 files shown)


drpngx:
@ebrevdo WDYT?

tensorflow/contrib/rnn/python/ops/gru_ops.py
```diff
@@ -129,13 +130,21 @@ class GRUBlockCell(rnn_cell_impl.RNNCell):
...
   """
 
-  def __init__(self, cell_size):
+  @deprecated_args(None, "cell_size is deprecated, use num_units instead",
+                   "cell_size")
+  def __init__(self, num_units=None, cell_size=None):
     """Initialize the Block GRU cell.
 
     Args:
-      cell_size: int, GRU cell size.
```
ebrevdo: must keep the old arg in the order it shows up in the signature; just say it's deprecated and means the same thing as num_units.

add a Raises: section with entry "ValueError: if both cell_size and num_units are not None; or both are None."

tensorflow/contrib/rnn/python/ops/gru_ops.py
```diff
@@ -129,13 +130,21 @@ class GRUBlockCell(rnn_cell_impl.RNNCell):
...
+      num_units: int, The number of units in the GRU cell
     """
-    self._cell_size = cell_size
+    if cell_size is not None:
```
ebrevdo: if (cell_size is None) == (num_units is None):
  raise ValueError("Exactly one of num_units or cell_size must be provided.")

tensorflow/contrib/rnn/python/ops/gru_ops.py
yongtang: Thanks @ebrevdo. The docstring has been updated.

tensorflow/contrib/rnn/python/ops/gru_ops.py
yongtang: @ebrevdo Thanks. Indeed that is much cleaner. Updated.

yongtang:
@ebrevdo Thanks for the review. The PR has been updated. Please take a look.

ebrevdo (approved):

drpngx:
Jenkins, test this please.

drpngx:
Timeout or reboot of the executor, probably.

Jenkins, test this please.

PR #13159, opened 2017-09-19
Refactoring: Pull out repeated shape-related code into shared function 

frreiss:
While looking for some example code to initialize an op's output shape from a shape provided in an attribute, I found a six-line snippet that is repeated in seven different places. It looks like some copying and pasting has happened in the past. This pull request pulls that shared code into a single function in `common_shape_fns.cc`.


caisq:

caisq (approved):

Issue #13129, opened 2017-09-18
random crashes while serving multiple frozen models in parallel using go api


anight:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:

Yes, I have written custom code

- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:

Linux x86_64 SLES12

- **TensorFlow installed from (source or binary)**:

Source, latest master at the moment

- **TensorFlow version (use command below)**:

('v1.3.0-rc1-2265-g6e7539b', '1.4.0-dev')
I also tried r1.3 with similar result.

- **Python version**: 

Python 2.7.9

- **Bazel version (if compiling from source)**:

Build label: 0.5.4
Build target: bazel-out/local-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Aug 25 10:00:00 2017 (1503655200)
Build timestamp: 1503655200
Build timestamp as int: 1503655200

- **CUDA/cuDNN version**:

Not used.

- **GPU model and memory**:

Not used.

- **Exact command to reproduce**:

$ ./tfcrash --n_models 16 --n_images 100

An output of this program can be different. The bug has random nature. In some cases the process just segfaults. Typical output is:

$ ./tfcrash --n_models 16 --n_images 100
2017/09/18 16:45:43 setting 8 cpu
2017/09/18 16:45:43 launching 16 models
2017/09/18 16:45:43 feeding 100 images
2017/09/18 16:45:43 waiting
2017/09/18 16:45:51 session.Run() failed: Expects arg[0] to be uint8 but INVALID is provided

Or:

$ ./tfcrash 
2017/09/18 16:57:54 setting 8 cpu
2017/09/18 16:57:54 launching 16 models
2017/09/18 16:57:54 feeding 100 images
2017/09/18 16:57:54 waiting
Segmentation fault (core dumped)

### Describe the problem

I'm trying to serve predictions from multiple frozen models that I have trained and generated previously using python script. My programming language for serving predictions is golang. I have found that sometimes my process crashes randomly. Exact conditions needed to reproduce this behaviour are unknown. It is also unknown if this bug related to golang bindings or tensorflow itself.
[...]

PR #13184, opened 2017-09-20
A fix for https://github.com/tensorflow/tensorflow/issues/13129

anight:
The problem was with usage of runtime.SetFinalizer() in golang bindings. When a variable become unreferenced GC can harvest it any moment after. According to this https://tip.golang.org/pkg/runtime/#SetFinalizer:

> For example, if p points to a struct that contains a file descriptor d, and p has a finalizer that closes that file descriptor, and if the last use of p in a function is a call to syscall.Write(p.d, buf, size), then p may be unreachable as soon as the program enters syscall.Write. The finalizer may run at that moment, closing p.d, causing syscall.Write to fail because it is writing to a closed file descriptor (or, worse, to an entirely different file descriptor opened by a different goroutine). To avoid this problem, call runtime.KeepAlive(p) after the call to syscall.Write.

In our case unreferenced variable was "feeds", i.e. input tensors. After investigations I found out that func (t *Tensor) finalize() was called a way before SessionRun() was finished and that was the cause of crash.

### DIFF
--- a/tensorflow/go/session.go
+++ b/tensorflow/go/session.go
@@ -89,6 +89,10 @@ func (s *Session) Run(feeds map[Output]*Tensor, fetches []Output, targets []*Ope
 		ptrOutput(c.fetches), ptrTensor(c.fetchTensors), C.int(len(fetches)),
 		ptrOperation(c.targets), C.int(len(targets)),
 		nil, status.c)
+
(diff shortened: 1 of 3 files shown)


anight:
Please ignore this one

PR #13185, opened 2017-09-20
Fix random crashes in SessionRun()

### LINKED ISSUE #13129: random crashes while serving multiple frozen models in parallel using go api

anight:
This patch fixes https://github.com/tensorflow/tensorflow/issues/13129

The problem was with usage of runtime.SetFinalizer() in golang bindings. When a variable become unreferenced GC can harvest it any moment after. According to this https://tip.golang.org/pkg/runtime/#SetFinalizer:

> For example, if p points to a struct that contains a file descriptor d, and p has a finalizer that closes that file descriptor, and if the last use of p in a function is a call to syscall.Write(p.d, buf, size), then p may be unreachable as soon as the program enters syscall.Write. The finalizer may run at that moment, closing p.d, causing syscall.Write to fail because it is writing to a closed file descriptor (or, worse, to an entirely different file descriptor opened by a different goroutine). To avoid this problem, call runtime.KeepAlive(p) after the call to syscall.Write.

In our case unreferenced variable was "feeds", i.e. input tensors. After investigations I have found out that func (t *Tensor) finalize() { C.TF_DeleteTensor(t.c) } was called a way before SessionRun() was finished and that was the cause of crash in my case.

### DIFF
--- a/tensorflow/go/session.go
+++ b/tensorflow/go/session.go
@@ -89,6 +89,10 @@ func (s *Session) Run(feeds map[Output]*Tensor, fetches []Output, targets []*Ope
 		ptrOutput(c.fetches), ptrTensor(c.fetchTensors), C.int(len(fetches)),
 		ptrOperation(c.targets), C.int(len(targets)),
 		nil, status.c)
+
+	// Make sure GC won't harvest input tensors until SessionRun() is finished
+	runtime.KeepAlive(feeds)
+
 	if err := status.Err(); err != nil {
 		return nil, err
 	}


tensorflow/go/session.go
```diff
@@ -89,6 +89,10 @@ func (s *Session) Run(feeds map[Output]*Tensor, fetches []Output, targets []*Ope
 		ptrOutput(c.fetches), ptrTensor(c.fetchTensors), C.int(len(fetches)),
 		ptrOperation(c.targets), C.int(len(targets)),
 		nil, status.c)
+
+	// Make sure GC won't harvest input tensors until SessionRun() is finished
+	runtime.KeepAlive(feeds)
```
asimshankar: Going by the documentation on `SetFinalizer` you pointed out, perhaps this should be:

```
runtime.KeepAlive(c)
```

Probably achieves the same effect, but seems more future proof if `c` is enhanced in the future?

asimshankar: Translating the documentation on `SetFinalizer` you linked to, shouldn't this be `runtime.KeepAlive(c)` instead?

anight: This does not seem right because "c" has no references to "feeds" either.
I have just checked with runtime.KeepAlive(c) and it keeps crashing, as expected.

asimshankar: Or is it that individual fields of `c` are being independently GCed?

asimshankar: Ah, sorry, yeah, you're right - I was not needing the `feeds` vs. `c.feeds` correctly.
Thanks very much!

asimshankar (commented):
Thanks for the fix!

asimshankar (approved):

asimshankar:
Jenkins, test this please

tensorflow/go/session.go
anight: That is because "c.feeds" does not reference "feeds". Instead it references t.c, which is different object for GC unrelated to "feeds" in general.

asimshankar: It may make sense to also call `runtime.KeepAlive(targets)` and `runtime.KeepAlive(fetches)`?

asimshankar: Would it also make sense to call `runtime.KeepAlive(fetches)` and `runtime.KeepAlive(targets)`?

anight: For Operation object there is no finalizer destructor defined, so that means op.c could only point to some constant node of a graph (have not checked this though). In that case Operation object can be harvested by GC independently from it's ".c" part and without any side effects.

asimshankar (approved):

asimshankar:
Jenkins, test this please

caisq:
The test failures are unrelated. Merging the PR.

PR #13193, opened 2017-09-20
Allow partial shape inference for `tf.nn.conv2d_transpose`

### LINKED ISSUE #8972: conv2d_transpose output shape more undefined than input shape
I have posted [this on StackOverflow](http://stackoverflow.com/questions/43113984/output-shape-of-tf-nn-conv2d-transpose-is-entirely-undefined-even-though-only-ba) already but haven't gotten an answer yet, plus it feels like a bug very similar to https://github.com/tensorflow/tensorflow/issues/5807 which is why I'm posting it here as well.

Basically the problem is that the output shape of `tf.nn.conv2d_transpose` is entirely undefined, even if e.g. only one dimension in the input is unknown. So in the following code snippet, the shape of `out` is `[3, 10, 5, 5]` as expected if using the static shape to get the size of the first dimension. However if you use the dynamic shape (commented line in the snippet below), then the shape of `out` is `[?, ?, ?, ?]` instead of `[?, 10, 5, 5]`.

This is a problem for me because I am using `out` in a batch-normalization layer with `tf.contrib.layers.python.layers.batch_norm` for which certain dimensions must be defined.

```
import tensorflow as tf

input_ = tf.Variable(tf.random_normal([3, 10, 5, 1]))
w = tf.get_variable('w', initializer=tf.truncated_normal([3, 3, 5, 1], mean=0.0, stddev=0.01, dtype=tf.float32))

# output_shape = [tf.shape(input_)[0], 10, 5, 5]
output_shape = [input_.get_shape()[0].value, 10, 5, 5]
out = tf.nn.conv2d_transpose(input_,
                             filter=w,
                             output_shape=tf.pack(output_shape),
                             strides=[1, 1, 1, 1],
                             padding='SAME')
```

I am using TF v0.12 on Ubuntu 14.04, Python 3.5.2, installed in Anaconda environment through pip.

yongtang:
This fix tries to address the issue raised in #8972 where it was not possible to infer a partial shape for
`tf.nn.conv2d_transpose`. In case any dimensions are not defined, the shape will be totally undefined.

This fix utilizes `tensor_util.constant_value_as_shape` so that partial shape is possible.

This fix fixes #8972.

### DIFF
--- a/tensorflow/python/kernel_tests/conv2d_transpose_test.py
+++ b/tensorflow/python/kernel_tests/conv2d_transpose_test.py
@@ -24,8 +24,12 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 from tensorflow.python.client import device_lib
 from tensorflow.python.framework import constant_op
 from tensorflow.python.framework import dtypes
+from tensorflow.python.ops import array_ops
 from tensorflow.python.ops import gradient_checker
 from tensorflow.python.ops import nn_ops
+from tensorflow.python.ops import random_ops
+from tensorflow.python.ops import variable_scope
+from tensorflow.python.ops import variables
 import tensorflow.python.ops.nn_grad  # pylint: disable=unused-import
 from tensorflow.python.platform import test
 
@@ -289,6 +293,17 @@ class Conv2DTransposeTest(test.TestCase):
 
         self.assertAllClose(cache_values, value)
 
+
+  def testConv2DTransposeShapeInference(self):
+    # Test case for 8972
+    initializer=random_ops.truncated_normal(
+        [3, 3, 5, 1], mean=0.0, stddev=0.01, dtype=dtypes.float32)
+    x = variables.Variable(random_ops.random_normal([3, 10, 5, 1]))
+    f = variable_scope.get_variable('f', initializer=initializer)
+    f_shape = array_ops.stack([array_ops.shape(x)[0], 10, 5, 5])
+    output = nn_ops.conv2d_transpose(
+      x, f, f_shape, strides=[1, 1, 1, 1], padding='SAME')
+    self.assertEqual(output.get_shape().as_list(), [None, 10, 5, 5])
 
 if __name__ == "__main__":
   test.main()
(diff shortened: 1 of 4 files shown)


caisq:

tensorflow/python/kernel_tests/conv2d_transpose_test.py
```diff
@@ -24,8 +24,12 @@
 from tensorflow.python.client import device_lib
```
cwhipkey: The C++ shape fn can do this correctly, but has a bug - it is explicitly checking input_tensor instead of just calling MakeShapeFromShapeTensor  (MakeShapeFromShapeTensor used to not be able to handle this case correctly, but now does).

Can you instead change core/ops/nn_ops.cc -> the InputTensorShapeOrUnknown function doesn't need to call input_tensor and check for nullptr, it can just call MakeShapeFromShapeTensor.

yongtang: Thanks @cwhipkey! The `MakeShapeFromShapeTensor` works great. I didn't notice it before. The PR has been updated.

yongtang:
Thanks @cwhipkey for the help. The PR has been updated. Please take a look.

tensorflow/python/kernel_tests/conv2d_transpose_test.py
cwhipkey: I think all the callers to InputTensorShapeOrUnknown would benefit from this change - would you mind changing that function instead, to not do the nullptr check and to have the WithRank addition?

yongtang: Thanks @cwhipkey, done.

yongtang:
Thanks @cwhipkey for the review. The PR has been updated. Please take a look.

cwhipkey (approved):
Can InputTensorShapeOrUnknown be removed now (or, make the changed call sites go back to calling that function, and make it do those 5 lines that are in each shape fn now)?

yongtang:
Thanks @cwhipkey. The `InputTensorShapeOrUnknown` and related unit tests have been removed.

sb2nov:
Jenkins, test this please.

Issue #13007, opened 2017-09-13
Feature: tf.decode_csv support NA values


facaiy:
I found many people prefer to using `NA`, `null` or `NULL` or other string as missing value in HIVE table. But `tf.decode_csv` cannot handle the case. So, I suggest to add `na_value` argument like [`pd.read_csv`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html).

PR #13205, opened 2017-09-21
Revert "Revert "EHN: csv supports missing value (#13008)" (#13051)"

facaiy:
issue: #13007
CF PR: #13008

This reverts commit b6f253429c475d1a6f80702feadfff8ff9409156.

### DIFF
--- a/tensorflow/core/kernels/decode_csv_op.cc
+++ b/tensorflow/core/kernels/decode_csv_op.cc
@@ -36,8 +36,8 @@ class DecodeCSVOp : public OpKernel {
     OP_REQUIRES_OK(ctx, ctx->GetAttr("use_quote_delim", &use_quote_delim_));
     OP_REQUIRES(ctx, delim.size() == 1,
                 errors::InvalidArgument("field_delim should be only 1 char"));
-
     delim_ = delim[0];
+    OP_REQUIRES_OK(ctx, ctx->GetAttr("na_value", &na_value_));
   }
 
   void Compute(OpKernelContext* ctx) override {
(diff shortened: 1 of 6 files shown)


facaiy:
cc @gunan @martinwicke @drpngx who I believe might be interested in this PR.

martinwicke:
The only issue I can see here is that na_value comes before name. I know that's where it belongs, but that's a breaking change for people who are not using named args (as they obviously should).

So, in the BUILD file, mark decode_csv "hidden", and then write a python wrapper which swaps the last two arguments. 

@josh11b @asimshankar This is an interesting case -- the fault here lies with the generated python wrapper. I suppose that means we need another wrapper around the wrapper for every argument change. It is all very ugly. @annarev this will all be so much better once the API generation is in!

facaiy:
Thank you for review, @martinwicke . Writing a python wrapper is quite easy for me, however I'm afraid that it brings unnecessary complexity. Hope we could find a more lightweight solution later.

martinwicke:
I agree, this is rather ugly. @annarev is working on a better way to generate our python API from the op definitions -- the new process could fix this. 

Possibly we need to give users more control how Op attrs are turned into python args, maybe by allowing .attr and .attr_after? It's all a little sad, but that's the cost of backwards compatibility.

facaiy:
Python wrapper has been added, and leave a todo comment as memos.

martinwicke:
Jenkins, test this please.

martinwicke (changes requested):
This change does not include any changes to the API. I assume that the api compatibility test will fail. Please run it locally and update the golden files to reflect the added argument.

facaiy:
Hi, golden file has been updated, and compatibility test passed as well on my local machine.
```
INFO: Elapsed time: 13.524s, Critical Path: 9.72s
//tensorflow/tools/api/tests:api_compatibility_test                      PASSED in 3.6s

Executed 1 out of 1 test: 1 test passes.
```

martinwicke:
Jenkins, test this please.

martinwicke (approved):

Issue #13154, opened 2017-09-19
BeamSearchDecoder should support an AttentionWrapper cell with alignment_history enabled


guillaumekln:
### System information
- **Have I written custom code**: yes
- **OS Platform and Distribution**: Ubuntu 16.04
- **TensorFlow installed from**: binary
- **TensorFlow version**: 1.3.0
- **Python version**: 2.7
- **Exact command to reproduce**: see the code snippet below.

### Describe the problem

Currently, setting `tf.contrib.seq2seq.AttentionWrapper`'s `alignment_history` argument to `True` and using this cell in a `tf.contrib.seq2seq.BeamSearchDecoder` does not work for 2 reasons:

1. In this configuration, the `tf.contrib.seq2seq.AttentionWrapper.state_size` property is invalid as it does not have the same structure as `zero_state` (see the code below). The [decoder state initialization](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py#L193) is failing because of this.
2. `tf.contrib.seq2seq.BeamSearchDecoder` [raises an error](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py#L123) when the state contains a `TensorArray`, which is the type currently used to gather alignments.

I believe this configuration should be supported as it is a standard use case for sequence to sequence models.

To address both of these limitations, it seems this `alignment_history` could be a `Tensor` on which alignments are repeatedly concatenated. Would it work?

### Source code / logs

This code sample reproduces 1. which is the error directly visible when using this configuration.

```python
import tensorflow as tf

batch_size = 2
num_units = 10

memory = tf.placeholder(tf.float32, shape=(None, None, num_units))

attention_mechanism = tf.contrib.seq2seq.LuongAttention(
  num_units,
  memory)

cell = tf.contrib.seq2seq.AttentionWrapper(
  tf.contrib.rnn.LSTMCell(num_units),
  attention_mechanism,
  alignment_history=True) # Set this to False to make it work.

tf.contrib.framework.nest.assert_same_structure(
  cell.zero_state(batch_size, dtype=tf.float32),
  cell.state_size)
```

It exits with this error:

```text
Traceback (most recent call last):
  File "<file>", line 19, in <module>
    cell.state_size)
  File "<dir>/local/lib/python2.7/site-packages/tensorflow/python/util/nest.py", line 199, in assert_same_structure
[...]

PR #13208, opened 2017-09-21
Use Tensors instead of TensorArrays for storing AttentionWrapper's alignment_history

guillaumekln:
I would like to propose this fix for #13154.

This is a breaking change as it replaces `TensorArray`s of the `alignment_history` field with batch major `Tensor`s.

Additionally, the helper function used to gather beams in `BeamSearchDecoder` had to be updated to allow keeping the original values dimensions.


ebrevdo:
This looks like it may kill performance.  I'm closing for now but let's continue discussing other solutions on the original bug report.

PR #13217, opened 2017-09-21
Fixes #12401: Use traceback instead of inspect for frame information

### LINKED ISSUE #12401: `import tensorflow.contrib.layers` takes very long time
I've used Cprofile to test my code, and I find that `import tensorflow.contrib.layers` takes very long time. Here is the quick screenshot of the result:

```
14495011 function calls (14451561 primitive calls) in 6.324 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   1023/1    0.058    0.000    6.328    6.328 {built-in method builtins.exec}
        1    0.000    0.000    6.328    6.328 test_import.py:10(<module>)
   1066/1    0.006    0.000    6.328    6.328 <frozen importlib._bootstrap>:966(_find_and_load)
   1066/1    0.005    0.000    6.328    6.328 <frozen importlib._bootstrap>:939(_find_and_load_unlocked)
   1360/1    0.001    0.000    6.328    6.328 <frozen importlib._bootstrap>:214(_call_with_frames_removed)
    759/1    0.002    0.000    6.328    6.328 {built-in method builtins.__import__}
    881/2    0.004    0.000    6.328    3.164 <frozen importlib._bootstrap>:659(_load_unlocked)
    833/2    0.003    0.000    6.328    3.164 <frozen importlib._bootstrap_external>:667(exec_module)
 6575/625    0.005    0.000    6.102    0.010 <frozen importlib._bootstrap>:996(_handle_fromlist)
... 
```
As seen above, it took more than 6 secs to just do `import tensorflow.contrib.layers`, and I wonder if this is expected. Also, I am pretty interested in why it takes so long time to do importing there.

**P.S** The profiling test runs on MacOS Sierra 10.12.5 with Tensorflow v1.2 and Python 3.5.

JettJones:
#12401 - import tensorflow.contrib.layers takes a very long time

I learned (while rebasing this diff) that the issue of slow imports of tensorflow.contrib.layers was fixed earlier this week: https://github.com/tensorflow/tensorflow/pull/11830. 🎉 Since the diff is ready,  I want to put up the diff for the sake of history, but I understand if at this point the further speedup is minor enough to ignore.

The speedup is a factor of 2, while the original fix is a factor of 100.   

----

My performance tests compared three versions:
- original implementation with inspect.stack()
- traceback(limit)
- current_frame + inspect

I also found that the original implementation was sensitive to the depth of the current stack, and a file importing another file generally increased that call depth by 5.  Which is to say, in practice, my code importing `tensorflow.config` at an import depth of 7 had much worse performance than the baseline test of just importing `tensorflow.config`

| Test  | Timing |
| -- | -- |
| python2 - inspect.stack |  7.8 sec |
| python2 - inspect.getframeinfo  |  0.12 sec   |
| python2 - traceback.extract_stack  |  0.04 sec   |
| python3 - inspect.stack |  22.8 sec |
| python3 - inspect.getframeinfo  |  0.14 sec   |
| python3 - traceback.extract_stack  |  0.07 sec   |

^ tests are the total of 1000 runs with a stack of 200, to give easy to compare numbers.
https://gist.github.com/JettJones/c236494013f22723c1822126df944b12

This functionality of adding a decorator name from the call stack is covered by the existing unit test:
* testSetsDecoratorNameToFunctionThatCallsMakeDecoratorIfAbsent

And pylint did not report any issues on these changes

### DIFF
--- a/tensorflow/contrib/framework/python/ops/arg_scope.py
+++ b/tensorflow/contrib/framework/python/ops/arg_scope.py
@@ -181,7 +181,7 @@ def add_arg_scope(func):
     return func(*args, **current_args)
   _add_op(func)
   setattr(func_with_args, '_key_op', _key_op(func))
-  return tf_decorator.make_decorator(func, func_with_args)
+  return tf_decorator.make_decorator(func, func_with_args, decorator_name='add_arg_scope')
 
 
 def has_arg_scope(func):
--- a/tensorflow/contrib/framework/python/ops/arg_scope.py
+++ b/tensorflow/contrib/framework/python/ops/arg_scope.py
@@ -181,7 +181,7 @@ def add_arg_scope(func):
     return func(*args, **current_args)
   _add_op(func)
   setattr(func_with_args, '_key_op', _key_op(func))
-  return tf_decorator.make_decorator(func, func_with_args, decorator_name='add_arg_scope')
+  return tf_decorator.make_decorator(func, func_with_args)
 
 
 def has_arg_scope(func):
--- a/tensorflow/python/util/tf_decorator.py
+++ b/tensorflow/python/util/tf_decorator.py
@@ -60,7 +60,7 @@ from __future__ import division
 from __future__ import print_function
 
 import functools as _functools
-import inspect as _inspect
+import traceback as _traceback
 
 
 def make_decorator(target,
@@ -83,7 +83,9 @@ def make_decorator(target,
     The `decorator_func` argument with new metadata attached.
   """
   if decorator_name is None:
-    decorator_name = _inspect.stack()[1][3]  # Caller's name.
+    frame = _traceback.extract_stack(limit=2)[0]
+    # frame name is tuple[2] in python2, and object.name in python3
+    decorator_name = getattr(frame, 'name', name[2]) # Caller's name
   decorator = TFDecorator(decorator_name, target, decorator_doc,
                           decorator_argspec)
   setattr(decorator_func, '_tf_decorator', decorator)
--- a/tensorflow/python/util/tf_decorator.py
+++ b/tensorflow/python/util/tf_decorator.py
@@ -85,7 +85,7 @@ def make_decorator(target,
   if decorator_name is None:
     frame = _traceback.extract_stack(limit=2)[0]
     # frame name is tuple[2] in python2, and object.name in python3
-    decorator_name = getattr(frame, 'name', name[2]) # Caller's name
+    decorator_name = getattr(frame, 'name', frame[2]) # Caller's name
   decorator = TFDecorator(decorator_name, target, decorator_doc,
                           decorator_argspec)
   setattr(decorator_func, '_tf_decorator', decorator)
--- a/tensorflow/python/util/tf_decorator.py
+++ b/tensorflow/python/util/tf_decorator.py
@@ -60,7 +60,7 @@ from __future__ import division
 from __future__ import print_function
 
 import functools as _functools
-import inspect as _inspect
+import traceback as _traceback
 
 
 def make_decorator(target,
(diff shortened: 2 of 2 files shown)


JettJones:
`I've signed it!`  - My company has signed the CLA, and my email address used on the commit is a member of the google group that we specified.  This will be my first time going through the process, so please let me know if some part of that is incorrect.  Thanks!

drpngx:
@charlesnicholson mind taking a look?

drpngx:
Jenkins, test this please.

drpngx:
I don't see your username in our CLA database, presumably because it doesn't crawl all under the company. @gunan please advise.

gunan:
@JettJones If you are covered under corporate CLA, you need to:
-Add your corporate email to your github account as a secondary email
-Use your corporate email
-Comment this, and only this to retrigger CLA check: "I signed it!"

JettJones:
I signed it!

gunan:
Jenkins, test this please.

yaroslavvb:
the only failure is `nn_grad_test`, sounds like the test is flaky:

`Expected: (max_error) < (2e-4), actual: 0.000205815 vs 0.0002`

gunan:
Correct, nn_grad_test issues are addressed internally, next push should fix the test.

gunan:
@josh11b @drpngx change passing all tests and is ready for review.

drpngx:
I'm good with the changes, perhaps @charlesnicholson should take a look.

tensorflow/contrib/framework/python/ops/arg_scope.py
```diff
@@ -181,7 +181,7 @@ def func_with_args(*args, **kwargs):
     return func(*args, **current_args)
   _add_op(func)
   setattr(func_with_args, '_key_op', _key_op(func))
-  return tf_decorator.make_decorator(func, func_with_args)
+  return tf_decorator.make_decorator(func, func_with_args, decorator_name='add_arg_scope')
```
charlesnicholson: Is this necessary? I.e. if this line is required, it means that `tf_decorator.make_decorator` isn't correctly getting the caller's name (`func_with_args`), which means that the new behavior isn't  working as intended.

charlesnicholson:
Pending resolution of the comment I left, this looks good. (I think it's vestigial but if not would indicate behavior regression)

tensorflow/contrib/framework/python/ops/arg_scope.py
JettJones: Passing in the decorator_name could be left out, and you're right - if this is the only call path that triggers the name look-up it could leave the changed behavior untested.  By my reading, the unit tests still cover inspection, and I will confirm other make_decorator calls as well.  

Originally my question was - why run additional call stack inspection when the caller could pass in its name? But after finding alternatives to `inspect.stack()` the overhead of the repeated lookup is no longer visible.

Long story short, I'll drop this change from my diff to ensure my main changes are well exercised.  Also - check into #13232 which is looking to provide this same fix to add_arg_scope.

drpngx: Should we add a test to document that behavior? (in another PR)

charlesnicholson: Not a huge deal to me, I just want to confirm that the changes to line 184 aren't /required/.

charlesnicholson: also @drpngx this functionality is tested, which is why I was curious about the change. Like I said, though, it's not important. I don't think it should hold up the PR.

charlesnicholson: @JettJones the functionality is there to minimize the amount of duplication required to get decorators to work. It's lower friction than having to write `def my_decorator():` and then later `my_decorator`: string duplication is resistant to rename refactoring, etc. I'm not religious about it, it's just a convenience.

JettJones: I didn't find another caller of `make_decorator` under contrib, so I rolled back passing in `decorator_name` explicitly.  Thanks for calling this out.

charlesnicholson:
This looks great, thanks for listening to my nits :)

@drpngx I think this should be merged ASAP given the speedup. Thanks again for taking the time to write + fix.

sb2nov:
Jenkins, test this please.

drpngx:
Not clear if it's a transient failure. Jenkins, test this please.

drpngx (approved):
As per @charlesnicholson comment.

sb2nov:
Jenkins, test this please.

drpngx:
@caisq Ready to merge.

caisq:
@drpngx Will merge when sync'ing to internal is done.

PR #13232, opened 2017-09-22
Getting rid from unnecessary call 'inspect.stack'

habibutsu:
Getting rid from unnecessary call 'inspect.stack' in making of decorator in 'tf.contrib.framework.add_arg_scope'

P.S.
On my laptop this call gives big overhead for initialization of tensorflow

### DIFF
--- a/tensorflow/contrib/framework/python/ops/arg_scope.py
+++ b/tensorflow/contrib/framework/python/ops/arg_scope.py
@@ -181,7 +181,7 @@ def add_arg_scope(func):
     return func(*args, **current_args)
   _add_op(func)
   setattr(func_with_args, '_key_op', _key_op(func))
-  return tf_decorator.make_decorator(func, func_with_args)
+  return tf_decorator.make_decorator(func, func_with_args, 'add_arg_scope')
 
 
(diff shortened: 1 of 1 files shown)


drpngx:
Can you take a look at #13217?

Thanks

habibutsu:
@drpngx  #13217 is Ok, but it still help to avoid unnecessary call of inspection of stack even if will used *traceback* instead *inspect* module

drpngx:
Can you comment on that thread?

habibutsu:
@rpngx what I should comment? That PR already merged and also this improvements though fixes same problem but making it in different places, so why we should work with both of this PR as one?

drpngx:
Can you figure out what the problem is with the CLA?

/CC: @charlesnicholson

charlesnicholson:
This code looks reasonable. It's probably worth taking out the inspect stuff from decorators and just requiring that the author pass the name of the decorator function in as a string. (maybe not in this PR though)

martinwicke:
@habibutsu Can you sign the CLA?

martinwicke:
Closing as stalled.

PR #13239, opened 2017-09-22
Add ReceptiveField class and coordinate conversion methods.

tillahoffmann:
This PR adds a `ReceptiveField` class which supports coordinate conversion from the input to the feature space and vice versa while maintaining backwards compatibility.

### DIFF
--- a/tensorflow/contrib/receptive_field/python/util/receptive_field.py
+++ b/tensorflow/contrib/receptive_field/python/util/receptive_field.py
@@ -27,12 +27,14 @@ import math
 from tensorflow.contrib.receptive_field.python.util import graph_compute_order
 from tensorflow.contrib.util import make_ndarray
 from tensorflow.python.platform import tf_logging as logging
+from tensorflow.python.framework import ops as framework_ops
+import numpy as np
 
 # White-listed layer operations, which do not affect the receptive field
 # computation.
 _UNCHANGED_RF_LAYER_OPS = [
     "Softplus", "Relu", "BiasAdd", "Mul", "Add", "Const", "Identity",
-    "VariableV2", "Sub", "Rsqrt", "ConcatV2"
+    "VariableV2", "Sub", "Rsqrt", "ConcatV2", "Log", "Pow", "RealDiv",
 ]
 
 # Different ways in which padding modes may be spelled.
@@ -304,6 +306,62 @@ def _get_effective_padding_node_input(stride, padding,
   return stride * effective_padding_output + padding
 
 
+class ReceptiveField:
+  """
+  Receptive field of a convolutioanl neural network.
+
+  Args:
+    size: Receptive field size.
+    stride: Effective stride.
+    padding: Effective padding.
+  """
+  def __init__(self, size, stride, padding):
+    self.size = np.asarray(size)
+    self.stride = np.asarray(stride)
+    self.padding = np.asarray(padding)
+
+  def compute_input_coordinates(self, x, axis=None):
+    """
+    Computes the center of the receptive field that generated a feature.
+
+    Args:
+      x: Coordinate of the feature.
+      axis: The dimensions for which to compute the input coordinates.
+
+    Returns:
+      y: Center of the receptive field that generated the feature.
+    """
+    # Use all dimensions
+    if axis is None:
+      axis = range(self.size.size)
+    # Ensure axis is a list because tuples have different indexing behavior
+    axis = list(axis)
+    return - self.padding[axis] + x * self.stride[axis] + \
+      (self.size[axis] - 1) / 2
+
+  def compute_feature_coordinates(self, y, axis=None):
+    """
+    Computes the position of a feature given the center of a receptive field.
+
+    Args:
+      y: Center of the receptive field.
+      axis: The dimensions for which to compute the feature coordinates.
+
+    Returns:
+      x: Coordinate of the feature.
+    """
+    # Use all dimensions
+    if axis is None:
+      axis = range(self.size.size)
+    # Ensure axis is a list because tuples have different indexing behavior
+    axis = list(axis)
+    return (y + self.padding[axis] + (1 - self.size[axis]) / 2) / \
+      self.stride[axis]
+
+  def __iter__(self):
+    return iter(np.concatenate([self.size, self.stride, self.padding]))
+
+
 def compute_receptive_field_from_graph_def(graph_def, input_node, output_node):
   """Computes receptive field (RF) parameters from a GraphDef object.
 
@@ -331,6 +389,16 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node):
       cannot be found. For network criterion alignment, see
       photos/vision/features/delf/g3doc/rf_computation.md
   """
+  # Convert a graph to graph_def if necessary
+  if isinstance(graph_def, framework_ops.Graph):
+    graph_def = graph_def.as_graph_def()
+
+  # Convert tensors to names
+  if isinstance(input_node, framework_ops.Tensor):
+    input_node = input_node.op.name
+  if isinstance(output_node, framework_ops.Tensor):
+    output_node = output_node.op.name
+
   # Computes order of computation for a given graph.
   name_to_order_node = graph_compute_order.get_compute_order(
       graph_def=graph_def)
@@ -480,6 +548,7 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node):
     raise ValueError("Output node was not found")
   if input_node not in rf_sizes_x:
     raise ValueError("Input node was not found")
-  return (rf_sizes_x[input_node], rf_sizes_y[input_node],
-          effective_strides_x[input_node], effective_strides_y[input_node],
-          effective_paddings_x[input_node], effective_paddings_y[input_node])
+  return ReceptiveField(
+    (rf_sizes_x[input_node], rf_sizes_y[input_node]),
+    (effective_strides_x[input_node], effective_strides_y[input_node]),
+    (effective_paddings_x[input_node], effective_paddings_y[input_node]))
--- a/tensorflow/contrib/receptive_field/python/util/receptive_field.py
+++ b/tensorflow/contrib/receptive_field/python/util/receptive_field.py
@@ -339,10 +339,10 @@ class ReceptiveField:
       ValueError: If the number of dimensions of the feature coordinates does
         not match the number of elements in `axis`.
     """
-    # Use all dimensions
+    # Use all dimensions.
     if axis is None:
       axis = range(self.size.size)
-    # Ensure axis is a list because tuples have different indexing behavior
+    # Ensure axis is a list because tuples have different indexing behavior.
     axis = list(axis)
     x = np.asarray(x)
     if x.shape[-1] != len(axis):
@@ -370,10 +370,10 @@ class ReceptiveField:
       ValueError: If the number of dimensions of the input center coordinates
         does not match the number of elements in `axis`.
     """
-    # Use all dimensions
+    # Use all dimensions.
     if axis is None:
       axis = range(self.size.size)
-    # Ensure axis is a list because tuples have different indexing behavior
+    # Ensure axis is a list because tuples have different indexing behavior.
     axis = list(axis)
     y = np.asarray(y)
     if y.shape[-1] != len(axis):
@@ -423,11 +423,11 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node,
       cannot be found. For network criterion alignment, see
       photos/vision/features/delf/g3doc/rf_computation.md
   """
-  # Convert a graph to graph_def if necessary
+  # Convert a graph to graph_def if necessary.
   if isinstance(graph_def, framework_ops.Graph):
     graph_def = graph_def.as_graph_def()
 
-  # Convert tensors to names
+  # Convert tensors to names.
   if isinstance(input_node, framework_ops.Tensor):
     input_node = input_node.op.name
   if isinstance(output_node, framework_ops.Tensor):
@@ -526,7 +526,7 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node,
 
       # Loop over this node's inputs and potentially propagate information down.
       for inp_name in node.input:
-        # Stop the propagation of the receptive field
+        # Stop the propagation of the receptive field.
         if any(inp_name.startswith(stop) for stop in stop_propagation):
           logging.vlog(3, "Skipping explicitly ignored node %s.", node.name)
           continue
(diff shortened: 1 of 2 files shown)


sb2nov:
Jenkins, test this please.

martinwicke:
@marksandler @andrefaraujo can you take a look?

andrefaraujo:
yes, will look into it soon. (next day or so)

andrefaraujo:
apologies on the delay for this, things got busy. Will make sure to review it next week.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -304,13 +308,72 @@ def _get_effective_padding_node_input(stride, padding,
...
+    self.stride = np.asarray(stride)
+    self.padding = np.asarray(padding)
+
+  def compute_input_coordinates(self, x, axis=None):
+    """
+    Computes the center of the receptive field that generated a feature.
+
+    Args:
+      x: Coordinate of the feature.
+      axis: The dimensions for which to compute the input coordinates.
```
andrefaraujo: Please explain/document what the default behavior is (ie, what happens if axis is None). My understanding is that it will return a list of centers for each dimension.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -304,13 +308,72 @@ def _get_effective_padding_node_input(stride, padding,
...
+    return - self.padding[axis] + x * self.stride[axis] + \
+      (self.size[axis] - 1) / 2
+
+  def compute_feature_coordinates(self, y, axis=None):
+    """
+    Computes the position of a feature given the center of a receptive field.
+
+    Args:
+      y: Center of the receptive field.
+      axis: The dimensions for which to compute the feature coordinates.
```
andrefaraujo: same here.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -304,13 +308,72 @@ def _get_effective_padding_node_input(stride, padding,
...
+    size: Receptive field size.
+    stride: Effective stride.
+    padding: Effective padding.
+  """
+  def __init__(self, size, stride, padding):
+    self.size = np.asarray(size)
+    self.stride = np.asarray(stride)
+    self.padding = np.asarray(padding)
+
+  def compute_input_coordinates(self, x, axis=None):
```
andrefaraujo: what about renaming this to "compute_input_center_coordinates"?
I believe the current form is ambiguous, since "coordinates" could correspond to the "bounding box" coordinates of the receptive field region (ie, top-left and bottom-right corners).

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -331,6 +394,18 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node):
       cannot be found. For network criterion alignment, see
       photos/vision/features/delf/g3doc/rf_computation.md
   """
+  # Convert a graph to graph_def if necessary
```
andrefaraujo: It seems like the function now supports graph_def to be a simple graph, and input and output nodes to be tensors. Please add a few comments to the docstring to reflect these changes.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -304,13 +308,72 @@ def _get_effective_padding_node_input(stride, padding,
...
+
+def compute_receptive_field_from_graph_def(graph_def, input_node, output_node,
+                                           stop_propagation=None):
   """Computes receptive field (RF) parameters from a GraphDef object.
 
   Args:
     graph_def: GraphDef object.
     input_node: Name of the input node from graph.
     output_node: Name of the output node from graph.
+    stop_propagation: List of operation  or scope names for which to stop the
```
andrefaraujo: I understand the logic for the need of stop_propagation. I guess it is useful in order to compute receptive field information for specific paths of a network. (in other words, in a CNN there are multiple ways to go from input to output; stop_propagation supports measuring RF information for only part of the network).
Are there some other cases where it is useful? Anyway, if you could just one or a couple more sentence to explain to the not-so-familiar reader when stop_propagation can be useful, it would be perfect.

tensorflow/contrib/receptive_field/python/util/receptive_field_test.py
```diff
@@ -150,6 +151,31 @@ def create_test_network_5():
   return g
 
 
+def create_test_network_6():
+  """Aligned network for test.
+
+  The graph corresponds to the example from the second figure in
+  go/cnn-rf-computation#arbitrary-computation-graphs
```
andrefaraujo: add at the end of the sentence: ", with added dropout."

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -27,12 +27,15 @@
...
+from tensorflow.python.framework import ops as framework_ops
+import numpy as np
 
 # White-listed layer operations, which do not affect the receptive field
 # computation.
 _UNCHANGED_RF_LAYER_OPS = [
     "Softplus", "Relu", "BiasAdd", "Mul", "Add", "Const", "Identity",
-    "VariableV2", "Sub", "Rsqrt", "ConcatV2"
+    "VariableV2", "Sub", "Rsqrt", "ConcatV2", "Log", "Pow", "RealDiv",
+    "Floor", "Ceil", "Round",
```
andrefaraujo: this list is getting big. Could you please sort the items by name?

tensorflow/contrib/receptive_field/python/util/receptive_field_test.py
```diff
@@ -220,6 +246,36 @@ def testComputeRFFromGraphDefNonSquareRF(self):
...
+    graph_def = create_test_network_6().as_graph_def()
+    input_node = 'input_image'
+    output_node = 'output'
+    # Compute the receptive field but stop the propagation for the random
+    # uniform variable of the dropout
+    (receptive_field_x, receptive_field_y, effective_stride_x,
+     effective_stride_y, effective_padding_x, effective_padding_y) = (
+         receptive_field.compute_receptive_field_from_graph_def(
+             graph_def, input_node, output_node,
+             ['Dropout/dropout/random_uniform']))
```
andrefaraujo: In this case, I believe the same behavior could be achieved by adding a new op "RandomUniform" to "_UNCHANGED_RF_LAYER_OPS". Correct?
This is just to check my understanding, I am not suggesting you to do this. This is a great test for the stop_propagation option.

andrefaraujo (changes requested):
Great addition, thanks tillahoffman@!

I have mostly small comments, could you please address them?

tensorflow/contrib/receptive_field/python/util/receptive_field_test.py
tillahoffmann: Almost. The implementation of `dropout` uses the shape of the input features to generate uniform random numbers of matching shape. That introduces a dependency of a `RandomUniform` operation on the input features but there is no meaningful way to propagate the receptive field. So the propagation of the receptive field needs to be stopped rather than passed through unchanged.

tillahoffmann:
@andrefaraujo, thanks for the feedback. Comments should be addressed.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -322,22 +321,34 @@ def __init__(self, size, stride, padding):
...
 
     Returns:
-      y: Center of the receptive field that generated the feature.
+      y: Center of the receptive field that generated the features.
+
+    Raises:
+      ValueError: If the number of dimensions of the feature coordinates does
+        not match the number of elements in `axis`.
     """
     # Use all dimensions
```
andrefaraujo: please use a period at the end of comment sentences.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -322,22 +321,34 @@ def __init__(self, size, stride, padding):
...
     if axis is None:
       axis = range(self.size.size)
     # Ensure axis is a list because tuples have different indexing behavior
```
andrefaraujo: same here.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -322,22 +321,34 @@ def __init__(self, size, stride, padding):
...
-      axis: The dimensions for which to compute the input coordinates.
+      x: An array of feature coordinates with shape `(..., d)`, where `d` is the
+        number of dimensions of the coordinates.
+      axis: The dimensions for which to compute the input center coordinates.
+        If `None` (the default), compute the input center coordinates for all
+        dimensions.
 
     Returns:
-      y: Center of the receptive field that generated the feature.
+      y: Center of the receptive field that generated the features.
```
andrefaraujo: What about: "Center of the receptive field that generated the features at the input of the network."

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -322,22 +321,34 @@ def __init__(self, size, stride, padding):
     self.stride = np.asarray(stride)
     self.padding = np.asarray(padding)
 
-  def compute_input_coordinates(self, x, axis=None):
+  def compute_input_center_coordinates(self, x, axis=None):
```
andrefaraujo: I am a little bothered by the fact that output feature coordinates are "x" and input center coordinates are "y". Ie, "y" at the input and "x" at the output, which is a little weird. I think swapping the two notations would be better for clarity, and it shouldn't be too hard since there not that many cases. Can you please do it?

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -346,17 +357,29 @@ def compute_feature_coordinates(self, y, axis=None):
     Computes the position of a feature given the center of a receptive field.
 
     Args:
-      y: Center of the receptive field.
+      y: An array of input center coordinates with shape `(..., d)`, where `d`
```
andrefaraujo: same in this function for swapping "x" and "y".

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -304,13 +307,102 @@ def _get_effective_padding_node_input(stride, padding,
...
+        dimensions.
+
+    Returns:
+      x: Coordinates of the features.
+
+    Raises:
+      ValueError: If the number of dimensions of the input center coordinates
+        does not match the number of elements in `axis`.
+    """
+    # Use all dimensions
```
andrefaraujo: missing period.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -304,13 +307,102 @@ def _get_effective_padding_node_input(stride, padding,
...
+    if axis is None:
+      axis = range(self.size.size)
+    # Ensure axis is a list because tuples have different indexing behavior
```
andrefaraujo: missing period.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -331,6 +423,18 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node):
       cannot be found. For network criterion alignment, see
       photos/vision/features/delf/g3doc/rf_computation.md
   """
+  # Convert a graph to graph_def if necessary
```
andrefaraujo: missing period.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -331,6 +423,18 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node):
...
+  if isinstance(graph_def, framework_ops.Graph):
+    graph_def = graph_def.as_graph_def()
+
+  # Convert tensors to names
```
andrefaraujo: missing period.

tensorflow/contrib/receptive_field/python/util/receptive_field.py
```diff
@@ -422,6 +526,10 @@ def compute_receptive_field_from_graph_def(graph_def, input_node, output_node):
 
       # Loop over this node's inputs and potentially propagate information down.
       for inp_name in node.input:
+        # Stop the propagation of the receptive field
```
andrefaraujo: missing period.

tensorflow/contrib/receptive_field/python/util/receptive_field_test.py
```diff
@@ -220,6 +246,36 @@ def testComputeRFFromGraphDefNonSquareRF(self):
     self.assertEqual(effective_padding_x, 0)
     self.assertEqual(effective_padding_y, 0)
 
+  def testComputeRFFromGraphDefStopPropagation(self):
+    graph_def = create_test_network_6().as_graph_def()
+    input_node = 'input_image'
+    output_node = 'output'
+    # Compute the receptive field but stop the propagation for the random
+    # uniform variable of the dropout
```
andrefaraujo: missing period.

tensorflow/contrib/receptive_field/python/util/receptive_field_test.py
andrefaraujo: hmmm... if we added the "RandomUniform" op to "_UNCHANGED_RF_LAYER_OPS", it would still work I guess? Maybe I am missing something here.
If I understand correctly, based on what you said and based on my understanding of dropout, the graph with Dropout will have a branch to generate random values which will be used to elementwise multiply the features, deactivating some of them. So if we simply propagated back the RF through this branch without changing it, the computation would still work?

andrefaraujo (changes requested):
Added some more comments and ideas, thanks again tillahoffmann@!

tensorflow/contrib/receptive_field/python/util/receptive_field_test.py
tillahoffmann: Yes, that's right. I confused myself.

andrefaraujo: thanks, glad I understand this now.

andrefaraujo (approved):
Thanks, @tillahoffmann! All LGTM :)

andrefaraujo:
@martinwicke , not sure if I need to do something else now. I approved but the "TF Test Suite" and "Ubuntu CC" checks have been pending for ~12hs now. Let me know if I need to do something.

martinwicke:
Jenkins, test this please.

tillahoffmann:
@martinwicke, do you know why the Google cloud source builds are failing?

martinwicke:
No, but it's definitely unrelated to this change.

PR #13247, opened 2017-09-22
Refactoring of canned estimators

davorrunje:
Class specific `model_fn()` used by each canned estimator are replaced by `common_model_fn()` function and moved to `utils.py`, together with common `classifier_head()` and `regression_head()` functions. Canned estimators are then refactored to use these common functions which significantly simplified their code and increased readability.


davorrunje:
Some (seemingly unrelated) tests are failing, I need to check why first.

PR #13252, opened 2017-09-22
GetConvolve*Algorithms return tensor-op algos

nluehr:
Attention: @jlebar and @zheng-xq

This is the follow-up on the cudnn7 patch. When enumerating convolution algorithms, it moves the tensor_ops toggle into the GetConvolve*Algorithms functions.

Also tensor_ops are no longer included in the returned algo list if they are not supported by the cuDNN version or GPU compute capability.


jlebar (approved):
This is wonderful; thank you for doing this.

sb2nov:
Jenkins, test this please.

PR #13270, opened 2017-09-24
Fix docs for tf.tanh, tf.sigmoid and tf.tan.

codrut3:
The API docs incorrectly state that tf.tanh accepts int32
or int64 inputs. This was referenced in issue #10376, however,
the fix to #10376 only removed the reference to qint32. A look
at the implementation in cwise_op_tanh.cc shows that int32 and
int64 are indeed not supported. This fix removes them from
the API docs.

Similarly, we remove int32, int64 and qint32 from tf.sigmoid
API docs.

Finally, tf.tan is defined only for float and double, so I
changed in math_ops.cc to use UNARY_REAL.

### DIFF
--- a/tensorflow/core/ops/math_ops.cc
+++ b/tensorflow/core/ops/math_ops.cc
@@ -362,7 +362,7 @@ REGISTER_OP("Cos").UNARY_COMPLEX().Doc(R"doc(
 Computes cos of x element-wise.
 )doc");
 
-REGISTER_OP("Tan").UNARY_REAL().Doc(R"doc(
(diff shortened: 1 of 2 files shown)


caisq:

codrut3:
Hi,

It seems changing UNARY() to UNARY_REAL() broke the backwards compatibility test, so I reverted that change. I hope all tests will pass now.

rmlarsen (approved):

rmlarsen:

sb2nov:
Jenkins, test this please.

PR #13277, opened 2017-09-24
Refactoring of canned estimators

davorrunje:
Class specific `model_fn()` used by each canned estimator are replaced by `common_model_fn()` and `common_combined_model_fn()` functions and moved to `utils.py`, together with common `classifier_head()` and `regression_head()` functions. Canned estimators are then refactored to use these common functions which simplified their code and increased readability.

### DIFF
--- a/tensorflow/python/estimator/BUILD
+++ b/tensorflow/python/estimator/BUILD
@@ -35,7 +35,6 @@ py_library(
         ":model_fn",
         ":parsing_utils",
         ":run_config",
-        ":training",
         "//tensorflow/python:util",
     ],
 )
@@ -124,28 +123,6 @@ py_test(
 )
 
 py_library(
-    name = "training",
-    srcs = ["training.py"],
-    srcs_version = "PY2AND3",
-    deps = [
-        ":estimator",
-        "//tensorflow/python:training",
-        "@six_archive//:six",
-    ],
-)
-
-py_test(
-    name = "training_test",
-    size = "small",
-    srcs = ["training_test.py"],
-    srcs_version = "PY2AND3",
-    deps = [
-        ":training",
-        "//tensorflow/python:client_testlib",
-    ],
-)
-
-py_library(
     name = "run_config",
     srcs = ["run_config.py"],
     srcs_version = "PY2AND3",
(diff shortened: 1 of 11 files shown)


roumposg (commented):
Thanks for working on that, but we do not want to do this refactoring. When we wrote the canned estimators, we made some explicit choices:
* Keep model_fn in the same file as the estimator, e.g. dnn_model_fn in the same file as DNNClassifier. There may be some duplicate code among. dnn.py. linear.py and dnn_linear_combined.py, but we are ok with that. The benefit is that one needs to only read one file to understand what DNN is doing.
* We don't want to compose model_fn. We broke it into logit_fn and head, but we don't want to compose one model_fn from multiple model_fn. The benefit, again, is that it is easier to read what the model_fn is doing.
* We want dnn_model_fn and linear_model_fn to evolve independently. E.g. some feature may make sense for one, but not for the other. Keeping them separate was an explicit design decision.

Let me know if that answers your concerns.

davorrunje:
@roumposg Thank you for the explanation of your design decision.

My primary motive for the refactorization is my own work which requires implementing my own Estimators. E.g. I would need a CNN estimator and with this refactorization I could easily implement one by simply implementing its own logit_fn and reusing common head and model_fn. In case I need to extend DNN estimator with batch normalization the same tactic would sufice. Most use cases I can think of use the same model_fn and only differ in logit_fn and head. Reimplementing the same model function each time would make code more difficult to maintain because it would require every change to be propagated through each Estimator.

Of course, I can always use this refactorization for my own estimators regardless of merging it with the master, but I feel it would be useful to others as well.

ispirmustafa:
davorrunje@
if you have head and logit_fn, will your model_fn look like as follows, rigth?
def model_fn(features, labels, mode):
  logits = logit_fn(features)
  return head.create_estimator_spec(...)

davorrunje:
@ispirmustafa yes, the `model_fn()`  is exactly the same as before externally, and internally it is really just combining of the existing ones. First, I noticed that `model_fn()` of DNN and linear estimators is the same and there is no need for code duplication. Then I noticed that combined `model_fn()` is just unrolling of that same function twice and can be written as a loop. This way one function can replace all three written so far.

The reason why I think it is a good idea to have one` model_fn()` is when you start writing your own estimators based on existing ones. I am teaching a class on deep learning using Tensorflow and I teach students different techniques and architectures. We start with the existing canned Estimators and then we learn how to extend them with new functionality. With the existing code base, we would need to make numerous copies of the same `model_fn()` because a typical use case requires new `logit_fn()` and the rest of the code remains the same. This code duplication becomes obvious after writing more than five different estimators. E.g. adding batch norm layers is one example and using convolutional layers is another, they all require changes to `logits_fn()` only.

davorrunje:
@ispirmustafa please let me know if you have more questions.

martinwicke:
@ispirmustafa any update on this?

ispirmustafa:
Hi @davorrunje 
First of all thank you for your focus on making expansions easy. I see how common_model_fn fits well to your use cases and training. We don't wanna add one more layer in to existing model_fns. 
Thanks again

PR #13280, opened 2017-09-24
Run as the host user in docker when running docker in CI scripts.

gunan:


gunan:
NVM, It looks like I was looking at some old builds.

Issue #12486, opened 2017-08-22
Feature request: stop requiring the same dtype for inputs in tf.shape_n


leandro-gracia-gil:
As for Tensorflow 1.3, [tf.shape_n](https://www.tensorflow.org/api_docs/python/tf/shape_n) takes a list of tensors as input to produce a list of shapes as output. However, it produces an error if tensors of different types are provided. As far as I can tell, whether tensors are the same type or not is completely irrelevant to the behavior of this function, making it an arbitrary constraint that limits its functionality for no particular reason.

Would it be possible to remove such restriction if there's no good reason to keep it?

leandro-gracia-gil:
Note: this should probably have a very low priority since there's a quite simple workaround:

```python
def shape_n(input, out_type=tf.int32):
  return [tf.cast(tf.shape(t), out_type) for t in input]
```

That's not exactly equivalent since it's not a single named op with n outputs, but n unnamed ops with a single output. Yet, the result should be the same.

In any case, I'm merely rising the point for any future updates.

ali01:
We currently don't have anybody working on this. It would be great if you could help us by working on this and submitting a PR. Let us know if you need further clarification. Thanks!

ibrahimsharaf:
Hi @leandro-gracia-gil @ali01, I am new to open source, how can I start working on this?

sushreebarsa:
@leandro-gracia-gil Other TensorFlow functions like tf.stack and tf.concat already support heterogeneous tensor inputs and automatically perform type casting if necessary.  It is a TF 1.x related issue which needs to be upgraded to the latest as older versions are not actively supported. Kindly refer to the migration doc [here](https://www.tensorflow.org/guide/migrate). Please create a new ticket if you still have a concern?
Thank you!

PR #13282, opened 2017-09-24
Relex restrictions on `tf.foldr` and `tf.foldl`

### LINKED ISSUE #12019: foldr is too restrictive:  dimension 0 in both shapes must be equal
Using TF 1.2.0-rc1 on Ubuntu 16.04.

There are cases where `tf.foldr` should work, but is unable to because of a restriction that the first shape dimensions be identical for all list elements.  Below is a self-contained example that demonstrates the problem:

```
import tensorflow as tf

def concat2(A, B):
    return tf.concat([A, B], axis=0)

A = tf.constant([[10,10]])             # A.shape => (1,2)
B = tf.constant([[20, 20], [30, 30]])  # B.shape => (2,2)

print(concat2(A, B).eval())              # => [[10, 10], [20, 20], [30, 30]]
print(tf.foldr(concat2, [A, B]).eval())  # => ERROR!
```

yongtang:
This fix tries to address the issue raised in #12019 where `tf.foldr` and `tf.foldl` only accpet first shape dimensions be identical for all list elements.

The issue comes from the implementation of `tf.foldr` and `tf.foldl` where the input elems is converted to tensor, then converted to tensor array:
https://github.com/tensorflow/tensorflow/blob/1e1b3d90295f9396a25b9cfd4c571f7949716182/tensorflow/python/ops/functional_ops.py#L100-L106

It is possible to bypass the conversion to tensor, and go directly to tensor array, as long as the input elems is Iterable (and pass `infer_shape=False`)

This fix is an proposal to relex the restriction on `tf.foldr` and `tf.foldl` so that a list of elements with different shapes could be passed.

This fix fixes #12019.

### DIFF
--- a/tensorflow/python/kernel_tests/functional_ops_test.py
+++ b/tensorflow/python/kernel_tests/functional_ops_test.py
@@ -120,6 +120,29 @@ class FunctionalOpsTest(test.TestCase):
         self.assertEqual(len(variables.trainable_variables()), 1)
         self.assertAllEqual(1282, r.eval())
 
+
+  def testFoldr_NoInferShape(self):
+    # Test case for 12019 (foldr)
+    with self.test_session():
+      a = constant_op.constant([[10, 10]])
+      b = constant_op.constant([[20, 20], [30, 30]])
+
+      r = functional_ops.foldr(
+          lambda a, b: array_ops.concat([b, a], axis = 0),
+          [a, b])
+      self.assertAllEqual([[10, 10], [20, 20], [30, 30]], r.eval())
+
+  def testFoldl_NoInferShape(self):
+    # Test case for 12019 (foldl)
+    with self.test_session():
+      a = constant_op.constant([[10, 10]])
+      b = constant_op.constant([[20, 20], [30, 30]])
+
+      r = functional_ops.foldl(
+          lambda a, b: array_ops.concat([a, b], axis = 0),
+          [a, b])
+      self.assertAllEqual([[10, 10], [20, 20], [30, 30]], r.eval())
+
   # pylint: disable=unnecessary-lambda
   def testFold_Grad(self):
     with self.test_session():
--- a/tensorflow/python/ops/functional_ops.py
+++ b/tensorflow/python/ops/functional_ops.py
@@ -97,13 +97,30 @@ def foldl(fn, elems, initializer=None, parallel_iterations=10, back_prop=True,
       varscope.set_caching_device(lambda op: op.device)
       varscope_caching_device_was_none = True
 
+    # Try to convert elems to tensor
+    elems_tensor = None
+    try:
+      elems_tensor = ops.convert_to_tensor(elems, name="elems")
+    except ValueError:
+      pass
+
     # Convert elems to tensor array.
-    elems = ops.convert_to_tensor(elems, name="elems")
-    n = array_ops.shape(elems)[0]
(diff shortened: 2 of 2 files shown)


yongtang:
Not sure about the input value of `tf.foldr` and `tf.foldl` so this PR is more about a proposal.

mrry:
@ebrevdo: Can you take a look at the `TensorArray` manipulation here and see if it looks alright?

I'm a little dubious about these functions, because they don't appear to accept `nest`-ed components, and that seems like an equally likely feature request that might conflict with this one in annoying ways....

ebrevdo:
It would indeed be better to extent these functions to support nested inputs and outputs in the same way that map_fn does.

facaiy:
I found that most of ops use `ops.convert_to_tensor` to convert its inputs, however the method is really strict (same dimension, same dtype etc), and brings some inconvenience, for example #12486. Is it possible to find a more general and solid solution? say, create some more loose helper method?

yongtang:
@mrry @ebrevdo Thanks. I will update the PR with nested input.

yongtang:
@facaiy I guess some of the restrictions might be related to proto definition. For example, the issue #12486 you referred is the op `ShapeN` which requires the same type `T` for input:
https://github.com/tensorflow/tensorflow/blob/de46da6f748d2ea8472154bb067a361e8238568d/tensorflow/core/ops/array_ops.cc#L2096-L2102

It might be possible to pass the proto such that same type is not required. Though that would require proto update I guess.

yongtang:
@mrry @ebrevdo I looked into the way `map_fn` handles nested structures. However, it seems that `map_fn`-like method (and `scan`) expects the same (nested) structure for all elems along dimension 0. But this is not the case raised in #12019.

In #12019, the input was
```
def concat2(A, B):
    return tf.concat([A, B], axis=0)

A = tf.constant([[10,10]])             # A.shape => (1,2)
B = tf.constant([[20, 20], [30, 30]])  # B.shape => (2,2)

print(concat2(A, B).eval())              # => [[10, 10], [20, 20], [30, 30]]
print(tf.foldr(concat2, [A, B]).eval())  # => ERROR!
```

My understanding is that the above scenario will not be compatible with nested structure as `map_fn`.

For that I am thinking we might consider:
- Update the `tf.foldl` and `tf.foldr` to support `map_fn`-like operations that allows nested structure.
- Add method `foldl` and `foldr` to `TensorArray` so that tensors with different shapes could be supported.

Let me know if the above makes sense or not. If it does, then I can update this PR for nested structure support of `tf.foldl` and `tf.foldr`, and create another PR to address #12019 with `foldl` and `foldr` to `TensorArray`.

/cc @edgimar

ebrevdo:
I'd prefer not making any more changes to the TensorArray API.  Can all the
logic be done in tf.foldl and tf.foldr?

martinwicke:
@yongtang any update?

martinwicke:
@yongtang and update on this? Or did you make another PR?

yongtang:
Sorry for the late response. I have been thinking about addressing the issue of  #12019  while at the same time limit the changes to `foldr/foldl` (vs. TensorArray) but couldn't find an easy way to get around that. For now I think it probably make sense to close this PR. I will keep the issue #12019 in mind and if I could find some way to make it work then will reopen.

Thanks all for the help and sorry for the inconvenience.

Issue #6604, opened 2017-01-02
add summary_feed_dict support to tf.train.Supervisor


bodokaiser:
Example use case:

> We want to use `tf.summary.image` on a (computed) tensor which requires a `feed_dict` of some input tensor.

Current work around:

> Manage `summary_op` manually (inside training loop). Cannot use any higher level training helper.

I could also provide a PR on interest.

bodokaiser:
One possible solution I see with `tf.train.MonitoredTrainingSession`:

```python
with tf.train.MonitoredTrainingSession(hooks=hooks) as session:
    while not session.should_stop():
        session.run(train, feed_dict={
            mr: batch[0].eval(session=session),
            us: batch[1].eval(session=session),
        })
```

that we take the `feed_dict` argument of `session.run` and pass it down to every `hook` where we add that `feed_dict` as default to `tf.train.SessionRunArgs`.

yaroslavvb:
What about using feed_fn like in FeedingQueueRunner?

yaroslavvb:
BTW, you can use supervisor `loop` construct to add a modified summary writer:

```
sv = tf.train.Supervisor(..., summary_op=None)
with sv.managed_session() as sess:
    sv.loop(60, lambda: sv.summary_computed(sess, sess.run(summary_op, feed_dict=...)))
    <your training loop>
```

bodokaiser:
This problem is not limited to `TensorLoggingHook` and `SummarySaverHook` but also to custom `SessionRunHook`s which at the moment provide the modular interface to organize such things around your training / testing evaluation.

As we already have support for `init_dict` and `init_fn` on many interfaces (e.g. `tf.train.Supervisor`, `tf.train.Scaffold` I guess it would be consistent to add something similar for evaluation.

At the moment I calls to `monitored_session.run(fetches, feed_dict=...)` already forward this information:

```python
class LoggerHook(tf.train.LoggingTensorHook):

    def before_run(self, run_context):
        run_args = super().before_run(run_context)
        feed_dict = run_context.original_args.feed_dict

       # extract arguments from `monitored_session.run(...)` out of `feed_dict`

        return run_args
```

carlthome:
Related: http://stackoverflow.com/questions/41366248/using-tf-train-supervisors-default-summary-op-on-validation-data

yaroslavvb:
[...]

PR #13284, opened 2017-09-25
MonitoredTrainingSession supports partial restore from checkpoint

immars:
As mentioned in #6604 , `MonitoredTrainingSession` is preferred to `Supervisor`, I think it is useful to support partial restoration from checkpoint files in `MonitoredTrainingSession`.

The main idea is bypassing checkpoint restoration of `SessionManager`, and adding a new `CheckpointRestorerHook` to restore data from checkpoint, after `tf.Session` is created and `init_op` is run.

A new ctor parameter `restore_var_list=None` added to explicitly specify which part of variables to be restored. If not set, all variables are restored by default behavior of `Saver`.

This way the user can train a model, save checkpoint file, change the model a bit and fine-tune the new model by specifying `restore_var_list` when he/she continues training.


sb2nov:
Thanks for this PR, can you add unit-tests for the new functionality.

ispirmustafa (changes requested):
Thank you Lisen!
You can put new hook under: tf.contrib.training.
We'll get feedback from users. No need to change the MTS interface for now.

immars:
@ispirmustafa @sb2nov ok, I'll push when ready. Thanks.

immars:
@ispirmustafa so what I need to do is to move `CheckpointRestorerHook` into contrib, revert my changes to `MonitoredTrainingSession`, and provide another version of `MonitoredTrainingSession` in contrib, is that correct?

Because if I move `CheckpointRestorerHook` into contrib, I guess what's in `tensorflow.training` should not depend on anything in `contrib`, right?

ispirmustafa:
@immars 
yes, please add `CheckpointRestorerHook` to tf.contrib.training.
I think you don't need to modify `MonitoredTrainingSession` or provide similar in contrib. Users can create your hook and send it to the  `MonitoredTrainingSession`, right?

immars:
@ispirmustafa user also need to bypass checkpoint restoration of `SessionManager`, while he/she cannot set parameter `checkpoint_dir` to None because it's also for `CheckpointSaverHook` to save newer checkpoints. Otherwise, `SessionManager` would still complain about missing variables because it happens right after `tf.Session` is created, before hooks can get access to the `sess`.

It seems to me that it's not possible to achieve this without modifying `MonitoredTrainingSession`, because currently the same `Saver` instance is used to restore and save checkpoints.

please see https://github.com/tensorflow/tensorflow/pull/13284/files#diff-078d8aa8dc9179be5e99435311ade88fR352 , about `restore_checkpoint_dir`.

ispirmustafa:
Hi @immars
User can use `MonitoredSession` directly instead of `MonitoredTrainingSession`. Please document in your hook, how should they use it and give an example of usage with `MonitoredSession`.
Having another `MonitoredTrainingSession` inside contrib will confuse users.

immars:
@ispirmustafa ok, I'll move changes into contrib, as well as a separate helper method as example.

I think A good example of usage with `MonitoredSession` would be similar to `MonitoredTrainingSession`, because the interface of `MonitoredTrainingSession` is more user friendly than `MonitoredSession`. I would probably wrap `MonitoredSession` and hooks up in something like, if not use directly, `MonitoredTrainingSession` in my code.

please see `PartialRestoreSession` in `partial_restore_session.py`.

sb2nov:
Jenkins, test this please.

immars:
Jenkins, test this please.

immars:
@ispirmustafa is there anything else I should do?

tensorflow/contrib/training/python/training/partial_restore_session.py
```diff
@@ -0,0 +1,210 @@
...
+        ckpt = saver.get_checkpoint_state(self._dir)
+      else:
+        return
+
+    # Loads the checkpoint.
+    self._saver.restore(session, ckpt.model_checkpoint_path)
+    self._saver.recover_last_checkpoints(ckpt.all_model_checkpoint_paths)
+
+
+def PartialRestoreSession(master='',  # pylint: disable=invalid-name
```
ispirmustafa: PartialRestoreSession is a replication of MonitoredTrainingSession. I think we should not introduce a duplicated functionality until we verify usefulness of CheckpointRestorerHook.

AFAIU, the part you want to reuse is the code block under 'if checkpoint_dir'. I still think we should only add the hook in this PR. You can give an example usage in the pydoc of the hook. The code is simpler then this function. 
Only confusing part may be the summary hooks. You can create a utility function 'create_summary_hooks' to share code.

ispirmustafa (changes requested):

martinwicke:
@immars can you remove the new session class as Mustafa requested? 

All MonitoredTrainingSession does is add a few hooks for MonitoredSession. No need to make every combination of hooks a new class.

martinwicke:
Closing as stalled. Please open a new PR if there is progress.

PR #13291, opened 2017-09-25
Re-instate the plugin BUILD

DavidNorman:
This change re-instates the plugin BUILD link which was removed in commit 7de939bb74c5edbc2f45e77a5d4696e70bb59e5b

### DIFF
--- a/tensorflow/compiler/jit/BUILD
+++ b/tensorflow/compiler/jit/BUILD
@@ -33,6 +33,7 @@ cc_library(
     deps = [
         ":xla_cpu_device",
         ":xla_cpu_jit",
+        "//tensorflow/compiler/plugin",
     ] + if_cuda_is_configured([
         ":xla_gpu_device",
         ":xla_gpu_jit",
--- a/tensorflow/compiler/plugin/BUILD
+++ b/tensorflow/compiler/plugin/BUILD
@@ -1,0 +1,42 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Configuration file for an XLA plugin.
+
+  please don't check in changes to this file. to prevent changes appearing
+  in git status, use:
+
+  git update-index --assume-unchanged tensorflow/compiler/plugin/BUILD
+
+  To add additional devices to the XLA subsystem, add targets to the
+  dependency list in the 'plugin' target. For instance:
+
+    deps = ["//tensorflow/compiler/plugin/example:plugin_lib"],
+
+  ** Please don't remove this file - it is supporting some 3rd party plugins **
+"""
+
+licenses(["notice"])
+
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "plugin",
+    deps = [
+        #"//tensorflow/compiler/plugin/example:example_lib",
+    ],
+)
--- a/tensorflow/compiler/plugin/README.md
+++ b/tensorflow/compiler/plugin/README.md
@@ -1,0 +1,16 @@
+3rd party XLA devices
+---------------------
+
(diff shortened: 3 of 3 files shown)


drpngx:
@kayzhu what do you think? This was introduced in 169160056.

kayzhu:
@DavidNorman

Hi David, sorry for removing this in my earlier change: I wasn't quite sure what the empty BUILD file was for, and whether / how you were using it. 

I wonder:
1. depending on your need, is there an alternative solution than introducing a new directory with an empty BUILD file?
2. if not, I would suggest adding more comments in the BUILD file, or add README.md to this directory itself to indicate how it is being used, and why it is unwise to remove it :)

Thanks!

DavidNorman:
Thanks.  I realize it looks a bit like a nothing.  However, since the repo keepers are (rightly) suggesting that our 3rd party code remains out of the main repo - I think it is important to have a sensible and clean way of integrating it.  Having to make edits to other BUILD files, where there are many more targets, seems messy.

It is a shame that there isn't a 'find and include a target if it exists' option in bazel.   I played around with a few possibilities, but as you can see in the conversations in the my original pull requests, they were considered more messy than this way.

I'll add a README.md file to that directory.

drpngx:
I'm not 100% sure that this is needed. Can we iterate internally with @jart or @gunan and @yifeif on this?

kayzhu:
Adding @hawkinsp as well who may have some idea too.

@drpngx Sounds good -- it's possible that they may be aware of a more standard way for this.

@DavidNorman thanks for the update! I will see if @jart @gunan @yifeif has any suggestions for a possibly better solution for this issue.

jart:
I have limited familiarity with XLA but I understand the problem you're describing.

If they were all first party, you could have a `config_setting` like we do for jemalloc where the user passes `--config=with_xla_foo=true` to Bazel, which then gets triggered by `select()`. But if the labels are unknowable, then Bazel's design is you can only really do it by having the user define a new rule at the tip of the build tree.

For example, if XLA compiler was a cc_binary, you'd have your own rule with the standard set of plugins and a shell main.cc file that statically registers them and invokes the compiler. But since I'm assuming XLA is baked into TensorFlow's big .so file, it's probably not as trivial as writing a cc_binary() to define a custom tip for the entire TensorFlow build.

In that case, have you considered doing what custom ops do? Plugin is an alwayslink=1 library with that C++11 lambda singleton registration hack, which a Python file dynamically links at runtime.

DavidNorman:
@jart - I've tried the dynamic linking thing but both bazel and protobuf are not up to the job.  protobuf doesn't like having multiple copies of itself in the same address space, and bazel didn't have the features for a DSO which is linked down to other DSOs.  or something like that.  i was in touch with a person from the bazel team who was fixing this, so it probably does now.

gunan:
I think a recent change by @allenlavoie together with some pending work will help with having multiple copies of protobuf. Ill let Allen comment more on dynamic linking.

allenlavoie:
In principle we could do the same thing for XLA extensions that we do for op libraries and filesystems: include the registries and symbols needed to register new things in //tensorflow:libtensorflow_framework.so, then use an equivalent of tf.load_op_library to load a shared object (that object references symbols in libtensorflow_framework.so and does the registration; we could also add an XLA dlopen() wrapper to the C API for access in other languages). Right now all of the XLA symbols get added to pywrap_tensorflow directly, so that would require some rearranging.

After the RTLD_GLOBAL removal you should need to explicitly include protobufs to break things; otherwise if you use the @protobuf_archive//:protobuf_headers you'll just pick up the version in libtensorflow_framework.so, which should work fine.

This sort of registration via dynamic linking would be a bit of work, so in the meantime I don't see a problem with doing plugins this way...

DavidNorman:
That's good news regarding protobuf.  I did try, for a week, to generate a plugin DSO (hence the name of the driver directory).  I can see that protobuf was just not designed to support dynamic loading. Why would it be given where it has come from.  It doesn't have 16 bit type support for the same reason.

I would be delighted to be able to release just a loadable driver though, so I will watch the neatening of the interfaces with interest. 

cheers for the responses.

DavidNorman:
Hi all - so until the DSO changes that are required to get proper plugins working - can we reinstate these changes?

thanks

allenlavoie:
To be clear, I'm in favor but don't have much to do with XLA development. @kayzhu do you have objections?

drpngx:
Jenkins, test this please.

kayzhu:
@allenlavoie No objections from me. I will take another look at it shortly for final approve if other folks are fine with this change too.

gunan:
Also no objections from me.

gunan:
And the failures are unrelated.

drpngx:
I'm good with that. Sorry to delay this, I thought there would be a simpler solution.

kayzhu (approved):

DavidNorman:
awesome.  thanks guys

PR #13299, opened 2017-09-25
Add optional CheckpointSaverListener to MonitoredSession

magnusja:


magnusja:
I signed it!

ispirmustafa (changes requested):
Thank you Magnus.
Please add unit test. 
also mention in documentation that this argument will be used if (save_checkpoint_secs and save_checkpoint_secs > 0)

asimshankar:
Could you elaborate on the motivation here? 

IIUC, this provides a hook to the checkpoint hook :). Is there an alternative possible here where you provide your own MyCheckpointHook that composes CheckpointSaverHook with your custom code?

magnusja:
@asimshankar Yes that's right ;)
I have not thought about that, but that would be an option as well. Would you prefer an CheckpointSaverHook as an optional argument which is used when not None instead of creating one?

asimshankar:
@martinwicke @ispirmustafa : Could you guys take a look? It seems this is very similar to a change that was proposed to `tf.Estimator` that was also adding such a hook.

ispirmustafa:
We're adding listener argument to `tf.estimator.Estimator.train` since there are edge cases related to creation of `CheckpointSaverHook`.  User can create his/her own `CheckpointSaverHook` before calling `MonitoredTrainingSession`. It's also not a common thing. So I recommend not to change `MonitoredTrainingSession`.

martinwicke:
Closing in response to Mustafa's comment.

PR #13312, opened 2017-09-26
Support TensorArray in BeamSearchDecoder state.

### LINKED ISSUE #13154: BeamSearchDecoder should support an AttentionWrapper cell with alignment_history enabled

guillaumekln:
#13208 attempted to fix #13154 by representing the `alignment_history` field with a `Tensor` instead of a `TensorArray`. However, @ebrevdo pointed out that this approach led to a quadratic time and space overhead.

This PR fixes the issue by directly adding the support for `TensorArray` in the `BeamSearchDecoder` state as proposed by @ebrevdo.

@ebrevdo Let me know what you think of this implementation. Thanks!


sb2nov:
Jenkins, test this please.

sb2nov:
Jenkins, test this please.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -120,13 +120,79 @@ def tile_batch(t, multiplier, name=None):
...
 
-def _check_maybe(t):
+def _map_function(fn, t):
+  """Maps the function fn on each element of the TensorArray t."""
+  loop_stop = lambda mapped, elems, i: (
+      i == elems.size())
+  loop_body = lambda mapped, elems, i: (
+      mapped.write(i, fn(elems.read(i))), elems, i + 1)
+  mapped = tensor_array_ops.TensorArray(t.dtype, size=0, dynamic_size=True)
+  mapped, _, _ = control_flow_ops.while_loop(
```
ebrevdo: i have to think about this.  it sounds like at each time step you're reordering the entire history.  is that right?

guillaumekln: The idea is to keep the beams in the correct order at each timestep.

I guess it could also be done at the end of the decoding but the implementation here is more generic than just alignment history. The `TensorArray` might not contain time-based series.

steven-hh-ding: Thank you very much for the PR. I am also working on the same problem. Instead of picking out the right beams at each time step, we use gather_tree to sort out the right beams to avoid a quadratic complexity. Unfortunately the gather_tree only supports int32, so we did a dirty cast. It is fine for us as the alignments are only used for visualization on tensorboard. We update the finalize function and skip tensorarray for _merge_batch_beams and _split_batch_beams. 

```python
  def finalize(self, outputs, final_state, sequence_lengths):
    predicted_ids = beam_search_ops.gather_tree(
        outputs.predicted_ids, outputs.parent_ids,
        sequence_length=sequence_lengths)
    def gather_tree_from_array(t):
        if not isinstance(t, tensor_array_ops.TensorArray):
            return t
        history = t.stack()
        history = rnn._transpose_batch_time(history)
        history = math_ops.cast(history * 10000, dtypes.int32)
        history = beam_search_ops.gather_tree(
            history, outputs.parent_ids,
            sequence_length=sequence_lengths)
        history = math_ops.cast(history / 10000, t.dtype)
        return history
    final_state = nest.map_structure(gather_tree_from_array, final_state)
    outputs = FinalBeamSearchDecoderOutput(
        beam_search_decoder_output=outputs, predicted_ids=predicted_ids)
    return outputs, final_state
```

ebrevdo:
We can also update gathertree to support more input dtypes.

steven-hh-ding:
Thank you ebrevdo. I deleted the previous post since it contains several errors in our old code. The gather_tree only support tensor rank of 3. The dtype is fine. We can use gather_tree to sort out the right index of beams for each time step. And collect the alignment history using the index. I will post our code here once it is tested.

ebrevdo:
Thanks!

steven-hh-ding:
```python
def _gather_tree_for_array(t, parent_ids, sequence_length):
  """
  Convert tensor array that contains tensor with unsorted beams.
  Each element has shape [batch*beam, depth]
  Return sorted beam tensor array contains elements of shape
  [batch, beam, depth]
  Padding has a value of 0.
  """
  # (time, batch, beam)
  time = array_ops.shape(parent_ids)[0]
  batch = array_ops.shape(parent_ids)[1]
  beam = array_ops.shape(parent_ids)[2]
  sorted_indx = array_ops.expand_dims(
      array_ops.expand_dims(math_ops.range(beam), 0), 0)
  sorted_indx = array_ops.tile(sorted_indx, [time, batch, 1])
  sorted_beams = beam_search_ops.gather_tree(
      step_ids=sorted_indx, parent_ids=parent_ids,
      sequence_length=sequence_length)
  # all index of beam increase by 1. (-1 padding index becomes 0)
  # 0 indicates a padding beam which has zero alignments.
  sorted_beams = sorted_beams + 1

  def collect(collector, i):
      # concate a padding alignment (zeros) for each batch
      value = t.read(i)
      padding = array_ops.zeros([batch, 1, array_ops.shape(value)[-1]])
      value = array_ops.concat([padding, value], axis=1)

      # collect value according to the sorted_beams
      # cannot use gather_helper as we increase the beam size by one
      # so the final_shape has a different beam size
      range_ = array_ops.expand_dims(
          math_ops.range(batch) * (beam + 1), 1)
      gather_indices = array_ops.reshape(sorted_beams[i] + range_, [-1])
      sorted_value = array_ops.gather(
          array_ops.reshape(
              value, [batch * (beam + 1), -1]),
          gather_indices)
      sorted_value = array_ops.reshape(sorted_value, [batch, beam, -1])

      collector = collector.write(i, sorted_value)
      return collector, i + 1

  collected = tensor_array_ops.TensorArray(
      size=t.size(), dynamic_size=True, dtype=dtypes.float32)
  collected, _ = control_flow_ops.while_loop(
      lambda _, i: i < t.size(),
      collect,
      loop_vars=(collected, 0),
      parallel_iterations=1)
  return collected
```

test code: (extended from testGatherTree)

```python
step_ids = _transpose_batch_time(
    [[[1, 2, 3], [4, 5, 6], [7, 8, 9], [-1, -1, -1]], 
    [[1, 2, 3], [4, 5, 6], [7, 8, 9], [-1, -1, -1]]])
parent_ids = _transpose_batch_time(
    [[[0, 0, 0], [0, 1, 1], [2, 1, 2], [-1, -1, -1]], 
    [[0, 0, 0], [0, 1, 1], [2, 1, 2], [-1, -1, -1]]])
sequence_length = [[3, 3, 3],[3, 1, 3]]

# make a dummy alignment history, which is the tiled step_ids
step_ids = ops.convert_to_tensor(step_ids)
alignment_length = array_ops.constant(10)
alignment_history = tensor_array_ops.TensorArray(
    size=0, dynamic_size=True, dtype=dtypes.float32)
alignment_history = alignment_history.unstack(
    math_ops.cast(array_ops.tile(array_ops.expand_dims(step_ids,-1), [1,1,1,alignment_length]), 
            dtypes.float32))

sorted_history = _gather_tree_for_array(
    alignment_history, parent_ids, sequence_length).stack()

sorted_step_ids = beam_search_ops.gather_tree(
    step_ids=step_ids, parent_ids=parent_ids,
    sequence_length=sequence_length)

sorted_step_ids = array_ops.tile(
    array_ops.expand_dims(sorted_step_ids,-1), 
    [1,1,1,alignment_length])
sorted_step_ids = math_ops.cast(
    clip_ops.clip_by_value(
        sorted_step_ids, 
        0, 
        math_ops.reduce_max(step_ids)),
    dtypes.float32)

# the sorted history should be the same as the tiled sorted step id
print(np.array_equal(sorted_history.eval(), sorted_step_ids.eval()))
```

ebrevdo:
Steven, very nice!  Guillaume, are you interested in incorporating this
into your PR?

guillaumekln:
Thank you @steven-hh-ding. I integrated your code with some minor restyling and renaming.

So now, `TensorArray`s are ignored in `_maybe_split_batch_beams`, `_maybe_merge_batch_beams`, and `_maybe_tensor_gather_helper` and beams are eventually sorted in `finalize()`. Is that more appropriate?

If yes, @ebrevdo can review for details.

steven-hh-ding:
@guillaumekln  Thank you! Please feel free to modify anything. It was done in a rush.

guillaumekln:
The implementation should be revised after some changes on master. Will try to work on this again in the coming days.

martinwicke:
@guillaumekln any updates? It has accumulated another conflict.

guillaumekln:
Yes, I looked at it again yesterday. I will update this thread again when it's ready for review. Thanks.

guillaumekln:
I rebased the branch on master and revised the implementation following the recent changes made to `gather_tree`.

@ebrevdo Could you take a look when you have some time? Thanks.

martinwicke:
ping @ebrevdo

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -121,10 +121,69 @@ def tile_batch(t, multiplier, name=None):
...
+        gather_from=t.read(i),
+        batch_size=batch_size,
+        range_size=beam_width,
+        gather_shape=[batch_size * beam_width, -1])
+    return collector.write(i, gathered), i + 1
+
+  collected = tensor_array_ops.TensorArray(
+      t.dtype, size=t.size(), dynamic_size=False)
+  collected, _ = control_flow_ops.while_loop(
+      lambda _, i: i < t.size(),
```
ebrevdo: would it make sense here to just do a ta.stack() at each time step and run the (much faster) gather_tree op on this?  this inner loop looks REALLY slow.

add a comment in the top docstring that if requesting the alignment history, the complexity goes from `O(time_steps)` to `O(time_steps^2)`.

also add a test with `has_attention=False`, `has_alignment_history=True`.

guillaumekln: I'm not sure to understand. This loop is only called at the end of the decoding in `finalize()` and iterates over `Tensor`s. As  `gather_tree` works on `step_ids`, here we gather a generated sequence of ids that we then use to reorder each element of the `TensorArray`.

Am I missing something here?

Regarding the test, when `has_attention` is `False` there is no `AttentionWrapper` and no `alignment_history` by extension.

ebrevdo: so then my question remains the same.  why not just stack() the TensorArray and perform the reordering using something like `gather_nd`?

ebrevdo: (which would likely be much faster...)

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -806,3 +873,22 @@ def _tensor_gather_helper(gather_indices, gather_from, batch_size,
...
+    t: A `TensorArray` of size `max_time` that contains `Tensor`s of shape
+      `[batch_size, beam_width, depth]`.
+    parent_ids: The parent ids of shape `[max_time, batch_size, beam_width]`.
+    sequence_length: The sequence length of shape `[batch_size, beam_width]`.
+
+  Returns:
+    A `TensorArray` where beams are sorted in each `Tensor` or `t` itself if it
+    is not a `TensorArray`.
+  """
+  if isinstance(t, tensor_array_ops.TensorArray):
```
ebrevdo: There's no guarantee that the TensorArray has shapes that are amenable to sorting.  what if it contains scalars?  what if the the shape is not `[batch_size, beam_width, depth]` but instead someone is storing previous time steps' state having shape `[?, depth]` where `?` changes from element to element?  This is perfectly valid - you can't stack this TensorArray but you *can* `concat` it.

In other words, this function is making some serious assumptions about the fact that only an `AttentionWrapper` might create TensorArrays.  Guard against it; enable by default, but if there's a failure raise a warning message that the user should disable the TensorArray sorting.

Additional checks you should use:

1. Only perform this if `t` has `infer_shape=True`.
2. Only perform this if `t.element_shape` is rank 3 and element_shape[2:3] == [beam_width, depth].

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
steven-hh-ding: It sounds like a good idea by stacking and gather_nd, which will indeed be faster than current loop. i guess we need to unstack as well so the returned type is still tensorarray.

guillaumekln: How would you generate the indices for `gather_nd` efficiently? (@steven-hh-ding If you want to take a look that would be appreciated.)

The idea behind this loop is to apply the same gathering code (i.e. `_tensor_gather_helper`) on Tensors, whether they are directly in the cell state or within the TensorArray.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
guillaumekln: Currently, the assumptions are:

1. the TensorArrays size is the time dimension
2. the TensorArrays contain Tensors with the same shape assumption as Tensors that are directly in the cell state

> Guard against it; enable by default, but if there's a failure raise a warning message that the user should disable the TensorArray sorting.

Do you mean to try/catch the call to `gather_tree_from_array` and expose a flag to disable any post process on TensorArrays?

ebrevdo:
Expose a flag to disable TensorArray reordering.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
steven-hh-ding: if i remember the shapes here correctly:


```python
time = 3
batch = 4
beam = 5
depth = 6

# [time, batch, beam, depth]
a = tf.constant(np.arange(time * batch * beam * depth), shape=[time, batch, beam, depth], dtype=tf.int64)

# [time, batch, beam] (reverse beam as example)
beam_ids = tf.constant(np.flip(np.arange(beam), 0), shape=[1,1,beam], dtype=tf.int64)
beam_ids = tf.tile(beam_ids, [3,4,1])

# we need [time, batch, beam, 3]
# assume beam_ids > -1
t_bt_ind = tf.where(tf.ones_like(beam_ids) > -1)[:,:-1]
t_bt_ind = tf.reshape(t_bt_ind, [time,batch,beam,-1])
t_bt_be_ind = tf.concat([t_bt_ind, tf.expand_dims(beam_ids, -1)], -1)

# gather
tf.gather_nd(a, t_bt_be_ind)

```

Please double-check. Another way is to combine range and tile; but it is hardly readable. We use this method a lot in our code (tf.where(constant)), it looks okay. ebrevdo may better comment on this. Thanks!

ebrevdo:
tf.where(tf.ones_like(beam_ids) > -1)[:,:-1]


that line looks relatively expensive.


i guess you can do tf.where(beam_ids > -1)

truthfully i like range and tile better because it's a much faster
kernel than tf.where (tf.where is *really* expensive, especially on
gpu)

steven-hh-ding:
Oh cool. Good to know. Thank you @ebrevdo . @guillaumekln so we better use range and tile to have the indexes  like below. Please double check. Thanks!

```python
t_ind = tf.tile(tf.reshape(tf.range(time), [-1, 1, 1, 1]), [1, batch, beam, 1])
be_ind = tf.tile(tf.reshape(tf.range(batch), [-1, 1, 1, 1]), [1, time, beam, 1])
be_ind = tf.transpose(be_ind, perm=[1,0,2,3])  
t_be_ind = tf.concat([t_ind, be_ind, tf.expand_dims(beam_ids, -1)], -1)

tf.gather_nd(a, t_be_ind)
```

guillaumekln:
@steven-hh-ding Thanks for your help on this.
@ebrevdo I pushed new commits trying to address your feedback. Hopefully we are getting close.

In particular, I added more tests to check the reordering on elements with shape:

* `[batch_size, beam_width]`
* `[batch_size, beam_width, s1]`
* `[batch_size * beam_width, s1]`
* `[batch_size, beam_width, s1, s2]`

Reordering does not happen when the elements' shape cannot be inferred. Otherwise, it is check that they can be reshaped to `[batch_size, beam_width, -1]`. If not, an exception in raised suggesting to disable `reorder_tensor_arrays`.

tensorflow/contrib/seq2seq/python/kernel_tests/attention_wrapper_test.py
```diff
@@ -200,6 +200,9 @@ def _testWithMaybeMultiAttention(self,
           self.assertEqual(
               (None, batch_size, None),
               tuple(state_alignment_history.get_shape().as_list()))
+        nest.assert_same_structure(
+            cell.state_size,
+            cell.zero_state(batch_size, dtype('float32')))
```
ebrevdo: dtypes.float32

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -121,10 +121,73 @@ def tile_batch(t, multiplier, name=None):
...
+      `[batch_size, beam_width, s]` or `[batch_size * beam_width, s]` where
+      `s` is the depth shape.
+    parent_ids: The parent ids of shape `[max_time, batch_size, beam_width]`.
+    sequence_length: The sequence length of shape `[batch_size, beam_width]`.
+
+  Returns:
+    A `TensorArray` of the same size and type as `t` and where beams are sorted
+    in each `Tensor` according to `parent_ids`.
+  """
+  max_time = array_ops.shape(parent_ids)[0]
```
ebrevdo: parent_ids.shape[0].value or array_ops.shape(parent_ids)[0]
...
same for the other 2.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -121,10 +121,73 @@ def tile_batch(t, multiplier, name=None):
...
+      math_ops.reduce_max(sequence_length, axis=1))
+  sorted_beam_ids = beam_search_ops.gather_tree(
+      step_ids=masked_beam_ids,
+      parent_ids=parent_ids,
+      max_sequence_lengths=max_sequence_lengths,
+      end_token=beam_width + 1)
+
+  # For out of range steps, simply copy the same beam.
+  sorted_beam_ids = array_ops.where(
+      math_ops.cast(mask, dtypes.bool), x=sorted_beam_ids, y=beam_ids)
```
ebrevdo: i don't think you need the cast here anymore; we recently modified WhereOp to accept integers etc.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -445,15 +520,49 @@ def _maybe_merge_batch_beams(self, t, s):
...
+
+    Returns:
+      A `TensorArray` where beams are sorted in each `Tensor` or `t` itself if it
+      is not a `TensorArray` or does not meet shape requirements.
+    """
+    if (not isinstance(t, tensor_array_ops.TensorArray)
+        or not t._infer_shape or not t._element_shape):  # pylint: disable=protected-access
+      return t
+    shape = t._element_shape[0]  # pylint: disable=protected-access
+    batch_size = tensor_util.constant_value(self._batch_size)
```
ebrevdo: this may be None;' in which case the test at L554 is ill defined.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -121,10 +121,73 @@ def tile_batch(t, multiplier, name=None):
...
+  # Gather from a tensor with collapsed additional dimensions.
+  gather_from = t.stack()
+  final_shape = array_ops.shape(gather_from)
+  gather_from = array_ops.reshape(
+      gather_from, [max_time, batch_size, beam_width, -1])
+  ordered = array_ops.gather_nd(gather_from, indices)
+  ordered = array_ops.reshape(ordered, final_shape)
+
+  # Return the result as a TensorArray.
+  ordered_ta = tensor_array_ops.TensorArray(
```
ebrevdo: i'm wondering if we need to go through this unstack.  maybe just say in the initializer docstring "if the TensorArray can be reordered, it will be reordered and returned back in reordered and stacked form (i.e., as a regular tensor)".  TensorArrays that cannot be reordered will be returned in non-stacked form.

?

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -445,15 +520,49 @@ def _maybe_merge_batch_beams(self, t, s):
...
+        `s` is the depth shape.
+      parent_ids: The parent ids of shape `[max_time, batch_size, beam_width]`.
+      sequence_length: The sequence length of shape `[batch_size, beam_width]`.
+
+    Returns:
+      A `TensorArray` where beams are sorted in each `Tensor` or `t` itself if it
+      is not a `TensorArray` or does not meet shape requirements.
+    """
+    if (not isinstance(t, tensor_array_ops.TensorArray)
+        or not t._infer_shape or not t._element_shape):  # pylint: disable=protected-access
```
ebrevdo: use tf_logging.warn or tf_logging.info to describe why the TensorArray is not being sorted.

ebrevdo: consider adding more constraints here; like t._element_shape.ndims is None or t._element_shape.ndims < 2

ebrevdo:
looking very good!  main concern now is with cases where the batch_size is not known.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
guillaumekln: We could do that if it is properly documented but it would make the API a bit less consistent. Taking `alignement_history` as an example, the user would first expect the same output type from a `BeamSearchDecoder` or a `BasicDecoder`.

What about stacking all TensorArrays in the cell state in `finalize`? It seems like that is what users do anyway.

Let me know what you think is best.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
guillaumekln: Should we make this part of the condition to ignore TensorArrays reordering?

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
steven-hh-ding: thank you @guillaumekln +1 for the non-stacked form. eventually i still need to stack it up. unstack here and stack in finalize seems redundant. i would prefer as less unnecessary oprs as possible.  it is up to ebrevdo and you though.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
guillaumekln: `tf.where` is actually calling `gen_math_ops._select` here which appears to only work on booleans.

ebrevdo:
@guillaumekln it looks very close to done.  looking forward to approving this PR.

guillaumekln:
* Regarding the case where the batch size is not know, does it make sense to just try/catch the call to `gather_tree_from_array` instead of trying to figure out the batch and beam dimensions?
* For the return value, I applied your suggestion to return the stacked form for reordered arrays and added a notice in the docstring.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
ebrevdo: Oops, right.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -445,15 +517,55 @@ def _maybe_merge_batch_beams(self, t, s):
...
+
+    Returns:
+      A `TensorArray` where beams are sorted in each `Tensor` or `t` itself if it
+      is not a `TensorArray` or does not meet shape requirements.
+    """
+    if not isinstance(t, tensor_array_ops.TensorArray):
+      return t
+    if (not t._infer_shape or not t._element_shape
+        or t._element_shape[0].ndims is None or t._element_shape[0].ndims < 1):  # pylint: disable=protected-access
+      tf.logging.warn("The cell state contains a TensorArray that is not "
```
ebrevdo: Consider printing `t.name` here.

ebrevdo: also; provide some more information about the element_shape in the message.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -445,15 +517,55 @@ def _maybe_merge_batch_beams(self, t, s):
...
+                      "amenable to sorting based on the beam search result. "
+                      "For a TensorArray to be sorted, its elements shape "
+                      "must be defined and have at least a rank of 1.")
+      return t
+    shape = t._element_shape[0]  # pylint: disable=protected-access
+    batch_size = tensor_util.constant_value(self._batch_size)
```
ebrevdo: not sure how i feel about this.  i think if batch_size is statically known, as are the entries in `shape`, its ok to raise a ValueError here.  however, if they're not, you should probably push down an `Assert` on the shape of the output of `ta.stack()` to compare the run-time batch size and rank at that point.

ebrevdo (changes requested):
pretty much; done; i'll ask @oahziur to double check and also ask you add some better handling of dynamic batch_size and rank assertions.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
guillaumekln: This is done.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -445,15 +518,62 @@ def _maybe_merge_batch_beams(self, t, s):
...
+                      "TensorArray to be sorted, its elements shape must be "
+                      "defined and have at least a rank of 1, but saw shape: %s"
+                      % (t.handle.name, shape))
+      return t
+    shape = t._element_shape[0]
+    # pylint: enable=protected-access
+    batch_size = tensor_util.constant_value(self._batch_size)
+    reshaped_shape = tensor_shape.TensorShape(
+        [batch_size, self._beam_width, None])
+    if (shape.ndims >= 1 and shape[0] == batch_size * self._beam_width
```
ebrevdo: tensor_util.constant_value can return `None` here, in which case this line will fail because `None` cannot be multiplied with `self._beam_width`.  a unit test with placeholders as inputs would have caught this, I think.

guillaumekln:
Sorry for the delay. I revised the static and dynamic dimensions checking as requested, although it makes some heavy conditionals. Let me know if that is OK with you.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -189,6 +189,45 @@ def _check_maybe(t):
...
+  """
+  error_message = ("TensorArray reordering expects elements to be "
+                   "reshapable to [batch_size, beam_size, -1] which is "
+                   "incompatible with the dynamic shape of %s elements. "
+                   "Consider setting reorder_tensor_arrays to False to disable "
+                   "TensorArray reordering during the beam search."
+                   % (t.name))
+  return control_flow_ops.Assert(
+      math_ops.logical_and(
+          math_ops.greater_equal(array_ops.rank(t), 2),
```
ebrevdo: here for ease of reading, you can use:

array_ops.rank(t) >= 2

below you can use:

array_ops.rank(t) > 2

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -560,20 +599,12 @@ def _maybe_sort_array_beams(self, t, parent_ids, sequence_length):
...
-    reshaped_shape = tensor_shape.TensorShape(
-        [batch_size, self._beam_width, None])
-    if (shape.ndims >= 1 and shape[0] == batch_size * self._beam_width
-        or shape.ndims >= 2
-        and shape[0] == batch_size and shape[1] == self._beam_width):
+    _check_static_batch_beam_maybe(
+        shape, tensor_util.constant_value(self._batch_size), self._beam_width)
+    t = t.stack()
+    with ops.control_dependencies(
+        [_check_batch_beam(t, self._batch_size, self._beam_width)]):
```
ebrevdo: this should only happen once at the very end of the beam search, so it's probably ok to have.  it would be too expensive if it ran for every frame.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
```diff
@@ -560,20 +599,12 @@ def _maybe_sort_array_beams(self, t, parent_ids, sequence_length):
...
     # pylint: enable=protected-access
-    batch_size = tensor_util.constant_value(self._batch_size)
-    reshaped_shape = tensor_shape.TensorShape(
-        [batch_size, self._beam_width, None])
-    if (shape.ndims >= 1 and shape[0] == batch_size * self._beam_width
-        or shape.ndims >= 2
-        and shape[0] == batch_size and shape[1] == self._beam_width):
+    _check_static_batch_beam_maybe(
+        shape, tensor_util.constant_value(self._batch_size), self._beam_width)
+    t = t.stack()
```
ebrevdo: you may want to move the stack inside the control_dependencies.  sometimes people use element_shapes that have an unknown primary dimension, and they use `.concat()` instead of `.stack()`.  if someone's doing that, then this stack will fail because the individual elements have different shapes.  if you move the stack inside the control dependencies, they'll get a much better error message at runtime.

ebrevdo:
looks great!  minor nits.  a unit test with input placeholders to ensure the dynamic checks work correctly would be super helpful.  otherwise LGTM.

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
guillaumekln: How would you check the dynamic element shapes on a non stacked array?

ebrevdo: sigh; good point.  you emit a warning now, right?  perhaps that's enough for anyone who will run into this.  this really needs a new kernel like TensorArrayAllShapesMatch or something; but that's out of scope for this PR.

guillaumekln: If the elements shape is defined but the first dimensions is unknown it will go up to this point and fail without warning. I don't know how we can cover this case properly for now.

rmlarsen:
@ebrevdo another look?

tensorflow/contrib/seq2seq/python/ops/beam_search_decoder.py
ebrevdo: i think i prefer turning it into a warning (use tf_logging.warn) and hope that the user sees it when their session fails to run.  that's probably enough.  having unknown first dimension is more likely than a user using concat on a tensorarray; so we don't need to be too conservative here.

ebrevdo:
OK; two final comments:
 - 1. convert an error into a warning
 - 2. add a small unit test.  it LGTM otherwise.  no more review is required from me once you've made those changes.

tensorflowbutler:

rmlarsen:
@guillaumekln any change you could address the last minor comments?

guillaumekln:
Yes, sorry for the delay. I had other priorities lately but I will finish this soon.

@ebrevdo Good call about unit testing the dynamic checks as it actually does not work as is. As all logical operands are evaluated at the graph build time (I missed that), expressions like:

```                     math_ops.equal(array_ops.shape(t)[2], beam_width)
```

will fail if `t` has a rank lower than 3.

ttrouill:
Hi, no pressure, but is there a chance this will be merged soon? Many thanks

guillaumekln:
Sorry, this was long overdue. I fixed the dynamic checks and added some tests for them. I also converted the error raised when static checks fail into a warning.

rmlarsen:
@ebrevdo another look, please?

ebrevdo (approved):
awesome!  thanks!

rmlarsen:
@guillaumekln It looks like there are some linter errors. Could you take a look: https://source.cloud.google.com/results/invocations/9513608d-b992-4986-a1e6-b120c7235214/targets/%2F%2Ftensorflow%2Ftools%2Fci_build:gen_ci_sanity_out/log

guillaumekln:
Thanks, I fixed them.

rmlarsen:
@guillaumekln Thanks!

ebrevdo:
Thanks @guillaumekln for this PR.  It's very nice and will help users be able to get attention alignment histories with beam search.

Issue #13065, opened 2017-09-15
maxpooling error while building tf_label_image_example


lalwaniabhi:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:
   No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
   Windows 10 (64 bit)
- **TensorFlow installed from (source or binary)**:
   Using CMake and instructions present at (tensorflow/tensorflow/contrib/cmake)
- **Python version**: 
   3.6.1
- **CUDA/cuDNN version**:
   CUDA-8.0.61
   cuDNN-5.1
- **GPU model and memory**:
   NVIDIA GeForce GTX TITAN X (382.05)
- **Exact command to reproduce**:
   MSBuild /p:Configuration=Release tf_label_image_example.vcxproj

### Describe the problem
i am facing this error while trying to build the project tf_label_image_example. I was able to build this project without GPU support, but, while building with GPU support, i am facing the following error.

"C:\Users\alalwani\tensorflow\tensorflow\contrib\cmake\build\tf_label_image_example.vcxproj" (default target) (1) ->
(Link target) ->
maxpooling_op.obj : error LNK2001: unresolved external symbol "public: bool __cdecl tensorflow::functor::MaxPoolForwardWithOptionalArgmax::operator()(struct Eigen::QInt8 const *,int,int,int,int,int,int,int,int,int
,int,int,int,struct Eigen::QInt8 *,__int64 *,struct Eigen::GpuDevice const &)" (??R?$MaxPoolForwardWithOptionalArgmax@UQInt8@Eigen@@@functor@tensorflow@@QEAA_NPEBUQInt8@Eigen@@hhhhhhhhhhhhpeau34@PEA_JAEBUGpuDevice@4@@z) [C:\Users\alalwa
ni\tensorflow\tensorflow\contrib\cmake\build\tf_label_image_example.vcxproj]
C:\Users\alalwani\tensorflow\tensorflow\contrib\cmake\build\Release\tf_label_image_example.exe : fatal error LNK1120: 1 unresolved externals [C:\Users\alalwani\tensorflow\tensorflow\contrib\cmake\build\tf_label_image_example.vcxproj]

reedwm:
@mrry can you take a look?

mrry:
[...]

PR #13321, opened 2017-09-26
Disable broken MaxPool<qint8> and MaxPoolV2<qint8> kernels on Windows.

mrry:
This is intended to unbreak the Windows GPU build (#13065).


av8ramit:
http://ci.tensorflow.org/view/Nightly/job/nightly-win/323/

mrry:
And now we wait....

av8ramit:
So it seems that this fix fixes the original issue, but now we have a similar one with a new error signature. Merging for now.

av8ramit (approved):

PR #13325, opened 2017-09-26
Give accumulate_n op a gradient (version 2)

frreiss:
This pull request is a restructuring of the changes in https://github.com/tensorflow/tensorflow/pull/13022 to address high-level comments. Two major changes to call out vis a vis the previous PR:
* Instead of replacing `accumulate_n`, this PR now adds a new op, `accumulate_n_v2`. This new op is defined under `contrib/framework`.
* Refactoring changes have been broken out into a separate PR, https://github.com/tensorflow/tensorflow/pull/13159

Overall, this pull request addresses issue #10607 by adding a gradient to the existing `accumulate_n` operator. I followed the approach suggested by @alextp: rewrite `accumulate_n` as an atomic op which has a gradient defined for it and which gets rewritten by the runtime into the current implementation. Previously, this op had been implemented in Python as a constellation of lower-level ops, some of which are not differentiable.

**Implementation Details**
I have added a new C++ op, `AccumulateNV2`, which serves as a placeholder for type inference and gradient computation. A new rewrite, implemented in `accumulate_n_optimizer.cc`, replaces this placeholder with a group of `AssignAdd` ops and some additional ops that create, initialize, and destroy temporary variables.

I wrote a new Python wrapper function `accumulate_n_v2` that has the same signature as the original `accumulate_n` function. Unlike the original, `accumulate_n_v2` only validates its arguments and creates an instance of the `AccumulateNV2` placeholder op.

**Testing**
I added a more complete set of tests for `accumulate_n` in a previous pull request (https://github.com/tensorflow/tensorflow/pull/12196) to ensure that the op would still be correct after the changes in the current pull request. I copied all of the existing tests for `accumulate_n` into a test suite for the new `accumulate_n_v2` op (see tensorflow/contrib/framework/python/ops/accumulate_n_v2_test.py). I also added one additional test to verify that `accumulate_n_v2` now has a gradient. All the tests under `//tensorflow/python/...` currently pass on my MacOS and Linux test machines.

**Things to Note**
The semantics of the new implementation are broadly the same as the original, with the exception of one corner case. The original implementation allowed all the inputs to `accumulate_n` to have an undefined shape. My new code requires that at least one input have a defined shape; or that the user provides a shape using the `shape` argument to the `accumulate_n` function.

I had to put the graph rewrite code into the main build to get the rewrite to compile and run properly. It looks like some part of the API for adding an optimizer pass instantiates some static global data structures and can't be called from a dynamically-linked library.


sb2nov:
Jenkins, test this please.

jhseu:
Reassigning to Alex who has more context from reviewing the previous change.

tensorflow/contrib/framework/python/ops/accumulate_n_v2.py
```diff
@@ -0,0 +1,113 @@
...
+    name: A name for the operation (optional).
+
+  Returns:
+    A `Tensor` of same shape and type as the elements of `inputs`.
+
+  Raises:
+    ValueError: If `inputs` don't all have same shape and dtype or the shape
+    cannot be inferred.
+  """
+  loader.load_op_library(
```
alextp: I think it's fine to link the op into core tensorflow, making it be in core:all_ops and not use the dynamic loading thing.

tensorflow/contrib/framework/python/ops/accumulate_n_v2.py
```diff
@@ -0,0 +1,113 @@
...
+      resource_loader.get_path_to_datafile("_accumulate_n_v2_op.so"))
+  _INPUTS_ERR_MSG = ValueError("inputs must be a list of at least one Tensor"
+                               "with the same dtype and shape")
+  if context.in_eager_mode():
+    # TODO(apassos) remove this once the lifetime of eager variables gets
+    # addressed.
+    raise ValueError("accumulate_n not supported in eager mode")
```
alextp: Better yet, just use add_n here.

alextp (approved):

alextp:
Jenkins, test this please.

tensorflow/contrib/framework/python/ops/accumulate_n_v2.py
frreiss: Thanks, putting the op in `math_ops.cc` simplifies the changes under `contrib` significantly. I'll update the PR with that change once I've rerun the regression suite.

tensorflow/contrib/framework/python/ops/accumulate_n_v2.py
frreiss: Good idea. I'll put a fallback to `add_n` for eager mode at the bottom of this function, after the argument validation code. I'll also add a test case modeled after `OpsTest.testExecuteIntAttr()` in `python/eager/ops_test.py`.

alextp:
Jenkins, test this please.

alextp (approved):

Issue #9560, opened 2017-05-01
tf.pow(x, y) will freeze for negative integer y


korepwx:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10, and macOS 10.12.4
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: 1.1.0 on both operation systems
- **Bazel version (if compiling from source)**: not compiled from source
- **CUDA/cuDNN version**: CPU only on Windows, while CUDA 8.0 on macOS
- **GPU model and memory**: Nvidia Titan X, 12GB
- **Exact command to reproduce**: a short piece of Python code
- **Python version**: 3.5.2 (Anaconda) on Windows, 3.6.0 (Anaconda) on macOS

### Describe the problem

tf.pow(x, y) will freeze for negative integer y (and of course, integer x).  It will not freeze for negative floating number y.

### Source code / logs

```python
import tensorflow as tf

with tf.Graph().as_default(), tf.Session().as_default():
    print(tf.pow(5, -2).eval())  # this will not stop
```

aselle:
I've confirmed and reproduced this, thanks for reporting.

yongtang:
I think this issue is related to issue #12156. Also related to PR #11852.

codrut3:
I've looked in the source code and the origin of this bug is in Eigen, specifically src/Core/MathFunctions.h. They specialize their template for pow to the case of both parameters integers, and provide an implementation which only works for positive y. The problem is the well-known bug-loop of shifting y to the left: if y < 0, then the bit sign extends and it continues indefinitely.

Here is the code from Eigen:
`template<typename ScalarX,typename ScalarY>
struct pow_impl<ScalarX,ScalarY, true>
{
  typedef ScalarX result_type;
  static EIGEN_DEVICE_FUNC inline ScalarX run(ScalarX x, ScalarY y)
  {
    ScalarX res(1);
    eigen_assert(!NumTraits<ScalarY>::IsSigned || y >= 0);
    if(y & 1) res *= x;
    y >>= 1;
    while(y)
    {
      x *= x;
      if(y&1) res *= x;
      y >>= 1;
    }
    return res;
  }
};`

However, it would be difficult to fix this: the template specifically requests that the return type is Scalar, and not Real, and changing this in Eigen might break things for other people.

[...]

PR #13334, opened 2017-09-27
Make tf.pow work for integer inputs.

codrut3:
Currently tf.pow freezes if both inputs are integer tensors and y contains negative values. For example
the call
`tf.pow(5, -2).eval()`
never finishes.

The cause is the implementation of pow in Eigen. To get around this issue I've changed the
underlying C++ functor to treat differently integer inputs. However, tf.pow now returns float tensors on
input integer tensors.

In order to pass the backwards compatibility test, I had to create a new op IntegralPow for the case of
integer inputs. This mirrors the implementation of abs using Abs and ComplexAbs. I've made this
op hidden.

I've also added tests for the pow function in python.

This should fix issue #9560.

### DIFF
--- a/tensorflow/core/kernels/cwise_op_pow.cc
+++ b/tensorflow/core/kernels/cwise_op_pow.cc
@@ -16,12 +16,17 @@ limitations under the License.
 #include "tensorflow/core/kernels/cwise_ops_common.h"
 
 namespace tensorflow {
-REGISTER7(BinaryOp, CPU, "Pow", functor::pow, float, Eigen::half, double, int32,
-          int64, complex64, complex128);
+REGISTER5(BinaryOp, CPU, "Pow", functor::pow, float, Eigen::half, double,
+          complex64, complex128);
+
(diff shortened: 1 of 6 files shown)


sb2nov:
Jenkins, test this please.

sb2nov:
@codrut3 can you look at the test failures.

sb2nov:
Jenkins, test this please.

codrut3:
I did, and there were two types of failures:
1. I was handling dtype incorrectly in math_ops.pow (I fixed this)
2. There are use cases where it's natural to expect tf.pow to return an int32 tensor.
    For example, one of the tests did x ** 2, for x a tf.int32 tensor.
    Another computed tf.pow(2, layer), with layer a tf.constant.

These are reasonable use cases, and I don't see a way around them. I don't see how to fix the issue
without returning float for all integer inputs.

So I was thinking of closing the pull request. I apologize for this.

sb2nov:
Thanks for the contribution, if you can come up with a work-around please re-open.

PR #13399, opened 2017-09-30
Fix the Docker GPU build (adds a symlink + library path)

allenlavoie:
I was running a slightly modified build script the last time this passed, so you may want to test it again.

### DIFF
--- a/tensorflow/tools/docker/Dockerfile.devel-gpu
+++ b/tensorflow/tools/docker/Dockerfile.devel-gpu
@@ -78,9 +78,11 @@ WORKDIR /tensorflow
 
 # Configure the build for our CUDA configuration.
 ENV CI_BUILD_PYTHON python
-ENV LD_LIBRARY_PATH /usr/local/cuda/extras/CUPTI/lib64:$LD_LIBRARY_PATH
+ENV LD_LIBRARY_PATH /usr/local/cuda/lib64/stubs:/usr/local/cuda/extras/CUPTI/lib64:$LD_LIBRARY_PATH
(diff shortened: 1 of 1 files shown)


gunan (approved):

flx42:
Hello,

I don't understand what this PR is solving. You shouldn't add the stubs to your `LD_LIBRARY_PATH`, you don't want to use them at runtime, just at build time.

gunan:
We were able to work around a bazel bug with this PR.
@allenlavoie has more context.

allenlavoie:
Some recent changes mean that we're linking in GPU code when generating op wrappers (while building).

The tools for Bazel genrules depend on libtensorflow_framework.so, and that depends on CUDA libraries when building with CUDA. It's likely avoidable by not linking libtensorflow_framework.so for these genrules, and instead relying on the cc_library rules directly.

So we need a stub to satisfy that (trivial/incidental) dependency while building. But you're right that we should avoid putting stubs in LD_LIBRARY_PATH in the final image. I'll fix it in another PR.

flx42:
Ok. I would suggest adding the fixups around the code that require them, e.g. something like that:
```
RUN export LD_LIBRARY_PATH=/usr/local/cuda/lib64/stubs:$LD_LIBRARY_PATH && \
        ln -s /usr/local/cuda/lib64/stubs/libcuda.so /usr/local/cuda/lib64/stubs/libcuda.so.1 && \
        bazel [...] && \
        rm /usr/local/cuda/lib64/stubs/libcuda.so.1
```
Thanks!

Issue #13419, opened 2017-09-30
BUG: variables outside won't update in DNNLinearCombinedRegressor


facaiy:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac 10.11.6
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: 1.3.0
- **Python version**: 3.5
- **Bazel version (if compiling from source)**: N/A
- **CUDA/cuDNN version**: N/A
- **GPU model and memory**: N/A
- **Exact command to reproduce**: N/A

### Describe the problem

Variables outside won't update for `DNNLinearCombinedRegressor`, while everything is OK for `DNNRegressor`.

The bug stems from that only variables in dnn / linear scope will be updated in the code below:

https://github.com/tensorflow/tensorflow/blob/107cc777af7880c140d089e44ad898a6ba929286/tensorflow/python/estimator/canned/dnn_linear_combined.py#L214-L227

### Source code / logs

This is a tiny code to see whether `w` is updated or not.

```python
import numpy as np

import tensorflow as tf
from tensorflow import feature_column as fc
from tensorflow.python.summary import summary

tf.logging.set_verbosity(tf.logging.DEBUG)


BATCH_SIZE = 4


def input_fn():
    x = tf.constant(np.random.randn(BATCH_SIZE, 4), dtype=tf.float32)

    w = tf.Variable(np.array([1, 2, 3, 4]).reshape((4, 1)), dtype=tf.float32, name="test/w")
    summary.scalar("test/w_0_0", w[0][0])
    summary.scalar("test/w_1_0", w[1][0])
    summary.scalar("test/w_2_0", w[2][0])
    summary.scalar("test/w_3_0", w[3][0])

    y = tf.matmul(x, w)
    label = tf.constant(np.random.randint(0, 1, size=(BATCH_SIZE,)))

    return {"y": y}, label

f = fc.numeric_column("y")

def gen_estimator(cls, model_dir):
    if cls == "dnn":
        return tf.estimator.DNNRegressor(
                feature_columns=[f],
                hidden_units=[2],
                model_dir=model_dir)
    else:
        return tf.estimator.DNNLinearCombinedRegressor(
                dnn_feature_columns=[f],
                dnn_hidden_units=[2],
                model_dir=model_dir)

[...]

PR #13420, opened 2017-09-30
Fix bug: variables outside won't update in DNNLinearCombinedRegressor

facaiy:
This fix is an proposal for #13419 

There are two solutions to fix the problem in my opinion:
+ all variables except of linear is optimized by dnn.
+ which one of optimizers does the variables left (except of linear and dnn) use is selected by user, perhaps a new argument is needed.

The PR is implemented by the first one for simplicity.

### How to test

+ [x] add test case: I have no idea of how to write it, hence any suggestions will be appreciated.
+ [ ] pass all tests.

### DIFF
--- a/tensorflow/python/estimator/canned/dnn_linear_combined.py
+++ b/tensorflow/python/estimator/canned/dnn_linear_combined.py
@@ -192,7 +192,7 @@ def _dnn_linear_combined_model_fn(
     global_step = training_util.get_global_step()
 
     if model == "combined":
-      all_vars = ops.trainable_variables()
+      all_vars = ops.get_collection(ops.GraphKeys.TRAINABLE_VARIABLES)
       linear_vars = ops.get_collection(
         ops.GraphKeys.TRAINABLE_VARIABLES,
         scope=linear_parent_scope)
--- a/tensorflow/python/estimator/canned/dnn_linear_combined.py
+++ b/tensorflow/python/estimator/canned/dnn_linear_combined.py
@@ -196,7 +196,7 @@ def _dnn_linear_combined_model_fn(
       linear_vars = ops.get_collection(
         ops.GraphKeys.TRAINABLE_VARIABLES,
         scope=linear_parent_scope)
-      # variables left is optimized by dnn
+      # variables left are optimized by dnn
       dnn_vars = list(set(all_vars) - set(linear_vars))
       train_ops = [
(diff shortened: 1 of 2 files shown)


facaiy:
Hi, @ispirmustafa . I am not sure whether I understand correctly, could you take a look (or verify)?

tensorflow/python/estimator/canned/dnn_linear_combined.py
```diff
@@ -179,31 +179,34 @@ def _dnn_linear_combined_model_fn(
...
-
-    train_op = control_flow_ops.group(*train_ops)
+
+    if model == "combined":
+      all_vars = ops.trainable_variables()
+      linear_vars = ops.get_collection(
+        ops.GraphKeys.TRAINABLE_VARIABLES,
+        scope=linear_parent_scope)
+      # variables left are optimized by dnn
+      dnn_vars = list(set(all_vars) - set(linear_vars))
```
facaiy: Perhaps it will be performance bottlenecks, better proposals are welcome and appreciated.

sb2nov:
Jenkins, test this please.

facaiy:
Hi, the failure is unrelated. Could you retest it? Anyway, I'll try to add a test case later.

ispirmustafa (commented):
Hi Yan,
We don't recommend having variables inside input_fn. It will not work in some cases such as TPUEstimator. 
So let's not do this change since it will allow to create variables inside input_fn.

facaiy:
Hi, @ ispirmustafa . I agree with you that `input_fn` is not recommend to create variable, perhaps `Dataset` will be better or not. Anyway, it is not important. What I concerned is that the limit might be too strict (for custom).

For example, In practice, sentence features are not uncommon to make some preprocessing. We might split them, and build a shallow convolution for these features. So what is the best solution for the case?
+ hack `input_fn`: easy, however rebuild need be careful (export model and load model).
+ inherit `_FeatureColumn`: private class, and more code.
+ create custom `Estimator`: more code.

As far as I known, feature engineering is fundamental to the traditional application of machine learning. However, it seems that we, unfortunately, cannot make some complicate preprocession for now. That's why I hope to removing the restriction.

facaiy:
How about creating a `Lambda` feature column, like `Lambda` layer of keras ?  It seems easier to customize our pre-procession.

PR #13421, opened 2017-09-30
Java API Generics Phase 3.

andrewcmyers:
- Added the utility Tensors class.
- Updated tests to use it.
- Updated scripts for generating Tensors.java and the types directory.
  Note that these are still run manually, but remain helpful because
  maintaining so many methods and their documentation is a headache.
- Added missing checking for attempts to create tensors from arrays of boxed
  primitives, and a test case.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Tensor.java
@@ -94,8 +94,9 @@ public final class Tensor<T> implements AutoCloseable {
    * Tensor<String> m = Tensor.create(matrix, String.class);
    * }</pre>
    *
-   * @param obj The object to convert to a Tensor<T>. Note that whether the it is compatible with
-   *     the type T is not checked by the type system.
+   * @param obj The object to convert to a Tensor<T>. Note that whether it is compatible with the
+   *     type T is not checked by the type system. For type-safe creation of tensors, use {@link
+   *     Tensors}.
    * @param type The class object representing the type T.
    * @throws IllegalArgumentException if {@code obj} is not compatible with the TensorFlow type
    *     system.
@@ -174,7 +175,7 @@ public final class Tensor<T> implements AutoCloseable {
    *
    * <p>Creates a Tensor with the given shape by copying elements from the buffer (starting from its
    * current position) into the tensor. For example, if {@code shape = {2,3} } (which represents a
-   * 2×3 matrix) then the buffer must have 6 elements remaining, which will be consumed by this
+   * 2x3 matrix) then the buffer must have 6 elements remaining, which will be consumed by this
    * method.
    *
    * @param shape the tensor shape.
@@ -457,7 +458,7 @@ public final class Tensor<T> implements AutoCloseable {
    * @param dst the destination buffer
    * @throws BufferOverflowException If there is insufficient space in the given buffer for the data
    *     in this tensor
-   * @throws IllegalArgumentException If the tensor datatype is not {@link Integer}
+   * @throws IllegalArgumentException If the tensor data type is not {@link Integer}
    */
   public void writeTo(IntBuffer dst) {
     if (dtype != DataType.INT32) {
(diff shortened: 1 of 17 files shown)


andrewcmyers:
@asimshankar Java Generics Phase 3 PR started.

Some issues to consider:
- Do we want to change the type parameter for strings?  It's genuinely unclear to me how often encoding mistakes will happen and how much TFString would actually help. The idea of getting more opinions sounds good; I can go either way.
- Should the scripts for code generation be ported to some other language? They are not part of the build process, so maybe it's not so critical.
- We will probably want parallel methods for creating constant Tensors, in `op.Constant`, folding together `Constant.createWithTensor` + `Tensors.create`. I'm thinking that can go into a later PR.

sb2nov:
Jenkins, test this please.

asimshankar:
Thanks for the PR. A few quick things:

- RE: scripts to generate: Yes, we should port to another language (or heck, use an `AnnotationProcessor` :)). But can defer that for now because, as you point out, the generation scripts are not part of the build process (and have to be manually invoked). However, I didn't see the generated `Tensors.java` in this PR, shouldn't that be included too?

- RE: Merging `Constant.createWithTensor` and `Tensors.create` sounds like a great idea, but again, something that we can do later on.

Please update the PR with the generated `Tensors.java`.
Thanks

tensorflow/java/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -637,11 +638,25 @@ private static void throwExceptionIfNotByteOfByteArrays(Object array) {
    * represent more than one TensorFlow data type; for example, 'byte' can represent both {@code
    * uint8} and {@code string}, with the latter being the default interpretation.
    */
-  private static DataType dataTypeOf(Object o) {
+  private static Class<?> baseObjType(Object o) {
```
asimshankar: Javadoc above needs to be updated.
(Or you can just remove it - 5 line private helper methods don't really need a Javadoc :)

tensorflow/java/src/main/java/org/tensorflow/Tensors.java
```diff
@@ -0,0 +1,427 @@
+package org.tensorflow;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/** Utility class offering convenience methods for creating Tensors in a type-safe way. */
```
asimshankar: Optional: How about wording this as:

```java
/** Type-safe factory methods for creating {@link Tensor} objects. */
```

tensorflow/java/src/main/java/org/tensorflow/Tensors.java
```diff
@@ -0,0 +1,427 @@
...
+public class Tensors {
```
asimshankar: Nit: `public final class`?

tensorflow/java/src/main/java/org/tensorflow/types/UInt8.java
```diff
@@ -17,5 +17,8 @@
 
 /** Represents an 8-bit unsigned integer. */
 public class UInt8 {
-  private UInt8() {}
+  private byte value;
```
asimshankar: I didn't quite follow the point of these private constructors and value, could you elaborate?

asimshankar (commented):
A few minor suggestions, otherwise looks great. Thanks!

FYI @kbsriram @karllessard

tensorflow/java/src/main/java/org/tensorflow/types/UInt8.java
andrewcmyers: This was setting up to support a future generic `scalarValue()` accessor. But I was getting ahead of myself...

asimshankar (approved):

asimshankar:
Thanks @andrewcmyers ! Much appreciated.

asimshankar:
Jenkins, test this please

asimshankar:
Jenkins, test this please

PR #13424, opened 2017-09-30
Support weight_column for time series regressor

terrytangyuan:


allenlavoie:
Right now the Estimators operate on a single series. What is getting weighted here? Parts of the series?

terrytangyuan:
I guess I should've spent some more time on the input pipeline. I was
assuming those estimators support multiple series (for example multivariate
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/timeseries/examples/multivariate.py)
so I could use part of the series (one of the columns in the data)  as
weights. Please correct me if I misunderstood something here.

allenlavoie:
Ah, so the idea is that each feature of the series can have its own weight? I suppose that could be useful. Could you add a unit test or two?

terrytangyuan:
I am basically trying to complete feature parity against the heads used by
non-ts estimators, e.g.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/estimator/canned/head.py#L786

Right now it's using the shared _weights() function in canned/head.py to
extract the weight column from the list of columns from input.

terrytangyuan:
Got distracted by my projects recently. I am going to close this one for now until I found more time.

PR #13451, opened 2017-10-02
Add support for CUBLAS_TENSOR_OP_MATH in fp16 GEMM

nluehr:
- Applies to matrix multiplications with fp16 input/output.
  Computations will fall back to pseudo-fp16 if tensor op math is
  disabled or not supported.
- Enabled by default, but can be disabled by setting the environment
  variable TF_DISABLE_TENSOR_OP_MATH=1.

### DIFF
--- a/tensorflow/stream_executor/cuda/cuda_blas.cc
+++ b/tensorflow/stream_executor/cuda/cuda_blas.cc
@@ -36,6 +36,7 @@ limitations under the License.
 #include <assert.h>
 #include <complex>
 
+#include "tensorflow/core/util/env_var.h"
 #include "tensorflow/stream_executor/cuda/cuda_activation.h"
 #include "tensorflow/stream_executor/cuda/cuda_gpu_executor.h"
 #include "tensorflow/stream_executor/cuda/cuda_helpers.h"
@@ -268,6 +269,11 @@ PERFTOOLS_GPUTOOLS_CUBLAS_WRAP(cublasSgemmEx)
 PERFTOOLS_GPUTOOLS_CUBLAS_WRAP(cublasGemmEx)
 #endif
 
+#if CUDA_VERSION >= 9000
+PERFTOOLS_GPUTOOLS_CUBLAS_WRAP(cublasGetMathMode)
+PERFTOOLS_GPUTOOLS_CUBLAS_WRAP(cublasSetMathMode)
+#endif
+
 }  // namespace wrap
 
 static string ToString(cublasStatus_t status) {
@@ -297,6 +303,18 @@ static string ToString(cublasStatus_t status) {
     default:
       return port::StrCat("<invalid cublas status: ", status, ">");
   }
+}
+
+// Decide whether to enable TENSOR_OP_MATH
+static bool TensorOpMathEnabled() {
+  static bool is_enabled = [] {
+    bool is_disabled;
+    TF_CHECK_OK(
+        tensorflow::ReadBoolFromEnvVar("TF_DISABLE_CUBLAS_TENSOR_OP_MATH",
+                                       /*default_val=*/false, &is_disabled));
+    return !is_disabled;
+  }();
+  return is_enabled;
 }
 
 // cuBLAS has interfaces that permit pointers to be passed from either the host
@@ -359,6 +377,66 @@ class ScopedCublasPointerMode {
   cublasPointerMode_t old_mode_;  // Prior cuBLAS pointer mode, to be restored.
   bool ok_;                       // Whether the change was successful.
 };
+
+#if CUDA_VERSION >= 9000
+// cuBLAS has interfaces that permit computations to use the Tensor Cores
+// available in Volta hardware. This must be enabled via the
+// cublasGet/SetMathMode APIs.
+//
+// This helper sets the cuBLAS math mode to a desired value for a cuBLAS call
+// you are about to perform in a given scope.
+//
+// The prior cuBLAS math mode is retained and restored when this object goes
+// out of scope.
+class ScopedCublasMathMode {
+ public:
+  // Note that, because the setting of the cublas math mode is fallible,
+  // construction of this scoped datatype must be paired with a call to
+  // Init().
+  //
+  // Parameters:
+  //  handle: The cublas library handle to act upon in setting the math mode.
+  explicit ScopedCublasMathMode(CUDAExecutor *parent, cublasHandle_t handle)
+      : parent_(parent), handle_(handle), ok_(false) {}
+
+  // Attempts the switch to the requested scoped math mode, new_mode.
+  //
+  // Note that when false is returned, an appropriate error has already been
+  // logged.
+  bool Init(cublasMath_t new_mode) {
+    cublasStatus_t ret = wrap::cublasGetMathMode(parent_, handle_, &old_mode_);
+    if (ret != CUBLAS_STATUS_SUCCESS) {
+      LOG(ERROR) << "failed to get old cublas math mode: " << ToString(ret);
+      return ok_ = false;
+    }
+
+    ret = wrap::cublasSetMathMode(parent_, handle_, new_mode);
+    if (ret != CUBLAS_STATUS_SUCCESS) {
+      LOG(ERROR) << "failed to set new cublas math mode: " << ToString(ret);
+      return ok_ = false;
+    }
+    return ok_ = true;
+  }
+
+  // Switches back to the prior math mode, if the switch operation was
+  // successful in the first place.
+  ~ScopedCublasMathMode() {
+    if (ok_) {
+      cublasStatus_t ret = wrap::cublasSetMathMode(parent_, handle_, old_mode_);
+      if (ret != CUBLAS_STATUS_SUCCESS) {
+        LOG(ERROR) << "failed to set former cublas math mode: "
+                   << ToString(ret);
+      }
+    }
+  }
+
+ private:
+  CUDAExecutor *parent_;   // Executor establishing this math mode for.
+  cublasHandle_t handle_;  // Handle to the cuBLAS instance of interest.
+  cublasMath_t old_mode_;  // Prior cuBLAS math mode, to be restored.
+  bool ok_;                // Whether the change was successful.
+};
+#endif  // CUDA_VERSION >= 9000
 
 bool CUDABlas::Init() {
   cublasStatus_t ret = wrap::cublasCreate(parent_, &blas_);
@@ -532,7 +610,7 @@ cudaDataType_t CUDAComputationType(blas::ComputationType ty) {
 template <typename FuncT, typename... Args>
 bool CUDABlas::DoBlasInternalImpl(FuncT cublas_func, Stream *stream,
                                   bool pointer_mode_host, bool err_on_failure,
-                                  Args... args) {
+                                  bool use_tensor_op_math, Args... args) {
   mutex_lock lock{mu_};
 
   CHECK(blas_ != nullptr);
@@ -545,7 +623,14 @@ bool CUDABlas::DoBlasInternalImpl(FuncT cublas_func, Stream *stream,
                                            : CUBLAS_POINTER_MODE_DEVICE)) {
     return false;
   }
-
+#if CUDA_VERSION >= 9000
+  ScopedCublasMathMode math_mode{parent_, blas_};
+  if (use_tensor_op_math) {
+    if (!math_mode.Init(CUBLAS_TENSOR_OP_MATH)) {
+      return false;
+    }
+  }
+#endif
   cublasStatus_t ret = cublas_func(parent_, blas_, args...);
   if (err_on_failure && ret != CUBLAS_STATUS_SUCCESS) {
     LOG(ERROR) << "failed to run cuBLAS routine " << cublas_func.kName << ": "
@@ -1762,14 +1847,26 @@ bool CUDABlas::DoBlasGemm(
                       "precondition violation";
     }
   }
-  // TODO(sesse): Consider supporting the Hgemm interface, which uses half
-  // calculations internally (faster on newer devices, such as Pascal and TX1,
-  // but less precise).
-  return DoBlasInternal(
+
+  bool use_tensor_ops = false;
+#if CUDA_VERSION >= 9000
+  int cc_major, cc_minor;
+  stream->parent()->GetDeviceDescription().cuda_compute_capability(&cc_major,
+                                                                   &cc_minor);
+
+  // GPUs < sm_70 don't support tensor cores
+  if (cc_major >= 7 && TensorOpMathEnabled()) {
+    use_tensor_ops = true;
+  }
+#endif
+
+  return DoBlasInternalImpl(
       wrap::cublasSgemmEx, stream, true /* = pointer_mode_host */,
-      CUDABlasTranspose(transa), CUDABlasTranspose(transb), m, n, k, &alpha,
-      CUDAMemory(a), SE_CUDA_DATA_HALF, lda, CUDAMemory(b), SE_CUDA_DATA_HALF,
-      ldb, &beta, CUDAMemoryMutable(c), SE_CUDA_DATA_HALF, ldc);
+      true /* = err_on_failure= */, use_tensor_ops, CUDABlasTranspose(transa),
+      CUDABlasTranspose(transb), m, n, k, &alpha, CUDAMemory(a),
+      SE_CUDA_DATA_HALF, lda, CUDAMemory(b), SE_CUDA_DATA_HALF, ldb, &beta,
+      CUDAMemoryMutable(c), SE_CUDA_DATA_HALF, ldc);
+
 #else
   LOG(ERROR) << "fp16 sgemm is not implemented in this cuBLAS version "
              << "(need at least CUDA 7.5)";
@@ -2031,6 +2128,26 @@ bool CUDABlas::DoBlasGemmWithProfilingImpl(
   return result;
 }
 
+static bool UsesTensorOps(blas::AlgorithmType algo) {
+#if CUDA_VERSION >= 9000
+  cublasGemmAlgo_t cublas_algo = static_cast<cublasGemmAlgo_t>(algo);
+  return cublas_algo >= CUBLAS_GEMM_DEFAULT_TENSOR_OP;
+#else
+  return false;
+#endif
+}
+
+template <typename InType>
+static bool TensorOpsAvailable(int cc_major) {
+#if CUDA_VERSION >= 9000
+  if (cc_major >= 7 && TensorOpMathEnabled() &&
+      std::is_same<InType, Eigen::half>::value) {
+    return true;
+  }
+#endif
+  return false;
+}
+
 template <typename InT, typename OutT, typename CompT>
 bool CUDABlas::DoBlasGemmWithAlgorithmImpl(
     Stream *stream, blas::Transpose transa, blas::Transpose transb, uint64 m,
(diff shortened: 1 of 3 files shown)


nluehr:
For review by @zheng-xq.

frankchn:
Jenkins, test this please.

martinwicke:
Jenkins, test this please. (running again in case this was infra)

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -1762,14 +1921,26 @@ bool CUDABlas::DoBlasGemm(
...
-      ldb, &beta, CUDAMemoryMutable(c), SE_CUDA_DATA_HALF, ldc);
+
+  bool use_tensor_ops = false;
+#if CUDA_VERSION >= 9000
+  int cc_major, cc_minor;
+  stream->parent()->GetDeviceDescription().cuda_compute_capability(&cc_major,
+                                                                   &cc_minor);
+
+  // GPUs < sm_70 don't support tensor cores
+  if (cc_major >= 7 && TensorOpMathEnabled()) {
```
zheng-xq: I'm not a big fan of parsing env-var on every matmul-calls. Either cache this globally, or have the caller cache it and pass it in. 

The problem with a global setting is that you will have to test it separately. Since there is no way to unset it to not affect other tests. 

With TF, this can be cached at the op construction time, and therefore save the trouble.

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -2069,10 +2242,10 @@ bool CUDABlas::DoBlasGemmWithAlgorithmImpl(
   static_assert(blas::kDefaultGemmAlgo == CUBLAS_GEMM_DFALT, "");
   bool result = DoBlasInternalFailureOK(
       wrap::cublasGemmEx, stream, /* pointer_mode_host = */ true,
-      CUDABlasTranspose(transa), CUDABlasTranspose(transb), m, n, k, &alpha,
-      CUDAMemory(a), cuda_in_type, lda, CUDAMemory(b), cuda_in_type, ldb, &beta,
-      CUDAMemoryMutable(c), CUDADataType<OutT>::type, ldc,
-      CUDAComputationType(computation_type),
+      false /* = use_tensor_ops */, CUDABlasTranspose(transa),
```
zheng-xq: This is the part that the matmul autotuning works. Should we add that to part of the autotuning? 

DoBlasGemmWithAlgorithmImpl

This feature is currently disabled by default through this env-var "TF_MATMUL_AUTOTUNE_ENABLE", due to large noise. But our plan is to enable that. So it is a good idea to make sure it still works.

zheng-xq (changes requested):

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -558,47 +641,53 @@ bool CUDABlas::DoBlasAsum(Stream *stream, uint64 elem_count,
...
+      incx, CUDAMemoryMutable(result));
 }
 
 bool CUDABlas::DoBlasAxpy(Stream *stream, uint64 elem_count, float alpha,
                           const DeviceMemory<float> &x, int incx,
                           DeviceMemory<float> *y, int incy) {
   return DoBlasInternal(wrap::cublasSaxpy, stream,
-                        true /* = pointer_mode_host */, elem_count, &alpha,
+                        true /* = pointer_mode_host */,
+                        false /* = use_tensor_ops */, elem_count, &alpha,
```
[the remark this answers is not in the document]
yzhwang: Second on XQ's comment. FYI, we have done this with DoFusedConvolve() to support fused conv: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/stream_executor/cuda/cuda_dnn.cc#L2461

tensorflow/stream_executor/cuda/cuda_blas.cc
yzhwang: I'd suggest that we add tensor_op support here, while include the setting in autotune in TensorFlow's op level by adding ways to propagate use_tensor_ops option to matmul op.

yzhwang (commented):
Hi Nathan, were you able to make some changes to this pull request? After that I can give another pass of review. Thank you!

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -360,6 +377,66 @@ class ScopedCublasPointerMode {
...
+  // Parameters:
+  //  handle: The cublas library handle to act upon in setting the math mode.
+  explicit ScopedCublasMathMode(CUDAExecutor *parent, cublasHandle_t handle)
+      : parent_(parent), handle_(handle), ok_(false) {}
+
+  // Attempts the switch to the requested scoped math mode, new_mode.
+  //
+  // Note that when false is returned, an appropriate error has already been
+  // logged.
+  bool Init(cublasMath_t new_mode) {
```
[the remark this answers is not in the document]
nluehr: The separate init function is used to better handle errors. This follows the existing approach used for ScopedCublasPointerMode.

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -360,6 +377,66 @@ class ScopedCublasPointerMode {
...
+    cublasStatus_t ret = wrap::cublasGetMathMode(parent_, handle_, &old_mode_);
+    if (ret != CUBLAS_STATUS_SUCCESS) {
+      LOG(ERROR) << "failed to get old cublas math mode: " << ToString(ret);
+      return ok_ = false;
+    }
+
+    ret = wrap::cublasSetMathMode(parent_, handle_, new_mode);
+    if (ret != CUBLAS_STATUS_SUCCESS) {
+      LOG(ERROR) << "failed to set new cublas math mode: " << ToString(ret);
```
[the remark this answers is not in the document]
nluehr: Propagation is already in place as the callers check for a false return value.

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -545,7 +622,13 @@ bool CUDABlas::DoBlasInternalImpl(FuncT cublas_func, Stream *stream,
                                            : CUBLAS_POINTER_MODE_DEVICE)) {
     return false;
   }
-
+#if CUDA_VERSION >= 9000
+  ScopedCublasMathMode math_mode{parent_, blas_};
```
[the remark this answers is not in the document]
nluehr: OK. We'll only call init when tensor-ops are needed.

tensorflow/stream_executor/cuda/cuda_blas.cc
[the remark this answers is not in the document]
nluehr: Done.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: The env-var is only checked on first access. TensorOpMathEnabled() caches the return value in a static variable.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: Added cuda 9.0 algos, including tensor ops, to the algorithm list. Also, added checks to DoBlasGemmWithAlgorithmImpl to return false if tensor_op algorithm is requested when not appropriate. These changes enable the auto-tuning path to use tensor-ops.

nluehr:
Just pushed.

tensorflow/stream_executor/cuda/cuda_blas.h
```diff
@@ -84,21 +84,33 @@ class CUDABlas : public blas::BlasSupport {
...
-                              /*err_on_failure=*/true, args...);
+                              /*err_on_failure=*/true, /*use_tensor_ops=*/false,
+                              args...);
+  }
+  template <typename FuncT, typename... Args>
+  bool DoBlasInternalWithTensorOps(FuncT cublas_func, Stream *stream,
+                                   bool pointer_mode_host, bool use_tensor_ops,
+                                   Args... args) {
+    return DoBlasInternalImpl(cublas_func, stream, pointer_mode_host,
+                              /*err_on_failure=*/true, /*use_tensor_ops=*/false,
```
yzhwang: The name DoBlasInternalWithTensorOps() suggests that use_tensor_ops should be set to true. Also, if we use this function, I think the argument use_tensor_ops is redundant.

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -1762,14 +1846,26 @@ bool CUDABlas::DoBlasGemm(
...
+    use_tensor_ops = true;
+  }
+#endif
+
+  return DoBlasInternalWithTensorOps(
       wrap::cublasSgemmEx, stream, true /* = pointer_mode_host */,
-      CUDABlasTranspose(transa), CUDABlasTranspose(transb), m, n, k, &alpha,
-      CUDAMemory(a), SE_CUDA_DATA_HALF, lda, CUDAMemory(b), SE_CUDA_DATA_HALF,
-      ldb, &beta, CUDAMemoryMutable(c), SE_CUDA_DATA_HALF, ldc);
+      use_tensor_ops, CUDABlasTranspose(transa), CUDABlasTranspose(transb), m,
```
yzhwang: DoBlasInternalWithTensorOps should set use_tensor_ops to be true, I think we should get rid of use_tensor_ops argument.

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -299,6 +305,17 @@ static string ToString(cublasStatus_t status) {
   }
 }
 
+// Decide whether to enable TENSOR_OP_MATH
+static bool TensorOpMathEnabled() {
+  static bool is_enabled = [] {
+    bool ret;
+    TF_CHECK_OK(tensorflow::ReadBoolFromEnvVar("TF_DISABLE_TENSOR_OP_MATH",
```
yzhwang: I noticed that in cuda_dnn.cc, there is also an env-var with exactly the same name: TF_DISABLE_TENSOR_OP_MATH. Ideally, we would want each env-var to only control the behavior of one op, this will make debugging easier. Could you consider using two different names such as TF_DISABLE_CUBLAS_TENSOR_OP_MATH and TF_DISABLE_CUDNN_TENSOR_OP_MATH?

tensorflow/stream_executor/cuda/cuda_blas.cc
yzhwang: I think you'd have to also add this code to DoBlasGemmWithProfiling() and DoBlasGemmWithAlgorithm() for fp16, so that it can also be used when turning on autotune.

tensorflow/stream_executor/cuda/cuda_blas.h
yzhwang: Please also change the interfaces of  DoBlasGemmWithProfiling() and DoBlasGemmWithAlgorithm() for fp16. Thank you!

yzhwang (changes requested):
Thank you for the changes! See my comments below.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: It is convenient to pass this as a flag, otherwise I'll have to do an if-else on the flag in order to make different function calls. Alternatively, I could remove DoBlasInternalWithTensorOps entirely and use direct calls to DoBlasInternalImpl. Would that be preferable?

tensorflow/stream_executor/cuda/cuda_blas.h
nluehr: Tensor op usage is encoded in the algorithm, so no modification of DoBlasGemmWithAlgorithm is needed.
DoBlasGemmWithProfiling wraps DoBlasGemm and also should not require further change as I understand the code.

tensorflow/stream_executor/cuda/cuda_blas.cc
yzhwang: I'm fine with either way. I noticed that later in DoBlasInternalWithTensorOps() you directly called this: return DoBlasInternalImpl(cublas_func, stream, pointer_mode_host, /*err_on_failure=*/true, /*use_tensor_ops=*/false, ...) Shouldn't it be /*use_tensor_ops=*/true?
Also I still think if you are using DoBlasInternalWithTensorOps(), it suggests you are using tensor ops, so probably just get rid of this function?

tensorflow/stream_executor/cuda/cuda_blas.h
yzhwang: Yes you're right, both functions call DoBlasGemm() so nothing should be done in those two functions. Thanks for the correction.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: TF_DISABLE_TENSOR_OP_MATH continues to disable tensor ops in both cublas and cudnn, but I added TF_DISABLE_CUBLAS_TENSOR_OP_MATH and TF_DISABLE_CUDNN_TENSOR_OP_MATH which will override the above and apply specifically to CUBLAS or CUDNN.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: Yes, that was a bug. Thanks for the catch. DoBlasGemm for fp16 now calls  DoBlasInternalImpl directly and DoBlasInternalWithTensorOps has been removed.

tensorflow/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -560,8 +560,13 @@ class ScopedFilterDescriptor {
...
-    TF_CHECK_OK(tensorflow::ReadBoolFromEnvVar("TF_DISABLE_TENSOR_OP_MATH",
-                                               /*default=*/false, &ret));
+    if (getenv("TF_DISABLE_CUDNN_TENSOR_OP_MATH") != nullptr) {
+      TF_CHECK_OK(tensorflow::ReadBoolFromEnvVar("TF_DISABLE_CUDNN_TENSOR_OP_MATH",
+                                                 /*default=*/false, &ret));
+    } else {
+      TF_CHECK_OK(tensorflow::ReadBoolFromEnvVar("TF_DISABLE_TENSOR_OP_MATH",
+                                                 /*default=*/false, &ret));
+    }
     return ret;
```
yzhwang: Should this also be !ret like in cuda_blas.cc?

yzhwang (commented):
Please also add description that you have changed GetBlasGemmAlgorithms() to include more internal algorithms for CUDA_VERSION >= 9000.

nluehr:
Done.

yzhwang (approved):
LGTM.

sb2nov:
@zheng-xq PTAL

nluehr:
Ping @zheng-xq. Are there additional changes needed?

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -299,6 +305,24 @@ static string ToString(cublasStatus_t status) {
   }
 }
 
+// Decide whether to enable TENSOR_OP_MATH
+static bool TensorOpMathEnabled() {
+  static bool is_enabled = [] {
```
zheng-xq: What's the point of having a lambda and evaluate it right away?

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -299,6 +305,24 @@ static string ToString(cublasStatus_t status) {
...
+    bool is_disabled;
+    if (getenv("TF_DISABLE_CUBLAS_TENSOR_OP_MATH") != nullptr) {
+      TF_CHECK_OK(
+          tensorflow::ReadBoolFromEnvVar("TF_DISABLE_CUBLAS_TENSOR_OP_MATH",
+                                         /*default=*/false, &is_disabled));
+    } else {
+      TF_CHECK_OK(tensorflow::ReadBoolFromEnvVar("TF_DISABLE_TENSOR_OP_MATH",
```
zheng-xq: In general, I'm not a fan of hierarchical flags, since their relationship is not clearly documented. If the name is confusing, why not just change the other one into "TF_DISABLE_CUDNN_TENSOR_OP_MATH"?

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -2049,6 +2152,15 @@ bool CUDABlas::DoBlasGemmWithAlgorithmImpl(
     return false;
   }
 
+#if CUDA_VERSION >= 9000
+  cublasGemmAlgo_t cublas_algo = static_cast<cublasGemmAlgo_t>(algorithm);
+  if (cublas_algo >= CUBLAS_GEMM_DFALT_TENSOR_OP &&
```
zheng-xq: Why not using "CUBLAS_GEMM_DEFAULT_TENSOR_OP".

Also I'm not a fan of comparing enums. Could you split "cublas_algo >= CUBLAS_GEMM_DFALT_TENSOR_OP" into a function? In fact, I'd prefer to return them one by one... So future changes have to update them.

tensorflow/stream_executor/cuda/cuda_blas.cc
```diff
@@ -2098,10 +2210,19 @@ bool CUDABlas::GetBlasGemmAlgorithms(
 // still return the out_algorithms. Caller needs to make sure that in this case,
 // the returned vector is empty.
 #if CUDA_VERSION >= 8000
-  for (cublasGemmAlgo_t algo :
-       {CUBLAS_GEMM_DFALT, CUBLAS_GEMM_ALGO0, CUBLAS_GEMM_ALGO1,
-        CUBLAS_GEMM_ALGO2, CUBLAS_GEMM_ALGO3, CUBLAS_GEMM_ALGO4,
-        CUBLAS_GEMM_ALGO5, CUBLAS_GEMM_ALGO6, CUBLAS_GEMM_ALGO7}) {
+  for (cublasGemmAlgo_t algo : {
+         CUBLAS_GEMM_DFALT, CUBLAS_GEMM_ALGO0, CUBLAS_GEMM_ALGO1,
```
zheng-xq: Could you split the list into its own function? So we know to update them, or even auto-gen them from the header file.

zheng-xq (commented):
A few more comments

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: This allows the static is_enabled to be set inline.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: Removed TF_DISABLE_TENSOR_OP_MATH.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: Changed to CUBLAS_GEMM_DEFAULT_TENSOR_OP and moved inequality check to a separate function.

tensorflow/stream_executor/cuda/cuda_blas.cc
nluehr: I'm not sure what is being suggested here. The current function seems pretty clean to me. Do you envision separate functions for CUBLAS 8 vs. 9 algos, or separate function for tensor op vs. non tensor op algos. Either of these would make the API higher up the stack more complicated.

zheng-xq:
This looks okay for now. Let's merge it in and improve it later. Thanks for the contribution!

zheng-xq:
LGTM

zheng-xq (approved):

drpngx:
Jenkins, test this please.

PR #13453, opened 2017-10-02
sampled version of sparse_softmax_cross_entropy_with_logits

TTrapper:
Following from the already implemented `sampled_softmax_loss`, this is a very similar implementation that performs sampling for `sparse_softmax_cross_entropy_with_logits`.


frankchn:
Jenkins, test this please.

drpngx:
@ebrevdo WDYT? Should we do this in contrib first?

TTrapper:
Hi folks, thank you for looking at this request. In regards to the test failures:

- **Linux CPU Tests:** `api_compatibility_test`
I take it this is to be expected given that the API would change. If these changes are accepted, how do I go about updating the goldens?

- **Linux CPU Tests (Python 3):** ` do_virtualenv_pip_test`
Not sure about this one, but I can investigate tomorrow.

tensorflow/python/ops/nn_impl.py
```diff
@@ -1258,3 +1258,101 @@ def sampled_softmax_loss(weights,
...
+        biases=biases,
+        labels=labels,
+        inputs=inputs,
+        ...,
+        partition_strategy="div")
+  elif mode == "eval":
+    logits = tf.matmul(inputs, tf.transpose(weights))
+    logits = tf.nn.bias_add(logits, biases)
+    labels_one_hot = tf.one_hot(labels, n_classes)
+    loss = tf.nn.sparse_softmax_cross_entropy_with_logits(
```
ebrevdo: The labels to this function are not one hot.

tensorflow/python/ops/nn_impl.py
```diff
@@ -1258,3 +1258,101 @@ def sampled_softmax_loss(weights,
...
+  ```
+
+  See our [Candidate Sampling Algorithms Reference]
+  (https://www.tensorflow.org/extras/candidate_sampling.pdf)
+
+  Also see Section 3 of [Jean et al., 2014](http://arxiv.org/abs/1412.2007)
+  ([pdf](http://arxiv.org/pdf/1412.2007.pdf)) for the math.
+
+  Args:
+    weights: A `Tensor` of shape `[num_classes, dim]`, or a list of `Tensor`
```
ebrevdo: Tensor or variable?

tensorflow/python/ops/nn_impl.py
```diff
@@ -1258,3 +1258,101 @@ def sampled_softmax_loss(weights,
...
+      num_sampled=num_sampled,
+      num_classes=num_classes,
+      num_true=1,
+      sampled_values=sampled_values,
+      subtract_log_q=True,
+      remove_accidental_hits=remove_accidental_hits,
+      partition_strategy=partition_strategy,
+      name=name)
+
+  # labels returned by _compute_sampled_logits are one_hot. Convert to indices.
```
ebrevdo: Seems wasteful, perhaps that function can be modified to optionally return labels as integers.

ebrevdo (changes requested):
This is a reasonable candidate for tf.contrib.nn, but not core until it gets some use.

tensorflow/python/ops/nn_impl.py
TTrapper: Thank you. Corrected.

tensorflow/python/ops/nn_impl.py
TTrapper: Certainly a Variable, but this is the terminology used in the documentation of similar functions (eg `sampled_softmax_loss`). I take it that the term Tensor is in inclusive of variable Tensors? Let me know if you would still like it changed.

tensorflow/python/ops/nn_impl.py
TTrapper: OK, I added an option to do just that.

TTrapper (commented):
Requested changes have been implemented/addressed.

TTrapper:
@frankchn I am new to this. Is there something I should do to have the `stat: awaiting response` label removed? Thanks!

frankchn:
Jenkins, test this please.

frankchn:
@ebrevdo PTAL again. Thanks!

tensorflow/python/ops/nn_impl.py
```diff
@@ -1046,13 +1054,19 @@ class biases.
...
-    # of ones. We then divide by num_true to ensure the per-example labels sum
-    # to 1.0, i.e. form a proper probability distribution.
-    out_labels = array_ops.concat([
-        array_ops.ones_like(true_logits) / num_true,
-        array_ops.zeros_like(sampled_logits)
-    ], 1)
+    if labels_as_indices:
+      # We want each row of labels to be the indices of the targets, which
+      # start at col 0 and end at col num_true-1.
+      out_labels = gen_array_ops.tile(
```
ebrevdo: i guess this should only be used for num_true == 1, right?  can you add a check and corresponding documentation?

alternatively you could just move this code to sparse_sampled_softmax_loss and revert this.

ebrevdo (dismissed):

asimshankar:
(Removing API review label as this is in `contrib`)

TTrapper (commented):
@ebrevdo just checking in. I'm not sure whether my above comment triggers the necessary notifications to get the ball rolling again.

tensorflow/python/ops/nn_impl.py
TTrapper: @drpngx just checking in because this has been sitting for some time. I responded to @ebrevdo 's review with the above comment. Thanks!

drpngx:
Thanks @TTrapper ! Let's see what @ebrevdo has to say.

ebrevdo:
If the special case code is only used by the sampled function, perhaps it's better to keep it in that function only?

TTrapper:
Thanks @ebrevdo. It's general enough to be potentially useful to others. I'm also weary of having code elsewhere that assumes `out_logits == concat[true_logits, sampled_logits]`. If that ever changes, then it is nice if changes to the `out_labels` are all in the same place as well.  Let me know if you agree.

TTrapper:
Hi @drpngx, @ebrevdo. Just checking in on my above comment. Thanks!

drpngx:
@TTrapper I think the recommendation was to move to contrib. That will still be available to everyone.

TTrapper:
@drpngx Thanks, I think @ebrevdo 's concern was that the proposed `sample_sparse_softmax_loss` was the only place where the special case logic for getting true indices was used. I have now removed that special case logic from `_compute_sampled_logits` and added the specific use case directly to `sampled_sparse_softmax_loss`.

drpngx:
I see.

Jenkins, test this please.

martinwicke (approved):

PR #13465, opened 2017-10-03
Remove unnecessary specification for default kernel name

yakigac:
This line doesn't effect in any cases.

The docker container for latest-py2 is working well even without this line.
Although it's written `= 'python2'`, the docker container for latest-py3 is working.

`parameterized_docker_build.sh` or other scripts are not overwriting the line.
It should be removed, or overwritten when the images are built.


frankchn:
Jenkins, test this please.

frankchn (approved):

PR #13483, opened 2017-10-04
Fix tf-signal tests on pip packages.

gunan:


frankchn (approved):

frankchn:
Jenkins, test this please.

rryan:
Sorry about this! Is there any way I should have noticed this? I'm pretty sure internally the pip tests were passing when I submitted.

gunan:
I think one way was pip_smoke_tests, but we need a better way to explain the situation in the presubmit tests.

PR #13495, opened 2017-10-04
Remove unneeded branch check

### LINKED ISSUE #6911: Switching branches forces ./configure and bazel rebuild
Right now switching branches requires running ./configure which then makes next `bazel test` rebuild everything.

In particular on 32 core Xeon, switching branches/`./configure` makes `bazel test learning/python/...` run 17 minutes longer because it's building stuff from scratch. This makes it hard for open-source contributors to work on more than one Pull Request at a time.

In particular, the error from `bazel test` when switching git branches is as follows:

```
ERROR: /home/yaroslav/tensorflow.git/tensorflow/tensorflow/core/BUILD:1194:1: Executing genrule //tensorflow/core:version_info_gen failed: bash failed:\
 error executing command /bin/bash -c ... (remaining 1 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited\
 with status 1.
  File "tensorflow/tools/git/gen_git_source.py", line 260, in <module>
    generate(args.generate)
  File "tensorflow/tools/git/gen_git_source.py", line 211, in generate
    (old_branch, new_branch))
RuntimeError: Run ./configure again, branch was 'refs/heads/match-once-fix' but is now 'refs/heads/symbolic-gradient-fix'
_
```

yaroslavvb:


yaroslavvb:
fixes #6911

frankchn:
Jenkins, test this please.

Timmmm:
Is there a reason this can't be merged, or has it just slipped through the net? It's a pretty annoying problem.

yaroslavvb:
ping @aselle

aselle (changes requested):
This is actually a *needed* branch check for the git version to be accurate at all. Without the check, nothing will force bazel to rebuild the version value. That means the reported version will be incorrect, and you will have no way to determine if this is incorrect. 

That being said, I get the point of not wanting to do a whole rebuild just to get the correct sha recorded when you are swapping between branches.

So, a reasonable compromise would be to do this branch check, and if it fails, return "inconsistent-git-version" as the version number or something, so at least you know that the value is garbage. (The whole point of this was to help with issue triage).

The bazel team pointed out to us that there is now a mechanism to do all of this without the symlink hackery by using --workspace_status_command. So if you are feeling especially ambitous, you could move to using that. I'd say, let's get this change in first though.

Timmmm:
I haven't really delved into the whole Bazel stamping thing yet, but shouldn't the git hash be a dependency of only a small part of the whole project. E.g. to rename the final artefacts or maybe to put in a `version.h` header that isn't `#include`d anywhere. I don't see why the entire project would need to depend on the git hash.

yaroslavvb:
@Timmmm updating GIT_VERSION just needs relinking pywrap_tensorflow.so which is fast, however, it's implemented through ./configure which touches touches a bunch of dependencies as a by-product, so building becomes slow.

@aselle PTAL

Timmmm:
So, I had a look, and as far as I understand the process, here is how it currently works:

In `tensorflow/tools/git/gen_git_source.py` there is a function `generate()` which writes `version_info.cc` containing this function: `const char* tf_git_version() {return "%s";}` where `%s` is the result of `git describe --long --tags`, e.g. `v1.3.0-rc1-4901-gd4c5301d52`. `generate()` is called when you run `python gen_git_source.py --generate`.

This is called by a [genrule](https://docs.bazel.build/versions/master/be/general.html#genrule) defined by `tf_version_info_genrule()` in `tensorflow.bzl` which is called in `core/BUILD`. The genrule has these sources:

```
      srcs=[
          clean_dep("//tensorflow/tools/git:gen/spec.json"),
          clean_dep("//tensorflow/tools/git:gen/head"),
          clean_dep("//tensorflow/tools/git:gen/branch_ref"),
      ],
```

`head` is symlinked to `.git/HEAD` which I guess means it is run every time the head or branch changes. There is a function in `configure.py` that calls `gen_git_source.py --configure` which writes `spec.json` and symlinks `head` and `branch_ref` to `.git/HEAD` and the current branch ref in `.git`. `spec.json` contains a flag to say if `git` is present, the name of the current branch and the full path to the repo.

So here's what I think happens:

1. You `git clone tensorflow`
2. You run `./configure`. This calls `gen_git_source.py --configure` which symlinks `gen/head` to `.git/HEAD`, and generates `spec.json` which contains the current branch name.
3. You run a Bazel command. Some target somewhere depends on `version_info.cc` which is generated by the `tf_version_info_genrule()` genrule. This genrule calls `gen_git_source.py --generate` which checks that the current branch is the same as the one in `spec.json`. If so it writes out `version_info.cc` using `git describe` and everyone is happy.
4. You switch branches. This causes `gen/head` to change.
5. You run a Bazel command. Since `gen/head` changed, `gen_git_source.py --generate` is called which checks the current branch vs `spec.json` and finds that it has changed! It makes you run `./configure` again.

This seems silly. I can't even see that the branch is used anywhere! I think a simpler design would be like this:

1. Remove the current branch from `spec.json`.
2. Remove the branch check from `generate()`.
3. Get rid of `gen/branch_ref`.
3. That's it.

If head changes, then `//tensorflow/tools/git:gen/head` should change, and `gen_git_source.py --generate` will be called again by the genrule and write a new `version_info.cc`. Nothing needs to know the branch name.

By the way as far as I can tell none of this will work on Windows anyway because it copies files rather than creating symlinks.

Let me know if I have totally misunderstood things!

martinwicke:
@aselle can you take another look?

aselle:
@Timmmm, your analysis is partially true. You need to track both changes of branches and commit changes to that branch. the gen/branch_ref tracks commits to the branch, and the gen/head tracks changes to what the current branch is. So if you get rid of gen/branch_ref, the git sha will not change when you commit a new change.

---workspace_status_command is the right way to go in any case.

aselle (approved):

Timmmm:
Ah yes good point - I was thinking that `.git/HEAD` contains the hash of HEAD but actually it contains something like `ref: refs/heads/master` and the hash is in `.git/refs/heads/master`.

yifeif:

drpngx:
Framework issue on the Sanity check? Retrying.

Jenkins, test this please.

drpngx:
Py3: flaky test :-(

```
FAIL: //tensorflow/core/platform/cloud:file_block_cache_test (see /tmpfs/tmp/bazel/execroot/org_tensorflow/bazel-out/k8-py3-opt/testlogs/tensorflow/core/platform/cloud/file_block_cache_test/test.log)
INFO: From Testing //tensorflow/core/platform/cloud:file_block_cache_test:
==================== Test output for //tensorflow/core/platform/cloud:file_block_cache_test:
Running test tensorflow/core/platform/cloud/file_block_cache_test on GPU 2
Running main() from test_main.cc
[==========] Running 11 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 11 tests from FileBlockCacheTest
[ RUN      ] FileBlockCacheTest.PassThrough
[       OK ] FileBlockCacheTest.PassThrough (0 ms)
[ RUN      ] FileBlockCacheTest.BlockAlignment
[       OK ] FileBlockCacheTest.BlockAlignment (0 ms)
[ RUN      ] FileBlockCacheTest.CacheHits
[       OK ] FileBlockCacheTest.CacheHits (1 ms)
[ RUN      ] FileBlockCacheTest.OutOfRange
[       OK ] FileBlockCacheTest.OutOfRange (0 ms)
[ RUN      ] FileBlockCacheTest.Inconsistent
[       OK ] FileBlockCacheTest.Inconsistent (0 ms)
[ RUN      ] FileBlockCacheTest.LRU
[       OK ] FileBlockCacheTest.LRU (0 ms)
[ RUN      ] FileBlockCacheTest.MaxStaleness
[       OK ] FileBlockCacheTest.MaxStaleness (0 ms)
[ RUN      ] FileBlockCacheTest.RemoveFile
[       OK ] FileBlockCacheTest.RemoveFile (0 ms)
[ RUN      ] FileBlockCacheTest.Prune
[       OK ] FileBlockCacheTest.Prune (2002 ms)
[ RUN      ] FileBlockCacheTest.ParallelReads
[       OK ] FileBlockCacheTest.ParallelReads (13 ms)
[ RUN      ] FileBlockCacheTest.CoalesceConcurrentReads
tensorflow/core/platform/cloud/file_block_cache_test.cc:490: Failure
Value of: WaitForNotificationWithTimeout(&notification, 10000)
  Actual: false
Expected: true
Timeout waiting for concurrent thread to start.
[  FAILED  ] FileBlockCacheTest.CoalesceConcurrentReads (111 ms)
[----------] 11 tests from FileBlockCacheTest (2128 ms total)

[----------] Global test environment tear-down
[==========] 11 tests from 1 test case ran. (2128 ms total)
[  PASSED  ] 10 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] FileBlockCacheTest.CoalesceConcurrentReads

 1 FAILED TEST
[10,621 / 10,624] 401 / 403 tests, 1 failed; Testing //tensorflow/core:common_runtime_gpu_gpu_bfc_allocator_test; 19s local ... (3 actions running)
INFO: Elapsed time: 1059.053s, Critical Path: 312.29s
INFO: Build completed, 1 test FAILED, 10624 total actions
```

GPU: test timeout `TIMEOUT: //tensorflow/python/keras:data_utils_test (Summary)`

PR #13496, opened 2017-10-05
Add tf.sysconfig.get_compile_flags() & tf.sysconfig.get_link_flags() for custom operators

alsrgv:
The goal is to make custom operators compilation a breeze.

```
In [1]: import tensorflow as tf
Couldn't import dot_parser, loading of dot files will not be possible.

In [2]: tf.sysconfig.get_compile_flags()
Out[2]:
['-I/home/devadmin/env/lib/python2.7/site-packages/tensorflow/include',
 '-I/home/devadmin/env/lib/python2.7/site-packages/tensorflow/include/external/nsync/public',
 '-D_GLIBCXX_USE_CXX11_ABI=1']

In [3]: tf.sysconfig.get_link_flags()
Out[3]:
['-L/home/devadmin/env/lib/python2.7/site-packages/tensorflow',
 '-ltensorflow_framework']
```


alsrgv:
Please advise if `version_info.cc` is a good place to probe `_GLIBCXX_USE_CXX11_ABI`, whether this level of exposure is OK and what's the best place to add tests.

frankchn:
Jenkins, test this please.

alsrgv:
Looks like test failure caught new functions in `tf.sysconfig`, which is expected.

```
You will need an API approval to make changes to the public TensorFlow API. This
includes additions to the API.

ERROR:tensorflow:2 differences found between API and golden.
ERROR:tensorflow:Issue 1	: Change detected in python object: tensorflow.
ERROR:tensorflow:Issue 2	: Change detected in python object: tensorflow.sysconfig.
```

alsrgv:
+ @jhseu

tensorflow/core/public/version.h
```diff
@@ -116,5 +116,7 @@ extern const char* tf_compiler_version();
 // The git commit designator when tensorflow was built
 // If no git repository, this will be "internal".
 extern const char* tf_git_version();
+// Value of the _GLIBCXX_USE_CXX11_ABI flag, or -1 if it's not set.
+extern const int tf_cxx11_abi();
```
drpngx: I would rename everything to `_flag`: `tf_cxx11_abi_flag`, `tf.sysconfig.CXX11_ABI_FLAG`

alsrgv:
Thanks for the feedback, @drpngx.  Updated.

drpngx:
Jenkins, test this please.

tensorflow/core/public/version.h
```diff
@@ -116,5 +116,7 @@ extern const char* tf_compiler_version();
 // The git commit designator when tensorflow was built
 // If no git repository, this will be "internal".
 extern const char* tf_git_version();
+// Value of the _GLIBCXX_USE_CXX11_ABI flag, or -1 if it's not set.
```
drpngx: Any reason it's not 0 when not set?

alsrgv: Allows to skip it in `tf.sysconfig.get_compile_flag()` if it's not defined in the system.  If you'd rather it always be there and set to 0 by default, I can change.

asimshankar:
Okay for API change, @drpngx : deferring to you to complete the review and approve.

tensorflow/python/platform/sysconfig.py
```diff
@@ -51,5 +53,32 @@ def get_lib():
...
   return _os_path.join(_os_path.dirname(tf.__file__))
 
+
+def get_compile_flags():
+  """Get the compilation flags for custom operators.
+
+  Returns:
+    The compilation flags.
+  """
+  import tensorflow as tf
```
drpngx: For this, I think you should import versions.py direction from the top.

tensorflow/python/platform/sysconfig.py
```diff
@@ -51,5 +53,32 @@ def get_lib():
...
+  flags = []
+  flags.append('-I%s' % get_include())
```
drpngx: These are gcc flags. It's not clear what to do on windows, since you can use MSVC and gcc-based compilers. I think we should assume MSVC-style flags on windows.

drpngx (changes requested):

tensorflow/python/platform/sysconfig.py
alsrgv: Can you clarify what do you mean?

tensorflow/python/platform/sysconfig.py
alsrgv: Is it OK to just check `platform.system() == 'Windows'`, and if so output MSVC-style flags?

tensorflow/python/platform/sysconfig.py
drpngx: What I meant to say is that you don't have to import tf here. I know it's done above but it's a special case. You can say `from tensorflow.python.framework.version import ...`

tensorflow/python/platform/sysconfig.py
drpngx: So, I think that will do for now.

tensorflow/core/public/version.h
drpngx: OK, sorry, now I get your point. Let's leave it to `-1`, then.

drpngx (changes requested):

drpngx:
/CC @gunan @jart FYI. This is like `pkgconfig`. While it will be hard to maintain, I think it's still useful. Normally we might have a standalone tool, but it make the packaging harder.

tensorflow/python/platform/sysconfig.py
alsrgv: Gotcha.  I think includes (/I) are pretty straightforward.  Is there any instruction that specifies which link options are necessary for Windows?  https://www.tensorflow.org/extend/adding_an_op#compile_the_op_using_your_system_compiler_tensorflow_binary_installation port for Windows.

tensorflow/python/platform/sysconfig.py
```diff
@@ -51,5 +53,32 @@ def get_lib():
...
   return _os_path.join(_os_path.dirname(tf.__file__))
 
+
+def get_compile_flags():
+  """Get the compilation flags for custom operators.
+
+  Returns:
+    The compilation flags.
+  """
+  from tensorflow.python.framework.versions import __cxx11_abi_flag__
```
alsrgv: @drpngx is this what you had in mind?

I've tried to put import on top of the module, too - but that causes __cxx11_abi_flag__ to pollute the namespace.

drpngx:
Jenkins, test this please.

tensorflow/python/platform/sysconfig.py
drpngx: You can say `import .. CXX_ABI_FLAG as _CXX_ABI_FLAG` on top.

alsrgv: If I do that, it's still there:

>>> dir (tf.sysconfig)
['_CXX11_ABI_FLAG', '__builtins__', '__doc__', '__file__', '__name__', '__package__', '_allowed_symbols', '_os_path', 'get_compile_flags', 'get_include', 'get_lib', 'get_link_flags']

drpngx: Yes, that's fine.

alsrgv: Ack, updated.

gunan:
AFAICR, @m3bm3b had a plan to create a makefile for custom ops, which we would package with the pip package.

alsrgv:
@gunan would it work if the custom op has dependencies on other libraries, such as NCCL, CUDA or MPI? MPI typically comes with "compilation flags" and "linking flags". Also, it is more challenging to run `make` than `gcc`/`clang` from `pip install`.

gunan:
You are right, the custom dependencies for custom ops can be confusing.
I think ultimately, we would move to distributing TF using RPM/DEB packages and also provide tools maybe not makefile but some tools to build custom ops easier.
But until then, this may work.

Its existence in the core API, I dont know.
Ill let API owners to decide on that.

alsrgv:
@drpngx, gentle ping.

drpngx (approved):

drpngx:
Sorry about the delay.

alsrgv:
@drpngx, thanks!

alsrgv:
Did the tests pass?  I see two failures:
1. `//tensorflow/tools/api/tests:api_compatibility_test` failed, which is expected.
2. `//tensorflow/contrib/data/python/kernel_tests:sloppy_transformation_dataset_op_test` failed, and I can't see any logs - I don't believe it could be related to this change.

gunan:
You need to update goldens and package them in this Pull Request.
Currently api compatibility test only checks the API status in python2 and linux setup, you may use docker to run api compatibility test and generate new goldens.

The other test failure should go away if you rebase your change on top of master.

alsrgv:
@gunan, got you. Updated.

gunan:
Jenkins, test this please.

alsrgv:
@drpngx, @gunan - thanks!

alsrgv:
Next step is to update master version of https://www.tensorflow.org/versions/r1.4/extend/adding_an_op#build_the_op_library.  Is it hosted in a public repository?  If so, I could help to do the change.

alsrgv:
Sorry guys, have to follow this up with a bugfix #14159.

After I tested this nightly build with gcc4 & 5, I realized that gcc4 -> gcc5 always requires ABI flag to be present with 0 value, even though it was undefined for gcc4.

This somewhat echoes @drpngx question whether we could have the default value of `_GLIBCXX_USE_CXX11_ABI` be 0, but we need to add that flag to complication flags even if it is 0.

Issue #13498, opened 2017-10-05
Problem with "AddControlInput" in python_api.h


eaplatanios:
@asimshankar @skye I'm observing an issue with `AddControlInput` in `python_api.h`. It seems to not be working for me actually and I'm not sure why. Everything is ok if I add the control dependency during op creation. However, if I add it right after, it's not enforced during execution. A simple example is creating a `switch` op, creating two constant ops and adding one control dependency for each on each switch output, and then feeding these two constant ops into a `merge`op. The result will be whichever constant op was fed first into the merge op (i.e., the control dependencies are not satisfied and both ops are executed). If I add the control dependency during op construction, all is good.

Note that the control input does show up in the `GraphDef` that I generated after the call. It's simply not enforced during execution.

Do you have any idea why that might be happening? I am aware that this is not part of the public API and thus is not stable, but it is still unexpected behavior. I am using that in my implementation of control flow ops and gradients (I ended up re-implementing it based on the Python version because I depend on it and the C++ functionality is currently very limited -- I'll update on this later on, once I release it).

eaplatanios:
FYI, if you want to replicate that behavior using the Python API, you can do the following:

1. Set `_USE_C_API` in `python/framework/ops.py` to `True`.
2. Move this (in `python/framework/ops.py`):
    ```python
    # Add this op to the current control flow context:
    self._control_flow_context = g._get_control_flow_context()  # pylint: disable=protected-access
    if self._control_flow_context is not None:
      # TODO(skyewm): consider refactoring this to call self._create_c_op()
      # first. This would require updating the TF_Operation's ID (see the
      # comment and self._id_value update below). The disadvantage of calling
      # AddOp() first is that we need to maintain Operation state that is
      # accessed by AddOp() in Python, e.g. the input Tensors.
      self._control_flow_context.AddOp(self)
    ```
    after the `_c_op` initialization.
3. Try running the following simple script:
    ```python
    p = tf.constant(True)
    t = lambda: tf.constant(True)
    f = lambda: tf.constant(False)
    r = tf.cond(p, t, f)
    print(r.eval(session=tf.Session()))
    ```

This will print `False`, even though it should be printing `True`. The reason is that the false branch output is fed into the merge op first.

Also CCing @itsmeolivia since I think she's been working on making the Python API to use the C API.

eaplatanios:
[...]

PR #13505, opened 2017-10-05
Fix for #13498

eaplatanios:
@skye I'm kind of indecisive it turns out. :P This is just your fix along with an equivalent fix for `RemoveEdge`. I don't know if you want to merge it given that you might make changes to your fix, but I'm putting here just in case. The `RemoveEdge` fix is pretty much equivalent and since the method is not used in the graph constructor, I didn't change its call sites to use `false` for the update node def argument. I'm not sure if that collides with anything else, but I guess the tests can tell us something (I didn't run them on my laptop because they take forever and I can't use it in the meantime).


skye:
Let's hold off on this until my patch is actually submitted :)

eaplatanios:
Sounds good! By the way, I just pushed a small update for `RemoveEdge`, because the previous version caused some problems.

caisq:

skye:
@eaplatanios you need to rebase. The AddControlEdge API changed during code review: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/graph/graph.h#L446

eaplatanios:
@skye I think I resolved this earlier today. Are there any changes I omitted?

skye:
Sorry I deleted that previous comment, I was somehow looking at the wrong thing.

tensorflow/core/graph/graph.cc
```diff
@@ -453,6 +453,20 @@ const Edge* Graph::AddControlEdge(Node* source, Node* dest,
   return AddEdge(source, kControlSlot, dest, kControlSlot);
 }
 
+void Graph::RemoveControlEdge(const Edge* e, bool update_node_def) {
```
skye: Do you have need for update_node_def=false? If not, let's just remove it. The contract with AddControLEdge() is that it will make sure the NodeDef is in sync with the added control edge (unless allow_duplicates=true, which is only in one weird specific case that we would ideally change). So we can assume that we should continue to keep the NodeDef in sync here.

tensorflow/core/graph/graph_constructor.cc
```diff
@@ -840,7 +840,7 @@ Status GraphConstructor::Convert() {
         back_edges_.push_back(
             EdgeInfo(inputs[i].name, inputs[i].index, node, i));
       } else if (inputs[i].index == Graph::kControlSlot) {
-        g_->AddControlEdge(inputs[i].node, node);
+        g_->AddControlEdge(inputs[i].node, node, /*update_node_def=*/false);
```
skye: Do you need to update this file? Also unit tests for your new method?

skye (changes requested):

eaplatanios:
No worries. It was actually super well timed. I like pushed the changes a few minutes before you sent me that. :)

tensorflow/core/graph/graph.cc
eaplatanios: Yeah I only added that option in order to be consistent with your previous change because I was unsure of where else it might be necessary to not keep the NodeDef in sync. I'll remove that.

tensorflow/core/graph/graph_constructor.cc
eaplatanios: These were just changes you had in your early draft implementation and I copied them over. I'll remove them and check that everything works fine. Also, I'll add tests for the new method once I get the chance this week. :)

eaplatanios:
@skye I added a simple test for the new method. I'm not sure what your convention is for unit tests. Please let me know if you want me to test for some specific condition other than the simple case I added. :)

tensorflow/core/graph/graph_test.cc
```diff
@@ -477,6 +491,44 @@ TEST_F(GraphTest, AddControlEdge) {
...
+  Node* b = FindNode("B");
+  Node* c = FindNode("C");
+
+  // Add a control edge.
+  const Edge* edge_1 = graph_.AddControlEdge(c, a);
+  const Edge* edge_2 = graph_.AddControlEdge(a, b);
+  ASSERT_TRUE(edge_1 != nullptr);
+  ASSERT_TRUE(edge_2 != nullptr);
+
+  const Edge* edge_1_obtained = FindInputControlEdge(a, edge_1);
```
skye: You can remove L509-521, as this is testing AddControlEdge, which is already tested above. I would then change FindInputControlEdge to "bool ControlEdgeExists()".

tensorflow/core/graph/graph_test.cc
```diff
@@ -477,6 +491,44 @@ TEST_F(GraphTest, AddControlEdge) {
...
+
+  const Edge* edge_2_obtained = FindInputControlEdge(b, edge_2);
+  ASSERT_TRUE(edge_2_obtained != nullptr);
+  EXPECT_EQ(edge_2_obtained->src(), edge_2->src());
+  EXPECT_EQ(edge_2_obtained->src_output(), Graph::kControlSlot);
+  EXPECT_EQ(edge_2_obtained->dst(), edge_2->dst());
+  EXPECT_EQ(edge_2_obtained->dst_input(), Graph::kControlSlot);
+
+  graph_.RemoveControlEdge(edge_1);
+  ASSERT_TRUE(FindInputControlEdge(a, edge_1) == nullptr);
```
skye: Here or possibly in the helper function you should check that the NodeDef doens't contain the control input either.

tensorflow/core/graph/graph_test.cc
```diff
@@ -477,6 +491,44 @@ TEST_F(GraphTest, AddControlEdge) {
...
+  ASSERT_TRUE(FindInputControlEdge(b, edge_2) != nullptr);
+
+  graph_.RemoveControlEdge(edge_2);
+  ASSERT_TRUE(FindInputControlEdge(a, edge_1) == nullptr);
+  ASSERT_TRUE(FindInputControlEdge(b, edge_2) == nullptr);
```
skye: Test removing a duplicate control edge

skye (changes requested):

eaplatanios:
@skye I made all of the suggested changes. :)

tensorflow/core/graph/graph.cc
```diff
@@ -453,6 +453,20 @@ const Edge* Graph::AddControlEdge(Node* source, Node* dest,
   return AddEdge(source, kControlSlot, dest, kControlSlot);
 }
 
+void Graph::RemoveControlEdge(const Edge* e) {
+  if (!e->src_->IsSource() && !e->dst_->IsSink()) {
+    e->dst_->MaybeCopyOnWrite();
+    auto* inputs = e->dst_->props_->node_def.mutable_input();
+    for (auto it = inputs->begin(); it != inputs->end(); ++it) {
+      if (*it == strings::StrCat("^", e->src_->name())) {
```
skye: nit: pull the strings::StrCat out of the for loop so you're not computing it every time (you could even not construct a string at all, but it's fine)

tensorflow/core/graph/graph.h
```diff
@@ -451,6 +451,10 @@ class Graph {
   // REQUIRES: The edge must exist.
   void RemoveEdge(const Edge* edge);
 
+  // Removes control edge `edge` from the graph.
```
skye: Add that this updates the node def

tensorflow/core/graph/graph_test.cc
```diff
@@ -118,6 +118,24 @@ class GraphTest : public ::testing::Test {
     LOG(FATAL) << name;
   }
 
+  bool ControlEdgeExistsInGraphOrNodeDef(const Node* src,
```
skye: This logic isn't quite right... you want to test that both the Edge* and the NodeDef input were removed, not either. Maybe divide this into two functions, or change it to "void ExceptControlEdgeRemoved" and it does both checks, or something.

tensorflow/core/graph/graph_test.cc
```diff
@@ -477,6 +495,42 @@ TEST_F(GraphTest, AddControlEdge) {
...
+  ASSERT_TRUE(!ControlEdgeExistsInGraphOrNodeDef(c, a));
+  ASSERT_TRUE(ControlEdgeExistsInGraphOrNodeDef(a, b));
+
+  graph_.RemoveControlEdge(edge_2);
+  ASSERT_TRUE(!ControlEdgeExistsInGraphOrNodeDef(c, a));
+  ASSERT_TRUE(!ControlEdgeExistsInGraphOrNodeDef(a, b));
+
+  // Test removing a duplicate control edge.
+  const Edge* edge_3 = graph_.AddControlEdge(c, a);
+  const Edge* edge_4 = graph_.AddControlEdge(c, a);
```
skye: Sorry, I was unclear about this... this won't actually add a duplicate edge unless you pass "allow_duplicates" here. This will be kind of a weird test case (I don't actually expect anyone to use allow_duplicates=true along with RemoveControlEdge), but I think it will be good to at least document the behavior in this test.

skye (changes requested):

tensorflow/core/graph/graph_test.cc
eaplatanios: This checks if it exists in either the graph or the NodeDef, meaning that it returns `false` only if it's been removed from both. Isn't that what we want?

tensorflow/core/graph/graph_test.cc
eaplatanios: Good point. I'll add a comment for this.

tensorflow/core/graph/graph.cc
eaplatanios: Good catch. :)

eaplatanios:
@skye I made the changes you suggested except the one about the boolean function which I think is correct, but could you please check and let me know if I'm confused about something? :)

tensorflow/core/graph/graph_test.cc
skye: Ah right, my bad.

tensorflow/core/graph/graph_test.cc
```diff
@@ -477,6 +496,47 @@ TEST_F(GraphTest, AddControlEdge) {
...
+  graph_.RemoveControlEdge(edge_2);
+  ASSERT_TRUE(!ControlEdgeExistsInGraphOrNodeDef(c, a));
+  ASSERT_TRUE(!ControlEdgeExistsInGraphOrNodeDef(a, b));
+
+  // Test removing a duplicate control edge.
+  // Note that unless allow_duplicates is true, the duplicate edge
+  // will not be added. That's why we expect edge_4 to be a null
+  // pointer. We are not testing with allow_duplicates set to true,
+  // as that is a highly unlikely use case that does not make much
+  // sense.
```
skye: I actually meant to test with allow_duplicates = true. But this is fine.

skye (approved):

tensorflow/core/graph/graph_test.cc
eaplatanios: I can add that too if you want. :)

skye: I need to run soon, and will be gone until next week, so this is fine, or do it in a separate PR and I'll review when I get back.

eaplatanios:
Sounds good! Have fun wherever you’re going! :)

gunan:
Jenkins, test this please.

gunan:
Jenkins, test this please.

PR #13517, opened 2017-10-06
Fix crash when `tf.pad` is used with int64 paddings.

### LINKED ISSUE #13506: tf.image.pad_to_bounding_box crashes when passed bounds with dtype int64
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**:from pip in virtualenv
- **TensorFlow version (use command below)**:v1.3.0-rc2-20-g0787eee 1.3.0
- **Python version**: '3.5.2 (default, Nov 17 2016, 17:05:23) \n[GCC 5.4.0 20160609]'
- **CUDA/cuDNN version**:
- **GPU model and memory**:

## Description

Passing arguments of type int64 to `tf.image.pad_to_bounding_box` triggers a crash of the python interpreter. This is a bug because the type required by `tf.image.pad_to_bounding_box` not documented anywhere and just causes a crash with a cryptic error message.

## Sources/Logs

The following snippet crashes the whole python interpreter with a core dump.

    import tensorflow as tf
    i = tf.constant([0, 0, 3, 3], dtype=tf.int64)
    img = tf.ones([1,1,1], dtype=tf.float32)
    sess = tf.Session()
    sess.run(tf.image.pad_to_bounding_box(img, i[0], i[1], i[2], i[3]))

And leaves the following 

    2017-10-05 13:51:24.789715: F tensorflow/core/framework/tensor.cc:493] Check failed: dtype() == expected_dtype (9 vs. 3)

yongtang:
This fix tries to fix the issue raised in #13506 where int64 data types for bounds in `tf.image.pad_to_bounding_box` crashes.

The reason of the crash is caused by the fact that int64 was directly converted into int32 without passing through kernel registeration.

This fix fixes the issue by adding `typename Tpadding` to the template and adds appropriate kernels.

This fix fixes #13506.

### DIFF
--- a/tensorflow/core/kernels/pad_op.cc
+++ b/tensorflow/core/kernels/pad_op.cc
@@ -218,14 +218,27 @@ TF_CALL_GPU_NUMBER_TYPES(DECLARE_GPU_SPECS);
                               .TypeConstraint<T>("T")             \
                               .TypeConstraint<int32>("Tpaddings") \
                               .HostMemory("paddings"),            \
-                          PadOp<GPUDevice, T>);                   \
+                          PadOp<GPUDevice, T, int32>);            \
+  REGISTER_KERNEL_BUILDER(Name("Pad")                             \
+                              .Device(DEVICE_GPU)                 \
+                              .TypeConstraint<T>("T")             \
+                              .TypeConstraint<int64>("Tpaddings") \
+                              .HostMemory("paddings"),            \
+                          PadOp<GPUDevice, T, int64>);            \
   REGISTER_KERNEL_BUILDER(Name("PadV2")                           \
                               .Device(DEVICE_GPU)                 \
                               .TypeConstraint<T>("T")             \
                               .TypeConstraint<int32>("Tpaddings") \
                               .HostMemory("paddings")             \
                               .HostMemory("constant_values"),     \
-                          PadOp<GPUDevice, T>)
+                          PadOp<GPUDevice, T, int32>)             \
+  REGISTER_KERNEL_BUILDER(Name("PadV2")                           \
+                              .Device(DEVICE_GPU)                 \
+                              .TypeConstraint<T>("T")             \
+                              .TypeConstraint<int64>("Tpaddings") \
+                              .HostMemory("paddings")             \
+                              .HostMemory("constant_values"),     \
+                          PadOp<GPUDevice, T, int64>)
 
 TF_CALL_GPU_NUMBER_TYPES(REGISTER_GPU_KERNEL);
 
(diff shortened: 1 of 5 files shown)


frankchn:
Jenkins, test this please.

rryan:
Happy to review this but I'm on vacation until 10/9.

rryan:
Jenkins, test this please.

tensorflow/core/kernels/pad_op.cc
```diff
@@ -177,11 +193,17 @@ TF_CALL_POD_TYPES(REGISTER_KERNEL);
 namespace functor {
 #define DECLARE_GPU_SPEC(T, Dims)                                         \
   template <>                                                             \
-  void Pad<GPUDevice, T, Dims>::operator()(                               \
+  void Pad<GPUDevice, T, int32, Dims>::operator()(                               \
```
rryan: please fix the alignment of the \ at the end of the line to match the other lines

tensorflow/core/kernels/pad_op.cc
```diff
@@ -177,11 +193,17 @@ TF_CALL_POD_TYPES(REGISTER_KERNEL);
...
       const GPUDevice& d, typename TTypes<T, Dims>::Tensor output,        \
       typename TTypes<T, Dims>::ConstTensor input,                        \
       Eigen::array<Eigen::IndexPair<int32>, Dims> paddings, T pad_value); \
-  extern template struct Pad<GPUDevice, T, Dims>;
+  extern template struct Pad<GPUDevice, T, int32, Dims>; \
+  template <>                                                             \
+  void Pad<GPUDevice, T, int64, Dims>::operator()(                               \
```
rryan: please fix the alignment of the \ at the end of the line to match the other lines

rryan (changes requested):
Thanks for the fix! Since this is really a bug in tf.pad, I'm renaming the PR to reflect that.

Could you please add a test to pad_op_test.py that paddings can be int64?
https://github.com/tensorflow/tensorflow/blob/48be6a56d5c49d019ca049f8c48b2df597594343/tensorflow/python/kernel_tests/pad_op_test.py

tensorflow/core/kernels/pad_op.cc
yongtang: Thanks @rryan. The alignment has been fixed.

tensorflow/core/kernels/pad_op.cc
yongtang: Thanks. Done.

yongtang:
Thanks @rryan for the review. The PR has been updated with alignment of `\` fixed. Additional unit tests have also been added in `pad_op_test.py` for `int64`. Without the PR the new test case will trigger the crash.

rryan (approved):

rryan:
LGTM, thanks!

rryan:
Jenkins, test this please.

benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

PR #13534, opened 2017-10-06
Fix unevaluated link in "Reading data" docs

jeffcarp:
In the pull-out in https://www.tensorflow.org/api_guides/python/reading_data#Reading_from_files
there's a link that's unevaluated due to beginning with a `$` instead of a `@`.


frankchn:
Jenkins, test this please.

frankchn (approved):

PR #13553, opened 2017-10-07
Add `<` operator for IndicesRowIterator

### LINKED ISSUE #12000: error C2678: binary '<': no operator found which takes a left-hand operand of type IndicesRowIterator
If you build a debug version of the current tensorflow version 1.3.0 on Windows  with CMAKE  following error is occurred:

(ClCompile target) ->
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.10.25017\include\xutility(978): error C2678: binary '<': no operator found which
takes a left-hand operand of type 'tensorflow::boosted_trees::utils::`anonymous-namespace'::IndicesRowIterator' (or there is no acceptable conversion) (compiling sou
rce file C:\Development\dev\test_projects\deeplearning\tensorflow\tensorflow\tensorflow\contrib\boosted_trees\lib\utils\sparse_column_iterable.cc) [C:\Development\de
v\test_projects\deeplearning\tensorflow_build64\tf_core_kernels.vcxproj]

The reason is **IndicesRowIterator::operator<()** is missing.
Adding these lines 

  **bool operator<( const IndicesRowIterator& other ) const {
     QCHECK_LT( iter_, other.iter_ );
     return ( row_idx_ < other.row_idx_ );
  }**

solves the problem.

###System information
tensorflow 1.3.0
Windows 10
VisualStudio 2017
CMake 3.8.1

yongtang:
This fix adds the `<` operator for IndicesRowIterator to address C2678 error in VS Debug mode, as was specified in #12000.

This fix fixes #12000.

### DIFF
--- a/tensorflow/contrib/boosted_trees/lib/utils/sparse_column_iterable.cc
+++ b/tensorflow/contrib/boosted_trees/lib/utils/sparse_column_iterable.cc
@@ -101,6 +101,11 @@ class IndicesRowIterator
     return (row_idx_ == other.row_idx_);
   }
 
+  bool operator<(const IndicesRowIterator& other) const {
+    QCHECK_LT(iter_, other.iter_);
+    return (row_idx_ < other.row_idx_);
+  }
+
   Eigen::Index row_idx() const { return row_idx_; }
 
  private:


caisq:
ping @ThomasColthurst

gunan:
@ThomasColthurst Could you review the change?
Jenkins, test this please.

tensorflow/contrib/boosted_trees/lib/utils/sparse_column_iterable.cc
```diff
@@ -101,6 +101,11 @@ class IndicesRowIterator
     return (row_idx_ == other.row_idx_);
   }
 
+  bool operator<(const IndicesRowIterator& other) const {
+    QCHECK_LT(iter_, other.iter_);
```
yk5: This should be QCHECK_EQ, to make sense, as in other operators.
Given that, how do you make sure this PR doesn't break things?
Can you add some test cases?

martinwicke: A test would be a good idea. Another option would be to simply throw an exception (NotImplementedError) to clarify that we don't actually expect this to be used.

yongtang:
It looks like another PR #14404 is making the same change. As #14404 has already been merged, will close this PR. Thanks all for the review!

PR #13570, opened 2017-10-08
update dnn.py to clarify outputs arg to predict fn

yufengg:
It is unclear what the valid values are for predict(). Since the 2 variants are predict_classes and predict_proba, it's easy to think that the valid values are "classes" and "proba", which is untrue.
I _think_ I've listed all the relevant values for DNNClassifier -- at least that's what the error msg told me when I used "proba" as my arg :D


martinwicke:
All of the contrib Estimator will be dropped in favor of tf.estimator.Estimator, which does not have a predict_classes methods, but retains the predict method returning all the outputs. I think while this change looks like a good idea, it would be confusing in this context.

PR #13571, opened 2017-10-08
top level estimator docs update to match dnn.py

yufengg:
see #13570


ispirmustafa (commented):
Valid values are not univarsal. it is determined by model_fn or the canned estimator.

PR #13572, opened 2017-10-08
Adding --verify_images to example image retrainer.

ianmaddox:
Adding --verify_images to allow soft checking of image files before bottleneck creation.
When enabled, invalid images are ignored with a warning instead of throwing an exception.

### DIFF
--- a/tensorflow/examples/image_retraining/retrain.py
+++ b/tensorflow/examples/image_retraining/retrain.py
@@ -96,6 +96,7 @@ from __future__ import print_function
 
 import argparse
 from datetime import datetime
+from PIL import Image
 import hashlib
 import os.path
 import random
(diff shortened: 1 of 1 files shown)


caisq:

petewarden (changes requested):
Sorry for the slow reply on this one. In general this is a great idea, and would make the script more usable. The dependency on PIL is a problem though. We don't require this library otherwise, and it's pretty heavyweight to install, especially on non-standard systems like the Pi: https://stackoverflow.com/questions/20176883/how-to-install-pil-with-jpeg-support-on-a-raspberry-pi

I'd be happy if you were able to replace this check using standard DecodeJpeg TF ops instead, since that wouldn't introduce a new dependency.

martinwicke:
@ianmaddox any updates? Marking as stalled for now.

drpngx:
@ianmaddox Feel free to reopen when you get a chance to take a look.

PR #13579, opened 2017-10-09
[XLA] Reorder the parameters in a map inline operation according to the parameter number

DavidNorman:
The inliner (not typically in use) has a fault where it doesn't spot that the parameters within the mapped operation are not in the same order as the actual parameter numbers on the operands.

i.e. the mapped operation looks like:

```
  Param(1)         Param(0)
      |              |
      -- Binary-op ---
            |
```

Previously the operands were passed to the clone in the order that they were supplied to the binary op - not the order as defined by the parameter numbers.

This fixes the issue (which could be potentially be called `b/35786417`)

### DIFF
--- a/tensorflow/compiler/xla/service/inliner.cc
+++ b/tensorflow/compiler/xla/service/inliner.cc
@@ -90,8 +90,12 @@ Status InlinerVisitor::HandleMap(
     // different than the map shape. Hence, a broadcast is needed, else the
     // cloned operand with new shape and operands work.
     if (root.opcode() != HloOpcode::kConstant) {
+      std::vector<HloInstruction*> params;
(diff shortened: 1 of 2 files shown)


caisq:

DavidNorman:
I don't think that the test failures are related to my change.

hawkinsp (commented):
Thanks for the contribution!

Looks good; would you mind adding a unit test to inliner_test.cc so we can be sure this doesn't break in the future?

DavidNorman:
sure thing

DavidNorman:
added

hawkinsp (approved):
Looks good! Thanks for the contribution!

caisq:

caisq:

PR #13598, opened 2017-10-09
Fix discrepancy between docs and registered kernels for `tf.ones_like`/`tf.zeros_like`

yongtang:
This fix tries to address the discrepancy between docs and registered kernels for `tf.ones_like`. From the implementations the `OnesLike`/`ZerosLike` are registered with all POD types. However, in the documentation several data types are missing (`uint8`, `int8`, `uint16`, `int16`, `bool`).

This fix addresses the issue by adding missing types to documentation.

### DIFF
--- a/tensorflow/core/ops/array_ops.cc
+++ b/tensorflow/core/ops/array_ops.cc
@@ -665,7 +665,9 @@ y: a tensor of the same shape and type as x but filled with zeros.
 REGISTER_OP("OnesLike")
     .Input("x: T")
     .Output("y: T")
-    .Attr("T: {float, double, int32, int64, complex64, complex128}")
+    .Attr(
+        "T: {float, double, int8, uint8, int16, uint16, int32, int64, "
+        "complex64, complex128, bool}")
     .SetShapeFn(shape_inference::UnchangedShape)
     .Doc(R"doc(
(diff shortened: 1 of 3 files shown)


tensorflow/core/ops/array_ops.cc
```diff
@@ -665,7 +665,9 @@ y: a tensor of the same shape and type as x but filled with zeros.
 REGISTER_OP("OnesLike")
     .Input("x: T")
     .Output("y: T")
-    .Attr("T: {float, double, int32, int64, complex64, complex128}")
+    .Attr(
```
ebrevdo: Is there a unit test that a ones_like(boolean tensor) actually runs?

yongtang:
I signed it!

tensorflow/core/ops/array_ops.cc
yongtang: The unit test for bool (and uint16) was not there so I updated the unit tests (included in `tensorflow/python/kernel_tests/constant_op_test.py`) and it should be covered now.

yongtang:
Thanks @ebrevdo for the review. The PR has been updated. Please take a look.

ebrevdo (approved):

gunan:
Jenkins, test this please.

PR #13599, opened 2017-10-10
Rewrite matrix_set_diag GPU kernel.

gunan:
Base new implementation on matrix_diag_op.


gunan:
Rasmus, I am now going to test this on windows. please feel free to throw this PR out if there should be a better way to implement this.

gunan:
Running windows GPU tests here:
http://ci.tensorflow.org/view/Experimental/job/tf-win-gpu/3/

Hopefully, by tomorrow morning we will have a result.

gunan:
Still fails on windows. Not sure what is missing.

rmlarsen:
As discussed, this is not the right direction to go.

rmlarsen:
Instead, I recommend using the recent rewrite of matrix_band_part as a template. Getting rid of the Eigen generator mechanism makes the code more readable and much faster.

PR #13602, opened 2017-10-10
Calculate largest max rather than smallest max in quantized_add_op.cc

Taneb:
I think this is a simple error that just didn't get spotted


martinwicke (approved):
How embarrassing.

martinwicke:
Jenkins, test this please.

Issue #13627, opened 2017-10-11
HDFS user impersonation


kouzant:
TensorFlow  does support HDFS filesystem but there is no way to specify as which user to access the filesystem.

The native hdfs library provides a function to set the user name, hdfsBuilderSetUserName() similar to hdfsBuilderSetNameNode(). There is also hdfsConnectAsUser() where you can specify the user, builds the hdfsBuilder struct, set the username among other arguments and return the FileSystem handle.

Currently I don't see any way to provide a username in TensorFlow and more specifically in tensorflow/core/platform/hadoop/hadoop_file_system.cc

I suppose that something like the following would be sufficient.
https://github.com/kouzant/tensorflow/commit/eacef5cb81d09d0490403fde33de8e5526f212ad

jimdowling:
We can install our own build of tensorflow. And submit a PR to tensorflow in the meantime.

reedwm:
Marking as contributions weclome since it seems @jimdowling will submit a PR.

/CC @jhseu in case he has any comments.

jimdowling:
Sorry, that was an internal comment for @kouzant  - who will submit the PR.

mihaimaruseac:
Closing since HDFS is no longer provided by TF itself

PR #13654, opened 2017-10-12
Enable setting HDFS user

kouzant:
Following issue https://github.com/tensorflow/tensorflow/issues/13627 add enable TensorFlow to access HDFS as a specific HDFS user instead of the the system user owning the Tf process

### DIFF
--- a/tensorflow/core/platform/hadoop/hadoop_file_system.cc
+++ b/tensorflow/core/platform/hadoop/hadoop_file_system.cc
@@ -60,6 +60,7 @@ class LibHDFS {
   std::function<int(const char*, char**)> hdfsConfGetStr;
   std::function<void(hdfsBuilder*, const char* kerbTicketCachePath)>
       hdfsBuilderSetKerbTicketCachePath;
+  std::function<void(hdfsBuilder*, const char*)> hdfsBuilderSetUserName;
   std::function<int(hdfsFS, hdfsFile)> hdfsCloseFile;
   std::function<tSize(hdfsFS, hdfsFile, tOffset, void*, tSize)> hdfsPread;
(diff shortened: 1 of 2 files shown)


kouzant:
I signed it!

caisq:

tensorflow/core/platform/hadoop/hadoop_file_system.cc
```diff
@@ -170,6 +172,12 @@ Status HadoopFileSystem::Connect(StringPiece fname, hdfsFS* fs) {
   if (ticket_cache_path != nullptr) {
     hdfs_->hdfsBuilderSetKerbTicketCachePath(builder, ticket_cache_path);
   }
+
+  const char* hdfs_user = getenv("HDFS_USER");
```
jhseu: HDFS_USER -> HADOOP_USER_NAME to match existing HDFS conventions?

kouzant: Yes you are right. I'll update it.

kouzant:
@jhseu Pushed the change

jhseu:
Are you sure it doesn't already work without this change? I vaguely recall testing it and it picked up the username from HADOOP_USER_NAME.

kouzant:
We have a use case where we run TensorFlowOnSpark from Apache Livy. When we set HADOOP_USER_NAME then we get an impersonation error saying that User_A cannot impersonate User_A, which is correct since we don't give permissions to User_A to impersonate other users, hence not itself.

Nevertheless, when I set HADOOP_USER_NAME Tensorflow does access HDFS as that user, so I suppose there is no need for a patch.

jhseu:
Thanks. As I understand it, we can close this out. Feel free to reopen if you disagree.

PR #13659, opened 2017-10-12
Enable vectorization on z13

namrata-ibm:
Enable vectorization on z13 to improve performance.


caisq:

benoitsteiner (approved):

Issue #13600, opened 2017-10-10
LayerNormBasicLSTMCell causes `bias key not found in checkpoint` when layer_norm=False


bxshi:
When initializing `LayerNormBasicLSTMCell`, it has a parameter `layer_norm` which controls whether we want to enable layer norm or not. I assume `layer_norm=True` should be set during training and `layer_norm=False` for evaluation. However, if I use this in an Estimator, due to the following line

https://github.com/tensorflow/tensorflow/blob/a2d9b3bf5f9e96bf459074d079b01e1c74b25afa/tensorflow/contrib/rnn/python/ops/rnn_cell.py#L1331

it will not initialize the `bias` term because `layer_norm=True`, resulting in a `NotFoundError` when loading the saved checkpoint with `layer_norm=False`.

What should be the expected behavior of this? Should this cell applies the bias anyway regardless the `layer_norm`? If we do not use the bias during training, I see no points to use it during inference.

bxshi:
It should be an easy fix. Please let me know if you want me to contribute on this.

tatatodd:
@ebrevdo Can you comment on this?

ebrevdo:
The original author should address the issue.

tatatodd:
@ebrevdo my apologies, but it looks like the original author was an intern and not on github, and you reviewed the change that introduced LayerNormBasicLSTMCell.

Do you have recommendations for anyone else who might have the context to evaluate this issue?

ebrevdo:
@lukaszkaiser can you recommend someone to review?  it's not clear to me that layer norm should be disabled at inference time.

lukaszkaiser:
That's a hard question. I'd suggest maybe we ask Erich Elsen? Erich: do you know this?

bxshi:
@lukaszkaiser My bad. Actually there is no need to disable `layer_norm` during inference. Closing the issue.

ekelsen:
For reference, @lukaszkaiser it looks like you meant me, but assigned Eric D. Nielsen instead :)  I just happened to randomly find this, sorry I didn't reply before.

PR #13689, opened 2017-10-13
Always add bias term to the output of LayerNormBasicLSTMCell.

bxshi:
These would solve the problem of referring untrained bias term during inference with `self._layer_norm=False`. 

Issue ref: #13600


ebrevdo:
Let's wait on the discussion in #13600 to resolve before advancing a PR?

PR #13690, opened 2017-10-13
Adding equals and hashCode to Shape.

mdaley:
Hello, I know this is a bit of a simple and trivial pull request.  However, for some other work I'm doing I wanted to compare `Shape` instances in some tests and I found that it didn't work in the standard way.  Hence this pull request to just add an `equals` method. It has some tests and also does `hashCode` because it's a common thing to do at the same time.

### DIFF
--- a/tensorflow/java/src/main/java/org/tensorflow/Shape.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Shape.java
@@ -119,7 +119,7 @@ public final class Shape {
 
   private boolean hasUnknownDimension() {
     if (shape == null) {
-      return false;
+      return true;
     }
 
     for (long dimension : shape) {
--- a/tensorflow/java/src/main/java/org/tensorflow/Shape.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Shape.java
@@ -77,6 +77,24 @@ public final class Shape {
     return shape[i];
   }
 
+  @Override
+  public int hashCode() {
+    return Arrays.hashCode(shape);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
+
+    if (obj instanceof Shape && Arrays.equals(this.shape, ((Shape) obj).shape)) {
+      return true;
+    }
+
+    return super.equals(obj);
+  }
+
   /** Succinct description of the shape meant for debugging. */
   @Override
   public String toString() {
--- a/tensorflow/java/src/main/java/org/tensorflow/Shape.java
+++ b/tensorflow/java/src/main/java/org/tensorflow/Shape.java
@@ -89,7 +89,7 @@ public final class Shape {
     }
 
     if (obj instanceof Shape && Arrays.equals(this.shape, ((Shape) obj).shape)) {
-      return true;
+      return !hasUnknownDimension();
     }
 
     return super.equals(obj);
@@ -116,4 +116,18 @@ public final class Shape {
   }
 
   private long[] shape;
+
+  private boolean hasUnknownDimension() {
+    if (shape == null) {
+      return false;
+    }
+
+    for (long dimension : shape) {
+      if (dimension == -1) {
+        return true;
+      }
+    }
+
+    return false;
+  }
 }
--- a/tensorflow/java/src/test/java/org/tensorflow/ShapeTest.java
+++ b/tensorflow/java/src/test/java/org/tensorflow/ShapeTest.java
@@ -81,13 +81,16 @@ public class ShapeTest {
 
   @Test
   public void equalsWorksCorrectly() {
-    assertEquals(Shape.make(-1, 2, 3), Shape.make(-1, 2, 3));
     assertEquals(Shape.scalar(), Shape.scalar());
     assertEquals(Shape.unknown(), Shape.unknown());
+    assertEquals(Shape.make(1, 2, 3), Shape.make(1, 2, 3));
 
     assertNotEquals(Shape.make(1,2), null);
     assertNotEquals(Shape.make(1,2), new Object());
-    assertNotEquals(Shape.make(-1, 2, 3), Shape.make(-1, 2, 4));
+    assertNotEquals(Shape.make(1, 2, 3), Shape.make(1, 2, 4));
+
+    assertNotEquals(Shape.make(-1), Shape.make(-1));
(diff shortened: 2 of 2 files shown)


tensorflow/java/src/test/java/org/tensorflow/ShapeTest.java
```diff
@@ -77,4 +80,36 @@ public void nodesInAGraph() {
...
+  @Test
+  public void hashCodeIsAsExpected() {
+    long[] d1 = new long[] {1, 2, 3, 4};
+    long[] d2 = new long[] {};
+
+    Shape s1 = Shape.make(1, 2, 3, 4);
+    Shape s2 = Shape.scalar();
+
+    assertEquals(Arrays.hashCode(d1), s1.hashCode());
+    assertEquals(Arrays.hashCode(d2), s2.hashCode());
```
asimshankar: These assertions are testing the implementation rather than behavior. Let's modify them to do the latter, so something like:

```java
assertEquals(Shape.make(1, 2, 3, 4).hashCode(), Shape.make(1, 2, 3, 4).hashCode());
assertNotEquals(Shape.make(1, 2).hashCode9), Shape.make(1, 3).hashCode());
```

tensorflow/java/src/test/java/org/tensorflow/ShapeTest.java
```diff
@@ -77,4 +80,36 @@ public void nodesInAGraph() {
...
+  public void equalsWorksCorrectly() {
+    Shape s1 = Shape.make(-1, 2, 3);
+    Shape s2 = s1;
+    Shape s3 = Shape.make(-1, 2, 3);
+    Shape s4 = Shape.make(-1, 2, 4);
+    Shape s5 = Shape.make(-1, 2, 3, 4);
+    Shape s6 = Shape.make(-1, 2);
+    Object o = new Object();
+
+    assertEquals(s1, s2);
```
asimshankar: Optional: The tests might be easier to follow if we don't have to refer back to what `s1`, `s3` etc. mean. Perhaps inline the `Shape` objects? So something like:

```java
assertEquals(Shape.make(-1, 2, 3), Shape.make(-1, 2, 3));
assertNotEquals(Shape.make(1, 2), new Object());
assertNotEquals(Shape.make(-1, 2, 3), Shape.make(-1, 2, 4));
assertEquals(Shape.scalar(), Shape.scalar());
```

asimshankar (changes requested):
Thanks for the contribution @mdaley . Looks find overall, just some minor nitpicks on the test.

Thanks!

tensorflow/java/src/test/java/org/tensorflow/ShapeTest.java
mdaley: Yes, good comment.  I'll make some changes.  Same with your comment below.  :-)

mdaley:
Hello @asimshankar.  Do my changes to the tests work for you?  You'll see I also added some stuff relating to the `scalar()` and `unknown()` variants of Shape.  I worried about potential NPE if the long array wrapped by shape was null.  Also, I have a question: should `Shape().unknown()` be the same as `Shape.unknown()`?  I guess that's a bit of a philosophical point; and the simplest answer is to say yes.

asimshankar:
Sorry for the delayed response @mdaley .
Great question about handling unknown shapes and dimensions. Since objects of the `Shape` class represent partially known shapes, I think it might be safer to consider shapes equal iff each dimension is known and equal (i.e., no -1).

Thanks for thinking about this (I should have in my first round of reviews).
The Python `TensorShape` class does not consider two shapes with unknown dimensions to be equal, and we should do the same here.

For example, in Python:

```python
import tensorflow as tf

x = tf.placeholder(tf.float32, [None, 1])
y = tf.placeholder(tf.float32, [None, 1])
assert not (x.shape == y.shape) # First dimension is unknown

x = tf.placeholder(tf.float32, [1, 1])
y = tf.placeholder(tf.float32, [1, 1])
assert x.shape == y.shape
```

(Implemented in [`TensorShape.__eq__`](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/python/framework/tensor_shape.py#L807), [`Dimension.__eq__`](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/python/framework/tensor_shape.py#L46))

Confusingly, it does treat two completely unknown shapes as equal:

```python
x = tf.placeholder(tf.float32)
y = tf.placeholder(tf.float32)
assert x.shape == y.shape
```

I'll dig into that a bit, but let's update the code to not treat shapes equal when one of the dimensions is unknown (i.e., one of the elements of the `shape` array is `-1`)

tensorflow/java/src/main/java/org/tensorflow/Shape.java
```diff
@@ -98,4 +116,18 @@ public String toString() {
   }
 
   private long[] shape;
+
+  private boolean hasUnknownDimension() {
+    if (shape == null) {
+      return false;
```
asimshankar: Let's make this `return true;`. After digging in a bit, I think that the fact that Python returns `True` for `TensorShape(None).__eq__(TensorShape(None))` is a bug.

asimshankar (changes requested):
Thanks for being thorough @mdaley .
One more minor change and I think we're good to go.

Very much appreciate your contribution.

mdaley:
Hi @asimshankar, I have made that final change.  I agree with what you say; are you going to raise that python-side bug that you were talking about?

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please.

asimshankar:
@mdaley : Yup, I'll see if the Python change is trivial (no unexpected test breakages, or concerns around change to a stable API), and if so just do them, else file a bug and list out any gotcha.

Thanks.

mdaley:
@asimshankar Asim, I've enjoyed working with you on this!  I shall have to find some more ways to be involved...

mdaley:
Oh, hope those failed tests weren't due to me!

gunan:
Jenkins, test this please.

mdaley:
Jenkins, test this please.

gunan:
Jenkins, test this please.

asimshankar:
Jenkins, test this please

gunan:
Linux CPU Tests Makefile failure OK to ignore, it is superseeded by Ubuntu Makefile

PR #13695, opened 2017-10-13
Add  'log_progress' argument for tf.estimator.Estimator's evaluate function

ursk:
The output of `estimator.evaluate` can be unnecessarily verbose with logging level INFO and many batches to evaluate on. The old `tf.contrib.learn.Estimator` had the ability to pass the `log_progress=False` argument along to to `tensorflow.python.training.evaluation` where it suppresses logging in the `after_run` function. This PR adds the same functionality for `tf.Estimator`.

### DIFF
--- a/tensorflow/python/training/evaluation.py
+++ b/tensorflow/python/training/evaluation.py
@@ -19,6 +19,7 @@ from __future__ import division
 from __future__ import print_function
 
 import time
+import math
 
 from tensorflow.python.framework import dtypes
 from tensorflow.python.framework import ops
@@ -91,8 +92,9 @@ class _StopAfterNEvalsHook(session_run_hook.SessionRunHook):
     self._num_evals = num_evals
     self._evals_completed = None
     self._log_progress = log_progress
-    # Log only every 10th evaluation if more than 100 total.
-    self._log_frequency = 1 if num_evals < 100 else 10
+    # Reduce logging frequency if there are 20 or more evaluations.
+    self._log_frequency = (1 if (num_evals is None or num_evals < 20)
+                           else math.floor(num_evals / 10.))
 
   def _set_evals_completed_tensor(self, updated_eval_step):
     self._evals_completed = updated_eval_step
@@ -108,7 +110,8 @@ class _StopAfterNEvalsHook(session_run_hook.SessionRunHook):
       if self._num_evals is None:
         logging.info('Evaluation [%d]', evals_completed)
       else:
-        if (evals_completed % self._log_frequency) == 1:
+        if ((evals_completed % self._log_frequency) == 0 or
+            (self._num_evals == evals_completed)):
           logging.info('Evaluation [%d/%d]', evals_completed, self._num_evals)
     if self._num_evals is not None and evals_completed >= self._num_evals:
       run_context.request_stop()
--- a/tensorflow/python/training/evaluation.py
+++ b/tensorflow/python/training/evaluation.py
@@ -91,6 +91,8 @@ class _StopAfterNEvalsHook(session_run_hook.SessionRunHook):
     self._num_evals = num_evals
     self._evals_completed = None
     self._log_progress = log_progress
+    # Log only every 10th evaluation if more than 100 total.
+    self._log_frequency = 1 if num_evals < 100 else 10
 
   def _set_evals_completed_tensor(self, updated_eval_step):
     self._evals_completed = updated_eval_step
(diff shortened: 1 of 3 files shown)


caisq:
@ursk Please see pylint errors at:
https://ci.tensorflow.org/job/tensorflow-pull-requests-sanity/9080/consoleFull

xiejw:
I think changing API is not an ideal solution to this. But this is an usability issue.

I would suggest we could change the behavior of the logging. For steps >= XXX, we only log evey 1/10th of the progress. 

ispirmustafa@, wdyt?

ursk:
@caisq, just pushed a commit to address the PyLint error. It's passing now. I wasn't aware of the TPU side of Estimator.

@xiejw I'm happy to follow your suggestion to long only a subset of the validation set. The usability issue is that e.g. with 50,000 validation images at a batch size of 128, we get 390 lines of logging, cutting that down to 39 lines would solve my usability issue. I'll go ahead and make that change if you confirm that's what we want.

xiejw:
The tpu_estimator is broken due to the fact you are using public API.

I think reducing the log frequency is better. Can you convert your change to that? With that being said, I think you need to update _StopAfterNEvalsHook and skip some logging steps (but remain the final one)

Thanks.

xiejw:
s/using/changing   public API. Sorry for the typo.

ursk:
Is this what you had in mind?

tensorflow/python/training/evaluation.py
```diff
@@ -91,6 +91,8 @@ def __init__(self, num_evals, log_progress=True):
     self._num_evals = num_evals
     self._evals_completed = None
     self._log_progress = log_progress
+    # Log only every 10th evaluation if more than 100 total.
+    self._log_frequency = 1 if num_evals < 100 else 10
```
xiejw: Two small changes:

1) num_evals could be None, so you have a if condition check here otherwise num_evals < 100 will trigger TypeError in python3

2) Probably I did not express clearly sorry. I meant 1/10 of the progress, so, I was thinking

math.floor(num_evals / 10.)

tensorflow/python/training/evaluation.py
```diff
@@ -106,7 +108,8 @@ def after_run(self, run_context, run_values):
       if self._num_evals is None:
         logging.info('Evaluation [%d]', evals_completed)
       else:
-        logging.info('Evaluation [%d/%d]', evals_completed, self._num_evals)
+        if (evals_completed % self._log_frequency) == 1:
+          logging.info('Evaluation [%d/%d]', evals_completed, self._num_evals)
```
xiejw: Here you need one more condition

if ((evals_completed % self._log_frequency) == 1 or
    self._num_evals == evals_completed):
  logging.info('Evaluation [%d/%d]', evals_completed, self._num_evals)

It is very important from usability perspective to print the final one.

ursk: Thanks for being so patient. Here's a couple of examples where we are at now:

With 390 steps:
```
INFO:tensorflow:Evaluation [39/390]
INFO:tensorflow:Evaluation [78/390]
INFO:tensorflow:Evaluation [117/390]
INFO:tensorflow:Evaluation [156/390]
INFO:tensorflow:Evaluation [195/390]
INFO:tensorflow:Evaluation [234/390]
INFO:tensorflow:Evaluation [273/390]
INFO:tensorflow:Evaluation [312/390]
INFO:tensorflow:Evaluation [351/390]
INFO:tensorflow:Evaluation [390/390]
```

With 25 steps:
```
INFO:tensorflow:Evaluation [2/25]
INFO:tensorflow:Evaluation [4/25]
INFO:tensorflow:Evaluation [6/25]
INFO:tensorflow:Evaluation [8/25]
INFO:tensorflow:Evaluation [10/25]
INFO:tensorflow:Evaluation [12/25]
INFO:tensorflow:Evaluation [14/25]
INFO:tensorflow:Evaluation [16/25]
INFO:tensorflow:Evaluation [18/25]
INFO:tensorflow:Evaluation [20/25]
INFO:tensorflow:Evaluation [22/25]
INFO:tensorflow:Evaluation [24/25]
INFO:tensorflow:Evaluation [25/25]
```

With 10 steps
```
INFO:tensorflow:Evaluation [1/10]
INFO:tensorflow:Evaluation [2/10]
INFO:tensorflow:Evaluation [3/10]
INFO:tensorflow:Evaluation [4/10]
INFO:tensorflow:Evaluation [5/10]
INFO:tensorflow:Evaluation [6/10]
INFO:tensorflow:Evaluation [7/10]
INFO:tensorflow:Evaluation [8/10]
INFO:tensorflow:Evaluation [9/10]
INFO:tensorflow:Evaluation [10/10]
```

Seems reasonable to me, although it's not always exactly 10 outputs, because of the rounding / truncation step.

xiejw (approved):
Thanks for the help here!

gunan:
Jenkins, test this please.

PR #13701, opened 2017-10-13
Fix typos in datasets guide

waprin:
I think `recommend`->`recommended` is definitely necessary.

`are` to `is` are apparently both technically accepted usages, but `is` is far more common and I believe `are` will sound jarring here to most native English speakers.

### DIFF
--- a/tensorflow/docs_src/programmers_guide/datasets.md
+++ b/tensorflow/docs_src/programmers_guide/datasets.md
@@ -44,7 +44,7 @@ To start an input pipeline, you must define a *source*. For example, to
 construct a `Dataset` from some tensors in memory, you can use
 `tf.data.Dataset.from_tensors()` or
 `tf.data.Dataset.from_tensor_slices()`. Alternatively, if your input
-data are on disk in the recommend TFRecord format, you can construct a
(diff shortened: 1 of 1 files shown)


tensorflow/docs_src/programmers_guide/datasets.md
```diff
@@ -44,7 +44,7 @@ To start an input pipeline, you must define a *source*. For example, to
 construct a `Dataset` from some tensors in memory, you can use
 `tf.data.Dataset.from_tensors()` or
 `tf.data.Dataset.from_tensor_slices()`. Alternatively, if your input
-data are on disk in the recommend TFRecord format, you can construct a
```
mrry: "data are on disk" is correct; "the recommend TFRecord format" is obviously wrong. Please revert the first change and keep the second one.

mrry (changes requested):

tensorflow/docs_src/programmers_guide/datasets.md
waprin: done

caisq (approved):

PR #13704, opened 2017-10-14
Clarify low-latency model stride length

daemon:
`first_filter_stride_y = 4` has no effect, since `first_filter_height = input_time_size`. If `cnn-one-fstride4` should be used, then `first_filter_stride_x` should be 4 instead. However, during my experimentation `cnn-one-fstride4` provided significantly worse results than the current TF impl, so I think the impl is fine.

This small fix just clarifies that the model isn't striding in the time dimension.


petewarden (approved):
Thanks for the fix, and sorry for the slow response. Looks good!

martinwicke:
Jenkins, test this please.

PR #13709, opened 2017-10-14
Tidy up tf_tests.cmake and reenable tests that are not failing anymore.

gunan:


mrry (approved):
Thanks for cleaning this up!

mrry:
Looks like tensorflow/python/training/coordinator_test.py failed :(. I think this is the same issue you fixed in `saver_test.py` (in ee50560b5fd2b1112e82377d7d094a0e6918f935) where `time.sleep()` doesn't sleep for long enough on Windows, and can lead to flakiness. Would it be easy to port the same solution?

mrry:
(Disabling the test in this PR is also a reasonable approach, though :).)

gunan:
I will follow this test internally.
I am really surprised it has not been more flaky.

PR #13721, opened 2017-10-14
Update documentation for uint16 support in `tf.resize_...` ops

yongtang:
This fix tries to address the different between the registered kernels and the documentation differnet for `uint16` support of:
`tf.resize_area`
`tf.resize_bicubic`
`tf.resize_bilinear`
`tf.resize_nearest_neighbor`

Though `uint16` is supported in the kernel, it is not documented in `image_ops.cc`. Unit test cases are also missing for `uint16` and `float16`.

This fix added the missing entries in tests and docs.


caisq:

vrv (approved):
I think the resize ops are only implemented for a few types on GPU, but CPU does register for this type, so this looks good.

PR #13727, opened 2017-10-15
AddSign and PowerSign optimizers for contrib.opt

kdubovikov:
Added optimizers with update rules from https://arxiv.org/pdf/1709.07417.pdf

### DIFF
--- a/tensorflow/contrib/opt/python/training/add_sign_optimizer.py
+++ b/tensorflow/contrib/opt/python/training/add_sign_optimizer.py
@@ -1,0 +1,71 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


kdubovikov:
I signed it!

vrv:
Thanks @kdubovikov for this PR, and for your interest in the paper!

It's worth noting that Irwan (the paper author) is already mostly done with a change to add these optimizers to GitHub.  I would be happy to take this change, but it is complicated due to the fact that most of our supported optimizers require C++ fused implementations for speed and distributed training reasons.

Given that this work is almost done, would you be okay if we closed this PR in favor of the C++ version that is coming soon?

kdubovikov:
Hi! Great news! I thought to develop C++ code later, but it’s clear that the author of the paper will manage this perfectly. Hope I’ll implement something more useful further along

vrv:
Great, I'll close it for now, but if we can't get the C++ changes in by the end of the month, I'll reopen.

PR #13731, opened 2017-10-15
Add GPU and CPU implementation of `tf.histogram_fixed_width`.

yongtang:
This fix adds the GPU and CPU implementation of `tf.histogram_fixed_width`.

The previous implementation was done in python. This fix adds C++ kernel for GPU and CPU.

The GPU version uses `CUB`'s API `cub::DeviceHistogram::HistogramRange`. The `range` is  constructed from the upper/lower limit and step size. (`HistogramEven` could not be used directly as the edge case is different).

The CPU version uses a transform to map the input into the bucket index, then did a bin count.

Note the output type of int64 on GPU is not supported yet as atomicAdd has no int64 at the moment.

### DIFF
--- a/tensorflow/core/kernels/histogram_op.cc
+++ b/tensorflow/core/kernels/histogram_op.cc
@@ -74,69 +74,69 @@ struct HistogramFixedWidthFunctor<CPUDevice, T, Tout> {
 template <typename Device, typename T, typename Tout>
 class HistogramFixedWidthOp : public OpKernel {
  public:
-  explicit HistogramFixedWidthOp(OpKernelConstruction* ctx) : OpKernel(ctx) {}
+  explicit HistogramFixedWidthOp(OpKernelConstruction* ctx) : OpKernel(ctx) {
+    OP_REQUIRES_OK(ctx, ctx->GetAttr("nbins", &nbins_));
+    OP_REQUIRES(
+        ctx, (nbins_ > 0),
+        errors::InvalidArgument("nbins should be a positive number, but got '",
+                                nbins_, "'"));
+  }
 
   void Compute(OpKernelContext* ctx) override {
     const Tensor& values_tensor = ctx->input(0);
     const Tensor& value_range_tensor = ctx->input(1);
-    const Tensor& nbins_tensor = ctx->input(2);
 
     OP_REQUIRES(ctx, TensorShapeUtils::IsVector(value_range_tensor.shape()),
                 errors::InvalidArgument("value_range should be a vector."));
     OP_REQUIRES(ctx, (value_range_tensor.shape().num_elements() == 2),
                 errors::InvalidArgument(
                     "value_range should be a vector of 2 elements."));
-    OP_REQUIRES(ctx, TensorShapeUtils::IsScalar(nbins_tensor.shape()),
-                errors::InvalidArgument("nbins should be a scalar."));
 
     const auto values = values_tensor.flat<T>();
     const auto value_range = value_range_tensor.flat<T>();
-    const int32 nbins = nbins_tensor.scalar<int32>()();
 
     OP_REQUIRES(
         ctx, (value_range(0) < value_range(1)),
         errors::InvalidArgument("value_range should satisfy value_range[0] < "
                                 "value_range[1], but got '[",
                                 value_range(0), ", ", value_range(1), "]'"));
-    OP_REQUIRES(
-        ctx, (nbins > 0),
-        errors::InvalidArgument("nbins should be a positive number, but got '",
-                                nbins, "'"));
 
     Tensor* out_tensor;
     OP_REQUIRES_OK(ctx,
-                   ctx->allocate_output(0, TensorShape({nbins}), &out_tensor));
+                   ctx->allocate_output(0, TensorShape({nbins_}), &out_tensor));
     auto out = out_tensor->flat<Tout>();
 
     OP_REQUIRES_OK(
         ctx, functor::HistogramFixedWidthFunctor<Device, T, Tout>::Compute(
-                 ctx, values, value_range, nbins, out));
+                 ctx, values, value_range, nbins_, out));
   }
+
+ private:
+  int nbins_;
 };
 
 #define REGISTER_KERNELS(type)                                           \
   REGISTER_KERNEL_BUILDER(Name("HistogramFixedWidth")                    \
                               .Device(DEVICE_CPU)                        \
                               .TypeConstraint<type>("T")                 \
-                              .TypeConstraint<int32>("Tout"),            \
+                              .TypeConstraint<int32>("dtype"),           \
                           HistogramFixedWidthOp<CPUDevice, type, int32>) \
   REGISTER_KERNEL_BUILDER(Name("HistogramFixedWidth")                    \
                               .Device(DEVICE_CPU)                        \
                               .TypeConstraint<type>("T")                 \
-                              .TypeConstraint<int64>("Tout"),            \
+                              .TypeConstraint<int64>("dtype"),           \
                           HistogramFixedWidthOp<CPUDevice, type, int64>)
 
 TF_CALL_REAL_NUMBER_TYPES(REGISTER_KERNELS);
 #undef REGISTER_KERNELS
 
 #if GOOGLE_CUDA
-#define REGISTER_KERNELS(type)                                \
-  REGISTER_KERNEL_BUILDER(Name("HistogramFixedWidth")         \
-                              .Device(DEVICE_GPU)             \
-                              .HostMemory("value_range")      \
-                              .HostMemory("nbins")            \
-                              .TypeConstraint<type>("T")      \
-                              .TypeConstraint<int32>("Tout"), \
+#define REGISTER_KERNELS(type)                                 \
+  REGISTER_KERNEL_BUILDER(Name("HistogramFixedWidth")          \
+                              .Device(DEVICE_GPU)              \
+                              .HostMemory("value_range")       \
+                              .TypeConstraint<type>("T")       \
+                              .TypeConstraint<int32>("dtype"), \
                           HistogramFixedWidthOp<GPUDevice, type, int32>)
 
 TF_CALL_GPU_NUMBER_TYPES(REGISTER_KERNELS);
--- a/tensorflow/core/kernels/histogram_op.cc
+++ b/tensorflow/core/kernels/histogram_op.cc
@@ -39,11 +39,12 @@ struct HistogramFixedWidthFunctor<CPUDevice, T, Tout> {
                         int32 nbins, typename TTypes<Tout, 1>::Tensor& out) {
     const CPUDevice& d = context->eigen_device<CPUDevice>();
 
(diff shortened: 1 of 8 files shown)


tensorflow/core/kernels/histogram_op.cc
```diff
@@ -0,0 +1,136 @@
...
+    const Tensor& nbins_tensor = ctx->input(2);
+
+    OP_REQUIRES(ctx, TensorShapeUtils::IsVector(value_range_tensor.shape()),
+                errors::InvalidArgument("value_range should be a vector."));
+    OP_REQUIRES(ctx, (value_range_tensor.shape().num_elements() == 2),
+                errors::InvalidArgument(
+                    "value_range should be a vector of 2 elements."));
+    OP_REQUIRES(ctx, TensorShapeUtils::IsScalar(nbins_tensor.shape()),
+                errors::InvalidArgument("nbins should be a scalar."));
+
```
rmlarsen: Check that nbins is positive and value range is non-empty.

tensorflow/core/kernels/histogram_op.cc
```diff
@@ -0,0 +1,136 @@
...
+    const auto values = values_tensor.flat<T>();
+    const auto value_range = value_range_tensor.flat<T>();
+    int32 nbins = nbins_tensor.scalar<int32>()();
```
rmlarsen: const

tensorflow/core/kernels/histogram_op.h
```diff
@@ -0,0 +1,41 @@
...
+==============================================================================*/
+
+#ifndef TENSORFLOW_HISTOGRAM_OP_H_
+#define TENSORFLOW_HISTOGRAM_OP_H_
+
+#include "tensorflow/core/framework/op_kernel.h"
+#include "tensorflow/core/framework/tensor_types.h"
+#include "tensorflow/core/framework/types.h"
+#include "tensorflow/core/lib/core/errors.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
```
rmlarsen: I don't think you need this include.

tensorflow/core/kernels/histogram_op.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace tensorflow {
+
```
rmlarsen: Get rid of empty lines between namespaces in this case.

tensorflow/core/kernels/histogram_op.h
```diff
@@ -0,0 +1,41 @@
...
+template <typename Device, typename T, typename Tout>
+struct HistogramFixedWidthFunctor {
+  static Status Compute(OpKernelContext* context,
+                        const typename TTypes<T, 1>::ConstTensor& values,
+                        const typename TTypes<T, 1>::ConstTensor& value_range,
+                        int32 nbins, typename TTypes<Tout, 1>::Tensor& out);
+};
+
+}  // end namespace functor
+
```
rmlarsen: Get rid of empty line.

tensorflow/core/kernels/histogram_op_gpu.cu.cc
```diff
@@ -0,0 +1,123 @@
...
+        pinned_allocator));
+    auto levels = levels_tensor.flat<T>();
+
+    const double step = static_cast<double>(value_range(1) - value_range(0)) /
+                        static_cast<double>(nbins);
+    double curr = static_cast<double>(value_range(0)) + step;
+    levels(0) = std::numeric_limits<T>::lowest();
+    for (int i = 1; i < nbins; i++) {
+      levels(i) = T(curr);
+      curr += step;
```
rmlarsen: Summing steps in curr is not as numerically stable as setting 

   levels(i) = T(static_cast<double>(value_range(0))  + (i + 1) * step);

since round-off may accumulate if nbins is large.

tensorflow/core/kernels/histogram_op_gpu.cu.cc
```diff
@@ -0,0 +1,123 @@
...
+
+    size_t temp_storage_bytes = 0;
+    const T* d_samples = values.data();
+    Tout* d_histogram = out.data();
+    int num_levels = levels.size();
+    T* d_levels = levels.data();
+    int num_samples = values.size();
+    const cudaStream_t& stream = GetCudaStream(context);
+
+    auto err = cub::DeviceHistogram::HistogramRange(
```
rmlarsen: Comment that this call is to get size of temp storage.

tensorflow/core/kernels/histogram_op_gpu.cu.cc
```diff
@@ -0,0 +1,123 @@
...
+        /* d_temp_storage */ NULL,
+        /* temp_storage_bytes */ temp_storage_bytes,
+        /* d_samples */ d_samples,
+        /* d_histogram */ d_histogram,
+        /* num_levels */ num_levels,
+        /* d_levels */ d_levels,
+        /* num_samples */ num_samples,
+        /* stream */ stream);
+    if (err != cudaSuccess) {
+      return errors::Internal("Could not launch HistogramFixedWidthKernel: ",
```
rmlarsen: Different error messages to distinguish between the two calls?

tensorflow/core/ops/math_ops.cc
```diff
@@ -2250,6 +2250,44 @@ product: Pairwise cross product of the vectors in `a` and `b`.
...
+    .Attr("Tout: {int32, int64} = DT_INT32")
+    .SetShapeFn([](InferenceContext* c) {
+      c->set_output(0, c->UnknownShapeOfRank(1));
+      return Status::OK();
+    })
+    .Doc(R"doc(
+Return histogram of values.
+
+Given the tensor `values`, this operation returns a rank 1 histogram counting
+the number of entries in `values` that fell into every bin.  The bins are
```
rmlarsen: nit^3: s/fell/fall/

rmlarsen (changes requested):
Thanks for the contribution! My comments are mostly nits.

rmlarsen:

tensorflow/core/kernels/histogram_op.cc
```diff
@@ -0,0 +1,136 @@
...
+template <typename T, typename Tout>
+struct HistogramFixedWidthFunctor<CPUDevice, T, Tout> {
+  static Status Compute(OpKernelContext* context,
+                        const typename TTypes<T, 1>::ConstTensor& values,
+                        const typename TTypes<T, 1>::ConstTensor& value_range,
+                        int32 nbins, typename TTypes<Tout, 1>::Tensor& out) {
+    const CPUDevice& d = context->eigen_device<CPUDevice>();
+
+    Tensor temp_tensor;
+    TF_RETURN_IF_ERROR(context->allocate_temp(DataTypeToEnum<int32>::value,
```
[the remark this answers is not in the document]
yongtang: Thanks @rmlarsen. Updated.

tensorflow/core/kernels/histogram_op.cc
```diff
@@ -0,0 +1,136 @@
...
+                                              TensorShape({values.size()}),
+                                              &temp_tensor));
+    auto temp = temp_tensor.flat<int32>();
```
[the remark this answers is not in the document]
yongtang: Thanks. Done.

tensorflow/core/kernels/histogram_op.cc
yongtang: Thanks. Done.

tensorflow/core/kernels/histogram_op.cc
yongtang: Done.

tensorflow/core/kernels/histogram_op.h
yongtang: Done. Removed.

tensorflow/core/kernels/histogram_op.h
yongtang: Done.

tensorflow/core/kernels/histogram_op.h
yongtang: Done.

tensorflow/core/kernels/histogram_op_gpu.cu.cc
yongtang: Updated. Thanks!

tensorflow/core/kernels/histogram_op_gpu.cu.cc
yongtang: Done.

tensorflow/core/kernels/histogram_op_gpu.cu.cc
yongtang: Done. Updated.

tensorflow/core/ops/math_ops.cc
yongtang: Done.

yongtang:
Thanks @rmlarsen for the review 👍 . The PR has been updated. Please take a look and let me know if there are any issues.

rmlarsen (approved):

rmlarsen:

yongtang:
The only failed test is on Mac OS X of `//tensorflow/contrib/distributions/poisson_lognormal_test`. I would assume this is unrelated?

rmlarsen:
@yongtang Yes, the failing test appears to be unrelated. Merging. Thanks for your contribution!

gunan:
I think the "poisson_lognormal_test" is broken due to this change.
```
TypeError: Expected int for argument 'nbins' not <tf.Tensor 'histogram/ToInt32:0' shape=() dtype=int32>.
```

https://source.cloud.google.com/results/invocation/28566b59-d961-458c-8978-b16dc248782f/%2F%2Ftensorflow%2Fcontrib%2Fdistributions:poisson_lognormal_test?page=log

I now also see these failures in our continuous builds. Could you take a look?

@vrv @hpucha looking at the state of things in the morning, we may want to revert this change temporarily to unblock everyone else.

yongtang:
@gunan Sorry for the breaks. I have created a PR #13776 that revert the change.

I will take a look at the failing test shortly.

yongtang:
Think I know where the issue is. Before this PR, the `tf.histogram_fixed_width` is only available in Python (no C++ for CPU or GPU). The python version of `tf.histogram_fixed_width` has a default value of `nbins=100`.

This PR tries to preserve the `nbins=100` to keep API backward-compatibility. As a result this PR uses attribute instead of `input` tensor to carry nbins. (You could specify default value `=100` with attribute).

But that somehow breaks the test utility code in `tensorflow/contrib/distributions/python/ops/test_util.py` where a tensor is passed in `nbins`.

The fix might be quite tricky. Need to preserve api compatibility and also pass a tensor. Let me think if there is a way to fix the issue without breaking one way or another.

The  PR #13776 is the revert. It could be merged first for a quick solution.

yongtang:
Think I could hide the `HistogramFixedWidth` in `hidden_ops.txt` and add wrapper, so that both api compatibility and test utility in contrib could be preserved. Will create the PR shortly.

PR #13732, opened 2017-10-15
Update CUB to 1.7.4

yongtang:
Was working on #13731 and noticed that the CUB version is 1.7.3.

This fix updates CUB to 1.7.4 which consists of bug fixes in radix sort.

### DIFF
--- a/tensorflow/contrib/cmake/external/cub.cmake
+++ b/tensorflow/contrib/cmake/external/cub.cmake
@@ -14,8 +14,8 @@
 # ==============================================================================
 include (ExternalProject)
 
-set(cub_URL http://mirror.bazel.build/github.com/NVlabs/cub/archive/1.7.3.zip)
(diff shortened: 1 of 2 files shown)


caisq:

caisq (approved):

mrry (changes requested):
Can you please apply the same change to [`cub.cmake`](https://github.com/tensorflow/tensorflow/blob/cffe603e93520318aaaf07c165cfaf1a3f13805a/tensorflow/contrib/cmake/external/cub.cmake#L17) so that Windows users get the benefits of the bug fix? Thanks!

yongtang:
Thanks @mrry @caisq. I have updated the PR with the CMAKE file changes. I don't have a Windows dev machine readily available though I tested with CMAKE Linux build and it works.

mrry (approved):
CMake changes LGTM, thanks!

caisq:

Issue #13449, opened 2017-10-02
[FeatureRequest] TFDBG change default limit depth for recursive checking tensors


kcyu1993:
Hi,

Recently, I was debugging with TFDBG for a neural network training. It is pretty useful, however, when I want to check some ill-conditioned tensors and their parent, by default TFDBG give me its parents up to a recursive level of 20, i.e. limit depth=20 by default. I could solve the problem by putting a "-d 1" to check its immediate parent each time but it is quite annoying to input this every time.

Would you consider adding a feature in this TFDBG allowing us to set a default value for recursive retrieval of a tensor's input nodes?

My TensorFlow version is
```('v1.3.0-rc2-20-g0787eee', '1.3.0')```

Here is an example of tracing a node's inputs when running the TFDBG currently 

``` 
Inputs to node "node name" (Depth limit = 20, controlled input included )
...
... some detail information
...
```
for command  ```>tfdbg li -c -r "node name" ```

caisq:
Thanks for the feature request @kcyu1993 . I'll try to get that implemented at HEAD this week.

PR #13736, opened 2017-10-16
tfdbg: add persistent config (cherry-pick to r1.4)

caisq:
* Add two persistent UI configurations backed by a file at ~/.tfdbg_config by default.
  * graph_recursion_depth, which controls the recursive output of li/lo commands.
  * mouse_mode, which controls the mouse state of the CursesUI.
* Add `config` command to set and inspect the persistent configuration. E.g.,
  * config show
  * config set graph_recursion_depth 3
  * config set mouse_mode False

Fixes: #13449
PiperOrigin-RevId: 172270804


gunan:
The new cherrypick policy is we only accept critical bugfixes or documentation updates into the release.
As the issue looks like a feature request, I will close the PR.
We can discuss internally whether to have this cherrypick, if you think we definitely need it, let's start an email thread.

PR #13748, opened 2017-10-16
Utility class for outputing/formatting generated source code

karllessard:
The following class is used in an upcoming PR for generating Java operation wrappers but is generic and useful enough to take place in the core library. This way, any source code generated from C++ can make use of it.


asimshankar:
For now, let's not include this in core but instead include it only with the Java code generator.
If and when we decide that the C++/Python code generator should use this too, we can move it then.

Sound fair?

karllessard:
It is ok with me but just to clarify why I thought it has its place in the core:

I'm already planning to have a JavaWriter which decorates this generic interface with Java specific features. I'd like to keep them in two separated classes, it is just a bit strange having both in the same package (but no big deal here). Also, I wanted to make sure its existence would be noticed by developers of the other clients better than if it's "hidden" in the java package.

That being said, if we decide to move this in the java package, I guess it would be easier to close this PR and push it with the upcoming Java stuff?

asimshankar:
Thanks for your understanding. Yeah, I think for now let's close this PR and push it with the Java stuff.

(I feel that any generalization in location or on structure of the code should occur only when needed - or at least when there is someone who is actively interested in integration. Since we don't have that yet for the C++/Python code generators, let's leave it in the Java dir and not worry about more generalization than required. That said, I'm okay with your proposed class structure)

Thanks!

karllessard:
Understood and agreed, thanks!

PR #13753, opened 2017-10-16
Fix dev remnants 1.4

av8ramit:


vrv:
These all look good, no tests needed since it's just metadata files.

PR #13755, opened 2017-10-16
Allow num parameter in tf.linspace to be int64.

codrut3:
According to the API, tf.linspace is defined for num int32 or int64.
However the C++ kernel only allows int32, even though the op in core/ops/math_ops permits int64 too.
So I slightly changed the kernel to allow int64 too. I also added tests for RangeOp and LinSpaceOp.

The following code shows the issue:

```
import numpy as np
import tensorflow as tf

with tf.device("/cpu:0"):
    step = tf.constant(1, tf.int64)
    x = tf.linspace(10.0, 12.0, step)

with tf.Session() as sess:
   x_np = sess.run(x)
   print(x_np)
```

### DIFF
--- a/tensorflow/core/kernels/sequence_ops_test.cc
+++ b/tensorflow/core/kernels/sequence_ops_test.cc
@@ -1,0 +1,148 @@
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "tensorflow/core/framework/allocator.h"
+#include "tensorflow/core/framework/fake_input.h"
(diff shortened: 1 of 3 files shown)


tensorflow/core/kernels/sequence_ops_test.cc
```diff
@@ -0,0 +1,148 @@
...
+#include "tensorflow/core/kernels/ops_testutil.h"
+#include "tensorflow/core/kernels/ops_util.h"
+#include "tensorflow/core/platform/test.h"
+
+namespace tensorflow {
+namespace {
+
+class RangeOpTest : public OpsTestBase {
+ protected:
+  void MakeOp(DataType variable_ref_type) {
```
vrv: Is there a better name than "variable_ref_type" here?  Should this be something like "Index type" ?

tensorflow/core/kernels/sequence_ops_test.cc
```diff
@@ -0,0 +1,148 @@
...
+                     .Input(FakeInput(variable_ref_type))
+                     .Input(FakeInput(variable_ref_type))
+                     .Finalize(node_def()));
+    TF_ASSERT_OK(InitOp());
+  }
+};
+
+class LinSpaceOpTest : public OpsTestBase {
+ protected:
+  void MakeOp(DataType variable_ref_type, DataType num_type) {
```
vrv: How about 'index_type' instead of  'num_type'?

vrv:
Overall looks great, just a minor variable naming nit, and then I can kick off tests :)

codrut3:
I changed the variable names, I hope it's fine now. Please let me know if there is anything else I should do.

vrv:
Great!

vrv (approved):

PR #13758, opened 2017-10-16
Fix ambiguous type comparison in s3_crypto.cc

case540:
We were seeing the following compilation error on Windows builds.

tensorflow/contrib/s3/s3_crypto.cc(74): error C2666:
'std::fpos<_Mbstatet>::operator ==': 3 overloads have similar conversions
could be 'bool std::fpos<_Mbstatet>::operator ==(std::streamoff) const'
or 'bool std::fpos<_Mbstatet>::operator ==(const std::fpos<_Mbstatet> &)


vrv:
Only failure appears to be the batch_matmul_test on windows, which is not a build failure, so merging.

PR #13770, opened 2017-10-17
the comment is confusing

liaolushen:
I find the comment say that embedding shape is [num_encoder_symbols x input_size], I was so confused and checked the source code, I believe it will be better to change the input_size to embedding_size


liaolushen:
I signed it!

vrv (approved):

PR #13781, opened 2017-10-17
Fix #13731 by adding HistogramdFixedWidth in hidden_ops.txt and create the python wrapper

yongtang:
This fix fixes the build breaks in #13731. The issue was caused by the fact that API compatibility requires a default value for `nbins=100` while the test utility code in contrib need to pass a tensor (not through attribute).

This fix adds `HistogramdFixedWidth` in hidden_ops.txt so that `histogram_fixed_width` is hidden in `gen_math_ops.py`. Then `histogram_fixed_width` calls the hidden `gen_math_ops._histogram_fixed_width`.

In this way, both api compatibility and test utility code in contrib requires no changes.

See (comment) https://github.com/tensorflow/tensorflow/pull/13731#issuecomment-337186002 for reference.


gunan:
Let's try if this fixes the problem.
Jenkins, test this please.

vrv (approved):

PR #13800, opened 2017-10-18
Fix MPI and Verbs compilation when not using GPUs

jbedorf:
Code compilation would fail when building without CUDA support, but with MPI and/or Verbs support. This PR adds the required dependencies directly into the MPI/Verbs BUILD files and fixes those compilation errors.


vrv:
Do you know what the compilation error says?  Just want to verify that this is the right dependency to add.

jbedorf:
This is the error:
```
ERROR: /home/jbedorf/codes/tensorflow/tensorflow/contrib/mpi/BUILD:60:1: C++ compilation of rule '//tensorflow/contrib/mpi:mpi_rendezvous_mgr' failed (Exit 1): gcc failed: error executing command 
(....)
In file included from ./tensorflow/core/platform/stream_executor.h:26:0,
                 from ./tensorflow/core/common_runtime/gpu/gpu_util.h:23,
                 from tensorflow/contrib/mpi/mpi_rendezvous_mgr.cc:30:
./tensorflow/stream_executor/dso_loader.h:32:30: fatal error: cuda/cuda_config.h: No such file or directory
 #include "cuda/cuda_config.h"
                              ^
compilation terminated.
```

vrv:
Thanks!  So it seems as though since mpi_rendezvous_mgr depends on gpu_util directly, it should depend on the library that is exported by gpu_util.

gpu_util looks to be exported by both core:gpu_runtime and core:gpu_runtime_impl, where the latter has a dependency on :stream_executor already.

Perhaps the bug that needs to be fixed is that core:gpu_runtime should depend on stream_executor, since one of its headers includes stream_executor.  Can you give that a shot?

vrv:
cc @allenlavoie since it looks like this split might have been done by him at some point and may have more context.

allenlavoie:
Adding a stream_executor dependency to gpu_runtime should be fine. stream_executor is a header-only rule and has a separate implementation rule (all these rules are split for the changes described in https://github.com/tensorflow/tensorflow/commit/5c7f9e31). May have just been an omission when I did the split.

jbedorf:
@vrv Adding the dependency to core:gpu_runtime also resulted in a successful built. The PR has been updated.

allenlavoie (approved):

vrv:

PR #13808, opened 2017-10-18
add swish activation function to python ops

lakshayg:
Swish activation function `swish(x) = x * sigmoid(x)` has
just been published by the Google Brain team. This is a
simple implementation of the same in TF's python interface.

Ref: https://export.arxiv.org/pdf/1710.05941

### DIFF
--- a/tensorflow/python/ops/math_ops.py
+++ b/tensorflow/python/ops/math_ops.py
@@ -50,6 +50,7 @@ See the @{$python/math_ops} guide.
 @@asinh
 @@acosh
 @@atanh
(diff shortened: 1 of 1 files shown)


vrv:
As the paper says, we already have a version to check in.  It's only slightly more complicated because you want to avoid keeping two activation buffers worth of memory for backprop by recomputing it in the backward pass.  I'll make sure @PrajitR gets his change in soon ;)

lakshayg:
Oh, I missed that part. Thanks for pointing this out :)

PR #13812, opened 2017-10-18
Add missing `uint16` type registration for ops `CropAndResize`/`CropAndResizeGradBoxes`

yongtang:
This fix adds missing `uint16` type registration in `image_ops.cc` for `CropAndResize` and `CropAndResizeGradBoxes`.

The kernel of `uint16` is available for `CropAndResize` and `CropAndResizeGradBoxes` though it is missing in `image_ops.cc`. This fix addresses this issue.

This fix also adds incomplete test cases for `CropAndResize`


vrv:
Nice!  It looks like far fewer kernels are registered for Grad though, and specifically only floating point values are supported on GPU, but this seems okay.

vrv (approved):

PR #13816, opened 2017-10-18
Disable flaky tests in cmake build.

gunan:


vrv (approved):
Are any of these actually being looked at? :(

gunan:
Jenkins, test this please.

vrv:
(The disabled tests; the list is super long now and doesn't seem to be shrinking much).

gunan:
Current test failure is a known issue, merging.

We can discuss what we can do about the disabled tests further.
Every now and then I revisit these and try them on my machine, and enable them like this:
https://github.com/gunan/tensorflow/commit/ac6442ee5227693a8a0f23cb52b821e54e777b54#diff-154d7a6e7930f8e143fbd4b8329c67d0
But we need more than what I can do by myself.

vrv:
(I can merge if you'd like?)

PR #13817, opened 2017-10-18
allow to run configure from a parent workspace

### LINKED ISSUE #12761: Allow to build tensorflow as a bazel external dependency.
It would be nice to be able to just add a @org_tensorflow in bazel and be able to build it as a dependency.  

At the best of my knowledge the syntaxnet dockerfile is the best example on having tensorflow as a submodule.
https://github.com/tensorflow/models/blob/c259259299db3b486ccdfd6cdec44b884623053a/syntaxnet/Dockerfile#L63

But still building it is not straightforward and probably does not work if you activate CUDA/XLA/MLK because those depends on other bazel subprojects.

Are there any plans to support this use case or I am missing something in the best practices to build tensorflow as a subpackage ?

Mistobaan:
With this patch you can configure tensorflow directly using bazel.

Fixes #12761 

```
export TF_NEED_CUDA=1
export TF_CUDA_VERSION=8.0
export TF_CUDA_COMPUTE_CAPABILITIES=6.0
export TF_CUDNN_VERSION=6
export CUDNN_INSTALL_PATH=/usr/lib/x86_64-linux-gnu
export PYTHON_BIN_PATH=`which python2`
export TF_NEED_GCP=1
export TF_NEED_HDFS=0
export TF_ENABLE_XLA=1
export TF_NEED_VERBS=1
export CC_OPT_FLAGS="-mavx -msse4.2 -mfpmath=both -DEIGEN_USE_VML"
export TF_NEED_MKL=1
export TF_DOWNLOAD_MKL=1
bazel run @org_tensorflow//:configure -- --tf_workspace= $(bazel info output_base)/external/org_tensorflow`" --workspace=$(PWD)
```

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -2,5 +2,20 @@ exports_files(
     [
         "LICENSE",
         "ACKNOWLEDGEMENTS",
+        "configure.py", 
+])
+
+py_binary(
+    name = "configure",
+    srcs = [
+        "configure.py",
     ],
+    data = [
+        "//tensorflow/tools/git:gen_git_source.py",
+        "//tools",
+    ],
+    args = [
+        "--workspace=.", 
+        "--tf_workspace=.", 
+    ]
 )
--- a/configure.py
+++ b/configure.py
@@ -32,10 +32,31 @@ except ImportError:
   from distutils.spawn import find_executable as which
 # pylint: enable=g-import-not-at-top
 
-_TF_BAZELRC = os.path.join(os.path.dirname(os.path.abspath(__file__)),
-                           '.tf_configure.bazelrc')
-_TF_WORKSPACE = os.path.join(os.path.dirname(os.path.abspath(__file__)),
-                             'WORKSPACE')
+import argparse
+parser = argparse.ArgumentParser()
+parser.add_argument("--workspace",
+                      type=str,
+                      default=os.path.dirname(os.path.abspath(__file__)),
+                      help="the absolute path to your active bazel workspace.")
+parser.add_argument("--tf_workspace",
+                      type=str,
+                      default=os.path.dirname(os.path.abspath(__file__)),
+                      help="the absolute path to the tensorflow workspace."\
+                          "Use `$(bazel info output_base)/external/org_tensorflow`" \
+                          "if you are importing tensorflow as a module.")
+args = parser.parse_args()
+
+_WORKSPACE = args.workspace
+_TF_WORKSPACE = args.tf_workspace
+
+if not os.path.exists(_WORKSPACE):
+  raise OSError("path not found %s" % _WORKSPACE)
+
+if not os.path.exists(_TF_WORKSPACE):
+  raise OSError("path not found %s" % _TF_WORKSPACE)
+
+
+_TF_BAZELRC = os.path.join(_WORKSPACE, '.tf_configure.bazelrc')
 _DEFAULT_CUDA_VERSION = '9.0'
 _DEFAULT_CUDNN_VERSION = '7'
 _DEFAULT_CUDA_COMPUTE_CAPABILITIES = '3.5,5.2'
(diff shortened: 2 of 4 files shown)


vrv:
@gunan do you know a better reviewer for this?

gunan:
Potentially @yifeif, but she is OOO for a few weeks. I will look into this.

configure.py
```diff
@@ -240,24 +263,24 @@ def setup_python(environ_cp):
   environ_cp['PYTHON_BIN_PATH'] = python_bin_path
 
   # Write tools/python_bin_path.sh
-  with open('tools/python_bin_path.sh', 'w') as f:
+  with open(os.path.join(_TF_WORKSPACE, 'tools/python_bin_path.sh'), 'w') as f:
```
yifeif: Looks like running `bazel run //:configure` from tf repo won't work because of this file. Is there any way around this?

vrv:
Ping for @Mistobaan

vrv:
One more ping for @Mistobaan ?

Mistobaan:
sorry @vrv being swamped with work. I will push an update in the next day

vrv:
Any updates @Mistobaan

Mistobaan:
@vrv I finally got the time to go over this once again.

```bash
bazel run //:configure -- --tf_workspace=$(PWD) --workspace=$(PWD)
```
This will write the bazelrc inside the current workspace. 
I haven't find a way to pass the full workspace path to the bazel rule other than explicitly settting them. We might change ./configure to run `bazel` with those parameters.

The nice thing is now importing tensorflow from another module works. See this bareminimum workspace: https://github.com/Mistobaan/sample_tf_bazel_import

gunan:
@yifeif could you take another look?

yifeif (approved):

yifeif:

angersson:
Looks like there's some incompatibility with the way the configure script checks for the TF workspace file (revealed by the failing CIs).

drpngx:
@Mistobaan could you address @angersson comment to fix the build? Also, please pull rebase and push again.

gunan:
Looks like the build issues are still there. But this is something we would like to get in.
Checking with @case540 to see if we already merged an equivalent change.

gunan:
@angersson @case540 
If let's look into merging this change. Did we already merge anything equivalent?
If not, let's either merge this, or move it to internal and merge it internally.

configure.py
```diff
@@ -244,33 +265,47 @@ def setup_python(environ_cp):
...
+      open(bazelrc_path, 'w').close()
 
-  remove_line_with('.bazelrc', 'tf_configure')
-  with open('.bazelrc', 'a') as f:
+  remove_line_with(bazelrc_path, 'tf_configure')
+  with open(bazelrc_path, 'a') as f:
     f.write('import %workspace%/.tf_configure.bazelrc\n')
 
 
+def run_gen_git_source(environ_cp):
```
case540: Dont think this is necessary anymore.

BUILD
```diff
@@ -2,5 +2,20 @@ exports_files(
...
+        "configure.py", 
+])
+
+py_binary(
+    name = "configure",
+    srcs = [
+        "configure.py",
     ],
+    data = [
+        "//tensorflow/tools/git:gen_git_source.py",
```
case540: This might not be a dependency anymore.

case540:
I don't think we have anything like this currently and I think its worth adding. configure script changed a bit so I think this change just needs some small updates.

An alternative would be to write some Bazel repo rule that acts like http_archive https://docs.bazel.build/versions/master/be/workspace.html#http_archive but then immediately runs configure.py . Then people using TF as an external dep would not have to manually run configure. Problem is it would add things to .bazelrc file, and I dont think the things added to .bazelrc would be used until the second time you build.

martinwicke:
@case540 @gunan, so what do we do?

gunan:
@case540 if @Mistobaan is OK with it, could you take over this change?

Mistobaan:
@gunan sure!

tensorflowbutler:
Nagging Assignee @vrv: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

Issue #11673, opened 2017-07-21
Error in tf.contrib.layers.batch_norm when center=False, data_format='NCHW' and zero_debias_moving_mean=True


Kipok:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**:  Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 14.04
- **TensorFlow installed from (source or binary)**: source
- **TensorFlow version (use command below)**:  1.2.1
- **Python version**:  3.4.3
- **Bazel version (if compiling from source)**: 0.5.2
- **CUDA/cuDNN version**:  8.0/6.0
- **GPU model and memory**: Tesla P100 (16 gb)

### Describe the problem
Batch norm layer fails with an error when both center=False, data_format='NCHW' and zero_debias_moving_mean=True arguments are used. It looks like the solution would be just adding additional if to check if beta is None in the same way it is done for gamma, but maybe there are some more dependencies.

### Source code / logs
The following code could be used to reproduce the issue:
```python
import tensorflow as tf
a = tf.placeholder(tf.float32, shape=(10, 10, 10, 10))
b = tf.contrib.layers.batch_norm(a, center=False, data_format='NCHW',
                                 zero_debias_moving_mean=True)
sess = tf.Session()
sess.run(tf.global_variables_initializer())
```
The error log:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/Documents/weight-normalization-exps/venv/lib/python3.4/site-packages/tensorflow/python/framework/op_def_library.py in apply_op(self, op_type_name, name, **keywords)
    489                 as_ref=input_arg.is_ref,
--> 490                 preferred_dtype=default_dtype)
    491           except TypeError as err:

~/Documents/weight-normalization-exps/venv/lib/python3.4/site-packages/tensorflow/python/framework/ops.py in internal_convert_to_tensor(value, dtype, name, as_ref, preferred_dtype)
    675         if ret is None:
--> 676           ret = conversion_func(value, dtype=dtype, name=name, as_ref=as_ref)
    677 

~/Documents/weight-normalization-exps/venv/lib/python3.4/site-packages/tensorflow/python/framework/constant_op.py in _constant_tensor_conversion_function(v, dtype, name, as_ref)
    120   _ = as_ref
--> 121   return constant(v, dtype=dtype, name=name)
    122 

[...]

PR #13829, opened 2017-10-19
Update layers.py

shreyneil:
I have added a check on beta with reference to the following issue raised.
https://github.com/tensorflow/tensorflow/issues/11673

Please verify and get back.
Thank you.

I will add update the unit test cases to verify this patch.

### DIFF
--- a/tensorflow/contrib/layers/python/layers/layers_test.py
+++ b/tensorflow/contrib/layers/python/layers/layers_test.py
@@ -3688,7 +3688,17 @@ class LegacyFullyConnectedTest(test.TestCase):
                                    'rank of x must be at least 2 not: 1'):
         x = constant_op.constant([[]], shape=[0])
         _layers.legacy_fully_connected(x, 2, activation_fn=nn_ops.softmax)
-
-
+  
+class zero_debias_moving_mean(test.TestCase):
+  
+  def Error_in_tf.contrib.layers.batch_norm_when(self):
+    import tensorflow as tf
+    a = tf.placeholder(tf.float32, shape=(10, 10, 10, 10))
+    b = tf.contrib.layers.batch_norm(a, center=False, data_format='NCHW',
+    zero_debias_moving_mean=True)
+    sess = tf.Session()
+    sess.run(tf.global_variables_initializer())
+    
+    
 if __name__ == '__main__':
   test.main()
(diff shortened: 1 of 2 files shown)


vrv (approved):

vrv:

vrv:
Please add the test as requested in the previous version of the CL, otherwise this looks great, thanks!

shreyneil:
https://github.com/tensorflow/tensorflow/pull/13864 contains the updated unit test case.

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -3688,7 +3688,17 @@ def test_1d_invalid_input(self):
                                    'rank of x must be at least 2 not: 1'):
         x = constant_op.constant([[]], shape=[0])
         _layers.legacy_fully_connected(x, 2, activation_fn=nn_ops.softmax)
-
-
+  
+class zero_debias_moving_mean(test.TestCase):
```
vrv: In this case, you can just add the test to the existing BatchNormTest suite, no need to add a new ones.

In general though, you can check the style of other class names in this file; one should use CamelCase test names here if you were adding a new class.

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -3688,7 +3688,17 @@ def test_1d_invalid_input(self):
...
+  
+  def Error_in_tf.contrib.layers.batch_norm_when(self):
```
vrv: test names should start with 'test':

def testBatchNormCenterFalse

tensorflow/contrib/layers/python/layers/layers_test.py
```diff
@@ -3688,7 +3688,17 @@ def test_1d_invalid_input(self):
...
+    import tensorflow as tf
```
vrv: In internal tests, you have to import the modules individually, you cannot import tensorflow here.  So you should use something like:

a = array_ops.placeholder(dtypes.float32, ...)

Take a look at the existing batch norm tests and use that as a guide for how to write the test where center=False

vrv (changes requested):

vrv:
It looks like you may be having trouble updating this PR to include all the changes  -- perhaps it makes sense to resend a new fresh PR with all of the requested changes (layers.py and layers_test.py change in the same PR).  Thanks!

PR #13831, opened 2017-10-19
Fix casting to `size_t` for mkl conv filter dims

jjangsangy:
Here's the output from clang regarding explicit casting `int64` to `size_t`.

```clang
tensorflow/core/kernels/mkl_conv_ops.cc:291:31: error: non-constant-expression cannot be narrowed from type 'int64' (aka 'long long') to 'size_t' (aka 'unsigned long') in initializer list [-Wc++11-narrowing]
```

After modifying those lines I was able to compile successfully for CPU w/MKL + XLA.


vrv:

vrv (approved):

PR #13836, opened 2017-10-19
Add "PlainSessionCreator"

ppwwyyxx:
Not sure if this is desirable, but I found this improves the usability of the `SessionCreator` interface. When I use a factory to create session, it's helpful that it can at least create session in the simplest way.
An example usage would be to write a "train" function that takes a SessionCreator and a bunch of other things.

### DIFF
--- a/tensorflow/python/training/monitored_session.py
+++ b/tensorflow/python/training/monitored_session.py
@@ -25,6 +25,7 @@ import sys
 import six
 
 from tensorflow.core.protobuf import config_pb2
+from tensorflow.python.client import session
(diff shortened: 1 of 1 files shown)


ispirmustafa (commented):
Thank you Yuxin.
PlainSessionCreator is equvelant to ChiefSessionCreator where you left other parameters as default (None).
Let's do not add this functionality.

ppwwyyxx:
Thanks! Didn't notice that due to all the complications in ChiefSessionCreator. Then it makes sense to close this PR.

PR #13846, opened 2017-10-20
Fix `tf.py_func()` and `Dataset.from_generator()` on Python 3.

mrry:
Cherry pick to make `Dataset.from_generator()` work on Python 3 when the generator yields (unicode) strings.

### DIFF
--- a/tensorflow/python/kernel_tests/dataset_from_generator_op_test.py
+++ b/tensorflow/python/kernel_tests/dataset_from_generator_op_test.py
@@ -216,6 +216,26 @@ class DatasetConstructorTest(test.TestCase):
         with self.assertRaises(errors.OutOfRangeError):
           sess.run(get_next)
 
+  def testFromGeneratorString(self):
+    def generator():
+      yield "foo"
+      yield b"bar"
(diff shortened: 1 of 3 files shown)


vrv:

mrry:
Hi folks... I realize this might be past the deadline for rc1, but I thought I'd send a PR just in case.

The bug it fixes is **not** severe enough to be a release blocker, because there's a workaround (i.e. explicitly yield `bytes` not `str` when using Python 3). However, if this PR is too late, would it be possible to update the release notes with a "known issue"?

case540:
This is unfortunately too late for RC1. I can update the release notes however. How would you word the issue..

Is something like this good?

* `Dataset.from_generator()` does not work in Python 3 when the generator yields (unicode) strings.

mrry:
How about "In Python 3, `Dataset.from_generator()` does not support Unicode strings. You must convert any strings to `bytes` objects before yielding them from the generator."

case540:
Sounds good. Will add to Release notes.

mrry:
Thanks!

Issue #13298, opened 2017-09-25
wrong tf.svd documentation in tensorflow 1.3.0 version


pswpswpsw:
In tf.svd documentation, it was said that the tensorflow implementation of svd is simply np.linalg.svd

in fact it was not the same, in tensorflow, the index for V matrix is transposed..different from np.linalg.svd default setting. Funny thing is, this type of problem is a famous bug in matlab code people usually write, which is row selection and transpose is not equal to transpose then column selection.

code demonstration:

```
import tensorflow as tf
tf.__version__

## benchmark using numpy svd
import numpy as np

x = np.array([1.0,2.0,3.0,4.0,5.0,6,7,8,9])
x = x.reshape(3,3)

u_,s_,v_=np.linalg.svd(x,compute_uv=True)
# print s_
print u_
print v_

r = 2
diagS = np.diag(s_)
print diagS
diagS = diagS[:r,:r]
x_rank1 = np.matmul(u_[:,:r], np.matmul(diagS, v_[:r,:]))

print 'original = \n',x
print 'rank1 approximation\n',x_rank1
```
Result is
```
[[-0.21483724  0.88723069  0.40824829]
 [-0.52058739  0.24964395 -0.81649658]
 [-0.82633754 -0.38794278  0.40824829]]
[[-0.47967118 -0.57236779 -0.66506441]
 [-0.77669099 -0.07568647  0.62531805]
 [-0.40824829  0.81649658 -0.40824829]]
[[  1.68481034e+01   0.00000000e+00   0.00000000e+00]
 [  0.00000000e+00   1.06836951e+00   0.00000000e+00]
 [  0.00000000e+00   0.00000000e+00   4.41842475e-16]]
original = 
[[ 1.  2.  3.]
 [ 4.  5.  6.]
 [ 7.  8.  9.]]
rank1 approximation
[[ 1.  2.  3.]
 [ 4.  5.  6.]
 [ 7.  8.  9.]]
```

But in tensorflow svd
```
X = tf.constant([1.0,2.0,3.0,4.0,5.0,6,7,8,9],shape=[3,3])

sess = tf.InteractiveSession()
    
print sess.run(X)

s,u,v = tf.svd(X)

print sess.run(u)
print sess.run(v)

diagS_tf = tf.diag(s)
# print sess.run(diagS_tf)

tmp_correct =  tf.matmul(diagS_tf[:r,:r], tf.transpose(v[:,:r]))
[...]

PR #13850, opened 2017-10-20
Updated tf.svd documentation to state " u * s * v' "

donigian:
Updated tf.svd documentation to state the index for V matrix is transposed.

### DIFF
--- a/tensorflow/python/ops/linalg_ops.py
+++ b/tensorflow/python/ops/linalg_ops.py
@@ -371,8 +371,9 @@ def svd(tensor, full_matrices=False, compute_uv=True, name=None):
   """Computes the singular value decompositions of one or more matrices.
 
   Computes the SVD of each inner matrix in `tensor` such that
-  `tensor[..., :, :] = u[..., :, :] * diag(s[..., :, :]) * transpose(v[..., :,
-  :])`
+  `tensor[..., :, :] = u[..., :, :] * diag(s[..., :, :]) * transpose(conj(v[..., :,
+  :])). ` Note: Currently only support complex SVD on CPU.
+
 
   ```python
   # a is a tensor.
@@ -403,12 +404,13 @@ def svd(tensor, full_matrices=False, compute_uv=True, name=None):
       `[..., M, M]`. Not returned if `compute_uv` is `False`.
     v: Right singular vectors. If `full_matrices` is `False` (default) then
       shape is `[..., N, P]`. If `full_matrices` is `True` then shape is
-      `[..., N, N]`. Not returned if `compute_uv` is `False`.
+      `[..., N, N]`. Not returned if `compute_uv` is `False`. 
 
   @compatibility(numpy)
-  Mostly equivalent to numpy.linalg.svd, except that the order of output
-  arguments here is `s`, `u`, `v` when `compute_uv` is `True`, as opposed to
-  `u`, `s`, `v` for numpy.linalg.svd.
+  Mostly equivalent to numpy.linalg.svd, except that the transposition and 
+  order of output arguments here is `s`, `u`, `v` when `compute_uv` is `True`, 
+  as opposed to `u`, `s`, `v` for numpy.linalg.svd. 
+
   @end_compatibility
   """
   # pylint: disable=protected-access


facaiy:
resolve #13298

rmlarsen:
You change is not correct. V stores the singular vectors as stated. It is correct that A \approx U \Sigma V^H. Notice that this convention is unlike some other SVD libraries that return V^H.

https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/python/kernel_tests/svd_op_test.py?rcl=171836140&l=120

rmlarsen:
Also as stated in 

https://www.tensorflow.org/api_docs/python/tf/svd

tensor[..., :, :] = u[..., :, :] * diag(s[..., :, :]) * transpose(v[..., :, :])

We should perhaps update that to 

tensor[..., :, :] = u[..., :, :] * diag(s[..., :, :]) * transpose(conj(v[..., :, :]))
(although we currently only support complex SVD on CPU)

rmlarsen:
@donigian if you are up for changing the documentation as stated above, it would be helpful. Thanks!

yaroslavvb:
What's needed is a fix to "compatibility" section of the docs, there a difference in order and transposition from numpy result, whereas docs only mentions order difference

```
@compatibility(numpy)
  Mostly equivalent to numpy.linalg.svd, except that the order of output
  arguments here is `s`, `u`, `v` when `compute_uv` is `True`, as opposed to
  `u`, `s`, `v` for numpy.linalg.svd.
  @end_compatibility
```

donigian:
oops @rmlarsen please see revised edits from what I was able to gather.

tensorflow/python/ops/linalg_ops.py
```diff
@@ -403,12 +404,13 @@ def svd(tensor, full_matrices=False, compute_uv=True, name=None):
       `[..., M, M]`. Not returned if `compute_uv` is `False`.
     v: Right singular vectors. If `full_matrices` is `False` (default) then
       shape is `[..., N, P]`. If `full_matrices` is `True` then shape is
-      `[..., N, N]`. Not returned if `compute_uv` is `False`.
+      `[..., N, N]`. Not returned if `compute_uv` is `False`.
```
rmlarsen: Please remove trailing space.

tensorflow/python/ops/linalg_ops.py
```diff
@@ -403,12 +404,13 @@ def svd(tensor, full_matrices=False, compute_uv=True, name=None):
...
+      `[..., N, N]`. Not returned if `compute_uv` is `False`. 
 
   @compatibility(numpy)
-  Mostly equivalent to numpy.linalg.svd, except that the order of output
-  arguments here is `s`, `u`, `v` when `compute_uv` is `True`, as opposed to
-  `u`, `s`, `v` for numpy.linalg.svd.
+  Mostly equivalent to numpy.linalg.svd, except that the transposition and 
+  order of output arguments here is `s`, `u`, `v` when `compute_uv` is `True`, 
+  as opposed to `u`, `s`, `v` for numpy.linalg.svd. 
+
```
rmlarsen: Remove empty line.

tensorflow/python/ops/linalg_ops.py
```diff
@@ -403,12 +404,13 @@ def svd(tensor, full_matrices=False, compute_uv=True, name=None):
...
     v: Right singular vectors. If `full_matrices` is `False` (default) then
       shape is `[..., N, P]`. If `full_matrices` is `True` then shape is
-      `[..., N, N]`. Not returned if `compute_uv` is `False`.
+      `[..., N, N]`. Not returned if `compute_uv` is `False`. 
 
   @compatibility(numpy)
-  Mostly equivalent to numpy.linalg.svd, except that the order of output
-  arguments here is `s`, `u`, `v` when `compute_uv` is `True`, as opposed to
-  `u`, `s`, `v` for numpy.linalg.svd.
+  Mostly equivalent to numpy.linalg.svd, except that the transposition and
```
rmlarsen: "transposition" of outputs does not make sense. How about this:

Mostly equivalent to numpy.linalg.svd, except that order of output arguments here is `s`, `u`, `v` when `compute_uv` is `True`, as opposed to `u`, `s`, `v` for numpy.linalg.svd. Also notice that this function returns `v` as opposed to `transpose(conj(v))`  for numpy.linalg.svd, which does not follow the mathematical definition of the SVD.

rmlarsen (changes requested):

rmlarsen:
@yaroslavvb the numpy api is just wrong :-)

rmlarsen:
@yaroslavvb the numpy api is just "wrong", as explained in the notes:

https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.linalg.svd.html

"The SVD is commonly written as a = U S V.H. The v returned by this function is V.H and u = U."

tensorflow/python/ops/linalg_ops.py
yaroslavvb: Maybe say "numpy.linalg.svd returns v" instead of "this function returns v" to be explicit?

@rmlarsen agree on numpy choosing suboptimal output format

rmlarsen: @yaroslavvb I agree that your formulation is clearer. Thanks.

rmlarsen: ugh actually, I think it should be tf.linalg.svd return v. numpy.linalg.svd returns v^H, but calls the argument "v"... :-)

vrv:
Closing for lack of activity, and it seems @rmlarsen is the expert who knows how this could be clarified properly.

rmlarsen:
FYI. I'll update the docstring internally. Should be pushed out in a day or two. Thanks for bringing up the issue @donigian and @yaroslavvb

Issue #13855, opened 2017-10-20
compile source code fail on mac


orange-jacky:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
macOS Sierra 10.12.6
- **TensorFlow installed from (source or binary)**:
source
- **TensorFlow version (use command below)**: 
1.3.0
- **Python version**: 
Python 2.7.10
- **Bazel version (if compiling from source)**:
macbookpro:tensorflow fredlee$ bazel version
Build label: 0.7.0-homebrew
Build target: bazel-out/darwin_x86_64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Oct 19 09:12:48 2017 (1508404368)
Build timestamp: 1508404368
Build timestamp as int: 1508404368
- **CUDA/cuDNN version**:
no
- **GPU model and memory**:
no 
- **Exact command to reproduce**:

step by step according to this guide
   step1:
       git clone https://github.com/tensorflow/tensorflow
  step2:
      cd tensorflow
      ./configure
  step3:
     create an example as (https://tensorflow.google.cn/api_guides/cc/guide)
  step4:
    bazel run -c opt //tensorflow/cc/example:example

  it return
      

> ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Target //tensorflow/cc/example:example failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 3083.675s, Critical Path: 306.19s
ERROR: Build failed. Not running target.

compile log:
[compile.log.zip](https://github.com/tensorflow/tensorflow/files/1401503/compile.log.zip)

****can you tell me how to make it work?** thanks**

**best wishes.**

allenlavoie:
There are two issues. First, the documentation is out of date (cc_binary -> tf_cc_binary to link in libtensorflow_framework.so). Second, even if it were updated at head, that's the 1.3 version of the documentation (CC @MarkDaoust; I'll send you a CL for both).

drpngx:
Thanks @allenlavoie !

Timmmm:
With `tf_cc_binary` in the `BUILD` file I get:

    ERROR: /..../tensorflow/tensorflow/cc/example/BUILD:1:1: name 'tf_cc_binary' is not defined (did you mean 'cc_binary'?).

allenlavoie:
[...]

PR #13858, opened 2017-10-20
Update the C++ API guide [Doc only, should be in 1.4]

allenlavoie:
- Includes a documentation fix for 1.4 (cc_binary -> tf_cc_binary to avoid undefined symbols).
- Adds the standard warning at the top that people may want the master branch.

Fixes https://github.com/tensorflow/tensorflow/issues/13855.


MarkDaoust (approved):
This looks reasonable to me.

Good idea adding that warning at the top.

PR #13862, opened 2017-10-20
Fix issues where int64 crops could not be passed to batch_to_space.

yongtang:
This fix tries to address the issue where int64 `crops` could not be passed to `batch_to_space` even though both int32 and int64 are specified as supported in the docs (tf.batch_to_space.__doc__)

The reason is that BatchToSpace kernel puts a constraint of int32 to crops data types.

This fix removed the constraint so that int64 `crops` could be supported.

NOTE: Just removing the constraint should work and it is not necessary to add specification to the kernel class template, as `SubtleMustCopyFlat` called in the class already correctly handled both int32 and int64 cases. Besides, other data types (e.g., float or double) will not be passed to the kernel as they are guarded by the specification in `array_ops.cc`.

### DIFF
--- a/tensorflow/core/kernels/batchtospace_op.cc
+++ b/tensorflow/core/kernels/batchtospace_op.cc
@@ -23,7 +23,6 @@ limitations under the License.
 
 #include "tensorflow/core/kernels/spacetobatch_functor.h"
 
-#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
 #include "tensorflow/core/framework/op.h"
 #include "tensorflow/core/framework/op_kernel.h"
 #include "tensorflow/core/framework/register_types.h"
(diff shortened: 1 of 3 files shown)


vrv (approved):

vrv:

tensorflow/core/kernels/batchtospace_op.cc
```diff
@@ -23,7 +23,6 @@ limitations under the License.
 
 #include "tensorflow/core/kernels/spacetobatch_functor.h"
 
-#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
```
vrv: Actually, I think we keep this include here because Eigen defines some #defines that must be defined prior to other headers (sadness).  Is it okay to put it back where it was?

yongtang: Thanks @vrv. Didn't know that as I was just using `clang-format -I` to reformat the .cc file.

The PR has been updated with `"third_party/eigen3/unsupported/Eigen/CXX11/Tensor"` back.
 Please take a look.

vrv:

tensorflow/core/kernels/batchtospace_op.cc
vrv: Yup, thanks :)

PR #13863, opened 2017-10-20
Fix crash when `int64` axis is passed to `tf.reduce_sum`

yongtang:
This fix tries to fix the crash triggered by `int64` axis passed to `tf.reduce_sum`:
```
ubuntu@ubuntu:~/tensorflow2$ (cd && python)
>>> import tensorflow as tf
>>> v = tf.reduce_sum([1,2,3], tf.constant(0, tf.int64))
2017-10-20 15:55:06.993430: F tensorflow/core/framework/tensor.cc:601] Check failed: dtype() == expected_dtype (9 vs. 3)
ubuntu@ubuntu:~/tensorflow2$
```

The issue is caused by the fact that shape inference in `common_shape_fns.cc` only assumes int32 without proper handling of diffent types. In `math_ops.cc` both int32 and int64 are mentioned.

NOTE that this fix does not address the issue that int64 is not supported. To allow int64 axis it is more than adding a template in `ReductionOp` as the type of the axis seems to be decided by some other ways in Eigen. Will investigate that later.

This fix merely fixed the crash so that an error message will return without exit from the python program "No OpKernel was registered to support Op 'Sum' with these attrs".

Still, its worth to at least allow the program not to exit in case an error happens.

### DIFF
--- a/tensorflow/core/framework/common_shape_fns.cc
+++ b/tensorflow/core/framework/common_shape_fns.cc
@@ -1050,22 +1050,44 @@ Status ReductionShape(InferenceContext* c) {
   }
 
   const int32 input_rank = c->Rank(input);
-  std::set<int32> true_indices;
-  auto reduction_indices = reduction_indices_t->flat<int32>();
-  for (int i = 0; i < reduction_indices_t->NumElements(); ++i) {
-    int32 reduction_index = reduction_indices(i);
-    if (reduction_index < -input_rank || reduction_index >= input_rank) {
-      return errors::InvalidArgument("Invalid reduction dimension ",
-                                     reduction_index, " for input with ",
-                                     input_rank, " dimensions.");
-    }
+  std::set<int64> true_indices;
+  if (reduction_indices_t->dtype() == DataType::DT_INT32) {
+    auto reduction_indices = reduction_indices_t->flat<int32>();
+    for (int i = 0; i < reduction_indices_t->NumElements(); ++i) {
+      int32 reduction_index = reduction_indices(i);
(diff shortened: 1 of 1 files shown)


tensorflow/core/framework/common_shape_fns.cc
```diff
@@ -1050,22 +1050,44 @@ Status ReductionShape(InferenceContext* c) {
...
-  for (int i = 0; i < reduction_indices_t->NumElements(); ++i) {
-    int32 reduction_index = reduction_indices(i);
-    if (reduction_index < -input_rank || reduction_index >= input_rank) {
-      return errors::InvalidArgument("Invalid reduction dimension ",
-                                     reduction_index, " for input with ",
-                                     input_rank, " dimensions.");
-    }
+  std::set<int64> true_indices;
+  if (reduction_indices_t->dtype() == DataType::DT_INT32) {
+    auto reduction_indices = reduction_indices_t->flat<int32>();
```
vrv: Can we pull this out into a little helper function templated on the index type and call it like:

if (...type == DT_INT32) {
  helper_function<int32>(...);
} else if (...type == DT_INT64) {
  helper_function<int64>(...);
} else {
  invalid...
}

yongtang: Thanks @vrv. The PR has been updated with a template helper function of `ReductionShapeHelper`. Please take a look.

vrv (approved):

vrv:

(Surprised this never came up before! thanks for the fix :)

PR #13867, opened 2017-10-20
added CMake install targets

Slonegg:
This way you can use cmake to install tensorflow into some location. It is easier to package tensorflow this way.

### DIFF
--- a/tensorflow/contrib/cmake/tf_shared_lib.cmake
+++ b/tensorflow/contrib/cmake/tf_shared_lib.cmake
@@ -94,3 +94,46 @@ endif()
 if(WIN32)
   add_dependencies(tensorflow tensorflow_static)
 endif(WIN32)
+
+install(TARGETS tensorflow
+        RUNTIME DESTINATION bin
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib)
+
+# install necessary headers
+# tensorflow headers
+install(DIRECTORY ${tensorflow_source_dir}/tensorflow/cc/
+        DESTINATION include/tensorflow/cc
+        FILES_MATCHING PATTERN "*.h")
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tensorflow/cc/
+        DESTINATION include/tensorflow/cc
+        FILES_MATCHING PATTERN "*.h")
+install(DIRECTORY ${tensorflow_source_dir}/tensorflow/core/
+        DESTINATION include/tensorflow/core
+        FILES_MATCHING PATTERN "*.h")
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tensorflow/core/
+        DESTINATION include/tensorflow/core
+        FILES_MATCHING PATTERN "*.h")
+install(DIRECTORY ${tensorflow_source_dir}/tensorflow/stream_executor/
+        DESTINATION include/tensorflow/stream_executor
+        FILES_MATCHING PATTERN "*.h")
+# google protobuf headers
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/protobuf/src/protobuf/src/google/
+        DESTINATION include/google
+        FILES_MATCHING PATTERN "*.h")
+# nsync headers
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/external/nsync/
+        DESTINATION include/external/nsync
+        FILES_MATCHING PATTERN "*.h")
+# Eigen directory
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/eigen/src/eigen/Eigen/
+        DESTINATION include/Eigen)
+# external directory
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/external/eigen_archive/
+        DESTINATION include/external/eigen_archive)
+# third_party eigen directory
+install(DIRECTORY ${tensorflow_source_dir}/third_party/eigen3/
+        DESTINATION include/third_party/eigen3)
+# unsupported Eigen directory
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/eigen/src/eigen/unsupported/Eigen/
+        DESTINATION include/unsupported/Eigen)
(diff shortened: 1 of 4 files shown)


tensorflow/contrib/cmake/tf_shared_lib.cmake
```diff
@@ -94,3 +94,42 @@ endif()
...
+        RUNTIME DESTINATION bin
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib)
+
+# install headers
+get_property(include_dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
+# remove root directories, since they contain everything
+list(REMOVE_DUPLICATES include_dirs)
+list(REMOVE_ITEM include_dirs ${tensorflow_source_dir} ${CMAKE_CURRENT_BINARY_DIR})
+# append core tensorflow header dirs and binary dir where generated headers reside
```
mrry: I think this is a superset of the public headers—or at least, I can't see where the filtering happens. Can you try to restrict it to the same set of headers as is installed with the PIP package? (AIUI, you'll also need to install some of the protobuf public headers, if they aren't already installed.)

tensorflow/contrib/cmake/tf_tools.cmake
```diff
@@ -147,3 +147,8 @@ target_link_libraries(${benchmark_model} PUBLIC
   ${tf_core_gpu_kernels_lib}
   ${tensorflow_EXTERNAL_LIBRARIES}
 )
+
+install(TARGETS ${proto_text} ${transform_graph} ${summarize_graph} ${compare_graphs} ${benchmark_model}
```
mrry: I'm not sure to what extent we'd consider these tools part of the public package. The `proto_text` tool is certainly for internal build use only; @petewarden would probably know better about the other ones.

mrry (changes requested):

tensorflow/contrib/cmake/tf_tools.cmake
Slonegg: I was going to just create an option tensorflow_BUILD_TOOLS in separate PR and build these tools conditionally.

tensorflow/contrib/cmake/tf_shared_lib.cmake
Slonegg: Thanks @mrry for the quick review. Since I was not quite sure which headers to install and it probably depends on the CMake options I've created universal solution which just installs all the headers used by tensorflow package with same directory structure. This way you also don't need to worry about adjusting this code when tensorflow package is changed. If you confirm that it is unlikely that more packages will be added to public headers set, then I'll try to restrict amount of headers installed. I know for sure that Eigen, Protobuf, nsync and protobuf generated headers are needed to compile trivial C++ program using tensorflow.

vrv:
Friendly ping for updates.

Slonegg:
@vrv I'm waiting for review from @petewarden and @mrry. Solution works, I can provide logs of tensorflow build and example program build.

tensorflow/contrib/cmake/tf_shared_lib.cmake
mrry: It's not unlikely that more headers will be added to the set of public headers in future changes, but the PIP package generated by the Bazel build is a reliable source of truth for what headers need to be included. (IIRC @av8ramit was looking into including these headers in the CMake build at one point, so perhaps he has something you could reuse?)

tensorflow/contrib/cmake/tf_tools.cmake
mrry: Right, but `proto_text` is needed *unconditionally* by the rest of the build process. As far as I know, it has no other uses, so it shouldn't be installed when we install TensorFlow.

tensorflow/contrib/cmake/tf_shared_lib.cmake
av8ramit: https://github.com/tensorflow/tensorflow/commit/fd584b5e9daf57c827b0ee652317ab0f205ae4bb#diff-10f12827f7259a4fd7dba7abbd8307b0 Hopefully this helps?

mrry: Oh sweet, I didn't know that was already in! (I think it might need to be updated for nsync, but apart from that it's just what we need.)

Slonegg: Thanks @av8ramit it was indeed helpful! @mrry I've added install targets based on the @av8ramit solution, needed to add some extra headers from tensorflow/cc and nsync.

tensorflow/contrib/cmake/tf_tools.cmake
Slonegg: Removed proto_text install target.

PinkySan:
What about porting the examples (mainly tf_tutorials_example_trainer and label_image) from an in-source-example to an installing-example? (Providing an example based on the installed tensorflow)

mrry (approved):
Looks good to me! Thanks for the contribution :).

Slonegg:
@PinkySan I think it is a good idea, though I want to merge this PR and few more pressing changes back to upstream first.

PinkySan:
@Slonegg What about providing an additional find_package module? It will be easier integrating tensorflow in existing cmake projects. (sorry for adding additional requrements. I was really looking forward to a PR like this. Good work!)

Slonegg:
@PinkySan I need to fix few things in tensorflow, so we can use it in our application.
1. Add ability to use external GRPC, zlib, Eigen and mb some other packages. Because right now it is impossible to compile your project with tensorflow if you're using different version of GRPC. I did something similar to GRPC package whether you specify tensorflow_<LIB>_PROVIDER variable. It will determine if tensorflow needs to download dependency and use it as a module or use find_package(<LIB>) to find <LIB> package.
2. Add tensorflow_BUILD_STATIC_LIB variable to create static library. With all kernels and no extra functionality tensorflow is several gigs in size. It's too heavy to deploy it as a shared library in microservice application. While if we link static tensorflow library, our final executable hopefully will be smaller.

As for what you are suggesting, I think tensorflow config will suffice. You will need to specify tensorflow_DIR variable to use find_package(tensorflow). We don't use find_package functionality in our projects a lot, since we're using conan package manager. And in conan recipe you can specify all necessary include dirs, libraries, etc. But in general, yes, package config or find module is desired.

mrry:
All of these sounds like great future contributions! Let's get this PR in as a starting point though :).

PR #13871, opened 2017-10-20
Remove deprecated 32bit IOS builds

powderluv:
Apple has stopped supporting 32bit iOS builds staring with iOS11.
Building these binaries are pretty much useless on iOS, but saves
a lot of time by just building arm64 and x86_64.

TEST: build ios libraries with ./build_all_ios.sh

### DIFF
--- a/tensorflow/contrib/makefile/compile_ios_protobuf.sh
+++ b/tensorflow/contrib/makefile/compile_ios_protobuf.sh
@@ -73,36 +73,6 @@ fi
 
 make distclean
 ./configure \
---host=i386-apple-${OSX_VERSION} \
---disable-shared \
---enable-cross-compile \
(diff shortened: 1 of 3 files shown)


vrv:
Pete perhaps you might know more here :)

petewarden:
Thanks for the PR! Unfortunately we still have a lot of internal and external apps that are shipping for iOS 10, and want to support old devices. I'd love to see if there was a clean way to make these architectures optional though, since you're right that for many developers building them is a waste of time.

powderluv:
@petewarden I have another PR #13920  which will allow building just the architecture you want. I will update it so the default builds all the current libs.

powderluv:
New PR #13959 is up which builds all the architectures by default or you can pass -a "arm64" for just one arch.

PR #13874, opened 2017-10-21
Update node_def.proto comments

tayo:
The device field had outdated comments.

Note: We could consider adding tpu as an example here, e.g. "gpu" | "cpu" | "tpu".  Thoughts?

### DIFF
--- a/tensorflow/core/framework/node_def.proto
+++ b/tensorflow/core/framework/node_def.proto
@@ -35,7 +35,7 @@ message NodeDef {
   // CONSTRAINT ::= ("job:" JOB_NAME)
   //              | ("replica:" [1-9][0-9]*)
   //              | ("task:" [1-9][0-9]*)
-  //              | ( ("gpu" | "cpu") ":" ([1-9][0-9]* | "*") )
+  //              | ("device:" ("gpu" | "cpu") ":" ([1-9][0-9]* | "*") )
   //
   // Valid values for this string include:
   // * "/job:worker/replica:0/task:1/device:GPU:3"  (full specification)


tensorflow/core/framework/node_def.proto
```diff
@@ -35,7 +35,7 @@ message NodeDef {
   // CONSTRAINT ::= ("job:" JOB_NAME)
   //              | ("replica:" [1-9][0-9]*)
   //              | ("task:" [1-9][0-9]*)
-  //              | ( ("gpu" | "cpu") ":" ([1-9][0-9]* | "*") )
+  //              | ("device:" ("gpu" | "cpu") ":" ([1-9][0-9]* | "*") )
```
vrv: Technically it can be /device:(^:)* (anything but ":"), but probably we can just say A-Za-z to keep things simple.

vrv:

tensorflow/core/framework/node_def.proto
tayo: Ack'd, sorry for delay vrv@.  We can correct this with another PR if you'd like

vrv:
Feel free to send an update :)

tadeegan:
Hi. Was this change announced in the release notes anywhere?

tayo:
Hi,

I did not see this mentioned in relnotes for 1.4.

vrv:
This just updated a stale comment, /device:FOO has been supported in the core codebase well before 1.0, though not all libraries have handled this form of the device name until more recently.

/gpu:0 and /cpu:0 will remain for the foreseeable future due to backwards compatibility, but new devices must use /device:$XPU:$id.

PR #13878, opened 2017-10-21
protobuf lib path bug fix for benckmark on osx

formath:
### bug info
If not fixed, `/usr/local/lib` will before `tensorflow/contrib/makefile/gen/protobuf-host/lib/` in the library search path. It will cause `undefined symbol` problem of protobuf when building `benchmark` because of incompatibility of system installed protobuf.

### fix
Have tested on maxOS 10.12.6

@vrv

### DIFF
--- a/tensorflow/contrib/makefile/Makefile
+++ b/tensorflow/contrib/makefile/Makefile
@@ -190,6 +190,10 @@ LIBFLAGS :=
 
 # If we're on OS X, make sure that globals aren't stripped out.
 ifeq ($(TARGET),OSX)
+ifeq ($(HAS_GEN_HOST_PROTOC),true)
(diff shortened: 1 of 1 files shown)


formath:
I signed it!

vrv:
Thanks, do you know if this is specific to Mac OS X or does it apply more generally?

formath:
Linux has already set this so has no problem. I have also tested IOS and its works right. Others, e.g, Android and PI, I don't have those environment so have no test.

vrv:

vrv (approved):

vrv:
Unrelated failures.

qichaotang:
view

PR #13884, opened 2017-10-21
Add `int64` type `multiples` support for `tf.tile`

yongtang:
In the doc of `tf.tile` (tf.tile.__doc__) both `int32` and `int64` are supported for `multiples`. However, the kernel for `int64` is not registered yet.

This fix adds the support of `int64` `multiples` so that the behavior matches the description of the docs.

### DIFF
--- a/tensorflow/core/kernels/tile_functor.h
+++ b/tensorflow/core/kernels/tile_functor.h
@@ -17,6 +17,7 @@ limitations under the License.
 #define TENSORFLOW_KERNELS_TILE_FUNCTOR_H_
 
 #include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
 #include "tensorflow/core/framework/tensor.h"
 #include "tensorflow/core/framework/tensor_types.h"
 #include "tensorflow/core/platform/types.h"
@@ -29,13 +30,13 @@ namespace internal {
 template <typename Device, typename T>
 void TileSimple(const Device& d, Tensor* out, const Tensor& in);
 
-template <typename Device, typename T, int NDIM>
+template <typename Device, typename Tmultiples, typename T, int NDIM>
 void TileUsingEigen(const Device& d, Tensor* out, const Tensor& in,
-                    const gtl::ArraySlice<int32>& broadcast_array) {
+                    const gtl::ArraySlice<Tmultiples>& broadcast_array) {
   auto x = in.tensor<T, NDIM>();
   auto y = out->tensor<T, NDIM>();
 
-  Eigen::array<int32, NDIM> b;
+  Eigen::array<Tmultiples, NDIM> b;
   for (int i = 0; i < NDIM; ++i) b[i] = broadcast_array[i];
   if (Eigen::internal::is_same<Device, Eigen::GpuDevice>::value) {
     // Use 32bit indexing to speed up the computations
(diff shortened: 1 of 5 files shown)


tensorflow/core/kernels/tile_functor.h
```diff
@@ -29,13 +30,13 @@ namespace internal {
 template <typename Device, typename T>
 void TileSimple(const Device& d, Tensor* out, const Tensor& in);
 
-template <typename Device, typename T, int NDIM>
+template <typename Device, typename Tmultiples, typename T, int NDIM>
```
vrv: Minor nit: it seems like T should come before Tmultiples since we usually put the 'main datatype' first and then the auxilliary one.

tensorflow/python/kernel_tests/shape_ops_test.py
```diff
@@ -411,14 +411,17 @@ def testScalar(self):
...
   def testSimple(self):
-    with self.test_session():
-      inp = np.random.rand(4, 1).astype(np.float32)
-      a = constant_op.constant(inp)
-      tiled = array_ops.tile(a, [1, 4])
-      result = tiled.eval()
-    self.assertEqual(result.shape, (4, 4))
-    self.assertEqual([4, 4], tiled.get_shape())
-    self.assertTrue((result == np.tile(inp, (1, 4))).all())
+    for use_gpu in False, True:
```
vrv: You can just do use_gpu=True; if a GPU is not available the test will run on CPU, and our test infrastructure runs all tests on CPU and GPU machines.

vrv (commented):
Overall looks good, one just one minor ordering suggestion

tensorflow/core/kernels/tile_functor.h
yongtang: Thanks @vrv, the order of the typename has been rearranged.

tensorflow/python/kernel_tests/shape_ops_test.py
yongtang: Thanks @vrv. Done.

yongtang:
Thanks @vrv for the review. The PR has been updated. Please take a look.

vrv:

vrv (approved):

tensorflow/core/kernels/tile_ops.cc
```diff
@@ -505,127 +508,73 @@ REGISTER_KERNEL_BUILDER(Name("Tile")
...
+  REGISTER_KERNEL_BUILDER(Name("Tile")                             \
+                              .Device(DEVICE_GPU)                  \
+                              .TypeConstraint<type>("T")           \
+                              .TypeConstraint<int64>("Tmultiples") \
+                              .HostMemory("multiples"),            \
+                          TileOp<GPUDevice, int64>);               \
+  REGISTER_KERNEL_BUILDER(Name("TileGrad")                         \
+                              .Device(DEVICE_GPU)                  \
+                              .TypeConstraint<type>("T")           \
+                              .TypeConstraint<int32>("Tmultiples") \
```
vrv: should we also add the int64 multiples kernel for TileGrad too?  I think we can do that in a separate PR, but it probably makes sense to do this for completeness.

vrv (approved):

tensorflow/core/kernels/tile_ops.cc
yongtang: @vrv Again thanks for the review and sorry for the `TileGrad`. Let me add a PR shortly.

PR #13894, opened 2017-10-22
Cauchy Distribution

cshenton:
Added the cauchy distribution to `contrib.distributions`.

The cauchy distribution is an example of a distribution with undefined moments. I couldn't find an existing implementation to reference against, so I'm not sure if my approach to this:
```python
if self.allow_nan_stats:
  return constant_op.constant(float("nan"), shape=self.batch_shape)
else:
  raise ValueError("`mean` is undefined for Cauchy distribution.")
```
is correct, please advise.

### DIFF
--- a/tensorflow/contrib/distributions/python/kernel_tests/cauchy_test.py
+++ b/tensorflow/contrib/distributions/python/kernel_tests/cauchy_test.py
@@ -1,4 +1,4 @@
-# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -86,14 +86,6 @@ class CauchyTest(test.TestCase):
     self._testParamStaticShapes(
         tensor_shape.TensorShape(sample_shape), sample_shape)
 
-  def testCauchyWithSoftplusScale(self):
-    with self.test_session():
-      loc = array_ops.zeros((10, 3))
-      rho = array_ops.ones((10, 3)) * -2.
-      cauchy = cauchy_lib.CauchyWithSoftplusScale(loc=loc, scale=rho)
-      self.assertAllEqual(loc.eval(), cauchy.loc.eval())
-      self.assertAllEqual(nn_ops.softplus(rho).eval(), cauchy.scale.eval())
-
   def testCauchyLogPDF(self):
     with self.test_session():
       batch_size = 6
@@ -104,16 +96,16 @@ class CauchyTest(test.TestCase):
 
       log_pdf = cauchy.log_prob(x)
       self.assertAllEqual(cauchy.batch_shape_tensor().eval(),
-                          log_pdf.get_shape())
+                          log_pdf.shape)
       self.assertAllEqual(cauchy.batch_shape_tensor().eval(),
                           log_pdf.eval().shape)
-      self.assertAllEqual(cauchy.batch_shape, log_pdf.get_shape())
+      self.assertAllEqual(cauchy.batch_shape, log_pdf.shape)
       self.assertAllEqual(cauchy.batch_shape, log_pdf.eval().shape)
 
       pdf = cauchy.prob(x)
-      self.assertAllEqual(cauchy.batch_shape_tensor().eval(), pdf.get_shape())
+      self.assertAllEqual(cauchy.batch_shape_tensor().eval(), pdf.shape)
       self.assertAllEqual(cauchy.batch_shape_tensor().eval(), pdf.eval().shape)
-      self.assertAllEqual(cauchy.batch_shape, pdf.get_shape())
+      self.assertAllEqual(cauchy.batch_shape, pdf.shape)
       self.assertAllEqual(cauchy.batch_shape, pdf.eval().shape)
 
       if not stats:
@@ -133,20 +125,20 @@ class CauchyTest(test.TestCase):
 
       log_pdf = cauchy.log_prob(x)
       log_pdf_values = log_pdf.eval()
-      self.assertEqual(log_pdf.get_shape(), (6, 2))
+      self.assertEqual(log_pdf.shape, (6, 2))
       self.assertAllEqual(cauchy.batch_shape_tensor().eval(),
-                          log_pdf.get_shape())
+                          log_pdf.shape)
       self.assertAllEqual(cauchy.batch_shape_tensor().eval(),
                           log_pdf.eval().shape)
-      self.assertAllEqual(cauchy.batch_shape, log_pdf.get_shape())
+      self.assertAllEqual(cauchy.batch_shape, log_pdf.shape)
       self.assertAllEqual(cauchy.batch_shape, log_pdf.eval().shape)
 
       pdf = cauchy.prob(x)
       pdf_values = pdf.eval()
-      self.assertEqual(pdf.get_shape(), (6, 2))
-      self.assertAllEqual(cauchy.batch_shape_tensor().eval(), pdf.get_shape())
+      self.assertEqual(pdf.shape, (6, 2))
+      self.assertAllEqual(cauchy.batch_shape_tensor().eval(), pdf.shape)
       self.assertAllEqual(cauchy.batch_shape_tensor().eval(), pdf_values.shape)
-      self.assertAllEqual(cauchy.batch_shape, pdf.get_shape())
+      self.assertAllEqual(cauchy.batch_shape, pdf.shape)
       self.assertAllEqual(cauchy.batch_shape, pdf_values.shape)
 
       if not stats:
@@ -164,9 +156,9 @@ class CauchyTest(test.TestCase):
 
       cauchy = cauchy_lib.Cauchy(loc=loc, scale=scale)
       cdf = cauchy.cdf(x)
-      self.assertAllEqual(cauchy.batch_shape_tensor().eval(), cdf.get_shape())
+      self.assertAllEqual(cauchy.batch_shape_tensor().eval(), cdf.shape)
       self.assertAllEqual(cauchy.batch_shape_tensor().eval(), cdf.eval().shape)
-      self.assertAllEqual(cauchy.batch_shape, cdf.get_shape())
+      self.assertAllEqual(cauchy.batch_shape, cdf.shape)
       self.assertAllEqual(cauchy.batch_shape, cdf.eval().shape)
       if not stats:
         return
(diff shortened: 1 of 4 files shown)


cshenton:
I signed it.

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+  def _log_prob(self, x):
+    return self._log_unnormalized_prob(x) - self._log_normalization()
+
+  def _prob(self, x):
+    return math_ops.exp(self._log_prob(x))
+
+  def _cdf(self, x):
+    return 1/math.pi * math_ops.atan(self._z(x)) + 0.5
+
+  def _log_cdf(self, x):
```
jvdillon: For this we should do:

math_ops.log1p(2 / np.pi * math_ops.atan(self._z(x))) - np.log(2)

That should be more accurate than log(self._cdf)

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+    return math_ops.log(self._cdf(x))
+
+  def _log_unnormalized_prob(self, x):
+    return -math_ops.log(1 + math_ops.square(self._z(x)))
+
+  def _log_normalization(self):
+    scale = self.scale * array_ops.ones_like(self.loc)
```
jvdillon: No need to force the broadcast, just:
return np.log(np.pi) + math_ops.log(self.scale)

Why? 
This broadcasting will happen automatically when _log_unnorm_prob is added to _log_normalization

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+  def _event_shape_tensor(self):
+    return constant_op.constant([], dtype=dtypes.int32)
+
+  def _event_shape(self):
+    return tensor_shape.scalar()
+
+  def _sample_n(self, n, seed=None):
+    shape = array_ops.concat([[n], self.batch_shape_tensor()], 0)
+    probs = random_ops.random_uniform(
+        shape=shape, minval=0., maxval=1., dtype=self.loc.dtype, seed=seed)
```
jvdillon: self.dtype instead of self.loc.dtype

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+    return self._scale
+
+  def _batch_shape_tensor(self):
+    return array_ops.broadcast_dynamic_shape(
+        array_ops.shape(self.loc),
+        array_ops.shape(self.scale))
+
+  def _batch_shape(self):
+    return array_ops.broadcast_static_shape(
+        self.loc.get_shape(),
```
jvdillon: .get_shape() is deprecated; use the new property .shape instead

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+    with ops.name_scope("standardize", values=[x]):
+      return (x - self.loc) / self.scale
+
+  def _inv_z(self, z):
+    """Reconstruct input `x` from a its normalized version."""
+    with ops.name_scope("reconstruct", values=[z]):
+      return z * self.scale + self.loc
+
+  def _log_survival_function(self, x):
+    return math_ops.log(self._survival_function(x))
```
jvdillon: Not needed; base class does this

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+    return self._quantile(probs)
+
+  def _log_prob(self, x):
+    return self._log_unnormalized_prob(x) - self._log_normalization()
+
+  def _prob(self, x):
+    return math_ops.exp(self._log_prob(x))
+
+  def _cdf(self, x):
+    return 1/math.pi * math_ops.atan(self._z(x)) + 0.5
```
jvdillon: Id probably just write:
math_ops.atan(self._z(x)) / np.pi + 0.5
but this is fine too.

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+
+  def _log_unnormalized_prob(self, x):
+    return -math_ops.log(1 + math_ops.square(self._z(x)))
+
+  def _log_normalization(self):
+    scale = self.scale * array_ops.ones_like(self.loc)
+    return math.log(math.pi) + math_ops.log(scale)
+
+  def _entropy(self):
+    scale = self.scale * array_ops.ones_like(self.loc)
```
jvdillon: More efficient to broadcast *after* calculations:

h = np.log(4 * np.pi) + math_ops.log(scale)
return h * array_ops.ones_like(self.loc)

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+
+  def _inv_z(self, z):
+    """Reconstruct input `x` from a its normalized version."""
+    with ops.name_scope("reconstruct", values=[z]):
+      return z * self.scale + self.loc
+
+  def _log_survival_function(self, x):
+    return math_ops.log(self._survival_function(x))
+
+  def _survival_function(self, x):
```
jvdillon: Not needed; base class does this

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+    return 1. - self._cdf(x)
+
+  def _mean(self):
+    if self.allow_nan_stats:
+      return constant_op.constant(float("nan"), shape=self.batch_shape)
```
jvdillon: This should be `self.batch_shape_tensor()`.  Can you ensure your test checks the mean when allow_nan_stats is True?

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,252 @@
...
+    else:
+      raise ValueError("`mean` is undefined for Cauchy distribution.")
+
+  def _stddev(self):
+    if self.allow_nan_stats:
+      return constant_op.constant(float("nan"), shape=self.batch_shape)
```
jvdillon: This should be `self.batch_shape_tensor()`.  Can you ensure your test checks the mean when allow_nan_stats is True?

jvdillon (changes requested):
Thank you! This is a much needed contribution. Im happy to see you grokked our idioms!

My only substantive comments relate to numerical stability, ie, use log1p where we can.

tensorflow/contrib/distributions/python/ops/cauchy.py
cshenton: Passing the result of `self.batch_shape_tensor()` here causes complaints about array indexing, so I'll do `float("nan") * tf.ones(self.batch_shape_tensor())`. That test is in there, also.

cshenton:
Thanks for the review. I've made those changes you requested. As an aside, I'm having some issues running unit tests locally. Docker container builds fine, collects the tests, then skips them all. Is it alright if I rely on the CI servers to do the heavy lifting?

ebrevdo:
Yes; that's fine.  But I think you may have to ask us to trigger it.

tensorflow/contrib/distributions/python/kernel_tests/cauchy_test.py
```diff
@@ -0,0 +1,440 @@
...
+    is_finite = np.isfinite(tensor.eval())
+    all_true = np.ones_like(is_finite, dtype=np.bool)
+    self.assertAllEqual(all_true, is_finite)
+
+  def _testParamShapes(self, sample_shape, expected):
+    with self.test_session():
+      param_shapes = cauchy_lib.Cauchy.param_shapes(sample_shape)
+      loc_shape, scale_shape = param_shapes["loc"], param_shapes["scale"]
+      self.assertAllEqual(expected, loc_shape.eval())
+      self.assertAllEqual(expected, scale_shape.eval())
```
ebrevdo: the two evals cause TF to run the graph twice.  can you do:

with self.test_session(use_gpu=True) as sess:
 ...
  self.assertAllEqual((expected, expected), sess.run((loc_shape, scale_shape))

cshenton:
Okay, I'll investigate getting the local tests working a bit further then.

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,221 @@
...
+      return (x - self.loc) / self.scale
+
+  def _inv_z(self, z):
+    """Reconstruct input `x` from a its normalized version."""
+    with ops.name_scope("reconstruct", values=[z]):
+      return z * self.scale + self.loc
+
+  def _mean(self):
+    if self.allow_nan_stats:
+      return float("nan") * array_ops.ones(self.batch_shape_tensor())
```
ebrevdo: this should match the dtype the cauchy distribution was created with.  you can use `dtype.as_numpy_dtype(np.nan)` for that.  also prefer using array_ops.fill over multiplying a scalar by ones.

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,221 @@
...
+    else:
+      raise ValueError("`mean` is undefined for Cauchy distribution.")
+
+  def _stddev(self):
+    if self.allow_nan_stats:
+      return float("nan") * array_ops.ones(self.batch_shape_tensor())
```
ebrevdo: ditto.

tensorflow/contrib/distributions/python/ops/cauchy.py
cshenton: Good to know, that's what I was trying to achieve with `constant_ops.constant` previously, but ran into issues. Didn't know about `fill`, thanks for the heads up!

cshenton:
Okay, now using `np.nan` with correct dtype, let me know if there's anything else.

tensorflow/contrib/distributions/python/ops/cauchy.py
```diff
@@ -0,0 +1,223 @@
...
+
+class Cauchy(distribution.Distribution):
+  """The Cauchy distribution with location `loc` and scale `scale`.
+
+  #### Mathematical details
+
+  The probability density function (pdf) is,
+
+  ```none
+  pdf(x; loc, scale) = 1 / (pi*scale*(1+((x-loc)/scale)**2))
```
jvdillon: Please put spaces around all binary operators (except exponentiation).  This makes it easier to read,

jvdillon (approved):
Looks good to me. Thanks!

tensorflow/contrib/distributions/python/ops/cauchy.py
vrv: @jvdillon for small edits like this you can make an edit directly to the PR usually.  This speeds up the process of accepting changes due to nitpicks.

jvdillon (approved):

cshenton:
Okay, fixed some fluff ups in `BUILD` which had the test and dependency paths wrong. Can confirm that the tests on this commit are all passing locally. Could someone please trigger the TF Test Suite again?

ebrevdo:
Jenkins, test this please.

jvdillon:
@ebrevdo maybe kickoff another test run?

PR #13899, opened 2017-10-22
nsync lib name change

formath:
Now, the lib name is `nsync.a`. It will cause problem `lnsync` not found in cmake. So, change it to `libnsync.a`


vrv:

vrv:
This doesn't seem to work, is there some kind of update that needs to be done first?

formath:
Yes. This pull should be merged firstly.
https://github.com/google/nsync/pull/2

vrv:
Can you do this instead? https://stackoverflow.com/questions/31038963/how-do-you-rename-a-library-filename-in-cmake

formath:
Fine. I will use absolute path for `nsync.a`.

PR #13904, opened 2017-10-22
Make special_math._ndtri work with partially-specified shapes

artsobolev:
Current implementation of the `special_math._ndtri` function does not work with tensors of partially-known shape, which prevents one from computing Normal distribution's quantile function for non-fixed size batches.


artsobolev:
I signed it!

jvdillon (approved):

vrv:

artsobolev:
@vrv, what happened to the tests? Looks like some kind of an internal CI error.

vrv:
Let's try it again, our CI infra is having lots of problems lately as we transition to a new system.

PR #13907, opened 2017-10-22
Add int64 padding support for MirrorPad

yongtang:
This fix adds int64 padding support for `MirrorPad`. In the `array_ops.cc` the `MirrorPad`/`MirrorPadGrad` has been specified as supporting int64 padding. The related kernels does not have the int64 padding registered though.

This fix adds the int64 padding support. This fix also adds additional test cases for coverage.

### DIFF
--- a/tensorflow/core/kernels/mirror_pad_op.cc
+++ b/tensorflow/core/kernels/mirror_pad_op.cc
@@ -18,10 +18,10 @@ limitations under the License.
 #define EIGEN_USE_THREADS
 
 #include "tensorflow/core/kernels/mirror_pad_op.h"
-
 #include <string>
 
 #include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
+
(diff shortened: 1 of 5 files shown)


vrv (approved):

vrv:

yongtang:
@vrv Thanks for the review and update in the PR. And sorry for missing the Eigen header issue.

On a side note, I did notice that if we put an additional empty line in between Eigen header and the rest of the header, then a `clang-format -i --style=google file.cc` will actually respect the order that Eigen header is at the top.

Maybe it is worth the effort to scan through and do that, so that any developer could safely run `clang-format` (or automatically run by jenkins) without worrying about the order mess-up?

vrv:
Yes, that sounds like a good idea!  But I also might be wrong about the eigen headers thing.  It was required at some point because Eigen on CPU required a define to use multiple threads, but someone might want to double check before doing this so we don't end up causing a lot of wasted work.

PR #13908, opened 2017-10-22
Add int64 shape support on GPU for stateless random ops.

yongtang:
This fix adds int64 shape support on GPU for stateless random ops `StatelessRandomUniform`, `StatelessRandomNormal`, `StatelessTruncatedNormal`.

The int64 shape for stateless random ops is already supported on CPU with int32/int64 processed properly through `MakeShape`.

However, on GPU a type constraint `.TypeConstraint<int32>("T")` has been improperly added. Such a type constraint actually prevents an int64 shape type to run on GPU. (As a comparision, no type constraint on CPU).

This fix removes the type constraint and allows int64 shape to be run on GPU. This fix also adds test cases for int64 shape support on stateless random ops.

### DIFF
--- a/tensorflow/contrib/stateless/python/kernel_tests/stateless_random_ops_test.py
+++ b/tensorflow/contrib/stateless/python/kernel_tests/stateless_random_ops_test.py
@@ -82,7 +82,7 @@ class StatelessOpsTest(test.TestCase):
 
   def testShapeType(self):
     with self.test_session(use_gpu=True):
-      for shape_dtype in [dtypes.int64]:
+      for shape_dtype in [dtypes.int32, dtypes.int64]:
         seed_t = array_ops.placeholder(dtypes.int64, shape=[2])
         seeds = [(x, y) for x in range(5) for y in range(5)] * 3
         for stateless_op, _ in CASES:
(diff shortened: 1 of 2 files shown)


tensorflow/contrib/stateless/python/kernel_tests/stateless_random_ops_test.py
```diff
@@ -79,6 +80,21 @@ def testDeterminism(self):
             for s1, v1 in values:
               self.assertEqual(s0 == s1, np.all(v0 == v1))
 
+  def testShapeType(self):
+    with self.test_session(use_gpu=True):
+      for shape_dtype in [dtypes.int64]:
```
vrv: Does it make sense to test both int64 and int32?

vrv:

vrv (approved):

tensorflow/contrib/stateless/python/kernel_tests/stateless_random_ops_test.py
yongtang: Sorry I missed that one. Thanks for the update in the PR and help!

PR #13909, opened 2017-10-22
Add int64 Tperm type support for `Transpose`

yongtang:
This fix adds int64 Tperm support for `Transpose`. In `array_ops.cc`, `Transpose` and `ConjugateTranspose` have been specified as accepting int32 and int64 perm types. However, only int32 kernels has been registered.

This fix adds the int64 perm support by removing the constraint on Tperm, resolve the type at runtime, and copying the data type accordingly to correctly handle the int64/int32 types.

Additional tests have been added as well.


vrv:

vrv (approved):

PR #13920, opened 2017-10-23
Update build_all_ios.sh to support per arch builds

powderluv:


powderluv:
Will update with a new commit defaulting the libs built to what we have currently have.

powderluv:
closing this PR will submit a new one with the ability to build just one arch.

PR #13923, opened 2017-10-23
Branch 173127955

benoitsteiner:


gunan:
@annarev is working on fixing the android failure.
@vrv FYI.

benoitsteiner:
Should we merge this as is then ? We can always merge the android failure later.

gunan:
Sounds good. Let's avoid increasing the skew.
Please feel free to merge right away. We can do a smaller push when the fix is in.

gunan (approved):

PR #13935, opened 2017-10-24
Fix mnist_softmax tutorial: W should be randomly initialized, should …

eeandrew:
Fix mnist_softmax tutorial: W should be randomly initialized, should not be initialized as zeros. Otherwise, you will get a symmetric NN.
```
# This is wrong and misleading. W should not initialized as zeros. 
# If you do so, you will get a symmetric network. 
W = tf.Variable(tf.zeros([784, 10]))
```
Instead, W should be randomly initialized
```
W = tf.Variable(tf.random_normal([784, 10]))
```
Actually, there is another mistake:
```
# This is not how we use mini-batch in practical environment
for _ in range(1000):
    batch_xs, batch_ys = mnist.train.next_batch(100)
    sess.run(train_step, feed_dict={x: batch_xs, y_: batch_ys})
```
Instead, mini-batch should be used as follows:
```
for _ in range(1000):
   for batch in range(550):
      batch_xs, batch_ys = mnist.train.next_batch(100)
      sess.run(train_step, feed_dict={x: batch_xs, y_: batch_ys})
```
But maybe for simplicity, we can ignore this mistake

### DIFF
--- a/tensorflow/examples/tutorials/mnist/mnist_softmax.py
+++ b/tensorflow/examples/tutorials/mnist/mnist_softmax.py
@@ -38,7 +38,7 @@ def main(_):
 
   # Create the model
   x = tf.placeholder(tf.float32, [None, 784])
-  W = tf.Variable(tf.zeros([784, 10]))
+  W = tf.Variable(tf.random_normal([784, 10]))
   b = tf.Variable(tf.zeros([10]))
   y = tf.matmul(x, W) + b
 


eeandrew:
I signed it!

JesseLivezey:
@eeandrew this no-hidden-layer network, multinomial or softmax logistic regression, is convex in the parameters, so initialization to zeros is fine. Symmetry-breaking is a problem in networks with hidden layers.

dr4b:
In terms of not breaking the docs this looks fine to me?  (Though I don't think I'm technically supposed to approve code sample changes.)

MarkDaoust:
Hi thanks for the PR.

This PR would only make sense on `master`, otherwise it would only be fixed in 1.4. 

But I'm not sure this one is worth pursuing:

* @JesseLivezey if has a good (if subtle), point. 
* We'll be moving and modernizing these examples shortly.

PR #13938, opened 2017-10-24
make "smart_cond" api public and reusable

### LINKED ISSUE #13903: Feature request: make `smart_cond` public API
------------------------

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS High Sierra
- **TensorFlow version (use command below)**: v1.3.0-24-g658866597
- **Python version**: 3.6.3
- **CUDA/cuDNN version**: None
- **GPU model and memory**: None
- **Exact command to reproduce**: None

### Describe the problem

Currently `tf.cond` does not work if predicate is a Python boolean. As a result, people frequently have to write conditional statements twice, one with `if` statement, and one with `tf.cond` call. There is a `smart_cond` in `tensorflow/python/layers/utils.py`, but it is not in the public API or searchable in documentation. Petition to make it public or just integrate the smartness in `tf.cond` altogether. It is not a big change and will not impact backwards compatibility.

michaelzqj:
Fix #13903 .

Move the implementation of "smart_cond" and "constant_value" from `tensorflow/python/layers/utils` to `tensorflow/python/ops/control_flow_ops` and add corresponding test and expose `smart_cond` to `tensorflow/python/ops/standard_ops`.

@martinwicke should I implement the API in this way or not? As I think ops should be lower level API than layers and `smart_cond` should be a kind of ops. It's my first time to contribute to TensorFlow. Please feel free to correct me if there's any problem with the design and codes.

### DIFF
--- a/tensorflow/python/layers/utils.py
+++ b/tensorflow/python/layers/utils.py
@@ -195,19 +195,7 @@ def smart_cond(pred, fn1, fn2, name=None):
   Raises:
     TypeError: If `fn1` or `fn2` is not callable.
   """
-  if not callable(fn1):
-    raise TypeError('`fn1` must be callable.')
-  if not callable(fn2):
-    raise TypeError('`fn2` must be callable.')
-
-  pred_value = constant_value(pred)
-  if pred_value is not None:
-    if pred_value:
-      return fn1()
-    else:
-      return fn2()
-  else:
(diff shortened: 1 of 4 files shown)


GeorgyZhou:
Oops, seems that my first PR could not be that smooth. It seems that the sanity checks failed at GPG signature verification.

Issue #13885, opened 2017-10-21
tf.reduce_mean is not compatible with np.mean


amirj:
[tf.reduce_mean](https://www.tensorflow.org/api_docs/python/tf/reduce_mean) emphasized that this function is compatible with numpy:

> Equivalent to np.mean

But it doesn't in the output type. Consider the following code for example:

```
import tensorflow as tf
x = tf.Variable([1, 0, 1, 0])
init = tf.global_variables_initializer()
sess = tf.Session()
sess.run(init)
print(sess.run(tf.reduce_mean(x)))

```

The output is zero. It seems that tf.reduce_mean infer the output type from the input tensor because casting the input tensor to float values, solve the problem. This attribute is not compatible to np.mean:

```
import numpy as np
print(np.mean([1,0,0,1]))
```

### System information
- **OS Platform and Distribution**: Linux Ubuntu 16.04
- **TensorFlow installed from (source or binary)**: Source
- **TensorFlow version (use command below)**: 1.3
- **Python version**: 3.6

yongtang:
In numpy, `numpy.mean` has a `dtype` that could be used to specify the output type:
https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.mean.html

 By default this is `dtype=float64` for integer types and same type as input for non-integer types.

It might be possible to add a dtype in tf.reduce_mean though a type cast is always needed I assume.

drpngx:
@MarkDaoust should we update the numpy compat docstring?

/CC @aselle

amirj:
@drpngx @MarkDaoust  _dtype_ is an optional parameter in both TF and Numpy. I think it should be clear in the document or adapt to Numpy. It's my honer to contribute in both cases.

MarkDaoust:
Sure, if you want to update the `@compatability(numpy)` note we'd welcome the PR.

But I don't think this is enough of a bug that we can break backwards compatibility to fix it.

DjangoPeng:
@drpngx I'd like to send a PR to add the `dtype` argument for `reduce_mean` funciton.

drpngx:
@DjangoPeng sounds good, it'll have to go to api review. Thank you!

MarkDaoust:
Fixed by #13941

PR #13941, opened 2017-10-24
add a note on numpy compatibility

amirj:
`np.mean` has a `dtype` parameter that could be used to specify the output type. By default this is `dtype=float64`. On the other hand, `tf.reduce_mean` has an aggressive type inference from `input_tensor`. This should be clear in the Numpy compatibility section.
https://github.com/tensorflow/tensorflow/issues/13885


martinwicke (approved):

martinwicke:
Jenkins, test this please.

PR #13974, opened 2017-10-25
Update device string comments in node_def.proto

tayo:
Clarifying comments for valid device string in NodeDef, as discussed in PR #13874.

Notes:
1. The device string is as returned by device.py to_string() 
2. parse_from_string() in device.py seems currently allows an empty string for device_type (line 167). This commit reflects that.  Is this actually intended behavior?
3. I notice our regex convention does not use '+' (e.g. XX* instead of X+).


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

Issue #13094, opened 2017-09-17
Feature Request - SRU (Simple Recurrent Unit) Cell


desire2020:
ArXiv path: https://arxiv.org/abs/1709.02755

This is a recently proposed, parallelization-friendly RNN cell. The author released his own PyTorch version of the SRU. We are looking forward to an offical tensorflow implementation with Cudnn optimizations.

tjingrant:
Hi, I'm interested in taking a crack at this. Will keep you updated.

buptpriswang:
Hi, it's interesting RNN cell, will keep you updated

argman:
Any progress?

tjingrant:
Will submit the initial PR early next week. 
Sorry I started late because I have to clear some build errors on master.

GeorgyZhou:
I am interested in this question as well. If @tjingrant have some problems, I can also join in.

tjingrant:
I apologize for keeping everyone interested waiting and will do my best to release a PR early this week. @GeorgyZhou: thanks for your offer, I will let you know if I need extra hands!

hadaev8:
Any news?

stegben:
I'm working on a fused version of SRU (process on whole time steps)

danijar:
Hi @stegben, what's the status of this? It sounds like the fused implementation is really what would make SRU most useful.

tjingrant:
To be honest, after implementing the non-fused version, I suspect whether it's worth going deeper. The [paper](https://openreview.net/forum?id=rJBiunlAW) has been rejected on the ground that it is a special case of Quasi-RNN which makes sense. It's probably better for everyone to work on a better version of Quasi-RNN than keeping implementing SRU.

raofengyun:
@tjingrant it's now published on EMNLP: https://aclanthology.info/papers/D18-1477/d18-1477  
we tried it on some dataset, and it really works good and fast, but we have to use pytorch, since the tf version SRUCell is NOT faster than GRUCell.  

could anyone implement a tf-version SRU instead of SRUCell?

dynamicwebpaige:
Closing, as this feature is outside the scope of TensorFlow Core and no longer relevant, from a research perspective. If I'm mistaken, please feel free to reopen and we can route to [TensorFlow Addons](https://www.github.com/tensorflow/addons). Thanks! 🙂

PR #13978, opened 2017-10-25
Initial SRU Implementation

tjingrant:
Hi, as per https://github.com/tensorflow/tensorflow/issues/13094, I'm submitting an initial PR for SRU implementation. I have checked the results with implementation [here](https://github.com/desire2020/SRU-tensorflow).

I intend to implement SRUBlock as well; but I have been distracted away by other high priority things until recently. So I will incrementally add the SRUBlock here as well.

In addition, it occurs to me that SRU only allows num_units equal to input_size. Can someone verify my interpretation? Otherwise we cannot perform equaiton 7 in the [paper](https://arxiv.org/pdf/1709.02755.pdf) because of last-dimension size mismatch.


ebrevdo (changes requested):
Could you look at the format of the simple lstm cell in core at master?  We recently converted it to use a build function, and it subclasses a special subclass of RNNCell.  It's constructor also takes a name= argument.  Could you modify the sru class to do the same?

tjingrant:
Just a reminder, quote from my original PR comment:

> In addition, it occurs to me that SRU only allows num_units equal to input_size. Can someone verify my interpretation? Otherwise we cannot perform equaiton 7 in the paper because of last-dimension size mismatch.

tjingrant:
@ebrevdo any update?

tensorflow/contrib/rnn/python/kernel_tests/core_rnn_cell_test.py
```diff
@@ -139,6 +139,21 @@ def testGRUCell(self):
         # Smoke test
         self.assertAllClose(res[0], [[0.156736, 0.156736]])
 
+  def testSRUCell(self):
+    with self.test_session() as sess:
+      with variable_scope.variable_scope(
+          "root", initializer=init_ops.constant_initializer(0.5)):
+        x = array_ops.zeros([1, 2])
+        m = array_ops.zeros([1, 2])
+        g, _ = rnn_cell_impl.SRUCell(2, \
```
ebrevdo: nit: don't use backslash here.

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -324,6 +323,78 @@ def __call__(self, inputs, state, scope=None):
...
+
+    self._bias = self.add_variable(
+        _BIAS_VARIABLE_NAME,
+        shape=[2 * self._num_units],
+        initializer=bias_initializer)
+
+    self._built = True
+
+  def call(self, inputs, state):
+    """Simple recurrent unit (SRU) with nunits cells."""
```
ebrevdo: s/nunits/num_units/.

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -324,6 +323,78 @@ def __call__(self, inputs, state, scope=None):
...
+
+    U = math_ops.matmul(inputs, self._kernel)
+    x_bar, f_intermediate, r_intermediate = \
```
ebrevdo: nit:

```python
x_bar, f_intermediate, r_intermediate = (
  array_ops.split(...))
```

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -324,6 +323,78 @@ def __call__(self, inputs, state, scope=None):
...
+      array_ops.split(value=U, num_or_size_splits=3, axis=1)
+
+    f_r = math_ops.sigmoid(nn_ops.bias_add(array_ops.concat([f_intermediate, \
```
ebrevdo: ditto regarding backslashes

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -324,6 +323,78 @@ def __call__(self, inputs, state, scope=None):
...
+      raise ValueError("Expected inputs.shape[-1] to be known, saw shape: %s"
+                       % inputs_shape)
+
+    input_depth = inputs_shape[1].value
+    self._kernel = self.add_variable(
+        _WEIGHTS_VARIABLE_NAME,
+        shape=[input_depth, 3 * self._num_units],
+        initializer=self._kernel_initializer)
+
+    bias_initializer = self._bias_initializer or \
```
ebrevdo:
```python
... = (
  self._bais_initializer or ...)
```

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -324,6 +323,78 @@ def __call__(self, inputs, state, scope=None):
...
+  """
+
+  def __init__(self,
+               num_units,
+               activation=None,
+               reuse=None,
+               kernel_initializer=None,
+               bias_initializer=None):
+    super(SRUCell, self).__init__(_reuse=reuse)
+    self._num_units = num_units
```
ebrevdo: set the input_spec to ensure inputs are matrices only.

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -324,6 +323,78 @@ def __call__(self, inputs, state, scope=None):
...
+    projection matrices.
+    bias_initializer: (optional) The initializer to use for the bias.
+  """
+
+  def __init__(self,
+               num_units,
+               activation=None,
+               reuse=None,
+               kernel_initializer=None,
+               bias_initializer=None):
```
ebrevdo: add a name=None argument, see the LSTMCell in core for an example.

ebrevdo (changes requested):

tjingrant:
@ebrevdo I've updated my code according to your comments.

tjingrant:
@ebrevdo any updates?

tjingrant:
@ebrevdo any update?

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -323,6 +322,71 @@ def __call__(self, inputs, state, scope=None):
...
+  def output_size(self):
+    return self._num_units
+
+  def build(self, inputs_shape):
+    if inputs_shape[1].value is None:
+      raise ValueError("Expected inputs.shape[-1] to be known, saw shape: %s"
+                       % inputs_shape)
+
+    input_depth = inputs_shape[1].value
+    assert input_depth == self._num_units, "SRU requires unit_num == input_size"
```
ebrevdo: raise a ValueError instead of asserting.

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -323,6 +322,71 @@ def __call__(self, inputs, state, scope=None):
...
+
+  Args:
+    num_units: int, The number of units in the SRU cell.
+    activation: Nonlinearity to use.  Default: `tanh`.
+    reuse: (optional) Python boolean describing whether to reuse variables
+     in an existing scope.  If not `True`, and the existing scope already has
+     the given variables, an error is raised.
+    kernel_initializer: (optional) The initializer to use for the weight and
+    projection matrices.
+    bias_initializer: (optional) The initializer to use for the bias.
```
ebrevdo: name: ...

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -323,6 +322,71 @@ def __call__(self, inputs, state, scope=None):
...
 
+class SRUCell(_LayerRNNCell):
+  """Training RNNs as Fast as CNNs (cf. https://arxiv.org/abs/1709.02755).
+
+  Args:
+    num_units: int, The number of units in the SRU cell.
+    activation: Nonlinearity to use.  Default: `tanh`.
+    reuse: (optional) Python boolean describing whether to reuse variables
+     in an existing scope.  If not `True`, and the existing scope already has
+     the given variables, an error is raised.
```
ebrevdo: reorder args to match __init__'s signature.

ebrevdo (approved):
approval conditioned on fixes to nits.

ebrevdo:
@tjingrant quick ping on this.

tjingrant:
@ebrevdo should be good now.

benoitsteiner:
Jenkins, test this please

ebrevdo:
Thanks!

gunan:
This PR introduces an API change. We will need to revert this and resubmit after an API review.

asimshankar:
@tjingrant : Could you add some more detail to the docstring, the current one liner seems a bit ... succinct :)

tjingrant:
@asimshankar yes, of course. But let's talk about it over here (https://github.com/tensorflow/tensorflow/pull/15436).

tensorflow/python/ops/rnn_cell_impl.py
```diff
@@ -289,6 +288,86 @@ def __call__(self, inputs, state, scope=None):
     # method.  See the class docstring for more details.
     return base_layer.Layer.__call__(self, inputs, state, scope=scope)
 
+class SRUCell(_LayerRNNCell):
```
ebrevdo: ick!  this *is* an API change.  this should have gone into contrib!  i thought it had been moved since the test had been moved to contrib as well!

ebrevdo:
let's revert this and resubmit with the SRUCell going into tf.contrib.rnn (not just the test!)

XiaoyuGuo123:
about this function 'tf.contrib.rnn.SRUCell(units)'   
How many states does the return value of this function call directly? I found it in debugging and it seems that there is only one state.
What should I do if I want to stack several layers of SRU?Because I don't understand that this function has only one return value. How does the internal state of c work when stacked on the SRU layers?

Issue #13989, opened 2017-10-26
Fused batch norm can be folded with atrous conv2d 


liyinhgqw:
### Describe the problem
As we can fold batch norm with convolution, we should also fold batch norm with atrous convolution, which has not been implemented.

smohan10:
Hi, 

I am working on folding batch norm with atrous convolution for quantization experiments. It seems the Space to Batch -> Conv -> Batch to Space -> Batch Norm -> Activation isn't getting folded correctly. There seems to be a shape incompatibility. Has anyone made progress on this issue?

tilakrayal:
@liyinhgqw,
Respected PR was closed. Could you please check the PR and provide the required information. Thank you!

PR #13990, opened 2017-10-26
Support fold batch norm for atrous conv2d

liyinhgqw:
As we can fold batch norm with convolution, we should also fold batch norm with atrous convolution, which has not been implemented.

issue: #13989


liyinhgqw:
I signed it!)

liyinhgqw:
I signed it!

liyinhgqw:
I signed it!

liyinhgqw:
I signed it!

PR #13998, opened 2017-10-26
Add customerized kernel implementation for clip_by_value

### LINKED ISSUE #7225: Feature request: equivalent of np.clip

yongtang:
This fix tries to address the issue raised in #7225 where `tf.clip_by_value` does not have a custom kernel and reused `tf.maximum` and `tf.mimimum`. In case scalar values are passed to `tf.clip_by_value`, unnecessary memory usage might incur.

This fix adds the customerized kernel implementation for `tf.clip_by_value`.

This fix fixes #7225.

### DIFF
--- a/tensorflow/core/kernels/cwise_op_clip.cc
+++ b/tensorflow/core/kernels/cwise_op_clip.cc
@@ -1,4 +1,4 @@
-/* Copyright 2015 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/tensorflow/core/kernels/cwise_op_clip.cc
+++ b/tensorflow/core/kernels/cwise_op_clip.cc
@@ -25,9 +25,9 @@ typedef Eigen::GpuDevice GPUDevice;
 //   Device: E.g., CPUDevice, GPUDevice.
 //   Functor: defined above. E.g., functor::clip.
 template <typename Device, typename T>
-class TenaryOp : public OpKernel {
+class ClipOp : public OpKernel {
  public:
-  explicit TenaryOp(OpKernelConstruction* ctx) : OpKernel(ctx) {}
+  explicit ClipOp(OpKernelConstruction* ctx) : OpKernel(ctx) {}
 
   void Compute(OpKernelContext* ctx) override {
     const Tensor& in0 = ctx->input(0);
@@ -178,7 +178,7 @@ INSTANTIATE_CPU(uint16);
 #define REGISTER_CPU_KERNEL(type)                                       \
   REGISTER_KERNEL_BUILDER(                                              \
       Name("ClipByValue").Device(DEVICE_CPU).TypeConstraint<type>("T"), \
-      TenaryOp<CPUDevice, type>);
+      ClipOp<CPUDevice, type>);
 
 REGISTER_CPU_KERNEL(Eigen::half);
 REGISTER_CPU_KERNEL(float);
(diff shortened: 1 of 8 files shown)


tensorflow/core/kernels/cwise_op_clip.cc
```diff
@@ -0,0 +1,224 @@
+/* Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
benoitsteiner: s./2015/2017/

tensorflow/python/ops/clip_ops.py
```diff
@@ -56,18 +57,10 @@ def clip_by_value(t, clip_value_min, clip_value_max,
...
-    _ = t.shape.merge_with(t_min.shape)
-
-    t_max = math_ops.maximum(t_min, clip_value_min, name=name)
-    _ = t.shape.merge_with(t_max.shape)
-
-  return t_max
+    return gen_math_ops._clip_by_value(t,
+                                       clip_value_min,
+                                       clip_value_max,
+                                       name=name)
```
benoitsteiner: Before your change, TF was automatically generating a gradient from the gradient of the maximum and minimum ops used to clip the value, but it can't do this for your new ClipByValue: therefore you need to define and register a gradient op for ClipByValue.

benoitsteiner (changes requested):

yongtang:
Thanks @benoitsteiner. I was relatively new to grad code base and didn't take the grad registration into consideration before. Let me take a look and update the pull request accordingly.

tensorflow/core/kernels/cwise_op_clip.cc
yongtang: Thanks. Done.

tensorflow/python/ops/clip_ops.py
yongtang: Thanks. The gradient registration for ClipByValue has been added.

yongtang:
Thanks @benoitsteiner for the review. The PR has been updated with grad registration added. Test cases for gradients of clip_by_value have also been added. Please take a look.

martinwicke:
@benoitsteiner, can you take another look?

benoitsteiner:
Jenkins, test this please.

tensorflow/core/kernels/cwise_op_clip.cc
```diff
@@ -0,0 +1,224 @@
...
+
+typedef Eigen::ThreadPoolDevice CPUDevice;
+typedef Eigen::GpuDevice GPUDevice;
+
+// Basic coefficient-wise tenary operations.
+// This is the case for example of the clip_by_value.
+//   Device: E.g., CPUDevice, GPUDevice.
+//   Functor: defined above. E.g., functor::clip.
+template <typename Device, typename T>
+class TenaryOp : public OpKernel {
```
benoitsteiner: You have a typo, this should be TernaryOp. I'm also wondering why it's not just called ClipOp since the Compute method can only handle clipping.

benoitsteiner (approved):

benoitsteiner:
@yongtang  Can you run tensorflow/core/api_def/update_api_def.sh to update our API checks ? I should be able to merge this PR once that's done.

yongtang:
@benoitsteiner Thanks for the help! The PR has been updated.

gunan:
Looks like this is also failing the API checks. Also reverting this PR.

PR #14011, opened 2017-10-26
iOS/RPi Add the ability to choose ANDROID_TYPES_FULL

powderluv:
Some networks require "full" types instead of "slim" so remove
the hard coding of SLIM in iOS and RPi. It still defaults to
building SLIM for them if not ENV var is specified but now
you can build with

ANDROID_TYPES="-D__ANDROID_TYPES_FULL" \
./tensorflow/contrib/makefile/build_all_ios.sh

TEST: Verify the  -D__ANDROID_TYPES_SLIM__ flag is default and
you can override with an env var

### DIFF
--- a/tensorflow/contrib/makefile/Makefile
+++ b/tensorflow/contrib/makefile/Makefile
@@ -44,6 +44,11 @@ ifdef HEXAGON_LIBS
 	endif
 endif # HEXAGON_LIBS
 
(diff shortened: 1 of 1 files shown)


benoitsteiner:
Jenkins, test this please.

benoitsteiner (approved):

Issue #13887, opened 2017-10-21
ValueError raised when using AdamOptimizer, but not for GradientDescentOptimizer


rohan-varma:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Sierra 10.12.1
- **TensorFlow installed from (source or binary)**: binary
- **TensorFlow version (use command below)**: v1.2.0-rc2-21-g12f033d 1.2.0
- **Python version**: 3.5.0

### Source code / logs

Taken from this issue: https://github.com/tensorflow/tensorflow/issues/6220#issuecomment-314688444, the below code reproduces the error.

```
import tensorflow as tf
import numpy as np
class SimpleModel:
    def __init__(self):
        self.loss = self.calc_loss()
        self.train = self.train_model(self.loss)
    def calc_loss(self):
        W = tf.get_variable("w", [1])
        b = tf.Variable(tf.zeros([1]))
        y = W * x_data + b
        return tf.reduce_mean(tf.square(y - y_data))
    def train_model(self, loss):
        return tf.train.AdamOptimizer(0.5).minimize(loss)
        #return tf.train.GradientDescentOptimizer(0.5)
x_data = np.random.rand(100).astype(np.float32)
y_data = x_data * 0.1 + 0.3
s1 = SimpleModel()
tf.get_variable_scope().reuse_variables()
s2 = SimpleModel()
```

Running this gives the error `ValueError: Variable w/Adam_2/ does not exist, or was not created with tf.get_variable(). Did you mean to set reuse=None in VarScope?`, but when `train_model` is changed to return the `GradientDescentOptimizer`, the code compiles correctly.

I've tried following all the advice in the issue that relates to this (https://github.com/tensorflow/tensorflow/issues/6220), but haven't had any luck.

kazemSafari:
I am having the same issue. I use tensorflow 1.3 and my operating system is ubuntu 16.4.

drpngx:
Did you look at this?
https://stackoverflow.com/questions/44440900/cant-access-tensorflow-adam-optimizer-namespace

kazemSafari:
I looked at it. It definitely explains exactly what  the problem is but does not give any solution at all.
I do not know how to create that subscope that he is talking about without tensorflow raising an error.

kazemSafari:
Hi again!
[...]

PR #14015, opened 2017-10-27
Set reuse=False instead of reuse=None, and add suggestion for tf.AUTO_REUSE

rohan-varma:
As per https://github.com/tensorflow/tensorflow/issues/13887 we'd like to change the message that contains "reuse=None" to "reuse=False" and suggest the use of tf.AUTO_REUSE.


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please.

Issue #14014, opened 2017-10-27
Proposal: Compiling TF 1.4.0 GPU w/ CMAKE on Linux x64


myungjoo:
Currently, I have been playing with the cmake scripts of TensorFlow-GPU to make it CMAKE-buildable in a bare-metal x64 Linux. (isolated environment, without the internet, w/ OBS)
(I'm trying CMAKE because I don't want to port Java)

It's almost done (need code clean) and I'll probably send Pull-Request next week. However, there are a few things I want to check before I write additional features besides simply make it able to build with CMAKE for TF-GPU in Linux.

1. Tensorflow is statically linking to pre-installed devel packages. I think in some cases, it might be better to do dynamic linking (use .so) to other libraries (to save some memory and storage size). May I simply add an option like "tensorflow_USE_SHARED_LIBS_CUDA"? (e.g., cuda, nccl, ...)

2. Tensorflow is downloading a lot of external packages and use them "statically", which is pretty awful for some people. I'd like to make it use shared libraries as well (maybe along with some version restrictions). May I simply add an option like "tensorflow_USE_SHARD_LIBS_JSONCPP", which disables downloading "JSONCPP" as well?

Some additional benefit with these might be reduced memory requirement for building tensorflow; memory consumption of tensorflow-build gets dangerous at the last step with ```ld```. Probably, for now, I may simply need to be satisfied with letting ```make``` reduce ```-j#``` only for ```ld``` steps.

CC: @leemgs
STATUS: failing at the last step. (without ```-Dtensorflow_BUILD_SHARED_LIB```, it works anyway)
```
...
[ 2457s] /usr/lib64/gcc/x86_64-tizen-linux-gnu/6.2.1/../../../../x86_64-tizen-linux-gnu/bin/ld: libtf_core_gpu_kernels.a(tf_core_gpu_kernels_generated_beam_search_ops_gpu.cu.cc.o): relocation R_X86_64_32 against `.data' can not be used when making a shared object; recompile with -fPIC
[ 2457s] /usr/lib64/gcc/x86_64-tizen-linux-gnu/6.2.1/../../../../x86_64-tizen-linux-gnu/bin/ld: libtf_core_gpu_kernels.a(tf_core_gpu_kernels_generated_resampler_ops_gpu.cu.cc.o): relocation R_X86_64_32 against `.data' can not be used when making a shared object; recompile with -fPIC
[...]

PR #14016, opened 2017-10-27
CMake-Linux support for GPU-build

myungjoo:
There had been some fixes for Linux
- Enable to specify or find related libraries (cuda/cudnn)
- Enable "find_package(CUDA)"
- Address the Linux Cmake-parser issue with TF_EXTRA_CUDA_CAPABILITIES=3.0,3.5,5.2
- Improve the method to find static libraries
- Enable tf_stream_executor for Linux-CMake build by linking to "libgomp"
- Adjust pywarp_tensorflow_lib configuration
- Hide build option not recognized at Linux toolchains
- Add resampler cu.cc files
- Don't use tf_core_kernels_cpu_only for Linux (it's for Windows only)

Tested with OBS (open build service) + rpmbuild for bare-metal (minimal packages are installed) x64 Linux.

Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>


myungjoo:
Addressing #14014

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -39,8 +39,19 @@ if (NOT WIN32)
   # Threads: defines CMAKE_THREAD_LIBS_INIT and adds -pthread compile option
   # for targets that link ${CMAKE_THREAD_LIBS_INIT}.
   find_package (Threads)
+
+  option(tensorflow_PATH_STATIC_LIB "Additional library search path for libcudnn_static.a, libnccl_static.a, libculibos.a" /usr/local/cuda/lib64/)
```
mrry: Do we need this option? For example, is there an underlying option that will cause `find_package(CUDA 8.0 REQUIRED)` to surface these same paths?

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -39,8 +39,19 @@ if (NOT WIN32)
...
+  option(tensorflow_CUDNN_INCLUDE "cudnn.h header install path" /usr/include/)
 endif()
 
+if (WIN32)
+  set(BOOL_WIN32 ON)
```
mrry: Do we need a separate `BOOL_WIN32` symbol when we already have `WIN32`?

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -219,64 +230,114 @@ if(UNIX)
 endif()
 
 if (tensorflow_ENABLE_GPU)
+  if (NOT WIN32)
+    # Default install paths for cuda libraries in Linux
+    list(APPEND CMAKE_LIBRARY_PATH /usr/local/cuda/lib64)
+    list(APPEND CMAKE_LIBRARY_PATH /usr/local/cuda/lib64/stubs)
```
mrry: Is there a way to get this information from `find_package(CUDA 8.0 REQUIRED)` or should this be an option with a default value?

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -219,64 +230,114 @@ if(UNIX)
...
+  )
+
+  if (WIN32)
     # tf assumes in various places header files to be in cuda/include. On windows the cuda sdk
     # installs them under cuda/version/include and to avoid that we need to change tf we copy a
     # few files to cuda/include
     FILE(COPY
       ${CUDA_TOOLKIT_TARGET_DIR}/include/cuda.h ${CUDA_TOOLKIT_TARGET_DIR}/include/cuComplex.h
-      ${CUDA_TOOLKIT_TARGET_DIR}/include/cublas_v2.h ${CUDNN_HOME}/include/cudnn.h
+      ${CUDA_TOOLKIT_TARGET_DIR}/include/cublas_v2.h ${CUDNN_HOME}/cudnn.h
```
mrry: This change to the source path for `cudnn.h` seems like it might break the Windows build. Can you revert it?

tensorflow/contrib/cmake/CMakeLists.txt
```diff
@@ -219,64 +230,114 @@ if(UNIX)
...
+    )
   endif(WIN32)
+
+  include_directories(${tensorflow_source_dir}/third_party/gpus)
+  # add cuda libraries to tensorflow_EXTERNAL_LIBRARIES
+  list(APPEND tensorflow_EXTERNAL_LIBRARIES ${CUDA_LIBRARIES})
+
+  # NOTE(mrry): Update these flags when the version of CUDA or cuDNN used
+  # in the default build is upgraded.
+  set(tensorflow_BUILD_INFO_FLAGS --build_config cuda --key_value
```
mrry: This should probably be in an `if(WIN32)` block.

tensorflow/contrib/cmake/tf_python.cmake
```diff
@@ -957,7 +960,7 @@ add_library(pywrap_tensorflow_internal SHARED
     $<TARGET_OBJECTS:tf_tools_transform_graph_lib>
     $<$<BOOL:${tensorflow_ENABLE_GRPC_SUPPORT}>:$<TARGET_OBJECTS:tf_core_distributed_runtime>>
     $<TARGET_OBJECTS:tf_core_kernels>
-    $<$<BOOL:${tensorflow_ENABLE_GPU}>:$<TARGET_OBJECTS:tf_core_kernels_cpu_only>>
+    $<$<BOOL:${tensorflow_ENABLE_GPU}>:$<$<BOOL:${BOOL_WIN32}>:$<TARGET_OBJECTS:tf_core_kernels_cpu_only>>>
```
mrry: I'm unsure why this particular target is WIN32 only. Can you please add a comment indicating why this is platform-dependent?

tensorflow/contrib/cmake/tf_tools.cmake
```diff
@@ -73,7 +73,7 @@ add_executable(${transform_graph}
     $<TARGET_OBJECTS:tf_core_direct_session>
     $<TARGET_OBJECTS:tf_tools_transform_graph_lib>
     $<TARGET_OBJECTS:tf_core_kernels>
-    $<$<BOOL:${tensorflow_ENABLE_GPU}>:$<TARGET_OBJECTS:tf_core_kernels_cpu_only>>
+    $<$<BOOL:${tensorflow_ENABLE_GPU}>:$<$<BOOL:${BOOL_WIN32}>:$<TARGET_OBJECTS:tf_core_kernels_cpu_only>>>
```
mrry: I'm not sure why this tool (and the subsequent ones) depend on the kernels at all. It seems like this might be making the executables *much* larger than they need to be! 

@petewarden Do any of these tools need to run kernels? (I think I've confirmed the `summarize_graph` doesn't, but I'm not sure about the others.)

mrry (changes requested):
Thanks for this contribution! It will be great to merge this and avoid gratuitous incompatibility between the Windows and Linux CMake builds :).

I do have a few comments, mostly around how we configure this. Please take a look.

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: Cudnn and nccl may be installed at a different path. (separated from Cuda)

Such issues get worse if users do not use popular Linux distro.

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: Is is because WIN32 cannot be used in '''<BOOL:${BOOL_WIN32}>:...'''

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: In my case, find_package(CUDA 8.0 REQUIRED) had failed if CMAKE_LIBRARY_PATH did not include /usr/local/cuda/lib64. This could be caused by incomplete packaging code of my own CUDA RPM packaging code; however, I'm not sure why yet.

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: Ah.. you are right. This should've not changed. I'll update the commit.

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: I first thought so. But, ```cuda_version_number=8.0``` and ```cudnn_version_number=6``` are still required for Linux and in Linux other "dll" parts are ignored anyway.

tensorflow/contrib/cmake/tf_python.cmake
myungjoo: That ```tf_core_kernels_cpu_only``` is for Windows only, which has sort of "workaround" for Windows that does not support GPU in specific features. In Linux, this ```tf_core_kernel_cpu_only``` is not necessary.

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: This is corrected with a1b7fe6

tensorflow/contrib/cmake/tf_tools.cmake
petewarden: The transform_graph tool does require ops to be linked in, since it can run transform rules like fold_constants that actually run parts of the graph.

mrry: Thanks for clarifying, Pete! Perhaps we can remove the kernels altogether from `summarize_graph` and `compare_graphs` then?

tensorflow/contrib/cmake/CMakeLists.txt
mrry: Perhaps we could move these inline paths to options? (It seems redundant with `tensorflow_PATH_STATIC_LIB`, for example.)

tensorflow/contrib/cmake/CMakeLists.txt
mrry: Perhaps we should have two options, `tensorflow_PATH_CUDNN_STATIC_LIB` and `tensorflow_PATH_NCCL_STATIC_LIB`, with some common default?

tensorflow/contrib/cmake/CMakeLists.txt
mrry: In that case, can you remove the `.dll` parts when building for Linux? It would be potentially confusing if `build_info.py` contained Windows-specific filenames.

mrry (changes requested):

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: Adding options for these (additional ```CMAKE_LIBRARY_PATH```) looks better.
However, they are not redundant with ```tensorflow_PATH_STATIC_LIB``` because they are not meant to be static libraries (.a); they are for shared libraries (.so).

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: Yup. That I'll do.

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: Ok. I'll add these two to override the value of ```tensorflow_PATH_STATIC_LIB```:
```
  option(tensorflow_PATH_CUDNN_STATIC_LIB "Override PATH_STATIC_LIB for libcudnn_static.a" ${tensorflow_PATH_STATIC_LIB})              
  option(tensorflow_PATH_NCCL_STATIC_LIB "Override PATH_STATIC_LIB for libnccl_static.a" ${tensorflow_PATH_STATIC_LIB}) 
```

tensorflow/contrib/cmake/CMakeLists.txt
myungjoo: I've added the option:
```
option(tensorflow_CUDA_LIBRARY_PATH "Designate the default CUDA library paths" /usr/local/cuda/lib64)
```

mrry (approved):
Looks good to me.

mrry:

mrry:
I just kicked off a Windows GPU build to check that there are no regressions there either: http://ci.tensorflow.org/view/All/job/tf-pr-win-cmake-gpu/15/

gunan:
@martinwicke also ready to merge

tensorflow/contrib/cmake/CMakeLists.txt
headupinclouds: @mrry : I'm trying to extend this update to finish CMake + OSX testing w/ (NVIDIA) GPU.   I had started with CUDA 8.0 and cuDNN 6.0, but given recent activity here, it seems better to target master w/ CUDA 9.0 + cuDNN 7.0.  Since this is closely related, It seemed better to keep related discussion in this thread, but I can create a new issue if preferred.   The OS X CUDA installation does not come with `nccl` (see `find` output below for contents).  NVIDIA has a (login based) `nccl` download for v2.0, but it only has Linux libraries.  I see three other options to support OS X + GPU Builds:

1. https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/nccl (build from bundled source v1.3.5)
2. https://github.com/nvidia/nccl (build from source v1.?)
3. add `option(tensroflow_USE_NCCL "Use NCCL lib" ON)` and update code

Shall I add the bundled `nccl` (v1.3.5?) files to the CMake build?.  I see notes about migrating to v2.0 in commit comments, which seems to be closed source.  If that's the case, then it may be best to try to make any internal tensorflow `nccl` operations optional in order to support single GPU configurations to start with. 

`find /Developer/NVIDIA/CUDA-9.0/lib -name "*.a"`
```
/Developer/NVIDIA/CUDA-9.0/lib/libcublas_device.a
/Developer/NVIDIA/CUDA-9.0/lib/libcublas_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libcudadevrt.a
/Developer/NVIDIA/CUDA-9.0/lib/libcudart_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libcufft_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libcufftw_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libculibos.a
/Developer/NVIDIA/CUDA-9.0/lib/libcurand_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libcusolver_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libcusparse_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppc_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppial_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppicc_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppicom_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppidei_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppif_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppig_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppim_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppist_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppisu_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnppitc_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnpps_static.a
/Developer/NVIDIA/CUDA-9.0/lib/libnvgraph_static.a
```

PR #14024, opened 2017-10-27
support DepthwiseConv2dNative op when use python/tools/optimize_for_i…

dnecho:
support DepthwiseConv2dNative op when use python/tools/optimize_for_inference.py


martinwicke:
@petewarden can you take a look?

petewarden (changes requested):
We're trying to encourage people to migrate to the Graph Transform Tool:
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms/#optimizing-for-deployment

If possible, it would be great if you could make this change in that path too, probably in this file (and the related tests):

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/graph_transforms/fold_batch_norms.cc

Let me know if that's possible, thanks for the contribution!

drpngx:
@dnecho feel free to reopen if you have more comments here.

PR #14031, opened 2017-10-27
Fix incorrect annotation tag in the docs of tf.Variable

yongtang:
In tf.Variable the annotation tag of `@compatiblity` (missing `i`) should be `@compatibility` , without the fix the rendering of docs may be incorrect. 

See `add_check_numerics_ops()` for an example of `@compatibility`:
https://github.com/tensorflow/tensorflow/blob/e43e514b7418756a828c6a0f60e43aa6a638e961/tensorflow/python/ops/numerics.py#L68-L72

### DIFF
--- a/tensorflow/python/ops/variables.py
+++ b/tensorflow/python/ops/variables.py
@@ -119,7 +119,7 @@ class Variable(object):
   various `Optimizer` classes use this collection as the default list of
   variables to optimize.
 
-  @compatiblity(eager)
+  @compatibility(eager)
   `tf.Variable` is not compatible with eager execution.  Use
   `tfe.Variable` instead which is compatable with both eager execution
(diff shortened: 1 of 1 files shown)


benoitsteiner (approved):

benoitsteiner:
Jenkins, test this please

yongtang:
Thanks @benoitsteiner for the review. The Jenkins CI error on `//tensorflow/tools/docs:build_docs_test` is because doc string for compatibility can only be specified at the method level, not at the class level (See Ln 120 in `tensorflow/tools/docs/pretty_docs.py`).

I have move the `@compatibility(eager)` from `Class` level docstring to `__init__` docstring. The test should pass now.

benoitsteiner:
Jenkins, test this please

PR #14040, opened 2017-10-27
Add go format check as part of the sanity check

yongtang:
In go, it is very common to format the code with `gofmt -s -w file.go`. This fix adds the gofmt check as part of the sanity check.

There was only one file `tensorflow/go/tensor.go` in the whole `tensorflow/go` directory that is not properly formatted.

This fix formatted `tensorflow/go/tensor.go` with `gofmt -s -w` as well.

It is also possible to check the format status in:
https://goreportcard.com/report/github.com/tensorflow/tensorflow

### DIFF
--- a/tensorflow/go/tensor.go
+++ b/tensorflow/go/tensor.go
@@ -101,7 +101,7 @@ func NewTensor(value interface{}) (*Tensor, error) {
 			return nil, bug("NewTensor incorrectly calculated the size of a tensor with type %v and shape %v as %v bytes instead of %v", dataType, shape, nbytes, buf.Len())
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


asimshankar:
Thanks for the PR @yongtang .

However, for now let's not do this in the sanity check. We currently don't enforce formatting for C++, Python, or Java in the sanity check and I'd prefer to not to special case go for now.

Happy to accept the change to `tensor.go` though.

Thanks!

yongtang:
Thanks @asimshankar. I have updated the PR and removed the sanity check part. Only the changes in `tensor.go` remains in the PR. Please take a look.

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please.

Issue #14001, opened 2017-10-26
layers-based cuDNN RNN functionality not working


alquraishi:
I'm trying to use the newly added layers-style cuDNN RNN functionality. I'm running TF 1.4.0-rc0 on Ubuntu with Pascal GPUs, compiled from source with CUDA 8 and cuDNN 7. When trying to import the relevant library:

`from tensorflow.contrib.cudnn_rnn.python.layers import cudnn_rnn`

I get the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/tensorflow-1.4.0-rc0/local/lib/python2.7/site-packages/tensorflow/contrib/cudnn_rnn/python/layers/cudnn_rnn.py", line 33, in <module>
    resource_loader.get_path_to_datafile("_cudnn_rnn_ops.so"))
  File "/usr/local/tensorflow-1.4.0-rc0/local/lib/python2.7/site-packages/tensorflow/contrib/util/loader.py", line 55, in load_op_library
    ret = load_library.load_op_library(path)
  File "/usr/local/tensorflow-1.4.0-rc0/local/lib/python2.7/site-packages/tensorflow/python/framework/load_library.py", line 56, in load_op_library
    lib_handle = py_tf.TF_LoadLibrary(library_filename, status)
  File "/usr/local/tensorflow-1.4.0-rc0/local/lib/python2.7/site-packages/tensorflow/python/framework/errors_impl.py", line 473, in __exit__
    c_api.TF_GetCode(self.status.status))
tensorflow.python.framework.errors_impl.NotFoundError: /usr/local/tensorflow-1.4.0-rc0/local/lib/python2.7/site-packages/tensorflow/contrib/cudnn_rnn/python/layers/_cudnn_rnn_ops.so: cannot open shared object file: No such file or directory
```

Note that the non-layer based cuDNN RNN functionality works fine. I.e. I can run this:

`from tensorflow.contrib.cudnn_rnn.python.ops import cudnn_rnn_ops`

and run cuDNN-based RNNs with no problem otherwise.

printdhruv:
@alquraishi Did you try to set dynamic path to solve the error `cannot open shared object file: No such file or directory`?
Can you fill the ISSUE_TEMPLATE [here](https://github.com/printdhruv/tensorflow/blob/master/ISSUE_TEMPLATE.md)?

alquraishi:
[...]

PR #14043, opened 2017-10-27
Update cudnn_rnn.py

printdhruv:
For addressing this issue,
https://github.com/tensorflow/tensorflow/issues/14001#issuecomment-340042335

Updating the path on line 33 to make import cudnn_rnn work. Change worked on Ubuntu.

### DIFF
--- a/tensorflow/contrib/cudnn_rnn/python/layers/cudnn_rnn.py
+++ b/tensorflow/contrib/cudnn_rnn/python/layers/cudnn_rnn.py
@@ -30,7 +30,7 @@ from tensorflow.python.platform import resource_loader
 from tensorflow.python.platform import tf_logging as logging
 
 _cudnn_rnn_ops_so = loader.load_op_library(
-    resource_loader.get_path_to_datafile("_cudnn_rnn_ops.so"))
+    resource_loader.get_path_to_datafile("../ops/_cudnn_rnn_ops.so"))
 
 CUDNN_RNN_UNIDIRECTION = cudnn_rnn_ops.CUDNN_RNN_UNIDIRECTION
 CUDNN_RNN_BIDIRECTION = cudnn_rnn_ops.CUDNN_RNN_BIDIRECTION


protoget:
LGTM.
I think something might stil bel missing for windows as it fails our release test.

printdhruv:
@protoget I just changed the path lookup? What you think the fix should be?

tensorflow/contrib/cudnn_rnn/python/layers/cudnn_rnn.py
```diff
@@ -30,7 +30,7 @@
 from tensorflow.python.platform import tf_logging as logging
 
 _cudnn_rnn_ops_so = loader.load_op_library(
-    resource_loader.get_path_to_datafile("_cudnn_rnn_ops.so"))
+    resource_loader.get_path_to_datafile("../ops/_cudnn_rnn_ops.so"))
```
asimshankar: Can we simply remove this line instead? The earlier line:

```python
from tensorflow.contrib.cudnn_rnn.python.ops import cudnn_rnn_ops
```

imports the `cudnn_rnn_ops` module, which in turn loads the right library anyway, right?

protoget: Yes I think it'd work. Let's see if it passes the test.

printdhruv: @protoget Did it pass on windows?

protoget: L33 has been removed in the nightly build. It isn't needed since cudnn_rnn.py imports cudnn_rnn_ops which imports the .so file already.

printdhruv: Great.

asimshankar: Yup, this was done in https://github.com/tensorflow/tensorflow/commit/a6a61884396ef1d51b01f8e13df21becb23fd0c8

which subsumes this PR.

Thanks!

PR #14056, opened 2017-10-28
Half Normal Distribution (and inverse error function)

cshenton:
## Inverse Error Function

Added `erfinv` as a simple function of `ndtri` (the api mirrors the approach taken in `scipy.special`. Mainly to prevent distributions with methods in terms of the inverse error function reimplementing their own wrappers around `ndtri`.

## Half Normal Distribution
```python
X ~ Normal(0.0, scale)
Y = |X|
# then Y ~ HalfNormal(scale)
```

Added the Half Normal distribution to `contrib.distributions`. Main things to look at is how I'm dealing with the pdf discontinuity.

A discontinuity in the pdf that sends values to 0.0 is easy to represent in `prob`, (by multiplying my a tensor with 0.0s at the relevant positions) than `log_prob` (which requires masking -inf wherever x < 0). I'd be interested if anyone has any suggestions for correct ways to implement the latter, as I think this is a case where `log_prob` is more numerically stable (along the support of the distribution that is).

### DIFF
--- a/tensorflow/contrib/distributions/python/kernel_tests/half_normal_test.py
+++ b/tensorflow/contrib/distributions/python/kernel_tests/half_normal_test.py
@@ -1,0 +1,317 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for initializers."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import importlib
+import numpy as np
+
+from tensorflow.python.framework import constant_op
+from tensorflow.python.framework import dtypes
+from tensorflow.python.framework import ops
+from tensorflow.python.framework import tensor_shape
+from tensorflow.python.ops import array_ops
+from tensorflow.python.ops import gradients_impl
+from tensorflow.python.ops import nn_ops
+from tensorflow.python.ops import variables
+from tensorflow.contrib.distributions.python.ops import half_normal as hn_lib
+from tensorflow.python.platform import test
+from tensorflow.python.platform import tf_logging
+
+
+def try_import(name):  # pylint: disable=invalid-name
+  module = None
+  try:
+    module = importlib.import_module(name)
+  except ImportError as e:
+    tf_logging.warning("Could not import %s: %s" % (name, str(e)))
+  return module
+
+stats = try_import("scipy.stats")
+
+
+class HalfNormalTest(test.TestCase):
+
+  def setUp(self):
+    self._rng = np.random.RandomState(123)
+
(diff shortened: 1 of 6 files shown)


cshenton:
Sanity checks failure seems to be unrelated to the content of the PR:
```
The command '/bin/sh -c /install/install_deb_packages.sh' returned a non-zero code: 100
ERROR: docker build failed. Dockerfile is at /var/lib/jenkins/workspace/tensorflow-pull-requests-sanity/tensorflow/tools/ci_build/Dockerfile.cpu
```

gunan:
Jenkins, test this please.

ebrevdo:
The details mostly look good.  Is there a reason you didn't use the [absolutevalue Bijector](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/bijectors/absolute_value_impl.py#L34) to implement this class?  It would take care of sampling and log_prob for you.

cshenton:
Thank for bringing that to my attention. If that's available, I'll implement a the `FoldedNormal` with `HalfNormal` as a special case.

As an aside, the docs state that bijectors implement bijective differentiable transforms, but the absolute value function is not a bijection. Am I right in guessing bijectors is moving towards being a more general set of transforms for distributions?

ebrevdo:
Yes; it's an extension now that bijector inverses (and inverse log det
jacobian functions) can return tuples - under the assumption that the
preimage is always len(tuple)-valued.  I believe that the special case of
t=0 should continue to work for HalfNormal due to the way
TransformedDistribution treats these values, but if not then you may have
to special case that value.

cshenton:
Fixed things up. Tests were previously failing since the scipy `halfnorm`'s first arg was not `scale` as I assumed, but `loc`, where loc is both the mean of the base distribution and the fold point.

Also I didn't end up using the bijector and subclassing `TransformedDistribution`, because the static shape of the outputs of `prob` and `log_prob` were both `<unknown>`.

cshenton:
Could someone manually trigger the test run? Container build failed on the sanity tests for some reason.
```
ERROR: docker build failed. Dockerfile is at /var/lib/jenkins/workspace/tensorflow-pull-requests-sanity/tensorflow/tools/ci_build/Dockerfile.cpu
Build was aborted
Aborted by unknown
```

ebrevdo:

cshenton:
Looks like one of the macOS tests timed out, so tests will need to be rerun. Apologies for the trouble. 
```
//tensorflow/python/kernel_tests:slice_op_test                          TIMEOUT in 301.0s
```

ebrevdo (approved):

tensorflow/contrib/distributions/python/kernel_tests/half_normal_test.py
```diff
@@ -0,0 +1,318 @@
...
+            variables.global_variables_initializer().run()
+            self.assertAllFinite(value)
+            self.assertAllFinite(grads[0])
+
+  def testHalfNormalEntropy(self):
+    with self.test_session():
+      scale = np.array([[1.0, 2.0, 3.0]])
+      halfnorm = hn_lib.HalfNormal(scale=scale)
+
+      expected_entropy = 0.5 * np.log(np.pi * scale ** 2.0 / 2.0) + 0.5
```
ebrevdo: add a reference for this calculation?

cshenton: Will do. Is a link to wikipedia fine? There's the 1961 paper on the folded normal distributions but that's paywalled so less easy to check.

ebrevdo:
Wiki is fine

cshenton:
Do the remaining checks need to be triggered, or is the Ubuntu CC just taking a really long time?

PR #14073, opened 2017-10-29
Explicitly pass label_count to confusion_matrix.

baranov1ch:
If batch size gets small enough, there can be not enough classes
in the validation batch, and auto-inference on each step may
return different sizes of the confustion matrices, and their
addition will fail


petewarden (approved):
Looks great, thanks for this change!

baranov1ch:
resolved in https://github.com/tensorflow/tensorflow/pull/14600

Issue #1419, opened 2016-03-07
Dependency required for user Reader ops?


sbyma:
When creating a Reader op inside core/user_ops that inherits from ReaderBase, Bazel responds with 

```
tensorflow/core/BUILD:325:1: undeclared inclusion(s) in rule '//tensorflow/core:user_ops_op_lib':
this rule is missing dependency declarations for the following files included by 'tensorflow/core/user_ops/user_reader_op.cc':
  'bazel-out/local_darwin-opt/genfiles/tensorflow/core/kernels/reader_base.pb.h'.
```

Is this because the user_ops_op_lib depends only on framework? user_reader_op.cc includes reader_base.h which includes reader_base.pb.h. What dependency should I add to make this work? I'm still figuring out Bazel atm. 

This is with latest source code.

sbyma:
I was able to resolve this by changing tensorflow/core/BUILD as follows at line 325:

```
cc_library(
    name = "user_ops_op_lib",
    srcs = glob(["user_ops/**/*.cc"]),
    copts = tf_copts(),
    linkstatic = 1,
    visibility = ["//visibility:public"],
    deps = [":framework", "//tensorflow/core/kernels:reader_base"],
    alwayslink = 1,
)

```

Works for now but a broader dependency might be better?

girving:
@keveman: What's the situation with user_ops these days?  Was it going through some flux?

keveman:
Not that I know of. There is #1378 for which I'll have a fix soon.

keveman:
@sbyma Can you please verify and close this issue?

crjensen21:
I'm getting a very similar error that I can't get around. If I add //tensorflow/core/kernels:reader_base to the Bazel dependencies, though, I get this error:

tensorflow/core/kernels:reader_base cannot depend on tensorflow/core:framework.

It seems both these builds have the other listed in their disallow_deps attribute. I can't find a workaround.

yaroslavvb:
I had the same issue recently when trying to add dependency to `tensorflow/core:core` from my custom op -- https://github.com/tensorflow/tensorflow/issues/7030#issuecomment-274882446

I used bazel query to make sure `core` doesn't depend on framework, so I'm not sure what's going on, my guess is that custom tf_custom_op Bazel macro does something funky
`bazel query "somepath(//tensorflow/core:core, //tensorflow/core:framework)"
`

My work-around is to use `g++` to compile the op instead of Bazel

[...]

PR #14098, opened 2017-10-30
Add KafkaReader for processing streaming data with Apache Kafka

yongtang:
This is a proposal to add KafkaReader so that it is possible to read data from Kafka like TextLineReader and TFRecordReader.

Apache Kafka is a widely used distributed streaming platform in open source community. The goal of this fix is to create a contrib Reader ops (inherits ReaderBase and is similiar to TextLineReader / TFRecordReader) so that it is possible to reader Kafka streaming data from TensorFlow in a similiar fashion.

This fix uses a C/C++ Apache Kafka client library librdkafka which is released under the 2-clause BSD license, and is widely used in a number of Kafka bindings such as Go, Python, C#/.Net, etc.

Below is a sample usage:
```python
import tensorflow as tf
import tensorflow.contrib.kafka as kafka

# The name of test:0:0:10
# test: topic
# 0: partition
# 0: offset
# 10: length (or -1 for unlimited)
#
# The parameters for KafkaReader
# servers: bootstrap.servers
# group: group.id
# eof: If True then read will block until the length of the data
#      has been read and ignore EOF error from Kafka

filename_queue = tf.train.string_input_producer(["test:0:0:10"])

reader = kafka.KafkaReader(servers="localhost", group="test", eof=True)
key, value = reader.read(filename_queue)
with tf.Session() as sess:
  coord = tf.train.Coordinator()
  threads = tf.train.start_queue_runners(coord=coord)
  for i in range(10):
    print(sess.run([key, value]))
  coord.request_stop()
  coord.join(threads)
```

NOTE: Because of #1419, the clean_deps of `tensorflow/core:framework` and `tensorflow/core:lib` has been commented out temporarily so that it is possible to build with ReaderBase. Any suggestions to address this issue is welcomed.


martinwicke:
@jhseu can you take a look?

tensorflow/contrib/kafka/kernels/kafka_reader_ops.cc
```diff
@@ -0,0 +1,190 @@
...
+
+  Status ReadLocked(string* key, string* value, bool* produced,
+                    bool* at_end) override {
+    if (limit_ >= 0 &&
+        (topic_partition_->offset() >= limit_ || offset_ >= limit_)) {
+      *at_end = true;
+      return Status::OK();
+    }
+    while (true) {
+      std::unique_ptr<RdKafka::Message> message(consumer_->consume(1000));
```
terrytangyuan: Not too familiar with `rdkafkacpp` but any reasons behind this hard-coded 1000? Is this a standard practice?

yongtang: Thanks @terrytangyuan. The `1000` is the timeout (in `ms`) for consume() call to return. The call will repeat in the loop. The value was used as in the demo code of librdkafka.

I have updated the PR so that an optional attribute `timeout` could be passed to control this value (instead of hard coded `1000`). Please take a look.

jhseu:
Seems reasonable to add, but can we implement this as a Dataset instead? See `tensorflow/core/kernels/reader_dataset_ops.cc`

jhseu:
The queue style of reading will be deprecated.

yongtang:
Thanks @jhseu for the feedback! Let me take a look and update the PR accordingly.

yongtang:
It seems that Dataset requires `tensorflow/core/kernels/dataset.cc` to link as a `.so`. However, that is problematic becuase the class of `Dataset` will exist in two places and the kernel will not recognize the one in `.so`.

That is quite similar to the issue faced in  #1419. Not sure how to get around it other than building the KafkaDataset into the core kernel.

jhseu:
@mrry @allenlavoie Derek or Allen, any thoughts on how we might make a dataset in contrib that gets exposed as a shared library? I'm guessing it would just require moving some of the classes over to the framework directory?

mrry:
@jhseu It should be possible in the next week or two. The current layering is bit more intricate than that, but a couple of changes in flight will enable us to remove the current dependency on `core/common_runtime` and allow building `DatasetOpKernel` and `DatasetBase` subclasses in a separate shared library.

yongtang:
Thanks @jhseu @mrry, please keep me posted on the changes in Dataset so that I could update the PR accordingly.

drpngx:
@jhseu @mrry have the changes been pushed? Thanks.

mrry:
Not yet: we're waiting on a CL from @rohan100jain that will sever the dependency between `captured_function.cc` and `tensorflow/core/common_runtime/...`, so we can split all the `tf.contrib.data` kernels into a separate library that only depends on `tensorflow/core/framework/...`.

drpngx:
Thanks! @rohan100jain please ping this thread once ready.

tensorflowbutler:

drpngx:
@mrry @rohan100jain was the CL submitted?

rohan100jain:
Yes that change was submitted. https://github.com/tensorflow/tensorflow/commit/9f4118d00fa9eb85f81a4eb3f96a5583ae5afcdc

drpngx:
OK, @jhseu I think it's ready for review.

tensorflow/contrib/kafka/kernels/kafka_dataset_ops.cc
```diff
@@ -0,0 +1,321 @@
...
+          topics_(std::move(topics)),
+          servers_(servers),
+          group_(group),
+          eof_(eof),
+          timeout_(timeout) {}
+
+    std::unique_ptr<IteratorBase> MakeIterator(
+        const string& prefix) const override {
+      return std::unique_ptr<IteratorBase>(
+          new Iterator({this, strings::StrCat(prefix, "::TextLine")}));
```
jhseu: Kafka instead of TextLine?

tensorflow/contrib/kafka/BUILD
```diff
@@ -0,0 +1,93 @@
...
+        "//tensorflow/python:platform",
+        "//tensorflow/python:state_ops",
+        "//tensorflow/python:training",
+        "//tensorflow/python/data/ops:dataset_ops",
+        "//tensorflow/python/data/ops:iterator_ops",
+        "//tensorflow/python/data/ops:readers",
+    ],
+)
+
+tf_py_test(
```
jhseu: Can you give instructions for running the test in a comment here?

tensorflow/contrib/kafka/BUILD
```diff
@@ -0,0 +1,93 @@
...
+    srcs = ["kernels/kafka_dataset_ops.cc"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//tensorflow/core:framework",
+        "//tensorflow/core:lib",
+        "//tensorflow/core:lib_internal",
+        "//tensorflow/core/kernels:bounds_check_lib",
+        "//tensorflow/core/kernels:dataset",
+        "//third_party/eigen3",
+        "@kafka//:kafka",
```
jhseu: Can we make this an optional dependency that's set through ./configure? The code can be wrapped around #ifdef if kafka is enabled. See other directories for examples.

yongtang:
Thanks @jhseu for the review. The PR has been updated with comments addressed. Please take a look and let me know if there are any issues.

jhseu (approved):

jhseu:
Thanks! This is greatly appreciated.

PR #14099, opened 2017-10-30
fix fused_batchnorm_test.py

jacquesqiao:
In file fused_batchnorm_test.py,  the function [_reference_training](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tests/fused_batchnorm_test.py#L39) use 
```python
element_count = np.size(x) / int(np.shape(x)[0])
```
to get element_count for `NHWC` data_format, but it should be 
```
element_count = np.size(x) / int(np.shape(x)[-1])
```
The last dimention is the channel number, the reason why it works now is that the test data is `[2,2,6,2]`   
```python
np.shape(x)[0] == np.shape(x)[-1] == 2
```

### DIFF
--- a/tensorflow/compiler/tests/fused_batchnorm_test.py
+++ b/tensorflow/compiler/tests/fused_batchnorm_test.py
@@ -36,7 +36,7 @@ class FusedBatchNormTest(XLATestCase):
     x_square = x * x
     x_square_sum = np.sum(x_square, (0, 1, 2))
     x_sum = np.sum(x, axis=(0, 1, 2))
-    element_count = np.size(x) / int(np.shape(x)[0])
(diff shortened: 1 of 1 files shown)


jacquesqiao:
I signed it!

jacquesqiao:
Can anyone help to review this PR? @tensorflow-jenkins

jacquesqiao:
what is the CI fail about? It does not have any error log.

gunan (approved):

gunan:
The issue looks like an interesting flake, I cant see how it can be related to your change. Will rerun tests to see what is going on.
Jenkins, test this please.

PR #14106, opened 2017-10-30
Batch normalization for LSTM

Andres-Hernandez:
A recurrent batch normalization is implemented based on https://arxiv.org/pdf/1603.09025.pdf. The new functionality allows for setting normalization for each of the inputs, the actual input or the state, and the cell prior to the activation function. To avoid a double implementation, tensorflow.python.layers.normalization.BatchNormalization is called to perform the normalization, and it can be controlled fully, but if no extra configuration is provided a good default configuraiton is used. By default no normalization is done. The only aspect mentioned in the paper that was not implemented, was that it was recommended there that the statistics be taken on a time step basis, rather than having all time steps share the same statistics. This would have necessitated a lot of tinkering in BatchNormalization, to the point that it would be more economical if that behaviour was desired, to simply port the parts needed into LSTM and adjust them as necessary, but that should not be part of the main branch.


martinwicke:
@ebrevdo, can you comment on this?

@Andres-Hernandez can you look into the conflicts? I am sorry it took us so long to get to this.

Andres-Hernandez:
I have resolved the conflicts, which were due to a refactoring of the LSTM cell which happened over the weekend between me getting my working copy and when I checked in on Monday. The refactoring that had been done including things like e.g. name as parameter and use of standard build function to create variables, instead of creating them in call the first time it is invoked. The new normalization code has been adapted to this more standard implementation.

ebrevdo (changes requested):
There is a lstm cell with layer norm in tf.contrib.rnn.  can you add a new cell in tf.contrib.rnn using the Layer norm cell as a template?  With unit tests?

martinwicke:
@Andres-Hernandez any progress? Marking as stalled for now.

drpngx:
@Andres-Hernandez feel free to reopen if you get a chance to reply. Thanks!

PR #14128, opened 2017-10-31
Fix typos in Linear Model Tutorial samples and install_sources.md

case540:
1. test_file_name is undefined (should be test_file.name)
2. train_file_name is undefined (should be train_file.name)

PiperOrigin-RevId: 173733442


case540:
So "Ubuntu Makefile" is broken for 1.4 branch. Ignore those failures.

av8ramit (approved):

PR #14137, opened 2017-10-31
[XLA] Rename symbols for OS/X only - it doesn't have sincos

DavidNorman:
On OS/X the sincos and sincosf are apparently not present.  There is a __sincos and __sincosf.  This change creates an alias for them on OS/X.


Timmmm:
The types should be as follows:

```
#if defined(__APPLE__)
static void sincos(double, double*, double*)  __attribute__((weakref ("__sincos")));
static void sincosf(float, float*, float*) __attribute__((weakref ("__sincosf")));
#endif
```

DavidNorman:
Updated - thanks @Timmmm

DavidNorman:
i believe this was merged with the other fix

PR #14166, opened 2017-11-02
Documentation fix - clarifying sum in softmax function.

nsuh:
This notational fix makes it more clear that you're summing over the classes in the denominator of the softmax function.

### DIFF
--- a/tensorflow/docs_src/get_started/mnist/beginners.md
+++ b/tensorflow/docs_src/get_started/mnist/beginners.md
@@ -184,7 +184,7 @@ $$\text{softmax}(evidence) = \text{normalize}(\exp(evidence))$$
 
 If you expand that equation out, you get:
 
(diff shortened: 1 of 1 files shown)


nsuh:
I signed it!

martinwicke (approved):

martinwicke (changes requested):
Actually, on second thought, I am not sure this is strictly an improvement. you say 1..10, the code would say 0..9. The real confusion probably stems from j being used earlier to index "over the pixels of our input image". Adding 1..10 gives some more hint in this contradictory situation, but not reusing this index may be better still (and probably good enough to skip the explicit 1..10).

martinwicke:
@nsuh any update? Marking as stalled.

drpngx:
@nsuh closing for now, feel free to reopen if you get a chance to address @martinwicke comments.

PR #14167, opened 2017-11-02
TensorFlow for NVIDIA Tegra devices with CUDA support

lihanchen:


tensorflow/contrib/makefile/Makefile
```diff
@@ -373,12 +373,67 @@ $(MARCH_OPTION) \
...
+
+		INCLUDES += \
+-Itensorflow/core/kernels \
+-I$(JETPACK)/cuda-8.0/include \
+-I$(JETPACK) \
+-I$(JETPACK)/cuDNN/aarch64 \
+-I$(JETPACK)/cuda-8.0/extras/CUPTI/include \
+-I. \
+-Itegra \
+-Itegra/external/cub_archive/cub/device
```
andrewharp: What is this for? Do we really need this whole archive?

tensorflow/core/kernels/conv_ops_gpu_2.cu.cc
```diff
@@ -26,11 +26,15 @@ namespace tensorflow {
 
 typedef Eigen::GpuDevice GPUDevice;
 template struct functor::InflatePadAndShuffle<GPUDevice, float, 4, int>;
+#ifndef NVIDIA_TEGRA
 template struct functor::InflatePadAndShuffle<GPUDevice, float, 4,
                                               Eigen::DenseIndex>;
```
andrewharp: Does Tegra not even support the compilation of DenseIndex?

tensorflow/contrib/makefile/build_all_android.sh
```diff
@@ -26,7 +26,7 @@ usage() {
   echo "-x [hexagon library path] copy and hexagon libraries in the specified path"
   echo "-a [architecture] Architecture of target android [default=armeabi-v7a] \
 (supported architecture list: \
-arm64-v8a armeabi armeabi-v7a mips mips64 x86 x86_64)"
+arm64-v8a armeabi armeabi-v7a mips mips64 x86 x86_64 Tegra)"
```
andrewharp: probably want to lower-case tegra here to stay consistent.

tensorflow/contrib/makefile/download_dependencies.sh
```diff
@@ -54,7 +54,7 @@ download_and_extract() {
   elif [[ "${url}" == *zip ]]; then
     tempdir=$(mktemp -d)
     tempdir2=$(mktemp -d)
-    wget -P ${tempdir} ${url}
+    wget --no-check-certificate -P ${tempdir} ${url}
```
andrewharp: Why is this necessary?

tensorflow/contrib/makefile/Makefile
lihanchen: -I. \ can be removed.

-I$(JETPACK)/cuda-8.0/extras/CUPTI/include and libcupti.so is used by tensorflow/core/platform/default/gpu_tracer.cc. If we don’t need gpu_tracer.cc we can remove this.

+-Itegra \ contains cuda_config.h is written by google. I don't think it plays any practical role, but it is included in many CUDA source files.

+-Itegra/external/cub_archive/cub/device is a lot of cuda headers.
Most of them can be found in Jetpack, but the way TF code include headers makes it unable to use the copy in Jetpack
For example, 
tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc:25: #include "external/cub_archive/cub/util_ptx.cuh"
There is a util.ptx.cuh in Jetpack_Android_3.0/cuda-8.0/targets/aarch64-linux-androideabi/include/thrust/system/cuda/detail/cub/util_ptx.cuh
However, I can’t handle the path in the inclusion “external/cub_archive/”.

Also, I notice there are at least two cuh files that TF needs but don’t exists in Jetpack.
tegra/external2/cub_archive/cub/device/device_segmented_radix_sort.cuh
tegra/external2/cub_archive/cub/device/device_segmented_reduce.cuh.cuh

tensorflow/core/kernels/conv_ops_gpu_2.cu.cc
lihanchen: For some reason, nvcc recognize DenseIndex as int, causing a duplicate definition error.

tensorflow/contrib/makefile/build_all_android.sh
lihanchen: I will change that

tensorflow/contrib/makefile/download_dependencies.sh
lihanchen: On my computer, it fails to download deps without this option:
downloading https://mirror.bazel.build/bitbucket.org/eigen/eigen/get/429aa5254200.tar.gz
downloading https://mirror.bazel.build/github.com/google/gemmlowp/archive/010bb3e71a26ca1d0884a167081d092b43563996.zip
--2017-11-03 14:25:06--  https://mirror.bazel.build/github.com/google/gemmlowp/archive/010bb3e71a26ca1d0884a167081d092b43563996.zip
Resolving mirror.bazel.build (mirror.bazel.build)... 130.211.22.235
Connecting to mirror.bazel.build (mirror.bazel.build)|130.211.22.235|:443... connected.
ERROR: cannot verify mirror.bazel.build's certificate, issued by ‘CN=Google Internet Authority G2,O=Google Inc,C=US’:
  Unable to locally verify the issuer's authority.
To connect to mirror.bazel.build insecurely, use `--no-check-certificate'.

lihanchen:
Hi,

I am added to the nvidia-google-contributors group, so the CLA should be fine.

tensorflow/contrib/makefile/Makefile
andrewharp: Yes, let's disable GPU-tracer for now and worry about supporting it later if necessary.

Do we need to use tegra/cuda_config.h vs third_party/toolchains/gpus/cuda/cuda/cuda_config.h?

Is "Jetpack_Android_3.0/cuda-8.0/targets/aarch64-linux-androideabi/include/thrust/system/cuda/detail/cub" completely analogous to "external/cub_archive"? If so maybe we can create a symlink instead?

For device_segmented_radix_sort.cuh and device_segmented_reduce.cuh.cuh, what happens if we just don't compile the including files?

lihanchen: The third_party/toolchains/gpus/cuda/cuda/cuda_config.h seems a little different, but it should be usable.

With a quick glance, the difference between Jetpack_Android_3.0/cuda-8.0/targets/aarch64-linux-androideabi/include/thrust/system/cuda/detail/cub and external/cub_archive are device_segmented_radix_sort.cuh and device_segmented_reduce.cuh.cuh

I will take a closer look on Monday.

If we don't include these two files. the compilation will fail. We need to exclude more than 10 source codes which seems unacceptable to me.

tensorflow/contrib/makefile/download_dependencies.sh
andrewharp: That's odd, but it seems unrelated to the CUDA functionality so should not be included in the PR. Is it possible that you're connecting through a proxy which is mangling the certificates?

lihanchen: I just verified that I don't need this flag from home (Comcast).
I will remove this change and figure out what is going on with my working machine

tensorflow/contrib/makefile/Makefile
```diff
@@ -687,6 +801,18 @@ $(PROTO_OBJS): $(PROTO_CC_SRCS) $(PBT_GEN_FILES)
 
 # Host compilation rules.
 
+# NVCC compilation rules for Tegra
+ifeq ($(BUILD_FOR_TEGRA),1)
+$(OBJDIR)%.cu.o: %.cu.cc
```
andrewharp: Would it be possible for you to verify this rule works with Make 3.81? I'm seeing that it's not getting matched and the build fails when it tries to use g++ to compile cu.cc files. I've tried moving it up and down in the rule list as well to see if it's an order thing, but it seems to make no difference.

tensorflow/contrib/makefile/Makefile
lihanchen: I tested that I can use third_party/toolchains/gpus/cuda/cuda/cuda_config.h.
For cub_archive, I found that several file is missing in Jetpack_Android_3.0/cuda-8.0/targets/aarch64-linux-androideabi/include/thrust/system/cuda/detail/cub, such as
dispatch/dispatch_reduce.cuh
tegra/external2/cub_archive/cub/device/device_segmented_radix_sort.cuhtegra/external2/cub_archive/cub/device/device_segmented_reduce.cuh.cuh
dispatch_reduce_by_key.cuh 
dispatch_scan.cuh

In conclusion, I think we need to download cub archive during compilation.
I got it by using bazel to build for Java:
./configure #choose Y for CUDA
bazel build --config opt \
  //tensorflow/java:tensorflow \
  //tensorflow/java:libtensorflow_jni

Can you point to me the url of cub archive?

tensorflow/contrib/makefile/Makefile
lihanchen: Make 3.81 was released in 2006.
I can't find a working binary and I had a lot of problem building from source.
However, can you try moving the

ifeq ($(BUILD_FOR_TEGRA),1)
$(OBJDIR)%.cu.o: %.cu.cc
	@mkdir -p $(dir $@)
	@mkdir -p $(dir $(DEPDIR)$*)
	$(NVCC) $(NVCCFLAGS) -Xcompiler "$(CXXFLAGS4NVCC) $(DEPFLAGS)" $(INCLUDES) -c $< -o $@

$(LIBDIR)libtfcuda.a: $(CUDA_CC_OBJS)
	@mkdir -p $(dir $@)
	$(AR) $(ARFLAGS) $@ $(CUDA_CC_OBJS)
endif


above

$(OBJDIR)%.o: %.cc | $(PBT_GEN_FILES)
	@mkdir -p $(dir $@)
	@mkdir -p $(dir $(DEPDIR)$*)
	$(CXX) $(CXXFLAGS) $(DEPFLAGS) $(INCLUDES) -c $< -o $@
	@mv -f $(DEPDIR)/$*.Td $(DEPDIR)/$*.d

It should work theoretically.

lihanchen: I tested on my machine. With above change, I can compile with GNU make 3.81

tensorflow/contrib/makefile/Makefile
andrewharp: For the bazel build the cub archive is downloaded via https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl#L772

If it's necessary, we can do a similar download of the archive for Make builds in download_dependencies.sh

andrewharp: Should be pretty simple to add it here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/makefile/download_dependencies.sh

lihanchen: For CUPTI, I found another dependancy chain that prevents me to remove libcupti.so
tensorflow/core/common_runtime/direct_session.cc -> tensorflow/core/platform/default/gpu_tracer.cc -> tensorflow/core/platform/default/gpu/cupti_wrapper.cc -> libcupti.so

Without direct_session, benchmark fails with
session.cc:58 Not found: No session factory registered for the given session options: {target: "" config: } Registered factories are {}

Should we keep libcupti.so?

lihanchen: Also, I noticed that TF is specifically looking for libcupti.so.8.0. It will throw an error if that lib doesn't exitst.
native : statusor.h:212 Non-OK-status: status_ status: Failed precondition: could not dlopen DSO: libcupti.so.8.0; dlerror: dlopen failed: library "libcupti.so.8.0" not found
I can't find where libcupti.so.8.0 is referenced in the source code.

drpngx:
@lihanchen could you pull rebase and push again? That makes it easier to review. Thanks!

drpngx:
@andrewharp could you take a look?

lihanchen:
@drpngx Hi. You asked me "@lihanchen could you pull rebase and push again? That makes it easier to review. Thanks!"
Do you mean to combine all the commits into one?

drpngx:
@lihanchen there were many commits so I thought it was a merge, but it wasn't. I have deleted that comment. You're good!

andrewharp:
Jenkins, test this please.

andrewharp (approved):

drpngx:
Jenkins, test this please.

drpngx:
Jenkins, test this please.

andrewharp:
Jenkins, test this please.

andrewharp:
Jenkins, test this please.

andrewharp:
Jenkins, test this please.

andrewharp:
@zheng-xq The failures seem to be unrelated, can we go ahead and merge this in?

zheng-xq:
If the test are failing at head, and can be fixed soon, maybe wait until that. But if it takes too long, it's okay to merge. Thanks!

andrewharp:
Jenkins, test this please.

PR #14172, opened 2017-11-02
Eliminate ambiguity

larrytin:
[batch_size] + shape means the sum of two matrix


tensorflow/python/feature_column/feature_column.py
```diff
@@ -586,7 +586,7 @@ def numeric_column(key,
     shape: An iterable of integers specifies the shape of the `Tensor`. An
       integer can be given which means a single dimension `Tensor` with given
       width. The `Tensor` representing the column will have the shape of
-      [batch_size] + `shape`.
```
gunan: I disagree. I think you need to see this as follows:
```
batch_size = 32
shape = [12, 22, 42]
[batch_size] + shape = [32, 12, 22, 42]
```
However, with your proposal:
```
(batch_size, shape) = (32, [12, 22, 42])
```

Thoughts?

PR #14183, opened 2017-11-02
improve production condition of debug log with VLOG

nutti:
This patch improve a debug logging **VLOG** .

New environment variable ```TF_CPP_MAX_VLOG_LEVEL``` is added to suppress log production with VLOG.
if ```TF_CPP_MAX_VLOG_LEVEL``` is set, VLOG only outputs log lower level than it.

For example, if an user sets environment variable ```TF_CPP_MAX_VLOG_LEVEL=2```, you can see the log only lower level than 2. And you can't see the log with level 1.

### Use case:

1. ```TF_CPP_MAX_VLOG_LEVEL=2``` and ```TF_CPP_MIN_VLOG_LEVEL=1```
   * ```VLOG(1)``` and ```VLOG(2)``` will output log
2. ```TF_CPP_MIN_VLOG_LEVEL=1```
   * Only ```VLOG(1)``` will output
3. ```TF_CPP_MAX_VLOG_LEVEL=3```
   * ```VLOG(3)``` , ```VLOG(4)``` ...  will output log

### DIFF
--- a/tensorflow/core/platform/default/logging.cc
+++ b/tensorflow/core/platform/default/logging.cc
@@ -29,8 +29,8 @@ limitations under the License.
 namespace tensorflow {
 namespace internal {
 
-LogMessage::LogMessage(const char* fname, int line, int severity)
-    : fname_(fname), line_(line), severity_(severity) {}
+LogMessage::LogMessage(const char* fname, int line, int severity, int64 vlog_lvl)
+    : fname_(fname), line_(line), severity_(severity), vlog_lvl_(vlog_lvl) {}
 
 #if defined(PLATFORM_POSIX_ANDROID)
 void LogMessage::GenerateLogMessage() {
@@ -86,8 +86,16 @@ void LogMessage::GenerateLogMessage() {
 	   localtime(&now_seconds));
 
   // TODO(jeff,sanjay): Replace this with something that logs through the env.
-  fprintf(stderr, "%s.%06d: %c %s:%d] %s\n", time_buffer, micros_remainder,
-	  "IWEF"[severity_], fname_, line_, str().c_str());
+  if (vlog_lvl_ != 0) {
+    // output VLOG
+    fprintf(stderr, "%s.%06d: %c V%lld %s:%d] %s\n", time_buffer,
+            micros_remainder, "IWEF"[severity_], vlog_lvl_, fname_, line_,
+            str().c_str());
+  } else {
+    // output LOG
+    fprintf(stderr, "%s.%06d: %c %s:%d] %s\n", time_buffer, micros_remainder,
+	    "IWEF"[severity_], fname_, line_, str().c_str());
+  }
 }
 #endif
 
--- a/tensorflow/core/platform/default/logging.h
+++ b/tensorflow/core/platform/default/logging.h
@@ -38,7 +38,7 @@ namespace internal {
 
 class LogMessage : public std::basic_ostringstream<char> {
  public:
-  LogMessage(const char* fname, int line, int severity);
+  LogMessage(const char* fname, int line, int severity, int64 vlog_lvl=0);
   ~LogMessage();
 
   // Returns the minimum log level for VLOG statements.
@@ -53,6 +53,7 @@ class LogMessage : public std::basic_ostringstream<char> {
   const char* fname_;
   int line_;
   int severity_;
+  int64 vlog_lvl_;
 };
 
 // LogMessageFatal ensures the process will exit in failure after
@@ -88,7 +89,7 @@ class LogMessageFatal : public LogMessage {
 
 #define VLOG(lvl)                        \
   if (TF_PREDICT_FALSE(VLOG_IS_ON(lvl))) \
-  ::tensorflow::internal::LogMessage(__FILE__, __LINE__, tensorflow::INFO)
+  ::tensorflow::internal::LogMessage(__FILE__, __LINE__, tensorflow::INFO, lvl)
 
 // CHECK dies with a fatal error if condition is not true.  It is *not*
 // controlled by NDEBUG, so the check will be executed regardless of


martinwicke:
I'm not sure I understand the idea here -- you can now limit logging to see only, say, warning, errors, and fatals. With this new variable, you'd be able to restrict it to only show warnings, but not errors or fatals? 

What's the use case for this?

nutti:
VLOG is always output as INFO log level, so I think VLOG's level (1, 2 ...) is different from LOG's level (info, warning, errors, fatal).
And there is highest level VLOG that is not important compared to the lower level.
Therefore, it is useful to limit output log level from VLOG.

martinwicke:
I still don't understand why one would want both a minimum and a maximum.

nutti:
I want to filter the output log with this improvement.

but Instead, i think that it is also good to display a log level on VLOG output.

martinwicke:
I'm sorry, I must be missing something really obvious -- there already is a minimum vlog level, this PR only adds a maximum. I don't understand why that is useful.

nutti:
Yes, I know there is a minimum VLOG level already.
But in the current interface,  I can not recognize what VLOG level is.
In order to recognize or filter the VLOG level, I added TF_CPP_MAX_VLOG_LEVEL which is the one of way to realize it.

drpngx:
@nutti I think the intent of vlog is to output anything below that level. Having the ability to suppress levels at lower verbosity is probably not something we want, unless I am misunderstanding something.

nutti:
@drpngx 

Yes, you are right. And I understand about the intention.
But the current VLOG does not output VLOG level, so I could not recognize what VLOG level is output.
So, I proposed to suppress the lower log level to recognize it. Of course, this is the best way to solve this problem.

Instead, if you approve to add VLOG level to the log message, it can also solve the problem too.

for example:

code:
```
VLOG(2) << "Debug";
```

VLOG output(V2 will be added):
2017-12-27 21:26:43.712063: I **V2** tensorflow/core/common_runtime/executor.cc:1557] Debug

But I did not which method is better.

martinwicke:
The code generating the actual warning message is in tensorflow/core/platform/default/logging.cc:78. Currently all VLOG messages are generated with INFO severity. Since the VLOG numbers don't map cleanly to INFO/WARNING etc., you'd have to add another field to LogMessage, set that field from the constructor used by VLOG, and then output what it contains. Adding this information would make more sense.

nutti:
Thanks for your instruction.

I fixed my code to output VLOG level as shown previous comment.
Please check this revision.

drpngx:
VLOG means something else, it's for `VLOG(x)` macros. In any case, I don't think we want both min and max. The max should include everything below it.

nutti:
Yes, I understood your intention.

So, I discarded previous commit which adds TF_CPP_MAX_VLOG_LEVEL.
Instead, I added VLOG level to VLOG messages as mentioned before.

This will enable to distinguish which VLOG level is.
And I also think it will be better way than adding max level.

tensorflowbutler:

drpngx:
Thanks! While this seems OK, I'd rather we didn't change the log format to be different from what we have in glog and our internal versions.

nutti:
OK, thanks again for the feedback.
Instead, how about to add VLOG Level to the tail or the head of messages?
I only want to recognize what VLOG level it is, and it is boring to check VLOG level by using ```TF_CPP_MIN_VLOG_LEVEL``` repeatedly.

Example 1:

```
2017-12-27 21:26:43.712063: I tensorflow/core/common_runtime/executor.cc:1557] Debug (VLOG:2)
```

Example 2:

```
2017-12-27 21:26:43.712063: I tensorflow/core/common_runtime/executor.cc:1557] (VLOG:2) Debug
```

tensorflowbutler:

tensorflowbutler:
Nagging Assignee @martinwicke: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

martinwicke:
@drpngx Do you think a change to the format as proposed (prefix, or suffix to the log message, not touching the [] part) would be acceptable? 

@nutti It would be nice to get information about VLOG level in the message, but log output is often processed by script etc., so it is very easy to break users with any change to it.

drpngx:
I guess `Example 2` will not break parsers. But now it makes it extra non-formalized information. I still don't think it's useful enough to change it for everyone.

nutti:
Thank you for your feedback.

I make sense about the effects of the change of VLOG format.
How about the addition of the output switch parameter instead?

For example, environment variable ```TF_CPP_OUTPUT_VLOG_LEVEL``` is a switch parameter in this case.
If you specify ```TF_CPP_OUTPUT_VLOG_LEVEL=TRUE```, you will get the log with VLOG level.
If you don't specify ```TF_CPP_OUTPUT_VLOG_LEVEL```, VLOG level will not output.
This method will not break the exist users.

Is there any other idea to output VLOG level?
Or perhaps, this idea will not be acceptable?

tensorflowbutler:
Nagging Assignee @martinwicke: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

martinwicke:
Ok, I will close this PR. It is truly regrettable that we have two separate and barely compatible logging mechanisms. I'm not sure whether this may change with absl? 

@yilei will absl-py logging use absl C++ logging, and provide a single mechanism for verbosity?

yilei:
Yes, we will make absl-py logging use absl C++ logging once it's released. [`logging.set_verbosity`](https://github.com/abseil/abseil-py/blob/ea8c4d2ddbf3fba610c4d613260561699b776db8/absl/logging/__init__.py#L259) or the -v/--verbosity flags are used to control verbosity.

nutti:
Thanks. I got it.
I'm looking forward to release new verbosity mechanism.

Issue #8926, opened 2017-04-03
Image Distortions should be able to be applied to batches. [Feature Request?]


jubjamie:
The tf.image distortion functions only accept a single image as an input whereas it would be much more useful to be able to place the distortion within your graph flow so any batches of images that pass through it get distorted too. I believe a current solution is as suggeested by @mrry [here on Stack Overflow](http://stackoverflow.com/questions/38920240/tensorflow-image-operations-for-batches) however it feels like, much like the other ops for images work, being able to take batches would be much more useful.

Is the solution proposed by @mrry the accepted method or should the distortion functions be taking batches?

cancan101:
I made a similar changed to `resize_image_with_crop_or_pad `in https://github.com/tensorflow/tensorflow/pull/7369.

jubjamie:
@cancan101 that would be great, something like that. The map_fn method is looking promising for the mean time (as i'm short on time to go and patch TF myself) but I had some PIL/Pillow tools built already in Python so might look at trying to feed in and out of that (although would much rather prefer an in-graph solution!)

aselle:
@mrry, since you commented on the Stackoverflow, could you comment on this?

mrry:
I think it makes sense to make the image API more consistent (in a similar manner to what @cancan101 did), but I'll defer to anyone who feels more ownership over the image API (e.g. @martinwicke, @shlens?).

martinwicke:
See also #1029, #521. @cancan101 did #7369 fix #521?

bhack:
And https://github.com/tensorflow/tensorflow/issues/8465?

zo7:
Just to chime in, it seems that some image ops (e.g. `tf.image.random_brightness()`, `tf.image.random_contrast()`) allow batches, while others (e.g. `tf.image.random_saturation()`, `tf.image.random_hue()`) do not – it would make sense to either make all the image ops accept batches, or at least note on the documentation that they don't.

(also, does the "awaiting tensorflower" flag mean a TF dev is working on this, or is this open to contributions?)

martinwicke:
Stale tags. Fixed now.

itsmeolivia:
Automatically closing due to lack of recent activity. Since this issue is old at this point, please reopen the issue if it still occurs when tried with the latest version of Tensorflow. Thank you.

bhack:
[...]

PR #14187, opened 2017-11-02
Remove `non-fused` version of `adjust_hue` as GPU kernel is already in place

yongtang:
Was looking into adding batch support for `tf.image.random_hue` (#8926) and noticed that the `non-fused` version of `adjust_hue` was still in place. The `non-fused` is for non-GPU support of `adjust_hue`.

As GPU kernel for `AdjustHue` has already been added in PR #6818, I think it makes sense to remove the non-fused version. Besides, the env `TF_ADJUST_HUE_FUSED` seems not in use anyway.

This fix removed `non-fused` version of `adjust_hue` as GPU kernel is already in place.


martinwicke (approved):

martinwicke:
Jenkins, test this please.

PR #14192, opened 2017-11-02
Dockerfile.gpu: use the runtime cuDNN v6 image

flx42:
The generated Docker image will be approximately 900 MB smaller.

The Dockerfile switched to the devel image a long time ago to
workaround a bug when looking up CUDA libraries. This problem has been
fixed in the meantime.

@gunan let me know if I'm missing something and there is a good reason to keep the devel image as a base.


gunan (approved):

gunan:
I cannot think of a good reason, thank you very much for the help!
Jenkins, test this please.

PR #14201, opened 2017-11-03
Update word2vec_basic.py in generate_batch

liuhonggithub:
File "/word2vec_***_basic.py", line 139, in generate_batch
    buffer[:] = data[:span]
TypeError: sequence index must be integer, not 'slice'

got above messages running in python 2.7

updating line 139 from "buffer[:] = data[:span]" to :
       buffer.extend(data[0:span])
The updating has passed my own test and can works.

thanks.


gunan:
This looks like it is sent against the release branch.
We do not accept PRs against the release branch. Could you send this against the master branch?

PR #14225, opened 2017-11-03
Clean up our libcuda stub when building the GPU Docker container

case540:


case540:
Welp, our builds are broken for 1.4 branch since the builds scripts aren't versioned with the branches.

New stuff was added to the ubilds which didn't exist before. Will have to force-submit this :(

PR #14237, opened 2017-11-04
Added an option to `kernel` to `skip_metropolis_step`.

### LINKED ISSUE #14221: bayeslfow.hmc - provide option for skipping the MH step
Currently `tf.contrib.bayesflow.hmc.kernel` returns directly the update x and value of the potential after the Metropolis-Hasting steps. It would be useful to have an option to omitting the MH step. This, for instance, is required for implementing HVI [1], where we want to propagate gradients through the HMC step and not reject any samples:

[1] https://arxiv.org/pdf/1410.6460.pdf

botev:
Fix #14221


botev:
I signed it!

tensorflow/contrib/bayesflow/python/ops/hmc_impl.py
```diff
@@ -477,6 +480,11 @@ def log_joint_partial(beta):
     accepted = math_ops.cast(
```
jvdillon: if skip_metropolis_step:
      return new_x, acceptance_probs, -log_potential_1, -grad_1


Then you dont need additional ops.

jvdillon (commented):
Can you update the unit-test to ensure the correct result is returned when `skip_metropolis_step=True`?

Also, do we want to return acceptance probs as all ones? (Either way, the `skip_metropolis_step` arg docstring should reflect what's returned for acceptance_probs since its ambiguous.)

tensorflow/contrib/bayesflow/python/ops/hmc_impl.py
botev: Should I make that with control flow(allowing for graph conditional) or leave it to be plain boolean?

botev:
@jvdillon Could you point me out where are the tests for this?

jvdillon:
HMC tests live in:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/bayesflow/python/kernel_tests/hmc_test.py

tensorflow/contrib/bayesflow/python/ops/hmc_impl.py
jvdillon: Plain boolean is ok for this. I doubt users will need to do this in a deferred way. And if they do, we can consider adding this feature at that time.

botev:
@jvdillon Do I have to do something for the Test suit to execute?

martinwicke:
Sorry, could you resolve the conflicts? The test ones is easy, but I don't want to mess with the _impl file.

jvdillon (approved):

botev:
@martinwicke I've resolved the conflicts.

martinwicke:
Jenkins, test this please.

botev:
I can't say that I understand what the failure on the CPU Test Python 3 is. Could someone take a look and elaborate on it?

martinwicke:
Infra failure. Nothing for you to worry about.

martinwicke:
The Contrib test does look like it could be related though: hmc_test failing?

botev:
I agree there is some chance, but all other tests pass... also I don't see any more comprehensive report on which part of the test has failed. Could you rerun it?

yifeif:
Here is the test log for the failed test https://source.cloud.google.com/results/invocations/d109530b-8561-434d-8315-6f4639cf70de/targets/%2F%2Ftensorflow%2Fcontrib%2Fbayesflow:hmc_test/log. @botev could you take a look?

martinwicke:
I think the link should actually be https://source.cloud.google.com/results/invocations/d109530b-8561-434d-8315-6f4639cf70de/targets/%2F%2Ftensorflow%2Fcontrib%2Fbayesflow:hmc_test?page=log

/log doesn't work for me, maybe it relies on a referrer.

jvdillon:
Im seeing "test log not found".

yifeif:
Could you try click on one of the blue `Details` link, and then clicked the failing target bayesflow:hmc_test, then TARGET LOG tab?

jvdillon:
(Looks like the accept_probs are quite different from the skip_accept_probs, although Im not sure why.)

botev:
Ah you are right, since the `hmc.kernel` samples the initial velocity vectors inside it means it is stochastic on each call. Does any of you know of any way to synchronize somehow the samples of `random_ops.random_normal(x_shape, dtype=x.dtype)` in two consecutive calls?

botev:
@jvdillon I've used `set_random_seed` in order to go around this. Does that seem reasonable? 
Also, need to see the results of the tests.

jvdillon:
The current impl has a design flaw; every function should take a seed so that way seed can be propagated all the way down to `kernel`. Unfortunately you'll have to make this change to get your cl in.

(Side-note: we're actually revamping HMC and our Metropolis Hastings code based on user feedback. Having said that, Id like to get you the feature you need ASAP and so we're still supporting this version.)

botev:
@jvdillon Could you run jenkins? I think my tests should now pass.

On a related note for improvements, you might want to have a distribution argument for the velocities rather than to always use a fixed Gaussian. 
Even, further following some of the new advances, like https://openreview.net/forum?id=B1n8LexRZ, you can add transformations in as well.

jvdillon:
@martinwicke: can you rerun tests for @botev.

@botev: thank you for this feedback! ill make sure to accommodate this in the new design. (Ive shared the new design with you as well.)

yifeif:

botev:
Ah, apparently this does not work, meaning I need to set the op seed, not the graph one.... should I just drop this test? Currently, the test is impossible to do due to:
```
m = random_ops.random_normal(x_shape, dtype=x.dtype)
```
in `kernel`.

drpngx:
@jvdillon WDYT?

ebrevdo:
Josh is on leave for a while; so this may stall for now.  @langmore can you recommend someone to finish review?

langmore:
I can finish it next week.  I'm out until then.

tensorflow/contrib/bayesflow/python/kernel_tests/hmc_test.py
```diff
@@ -348,6 +348,82 @@ def testAIS2(self):
...
+    feed_dict[step_size] = 0.4
+
+    sample, acceptance_probs, _, _ = hmc.kernel(step_size, 5, initial_draws,
+                                                log_gamma_log_prob, event_dims)
+    skip_sample, skip_acceptance_probs, _, _ = hmc.kernel(
+      step_size, 5, initial_draws, log_gamma_log_prob, event_dims,
+      skip_metropolis_step=True)
+
+    # Need to set the same seed to make the velocity samples same
+    seed = np.random.randint(0, np.iinfo(np.uint32).max)
```
langmore: Rather than making `seed` random, just choose some fixed integer.  That will prevent flakes in the rare event when you get an unlucky seed.

tensorflow/contrib/bayesflow/python/ops/hmc_impl.py
```diff
@@ -308,7 +308,7 @@ def log_prob_beta(x):
 
 
 def kernel(step_size, n_leapfrog_steps, x, target_log_prob_fn, event_dims=(),
```
langmore: To make your test pass, you'll have to do a few things.  The end result will be simpler.

First, add the import and a separate alias
```import tensorflow.python.ops.distributions import util as distribution_util```

Second, to `kernel()` and every function called by `kernel()`, you'll have to add a kwarg `seed` (make it the last kwarg), with default value `None`.

Third, every time a random number is generated, you need to add a "salted" version of `seed`.  E.g.
```m = random_ops.random_normal(x_shape, dtype=x.dtype, seed=distribution_util.gen_new_seed(seed, salt="m")```
Make sure the salt is unique for every call.  Also make sure that if one function calls another, a salted version of the seed is passed on.

Finally, inside your test, you can use the same seed for both calls to `kernel()`.  This should make things work.

Have fun!

botev: I'm not sure that's my definition of fun :D

rmlarsen:
@botev can you address @langmore's comments?

langmore:
Actually, internally we are getting closer to a new version since Josh came back earlier than expected.  @jvdillon , would you say that @botev is better off not bothering with these changes in this version?

botev:
I'm a bit busy until next week, so won't do it until then. If by any chance indeed there is no point bothering let me know.

langmore:
Yes @botev , Josh has confirmed that the new version will be in soon enough that you need not worry.  I'll try to update this thread when the new version is in.

botev:
One thing when you design that is to mention currently there is an inconsistency between `leapfrog_step` and `kernel`. Specifically one accepts the log_density and its gradient, while the other accepts the "potential" which is the negative log_density. I think it would be better both to accept the same thing.

botev:
@jvdillon  Given that this issue #5802 is still open, is there any way currently to run anything in the bayesflow module with:

1. The log-joint has more than one variable (currently it seems that `leapfrog_step` assumes that the inputs cannot be a list)
2. Don't have access to a function mapping x -> log p(x), but rather we have x and the graph node representing log p(x)? E.g. are there any workarounds for constructing a function from the graph? If so where can I find examples?

Atm given that I just found out about these it's making my life hell, as we want to test HMC on a small network...

martinwicke:
@jvdillon @langmore what should happen here?

tensorflowbutler:
Nagging Assignee @ebrevdo: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

jvdillon:
We're just about wrapping up a major refactoring of HMC (and MCMC) code.  The new version will have an UncalibratedHMC (which has no MetropolisHastings adjustment) as well as what is currently HMC.

Also, we're moving MCMC code to: github.com/tensorflow/probability

PR #14246, opened 2017-11-04
typo fixed in CONTRIBUTING.md

iossefy:
The noun phrase [contribution - decision - code] seems to be missing a determiner before it. Consider adding an article.


martinwicke:
The indeterminate "code", as in, "some of the code", is ok here.

Issue #13641, opened 2017-10-11
Extend SVD gradient to support backpropagating through complex and (strongly) rectangular U and V


rmlarsen:
This initial version of SVD gradients has the following restrictions:
  Only supports statically known inner matrix dimensions m and n.

Backpropagating through U and V (i.e. backpropagating through SVD nodes with compute_uv=True) has further restrictions:
  a) Only supports real tensors.
  b) Only supports square and "almost square" matrices where the number of rows and columns differ by at most 1.
  c) full_matrices must be true also. This does not currently have severe implications, given the restriction in b).

Support for dynamic shapes and a) (I think) is straightforward to fix.  But b) is probably a deeper issue having to do with the (lack of) uniqueness of the decomposition, and will require some analysis. I think that if we understand b), we can get around the restriction in c) as well.

I'm marking this as contributions welcome, in the hope that somebody with better math skills than myself will help out :-)

rmlarsen:
It looks like the implementation on autograd is better in several respects. That is probably a good starting point for working on this: https://github.com/HIPS/autograd/blob/master/autograd/numpy/linalg.py#L120

guivenca:
This article studies the analytical computation of the Jacobian of the SVD decomposition, maybe it can be of use:
"Estimating the Jacobian of the Singular Value Decomposition : Theory and Applications " by Theodore Papadopoulo and Manolis L.A. Lourakis
https://hal.inria.fr/inria-00072686/document
Section 2.3.1 analyzes the degenerate case with repeated singular values. Their suggestion is to use a least square approach to obtain the "minimum norm" Jacobian.

That's probably why the finite difference gradient doesn't match the analytical one: even if it's able to compute a gradient, it may be only one of many.

rmlarsen:
@guivenca thanks for the reference! I was thinking that such a minimal norm solution might be the right approach, but didn't work out the math. I'll study this paper.

davharris:
@rmlarsen I just wanted to say that I really appreciate your work on this!

vnavkal:
Would a "minimum norm" Jacobian actually be the one we need?  It seems to me that we need the Jacobian of the specific SVD function that's used.  I suspect this is an obstacle not only to removing the "almost square" restriction but also to extending support to complex tensors, since in the complex case there are always infinitely many choices of U and V.

[...]

PR #14259, opened 2017-11-05
Implement backpropagation for SVD when full_matrices is False

vnavkal:
This pull request implements backpropagation for the singular value decomposition operation when `full_matrices` is False.  A request for this feature is in #13641.

The algorithm I use is the same one as in the [autograd implementation](https://github.com/HIPS/autograd/blob/77387abc6d5d57a2688d6fa04e932f4a9d5613d9/autograd/numpy/linalg.py#L127-L217).  There is a very nice derivation of the formula, written (I think) by @j-towns, at https://j-towns.github.io/papers/svd-derivative.pdf.  (@j-towns I hope you don't mind that I include this link here; I found it in the description of your autograd pull request that added this feature.)

Thanks in advance for taking the time to review this!

### DIFF
--- a/tensorflow/python/ops/linalg_grad.py
+++ b/tensorflow/python/ops/linalg_grad.py
@@ -268,13 +268,14 @@ def _SelfAdjointEigV2Grad(op, grad_e, grad_v):
 
 @ops.RegisterGradient("Svd")
 def _SvdGrad(op, grad_s, grad_u, grad_v):
-  """Gradient for Svd based on Giles' algorithm. Reference at top of file."""
+  """Gradient for the singular value decomposition
 
-  if op.get_attr("compute_uv") and not op.get_attr("full_matrices"):
-    raise NotImplementedError(
-        "SVD gradient is not implemented for compute_uv=True and "
-        "full_matrices=False.")
-
+  The derivation for the compute_uv=False case, and most of
+  the derivation for the full_matrices=True case, are in
+  Giles' paper (see reference at top of file).  A derivation for
+  the full_matrices=False case is available at
+  https://j-towns.github.io/papers/svd-derivative.pdf
+  """
   a = op.inputs[0]
   a_shape = a.get_shape().with_rank_at_least(2)
 
@@ -300,7 +301,7 @@ def _SvdGrad(op, grad_s, grad_u, grad_v):
         "SVD gradient has not been implemented for input with unknown "
         "inner matrix shape.")
 
-  if not op.get_attr("full_matrices") or not op.get_attr("compute_uv"):
+  if not op.get_attr("compute_uv"):
     s, u, v = linalg_ops.svd(a, compute_uv=True, full_matrices=True)
   else:
     s = op.outputs[0]
@@ -334,9 +335,10 @@ def _SvdGrad(op, grad_s, grad_u, grad_v):
     # multiple singular values with value zero. I am not sure if this is a true
     # instability or if it simply throws off the finite difference gradient
     # checker.
-    if abs(m - n) > 1:
+    if op.get_attr("full_matrices") and abs(m - n) > 1:
       raise NotImplementedError(
-          "svd gradient is not implemented for abs(m - n) > 1")
+          "svd gradient is not implemented for abs(m - n) > 1 "
+          "when full_matrices is True")
     s_mat = array_ops.matrix_diag(s)
     s2 = math_ops.square(s)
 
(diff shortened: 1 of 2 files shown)


tensorflow/python/ops/linalg_grad.py
```diff
@@ -334,9 +335,10 @@ def _SvdGrad(op, grad_s, grad_u, grad_v):
     # multiple singular values with value zero. I am not sure if this is a true
     # instability or if it simply throws off the finite difference gradient
     # checker.
```
vnavkal: How should I update this comment?  I personally feel there isn't much hope for the case `abs(m-n) > 1` when `full_matrices=True`, because the gradient would need to be consistent with the particular choices of U and V made by the SVD algorithm.  So my preference is to remove the comment.  But if others prefer for me to leave it here, I can instead change `numerically stable for abs(m-n) > 1` to `numerically stable for abs(m-n) > 1 and full_matrices=True`.

j-towns:
>@j-towns I hope you don't mind that I include this link here; I found it in the description of your autograd pull request that added this feature.

No problem. Glad you found it useful!

rmlarsen:
Will review this on the next couple of days. Thanks for adding this!

tensorflow/python/ops/linalg_grad.py
```diff
@@ -352,32 +354,45 @@ def _SvdGrad(op, grad_s, grad_u, grad_v):
...
+
     u_gu = math_ops.matmul(u, grad_u, adjoint_a=True)
-    v_gv = math_ops.matmul(v, grad_v, adjoint_a=True)
+    v_gv = math_ops.matmul(v1, grad_v1, adjoint_a=True)
 
-    if m == n:
-      f_u = f * u_gu
-      f_v = f * v_gv
-    else:
-      dv2 = array_ops.matrix_transpose(v_gv[..., m:n, :m]) - v_gv[..., :m, m:n]
```
vnavkal: I should explain what happened to `dv2`.  `dv2` has two terms:
1. `array_ops.matrix_transpose(v_gv[..., m:n, :m])` and
2. `-v_gv[..., :m, m:n]`.

When `grad_a_nouv` gets right-multiplied by the adjoint of `v`, the submatrix `dv2` gets right-multiplied by the adjoint of `v[..., :, m:n]`.  After right-multiplication by `v[..., :, m:n]`,
1. the first term of `dv2` becomes the first term of `term2_nous`: `math_ops.matmul(grad_v1, proj_v1_perp, adjoint_a=True)`
2. the second term of `dv2` becomes the second term of `term2_nous`: `-math_ops.matmul(v1t_gv2, v2, adjoint_b=True)`

To see why the first terms match, use the identity
```
math_ops.matmul(v[..., :, :m], v[..., :, :m], adjoint_b=True)
   + math_ops.matmul(v[..., :, m:n], v[..., :, m:n], adjoint_b=True)
       == linalg_ops.eye(n, dtype=v.dtype)
```
which follows from the fact that `v` is unitary.

My motivation for separating the terms of `dv2` was to make clear what the contribution to `grad_a` is from the "redundant" columns of V when `full_matrices=True`.

If the math here isn't clear, just let me know and I can write it down and upload it.

vnavkal:
sure, no hurry, thanks!

martinwicke:
ping @rmlarsen

tensorflow/python/ops/linalg_grad.py
rmlarsen: I'm OK with removing the comment.

rmlarsen:

tensorflow/python/ops/linalg_grad.py
rmlarsen: I'm OK with removing the comment.

tensorflow/python/ops/linalg_grad.py
```diff
@@ -300,7 +301,7 @@ def _SvdGrad(op, grad_s, grad_u, grad_v):
         "SVD gradient has not been implemented for input with unknown "
         "inner matrix shape.")
 
-  if not op.get_attr("full_matrices") or not op.get_attr("compute_uv"):
+  if not op.get_attr("compute_uv"):
     s, u, v = linalg_ops.svd(a, compute_uv=True, full_matrices=True)
```
rmlarsen: Remove full_matrices=True here?

tensorflow/python/ops/linalg_grad.py
```diff
@@ -352,32 +354,45 @@ def _SvdGrad(op, grad_s, grad_u, grad_v):
...
 
-    if m != n:
-      grad_a_nouv = array_ops.concat(
-          [grad_a_nouv, math_ops.matmul(s_inv_mat, dv2)], -1)
+    term1 = math_ops.matmul(u, math_ops.matmul(term1_nouv, v1, adjoint_b=True))
+
+    if m == n:
+      grad_a_before_transpose = term1
+    else:
+      proj_v1_perp = (linalg_ops.eye(n, dtype=v.dtype)
```
rmlarsen: proj_v1_perp is large potentially very large. I think it would be more efficient to compute term2_nous as

grad_v1 - math_ops.matmul(math_ops.matmul(grad_v1, v1), v1, adjoint_b=True)

rmlarsen (changes requested):

tensorflow/python/ops/linalg_grad.py
vnavkal: good point.  I updated this part of the calculation so that it doesn't use an `n` by `n` matrix.  (Note that in the formula you wrote, `grad_v1` needs to be replaced by its adjoint.)

tensorflow/python/ops/linalg_grad.py
vnavkal: I kept it this way to preserve existing behavior; my pull request changes only the logic for the case when `compute_uv` is `True`.  However, I think you're right that it could be removed, and `v[..., :, :m]` could be replaced with `v` [here](https://github.com/tensorflow/tensorflow/blob/17e725c0558581cba19bd6c409698b2c3f88efe5/tensorflow/python/ops/linalg_grad.py#L324) and [here](https://github.com/tensorflow/tensorflow/blob/17e725c0558581cba19bd6c409698b2c3f88efe5/tensorflow/python/ops/linalg_grad.py#L328).

Would it make sense for me to make this change in a separate pull request, since it's not directly related to this one?  I can also just add it to this one if you prefer.

rmlarsen (approved):

rmlarsen:

rmlarsen:

helwilliams:
Hello I am trying to use SVD single components in my training process, however, occasionally the training I get the following error maybe due to back propagation of SVD being unstable, I am using compute_uv=False as I only care about s for training   
```
 InvalidArgumentError (see above for traceback): slice index 0 of dimension 0 out of bounds.
         [[node worker_0/strided_slice (defined at g:\NiftyNet\niftynet\layer\loss_added.py:61)  = StridedSlice[Index=DT_INT32, T=DT_FLOAT, begin_mask=0, ellipsis_mask=0, end_mask=0, new_axis_mask=0, shrink_axis_mask=1, _device="/job:localhost/replica:0/task:0/device:GPU:0"](worker_0/Svd, ConstantFolding/worker_0/gradients/worker_0/NoNewNet/R4/120_bn_relu/bn_/batch_norm/add_grad/BroadcastGradientArgs-folded-1, worker_0/strided_slice/stack_1, worker_0/strided_slice/stack_1)]]

```
any advice on how to bypass this problem would be great, thanks and thanks for your work on the tool.

PR #14286, opened 2017-11-06
fix InvalidArgumentError when using cv2 with tf.py_func()

EncodeTS:
If we use `cv2.imread(filename, cv2.IMREAD_GRAYSCALE)` here, we'll get an `InvalidArgumentError: TypeError: bad argument type for built-in operation`, using `cv2.imread(filename.decode(), cv2.IMREAD_GRAYSCALE)` instead solves this issue.


jhseu (approved):

jhseu:
Jenkins, test this please

EncodeTS:
I notice that the PR still hasn't been merged, what else should I do?

drpngx:
No, good to go.

PR #14287, opened 2017-11-06
Add layer scope to tf.contrib.layers.spatial_softmax

lospooky:
`tensorflow.contrib.layers.spatial_softmax` is lacking a layer-level scope

| before | after |
|---|---|
|before|after|

### DIFF
--- a/tensorflow/contrib/layers/python/layers/layers.py
+++ b/tensorflow/contrib/layers/python/layers/layers.py
@@ -2648,51 +2648,54 @@ def spatial_softmax(features,
     ValueError: If unexpected data_format specified.
     ValueError: If num_channels dimension is unspecified.
   """
-  shape = array_ops.shape(features)
-  static_shape = features.shape
-  if data_format == DATA_FORMAT_NHWC:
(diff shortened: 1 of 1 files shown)


martinwicke:
This may break checkpoints by changing variable names (unless I misunderstand). You could use a name_scope instead, that is harmless.

lospooky:
On one hand, yes, that would break checkpoints. On the other, wouldn't it be a bit against the general behavior of the `layers` api to create variables not in their own scope but in the global one?

No problem though if you'd prefer a `name_scope` however.

martinwicke:
That is true, variable_scope would have been the right thing to do. I am wary of breaking people though, so I'd rather stick with the safer fix. 

@sguada, do you agree with name_scope here? Or should we try for variable_scope.

sguada:
Since it creates a variable it should have a variable_scope.

martinwicke:
Are you ok with breaking checkpoints using this? That's what a variable_scope would do. If you're ok with it, I am too.

sguada (approved):
I'm ok with this change.

martinwicke:
Jenkins, test this please.

Issue #13466, opened 2017-10-03
contrib/verbs only works on GPU


dariavel:
### System information
- **OS Linux readhat** 7.3
- **TensorFlow** installed from source
- **TensorFlow version** 1.3.0
- **Python version**:  2.7.5
- **Bazel version**:
Build label: 0.5.4
Build target: bazel-out/local-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Aug 25 10:00:00 2017 (1503655200)
Build timestamp: 1503655200
Build timestamp as int: 1503655200

- **CUDA/cuDNN version**: no CUDA
- **GPU model and memory**: no GPU
- **Exact command to reproduce**: Compilation with the following command fails:

 bazel build --config=opt //tensorflow/tools/pip_package:build_pip_package

ERROR: /root/tensorflow/tensorflow/contrib/verbs/BUILD:133:1: C++ compilation of rule '//tensorflow/contrib/verbs:rdma' failed (Exit 1).
In file included from ./tensorflow/core/platform/stream_executor.h:26:0,
                 from ./tensorflow/core/common_runtime/gpu/gpu_util.h:23,
                 from tensorflow/contrib/verbs/rdma.cc:24:
./tensorflow/stream_executor/dso_loader.h:32:30: fatal error: cuda/cuda_config.h: No such file or directory
 #include "cuda/cuda_config.h"
                              ^
compilation terminated.

The ./configure is done with all the defaults except enabling verbs.

Reverting the following commit fixes compilation:
commit 0d864630161d9f3b9eaef0b7c6ce7443654df97a
Author: A. Unique TensorFlower <gardener@tensorflow.org>
Date:   Fri Sep 22 13:48:33 2017 -0700

    Move GPU-specific dependencies of core/grappler:devices into
    cuda_deps.

    Fix #includes and deps of contrib/verbs:verbs_util, in particular
    removing an unnecessary #include of gpu_util.h that relied on a
    transitive dependency through :devices.

    PiperOrigin-RevId: 169732234

@junshi15

junshi15:
The existing verbs implementation assumes GPU usage. For example the following line calls GPUUtil.

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/verbs/rdma.cc#L714

You can make it work with CPU-only cases, with some code change, I suppose.

tatatodd:
[...]

PR #14290, opened 2017-11-06
Adding connectivity check, compilation fix and some code refactoring to verbs 

dariavel:
1. Connectivity check - checking the correctness of the RDMA configuration parameters by pinging on each channel.
2. Compilation with verbs and without CUDA fix (contrib/verbs only works on GPU #13466)
3. Code refactoring: 
  Call done in case of not OK status fix  
  Replace hardcoded 100 with RDMA_QP_QUEUE_DEPTH

### DIFF
--- a/tensorflow/contrib/verbs/rdma.h
+++ b/tensorflow/contrib/verbs/rdma.h
@@ -107,6 +107,7 @@ class RdmaAdapter {
   ~RdmaAdapter();
   // Adapter name, e.g. mlx5_0.
   string name() const;
+  void StartPolling();
   void Process_CQ();
 
  protected:
--- a/tensorflow/contrib/verbs/rdma.h
+++ b/tensorflow/contrib/verbs/rdma.h
@@ -162,6 +162,15 @@ class RdmaChannel {
   void RemoveRecvCallback(const string& key);
   void RunRecvCallback(const string& key);
   static const int kNumMessageBuffers = 4;
+  static const int PingRecvWrid = 0;
+
+ private:
+  static const int PingBuffSize = 1024;
+  char ping_buff_[PingBuffSize];
+  struct ibv_mr* mr_;
+  struct ibv_sge ping_sge_list_;
+  int PingPostRecv();
+  int PingPostSend();
 
  protected:
   const RdmaAdapter* adapter_;
--- a/tensorflow/contrib/verbs/rdma.h
+++ b/tensorflow/contrib/verbs/rdma.h
@@ -162,11 +162,11 @@ class RdmaChannel {
   void RemoveRecvCallback(const string& key);
   void RunRecvCallback(const string& key);
   static const int kNumMessageBuffers = 4;
-  static const int PingRecvWrid = 0;
+  static const int kPingRecvWrid = 0;
 
  private:
-  static const int PingBuffSize = 1024;
-  char ping_buff_[PingBuffSize];
+  static const int kPingBuffSize = 1024;
+  char ping_buff_[kPingBuffSize];
   struct ibv_mr* mr_;
   struct ibv_sge ping_sge_list_;
   int PingPostRecv();
--- a/tensorflow/contrib/verbs/rdma_mgr.cc
+++ b/tensorflow/contrib/verbs/rdma_mgr.cc
@@ -115,7 +115,6 @@ void RdmaMgr::SetupChannels() {
   }
 }
 
-
 // Check connectivity by pinging every channel
 bool RdmaMgr::ConnectivityCheck() {
   int i, rcnt = 0, scnt = 0;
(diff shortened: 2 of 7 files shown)


dariavel:
This PR fixes the #13466 (contrib/verbs only works on GPU) issue.

dariavel:
@junshi can you please review?

jhseu:
@noaezra , can you confirm here that you're ok with your commit being submitted?

tensorflow/contrib/verbs/rdma.h
```diff
@@ -161,6 +162,15 @@ class RdmaChannel {
   void RemoveRecvCallback(const string& key);
   void RunRecvCallback(const string& key);
   static const int kNumMessageBuffers = 4;
+  static const int PingRecvWrid = 0;
+
+ private:
+  static const int PingBuffSize = 1024;
```
poxvoculi: Google style for const names like this is kPingBuffSize.

tensorflow/contrib/verbs/rdma.h
```diff
@@ -161,6 +162,15 @@ class RdmaChannel {
   void RemoveRecvCallback(const string& key);
   void RunRecvCallback(const string& key);
   static const int kNumMessageBuffers = 4;
+  static const int PingRecvWrid = 0;
```
poxvoculi: kPingRecvWrid

tensorflow/contrib/verbs/rdma_rendezvous_mgr.cc
```diff
@@ -58,20 +60,13 @@ void RdmaRemoteRendezvous::RecvFromRemoteAsync(
   // parse src_name and dst_name
   string src_name, dst_name, unused;
   if (!DeviceNameUtils::SplitDeviceName(parsed.src_device, &src_name,
+                                        &unused) ||
+      !DeviceNameUtils::SplitDeviceName(parsed.dst_device, &dst_name,
                                         &unused)) {
     s = errors::Internal("Could not parse src name.");
```
poxvoculi: Looks like the error message is misleading if it's the dst_name that won't parse.

poxvoculi:
I'm okay with this change, but I don't have a well-informed opinion as to how appropriate it is.  @junshi15 should give his opinion.  Is there a background thread other than the cited bug discussing this change?

tensorflow/contrib/verbs/rdma_mgr.cc
```diff
@@ -115,6 +115,57 @@ void RdmaMgr::SetupChannels() {
...
+        RdmaChannel* rc =
+            reinterpret_cast<RdmaChannel*>(rdma_adapter_->wc_[i].wr_id);
+        CHECK(s == IBV_WC_SUCCESS) << ": " << ibv_wc_status_str(
+                                                  rdma_adapter_->wc_[i].status)
+                                   << "(" << rdma_adapter_->wc_[i].status
+                                   << ") to " << rc->remote_name_;
+        ++scnt;
+      }
+    }  // for
+  }    // while
```
junshi15: Can this while loop hang if one of the remote node is down?

junshi15:
@dariavel The patch looks good, except that I have one question above.

noaezra:
@jhseu # "noaezra , can you confirm here that you're ok with your commit being submitted?"
It's ok that my commit is being submitted.

tensorflow/contrib/verbs/rdma_mgr.cc
dariavel: If the remote node is down we will get an error earlier in channel creation or when sending ping. 
The only option for the loop to hang is when the remote node failes between receiving ping to sending reply.
In this case we will successfully send the ping to remote, it will receive it and go down for some reason before sending reply so we'll be polling the completion queue forever, similar to the case when remote node goes down after connectivity check and we are stuck in the while(true) [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/verbs/rdma.cc#L443).

tensorflow/contrib/verbs/rdma_rendezvous_mgr.cc
dariavel: It's scr or dst name. Fixed in [d43d00b](https://github.com/tensorflow/tensorflow/pull/14290/commits/d43d00be13ff271eb8a2e6a14eb7ac01a51934ff).

tensorflow/contrib/verbs/rdma_mgr.cc
junshi15: Make sense. Thanks for explanation.

poxvoculi (approved):

noaezra:
@jhseu @poxvoculi @junshi15
can you please re-run the tests?

tatatodd:
Jenkins, test this please.

shamoya:
Thanks @tatatodd . looks like the failure is not related to the PR.
Can we merge or re-run ?

tatatodd:
Jenkins, test this please.

PR #13242, opened 2017-09-22
Add op `tf.contrib.ffmpeg.decode_video`

yongtang:
This fix tries to address the request raised in #6265 where it was not possible to decode video like the existing op of `decode_audio`.

This fix adds the support of `tf.contrib.ffmpeg.decode_video` by invoking ffmpeg the same fashion as `tf.contrib.ffmpeg.decode_audo` so that video could be stored in the tensor `[frames, height, width, channel]`. At the moment, the output format is `RGB24`.

This fix fixes #6265.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

PR #14296, opened 2017-11-06
Add support for grayscale bmp image

### LINKED ISSUE #13942: Support grayscale .bmp images
The current implementation only supports 3 (RGB) or 4 (ARGB) channel BMP images. However, grayscale images are also often used for machine learning stuff. Therefore it would be great to be able to read 1 channel BMP images.

See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/decode_bmp_op.cc#L71

yongtang:
This fix tries to address the issue raised in #13942 to support grayscale bmp image. Previously only channels of 3 or 4 are supported in bmp decoding. This fix adds the support to have 1 channel grayscale image.

This fix fixes #13942.

### DIFF
--- a/tensorflow/python/kernel_tests/decode_bmp_op_test.py
+++ b/tensorflow/python/kernel_tests/decode_bmp_op_test.py
@@ -64,6 +64,40 @@ class DecodeBmpOpTest(test.TestCase):
       decoded = decode.eval()
       self.assertAllEqual(decoded, img_bytes)
 
+  def testGrayscale(self):
+    img_bytes = [[[255], [0]], [[255], [0]]]
+    encoded_bytes = [
+        0x42, 0x40,
+        0x3d, 0, 0, 0,
+        0, 0,
+        0, 0,
+        0x36, 0, 0, 0,
+        0x28, 0, 0, 0,
+        0x2, 0, 0, 0,
+        0x2, 0, 0, 0,
+        0x1, 0,
+        0x8, 0,
+        0, 0, 0, 0,
+        0x10, 0, 0, 0,
+        0x13, 0xb, 0, 0,
+        0x13, 0xb, 0, 0,
+        0, 0, 0, 0,
+        0, 0, 0, 0,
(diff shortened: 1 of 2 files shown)


tensorflow/python/kernel_tests/decode_bmp_op_test.py
```diff
@@ -64,6 +67,16 @@ def testex1(self):
       decoded = decode.eval()
       self.assertAllEqual(decoded, img_bytes)
 
+  def testGrayscale(self):
+    prefix_path = "tensorflow/core/lib"
+    # Read a grayscale 512*512 bmp and verify shape
+    path = os.path.join(prefix_path, "bmp", "testdata", "lena512.bmp")
```
martinwicke: We'd like to avoid binary files as much as possible. Since this is testing the decoder, not the reader, could you just use code to define test data (as done above).

martinwicke (changes requested):

tensorflow/python/kernel_tests/decode_bmp_op_test.py
yongtang: Thanks @martinwicke. The PR has been updated with binary file `lena512.bmp` removed.

yongtang:
It looks like tests passed as well.
One is the timeout error of 
```
//tensorflow/python/kernel_tests:slice_op_test
```
and `Ubuntu contrib`/`MacOs Contrib` error. It looks like they are very similar to the error in https://github.com/tensorflow/tensorflow/pull/13242#issuecomment-343700709

Maybe they are unrelated as well?

martinwicke (approved):

martinwicke:
Thanks! I'm rerunning the GPU tests here: http://ci.tensorflow.org/job/tensorflow-pull-requests-gpu/7497/

The slide_op_test is indeed unrelated.

martinwicke:
Passed. Thanks!

PR #14297, opened 2017-11-06
Fix batch dataset op test

martinwicke:


caisq (approved):

martinwicke:
I changed my mind. I will disable the test in pip. We can choose to re-enable it, but ideally we would come up with a solution that uses bazel to include all the required deps.

martinwicke:
Jenkins, test this please.

tensorflow/contrib/data/python/kernel_tests/BUILD
```diff
@@ -11,6 +11,7 @@ py_test(
     size = "small",
     srcs = ["batch_dataset_op_test.py"],
     srcs_version = "PY2AND3",
+    tags = ["nopip"],
```
caisq: --> no_pip

PR #14317, opened 2017-11-07
added reuse in kwargs checking for reusing shared variable

kamwoh:
When I am implementing DeepID2 from "Deep Learning Face Representation by Joint
Identification-Verification by Yi Sun, Xiaogang Wang and Xiaoou Tang", I realise that using same architecture for two inputs (two face image) and sharing weight (if I didn't misunderstand) when feedforward, there is a layer called locally connected convolution, and only keras implemented LocallyConnected2D, when I trying to reuse weight, but failed because the layer didn't allow "reuse" as a kwargs, so when I added it, it seems working (I am not expert in deep learning, so I am not sure if I am doing it correct, so I'm always trying any way)

### DIFF
--- a/tensorflow/python/keras/_impl/keras/engine/topology.py
+++ b/tensorflow/python/keras/_impl/keras/engine/topology.py
@@ -134,6 +134,7 @@ class Layer(tf_base_layers.Layer):
         'name',
         'trainable',
         'weights',
+        'reuse'
     }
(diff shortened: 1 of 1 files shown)


kamwoh:
I signed it!

jhseu:
We're not accepting commits to the 1.4 branch. If you still want to do this change, can you make it in the master branch?

PR #14330, opened 2017-11-07
Add `decode_libsvm` for libsvm format support

### LINKED ISSUE #14313: Feature request: add tf.decode_libsvm op
Hi, since libsvm format is used widespreadly to store sparse data, and is supported by many main-stream frameworks:
+ spark: [LibSVMDataSource](https://spark.apache.org/docs/2.0.0/api/java/org/apache/spark/ml/source/libsvm/LibSVMDataSource.html)
+ sklearn: [load_svmlight_file](http://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_svmlight_file.html)
+ xgboost: [LibSVM format](https://github.com/dmlc/xgboost/blob/master/demo/binary_classification/README.md)

Could tensorflow support to parse libsvm format like [`tf.decode_csv`](https://www.tensorflow.org/versions/master/api_docs/python/tf/decode_csv)?

yongtang:
This fix is an effort to add libsvm format support with the implementation of `decode_libsvm`, as was proposed in #14313.

The implementation is done in contrib with, e.g.,
```
label, feature = tf.contrib.libsvm.decode_libsvm(input, num_features=6)
```

where the input is a string tensor and the output is a tuple of label and feature.

The label tensor is the same shape as input. The feature tensor is of the shape `[input_shape, num_features]`.

This fix fixes #14313.

### DIFF
--- a/tensorflow/contrib/libsvm/BUILD
+++ b/tensorflow/contrib/libsvm/BUILD
@@ -1,0 +1,102 @@
+package(
+    default_visibility = ["//visibility:private"],
+)
+
+licenses(["notice"])  # Apache 2.0
+
+exports_files(["LICENSE"])
+
+load("//tensorflow:tensorflow.bzl", "tf_custom_op_library")
+load("//tensorflow:tensorflow.bzl", "tf_gen_op_libs")
+load("//tensorflow:tensorflow.bzl", "tf_gen_op_wrapper_py")
+load("//tensorflow:tensorflow.bzl", "tf_kernel_library")
+load("//tensorflow:tensorflow.bzl", "tf_custom_op_py_library")
+load("//tensorflow:tensorflow.bzl", "tf_py_test")
+
+tf_custom_op_library(
+    name = "python/ops/_libsvm_ops.so",
+    srcs = [
+        "kernels/decode_libsvm_op.cc",
+        "ops/libsvm_ops.cc",
+    ],
+    deps = [
+        "//tensorflow/core/kernels:bounds_check_lib",
+    ],
+)
+
+tf_kernel_library(
+    name = "libsvm_kernels",
+    srcs = ["kernels/decode_libsvm_ops.cc"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//tensorflow/core:framework",
+        "//tensorflow/core:lib",
+        "//tensorflow/core:lib_internal",
+        "//tensorflow/core/kernels:bounds_check_lib",
+    ],
+)
+
+tf_gen_op_libs(
+    op_lib_names = ["libsvm_ops"],
+    deps = [
+        "//tensorflow/core:lib",
+    ],
+)
+
+tf_gen_op_wrapper_py(
+    name = "libsvm_ops",
+    deps = [":libsvm_ops_op_lib"],
+)
+
+tf_custom_op_py_library(
+    name = "libsvm",
+    srcs = [
+        "__init__.py",
+        "python/ops/libsvm_ops.py",
+    ],
+    dso = [
+        ":python/ops/_libsvm_ops.so",
+    ],
+    kernels = [
+        ":libsvm_kernels",
+        ":libsvm_ops_op_lib",
+    ],
+    srcs_version = "PY2AND3",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":libsvm_ops",
+        "//tensorflow/contrib/util:util_py",
+        "//tensorflow/python:array_ops",
+        "//tensorflow/python:control_flow_ops",
+        "//tensorflow/python:framework",
+        "//tensorflow/python:framework_for_generated_wrappers",
+        "//tensorflow/python:platform",
+        "//tensorflow/python:state_ops",
+        "//tensorflow/python:training",
+    ],
+)
+
+tf_py_test(
+    name = "decode_libsvm_op_test",
+    srcs = ["python/kernel_tests/decode_libsvm_op_test.py"],
+    additional_deps = [
+        ":libsvm",
+        "//third_party/py/numpy",
+        "//tensorflow/python:client_testlib",
+        "//tensorflow/python:framework",
+        "//tensorflow/python:framework_test_lib",
+        "//tensorflow/python:platform_test",
+    ],
+)
+
+filegroup(
+    name = "all_files",
+    srcs = glob(
+        ["**/*"],
+        exclude = [
+            "**/METADATA",
+            "**/OWNERS",
+        ],
+    ),
+    visibility = ["//tensorflow:__subpackages__"],
+)
--- a/tensorflow/contrib/libsvm/BUILD
+++ b/tensorflow/contrib/libsvm/BUILD
@@ -26,7 +26,7 @@ tf_custom_op_library(
 
 tf_kernel_library(
     name = "libsvm_kernels",
-    srcs = ["kernels/decode_libsvm_ops.cc"],
+    srcs = ["kernels/decode_libsvm_op.cc"],
     visibility = ["//visibility:public"],
     deps = [
         "//tensorflow/core:framework",
--- a/tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
+++ b/tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
@@ -28,6 +28,9 @@ class DecodeLibsvmOp : public OpKernel {
  public:
   explicit DecodeLibsvmOp(OpKernelConstruction* ctx) : OpKernel(ctx) {
     OP_REQUIRES_OK(ctx, ctx->GetAttr("num_features", &num_features_));
+    OP_REQUIRES(ctx, (num_features_ >= 1),
+                errors::InvalidArgument("Invalid number of features \"",
+                                        num_features_, "\""));
   }
 
   void Compute(OpKernelContext* ctx) override {
--- a/tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
+++ b/tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
@@ -36,16 +36,13 @@ class DecodeLibsvmOp : public OpKernel {
     const auto& input_flat = input_tensor->flat<string>();
 
     Tensor* label_tensor;
-    Tensor* feature_tensor;
     OP_REQUIRES_OK(ctx,
                    ctx->allocate_output(0, TensorShape({input_flat.size()}),
                                         &label_tensor));
-    OP_REQUIRES_OK(ctx, ctx->allocate_output(
-                            1, TensorShape({input_flat.size(), num_features_}),
-                            &feature_tensor));
-
     auto label = label_tensor->flat<int64>();
-    auto feature = feature_tensor->matrix<T>();
+
+    std::vector<T> out_values;
+    std::vector<std::pair<int64, int64>> out_indices;
     for (int i = 0; i < input_flat.size(); ++i) {
       std::vector<string> entries =
           str_util::Split(input_flat(i), " ", str_util::SkipEmpty());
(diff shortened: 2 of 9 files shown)


mrry:
Thanks for contributing this! One suggestion: since libsvm format is primarily used for representing sparse data, would it make sense for this op to return a `tf.SparseTensor` of features?

yongtang:
@mrry Thanks for the feedback. Let me take a look and update the PR.

tensorflow/contrib/libsvm/ops/libsvm_ops.cc
```diff
@@ -0,0 +1,56 @@
...
+REGISTER_OP("DecodeLibsvm")
+    .Input("input: string")
+    .Output("label: int64")
+    .Output("feature: dtype")
+    .Attr("dtype: {float, double, int32, int64} = DT_FLOAT")
+    .Attr("num_features: int >= 1")
+    .SetShapeFn([](InferenceContext* c) {
+      c->set_output(0, c->input(0));
+
+      int32 num_features;
```
facaiy: int64?

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
```diff
@@ -0,0 +1,112 @@
...
+#include "tensorflow/core/lib/strings/numbers.h"
+#include "tensorflow/core/lib/strings/str_util.h"
+
+namespace tensorflow {
+
+template <typename T>
+class DecodeLibsvmOp : public OpKernel {
+ public:
+  explicit DecodeLibsvmOp(OpKernelConstruction* ctx) : OpKernel(ctx) {
+    OP_REQUIRES_OK(ctx, ctx->GetAttr("num_features", &num_features_));
```
facaiy: Perhaps we'd better check num_features > 1?

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
```diff
@@ -0,0 +1,112 @@
...
+
+    auto label = label_tensor->flat<int64>();
+    auto feature = feature_tensor->matrix<T>();
+    for (int i = 0; i < input_flat.size(); ++i) {
+      std::vector<string> entries =
+          str_util::Split(input_flat(i), " ", str_util::SkipEmpty());
+      OP_REQUIRES(ctx, (entries.size() > 0),
+                  errors::InvalidArgument("No entries found for input[", i,
+                                          "]: \"", input_flat(i), "\""));
+      int64 label_value;
```
facaiy: I think int32 is large enough for label.

facaiy:
As said by @mrry, `SparseTensor` for features is really what I need. I am looking forward to the improvement. Thanks for your work, @yongtang.

yongtang:
@mrry @facaiy The PR has been updated by returning sparse tensors. Please take a look and let me know if there are any issues.

tensorflow/contrib/libsvm/BUILD
```diff
@@ -0,0 +1,102 @@
+package(
```
mrry: Adding a new `BUILD` file will require complementary changes in `tensorflow/contrib/cmake/tf_core_kernels.cmake` and `tensorflow/contrib/cmake/tf_python.cmake`.

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
```diff
@@ -0,0 +1,153 @@
...
+  void Compute(OpKernelContext* ctx) override {
+    const Tensor* input_tensor;
+    OP_REQUIRES_OK(ctx, ctx->input("input", &input_tensor));
+    const auto& input_flat = input_tensor->flat<string>();
+
+    Tensor* label_tensor;
+    OP_REQUIRES_OK(ctx,
+                   ctx->allocate_output(0, TensorShape({input_flat.size()}),
+                                        &label_tensor));
+    auto label = label_tensor->flat<int64>();
```
mrry: Nit: since `label_tensor` is always a vector, use `label_tensor->vec<int64>();`.

tensorflow/contrib/libsvm/ops/libsvm_ops.cc
```diff
@@ -0,0 +1,58 @@
...
+REGISTER_OP("DecodeLibsvm")
+    .Input("input: string")
+    .Output("label: int64")
+    .Output("feature_indices: int64")
+    .Output("feature_values: dtype")
+    .Output("feature_shape: int64")
+    .Attr("dtype: {float, double, int32, int64} = DT_FLOAT")
+    .Attr("num_features: int >= 1")
+    .SetShapeFn([](InferenceContext* c) {
+      c->set_output(0, c->input(0));
```
mrry: This says that output 0 has the same shape as input 0, but in the implementation output 0 is always a vector. In that case, add an assertion that `c->input(0)` is a vector as well.

(Alternatively, you could allow `c->input(0)` to be a scalar or a vector, but that would require some changes to the op implementation, and could be added later.)

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
mrry: In the implementation [I've seen](https://ece.uwaterloo.ca/~nnikvand/Coderep/libsvm-3.16/tools/checkdata.py) it appears that labels are treated as floating-point values. This was initially surprising to me, but I suppose it is possible that you'd have a dataset that maps sparse features to continuous values....

(There's also a multi-label form, but that could be handled by a separate op that returns *two* `tf.SparseTensor` objects.)

tensorflow/contrib/libsvm/python/kernel_tests/decode_libsvm_op_test.py
```diff
@@ -0,0 +1,50 @@
...
+from tensorflow.python.platform import test
+
+
+class DecodeLibsvmOpTest(test.TestCase):
+
+  def testBasic(self):
+    with self.test_session() as sess:
+      content = ["1 1:3.4 2:0.5 4:0.231",
+                 "1 2:2.5 3:0.1 5:0.503",
+                 "2 3:2.5 2:0.1 1:0.105"]
```
mrry: Can you add a test with `nan` and `inf` in floating-point data? This seems to be allowed:

https://ece.uwaterloo.ca/~nnikvand/Coderep/libsvm-3.16/tools/checkdata.py

tensorflow/contrib/libsvm/python/ops/libsvm_ops.py
```diff
@@ -0,0 +1,47 @@
...
+
+  Args:
+    content: A `Tensor` of type `string`. Each string is a record/row in
+      the Libsvm format.
+    num_features: The number of features.
+    dtype: The type of the output feature tensor. Default to tf.float32.
+
+  Returns:
+    label: A `Tensor` of the same shape as content.
+    feature: A `Tensor` of the shape `[input_shape, num_features]`.
```
mrry: This line is outdated: instead of returning the separate components, build a `tf.SparseTensor` from them and return that.

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
```diff
@@ -0,0 +1,153 @@
...
+      label(i) = label_value;
+      for (int j = 1; j < entries.size(); j++) {
+        std::vector<string> pair = str_util::Split(entries[j], ":");
+        OP_REQUIRES(
+            ctx, (pair.size() == 2),
+            errors::InvalidArgument("Invalid feature \"", entries[j], "\""));
+        int64 feature_index;
+        OP_REQUIRES(
+            ctx, strings::safe_strto64(pair[0].c_str(), &feature_index),
+            errors::InvalidArgument("Feature format incorrect: ", entries[j]));
```
mrry: Also check that `feature_index >= 0`. Technically they must also be in ascending order, but I suppose we don't need to enforce that here.

mrry (changes requested):
Thanks for the changes. I've looked into the LIBSVM format a bit more, and I there seem to be some other legitimate inputs that this should handle.

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
facaiy: I agree that ascending order is too strict.

tensorflow/contrib/libsvm/python/kernel_tests/decode_libsvm_op_test.py
facaiy: The idea is cool, however, perhaps it is better to keep it simple at first. And if the feature is requested later, we can support it in the future.

tensorflow/contrib/libsvm/BUILD
yongtang: Thanks @mrry. The PR has been updated with cmake file added. I don't have a Windows dev machine setup yet but will try to see if I could fix any issues if encountered. May get a Windows dev machine later.

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
yongtang: Thanks @mrry. Actually my intention was to allow label to match any shape (e.g., `[3, 2]`) from input. However, there was one bug in the previous update that uses `TensorShape({input_flat.size()})`. Now the PR has been updated using `input_tensor->shape()` and the issue should be resolved.

I added an additional test case to show that an input tensor of shape `[3, 2]` works as well. (as opposed to a vector).

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
yongtang: Thanks. I updated the PR so that an additional attribute `label_dtype` could be provided to control the data type of the label.

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
yongtang: Thank. `feature_index >= 0` check has been added.

tensorflow/contrib/libsvm/ops/libsvm_ops.cc
yongtang: Thanks @mrry. The PR has been updated so that the input could be any shape.

tensorflow/contrib/libsvm/python/kernel_tests/decode_libsvm_op_test.py
yongtang: Additional test cases have been added.

tensorflow/contrib/libsvm/python/ops/libsvm_ops.py
yongtang: Thanks @mrry. Updated.

yongtang:
Thanks for the review and feedback. The PR has been updated. Please take a look.

tensorflow/contrib/libsvm/python/ops/libsvm_ops.py
```diff
@@ -0,0 +1,51 @@
...
+    dtype: The type of the output feature tensor. Default to tf.float32.
+    label_dtype: The type of the output label tensor. Default to tf.int64.
+
+  Returns:
+    label: A `Tensor` of the same shape as content.
+    feature: A `SparseTensor` of the shape `[input_shape, num_features]`.
+  """
+  label, indices, values, shape = gen_libsvm_ops.decode_libsvm(
+      content, num_features, dtype=dtype, label_dtype=label_dtype)
+  return label, sparse_tensor.SparseTensor(indices, values, shape)
```
mrry: Many TensorFlow APIs order their arguments and return values as `features, labels` rather than `labels, features` (e.g. `tf.estimator.Estimator`). I realize that the Libsvm format puts the label first, but I think we should be consistent with the other APIs and swap the order here.

mrry (changes requested):
Thanks for making the changes. I just have one last request about the API, and then we should be good to merge.

tensorflow/contrib/libsvm/python/ops/libsvm_ops.py
yongtang: Thanks @mrry. The return values have been updated to `features, labels`.

yongtang:
Thanks @mrry for the review. The PR has been updated. Please take a look and let me know if there are any issues.

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
```diff
@@ -0,0 +1,178 @@
...
+    Tensor* label_tensor;
+    OP_REQUIRES_OK(
+        ctx, ctx->allocate_output(0, input_tensor->shape(), &label_tensor));
+    auto label = label_tensor->flat<Tlabel>();
+
+    std::vector<T> out_values;
+    std::vector<std::pair<int64, int64>> out_indices;
+    for (int i = 0; i < input_flat.size(); ++i) {
+      std::vector<string> entries =
+          str_util::Split(input_flat(i), " ", str_util::SkipEmpty());
```
snnn: Here has a performance problem:  For each mini-batch, how many memory allocations occurred at here?

mrry: +1. It would probably be more efficient to loop over `input_flat(i)` to identify the start and end positions for each label/feature, then use `StringPiece` instead of `string` as the argument to `ConvertHelper<T>()`.

(It would also be fine to iterate on this in a future PR.)

mrry (approved):

mrry:

tensorflow/contrib/libsvm/kernels/decode_libsvm_op.cc
yongtang: Thanks @mrry @snnn, I will have a follow up PR for that after this PR is merged.

yongtang:
The PR has been rebased and pushed to resolve the merge conflict.

yongtang:
All tests passed except `//tensorflow/python/keras:data_utils_test` in ubuntu. The error code is `timeout`. I assume this is unrelated?

mrry:
This looks good to merge. I'll kick off the tests one more time to try and get a clean bill of health.

yifeif:

yongtang:
Thanks all for the help! 👍 🎉

mrry:
Woohoo! Glad to see this PR land!

facaiy:
Many thanks, @yongtang !

PR #12769, opened 2017-09-02
Allow tfcompile_flags to be a list

yongtang:
This fix tries to fix the issue raised in #12767 where
it was not possible to specify tfcompile_flags as a list:
```
tf_library(
  ...
  tfcompile_flags = ["--target_cpu='core-avx2'", "--xla_enable_fast_math=false"]
)
```
will crash upon build with '+' operator applied to incompatible types (select of string, list)

This is inconsistent with other rules like 'copts' in cc_binary.

The issue is from tfcompile.bzl:
```
" " + (tfcompile_flags or "")),
```

This fix uses `" ".join(tfcompile_flags or [])` instead so that it
is possible to specify the list.

This fix fixes #12767.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

PR #14333, opened 2017-11-07
Revert "Allow tfcompile_flags to be a list"

martinwicke:
Reverts tensorflow/tensorflow#12769, which breaks everything.


martinwicke:
@yongtang I have to revert this. I think the test invocation was old. Please reopen? I believe the change is still a good one.

yongtang:
Thanks @martinwicke. The PR has been created in #14337. Please take a look.

PR #14341, opened 2017-11-08
Add checkpoint file prefix check

### LINKED ISSUE #9465: Misleading error message when running restore op with path to an .index file
Using V2 checkpoint files, calling `saver.restore()` with path to the `.index` returns the following error: 

    Error loading checkpoint from /home/peci1/tradr-git/tradr-ws/src/tradr-simulation/safe_exploration/scripts/ddpg/models/my-model.index: Not found: Tensor name "ActorFullyConnected1/b" not found in checkpoint files /home/peci1/tradr-git/tradr-ws/src/tradr-simulation/safe_exploration/scripts/ddpg/models/my-model.index

I think it'd be nicer to get an error saying "File ... is not a checkpoint. Please, remove the .index extension". This is roughly what you get if you pass any other file.

tedhtchang:
V2 format refers checkpoint files by their file name prefix whereas V1 format refers checkpoint by actual file. I added a check to verify the V2 checkpoint reference is not a file. Also, the return value for saver.save() has changed for V2 so I corrected the doc string.

### DIFF
--- a/tensorflow/contrib/slim/python/slim/evaluation_test.py
+++ b/tensorflow/contrib/slim/python/slim/evaluation_test.py
@@ -236,7 +236,7 @@ class SingleEvaluationTest(test.TestCase):
   def _prepareCheckpoint(self, checkpoint_path):
     init_op = control_flow_ops.group(variables.global_variables_initializer(),
                                      variables.local_variables_initializer())
-    saver = saver_lib.Saver(write_version=saver_pb2.SaverDef.V1)
+    saver = saver_lib.Saver()
     with self.test_session() as sess:
       sess.run(init_op)
       saver.save(sess, checkpoint_path)
--- a/tensorflow/python/tools/freeze_graph.py
+++ b/tensorflow/python/tools/freeze_graph.py
@@ -72,7 +72,8 @@ def freeze_graph_with_def_protos(input_graph_def,
                                  variable_names_blacklist="",
                                  input_meta_graph_def=None,
                                  input_saved_model_dir=None,
-                                 saved_model_tags=None):
+                                 saved_model_tags=None,
+                                 checkpoint_version=saver_pb2.SaverDef.V2):
   """Converts all variables in a graph and checkpoint into constants."""
   del restore_op_name, filename_tensor_name  # Unused by updated loading code.
 
@@ -100,7 +101,8 @@ def freeze_graph_with_def_protos(input_graph_def,
     _ = importer.import_graph_def(input_graph_def, name="")
   with session.Session() as sess:
     if input_saver_def:
-      saver = saver_lib.Saver(saver_def=input_saver_def)
+      saver = saver_lib.Saver(saver_def=input_saver_def,
+                              write_version=checkpoint_version)
       saver.restore(sess, input_checkpoint)
     elif input_meta_graph_def:
       restorer = saver_lib.import_meta_graph(
@@ -124,7 +126,8 @@ def freeze_graph_with_def_protos(input_graph_def,
           # 'global_step' or a similar housekeeping element) so skip it.
           continue
         var_list[key] = tensor
-      saver = saver_lib.Saver(var_list=var_list)
+      saver = saver_lib.Saver(var_list=var_list,
+                              write_version=checkpoint_version)
       saver.restore(sess, input_checkpoint)
       if initializer_nodes:
         sess.run(initializer_nodes.split(","))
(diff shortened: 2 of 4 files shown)


martinwicke (approved):

drpngx:
Apparently that breaks existing behavior.

@tedhtchang could you check the tests?

tedhtchang:
In a freez_graph_test.py log, for example:
https://source.cloud.google.com/results/invocations/4266b4b2-9313-48dd-a120-8579baf5a4e6/targets/%2F%2Ftensorflow%2Fpython%2Ftools:freeze_graph_test/log
"File "/tmp/botexec/bazel-out/k8-opt/bin/tensorflow/python/tools/freeze_graph_test.runfiles/org_tensorflow/tensorflow/python/tools/freeze_graph_test.py", line 151, in testFreezeGraphV1
    self._testFreezeGraph(saver_pb2.SaverDef.V1)

My code has this logic if (os.path.isfile(save_path) and self._write_version != saver_pb2.SaverDef.V1). I am not sure how testFreezeGraphV1 test triggers this error. Let me take a look freez_graph_test.py first.

tedhtchang:
@drpngx The problem is freez_graph_test.py does not pass [saver_pb2.SaverDef.V1](https://github.com/tensorflow/tensorflow/blob/f99275a6a309699c73e1bbebd89ba9aa32e79aa3/tensorflow/python/tools/freeze_graph_test.py#L151) into freez_graph.py. And in freez_graph.py, it does not explicitly initialize the Saver() to use [write_version=saver_pb2.SaverDef.V1](https://github.com/tensorflow/tensorflow/blob/2ebfeda9bad9d1f867205fa8de3db564953ebc23/tensorflow/python/tools/freeze_graph.py#L103) but still be able to restore a V1 checkpoint file.
Which solution do you prefer?
1. Drop the testFreezeGraphV1 test since V1 will be deprecated.
2. Introduce additional argument (write_version=saver_pb2.SaverDef.V2) to the freez_test.py
3. Drop this PR ;however people will not get the useful message when passing any V2 checkpoint file to the restore function.

drpngx:
Option 2 seems to be the best. Thank you!

tedhtchang:
@drpngx Two other failed [tests](https://source.cloud.google.com/results/invocations/35acfd09-b536-4b71-b689-aed4f35f81cb/targets/%2F%2Ftensorflow%2Fcontrib%2Fslim:evaluation_test/log), testAdditionalHooks and testRestoredModelPerformance, also have same problem that the Saver is initialized with [write_version=saver_pb2.SaverDef.V1](https://github.com/tensorflow/tensorflow/blob/4cb0c13c7779da536cac6c682180c5757611b384/tensorflow/contrib/slim/python/slim/evaluation_test.py#L239). I will open another PR to remove write_version=saver_pb2.SaverDef.V1 because I believe it is also fine with V2.

tensorflowbutler:

drpngx:
@tedhtchang is the other PR fixing a test in this one? We try not to submit PRs that have broken tests, to make it easier to bisect. If that's the case, then please fold the other PR into this one.

tedhtchang:
@drpngx I folded other PRs into this one.

rmlarsen:
@drpngx is this good to go?

drpngx:
What I'm saying is that we shouldn't pass the checkpoint version to the `freeze_graph_with_def_protos` that will cause it to *write* in the `V1` format.

drpngx:
OK on second thoughts actually, this is good.

rmlarsen:
@drpngx thanks!

FloFoer:
How am i supposed to use this with the Tensorflow Object Detection API? 
There i have a "Fine-Tune Checkpoint" which is an option in the configuration file that points to a single file. I use the Inception v2 Checkpoint from [here](https://github.com/tensorflow/models/tree/master/research/slim) - again, a single file. 
Previously (r1.5) the training worked just fine with that. However, using TF from the current master or r1.6 (which includes this change here, other than r1.5) it only produces the ValueError you raise in saver.py at line 1749. 
Is the API simply missing support for this and i should post it there or am i doing it wrong?

martinwicke:
@tedhtchang @drpngx it does look like this may be too restrictive. The inception_v2 checkpoint is a v1 checkpoint, and that should be read automatically without raising an error. 

On other words, we cannot assume that we read only v2 checkpoints just because we want to write only v2 checkpoints. Removing the check in saver.py should be enough to fix this, the other changes are still good (I think)

@av8ramit I think this will require a cherry-pick for 1.6 to fix the backwards incompatible change in behavior.

av8ramit:
@gunan who owns the rc1 including cherrypicks.

drpngx:
Yeah, I didn't realize v1 could also be single files. Sorry. We should simply revert the PR.

PR #14343, opened 2017-11-08
Add curses install note for Windows in tfdbg docs

pvaneck:
On Windows, using the curses UI with tfdbg is a nicer experience than the suggested readline, so I thought it would be worthwhile to include a note in the documentation about Windows curses installation.

### DIFF
--- a/tensorflow/docs_src/programmers_guide/debugger.md
+++ b/tensorflow/docs_src/programmers_guide/debugger.md
@@ -9,10 +9,16 @@ lets you view the internal structure and states of running TensorFlow graphs
 during training and inference, which is difficult to debug with general-purpose
 debuggers such as Python's `pdb` due to TensorFlow's computation-graph paradigm.
 
-> NOTE: The system requirements of tfdbg on supported external platforms include
-> the following. On Mac OS X, the `ncurses` library is required. It can be
-> installed with `brew install homebrew/dupes/ncurses`. On Windows, `pyreadline`
-> is required. If you use Anaconda3, you can install it with a command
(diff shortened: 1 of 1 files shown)


tensorflow/docs_src/programmers_guide/debugger.md
```diff
@@ -9,10 +9,16 @@ lets you view the internal structure and states of running TensorFlow graphs
...
 
-> NOTE: The system requirements of tfdbg on supported external platforms include
-> the following. On Mac OS X, the `ncurses` library is required. It can be
-> installed with `brew install homebrew/dupes/ncurses`. On Windows, `pyreadline`
-> is required. If you use Anaconda3, you can install it with a command
+> NOTE: TensorFlow debugger uses a
+> [curses](https://en.wikipedia.org/wiki/Curses_\(programming_library\))-based
+> text user interface. On Mac OS X, the `ncurses` library is required and can
+> be installed with `brew install homebrew/dupes/ncurses`. On Windows, an
+> unofficial curses package for various Python versions can be downloaded
```
caisq: I personally find that distribution of curses on Windows not 100% reliable. For example, many font attributes, such as background colors and underscores, do not get rendered correctly on some machines. So while I'm okay with adding mention of that unofficial distribution of curses, I'd suggest moving it after the Window pyreadilne option.

caisq: Also, please add warning to the user that the Windows distribution may not work as reliably as curses on Linux and Mac.

caisq (approved):

caisq:
@pvaneck Thanks!

PR #14351, opened 2017-11-08
memmap changes

cloudbank:
@andrewharp  @martinwicke 
Here is the new PR with changes as discussed in 12922

### DIFF
--- a/tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
+++ b/tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
@@ -15,20 +15,26 @@ limitations under the License.
 
 package org.tensorflow.contrib.android;
 
+import android.content.Context;
 import android.content.res.AssetManager;
 import android.os.Build.VERSION;
 import android.os.Trace;
 import android.text.TextUtils;
 import android.util.Log;
 import java.io.ByteArrayOutputStream;
+import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.DoubleBuffer;
 import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 import java.nio.LongBuffer;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel;
 import java.util.ArrayList;
 import java.util.List;
 import org.tensorflow.Graph;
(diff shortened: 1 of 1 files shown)


cloudbank:
@martinwicke @andrewharp  Could someone be assigned to review, and we could finish it?

martinwicke:
@andrewharp can you continue to look at this? It's a continuation of another PR.

cloudbank:
@andrewharp I added a check for res/raw and if model file is not in there, check in the assets.  Can you review it please

cloudbank:
@martinwicke can anyone else review it?

martinwicke:
@petewarden can you take a look or find someone who can?

cloudbank:
@martinwicke  What can I do to move this along?

andrewharp:
Sorry, have been on vacation the past couple of weeks. Will take a look as soon as I can.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -50,6 +56,117 @@
...
+    this.runner = sess.runner();
+    this.modelName = modelName;
+    MappedByteBuffer buffer = null;
+    FileChannel fileChannel = null;
+    RandomAccessFile raf = null;
+    File cacheFile = null;
+    try {
+      cacheFile = createCacheFile(ctx, modelName, true);
+      if (cacheFile != null) {
+        //maybe do fis
```
andrewharp: Space after // and please spell out the acronym

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -50,6 +56,117 @@
...
+        raf = new RandomAccessFile(cacheFile, "r");
+        fileChannel = raf.getChannel();
+      } else {
+        //try assets
+        if (!modelName.endsWith(".pb")) {
```
andrewharp: Not sure if we should assume this, and in any case it seems redundant with the instance in createFileCache()

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -50,6 +56,117 @@
...
+      }
+      inputStream = context.getAssets().open(filename);
+    }
+    File cacheFile = new File(context.getCacheDir(), filename);
+    if (cacheFile.createNewFile() == false) {
+      cacheFile.delete();
+      cacheFile.createNewFile();
+    }
+    FileOutputStream fileOutputStream = new FileOutputStream(cacheFile);
+    byte[] buffer = new byte[1024 * 512];
```
andrewharp: Any particular reason for this number? If so, please comment.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -50,6 +56,117 @@
...
+    InputStream inputStream = null;
+    if (res) {
+      int resourceId = context.getResources().getIdentifier(filename, "raw", context.getPackageName());
+      if (resourceId == 0) {
+        return null;
+      }
+      inputStream = context.getResources().openRawResource(resourceId);
+    } else {
+      if (!filename.endsWith(".pb")) {
+        filename += ".pb";
```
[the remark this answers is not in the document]
cloudbank: @andrewharp  For instance, in your demo since you use assets, you have the filename with the extension, but I use raw and it does not use the extension.  I am simply checking if they have it before I try to pull the file from assets.  What assumption are you referring to?

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
cloudbank: @andrewharp I will remove this

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
cloudbank: @andrewharp this is a redundant typo and I will remove

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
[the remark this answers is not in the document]
andrewharp: Is it a requirement to not use an extension on files in raw?

The assumption this seems to be making is that any model loaded from assets will have a .pb extension, even if the calling site doesn't request it. For example, if I name my graph "model.bin_proto" for whatever reason it will now try to load "model.bin_proto.pb". I don't think we want to force this.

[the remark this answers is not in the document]
cloudbank: Yes. Res/raw requires no extension while assets requires one. I am only making this worst case assumption: I passed in a model name without extension and now, if I cannot find it in raw, I want to ensure if it is in assets, I have added its extension to the filename passed in before I try to find it there.  Your demo code for instance:
private static final String MODEL_FILE = "file:///android_asset/rounded_graph.pb";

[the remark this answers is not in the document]
andrewharp: Just to be clear, res/raw does not require an extension, or requires that the file has no extension? That is, is it possible to put a file with an extension in there?

Regarding loading from assets, it seems the caller should have the freedom to make the extension whatever they want and not be forced to use ".pb". Just because the examples use models ending with pb does not mean we should assume this is always the case.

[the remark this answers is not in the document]
cloudbank: Res/raw file are accessible by resource id or filename without the extension.  I did not consider that files outside of .pb would be involved.  Protobufs used by tensorflow are usually binary and hold that extension.  I would have to do more research to find out if there is a use case for that; I tend to agree it is a bit brittle and can move that check outside of the interface.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
cloudbank: @andrewharp For Android, it might be a bit large.  I see that the cache is recommended to max out at 1MB.  I guess I should change it to use internal storage.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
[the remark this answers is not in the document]
andrewharp: Yes, please remove the forced file extension and address the other comments and I think this is good to go.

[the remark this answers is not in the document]
cloudbank: @andrewharp I am still testing it, but should be done shortly.  The changes to aapt make the res/raw access even more complicated, but there is a workaround.

cloudbank:
@andrewharp I update the file to remove the extension check and try the res/raw first with a fallback to assets.

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -50,6 +56,112 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to open graph file: " + modelFile, e);
+    }
+    try {
+      buffer = fileChannel.map(FileChannel.MapMode.READ_ONLY, 0, fileChannel.size());
+      if (!buffer.isLoaded()) {
+        buffer.load();
+      }
+      byte[] graphDef = new byte[buffer.limit()];
+      buffer.get(graphDef);
```
andrewharp: Isn't this a copy operation?

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -50,6 +56,112 @@
...
+    FileChannel fileChannel = null;
+    RandomAccessFile raf = null;
+    File copy = null;
+    try {
+      copy = copyFile(ctx, modelName, true);
+      if (copy != null) {
+        raf = new RandomAccessFile(copy, "r");
+        fileChannel = raf.getChannel();
+      } else {
+        //try assets
```
andrewharp: space after //

tensorflow/contrib/android/java/org/tensorflow/contrib/android/TensorFlowInferenceInterface.java
```diff
@@ -50,6 +56,112 @@
...
+      loadGraph(graphDef, g);
+      Log.i(TAG, "Successfully loaded model from map" + modelName);
+    } catch (IOException ioe) {
+      throw new RuntimeException("Failed to map  model from" + modelName, ioe);
+    } finally {
+      try {
+        if (raf != null) {
+          raf.close();
+        }
+        if (copy != null) {
```
andrewharp: Do we want to delete this? What is the copy time -- maybe it would be better to keep it cached if possible for faster loading the next time?

andrewharp (changes requested):
So currently, this gives us:

`TensorFlowInferenceInterface(Context ctx, String modelName)`
- copies from in order of priority:
  - res/raw
  - assets
- memmaps copied file
- copies to buffer
- loads buffer into graph

`TensorFlowInferenceInterface(AssetManager manager, String modelName)`
- opens input stream on model in, in order of priority:
  - assets (if ANDROID_ASSET is prefixed, strictly enforces that model come from assets)
  - disk
- loads graph from inputstream


To try and make the behavior as consistent as possible, maybe we should have the context method also load from disk as well? Then at least it's mostly just a superset of the behavior of the AssetManager version.


Questions:
How long does the filecopy take?

Hadn't  we decided to try memmapping directly from uncompressed assets? Or is that only possible in native code?

drpngx:
@cloudbank could you address @andrewharp comments?

cloudbank:
@drpngx  Yes.  I am refactoring.

drpngx:
Nice, looking forward for the next push!

tensorflowbutler:

drpngx:
@cloudbank any luck with this?

cloudbank:
I have been busy elsewhere.  I will work on it this week.

Best regards,

Sabine G. Vogel

tensorflowbutler:

tensorflowbutler:
Nagging Assignees @jhseu, @yifeif: It has been 14 days with no activity and this issue has an assignee. Please update the label and/or status accordingly.

jhseu:
Closing due to inactivity, but feel free to send a new pull request with the requested changes.

Issue #14284, opened 2017-11-06
Estimator from Keras by model_to_estimator cannot export


yjmade:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 10.12
- **TensorFlow installed from (source or binary)**: Source
- **TensorFlow version (use command below)**: 1.4 or Master 

### Describe the problem
when calling model_to_estimator, the model_fn create by _create_keras_model_fn didn't set ```export_outputs``` in the returned ```EstimatorSpec```, see [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/_impl/keras/estimator.py#L182). That make it unable to export to SavedModel, the following error raised when call ```estimator.export_savedmodel()```

```
packages/tensorflow/python/estimator/export/export.py", line 193, in build_all_signature_defs
    raise ValueError('export_outputs must be a dict.')
ValueError: export_outputs must be a dict.
```

I would like contrib a PR for fixing this if is OK

bignamehyp:
@yifeif , can you please take a look at this issue? Thanks,

yifeif:
Thanks @yjmade. A PR is more than welcomed! Feel free to request a review from me whenever you have it ready. cc: @fchollet.

yjmade:
PR Merged, close now

PR #14354, opened 2017-11-08
Fix estimator convert from Keras to export_savedmodel() #14284

yjmade:
PR for #14284 
[FIX]the estimator generate by tf.keras.model_to_estimator() cannot export saved_model because the model_fn provided by _create_keras_model_fn wasn't set export_outputs in the returned EstimatorSpec. Here I provide a default export_outputs with serve_default key and Predict API, and the result inside is same as predictions

[FIX]_save_first_checkpoint call saver.save with only a path and without filename, that make the saved ckpt files with name like `{model_dir}/.meta` and `{model_dir}/.index`, which is not be able to found by latest_checkpoint("{model_dir}"). As state by Saver.save(), save_path should be a path to the checkpoint name. So to fix this, I change the name to `{model_dir}/keras_model.ckpt`


tensorflow/python/keras/_impl/keras/estimator.py
```diff
@@ -33,6 +35,9 @@
 from tensorflow.python.platform import tf_logging as logging
 from tensorflow.python.training import saver as saver_lib
 from tensorflow.python.training import training_util
+from tensorflow.python.saved_model import signature_constants
```
yifeif: Add new dependency to tensorflow/python/keras/BUILD.

yjmade:
Thanks @yifeif for the review. The PR has been updated with review comments addressed. Please take a look.

yifeif (approved):

yifeif:
Thanks @yjmade!

PR #14367, opened 2017-11-08
golang: added Session.ListDevices method

anight:
Implemented Session.ListDevices() method


anight:
Please ignore this one

PR #14370, opened 2017-11-08
Change `NHWC`/`NCHW` to `NWC`/`NCW` for conv1d

yongtang:
While working on #13105 I noticed that in the current code base `conv1d` uses `NHWC`/`NCHW` which should really be `NWC`/`NCW`.

This fix addresses this issue and keep `NHWC`/`NCHW` compatible internally so that users will not be impacted.

### DIFF
--- a/tensorflow/python/ops/nn_ops.py
+++ b/tensorflow/python/ops/nn_ops.py
@@ -2260,9 +2260,9 @@ def conv1d(value, filters, stride, padding,
 
   Given an input tensor of shape
     [batch, in_width, in_channels]
-  if data_format is "NHWC", or
+  if data_format is "NWC", or
     [batch, in_channels, in_width]
-  if data_format is "NCHW",
+  if data_format is "NCW",
   and a filter / kernel tensor of shape
   [filter_width, in_channels, out_channels], this op reshapes
   the arguments to pass them to conv2d to perform the equivalent
@@ -2287,9 +2287,9 @@ def conv1d(value, filters, stride, padding,
       the filter is moved right at each step.
     padding: 'SAME' or 'VALID'
     use_cudnn_on_gpu: An optional `bool`.  Defaults to `True`.
-    data_format: An optional `string` from `"NHWC", "NCHW"`.  Defaults
-      to `"NHWC"`, the data is stored in the order of
-      [batch, in_width, in_channels].  The `"NCHW"` format stores
+    data_format: An optional `string` from `"NWC", "NCW"`.  Defaults
+      to `"NWC"`, the data is stored in the order of
+      [batch, in_width, in_channels].  The `"NCW"` format stores
       data as [batch, in_channels, in_width].
     name: A name for the operation (optional).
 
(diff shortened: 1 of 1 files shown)


jhseu:
Not sure we'll be able to change this since it's not backwards-compatible.

tensorflow/python/ops/nn_ops.py
```diff
@@ -2202,9 +2202,9 @@ def conv1d(value, filters, stride, padding,
       the filter is moved right at each step.
     padding: 'SAME' or 'VALID'
     use_cudnn_on_gpu: An optional `bool`.  Defaults to `True`.
-    data_format: An optional `string` from `"NHWC", "NCHW"`.  Defaults
-      to `"NHWC"`, the data is stored in the order of
-      [batch, in_width, in_channels].  The `"NCHW"` format stores
+    data_format: An optional `string` from `"NWC", "NCW"`.  Defaults
```
martinwicke: We cannot change the signature this way, as @jhseu said, this is a backwards incompatible change. Instead, you could accept both NHWC and NWC. You can even only document NWC to push users away from NHWC. Optionally, you can use @deprecated_arg_values to mark NWHC explicitly deprecated (will trigger a warning if used).

martinwicke (changes requested):

tensorflow/python/ops/nn_ops.py
yongtang: Thanks @martinwicke @jhseu for the review. The current PR actually accept both `NWC` and `NHWC`, (see Ln 2219):
```
    if data_format is None or data_format == "NHWC" or data_format == "NWC":
      data_format = "NHWC"
      spatial_start_dim = 1
      strides = [1, 1, stride, 1]
    elif data_format == "NCHW" or data_format == "NCW":
      data_format = "NCHW"
      spatial_start_dim = 2
      strides = [1, 1, 1, stride]
    else:
```

I think `@deprecated_arg_values` is good. Will update the PR to add it.

yongtang:
Thanks @martinwicke for the tip of `@deprecated_arg_values`. The PR has been updated with `@deprecated_arg_values` added. Please take a look.

martinwicke (approved):

martinwicke:
Jenkins, test this please.

asimshankar:
(From API review)

Is using "width" for 1D somehow better than "height"? (Or the other way around).
Sounds like `W` or `H` are also inappropriate for 1D values.

I'd be tempted to leave things as-is for now?

yongtang:
Thanks @asimshankar for the review. The NCW/NWC was mainly to following other ops in nn like `convolution`:
https://www.tensorflow.org/api_docs/python/tf/nn/convolution


data_format: A string or None. Specifies whether the channel dimension of the input and output is the last dimension (default, or if data_format does not start with "NC"), or the second dimension (if data_format starts with "NC"). For N=1, the valid values are "NWC" (default) and "NCW". For N=2, the valid values are "NHWC" (default) and "NCHW". For N=3, the valid values are "NDHWC" (default) and "NCDHW".


I also noticed that a similar op `conv1d_transpose` (as opposed to `conv1d` in this PR) is using `NCW/NWC` (in Ln 2328 of the same file). That was why I opened this PR.

Not sure what is the best way to handle the inconsistencies but happy to do one way or another.

martinwicke:
Ok, checked with @asimshankar, this is fine.

Issue #14054, opened 2017-10-28
[Keras] TimeDistributed wrapper error:  'NoneType' object has no attribute 'as_list'


footh:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in TensorFlow)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04
- **TensorFlow installed from (source or binary)**: source
- **TensorFlow version (use command below)**: built from master pulled on 10/26
- **Python version**: 3.6
- **CUDA/cuDNN version**: N/A
- **GPU model and memory**: N/A
- **Exact command to reproduce**: See below

### Describe the problem
Issue arises when using TimeDistributed wrapper in release *1.4*. See simple code example below which works fine in version *1.3*.

### Source code / logs
```
def td():
    model = tf.keras.models.Sequential()
    model.add(tf.keras.layers.Dense(8, input_shape=(16,)))
    model.add(tf.keras.layers.Dense(4))
    model.summary()
    
    frame_input = tf.keras.layers.Input(shape=(10, 16))
    x = tf.keras.layers.TimeDistributed(model)(frame_input)
    x = tf.keras.layers.Flatten()(x)
    
    full_model = tf.keras.models.Model(inputs=frame_input, outputs=x)
    full_model.summary()
```
produces this trace:

```
     x = tf.keras.layers.TimeDistributed(model)(frame_input)
  File "/Users/guest/dev/anaconda3/envs/tensorflow/lib/python3.6/site-packages/tensorflow/python/keras/_impl/keras/engine/topology.py", line 252, in __call__
    output = super(Layer, self).__call__(inputs, **kwargs)
  File "/Users/guest/dev/anaconda3/envs/tensorflow/lib/python3.6/site-packages/tensorflow/python/layers/base.py", line 575, in __call__
    outputs = self.call(inputs, *args, **kwargs)
  File "/Users/guest/dev/anaconda3/envs/tensorflow/lib/python3.6/site-packages/tensorflow/python/keras/_impl/keras/layers/wrappers.py", line 238, in call
    output_shape = self._compute_output_shape(input_shape).as_list()
  File "/Users/guest/dev/anaconda3/envs/tensorflow/lib/python3.6/site-packages/tensorflow/python/keras/_impl/keras/layers/wrappers.py", line 193, in _compute_output_shape
    child_input_shape).as_list()
AttributeError: 'NoneType' object has no attribute 'as_list'
[...]

PR #14377, opened 2017-11-08
Fix indenting on cache check in Network layer class

footh:
Resolves [#14054](https://github.com/tensorflow/tensorflow/issues/14054)

The `else` and ensuing code on the cache check was indented one spot too far. This resulted in the cache hit code not being executed when there was a cache hit (and the cache miss code never being called because there was a cache hit).


fchollet (approved):
LGTM

martinwicke:
I apologize that this has languished for so long.

This code has moved to network.py. Could you rebase?

@fchollet, should we make this fix internally?

footh:
Looking at the conflict (lines 2125-2135), it looks like this issue has already been resolved. So this PR is probably moot at this point.

footh:
Oops, no, I take that back. I just reread your comment and see the bug is in the network.py class. I will rebase.

tensorflow/python/layers/base.py
```diff
@@ -1358,7 +1358,6 @@ def __repr__(self):
                                                          self.get_shape(),
                                                          self.dtype.name)
 
-
```
martinwicke: I think we want this blank line.

martinwicke (changes requested):

martinwicke (approved):

tensorflow/python/layers/base.py
martinwicke: Never mind.

martinwicke:
Thanks! Testing now.

yifeif:

PR #14385, opened 2017-11-08
golang: added Session.ListDevices method

anight:
Implemented Session.ListDevices() method

### DIFF
--- a/tensorflow/go/session.go
+++ b/tensorflow/go/session.go
@@ -65,6 +65,53 @@ func NewSession(graph *Graph, options *SessionOptions) (*Session, error) {
 	return s, nil
 }
 
+type Device struct {
+	Name, Type  string
+	MemoryLimit int64
+}
+
+// Return list of devices associated with a Session
+func (s *Session) ListDevices() ([]Device, error) {
+	var devices []Device
+
+	status := newStatus()
+	devices_list := C.TF_SessionListDevices(s.c, status.c)
+	if err := status.Err(); err != nil {
+		return nil, fmt.Errorf("TF_SessionListDevices() failed: %v", err)
+	}
+	defer C.TF_DeleteDeviceList(devices_list)
+
+	for i := 0; i < int(C.TF_DeviceListCount(devices_list)); i++ {
+		status = newStatus()
+		device_name := C.TF_DeviceListName(devices_list, C.int(i), status.c)
+		if err := status.Err(); err != nil {
+			return nil, fmt.Errorf("TF_DeviceListName(index=%d) failed: %v", i, err)
+		}
+
+		status = newStatus()
+		device_type := C.TF_DeviceListType(devices_list, C.int(i), status.c)
+		if err := status.Err(); err != nil {
+			return nil, fmt.Errorf("TF_DeviceListType(index=%d) failed: %v", i, err)
+		}
+
+		status = newStatus()
+		memory_limit := C.TF_DeviceListMemoryBytes(devices_list, C.int(i), status.c)
+		if err := status.Err(); err != nil {
+			return nil, fmt.Errorf("TF_DeviceListMemoryBytes(index=%d) failed: %v", i, err)
+		}
+
+		device := Device{
+			Name:        C.GoString(device_name),
+			Type:        C.GoString(device_type),
+			MemoryLimit: int64(memory_limit),
+		}
+
+		devices = append(devices, device)
+	}
+
+	return devices, nil
+}
+
 // Run the graph with the associated session starting with the supplied feeds
 // to compute the value of the requested fetches. Runs, but does not return
 // Tensors for operations specified in targets.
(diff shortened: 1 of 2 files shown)


tensorflow/go/session.go
```diff
@@ -65,6 +65,53 @@ func NewSession(graph *Graph, options *SessionOptions) (*Session, error) {
...
+
+	status := newStatus()
+	devices_list := C.TF_SessionListDevices(s.c, status.c)
+	if err := status.Err(); err != nil {
+		return nil, fmt.Errorf("TF_SessionListDevices() failed: %v", err)
+	}
+	defer C.TF_DeleteDeviceList(devices_list)
+
+	for i := 0; i < int(C.TF_DeviceListCount(devices_list)); i++ {
+		status = newStatus()
```
asimshankar: Do we need to create a `newStatus()` multiple times?Seems like we can re-use the one `status` created in line 77?

tensorflow/go/session.go
```diff
@@ -65,6 +65,53 @@ func NewSession(graph *Graph, options *SessionOptions) (*Session, error) {
 	return s, nil
 }
 
+type Device struct {
```
asimshankar: Mind adding some commentary to make godoc nice?

tensorflow/go/session.go
```diff
@@ -65,6 +65,53 @@ func NewSession(graph *Graph, options *SessionOptions) (*Session, error) {
...
+	Name, Type  string
+	MemoryLimit int64
```
asimshankar: Perhaps call this `MemoryLimitBytes` to make the units clearer?

tensorflow/go/session.go
```diff
@@ -65,6 +65,53 @@ func NewSession(graph *Graph, options *SessionOptions) (*Session, error) {
...
+}
+
+// Return list of devices associated with a Session
+func (s *Session) ListDevices() ([]Device, error) {
+	var devices []Device
+
+	status := newStatus()
+	devices_list := C.TF_SessionListDevices(s.c, status.c)
+	if err := status.Err(); err != nil {
+		return nil, fmt.Errorf("TF_SessionListDevices() failed: %v", err)
```
asimshankar: Listing the C API method in the error string might be confusing to a Go user since most users should have no idea what `TF_Session*` means.

Perhaps rephrase the error messages? For example, this one can perhaps just be `return nil, err`, while the ones in the loop can be `return nil, fmt.Errorf("failed to fetch name for device #%d: %v", i, err)` etc.?

tensorflow/go/session_test.go
```diff
@@ -283,3 +283,30 @@ func TestSessionConfig(t *testing.T) {
...
+	devices, err := s.ListDevices()
+	if err != nil {
+		t.Fatalf("ListDevices(): %v", err)
+	}
+
+	if len(devices) == 0 {
+		t.Fatalf("no devices detected")
+	}
+
+	var cpus, gpus int
```
asimshankar: Perhaps remove this line onwards (since it is effectively dead code?)

asimshankar (changes requested):
Thanks for the contribution!
Some minor suggestions.

(Curious about your use case as well, if you have some code/blog-post etc. that you'd like to share, that would be great. If not, that's okay too :)

tensorflow/go/session.go
anight: Done, thanks.

tensorflow/go/session.go
anight: Done, thanks.

tensorflow/go/session.go
anight: A rationale behind this type of error message texts is simple. By copying exact failed function name to an error message we allow easier code search in case of potential problems. If we leave a less distinctive text like "failed to fetch name ..." then future developers potentially might have found it harder to find exact place in the code by searching through all similar messages. And yes, these error messages are for developers eyes mostly, because normally these functions do not fail.

tensorflow/go/session.go
anight: Done, thanks. I tried to find some info on how safe is to re-use statuses and I found nothing. Other places in code does so though.

tensorflow/go/session_test.go
anight: Done, thanks.

anight:
My use case was I was trying to balance session creation manually across multiple GPUs so I needed to fetch a number of available GPUs together with available memory. So I decided to create an empty session at the start and then fetch all info via ListDevices().

asimshankar (approved):
Thanks!

asimshankar:
Jenkins, test this please

jhseu:
Jenkins, test this please

caisq:

yifeif:

yifeif:
windows rerun at http://ci.tensorflow.org/job/tensorflow-pr-win-cmake-py/5945/

PR #14397, opened 2017-11-09
remove duplicated code

horance-liu:
I'm remove duplicated code, and to avoid missing when adding new error code.


jhseu:
Seems like it was intentionally done this way, but adding @skye to chime in.

skye:
@josh11b actually wrote this, but I agree we should keep it as-is. It's actually less error-prone in the long run, as it doesn't depend on TF_Code and the proto being in sync (and will fail if we add a new code to the proto without adding it to TF_Code).

horance-liu:
@skye @jhseu It has delayed for a long time. the modification will be less error-prone in future, and remove duplicated code implementation.

jhseu:
Closing as per Skye's comments.

PR #14401, opened 2017-11-09
CMake: configure default string values of options properly

### LINKED ISSUE #14400: [CMake] Default values of some options are not properly set
```tensorflow_PATH_CUDNN_STATIC_LIB``` and ```tensorflow_PATH_NCCL_STATIC_LIB``` are supposed to be configured as the value of ```tensorflow_PATH_STATIC_LIB``` by default.

However, because cmake options's default values are either OFF or ON, this doesn't work as supposed.

I'll post a Pull Request addressing this issue in a few minutes.

myungjoo:
Because cmake configures defaults values as ON or OFF only,
string values as default doesn't work.

Thus, when it is set "OFF", we need to re-set the values.

Fixes #14400

Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>


mrry (approved):
Thanks for the contribution!

mrry:

PR #14409, opened 2017-11-09
added patch needed for jetson tx2/1

tano297:
arm64 does not support NUMA, so TryToReadNumaNode in cuda_gpu_executor.cc has to be modified accordingly to avoid a failure at runtime. 

Currently everybody using a jetson board needs to apply this patch manually, which is a pain if you forget, because it fails at runtime, and not at build. This means that you compile for a couple of hours only to realize you should've done this first.


tano297:
I signed it!

jhseu (approved):

jhseu:
Jenkins, test this please

tano297:
How should we move forward with this? @jhseu

PR #14418, opened 2017-11-09
enum34 is only required for Python < 3.4

boegel:
I ran into problems when building TensorFlow 1.4.0 on top of Python 3.6.3 because `enum34` is pulled in, but not compatible with Python 3.6, i.e. `python -c 'import tensorflow'` lead to:

```
AttributeError: module 'enum' has no attribute 'IntFlag'
```

See also https://bitbucket.org/stoneleaf/enum34/issues/19.

### DIFF
--- a/tensorflow/tools/pip_package/setup.py
+++ b/tensorflow/tools/pip_package/setup.py
@@ -33,7 +33,6 @@ _VERSION = '1.4.0'
 
 REQUIRED_PACKAGES = [
     'absl-py',
-    'enum34 >= 1.1.6',
     'numpy >= 1.12.1',
     'six >= 1.10.0',
     'protobuf >= 3.4.0',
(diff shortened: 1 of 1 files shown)


gunan:
How would this work if we build TF on python 3.5, but it ends up getting installed on a python 3.3 system?

boegel:
@gunan Good question, I didn't take that into account...

In that case I guess you can keep requiring `enum34` for both Python 3.4 and 3.5 too.

But for Python 3.6, it's wrong to install it. The installation works fine, but as soon as you try to use TensorFlow, it will crash hard on top of Python 3.6 with the error I mentioned.

caisq:
@boegel Interesting finding. @gunan @av8ramit  I think this has to do with the fact that we don't do real testing on python 3.6 (on Linux or Mac) currently.

gunan:
@caisq we are aware. This is already being worked on.
This is slowed down due to the confusion about multiple python versions supported. A protocol about what python versions are going to be supported is being prepared, and we will clarify our python 3 support story with that going forward.

@boegel one of my concerns is the evaluation time of this file (which I do not fully know).
Is it being evaluated when we are building the pip package? So if we build the pip package on a py 3.6 system,  then if people end up installing this on a py 3.4 system, do they not get py 3.4?
I guess if setup.py gets evaluated during installation, this should be OK.
@yifeif in case she can answer this concern.

yifeif:
Looks like we can also do something like https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies
I would expect this to be evaluated during installation.

gunan:
Great, let's go with that instead then.

@boegel, would you like to update your PR? we will also need you to take care of the Google CLA.
or we can make this change on our end.

gunan:
Closing in favor of #14730 since that one has the CLA taken care of.

PR #14426, opened 2017-11-10
Fix absl flag initialization in cloud_tpu_profiler

jart:
This fixes a regression caused by 2652704b576adc16b4d735f651cea1024e88b72e where the command would not run. See also: tensorflow/tensorboard#716

This is caused by the PIP generated program wrapper not invoking the `tf.app.run` that normally goes in the `if __main__` clause of a script. That runner basically initializes flags. But as far as I can tell, this is the only pip console script in TensorFlow that still uses flags. The other two appear to have migrated to argparse. So this change should be sufficient.

CC: @nfelt @PrashantJalan @yifeif


jhseu (approved):

jhseu:
Jenkins, test this please

yifeif:

PR #14443, opened 2017-11-10
fix the distributed training problem for the estimator api

timefly666:
Current version will fail as I tried to use the `tf.estimator.train_and_evaluate` api for distributed training on the same machine. All the gpus will be occupied even when I  set `session_config.gpu_options.allow_growth = True` for `tf.estimator.RunConfig`.

### DIFF
--- a/tensorflow/python/estimator/training.py
+++ b/tensorflow/python/estimator/training.py
@@ -670,11 +670,19 @@ class _TrainingExecutor(object):
             'RunConfig or set the TF_CONFIG environment variable.')
 
     logging.info('Start Tensorflow server.')
+
+    if config.session_config is None:
+      session_config=config_pb2.ConfigProto(log_device_placement=False)
+    else:
+      session_config=config_pb2.ConfigProto(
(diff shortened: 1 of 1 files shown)


martinwicke:
@xiejw FYI -- I think this makes sense: the gpu options set in TF_CONFIG should be used. If one wants to vary them by worker/PS, once can send different TF_CONFIG (though that would only rarely be necessary). Passing GPU options to non-GPU workers should be harmless.

martinwicke (approved):

yifeif:

yifeif:
@Dong--Jian could you update the change to handle when config.session_config is None? Thanks!

Dong--Jian:
Thanks for the reminder. I add the corresponding code.

drpngx:
@Dong--Jian could you rebase and push again?

Dong--Jian:
I have rebased with the master.

drpngx:
Jenkins, test this please.

PR #14447, opened 2017-11-10
PrefetchDataset with buffer_size==0 results in deadlock

boeddeker:
Hey, 

I experimented with `tf.data.Dataset.prefetch` and found that an assert inside the code is missing.
When `buffer_size` is zero, I got a deadlock.
Here is a toy example to reproduce my bug:

```python
import tensorflow as tf
def test(buffer_size):
    ds = tf.data.Dataset.range(5)
    ds = ds.prefetch(buffer_size=buffer_size)
    iterator = ds.make_one_shot_iterator()
    entry = iterator.get_next()
    with tf.Session() as sess:
        try:
            while True:
                print(sess.run(entry))
        except tf.errors.OutOfRangeError:
            pass
        
test(1)
test(0)  # deadlock
```
and here my tensorflow version
```bash
$ pip show tensorflow
Name: tensorflow
Version: 1.4.0
Summary: TensorFlow helps the tensors flow
Home-page: https://www.tensorflow.org/
Author: Google Inc.
Author-email: opensource@google.com
License: Apache 2.0
Location: .../lib/python3.6/site-packages
Requires: protobuf, numpy, wheel, tensorflow-tensorboard, absl-py, six, enum34
```

### DIFF
--- a/tensorflow/python/data/ops/dataset_ops.py
+++ b/tensorflow/python/data/ops/dataset_ops.py
@@ -1604,6 +1604,7 @@ class PrefetchDataset(Dataset):
 
   def __init__(self, input_dataset, buffer_size):
     """See `Dataset.prefetch()` for details."""
+    assert buffer_size > 0, "The buffer_size ({}) has to be > 0.".format(buffer_size)
     super(PrefetchDataset, self).__init__()
     self._input_dataset = input_dataset
     self._buffer_size = ops.convert_to_tensor(buffer_size, dtype=dtypes.int64,


boeddeker:
I signed it!

facaiy:
Good! Perhaps it's better to check `buffer_size` at c++ side?

https://github.com/tensorflow/tensorflow/blob/e951547c9ba5f883f3be9bd9b1a79ccc85b29629/tensorflow/core/kernels/prefetch_dataset_op.cc#L36-L40

boeddeker:
Yes, it would be better to have the assert in the C++ side.
But I am not sure if then the traceback gets lost.
Is the C++ exception traceback readable? I have sometimes error msgs that are unrelated to the error.
If yes, only C++ else C++ and python.

I am not familiar with the C++ code (and style), so maybe someone other can make the changes?

tensorflow/python/data/ops/dataset_ops.py
```diff
@@ -1604,6 +1604,7 @@ class PrefetchDataset(Dataset):
 
   def __init__(self, input_dataset, buffer_size):
     """See `Dataset.prefetch()` for details."""
+    assert buffer_size > 0, "The buffer_size ({}) has to be > 0.".format(buffer_size)
```
mrry: Instead of this assertion, can you add the following line to the C++ implementation [here](https://github.com/tensorflow/tensorflow/blob/047d7965d2877d7b55f4cdb0d0abdcd733f266a9/tensorflow/core/kernels/prefetch_dataset_op.cc#L39)?

```c++
OP_REQUIRES(ctx, buffer_size > 0, errors::InvalidArgument("buffer_size must be > 0"));
```

mrry (changes requested):
Thanks for the contribution! The assertion doesn't quite cover all the cases, and will cause problems if e.g. the buffer size is a `tf.Tensor`. I've added a comment to suggest where the check should go instead.

tensorflow/python/data/ops/dataset_ops.py
facaiy: The fix makes sense. Do we need a timeout test case for `buffer_size=0` on python side?

mrry: It's complicated by the fact that `buffer_size` can be a `tf.Tensor`. The `assert` will fail in that case... in fact I think it will trigger a different exception about using a `Tensor` as a boolean.

The most thorough way to do it would involve the following steps:

1. Convert `buffer_size` to a `tf.Tensor` to validate its type etc.
2. Attempt to convert it back using `tensor_util.constant_value()`.
3. If the `constant_value()` is not `None`, compare it to `0`.

facaiy: I mean your fix at c++ side is correct, because `buffer_size` is converted to an int64 there. Moreover, do we need to add a test?

mrry: Ahh, yes, that would be a good idea!

facaiy: So we'd better create a`tensorflow/tensorflow/python/kernel_tests/prefetch_dataset_op_test.py` and add a test case, right?

[range_dataset_op_test.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/range_dataset_op_test.py) is a good example, and then we can register the python file in the `BUILD` like this:

https://github.com/tensorflow/tensorflow/blob/054b515feec0a3fca4cfb1f29adbf423c9027c3a/tensorflow/python/kernel_tests/BUILD#L2925-L2929

I hope the discussion could be useful for your work, @boeddeker.

boeddeker: Thanks for the suggestions. I try to do the modifications on monday. 
But I have installed tensorflow from a whl (I am new to tf), so all my changed will be a guess if the code is correct.

If it is less work for you, I am fine when someone other implements the changes.

facaiy: @boeddeker I understand it. Don't worry, I can take over the issue if you like.

boeddeker: @facaiy Thanks, that would be great.

facaiy:
@boeddeker Hi, might we close the PR since #14508 has been merged. Thanks.

boeddeker:
@facaiy Hi, thanks for the fix. I have closed this PR.

PR #14468, opened 2017-11-10
Added Ubuntu 16.04 Dockerfile with TF 1.4 optimized for CPU with Inte…

claynerobison:
…l(R) MKL
This is an Ubuntu 16.04 Container running Tensorflow 1.4 with Intel MKL.
To test on a system with Docker installed, do this as root:

Build the container (Needs to be done only once; takes about 20 minutes)
```
# cd <tensorflow-root>tensorflow/tools/docker
# docker build . -t intelaipg/tensorflow -f Dockerfile.devel-cpu-mkl
```
Then run the container as root. :

```# docker run -it --mount type=bind,source=/root,target=/host intelaipg/tensorflow```

This will bind a directory in the container at /host to /root on the host machine.

Run your tests and benchmarks.

### DIFF
--- a/tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
+++ b/tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
@@ -1,0 +1,127 @@
+FROM ubuntu
+
+LABEL maintainer="Clayne Robison<clayne.b.robison@intel.com>"
+
+# It is possible to override these for releases.
+ARG TF_BRANCH=r1.4
+ARG BAZEL_VERSION=0.7.0
+ARG WHL_DIR=/whl
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential \
+        curl \
+        git \
+        golang \
+        libcurl3-dev \
+        libfreetype6-dev \
+        libzmq3-dev \
+        pkg-config \
+        python-dev \
+        python-pip \
+        rsync \
+        software-properties-common \
+        unzip \
+        zip \
+        zlib1g-dev \
+        openjdk-8-jdk \
+        openjdk-8-jre-headless \
+        wget \
+	locate \
+	vim \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN pip --no-cache-dir install --upgrade \
+        pip setuptools
+
+RUN pip --no-cache-dir install \
+        ipykernel \
+        matplotlib \
+        numpy \
+        scipy \
+        sklearn \
+        pandas \
+        wheel \
+        && \
+    python -m ipykernel.kernelspec
+
+# Set up Bazel.
+
+# Running bazel inside a `docker build` command causes trouble, cf:
+#   https://github.com/bazelbuild/bazel/issues/134
+# The easiest solution is to set up a bazelrc file forcing --batch.
+RUN echo "startup --batch" >>/etc/bazel.bazelrc
+# Similarly, we need to workaround sandboxing issues:
+#   https://github.com/bazelbuild/bazel/issues/418
+RUN echo "build --spawn_strategy=standalone --genrule_strategy=standalone" \
+    >>/etc/bazel.bazelrc
+WORKDIR /
+RUN mkdir /bazel && \
+    cd /bazel && \
+    wget --quiet https://github.com/bazelbuild/bazel/releases/download/$BAZEL_VERSION/bazel-$BAZEL_VERSION-installer-linux-x86_64.sh && \
+    wget --quiet https://raw.githubusercontent.com/bazelbuild/bazel/master/LICENSE && \
+    chmod +x bazel-*.sh && \
+    ./bazel-$BAZEL_VERSION-installer-linux-x86_64.sh && \
+    rm -f /bazel/bazel-$BAZEL_VERSION-installer-linux-x86_64.sh
+
+# Download and build TensorFlow.
+WORKDIR /
+RUN git clone https://github.com/tensorflow/tensorflow.git && \
+    cd tensorflow && \
+    git checkout ${TF_BRANCH}
+WORKDIR /tensorflow
+
+# Configure the build for CPU with MKL by accepting default build options and
+# setting library locations
+ENV CI_BUILD_PYTHON=python \
+   LD_LIBRARY_PATH=${LD_LIBRARY_PATH} \
+    PYTHON_BIN_PATH=/usr/bin/python \
+    PYTHON_LIB_PATH=/usr/local/lib/python2.7/dist-packages \
+    CC_OPT_FLAGS='-march=native' \
+    TF_NEED_JEMALLOC=0 \
+    TF_NEED_GCP=0 \
+    TF_NEED_CUDA=0 \
+    TF_NEED_HDFS=0 \
+    TF_NEED_S3=0 \
+    TF_NEED_OPENCL=0 \
+    TF_NEED_GDR=0 \
+    TF_ENABLE_XLA=0 \
+    TF_NEED_VERBS=0 \
+    TF_NEED_MPI=0
+RUN ./configure; exit 0 # ./configure whines unnecessarily
+
+# Build and Install TensorFlow.
+RUN LD_LIBRARY_PATH=${LD_LIBRARY_PATH} \
+    bazel build --config=mkl \
+                --copt="-mfma" \
+                --copt="-mavx2" \
+		--config="opt" \
+               --copt="-march=native" \
+                --copt="-O3" \
+                //tensorflow/tools/pip_package:build_pip_package && \
+    mkdir ${WHL_DIR} && \
+    bazel-bin/tensorflow/tools/pip_package/build_pip_package ${WHL_DIR}
+
+# Clean up Bazel cache when done, but leave the whl.
+RUN pip --no-cache-dir install --upgrade ${WHL_DIR}/tensorflow-*.whl && \
+    rm -rf /root/.cache
+
+WORKDIR /root
+
+#add welcome message with instructions
+
+RUN echo '[ ! -z "$TERM" -a -r /etc/motd ] && cat /etc/issue && cat /etc/motd' \
+	>> /etc/bash.bashrc \
+	; echo "\
+||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||\n\
+|								\n\
+| Docker container running Ubuntu				\n\
+| with TensorFlow ${TF_BRANCH} optimized for CPU		\n\
+| with Intel(R) MKL						\n\
+|								\n\
+||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||\n\
+\n "\
+	> /etc/motd
+
+CMD ["bash"]
--- a/tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
+++ b/tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
@@ -1,33 +1,16 @@
-FROM ubuntu
+FROM tensorflow/tensorflow:latest-devel
 
 LABEL maintainer="Clayne Robison<clayne.b.robison@intel.com>"
 
-# It is possible to override these for releases.
+# These arguments are parameterized. Use --build-args to override.
 ARG TF_BRANCH=r1.4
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tensorflow/tools/docker/Dockerfile.devel-cpu-mkl [hunk 1]
```diff
@@ -0,0 +1,127 @@
...
+        python-pip \
+        rsync \
+        software-properties-common \
+        unzip \
+        zip \
+        zlib1g-dev \
+        openjdk-8-jdk \
+        openjdk-8-jre-headless \
+        wget \
+	locate \
```
inc0: nit: indents aren't equal

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,127 @@
...
+	vim \
```
inc0: we need vim?

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,127 @@
...
+        pip setuptools
+
+RUN pip --no-cache-dir install \
+        ipykernel \
+        matplotlib \
+        numpy \
+        scipy \
+        sklearn \
+        pandas \
+        wheel \
```
inc0: regular docker image also have jupyter stack installed, I think it's good idea to do it, that'll also mean EXPOSE 8888

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,127 @@
...
+    TF_NEED_GCP=0 \
+    TF_NEED_CUDA=0 \
+    TF_NEED_HDFS=0 \
+    TF_NEED_S3=0 \
+    TF_NEED_OPENCL=0 \
+    TF_NEED_GDR=0 \
+    TF_ENABLE_XLA=0 \
+    TF_NEED_VERBS=0 \
+    TF_NEED_MPI=0
+RUN ./configure; exit 0 # ./configure whines unnecessarily
```
inc0: but it might whine necessarily... however I don't have good idea how to tell one from another:) still zen of python's "errors should never pass silently"...

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,127 @@
...
+#add welcome message with instructions
+
+RUN echo '[ ! -z "$TERM" -a -r /etc/motd ] && cat /etc/issue && cat /etc/motd' \
+	>> /etc/bash.bashrc \
+	; echo "\
+||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||\n\
+|								\n\
+| Docker container running Ubuntu				\n\
+| with TensorFlow ${TF_BRANCH} optimized for CPU		\n\
+| with Intel(R) MKL						\n\
```
inc0: besides the point a bit, but can we re-use some existing Dockerfile and FROM it to avoid duplication?

inc0 (changes requested):

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: Good catch.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: Yes. The purpose of this container is to enable engineers to get a common build with common components to work at the command line. No intention of using Jupyter and web interface.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: See above. It's not the purpose of this container.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: Good point. I'll remove that check. It was used primarily with TF 1.3, which wasn't happy on ubuntu 16.04.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: Suggestions?

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
inc0: Well, usually this isn't done this way. If someone builds app they'll write their own Dockerfile, FROM this one and use COPY or ADD to put code inside. Editing anything inside container is considered antipattern as image is not easy to reproduce.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
inc0: Well if you use FROM original tensorflow image and just replace tensorflow libs with built ones you'll keep API of original container with significantly less code

gunan: I think this may need a larger refactoring, I would like to move to something like this for all TF containers eventually, but I think it is outside the scope of this change.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,127 @@
...
+    TF_NEED_MPI=0
+RUN ./configure; exit 0 # ./configure whines unnecessarily
+
+# Build and Install TensorFlow.
+RUN LD_LIBRARY_PATH=${LD_LIBRARY_PATH} \
+    bazel build --config=mkl \
+                --copt="-mfma" \
+                --copt="-mavx2" \
+		--config="opt" \
+               --copt="-march=native" \
```
gunan: There is a lot of redundancy here. I would say let's just have `-march=ivybridge`, we can have the rest maybe configurable through an ARG. There are a lot of machines without AVX2, and it would be nice if the default worked for most of the users.

gunan (changes requested):

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: True.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,127 @@
+FROM ubuntu
+
+LABEL maintainer="Clayne Robison<clayne.b.robison@intel.com>"
+
+# It is possible to override these for releases.
+ARG TF_BRANCH=r1.4
```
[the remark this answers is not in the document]
claynerobison: Not really. I'd rather use a TF verison I know works with MKL, which 1.4 does.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: Now that you mention it, when building with MKL, the runtime detects the platform it runs on and adapts, meaning these switches only make a difference when the path is EIGEN. I'll remove the redundancy, but I'd prefer arch=native because the compiler will detect what optimizations it can use for EIGEN paths.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -1,73 +1,39 @@
...
+# Use the specified Bazel version
 WORKDIR /
-RUN mkdir /bazel && \
-    cd /bazel && \
-    wget --quiet https://github.com/bazelbuild/bazel/releases/download/$BAZEL_VERSION/bazel-$BAZEL_VERSION-installer-linux-x86_64.sh && \
-    wget --quiet https://raw.githubusercontent.com/bazelbuild/bazel/master/LICENSE && \
+RUN cd /bazel && \
+    rm -rf * && \
+    curl -H "User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36" -fSsL -O https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh && \
+    curl -H "User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36" -fSsL -o /bazel/LICENSE.txt https://raw.githubusercontent.com/bazelbuild/bazel/master/LICENSE && \
```
inc0: why you need headers? looks strange...

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -1,73 +1,39 @@
...
-        unzip \
-        zip \
-        zlib1g-dev \
-        openjdk-8-jdk \
-        openjdk-8-jre-headless \
-        wget \
-	locate \
-	vim \
+        vim \
+        emacs \
```
inc0: while I still think we shouldn't have these, it's ok;)

inc0 (commented):
Just few nits, but I think it's good enough

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,96 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+RUN pip --no-cache-dir install --upgrade \
+        pip setuptools
+
+RUN pip --no-cache-dir install wheel 
+
+# Use the specified Bazel version
+WORKDIR /
+RUN cd /bazel && \
```
gunan: tensorflow:latest-devel should already have bazel installed.
Do we need to upgrade bazel for MKL?

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: Just using the same command from the Tensorflow Dockerfile.devel that this is based on: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docker/Dockerfile.devel#L64. Pulling a different rev of bazel.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: @gunan Do you want me to make a change here? It's not clear from your comment.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
claynerobison: I thought that we had problems building MKL with it, but I just verified that it's OK. I'll remove this and use the version that comes with the base tensorflow container.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
```diff
@@ -0,0 +1,85 @@
...
+
+RUN pip --no-cache-dir install wheel 
+
+# Download and build TensorFlow.
+WORKDIR /
+RUN rm -rf tensorflow && \
+    git clone https://github.com/tensorflow/tensorflow.git && \
+    cd tensorflow && \
+    git checkout ${TF_BRANCH}
+WORKDIR /tensorflow
```
gunan: I just noticed. our devel images may already have TF checked out in this location. Could you make sure this wont break? Just running docker build once with this dockerfile should be sufficient.

tensorflow/tools/docker/Dockerfile.devel-cpu-mkl
gunan: No need for change. This is something I will need to look at.

gunan (approved):

gunan:
Jenkins, test this please.

claynerobison:
These failures don't seem to be related to this pull request. What do I need to do to get this approved?

gunan:
The PR is already approved. and the new file should not be covered by any of the failing tests.
So I will go ahead and merge.

Issue #13558, opened 2017-10-07
segfaults in GPU tf.matrix_inverse


yaroslavvb:
I'm running into segfaults in tf.matrix_inverse
I'm adding identity*0.001 so matrices should be invertible, and same procedure works fine in numpy and in TensorFlow CPU version.

https://github.com/yaroslavvb/stuff/blob/master/inverse_segfault.py
`python inverse_segfault.py`

This non-deterministically crashes after 1-2 seconds with various backtraces.

IE

```
#0  0x0000000000000001 in ?? ()
#1  0x00007fe90ed9c652 in tensorflow::Tensor::TotalBytes() const ()
   from /home/yaroslav/anaconda3/envs/oct10/lib/python3.5/site-packages/tensorflow/python/../libtensorflow_framework.so
#2  0x00007fe90ed9c7d6 in tensorflow::Tensor::tensor_data() const ()
   from /home/yaroslav/anaconda3/envs/oct10/lib/python3.5/site-packages/tensorflow/python/../libtensorflow_framework.so
#3  0x00007fe9137adda3 in bool tensorflow::internal::TransposeUsingTile<unsigned int>(Eigen::GpuDevice const&, tensorflow::Tensor const&, tensorflow::gtl::ArraySlice<int>, tensorflow::Tensor*) ()
   from /home/yaroslav/anaconda3/envs/oct10/lib/python3.5/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so
#4  0x00007fe9137a696c in tensorflow::Status tensorflow::DoTranspose<Eigen::GpuDevice>(Eigen::GpuDevice const&, tensorflow::Tensor const&, tensorflow::gtl::ArraySlice<int>, tensorflow::Tensor*) ()
   from /home/yaroslav/anaconda3/envs/oct10/lib/python3.5/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so
#5  0x00007fe911ada0fd in tensorflow::SvdOpGpu<float>::PerformSVD_MgeqN(tensorflow::OpKernelContext*, std::function<void ()>, long long, long long, long long, tensorflow::gtl::ArraySlice<int> const&, tensorflow::Tensor const&, tensorflow::Tensor*, tensorflow::Tensor*, tensorflow::Tensor*) ()
   from /home/yaroslav/anaconda3/envs/oct10/lib/python3.5/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so
[...]

PR #14492, opened 2017-11-12
Add a mutex in cuda_solvers GetrfImpl.

codrut3:
This should solve issue #13558. Because NVIDIA does not publish the code, it is not possible to prove
the bug is in the cuSolver function. However, I have removed / commented code from
matrix_inverse_op.cpp until essentially only the call to solver->Getrf remained, and from @yaroslavvb' 
project mostly everything except the regularized_inverse calls, and the segmentation fault still
occurred. It did not appear when there was a single thread calling regularized_inverse. The segmentation fault always occurs without my change, and never after introducing the
lock (although that in itself does not mean the bug is solved).

Note that GetrsImpl is already protected with a lock, so apparently this issue was known.

Finally, @yaroslavvb has a second stacktrace which does not involve the Getrf function.
I could not reproduce this error, but it may be a different bug.


yaroslavvb:
Impressive investigation!

codrut3:
Hi @rmlarsen,

Please, would you have some time to look at this PR?

codrut3:
Only one test fails: //tensorflow/core:util_saved_tensor_slice_util_test. It does not seem related to my change (it doesn't even use CUDA). Also, I can't reproduce this error on my computer.

rmlarsen (approved):

rmlarsen:
@codrut3 I have sadly had to add similar mutexes due to races in cusolver. NVIDIA is aware of the problem and I believe the situation should be improved in Cuda9.

rmlarsen:

drpngx:
OK, we are moving to cuda9. could you update and test with the new version?

codrut3:
Hi @drpngx ,

Someone else added in the meantime the mutex, so the change is no longer needed. However, my PR also fixes a typo, at least this could merged. Because it's just a comment change, I don't think it requires testing.

drpngx:
We have to test everything. You never know when something is going to break.

Jenkins, test this please.

drpngx:
/CC @gunan Ran out of disk space on the Jenkins build here too.

```
devmapper: Thin Pool has 968455 free data blocks which is less than minimum required 983040 free data blocks. Create more free space in thin pool or use dm.min_free_space option to change behavior
```
Jenkins, test this please.

PR #14497, opened 2017-11-12
use with when calling TFRecordWriter

FirefoxMetzger:
TFRecordWriter supports __enter__ and __exit__ . 
Calling it through with is more pythonic and does cleanup in case something throws an uncaught exception.


qmick (approved):

FirefoxMetzger:
@qmick Anything I can do to progress the this PR?
It seems that the TF Test Suite never finishes.

qmick:
Don't worry about that. You just need to wait until a reviewer with write access  approve this PR, then he/she will start the test.

jhseu (approved):

PR #14499, opened 2017-11-12
fix assert_shallow_structure for dicts

boeddeker:
The function `tensorflow.python.data.util.nest.flatten_up_to` used in `tf.data.Dataset.from_generator` does not compare the dict keys (only the length of the dict)
```python
tf_tree = dict(A=tf.placeholder(tf.float32), B=tf.placeholder(tf.float64))
np_tree = dict(A=np.zeros([1], np.float32), C=np.zeros([2], np.float64))
nest.flatten_up_to(tf_tree, np_tree)  # Fails now
```
and ignores nested dict's (Bug, iterate over keys in `tensorflow.python.data.util.assert_shallow_structure`)
```python
tf_tree = dict(A=tf.placeholder(tf.float32), B=dict(C=tf.placeholder(tf.float64)))
np_tree = dict(A=np.zeros([1], np.float32), B=dict(D=np.zeros([2], np.float64)))
nest.flatten_up_to(tf_tree, np_tree)  # Fails now
```

This PR fix `tensorflow.python.data.util.assert_shallow_structure` that is used in `tensorflow.python.data.util.nest.flatten_up_to`, but there may also other function that have such a bug.

### DIFF
--- a/tensorflow/python/data/util/nest.py
+++ b/tensorflow/python/data/util/nest.py
@@ -367,6 +367,16 @@ def assert_shallow_structure(shallow_tree, input_tree, check_types=True):
           "structure has length %s, while shallow structure has length %s."
           % (len(input_tree), len(shallow_tree)))
 
+    if check_types and isinstance(shallow_tree, dict):
+      if set(input_tree) != set(shallow_tree):
+        raise ValueError(
+          "The two structures don't have the same keys. Input "
+          "structure has keys %s, while shallow structure has keys %s."
+          % (list(_six.iterkeys(input_tree)),
+             list(_six.iterkeys(shallow_tree))))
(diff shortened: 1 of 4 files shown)


mrry (changes requested):
Thanks for the fix! That `tf.data` code has a fork of some of the `nest.py` code that includes this bug. Could you add the same fix [here](https://github.com/tensorflow/tensorflow/blob/9089ab59826ce656b6f8026b08471ca6696cacf9/tensorflow/python/util/nest.py#L454)?

boeddeker:
Sure, I made the modification also in the other `nest.py`.
I fixed also a wrong key the `nest_test.py`.

mrry (approved):
Looks great... thanks for the contribution!

mrry:

